PR #6563, opened 2022-11-09
copier: Don't block reset during xruns

ranj063:
When an xrun happens, the pipeline is reset to allow for a clean D3 entry. So remove the check for state in copier_reset() to allow the copier component to be reset during xruns.

### DIFF
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -644,11 +644,6 @@ static int copier_reset(struct comp_dev *dev)
 
 	comp_dbg(dev, "copier_reset()");
 
-	if (dev->state == COMP_STATE_ACTIVE) {
-		comp_info(dev, "copier_config(): Component is in active state. Ignore resetting");
-		return 0;
-	}
-
 	cd->input_total_data_processed = 0;
 	cd->output_total_data_processed = 0;
 


src/audio/copier/copier.c
```diff
@@ -644,11 +644,6 @@ static int copier_reset(struct comp_dev *dev)
 
 	comp_dbg(dev, "copier_reset()");
 
-	if (dev->state == COMP_STATE_ACTIVE) {
-		comp_info(dev, "copier_config(): Component is in active state. Ignore resetting");
-		return 0;
-	}
-
```
kv2019i, line 651: @RanderWang Can you review? This was part of your original e1dbb5b4fc04a commit.

kv2019i (approved):
We basicly now have two usages of reset. Reset after pipeline stop is completed, and recent addition of xrun handling. I guess the question is, from component interface perspective, is the ACTIVE->RESET transition allowed or not. We do have now pipeline_xrun_recover() which relies on this, so if we keep this, this patch is mandatory.

RanderWang (approved):

iuliana-prodan (approved):

lgirdwood (approved):

PR #6564, opened 2022-11-10
Ssp virtual dai

brentlu:
Rename SSP stream names in topology file which will be linked to DAI Links in machine driver side.


brentlu:
Linux side: https://github.com/thesofproject/linux/pull/4001

plbossart (changes requested):
That's a NAK from me @brentlu, as I mentioned in the kernel PR this is not a backwards-compatible change. An update of the kernel would need to be synchronized with a topology update, and that's a big no-no

PR #6566, opened 2022-11-10
rtnr: Fix log formatting for unsigned parameters

MingJenTai:
This PR fixes the format for logging unsigned parameters.

### DIFF
--- a/src/audio/rtnr/rtnr.c
+++ b/src/audio/rtnr/rtnr.c
@@ -527,7 +527,7 @@ static int rtnr_set_config_bytes(struct comp_dev *dev,
 	 * the whole config data is received.
 	 */
 	if (size < sizeof(cd->config)) {
-		comp_err(dev, "rtnr_set_config_data(): invalid size %d",
+		comp_err(dev, "rtnr_set_config_data(): invalid size %u",
 			 size);
 		return -EINVAL;
 	}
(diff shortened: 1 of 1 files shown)


marc-hb:
Which toolchain found this?

MingJenTai:
> Which toolchain found this?

Hi @yaochunhung, are you able to answer this? Thanks!

yaochunhung:
@MingJenTai I just think it should be %u for unsigned int, not found any issue on any toolchian. thanks.

marc-hb:
I'm not sure this change is a good idea.

As the C language has no `None` special value, sometimes people use negative values as error codes. Example: `unsigned size = -1`. In that case it's much more readable to display these negatives values as `-1` or `-2`; especially in an error message like here.

Of course it would be better to change the type and switch to `int size = -1` to formerly support error values without trying to understand C's impenetrable conversion rules but that may require many more changes all across the code.

yaochunhung:
Thanks. I know your concern now. But I have a little bit confusion because I see there is example used uint32_t  and print %u,
https://github.com/thesofproject/sof/blob/main/src/audio/buffer.c#L38

lgirdwood:
@MingJenTai I assume both are unsigned types then this is good for me. Fwiw, Zephyr build can be strict here, but I dont think we have tried RTNR with Zephyr config yet.

marc-hb:
> But I have a little bit confusion because I see there is example 

I'm pretty sure you can find an example for anything.

MingJenTai:
> @MingJenTai I assume both are unsigned types then this is good for me. Fwiw, Zephyr build can be strict here, but I dont think we have tried RTNR with Zephyr config yet.

Yes both are unsigned types. Not sure about what's the common formatting of unsigned variables should be used here but at least the compiler will not report warning/error about this.

lgirdwood (approved):

PR #6567, opened 2022-11-10
topology1:renoir: add DTS + EQ topology

joechengxperi:
Add a new m4 to support DTS SDK -> EQ processing

### DIFF
--- a/tools/topology/topology1/sof-rn-rt5682-rt1019.m4
+++ b/tools/topology/topology1/sof-rn-rt5682-rt1019.m4
@@ -21,7 +21,7 @@ include(`platform/amd/acp.m4')
 # Playback pipeline 1 on PCM 0 using max 2 channels of s16le.
 # Schedule 96 frames per 2000us deadline on core 0 with priority 0
 PIPELINE_PCM_ADD(
-ifdef(`DTS', sof/pipe-eq-iir-dts-codec-playback.m4, sof/pipe-passthrough-playback.m4),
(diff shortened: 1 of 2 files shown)


tools/topology/topology1/sof-rn-rt5682-rt1019.m4
```diff
@@ -21,7 +21,7 @@ include(`platform/amd/acp.m4')
 # Playback pipeline 1 on PCM 0 using max 2 channels of s16le.
 # Schedule 96 frames per 2000us deadline on core 0 with priority 0
 PIPELINE_PCM_ADD(
-ifdef(`DTS', sof/pipe-eq-iir-dts-codec-playback.m4, sof/pipe-passthrough-playback.m4),
+ifdef(`DTS', sof/pipe-dts-codec-eq-iir-playback.m4, sof/pipe-passthrough-playback.m4),
```
johnylin76, line 24: Will "DTS before EQIIR" be the common case (or even more native in algorithm)?
If yes, we should land on main first.

johnylin76 (approved):
LGTM with one question

tools/topology/topology1/sof-rn-rt5682-rt1019.m4
joechengxperi: No, this change is for Dewatt to align the CRAS tuning result. I'll still land this commit on main for consistency.

PR #6580, opened 2022-11-11
[SKIP CI][test only] use same zephyr snapshot as mtl-002

mengdonglin:
Please don't merge. This is only to compare test result between SOF main branch and mtl-002-drop-stable.


mengdonglin:
I'll wait for upstream Zephyr updates for SOF main branch. This PR is not needed.

PR #6583, opened 2022-11-13
topology2: fix audio format bit depth in deepbuffer copier

yongzhi1:
Fix out_valid_bit_depth for 16-bit and 24-bit audio formats.

### DIFF
--- a/tools/topology/topology2/include/pipelines/cavs/deepbuffer-playback.conf
+++ b/tools/topology/topology2/include/pipelines/cavs/deepbuffer-playback.conf
@@ -53,7 +53,7 @@ Class.Pipeline."deepbuffer-playback" {
 			# 16-bit 48KHz 2ch
 			Object.Base.audio_format.1 {
 				out_bit_depth		32
-				out_valid_bit_depth	24
+				out_valid_bit_depth	32
 				# 100ms buffer
 				dma_buffer_size "$[$ibs * $DEEPBUFFER_FW_DMA_MS]"
 			}
@@ -62,7 +62,7 @@ Class.Pipeline."deepbuffer-playback" {
 				in_bit_depth		32
 				in_valid_bit_depth	24
 				out_bit_depth		32
-				out_valid_bit_depth	24
+				out_valid_bit_depth	32
 				# 100ms buffer
 				dma_buffer_size "$[$obs * $DEEPBUFFER_FW_DMA_MS]"
 			}


lgirdwood (approved):
@RanderWang @ranj063 pls review

RanderWang (commented):
why use  32bit valid sample size ? Internal pipeline uses 24bit valid bits to do calculation.

ranj063:
> why use 32bit valid sample size ? Internal pipeline uses 24bit valid bits to do calculation.

but this change is needed to make the format matching to work in the kernel isnt it?

RanderWang:
why use  32bit valid sample size ? Internal pipeline uses 24bit valid bits to do calculation.

> > why use 32bit valid sample size ? Internal pipeline uses 24bit valid bits to do calculation.
> 
> but this change is needed to make the format matching to work in the kernel isnt it?

yes. @yongzhi1  What's the problem now ?

yongzhi1:
Hi @RanderWang , we observed no sound coming from "Deepbuffer Amps" when playing 16-bit material.

ranj063:
> Hi @RanderWang , we observed no sound coming from "Deepbuffer Amps" when playing 16-bit material.

@yongzhi1 it would be helpful if you can file a bug that this PR fixes so we can see why this is needed

yongzhi1:
> @yongzhi1 it would be helpful if you can file a bug that this PR fixes so we can see why this is needed

Sure, will add the PR # here.

yongzhi1:
The issue seems to be deepbuffer-playback only supports  $HDA_HOST_OUTPUT_CLASS node_type, which works fine for HDA and SDW but not I2S.

plbossart:
> The issue seems to be deepbuffer-playback pipeline only supports $HDA_HOST_OUTPUT_CLASS node_type, which works fine for HDA and SDW but not I2S.

That seems very odd @yongzhi1, @RanderWang just added the deep buffer for the nocodec topology which does use I2S/SSP.

RanderWang:
> > The issue seems to be deepbuffer-playback pipeline only supports $HDA_HOST_OUTPUT_CLASS node_type, which works fine for HDA and SDW but not I2S.
> 
> That seems very odd @yongzhi1, @RanderWang just added the deep buffer for the nocodec topology which does use I2S/SSP.

@plbossart  yong is doing experiment and pick my change to his private branch

yongzhi1:
Learned a clever way to do it from @RanderWang , updated top level conf instead https://github.com/yongzhi1/sof/commit/09966aa13d16e3baaad5c0857e75f1ba1010f927, so no need to hard-code the changes here, thanks all for reviewing the changes.

PR #6284, opened 2022-09-09
platform: Add suport for a new intel Meteorlake platform

softwarecki:
This PR adds support for the new intel Meteorlake platform. It can be merged after switching to a new Zephyr version with IPC driver.

PR #6586, opened 2022-11-14
module_adapter:dts: support module life cycle change

joechengxperi:
1. Add a free memory function to support PR#6230 and PR#6331, the latest prepare/reset API flow update of module_adapter.
2. This requires DTS library v1.0.7 to support this change.
3. This is the backport from PR #6587


lgirdwood (commented):
@joechengxperi is this a backport of a patch in main ?

markbartonxperi (approved):

lgirdwood:
@thesofproject/amd pls review/merge

joechengxperi:
@johnylin76 @balakishorepati could you please review this PR? Thanks.

johnylin76:
Hi Joe,

I just uploaded https://github.com/thesofproject/sof/pull/6824 which also includes your commits. We should land them after the module_adapter general commits.

I also tested them on device with the latest SDK. However it will fail on writing config blobs which breaks the whole audio. I have blurred memory that you ever mentioned the config blob format may require to change from the existing one. Could that be the root cause?

joechengxperi:
> Hi Joe,
> 
> I just uploaded #6824 which also includes your commits. We should land them after the module_adapter general commits.
> 
> I also tested them on device with the latest SDK. However it will fail on writing config blobs which breaks the whole audio. I have blurred memory that you ever mentioned the config blob format may require to change from the existing one. Could that be the root cause?

Hi Johny,

Yes, for #6824, there is one value required to change in the dts_codec_adapater.m4 to avoid the crash with latest SDK. I found it last week when testing rpl-001-drop-stable and haven't committed it yet.

johnylin76:
> 



> > Hi Joe,
> > I just uploaded #6824 which also includes your commits. We should land them after the module_adapter general commits.
> > I also tested them on device with the latest SDK. However it will fail on writing config blobs which breaks the whole audio. I have blurred memory that you ever mentioned the config blob format may require to change from the existing one. Could that be the root cause?
> 
> Hi Johny,
> 
> Yes, for #6824, there is one value required to change in the dts_codec_adapater.m4 to avoid the crash with latest SDK. I found it last week when testing rpl-001-drop-stable and haven't committed it yet.

Do you mean the value change for CA_SETUP_CONTROLBYTES (default config) in m4/dts_codec_adapater.m4? Does dts_sdk_*.bin require the change as well?

johnylin76:
> > > > Hi Joe,
> > > > I just uploaded #6824 which also includes your commits. We should land them after the module_adapter general commits.
> > > > I also tested them on device with the latest SDK. However it will fail on writing config blobs which breaks the whole audio. I have blurred memory that you ever mentioned the config blob format may require to change from the existing one. Could that be the root cause?
> > > 
> > > 
> > > Hi Johny,
> > > Yes, for #6824, there is one value required to change in the dts_codec_adapater.m4 to avoid the crash with latest SDK. I found it last week when testing rpl-001-drop-stable and haven't committed it yet.
> > 
> > 
> > Do you mean the value change for CA_SETUP_CONTROLBYTES (default config) in m4/dts_codec_adapater.m4? Does dts_sdk_*.bin require the change as well?
> > By the way, is it reproducible by DTS uprev on mt8195/v0.4 branch?
> 
> For the updated dts_sdk_*.bin, it's already addressed. I found the issue on rpl-001-drop-stable branch only since the commit [5bebd28](https://github.com/thesofproject/sof/commit/5bebd287554ff7aa87828917b049ab94310ddc0e) is only merged in this branch, not in other production branches.

I see. Here is another clue. Previously I have verified DTS on rpl-001 build (with the original DTS SDK, without  https://github.com/thesofproject/sof/pull/6824) and it looks good.

johnylin76:
> > > > > Hi Joe,
> > > > > I just uploaded #6824 which also includes your commits. We should land them after the module_adapter general commits.
> > > > > I also tested them on device with the latest SDK. However it will fail on writing config blobs which breaks the whole audio. I have blurred memory that you ever mentioned the config blob format may require to change from the existing one. Could that be the root cause?
> > > > 
> > > > 
> > > > Hi Johny,
> > > > Yes, for #6824, there is one value required to change in the dts_codec_adapater.m4 to avoid the crash with latest SDK. I found it last week when testing rpl-001-drop-stable and haven't committed it yet.
> > > 
> > > 
> > > Do you mean the value change for CA_SETUP_CONTROLBYTES (default config) in m4/dts_codec_adapater.m4? Does dts_sdk_*.bin require the change as well?
> > > By the way, is it reproducible by DTS uprev on mt8195/v0.4 branch?
> > 
> > 
> > For the updated dts_sdk_*.bin, it's already addressed. I found the issue on rpl-001-drop-stable branch only since the commit [5bebd28](https://github.com/thesofproject/sof/commit/5bebd287554ff7aa87828917b049ab94310ddc0e) is only merged in this branch, not in other production branches.
> 
> I see. Here is another clue. Previously I have verified DTS on rpl-001 build (with the original DTS SDK, without #6824) and it looks good.
> 
> One thing should be noted as well is that [5bebd28](https://github.com/thesofproject/sof/commit/5bebd287554ff7aa87828917b049ab94310ddc0e) removed "DTS Codec Runtime" control, so you need to send config blob to "DTS Codec Setup" instead.

I suspect the root cause is that "param->id = 0" is negligible for original SDK, but is meaningful for v1.0.8 SDK (perhaps?)
After [5bebd28](https://github.com/thesofproject/sof/commit/5bebd287554ff7aa87828917b049ab94310ddc0e) `CA_SETUP_CONTROLBYTES` will set a 3-u32 param {0x0, 0xc, 0x0} to SDK. As I know in dts_sdk_*.bin the param uses id=1 {0x1, 0x2c0, ....}

joechengxperi:
> 
Yes, you're right. But I'm not sure that you mentioned it was good without #6284, I might need to check it.

johnylin76 (approved):

PR #6587, opened 2022-11-14
module_adapter:dts: support module life cycle change

joechengxperi:
1. Add a free memory function to support PR#6330 and PR#6331, the latest prepare/reset API flow update of module_adapter.
2. This requires DTS library v1.0.7 to support this change.


lgirdwood (approved):
@joechengxperi best to get this merged first before merging with production branch.
@ranj063 good for you ?

src/audio/module_adapter/module/dts.c
```diff
@@ -334,16 +346,16 @@ static int dts_codec_apply_config(struct processing_module *mod)
...
 			if (param_data_size) {
 				dts_result = dtsSofInterfaceApplyConfig(codec->private, param->id,
 					param->data, param_data_size);
-				ret = dts_effect_convert_sof_interface_result(dev, dts_result);
-				if (ret) {
+				if (dts_result != DTS_SOF_INTERFACE_RESULT_SUCCESS) {
 					comp_err(dev,
 						"dts_codec_apply_config() dtsSofInterfaceApplyConfig failed %d",
 						dts_result);
-					return ret;
```
ranj063, line 346: @joechengxperi how is this change related to the addition of the free memory callback?

src/audio/module_adapter/module/dts.c
```diff
@@ -116,8 +131,6 @@ static int dts_effect_populate_buffer_configuration(struct comp_dev *dev,
 	buffer_config->sampleRate = rate;
 	buffer_config->numChannels = channels;
 	buffer_config->periodInFrames = dev->frames;
-	/* totalBufferLengthInBytes will be populated in dtsSofInterfacePrepare */
-	buffer_config->totalBufferLengthInBytes = 0;
```
ranj063, line 120: same here? Is this related to the free memory callback?

markbartonxperi (approved):

johnylin76:
+1 for comments from @ranj063, they should be separated into 2 commits
1. Required code update for uprev DTS library v1.0.7
2. Fix to support new reset/prepare module_adapter API

The change of current commit is more like bullet#1 to me.
In some extent bullet#1 may fix the module_adapter API (bullet#2) by adding the free_memory interface.
Since free_memory is called from DTS lib so we don't see it in the wrapper code. Perhaps adding some comments in code to help reviewer to understand.

I would like to confirm one more thing, which kind(s) of memory will be freed while module reset? data_buffer, config_buffer, or both?
When is config_buffer allocated, on apply_config()? (config_buffer here means the internal buffer for DTS lib to store the config data)
A key point is, there is no guarantee that apply_config() will be called in the duration from module reset to next-time prepare. If config_buffer is freed on reset, then there comes the problem

src/audio/module_adapter/module/dts.c
johnylin76: you can treat it as code update for library uprev, see my comment https://github.com/thesofproject/sof/pull/6587#issuecomment-1314793068

ranj063: > No, this line is not related to the memory callback. We've found it's redundant here and removed it. Should we separate it to another commit or PR?

yes, please

joechengxperi:
> +1 for comments from @ranj063, they should be separated into 2 commits
> 
> 1. Required code update for uprev DTS library v1.0.7
> 2. Fix to support new reset/prepare module_adapter API
> 
> The change of current commit is more like bullet#1 to me. In some extent bullet#1 may fix the module_adapter API (bullet#2) by adding the free_memory interface. Since free_memory is called from DTS lib so we don't see it in the wrapper code. Perhaps adding some comments in code to help reviewer to understand.
> 
> I would like to confirm one more thing, which kind(s) of memory will be freed while module reset? data_buffer, config_buffer, or both? When is config_buffer allocated, on apply_config()? (config_buffer here means the internal buffer for DTS lib to store the config data) A key point is, there is no guarantee that apply_config() will be called in the duration from module reset to next-time prepare. If config_buffer is freed on reset, then there comes the problem



> +1 for comments from @ranj063, they should be separated into 2 commits
> 
> 1. Required code update for uprev DTS library v1.0.7
> 2. Fix to support new reset/prepare module_adapter API
> 
> The change of current commit is more like bullet#1 to me. In some extent bullet#1 may fix the module_adapter API (bullet#2) by adding the free_memory interface. Since free_memory is called from DTS lib so we don't see it in the wrapper code. Perhaps adding some comments in code to help reviewer to understand.
> 
> I would like to confirm one more thing, which kind(s) of memory will be freed while module reset? data_buffer, config_buffer, or both? When is config_buffer allocated, on apply_config()? (config_buffer here means the internal buffer for DTS lib to store the config data) A key point is, there is no guarantee that apply_config() will be called in the duration from module reset to next-time prepare. If config_buffer is freed on reset, then there comes the problem

It's actually used in prepare(). With the module reset change, the different buffer configurations could be propagated to 3rd party module now. If this happens, the SDK instance or data buffers may be freed and reallocated according to new configuration. Current configuration won't be affected as there is local copy in the SDK.

johnylin76:
> > +1 for comments from @ranj063, they should be separated into 2 commits
> > 
> > 1. Required code update for uprev DTS library v1.0.7
> > 2. Fix to support new reset/prepare module_adapter API
> > 
> > The change of current commit is more like bullet#1 to me. In some extent bullet#1 may fix the module_adapter API (bullet#2) by adding the free_memory interface. Since free_memory is called from DTS lib so we don't see it in the wrapper code. Perhaps adding some comments in code to help reviewer to understand.
> > I would like to confirm one more thing, which kind(s) of memory will be freed while module reset? data_buffer, config_buffer, or both? When is config_buffer allocated, on apply_config()? (config_buffer here means the internal buffer for DTS lib to store the config data) A key point is, there is no guarantee that apply_config() will be called in the duration from module reset to next-time prepare. If config_buffer is freed on reset, then there comes the problem
> 
> It's actually used in prepare(). With the module reset change, the different buffer configurations could be propagated to 3rd party module now. If this happens, the SDK instance or data buffers may be freed and reallocated according to new configuration. Current configuration won't be affected as there is local copy in the SDK.

Looks good to me. So it implies that data buffer is retained when the required size is unchanged, right? Would the residual contents in data buffer be cleaned up?

joechengxperi:
> Looks good to me. So it implies that data buffer is retained when the required size is unchanged, right? Would the residual contents in data buffer be cleaned up for such cases? (just a reminder. It doesn't have to be cleaned up, depending on the usage for processing underneath.

Yes, data buffer won't be changed if there is nothing changed. The residual contents in data buffer would be flushed in module reset.

src/audio/module_adapter/module/dts.c
joechengxperi: @ranj063 I've separated it to another commit. Please let me know if it's good now. Thanks.

src/audio/module_adapter/module/dts.c
```diff
@@ -350,15 +347,16 @@ static int dts_codec_apply_config(struct processing_module *mod)
...
+
 			if (param_data_size) {
 				dts_result = dtsSofInterfaceApplyConfig(codec->private, param->id,
 					param->data, param_data_size);
-				ret = dts_effect_convert_sof_interface_result(dev, dts_result);
-				if (ret) {
+				if (dts_result != DTS_SOF_INTERFACE_RESULT_SUCCESS) {
 					comp_err(dev,
 						"dts_codec_apply_config() dtsSofInterfaceApplyConfig failed %d",
 						dts_result);
```
johnylin76, line 359: why is it changed to suppress error on apply config?

markbartonxperi: As we've integrated against various platforms, we've encountered a few occurrences where invalid configuration data is sent at startup, which we would detect and report to the module adapter, which then results in the DSP not starting, causing delays in the integration effort.
I think it's better here for us to absorb the error and create a log message.

johnylin76: I believe the invalid config at startup is caused by the outdated topology which still have setup and runtime config controls for module_adapter. The module_adapter c code had removed setup config, so those bytes fed by old topology will be passed here and causes errors.

markbartonxperi: Yes. This is one of the scenarios. And I think a good example of the motivation for this change.
As the module adapter continues to change shape on the mainstream branch and those changes get distributed to some of the per-platform branches, I think we (DTS) struggle to understand when, where and, sometimes, how we should make corresponding changes to our topology files. This change isolates us from that - we now only need to confirm that our configuration blob delivered via UCM is applied successfully, and know that all other blobs we don't care about will be quietly rejected is they aren't suitable

johnylin76: I beg to differ. It is indeed a hard task to maintain the module working on multiple production branches. We are taking progress on migrating the elder branch (e.g. adl-004-drop-stable) to more up-to-date one (rpl-001-drop-stable) to alleviate the code difference from main. Even though the varied designs are required on different branches, we should still keep the correct one on mainstream.

From another aspect, the suppressed failure could be a potential landmine which causes the fatal error at the latter timestamp. It will make debugging more difficult because the SOF log is not achievable from the normal (official) OS images (we need to enable debug_trace feature in kernel config and build/deploy kernel). We should raise the error promptly unless we have fair confidence the error is harmless.

joechengxperi: @johnylin76 
I've already reverted this part to previous one. Please review. Thanks.

johnylin76 (approved):

lgirdwood (approved):
@ranj063 good for you ?

markbartonxperi (approved):

lgirdwood:
@ranj063 ping

johnylin76:
@ranj063 ping.

Hi @lgirdwood , since the change in this PR is only inside DTS module code. Could we merge it without the approval from @ranj063 (due to long vacation perhaps?)

PR #6590, opened 2022-11-14
sparse: preserve build output in sparse CI test

lyakh:
Build output is needed to actually be able to fix warnings and errors.

### DIFF
--- a/scripts/parse_sparse_output.sh
+++ b/scripts/parse_sparse_output.sh
@@ -13,8 +13,6 @@
 
 main()
 {
-    >&2 printf 'Reminder: to see ALL warnings you must as usual build _from scratch_\n'
-
     # To reproduce an 'error: ' and test this script try commenting out
     # `defined(__CHECKER__)` in common.h.
     #
@@ -22,10 +20,13 @@ main()
     # script try deleting a __sparse_cache annotation like the one in
     # src/audio/mixer/mixer.c
 
-    ! grep -i  \
-        -e '[[:space:]]error:[[:space:]]'  \
-        -e '[[:space:]]warning:[[:space:]].*different address space' \
-
+    ret=0
+    while read -r line; do
+        echo "$line" | grep -q \
+            -e 'error:[[:space:]]\|warning:[[:space:]].*different address spaces' && ret=1
+        printf '%s\n' "$line"
+    done
+    return $ret
 }
 
 main "$@"


scripts/parse_sparse_output.sh
```diff
@@ -13,19 +13,20 @@
...
 
-    ! grep -i  \
-        -e '[[:space:]]error:[[:space:]]'  \
-        -e '[[:space:]]warning:[[:space:]].*different address space' \
-
+    ret=0
+    while read line; do
+        echo "$line" | grep -q \
+            -e 'error:[[:space:]]\|warning:[[:space:]].*different address spaces' && ret=1
+        echo "$line"
```
marc-hb, line 27: Nit: `printf` is more portable
```suggestion
        printf '%s\n' "$line"
```

scripts/parse_sparse_output.sh
```diff
@@ -13,19 +13,20 @@
...
     # To reproduce the 'different address space' warning and test this
     # script try deleting a __sparse_cache annotation like the one in
     # src/audio/mixer/mixer.c
 
-    ! grep -i  \
-        -e '[[:space:]]error:[[:space:]]'  \
-        -e '[[:space:]]warning:[[:space:]].*different address space' \
-
+    ret=0
+    while read line; do
```
marc-hb, line 24:
```suggestion
    while read -r line; do
```

marc-hb (changes requested):
I don't think this is a good idea yet because of the very large number of current warnings. When the build fails users will see a very large output with many warnings and they will not know which warnings are considered more important and must be fixed.

To see the build output, simply don't use the script. That's more user-friendly than checking the exit status.

lyakh:
> I don't think this is a good idea yet because of the very large number of current warnings. When the build fails users will see a very large output with many warnings and they will not know which warnings are considered more important and must be fixed to get the build passing.
> 
> To see the full build output, simply don't use this filter script. That's more user-friendly than checking the exit status.

@marc-hb CI isn't only useful for giving the result - pass or fail - but also for showing what passed or failed. I don't think I have to advocate for that. You can easily imagine what would be if our on-device failures would simply say: "failed. Run locally to see the error."
But I'm open to improvements. I would add text in the failure case explaining what should be fixed. We could also add shell escapes to mark those lines red, but I don't think that would be a good idea.

lyakh:
@marc-hb and now you see why we need output: https://github.com/thesofproject/sof/actions/runs/3463308747/jobs/5783358076

marc-hb:
> CI isn't only useful for giving the result - pass or fail - but also for showing what passed or failed
> You can easily imagine what would be if our on-device failures would simply say: "failed. Run locally to see the error."

I think we're talking across each other. The current script shows in CI the sparse "warnings-treated-as-errors", only those and all those. So your second sentence does not make sense to me. Did you use the current filter script much?

Other, non-sparse checks typically show more logs but they can because:
- unlike sparse, other checks label "warnings=PASS" and "errors=FAIL" correctly
- unlike this check, other checks stop on the first error. When a build fails everyone looks only at the last few lines. No one scrolls up for pages and pages looking for an ERROR. Even test frameworks (!= builds) print a summary at the end.

>  I would add text in the failure case explaining what should be fixed. 

This would indeed make a huge difference and address almost all my concerns. However I've seen situations like this before and many people get confused when the build error(s) that matter are not in the last few lines and they have to scroll up.

How about this simple variation to highlight the warnings-treated-as-error: compile with sparse TWICE in .github/workflows. (compilation does not take that long)
- Do not use this filter the first time. Full logs.
- Compile a second time with the existing filter showing only the warnings-treated-as-error **at the end**.

marc-hb:
> @marc-hb and now you see why we need output: https://github.com/thesofproject/sof/actions/runs/3463308747/jobs/5783358076

I don't understand sorry. This page looks like a good example of what not to do because the errors are not at the end and very difficult to find.

lyakh:
> > @marc-hb and now you see why we need output: https://github.com/thesofproject/sof/actions/runs/3463308747/jobs/5783358076
> 
> I don't understand sorry. This page looks like a good example of what not to do because the errors are not at the end and very difficult to find.

@marc-hb what I see in that output is a lot of
```
/zep_workspace/zephyr/include/zephyr/toolchain.h:139:4: error: "Kconfig/toolchain endianness mismatch:"
/zep_workspace/zephyr/include/zephyr/toolchain.h:148:8: error: "CONFIG_LITTLE_ENDIAN but __ORDER_BIG_ENDIAN__"
```
errors and not a single relevant warning or error, which means the sparse run isn't configured correctly at all, doesn't it?

lyakh:
@marc-hb for an example compare to https://github.com/thesofproject/sof/actions/runs/3360977294/jobs/5570808747 - there we do have a correct sparse run and relevant warnings / errors:
```
/zep_workspace/zephyr/soc/xtensa/intel_adsp/cavs/multiprocessing.c:73:25: warning: incorrect type in argument 1 (different address spaces)
/zep_workspace/zephyr/soc/xtensa/intel_adsp/cavs/multiprocessing.c:73:25:    expected void __cache *ptr
/zep_workspace/zephyr/soc/xtensa/intel_adsp/cavs/multiprocessing.c:73:25:    got void *
```
Also, I'd like to see the complete sparse output because we actually want to fix all other warnings like
```
/zep_workspace/zephyr/soc/xtensa/intel_adsp/cavs/power.c:25:1: warning: symbol 'log_const_soc' was not declared. Should it be static?
```
too

lyakh:
> * Do not use this filter the first time. Full logs. Exit status 0 because sparse does not know any better.
> 
> * Compile a second time with the existing filter showing only the warnings-treated-as-error **at the end**. Exit status 1 if any.

@marc-hb Personally I'd find it a bit confusing seeing (warning / error) output twice, and even though it is relatively fast, it's still somewhere around 30 seconds, I'd say. Couldn't we use `tee` to save output in a temporary file and then grep it again for the second output? It would certainly be better than not having the output at all, I could live with this.

marc-hb:
> 



> @marc-hb what I see in that output is a lot of
> 
> ```
> /zep_workspace/zephyr/include/zephyr/toolchain.h:139:4: error: "Kconfig/toolchain endianness mismatch:"
> /zep_workspace/zephyr/include/zephyr/toolchain.h:148:8: error: "CONFIG_LITTLE_ENDIAN but __ORDER_BIG_ENDIAN__"
> ```
> 
> a lot of error: "Kconfig/toolchain endianness mismatch:"
> errors and not a single relevant warning or error, which means the sparse run isn't configured correctly at all, doesn't it?

This is a recent, sparse-specific problem and because it is so noisy it must be fixed first before this PR or any other sparse-related PR. If this cannot be fixed on the sparse side because of some sparse limitation then this new check can at least be disabled in Zephyr when sparse is used.

marc-hb:
> Also, I'd like to see the complete sparse output because we actually want to fix all other warnings like

s/we/you. I think I understand the huge value sparse brings to the table but for most people it's just another _unknown_ and annoying CI check that gets in the way of them getting their job done. We cannot let this vast majority struggle to make the difference between the 3 or 4 sparse-warnings-currently-treated-as-error versus the other, hundreds of sparse warnings-PASS. People must be given extremely clear pointers to what regressions in their PR they need to fix to paint CI green, otherwise they will simply start ignoring sparse. 

Of course we can and should gradually expand the list of sparse-warnings-currently-treated-as-error but that's a different topic.

Sparse experts like you who want to fix the sparse warnings-PASS (in order to expand the list) really don't need to look at CI results because they're already running sparse locally. CI results must always be designed for and readable by the masses, not for the rare experts in this or that tool.

marc-hb:
> Couldn't we use tee to save output in a temporary file and then grep it again for the second output?

Sounds good to me.

lyakh:
> > 
> 
> > @marc-hb what I see in that output is a lot of
> > ```
> > /zep_workspace/zephyr/include/zephyr/toolchain.h:139:4: error: "Kconfig/toolchain endianness mismatch:"
> > /zep_workspace/zephyr/include/zephyr/toolchain.h:148:8: error: "CONFIG_LITTLE_ENDIAN but __ORDER_BIG_ENDIAN__"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > errors and not a single relevant warning or error, which means the sparse run isn't configured correctly at all, doesn't it?
> 
> This is a recent, sparse-specific problem and because it is so noisy it must be fixed first before this PR or any other sparse-related PR. If this cannot be fixed on the sparse side because of some sparse limitation then this new check can at least be disabled in Zephyr when sparse is used.

@marc-hb not sure I understand correctly what you're saying. Are you saying, that this is "just" an additional error now printed by the new version of sparse and otherwise it functions as before or that it changed now and actually stopped working correctly with our Xtensa Zephyr builds? It seems like the latter to me. For example comparing those two runs that I referenced above - the old "all green" one prints for cbprintf_packaged.c were
```
[34/304] Building C object zephyr/CMakeFiles/zephyr.dir/lib/os/cbprintf_packaged.c.obj
/zep_workspace/zephyr/lib/os/cbprintf_packaged.c:18:1: warning: symbol 'log_const_cbprintf_package' was not declared. Should it be static?
/zep_workspace/zephyr/lib/os/cbprintf_packaged.c:881:6: warning: symbol 'is_ptr' was not declared. Should it be static?
/zep_workspace/zephyr/lib/os/cbprintf_packaged.c:1003:33: warning: dubious: x & !y
/zep_workspace/zephyr/lib/os/cbprintf_packaged.c:1003:33: warning: Variable length array is used.
/zep_workspace/zephyr/lib/os/cbprintf_packaged.c:1003:33: warning: Variable length array is used.
```
and now it's
```
[34/306] Building C object zephyr/CMakeFiles/zephyr.dir/lib/os/cbprintf_packaged.c.obj
/zep_workspace/zephyr/lib/os/cbprintf_packaged.c: note: in included file (through /zep_workspace/zephyr/include/zephyr/sys/errno_private.h, /zep_workspace/zephyr/lib/libc/minimal/include/errno.h):
/zep_workspace/zephyr/include/zephyr/toolchain.h:139:4: error: "Kconfig/toolchain endianness mismatch:"
/zep_workspace/zephyr/include/zephyr/toolchain.h:148:8: error: "CONFIG_LITTLE_ENDIAN but __ORDER_BIG_ENDIAN__"
```
I'll see if I can make a test PR to revert those un-grening changes and see if the output recovers but logically - yes, would seem unlikely to me that manipulating output should change the actual execution.

lgirdwood:
@lyakh @marc-hb some developers wont have sparse locally installed or configured for their OS. My expectation is that the CI report should provide all the information needed to fix the errors. Additionally the CI sparse output is also useful when compared to local sparse output as a means of improving and upgrading our sparse integration.

lyakh:
@marc-hb I think I know why https://git.kernel.org/pub/scm/devel/sparse/sparse.git/tree/target-xtensa.c#n25
```
const struct target target_xtensa = {
	.mach = MACH_XTENSA,
	.bitness = ARCH_LP32,
	.big_endian = true,
```

lyakh:
https://marc.info/?l=linux-sparse&m=166861736724543&w=2

marc-hb:
> This is a recent, sparse-specific problem...

It's recent because of a recently added check in upstream Zephyr, not because anything changed recently in sparse or SOF. Sorry I assumed everybody knew that.

marc-hb:
> https://marc.info/?l=linux-sparse&m=166861736724543&w=2

@lyah should we maintain a "under review" sparse branch in say http://github.com/thesofproject/sparse  ? This would help with handling errors, warning and exit codes too. Building sparse is quick and it seems to have very few dependencies, it may be easy to build it on the fly for every PR and break away from the current Zephyr image dependency. I can give that a try.

lyakh:
> > https://marc.info/?l=linux-sparse&m=166861736724543&w=2
> 
> @lyah should we maintain a "under review" sparse branch in say [http://github.com/thesofproject/sparse](https://github.com/thesofproject/sparse) ? This would help with handling errors, warning and exit codes too. Building sparse is quick and it seems to have very few dependencies, it may be easy to build it on the fly for every PR and break away from the current Zephyr image dependency. I can give that a try.

@marc-hb would be an option, sure

marc-hb:
> Couldn't we use tee to save output in a temporary file and then grep it again for the second output? It would certainly be better than not having the output at all, I could live with this.

Done in #6590

> @lyah should we maintain a "under review" sparse branch in say [http://github.com/thesofproject/sparse](https://github.com/thesofproject/sparse) ? This would help with handling errors, warning and exit codes too. Building sparse is quick and it seems to have very few dependencies, it may be easy to build it on the fly for every PR and break away from the current Zephyr image dependency. I can give that a try.

Done with your endianness fix in #6590 

#6590 does not solve world hunger or climate change but I think it comes pretty close.

- #6590

lgirdwood:
@lyakh #6646 now merged - is this ready now or will there be more dependency PRs ?

lyakh:
alternative #6646 has been merged

PR #6600, opened 2022-11-15
Use ops wrapper instead of calling the ops directly for get_attribute

aiChaoSONG:
the wrapper checks if the ops exist, and call the ops only if it exists, which is safer


lgirdwood (approved):

lgirdwood:
@aiChaoSONG thanks good find.

paulstelian97 (approved):
We had the wrapper and didn't use it? This project deserves a linter that can find issues like this.

PR #6556, opened 2022-11-08
Audio: SRC: Refinements for common 2-stage and generic C FIR

singalsu:
After changing this per review feedback the three commits are
- Audio: SRC: Optimize generic C fir data pointer wrap
- Audio: SRC: Refine input and output blocks count calculation
- Audio: SRC: Delete unused function parameter fir_delay_length

Only the first commit improves performance. The latter two have no measurable impact.

This PR saves in CML gcc build 2%, or 2.0 MCPS from 46.2 to 45.3
MCPS in stereo 44.1 kHz to 48 kHz conversion.

PR #6608, opened 2022-11-16
Audio: Mixin_mixout: Optimize the mixin processing function

andrula-song:
Use audio_stream_samples_without_wrap_sx to replace the complicated boundary detection can save at least 4.4% MCPS. Separate the mix function to normal mode and remap mode for better performance of optimization, and add HiFi3 version of mix processing functions

Signed-off-by: Andrula Song <xiaoyuan.song@intel.com>

### DIFF
--- a/src/audio/mixin_mixout/mixin_mixout.c
+++ b/src/audio/mixin_mixout/mixin_mixout.c
@@ -116,16 +116,9 @@ struct mixin_data {
 	 * private data as ipc4_base_module_cfg!
 	 */
 	struct ipc4_base_module_cfg base_cfg;
-
-	void (*mix_channel)(struct audio_stream __sparse_cache *sink, uint8_t sink_channel_index,
-			    uint8_t sink_channel_count, uint32_t start_frame, uint32_t mixed_frames,
-			    const struct audio_stream __sparse_cache *source,
-			    uint8_t source_channel_index, uint8_t source_channel_count,
-			    uint32_t frame_count, uint16_t gain);
-
-	void (*mute_channel)(struct audio_stream __sparse_cache *stream, uint8_t channel_index,
-			     uint32_t start_frame, uint32_t mixed_frames, uint32_t frame_count);
-
+	normal_mix_func normal_mix_channel;
+	remap_mix_func remap_mix_channel;
+	mute_func mute_channel;
 	struct mixin_sink_config sink_config[MIXIN_MAX_SINKS];
 };
 
@@ -278,287 +271,6 @@ static void reset_mixed_data_info_frame_counters(struct mixed_data_info __sparse
 		mdi->source_info[i].consumed_yet_not_produced_frames = 0;
 }
 
-#if CONFIG_FORMAT_S16LE
-/* Instead of using sink->channels and source->channels, sink_channel_count and
- * source_channel_count are supplied as parameters. This is done to reuse the function
- * to also mix an entire stream. In this case the function is called with fake stream
- * parameters: multichannel stream is treated as single channel and so the entire stream
- * contents is mixed.
- */
-static void mix_channel_s16(struct audio_stream __sparse_cache *sink, uint8_t sink_channel_index,
-			    uint8_t sink_channel_count, uint32_t start_frame, uint32_t mixed_frames,
-			    const struct audio_stream __sparse_cache *source,
-			    uint8_t source_channel_index, uint8_t source_channel_count,
-			    uint32_t frame_count, uint16_t gain)
-{
-	int16_t *dest, *src;
-	uint32_t frames_to_mix, frames_to_copy;
-
-	/* audio_stream_wrap() is required and is done below in a loop */
-	dest = (int16_t *)sink->w_ptr + start_frame * sink_channel_count + sink_channel_index;
-	src = (int16_t *)source->r_ptr + source_channel_index;
-
-	assert(mixed_frames >= start_frame);
-	frames_to_mix = MIN(mixed_frames - start_frame, frame_count);
-	frames_to_copy = frame_count - frames_to_mix;
-
-	/* we do not want to use something like audio_stream_frames_without_wrap() here
-	 * as it uses stream->channels internally. Also we would like to avoid expensive
-	 * division operations.
-	 */
-	while (frames_to_mix) {
-		src = audio_stream_wrap(source, src);
-		dest = audio_stream_wrap(sink, dest);
-
-		if (gain == IPC4_MIXIN_UNITY_GAIN)
-			while (frames_to_mix && src < (int16_t *)source->end_addr &&
-			       dest < (int16_t *)sink->end_addr) {
-				*dest = sat_int16((int32_t)*dest + (int32_t)*src);
-				src += source_channel_count;
-				dest += sink_channel_count;
-				frames_to_mix--;
-			}
-		else
-			while (frames_to_mix && src < (int16_t *)source->end_addr &&
-			       dest < (int16_t *)sink->end_addr) {
-				*dest = sat_int16((int32_t)*dest +
-						  q_mults_16x16(*src, gain,
-								IPC4_MIXIN_GAIN_SHIFT));
-				src += source_channel_count;
-				dest += sink_channel_count;
-				frames_to_mix--;
-			}
-	}
-
-	while (frames_to_copy) {
-		src = audio_stream_wrap(source, src);
-		dest = audio_stream_wrap(sink, dest);
-
-		if (gain == IPC4_MIXIN_UNITY_GAIN)
-			while (frames_to_copy && src < (int16_t *)source->end_addr &&
-			       dest < (int16_t *)sink->end_addr) {
-				*dest = *src;
-				src += source_channel_count;
-				dest += sink_channel_count;
-				frames_to_copy--;
-			}
-		else
-			while (frames_to_copy && src < (int16_t *)source->end_addr &&
-			       dest < (int16_t *)sink->end_addr) {
-				*dest = (int16_t)q_mults_16x16(*src, gain, IPC4_MIXIN_GAIN_SHIFT);
-				src += source_channel_count;
-				dest += sink_channel_count;
-				frames_to_copy--;
-			}
-	}
-}
-
-static void mute_channel_s16(struct audio_stream __sparse_cache *stream, uint8_t channel_index,
-			     uint32_t start_frame, uint32_t mixed_frames, uint32_t frame_count)
-{
-	uint32_t skip_mixed_frames;
-	int16_t *ptr;
-
-	assert(mixed_frames >= start_frame);
-	skip_mixed_frames = mixed_frames - start_frame;
-
-	if (frame_count <= skip_mixed_frames)
-		return;
-	frame_count -= skip_mixed_frames;
-
-	/* audio_stream_wrap() is needed here and it is just below in a loop */
-	ptr = (int16_t *)stream->w_ptr + mixed_frames * stream->channels + channel_index;
-
-	while (frame_count) {
-		ptr = audio_stream_wrap(stream, ptr);
-		while (frame_count && ptr < (int16_t *)stream->end_addr) {
-			*ptr = 0;
-			ptr += stream->channels;
-			frame_count--;
-		}
-	}
-}
-#endif	/* CONFIG_FORMAT_S16LE */
-
-#if CONFIG_FORMAT_S24LE
-/* Instead of using sink->channels and source->channels, sink_channel_count and
- * source_channel_count are supplied as parameters. This is done to reuse the function
- * to also mix an entire stream. In this case the function is called with fake stream
- * parameters: multichannel stream is treated as single channel and so the entire stream
- * contents is mixed.
- */
-static void mix_channel_s24(struct audio_stream __sparse_cache *sink, uint8_t sink_channel_index,
-			    uint8_t sink_channel_count, uint32_t start_frame, uint32_t mixed_frames,
-			    const struct audio_stream __sparse_cache *source,
-			    uint8_t source_channel_index, uint8_t source_channel_count,
-			    uint32_t frame_count, uint16_t gain)
-{
-	int32_t *dest, *src;
-	uint32_t frames_to_mix, frames_to_copy;
-
-	/* audio_stream_wrap() is required and is done below in a loop */
-	dest = (int32_t *)sink->w_ptr + start_frame * sink_channel_count + sink_channel_index;
-	src = (int32_t *)source->r_ptr + source_channel_index;
-
-	assert(mixed_frames >= start_frame);
-	frames_to_mix = MIN(mixed_frames - start_frame, frame_count);
-	frames_to_copy = frame_count - frames_to_mix;
-
-	/* we do not want to use something like audio_stream_frames_without_wrap() here
-	 * as it uses stream->channels internally. Also we would like to avoid expensive
-	 * division operations.
-	 */
-	while (frames_to_mix) {
-		src = audio_stream_wrap(source, src);
-		dest = audio_stream_wrap(sink, dest);
-
-		if (gain == IPC4_MIXIN_UNITY_GAIN)
-			while (frames_to_mix && src < (int32_t *)source->end_addr &&
-			       dest < (int32_t *)sink->end_addr) {
-				*dest = sat_int24(sign_extend_s24(*dest) + sign_extend_s24(*src));
-				src += source_channel_count;
-				dest += sink_channel_count;
-				frames_to_mix--;
-			}
-		else
-			while (frames_to_mix && src < (int32_t *)source->end_addr &&
-			       dest < (int32_t *)sink->end_addr) {
-				*dest = sat_int24(sign_extend_s24(*dest) +
-						  (int32_t)q_mults_32x32(sign_extend_s24(*src),
-						  gain, IPC4_MIXIN_GAIN_SHIFT));
-				src += source_channel_count;
-				dest += sink_channel_count;
-				frames_to_mix--;
-			}
-	}
-
-	while (frames_to_copy) {
-		src = audio_stream_wrap(source, src);
-		dest = audio_stream_wrap(sink, dest);
-
-		if (gain == IPC4_MIXIN_UNITY_GAIN)
-			while (frames_to_copy && src < (int32_t *)source->end_addr &&
-			       dest < (int32_t *)sink->end_addr) {
-				*dest = *src;
-				src += source_channel_count;
-				dest += sink_channel_count;
-				frames_to_copy--;
-			}
-		else
-			while (frames_to_copy && src < (int32_t *)source->end_addr &&
-			       dest < (int32_t *)sink->end_addr) {
-				*dest = (int32_t)q_mults_32x32(sign_extend_s24(*src),
-							       gain, IPC4_MIXIN_GAIN_SHIFT);
-				src += source_channel_count;
-				dest += sink_channel_count;
-				frames_to_copy--;
-			}
-	}
-}
-
-#endif	/* CONFIG_FORMAT_S24LE */
-
-#if CONFIG_FORMAT_S32LE
-/* Instead of using sink->channels and source->channels, sink_channel_count and
- * source_channel_count are supplied as parameters. This is done to reuse the function
- * to also mix an entire stream. In this case the function is called with fake stream
- * parameters: multichannel stream is treated as single channel and so the entire stream
- * contents is mixed.
- */
-static void mix_channel_s32(struct audio_stream __sparse_cache *sink, uint8_t sink_channel_index,
-			    uint8_t sink_channel_count, uint32_t start_frame, uint32_t mixed_frames,
-			    const struct audio_stream __sparse_cache *source,
-			    uint8_t source_channel_index, uint8_t source_channel_count,
-			    uint32_t frame_count, uint16_t gain)
-{
-	int32_t *dest, *src;
-	uint32_t frames_to_mix, frames_to_copy;
-
-	/* audio_stream_wrap() is required and is done below in a loop */
-	dest = (int32_t *)sink->w_ptr + start_frame * sink_channel_count + sink_channel_index;
-	src = (int32_t *)source->r_ptr + source_channel_index;
-
-	assert(mixed_frames >= start_frame);
-	frames_to_mix = MIN(mixed_frames - start_frame, frame_count);
-	frames_to_copy = frame_count - frames_to_mix;
-
-	/* we do not want to use something like audio_stream_frames_without_wrap() here
-	 * as it uses stream->channels internally. Also we would like to avoid expensive
-	 * division operations.
-	 */
-	while (frames_to_mix) {
-		src = audio_stream_wrap(source, src);
-		dest = audio_stream_wrap(sink, dest);
-
-		if (gain == IPC4_MIXIN_UNITY_GAIN)
-			while (frames_to_mix && src < (int32_t *)source->end_addr &&
-			       dest < (int32_t *)sink->end_addr) {
-				*dest = sat_int32((int64_t)*dest + (int64_t)*src);
-				src += source_channel_count;
-				dest += sink_channel_count;
-				frames_to_mix--;
-			}
-		else
-			while (frames_to_mix && src < (int32_t *)source->end_addr &&
-			       dest < (int32_t *)sink->end_addr) {
-				*dest = sat_int32((int64_t)*dest +
-						  q_mults_32x32(*src, gain, IPC4_MIXIN_GAIN_SHIFT));
-				src += source_channel_count;
-				dest += sink_channel_count;
-				frames_to_mix--;
-			}
-	}
-
-	while (frames_to_copy) {
-		src = audio_stream_wrap(source, src);
-		dest = audio_stream_wrap(sink, dest);
-
-		if (gain == IPC4_MIXIN_UNITY_GAIN)
-			while (frames_to_copy && src < (int32_t *)source->end_addr &&
-			       dest < (int32_t *)sink->end_addr) {
-				*dest = *src;
-				src += source_channel_count;
-				dest += sink_channel_count;
-				frames_to_copy--;
-			}
-		else
-			while (frames_to_copy && src < (int32_t *)source->end_addr &&
-			       dest < (int32_t *)sink->end_addr) {
-				*dest = (int32_t)q_mults_32x32(*src, gain, IPC4_MIXIN_GAIN_SHIFT);
-				src += source_channel_count;
-				dest += sink_channel_count;
-				frames_to_copy--;
-			}
-	}
-}
-
-static void mute_channel_s32(struct audio_stream __sparse_cache *stream, uint8_t channel_index,
-			     uint32_t start_frame, uint32_t mixed_frames, uint32_t frame_count)
-{
-	uint32_t skip_mixed_frames;
-	int32_t *ptr;
-
-	assert(mixed_frames >= start_frame);
-	skip_mixed_frames = mixed_frames - start_frame;
-
-	if (frame_count <= skip_mixed_frames)
-		return;
-	frame_count -= skip_mixed_frames;
-
-	/* audio_stream_wrap() is needed here and it is just below in a loop */
-	ptr = (int32_t *)stream->w_ptr + mixed_frames * stream->channels + channel_index;
-
-	while (frame_count) {
-		ptr = audio_stream_wrap(stream, ptr);
-		while (frame_count && ptr < (int32_t *)stream->end_addr) {
-			*ptr = 0;
-			ptr += stream->channels;
-			frame_count--;
-		}
-	}
-}
-#endif	/* CONFIG_FORMAT_S32LE */
-
 static int mix_and_remap(struct comp_dev *dev, const struct mixin_data *mixin_data,
 			 uint16_t sink_index, struct audio_stream __sparse_cache *sink,
 			 uint32_t start_frame, uint32_t mixed_frames,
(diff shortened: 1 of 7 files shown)


andrula-song:
here are the performance test results of xtensa simulater:
res1
compared with the original C version:
for 16bit normal mode mix, optimized C version can save about 37.6% cycles and hifi version can save about 76.7% cycles;
for 24bit normal mode mix, optimized C version can save about 50.7% cycles and hifi version can save about 84.1% cycles;
for 32bit normal mode mix, optimized C version can save about 44.0% cycles and hifi version can save about 84.6% cycles;

for 16bit remap mode mix, optimized C version can save about 9.1% cycles and hifi version can save about 20.1% cycles;
for 24bit remap mode mix, optimized C version can save about 4.4% cycles and hifi version can save about 47.8% cycles;
for 32bit remap mode mix, optimized C version can save about 5.0% cycles and hifi version can save about 57.1% cycles;

for 16bit mute, optimized C version can save about 10.9% cycles and hifi version can save about 35.5% cycles;
for 16bit mute, optimized C version can save about 9.6% cycles and hifi version can save about 36.5% cycles;

aiChaoSONG:
from your comments, I think several things are done in a single commit, is it possible to separate your work in several commits, with each only doing one thing.

andrula-song:
> from your comments, I think several things are done in a single commit, is it possible to separate your work in several commits, with each only doing one thing.
actually, all things are processing functions optimization and I think it is okay to have one commit to cover these.

src/audio/mixin_mixout/mixin_mixout_generic.c
```diff
@@ -0,0 +1,421 @@
...
+		nmax = audio_stream_samples_without_wrap_s16(source, src);
+		n = MIN(left_frames, nmax);
+		nmax = audio_stream_samples_without_wrap_s16(sink, dst);
+		n = MIN(n, nmax);
+		/* frames is the processed frame count in this loop*/
+		frames = 0;
+		for (i = 0; i < n; i++) {
+			*dst = sat_int16((int32_t)*dst + (int32_t)*src++);
+			dst++;
+			frames++;
```
lyakh, line 48: you don't need to count `frames`. You know, that after this loop completes, `frames` will be exactly equal to `n`. You can remove `frames` and just do `left_frames -= n` in the `for` statement at line 35.

src/audio/mixin_mixout/mixin_mixout_generic.c
```diff
@@ -0,0 +1,421 @@
...
+
+	/* audio_stream_wrap() is required and is done below in a loop */
+	int16_t *dst = (int16_t *)sink->w_ptr + start_frame;
+	int16_t *src = (int16_t *)source->r_ptr;
+
+	assert(mixed_frames >= start_frame);
+	frames_to_mix = MIN(mixed_frames - start_frame, frame_count);
+	frames_to_copy = frame_count - frames_to_mix;
+
+	for (left_frames = frames_to_mix; left_frames; left_frames -= frames) {
```
lyakh, line 35: we understand, that this loop is heavily relying on function parameters being "correct." If in some manner frame counts from function parameters don't match source and sink buffer states, this loop will malfunction, right? Personally I would put some guards there to protect from this loop running endlessly or e.g. from `left_frames` becoming "negative."

src/include/ipc4/mixin_mixout.h
```diff
@@ -85,4 +102,97 @@ struct ipc4_mixer_mode_config {
...
+static inline remap_mix_func remap_mix_get_processing_function(int fmt)
+{
+	int i;
+
+	/* map the remap mode mixin_mixout function for source and sink buffers */
+	for (i = 0; i < mix_count; i++) {
+		if (fmt != mix_func_map[i].frame_fmt)
+			continue;
+
+		return mix_func_map[i].remap_func;
```
lyakh, line 173:
```
if (fmt == mix_func_map[i].frame_fmt)
	return mix_func_map[i].remap_func;
```

src/audio/mixin_mixout/mixin_mixout_generic.c
andrula-song: yeah, thanks!

andrula-song:
Thanks  for the reminding of @lyakh!
Here are the latest improvement of normal mode generic C version:
for 16bit normal mode mix, optimized C version can save about 41.9% cycles;
for 24bit normal mode mix, optimized C version can save about 54.3% cycles;
for 32bit normal mode mix, optimized C version can save about 46.5% cycles.

btian1:
> nd and HiFi3 ve

two and and

btian1:
Did you get real time performance improvement data? then can have a compare with simulator result.

src/audio/mixin_mixout/mixin_mixout_generic.c
```diff
@@ -0,0 +1,403 @@
...
+	for (left_frames = frames_to_mix; left_frames > 0; left_frames -= n) {
+		src = audio_stream_wrap(source, src);
+		dst = audio_stream_wrap(sink, dst);
+		/* calculate the remaining samples*/
+		nmax = audio_stream_samples_without_wrap_s16(source, src);
+		n = MIN(left_frames, nmax);
+		nmax = audio_stream_samples_without_wrap_s16(sink, dst);
+		n = MIN(n, nmax);
+		for (i = 0; i < n; i++) {
+			*dst = sat_int16((int32_t)*dst + (int32_t)*src++);
```
btian1, line 44: for (i = 0; i < n; i++) 
			*dst++ = sat_int16((int32_t)*dst + (int32_t)*src++);

looks better?

andrula-song: There will be compilation errors when writing like this

andrula-song:
> Did you get real time performance improvement data? then can have a compare with simulator result.

it seems like too many mtrace log are lost, I haven't seen what I added.

lgirdwood (commented):
LGTM but it will need @singalsu to review the HiFi logic.

src/audio/mixin_mixout/mixin_mixout_generic.c
```diff
@@ -0,0 +1,403 @@
...
+		for (i = 0; i < n; i++)
+			*dst++ = *src++;
+	}
+}
+
+static void remap_mix_channel_s16(struct audio_stream __sparse_cache *sink,
+				  uint8_t sink_channel_index,  uint8_t sink_channel_count,
+				  uint32_t start_frame, uint32_t mixed_frames,
+				  const struct audio_stream __sparse_cache *source,
+				  uint8_t source_channel_index, uint8_t source_channel_count,
```
singalsu, line 65: These uint8_t types are used in the hot loop code. Have you tried if change to int improves code speed? Xtensa is slow with 8 bit types and also I don't see specific need for unsigned.

singalsu: With lots of parameters to function a struct reference may be faster?

src/audio/mixin_mixout/mixin_mixout_generic.c
```diff
@@ -0,0 +1,403 @@
...
+ * parameters: multichannel stream is treated as single channel and so the entire stream
+ * contents is mixed.
+ */
+static void normal_mix_channel_s16(struct audio_stream __sparse_cache *sink, uint32_t start_frame,
+				   uint32_t mixed_frames,
+				   const struct audio_stream __sparse_cache *source,
+				   uint32_t frame_count, uint16_t gain)
+{
+	uint32_t frames_to_mix, frames_to_copy, left_frames;
+	uint32_t n, nmax, i;
```
singalsu, line 25: Up to you but I'm avoiding unsigned types in hot code parts, it's never producing faster code than signed with normal arithmetic.

src/audio/mixin_mixout/mixin_mixout_generic.c
```diff
@@ -0,0 +1,403 @@
...
+		return;
+	frame_count -= skip_mixed_frames;
+	channel_count = stream->channels;
+	/* audio_stream_wrap() is needed here and it is just below in a loop */
+	ptr = (int16_t *)stream->w_ptr + mixed_frames * stream->channels + channel_index;
+
+	for (left_frames = frame_count; left_frames; left_frames -= frames) {
+		ptr = audio_stream_wrap(stream, ptr);
+		n = audio_stream_samples_without_wrap_s16(stream, ptr);
+		n =  MIN(left_frames * channel_count, n);
```
singalsu, line 135: I'm not putting equations to MIN() because of risk of becoming evaluated twice.

src/audio/mixin_mixout/mixin_mixout_generic.c
```diff
@@ -0,0 +1,403 @@
...
+			*dst++ = *src++;
+	}
+}
+
+static void remap_mix_channel_s24(struct audio_stream __sparse_cache *sink,
+				  uint8_t sink_channel_index,  uint8_t sink_channel_count,
+				  uint32_t start_frame, uint32_t mixed_frames,
+				  const struct audio_stream __sparse_cache *source,
+				  uint8_t source_channel_index, uint8_t source_channel_count,
+				  uint32_t frame_count, uint16_t gain)
```
singalsu, line 199: Here the same uint8_t parameters galore issue. Struct with int for generic numbers?

src/audio/mixin_mixout/mixin_mixout_generic.c
```diff
@@ -0,0 +1,403 @@
...
+			*dst++ = *src++;
+	}
+}
+
+static void remap_mix_channel_s32(struct audio_stream __sparse_cache *sink,
+				  uint8_t sink_channel_index,  uint8_t sink_channel_count,
+				  uint32_t start_frame, uint32_t mixed_frames,
+				  const struct audio_stream __sparse_cache *source,
+				  uint8_t source_channel_index, uint8_t source_channel_count,
+				  uint32_t frame_count, uint16_t gain)
```
singalsu, line 305: Same

src/audio/mixin_mixout/mixin_mixout_hifi3.c
```diff
@@ -0,0 +1,533 @@
...
+	ae_int16x4 *out;
+	ae_valign inu = AE_ZALIGN64();
+	ae_valign outu1 = AE_ZALIGN64();
+	ae_valign outu2 = AE_ZALIGN64();
+	/* audio_stream_wrap() is required and is done below in a loop */
+	ae_int16 *dst = (ae_int16 *)sink->w_ptr + start_frame;
+	ae_int16 *src = (ae_int16 *)source->r_ptr;
+
+	assert(mixed_frames >= start_frame);
+	frames_to_mix = MIN(mixed_frames - start_frame, frame_count);
```
singalsu, line 38: Arithmetic for MIN() argument, check if OK

src/audio/mixin_mixout/mixin_mixout_hifi3.c
```diff
@@ -0,0 +1,533 @@
...
+		/* calculate the remaining samples*/
+		nmax = audio_stream_samples_without_wrap_s16(source, src);
+		n = MIN(left_frames, nmax);
+		nmax = audio_stream_samples_without_wrap_s16(sink, dst);
+		n = MIN(n, nmax);
+		in = (ae_int16x4 *)src;
+		out = (ae_int16x4 *)dst;
+		inu = AE_LA64_PP(in);
+		outu1 = AE_LA64_PP(out);
+		for (i = 0; i < n; i += 4) {
```
singalsu, line 54: Please add comment where it's ensured that frames_to_mix and frame_count have multiple of four samples

src/audio/mixin_mixout/mixin_mixout_hifi3.c
```diff
@@ -0,0 +1,533 @@
...
+	AE_SETCBEGIN0(source->addr);
+	AE_SETCEND0(source->end_addr);
+
+	for (left_frames = frames_to_mix; left_frames; left_frames -= frames) {
+		/* audio_stream_wrap() is required and is done below in a loop */
+		dst = audio_stream_wrap(sink, dst);
+
+		/* calculate the remaining samples of sink*/
+		nmax = audio_stream_samples_without_wrap_s16(sink, dst);
+		n = MIN(left_frames * sink_channel_count, nmax);
```
singalsu, line 123: Check aritmetic in MIN() argument

singalsu: Have you tried AE_MIN32()

src/audio/mixin_mixout/mixin_mixout_hifi3.c
```diff
@@ -0,0 +1,533 @@
...
+
+		/* frames is the processed frame count in this loop*/
+		frames = 0;
+		for (i = 0; i < n; i += source_channel_count) {
+			AE_L16_XC(in, src, inoff);
+			AE_L16_XP(out, dst, 0);
+			/* Q1.15 * Q1.15 to Q2.30*/
+			temp = AE_MULF16SS_00(in, gain_v);
+			temp = AE_SRAA32(temp, shift);
```
singalsu, line 132: AE_SRAI32() because shift is immediate as IPC4_MIXIN_GAIN_SHIFT + 1

src/audio/mixin_mixout/mixin_mixout_hifi3.c
```diff
@@ -0,0 +1,533 @@
...
+			out1 = AE_SEXT32X2D16_10(out);
+			temp = AE_ADD32S(temp, out1);
+			temp = AE_SLAA32S(temp, 16);
```
singalsu, line 135: 16 is immediate so AE_SLAI32S(). It saves a HiFi register unless complier is smart enough to change this to instruction with immediate.

src/audio/mixin_mixout/mixin_mixout_hifi3.c
```diff
@@ -0,0 +1,533 @@
...
+		nmax = audio_stream_samples_without_wrap_s16(sink, dst);
+		n = MIN(left_frames * sink_channel_count, nmax);
+		frames = 0;
+
+		for (i = 0; i < n; i += source_channel_count) {
+			AE_L16_XC(in, src, inoff);
+			AE_L16_XP(out, dst, 0);
+			temp = AE_MULF16SS_00(in, gain_v);
+			temp = AE_SRAA32(temp, shift);
+			temp = AE_SLAA32S(temp, 16);
```
singalsu, line 153: same^2

src/audio/mixin_mixout/mixin_mixout_hifi3.c
```diff
@@ -0,0 +1,533 @@
...
+	int off = stream->channels * sizeof(ae_int16);
+	ae_int16 *ptr;
+	ae_int16x4 zero = AE_ZERO16();
+
+	assert(mixed_frames >= start_frame);
+	skip_mixed_frames = mixed_frames - start_frame;
+
+	if (frame_count <= skip_mixed_frames)
+		return;
+	frame_count -= skip_mixed_frames;
```
singalsu, line 174: No point to use unsigned types when doing subtracting them etc. The DSP ALU computes signed integers natively so why not use them.

src/audio/mixin_mixout/mixin_mixout_hifi3.c
```diff
@@ -0,0 +1,533 @@
...
+		AE_SA64POS_FP(outu2, out);
+	}
+}
+
+static void remap_mix_channel_s24(struct audio_stream __sparse_cache *sink,
+				  uint8_t sink_channel_index,  uint8_t sink_channel_count,
+				  uint32_t start_frame, uint32_t mixed_frames,
+				  const struct audio_stream __sparse_cache *source,
+				  uint8_t source_channel_index, uint8_t source_channel_count,
+				  uint32_t frame_count, uint16_t gain)
```
singalsu, line 263: A lot of uint8_t parameters to slow down the code here as well

src/audio/mixin_mixout/mixin_mixout_hifi3.c
```diff
@@ -0,0 +1,533 @@
...
+	AE_L16_IP(gain_v, pgain, 0);
+	/* set source as circular buffer, hifi3 only have 1 circular buffer*/
+	AE_SETCBEGIN0(source->addr);
+	AE_SETCEND0(source->end_addr);
+
+	for (left_frames = frames_to_mix; left_frames; left_frames -= frames) {
+		dst = audio_stream_wrap(sink, dst);
+		/* calculate the remaining samples*/
+		nmax = audio_stream_samples_without_wrap_s24(sink, dst);
+		n = MIN(left_frames * sink_channel_count, nmax);
```
singalsu, line 294: AE_MIN32()?

src/audio/mixin_mixout/mixin_mixout_hifi3.c
```diff
@@ -0,0 +1,533 @@
...
+
+		for (i = 0; i < n; i += source_channel_count) {
+			AE_L32_XC(in, src, inoff);
+			AE_L32_XP(out, dst, 0);
+			tmp = AE_MUL32X16_H0(in, gain_v);
+
+			/* shift should be -(IPC4_MIXIN_GAIN_SHIFT
+			 * - 16(to keep the valid LSB bits of ae_int64))
+			 */
+			tmp = AE_SLAA64S(tmp, shift);
```
singalsu, line 461: AE_SLAI64S() with immediate instead of variable shift

src/audio/mixin_mixout/mixin_mixout_hifi3.c
```diff
@@ -0,0 +1,533 @@
...
+		dst = audio_stream_wrap(sink, dst);
+		nmax = audio_stream_samples_without_wrap_s32(sink, dst);
+		n = MIN(left_frames * sink_channel_count, nmax);
+		/* frames is the processed frame count in this loop*/
+		frames = 0;
+
+		for (i = 0; i < n; i += source_channel_count) {
+			AE_L32_XC(in, src, inoff);
+			tmp = AE_MUL32X16_H0(in, gain_v);
+			tmp = AE_SLAA64S(tmp, shift);
```
singalsu, line 480: same

singalsu (changes requested):

src/audio/mixin_mixout/mixin_mixout_generic.c
marc-hb: I don't see why these `(uint32_t)` casts are needed. Casts hide bugs, especially when dealing with pointers and pointer arithmetic. Remove these casts or add a comment explaining why they are required.

marc-hb: Unsigned overflow is defined in the standard (and the arguments may even be promoted to `int` before the addition takes, can't remember the details).

src/audio/mixin_mixout/mixin_mixout_generic.c
marc-hb: > and also I don't see specific need for unsigned.

One of the big differences between signed and unsigned is signed has no well-defined overflow and underflow. I don't know whether overflow matters here.

But the worst things that can happen (including security issues) is when mixing signed and unsigned together.

src/audio/mixin_mixout/mixin_mixout_generic.c
marc-hb: > *dst++ = sat_int16((int32_t)*dst + (int32_t)*src++);

Wow, just noticed this. This is incredibly dangerous because you cannot tell whether the `dst` on the right side will be incremented by the left side before or after the right side is evaluated.

> There will be compilation errors when writing like this

Which one?

andrula-song: >you are right , if make it like *dst++ = sat_int16((int32_t)*dst + (int32_t)*src++); would lesd to compilation errors

src/audio/mixin_mixout/mixin_mixout_generic.c
andrula-song: > With lots of parameters to function a struct reference may be faster?

If use a struct, we need to assign value to it before we call the function. And I found that not in all cases, the fewer parameters, the better, like https://github.com/thesofproject/sof/pull/6556. For example, test by xtensa simulator, if I do not remove the useless parameters of the normal mode functions, I will use fewer cycles than I remove it.

andrula-song: > But the worst things that can happen (including security issues) is when mixing signed and unsigned together.
our audio stream will always be singed type.

src/audio/mixin_mixout/mixin_mixout_hifi3.c
andrula-song: it doesn't need to be multiple of four samples, since each time I update the src and dst base on n.

src/audio/mixin_mixout/mixin_mixout_hifi3.c
andrula-song: since AE_MIN32() needs two ae_int32x2, I 'd like AE_MIN_32_signed

src/audio/mixin_mixout/mixin_mixout_hifi3.c
andrula-song: Almost forget that, I'd like to use another PR to use the frame limits API to add limits of processing limits. Thanks!

src/audio/mixin_mixout/mixin_mixout_hifi3.c
andrula-song: it's ok to do so.

src/audio/mixin_mixout/mixin_mixout_generic.c
andrula-song: Since the intrinsic AE_MIN_32_signed is faster than MIN(), so I will use AE_MIN_32_signed, and the arguments of this instruction need to be signed.

src/audio/mixin_mixout/mixin_mixout_hifi3.c
andrula-song: I thought that shift can avoid the recalculation of  IPC4_MIXIN_GAIN_SHIFT + 1, but from the simulator,   AE_SRAA32(temp, shift) is slower than  AE_SRAI32(temp, IPC4_MIXIN_GAIN_SHIFT + 1), thanks!

src/audio/mixin_mixout/mixin_mixout_generic.c
andrula-song: change uint8_t to int32_t can have save about 0.6% cycles in remap_mix_channel_s16, so I will change the parameters to int32_t, but not struct.

src/audio/mixin_mixout/mixin_mixout_generic.c
```diff
@@ -0,0 +1,403 @@
...
+		nmax = audio_stream_samples_without_wrap_s16(sink, dst);
+		n = MIN(n, nmax);
+		/* frames is the processed frame count in this loop*/
+		frames = 0;
+		for (i = 0; i < n; i += source_channel_count) {
+			*dst = sat_int16((int32_t)*dst +
+			       q_mults_16x16(*src, gain, IPC4_MIXIN_GAIN_SHIFT));
+			src += source_channel_count;
+			dst += sink_channel_count;
+			frames++;
```
lyakh, line 95: still needlessly counting `frames` - here and in other locations

src/audio/mixin_mixout/mixin_mixout_generic.c
lyakh: ack, the compiler will likely optimise this, but better avoid this with our "trivial" definition of `MIN()`. More complex definitions avoid this by using local variables to evaluate arguments only once. Since we don't do that, better not use expressions in `MIN()` arguments. Also there's a superfluous space there.

src/audio/mixin_mixout/mixin_mixout_generic.c
lyakh: you think a single division is more expensive than possibly hundreds of increments? But ok, not really important.

src/audio/mixin_mixout/mixin_mixout_generic.c
marc-hb: Don't "resolve" and hide unfinished conversations.

It's not required to resolve all conversations.

src/audio/mixin_mixout/mixin_mixout_generic.c
andrula-song: actually, it is because the author of original functions comments he doesn't  want a division, so I avoid to use division here.

src/audio/mixin_mixout/mixin_mixout_generic.c
andrula-song: Sorry, since I remove that two (uint32_t), I thought this has done.

marc-hb: Some threads have more than one topic.

andrula-song:
hi @singalsu ， can you help to review? Thanks!

btian1:
> 

you don't need add any new traces, just enable -d -o app/perf.overlay.conf to check, there is traces like:
[    1.267010] <inf> component: comp:1 0x30000 perf comp_copy samples 48 period 1000 cpu avg 171 peak 221
[    1.268125] <inf> component: comp:0 0x20000 perf comp_copy samples 48 period 1000 cpu avg 380 peak 620

it should be mixin mixout data.

if not able to see those logs, means your mtrace still need more configuration.

src/audio/mixin_mixout/mixin_mixout_generic.c
btian1: normally, divide(int) takes up to 32 cycles in dsp, and divide also can be done with >> partially.
The code looks complex, still because the wrapper buffer usage

andrula-song:
> > 
> 
> you don't need add any new traces, just enable -d -o app/perf.overlay.conf to check, there is traces like: [ 1.267010] component: comp:1 0x30000 perf comp_copy samples 48 period 1000 cpu avg 171 peak 221 [ 1.268125] component: comp:0 0x20000 perf comp_copy samples 48 period 1000 cpu avg 380 peak 620
> 
> it should be mixin mixout data.
> 
> if not able to see those logs, means your mtrace still need more configuration.

there are the  test result on my tgl: 
the original: [   12.541026] <inf> component: comp:1 0x20000 perf comp_copy samples 48 period 1000 cpu avg 374 peak 390
the optimized: [   91.384518] <inf> component: comp:5 0x20000 perf comp_copy samples 48 period 1000 cpu avg 234 peak 261
the original version average is 374, the max peak is 390; the  optimized version  worst average is 234 , the max peak is 261. 
From the perspective of the entire mixin, the overall improvement is 37.4%.
And it's very strange that there is a huge difference  between your peak and average.  What's is your device?

src/audio/mixin_mixout/mixin_mixout_generic.c
singalsu: There's no policy to have generic integers as unsigned when possible.

src/audio/mixin_mixout/mixin_mixout_generic.c
singalsu: OK, but be careful when mixing unsigned and signed arithmetic. It's safest to use only signed types. It's very rare that we need to overflow the integers (a stupid shortcoming in C standard to claim it's undefined) or need the 2x positive range of unsigned.

singalsu (approved):

lgirdwood (approved):

src/audio/mixin_mixout/mixin_mixout_generic.c
```diff
@@ -0,0 +1,417 @@
...
+
+	for (left_frames = frames_to_copy; left_frames > 0; left_frames -= n) {
+		src = audio_stream_wrap(source, src);
+		dst = audio_stream_wrap(sink, dst);
+		nmax = audio_stream_samples_without_wrap_s32(source, src);
+		n = MIN(left_frames, nmax);
+		nmax = audio_stream_samples_without_wrap_s32(sink, dst);
+		n = MIN(n, nmax);
+		for (i = 0; i < n; i++)
+			*dst++ = *src++;
```
aiChaoSONG, line 306: do you think a memory copy will be faster here?

marc-hb: Good catch, it's very strange to see `*dst++ = *src++;` in a PR named "Optimize"

https://docs.zephyrproject.org/apidoc/latest/string_8h.html
https://github.com/thesofproject/sof/blob/main/src/arch/xtensa/include/arch/string.h
https://github.com/thesofproject/sof/blob/main/zephyr/include/rtos/string.h

andrula-song: test on xtensa simulator and my TGL several times, use __vec_memcpy to replace the L/S performace worse!
So just change generic C to use memcpy_s， the latest results are:
for 16bit normal mode mix, optimized C version can save about 56.9% cycles;
for 24bit normal mode mix, optimized C version can save about 62.9% cycles;
for 32bit normal mode mix, optimized C version can save about 55.5% cycles.

src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -1021,24 +734,13 @@ static int mixin_prepare(struct comp_dev *dev)
...
 	case SOF_IPC_FRAME_S24_4LE:
-		md->mix_channel = mix_channel_s24;
-		md->mute_channel = mute_channel_s32;	/* yes, 32 is correct */
-		break;
-#endif /* CONFIG_FORMAT_S24LE */
-#if CONFIG_FORMAT_S32LE
 	case SOF_IPC_FRAME_S32_LE:
-		md->mix_channel = mix_channel_s32;
-		md->mute_channel = mute_channel_s32;
+		md->normal_mix_channel = normal_mix_get_processing_function(fmt);
```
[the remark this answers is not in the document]
andrula-song: Actually only if the format is not one of SOF_IPC_FRAME_S16_LE, SOF_IPC_FRAME_S24_4LE, SOF_IPC_FRAME_S32_LE would return null,  it is the case of default, so it is not necessary to add the error handled here. But add here as you like.

src/audio/mixin_mixout/mixin_mixout_generic.c
andrula-song: Ok, change those parameters to int32_t.

src/audio/mixin_mixout/mixin_mixout_generic.c
marc-hb: > use __vec_memcpy to replace the L/S performace worse!

This looks like a very serious bug. Please file a bug and add a link to it here in the source.

btian1:
> 

I am using 16bit for the test, and the difference is just come from peak, it maybe environment related, so it is not issue.
Let's continue to merge.

btian1 (approved):

PR #6611, opened 2022-11-16
Audio: SRC: Generic C SRC FIR core improvement

singalsu:
This patch compacts the FIR code a bit per discussion in other SRC
    refine patch review where the current pointer arithmetic and code
    style was not liked. There is no measurable MCPS impact in the 2ch
    and Nch versions of core. It's not slower than original, but possibly
    one perf comp_copy trace unit faster (0.03%).
    
### DIFF
--- a/src/audio/src/src_generic.c
+++ b/src/audio/src/src_generic.c
@@ -59,22 +59,17 @@ static inline void fir_filter_generic(int32_t *rp, const void *cp, int32_t *wp0,
 		 * output shift includes the shift by 15 for Qx.46 to
 		 * Qx.31.
 		 */
-		for (i = 0; i < n1; i++) {
-			y0 += (int64_t)(*coef) * (*data);
-			data++;
-			y1 += (int64_t)(*coef) * (*data);
-			data++;
-			coef++;
+		for (i = 0; i < n1; i++, coef++, data += 2) {
+			y0 += (int64_t)(*coef) * data[0];
+			y1 += (int64_t)(*coef) * data[1];
 		}
+
 		if (data == fir_end)
 			data = fir_start;
 
-		for (i = 0; i < n2; i++) {
-			y0 += (int64_t)(*coef) * (*data);
-			data++;
-			y1 += (int64_t)(*coef) * (*data);
-			data++;
-			coef++;
+		for (i = 0; i < n2; i++, coef++, data += 2) {
+			y0 += (int64_t)(*coef) * data[0];
+			y1 += (int64_t)(*coef) * data[1];
 		}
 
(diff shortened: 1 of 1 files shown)


singalsu:
I just changed the PR title to be more generic. 

The cache performance issue with N channel version needs a larger rewrite so better to have this as a separate PR. There's possibly need to separate 1 ch and N ch cases.

lgirdwood (commented):
@andrula-song @lyakh pls review

andrula-song (approved):

src/audio/src/src_generic.c
```diff
@@ -60,21 +60,15 @@ static inline void fir_filter_generic(int32_t *rp, const void *cp, int32_t *wp0,
...
 			data = fir_start;
 
 		for (i = 0; i < n2; i++) {
-			y0 += (int64_t)(*coef) * (*data);
-			data++;
-			y1 += (int64_t)(*coef) * (*data);
-			data++;
-			coef++;
+			y0 += (int64_t)(*coef) * (*(data++));
+			y1 += (int64_t)(*(coef++)) * (*(data++));
```
btian1, line 71: I would suggest remove (), due to () may bring understanding trouble to developer, * or () which have higher priority?
y1 += (int64_t)(*coef++ * *data++);

lyakh: we seem to be in a rather subjective area here. My personal order of preference:
1. `for (i=0; i<5; i++, ptr++) y=*ptr;`
2. `for (i=0; i<5; i++) {y=*ptr; ptr++;}`
3. `for (i=0; i<5; i++) y=*(ptr++);`
4. `for (i=0; i<5; i++) y=*ptr++;`

but that's just me. I can live with either of them, so please choose whatever you, guys, who work with this code most frequently, feel most comfortable with. I also don't think this choice would make any performance difference. Possibly only in loops like the one above, where `data` is incremented twice. *Maybe* doing
```
for (i=0; i<5; i++) {
	y0 += data[0];
	y1 += data[1];
	data += 2;
}
```
would generate 1 or 2 asm instructions fewer, or maybe it won't

lgirdwood: Fwiw, the latter from @lyakh comment is probably easier for compilers to track/vectorize as the offsets are constant wrt loop iteration.

lyakh (approved):
wow, I like this change! But I should say, that most of this is subjective. I'm sure many developers would prefer the original style or something yet different. So, I'd say it's up to those who work with this code the most to pick up the style they prefer (within the project style guidelines of course)

lgirdwood (approved):

PR #6538, opened 2022-11-04
dma: ipc4: dai: ace: Use extended Zephyr DMA API

softwarecki:
The dma_get_attribute function was renamed to dma_get_attribute_legacy. New version of it was added to Zephyr DMA API. It is used when using zephyr native drivers. The local copy of get_attribute functions for DW and ALH DMAs in ACE is no longer necessary and was removed.
The Zephyrs dma_get_status function was extended to support read of a linear link position. It is now used in dai_position and dai_dma_position_update funtions for ipc4.

**This PR require merged https://github.com/zephyrproject-rtos/zephyr/pull/51957.**

PR #6613, opened 2022-11-16
add support L3 (IMR) heap

marcinszkudlinski:
L3 memory should support have support
for allocating/freeing memory
A heap support in L3 is introduced by this PR

NOTE! the IMR memory layout definitions will be soon introduced in Zephyr
so the PR won't compile till a zephyr's PR below is merged 

- [x] https://github.com/zephyrproject-rtos/zephyr/pull/47840
- [x] https://github.com/zephyrproject-rtos/zephyr/pull/52477
- [x] https://github.com/thesofproject/sof/pull/6538 (fix for CI failures and newest zephyr usage)
- [x] https://github.com/thesofproject/sof/pull/6678 (warnign removal - currently included in this PR)


marcinszkudlinski:
Update: conflicts resolved (changes from 371d35108a75cd8788cfadcb00ae9a3028677f22 - Memory: add support for memory heap profiling) imported

src/lib-zephyr/alloc.c
```diff
@@ -0,0 +1,280 @@
+// SPDX-License-Identifier: BSD-3-Clause
```
lgirdwood, line 1: Not following why we need another Zephyr directory ?  See https://github.com/thesofproject/sof/pull/6619

src/lib-zephyr/alloc.c
```diff
@@ -0,0 +1,280 @@
...
+LOG_MODULE_REGISTER(mem_allocator, CONFIG_SOF_LOG_LEVEL);
+
+/*
+ * Memory - Create Zephyr HEAP for SOF.
+ *
+ * Currently functional but some items still WIP.
+ */
+
+#ifndef HEAP_RUNTIME_SIZE
+#define HEAP_RUNTIME_SIZE	0
```
lgirdwood, line 50: looks like this is a xtos only option ? I wonder where we are still using in Zephyr config ?

src/lib-zephyr/alloc.c
```diff
@@ -0,0 +1,280 @@
...
+/* The Zephyr heap */
+
+#ifdef CONFIG_IMX
+#define HEAPMEM_SIZE		(HEAP_SYSTEM_SIZE + HEAP_RUNTIME_SIZE + HEAP_BUFFER_SIZE)
+
+/*
+ * Include heapmem variable in .heap_mem section, otherwise the HEAPMEM_SIZE is
+ * duplicated in two sections and the sdram0 region overflows.
+ */
+__section(".heap_mem") static uint8_t __aligned(64) heapmem[HEAPMEM_SIZE];
```
lgirdwood, line 67: cache line size instead of 64 ?

src/lib-zephyr/alloc.c
marcinszkudlinski: There was an idea - having 2 dirs lib and zephyr-lib/ and providing same services natively / using zephyr. 
#6619  is another idea of same thing

src/lib-zephyr/alloc.c
marcinszkudlinski: IMX platform only - to be removed

src/lib-zephyr/alloc.c
marcinszkudlinski: IMX platform only - to be removed

src/lib-zephyr/alloc.c
marcinszkudlinski: moved to zephyr/lib

zephyr/lib/alloc.c
```diff
@@ -68,12 +68,109 @@ extern char _end[], _heap_sentry[];
...
+static bool is_l3_heap_pointer(void *ptr)
+{
+	uintptr_t l3_heap_start = get_l3_heap_start();
+	uintptr_t l3_heap_end = l3_heap_start + get_l3_heap_size();
+
+	if (POINTER_TO_UINT(ptr) >= l3_heap_start &&
+	    POINTER_TO_UINT(ptr) < l3_heap_end) {
+		return true;
+	}
+	return false;
```
lyakh, line 120: could be `return POINTER_TO_UINT(ptr) >= l3_heap_start && POINTER_TO_UINT(ptr) < l3_heap_end;`

zephyr/lib/alloc.c
```diff
@@ -68,12 +68,109 @@ extern char _end[], _heap_sentry[];
...
+
+	key = k_spin_lock(&l3_heap.lock);
+	ret = sys_heap_aligned_alloc(&l3_heap.heap, L3_MEM_PAGE_SIZE, size);
+	k_spin_unlock(&l3_heap.lock, key);
+	return ret;
+}
+
+/**
+ * Allocate buffer from the L3 memory heap.
+ * @param mem SPointer to memory block to be freed.
```
lyakh, line 142: typo?

zephyr/lib/alloc.c
```diff
@@ -68,12 +68,109 @@ extern char _end[], _heap_sentry[];
...
+	 */
+	uintptr_t l3_heap_start = get_l3_heap_start();
+	uintptr_t l3_heap_end = l3_heap_start + get_l3_heap_size();
+
+	if ((uintptr_t)mem < l3_heap_start && (uintptr_t)mem >= l3_heap_end)
+		return;
+
+	k_spinlock_key_t key = k_spin_lock(&l3_heap.lock);
+
+	sys_heap_free(&l3_heap.heap, mem);
```
lyakh, line 158: are these allocations always going to be uncached?

marcinszkudlinski: good point! it is always CACHED (adress area 0xA1...)

good that currently IMR heap is used exclusively by core 0
Anyway "cache resistance" as in HPSRAM heap is needed here 

ANWYAY, for now I can do a workaround for this
 - an assert limiting alloc/free of IMR heap to core 0
 - open a defect/feture req for multicore implementation

we do need IMR heap for D3 a.k.a "imr boot" flow, badly needed for chrome team :)

zephyr/lib/alloc.c
```diff
@@ -47,27 +47,9 @@ extern struct tr_ctx zephyr_tr;
...
-#ifndef HEAP_RUNTIME_SIZE
-#define HEAP_RUNTIME_SIZE	0
-#endif
-
-/* system size not declared on some platforms */
-#ifndef HEAP_SYSTEM_SIZE
-#define HEAP_SYSTEM_SIZE	0
-#endif
-
 /* The Zephyr heap */
```
kv2019i, line 51: Hmm, sure it's ok to just remove them?

zephyr/lib/alloc.c
```diff
@@ -47,27 +47,9 @@ extern struct tr_ctx zephyr_tr;
...
-#define HEAPMEM_SIZE		(HEAP_SYSTEM_SIZE + HEAP_RUNTIME_SIZE + HEAP_BUFFER_SIZE)
-
-/*
- * Include heapmem variable in .heap_mem section, otherwise the HEAPMEM_SIZE is
- * duplicated in two sections and the sdram0 region overflows.
- */
-__section(".heap_mem") static uint8_t __aligned(64) heapmem[HEAPMEM_SIZE];
-
-#elif CONFIG_ACE
+#ifdef CONFIG_ACE
```
kv2019i, line 54: ... uhm, wouldn't this break SOF Zephyr on IMX?

src/include/ipc/topology.h
```diff
@@ -77,14 +77,16 @@ struct sof_ipc_comp {
...
+#define SOF_MEM_CAPS_RAM		BIT(0)
+#define SOF_MEM_CAPS_ROM		BIT(1)
+#define SOF_MEM_CAPS_EXT		BIT(2) /**< external */
+#define SOF_MEM_CAPS_LP			BIT(3) /**< low power */
+#define SOF_MEM_CAPS_HP			BIT(4) /**< high performance */
+#define SOF_MEM_CAPS_DMA		BIT(5) /**< DMA'able */
+#define SOF_MEM_CAPS_CACHE		BIT(6) /**< cacheable */
+#define SOF_MEM_CAPS_EXEC		BIT(7) /**< executable */
+#define SOF_MEM_CAPS_L3			BIT(8) /**< L3 memory */
+
```
kv2019i, line 89: This is bit of a historical baggage as this definition is used directly in the IPC3 SOF_IPC_TPLG_BUFFER_NEW (for caps field) and the same field is used internally in FW to manage memory region capabilities. 

Hmm, this change is probably still best. Capability makes sense in general (not tied to IPC version or platform), even if it is used on some platforms. This does mean (see next comment)...

src/include/kernel/abi.h
```diff
@@ -30,7 +30,7 @@
 /** \brief SOF ABI version major, minor and patch numbers */
 #define SOF_ABI_MAJOR 3
 #define SOF_ABI_MINOR 24
-#define SOF_ABI_PATCH 0
+#define SOF_ABI_PATCH 1
```
kv2019i, line 33: If we need to change structs shared with IPC3, this needs a MINOR bump and be tagged with "ABI".

src/platform/meteorlake/include/platform/lib/memory.h
```diff
@@ -48,9 +48,10 @@
 #define PLATFORM_HEAP_SYSTEM_SHARED	1
 #define PLATFORM_HEAP_BUFFER		2
 
-/* boot loader in IMR */
-#define IMR_BOOT_LDR_MANIFEST_BASE	0xA1042000
-#define IMR_BOOT_LDR_MANIFEST_SIZE	0x6000
+/**
+ * size of HPSRAM system heap
+ */
+#define HEAPMEM_SIZE 0x40000
```
kv2019i, line 54: I was considering to submit this separate (remove IMR_BOOT), but it seems you have it here. I guess works like this as well, but we'll have build warnings upon next Zephyr baseline update as long as we have the define in memory.h, so in general better to do those changes in SOF upfront that can be done.

zephyr/lib/alloc.c
```diff
@@ -70,12 +70,108 @@ extern char _end[], _heap_sentry[];
...
+	void *ret;
+
+	key = k_spin_lock(&l3_heap.lock);
+	ret = sys_heap_aligned_alloc(&l3_heap.heap, L3_MEM_PAGE_SIZE, size);
+	k_spin_unlock(&l3_heap.lock, key);
+	return ret;
+}
+
+/**
+ * Allocate buffer from the L3 memory heap.
```
kv2019i, line 142: Free buffer...?

zephyr/lib/alloc.c
```diff
@@ -70,12 +70,108 @@ extern char _end[], _heap_sentry[];
...
+void l3_free(void *mem)
+{
+	/*
+	 * first check to which heap the mem argument belongs to
+	 */
+	uintptr_t l3_heap_start = get_l3_heap_start();
+	uintptr_t l3_heap_end = l3_heap_start + get_l3_heap_size();
+
+	if ((uintptr_t)mem < l3_heap_start && (uintptr_t)mem >= l3_heap_end)
+		return;
```
kv2019i, line 155: Could this L148-155 be just "ASSERT_NO_MSG(is_l3_heap_pointer(mem));"
It seems most places already check before calling l3_free and sys_heap_free() will check for validity as well.

kv2019i (changes requested):
Looks pretty clean in general, but a few issues that need addressing in the cleanup patch and how the new capability is added to public headers. Please see inline.

zephyr/lib/alloc.c
marcinszkudlinski: ok, leaving as it was, does not hurt anyway

zephyr/lib/alloc.c
marcinszkudlinski: obsolete, l3_free removed

zephyr/lib/alloc.c
marcinszkudlinski: fixed, handling of l3 heap redesigned

marcinszkudlinski:
CHANGES
- L3 heap is now handled by the same procedures as main heap - and is cache / therad safe now
- some 3rd party platforms reintroduced to alloc.c 
- dupliocated memory definitions removed from TGL as well as from MTL

marcinszkudlinski:
sorry, accidental force-push. Cleaned up.

src/platform/intel/cavs/include/cavs/lib/memory.h
```diff
@@ -35,13 +38,6 @@
 
 #define MAX_MEMORY_SEGMENTS		PLATFORM_HPSRAM_SEGMENTS
 
-#if CONFIG_LP_MEMORY_BANKS
-#define LP_SRAM_SIZE \
-	(CONFIG_LP_MEMORY_BANKS * SRAM_BANK_SIZE)
-#else
-#define LP_SRAM_SIZE 0
-#endif
-
```
kv2019i, line 44: This will break XTOS build of Intel cAVS platforms.

I'm trying an alternative at https://github.com/thesofproject/sof/pull/6678  -- let's see if it goes through the checks.

kv2019i (commented):
Thanks @marcinszkudlinski .. only thing that remains is the cAVS legacy builds. Let's see what the CI things about it (about this and my #6678). If we go the route of removing Intel cAVS XTOS builds, that should be done first.

marcinszkudlinski:
@kv2019i the PR is now rebased to https://github.com/thesofproject/sof/pull/6678

kv2019i (approved):
Thanks, seems good now!

src/platform/tigerlake/include/platform/lib/memory.h
```diff
@@ -51,6 +51,11 @@
 #define HOST_WIN_SIZE		0x00000008
 #define HOST_WIN_BASE(x)	(0x00071A00 + x * HOST_WIN_SIZE)
 
+/*
+ * size of HPSRAM system heap
+ */
+#define HEAPMEM_SIZE 0x40000
```
kv2019i, line 57: @marcinszkudlinski D'oh, sorry, missed this in the review and focused on the IMR boot definitions. Please just drop this HEAPMEM_SIZE, we don't need it for legacy platforms. The definition in zephyr alloc.c "#define HEAPMEM_SIZE ((uint8_t *)_heap_sentry - heapmem)" handled the cAVS platforms.

zephyr/lib/alloc.c
```diff
@@ -28,6 +28,7 @@
 #include <version.h>
 #include <zephyr/sys/__assert.h>
 #include <soc.h>
+#include <adsp_memory.h>
```
kv2019i, line 29: What do you actually need from adsp_memory.h? You are not adding anything here to alloc.c?

kv2019i (commented):
@marcinszkudlinski Hmm, CI results still a bit brutal. Please check one thing inline, that could fix at least one category of issues.

zephyr/lib/alloc.c
marcinszkudlinski: > What do you actually need from adsp_memory.h? You are not adding anything here to alloc.c?

IMR_L3_HEAP_BASE
IMR_L3_HEAP_SIZE
IMR_L3_HEAP_BASE
IMR_L3_HEAP_SIZE

taken from DeviceTree

src/platform/tigerlake/include/platform/lib/memory.h
marcinszkudlinski:
```
#elif CONFIG_ACE

/*
 * System heap definition for ACE is defined below.
 * It needs to be explicitly packed into dedicated section
 * to allow memory management driver to control unused
 * memory pages.
 */
__section(".heap_mem") static uint8_t __aligned(PLATFORM_DCACHE_ALIGN) heapmem[**HEAPMEM_SIZE**];
```
do we? For MTL we need to have limited heap size to keep HPSRAM free for loadable libraries

kv2019i: @marcinszkudlinski I mean for older platform we size the heap based on linker script markers. So you'll need HEAPMEM_SIZE only for MTL. I guess we could reserve space to dynamic libs via linker script markers as well (just an idea).
This should still help to reduce conflicts.

zephyr/lib/alloc.c
kv2019i: Ack, might be better to have the adsp_memory.h inclusion in the patch where these start to be used. But right, I get it now.

kv2019i: @marcinszkudlinski And another problem. adsp_memory.h is an Intel specific file, so this would not work with e.g. nxp targets (that use Zephyr as well) and share the alloc.c code.  So I think the IMR_L3_HEAP definitions need to be made conditional on the L3_HEAP feature and perhaps pull the values directly from DT (in alloc.c).

aborisovich:
Still checkpatch to address

zephyr/lib/alloc.c
marc-hb: > adsp_memory.h is an Intel specific file, so this would not work with e.g. nxp targets (that use Zephyr as well) and share the alloc.c code. 

This "would" not work or this _did_ not work? (URL to build logs?)

I hope it's the latter otherwise we have a big IMX coverage gap in `.github/` @dbaluta

marc-hb:
> Still checkpatch to address

I think this can be rephrased in a slightly more positive way :-)

Right now _ALL CI checks are green_ except for a single, small `--subjective` checkpatch warning at https://github.com/thesofproject/sof/actions/runs/3575165092/jobs/6011351397 and the usual cache coherency warnings on MTL at https://github.com/thesofproject/sof/actions/runs/3575165077/jobs/6011349216

The checkpatch warning can probably be fixed but I doubt anyone will lose sleep over it.

PS: Please always link to build logs because they're hard/impossible to find after a force push.

src/platform/meteorlake/include/platform/lib/memory.h
btian1: here HP heap size is 256KB, how to align this with print out 1362KB total with zephyr build?

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -33,7 +33,7 @@ CONFIG_INTEL_ADSP_TIMER=y
 CONFIG_MM_DRV_INTEL_ADSP_TLB_REMAP_UNUSED_RAM=y
```
kv2019i, line 33: @marcinszkudlinski No need to update this PR, but the git commit messages look a bit odd. The text should be line-wrapped at 75 columns (originating from Linux coding guidelines). I.e.

"L3 memory should support have support
for allocating/freeing memory
A heap support in L3 is introduced by
this commit"

-> 

""
L3 memory should support have support for allocating/freeing
memory A heap support in L3 is introduced by this commit
""

kv2019i (approved):
Thanks, looks good now. And CI job still running, but others look good.

marcinszkudlinski:
Legacy platforms compilation cleared, please proceed with merging

abonislawski (approved):
It would be nice if the commit titles were slightly modified to match the current naming convention

PR #6614, opened 2022-11-16
platform: kconfig: Introduce MP_NUM_CPUS config

paulstelian97:
This config is added as a refactor to make xtos behave more like Zephyr. It is needed to avoid a warning on the later use of MP_NUM_CPUS in CORE_COUNT on xtos builds.

This configuration would end up having a double definition when performing Zephyr builds; that should be fine.

Draft because I also intend to deprecate the regular CORE_COUNT config, so that MP_NUM_CPUS is used by all of the code within SOF instead of it.

### DIFF
--- a/src/platform/Kconfig
+++ b/src/platform/Kconfig
@@ -304,6 +304,13 @@ config MAX_CORE_COUNT
 	help
 	  Maximum number of cores per configuration
 
+config MP_NUM_CPUS
+	int "Number of cores (Zephyr compatibility)"
+	default MAX_CORE_COUNT
+	range 1 MAX_CORE_COUNT
+	help
+	  Number of cores defined on Zephyr. Used for compatibility
+
 config CORE_COUNT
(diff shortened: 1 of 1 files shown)


src/platform/Kconfig
```diff
@@ -304,6 +304,13 @@ config MAX_CORE_COUNT
 	help
 	  Maximum number of cores per configuration
 
+config MP_NUM_CPUS
```
lgirdwood, line 307: Is there any way we can avoid the duplication with Zephy ? i.e.  could the Kconfig item depend on XTOS == true and would reusing the Zephyr Kconfig symbol for xtos help here ? Just want to avoid any confusion around Zephyr Kconfig (which we had in the past and had to fix). @kv2019i any comments ?

paulstelian97:
Closing for now, no focus on this. Will reopen when I go back to the big issue in which this is a part of.

PR #6618, opened 2022-11-16
ipc4: handler: Always invoke platform_context_save() when entering D3

ranj063:
The host driver ignores the IPC errors and continues to power off the DSP after sending the request to power off the primary core. So, always platform_context_save() even in case of errors to avoid unexpected behaviour during resume.

### DIFF
--- a/src/ipc/ipc4/handler.c
+++ b/src/ipc/ipc4/handler.c
@@ -839,16 +839,12 @@ static int ipc4_module_process_dx(struct ipc4_message_request *ipc4)
 
 	/* Deactivating primary core if requested.  */
 	if (dx_info.core_mask & BIT(PLATFORM_PRIMARY_CORE_ID)) {
-		if (cpu_enabled_cores() & ~BIT(PLATFORM_PRIMARY_CORE_ID)) {
+		if (cpu_enabled_cores() & ~BIT(PLATFORM_PRIMARY_CORE_ID))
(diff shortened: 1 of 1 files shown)


lgirdwood (commented):
This means the host driver wont know the BUSY FW state at D3 ? I think we still need to return the BUSY status (as the driver would know to take any corrective action).

tmleman (changes requested):
How it will behave in a situation when we do save/restore context of a active ppl? @marcinszkudlinski, any thoughts?

I think that if we want to power-off DSP after error on SetDx ipc, power-up should be done with purge option enabled.

marcinszkudlinski (changes requested):
We can't go to D3 if pipelines are running. 
FW must fail the flow in case it is in a wrong state - it won't be able to restore correctly from IMR 
Driver, however, may power it off regardless of IPC status and perform a full FW load / boot flow. Anyway, the firmware MUST NOT try to enter D3 if in wrong state

ranj063:
> We can't go to D3 if pipelines are running. FW must fail the flow in case it is in a wrong state - it won't be able to restore correctly from IMR Driver, however, may power it off regardless of IPC status and perform a full FW load / boot flow. Anyway, the firmware MUST NOT try to enter D3 if in wrong state

Thanks for the confirmation @marcinszkudlinski. Let me see how to handle this on this host then.

PR #6623, opened 2022-11-17
Tools: Topology1: Add sof-hda-generic topology with SRC

singalsu:
This patch adds to sof-hda-generic.m4 macros HSSFX, HSSFX_FILTER1, and HSSFX_FILTER2 those can be used to customize stream effect for PCM 30. The macro defaults to volume so the default topology build is not impacted.

The CMakeLists.txt in development is updated to build topologies sof-hda-generic-src.tplg, sof-hda-generic-2ch-src.tplg, and sof-hda-generic-4ch-src.tplg. The stream effect is set with HSSFX to src-volume.

The pipeline pipe-host-src-volume-playback.m4 is added. It is similar as pipe-host-volume-playback.m4 but a sample rate converter (SRC) is added after PCM, before volume. The PCM capability is changed to min 8000, max 192000 Hz.

### DIFF
--- a/tools/topology/topology1/development/CMakeLists.txt
+++ b/tools/topology/topology1/development/CMakeLists.txt
@@ -33,7 +33,6 @@ set(TPLGS
 	"sof-imx8mp-compr-pcm-cap-wm8960\;sof-imx8mp-compr-pcm-cap-wm8960"
 	"sof-apl-nocodec-demux-eq-4ch4ch\;sof-apl-nocodec-demux-eq-4ch4ch"
 	"sof-apl-nocodec-demux-eq-2ch4ch\;sof-apl-nocodec-demux-eq-2ch4ch"
-
 	"sof-hda-generic-kwd\;sof-hda-generic-2ch-kwd\;-DCHANNELS=2\;-DDYNAMIC=1"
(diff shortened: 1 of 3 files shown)


lgirdwood (commented):
@singalsu whats the default impact of this change i.e. is the customization an opt-in for only certain devices ?

singalsu:
> @singalsu whats the default impact of this change i.e. is the customization an opt-in for only certain devices ?

We don't have a way to test SRC elsewhere but in APL UP2 and since the platform is no more used in tests, there's no tests done for SRC. Though we do have a bug in HDA drivers system (pipelines propagate wrong HDA DAI rate) so SRC can't work correctly. But to fix things a topology to see the issue is useful.

lgirdwood:
@singalsu ok, so which platforms does this impact ?

singalsu:
> @singalsu ok, so which platforms does this impact ?

The change impacts none of the platforms. It creates to development directory three new topologies those can be used optionally.

The sof-hda-generic.m4 that is edited is used by all HDA topologies, so need to be very careful with it. But checked I by diffing before and after the .conf files to ensure there's only white space changes.

lgirdwood (approved):

ranj063 (approved):

PR #6624, opened 2022-11-17
hw hda chain managment

makarukp:
This is only temporary draft PR (work in progress)


src/include/ipc/topology.h
```diff
@@ -28,6 +28,7 @@ enum sof_comp_type {
 	SOF_COMP_NONE = 0,
 	SOF_COMP_HOST,
 	SOF_COMP_DAI,
+	SOF_COMP_CHAIN_DMA,
```
plbossart, line 31: can we do this @ranj063 @jsarha ? This would break backwards-compatibility with the kernel. This should be added line 52

I also thought we'd stop adding new component types and use UUIDs now?

lgirdwood: Yes - we should use UUIDs now. IIUC Windows should also use an UUID for this ? @marcinszkudlinski ?

jsarha: My understanding of the topology system is still quite limited, but while browsing the code I could not immediately tell why this would not work with the current driver side implementation. The component IDs are not present in IPC4 messages.

I also tried to check this with brute force method by testing it, but the build failed with this error:

/home/oku/work/sof-zephyr-new/sof/src/ipc/ipc4/helper.c:593:10: fatal error: sof/mem_win_debug.h: No such file or directory
  593 | #include "sof/mem_win_debug.h"
      |          ^~~~~~~~~~~~~~~~~~~~~

Any idea what I am missing?

src/ipc/ipc4/handler.c
```diff
@@ -458,36 +459,101 @@ static int ipc4_load_library(struct ipc4_message_request *ipc4)
...
+	const bool allocate = cdma.primary.r.allocate;
+	const bool enable = cdma.primary.r.enable;
+	const bool scs = cdma.primary.r.scs;
+
+	MEM_WIN_DEBUG0(0xaaa0,0,0,0);
+	int comp_id = IPC4_COMP_ID(host_dma_id + IPC4_MAX_MODULE_COUNT,
+			link_dma_id);
+
+    if (allocate)
+    {
```
lgirdwood, line 486: coding style is 
```c
if (condition) {
        some logic
}
```

src/ipc/ipc4/handler.c
```diff
@@ -458,36 +459,101 @@ static int ipc4_load_library(struct ipc4_message_request *ipc4)
...
+		ret = ipc4_create_io_driver_for_chain(&cdma);
+		struct ipc_comp_dev* cdma_comp = ipc_get_comp_by_id(ipc, comp_id);
+		MEM_WIN_DEBUG0(0xaab0,cdma_comp,0,0);
```
lgirdwood, line 489: debug ?

src/ipc/ipc4/handler.c
```diff
@@ -458,36 +459,101 @@ static int ipc4_load_library(struct ipc4_message_request *ipc4)
...
+		assert(cdma_comp != NULL);
```
lgirdwood, line 490: pls dont assert unless FW is unrecoverable. We should emit some error and return the error code up the stack.

src/ipc/ipc4/handler.c
```diff
@@ -458,36 +459,101 @@ static int ipc4_load_library(struct ipc4_message_request *ipc4)
...
+		MEM_WIN_DEBUG0(0xaab2,drv,0,0);
+    }
+    else
+    {
+		const struct comp_driver *drv;
+		drv = ipc4_get_drv((uint8_t *)&uuid);
+		struct ipc_comp_dev* cdma_comp = ipc_get_comp_by_id(ipc, comp_id);
+        if (enable)
+        {
+            return IPC4_INVALID_REQUEST;
```
lgirdwood, line 520: indentaion looks off, we should be using tab size 8.

src/ipc/ipc4/handler.c
```diff
@@ -458,36 +459,101 @@ static int ipc4_load_library(struct ipc4_message_request *ipc4)
...
+			drv->ops.trigger(cdma_comp->cd, COMP_TRIGGER_PAUSE);
+        }
+        if (ret == IPC4_SUCCESS)
+        {
+            //ret = chain_dma_mgr->RemoveChain(host_dma_id);
+			drv->ops.free(cdma_comp->cd);
+        }
+    }
+
+	//PREVIOUS IMPLEMENTATION
```
lgirdwood, line 534: do we need to keep this ?

src/ipc/ipc4/helper.c
```diff
@@ -590,6 +590,43 @@ static int construct_config(struct ipc4_copier_module_cfg *copier_cfg, uint32_t
...
+
+	ret = process_dma_index(cdma->primary.r.host_dma_id, &dir, &host_chan);
+	if (ret != IPC4_SUCCESS)
+		return ret;
+
+	ret = process_dma_index(cdma->primary.r.link_dma_id, &dir, &link_chan);
+	if (ret != IPC4_SUCCESS)
+		return ret;
+
+	dev = drv->ops.chain_dma_create(drv, (uint8_t)host_chan, (uint8_t)link_chan,
```
lgirdwood, line 617: Lets use the API here (which does the NULL check etc).

src/ipc/ipc4/helper.c
```diff
@@ -590,6 +590,43 @@ static int construct_config(struct ipc4_copier_module_cfg *copier_cfg, uint32_t
...
+					cdma->extension.r.fifo_size, cdma->primary.r.scs);
+
+	if (!dev)
+		return NULL;
+
+	ipc4_add_comp_dev(dev);
+	//dev->ipc_config.id
```
lgirdwood, line 624: needed ?

src/ipc/ipc4/helper.c
```diff
@@ -590,6 +590,43 @@ static int construct_config(struct ipc4_copier_module_cfg *copier_cfg, uint32_t
...
+
+	return IPC4_SUCCESS;
+}
+
+//deprecated
```
lgirdwood, line 629: can it be removed ?

src/ipc/ipc4/helper.c
```diff
@@ -793,6 +831,7 @@ int ipc4_trigger_chain_dma(struct ipc *ipc, struct ipc4_chain_dma *cdma)
 	return IPC4_SUCCESS;
 }
 
+#include "mem_win_debug.h"
```
lgirdwood, line 834: it looks like a of of debug logic has been left in this PR and could be removed ?

lgirdwood (commented):
Some cleanup of debug, code style and indentation is needed. I would split out the west update into a separate PR so this can be separately tested by validation prior to merge.

src/include/ipc/topology.h
jsarha: Managed to build the branch by removeing the win debug code, but the FW still does not work. e.g. its not only HDMI audio that is broken, but nothing really works. I am probably missing something.

src/ipc/ipc4/handler.c
makarukp: This is only draft under debug.

makarukp:
This is only draft PR under debug. 
Please do not test it yet!

src/include/ipc/topology.h
makarukp: SOF_COMP_CHAIN_DMA removed.
Resolved in:
image

src/ipc/ipc4/handler.c
makarukp: Rsolved in:
image

src/ipc/ipc4/helper.c
makarukp: Reolved in:
image

src/ipc/ipc4/helper.c
makarukp: Resolved in:
image

makarukp:
I have to close that PR. Continuation of issue:
image

PR #6626, opened 2022-11-17
[SKIP CI] .github: move sparse job to a separate workflow

marc-hb:
Zero code change, pure move to a new file.

Similarity is not a very strong reason for the sparse job to be in the same file as regular Zephyr compilation.

Unlike plain Zephyr compilation, the sparse job is very much a "work in progress" which is currently failing all the time. Moving it to a different file/workflow provides more configuration flexibility.


marc-hb:
checkpatch has the usual and stupid "WARNING: A patch subject line should describe the change not the tool that found it". Everything else is fine.

lgirdwood (approved):

PR #6628, opened 2022-11-17
[SKIP CI] .github: create Sparse Zephyr workflow

marc-hb:
First attempt in https://github.com/thesofproject/sof/pull/6626  but afraid brand new workflows have to be created from the web UI first...


marc-hb:
> but afraid brand new workflows have to be created from the web UI first...

Nope, there was a YAML mistake in 6626. As usual no error message from github at all :-(

PR #6636, opened 2022-11-18
Audio: Simplify volume ramp functions

singalsu:
This patch saves about 15% of processing time or about 0.9 MCPS
    during volume ramp.
    
    The volume_ramp() functions are inlined to volume_process() function. The
    other use of function in idle is replaced by volume_ramp_check().
    
    In ramp calculation the division in ramp time calculation is converted
    to faster multiplication with inverse of sample rate.
    
    The unnecessary checks for vol greater than vol_max or vol_min are
    eliminated by ensuring that tvolume[] is limited to vol_min and vol_max.
    
    The clear of ramp_coef[] at ramp finish is unnecessary. It's
    internal for linear ramp and is not used by the other ramp type.
    
    The other optimization is to pass to ramp math functions struct
    vol_data pointer instead of processing_module pointer to direct
    access to needed state variables.
    
### DIFF
--- a/src/audio/module_adapter/module/volume/volume.c
+++ b/src/audio/module_adapter/module/volume/volume.c
@@ -232,10 +232,8 @@ static const struct comp_zc_func_map zc_func_map[] = {
  * \param[in] ramp_time Time spent since ramp start as milliseconds Q29.3
  * \param[in] channel Current channel to update
  */
-static int32_t volume_linear_ramp(struct processing_module *mod, int32_t ramp_time, int channel)
+static inline int32_t volume_linear_ramp(struct vol_data *cd, int32_t ramp_time, int channel)
 {
-	struct vol_data *cd = module_get_private_data(mod);
-
 	return cd->rvolume[channel] + ramp_time * cd->ramp_coef[channel];
 }
 #endif
@@ -248,10 +246,8 @@ static int32_t volume_linear_ramp(struct processing_module *mod, int32_t ramp_ti
  * \param[in] channel Current channel to update
  */
 
-static int32_t volume_windows_fade_ramp(struct processing_module *mod, int32_t ramp_time,
-					int channel)
+static inline int32_t volume_windows_fade_ramp(struct vol_data *cd, int32_t ramp_time, int channel)
 {
-	struct vol_data *cd = module_get_private_data(mod);
 	int32_t time_ratio; /* Q2.30 */
 	int32_t pow_value; /* Q2.30 */
 	int32_t volume_delta = cd->tvolume[channel] - cd->rvolume[channel]; /* Q16.16 */
@@ -264,86 +260,92 @@ static int32_t volume_windows_fade_ramp(struct processing_module *mod, int32_t r
 
 /**
  * \brief Ramps volume changes over time.
- * \param[in,out] mod Volume processing module handle
-
+ * \param[in,out] vol_data Volume component data
  */
-static void volume_ramp(struct processing_module *mod)
+
+/* Note: Using inline saves 0.4 MCPS */
+static inline void volume_ramp(struct vol_data *cd)
 {
-	struct vol_data *cd = module_get_private_data(mod);
-	struct comp_dev *dev = mod->dev;
-	int32_t vol;
-	int32_t ramp_time;
+	int32_t new_vol;
+	int32_t tvolume;
+	int32_t volume;
 	int i;
-	bool ramp_finished = true;
 
+	cd->ramp_finished = true;
 	cd->copy_gain = true;
-	/* No need to ramp in idle state, jump volume to request. */
-	if (dev->state == COMP_STATE_READY) {
-		for (i = 0; i < PLATFORM_MAX_CHANNELS; i++)
-			cd->volume[i] = cd->tvolume[i];
-
-		cd->ramp_finished = true;
-		return;
-	}
-
-	/* The first is set and cleared to indicate ongoing ramp, the
-	 * latter is set once to enable self launched ramp only once
-	 * in stream start.
-	 */
-	cd->vol_ramp_active = true;
 
 	/* Current ramp time in Q29.3 milliseconds. Note that max. ramp length
 	 * can be 1.3s at 192 kHz rate and 5.5s at 48 kHz rate without
-	 * exceeding int32_t range. Value 8000 is 1000 for converting to
-	 * milliseconds times 8 (2 ^ 3) for fraction.
+	 * exceeding int32_t range. Inverse of sample rate is 1000/sample_rate
+	 * for milliseconds.
 	 */
-	ramp_time = cd->vol_ramp_elapsed_frames * 8000 / cd->sample_rate;
+#if defined CONFIG_COMP_VOLUME_WINDOWS_FADE || defined CONFIG_COMP_VOLUME_LINEAR_RAMP
+	int32_t ramp_time = Q_MULTSR_32X32((int64_t)cd->vol_ramp_elapsed_frames,
+					   cd->sample_rate_inv, 0, 31, 3);
+#endif
 
 	/* Update each volume if it's not at target for active channels */
 	for (i = 0; i < cd->channels; i++) {
 		/* skip if target reached */
-		if (cd->volume[i] == cd->tvolume[i])
+		volume = cd->volume[i];
+		tvolume = cd->tvolume[i];
+		if (volume == tvolume)
 			continue;
 
 		/* Update volume gain with ramp. The ramp gain value is
 		 * calculated from previous gain and ramp time. The slope
 		 * coefficient is calculated in volume_set_chan().
 		 */
-		vol = cd->ramp_func(mod, ramp_time, i);
-		if (cd->volume[i] < cd->tvolume[i]) {
+#if defined CONFIG_COMP_VOLUME_WINDOWS_FADE && defined CONFIG_COMP_VOLUME_LINEAR_RAMP
+		if (cd->ramp_type == SOF_VOLUME_WINDOWS_FADE)
+			new_vol = volume_windows_fade_ramp(cd, ramp_time, i);
+		else
+			new_vol = volume_linear_ramp(cd, ramp_time, i);
+#elif defined CONFIG_COMP_VOLUME_WINDOWS_FADE
+		new_vol = volume_windows_fade_ramp(cd, ramp_time, i);
+#elif defined CONFIG_COMP_VOLUME_LINEAR_RAMP
+		new_vol = volume_linear_ramp(cd, ramp_time, i);
+#else
+		new_vol = tvolume;
+#endif
+		if (volume < tvolume) {
 			/* ramp up, check if ramp completed */
-			if (vol >= cd->tvolume[i] || vol >= cd->vol_max) {
-				cd->ramp_coef[i] = 0;
-				cd->volume[i] = cd->tvolume[i];
-			} else {
-				ramp_finished = false;
-				cd->volume[i] = vol;
-			}
+			if (new_vol < tvolume)
+				cd->ramp_finished = false;
+			else
+				new_vol = tvolume;
 		} else {
 			/* ramp down */
-			if (vol <= 0) {
-				/* cannot ramp down below 0 */
-				cd->volume[i] = cd->tvolume[i];
-				cd->ramp_coef[i] = 0;
-			} else {
-				/* ramp completed ? */
-				if (vol <= cd->tvolume[i] ||
-				    vol <= cd->vol_min) {
-					cd->ramp_coef[i] = 0;
-					cd->volume[i] = cd->tvolume[i];
-				} else {
-					ramp_finished = false;
-					cd->volume[i] = vol;
-				}
-			}
+			if (new_vol > tvolume)
+				cd->ramp_finished = false;
+			else
+				new_vol = tvolume;
 		}
-	}
-
-	if (ramp_finished) {
-		cd->ramp_finished = true;
-		cd->vol_ramp_active = false;
+		cd->volume[i] = new_vol;
 	}
 }
+
+/**
+ * \brief Ramps volume changes over time.
+ * \param[in,out] mod Volume processing module handle
+ */
+#if CONFIG_IPC_MAJOR_3
+static void volume_ramp_check(struct processing_module *mod)
+{
+	struct vol_data *cd = module_get_private_data(mod);
+	struct comp_dev *dev = mod->dev;
+	int i;
+
+	/* No need to ramp in idle state, jump volume to request. */
+	cd->ramp_finished = false;
+	if (dev->state == COMP_STATE_READY) {
+		for (i = 0; i < PLATFORM_MAX_CHANNELS; i++)
+			cd->volume[i] = cd->tvolume[i];
+
+		cd->ramp_finished = true;
+	}
+}
+#endif
 
 /**
  * \brief Reset state except controls.
@@ -359,10 +361,9 @@ static void reset_state(struct vol_data *cd)
 
 	cd->channels = 0;
 	cd->ramp_finished = true;
-	cd->vol_ramp_active = false;
 	cd->vol_ramp_frames = 0;
 	cd->vol_ramp_elapsed_frames = 0;
-	cd->sample_rate = 0;
+	cd->sample_rate_inv = 0;
 	cd->copy_gain = true;
 }
 
@@ -437,21 +438,17 @@ static int volume_init(struct processing_module *mod)
 		cd->muted[i] = false;
 	}
 
-	cd->ramp_type = vol->ramp;
-	cd->initial_ramp = vol->initial_ramp;
-
 	switch (cd->ramp_type) {
 #if CONFIG_COMP_VOLUME_LINEAR_RAMP
 	case SOF_VOLUME_LINEAR:
 	case SOF_VOLUME_LINEAR_ZC:
-		cd->ramp_func = &volume_linear_ramp;
-		break;
 #endif
 #if CONFIG_COMP_VOLUME_WINDOWS_FADE
 	case SOF_VOLUME_WINDOWS_FADE:
-		cd->ramp_func = &volume_windows_fade_ramp;
-		break;
 #endif
+		cd->ramp_type = vol->ramp;
+		cd->initial_ramp = vol->initial_ramp;
+		break;
 	default:
 		comp_err(dev, "volume_new(): invalid ramp type %d", vol->ramp);
 		rfree(cd);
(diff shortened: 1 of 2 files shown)


singalsu:
> @singalsu whats the MCPS improvement ?

Currently tiny, only a fraction of MCPS from about +3 MCPS when ramp is executed. I wonder if the cache trash impact from this function to hot code part in volume shadows all improvement. Experiment with doubled update period 125 us -> 250 us could be worth it. I was maybe too critical when I tried to hear zipper effect i volume adjust. Opinion from a larger group could be useful.

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -293,49 +291,39 @@ static void volume_ramp(struct processing_module *mod)
 
 	/* Current ramp time in Q29.3 milliseconds. Note that max. ramp length
 	 * can be 1.3s at 192 kHz rate and 5.5s at 48 kHz rate without
-	 * exceeding int32_t range. Value 8000 is 1000 for converting to
-	 * milliseconds times 8 (2 ^ 3) for fraction.
+	 * exceeding int32_t range. Inverse of sample rate is 1000/sample_rate
+	 * for milliseconds.
```
btian1, line 295: I am not quite understanding why ramp time still need a  Q format, possible to set ramp time with a fixed like 20-50ms? ramp functionality is for reduce pop noise and make volume change smooth, if we have fixed time, then based on sample rate, we will know how many samples need process with ramp, then a counter will be set, it would be much simpler.

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -678,18 +667,18 @@ static inline int volume_set_chan(struct processing_module *mod, int chan,
...
 	 * can be useful to prevent totally muted small volume gain.
 	 */
-	if (v < VOL_MIN) {
+	if (v < cd->vol_min) {
 		/* No need to fail, just trace the event. */
 		comp_warn(mod->dev, "volume_set_chan: Limited request %d to min. %d",
-			  v, VOL_MIN);
-		v = VOL_MIN;
+			  v, cd->vol_min);
+		v = cd->vol_min;
```
btian1, line 674: there are other place also set vol_min and max, we are better to check condition is possible, if not possible, I would suggest remove the check.

btian1:
> 

For ramp, I  reported a ramp disfunctional bug to sof kernel team, Rander is working on it:
https://github.com/thesofproject/linux/issues/4026
I hard coded ramp in local and test performance with real time, it is around 0.9 for window ramp.

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -293,49 +291,39 @@ static void volume_ramp(struct processing_module *mod)
...
-		vol = cd->ramp_func(mod, ramp_time, i);
-		if (cd->volume[i] < cd->tvolume[i]) {
+		new_vol = cd->ramp_func(cd, ramp_time, i);
+		if (volume < tvolume) {
 			/* ramp up, check if ramp completed */
-			if (vol >= cd->tvolume[i] || vol >= cd->vol_max) {
-				cd->ramp_coef[i] = 0;
-				cd->volume[i] = cd->tvolume[i];
-			} else {
+			if (new_vol >= tvolume)
```
btian1, line 315: ramp_func should make sure this can't be happen.

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -1092,14 +1079,14 @@ static int volume_process(struct processing_module *mod,
 			frames = cd->vol_ramp_frames;
 		}
 
-		/* copy and scale volume */
-		cd->scale_vol(mod, &input_buffers[0], &output_buffers[0], frames);
+		if (!cd->ramp_finished)
```
btian1, line 1082: I like put it ahead of volume, and if the condition based on ramp_sample_left will be better.

src/include/sof/audio/volume.h
```diff
@@ -146,6 +148,7 @@ struct vol_data {
 	int32_t vol_min;			/**< minimum volume */
 	int32_t vol_max;			/**< maximum volume */
 	int32_t	vol_ramp_range;			/**< max ramp transition */
+	int32_t sample_rate_inv;		/**< 1000x inverse of sample rate as Q1.31 */
```
btian1, line 151: better find another way to set ramp time instead of add new var to do calculation?
any concerns to set ramp time with a fixed or a table with different value, like 8k to 100ms, 48k use 20ms? just a example.

singalsu: The rate inverse coefficient is calculated only once in prepare(). After that one in ramp update the sample counter is multiplied by the coefficient. The ramp functions need time, so I don't see how this part could be made any simpler with lookup table. The ramp length is arbitrary with 1 ms steps. I'd not like to pre-define some ramp times.

src/audio/module_adapter/module/volume/volume.c
singalsu: Ramp_func does not use tvolume, but rvolume, it may hit exactly tvolume or exceed it slightly due limited precision. I don't see improvement from moving this check to ramp_func. Or is there, how?

src/audio/module_adapter/module/volume/volume.c
singalsu: Yep, agree. I'd like to get rid of duplicate ramp_finished and vol_ramp_active but I'm not sure how to do it. So I'm keeping this still draft. Also I have problems with check_volume_levels.sh stability so I'm not 100% this in stress test always passes.

singalsu:
@lgirdwood @btian1 Thanks for reviewing. Please check this new version also.

I took today another shot to this and got some better 0.5 MCPS saving during ramp. The linear ramp still adds 2.5 MCPS to normal volume load, almost double. Windows fade probably more, I haven't measured.

src/audio/module_adapter/module/volume/volume.c
singalsu: Ramp is relative time and independent from stream rate. If the linear and windows ramp would take sample count as parameter they would any way need to compensate the ramp step to stream sample rate. Also the ramp update rate is independent of sample rate. The fastest ramp update is 8 kHz (to avoid zipper noise) independently of stream rate. But we have also 2 kHz and 1 kHz ramp updates.

singalsu: So, in other words, simplification of this is hard.

src/audio/module_adapter/module/volume/volume.c
singalsu: Yes possibly, this is not speed critical code. I'll see if this can be improved.

singalsu:
> For ramp, I reported a ramp disfunctional bug to sof kernel team, Rander is working on it: [thesofproject/linux#4026](https://github.com/thesofproject/linux/issues/4026) I hard coded ramp in local and test performance with real time, it is around 0.9 for window ramp.

If it was direct table lookup, it removes totally ramp customization and sample rate independence.

src/include/sof/audio/volume.h
```diff
@@ -231,6 +234,71 @@ static inline void peak_vol_update(struct vol_data *cd)
...
+			/* ramp down */
+			if (new_vol <= tvolume)
+				new_vol = tvolume;
+			else
+				ramp_finished = false;
+		}
+		cd->volume[i] = new_vol;
+	}
+
+	if (ramp_finished) {
```
btian1, line 296: ramp_finished is for all channel, however, setting is based on per channel, i.e, as long as one channel  need ramp, ramp is happening, is it expected with design?

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -261,86 +257,18 @@ static int32_t volume_windows_fade_ramp(struct processing_module *mod, int32_t r
 }
 #endif
 
-/**
- * \brief Ramps volume changes over time.
- * \param[in,out] mod Volume processing module handle
-
- */
-static void volume_ramp(struct processing_module *mod)
```
lyakh, line 269: moving this function to a header shouldn't be needed, the compiler should inline it automatically in .c too when building with optimisation. Could you check that keeping the function here doesn't reduce your performance gains? Keeping the function here would also help reduce the size of the patch and make changes clearer. In fact it would be even better to split this patch into separate optimisations, but at least making changes more obvious by keeping the function in place would be good

singalsu: I actually saw performance improvement from the change. It was like inline wasn't happening.

Yep, such changes are good for separate commit.

I'm actually now thinking to remove the function call pointer for ramp function and change it to if - else. Then I could inline the ramp math also.

singalsu:
> > 
> 
> When 2.5 mcps happened, did you add ZC function? if added, then we can try to optimize it. Also linear itself should not cause so big mcps cost.

The 2.5 MCPS is with linear. The high cost is surprising and I think there must be some cache performance issue. I will test by inlining also the ramp math (remove function call pointer).

src/include/sof/audio/volume.h
singalsu: The ramp channels loop function is called until every channel has completed. However typically the channel gains are the same value so there is not much help if try to optimize the single channel update. Also finding similar channels for 2ch might not help much. With larger channels count maybe.

src/audio/module_adapter/module/volume/volume.c
singalsu: Sorry, no improvement from function move to volume.h. The reason was some other change, I'll return the function to volume.c.

singalsu:
I in-lined the ramp math too, now the saving is about 0.9 MCPS. Below is a figure of two PGAs 1.2 and 30.67 from s32 playback to hda-generic headset. Upper pictures are before, lower are after. The higher part of the MCPS plot is 20 ms ramp, the lower part is the normal operation after ramp.

Screenshot from 2022-11-28 17-38-05

singalsu:
I think the above per 1 ms trace computed MCPS is not very suitable for low loads. But maybe the observation of 1.5x load during ramp is normal. Below is the per 1024 ms output performance counters based MCPS (trace value x 400/38400). If average is 2.3 or 2.6 MCPS then during ramp the load would be 3.5 or 3.9 MCPS.

Screenshot from 2022-11-28 18-20-23

src/audio/module_adapter/module/volume/volume.c
btian1: if slightly exceed, is this also expected? I mean it may not have impact on final volume, 10000/10001, there is no difference for volume output.

I am also fine to keep this.

src/audio/module_adapter/module/volume/volume.c
btian1: > 

0.5 MCPS, how this calculated out? this patch more focused on code refine, I don't think there is 0.5 mcps improvement. window ramp itself cost about 0.9 as I measured without this PR.

src/audio/module_adapter/module/volume/volume.c
singalsu: The ramp time is quantized to multiple of 125 us, so the final point can be off a bit. The more, the steeper the ramp is. The check if the ramp is up or down is needed to clamp the gain to not go over or under so I don't see simplification possibility for this.

src/audio/module_adapter/module/volume/volume.c
singalsu: The 0.5 MCPS was from that time 1 ms copy() length traced timestamps. Now with more code in-lined the improvement seems to be 0.9 MCPS same way.  I tried also perf counters but it's not suitable  to check a 20 ms long operation mode. If I would change in topology ramp to longer the load would be lower because the ramp update is less frequent (1/2 or 1/4) with the less steep ramps. It seems that the overhead of 1 ms traces is quite large. I get base 4.3 - 4.5 MCPS that way for volume while perf counters give 2.3 - 2.6 MCPS. I assumed that perf counters measure 38.4 MHz timer ticks and that DSP clock in TGL-Zephyr-IPC3 is 400 MHz. Is that correct?

singalsu: Also please check in the bottom of this discussion the plots, there's the evidence of MCPS saving.

src/audio/module_adapter/module/volume/volume.c
btian1: you can create an jira ticket or github issue, then we can do it when we have free time: to simply those flags.

src/audio/module_adapter/module/volume/volume.c
btian1: yes, this is due to log print takes a lot of time, if you use 1ms to measure, you will get overestimated data, 1024ms is better, maybe in future, we can continue enlarge it to 10240ms for more accurate data, log print is really a big overhead to sof.

btian1:
With this PR, can we say we lower down volume module 20% performance? perviously, I downed 5%, so if this PR can save 10%, left 5-10% can be saved by more advanced HIFI4 architecture.

singalsu:
> With this PR, can we say we lower down volume module 20% performance? perviously, I downed 5%, so if this PR can save 10%, left 5-10% can be saved by more advanced HIFI4 architecture.

Maybe more safe to say this reduces by 14-16% the worst-case.

Also I tried as quick test to change the linear ramp multiplication to shift as fast approximation of multiply but the saving was minimal. So I didn't include any intrinsic multiplications. What might work additionally could be if we could change for selected components like volume the xcc optimization to -O3 and enable HiFi instructions for C. Now the optimization is -O2 for entire SOF.

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -1225,21 +1208,15 @@ static int volume_prepare(struct processing_module *mod)
...
 	 * to previous volume request. Copy() checks for ramp finished
 	 * and executes it if it has not yet finished as result of
 	 * driver commands. Ramp is not constant rate to ensure it lasts
 	 * for entire topology specified time.
 	 */
-	cd->ramp_finished = true;
+	cd->ramp_finished = false;
 	cd->channels = sink_c->stream.channels;
-	cd->sample_rate = sink_c->stream.rate;
+	cd->sample_rate_inv = (int32_t)((int64_t)1000 * INT32_MAX / sink_c->stream.rate);
```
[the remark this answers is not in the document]
singalsu: Done!

lgirdwood:
@singalsu looks like there are some build failures.

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -263,84 +259,83 @@ static int32_t volume_windows_fade_ramp(struct processing_module *mod, int32_t r
 
 /**
  * \brief Ramps volume changes over time.
- * \param[in,out] mod Volume processing module handle
-
+ * \param[in,out] vol_data Volume component data
  */
-static void volume_ramp(struct processing_module *mod)
+static inline void volume_ramp(struct vol_data *cd)
```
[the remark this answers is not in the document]
lyakh: hm, interesting, I'd literary expect the compiler to do that for us automatically. Since the gain isn't huge, to confirm that there's really a difference - could you perhaps just build the firmware with and without `inline` and check that the resulting image changed (in size)? At least `volume_windows_fade_ramp()` really shouldn't need `inline` since you take its address:
```
	cd->ramp_func =  &volume_windows_fade_ramp;
```

lgirdwood:
@singalsu stiil some build failure
```
[ 70%] Building C object CMakeFiles/sof.dir/src/audio/module_adapter/module/volume/volume.c.o
[ 70%] Building C object CMakeFiles/sof.dir/src/audio/module_adapter/module/volume/volume.c.o
staging/sof-tplg: directory
staging/sof-tplg
|-- sof-adl-sdw-max98373-rt5682.tplg
|-- sof-adl-max98360a-nau8825.tplg
|-- sof-adl-max98360a-rt5682-4ch.tplg
|-- sof-adl-max98360a-rt5682.tplg
|-- sof-adl-rt1015-nau8825.tplg
|-- sof-adl-max98390-rt5682.tplg
|-- sof-tgl-sdw-max98373-rt5682.tplg
|-- sof-tgl-max98357a-rt5682-pdm1-drceq.tplg
|-- sof-adl-max98360a-rt5682-2way.tplg
├── ...
..
mkdir -p staging/tools
cd /home/sof/work/sof.git/installer/../installer-builds/build_tools && \
  cp -p ctl/sof-ctl  logger/sof-logger  probes/sof-probes /home/sof/work/sof.git/installer/staging/tools
[ 70%] Building C object CMakeFiles/sof.dir/src/audio/copier/copier.c.o
[ 70%] Building C object CMakeFiles/sof.dir/src/audio/copier/copier.c.o
/home/sof/work/sof.git/src/audio/module_adapter/module/volume/volume.c: In function 'volume_ramp':
/home/sof/work/sof.git/src/audio/module_adapter/module/volume/volume.c:267:10: error: variable 'ramp_time' set but not used [-Werror=unused-but-set-variable]
  267 |  int32_t ramp_time;
      |          ^~~~~~~~~
/home/sof/work/sof.git/src/audio/module_adapter/module/volume/volume.c: In function 'volume_ramp':
/home/sof/work/sof.git/src/audio/module_adapter/module/volume/volume.c:267:10: error: variable 'ramp_time' set but not used [-Werror=unused-but-set-variable]
  267 |  int32_t ramp_time;
      |          ^~~~~~~~~
make[4]: *** [CMakeFiles/sof.dir/build.make:941: CMakeFiles/sof.dir/src/audio/module_adapter/module/volume/volume.c.o] Error 1
make[4]: *** Waiting for unfinished jobs....
make[4]: *** [CMakeFiles/sof.dir/build.make:941: CMakeFiles/sof.dir/src/audio/module_adapter/module/volume/volume.c.o] Error 1
make[4]: *** Waiting for unfinished jobs....
Scanning dependencies of target bootloader_dump
```

lgirdwood:
@singalsu can you check CI, thanks !

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -231,10 +231,8 @@ static const struct comp_zc_func_map zc_func_map[] = {
  * \param[in] ramp_time Time spent since ramp start as milliseconds Q29.3
  * \param[in] channel Current channel to update
  */
-static int32_t volume_linear_ramp(struct processing_module *mod, int32_t ramp_time, int channel)
+static inline int32_t volume_linear_ramp(struct vol_data *cd, int32_t ramp_time, int channel)
```
btian1, line 234: I have a comment here, seems we always use cd to indicate data, however, for copier, use cd make sense, how about here to use vd to indicate it is volume data?

lgirdwood (approved):

lgirdwood:
@singalsu still a failure - best to ping @wszypelt or @lrudyX

singalsu:
> @singalsu still a failure - best to ping @wszypelt or @lrudyX

Thanks, yes there's a gain test failure. I'll find out what exactly fails.

src/audio/module_adapter/module/volume/volume.c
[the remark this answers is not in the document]
lgirdwood: ok, so we dont have time to figure out why we get the XCC speed improvement with inline here, lets just mark this with an inline comment so that its obvious.

lgirdwood (commented):
LGTM, lets juts add the inline comment and we are good.

singalsu:
> LGTM, lets juts add the inline comment and we are good.

I just added the comment. Let's see if it passes now the CI.

btian1:
CI still failed, and title still have don't merge, please change title accordingly.

singalsu:
> CI still failed, and title still have don't merge, please change title accordingly.

The fail with TGLU_RVP_NOCODEC_IPC4ZPH seems to be a suspend test failure. I doubt it's related. The Internal Intel CI check that earlier failed has passed now.

lgirdwood (approved):

lgirdwood:
@andrula-song @kv2019i pls review

btian1:
> 

yes, same failure on my PR as well.

btian1 (approved):

andrula-song (approved):

kv2019i:
One system PM failure in https://sof-ci.01.org/sofpr/PR6636/build4370/devicetest/index.html , not related to volume ramp. Proceeding with merge.

PR #6641, opened 2022-11-21
Heap profiling: fix regression for heap profiling print with ipc3 case

btian1:
In zephyr ipc3 case, the first heap allocation come too early and zephyr logging even is not initizlized, it cause sof boot failure.

Limit heap status collection for zephyr+ipc4 only.

Signed-off-by: Baofeng Tian <baofeng.tian@intel.com>


singalsu (approved):
This works for me. Thanks!

dbaluta:
This is OK and we could merge it. The ideal fix would also solve the IPC3 issue.

dbaluta (approved):

lgirdwood (approved):

PR #6646, opened 2022-11-22
Fix sparse build by using our own sparse branch

marc-hb:
Use our own sparse branch, we cannot wait for critical sparse fixes like @lyakh's
`xtensa: switch to little endianness` https://marc.info/?l=linux-sparse&m=166861736724543
and others to be merged

### DIFF
--- a/.github/workflows/sparse-zephyr.yml
+++ b/.github/workflows/sparse-zephyr.yml
@@ -57,4 +57,6 @@ jobs:
           ZSDK=$(cat zsdk_location); _RCC=${{ matrix.platforms.real_cc }}
           REAL_CC="$ZSDK/$_RCC" ./sof/zephyr/docker-run.sh                          \
              ./sof/zephyr/docker-build.sh ${{ matrix.platforms.platform }}          \
-             --cmake-args=-DSPARSE=y | ./sof/scripts/parse_sparse_output.sh
+             --cmake-args=-DSPARSE=y 2>&1 | tee _.log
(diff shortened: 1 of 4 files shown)


marc-hb:
https://github.com/thesofproject/sof/actions/runs/3519164757/jobs/5898822295 is passing for TGL and showing `different address space` errors for MTL. Both show the complete log.

lyakh (approved):
looks good, thanks @marc-hb . Just one idea: would it be better to use the sparse fork from thesofproject directly? We could make a branch there for our waiting-to-be-merged patches. Although I don't expect many of them, after merging this one. Would we switch back to upstream sparse after this patch is merged?

lgirdwood (approved):

marc-hb:
> Just one idea: would it be better to use the sparse fork from thesofproject directly?

Absolutely yes, in fact I wrote a `TODO` in the `sparse-zephyr.yml` file about that. But first can you help update the https://github.com/thesofproject/sparse main branch with what you know we need?

marc-hb:
> Would we switch back to upstream sparse after this patch is merged?

Any switch is possible and easy in the future. Note both https://github.com/lucvoo/sparse and the version in the `zephyr-build` image usually lag behind https://git.kernel.org/pub/scm/devel/sparse/sparse.git

cc: @stephanosio, @tejlmand

marc-hb:
https://github.com/thesofproject/sparse does not have ANY patch right now because I'm not sure how we should manage it. So it can't be used yet. Are force-pushes to the main branch OK? Probably not, maybe we should use a non-default branch.

PR #6647, opened 2022-11-22
xt-clang: Do not use -std=gnu99 for C++

jxstelter:
Option -std=gnu99 applied to C++ files trigerrs warning for xcc compiler however throws error for xt-clang.
This patch limits usage of this flag to C, ASM code only.

### DIFF
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -895,7 +895,8 @@ find_package(Python3 COMPONENTS Interpreter)
 set(PYTHON3 "${Python3_EXECUTABLE}")
 
 # SOF uses GNU C99 extensions. TODO other flags required ?
-target_compile_options(SOF INTERFACE -std=gnu99 -fno-inline-functions)
+target_compile_options(SOF INTERFACE -fno-inline-functions)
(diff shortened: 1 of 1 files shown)


lgirdwood (approved):
@jxstelter do we need to set a minimum C++ standard ? I think @cujomalainey has some recent C++ code that may need a newish C++ standard.

jxstelter:
> @jxstelter do we need to set a minimum C++ standard ? I think @cujomalainey has some recent C++ code that may need a newish C++ standard.

I am not sure. I noticed that warnings on C++ files regarding std99 are treated as errors with xt-clang. This will block us when switching to newer xtensa toolchains.

lgirdwood:
@jxstelter can you check internal CI (need to use internal link atm). Thanks.

dbaluta (approved):

aborisovich (approved):
Looks great, @lgirdwood i would set C++ version in other PR

marc-hb:
Comments fix submitted in #6650

cujomalainey:
@perahgren who is using the C++ code from our side

jxstelter:
> @perahgren who is using the C++ code from our side

SOF for Intel ACE platform is using it.

PR #6649, opened 2022-11-22
lib: MCPS budget manager

kfrydryx:
Add initial implementation of mcps budget management.

Ideally it should be used to control cpu clock, so that it runs on the lowest frequency suitable for performed tasks.
Different parts of FW can request/free some budget using `core_cps_adjust` and clock should be changed accordingly. (i.e. when setting pipeline to run we should request budget required by its modules).

Currently there is no proper performance data to be used, so it's advised to request some artificial, high budget on platform init to keep the clock running fast and avoid performance issues.

Requires https://github.com/zephyrproject-rtos/zephyr/pull/52205 to work on mtl

### DIFF
--- a/src/audio/base_fw.c
+++ b/src/audio/base_fw.c
@@ -8,6 +8,8 @@
 #include <ipc4/pipeline.h>
 #include <ipc4/logging.h>
 #include <sof_versions.h>
+#include <sof/lib/cpu-clk-manager.h>
+#include <sof/lib/cpu.h>
 
 #if CONFIG_ACE_V1X_ART_COUNTER || CONFIG_ACE_V1X_RTC_COUNTER
 #include <zephyr/device.h>
@@ -323,6 +325,71 @@ static uint32_t basefw_get_ext_system_time(uint32_t *data_offset, char *data)
 	return IPC4_UNAVAILABLE;
 }
 
+static int basefw_register_kcps(bool first_block,
+				bool last_block,
+				uint32_t data_offset_or_size,
+				const char *data)
+{
+	if (!(first_block && last_block))
+		return -EINVAL;
+
+	/* value of kcps to request on core 0. Can be negative */
+	if (core_kcps_adjust(0, *(int32_t *)data))
+		return -EINVAL;
+
+	return 0;
+}
+
+static int basefw_kcps_allocation_request(struct ipc4_resource_kcps *request)
+{
+	if (core_kcps_adjust(request->core_id, request->kcps))
+		return -EINVAL;
+
+	return 0;
+}
+
+static int basefw_resource_allocation_request(bool first_block,
+					      bool last_block,
+					      uint32_t data_offset_or_size,
+					      const char *data)
+{
+	struct ipc4_resource_request *request;
+
+	if (!(first_block && last_block))
+		return -EINVAL;
+
+	request = (struct ipc4_resource_request *)data;
+
+	switch (request->ra_type) {
+	case IPC4_RAT_DSP_KCPS:
+		return basefw_kcps_allocation_request(&request->ra_data.kcps);
+	case IPC4_RAT_MEMORY:
+		return -EINVAL;
+	default:
+		return -EINVAL;
+	}
+}
+
+static int basefw_power_state_info_get(uint32_t *data_offset, char *data)
+{
+	struct ipc4_tuple *tuple = (struct ipc4_tuple *)data;
+	uint32_t core_kcps[CONFIG_CORE_COUNT] = {0};
+	int core_id;
+
+	set_tuple_uint32(tuple, IPC4_ACTIVE_CORES_MASK, cpu_enabled_cores());
+	tuple = next_tuple(tuple);
+
+	for (core_id = 0; core_id < CONFIG_CORE_COUNT; core_id++) {
+		if (cpu_is_core_enabled(core_id))
+			core_kcps[core_id] = core_kcps_get(core_id);
+	}
+
+	set_tuple(tuple, IPC4_CORE_KCPS, sizeof(core_kcps), core_kcps);
+	tuple = next_tuple(tuple);
+	*data_offset = (int)((char *)tuple - data);
+	return 0;
+}
+
 static int basefw_get_large_config(struct comp_dev *dev,
 				   uint32_t param_id,
 				   bool first_block,
@@ -358,6 +425,9 @@ static int basefw_get_large_config(struct comp_dev *dev,
 		} else {
 			return ret;
 		}
+	break;
+	case IPC4_POWER_STATE_INFO_GET:
+		return basefw_power_state_info_get(data_offset, data);
 	/* TODO: add more support */
 	case IPC4_DSP_RESOURCE_STATE:
 	case IPC4_NOTIFICATION_MASK:
@@ -366,7 +436,6 @@ static int basefw_get_large_config(struct comp_dev *dev,
 	case IPC4_PIPELINE_PROPS_GET:
 	case IPC4_SCHEDULERS_INFO_GET:
 	case IPC4_GATEWAYS_INFO_GET:
-	case IPC4_POWER_STATE_INFO_GET:
 	case IPC4_LIBRARIES_INFO_GET:
 	case IPC4_PERF_MEASUREMENTS_STATE:
 	case IPC4_GLOBAL_PERF_DATA:
@@ -394,6 +463,11 @@ static int basefw_set_large_config(struct comp_dev *dev,
 						last_block, data_offset, data);
 	case IPC4_ENABLE_LOGS:
 		return ipc4_logging_enable_logs(first_block, last_block, data_offset, data);
+	case IPC4_REGISTER_KCPS:
+		return basefw_register_kcps(first_block, last_block, data_offset, data);
+	case IPC4_RESOURCE_ALLOCATION_REQUEST:
+		return basefw_resource_allocation_request(first_block, last_block, data_offset,
+							  data);
 	default:
 		break;
 	}
--- a/src/include/ipc4/base_fw.h
+++ b/src/include/ipc4/base_fw.h
@@ -667,3 +667,8 @@ struct ipc4_log_state_info {
 	 */
 	uint32_t logs_mask[IPC4_MAX_LIBS_COUNT];
 } __attribute__((packed, aligned(4)));
+
+enum ipc4_power_state_type {
+	IPC4_ACTIVE_CORES_MASK = 0,
+	IPC4_CORE_KCPS = 1,
+};
--- a/src/include/sof/lib/cpu-clk-manager.h
+++ b/src/include/sof/lib/cpu-clk-manager.h
@@ -1,0 +1,57 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright(c) 2022 Intel Corporation. All rights reserved.
+ *
+ * Author: Krzysztof Frydryk <frydryk.krzysztof@intel.com>
+ */
+
+/**
+ * \file include/sof/lib/cpu-clk-manager.h
+ * \brief CPU clk manager definition
+ * \author Krzysztof Frydryk <krzysztofx.frydryk@intel.com>
+ */
+
+#ifndef __SOF_LIB_CPU_CLK_MANAGER_H__
+#define __SOF_LIB_CPU_CLK_MANAGER_H__
+
+#include <rtos/spinlock.h>
+#include <sof/sof.h>
+#include <stdint.h>
+
+/**
+ * \brief CPS budget data.
+ */
+struct kcps_budget_data {
+	/* uncache only */
+	int kcps_consumption[CONFIG_CORE_COUNT];	/* Sum of declared consumptions on core */
+	struct k_spinlock lock;
+};
+
+/**
+ * \brief Declare KCPS usage on core
+ *
+ * Declare KCPS (kilo cycles per second) CPU usage on core and adjust
+ * CPU clock according to sum of declared consumptions.
+ *
+ * Declared consumption can be negative in case of freeing CPS.
+ *
+ * @param core The core to which consumption should be pinned
+ * @param kcps_delta declared usage. Can be negative.
+ */
+int core_kcps_adjust(int core, int kcps_delta);
+
+/**
+ * \brief Get KCPS usage on core
+ *
+ * Get declared KCPS usage on core
+ *
+ * @param core The core which consumption should be returned
+ */
+int core_kcps_get(int core);
+
+/**
+ * \brief Init KCPS budget mechanism
+ */
+int kcps_budget_init(void);
+
+#endif /*__SOF_LIB_CPU_CLK_MANAGER_H__ */
--- a/src/lib/cpu-clk-manager.c
+++ b/src/lib/cpu-clk-manager.c
@@ -1,0 +1,95 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// Copyright(c) 2022 Intel Corporation. All rights reserved.
+//
+// Author: Krzysztof Frydryk <frydryk.krzysztof@intel.com>
+
+#include <rtos/spinlock.h>
+#include <sof/sof.h>
+#include <stdint.h>
+#include <sof/lib/cpu-clk-manager.h>
+#include <rtos/clk.h>
+#include <sof/sof.h>
+#include <sof/math/numbers.h>
+#include <errno.h>
+#ifdef __ZEPHYR__
+#include <zephyr/sys/util.h>
+#endif /* __ZEPHYR__ */
+
+static struct kcps_budget_data kcps_data;
+
+static int request_freq_change(unsigned int core, int freq)
+{
+	int current_freq;
+	int selected_freq_id;
+	struct clock_info *clk;
+
+	clk = clocks_get() + core;
+
+	for (selected_freq_id = 0; selected_freq_id < clk->freqs_num; selected_freq_id++) {
+		if (freq < clk->freqs[selected_freq_id].freq)
+			break;
+	}
+
+	/* don't change clock frequency if already using proper clock */
+	current_freq = clock_get_freq(core);
+	if (clk->freqs[selected_freq_id].freq != current_freq)
+		clock_set_freq(core, freq);
+
+	return 0;
+}
+
+static int max_core_consumption(void)
+{
+	int result = 0;
+	unsigned int core;
+
+	for (core = 0; core < CONFIG_CORE_COUNT; core++)
+		result = MAX(result, kcps_data.kcps_consumption[core]);
+
+	return result;
+}
+
+int core_kcps_adjust(int adjusted_core_id, int kcps_delta)
+{
+	k_spinlock_key_t key;
+	int freq;
+	unsigned int core_id;
+	int ret = 0;
+
+	key = k_spin_lock(&kcps_data.lock);
+	kcps_data.kcps_consumption[adjusted_core_id] += kcps_delta;
+
+	/* set clock according to maximum requested mcps budget */
+	freq = max_core_consumption();
+
+	for (core_id = 0; core_id < CONFIG_CORE_COUNT; core_id++) {
+		/* Convert kcps to cps */
+		ret = request_freq_change(core_id, freq * 1000);
+		if (ret < 0)
+			goto out;
+	}
+
+out:
+	k_spin_unlock(&kcps_data.lock, key);
+
+	return ret;
+}
+
+int core_kcps_get(int core)
+{
+	k_spinlock_key_t key;
+	int ret;
+
+	key = k_spin_lock(&kcps_data.lock);
+	ret = kcps_data.kcps_consumption[core];
+	k_spin_unlock(&kcps_data.lock, key);
+	return ret;
+}
+
+int kcps_budget_init(void)
+{
+	k_spinlock_init(&kcps_data.lock);
+
+	return 0;
+}
--- a/src/lib/CMakeLists.txt
+++ b/src/lib/CMakeLists.txt
@@ -23,4 +23,5 @@ add_local_sources(sof
 	dma.c
 	dai.c
 	wait.c
+	cpu-clk-manager.c
 )
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -549,6 +549,7 @@ zephyr_library_sources(
 	${SOF_LIB_PATH}/clk.c
 	${SOF_LIB_PATH}/notifier.c
 	${SOF_LIB_PATH}/pm_runtime.c
+	${SOF_LIB_PATH}/cpu-clk-manager.c
 	${SOF_LIB_PATH}/dma.c
 	${SOF_LIB_PATH}/dai.c
 


src/platform/intel/ace/lib/clk.c
```diff
@@ -12,8 +12,27 @@
...
+#include <zephyr/device.h>
+#include <zephyr/drivers/clock_control.h>
+#endif
+
 static SHARED_DATA struct clock_info platform_clocks_info[NUM_CLOCKS];
 
+
+static int select_cpu_freq(int clock, uint32_t hz)
+{
+#if CONFIG_ZEPHYR_NATIVE_DRIVERS
```
abonislawski, line 25: CONFIG_ZEPHYR_NATIVE_DRIVERS not needed in this file

abonislawski (changes requested):
Why there are no users of this manager? Even not supporting IPC to set CoreCpsAdjust from IPC data?

src/include/sof/lib/cps-budget-manager.h
```diff
@@ -0,0 +1,24 @@
...
+
+#ifndef __SOF_LIB_CPS_BUDGET_MANAGER_H__
+#define __SOF_LIB_CPS_BUDGET_MANAGER_H__
+
+#include <rtos/spinlock.h>
+#include <sof/sof.h>
+#include <stdint.h>
+
+struct cps_budget_data {
+	int kcps_consumtion[CONFIG_CORE_COUNT];
```
lgirdwood, line 16: This is not recommended on multicore as the array[CORE_COUNT] can easily be clobbered. This structure should either use coherent API or be uncache only (with the correct SPARSE types).

src/lib/cps-budget-manager.c
```diff
@@ -0,0 +1,72 @@
...
+
+static int request_freq_change(int core, int freq)
+{
+	int current_freq;
+	int selected_freq_id = 0;
+	struct clock_info *clk;
+
+	clk = clocks_get() + core;
+
+	while (freq > clk->freqs[selected_freq_id].freq)
```
lgirdwood, line 28: Needs to be a for loop with an array size end condition.

src/lib/cps-budget-manager.c
```diff
@@ -0,0 +1,72 @@
...
+#include <stdint.h>
+#include <sof/lib/cps-budget-manager.h>
+#include <rtos/clk.h>
+#include <sof/sof.h>
+#include <sof/math/numbers.h>
+#ifdef __ZEPHYR__
+#include <zephyr/sys/util.h>
+#endif
+
+static struct cps_budget_data cps_data;
```
lgirdwood, line 18: Given my previous statement, its worth saying in a comment that this should be in uncache (which it should be on Zephyr) so no one is tempted to move it to cache.

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -64,3 +64,5 @@ CONFIG_PM_DEVICE_RUNTIME=y
 CONFIG_PM_DEVICE_POWER_DOMAIN=n
 CONFIG_COMP_KPB=y
 
+CONFIG_CLOCK_CONTROL_ADSP=y
+CONFIG_CLOCK_CONTROL=y
```
lgirdwood, line 68: newline

src/platform/intel/ace/lib/clk.c
```diff
@@ -12,8 +12,24 @@
...
+#include <zephyr/drivers/clock_control.h>
+#endif
+
 static SHARED_DATA struct clock_info platform_clocks_info[NUM_CLOCKS];
 
+static int select_cpu_freq(int clock, int hz)
+{
+	const struct device *clock_ctrl = device_get_binding("clkctl");
+
+	if (!clock_ctrl)
```
lgirdwood, line 26: would be nice to have something in the logs here.

abonislawski (commented):
Please add basic usage for this driver/manager in base_fw (set&get kcps data)

btian1 (commented):
whether need a extra function for estimate current workload performance based on pipeline num and component name/num? then consolidate these data and set freq accordingy?

src/audio/base_fw.c
```diff
@@ -323,6 +325,68 @@ static uint32_t basefw_get_ext_system_time(uint32_t *data_offset, char *data)
...
+	power_state_info->active_cores_mask = cpu_enabled_cores();
+	power_state_info->type2 = CORE_KCPS;
+	power_state_info->len2 = CONFIG_CORE_COUNT * sizeof(uint32_t);
+
+	for (int i = 0; i < CONFIG_CORE_COUNT; i++)
+		power_state_info->core_kcps[i] = core_kcps_get(i);
+
+	*data_offset = sizeof(struct ipc4_power_state_info);
+
+	return IPC4_SUCCESS;
```
tmleman, line 387: **struct _ipc4_power_state_info_** contains Type + Length + Value, this looks like TVL, so maybe use tlv.

```c
struct ipc4_tuple *tuple = (struct ipc4_tuple *)data;
uint32_t core_kcps[CONFIG_CORE_COUNT] = {0, 0, 0};

set_tuple_uint32(tuple, ACTIVE_CORES_MASK, cpu_enabled_cores());
tuple = next_tuple(tuple);

for (int core_id = 0; core_id < CONFIG_CORE_COUNT; core_id++) {
    if (arch_cpu_active(core_id))
        core_kcps[core_id] = core_kcps_get(core_id);
}

set_tuple(tuple, CORE_KCPS, sizeof(core_kcps), core_kcps);
tuple = next_tuple(tuple);
*data_offset = (int)((char *)tuple - data);
return IPC4_SUCCESS;
```

tmleman (changes requested):

src/lib/cps-budget-manager.c
kfrydryx: is a comment there sufficient or do we have sth else (sparse?) for uncache specifically?

src/include/sof/lib/cps-budget-manager.h
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+#ifndef __SOF_LIB_CPS_BUDGET_MANAGER_H__
+#define __SOF_LIB_CPS_BUDGET_MANAGER_H__
+
+#include <rtos/spinlock.h>
+#include <sof/sof.h>
+#include <stdint.h>
+
+struct cps_budget_data {
```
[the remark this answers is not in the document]
kfrydryx: cycles per second and kilo cycles per second. cps = 1000 * kcps. Nothing special inside, though I did unify the naming to use kcps where possible.

src/audio/base_fw.c
kfrydryx: thx, didn't notice tuple there being TLV.

kfrydryx:
@btian1 The performance numbers can come in multiple forms. We definitely need some base usage for basefw. For audio modules their usage should come in ipc4 message (the values there are unmeasured yet afaik). Maybe some other parts of fw will have to request budget for themselves. That'll come with future PRs.

src/audio/base_fw.c
```diff
@@ -323,6 +325,69 @@ static uint32_t basefw_get_ext_system_time(uint32_t *data_offset, char *data)
...
+	if (!(first_block && last_block))
+		return IPC4_INVALID_REQUEST;
+
+	if (core_kcps_adjust(0, *(int *)data))
+		return IPC4_FAILURE;
+
+	return IPC4_SUCCESS;
+}
+
+static int basefw_KCPS_allocation_request(struct ipc4_resource_kcps *request)
```
tmleman, line 342: This "\_KCPS\_" just looks weird.

btian1:
> 

Thanks, component performance data will come out after CI team integrate the data, to enable these data, just add -o app/perf_overlay.conf, it need extra computation, please stay tuned, or ping me offline, I will introduce these to CI team tomorrow, and then CI will use script get component/pipeline/core performance data(mcps/memory used)

src/audio/base_fw.c
btian1: agree, this is due to people know mcps more, maybe add a comments is better.

tmleman:
@kfrydryx why don't you fix those failing checks?

kfrydryx:
@tmleman I'm on it. IPC fuzzer and testbench are probably not caused by this PR though (uint32_t is used before stdint include in spinlock.h). Will check in a separate draft.

tmleman:
@kfrydryx I see it differently :-)

```sh
In file included from /src/sof/src/lib/kcps-budget-manager.c:7:
/src/sof/xtos/include/rtos/spinlock.h:17:9: error: unknown type name 'uint32_t'
typedef uint32_t k_spinlock_key_t;
        ^
```

src/lib/kcps-budget-manager.c
```diff
@@ -0,0 +1,87 @@
...
+	int freq;
+
+	key = k_spin_lock(&kcps_data.lock);
+	kcps_data.kcps_consumtion[core] += kcps_delta;
+
+	/* set clock according to maximum requested mcps budget */
+	freq = max_core_consumption();
+
+	for (int i = 0; i < CONFIG_CORE_COUNT; i++)
+		request_freq_change(i, freq);
```
kv2019i: Shouldn't we only modify the frequency of the core this request is for?

src/include/sof/lib/kcps-budget-manager.h
```diff
@@ -0,0 +1,26 @@
...
+
+#include <rtos/spinlock.h>
+#include <sof/sof.h>
+#include <stdint.h>
+
+struct kcps_budget_data {
+	/* uncache only */
+	int kcps_consumtion[CONFIG_CORE_COUNT];
+	struct k_spinlock lock;
+};
```
kv2019i, line 19: I think this is ok with a comment. This has been a problematic approach if there is any need to access the data from multiple cores and performance is a concern (requiring cached access). The "kpcps_consumtion" items are here on the same cacheline, making this impossible to make multicore safe if anyone is accessing this data via a cached pointer.

src/include/sof/lib/kcps-budget-manager.h
```diff
@@ -0,0 +1,26 @@
...
+#ifndef __SOF_LIB_CPS_BUDGET_MANAGER_H__
+#define __SOF_LIB_CPS_BUDGET_MANAGER_H__
+
+#include <rtos/spinlock.h>
+#include <sof/sof.h>
+#include <stdint.h>
+
+struct kcps_budget_data {
+	/* uncache only */
+	int kcps_consumtion[CONFIG_CORE_COUNT];
```
kv2019i, line 17: typo: consumtion -> consumption

src/lib/kcps-budget-manager.c
```diff
@@ -0,0 +1,87 @@
...
+{
+	int result = 0;
+
+	for (int i = 0; i < CONFIG_CORE_COUNT; i++)
+		result = MAX(result, kcps_data.kcps_consumtion[i]);
+
+	return result;
+}
+
+int core_kcps_adjust(int core, int kcps_delta)
```
kv2019i, line 52: What is the expected call model? Is it allowed to call this from any core, or always from core0? I'm guessing as there is locking, the assumption is that any core can call this, right? This would be good to document in the PR.

src/audio/base_fw.c
```diff
@@ -323,6 +325,69 @@ static uint32_t basefw_get_ext_system_time(uint32_t *data_offset, char *data)
...
 
+static int basefw_register_kcps(bool first_block,
+				bool last_block,
+				uint32_t data_offset_or_size,
+				const char *data)
+{
+	if (!(first_block && last_block))
+		return IPC4_INVALID_REQUEST;
+
+	if (core_kcps_adjust(0, *(int *)data))
```
kv2019i, line 336: This would need some inline documentation. First the cast to "int" looks not right, how can we assume the "int" type matches the data received via IPC. Should this be a fixed-width type like "int32_t"?

Second, why this is adding core 0 here, is this implicit part of the IPC4 protocol?

src/include/sof/lib/kcps-budget-manager.h
```diff
@@ -0,0 +1,26 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright(c) 2022 Intel Corporation. All rights reserved.
+ *
+ * Author: Krzysztof Frydryk <frydryk.krzysztof@intel.com>
+ */
+
+#ifndef __SOF_LIB_CPS_BUDGET_MANAGER_H__
+#define __SOF_LIB_CPS_BUDGET_MANAGER_H__
```
kv2019i, line 9: The name is a bit challenging, but I can't come up with a better one. "DSP Cycle Budget Manager" would be more explicit, but "CPS" is used in the IPC messages, so maybe this is good enough.

kv2019i (commented):
PR updated while I was revieing, so posting my older comments here (not all may apply).

src/include/sof/lib/kcps-budget-manager.h
kfrydryx: Performance shouldn't be a concern there. It's not a lot of data and should be used only when stuff is initializing or deinitializing.

src/lib/kcps-budget-manager.c
kfrydryx: This might not be generally true, but currently on all supported platforms the `core` information is ignored when actually setting cpu clock, meaning there is one selection for all cores. So I suppose we'd like to inform all cores (`clock_set_freq` sends notifications about clock change). Alternatively I can make this behavior platform specific.

lgirdwood:
@kfrydryx btw, is there any CI test that uses code in this PR or is this currently locally tested ? i.e will this code be invoked today if CI runs a low MCPS pipeline ?

src/audio/base_fw.c
```diff
@@ -394,6 +463,10 @@ static int basefw_set_large_config(struct comp_dev *dev,
 						last_block, data_offset, data);
 	case IPC4_ENABLE_LOGS:
 		return ipc4_logging_enable_logs(first_block, last_block, data_offset, data);
+	case IPC4_REGISTER_KCPS:
```
pblaszko, line 466: @kfrydryx did you try to clarify with ACE FW/Windows SW why do we have two IPC's doing actually the same?

pblaszko (dismissed):

src/audio/base_fw.c
```diff
@@ -394,6 +463,10 @@ static int basefw_set_large_config(struct comp_dev *dev,
...
+		return basefw_register_kcps(first_block, last_block, data_offset, data);
+	case IPC4_RESOURCE_ALLOCATION_REQUEST:
```
pblaszko, line 468: @kfrydryx did you try to clarify with ACE FW/Windows SW why do we have two IPC's doing actually the same?

src/lib/kcps-budget-manager.c
```diff
@@ -0,0 +1,87 @@
...
+
+	for (selected_freq_id = 0; selected_freq_id < clk->freqs_num; selected_freq_id++) {
+		if (freq < clk->freqs[selected_freq_id].freq)
+			break;
+	}
+
+	/* don't change clock frequency if already using proper clock */
+	current_freq = clock_get_freq(core);
+	if (clk->freqs[selected_freq_id].freq != current_freq)
+		clock_set_freq(core, freq);
```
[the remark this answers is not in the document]
kfrydryx: It's a void function `void clock_set_freq(int clock, uint32_t hz)`

src/lib/kcps-budget-manager.c
```diff
@@ -0,0 +1,87 @@
...
+	int core_id;
+
+	key = k_spin_lock(&kcps_data.lock);
+	kcps_data.kcps_consumption[core] += kcps_delta;
+
+	/* set clock according to maximum requested mcps budget */
+	freq = max_core_consumption();
+
+	for (core_id = 0; core_id < CONFIG_CORE_COUNT; core_id++)
+		request_freq_change(core_id, freq);
```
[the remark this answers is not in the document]
kfrydryx: ok

src/audio/base_fw.c
kfrydryx: I did not. register_kcps is not used in internal testing, though the call is implemented. Both are handled by closed source fw.

kfrydryx:
@lgirdwood This code is not used yet anywhere else, other than in those IPC requests. There are tests in internal CI calling them (though they are not made exclusively to test this functionality).

src/audio/base_fw.c
btian1: is 1 mcps == 1000 kcps？

kfrydryx: yes, exactly. Just like in metric system. 1 MCPS = 1000 KCPS = 1000 000 CPS.

src/lib/cps-budget-manager.c
lyakh: @lgirdwood don't think so. An address space has to be specified explicitly for sparse to check it, all objects, that don't specify one explicitly are assumed to belong to the default address space - regardless whether static or automatic.

src/lib/kcps-budget-manager.c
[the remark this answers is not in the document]
pblaszko: yes, my fault, sorry

src/include/ipc4/base_fw.h
```diff
@@ -667,3 +667,8 @@ struct ipc4_log_state_info {
 	 */
 	uint32_t logs_mask[IPC4_MAX_LIBS_COUNT];
 } __attribute__((packed, aligned(4)));
+
+enum ipc4_power_state_type {
+	ACTIVE_CORES_MASK = 0,
```
lgirdwood, line 672: can we prefix to avoid any naming collisions/confusion ?

src/include/sof/lib/kcps-budget-manager.h
```diff
@@ -0,0 +1,26 @@
...
+#include <sof/sof.h>
+#include <stdint.h>
+
+struct kcps_budget_data {
+	/* uncache only */
+	int kcps_consumption[CONFIG_CORE_COUNT];
+	struct k_spinlock lock;
+};
+
+int core_kcps_adjust(int core, int kcps_delta);
```
lgirdwood, line 21: We need to add doxygen for these new APIs.

lgirdwood (commented):
@kfrydryx I assume this should be our single way/API to control CPU frequency/source (when it's enabled to avoid collisions with the manual API) ?  If so, I would recommend 
1) We check current CPU frequency/source is correct before we set new frequency/source.
2) I would call it cpu-clk-manager.c/.h, the word "budget" suggest scheduling and this PR is about clock frequency/source.

src/include/ipc4/base_fw.h
kfrydryx: Sure
Also:
File name changed.
I was thinking about having a pause/resume mechanism for overriding the managers clock choice, but according to @mmaka1 it is currently sufficient. This can be easily worked around by requesting artificial kcps budget. We were also considering an approach of holding a reference of who requested the clock and decide on the change on clock driver level, but left that topic for later.

src/include/sof/lib/cpu-clk-manager.h
```diff
@@ -0,0 +1,57 @@
...
+ * Author: Krzysztof Frydryk <frydryk.krzysztof@intel.com>
+ */
+
+/**
+  * \file include/sof/lib/cpu-clk-manager.h
+  * \brief CPU clk manager definition
+  * \author Krzysztof Frydryk <krzysztofx.frydryk@intel.com>
+  */
+
+#ifndef __SOF_LIB_CPS_BUDGET_MANAGER_H__
```
mmaka1, line 14: Pls update the pre-proc symbols to reflect the new (nice) name.

src/lib/cpu-clk-manager.c
```diff
@@ -0,0 +1,95 @@
...
+		result = MAX(result, kcps_data.kcps_consumption[core]);
+
+	return result;
+}
+
+int core_kcps_adjust(int core, int kcps_delta)
+{
+	k_spinlock_key_t key;
+	int freq;
+	int core_id;
```
mmaka1, line 57: Using `core` and `core_id` as symbols is a bit confusing, especially that the first variable contains a core id too. `core` could be renamed as `adjusted_core_id` or sth similar.

src/lib/cpu-clk-manager.c
```diff
@@ -0,0 +1,95 @@
...
+	int result = 0;
+	int core;
+
+	for (core = 0; core < CONFIG_CORE_COUNT; core++)
+		result = MAX(result, kcps_data.kcps_consumption[core]);
+
+	return result;
+}
+
+int core_kcps_adjust(int core, int kcps_delta)
```
mmaka1, line 53: Not sure why `core` is of signed int type (also in many other places in the code) while this piece of the code (as many others) works only for sequential values in 0..CONFIG_CORE_COUNT-1 range. Note that `cpu_get_id()` is now mapped to Zephyr's `arch_proc_id()` that returns `uint32_t`. Note that the latter does not guarantee the condition mentioned above (see https://docs.zephyrproject.org/latest/hardware/porting/arch.html#c.arch_proc_id), so indexing `kcps_consumption` with `core` might not be safe, unless there is some integration note but I did not find anything in place the `cpu_get_id()` is declared/defined.

src/lib/cpu-clk-manager.c
```diff
@@ -0,0 +1,95 @@
...
+	kcps_data.kcps_consumption[core] += kcps_delta;
+
+	/* set clock according to maximum requested mcps budget */
+	freq = max_core_consumption();
+
+	for (core_id = 0; core_id < CONFIG_CORE_COUNT; core_id++) {
+		/* Convert kcps to cps */
+		ret = request_freq_change(core_id, freq * 1000);
+		if (ret < 0) {
+			k_spin_unlock(&kcps_data.lock, key);
```
mmaka1, line 70: Defining multiple exit points and requiring serious cleanup like `unlock` in all that places might be less readable that setting error code to the local variable and jump to the label near the end of hte function.

src/include/sof/lib/cpu-clk-manager.h
kfrydryx: Sorry for the late answer, sure. Missed that.

src/lib/cpu-clk-manager.c
kfrydryx: agree, changed

src/lib/cpu-clk-manager.c
kfrydryx: ok

src/lib/cpu-clk-manager.c
kfrydryx: That's the way it's done in other parts of code. I also couldn't find any spec guaranteeing core ids to be continues and starting at 0. I can imagine something like platform specific iterator for enumerating cores, but that is probably not the place for such discussion. And it would require refactoring other code too.
I did change the iterators to be unsigned, just for the sake of it :)

src/audio/base_fw.c
```diff
@@ -323,6 +325,71 @@ static uint32_t basefw_get_ext_system_time(uint32_t *data_offset, char *data)
...
+				const char *data)
+{
+	if (!(first_block && last_block))
+		return IPC4_INVALID_REQUEST;
+
+	/* value of kcps to request on core 0. Can be negative */
+	if (core_kcps_adjust(0, *(int32_t *)data))
+		return IPC4_FAILURE;
+
+	return IPC4_SUCCESS;
```
lgirdwood, line 340: This needs to be POSIX and do the IPC4 translation in the IPC layer.

btian1:
Is there a way to adjust DSP clock at runtime instead of at beginning? this is because beginning performance is poor at the first 1-3 ms running.

src/audio/base_fw.c
kfrydryx: I see, in base_fw.c they are mixed up. I suppose that's just lack of cleanup?

kfrydryx:
@btian1 Well... that's what this is for. The desired model is to register everything that is running on the fw and change clock accordingly at runtime. At the moment there's not enough performance data for this to make sense, so that PR consists only of base mechanism and some IPC handling as usage.
Also, at least for MTL, I believe the clock is hardcode selected to fastest at boot.

btian1:
> 

ok，we will have mcps data later, Q1 will happen, for one single playback, expectation is change clock one time except there are new pipe join, since change clock also takes time.

Also I would echo Liam's suggestion, title and file name better come up with: cpu-clk-manager.c/.h or adsp-clk-manager.c/h

lgirdwood (approved):

btian1 (approved):

tmleman (approved):

lgirdwood:
@pblaszko good for you now ?

abonislawski (approved):

mwasko (approved):

PR #6651, opened 2022-11-22
ipc4: Fix chain DMA trigger handling

jsarha:
We need to pass the pipeline_trigger() return value through in ipc4_trigger_chain_dma() for the possibly delayed reply message sending to work correctly.

Signed-off-by: Jyri Sarha <jyri.sarha@intel.com>

### DIFF
--- a/src/ipc/ipc4/helper.c
+++ b/src/ipc/ipc4/helper.c
@@ -722,12 +722,13 @@ int ipc4_create_chain_dma(struct ipc *ipc, struct ipc4_chain_dma *cdma)
 	return ret;
 }
 
-int ipc4_trigger_chain_dma(struct ipc *ipc, struct ipc4_chain_dma *cdma)
+int ipc4_trigger_chain_dma(struct ipc *ipc, struct ipc4_chain_dma *cdma,
+			   bool *delay)
 {
 	struct ipc_comp_dev *ipc_pipe;
 	struct comp_dev *host;
 	uint32_t pipeline_id;
-	int ret;
+	int ret = 0;
 
 	pipeline_id = IPC4_COMP_ID(cdma->primary.r.host_dma_id + IPC4_MAX_MODULE_COUNT,
 				   cdma->primary.r.link_dma_id);
@@ -746,6 +747,10 @@ int ipc4_trigger_chain_dma(struct ipc *ipc, struct ipc4_chain_dma *cdma)
 			if (ret < 0) {
 				tr_err(&ipc_tr, "failed to disable chain dma %d", ret);
 				return IPC4_BAD_STATE;
+			} else if (ret == PPL_STATUS_SCHEDULED) {
+				/* pipeline reset will be done by schedule thread */
+				*delay = true;
+				return IPC4_SUCCESS;
 			}
 		}
 
@@ -788,6 +793,9 @@ int ipc4_trigger_chain_dma(struct ipc *ipc, struct ipc4_chain_dma *cdma)
 			return IPC4_BAD_STATE;
 		}
 	}
+
+	if (ret == PPL_STATUS_SCHEDULED)
+		*delay = true;
 
 	return IPC4_SUCCESS;
 }
--- a/src/include/sof/ipc/topology.h
+++ b/src/include/sof/ipc/topology.h
@@ -51,7 +51,7 @@ struct comp_dev *ipc4_get_comp_dev(uint32_t comp_id);
 int ipc4_add_comp_dev(struct comp_dev *dev);
 const struct comp_driver *ipc4_get_drv(uint8_t *uuid);
 int ipc4_create_chain_dma(struct ipc *ipc, struct ipc4_chain_dma *cdma);
-int ipc4_trigger_chain_dma(struct ipc *ipc, struct ipc4_chain_dma *cdma);
+int ipc4_trigger_chain_dma(struct ipc *ipc, struct ipc4_chain_dma *cdma, bool *delay);
 #else
 #error "No or invalid IPC MAJOR version selected."
 #endif
--- a/src/ipc/ipc4/handler.c
+++ b/src/ipc/ipc4/handler.c
@@ -454,6 +454,7 @@ static int ipc4_process_chain_dma(struct ipc4_message_request *ipc4)
 {
 	struct ipc4_chain_dma cdma;
 	struct ipc *ipc = ipc_get();
+	bool delay = false;
 	int ret;
 
 	memcpy_s(&cdma, sizeof(cdma), ipc4, sizeof(cdma));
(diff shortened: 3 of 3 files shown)


jsarha:
With this PR together with https://github.com/thesofproject/linux/pull/3953 sof-test/test-case/multiple-pause-resume.sh test passes also with SOF FW. However, I am still not sure if !cdma->primary.r.enable && !cdma->primary.r.allocate message is handled correctly if the stream is active. Please check @lyakh and @RanderWang !

RanderWang:
> 

https://github.com/thesofproject/sof/commit/06226d00b034921fdb831d7de9c4b240c485248c

> With this PR together with [thesofproject/linux#3953](https://github.com/thesofproject/linux/pull/3953) sof-test/test-case/multiple-pause-resume.sh test passes also with SOF FW. However, I am still not sure if !cdma->primary.r.enable && !cdma->primary.r.allocate message is handled correctly if the stream is active. Please check @lyakh and @RanderWang !

@jsarha please check this commit https://github.com/thesofproject/sof/commit/06226d00b034921fdb831d7de9c4b240c485248c. We need to first pause and then check "ret != PPL_STATUS_SCHEDULED" to do reset.

ujfalusi (approved):
@jsarha, good catch!

abonislawski (approved):

jsarha:
> 
> @jsarha please check this commit [06226d0](https://github.com/thesofproject/sof/commit/06226d00b034921fdb831d7de9c4b240c485248c). We need to first pause and then check "ret != PPL_STATUS_SCHEDULED" to do reset.

Thanks @RanderWang ,  but I still do not see how ipc_pipeline_free() would get called. Or will it be cleaned up automatically by some later step?

src/ipc/ipc4/helper.c
```diff
@@ -764,7 +764,7 @@ int ipc4_trigger_chain_dma(struct ipc *ipc, struct ipc4_chain_dma *cdma)
 			}
 		}
 
-		return IPC4_SUCCESS;
+		return ret;
```
lyakh, line 767: sorry, don't understand, I don't see any path where a non-success return code would be lost? It looks like everywhere `ret` is in the "negative POSIX or 0" error space, and this seems to mix it with IPC4_* error codes. Yet another one...

lyakh (changes requested):

lyakh:
I think the calling of `ipc4_trigger_chain_dma()` is wrong:
```
	ret = ipc4_trigger_chain_dma(ipc, &cdma);
	if (ret != PPL_STATUS_SCHEDULED) {
```
`ipc4_trigger_chain_dma()` returns IPC4 error codes, so, numerically `PPL_STATUS_SCHEDULED` is equal to 2 and is equal to `IPC4_UNKNOWN_MESSAGE_TYPE` which is probably not what we want to check

jsarha:
> I think the calling of `ipc4_trigger_chain_dma()` is wrong:
> 
> ```
> 	ret = ipc4_trigger_chain_dma(ipc, &cdma);
> 	if (ret != PPL_STATUS_SCHEDULED) {
> ```
> 
> `ipc4_trigger_chain_dma()` returns IPC4 error codes, so, numerically `PPL_STATUS_SCHEDULED` is equal to 2 and is equal to `IPC4_UNKNOWN_MESSAGE_TYPE` which is probably not what we want to check

@lyakh AFAIU trigger handling requires delayed reply handling. At least pipeline_trigger() may return PPL_STATUS_SCHEDULED, shouldn't we handle that properly?

RanderWang:
> > I think the calling of `ipc4_trigger_chain_dma()` is wrong:
> > ```
> > 	ret = ipc4_trigger_chain_dma(ipc, &cdma);
> > 	if (ret != PPL_STATUS_SCHEDULED) {
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `ipc4_trigger_chain_dma()` returns IPC4 error codes, so, numerically `PPL_STATUS_SCHEDULED` is equal to 2 and is equal to `IPC4_UNKNOWN_MESSAGE_TYPE` which is probably not what we want to check
> 
> @lyakh AFAIU trigger handling requires delayed reply handling. At least pipeline_trigger() may return PPL_STATUS_SCHEDULED, shouldn't we handle that properly?

@jsarha  I understood why IPC4_SUCCESS was returned. lyakh is correct, we can't return PPL_STATUS_SCHEDULED directly. You can add a function params like "bool *delay" for ipc4_trigger_chain_dma, then return the status in delay

RanderWang:
> > @jsarha please check this commit [06226d0](https://github.com/thesofproject/sof/commit/06226d00b034921fdb831d7de9c4b240c485248c). We need to first pause and then check "ret != PPL_STATUS_SCHEDULED" to do reset.
> 
> Thanks @RanderWang , but I still do not see how ipc_pipeline_free() would get called. Or will it be cleaned up automatically by some later step?

@jsarha  I got your idea. For normal pipeline there will be a pipeline free ipc msg, but none for chain dma.
````c
	ret = ipc4_trigger_chain_dma(ipc, &cdma);

	/* it is not scheduled in another thread */
	if (ret != PPL_STATUS_SCHEDULED) {
		msg_data.delayed_reply = 0;
		msg_data.delayed_error = 0;
	} else if (!cdma->primary.r.allocate) {
               ipc_wait_for_compound_msg();
		ret = ipc_pipeline_free(ipc, pipeline_id);
                if (ret < 0) {
				tr_err(&ipc_tr, "failed to free chain dma %d", ret);
				ret = IPC4_BAD_STATE;
		} else
                                 ret = IPC4_SUCCESS;
	}
````

File: src/ipc/ipc4/helper.c [hunk 2]
```diff
@@ -746,6 +747,9 @@ int ipc4_trigger_chain_dma(struct ipc *ipc, struct ipc4_chain_dma *cdma)
 			if (ret < 0) {
 				tr_err(&ipc_tr, "failed to disable chain dma %d", ret);
 				return IPC4_BAD_STATE;
+			} else if (ret == PPL_STATUS_SCHEDULED) {
+				*delay = true;
+				return IPC4_SUCCESS;
```
RanderWang, line 752: add comments here: pipeline reset will be done by schedule thread

src/ipc/ipc4/helper.c
```diff
@@ -746,6 +747,9 @@ int ipc4_trigger_chain_dma(struct ipc *ipc, struct ipc4_chain_dma *cdma)
 			if (ret < 0) {
 				tr_err(&ipc_tr, "failed to disable chain dma %d", ret);
 				return IPC4_BAD_STATE;
+			} else if (ret == PPL_STATUS_SCHEDULED) {
```
lyakh, line 750: no `else` needed

lyakh (approved):

jsarha:
> @jsarha I got your idea. For normal pipeline there will be a pipeline free ipc msg, but none for chain dma.
> 
> ```c
> 	ret = ipc4_trigger_chain_dma(ipc, &cdma, &delay);
>         if (ret != IPC4_SUCCUESS)
>                return ret;
> 
> 	/* it is not scheduled in another thread */
> 	if (!delay) {
> 		msg_data.delayed_reply = 0;
> 		msg_data.delayed_error = 0;
> 	} else if (!cdma->primary.r.allocate) {
>               /* waiting for pipeline reset done */
>                ipc_wait_for_compound_msg();
> 		ret = ipc_pipeline_free(ipc, pipeline_id);
>                 if (ret < 0) {
> 				tr_err(&ipc_tr, "failed to free chain dma %d", ret);
> 				ret = IPC4_BAD_STATE;
> 		} else
>                                  ret = IPC4_SUCCESS;
> 	}
> ```

This should never be true: (ret != IPC4_SUCCESS && delay), so there is really no need to check right after ipc4_trigger_chain_dma() call, and even in an error case we should reset msg_data.delayed_reply and delayed_error, right?

jsarha:
> yes, no harmful to reset msg_data. Checking the ret is to make it easy to understand.

@RanderWang but since we have set msg_data.delayed_reply to 1 isn't clearing it something we have to do, also in error case?

RanderWang (approved):
LGTM

jsarha:
Latest version just adds one newline between variable declaration and code.

kv2019i:
Approvals in place, tests pass, no more comments in last two days, proceeding with merge.

PR #6654, opened 2022-11-23
[SKIP CI] Added daily tests workflow

aborisovich:
Daily tests workflow triggers "Main Actions" workflow from ./.github/workflows/pull-requests.yml on the main branch
in the time interval defined in schedule/cron node.
Cron value may be set to run in some desired time intervals.
Cron values online configurator: https://crontab.guru

Github references:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule
https://docs.github.com/en/actions/using-workflows/reusing-workflows

### DIFF
--- a/.github/workflows/daily-tests.yml
+++ b/.github/workflows/daily-tests.yml
@@ -1,0 +1,20 @@
+---
+# https://docs.github.com/en/actions/using-workflows/reusing-workflows
+# https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule
+
+name: Daily tests
+
+# yamllint disable-line rule:truthy
+on:
+  # To configure goto https://crontab.guru/
+  schedule:
+    # daily at 00:00 UTC
+    - cron: '0 0 * * *'
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  run-all-main-actions:
+    uses: ./.github/workflows/pull-request.yml
+  run-zephyr-builds:
+    uses: ./.github/workflows/zephyr.yml
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -28,6 +28,8 @@ on:
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
+  # Allows to call this forkflow from other workflows
+  workflow_call:
 
 jobs:
 
--- a/.github/workflows/zephyr.yml
+++ b/.github/workflows/zephyr.yml
@@ -5,7 +5,7 @@ name: Zephyr
 # 'workflow_dispatch' allows running this workflow manually from the
 # 'Actions' tab
 # yamllint disable-line rule:truthy
-on: [push, pull_request, workflow_dispatch]
+on: [push, pull_request, workflow_dispatch, workflow_call]
 
 jobs:
   build:


.github/workflows/daily-tests.yml
```diff
@@ -0,0 +1,15 @@
+# https://docs.github.com/en/actions/using-workflows/reusing-workflows
```
lgirdwood, line 1: can you update the commit message subject and add some context. Thanks.

aborisovich:
It looks like this workflow will not run until is merged to main branch.
I'll remove on: shedule node leaving only workflow_dispatch node to run action manually from actions tab and verify it is working correctly.

aborisovich:
Looks better now, "Main Actions" and "Zephyr" workflows had to be added new node:
```
on:
  workflow_call
```
so they can be called by Daily Tests workflow

aborisovich:
I'll remove "on pull-request trigger" and leave on: schedule to check whether this will be executed in time intervals on main branch.

aborisovich:
Still no sign of the "Daily Tests" action in Action tab...
But according to Github docs it may be delayed due to high traffic of other Actions currently running...
Let's wait a bit.
image

aborisovich:
It looks like sheduling will not work until this workflow is on the master branch (PR is merged).
I'll add "workflow_dispatch" to Daily Tests to check if it is working.

aborisovich:
@marc-hb is it possible that new workflow is not available in Github Actions tab despite having "workflow_dispatch" trigger?
I can add "on pull request" trigger as a test to check whether all sub-workflows are starting and then remove it when we merge.

aborisovich:
All checks starting with "Daily tests /" are new workflow.
Awaiting reviews from @lgirdwood and @marc-hb .

I've consulted today Intel Internal CI team about how to trigger those workflows from Daily tests:
image
received response that this can be done but not now as it requires manual work with Github REST API.
Currently we have higher priority tasks related to internal CI.

Looks like new workflow finally appeared in the Actions tab:
image

Basically, with this new workflow we can at least daily check whether all platforms are building and whether build with Zephyr main HEAD passes.
It would be great if sof-ci checks:
image
could also be triggered by new Daily tests but I don't know how much work is required on the backend side on sof-ci.
However, it does listen for new PRs so I guess it can also listen for new action to appear?

aborisovich:
I think we can ignore codestyle check...
I've never seen anyone starting Action document with "---" characters on top.
Also, this warning related to "truthy" values seems wrong (there is "on" statement in line 6).

lgirdwood (approved):
This should speed up the PR CI.

marc-hb (changes requested):
Surprisingly short and simple, thanks! Only need a very easy fix for the two minor yamllint warnings as far as I'm concerned.

I wonder where the results will appear?

> It would be great if sof-ci/jenkins checks could also be triggered by new Daily tests but I don't know how much work is required on the backend side on sof-ci.


I don't know either, @greg-intel, @keqiaozhang , @fredoh9 , @aiChaoSong?

Note sof-ci/jenkins already runs some more extensive, daily sof-test plans internally but the results can unfortunately not be made public as a whole.

> However, it does listen for new PRs so I guess it can also listen for new action to appear?

AFAIK sof-ci/jenkins is continuously polling Github, looking for recent PR changes.

> I think we can ignore codestyle check...

You forgot the link, I'm guessing you're referring to https://github.com/thesofproject/sof/actions/runs/3532835426/jobs/5927687763
Always link to test results when discussing them as they're very hard/impossible to find again after a force-push.

We cannot ignore these minor yamllint warnings because this would make this check fail every time on every force push of every PR in the future. These two warnings are super easy to silence, just look at other files in the same directory.

marc-hb:
> I wonder where the results will appear?

Wait, why is this drowning the results of this Pull Request in dozens of new "Daily tests" line in? This does not look right, this pull request is not a daily test...

marc-hb:
> I wonder where the results will appear?

Here I guess: https://github.com/thesofproject/sof/actions

aborisovich:
Fixed checkpatch/yamling warnings.

marc-hb (approved):

aborisovich:
Ok, removed test code (`on: pull_request`), ready to go.
It's 23:44 of UTC time so if you merge quickly we might observe triggering this action by Github scheduler at 00:00 ;-)

marcinszkudlinski:
please rebase on newest main and re-run checks with newest sof/zephyr

marc-hb:
Looking great now after 2 days, thank you @aborisovich : https://github.com/thesofproject/sof/actions/workflows/daily-tests.yml

PR #6656, opened 2022-11-23
hda: hw support for hda chain

makarukp:
This PR enables hardware support for HDA chain transfers (without deploying internal pipelines):
Content:
- hda chain management
- hda chain tasks and DMA logic implementation


src/audio/pipeline/chain_dma.c
```diff
@@ -0,0 +1,95 @@
...
+static const struct comp_driver comp_chain_dma = {
+		.uid = SOF_RT_UUID(chain_dma_uuid),
+		.tctx = &chain_dma_tr,
+		.ops = {
+				.create = chain_dma_mng_create,
+				.chain_dma_create = chain_dma_create,
+				.chain_dma_trigger = chain_dma_trigger,
+				.chain_dma_remove = chain_dma_remove,
+		},
+};
```
lyakh, line 83: this is a rare thing to say, but this looks like too much indentation

src/audio/pipeline/chain_dma.c
```diff
@@ -0,0 +1,95 @@
...
+
+static SHARED_DATA struct comp_driver_info comp_chain_dma_info = {
+		.drv = &comp_chain_dma,
```
lyakh, line 86: ditto

src/audio/pipeline/chain_dma.c
```diff
@@ -0,0 +1,95 @@
...
+};
+
+UT_STATIC void sys_comp_chain_dma_init(void)
+{
+		comp_register(platform_shared_get(&comp_chain_dma_info,
```
lyakh, line 91: here as well

src/include/sof/audio/chain_dma.h
```diff
@@ -0,0 +1,41 @@
...
+#include <sof/common.h>
+#include <sof/ipc/topology.h>
+#include <sof/ipc/common.h>
+#include <ipc/dai.h>
+
+static const uint32_t MAX_CHAIN_NUMBER = DAI_NUM_HDA_OUT + DAI_NUM_HDA_IN;
+
+/* chain dma component private data */
+struct chain_dma_data {
+	bool first_data_received_;
```
lyakh, line 14: why are you using underscore at the end? Don't think this is a common practice in SOF

src/include/sof/audio/chain_dma.h
```diff
@@ -0,0 +1,41 @@
...
+	//! Node id of host HD/A DMA
```
lyakh, line 15: comment style

src/include/sof/audio/chain_dma.h
```diff
@@ -0,0 +1,41 @@
...
+					uint32_t fifo_size, bool scs);
+
+int chain_dma_start(struct comp_dev *dev, uint8_t host_dma_id);
+
+int chain_dma_pause(struct comp_dev *dev, uint8_t host_dma_id);
+
+int chain_dma_remove(struct comp_dev *dev, uint8_t host_dma_id);
+
+int chain_dma_trigger(struct comp_dev *dev,
+					uint8_t host_dma_id, int cmd);
```
lyakh, line 37: indentation after the opening parenthesis everywhere please

src/include/sof/audio/chain_dma.h
```diff
@@ -0,0 +1,41 @@
...
+
+void SetReadPointer(uint32_t read_pointer);
```
lyakh, line 39: camel case

src/include/sof/audio/component.h
```diff
@@ -307,6 +307,16 @@ struct comp_ops {
...
 
+	int (*chain_dma_create)(struct comp_dev *dev,
+					uint8_t host_dma_id, uint8_t link_dma_id,
+					uint32_t fifo_size, bool scs);
+
+	int (*chain_dma_remove)(struct comp_dev *dev,
+					uint8_t host_dma_id);
+
+	int (*chain_dma_trigger)(struct comp_dev *dev,
+					uint8_t host_dma_id, int cmd);
```
lyakh, line 318: can a component have both - "normal" and chain DMA operations? If not, would it be possible to reuse `.create()`, `.free()`, and `.trigger()`? I see that these function have an additional host_dma_id parameter, but maybe it can be obtained from the device?

src/ipc/ipc4/helper.c
```diff
@@ -589,6 +590,92 @@ static int construct_config(struct ipc4_copier_module_cfg *copier_cfg, uint32_t
...
+{
+	const struct sof_uuid uuid = {0x6a0a274f, 0x27cc, 0x4afb, {0xa3, 0xe7, 0x34,
+			0x44, 0x72, 0x3f, 0x43, 0x2e}};
+
+	const struct comp_driver *drv;
+	struct comp_dev *dev;
+	int ret = IPC4_INVALID_REQUEST;
+
+	drv = ipc4_get_drv((uint8_t *)&uuid);
+	 if (!drv)
```
lyakh, line 603: indentation

src/ipc/ipc4/helper.c
```diff
@@ -589,6 +590,92 @@ static int construct_config(struct ipc4_copier_module_cfg *copier_cfg, uint32_t
...
+	if (!dev)
+		return IPC4_INVALID_RESOURCE_ID;
+
+	const uint32_t comp_id = IPC4_COMP_ID(IPC4_MAX_MODULE_COUNT, 0);
+	dev->ipc_config.id = comp_id;
+
+	list_init(&dev->bsource_list);
+	list_init(&dev->bsink_list);
+
+	ret = ipc4_add_comp_dev(dev);
```
lyakh, line 618: `return ipc4_add_comp_dev(dev);` and remove `ret`

src/ipc/ipc4/helper.c
```diff
@@ -589,6 +590,92 @@ static int construct_config(struct ipc4_copier_module_cfg *copier_cfg, uint32_t
...
 }
 
+int ipc4_create_mng_for_chain()
+{
+	const struct sof_uuid uuid = {0x6a0a274f, 0x27cc, 0x4afb, {0xa3, 0xe7, 0x34,
+			0x44, 0x72, 0x3f, 0x43, 0x2e}};
+
+	const struct comp_driver *drv;
+	struct comp_dev *dev;
+	int ret = IPC4_INVALID_REQUEST;
```
lyakh, line 600: superfluous initialisation (and the variable too)

src/ipc/ipc4/helper.c
```diff
@@ -589,6 +590,92 @@ static int construct_config(struct ipc4_copier_module_cfg *copier_cfg, uint32_t
...
+
+int ipc4_chain_dma_state(struct comp_dev *dev, struct ipc4_chain_dma *cdma)
+{
+	const uint8_t link_dma_id = cdma->primary.r.link_dma_id;
+	const uint8_t host_dma_id = cdma->primary.r.host_dma_id;
+	const uint32_t fifo_size = cdma->extension.r.fifo_size;
+	const bool allocate = cdma->primary.r.allocate;
+	const bool enable = cdma->primary.r.enable;
+	const bool scs = cdma->primary.r.scs;
+	int ret = IPC4_INVALID_REQUEST;
```
lyakh, line 631: superfluous init

src/ipc/ipc4/helper.c
```diff
@@ -589,6 +590,92 @@ static int construct_config(struct ipc4_copier_module_cfg *copier_cfg, uint32_t
 	return IPC4_SUCCESS;
 }
 
+int ipc4_create_mng_for_chain()
```
lyakh, line 593: `int ipc4_chain_dma_create_mng(void)`

src/ipc/ipc4/helper.c
```diff
@@ -589,6 +590,92 @@ static int construct_config(struct ipc4_copier_module_cfg *copier_cfg, uint32_t
...
+			return ret;
+		}
+		if (enable)
+		{
+			ret = drv->ops.chain_dma_trigger(dev, host_dma_id, COMP_TRIGGER_START);
+		}
+		else
+		{
+			ret = drv->ops.chain_dma_trigger(dev, host_dma_id, COMP_TRIGGER_PAUSE);
+		}
```
lyakh, line 659: please remove all the superfluous braces

src/ipc/ipc4/helper.c
```diff
@@ -589,6 +590,92 @@ static int construct_config(struct ipc4_copier_module_cfg *copier_cfg, uint32_t
...
+			return IPC4_INVALID_REQUEST;
+		}
+		else
+		{
+			ret = drv->ops.chain_dma_trigger(dev, host_dma_id, COMP_TRIGGER_PAUSE);
+		}
+		if (ret == IPC4_SUCCESS)
+		{
+			drv->ops.chain_dma_remove(dev, host_chan);
+		}
```
lyakh, line 674: braces here too

src/ipc/ipc4/helper.c
```diff
@@ -589,6 +590,92 @@ static int construct_config(struct ipc4_copier_module_cfg *copier_cfg, uint32_t
...
+			ret = drv->ops.chain_dma_trigger(dev, host_dma_id, COMP_TRIGGER_PAUSE);
+		}
+	}
+	else
+	{
+		if (enable)
+		{
+			return IPC4_INVALID_REQUEST;
+		}
+		else
```
lyakh, line 667: `else` unneeded

src/audio/pipeline/chain_dma.c
```diff
@@ -0,0 +1,95 @@
...
+
+	struct chain_dma_data *cd = comp_get_drvdata(dev);
+	size_t size = sizeof(*cd);
+	cd = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, size);
+	if (!cd)
+		goto error;
+
+	return dev;
+
+	error:
```
lyakh, line 69: label at offset 0 please

lyakh (changes requested):

src/include/sof/audio/component.h
makarukp: It is possible. Originally I was thinking about that, however, I decided to treat device instance as manager which contain all information about all possible chains. Finally I don't see any problem to change it to one chain per device.

lyakh: interesting. I don't know much about these chains, what's their life-cycle? Aren't they also just components in pipelines, instantiated from the topology? If yes - then it would be logical to also treat them as normal components - one object per instance?

makarukp: Their lifecycle is independent from pipelines and outside of their scope, but from the logical point of view it is reasonable to treat them as separate components and identify each of them from ipc handler level. Maximum possible number of chains is defined as sum of HDA host inputs + HDA host outputs (platform dependent).

zephyr/CMakeLists.txt
```diff
@@ -564,6 +564,7 @@ zephyr_library_sources(
 	${SOF_AUDIO_PATH}/pipeline/pipeline-schedule.c
 	${SOF_AUDIO_PATH}/pipeline/pipeline-stream.c
 	${SOF_AUDIO_PATH}/pipeline/pipeline-xrun.c
+	${SOF_AUDIO_PATH}/pipeline/chain_dma.c
```
lgirdwood, line 567: Will need a Kconfig for this as it's probably and Intel only feature. We can make it default = y for TGL, ADL and MTL though.

src/audio/pipeline/chain_dma.c
```diff
@@ -0,0 +1,107 @@
+// SPDX-License-Identifier: BSD-3-Clause
```
lgirdwood, line 1: Since this is a processing component/module (since it's using the comp APIs) it would be best to be placed with the others (i.e. not in core pipeline directory).

zephyr/CMakeLists.txt
makarukp: Done

src/audio/chain_dma.c
```diff
@@ -0,0 +1,107 @@
...
+					 const void *ipc_specific_config)
+{
+	const struct ipc4_chain_dma *cdma = (struct ipc4_chain_dma *)ipc_specific_config;
+	struct comp_dev *dev;
+
+	dev = comp_alloc(drv, sizeof(*dev));
+	if (!dev)
+		return NULL;
+
+	struct chain_dma_data *cd;
```
RanderWang, line 32: move this definition up to the header of this function ?

src/audio/chain_dma.c
```diff
@@ -0,0 +1,107 @@
...
+{
+	switch (cmd) {
+	case COMP_TRIGGER_START:
+		chain_dma_start(dev);
+		break;
+	case COMP_TRIGGER_PAUSE:
+		chain_dma_pause(dev);
+		break;
+	default:
+		return 0;
```
RanderWang, line 74: break or return IPC4_SUCCESS ?

lgirdwood:
@makarukp is this PR still pending validation or implementation completion or external dependencies ?

src/audio/chain_dma.c
```diff
@@ -0,0 +1,103 @@
...
+	return dev;
+
+error:
+	rfree(dev);
+	return NULL;
+}
+
+int chain_dma_start(struct comp_dev *dev)
+{
+	return IPC4_SUCCESS;
```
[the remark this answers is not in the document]
makarukp: The content of these functions will be added in next commit which I am currently working on.

makarukp:
@lgirdwood This PR is still under development, it requires at lease one more commit with code completion before it is ready for validation.

makarukp:
Code is completed. PR is ready for validation.

src/ipc/ipc4/helper.c
```diff
@@ -658,6 +659,74 @@ static int construct_config(struct ipc4_copier_module_cfg *copier_cfg, uint32_t
...
+	if (allocate) {
+		if (enable)
+			ret = drv->ops.trigger(dev, COMP_TRIGGER_START);
+		else
+			ret = drv->ops.trigger(dev, COMP_TRIGGER_PAUSE);
+	} else {
+		if (enable)
+			return IPC4_INVALID_REQUEST;
+		ret = drv->ops.trigger(dev, COMP_TRIGGER_PAUSE);
+		if (ret == IPC4_SUCCESS)
```
ujfalusi, line 725: Is this condition right? Not `if (ret != IPC4_SUCCESS)` ?

ujfalusi: I see, this is the stop and release the ChainDMA and only free if the trigger was succesfull?

RanderWang:
@makarukp why not remove the original chain hda implementation ? This PR is a native solution for chain dma.

src/ipc/ipc4/helper.c
makarukp: I meant that if 'trigger pause' return any error, remove flow stops and return error without removing chain. I could transform it into more readible form:

```suggestion
		/* remove chain */
		ret = drv->ops.trigger(dev, COMP_TRIGGER_PAUSE);
		if (ret)
		        return ret;
		drv->ops.free(dev);
```
Does it look better ?

makarukp: > 

yes, exactly

makarukp: > I see, this is the stop and release the ChainDMA and only free if the trigger was succesfull?



> I see, this is the stop and release the ChainDMA and only free if the trigger was succesfull?

yes, exactly

makarukp:
> @makarukp why not remove the original chain hda implementation ? This PR is a native solution for chain dma.

Sure, I could remove the original implementation if it does not cover any special cases where hw chaining is not supported.

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+		return out_read_pos - in_read_pos;
+
+	return (buff_size - in_read_pos) + out_read_pos;
+}
+
+int get_connector_node_id(uint32_t dma_id, bool host_type,
+			  union ipc4_connector_node_id *connector_node_id)
+{
+	uint8_t type = host_type ? ipc4_hda_host_output_class : ipc4_hda_link_output_class;
+	uint8_t dma_idx = dma_id;
```
lyakh, line 499: you can just use `dma_id`

src/ipc/ipc4/helper.c
```diff
@@ -669,6 +669,7 @@ int ipc4_create_mng_for_chain(struct ipc4_chain_dma *cdma)
 	struct comp_dev *dev;
 
 	drv = ipc4_get_drv((uint8_t *)&uuid);
+
 	if (!drv)
 		return IPC4_INVALID_RESOURCE_ID;
```
lyakh, line 675: please drop unrelated white-space changes. In fact in my subjective perception most of the SOF code doesn't have an empty line before checking return code like
```
ret = fn();
if (ret < 0)
```

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
 {
 	switch (cmd) {
 	case COMP_TRIGGER_START:
-		return chain_dma_start(dev);
+		return chain_task_start(dev);
 	case COMP_TRIGGER_PAUSE:
-		return chain_dma_pause(dev);
+		return chain_task_pause(dev);
 	default:
 		return IPC4_ERROR_INVALID_PARAM;
```
lyakh, line 143: all component operations should return POSIX error codes, please fix.

lyakh (changes requested):

src/ipc/ipc4/helper.c
```diff
@@ -658,6 +659,74 @@ static int construct_config(struct ipc4_copier_module_cfg *copier_cfg, uint32_t
...
+	const struct comp_driver *drv = dev->drv;
+
+	uint32_t dir, host_chan, link_chan;
+
+	ret = process_dma_index(host_dma_id, &dir, &host_chan);
+
+	if (ret != IPC4_SUCCESS)
+		return ret;
+
+	ret = process_dma_index(link_dma_id, &dir, &link_chan);
```
RanderWang, line 710: why call process_dma_index ? these parameters are not used later

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+	int ret;
+
+	comp_info(dev, "chain_task_start(), host_dma_id = 0x%08x", cd->host_connector_node_id.dw);
+	if (cd->chain_task.state == SOF_TASK_STATE_QUEUED)
+		return IPC4_SUCCESS;
+
+	key = k_spin_lock(&drivers->lock);
+	if (!(cd->chain_task.state == SOF_TASK_STATE_COMPLETED
+	    || cd->chain_task.state == SOF_TASK_STATE_INIT))
+		return IPC4_BAD_STATE;
```
RanderWang, line 83: return IPC4_BAD_STATE; without spin_unlock ?

src/audio/chain_dma.c
```diff
@@ -0,0 +1,536 @@
...
+	cd->first_data_received = false;
+	cd->under_over_run_notification_sent = false;
+	cd->cs = scs ? 2 : 4;
+	cd->chain_task.state = SOF_TASK_STATE_INIT;
+
+	comp_set_drvdata(dev, cd);
+
+	ret = chain_task_init(dev, host_dma_id, link_dma_id, fifo_size);
+	if (ret)
+		goto error;
```
softwarecki, line 60: cd is not released if error occurred.

src/audio/chain_dma.c
```diff
@@ -0,0 +1,536 @@
...
+
+	/* retrieve current status of host DMA */
+	ret = dma_get_status(cd->chan_host->dma->z_dev, cd->chan_host->index, &stat);
+	if (ret < 0)
+		tr_err(&chain_dma_tr, "chain_task_run(): dma_get_status() failed, ret = %u",
+		       ret);
+
+	host_avail_bytes = stat.pending_length;
+	host_free_bytes = stat.free;
+	host_read_pos = stat.read_position;
```
softwarecki, line 183: Can we use values from the stat variable if dma_get_status returns an error?

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+	    || cd->chain_task.state == SOF_TASK_STATE_INIT))
+		return IPC4_BAD_STATE;
+	cd->under_over_run_notification_sent = false;
+
+	ret = chain_link_start(dev);
+	if (ret)
+		goto error;
+
+	ret = chain_host_start(dev);
+	if (ret)
```
RanderWang, line 91: the start order is the same for both playback & capture ?

RanderWang: if host start is failed, do we need to stop link ?

softwarecki (changes requested):

src/audio/chain_dma.c
```diff
@@ -18,13 +21,21 @@ DECLARE_SOF_RT_UUID("chain_dma", chain_dma_uuid, 0x6a0a274f, 0x27cc, 0x4afb,
 		    0xa3, 0xe7, 0x34, 0x44, 0x72, 0x3f, 0x43, 0x2e);
 DECLARE_TR_CTX(chain_dma_tr, SOF_UUID(chain_dma_uuid), LOG_LEVEL_INFO);
 
-static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
+static struct comp_dev *chain_task_create(const struct comp_driver *drv,
```
ujfalusi, line 24: Hrm, you are fixing what you have added in the first patch? Why not do it properly from the start?

ujfalusi: iow, squash the patches.

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+	struct comp_driver_list *drivers = comp_drivers_get();
+	int ret;
+
+	comp_info(dev, "chain_task_start(), host_dma_id = 0x%08x", cd->host_connector_node_id.dw);
+	if (cd->chain_task.state == SOF_TASK_STATE_QUEUED)
+		return IPC4_SUCCESS;
+
+	key = k_spin_lock(&drivers->lock);
+	if (!(cd->chain_task.state == SOF_TASK_STATE_COMPLETED
+	    || cd->chain_task.state == SOF_TASK_STATE_INIT))
```
ujfalusi, line 82: The `||` belongs at the end of the previous line.

src/audio/chain_dma.c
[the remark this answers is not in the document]
ujfalusi: POSIX error codes?

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+	case SOF_IPC_STREAM_CAPTURE:
+		dma_cfg_host->channel_direction = MEMORY_TO_HOST;
+		dma_block_cfg_host->source_address = (uint32_t)addr;
+		dma_cfg_link->channel_direction = PERIPHERAL_TO_MEMORY;
+		dma_block_cfg_link->dest_address = (uint32_t)addr;
+		break;
+	}
+
+	/* get host DMA channel */
+	int channel = cd->host_connector_node_id.f.v_index;
```
kv2019i, line 444: @makarukp  Quick drive-by comment. The establish practice in SOF is to declare local variables at start of code blocks.

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
 	return NULL;
 }
 
-int chain_dma_start(struct comp_dev *dev)
+int chain_task_start(struct comp_dev *dev)
 {
+	struct chain_dma_data *cd = comp_get_drvdata(dev);
+	k_spinlock_key_t key;
+	struct comp_driver_list *drivers = comp_drivers_get();
+	int ret;
```
ujfalusi, line 74: Please re-arrange the variables to reverse xmas tree (declarations longest to shortest)

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+
+	if (chain_stop(dev) < 0) {
+		k_spin_unlock(&drivers->lock, key);
+		return IPC4_FAILURE;
+	}
+
+	cd->chain_task.state = SOF_TASK_STATE_COMPLETED;
+	k_spin_unlock(&drivers->lock, key);
+	schedule_task_free(&cd->chain_task);
+	pm_policy_state_lock_put(PM_STATE_RUNTIME_IDLE, PM_ALL_SUBSTATES);
```
ujfalusi, line 130: The `pm_policy_state_lock_get()` is protected by the `drivers->lock` while here it is outside?

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
 	rfree(cd);
 	rfree(dev);
 }
 
+static enum task_state chain_task_run(void *data)
+{
+	struct dma_status stat;
+	size_t link_avail_bytes, link_free_bytes, host_avail_bytes, host_free_bytes;
+	uint32_t link_read_pos, host_read_pos;
+	struct chain_dma_data *cd = (struct chain_dma_data *)data;
```
ujfalusi, line 161: reverse xmas tree whenever possible.

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+
+	/* retrieve current status of link DMA */
+	int ret = dma_get_status(cd->chan_link->dma->z_dev, cd->chan_link->index,
+				 &stat);
+	if (ret < 0)
+		tr_err(&chain_dma_tr, "chain_task_run(): dma_get_status() failed, ret = %u",
+			 ret);
```
ujfalusi, line 168: alignment to parenthesis

ujfalusi: So, it is an error, yet we continue? So it is not an error, or?

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+	link_avail_bytes = stat.pending_length;
+	link_free_bytes = stat.free;
+	link_read_pos = stat.read_position;
+
+
+	/* retrieve current status of host DMA */
+	ret = dma_get_status(cd->chan_host->dma->z_dev, cd->chan_host->index, &stat);
+	if (ret < 0)
+		tr_err(&chain_dma_tr, "chain_task_run(): dma_get_status() failed, ret = %u",
+		       ret);
```
ujfalusi, line 179: Is ti error or not?

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+	uint32_t link_type = cd->link_connector_node_id.f.dma_type;
+
+	if (link_type == ipc4_hda_link_input_class) {
+		const size_t increment = MIN(host_free_bytes, link_avail_bytes);
+
+		ret = dma_reload(cd->chan_host->dma->z_dev, cd->chan_host->index,
+				 0, 0, increment);
+		if (ret < 0)
+			tr_err(&chain_dma_tr, "chain_task_run(): dma_reload() host failed, ret = %u",
+			       ret);
```
ujfalusi, line 194: here as well, is it error or not?

also break the lines:
```c
			tr_err(&chain_dma_tr,
			       "chain_task_run(): dma_reload() link failed, ret = %u",
			       ret);
```

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+	} else {
+		const size_t half_buff_size = cd->dma_buffer->stream.size / 2;
+
+		if ((!cd->first_data_received) && (host_avail_bytes > half_buff_size)) {
+			ret = dma_reload(cd->chan_link->dma->z_dev,
+					 cd->chan_link->index, 0, 0,
+					 half_buff_size);
+			if (ret < 0)
+				tr_err(&chain_dma_tr, "chain_task_run(): dma_reload() link failed, ret = %u",
+				       ret);
```
ujfalusi, line 210: make these lines shorter by breaking it down.
and is this error or warn?

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+			cd->first_data_received = true;
```
ujfalusi, line 211: the `dma_reload()` failed and we say that everything is OK?

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+				tr_err(&chain_dma_tr, "chain_task_run(): dma_reload() host failed, ret = %u",
+				       ret);
+
+			if ((host_avail_bytes >= half_buff_size) &&
+			    (link_free_bytes >= half_buff_size)) {
+				ret = dma_reload(cd->chan_link->dma->z_dev, cd->chan_link->index,
+						 0, 0, half_buff_size);
+				if (ret < 0)
+					tr_err(&chain_dma_tr, "chain_task_run(): dma_reload() link failed, ret = %u",
+					       ret);
```
ujfalusi, line 230: Lines are too long, please make them shorter, reword and/or break them up.

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+	return SOF_TASK_STATE_RESCHEDULE;
+}
+
+int chain_link_start(struct comp_dev *dev)
+{
+	struct chain_dma_data *cd = comp_get_drvdata(dev);
+	int err = dma_start(cd->chan_link->dma->z_dev, cd->chan_link->index);
+
+	if (err < 0)
+		return err;
```
ujfalusi, line 244: this is hard to read like this, separate the declaration and the call to dma_start().

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+
+	comp_info(dev, "chain_start(): dma_start() link chan_index = %u",
```
ujfalusi, line 246: this is the `chain_link_start()` call, probably: "chain_link_start() link chan_index = %u" ?

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+int chain_host_start(struct comp_dev *dev)
+{
+	struct chain_dma_data *cd = comp_get_drvdata(dev);
+	int err = dma_start(cd->chan_host->dma->z_dev, cd->chan_host->index);
+
+	if (err < 0)
+		return err;
+
+	comp_info(dev, "chain_host_start(): dma_start() host chan_index = %u",
+		  cd->chan_host->index);
```
ujfalusi, line 260: same comments as for the link side.

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+
+	if (err < 0)
+		return err;
+	comp_info(dev, "chain_stop(): dma_stop() link chan_index = %u",
+		  cd->chan_link->index);
+
+	err = dma_stop(cd->chan_host->dma->z_dev, cd->chan_host->index);
+	if (err < 0)
+		return err;
+	comp_info(dev, "chain_stop(): dma_stop() host chan_index = %u",
```
ujfalusi, line 277: here as well.

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+}
+
+int chain_stop(struct comp_dev *dev)
+{
+	struct chain_dma_data *cd = comp_get_drvdata(dev);
+	int err = dma_stop(cd->chan_link->dma->z_dev, cd->chan_link->index);
+
+	if (err < 0)
+		return err;
+	comp_info(dev, "chain_stop(): dma_stop() link chan_index = %u",
```
ujfalusi, line 271: here a blank line after the if block would make things cleaner.

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+int chain_task_init(struct comp_dev *dev, uint8_t host_dma_id, uint8_t link_dma_id,
+		    uint32_t fifo_size)
+{
+	struct comp_buffer __sparse_cache *buffer_c;
+	struct chain_dma_data *cd = comp_get_drvdata(dev);
+	uint32_t dir;
+	int ret;
+
+	schedule_task_init_ll(&cd->chain_task,
+			      SOF_UUID(chain_dma_uuid),
```
ujfalusi, line 305: You can move the SOF_UUID up one line?

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+		cd->stream_direction = SOF_IPC_STREAM_CAPTURE;
+	}
+
+	/* request HDA DMA with shared access privilege */
+	dir = (cd->stream_direction == SOF_IPC_STREAM_PLAYBACK) ?
+		DMA_DIR_HMEM_TO_LMEM : DMA_DIR_LMEM_TO_HMEM;
+
+	cd->dma_host = dma_get(dir, 0, DMA_DEV_HOST, DMA_ACCESS_SHARED);
+
+	if (!cd->dma_host) {
```
ujfalusi, line 335: No need for the blank line

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+		comp_err(dev, "chain_task_init(): dma_get() returned NULL");
+		return IPC4_FAILURE;
+	}
+
+	/* retrieve DMA buffer address alignment */
+	uint32_t addr_align;
+	int err = dma_get_attribute(cd->dma_host, DMA_ATTR_BUFFER_ADDRESS_ALIGNMENT, &addr_align);
+
+	if (err < 0) {
+		comp_err(dev, "chain_task_init(): could not get dma buffer address alignment, err = %d",
```
ujfalusi, line 355: Long line, move the string to a new one to shorten it (can it be re-worded to shorten it even further?)

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+	}
+
+	fifo_size = ALIGN_UP_INTERNAL(fifo_size, addr_align);
+
+	if (cd->link_connector_node_id.f.dma_type == ipc4_hda_link_input_class) {
+		/* Increasing buffer size for capture path as L1SEN exit takes sometimes
+		 * more than expected.
+		 */
+		fifo_size *= 5;
+	}
```
ujfalusi, line 367: The curly brackets can be dropped and the comment moved before the if with a small reworking to include the reason why the fifo_size needs to be multiplied by 5.

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+	if (cd->link_connector_node_id.f.dma_type == ipc4_hda_link_output_class) {
+		/* Increasing buffer size for playback path as L1SEN exit takes sometimes
+		 * more that expected
+		 * Note, FIFO size must be smaller than half of host buffer size
+		 * (20ms ping pong) to avoid problems with position reporting
+		 * Size increase from default 2ms to 5ms is enough.
+		 */
+		fifo_size /= 2;
+		fifo_size *= 5;
+	}
```
ujfalusi, line 378: if it is
`cd->link_connector_node_id.f.dma_type != ipc4_hda_link_input_class && cd->link_connector_node_id.f.dma_type != ipc4_hda_link_output_class` then the fifo_size does not need to be modified and it is valid?

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+	return IPC4_SUCCESS;
+}
+
+int chain_init(struct comp_dev *dev, void *addr, size_t length)
+{
+	struct chain_dma_data *cd = comp_get_drvdata(dev);
+	struct dma_config *dma_cfg_host = &cd->z_config_host;
+	struct dma_config *dma_cfg_link = &cd->z_config_link;
+	struct dma_block_config *dma_block_cfg_host = &cd->dma_block_cfg_host;
+	struct dma_block_config *dma_block_cfg_link = &cd->dma_block_cfg_link;
```
ujfalusi, line 409: reverse xmas tree

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+	dma_cfg_host->source_data_size = cd->cs;
+	dma_cfg_host->dest_data_size = cd->cs;
+	dma_cfg_host->head_block = dma_block_cfg_host;
+	dma_cfg_link->block_count = 1;
+	dma_cfg_link->source_data_size = cd->cs;
+	dma_cfg_link->dest_data_size = cd->cs;
+	dma_cfg_link->head_block  = dma_block_cfg_link;
+
+	dma_block_cfg_host->block_size = length;
+	dma_block_cfg_link->block_size = length;
```
ujfalusi, line 426: I would collect the host and the link struct clearing and setup into separate blocks and not mixed them

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+		comp_err(dev, "chain_init(): dma_request_channel() failed");
+		cd->chan_host = NULL;
+		return IPC4_FAILURE;
+	}
+
+	cd->chan_host = &cd->dma_host->chan[channel];
+
+	int err = dma_config(cd->dma_host->z_dev, cd->chan_host->index, dma_cfg_host);
+
+	if (err < 0) {
```
ujfalusi, line 457: drop the blank line and move the `int ret;` declaration at the start of the function?

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+		dma_block_cfg_host->source_address = (uint32_t)addr;
+		dma_cfg_link->channel_direction = PERIPHERAL_TO_MEMORY;
+		dma_block_cfg_link->dest_address = (uint32_t)addr;
+		break;
+	}
+
+	/* get host DMA channel */
+	int channel = cd->host_connector_node_id.f.v_index;
+
+	channel = dma_request_channel(cd->dma_host->z_dev, &channel);
```
ujfalusi, line 446: It would be easier to follow if you do:
```c
	channel = dma_request_channel(cd->dma_host->z_dev,
				      &cd->host_connector_node_id.f.v_index);
...
	channel = dma_request_channel(cd->dma_host->z_dev,
				      &cd->link_connector_node_id.f.v_index);
```

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+	if (channel < 0) {
+		comp_err(dev, "chain_init(): dma_request_channel() failed");
+		cd->chan_host = NULL;
```
ujfalusi, line 449: Was cd->chan_host not NULL before? Is that OK to override it?

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+
+int get_connector_node_id(uint32_t dma_id, bool host_type,
+			  union ipc4_connector_node_id *connector_node_id)
+{
+	uint8_t type = host_type ? ipc4_hda_host_output_class : ipc4_hda_link_output_class;
+	uint8_t dma_idx = dma_id;
+	uint32_t host_output_quantity = DAI_NUM_HDA_OUT;
+	uint32_t host_input_quantity = DAI_NUM_HDA_IN;
+
+	if (dma_idx >= host_output_quantity) {
```
ujfalusi, line 503: You can use the defines in place?

src/audio/chain_dma.c
```diff
@@ -37,56 +48,478 @@ static struct comp_dev *chain_dma_create(const struct comp_driver *drv,
...
+		type = host_type ? ipc4_hda_host_input_class : ipc4_hda_link_input_class;
+		dma_idx = dma_id - host_output_quantity;
+		if (!(dma_idx < host_input_quantity))
+			return IPC4_ERROR_INVALID_PARAM;
+	}
```
ujfalusi, line 508: It might be easier to follow if you do this with if () {} else {} ?

src/include/sof/audio/chain_dma.h
```diff
@@ -12,32 +12,77 @@
...
+
+int chain_task_release(struct comp_dev *dev);
+
+void chain_task_free(struct comp_dev *dev);
+
+static enum task_state chain_task_run(void *data);
+
+int chain_task_trigger(struct comp_dev *dev, int cmd);
+
+int chain_task_init(struct comp_dev *dev, uint8_t host_dma_id, uint8_t link_dma_id,
```
ujfalusi, line 63: No need for the blank lines.

src/ipc/ipc4/handler.c
```diff
@@ -456,48 +456,39 @@ static int ipc4_load_library(struct ipc4_message_request *ipc4)
...
 	memcpy_s(&cdma, sizeof(cdma), ipc4, sizeof(cdma));
 
-	if (cdma.primary.r.allocate && cdma.extension.r.fifo_size) {
-		ret = ipc4_create_chain_dma(ipc, &cdma);
-		if (ret) {
-			tr_err(&ipc_tr, "failed to create chain dma %d", ret);
-			return ret;
-		}
+	int comp_id = IPC4_COMP_ID(cdma.primary.r.host_dma_id
+				   + IPC4_MAX_MODULE_COUNT, 0);
```
ujfalusi, line 467: this could fit in one line.

src/ipc/ipc4/handler.c
```diff
@@ -456,48 +456,39 @@ static int ipc4_load_library(struct ipc4_message_request *ipc4)
...
 
-		/* if enable is not set, chain dma pipeline is not going to be triggered */
-		if (!cdma.primary.r.enable)
+	struct ipc_comp_dev *cdma_comp = ipc_get_comp_by_id(ipc, comp_id);
```
ujfalusi, line 469: Move the declarations to the start of the function?

src/include/sof/audio/chain_dma.h
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// Copyright(c) 2022 Intel Corporation. All rights reserved.
+//
+// Author: Piotr Makaruk <piotr.makaruk@intel.com>
+
```
tlissows, line 6: Consider using 'include guard' for this file.

`#ifndef __SOF_CHAIN_DMA_H__`
`#define __SOF_CHAIN_DMA_H__`
`#endif /* __SOF_CHAIN_DMA_H__ */`

marc-hb: I think @tlissows means this:
```
#ifndef __SOF_CHAIN_DMA_H__
#define __SOF_CHAIN_DMA_H__

  < entire file goes here >

#endif /* __SOF_CHAIN_DMA_H__ */
```

src/audio/chain_dma.c
makarukp: Yes. Currently dma_get_status() (for hda) returns always 0, however, there is plan to implement xrun reporting with error(but not critical one) which has to be properly handled using stat in  chain_task_run() so the task should be continued.
Analogical example: dw_dma_get_status() in Zephyr.

src/audio/chain_dma.c
makarukp: Not sure if I well understood you... However, I simplified that function. Does it look better ?
`
int get_connector_node_id(uint32_t dma_id, bool host_type,
			  union ipc4_connector_node_id *connector_node_id)
{
	uint8_t type = host_type ? ipc4_hda_host_output_class : ipc4_hda_link_output_class;

	if (dma_id >= DAI_NUM_HDA_OUT) {
		type = host_type ? ipc4_hda_host_input_class : ipc4_hda_link_input_class;
		dma_id = dma_id - DAI_NUM_HDA_OUT;
		if (dma_id >= DAI_NUM_HDA_IN)
			return -EINVAL;
	}
	connector_node_id->dw = 0;
	connector_node_id->f.dma_type = type;
	connector_node_id->f.v_index = dma_id;

	return 0;
}
`

src/audio/chain_dma.c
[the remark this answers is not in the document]
makarukp: Currently dma_get_status() (for hda) returns always 0, however, there is plan to implement xrun reporting with error(but not critical one) which has to be properly handled using stat in chain_task_run() so the task should be continued.
Analogical example: dw_dma_get_status() in Zephyr.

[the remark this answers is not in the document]
makarukp: The same case with dma_reload().

src/audio/chain_dma.c
makarukp: This is not critical error (xrun detection), so task flow should continue and report in logs.

src/audio/chain_dma.c
makarukp: It could return error when xrun is detected. Then the task should report an error and continue with flow.

src/audio/chain_dma.c
makarukp: If exists, not a critical error as well.

src/audio/chain_dma.c
makarukp: These all looks more like warning, so I change it accordingly.

src/audio/chain_dma.c
makarukp: These all looks more like warning, so I change it accordingly.

src/audio/chain_dma.c
makarukp: Only 2 classes of dma link type are possible:  ipc4_hda_link_input_class or ipc4_hda_link_output_class. It is always being modified. Validation node id and type assignment is realized in get_connector_node_id(). I replaced it with switch. Is it more clear ?

src/audio/chain_dma.c
```diff
@@ -0,0 +1,596 @@
...
+
+	ret = chain_task_init(dev, host_dma_id, link_dma_id, fifo_size);
+	if (ret)
+		goto error2;
+
+	return dev;
+
+error2:
+	rfree(cd);
+	goto error1;
```
softwarecki, line 63: goto not needed

src/audio/chain_dma.c
```diff
@@ -0,0 +1,596 @@
...
+		if (ret)
+			goto error2;
+	} else
+	{
+		ret = chain_link_start(dev);
+		if (ret)
+			goto error1;
+		ret = chain_host_start(dev);
+		if (ret)
+			goto error3;
```
softwarecki, line 108: error 3 is used only here. Maybe instead of jumping between labels call chain_link_stop here and jump to the error handler?

src/audio/chain_dma.c
```diff
@@ -0,0 +1,596 @@
...
+		k_spin_unlock(&drivers->lock, key);
+		return 0;
+	case SOF_TASK_STATE_COMPLETED:
+		break;
+	case SOF_TASK_STATE_INIT:
+		break;
+	default:
+		comp_err(dev, "chain_task_start(), bad state transition");
+		k_spin_unlock(&drivers->lock, key);
+		return -EINVAL;
```
softwarecki, line 90: looks like goto error1

src/audio/chain_dma.c
```diff
@@ -0,0 +1,592 @@
...
+	k_spin_unlock(&drivers->lock, key);
+
+	return 0;
+error2:
+	chain_host_stop(dev);
+	goto error1;
+error3:
+	chain_link_stop(dev);
+	goto error1;
+error1:
```
ujfalusi, line 170: Labels should have meaningful names but at least they should be in order?

src/audio/chain_dma.c
```diff
@@ -0,0 +1,592 @@
...
+	pm_policy_state_lock_put(PM_STATE_RUNTIME_IDLE, PM_ALL_SUBSTATES);
+	k_spin_unlock(&drivers->lock, key);
+
+	return 0;
+error2:
+	chain_host_stop(dev);
+	goto error1;
+error3:
+	chain_link_stop(dev);
+	goto error1;
```
ujfalusi, line 169: This does not make sense:
you jump here from line 155 _after_ the `chain_link_stop()` has been successful but the `chain_host_stop()` failed and you do another `chain_link_stop()`?

src/audio/chain_dma.c
```diff
@@ -0,0 +1,592 @@
...
+	if (cd->chain_task.state == SOF_TASK_STATE_FREE)
+		return 0;
+
+	key = k_spin_lock(&drivers->lock);
+	cd->first_data_received = false;
+	if (cd->stream_direction == SOF_IPC_STREAM_PLAYBACK)
+	{
+		ret = chain_host_stop(dev);
+		if (ret < 0)
+			goto error1;
```
ujfalusi, line 145: So, if the host fail to stop, we don't bother with the link?

src/audio/chain_dma.c
```diff
@@ -0,0 +1,592 @@
...
+		ret = chain_link_stop(dev);
+		if (ret < 0)
+			goto error2;
+	} else {
+		ret = chain_link_stop(dev);
+		if (ret < 0)
+			goto error1;
```
ujfalusi, line 152: So, if the link fail to stop, we don't bother with the host?

src/audio/chain_dma.c
makarukp: I see... but compiler cannot take address of bit-field 'v_index', can we stay with anything like previous approach with declaration on the beginning ?

src/audio/chain_dma.c
```diff
@@ -0,0 +1,587 @@
...
+void chain_task_free(struct comp_dev *dev)
+{
+	struct chain_dma_data *cd = comp_get_drvdata(dev);
+
+	chain_release(dev);
+	rfree(cd);
+	rfree(dev);
+}
+
+static enum task_state chain_task_run(void *data)
```
lgirdwood, line 193: fwiw, this is quite a complex function - it would benefit from more inline comments to help everyone follow.

src/ipc/ipc4/helper.c
```diff
@@ -658,6 +659,61 @@ static int construct_config(struct ipc4_copier_module_cfg *copier_cfg, uint32_t
...
+	const bool enable = cdma->primary.r.enable;
+	int ret;
+
+	if (!dev)
+		return -EINVAL;
+	const struct comp_driver *drv = dev->drv;
+
+	if (allocate) {
+		if (enable)
+			ret = drv->ops.trigger(dev, COMP_TRIGGER_START);
```
lgirdwood, line 700: can we use the wrappers here rather than deref the ops pointers.

makarukp: Not sure if I understood you well, do you mean defining in chain_dma_state func something like this ? 
`
int (*trigger)(struct comp_dev *dev, int cmd);
trigger = &drv->ops.trigger;
`

lgirdwood: We should have static inline wrapper function for all DAI/DMA ops-> (as some are optional ops and we safely handle).
See component_ext.h i.e.
```c
/** See comp_ops::trigger */
static inline int comp_trigger(struct comp_dev *dev, int cmd)
{
	assert(dev->drv->ops.trigger);

	return (dev->is_shared && !cpu_is_me(dev->ipc_config.core)) ?
		comp_trigger_remote(dev, cmd) : dev->drv->ops.trigger(dev, cmd);
}
```

src/audio/chain_dma.c
```diff
@@ -0,0 +1,596 @@
...
+			return -EINVAL;
+	}
+	connector_node_id->dw = 0;
+	connector_node_id->f.dma_type = type;
+	connector_node_id->f.v_index = dma_id;
+
+	return 0;
+}
+
+static const struct comp_driver comp_chain_dma = {
```
lgirdwood, line 576: @makarukp the comp_driver model/API is being deprecated in favor of the module API, i.e. these 3 ops will need changed to be module API ops either now or later.

makarukp: I see... Do you mean by changing it to module_interface from module_interface.h like in mixin_mixout module ?

lgirdwood:
@makarukp can you check CI thanks.

marc-hb:
Only non-Intel platforms compile in https://github.com/thesofproject/sof/actions/runs/3883238358/jobs/6624317343

I don't see the point of submitting code that does not compile in any configuration but maybe I'm missing something.

In any case always submit PRs as a Github "draft" first to avoid spamming reviewers. You can also convert to a draft later but that does not automatically unsubscribe reviewers.

src/audio/chain_dma.c
makarukp: I decided that due to lack of extended module api (extra dependency under discussion), transfer to module API will be added in another PR.

dabekjakub (approved):

src/ipc/ipc4/handler.c
```diff
@@ -460,49 +460,41 @@ static int ipc4_load_library(struct ipc4_message_request *ipc4)
...
-		}
-	}
+	cdma_comp = ipc_get_comp_by_id(ipc, comp_id);
+	if (!cdma_comp)
+		return IPC4_FAILURE;
 
-	return ret;
+	ret = ipc4_chain_dma_state(cdma_comp->cd, &cdma);
+	if (ret < 0)
+		return IPC4_FAILURE;
```
lyakh, line 489: don't we need to revert the effect of `ipc4_create_mng_for_chain()` in case of an error?

src/ipc/ipc4/handler.c
```diff
@@ -460,49 +460,41 @@ static int ipc4_load_library(struct ipc4_message_request *ipc4)
...
+
+	if (!cdma.primary.r.allocate && !cdma.primary.r.allocate)
```
lyakh, line 491: something's wrong here

src/ipc/ipc4/handler.c
makarukp: Sure, but I think only in case of first COMP_TRIGGER_START failure, otherwise we may return IPC4_INVALID_CHAIN_STATE_TRANSITION.

src/audio/CMakeLists.txt
```diff
@@ -107,6 +107,11 @@ if(NOT CONFIG_LIBRARY)
 	endif()
 	subdirs(pipeline)
 	add_subdirectory(google)
+	if(CONFIG_COMP_CHAIN_DMA)
+		add_local_sources(sof
```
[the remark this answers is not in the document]
makarukp: I simplified it to one line, but  according to rules syntax is 'add_local_sources(${lib_name} ${ARGN})' correct ;)

mwasko (approved):
The code looks good and functionality has been verified. @lyakh if you have no more comments then please let us know so we can proceed with the merge.

src/audio/chain_dma.c
```diff
@@ -0,0 +1,588 @@
...
+		if (cd->link_connector_node_id.f.dma_type != ipc4_hda_link_input_class)
+			return -EINVAL;
+		cd->stream_direction = SOF_IPC_STREAM_CAPTURE;
+	}
+
+	/* request HDA DMA with shared access privilege */
+	dir = (cd->stream_direction == SOF_IPC_STREAM_PLAYBACK) ?
+		DMA_DIR_HMEM_TO_LMEM : DMA_DIR_LMEM_TO_HMEM;
+
+	cd->dma_host = dma_get(dir, 0, DMA_DEV_HOST, DMA_ACCESS_SHARED);
```
lyakh, line 388: don't we need matching `dma_put()` for these?

src/audio/chain_dma.c
```diff
@@ -0,0 +1,588 @@
...
+
+	/* clear dma buffer */
+	buffer_c = buffer_acquire(cd->dma_buffer);
+	buffer_zero(buffer_c);
+	buff_addr = cd->dma_buffer->stream.addr;
+	buff_size = cd->dma_buffer->stream.size;
+	buffer_release(buffer_c);
+
+	ret = chain_init(dev, buff_addr, buff_size);
+	if (ret)
```
lyakh, line 450: `buffer_free()`?

src/ipc/ipc4/helper.c
```diff
@@ -659,6 +662,61 @@ static int construct_config(struct ipc4_copier_module_cfg *copier_cfg, uint32_t
...
+			0x44, 0x72, 0x3f, 0x43, 0x2e}};
+	const struct comp_driver *drv;
+	struct comp_dev *dev;
+
+	drv = ipc4_get_drv((uint8_t *)&uuid);
+	if (!drv)
+		return -EINVAL;
+
+	dev = drv->ops.create(drv, (const struct comp_ipc_config *)NULL,
+			      (const void *)cdma);
```
lyakh, line 533: both typecasts aren't needed

src/ipc/ipc4/handler.c
```diff
@@ -460,49 +460,53 @@ static int ipc4_load_library(struct ipc4_message_request *ipc4)
...
-	} else if (!cdma.primary.r.allocate) {
-		uint32_t pipeline_id = IPC4_COMP_ID(cdma.primary.r.host_dma_id
-						    + IPC4_MAX_MODULE_COUNT,
-						    cdma.primary.r.link_dma_id);
-
-		/* waiting for pipeline reset done */
-		ipc_wait_for_compound_msg();
-		ret = ipc_pipeline_free(ipc, pipeline_id);
+		cdma_comp = ipc_get_comp_by_id(ipc, comp_id);
+		if (!cdma_comp)
```
lyakh, line 483: `ipc4_create_mng_for_chain()` doesn't have to be reverted?

makarukp:
@lyakh What do you think about such approach of critical error handling in task in response to 'dma_get_status()' ?
'
	/* Status can return -ENODATA and current status if xrun occurs, then it is not critical
	 * and flow shall continue. In worst case, when channel index exceeds max value, status can
	 * return -EINVAL, then it is critical error.
	 */
	ret = dma_get_status(cd->chan_link->dma->z_dev, cd->chan_link->index, &stat);
	switch (ret) {
	case 0:
		break;
	case -ENODATA:
		tr_warn(&chain_dma_tr, "chain_task_run(): dma_get_status() xrun occured, ret = %u",
		ret);
		break;
	default:
		tr_err(&chain_dma_tr, "chain_task_run(): dma_get_status() error, ret = %u", ret);
	}
'

lyakh:
> Status can return -ENODATA and current status if xrun occurs

According to https://www.alsa-project.org/alsa-doc/alsa-lib/pcm.html `-EPIPE` is used by ALSA to indicate an xrun?

makarukp:
@lyakh Accually -EINVAL can be returned in current implementation only by dw_dma_get_status() which is used by GPDMA in Zephyr. DMA uses intel_adsp_hda_dma_status() and there is only assertion  `__ASSERT(channel < cfg->dma_channels, "Channel does not exist");` .

src/audio/chain_dma.c
```diff
@@ -0,0 +1,588 @@
...
+	rfree(dev);
+	return NULL;
+}
+
+int chain_task_start(struct comp_dev *dev)
+{
+	struct comp_driver_list *drivers = comp_drivers_get();
+	struct chain_dma_data *cd = comp_get_drvdata(dev);
+	k_spinlock_key_t key;
+
```
kv2019i, line 128: Minor: extra line L74

src/ipc/ipc4/handler.c
```diff
@@ -460,49 +460,53 @@ static int ipc4_load_library(struct ipc4_message_request *ipc4)
...
+		return IPC4_FAILURE;
+
+	ret = ipc4_chain_dma_state(cdma_comp->cd, &cdma);
+	if (ret < 0)
+		return IPC4_INVALID_CHAIN_STATE_TRANSITION;
+
+	if (!cdma.primary.r.allocate && !cdma.primary.r.enable)
+		list_item_del(&cdma_comp->list);
+
+	return  IPC4_SUCCESS;
```
kv2019i, line 508: Minor: extra whitespace, should be "return IPC4_SUCCESS"

app/boards/intel_adsp_cavs25_tgph.conf
```diff
@@ -1,5 +1,6 @@
 CONFIG_TIGERLAKE=y
```
kv2019i, line 1: typo: in git commit summary: s/thl-h/tgl-h/. 
Also as a style issue, the git summary should follow "FOO: BAR: blaa-baa" style, i.e. space after each prefix. We've not been very consistent with the "app/" folder commits so far, so this is not a big deal, but a note for future commits.

src/audio/chain_dma.c
```diff
@@ -0,0 +1,656 @@
...
+	/* local link DMA config */
+	struct dma *dma_link;
+	struct dma_chan_data *chan_link;
+	struct dma_config z_config_link;
+	struct dma_block_config dma_block_cfg_link;
+
+	struct comp_buffer *dma_buffer;
+};
+
+int chain_host_start(struct comp_dev *dev)
```
lyakh, line 65: now that you removed the header, please make all the functions in this file `static`

src/ipc/ipc4/handler.c
```diff
@@ -460,49 +460,55 @@ static int ipc4_load_library(struct ipc4_message_request *ipc4)
...
-		}
+	comp_id = IPC4_COMP_ID(cdma.primary.r.host_dma_id + IPC4_MAX_MODULE_COUNT, 0);
+	cdma_comp = ipc_get_comp_by_id(ipc, comp_id);
 
-		/* if enable is not set, chain dma pipeline is not going to be triggered */
-		if (!cdma.primary.r.enable)
-			return ret;
-	}
+	if (!cdma_comp && cdma.primary.r.allocate && cdma.primary.r.enable) {
+		ret = ipc4_create_mng_for_chain(&cdma);
```
lyakh, line 478: `ipc4_chain_manager_create()` seems to be a better name for this, but it's already there

src/audio/chain_dma.c
lgirdwood: @lyakh this looks resolved now, error code is checked ?

src/audio/chain_dma.c
lgirdwood: @ujfalusi resolved ?

lgirdwood (approved):

lyakh (approved):

mwasko (approved):

marc-hb:
This apparently caused regressions #6975 and #6976.

It was also merged while failing a large number of tests: https://sof-ci.01.org/sofpr/PR6656/build3409/devicetest/index.html

lgirdwood:
There is a lot of test failures on ADL capture, I've created a revert PR that can be merged today prior to daily test runs if we don't have fixes ready in time.

lgirdwood:
@makarukp @mwasko @kv2019i @lyakh fyi - does anyone have fix PRs in progress otherwise we will have to revert ?

mwasko:
> @makarukp @mwasko @kv2019i @lyakh fyi - does anyone have fix PRs in progress otherwise we will have to revert ?

@lgirdwood, @kv2019i the regression is affecting only cAVS targets which are not POR in Zephyr configuration. The tests were passing before on cAVS because they were using w/a in form of copier-copier solution which is not correct implementation. We should treat it as we have enabled feature for MTL and for cAVS it is a to do action (on best effort basis).

lgirdwood:
> > @makarukp @mwasko @kv2019i @lyakh fyi - does anyone have fix PRs in progress otherwise we will have to revert ?
> 
> @lgirdwood, @kv2019i the regression is affecting only cAVS targets which are not POR in Zephyr configuration. The tests were passing before on cAVS because they were using w/a in form of copier-copier solution which is not correct implementation. We should treat it as we have enabled feature for MTL and for cAVS it is a to do action (on best effort basis).

I'm fine to make this a MTL only feature, but it should not break legacy features on other platforms. Can this be fixed with a Kconfig ?

makarukp:
> > > @makarukp @mwasko @kv2019i @lyakh fyi - does anyone have fix PRs in progress otherwise we will have to revert ?
> > 
> > 
> > @lgirdwood, @kv2019i the regression is affecting only cAVS targets which are not POR in Zephyr configuration. The tests were passing before on cAVS because they were using w/a in form of copier-copier solution which is not correct implementation. We should treat it as we have enabled feature for MTL and for cAVS it is a to do action (on best effort basis).
> 
> I'm fine to make this a MTL only feature, but it should not break legacy features on other platforms. Can this be fixed with a Kconfig ?

Unfortunately it is not possible to fix it with Kconfig because legacy approach was erased by hw chain dma PR. 

As for https://github.com/thesofproject/sof/issues/6975#event-8319185716 issue it does not look like a big deal. I will prepare debug FW which could probably solve that problem or provide mor information. 

As for https://github.com/thesofproject/sof/issues/6976#event-8319192834 issue I don't know what happened and still thinking about further steps for debug ...

makarukp:
I prepared fix for #6975:
https://github.com/thesofproject/sof/pull/6983

PR #6661, opened 2022-11-24
sof: common: namespace common DIV_ROUND_UP macro

gmarull:
DIV_ROUND_UP is a common macro exposed in public headers without namespacing. Change the name to SOF_DIV_ROUND_UP to avoid collisions with other systems/libraries.

See https://github.com/zephyrproject-rtos/zephyr/pull/52488 for more context.

Signed-off-by: Gerard Marull-Paretas <gerard.marull@nordicsemi.no>


sofci:

lgirdwood:
test this please

lgirdwood:
@singalsu is the Zephyr macro a duplication of teh SOF macro - if so we should use the Zephyr macro directly and only redefine for xtos.

gmarull:
> @singalsu is the Zephyr macro a duplication of teh SOF macro - if so we should use the Zephyr macro directly and only redefine for xtos.

Zephyr wants to introduce such macro, but sof already defines it. But Zephyr includes a sof fork as a module, so it's kind of a chicken or the egg first problem. If you're happy doing:

```c
#ifdef __ZEPHYR__
#include <zephyr/sys/util.h>
#else
#define DIV_ROUND_UP(..._)
#endif
```

we can first patch Zephyr sof fork, add the feature, and then backport the commit upstream.

abonislawski:
Please stay with the original name, just like ARRAY_SIZE
```
/* use same syntax as Linux for simplicity */
#ifndef __ZEPHYR__ /* Already present and compatible via Zephyr headers */
#define ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))
#endif
```

gmarull:
> Please stay with the original name, just like ARRAY_SIZE
> 
> ```
> /* use same syntax as Linux for simplicity */
> #ifndef __ZEPHYR__ /* Already present and compatible via Zephyr headers */
> #define ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))
> #endif
> ```

This is missing `#include <zephyr/sys/util.h>`

gmarull:
Opened https://github.com/zephyrproject-rtos/sof/pull/22 (can be backported to upstream once Zephyr is updated).

lgirdwood (commented):
@gmarull can you update this PR to align with the PR you posted to the Zephyr SOF module. I think we are good at that point.
Fwiw, we are in the middle of an RTOS abstraction update, my expectation is we will be able to cleanly separate similar collisions when the abstraction is done.

gmarull:
> @gmarull can you update this PR to align with the PR you posted to the Zephyr SOF module. I think we are good at that point. Fwiw, we are in the middle of an RTOS abstraction update, my expectation is we will be able to cleanly separate similar collisions when the abstraction is done.

I can't, since Zephyr needs to be updated first. And to update Zephyr, I need to patch SOF Zephyr fork before... (I'm not sure why Zephyr pulls SOF as a module, TBH).

lgirdwood:
> > @gmarull can you update this PR to align with the PR you posted to the Zephyr SOF module. I think we are good at that point. Fwiw, we are in the middle of an RTOS abstraction update, my expectation is we will be able to cleanly separate similar collisions when the abstraction is done.
> 
> I can't, since Zephyr needs to be updated first. And to update Zephyr, I need to patch SOF Zephyr fork before... (I'm not sure why Zephyr pulls SOF as a module, TBH).

@kv2019i @nashif chicken/egg how do we solve ?

kv2019i:
@gmarull wrote:
>> I can't, since Zephyr needs to be updated first. And to update Zephyr, I need to patch SOF Zephyr fork before... (I'm not sure why Zephyr pulls SOF as a module, TBH).
> 
> @kv2019i @nashif chicken/egg how do we solve ?

I think we need to do in two steps. E.g. let's merge this first (introducing SOF_DIV_ROUND_UP), second the SOF module can be updated in Zephyr, and third SOF can be moved to use Zephyr DIV_ROUND_UP directly.

gmarull:
> @lgirdwood wrote:
> 
> > @gmarull wrote:
> > > I can't, since Zephyr needs to be updated first. And to update Zephyr, I need to patch SOF Zephyr fork before... (I'm not sure why Zephyr pulls SOF as a module, TBH).
> > 
> > 
> > @kv2019i @nashif chicken/egg how do we solve ?
> 
> I think we need to do in two steps. E.g. let's merge this first (introducing SOF_DIV_ROUND_UP), second the SOF module can be updated in Zephyr, and third SOF can be moved to use Zephyr DIV_ROUND_UP directly.

I'm fine with any approach, rebased PR so it should be ready. Btw, I think the existing circular dependency you guys have should be fixed.

kv2019i (approved):
Thanks @gmarull !

kv2019i:
@gmarull wrote:
> I'm fine with any approach, rebased PR so it should be ready. Btw, I think the existing circular dependency you guys have should be fixed.

Ack. The dual use of SOF is indeed a bit complicated. SOF repository contains both a pipeline framework (that may be useful for other Zephyr applications), but also the SOF application (that should not live in Zephyr repo tree). Discussions are ongoing, but no conclusion yet.

lgirdwood (approved):

lgirdwood:
CI is stuck - rerun.

lgirdwood:
SOFCI TEST

kv2019i:
Checkpatch warning known, proceeding with merge.

PR #6671, opened 2022-11-25
zephyr: ace: power related small changes

tmleman:
* preventing power state change during ipc processing,
* waking-up secondary cores when forcing them to do transition into D3 state,
* adding logs inside power related functions.

### DIFF
--- a/src/ipc/ipc-zephyr.c
+++ b/src/ipc/ipc-zephyr.c
@@ -28,6 +28,8 @@
 #include <rtos/spinlock.h>
 #include <ipc/header.h>
 
+#include <zephyr/pm/policy.h>
+
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
@@ -69,6 +71,12 @@ static bool message_handler(const struct device *dev, void *arg, uint32_t data, 
 	increment_ipc_received_counter();
 #endif
 
+	/* Preventing d0i3 state during IPC processing in case if a host previously allowed for
+	 * power state transitions. Transition can occur only in the Idle thread and this lock acts
+	 * as a safeguard in case of the remaining threads to be non active
+	 * (e.g. waiting for a semaphore).
+	 */
+	pm_policy_state_lock_get(PM_STATE_RUNTIME_IDLE, PM_ALL_SUBSTATES);
 	ipc_schedule_process(ipc);
 
 	k_spin_unlock(&ipc->lock, key);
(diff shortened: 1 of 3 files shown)


src/platform/intel/ace/lib/pm_runtime.c
```diff
@@ -52,6 +54,8 @@ const struct pm_state_info *pm_policy_next_state(uint8_t cpu, int32_t ticks)
 			/* TODO: PM_STATE_RUNTIME_IDLE requires substates to be defined
 			 * to handle case with enabled PG andf disabled CG.
 			 */
+			tr_info(&power_tr, "transition to state %x (min_residency = %u, exit_latency = %u)",
```
btian1, line 57: Once confirmed this log functionality, could change it to tr_dbg? log print always take a lot of cycles, especially in core enabling and ipc message handling, even one line log could cause ipc timeout.

marcinszkudlinski (approved):

src/ipc/ipc-zephyr.c
```diff
@@ -69,6 +71,7 @@ static bool message_handler(const struct device *dev, void *arg, uint32_t data,
 	increment_ipc_received_counter();
 #endif
 
+	pm_policy_state_lock_get(PM_STATE_RUNTIME_IDLE, PM_ALL_SUBSTATES);
```
lyakh, line 74: could you explain how the power state transition was possible before this commit?

lgirdwood: ack - @tmleman I think we should spell this out as an inline comment so that no one is tempted to change PM state during IPC processing. Is suspect we can only change PM state when entering idle ?

lgirdwood (commented):
LGTM, Btw, I assume the PM driver spells out the conditions where PM state changes are allowed ?

src/platform/intel/ace/lib/pm_runtime.c
tmleman: ACK

abonislawski: Nice to have %#x for hex

abonislawski (approved):

lgirdwood (approved):

lgirdwood:
@tmleman can you check internal CI. Thanks !

tmleman:
> @tmleman can you check internal CI. Thanks !

@lgirdwood it was a problem with CI platform.

Issue #6669, opened 2022-11-25
[BUG] mtrace dropping log messages 


kv2019i:
**Describe the bug**
The mtrace logging backend loses log messages in fairly common use-cases. As the backend is based on use of a small SRAM window, it simply cannot handle a large bandwidth of messages, but performance should be sufficient with ERR/WARN/INFO level logs that are currently the default in SOF Zephyr builds.

**To Reproduce**
~/sof-test/test-case/multiple-pipeline.sh -f p -l 5

**Reproduction Rate**
Roughly 20% log messages lost in the above multiple-pipeline test.

**Expected behavior**
<1% lost rate in defaul config

**Impact**
Makes debugging more difficult using FW logs

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel: sof-dev c693d0caf6fa1b416d08d48eacbb5db96a8fac58
    * SOF: main 880857b3f7d144e8aaa3e01b5a271b028d59b56f
2) Name of the topology file
    * Topology: sof-tgl-nocodec, IPC4 test
3) Name of the platform(s) on which the bug is observed.
    * Platform: TGL cavs2.5

**Screenshots or console output**

kv2019i:
FYI to @andrula-song  @ranj063 as you've both reported this.

kv2019i:
With the linked two draft PRs (one to FW and one to kernel), I'm able to get log loss rate below 1% (from 20%) in the multi-pipeline test case.  Ideally we could increase the SRAM allocation a bit, but that is out of the cards for the moment.

lgirdwood:
> With the linked two draft PRs (one to FW and one to kernel), I'm able to get log loss rate below 1% (from 20%) in the multi-pipeline test case. Ideally we could increase the SRAM allocation a bit, but that is out of the cards for the moment.

Kconfig option ? we could encode some bits in the IPC flow control messages to warn about missing messages etc....

kv2019i:
With latest versions, only the FW change #6672 is needed. If this proves to work, I'll close the kernel PR for future use. We might add an opt-in debug mode to kernel to poll the sram window instead of using IPC messages, but for now it seems this is not needed.

PR #6672, opened 2022-11-25
ipc4: logging: mtrace: stop throttling if running out of space

kv2019i:
When firmware is executing a use-case sequence where system moves from:

  a) pipelines established and running with low rate of logs
     emitted

and moves to

  b) one or more pipelines change state and flurry of IPC
     activity occurs

The IPC notification throttling may be too slow to react, leading to many lost log messages that cover activity around system moving to (b) state.

Add logic to notice this situation and alert the host if available space in the SRAM log buffer gets low, even if throttling would otherwise block the IPC notification.

Link: https://github.com/thesofproject/sof/issues/6669

### DIFF
--- a/src/ipc/ipc4/logging.c
+++ b/src/ipc/ipc4/logging.c
@@ -42,12 +42,6 @@ LOG_MODULE_REGISTER(mtrace, CONFIG_SOF_LOG_LEVEL);
  */
 #define IPC4_MTRACE_AGING_TIMER_MIN_MS 100
 
-/**
- * Shortest time between IPC notifications sent to host. This
- * is used as protection against a flood of log messages.
- */
-#define IPC4_MTRACE_NOTIFY_MIN_DELTA_MS 10
-
 /* bb2aa22e-1ab6-4650-8501-6e67fcc04f4e */
 DECLARE_SOF_UUID("mtrace-task", mtrace_task_uuid, 0xbb2aa22e, 0x1ab6, 0x4650,
 		 0x85, 0x01, 0x6e, 0x67, 0xfc, 0xc0, 0x4f, 0x4e);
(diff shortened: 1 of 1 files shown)


ranj063:
@kv2019i Thank you so much for this. This PR makes mtrace so much more useful now!

src/ipc/ipc4/logging.c
```diff
@@ -78,7 +85,8 @@ static void mtrace_log_hook(size_t written, size_t space_left)
 
 	delta = k_uptime_get() - mtrace_notify_last_sent;
 
-	if (delta < IPC4_MTRACE_NOTIFY_MIN_DELTA_MS)
+	if (delta < IPC4_MTRACE_NOTIFY_MIN_DELTA_MS &&
+	    space_left > NOTIFY_BUFFER_CRIT_THRESHOLD)
```
btian1, line 89: ipc message cost more cycles? disble ipc message also cause log missing?

kv2019i: @btian1 Sending the IPC message is actually not that costly.

kv2019i:
V2 update:
- marking as not-draft
- simplified the approach -> remove the throttling altogether (with change on Zephyr side, this is not mandatory anymore, and it seems this is doing more harm than good, so better to just drop and save some DSP cycles while at it)
- let's see if the CI is happy.. there will be a small new burst of notify IPCs if logs are enabled in FW but host side is does not have a mtrace reader running, but this is a short transient

lgirdwood (commented):
Does this need to be a debug Kconfig i.e. when ```tr_dbg()``` level is enabled.
Thinking aloud, if logs are flooded during normal tr_level then are we sending too much ? and should we use ```tr_dbg()``` for more ?

kv2019i:
@lgirdwood I think this should be the default. Even after this PR, DSP won't send IPC notifications unless >50% of space is filled with logs. The host always reads everything out when it wakes up, so throttling will provide limited benefits.

This used to be more important as the Zephyr logging backend kept calling the hook function even when the buffer was full. This led to flood of IPC notifies in fairly common use-scenarios. This no longer happens, so throttling is not as important.

lyakh (approved):

abonislawski (approved):

Issue #6653, opened 2022-11-23
[BUG] failed to bind modules in multiple pipeline capture test with multi stream capture topology


bardliao:
**Describe the bug**
"failed to bind modules copier.module.8.2:1 -> copier.module.17.2:0" error will happen when I run "~/sof-test/test-case/multiple-pipeline.sh -f c -l 200 -w 1" test with SSP multi stream capture topology.

**To Reproduce**
 run "~/sof-test/test-case/multiple-pipeline.sh -f c -l 200 -w 1"

**Reproduction Rate**
\> 90%

**Expected behavior**
No error on the test

**Impact**
We will see CI test fail if https://github.com/thesofproject/sof/pull/6601 is merged

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel: topic/sof-dev branch 6992e1c067144d28e6461237aae84798898e6787 + https://github.com/thesofproject/linux/pull/4022
    * SOF: main branch e430629a1e9e4300bee7584c6b53c5f0412d5bb6 + https://github.com/thesofproject/sof/pull/6601
2) Name of the topology file
    * Topology: sof-tgl-nocodec.tplg
3) Name of the platform(s) on which the bug is observed.
    * Platform: TGL

**Screenshots or console output**
```
2022-11-23 09:32:29 UTC [INFO] ===== Testing: (Loop: 28/200) =====
2022-11-23 09:32:29 UTC [INFO] /home/test/sof-test/test-case/multiple-pipeline.sh will use topology /lib/firmware/intel/avs-tplg/sof-tgl-nocodec.tplg to run the test case
2022-11-23 09:32:29 UTC [INFO] Pipeline list to ignore is specified, will ignore 'pcm=HDA Digital' in test case
2022-11-23 09:32:29 UTC [INFO] Run command to get pipeline parameters
2022-11-23 09:32:29 UTC [COMMAND] sof-tplgreader.py /lib/firmware/intel/avs-tplg/sof-tgl-nocodec.tplg -f 'type:capture & ~pcm:Amplifier Reference' -b ' pcm:HDA Digital' -s 0 -e
2022-11-23 09:32:29 UTC [INFO] Testing: Port0 [hw:0,0]
2022-11-23 09:32:29 UTC [COMMAND] arecord   -D hw:0,0 -c 2 -r 48000 -f S16_LE /dev/null -q
[...]

PR #6674, opened 2022-11-28
ipc4: helper: disable interrupts during module binding

ranj063:
Disable interrupts until the intermediate buffer is connected to both the source and the sink components during module binding. This is needed to prevent IPC timeouts caused due to the IPC task getting preempted when the buffer is only connected to the source and the sink has not been connected yet.

Fixes https://github.com/thesofproject/sof/issues/6653

### DIFF
--- a/src/ipc/ipc4/helper.c
+++ b/src/ipc/ipc4/helper.c
@@ -359,23 +359,23 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 				  PPL_CONN_DIR_COMP_TO_BUFFER);
 	if (ret < 0) {
 		tr_err(&ipc_tr, "failed to connect src %d to internal buffer", src_id);
-		goto err;
+		goto free;
 	}
+
+	ret = comp_bind(source, bu);
+	if (ret < 0)
+		goto e_src_bind;
 
 	ret = comp_buffer_connect(sink, sink->ipc_config.core, buffer,
 				  PPL_CONN_DIR_BUFFER_TO_COMP);
 	if (ret < 0) {
 		tr_err(&ipc_tr, "failed to connect internal buffer to sink %d", sink_id);
-		goto err;
+		goto e_sink_connect;
 	}
-
-	ret = comp_bind(source, bu);
-	if (ret < 0)
-		return IPC4_INVALID_RESOURCE_ID;
 
 	ret = comp_bind(sink, bu);
 	if (ret < 0)
-		return IPC4_INVALID_RESOURCE_ID;
+		goto e_sink_bind;
 
 	/* update direction for sink component if it is not set already */
 	if (!sink->direction_set && source->direction_set) {
@@ -391,11 +391,18 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 
 	/* both sink and source components should have their direction set during module binding */
 	if (!sink->direction_set || !source->direction_set)
-		return IPC4_INVALID_RESOURCE_STATE;
+		goto err;
 
 	return IPC4_SUCCESS;
-
 err:
+	comp_unbind(sink, bu);
+e_sink_bind:
+	pipeline_disconnect(sink, buffer, PPL_CONN_DIR_BUFFER_TO_COMP);
+e_sink_connect:
+	comp_unbind(source, bu);
+e_src_bind:
+	pipeline_disconnect(source, buffer, PPL_CONN_DIR_COMP_TO_BUFFER);
+free:
 	buffer_free(buffer);
 	return IPC4_INVALID_RESOURCE_STATE;
 }
--- a/src/ipc/ipc4/helper.c
+++ b/src/ipc/ipc4/helper.c
@@ -419,7 +419,7 @@ int ipc_comp_disconnect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 	struct comp_dev *src, *sink;
 	struct list_item *sink_list;
 	uint32_t src_id, sink_id, buffer_id;
-	int ret;
+	int ret, ret1;
 
 	bu = (struct ipc4_module_bind_unbind *)_connect;
 	src_id = IPC4_COMP_ID(bu->primary.r.module_id, bu->primary.r.instance_id);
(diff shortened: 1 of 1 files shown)


src/ipc/ipc4/helper.c
```diff
@@ -366,6 +369,8 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 		goto err;
 	}
 
+	irq_local_enable(flags);
```
pblaszko, line 372: You need to enable interrupts back in case of `goto err;`

pblaszko (changes requested):

src/ipc/ipc4/helper.c
ranj063: Thanks @pblaszko fixed now. In fact I also added another patch to free the allocated buffer in case of all errors.

pblaszko: Thanks! Now looks good to me.

pblaszko (approved):

kv2019i (approved):

ranj063:
SOFCI TEST

src/ipc/ipc4/helper.c
```diff
@@ -372,12 +372,16 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
-		return IPC4_INVALID_RESOURCE_ID;
+	if (ret < 0) {
+		ret = IPC4_INVALID_RESOURCE_ID;
+		goto free;
+	}
 
 	ret = comp_bind(sink, bu);
-	if (ret < 0)
-		return IPC4_INVALID_RESOURCE_ID;
+	if (ret < 0) {
```
marcinszkudlinski, line 381: Still memory leak is possible - source has been bind, sink not, and procedure fail in the middle
not sure what happens (crash?) if buffer id freed when source bind was a success

src/ipc/ipc4/helper.c
marcinszkudlinski, line 372: Please comment it why this is a critical section
and
> **at the moment** SOF only supports running all connected pipelines on the same core.

it will change soon enough, IPC task will always be binded to core 0, pipelines will work on all cores, what then? 
do we really want to do such temporary changes?

marcinszkudlinski (changes requested):
@mwasko @mmaka1 please take a look at this

src/ipc/ipc4/helper.c
lyakh: Firstly we probably want that comment about why we're locking here in-line, not just as a commit message. Secondly, @marcinszkudlinski I'm not sure we need a spinlock here. It's good to prepare for the future, but we don't know yet what it will be like. I think when we'll have to run connected pipelines or components on different cores, we'll have to protect the buffer list header throughout the code, and that will be **a lot** of protecting. And then most (or all) of that code runs in threads, so we might decide to use mutexes instead of spinlocks.
Another possibility to fix this could be finding the code that crashes when the buffer isn't fully connected yet, and fixing it to ignore such half-connected buffers.

src/ipc/ipc4/helper.c
ranj063: > FWIW I don't see a single `.bind()` implementation. Maybe time to remove it

@lyakh the mixinmixout uses the bind/unbind. I am about to remove it but not there yet

src/ipc/ipc4/helper.c
ranj063: > Another possibility to fix this could be finding the code that crashes when the buffer isn't fully connected yet, and fixing it to ignore such half-connected buffers.

I thought about this but then decided that the solution has to be generic to allow module developers to not worry about it.

lyakh: > Please comment it why this is a critical section and
> 
> > **at the moment** SOF only supports running all connected pipelines on the same core.
> 
> it will change soon enough, IPC task will always be binded to core 0, pipelines will work on all cores, what then? do we really want to do such temporary changes?

@marcinszkudlinski I think this isn't quite correct. Core 0 will manage the IPC queue, but IPC processing will be done on respective cores, just as it is done now

src/ipc/ipc4/helper.c
lyakh: @ranj063 you're right, thanks, I misinterpreted my grep results

src/ipc/ipc4/helper.c
ranj063: > you think it would be a specific module, not some common code? Then maybe make it common... Just another option

@lyakh my original suspicion was the copier module and still is. But im not able to pinpoint exactly what in the copier module code is crashing to cause this problem. I thought it was likely the sink dev being NULL but even that doesnt help

ranj063: @lyakh @marcinszkudlinski thinking a bit more about the issue of connected pipelines being on different cores, I think this whole function ipc_comp_connect() would need some surgery to support that isnt it? This function the way it is written can only really work is both sides of the buffer are on the same core. If thats not the case, each connection would have to be handled on the core that it belongs to and then there's the question on which core to allocate the buffer itself. So, I think its too premature to talk about that at this point.

ranj063:
@lyakh @marcinszkudlinski I've update to address both your comments to fix the error paths now.

src/ipc/ipc4/helper.c
```diff
@@ -401,13 +406,20 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
+	}
 
 	return IPC4_SUCCESS;
 
 err:
 	irq_local_enable(flags);
+free:
+	pipeline_disconnect(source, buffer, PPL_CONN_DIR_COMP_TO_BUFFER);
+	pipeline_disconnect(sink, buffer, PPL_CONN_DIR_BUFFER_TO_COMP);
 	buffer_free(buffer);
```
lyakh, line 423: can we make it
```
e_sink:
	comp_unbind(sink, bu);
e_source:
	comp_unbind(source, bu);
	pipeline_disconnect(source, buffer, PPL_CONN_DIR_BUFFER_TO_COMP);
free:
	buffer_free(buffer);
```
and then jump to the correct label above?

ranj063: @lyakh I can't quite do that but i've squashed the 2 commits now to perform both connect and bind with the IRQ disabled and jump to the right label in case of errors. Let me know how it looks

marcinszkudlinski (approved):

PR #6688, opened 2022-11-29
Audio: SRC: Increase stage buffer size and change copy() run condition

singalsu:
The main changes are commits 1 and 2 to SRC code. First prevents a pipeline freeze. The second fixes an overflow with 16 bit audio data.

The rest of changes are improvements to src_test.m script. E.g. call testbench with copy limit of 5 min to stop frozen SRC and minor issues fixes and speedup with reduce of printing and plot window opening.

### DIFF
--- a/tools/test/audio/src_test.m
+++ b/tools/test/audio/src_test.m
@@ -465,7 +465,7 @@ end
 
 function src_test_result_print(t, testverbose, testacronym, ph)
 tstr = sprintf('%s SRC %d, %d', testverbose, t.fs1, t.fs2);
-if nargin > 3
+if nargin > 3 && ~isempty(ph)
         title(ph, tstr);
 else
         title(tstr);
--- a/tools/test/audio/src_test.m
+++ b/tools/test/audio/src_test.m
@@ -456,6 +456,7 @@ end
 function test = test_run_src(test, t)
 test.fs_in = test.fs1;
 test.fs_out = test.fs2;
+test.extra_opts = '-C 300000'; % Limit to 5 min max, assume 1 ms scheduling
 delete_check(1, test.fn_out);
 test = test_run(test);
 end
--- a/tools/test/audio/src_test.m
+++ b/tools/test/audio/src_test.m
@@ -7,8 +7,8 @@ function [n_fail, n_pass, n_na] = src_test(bits_in, bits_out, fs_in_list, fs_out
 %
 % bits_in    - input word length
 % bits_out   - output word length
-% fs_in      - vector of rates in
-% fs_out     - vector of rates out
+% fs_in      - vector of rates in, default 8 to 192 kHz
+% fs_out     - vector of rates out, default 8 to 192 kHz
 % full_test  - set to 0 for chirp only, 1 for all, default 1
 % show_plots - set to 1 to see plots, default 0
 %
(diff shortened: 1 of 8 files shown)


tools/test/topology/test-playback.m4
```diff
@@ -152,8 +152,8 @@ SectionGraph."pipe-sof-fourth-pipe" {
 
 # DAI configuration
 
-# Use 3 periods for SRC DAI buffer, otherwise 2 periods
-ifelse(TEST_PIPE_NAME, `src', `define(TEST_DAI_PERIODS, `3')', `define(TEST_DAI_PERIODS, `2')')
+# Use 4 periods for SRC DAI buffer, otherwise 2 periods
```
singalsu, line 155: These are TBD, maybe 3 & 3 could work. I'll test. I did this before noticing the stage1 and stage2 conditions.

singalsu: Yes, the existing 3 source and 3 sink buffers were sufficient. I removed this topology commit.

tools/test/audio/src_test.m
```diff
@@ -413,7 +413,9 @@
 %% Analyze
 test.fs = t.fs2;
 test = chirp_test_analyze(test);
-src_test_result_print(t, 'Chirp', 'chirpf');
+if ~test.fail
```
singalsu, line 416: Mistake here, the fail is -1 if the chirp file doesn't exist. Need to plot & store a failed chirp spectra.

singalsu: fixed now

lgirdwood (approved):
Approve the C part.

gongjun-song:
This PR indeed fix hang issue, but I tested src_test.sh on the default SOF and it didn't pass all the test cases.
```
Number of passed tests = 864
Number of failed tests = 2
Number of non-applicable tests = 156
Number of skipped tests = 0

ERROR: TEST FAILED!!!
Error: SRC test with 16 bits data failed.
Test failed.
```

singalsu:
> This PR indeed fix hang issue, but I tested src_test.sh on the default SOF and it didn't pass all the test cases.

Thanks for report! I ran the 32 bit test. There is likely some pass criteria that the 16 bit version cannot pass, I'll try to address it. Please use meanwhile 24 or 32 bit tests.

gongjun-song:
> > This PR indeed fix hang issue, but I tested src_test.sh on the default SOF and it didn't pass all the test cases.
> 
> Thanks for report! I ran the 32 bit test. There is likely some pass criteria that the 16 bit version cannot pass, I'll try to address it. Please use meanwhile 24 or 32 bit tests.

I disabled 16 bit and only test 24bit and 32bit, all passed.
```
Number of passed tests = 866
Number of failed tests = 0
Number of non-applicable tests = 156
Number of skipped tests = 0

Test passed.
Success: SRC test with 24 and 32 bits data passed.
Test passed.
```

singalsu:
@gongjun-song @lgirdwood The 16 bit issue was a real bug, good catch! I added as 2nd commit the bug fix to polyphase filter where the saturation was missing for 16 bit audio data round.

gongjun-song:
New commits have fixed 16-bit issue, all passed!
```
Number of passed tests = 866
Number of failed tests = 0
Number of non-applicable tests = 156
Number of skipped tests = 0

Test passed.
Success: SRC test with 16, 24 and 32 bits data passed.
Test passed.
```

lgirdwood (approved):
@gongjun-song thanks for the test results.

PR #6705, opened 2022-12-01
tools: add nhlt acpi table parsing tool

juimonen:
Add tool to parse nhlt acpi table. Tool can be used to parse blobs generated by alsatplg or even blobs from bios. It will print out nhlt endpoint headers and ssp and dmic blob internals.

example usage:
./sof-nhlt -i nhlt.bin -d 1 (use dmic hw 1 i.e. all prior to mtl)

./sof-nhlt -i nhlt.bin -d 3 (use dmic hw 3 i.e. mtl onwards)


tools/nhlt/nhlt.h
```diff
@@ -0,0 +1,132 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright(c) 2022 Intel Corporation. All rights reserved.
+ *
+ * Author: Jaska Uimonen <jaska.uimonen@linux.intel.com>
+ */
```
marc-hb, line 6: Please add a link to either the spec or the reference code

marc-hb (commented):
A parser in... bare C? In 2022?

 :-(


https://pypi.org/project/construct/, plenty others

marc-hb:
> A (user space) parser in... bare C? In 2022?

So this is what happens with C = the language of memory corruption. No need to look far at all:
- https://github.com/thesofproject/rimage/issues/115
- https://github.com/thesofproject/rimage/issues/80
- https://github.com/thesofproject/rimage/pull/28
- https://github.com/thesofproject/rimage/pull/27

Etc.

https://alexgaynor.net/2019/aug/12/introduction-to-memory-unsafety-for-vps-of-engineering/

juimonen:
@marc-hb yeah this is oldies/goldies from my hard drive, which has gained quite big popular demand with topology2 blobs. Tired of sending tar balls to people :) but yeah would be much better to do do in python etc. just don't have time right now.

marc-hb:
> yeah would be much better to do do in python etc. just don't have time right now.

I took a few hours, learned Construct and in 100 lines I have "round-trip" library that can not just pretty-print but also interactively edit any field and write back. Demo at:

https://github.com/marc-hb/sof/tree/NHLT-construct/tools/NHLT

There is practically no code, it's almost entirely declaring the structs and that's it.

Disclaimer: it's only parsing the standard NHLT data for now, not the DMIC blobs yet. That should not take more than a couple hundreds more lines.

marc-hb:
@miRoox had already and successfully used Construct for https://github.com/thesofproject/sof-test/blob/main/tools/tplgtool2.py. Thanks for pioneering this!

lgirdwood:
Why is the python decoder in sof-test and not in sof, isn't it a developer tool ?

lgirdwood:
> > Why is the python decoder in sof-test and not in sof, isn't it a developer tool ?
> 
> The first use case was creating topology graphs in test results. I guess that's why.
> 
> Does it matter that much? Are there any (bad!) people cloning sof but not sof-test? :-)

This should be moved out of sof-test into the new tools repo when its' created.

juimonen:
@aborisovich I will definitely consider this as an insult, but as you wish.

plbossart:
> We need to change the mindset and approach.

And you need to start working on your comments @aborisovich. It's not the first time and it's not going to make you become more respectable in an open-source public development.

PR #6716, opened 2022-12-05
math: Fix warnings about uninitialized variables

piotrkmx:
Minor changes to avoid warnings about uninitialized variables given by Klocwork.

This changes may slightly affect performance. Please reconsider if we really need them.

Signed-off-by: Piotr Kmiecik <piotrx.kmiecik@intel.com>

### DIFF
--- a/src/math/fir_hifi3.c
+++ b/src/math/fir_hifi3.c
@@ -95,9 +95,9 @@ void fir_32x16_hifi3(struct fir_state_32x16 *fir, ae_int32 x, ae_int32 *y,
 	ae_f64 a;
 	ae_valign u;
 	ae_f32x2 data2;
-	ae_f16x4 coefs;
-	ae_f32x2 d0;
-	ae_f32x2 d1;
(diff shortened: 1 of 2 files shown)


src/math/fir_hifi3.c
```diff
@@ -95,9 +95,9 @@ void fir_32x16_hifi3(struct fir_state_32x16 *fir, ae_int32 x, ae_int32 *y,
 	ae_f64 a;
 	ae_valign u;
 	ae_f32x2 data2;
-	ae_f16x4 coefs;
-	ae_f32x2 d0;
-	ae_f32x2 d1;
+	ae_f16x4 coefs = AE_ZERO16(); /* Note: Init is not needed */
```
marcinszkudlinski, line 98: don;t, set as not an issue in static analyser tool

marcinszkudlinski (approved):

src/math/fir_hifi3.c
marcinszkudlinski: damn should be reject

PR #6719, opened 2022-12-05
ipc4: Fix error handling issues in ipc4

jxstelter:
This patch fixes issues in error handling flows.

### DIFF
--- a/src/ipc/ipc4/handler.c
+++ b/src/ipc/ipc4/handler.c
@@ -290,6 +290,9 @@ static int set_pipeline_state(uint32_t id, uint32_t cmd, bool *delayed)
 			}
 			if (ret == PPL_STATUS_SCHEDULED)
 				*delayed = true;
+			break;
+		default:
+			ret = 0;
 		}
 
 		/*


src/ipc/ipc4/helper.c
```diff
@@ -627,6 +627,11 @@ int ipc4_create_chain_dma(struct ipc *ipc, struct ipc4_chain_dma *cdma)
 
 	ipc_pipe = ipc_get_comp_by_id(ipc, pipeline_id);
 
+	if (!ipc_pipe) {
```
ranj063, line 630: this too is not correct. The pipeline is create just above. If it wasnt created successfully, we'd never get here right?

src/ipc/ipc4/handler.c
```diff
@@ -215,7 +215,7 @@ static int set_pipeline_state(uint32_t id, uint32_t cmd, bool *delayed)
 	struct ipc_comp_dev *host = NULL;
 	struct ipc *ipc = ipc_get();
 	int status;
-	int ret;
+	int ret = 0;
```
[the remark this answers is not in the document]
jxstelter: I think you can return in line #306 for case SOF_IPC4_PIPELINE_STATE_RESET. If status is not in (COMP_STATE_INIT, COMP_STATE_READY, COMP_STATE_ACTIVE or COMP_STATE_PAUSED) and *delayed  is set. Don't know if this is real scenario.

src/ipc/ipc4/helper.c
jxstelter: Yes, I missed the call inside ipc4_create_pipeline() function. Reverting back.

src/ipc/ipc4/handler.c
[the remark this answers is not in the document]
ranj063: Even with the return on line 306, I think ret would be set in every switch/case just above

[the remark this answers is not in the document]
ranj063: @jxstelter i think a better fix would be to add a default cae in the switch above.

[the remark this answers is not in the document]
jxstelter: fixed

pblaszko (approved):

wszypelt:
@jxstelter Request to ignore the results for the internal intel CI test, changes are currently being made, tests will be repeated in 2 hours

ranj063 (approved):

src/ipc/ipc4/handler.c
```diff
@@ -290,6 +290,9 @@ static int set_pipeline_state(uint32_t id, uint32_t cmd, bool *delayed)
 			}
 			if (ret == PPL_STATUS_SCHEDULED)
 				*delayed = true;
+			break;
+		default:
+			ret = 0;
```
marcinszkudlinski, line 295: sure that default is "0 == SUCCESS"? in case a state has not been found maybe better to return some error code?

jxstelter: I wanted to keep identical behavior. Don't know whether not hitting any of these statuses is wrong since we have more states defined than these under switch.

lgirdwood: I think we should have a IPC4_SUCESS value here ? and maybe it's 0. Btw, do we need a break or fallthough here ?
```diff
diff --git a/src/ipc/ipc4/handler.c b/src/ipc/ipc4/handler.c
index 35f9361ac..2a4baf4b5 100644
--- a/src/ipc/ipc4/handler.c
+++ b/src/ipc/ipc4/handler.c
@@ -309,6 +309,7 @@ static int set_pipeline_state(struct ipc_comp_dev *ppl_icd, uint32_t cmd,
        case SOF_IPC4_PIPELINE_STATE_EOS:
                if (status != COMP_STATE_ACTIVE)
                        return IPC4_INVALID_REQUEST;
+               /* fallthrough */
        case SOF_IPC4_PIPELINE_STATE_SAVED:
        case SOF_IPC4_PIPELINE_STATE_ERROR_STOP:
        default:
```

marcinszkudlinski: break or fallthrough, 0 or IPC4_SUCCESS - just a code style. 

My concern is if it should not be like:
``` 
                case OTHER_ALLOWED_STATE1:
                case OTHER_ALLOWED_STATE2:
                       ret = 0;
                       break;
		default:
			tr_err(&ipc_tr, "ipc: current status %d", status);
			return IPC4_INVALID_REQUEST;
		}
```

lgirdwood: ok, @jxstelter can we look at the fallthrough/break update in a new PR.

lgirdwood:
@marcinszkudlinski @jxstelter lots of build failures.

PR #6721, opened 2022-12-06
topology2: cavs-rt5682: add deep buffer support on headsets

yongzhi1:
Deep buffer is mixed with ssp1 stream.
Override mixout copier audio_format.1 from 32/24 to 32/32 for I2S.

### DIFF
--- a/tools/topology/topology2/cavs-rt5682.conf
+++ b/tools/topology/topology2/cavs-rt5682.conf
@@ -15,6 +15,7 @@
 <passthrough-be.conf>
 <host-copier-gain-mixin-playback.conf>
 <mixout-gain-dai-copier-playback.conf>
+<deepbuffer-playback.conf>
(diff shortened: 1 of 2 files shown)


yongzhi1:
> @yongzhi1 what's the point of trying to add a deep-buffer for amplifiers? we are not going to reduce power consumption much, are we?

Yes, I agree. It's easier to test remotely w/o jack insertion right now, will move deep-buffer to headsets later.

plbossart:
> Yes, I agree. It's easier to test remotely w/o jack insertion right now, will move deep-buffer to headsets later.

Ah, that makes sense indeed if the default is speakers. Good suggestion. 

In theory once we have the split host- and dai parts with a mixer on each dai pipeline it's straightforward to change the topology.

lgirdwood (approved):

PR #6722, opened 2022-12-06
[SKIP CI] sparse: temporarily ignore MTL address errors and add to daily tests

marc-hb:
6 commits, see messages

### DIFF
--- a/scripts/parse_sparse_output.sh
+++ b/scripts/parse_sparse_output.sh
@@ -27,7 +27,10 @@ main()
 
     sparse_errors+=(-e '[[:space:]]error:[[:space:]]')
 
-    sparse_errors+=(-e '[[:space:]]warning:[[:space:]].*different address space')
+    # TODO: fix mtl source code
+    if test 'mtl' != "$platform"; then
+        sparse_errors+=(-e '[[:space:]]warning:[[:space:]].*different address space')
+    fi
 
     ! grep -v 'alsatplg.*topology2.*skip' | grep -i  "${sparse_errors[@]}"
 
--- a/.github/workflows/daily-tests.yml
+++ b/.github/workflows/daily-tests.yml
@@ -18,3 +18,5 @@ jobs:
     uses: ./.github/workflows/pull-request.yml
   run-zephyr-builds:
     uses: ./.github/workflows/zephyr.yml
+  sparse-zephyr:
+    uses: ./.github/workflows/sparse-zephyr.yml


marc-hb:
MTL warnings are still visible in https://github.com/thesofproject/sof/actions/runs/3625503925/jobs/6113619508 but they don't "break the build" any more.

https://github.com/thesofproject/sof/actions/runs/3625503919/jobs/6113620156 has the stupid "we don't talk about sparse" checkpatch warning and nothing else.

scripts/parse_sparse_output.sh
```diff
@@ -27,7 +27,10 @@ main()
 
     sparse_errors+=(-e '[[:space:]]error:[[:space:]]')
 
-    sparse_errors+=(-e '[[:space:]]warning:[[:space:]].*different address space')
+    # TODO: fix mtl source code
+    if test 'mtl' != "$platform"; then
+        sparse_errors+=(-e '[[:space:]]warning:[[:space:]].*different address space')
+    fi
 
     ! grep -v 'alsatplg.*topology2.*skip' | grep -i  "${sparse_errors[@]}"
```
lyakh, line 36: this means, we will see complete MTL sparse output including "address space" warnings in CI, it just won't be a failure? Could you maybe remove "SKIP CI" to let us see what it will look like?

lyakh: Indeed, thanks! I'm not sure whether it's a good idea to green-wash our failures, but as long as we have the output, I personally can live with it

marc-hb: I don't like it either but it's been red for too long now. I also made sure it's very easy to revert.

Can you approve then?

marc-hb: @lgirdwood re-submitting the other patches is not an issue but does anyone have any idea how to fix the current warnings on MTL and a line of sight on a fix? Right now we are actively training users to ignore the results on MTL and we are blind to possibly WORSE MTL regressions.

lgirdwood: Will be fixed after v2.4 is tagged, IOW early next week.

marc-hb:
I resubmitted the cleanup commits that don't change anything in this new PR:
- #6769

> I'm not sure whether it's a good idea to green-wash our failures

I am sure it's a bad idea to "red-wash" our sparse warnings on MTL, which is what we have been doing for a few weeks now. In other words: if the sparse analysis degrades from "we have warnings" to "it does not even build" then no one will notice. "Red-washed".

This of course on top of training people to ignore warnings in the first place.

lgirdwood:
@lyakh how are the SPARSE MTL fixes in Zephyr coming ?

marc-hb:
I believe all sparse warnings are fixed in the Zephyr branch but upgrading Zephyr is blocked by https://github.com/zephyrproject-rtos/zephyr/pull/53327#issuecomment-1415298902

marc-hb:
Actually let's keep this for adding to daily tests

marc-hb:
address space warnings are all fixed now. Adding sparse to daily tests in new
- #7203

marc-hb:
>  Right now we are actively training users to ignore the results on MTL and we are blind to possibly WORSE MTL regressions.

There was an address space regression and no one noticed:
https://github.com/thesofproject/sof/pull/7089#issuecomment-1468873034

Issue #6679, opened 2022-11-28
[BUG] MOD_DX IPC timeout 


ranj063:
**Describe the bug**
The MOD_DX IPC times out when entering D3

**To Reproduce**
check-suspend-resume-with-audio.sh -l 1 -m capture

**Reproduction Rate**
TGL: Very low < 1-5%. Only seen it a handful of times so far.
MTL: about 50%

**Expected behavior**
MOD_DX IPC should always send a reply back to the host.

3) Name of the platform(s) on which the bug is observed.
    * Platform: TGL, MTL

https://sof-ci.01.org/sofpr/PR6672/build2523/devicetest/index.html?model=TGLU_RVP_NOCODEC_IPC4ZPH&testcase=check-suspend-resume-with-capture

miRoox:
This issue is also found in MTL platforms, and the reproduction rate seems higher than TGL.

---

Found in test report 18581:
Kernel Commit: 
[321028577de9](https://github.com/thesofproject/linux/commit/321028577de9)
SOF Commit: 
[4a9a99aeb24a](https://github.com/thesofproject/sof/commit/4a9a99aeb24a)

[dmesg.txt](https://github.com/thesofproject/sof/files/10162059/dmesg.txt)

ranj063:
even on TGL check-suspend-resume-with-audio.sh -l 20 -m capture fails 100% of the time

keqiaozhang:
This issue is not reproducible in CI. Closing this bug.

PR #6440, opened 2022-10-18
soc: ace: fix IPC D3 power state entry

aborisovich:
**Fist commit**
Patch fixes D3 issue - now platform enters and wakes up from D3 correctly. 
It had been achieved using Zephyr cpu function call that redirects D3 flow to power down assembly code
sending IPC response to host just before powering down.

~**Second commit**~
~Adds pm notifiers to Zephyr power manager to allocate IMR memory inside SOF code in order to save LPSRAM and HPSRAM memory before powering down.~
Edit: dropped for now due to missing implementation of devices context restore.
https://github.com/zephyrproject-rtos/zephyr/pull/53105 will be used instead to skip context restore for now.
It will be implemented soon<sup>tm</sup>.

Requires:

- [x] https://github.com/thesofproject/sof/pull/6360
- [x] https://github.com/thesofproject/sof/pull/6041
- [x] https://github.com/zephyrproject-rtos/zephyr/pull/51414
- [x] https://github.com/thesofproject/sof/pull/6489
- [x] https://github.com/zephyrproject-rtos/zephyr/pull/47840
- [x] https://github.com/thesofproject/sof/pull/6613
- [x] https://github.com/thesofproject/sof/pull/6697
- [x] https://github.com/zephyrproject-rtos/zephyr/pull/53105
- [x] https://github.com/thesofproject/sof/pull/6828

Signed-off-by: Andrey Borisovich <andrey.borisovich@intel.com>
Signed-off-by: Marcin Szkudlinski <marcin.szkudlinski@intel.com>
Signed-off-by: Tomasz Leman <tomasz.m.leman@intel.com>

PR #6725, opened 2022-12-06
Fix D3 sequence for TGL

ranj063:
This makes the D3 sequence for IPC3 and IPC4 identical for TGL and fixes https://github.com/thesofproject/sof/issues/6679 on TGL

### DIFF
--- a/src/ipc/ipc4/handler.c
+++ b/src/ipc/ipc4/handler.c
@@ -878,6 +878,10 @@ static int ipc4_module_process_dx(struct ipc4_message_request *ipc4)
 		/* do platform specific suspending */
 		platform_context_save(sof_get());
 
+#if !defined(CONFIG_LIBRARY) && defined(CONFIG_CAVS)
+		arch_irq_lock();
+		platform_timer_stop(timer_get());
+#endif
 		ipc_get()->pm_prepare_D3 = 1;
 	}
 


miRoox:
This patch seems not helpful for MTL, https://github.com/thesofproject/sof/issues/6679 is still reproducible with it on MTL.

src/ipc/ipc4/handler.c
```diff
@@ -878,6 +878,10 @@ static int ipc4_module_process_dx(struct ipc4_message_request *ipc4)
 		/* do platform specific suspending */
 		platform_context_save(sof_get());
 
+#if !defined(CONFIG_LIBRARY)
+		arch_irq_lock();
+		platform_timer_stop(timer_get());
+#endif
 		ipc_get()->pm_prepare_D3 = 1;
```
tmleman, line 885:
```c
#if !defined(CONFIG_LIBRARY) && !defined(CONFIG_PM)
```

On MTL  function **_platform_timer_stop_** does nothing from what I've seen.
Interrupts will be disabled later by the PM.

tmleman (changes requested):
[PR6440](https://github.com/thesofproject/sof/pull/6440) this patch is required to test D3 on MTL at all.

tmleman:
@aborisovich @marcinszkudlinski fyi

src/ipc/ipc4/handler.c
ranj063: @tmleman may I ask why you are suggesting !defined(CONFIG_PM)? Did you mean defined(CONFIG_PM) instead? Would defined(CONFIG_CAVS) be a better option for TGL?

tmleman: **!defined(CONFIG_PM)** and **defined(CONFIG_CAVS)** will have the same result. In my opinion, this code should not go to MTL. It will also probably not be needed once CAVS platforms start using the Zephyr power manager.

**defined(CONFIG_CAVS)** is fine for me too.

tmleman (approved):

lgirdwood (approved):

PR #6728, opened 2022-12-06
module_adapter: cadence: Treat nonfatal errors as nonfatal

paulstelian97:
The processing function can at times return a nonfatal error. Such errors should not be treated as hard errors.

This commit changes it so that they are instead treated as a warning.

Signed-off-by: Paul Olaru <paul.olaru@nxp.com>

Draft because of formatting issues and because I did not change this in all places, just in one of them.


src/audio/module_adapter/module/cadence.c
```diff
@@ -632,11 +632,15 @@ cadence_codec_process(struct processing_module *mod,
 	}
 
 	API_CALL(cd, XA_API_CMD_EXECUTE, XA_CMD_TYPE_DO_EXECUTE, NULL, ret);
-	if (ret != LIB_NO_ERROR) {
+	//if (ret != LIB_NO_ERROR) {
```
iuliana-prodan, line 635: You might add a define and use it where necessary, like:
```
#define XA_CHECK_ERROR(dev, err)
({
    if (err != LIB_NO_ERROR)
    {
        if (XA_ERROR_SEVERITY(err))
        {
            comp_err(dev, "error: %x), err);
            return err;
        }
        comp_warn(dev, "warning: %x), err);
    }
    err;
})
```

paulstelian97: I do like this idea, thanks. Will do in the next submission.

lyakh, line 635: please, remove commented out code

paulstelian97: Will do, it was one of the formatting issues which made me make this a draft.

paulstelian97:
Version 2, still has formatting issues but has refactored and added a macro for the error checking.

src/audio/module_adapter/module/cadence.c
```diff
@@ -21,6 +21,19 @@ DECLARE_SOF_RT_UUID("cadence_codec", cadence_uuid, 0xd8218443, 0x5ff3, 0x4a4c,
 
 DECLARE_TR_CTX(cadence_tr, SOF_UUID(cadence_uuid), LOG_LEVEL_INFO);
 
+/* Define a macro for error checking from Cadence API functions */
+
+#define XA_CHECK_ERROR(dev, err, fmt_extra, ret_statement, ...) ({ \
+	if (err != LIB_NO_ERROR) { \
+		if (XA_ERROR_SEVERITY(err)) {\
+			comp_err(dev, "Cadence API error %x" fmt_extra, err, ##__VA_ARGS__); \
+			ret_statement; \
```
dbaluta, line 30: I don't understand how this works. 

It looks like you replaced
```
#define LIB_IS_FATAL_ERROR(e) ((e) & XA_FATAL_ERROR)
```
with:

```
#define XA_ERROR_SEVERITY(code) (((code) & XA_FATAL_ERROR) != 0)
```

iuliana-prodan: LIB_IS_FATAL_ERROR and XA_ERROR_SEVERITY are similar.
We should use XA_ERROR_SEVERITY - this is from Cadence API - https://github.com/thesofproject/sof/blob/main/src/include/sof/audio/cadence/xa_error_standards.h#L56

> Also, I don't really like XA_CHECK_ERROR .

Can you elaborate on this?

miRoox:
SOFCI TEST

src/audio/module_adapter/module/cadence.c
paulstelian97: I can use LIB_IS_FATAL_ERROR instead of XA_ERROR_SEVERITY, in the end it doesn't truly matter so I can do that before my next push.

dbaluta: Use can use XA_ERROR_SEVERITY instead of 

> LIB_IS_FATAL_ERROR and XA_ERROR_SEVERITY are similar. We should use XA_ERROR_SEVERITY - this is from Cadence API - https://github.com/thesofproject/sof/blob/main/src/include/sof/audio/cadence/xa_error_standards.h#L56
> 
> > Also, I don't really like XA_CHECK_ERROR .
> 
> Can you elaborate on this?

I can agree we can use XA_ERROR_SEVERITY instead of LIB_IS_FATAL_ERROR but the name XA_ERROR_SEVERITY isn't really very useful. You don't know exactly what it does by only just looking at the name.

With LIB_IS_FATAL_ERROR you know you are testing for a FATAL error.

As for XA_CHECK_ERROR. I don't like it because it can just get just "return" from the current function and you don't really expect macros to do that. 

Just look at the call:

```XA_CHECK_ERROR(dev, ret, " while applying parameter %d with value %d", return ret, param->id, *(int32_t *)param->data);
```

It does have a 'return ret' as a parameter.

paulstelian97: I guess having manual checks instead (e.g. if IS_FATAL_ERROR <do something> else if is an error <do something else>) as in having this macro unwound is an idea?

I guess I'll remove commented out code, try to change this and open this up for wider review.

paulstelian97:
Version 3: Simplified and removed the new macro which had nonlocal flow control. Now the fix is strictly for the original issue, and no refactor is being done anymore.

paulstelian97:
Version 4: Refactored previous patch to be more in line with preexisting code in the same file. No functional change.

dbaluta (approved):

lgirdwood (approved):

lgirdwood:
@wszypelt not expecting this to impact CI (as I dont think we test Cadence Libs atm). Good to merge ?

wszypelt:
@lgirdwood Multi-Core support failed in tests, give me 20 min to check this again.

wszypelt:
@lgirdwood @paulstelian97 
It seems that we have a problem with dut and some tests fail, I'm re-testing internal intel CI

PR #6730, opened 2022-12-06
Cache: fix allocation and address prefetch

lyakh:
1. Objects, handled using the coherent API, must be allocated as cached.
2. Make sure data-cache prefetch doesn't race against buffer linking and unlinking

### DIFF
--- a/src/audio/buffer.c
+++ b/src/audio/buffer.c
@@ -7,6 +7,7 @@
 
 #include <sof/audio/buffer.h>
 #include <sof/audio/component.h>
+#include <sof/common.h>
 #include <rtos/interrupt.h>
 #include <rtos/alloc.h>
 #include <rtos/cache.h>
@@ -40,9 +41,12 @@ struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t align)
 		return NULL;
 	}
 
-	/* allocate new buffer */
+	/*
+	 * allocate new buffer, align the allocation size to a cache line for
+	 * the coherent API
+	 */
 	buffer = rzalloc(SOF_MEM_ZONE_RUNTIME_SHARED, 0, SOF_MEM_CAPS_RAM,
-			 sizeof(*buffer));
+			 ALIGN_UP(sizeof(*buffer), PLATFORM_DCACHE_ALIGN));
 	if (!buffer) {
 		tr_err(&buffer_tr, "buffer_alloc(): could not alloc structure");
 		return NULL;
--- a/src/audio/mixin_mixout/mixin_mixout.c
+++ b/src/audio/mixin_mixout/mixin_mixout.c
@@ -198,8 +198,10 @@ static struct comp_dev *mixout_new(const struct comp_driver *drv,
 
 	memcpy_s(&md->base_cfg, sizeof(md->base_cfg), spec, sizeof(md->base_cfg));
 
+	/* align the allocation size to a cache line for the coherent API */
 	md->mixed_data_info = rzalloc(SOF_MEM_ZONE_RUNTIME_SHARED, 0, SOF_MEM_CAPS_RAM,
-				      sizeof(struct mixed_data_info));
+				      ALIGN_UP(sizeof(struct mixed_data_info),
+					       PLATFORM_DCACHE_ALIGN));
 	if (!md->mixed_data_info) {
 		rfree(md);
 		rfree(dev);
--- a/src/audio/pipeline/pipeline-graph.c
+++ b/src/audio/pipeline/pipeline-graph.c
@@ -185,6 +185,8 @@ int pipeline_connect(struct comp_dev *comp, struct comp_buffer *buffer,
 		     int dir)
 {
 	struct list_item *comp_list;
+	struct list_item __sparse_cache *needs_sync;
+	bool further_buffers_exist;
 	uint32_t flags;
 
 	if (dir == PPL_CONN_DIR_COMP_TO_BUFFER)
@@ -201,10 +203,14 @@ int pipeline_connect(struct comp_dev *comp, struct comp_buffer *buffer,
 	 * overwriting the modified header. FIXME: this is still a problem with
 	 * different cores.
 	 */
-	if (!list_is_empty(comp_list))
-		dcache_writeback_invalidate_region(uncache_to_cache(comp_list->next),
-						   sizeof(struct list_item));
+	further_buffers_exist = !list_is_empty(comp_list);
+	needs_sync = uncache_to_cache(comp_list->next);
+	if (further_buffers_exist)
+		dcache_writeback_region(needs_sync, sizeof(struct list_item));
+	/* The cache line can be prefetched here, invalidate it after prepending */
 	list_item_prepend(buffer_comp_list(buffer, dir), comp_list);
+	if (further_buffers_exist)
+		dcache_invalidate_region(needs_sync, sizeof(struct list_item));
 	buffer_set_comp(buffer, comp, dir);
 	irq_local_enable(flags);
 
@@ -214,6 +220,8 @@ int pipeline_connect(struct comp_dev *comp, struct comp_buffer *buffer,
 void pipeline_disconnect(struct comp_dev *comp, struct comp_buffer *buffer, int dir)
 {
 	struct list_item *buf_list, *comp_list;
+	struct list_item __sparse_cache *needs_sync_prev, *needs_sync_next;
+	bool buffers_after_exist, buffers_before_exist;
 	uint32_t flags;
 
 	if (dir == PPL_CONN_DIR_COMP_TO_BUFFER)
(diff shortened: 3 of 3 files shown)


lgirdwood (commented):
@lyakh does this fix or improve a particular issue ?

lyakh:
SOFCI TEST

src/audio/buffer.c
```diff
@@ -40,9 +41,12 @@ struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t align)
...
 	}
 
-	/* allocate new buffer */
+	/*
+	 * allocate new buffer, align the allocation size to a cache line for
+	 * the coherent API
+	 */
 	buffer = rzalloc(SOF_MEM_ZONE_RUNTIME_SHARED, 0, SOF_MEM_CAPS_RAM,
-			 sizeof(*buffer));
+			 ALIGN_UP(sizeof(*buffer), PLATFORM_DCACHE_ALIGN));
```
lgirdwood, line 49: Do we have a ```rzalloc_aligned()``` if not, would be cleaner (and it could align at beginning and end).

lyakh: @lgirdwood no, we don't, and the beginning is always aligned

lyakh:
CI: I assume https://sof-ci.01.org/sofpr/PR6730/build2728/devicetest/index.html?model=TGLU_RVP_NOCODEC_IPC4ZPH&testcase=check-runtime-pm-status and two more kmod failures on TGLU_RVP_NOCODEC_IPC4ZPH are results of the known regression, reported earlier today

src/audio/pipeline/pipeline-graph.c
```diff
@@ -201,10 +203,14 @@ int pipeline_connect(struct comp_dev *comp, struct comp_buffer *buffer,
 	 * overwriting the modified header. FIXME: this is still a problem with
 	 * different cores.
 	 */
-	if (!list_is_empty(comp_list))
-		dcache_writeback_invalidate_region(uncache_to_cache(comp_list->next),
-						   sizeof(struct list_item));
+	further_buffers_exist = !list_is_empty(comp_list);
+	needs_sync = uncache_to_cache(comp_list->next);
+	if (further_buffers_exist)
+		dcache_writeback_region(needs_sync, sizeof(struct list_item));
```
lgirdwood, line 209: This will WB the contents around list_item as we only WB/INV on cache line alignment and in multiples of cache line size.

lyakh: @lgirdwood sure, that's exactly what's needed. We need to sync the list head and we cannot avoid syncing the respective cache line, and that's ok

src/audio/pipeline/pipeline-graph.c
```diff
@@ -214,6 +220,8 @@ int pipeline_connect(struct comp_dev *comp, struct comp_buffer *buffer,
 void pipeline_disconnect(struct comp_dev *comp, struct comp_buffer *buffer, int dir)
 {
 	struct list_item *buf_list, *comp_list;
+	struct list_item __sparse_cache *needs_sync_prev, *needs_sync_next;
+	bool buffers_after_exist, buffers_before_exist;
```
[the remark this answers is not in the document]
lyakh: @lgirdwood sorry, coherent API for what? This handles buffer objects and we already are using the coherent API with them, and that's the reason why we need this machinery

marc-hb:
> CI: I assume https://sof-ci.01.org/sofpr/PR6730/build2728/devicetest/index.html?model=TGLU_RVP_NOCODEC_IPC4ZPH&testcase=check-runtime-pm-status and two more kmod failures on TGLU_RVP_NOCODEC_IPC4ZPH are results of the known regression, reported earlier today

Yes looks like system misconfiguration #6735, should be fixed now.

marc-hb:
SOFCI TEST

src/audio/pipeline/pipeline-graph.c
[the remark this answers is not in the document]
lyakh: @lgirdwood Do you mean adding something like `buffer_connect()` and `buffer_disconnect()` to go along with `buffer_acquire()` and `buffer_release()` to include this cache synchronisation internally? That might be an idea. Can we do this: merge this as is, then I'll try to unify all cases where buffers are linked and unlinked, then we'll add inline functions for this cache synchronisation?

lgirdwood (approved):

src/audio/pipeline/pipeline-graph.c
```diff
@@ -232,13 +240,20 @@ void pipeline_disconnect(struct comp_dev *comp, struct comp_buffer *buffer, int
...
+	if (buffers_after_exist)
+		dcache_writeback_region(needs_sync_next, sizeof(struct list_item));
+	if (buffers_before_exist)
+		dcache_writeback_region(needs_sync_prev, sizeof(struct list_item));
+	/* buffers before or after can be prefetched here */
 	list_item_del(buf_list);
+	if (buffers_after_exist)
+		dcache_invalidate_region(needs_sync_next, sizeof(struct list_item));
+	if (buffers_before_exist)
+		dcache_invalidate_region(needs_sync_prev, sizeof(struct list_item));
```
kv2019i, line 256: I now see your point @lyakh that maybe it would be easier to have everything cached :D 

This does kind of scream that this would need to be abstracted somewhere. Basicly it's problematic to have the list members in the buffer struct, and basicly they cannot be used without doing the full dance above.

It still seems this indeed needs to be done, so I'm ok to merge this and then try to improve the abstraction/interface in follow-ups.

kv2019i (commented):
@lyakh  Second commit good (but some comments for future work). I have some opens on the first one, this seems like duplicated alignment of size.

src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -190,7 +190,9 @@ static struct comp_dev *mixout_new(const struct comp_driver *drv,
 
 	dev->ipc_config = *config;
 
-	md = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*md));
+	/* align the allocation size to a cache line for the coherent API */
+	md = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
+		     ALIGN_UP(sizeof(*md), PLATFORM_DCACHE_ALIGN));
```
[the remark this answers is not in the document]
lyakh: @kv2019i no, the alignment that those allocations do (as well as all our allocations) is of the start address. But we also need to align the end address to avoid a part of this allocation sharing a cache-line with a header of the next allocation. This is done with cached allocation, check in alloc.c. But since here we allocate uncached, we align the size manually.

src/audio/pipeline/pipeline-graph.c
lyakh: @kv2019i yes, I have a follow-up PR prepared after this one to put this in a separate function. You still have a function with some pretty complex cache sync, but at least it's separate... And I then reuse those two functions for buffer.c and module_adapter.c

src/audio/mixin_mixout/mixin_mixout.c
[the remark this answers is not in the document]
kv2019i: @lyakh Hmm, unless SOF_MEM_FLAG_COHERENT is passed, SOF_MEM_ZONE_RUNTIME allocs are always cached, and heap_alloc_aligned_cached specifically rounds up the allocation size. Otherwise we'd have issues with Zephyr heap metadata.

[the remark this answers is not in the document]
lyakh: @kv2019i https://github.com/thesofproject/sof/blob/73f6171bb3c8c5b1d0963883014508022445b997/src/audio/buffer.c#L44 allocates as `SHARED` i.e. uncached, right? So that one won't be aligned up. As for mixin-mixout - I think I took the wrong allocation :-/ It's https://github.com/thesofproject/sof/blob/73f6171bb3c8c5b1d0963883014508022445b997/src/audio/mixin_mixout/mixin_mixout.c#L201-L202 that has to be changed - that's the allocation, used with the coherent API... I'll change that... Does that sound correct?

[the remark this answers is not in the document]
kv2019i: @lyakh Ack, _SHARED is uncached.

src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -198,8 +198,10 @@ static struct comp_dev *mixout_new(const struct comp_driver *drv,
 
 	memcpy_s(&md->base_cfg, sizeof(md->base_cfg), spec, sizeof(md->base_cfg));
 
+	/* align the allocation size to a cache line for the coherent API */
 	md->mixed_data_info = rzalloc(SOF_MEM_ZONE_RUNTIME_SHARED, 0, SOF_MEM_CAPS_RAM,
```
kv2019i, line 202: This is now correct. I do wonder if we should actually enforce coherent objects to be initialized with a cached object. It seems very error prone that user needs to allocate the object in a very specific way (that is really specific to coherent API and should be hidden by it). @lyakh @ranj063 @lgirdwood ?

kv2019i (approved):
Thanks, looks good now. I do have one concern on the coherent interface semantics, see comment inline. But this is not in scope of this PR.

src/audio/mixin_mixout/mixin_mixout.c
lyakh: @kv2019i I tried that and I removed that attempt. I'm not 100% sure any more why. I think one of the reasons was that you actually want to store the uncached pointer to that memory, that's what you use. You pass the uncached pointer around, you use it to *acquire* a cached version, then you *release* it, and you never store the cached pointer. So, if you allocate a cached pointer, then you use `cache_to_uncache()` to get its uncached alias and store it. Then in the end you free it, for which you now only have your uncached pointer, so you have to convert it back to cached or otherwise `heap_free()` won't synchronise its cache... You could have `coherent_init()` allocate memory for you and add `coherent_free()` to correctly free it, but the coherent API actually only knows about its header, not the whole object... So, it's possible but not particularly elegant

PR #6731, opened 2022-12-06
ipc4: copier: remove redundant assignment

serhiy-katsyuba-intel:
Removes variable cd->converter[i] assigning to self. This fixes error reported by static code analyzer.

### DIFF
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -610,15 +610,16 @@ static int copier_prepare(struct comp_dev *dev)
 	if (ret == COMP_STATUS_STATE_ALREADY_SET)
 		return PPL_STATUS_PATH_STOP;
 
-	if (cd->endpoint_num) {
+	for (i = 0; i < cd->endpoint_num; i++) {
+		ret = cd->endpoint[i]->drv->ops.prepare(cd->endpoint[i]);
+		if (ret < 0)
+			break;
+	}
 
-		for (i = 0; i < cd->endpoint_num; i++) {
-			ret = cd->endpoint[i]->drv->ops.prepare(cd->endpoint[i]);
-			if (ret < 0)
-				break;
-		}
-	} else {
-		/* set up format conversion function */
+	if (!cd->endpoint_num) {
+		/* set up format conversion function for pin 0, for other pins (if any)
+		 * format is set in IPC4_COPIER_MODULE_CFG_PARAM_SET_SINK_FORMAT handler
+		 */
 		cd->converter[0] = get_converter_func(&cd->config.base.audio_fmt,
 							      &cd->config.out_fmt, ipc4_gtw_none,
 							      ipc4_bidirection);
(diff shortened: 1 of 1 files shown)


src/audio/copier/copier.c
```diff
@@ -610,25 +610,23 @@ static int copier_prepare(struct comp_dev *dev)
 	if (ret == COMP_STATUS_STATE_ALREADY_SET)
 		return PPL_STATUS_PATH_STOP;
 
-	if (cd->endpoint_num) {
+	for (i = 0; i < cd->endpoint_num; i++) {
```
ranj063, line 613: there's a similar check on line 1010. Can you please fix that too?

ranj063: and please lets be respectful in commit messages. maybe say remove redundant assignment instead?

serhiy-katsyuba-intel: > there's a similar check on line 1010. Can you please fix that too?

That should be done as a separate commit (and might be done as a separate PR). The purpose of this commit is just to remove "cd->converter[i] = cd->converter[i];" redundant assignment.

serhiy-katsyuba-intel: > and please lets be respectful in commit messages. maybe say remove redundant assignment instead?

You are right, "redundant" is more appropriate. Re-edited commit message.

marcinszkudlinski (approved):

src/audio/copier/copier.c
```diff
@@ -610,25 +610,23 @@ static int copier_prepare(struct comp_dev *dev)
...
 		cd->converter[0] = get_converter_func(&cd->config.base.audio_fmt,
 							      &cd->config.out_fmt, ipc4_gtw_none,
 							      ipc4_bidirection);
 		if (!cd->converter[0]) {
 			comp_err(dev, "can't support for in format %d, out format %d",
 				 cd->config.base.audio_fmt.depth,  cd->config.out_fmt.depth);
 			ret = -EINVAL;
-		} else {
-			for (i = 1; i < IPC4_COPIER_MODULE_OUTPUT_PINS_COUNT; i++)
-				cd->converter[i] = cd->converter[i];
```
fkwasowi, line 631: If we remove the conversion function setting for the other pins, then in the do_conversion_copy() function if the SINK_QUEUE_ID is different from zero, the conversion will not be performed.

serhiy-katsyuba-intel: Quote from spec:
"... Initial configuration data allows setup audio format of main Copier's pins, input pin #0 and output pin #0 and prepare PCM conversion routine if any is required. However Copier supports up to COPIER_MODULE_OUTPUT_PINS_COUNT output pins. Before any additional output pin is used in binding operation, the host driver has to send run-time parameter to setup sink format (COPIER_MODULE_CFG_PARAM_SET_SINK_FORMAT) first to setup a PCM conversion routine if any is required. ..."

So conversion functions for pins 1 - 3 (if present) are set in copier_set_sink_fmt() -- a handler of large param set COPIER_MODULE_CFG_PARAM_SET_SINK_FORMAT driver request. That call should be done before bind.

The problem we still have, is if COPIER_MODULE_CFG_PARAM_SET_SINK_FORMAT has not been sent by driver the firmware will crash in do_conversion_copy(). It would be nice to verify (in copier_prepare() or copier_params() ?) if for all sinks conversion functions are initialized. If not -- log and return error instead of crash. But this could be addressed in a separate PR.

serhiy-katsyuba-intel: > It seems to me that it should be:
for (i = 1; i < IPC4_COPIER_MODULE_OUTPUT_PINS_COUNT; i++)
cd->converter[i] = cd->converter[0];

No, copier_prepare() is called before pipeline start, so this will replace conversion functions set in copier_set_sink_fmt() with converter[0], which is wrong.

fkwasowi (approved):

PR #6733, opened 2022-12-06
Parameterize google rtc audio processing

perahgren:
Changes in the GoogleRtcAudioProcessing component to
-Allow sample rate and number of channels to explicitly be specified.  
-Update the memory management.
-Allow parameters to be passed.

### DIFF
--- a/src/audio/google_rtc_audio_processing.c
+++ b/src/audio/google_rtc_audio_processing.c
@@ -34,8 +34,9 @@
 
 #include <google_rtc_audio_processing.h>
 #include <google_rtc_audio_processing_platform.h>
+#include <google_rtc_audio_processing_sof_message_reader.h>
 
-#define GOOGLE_RTC_AUDIO_PROCESSING_SAMPLERATE 48000
+#define GOOGLE_RTC_AUDIO_PROCESSING_FREQENCY_TO_PERIOD_FRAMES 100
 
 LOG_MODULE_REGISTER(google_rtc_audio_processing, CONFIG_SOF_LOG_LEVEL);
 
@@ -52,13 +53,15 @@ struct google_rtc_audio_processing_comp_data {
 	struct comp_buffer *output;
 	uint32_t num_frames;
 	int num_aec_reference_channels;
+	int num_capture_channels;
 	GoogleRtcAudioProcessingState *state;
 	int16_t *aec_reference_buffer;
-	int16_t aec_reference_frame_index;
+	int aec_reference_frame_index;
 	int16_t *raw_mic_buffer;
-	int16_t raw_mic_buffer_index;
+	int raw_mic_buffer_frame_index;
 	int16_t *output_buffer;
-	int16_t output_buffer_index;
+	int output_buffer_frame_index;
+	uint8_t *memory_buffer;
 	struct comp_data_blob_handler *tuning_handler;
 	bool reconfigure;
 };
@@ -113,11 +116,110 @@ static int google_rtc_audio_processing_reconfigure(struct comp_dev *dev)
 		  config, size);
 
 	cd->reconfigure = false;
-	ret = GoogleRtcAudioProcessingReconfigure(cd->state, config, size);
-	if (ret) {
-		comp_err(dev, "GoogleRtcAudioProcessingReconfigure failed: %d",
-			 ret);
-		return -EINVAL;
+
+	uint8_t *google_rtc_audio_processing_config;
+	size_t google_rtc_audio_processing_config_size;
+	int num_capture_input_channels;
+	int num_capture_output_channels;
+	float aec_reference_delay;
+	float mic_gain;
+	bool google_rtc_audio_processing_config_present;
+	bool num_capture_input_channels_present;
+	bool num_capture_output_channels_present;
+	bool aec_reference_delay_present;
+	bool mic_gain_present;
+
+	GoogleRtcAudioProcessingParseSofConfigMessage(config, size,
+						      &google_rtc_audio_processing_config,
+						      &google_rtc_audio_processing_config_size,
+						      &num_capture_input_channels,
+						      &num_capture_output_channels,
+						      &aec_reference_delay,
+						      &mic_gain,
+						      &google_rtc_audio_processing_config_present,
+						      &num_capture_input_channels_present,
+						      &num_capture_output_channels_present,
+						      &aec_reference_delay_present,
+						      &mic_gain_present);
+
+	if (google_rtc_audio_processing_config_present) {
+		comp_info(dev,
+			  "google_rtc_audio_processing_reconfigure(): Applying config of size %zu bytes",
+			  google_rtc_audio_processing_config_size);
+
+		ret = GoogleRtcAudioProcessingReconfigure(cd->state,
+							  google_rtc_audio_processing_config,
+							  google_rtc_audio_processing_config_size);
+		if (ret) {
+			comp_err(dev, "GoogleRtcAudioProcessingReconfigure failed: %d",
+				 ret);
+			return ret;
+		}
+	}
+
+	if (num_capture_input_channels_present || num_capture_output_channels_present) {
+		if (num_capture_input_channels_present && num_capture_output_channels_present) {
+			if (num_capture_input_channels != num_capture_output_channels) {
+				comp_err(dev, "GoogleRtcAudioProcessingReconfigure failed: unsupported channel counts");
+				return -EINVAL;
+			}
+			cd->num_capture_channels = num_capture_input_channels;
+		} else if (num_capture_input_channels_present) {
+			cd->num_capture_channels = num_capture_output_channels;
+		} else {
+			cd->num_capture_channels = num_capture_output_channels;
+		}
+		comp_info(dev,
+			  "google_rtc_audio_processing_reconfigure(): Applying num capture channels %d",
+			  cd->num_capture_channels);
+
+
+		ret = GoogleRtcAudioProcessingSetStreamFormats(cd->state,
+							       CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_SAMPLE_RATE_HZ,
+							       cd->num_capture_channels,
+							       cd->num_capture_channels,
+							       CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_SAMPLE_RATE_HZ,
+							       cd->num_aec_reference_channels);
+
+		if (ret) {
+			comp_err(dev, "GoogleRtcAudioProcessingSetStreamFormats failed: %d",
+				 ret);
+			return ret;
+		}
+	}
+
+	if (aec_reference_delay_present || mic_gain_present) {
+		float *capture_headroom_linear_use = NULL;
+		float *echo_path_delay_ms_use = NULL;
+
+		if (mic_gain_present) {
+			capture_headroom_linear_use = &mic_gain;
+
+			/* Logging of linear headroom, using integer workaround to the broken printout of floats */
+			comp_info(dev,
+				  "google_rtc_audio_processing_reconfigure(): Applying capture linear headroom: %d.%d",
+				  (int)mic_gain, (int)(100 * mic_gain) - 100 * ((int)mic_gain));
+		}
+		if (aec_reference_delay_present) {
+			echo_path_delay_ms_use = &aec_reference_delay;
+
+			/* Logging of delay, using integer workaround to the broken printout of floats */
+			comp_info(dev,
+				  "google_rtc_audio_processing_reconfigure(): Applying aec reference delay: %d.%d",
+				  (int)aec_reference_delay,
+				  (int)(100 * aec_reference_delay) -
+				  100 * ((int)aec_reference_delay));
+		}
+
+		ret = GoogleRtcAudioProcessingParameters(cd->state,
+							 capture_headroom_linear_use,
+							 echo_path_delay_ms_use);
+
+		if (ret) {
+			comp_err(dev, "GoogleRtcAudioProcessingParameters failed: %d",
+				 ret);
+			return ret;
+		}
 	}
 
 	return 0;
@@ -205,6 +307,7 @@ static struct comp_dev *google_rtc_audio_processing_create(
 {
 	struct comp_dev *dev;
 	struct google_rtc_audio_processing_comp_data *cd;
+	int ret;
 
 	comp_cl_info(drv, "google_rtc_audio_processing_create()");
 
@@ -225,22 +328,54 @@ static struct comp_dev *google_rtc_audio_processing_create(
 	if (!cd->tuning_handler)
 		goto fail;
 
-	cd->state = GoogleRtcAudioProcessingCreate();
+	cd->num_aec_reference_channels = CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_NUM_AEC_REFERENCE_CHANNELS;
+	cd->num_capture_channels = CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_NUM_CHANNELS;
+	cd->num_frames = CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_SAMPLE_RATE_HZ /
+		GOOGLE_RTC_AUDIO_PROCESSING_FREQENCY_TO_PERIOD_FRAMES;
+
+	if (CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_MEMORY_BUFFER_SIZE_BYTES > 0) {
+		cd->memory_buffer = rballoc(0, SOF_MEM_CAPS_RAM,
+					    CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_MEMORY_BUFFER_SIZE_BYTES *
+					    sizeof(cd->memory_buffer[0]));
+		if (!cd->memory_buffer) {
+			comp_err(dev, "google_rtc_audio_processing_create: failed to allocate memory buffer");
+
+			goto fail;
+		}
+
+		GoogleRtcAudioProcessingAttachMemoryBuffer(cd->memory_buffer, CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_MEMORY_BUFFER_SIZE_BYTES);
+	}
+
+	cd->state = GoogleRtcAudioProcessingCreateWithConfig(CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_SAMPLE_RATE_HZ,
+							     cd->num_capture_channels,
+							     cd->num_capture_channels,
+							     CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_SAMPLE_RATE_HZ,
+							     cd->num_aec_reference_channels,
+							     /*config=*/NULL, /*config_size=*/0);
+
 	if (!cd->state) {
 		comp_err(dev, "Failed to initialized GoogleRtcAudioProcessing");
 		goto fail;
 	}
-	cd->num_aec_reference_channels = 2;
-	cd->num_frames = GOOGLE_RTC_AUDIO_PROCESSING_SAMPLERATE *
-					 GoogleRtcAudioProcessingGetFramesizeInMs(cd->state) / 1000;
+
+	float capture_headroom_linear = CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_MIC_HEADROOM_LINEAR;
+	float echo_path_delay_ms = CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_ECHO_PATH_DELAY_MS;
+	ret = GoogleRtcAudioProcessingParameters(cd->state,
+						 &capture_headroom_linear,
+						 &echo_path_delay_ms);
+
+	if (!cd->state) {
+		comp_err(dev, "Failed to apply GoogleRtcAudioProcessingParameters");
+		goto fail;
+	}
 
 	cd->raw_mic_buffer = rballoc(
 		0, SOF_MEM_CAPS_RAM,
-		cd->num_frames * sizeof(cd->raw_mic_buffer[0]));
+		cd->num_frames * cd->num_capture_channels * sizeof(cd->raw_mic_buffer[0]));
 	if (!cd->raw_mic_buffer)
 		goto fail;
-	bzero(cd->raw_mic_buffer, cd->num_frames * sizeof(cd->raw_mic_buffer[0]));
-	cd->raw_mic_buffer_index = 0;
+	bzero(cd->raw_mic_buffer, cd->num_frames * cd->num_capture_channels * sizeof(cd->raw_mic_buffer[0]));
+	cd->raw_mic_buffer_frame_index = 0;
 
 	cd->aec_reference_buffer = rballoc(
 		0, SOF_MEM_CAPS_RAM,
@@ -248,16 +383,16 @@ static struct comp_dev *google_rtc_audio_processing_create(
 		cd->num_aec_reference_channels);
 	if (!cd->aec_reference_buffer)
 		goto fail;
-	bzero(cd->aec_reference_buffer, cd->num_frames * sizeof(cd->aec_reference_buffer[0]));
+	bzero(cd->aec_reference_buffer, cd->num_frames * cd->num_aec_reference_channels * sizeof(cd->aec_reference_buffer[0]));
 	cd->aec_reference_frame_index = 0;
 
 	cd->output_buffer = rballoc(
 		0, SOF_MEM_CAPS_RAM,
-		cd->num_frames * sizeof(cd->output_buffer[0]));
+		cd->num_frames * cd->num_capture_channels * sizeof(cd->output_buffer[0]));
 	if (!cd->output_buffer)
 		goto fail;
 	bzero(cd->output_buffer, cd->num_frames * sizeof(cd->output_buffer[0]));
-	cd->output_buffer_index = 0;
+	cd->output_buffer_frame_index = 0;
 
 	/* comp_is_new_data_blob_available always returns false for the first
 	 * control write with non-empty config. The first non-empty write may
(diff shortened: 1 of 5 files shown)


src/audio/google_rtc_audio_processing.c
```diff
@@ -225,14 +330,36 @@ static struct comp_dev *google_rtc_audio_processing_create(
 	if (!cd->tuning_handler)
 		goto fail;
 
-	cd->state = GoogleRtcAudioProcessingCreate();
+	if (MEMORY_BUFFER_BYTES > 0) {
```
cujomalainey, line 333: always true as this is hard coded

src/audio/google_rtc_audio_processing.c
```diff
@@ -225,14 +330,36 @@ static struct comp_dev *google_rtc_audio_processing_create(
...
+					    MEMORY_BUFFER_BYTES * sizeof(cd->memory_buffer[0]));
+		if (!cd->memory_buffer) {
+			comp_err(dev, "google_rtc_audio_processing_create: failed to allocate memory buffer");
+
+			goto fail;
+		}
+
+		GoogleRtcAudioProcessingAttachMemoryBuffer(cd->memory_buffer, MEMORY_BUFFER_BYTES);
+	} else {
+		cd->memory_buffer = NULL;
```
cujomalainey, line 344: this should be impossible to get to, and if you did modify ti to be accessible you will leak here

src/audio/google_rtc_audio_processing.c
```diff
@@ -275,6 +402,10 @@ static struct comp_dev *google_rtc_audio_processing_create(
 	if (cd) {
 		rfree(cd->output_buffer);
 		rfree(cd->aec_reference_buffer);
+		if (cd->memory_buffer) {
```
cujomalainey, line 405: not sure about your API call but memory buffer does not need to be in the if as rfree(NULL) is a no-op

src/audio/google_rtc_audio_processing_mock.c
```diff
@@ -45,6 +54,35 @@ GoogleRtcAudioProcessingState *GoogleRtcAudioProcessingCreate()
...
+									int config_size)
+{
+	struct GoogleRtcAudioProcessingState *s =
+		rballoc(0, SOF_MEM_CAPS_RAM, sizeof(GoogleRtcAudioProcessingState));
+	if (!s)
+		return NULL;
+	s->num_capture_channels = num_capture_input_channels;
+	s->num_aec_reference_channels = num_render_channels;
+	s->num_output_channels = num_capture_output_channels;
+	s->num_frames = capture_sample_rate_hz * 10 / 1000;
```
cujomalainey, line 72: what is 10/1000? is it a time conversion?

src/audio/google_rtc_audio_processing_mock.c
```diff
@@ -53,6 +91,35 @@ void GoogleRtcAudioProcessingFree(GoogleRtcAudioProcessingState *state)
...
+	s->aec_reference = rballoc(0,
+				   SOF_MEM_CAPS_RAM,
+				   sizeof(s->aec_reference[0]) *
+				   s->num_frames *
+				   s->num_aec_reference_channels
+				   );
+	return 0;
```
cujomalainey, line 113: this looks very similar to above, can we maybe reuse it with helpers or inlines to mask out variables?

src/audio/google_rtc_audio_processing_mock.c
```diff
@@ -94,3 +161,30 @@ int GoogleRtcAudioProcessingAnalyzeRender_int16(
 		 data, buffer_size);
 	return 0;
 }
+
+void GoogleRtcAudioProcessingParseSofConfigMessage(uint8_t *message,
```
cujomalainey, line 165: this does not appear to be parsing

perahgren: No, the mock does not do the parsing and instead just flags that no parameters were specified.

src/audio/google_rtc_audio_processing_mock.c
perahgren: 10 is the frame size in ms, and the computation produces in the frame size in samples.
This now changed this to make it more clear.

src/audio/google_rtc_audio_processing.c
perahgren: I don't follow.
If MEMORY_BUFFER is zero, this simply initializes memory_buffer to NULL, and no allocation is done of memory_buffer. Therefore there should not be any leakage.

src/audio/google_rtc_audio_processing.c
```diff
@@ -34,8 +34,18 @@
...
-#define GOOGLE_RTC_AUDIO_PROCESSING_SAMPLERATE 48000
+#define SAMPLE_RATE_HZ 48000
+#define SAMPLES_PER_CHANNEL (SAMPLE_RATE_HZ / 100)
+#define DEFAULT_NUM_AEC_REFERENCE_CHANNELS 2
+#define DEFAULT_NUM_CAPTURE_INPUT_CHANNELS 1
+#define DEFAULT_NUM_CAPTURE_OUTPUT_CHANNELS 1
+
+#define MEMORY_BUFFER_STEREO_SIZE_BYTES 250000
+#define MEMORY_BUFFER_MONO_SIZE_BYTES 200000
+#define MEMORY_BUFFER_BYTES (DEFAULT_NUM_CAPTURE_OUTPUT_CHANNELS == 1 ? \
```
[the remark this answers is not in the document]
perahgren: Yes. But my concern with that is that this size needs to be defined compile-time, so if a macro is defined, it mustn't be used on any runtime-specified parameters. 
With that I'm not convinced that adding a macro is the correct thing to do.

marc-hb:
@perahgren since you're doing "rapide-fire force pushes" I change the PR title and added the (magic and cryptic, sorry) [SKIP CI] prefix to stop the saturation of our limited hardware resources at sof-ci/jenkins/pr-device-test

src/audio/google_rtc_audio_processing.c
```diff
@@ -34,8 +34,18 @@
 
 #include <google_rtc_audio_processing.h>
 #include <google_rtc_audio_processing_platform.h>
+#include <google_rtc_audio_processing_sof_message_reader.h>
 
-#define GOOGLE_RTC_AUDIO_PROCESSING_SAMPLERATE 48000
+#define SAMPLE_RATE_HZ 48000
+#define SAMPLES_PER_CHANNEL (SAMPLE_RATE_HZ / 100)
```
lgirdwood, line 40: I would be careful with these generic names, best to use the prefix to prevent collisions with global names.

perahgren: Thanks for the headsup! I've now added the GOOGLE_RTC_AUDIO_PROCESSING prefix to avoid that

src/audio/google_rtc_audio_processing.c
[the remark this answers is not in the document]
cujomalainey: You could add a complie time assert on the final calculated value to make sure it is compile time available. Also why does it need to be compile time ready?

src/audio/google_rtc_audio_processing.c
cujomalainey: It's written as if it can chance at runtime but that is weird given it's a constant

src/audio/google_rtc_audio_processing_mock.c
cujomalainey: Haha missed I moved into the mock file, thanks for the pointer

src/audio/google_rtc_audio_processing.c
```diff
@@ -34,8 +34,19 @@
 
 #include <google_rtc_audio_processing.h>
 #include <google_rtc_audio_processing_platform.h>
+#include <google_rtc_audio_processing_sof_message_reader.h>
 
-#define GOOGLE_RTC_AUDIO_PROCESSING_SAMPLERATE 48000
+#define GOOGLE_RTC_AUDIO_PROCESSING_SAMPLE_RATE_HZ 48000
+#define GOOGLE_RTC_AUDIO_PROCESSING_SAMPLES_PER_CHANNEL (GOOGLE_RTC_AUDIO_PROCESSING_SAMPLE_RATE_HZ / 100)
```
cujomalainey, line 40: PERIOD_SAMPLES_PER_CHANNEL

src/audio/google_rtc_audio_processing.c
```diff
@@ -34,8 +34,19 @@
...
+#define GOOGLE_RTC_AUDIO_PROCESSING_DEFAULT_NUM_AEC_REFERENCE_CHANNELS 2
+#define GOOGLE_RTC_AUDIO_PROCESSING_DEFAULT_NUM_CAPTURE_INPUT_CHANNELS 1
```
cujomalainey, line 42: How hard will it be to make this dynamic?

src/audio/google_rtc_audio_processing.c
```diff
@@ -113,11 +127,104 @@ static int google_rtc_audio_processing_reconfigure(struct comp_dev *dev)
...
+						      &num_capture_output_channels,
+						      &aec_reference_delay,
+						      &mic_gain,
+						      &google_rtc_audio_processing_config_present,
+						      &num_capture_input_channels_present,
+						      &num_capture_output_channels_present,
+						      &aec_reference_delay_present,
+						      &mic_gain_present);
+
+	if (num_capture_input_channels_present || num_capture_output_channels_present) {
```
cujomalainey, line 156: isnt the config invalid if these are 0?

src/audio/google_rtc_audio_processing.c
```diff
@@ -113,11 +127,104 @@ static int google_rtc_audio_processing_reconfigure(struct comp_dev *dev)
...
+							       GOOGLE_RTC_AUDIO_PROCESSING_SAMPLE_RATE_HZ,
+							       cd->num_capture_input_channels,
+							       cd->num_capture_output_channels,
+							       GOOGLE_RTC_AUDIO_PROCESSING_SAMPLE_RATE_HZ,
+							       cd->num_aec_reference_channels);
+
+		if (ret) {
+			comp_err(dev, "GoogleRtcAudioProcessingSetStreamFormats failed: %d",
+				 ret);
+			return -EINVAL;
```
cujomalainey, line 180: is the library returning a linux errno? if yes we should just return that

src/audio/google_rtc_audio_processing.c
```diff
@@ -113,11 +127,104 @@ static int google_rtc_audio_processing_reconfigure(struct comp_dev *dev)
...
+
+			comp_info(dev,
+				  "google_rtc_audio_processing_reconfigure(): Applying capture linear headroom: %d.%d",
+				  (int)mic_gain, (int)(100 * mic_gain) - 100 * ((int)mic_gain));
+		}
+		if (aec_reference_delay_present) {
+			echo_path_delay_ms_use = &aec_reference_delay;
+			comp_info(dev,
+				  "google_rtc_audio_processing_reconfigure(): Applying aec reference delay: %d.%d",
+				  (int)aec_reference_delay,
```
cujomalainey, line 199: not sure i grasp what is going on here, you appear to be logging a pointer, is that intended?

src/audio/google_rtc_audio_processing.c
```diff
@@ -325,7 +460,7 @@ static int google_rtc_audio_processing_prepare(struct comp_dev *dev)
 
 		if (source_c->source->pipeline->pipeline_id != dev->pipeline->pipeline_id) {
 			cd->aec_reference = source;
-			aec_channels = sourcs_c->stream.channels;
```
cujomalainey, line 328: this should be its own commit so it can be cherrypicked without your change if needed

src/audio/google_rtc_audio_processing_mock.c
```diff
@@ -22,29 +23,79 @@ struct GoogleRtcAudioProcessingState {
...
+		       int num_capture_input_channels,
+		       int num_capture_output_channels,
+		       int render_sample_rate_hz,
+		       int num_render_channels)
+{
+	s->num_capture_channels = num_capture_input_channels;
+	s->num_output_channels = num_capture_output_channels;
+	s->num_frames = capture_sample_rate_hz / GOOGLE_RTC_AUDIO_PROCESSING_FRAMES_PER_SECOND;
+
+	if (s->num_aec_reference_channels != num_render_channels ||
```
cujomalainey, line 37: brackets for clarity on logic

cujomalainey (changes requested):

src/audio/google_rtc_audio_processing.c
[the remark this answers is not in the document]
perahgren: Yes, I think you are right, and my reasons for keeping this as a #define does not really hold.
I changed it to a function instead though (because I would prefer to not use a macro).

The information needs to be available at the call to _create though. However, if the buffer allocated is too small, the functionality fallbacks to using the SOF allocation routines.

src/audio/google_rtc_audio_processing.c
perahgren: The code is designed to work also when setting either GOOGLE_RTC_AUDIO_PROCESSING_MEMORY_BUFFER_STEREO_SIZE_BYTES
GOOGLE_RTC_AUDIO_PROCESSING_MEMORY_BUFFER_MONO_SIZE_BYTES to zero.

It looks slightly different now though

src/audio/google_rtc_audio_processing.c
perahgren: Nice! I removed that.

src/audio/google_rtc_audio_processing.c
perahgren: Done

src/audio/google_rtc_audio_processing.c
perahgren: It is doable, but there are some requirements involved.

The trickiest part is that these needs to be available at _create, which means that when they are needed, there is yet no way to pass values from outside of the firmware.

src/audio/google_rtc_audio_processing.c
perahgren: No, these values are only optional in the config. 
The config can now carry different optional payloads, and the presence of those is flagged using the bools
The channel counts passed in the config can be 0, but such will be rejected by GoogleRtcAudioProcessingSetStreamFormats (and an error code will be returned)

src/audio/google_rtc_audio_processing.c
perahgren: Good point! Done.

src/audio/google_rtc_audio_processing.c
perahgren: What I wanted to achieve is to print out a floating point value.
However, it seems like with comp_info, using #f for floating point values is broken, and just printing random values.

This was the best I could do to achieve a log-line with a floating point value

src/audio/google_rtc_audio_processing.c
perahgren: Done. Reverted my change.

src/audio/google_rtc_audio_processing_mock.c
perahgren: Like this?

src/audio/google_rtc_audio_processing.c
cujomalainey: Got it, in the meantime how about kconfig? We should probably also have checks to make sure the pipeline isn't outside out spec

src/audio/google_rtc_audio_processing.c
cujomalainey: Might be worth adding a comment since its not particularly clear

src/audio/google_rtc_audio_processing.c
```diff
@@ -225,14 +337,36 @@ static struct comp_dev *google_rtc_audio_processing_create(
...
 		goto fail;
 
-	cd->state = GoogleRtcAudioProcessingCreate();
+	cd->num_aec_reference_channels = GOOGLE_RTC_AUDIO_PROCESSING_DEFAULT_NUM_AEC_REFERENCE_CHANNELS;
+	cd->num_capture_input_channels = GOOGLE_RTC_AUDIO_PROCESSING_DEFAULT_NUM_CAPTURE_INPUT_CHANNELS;
+	cd->num_capture_output_channels = GOOGLE_RTC_AUDIO_PROCESSING_DEFAULT_NUM_CAPTURE_OUTPUT_CHANNELS;
+	cd->num_frames = GOOGLE_RTC_AUDIO_PROCESSING_PERIOD_SAMPLES_PER_CHANNEL;
+
+	int memory_buffer_size_bytes = get_memory_buffer_size_bytes(cd->num_capture_output_channels);
+	if (memory_buffer_size_bytes > 0) {
```
cujomalainey, line 346: based on the function call, this appears to be always true

src/audio/google_rtc_audio_processing.c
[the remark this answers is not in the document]
cujomalainey: My biggest issue is that it detaches the pipeline expectations from the component. I think it would be wiser to have the component check the pipeline against its needs rather than assume based on a compilation config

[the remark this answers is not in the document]
cujomalainey: That being said, we do have the issue right now where DMIC is more channels than we actually care about. Might be worth fixing up a demux properly. @sathya-nujella FYI

src/audio/google_rtc_audio_processing_mock.c
cujomalainey: Might be worth defineing that as it is not inherently clear

src/audio/google_rtc_audio_processing_mock.c
```diff
@@ -22,29 +23,79 @@ struct GoogleRtcAudioProcessingState {
...
+		       int num_render_channels)
+{
+	s->num_capture_channels = num_capture_input_channels;
+	s->num_output_channels = num_capture_output_channels;
+	s->num_frames = capture_sample_rate_hz / GOOGLE_RTC_AUDIO_PROCESSING_FRAMES_PER_SECOND;
+
+	if ((s->num_aec_reference_channels != num_render_channels) ||
+	    !s->aec_reference) {
+		s->num_aec_reference_channels = num_render_channels;
+		if (s->aec_reference) {
```
cujomalainey, line 40: if not needed

perahgren: True. Removed that.

src/audio/google_rtc_audio_processing.c
perahgren: Good point.
Done

src/audio/google_rtc_audio_processing_mock.c
perahgren: Done.

src/audio/google_rtc_audio_processing.c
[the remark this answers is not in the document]
perahgren: The scheme here is fairly easily extended to react on the pipeline needs.
Right now, however, the component _copy code is so hardcoded to always process mono mic channels, so I'd prefer to postpone that change until the _copy code is properly extended.

src/audio/google_rtc_audio_processing.c
cujomalainey: I think 1 is fine but then it needs to be clear it can be modified, e.g. an optional kconfig

cujomalainey (approved):
one nit, also approval with the agreement we move away from hard coded defines in the future

src/audio/google_rtc_audio_processing_mock.c
```diff
@@ -22,29 +24,76 @@ struct GoogleRtcAudioProcessingState {
...
-	s->num_capture_channels = 1;
-	s->num_aec_reference_channels = 2;
-	s->num_output_channels = 1;
-	s->num_frames = GOOGLE_RTC_AUDIO_PROCESSING_SAMPLE_RATE_HZ * 10 / 1000;
+	s->num_capture_channels = num_capture_input_channels;
+	s->num_output_channels = num_capture_output_channels;
+	s->num_frames = capture_sample_rate_hz / GOOGLE_RTC_AUDIO_PROCESSING_FRAMES_PER_SECOND;
+
+	s->num_aec_reference_channels = num_render_channels;
+	if (s->aec_reference) {
```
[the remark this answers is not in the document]
perahgren: I think that is needed, since SetFormats may be called with it being NULL: i.e., from line 72.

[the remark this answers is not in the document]
perahgren: Well spotted. That indeed was wrong.
Done

src/audio/google_rtc_audio_processing_mock.c
```diff
@@ -22,29 +24,76 @@ struct GoogleRtcAudioProcessingState {
...
+		       int render_sample_rate_hz,
+		       int num_render_channels)
+{
+	state->num_capture_channels = num_capture_input_channels;
+	state->num_output_channels = num_capture_output_channels;
+	state->num_frames = capture_sample_rate_hz / GOOGLE_RTC_AUDIO_PROCESSING_FRAMES_PER_SECOND;
+
+	state->num_aec_reference_channels = num_render_channels;
+	if (state->aec_reference) {
+		rfree(state->aec_reference);
```
cujomalainey, line 40: `if` isn't needed

perahgren: I think it is needed, since SetFormats could both be called with state->aec_reference allocated, and being NULL.

E.g., on line 72 it is NULL, and on line 112 it is allocated.

src/audio/google_rtc_audio_processing_mock.c
```diff
@@ -22,29 +24,74 @@ struct GoogleRtcAudioProcessingState {
...
+	state->num_output_channels = num_capture_output_channels;
+	state->num_frames = capture_sample_rate_hz / GOOGLE_RTC_AUDIO_PROCESSING_FRAMES_PER_SECOND;
+
+	state->num_aec_reference_channels = num_render_channels;
+	rfree(state->aec_reference);
+	state->aec_reference = rballoc(0,
+				       SOF_MEM_CAPS_RAM,
+				       sizeof(state->aec_reference[0]) *
+				       state->num_frames *
+				       state->num_aec_reference_channels);
```
lyakh, line 44: would `rbrealloc()` work for you?

src/audio/google_rtc_audio_processing_mock.c
```diff
@@ -53,21 +100,46 @@ void GoogleRtcAudioProcessingFree(GoogleRtcAudioProcessingState *state)
...
+	return state->num_frames *
+		GOOGLE_RTC_AUDIO_PROCESSING_MS_PER_SECOND /
+		GOOGLE_RTC_AUDIO_PROCESSING_SAMPLE_RATE_HZ;
 }
 
 int GoogleRtcAudioProcessingReconfigure(
-	GoogleRtcAudioProcessingState * const state, const uint8_t *const config,
-	int config_size)
+					GoogleRtcAudioProcessingState * const state, const uint8_t *const config,
+					int config_size)
```
lyakh, line 135: if you're aligning these parameters now, you could just put the first parameter one line above and move all other parameters onto the second line. Same below

perahgren: True. 
Done.

src/audio/google_rtc_audio_processing_mock.c
cujomalainey: @lyakh given this is test code im inclined to reject this change. If it was in the adapter then I would agree

src/audio/google_rtc_audio_processing.c
```diff
@@ -275,6 +410,8 @@ static struct comp_dev *google_rtc_audio_processing_create(
 	if (cd) {
 		rfree(cd->output_buffer);
 		rfree(cd->aec_reference_buffer);
+		GoogleRtcAudioProcessingDetachMemoryBuffer();
```
cujomalainey, line 413: where is `cd->state` cleaned up?

perahgren: Good find! It is not at all cleaned up here, and has never been.

I added a conditional cleanup.
PTAL.

cujomalainey (approved):

src/audio/google_rtc_audio_processing.c
```diff
@@ -293,6 +433,8 @@ static void google_rtc_audio_processing_free(struct comp_dev *dev)
 	cd->state = NULL;
 	rfree(cd->output_buffer);
 	rfree(cd->aec_reference_buffer);
+	GoogleRtcAudioProcessingDetachMemoryBuffer();
+	rfree(cd->memory_buffer);
```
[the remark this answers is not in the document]
cujomalainey: nvm im blind

cujomalainey (approved):

src/audio/google_rtc_audio_processing.c
```diff
@@ -34,8 +34,9 @@
 
 #include <google_rtc_audio_processing.h>
 #include <google_rtc_audio_processing_platform.h>
+#include <google_rtc_audio_processing_sof_message_reader.h>
 
-#define GOOGLE_RTC_AUDIO_PROCESSING_SAMPLERATE 48000
+#define GOOGLE_RTC_AUDIO_PROCESSING_FRAMES_PER_SECOND 100
```
cujomalainey, line 39: I think this name is false, based on the usage it looks like its better to make this macro called FREQENCY_TO_PERIOD_FRAMES

src/audio/google_rtc_audio_processing_mock.c
```diff
@@ -13,6 +13,8 @@
 #include "ipc/topology.h"
 
 #define GOOGLE_RTC_AUDIO_PROCESSING_SAMPLE_RATE_HZ 48000
```
cujomalainey, line 15: this is now a config

src/audio/google_rtc_audio_processing_mock.c
```diff
@@ -13,6 +13,8 @@
...
+#define GOOGLE_RTC_AUDIO_PROCESSING_FRAMES_PER_SECOND 100
```
cujomalainey, line 16: this is redefined, should it go in a header?

src/audio/google_rtc_audio_processing_mock.c
perahgren: Well spotted. I now changed to using that.

src/audio/google_rtc_audio_processing_mock.c
perahgren: I would prefer to not put it into the header. google_rtc_audio_processing.h and instead redefine it here.

src/audio/google_rtc_audio_processing.c
perahgren: Sounds good.
Done.

cujomalainey (approved):

cujomalainey:
Merging as this is a google isolated change. Note to readers, next week I will move this to our own folder with our own kconfig.

PR #6736, opened 2022-12-07
ipc4: only primary core sends ipc to host

pblaszko:
Remove option to immediately send IPC in ipc_send_msg().

In current implementation, secondary core can directly write IPC mailbox and registers. It may lead to conflict with IPC's that are beeing send from queue on primary core.

All IPC responses and notifications should be queued on primary core to synchronize two types of DSP to host messages between cores.

Changing flow only for IPC4/Zephyr. Leaving XTOS path unchanged.

### DIFF
--- a/src/ipc/ipc-common.c
+++ b/src/ipc/ipc-common.c
@@ -188,13 +188,23 @@ void ipc_msg_send(struct ipc_msg *msg, void *data, bool high_priority)
 		assert(!ret);
 	}
 
+	/*
+	 * note: This function can be executed in LL or EDF context, from any core.
+	 * In Zephyr builds, there is IPC queue that is always handled by the primary core,
+	 * whereas submitting to the queue is allowed from any core. Therefore disable option
+	 * of sending IPC immediately by any context/core to secure IPC registers/mailbox
+	 * access.
+	 */
+#ifndef __ZEPHYR__
 	/* try to send critical notifications right away */
 	if (high_priority) {
 		ret = ipc_platform_send_msg(msg);
-		if (!ret)
-			goto out;
+		if (!ret) {
+			k_spin_unlock(&ipc->lock, key);
+			return;
+		}
 	}
-
+#endif
 	/* add to queue unless already there */
 	if (list_is_empty(&msg->list)) {
 		if (high_priority)
@@ -205,7 +215,6 @@ void ipc_msg_send(struct ipc_msg *msg, void *data, bool high_priority)
 
 	schedule_ipc_worker();
 
-out:
 	k_spin_unlock(&ipc->lock, key);
 }
 


src/ipc/ipc-common.c
```diff
@@ -188,13 +188,16 @@ void ipc_msg_send(struct ipc_msg *msg, void *data, bool high_priority)
 		assert(!ret);
 	}
 
+#ifndef __ZEPHYR__
```
lgirdwood, line 191: Difficult to understand the code, it may be easier to follow if we 
```c
if (cpu_is_me() != PRIMARY) {
    // some error log and return value
}
```
This way we will catch anyone that attempts this today and it could be extended in the future to allow more IPC capable cores if needed.

miRoox:
SOFCI TEST

src/ipc/ipc-common.c
lyakh: @lgirdwood I don't think this code does anything depending on the core, where it's running. It just makes sure, that *no* IPC ever gets sent immediately from here but always from the queue. It is the IPC queue that is always handled by the primary core, whereas submitting to the queue is allowed from any core.

pblaszko: @lgirdwood My intention was to queue all IPC's to Host, regardless if this is a primary or secondary core. Even in context of single core, notifications and IPC responses are asynchronous: EDF thread may want send IPC response and be interrupted by LL thread which want to send notification. For single core, current synchronization mechanisms can secure this flow, but for multicore the situation gets more complicated. That is why I think its more consistent and secure if all IPC's are queued on queue executed only on primary core.
Normally I would just remove the part of code trying to send IPC immediately, but there is still XTOS configuration where I think sending IPC's by secondary core was allowed. I did not want to change flow of IPC3/XTOS, as I am focusing on IPC4/Zephyr.

kv2019i: @pblaszko Please add an inline comment why you ifdef the code out on Zephyr. Code is pretty hard to follow otherwise.

kv2019i (changes requested):
Hmm, let's see the CI results on TGL. This is a pretty big change to IPC timing and might cause unexpected issues. But if results are good, then seems ok.

tmleman (approved):
LGTM

src/ipc/ipc-common.c
pblaszko: > @lgirdwood I don't think this code does anything depending on the core, where it's running. It just makes sure, that _no_ IPC ever gets sent immediately from here but always from the queue. It is the IPC queue that is always handled by the primary core, whereas submitting to the queue is allowed from any core.

@lyakh Your comment catch the essence of the change, so if you don't mind, I put it to inline comment :)

pblaszko:
> Hmm, let's see the CI results on TGL. This is a pretty big change to IPC timing and might cause unexpected issues. But if results are good, then seems ok.

@kv2019i I added inline comment.
What CI do you mean? Of course I am open to test it by any possible way. I think all Internal Intel CI pass.

kv2019i:
@pblaszko wrote:
> @kv2019i I added inline comment. What CI do you mean? Of course I am open to test it by any possible way. I think all Internal Intel CI pass.

Just the CI tests run here on the pull-request. Seem to be clean!

ranj063 (approved):

src/ipc/ipc-common.c
lyakh: @pblaszko sure, feel free

lgirdwood (approved):

PR #6738, opened 2022-12-07
fix Klocwork errors and criticals from logger sources

aborisovich:
Files tools\logger\logger.c and tools\logger\convert.c had been verified for issues and had been fixed.

### DIFF
--- a/tools/logger/convert.c
+++ b/tools/logger/convert.c
@@ -131,7 +131,8 @@ static const char *format_uid(uint32_t uid_ptr, int use_colors, bool be, bool up
 	if (uid_ptr < uids_dict->base_address ||
 	    uid_ptr >= uids_dict->base_address + uids_dict->data_length) {
 		str = calloc(1, strlen(BAD_PTR_STR) + 1 + 6);
-		sprintf(str, BAD_PTR_STR, uid_ptr);
+		if (str)
+			sprintf(str, BAD_PTR_STR, uid_ptr);
 	} else {
 		uid_entry = get_uuid_entry(uid_ptr);
 		str = format_uid_raw(uid_entry, use_colors, 1, be, upper);
@@ -358,8 +359,10 @@ static inline void print_table_header(void)
 		/* e.g.: ktime=4263.487s @ 2021-04-27 14:21:13 -0700 PDT */
 		fprintf(out_fd, "\tktime=%lu.%03lus",
 			ktime.tv_sec, ktime.tv_nsec / 1000000);
-		if (strftime(date_string, sizeof(date_string),
-			     "%F %X %z %Z", localtime(&epoc_secs)))
+		struct tm *info = localtime(&epoc_secs);
+
+		if (info && strftime(date_string, sizeof(date_string),
+				    "%F %X %z %Z", localtime(&epoc_secs)))
 			fprintf(out_fd, "  @  %s", date_string);
 	}
 
@@ -961,7 +964,8 @@ static int verify_ldc_checksum(const uint32_t ldc_sum)
 	count = fread(&ver, sizeof(ver), 1, global_config->version_fd);
 	if (!count) {
 		log_err("Error while reading %s.\n", global_config->version_file);
-		return -ferror(global_config->version_fd);
+		int result = -ferror(global_config->version_fd);
+		return result;
 	}
 
 	/* compare source hash value from version file and ldc file */
@@ -1017,6 +1021,10 @@ static int dump_ldc_info(void)
 	while (remaining > 0) {
 		name = format_uid_raw(&uid_ptr[cnt], 0, 0, false, false);
 		uid_addr = get_uuid_key(&uid_ptr[cnt]);
+		if (!name) {
+			log_err("Failed to allocate memory for uid name\n");
+			exit(1);
+		}
 		fprintf(out_fd, "\t%p  %s\n", (void *)uid_addr, name);
 
 		if (name) {
@@ -1045,19 +1053,25 @@ int convert(struct convert_config *config)
 	count = fread(&snd, sizeof(snd), 1, config->ldc_fd);
 	if (!count) {
 		log_err("Error while reading %s.\n", config->ldc_file);
-		return -ferror(config->ldc_fd);
+		int result = -ferror(config->ldc_fd);
+
+		config = NULL;
+		return result;
 	}
 
 	if (strncmp((char *) snd.sig, SND_SOF_LOGS_SIG, SND_SOF_LOGS_SIG_SIZE)) {
 		log_err("Invalid ldc file signature.\n");
+		config = NULL;
 		return -EINVAL;
 	}
 
 	if (global_config->version_fw && /* -n option */
 	    !global_config->dump_ldc) {
 		ret = verify_ldc_checksum(global_config->logs_header->version.src_hash);
-		if (ret)
+		if (ret) {
+			config = NULL;
 			return ret;
+		}
 	}
 
 	/* default logger and ldc_file abi verification */
@@ -1073,6 +1087,7 @@ int convert(struct convert_config *config)
 			SOF_ABI_VERSION_MAJOR(snd.version.abi_version),
 			SOF_ABI_VERSION_MINOR(snd.version.abi_version),
 			SOF_ABI_VERSION_PATCH(snd.version.abi_version));
+		config = NULL;
 		return -EINVAL;
 	}
 
@@ -1081,7 +1096,10 @@ int convert(struct convert_config *config)
 	count = fread(&uids_hdr, sizeof(uids_hdr), 1, config->ldc_fd);
 	if (!count) {
 		log_err("Error while reading uuids header from %s.\n", config->ldc_file);
-		return -ferror(config->ldc_fd);
+		int result = -ferror(config->ldc_fd);
+
+		config = NULL;
+		return result;
 	}
 	if (strncmp((char *)uids_hdr.sig, SND_SOF_UIDS_SIG,
 		    SND_SOF_UIDS_SIG_SIZE)) {
@@ -1118,5 +1136,6 @@ int convert(struct convert_config *config)
 	ret = logger_read();
 out:
 	free(config->uids_dict);
+	config = NULL;
 	return ret;
 }
--- a/tools/logger/logger.c
+++ b/tools/logger/logger.c
@@ -438,7 +438,7 @@ int main(int argc, char *argv[])
 
 	if (config.input_std) {
 		config.in_fd = stdin;
-	} else if (baud) {
+	} else if (baud && config.in_file) {
 		config.serial_fd = configure_uart(config.in_file, baud);
 		if (config.serial_fd < 0) {
 			ret = -config.serial_fd;
@@ -507,6 +507,9 @@ out:
 
 	if (config.version_fd)
 		fclose(config.version_fd);
+
+	if (config.in_file)
+		fclose(config.in_file);
 
 	return ret;
 }


tools/logger/convert.c
```diff
@@ -346,7 +347,8 @@ static inline void print_table_header(void)
 			timestamp_width(global_config->time_precision);
 		snprintf(time_fmt, sizeof(time_fmt), "%%-%ds(us)%%%ds  ",
 			 ts_width, ts_width);
-		fprintf(out_fd, time_fmt, " TIMESTAMP", "DELTA");
+		sprintf(time_fmt, "%s", time_fmt, " TIMESTAMP", "DELTA");
+		fprintf(out_fd, time_fmt);
```
paulstelian97, line 351: This is explicitly bad -- what if we end up with a format specifier inside "time_fmt"? Also unsafe sprintf without length specifier AND having a string as both input and output.

tools/logger/convert.c
```diff
@@ -517,7 +520,7 @@ static void print_entry_params(const struct log_entry_header *dma_log,
 			raw_output && strlen(ids) ? "-" : "",
 			ids);
 		if (time_precision >= 0)
-			fprintf(out_fd, time_fmt,
+			fprintf(out_fd, "%s", time_fmt,
 				to_usecs(dma_log->timestamp - timestamp_origin), dt);
```
paulstelian97, line 524: time_fmt seems to itself be a format specifier; you're changing semantics here unfortunately.

tools/logger/convert.c
```diff
@@ -961,7 +964,8 @@ static int verify_ldc_checksum(const uint32_t ldc_sum)
 	count = fread(&ver, sizeof(ver), 1, global_config->version_fd);
 	if (!count) {
 		log_err("Error while reading %s.\n", global_config->version_file);
-		return -ferror(global_config->version_fd);
+		int result = -ferror(global_config->version_fd);
+		return result;
```
paulstelian97, line 968: Pointless refactor.

tools/logger/convert.c
```diff
@@ -1016,10 +1020,12 @@ static int dump_ldc_info(void)
 
 	while (remaining > 0) {
 		name = format_uid_raw(&uid_ptr[cnt], 0, 0, false, false);
+		if(!name)
+			name = "-"
 		uid_addr = get_uuid_key(&uid_ptr[cnt]);
 		fprintf(out_fd, "\t%p  %s\n", (void *)uid_addr, name);
 
-		if (name) {
+		if (name && name != "-") {
```
paulstelian97, line 1028: You're making a value special without it needing to be special. I don't agree at all with this specific change.

tools/logger/convert.c
```diff
@@ -1045,7 +1051,8 @@ int convert(struct convert_config *config)
 	count = fread(&snd, sizeof(snd), 1, config->ldc_fd);
 	if (!count) {
 		log_err("Error while reading %s.\n", config->ldc_file);
-		return -ferror(config->ldc_fd);
+		int result = -ferror(config->ldc_fd);
+		return result;
```
paulstelian97, line 1055: Pointless refactor.

tools/logger/convert.c
```diff
@@ -1081,7 +1088,8 @@ int convert(struct convert_config *config)
 	count = fread(&uids_hdr, sizeof(uids_hdr), 1, config->ldc_fd);
 	if (!count) {
 		log_err("Error while reading uuids header from %s.\n", config->ldc_file);
-		return -ferror(config->ldc_fd);
+		int retult = -ferror(config->ldc_fd);
+		return result;
```
paulstelian97, line 1092: Compiler error (due to typo) and it's a pointless refactor anyway.

paulstelian97 (changes requested):
Many of these changes are either pointless or outright breaking functionality. I am strongly against them. If we thin out this PR to only the useful changes it would be less than a quarter in size to be honest.

tools/logger/convert.c
aborisovich: This is not pointless refactor. Calling return on ferror causes a dangling pointer to global_config, which is no longer in scope. And static analysis tools reports that fact.

paulstelian97: This specific change would also have the exact same dangling pointer because it literally compiles to the same thing if you have at least -O1...

tools/logger/convert.c
paulstelian97: It is pointless because it fixes absolutely nothing here.

tools/logger/convert.c
aborisovich: Ok, so please propose an alternative solution. What do we do if name is NULL?
Skip printing at all?

tools/logger/convert.c
paulstelian97: Yes here specifically the new code is the wrong one -- time_fmt is supposed to be a format specifier (computed on line 347 of the original code). You are NOT supposed to escape it here, because it's not accidental but intentional.

Further, your new line 351 passes a format string without arguments (time_fmt will have format specifiers), which INTRODUCES the very vulnerability you're mentioning.

marc-hb: @paulstelian97 is correct, this change will break the (deprecated) sof-logger. Good catch thanks.

Static analyzers are not always right.

marc-hb (changes requested):
> @paulstelian97 I appreciate your review. Anyway, I have to dismiss it now. 

Don't break sof-logger in any branch. Either it's used and you cannot break it, or it's not used in that branch and breaking it is pointless (and you should stop scanning it).

Static analyzers are not always right. Klocwork itself knows this and has a documented process to mark false positive. Use that instead of randomly patching the code.

tools/logger/convert.c
marc-hb: `time_fmt` is (safely until proven otherwise) constructed at the line before but Klocwork is not capable of analyzing that. In this case it's only capable of a dumb check. This is a classic case of a false positive.

tools/logger/convert.c
marc-hb: This error is fatal, the tool exits when it happens. KW is not smart enough to see that, this is yet another false positive.

Indeed this change makes absolutely no difference, I hope KW is not so dumb to be deceived by it.

tools/logger/convert.c
marc-hb: Probably the only reason `name` could be NULL is because some `malloc` failed. In that very unlikely case the program should just stop and not print garbage. Please use some kind of `assert` either here or probably better inside `log_vasprintf()` (in case KW can understand that)

marc-hb:
I just noticed this code does not even compile: https://github.com/thesofproject/sof/actions/runs/3639665715/jobs/6143338384

aborisovich:
Converted to draft, fixing typos and warnings. Pushed in hurry as we are in hurry with release.
Thank @marc-hb for review I'll add asserts and fix it.

Now, I would like to elaborate on following comments:

> Compiler error (due to typo) and it's a pointless refactor anyway.

> Pointless refactor.

by @paulstelian97 

> Static analyzers are not always right. Klocwork itself knows this and has a documented process to mark false positives. Use that KW process instead of randomly patching the code.
by @marc-hb 

And those are related specifically to this case:
image

I assume you know what a variable scope is.
I also assume that you understand the difference in C language between "something is working" and "something is working but in language terms is undefined behavior". I also assume, that you know what a "dangling pointer" is.
Ok, lets start from a reference to what [return statement](https://learn.microsoft.com/en-us/cpp/c-language/return-statement-c?view=msvc-170 ) does.
I will quote from those docs:
image
This:

> ends the execution of the function

means that it ends **function scope**.

Now, let's go to the case where return statement returns the expression (emphasis mine):
image


This means, that the **function scope** ends with return statement.
Stack frame has additional slot in bytes reserved only for the value returned by the function. Let's illustrate it like so:
image

This means following:

```
int func_a()
{
    return func_b();
}
 int func_b()  // THIS IS NO LONGER SCOPE OF A. Expression of func_b is evaluated and return value placed into the stack trace return window.
{
    return X;
}
```
And more graphically:
image

This means, that function arguments passed to func_b, if they are pointer types which are copied by value (in C++ there are references whose lifetime is **expanded** in this situation!), then those are **dangling pointers** - a pointers to local variables of a func_a that are outside of the scope during execution of func_b.
More graphically:
image

This is what happens here in:
image

You pass a `global_config->version_fd` to ferror function what results in a dangling pointer.
Please undestand that:
```
int a = ferror(...);
return a;
```
is not the same as:
`return ferror(...);`

Compiler might optimize it if `ferror` is inlined or if `ferror` has no arguments but also MAY NOT optimize it.
Especially in situaltions like this one.

Next time you write **something not very kind** @paulstelian97 please make sure **you know what you talk about**.
Same @marc-hb. Please **make sure you know what you are saying**. Some things are not as trivial as they look.

tools/logger/convert.c
kfrydryx: The error in this return rather comes from line 1048, `config->logs_header = &snd;`. So on return, we're left with `config->logs_header` (config is passed here as a ptr from scope higher) and local variable address (now dangling) from local scope.

paulstelian97:
The problem with your example here is that

```
int a = ferror(...);
return a;
```

CAN be optimized to generate the exact same code as

```
return ferror(...);
```

The C standard doesn't guarantee any local variable actually _gets_ a stack frame allocation (unless you take its address and that address isn't optimized out for some different reason)

Also... "dangling pointer" I feel like it doesn't mean what you think it means. version_fd is a field of the thing pointed at by "global_config", which doesn't get automatically freed in any way (global_config doesn't point to anywhere on the stack), as such version_fd remains valid (if it's a pointer in the first place anyway). You can't have dangling pointers if the memory they point to isn't freed (and thus remains valid) -- a pointer to valid memory cannot by definition be a dangling pointer.

The return statement ends the execution of the function after the expression has been evaluated (so it knows what value to return).

aborisovich:
@marc-hb 

> Probably the only reason name could be NULL is because some malloc failed. In that very unlikely case the program should just stop and not print garbage. Please use some kind of assert either here or probably better inside log_vasprintf() (in case KW can understand that)

Default GCC asserts does not work with -NDEBUG flag and I would like it to work with production code when memory allocation fails. Including sof/debug/panic.h fails also since logger does not have architecture defined (its not a platform so...).
Any suggestions on error handling here?

paulstelian97:
This is a tool, so for error handling you can probably print an error message and call the exit() function.

I have looked at your past commit history, this is your first time in C code (unless I'm missing some old work) on this repo. Sorry for making the "welcome" be so rough; I just have to point out that C doesn't behave pretty much at all like Python and there's a LOT of gotchas that seem to have stumped you when sending this pull request.

tools/logger/convert.c
paulstelian97: Now the issue is still weird -- why do you set a local variable to NULL? That thing can get optimized out and you _still_ get the same code. If "config" is a global variable (so that this change does become meaningful) disregard.

tools/logger/logger.c
```diff
@@ -438,7 +438,7 @@ int main(int argc, char *argv[])
 
 	if (config.input_std) {
 		config.in_fd = stdin;
-	} else if (baud) {
+	} else if (baud && config.in_file) {
```
paulstelian97, line 441: I think this is the only meaningfully _good_ change I can see.

paulstelian97 (changes requested):
Still not great, but a bit better than last time. See individual comments.

aborisovich:
> I have looked at your past commit history, this is your first time in C code (unless I'm missing some old work -- EDIT: I did find a few basically trivial commits such as a missing cast that did get accepted) on this repo. Sorry for making the "welcome" be so rough; I just have to point out that C doesn't behave pretty much at all like Python and there's a LOT of gotchas that seem to have stumped you when sending this pull request.

@paulstelian97 what are you talking about?
I'm developing MTL firmware for more than a year now in C... I don't know how you checked your commit history.
I guess it had a lot of squashes during upstreams to you know, commit history is not everything. Better not judge people by the book.

aborisovich:
> Anyway you claim that some things are meaningful when my experience says otherwise so I would like you to tell me exactly why those things _are_ meaningful.

I'll gladly elaborate on this on Saturday in my free time since we need to finish release asap and production world will not be happy that I prolong the process due to scientific debates :-)
Let's continue with the review meantime.

aborisovich:
Next iteration:
* added name != NULL check using logging and exit() as @paulstelian97 suggested
* added config = NULL assignment to fix errors reported by static analysis tool that reference to local is returned through argument

paulstelian97 (commented):
I guess I'm gonna allow these changes (there's no longer any actively-breaking change, just two changes that prevent crashes and a good chunk of changes that do nothing). They are still meaningless (you trust the tool more than the experts, which is why I can't outright approve these changes) AND have not truly made an attempt to explain why the tool is right over the experts for most of these changes. I will no longer _block_ the merge however.

I suggest you use more well known static analysis tools. Coverity is an example. And when a tool reports an error, I suggest you get a _full_ understanding of what it's saying to be able to evaluate whether it's right or wrong, rather than trust it (I always do that, rather than blindly trust the tools -- I do my best to understand its point and if I fail to understand I just... do more research to try to figure it out, rather than trust that it's right). No tool gets all things right (the best tools will just never _miss_, but _will_ report errors that aren't errors)

aborisovich:
> I guess I'm gonna allow these changes (there's no longer any actively-breaking change, just two changes that prevent crashes and a good chunk of changes that do nothing). They are still meaningless (you trust the tool more than the experts, which is why I can't outright approve these changes) AND have not truly made an attempt to explain why the tool is right over the experts for most of these changes. I will no longer block the merge however.

You are  not going to allow anything. Your review had been dismissed and this is not SOF main branch but a product release branch. You are not only maintainer here and I think @lgirdwood should do the review. You should certainly be more careful with what you say and do as a maintainer of this project. Further comments on my developer career I will take as an offence.

paulstelian97:
I will not make comments on your developer career, that I can agree with. And while I don't maintain this branch, I believe that marc does, and Liam isn't the _only_ maintainer that can deal with these branches.

I'm cool with Intel guys commenting and reviewing patches on i.MX stable branches just fine, and WILL take into account what they are saying. You don't seem to take stuff into account from me OR from marc, so I strongly doubt Liam will be of a different opinion. The only real difference is he'll be more polite than myself in saying most of these changes are either pointless or wrong.

aborisovich:
After consultations with team turns out that logger is not part of the release for MTL, closing

marcinszkudlinski:
I'm closing this PR
after discussions we reailzed that logger is not a tool intended for MTL, so is not a part od MTL-002-stable release

plbossart:
> Next time you write **something not very kind** @paulstelian97 please make sure **you know what you talk about**.

Unacceptable tone @aborisovich. This is not how we work with partners and colleagues.

tools/logger/convert.c
paulstelian97: Function parameters are still local variables actually.

marc-hb:
>  Please understand that:
```
int a = ferror(pointer_to_local_stack);
return a;
```
> is not the same as:
```
return ferror(pointer_to_local_stack);
```

This is an interesting C quizz. Please correct me but I think and hope your long and illustrated explanation can be summarized with just these 3 words: "tail call optimization"? I would naively trust compilers to "do the Right Thing" and NOT perform any tail call optimization when the function argument contains any reference to the local stack but maybe that's too naive. C is a terrifying minefield of unintuitive and undefined behaviors so maybe this is yet another one of them?

Also, maybe most compilers are smart enough but a few are not (or they are buggy) and Klocwork is being very conservative and erring on the "false positive" side? It would suit its role.

I'm also a bit sceptical that a (too?) simple intermediate variable trick is enough to defeat tail call optimization, now _that_ seems a bit too naive too?

**Anyway `config` is NOT a pointer to the local stack in most (all?) cases found by KW so these were definitely false positives.** As noted by @paulstelian97, nothing is ever dangling here because it always points to something in a higher scope.

I see sof-logger is removed in that branch by #6754 and that's OK but you still need to fix the KW process because there will always be other false positives that must be disposed (after very careful team review) and it won't always be possible to change or delete the code to avoid them.

Static analyzers _always_ produce false positives which is why they all have a process to ignore some warnings and KW is no exception.

> I'll gladly elaborate on this on Saturday in my free time since we need to finish release asap and production world will not be happy that I prolong the process due to scientific debates :-)

I'm looking forward to Saturday as long as we all focus 100% on the technical questions and stop all personal comments! https://www.sofproject.org/community/

marc-hb:
Interesting experimentations:

https://quuxplusone.github.io/blog/2021/01/09/tail-call-optimization/

tl;dr: some compilers perform tail call optimizations but they stop when things "get too complicated", likely out of fear of breaking things.

https://godbolt.org/z/vcY3v9 - See assembly generated "live" by a few compilers

marc-hb:
BTW different static analyzers find different bugs and the CMake maintainer in Zephyr has submitted a pull request to support any of them generically:
- https://github.com/zephyrproject-rtos/zephyr/pull/52671

This is very promising, looking forward to it!

Different static analyzers also emit different _false positives_ that may conflict with each other: write the code one way and analyzer 1 complains, write it differently and now analyzer 5 complains! "Static analyzer whack-a-mole" gets nowhere fast. It's OK to simplify the code to help _all_ analyzers (and humans!) understand it. It's not OK to make it more complicated to silence a single analyzer. That's why many static analyzers offer a way to filter out false positives.

marc-hb:
Fixes submitted to the main branch by @kv2019i (thanks!)
- #6844, 
- #6846,
- #6847,
- #6848

marc-hb:
> Default GCC asserts does not work with -NDEBUG flag and I would like it to work with production code when memory allocation fails. Including sof/debug/panic.h fails also since logger does not have architecture defined (its not a platform so...).
Any suggestions on error handling here?

Good catch. `abort()` is a good alternative, see Linux demo in https://github.com/thesofproject/sof/pull/6848#issuecomment-1358632925

`exit(non-zero)` is good when failing input sanitization or other situations when the program is still "in control". Out of memory is an "out of control" situation (and will practically never happen with sof-logger)

marc-hb:
Making global_config truly global with a minimal number of lines changed in:
- https://github.com/thesofproject/sof/pull/6862

PR #6746, opened 2022-12-08
memory: intel: increase task stack size for specific modules

johnylin76:
Larger stack context stack size is required for RTNR and GOOGLE_RTC_AUDIO integration on Intel platforms.

### DIFF
--- a/src/platform/intel/cavs/include/cavs/lib/memory.h
+++ b/src/platform/intel/cavs/include/cavs/lib/memory.h
@@ -68,7 +68,12 @@
 #define HEAP_BUF_ALIGNMENT		PLATFORM_DCACHE_ALIGN
 
 /** \brief EDF task's default stack size in bytes. */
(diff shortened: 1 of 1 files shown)


cujomalainey (approved):

cujomalainey:
@lgirdwood on zephyr is there a way we can bundle this information with the component so we can check the stack of the thread is compatible before we even try to instantiate it?

lgirdwood:
> @lgirdwood on zephyr is there a way we can bundle this information with the component so we can check the stack of the thread is compatible before we even try to instantiate it?

Good question, @mwasko @marcinszkudlinski do you guys know ?

lgirdwood (approved):
IIUC, this could be planned for the manifest and configured via the toml config for each module.

marcinszkudlinski:
@cujomalainey there's a stack_size field in a loadable module manifest, so there will be a simple way to request a desired stack space.

PR #6751, opened 2022-12-08
Ipc4 : add ipc4 support in smart amp test

RanderWang:
Chao cooperates with me to develop this feature. Now it works with nocodec smart amp topology

### DIFF
--- a/src/include/sof/samples/audio/smart_amp_test.h
+++ b/src/include/sof/samples/audio/smart_amp_test.h
@@ -12,6 +12,12 @@
 #include <sof/audio/component.h>
 #include <sof/audio/data_blob.h>
 
+#if CONFIG_IPC_MAJOR_4
+#include <ipc4/base-config.h>
+
+#define SOF_SMART_AMP_FEEDBACK_QUEUE_ID        1
+#endif
+
 #define SMART_AMP_MAX_STREAM_CHAN   8
 
 /** IPC blob types */
@@ -81,10 +87,31 @@ typedef int(*smart_amp_proc)(struct comp_dev *dev,
  */
 
 struct sof_smart_amp_config {
+#if CONFIG_IPC_MAJOR_4
+	struct ipc4_base_module_cfg base;
+	uint16_t num_input_pin_fmt;
+	uint16_t num_output_pin_fmt;
+	uint8_t reserved[8];
+	/* length of data after output_pin_fmt */
+	uint32_t priv_param_len;
+	struct ipc4_input_pin_format input_pin_fmt[2];
+	struct ipc4_output_pin_format output_pin_fmt;
+#else
 	uint32_t size;
+#endif
 	uint32_t feedback_channels;
 	int8_t source_ch_map[PLATFORM_MAX_CHANNELS];
 	int8_t feedback_ch_map[PLATFORM_MAX_CHANNELS];
 };
+
+#if CONFIG_IPC_MAJOR_4
+
+enum smart_amp_config_params {
+	SMART_AMP_SET_MODEL = 1,
+	SMART_AMP_SET_CONFIG = 2,
+	SMART_AMP_GET_CONFIG = 3
+};
+
+#endif
 
 #endif /* __SOF_AUDIO_SMART_AMP_H__ */
--- a/src/samples/audio/smart_amp_test.c
+++ b/src/samples/audio/smart_amp_test.c
@@ -276,7 +276,7 @@ static int smart_amp_verify_params(struct comp_dev *dev,
 
 	ret = comp_verify_params(dev, BUFF_PARAMS_CHANNELS, params);
 	if (ret < 0) {
-		comp_err(dev, "volume_verify_params() error: comp_verify_params() failed.");
+		comp_err(dev, "smart_amp_verify_params() error: comp_verify_params() failed.");
 		return ret;
 	}
 
--- a/src/samples/audio/smart_amp_test.c
+++ b/src/samples/audio/smart_amp_test.c
@@ -11,6 +11,14 @@
 #include <sof/ipc/msg.h>
 #include <sof/ut.h>
 
+#ifndef __ZEPHYR__
+#include <rtos/mutex.h>
+#endif
+
+#if CONFIG_IPC_MAJOR_4
+#include <ipc4/module.h>
+#endif
+
 static const struct comp_driver comp_smart_amp;
 
 LOG_MODULE_REGISTER(smart_amp_test, CONFIG_SOF_LOG_LEVEL);
@@ -32,6 +40,8 @@ struct smart_amp_data {
 	struct comp_buffer *feedback_buf; /**< feedback source buffer */
 	struct comp_buffer *sink_buf; /**< sink buffer */
 
+	struct k_mutex lock; /**< protect feedback_buf updated */
+
 	smart_amp_proc process;
 
 	uint32_t in_channels;
@@ -42,10 +52,12 @@ static struct comp_dev *smart_amp_new(const struct comp_driver *drv,
 				      const struct comp_ipc_config *config,
 				      const void *spec)
 {
-	struct comp_dev *dev = NULL;
+#if CONFIG_IPC_MAJOR_3
 	const struct ipc_config_process *ipc_sa = spec;
-	struct smart_amp_data *sad = NULL;
-	struct sof_smart_amp_config *cfg;
+#endif
+	const struct sof_smart_amp_config *cfg;
+	struct smart_amp_data *sad;
+	struct comp_dev *dev;
 	size_t bs;
 
 	dev = comp_alloc(drv, sizeof(*dev));
@@ -64,8 +76,15 @@ static struct comp_dev *smart_amp_new(const struct comp_driver *drv,
 	if (!sad->model_handler)
 		goto sad_fail;
 
+	k_mutex_init(&sad->lock);
+
+#if CONFIG_IPC_MAJOR_4
+	cfg = spec;
+	bs = sizeof(struct sof_smart_amp_config);
+#else
 	cfg = (struct sof_smart_amp_config *)ipc_sa->data;
 	bs = ipc_sa->size;
+#endif
 
 	if ((bs > 0) && (bs < sizeof(struct sof_smart_amp_config))) {
 		comp_err(dev, "smart_amp_new(): failed to apply config");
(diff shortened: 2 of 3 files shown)


aiChaoSONG:
The IPC4 smart_amp feature is verified with topology and linux kernel patches, which I will send patch later.

RanderWang:
will fix compiling issue for mutex. It is only supported by zephyr

src/samples/audio/smart_amp_test.c
```diff
@@ -32,20 +36,65 @@ struct smart_amp_data {
 	struct comp_buffer *feedback_buf; /**< feedback source buffer */
 	struct comp_buffer *sink_buf; /**< sink buffer */
 
+#ifndef __ZEPHYR__
+		struct k_spinlock lock; /**< protect feedback_buf updated */
+		k_spinlock_key_t key;
+#else
+		struct k_mutex lock;
```
lyakh, line 43: wrong indentation

src/samples/audio/smart_amp_test.c
```diff
@@ -86,6 +142,212 @@ static struct comp_dev *smart_amp_new(const struct comp_driver *drv,
...
+		source_buffer = container_of(blist, struct comp_buffer, sink_list);
+		buffer_c = buffer_acquire(source_buffer);
+		if (IPC4_SINK_QUEUE_ID(buffer_c->id) == SOF_SMART_AMP_FEEDBACK_QUEUE_ID) {
+			sa_lock(sad);
+			sad->feedback_buf = source_buffer;
+
+			buffer_c->stream.channels = sad->config.feedback_channels;
+			buffer_c->stream.rate = sad->config.base.audio_fmt.sampling_frequency;
+			buffer_release(buffer_c);
+			sa_unlock(sad);
```
lyakh, line 316: locks should be taken and released symmetrically. `buffer_acquire()` can potentially take a lock if the buffer is shared, so `buffer_unlock()` should come after `sa_unlock()`.  Also below you acquire the buffer inside `sa_lock()` blocks. I know, that this is just a test, so probably it cannot race, right? Still it would be better to have that consistent to not even look like we can get an AB-BA deadlock

lyakh (changes requested):

src/samples/audio/smart_amp_test.c
RanderWang: thanks!

src/samples/audio/smart_amp_test.c
RanderWang: good catch!  thanks! refined it.

src/samples/audio/smart_amp_test.c
```diff
@@ -32,20 +36,65 @@ struct smart_amp_data {
...
+
+static void sa_unlock(struct smart_amp_data *sad)
+{
+	k_mutex_unlock(&sad->lock);
+}
+
+static void sa_lock_init(struct smart_amp_data *sad)
+{
+	k_mutex_init(&sad->lock);
+}
```
lgirdwood, line 67: Can we revisit this in a subsequent PR (as there are others doing the same) and use k_mutex_ API directly (where we will define this for xtos as a spinlock).

src/ipc/ipc4/helper.c
```diff
@@ -478,13 +478,18 @@ int ipc_comp_disconnect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 	 * pipelines need to be on the same core.
 	 */
 	irq_local_disable(flags);
-	pipeline_disconnect(src, buffer, PPL_CONN_DIR_COMP_TO_BUFFER);
-	pipeline_disconnect(sink, buffer, PPL_CONN_DIR_BUFFER_TO_COMP);
+	if (src->pipeline->status < COMP_STATE_ACTIVE)
```
lgirdwood, line 481: Can you circle back and add inline comments here for this so it's easy to understand.

src/ipc/ipc4/helper.c
```diff
@@ -478,13 +478,18 @@ int ipc_comp_disconnect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
+
+	if (sink->pipeline->status < COMP_STATE_ACTIVE)
+		pipeline_disconnect(sink, buffer, PPL_CONN_DIR_BUFFER_TO_COMP);
+
 	ret = comp_unbind(src, bu);
 	ret1 = comp_unbind(sink, bu);
 	irq_local_enable(flags);
 
-	buffer_free(buffer);
+	if (src->pipeline->status < COMP_STATE_ACTIVE && sink->pipeline->status < COMP_STATE_ACTIVE)
```
lgirdwood, line 491: ditto - needs comments

lgirdwood (approved):
Some of the logic is not obvious for beginners, can you create a new PR where you add more comments to the ipc4 helper in general.

src/samples/audio/smart_amp_test.c
```diff
@@ -555,14 +830,22 @@ static const struct comp_driver comp_smart_amp = {
...
+		.create			= smart_amp_new,
+		.free			= smart_amp_free,
+		.params			= smart_amp_params,
+		.prepare		= smart_amp_prepare,
+#if CONFIG_IPC_MAJOR_4
+		.set_large_config	= smart_amp_set_large_config,
+		.get_large_config	= smart_amp_get_large_config,
+		.get_attribute		= smart_amp_get_attribute,
+		.bind			= smart_amp_bind,
+		.unbind			= smart_amp_unbind,
```
ranj063, line 842: @RanderWang I keep repeating this every time a module is converted to IPC4. bind/unbind are not compatible with the module adapter interface. If we keep adding this si every new component with IPC4, it will take us forever to deprecate comp_driver. Can we please find a way to not do this and make our own lives difficult later on?

RanderWang: @ranj063 this interface is efficient a way to deal with event.  Without it , what I can figure out is that in each smart copy or smart trigger we need to check whether there is a feedback buffer connected or disconnected. I don't like this solution. If we can add a event processor for each module, we can also support this.

src/samples/audio/smart_amp_test.c
RanderWang: sure, thanks

src/ipc/ipc4/helper.c
RanderWang: sure, thanks

RanderWang:
add comments for unbind

src/samples/audio/smart_amp_test.c
ranj063: @RanderWang don't we have notifications for this?

src/samples/audio/smart_amp_test.c
lyakh: AB-BA is still here: in this function buffer-acquisition is outside of `sa_lock()`, and in `smart_amp_trigger()` it's the other way round. Maybe you could just take the `sa_lock()` outside of the loop here?

src/samples/audio/smart_amp_test.c
lgirdwood: ok, since we are updating the PR due to other issues, let fix this one too. Please add the k_mutex() calls to xtos/include/rtos/spinlock.h (and map them to use spinlocks).

src/include/sof/samples/audio/smart_amp_test.h
```diff
@@ -81,10 +87,31 @@ typedef int(*smart_amp_proc)(struct comp_dev *dev,
...
 
 struct sof_smart_amp_config {
+#if CONFIG_IPC_MAJOR_4
+	struct ipc4_base_module_cfg base;
+	uint16_t num_source_pin_fmt;
+	uint16_t num_sink_pin_fmt;
+	uint8_t reserved[8];
+	/* length of data after sink_pin_fmt */
+	uint32_t priv_param_len;
+	struct ipc4_input_pin_format source_pin_fmt[2];
```
aiChaoSONG, line 97: For a widget, sink pin is the input pin of the widget, and source pin is the output pin of the widget. I use the terms `sink` and `source` because they are already used in widget queue allocation in the kernel, and we have num_sink_pins (input) and num_source_pins (output) topology token. 

So the structure here for IPC4 should be:
```
uint16_t num_sink_pin_fmt; /* number of input pin formats */
uint16_t num_source_pin_fmt; /* number of output pin formats */
uint8_t reserved[8];
/* length of data after sink_pin_fmt */
uint32_t priv_param_len;
struct ipc4_input_pin_format sink_pin_fmt[2];
struct ipc4_output_pin_format source_pin_fmt;
```
of course, the firmware doesn't need to be the same as linux, you can use num_input_pin_fmt/num_output_pin_fmt

src/samples/audio/smart_amp_test.c
lyakh: @lgirdwood are we sure that this is safe? XTOS also has a real thread context, IIRC it's used for EDF tasks. Are we sure that we don't have code in SOF, that maps to an EDF task in XTOS and uses a mutex for locking? Or would it be safe to take a spinlock and potentially go to sleep in such a task?

src/samples/audio/smart_amp_test.c
RanderWang: @ranj063  sure. If it is not fit for module_adapter, we need to remove comp_bind interface and use notification to replace it (a new notification like MODULE_BIND/UNBIND). I found that comp_bind is used by mixin_out. I don't want to mix the rework with this smart_amp_test PR. Can I submit the rework in another PR ?

src/include/sof/samples/audio/smart_amp_test.h
RanderWang: It is a common usage that source is input and sink is output in fw, so I want to align with it

src/samples/audio/smart_amp_test.c
RanderWang: yes, it is a potential risk

src/samples/audio/smart_amp_test.c
RanderWang: do we need to maintain XTOS ? or switch to zephyr completely ?  if so we can do nothing for k_mutex for XTOS

src/include/sof/samples/audio/smart_amp_test.h
RanderWang: it is common use sink for output in fw, do we need to use sink_pin for input ?

src/samples/audio/smart_amp_test.c
kv2019i: @RanderWang Now I get the problem. Maybe define "struct k_mutex" so that there's both struct k_spinlock and k_spinlock_key_t as members of it when you build for XTOS. So then you can implement k_mutex_lock() with k_spin_lock() and store the key in XTOS version of "struct k_mutex". Would this work?

src/samples/audio/smart_amp_test.c
lyakh: right, let's get it correct then. It should be ok to have `sa_lock()` outside of buffer locking since with Zephyr they're both taking mutexes and with XTOS - spinlocks

ranj063: @RanderWang can you do this in prepare() instead of bind?

RanderWang: @ranj063 we can't do it in prepare for playback stream since the feedback buffer is shared by capture feedback stream, so it can be available only when the capture feedback stream is started.

RanderWang: I will submit a patch later that remove bind/unbind interface and use notifier to process bind/unbind event. please check https://github.com/thesofproject/sof/pull/6751#discussion_r1046773321

src/samples/audio/smart_amp_test.c
RanderWang: @kv2019i @lgirdwood  thanks very much !  completely address my concern.

RanderWang:
update mutex support. Thanks!

mengdonglin:
> update mutex support. Thanks!

@lgirdwood @kv2019i @lyakh Can you help review the updated PR?

lgirdwood (approved):

kv2019i (approved):
I'm not familiar with the existing smart-amp-test module, so cannot fully review how safe the locking changes are, but the changes look ok. The conditional disconnect starts to look complex, this may need further work, but can be done in separate PRs.

lgirdwood:
@wszypelt can you check CI, I dont think we are testing dummy smart amp today ? (unless I'm out of date with the test matrix)

src/ipc/ipc4/helper.c
```diff
@@ -478,13 +478,29 @@ int ipc_comp_disconnect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
+	 * , a buffer is shared between two streams. One of the two streams will be stopped
+	 * first and will send a unbind ipc msg to the disconnect buffer which can result to
+	 * null pointer error for active stream if the buffer is disconnected from it. The buffer
+	 * will be disconnected when the active pipeline is deleted.
+	 */
+	if (src->pipeline->status < COMP_STATE_ACTIVE)
+		pipeline_disconnect(src, buffer, PPL_CONN_DIR_COMP_TO_BUFFER);
+
+	if (sink->pipeline->status < COMP_STATE_ACTIVE)
+		pipeline_disconnect(sink, buffer, PPL_CONN_DIR_BUFFER_TO_COMP);
```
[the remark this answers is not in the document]
ranj063: @RanderWang why is this needed? ipc_comp_disconnect() will only be called when 2 connected components from separate pipelines are disconnected. You should disconnecte and unbind here.

src/audio/copier/copier.c
```diff
@@ -887,7 +887,7 @@ static int copier_copy(struct comp_dev *dev)
 		sink_c = buffer_acquire(sink);
 		sink_dev = sink_c->sink;
 		processed_data.sink_bytes = 0;
-		if (sink_dev->state == COMP_STATE_ACTIVE) {
+		if (sink_dev && sink_dev->state == COMP_STATE_ACTIVE) {
```
ranj063, line 890: @RanderWang this can never happen, right? During comp_disconnect, we have disabled interrupts. So it is not possible that the buffer is dsconnected from the sink dev but it hasn't been deleted from the list and freed.

src/samples/audio/smart_amp_test.c
```diff
@@ -504,7 +504,7 @@ static int smart_amp_verify_params(struct comp_dev *dev,
 
 	ret = comp_verify_params(dev, BUFF_PARAMS_CHANNELS, params);
 	if (ret < 0) {
-		comp_err(dev, "volume_verify_params() error: comp_verify_params() failed.");
+		comp_err(dev, "smart_amp_verify_params() error: comp_verify_params() failed.");
```
ranj063, line 507: can you please reorder the patches in the PR to do: typos first, pre-req fixes next, new API definitions next followed by the IPC4 support for the smartamp module? spare a thought for CI and git bisect please.

src/samples/audio/smart_amp_test.c
```diff
@@ -86,6 +105,213 @@ static struct comp_dev *smart_amp_new(const struct comp_driver *drv,
...
+
+	/* searching for feedback source buffers */
+	list_for_item(blist, &dev->bsource_list) {
+		source_buffer = container_of(blist, struct comp_buffer, sink_list);
+
+		k_mutex_lock(&sad->lock, K_FOREVER);
+		buffer_c = buffer_acquire(source_buffer);
+		if (IPC4_SINK_QUEUE_ID(buffer_c->id) == SOF_SMART_AMP_FEEDBACK_QUEUE_ID) {
+			sad->feedback_buf = source_buffer;
+			buffer_c->stream.channels = sad->config.feedback_channels;
```
ranj063, line 276: @RanderWang Im not sure where my comment went but I was wondering can be done in prepare instead of bind?

RanderWang: it is at https://github.com/thesofproject/sof/pull/6751#discussion_r1049219248

src/audio/copier/copier.c
RanderWang: in different stream case, not in the same stream.

src/samples/audio/smart_amp_test.c
RanderWang: sure, thanks

src/ipc/ipc4/helper.c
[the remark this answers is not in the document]
RanderWang: consider different streams not different pipelines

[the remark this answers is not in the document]
RanderWang: A buffer is shared between playback stream and capture stream. I meant playback stream is still working, but capture stream is stopped. if the buffer is using by playback stream, the disconnection will result to null pointer issue

RanderWang:
Remove pipeline state check in comp_disconnect function since we have CI issue . Currently we use mutex in unbind function so we can make sure that the buffer is not used by smart_amp. According to Kai, reference count is better than state check. We will remove unbind interface and check the multiple stream issue again.

src/audio/copier/copier.c
RanderWang: I meant the buffer is not disconnected for the state check in comp_disconnect.  Ignore it since I remove the logic

src/samples/audio/smart_amp_test.c
```diff
@@ -42,10 +52,12 @@ static struct comp_dev *smart_amp_new(const struct comp_driver *drv,
...
 				      const void *spec)
 {
-	struct comp_dev *dev = NULL;
+#if CONFIG_IPC_MAJOR_3
 	const struct ipc_config_process *ipc_sa = spec;
+#endif
+	const struct sof_smart_amp_config *cfg;
 	struct smart_amp_data *sad = NULL;
-	struct sof_smart_amp_config *cfg;
+	struct comp_dev *dev = NULL;
```
ranj063, line 60: init is redundant

src/samples/audio/smart_amp_test.c
```diff
@@ -86,6 +105,213 @@ static struct comp_dev *smart_amp_new(const struct comp_driver *drv,
...
+
+	if (cfg_size > *data_size) {
+		comp_err(dev, "smart_amp_get_config(): wrong config size %d",
+			 *data_size);
+	}
+
+	*data_size = cfg_size;
+
+	ret = memcpy_s(data, cfg_size, &sad->config, cfg_size);
+	assert(!ret);
```
ranj063, line 193: i know this is an irrecoverable failure but can we please limit the use of asserts and return error? simplyy ``` return memcpy_s(...)"

RanderWang:
@lgirdwood  updated, pass CI test

src/samples/audio/smart_amp_test.c
```diff
@@ -86,6 +105,213 @@ static struct comp_dev *smart_amp_new(const struct comp_driver *drv,
...
+
+	comp_dbg(dev, "smart_amp_set_large_config()");
+
+	switch (param_id) {
+	case SMART_AMP_SET_MODEL:
+		return ipc4_comp_data_blob_set(sad->model_handler,
+					       first_block,
+					       last_block,
+					       data_offset,
+					       data);
```
ranj063, line 215: can you please condense to 2 lines?

RanderWang: sure, thanks

src/samples/audio/smart_amp_test.c
RanderWang: sure, thanks

src/samples/audio/smart_amp_test.c
RanderWang: sure, thanks

RanderWang:
refined code style

src/samples/audio/smart_amp_test.c
lyakh: We're already doing this in the coherent API - storing the spinlock and the key in the same persistent structure. But that made me rethink that a bit - is that actually safe, because normally you have those keys as local variables - you store your returned spinlock status (CPU interrupt state) and then you restore it from the same variable, making sure nobody else overwrites that variable in the meantime. So having that variable as persistent and available to all might sound risky - you don't want to have another context overwrite that key and then restore a wrong context. But I think it should be safe in this context because another context will only overwrite the key **after** it has taken the spinlock, so the spinlock protects the key too. Just something to be aware of. Alternatively we could wrap `k_mutex_lock()` to return a status, but I don't think we want to do that, we want to use Zephyr APIs directly.

src/samples/audio/smart_amp_test.c
```diff
@@ -86,6 +105,200 @@ static struct comp_dev *smart_amp_new(const struct comp_driver *drv,
...
+
+	switch (type) {
+	case COMP_ATTR_BASE_CONFIG:
+		*(struct ipc4_base_module_cfg *)value = sad->config.base;
+		break;
+	default:
+		return -EINVAL;
+	}
+
+	return 0;
```
lyakh, line 243: nitpick: if you do a revision, you could move this to line 238 instead of `break`

lyakh (approved):

src/samples/audio/smart_amp_test.c
RanderWang: sure, thanks

RanderWang:
update commit message to wrap at line of 75 chars

ranj063 (approved):

PR #6755, opened 2022-12-08
src: avoid restless warn messages from comp_get_data_blob()

johnylin76:
At present codes of rtnr and google-rtc-audio-processing, a reconfigure() function is called per copy() routine to dynamically update config blob in COMP_STATE_ACTIVE state. The logic has no problem in practice.

However, if there are cases that config blob is always empty or arrived late, hundreds of warning message lines per second will be produced by comp_get_data_blob() called from reconfigure(), restlessly complaining that [data blob is not set](https://github.com/thesofproject/sof/blob/main/src/audio/data_blob.c#L84).

This commit adds a pre-check for early-return under such cases to skip comp_get_data_blob() calls with warn messages.

### DIFF
--- a/src/audio/google_rtc_audio_processing.c
+++ b/src/audio/google_rtc_audio_processing.c
@@ -98,6 +98,19 @@ static int google_rtc_audio_processing_reconfigure(struct comp_dev *dev)
 
 	comp_dbg(dev, "google_rtc_audio_processing_reconfigure()");
 
+	if (!comp_is_current_data_blob_valid(cd->tuning_handler) &&
+	    !comp_is_new_data_blob_available(cd->tuning_handler)) {
+		/*
+		 * The data blob hasn't been available once so far.
+		 *
+		 * This looks redundant since the same check will be done in
+		 * comp_get_data_blob() below. But without this early return,
+		 * hundreds of warn message lines are produced per second by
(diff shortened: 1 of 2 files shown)


perahgren (approved):

cujomalainey (approved):
nit brackets not needed

also I think you meant "relentless" but otherwise thanks for fix :)

MingJenTai (approved):

src/audio/google_rtc_audio_processing.c
```diff
@@ -98,6 +98,19 @@ static int google_rtc_audio_processing_reconfigure(struct comp_dev *dev)
...
+	    !comp_is_new_data_blob_available(cd->tuning_handler)) {
+		/*
+		 * The data blob hasn't been available once so far.
+		 *
+		 * This looks redundant since the same check will be done in
+		 * comp_get_data_blob() below. But without this early return,
+		 * hundreds of warn message lines are produced per second by
+		 * comp_get_data_blob() calls until the data blob is arrived.
+		 */
+		return 0;
```
lyakh, line 111: are we sure we wouldn't be hiding a problem here? Maybe it would be better to do something like "warn once" or rate-limited warning?

lgirdwood (approved):
We dont have warn_once() atm, so is better than flooding the logs.

PR #6757, opened 2022-12-09
Audio: Math: Add HiFi3 version implementation of FFT

andrula-song:
Add HiFi3 version implementation of FFT. Test with xtensa
    simulator, we can save about 52.7% cycles for 16bit processing
    and 58.1% for 32bit processing.


andrula-song:
here are the test result of xtensa simulator;
fft

src/math/fft/fft_16.c
```diff
@@ -10,6 +10,54 @@
...
+{
+	int n_rnd = -n - 1;
+
+	if (n >= 0) {
+		/* need saturation handling */
+		output->real = sat_int16((int32_t)input->real << n);
+		output->imag = sat_int16((int32_t)input->imag << n);
+	} else {
+		output->real = sat_int16((((int32_t)input->real >> n_rnd) + 1) >> 1);
+		output->imag = sat_int16((((int32_t)input->imag >> n_rnd) + 1) >> 1);
```
lyakh, line 59: you can move `n_rnd` under the `else` clause - it isn't needed in the `n >= 0` case. Then you could maybe save a couple of cycles by doing
```
int n1 = -n;
int n2 = 1 << (n1 - 1);

output->real = sat_int16(((int32_t)input->real + n2) >> n1);
output->imag = sat_int16(((int32_t)input->imag + n2) >> n1);
```

lgirdwood (commented):
@singalsu pls review hifi.

src/math/fft/fft_16.c
```diff
@@ -8,8 +8,60 @@
...
 #include <sof/common.h>
 #include <sof/math/fft.h>
+
+#ifdef FFT_GENERIC
 #include <sof/audio/coefficients/fft/twiddle_16.h>
 
+/*
+ * Helpers for 16 bit FFT calculation
+ */
+static inline void icomplex16_add(struct icomplex16 *in1, struct icomplex16 *in2,
```
marc-hb, line 18: I think you could also make these functions compatible with const arguments:
```suggestion
static inline void icomplex16_add(const struct icomplex16 *in1, const struct icomplex16 *in2,
```
Same for all the other ones.

andrula-song: Thanks, done.

src/math/fft/fft_16_hifi3.c
```diff
@@ -0,0 +1,139 @@
...
+		 * the shrink we did in the FFT transform.
+		 */
+		in16x4 = (ae_int16x4 *)plan->outb16;
+		out16x4 = (ae_int16x4 *)plan->outb16;
+		n = size >> 1;
+		inu = AE_LA64_PP(in16x4);
+		/* shift 2 samples per loop */
+		for (i = 0; i < n; i++) {
+			AE_LA16X4_IP(sample, inu, in16x4);
+			sample = AE_SLAA16S(sample, len);
```
singalsu, line 122: This is not doing the round in right shift? I don't remember which FFT lengths cause a right shift and I'm not sure how important it's for 16 bit IFFT accuracy bit it seems to be missing.

src/math/fft/fft_32_hifi3.c
```diff
@@ -0,0 +1,120 @@
...
+			AE_S32_L_IP(sample, in, sizeof(struct icomplex32));
+		}
+	}
+
+	/* step 1: re-arrange input in bit reverse order, and shrink the level to avoid overflow */
+	inu = AE_LA64_PP(inx);
+	for (i = 1; i < size; ++i) {
+		AE_LA32X2_IP(sample, inu, inx);
+		sample = AE_SRAA32S(sample, len);
+		out = (ae_int32x2 *)&outx[plan->bit_reverse_idx[i]];
```
singalsu, line 60: Is this type cast necessary? outx seems to be int32x2 type already.

singalsu (commented):
Good work! Two questions:

src/math/fft/fft_32_hifi3.c
andrula-song: yeah, thanks

src/math/fft/fft_16_hifi3.c
andrula-song: would len be a negative integer？ I have done the round and saturation at step 1 for right shift.

lgirdwood (approved):
Approved C parts.

src/math/fft/fft_16_hifi3.c
singalsu: Yep, you are right. Len is always positive, log2() of FFT length.

singalsu (approved):
Looks good, thanks!

kv2019i:
Thanks all, merging. The one alsabat failure is a known one.

PR #6758, opened 2022-12-09
xtensa-build-zephyr.py: objcopy --remove-section .comment zephyr.strip

marc-hb:
Make stripped ELF files compiled by XCC identical across different machines.

Some Xtensa compilers (ab?)use the .ident / .comment section and append the typically absolute and not
reproducible /path/to/the.c file after the usual
compiler ID.

https://sourceware.org/binutils/docs/as/Ident.html

strip --strip-all does not remove the .comment section. Remove ourselves like some gcc test scripts do:

 https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=c7046906c3ae

### DIFF
--- a/scripts/xtensa-build-zephyr.py
+++ b/scripts/xtensa-build-zephyr.py
@@ -734,13 +734,22 @@ def install_platform(platform, sof_platform_output_dir):
 		InstFile(BIN_NAME + ".map"),
 
 		# CONFIG_BUILD_OUTPUT_STRIPPED
-		InstFile(BIN_NAME + '.strip', optional=True),
+		# Renaming ELF files highlights the workaround below that strips the .comment section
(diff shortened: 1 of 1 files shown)


marc-hb:
Note this commit does not affect `.ri` files at all, it only makes finding differences easier.

lgirdwood:
@marc-hb I assume this wont strip the image ELF file (i.e. the file we point gdb to when debugging) ?

marc-hb:
> @marc-hb I assume this wont strip the image ELF file (i.e. the file we point gdb to when debugging) ?

Correct, unstripped files are left unmodified, always have been.

lgirdwood (approved):

lgirdwood:

marc-hb:
We need this for the 2.4 release branch

kv2019i (approved):

PR #6769, opened 2022-12-10
[SKIP CI] .github: cosmetic sparse fixes

marc-hb:
Zero functional change, pure cleanups / minor refactor.

This PR is a subset of rejected #6722, it's the subset that does nothing and only re-organizes the code.

cc:
- #6722


marc-hb:
sparse output is as expected in https://github.com/thesofproject/sof/actions/runs/3662044536/jobs/6190838328

PR #6770, opened 2022-12-11
Created CMakePresets.json for Zephyr application

aborisovich:
Alternative way of building the SOF project as Zephyr application is using CMake overlays. This provides similar functionality to xtensa-build-zephyr.py but is solution native to CMake. In this commit added TGL and MTL builds using Xtensa toolchain. Only elf file is built, integration with Rimage is TODO.

To test (example):
```
cd sof/app
cmake -S . --preset mtl-xtensa-linux
cmake --build --preset mtl-xtensa-linux
```
This will build .elf file without signing.
Rimage integration is TODO.
@marc-hb can you please test it on linux?
What toolchain are NXP platforms using?

### DIFF
--- a/app/CMakePresets.json
+++ b/app/CMakePresets.json
@@ -1,0 +1,171 @@
+{
+	"version": 3,
+	"cmakeMinimumRequired": {
+	  "major": 3,
+	  "minor": 21,
+	  "patch": 0
+	},
+
+	"configurePresets": [
+	{
+		"name": "common-settings",
+		"description": "List of common settings used by all configurations",
+		"hidden": true,
+		"generator": "Ninja",
+		"environment": {
+			"ZEPHYR_BASE": "${sourceDir}/../../zephyr"
+		}
+	},
+	{
+		"name": "mtl-xtensa-windows",
+		"displayName": "MTL Xtensa",
+		"inherits": [ "mtl-xtensa", "xtensa-toolchain-windows" ],
+		"environment": {
+			"XTENSA_TOOLS_VERSION": "RI-2020.5-win32"
+		}
+	},
+	{
+		"name": "mtl-xtensa-linux",
+		"displayName": "MTL Xtensa",
+		"inherits": [ "mtl-xtensa", "xtensa-toolchain-linux" ],
+		"environment": {
+			"XTENSA_TOOLS_VERSION": "RI-2020.5-linux"
+		}
+	},
+	{
+		"name": "mtl-xtensa",
+		"hidden": true,
+		"description": "Meteorlake build using Xtensa toolchain",
+		"inherits": "common-settings",
+		"binaryDir": "${sourceDir}/../../build-mtl",
+		"cacheVariables": {
+			"BOARD": "intel_adsp_ace15_mtpm",
+			"ZEPHYR_TOOLCHAIN_VARIANT":"xcc"
+		},
+		"environment": {
+			"XTENSA_CORE": "ace10_LX7HiFi4_RI_2020_5"
+		}
+	},
+	{
+		"name": "tgl-xtensa-windows",
+		"displayName": "TGL Xtensa",
+		"inherits": [ "tgl-xtensa", "xtensa-toolchain-windows" ],
+		"environment": {
+			"XTENSA_TOOLS_VERSION": "RG-2017.8-Win32"
+		}
+	},
+	{
+		"name": "tgl-xtensa-linux",
+		"displayName": "TGL Xtensa",
+		"inherits": [ "tgl-xtensa", "xtensa-toolchain-linux" ],
+		"environment": {
+			"XTENSA_TOOLS_VERSION": "RG-2017.8-linux"
+		}
+	},
+	{
+		"name": "tgl-xtensa",
+		"hidden": true,
+		"description": "Tigerlake build using Xtensa toolchain",
+		"inherits": "common-settings",
+		"binaryDir": "${sourceDir}/../../build-tgl",
+		"cacheVariables": {
+			"BOARD": "intel_adsp_cavs25",
+			"ZEPHYR_TOOLCHAIN_VARIANT":"xcc",
+			"OVERLAY_CONFIG": "overlays/tgl/ipc4_overlay.conf"
+		},
+		"environment": {
+			"XTENSA_CORE": "cavs2x_LX6HiFi3_2017_8"
+		}
+	},
+	{
+		"name": "xtensa-toolchain-windows",
+		"hidden": true,
+		"inherits": [ "windows", "xtensa-toolchain" ],
+		"environment": {
+			"XTENSA_INSTALL_PATH": "C:/usr/xtensa"
+		}
+	},
+	{
+		"name": "xtensa-toolchain-linux",
+		"hidden": true,
+		"inherits": [ "linux", "xtensa-toolchain" ],
+		"environment": {
+			"XTENSA_INSTALL_PATH": "~/xtensa"
+		}
+	},
+	{
+		"name": "xtensa-toolchain",
+		"description": "Common settings for Xtensa tools",
+		"hidden": true,
+		"environment": {
+			"XTENSA_TOOLS_DIR": "$env{XTENSA_INSTALL_PATH}/XtDevTools/install/tools",
+			"XTENSA_BUILDS_DIR": "$env{XTENSA_INSTALL_PATH}/XtDevTools/install/builds",
+			"XTENSA_TOOLCHAIN_PATH": "$env{XTENSA_TOOLS_DIR}/$env{XTENSA_TOOLS_VERSION}",
+			"XTENSA_TOOLS": "$env{XTENSA_TOOLS_DIR}/$env{XTENSA_TOOLS_VERSION}/XtensaTools",
+			"XTENSA_SYSTEM": "$env{XTENSA_BUILDS_DIR}/$env{XTENSA_TOOLS_VERSION}/$env{XTENSA_CORE}/config"
+		}
+	},
+	{
+		"name": "windows",
+		"description": "Indicates that preset is Windows specific",
+		"hidden": true,
+		"condition": {
+			"type": "equals",
+			"lhs": "${hostSystemName}",
+			"rhs": "Windows"
+		}
+	},
+	{
+		"name": "linux",
+		"description": "Indicates that preset is Linux specific",
+		"hidden": true,
+		"condition": {
+			"type": "equals",
+			"lhs": "${hostSystemName}",
+			"rhs": "Linux"
+		}
+	}
+	],
+
+	"buildPresets": [
+	{
+		"name": "mtl-xtensa-windows",
+		"displayName": "MTL Xtensa",
+		"description": "Build MTL with Xtensa toolchain",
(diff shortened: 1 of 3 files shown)


lgirdwood:
@dbaluta @iuliana-prodan @paulstelian97 fyi - can you comment on the NXP questions.

dbaluta (approved):

dbaluta:
This looks useful. Please add a prefix to commit subject. e.g `app: Create CMakePresets.json....`

marc-hb:
I never heard about "CMake presets" before but it looks so much better that embedding these values in our Python wrapper script, thanks!

@tejlmand, @mbolivar-nordic what is your Zephyr perspective on **CMake presets** if any? I mean in general, no need to look at the changes in this PR. In other words, "Zephyric" are CMake presets? Yes/no/indifferent?

@aborisovich I'm much less impressed by the huge amount of scripting copy/paste/diverge besides the presets themselves but as this is still a draft I'm assuming it's temporary and just for testing purposes?

If this works well then please get rid of the duplicate values embedded in xtensa-build-zephyr.py. We don't want copy/paste/diverge of presets.

> Rimage integration is TODO

Let's keep that in `xtensa-build-zephyr.py` for now and fix one problem at a time. _Continuous_ integration.


I'm afraid we'll never completely get rid of that script, there are a few things that `west` will/should most likely never do like building for multiple products and laying out the `/lib/firmware/` directory structure but the smaller and more optional `xtensa-build-zephyr.py` gets, the better.

marc-hb:
> @marc-hb can you please test it on linux?

If it works in Github Actions then Linux should be just fine.

BTW any chance of seeing Windows compilation some day? I mean something like this: https://github.com/actions/runner-images/blob/main/images/win/Windows2019-Readme.md
https://docs.zephyrproject.org/latest/develop/toolchains/zephyr_sdk.html#install-zephyr-sdk-on-windows

aborisovich:
I'll also add configurations for building with Zephyr SDK toolchain.
Also note that with this approach, we set some default paths for Xtensa installation on both Windows and Linux.
If the end-user has installed Xtensa to other custom directories, he/she would be forced to implement own CMakeUserPresets.json file that overwrites paths to xtensa tools and cores (see [CMakePresets - introduction](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#introduction))

aborisovich:
> > @marc-hb can you please test it on linux?
> 
> If it works in Github Actions then Linux should be just fine.
> 
> BTW any chance of seeing Windows compilation some day? I mean something like this: https://github.com/actions/runner-images/blob/main/images/win/Windows2019-Readme.md https://docs.zephyrproject.org/latest/develop/toolchains/zephyr_sdk.html#install-zephyr-sdk-on-windows

Yes! I totally forgot about it. Thanks for reminding, I need to add it asap. Lots of people works on Windows setups now and we may break it with every commit!

app/CMakePresets.json
```diff
@@ -0,0 +1,171 @@
...
+		"displayName": "TGL Xtensa",
+		"description": "Build TGL with Xtensa toolchain",
+		"configurePreset": "tgl-xtensa-windows"
+	},
+	{
+		"name": "tgl-xtensa-windows-rebuild",
+		"displayName": "TGL Xtensa rebuild",
+		"description": "Cleans build files and rebuilds TGL with Xtensa toolchain",
+		"configurePreset": "tgl-xtensa-windows",
+		"cleanFirst": true
```
marc-hb, line 168: That's an incredibly verbose and duplicated way to just delete the build directory.

Looks like this combinatorial explosion:
https://gitlab.kitware.com/cmake/cmake/-/issues/22538

app/CMakePresets.json
```diff
@@ -0,0 +1,171 @@
...
+	  "minor": 21,
+	  "patch": 0
+	},
+
+	"configurePresets": [
+	{
+		"name": "common-settings",
+		"description": "List of common settings used by all configurations",
+		"hidden": true,
+		"generator": "Ninja",
```
marc-hb, line 14: Why is this required?

app/CMakePresets.json
```diff
@@ -0,0 +1,171 @@
...
+		"environment": {
+			"XTENSA_INSTALL_PATH": "C:/usr/xtensa"
+		}
+	},
+	{
+		"name": "xtensa-toolchain-linux",
+		"hidden": true,
+		"inherits": [ "linux", "xtensa-toolchain" ],
+		"environment": {
+			"XTENSA_INSTALL_PATH": "~/xtensa"
```
marc-hb, line 93: Please keep using XTENSA_TOOLS_ROOT for now. One change at a time. Needlessly breaking the build of every CI and every developer is not going to make CMake presets popular.

marc-hb (changes requested):
Sorry I only had a very brief look at this yesterday and I didn't realize there was  more than just moving some hardcoded constants from Python to CMake. That's a bigger change than I expected and I'm not sure it's all good. Can we please start small with CMake presets and only move the hardcoded constants away from Python first? Then we can discuss the other changes but not all at once; only one big / backward-incompatible change per Pull Request.


> It's worth to note that CMake presets replace west commands in functionality (I mean west build).

This looks like a problem. `west build/flash/other` is the One True CLI swiss army knife used all the time by 99% of the Zephyr developers and 100% of Zephyr automation it must stay functional and able to build the same thing as everyone else using presets. `west` is invoking `cmake` so surely `west` can use CMake presets, no?


>  You see, when you run west build west tool sets environmental variable ZEPHYR_BASE which we use.

We use ZEPHYR_BASE only in one place: sof/app/CMakeLists.txt

There is an alternative but I never used it:
https://docs.zephyrproject.org/latest/build/zephyr_cmake_package.html#zephyr-cmake-package-export-west

> but this works only with "our" directory layout - when SOF is "next to" Zephyr directory.

This is enforced by sof/west.yml anyway so I don't really see the problem.

> Note - in Zephyr CI SOF is cloned to modules directory just like defined in the zephyr/west.yml manifest. So this directory layout no longer works.

Can you share the twister command that fails? I lost track of how zephyr/west.yml uses the SOF repo sorry.

> I'll also add configurations for building with Zephyr SDK toolchain.

BTW that toolchain never required any explicit configuration, installing it in the home directory was always enough.

> Also note that with this approach, we set some default paths for Xtensa installation on both Windows and Linux.
If the end-user has installed Xtensa to other custom directories, he/she would be forced to implement own CMakeUserPresets.json file that overwrites paths to xtensa tools and cores

This looks like a big regression. It used to work with a single XTENSA_TOOLS_ROOT environment variable, why can't we keep that? At least for now; this seems unrelated to CMake presets. One change at a time.



BTW CMake presets do not seem perfect, just found this: https://gitlab.kitware.com/cmake/cmake/-/issues/22538

mbolivar-nordic:
>  @mbolivar-nordic what is your Zephyr perspective on CMake presets if any? I mean in general, no need to look at the changes in this PR. In other words, how "zephyric" are CMake presets? Yes/no/indifferent?

I've looked at them briefly and they seem interesting and that's the level of analysis I have done

aborisovich:
Yeah, west can use Cmake presets, we just need to override default build command for west build. Then west build would call cmake presents. I'll also remove those xtensa default paths. Anyway moved this as P2 now, working hard on Windows CI, should be done before Christmas 😊

tejlmand:
> @tejlmand, @mbolivar-nordic what is your Zephyr perspective on **CMake presets** if any? I mean in general, no need to look at the changes in this PR. In other words, how "zephyric" are CMake presets? Yes/no/indifferent?

@marc-hb I have looked at them and done some prototyping in Zephyr context.
I find presets promising and flexible, but care must be taken.

For example it would be possible to move content of `sample.yml` and `testcase.yml` into preset files and thus just maintain a single file.

However, in Zephyr itself it would request people to shift to a different workflow, and that's the main reason I haven't pushed for such change (as well as concern of misuse of the feature).

With flexibility of presets comes great responsibility as you can start pushing all sorts of settings information into the preset file.
Settings that rightfully should be provided by the user, like `BOARD`.
This means a lot of noise for many people may creep into a preset file.

As also seen in this PR, a 171 line json file 😮 

For those curios, I have some simple prototyping here of using preset file with Zephyr and migrating a `sample.yaml` to presets:
https://github.com/nrfconnect/sdk-zephyr/pull/469/commits/3237110cb770ca803226e2729b4401359a354ff1
https://github.com/nrfconnect/sdk-nrf/pull/3979

lgirdwood:
@aborisovich any update ?

aborisovich:
Long forgotten due to other pressing issues we are late for ;-)
But your right, stale PRs should are not good for the project. I'll find the time tomorrow for this one.

marc-hb:
> But your right, stale PRs should are not good for the project.

On the other hand, maintainers should not let github **drafts** like this one deprive other, real PRs of their precious and very limited time :-)

2/3 of the currently open SOF PRs are drafts / work not ready for formal review yet! https://github.com/thesofproject/sof/pulls?q=is%3Apr+is%3Aopen+-is%3Adraft

aborisovich:
Working on it in my free time, but this PR requires much more work due to integration with west.
I think best architectural approach would be to:
`xtensa-build-zephyr.py` -> `west build` ->`CmakePreset`.
So, I need to wrote `west build` command extension for SOF in python that would overwrite `west build` command provided by Zephyr. Moreover, since CMakePresets are used by multiple IDEs like Visual Studio, CLion and some common editors I think...
Those IDEs typically provide GUI buttons like "build" or "run" that use CmakePresets underneath.
So we also would need to make sure that overriden `west build` command in SOF does not expand functionalities over CMakePresets (that `west build` == `Cmake preset` feature wise).
Then again, people building SOF from Zephyr repo as "Starting point" will clone SOF as module of Zephyr, but `west` tool keeps being initialized to zephyr directory in such scenario. So, if SOF has overridden `west build`, it would not be used.
So the thing is not trivial as you see...

aborisovich:
But, keeping discussion alive, I don't think @marc-hb @tejlmand that this:

> Looks like this combinatorial explosion:
> https://gitlab.kitware.com/cmake/cmake/-/issues/22538

is a big issue here. I mean - what are the consequences of those?
You look at it from "development good practices" perspective but is it really the right thing to do?
Let me elaborate - this file is not something that is browsed, read and developed continuously. So there is no pain if it has 1000 lines if you do not work on the file. Granted, if something changes in build system configuration you have to make sure very carefully that every "repeated" build configuration is changed accordingly but this is the only downside.
What this file is supposed to do, is to provide a friendly name for every variation of action you execute on the build system.
Another aspect - please note that IDEs GUIs that use CmakePresets typically display a selection of buildConfigurations based on selection of configure section. Let me explain what I mean:
If you select "MTL Xtensa" configure preset:
image
Then, for build preset only those matching selected configure preset are displayed:
(Note - no: "TGL Build" or "TGL rebuild" options show now):
image

To sum up:
I agree that current CmakePresets implementation is "tedious" due to the "combinatorial explosion", but this is not a reason painful enough to avoid implementing it nowadays in SOF. Lets implement it now as it is and with next CMake version improvements in future we will also improve how the file looks. I saw that CMake team is working hard on finding a solution to combinatorial explosion @marc-hb mentioned.

aborisovich:
Closing for now no time for this. Will reopen when re-introduced.

Issue #6615, opened 2022-11-16
[BUG] addition of AMD_HS breaks backwards compatibility


plbossart:
https://lore.kernel.org/alsa-devel/36a45c7a-820a-7675-d740-c0e83ae2c417@collabora.com/

We need to move AMD_HS to the end of the dai list

mengdonglin:
Fixes merged.

PR #6772, opened 2022-12-12
backport: dai: move AMD_HS to end of list to restore backwards-compatibility

Zames-Chang:
The addition of AMD_HS breaks Mediatek platforms by using an index previously allocated to Mediatek. This is a backwards-compatibility issue and needs to be fixed. All firmware released by AMD needs to be re-generated and re-distributed.

Fixes: 83ee5b49a987 ("dai: support for amd specific hs dai id")
Link: https://github.com/thesofproject/sof/issues/6615
Link: https://lore.kernel.org/alsa-devel/36a45c7a-820a-7675-d740-c0e83ae2c417@collabora.com/
Reported-by: AngeloGioacchino Del Regno <angelogioacchino.delregno@collabora.com>
Signed-off-by: Pierre-Louis Bossart <pierre-louis.bossart@linux.intel.com>

### DIFF
--- a/src/include/ipc/dai.h
+++ b/src/include/ipc/dai.h
@@ -89,8 +89,8 @@ enum sof_ipc_dai_type {
 	SOF_DAI_AMD_BT,			/**< Amd BT */
 	SOF_DAI_AMD_SP,			/**< Amd SP */
 	SOF_DAI_AMD_DMIC,		/**< Amd DMIC */
+	SOF_DAI_MEDIATEK_AFE,           /**< Mtk AFE */
 	SOF_DAI_AMD_HS,			/**< Amd HS */
-	SOF_DAI_MEDIATEK_AFE            /**< Mtk AFE */
 };
 
 /* general purpose DAI configuration */


chunxu-li (approved):

lgirdwood (approved):

marc-hb (commented):
Not the time to change many lines in this PR but in the future wouldn't it better to explicitly initialize each value instead of only relying on the declaration order? This would have avoided this regression (smaller PRs with less than 80% copy/paste would have helped too; I digress)

Speaking of the longer term: doesn't the main branch need a fix too? The damage control strategy is not clear to me.

My 2 cents.

marc-hb:
> Speaking of the longer term: doesn't the main branch need a fix too? 

Just spotted "backport" in the title, sorry.

@Zames-Chang next time please use the `-x` option: `git cherry-pick -x` c601f7d1bc0c

Zames-Chang:
@marc-hb Thanks for reminder, I will use cherry-pick next time.

@tinghan-shen for vis

PR #6773, opened 2022-12-12
comp: fix binding modules

fkwasowi:
Current ipc_comp_connect() connects the source first, then calls comp_bind
and only then connects the sink. This leads to errors with modules that
assume sink to be connected at bind. This is a valid assumption, so fix
the order in ipc_comp_connect().

### DIFF
--- a/src/ipc/ipc4/helper.c
+++ b/src/ipc/ipc4/helper.c
@@ -371,9 +371,6 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 		goto free;
 	}
 
-	ret = comp_bind(source, bu);
-	if (ret < 0)
-		goto e_src_bind;
 
 	ret = comp_buffer_connect(sink, sink->ipc_config.core, buffer,
 				  PPL_CONN_DIR_BUFFER_TO_COMP);
@@ -381,6 +378,10 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 		tr_err(&ipc_tr, "failed to connect internal buffer to sink %d", sink_id);
 		goto e_sink_connect;
 	}
+
+	ret = comp_bind(source, bu);
+	if (ret < 0)
+		goto e_src_bind;
 
 	ret = comp_bind(sink, bu);
 	if (ret < 0)
@@ -408,10 +409,10 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 err:
 	comp_unbind(sink, bu);
 e_sink_bind:
-	pipeline_disconnect(sink, buffer, PPL_CONN_DIR_BUFFER_TO_COMP);
-e_sink_connect:
 	comp_unbind(source, bu);
 e_src_bind:
+	pipeline_disconnect(sink, buffer, PPL_CONN_DIR_BUFFER_TO_COMP);
+e_sink_connect:
 	pipeline_disconnect(source, buffer, PPL_CONN_DIR_COMP_TO_BUFFER);
 free:
 	irq_local_enable(flags);


dbaluta (approved):

lgirdwood (approved):

lgirdwood:
@fkwasowi can you check internal CI thanks !

src/ipc/ipc4/helper.c
```diff
@@ -382,6 +379,10 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 		goto e_sink_connect;
 	}
 
+	ret = comp_bind(source, bu);
+		if (ret < 0)
```
ranj063, line 383: you'd also need to fix the error paths below. e_sink_connect label no longer needs to do an unbind right?

ranj063:
Is there an issue filed so we can see when this problem occurs?

src/ipc/ipc4/helper.c
kv2019i: Some whitespace error on L383 and L384, indentation seems off.

kv2019i (changes requested):
Looks like a good fix, but it would be helpful to explain why this has worked before and why the change is needed now. So maybe something like:

```
comp: fix module bind order

Current ipc_comp_connect() connects the source first, then calls comp_bind
and only then connects the sink. This leads to errors with modules that
assume sink to be connected at bind. This is a valid assumption, so fix
the order in ipc_comp_connect().
```

And some indentation issue in code. Maybe a github UI artifact, but please check.

fkwasowi:
> Is there an issue filed so we can see when this problem occurs?

There is no issue created for this defect. The problem occurs, for example, during a test scenario \conv_fw_003_kd\test_01_kd_topologies

fkwasowi:
> Looks like a good fix, but it would be helpful to explain why this has worked before and why the change is needed now. So maybe something like:
> 
> ```
> comp: fix module bind order
> 
> Current ipc_comp_connect() connects the source first, then calls comp_bind
> and only then connects the sink. This leads to errors with modules that
> assume sink to be connected at bind. This is a valid assumption, so fix
> the order in ipc_comp_connect().
> ```
> 
> And some indentation issue in code. Maybe a github UI artifact, but please check.

I have changed the description

src/ipc/ipc4/helper.c
fkwasowi: whitespaces fixed

src/ipc/ipc4/helper.c
```diff
@@ -382,6 +379,10 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 		goto e_sink_connect;
 	}
 
+	ret = comp_bind(source, bu);
+	if (ret < 0)
+		goto e_src_bind;
```
kv2019i, line 384: Error handling in e_src_bind and e_sink_connect implementation needs to be changed when the patch changes the order.

kv2019i (changes requested):
Thanks @fkwasowi ! I think the goto handling is still missing, that needs to be changed to reflect the new order.

src/ipc/ipc4/helper.c
fkwasowi: I have changed the order

src/ipc/ipc4/helper.c
```diff
@@ -371,17 +371,18 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
 				  PPL_CONN_DIR_BUFFER_TO_COMP);
 	if (ret < 0) {
 		tr_err(&ipc_tr, "failed to connect internal buffer to sink %d", sink_id);
-		goto e_sink_connect;
+		goto e_src_bind;
 	}
 
+	ret = comp_bind(source, bu);
+	if (ret < 0)
+		goto e_sink_connect;
```
kv2019i, line 384: And have "e_source_bind" here.

src/ipc/ipc4/helper.c
```diff
@@ -371,17 +371,18 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
-	ret = comp_bind(source, bu);
-	if (ret < 0)
-		goto e_src_bind;
 
 	ret = comp_buffer_connect(sink, sink->ipc_config.core, buffer,
 				  PPL_CONN_DIR_BUFFER_TO_COMP);
 	if (ret < 0) {
 		tr_err(&ipc_tr, "failed to connect internal buffer to sink %d", sink_id);
-		goto e_sink_connect;
+		goto e_src_bind;
```
kv2019i, line 379: So actually, please keep this as "goto e_sink_connect;" and then
at end of function, have:
```
err:
	comp_unbind(sink, bu);
e_sink_bind:
	comp_unbind(source, bu);
e_src_bind:
e_sink_connect:
	pipeline_disconnect(source, buffer, PPL_CONN_DIR_COMP_TO_BUFFER);
```

kv2019i (changes requested):
Still some tuning in the error case, please see my inline comment.

src/ipc/ipc4/helper.c
fkwasowi: Done.Is it now correct?

kv2019i (approved):
Thanks @fkwasowi , looks good now!

fkwasowi:
SOFCI TEST

ranj063 (approved):

libinyang:
Patch looks good to me. I have tested on WoV pipeline. This can fix https://github.com/thesofproject/sof/issues/6784

libinyang (approved):

PR #6775, opened 2022-12-12
comp_update: null pointer buffer prevent

pjdobrowolski:
Added NULL pointer prevent.

### DIFF
--- a/src/audio/buffer.c
+++ b/src/audio/buffer.c
@@ -179,8 +179,7 @@ void buffer_free(struct comp_buffer *buffer)
 		.buffer = buffer,
 	};
 
-	if (!buffer)
-		return;
+	assert(buffer != NULL);
 
 	buf_dbg(buffer, "buffer_free()");
 
@@ -209,6 +208,8 @@ void buffer_free(struct comp_buffer *buffer)
  */
 void comp_update_buffer_produce(struct comp_buffer __sparse_cache *buffer, uint32_t bytes)
 {
+	assert(buffer != NULL);
+
 	struct buffer_cb_transact cb_data = {
 		.buffer = buffer,
 		.transaction_amount = bytes,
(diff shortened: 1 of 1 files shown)


paulstelian97 (approved):
This is hot path code, so it's quite possible that the NULL check will have a performance impact.

That said, I think it's better to be safe and do the check anyway, but I recommend doing a performance verification to see if there is an issue there. My intuition says there won't be any issue (hence the approve) but I recommend checking anyway.

pjdobrowolski:
> This is hot path code, so it's quite possible that the NULL check will have a performance impact.
> 
> That said, I think it's better to be safe and do the check anyway, but I recommend doing a performance verification to see if there is an issue there. My intuition says there won't be any issue but I recommend checking anyway.

Do you have any particular performance check in mind?

pblaszko (approved):

paulstelian97:
Well the copy() function of all components calls one of these two functions, which happens once per period (which varies between 1 and several milliseconds, typically) per component. I'd guess that a platform with many active pipelines/components active would be the most impacted one, as it would go through this code the most times per period.

Now an actual check of the cycle counts for the copy() function inside components can be done. Again, I feel like it shouldn't have that huge of an effect since we don't typically border on having an issue with this anyway (my expectation is that it wouldn't be more than a 1% increase, and components which do more processing are affected less). I would re-approve because I don't expect any big issues to show up here.

src/audio/buffer.c
```diff
@@ -209,6 +209,9 @@ void buffer_free(struct comp_buffer *buffer)
  */
 void comp_update_buffer_produce(struct comp_buffer __sparse_cache *buffer, uint32_t bytes)
 {
+	if (!buffer)
```
marcinszkudlinski, line 212: NULL here is a most unlike event and means - something is really wrong. Return without any error code may lead to undefined and had to debug problems. 
Assert and controlled crash is a better sollution, just keep in mind that ASSERT_NO_MSG (zephyr assert) is currently off and doesn't provide any protection - use SOF assert() macro

marcinszkudlinski (changes requested):

pjdobrowolski:
Oh, I thought that you have some metrics or UT in hand. Well, we do validation on regular basics so I won't worry about that.

src/audio/buffer.c
paulstelian97: Ah, true, I should take this into account when I do my own reviewing. A few other related static analysis "fixes" also had this issue of silencing errors instead of properly logging/propagating them.

pjdobrowolski: Done

PR #6776, opened 2022-12-12
buffer: NULL pointer protection

pjdobrowolski:
comp_update might point to NULL in certain situations.

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -677,6 +677,8 @@ int module_adapter_copy(struct comp_dev *dev)
 	}
 
 	if (mod->simple_copy) {
+		if (!source_c)
+			return -ENODATA;
 		comp_update_buffer_consume(source_c, mod->input_buffers[0].consumed);
 		buffer_release(sink_c);
 		buffer_release(source_c);


marcinszkudlinski (approved):

src/audio/module_adapter/module_adapter.c
```diff
@@ -677,6 +677,8 @@ int module_adapter_copy(struct comp_dev *dev)
 	}
 
 	if (mod->simple_copy) {
+		if(!source_c)
```
ranj063, line 680: @pjdobrowolski Could you please laborate the "certain situations" part? And even when this is true isnt it a bit too late to be checking for this? Can we check this on line 617?

ranj063: Ideally, filing an issue and then linking the PR would be the best way to help with reviews.

src/audio/module_adapter/module_adapter.c
```diff
@@ -677,6 +677,8 @@ int module_adapter_copy(struct comp_dev *dev)
 	}
 
 	if (mod->simple_copy) {
+		if (!source_c)
+			return -ENODATA;
```
kv2019i, line 681: Aa, so this is basicly to keep static checkers happy? You do need to "goto out" in this case to free the buffers on error.

kv2019i (changes requested):
One thing more, you do need to go through clean up via "out:" on error.

src/audio/module_adapter/module_adapter.c
pjdobrowolski: @ranj063 this issue is a product of static analysis tool. I've checked your suggestions but I'm afraid that that is only solution.  
@kv2019i That PR is for private release branch for mtl and there was no need of opening issue.

src/audio/module_adapter/module_adapter.c
pjdobrowolski: It is my first time using that tool and I thought that it would catch such trivial mistake.

paulstelian97: That's the thing with static analysis tools -- they lack context on what is correct and thus they only help you not crash/get other forms of UB, but NOT actually write the correct code when doing so.

ujfalusi: But this is invalid report, see:
```
if (mod->simple_copy) { // line 612
...
	source_c = buffer_acquire(source);
...
} else {
...
}

if (mod->simple_copy) { // line 679, here
	// source_c can not be NULL.
}

PR #6777, opened 2022-12-13
audio: move google components to a subfolder

cujomalainey:
just some minor house cleaning


lkoenig (approved):

ranj063 (approved):
now you can stop adding the "google_"prefix for modules in the future with this change

PR #6785, opened 2022-12-13
kpb: add support 24/24bit buffering for 1-6 channels

fkwasowi:
FW infrastructure shall support buffering of historic data from 1ch up to 6 channels 24bit samples in 24bit container. 
KPB module shall support on input 3 to 6 channels (1-4 mics + 2ch reference), 16kHz, 16/24bit . Output pins shall support: 
For buffered pin same number of channels as on input, 16kHz, bit depth same as on input.

### DIFF
--- a/src/audio/kpb.c
+++ b/src/audio/kpb.c
@@ -814,8 +814,7 @@ static int kpb_prepare(struct comp_dev *dev)
 		ret = -EIO;
 	}
 
-	/* Disallow sync_draining_mode for now */
-	kpb->sync_draining_mode = false;
+	kpb->sync_draining_mode = true;
 
 	kpb_change_state(kpb, KPB_STATE_RUN);
 
--- a/src/audio/kpb.c
+++ b/src/audio/kpb.c
@@ -224,7 +224,7 @@ static void kpb_set_params(struct comp_dev *dev,
 	params->sample_valid_bytes =
 		kpb->ipc4_cfg.base_cfg.audio_fmt.valid_bit_depth / 8;
 	params->buffer_fmt = kpb->ipc4_cfg.base_cfg.audio_fmt.interleaving_style;
-	params->buffer.size = kpb->ipc4_cfg.base_cfg.obs * KPB_MAX_BUFF_TIME * 2;
+	params->buffer.size = kpb->ipc4_cfg.base_cfg.obs * KPB_MAX_BUFF_TIME * params->channels;
 
 	params->host_period_bytes = params->channels *
 				    params->sample_container_bytes *
@@ -704,7 +704,7 @@ static int kpb_prepare(struct comp_dev *dev)
 	struct comp_data *kpb = comp_get_drvdata(dev);
 	int ret = 0;
 	int i;
-	size_t hb_size_req = KPB_MAX_BUFFER_SIZE(kpb->config.sampling_width);
+	size_t hb_size_req = KPB_MAX_BUFFER_SIZE(kpb->config.sampling_width, kpb->config.channels);
 
 	comp_dbg(dev, "kpb_prepare()");
 
@@ -1807,13 +1807,13 @@ static inline bool validate_host_params(struct comp_dev *dev,
 		comp_err(dev, "kpb: host_period_size (%d) cannot be 0 and host_buffer_size (%d) cannot be 0",
 			 host_period_size, host_buffer_size);
 		return false;
-	} else if (HOST_BUFFER_MIN_SIZE(hb_size_req) >
+	} else if (HOST_BUFFER_MIN_SIZE(hb_size_req, kpb->config.channels) >
 		   host_buffer_size) {
 		/* Host buffer size is too small - history data
 		 * may get overwritten.
 		 */
 		comp_err(dev, "kpb: host_buffer_size (%d) must be at least %d",
-			 host_buffer_size, HOST_BUFFER_MIN_SIZE(hb_size_req));
+			 host_buffer_size, HOST_BUFFER_MIN_SIZE(hb_size_req, kpb->config.channels));
 		return false;
 	} else if (kpb->sync_draining_mode) {
 		/* Sync draining allowed. Check if we can perform draining
--- a/src/audio/kpb.c
+++ b/src/audio/kpb.c
@@ -107,7 +107,7 @@ static void kpb_free_history_buffer(struct history_buffer *buff);
 static inline bool kpb_is_sample_width_supported(uint32_t sampling_width);
 static void kpb_copy_samples(struct comp_buffer __sparse_cache *sink,
 			     struct comp_buffer __sparse_cache *source, size_t size,
-			     size_t sample_width);
+			     size_t sample_width, uint32_t channels);
 static void kpb_drain_samples(void *source, struct audio_stream __sparse_cache *sink,
 			      size_t size, size_t sample_width);
 static void kpb_buffer_samples(const struct audio_stream __sparse_cache *source,
@@ -914,6 +914,7 @@ static int kpb_copy(struct comp_dev *dev)
 	size_t sample_width = kpb->config.sampling_width;
 	struct draining_data *dd = &kpb->draining_task_data;
 	uint32_t avail_bytes;
+	uint32_t channels = kpb->config.channels;
 
 	comp_dbg(dev, "kpb_copy()");
 
@@ -965,7 +966,7 @@ static int kpb_copy(struct comp_dev *dev)
 			break;
 		}
 
-		kpb_copy_samples(sink_c, source_c, copy_bytes, sample_width);
+		kpb_copy_samples(sink_c, source_c, copy_bytes, sample_width, channels);
 
 		/* Buffer source data internally in history buffer for future
 		 * use by clients.
@@ -1025,7 +1026,7 @@ static int kpb_copy(struct comp_dev *dev)
 			break;
 		}
 
-		kpb_copy_samples(sink_c, source_c, copy_bytes, sample_width);
+		kpb_copy_samples(sink_c, source_c, copy_bytes, sample_width, channels);
 
 		comp_update_buffer_produce(sink_c, copy_bytes);
 		comp_update_buffer_consume(source_c, copy_bytes);
@@ -1593,6 +1594,77 @@ out:
 	return SOF_TASK_STATE_COMPLETED;
 }
 
+#ifdef KPB_HIFI3
+static void kpb_convert_24b_to_32b(const void *linear_source, int ioffset,
+				   struct audio_stream __sparse_cache *sink, int ooffset,
+				   unsigned int n_samples)
+{
+	int ssize = audio_stream_sample_bytes(sink);
+	uint8_t *in = (uint8_t *)linear_source + ioffset * ssize;
+	uint8_t *out = audio_stream_wrap(sink, (uint8_t *)sink->w_ptr + ooffset * ssize);
+	ae_int32x2 *buf_end;
+	ae_int32x2 *buf;
+
+	buf = (ae_int32x2 *)(sink->addr);
+	buf_end = (ae_int32x2 *)(sink->end_addr);
+	ae_int32x2 *out_ptr = (ae_int32x2 *)buf;
+
+	AE_SETCBEGIN0(buf);
+	AE_SETCEND0(buf_end);
+	out_ptr = (ae_int32x2 *)out;
+
+	ae_valign align_in = AE_LA64_PP(in);
+	int i = 0;
+	ae_int24x2 d24 = AE_ZERO24();
+
+	if (!IS_ALIGNED((uintptr_t)out_ptr, 8)) {
+		AE_LA24_IP(d24, align_in, in);
+		ae_int32x2 d320 = d24;
+		int higher = AE_MOVAD32_H(d320);
+		*(ae_int32 *)(out_ptr) = higher << 8;
+		out_ptr = (ae_int32x2 *)(out + 4);
+		++i;
+	}
+	/* process two samples in single iteration to increase performance */
+	while (i < (int)n_samples - 1) {
+		AE_LA24X2_IP(d24, align_in, in);
+		ae_int32x2 d320 = d24;
+
+		d320 = AE_SLAI32(d320, 8);
+		AE_S32X2_XC(d320, out_ptr, 8);
+		i += 2;
+	}
+	if (i != (int)n_samples) {
+		AE_LA24X2_IP(d24, align_in, in);
+		ae_int32x2 d320 = d24;
+		int higher = AE_MOVAD32_H(d320);
+		*(ae_int32 *)(out_ptr) = higher << 8;
+	}
+}
+#else
+static void kpb_convert_24b_to_32b(const void *source, int ioffset,
+				   struct audio_stream __sparse_cache *sink,
+				   int ooffset, unsigned int samples)
+{
+	int ssize = audio_stream_sample_bytes(sink);
+	uint8_t *src = (uint8_t *)source + ioffset * 3;
+	int32_t *dst = audio_stream_wrap(sink, (uint8_t *)sink->w_ptr + ooffset * ssize);
+	int processed;
+	int nmax, i, n;
+
+	for (processed = 0; processed < samples; processed += n) {
+		dst = audio_stream_wrap(sink, dst);
+		n = samples - processed;
+		nmax = KPB_BYTES_TO_S32_SAMPLES(audio_stream_bytes_without_wrap(sink, dst));
+		n = MIN(n, nmax);
+		for (i = 0; i < n; i += 1) {
+			*dst = (src[2] << 16) | (src[1] << 8) | src[0];
+			dst++;
+			src += 3;
+		}
+	}
+}
+#endif
 /**
  * \brief Drain data samples safe, according to configuration.
  *
@@ -1616,6 +1688,9 @@ static void kpb_drain_samples(void *source, struct audio_stream __sparse_cache *
 #endif /* CONFIG_FORMAT_S16LE */
 #if CONFIG_FORMAT_S24LE || CONFIG_FORMAT_S32LE
 	case 24:
+		samples = size / ((sample_width >> 3) * sink->channels);
+		kpb_convert_24b_to_32b(source, 0, sink, 0, samples);
+		break;
 	case 32:
 		samples = KPB_BYTES_TO_S32_SAMPLES(size);
 		audio_stream_copy_from_linear(source, 0, sink, 0, samples);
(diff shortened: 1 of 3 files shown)


src/audio/kpb.c
```diff
@@ -1593,6 +1594,44 @@ static enum task_state kpb_draining_task(void *arg)
...
 }
 
+void kpb_convert_24b_to_32b(const void *linear_source, int ioffset,
+			    struct audio_stream __sparse_cache *sink, int ooffset,
+			    unsigned int n_samples)
+{
+	int ssize = audio_stream_sample_bytes(sink);
+	uint8_t *in = (uint8_t *)linear_source + ioffset * ssize;
+	uint8_t *out = audio_stream_wrap(sink, (uint8_t *)sink->w_ptr + ooffset * ssize);
+	ae_int32x2 *out_ptr = (ae_int32x2 *)(out);
```
marcinszkudlinski, line 1604: please add also a GCC version - without HiFi specific code

marcinszkudlinski (changes requested):

src/platform/meteorlake/include/platform/lib/memory.h
```diff
@@ -56,7 +56,7 @@
 /**
  * size of HPSRAM system heap
  */
-#define HEAPMEM_SIZE 0x40000
+#define HEAPMEM_SIZE 0xD0000
```
lgirdwood, line 59: Btw, not sure if this is really used now with Zephyr RTOS config. e.g. it should come from Zephyr, but this could still be on the TODO list.

src/audio/kpb.c
lgirdwood: ack - we separate the arch generic/hifi code using different files or with ifdefs.
@andrula-song @singalsu pls review the HiFi code.

src/audio/kpb.c
```diff
@@ -1648,11 +1707,15 @@ static void kpb_buffer_samples(const struct audio_stream __sparse_cache *source,
 		samples_count = KPB_BYTES_TO_S16_SAMPLES(size);
 		samples_offset = KPB_BYTES_TO_S16_SAMPLES(offset);
 		audio_stream_copy_to_linear(source, samples_offset,
-					    sink, 0, samples_count);
+				    sink, 0, samples_count);
```
lgirdwood, line 1710: looks like indentation is broken here.

src/audio/kpb.c
```diff
@@ -1648,11 +1707,15 @@ static void kpb_buffer_samples(const struct audio_stream __sparse_cache *source,
...
 		break;
 #endif
 #if CONFIG_FORMAT_S24LE || CONFIG_FORMAT_S32LE
 	case 24:
+		samples_count = size / ((sample_width / 8) * source->channels);
+		samples_offset = offset / ((sample_width / 8) * source->channels);
```
lgirdwood, line 1716: best to avoid divisions where possible for performance.  I think in this case the sample_width will always be 4 (since IIRC we use a 4 byte container for 24 bit) ?

src/audio/kpb.c
```diff
@@ -1716,6 +1779,36 @@ static inline bool kpb_is_sample_width_supported(uint32_t sampling_width)
...
+	ae_int32x2 vs = AE_ZERO32();
+
+	if (!IS_ALIGNED((uintptr_t)sin, 8)) {
+		AE_L32_IP(vs, (const ae_int32 *)(sin), 4);
+		AE_S32_L_IP(vs, (ae_int32 *)(sout), 4);
+		n_samples--;
+	}
+	ae_valign align_out = AE_ZALIGN64();
+
+	for (size_t i = 0; i < n_samples / 2; i++) {
```
lgirdwood, line 1801: re avoid divisions, this can be shifted which is cheap on xtensa.

src/audio/kpb.c
```diff
@@ -1741,6 +1835,8 @@ static void kpb_copy_samples(struct comp_buffer __sparse_cache *sink,
 #endif
 #if CONFIG_FORMAT_S24LE || CONFIG_FORMAT_S32LE
 	case 24:
+		samples = size / ((sample_width / 8) * channels);
```
lgirdwood, line 1838: ditto re constants and divisions.

lgirdwood (commented):
LGTM, mostly minor stuff - needs hifi review

src/audio/kpb.c
```diff
@@ -1648,11 +1707,15 @@ static void kpb_buffer_samples(const struct audio_stream __sparse_cache *source,
...
-					    sink, 0, samples_count);
+				    sink, 0, samples_count);
 		break;
 #endif
 #if CONFIG_FORMAT_S24LE || CONFIG_FORMAT_S32LE
 	case 24:
+		samples_count = size / ((sample_width / 8) * source->channels);
+		samples_offset = offset / ((sample_width / 8) * source->channels);
+		kpb_convert_32b_to_24b(source, samples_offset,
+				       sink, 0, samples_count);
```
tlissows, line 1718: Consider adding _fallthrough_ annotation.

src/audio/kpb.c
```diff
@@ -1741,6 +1835,8 @@ static void kpb_copy_samples(struct comp_buffer __sparse_cache *sink,
 #endif
 #if CONFIG_FORMAT_S24LE || CONFIG_FORMAT_S32LE
 	case 24:
+		samples = size / ((sample_width / 8) * channels);
+		kpb_copy_24b_in_32b(istream, 0, ostream, 0, samples);
```
tlissows, line 1839: ditto consider adding fallthrough annotation.

src/audio/kpb.c
```diff
@@ -1593,6 +1594,44 @@ static enum task_state kpb_draining_task(void *arg)
 	return SOF_TASK_STATE_COMPLETED;
 }
 
+void kpb_convert_24b_to_32b(const void *linear_source, int ioffset,
```
tlissows, line 1597: Consider declaring this function as static if it is not going to be used outside the module.

src/audio/kpb.c
```diff
@@ -1627,6 +1668,24 @@ static void kpb_drain_samples(void *source, struct audio_stream __sparse_cache *
 	}
 }
 
+void kpb_convert_32b_to_24b(const struct audio_stream __sparse_cache *source, int ioffset,
```
tlissows, line 1671: Consider declaring this function as static if it is not going to be used outside the module.

src/audio/kpb.c
```diff
@@ -1716,6 +1779,36 @@ static inline bool kpb_is_sample_width_supported(uint32_t sampling_width)
 	return ret;
 }
 
+void kpb_copy_24b_in_32b(const struct audio_stream __sparse_cache *source, uint32_t ioffset,
```
tlissows, line 1782: Consider declaring this function as static if it is not going to be used outside the module.

src/platform/meteorlake/include/platform/lib/memory.h
kv2019i: @lgirdwood ACE and IMX are still using HEAPMEM_SIZE from the headers. For cAVS, we are taking the values from linker scripts. All variants are in sof/zephyr/lib/alloc.c

src/audio/kpb.c
```diff
@@ -1716,6 +1779,36 @@ static inline bool kpb_is_sample_width_supported(uint32_t sampling_width)
...
+		n_samples--;
+	}
+	ae_valign align_out = AE_ZALIGN64();
+
+	for (size_t i = 0; i < n_samples / 2; i++) {
+		AE_L32X2_IP(vs, sin, 8);
+		AE_SA32X2_IP(vs, align_out, sout);
+	}
+	AE_SA64POS_FP(align_out, sout);
+	if (n_samples % 2) {
```
lyakh, line 1806: I'd actually hope that the compiler is smart enough by now to replace `/ 2` with `>> 1` then it should also replace this with `& 1` but if we decide to perform this optimisation everywhere manually, then we should probably also replace this.

src/audio/kpb.c
```diff
@@ -814,7 +814,7 @@ static int kpb_prepare(struct comp_dev *dev)
 	}
 
 	/* Disallow sync_draining_mode for now */
-	kpb->sync_draining_mode = false;
+	kpb->sync_draining_mode = true;
```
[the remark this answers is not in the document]
lgirdwood: @fkwasowi the comment needs aligned to the code update as the comment is now wrong.

src/audio/kpb.c
lgirdwood: ack - if it's an internal API we need static.

src/audio/kpb.c
lgirdwood: still broken - sometimes easier to use ```git gui``` here to discard these lines as white space changes are difficult to see.

src/audio/kpb.c
lgirdwood: ```sample_width >> 3``` shifts are faster than divisions on xtensa (compiler should know, but lets be explicit)

src/audio/kpb.c
lgirdwood: need to update, this variable is also not changed in the loops so could be calculated before the loop (again compiler should know better, but is not great so lets help it).

src/audio/kpb.c
```diff
@@ -1593,6 +1594,70 @@ static enum task_state kpb_draining_task(void *arg)
...
+	int nmax, i, n;
+
+	for (processed = 0; processed < samples; processed += n) {
+		dst = audio_stream_wrap(sink, dst);
+		n = samples - processed;
+		nmax = audio_stream_bytes_without_wrap(sink, dst) >> BYTES_TO_S32_SAMPLES;
+		n = MIN(n, nmax);
+		for (i = 0; i < n; i += 1) {
+			*dst = (*(src + 2) << 24) | (*(src + 1) << 16) | (*(src + 0) << 8);
+			*dst >>= 8;
```
lyakh, line 1654: can we just make this
```
*dst = (src[2] << 16) | (src[1] << 8) | src[0];
```

src/platform/meteorlake/include/platform/lib/memory.h
iganakov, line 59: I think it is good to provide more precise explanation in commit message why you're increasing HEAPMEM_SIZE by 0x90000, i.e. in case of audio format 16000Hz/6ch/24bit history_buffer_size = 16 * 6 * 3 * 2100s = 604800 bytes (0x93A80)

iganakov (changes requested):

src/audio/kpb.c
tlissows: With the new approach this line should be removed:
`*dst >>= 8;`

tlissows (changes requested):

marcinszkudlinski (approved):

iganakov (approved):

tlissows (approved):

.xxproject
```diff
@@ -0,0 +1,2 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<xxProperties/>
```
lyakh: left-over? Should be removed I guess

src/audio/kpb.c
```diff
@@ -1592,6 +1593,69 @@ static enum task_state kpb_draining_task(void *arg)
...
+
+	if (!IS_ALIGNED((uintptr_t)out_ptr, 8)) {
+		AE_LA24_IP(d24, align_in, in);
+		ae_int32x2 d320 = d24;
+		int higher = AE_MOVAD32_H(d320);
+		*(ae_int32 *)(out_ptr) = higher << 8;
+		out_ptr = (ae_int32x2 *)(out + 4);
+		++i;
+	}
+	// process two samples in single iteration to increase performance
```
lyakh, line 1618: comment style

src/audio/kpb.c
```diff
@@ -1592,6 +1593,69 @@ static enum task_state kpb_draining_task(void *arg)
...
+		int higher = AE_MOVAD32_H(d320);
+		*(ae_int32 *)(out_ptr) = higher << 8;
+	}
+}
+#else
+static void kpb_convert_24b_to_32b(const void *source, int ioffset,
+				   struct audio_stream __sparse_cache *sink,
+				   int ooffset, unsigned int samples)
+{
+	comp_cl_info(&comp_kpb, "kpb_pcm_convert_s24_c24_to_s24_c32() entry, samples %d", samples);
```
lyakh, line 1639: please make it `_dbg()`, this is too noisy

src/audio/kpb.c
```diff
@@ -1592,6 +1593,69 @@ static enum task_state kpb_draining_task(void *arg)
...
+	int ssize = audio_stream_sample_bytes(sink);
+	uint8_t *src = (uint8_t *)source + ioffset * 3;
+	int32_t *dst = audio_stream_wrap(sink, (int32_t *)sink->w_ptr + ooffset * ssize);
```
lyakh, line 1642: is this correct? It seems to add number of bytes to a 32-bit pointer, i.e. for each byte it will skip 4 bytes. Has the gcc version been tested? Could this be done?

src/audio/kpb.c
```diff
@@ -1626,6 +1692,51 @@ static void kpb_drain_samples(void *source, struct audio_stream __sparse_cache *
...
+		AE_SA24_L_IP(d24, align_out, out);
+	}
+	AE_SA64POS_FP(align_out, out);
+}
+#else
+static void kpb_convert_32b_to_24b(const struct audio_stream __sparse_cache *source, int ioffset,
+				   void *sink, int ooffset, unsigned int samples)
+{
+	int ssize = audio_stream_sample_bytes(source);
+	int32_t *src = audio_stream_wrap(source, (int32_t *)source->r_ptr + ioffset * ssize);
```
lyakh, line 1718: same question here

src/audio/kpb.c
```diff
@@ -1626,6 +1692,51 @@ static void kpb_drain_samples(void *source, struct audio_stream __sparse_cache *
...
+		n = samples - processed;
+		nmax = KPB_BYTES_TO_S32_SAMPLES(audio_stream_bytes_without_wrap(source, src));
+		n = MIN(n, nmax);
+		for (i = 0; i < n; i += 1) {
+			*dst = *src & 0xFF;
+			dst++;
+			*dst = (*src >> 8) & 0xFF;
+			dst++;
+			*dst = (*src >> 16) & 0xFF;
+			dst++;
```
lyakh, line 1734: I would write it as
```
	dst[0] = *src & 0xFF;
	dst[1] = (*src >> 8) & 0xFF;
	dst[2] = (*src >> 16) & 0xFF;
	dst += 3;
```
but I'm not sure if the compiler would generate better code from this

src/audio/kpb.c
lyakh: it's ok to decide not to address this, but then let's leave it unresolved too

lyakh (changes requested):

src/audio/kpb.c
```diff
@@ -1593,6 +1593,68 @@ static enum task_state kpb_draining_task(void *arg)
...
+	int i = 0;
+	ae_int24x2 d24 = AE_ZERO24();
+
+	if (!IS_ALIGNED((uintptr_t)out_ptr, 8)) {
+		AE_LA24_IP(d24, align_in, in);
+		ae_int32x2 d320 = d24;
+		int higher = AE_MOVAD32_H(d320);
+		*(ae_int32 *)(out_ptr) = higher << 8;
+		out_ptr = (ae_int32x2 *)(out + 4);
+		++i;
```
andrula-song, line 1616: it seems like the address jump of output stream hasn't been taken into consideration?

tmleman (approved):

fkwasowi:
SOFCI TEST

fkwasowi:
SOFCI TEST

src/audio/kpb.c
lyakh: could you please reply before resolving, thanks.

lyakh (changes requested):

lgirdwood:
@fkwasowi you may want to add .xxproject file type to gitignore. That way it will always be ignored by git.

src/audio/kpb.c
```diff
@@ -1593,6 +1593,68 @@ static enum task_state kpb_draining_task(void *arg)
...
+	ae_valign align_in = AE_LA64_PP(in);
+	int i = 0;
+	ae_int24x2 d24 = AE_ZERO24();
+
+	if (!IS_ALIGNED((uintptr_t)out_ptr, 8)) {
+		AE_LA24_IP(d24, align_in, in);
+		ae_int32x2 d320 = d24;
+		int higher = AE_MOVAD32_H(d320);
+		*(ae_int32 *)(out_ptr) = higher << 8;
+		out_ptr = (ae_int32x2 *)(out + 4);
```
andrula-song, line 1615: and for those instructions, 
"		ae_int32x2 d320 = d24;
		int higher = AE_MOVAD32_H(d320);
		*(ae_int32 *)(out_ptr) = higher << 8;
		out_ptr = (ae_int32x2 *)(out + 4); "
we can use AE_S32F24_L_IP(d24, out_ptr, 4); to replace that.

src/audio/kpb.c
```diff
@@ -1593,6 +1593,68 @@ static enum task_state kpb_draining_task(void *arg)
...
+	}
+	/* process two samples in single iteration to increase performance */
+	while (i < (int)n_samples - 1) {
+		AE_LA24X2_IP(d24, align_in, in);
+		ae_int32x2 d320 = d24;
+
+		d320 = AE_SLAI32(d320, 8);
+		AE_S32X2_XC(d320, out_ptr, 8);
+		i += 2;
+	}
```
andrula-song, line 1626: here too, use AE_S32X2F24_IP or AE_S32X2F24_XC

src/audio/kpb.c
```diff
@@ -1593,6 +1593,68 @@ static enum task_state kpb_draining_task(void *arg)
...
+	if (i != (int)n_samples) {
+		AE_LA24X2_IP(d24, align_in, in);
+		ae_int32x2 d320 = d24;
+		int higher = AE_MOVAD32_H(d320);
+		*(ae_int32 *)(out_ptr) = higher << 8;
```
andrula-song, line 1631: here too,  use AE_S32F24_L_IP(d24, out_ptr, 4)

src/audio/kpb.c
lyakh: isn't a `break` missing here in fact? Looks very much like it is.

src/audio/kpb.c
```diff
@@ -1715,6 +1845,65 @@ static inline bool kpb_is_sample_width_supported(uint32_t sampling_width)
...
+	uint8_t *in = audio_stream_wrap(source, (uint8_t *)source->r_ptr + ioffset * ssize);
+	uint8_t *out = audio_stream_wrap(sink, (uint8_t *)sink->w_ptr + ooffset * ssize);
+
+	const ae_int32x2 *sin = (const ae_int32x2 *)(in);
+	ae_int32x2 *sout = (ae_int32x2 *)(out);
+	ae_int32x2 vs = AE_ZERO32();
+
+	if (!IS_ALIGNED((uintptr_t)sin, 8)) {
+		AE_L32_IP(vs, (const ae_int32 *)(sin), 4);
+		AE_S32_L_IP(vs, (ae_int32 *)(sout), 4);
```
lyakh, line 1863: you really don't need parentheses around `sin` / `sout` / `in` / `out` - there are multiple such locations in the code

src/audio/kpb.c
lyakh: ...also unless a `break` is missing here too

src/audio/kpb.c
andrula-song: since the H part is the same with L part of d24, so both  AE_S32F24_L_IP(d24, out_ptr, 4); or  AE_S32F24_H_IP(d24, out_ptr, 4); is work.

src/audio/kpb.c
andrula-song: while use AE_S32X2F24*, please set parameter 3 as 8 instead of 4, for example, AE_S32X2F24_IP(d24, out_ptr, 8);

src/audio/kpb.c
andrula-song: here use AE_S32F24_H_IP(d24, out_ptr, 4) is because you load 2 24bit (the L and H part of d24 is different)

lyakh (approved):
please also at least lightly run some tests with a gcc build

mwasko:
SOFCI TEST

mwasko (changes requested):
It looks like MTL configuration does not build correctly  with this changes - CI Failue. @fkwasowi please check the error and fix the problem.

abonislawski:
Please run tests with both standard and HiFi4 paths

fkwasowi:
SOFCI TEST

fkwasowi:
SOFCI TEST

fkwasowi:
SOFCI TEST

fkwasowi:
> It looks like there is some KD test failure in TGL XTOS IPC3 configuration. @fkwasowi please take a look at this.

The multiple-pause-resume-5.sh test fails, but this is independent of the PR change, as there is no kpb module in the topology in this test.

fkwasowi:
SOFCI TEST

src/include/sof/audio/kpb.h
```diff
@@ -12,6 +12,14 @@
 #include <user/trace.h>
 #include <stdint.h>
 
+#if defined(__XCC__)
+
+#include <xtensa/config/core-isa.h>
+#if XCHAL_HAVE_HIFI3 || XCHAL_HAVE_HIFI4
+#define KPB_HIFI4
```
abonislawski, line 19: so this is HIFI3 actually, valid for TGL too, please rename

abonislawski (changes requested):
the problem was in kd tgl test but looks like its okay after yesterday push.
Pause-resume is fine for this PR

src/include/sof/audio/kpb.h
fkwasowi: Done

kv2019i:
The multiple-pause failure is likely not related to this PR:
[  616.962191] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc timed out for 0x13000004|0x1
... but is basicly this https://github.com/thesofproject/sof/issues/6864

@abonislawski  @fkwasowi let me know when this is ready and should we ignore Michal's -1 to do merge this week.

abonislawski (approved):
> let me know when this is ready

waiting only for green internal CI

PR #6786, opened 2022-12-13
topology2: add sdw spk feedback pcm

bardliao:
This commit create a PCM that can pass amplifier feedback to user space.

### DIFF
--- a/tools/topology/topology2/platform/intel/alh-2nd-spk.conf
+++ b/tools/topology/topology2/platform/intel/alh-2nd-spk.conf
@@ -12,4 +12,19 @@ Object.Pipeline {
 			node_type $ALH_LINK_OUTPUT_CLASS
 		}
 	}
+	passthrough-be.$ALH_2ND_SPK_IN_ID {
+		direction	"capture"
+		index $ALH_2ND_SPK_IN_ID
+		copier_type "ALH"
+		Object.Widget.pipeline.1 {
+			stream_name	'SPK IN stream'
+		}
+		Object.Widget.copier.1 {
+			stream_name	$SDW_SPK_IN_STREAM
+			dai_type	"ALH"
+			copier_type	"ALH"
+			type		"dai_out"
+			node_type $ALH_LINK_INPUT_CLASS
+		}
+	}
 }
--- a/tools/topology/topology2/platform/intel/sdw-amp-generic.conf
+++ b/tools/topology/topology2/platform/intel/sdw-amp-generic.conf
@@ -2,8 +2,11 @@
 
 Define {
 	SDW_SPK_STREAM 'SDW1-Playback'
+	SDW_SPK_IN_STREAM 'SDW1-Capture'
 	ALH_2ND_SPK_ID 22
+	ALH_2ND_SPK_IN_ID 32
 	SDW_AMP_BE_ID 2
+	SDW_AMP_IN_BE_ID 3
 }
 
 Object.Dai {
@@ -17,6 +20,18 @@ Object.Dai {
 
 		Object.Base.hw_config."ALH514" {
 			id	0
+		}
+	}
+	ALH."515" {
+		id 		$SDW_AMP_IN_BE_ID
+		direction	"capture"
+		name		$SDW_SPK_IN_STREAM
+		default_hw_conf_id	1
+		rate			48000
+		channels		2
+
+		Object.Base.hw_config."ALH515" {
+			id	1
 		}
 	}
 }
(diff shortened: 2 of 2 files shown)


RanderWang:
in 2ch playback case, feedback will be 4ch or 2ch ?

bardliao:
> in 2ch playback case, feedback will be 4ch or 2ch ?

It depends on codec's capability. Usually, an amplifier can provide 1ch for I data and 1ch for V data. So, 2ch for 1 amplifier and 4ch for 2 amplifiers. However, we may not have enough bandwidth for 2 or 4 ch playback and 4 ch capture.
So, I suggest that we set 2ch for the feedback capture first, and add 4ch feedback capture support after we solve the sdw bandwidth issue.

lgirdwood:
adding @plbossart

plbossart:
@bardliao how does this work if the two amplifiers are on separate links? does the gateway/copier stuff automagically group the feedback in a 4ch stream?

bardliao:
> @bardliao how does this work if the two amplifiers are on separate links? does the gateway/copier stuff automagically group the feedback in a 4ch stream?

Yes, I think so. But I can't test it with 4ch due to SoundWire bus bandwidth limitation.

RanderWang:
> > @bardliao how does this work if the two amplifiers are on separate links? does the gateway/copier stuff automagically group the feedback in a 4ch stream?
> 
> Yes, I think so. But I can't test it with 4ch due to SoundWire bus bandwidth limitation.

yes, It is supported by ref fw with channel map setting for each amplifier, but now cSOF doesn't support it. We need to add it in shopping list.

RanderWang (approved):

lgirdwood (approved):

plbossart:
> > @bardliao how does this work if the two amplifiers are on separate links? does the gateway/copier stuff automagically group the feedback in a 4ch stream?
> 
> Yes, I think so. But I can't test it with 4ch due to SoundWire bus bandwidth limitation.

what prevents you from running faster? https://github.com/thesofproject/linux/pull/4057 should help, no?

plbossart:
> > > @bardliao how does this work if the two amplifiers are on separate links? does the gateway/copier stuff automagically group the feedback in a 4ch stream?
> > 
> > 
> > Yes, I think so. But I can't test it with 4ch due to SoundWire bus bandwidth limitation.
> 
> what prevents you from running faster? [thesofproject/linux#4057](https://github.com/thesofproject/linux/pull/4057) should help, no?

Also can we not limit the resolution to e.g. 16 bits for the feedback stream. It's usually not really demanding in terms of bit depth.

RanderWang:
> > > > @bardliao how does this work if the two amplifiers are on separate links? does the gateway/copier stuff automagically group the feedback in a 4ch stream?
> > > 
> > > 
> > > Yes, I think so. But I can't test it with 4ch due to SoundWire bus bandwidth limitation.
> > 
> > 
> > what prevents you from running faster? [thesofproject/linux#4057](https://github.com/thesofproject/linux/pull/4057) should help, no?
> 
> Also can we not limit the resolution to e.g. 16 bits for the feedback stream. It's usually not really demanding in terms of bit depth.

It is a issue that driver fixes the BE format to 32 bits in dai fixup. This makes driver simple and unified, but it will introduce bandwidth issue for SDW bus.  Actually we only need to fix 32bits BE format for I2S and use runtime format for other gateways. Also we have debugged the HDA format issue that we use 32bits but codec use 24bits and we don't need to use 32bit for most 16bit pcm streams. If it is reasonable, I can submit a patch to only fix I2S to 32bits in driver (not just change a word, need to adjust the format computation for each gateway).

bardliao:
> > > @bardliao how does this work if the two amplifiers are on separate links? does the gateway/copier stuff automagically group the feedback in a 4ch stream?
> > 
> > 
> > Yes, I think so. But I can't test it with 4ch due to SoundWire bus bandwidth limitation.
> 
> what prevents you from running faster? [thesofproject/linux#4057](https://github.com/thesofproject/linux/pull/4057) should help, no?

Thanks, @plbossart it helps. Unfortunately, it doesn't work as I expect. Only channel 1 and 2 have data, channel 3 and 4 are empty.

kv2019i (approved):

plbossart:
Not sure why this was merged @kv2019i, clearly there are some issues here.

plbossart:
> > > > > @bardliao how does this work if the two amplifiers are on separate links? does the gateway/copier stuff automagically group the feedback in a 4ch stream?
> > > > 
> > > > 
> > > > Yes, I think so. But I can't test it with 4ch due to SoundWire bus bandwidth limitation.
> > > 
> > > 
> > > what prevents you from running faster? [thesofproject/linux#4057](https://github.com/thesofproject/linux/pull/4057) should help, no?
> > 
> > 
> > Also can we not limit the resolution to e.g. 16 bits for the feedback stream. It's usually not really demanding in terms of bit depth.
> 
> It is a issue that driver fixes the BE format to 32 bits in dai fixup. This makes driver simple and unified, but it will introduce bandwidth issue for SDW bus. Actually we only need to fix 32bits BE format for I2S and use runtime format for other gateways. Also we have debugged the HDA format issue that we use 32bits but codec use 24bits and we don't need to use 32bit for most 16bit pcm streams. If it is reasonable, I can submit a patch to only fix I2S to 32bits in driver (not just change a word, need to adjust the format computation for each gateway).

Yes, this doesn't look quite right to always use 32 bits. This is specifically problematic for this feedback, and in general 32-bits on a link is just madness. 32 bits makes sense as an intermediate format when there's processing involved in the DSP, so that there's enough headroom to avoid saturation. No one can hear the difference between 32 and 24 bits once the data reach the amplifier.

kv2019i:
@plbossart There were two approvals and tests passed, so I did not see a reason to not merge. Please @bardliao submit fixes as follow-up or revert if this takes longer to address.

bardliao:
@plbossart @kv2019i The fixup topology is https://github.com/thesofproject/sof/pull/6854. However, we need to set different ch_mask for different amplifiers on kernel. I will create a PR on kernel side.

PR #6787, opened 2022-12-13
topology: hdmi-generic: change HDMI PCM id

bardliao:
Define HDMI_PCM_ID macro so that we can have different pcm id for different topology. And change the hdmi pcm id of sof-hda-generic.tplg to align with UCM.


tools/topology/topology2/sof-hda-generic.conf
```diff
@@ -58,6 +58,10 @@ Define {
 	HDMI1_ID	1
 	HDMI2_ID	2
 	HDMI3_ID	3
+	HDMI1_PCM_ID	3
+	HDMI2_PCM_ID	4
+	HDMI3_PCM_ID	5
+	HDMI4_PCM_ID	6
```
plbossart, line 64: IIRC we never enabled 4 streams for HDMI in sof-hda-generic, did we?

And also I am confused by the HDMI1_ID being 1, is this correct? I thought we had the HDaudio codec Dailink first, and HDMI last?

bardliao: > IIRC we never enabled 4 streams for HDMI in sof-hda-generic, did we?

I will remove HDMI4_PCM_ID from sof-hda-generic.conf.

> 
> And also I am confused by the HDMI1_ID being 1, is this correct? I thought we had the HDaudio codec Dailink first, and HDMI last?

No, we have HDMI first. see [skl_hda_be_dai_links](https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/intel/boards/skl_hda_dsp_common.c#L82)
```
iDisp1
iDisp2
iDisp3
Analog Playback and Capture
Digital Playback and Capture
dmic01
dmic16k
```

plbossart: ok

tools/topology/topology2/platform/intel/hdmi-default.conf
```diff
@@ -4,6 +4,10 @@ Define {
 	HDMI2_ID			6
 	HDMI3_ID			7
 	HDMI4_ID			8
+	HDMI1_PCM_ID			5
+	HDMI2_PCM_ID			6
+	HDMI3_PCM_ID			7
+	HDMI4_PCM_ID			8
```
plbossart, line 10: I am still confused on this one. Here we're adding the defintions...

tools/topology/topology2/platform/intel/hdmi-generic.conf
```diff
@@ -121,7 +121,7 @@ Object.PCM {
 	}
 	pcm.7 {
 		name HDMI3
-		id 7
+		id $HDMI3_PCM_ID
```
plbossart, line 124: but here we are not adding the HDMI4 in a file that's supposed to be generic. So doesn't this result in HDMI4 NEVER being added?

FWIW, I would recommend not adding HDMI4, it's useless in practice and it makes our life problematic. If we can we should limit all topology2 new files to only use HDMI1..3

bardliao: > but here we are not adding the HDMI4 in a file that's supposed to be generic. So doesn't this result in HDMI4 NEVER being added?

HDMI4 pcm will be added in hdmi-4th.conf if needed.

> 
> FWIW, I would recommend not adding HDMI4, it's useless in practice and it makes our life problematic. If we can we should limit all topology2 new files to only use HDMI1..3

Even for SoundWire + HDMI topology?

plbossart: yes, this is painful so we should only allow for 3 HDMI streams. Adding @kv2019i to validate this direction.

kv2019i (commented):
Comments inline.

plbossart:
@kv2019i for sof-sdw we only have 3 devices in the UCM conf. So there's no issue if we stop using the 4th pipe for audio for new platforms

````
# Use case Configuration for sof-soundwire

Include.hdmi.File "/codecs/hda/hdmi.conf"

If.hdmi1 {
	Condition { Type AlwaysTrue }
	True.Macro.hdmi1.HDMI {
		Number 1
		Device 5
		Priority 500
	}
}

If.hdmi2 {
	Condition { Type AlwaysTrue }
	True.Macro.hdmi2.HDMI {
		Number 2
		Device 6
		Priority 600
	}
}

If.hdmi3 {
	Condition { Type AlwaysTrue }
	True.Macro.hdmi2.HDMI {
		Number 3
		Device 7
		Priority 700
	}
}
````

**Comment by bardliao** [comment]
@plbossart I added a commit to change NUM_HDMIS from 4 to 3 for all existing topologies, but keep hdmi-4th.conf in case we want to use it in the future. I also add other commits to fix the pipeline id conflict between sdw speaker and hdmi and add hdmi support to the existing sdw topologies.

File: tools/topology/topology2/platform/intel/hdmi-default.conf [hunk 4]
```diff
@@ -4,28 +4,28 @@ Define {
...
+	HDMI4_DAI_PIPELINE_ID		81
+	HDMI1_PCM_CAPS			"hdmi1 Playback"
+	HDMI1_PIPELINE_STREAM_NAME	"hdmi1 stream"
+	HDMI2_PCM_CAPS			"hdmi2 Plbyback"
+	HDMI2_PIPELINE_STREAM_NAME	"hdmi2 stream"
+	HDMI3_PCM_CAPS			"hdmi3 Playback"
+	HDMI3_PIPELINE_STREAM_NAME	"hdmi3 stream"
+	HDMI4_PCM_CAPS			"hdmi4 Playback"
+	HDMI4_PIPELINE_STREAM_NAME	"hdmi4 stream"
+	HDMI1_HOST_PIPELINE_SINK	"copier.host.50.1"
```
lgirdwood, line 27: @bardliao @jsarha I assume we can change 
```"copier.host.50.1"``` to ```"copier.host.$HDMI1_HOST_PIPELINE_ID.1"```
i.e. we dereference the ```HDMI1_HOST_PIPELINE_ID``` assignment above ?

File: tools/topology/topology2/platform/intel/hdmi-default.conf [hunk 5]
```diff
@@ -4,28 +4,28 @@ Define {
...
-	HDMI4_DAI_PIPELINE_SRC		"copier.HDA.27.1"
+	HDMI1_HOST_PIPELINE_ID		50
+	HDMI1_DAI_PIPELINE_ID		51
+	HDMI2_HOST_PIPELINE_ID		60
+	HDMI2_DAI_PIPELINE_ID		61
+	HDMI3_HOST_PIPELINE_ID		70
+	HDMI3_DAI_PIPELINE_ID		71
+	HDMI4_HOST_PIPELINE_ID		80
+	HDMI4_DAI_PIPELINE_ID		81
+	HDMI1_PCM_CAPS			"hdmi1 Playback"
```
plbossart, line 19: HDMI in capital letters

tools/topology/topology2/platform/intel/hdmi-default.conf
```diff
@@ -4,28 +4,28 @@ Define {
...
+	HDMI1_PIPELINE_STREAM_NAME	"hdmi1 stream"
+	HDMI2_PCM_CAPS			"hdmi2 Plbyback"
```
plbossart, line 21: typo: Playback

tools/topology/topology2/platform/intel/hdmi-generic.conf
plbossart: we can simplify for 3 HDMIs for all SoundWire topologies. It's fine if the machine driver exposes 4 and the topology uses only 3. What doesn't work is the other way around, a topology using more dailinks than what is exposed by the machine driver.

tools/topology/topology2/sof-ace-tplg/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ DEEPBUFFER_FW_DMA_MS=100"
 "cavs-sdw\;mtl-sdw"
 
 # SDW + HDMI topology with passthrough pipelines
-"cavs-sdw\;mtl-sdw-hdmi\;NUM_HDMIS=4"
+"cavs-sdw\;mtl-sdw-hdmi\;NUM_HDMIS=3"
```
plbossart, line 22: this topology name needs to be fixed or this needs to moved to a different develop, it's not used by the kernel.

plbossart (changes requested):
couple of comments below @bardliao

tools/topology/topology2/platform/intel/hdmi-default.conf
bardliao: > @bardliao @jsarha I assume we can change `"copier.host.50.1"` to `"copier.host.$HDMI1_HOST_PIPELINE_ID.1"` i.e. we dereference the `HDMI1_HOST_PIPELINE_ID` assignment above ?

That is what I was looking for. But according to @ranj063, we don't support it yet.

bardliao:
@kv2019i @miRoox Can you take a look the "topology2: move development topologies to development folder" commit? Some topologies are moved to development folder. I think CI will be impacted.

tools/topology/topology2/platform/intel/hdmi-default.conf
lgirdwood: > > @bardliao @jsarha I assume we can change `"copier.host.50.1"` to `"copier.host.$HDMI1_HOST_PIPELINE_ID.1"` i.e. we dereference the `HDMI1_HOST_PIPELINE_ID` assignment above ?
> 
> That is what I was looking for. But according to @ranj063, we don't support it yet.

ok, got you - @jsarha can you confirm this is in your topology improvement work.

lgirdwood:
@plbossart good for you ?

plbossart (approved):

plbossart:
@ranj063 can you please double-check this PR?

lgirdwood (approved):

marc-hb:
This caused regression
- #6947

Tentative fix that disables HDMI:
- #6946

kv2019i:
This also causes regression on ADL SDW config and was picked up by CI -> https://sof-ci.01.org/sofpr/PR6931/build3353/devicetest/index.html?model=ADLP_RVP_SDW_IPC4ZPH&testcase=verify-kernel-boot-log  and https://sof-ci.01.org/sofpr/PR6787/build3161/devicetest/index.html?model=ADLP_RVP_SDW_IPC4ZPH&testcase=verify-kernel-boot-log

@bardliao  @lgirdwood  Unless we have a quick fix, we need to revert his.

lgirdwood:
Ok, lets revert. @bardliao I guess you have enough data to root cause.

marc-hb:
Revert submitted by @lgirdwood  in #6953

Indeed this was merged despite ADLP_RVP_SDW_IPC4ZPH totally failing https://sof-ci.01.org/sofpr/PR6787/build3161/devicetest/index.html !

@bardliao please convert failing PRs to "draft" to block merges of failing PRs and help overloaded reviewers (also: please submit a revert if there's no quick fix)

bardliao:
> Revert submitted by @lgirdwood in #6953
> 
> Indeed this was merged despite ADLP_RVP_SDW_IPC4ZPH totally failing in https://sof-ci.01.org/sofpr/PR6787/build3161/devicetest/index.html !
> 
> @bardliao please convert failing PRs to "draft" to block merges of failing PRs and help overloaded reviewers (also: please submit a revert if there's no quick fix)

@marc-hb Sorry for the regressions. The problem is that as I mentioned on https://github.com/thesofproject/sof/pull/6787#issuecomment-1359240839, the CI will be impacted, so the PR test result was not reliable for this PR. 
I do post fix PRs in some hour when I was informed that this PR causes some regressions and that's why I didn't submit a revert.

tools/topology/topology2/platform/intel/hdmi-default.conf
jsarha: @bardliao @lgirdwood that is not exactly what I trying to achieve just yet.  So far I've been looking for a way to implement variable substitution for alsa control names, not generic widget name substitution.

marc-hb:
> the CI will be impacted, so the PR test result was not reliable for this PR.

In situations with a CI dependency like this we should coordinate and make the CI changes first (after local testing and approvals). Then use "SOFCI TEST" in 6787. If the tests pass then 6787 can be merged, if tests still don't pass then CI changes should be reverted.

Another possibility is to temporarily disable some tests in some configurations.

> I do post fix PRs in some hour when I was informed that this PR causes some regressions and that's why I didn't submit a revert.

"Quick" fixes are great except they must be tested, reviewed etc. which means they're almost never quick.

PR #6788, opened 2022-12-13
module adapter: volume: Validation of channels count

softwarecki:
The module operates on arrays whose size is based on SOF_IPC_MAX_CHANNELS. Added check if the number of channels given in the configuration does not exceed the number of maximum supported channels.

Signed-off-by: Adrian Warecki <adrian.warecki@intel.com>

### DIFF
--- a/src/audio/module_adapter/module/volume/volume.c
+++ b/src/audio/module_adapter/module/volume/volume.c
@@ -551,6 +551,12 @@ static int volume_init(struct processing_module *mod)
 	uint8_t channel;
 	uint32_t instance_id;
 
+	channels_count = mod->priv.cfg.base_cfg.audio_fmt.channels_count;
+	if (channels_count > SOF_IPC_MAX_CHANNELS) {
+		comp_err(dev, "volume_init(): Invalid channels count %u", channels_count);
+		return -EINVAL;
+	}
+
 	cd = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(struct vol_data));
 	if (!cd)
 		return -ENOMEM;
@@ -567,8 +573,6 @@ static int volume_init(struct processing_module *mod)
 	}
 
 	md->private = cd;
-
-	channels_count = mod->priv.cfg.base_cfg.audio_fmt.channels_count;
 
 	for (channel = 0; channel < channels_count ; channel++) {
 		if (vol->config[0].channel_id == IPC4_ALL_CHANNELS_MASK)
@@ -922,7 +926,8 @@ static int volume_set_config(struct processing_module *mod, uint32_t config_id,
 	struct module_data *md = &mod->priv;
 	struct comp_dev *dev = mod->dev;
 	struct ipc4_peak_volume_config cdata;
-	int i, ret;
+	int ret;
+	uint32_t i, channels_count;
 
 	comp_dbg(dev, "volume_set_config()");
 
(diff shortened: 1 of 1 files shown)


src/audio/module_adapter/module/volume/volume.c
```diff
@@ -569,6 +569,12 @@ static int volume_init(struct processing_module *mod)
 	md->private = cd;
 
 	channels_count = mod->priv.cfg.base_cfg.audio_fmt.channels_count;
+	if (channels_count >= SOF_IPC_MAX_CHANNELS) {
```
ranj063, line 572: why not do this check before line 554 so you dont have to free anything

marcinszkudlinski (approved):

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -551,6 +551,12 @@ static int volume_init(struct processing_module *mod)
 	uint8_t channel;
 	uint32_t instance_id;
 
+	channels_count = mod->priv.cfg.base_cfg.audio_fmt.channels_count;
+	if (channels_count >= SOF_IPC_MAX_CHANNELS) {
```
makarukp, line 555: Shouldn't be "channels_count > SOF_IPC_MAX_CHANNELS" ?

makarukp (changes requested):

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -941,13 +946,20 @@ static int volume_set_config(struct processing_module *mod, uint32_t config_id,
...
 
+	if ((cdata.channel_id != IPC4_ALL_CHANNELS_MASK) &&
+	    (cdata.channel_id >= SOF_IPC_MAX_CHANNELS))
+		return -EINVAL;
+
 	init_ramp(cd, cdata.curve_duration, cdata.target_volume);
 	cd->ramp_finished = true;
 
+	channels_count = mod->priv.cfg.base_cfg.audio_fmt.channels_count;
+	assert(channels_count < SOF_IPC_MAX_CHANNELS);
```
abonislawski, line 957: Why assert here? Another check looks different
`if (channels_count >= SOF_IPC_MAX_CHANNELS)`

makarukp: I think that according to previous check (channels_count <= SOF_IPC_MAX_CHANNELS) should be asserted

abonislawski: I mean why we have _assert_ here instead of the same _if_,
FW should return error if set config failed instead of crash.

lgirdwood: ack, this is non fatal (just a bad topology) so FW should not crash but report up the stack.

abonislawski (approved):

makarukp (approved):

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -551,6 +551,12 @@ static int volume_init(struct processing_module *mod)
 	uint8_t channel;
```
kv2019i, line 551: My usual nitpick @softwarecki why the weird 40-50 line word-wrapping :)

kv2019i (approved):

kv2019i:
Subjective checkpatch complains about extra parenthesis, but seems something at least I can live with. Proceeding with merge.

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -941,13 +946,25 @@ static int volume_set_config(struct processing_module *mod, uint32_t config_id,
 	cdata = *(const struct ipc4_peak_volume_config *)fragment;
 	cdata.target_volume = convert_volume_ipc4_to_ipc3(dev, cdata.target_volume);
 
+	if ((cdata.channel_id != IPC4_ALL_CHANNELS_MASK) &&
+	    (cdata.channel_id >= SOF_IPC_MAX_CHANNELS)) {
```
ranj063, line 950: why is this check not consistent? >= here but just > in the other 2 changes?

softwarecki: Here is checked channel_id, in another places channels_count was checked.

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -551,6 +551,12 @@ static int volume_init(struct processing_module *mod)
 	uint8_t channel;
 	uint32_t instance_id;
 
+	channels_count = mod->priv.cfg.base_cfg.audio_fmt.channels_count;
+	if (channels_count > SOF_IPC_MAX_CHANNELS) {
```
lyakh, line 555: does this have to be checked twice? Can the configuration be supplied both in `.init()` and in `.set_configuration()`?

kv2019i: @lyakh Given they are separate ops, it seems hard to be absolutely sure and seems safer to check.

PR #6791, opened 2022-12-13
ace: fix cache-uncache conversion macros for sparse

lyakh:
cache_to_uncache() and uncache_to_cache() must be inline functions with proper argument annotations. Fix ACE definitions by alanogy with cAVS.

### DIFF
--- a/src/library_manager/lib_manager.c
+++ b/src/library_manager/lib_manager.c
@@ -12,6 +12,7 @@
 #include <sof/audio/buffer.h>
 #include <sof/audio/component_ext.h>
 #include <sof/common.h>
+#include <sof/compiler_attributes.h>
 #include <sof/ipc/topology.h>
 
 #include <sof/sof.h>
@@ -54,19 +55,19 @@ struct lib_manager_dma_ext {
 
 static struct ext_library loader_ext_lib;
 
-static int lib_manager_load_data_from_storage(uint8_t *vma, uint8_t *s_addr,
+static int lib_manager_load_data_from_storage(void __sparse_cache *vma, void *s_addr,
 					      uint32_t size, uint32_t flags)
 {
-	int ret = sys_mm_drv_map_region(vma, POINTER_TO_UINT(NULL),
+	int ret = sys_mm_drv_map_region((__sparse_force void *)vma, POINTER_TO_UINT(NULL),
 					size, SYS_MM_MEM_PERM_RW | SYS_MM_MEM_PERM_EXEC);
 	if (ret < 0)
 		return ret;
 
-	ret = memcpy_s((void *)vma, size, (void *)s_addr, size);
+	ret = memcpy_s((__sparse_force void *)vma, size, s_addr, size);
 	if (ret < 0)
 		return ret;
 
-	dcache_writeback_region((void *)vma, size);
+	dcache_writeback_region(vma, size);
 
 	/* TODO: Change attributes for memory to FLAGS  */
 	return 0;
@@ -78,10 +79,12 @@ static int lib_manager_load_module(uint32_t module_id, struct sof_man_module *mo
 	struct ext_library *ext_lib = ext_lib_get();
 	uint32_t lib_id = LIB_MANAGER_GET_LIB_ID(module_id);
 	size_t load_offset = (size_t)((void *)ext_lib->desc[lib_id]);
-	void *va_base_text = (void *)mod->segment[SOF_MAN_SEGMENT_TEXT].v_base_addr;
+	void __sparse_cache *va_base_text = (void __sparse_cache *)
+		mod->segment[SOF_MAN_SEGMENT_TEXT].v_base_addr;
 	void *src_txt = (void *)(mod->segment[SOF_MAN_SEGMENT_TEXT].file_offset + load_offset);
 	size_t st_text_size = mod->segment[SOF_MAN_SEGMENT_TEXT].flags.r.length;
-	void *va_base_rodata = (void *)mod->segment[SOF_MAN_SEGMENT_RODATA].v_base_addr;
+	void __sparse_cache *va_base_rodata = (void __sparse_cache *)
+		mod->segment[SOF_MAN_SEGMENT_RODATA].v_base_addr;
 	void *src_rodata =
 		(void *)(mod->segment[SOF_MAN_SEGMENT_RODATA].file_offset + load_offset);
 	size_t st_rodata_size = mod->segment[SOF_MAN_SEGMENT_RODATA].flags.r.length;
@@ -125,8 +128,8 @@ static int lib_manager_load_module(uint32_t module_id, struct sof_man_module *mo
 	return 0;
 
 err:
-	sys_mm_drv_unmap_region(va_base_text, st_text_size);
-	sys_mm_drv_unmap_region(va_base_rodata, st_rodata_size);
+	sys_mm_drv_unmap_region((__sparse_force void *)va_base_text, st_text_size);
+	sys_mm_drv_unmap_region((__sparse_force void *)va_base_rodata, st_rodata_size);
 
 	return ret;
 }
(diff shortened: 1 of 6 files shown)


lgirdwood:
@lyakh looks like this redues teh MTL sparse error to Zephyr only
```
Reminder: to see ALL warnings you must as usual build _from scratch_
/zep_workspace/zephyr/soc/xtensa/intel_adsp/ace/power.c:190:17: warning: incorrect type in argument 1 (different address spaces)
/zep_workspace/zephyr/subsys/logging/backends/log_backend_adsp_mtrace.c:69:13: warning: incorrect type in argument 1 (different address spaces)
/zep_workspace/zephyr/subsys/logging/backends/log_backend_adsp_mtrace.c:73:9: warning: incorrect type in argument 1 (different address spaces)
/zep_workspace/zephyr/subsys/logging/backends/log_backend_adsp_mtrace.c:78:67: warning: incorrect type in argument 1 (different address spaces)
/zep_workspace/zephyr/drivers/mm/mm_drv_intel_adsp_mtl_tlb.c:374:20: warning: incorrect type in initializer (different address spaces)
/zep_workspace/zephyr/drivers/mm/mm_drv_intel_adsp_mtl_tlb.c:438:20: warning: incorrect type in initializer (different address spaces)
/zep_workspace/zephyr/drivers/mm/mm_drv_intel_adsp_mtl_tlb.c:529:24: warning: incorrect type in initializer (different address spaces)
/zep_workspace/zephyr/drivers/mm/mm_drv_intel_adsp_mtl_tlb.c:530:24: warning: incorrect type in initializer (different address spaces)
/zep_workspace/zephyr/drivers/mm/mm_drv_intel_adsp_mtl_tlb.c:542:18: warning: incorrect type in assignment (different address spaces)
/zep_workspace/zephyr/drivers/mm/mm_drv_intel_adsp_mtl_tlb.c:543:18: warning: incorrect type in assignment (different address spaces)
/zep_workspace/zephyr/drivers/mm/mm_drv_intel_adsp_mtl_tlb.c:640:24: warning: incorrect type in initializer (different address spaces)
/zep_workspace/zephyr/drivers/mm/mm_drv_intel_adsp_mtl_tlb.c:641:24: warning: incorrect type in initializer (different address spaces)
/zep_workspace/zephyr/drivers/mm/mm_drv_intel_adsp_mtl_tlb.c:802:33: warning: incorrect type in argument 1 (different address spaces)
Error: Process completed with exit code 1.
```

lyakh:
> @lyakh looks like this redues teh MTL sparse error to Zephyr only

@lgirdwood exactly. The fixes aren't perfect, particularly to the lib-manager. I cut the propagation of `__sparse_cache` attributes at some point, by forced type-casts, but maybe we rather should propagate them further. E.g. maybe we should adapt `sys_mm_drv_*()` functions to use cached addresses on Xtensa, but I don't really know whether those addresses really must be cached. Also `ext_lib_get()->desc` pointers really are cached, but labeling them as such was a bit difficult, so I left them as "generic" / uncached for now too. If desired, we can change that in a follow-up

src/platform/intel/ace/include/ace/lib/mailbox.h
```diff
@@ -39,29 +39,29 @@
 static inline void mailbox_sw_reg_write(size_t offset, uint32_t src)
 {
 	volatile uint32_t *ptr;
-	volatile uint32_t __sparse_cache *ptr_c;
+	uint32_t __sparse_cache *ptr_c;
 
-	ptr_c = (volatile uint32_t __sparse_cache *)(MAILBOX_SW_REG_BASE + offset);
+	ptr_c = (uint32_t __sparse_cache *)(MAILBOX_SW_REG_BASE + offset);
```
kv2019i, line 44: This looks a bit weird to me (we are casting volatile back), but compiler seems to be happy, so I guess this is ok. You live and learn.

src/platform/intel/ace/include/ace/lib/memory.h
```diff
@@ -40,14 +40,19 @@
 #define SRAM_ALIAS_OFFSET		0x60000000
```
kv2019i, line 40: typoe: alanogy in commit message

src/library_manager/lib_manager.c
```diff
@@ -366,11 +379,10 @@ int lib_manager_register_module(struct sof_man_fw_desc *desc, int module_id)
 	return ret;
 }
 
-static void lib_manager_dma_buffer_update(struct lib_manager_dma_buf *buffer, uintptr_t addr,
+static void lib_manager_dma_buffer_update(struct lib_manager_dma_buf *buffer,
 					  uint32_t size)
 {
 	buffer->size = size;
-	buffer->addr = addr;
 	buffer->w_ptr = buffer->addr;
```
kv2019i, line 386: @lyakh  Is this removal intentional? This seems like a bigger change than just adding annotations?

src/library_manager/lib_manager.c
```diff
@@ -388,14 +400,12 @@ static int lib_manager_dma_buffer_init(struct lib_manager_dma_buf *buffer, uint3
...
+	dcache_writeback_region((__sparse_force void __sparse_cache *)buffer->addr, size);
 
-	/* initialise the DMA buffer */
+	tr_dbg(&lib_manager_tr, "lib_manager_dma_buffer_init(): %#lx, %#lx",
+	       buffer->addr, buffer->end_addr);
 
-	lib_manager_dma_buffer_update(buffer, buffer->addr, size);
-	tr_dbg(&lib_manager_tr, "lib_manager_dma_buffer_init(): %p, %p",
-	       (void *)buffer->addr, (void *)buffer->end_addr);
 	return 0;
```
kv2019i, line 409: Ditto here.

src/library_manager/lib_manager.c
```diff
@@ -509,98 +519,98 @@ static int lib_manager_load_data_from_host(struct lib_manager_dma_ext *dma_ext,
...
 
 	while (copied_bytes < dst_size) {
+		uint32_t bytes_to_copy;
+		int ret;
+
 		if ((dst_size - copied_bytes) > MAN_MAX_SIZE_V1_8)
 			bytes_to_copy = MAN_MAX_SIZE_V1_8;
 		else
-			bytes_to_copy = (dst_size - copied_bytes);
+			bytes_to_copy = dst_size - copied_bytes;
```
kv2019i, line 542: Unrelated changes creeping in.

src/platform/intel/ace/include/ace/lib/mailbox.h
lyakh: well, first of all you know "volatile is evil." But the point here is that that pointer is never used to actually access memory, so no need for `volatile` there at all

src/library_manager/lib_manager.c
lyakh: yes, intentional. Yeah, there are some fixes like this. This assignment is redundant, this function was always called with `(buf, buf->addr,...)` so it was an assignment to itself. That's why I removed that `addr` argument completely. The change seemed small enough to me to integrate it...

src/library_manager/lib_manager.c
lyakh: ok, I could split this into 2: a clean up and sparse stuff, would that help?

kv2019i: @lyakh you can keep it, it's just a bit distracting to have these in bigger patches.

src/platform/intel/ace/include/ace/lib/mailbox.h
marc-hb: > "volatile is evil."

I think "harmful" is more popular hence a better search keyword: https://www.bing.com/search?q=volatile+considered+harmful

src/library_manager/lib_manager.c
marc-hb: It's not just a reviewer distraction, it also means no validation will never test these separately.

marc-hb (commented):
>  looks like this redues teh MTL sparse error to Zephyr only

You forgot the link to the corresponding logs: https://github.com/thesofproject/sof/actions/runs/3695828516/jobs/6258707347

src/library_manager/lib_manager.c
kv2019i: Ack, @jxstelter @pjdobrowolski @softwarecki can you check?

kv2019i (approved):
Ack, thanks @lyakh . I'd like some of the lib-manager authors to go through the changes as well.

src/library_manager/lib_manager.c
```diff
@@ -509,98 +519,98 @@ static int lib_manager_load_data_from_host(struct lib_manager_dma_ext *dma_ext,
...
 
-	/* allocate new buffer */
-	local_add = rballoc_align(0, caps, size, addr_align);
+	/* allocate new buffer: cached alias */
+	local_add = (__sparse_force void __sparse_cache *)rballoc_align(0, caps, size, addr_align);
 #endif
 	if (!local_add) {
 		tr_err(&lib_manager_tr, "dma_buffer_init(): alloc failed");
-		return -ENOMEM;
+		return NULL;
```
softwarecki, line 577: You changed the function declarations to return void, so why are you returning a value?

src/library_manager/lib_manager.c
```diff
@@ -509,98 +519,98 @@ static int lib_manager_load_data_from_host(struct lib_manager_dma_ext *dma_ext,
...
 	}
 
-	*address = (uint32_t)local_add;
-	return 0;
+	return local_add;
```
softwarecki, line 580: same as above

src/library_manager/lib_manager.c
lyakh: @softwarecki it's returning `void *`

softwarecki: my mistake, sorry

src/platform/intel/ace/include/ace/lib/mailbox.h
lyakh: > if I am sharing memory between devices (to a context where the compiler doesn't see that the memory will be accessed) then volatile is appropriate.

@paulstelian97 not sure which "devices" you mean. I think `volatile` is supposed to tell the compiler that contents of that memory can change or have effects outside of the scope of this code, so don't optimise access out.
So e.g.
```
*v = 1;
*v = 2;
```
would normally look redundant to the compiler and it might throw the first assignment away, but if the `v` pointer is declared as `volatile`, it won't. Similar for
```
X = *v;
Y = *v;
```
the compiler could reuse the value already read from `v` in the first line to effectively make the second like `Y = X;` but if `v` is a `volatile` pointer, it won't.
In this case that pointer points to memory, shared with the host, so the compiler has no way to know that the host can modify that memory. But to the compiler `ptr` and `ptr_c` are two different pointers, pointing to two different addresses - with some fixed offset. The compiler has no idea that it's "the same" memory. So whatever we do with one of the two pointers will not tell anything to the compiler that something also affects the other pointer. And since `ptr_c` is never used to access memory, since it is never dereferenced, the `volatile` qualifier for it has no effect.

softwarecki (approved):

src/library_manager/lib_manager.c
jxstelter: Looks good.

jxstelter (approved):

PR #6794, opened 2022-12-13
ipc4: fix crash when using ALH DAI

serhiy-katsyuba-intel:
dma_buffer is not yet created when ipc_dai_data_config() is called. Accessing dma_buffer->stream leads to crash.

Anyway, there is no need to set dd->dma_buffer->stream.frame_fmt = SOF_IPC_FRAME_S32_LE as this is done just after ipc_dai_data_config() in dai_params() by calling buffer_set_params().

Somewhat similar fix for IPC3 was done some time ago: 0ba90d8e2a.

### DIFF
--- a/src/ipc/ipc4/dai.c
+++ b/src/ipc/ipc4/dai.c
@@ -114,13 +114,12 @@ int ipc_dai_data_config(struct comp_dev *dev)
 		 * all formats, such as 8/16/24/32 bits.
 		 */
 		dev->ipc_config.frame_fmt = SOF_IPC_FRAME_S32_LE;
-		dd->dma_buffer->stream.frame_fmt = dev->ipc_config.frame_fmt;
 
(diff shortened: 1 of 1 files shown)


serhiy-katsyuba-intel:
Interestingly, the crash happens on dereferencing dma_buffer in comp_dbg() when building with CONFIG_SOC_LOG_LEVEL_DBG=y. It's strange that crash does not happen few lines above on dd->dma_buffer->stream.frame_fmt = dev->ipc_config.frame_fmt; assignment. This might be explained as compiler optimizes out that assignment, however, I do not understand how such optimization possible.

abonislawski (approved):

ranj063 (approved):

RanderWang (approved):

kv2019i (approved):
Seems good!

marcinszkudlinski (approved):

PR #6795, opened 2022-12-13
AEC fixes spotted with GCC that were missed with XCC

cujomalainey:


src/audio/google/google_rtc_audio_processing.c
```diff
@@ -377,7 +377,7 @@ static struct comp_dev *google_rtc_audio_processing_create(
 						 &capture_headroom_linear,
 						 &echo_path_delay_ms);
 
-	if (!cd->state) {
+	if (!cd->state || ret < 0) {
```
perahgren, line 380: This should be ret != 0 I think

perahgren (commented):
Thanks for fixing this!

src/audio/google/google_rtc_audio_processing.c
perahgren: Oups, sorry, realized that it should be 
if (ret < 0) {
Since we've already checked the state on line 369

perahgren (approved):

ranj063 (approved):

lgirdwood (approved):

PR #6797, opened 2022-12-14
module_adapter: free memory resource

barry-waves:
Since module-specific prepare API will be called every time on prepare
 of module_adapter (stream start trigger), the reset API should
free all memory that was allocated during the prepare API.

Signed-off-by: barry.jan <barry.jan@waves.com>


lgirdwood (approved):
@ranj063 @dbaluta pls review.

ranj063 (approved):

PR #6799, opened 2022-12-14
Topology2: add smart amp nocodec test topology

aiChaoSONG:
This patch adds the smart amp nocodec topology, this topology should be used with sof main and   https://github.com/thesofproject/linux/pull/4121

sof-mtl-nocodec


aiChaoSONG:
@marc-hb This topology is just like the nocodec topology, and we do have the smart_amp module in the firmware. I think it is good. I am not sure if we want to move all nocodec to a new folder in the future, if yes, this one can be move with all nocodec topologies

plbossart (changes requested):

plbossart:
> we do have the smart_amp module in the firmware

That is in addition a fundamentally wrong assumption. We should not have test modules in the base firmware. It's ok as a temporary step while the firmware team works on generating separate libraries, but it cannot be the long term direction.

tools/topology/topology2/avs-tplg/CMakeLists.txt
```diff
@@ -40,6 +40,10 @@ DMIC1_ID=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1"
 PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-tgl-nocodec.bin,DEEPBUFFER_FW_DMA_MS=100"
 "cavs-nocodec\;sof-adl-nocodec\;NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,\
 PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-adl-nocodec.bin,DEEPBUFFER_FW_DMA_MS=100"
+
+# smart amp nocodec test topology
+"cavs-smart-amp-nocodec\;cavs-tgl-smart-amp-nocodec\;PREPROCESS_PLUGINS=nhlt,\
+NHLT_BIN=nhlt-cavs-tgl-smart-amp-nocodec.bin"
```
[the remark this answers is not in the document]
aiChaoSONG: @plbossart @marc-hb Now I add a new sub-folder for cavs test topologies, and move this topology into it.

> We should not have test modules in the base firmware. It's ok as a temporary step while the firmware team works on generating separate libraries, but it cannot be the long term direction.

agreed and sounds good to me, I think finally, we will make the smart amp test module a loadable one.

tools/topology/topology2/include/components/smart_amp.conf
```diff
@@ -0,0 +1,107 @@
...
+		#
+		unique	"instance"
+	}
+
+	# Default attribute values for smart_amp widget
+
+	# uuid: 167a961e-8ae4-11ea-89f1-000c29ce1635
+	uuid			"1e:96:7a:16:e4:8a:ea:11:89:f1:00:0c:29:ce:16:35"
+	type			"effect"
+	no_pm 			"true"
```
plbossart, line 101: what is that?

tools/topology/topology2/include/components/smart_amp.conf
```diff
@@ -0,0 +1,107 @@
...
+	core_id			0
+	cpc 			5000
+	is_pages		1
+	num_sink_pins	2
```
plbossart, line 105: don't we need to specific which pin is used for what? I vaguely recall there was a special mapping needed

tools/topology/topology2/include/pipelines/cavs/smart-amp-playback.conf
```diff
@@ -0,0 +1,111 @@
+#
+# cAVS smart_amp playback pipeline
```
plbossart, line 2: why the reference to cAVS?

tools/topology/topology2/include/pipelines/cavs/smart-amp-playback.conf
```diff
@@ -0,0 +1,111 @@
...
+				dma_buffer_size "$[$ibs * 2]"
+			}
+		}
+
+		pipeline."1" {
+			priority	0
+			lp_mode		0
+		}
+
+		smart_amp."1" {
```
plbossart, line 82: no this is wrong. We should have a different pipeline for host and dai, and the smart-amp should be part of the dai pipeline. Keep in mind we do want to have multiple PCMs for playback, all mixed with the result processed by the smart amp.

tools/topology/topology2/cavs-smart-amp-nocodec.conf
```diff
@@ -0,0 +1,235 @@
+<searchdir:include>
```
plbossart, line 1: why is this file not in a test directory? It's at the top level of topology2, this is wrong.

tools/topology/topology2/cavs-smart-amp-nocodec.conf
```diff
@@ -0,0 +1,235 @@
...
+
+						0x04, 0x00, 0x00, 0x00, 0x00, 0x01, 0xff, 0xff,
+						0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x00, 0x01,
+						0xff, 0xff, 0xff, 0xff"
+				}
+			}
+		}
+	}
+
+	passthrough-be.1 {
```
plbossart, line 128: I don't want to see ANY references to passthrough moving forward. I have repeated over and over that we do not do passthrough pipelines any longer. We want mixers on playback and demux on capture. Get on with the program please.

plbossart (changes requested):
this needs a lot of work, not ready at all for merge.

tools/topology/topology2/include/components/smart_amp.conf
marc-hb: I don't know what is either but I know this would be much better to avoid double and triple negations:
```suggestion
	pm 			"false"
```
(I hope the default value can be either "true" or "false")

tools/topology/topology2/avs-tplg/CMakeLists.txt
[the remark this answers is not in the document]
marc-hb: > Now I add a new sub-folder for cavs test topologies, and move this topology into it

Thanks! Validation can copy / "merge" this test folder into the production folder. I hope we can keep the filenames different and collision-free.

No moving of files one by one please.

tools/topology/topology2/include/components/smart_amp.conf
ranj063: sorry this cannot be changed. no_pm is an existing terminology used by ALSA topology and must be set to true for all widgets

tools/topology/topology2/avs-tplg/CMakeLists.txt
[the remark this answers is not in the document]
plbossart: the topology was not moved in the update, it's in tools/topology/topology2/cavs-smart-amp-nocodec.conf
Difficult to be more confusing.

tools/topology/topology2/include/common/tokens.conf
```diff
@@ -25,6 +25,10 @@ Object.Base.VendorToken {
 		# output audio format in the init instance ipc4 message
 		# payload.
 		payload_with_output_fmt	415
+		# Token used to tell which init blob style is used for the widget:
+		#	- base module config + module specifc extension
+		#	- base module config + base config extension
+		init_blob_style		416
```
ranj063, line 31: separate patch please

tools/topology/topology2/cavs-smart-amp-nocodec.conf
```diff
@@ -0,0 +1,235 @@
...
+						0x00, 0x00, 0x00, 0x00, 0x04, 0x20, 0x00, 0x00,
+
+						0x00, 0x00, 0x00, 0x00, 0x00, 0x03, 0x00, 0x00,
+						0x80, 0xbb, 0x00, 0x00, 0x20, 0x00, 0x00, 0x00,
+						0x10, 0xff, 0xff, 0xff, 0x01, 0x00, 0x00, 0x00,
+						0x00, 0x00, 0x00, 0x00, 0x04, 0x20, 0x00, 0x00,
+
+						0x04, 0x00, 0x00, 0x00, 0x00, 0x01, 0xff, 0xff,
+						0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x00, 0x01,
+						0xff, 0xff, 0xff, 0xff"
```
ranj063, line 122: can this be moved out to a separate conf file and included here? It is polluting the top-level tplg file

ranj063: in fact why can't all of this go into smart_amp.conf and we should only set the name here in the top-level

tools/topology/topology2/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ add_custom_target(abi_target
 	DEPENDS abi.conf
 )
 
-add_dependencies(topologies2 topology2_cavs topology2_ace)
+add_dependencies(topologies2 topology2_cavs topology2_cavs_test topology2_ace)
```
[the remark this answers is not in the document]
marc-hb: This was discussed above @ranj063. Because we don't want to release it in sof-bin. It can be tested regularly either way.

aiChaoSONG:
replacement PR #6996, close

PR #6802, opened 2022-12-14
sof debug and trace: change sof log level for some modules

btian1:
Change some log level from info to dbg, those changed place are not reporting errors, just information, so no risk and can be change back when debugging.

From performance perspective, before and after:
Before: [11.86] ll_schedule: ll timer avg 2329, max 2887, overruns 0
After:   [11.52] ll_schedule: ll timer avg 2226, max 2789, overruns 0

Signed-off-by: Baofeng Tian <baofeng.tian@intel.com>

### DIFF
--- a/src/audio/dai-zephyr.c
+++ b/src/audio/dai-zephyr.c
@@ -485,7 +485,7 @@ static int dai_playback_params(struct comp_dev *dev, uint32_t period_bytes,
 		fifo = dai_get_fifo(dd->dai, dev->direction,
 				    dd->stream_id);
 
-		comp_info(dev, "dai_playback_params() fifo 0x%x", fifo);
+		comp_dbg(dev, "dai_playback_params() fifo 0x%x", fifo);
 
 		err = dma_sg_alloc(&config->elem_array, SOF_MEM_ZONE_RUNTIME,
 				   config->direction,
@@ -614,7 +614,7 @@ static int dai_capture_params(struct comp_dev *dev, uint32_t period_bytes,
 		fifo = dai_get_fifo(dd->dai, dev->direction,
 				    dd->stream_id);
 
-		comp_info(dev, "dai_capture_params() fifo 0x%x", fifo);
+		comp_dbg(dev, "dai_capture_params() fifo 0x%x", fifo);
 
 		err = dma_sg_alloc(&config->elem_array, SOF_MEM_ZONE_RUNTIME,
 				   config->direction,
@@ -844,7 +844,7 @@ static int dai_config_prepare(struct comp_dev *dev)
 	}
 
 	channel = dai_config_dma_channel(dev, dd->dai_spec_config);
-	comp_info(dev, "dai_config_prepare(), channel = %d", channel);
+	comp_dbg(dev, "dai_config_prepare(), channel = %d", channel);
 
 	/* do nothing for asking for channel free, for compatibility. */
 	if (channel == DMA_CHAN_INVALID) {
@@ -863,8 +863,8 @@ static int dai_config_prepare(struct comp_dev *dev)
 	dd->chan = &dd->dma->chan[channel];
 	dd->chan->dev_data = dd;
 
-	comp_info(dev, "dai_config_prepare(): new configured dma channel index %d",
-		  dd->chan->index);
+	comp_dbg(dev, "dai_config_prepare(): new configured dma channel index %d",
+		 dd->chan->index);
 
 	/* setup callback */
 	notifier_register(dev, dd->chan, NOTIFIER_ID_DMA_COPY,
@@ -879,7 +879,7 @@ static int dai_prepare(struct comp_dev *dev)
 	struct comp_buffer __sparse_cache *buffer_c;
 	int ret;
 
-	comp_info(dev, "dai_prepare()");
+	comp_dbg(dev, "dai_prepare()");
 
 	ret = dai_config_prepare(dev);
 	if (ret < 0)
@@ -930,7 +930,7 @@ static int dai_reset(struct comp_dev *dev)
 	struct dai_data *dd = comp_get_drvdata(dev);
 	struct dma_sg_config *config = &dd->config;
 
-	comp_info(dev, "dai_reset()");
+	comp_dbg(dev, "dai_reset()");
 
 	/*
 	 * DMA channel release should be skipped now for DAI's that support the two-step stop
--- a/src/audio/pipeline/pipeline-graph.c
+++ b/src/audio/pipeline/pipeline-graph.c
@@ -218,9 +218,9 @@ void pipeline_disconnect(struct comp_dev *comp, struct comp_buffer *buffer, int 
 	uint32_t flags;
 
 	if (dir == PPL_CONN_DIR_COMP_TO_BUFFER)
-		comp_info(comp, "disconnect buffer %d as sink", buffer->id);
+		comp_dbg(comp, "disconnect buffer %d as sink", buffer->id);
 	else
-		comp_info(comp, "disconnect buffer %d as source", buffer->id);
+		comp_dbg(comp, "disconnect buffer %d as source", buffer->id);
 
 	irq_local_disable(flags);
 	buf_list = buffer_comp_list(buffer, dir);
@@ -247,7 +247,7 @@ void pipeline_disconnect(struct comp_dev *comp, struct comp_buffer *buffer, int 
 /* pipelines must be inactive */
 int pipeline_free(struct pipeline *p)
 {
-	pipe_info(p, "pipeline_free()");
+	pipe_dbg(p, "pipeline_free()");
 
 	/*
 	 * pipeline_free should always be called only after all the widgets in the pipeline have
@@ -311,7 +311,7 @@ int pipeline_complete(struct pipeline *p, struct comp_dev *source,
 #endif
 	int ret;
 
-	pipe_info(p, "pipeline complete, clock freq %dHz", freq);
+	pipe_dbg(p, "pipeline complete, clock freq %dHz", freq);
 
 	/* check whether pipeline is already completed */
 	if (p->status != COMP_STATE_INIT) {
@@ -398,7 +398,7 @@ int pipeline_reset(struct pipeline *p, struct comp_dev *host)
 	};
 	int ret;
 
-	pipe_info(p, "pipe reset");
+	pipe_dbg(p, "pipe reset");
 
 	ret = walk_ctx.comp_func(host, NULL, &walk_ctx, host->direction);
 	if (ret < 0) {
--- a/src/audio/selector/selector.c
+++ b/src/audio/selector/selector.c
@@ -176,7 +176,7 @@ static struct comp_dev *selector_new(const struct comp_driver *drv,
 	struct comp_data *cd;
 	int ret;
 
-	comp_cl_info(&comp_selector, "selector_new()");
+	comp_cl_dbg(&comp_selector, "selector_new()");
 
 	dev = comp_alloc(drv, sizeof(*dev));
 	if (!dev)
@@ -206,7 +206,7 @@ static void selector_free(struct comp_dev *dev)
 {
 	struct comp_data *cd = comp_get_drvdata(dev);
 
-	comp_info(dev, "selector_free()");
+	comp_dbg(dev, "selector_free()");
 
 	rfree(cd);
 	rfree(dev);
@@ -224,7 +224,7 @@ static int selector_params(struct comp_dev *dev,
 {
 	int err;
 
-	comp_info(dev, "selector_params()");
+	comp_dbg(dev, "selector_params()");
 
 	err = selector_verify_params(dev, params);
 	if (err < 0) {
@@ -250,7 +250,7 @@ static int selector_ctrl_set_data(struct comp_dev *dev,
 
 	switch (cdata->cmd) {
 	case SOF_CTRL_CMD_BINARY:
-		comp_info(dev, "selector_ctrl_set_data(), SOF_CTRL_CMD_BINARY");
+		comp_dbg(dev, "selector_ctrl_set_data(), SOF_CTRL_CMD_BINARY");
 
 		cfg = (struct sof_sel_config *)
 		      ASSUME_ALIGNED(&cdata->data->data, 4);
@@ -286,7 +286,7 @@ static int selector_ctrl_get_data(struct comp_dev *dev,
 
 	switch (cdata->cmd) {
 	case SOF_CTRL_CMD_BINARY:
-		comp_info(dev, "selector_ctrl_get_data(), SOF_CTRL_CMD_BINARY");
+		comp_dbg(dev, "selector_ctrl_get_data(), SOF_CTRL_CMD_BINARY");
 
 		/* Copy back to user space */
 		ret = memcpy_s(cdata->data->data, ((struct sof_abi_hdr *)
@@ -321,7 +321,7 @@ static int selector_cmd(struct comp_dev *dev, int cmd, void *data,
 	struct sof_ipc_ctrl_data *cdata = ASSUME_ALIGNED(data, 4);
 	int ret = 0;
 
-	comp_info(dev, "selector_cmd()");
+	comp_dbg(dev, "selector_cmd()");
 
 	switch (cmd) {
 	case COMP_CMD_SET_DATA:
@@ -331,10 +331,10 @@ static int selector_cmd(struct comp_dev *dev, int cmd, void *data,
 		ret = selector_ctrl_get_data(dev, cdata, max_data_size);
 		break;
 	case COMP_CMD_SET_VALUE:
-		comp_info(dev, "selector_cmd(), COMP_CMD_SET_VALUE");
+		comp_dbg(dev, "selector_cmd(), COMP_CMD_SET_VALUE");
 		break;
 	case COMP_CMD_GET_VALUE:
-		comp_info(dev, "selector_cmd(), COMP_CMD_GET_VALUE");
+		comp_dbg(dev, "selector_cmd(), COMP_CMD_GET_VALUE");
 		break;
 	default:
 		comp_err(dev, "selector_cmd(): invalid command");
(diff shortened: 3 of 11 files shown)


src/audio/dai-zephyr.c
```diff
@@ -477,15 +477,15 @@ static int dai_playback_params(struct comp_dev *dev, uint32_t period_bytes,
 	config->is_scheduling_source = comp_is_scheduling_source(dev);
 	config->period = dev->pipeline->period;
 
-	comp_info(dev, "dai_playback_params() dest_dev = %d stream_id = %d src_width = %d dest_width = %d",
-		  config->dest_dev, dd->stream_id,
-		  config->src_width, config->dest_width);
+	comp_dbg(dev, "dai_playback_params() dest_dev = %d stream_id = %d src_width = %d dest_width = %d",
```
lgirdwood, line 480: lets keep this one as its low frequency and useful info.

lgirdwood (commented):
@abonislawski @juimonen @softwarecki pls take a look, there is some low hanging fruit here that give us a few % performance improvement.

src/audio/dai-zephyr.c
```diff
@@ -606,15 +606,15 @@ static int dai_capture_params(struct comp_dev *dev, uint32_t period_bytes,
 		config->dest_width = config->src_width;
 	}
 
-	comp_info(dev, "dai_capture_params() src_dev = %d stream_id = %d src_width = %d dest_width = %d",
-		  config->src_dev, dd->stream_id,
-		  config->src_width, config->dest_width);
```
kv2019i, line 611: I'd keep this. Not having this can slow down CI results analysis.

src/audio/pipeline/pipeline-graph.c
```diff
@@ -115,8 +115,8 @@ struct pipeline *pipeline_new(uint32_t pipeline_id, uint32_t priority, uint32_t
 	struct pipeline *p;
 	int ret;
 
-	pipe_cl_info("pipeline new pipe_id %d priority %d",
-		     pipeline_id, priority);
```
kv2019i, line 119: I'd keep this as well for CI.

src/audio/pipeline/pipeline-stream.c
```diff
@@ -272,7 +272,7 @@ int pipeline_trigger(struct pipeline *p, struct comp_dev *host, int cmd)
 {
 	int ret;
 
-	pipe_info(p, "pipe trigger cmd %d", cmd);
```
kv2019i, line 275: I'd keep this, we use this a lot of analyse CI results.

src/ipc/ipc-helper.c
```diff
@@ -40,8 +40,8 @@ struct comp_buffer *buffer_new(const struct sof_ipc_buffer *desc)
 {
 	struct comp_buffer *buffer;
 
-	tr_info(&buffer_tr, "buffer new size 0x%x id %d.%d flags 0x%x",
-		desc->size, desc->comp.pipeline_id, desc->comp.id, desc->flags);
```
kv2019i, line 44: Keep this as well.

src/schedule/zephyr_domain.c
```diff
@@ -202,8 +202,8 @@ static int zephyr_domain_unregister(struct ll_schedule_domain *domain,
 
 	k_spin_unlock(&domain->lock, key);
 
-	tr_info(&ll_tr, "zephyr_domain_unregister domain->type %d domain->clk %d",
-		domain->type, domain->clk);
```
kv2019i, line 206: I'd keep this. this is unfrequent and useful for CI analysis.

src/schedule/zephyr_ll.c
```diff
@@ -264,8 +264,8 @@ static int zephyr_ll_task_schedule_common(struct zephyr_ll *sch, struct task *ta
 
 	zephyr_ll_assert_core(sch);
 
-	tr_info(&ll_tr, "task add %p %pU priority %d flags 0x%x", task, task->uid,
-		task->priority, task->flags);
```
kv2019i, line 268: I'd keep this. this is unfrequent and useful for CI analysis.

kv2019i (changes requested):
Thanks @btian1 ! There's clearly a lot of unnecessary traffic at info level. I would like to keep some of the logs at info level. Basicly we do need to have essential info output to mtrace, so we can analyze CI results more efficiently, and if we drop these to dbg levels, basicly one has to reproduce locally with a custom build, and this is not effective.

kv2019i:
@btian1 wrote:
> From performance perspective, before and after: Before:[3.3] ll_schedule: ll timer avg 2577, max 5011, overruns 0 after: [7.4] ll_schedule: ll timer avg 2524, max 4996, overruns 0

Just noting that I think it's ok to have "max" spikes as long as we 
 a) understand where they come from,
 b) they are within our scheduling buffer (e.g. not causing audible glitches to actual workload)

In other words, I think it's ok to take some hit in the max values to emit logs at key places in pipeline setup/teardown, as long as the "avg" stays impacts. If we have high-frequency logs in some DSP processing loop, that will show up in "avg" and that is a complete no-go in default build.

src/ipc/ipc4/dai.c
```diff
@@ -298,8 +298,8 @@ int dai_config(struct comp_dev *dev, struct ipc_config_dai *common_config,
 	    dd->ipc_config.type != common_config->type)
 		return 0;
 
-	comp_info(dev, "dai_config() dai type = %d index = %d dd %p",
-		  common_config->type, common_config->dai_index, dd);
+	comp_dbg(dev, "dai_config() dai type = %d index = %d dd %p",
```
abonislawski, line 301: please keep this one

src/lib/dma.c
```diff
@@ -116,9 +116,9 @@ struct dma *dma_get(uint32_t dir, uint32_t cap, uint32_t dev, uint32_t flags)
 
 	dmin->sref++;
 
-	tr_info(&dma_tr, "dma_get() ID %d sref = %d busy channels %ld",
-		dmin->plat_data.id, dmin->sref,
-		atomic_read(&dmin->num_channels_busy));
+	tr_dbg(&dma_tr, "dma_get() ID %d sref = %d busy channels %ld",
```
abonislawski, line 119: and this

src/lib/dma.c
```diff
@@ -254,9 +254,9 @@ struct dma *dma_get(uint32_t dir, uint32_t cap, uint32_t dev, uint32_t flags)
 	if (!ret)
 		dmin->sref++;
 
-	tr_info(&dma_tr, "dma_get() ID %d sref = %d busy channels %ld",
-		dmin->plat_data.id, dmin->sref,
-		atomic_read(&dmin->num_channels_busy));
+	tr_dbg(&dma_tr, "dma_get() ID %d sref = %d busy channels %ld",
```
abonislawski, line 257: and this

abonislawski (changes requested):
Agree with @kv2019i, please restore some useful logs.
Also please restore dma/dai usage logs:

src/ipc/ipc4/dai.c
btian1: > 

Yes, I just put it here to let people know there is max performance data for first second, after that, max will return to normal, like below:

[   16.460291] <inf> ll_schedule: ll timer avg 2576, max 3136, overruns 0
[    9.485483] <inf> ll_schedule: ll timer avg 2523, max 3084, overruns 0

btian1:
Thank you for your review, I will change back based on your comments.

src/ipc/ipc-helper.c
btian1: I reverted to info, this is pipeline allocate buffer, always see it, personal think, it is not so important.

src/ipc/ipc4/dai.c
abonislawski: @btian1 so this specific log can affect performance?

src/audio/component.c
```diff
@@ -62,8 +62,8 @@ int comp_set_state(struct comp_dev *dev, int cmd)
 	int requested_state = comp_get_requested_state(cmd);
 
 	if (dev->state == requested_state) {
-		comp_info(dev, "comp_set_state(), state already set to %u",
-			  dev->state);
+		comp_dbg(dev, "comp_set_state(), state already set to %u",
+			 dev->state);
```
lyakh, line 66: not sure about this one. Indeed in some cases it isn't an error, but it has also proven useful sometimes when debugging. Maybe we should check all cases where this gets printed now and for all those cases check in the caller? Or add a wrapper `comp_set_state_if_different()` or something and then maybe even change this to `comp_warn()`

src/audio/dai-zephyr.c
```diff
@@ -449,7 +449,7 @@ static int dai_playback_params(struct comp_dev *dev, uint32_t period_bytes,
 	struct dma_block_config *dma_block_cfg;
 	struct dma_block_config *prev = NULL;
 	struct comp_buffer __sparse_cache *dma_buf = buffer_acquire(dd->dma_buffer),
-		*local_buf = buffer_acquire(dd->local_buffer);
+	*local_buf = buffer_acquire(dd->local_buffer);
```
lyakh, line 452: this is both wrong and unrelated, please drop

src/audio/pipeline/pipeline-graph.c
```diff
@@ -189,9 +189,9 @@ int pipeline_connect(struct comp_dev *comp, struct comp_buffer *buffer,
 	uint32_t flags;
 
 	if (dir == PPL_CONN_DIR_COMP_TO_BUFFER)
-		comp_info(comp, "connect buffer %d as sink", buffer->id);
+		comp_dbg(comp, "connect buffer %d as sink", buffer->id);
 	else
-		comp_info(comp, "connect buffer %d as source", buffer->id);
+		comp_dbg(comp, "connect buffer %d as source", buffer->id);
```
lyakh, line 194: these are actually useful

src/audio/selector/selector.c
```diff
@@ -882,10 +882,10 @@ static int selector_prepare(struct processing_module *mod)
...
 	 * reduce channel count between source and sink
 	 */
-	comp_info(dev, "selector_prepare(): sourceb->schannels = %u",
-		  source_c->stream.channels);
-	comp_info(dev, "selector_prepare(): sinkb->channels = %u",
-		  sink_c->stream.channels);
+	comp_dbg(dev, "selector_prepare(): sourceb->schannels = %u",
+		 source_c->stream.channels);
+	comp_dbg(dev, "selector_prepare(): sinkb->channels = %u",
+		 sink_c->stream.channels);
```
lyakh, line 888: I haven't worked with selector yet, but I would imagine that this could be useful too, you'd save some if you merge the two into one

src/ipc/ipc4/handler.c
```diff
@@ -1040,7 +1040,7 @@ void ipc_cmd(struct ipc_cmd_hdr *_hdr)
 		return;
 
 	if (cpu_is_primary(cpu_get_id()))
-		tr_info(&ipc_tr, "rx\t: %#x|%#x", in->primary.dat, in->extension.dat);
+		tr_dbg(&ipc_tr, "rx\t: %#x|%#x", in->primary.dat, in->extension.dat);
```
lyakh, line 1043: NAK, please, keep this

src/audio/component.c
```diff
@@ -62,8 +62,8 @@ int comp_set_state(struct comp_dev *dev, int cmd)
 	int requested_state = comp_get_requested_state(cmd);
```
lyakh, line 62: performance difference is very much test-dependent. If you do a lot of starting / stopping with very short streaming times, or if you do parallel asynchronous multi-PCM testing, your logging contribution will be rather high, however if you just start a single stream, let it run for a minute, then stop - the saving will be much lower, because obviously prints are done during set up and tear down and very few during streaming

lyakh (changes requested):

src/audio/component.c
btian1: That's correct, and I also agree that some logs are useful, just one thinking: how about use debug build to perform test(with dbg level log print), and keep release build as much less logs?  I guess there maybe something wrong for timing due to too much dbg level logs, it may cause timeout for some cases.

Bios does not have any logs for production build, and validation team use debug build perform daily test.

src/audio/component.c
btian1: ok, maybe warn is more suitable here, since it is not error, use info seems too light, afterall, it is not a right operation.

src/ipc/ipc4/handler.c
btian1: I know this is debug ipc key message, I think this part code already matured, and no need further debug, this is the reason for remove this print.

src/audio/component.c
lyakh: unfortunately we have too mane debug prints, so if you enable them all, debugging becomes way too difficult. I think it enables printing during streaming, so your log is flooded, messages get dropped, or the whole run slows down to timeouts, etc. Maybe we need verbose debug for debugging during streaming to distinguish it from "slightly more verbose than usual but still usable debug logging" but we don't have it so far

src/ipc/ipc4/handler.c
lyakh: no, it isn't because we are unsure whether this print is working, it is because this is the primary means of matching firmware and kernel logs, please, keep.

btian1:
> @btian1 what does the periodic stream runtime look like with the latest update applied. Can you paste - I want to make sure we are disabling the low value recurrent trace whilst we keep the low frequency or high value data..

@lgirdwood ， thanks for your sharp catch, previously, I tested on a adlp device, data is based on adlp + debug key.
Now, I returned this device, and use an TGL with product key device, I got much better data even reverted some log changes(required by reviewer):

 [11.86] <inf> ll_schedule: ll timer avg 2329, max 2887, overruns 0
 [11.52] <inf> ll_schedule: ll timer avg 2226, max 2789, overruns 0

It is around 5% percent reduction with this PR, I feel happy with this data, previous data only have near 3% reduction with debug key.

Any difference between debug key and community key regarding performance?

Thanks
Tim

src/audio/component.c
```diff
@@ -62,7 +62,7 @@ int comp_set_state(struct comp_dev *dev, int cmd)
...
 
 	if (dev->state == requested_state) {
-		comp_info(dev, "comp_set_state(), state already set to %u",
+		comp_warn(dev, "comp_set_state(), state already set to %u",
```
kv2019i, line 65: This is a bit unexpected and does not match the git commit message. I can't comment whether this should be a warning or not and we'd need to get ack from author of this component.c code.

src/audio/pipeline/pipeline-graph.c
```diff
@@ -218,9 +218,9 @@ void pipeline_disconnect(struct comp_dev *comp, struct comp_buffer *buffer, int
 	uint32_t flags;
 
 	if (dir == PPL_CONN_DIR_COMP_TO_BUFFER)
-		comp_info(comp, "disconnect buffer %d as sink", buffer->id);
+		comp_dbg(comp, "disconnect buffer %d as sink", buffer->id);
 	else
-		comp_info(comp, "disconnect buffer %d as source", buffer->id);
+		comp_dbg(comp, "disconnect buffer %d as source", buffer->id);
```
kv2019i, line 223: I was ok with this, but didn't someone ask to keep these?

src/lib/dma.c
kv2019i: @btian1  Please don't mark as resolved without addressing the comment.

src/audio/component.c
kv2019i: @btian1 Yup, @lyakh has a point. In practical terms, if you want to get these logs back, that is not easy as current "dbg" is so crowded. But I do think we have too much at "info" so this PR is needed. It's now a question of going through the log entries and evaluating per log, whether it should be kept or not.

@btian1 And one can of course ship with traces disabled, that's just a matter of a Kconfig option. Some SOF release binaries do not have any logs enabled. But e.g. the public upstream SOF releases for Intel hardware do have logs enabled as the performance impact has been within acceptable limits. But at code level, we should not assume anything about this, but rather enable both options to be used.

kv2019i (changes requested):
Starts to look good!  A few more comments inline...

src/audio/component.c
btian1: ok, let me change to info, from performance view, there is no difference for info and warn.

src/lib/dma.c
btian1: this should be missed during push, let me change it back.

src/audio/component.c
btian1: ok, let's work step by step to get logging better.

lyakh:
> @btian1 Yup, @lyakh has a point. In practical terms, if you want to get these logs back, that is not easy as current "dbg" is so crowded. But I do think we have too much at "info" so this PR is needed. It's now a question of going through the log entries and evaluating per log, whether it should be kept or not.

@kv2019i can we make an IPC for this? We would boot with logging enabled because otherwise you lose all the early logging irrecoverably. But then unless an IPC comes to keep logging, it's disabled except for warnings and errors?

lgirdwood:
> > @btian1 Yup, @lyakh has a point. In practical terms, if you want to get these logs back, that is not easy as current "dbg" is so crowded. But I do think we have too much at "info" so this PR is needed. It's now a question of going through the log entries and evaluating per log, whether it should be kept or not.
> 
> @kv2019i can we make an IPC for this? We would boot with logging enabled because otherwise you lose all the early logging irrecoverably. But then unless an IPC comes to keep logging, it's disabled except for warnings and errors?

I think Zephyr has more log levels which will help, but we need to map the Zephyr log levels to legacy at some point.
@btian1 I would update this PR to only change the agreed messages, and keep all the messages that are under discussion for future alignment.

src/lib/dma.c
btian1: > ed to map the Zephyr log levels to legacy at some point.
> @btian1 I would update this PR to o

Thanks, latest version already include agreed log, and for logs under discussion, I already reverted back to origin.
Hopefully, this PR can be merged before your holiday, @kv2019i @lgirdwood

kv2019i (approved):
Thanks @btian1 !

lyakh (approved):
looks ok now, thanks

abonislawski (approved):

wszypelt:
@kv2019i I turned on the tests again, we'll see what comes of it within 1 hour

kv2019i:
Thanks @wszypelt , now good! One known issue with TGL alsabat, otherwise good. Proceeding with merge.

PR #6803, opened 2022-12-14
buffer: NULL pointer protection

pjdobrowolski:
Static analysis discover scenario in which comp_update might point to NULL in certain situations. Additional condition preventing it is added.

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -677,6 +677,12 @@ int module_adapter_copy(struct comp_dev *dev)
 	}
 
 	if (mod->simple_copy) {
+		if (!source_c) {
+			ret = -ENODATA;
+			comp_err(dev, "comp_update_buffer_consume() %x: NULL dereference error",
+				 ret);
+			goto out;
+		}
 		comp_update_buffer_consume(source_c, mod->input_buffers[0].consumed);
 		buffer_release(sink_c);
 		buffer_release(source_c);


src/audio/module_adapter/module_adapter.c
```diff
@@ -677,6 +677,12 @@ int module_adapter_copy(struct comp_dev *dev)
 	}
 
 	if (mod->simple_copy) {
+		if (!source_c) {
+			ret = -ENODATA;
+			comp_err(dev, "comp_update_buffer_consume() %x: NULL dereference error",
+				 ret);
+			goto out;
+		}
```
ujfalusi, line 685: `source_c` can not be NULL, it is assigned at line 616 under the same if condition at line 612.

The static analyzer got this wrong by not understanding the context.

ujfalusi: And even if `source_c` is NULL (which it can not be), the `sink_c` might not be and still need to be released?

ujfalusi: Furthermore: at the `out:` label we have:
```c
if (mod->simple_copy) {
	buffer_release(sink_c);
	buffer_release(source_c);
}
```
where `source_c` is still NULL if we jump from here, is that fine?

pjdobrowolski: You're right. However in future there might be another check and it could pop ap again. My first approach was to add NULL prevent condition in comp_update_buffer_consume() because this function is causing that critical error. However exiting without dbg message was not good enough. What is your suggestion?

lgirdwood: Are we able to move on as a false positive today ?

pjdobrowolski: Ok, I'm closing it as false positive.

marc-hb:
> However in future there might be another check and it could pop ap again.

A decent static analyzer tool should be able to remember false positives.

> source_c can not be NULL, it is assigned at line 616 under the same if condition at line 612.

A decent, professionally supported static analyzer also has a channel to report false positives as analysis bugs.

Of course there will always be complex situations and false positives that cannot be fixed. When possible it does not hurt to simplify the code, this also helps humans review it and its correctness.

pjdobrowolski:
Yeah, right but for mtl 02 we don't have a ready recipes for false positives.

marc-hb:
Not sure what a "recipe" is. In the tool it should hopefully be as simple as one "false positive" click. The hard part is not in the tool: reviewing whether it's actually a false positive. If that was easy then it wouldn't be a false positive :-)

PR #6805, opened 2022-12-14
ipc4: copier: remove redundant assignment

serhiy-katsyuba-intel:
Removes variable cd->converter[i] assigning to self. This fixes error reported by static code analyzer.

### DIFF
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -610,15 +610,16 @@ static int copier_prepare(struct comp_dev *dev)
 	if (ret == COMP_STATUS_STATE_ALREADY_SET)
 		return PPL_STATUS_PATH_STOP;
 
-	if (cd->endpoint_num) {
+	for (i = 0; i < cd->endpoint_num; i++) {
+		ret = cd->endpoint[i]->drv->ops.prepare(cd->endpoint[i]);
+		if (ret < 0)
+			break;
+	}
 
-		for (i = 0; i < cd->endpoint_num; i++) {
-			ret = cd->endpoint[i]->drv->ops.prepare(cd->endpoint[i]);
-			if (ret < 0)
-				break;
-		}
-	} else {
-		/* set up format conversion function */
+	if (!cd->endpoint_num) {
+		/* set up format conversion function for pin 0, for other pins (if any)
+		 * format is set in IPC4_COPIER_MODULE_CFG_PARAM_SET_SINK_FORMAT handler
+		 */
 		cd->converter[0] = get_converter_func(&cd->config.base.audio_fmt,
 							      &cd->config.out_fmt, ipc4_gtw_none,
 							      ipc4_bidirection);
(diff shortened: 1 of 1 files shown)


lgirdwood (approved):
@serhiy-katsyuba-intel btw would be nice to have a macro for the 0 in ```endpoint[0]``` throught the copier code, I assume its the default endpoint ?

serhiy-katsyuba-intel:
> @serhiy-katsyuba-intel btw would be nice to have a macro for the 0 in `endpoint[0]` throught the copier code, I assume its the default endpoint ?

We have :) It's IPC4_COPIER_GATEWAY_PIN. But it is not consistently used in copier code, in some places that define is used in other places 0.

lgirdwood:
@serhiy-katsyuba-intel can you check CI. Thanks !

serhiy-katsyuba-intel:
> @serhiy-katsyuba-intel can you check CI. Thanks !

The CI failures are unrelated to the changes made, looks like CI problem.

ranj063 (approved):

src/audio/copier/copier.c
```diff
@@ -610,25 +610,23 @@ static int copier_prepare(struct comp_dev *dev)
 	if (ret == COMP_STATUS_STATE_ALREADY_SET)
 		return PPL_STATUS_PATH_STOP;
 
-	if (cd->endpoint_num) {
+	for (i = 0; i < cd->endpoint_num; i++) {
+		ret = cd->endpoint[i]->drv->ops.prepare(cd->endpoint[i]);
+		if (ret < 0)
+			break;
```
kv2019i, line 616: This has a slight behaviour change. If error happens, before we skipped running prepare on pre-patch L616, but now we still. Is this intentional? If not, check on "&& !ret" on L619 will do.

kv2019i (commented):
@serhiy-katsyuba-intel Otherwise good, but there's on change in behaviour. Is this intentional?

src/audio/copier/copier.c
lyakh: @kv2019i sorry, not sure why you think the behaviour is changed here - I don't see it, but I'd agree that it would be better to avoid unnecessary changes. Just remove lines 629-631, don't change anything else

kv2019i: @lyakh Ack, the scenario I have above cannot happen. If we execute prepare(), endpoint_num must be non-zero, so we never go to the branch starting at L619. So ok, this is good.

kv2019i (approved):

tmleman (approved):

marc-hb:
> > @serhiy-katsyuba-intel can you check CI. Thanks !

> The CI failures are unrelated to the changes made, looks like CI problem.

Which failures?

kv2019i:
@wszypelt Can you check the merge/build fail? It seems only log upload failed and rest of the test job was successful.

pblaszko (approved):

paulstelian97 (approved):
I do agree that there might be an issue with this patch (namely if you get an error in both the initial loop (new line 613) and in the get_converter_func function (call is on new line 623) the latter will hide the former. That said, it's still an error so it might be fine (the debugging shouldn't get _too_ much harder). I'll say this is fine.

lgirdwood:
@kv2019i for v2.4 ?

PR #6808, opened 2022-12-14
cherry-pick: Fix performance counter regression

johnylin76:
Cherry picked from commit 8e59c9617376597459000ea956490a5328b68033 in main.
Since rpl-001-drop-stable is upon tag:v2.2, it lacks 8e59c9617376597459000ea956490a5328b68033 which is tag:v2.3.
However, this commit fix should be merged to rpl-001-drop-stable otherwise the performance counter is not usable for xtos use cases.


cujomalainey (approved):

johnylin76:
@mwasko is this ok to merge now?

PR #6809, opened 2022-12-14
module_adapter: fix large blob processing

iganakov:
Use correct IPC4 mailbox size. Fix `fragment_size`
calculation for the first and middle config chunks.

Signed-off-by: Ievgen Ganakov <ievgen.ganakov@intel.com>

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -936,8 +936,11 @@ int module_set_large_config(struct comp_dev *dev, uint32_t param_id, bool first_
 		fragment_size = data_offset_size;
 		break;
 	case MODULE_CFG_FRAGMENT_MIDDLE:
+		fragment_size = MAILBOX_DSPBOX_SIZE;
+		break;
 	case MODULE_CFG_FRAGMENT_FIRST:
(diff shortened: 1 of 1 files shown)


ranj063 (approved):

jxstelter (approved):

src/audio/module_adapter/module_adapter.c
```diff
@@ -936,8 +936,11 @@ int module_set_large_config(struct comp_dev *dev, uint32_t param_id, bool first_
 		fragment_size = data_offset_size;
```
kv2019i, line 936: Minor: why the strange word-wrapping in commit message? 

```
Use correct IPC4 mailbox size. Correct calculation of fragment_size.
```

Nicely fits on one line (with the 74-column typical line-wrap).

Anyways, not blocking PR for this, but just noting this as this seems to repeat in many PRs.

kv2019i (approved):

kv2019i:
Sparse-zephyr fails are known fails in current main, so proceeding with merge.

src/audio/module_adapter/module_adapter.c
iganakov: @kv2019i  Good point. I'll bare this in mind in the future PRs

PR #6814, opened 2022-12-15
xtensa-build-zephyr.py: checksum deterministic configs.c

marc-hb:
xtensa-build-zephyr.py: checksum deterministic configs.c

.config is not deterministic because it has absolute paths in comments
and its order seems hard to predict. configs.c has the same information
generated in a determistic way.


lgirdwood (approved):

kv2019i (approved):

PR #6815, opened 2022-12-15
Tools: Topology1: Add to development topologies DRC for UP2 board

singalsu:
This patch adds to development topologies build of minimal topologies sof-apl-pcm512x-drc.tplg and sof-apl-pcm512x-mdrc.tplg for testing.


singalsu:
@ShriramShastry You can use these topologies to test DRC in up-squared board. Also here's the example of how to generate topology for any of the processing components.

ShriramShastry:
> @ShriramShastry You can use these topologies to test DRC in up-squared board. Also here's the example of how to generate topology for any of the processing components.

Thank you very much.
I use the topology and begin the DRC tests.

singalsu:
Now non-draft

PR #6819, opened 2022-12-16
xtensa-build-zephyr.py: don't leave the install of an old build behind

marc-hb:
In theory a failure from this script should immediately stop the show and no one should then use the staging directory.

In practice we cannot make sure, especially not with shell scripts. So make sure no previous build of a requested platform is left behind.

### DIFF
--- a/scripts/xtensa-build-zephyr.py
+++ b/scripts/xtensa-build-zephyr.py
@@ -477,6 +477,26 @@ def get_build_and_sof_version(abs_build_dir):
 
 	return sof_fw_version, sof_build_version
 
+def rmtree_if_exists(directory):
+	"This is different from ignore_errors=False because it deletes everything or nothing"
+	if os.path.exists(directory):
+		shutil.rmtree(directory)
+
+def clean_staging(platform):
+	print(f"Cleaning {platform} from {STAGING_DIR}")
+
+	rmtree_if_exists(STAGING_DIR / "sof-info" / platform)
+
+	sof_output_dir = STAGING_DIR / "sof"
+
+	# --use-platform-subdir
+	rmtree_if_exists(sof_output_dir / platform)
+
+	# Remaining .ri and .ldc files
+	for f in sof_output_dir.glob(f"**/sof-{platform}.*"):
+		os.remove(f)
+
+
 STAGING_DIR = None
 def build_platforms():
 	global west_top, SOF_TOP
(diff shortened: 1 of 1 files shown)


scripts/xtensa-build-zephyr.py
```diff
@@ -477,6 +477,26 @@ def get_build_and_sof_version(abs_build_dir):
...
+	rmtree_if_exists(STAGING_DIR / "sof-info" / platform)
+
+	sof_output_dir = STAGING_DIR / "sof"
+
+	# --use-platform-subdir
+	rmtree_if_exists(sof_output_dir / platform)
+
+	# Remaining .ri and .ldc files
+	for f in sof_output_dir.glob(f"**/sof-{platform}.*"):
+		os.remove(f)
```
kv2019i, line 497: Wouldn't this delete all platforms, even those not build in this run? Or that is the intention?

kv2019i (commented):
This seems pretty scary. I'm guessing you've double-checked and tripled checked this can in no circumstances go haywire.

scripts/xtensa-build-zephyr.py
marc-hb: It starts by cleaning only the platforms _requested_ in this run, see the `{platform}` argument.

marc-hb:
Unrelated crash in https://sof-ci.01.org/sofpr/PR6819/build2937/devicetest/index.html

marc-hb:
SOFCI TEST

aborisovich (commented):
I'll test this on Windows and let you know asap.

miRoox (approved):

aborisovich (approved):
Windows CI works, change works as intended.
Performed following test:
1. execute `python sof\scripts\xtensa-build-zephyr.py tgl -i IPC4`
2. execute `python sof\scripts\xtensa-build-zephyr.py mtl`
3. execute `python sof\scripts\xtensa-build-zephyr.py mtl`

Result:
After step 2 and step 3 TGL was build and output was not removed.
After step 3 MTL binaries from build-sof-staging have updated timestamps (removed binaries from step 2).
PR is good to go.

Still I'll rumble a bit again on our pythons, just FYI as a future refactor material:
1. Copying FW binaries and tools to `build-sof-staging` should be done as [CMake Install](https://cmake.org/cmake/help/latest/command/install.html) step.
2. Then, what @marc-hb done in this PR should be done using
[CMake option](https://cmake.org/cmake/help/v3.1/manual/cmake.1.html#options) --clean-first.
This flag was introduced quite recently but it does exactly what @marc-hb did here.
It removes files installed as part of install step, removes objects and binaries created by build, and most importantly, does not delete toolchain check discovery files so you basically can easily rebuild skipping this process:
image

marc-hb:
Thanks a lot for the review and testing.

> Copying FW binaries and tools to build-sof-staging should be done as [CMake Install](https://cmake.org/cmake/help/latest/command/install.html) step.

That seems more logical but:
- This /lib/firmware/ "installation" looks nothing like installing regular C/C++ software. I'm afraid this would require twisting CMake's arm a lot. CMake's "install" documentation is full of concepts totally irrelevant here like headers, static/shared libraries, executables,... There is a "custom installation logic" that can delegate to an external (Python?!) script but then everything is hidden from CMake and do features like `--clean-first` still work?
- It's hard enough to find people willing to make any change to build and test scripts; at least Python scripts are less scary and much more popular than CMake scripts. In fact CMake's new and main competitor "meson" adopted a Python-looking syntax.
- We will always need this script to at least iterate on multiple platforms. I hope it can become smaller (the XCC versions should especially be moved) but it will never be zero.

Issue #6709, opened 2022-12-02
[BUG] HDA audio output garbled on MTL & cAVS2.5 with native Zephyr DMA+DAI


kv2019i:
**Describe the bug**
Test audio playback to HDA headset codec. 

**To Reproduce**
aplay file.wav

**Reproduction Rate**
100%

**Expected behavior**
Clean audio output

**Impact**
Showstopper

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel:  c693d0caf6fa
    * SOF: f6f9b62d5a1
2) Name of the topology file
    * Topology: sof-hda-generic-2ch.tplg
3) Name of the platform(s) on which the bug is observed.
    * Platform: TGL

**Screenshots or console output**

**Notes**
Originally reported by @singalsu  on UP i11.

kv2019i:
FYI @mengdonglin @lgirdwood , this potentially affects sof-2.4 stable branch as well.

marc-hb:
@singalsu , @singalsu is this too small for https://github.com/thesofproject/sof-test/blob/main/test-case/check-alsabat.sh to notice?

kv2019i:
@marc-hb alsabat should definitely pick this up.

kv2019i:
Hmm, something is off here. @singalsu , @ujfalusi both got the audio problem. I just retried building kernel (sof-dev of today 41f9d4fe31be334ece6ec0a1829b730342f9c900), rebuild tplg (main of today 7029fb7b015912e9459f7146c2b23153281e98af) and rebuild FW (main of today 7029fb7b015912e9459f7146c2b23153281e98af) and I still get the garbled audio.

Dmic seems to be ok, so at least some of the DMAs are running as expected. The FW log also seems ok, performance is at expected level and no xruns are reported.

mengdonglin:
@kv2019i @singalsu @ujfalusi @keqiaozhang Do you have PulseAudio or Pipewire enabled during your test?

kv2019i:
@mengdonglin Not enabled, this happens with plain aplay to native PCM, so audio servers should not interfere.

Also tested today:
- forcing DMI-L1 off (-> no impact)
- older test branch for  native drivers: SOF c04b406fd0a1085ba5b1dee888c2bbcf0c601ba0 , Zephyr bd60b9289a635aedf80fc997e6b2a2de56fe5786 (-> issue still seen)

[...]

PR #6823, opened 2022-12-16
dai-zephyr: check max block count supported by DMA

abonislawski:
check max block count supported by DMA to prepare a correct config for DMA

Fixes: https://github.com/thesofproject/sof/issues/6709

OPEN: currently it is recalculating block count to max supported but it could also return error and recalculate to 1 only for HDA

### DIFF
--- a/src/audio/dai-zephyr.c
+++ b/src/audio/dai-zephyr.c
@@ -452,7 +452,7 @@ static int dai_playback_params(struct comp_dev *dev, uint32_t period_bytes,
 		*local_buf = buffer_acquire(dd->local_buffer);
 	uint32_t local_fmt = local_buf->stream.frame_fmt;
 	uint32_t dma_fmt = dma_buf->stream.frame_fmt;
-	uint32_t fifo;
+	uint32_t fifo, max_block_count, buf_size;
 	int i, err = 0;
 
 	buffer_release(local_buf);
@@ -486,6 +486,27 @@ static int dai_playback_params(struct comp_dev *dev, uint32_t period_bytes,
 				    dd->stream_id);
 
 		comp_info(dev, "dai_playback_params() fifo 0x%x", fifo);
+
+		err = dma_get_attribute(dd->dma->z_dev, DMA_ATTR_MAX_BLOCK_COUNT,
+					&max_block_count);
+		if (err < 0) {
+			comp_err(dev, "dai_playback_params(): could not get dma attr max block count, err = %d", err);
+			goto out;
+		}
+
+		if (max_block_count < period_count) {
+			comp_dbg(dev, "dai_playback_params(): block count = %d not supported by DMA", period_count);
+			buf_size = period_count * period_bytes;
+			do {
+				if (IS_ALIGNED(buf_size, max_block_count)) {
+					period_count = max_block_count;
+					period_bytes = buf_size / period_count;
+					break;
+				} else {
+					comp_warn(dev, "dai_playback_params(): alignment error for buffer = %d, block count = %d", buf_size, max_block_count);
+				}
+			} while (--max_block_count > 0);
+		}
 
 		err = dma_sg_alloc(&config->elem_array, SOF_MEM_ZONE_RUNTIME,
 				   config->direction,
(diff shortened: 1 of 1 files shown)


src/audio/dai-zephyr.c
```diff
@@ -452,7 +452,7 @@ static int dai_playback_params(struct comp_dev *dev, uint32_t period_bytes,
 		*local_buf = buffer_acquire(dd->local_buffer);
```
kv2019i, line 452: The commit message could be a bit more descriptive why this change is done.

src/audio/dai-zephyr.c
```diff
@@ -487,6 +487,20 @@ static int dai_playback_params(struct comp_dev *dev, uint32_t period_bytes,
...
+		if (err < 0) {
+			comp_err(dev, "dai_playback_params(): could not get dma attr max block count, err = %d", err);
+			goto out;
+		}
+
+		if (max_block_count < period_count) {
+			comp_info(dev, "dai_playback_params(): block count = %d not supported by DMA", period_count);
+			buf_size = period_count * period_bytes;
+			period_count = max_block_count;
+			period_bytes = buf_size / period_count;
```
kv2019i, line 501: Maybe add an assert for the case where this does not divide evenly? Should never happen of course.

src/audio/dai-zephyr.c
```diff
@@ -616,6 +630,20 @@ static int dai_capture_params(struct comp_dev *dev, uint32_t period_bytes,
...
+		if (err < 0) {
+			comp_err(dev, "dai_capture_params(): could not get dma attr max block count, err = %d", err);
+			goto out;
+		}
+
+		if (max_block_count < period_count) {
+			comp_info(dev, "dai_capture_params(): block count = %d not supported by DMA", period_count);
+			buf_size = period_count * period_bytes;
+			period_count = max_block_count;
+			period_bytes = buf_size / period_count;
```
kv2019i, line 644: Ditto for the assert.

kv2019i (commented):
Mostly good, a few minor comments inline. I think this is at least better than having HDA specific code in dai-zephyr.c.

src/audio/dai-zephyr.c
lyakh: hmm, maybe there's some implicit rule that I'm not aware of, but by just looking at this, it might well not divide evenly. We enter here when the caller requested more periods than the DMA can handle. We then calculate a buffer size from that larger block count and divide it by the the DMA maximum block count. So if max was 16 and the caller requested 17 and the period size isn't a multiple of 16, this won't divide well. Or is this meant for cases where the requested period count is a multiple of the block count, fitting multiple periods in a single DMA block?

abonislawski: The only one real use case is limiting period_count to 1 for HDA DAI so more like as you said - fitting multiple period in a single DMA block, I thought we can make this more generic.
Division problem we can quickly solve with assert or ALIGN_DOWN but Im now not sure if this will result in a valid config, lets say 4*128 is still valid if requested 8*64?

tmleman (approved):

marc-hb:
Re-running to clear https://sof-ci.01.org/sofpr/PR6823/build2944/boottest

https://sof-ci.01.org/sofpr/PR6823/build2944/devicetest/index.html is all green!

marc-hb:
SOFCI TEST

src/audio/dai-zephyr.c
abonislawski: Updated with alignment check and period count decreasing, please take a look if it's not over complicated

src/audio/dai-zephyr.c
kv2019i, line 452: @abonislawski Can you update the commit message a bit as well and maybe add tag for the bug, i.e. "Link: https://github.com/thesofproject/sof/issues/6709" ?

kv2019i (approved):
Thanks! It would be good to update git commit message though.

src/audio/dai-zephyr.c
abonislawski: Updated
PR linked to 6709

kv2019i:
Test results are good, including alsabat test now passing https://sof-ci.01.org/sofpr/PR6823/build3247/devicetest/index.html?model=TGLU_UP_HDA_IPC4ZPH&testcase=check-alsabat-headset-playback

There are multiple checkpatch warnings from long log/trace lines, but I think this is ok as it is not recommended to split string literals across multiple lines.

As this is a critical fix, proceeding with merge.

PR #6825, opened 2022-12-16
zephyr: ipc: refactor of pg prevent

tmleman:
Preventing d0i3 state during IPC processing in case if a host previously allowed for power state transitions. Transition can occur only in the Idle thread and this lock acts as a safeguard in case of the remaining threads to be non active (e.g. waiting for a semaphore).

The advantage of this solution is that we do not risk bypassing the prevent. PM state lock (get and put) require that their calls be balanced. One more put or get will result with hard to debug error. For example we can end up with permanent prevent on power gating.

Original code introduced here: ae4e037b0

### DIFF
--- a/src/ipc/ipc-zephyr.c
+++ b/src/ipc/ipc-zephyr.c
@@ -28,8 +28,6 @@
 #include <rtos/spinlock.h>
 #include <ipc/header.h>
 
-#include <zephyr/pm/policy.h>
-
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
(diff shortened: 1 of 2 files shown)


src/ipc/ipc-zephyr.c
```diff
@@ -141,7 +132,6 @@ enum task_state ipc_platform_do_cmd(struct ipc *ipc)
 	}
 #endif
 
-	pm_policy_state_lock_put(PM_STATE_RUNTIME_IDLE, PM_ALL_SUBSTATES);
 	return SOF_TASK_STATE_COMPLETED;
```
lgirdwood, line 135: My understanding is that this would be a balanced call with get()/put() ? Can you describe the case where this was breaking ?

tmleman: It wasn't breaking. But when revisiting this code I realized that it can be done more easily. This way we remove the risk that put/get calls will not be balanced and it we generate less instructions.

lgirdwood (commented):
@marcinszkudlinski @kv2019i @ranj063 pls review.

kv2019i (approved):
Looks good to me!

lgirdwood (approved):

PR #6830, opened 2022-12-16
ipc4: copier: Add support for ALH multi-gateways

serhiy-katsyuba-intel:
Adds ALH multi-gateway feature: functionality to mux/demux data from/into multiple gateways.

### DIFF
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -52,11 +52,30 @@ static pcm_converter_func get_converter_func(const struct ipc4_audio_format *in_
 					     enum ipc4_gateway_type type,
 					     enum ipc4_direction_type);
 
+static uint32_t bitmask_to_nibble_channel_map(uint8_t bitmask)
+{
+	int i;
+	int channel_count = 0;
+	uint32_t nibble_map = 0;
+
+	for (i = 0; i < 8; i++)
+		if (bitmask & BIT(i)) {
+			nibble_map |= i << (channel_count * 4);
+			channel_count++;
+		}
+
+	/* absent channel is represented as 0xf nibble */
+	nibble_map |= 0xFFFFFFFF << (channel_count * 4);
+
+	return nibble_map;
+}
+
 static int create_endpoint_buffer(struct comp_dev *parent_dev,
 				  struct copier_data *cd,
 				  struct comp_ipc_config *config,
 				  const struct ipc4_copier_module_cfg *copier_cfg,
 				  enum ipc4_gateway_type type,
+				  bool create_multi_endpoint_buffer,
 				  int index)
 {
 	enum sof_ipc_frame __sparse_cache in_frame_fmt, out_frame_fmt;
@@ -65,7 +84,7 @@ static int create_endpoint_buffer(struct comp_dev *parent_dev,
 	struct sof_ipc_buffer ipc_buf;
 	struct comp_buffer *buffer;
 	uint32_t buf_size;
-	uint32_t mask;
+	uint32_t chan_map;
 	int i;
 
 	audio_stream_fmt_conversion(copier_cfg->base.audio_fmt.depth,
@@ -108,7 +127,7 @@ static int create_endpoint_buffer(struct comp_dev *parent_dev,
 			buf_size = copier_cfg->base.obs * 2;
 		}
 
-		mask = copier_cfg->out_fmt.ch_map;
+		chan_map = copier_cfg->out_fmt.ch_map;
 	} else {
 		if (config->type == SOF_COMP_HOST) {
 			config->frame_fmt = out_frame_fmt;
@@ -120,7 +139,7 @@ static int create_endpoint_buffer(struct comp_dev *parent_dev,
 			buf_size = copier_cfg->base.ibs * 2;
 		}
 
-		mask = copier_cfg->base.audio_fmt.ch_map;
+		chan_map = copier_cfg->base.audio_fmt.ch_map;
 	}
 
 	parent_dev->ipc_config.frame_fmt = config->frame_fmt;
@@ -140,22 +159,40 @@ static int create_endpoint_buffer(struct comp_dev *parent_dev,
 	buffer->stream.valid_sample_fmt = valid_fmt;
 	buffer->buffer_fmt = copier_cfg->base.audio_fmt.interleaving_style;
 
-	if (type == ipc4_gtw_alh) {
-		struct sof_alh_configuration_blob *alh_blob;
-
+	/* For ALH multi-gateway case, configuration blob contains struct ipc4_alh_multi_gtw_cfg
+	 * with channel map and channels number for each individual gateway.
+	 */
+	if (type == ipc4_gtw_alh && is_multi_gateway(copier_cfg->gtw_cfg.node_id) &&
+	    !create_multi_endpoint_buffer) {
 		if (copier_cfg->gtw_cfg.config_length) {
-			alh_blob = (struct sof_alh_configuration_blob *)
-				copier_cfg->gtw_cfg.config_data;
-			mask = alh_blob->alh_cfg.mapping[index].channel_mask;
+			int channels;
+			const struct sof_alh_configuration_blob *alh_blob =
+				(const struct sof_alh_configuration_blob *)
+					copier_cfg->gtw_cfg.config_data;
+			uint8_t chan_bitmask = alh_blob->alh_cfg.mapping[index].channel_mask;
+
+			channels = popcount(chan_bitmask);
+			if (channels < 1 || channels > SOF_IPC_MAX_CHANNELS) {
+				comp_err(parent_dev, "Invalid channels mask: 0x%x", chan_bitmask);
+				return -EINVAL;
+			}
+			buffer->stream.channels = channels;
+			chan_map = bitmask_to_nibble_channel_map(chan_bitmask);
+		} else {
+			comp_err(parent_dev, "No ipc4_alh_multi_gtw_cfg found in blob!");
+			return -EINVAL;
 		}
 	}
 
 	for (i = 0; i < SOF_IPC_MAX_CHANNELS; i++)
-		buffer->chmap[i] = (mask >> i * 4) & 0xf;
+		buffer->chmap[i] = (chan_map >> i * 4) & 0xf;
 
 	buffer->hw_params_configured = true;
 
-	cd->endpoint_buffer[cd->endpoint_num] = buffer;
+	if (create_multi_endpoint_buffer)
+		cd->multi_endpoint_buffer = buffer;
+	else
+		cd->endpoint_buffer[cd->endpoint_num] = buffer;
 
 	return 0;
 }
@@ -182,7 +219,7 @@ static int create_host(struct comp_dev *parent_dev, struct copier_data *cd,
 
 	config->type = SOF_COMP_HOST;
 
-	ret = create_endpoint_buffer(parent_dev, cd, config, copier_cfg, ipc4_gtw_host, 0);
+	ret = create_endpoint_buffer(parent_dev, cd, config, copier_cfg, ipc4_gtw_host, false, 0);
 	if (ret < 0)
 		return ret;
 
@@ -242,7 +279,7 @@ static int init_dai(struct comp_dev *parent_dev,
 	int ret;
 
 	cd = comp_get_drvdata(parent_dev);
-	ret = create_endpoint_buffer(parent_dev, cd, config, copier, type, index);
+	ret = create_endpoint_buffer(parent_dev, cd, config, copier, type, false, index);
 	if (ret < 0)
 		return ret;
 
@@ -306,7 +343,6 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
 	struct sof_uuid id = {0xc2b00d27, 0xffbc, 0x4150, {0xa5, 0x1a, 0x24,
 				0x5c, 0x79, 0xc5, 0xe5, 0x4b}};
 	int dai_index[IPC4_ALH_MAX_NUMBER_OF_GTW];
-	struct sof_alh_configuration_blob *alh_blob;
 	union ipc4_connector_node_id node_id;
 	enum ipc4_gateway_type type;
 	const struct comp_driver *drv;
@@ -366,12 +402,24 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
 		 * }
 		 */
 		 /* get gtw node id in config data */
-		if (copier->gtw_cfg.config_length) {
-			alh_blob = (struct sof_alh_configuration_blob *)copier->gtw_cfg.config_data;
-			dai_count = alh_blob->alh_cfg.count;
-			for (i = 0; i < dai_count; i++)
-				dai_index[i] =
+		if (is_multi_gateway(node_id)) {
+			if (copier->gtw_cfg.config_length) {
+				const struct sof_alh_configuration_blob *alh_blob =
+					(const struct sof_alh_configuration_blob *)
+						copier->gtw_cfg.config_data;
+
+				dai_count = alh_blob->alh_cfg.count;
+				if (dai_count > IPC4_ALH_MAX_NUMBER_OF_GTW || dai_count < 0) {
+					comp_err(parent_dev, "Invalid dai_count: %d", dai_count);
+					return -EINVAL;
+				}
+				for (i = 0; i < dai_count; i++)
+					dai_index[i] =
 					IPC4_ALH_DAI_INDEX(alh_blob->alh_cfg.mapping[i].alh_id);
+			} else {
+				comp_err(parent_dev, "No ipc4_alh_multi_gtw_cfg found in blob!");
+				return -EINVAL;
+			}
 		} else {
 			dai_index[dai_count - 1] = IPC4_ALH_DAI_INDEX(node_id.f.v_index);
 		}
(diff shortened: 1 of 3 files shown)


src/audio/copier/copier.c
```diff
@@ -52,11 +52,30 @@ static pcm_converter_func get_converter_func(const struct ipc4_audio_format *in_
 					     enum ipc4_gateway_type type,
 					     enum ipc4_direction_type);
 
+static uint32_t bitmask_to_nibble_channel_map(uint8_t bitmask)
+{
+	int i, channel_count = 0;
+	uint32_t nibble_map = 0;
+
+	for (i = 0; i < 8; i++)
+		if (bitmask & (1 << i)) {
```
plbossart, line 61: BIT(i)

src/audio/copier/copier.c
```diff
@@ -52,11 +52,30 @@ static pcm_converter_func get_converter_func(const struct ipc4_audio_format *in_
...
+			nibble_map |= i << (channel_count * 4);
```
plbossart, line 62: how does this work for i == 0 ?

src/audio/copier/copier.c
```diff
@@ -140,22 +159,34 @@ static int create_endpoint_buffer(struct comp_dev *parent_dev,
...
-			alh_blob = (struct sof_alh_configuration_blob *)
-				copier_cfg->gtw_cfg.config_data;
-			mask = alh_blob->alh_cfg.mapping[index].channel_mask;
+			const struct sof_alh_configuration_blob *alh_blob =
+				(const struct sof_alh_configuration_blob *)
+					copier_cfg->gtw_cfg.config_data;
+			uint8_t chan_bitmask = alh_blob->alh_cfg.mapping[index].channel_mask;
+
+			chan_map = bitmask_to_nibble_channel_map(chan_bitmask);
+			buffer->stream.channels = popcount(chan_bitmask);
```
plbossart, line 174: shouldn't there be some sort of range check on the blob, specially that the map makes sense?

src/audio/copier/copier.c
```diff
@@ -780,6 +826,178 @@ static int copier_comp_trigger(struct comp_dev *dev, int cmd)
...
+	int32_t *r_ptr = audio_stream_wrap(src, (int32_t *)src->r_ptr + src_channel);
+	int32_t *w_ptr = audio_stream_wrap(dst, (int32_t *)dst->w_ptr + dst_channel);
+	int i;
+
+	for (i = 0; i < frame_count; i++) {
+		*w_ptr = *r_ptr;
+		r_ptr = audio_stream_wrap(src, r_ptr + src->channels);
+		w_ptr = audio_stream_wrap(dst, w_ptr + dst->channels);
+	}
+}
```
plbossart, line 865: don't we already have a library for such buffer copies?

src/audio/copier/copier.c
```diff
@@ -780,6 +826,178 @@ static int copier_comp_trigger(struct comp_dev *dev, int cmd)
...
+
+	comp_update_buffer_consume(multi_buf_c, avail_bytes);
+	buffer_release(multi_buf_c);
+}
+
+static void mux_into_multi_endpoint_buffer(struct copier_data *cd)
+{
+	struct comp_buffer __sparse_cache *multi_buf_c, *endp_buf_c;
+	int endp_idx, endp_channel;
+	uint32_t avail_frames = UINT32_MAX, container_size, bytes_produced;
```
plbossart, line 922: don't put on the same lines variables that you initialize and other that you don't initialize. This is painful to read and review.
I personally err on the side of one variable per line. Some coding standards also require this.

src/audio/copier/copier.c
```diff
@@ -1007,7 +1219,24 @@ static int copier_params(struct comp_dev *dev, struct sof_ipc_stream_params *par
 	if (cd->endpoint_num) {
 		for (i = 0; i < cd->endpoint_num; i++) {
 			update_internal_comp(dev, cd->endpoint[i]);
-			ret = cd->endpoint[i]->drv->ops.params(cd->endpoint[i], params);
+
+			/* For ALH multi-gateway case, params->channels is a total multiplexed
+			 * number of channels. Demultiplexed number of channels for each individual
+			 * gateway comes in blob's struct ipc4_alh_multi_gtw_cfg.
```
plbossart, line 1225: in 100% of the cases today, we end-up sending the SAME information across multiple links.

All these channel changes are really useless in practice, at least for playback.

src/audio/copier/copier.c
```diff
@@ -367,12 +397,24 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
 		 * }
 		 */
 		 /* get gtw node id in config data */
-		if (copier->gtw_cfg.config_length) {
-			alh_blob = (struct sof_alh_configuration_blob *)copier->gtw_cfg.config_data;
-			dai_count = alh_blob->alh_cfg.count;
-			for (i = 0; i < dai_count; i++)
-				dai_index[i] =
+		if (is_multi_gateway(node_id)) {
+			if (copier->gtw_cfg.config_length) {
```
kv2019i, line 401: Is this this intentional? Before we went to this branch if config_length was non-zero, not we go only if "is_multi_gateway()" is true.

src/audio/copier/copier.c
```diff
@@ -780,6 +826,178 @@ static int copier_comp_trigger(struct comp_dev *dev, int cmd)
...
+		for (endp_channel = 0; endp_channel < endp_buf_c->stream.channels;
+				endp_channel++) {
+			int multi_buf_channel = endp_buf_c->chmap[endp_channel];
+
+			if (container_size == 2) {
+				copy_single_channel_c16(&multi_buf_c->stream, multi_buf_channel,
+							&endp_buf_c->stream, endp_channel,
+							avail_frames);
+			} else {
+				assert(container_size == 4);
```
kv2019i, line 958: Hmm, audio stream config is coming from outside (which is not trusted), so assert() is doesn't seem like the correct option here.

kv2019i (commented):
Thanks @serhiy-katsyuba-intel ! A few comments inline.

src/audio/copier/copier.c
serhiy-katsyuba-intel: It does work. nibble_map is a channel map in which each channel is represented by 4 bits (total 8 channels in uint32_t) as value from 0 to 7 and a special value 0xf means no channel (mute). For example, map for stereo would be 0x00000010.
Here for i == 0 (channel 0) value 0 will be ORed, i.e., OR will do nothing. However, since nibble_map is initialized with 0, the result will be 0 and this is what we want (channel 0 is represented as 0x0 nibble).

src/audio/copier/copier.c
serhiy-katsyuba-intel: Yes, that if is_multi_gateway(...) was added intentionally. Quote from the spec:

_... Configuration BLOB depends on the gateway type:
Single gateway: unspecified,
Multi gateway: set to AlhMultiGtwCfg,
External VAD gateway: set to ExtVadGtwCfg. ..._

So checking for blob length is not enough, it might be ExtVadGtwCfg instead of AlhMultiGtwCfg, or something unspecified for single gateway.

src/audio/copier/copier.c
serhiy-katsyuba-intel: No, seems we do not have such functionality in the library. This is a coping of particular channel from one stream to another. Here it is used to multiplex or demultiplex streams according to channel map.

src/audio/copier/copier.c
```diff
@@ -780,6 +826,178 @@ static int copier_comp_trigger(struct comp_dev *dev, int cmd)
...
+{
+	int16_t *r_ptr = audio_stream_wrap(src, (int16_t *)src->r_ptr + src_channel);
+	int16_t *w_ptr = audio_stream_wrap(dst, (int16_t *)dst->w_ptr + dst_channel);
+	int i;
+
+	for (i = 0; i < frame_count; i++) {
+		*w_ptr = *r_ptr;
+		r_ptr = audio_stream_wrap(src, r_ptr + src->channels);
+		w_ptr = audio_stream_wrap(dst, w_ptr + dst->channels);
+	}
```
lyakh, line 848: we're now using loops to avoid wrapping on each iteration, see earlier work by @singalsu

serhiy-katsyuba-intel: Reimplemented to avoid wrapping on each iteration.

src/audio/copier/copier.c
serhiy-katsyuba-intel: Not sure if I understand the comment correctly. This change here is required: when multiplexing/demultiplexing of multiple links is used, struct sof_ipc_stream_params contains multiplexed stream channel count, individual links will have less channels. Channel count for such demultiplexed links is received from channel map found in struct ipc4_alh_multi_gtw_cfg (comes in blob).
If ALH multi-gateway feature is not used, we only execute one "if" in copier_copy(), so the changes does not really affect performance in this case.

src/audio/copier/copier.c
serhiy-katsyuba-intel: Replaced with switch() and error reporting for unsupported container size.

src/audio/copier/copier.c
```diff
@@ -780,6 +836,249 @@ static int copier_comp_trigger(struct comp_dev *dev, int cmd)
...
+	for (endp_idx = 0; endp_idx < cd->endpoint_num; endp_idx++) {
+		uint32_t bytes_produced;
+
+		endp_buf_c = buffer_acquire(cd->endpoint_buffer[endp_idx]);
+
+		for (endp_channel = 0; endp_channel < endp_buf_c->stream.channels;
+				endp_channel++) {
+			int multi_buf_channel = endp_buf_c->chmap[endp_channel];
+
+			switch (container_size) {
```
lgirdwood, line 952: container_size is constant within both loops. Could be optimized, especially if this is called as part of pipeline processing.

src/audio/copier/copier.c
ranj063: so if bitmask is 0x1, nibble_map should be 0xFFFFFFF0?

serhiy-katsyuba-intel: > so if bitmask is 0x1, nibble_map should be 0xFFFFFFF0?

Yes.

src/audio/copier/copier.c
serhiy-katsyuba-intel: Modified to select channel copy func at copier_prepare().

serhiy-katsyuba-intel:
@lgirdwood , checkpatch warning is kind of false positive:
```
WARNING: do not add new typedefs
#564: FILE: src/include/ipc4/copier.h:324:
+typedef void (* channel_copy_func)(struct audio_stream __sparse_cache *dst,
```
There is no non-ugly way to avoid adding typedef there.

src/audio/copier/copier.c
ranj063: Can this function be simplified then as 
```
static uint32_t bitmask_to_nibble_channel_map(uint8_t bitmask)
{
	uint32_t nibble_map;
	int channel_count = popcount(bitmask);
	int i;
	
	nibble_map = GEN_MASK(31, channel_count);
	for (i = 0; i < channel_count; i++)
		nibble_map |= i << (i * 4);

	return nibble_map;
}
````

Comment (serhiy-katsyuba-intel) [reply src/audio/copier/copier.c:62]: No. Bitmask could be like 00000101 binary, meaning two channels with number 0 and 2. This potentially could be later multiplexed with some other channels coming from other gateway to create, for example, a 4 channel stream with channels 0,1,2,3.

Comment (ranj063) [reply src/audio/copier/copier.c:62]: So irrespective of the bitmask 0b11 or 0b101, the nibble mask remains the same 0xFFFFFF10?

Comment (serhiy-katsyuba-intel) [reply src/audio/copier/copier.c:62]: No, it are 0xFFFFFF10 and 0xFFFFFF20 respectively.

**Review by jxstelter — APPROVED** [review]

**Review by softwarecki — APPROVED** [review]

**Review by tmleman — APPROVED** [review]

**Review by wszypelt — APPROVED** [review]

**Review by lgirdwood — APPROVED** [review]

**Comment by lgirdwood** [comment]
@bardliao fyi - your WOV testing can now start.

**Comment by bardliao** [comment]
> @bardliao fyi - your WOV testing can now start.

@lgirdwood I did a quick test with my echo reference topology. It looks like we still have something need to do.
Below is the error from kernel and firmware.
```
[  307.594426] snd_sof:sof_ipc4_widget_setup: sof-audio-pci-intel-tgl 0000:00:1f.3: Create widget copier.ALH.31.1 instance 0 - pipe 31 - core 0
[  307.594428] snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx      : 0x40000004|0x36: MOD_INIT_INSTANCE [data size: 216]
[  307.594826] snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx reply: 0x60000068|0x36: MOD_INIT_INSTANCE
[  307.594828] sof-audio-pci-intel-tgl 0000:00:1f.3: FW reported error: 104 - Other failure of module instance initialization request
[  307.594837] sof-audio-pci-intel-tgl 0000:00:1f.3: ipc error for msg 0x40000004|0x36
[  307.594842] sof-audio-pci-intel-tgl 0000:00:1f.3: failed to create module copier.ALH.31.1
```
```
[    0.073370] <inf> ipc: rx    : 0x40000004|0x36
[    0.073401] <inf> ipc: buffer new size 0x600 id 0.0 flags 0x0
[    0.073436] <err> dai_comp: dai_new(): dai_get() failed to create DAI.
[    0.073465] <err> copier: comp:0 0x40000 failed to create dai
[    0.073476] <err> copier: unable to create dai
[    0.073488] <err> ipc: error: failed to init module 4 : 0
[    0.073491] <err> ipc: ipc4: MODULE_MSG failed with err 104
```

I will take a look after new year.

**Comment by serhiy-katsyuba-intel** [comment]
@bardliao , to use multi-gateway feature this Zephyr commit is also required: https://github.com/zephyrproject-rtos/zephyr/pull/53066 . It adds more ALH devices to MTL device tree, previously only 2 ALH devices were in device tree. The commit is already merged to Zephyr, however, our thesofproject/sof repo uses some older Zephyr (from revision in west.yaml). You can try to cherry-pick that commit to your Zephyr sources, or, hopefully, Zephyr revision will be updated in thesofproject/sof soon.

**Comment by bardliao** [comment]
@serhiy-katsyuba-intel I tested on TGL devices and I apply the same change to `dts/xtensa/intel/intel_adsp_cavs25.dtsi`. 
However, current alh dai_index for SoundWire link 1 is 0x100 and for SoundWire link 2 is 0x200, but it seems that the valid dai_index are 0, 1, 2, and 3.
After applying below change, playback works.
```patch
diff --git a/src/audio/copier/copier.c b/src/audio/copier/copier.c
index 190e25d508a3..4c18d6dd6b24 100644
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -415,7 +415,7 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
                                }
                                for (i = 0; i < dai_count; i++)
                                        dai_index[i] =
-                                       IPC4_ALH_DAI_INDEX(alh_blob->alh_cfg.mapping[i].alh_id);
+                                       IPC4_ALH_DAI_INDEX(alh_blob->alh_cfg.mapping[i].alh_id) >> 8;
                        } else {
                                comp_err(parent_dev, "No ipc4_alh_multi_gtw_cfg found in blob!");
                                return -EINVAL;
```
I am not sure if the above change makes sense.

I will get IO error when I do multi-gateways capture. I can see a lot of `no bytes to copy` and `nothing to copy` in the trace.
```
[    0.079703] <inf> host_comp: comp:1 0x40001 no bytes to copy, 0 avail in buffer, 768 free in DMA
[    0.079736] <wrn> dai_comp: comp:0 0x40000 dai_copy(): nothing to copy
[    0.079743] <wrn> dai_comp: comp:0 0x40000 dai_copy(): nothing to copy
[    0.080711] <inf> host_comp: comp:1 0x40001 no bytes to copy, 0 avail in buffer, 768 free in DMA
[    0.080725] <wrn> dai_comp: comp:0 0x40000 dai_copy(): nothing to copy
[    0.080731] <wrn> dai_comp: comp:0 0x40000 dai_copy(): nothing to copy
[    0.081676] <inf> host_comp: comp:1 0x40001 no bytes to copy, 0 avail in buffer, 768 free in DMA
[    0.081690] <wrn> dai_comp: comp:0 0x40000 dai_copy(): nothing to copy
[    0.081696] <wrn> dai_comp: comp:0 0x40000 dai_copy(): nothing to copy
```
The full FW trace is [out.txt](https://github.com/thesofproject/sof/files/10334410/out.txt)

bardliao:
@serhiy-katsyuba-intel In fact, this is an common issue not just multi-gateways issue. See https://github.com/thesofproject/sof/issues/6903

serhiy-katsyuba-intel:
@bardliao , that IPC4_ALH_DAI_INDEX() subtracts 7 from its supplied value (I do not know why). So it kind of expects v_index values starting from 7. Not sure if it's because of a bug or made intentionally.

bardliao:
@serhiy-katsyuba-intel The dai_index issue was fixed by [intel_adsp_cavs25.diff.txt](https://github.com/thesofproject/sof/files/10336488/intel_adsp_cavs25.diff.txt) I can get the same result as IPC4 tester now.

PR #6834, opened 2022-12-18
Happy new year SOF - Windows CI is coming to town!

aborisovich:
### Adding Windows CI for Zephyr platforms by adding new "build-windows" job.

**Content**

Five commits:

1. Changed anytree.RenderTree style to ASCII style in the `xtensa-build-zephyr.py`. Github action console was failing to print fancy lines used in the previous style.
2. Added build-windows job to existing Zephyr workflow.
This builds SOF with Zephyr on Windows using Zephyr SDK and MSYS2.
"Manual environment setup" is done when comparing to linux build step (that uses Zephyr Docker container with ready-to-go environment). Windows build uses Github cache system to speed up the job and avoid excessive downloads from external servers.

3. Renamed Zephyr "build" job to "build-linux" job.

4. Added concurrency group to Zephyr workflow so the previous actions are stopped when new push to "similar context" is done. Eg. pushing fast to pull request will now stop previous uncompleted actions from running and start a new one with newest push. Same rule applies to triggering workflow manually with workflow_dispatch and daily tests.

5. Added build artifacts upload. Github allows to store build artifacts from jobs for 90 days.
This may be useful to:
  -examine compile commands by downloading compile_commands.json file
  -downloading fw .ri file to run it on the simulator
  -downloading for future Zephyr unit tests

**Some context and implementation decisions**
* [Zephyr SDK](https://github.com/zephyrproject-rtos/sdk-ng/releases/tag/v0.15.2) is downloaded once from Zephyr website then from [Github cache](https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28) (10GB total per project as of day of writing).
* [Wget](https://gnuwin32.sourceforge.net/packages/wget.htm) and [unzip](https://gnuwin32.sourceforge.net/packages/unzip.htm) tools are not present on Windows by default. Downloading from
gnuwin32.sourceforge.net once, then from [Github cache](https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28).
* Q: "Why not using wget and unzip from MSYS2 which is used anyway?
A: MSYS2 packages may not be cached in Github and we would re-download them every time.
* Q: "GCC and OpenSSL is provided by MSYS2. You just wrote that MSYS2 may not be cached. Why not just download windows binaries and cache them?"
A: GCC downloaded from other packages like chocolatey is still missing some linux specific system headers and compilation fails (rimage and sof-logger). OpenSSL used by rimage is currently taken from MSYS2 installation in rimage CMake files.
We had tried before to use OpenSSL binaries for Windows with rimage but there are implementation differences and it would require fixing up the code.

### DIFF
--- a/.github/workflows/zephyr.yml
+++ b/.github/workflows/zephyr.yml
@@ -77,6 +77,14 @@ jobs:
              ./sof/zephyr/docker-build.sh --cmake-args=-DEXTRA_CFLAGS=-Werror
              --cmake-args=--warn-uninitialized ${{ matrix.IPC_platforms }}
 
+      - name: Upload build artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: linux-build-output
+          path: |
+            ${{ github.workspace }}/workspace/build-sof-staging
+            ${{ github.workspace }}/workspace/**/compile_commands.json
+
   build-windows:
     runs-on: windows-latest
     strategy:
@@ -206,3 +214,11 @@ jobs:
           --no-interactive
           --cmake-args=-DEXTRA_CFLAGS=-Werror
           --cmake-args=--warn-uninitialized ${{ matrix.platforms }}
+
+      - name: Upload build artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: windows-build-output
+          path: |
+            ${{ github.workspace }}/workspace/build-sof-staging
+            ${{ github.workspace }}/workspace/**/compile_commands.json
--- a/.github/workflows/zephyr.yml
+++ b/.github/workflows/zephyr.yml
@@ -10,7 +10,6 @@ on: [push, pull_request, workflow_dispatch, workflow_call]
 jobs:
   build:
     runs-on: ubuntu-20.04
-
     strategy:
       fail-fast: false
       matrix:
@@ -70,3 +69,133 @@ jobs:
         run: cd workspace && ./sof/zephyr/docker-run.sh
              ./sof/zephyr/docker-build.sh --cmake-args=-DEXTRA_CFLAGS=-Werror
              --cmake-args=--warn-uninitialized ${{ matrix.IPC_platforms }}
+
+  build-windows:
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      # FIXME: Create common matrix for build-linux and build-windows
+      matrix:
+        # Using groups to avoid spamming the small results box with too
+        # many lines. Pay attention to COMMAS.
+        platforms: [
+          # - IPC3 default
+          imx8 imx8x imx8m,
+          tgl tgl-h,  # UNSUPPORTED! Will be removed
+          # - IPC4 default
+          mtl,
+          # Very few IPC3 platforms support IPC4 too.
+          -i IPC4 tgl tgl-h,
+        ]
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          path: ./workspace/sof
+
+      # Cache artifacts so we do not overload external servers with downloads
+      # Remember to change step key if you change the tools so old cache is not restored,
+      # or delete cache manually in Github Actions tab
+      - name: Cache unzip
+        id: cache-unzip
+        uses: actions/cache@v3.0.11
+        with:
+          path: unzip-5.51-1-bin.zip
+          key: ${{ runner.os }}-cache-unzip-5-51-1
+
+      - name: Cache wget
+        id: cache-wget
+        uses: actions/cache@v3.0.11
+        with:
+          path: wget-1.11.4-1-bin.zip
+          key: ${{ runner.os }}-cache-wget-1-11-4-1
+
+      - name: Cache Zephyr SDK
+        id: cache-zephyr-sdk
+        uses: actions/cache@v3.0.11
+        with:
+          path: zephyr-sdk-0.15.2_windows-x86_64.zip
+          key: ${{ runner.os }}-cache-zephyr-sdk-0-15-2
+
+      # Unzip is needed by Zephyr SDK setup.cmd installation script
+      - name: Download unzip
+        if: ${{ steps.cache-unzip.outputs.cache-hit != 'true' }}
+        run: |
+          curl -L -o unzip-5.51-1-bin.zip `
+          https://gnuwin32.sourceforge.net/downlinks/unzip-bin-zip.php
+
+      # Wget is needed by Zephyr SDK setup.cmd installation script
+      - name: Download wget
+        if: ${{ steps.cache-wget.outputs.cache-hit != 'true' }}
+        run: |
+          curl -L -O http://downloads.sourceforge.net/gnuwin32/wget-1.11.4-1-bin.zip
+
+      - name: Download Zephyr SDK
+        if: ${{ steps.cache-zephyr-sdk.outputs.cache-hit != 'true' }}
+        run: |  # yamllint disable-line rule:line-length
+          curl -L -O `
+          https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v0.15.2/zephyr-sdk-0.15.2_windows-x86_64.zip
+
+      # Unzips every .zip package to directory matching its name without extension
+      - name: Unzip downloaded packages
+        run: 7z x *.zip -o*
+
+      - name: Add unzip to system PATH
+        run: |
+          echo "${{ github.workspace }}/unzip-5.51-1-bin/bin" | `
+          Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+
+      - name: Add wget to system PATH
+        run: |
+          echo "${{ github.workspace }}/wget-1.11.4-1-bin/bin" | `
+          Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+
+      # Install Zephyr SDK - all toolchains including Host Tools
+      # and registering CMake package in the registry
+      # setup.cmd may not be called in from msys shell as it does not parse
+      # forward slash script input arguments correctly.
+      - name: Install Zephyr SDK
+        run: zephyr-sdk-0.15.2_windows-x86_64/zephyr-sdk-0.15.2/setup.cmd /t all /h /c
+
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.8'
+
+      - name: West install
+        run: pip3 install west
+
+      - name: West clone
+        working-directory: ${{ github.workspace }}/workspace
+        run: west init -l sof && west update
+
+      # Call Setup Python again to save the PIP packages in cache
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        id: cache-python
+        with:
+          python-version: '3.8'
+          cache: 'pip'
+          cache-dependency-path: workspace/zephyr/scripts/requirements.txt
+
+      # All requirements will be satisfied if the restored cache matches existing state
+      - name: Validate python PIP cache
+        working-directory: ${{ github.workspace }}/workspace
+        run: pip install -r zephyr/scripts/requirements.txt
+
+      # MSYS2 provides gcc x64_86 toolchain & openssl
+      - name: Initialize MSYS2
+        uses: msys2/setup-msys2@v2
+        with:
+          msystem: MSYS
+          install: gcc openssl-devel
+          path-type: inherit
+
+      - name: Build
+        shell: msys2 {0}
+        working-directory: ${{ github.workspace }}/workspace
+        run: python sof/scripts/xtensa-build-zephyr.py
+          --no-interactive
+          --cmake-args=-DEXTRA_CFLAGS=-Werror
+          --cmake-args=--warn-uninitialized ${{ matrix.platforms }}
--- a/.github/workflows/zephyr.yml
+++ b/.github/workflows/zephyr.yml
@@ -8,7 +8,7 @@ name: Zephyr
 on: [push, pull_request, workflow_dispatch, workflow_call]
 
 jobs:
-  build:
+  build-linux:
     runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
--- a/.github/workflows/zephyr.yml
+++ b/.github/workflows/zephyr.yml
@@ -7,6 +7,13 @@ name: Zephyr
 # yamllint disable-line rule:truthy
 on: [push, pull_request, workflow_dispatch, workflow_call]
 
+# Specifies group name that stops previous wokrflows if the name matches
+concurrency:
+  # eg. "Zephyr-pull_request-my_fork_branch_to_merge"
+  # eg. "Zephyr-push-refs/heads/my_branch_merging"
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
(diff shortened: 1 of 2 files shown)


aborisovich:
It looks like we have some bugs in the python `xtensa-build-zephyr.py` script on Windows during last "install step".

aborisovich:
Converting back to draft, looks like this PR is still not ready 😞

.github/workflows/zephyr.yml
```diff
@@ -7,6 +7,13 @@ name: Zephyr
 # yamllint disable-line rule:truthy
 on: [push, pull_request, workflow_dispatch, workflow_call]
 
+# Specifies group name that stops previous wokrflows if the name matches
+concurrency:
+  # eg. "Zephyr-pull_request-my_fork_branch_to_merge"
+  # eg. "Zephyr-push-refs/heads/my_branch_merging"
```
kv2019i, line 13: typoe: wokrflows

The examples are nice, help to review!

.github/workflows/zephyr.yml
```diff
@@ -70,3 +69,134 @@ jobs:
...
+        # Using groups to avoid spamming the small results box with too
+        # many lines. Pay attention to COMMAS.
+        platforms: [
+          # - IPC3 default
+          imx8 imx8x imx8m,
+          tgl tgl-h,  # UNSUPPORTED! Will be removed
+          # - IPC4 default
+          mtl,
+          # Very few IPC3 platforms support IPC4 too.
+          -i IPC4 tgl tgl-h,
```
kv2019i, line 94: I wonder if threre's a way to share the platform list with Windows? Not a huge thing though, can be handled later.

kv2019i (approved):
Thanks @aborisovich , this is a nice addition! Few minor comments inline, but nothing blocking.

.github/workflows/zephyr.yml
aborisovich: Yeah, I've spend way too much time trying to implement this and finally gave up 😞 
Github right now does not allow to create a common matrix for few jobs.
What can be done (I've tried it), is to create a job A that needs job B to complete and passing input from one job to another. So the common matrix can be implemented like so:
1. create a "read-matrix" job that reads a matrix from json file
2. add build-windows and build-linux job that needs "read-matrix" job to complete and receives that matrix as job output
3. build-linux assigns received matrix as own job matrix filtering out windows array elements, same build-windows filtering out linux elements.

Unfortunately, parsing the matrix from json is very buggy - the moment where you read the json and try to assign received matrix as job matrix. Basically GH is picky about every single whitespace in the json file for some reason.
I had to remove all whitespaces (including space characters) from my json file and still there were issues parsing "-i IPC4 TGL" (since it has spaces inside).

aborisovich:
Ok  build-windows is passing now. Found the cause of the issue causing python script to fail.
Added commit a9fdc68a400b050d103e2804c00e1af0703ed643 to fix that. Funny fact - turned out that Github action console displaying the logs can't handle fancy lines drawn by anytree graph:
image
Changed the style to ASCII so now it will look like this:
image

.github/workflows/zephyr.yml
aborisovich: Fixed 👍

aborisovich:
Added also commit 29d63e1759c3700921be3e7fe8bdeb6f535d0a9c that uploads builds artifacts that can be downloaded from Github Actions tab here:
image
But uploading whole `build-platform` directory and `build-sof-staging` directory takes ~2m 30 seconds.
Maybe we should reduce to uploading only `build-sof-staging` directory and `compile_commands.json` file to limit this time?
This windows build is much longer than linux anyway...

aborisovich:
@marc-hb what do you think? Looking forward for your feedback.

aborisovich:
Ok, rerun the CI - caching is working, Python PIP packages are restored from cache too ❤️ 
image
image

.github/workflows/zephyr.yml
```diff
@@ -70,3 +69,135 @@ jobs:
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.8'
+          cache: 'pip'
+          cache-dependency-path: "**\requirements*.txt"
+
+      - name: West install
+        run: pip3 install west
+
+      - if: ${{ matrix.zephyr_revision }} == 'main'
```
marc-hb, line 179: Maybe drop this to simplify and reduce the duplication with Linux. Building the manifest version should be enough with Windows, Linux should be enough to find upstream regressions. This will also reduce the number of "checks" in the too small window, it's getting too big.

.github/workflows/zephyr.yml
```diff
@@ -70,3 +69,135 @@ jobs:
...
+          Set-Content -path .\sof\submanifests\zephyr-override.yml
+
+      - name: West clone
+        working-directory: ${{ github.workspace }}\workspace
+        run: west init -l sof && west update
+
+      # All requirements will be satisfied if the restored cache matches existing state
+      - name: Validate python PIP cache
+        working-directory: ${{ github.workspace }}\workspace
+        run: pip install -r zephyr\scripts\requirements.txt
```
marc-hb, line 194: See comment above about forward slashes. This could make quoting even harder.

.github/workflows/zephyr.yml
```diff
@@ -70,3 +69,135 @@ jobs:
...
+      - name: Install Zephyr SDK
+        run: zephyr-sdk-0.15.2_windows-x86_64\zephyr-sdk-0.15.2\setup.cmd /t all /h /c
+
+      # Cache dependency path is always relative to ${{ github.workspace }}
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.8'
+          cache: 'pip'
+          cache-dependency-path: "**\requirements*.txt"
```
marc-hb, line 174: Please don't install a lot of "random" stuff
```suggestion
    cache-dependency-path: |
      sof/scripts/requirements.txt
      zephyr/scripts/requirements-base.txt
      zephyr/scripts/requirements-build-test.txt
```

.github/workflows/zephyr.yml
```diff
@@ -70,3 +69,135 @@ jobs:
...
+          tgl tgl-h,  # UNSUPPORTED! Will be removed
+          # - IPC4 default
+          mtl,
+          # Very few IPC3 platforms support IPC4 too.
+          -i IPC4 tgl tgl-h,
+        ]
+        zephyr_revision: [
+          manifest_revision,
+          main,
+        ]
```
marc-hb, line 99: Drop this matrix, see below.

.github/workflows/zephyr.yml
```diff
@@ -70,3 +69,135 @@ jobs:
...
+      - if: ${{ steps.cache-unzip.outputs.cache-hit != 'true' }}
+        name: Download unzip
+        run: |
+          curl -L -o unzip-5.51-1-bin.zip `
+          https://gnuwin32.sourceforge.net/downlinks/unzip-bin-zip.php
+
+      - if: ${{ steps.cache-wget.outputs.cache-hit != 'true' }}
+        name: Download wget
+        run: |
+          curl -L -O http://downloads.sourceforge.net/gnuwin32/wget-1.11.4-1-bin.zip
```
marc-hb, line 141: Downloading `wget` with `curl` looks very strange :-) Can you add a one-line comment to tell who needs wget?

> A: MSYS2 packages may not be cached in Github and we would re-download them every time.

Promote this to a comment in the source.

.github/workflows/zephyr.yml
```diff
@@ -70,3 +69,135 @@ jobs:
...
+
+      - if: ${{ steps.cache-zephyr-sdk.outputs.cache-hit != 'true' }}
+        name: Download Zephyr SDK
+        run: |  # yamllint disable-line rule:line-length
+          curl -L -O `
+          https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v0.15.2/zephyr-sdk-0.15.2_windows-x86_64.zip
+
+      # Unzips every .zip package to directory matching its name without extension
+      - name: Unzip downloaded packages
+        run: 7z x *.zip -o*
```
marc-hb, line 151: Similarly, installing unzip with 7z feels very strange. Who needs unzip?

marc-hb (changes requested):
Thanks @aborisovich , this will make a huge difference!

I see nothing _individually_ blocking this but I'm "requesting changes" only because of the number of suggestions I'm making.

> A: GCC downloaded from other packages like chocolatey is still missing some linux specific system headers and compilation fails (rimage and sof-logger). 

Could you file bugs about these? Not expecting them to be fixed soon but so people can google the corresponding error messages.



> Maybe we should reduce to uploading only build-sof-staging directory and compile_commands.json file to limit this time?

Yes build-sof-staging should have all the critical bits and the rest of the build should now be reproducible anyway.

> On the other hand, Windows CI is still much faster than tests on hardware so maybe I should not bother too much?

Yes and no. Yes because it's always great not to be the last one. No because the faster the feedback and the faster developers can start fixing issues - even before other tests are finished.


BTW please don't use screenshots when it's only text:

https://docs.zephyrproject.org/latest/develop/getting_started/index.html#use-copy-paste

.github/workflows/zephyr.yml
```diff
@@ -7,6 +7,13 @@ name: Zephyr
 # yamllint disable-line rule:truthy
 on: [push, pull_request, workflow_dispatch, workflow_call]
 
+# Specifies group name that stops previous wokrflows if the name matches
+concurrency:
+  # eg. "Zephyr-pull_request-my_fork_branch_to_merge"
+  # eg. "Zephyr-push-refs/heads/my_branch_merging"
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.head_ref || github.ref }}
```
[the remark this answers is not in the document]
aborisovich: This one is actually stolen from Zephyr project and is a very smart one ;-)
This part:
`${{ github.head_ref || github.ref }}`
is a "[fallback value](https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value)". Let me quote github.ref value docs:

> The fully-formed ref of the branch or tag that triggered the workflow run. For workflows triggered by push, this is the branch or tag ref that was pushed. For workflows triggered by pull_request, this is the pull request merge branch. For workflows triggered by release, this is the release tag created. For other triggers, this is the branch or tag ref that triggered the workflow run. This is only set if a branch or tag is available for the event type. The ref given is fully-formed, meaning that for branches the format is refs/heads/<branch_name>, for pull requests it is refs/pull/<pr_number>/merge, and for tags it is refs/tags/<tag_name>. For example, refs/heads/feature-branch-1.

So, github.ref would be valid for both "push" and "pull_request" events.
Right now we have a duplication in the events triggering - every PR starts a Zephyr workflow for both "push" and "pull_request" events what is a colossal waist of resources.
So thanks to using ${{ github.head_ref || github.ref }} we avoid duplication of those events in the PRs.
What is happening in detail:
1. You push to fork branch X on created PR to some other branch
2. Both event triggers for Zephyr workflows are valid - "on: push: and "on: pull_request", so they start two copies of the workflow (current implementation).
3. This concurrency group sees, that "on: push" event trigger has valid "github.head_ref" property since the push is done to the branch that has pull request assigned.
4. Thanks to this, workflow triggered by "on: push" has the very same group name that workflow created by "on: pull_request" event.
5. Workflows with group names matching are not being run accourding to concurrency keyword definition.
So, in the pull request only workflows triggered by "on: pull_request" are being launched.

What happens in other scenarios (non Pull Request):
1. When you merge a pull request or run daily tests triggered by scheduler using "on: workflow_call" or invoke Github Action manually using "on: workflow_dispatch", concurrency group name does not have a valid ${{ github.head_ref }} property.
2. Fallback value is used ${{ github.ref }} to establish group name.
3. If you retrigger the run - rerun every action on the very same commit it will cancel old run and start a new one as the group names match.

> "This is an interesting optimization but AFAIK it's only going to save "free" resources."

This saves not only "free" resources, it saves a really lot of resources.
I often wait for CI results because all Github runners are busy and not picking up workflows from my PRs due to high traffic. Now thanks to deduplication of the "pull_request" and "push" events we have more free runners for the project.
Secondly, I often wait for CI to pick up my workflows on PRs because I push frequently and impatiently.
And when I push to fix a single typo I need to wait for all my previous workflows to complete to start new ones.
This optimization  would stop all my previous running workflows and start new ones without waiting.
This is really useful feature (which I would not be smart enough to implement - as previously said stolen from Zephyr project).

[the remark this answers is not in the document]
aborisovich: This should be probably copy pasted to all other workflows in our project in some other PR.

.github/workflows/zephyr.yml
```diff
@@ -208,3 +208,9 @@ jobs:
           --no-interactive
           --cmake-args=-DEXTRA_CFLAGS=-Werror
           --cmake-args=--warn-uninitialized ${{ matrix.platforms }}
+
+      - name: Upload build artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: build-output
+          path: ${{ github.workspace }}\workspace\build-*
```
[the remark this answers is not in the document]
aborisovich: I'll add artifacts upload for linux too ;-)

.github/workflows/zephyr.yml
```diff
@@ -70,3 +69,135 @@ jobs:
...
+        with:
+          fetch-depth: 0
+          path: ./workspace/sof
+
+      # Cache artifacts so we do not overload external servers with downloads
+      # Remember to change step key if you change the tools so old cache is not restored,
+      # or delete cache manually in Github Actions tab
+      - name: Cache unzip
+        id: cache-unzip
+        uses: actions/cache@v3.1.0-beta.3
```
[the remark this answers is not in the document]
aborisovich: Probably not, it was just provided in Github docs. It surprised me too but I just took it if they are using it in the Github official documentation. But yes I think maybe it's just automated generation, I will reduce to stable 3.1.0 version.

.github/workflows/zephyr.yml
aborisovich: Unfortunately, setup.cmd script used to install Zephyr SDK requires it.

.github/workflows/zephyr.yml
aborisovich: see above.

.github/workflows/zephyr.yml
aborisovich: So, it's not exatcly possible to not install "random" stuff.
Because ultimately we need to call PIP install -r zephyr/scripts/requirements.txt that includes many more:
```
-r requirements-base.txt
-r requirements-build-test.txt
-r requirements-doc.txt
-r requirements-run-test.txt
-r requirements-extras.txt
-r requirements-compliance.txt
```
So I'll just remove wildstars and switch to concrete file (maybe this will avoid some randomly laying around requirements.tx files).

aborisovich:
@marc-hb I've fixed all the things you mentioned.
* converted to forward slashes
* added comments and explanations
* fixed python caching
* removed revisions from the matrix
* made artifacts upload smaller and added to build-linux job
Caches are working as expected. Artifacts size of windows is 5MB bigger than linux (what is interesting taken we build same platforms). Good to go now?

marc-hb (approved):

marc-hb:
> @marc-hb I've fixed all the things you mentioned.

I've spotted a few left but can be addressed later.

This will make a massive difference, thanks again.

lgirdwood (approved):

PR #6835, opened 2022-12-19
platform: ace: dma: update buffer period count

ArsenEloglian:
Double the dma buffer period count as it
should be 128 for 1 channel, 256 for 2, etc.

Signed-off-by: Arsen Eloglian <ArsenX.Eloglian@intel.com>

### DIFF
--- a/src/platform/intel/ace/lib/dma.c
+++ b/src/platform/intel/ace/lib/dma.c
@@ -18,7 +18,7 @@
 #include <zephyr/device.h>
 
 #define DW_DMA_BUFFER_PERIOD_COUNT	0x4
-#define HDA_DMA_BUFFER_PERIOD_COUNT	2
+#define HDA_DMA_BUFFER_PERIOD_COUNT	4
 
 SHARED_DATA struct dma dma[] = {
 #if DT_NODE_HAS_STATUS(DT_NODELABEL(lpgpdma0), okay)


src/platform/intel/ace/lib/dma.c
```diff
@@ -56,7 +56,7 @@ SHARED_DATA struct dma dma[] = {
 		.caps		= DMA_CAP_HDA,
 		.devs		= DMA_DEV_HOST,
 		.channels	= DT_PROP(DT_NODELABEL(hda_host_in), dma_channels),
-		.period_count	= HDA_DMA_BUFFER_PERIOD_COUNT,
+		.period_count	= 2 * HDA_DMA_BUFFER_PERIOD_COUNT,
```
kv2019i, line 59: A couple of points to commit message:

1. Please use "platform: ace: update buffer period count"
This helps reviewers to immediately see this is a change specific to one platform.

2. The reason for change is still a bit lacking. Double compared to what? Why not just increase HDA_DMA_BUFFER_PERIOD_COUNT byt two? These maybe obvious, but please explain in git commit message.

kv2019i (changes requested):
Some comments inline.

lgirdwood (approved):
Do we need the same fix for TGL ?

tmleman (approved):

marcinszkudlinski (approved):

softwarecki (approved):

lgirdwood:
@kv2019i update good for you ?

kv2019i (approved):
Thanks @ArsenEloglian , good now!

PR #6836, opened 2022-12-19
Convert the mixin module to use the module adapter

ranj063:


lgirdwood:
@ranj063 #6833 now merged, this will need a rebase.

lgirdwood (approved):

lgirdwood:
@ranj063 can you check internal CI thanks.

ranj063:
> @ranj063 can you check internal CI thanks.

@lgirdwood looking into it. Looks like the gain quality test is failing.

lgirdwood:
> > @ranj063 can you check internal CI thanks.
> 
> @lgirdwood looking into it. Looks like the gain quality test is failing.

Thanks, still showing an issue.

ranj063:
SOFCI TEST

ranj063:
SOFCI TEST

ranj063:
SOFCI TEST

ranj063:
SOFCI TEST

src/audio/module_adapter/module_adapter.c
```diff
@@ -623,54 +630,126 @@ static void module_adapter_process_output(struct comp_dev *dev)
...
+		buffer_stream_invalidate(source_c[i], c.frames * c.source_frame_bytes);
+
+		/*
+		 * note that the size is in number of frames not the number of
+		 * bytes
+		 */
+		mod->input_buffers[num_input_buffers].size = c.frames;
+		mod->input_buffers[num_input_buffers].consumed = 0;
+
+		mod->input_buffers[num_input_buffers].data = &source_c[i]->stream;
```
lyakh, line 664: and here's why: to me this looks like your function receiver two arrays of pointers as parameters, i.e. it made local copies of those arrays. And here you store pointers to those (local copied of) array elements to global data. Maybe that isn't what the compiler does in this case, but it does look confusing to me.

src/audio/module_adapter/module_adapter.c
```diff
@@ -623,54 +630,126 @@ static void module_adapter_process_output(struct comp_dev *dev)
...
+			continue;
+		}
+
+		comp_get_copy_limits_frame_aligned(source_c[0], sinks_c[i], &c);
+
+		min_frames = MIN(min_frames, c.frames);
+		source_frame_bytes = c.source_frame_bytes;
+
+		mod->output_buffers[num_output_buffers].size = 0;
+		mod->output_buffers[num_output_buffers].data = &sinks_c[i]->stream;
```
lyakh, line 701: same here

src/audio/module_adapter/module_adapter.c
```diff
@@ -623,54 +630,126 @@ static void module_adapter_process_output(struct comp_dev *dev)
...
-
-			mod->input_buffers[num_input_buffers].data = &source_c[i]->stream;
-			num_input_buffers++;
-			i++;
+			source_c[i++] = buffer_acquire(source);
+		}
+		if (mod->num_output_buffers == 1) {
+			num_input_buffers = module_single_sink_setup(dev, source_c, sinks_c);
+			if (sinks_c[0]->sink && sinks_c[0]->sink->state == dev->state)
+				num_output_buffers = 1;
```
lyakh, line 751: so what does it mean in this case `if (!sinks_c[0]->sink)`?

src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -90,56 +84,45 @@ struct mixout_data {
...
-	audio_stream_fmt_conversion(md->base_cfg.audio_fmt.depth,
-				    md->base_cfg.audio_fmt.valid_bit_depth,
+	audio_stream_fmt_conversion(mod->priv.cfg.base_cfg.audio_fmt.depth,
+				    mod->priv.cfg.base_cfg.audio_fmt.valid_bit_depth,
 				    &frame_fmt, &valid_fmt,
-				    md->base_cfg.audio_fmt.s_type);
+				    mod->priv.cfg.base_cfg.audio_fmt.s_type);
 
 	dev->ipc_config.frame_fmt = frame_fmt;
+	mod->simple_copy = true;
```
lyakh, line 123: mixin / mixout can have both multiple sources and sinks, cannot they? Then it cannot be "simple?"

src/audio/module_adapter/module_adapter.c
```diff
@@ -626,54 +633,124 @@ static void module_adapter_process_output(struct comp_dev *dev)
 	}
 }
 
+static uint32_t
+module_single_sink_setup(struct comp_dev *dev,
+			 struct comp_buffer __sparse_cache *source_c[PLATFORM_MAX_STREAMS],
+			 struct comp_buffer __sparse_cache *sinks_c[PLATFORM_MAX_STREAMS])
```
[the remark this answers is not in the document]
lgirdwood: @lyakh I think this is to be compatible with the module API, i.e. the call to process will pass an array of inputs/outputs.
@ranj063 doesn't the module API use variable size arrays instead of fixed size (as here). I think it's best to align.

src/audio/mixin_mixout/mixin_mixout.c
ranj063: No kixin is single source multi sink and mixout is single sink multi source

src/audio/module_adapter/module_adapter.c
ranj063: It means the sink device is not connected or it is not active yet. This check is quite important for the mixin

src/audio/module_adapter/module_adapter.c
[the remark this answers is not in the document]
ranj063: Sure I can make these pointers

src/audio/module_adapter/module_adapter.c
lyakh: the device isn't connected but the buffer is already there?

ranj063: The first condition is unlikely @lyakh . It's only there as a paranoia check.

lyakh: Sorry, I'm not a big fan of such "paranoia checks." I understand this software is complex and probably no single person in the world can know all the paths that can get executed by it. In cases like this maybe at least we can put an explicit check if we don't know if that can happen or not: if we don't want to make this an assert and risk an exception, at least just adding an error print could be enough.

src/include/sof/audio/module_adapter/module/generic.h
```diff
@@ -184,8 +184,8 @@ struct processing_module {
 	uint32_t num_input_buffers; /**< number of input buffers */
 	uint32_t num_output_buffers; /**< number of output buffers */
 	/*
-	 * flag set by a module when it has N input buffer and 1 output buffer and produces
-	 * period_bytes every copy
+	 * flag set by a module produces period_bytes every copy. It can be used by modules
```
kv2019i, line 187: grammar: "by a module that produces"

src/audio/module_adapter/module_adapter.c
```diff
@@ -623,54 +630,126 @@ static void module_adapter_process_output(struct comp_dev *dev)
...
 	struct comp_buffer *source, *sink;
 	struct comp_buffer __sparse_cache *source_c[PLATFORM_MAX_STREAMS];
+	struct comp_buffer __sparse_cache *sinks_c[PLATFORM_MAX_STREAMS];
 	struct comp_buffer __sparse_cache *sink_c = NULL;
-	struct comp_copy_limits c;
 	struct list_item *blist;
 	size_t size = MAX(mod->deep_buff_bytes, mod->period_bytes);
 	uint32_t min_free_frames = UINT_MAX;
 	uint32_t num_input_buffers = 0;
+	uint32_t num_output_buffers = 0;
```
kv2019i, line 729: Not really specific to this PR, but the "copy()" starts to be very complex. I wonder if some of this logic could be moved to some less-often-run function. Could we even have a separate copy for the "simple_copy case"?

src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -626,6 +600,70 @@ static int mixout_reset(struct processing_module *mod)
...
+	struct comp_dev *dev = mod->dev;
+	struct list_item *blist;
+	int ret;
+
+	comp_dbg(dev, "mixin_params()");
+
+	base_module_cfg_to_stream_params(&mod->priv.cfg.base_cfg, params);
+
+	/* Buffers between mixins and mixouts are not used (mixin writes data directly to mixout
+	 * sink). But, anyway, let's setup these buffers properly just in case.
```
kv2019i, line 622: This feels quit out-of-place, but as this is existing comment (and code) moved as-is, not need to block because of this.

src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -121,6 +121,7 @@ static int mixin_init(struct processing_module *mod)
 
 	dev->ipc_config.frame_fmt = frame_fmt;
 	mod->simple_copy = true;
+	mod->skip_src_buffer_invalidate = true;
```
kv2019i, line 124: Minor thing, but the commit leaves a bit open why the modules need the ability to do this. This does add two code branches to a very common code path (the module process). I wonder if it's more efficient to just remove the extra writebacks from modules, and if there a few cases where this can't be done, pay the cost only for these cases.

kv2019i (commented):
Nothing major, some minor comments, please see inline.

kv2019i:
@wszypelt "System/merge/codecheck" error hit with this one as well..

src/audio/mixin_mixout/mixin_mixout.c
ranj063: @kv2019i I think we need to clean this up for sure to not allow modules to do the WB/invalidates. But for now, I want to leave it as is to reduce the amount of changes.

src/audio/module_adapter/module_adapter.c
```diff
@@ -623,54 +630,126 @@ static void module_adapter_process_output(struct comp_dev *dev)
...
+{
+	struct processing_module *mod = comp_get_drvdata(dev);
+	struct comp_copy_limits c;
+	struct list_item *blist;
+	uint32_t num_input_buffers = 0;
+	int i = 0;
+
+	list_for_item(blist, &dev->bsource_list) {
+		/* check if the source dev is in the same state as the dev */
+		if (!source_c[i]->source || source_c[i]->source->state != dev->state) {
```
lyakh, line 626: I still don't like this. We're in `.copy()` i.e. in the pipeline task, actually performing streaming, right? These buffers can only be modified when modifying topology, i.e. either when connecting or when disconnecting pipelines. Those actions should also only be done in a pipeline task context. As long as we only allow connected pipelines on the same core, these tasks will run sequentially without preempting each other. So within one scheduling invocation these buffers remain consistent.
Only when we move to connecting different cores, then this will become a problem. But even then this will be a larger problem, requiring proper robust locking. And if we ever hit this condition, it means we were in an inconsistent state and it can bite us elsewhere. These checks seem only to reduce the probability of hitting a problem, making it more difficult to debug and fix. Really don't seem useful to me. Same below of course.

kv2019i: @lyakh I'm fully with you that this does not seem right in copy (we should have less frequent state change callbacks where module can react to changes in its configuration or the configuration of its immediate connections, and NOT do this check in each copy), but it's still a fact that currently this is hit in pause-resume tests (with sof-tests), so it's not like this is a corner case at this point.

src/audio/module_adapter/module_adapter.c
```diff
@@ -814,17 +793,42 @@ int module_adapter_copy(struct comp_dev *dev)
...
 
 	if (mod->simple_copy) {
+		/* consume from all active input buffers */
+		for (i = 0; i < num_input_buffers; i++) {
+			struct comp_buffer __sparse_cache *src_c;
+
+			src_c = attr_container_of(mod->input_buffers[i].data,
+						  struct comp_buffer __sparse_cache,
+						  stream, __sparse_cache);
+			comp_update_buffer_consume(src_c, mod->input_buffers[i].consumed);
```
kv2019i, line 803: Yup, this is cleaner,

kv2019i (approved):
Thanks, this is cleaner than my https://github.com/thesofproject/sof/pull/6909 . 

The concern raised by Guennadi on whether copy()/process() has to have so much conditional code, remains on open, but not really specific to this PR. Given I can easily hit these conditions with sof-test, it would seem this requires a separate PR (if someone disagrees, please chime in).

src/audio/module_adapter/module_adapter.c
ranj063: oh i missed these 2. fixed now

src/audio/mixin_mixout/mixin_mixout.c
lyakh: I must either remember something wrong then or somewhere in the code there's a location that implicitly assumes that. Will try to make a note of it next time I come across it (if it really exists)

src/audio/mixin_mixout/mixin_mixout.c
lyakh: the comment could be a bit "kinder" though at least, stating some "better sounding" reason than "just in case," e.g. "for correctness," "for consistency" etc. But yes, it's just a verbatim copy of the current comment

lyakh (approved):

src/audio/module_adapter/module_adapter.c
lyakh: > @lyakh I'm fully with you that this does not seem right in copy (we should have less frequent state change callbacks where module can react to changes in its configuration or the configuration of its immediate connections, and NOT do this check in each copy), but it's still a fact that currently this is hit in pause-resume tests (with sof-tests), so it's not like this is a corner case at this point.

@kv2019i if it's so easy to hit, then we really need to investigate and fix these properly. Maybe someone could add a task to our planning ;-)

src/audio/mixin_mixout/mixin_mixout.c
lgirdwood: The cache work needs to go in the module core, however, agree we should do teh module conversion first and then do the cache work (when hopefully the Zephyr cache API is ready).

lgirdwood (approved):

lgirdwood:
@ranj063 btw, what's module adapter plan for N:1 and M:N based processing optimisation ?

ranj063:
> @ranj063 btw, what's module adapter plan for N:1 and M:N based processing optimisation ?

@lgirdwood N:1 is already supported by the module adapter. M:N is the only open. I dont have a line of sight for this one due to the lack of a module for this type. Maybe what we need to think of extending the passthrough module to support M:N confg instead of 1:1 so we can test this. WHat do you think?

PR #6838, opened 2022-12-19
ipc4: Fix error handling issues in ipc4

jxstelter:
The set_pipeline_state() function could potentially end up with uninitialised ret variable. Set it to 0 to avoid returning uninitialised.
Fix from mtl-002-drop-stable: [https://github.com/thesofproject/sof/pull/6719](https://github.com/thesofproject/sof/pull/6719)

### DIFF
--- a/src/ipc/ipc4/handler.c
+++ b/src/ipc/ipc4/handler.c
@@ -275,6 +275,9 @@ static int set_pipeline_state(struct ipc_comp_dev *ppl_icd, uint32_t cmd,
 			}
 			if (ret == PPL_STATUS_SCHEDULED)
 				*delayed = true;
+			break;
+		default:
+			ret = 0;
 		}
 
 		/*


ranj063 (approved):

src/ipc/ipc4/handler.c
```diff
@@ -275,6 +275,9 @@ static int set_pipeline_state(struct ipc_comp_dev *ppl_icd, uint32_t cmd,
 			}
 			if (ret == PPL_STATUS_SCHEDULED)
 				*delayed = true;
+			break;
+		default:
+			ret = 0;
```
ranj063, line 280: would it be better to just return 0 instead?

jxstelter:
Closing. Already fixed in other commit 40e6c1ddc061e04687a5d48e38dc02c37ace0703

src/ipc/ipc4/handler.c
jxstelter: n/a. Fixed in other commit.

PR #6842, opened 2022-12-19
module_adapter: return 0 when component state is equal to requested state

fkwasowi:
When dev->state == requested_state in comp_set_state then COMP_STATUS_STATE_ALREADY_SET (status 1) is returned. Then the pipeline trigger is aborted and status 1 is returned to the pipeline_task_cmd function (status 1 in this function means PPL_STATUS_PATH_STOP). Finally, the status of the pipeline does not change and the IPC response returns success. which causes errors with subsequent changes in the pipeline state (INVALID_IPC_REQUEST)

### DIFF
--- a/src/audio/component.c
+++ b/src/audio/component.c
@@ -71,7 +71,11 @@ int comp_set_state(struct comp_dev *dev, int cmd)
 	if (dev->state == requested_state) {
 		comp_info(dev, "comp_set_state(), state already set to %u",
 			  dev->state);
+#ifdef CONFIG_IPC_MAJOR_4
+		return 0;
+#else
 		return COMP_STATUS_STATE_ALREADY_SET;
+#endif
 	}
 
 	switch (cmd) {


src/audio/component.c
```diff
@@ -71,7 +71,7 @@ int comp_set_state(struct comp_dev *dev, int cmd)
 	if (dev->state == requested_state) {
 		comp_info(dev, "comp_set_state(), state already set to %u",
 			  dev->state);
-		return COMP_STATUS_STATE_ALREADY_SET;
+		return 0;
```
ranj063, line 74: @fkwasowi I think this might end up breaking ipc3. Why not check for the already _set return value in pipeline_task_cmd() for ipc4 and return 0 there?

ranj063: or alternatively check for the IPC version in comp_set_state() and return the appropriate return value for each IPC version ie COMP_STATUS_STATE_ALREADY_SET for IPC3 and 0 for IPC4

ranj063 (approved):

kv2019i (approved):
The prefix seems odd, this should be "component: " right? 

OTOH, looking at git history. the prefix usage for this source file has been pretty inconsistent, so not blocking for this. @lyakh , seems good?

lgirdwood (approved):

lyakh:
@kv2019i @lgirdwood Actually I'd say it's a bug in how almost all component driver `.trigger()` methods do `return comp_set_state(dev, cmd);` In `pipeline_comp_trigger()` we have
```
	err = comp_trigger(current, ppl_data->cmd);
	switch (err) {
	case 0:
		break;
	case PPL_STATUS_PATH_STOP:
		current->pipeline->trigger.aborted = true;
		COMPILER_FALLTHROUGH;
	case PPL_STATUS_PATH_TERMINATE:
		return PPL_STATUS_PATH_STOP;
	default:
		return err;
	}
```
So we expect one of `PPL_*` status codes from `comp_trigger()` and that looks logical. So I would say it's all the `.trigger()` methods that are wrong when returning a completely unrelated `COMP_STATUS_STATE_ALREADY_SET`. Other component driver operations do it correctly, e.g. `eq_iir_prepare()`:
```
	ret = comp_set_state(dev, COMP_TRIGGER_PREPARE);
	if (ret < 0)
		return ret;

		return PPL_STATUS_PATH_STOP;
```

lgirdwood:
@fkwasowi can you check above comment by @lyakh - looks like we need to make this change in more places ?

PR #6846, opened 2022-12-19
sof-logger: print error if -u uart option is given with no infile

kv2019i:
sof-logger -u 115200 -d /lib/firmware/sof-foo.ldc

Leads to silent failure as a NULL is passed to open(). Add explicit error handling for this case.

### DIFF
--- a/tools/logger/logger.c
+++ b/tools/logger/logger.c
@@ -436,6 +436,11 @@ int main(int argc, char *argv[])
 	if (!config.in_file && !config.dump_ldc)
 		config.in_file = "/sys/kernel/debug/sof/etrace";
 
+	if (!config.in_file && baud) {
+		fprintf(stderr, "error: No UART device specified\n");
+		usage();
+	}
+
 	if (config.input_std) {
 		config.in_fd = stdin;
 	} else if (baud) {


paulstelian97 (approved):
I'm surprised this is in here in the first place. There may be another place to check this but here is good too and should be no issue.

paulstelian97:
Oh actually maybe a warning that e.g. the uart isn't used and the regular etrace/DMA trace file is being read instead could be good? Since you're not aborting the program...

tools/logger/logger.c
```diff
@@ -436,6 +436,11 @@ int main(int argc, char *argv[])
 	if (!config.in_file && !config.dump_ldc)
 		config.in_file = "/sys/kernel/debug/sof/etrace";
 
+	if (!config.in_file && baud) {
+		fprintf(stderr, "error: Input file not specified with -u option\n");
```
marc-hb, line 440:
```suggestion
		fprintf(stderr, "error: Input file not specified with -i option\n");
```

marc-hb (changes requested):

ranj063 (approved):

kv2019i:
@paulstelian97 wrote: 
> Oh actually maybe a warning that e.g. the uart isn't used and the regular etrace/DMA trace file is being read 
> instead could be good? Since you're not aborting the program...

....a bit of surprise (to me as well), but usage() actually calls exit, so the program does exit here.

marc-hb (approved):

tools/logger/logger.c
marc-hb: @kv2019i  you thumbed up my comment and resolved it but made no change.

PS: thanks to eagle-eyed @ranj063 for spotting this.

Maybe even better:
```suggestion
		fprintf(stderr, "error: Input device not specified with -i option\n");
```

marc-hb (changes requested):

tools/logger/logger.c
kv2019i: @marc-hb  Sorry, I didn't you had a different issue. Ok, I'll rephrase that message.

kv2019i:
@marc-hb Now updated.

lgirdwood (approved):

marc-hb (approved):

lyakh (approved):

PR #6848, opened 2022-12-19
sof-logger: ensure NULL is not passed to printf/fprintf

kv2019i:
Due to allocation failures, or invalid content in dictionary,
"params" entries in "struct proc_ldc_entry" may be NULL.

In print_entry_params(), the NULL entries may be passed
as arguments to fprintf(). While e.g. glibc handles these without
error, this is not guaranteed behaviour and may result in segfault
on some platforms.

Fix the issue by replacing any NULL entries in process_params()
and separately handling the NULL case in dump_ldc_info().

### DIFF
--- a/tools/logger/convert.c
+++ b/tools/logger/convert.c
@@ -65,6 +65,8 @@ static const char *BAD_PTR_STR = "<bad uid ptr 0x%.8x>";
 #define UUID_LOWER "%s%s%s<%08x-%04x-%04x-%02x%02x-%02x%02x%02x%02x%02x%02x>%s%s%s"
 #define UUID_UPPER "%s%s%s<%08X-%04X-%04X-%02X%02X-%02X%02X%02X%02X%02X%02X>%s%s%s"
 
+static const char *missing = "<missing>";
+
 static int read_entry_from_ldc_file(struct ldc_entry *entry, uint32_t log_entry_address);
 
 char *format_uid_raw(const struct sof_uuid_entry *uid_entry, int use_colors, int name_first,
@@ -249,6 +251,8 @@ static void process_params(struct proc_ldc_entry *pe,
 			/* check for string printing, because it leads to logger crash */
 			log_err("String printing is not supported\n");
 			pe->params[i] = (uintptr_t)log_asprintf("<String @ 0x%08x>", raw_param);
+			if (!pe->params[i])
+				abort();
 			pe->subst_mask |= 1 << i;
 			++i;
 			p += 2;
@@ -257,6 +261,8 @@ static void process_params(struct proc_ldc_entry *pe,
 			/* substitute UUID entry address with formatted string pointer from heap */
 			pe->params[i] = (uintptr_t)asprintf_uuid(p, raw_param, use_colors,
 								 &uuid_fmt_len);
+			if (!pe->params[i])
+				abort();
 			pe->subst_mask |= 1 << i;
 			++i;
 			/* replace uuid formatter with %s */
@@ -268,7 +274,12 @@ static void process_params(struct proc_ldc_entry *pe,
 			/* %pQ format specifier */
 			/* substitute log entry address with formatted entry text */
 			pe->params[i] = (uintptr_t)asprintf_entry_text(raw_param);
-			pe->subst_mask |= 1 << i;
+
+			if (pe->params[i])
+				pe->subst_mask |= 1 << i;
+			else
+				pe->params[i] = (uintptr_t)missing;
+
 			++i;
 
 			/* replace entry formatter with %s */
@@ -1021,7 +1032,7 @@ static int dump_ldc_info(void)
 	while (remaining > 0) {
 		name = format_uid_raw(&uid_ptr[cnt], 0, 0, false, false);
 		uid_addr = get_uuid_key(&uid_ptr[cnt]);
-		fprintf(out_fd, "\t%p  %s\n", (void *)uid_addr, name);
+		fprintf(out_fd, "\t%p  %s\n", (void *)uid_addr, name ? name : missing);
 
 		if (name) {
 			free(name);


kv2019i:
V2: 
- oops, compiler did not like "stdderr", fixed. thanks @ranj063 for spotting

ranj063 (approved):

marc-hb (dismissed):
I see `format_uid_raw()` being used in at least one other place.

Morever `format_uid_raw()` is not even calling `malloc()` directly, there is more indirection.

=> Please fix `log_vasprintf()` once for all instead; that is the function not checking the `malloc()` return value.

marc-hb:
Also, please use `abort()` instead of `exit()` in cases like this because it creates a convenience coredump. I just tested it:

```diff
--- a/tools/logger/logger.c
+++ b/tools/logger/logger.c
@@ -86,6 +86,8 @@ static void usage(void)
        fprintf(stdout, "%s:\t -F filter\t\tUpdate trace filter, format: "
                "<level>=<comp1>[, <comp2>]\n",
                APP_NAME);
+
+       abort();
        exit(0);
 }
```

```shell

./tools/build_tools/logger/sof-logger   


Aborted (core dumped)

coredumpctl list
TIME                           PID  UID  GID SIG     COREFILE EXE                                                             SIZE
Mon 2022-12-19 23:45:19 UTC 236154 1005 1005 SIGABRT present  /home/me/SOF/sof/tools/build_tools/logger/sof-logger 16.8K

coredumpctl gdb 236154

(gdb) backtrace
#4  0x0000563e885ef7cb in usage () at /home/me/SOF/sof/tools/logger/logger.c:90
#5  0x0000563e885f02b4 in main (argc=2, argv=0x7fff61f5dea8) at /home/me/SOF/sof/tools/logger/logger.c:392

```

marc-hb:
`abort()` is debug-friendly on Windows too:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/abort?view=msvc-170

> When the program is compiled in debug mode, the message box displays options to Abort, Retry, or Ignore. If the user chooses Abort, the program terminates immediately and returns an exit code of 3. If the user chooses Retry, a debugger is invoked for just-in-time debugging, if available

kv2019i:
@marc-hb wrote:
> @@ -86,6 +86,8 @@ static void usage(void)
>         fprintf(stdout, "%s:\t -F filter\t\tUpdate trace filter, format: "
>                 "<level>=<comp1>[, <comp2>]\n",
>                 APP_NAME);
> +
> +       abort();

Coredumping when wrong option passed is not really a good option right (in usage()). 

I'm not a fan of abort() in application code, it really should be reserved to system library / runtime code in most cases.

Let me check if I can clean this up.

lgirdwood (approved):

tools/logger/convert.c
```diff
@@ -284,6 +285,18 @@ static void process_params(struct proc_ldc_entry *pe,
 			p += 2;
 		}
 	}
+
+	/* ensure no NULLs in format strings */
+	for (j = 0; j < i; j++) {
+		if (!pe->params[j]) {
+			pe->params[j] = (uintptr_t)log_asprintf("<missing>");
+			pe->subst_mask |= 1 << j;
```
lyakh, line 293: you could just use a single `static const char missing[] = "<missing>";` string and not set `subst_mask`

kv2019i: @lyakh Actually yes, that would be potentially cleaner. This still requires mucking with the subst_mask (which is a bit iffy as in theory subst_mask is not just to indicate how the object was allocated). Let me update the patch.

kv2019i:
V3 update:
 - implemented @lyakh 's idea

tools/logger/convert.c
```diff
@@ -284,6 +287,19 @@ static void process_params(struct proc_ldc_entry *pe,
...
+	/* ensure no NULLs in format strings */
+	for (j = 0; j < i; j++) {
+		if (!pe->params[j]) {
+			pe->params[j] = (uintptr_t)missing;
+			/* ensure object is no free()'ed */
+			pe->subst_mask &= ~(1 << j);
+		}
+
+		if (!pe->params[j])
+			abort();
```
kv2019i, line 300: Actually L299-300 is not needed after I dropped the dynamic alloc of "<missing>". V4 coming...

kv2019i:
V4:
- dropped the no-longer-needed check in V3, no other changes

marc-hb:
https://sof-ci.01.org/sofpr/PR6848/build3091/devicetest/index.html has a worryingly large number of failures but they're all unrelated - sof-logger is not tested anymore on the main branch

marc-hb:
> I looked at this again and misc.c is clearly designed to handle out of memory cases, so I don't think it's appropriate to abort there.

I think you're making this much more complicated than it needs to be.

So we have `f1()` and `f2()` calling `log_vasprintf()` which invokes `malloc()` and does something stupid when it returns `NULL`.

I think you're saying "`f1()` does not trust `log_vasprintf()`, so let's make sure `f2()` does not either".

If `malloc()` starts failing in this tool (practically: never) then all bets are off and `log_vasprintf()` should `abort()` ASAP.

kv2019i:
@marc-hb wrote:
> I think you're making this much more complicated than it needs to be.
[...]
> If `malloc()` starts failing in this tool (in practice: never) then all bets are off and `log_vasprintf()` should `abort()` ASAP. One-line fix.

It's not just malloc fails, but asprintf_entry_text() (calling read_entry_from_ldc_file()) may return NULL for non-alloc failures as well.

lyakh (approved):

tools/logger/convert.c
```diff
@@ -284,6 +287,16 @@ static void process_params(struct proc_ldc_entry *pe,
...
 	}
+
+	/* ensure no NULLs in format strings */
+	for (j = 0; j < i; j++) {
+		if (!pe->params[j]) {
+			pe->params[j] = (uintptr_t)missing;
+			/* ensure object is no free()'ed */
+			pe->subst_mask &= ~(1 << j);
+		}
+	}
```
paulstelian97, line 298: This feels sketchy. Do we even support _strings_ in the parameters in the first place? I thought only stuff like %d (%x, %o) and %p are supported in parameters. I feel like this is a correctness issue. But then you would also have to scan the format string to "ban" the %s specifier which is more complicated.

tools/logger/convert.c
```diff
@@ -284,6 +287,16 @@ static void process_params(struct proc_ldc_entry *pe,
 			p += 2;
 		}
 	}
+
+	/* ensure no NULLs in format strings */
+	for (j = 0; j < i; j++) {
+		if (!pe->params[j]) {
+			pe->params[j] = (uintptr_t)missing;
```
jsarha, line 294: Mixing malloced pointers with static pointers is a bit scary, but I guess subst_mask is enough to make sure free is never called for "<missing>" .

jsarha (approved):
Looks Ok. Then again being scared about a null parameter for format string and dealing with risking heap corruption is counter productive, as printf deals gracefully with NULL and even if it did not, finding the error would be easy. In case of heap corruption its not. But this is what it is these static analysis errors.

tools/logger/convert.c
lgirdwood: strings not supported today and unlikely to be since this is all free with the Zephyr logger.

tools/logger/convert.c
kv2019i: @jsarha Just a note: the code is already doing that.

lgirdwood: ok, looks good then.

lgirdwood (approved):
Lets fix the potential NULL issue today, we can always come back and circle back on other updates if needed.

tools/logger/convert.c
kv2019i: @paulstelian97 we do support strings. Here's example print for params when subst_mask is set:

```
params[0] 0x9E19E370 (subst_mas clear)
params[1] 'dma-trace-task <2b972272-c5b1-4b7e-926f-0fc5cb4c4690>' (subst_mask set)
```

And matching log output:
```
[         131.770828] (         131.770828) c0 ll-schedule        ./schedule/ll_schedule.c:513  INFO task add 0x9e19e370 dma-trace-task <2b972272-c5b1-4b7e-926f-0fc5cb4c4690>
```

It's still sketchy. Let me see if I can do a simpler change.

kv2019i:
V5 update:
 - rebased and simplified implementation to abort on plain alloc fails and only handle the ldc entry lookup fail with explicit error handling

tools/logger/convert.c
```diff
@@ -268,7 +274,12 @@ static void process_params(struct proc_ldc_entry *pe,
 			/* %pQ format specifier */
 			/* substitute log entry address with formatted entry text */
 			pe->params[i] = (uintptr_t)asprintf_entry_text(raw_param);
-			pe->subst_mask |= 1 << i;
+
+			if (pe->params[i])
+				pe->subst_mask |= 1 << i;
+			else
+				pe->params[i] = (uintptr_t)missing;
+
```
kv2019i, line 282: This branch is now only used for one trace in SOF:

```
./trace/trace.c:139:                 _TRACE_INV_ID, _TRACE_INV_ID, "Suppressed %u similar messages: %pQ",
```

But still, for correctness, the entry may not be found.

tools/logger/convert.c
paulstelian97: I'm confused how strings can be supported because sof-logger uses dictionary logging and I'm not aware of any support for strings in there. Zephyr logging I think is not based on dictionary logging by default?

lgirdwood: @paulstelian97 I think Zephyr now has the string dictionary support, initially all the strings were baked into the Zephyr image (sorry, I don't know the Zephyr internals here). xtos config does not support ```%s``` strings as we cant do stuff like ```__func__``` in our trace today, only strings are in the trace body via the dictionary.

lgirdwood (approved):

tools/logger/convert.c
paulstelian97: Okay, I want to learn about the Zephyr logger and see if (after NXP migrates to fully Zephyr) we can use that. Right now we still have to use sof-logger due to still keeping the xtos builds (Zephyr isn't fully stable for us, there's a good few quirks left to fix)

kv2019i: @paulstelian97 @lgirdwood There's still code to handle "%s" but this is not used by any code in SOF FW and sof-logger emits a warning if one is seen. But we have two indirect usages of "%s", FW traces with "%pUx" and "%pQ" are converted to "%s" and the argument passed from FW (in raw log stream) is converter to a string that is passed as argument to (f)printf. And these steps can hit an alloc failure, or just missing conversion, leading to NULL being used. 

But right, "plain %s" usage where you'd have %s used directly in FW code, is not there.

PR #6828, opened 2022-12-16
west.yml: upgrade zephyr to af6d827b64

tmleman:
Zepych update: total of 73 commits.

Changes include adding power domains to DMA interfaces and allowing to skip context save during d3.

Dependencies:

- [x] https://github.com/thesofproject/sof/pull/6849

Signed-off-by: Tomasz Leman <tomasz.m.leman@intel.com>

PR #6849, opened 2022-12-19
audio: dai-zephyr/host-zephyr: stop dma only once

tmleman:
Until now, one dma channel could be stopped multiple times without any problems. From commit af6d827b64 on zephyr it is required that the start and stop calls be perfectly balanced.

Zephyr device power manager stores an internal counter. Counter is incremented if device is used and decremented on device release. In this case when we call dma_start() and dma_stop().

If dma_stop() is called more times than dma_start() power manager will reaport a error -EALREADY.

### DIFF
--- a/src/audio/dai-zephyr.c
+++ b/src/audio/dai-zephyr.c
@@ -1014,9 +1014,11 @@ static int dai_comp_trigger_internal(struct comp_dev *dev, int cmd)
 		/* only start the DAI if we are not XRUN handling */
 		if (dd->xrun == 0) {
 			/* recover valid start position */
-			ret = dma_stop(dd->chan->dma->z_dev, dd->chan->index);
-			if (ret < 0)
-				return ret;
+			if (dev->state == COMP_STATE_ACTIVE) {
+				ret = dma_stop(dd->chan->dma->z_dev, dd->chan->index);
+				if (ret < 0)
+					return ret;
+			}
 
 			/* dma_config needed after stop */
 			ret = dma_config(dd->chan->dma->z_dev, dd->chan->index, dd->z_config);
@@ -1054,12 +1056,23 @@ static int dai_comp_trigger_internal(struct comp_dev *dev, int cmd)
 		dai_trigger_op(dd->dai, cmd, dev->direction);
 #else
 		dai_trigger_op(dd->dai, cmd, dev->direction);
-		ret = dma_stop(dd->chan->dma->z_dev, dd->chan->index);
+		if (dev->state == COMP_STATE_ACTIVE) {
+			ret = dma_stop(dd->chan->dma->z_dev, dd->chan->index);
+		} else {
+			comp_warn(dev, "dma was stopped earlier");
+			ret = 0;
+		}
 #endif
 		break;
 	case COMP_TRIGGER_PAUSE:
 		comp_dbg(dev, "dai_comp_trigger_internal(), PAUSE");
-		ret = dma_suspend(dd->chan->dma->z_dev, dd->chan->index);
+		if (dev->state == COMP_STATE_ACTIVE) {
+			ret = dma_suspend(dd->chan->dma->z_dev, dd->chan->index);
+		} else {
+			comp_warn(dev, "dma was stopped earlier");
+			ret = 0;
+		}
+
 		dai_trigger_op(dd->dai, cmd, dev->direction);
 		break;
 	case COMP_TRIGGER_PRE_START:
--- a/src/audio/host-zephyr.c
+++ b/src/audio/host-zephyr.c
@@ -627,10 +627,13 @@ static int host_trigger(struct comp_dev *dev, int cmd)
 		break;
 	case COMP_TRIGGER_STOP:
 	case COMP_TRIGGER_XRUN:
-		ret = dma_stop(hd->chan->dma->z_dev, hd->chan->index);
-		if (ret < 0)
-			comp_err(dev, "host_trigger(): dma stop failed: %d",
-				 ret);
+		if (dev->state == COMP_STATE_ACTIVE) {
+			ret = dma_stop(hd->chan->dma->z_dev, hd->chan->index);
+			if (ret < 0)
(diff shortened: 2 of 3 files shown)


src/audio/dai-zephyr.c
```diff
@@ -1054,7 +1054,12 @@ static int dai_comp_trigger_internal(struct comp_dev *dev, int cmd)
 		dai_trigger_op(dd->dai, cmd, dev->direction);
 #else
 		dai_trigger_op(dd->dai, cmd, dev->direction);
-		ret = dma_stop(dd->chan->dma->z_dev, dd->chan->index);
```
ranj063, line 1057: the prefix in the commit title is wrong. This isn't in zephyr/dma, its actually audio:dai-zephyr/host-zephyr

src/audio/dai-zephyr.c
```diff
@@ -1054,7 +1054,12 @@ static int dai_comp_trigger_internal(struct comp_dev *dev, int cmd)
...
+		if (dev->state == COMP_STATE_ACTIVE) {
+			ret = dma_stop(dd->chan->dma->z_dev, dd->chan->index);
+		} else {
+			comp_warn(dev, "dma was stopped earlier");
```
ranj063, line 1060: @tmleman may I ask why we end up stopping the same DMA channel more than once anyway?

src/audio/host-zephyr.c
```diff
@@ -1048,7 +1051,8 @@ static int host_reset(struct comp_dev *dev)
 	comp_dbg(dev, "host_reset()");
 
 	if (hd->chan) {
-		dma_stop(hd->chan->dma->z_dev, hd->chan->index);
+		if (dev->state == COMP_STATE_ACTIVE)
```
ranj063, line 1054: is this check necessary? Is there a possibility of a reset while the comp is still active?

src/audio/dai-zephyr.c
```diff
@@ -1054,7 +1054,12 @@ static int dai_comp_trigger_internal(struct comp_dev *dev, int cmd)
 		dai_trigger_op(dd->dai, cmd, dev->direction);
 #else
 		dai_trigger_op(dd->dai, cmd, dev->direction);
-		ret = dma_stop(dd->chan->dma->z_dev, dd->chan->index);
+		if (dev->state == COMP_STATE_ACTIVE) {
+			ret = dma_stop(dd->chan->dma->z_dev, dd->chan->index);
```
kv2019i, line 1058: I guess we have to do this now. I'm a bit worried about coupling device state and DMA state this way. This opens a problem if some state is set to not ACTIVE in some other piece of code. Then dma_stop() is never called.  Probably this is ok now, but it's actually pretty hard to verify this is the case by reading current dai-zephyr.c code.

OTOH, it seems there are no calls to get dma channel status with the public API. Right @teburd @juimonen ? dma_get_status() does not seem fit to gate calls to dma_stop().

kv2019i (approved):
Thanks @tmleman ! I bit reluctant approval on this one. I think we have to make this change and I couldn't figure out a better way. Still, this seems to add brittleness that might bite us later. If anyone has better suggestions, please chime in.

src/audio/dai-zephyr.c
tmleman: For example in case of dai-zephyr, when we set PPL to pause it triggers COMP_TRIGGER_PAUSE. This leads to the function dma_suspend being called. Which is really just a call to function intel_adsp_hda_dma_stop.

```c
	.start = intel_adsp_hda_dma_start,
	.stop = intel_adsp_hda_dma_stop,
	.suspend = intel_adsp_hda_dma_stop,
	.get_status = intel_adsp_hda_dma_status,
```

Later when we do PPL reset we are triggering COMP_TRIGGER_STOP and that calls dma_stop again.

src/audio/host-zephyr.c
tmleman: I'm not sure. If not, the call to dma_stop should be removed from here because dma has been stopped during PPL pause.

src/audio/dai-zephyr.c
ranj063: Thanks for the clarification @tmleman

src/audio/host-zephyr.c
ranj063: I think we should leave the check here. I think for IPC4, it is not possible but maybe we need it for IPC3.

src/audio/dai-zephyr.c
abonislawski: This will work but the best way would be to balance this in zephyr driver, in gpdma we already have channel internal state so calling dma_stop several times will not result in unbalanced pm state.

```
	/* Validate the channel state */
	if (chan_data->state != DW_DMA_ACTIVE &&
	    chan_data->state != DW_DMA_SUSPENDED) {
		ret = -EINVAL;
		goto out;
	}
```
but there is no such internal state check in HDA zephyr driver so we need to take care of this in SOF (this PR) **or** track device state in Zephyr (HDA driver)

kv2019i: @abonislawski But as SOF code should work with any Zephyr driver, just fixing HDA is not enough, we'd need to have this as a guarantee at Zephyr dma.h level.

abonislawski: at dma.h level we can only expose current state

kv2019i:
@abonislawski Ok to merge this? We can't proceed with #6828 without this....

tmleman:
@kv2019i @abonislawski @ranj063 I had to make additional changes, I encourage you to take a look again.

Here you can check how it behave with updated zephyr kernel https://github.com/thesofproject/sof/pull/6853

abonislawski:
> @abonislawski Ok to merge this? We can't proceed with #6828 without this....

Im ok but please wait for CI

lgirdwood (approved):

aborisovich (approved):

PR #6851, opened 2022-12-19
Kw fix 278

jsarha:
Fix a static analysis issue, which is real in a sense. A search by ipc_get_comp_by_id() may fail if the id is not found. However, that is kind of unlikely since the comp with the exactly same id was created just couple of lines before.

### DIFF
--- a/src/ipc/ipc4/helper.c
+++ b/src/ipc/ipc4/helper.c
@@ -696,6 +696,11 @@ int ipc4_create_chain_dma(struct ipc *ipc, struct ipc4_chain_dma *cdma)
 	}
 
 	ipc_pipe = ipc_get_comp_by_id(ipc, pipeline_id);
+	if (!ipc_pipe) {
+		tr_err(&comp_tr, "ipc_pipe not found with pipeline_id %d",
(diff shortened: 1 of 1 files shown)


src/ipc/ipc4/helper.c
```diff
@@ -696,6 +696,11 @@ int ipc4_create_chain_dma(struct ipc *ipc, struct ipc4_chain_dma *cdma)
 	}
 
 	ipc_pipe = ipc_get_comp_by_id(ipc, pipeline_id);
+	if (!ipc_pipe) {
```
ranj063, line 699: @jsarha this was already discussed and qualified as a false positive. The pipeline is created just above on line 691 and if it failed, we would never get here

jsarha:
According to @ranj063, this issue is already handled (called false positive).

src/ipc/ipc4/helper.c
paulstelian97: Maybe an assert could be good here though. Full validation? Probably not, but an assert to make it clear that you know it's good should be there. This isn't hot code anyway.

PR #6858, opened 2022-12-20
Static code analysis

lyakh:
more fixes

### DIFF
--- a/src/audio/selector/selector_generic.c
+++ b/src/audio/selector/selector_generic.c
@@ -216,6 +216,8 @@ static void sel_s16le(struct processing_module *mod, struct input_stream_buffer 
 	int processed = 0;
 	int source_frame_bytes = audio_stream_frame_bytes(source);
 	int sink_frame_bytes = audio_stream_frame_bytes(sink);
+	int n_chan_source = MIN(SEL_SOURCE_CHANNELS_MAX, source->channels);
+	int n_chan_sink = MIN(SEL_SINK_CHANNELS_MAX, sink->channels);
 
 	while (processed < frames) {
 		n = frames - processed;
@@ -224,7 +226,7 @@ static void sel_s16le(struct processing_module *mod, struct input_stream_buffer 
 		nmax = audio_stream_bytes_without_wrap(sink, dest) / sink_frame_bytes;
 		n = MIN(n, nmax);
 		for (i = 0; i < n; i++) {
-			process_frame_s16le(dest, sink->channels, src, source->channels,
+			process_frame_s16le(dest, n_chan_sink, src, n_chan_source,
 					    &cd->coeffs_config);
 			src += source->channels;
 			dest += sink->channels;
@@ -285,6 +287,8 @@ static void sel_s32le(struct processing_module *mod, struct input_stream_buffer 
 	int processed = 0;
 	int source_frame_bytes = audio_stream_frame_bytes(source);
 	int sink_frame_bytes = audio_stream_frame_bytes(sink);
+	int n_chan_source = MIN(SEL_SOURCE_CHANNELS_MAX, source->channels);
+	int n_chan_sink = MIN(SEL_SINK_CHANNELS_MAX, sink->channels);
 
 	while (processed < frames) {
 		n = frames - processed;
@@ -293,7 +297,7 @@ static void sel_s32le(struct processing_module *mod, struct input_stream_buffer 
 		nmax = audio_stream_bytes_without_wrap(sink, dest) / sink_frame_bytes;
 		n = MIN(n, nmax);
 		for (i = 0; i < n; i++) {
-			process_frame_s32le(dest, sink->channels, src, source->channels,
+			process_frame_s32le(dest, n_chan_sink, src, n_chan_source,
 					    &cd->coeffs_config);
 			src += source->channels;
 			dest += sink->channels;
--- a/src/ipc/ipc4/handler.c
+++ b/src/ipc/ipc4/handler.c
@@ -465,7 +465,7 @@ static int ipc4_process_chain_dma(struct ipc4_message_request *ipc4)
 	bool delay = false;
 	int ret;
 
-	memcpy_s(&cdma, sizeof(cdma), ipc4, sizeof(cdma));
+	memcpy_s(&cdma, sizeof(cdma), ipc4, sizeof(*ipc4));
 
 	if (cdma.primary.r.allocate && cdma.extension.r.fifo_size) {
 		ret = ipc4_create_chain_dma(ipc, &cdma);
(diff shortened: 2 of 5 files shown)


src/audio/selector/selector_generic.c
```diff
@@ -254,6 +254,10 @@ static void process_frame_s32le(int32_t dst[], int dst_channels,
 	int64_t accum;
 	int i, j;
 
+	if (dst_channels > SEL_SINK_CHANNELS_MAX ||
+	    src_channels > SEL_SOURCE_CHANNELS_MAX)
+		return;
```
paulstelian97, line 259: I actually wonder if this is what should be done. I'd consider returning an error or clamping the value to the maximum allowed one here.

tools/logger/convert.c
```diff
@@ -1125,5 +1131,8 @@ int convert(struct convert_config *config)
 	ret = logger_read();
 out:
 	free(config->uids_dict);
+clean_up:
+	config->logs_header = NULL;
+	global_config = NULL;
```
paulstelian97, line 1136: Technically the second one doesn't point to the stack (so it's technically unneeded), but I'd say setting it to null is possibly fine here so I don't actually have an issue with this.

tools/logger/logger.c
```diff
@@ -444,6 +444,10 @@ int main(int argc, char *argv[])
 	if (config.input_std) {
 		config.in_fd = stdin;
 	} else if (baud) {
+		if (!config.in_file) {
+			ret = -EINVAL;
+			goto out;
+		}
```
paulstelian97, line 450: I've seen another change to this part of the file already (similar in spirit, even). Try not to duplicate work!

paulstelian97 (changes requested):
Blocking due to conflicting change with #6846

src/audio/selector/selector_generic.c
lgirdwood: @lyakh is this aligning with the other frame formats in selector or should this be done by the caller ?

lgirdwood (commented):
Looks good, but need to check selector.c for right place for src/dst count.

tools/logger/convert.c
marc-hb: global_config is not owned by this function so this looks wrong and possibly buggy. Did you test this? sof-logger is not used by Intel on this branch.

src/audio/selector/selector_generic.c
lyakh: this functions returns `void`

tools/logger/convert.c
lyakh: technically inside this function we don't know where the `config` parameter comes from, but in fact it is on stack in the caller's function. A better fix would be to get rid of those global variables and to pass `config` as a parameter to all functions.

src/audio/selector/selector_generic.c
lgirdwood: @lyakh I mean do it here in the caller. This way we can check all users.
```c
static int selector_process(struct processing_module *mod,
			    struct input_stream_buffer *input_buffers,
			    int num_input_buffers,
			    int num_output_buffers)
{
	struct comp_data *cd = module_get_private_data(mod);
	uint32_t avail_frames = input_buffers[0].size;

	comp_dbg(mod->dev, "selector_process()");

	if (!avail_frames)
		return PPL_STATUS_PATH_STOP;

	/* copy selected channels from in to out */
	cd->sel_func(mod, input_buffers, output_buffers, avail_frames);

	return 0;
}
```

tools/logger/convert.c
lyakh: @marc-hb this function is called only in one place. And after it completes `main()` completes too.

src/audio/selector/selector_generic.c
lgirdwood: Its a cheap check here regardless of single/multi channel and static analysis can be closed for this one.

tools/logger/convert.c
```diff
@@ -1052,19 +1052,21 @@ int convert(struct convert_config *config)
 	count = fread(&snd, sizeof(snd), 1, config->ldc_fd);
 	if (!count) {
 		log_err("Error while reading %s.\n", config->ldc_file);
-		return -ferror(config->ldc_fd);
+		ret = -ferror(config->ldc_fd);
```
kv2019i, line 1055: @lyakh I guess one alternative would be to make "struct snd_sof_Logs_header snd" a static global variable. Not much cleaner, but potentially less lines to change.

kv2019i (approved):
Good fixes @lyakh ! One alternative implementation proposal in inline comments, but no showstoppers. I think these can go in as they are.

marc-hb (changes requested):
> A better fix would be to get rid of those global variables and to pass config as a parameter to all functions.

That was more or less the case in the past. Then commit 5b29dae9c870fe said "enough spaghetti code, let's just use a single global config". But it stopped in the middle of the bridge and left this mess.

I'm trying now a simpler fix that finishes the job and makes a single, global `config` with a minimal number of lines changed.

Please resubmit this PR without this commit, this PR has too many unrelated commits anyway.

marc-hb:
> I'm trying now a simpler fix that finishes the job and makes a single, global config with a minimal number of lines changed.

Done:
- #6862

tools/logger/convert.c
lyakh: I'd much rather avoid adding more global variables. We already have a couple in the logger and I think we should remove them

src/ipc/ipc4/handler.c
```diff
@@ -465,7 +465,7 @@ static int ipc4_process_chain_dma(struct ipc4_message_request *ipc4)
 	bool delay = false;
 	int ret;
 
-	memcpy_s(&cdma, sizeof(cdma), ipc4, sizeof(cdma));
+	memcpy_s(&cdma, sizeof(cdma), ipc4, sizeof(*ipc4));
```
paulstelian97, line 468: This is an unchecked memcpy_s call which _may_ fail and I believe in case of failure there's an issue (the destination gets zeroed out, so you need to pay attention to that to avoid bugs like this). If this returns an error you should propagate or assert it doesn't happen (using the non-removable SOF asserts). Ditto on all other memcpy_s calls in this patch.

paulstelian97 (dismissed):

src/ipc/ipc4/handler.c
lyakh: this hasn't been changed by this commit. This commit doesn't increase chances of  `memcpy_s()` to fail

lgirdwood: Can we take the memcpy_s check as a separate patch/PR. This needs to be fixed first.

lgirdwood (approved):
@mwasko fyi - pls cherry-pick for mtl if needed.
@lyakh can you circle back and update the memcpy_s with checks in another PR.

mwasko:
> @mwasko fyi - pls cherry-pick for mtl if needed.

@marcinszkudlinski, @abonislawski FYI

PR #6860, opened 2022-12-20
topology2: add auxiliary data classes for ssp

juimonen:
Auxiliary data can be added at the end of ssp data blob to enable for example early clocks for certain platform. Add aux_config.conf that includes all possible aux data classes. These can be used inside hw_configs and will be embedded as tlv data after basic data blob.


tools/topology/topology2/include/dais/aux_config.conf
```diff
@@ -0,0 +1,280 @@
...
+	DefineAttribute."clock_source" {}
+
+	attributes {
+		!constructor [
+			"id"
+		]
+
+		unique	"name"
+	}
+}
```
lgirdwood, line 280: newline

lgirdwood (commented):
A lot of this look generic whilst some is obviously Intel specific - can this be split ?

tools/topology/topology2/include/dais/aux_config.conf
juimonen: should be fixed now

lgirdwood:
> > A lot of this look generic whilst some is obviously Intel specific - can this be split ?
> 
> so this _is_ very very intel specific... I think we did some of the stuff with ipc3 already, so I think some items are overlapping with old sof parameters. I don't know should this be moved to some intel directory then?

yes and lets call the file ```intel_something```

juimonen:
@lgirdwood so classes are now in include/dais/intel/intel_ssp_aux_data.conf
we had a intel dir in platform, but new class definitions didn't really seem to fit there (at least from my point of view).

lgirdwood (approved):

kv2019i (approved):

PR #6861, opened 2022-12-20
Audio: Mux: Convert mux and demux to module adapter

singalsu:
Please find the change descriptions from commits.

### DIFF
--- a/src/audio/mux/mux.c
+++ b/src/audio/mux/mux.c
@@ -409,12 +409,8 @@ static void mux_prepare_active_look_up(struct comp_data *cd,
 				       const struct audio_stream __sparse_cache **sources)
 {
 	const struct audio_stream __sparse_cache *source;
-	uint8_t active_elem;
-	uint8_t elem;
-
-	cd->active_lookup.num_elems = 0;
-
-	active_elem = 0;
+	int elem;
+	int active_elem = 0;
 
 	/* init pointers */
 	for (elem = 0; elem < cd->lookup[0].num_elems; elem++) {
@@ -422,16 +418,15 @@ static void mux_prepare_active_look_up(struct comp_data *cd,
 		if (!source)
 			continue;
 
-		if ((cd->lookup[0].copy_elem[elem].in_ch >
-		    (source->channels - 1)) ||
-		    (cd->lookup[0].copy_elem[elem].out_ch >
-		    (sink->channels - 1)))
+		if (cd->lookup[0].copy_elem[elem].in_ch >= source->channels ||
+		    cd->lookup[0].copy_elem[elem].out_ch >= sink->channels)
 			continue;
 
 		cd->active_lookup.copy_elem[active_elem] = cd->lookup[0].copy_elem[elem];
 		active_elem++;
-		cd->active_lookup.num_elems = active_elem;
 	}
+
+	cd->active_lookup.num_elems = active_elem;
 }
 
 static void demux_prepare_active_look_up(struct comp_data *cd,
@@ -439,26 +434,20 @@ static void demux_prepare_active_look_up(struct comp_data *cd,
 					 const struct audio_stream __sparse_cache *source,
 					 struct mux_look_up *look_up)
 {
-	uint8_t active_elem;
-	uint8_t elem;
-
-	cd->active_lookup.num_elems = 0;
-
-	active_elem = 0;
+	int elem;
+	int active_elem = 0;
 
 	/* init pointers */
 	for (elem = 0; elem < look_up->num_elems; elem++) {
-		if ((look_up->copy_elem[elem].in_ch >
-		    (source->channels - 1)) ||
-		    (look_up->copy_elem[elem].out_ch >
-		    (sink->channels - 1)))
+		if (look_up->copy_elem[elem].in_ch >= source->channels ||
+		    look_up->copy_elem[elem].out_ch >= sink->channels)
 			continue;
 
-		cd->active_lookup.copy_elem[active_elem] =
-			look_up->copy_elem[elem];
+		cd->active_lookup.copy_elem[active_elem] = look_up->copy_elem[elem];
 		active_elem++;
-		cd->active_lookup.num_elems = active_elem;
 	}
+
+	cd->active_lookup.num_elems = active_elem;
 }
 
 /* process and copy stream data from source to sink buffers */
--- a/src/audio/mux/mux.c
+++ b/src/audio/mux/mux.c
@@ -33,6 +33,7 @@
 static const struct comp_driver comp_mux;
 
 LOG_MODULE_REGISTER(muxdemux, CONFIG_SOF_LOG_LEVEL);
+
 #if CONFIG_IPC_MAJOR_3
 /* c607ff4d-9cb6-49dc-b678-7da3c63ea557 */
 DECLARE_SOF_RT_UUID("mux", mux_uuid, 0xc607ff4d, 0x9cb6, 0x49dc,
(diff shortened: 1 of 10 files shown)


src/audio/mux/mux.c
```diff
@@ -383,427 +349,186 @@ static void set_mux_params(struct comp_dev *dev,
...
 {
-	struct comp_data *cd = comp_get_drvdata(dev);
 	const struct audio_stream __sparse_cache *source;
-	uint8_t active_elem;
-	uint8_t elem;
-
-	cd->active_lookup.num_elems = 0;
-
-	active_elem = 0;
+	int elem;
```
singalsu, line 398: This and rest of of changes in this function should go into a code simplification commit, not related to module adapter convert.

src/audio/mux/mux.c
```diff
@@ -383,427 +349,186 @@ static void set_mux_params(struct comp_dev *dev,
...
 					 struct mux_look_up *look_up)
 {
-	struct comp_data *cd = comp_get_drvdata(dev);
-	uint8_t active_elem;
-	uint8_t elem;
-
-	cd->active_lookup.num_elems = 0;
-
-	active_elem = 0;
+	int elem;
```
singalsu, line 423: ditto

src/audio/mux/mux.c
```diff
@@ -234,53 +193,60 @@ static int build_config(struct comp_data *cd, struct comp_dev *dev)
...
 	int ret;
 
-	comp_cl_info(&comp_mux, "mux_new()");
+	comp_info(dev, "mux_init()");
 
-	dev = comp_alloc(drv, sizeof(*dev));
-	if (!dev)
-		return NULL;
-	dev->ipc_config = *config;
+	cd = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
```
singalsu, line 208: Seems to be wrong size for cd, since config is a pointer with model handler for binary blob.

singalsu: Fixed in latest push, actually needed to revert back to struct instead of pointer.

lgirdwood:
@singalsu ping, any ETA ?

singalsu:
> @singalsu ping, any ETA ?

I have problem with model handler in getting the demux blob. I need to find out if the problem is slim driver tests specific or generic with Linux driver too. Right now I'm working with SRC freeze/crash fix. I hope could get this PR fixed to pass CI and ready next week.

singalsu:
@lgirdwood It's confirmed the issue is in CI system/merge/build step. There is no way to work around the issue in FW code while using model handler. I can meanwhile clean up this PR and wait for the CI fix to be done.

src/include/sof/audio/module_adapter/module/generic.h
```diff
@@ -74,6 +75,47 @@ UT_STATIC void sys_comp_module_##adapter##_init(void) \
...
+		.get_large_config = module_get_large_config,\
+		.get_attribute = module_adapter_get_attribute,\
+	}, \
+}; \
+\
+static SHARED_DATA struct comp_driver_info comp_module_##adapter##_info = { \
+	.drv = &comp_##adapter##_module, \
+}; \
+\
+UT_STATIC void sys_comp_module_##adapter##_init(void); \
```
lyakh, line 109: this shouldn't be needed here: this macro should be used in C files, not in headers, typically close to the end of the file, so adding this declaration there won't help anyone

src/include/sof/audio/module_adapter/module/generic.h
```diff
@@ -74,6 +75,47 @@ UT_STATIC void sys_comp_module_##adapter##_init(void) \
 \
 DECLARE_MODULE(sys_comp_module_##adapter##_init)
 
+#define DECLARE_MODULE_ADAPTER_STEP1(adapter, uuid, tr) \
```
lyakh, line 78: sorry, could you explain why this is needed? Why the current `DECLARE_MODULE_ADAPTER()` isn't enough?

src/include/sof/audio/component.h
```diff
@@ -696,16 +696,25 @@ static inline struct comp_dev *comp_alloc(const struct comp_driver *drv,
 
 #if defined UNIT_TEST || defined __ZEPHYR__
 #define DECLARE_MODULE(init)
+#define DECLARE_TWO_MODULES(initfirst, initsecond)
```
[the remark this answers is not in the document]
singalsu: It's a testbench issue and explained in the commit message (There can't be two _module_init()s in a dynamic library). Maybe some variable arguments length and loop in C macro pre-processing to handle them could work but I don't know how to do such.

The demux/mux can't be run in current testbench without this. It's up to reviewers to decide how important it is. SOF as firmware works without this.

src/include/sof/audio/module_adapter/module/generic.h
```diff
@@ -66,6 +66,7 @@ static SHARED_DATA struct comp_driver_info comp_module_##adapter##_info = { \
 	.drv = &comp_##adapter##_module, \
 }; \
 \
+UT_STATIC void sys_comp_module_##adapter##_init(void); \
```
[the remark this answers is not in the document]
singalsu: Testbench usage with dynamic libraries for components needed the function prototype. Also this is not necessary if we don't run this in testbench. Personally I find testbench important with less hardware available for us. But also we can change testbench to static build and different component initialization, and wait. What do you think?

src/include/sof/audio/module_adapter/module/generic.h
singalsu: More components may need changes to be able to load to testbench. Remove this then?

src/include/sof/audio/module_adapter/module/generic.h
singalsu: It's described in commit message, did you notice it. I can't explain better.

src/include/sof/audio/module_adapter/module/generic.h
lyakh: I mean just this line - it's a function forward declaration. This is a part of a macro, the macro is used in .c files. So you end up having a function declaration and its definition that immediately follows it, so the declaration isn't needed there.

src/include/sof/audio/component.h
```diff
@@ -696,16 +696,25 @@ static inline struct comp_dev *comp_alloc(const struct comp_driver *drv,
...
 #elif CONFIG_LIBRARY
 /* In case of shared libs components are initialised in dlopen */
 #define DECLARE_MODULE(init) __attribute__((constructor)) \
 	static void _module_init(void) { init(); }
+#define DECLARE_TWO_MODULES(initfirst, initsecond) __attribute__((constructor)) \
+	static void _module_init(void) { initfirst(); initsecond(); }
```
lyakh, line 705: ah, this is the reason for this `DECLARE_TWO_MODULES()`! I see, sorry, I overlooked it. What if you had
```
#define DECLARE_MODULE(init) __attribute__((constructor)) \
	static void _module_init_##init(void) { init(); }
```
would it work then? You can have two functions with `__attribute__((constructor))` in a single file, cannot you?

src/include/sof/audio/module_adapter/module/generic.h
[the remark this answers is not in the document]
lyakh: ah, you mean the compiler is complaining that a global function has no prototype? In fact I'm not sure - how are this `DECLARE_MODULE()` functions called with testbench? They are executed, right? If not, then we can make them all `static` - you could try to just replace `UT_STATIC` here with `static` and see if it still works

src/include/sof/audio/component.h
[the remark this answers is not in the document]
lgirdwood: @singalsu I have a fix for this in my plugin branch, pls sync with me tomorrow.

src/include/sof/audio/component.h
singalsu: Yep, this is exactly the solution that I copied from Liam's plugin branch work. I'll push the new version next and make it non-draft. The CI test that previously failed passes now.

src/include/sof/audio/component.h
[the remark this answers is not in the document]
singalsu: I took your patch from plugin branch as such, it's the 3rd commit. Can you please check it's OK to include.

lgirdwood (approved):
Had a quick look, @ranj063 @lyakh @jsarha pls review for relevant areas of expertise.

lyakh (approved):

kv2019i (approved):
This is a big PR, but nothing popped up in the review and great to see cmocka support added for mux.

jsarha:
I should turbo charge my review routine. Just reached the end  (and found nothing to complain), only to find out the PR is merged.

PR #6862, opened 2022-12-20
logger: make "global_config" truly global

marc-hb:
Finish the job that commit 5b29dae9c870 ("logger: Create global convert_config variable to avoid spaghetti code.") started but did not finish, leaving behind a supposedly "global" variable that is actually a confusing global pointer to a struct local to the main() function.

This confuses some static analyzer complaining that stack values are being returned, see #6858 and #6738. This is a false positive because the main()'s stack lifespan is the same as a global but let's simplify things anyway.

Also stop using 'extern' in .c files, use a proper .h file instead.

### DIFF
--- a/tools/logger/convert.c
+++ b/tools/logger/convert.c
@@ -1032,7 +1032,7 @@ static int dump_ldc_info(void)
 
 int convert(void)
 {
-	struct snd_sof_logs_header snd;
+	struct snd_sof_logs_header * const logs_hdr = malloc(sizeof(*logs_hdr));
 	struct snd_sof_uids_header uids_hdr;
 	int count, ret = 0;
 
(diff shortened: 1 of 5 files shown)


marc-hb:
Intel CI does not test sof-logger anymore  but I have of course tested this locally.

kv2019i (approved):
Not sure the original move to use more global data is a good way to refactor code like this, but agreed, better to use real global object if the design is such.

tools/logger/convert.c
```diff
@@ -1032,55 +1032,58 @@ static int dump_ldc_info(void)
 
 int convert(void)
 {
-	struct snd_sof_logs_header snd;
+	struct snd_sof_logs_header * const logs_hdr = malloc(sizeof(*logs_hdr));
```
jsarha, line 1035: It would be nice to have a symmetric free()-call if something is malloced, or use a global if one does not bother to do that.

jsarha (approved):
I do not think going from a pointer passed to each function to global variable is an improvement, but that approach was taken before this PR. One nitpick about calling malloc() without symmetric free() call.

paulstelian97 (approved):

lgirdwood (approved):

marc-hb:
> Not sure the original move to use more global data is a good way to refactor code like this

> I do not think going from a pointer passed to each function to global variable is an improvement,

For the record I don't think 5b29dae9c870f was a good idea either but going back would require changing many more lines.

PR #6867, opened 2022-12-21
zephyr: ll: watchdog: ace: Add watchdog support in ll domain

softwarecki:
Added watchdog support to zephyrs ll scheduler domain. Enabled watchdog drivers for mtl platform.


src/schedule/zephyr_domain.c
```diff
@@ -55,6 +56,14 @@ struct zephyr_domain {
 /* perf measurement windows size 2^x */
 #define CYCLES_WINDOW_SIZE	10
 
+#if CONFIG_WDT_DW
```
abonislawski, line 59: just to make sure - this will be disabled with CONFIG_WATCHDOG=n?

src/schedule/zephyr_domain.c
```diff
@@ -67,6 +76,12 @@ static void zephyr_domain_thread_fn(void *p1, void *p2, void *p3)
 		/* immediately go to sleep, waiting to be woken up by the timer */
 		k_sem_take(&dt->sem, K_FOREVER);
 
+#if CONFIG_WDT_DW
+		// Feed the watchdog
```
abonislawski, line 80: stick to one comment style
/ *

src/schedule/zephyr_domain.c
```diff
@@ -92,6 +107,12 @@ static void zephyr_domain_thread_fn(void *p1, void *p2, void *p3)
 			runs = 0;
 			overruns = 0;
 		}
+
+#if CONFIG_WDT_DW
+		// Feed the watchdog
```
abonislawski, line 112: / *

src/schedule/zephyr_domain.c
softwarecki: yes, it will

src/schedule/zephyr_domain.c
```diff
@@ -55,6 +56,14 @@ struct zephyr_domain {
 /* perf measurement windows size 2^x */
 #define CYCLES_WINDOW_SIZE	10
 
+#if CONFIG_WDT_DW
+#define GET_DEVICE_LIST(node) DEVICE_DT_GET(node),
+
+const struct device *watchdog[CONFIG_MAX_CORE_COUNT] = {
```
lgirdwood, line 62: This needs to be in uncache section and aligned for cache line/size - can you confirm.

softwarecki: It is an array of pointers to the struct device. Its constant and don't think it need any special handling.

src/schedule/zephyr_domain.c
```diff
@@ -92,6 +107,12 @@ static void zephyr_domain_thread_fn(void *p1, void *p2, void *p3)
 			runs = 0;
 			overruns = 0;
 		}
+
+#if CONFIG_WDT_DW
+		/* Feed the watchdog */
+		if (watchdog[core])
+			wdt_feed(watchdog[core], 0);
+#endif
```
lyakh, line 115: I think usually just triggering once should be enough

src/schedule/zephyr_domain.c
lgirdwood: This depends on what other data is sharing the same cache line. If its a one time write, then making sure nothing else is on the same cache line is still needed

lyakh: @softwarecki as defined here it's an array of constant pointers, not a constant array. I.e. you cannot use pointers in this array to modify devices that it points to, but you can well modify contents of this array. The array itself isn't constant. If you want to make the array constant instead you need
```
static struct device * const watchdog[CONFIG_MAX_CORE_COUNT] = {
```
(btw, yes, you want this array `static`). If you want both - constant array of constant pointers you need
```
static const struct device * const watchdog[CONFIG_MAX_CORE_COUNT] = {
```
Now as is the array isn't constant, so it ends up in .bss, and as such it is uncached, it shares cache lines in the same section with other uncached objects, so cache is never used for them and can never go out of sync. If you make the array itself constant, it will go to .rodata, and that will be cached. But then it will be sharing cache lines with other constant data and as such that data is never modified, so again, no loss of synchronisation is possible. So, @lgirdwood I don't think we need any special handling here. If we deed, we'd need it in many other locations throughout the code and things would be broken right now.

src/schedule/zephyr_domain.c
```diff
@@ -168,6 +189,18 @@ static int zephyr_domain_register(struct ll_schedule_domain *domain,
...
 		k_timer_start(&zephyr_domain->timer, start, K_USEC(LL_TIMER_PERIOD_US));
+
+#if CONFIG_WDT_DW
+		if (watchdog[core]) {
+			/* Watchdog timeout configuration. */
+			const struct wdt_timeout_cfg watchdog_config = {
+				.window.max = 2 * LL_TIMER_PERIOD_US / 1000,
+			};
+
+			wdt_install_timeout(watchdog[core], &watchdog_config);
```
kv2019i, line 200: I don't see the wdt ever uninstalled? Wouldn't the watchdog fire when the ll-scheduler ends? In most cases, the system will go to power-off state, but I'd still expect to have symmetry here that if watchdog is installed for a thread, it's removed when the thread exits cleanly.

kv2019i (commented):
@softwarecki One question inline.

I'd also make the per-core array static.

src/schedule/zephyr_domain.c
lgirdwood: I'm fine if the data is all const, but my understanding here is that this is NOT the case ? and that ```CONFIG_MAX_CORE_COUNT``` means this data will be shared across cores (and we need to spell out this data should be in uncache and never directly accessed via cache (unless we have an access mechanism)). Can someone confirm usage ?

softwarecki:
This PR was replaced by https://github.com/thesofproject/sof/pull/6959

PR #6870, opened 2022-12-21
memcpy_s(): fix prototype and add IPC4 error checking

lyakh:


paulstelian97 (approved):
Try not to just add formatting changes in the middle of the patch, it makes review work harder than necessary. In this case it's little enough that it's fine, so approving for the actual fix.

paulstelian97 (approved):

lgirdwood (approved):
This also aligns the APIs more with POSIX

lgirdwood:
@marcinszkudlinski fyi - may want to cherry-pick for mtl

src/include/sof/debug/panic.h
```diff
@@ -35,14 +35,18 @@ void panic_dump(uint32_t p, struct sof_ipc_panic_info *panic_info,
 #define sof_panic(x) k_panic()
 
 #define assert(x) __ASSERT_NO_MSG(x)
+#if CONFIG_ASSERT
+#define assert_used(cond) assert(cond)
+#else
+#define assert_used(cond) (void)(cond)
```
kv2019i, line 41: Hmm, the rationale is missing bits. So is the intention to avoid compiler warnings when building without CONFIG_ASSERT? 

I'm not sure about the name. I don't recall "assert_used()" in widespread use and if we add our custom semantics, I'd rather add some "sof_" prefix to make it clear this is a SOF specific construct, not to be mixed with Zephyr and/or system runtime.

kv2019i: And why not use __ASSERT_EVAL() directly? Or if we need this in XTOS code, have assert_eval() to align with Zephyr naming? Probably more future-efficient to convert code to use native Zephyr macro and add macros for XTOS in the wrappers.

kv2019i (changes requested):
Not keen on the naming, see comments inline.

src/include/sof/debug/panic.h
lyakh: @kv2019i the rationale is both - avoiding compiler warnings and also actually executing the code. Currently we have a discrepancy: with XTOS the code under `assert()` is always evaluated, with Zephyr in the default configuration of `CONFIG_ASSERT=n` it isn't. So that instead of
```
int ret = fn();
assert(!ret);
```
one could just write
```
assert(!fn());
```
and get the same result. No, I don't like the name a lot either, I couldn't come up with a better one. The reason why I didn't use `__ASSERT_EVAL()` is very simple - I didn't know about it. I did try to search for an existing macro, but missed this one. And no wonder - it's only used once in Zephyr... I also find its semantics a bit cumbersome:
```
  __ASSERT_EVAL ((void) foo(),
  		 int x = foo(),
                 x != 0,
                 "foo() returned zero!");
```
so you write your actual execution code (in this case calling `foo()`) twice, making it look like `foo()` is called twice or something. With all that in mind, maybe we just have to redefine the actual `assert()` under Zephyr to always evaluate the argument like it currently does under XTOS?

kv2019i: @lyakh We spent effort not long ago to get rid of all side-effects from existing assert() code, so we could disable asserts from production builds. The standard assert is designed to be opt-in and code should not rely on it being evaluated and I think we should stick to standard definition and discourage use of assert for error handling.  

My personal experience is that the standard assert is ill fit for bigger projects, and the line between optional checks and error handling is too brittle in bigger projects, and errors crop up. In SOF, we already have too many asserts and they are in places that hurt performance.

If we want to use macros for error handling, I'd rather add something explicit like "BUG_ON()" which is used in Linux. Then there's not ambiguity on whether the check is present in all builds or not.
 
PS The __ASSERT_EVAL is indeed clumsy, so probably better to define something on SOF side.

src/ipc/ipc4/handler.c
```diff
@@ -463,9 +463,10 @@ static int ipc4_process_chain_dma(struct ipc4_message_request *ipc4)
 	struct ipc4_chain_dma cdma;
 	struct ipc *ipc = ipc_get();
 	bool delay = false;
-	int ret;
+	int ret = memcpy_s(&cdma, sizeof(cdma), ipc4, sizeof(*ipc4));
 
-	memcpy_s(&cdma, sizeof(cdma), ipc4, sizeof(*ipc4));
+	if (ret < 0)
+		return IPC4_FAILURE;
```
kv2019i, line 469: @lyakh 
- commit message: ""so it makes no sense to add gradeful run-time error handling there. Instead
add assert()s to protect against possible future human error.""
- actual patch: adds runtime error handling

Hmmm.

kv2019i (changes requested):
Hmm, puzzled by commit message, see inline.

src/ipc/ipc4/handler.c
lyakh: @kv2019i haha, the reader is invited to read between lines ;-) Corrected, thanks.

kv2019i (approved):

lgirdwood:
@lrudyX @wszypelt can you share the codecheck error, this PR is expected to improve the stability/security.

kv2019i:
Codecheck now clean, ready to go @lgirdwood @lyakh ?

lgirdwood (approved):

PR #6876, opened 2022-12-21
sparse: fix address space mismatches

lyakh:
Fix two new address space regressions


marc-hb:
sparse+TGL passed in https://github.com/thesofproject/sof/pull/6833. Are these files compiled only with MTL, not with TGL?

Just making sure we don't have yet another "green failure".

lyakh:
> sparse+TGL passed in #6833. Are these files compiled only with MTL, not with TGL?
> 
> Just making sure we don't have yet another "green failure".

@marc-hb yes, these files are only build with IPC4

lgirdwood (approved):

ranj063 (approved):

PR #6877, opened 2022-12-21
Revert "Audio: Copier: Add missing DMIC DAI index mapping to create_dai()"

serhiy-katsyuba-intel:
This reverts commit 8d166a432f68ccd9ee657a58f379ed4f0f22b888.

This restores working DMIC functionality on Windows platform broken by #6668.

**_ATTENTION: This will break DMIC1 (DMIC with DAI index 1) functionality on Linux platform, DMIC0 should still work fine!_** Non working DMIC1 should be fixed on Linux driver side, not this firmware.

### DIFF
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -378,7 +378,6 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
 
 		break;
 	case ipc4_dmic_link_input_class:
-		dai_index[dai_count - 1] = (dai_index[dai_count - 1] >> 5) & 0x7;
 		dai.type = SOF_DAI_INTEL_DMIC;
 		dai.is_config_blob = true;
 		type = ipc4_gtw_dmic;


juimonen:
I think the "fix" would be to extract the least significant 8 bits and use that as dai_index. The node_id can have other bits set in it from kernel side.

serhiy-katsyuba-intel:
> I think the "fix" would be to extract the least significant 8 bits and use that as dai_index. The node_id can have other bits set in it from kernel side.

dai_index comes from node_id.f.v_index, it's 8 bit:
dai_index[dai_count - 1] = node_id.f.v_index;

juimonen:
> 

yes correct. I'll make the corresponding kernel PR.

juimonen:
needs to be synced with https://github.com/thesofproject/sof/pull/6877. This is for windows compatibility.

lgirdwood:
@ujfalusi @plbossart fyi - how long has kernel had the wrong ID mapping ?
@serhiy-katsyuba-intel @juimonen the FW PR which introduced this was 2 days ago, so I dont think we have any legacy issues to block, this revert on FW side.

serhiy-katsyuba-intel:
> @ujfalusi @plbossart fyi - how long has kernel had the wrong ID mapping ?

Since March 2022. However, the problem affects only DMIC1. DMIC0 works fine with and without #6668. DMIC1 seems was not tested on Linux until recently.

juimonen:
@lgirdwood no I don't think so. I made https://github.com/thesofproject/linux/pull/4114, that should work with this revert. Should be  tested however.

ranj063:
> @lgirdwood no I don't think so. I made [thesofproject/linux#4114](https://github.com/thesofproject/linux/pull/4114), that should work with this revert. Should be tested however.

@libinyang Could you please test with this revert and the linux PR from @juimonen. Thanks!

ranj063 (approved):

singalsu (approved):

lgirdwood:
ok, @libinyang pls let us know if this PR and the kernel PR work for you.

tmleman (approved):

wszypelt (approved):

lgirdwood:
@libinyang ping - good to merge ? @mengdonglin fyi.

wszypelt:
@libinyang, @lgirdwood  good to merge?

lgirdwood:
@libinyang pls shout if there are any issues and we can address

PR #6783, opened 2022-12-13
topology2: add multiple capture support with DMIC in nocodec path

RanderWang:
It is a simple pipeline for AEC according to the requirement of chrome team. In future one capture stream will be for AEC and anther one will be sent back to user directly.
AEC pcm id is 27.  Table shared by Pierre

PCM0   playback	Jack Out
PCM1   capture 	Jack In
PCM2   playback	Speaker
PCM3   capture	Amplifier Reference (typically based on I/V feedback)
PCM4   capture 	Microphone
PCM5   playback	HDMI 1
PCM6   playback	HDMI 2
PCM7   playback	HDMI 3
PCM8   playback	HDMI 4
PCM9   RESERVED
PCM10  capture	DMIC
PCM11  capture	DMIC16kHz
PCM12  capture 	BufferedMic (aka KPB)
PCM13  playback Bluetooth
PCM14  capture 	Bluetooth
PCM15  RESERVED
PCM16  RESERVED
PCM17  RESERVED
PCM18  RESERVED
PCM19  RESERVED
PCM20  capture LowLatency
PCM21  capture RawPDM
PCM22  capture Ultrasonics
PCM23  capture RawPCM
PCM24  capture Voice  (Communication with a human)
PCM25  capture Speech (interaction with a machine)
PCM26  capture Context (aka Acoustic Context Analysis - ACA)
PCM27  capture Reference (e.g. for AEC. May be based on I/V feedback or on playback data)
PCM28  RESERVED
PCM29  RESERVED
PCM30  playback	LowLatency
PCM31  playback	DeepBuffer (for low-power)
PCM32  playback	Media (e.g. movies)
PCM33  playback	Voice (Communication with a human)
PCM34  playback	Ultrasonics

Issue #6657, opened 2022-11-23
[BUG][MTL-002] Recorded file from echo-reference dev has distortion and noise


yongzhi1:
**Describe the bug**
Run echo ref unit test with NOCODEC config, the captured file shows distortion and noise when examined on Audacity. 

**To Reproduce**
Start playback first:
```
aplay -Dhw:0,0 -c2 -r48000 -fS16_LE /home/sample.wav
```
Run capture and check the captured file on desktop:
```
arecord -Dhw:0,13 /home/loop.wav -c2 -r48000 -fs16_le -d10 
```

**Reproduction Rate**
100%

**Expected behavior**
The issue is not reproducible with closed src ACE FW.

**Impact**
What impact does this issue have on your progress (e.g., annoyance, showstopper)

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel: https://github.com/bardliao/linux/commits/echo-reference-nocodec
    * SOF: https://github.com/thesofproject/sof/tree/mtl-002-drop-stable
2) Name of the topology file
    * Topology: https://github.com/bardliao/sof/commits/test-echo-reference-nocodec-2

![sof-mtl-nocodec](https://user-images.githubusercontent.com/19287706/203659287-5d25be26-8ef2-45cc-b79f-b4545d7694e8.png)

3) Name of the platform(s) on which the bug is observed.
    * Platform: MTL-RVP

**Screenshots or console output**
[mtrace-echo-ref-mtl-002.txt](https://github.com/thesofproject/sof/files/10079534/mtrace-echo-ref-mtl-002.txt)
https://user-images.githubusercontent.com/19287706/203658196-f7212407-3407-497d-b893-ab293443ce5d.mp4

lgirdwood:
> Run echo ref unit test with NOCODEC config, the captured file shows distortion and noise when examined on Audacity.

@yongzhi1 can you take a screenshot of the audacity waveform and also attach a small raw PCM clip. Compressed audio cant be used to root cause distortion.

mengdonglin:
@yongzhi1 Which Platform and codec are you using in the test?

yongzhi1:
Hi @lgirdwood, here is the screenshot for the waveform, the lower track was captured with ACE, in both cases LR_48k.wav was used

[...]

PR #6601, opened 2022-11-15
topology2: add ssp multi stream capture 

bardliao:
Add a multi capture stream on SSP port. This will cause IO error with "~/sof-test/test-case/multiple-pause-resume.sh -r 50" test.
Creating this PR for people who want to reproduce the issue.

PR #6880, opened 2022-12-21
topology2: add reference capture for MAX98357A speaker amp

yongzhi1:
Use lbm_mode to loopback playback data for PCM27.


yongzhi1:
This is for MAX98357A speaker amp which does not have IV feedback capability, I am not sure about any side effect of lbm_mode, other than that any comments/suggestion?

bardliao:
@yongzhi1 Using LBM makes sense to me.

yongzhi1:
> @yongzhi1 Using LBM makes sense to me.

Ack, tested working with arecord based on sof-dev kernel.

lgirdwood:
> This is for MAX98357A speaker amp which does not have IV feedback capability, I am not sure about any side effect of lbm_mode, other than that any comments/suggestion?

Not sure why we need to fake IV feedback here with loop back ? Can you provide context.

Fwiw, The code looks fine so if not suitable for this use case could be used for testing.

plbossart:
In addition to @lgirdwood the open is "Why do we need an echo reference"?
This is only needed if there is a non-linear processing such as DRC or Waves or others on the speaker playback path, or if you wanted to use AEC in the DSP but that's not what this topology change does.
@yongzhi1 can you add a bit more context here?

yongzhi1:
> Not sure why we need to fake IV feedback here with loop back ? Can you provide context.

My understanding is that we did the same for max98360a, we now get the loopback data at I2S instead of MUXDEMUX:
sof-adl-max98360a-rt5682
https://github.com/thesofproject/sof/blob/main/tools/topology/topology1/sof-tgl-max98357a-rt5682.m4
https://github.com/thesofproject/linux/commit/f316c9d9ba8ea08d6994bc5ba8fa276eab186208

plbossart:
> > Not sure why we need to fake IV feedback here with loop back ? Can you provide context.
> 
> My understanding is that we did the same for max98360a, we now get the loopback data at I2S instead of MUXDEMUX as shown in sof-adl-max98360a-rt5682.tplg:
> 
> sof-adl-max98360a-rt5682 https://github.com/thesofproject/sof/blob/main/tools/topology/topology1/sof-tgl-max98357a-rt5682.m4 [thesofproject/linux@f316c9d](https://github.com/thesofproject/linux/commit/f316c9d9ba8ea08d6994bc5ba8fa276eab186208)

nope, this is using the loopback with the demux. The SSP0 is only used because we need a DPCM backend, this is the reason why we had issues with the updated kernel where the .dpcm_capture field was not set. https://github.com/thesofproject/linux/issues/4083

yongzhi1:
> nope, this is using the loopback with the demux. The SSP0 is only used because we need a DPCM backend, this is the reason why we had issues with the updated kernel where the .dpcm_capture field was not set. [thesofproject/linux#4083](https://github.com/thesofproject/linux/issues/4083)

Hi, @plbossart, yes, totally agree with above, think @bardliao also followed same approach to use copier-as-demux initially  https://github.com/thesofproject/sof/issues/6657.

yongzhi1:
> In addition to @lgirdwood the open is "Why do we need an echo reference"? This is only needed if there is a non-linear processing such as DRC or Waves or others on the speaker playback path, or if you wanted to use AEC in the DSP but that's not what this topology change does. @yongzhi1 can you add a bit more context here?

Hi, @plbossart , to use ACE in the DSP, I’d be tempted to follow https://github.com/thesofproject/sof/pull/6601, but it's a little over-kill to have "three copier.module for a simple demux functionality.".

Hi, @sathya-nujella , what's your opinion on above?

sathya-nujella:
> In addition to @lgirdwood the open is "Why do we need an echo reference"? This is only needed if there is a non-linear processing such as DRC or Waves or others on the speaker playback path, or if you wanted to use AEC in the DSP but that's not what this topology change does. @yongzhi1 can you add a bit more context here?

Yes. In playback path Post Processing/non-linear processing module would be enabled. Hence the requirement is to give this as  echo ref PCM to CRAS/user space.

bardliao:
@plbossart This PR is trying to get reference with loop back mode which should have the same result as getting reference from a demux like sof-adl-max98360a-rt5682.tplg does and will be simpler.

lgirdwood:
@bardliao @yongzhi1 is this PR for the "dummy" module testing only ? i.e. testing all the paths/PCMs ?

yongzhi1:
V2:

Added copier module in the pipeline to use as demux.

sof-mtl-max98357a-rt5682

lgirdwood:
> @bardliao @yongzhi1 is this PR for the "dummy" module testing only ? i.e. testing all the paths/PCMs ?

@yongzhi1 sorry, I'm not clear if we are looping back real smart amp feedback data here or fake (using the LBM) here.

yongzhi1:
> @yongzhi1 sorry, I'm not clear if we are looping back real smart amp feedback data here or fake (using the LBM) here.

Hi, @lgirdwood, MAX98357A is a dumb amp that does not support feedback data in PCM, so it's "fake" (not true IV).

lgirdwood:
> > @yongzhi1 sorry, I'm not clear if we are looping back real smart amp feedback data here or fake (using the LBM) here.
> 
> Hi, @lgirdwood, MAX98357A is a dumb amp that does not support feedback data in PCM, so it's "fake" (not true IV).

Thanks @yongzhi1, got you, that was confusing me. @plbossart is this the OK for you ?

tools/topology/topology2/cavs-rt5682.conf
```diff
@@ -107,10 +108,12 @@ Object.Dai {
 	SSP."1" {
 		id 		$SPK_ID
 		dai_index	1
-		direction	"playback"
+		direction	"duplex"
 		name		SSP1-Codec
 		default_hw_conf_id	0
 		sample_bits		32
+		# used for ref cap
+		quirks			"lbm_mode"
```
plbossart, line 116: why are we using LBM mode if there's a demux?

plbossart (commented):
not getting why there's both LBM and demux (with the copier).

tools/topology/topology2/cavs-rt5682.conf
yongzhi1: Hi, @plbossart , the demux is put on capture pipeline to connect with ACE module in the future.

plbossart: can you share the picture of the new topology, will be easier to see. Thanks.

yongzhi1: Hi, @plbossart 

Here is the picture with copier.module.8.2 added:
https://github.com/thesofproject/sof/pull/6880#issuecomment-1376031879

plbossart: Are you saying that copier.module.8.2 would have at some point additional sinks connected? I am struggling between what you want now and what will be done later.

yongzhi1: Hi, @plbossart , yes, that's my current thoughts, something like this (sorry for the ugly flow chart):
```mermaid
graph TD;
   SSP.IN-->copier.module.8.2;
   copier.module.8.2 --> echo-ref-PCM;
   copier.module.8.2 --> AEC;
    subgraph ide1 [future]

   PDM --> AEC;
   AEC --> DMIC48K;
end
```

tools/topology/topology2/cavs-rt5682.conf
```diff
@@ -107,10 +108,12 @@ Object.Dai {
 	SSP."1" {
 		id 		$SPK_ID
 		dai_index	1
-		direction	"playback"
+		direction	"duplex"
 		name		SSP1-Codec
 		default_hw_conf_id	0
 		sample_bits		32
+		# used for ref cap
```
ranj063, line 115: can we expand the comment to say reference capture please?

tools/topology/topology2/cavs-rt5682.conf
```diff
@@ -232,6 +235,38 @@ Object.Pipeline {
...
+				in_bit_depth		32
+				in_valid_bit_depth	32
+				out_bit_depth		32
+				out_valid_bit_depth	32
+				dma_buffer_size "$[$ibs * 2]"
+			}
+		}
+
+		Object.Widget.copier."2" {
+			stream_name	"SSP0-Codec"
```
ranj063, line 267: why do you need to set this?

tools/topology/topology2/cavs-rt5682.conf
```diff
@@ -262,6 +297,18 @@ Object.PCM {
 			formats	'S16_LE,S24_LE,S32_LE'
 		}
 	}
+
+	pcm."27" {
```
ranj063, line 301: pcm.2 please. Lets keep it incremental as it is really only an instance ID.

tools/topology/topology2/cavs-rt5682.conf
```diff
@@ -262,6 +297,18 @@ Object.PCM {
...
+		name	"echo-reference"
+		id	27
+		direction	"capture"
+		Object.Base.fe_dai."echo-reference" {}
```
ranj063, line 305: can we please wait till https://github.com/thesofproject/sof/pull/6986 merged and update this to use instance ID to instantiate?

tools/topology/topology2/cavs-rt5682.conf
```diff
@@ -262,6 +297,18 @@ Object.PCM {
...
+
+		Object.PCM.pcm_caps."capture" {
```
ranj063, line 307: same here

tools/topology/topology2/cavs-rt5682.conf
```diff
@@ -232,6 +235,38 @@ Object.Pipeline {
 			}
 		}
 	}
+
+	passthrough-capture.2 {
+		index	7
+
+		Object.Widget.copier.1 {
+			stream_name	'Echo reference'
```
ranj063, line 243: Can you make this string "Echo reference" a variable defined in the Define {} block so you can use it here and in the pcm caps object below?

tools/topology/topology2/cavs-rt5682.conf
```diff
@@ -289,4 +336,16 @@ Object.Base {
 		source	"copier.SSP.6.1"
 		sink	"copier.host.5.1"
 	}
+
+	route."6" {
+		source	"copier.SSP.8.1"
+		sink	"copier.module.8.2"
```
ranj063, line 342: this route can go into the pipeline definition passthrough-copier-be.conf

tools/topology/topology2/cavs-rt5682.conf
yongzhi1: Ack

tools/topology/topology2/cavs-rt5682.conf
yongzhi1: Here I just followed the example of similar place like:
https://github.com/thesofproject/sof/blob/main/tools/topology/topology2/cavs-nocodec.conf#L289
It's more out of convention than necessity.

tools/topology/topology2/cavs-rt5682.conf
yongzhi1: Here I followed https://github.com/thesofproject/sof/pull/6783#issue-1493569730

ranj063: @yongzhi1 the 27 is set below in the ID attribute. What I mean was the instance ID here doesnt need to match it

tools/topology/topology2/cavs-rt5682.conf
ranj063: lets not set it blindly if it is not needed,

tools/topology/topology2/cavs-rt5682.conf
yongzhi1: Ack

tools/topology/topology2/cavs-rt5682.conf
yongzhi1: Let me think about it, thanks for your review!!

tools/topology/topology2/cavs-rt5682.conf
yongzhi1: Sure

yongzhi1:
V3 update:
1. rebase the code.
2. add variables for  "Echo reference" and pipeline ID etc.
3. remove useless stream_name "SSP0-Codec".
4. echo ref related PCM, Dai instance & route index all start from 200.
5. add INCLUDE_ECHO_REF for conditional include.

tools/topology/topology2/include/pipelines/cavs/passthrough-copier-be.conf
```diff
@@ -0,0 +1,104 @@
...
+# }
+#
+# Where N is the unique pipeline ID within the same alsaconf node.
+#
+
+<include/common/audio_format.conf>
+<include/components/copier.conf>
+<include/components/pipeline.conf>
+
+Class.Pipeline."passthrough-copier-be" {
```
ranj063, line 23: can we call it dai-copier-buffer.conf? all copiers are capable of format conversion. So its not just passthrough anymore

tools/topology/topology2/platform/intel/i2s-dumb-amp-echo-ref.conf
```diff
@@ -0,0 +1,73 @@
+#
+# Pcm, Dai instance & route index all start from 200 to avoid collisions
```
ranj063, line 2: I have a PR submitted to alsautils https://github.com/alsa-project/alsa-utils/pull/188 to avoid such collisions in the alsatplg compiler instead of having to do this manually.

ranj063, line 2: Can we not use "dumb" in the name of the file please. Can you call this simply speaker-echo-ref.conf? What makes this exclusive to not-so-smart amps? When needed, the same file could be included in the case of smart amps too right?

tools/topology/topology2/platform/intel/i2s-dumb-amp-echo-ref.conf
```diff
@@ -0,0 +1,73 @@
...
+	}
+}
+
+Object.PCM {
+	pcm."200" {
+		name	"EchoRef"
+		id	27
+		direction	"capture"
+		Object.Base.fe_dai.1 {
+			name	"EchoRef"
```
ranj063, line 53: one feature of tplg2 that is not so widely known is when there's an object embedded within another and there's a shared attribute between them, like name in this case,  and they have the same value, there's no need to repeat it inside the child object. So you can just simple have Object.Base.fe_dai.1 {} and it will work fine.

tools/topology/topology2/platform/intel/i2s-dumb-amp-echo-ref.conf
```diff
@@ -0,0 +1,73 @@
...
+		}
+
+		Object.PCM.pcm_caps.1 {
+			direction	"capture"
```
ranj063, line 57: similarly for direction here

tools/topology/topology2/platform/intel/i2s-dumb-amp-echo-ref.conf
yongzhi1: Ack, it's a good learning.

tools/topology/topology2/platform/intel/i2s-dumb-amp-echo-ref.conf
yongzhi1: Ack, will update in V4.

tools/topology/topology2/platform/intel/i2s-dumb-amp-echo-ref.conf
yongzhi1: I agree the file name is rather dumb, for amps with IV data, we do not need the lbm_mode. Will update the file name.

tools/topology/topology2/include/pipelines/cavs/passthrough-copier-be.conf
yongzhi1: Ack,

tools/topology/topology2/cavs-rt5682.conf
```diff
@@ -113,7 +125,7 @@ Object.Dai {
 			tdm_slot_width	32
 		}
 	}
-	SSP."1" {
+	SSP."200" {
```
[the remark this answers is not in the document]
yongzhi1: Ack, the 200 serves as a warning that this instance was referenced elsewhere, do not change one place without changing another .. . will use incremental values for this and other places (route index etc.)

yongzhi1:
V5 update:
1. rename 2 new files
2. remove redundant settings (i.e. name, direction) in child object.
3. use incremental values for PCM instance id at top file and included file speaker-echo-ref.conf.

tools/topology/topology2/include/pipelines/cavs/dai-copier-be.conf
```diff
@@ -0,0 +1,104 @@
...
+				dma_buffer_size		"$[$obs * 2]"
+			}
+		}
+
+		copier."2" {
+			type	buffer
+			copier_type	"module"
+			period_sink_count	2
+			period_source_count	0
+			num_audio_formats	2
```
ranj063, line 79: there's only 1 audio format

tools/topology/topology2/include/pipelines/cavs/dai-copier-be.conf
```diff
@@ -0,0 +1,104 @@
...
+				out_valid_bit_depth	24
+				dma_buffer_size		"$[$obs * 2]"
+			}
+		}
+
+		copier."2" {
+			type	buffer
+			copier_type	"module"
+			period_sink_count	2
+			period_source_count	0
```
ranj063, line 78: are these period counts really useful?

tools/topology/topology2/platform/intel/speaker-echo-ref.conf
```diff
@@ -0,0 +1,66 @@
+Define {
+	ECHO_REF_STREAM_NAME	"Echo reference"
+}
+
+Object.Dai {
+	# override amp instance's i2s config for echo reference
+	SSP."1" {
```
ranj063, line 7: @yongzhi1 lets not do this. My alsauilts PR will not allow object overrides. You need to set the direction and lbm mode to duplex is INCLUDE_ECHO_REF is true in the top-level file.

ranj063 (approved):

tools/topology/topology2/include/pipelines/cavs/dai-copier-be.conf
yongzhi1: Ack, cut & paste error :)
I saw if was fixed by https://github.com/thesofproject/sof/commit/48afcefc756edbbd5596f3a74f1106b028f8832d

tools/topology/topology2/include/pipelines/cavs/dai-copier-be.conf
yongzhi1: I must admit I do not know.

tools/topology/topology2/platform/intel/speaker-echo-ref.conf
yongzhi1: The lbm_mode is the only valid value for SSP quirks, I will think about your suggestion.
`
			!valid_values [
				"lbm_mode"
			]
`

ranj063: @yongzhi1 what you need is something like this. You will need the latest upstream alsa-utils for this to work.
```
diff --git a/tools/topology/topology2/cavs-rt5682.conf b/tools/topology/topology2/cavs-rt5682.conf
index b04957d99..e13bc362c 100644
--- a/tools/topology/topology2/cavs-rt5682.conf
+++ b/tools/topology/topology2/cavs-rt5682.conf
@@ -116,7 +116,15 @@ Object.Dai {
        SSP."1" {
                id              $SPK_ID
                dai_index       $SPEAKER_SSP_DAI_INDEX
+IncludeByKey.INCLUDE_ECHO_REF {
+"false" {
                direction       "playback"
+       }
+"true" {
+               direction       "duplex"
+               quirks          "lbm_mode"
+       }
+}
                name            $SPEAKER_CODEC_NAME
                default_hw_conf_id      0
                sample_bits             32
```

yongzhi1: Ack, fortunately I waited a while before attempting the changes, the new capability arrived to rescue at the right moment: 
https://github.com/alsa-project/alsa-utils/commit/1350900246e0462b584331be0df24e177e7b25be

tools/topology/topology2/include/pipelines/cavs/dai-copier-be.conf
ranj063: lets remove it. im pretty sure these are not used in ipc4

tools/topology/topology2/platform/intel/speaker-echo-ref.conf
```diff
@@ -0,0 +1,58 @@
...
+				out_bit_depth		32
+				out_valid_bit_depth	32
+				dma_buffer_size "$[$ibs * 2]"
+			}
+		}
+	}
+}
+
+Object.PCM {
+	pcm."2" {
```
ranj063, line 36: BTW @yongzhi1 if you want to give it a try. you can start your object instances from 1 for the pcm and routes in this conf file with this PR https://github.com/alsa-project/alsa-utils/pull/188. The compiler will take care of avoiding clashes with the top-level file objects

yongzhi1:
Pushed for SOFCI TEST.

ranj063:
SOFCI TEST

ranj063 (approved):

lgirdwood (approved):

tools/topology/topology2/platform/intel/speaker-echo-ref.conf
lgirdwood: Lets do this as part of the topology2 feature cleanup. @jsarha fyi.

bardliao (approved):

kv2019i:
@yongzhi1 @ranj063 @lgirdwood I think we have a problem. This doesn't build and for some reason, pr-tools-build is not running topology2 check. FYI @marc-hb @greg-intel @fredoh9 @keqiaozhang

juimonen:
so we need very latest alsa-utils for this, where ever it needs to be updated...

kv2019i:
Yes, the tplg2 coverage seems ok after all, so this was a false alarm. One does need to have a recent alsa-utils to have the build work, but the CI checks seem good with regard to this.

ranj063:
> @yongzhi1 @ranj063 @lgirdwood I think we have a problem. This doesn't build and for some reason, pr-tools-build is not running topology2 check. FYI @marc-hb @greg-intel @fredoh9 @keqiaozhang

@kv2019i the CI docer has been updated with the latest alsa-utils. The build did pass right?

marc-hb:
Yes this build ONLY if you update your docker image maintained by @fredoh9 

```
docker pull thesofproject/sof 
./scripts/docker-run.sh ./scripts/build-tools.sh
```

marc-hb:
> or some reason, pr-tools-build is not running topology2 check.

URL?

PR #6762, opened 2022-12-09
V2.2 mux

lyakh:
This is a back-port of #6131 with 3 more missing backported commits, it should fix topologies with a mux in capture paths

PR #6502, opened 2022-10-28
[TEST ONLY] mux: (debug) remove all unused code

lyakh:
mux-demux is only used in its demux function in capture streams, disable the rest of the code since it isn't known to work.

PR #6882, opened 2022-12-22
mux: consider sinks on same direction only for prepare/reset propagation

johnylin76:
The change of mux component codes from PR#6762 added the condition on capture to propagate prepare/reset command to upstream only if there are no active sinks. The purpose is to support multiple endpoints sourced from one capture device.

However, the regression is observed with audio malfunction on DSP for some specific brya devices. After triage, we found that the regression only affects devices applying smart_amp on DSP.  

For topologies integrated with smart_amp component (as figured below), a demux is placed on echo reference (capture) pipeline, providing sink linkages to both PCMC and smart_amp. smart_amp is located on playback pipeline whose direction is opposite from demux.

```
 PCMP ----> smart_amp ----> SSP(A)  # playback ppl
             ^
             |
 PCMC <---- demux <----- SSP(B)     # echo reference ppl
```

In practice, the playback pipeline including smart_amp can be set to active or stopped independently.
While triggering prepare or reset to echo referene pipeline, the smart_amp (one of the sink component for demux) may be in active state, which causes the command propagation to upstream components is blocked. This is not the expected behavior (prepare/reset should be propagated across demux through the echo reference pipeline regardless of which state smart_amp is.

In other words, the condition for mux prepare/reset propagation on capture should only regard sinks on the same
direction (capture), skipping ones on playback.

### DIFF
--- a/src/audio/mux/mux.c
+++ b/src/audio/mux/mux.c
@@ -629,6 +629,11 @@ static int mux_reset(struct comp_dev *dev)
 	comp_info(dev, "mux_reset()");
 
 	if (dir == SOF_IPC_STREAM_PLAYBACK) {
+		/*
+		 * On playback path, stop propagating reset signal to downstream if any of the
+		 * source is active. The condition is only applicable to comp MUX (multi-source to
+		 * single-sink).
+		 */
 		list_for_item(blist, &dev->bsource_list) {
 			struct comp_buffer *source = container_of(blist, struct comp_buffer,
 								  sink_list);
@@ -643,16 +648,33 @@ static int mux_reset(struct comp_dev *dev)
 				return PPL_STATUS_PATH_STOP;
 		}
 	} else {
+		/*
+		 * On capture path, stop propagating reset signal to upstream if any of the sink is
+		 * active. The condition is only applicable to comp DEMUX (single-source to
+		 * multi-sink).
+		 *
+		 * One exception is that the active sink of DEMUX is on playback path, e.g. DEMUX
+		 * on Echo Reference path (capture) linked to the sink SMART_AMP on Speakers path
+		 * (playback). In practice, the playback sink state should not affect the reset
+		 * process through the capture path DEMUX locates on.
+		 */
 		list_for_item(blist, &dev->bsink_list) {
 			struct comp_buffer *sink = container_of(blist, struct comp_buffer,
 								source_list);
 			struct comp_buffer __sparse_cache *sink_c = buffer_acquire(sink);
 			int state = -1;
+			int direction = -1;
 
-			if (sink_c->sink)
+			if (sink_c->sink) {
 				state = sink_c->sink->state;
+				direction = sink_c->sink->direction;
+			}
 
 			buffer_release(sink_c);
+
+			/* no need to regard the playback sink state */
+			if (direction == SOF_IPC_STREAM_PLAYBACK)
+				continue;
 
 			if (state > COMP_STATE_READY)
 				/* should not reset the upstream components */
@@ -719,11 +741,18 @@ static int mux_prepare(struct comp_dev *dev)
 								source_list);
 			struct comp_buffer __sparse_cache *sink_c = buffer_acquire(sink);
 			int state = -1;
+			int direction = -1;
 
-			if (sink_c->sink)
+			if (sink_c->sink) {
 				state = sink_c->sink->state;
+				direction = sink_c->sink->direction;
+			}
 
 			buffer_release(sink_c);
+
+			/* no need to regard the playback sink state */
+			if (direction == SOF_IPC_STREAM_PLAYBACK)
+				continue;
 
 			/* only prepare upstream if we have no active sinks */
 			if (state == COMP_STATE_PAUSED || state == COMP_STATE_ACTIVE)


johnylin76:
@lyakh @plbossart PTAL, thanks

cujomalainey:
Should this be going to main?

src/audio/mux/mux.c
```diff
@@ -634,9 +634,14 @@ static int mux_reset(struct comp_dev *dev)
 								  sink_list);
 			struct comp_buffer __sparse_cache *source_c = buffer_acquire(source);
 			int state = source_c->source->state;
+			int direction = source_c->source->direction;
 
 			buffer_release(source_c);
 
+			/* the state is independent of loopback-typed sources */
```
plbossart, line 641: I don't understand what that repeated comment means.

In  addition, the playback case was NOT modified by PR #6762 in this function.

src/audio/mux/mux.c
```diff
@@ -719,12 +731,19 @@ static int mux_prepare(struct comp_dev *dev)
...
+			if (sink_c->sink) {
 				state = sink_c->sink->state;
+				direction = sink_c->sink->direction;
+			}
 
 			buffer_release(sink_c);
 
+			/* the state is independent of loopback-typed sinks */
+			if (direction == SOF_IPC_STREAM_PLAYBACK)
+				continue;
```
plbossart, line 745: And in this case you only deal loopbacks with the capture path

plbossart (changes requested):
The commit message is beyond convoluted, comments are very hard to understand, and some of the changes not related to PR#6762 so it's very hard to review. The last sentence takes the cake "the state machine for demux should be independent of the reversed sink."  What is a reversed sink?

The other problem is that there's clearly a test coverage issue, I don't know how we would verify if all the cases are actually covered. Keep in mind the multi-capture and multi-playback cases did NOT include a requirement for loopback. One would assume that the validation would include first an iso-functionality recheck with more PCMs added, before getting into advanced usages. It's not very wise IMHO to shoot straight for more advanced cases without first checking the basic cases.

plbossart:
> Should this be going to main?

'main' does not support XTOS/IPC3 for Intel any longer, so this has to be on a separate branch.

That said, this sort of changes is absolutely NOT suitable for a 'stable' branch, the branch management clearly makes no sense at this point if 'stable' adds new capabilities and new topologies that were not discussed before.

johnylin76:
I just rephrased the commit message and also revised the change. The issue is only from demux on capture path, so we should not touch the code for playback path cases.

johnylin76:
> The other problem is that there's clearly a test coverage issue, I don't know how we would verify if all the cases are actually covered. Keep in mind the multi-capture and multi-playback cases did NOT include a requirement for loopback. One would assume that the validation would include first an iso-functionality recheck with more PCMs added, before getting into advanced usages. It's not very wise IMHO to shoot straight for more advanced cases without first checking the basic cases.

Agreed. We have the similar expectation that only supporting uni-directional multiple endpoints (multi-playback or multi-capture). We have identified the regression is from smart_amp-integrated topology, which has a demux on capture (echo reference pipeline). It's the existing  use cases on xtos production builds. So we should still maintain the behavior as it is on developing multiple endpoints feature support. We can help on monitoring the regression on xtos production builds if exists.

src/audio/mux/mux.c
johnylin76: Just revised that. And we have further identified that the issue can only hit by demux on capture path, which has the sink component located on the playback path (e.g. echo reference for smart amp).

src/audio/mux/mux.c
```diff
@@ -648,12 +648,22 @@ static int mux_reset(struct comp_dev *dev)
 								source_list);
 			struct comp_buffer __sparse_cache *sink_c = buffer_acquire(sink);
 			int state = -1;
+			int direction = -1;
 
-			if (sink_c->sink)
+			if (sink_c->sink) {
 				state = sink_c->sink->state;
+				direction = sink_c->sink->direction;
+			}
```
lyakh, line 656: `mux_reset()` is currently also shared between mux and demux, whereas the echo reference functionality is only relevant for the demux, right? Would it maybe make sense to separate `mux_reset()` and `demux_reset()` and only modify the latter for this case? See https://github.com/thesofproject/sof/pull/6502/commits/02968c64c1e2dda1c369d844a118e0c81b8d9ccb for an example attempt to similarly untangle `mux_trigger()`

johnylin76: done

lgirdwood:
@lyakh good now ?

src/audio/mux/mux.c
```diff
@@ -660,10 +660,63 @@ static int mux_reset(struct comp_dev *dev)
...
+static int demux_reset(struct comp_dev *dev)
+{
+	int dir = dev->pipeline->source_comp->direction;
+	struct list_item *blist;
+	struct comp_data *cd = comp_get_drvdata(dev);
+
+	comp_info(dev, "demux_reset()");
+
+	if (dir == SOF_IPC_STREAM_PLAYBACK) {
+		list_for_item(blist, &dev->bsource_list) {
```
lyakh, line 678: demux has only one source, doesn't it? And in fact I don't think we do demux for playback as well as mux for capture, so we could just return errors here and for capture in `mux_reset()` above. Otherwise this introduces some significant code duplication. As an alternative, if you (understandably) don't want to introduce changes to a legacy development branch, we could add static functions with this common code to be called from both `mux_reset()` and `demux_reset()`

lyakh:
> @lyakh good now ?

@lgirdwood hm, now we have code duplication. We can use this, provided concerns from @plbossart are also addressed, or we could try to reduce the code duplication. One big simplification would be if we could make it explicit, that we only support mux-ing on playback pipelines and demux-ing in capture. But none of these considerations are critical.

src/audio/mux/mux.c
johnylin76: > demux has only one source, doesn't it? And in fact I don't think we do demux for playback as well as mux for capture, so we could just return errors here and for capture in `mux_reset()` above. Otherwise this introduces some significant code duplication. As an alternative, if you (understandably) don't want to introduce changes to a legacy development branch, we could add static functions with this common code to be called from both `mux_reset()` and `demux_reset()`

If so I would prefer the change of the previous version (before the reset splitting) for code duplication considering, which is:
```
--- a/src/audio/mux/mux.c
+++ b/src/audio/mux/mux.c
@@ -648,12 +647,22 @@ static int mux_reset(struct comp_dev *dev)
                                                                source_list);
                        struct comp_buffer __sparse_cache *sink_c = buffer_acquire(sink);
                        int state = -1;
+                       int direction = -1;
 
-                       if (sink_c->sink)
+                       if (sink_c->sink) {
                                state = sink_c->sink->state;
+                               direction = sink_c->sink->direction;
+                       }
 
                        buffer_release(sink_c);
 
+                       /*
+                        * the sink component is on the opposite direction (PLAYBACK for this case),
+                        * which should be irrelevant to the propagation for reset across the demux.
+                        */
+                       if (direction == SOF_IPC_STREAM_PLAYBACK)
+                               continue;
+
                        if (state > COMP_STATE_READY)
                                /* should not reset the upstream components */
                                return PPL_STATUS_PATH_STOP;
```

because the original code already specified the applicable cases implicitly.
The PPL_STATUS_PATH_STOP checkers under the [if-else clause](https://github.com/thesofproject/sof/blob/rpl-001-drop-stable/src/audio/mux/mux.c#L631) in mux_reset() handle reset propagation in respective cases:
- if dir==PLAYBACK (checking if any active source exists) 
  - for mux (N source to 1 sink): it's possible to meet the checking condition.
  - for demux (1 source to N sink): no way to meet since the propagation flow is from source
- else (dir==CAPTURE; checking if any active sink exists)
  - for mux (N source to 1 sink): no way to meet since the propagation flow is from sink
  - for demux (1 source to N sink): it's possible to meet
 
That is, the original code has the intention for only "mux on playback" and "demux on capture". We don't really need to separate to mux_reset and demux_reset. (I do think adding more comments for explanation would be nice because it is so implicit.)

@lyakh @plbossart thoughts?

lyakh: > I do think adding more comments for explanation would be nice because it is so implicit.

Ok, well, yes, at least your comment above "...to the propagation for reset across the demux" should make it clearer why it only mentions the demux. So, yes, maybe it's better for now to use this smaller change and consider for the future a cleanup, making the limitation of "mux for playback and demux for capture" explicit

johnylin76: Updated the commit with more detailed comments.

src/audio/mux/mux.c
```diff
@@ -629,6 +629,11 @@ static int mux_reset(struct comp_dev *dev)
 	comp_info(dev, "mux_reset()");
 
 	if (dir == SOF_IPC_STREAM_PLAYBACK) {
+		/*
+		 * On playback path, stop propagating reset signal to downstream if any of the
+		 * source is active. The condition is only applicable to comp MUX (multi-source to
+		 * single-sink).
+		 */
```
lyakh, line 636: There's no explicit check for the mux / demux type here. We just implicitly use the fact, that in case of a demux we would be coming into reset from the upstream source direction, so the single source is guaranteed to already be inactive? So the below condition `if (state > COMP_STATE_READY)` can only trigger for a mux?

src/audio/mux/mux.c
```diff
@@ -643,17 +648,34 @@ static int mux_reset(struct comp_dev *dev)
...
+		/*
+		 * On capture path, stop propagating reset signal to upstream if any of the sink is
+		 * active. The condition is only applicable to comp DEMUX (single-source to
+		 * multi-sink).
+		 *
+		 * One exception is that the active sink of DEMUX is on playback path, e.g. DEMUX
+		 * on Echo Reference path (capture) linked to the sink SMART_AMP on Speakers path
+		 * (playback). In practice, the playback sink state should not affect the reset
+		 * process through the capture path DEMUX locates on.
+		 */
```
lyakh, line 660: yes, clearer now, thanks

src/audio/mux/mux.c
johnylin76: AFAIK reset trigger starts from PCM and propagates downstream/upstream for playback/capture. It implies that for DEMUX on playback, reset can be received only when its (single) source got reset. Same concept for MUX on capture.

However I'm not sure if the trigger/propagation rule will be changed in the future. For example of MUX on capture, would it be possible to reset a PCM-less pipeline which presents as a branch of MUX upstream in multiple endpoints scheme? (like PPL#3 in the following graph)
```
PCMC --- PPL#1 ---> MUX --- PPL#2 ---> DAI#2
                       \--> PPL#3 ---> DAI#3

johnylin76:
Is it ok to merge now?

lyakh (approved):

src/audio/mux/mux.c
lgirdwood: @johnylin76 any comment here ? am I wrong in my assumptions here ?

johnylin76:
@mwasko could you help on merging it? Thanks

PR #6886, opened 2022-12-23
xtensa-build-zephyr.py: convert sof-info/ text files from CRLF to LF

marc-hb:
Building on Windows generates CRLF text files. Standardize checksums to
LF files.

As usual, files in the build directory are left untouched.


marc-hb:
One model was not available in https://sof-ci.01.org/sofpr/PR6886/build3181/devicetest/index.html, everything else passed. This PR does not affect hardware testing in any way.

lgirdwood (approved):

lgirdwood:
Good find, this should align the checksums.

PR #6887, opened 2022-12-23
hda: Hw hda chain xrun handling

makarukp:
This is draft which proposes xrun handling in chain dma. Please do not merge.

To enable that feature below PRs has to be merged:
https://github.com/thesofproject/sof/pull/6656
https://github.com/zephyrproject-rtos/zephyr/pull/53327

### DIFF
--- a/src/audio/chain_dma.c
+++ b/src/audio/chain_dma.c
@@ -24,6 +24,11 @@
 #include <sof/ut.h>
 #include <zephyr/pm/policy.h>
 #include <rtos/init.h>
+#if CONFIG_IPC4_XRUN_NOTIFICATIONS_ENABLE
+#include <ipc4/notification.h>
+#include <sof/ipc/msg.h>
+#include <ipc/header.h>
+#endif
 
 static const struct comp_driver comp_chain_dma;
 static const uint32_t max_chain_number = DAI_NUM_HDA_OUT + DAI_NUM_HDA_IN;
@@ -47,6 +52,10 @@ struct chain_dma_data {
 	enum sof_ipc_stream_direction stream_direction;
 	/* container size in bytes */
 	uint8_t cs;
+#if CONFIG_IPC4_XRUN_NOTIFICATIONS_ENABLE
+	bool xrun_notification_sent;
+	struct ipc_msg *msg_xrun;
+#endif
 
 	/* local host DMA config */
 	struct dma *dma_host;
@@ -131,6 +140,34 @@ static size_t chain_get_transferred_data_size(const uint32_t out_read_pos, const
 	return buff_size - in_read_pos + out_read_pos;
 }
 
+#if CONFIG_IPC4_XRUN_NOTIFICATIONS_ENABLE
+static void handle_xrun(struct chain_dma_data *cd)
+{
+	int ret;
+
+	if (cd->link_connector_node_id.f.dma_type == ipc4_hda_link_output_class &&
+	    !cd->xrun_notification_sent) {
+		tr_warn(&chain_dma_tr, "handle_xrun(): underrun detected");
+		xrun_notif_msg_init(cd->msg_xrun, cd->link_connector_node_id.dw,
+				    SOF_IPC4_GATEWAY_UNDERRUN_DETECTED);
+		ipc_msg_send(cd->msg_xrun, NULL, true);
+		cd->xrun_notification_sent = true;
+	} else if (cd->link_connector_node_id.f.dma_type == ipc4_hda_link_input_class &&
+		   !cd->xrun_notification_sent) {
+		tr_warn(&chain_dma_tr, "handle_xrun(): overrun detected");
+		xrun_notif_msg_init(cd->msg_xrun, cd->link_connector_node_id.dw,
+				    SOF_IPC4_GATEWAY_OVERRUN_DETECTED);
+		ipc_msg_send(cd->msg_xrun, NULL, true);
+		cd->xrun_notification_sent = true;
+	} else {
+		/* if xrun_notification_sent is already set, then it means that link was
+		 * able to reach stability therefore next underrun/overrun should be reported.
+		 */
+		cd->xrun_notification_sent = false;
+	}
+}
+#endif
+
 static enum task_state chain_task_run(void *data)
 {
 	size_t link_avail_bytes, link_free_bytes, host_avail_bytes, host_free_bytes;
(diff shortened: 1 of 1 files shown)


lgirdwood:
@makarukp fyi - some conflicts so CI wont run.

src/audio/chain_dma.c
```diff
@@ -47,6 +52,10 @@ struct chain_dma_data {
 	enum sof_ipc_stream_direction stream_direction;
 	/* container size in bytes */
 	uint8_t cs;
+#if CONFIG_IPC_MAJOR_4
+	bool under_over_run_notification_sent;
```
lyakh, line 56: don't we usually call "onder-over-runs" just "xruns?"

src/audio/chain_dma.c
```diff
@@ -224,6 +236,45 @@ static size_t chain_get_transferred_data_size(const uint32_t out_read_pos, const
...
+{
+	struct ipc4_resource_event_data_notification *notif_data = msg_xrun->tx_data;
+
+	notif_data->resource_id = resource_id;
+	notif_data->event_type = event_type;
+	notif_data->resource_type = GATEWAY;
+	notif_data->dws[0] = 0;
+	notif_data->dws[1] = 0;
+	notif_data->dws[2] = 0;
+	notif_data->dws[3] = 0;
```
lyakh, line 250: `dws[]` has 6 elements, you set 4 of them to 0 here. Is the intention of this to make it deterministic? Do you need to zero out all of them? Use `memset()`?

src/audio/chain_dma.c
```diff
@@ -243,6 +294,9 @@ static enum task_state chain_task_run(void *data)
 	case -EPIPE:
 		tr_warn(&chain_dma_tr, "chain_task_run(): dma_get_status() link xrun occurred,"
 			" ret = %u", ret);
+#if CONFIG_IPC_MAJOR_4
```
lyakh, line 297: is chain DMA actually available with IPC3? Has it been tested with it? It seems IPC4-only to me, in fact we'd want a dependency somewhere in Kconfig I guess

src/include/ipc4/notification.h
```diff
@@ -42,6 +42,17 @@ enum sof_ipc4_notification_type {
...
+	SOF_IPC4_GATEWAY_UNDERRUN_DETECTED	= 6,
+	SOF_IPC4_GATEWAY_OVERRUN_DETECTED	= 7,
+};
+
+/* Resource Type - source of the event */
+enum sof_ipc4_resource_type {
+	/* Gateway */
+	GATEWAY = 2,
+};
+
```
kv2019i, line 55: I guess these are generic definitions and could be a separate patch that adds the gateway resource event types to the common header. And then a second chain-dma specific patch that uses these flags for chain-dma.

src/audio/chain_dma.c
```diff
@@ -224,6 +236,45 @@ static size_t chain_get_transferred_data_size(const uint32_t out_read_pos, const
...
+
+static void handle_xrun(struct chain_dma_data *cd)
+{
+	if ((cd->link_connector_node_id.f.dma_type == ipc4_hda_link_output_class)
+	    && (!cd->under_over_run_notification_sent)) {
+		tr_warn(&chain_dma_tr, "handle_xrun(): underrun detected");
+		xrun_notif_msg_init(cd->msg_xrun, cd->link_connector_node_id.dw,
+				    SOF_IPC4_GATEWAY_UNDERRUN_DETECTED);
+		ipc_msg_send(cd->msg_xrun, NULL, true);
+		cd->under_over_run_notification_sent = true;
```
kv2019i, line 261: This probably should be good to put behind a Kconfig. At least with current Linux SOF driver, I don't see what we could do with these messages. Is driver supposed to react to these, or are these solely for debugging?

kv2019i (changes requested):
A few minor notes. I think a bigger one is what is the purpose of this. We've not send xrun messages before for pipelines, so why send specifically for chain-dma...? Still a potentially good new tool, but something that needs clarification.

src/audio/chain_dma.c
makarukp: We should treat it as a useful information for debugging. Xrun notification message should be possible to be masked by driver (even in production fw), however we do not have this functionality right now in converge. In such case I will create new define in Kconfig for blocking xrun notifications.

src/audio/chain_dma.c
```diff
@@ -589,6 +643,7 @@ static struct comp_dev *chain_task_create(const struct comp_driver *drv,
 	const uint32_t host_dma_id = cdma->primary.r.host_dma_id;
 	const uint32_t link_dma_id = cdma->primary.r.link_dma_id;
 	const uint32_t fifo_size = cdma->extension.r.fifo_size;
+	union ipc4_notification_header header;
```
mbukowsk, line 646: Unused variable if CONFIG_IPC_MAJOR_4 is not defined

makarukp:
To finalize that PR extra dependencies required:
#7023 
#7024

gkbldcig:

makarukp:
Please do not close, work in progress...

src/audio/chain_dma.c
softwarecki: I suggest move this function to some file under ipc/ipc4

makarukp: @softwarecki I agree that 'xrun_notif_msg_init' could be used as generic function and it would be reasonable to move it to ipc/ipc4, however, I am not sure if 'handle_xrun' is good candidate as it is specific handling dedicated for chain dma interface.

makarukp:
Requires #7164

src/include/ipc4/notification.h
makarukp: @kv2019i Done in #7023

src/audio/chain_dma.c
```diff
@@ -131,6 +140,34 @@ static size_t chain_get_transferred_data_size(const uint32_t out_read_pos, const
 	return buff_size - in_read_pos + out_read_pos;
 }
 
+#if CONFIG_XRUN_NOTIFICATIONS_ENABLE
```
lgirdwood, line 143: This is an IPC4 specific code block, should the Kconfig not be ```CONFIG_IPC4_XRUN_NOTIFICATIONS_ENABLE```

kv2019i (approved):
Thanks, looks good now!

softwarecki (approved):

src/audio/chain_dma.c
```diff
@@ -131,6 +140,34 @@ static size_t chain_get_transferred_data_size(const uint32_t out_read_pos, const
 	return buff_size - in_read_pos + out_read_pos;
 }
 
+#if CONFIG_IPC4_XRUN_NOTIFICATIONS_ENABLE
+static void handle_xrun(struct chain_dma_data *cd)
+{
+	int ret;
```
dnikodem, line 146: unnecessary variable?

dnikodem (approved):

PR #6893, opened 2022-12-26
sof-ctl: add support for ipc4 mode

libinyang:
This PR adds the sof-ctl support for ipc4 mode. User can run below command to generate the init instance data blob for ipc4:
```
# ./tools/build_tools/ctl/sof-ctl -I 'IPC4' -T 0 -g $size -b
```

### DIFF
--- a/src/include/kernel/abi.h
+++ b/src/include/kernel/abi.h
@@ -62,4 +62,7 @@
 /** \brief SOF ABI magic number "SOF\0". */
 #define SOF_ABI_MAGIC		0x00464F53
 
+/* SOF IPC4 ABI magic number "SOF4". */
+#define SOF_IPC4_ABI_MAGIC	0x34464F53
+
 #endif /* __KERNEL_ABI_H__ */
--- a/src/include/kernel/header.h
+++ b/src/include/kernel/header.h
@@ -36,4 +36,26 @@ struct sof_abi_hdr {
 	uint32_t data[0];	/**< Component data - opaque to core */
 } __attribute__((packed));
 
+/**
+ * struct sof_ipc4_abi_hdr - Used by any bespoke component data structures or binary blobs.
+ * @magic: The magic number for the header. The value is 'S', 'O', 'F', '4'
+ * @size: The size in bytes of the data, excluding this struct
+ * @abi: SOF ABI version
+ * @blob_type: Type of blob: INIT_INSTANCE, CONFIG_SET or LARGE_CONFIG_SET
+ *             The value is of type enum sof_ipc4_module_type
+ * @param_id: ID indicating which parameter to update with the new data. The validity of
+ *            the param_id with blob_type is dependent on the module implementation.
+ * @reserved: Reserved for future use
+ * @data: Component data - opaque to core
+ */
+struct sof_ipc4_abi_hdr {
+	uint32_t magic;
+	uint32_t size;
+	uint32_t abi;
+	uint32_t blob_type;
+	uint32_t param_id;
+	uint32_t reserved[3];
+	uint32_t data[];
+}  __packed;
+
 #endif /* __KERNEL_HEADER_H__ */
--- a/tools/ctl/ctl.c
+++ b/tools/ctl/ctl.c
@@ -41,6 +41,13 @@ struct ctl_data {
 	bool no_abi;
 	/* component specific type, default 0 */
 	uint32_t type;
+	/* ipc version, 0: IPC3 or 1: IPC4, default 0 for IPC3 */
+	uint8_t ipc_version;
+	/* ipc message cmd type, used in IPC4 message: */
+	/*  0: init_instance, 2: set_config, 4 set_largeconfig */
+	uint8_t blob_type;
+	/* ipc message param_id value in extension field, used in IPC4 message */
+	uint32_t param_id;
 	/* set or get control value */
 	bool set;
 	/* print ABI header */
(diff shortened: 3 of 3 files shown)


src/include/kernel/abi.h
```diff
@@ -62,4 +62,7 @@
 /** \brief SOF ABI magic number "SOF\0". */
 #define SOF_ABI_MAGIC		0x00464F53
 
+/* SOF IPC4 ABI magic number "SOF4". */
+#define SOF_IPC4_ABI_MAGIC	0x34464F53
```
lgirdwood, line 66: Need to watch this as there may be some users who will ```printf(%s)``` this an rely on the NULL termination. Can you check for any code that prints abi.

src/include/kernel/header.h
```diff
@@ -36,4 +36,26 @@ struct sof_abi_hdr {
...
+ * @size: The size in bytes of the data, excluding this struct
+ * @abi: SOF ABI version
+ * @blob_type: Type of blob: INIT_INSTANCE, CONFIG_SET or LARGE_CONFIG_SET
+ *             The value is of type enum sof_ipc4_module_type
+ * @param_id: ID indicating which parameter to update with the new data. The validity of
+ *            the param_id with blob_type is dependent on the module implementation.
+ * @reserved: Reserved for future use
+ * @data: Component data - opaque to core
+ */
+struct sof_ipc4_abi_hdr {
```
lgirdwood, line 51: this looks very similar to ipc3 version, can we use the same structure and use magic to determine where to treat things differently

kv2019i: Ack on that. It seems "type" can be used differently based on "abi" value, and the first reserved 32bit word can be used as param_id (ignored when abi is IPC3 or older).

tools/ctl/ctl.c
```diff
@@ -41,6 +41,13 @@ struct ctl_data {
 	bool no_abi;
 	/* component specific type, default 0 */
 	uint32_t type;
```
kv2019i, line 43: So this could have different semantics for IPC4.

tools/ctl/ctl.c
```diff
@@ -41,6 +41,13 @@ struct ctl_data {
...
+	/* ipc version, 0: IPC3 or 1: IPC4, default 0 for IPC3 */
+	uint8_t ipc_version;
```
kv2019i, line 45: Why not use "abi.major" directly. When we add IPC5, it's awkward to have this mapping (IPC5 would be 2).

kv2019i (changes requested):
Some comments inline.

src/include/kernel/abi.h
libinyang: @lgirdwood Thanks for reminder. I will have a check.

src/include/kernel/header.h
libinyang: We removed ```type``` in the struct sof_abi_hdr. So the member offset is different between sof_abi_hdr and sof_ipc4_abi_hdr. In this case, how can we use the same structure to refer the member, for example, ```size```?
sof_abi_hdr->size offset is 8 while sof_ipc4_abi_hdr->size offset is 4.

tools/ctl/ctl.c
libinyang: ```ipc_version``` is an input from user. And ABI version determined by ipc_version. So without ipc_version being set, we don't know the ABI version.

tools/ctl/ctl.c
```diff
@@ -571,12 +695,21 @@ int main(int argc, char *argv[])
...
+		if (ctl_data->ipc_version == 4)
+			ctl_data->ctrl_size = sizeof(struct sof_ipc4_abi_hdr);
+		else /* only ipc3 and ipc4 is supported */
+			ctl_data->ctrl_size = sizeof(struct sof_abi_hdr);
 		buffer_alloc(ctl_data);
 		header_init(ctl_data);
-		hdr = (struct sof_abi_hdr *)
-			&ctl_data->buffer[BUFFER_ABI_OFFSET];
-		hdr->size = ctl_data->print_abi_size;
+		if (ctl_data->ipc_version == 4) {
```
cujomalainey, line 704: can we use either a switch of ops callback. That will prevent partial implementations if for some reason we move to another version in the future

tools/ctl/ctl.c
```diff
@@ -551,6 +643,38 @@ int main(int argc, char *argv[])
 			ctl_data->print_abi_header = true;
 			ctl_data->print_abi_size = atoi(optarg);
 			break;
+		case 'I':
+		{
+			char *ipcv = optarg;
+
+			if (!strcmp(ipcv, "IPC3")) {
```
cujomalainey, line 650: why not just "3"?

cujomalainey (changes requested):

src/include/kernel/header.h
lgirdwood: @libinyang ```type``` should be unused in ipc4 and ```blob_type``` and ```param_id``` unused in ipc3.
@ujfalusi is the kernel using this new ABI IPC4 struct today ?

lgirdwood: @libinyang can you align this structure in the kernel before we send it upstream, this needs to be the same structure as IPC3 albeit with different magic and use of reserved fields for IPC4 usage. @plbossart fyi.

ujfalusi: > @ujfalusi is the kernel using this new ABI IPC4 struct today ?

Yes, it is merged via this PR: https://github.com/thesofproject/linux/pull/3918
Unfortunately lots of comments and discussion gone now, especially the ones where I was suggesting to extend/change  the existing sof_abi_hdr struct for IPC4 support.

src/include/kernel/header.h
```diff
@@ -36,4 +36,26 @@ struct sof_abi_hdr {
...
+	uint32_t magic;
+	uint32_t size;
+	uint32_t abi;
+	uint32_t blob_type;
+	uint32_t param_id;
```
singalsu, line 56: Param_id seems to be an uint8_t in current Peter's patch https://github.com/thesofproject/linux/pull/4132

ujfalusi: @singalsu, I will have an updated PR as soon as I get consensus on the kernel side.
The param_id is 8bit in IPC messages, this is why I have decided to not use the whole u32 for it on the kernel side.

gkbldcig:

lgirdwood:
@ujfalusi where are we with alignment with kernel and reuse of struct for IPC3 ?

ujfalusi:
> @ujfalusi where are we with alignment with kernel and reuse of struct for IPC3 ?

Yes, there is: https://github.com/thesofproject/linux/blob/54494c4753c8406ff49183dd08014f667f31c920/include/uapi/sound/sof/header.h#L14-L38

I'm will have a replacement PR, unless @libinyang wants to do the honors and rework this PR.

ujfalusi:
@lgirdwood, tentative header update: https://github.com/ujfalusi/sof/commit/c199c1e66b4015cf076bf3927831a06428e1a318 looks like this atm.

Issue #6896, opened 2022-12-30
[BUG][Zephyr] assert hit in sof-test run with dmic or hda DAIs


kv2019i:
**Describe the bug**
When CONFIG_ASSERT=y is enabled in build (e.g. with https://github.com/thesofproject/sof/pull/6530), assert is hit in multiple test cases with sof-test and on multiple platforms.

**To Reproduce**
Apply #6530 , run sof-test.

**Reproduction Rate**
100%

**Expected behavior**
Asserts should not be hit!

**Impact**
Unable to enable CONFIG_ASSERT in CI by default due to many failures.

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel: e3707e4dd275bc731f6ed4067a2694534675ca52 (sof-dev of today, "ASoC: Intel: soc-acpi: add configuration for variant of 0C11 product")
    * SOF: 14474447d41e2dfd3879a16fe2fa6f3dee041509 ("topology2: cavs-nocodec add wov support")
2) Name of the topology file
    * Topology: sof-hda-generic.tplg (but based on SOF Ci results affects also many other topologies)
3) Name of the platform(s) on which the bug is observed.
    * Platform: Intel TGL, ADL, MTL

**Screenshots or console output**

PR #6897, opened 2022-12-30
dai-zephyr: handle NULL case when calling dai_config_get()

kv2019i:
As per interface documentation of dai_config_get(), the function may return NULL if the device is not configured for the requested direction. Handle this case explicitly.

Currently no upstream Zephyr driver returns NULL, but e.g. dmic driver returns invalid configuration when dai_config_get() is called with DAI_DIR_TX as direction.


lyakh (approved):

marcinszkudlinski (approved):

aiChaoSONG (approved):

lgirdwood (approved):
@wszypelt can you check the internal check. Thanks !

Issue #6864, opened 2022-12-21
[BUG] ipc timed out for 0x13000004|0x1 GLB_SET_PIPELINE_STATE on IPC4 NOCODEC and SDW platforms


keqiaozhang:
**Describe the bug**
We observed this issue in daily test, it seems that this issue only happens on IPC4 NOCODEC and SDW platforms, we didn't observe this issue on HDA platform. DSP dead after ipc timed out, so there're no useful error logs in mtrace. 
The reproduction rate is about 20%. but I'm not sure if it's a regression or not, will do more tests to confirm it.

**dmesg**
```
[ 5182.381164] kernel: snd_sof:sof_ipc4_trigger_pipelines: sof-audio-pci-intel-tgl 0000:00:1f.3: trigger cmd: 4 state: 4
[ 5182.381751] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx reply: 0x33000000|0x1: GLB_SET_PIPELINE_STATE
[ 5182.381849] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx done : 0x13000003|0x1: GLB_SET_PIPELINE_STATE [data size: 12]
[ 5182.381895] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx      : 0x13000004|0x1: GLB_SET_PIPELINE_STATE [data size: 12]
[ 5182.382154] kernel: snd_sof_intel_hda_common:hda_dsp_stream_trigger: sof-audio-pci-intel-tgl 0000:00:1f.3: FW Poll Status: reg[0x160]=0x20140000 successful
[ 5182.382214] kernel: snd_sof:sof_pcm_trigger: sof-audio-pci-intel-tgl 0000:00:1f.3: pcm: trigger stream 0 dir 0 cmd 4
[ 5182.382238] kernel: snd_sof_intel_hda_common:hda_dsp_stream_trigger: sof-audio-pci-intel-tgl 0000:00:1f.3: FW Poll Status: reg[0x160]=0x2014001e successful
[ 5182.382244] kernel: snd_sof:sof_ipc4_trigger_pipelines: sof-audio-pci-intel-tgl 0000:00:1f.3: trigger cmd: 4 state: 4
[...]

PR #6900, opened 2023-01-02
mixin-mixout: handle congestion corner case without assert

kv2019i:
If mixout is not able to process data, the "assert(sinks_free_frames >= start_frame)" condition is hit.

This is not correct usage of asserts as the condition may be hit due to performance reasons (and is frequently hit in pause-resume stress test with sof-test), and production builds will have asserts disabled, leading to undefined behaviour.

Instead of assert, add a check, emit a warning and truncate the amount of copied frames, to ensure deterministic behaviour in all conditions.

### DIFF
--- a/src/audio/mixin_mixout/mixin_mixout.c
+++ b/src/audio/mixin_mixout/mixin_mixout.c
@@ -423,7 +423,6 @@ static int mixin_process(struct processing_module *mod,
 		 * performance problems with processing data on time in mixout.
 		 */
 		start_frame = mixout_data->pending_frames[source_index];
-		assert(sinks_free_frames >= start_frame);
 
 		sink_c = buffer_acquire(sink);
 


src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -439,7 +439,11 @@ static int mixin_copy(struct comp_dev *dev)
 		 * performance problems with processing data on time in mixout.
 		 */
 		start_frame = mixout_data->pending_frames[source_index];
-		assert(sinks_free_frames >= start_frame);
+		if (start_frame > frames_to_copy) {
+			comp_warn(dev, "mixin copy: no space, need %u, space for %u",
+				  start_frame, frames_to_copy);
+			frames_to_copy = start_frame;
+		}
```
lyakh, line 446: Do I understand it correctly, that previous the `assert()` was checking, that in the current sink the number of frames still pending from previous copies wasn't "too large," specifically that it wasn't larger than the number of free frames over all sinks? Which is similar to saying, that no sink is more than half full (assuming they all have the same size buffers)? Now you replace that check with a different one, namely checking that the number of still pending frames isn't larger than what we want to copy now? Which might actually be even easier to hit - if your last transfer is small and you still have a few pending frames from the previous run.

lgirdwood (commented):
@kv2019i do we need to pass anything up the stack for host to take corrective action ?

src/audio/mixin_mixout/mixin_mixout.c
kv2019i: @lyakh  Correct. "frames_to_copy" already taken into account as "sinks_free_frames = MIN(sinks_free_frames, free_frames - pending_frames);" is done earlier in the function. 

So old check assumes "start_frame" cannot ever be larger than sinks_free_frames. This is not correct and you either have undefined behaviour (this currently happens) or assert is hit (when we enable CONFIG_ASSERT=y).

My modified code takes account both  how much we have leftovers in source, and how much we have space in output.

kv2019i:
@lgirdwood wrote:
> @kv2019i do we need to pass anything up the stack for host to take corrective action ?

Not sure how to do this as the module really doesn't know why data flow is abnormal. The assert is mostly hit in pause-resume tests with lot of stream state transitions, so these are likely cases where the stream is anyways terminated. At least no failures are seen with this patch. At least this is better than running the process() with invalid parameters (start_frames > space in sinks).

@ranj063 is working on the https://github.com/thesofproject/sof/pull/6836 , but does not seem to touch this part specifically.

lgirdwood (approved):

lgirdwood:
@lyakh good for you ?
@kv2019i do we need any other PRs to be merged before this will pass CI or is the failure unexpected ?

kv2019i:
@lgirdwood I need to spend more time and respond to @lyakh 's concern. Let's not merge before that.

gkbldcig:

kv2019i:
V2 (Feb 3rd):
- modified patch based on review feedback
- it turns out the assert condition is in fact incorrect and can simply be removed

src/audio/mixin_mixout/mixin_mixout.c
kv2019i: @lyakh Boy, this took a long time. But thanks, you are right, my fix was not correct. Looking at this again, it seems the problem is with the assert and it's actually not needed. If "start_frame == 10" and "frames_to_copy == 5", the "sinks_free_frames" can still be "3", "5" or "10", or are valid values. The "start_frame" is specific to a single mixout buffer, while the other variables are calculcated as a minimum across all mixout buffers. So it's perfectly possible that a single mixout buffer has more already mixed data, than what can be written in this cycle by mixin. 

So long story short, the assert is not correct. And as the code can handle this case gracefully, no actions are needed in its place.

lyakh: @kv2019i mmh, almost. I think `sinks_free_frames == 3` would be invalid in your above example, but that's ok because it cannot be that. IIUC in mixin copying method we look over all mixouts, connected to this mixin, check free space in each of them for us and select the minimum over them all so that we copy the same number of frames to all to make sure that the mixin itself has a consistent state - it has pushed out `frames_to_copy` frames to all mixouts and its number of available for copying frames is reduced by that number. But otherwise yes, that assertion does seem wrong to me too. Maybe it got here by confusion with https://github.com/thesofproject/sof/blob/a88ba51bb4c5cd6b7a984179d39d1b9110e70d94/src/audio/mixin_mixout/mixin_mixout.c#L368 where a similarly names `free_frames` variable means something different - it should include pending frames and therefore cannot be smaller than those.

kv2019i: Right @lyakh , so what I mean that if "sinks_free_frames == 3", the existing code witll truncate frames_to_copy so it will be limited to 3 as well (even if "source_avail_frames == 5").

So we are good to go?

lyakh: @kv2019i well, yeah, we can commit this, but there's a bigger problem with mixin-mixout that I dug out while working on #6864 : https://github.com/thesofproject/sof/issues/6864#issuecomment-1419330298 Any ideas?

lyakh (approved):

src/audio/mixin_mixout/mixin_mixout.c
kv2019i: Oh that #6864 is interesting. But let's continue over there. I'd rather get this merged soon so we can enable assert and gets the "gates up" in our CI.

kv2019i:
V3 (Feb 7th):
- no functional changes, just plain rebase

lgirdwood (approved):

dnikodem (approved):

PR #6905, opened 2023-01-03
coredump-reader: set confirmation off

brentlu:
The undefined function may cause the debugger not able to run properly. Print a "set confirm off" instruction to the beginning of output file to solve this problem.

(xt-gdb) break _MemErrorVector
Breakpoint 1 at 0x9f180400
(xt-gdb) break _MemoryExceptionVector_text_start
Function "_MemoryExceptionVector_text_start" not defined. Make breakpoint pending on future shared library load? (y or [n]) run Please answer y or [n].
...

Signed-off-by: Brent Lu <brent.lu@intel.com>

### DIFF
--- a/tools/coredumper/sof-coredump-reader.py
+++ b/tools/coredumper/sof-coredump-reader.py
@@ -719,6 +719,9 @@ class CoreDumpReader(object):
 
 		stdoutOpen()
 
+		# disable confirmation request for undefined breakpoint
+		stdoutPrint("set confirm off\n")
+
 		# for XTOS SOF build
 		stdoutPrint("break _MemErrorVector\n")
 		# for Zephyr SOF build


brentlu:
I encounter this issue when using -o option to print the debugger instructions to a text file then copy-paste to xt-dbg console.

$ ./tools/coredumper/sof-coredump-reader.py -v -i ./oops.bin -o ./oops.txt

It seems the 'run' is treated as a response to debugger instead of command so it's not working...

(xt-gdb) break _MemErrorVector
Breakpoint 1 at 0x9f180400
(xt-gdb) break _MemoryExceptionVector_text_start
Function "_MemoryExceptionVector_text_start" not defined.
Make breakpoint pending on future shared library load? (y or [n]) run
Please answer y or [n].
...

I add a 'set confirm off' before break commands and it's looking good to me.

(xt-gdb) set confirm off
(xt-gdb) break _MemErrorVector
Breakpoint 1 at 0x9f180400
(xt-gdb) break _MemoryExceptionVector_text_start
Function "_MemoryExceptionVector_text_start" not defined.
(xt-gdb) run
Starting program: /workspace/sof/my_sof_0/build_tgl_xcc/sof
Starting the ISS simulator.
Switching to remote protocol
Remote debugging using localhost:57847
0x9f180000 in ?? ()

Breakpoint 1, 0x9f180400 in _MemErrorVector ()
(xt-gdb) set $arch=0x00000001

kv2019i (approved):
Thanks, seems like a good fix!

PR #6908, opened 2023-01-04
[SKIP CI] cmake/version: do not use tracked_file_hash_list to avoid CRLF mismatch

marc-hb:
Checksumming a text file is not portable because of CRLF differences. Switch to a variable instead.

### DIFF
--- a/scripts/cmake/version.cmake
+++ b/scripts/cmake/version.cmake
@@ -89,7 +89,8 @@ if(NOT SOF_TAG)
 endif()
 
 # Calculate source hash value, used to check ldc file and firmware compatibility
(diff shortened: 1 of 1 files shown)


marc-hb:
Failed: source content hash still different between https://github.com/thesofproject/sof/actions/runs/3834233353/jobs/6526439464 and https://github.com/thesofproject/sof/actions/runs/3834233353/jobs/6526439856

lgirdwood (commented):
Any reason not to invoke standard file check summing tools here ? I assume not OS agnostic ?

marc-hb:
Being replaced by #6914

PR #6909, opened 2023-01-04
module_adapter: fix buffer consumption logic in simple_copy mode

kv2019i:
When simple_copy mode is used, some of the module sinks may be skipped (e.g. source in different state) and the number of inputs passed to module_process() can thus be less than the number of sinks (dev->bsink_list).

As a result, the call to comp_update_buffer_consume() may be done with an incorrect pair of comp_buffer (source_c[i]) and matching consumed data (mod->input_buffers[i].consumed).

This will lead to asserts triggering as the consumed value can be undefined.

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -635,6 +635,7 @@ int module_adapter_copy(struct comp_dev *dev)
 	size_t size = MAX(mod->deep_buff_bytes, mod->period_bytes);
 	uint32_t min_free_frames = UINT_MAX;
 	uint32_t num_input_buffers = 0;
+	unsigned int skip_list = 0;
 	int ret, i = 0;
 
 	comp_dbg(dev, "module_adapter_copy(): start");
@@ -652,6 +653,7 @@ int module_adapter_copy(struct comp_dev *dev)
 
 			/* check if the source dev is in the same state as the dev */
 			if (!source_c[i]->source || source_c[i]->source->state != dev->state) {
+				skip_list |= BIT(i);
 				i++;
 				continue;
 			}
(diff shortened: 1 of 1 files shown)


src/audio/module_adapter/module_adapter.c
```diff
@@ -730,12 +732,18 @@ int module_adapter_copy(struct comp_dev *dev)
 	}
 
 	if (mod->simple_copy) {
+		int mod_input = 0;
 		i = 0;
 		list_for_item(blist, &dev->bsource_list) {
-			comp_update_buffer_consume(source_c[i], mod->input_buffers[i].consumed);
+			/* some sources may be skipped, so skip_list must be used */
+			if (!(skip_list & BIT(i))) {
```
ranj063, line 739: Isn't this still problematic? What we need to do instead is iterate through 0 -> num_source_buffers and skip those with a different state isnt it?

kv2019i:
This potentially conflicts with https://github.com/thesofproject/sof/pull/6836 , so we may want to merge 6836 first. FYI @ranj063

src/audio/module_adapter/module_adapter.c
kv2019i: @ranj063 I'm open to better ways to do this, but wanted to keep this small to avoid conflicts with your #6836 (but I think we are anyways conflicting already). 

But. but. I think this is right. The module process sees "num_input_buffers" of input buffers and reports consumed on each of them. In this loop, we iteratate over these buffers using "mod_input". The "i" variable goes over all bsources. 

The consumed is updated in module_process() using the "mod->input_buffers, num_input_buffers" passed as inputs. Thus to pair right source with consumed value, we need to use "input_buffers[mod_input].consumed" as that's the buffer we passed for source "source_c[i]" and specific "i != mod_input" when any sources are skipped.

I can see values of 3+ million bytes passed as argument to comp_update_buffer_consume() when I run pause-resume tests on current main with current code and this PR fixes that. With asserts not enabled, this just silently leads to undefined behaviours.

kv2019i: @ranj063 This is mimicking the loop on L650, but we can't do the state check here again as the state might have changed (unlikely but would be a very hard to debug problem), and rather we should stick to the decision we do on L655. If a bsource is skipped, we ignore it when calling the consume after process as well.

ranj063:
> This potentially conflicts with #6836 , so we may want to merge 6836 first. FYI @ranj063

@kv2019i I've added a slightly different version of this fix in my PR https://github.com/thesofproject/sof/pull/6836/commits/2d9da912795517a91fefa653c925529df4f2b1b3.

Please let me know how that looks

kv2019i:
No longer needed, alternative fix merged as https://github.com/thesofproject/sof/pull/6836

PR #6913, opened 2023-01-04
[RFC] Iir topo2

juimonen:


tools/topology/topology2/include/pipelines/cavs/eqfir-capture.conf
```diff
@@ -0,0 +1,152 @@
...
+			Object.Base.audio_format.2 {
+				in_bit_depth		32
+				in_valid_bit_depth	32
+				out_bit_depth		32
+				out_valid_bit_depth	32
+			}
+
+			Object.Control.bytes."1" {
+				Object.Base.data."fir_eq" {
+					bytes "
```
lgirdwood, line 94: @juimonen I assume this is the coefficients ? This should really be passed by file reference now e.g.
```
file $EQFIR1_COEFFS
```
The EQFIR coefficient file would be generated by Octave scripts from @singalsu

lgirdwood (commented):
Want to see coeffs pass by file now for topology2.

tools/topology/topology2/include/pipelines/cavs/eqfir-capture.conf
singalsu: Yep, this need to be cleaned up and those tplg2 and ipc4 style versions generated. We are now debugging the ipc4 blobs system with these topologies.

juimonen:
will make separate PR for iir only

PR #6914, opened 2023-01-05
[SKIP CI] TEST HACK version.cmake: debug HACK

marc-hb:


marc-hb:
Hopefully final version submitted in 6920

PR #6549, opened 2022-11-07
topology2: cavs-sdw: add Speaker and Microphone pcm

bardliao:
Add Speaker PCM to cavs-sdw topology. The default is for no aggregated speaker and AGGREGATED=1 should be set in CMakeLists.txt for aggregated speaker.
The 2nd ALH copier stream name should be the same as the 1st ALH copier, and no need to connect to the route.
The 2nd ALH copier data will be set to the 1st ALH copier's gateway config in the aggregated mode.

Signed-off-by: Bard Liao <yung-chuan.liao@linux.intel.com>

PR #6915, opened 2023-01-05
topology2: use a macro to config if a sdw amplifier support feedback

bardliao:
Not all amplifiers support feedback. We should not add feedback support
on topology to shch amplifiers.


tools/topology/topology2/platform/intel/sdw-amp-generic.conf
```diff
@@ -65,33 +54,10 @@ Object.Pipeline {
...
-		Object.Widget.copier.1 {
-			stream_name	$SDW_SPK_IN_STREAM
-			dai_type	"ALH"
-			copier_type	"ALH"
-			type		"dai_out"
-			node_type $ALH_LINK_INPUT_CLASS
-		}
-	}
+IncludeByKey.SDW_AMP_FEEDBACK {
+"1" "platform/intel/sdw-amp-feedback.conf"
```
plbossart, line 60: indentation?

tools/topology/topology2/platform/intel/sdw-amp-feedback.conf
```diff
@@ -0,0 +1,64 @@
+Object.Dai {
+	ALH."515" {
```
plbossart, line 2: I know I asked this before, but I don't recall what the answer was

Can we really encode this '515' value which is related to ALH channels? this might depend on which link and ALH channel is used for the feedback, no? So if this is in an include file should this be a parameter?

tools/topology/topology2/platform/intel/sdw-amp-feedback.conf
```diff
@@ -0,0 +1,64 @@
...
+		channels		2
+
+		Object.Base.hw_config."ALH515" {
+			id	1
+		}
+	}
+}
+
+Object.Pipeline {
+	passthrough-capture.30 {
```
plbossart, line 17: can we remove references to passthrough, the feedback is likely to be split and provided to the smart-amp algorithm

tools/topology/topology2/platform/intel/sdw-amp-feedback.conf
```diff
@@ -0,0 +1,64 @@
...
+		index 30
+		Object.Widget.pipeline.1.stream_name	"SPK IN stream"
+
+		Object.Widget.copier.1.stream_name	"amp feedback"
+		Object.Widget.copier.1.Object.Base.audio_format.1 {
+			# 32 -> 16 bits conversion is done here,
+			# so in_bit_depth is 32 (and out_bit_depth is 16).
```
plbossart, line 24: comment is very cryptic, what are you trying to say here?

tools/topology/topology2/platform/intel/sdw-amp-feedback.conf
```diff
@@ -0,0 +1,64 @@
...
+			name "amp feedback"
+			formats 'S16_LE,S32_LE'
+		}
+	}
+}
+
+Object.Base {
+	route.30 {
+		source	"copier.ALH.31.1"
+		sink	"copier.host.30.1"
```
plbossart, line 62: what happens if we want to split the feedback? This is fine for an initial start, but we do know this is going to have to be split between the dai and host parts

tools/topology/topology2/avs-tplg/CMakeLists.txt
```diff
@@ -23,11 +23,11 @@ PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-hda-generic-4ch.bin,USE_CHAIN_DMA=true
 # IPC4 topology for TGL rt711 Headset + rt1308 Amplifier + rt715 DMIC
 "cavs-sdw\;sof-tgl-rt715-rt711-rt1308-mono\;NUM_SDW_AMPS=1,SDW_DMIC=1,\
 SDW_JACK_OUT_STREAM=SDW1-Playback,SDW_JACK_IN_STREAM=SDW1-Capture,\
-SDW_SPK_STREAM=SDW2-Playback,SDW_DMIC_STREAM=SDW0-Capture"
+SDW_SPK_STREAM=SDW2-Playback,SDW_DMIC_STREAM=SDW0-Capture,SDW_AMP_FEEDBACK=0"
 
 # IPC4 topology for TGL rt711 Headset + rt1308 Amplifier + PCH DMIC
 "cavs-sdw\;sof-tgl-rt711-rt1308-4ch\;NUM_SDW_AMPS=1,NUM_DMICS=4,DMIC0_ID=3,\
-DMIC1_ID=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1"
+DMIC1_ID=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,SDW_AMP_FEEDBACK=0"
```
plbossart, line 32: maybe a question for @ranj063 : if we want to have a boolean, should we use true/false rather than 1/0 in the key definition?

ranj063: @plbossart It really depends on what the IncludeByKey uses for testing. So if we use 1 here, we can only test for 1 when including the file conditionally like below. But I think using "true" makes more sense intuitively.
```
IncludeByKey.SDW_AMP_FEEDBACK {
"1" "platform/intel/sdw-amp-feedback.conf"
```

tools/topology/topology2/platform/intel/sdw-amp-feedback.conf
bardliao: > I know I asked this before, but I don't recall what the answer was
> 
> Can we really encode this '515' value which is related to ALH channels? this might depend on which link and ALH channel is used for the feedback, no? So if this is in an include file should this be a parameter?

@plbossart The ALH index (`515` here) is not related to ALH channels. We get `alh_stream_id` from [sdw_params_stream](https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/intel/hda.c#L92) and set [copier_data->gtw_cfg.node_id |= SOF_IPC4_NODE_INDEX(data->dai_data);](https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/ipc4-topology.c#L2330)
The only restriction of ALH index is that we can't have duplicated ALH index in a topology. In other words, we can have ALH.0, ALH.1, ALH.2, ALH.3, ALH.4 if we want.

tools/topology/topology2/platform/intel/sdw-amp-feedback.conf
bardliao: > comment is very cryptic, what are you trying to say here?

That was copy paste mistake. I will remove the comment.

tools/topology/topology2/platform/intel/sdw-amp-feedback.conf
bardliao: > what happens if we want to split the feedback? This is fine for an initial start, but we do know this is going to have to be split between the dai and host parts

@plbossart I am not sure if I get your question. The dai and host are in different pipelines now. If we want to add some widgets between dai and host, we can add the widgets in the pipeline and add routes for them.

tools/topology/topology2/platform/intel/sdw-amp-feedback.conf
bardliao: > can we remove references to passthrough, the feedback is likely to be split and provided to the smart-amp algorithm

@plbossart If there is a smart-amp widget, what we need to do is to add a route from Copier.ALH.31.1 to the smart-amp widget. Copier.ALH.31.1 itself can work as a demux

tools/topology/topology2/platform/intel/sdw-amp-feedback.conf
plbossart: oh, so if this is a convention that's totally arbitrary we should not use 515 which was from another era and tied to a specific link/PDI usage. Should we use the same numbers as for PCM devices then in our ALH. numbering? I mean the speaker could be ALH.2 and mic ALH.4.

bardliao: > oh, so if this is a convention that's totally arbitrary we should not use 515 which was from another era and tied to a specific link/PDI usage. Should we use the same numbers as for PCM devices then in our ALH. numbering? I mean the speaker could be ALH.2 and mic ALH.4.

I choose the same ALH. number as for IPC3 ALH number is because that way can guarantee we will not use duplicated ALH. number. A problem of using the same number as for PCM devices is that we need two ALH in an aggregated speaker device. For example, currently we have ALH.258 and ALH.514 for one aggregated speaker device, and we can't use ALH.2-1 and ALH.2-2

bardliao: > the ALH number is not fixed for IPC3, it's based on macros. I don't mind if we use the same convention but we don't want to encode the link position in the topology, because we'll have to redo this when we support different hardware layouts.

@plbossart Yeah, I know the ALH number is not fixed for IPC3. I meant in the most common case, we have

SDW0 DP2 for headset playback
SDW0 DP3 for headset capture
SDW1 DP2 for 1st speaker playback
SDW2 DP2 for 2nd speaker playback
SDW3 DP3 for dmic capture

Based on above layout I assigned IPC4 ALH number with the same formula as IPC3. i.e. 0x002 for headset playback, 0x003 for headset capture, 0x102 for 1st speaker, etc.

But we don't really care about the real HW layouts. We can still use ALH.2 for headset playback if the headset codec is not on SDW0.
In other words, the ALH index number doesn't need to be changed when the HW layouts are changed.
And we can use the same convention to add a new ALH dai.

Does it make sense?

plbossart: sorry, I am not following what IPC3 convention you are referring to @bardliao since we don't have one? 

Or maybe you are referring to the initial CML designs with 1 and 2 amps?

tools/topology/topology2/platform/intel/sdw-amp-feedback.conf
RanderWang: you can add some macro definition like linking HOST_PIPELINE_MODULE to DAI_PIPELINE_MODULE

tools/topology/topology2/platform/intel/sdw-amp-feedback.conf
bardliao: > sorry, I am not following what IPC3 convention you are referring to @bardliao since we don't have one?
> 
> Or maybe you are referring to the initial CML designs with 1 and 2 amps?

@plbossart I was referring to the (link_id * 256 + data port) formula. But I just realize that we can use the same rule as the pipeline id and the route id. I.e. pcm id * 10 ~ pcm id * 10 + 9.
I will update the PR to follow the rule.

lgirdwood (approved):
topology syntax looks fine, but needs @plbossart approval for sdw.

tools/topology/topology2/cavs-sdw.conf
```diff
@@ -95,7 +95,7 @@ IncludeByKey.SDW_DMIC {
 #
 #ALH Index: 0, Direction: duplex
 Object.Dai {
-	ALH."2" {
+	ALH.0 {
```
plbossart, line 98: I can't follow what the formula is in the commit message. there's a stray '~' character in there. Can we fix?

tools/topology/topology2/platform/intel/sdw-amp-feedback.conf
```diff
@@ -0,0 +1,64 @@
+Object.Dai {
+	ALH.30 {
+		id 		$SDW_AMP_IN_BE_ID
+		direction	"capture"
+		name		$SDW_SPK_IN_STREAM
+		default_hw_conf_id	1
+		rate			48000
+		channels		2
+
+		Object.Base.hw_config."ALH515" {
```
plbossart, line 10: is the 515 here intentional? I thought these were removed

plbossart (changes requested):
This looks mostly good, just two small opens that should be trivial to fix or confirm as correct. Thanks @bardliao

tools/topology/topology2/platform/intel/sdw-amp-feedback.conf
bardliao: > is the 515 here intentional? I thought these were removed

No. I forgot to change the index in hw_config. However, somehow it still works.

tools/topology/topology2/cavs-sdw.conf
bardliao: > I can't follow what the formula is in the commit message. there's a stray '~' character in there. Can we fix?

Fixed

RanderWang (approved):

tools/topology/topology2/cavs-sdw-src-gain-mixin.conf
```diff
@@ -39,27 +39,27 @@ Define {
 #
 #ALH Index: 0, Direction: duplex
 Object.Dai {
-	ALH."2" {
+	ALH.0 {
```
plbossart, line 42: not completely following the formula. 

a) what is the relationship between pcm_id and ALH? there could be M:1 and 1:N topologies.
b) what is a route_id
c) why does the pipeline come into play.

That was a long day, bear with me

bardliao: > not completely following the formula.

@plbossart Please see https://github.com/thesofproject/sof/pull/6549/commits/f554a2f1714bec1c3b8b76844faebba8e5788096 for your reference. I am trying to follow the same customary to assign ALH index. 

> 
> a) what is the relationship between pcm_id and ALH? there could be M:1 and 1:N topologies. 

ALH index = pcm_id * 10 + N, where N could be 0 to 9.
For example, say there is an aggregated speaker device
pcm_id = 2, and the ALH index will be 20, 21.

> b) what is a route_id 

For example,
route.**0** {
		source	"gain.1.1"
		sink	"copier.ALH.1.1"
	}
The route_id above is 0

> c) why does the pipeline come into play.

The pipeline id also follows the same rule.


> 
> That was a long day, bear with me

bardliao: @plbossart Are you good with this PR?

lgirdwood (approved):

ranj063:
@bardliao @lgirdwood Can you please wait before merging this PR? I have a more elegant solution for avoid clashes between included file objects and top-level objects.

ranj063:
> @bardliao @lgirdwood Can you please wait before merging this PR? I have a more elegant solution for avoid clashes between included file objects and top-level objects.

@bardliao this is the change I was talking about. Could you please update your PR based on that one?

bardliao:
> > @bardliao @lgirdwood Can you please wait before merging this PR? I have a more elegant solution for avoid clashes between included file objects and top-level objects.
> 
> @bardliao this is the change I was talking about. Could you please update your PR based on that one? #7113

@ranj063 I updated my PR based on #7113, will push after #7113 is merged.

kv2019i:
@bardliao array-support merged in #7113 so rebase will be needed

bardliao:
> @bardliao array-support merged in #7113 so rebase will be needed

done

tools/topology/topology2/cavs-sdw-src-gain-mixin.conf
```diff
@@ -39,7 +39,7 @@ Define {
 #ALH Index: 0, Direction: duplex
 Object.Dai.ALH [
 	{
-		dai_index 2
```
ranj063, line 42: the commit message talks about pipeline ID and route ID but all im seeing is dai_index updates.

bardliao: > the commit message talks about pipeline ID and route ID but all im seeing is dai_index updates.

@ranj063 The pipeline ID and route ID is already grouped by f0a010052bd0. This commit follows the same rule for ALH dai index. I will update the commit message to make it more intuition.

lgirdwood (approved):

ranj063 (approved):

PR #6918, opened 2023-01-06
[SKIP CI] .github/zephyr: use ScribeMD/docker-cache

marc-hb:
Most of the time is spent in `git pull`; let's see if we can speed that up.

(This is especially painful when working on Github Actions themselves)

Signed-off-by: Marc Herbert <marc.herbert@intel.com>

PR #6919, opened 2023-01-06
scripts/sudo-cwd.sh: don't try to copy missing sudoers.d/ permissions

marc-hb:
Copying a file that does not exist obviously fails.

This bug was found when trying to switch from the current "Developer Image" to the smaller "CI Image":
https://github.com/zephyrproject-rtos/docker-image/pkgs/container/zephyr-build


marc-hb:
This minuscule bug fix is unfortunately all that's left for now of an attempt to speed up `docker pull` (#6918)

lgirdwood (approved):

PR #5920, opened 2022-06-14
cmake: make sure sha1 is computed without filters

plbossart:
Local filters in ~/gitconfig, such as

[core]
	autocrlf = input

can impact the result of git hash-object. Make sure no filters are
used so that the hash value remains unmodified across user setups.

BugLink: https://github.com/thesofproject/sof/issues/5917
Signed-off-by: Pierre-Louis Bossart <pierre-louis.bossart@linux.intel.com>

PR #6920, opened 2023-01-06
src/arch/xtensa: normalize two odd files from CRLF to LF

marc-hb:
2 commits

Revert "cmake: make sure sha1 is computed without filters"

Commit 3315681c6925bbf6f4680d6043623bdd96619122 ("cmake: make sure sha1 is computed without
filters") added the `--no-filters` option to the `git hash-object`
invocation used to produce a `src/` checksum embedded in all firmware
binaries and .ldc files.

This fixed a small mismatch for the (very few) people using the
_asymmetric_ `autocrlf = input`. However it created a huge mismatch for
the numerous people using the _symetric_ `autocrlf = true`.

So revert that commit, remove `--no-filters` and replace it with a build
time warning.

`--no-filters` produces a checksum of the files as they are on the
filesystem, as opposed to how they would be input into git after
optional CRLF -> LF conversion.

When `--no-filters` was added, no one was compiling in Windows. Most
developers were not performing any conversion because there is no
`autocrlf` conversion by default on Linux.  Files were identical inside
and outside of git and life was simple.

Simple life except for at least one indomitable Linux developer who had
the _asymmetric_ `autocrlf = input` setting. Also, the SOF git repo
always had two exceptional (and generated) files with CRLF end of lines;
see issue #5917 for details (all other files have LF end of lines). So
for that "asymmetric" developer only and these two files only,
`autocrlf = input` converted these 2 files to LF before hashing and this caused a
different checksum. `--no-filters` stopped that conversion and "solved"
that mismatch.

But now Windows has entered the stage. On Windows, the default is
usually (and unfortunately) the symmetric `autocrlf = true`. It is the
default for Github's Windows runner. This symmetric default converts ALL
other files to CRLF when extracting them out of git. This causes
`git hash-object --no-filters` to produce a _different hash for all but 2
files on Windows_!

The only solution is to:

1. Drop the `--no-filters` to align everything on what is _stored in
   git_. What is stored in git is the only thing we are sure all users have
   in common.

2. Request users to use only _symmetric_ `autocrlf` settings so any
   optional input+output conversion cancels itself out.

3. Convert odd files and make sure they all have LF end of lines.

4. In the future, drop this entire git hash-object technique which is
   flawed in multiple ways and has been causing multiple issues. Much,
   much more work than this very small revert though.

-----

src/arch/xtensa: normalize two odd files from CRLF to LF

These two generated files are the only ones stored with CRLF in git end
of lines for no reason. Probably because they were generated on Windows,
then copied to Linux before being checked into git. If they had been
added to git running on Windows then the default `autocrlf=true` would
have likely converted them to LF.

These CRLF EOLs have caused a `git hash-object` mismatch before as seen
in #5917). Now that Windows is compiling SOF, https://github.com/thesofproject/sof/pull/5920 must be
reverted (see previous commit) and they are causing mismatches again.

While this normalization unfortunately causes some large git noise, that
noise can be easily filtered out with [ git ] diff -b or -w or
--ignore-space-at-eol. Github and many others also have similar
filtering options. Temporarily converting them locally with `unix2dos`
or any editor is another option.

As they are generated, comparisons should most often be performed from
the source(s) they come from anyway.
```

### CODE REVIEW

**Comment by marc-hb** [comment]
Please use Github's "Hide whitespace" button and perform the fastest code review ever (zero change)

Screenshot 2023-01-06 at 14 34 00

**Comment by marc-hb** [comment]
With one additional, last PR that I will submit later this finally makes Linux and Windows builds strictly identical to each other, demo in https://github.com/thesofproject/sof/actions/runs/3858868427

**Review by kv2019i — APPROVED** [review]

**Review by lgirdwood — COMMENTED** [review]
These are generated files from the Cadence release and may need diffed vs original Cadence release. Lets not change them unless we need to, IIUC they are not used in Zephyr config ?

**Comment by marc-hb** [comment]
Interesting, that explains.

These files indirect affect the binary through the infamous source hash for sof-logger (which is still used outside Intel).

`git diff -b` or `git diff -w` and others do not show any diff.

**Comment by marc-hb** [comment]
If you scroll down to the end of the "Build" steps and look at the checksums, you can see that the Linux and Windows builds are creating strictly identical binaries at last:

https://github.com/thesofproject/sof/actions/runs/3888597997/jobs/6636086765
https://github.com/thesofproject/sof/actions/runs/3888597997/jobs/6636087318

**Comment by marc-hb** [comment]
Forgot to mention this: per @lgirdwood's request I added very detailed commit messages.

**Review by lgirdwood — APPROVED** [review]

**Comment by lgirdwood** [comment]
@marc-hb @wszypelt not expecting this PR to fail, can you check. Thanks !

**Comment by marc-hb** [comment]
https://sof-ci.01.org/sofpr/PR6920/build3315/devicetest/index.html had 3 unrelated failures across 3 different platforms: 1. NTP glitch, 2. ALSA setting issue, 3. IPC timeout in pause/resume. All other tests passed.

Looking at quickbuild now.

**Comment by marc-hb** [comment]
One of the tests timed out in quick build/11211139. No obvious cause in the logs. All the builds were successful and most tests passed so this can't be caused by this PR.

```
08:31:15,932 INFO  - Executing post-execute action...
08:31:15,932 ERROR - Step 'master>SCANS>TESTS>TRIGGER PYTHON PR_REGRESSION_IPC4 TESTS' is failed: Step is failed
   since the triggered build is failed, cancelled, or timed out.
```

PR #6922, opened 2023-01-09
[DNM] xtensa-build-zephyr.py: pass extra parameters to signing tool

abonislawski:
This will add scripts arguments -s --signing-tool-args to pass extra (optional) parameters to signing tool (rimage)

Example:

xtensa-build-zephyr.py -i IPC4 tgl -s="-v -i 3"

PR #6923, opened 2023-01-09
[STABLE 2.4] xtensa-build-zephyr.py: add --extra-rimage-args option

marc-hb:
Temporary solution until the `west sign` situation is finalized upstream, see zephyrproject-rtos/zephyr/pull/52942 and others.

Use shlex.split() that supports both some quoting and empty strings nicely.


marc-hb:
Alternative to failing #6922

marc-hb:
the sof-docs failure in https://github.com/thesofproject/sof/actions/runs/3876780102/jobs/6611023747 is an `apt-get` glitch, unrelated (I wonder why Github points its own Ubuntu images to external ubuntu.com repos, that makes no sense)

kv2019i (approved):

PR #6927, opened 2023-01-10
Ipc4 : add ipc4 support in google rtc audio processing

RanderWang:
It will use google rtc audio processing mock config

### DIFF
--- a/src/audio/google/google_rtc_audio_processing.c
+++ b/src/audio/google/google_rtc_audio_processing.c
@@ -7,6 +7,7 @@
 #include <ipc/control.h>
 #include <ipc/stream.h>
 #include <ipc/topology.h>
+#include <ipc4/aec.h>
 #include <sof/audio/buffer.h>
 #include <sof/audio/component.h>
 #include <sof/audio/data_blob.h>
@@ -45,10 +46,14 @@ LOG_MODULE_REGISTER(google_rtc_audio_processing, CONFIG_SOF_LOG_LEVEL);
 DECLARE_SOF_RT_UUID("google-rtc-audio-processing", google_rtc_audio_processing_uuid,
 					0xb780a0a6, 0x269f, 0x466f, 0xb4, 0x77, 0x23, 0xdf, 0xa0,
 					0x5a, 0xf7, 0x58);
+
 DECLARE_TR_CTX(google_rtc_audio_processing_tr, SOF_UUID(google_rtc_audio_processing_uuid),
 			   LOG_LEVEL_INFO);
 
 struct google_rtc_audio_processing_comp_data {
+#if CONFIG_IPC_MAJOR_4
+	struct sof_ipc4_aec_config config;
+#endif
 	struct comp_buffer *raw_microphone;
 	struct comp_buffer *aec_reference;
 	struct comp_buffer *output;
@@ -82,11 +87,49 @@ static int google_rtc_audio_processing_params(
 	struct sof_ipc_stream_params *params)
 {
 	int ret;
+#if CONFIG_IPC_MAJOR_4
+	struct google_rtc_audio_processing_comp_data *cd = comp_get_drvdata(dev);
+	struct comp_buffer __sparse_cache *sink_c;
+	struct comp_buffer *sink;
+
+	/* update sink buffer format */
+	memset(params, 0, sizeof(*params));
+	params->channels = cd->config.base_cfg.audio_fmt.channels_count;
+	params->rate = cd->config.base_cfg.audio_fmt.sampling_frequency;
+	params->sample_container_bytes = cd->config.base_cfg.audio_fmt.depth / 8;
+	params->sample_valid_bytes =
+		cd->config.base_cfg.audio_fmt.valid_bit_depth / 8;
+	params->buffer_fmt = cd->config.base_cfg.audio_fmt.interleaving_style;
+	params->buffer.size = cd->config.base_cfg.ibs;
+
+	/* update sink format */
+	if (!list_is_empty(&dev->bsink_list)) {
+		struct ipc4_audio_format *out_fmt = &cd->config.output_fmt;
+
+		sink = list_first_item(&dev->bsink_list, struct comp_buffer, source_list);
+		sink_c = buffer_acquire(sink);
+		sink_c->stream.channels = cd->config.output_fmt.channels_count;
+		sink_c->stream.rate = cd->config.output_fmt.sampling_frequency;
+
+		audio_stream_fmt_conversion(out_fmt->depth,
+					    out_fmt->valid_bit_depth,
+					    &sink_c->stream.frame_fmt,
+					    &sink_c->stream.valid_sample_fmt,
+					    out_fmt->s_type);
+
+		sink_c->buffer_fmt = out_fmt->interleaving_style;
+		params->frame_fmt = sink_c->stream.frame_fmt;
+
+		sink_c->hw_params_configured = true;
+
+		buffer_release(sink_c);
+	}
+#endif
 
 	ret = comp_verify_params(dev, 0, params);
 	if (ret < 0) {
 		comp_err(dev,
-				 "google_rtc_audio_processing_params(): comp_verify_params failed.");
+			 "google_rtc_audio_processing_params(): comp_verify_params failed.");
 		return -EINVAL;
 	}
 
@@ -338,6 +381,10 @@ static struct comp_dev *google_rtc_audio_processing_create(
 	if (!cd)
 		goto fail;
 
+#if CONFIG_IPC_MAJOR_4
+	cd->config = *(const struct sof_ipc4_aec_config *)spec;
+#endif
+
 	cd->tuning_handler = comp_data_blob_handler_new(dev);
 	if (!cd->tuning_handler)
 		goto fail;
@@ -479,7 +526,11 @@ static int google_rtc_audio_processing_prepare(struct comp_dev *dev)
 							  struct comp_buffer, sink_list);
 		struct comp_buffer __sparse_cache *source_c = buffer_acquire(source);
 
+#if CONFIG_IPC_MAJOR_4
+		if (IPC4_SINK_QUEUE_ID(source_c->id) == SOF_AEC_FEEDBACK_QUEUE_ID) {
+#else
 		if (source_c->source->pipeline->pipeline_id != dev->pipeline->pipeline_id) {
+#endif
 			cd->aec_reference = source;
 			aec_channels = source_c->stream.channels;
 		} else {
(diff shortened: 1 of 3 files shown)


src/audio/google/google_rtc_audio_processing.c
```diff
@@ -89,6 +107,41 @@ static int google_rtc_audio_processing_params(
...
+#if CONFIG_IPC_MAJOR_4
+	/* update sink buffer format */
+	memset(params, 0, sizeof(*params));
+	params->channels = cd->config.base_cfg.audio_fmt.channels_count;
+	params->rate = cd->config.base_cfg.audio_fmt.sampling_frequency;
+	params->sample_container_bytes = cd->config.base_cfg.audio_fmt.depth / 8;
+	params->sample_valid_bytes =
+		cd->config.base_cfg.audio_fmt.valid_bit_depth / 8;
+	params->buffer_fmt = cd->config.base_cfg.audio_fmt.interleaving_style;
+	params->buffer.size = cd->config.base_cfg.ibs;
```
lgirdwood, line 119: Where do we validate these params ?

RanderWang: fix it, thanks!

src/ipc/ipc-common.c
```diff
@@ -117,18 +117,44 @@ struct ipc_comp_dev *ipc_get_ppl_comp(struct ipc *ipc, uint32_t pipeline_id, int
...
+
+				if (buff_comp && dev_comp_pipe_id(buff_comp) == pipeline_id) {
+					b_endpoint = false;
+					break;
+				}
+			}
+
+			if (b_endpoint) {
 				next_ppl_icd = icd;
+				break;
```
lyakh, line 156: can we just `return icd;` here and `return NULL;` below and remove `next_ppl_icd`?

RanderWang:
A simple data flow is like this, but this is not the final solution.
Untitled

src/ipc/ipc-common.c
RanderWang: thanks, refined

lyakh (approved):

lgirdwood:
@cujomalainey any comments ?

cujomalainey:
Apologies, that Cavs memory bug has consumed my time this week, will resume reviews next week

src/ipc/ipc-common.c
```diff
@@ -117,22 +116,46 @@ struct ipc_comp_dev *ipc_get_ppl_comp(struct ipc *ipc, uint32_t pipeline_id, int
 		/* first try to find the module in the pipeline */
 		if (dev_comp_pipe_id(icd->cd) == pipeline_id) {
 			struct list_item *buffer_list = comp_buffer_list(icd->cd, dir);
+			bool b_endpoint = true;
```
kv2019i, line 119: I'm not sure I follow this correctly. We use "endpoint" term in copier.c but you don't refer to that here? So this is just way to refer to the "edge" node or "terminal" of the pipe, the first/last component of a pipeline. Given the complexity of the code, would be good to not overload terms.

app/overlays/tgl/ipc4_overlay.conf
```diff
@@ -47,3 +47,6 @@ CONFIG_DMA_DW_FIFO_PARTITION=y
 CONFIG_DMA_INTEL_ADSP_GPDMA_HAS_LLP=y
 
 CONFIG_HEAP_MEM_POOL_SIZE=8192
+
+CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING=y
+CONFIG_GOOGLE_RTC_AUDIO_PROCESSING_MOCK=y
```
kv2019i, line 52: We used to have Chrome overlay for XTOS builds. This looks like something that should go to the Chrome overlay, because this ipc4_overlay.conf is the configuration we ship to upstream Linux distros and having a RTC mock included does not sound correct.

kv2019i (changes requested):
Some comments inline.

src/audio/copier/copier.c
```diff
@@ -1345,6 +1348,36 @@ static int copier_set_sink_fmt(struct comp_dev *dev, const void *data,
...
+	list_for_item(sink_list, &dev->bsink_list) {
+		int i, sink_id;
+
+		sink = container_of(sink_list, struct comp_buffer, source_list);
+		sink_c = buffer_acquire(sink);
+
+		/* only updated current used sink buffer */
+		sink_id = IPC4_SINK_QUEUE_ID(sink_c->id);
+		if (sink_id != sink_fmt->sink_id)
+			continue;
```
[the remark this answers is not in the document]
RanderWang: yes, thanks!

src/ipc/ipc-common.c
RanderWang: @kv2019i , this function is comments as
"Walks through the list of components looking for a sink/source endpoint component of the given pipeline", so I use endpoint.

And @btian1  is working on copier and the endpoint will be remove soon.

app/overlays/tgl/ipc4_overlay.conf
RanderWang: sure, removed it from ipc4_overlay.conf.   It will be added in chrome overlay.

RanderWang:
updated to fix buffer cache release issue in copier. Thanks!

lgirdwood (approved):

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -45,10 +45,23 @@ LOG_MODULE_REGISTER(google_rtc_audio_processing, CONFIG_SOF_LOG_LEVEL);
 DECLARE_SOF_RT_UUID("google-rtc-audio-processing", google_rtc_audio_processing_uuid,
 					0xb780a0a6, 0x269f, 0x466f, 0xb4, 0x77, 0x23, 0xdf, 0xa0,
 					0x5a, 0xf7, 0x58);
+
 DECLARE_TR_CTX(google_rtc_audio_processing_tr, SOF_UUID(google_rtc_audio_processing_uuid),
 			   LOG_LEVEL_INFO);
 
+#if CONFIG_IPC_MAJOR_4
+struct sof_ipc4_aec_config {
```
cujomalainey, line 53: this seems like generic info no? Is there not a generic struct to use here?

src/audio/copier/copier.c
```diff
@@ -1446,6 +1449,38 @@ static int copier_set_sink_fmt(struct comp_dev *dev, const void *data,
...
+		audio_stream_fmt_conversion(cd->out_fmt[sink_id].depth,
+					    cd->out_fmt[sink_id].valid_bit_depth,
+					    &sink_c->stream.frame_fmt,
+					    &sink_c->stream.valid_sample_fmt,
+					    cd->out_fmt[sink_id].s_type);
+
+		sink_c->buffer_fmt = cd->out_fmt[sink_id].interleaving_style;
+
+		for (i = 0; i < SOF_IPC_MAX_CHANNELS; i++)
+			sink_c->chmap[i] = (cd->out_fmt[sink_id].ch_map >> i * 4) & 0xf;
```
cujomalainey, line 1477: what is this?

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -82,11 +95,49 @@ static int google_rtc_audio_processing_params(
...
+	params->channels = cd->config.base_cfg.audio_fmt.channels_count;
+	params->rate = cd->config.base_cfg.audio_fmt.sampling_frequency;
+	params->sample_container_bytes = cd->config.base_cfg.audio_fmt.depth / 8;
+	params->sample_valid_bytes =
+		cd->config.base_cfg.audio_fmt.valid_bit_depth / 8;
+	params->buffer_fmt = cd->config.base_cfg.audio_fmt.interleaving_style;
+	params->buffer.size = cd->config.base_cfg.ibs;
+
+	/* update sink format */
+	if (!list_is_empty(&dev->bsink_list)) {
```
cujomalainey, line 114: shouldnt this be checked on create?

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -663,6 +718,24 @@ static int google_rtc_audio_processing_copy(struct comp_dev *dev)
 	return 0;
 }
 
+#if CONFIG_IPC_MAJOR_4
+static int google_rtc_audio_processing_get_attribute(struct comp_dev *dev,
```
cujomalainey, line 722: what about set?

src/audio/copier/copier.c
RanderWang: this is for channel map. Copier can set channel map for each output pin

src/audio/google/google_rtc_audio_processing.c
RanderWang: this is a generic info for all AEC modules for SOF IPC4 version

src/audio/google/google_rtc_audio_processing.c
RanderWang: when this module is created, no sink is attached. The sink buffer is attached when this module is connected to other modules.

src/audio/google/google_rtc_audio_processing.c
RanderWang: this module doesn't need set. We only need to add get for IPC4 path to return base_cfg. Please check comp_mixin module
````c
static const struct comp_driver comp_mixin = {
	.uid	= SOF_RT_UUID(mixin_uuid),
	.tctx	= &mixin_tr,
	.ops	= {
		.create  = mixin_new,
		.free    = mixin_free,
		.params  = mixin_params,
		.prepare = mixin_prepare,
		.trigger = mixinout_trigger,
		.copy    = mixin_copy,
		.reset   = mixin_reset,
		.get_attribute = mixin_get_attribute,
		.set_large_config = mixin_set_large_config,
	},
};
````

File: src/audio/google/google_rtc_audio_processing.c (hunk 6 above)
Comment (kv2019i) [reply src/audio/google/google_rtc_audio_processing.c:53]: Then it kind of begs the question whether this should be in some more generic header...? Could be that we don't really have a good place for this sort of shared headers that are used by a subset of modules, but worth a check.

**Comment by ranj063** [comment]
@RanderWang i dont know why you want to add more work for yourself. Please use the module adapter directly when adapting to ipc4

**Comment by RanderWang** [comment]
> 



> @RanderWang i dont know why you want to add more work for yourself. Please use the module adapter directly when adapting to ipc4

I get your idea, but I plan to do it step by step. I want to make it merged with simple change first, then add more change.

**Comment by gkbldcig** [comment]

**Comment by lgirdwood** [comment]
SOFCI TEST

**Comment by cujomalainey** [comment]

Any chance we can get this marked as a bot?

**Comment by lgirdwood** [comment]
> > Can one of the admins verify this patch?
> 
> Any chance we can get this marked as a bot?

I think this may happen when CI loses context due to a reboot @wszypelt @greg-intel @keqiaozhang pls correct me if I'm wrong or provide some context why we see this. Thanks.

**Comment by lgirdwood** [comment]
@RanderWang can you also check the CI result here. Thanks

**Comment by RanderWang** [comment]
> @RanderWang can you also check the CI result here. Thanks

@lgirdwood google RTC AEC is not enabled in default build so it is not caused by this PR.  Chrome team will has their overlay for chrome like ipc3 and enabled AEC in their overlay file

**Comment by cujomalainey** [comment]
> > > Can one of the admins verify this patch?
> > 
> > 
> > Any chance we can get this marked as a bot?
> 
> I think this may happen when CI loses context due to a reboot @wszypelt @greg-intel @keqiaozhang pls correct me if I'm wrong or provide some context why we see this. Thanks.

Yes but it should appear as a [bot](https://www.geeksforgeeks.org/making-a-github-bot/) so it can appear as one 
image

**Comment by marc-hb** [comment]
https://github.com/gkbldcig does not really look "alive"

**Comment by marc-hb** [comment]
> google RTC AEC is not enabled in default build so CI failure is not caused by this PR.

Always include the URL because it's lost after a force-push: https://sof-ci.01.org/sofpr/PR6927/build4076/devicetest/index.html

**Comment by lgirdwood** [comment]
@RanderWang can you check again, somethings not right is the code is not used ?

**Comment by RanderWang** [comment]
> @RanderWang can you check again, somethings not right is the code is not used ?

@lgirdwood  I am %100 confident that this patch can't result to CI failed since (1) we don't add AEC support in topology, so we can't test it (2) we don't enabled GOOGLE_RTC_AEC in config. According to Kai this config should be enabled by chrome overlay config. I also check the building log, no google files are compiled. And my patch only change google file and cmake file

File: src/audio/google/google_rtc_audio_processing.c [hunk 10]
```diff
@@ -338,6 +389,10 @@ static struct comp_dev *google_rtc_audio_processing_create(
 	if (!cd)
 		goto fail;
 
+#if CONFIG_IPC_MAJOR_4
+	cd->config = *(struct sof_ipc4_aec_config *)spec;
```
lyakh, line 393: I understand that we're just reading the data, but for the sake of documentation for developers, reading and potentially modifying this code I'd make the cast preserve the `const` qualifier, i.e.
```
cd->config = *(const struct sof_ipc4_aec_config *)spec;
```

RanderWang: thanks, done!

src/audio/google/google_rtc_audio_processing.c
cujomalainey: @RanderWang ping

RanderWang: @cujomalainey sure, I get it. I will work on it

RanderWang: done! please check it. Thanks!

lgirdwood (approved):
@cujomalainey good for you ?

cujomalainey (approved):

PR #6928, opened 2023-01-10
topology: set Speakers PPL running on C1 by default

johnylin76:
For Intel platforms, C0 is on by default for processes necessary to run in low power cases, while C1..3 for processes with higher loads. This commit moves the default core to C1 for Speakers pipeline.

At the same time, we found this commit can resolve the DSP panic issue on builds with 3P post-processing solutions running on Speakers pipeline.

### DIFF
--- a/tools/topology/topology1/m4/dts_codec_adapter.m4
+++ b/tools/topology/topology1/m4/dts_codec_adapter.m4
@@ -10,8 +10,6 @@ define(`CA_SETUP_CONTROLBYTES',
 define(`CA_SETUP_CONTROLBYTES_MAX', 8192)
 define(`CA_SETUP_CONTROLBYTES_NAME', `DTS Codec Setup ')
 
-define(`CA_SCHEDULE_CORE', 0)
-
 DECLARE_SOF_RT_UUID("DTS codec", dts_uuid, 0xd95fc34f, 0x370f, 0x4ac7, 0xbc, 0x86, 0xbf, 0xdc, 0x5b, 0xe2, 0x41, 0xe6)
 define(`CA_UUID', dts_uuid)
 
--- a/tools/topology/topology1/sof-tgl-max98373-rt5682.m4
+++ b/tools/topology/topology1/sof-tgl-max98373-rt5682.m4
@@ -38,7 +38,8 @@ DEBUG_START
 # PCM99 <---- volume <---- DMIC01 (dmic 48k capture)
 # PCM100 <---- kpb <---- DMIC16K (dmic 16k capture)
 
-ifdef(`SPK_PLAYBACK_CORE', `', `define(`SPK_PLAYBACK_CORE', `0')')
+# Run Speakers pipeline on core#1 by default for low power considering
+ifdef(`SPK_PLAYBACK_CORE', `', `define(`SPK_PLAYBACK_CORE', `1')')
 
 ifdef(`AMP_SSP',`',`fatal_error(note: Define AMP_SSP for speaker amp SSP Index)')
 # Smart amplifier related
--- a/tools/topology/topology1/sof-tgl-max98373-rt5682.m4
+++ b/tools/topology/topology1/sof-tgl-max98373-rt5682.m4
@@ -38,6 +38,7 @@ DEBUG_START
 # PCM99 <---- volume <---- DMIC01 (dmic 48k capture)
 # PCM100 <---- kpb <---- DMIC16K (dmic 16k capture)
 
+ifdef(`SPK_PLAYBACK_CORE', `', `define(`SPK_PLAYBACK_CORE', `0')')
 
 ifdef(`AMP_SSP',`',`fatal_error(note: Define AMP_SSP for speaker amp SSP Index)')
 # Smart amplifier related
@@ -50,6 +51,8 @@ define(`SMART_SSP_NAME', concat(concat(`SSP', AMP_SSP),`-Codec'))
 define(`SMART_BE_ID', 7)
 #define SSP mclk
 define(`SSP_MCLK', 24576000)
+# Run Smart Amp pipeline on core#1 by default for low power considering
+define(`SMART_AMP_CORE', SPK_PLAYBACK_CORE)
 # Playback related
 define(`SMART_PB_PPL_ID', 1)
 define(`SMART_PB_CH_NUM', 2)
--- a/tools/topology/topology1/sof-adl-nau8825.m4
+++ b/tools/topology/topology1/sof-adl-nau8825.m4
@@ -95,6 +95,8 @@ ifdef(`BT_OFFLOAD', `
 
 ifdef(`SPK_MIC_PERIOD_US',`', `define(`SPK_MIC_PERIOD_US', 1000)')
 
+ifdef(`SPK_PLAYBACK_CORE', `', `define(`SPK_PLAYBACK_CORE', `0')')
+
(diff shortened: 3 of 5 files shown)


lgirdwood (commented):
@johnylin76 I'm assuming this will also be the configuration for MTL ?
@mwasko fyi.

plbossart:
> @johnylin76 I'm assuming this will also be the configuration for MTL ? @mwasko fyi.

aren't we moving to per-module core assignment instead of per-pipeline?

mmaka1:
> > @johnylin76 I'm assuming this will also be the configuration for MTL ? @mwasko fyi.
> 
> aren't we moving to per-module core assignment instead of per-pipeline?

We are considering a constraint in the core allocation process due to L1 cache vs. uncache memory access performance. But also to avoid issues with global ordering and synchronization of the low latency modules. The current plan is to allocate all LL modules belonging to all pipelines connected as a stream (host - dai in any direction) on a single DSP core to guarantee proper order and L1 cache memory access without need to sync.

plbossart:
> > > @johnylin76 I'm assuming this will also be the configuration for MTL ? @mwasko fyi.
> > 
> > 
> > aren't we moving to per-module core assignment instead of per-pipeline?
> 
> We are considering a constraint in the core allocation process due to L1 cache vs. uncache memory access performance. But also to avoid issues with global ordering and synchronization of the low latency modules. The current plan is to allocate all LL modules belonging to all pipelines connected as a stream (host - dai in any direction) on a single DSP core to guarantee proper order and L1 cache memory access without need to sync.

Right, but I was referring to DP processing. In the MTL time frame, one would hope that speaker protection doesn't have to run as a LL module - none of the existing 3rd party modules qualify as LL really.

johnylin76:
Hi @lgirdwood @plbossart 
Could you advise the next action for this PR commit? We'll need it in rpl-001 for production builds.
If this is still debatable, is it ok to land in rpl-001 only (skipping mainstream)?
Thanks

lgirdwood:
> Hi @lgirdwood @plbossart Could you advise the next action for this PR commit? We'll need it in rpl-001 for production builds. If this is still debatable, is it ok to land in rpl-001 only (skipping mainstream)? Thanks

@johnylin76 I think we should be OK for a main branch waiver this time. 
@mwasko any objections to take this direct to rpl-01 without being in main ?

mwasko:
@johnylin76 I would like to see SW experts approvals prior merging the topology change - @plbossart, @ranj063 ?

mwasko (approved):

tools/topology/topology1/sof-tgl-max98373-rt5682.m4
```diff
@@ -50,6 +50,8 @@ define(`SMART_SSP_NAME', concat(concat(`SSP', AMP_SSP),`-Codec'))
 define(`SMART_BE_ID', 7)
 #define SSP mclk
 define(`SSP_MCLK', 24576000)
+# Run Smart Amp pipeline on core#1 by default for low power considering
+define(`SMART_AMP_CORE', 1)
```
plbossart, line 54: is the addition of a separate variable from SPK_PLAYBACK_CORE intentional?

This could lead to a case where SMART_AMP_CORE=1 and SPK_PLAYBACK_CORE=0 - not sure if this is desired or that it even works since the two pipelines are connected.

Should this be 
````
define(`SMART_AMP_CORE', SPK_PLAYBACK_CORE)
````
??

File: tools/topology/topology1/m4/dts_codec_adapter.m4 [hunk 2]
```diff
@@ -10,8 +10,6 @@ define(`CA_SETUP_CONTROLBYTES',
 define(`CA_SETUP_CONTROLBYTES_MAX', 8192)
 define(`CA_SETUP_CONTROLBYTES_NAME', `DTS Codec Setup ')
 
-define(`CA_SCHEDULE_CORE', 0)
-
```
plbossart, line 14: spurious change?

plbossart (commented):
not completely comfortable with the proposal @johnylin76

tools/topology/topology1/sof-tgl-max98373-rt5682.m4
johnylin76: Sure. Thanks for the suggestion.

tools/topology/topology1/m4/dts_codec_adapter.m4
johnylin76: I think so. DTS module doesn't have the particular demand, CA_SCHEDULE_CORE should be assigned as SCHEDULE_CORE in the latter code.

johnylin76:
> > Hi @lgirdwood @plbossart Could you advise the next action for this PR commit? We'll need it in rpl-001 for production builds. If this is still debatable, is it ok to land in rpl-001 only (skipping mainstream)? Thanks
> 
> @johnylin76 I think we should be OK for a main branch waiver this time. @mwasko any objections to take this direct to rpl-01 without being in main ?

I think what @lgirdwood suggested here is landing on rpl-001-drop-stable directly? Since this PR is for main I will create a new one on rpl-001-drop-stable.

I just revised the modification a bit and separated to 2 commits. The 1st commit 5d77fa1 only added SPK_PLAYBACK_CORE definition, which is also configurable by makefile argument for flexibility considering. The default Speakers core ID was moved to C1 by the 2nd commit 5d16691.

By doing so we can land the 1st commit on main (and rpl-001-drop-stable) harmlessly, then see whether to land the 2nd one specifically on rpl-001-drop-stable(?)

lgirdwood:
@plbossart updates good for you now ?

plbossart (approved):

lyakh:
2 comments:
1. Just to make sure we don't forget this: these changes aren't tested by the CI - the PR CI now only tests IPC4 / topology2
2. This has already been mentioned before, but just to make sure - this doesn't move some pipelines in a connected topology to another core, while keeping other pipelines in the same stream on the original core? Are any other pipelines connected to the one, changed here? If yes - they're also running on core 1, aren't they? So before this change pipelines would run on different cores?

johnylin76:
> 2 comments:
> 
> 1. Just to make sure we don't forget this: these changes aren't tested by the CI - the PR CI now only tests IPC4 / topology2

Thanks for the note.

> 2. This has already been mentioned before, but just to make sure - this doesn't move some pipelines in a connected topology to another core, while keeping other pipelines in the same stream on the original core? Are any other pipelines connected to the one, changed here? If yes - they're also running on core 1, aren't they? So before this change pipelines would run on different cores?

It's confirmed that the connected pipelines run on the same core both before and after this change. For cases having other pipelines connected to Spkr (e.g. `GOOGLE_RTC_AUDIO_PROCESSING`), `SPK_PLAYBACK_CORE` will be overwritten to align as the connected pipeline (`DMIC0`), which is already running on core 1.

johnylin76:
@lgirdwood @lyakh is it all good to merge now?

lyakh:
> @lgirdwood @lyakh is it all good to merge now?

No objections against these changes in principle, but just a note, that IPC3 / topology1 aren't actually supported on "main" any more

lgirdwood:
> > @lgirdwood @lyakh is it all good to merge now?
> 
> No objections against these changes in principle, but just a note, that IPC3 / topology1 aren't actually supported on "main" any more

On Intel platforms, other platforms still support IPC3 and topology1.

PR #6930, opened 2023-01-10
copier: fix allocation and configuration size

lyakh:
Copier configuration size must include the variable size array in struct ipc4_copier_gateway_cfg.

### DIFF
--- a/src/include/ipc4/copier.h
+++ b/src/include/ipc4/copier.h
@@ -327,6 +327,11 @@ typedef void (* channel_copy_func)(struct audio_stream __sparse_cache *dst,
 				   int src_channel, int frame_count);
 
 struct copier_data {
+	/*
+	 * struct ipc4_copier_module_cfg actually has variable size, but we
+	 * don't need the variable size array at the end, we won't be copying it
+	 * from the IPC data.
+	 */
 	struct ipc4_copier_module_cfg config;
 	struct comp_dev *endpoint[IPC4_COPIER_MODULE_OUTPUT_PINS_COUNT];
 	struct comp_buffer *endpoint_buffer[IPC4_COPIER_MODULE_OUTPUT_PINS_COUNT];
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -495,7 +495,6 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
 	struct ipc *ipc = ipc_get();
 	struct copier_data *cd;
 	struct comp_dev *dev;
-	size_t size, config_size;
 	int i;
 
 	comp_cl_dbg(&comp_copier, "copier_new()");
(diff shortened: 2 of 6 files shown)


RanderWang:
@lyakh  Now copier doesn't save gateway config (blob data) in variable array since it is a little huge like DMIC blob. Why do we need to save it ? These blob data is sent dai directly from memory windows

lyakh:
> @lyakh Now copier doesn't save gateway config (blob data) in variable array since it is a little huge like DMIC blob. Why do we need to save it ? These blob data is sent to dai directly from memory windows

@RanderWang ah, so it is on purpose? Ok, then we need an explanation there, I'll add a comment. And we don't need that cache-invalidation, I'll remove it. And in principle this PR isn't wrong, right? So it is strange why it's causing PR failures. I'll investigate that.

src/include/ipc4/copier.h
```diff
@@ -350,6 +349,7 @@ struct copier_data {
 	pcm_converter_func converter[IPC4_COPIER_MODULE_OUTPUT_PINS_COUNT];
 	uint64_t input_total_data_processed;
 	uint64_t output_total_data_processed;
+	struct ipc4_copier_module_cfg config;
 };
```
dbaluta, line 354: I wonder how did this work until now?

lgirdwood: Wonder why this passed build ? @lyakh have you been able to build this with GCC ?

lgirdwood (approved):
@mwasko some fixes that may be needed for mtl03

src/include/ipc4/copier.h
lyakh: @dbaluta as @RanderWang explained, it wasn't used. And indeed in some cases that array is large, and if we don't need to keep that configuration, it's ok to not do so, but then we need to make it explicit in the code.
@lgirdwood sure, it builds both ways, because that array isn't declared as variable size, it's declared as size 1 because of problems with some compilers I guess

lyakh:
Now this PR is causing CI failures like https://sof-ci.01.org/sofpr/PR6930/build3364/devicetest/index.html?model=TGLU_RVP_NOCODEC_IPC4ZPH&testcase=check-capture-3times but I think it isn't the fault of this PR that those tests are now failing, it is because it's fixing something. E.g. in the SSP case I narrowed down the failure to calling `dai_ssp_get_properties()` - before this PR this function wasn't called, because `pipeline_comp_trigger()` wasn't using the correct pointer to the DAI. Now the pointer is correct and the function is called. And although the function's name has a "get" in it, it has a side effect... It initialises the SSP properties structure. Same is done with HDA and SDW. So, I think, parameters, that were set during DAI configuration now get reset to initial values. @juimonen do you know what's the right fix for this?

kv2019i (commented):
Code looks good. Not approving until the issue found by CI is sorted out.

lyakh:
Only CI failures now are the known SDW failure https://sof-ci.01.org/sofpr/PR6930/build3399/devicetest/index.html?model=ADLP_RVP_SDW_IPC4ZPH&testcase=check-capture-10sec like in https://sof-ci.01.org/sofpr/PR6785/build3402/devicetest/index.html?model=ADLP_RVP_SDW_IPC4ZPH&testcase=check-capture-10sec and an alsabat failure. There's also a typo in a commit message, let me fix that.

kv2019i (approved):

PR #6931, opened 2023-01-10
memory: Add virtual memory regions gathering and structs

dabekjakub:
Add virtual memory regions info gathering from zephyr and introduce structs agregating this information.
Add proper defines for memory managment.

Signed-off-by: Jakub Dabek <jakub.dabek@intel.com>

### DIFF
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -450,6 +450,11 @@ if (CONFIG_ACE_VERSION_1_5)
 		${SOF_SRC_PATH}/schedule/zephyr_ll.c
 	)
 
+	# Sources for virtual heap management
+	zephyr_library_sources(
+		lib/regions_mm.c
+	)
+
 	zephyr_library_sources_ifdef(CONFIG_CAVS_LPS
 		${SOF_PLATFORM_PATH}/intel/ace/lps_wait.c
 	)
--- a/zephyr/include/sof/lib/regions_mm.h
+++ b/zephyr/include/sof/lib/regions_mm.h
@@ -1,0 +1,40 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright(c) 2022 Intel Corporation. All rights reserved.
+ *
+ * Author: Jakub Dabek <jakub.dabek@intel.com>
+ */
+
+#ifndef ZEPHYR_LIB_REGIONS_MM_H_
+#define ZEPHYR_LIB_REGIONS_MM_H_
+
+#include <adsp_memory.h>
+#include <adsp_memory_regions.h>
+#include <zephyr/drivers/mm/system_mm.h>
+#include <zephyr/sys/mem_blocks.h>
+#include <zephyr/init.h>
+#include <ipc/topology.h>
+
+/*
+ * Struct containing information on virtual memory heap.
+ * Information about allocated physical blocks is stored in
+ * physical_blocks_allocators variable and uses zephyr memblocks api.
+ */
+struct virtual_memory_heap {
+	/* zephyr provided virtual region */
+	struct sys_mm_drv_region *virtual_region;
+	/* physical pages allocators represented in memblocks */
+	struct sys_multi_mem_blocks physical_blocks_allocators;
+	/* SOF memory capability */
+	uint32_t memory_caps;
+};
+
+/* Available externaly array containing all information on virtual heaps
+ * Used to control physical allocations and overall virtual to physicall
+ * mapping on sof side (zephyr handles the actual assigning physical memory
+ * sof only requests it).
+ */
+extern struct virtual_memory_heap
+	vm_heaps[CONFIG_MP_MAX_NUM_CPUS + VIRTUAL_REGION_COUNT];
+
+#endif /* ZEPHYR_LIB_REGIONS_MM_H_ */
--- a/zephyr/lib/regions_mm.c
+++ b/zephyr/lib/regions_mm.c
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/*
+ * Copyright(c) 2022 Intel Corporation. All rights reserved.
+ *
+ * Author: Jakub Dabek <jakub.dabek@intel.com>
+ */
+
+#include <sof/lib/regions_mm.h>
+
+struct virtual_memory_heap
+	vm_heaps[CONFIG_MP_MAX_NUM_CPUS + VIRTUAL_REGION_COUNT];
+
+/**
+ * @brief Fill vm_heaps array with information from zephyr.
+ *
+ * Virtual memory regions calculated in zephyr are translated here
+ * to a struct that will keep all information on current allocations
+ * and virtual to physical mappings that are related to heaps.
+ * System heap is not a part of this information. It only refers to
+ * virtual first heaps.
+ * Has to be initialized after calculations for regions is done in zephyr.
+ */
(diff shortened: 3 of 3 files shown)


marc-hb:
To please checkpatch and fix https://github.com/thesofproject/sof/actions/runs/3886647798/jobs/6631992187 please have a look at brand new
- #6934

marc-hb:
Not a single configuration compiled in https://github.com/thesofproject/sof/actions/runs/3886647797/jobs/6631994411 

Please submit as a Github "draft" and keep it a draft until compilation issues are fixed to avoid spamming reviewers, thanks!

dabekjakub:
I can build the changes localy with no problem that is why i pushed it - was doing checkpatch errors first didn't think it would spam reviewers - sorry.

marc-hb:
> I can build the changes localy with no problem that is why i pushed it -

Configuration differences happen no matter how hard we try to chase them: that's one of the reasons why I start ALL my pull requests as "drafts". It's a no-brainer, should be the default.

dabekjakub:
Current Ci problems seem to be generic - I am currently resolving those offline with CI owners.

zephyr/CMakeLists.txt
```diff
@@ -458,6 +463,7 @@ if (CONFIG_ACE_VERSION_1_5)
 
 	zephyr_include_directories(${SOF_PLATFORM_PATH}/intel/ace/include)
 	zephyr_include_directories(${SOF_PLATFORM_PATH}/meteorlake/include)
+
```
lyakh, line 466: unneeded

zephyr/include/sof/lib/regions_mm.h
```diff
@@ -0,0 +1,35 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright(c) 2022 Intel Corporation. All rights reserved.
+ *
+ * Author: Jakub Dabek <jakub.dabek@intel.com>
+ */
+
```
lyakh, line 7: we probably want header guards here

zephyr/lib/regions_mm.c
```diff
@@ -0,0 +1,49 @@
...
+ * and virtual to physical mappings that are related to heaps.
+ * System heap is not a part of this information. It only refers to
+ * virtual first heaps.
+ * Has to be initialized after calculations for regions is done in zephyr.
+ */
+static int virtual_heaps_init(const struct device *unused)
+{
+	ARG_UNUSED(unused);
+
+	struct sys_mm_drv_region *virtual_memory_regions = { 0 };
```
lyakh, line 27: superfluous initialisation, you overwrite it 1 code line below

zephyr/lib/regions_mm.c
```diff
@@ -0,0 +1,49 @@
...
+
+	for (size_t i = 0;
+		i < CONFIG_MP_MAX_NUM_CPUS + VIRTUAL_REGION_COUNT;
+		i++) {
+		vm_heaps[i].virtual_region = &virtual_memory_regions[i];
+		if (virtual_memory_regions[i].attr & MEM_REG_ATTR_CORE_HEAP)
+			vm_heaps[i].memory_caps =
+				SOF_MEM_CAPS_RAM | SOF_MEM_CAPS_HP | SOF_MEM_CAPS_CACHE;
+
+		if (virtual_memory_regions[i].attr & MEM_REG_ATTR_SHARED_HEAP)
```
lyakh, line 39: You're overwriting `vm_heaps[i].memory_caps` in each of these `if`s, right? Can `.attr` have more than one of these flags set? If so, it seems that `MEM_REG_ATTR_OPPORTUNISTIC_MEMORY` has the highest priority, since it overrides the other two, and `MEM_REG_ATTR_CORE_HEAP` has the lowest priority. So you revert the order and use `else if` to make the priorities clearer.

dabekjakub: Each of those regions is a distinct entity so none of these has a priority perse. Zephyr provides information and i translate it here so we can use it further in allocation.
Those Attributes are work in progress - they are due to change if future implementation uses theme. However elseif is a good point.

lyakh: but you should clearly define (1) whether those attributes are mutually exclusive and (2) what are their priorities. The way you had it implemented in the previous version meant, that they possibly weren't mutually exclusive and the priorities were (from the highest to the lowest):
```
MEM_REG_ATTR_OPPORTUNISTIC_MEMORY
MEM_REG_ATTR_SHARED_HEAP
MEM_REG_ATTR_CORE_HEAP
```
In your today's version you exactly reverted priorities. Is this intended? OTOH, if these attributes are mutually exclusive, then priorities don't matter, then I would suggest to do
```
switch (virtual_memory_regions[i].attr & MEM_REG_ATTR_MASK) {
case MEM_REG_ATTR_SHARED_HEAP:
...
```

zephyr/lib/regions_mm.c
```diff
@@ -0,0 +1,48 @@
...
+/*
+ * Copyright(c) 2022 Intel Corporation. All rights reserved.
+ *
+ * Author: Jakub Dabek <jakub.dabek@intel.com>
+ */
+
+#include <sof/lib/regions_mm.h>
+
+struct virtual_memory_heap
+	vm_heaps[CONFIG_MP_MAX_NUM_CPUS + VIRTUAL_REGION_COUNT] = { {0} };
```
lyakh, line 11: this initialisation is also redundant - global and static data is always initialised to 0 by default

zephyr/lib/regions_mm.c
dabekjakub: Yea they are actually mutually exclusive and there are no priorities - need to change zephyr implementation to enum later or is there any other way You suggest ?

zephyr/lib/regions_mm.c
```diff
@@ -0,0 +1,54 @@
...
+
+	struct sys_mm_drv_region *virtual_memory_regions =
+		(struct sys_mm_drv_region *)sys_mm_drv_query_memory_regions();
+
+	for (size_t i = 0;
+		i < CONFIG_MP_MAX_NUM_CPUS + VIRTUAL_REGION_COUNT;
+		i++) {
+		vm_heaps[i].virtual_region = &virtual_memory_regions[i];
+
+		switch (virtual_memory_regions[i].attr)	{
```
lyakh, line 35: oh, so it's not just a bit set in the attribute, the attribute is actually equal to one of those bits. I thought you'd need to do something like
```
#define MEM_REG_ATTR_MASK (MEM_REG_ATTR_OPPORTUNISTIC_MEMORY | MEM_REG_ATTR_SHARED_HEAP | MEM_REG_ATTR_CORE_HEAP)
switch (virtual_memory_regions[i].attr & MEM_REG_ATTR_MASK) {
```
but if you don't - all the better

lyakh (approved):

kv2019i (approved):

dabekjakub:
Can anyone rerun the CI on this pr ? @lgirdwood @mengdonglin I am 99% sure that it was an issue with CI not with the commit. (CI passed some time back but failed again after cosmetic changes)

kv2019i:
@dabekjakub wrote:
> Can anyone rerun the CI on this pr ? @lgirdwood @mengdonglin I am 99% sure that it was an issue with CI not with the commit. (CI passed some time back but failed again after cosmetic changes)

I think we can merge without a full rerun. The CI failures are known failures in the ADLP SDW configuration and not related to this PR.

PR #6932, opened 2023-01-10
[Skip CI] owners: fix google audio component path

cujomalainey:
Forgot to update this after moving our components to a folder


lgirdwood (approved):
@mwasko fyi - we still have some Intel guys as code owners that no longer work on SOF. Best that we can update the rights Intel folks for each feature.

PR #6933, opened 2023-01-10
[Skip CI] owners: remove departed users

cujomalainey:
Make the codeowners check happy

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -14,10 +14,8 @@ src/include/ipc/**			@thesofproject/steering-committee
 src/include/ipc/**			@randerwang @marcinszkudlinski
 src/include/kernel/**			@thesofproject/steering-committee
 src/include/user/**			@thesofproject/steering-committee
-src/include/sof/debug/gdb/*		@mrajwa
-src/include/sof/audio/kpb.h		@mrajwa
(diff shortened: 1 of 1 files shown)


lgirdwood (approved):
@mwasko @kv2019i need to update Intel users.
@thesofproject/amd @thesofproject/google @thesofproject/nxp @thesofproject/mediatek fyi - if you have any other codeowner cleanup

lgirdwood:
@cujomalainey looks like a conflict with your previous commit.

cujomalainey:
@lgirdwood we should do an org purge before that so we catch any users lingering that might not be relevant. Also I think we should rely on teams over people, but that is just my opinion

cujomalainey:
Also, I have no need to land this, I just wanted to point out the issues

mwasko (changes requested):
By default please replace @mrajwa with @abonislawski as a new codeowner from Intel.

marc-hb:
> By default please replace @mrajwa with @abonislawski as a new codeowner from Intel.

Done in #6949 because this was blocking #6949 (cause the  @mrajwa  account does not exist anymore, I didn't understand this from this PR).

This removal of @mrajwa can be abandoned.

PR #6941, opened 2023-01-11
mtl: Update zephyr and use new register name

softwarecki:
Some registers have been renamed in Zephyrs commit https://github.com/zephyrproject-rtos/zephyr/commit/21f278c04bc258eb344ac5b2123b49d760b5b71d. This PR changes the references from old to new names and update zephyr revision in west.

### DIFF
--- a/src/platform/intel/ace/lib/pm_runtime.c
+++ b/src/platform/intel/ace/lib/pm_runtime.c
@@ -90,8 +90,8 @@ void platform_pm_runtime_disable(uint32_t context, uint32_t index)
 		tr_dbg(&power_tr, "putting prevent on d0i3");
 		pm_policy_state_lock_get(PM_STATE_RUNTIME_IDLE, PM_ALL_SUBSTATES);
 		/* Disable power gating when preventing */
-		DFDSPBRCP.bootctl[PLATFORM_PRIMARY_CORE_ID].bctl |=
-			DFDSPBRCP_BCTL_WAITIPCG | DFDSPBRCP_BCTL_WAITIPPG;
+		DSPCS.bootctl[PLATFORM_PRIMARY_CORE_ID].bctl |=
+			DSPBR_BCTL_WAITIPCG | DSPBR_BCTL_WAITIPPG;
 		break;
 	default:
 		break;
--- a/west.yml
+++ b/west.yml
@@ -43,7 +43,7 @@ manifest:
 
     - name: zephyr
       repo-path: zephyr
-      revision: e4fcb32451c587a3e4ba7f8bf3fc602b16f9652b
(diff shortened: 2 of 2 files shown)


juimonen (commented):
we need the zephyr update in this PR?

juimonen:
talking to myself... ok west is pointing to old working zephyr, so yes.

softwarecki:
> we need the zephyr update in this PR?

@juimonen It is a good question. We can use #ifdef to detect which name is used by Zephyr and then we don't need update zephyrs revision.

juimonen:
@softwarecki I think west update is ok. sorry, I'm personally using upstream (or other than west sha1's) mostly, so didn't understand the issue...

softwarecki:
In the meantime, I have prepared a second solution (https://github.com/softwarecki/sof/commit/8dcb88553f92335156961292448c8eb8f8710725) for the problem. If someone is opposed to update the zephyr version, we can use it

lyakh (commented):
bisection is broken between these two patches, and if you revert the order it will be broken too. What do we do in such cases? Accept the breakage or make it a single commit?

juimonen (approved):
this is fine for me, don't see too much other way of not breaking bisect.

lgirdwood (approved):
Good find - will make it easier to align.

abonislawski (commented):
Any reason why we have platform_pm_runtime_enable() fully on zephyr side but platform_pm_runtime_disable() with extra registry set on sof-side?

Or maybe this is not needed anymore? I can see PCG and PPG bits turning on and off already on zephyr side

tmleman:
> Any reason why we have platform_pm_runtime_enable() fully on zephyr side but platform_pm_runtime_disable() with extra registry set on sof-side?
> 
> Or maybe this is not needed anymore? I can see PCG and PPG bits turning on and off already on zephyr side

Is a safety measure. Just to be sure that core will not enter PG state during WAITI. Right now it is safe to assume that PG and CG is only enabled during D3 and D0i3 entry flow. I think we can delete this. But this should be done in separate PR.

aborisovich (approved):

aborisovich:
It normal that half of the SOF CI tests for ADLP_RVP_SDW_IPC4ZPH fails?

kv2019i:
@aborisovich No, we have a regression affecting SDW configurations (not related to this PR):
https://sof-ci.01.org/sofpr/PR6941/build3352/devicetest/index.html?model=ADLP_RVP_SDW_IPC4ZPH&testcase=verify-kernel-boot-log

@bardliao @ujfalusi @lyakh FYI, I understood this is under work.

kv2019i:
Proceeding with merge.

kv2019i:
@abonislawski Let's handle changes in a separate PR. The register name change seems independent change and we can do it here.

PR #6943, opened 2023-01-11
[SKIP CI] .github/zephyr: add new manifest-check

marc-hb:
Other jobs typically use `west update --narrow` which is faster but also
able to fetch "wild" SHA1s from any random place!  It is useful for
testing unmerged Zephyr commits but risks accepting "invalid" zephyr
commits; this will not.

### DIFF
--- a/.github/workflows/sparse-zephyr.yml
+++ b/.github/workflows/sparse-zephyr.yml
@@ -61,8 +61,9 @@ jobs:
                west update --narrow --fetch-opt=--depth=5
 
       # Not strictly necessary but saves a lot of scrolling in the next step
-      # TODO, research caching:
-      # https://stackoverflow.com/questions/66421411/how-to-run-cached-docker-image-in-github-action
+      # Caching a 12G image is unfortunately not possible:
+      #   https://github.com/ScribeMD/docker-cache/issues/304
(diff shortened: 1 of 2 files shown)


marc-hb:
Demo where the new manifest-check is failing while other builds are passing:
- https://github.com/thesofproject/sof/pull/6942
- https://github.com/thesofproject/sof/actions/runs/3897627232/jobs/6655494808

greg-intel (approved):

paulstelian97 (approved):

.github/workflows/sparse-zephyr.yml
```diff
@@ -61,8 +61,9 @@ jobs:
                west update --narrow --fetch-opt=--depth=5
 
       # Not strictly necessary but saves a lot of scrolling in the next step
-      # TODO, research caching:
-      # https://stackoverflow.com/questions/66421411/how-to-run-cached-docker-image-in-github-action
+      # Caching a 12G image is unfortunately not possible:
```
lgirdwood, line 64: Why is the image 12G ? Are we carrying stuff we dont need anymore like old GCC versions for legacy HW ?

lgirdwood (approved):

PR #6948, opened 2023-01-12
ipc4: increase max pipeline count to 32

RanderWang:
Now each stream is built with at least two pipelines and this can't support more than 8 streams, 6 streams in normal. This results to CI test failure since it will tests all streams simultaneously although current topology is simple.

Now increase it to 32 for future enhancement.
sof-tgl-nocodec

### DIFF
--- a/src/include/ipc4/base_fw.h
+++ b/src/include/ipc4/base_fw.h
@@ -42,7 +42,7 @@
 #define IPC4_MAX_LIBS_COUNT 16
 
 /* Max pipeline count supported by ipc4 */
-#define IPC4_MAX_PPL_COUNT 16
+#define IPC4_MAX_PPL_COUNT 32
 
 enum ipc4_basefw_params {
 	/* Use LARGE_CONFIG_GET to retrieve fw properties as TLV structure


dbaluta:
32 should be enough for anybody :)

plbossart:
> 32 should be enough for anybody :)

32 is already what can be reached by using all HDAudio DMA and splitting the path in host- and dai- pipelines.

ranj063:
@lgirdwood may be a dumb question, but may I ask why we need to restrict the max number of pipelines. I mean, of course there's a limit which is determined by the available memory but given that we have dynamic pipelines, this max seems like an artificial limit. We could in thoery be OK to have more than 32 pipelines in topology assuming not all of them are running at the same time right?

lyakh:
SDW CI failure seems to be a recent regression: https://sof-ci.01.org/sofpr/PR6931/build3353/devicetest/index.html has the same

kv2019i (approved):
This seems harmless, but which CI test actually fails because of this? I cannot see any usage of this value in FW side and nothing on kernel side. Is it some Intel FW test that is failing?

serhiy-katsyuba-intel (approved):

lgirdwood:
Looks like we have a dead DUT. Restart CI.

lgirdwood:
SOFCI TEST

kv2019i:
The one failure left looks to be caused by https://github.com/thesofproject/sof/issues/6864 , so seems we are good with this one. I do think we need to return to the topic of whether we can lift this limitation altogether. It seems there are not very good apparenty reasons for the limit in current code.

PR #6949, opened 2023-01-12
[SKIP CI] CODEOWNERS: be more selective when matching CMakeLists.txt files

marc-hb:
I don't want to be spammed any more every time someone adds or renames a source file.


marc-hb:
> I don't want to be spammed any more every time someone adds or renames a source file.

I tried to "unsubscribe" in every possible way but [Github only "tolerates" force-pushes](https://github.com/zephyrproject-rtos/zephyr/issues/39194) and every force push re-subscribes me every time.

cujomalainey (approved):

lgirdwood (approved):

Issue #3730, opened 2021-01-06
How do we use tags on master branch?


dbaluta:
This questions starts from sof-logger output:

``
ERROR FW ABI 0x3012001 DBG ABI 0x5002001 tag v1.5-rc1-996-g69073c6ae4db src hash 0xc4ec3834 (ldc hash 0xc4ec3834)
``
This output looks to be correct because:

``
GIT_TAG = git describe --tags --abbrev=12 --match v*
``
and 
``
$ git describe --tags --abbrev=12 --match v
v1.5-rc1-996-g69073c6ae4db
``
So, the question is how do we actually use tags? And what's their relation with our releases.

Cc: @xiulipan @marc-hb @paulstelian97 @TE-N-ShengjiuWang 
@lgirdwood @plbossart

paulstelian97:
I think a tag is made for every RC and release, and when doing git describe the most recent tag that points to a commit on the branch is selected. Not sure what happens if two tags point to the same commit though, but I mean I think we shouldn't get into that case anyway.

dbaluta:
@paulstelian97 OK, it could be an explanation. But I have more questions:
- it means that TAG for master doesn't have a meaning? 
- and why do we need both tag and src hash?

lgirdwood:
I would expect we would have the same rules as Linux here, but it looks like we are always adding the commit ID (even when commit is a tag) ?

paulstelian97:
I think src hash is used to identify a specific commit after the tag (tags do NOT move with new commits like branches do).

Tags point to fixed, specific commits. Tags for master are generally releases or RCs at least for this project.

In Git, tags and branches are almost the same, except branches evolve as new commits are added while tags stay attached to a specific commit. Checking out a tag always puts you in "detached HEAD" in fact.

dbaluta:
@paulstelian97 @lgirdwood First thing we need to figure out why do we need the tag information in logs. I suspect that this is an easy way for the reader to identify release version.

lgirdwood:
@ktrzcinx @abonislawski @slawblauciak any idea why the tag info is in logs ?

marc-hb:
I see (at least) three distinct questions:
1. Why would anyone _not_ want git version information in logs that get attached to test runs and bug reports?
2. Fine-tuning of `git describe` parameters. The current `git describe` parameters look OK to me, please be specific about how you would like to change them and why.
[...]

PR #6952, opened 2023-01-13
[WIP] scripts/cmake: update fw versioning process

dnikodem:
FW versioning in SOF is based on git tag future. It was necessary to update the regex command that recognizes the git tag structure for correct fw version recognition:
- SOF_MAJOR
- SOF_MINOR
- SOF_MICRO
- SOF_BUILD

Signed-off-by: Damian Nikodem <damian.nikodem@intel.com>

### DIFF
--- a/scripts/cmake/version.cmake
+++ b/scripts/cmake/version.cmake
@@ -52,35 +52,40 @@ if(EXISTS ${TARBALL_VERSION_SOURCE_PATH})
 	message(STATUS "Found ${TARBALL_VERSION_SOURCE_PATH}")
 else()
 	# execute_process() errors are not fatal by default!
-	execute_process(
-	        COMMAND git describe --tags --abbrev=12 --match v* --dirty
-		WORKING_DIRECTORY ${SOF_ROOT_SOURCE_DIRECTORY}
-		OUTPUT_VARIABLE GIT_TAG
-		OUTPUT_STRIP_TRAILING_WHITESPACE
-	)
-
 	execute_process(COMMAND git log --pretty=format:%h -1
 		WORKING_DIRECTORY ${SOF_ROOT_SOURCE_DIRECTORY}
 		OUTPUT_VARIABLE GIT_LOG_HASH
 		OUTPUT_STRIP_TRAILING_WHITESPACE
 	)
+
+	execute_process(
+	        COMMAND git describe --tags ${GIT_LOG_HASH} --abbrev=12
+		WORKING_DIRECTORY ${SOF_ROOT_SOURCE_DIRECTORY}
+		OUTPUT_VARIABLE GIT_TAG
+		OUTPUT_STRIP_TRAILING_WHITESPACE
+	)
 endif()
 
-if(NOT GIT_TAG MATCHES "^v")
+if(NOT GIT_TAG)
 	message(WARNING
 	  "git describe found ${GIT_TAG} / nothing starting with 'v'. Shallow clone?")
 	set(GIT_TAG "v0.0-0-g0000")
 endif()
 
 message(STATUS "GIT_TAG / GIT_LOG_HASH : ${GIT_TAG} / ${GIT_LOG_HASH}")
+string(REGEX MATCH "v([0-9]+).?([0-9])[\.]?((-hotfix|-rc)?([0-9]+))?-?(([0-9]+))?" ignored "${GIT_TAG}")
 
-string(REGEX MATCH "^v([0-9]+)[.]([0-9]+)([.]([0-9]+))?" ignored "${GIT_TAG}")
 set(SOF_MAJOR ${CMAKE_MATCH_1})
 set(SOF_MINOR ${CMAKE_MATCH_2})
-set(SOF_MICRO ${CMAKE_MATCH_4})
+set(SOF_MICRO ${CMAKE_MATCH_5})
+set(SOF_BUILD ${CMAKE_MATCH_7})
 
 if(NOT SOF_MICRO MATCHES "^[0-9]+$")
 	set(SOF_MICRO 0)
+endif()
+
+if(NOT SOF_BUILD MATCHES "^[0-9]+$")
+	set(SOF_BUILD 0)
 endif()
 
 string(SUBSTRING "${GIT_LOG_HASH}" 0 5 SOF_TAG)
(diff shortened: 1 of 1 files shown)


dabekjakub (approved):

marc-hb (changes requested):
The changes go far beyond updating the regex without explaining why in the commit message. Also, don't remove `--dirty`, it has caught several CI and release problems in the past.

Regular expressions are notoriously buggy: please provide an example of the parsing issue fixed and explain how this fixes it.

dnikodem:
> Regular expressions are notoriously buggy: please provide an example of the parsing issue fixed and explain how this fixes it.

Marc, thank you for your comments,

About "--dirty", I did not know it was useful. I will restore this parameter, thanks.

The main problem here is that we are using many different versions of git tag to marked branch/releases. That's why I decided to change the formula for catching firmware versions. For example, some recent tags used:
- v2.4.1
- releases/mtl/v2.0-hotfix3
- v2.1.1a
- releases/rpl/v1.0
- v1.9.3-tplg2
- v.2.2-rc1

With the previous formula we only considered tags beginning with "v". This didn't work with some of the tags mentioned above. Now, it should work with tags like below:
- v2.0
- v2.0.1
- v2.0-hotfix1
- v2.0-rc1
- releases/mtl/v2.0
- releases/mtl/v2.0.1
- releases/mtl/v2.0-hotfix1

where:
 - SOF_MAJOR=2, 
 - SOF_MINOR=0, 
 - SOF_MICRO=1 (if existing in tag e.g. v2.0.1, v2.0.-rc1 etc., otherwise 0). 
In addition, SOF_BUILD will indicate the commits number since the last known tag (if newest commit was tagged then SOF_BUILD = 0)

marc-hb:
Thanks, I understand a bit better now.

I think we should simply stop parsing git tags, it's too brittle. Everyone and everything should be able to create any temporary (and impossible to parse) tag in their workspace without breaking functionality!

This would also solve better the previously discussed issue
- #3730

`git describe` is still extremely useful; it's still critical to have in the test logs. However no functionality should depend on it.

This exactly how Zephyr solved that problem. On one hand it has `BUILD_VERSION = $(git describe ...)` defined in `zephyr/cmake/gen_version_h.cmake`. `BUILD_VERSION` is not parsed and does not affect functionality.

On the other hand it has [`zephyr/VERSION`](https://github.com/zephyrproject-rtos/zephyr/commits/main/VERSION) which has versions checked in git and manually updated.

The Linux kernel does the same in https://github.com/thesofproject/linux/commits/topic/sof-dev/Makefile

marc-hb:
If we really really wanted to parse `git describe` (we must not), then we should not waste time debugging super error-prone regular expressions. We should instead re-use existing code like for instance Python's standard "packaging" library:

```shell
ipython3

from packaging.version import parse as parse_version

In [16]: [ parse_version(x).release for x in [ 'v2.4.1', 'v2.2.1a', 'v2.2-rc1' ]]
Out[16]: [(2, 4, 1), (2, 2, 1), (2, 2)]

In [17]: [ parse_version(x).pre     for x in [ 'v2.4.1', 'v2.2.1a', 'v2.2-rc1' ]]
Out[17]: [None, ('a', 0), ('rc', 1)]

```

marc-hb:
Separating `git describe` and functionality would also solve some of the problems with shallow clones, see for instance b5219d844ce4, recent 89ddee6961b4, others.

dnikodem:
> > > With the previous formula we only considered tags beginning with "v". This didn't work with some of the tags mentioned above. Now, it should work with tags like below:
> > > 
> > > * v2.0
> > > * v2.0.1
> > > * v2.0-hotfix1
> > > * v2.0-rc1
> > >* releases/mtl/v2.0
> > >* releases/mtl/v2.0.1
> > >* releases/mtl/v2.0-hotfix1
> > 
> > Not keen on the last ones, the v prefix is pretty well understood to mean a released version. We can suffix any desired metadata after the v1.2.3-somesuffix

Thanks for your input Liam, 
The above examples are only taken from our repository.  Based on them, I prepared the appropriate regex formula for capturing firmware versions. 
I agree that the flexibility of creating git tags can make the regex formula insufficient in future. Maybe we should consider some kind of standardization for git tag? 

> > 
> > > where:
> > > 
> > > * SOF_MAJOR=2,
> > > * SOF_MINOR=0,
> > > * SOF_MICRO=1 (if existing in tag e.g. v2.0.1, v2.0.-rc1 etc., otherwise 0).
> > >   In addition, SOF_BUILD will indicate the commits number since the last known tag (if newest commit was tagged then SOF_BUILD = 0)
> > 
> > So this is no longer a build count, but the number of commits after tag ? The former is useful for developers (but I think we have build date/time encoded somewhere if enabled ?), I've no objections to count if it also aligns with how things are done in Windows.

Yes, in my proposal for change, the SOF_BUILD will be points us the number of commits after last known/recognized git tag. Currently, SOF_BUILD is always equal to 1 on SOF (this is why I decided to use this variable)

Another thing I noticed is that currently fw versioning is not working properly on SOF. When creating a fw binary, the process does not properly recognize the last git tag so we need a solution that will help us patch this problem.

However, since the above solution is not perfect, we can plan here or in another discussion how to solve the versioning problem better. What do you think about this?

dnikodem:
> If we really really wanted to parse `git describe` (we must not), then we should not waste time debugging super error-prone regular expressions. We should instead re-use existing code like for instance Python's standard [and well documented](https://peps.python.org/pep-0440/) "packaging" library:
> 
> ```shell
> ipython3
> 
> from packaging.version import parse as parse_version
> 
> In [16]: [ parse_version(x).release for x in [ 'v2.4.1', 'v2.2.1a', 'v2.2-rc1' ]]
> Out[16]: [(2, 4, 1), (2, 2, 1), (2, 2)]
> 
> In [17]: [ parse_version(x).pre     for x in [ 'v2.4.1', 'v2.2.1a', 'v2.2-rc1' ]]
> Out[17]: [None, ('a', 0), ('rc', 1)]
> ```

It looks interesting, however we currently have a problem with fw  ersioning on SOF - it is not working properly. I think my proposal could solve this problem temporarily. In the meantime, we could work out a better working solution. What do you think about this?

dnikodem:
I closed this PR draft - I will prepare a new solution based on a manually changed *.h file containing the current version of FW. New solution will be prepared in the new PR.

marc-hb:
@dnikodem do you still intend to fix this?

dnikodem:
> @dnikodem do you still intend to fix this?

this task is waiting in my queue - I am now unable to specify when I will be able to take care of it again.

marc-hb:
Submitted in #7532. Still needs a bit more testing but should be ready to review.

PR #6966, opened 2023-01-18
scripts/xtensa-build-zephyr.py: checksum [sof_]version.h files

marc-hb:
`git describe` differences happen, especially when trying to optimize cloning and this is for instance what happened in
https://github.com/thesofproject/sof/pull/6950#issuecomment-1396150533 where zephyr tags were fetched on Linux but not on Windows

This also makes plain git version differences more obvious; no need to scroll all the way up and scan the build logs.

### DIFF
--- a/scripts/xtensa-build-zephyr.py
+++ b/scripts/xtensa-build-zephyr.py
@@ -825,7 +825,12 @@ def install_platform(platform, sof_platform_output_dir):
 BIN_NAME = 'zephyr'
 
 CHECKSUM_WANTED = [
-	'*.ri',     # Some .ri files have a non-deterministic signature, others not
+	# Some .ri files have a deterministic signature, others use
+	# a cryptographic salt. Even for the latter a checksum is still
+	# useful to match an artefact with a specific build log.
+	'*.ri',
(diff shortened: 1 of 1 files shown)


marc-hb:
https://github.com/thesofproject/sof/actions/runs/3953435969/jobs/6769706666 shows a `zephyr_version.h` difference which should be fixed by
- #6950

lgirdwood (approved):

dbaluta (approved):

marc-hb:
While @aborisovich 's review is always desired,  this change is 100% OS-independent and the Windows builds in https://github.com/thesofproject/sof/actions/runs/3953435969/jobs/6769706666 show the versions .h files checksummed as expected.

kv2019i (approved):

aborisovich:
> While @aborisovich 's review is always desired,  this change is 100% OS-independent and the Windows builds in https://github.com/thesofproject/sof/actions/runs/3953435969/jobs/6769706666 show the versions .h files checksummed as expected.

Sorry guys, I've been on vacation, returning on Monday :-)

PR #6967, opened 2023-01-19
topology2: sof-mtl-rt711-4ch: specify PLATFORM=mtl

bardliao:
It worked because we defined DMIC_DRIVER_VERSION 3 as the default value. We have to specify PLATFORM=mtl when the default value was removed.

Fixes: #6968


miRoox (approved):
Verified this patch can fix https://github.com/thesofproject/sof/issues/6968

kv2019i (approved):

Issue #6963, opened 2023-01-18
[Need Info] Pipeline creation failing on Apollo Lake / Gemini Lake after booting SOF


coolstar:
Hi there, I am writing an audio driver for a non-Linux OS and am hoping to use SOF (mainly for Gemini Lake chromebooks, but my test hardware is Apollo Lake chromebooks)

The intel-signed APL sof binary does have working audio on my chromebook in Linux, however with my own driver on the other OS, I'm noticing a strange issue on Apollo Lake

After booting SOF (and successfully getting the FW_READY message)  I'm attempting to create a pipeline with the following parameters:

```c
	uint32_t pipeline_id = 1;
	uint32_t comp_id = 6;

	struct sof_ipc_pipe_new pipe_msg;
	bzero(&pipe_msg, sizeof(pipe_msg));

	pipe_msg.hdr.cmd = SOF_IPC_GLB_TPLG_MSG | SOF_IPC_TPLG_PIPE_NEW;
	pipe_msg.hdr.size = sizeof(pipe_msg);

	pipe_msg.comp_id = comp_id;
	pipe_msg.pipeline_id = pipeline_id;
	pipe_msg.sched_id = comp_id - 1;

	pipe_msg.period = 1000;
	pipe_msg.period_mips = 5000;

	pipe_msg.time_domain = SOF_TIME_DOMAIN_TIMER;

	struct sof_ipc_comp_reply reply;
	int status = ipc3_tx_msg(&pipe_msg, sizeof(pipe_msg), &reply, sizeof(reply), FALSE);
	printf("Create Pipe? 0x%x %d\n", status, reply.rhdr.error);
```

However, even though the IPC handler returns success (you can see the status parameter returns 0, SOF is returning -22)

My driver code is working on Comet Lake / Jasper Lake / Tiger Lake / Alder Lake, so not sure if there's anything special needed for APL / GLK. The same issue happens on a Gemini Lake chromebook w/ community-signed firmware (though unfortunately I don't have GLK hardware on hand so it's not as easy for me to test it there)

Also worth noting that the IPC handler sends a PM_GATE message, and that returns successful.

```c
	struct sof_ipc_pm_gate pm_gate;
	bzero(&pm_gate, sizeof(pm_gate));

	/* configure pm_gate ipc message */
	pm_gate.hdr.size = sizeof(pm_gate);
	pm_gate.hdr.cmd = SOF_IPC_GLB_PM_MSG | SOF_IPC_PM_GATE;
[...]

PR #6977, opened 2023-01-20
Apollo Lake needs MCLK 1 for working da7219

coolstar:
Headphone output requires MCLK_ID to be set to 1

Fix for headphone issue noticed in #6963

Tested: https://www.youtube.com/watch?v=mtPHPJKPp6o

### DIFF
--- a/tools/topology/topology1/sof-apl-da7219.m4
+++ b/tools/topology/topology1/sof-apl-da7219.m4
@@ -157,21 +157,21 @@ PCM_PLAYBACK_ADD(HDMI3, 7, PIPELINE_PCM_7)
 # BE configurations - overrides config in ACPI if present
 #
 
-#SSP 5 (ID: 0) with 19.2 MHz mclk with MCLK_ID 0 (unused), 1.536 MHz blck
+#SSP 5 (ID: 0) with 19.2 MHz mclk with MCLK_ID 1 (unused), 1.536 MHz blck
 DAI_CONFIG(SSP, 5, 0, SSP5-Codec,
 	SSP_CONFIG(I2S, SSP_CLOCK(mclk, 19200000, codec_mclk_in),
 		SSP_CLOCK(bclk, 1536000, codec_slave),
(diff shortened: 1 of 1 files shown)


sofci:

gkbldcig:

tools/topology/topology1/sof-apl-da7219.m4
```diff
@@ -157,21 +157,21 @@ PCM_PLAYBACK_ADD(HDMI3, 7, PIPELINE_PCM_7)
 # BE configurations - overrides config in ACPI if present
 #
 
-#SSP 5 (ID: 0) with 19.2 MHz mclk with MCLK_ID 0 (unused), 1.536 MHz blck
+#SSP 5 (ID: 0) with 19.2 MHz mclk with MCLK_ID 1 (unused), 1.536 MHz blck
```
lyakh, line 160: if it *has* to be 1, then it's probably not "unused?"

coolstar: GLK needs to be updated too in that case, since it also has to be 1 there (and is already)

plbossart: no, I think the comment was intentional. The MCLK is only relevant for the headset codec IIRC, it's not used by the amplifier.
and while in theory we could use 2 MCLKs for the two different SSPs, I am not sure this was ever tested.

plbossart: the topology uses SSP5 (speaker - maxim98357a, BE link 0) so this is correct and doesn't need a change.

plbossart (approved):
LGTM thanks @coolstar

IIRC I could never get the headset detection to work on my Reef device, maybe that was the problem. I bricked that device so can't test this unfortunately, that was the first SOF device to ever be tested... 2018?

kv2019i (approved):

kv2019i (approved):
Thanks @coolstar , looks good to go!

kv2019i:
SOFCI TEST

PR #6978, opened 2023-01-20
Revert "hda: hw support for hda chain"

lgirdwood:
Reverts thesofproject/sof#6656

Lots of CI failures in capture tests.


kv2019i:
Not needed, handled via https://github.com/thesofproject/sof/pull/6979

PR #6986, opened 2023-01-23
topology2: Add instance attribute for all classes

ranj063:
In preparation for making it easier to write topology conf files that can be conditionally included without having to worry about conflicting node ID's for objects between the included file and the top-level conf file, modify all classes to include an instance attribute. This means that irrespective of the object type, all objects will be instantiated as follows:

Object.Widget.gain.1 {} or Object.Base.pdm_config.2 {} etc

The instance ID's are typically only meant for the alsatplg compiler to differentiate the nodes in the conf file and are not relevant for the kernel or the firmware. This change will allow the alsatplg compiler to be modified to automatically make the node ID's unique before conditionally including conf files.

### DIFF
--- a/tools/topology/topology2/cavs-mixin-mixout-hda.conf
+++ b/tools/topology/topology2/cavs-mixin-mixout-hda.conf
@@ -87,12 +87,16 @@ Object.PCM {
 	pcm.0 {
 		id 0
 		name 'HDA Analog'
-		Object.Base.fe_dai.'HDA Analog' {}
-		Object.PCM.pcm_caps.playback {
+		Object.Base.fe_dai.1 {
+			name "HDA Analog"
+		}
+		Object.PCM.pcm_caps.1 {
+			direction	"playback"
 			name $ANALOG_PLAYBACK_PCM
 			formats 'S32_LE,S24_LE,S16_LE'
 		}
-		Object.PCM.pcm_caps.capture {
+		Object.PCM.pcm_caps.2 {
+			direction	"capture"
 			name $ANALOG_CAPTURE_PCM
 			formats 'S32_LE,S24_LE,S16_LE'
 		}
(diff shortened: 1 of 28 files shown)


juimonen (approved):

jsarha (commented):
Looks consistent and I think I even have an idea why this is needed, but I have trouble building the typologies from this PR. I even updated my alsa-utils and run west update, but build-tools.sh -T still fails:

 + cmake --build /home/oku/work/sof-zephyr/sof/tools/build_tools -- -j 28 topologies
[0/24] Generating cavs-sdw.tplg
ALSA lib pre-process-object.c:1397:(tplg_construct_object_name) Argument name not set for object 'manifest.sof_manifest'

ALSA lib pre-process-object.c:1715:(tplg_build_object) Failed to construct object name for sof_manifest

ALSA lib pre-process-object.c:1885:(tplg_pre_process_objects) Error building object Base.manifest.sof_manifest

Unable to pre-process configuration
FAILED: topology/topology2/avs-tplg/cavs-sdw.tplg

ranj063:
> Looks consistent and I think I even have an idea why this is needed, but I have trouble building the typologies from this PR. I even updated my alsa-utils and run west update, but build-tools.sh -T still fails:

@jsarha you dont need any alsautils changes with this PR and the tools build is passing in the PR. Are you sure you dont have any local manual changes that are breaking your build?

juimonen:
@jsarha this PR builds for me at least.

kv2019i (approved):

fredoh9:
I can build tools and topologies with the SOF docker image.

ranj063:
SOFCI TEST

jsarha (approved):
Compile problem was just a user error. Fixed with git clean -xdf.

marc-hb:
> This PR is the only change between these two commits.

That's wrong sorry. There was also
- #6987

marc-hb:
False alarm, sorry for the noise. Daily tests landed on a faulty device (internal issue 355)

PR #6990, opened 2023-01-24
cavs: dma: increase HDA DMA period count to 4

kv2019i:
Align HDA DMA platform period count with value used on Intel ACE platforms. This only affects operation with native Zephyr DMA drivers.

### DIFF
--- a/src/platform/intel/cavs/lib/dma.c
+++ b/src/platform/intel/cavs/lib/dma.c
@@ -109,7 +109,7 @@ static const struct dw_drv_plat_data dmac1 = {
 };
 
 /* DMA number of buffer periods */
-#define HDA_DMA_BUFFER_PERIOD_COUNT	2
+#define HDA_DMA_BUFFER_PERIOD_COUNT	4
 
 #if CONFIG_DMA_HW_LLI
 #define DW_DMA_BUFFER_PERIOD_COUNT	4


juimonen (approved):

plbossart:
I for one never understood why 4 periods with 4 smaller buffers are better than 2 periods twice as big...

src/platform/intel/cavs/lib/dma.c
```diff
@@ -109,7 +109,7 @@ static const struct dw_drv_plat_data dmac1 = {
 };
 
 /* DMA number of buffer periods */
-#define HDA_DMA_BUFFER_PERIOD_COUNT	2
+#define HDA_DMA_BUFFER_PERIOD_COUNT	4
```
ranj063, line 112: what problem does it fix or improve @kv2019i

juimonen: mono 8khz "doesn't transfer" without this. I think without this the buffer size is going under some dma transfer limit -> no data moving.

kv2019i: @ranj063 @plbossart We can dig deeper before merging. This is now aligning with intel/ace/lib/dma.c (as I don't know full reasons why it is 4 there, but apparently that works better).

lgirdwood (approved):

src/platform/intel/cavs/lib/dma.c
plbossart: Why @lgirdwood would a change in buffer sizes prevent races? Not following at all.

PR #6992, opened 2023-01-25
Audio: EQ IIR: Use the new module interface

singalsu:
Use the module interface instead of the comp drv interface for the EQ IIR component.


src/audio/Kconfig
```diff
@@ -140,7 +140,9 @@ config COMP_IIR
 	bool "IIR component"
 	select COMP_BLOB
 	default y
+	depends on COMP_MODULE_ADAPTER
 	select MATH_IIR_DF1
+
```
lyakh, line 145: this empty line is quite unusual, don't remember seeing it before - before "help" in Kconfig

src/audio/eq_iir/eq_iir.c
```diff
@@ -504,15 +512,14 @@ static int eq_iir_init_coef(struct sof_eq_iir_config *config,
...
-				     i);
+			comp_info(mod->dev, "eq_iir_init_coef(), ch %d is set to bypass", i);
 			iir_reset_df1(&iir[i]);
 			continue;
 		}
 
 		if (resp >= config->number_of_responses) {
-			comp_cl_info(&comp_eq_iir, "eq_iir_init_coef(), requested response %d exceeds defined",
-				     resp);
+			comp_info(mod->dev, "eq_iir_init_coef(), requested response %d exceeds defined",
```
lyakh, line 521: shouldn't this be at least a warning, or even an error, since it returns an error code?

src/audio/eq_iir/eq_iir.c
```diff
@@ -522,14 +529,13 @@ static int eq_iir_init_coef(struct sof_eq_iir_config *config,
 		if (s > 0) {
 			size_sum += s;
 		} else {
-			comp_cl_info(&comp_eq_iir, "eq_iir_init_coef(), sections count %d exceeds max",
-				     eq->num_sections);
+			comp_info(mod->dev, "eq_iir_init_coef(), sections count %d exceeds max",
+				  eq->num_sections);
```
lyakh, line 533: ditto

src/audio/eq_iir/eq_iir.c
```diff
@@ -587,85 +593,76 @@ static int eq_iir_setup(struct comp_data *cd, int nch)
...
 	if (!cd)
-		goto fail;
-
-	comp_set_drvdata(dev, cd);
+		return -ENOMEM;
 
+	md->private = cd;
 	cd->eq_iir_func = NULL;
 	cd->iir_delay = NULL;
 	cd->iir_delay_size = 0;
```
lyakh, line 620: three lines above are superfluous after an `rzalloc()`, not introduced by this patch though

singalsu:
@lyakh Thanks for review! I just pushed IPC4 fixed version, plays and controls here now OK. I'll address your findings from previous version next.

src/audio/Kconfig
singalsu: Oops, fixed.

src/audio/eq_iir/eq_iir.c
singalsu: Yep, good catch.

src/audio/eq_iir/eq_iir.c
singalsu: Fixed too! These are real errors with no way to recover.

src/audio/eq_iir/eq_iir.c
singalsu: Yep, saving code size is good.

zephyr/wrapper.c
```diff
@@ -297,4 +297,3 @@ int poll_for_register_delay(uint32_t reg, uint32_t mask,
 	}
 	return 0;
 }
-
```
kv2019i, line 300: This is now superfluous.

src/audio/eq_iir/eq_iir.c
```diff
@@ -788,66 +799,45 @@ static int eq_iir_prepare(struct processing_module *mod)
...
 
-	source_period_bytes = audio_stream_period_bytes(&source_c->stream, dev->frames);
-	sink_period_bytes = audio_stream_period_bytes(&sink_c->stream, dev->frames);
-	if (sink_c->stream.size < sink_period_bytes) {
-		comp_err(dev, "eq_iir_prepare(): sink buffer size %d is insufficient < %d",
-			 sink_c->stream.size, sink_period_bytes);
-		buffer_release(sink_c);
-		buffer_release(source_c);
-		return -ENOMEM;
-	}
```
kv2019i, line 911: Commit description doesn't mention why this stream.size check is removed? Is this related to passthrough use?

src/audio/eq_iir/eq_iir.c
```diff
@@ -792,16 +828,24 @@ static int eq_iir_prepare(struct processing_module *mod)
 
 	comp_dbg(dev, "eq_iir_prepare()");
 
+	/* TODO: This is untested, may not work in prepare() */
```
kv2019i, line 831: Hmm, this TODO raises more questions than it answers. We use this code path every time, so isn't it problematic if is not tested...?

kv2019i (commented):
@singalsu Some minor comment/questions inline

lyakh (approved):
approving, subject to addressing comments from others, of course

src/audio/CMakeLists.txt
```diff
@@ -166,7 +166,7 @@ set(mixer_sources ${mixer_src})
 set(src_sources src/src.c src/src_generic.c)
 set(asrc_sources asrc/asrc.c asrc/asrc_farrow.c asrc/asrc_farrow_generic.c)
 set(eq-fir_sources module_adapter/module_adapter.c module_adapter/module/generic.c eq_fir/eq_fir.c eq_fir/eq_fir_generic.c)
-set(eq-iir_sources eq_iir/eq_iir.c)
+set(eq-iir_sources module_adapter/module_adapter.c module_adapter/module/generic.c eq_iir/eq_iir.c)
```
lgirdwood, line 169: @singalsu pls see my plugin branch, I have a fix for this in testbench build. We can merge that incrementally though.

lgirdwood (commented):
@singalsu I assume we have CI topologies that can test this ?

src/audio/eq_iir/eq_iir.c
singalsu: I should have been more specific. Setting different word length for source and sink may not work with module adapter since it's hiding params() currently from clients. The purpose of this params function is the check if IIR has conversion for source -> sink format. If yes, keep it. Otherwise set sink format to source. @ranj063 Do you have thoughts for this?

singalsu: By the the time at prepare() I don't think there can be any more different source and sink format. So this may be useless here.

src/audio/eq_iir/eq_iir.c
singalsu: Module adapter is doing it so no need in client. Yep, it should be in the commit message.

src/audio/eq_iir/eq_iir.c
ranj063: @singalsu this line here in verify_params seems wrong:
https://github.com/thesofproject/sof/blob/478a2bee5c880aafa3d1aab01e7812e202d78357/src/audio/eq_iir/eq_iir.c#L690

the eq_iir_find_func() returns a function pointer but buffer_flag is an uint32_t. Im not sure what the original intention was for the buffer_flag to be but seems pretty useless to me.

src/audio/eq_iir/eq_iir.c
singalsu: Also this change needs to go to first commit.

src/audio/eq_iir/eq_iir.c
singalsu: Yep, that seems to be totally broken as well. Behavior is random depending on function address bits that matches flag definitions. The intention might have been to set BUFF_PARAMS_BUFFER_FMT if address is not NULL.
I feel I should remove this call and the eq_iir_verify_params() function. It will remove the format conversion capability from IIR. SOF v2.2 might need a fix. IPC3 will work with format conversion in host/DAI. In IPC4 topologies the pipelines seem to be fixed 32 bit format and all conversion happens in host/DAI copiers.

Do you agree @ranj063 and @kv2019i ?

singalsu: Oops, I missed that it's `buffer_flag = eq_iir_find_func() ? BUFF_PARAMS_FRAME_FMT : 0;` I'll check with IPC3 topology format conversion in IIR before and after module adapter to help decide what to do.

singalsu: I just checked with git main that with IPC3 and sof-hda-generic-4ch.tplg the 48 kHz capture with `arecord -Dhw:0,6 -f S16_LE -c 4 -r 48000 -d 10 rec.wav` the IIR converts the format from 32 to 16 bit. It gives the best 16 bit quality from 32 bit DAI. The trace evidence is "eq_iir_prepare(), source_format=2, sink_format=0". With this module adapter version of PR with eq_iir_verify_params() called from prepare() the trace is "eq_iir_prepare(), source_format=0, sink_format=0" that means that DAI calls PCM converter and IIR is all 16 bits. The parameters verification in prepare() does no more have  opportunity to volunteer for format conversion.

@lgirdwood @ranj063 @kv2019i It's clear that enabling format conversion from prepare does not work. Should we add to module adapter a params() hook for components those can change stream format with IPC3? Or Drop this feature for now from IPC3 support of git main and see later how this could be re-introduced with IPC4 if needed?

kv2019i: @singalsu I think this could be kept as well. With current IPC4 topologies (and kernel code), the source format is fixed, so the conversion capability is neverused (IIR will always see S32 from the DAI). I currently have a kernel PR in review to change specifically this, so you could define a topology, where DAI provides other formats.  So then at prepare you'd see  also different source formats.

But the other question is whether there is need to have this code in other components. In IPC4, we have a copier at the DAI and the FE PCM, and they both can convert. So I guess the main benefit would be save a bit of compute by pushing the conversion from DAI copier to IIR output. But not sure this is worth the additional code (to maintain and have in the module).

singalsu: Thanks! OK, I can keep it there since seems to be no harm. Though the conversion feature will break with module_adapter since the params() walk handling is not complete.

src/audio/eq_iir/eq_iir.c
```diff
@@ -710,288 +704,211 @@ static int eq_iir_verify_params(struct comp_dev *dev,
...
-	comp_get_copy_limits(source_c, sink_c, &cl);
+/**
+ * \brief Set EQ IIR frames alignment limit.
+ * \param[in,out] source Structure pointer of source.
+ * \param[in,out] sink Structure pointer of sink.
+ */
+static void eq_iir_set_alignment(struct audio_stream *source, struct audio_stream *sink)
+{
+	const uint32_t byte_align = 1;
+	const uint32_t frame_align_req = 1;
```
andrula-song, line 776: why not set 	const uint32_t byte_align = 8;  const uint32_t frame_align_req = 2;?

kv2019i (approved):
Looks good. @singalsu Please check the one comment from Andrula.

zephyr/wrapper.c
singalsu: Yep, fixed now

src/audio/eq_iir/eq_iir.c
singalsu: Done

src/audio/CMakeLists.txt
singalsu: Thanks, good to clean this up. I will do it after this PR for all module adapter components.

src/audio/eq_iir/eq_iir.c
singalsu: Thanks, it's fixed now.

src/audio/eq_iir/eq_iir.c
```diff
@@ -730,6 +764,39 @@ static int eq_iir_get_config(struct processing_module *mod,
...
+	if (ret < 0) {
+		comp_err(mod->dev, "eq_iir_new_blob(), failed IIR setup");
+		return ret;
+	} else if (cd->iir_delay_size) {
+		comp_dbg(mod->dev, "eq_iir_new_blob(), active");
+#if CONFIG_IPC_MAJOR_3
+		cd->eq_iir_func = eq_iir_find_func(source_format, sink_format, fm_configured,
+						   ARRAY_SIZE(fm_configured));
+#endif
+#if CONFIG_IPC_MAJOR_4
```
lyakh, line 783: using `#elif` helps show that IPC3 and IPC4 are mutually exclusive. Elsewhere too of course

src/audio/eq_iir/eq_iir.c
```diff
@@ -788,24 +894,33 @@ static int eq_iir_prepare(struct processing_module *mod)
...
 
 	comp_dbg(dev, "eq_iir_prepare()");
 
+#if CONFIG_IPC_MAJOR_3
 	ret = eq_iir_verify_params(dev, mod->stream_params);
 	if (ret < 0)
 		return ret;
+#endif
+
+#if CONFIG_IPC_MAJOR_4
```
lyakh, line 907: e.g. also here `#elif'

lyakh (commented):
a tiny nitpick

src/audio/eq_iir/eq_iir.c
singalsu: Yep, fix coming

src/audio/eq_iir/eq_iir.c
singalsu: yep, tried to address all of those

PR #6993, opened 2023-01-25
lib_manager: lib_id mismatch prevention

pjdobrowolski:
It is possible that lib_id from IPC would not match with lib_id of module. That can cause NULL pointer and FW crash.

### DIFF
--- a/src/library_manager/lib_manager.c
+++ b/src/library_manager/lib_manager.c
@@ -256,6 +256,11 @@ uint32_t lib_manager_allocate_module(const struct comp_driver *drv,
 	tr_dbg(&lib_manager_tr, "lib_manager_allocate_module() mod_id: 0x%x", ipc_config->id);
 
 	desc = lib_manager_get_library_module_desc(module_id);
+	if (!desc) {
+		tr_err(&lib_manager_tr, "lib_manager_get_library_module_desc() failed: NULL");
+		return 0;
+	}
+
 	mod = (struct sof_man_module *)((char *)desc + SOF_MAN_MODULE_OFFSET(entry_index));
 
 	ret = lib_manager_load_module(module_id, mod, desc);
@@ -311,6 +316,8 @@ struct sof_man_fw_desc *lib_manager_get_library_module_desc(int module_id)
 	struct ext_library *_ext_lib = ext_lib_get();
 	uint8_t *buffptr = (uint8_t *)_ext_lib->desc[lib_id];
 
+	if (!buffptr)
+		return NULL;
 	return (struct sof_man_fw_desc *)(buffptr + SOF_MAN_ELF_TEXT_OFFSET);
 }
 


src/library_manager/lib_manager.c
```diff
@@ -310,7 +316,8 @@ struct sof_man_fw_desc *lib_manager_get_library_module_desc(int module_id)
 	uint32_t lib_id = LIB_MANAGER_GET_LIB_ID(module_id);
 	struct ext_library *_ext_lib = ext_lib_get();
 	uint8_t *buffptr = (uint8_t *)_ext_lib->desc[lib_id];
-
+	if (buffptr == NULL)
+		return -EINVAL;
 	return (struct sof_man_fw_desc *)(buffptr + SOF_MAN_ELF_TEXT_OFFSET);
```
dbaluta, line 321: Something doesn't look right here. Functions expects a return of type (struct sof_man_fw_desc*) but you return -EINVAL.

src/library_manager/lib_manager.c
```diff
@@ -256,6 +256,12 @@ uint32_t lib_manager_allocate_module(const struct comp_driver *drv,
 	tr_dbg(&lib_manager_tr, "lib_manager_allocate_module() mod_id: 0x%x", ipc_config->id);
 
 	desc = lib_manager_get_library_module_desc(module_id);
+
```
ranj063, line 259: no need for a newline

src/library_manager/lib_manager.c
```diff
@@ -310,7 +316,8 @@ struct sof_man_fw_desc *lib_manager_get_library_module_desc(int module_id)
 	uint32_t lib_id = LIB_MANAGER_GET_LIB_ID(module_id);
 	struct ext_library *_ext_lib = ext_lib_get();
 	uint8_t *buffptr = (uint8_t *)_ext_lib->desc[lib_id];
-
+	if (buffptr == NULL)
```
ranj063, line 319: and you need the newline here

src/library_manager/lib_manager.c
```diff
@@ -256,6 +256,12 @@ uint32_t lib_manager_allocate_module(const struct comp_driver *drv,
 	tr_dbg(&lib_manager_tr, "lib_manager_allocate_module() mod_id: 0x%x", ipc_config->id);
 
 	desc = lib_manager_get_library_module_desc(module_id);
+
+	if (desc < 0) {
```
ranj063, line 260: if (!desc) instead

src/library_manager/lib_manager.c
ranj063, line 319: if (!buffptr)
    return NULL;

jxstelter (approved):

dbaluta (approved):

marcinszkudlinski (approved):

dnikodem (approved):

tmleman (approved):

abonislawski (approved):

src/library_manager/lib_manager.c
```diff
@@ -311,6 +316,8 @@ struct sof_man_fw_desc *lib_manager_get_library_module_desc(int module_id)
 	struct ext_library *_ext_lib = ext_lib_get();
 	uint8_t *buffptr = (uint8_t *)_ext_lib->desc[lib_id];
 
+	if (!buffptr)
```
lyakh, line 319: @pjdobrowolski not sure what "lib_id from IPC would not match with lib_id of module" from the commit message means - does it mean, that it can be a valid ID but for a different library or an invalid ID altogether? If the latter - can it cause beyond array bound access? Or is it verified before? @kv2019i maybe we need a follow-up fix up for this PR?

pjdobrowolski: There is no ID validation in FW. If we call module ID which is not available in library we would access to memory with random data.

PR #6995, opened 2023-01-26
Audio: SRC: Make stage buffer size to be a multiple of 8

dbaluta:
sbuf length is empirically found value later used with some instruction which need to addresses to be align to 8-bytes.

See: src/audio/src/src_hifi4.c:

220	/* Load two data samples from two channels */
221	AE_L32X2F24_XC(d0, dp, inc); /* r0, l0 */

In order to help the euristic align sbug length to 8.

Signed-off-by: Daniel Baluta <daniel.baluta@nxp.com>

### DIFF
--- a/src/audio/src/src.c
+++ b/src/audio/src/src.c
@@ -141,7 +141,7 @@ static int src_buffer_lengths(struct comp_dev *dev, struct comp_data *cd,
 	struct src_param *a;
 	int fs_in, fs_out;
 	int source_frames;
-	int r1;
+	int r1, n;
 
 	a = &cd->param;
 	fs_in = cd->source_rate;
@@ -206,8 +206,8 @@ static int src_buffer_lengths(struct comp_dev *dev, struct comp_data *cd,
 		 * variable number of blocks to process per each stage
 		 * there is no equation known for minimum size.
 		 */
-		a->sbuf_length = 2 * nch * stage1->blk_out * r1;
-		a->sbuf_length += a->sbuf_length >> 3;
+		n = 2 * stage1->blk_out * r1;
+		a->sbuf_length = nch * (n + (n >> 3));
 	}
 
 	a->src_multich = a->fir_s1 + a->fir_s2 + a->out_s1 + a->out_s2;
@@ -771,7 +771,12 @@ static int src_params_general(struct comp_dev *dev, struct comp_data *cd,
 		goto out;
 	}
 
-	delay_lines_size = sizeof(int32_t) * cd->param.total;
+	/*
+	 * delay_lines_size is used to compute buffer_start which needs to
+	 * be aligned to 8 bytes as required by some Xtensa
+	 * instructions (e.g AE_L32X2F24_XC)
+	 */
+	delay_lines_size = ALIGN_UP(sizeof(int32_t) * cd->param.total, 8);
 	if (delay_lines_size == 0) {
 		comp_err(dev, "src_params(): delay_lines_size = 0");
 
@@ -792,7 +797,7 @@ static int src_params_general(struct comp_dev *dev, struct comp_data *cd,
 
 	/* Clear all delay lines here */
 	memset(cd->delay_lines, 0, delay_lines_size);
-	buffer_start = cd->delay_lines + cd->param.sbuf_length;
+	buffer_start = cd->delay_lines + ALIGN_UP(cd->param.sbuf_length, 2);
 
 	/* Initialize SRC for actual sample rate */
 	n = src_polyphase_init(&cd->src, &cd->param, buffer_start);


paulstelian97 (approved):
While this looks helpful to me, I'm not 100% sure that making the size be aligned guarantees that the _start_ of the buffer is also properly aligned.

I will approve anyway (it's good to do this) but I'm not sure it will _actually_ solve the original issue.

dbaluta:
@paulstelian97 yes, you are right. The code works because the memory is allocated using:
`rballoc` which uses ```rballoc_align(flags, caps, bytes, PLATFORM_DCACHE_ALIGN)``` and for most platforms PLATFORM_DCACHE_ALIGN is a multiple of 8.

src/audio/src/src.c
```diff
@@ -208,6 +208,7 @@ static int src_buffer_lengths(struct comp_dev *dev, struct comp_data *cd,
 		 */
 		a->sbuf_length = 2 * nch * stage1->blk_out * r1;
 		a->sbuf_length += a->sbuf_length >> 3;
+		a->sbuf_length = ALIGN_UP(a->sbuf_length, 8);
```
iuliana-prodan, line 211: Please add a comment here to explain why this is needed?

iuliana-prodan:
There are some typos in the commit message, please update it.

src/audio/src/src.c
dbaluta: @iuliana-prodan good point. added a comment to better explain this.

lgirdwood:
@singalsu @andrula-song any comments ?

dbaluta:
SOFCI TEST

andrula-song:
should we use rballoc_align to alloc memory for cd->delay_lines in function src_params_general to ensure cd->delay_lines is 8 byte aligned? Since the memory size is a->total = a->sbuf_length + a->src_multich;

dbaluta:
I think we can later do that as an improvement, but for now I don't think it is really necessary:

@andrula-song, we have

``»       cd->delay_lines = rballoc(0, SOF_MEM_CAPS_RAM, delay_lines_size);
``

and rballoc calls ``rballoc_align(flags, caps, bytes, PLATFORM_DCACHE_ALIGN);``

And then:

```
src/platform/amd/rembrandt/include/platform/lib/memory.h:#define PLATFORM_DCACHE_ALIGN	128
src/platform/amd/renoir/include/platform/lib/memory.h:#define PLATFORM_DCACHE_ALIGN	64
src/platform/imx8/include/platform/lib/memory.h:#define PLATFORM_DCACHE_ALIGN	32
src/platform/imx8m/include/platform/lib/memory.h:#define PLATFORM_DCACHE_ALIGN	32
src/platform/imx8ulp/include/platform/lib/memory.h:#define PLATFORM_DCACHE_ALIGN	32
src/platform/intel/ace/include/ace/lib/memory.h:#define PLATFORM_DCACHE_ALIGN		DCACHE_LINE_SIZE
src/platform/intel/cavs/include/cavs/lib/memory.h:#define PLATFORM_DCACHE_ALIGN	DCACHE_LINE_SIZE
src/platform/posix/include/platform/lib/memory.h:#define PLATFORM_DCACHE_ALIGN 64
src/platform/library/include/platform/lib/memory.h:#define PLATFORM_DCACHE_ALIGN	sizeof(void *)
src/platform/mt8186/include/platform/lib/memory.h:#define PLATFORM_DCACHE_ALIGN sizeof(void *)
src/platform/mt8188/include/platform/lib/memory.h:#define PLATFORM_DCACHE_ALIGN sizeof(void *)
src/platform/mt8195/include/platform/lib/memory.h:#define PLATFORM_DCACHE_ALIGN sizeof(void *)
src/platform/baytrail/include/platform/lib/memory.h:#define PLATFORM_DCACHE_ALIGN	sizeof(void *)
src/platform/haswell/include/platform/lib/memory.h:#define PLATFORM_DCACHE_ALIGN	sizeof(void *)

Platforms of interest are already aligning the memory to multiple of 8 bytes (support for baytrail and hsw is being removed).

We still need my change because the address of interest is at:

``»       buffer_start = cd->delay_lines + cd->param.sbuf_length;
``

So, even if cd->delay_lines is forced to be aligned at 8-byte boundry `cd->param.sbuf_length` it also needs to be aligned as current PR does.

File: src/audio/src/src.c (hunk 1 above)
Comment (lgirdwood) [reply src/audio/src/src.c:216]: @singalsu @andrula-song it seems like we need some arch/platform specific config here as this will hit on differently on each target and this is generic C code. e.g.
```c
#define ARCH_SIMD_DATA_ALIGN    8
```

lgirdwood (commented):
Think we need to approach this in generic manner to support all HiFi versions and target configs..

singalsu (approved):
Thanks Daniel, good catch!

src/audio/src/src.c
singalsu: The src_config.h should for HiFiX or other SIMD implementations return the buffer length align constant. We can add it later after merge of this. I assume this fixes the bug that was reported by Iuliana.

andrula-song (approved):

src/audio/src/src.c
dbaluta: @singalsu yes this fixes the bug reported by @iuliana-prodan.

dbaluta:
SOFCI TEST

@lgirdwood  CI stuck?

kv2019i:
@dbaluta Please rebase. We had a regression in topologies (two PRs passed CI on their own, but when both merged, the combination did not work and this broke CI). You changes are on top of the broken commit, so the tools build fails and on-device testing is not done (as there's no binary to test). Sorry for the hassle!

iuliana-prodan (approved):

src/audio/src/src.c
lgirdwood: Ok, lets follow up as an improvement PR.

lgirdwood (approved):

tmleman (approved):

lgirdwood:
@wszypelt @lrudyX good to merge ? CI been pending for a day. Thanks !

wszypelt:
@lgirdwood  
Results should be in within an hour

lgirdwood:
@wszypelt can you share the test result as @dbaluta has no way of reading the the test failure data for SRC. Thanks

wszypelt:
@lgirdwood it looks strange, the KD test passes but there are crush dumps, I will send the logs to @dbaluta and @andrula-song

dbaluta:
@wszypelt i don't this PR touches the KD code. Am I right? @lgirdwood  can we merge this?

src/audio/src/src.c
```diff
@@ -208,6 +208,12 @@ static int src_buffer_lengths(struct comp_dev *dev, struct comp_data *cd,
 		 */
 		a->sbuf_length = 2 * nch * stage1->blk_out * r1;
 		a->sbuf_length += a->sbuf_length >> 3;
+
+		/* sbuf_length is used to compute buffer_start which needs to
+		 * be aligned to 8 bytes as required by some Xtensa
+		 * instructions (e.g AE_L32X2F24_XC)
+		 */
+		a->sbuf_length = ALIGN_UP(a->sbuf_length, 8);
```
singalsu, line 216: Sorry I overlooked this in review and noticed my previous patch made a big mistake. The sbuf_length is not size in bytes but length as samples. The value must be an integer multiple of channel count. Making it multiple of 8 would not work for e.g. 6 channels.

So something like

```
n = 2 * stage1->blk_out * r1;
a->sbuf_length = nch * (n + (n >> 3));
```

@dbaluta Could you please try if above patch works for the bug you found.

singalsu: Still one thing missing. If channels count is odd, then buffer_start would not be 8 bytes aligned. So there's need in src_params_general() for

`delay_lines_size = ALIGN_UP(sizeof(int32_t) * cd->param.total, 8);
`
and get bugger_start to same 8 bytes align

`buffer_start = ALIGN_UP(cd->delay_lines + cd->param.sbuf_length, 8);
`

singalsu: The last macro usage does not seem to work, maybe this instead

`buffer_start = cd->delay_lines + ALIGN_UP(cd->param.sbuf_length, 2); `

dbaluta: @singalsu following change based on your suggestion works for me:

```
--- a/src/audio/src/src.c
+++ b/src/audio/src/src.c
@@ -141,7 +141,7 @@ static int src_buffer_lengths(struct comp_dev *dev, struct comp_data *cd,
        struct src_param *a;
        int fs_in, fs_out;
        int source_frames;
-       int r1;
+       int r1, n;
 
        a = &cd->param;
        fs_in = cd->source_rate;
@@ -206,8 +206,8 @@ static int src_buffer_lengths(struct comp_dev *dev, struct comp_data *cd,
                 * variable number of blocks to process per each stage
                 * there is no equation known for minimum size.
                 */
-               a->sbuf_length = 2 * nch * stage1->blk_out * r1;
-               a->sbuf_length += a->sbuf_length >> 3;
+               n = 2 * stage1->blk_out * r1;
+               a->sbuf_length = nch * (n + (n >> 3));
        }
 
        a->src_multich = a->fir_s1 + a->fir_s2 + a->out_s1 + a->out_s2;
@@ -771,7 +771,7 @@ static int src_params_general(struct comp_dev *dev, struct comp_data *cd,
                goto out;
        }
 
-       delay_lines_size = sizeof(int32_t) * cd->param.total;
+       delay_lines_size = ALIGN_UP(sizeof(int32_t) * cd->param.total, 8);
        if (delay_lines_size == 0) {
                comp_err(dev, "src_params(): delay_lines_size = 0");
 
@@ -792,7 +792,7 @@ static int src_params_general(struct comp_dev *dev, struct comp_data *cd,
 
        /* Clear all delay lines here */
        memset(cd->delay_lines, 0, delay_lines_size);
-       buffer_start = cd->delay_lines + cd->param.sbuf_length;
+       buffer_start = cd->delay_lines + ALIGN_UP(cd->param.sbuf_length, 2);
 
        /* Initialize SRC for actual sample rate */
        n = src_polyphase_init(&cd->src, &cd->param, buffer_start);

```

Anyhow, I'm not sure why this is entirely correct.

We need bufer_start to be aligned to 8 but we compute it as:

```
buffer_start = cd->delay_lines + ALIGN_UP(cd->param.sbuf_length, 2); 
```

singalsu: Also if doing that way, the param.total could be changed to param.total_size_aligned.

dbaluta: @singalsu OK, now I understand the explanation. Will go with my patch above.

singalsu (approved):
Looks good, thanks!

PR #6999, opened 2023-01-27
topology2: fix bt offload pipeline build

juimonen:
Fix bt offload pipeline with instance ids. This was an unfortunate merge error with recent instance id topology2 commit.


lgirdwood (approved):

kv2019i (approved):

marc-hb:
Which commit does this `Fixes:` and how did that escape CI?

marc-hb:
There was "logical" conflict between 4476171620b63928 and ae8ca22a590d3. They both passed separately but not together - and CI never tested them together.

More in #6971.

These two commit SHA1s should definitely have been present in the commit message (which I now finally understand)

PR #7000, opened 2023-01-27
Remove legacy Intel platforms

lyakh:


scripts/docker_build/sof_builder/Dockerfile
```diff
@@ -88,7 +88,7 @@ RUN cd "$HOME" && \
 	git clone $CLONE_DEFAULTS --branch sof-gcc10x $CT_NG_REPO && \
 	cd crosstool-ng && \
 	./bootstrap && ./configure --prefix=`pwd` && make && make install && \
-	for arch in byt hsw apl cnl imx imx8m imx8ulp rn rmb mt8186 mt8195; do \
+	for arch in hsw apl cnl imx imx8m imx8ulp rn rmb mt8186 mt8195; do \
```
plbossart, line 91: can we also remove hsw please? It was never supported...

scripts/qemu-check.sh
```diff
@@ -3,7 +3,7 @@
 # Copyright(c) 2018 Intel Corporation. All rights reserved.
 set -e
 
-SUPPORTED_PLATFORMS=(byt cht bdw hsw apl icl skl kbl cnl imx8 imx8x imx8m)
+SUPPORTED_PLATFORMS=(bdw hsw apl icl skl kbl cnl imx8 imx8x imx8m)
```
plbossart, line 6: and bdw as well btw.

plbossart: skl kbl are also questionable now that I read the whole line

lyakh:
@plbossart this is just the first step, more to come

fredoh9:
Thank you for the update. Yesterday I updated SOF Docker image, every platform has 15-20min build time. It will save the build time a lot.

marc-hb:
> Thank you for the update. Yesterday I updated SOF Docker image, every platform has 15-20min build time. It will save the build time a lot.

You're talking about the toolchains, right?

`installer/GNUmakefile` has always been building all platforms in parallel in a couple minutes on my not very fast workstation.

lyakh:
There are some scripts / tests that don't contain Intel platforms newer than cAVS2.0, i.e. after we complete legacy clean up they won't contain any Intel hardware. Do we need those scripts or are they deprecated? Those that I've found are:
* scripts/qemu-check.sh
* scripts/docker_build/sof_builder/Dockerfile
* tools/fuzzer

Then there's the tone generator, the test for which is configured for Baytrail
* tools/test/topology/test-tone-playback.m4

And one more test set is using an Apollo Lake configuration
* tools/test/topology

lyakh:
some statistics:
```
sof$ git diff HEAD~6 --stat | grep "files changed"
 342 files changed, 84 insertions(+), 40092 deletions(-)
```

lyakh:
"merge commit" only to resolve a conflict to run CI jobs, will be rebased properly before upstreaming of course

src/platform/Kconfig
```diff
@@ -5,7 +5,7 @@ menu "Platform"
 choice
```
kv2019i, line 5: Would be good to add a mention that these platforms will be supported in the stable-v2.2 branch.

installer/GNUmakefile
```diff
@@ -24,13 +24,13 @@ SIGNED_list  ?= tgl tgl-h
 ALIAS_OTHER_KEY_list += adl adl-s rpl rpl-s
 
 # Not supported in the main branch anymore, go to stable-v2.3
-# UNSIGNED_list += bdw
+# UNSIGNED_list += byt bdw cht hsw
```
kv2019i, line 27: I wonder if we could start with one platform and get the merged. This is getting bewildering to review. Here it seems we are adding back support in middle of the series.

tools/coredumper/sof-coredump-reader.py
```diff
@@ -56,7 +56,7 @@ def stdout_print(*args, **kwargs):
 	]]
 ]
 
-# Exception casues:
+# Exception causes:
```
kv2019i, line 59: I can see why you are doing this, but this is bumping up the review effort. I'm mentally scanning for changes that only cover platforms, so this kind of stuff jumps out.

.travis.yml
```diff
@@ -60,21 +57,12 @@ jobs:
         - echo CONFIG_HAVE_AGENT=n > src/arch/xtensa/configs/override/no-agent.config
         - ./scripts/docker-run.sh ./scripts/xtensa-build-all.sh -o no-agent -r $PLATFORM
         - ./scripts/docker-qemu.sh ../sof.git/scripts/qemu-check.sh $PLATFORM
-      env: PLATFORM='byt cht'
+      env: PLATFORM='icl'
```
kv2019i, line 60: After whole series applied, "icl" is still left here.

kv2019i (commented):
Wow, that is a thorough job (and a lot to review). A few minor notes. As a generic comment, I'd like to see a "supported in stable-v2.2" series comment in all the commits to give some more hints on what this is about.

tools/coredumper/sof-coredump-reader.py
marc-hb: +1

.travis.yml
marc-hb: Don't touch `.travis.yml`, it has not been used since we moved to Github Actions. There's another PR to finalize the transfer to Github Actions and remove it.

marc-hb:
> I wonder if we could start with one platform and get the merged. This is getting bewildering to review.

Agree 200%, it's hard to merely scroll up and down. Focus on one or two platforms to set a template, ease the review pain and give all various CIs time to adjust: they should be able to remove all deprecated platforms with a small number of changed lines. Most of this has been done already.

After the first removal is merged and CI back to normal then removing all the other ones should be easy and require barely any review.

tools/coredumper/sof-coredump-reader.py
lyakh: it is related to this patch - it removes "sue" ;-) Obviously, that's why I hit this and other typos fixed in this PR, personally I'd keep these changes - they are tiny and obvious comments, but can trivially be extracted too...

installer/GNUmakefile
lyakh: It's a comment. It's true, I wanted to move it back to not remove them in the first place, but forgot. It has a comment in line 26 that these are the platforms kept in 2.3, so no reason to remove them

lyakh:
> > I wonder if we could start with one platform and get the merged. This is getting bewildering to review.
> 
> Agree 200%, it's hard to merely scroll up and down. Focus on one or two platforms to set a template, ease the review pain and give all various CIs time to adjust: they should be able to remove all deprecated platforms with a small number of changed lines. Most of this has been done already.
> 
> After the first removal is merged and CI back to normal then removing all the other ones should be easy and require barely any review.

we can merge them individually, sure, but that won't make them or the review effort smaller. To avoid redundant reviewing, if after reviewing one of them you notice something that you think is likely to repeat in other patches you can just comment that and stop reviewing. But really, there isn't much to review here. Most of these changes are removing complete files and directories.

marc-hb:
> we can merge them individually, sure, but that won't make them or the review effort smaller.

Not each platform individually. Just one platform first, properly reviewed thanks to a reasonable size and well tested in all CIs over a few days after merge.

Then mass duplication to all the other platforms at once in a giant PR that is barely looked at and rubberstamped.

lgirdwood:
@lyakh can we make this PR byt/cht only as we can see impact of the merge on CI and adjust one by one.

lyakh:
> @lyakh can we make this PR byt/cht only as we can see impact of the merge on CI and adjust one by one.

@lgirdwood sure, I'll update it

lgirdwood:
@lyakh its probably a lot easier to split this PR up into one PR per patch above for review and merging and do one by one.

lyakh:
> @lyakh its probably a lot easier to split this PR up into one PR per patch above for review and merging and do one by one.

@lgirdwood #7080

lgirdwood (approved):

PR #7006, opened 2023-01-31
Topology2: cavs-nocodec: remove multi-stream capture on SSP

bardliao:
multi-stream capture is not needed on SSP port and it will make smart amp topology be more complexity.


plbossart (changes requested):
One bridge too far @bardliao, there is no going-back to passthrough

I also don't get what the relationship between multi-stream capture and Smart-amp can be, they must use different SSPs. The multi-stream capture is only intended on the headset capture path.

PR #7010, opened 2023-01-31
topology2: bt-generic: fix to parameter overrides 

kv2019i:


tools/topology/topology2/platform/intel/bt-generic.conf
```diff
@@ -205,12 +205,12 @@ Object.Base {
...
 Object.PCM {
 	pcm."BT" {
-		name		"Port2"
+		name		$BT_PCM_NAME
 		id 		$BT_PCM_ID
 		direction	"duplex"
 
 		Object.Base.fe_dai.1 {
-			name	"Port2"
+			name	$BT_PCM_NAME
```
ranj063, line 213: you dont even need this. IT will be inferited from the parent pcm object

kv2019i: Ack, will drop.

ranj063 (approved):

juimonen (approved):
lgtm

PR #7011, opened 2023-01-31
ipc4: add IPC4 gateway

serhiy-katsyuba-intel:
Adds IPC4 gateway. IPC4 gateway transfers audio between host and DSP without using DMA: via memory window (audio payload) and IPC messages (commands).

### DIFF
--- a/src/audio/CMakeLists.txt
+++ b/src/audio/CMakeLists.txt
@@ -58,6 +58,11 @@ if(NOT CONFIG_LIBRARY)
 			switch.c
 		)
 	endif()
+	if(CONFIG_IPC4_GATEWAY)
+		add_local_sources(sof
+			ipcgtw.c
+		)
+	endif()
 	if(CONFIG_COMP_DAI)
 		add_local_sources(sof
 			dai-legacy.c
--- a/src/audio/Kconfig
+++ b/src/audio/Kconfig
@@ -33,6 +33,15 @@ config COMP_CHAIN_DMA
 	  help
 	    Chain DMA support in hardware
 
+config IPC4_GATEWAY
+        bool "IPC4 Gateway"
+        default y
+        depends on IPC_MAJOR_4
+        help
+          Select for IPC4 Gateway. IPC4 Gateway is used to transfer audio between
+          host and DSP without using DMA: via memory window (audio payload) and
+          IPC4 messages (set/get/flush commands).
+
 config COMP_ARIA
         bool "ARIA component"
         default n
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -266,6 +266,26 @@ e_buf:
 	return ret;
 }
 
+static enum sof_ipc_stream_direction
+	get_gateway_direction(enum ipc4_connector_node_id_type node_id_type)
+{
+	/* WARNING: simple "% 2" formula that was used before does not work for all
+	 * interfaces: at least it does not work for IPC gateway. But it may also
+	 * does not work for other not yet supported interfaces. And so additional
+	 * cases might be required here in future.
+	 */
+	switch (node_id_type) {
+	/* from DSP to host */
+	case ipc4_ipc_output_class:
+		return SOF_IPC_STREAM_CAPTURE;
+	/* from host to DSP */
+	case ipc4_ipc_input_class:
+		return SOF_IPC_STREAM_PLAYBACK;
+	default:
+		return node_id_type % 2;
+	}
+}
+
 static int init_dai(struct comp_dev *parent_dev,
 		    const struct comp_driver *drv,
 		    struct comp_ipc_config *config,
@@ -361,7 +381,7 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
 	dai_count = 1;
 	node_id = copier->gtw_cfg.node_id;
 	dai_index[dai_count - 1] = node_id.f.v_index;
-	dai.direction = node_id.f.dma_type % 2;
+	dai.direction = get_gateway_direction(node_id.f.dma_type);
 	dai.is_config_blob = true;
 	dai.sampling_frequency = copier->out_fmt.sampling_frequency;
 
@@ -457,6 +477,72 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
 	return 0;
 }
 
+#if CONFIG_IPC4_GATEWAY
+static int create_ipcgtw(struct comp_dev *parent_dev, struct copier_data *cd,
+			 struct comp_ipc_config *config,
+			 const struct ipc4_copier_module_cfg *copier)
+{
+	const struct comp_driver *drv;
+	const struct sof_uuid uuid = {0xa814a1ca, 0x0b83, 0x466c, {0x95, 0x87, 0x2f,
+				      0x35, 0xff, 0x8d, 0x12, 0xe8}};
+	int ret;
+	struct comp_dev *dev;
+
+	drv = ipc4_get_drv((uint8_t *)&uuid);
+	if (!drv)
+		return -EINVAL;
+
+	/* create_endpoint_buffer() uses this value to choose between input and
+	 * output formats in copier config to setup buffer. For this purpose
+	 * IPC gateway should be handled similarly as host gateway.
+	 */
+	config->type = SOF_COMP_HOST;
+
+	ret = create_endpoint_buffer(parent_dev, cd, config, copier, ipc4_gtw_none, false, 0);
+	if (ret < 0)
+		return ret;
+
+	dev = drv->ops.create(drv, config, &copier->gtw_cfg);
+	if (!dev) {
+		ret = -EINVAL;
+		goto e_buf;
+	}
+
+	list_init(&dev->bsource_list);
+	list_init(&dev->bsink_list);
+
+	if (cd->direction == SOF_IPC_STREAM_PLAYBACK) {
+		comp_buffer_connect(dev, config->core, cd->endpoint_buffer[cd->endpoint_num],
+				    PPL_CONN_DIR_COMP_TO_BUFFER);
+		cd->bsource_buffer = false;
+	} else {
+		comp_buffer_connect(dev, config->core, cd->endpoint_buffer[cd->endpoint_num],
+				    PPL_CONN_DIR_BUFFER_TO_COMP);
+		cd->bsource_buffer = true;
+	}
+
+	cd->converter[IPC4_COPIER_GATEWAY_PIN] =
+		get_converter_func(&copier->base.audio_fmt,
+				   &copier->out_fmt,
+				   ipc4_gtw_host, IPC4_DIRECTION(cd->direction));
+	if (!cd->converter[IPC4_COPIER_GATEWAY_PIN]) {
+		comp_err(parent_dev, "failed to get converter for IPC gateway, dir %d",
+			 cd->direction);
+		ret = -EINVAL;
+		drv->ops.free(dev);
+		goto e_buf;
+	}
+
+	cd->endpoint[cd->endpoint_num++] = dev;
+
+	return 0;
+
+e_buf:
+	buffer_free(cd->endpoint_buffer[cd->endpoint_num]);
+	return ret;
+}
+#endif
+
 static int init_pipeline_reg(struct comp_dev *dev)
 {
 	struct copier_data *cd = comp_get_drvdata(dev);
@@ -536,7 +622,7 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
 	node_id = copier->gtw_cfg.node_id;
 	/* copier is linked to gateway */
 	if (node_id.dw != IPC4_INVALID_NODE_ID) {
-		cd->direction = node_id.f.dma_type % 2;
+		cd->direction = get_gateway_direction(node_id.f.dma_type);
 
 		switch (node_id.f.dma_type) {
 		case ipc4_hda_host_output_class:
@@ -573,6 +659,21 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
 				ipc_pipe->pipeline->source_comp = dev;
 
 			break;
+#if CONFIG_IPC4_GATEWAY
+		case ipc4_ipc_output_class:
+		case ipc4_ipc_input_class:
+			if (create_ipcgtw(dev, cd, &dev->ipc_config, copier)) {
+				comp_cl_err(&comp_copier, "unable to create IPC gateway");
+				goto error_cd;
+			}
+
+			if (cd->direction == SOF_IPC_STREAM_PLAYBACK)
+				ipc_pipe->pipeline->source_comp = dev;
+			else
+				ipc_pipe->pipeline->sink_comp = dev;
+
+			break;
+#endif
 		default:
 			comp_cl_err(&comp_copier, "unsupported dma type %x",
 				    (uint32_t)node_id.f.dma_type);
--- a/src/audio/ipcgtw.c
+++ b/src/audio/ipcgtw.c
@@ -1,0 +1,371 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// Copyright 2023 Intel Corporation. All rights reserved.
+
+#include <sof/audio/component_ext.h>
+#include <sof/trace/trace.h>
+#include <sof/lib/uuid.h>
+#include <sof/lib/memory.h>
+#include <sof/ut.h>
+#include <rtos/init.h>
+#include <ipc4/ipcgtw.h>
+#include <ipc4/copier.h>
+
+static const struct comp_driver comp_ipcgtw;
+
+LOG_MODULE_REGISTER(ipcgtw, CONFIG_SOF_LOG_LEVEL);
+
+/* a814a1ca-0b83-466c-9587-2f35ff8d12e8 */
+DECLARE_SOF_RT_UUID("ipcgw", ipcgtw_comp_uuid, 0xa814a1ca, 0x0b83, 0x466c,
+		    0x95, 0x87, 0x2f, 0x35, 0xff, 0x8d, 0x12, 0xe8);
+
+DECLARE_TR_CTX(ipcgtw_comp_tr, SOF_UUID(ipcgtw_comp_uuid), LOG_LEVEL_INFO);
+
+/* Host communicates with IPC gateways via global IPC messages. To address a particular
+ * IPC gateway, its node_id is sent in message payload. Hence we need to keep a list of existing
+ * IPC gateways and their node_ids to search for a gateway host wants to address.
+ */
+struct ipcgtw_data {
+	union ipc4_connector_node_id node_id;
+	struct comp_dev *dev;
+	struct list_item item;
+
+	/* IPC gateway buffer size comes in blob at creation time, we keep size here
+	 * to resize buffer later at ipcgtw_params().
+	 */
+	uint32_t buf_size;
+};
+
+/* List of existing IPC gateways */
+static struct list_item ipcgtw_list_head = { 0 };
+
+static struct comp_dev *ipcgtw_new(const struct comp_driver *drv,
+				   const struct comp_ipc_config *config,
+				   const void *spec)
+{
+	struct comp_dev *dev;
+	struct ipcgtw_data *ipcgtw_data;
+	const struct ipc4_copier_gateway_cfg *gtw_cfg = spec;
+	const struct ipc4_ipc_gateway_config_blob *blob;
+
+	comp_cl_dbg(&comp_ipcgtw, "ipcgtw_new()");
+
+	if (!gtw_cfg->config_length) {
+		comp_cl_err(&comp_ipcgtw, "ipcgtw_new(): empty ipc4_gateway_config_data");
+		return NULL;
+	}
+
+	dev = comp_alloc(drv, sizeof(*dev));
+	if (!dev)
+		return NULL;
+	dev->ipc_config = *config;
+
+	ipcgtw_data = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*ipcgtw_data));
+	if (!ipcgtw_data) {
+		rfree(dev);
+		return NULL;
+	}
+
+	comp_set_drvdata(dev, ipcgtw_data);
+
+	ipcgtw_data->node_id = gtw_cfg->node_id;
+	ipcgtw_data->dev = dev;
+
+	blob = (const struct ipc4_ipc_gateway_config_blob *)
+		((const struct ipc4_gateway_config_data *)gtw_cfg->config_data)->config_blob;
+
+	/* Endpoint buffer is created in copier with size specified in copier config. That buffer
+	 * will be resized to size specified in IPC gateway blob later in ipcgtw_params().
+	 */
+	comp_cl_dbg(&comp_ipcgtw, "ipcgtw_new(): buffer_size: %u", blob->buffer_size);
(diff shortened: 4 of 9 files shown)


src/audio/Kconfig
```diff
@@ -33,6 +33,15 @@ config COMP_CHAIN_DMA
 	  help
 	    Chain DMA support in hardware
 
+config COMP_IPCGTW
```
plbossart, line 36: COMP_IPC_GTW for the rest of us with prescription glasses?

src/audio/copier/copier.c
```diff
@@ -266,6 +266,26 @@ static int create_host(struct comp_dev *parent_dev, struct copier_data *cd,
...
+	 */
+	switch (node_id_type) {
+	/* from DSP to host */
+	case ipc4_ipc_output_class:
+		return SOF_IPC_STREAM_CAPTURE;
+	/* from host to DSP */
+	case ipc4_ipc_input_class:
+		return SOF_IPC_STREAM_PLAYBACK;
+	default:
+		return node_id_type % 2;
```
plbossart, line 285: humm, is this really needed?  I don't get what additional directions we might see beyond capture and playback.

And if you return an enum it should be from the enum list, not based on a formula.

src/audio/copier/copier.c
```diff
@@ -457,6 +477,74 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
...
+
+	if (cd->direction == SOF_IPC_STREAM_PLAYBACK) {
+		comp_buffer_connect(dev, config->core, cd->endpoint_buffer[cd->endpoint_num],
+				    PPL_CONN_DIR_COMP_TO_BUFFER);
+		cd->bsource_buffer = false;
+	} else {
+		comp_buffer_connect(dev, config->core, cd->endpoint_buffer[cd->endpoint_num],
+				    PPL_CONN_DIR_BUFFER_TO_COMP);
+		cd->bsource_buffer = true;
+	}
```
plbossart, line 522: between this code and the commit message, I am a bit lost of the problem statement.

In what cases would we send 'audio' over a SRAM window with IPC messages. We use DMAs for this.

Please clarify the usages that are improved by this solution.

src/audio/ipcgtw.c
```diff
@@ -0,0 +1,395 @@
...
+/* Incoming IPC gateway message */
+struct ipc4_ipc_gateway_cmd_data {
+	/* node_id of the target gateway */
+	union ipc4_connector_node_id node_id;
+	/* Payload (actual size is in the header extension.r.data_size) */
+	uint8_t payload[0];
+};
+
+/* Reply to IPC gateway message */
+struct ipc4_ipc_gateway_cmd_data_reply {
```
plbossart, line 51: can we use one convention?

ipc_gateway
ipc_gtw
ipcgtw

Pick one.

plbossart (commented):
You're missing a clear explanation as to why we would ever send 'audio' over IPC+SRAM channels, instead of using the many DMA channels we have available.

kv2019i (commented):
Code in general looks ok (aside the comments already added). One open is that there are no in-tree users, so this won't even get compile tested now. Enable for some configuration?

src/audio/copier/copier.c
```diff
@@ -457,6 +477,74 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
...
+
+	cd->converter[IPC4_COPIER_GATEWAY_PIN] =
+		get_converter_func(&copier->base.audio_fmt,
+				   &copier->out_fmt,
+				   ipc4_gtw_host, IPC4_DIRECTION(cd->direction));
+	if (!cd->converter[IPC4_COPIER_GATEWAY_PIN]) {
+		comp_err(parent_dev, "failed to get converter for IPC gateway, dir %d",
+			 cd->direction);
+		ret = -EINVAL;
+		goto e_buf;
```
ranj063, line 532: don't you also have to free dev in case of error?

src/audio/ipcgtw.c
```diff
@@ -0,0 +1,397 @@
...
+	} else {
+		comp_cl_warn(&comp_ipcgtw, "ipcgtw_reset(): no buffer found");
+	}
+
+	comp_set_state(dev, COMP_TRIGGER_RESET);
+
+	return 0;
+}
+
+static const struct comp_driver comp_ipcgtw = {
```
ranj063, line 373: pleae don't use the comp_drv ops. Any new module should use the module adapter

serhiy-katsyuba-intel:
> You're missing a clear explanation as to why we would ever send 'audio' over IPC+SRAM channels, instead of using the many DMA channels we have available.

There is a management decision to achieve feature parity with close source ACE FW. Basically, thesofproject should support everything we described in specs, implemented in closed source ACE FW and have covered with tests.

src/audio/copier/copier.c
serhiy-katsyuba-intel: > humm, is this really needed? I don't get what additional directions we might see beyond capture and playback.
> 
> And if you return an enum it should be from the enum list, not based on a formula.

The "% 2" was not my invention, it was used before in original code. However, it worked incorrectly for ipc4_ipc_output_class and ipc4_ipc_input_class. Hence the formula was moved to a separate function with switch added to handle ipc4_ipc_output_class and ipc4_ipc_input_class separately.

Of course, its better to do not use such kind of formulas with enums but have a big switch and return enum value. However, with such correct approach I was not sure what to do with ipc4_gpio_class, ipc4_hda_host_inout_class, etc. What direction should be returned for them? We do not (yet?) use those gateway types. So I decided leave the formula for rest of the gateways as it was working before and added a comment that if the formula did not work properly for  ipc4_ipc_output_class and ipc4_ipc_input_class, it might also not work properly for some other gateways that will be enabled in future.

src/audio/ipcgtw.c
serhiy-katsyuba-intel: I'm consistent here. I use ipcgtw everywhere except for IPC4 structs defined in spec. For those they are declared in C code as in spec but with ipc4_ prefix and with C convension. For example, scruct IpcGatewayCmd in spec is called ipc4_ipc_gateway_cmd in code. Hence, there are ipcgtw in some places (mostly in function names, variables) and ipc_gateway or something else in IPC4 structs and enums.

ranj063:
> There is a management decision to achieve feature parity with close source ACE FW. Basically, thesofproject should support everything we described in specs, implemented in closed source ACE FW and have covered with tests.

This aside, I think it will be useful to mention one or two practical usages for this component

src/ipc/ipc-common.c
```diff
@@ -182,7 +182,7 @@ void ipc_msg_send(struct ipc_msg *msg, void *data, bool high_priority)
 	key = k_spin_lock(&ipc->lock);
 
 	/* copy mailbox data to message if not already copied */
-	if ((msg->tx_size > 0 && msg->tx_size < SOF_IPC_MSG_MAX_SIZE) &&
```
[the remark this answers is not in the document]
serhiy-katsyuba-intel: Fixed in new update.

src/audio/copier/copier.c
```diff
@@ -266,6 +266,26 @@ static int create_host(struct comp_dev *parent_dev, struct copier_data *cd,
 	return ret;
 }
 
+static enum sof_ipc_stream_direction
+	get_gateway_direction(enum ipc4_connector_node_id_type node_id_type)
+{
+	/* WARNING: simple "% 2" formula that was used before does not work for all
+	 * interfaces: at least it does not work for IPC gateway. But it may also
+	 * does not work for other not yet supported interfaces. And so additional
+	 * "case"s might be required here in future.
```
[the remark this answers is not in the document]
serhiy-katsyuba-intel: Fixed in new update.

src/audio/copier/copier.c
serhiy-katsyuba-intel: Fixed in new update. NOTE: the code was copy-pasted from create_host(). That func still has that problem. Probably same for create_dai().

src/audio/ipcgtw.c
serhiy-katsyuba-intel: Yes, but I made IPC gateway similarly to existing host and DAI gateways. They use comp_drv ops API. These are not real components, i.e., those ops do not called from pipeline but from copier itself. So copier_copy() calls endpoint[i]->drv->ops.copy(), same for .params(), .prepare(), .create(), etc. So to use new module adapter API, first, copier, host-zephyr and dai-zephyr (not sure about host-legacy and dai-legacy) have to be modified to use module adapter API. That's why I use existing comp_drv ops API.

src/audio/CMakeLists.txt
```diff
@@ -58,6 +58,11 @@ if(NOT CONFIG_LIBRARY)
 			switch.c
 		)
 	endif()
+	if(CONFIG_COMP_IPCGTW)
```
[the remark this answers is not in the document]
serhiy-katsyuba-intel: Yes, that is just a copier gateway, not a real pipeline component. I chose _COMP_ similarly to CONFIG_COMP_DAI -- another gateway. I see almost all options in Kconfig uses _COMP_, even CONFIG_COMP_BLOB -- also seems not a real component.

src/audio/ipcgtw.c
```diff
@@ -0,0 +1,397 @@
...
+
+static inline struct comp_buffer *get_buffer(struct comp_dev *dev)
+{
+	if (dev->direction == SOF_IPC_STREAM_PLAYBACK) {
+		if (list_is_empty(&dev->bsink_list))
+			return NULL;
+		return list_first_item(&dev->bsink_list, struct comp_buffer, source_list);
+	}
+
+	assert(dev->direction == SOF_IPC_STREAM_CAPTURE);
```
kv2019i, line 186: Not related directly to this PR, but @marcinszkudlinski and @aborisovich , here's one good example for the discussion whether to enable asserts by default in all builds.

- If asserts are not enabled in production builds, this is completely ok and I wouldn't think about it more in a review.

- If asserts are always enabled, then this starts to be more complicated. There's no logic directly on the direction in this function, and this is inline helper so can be called on potentially hot codepaths, so not sure this is needed.

This specific case seems ok in the end (as it's not called from any audio loops directly), but a good example that if we do enable assert across the board, then that will have implications to their usage.

marcinszkudlinski: Yes and no
Asserts should be used in 2 cases
1) when there's absolutely no way to recover - controlled system panic is better than crash
2) in situations that "not gonna happen" - in this particular case the field may get 2 values, we really need to panic if it's set to 3 cause it is a defect (memory corruption?) the sooner we know about it the better. 

And IMHO both cases should be enabled in production

Completely separate matter is performance impact of asserts. 

IDEA - maybe 2 levels of assets? Debug asserts and "hard asserts" ? First ones disabled in production?

kv2019i:
@serhiy-katsyuba-intel wrote:
> @kv2019i wrote:
>> ... One open is that there are no in-tree users, so this won't even get compile tested now. Enable for some configuration?
> 
> ipcgtw.c is added to both zephyr/CMakeLists.txt and audio/CMakeLists.txt. CONFIG_COMP_IPCGTW in Kconfig is 'y' by default. So the code does compile. If fact, I got CI error for TGL without Zephyr compilation on first PR version then forgot to 

Ack, I missed the y by default part (by default enabled for all IPC4 builds). So then this is good.

kv2019i (approved):

src/audio/ipcgtw.c
kv2019i: @marcinszkudlinski Ack on 1+2 but the problem is specifically on the performance aspects. For (2) type of checks, there's no upper limit of how paranoid you can get. But let's continue the discussion in other PRs related to asserts, I just wanted to raise this specific point as a good example to think about.

src/audio/Kconfig
lgirdwood: @serhiy-katsyuba-intel best to say ```CONFIG_IPC4_GATEWAY```  as this make its obvious for everyone its an IPC4 config for gateways.

src/audio/Kconfig
```diff
@@ -33,6 +33,15 @@ config COMP_CHAIN_DMA
...
+        bool "IPC Gateway"
```
lgirdwood, line 37: Best to also say "IPC4 Gateway"

src/audio/Kconfig
```diff
@@ -33,6 +33,15 @@ config COMP_CHAIN_DMA
...
+        default y
+        depends on IPC_MAJOR_4
+        help
+          Select for IPC Gateway. IPC Gateway is used to transfer audio between
+          host and DSP via memory window (audio payload) and IPC messages
+          (set/get/flush commands).
```
lgirdwood, line 43: and I would also make it clear if this means that DMA is not used as I think NXP use a similar solution for IPC3.

src/audio/copier/copier.c
```diff
@@ -457,6 +477,72 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
...
+		return -EINVAL;
+
+	/* create_endpoint_buffer() uses this value to choose between input and
+	 * output formats in copier config to setup buffer. For this purpose
+	 * IPC gateway should be handled similarly as host gateway.
+	 */
+	config->type = SOF_COMP_HOST;
+
+	int ret = create_endpoint_buffer(parent_dev, cd, config, copier, ipc4_gtw_none, false, 0);
+
```
lgirdwood, line 501: extra newline

src/audio/copier/copier.c
```diff
@@ -457,6 +477,72 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
...
+	if (ret < 0)
```
lgirdwood, line 502: It would be good to have some error message here for ease of debug

src/audio/ipcgtw.c
```diff
@@ -0,0 +1,397 @@
...
+	 * to resize buffer later at ipcgtw_params().
+	 */
+	uint32_t buf_size;
+};
+
+/* List of existing IPC gateways */
+static struct list_item ipcgtw_list_head = { 0 };
+
+/* Incoming IPC gateway message */
+struct ipc4_ipc_gateway_cmd_data {
```
lgirdwood, line 43: Shouldn't this be in the IPC4 header file ?

src/audio/ipcgtw.c
```diff
@@ -0,0 +1,397 @@
...
+	/* node_id of the target gateway */
+	union ipc4_connector_node_id node_id;
+	/* Payload (actual size is in the header extension.r.data_size) */
+	uint8_t payload[0];
+};
+
+/* Reply to IPC gateway message */
+struct ipc4_ipc_gateway_cmd_data_reply {
```
lgirdwood, line 51: ditto, re IPC4 header file

src/audio/ipcgtw.c
lgirdwood: @ranj063 @serhiy-katsyuba-intel it sounds like we need to update the comp_driver APIs to remove any pipeline helpers and align the docs and APIs to be called as "low level" after all the high level modules are updated to module API.

src/include/ipc4/ipcgtw.h
```diff
@@ -0,0 +1,49 @@
...
+		struct {
+			uint32_t data_size : 30;
+			uint32_t _reserved_0 : 2;
+		} r;
+	} extension;
+} __packed __aligned(4);
+
+/**< Values of ipc4_ipcgtw_cmd::primary.r.cmd */
+enum {
+	IPCGWCMD_GET_DATA = 1,
```
lgirdwood, line 41: We should prefix this with IPC4_

src/audio/ipcgtw.c
marc-hb: The name `assert` is already used and defined by the C standard, the Zephyr documentation and probably others. They both clearly expect assert to be **disabled in "production"**:
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/assert.h.html
https://docs.zephyrproject.org/1.14.0/reference/kconfig/CONFIG_ASSERT.html

HOWEVER they don't define what "production" is. Is Continuous Integration on the main branch production? Unlike a release branch, I would say Continuous Integration on the main branch is NOT production.

> Ack on 1+2 but the problem is specifically on the performance aspects.

Exactly: the only reason to disable asserts is to save memory or CPU or both = performance. I think "production" implies "performance". So a "performance" build configured for benchmarks and detecting regressions should also disable asserts... even on the main branch!

So the answer to "enabled by default?" question is: BOTH. It depends what gets tested, what is the purpose of the build.

> IDEA - maybe 2 levels of assets? Debug asserts and "hard asserts" ?


Good idea. `ALWAYS_ASSERT()`?

tmleman (approved):

src/audio/Kconfig
serhiy-katsyuba-intel: Updated in new version.

src/audio/Kconfig
serhiy-katsyuba-intel: Updated in new version.

src/audio/copier/copier.c
serhiy-katsyuba-intel: Fixed.

src/audio/ipcgtw.c
serhiy-katsyuba-intel: Fixed in new version.

src/include/ipc4/ipcgtw.h
serhiy-katsyuba-intel: Fixed in new version.

src/audio/copier/copier.c
serhiy-katsyuba-intel: create_endpoint_buffer() (and buffer_alloc() inside it) do log in case of error. So it is not very important to also log an error here. The code was copy-pasted from create_host(), to fix error log it would be better to fix it also in other places where create_endpoint_buffer() is called: create_host() and init_dai(). So I left it as is, hope it's OK.

serhiy-katsyuba-intel:
SOFCI TEST

serhiy-katsyuba-intel:
To whom it may: seems there is a network connectivity issue on CI side. Can we re-trigger CI? Tagging: @wszypelt , @lgirdwood .

src/audio/copier/copier.c
plbossart: @serhiy-katsyuba-intel ping?

lgirdwood (approved):

src/audio/ipcgtw.c
```diff
@@ -0,0 +1,377 @@
...
+		struct ipcgtw_data *data = list_item(item, struct ipcgtw_data, item);
+
+		if (data->node_id.dw == node_id.dw)
+			return data->dev;
+	}
+
+	return NULL;
+}
+
+static inline void audio_stream_copy_bytes_from_linear(const void *linear_source,
```
abonislawski, line 122: I think we already have such func in audio_stream.h? or Im missing something?

https://github.com/thesofproject/sof/blob/df5eaf07cdd83c0928c58dc87f00c58c3e1032c6/src/include/sof/audio/audio_stream.h#LL734C10-L734C10

src/audio/ipcgtw.c
```diff
@@ -0,0 +1,377 @@
...
+		bytes_copied = MIN(bytes, bytes_snk);
+		memcpy_s(snk, bytes_copied, src, bytes_copied);
+		bytes -= bytes_copied;
+		src += bytes_copied;
+		snk = audio_stream_wrap(sink, snk + bytes_copied);
+	}
+}
+
+static inline
+void audio_stream_copy_bytes_to_linear(const struct audio_stream __sparse_cache *source,
```
abonislawski, line 141: same here

abonislawski (changes requested):

src/audio/copier/copier.c
```diff
@@ -457,6 +477,72 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
...
+	int ret;
+	struct comp_dev *dev;
+
+	drv = ipc4_get_drv((uint8_t *)&uuid);
+	if (!drv)
+		return -EINVAL;
+
+	/* create_endpoint_buffer() uses this value to choose between input and
+	 * output formats in copier config to setup buffer. For this purpose
+	 * IPC gateway should be handled similarly as host gateway.
```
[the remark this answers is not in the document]
serhiy-katsyuba-intel: That is a thing implemented in host-zephyr.c (and host-legacy.c) called in sources "comp_host". It is used to transfer audio via DMA between host and DSP. I'm not sure what is a proper name for it. It's not a real component or module like other components/modules: it's comp_driver.ops do not called from pipeline like it is done for other components but are only called from copier functions, so it is copier's gateway (aka endpoint).

src/audio/ipcgtw.c
```diff
@@ -0,0 +1,377 @@
...
+	struct list_item item;
+
+	/* IPC gateway buffer size comes in blob at creation time, we keep size here
+	 * to resize buffer later at ipcgtw_params().
+	 */
+	uint32_t buf_size;
+};
+
+/* List of existing IPC gateways */
+static struct list_item ipcgtw_list_head = { 0 };
```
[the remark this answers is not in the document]
serhiy-katsyuba-intel: Right, however, I think the way it is implemented now is better. Initializing with = {0} indicates the code depends on that 0, leaving default initialization could be interpreted as code does not care of initialization value.
Then, initializing list with {&ipcgtw_list_head, &ipcgtw_list_head} requires a knowledge of how list is actually implemented. Seems calling list_init() is preferred.

src/audio/ipcgtw.c
serhiy-katsyuba-intel: Those in audio_stream.h does work with samples, these are working with bytes. So a separate implementation was required. These new functions could be potentially moved to audio_stream.h to be reused if also required somewhere else, however, I was not sure about it and left them here.

abonislawski (approved):

src/audio/ipcgtw.c
```diff
@@ -0,0 +1,377 @@
...
+		rfree(dev);
+		return NULL;
+	}
+
+	comp_set_drvdata(dev, ipcgtw_data);
+
+	ipcgtw_data->node_id = gtw_cfg->node_id;
+	ipcgtw_data->dev = dev;
+
+	if (!gtw_cfg->config_length) {
```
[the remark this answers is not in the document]
serhiy-katsyuba-intel: Fixed in an updated version.

src/audio/ipcgtw.c
```diff
@@ -0,0 +1,377 @@
...
+	/* Endpoint buffer is created in copier with size specified in copier config. That buffer
+	 * will be resized to size specified in IPC gateway blob later in ipcgtw_params().
+	 */
+	comp_cl_dbg(&comp_ipcgtw, "ipcgtw_new(): buffer_size: %u", blob->buffer_size);
+	ipcgtw_data->buf_size = blob->buffer_size;
+
+	if (!ipcgtw_list_head.next)
+		list_init(&ipcgtw_list_head);
+
+	list_init(&ipcgtw_data->item);
```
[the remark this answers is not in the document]
serhiy-katsyuba-intel: Fixed in an updated version.

src/audio/copier/copier.c
serhiy-katsyuba-intel: IPC Gateway was used with some hardware like SueCreek there was no DMA. IPC Gateway has only dependency on IPC, so during bring-up of new hardware, once IPC is working one can start to run some tests/debugging with audio coming to/from host.

lgirdwood (approved):

src/audio/ipcgtw.c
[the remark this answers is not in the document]
lyakh: too late, but #7059

PR #7013, opened 2023-01-31
topology2: copier: Add a new attribute

ranj063:
The dma_buffer_size is only applicable from the host/dai copiers and can be calculated using the ibs/obs in the base_cfg in the kernel.

So in preparation for removing the dma_buffer_size token from the audio format objects, add a new attribute, deep_buffer_dma_ms, in the copier that will be used to the send the deep buffer DMA size to the kernel.

### DIFF
--- a/tools/topology/topology2/include/common/tokens.conf
+++ b/tools/topology/topology2/include/common/tokens.conf
@@ -180,5 +180,6 @@ Object.Base.VendorToken {
 	"18" {
 		name "sof_tkn_copier"
 		node_type		1980
+		deep_buffer_dma_ms	1981
 	}
 }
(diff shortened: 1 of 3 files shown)


tools/topology/topology2/include/common/tokens.conf
```diff
@@ -180,5 +180,6 @@ Object.Base.VendorToken {
 	"18" {
 		name "sof_tkn_copier"
 		node_type		1980
+		deep_buffer_dma_ms	1981
```
plbossart, line 183: we need to reserve 1984 for 'no privacy'

plbossart (approved):

aiChaoSONG (approved):

RanderWang (approved):

kv2019i (approved):

PR #7014, opened 2023-02-01
Prepare for modules with multiple input/output pins

ranj063:


tools/topology/topology2/include/common/tokens.conf
```diff
@@ -22,6 +22,8 @@ Object.Base.VendorToken {
 		# the widget name that the queue is connected from/to.
 		sink_pin_binding_wname	413
 		src_pin_binding_wname	414
+		num_input_audio_formats	415
+		num_output_audio_formats	416
```
plbossart, line 26: num_sink_audio_formats and num_source_audio_formats for consistency?

tools/topology/topology2/include/components/cavs/src_format.conf
```diff
@@ -1,5 +1,7 @@
 #src format array
 			num_audio_formats 42
+			num_input_audio_formats 42
+			num_output_audio_formats 42
```
plbossart, line 4: the answer is always 42.

tools/topology/topology2/include/common/tokens.conf
```diff
@@ -162,6 +162,7 @@ Object.Base.VendorToken {
 		in_interleaving_style	1906
 		in_fmt_cfg		1907
 		in_sample_type		1908
+		pin_index		1909
```
plbossart, line 165: does this belong in this PR? I don't see the link with input/output

ranj063: It's a token needed for the pin_index attributes in the new classes

dbaluta:
@ranj063 Care to explain a little bit better what is an **output Pin**? An output pin is a data line of a DAI?

ranj063:
> @ranj063 Care to explain a little bit better what is an **output Pin**? An output pin is a data line of a DAI?

@dbaluta this is purely from a module pov. A module could have multiple input and output modules connected to it. The term pin refers to the connection between modules and is used to specify the audio formats required at each end of the connection.

kv2019i (approved):

juimonen (approved):

bardliao (approved):

tools/topology/topology2/include/common/sink_audio_format.conf
```diff
@@ -0,0 +1,137 @@
+# Class definition for sink pin audio format object
```
ujfalusi, line 1: I'm confused.
Don't we have support for pins other than index 0? I would thin that they somehow work and somehow they actually get audio format, no?

Isn't the problem lies that we have only one set of audio format definition, which is global for the whole module and for all of it's pins?

ranj063: @ujfalusi we dont. The only audio formats we use today are for pin 0 and always intput and output together in pairs. Clearly this wont work

ujfalusi (approved):

kv2019i:
One timeout in https://sof-ci.01.org/sofpr/PR7014/build3661/devicetest/index.html , not related to this PR, proceeding with merge.

PR #7016, opened 2023-02-01
[ONLY FOR TEST] Test rimage PR#128

aiChaoSONG:


aiChaoSONG:
Test result is good, and rimage pr merged, close.

PR #7018, opened 2023-02-01
SOF native loadable modules

pjdobrowolski:
This is a PR with FW mechanism of loading native loadable modules. For easier development I used existing intel modules mechanism based on module adapter and library manager. To identify SOF module API module version  is being checked. All modules above version 5.0.0 are SOF modules. Version number is being dictate by IADK modules.


src/audio/module_adapter/iadk/iadk_system_agent.cpp
```diff
@@ -20,7 +20,7 @@
 #include <system_agent_interface.h>
 #include <module_initial_settings_concrete.h>
 #include <iadk_module_adapter.h>
-#include <system_agent.h>
+#include <iadk_system_agent.h>
```
lgirdwood, line 23: Looks like an unrelated rimage update was pulled into this patch.

src/include/sof/audio/module_adapter/library/system_agent.h
```diff
@@ -0,0 +1,37 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright(c) 2022 Intel Corporation. All rights reserved.
+ */
+
+#ifndef _SYSTEM_AGENT_H
+#define _SYSTEM_AGENT_H
```
lgirdwood, line 7: Is this for generic agent or IADK C++ agent ? Need to be careful for partitioning.

src/include/sof/audio/module_adapter/library/system_agent.h
```diff
@@ -0,0 +1,37 @@
...
+    uint32_t log_handle;
+    uint32_t const core_id;
+    uint32_t module_id;
+    uint32_t instance_id;
+    uint32_t module_size;
+    void *module_handle;};
+
+void *system_agent_start(uint32_t entry_point, uint32_t module_id, uint32_t instance_id,
+			 uint32_t core_id, uint32_t log_handle, void *mod_cfg);
+#endif
```
lgirdwood, line 37: Looks line a newline is needed here.

src/include/sof/audio/module_adapter/module/loadable_modules.h
```diff
@@ -0,0 +1,42 @@
...
+// 	(comp_dynamic_module)->ops.cmd = module_adapter_cmd; \
+// 	(comp_dynamic_module)->ops.trigger = module_adapter_trigger; \
+// 	(comp_dynamic_module)->ops.reset = module_adapter_reset; \
+// 	(comp_dynamic_module)->ops.free = module_adapter_free; \
+// 	(comp_dynamic_module)->ops.set_large_config = module_set_large_config;\
+// 	(comp_dynamic_module)->ops.get_large_config = module_get_large_config;\
+// 	(comp_dynamic_module)->ops.get_attribute = module_adapter_get_attribute; \
+// } while (0)
+
+#endif /* __SOF_AUDIO_IADK_MODULES__ */
```
lgirdwood, line 42: comment does not match top comment.

lgirdwood (commented):
I know still WIP, but need to make sure we keep partition between generic C module API and IADK module API. i.e. we should be able to use either API depending on the module.

src/audio/module_adapter/library/native_system_agent.c
```diff
@@ -0,0 +1,29 @@
+#include <errno.h>
```
lgirdwood, line 1: Needs a commit message describing the code being added in this patch.

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -86,11 +86,13 @@ static int iadk_modules_init(struct processing_module *mod)
 	desc = lib_manager_get_library_module_desc(module_id);
 	struct sof_man_module *module_entry =
 		    (struct sof_man_module *)((char *)desc + SOF_MAN_MODULE_OFFSET(0));
+	void *mod_adp;
```
lgirdwood, line 89: ditto commit message.

lgirdwood (commented):
Need some commit messages to help reviewers understand the feature in preparation for feature review.

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -247,9 +304,17 @@ static int iadk_modules_get_configuration(struct processing_module *mod, uint32_
...
+		struct module_interface *mod_in =
+					(struct module_interface *)mod->priv.module_adapter;
+
+		return mod_in->get_configuration(mod, config_id, data_offset_size,
+						 fragment, fragment_size);
+		} else {
+			return iadk_wrapper_get_configuration(mod->priv.module_adapter, config_id,
 					      MODULE_CFG_FRAGMENT_SINGLE, *data_offset_size,
 					      fragment, fragment_size);
+		}
```
lyakh, line 317: indentation

src/include/sof/audio/module_adapter/library/native_system_agent.h
```diff
@@ -0,0 +1,27 @@
...
+ * Author: Pawel Dobrowolski <pawelx.dobrowolski@intel.com>
+ */
+
+#ifndef __NATIVE_SYSTEM_AGENT__
+#define __NATIVE_SYSTEM_AGENT__
+
+#include <sof/audio/module_adapter/module/module_interface.h>
+#include <system_service.h>
+
+struct NativeSystemAgent {
```
lgirdwood, line 14: naming code style.

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -44,6 +46,7 @@ DECLARE_SOF_RT_UUID("iadk_modules", intel_uuid, 0xee2585f2, 0xe7d8, 0x43dc,
 		    0x90, 0xab, 0x42, 0x24, 0xe0, 0x0c, 0x3e, 0x84);
 DECLARE_TR_CTX(intel_codec_tr, SOF_UUID(intel_uuid), LOG_LEVEL_INFO);
 
+static uint8_t is_loadable;
```
lgirdwood, line 49: Whats the use for this ? it seems like we are missing a handle structure for this API that would allow us to carry context across API calls ?

pjdobrowolski: Correct. I had yesterday discussion with @serhiy-katsyuba-intel about it and how to handle it. His suggestion is to add position in module interface with information what kind of library we are dealing with.

src/audio/module_adapter/library/native_system_agent.c
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/*
+ * Copyright(c) 2023 Intel Corporation. All rights reserved.
```
dbaluta, line 3: Please use proper commit message prefix. E.g 

``module_adapter: native_system_agent: Introduce call and interface``

src/include/sof/audio/module_adapter/library/native_system_agent.h
```diff
@@ -0,0 +1,27 @@
...
+	AdspSystemService system_service;
+	uint32_t log_handle;
```
dbaluta, line 16: Lets not use CamelCase naming, unless this is somehow a structure that is taken from a 3rd party header which cannot be modified.

Other than that we can always use `struct adsp_system_service`.

I think adding a small description of what are we trying to do and a small information about the design of the solution would greatly help the review.

src/audio/module_adapter/library/native_system_agent.c
```diff
@@ -0,0 +1,37 @@
...
+				uint32_t core_id, uint32_t log_handle, void *mod_cfg,
+				void *mod_interface, void *mod_uuid)
+{
+	struct NativeSystemAgent native_system_agent;
+
+	native_system_agent.module_id = module_id;
+	native_system_agent.instance_id = instance_id;
+	native_system_agent.core_id = core_id;
+	native_system_agent.log_handle = log_handle;
+	void *system_agent_p = (void *)(&native_system_agent);
```
lyakh, line 31: `void *system_agent_p = &native_system_agent;` would be quite enough

src/audio/module_adapter/library/native_system_agent.c
```diff
@@ -0,0 +1,37 @@
...
+
+	native_create_instance_f ci = (native_create_instance_f)(entry_point);
+
+	return ci(module_id, instance_id, core_id, mod_cfg, NULL, &system_agent_p, &mod_interface,
+		  &mod_uuid);
```
lyakh, line 36: usually you pass pointers to a function either if it's a pointer to a larger object to not send the whole object, or when you expect the function to modify the object, to which the pointer points. Here you pass 3 pointers to local variables and those variables aren't used after the function returns. Wouldn't it be enough to pass those variables directly?

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -80,8 +83,26 @@ static int iadk_modules_init(struct processing_module *mod)
...
+		    (struct sof_man_module *)((char *)desc + SOF_MAN_MODULE_OFFSET(0));
+
+	is_loadable = module_entry->type.lib_type;
+	void *mod_adp;
+
+	/* Check if module is sof loadable */
+	if (is_loadable) {
+		mod_adp = native_system_agent_start(md->module_entry_point, module_id, instance_id,
+						    0, log_handle, &mod_cfg,
+						    (void *)&dev->drv->ops, (void *)dev->drv->uid);
```
lyakh, line 100: type-casting to `void *` of other non-constant pointers shouldn't be needed

src/audio/module_adapter/module/iadk_modules.c
pjdobrowolski: I've changed method discovering what module is being loaded by adding bool variable to processing module. I think that it is cleaner now.

src/include/sof/audio/module_adapter/library/native_system_agent.h
lgirdwood: @mwasko do you have any docs for Native System Agent ?

mwasko: @lgirdwood  there is only documentation for IADK Modules Adapter, but it can be used as a reference:
https://thesofproject.github.io/latest/architectures/firmware/intel/ace/iadk_modules.html

Currently in SOF we support only IADK loadable modules. The code that @pjdobrowolski is implementing is to provide support for SOF loadable modules written in C that implement Module API interface directly.

lgirdwood: ah, got you - so this change allows module API loadable modules alongside IADK loadable modules.

src/audio/module_adapter/library/native_system_agent.c
lgirdwood: ack. @pjdobrowolski can you do this in next uprev. Thanks !

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -80,8 +82,28 @@ static int iadk_modules_init(struct processing_module *mod)
...
-	void *mod_adp = system_agent_start(md->module_entry_point, module_id,
-					   instance_id, 0, log_handle, &mod_cfg);
+	/* Check if native Zephyr lib is loaded */
+	struct sof_man_fw_desc *desc;
+
+	desc = lib_manager_get_library_module_desc(module_id);
+	struct sof_man_module *module_entry =
+		    (struct sof_man_module *)((char *)desc + SOF_MAN_MODULE_OFFSET(0));
+
+	sof_module_build_info *mod_buildinfo = (sof_module_build_info *)(module_entry->segment[0].v_base_addr);
```
lgirdwood, line 92: I think we have a TEXT macro for segment index 0

src/include/sof/audio/module_adapter/module/generic.h
```diff
@@ -207,6 +207,11 @@ struct processing_module {
 	 */
 	bool skip_src_buffer_invalidate;
 
+	/*
+	 * flag to insure that module is loadable
+	 */
+	bool is_loadable;
```
lgirdwood, line 213: Is this used in the previous commit ? if so, needs squashed.

src/include/sof/audio/module_adapter/library/native_system_agent.h
pjdobrowolski: yes indeed. I've called it native because I had to came out with unique name, not just another sof_system_agent.

src/audio/module_adapter/library/native_system_agent.c
pjdobrowolski: done

src/audio/module_adapter/library/native_system_agent.c
pjdobrowolski: done

src/audio/module_adapter/module/iadk_modules.c
pjdobrowolski: done

src/audio/module_adapter/module/iadk_modules.c
pjdobrowolski: done

src/audio/module_adapter/module/iadk_modules.c
pjdobrowolski: done

src/include/sof/audio/module_adapter/library/native_system_agent.h
pjdobrowolski: @dbaluta I use existing structure with such naming. Do I have to refactor that structure in this PR or do it in seperate one?

src/include/sof/audio/module_adapter/module/generic.h
pjdobrowolski: done

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -8,7 +8,9 @@
 #include <sof/audio/module_adapter/module/iadk_modules.h>
```
kv2019i, line 8: minor nit: cSOF is a term that has no meaning in SOF project, so please just use "SOF module".

kv2019i (approved):
No major issues with this. More documentation would be nice, but https://thesofproject.github.io/latest/architectures/firmware/intel/ace/iadk_modules.html provides context at least.

src/include/sof/audio/module_adapter/library/native_system_agent.h
lgirdwood: I think best to make the change and then we can merge.

lgirdwood (commented):
Still some small comments to fix, I would do the name change suggested by @dbaluta as this clears all the review opens.

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -261,8 +330,15 @@ static int iadk_modules_get_configuration(struct processing_module *mod, uint32_
...
-	return iadk_wrapper_set_processing_mode(mod->priv.module_adapter, mode);
+{	if (mod->is_loadable) {
+		struct module_interface *mod_in =
+					(struct module_interface *)mod->priv.module_adapter;
+
+		return mod_in->set_processing_mode(mod, mode);
+	} else {
+		return iadk_wrapper_set_processing_mode(mod->priv.module_adapter, mode);
+	}
+
```
lyakh, line 341: superfluous empty line

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -226,11 +279,18 @@ static int iadk_modules_set_configuration(struct processing_module *mod, uint32_
...
+		struct module_interface *mod_in =
+					(struct module_interface *)mod->priv.module_adapter;
+
+		return mod_in->set_configuration(mod, config_id, pos, data_offset_size, fragment,
+						 fragment_size, response, response_size);
+	} else {
+		return iadk_wrapper_set_configuration(mod->priv.module_adapter, config_id, pos,
+						      data_offset_size, fragment, fragment_size,
+						      response, response_size);
+	}
```
lyakh, line 292: in cases, where you `return` in the `if` clause, you don't need an `else`

src/audio/module_adapter/module/iadk_modules.c
pjdobrowolski: I don't get it how it should be proper.

src/audio/module_adapter/module/iadk_modules.c
pjdobrowolski: done

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -125,8 +160,15 @@ static int iadk_modules_prepare(struct processing_module *mod)
 	comp_info(dev, "iadk_modules_prepare()");
 
 	/* Call module specific prepare function if exists. */
-	ret = iadk_wrapper_prepare(mod->priv.module_adapter);
-	return 0;
+	if (mod->is_loadable) {
+		struct module_interface *mod_in =
+					(struct module_interface *)mod->priv.module_adapter;
+
+		ret = mod_in->prepare(mod);
```
lgirdwood, line 167: Don't we have module API wrappers here that check for NULL ptrs before the dereference ?

zephyr/CMakeLists.txt
```diff
@@ -613,6 +613,8 @@ zephyr_library_sources_ifdef(CONFIG_LIBRARY_MANAGER
 
 zephyr_include_directories_ifdef(CONFIG_INTEL_MODULES
 	${SOF_SRC_PATH}/include/sof/audio/module_adapter/iadk/
+	${SOF_SRC_PATH}/include/sof/audio/module_adapter/library/
+
```
lgirdwood, line 617: newline not needed ?

lgirdwood (commented):
Just small opens for me.

src/audio/module_adapter/module/iadk_modules.c
lgirdwood: ```ret = module_adapter_prepare(mod);``` ?

pjdobrowolski: In SOF modules we don't you module adapter but we call modules directly without component api wrapper.

zephyr/CMakeLists.txt
pjdobrowolski: done

src/audio/module_adapter/module/iadk_modules.c
pjdobrowolski: I'm not sure about check for NULL ptrs. There are some macros calling priv data but not suitable for SOF modules.

lgirdwood: Not understanding why we dont need to call the module API and use the op ptr directly here ?
The call does a lot more than cal the op. In fact, it does need to check the op ptr before use if prepare() op is optional.

```c
int module_prepare(struct processing_module *mod)
{
	int ret;

	comp_dbg(dev, "module_prepare() start");

	if (mod->priv.state == MODULE_IDLE)
		return 0;
	if (mod->priv.state < MODULE_INITIALIZED)
		return -EPERM;

	ret = md->ops->prepare(mod);
	if (ret) {
		comp_err(dev, "module_prepare() error %d: module specific prepare failed, comp_id %d",
			 ret, dev_comp_id(dev));
		return ret;
	}

	/* After prepare is done we no longer need runtime configuration
	 * as it has been applied during the procedure - it is safe to
	 * free it.
	 */
	if (md->cfg.data)
		rfree(md->cfg.data);

	md->cfg.avail = false;
	md->cfg.data = NULL;

	md->state = MODULE_IDLE;
	comp_dbg(dev, "module_prepare() done");

	return ret;
}
```

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -226,11 +283,17 @@ static int iadk_modules_set_configuration(struct processing_module *mod, uint32_
...
+					(struct module_interface *)mod->priv.module_adapter;
+
+		return mod_in->set_configuration(mod, config_id, pos, data_offset_size, fragment,
+						 fragment_size, response, response_size);
+	}
 	return iadk_wrapper_set_configuration(mod->priv.module_adapter, config_id, pos,
 					      data_offset_size, fragment, fragment_size,
 					      response, response_size);
 }
-
```
lyakh, line 233: drive-by, please, drop

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -247,9 +310,17 @@ static int iadk_modules_get_configuration(struct processing_module *mod, uint32_
...
+
+		return mod_in->get_configuration(mod, config_id, data_offset_size,
+						 fragment, fragment_size);
+	}
 	return iadk_wrapper_get_configuration(mod->priv.module_adapter, config_id,
-					      MODULE_CFG_FRAGMENT_SINGLE, *data_offset_size,
-					      fragment, fragment_size);
+					      MODULE_CFG_FRAGMENT_SINGLE,
+					      *data_offset_size, fragment,
+					      fragment_size);
```
lyakh, line 323: no change here, right? Better avoid unneeded unrelated changes, they only make reviews more difficult, we spend time trying to identify what you're changing here

src/audio/module_adapter/module/iadk_modules.c
pjdobrowolski: "Not understanding why we dont need to call the module API and use the op ptr directly here ?" Because that op is delivered by module itself. Mechanism is the same as
`static struct module_interface up_down_mixer_interface = {
	.init  = up_down_mixer_init,
	.prepare = up_down_mixer_prepare,
	.process = up_down_mixer_process,
	.reset = up_down_mixer_reset,
	.free = up_down_mixer_free
};`

src/audio/module_adapter/module/iadk_modules.c
pjdobrowolski: ok

src/audio/module_adapter/module/iadk_modules.c
lgirdwood: Sorry, I'm confused. This change bypasses the state management in the module API by calling the op directly.
The module adpater should be the base class here and the iadk the derived class. i.e. the extra features the iadk provides are built on top of the module adaptor and NOT alongside the module adaptor.

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -80,8 +83,33 @@ static int iadk_modules_init(struct processing_module *mod)
...
+
+	void *mod_adp;
+
+	/* Check if module is FDK*/
+	if (mod_buildinfo->api_version_number.fields.major < _MAJOR_API_MODULE_VERSION) {
+		mod_adp = system_agent_start(md->module_entry_point, module_id,
+					     instance_id, 0, log_handle, &mod_cfg);
+	} else {
+		/* If not start agent for sof loadable */
+		mod->is_loadable = true;
```
ranj063, line 108: This is misleading isnt it? Whether a module is built with IADK or is a native SOF module, it is loadable nevertheless right? Do you want to call this mod->is_native_sof?

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -102,7 +130,14 @@ static int iadk_modules_init(struct processing_module *mod)
 	md->mpd.out_buff_size = src_cfg->obs;
 
 	/* Call module specific init function if exists. */
-	ret = iadk_wrapper_init(mod->priv.module_adapter);
+	if (mod->is_loadable) {
```
ranj063, line 133: it is also quite confusing to use the iadk_module_init() and this whole file in general for non-IADK modules as well. If it is generic, maybe we need to start with renaming the file and the functions first

src/audio/module_adapter/module/iadk_modules.c
ranj063: @pjdobrowolski what you are missing is the call to module_init() instead of mod_in->init() Once you do this you can call all the other wrappers such as module_prepare() as well, right?

src/include/sof/audio/module_adapter/module/generic.h
ranj063, line 213: why doesnt this apply to IADK modules. Aren't they loadable also?

kv2019i:
@pjdobrowolski Can you address the comments? I'm good with the PR, but we need sign-off from other reviewers.

src/include/sof/audio/module_adapter/library/module_api_ver.h
```diff
@@ -0,0 +1,34 @@
...
+#ifndef __MODULE_API_VER_H__
+#define __MODULE_API_VER_H__
+
+/*
+ *  Api version 5.0.0 for sof loadable modules
+ */
+
+#define _MAJOR_API_MODULE_VERSION 5
+#define _MIDDLE_API_MODULE_VERSION 0
+#define _MINOR_API_MODULE_VERSION 0
```
lgirdwood, line 17: We should align the prefix with the structures i.e. ```SOF_MODULE_API_VERSION_MAJOR``` that way all module API macros and APIS will start with sof_module_api_ or similar i.e. it obvious to the developer using the API .

src/audio/module_adapter/module/iadk_modules.c
lgirdwood: @pjdobrowolski any update here ? iadk_modules must call the lower base class module API. This ensure we have conssitency across all state machines and de-duplication..

lgirdwood (commented):
Need the module base api/class usage resolved.

sys-pt1s:

lgirdwood:
@pjdobrowolski any update - need to unblock.

pjdobrowolski:
@ranj063 @lgirdwood Please look at WIP code of loadable up_down_mix module https://github.com/abonislawski/sof/commits/sof_module_build_rebase . It might clear up a little how we are passing pointers to module api. This code is working prototype where we can build loadable separate .ri binary. I'm preparing documentation to deliver full information about that solution.

src/audio/module_adapter/library/native_system_agent.c
```diff
@@ -0,0 +1,71 @@
...
+		.system_service = {
+			.LogMessage  = SystemServiceLogMessage,
+			.SafeMemcpy = SystemServiceSafeMemcpy,
+			.SafeMemmove = SystemServiceSafeMemmove,
+			.VecMemset = SystemServiceVecMemset,
+			.NotificationCreate = SystemServiceCreateNotification,
+			.NotificationSend = SystemServiceSendNotificationMessage,
+			.GetInterface = SystemServiceGetInterface
+		},
+	};
```
lyakh, line 53: one TAB too many in lines 44-53

src/audio/module_adapter/library/native_system_agent.c
```diff
@@ -0,0 +1,71 @@
...
+	native_sys_agent.module_id = module_id;
+	native_sys_agent.instance_id = instance_id;
+	native_sys_agent.core_id = core_id;
+	native_sys_agent.log_handle = log_handle;
+
+	void *system_agent_p = &native_sys_agent;
+
+	*sys_service = &native_sys_agent.system_service;
+
+	native_create_instance_f ci = (native_create_instance_f)(entry_point);
```
lyakh, line 68: superfluous parentheses around `entry_point`. Same as `(int)x`

lyakh (approved):

lyakh:
> Need the module base api/class usage resolved.

@pjdobrowolski I think this comment (also explained in https://github.com/thesofproject/sof/pull/7018#discussion_r1135885442) is still unresolved, right?

pjdobrowolski:
@lyakh I've shared repo with completed loadable modules integration (FW+build scripts + ported module) yesterday which explains full flow and why wrappers are not the way how it is designed to work.

lyakh:
> @lyakh I've shared repo with completed loadable modules integration (FW+build scripts + ported module) yesterday which explains full flow and why wrappers are not the way how it is designed to work. Here is up today version https://github.com/abonislawski/sof/commits/sof_module_build_rebase2

Sorry, couldn't find that explanation there upon a quick scan, could you link to it directly?

pjdobrowolski:
Ok, The concept behind loadable modules is that modules are not being changed much and point of exchanging information is defined in module entry point function. FW is passing native system services, and required module data and module is returning pointer to module interface.

Loadable modules just like build in modules have defined module interface https://github.com/abonislawski/sof/blob/sof_module_build_rebase2/lmdk/modules/up_down_mixer/up_down_mixer.c#L438
which is being returned in call off module entry point defined in module manifest and returned https://github.com/abonislawski/sof/blob/sof_module_build_rebase2/lmdk/include/loadable_processing_module.h#L109 .

FW when receive that pointer just cast it to know structure. Just like in modules build in we expect that some data is not filled up.

src/include/sof/audio/module_adapter/library/module_api_ver.h
kv2019i: @pjdobrowolski This is still not addressed, right? Please add a comment if you think this should not be changed, otherwise it's impossible to track the PR status.

src/audio/module_adapter/module/iadk_modules.c
kv2019i: Not sure @pjdobrowolski you are talking about different things, but I tried to recap, and it would seem this is still a valid concern. When module adapter is used with built-in modules, all calls go via common wrapppers (e.g. module_prepare()). I don't see why we wouldn't use these common wrappers also when the module is loaded.

I mean, e.g. for upmix, it implements prepare() for sure, but when it is built in, we still call upmix prepare via module_prepare(), right?

kv2019i (commented):
@pjdobrowolski Can you address the open questions, I think those still need to be addressed before we can merge.

src/audio/module_adapter/library/native_system_agent.c
```diff
@@ -0,0 +1,71 @@
...
+
+/* The create_instance_f is a function call type known in module. The module entry_point
+ * points to this type of function which starts module creation.
+ */
+
+void SystemServiceLogMessage(AdspLogPriority log_priority, uint32_t log_entry,
+			     AdspLogHandle const *log_handle, uint32_t param1,
+			     uint32_t param2, uint32_t param3, uint32_t param4);
+
+AdspErrorCode SystemServiceSafeMemcpy(void *RESTRICT dst, size_t maxlen,
```
lgirdwood, line 21: I suspect these are IADK APIs to be used in the IADK module layer ? if so, It's probably worth spelling this out in the header so that everyone knows the APIs have external users for compatibility.

src/audio/module_adapter/library/native_system_agent.c
```diff
@@ -0,0 +1,71 @@
...
+AdspErrorCode SystemServiceCreateNotification(NotificationParams *params,
+					      uint8_t *notification_buffer,
+					      uint32_t notification_buffer_size,
+					      AdspNotificationHandle *handle);
+
+AdspErrorCode SystemServiceSendNotificationMessage(NotificationTarget notification_target,
+						   AdspNotificationHandle message,
+						   uint32_t actual_payload_size);
+
+AdspErrorCode SystemServiceGetInterface(AdspIfaceId id, SystemServiceIface **iface);
```
lgirdwood, line 38: Shouldn't these declarations be in the header ?

lgirdwood: and have some doxygen descriptions ?

lgirdwood (changes requested):
Needs review comments to be resolved.

src/include/sof/audio/module_adapter/library/module_api_ver.h
```diff
@@ -0,0 +1,34 @@
...
+
+typedef union sof_module_api_version {
+	uint32_t full;
+	struct {
+		uint32_t minor    : 10;
+		uint32_t middle   : 10;
+		uint32_t major    : 10;
+		uint32_t reserved : 2;
+	} fields;
+} sof_module_api_version;
```
aborisovich, line 27: How are we going to use it?
What clients read/validate this?

kv2019i:
@pjdobrowolski Can you update on the status? Are all issues addressed? Please work with the revieiwers who have given feedback so far and ping to them to check again if the PR is ready. Let me know if you don't get traction.

pjdobrowolski:
I'm addressing issues. Quite big refactor is being made although module adapter issue with wrappers is not addressed yet and it is waiting for discussion.

src/include/sof/audio/module_adapter/library/module_api_ver.h
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+#ifndef __MODULE_API_VER_H__
+#define __MODULE_API_VER_H__
+
+/*
+ *  Api version 5.0.0 for sof loadable modules
+ */
+
+#define SOF_MAJOR_API_MODULE_VERSION 5
```
lgirdwood, line 15: Can we prefix the version like the structures i.e. ```SOF_MODULE_API_MAJOR_VERSION```

src/include/sof/audio/module_adapter/library/module_api_ver.h
```diff
@@ -0,0 +1,34 @@
...
+		uint32_t middle   : 10;
+		uint32_t major    : 10;
+		uint32_t reserved : 2;
+	} fields;
+} sof_module_api_version;
+
+typedef const struct {
+	uint32_t format;
+	sof_module_api_version api_version_number;
+} sof_module_build_info;
```
lgirdwood, line 32: ```sof_module_api_build_info``` ?

src/audio/module_adapter/iadk/system_agent.cpp
```diff
@@ -16,7 +16,7 @@
 #include <utilities/array.h>
 #include <adsp_error_code.h>
 #include <logger.h>
-#include <system_service.h>
+#include <native_system_service.h>
```
lgirdwood, line 19: Looks like the rimage submodule has been updated and should be another patch unless its needed here ?

lgirdwood (commented):
Minor naming.

src/audio/module_adapter/library/native_system_service.c
```diff
@@ -123,10 +121,19 @@ AdspErrorCode SystemServiceSendNotificationMessage(NotificationTarget notificati
...
 }
 
+struct native_system_service_api native_system_service = {
+	.log_message  = native_system_service_log_message,
+	.safe_memcpy = native_system_service_safe_memcpy,
+	.safe_memmove = native_system_service_safe_memmove,
+	.vec_memset = native_system_service_vec_memset,
+	.notification_create = native_system_service_create_notification,
+	.notification_send = native_system_service_send_notif_msg,
+	.get_interface = native_system_service_get_interface
```
lgirdwood, line 138: Long term we can use Zephyr runtime linking when it's ready so modules can be more compatible with base FW Kconfig changes.

src/include/sof/audio/module_adapter/library/native_system_service.h
```diff
@@ -0,0 +1,106 @@
...
+
+/*! \brief Defines notification targets supported by ADSP system */
+typedef enum _notification_target {
+	DSP_TO_HOST = 1,  /*!< Notification target is HOST */
+	DSP_TO_ISH = 2    /*!< Notification target is ISH */
+} adsp_notification_target;
+
+/*! \brief Defines notification types supported by ADSP system */
+typedef enum _notification_type {
+	VOICE_COMMAND_NOTIFICATION = 4,	/*!< intel_adsp define corresponding to PHRASE_DETECTED
```
lgirdwood, line 51: Can we have a comment in each notification type to preserve ID for compatability. i.e. new types should start at value X

src/include/sof/audio/module_adapter/library/native_system_service.h
```diff
@@ -0,0 +1,106 @@
...
+				      * in case there is no data.
+				      */
+	uint32_t event_data[];       /*!< Optional event data (size set to 0 as it is optional
+				      * data)
+				      */
+} module_event_notification;
+
+/*! \brief Defines notification targets supported by ADSP system */
+typedef enum _notification_target {
+	DSP_TO_HOST = 1,  /*!< Notification target is HOST */
```
lgirdwood, line 45: Can we prefix these as they are quite generic. e.g. ```NOTIFICATION_TARGET_```

src/include/sof/audio/module_adapter/library/native_system_service.h
lgirdwood: Ditto, need a prefix.

src/include/sof/audio/module_adapter/library/native_system_service.h
```diff
@@ -0,0 +1,106 @@
...
+	AUDIO_CLASSIFIER_RESULTS = 9,	/*!< intel_adsp define corresponding to FW_AUD_CLASS_RESULT
+					 * notification
+					 */
+	MODULE_EVENT_NOTIFICATION = 12,	/*!< intel_adsp define corresponding to MODULE_NOTIFICATION
+					 * notification
+					 */
+} notification_type;
+
+typedef enum _adsp_iface_id {
+	IfaceIdGNA = 0x1000,			/*!< Reserved for ADSP system */
```
lgirdwood, line 63: Comment to preserve IDs for compatability, prefix and coding style.

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -80,8 +83,33 @@ static int iadk_modules_init(struct processing_module *mod)
...
-					   instance_id, 0, log_handle, &mod_cfg);
+	/* Check if native Zephyr lib is loaded */
+	struct sof_man_fw_desc *desc;
+
+	desc = lib_manager_get_library_module_desc(module_id);
+	if (!desc) {
+		comp_err(dev, "iadk_modules_init(): Failed to load manifest");
+		return -ENOMEM;
+	}
+	struct sof_man_module *module_entry =
```
lgirdwood, line 95: Do we validate entry is in the module memery region ?

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -228,6 +284,13 @@ static int iadk_modules_set_configuration(struct processing_module *mod, uint32_
 					  size_t fragment_size, uint8_t *response,
 					  size_t response_size)
 {
+	if (mod->is_native_sof) {
+		struct module_interface *mod_in =
+					(struct module_interface *)mod->priv.module_adapter;
+
+		return mod_in->set_configuration(mod, config_id, pos, data_offset_size, fragment,
```
lgirdwood, line 290: why no else here for iadk call ?

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -249,6 +312,13 @@ static int iadk_modules_get_configuration(struct processing_module *mod, uint32_
 					  uint32_t *data_offset_size, uint8_t *fragment,
 					  size_t fragment_size)
 {
+	if (mod->is_native_sof) {
+		struct module_interface *mod_in =
+					(struct module_interface *)mod->priv.module_adapter;
+
+		return mod_in->get_configuration(mod, config_id, data_offset_size,
+						 fragment, fragment_size);
+	}
```
lgirdwood, line 320: ditto, why is this call different for other IADK calls ? in that we call both native and IADK APIs ? 
Best to explain in inline comments.

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -264,6 +334,12 @@ static int iadk_modules_get_configuration(struct processing_module *mod, uint32_
 static int iadk_modules_set_processing_mode(struct processing_module *mod,
 					    enum module_processing_mode mode)
 {
+	if (mod->is_native_sof) {
+		struct module_interface *mod_in =
+					(struct module_interface *)mod->priv.module_adapter;
+
+		return mod_in->set_processing_mode(mod, mode);
```
lgirdwood, line 340: ditto

src/audio/module_adapter/module/iadk_modules.c
```diff
@@ -287,6 +363,12 @@ static enum module_processing_mode iadk_modules_get_processing_mode(struct proce
  */
 static int iadk_modules_reset(struct processing_module *mod)
 {
+	if (mod->is_native_sof) {
+		struct module_interface *mod_in =
+					(struct module_interface *)mod->priv.module_adapter;
+
+		return mod_in->reset(mod);
```
lgirdwood, line 369: ditto

src/include/sof/audio/module_adapter/module/generic.h
```diff
@@ -212,6 +212,9 @@ struct processing_module {
 	 */
 	bool stream_copy_single_to_single;
 
+	/* flag to insure that module is loadable */
```
lgirdwood, line 215: this is really more, I would explain the high level differences here in an inline comment.

src/audio/module_adapter/module/iadk_modules.c
lyakh: @lgirdwood because in above cases there's a single `return` statement in the end and they only assign `ret` on `if` branches, whereas here there's a `return` directly under the `if` so you don't need an `else`. Of 3 variants:
1.
```
if (a)
	ret = f1();
else
	ret = f2();
return ret;
```
2.
```
if (a)
	return f1();
else
	return f2();
```
3.
```
if (a)
	return f1();
return f2();
```
we try to avoid (2) (and I think checkpatch complains about it), both (1) and (3) are acceptable by the style, personally I prefer (3).

lgirdwood: @lyakh I mean in reference to some ops calling both native and iadk apis and some ops calling either iadk or native api. i.e. we need som comments to explain.

src/include/sof/audio/module_adapter/library/native_system_service.h
lgirdwood: @pjdobrowolski we still need a comment that says something like "dont change enum values as they are needed for compatability, please add new items with next incremental ID"

src/include/sof/audio/module_adapter/library/native_system_service.h
```diff
@@ -0,0 +1,109 @@
...
+				      */
+} module_event_notification;
+
+/*! \brief Defines notification targets supported by ADSP system */
+typedef enum _notification_target {
+	NOTIFICATION_TARGET_DSP_TO_HOST = 1,  /*!< Notification target is HOST */
+	NOTIFICATION_TARGET_DSP_TO_ISH = 2    /*!< Notification target is ISH */
+} adsp_notification_target;
+
+/*! \brief Defines notification types supported by ADSP system */
```
lgirdwood, line 49: ditto here, we need to say not to change enum values for compatability.

src/audio/module_adapter/module/iadk_modules.c
lgirdwood: pls ignore, these are now visible in the new diff and I have coffee.

lgirdwood (commented):
For me we just need to comment the enum values cant be changed to preserve compatability.

lgirdwood (approved):

lgirdwood:
@pjdobrowolski looks like some CI issues need to be fixed.

marc-hb:
I had a quick look and I didn't see any CI issue, only "real" issues like this compilation issue: https://github.com/thesofproject/sof/actions/runs/5475101798/jobs/9970738368?pr=7018

lyakh (changes requested):
making sure this doesn't get merged by accident before #7867 is clarified, also so far the main effort is making this to work on the mtl-005 branch, only after that we can port it to "main"

pjdobrowolski:
@lyakh From my perspective clearifying #7867 will took too much time and has nothing to do with main branch. That feature should be merged piece by piece and only that way we can be sure that quality is best. Blocking that PR by that will make more harm and it is putting me in question if that feature is working or is bag of trash. 

From beginning we pick super simple module to port which is up down mixer to check if it all work and now you want me to wait another months for smart amp to be "clear"? Sorry but that way this feature will took years to merge.

lyakh:
> @lyakh From my perspective clearifying #7867 will took too much time and has nothing to do with main branch. That feature should be merged piece by piece and only that way we can be sure that quality is best. Blocking that PR by that will make more harm and it is putting me in question if that feature is working or is bag of trash.
> 
> From beginning we pick super simple module to port which is up down mixer to check if it all work and now you want me to wait another months for smart amp to be "clear"? Sorry but that way this feature will took years to merge.

@pjdobrowolski I think code should be merged when it's reviewed and tested. Originally I don't know why it was decided to add smart-amp-test as a second loadable module example in addition to the up-down mixer. Maybe it was decided, that the up-down mixer is too simple and doesn't provide enough test coverage, or maybe because we couldn't easily test it under Linux, not sure. In principle I would be fine with first just merging the up-down mixer example as long as we review and test it. But this PR doesn't even include the up-down mixer example, so it cannot be tested as such?

lgirdwood:
@pjdobrowolski there is an issue loading modules at that is different vs Windows. Need to align at sync tomorrrow before we merge.

src/include/sof/audio/module_adapter/library/module_api_ver.h
```diff
@@ -0,0 +1,34 @@
...
+	struct {
+		uint32_t minor    : 10;
+		uint32_t middle   : 10;
+		uint32_t major    : 10;
+		uint32_t reserved : 2;
+	} fields;
+} sof_module_api_version;
+
+typedef const struct {
+	uint32_t format;
```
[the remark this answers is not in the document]
pjdobrowolski: It was taken from legacy API format. I don't really have an answer for that question because all modules have that parameter set to '0'. For me it is fine to drop it.

src/include/sof/audio/module_adapter/library/module_api_ver.h
```diff
@@ -0,0 +1,34 @@
...
+	sof_module_api_version api_version_number;
+} sof_module_api_build_info;
```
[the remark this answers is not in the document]
pjdobrowolski: That struct is the same in IADK modules and it is not packed https://github.com/thesofproject/converged-sof-modules/blob/main/FW/src/intel_adsp/include/adsp_stddef.h#L61 does packing it to fixed size won't harm anything?

src/include/sof/audio/module_adapter/library/module_api_ver.h
```diff
@@ -0,0 +1,34 @@
...
+	uint32_t full;
+	struct {
+		uint32_t minor    : 10;
+		uint32_t middle   : 10;
+		uint32_t major    : 10;
+		uint32_t reserved : 2;
+	} fields;
+} sof_module_api_version;
+
+typedef const struct {
```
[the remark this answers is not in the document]
pjdobrowolski: https://github.com/thesofproject/converged-sof-modules/blob/main/FW/src/intel_adsp/include/adsp_stddef.h#L61 so what is yours suggestion?

src/include/sof/audio/module_adapter/library/native_system_service.h
pjdobrowolski: Done

src/audio/module_adapter/module/iadk_modules.c
pjdobrowolski: We can, but it is not part of that PR.

src/include/sof/audio/module_adapter/library/module_api_ver.h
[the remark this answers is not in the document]
lyakh: we just use `struct s` or `union u` in variable definitions directly

src/include/sof/audio/module_adapter/library/module_api_ver.h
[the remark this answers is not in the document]
lyakh: @pjdobrowolski adding a packed attribute might or might not break it. Say, if you had
```
struct x {
    uint16_t s;
    uint32_t i;
};
```
without packed, then "usually" compilers would align `i` on a 32-bit boundary and thus leave a 16-bit gap in the structure. If this is a part of an API and, say, is sent between the host and the DSP, then 2 different compilers are used. If they both happen to apply the same alignment in this case, it will work. But it's unreliable and it might change, at which point it will break. Now if you did have this case and it did work and now you add a packed attribute, that gap will disappear, and it will break the ABI because it will change the layout of the structure.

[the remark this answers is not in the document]
pjdobrowolski: For that point it is working and I wouldn't change anything about that.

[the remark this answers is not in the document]
lyakh: I would check how the compiler(s) pack it now, i.e. if there are any gaps. Probably there aren't, then I'd add the attribute to make sure it stays that way. If there are gaps (unlikely), I'd add "reserved" fields there and also a packed attribute to also make the structure immutable. But again - all that is needed only if it's indeed a part of the ABI

src/include/sof/audio/module_adapter/library/module_api_ver.h
```diff
@@ -0,0 +1,34 @@
...
+		uint32_t middle   : 10;
+		uint32_t major    : 10;
+		uint32_t reserved : 2;
+	} fields;
+} sof_module_api_version;
+
+struct {
+	uint32_t format;
+	sof_module_api_version api_version_number;
+} sof_module_api_build_info;
```
lyakh, line 32: also here looks like you forgot to remove the type name and removed the structure name instead, which made it an object definition instead of a structure declaration

src/include/sof/audio/module_adapter/library/module_api_ver.h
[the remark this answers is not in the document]
pjdobrowolski: @lyakh your are testing loadable modules now, can you check if it is necessary? It is just one macro and I would like to switch into dynamic memory allocation for loadable modules now.

src/include/sof/audio/module_adapter/library/native_system_service.h
```diff
@@ -0,0 +1,114 @@
...
+/*! \file native_system_service.h */
+#ifndef NATIVE_SYSTEM_SERVICE_H
+#define NATIVE_SYSTEM_SERVICE_H
+#include "logger.h"
+#include "adsp_stddef.h"
+#include "adsp_error_code.h"
+#include <stdint.h>
+
+/*! \brief This struct defines the obfuscating type for notifications. */
+typedef struct _adsp_notification_handle {} *adsp_notification_handle;
```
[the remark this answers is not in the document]
pjdobrowolski: it must be for now. After that PR I would like to clear it and CamelCases warnings. For github commits continuity that refactor must be done in another PR  for i.e. easier reverts

lyakh (approved):

lgirdwood (approved):

PR #7021, opened 2023-02-01
Audio: EQ FIR: Get in prepare() channels count from sink stream

singalsu:
In IPC4 prepare() the source stream channels count is not always set, it depends on topology. However the sink channels count has been just set by IPC4 params function that gets it from component's base configuration.


lgirdwood (commented):
@ranj063 wont channels come from an IPC4 module message ?

ranj063:
> @ranj063 wont channels come from an IPC4 module message ?
@lgirdwood yes they do and they are available as part of the base_cfg and the buffer is updated to use that channel count from the base cfg in eq_fir_params(). SO this PR does the right thing

ranj063 (approved):

andrula-song (approved):

lgirdwood (approved):

PR #7023, opened 2023-02-01
ipc4: add resource event notification structures

makarukp:
Provide unified structure for events that may be raised by identifiable
entity from inside the FW(like a module instance identifiable by module
instance ID or a pipeline identifiable by pipeline ID). All enums are
aligned with ipc4 protocol. Receiver of such notification is host.

Signed-off-by: Piotr Makaruk [piotr.makaruk@intel.com](mailto:piotr.makaruk@intel.com)

### DIFF
--- a/src/include/ipc4/notification.h
+++ b/src/include/ipc4/notification.h
@@ -42,6 +42,67 @@ enum sof_ipc4_notification_type {
 	SOF_IPC4_MANAGEMENT_SERVICE		= 16,
 };
 
+/**
+ * \brief Resource Event Notification provides unified structure for events that may be raised by
+ * identifiable entity from inside the FW.
+ */
+enum sof_ipc4_resource_event_type {
+	/* MCPS budget for the module exceeded */
+	SOF_IPC4_BUDGET_VIOLATION		= 0,
+	/* Underrun detected by the Mixer */
+	SOF_IPC4_MIXER_UNDERRUN_DETECTED	= 1,
+	/* Stream data segment completed by the gateway */
+	SOF_IPC4_STREAM_DATA_SEGMENT		= 2,
+	/* Error caught during data processing */
+	SOF_IPC4_PROCESS_DATA_ERROR		= 3,
+	/* Stack overflow in a module instance */
+	SOF_IPC4_STACK_OVERFLOW			= 4,
+	/* KPB changed buffering mode */
+	SOF_IPC4_BUFFERING_MODE_CHANGED		= 5,
+	/* Underrun detected by gateway. */
+	SOF_IPC4_GATEWAY_UNDERRUN_DETECTED	= 6,
+	/* Overrun detected by gateway */
+	SOF_IPC4_GATEWAY_OVERRUN_DETECTED	= 7,
+	/* DP task missing the deadline */
+	SOF_IPC4_EDF_DOMAIN_UNSTABLE		= 8,
+	/* Watchdog notification */
+	SOF_IPC4_WATCHDOG_EVENT			= 9,
+	/* IPC gateway reached high threshold */
+	SOF_IPC4_GATEWAY_HIGH_THRES		= 10,
+	/* IPC gateway reached low threshold */
+	SOF_IPC4_GATEWAY_LOW_THRES		= 11,
+	/* Bit Count Error detected on I2S port */
+	SOF_IPC4_I2S_BCE_DETECTED		= 12,
+	/* Clock detected/loss on I2S port */
+	SOF_IPC4_I2S_CLK_STATE_CHANGED		= 13,
+	/* I2S Sink started/stopped dropping data in non-blk mode */
+	SOF_IPC4_I2S_SINK_MODE_CHANGED		= 14,
+	/* I2S Source started/stopped generating 0's in non-blk mode */
+	SOF_IPC4_I2S_SOURCE_MODE_CHANGED	= 15,
+	/* Frequency drift exceeded limit in SRE */
+	SOF_IPC4_SRE_DRIFT_TOO_HIGH		= 16,
+	/* The notification should be sent only once after exceeding threshold or aging timer. */
+	SOF_IPC4_TELEMETRY_DATA_STATUS		= 17,
+	/* SNDW debug notification e.g. external VAD detected */
+	SOF_IPC4_SNDW_DEBUG_INFO		= 18,
+	/* Invalid type */
+	SOF_IPC4_INVALID_RESORUCE_EVENT_TYPE	= 19,
+};
+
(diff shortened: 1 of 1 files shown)


src/include/ipc4/notification.h
```diff
@@ -42,6 +42,68 @@ enum sof_ipc4_notification_type {
...
+	/* VAD event detected */
+	SOF_IPC4_SNDW_DEBUG_INFO		= 18,
+	/* Invalid type */
+	SOF_IPC4_INVALID_RESORUCE_EVENT_TYPE,
+};
+
+/* Resource Type - source of the event */
+enum sof_ipc4_resource_type {
+	/* Module instance */
+	MODULE_INSTANCE = 0,
```
lgirdwood, line 96: These will need a prefix

src/include/ipc4/notification.h
```diff
@@ -168,3 +230,22 @@ static inline void ipc4_notification_watchdog_init(struct ipc4_watchdog_timeout_
 	notif->primary.r.rsp = SOF_IPC4_MESSAGE_DIR_MSG_REQUEST;
 	notif->primary.r.msg_tgt = SOF_IPC4_MESSAGE_TARGET_FW_GEN_MSG;
 }
+
+/* Input Data Payload */
+union ipc4_resource_event_data {
+	/* Raw data */
+	uint32_t dws[6];
+} __packed __aligned(8);
```
lgirdwood, line 238: This should really be sized, i.e. we should use the reserved field in the parent to state the size here.

src/include/ipc4/notification.h
```diff
@@ -42,6 +42,68 @@ enum sof_ipc4_notification_type {
...
+	SOF_IPC4_I2S_CLK_STATE_CHANGED		= 13,
+	/* I2S Sink started/stopped dropping data in non-blk mode */
+	SOF_IPC4_I2S_SINK_MODE_CHANGED		= 14,
+	/* I2S Source started/stopped generating 0's in non-blk mode */
+	SOF_IPC4_I2S_SOURCE_MODE_CHANGED	= 15,
+	/* Frequency drift exceeded limit in SRE */
+	SOF_IPC4_SRE_DRIFT_TOO_HIGH		= 16,
+	/* The notification should be sent only once after exceeding threshold or aging timer */
+	SOF_IPC4_TELEMETRY_DATA_STATUS		= 17,
+	/* VAD event detected */
```
dnikodem, line 87: Could you change this comment to some more "generic"? Like: "SNDW debug notification e.g. external VAD detected"

src/include/ipc4/notification.h
dabekjakub: Agreed Union here looks superficial. Was there any specific intent for it here ? If not it should go as Liam sugested

dabekjakub (changes requested):

lyakh:
Personally I have a very poor imagination. I find reviewing "preparation" PRs with no immediate users difficult. This is one of them. In this and in similar cases I'd much rather see at least a single user of the proposed preparation at least as a draft PR. Could you maybe at least push a draft PR with this commit and further ones adding users? It would really help understand this. Also maybe we don't need all the defines at ones. This isn't an exported API, is it? It's completely internal within SOF, so we don't need all the notification values in advance. We can add them gradually as users emerge.

makarukp:
@lyakh 
I understand your point. I hope now it is more readable.

src/include/ipc4/notification.h
```diff
@@ -168,3 +183,22 @@ static inline void ipc4_notification_watchdog_init(struct ipc4_watchdog_timeout_
 	notif->primary.r.rsp = SOF_IPC4_MESSAGE_DIR_MSG_REQUEST;
 	notif->primary.r.msg_tgt = SOF_IPC4_MESSAGE_TARGET_FW_GEN_MSG;
 }
+
+/* Input Data Payload */
+union ipc4_resource_event_data {
+	/* Raw data */
+	uint32_t dws[6];
```
lgirdwood, line 196: @makarukp what the answer re having a size for this data i.e. this would allow us to easily extend it as needed in the future. We have the reserved filed in teh parent structure. For backwards compat we could say size 0 means this structure is size 6, i.e size = size_of_data + 6 words.

src/include/ipc4/notification.h
```diff
@@ -47,12 +47,18 @@ enum sof_ipc4_notification_type {
  * identifiable entity from inside the FW.
  */
 enum sof_ipc4_resource_event_type {
+	/* Underrun detected by gateway. */
+	SOF_IPC4_GATEWAY_UNDERRUN_DETECTED      = 6,
+	/* Overrun detected by gateway */
+	SOF_IPC4_GATEWAY_OVERRUN_DETECTED       = 7,
 	/* Invalid type */
 	SOF_IPC4_INVALID_RESORUCE_EVENT_TYPE,
```
lyakh, line 55: this mix of automatic and explicit values looks strange

src/include/ipc4/notification.h
```diff
@@ -168,3 +229,26 @@ static inline void ipc4_notification_watchdog_init(struct ipc4_watchdog_timeout_
...
+
+struct ipc4_resource_event_data_notification {
+	/* Type of originator (see sof_ipc4_resource_type) */
+	uint32_t resource_type;
+	/* ID of resource firing event */
+	uint32_t resource_id;
+	/* Type of fired event (see sof_ipc4_resource_event_type enum) */
+	uint32_t event_type;
+	/* Explicit alignment as ipc4_resource_event_data contains 64bit fields and needs alignment */
+	uint32_t reserved0;
```
lgirdwood, line 251: dont we need to use ```reserved0``` for size ? (or even part of ```reserved0``` like 16bits)

makarukp: I think we will be able to deduce all the sizes by the event_type and appropriate size of future structures for each one. I think reserved0 as a size is not necessary.

lgirdwood: using types implies that all ingredients in FW/SW and tooling need to know about the type and map it to size (even just to pass that data through to the next layer in the stack - e.g. FW passing event to userspace means kernel needs to map the type to size). 
Using all or some of reserved0 allows SW not to care about type (when there is no direct data usage) and just memcpy() the data to the next layer.

lgirdwood: @makarukp thanks for explaining this only used for resource info and not for general purpose event messaging to host. All good to go.

lgirdwood (approved):

lgirdwood:
@dabekjakub changes good for you ?

dnikodem (approved):

mwasko (approved):

PR #7027, opened 2023-02-01
board: intel_adsp_ace15_mtpm xtensa toolchain upgrade and enable xt-clang

aborisovich:
Release toolchain for Intel Meteorlake board is Xtensa RI-2022.10.
Upgraded toolchain to final version. New Xtensa toolchains does not support obsolete xt-xcc driver anymore,
so we need to switch to new xt-clang driver.
In this PR:

- modified `xtensa-build-system.py` to build mtl platform with new Xtensa toolchain version RI-2022.10 and core ace10_LX7HiFi4_2022_10
- modified processes and `xtensa-build-system.py` to use new compiler **xt-clang** as the xt-xcc is not available anymore
in new Xtensa toolchains.
- added requirement to SOF project to use C++17 standard. This was caused by changes in Zephyr https://github.com/zephyrproject-rtos/zephyr/pull/53405 that require C++14 standard.
MTL has cpp code that uses those headers but we can move straight away to C++17 as the newest standard supported by xt-clang in this toolchain.
- please note: commit 9ae133d1abedd0ab8ebc8999954cf249648e84e7 will be dropped from this PR before merging. It contains temporary code to test new toolchain in Intel Internal CI & SOF CI and also provide debug logging. 

Merging this PR will require synchronization with the Intel Internal CI team so they can change environmental variables on their side that would select new toolchain as default in production.

Dependencies:

- [x] https://github.com/zephyrproject-rtos/zephyr/pull/54333 NOT MANDATORY
- [x] https://github.com/zephyrproject-rtos/zephyr/pull/54836
- [x] https://github.com/thesofproject/sof/pull/7096
- [x] Changes in the Intel internal CI.
- [x] full scope tests for MTL to prove lack of regression

### DIFF
--- a/scripts/xtensa-build-zephyr.py
+++ b/scripts/xtensa-build-zephyr.py
@@ -76,25 +76,29 @@ platform_list = [
 		"name": "apl",
 		"PLAT_CONFIG": "intel_adsp_cavs15",
 		"XTENSA_CORE": "X4H3I16w2D48w3a_2017_8",
-		"XTENSA_TOOLS_VERSION": f"RG-2017.8{xtensa_tools_version_postfix}"
+		"XTENSA_TOOLS_VERSION": f"RG-2017.8{xtensa_tools_version_postfix}",
+		"DEFAULT_TOOLCHAIN_VARIANT": "xcc"
 	},
 	{
 		"name": "cnl",
 		"PLAT_CONFIG": "intel_adsp_cavs18",
 		"XTENSA_CORE": "X6H3CNL_2017_8",
-		"XTENSA_TOOLS_VERSION": f"RG-2017.8{xtensa_tools_version_postfix}"
+		"XTENSA_TOOLS_VERSION": f"RG-2017.8{xtensa_tools_version_postfix}",
+		"DEFAULT_TOOLCHAIN_VARIANT": "xcc"
 	},
 	{
 		"name": "icl",
 		"PLAT_CONFIG": "intel_adsp_cavs20",
 		"XTENSA_CORE": "X6H3CNL_2017_8",
-		"XTENSA_TOOLS_VERSION": f"RG-2017.8{xtensa_tools_version_postfix}"
+		"XTENSA_TOOLS_VERSION": f"RG-2017.8{xtensa_tools_version_postfix}",
+		"DEFAULT_TOOLCHAIN_VARIANT": "xcc"
 	},
 	{
 		"name": "jsl",
 		"PLAT_CONFIG": "intel_adsp_cavs20_jsl",
 		"XTENSA_CORE": "X6H3CNL_2017_8",
-		"XTENSA_TOOLS_VERSION": f"RG-2017.8{xtensa_tools_version_postfix}"
+		"XTENSA_TOOLS_VERSION": f"RG-2017.8{xtensa_tools_version_postfix}",
+		"DEFAULT_TOOLCHAIN_VARIANT": "xcc"
 	},
 	{
 		"name": "tgl",
@@ -103,6 +107,7 @@ platform_list = [
 		"IPC4_RIMAGE_DESC": "tgl-cavs.toml",
 		"XTENSA_CORE": "cavs2x_LX6HiFi3_2017_8",
 		"XTENSA_TOOLS_VERSION": f"RG-2017.8{xtensa_tools_version_postfix}",
+		"DEFAULT_TOOLCHAIN_VARIANT": "xcc",
 		"RIMAGE_KEY": pathlib.Path(SOF_TOP, "keys", "otc_private_key_3k.pem")
 	},
 	{
@@ -112,13 +117,15 @@ platform_list = [
 		"IPC4_RIMAGE_DESC": "tgl-h-cavs.toml",
 		"XTENSA_CORE": "cavs2x_LX6HiFi3_2017_8",
 		"XTENSA_TOOLS_VERSION": f"RG-2017.8{xtensa_tools_version_postfix}",
+		"DEFAULT_TOOLCHAIN_VARIANT": "xcc",
 		"RIMAGE_KEY": pathlib.Path(SOF_TOP, "keys", "otc_private_key_3k.pem")
 	},
 	{
 		"name": "mtl",
 		"PLAT_CONFIG": "intel_adsp_ace15_mtpm",
-		"XTENSA_CORE": "ace10_LX7HiFi4_RI_2020_5",
-		"XTENSA_TOOLS_VERSION": f"RI-2020.5{xtensa_tools_version_postfix}",
+		"XTENSA_CORE": "ace10_LX7HiFi4_2022_10",
+		"XTENSA_TOOLS_VERSION": f"RI-2022.10{xtensa_tools_version_postfix}",
+		"DEFAULT_TOOLCHAIN_VARIANT": "xt-clang",
 		"RIMAGE_KEY": pathlib.Path(SOF_TOP, "keys", "otc_private_key_3k.pem")
 	},
 	# NXP platforms
@@ -542,7 +549,8 @@ def build_platforms():
 					"or is not a directory")
 
 			# set variables expected by zephyr/cmake/toolchain/xcc/generic.cmake
-			os.environ["ZEPHYR_TOOLCHAIN_VARIANT"] = "xcc"
+			os.environ["ZEPHYR_TOOLCHAIN_VARIANT"] = os.environ.get("ZEPHYR_TOOLCHAIN_VARIANT",
+				platform_dict["DEFAULT_TOOLCHAIN_VARIANT"])
 			XTENSA_TOOLCHAIN_PATH = str(pathlib.Path(xtensa_tools_root_dir, "install",
 				"tools").absolute())
 			os.environ["XTENSA_TOOLCHAIN_PATH"] = XTENSA_TOOLCHAIN_PATH
--- a/west.yml
+++ b/west.yml
@@ -43,7 +43,7 @@ manifest:
 
     - name: zephyr
       repo-path: zephyr
-      revision: 9af2789cad17924298d705b99a1bca5f35ea88e2
+      revision: e3ae110a05d3427647a03fcbeff4478c195c5a48
       remote: zephyrproject
       # Import some projects listed in zephyr/west.yml@revision
       #
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -101,6 +101,10 @@ endif()
 
 # SOF module init
 zephyr_library_named(modules_sof)
+
+# Zephyr C++ code requires 14 or newer standard
+set_property(TARGET modules_sof PROPERTY CXX_STANDARD 17)
+
 zephyr_include_directories(
 	include
 )


aborisovich:
I do not understand why zephyr builds CI is failing. It uses Zephyr SDK so no change done. It stops with weird error during west update step....

marc-hb:
That's probably because we use `west update --narrow` to speed things up. Try either the full, 40 digit SHA1 or `pull/54333/head` or `pull/54333/head` (without changing the remote)

scripts/xtensa-build-zephyr.py
```diff
@@ -542,7 +542,10 @@ def build_platforms():
 					"or is not a directory")
 
 			# set variables expected by zephyr/cmake/toolchain/xcc/generic.cmake
-			os.environ["ZEPHYR_TOOLCHAIN_VARIANT"] = "xcc"
+			if not os.getenv("ZEPHYR_TOOLCHAIN_VARIANT"):
+				raise RuntimeError(f"Platform {platform} uses Xtensa toolchain."
+					"\nPlease set ZEPHYR_TOOLCHAIN_VARIANT environmental variable to"
+					"'xcc' or 'xcc-clang'")
```
marc-hb, line 548: Can you please provide a default value here?
```suggestion
if not os.getenv("ZEPHYR_TOOLCHAIN_VARIANT"):
    os.environ["ZEPHYR_TOOLCHAIN_VARIANT"] = "xcc-clang"
```

aborisovich: I don't think default value is a good idea here especially one set to xt-clang. Script end-user if not set this variable could be trying to compile TGL with xt-clang what is possible but what we do not support (well at least I had not checked it for compiler errors/warnings). Also TGL does not have xcc-clang toolchain enabled in Zephyr board yaml file. I'm not sure what would happen.

marc-hb: > Is this value needed by some CI systems?

If it's optional then why fail when it's missing?

> if not set this variable could be trying to compile TGL with xt-clang what is possible but what we do not support [...]

It's OK to fail in unsupported cases.  It's not OK to require a new, "evil" environment variable for the default case.

>  I'm not sure what would happen.

If the error message for some unsupported case is too cryptic, the few people who want/need that case can always submit a better error message.

> Also TGL does not have xcc-clang toolchain enabled in Zephyr board yaml file.

BTW the `platform_list` has always provided platform-specific default values. In fact `platform_list` is probably the main value of this script: platform-specific default values that allow building multiple platforms at once without any painful configuration tricks. It will be a real regression to lose that.

aborisovich: Hmm, yeah you are right.
Maybe approach with expanding platform_list values would be much better solution (we just did not have to distinguish between Xtensa compilers before). I'll just add new field to platform_list that would provide info that for TGL default toolchain is xt-xcc and for MTL is xt-clang. That would make much more sense.

aborisovich: @marc-hb does this looks better now?

scripts/xtensa-build-zephyr.py
```diff
@@ -542,7 +549,7 @@ def build_platforms():
 					"or is not a directory")
 
 			# set variables expected by zephyr/cmake/toolchain/xcc/generic.cmake
-			os.environ["ZEPHYR_TOOLCHAIN_VARIANT"] = "xcc"
+			os.environ["ZEPHYR_TOOLCHAIN_VARIANT"] = platform_dict["ZEPHYR_TOOLCHAIN_VARIANT"]
```
marc-hb, line 552:
```suggestion
            if not os.environ["ZEPHYR_TOOLCHAIN_VARIANT"]:
			       os.environ["ZEPHYR_TOOLCHAIN_VARIANT"] = platform_dict["ZEPHYR_TOOLCHAIN_VARIANT"]
```

scripts/xtensa-build-zephyr.py
```diff
@@ -76,25 +76,29 @@
 		"name": "apl",
 		"PLAT_CONFIG": "intel_adsp_cavs15",
 		"XTENSA_CORE": "X4H3I16w2D48w3a_2017_8",
-		"XTENSA_TOOLS_VERSION": f"RG-2017.8{xtensa_tools_version_postfix}"
+		"XTENSA_TOOLS_VERSION": f"RG-2017.8{xtensa_tools_version_postfix}",
+		"ZEPHYR_TOOLCHAIN_VARIANT": "xcc"
```
marc-hb, line 80:
```suggestion
		"DEFAULT_XCC_VARIANT": "xcc"
```

marc-hb (commented):
> Intel Meteorlake release toolchain is Xtensa RI-2021.7 version. New toolchain does not support xt-xcc driver anymore so we are forced to switch to new xt-clang driver.

Wait, this sounds like a "Flag Day" change? I mean: there is no "transition" toolchain that supports both `xt-xcc` and `xt-clang`, it's either one or the other? That sounds bad, it sounds like only the "before" version of this script will be compatible with older toolchains and only the "after" version of this script will be compatible with newer toolchains? OK now maybe I understand why you were asking for a new input...

Assuming I'm understanding all this correctly, I think we need both a default value for convenience _and_ a new input parameter so users can use git across the "flag day" without having to switch toolchain at the same time. See inline suggestions.

cc: @dcpleung, @aasthagr

scripts/xtensa-build-zephyr.py
aborisovich: Yeah, this is an important one thanks

scripts/xtensa-build-zephyr.py
aborisovich: Here I would remain with those names simply because they are "all wrong".
So basically there is no such thing as "xcc variant".
We have two drivers in Xtensa toolchains (well until RI-2021.7):
"xt-xcc" and "xt-clang".
Somebody in Zephyr project, called the "xt-xcc" driver as "xcc".
Then the Clang came out to the light and they had to distinguish between native POSIX LLVM Clang and Xtensa Clang.
So they named it "xcc-clang". Bad naming indeed but xt-clang is certainly not a variant of "xcc".
Closest to the true would be `DEFAULT_XTENSA_TOOLCHAIN_VARIANT` naming.
But again it would have values, which are not matching the actual driver names.
So I would just stick with ZEPHYR_TOOLCHAIN_VARIANT to avoid any further confusion.

marc-hb: Very interesting, thanks for sharing! The main thing I was requesting is `DEFAULT` to show that it can be temporarily overriden with the environment, so people don't have uninstall/re-install a toolchain when using git checkout.

So how about `DEFAULT_TOOLCHAIN_VARIANT`? This script does not know anything besides `ZEPHYR`.

fredoh9:
I installed new toolchains, RI-2021.7 and RI-2022.10 (both builds and tools). I was able to build MTL fw. Then pull this PR  and try again. But build failed.

```
:~/work/zwork/sof$ ./scripts/xtensa-build-zephyr.py mtl
Found west: /home/fredoh/.local/bin/west
Building platforms: mtl
In dir: /home/fredoh/work/zwork; running command: west topdir
In dir: /home/fredoh/work/zwork; running command: west config manifest.path
In dir: /home/fredoh/work/zwork; running command: west config manifest.file
West workspace: /home/fredoh/work/zwork
West manifest path: /home/fredoh/work/zwork/sof/west.yml
SOF_TOP=/home/fredoh/work/zwork/sof
west_top=/home/fredoh/work/zwork
Cleaning mtl from /home/fredoh/work/zwork/build-sof-staging
XTENSA_TOOLCHAIN_PATH=/home/fredoh/xcc/install/tools
TOOLCHAIN_VER=RI-2021.7-linux
XTENSA_SYSTEM=/home/fredoh/xcc/install/builds/RI-2021.7-linux/ace10_LX7HiFi4/config
In dir: /home/fredoh/work/zwork; running command: west build --build-dir build-mtl --board intel_adsp_ace15_mtpm /home/fredoh/work/zwork/sof/app --
-- west build: generating a build system
Loading Zephyr default modules (Zephyr base (cached)).
-- Application: /home/fredoh/work/zwork/sof/app
-- Cache files will be written to: /home/fredoh/.cache/zephyr
-- Zephyr version: 3.2.99 (/home/fredoh/work/zwork/zephyr)
-- Found west (found suitable version "0.14.0", minimum required is "0.7.1")
-- Board: intel_adsp_ace15_mtpm
-- Found toolchain: xcc-clang (/home/fredoh/xcc/install/tools)
CMake Error at /home/fredoh/work/zwork/zephyr/cmake/compiler/xcc/generic.cmake:25 (message):
  Executing the below command failed.  Are permissions set correctly?

  '/home/fredoh/xcc/install/tools/RI-2021.7-linux/XtensaTools/bin/xt-clang
  --version'

Call Stack (most recent call first):
  /home/fredoh/work/zwork/zephyr/cmake/compiler/xcc-clang/generic.cmake:3 (include)
  /home/fredoh/work/zwork/zephyr/cmake/modules/FindHostTools.cmake:107 (include)
  /home/fredoh/work/zwork/zephyr/cmake/modules/dts.cmake:8 (find_package)
  /home/fredoh/work/zwork/zephyr/cmake/modules/zephyr_default.cmake:108 (include)
  /home/fredoh/work/zwork/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:66 (include)
  /home/fredoh/work/zwork/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:97 (include_boilerplate)
  CMakeLists.txt:5 (find_package)
...
```

fredoh9:
Looks like the tool install didn't finish properly. Now I'm able to build MTL fw.

```
:~/work/zwork/sof$ /home/fredoh/xcc/install/tools/RI-2021.7-linux/XtensaTools/bin/xt-clang --version
XtensaTools-14.07.20210809689958 clang version 10.0.1
Target: xtensa-unknown-unknown
Thread model: single
InstalledDir: /home/fredoh/xcc/install/tools/RI-2021.7-linux/XtensaTools/llvm/bin
```

marc-hb:
> or specify a different registry of Xtensa cores (with the --xtensa-system
option or the XTENSA_SYSTEM environment variable).

That's why all our build scripts define XTENSA_SYSTEM, so you don't have to and you don't have to "fix the registry".

marc-hb:
SOFCI TEST

fredoh9:
SOFCI TEST

fredoh9:
SOFCI TEST

fredoh9:
All build machines have new toolchain, RI-2021.7 and ace10_LX7HiFi4

fredoh9:
By the way, are you planning to update with RI-2022.10 also? If we have a plan to upgrade to RI-2022.10 sooner or later let me know also. I can prepare for that also.

lgirdwood (commented):
@aborisovich any ETA for non draft ?

aborisovich:
> By the way, are you planning to update with RI-2022.10 also? If we have a plan to upgrade to RI-2022.10 sooner or later let me know also. So that we can prepare for that also.

Hi, yes I've asked our process team to install RI-2022.10 straight away so I guess we will just skip 2021 toolchain. So please prepare for 2022.10 on sof CI side.

aborisovich:
> @aborisovich any ETA for non draft ?

I can make it normal Open PR but it has dependency to Zephyr PR with xt-clang introduction and "-no-pie" unsupported flag enforced by Zephyr toolchain. We will discuss those hanging PRs tomorrow with Zephyr team. I think code freeze is ended and those should be merged soon.

dcpleung:
https://github.com/zephyrproject-rtos/zephyr/pull/54836 would probably have a higher chance of getting merged sooner.

src/audio/selector/selector.c
```diff
@@ -579,7 +579,7 @@ static void build_config(struct comp_data *cd, struct module_config *cfg)
 	enum sof_ipc_frame __sparse_cache frame_fmt, valid_fmt;
 	int i;
 
-	frame_fmt = cfg->base_cfg.audio_fmt.depth;
+	frame_fmt = (int)cfg->base_cfg.audio_fmt.depth;
```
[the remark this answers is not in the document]
aborisovich: Well, when `sof/lib/dai.h` is included, `depth` is a `uint32_t` type member of `dai_plat_fifo_data` struct.
However when we compile with IPC4 config option,  header `src\include\ipc4\base-config.h` is uder and `depth` is `enum ipc4_bit_depth` type what causes the warning:

> implicit conversion from enumeration type 'enum ipc4_bit_depth' to different enumeration type 'enum sof_ipc_frame' [-Wenum-conversion]

Enums are implicitly cast to integral types but not to other enums. But thanks for the catch since I should cast to `(uint32_t)` here to avoid value overflow.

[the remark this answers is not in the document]
marc-hb: > However when we compile with IPC4 config option depth is enum ipc4_bit_depth instead of uint32_t

That definitely deserves a (short) comment IMHO.

marc-hb:
> I can make it normal Open PR but it has dependency to Zephyr PR with xt-clang introduction and "-no-pie" unsupported flag enforced by Zephyr toolchain.

@aborisovich could you fast-track the C code changes in a new, separate PR? It's easy enough for users to locally hack one-line in the python script + ignore the `-no-pie` warnings. Fixing C code is probably more time-consuming.

Once the C code is fixed it's also easier for Zephyr people like @dcpleung to give SOF a quick try.

aborisovich:
> [zephyrproject-rtos/zephyr#54836](https://github.com/zephyrproject-rtos/zephyr/pull/54836) would probably have a higher chance of getting merged sooner.

PRs like those give me joy. Finally those names had been fixed, that's what I've been talking about @marc-hb .

aborisovich:
> > I can make it normal Open PR but it has dependency to Zephyr PR with xt-clang introduction and "-no-pie" unsupported flag enforced by Zephyr toolchain.
> 
> @aborisovich could you fast-track the C code changes in a new, separate PR? It's easy enough for users to locally hack one-line in the python script and ignore the `-no-pie` warnings. Fixing C code is probably more time-consuming.
> 
> Once the C code is fixed it's also easier for Zephyr people like @dcpleung to give SOF a quick try.

Sure, I'll create new PR which we can instantly merge without any dependencies.

src/audio/selector/selector.c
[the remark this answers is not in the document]
aborisovich: Done in https://github.com/thesofproject/sof/pull/7096 .

aborisovich:
Created new PR https://github.com/thesofproject/sof/pull/7096 just for fixing C warnings.

aborisovich:
Ok, got info that our Intel Internal CI has new toolchains set up.
@fredoh9 what is the status of SOF CI? Are we ready there?

aborisovich:
Unfortunately there are some issues in Zephyr PR to which we have a dependency...
Looks like https://github.com/zephyrproject-rtos/zephyr/pull/54445 does not build with xt-clang on Windows (as it tries to select non Xtensa linker). Setting TOOLCHAIN_HOME variable does not help at all.
https://github.com/zephyrproject-rtos/zephyr/pull/54445#issuecomment-1434714264

lgirdwood:
@aborisovich I think we have some Zephyr merge breakages in general today, @kv2019i and @abonislawski are fixing. I would wait until Zephyr settles by mid week.

aborisovich:
### Hot news
There are two messages -  a good one and a bad one.

**A good one**
We can carry on with this PR since "-no-pie" flag had been removed in https://github.com/zephyrproject-rtos/zephyr/pull/54836 . We have toolchains installed on the CI machines and I will start testing those in a minute.

**A bad one**
### We should really merge this PR fast since 8 hours ago Zaphyr had broken completely compatibility with our xt-xcc toolchain for MTL code!
(emphasis mine)
This happened in the https://github.com/zephyrproject-rtos/zephyr/pull/53405 .
MTL has some code written in C++ that includes `zephyr/sys/util.h` header.
For xt-xcc error looks like this:
[xt-xcc.log](https://github.com/thesofproject/sof/files/10809855/xt-xcc.log)


New change requires at least C++14 standard.
From https://en.cppreference.com/w/cpp/language/constexpr partial quotes:

> constexpr function
> A constexpr function must satisfy the following requirements:
> (...)
> * the function body must be either deleted or defaulted or contain only the following: (until C++14)
>   - null statements
>   - plain semicolons
>   - static_assert](https://en.cppreference.com/w/cpp/language/static_assert) declarations
>   - typedef declarations and alias declarations that do not define classes or enumerations
>   - using declarations
>   - using directives
>   - if the function is not a constructor, exactly one return statement

For xt-clang with -std=c++11 we got:
[xt-clang C++11.log](https://github.com/thesofproject/sof/files/10809859/xt-clang.C%2B%2B11.log)
(warnings that we had violated rules above and functionality is provided using compiler extension).

aborisovich:
Since we are forced for C++ version upgrade, can we set C++17 straight away?
Amount of C++ code is really small right now and it does not conflict with Zephyr anyhow.
This is latest supported standard for xt-clang as of today.

scripts/xtensa-build-zephyr.py
```diff
@@ -535,19 +543,27 @@ def build_platforms():
...
+
 			xtensa_tools_root_dir = pathlib.Path(xtensa_tools_root_dir)
 			if not xtensa_tools_root_dir.is_dir():
 				raise RuntimeError(f"Platform {platform} uses Xtensa toolchain."
 					"\nVariable XTENSA_TOOLS_VERSION points path that does not exist\n"
 					"or is not a directory")
 
 			# set variables expected by zephyr/cmake/toolchain/xcc/generic.cmake
-			os.environ["ZEPHYR_TOOLCHAIN_VARIANT"] = "xcc"
+			if not os.environ["ZEPHYR_TOOLCHAIN_VARIANT"]:
```
marc-hb, line 556: This should fix the build failures in 
```suggestion
			if not os.environ.get("ZEPHYR_TOOLCHAIN_VARIANT"):
```
```
    if not os.environ["ZEPHYR_TOOLCHAIN_VARIANT"]:
  File "/usr/lib/python3.8/os.py", line 675, in __getitem__
    raise KeyError(key) from None
```

marc-hb: Or, even better
```suggestion
os.environ["ZEPHYR_TOOLCHAIN_VARIANT"] =
    os.environ.get("ZEPHYR_TOOLCHAIN_VARIANT", platf_stuff["DEFAULT_TOOLCHAIN_VARIANT"])
```

aborisovich: Thanks marc, I'll fix that but I'm currently struggling to replace environmental variables for execute_command(build_cmd, cwd=west_top)... I guess i need to pass there my custom env

marc-hb: This is currently failing on Linux SOFCI so better fix it sooner than later if you want some results?

I don't understand, we've been setting `os.environ["ZEPHYR_TOOLCHAIN_VARIANT"] = "xcc"` in this script since forever and it has been working since forever, at least on Linux...

Do you mean on Windows? I don't know about the environment on Windows (except it's very different from Linux) but on Linux all forked child processes automatically inherit the entire environment. That's why environment variable "hacks" are constantly abused on Linux: because it's so convenient.

aborisovich: Marc I'm trying to hook up to our Internal CI.
The fix you send me is great and I already applied it but it is not enough for me to push, since it would trigger a working build... using xt-xcc (what is not my goal). Take a look at what our QB is doing with environmental variables before xtensa-build-zephyr.py is called. I need to override those to trigger xt-clang toolchain selection. I'll do this a bit later today since I confused timezones and I'm unable to operate any longer ;-)

aborisovich:
Ok, it looks like everything works correctly.
Internal Intel CI System used xt-clang and completed the build ✔️ 
sof-ci/jenkins/pr-fw-build also used xt-clang and completed the build ✔️ 
I just need to drop now the last commit 9ae133d1abedd0ab8ebc8999954cf249648e84e7 
SOF CI device tests failed check-suspend-resume-with-capture.sh on ADLP_RVP_SDW_IPC4ZPH and TGLU_RVP_NOCODEC_IPC4ZPH but those seems unrelated as TGL had been build using existing xt-xcc toolchain.
@marc-hb @kv2019i, @lgirdwood and others please carry on with code review so we can make last polishes and merge.
Merging will require some synchronization with Internal Intel CI team as they need to change some environmental variables on their side.

pjdobrowolski (approved):

jxstelter (approved):

scripts/xtensa-build-zephyr.py
```diff
@@ -287,6 +294,7 @@ def execute_command(*run_args, **run_kwargs):
 	if run_kwargs.get('shell') or not isinstance(command_args, list):
 		raise RuntimeError("Do not rely on non-portable shell parsing")
 
+	args.verbose = 2
```
softwarecki, line 297: verbose as default?

softwarecki (approved):

scripts/xtensa-build-zephyr.py
```diff
@@ -532,8 +540,11 @@ def build_platforms():
 
 		platform_dict = [x for x in platform_list if x["name"] == platform][0]
 		xtensa_tools_root_dir = os.getenv("XTENSA_TOOLS_ROOT")
-		# when XTENSA_TOOLS_ROOT environmental variable is set,
-		# use user installed Xtensa tools not Zephyr SDK
+
+		# for some reason CI does not set XTENSA_TOOLS_ROOT while using xtensa toolchain...
+		if not xtensa_tools_root_dir and os.environ.get("XTENSA_INSTALL_PATH"):
+			# and has specific XTENSA_INSTALL_PATH set instead...
+			xtensa_tools_root_dir = pathlib.Path(os.environ.get("XTENSA_INSTALL_PATH"), "..", "..")
```
marc-hb, line 547: Note for other reviewers: this is part of the temporary, DO NOT MERGE commit. Do not review. It looked "too good" so I wasted a bit of time reviewing it, next time please add some uglier comments :-)

marc-hb (approved):
Approved except of course the DO NOT MERGE commit. Demoting to draft to avoid accidental merges, it happened :-) It's just one click both ways.

I'm not happy with the new C++17 requirement because it's not clear why it's needed: the author claims the requirement is still C++11 in 
https://github.com/zephyrproject-rtos/zephyr/pull/53405#issuecomment-1441187586


> added requirement to SOF project to use C++17 standard. This was caused by changes in Zephyr lib: os: PoT utilities, hash function, hash map (hash table) zephyrproject-rtos/zephyr#53405 that require C++14 standard.

@aborisovich can you please file a bug for this with (hopefully easy and open-source) reproduction steps? 

> SOF CI device tests failed check-suspend-resume-with-capture.sh on ADLP_RVP_SDW_IPC4ZPH and TGLU_RVP_NOCODEC_IPC4ZPH but those seems unrelated as TGL had been build using existing xt-xcc toolchain.

We have had suspend/resume failures since forever.

Always copy the URL because it gets lost after force push:
https://sof-ci.01.org/sofpr/PR7027/build4128/devicetest/index.html

aborisovich:
> I'm not happy with the new C++17 requirement because it's not clear why it's needed: the author claims the requirement is still C++11 in
> https://github.com/zephyrproject-rtos/zephyr/pull/53405#issuecomment-1441187586
> 
> But it's just a one-line, super easy to change later.

I've submitted the bug https://github.com/zephyrproject-rtos/zephyr/issues/55204 explaining the situation and containing the proof that Zephyr code requires C++14 after the PR we mentioned earlier.
I can set a language standard in this PR to C++14, but since we need to upgrade it anyway, and the existing C++ code works with the C++17 standard, I would like to upgrade to it straight away.
When this PR is merged we can also work on the upgrading C to C17 standard 😃 (but this will require much more work obviously).

aborisovich:
Results from validation team proved no regression on MTL full scope. Removed [DO NOT MERGE] commit, Intel Internal CI had switched to use xt-clang. Current build run will use new toolchain. Waiting for the results and if everything good we can merge @lgirdwood .

aborisovich:
Ok all green. `sof-ci/jenkins/pr-fw-build` & `Internal Intel CI System/merge/build` used new xt-clang toolchain.

PR #7028, opened 2023-02-01
Update fuzzing rig, make component creation work

andyross:
Come back to the fuzzing rig, dust off my brain (been a rough January here for me, sorry), and update to hopefully finish off IPC coverage.

As mentioned earlier, TPLG_COMP_NEW commands weren't being covered (and if we can't create components, we aren't able to get coverage on anything else!) because of the way they work.  The input wants to match a 128 bit UUID value in the command vs. a table of drivers, and that haystack is too deep for even fuzzing to find the needle.  This plays a trick by using an otherwise-ignored bit of fuzz input to encode a "create a whiteboxed command" metacommand.  When triggered, this adds Just Enough Correct Input to get the code through the driver selection stage of the IPC parser, but leaves the other random bits random.   There were a bunch of ancestor variants here which were terrible hacks, but I'm actually kinda proud of how this finally turned out.

Anyway, now we get all the way to the individual component creation paths, essentially completing coverage of all the IPC3 input paths (IPC4 isn't default in the tree yet, but AFAICT the protocol bits the fuzzing rig touches are all identical and it "should" work fine, modulo fuzz-detected-bugs, when we cut over; it's on my list to figure out ahead of time).

Interestingly I didn't stumble on any new bugs doing this, which makes me a little worried.  But I've manually traced out every creation path I can find and know they're all being hit.

We're also not hitting as much of the runtime behavior as I would have liked.  The fuzzing rig is successfully creating components, but after just a handful it's out of memory and the resulting random collection of junk doesn't actually try to do anything.  Maybe that's just asking too much of fuzzing as a technique; it's really only intended for input validation code.

### DIFF
--- a/scripts/fuzz.sh
+++ b/scripts/fuzz.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+set -e
+
+# Simple wrapper around a libfuzzer test run, as much for
+# documentation as direct use.  The idea here is really simple: build
+# for the Zephyr "native_posix" board (which is a just a x86
+# executable for the build host, not an emulated device) and run the
+# resulting zephyr.exe file.  This specifies a "fuzz_corpus" directory
+# to save the seeds that produce useful coverage output for use in
+# repeated runs (these are not particularly large, we might consider
+# curating and commiting such a seed directory to the tree).
+#
+# The tool will run until it finds a failure condition.  You will see
+# MANY errors on stdout from all the randomized input.  Don't try to
+# capture this, either let it output to a terminal or arrange to keep
+# only the last XXX lines after the tool exits.
+#
+# The only prerequisite to install is a clang compiler on the host.
+# Versions 12+ have all been observed to work.
+#
+# You will need the kconfigs specified below for correct operation,
+# but can add more at the end of this script's command line to
+# duplicate configurations as needed.  Alternatively you can pass
+# overlay files in kconfig syntax via -DOVERLAY_CONFIG=..., etc...
+
+export SOF_TOP=$(cd "$(dirname "$0")/.." && pwd)
+
+export ZEPHYR_BASE=$SOF_TOP/../zephyr
+export ZEPHYR_TOOLCHAIN_VARIANT=llvm
+
+main()
+{
+  west build -p -b native_posix $SOF_TOP/app/ -- \
+    -DCONFIG_ASSERT=y \
+    -DCONFIG_SYS_HEAP_BIG_ONLY=y \
+    -DCONFIG_ZEPHYR_NATIVE_DRIVERS=y \
+    -DCONFIG_ARCH_POSIX_LIBFUZZER=y \
+    -DCONFIG_ARCH_POSIX_FUZZ_TICKS=100 \
+    -DCONFIG_ASAN=y "$@"
+
+  mkdir -p ./fuzz_corpus
+  build/zephyr/zephyr.exe ./fuzz_corpus
+}
+
+main "$@"
--- a/src/platform/posix/include/platform/platform.h
+++ b/src/platform/posix/include/platform/platform.h
@@ -24,6 +24,8 @@
 
 #define PLATFORM_DEFAULT_DELAY 12
 
+struct sof;
+
 void posix_dma_init(struct sof *sof);
 void posix_dai_init(struct sof *sof);
 
--- a/src/platform/posix/ipc.c
+++ b/src/platform/posix/ipc.c
@@ -7,6 +7,7 @@
 #include <sof/ipc/common.h>
 #include <sof/ipc/schedule.h>
 #include <sof/schedule/edf_schedule.h>
+#include <sof/audio/component_ext.h>
 
 // 6c8f0d53-ff77-4ca1-b825-c0c4e1b0d322
 DECLARE_SOF_UUID("posix-ipc-task", ipc_task_uuid,
@@ -46,6 +47,8 @@ static uint8_t fuzz_in_sz;
 static void fuzz_isr(const void *arg)
 {
 	size_t rem, i, n = MIN(posix_fuzz_sz, sizeof(fuzz_in) - fuzz_in_sz);
+	bool comp_new = false;
+	int comp_idx = 0;
 
 	for (i = 0; i < n; i++)
 		fuzz_in[fuzz_in_sz++] = posix_fuzz_buf[i];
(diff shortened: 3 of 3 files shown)


scripts/fuzz.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+set -e
```
marc-hb, line 2: Like in most other scripts:
```shell
SOF_TOP=$(cd "$(dirname "$0")/.." && pwd)
```

scripts/fuzz.sh
```diff
@@ -0,0 +1,37 @@
...
+# only the last XXX lines after the tool exits.
+#
+# The only prerequisite to install is a clang compiler on the host.
+# Versions 12+ have all been observed to work.
+#
+# You will need the kconfigs specified below for correct operation,
+# but can add more at the end of this script's command line to
+# duplicate configurations as needed.
+
+export ZEPHYR_BASE=$SOF_TOP/../zephyr
```
marc-hb, line 25: I haven't used ZEPHYR_BASE for a long time, are you sure it's required in this case? Can this script be run outside a west workspace?

scripts/fuzz.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
marc-hb, line 1: I think you generally disagree but asking anyway: can you please move this to a `main "$@"` function? Rationale in https://github.com/thesofproject/sof-test/issues/740

scripts/fuzz.sh
```diff
@@ -0,0 +1,37 @@
...
+export ZEPHYR_BASE=$SOF_TOP/../zephyr
+export ZEPHYR_TOOLCHAIN_VARIANT=llvm
+
+west build -p -b native_posix $SOF_TOP/app/ -- \
+  -DCONFIG_ASSERT=y \
+  -DCONFIG_SYS_HEAP_BIG_ONLY=y \
+  -DCONFIG_ZEPHYR_NATIVE_DRIVERS=y \
+  -DCONFIG_ARCH_POSIX_LIBFUZZER=y \
+  -DCONFIG_ARCH_POSIX_FUZZ_TICKS=100 \
+  -DCONFIG_ASAN=y "$@"
```
marc-hb, line 34: Would it make sense to have all these in some kind of `defconfig` file? Overkill / too cumbersome?

scripts/fuzz.sh
andyross: I do, but this is your playground so no arguments here.

scripts/fuzz.sh
andyross: It's not needed in the default setup.  But I have like a thousand zephyr trees floating around and do dumb stuff like symlinking module directories across them.  If you have ZEPHYR_BASE set to the **wrong** tree, you get very funny results (like, the build works, but uses the wrong SOF tree!).  So I guess I'd see this as a best practices kind of thing to ensure that the tree this script is running from is the one being built.

scripts/fuzz.sh
andyross: I defer to you guys on that?  There's a thousand ways to do kconfig and this is the one that steps on the fewest toes.  Another option would be to create a new kconfig like CONFIG_SOF_FUZZER that then selects this stuff.

(Also I'm noticing that heap one, which doesn't belong.  I forget the details now, but the SOF "posix" platform creates a heap of the same size as the hardware platforms, but for some reason ends up building a zephyr without support for the "big" heap code and so gets runtime errors.  It's just a kconfig mismatch somewhere I worked around and then forgot about.)

lgirdwood: @kv2019i any comment on this re our Zephyr configs ?

kv2019i: I think it's fine here. An overlay file like sof/app/perf_overlay.conf (passed via -DOVERLAY_CONFIG= to west) would be perhaps be a good match, but as long as the assumption is the users all use this helper script anyways, I think the definitions are fine here.

kv2019i (commented):
No real objections, seems good to me.

scripts/fuzz.sh
marc-hb: > An overlay file like sof/app/perf_overlay.conf (passed via -DOVERLAY_CONFIG= to west) would be perhaps be a good match

@andyross can you add this as a one-line comment? Thx!

marc-hb:
Some of the checkpatch warnings seem easy to fix: https://github.com/thesofproject/sof/actions/runs/4069355263/jobs/7008988232

src/platform/posix/include/platform/platform.h
```diff
@@ -24,6 +24,8 @@
 
 #define PLATFORM_DEFAULT_DELAY 12
 
+struct sof;
```
lyakh, line 27: `extern`?

kv2019i: I think this is ok as a forward declaration.

kv2019i (approved):
Looks good, thanks @andyross !

lgirdwood (approved):

scripts/fuzz.sh
lgirdwood: @andyross your playground too !

andyross:
Cosmetic fixups as requested.  I have another series coming up pretty soon I think, but may ping you guys offline depending on results.

marc-hb (approved):

src/platform/posix/ipc.c
```diff
@@ -58,20 +61,74 @@ static void fuzz_isr(const void *arg)
...
+
 	// The first dword is a size value which fuzzing will stumble
-	// on only one time in 24M, fill it in manually.
+	// on only rarely, fill it in manually.
 	*(uint32_t *)global_ipc->comp_data = msgsz;
-	for (i = 0; i < n; i++) {
-		uint8_t *cmd = global_ipc->comp_data; // why is it a void*?
 
-		cmd[i + 4] = fuzz_in[i + 1];
+	struct sof_ipc_comp *cc = global_ipc->comp_data;
```
marc-hb, line 90:
```
src/platform/posix/ipc.c:90:30: warning: unused variable ‘cc’ [-Wunused-variable]
   90 |         struct sof_ipc_comp *cc = global_ipc->comp_data;
      |                              ^~
```

marc-hb (commented):
I just gave this a quick run. It works for me except the totally cryptic, DTS (!) error message when missing clang and llvm. Most likely unrelated to this.

src/platform/posix/include/platform/platform.h
lyakh: IIRC there is a weird (IMHO) C feature, that yes, you can use this in a header and the compiler will "guess what you actually mean" and only create one instance of the object, but I'd rather use an explicit `extern` the same way as we do everywhere, particularly since at least me personally I'm not sure which C standard we settled with and from which version that feature was introduced

marc-hb: There is a very good description of that portability issue at the end of the chapter "Declarations" in the Harbison & Steele (which every C developer should have IMHO). For maximum portability across various toolchains (which I think matters to SOF), they definitely agree with @lyakh .

Follow-up PR? Could also fix some other checkpatch warnings inhttps://github.com/thesofproject/sof/actions/runs/4118174248/jobs/7110383634

plbossart: humm, the use of 'struct foo' is very very common in Linux headers. I'd be surprised if you saw any 'extern' in any of the ALSA/ASoC headers...

andyross: FWIW, this is a type declaration, not an object definition.  All it means is that "struct sof exists as a type and you can legally take a pointer to it".  You see this construct in C++ a lot, because it avoids the need to pull in the very expensive headers that define types.   The reason it was introduced here is that I had a spot further down where I declare a function that takes a "struct sof *" as an argument, and was too lazy to figure out where that was declared.

marc-hb: > FWIW, this is a type declaration, not an object definition.

Thanks @andyross , I read too fast and stand corrected (and I think you meant: "not an _object_... declaration" :-)

lyakh: > FWIW, this is a type declaration, not an object definition.

ouch... /me looking at those "computer glasses" and contemplates if he should wear them more frequently... Then hides away

lyakh: no, of course it's a perfectly fine forward type declaration. And I myself frequently suggest it when just a pointer to that type is needed - exactly as @andyross explained, and I think that's usually preferred over including a header just for that. And that's also why it's used in Linux, in SOF, etc. I just somehow misread it as `struct sof sof;`...

PR #7033, opened 2023-02-02
Math: Library: Hifi3 exponential function dev

ShriramShastry:
This check-in adds a HiFi version of the fixed point exponential function, which has the same operating range and accuracy as the fixed point exponential function but performs four times quicker.

image

### DIFF
--- a/src/include/sof/math/exp_fcn.h
+++ b/src/include/sof/math/exp_fcn.h
@@ -13,8 +13,12 @@
 #if defined(__XCC__)
 /* HiFi */
 #include <xtensa/config/core-isa.h>
-#if XCHAL_HAVE_HIFI4 == 1
+#if XCHAL_HAVE_HIFI5 == 1
+#define SOFM_EXPONENTIAL_HIFI5 1
+#elif XCHAL_HAVE_HIFI4 == 1
 #define SOFM_EXPONENTIAL_HIFI4 1
+#elif XCHAL_HAVE_HIFI3 == 1
+#define SOFM_EXPONENTIAL_HIFI3 1
 #endif
 #else
 /* !XCC */
--- a/src/math/CMakeLists.txt
+++ b/src/math/CMakeLists.txt
@@ -15,7 +15,7 @@ if(CONFIG_SQRT_FIXED)
 endif()
 
 if(CONFIG_MATH_EXP)
-        add_local_sources(sof exp_fcn.c exp_fcn_hifi4.c)
+        add_local_sources(sof exp_fcn.c exp_fcn_hifi.c)
 endif()
 
 if(CONFIG_MATH_DECIBELS)
--- a/src/math/exp_fcn_hifi.c
+++ b/src/math/exp_fcn_hifi.c
@@ -9,19 +9,29 @@
 #include <sof/math/exp_fcn.h>
 #include <sof/common.h>
 #include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
 #include <stdint.h>
 #include <stddef.h>
+#include <errno.h>
 
-#if defined(SOFM_EXPONENTIAL_HIFI4)
+#if defined(SOFM_EXPONENTIAL_HIFI3) || defined(SOFM_EXPONENTIAL_HIFI4) || \
+	    defined(SOFM_EXPONENTIAL_HIFI5)
+
+#if XCHAL_HAVE_HIFI5
+#include <xtensa/tie/xt_hifi5.h>
+#elif XCHAL_HAVE_HIFI4
 #include <xtensa/tie/xt_hifi4.h>
+#else
+#include <xtensa/tie/xt_hifi3.h>
+#endif
 
+#define SOFM_CONVERG_ERROR 28823037624320LL /* error smaller than 1e-4,1/2 ^ -44.7122876209085 */
 #define SOFM_BIT_MASK_LOW_Q27P5 0x0000000008000000
-#define SOFM_BIT_MASK_Q62P2 0x4000000000000000
-#define SOFM_CONVERG_ERROR 0x1A36E2EB4000LL /* error smaller than 1e-4,1/2 ^ -44.7122876209085 */
-#define SOFM_QUOTIENT_SCALE 0x400000000
+#define SOFM_BIT_MASK_Q62P2 0x4000000000000000LL
+#define SOFM_QUOTIENT_SCALE BIT(30)
 #define SOFM_TERMS_Q23P9 0x800000
-#define SOFM_LSHIFT_BITS BIT(13)
-
+#define SOFM_LSHIFT_BITS 0x2000
 /*
  * Arguments	: int64_t in_0
  *		  int64_t in_1
@@ -75,10 +85,13 @@
  * not result in an overflow.
  *
  */
-static void mul_s64(ae_int64 in_0, ae_int64 in_1, ae_int64 *ptroutbitshi, ae_int64 *ptroutbitslo)
+static void mul_s64(ae_int64 in_0, ae_int64 in_1, ae_int64 *__restrict__ ptroutbitshi,
+		    ae_int64 *__restrict__ ptroutbitslo)
 {
 	ae_int64 producthihi, producthilo, productlolo;
-	ae_int64 producthi, carry, product_hl_lh_h, product_hl_lh_l;
+	ae_int64 producthi, productlo, product_hl_lh_h, product_hl_lh_l, carry;
+
+#if (SOFM_EXPONENTIAL_HIFI4 == 1 || SOFM_EXPONENTIAL_HIFI5 == 1)
 
 	ae_int32x2 in0_32 = AE_MOVINT32X2_FROMINT64(in_0);
 	ae_int32x2 in1_32 = AE_MOVINT32X2_FROMINT64(in_1);
@@ -111,8 +124,52 @@ static void mul_s64(ae_int64 in_0, ae_int64 in_1, ae_int64 *ptroutbitshi, ae_int
 	producthi = AE_ADD64(producthihi, carry);
 
 	producthi = AE_ADD64(producthi, product_hl_lh_h);
-
 	*ptroutbitslo = productlolo;
+#elif SOFM_EXPONENTIAL_HIFI3 == 1
+
+	ae_int64 producthi_1c;
+	ae_int64 producthi_2c;
+	ae_int64 productlo_2c;
+
+	ae_int64 s0 = AE_SRLI64(in_0, 63);
+	ae_int64 s1 = AE_SRLI64(in_1, 63);
+	bool x_or = (bool)((int)(int64_t)s0 ^ (int)(int64_t)s1);
+
+	ae_int32x2 in0_32 = AE_MOVINT32X2_FROMINT64(AE_ABS64(in_0));
+	ae_int32x2 in1_32 = AE_MOVINT32X2_FROMINT64(AE_ABS64(in_1));
+
+	producthihi = AE_MUL32_HH(in0_32, in1_32);
+	producthilo = AE_MUL32U_LL(in1_32, AE_MOVINT32X2_FROMINT64
+				  ((AE_MOVINT64_FROMINT32X2(in0_32) >> 32)));
+	producthilo += AE_MUL32U_LL(AE_MOVINT32X2_FROMINT64
+				  ((AE_MOVINT64_FROMINT32X2(in1_32) >> 32)), in0_32);
+	productlolo = AE_MUL32U_LL(in0_32, in1_32);
+
+	product_hl_lh_h = AE_SRAI64(producthilo, 32);
+	product_hl_lh_l = AE_SLAI64(producthilo, 32);
+
+	productlo = AE_ADD64(productlolo, product_hl_lh_l);
+	producthi = AE_ADD64(producthihi, product_hl_lh_h);
+
+	carry = AE_ADD64(AE_SRLI64(productlolo, 1), AE_SRLI64(product_hl_lh_l, 1));
+	carry = AE_SRLI64(carry, 63);
+	producthi = AE_ADD64(producthi, carry);
+
+	producthi_1c = AE_NOT64(producthi);
+	producthi_2c = AE_NEG64(producthi);
+	productlo_2c = AE_NEG64(productlo);
+
+	if (x_or) {
+		if (productlo == (ae_int64)0) {
+			producthi = producthi_2c;
+		} else {
+			producthi = producthi_1c;
+			productlo = productlo_2c;
+		}
+	}
+	*ptroutbitslo = productlo;
+#endif //(XCHAL_HAVE_HIFI4 || XCHAL_HAVE_HIFI5)
+
 	*ptroutbitshi = producthi;
 }
 
@@ -121,16 +178,15 @@ static void mul_s64(ae_int64 in_0, ae_int64 in_1, ae_int64 *ptroutbitshi, ae_int
  *		  int64_t b
  * Return Type	: int64_t
  */
-static int64_t mul_s64_sr_sat_near(int64_t a, int64_t b)
+static int64_t lomul_s64_sr_sat_near(int64_t a, int64_t b)
 {
-	ae_int64 result;
 	ae_int64 u64_chi;
 	ae_int64 u64_clo;
 	ae_int64 temp;
 
 	mul_s64(a, b, &u64_chi, &u64_clo);
 
-	ae_int64 roundup = AE_AND64(u64_clo, SOFM_SOFM_BIT_MASK_LOW_Q27P5);
+	ae_int64 roundup = AE_AND64(u64_clo, SOFM_BIT_MASK_LOW_Q27P5);
 
 	roundup = AE_SRLI64(roundup, 27);
 	temp = AE_OR64(AE_SLAI64(u64_chi, 36), AE_SRLI64(u64_clo, 28));
@@ -187,8 +243,6 @@ int32_t sofm_exp_int32(int32_t x)
 	ae_int64 ts = SOFM_TERMS_Q23P9;
 	ae_int64 mp = (x + SOFM_LSHIFT_BITS) >> 14; /* x in Q50.14 */;
 	xtbool flag;
-
-	int64_t qt_temp;
 	int64_t b_n;
 
 	mul_s64(mp, SOFM_BIT_MASK_Q62P2, &outhi, &outlo);
@@ -198,7 +252,7 @@ int32_t sofm_exp_int32(int32_t x)
 
 	ts = AE_ADD64(ts, temp);
 
-	mp = mul_s64_sr_sat_near(mp, (int64_t)x);
+	mp = lomul_s64_sr_sat_near(mp, (int64_t)x);
 
 	for (b_n = 0; b_n < 64;) {
 		AE_L64_IP(onebyfact, ponebyfact_Q63, 8);
@@ -209,7 +263,7 @@ int32_t sofm_exp_int32(int32_t x)
 		temp = AE_SRAI64(AE_ADD64(qt, SOFM_QUOTIENT_SCALE), 35);
 		ts = AE_ADD64(ts, temp);
 
-		mp = mul_s64_sr_sat_near(mp, (int64_t)x);
+		mp = lomul_s64_sr_sat_near(mp, (int64_t)x);
 
 		const ae_int64 sign = AE_NEG64(qt);
 
--- a/src/math/exp_fcn.c
+++ b/src/math/exp_fcn.c
@@ -12,6 +12,8 @@
 #include <rtos/bit.h>
 #include <stdbool.h>
 #include <stdint.h>
+#include <stddef.h>
+#include <errno.h>
 
 #if defined(EXPONENTIAL_GENERIC)
 
--- a/test/cmocka/src/math/arithmetic/CMakeLists.txt
+++ b/test/cmocka/src/math/arithmetic/CMakeLists.txt
@@ -13,7 +13,7 @@ cmocka_test(base2_logarithm
 cmocka_test(exponential
 	exponential.c
 	${PROJECT_SOURCE_DIR}/src/math/exp_fcn.c
-	${PROJECT_SOURCE_DIR}/src/math/exp_fcn_hifi4.c
+	${PROJECT_SOURCE_DIR}/src/math/exp_fcn_hifi.c
 )
 cmocka_test(square_root
 	square_root.c
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -661,7 +661,7 @@ zephyr_library_sources_ifdef(CONFIG_SQRT_FIXED
 
 zephyr_library_sources_ifdef(CONFIG_MATH_EXP
 	${SOF_MATH_PATH}/exp_fcn.c
-	${SOF_MATH_PATH}/exp_fcn_hifi4.c
+	${SOF_MATH_PATH}/exp_fcn_hifi.c
 )
 
 zephyr_library_sources_ifdef(CONFIG_COMP_UP_DOWN_MIXER


src/include/sof/math/exp_fcn_hifi4.h
```diff
@@ -0,0 +1,36 @@
...
+#include <xtensa/config/core-isa.h>
+#if XCHAL_HAVE_HIFI4 == 1
+#undef EXPONENTIAL_GENERIC
+#define EXPONENTIAL_HIFI4
+#endif
+
+#endif
+
+/* define constants */
+#define BIT_MASK_LOW_Q27P5 0x0000000008000000
```
lgirdwood: Will need a prefix for these since these are in public API header e.g.
```MATH_EXP_```
or move them to private header.
@singalsu what prefix are you using today for maths APIs/Macros ?

File: src/math/exp_fcn_hifi4.c [hunk 2]
```diff
@@ -0,0 +1,213 @@
...
+	mul_s64(mp, BIT_MASK_Q62P2, &outhi, &outlo);
+	qt = AE_OR64(AE_SLAI64(outhi, 46), AE_SRLI64(outlo, 18));
+
+	temp = AE_SRAI64(AE_ADD64(qt, QUOTIENT_SCALE), 35);
+
+	ts = AE_ADD64(ts, temp);
+
+	mp = mul_s64_sr_sat_near(mp, (int64_t)x);
+
+	for (b_n = 0; b_n < ARRAY_SIZE(onebyfact_Q63); b_n++) {
```
lgirdwood: this may need some inline comments to explain the intrinsic steps.

test/cmocka/src/math/arithmetic/exponential_hifi4.c
```diff
@@ -0,0 +1,79 @@
...
+
+#include <sof/math/exp_fcn_hifi4.h>
+#include <sof/audio/format.h>
+#include <rtos/string.h>
+#include <sof/common.h>
+
+#define ULP_TOLERANCE 5.60032793
+#define ULP_SCALE 0.0000999999999749
+/* testvector in Q4.28, -5 to +5 */
+	static const int32_t iv[128] = {
```
lgirdwood: indentation looks out here. The static declaration should be column 0

lgirdwood (commented):
Hifi parts need review by @singalsu and @andrula-song

test/cmocka/src/math/arithmetic/exponential_hifi.c
```diff
@@ -0,0 +1,83 @@
...
+	    940580929,	 961717579,   982854229,   1003990879,	1025127529,  1046264179,
+	    1067400829,	 1088537479,  1109674129,  1130810779,	1151947429,  1173084079,
+	    1194220729,	 1215357380,  1236494030,  1257630680,	1278767330,  1299903980,
+	    1321040630,	 1342177280};
+#endif
+
+static void test_math_arithmetic_exponential_hifi4_fixed(void **state)
+{
+	(void)state;
+#if (EXPONENTIAL_HIFI3 == 1 || EXPONENTIAL_HIFI4 == 1 EXPONENTIAL_HIFI5 == 1)
```
andrula-song, line 53: #if (EXPONENTIAL_HIFI3 == 1 || EXPONENTIAL_HIFI4 == 1 || EXPONENTIAL_HIFI5 == 1)?

test/cmocka/src/math/arithmetic/exponential_hifi.c
```diff
@@ -0,0 +1,83 @@
...
+#include <sof/math/exp_fcn_hifi.h>
+#include <sof/audio/format.h>
+#include <rtos/string.h>
+#include <sof/common.h>
+
+#define ULP_TOLERANCE 5.60032793
+#define ULP_SCALE 0.0000999999999749
+/* testvector in Q4.28, -5 to +5 */
+#if (EXPONENTIAL_HIFI3 == 1 || EXPONENTIAL_HIFI4 == 1 EXPONENTIAL_HIFI5 == 1)
+	static const int32_t iv[128] = {
```
andrula-song, line 25: #if (EXPONENTIAL_HIFI3 == 1 || EXPONENTIAL_HIFI4 == 1 || EXPONENTIAL_HIFI5 == 1)?

andrula-song (changes requested):

src/include/sof/math/exp_fcn_hifi4.h
ShriramShastry: Please accept my sincere apologies;  I am unaware of the headers for the public API.
If you could leave me a note with some instructions, a website, etc., I would be grateful.

test/cmocka/src/math/arithmetic/exponential_hifi.c
ShriramShastry: Thanks.  I 'II fix .

test/cmocka/src/math/arithmetic/exponential_hifi.c
ShriramShastry: Yes, will do here as well. Thank you

src/math/exp_fcn_hifi.c
```diff
@@ -0,0 +1,270 @@
...
+		    ae_int64 *__restrict__ ptroutbitslo)
+{
+	ae_int64 producthihi, producthilo, productlolo;
+	ae_int64 producthi, carry, product_hl_lh_h, product_hl_lh_l, productlo;
+
+#if (EXPONENTIAL_HIFI4 == 1 || EXPONENTIAL_HIFI5 == 1)
+
+	ae_int32x2 in0_32 = AE_MOVINT32X2_FROMINT64(in_0);
+	ae_int32x2 in1_32 = AE_MOVINT32X2_FROMINT64(in_1);
+
```
andrula-song, line 91: ae_int32x2 in0_32 = AE_MOVINT32X2_FROMINT64(in0);
	ae_int32x2 in1_32 = AE_MOVINT32X2_FROMINT64(in1);

src/include/sof/math/exp_fcn_hifi4.h
singalsu: We don't have the MATH_ or math_ prefix in most of src/math and src/include/sof/math stuff. I think it would be good. @ShriramShastry could you do the math prefix add for you exp functions PRs. I can do a PR to address the existing functions similarly to get this cleaned up.

singalsu: Math_ could leave conflicts, so SOFM_ and sofm_ should be good. Can you add those Sriram?

src/math/exp_fcn_hifi.c
ShriramShastry: done

test/cmocka/src/math/arithmetic/exponential_hifi.c
lgirdwood: There should be a top level CONFIG_MATH_EXP in Kconfig and then the architecture level HIFI kconfigs can be used.

test/cmocka/src/math/arithmetic/exponential.c
```diff
@@ -68,7 +69,11 @@ static void test_math_arithmetic_exponential_fixed(void **state)
 	for (i = 0; i < 256; i++) {
 		gen_exp_testvector(a_tmp, b_tmp, &r, &b_i);
 		a_i = (double)b_i / (1 << 28);
+#if (EXPONENTIAL_HIFI4 == 1)
```
btian1, line 72: EXPONENTIAL_HIFI4, suggest this one longer, I remember there is another exp module, we need different with it.
also, does this hifi only works for HIFI4 dsp? if also works for HIFI3, please add HIFI3, it is an optimization, no need use EXPONENTIAL_HIFI4, how about just use:
XCHAL_HAVE_HIFI3 == 1 || XCHAL_HAVE_HIFI4 == 1

ShriramShastry: Ahaah, Please give it a little more time. I 'm in the  middle of resolving the merge conflict. I will notify once I am done. 
Thank you

ShriramShastry: @btian1 , I've made my next commit, and you can review the HIFi3 improvements on top of the HiFi4 and HiFi5 versions .

test/cmocka/src/math/arithmetic/exponential_hifi.c
ShriramShastry: Done !! I've incorporated your suggestion. Should I change others, such as `CONFIG_SQRT_FIXED, CONFIG_CORDIC_FIXED,` and so on?

src/math/exp_fcn_hifi.c
```diff
@@ -97,6 +118,49 @@ static void mul_s64(ae_int64 in_0, ae_int64 in_1, ae_int64 *ptroutbitshi, ae_int
...
+	ae_int64 s0 = AE_SRLI64(in_0, 63);
+	ae_int64 s1 = AE_SRLI64(in_1, 63);
+	bool x_r = (bool)((int)(int64_t)s0 ^ (int)(int64_t)s1);
+
+	ae_int32x2 in0_32 = AE_MOVINT32X2_FROMINT64(AE_ABS64(in_0));
+	ae_int32x2 in1_32 = AE_MOVINT32X2_FROMINT64(AE_ABS64(in_1));
+
+	producthihi = AE_MUL32_HH(in0_32, in1_32);
+	producthilo = AE_MUL32U_LL(in1_32, AE_MOVINT32X2_FROMINT64
+				  ((AE_MOVINT64_FROMINT32X2(in0_32) >> 32)));
```
singalsu, line 136: Would AE_SRAI64(x, 32) be faster here? Or would some select function AE_SEL32 do this high word get directly?

src/include/sof/math/exp_fcn_hifi.h
```diff
@@ -0,0 +1,53 @@
...
+#include <stdint.h>
+#include <stddef.h>
+
+#define EXPONENTIAL_GENERIC 1
+#if defined(__XCC__)
+
+#include <xtensa/config/core-isa.h>
+#if XCHAL_HAVE_HIFI5 == 1
+#undef  EXPONENTIAL_GENERIC
+#undef EXPONENTIAL_HIFI3
```
cujomalainey, line 21: why not just have one define and check equality in your code? E.g.
EXPONENTIAL_HIFI == 5

src/math/exp_fcn_hifi4.c
```diff
@@ -0,0 +1,215 @@
...
+	qt = AE_OR64(AE_SLAI64(outhi, 46), AE_SRLI64(outlo, 18));
+
+	temp = AE_SRAI64(AE_ADD64(qt, QUOTIENT_SCALE), 35);
+
+	ts = AE_ADD64(ts, temp);
+
+	mp = mul_s64_sr_sat_near(mp, (int64_t)x);
+
+	for (b_n = 0; b_n < 64;) {
+		ae_int64 sign;
```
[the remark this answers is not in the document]
cujomalainey: i think we are moving towards a c99 standard so this is no longer mandatory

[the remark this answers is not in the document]
marc-hb: This is compatible with C89 because it's at the start of a `{` block.

Reminder: we're in 2023 :-)

C99 would be this:
```C
	for (b_n = 0; b_n < 64;) {

                 // some code
    
		const ae_int64 sign = ...

```

src/include/sof/math/exp_fcn_hifi.h
marc-hb: There's also no need to #undef something that has never been defined before, this is very verbose and hard to read.

test/cmocka/src/math/arithmetic/exponential.c
```diff
@@ -68,7 +69,11 @@ static void test_math_arithmetic_exponential_fixed(void **state)
...
+		accum = sofm_hifi_exp_int32(b_i);
```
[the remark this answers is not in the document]
marc-hb: +1 and this may allow getting rid of most of the macro complexity above.

But wait... cmocka NEVER has HIFIx anyway, has it?

marc-hb (changes requested):

ShriramShastry (commented):
Thank you for looking at the code. I've made the necessary changes.

src/include/sof/math/exp_fcn_hifi.h
```diff
@@ -0,0 +1,45 @@
...
+#define __SOFM_EXP_HIFI_H__
+
+#include <sof/platform.h>
+#include <stdint.h>
+#include <stddef.h>
+
+#undef EXPONENTIAL_GENERIC
+#undef EXPONENTIAL_HIFI3
+#undef EXPONENTIAL_HIFI4
+#undef EXPONENTIAL_HIFI5
```
marc-hb, line 18: What is the purpose of all these `#undef`s? To keep things simple macros should have a single source of truth. They should not be defined in different places fighting with each other and `#undef`-ining each other.

src/include/sof/math/exp_fcn_hifi.h
```diff
@@ -0,0 +1,45 @@
...
+#if XCHAL_HAVE_HIFI5 == 1
+#define EXPONENTIAL_HIFI5 1
+
+#elif XCHAL_HAVE_HIFI4 == 1
+#define EXPONENTIAL_HIFI4 1
+
+#elif XCHAL_HAVE_HIFI3 == 1
+#define EXPONENTIAL_HIFI3 1
+#else
+#error "No HIFI3 found. Cannot build XCC EXP - HiFi module."
```
marc-hb, line 31:
```suggestion
#error "No XCHAL_HAVE_HIFIx found. Cannot build XCC EXP - HiFi module."
```

src/include/sof/math/exp_fcn_hifi.h
```diff
@@ -0,0 +1,45 @@
...
+#endif
+#else
+/* GCC */
+#if !defined(EXPONENTIAL_HIFI3) || !defined(EXPONENTIAL_HIFI4) || !defined(EXPONENTIAL_HIFI5)
```
marc-hb, line 35: How could `EXPONENTIAL_HIFIx` be defined when `__XCC__` is not?

src/include/sof/math/exp_fcn_hifi.h
```diff
@@ -0,0 +1,45 @@
...
+#define EXPONENTIAL_GENERIC 1
+#endif
+#endif
+
+#if defined(EXPONENTIAL_HIFI3) || defined(EXPONENTIAL_HIFI4) || defined(EXPONENTIAL_HIFI5)
+
+int32_t sofm_exp_int32(int32_t x);
```
marc-hb, line 42: Move this to some existing `#ifdef` section to minimize their number.

marc-hb (changes requested):

marc-hb:
> compiler display error as we cannot assign to variable 'sign' with const-qualified type 'const ae_int64' as this is changing

Only when you separate the declaration and the definition. I think you're confusing "change" with "initialization". 

This works in C99:

```C
	for (b_n = 0; b_n < 64;) {
		AE_L64_IP(onebyfact, ponebyfact_Q63, 8);

                ....

		const ae_int64 sign = AE_NEG64(qt);

		flag = AE_LT64(qt, 0);
		AE_MOVT64(qt, sign, flag);

```

marc-hb:
> What is the purpose of all these #undefs? To keep things simple macros should have a single source of truth. They should not be defined in different places fighting with each other and #undef-ining each other.

#7214 is a good example of how to use a lot less and simpler macros to achieve what seems to be the same thing. It has no `#undef`.

johnylin76 (commented):
The function design looks good to me.
The only IMO comment is that I am more fond of codes split into .c/.h files for each implementation (generic/HIFI2/3/...), for example like FIR. Although it may bring some duplicated code it's much easier for code tracing.

marc-hb:
The last series https://github.com/thesofproject/sof/commits/714925fbcfa3eb3cd6331e604d3a81b4a85f8ce4 that you just pushed is a totally unreviewable git mess. There are 4 commits with more or less the same commit message and no logic about which code change belongs to which commit. The [3rd commit](https://github.com/thesofproject/sof/pull/7033/commits/9f4064847b076112057864209ffa99f47ecca763) moves the `rimage` submodule backwards 45 commits?!

Screenshot 2023-05-30 at 10 04 44


Screenshot 2023-05-30 at 10 04 19

ShriramShastry:
> The last series https://github.com/thesofproject/sof/commits/714925fbcfa3eb3cd6331e604d3a81b4a85f8ce4 that you just pushed is a totally unreviewable git mess. There are 4 commits with more or less the same commit message and no logic about which code change belongs to which commit. The [3rd commit](https://github.com/thesofproject/sof/pull/7033/commits/9f4064847b076112057864209ffa99f47ecca763) moves the `rimage` submodule backwards 45 commits?!
> 
> https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes
> 
> Screenshot 2023-05-30 at 10 04 44 Screenshot 2023-05-30 at 10 04 19

Thank you!! I made a mistake; please allow me some time to amend and push the change.

src/include/sof/math/exp_fcn_hifi4.h
```diff
@@ -0,0 +1,35 @@
...
+ *
+ */
+#ifndef __SOF_MATH_EXPONENTIAL_HIFI4_H__
+#define __SOF_MATH_EXPONENTIAL_HIFI4_H__
+
+#include <stdint.h>
+#include <stddef.h>
+
+#undef EXPONENTIAL_GENERIC
+#undef EXPONENTIAL_HIFI4
```
marc-hb, line 15: How could these be defined before any SOF `#include`?

Stop undef-ining things "just in case"; we want the pre-processor to catch multiple definitions and fail.

src/include/sof/math/exp_fcn_hifi4.h
```diff
@@ -0,0 +1,35 @@
...
+#if defined(__XCC__)
+
+#include <xtensa/config/core-isa.h>
+#if XCHAL_HAVE_HIFI4 == 1
+ #define EXPONENTIAL_HIFI4 1
+#else
+ #define EXPONENTIAL_GENERIC 1
```
marc-hb, line 22: This looks more or less like a duplicate of `exp_fcn_hifi.h`. Why not just include that file?

src/include/sof/math/exp_fcn_hifi.h
```diff
@@ -0,0 +1,36 @@
...
+#elif XCHAL_HAVE_HIFI4 == 1
+#define EXPONENTIAL_HIFI4 1
+
+#elif XCHAL_HAVE_HIFI3 == 1
+#define EXPONENTIAL_HIFI3 1
+#else
+#error "No XCHAL_HAVE_HIFIx found. Cannot build XCC EXP - HiFi module."
+#endif
+#else
+/* GCC */
```
marc-hb, line 30:
```suggestion
/* not XCC */
```
This could be clang or anything.

marc-hb (dismissed):
`#undef` do not seem required

ShriramShastry (commented):
Thank you for your feedback; I made the necessary improvements.

src/include/sof/math/exp_fcn_hifi4.h
```diff
@@ -0,0 +1,31 @@
...
+#include <stddef.h>
+
+#if defined(__XCC__)
+/* HiFi */
+#include <xtensa/config/core-isa.h>
+#if XCHAL_HAVE_HIFI4 == 1
+#define EXPONENTIAL_HIFI4 1
+
+#else
+#error "No XCHAL_HAVE_HIFIx found. Cannot build XCC EXP - HiFi module."
```
singalsu, line 21: Does this cause problem with HiFi3 build that has XCC but no HIFI4?

src/include/sof/math/exp_fcn_hifi.h
```diff
@@ -0,0 +1,36 @@
...
+#if defined(__XCC__)
+/* HiFi */
+#include <xtensa/config/core-isa.h>
+#if XCHAL_HAVE_HIFI5 == 1
+#define EXPONENTIAL_HIFI5 1
+
+#elif XCHAL_HAVE_HIFI4 == 1
+#define EXPONENTIAL_HIFI4 1
+
+#elif XCHAL_HAVE_HIFI3 == 1
```
singalsu, line 24: The 2nd commit adds handling of HIFI3 but I think the first commit would break bisect with Hifi3.

src/math/CMakeLists.txt
```diff
@@ -14,12 +14,8 @@ if(CONFIG_SQRT_FIXED)
         add_local_sources(sof sqrt_int16.c)
 endif()
 
-if(CONFIG_EXP_FIXED)
-        add_local_sources(sof exp_fcn.c)
-endif()
-
-if(CONFIG_EXP_FIXED)
-        add_local_sources(sof exp_fcn_hifi4.c)
+if(CONFIG_MATH_EXP)
```
singalsu, line 17: Should this change be in first commit? It's now changed in both commits.

src/include/sof/math/exp_fcn_hifi.h
marc-hb: So why are testing it if you know it's not defined?

src/include/sof/math/exp_fcn_hifi4.h
marc-hb: yet another reason not to partially duplicate [exp_fcn_hifi.h](https://github.com/thesofproject/sof/pull/7033/files#diff-e3d5145d9d47f335c1fa1b9eaccd78946878c3fcb340053c82e0279574f51b23) but to include it instead.

ShriramShastry (commented):
Thank you for reviewing; please read my inline responses.

src/include/sof/math/exp_fcn_hifi.h
marc-hb: HiFi2 does not matter. Right now you have this code:

```C
#if defined(__XCC__)
...
#  define EXPONENTIAL_HIFI4
...
#else
...
#  if !defined(EXPONENTIAL_HIFI4) ...
...
#endif
```

EXPONENTIAL_HIFI4 is defined only in the top-level `#if` (as it should).  So of course it's not going to be defined in the top-level `#else`, testing it there makes no sense and is just super-confusing for the reader.

src/include/sof/math/exp_fcn_hifi4.h
marc-hb: Why not just include `exp_fcn_hifi.h` here?

ShriramShastry: Sorry, I missed your comment here. The commit only applies to HiFi4 code.

src/math/exp_fcn_hifi.c
```diff
@@ -0,0 +1,276 @@
...
+	AE_ADD72(ep_lolo, productlolo, ep_HL_LH, product_hl_lh_l);
+
+	carry = AE_SRAI72(ep_lolo, productlolo, 32);
+
+	carry = AE_SRLI64(carry, 32);
+	producthi = AE_ADD64(producthihi, carry);
+
+	producthi = AE_ADD64(producthi, product_hl_lh_h);
+	*ptroutbitslo = productlolo;
+#elif EXPONENTIAL_HIFI3 == 1
```
marc-hb, line 123: HIFI3 shares no code with HIFI4 and HIFI5, it is effectively a different function so keep it separate for clarity. Something like  this:

```C
#if (EXPONENTIAL_HIFI4 == 1 || EXPONENTIAL_HIFI5 == 1)
#  define MUL_S64(...) mul_s64_hifi45(...)
#elif EXPONENTIAL_HIFI3 == 1
#  define MUL_S64(...) mul_s64_hifi3(...)
#else
#error should not happen
#fi
```

lgirdwood (commented):
@andrula-song pls review.

src/math/exp_fcn_hifi.c
marc-hb: Please don't concatenate these two, 99% different `mul_s64()` functions into one.  Give each a different name.

src/include/sof/math/exp_fcn.h
```diff
@@ -13,8 +13,12 @@
 #if defined(__XCC__)
 /* HiFi */
 #include <xtensa/config/core-isa.h>
-#if XCHAL_HAVE_HIFI4 == 1
+#if XCHAL_HAVE_HIFI5 == 1
```
singalsu, line 16: Update also commit message of this patch, since this is adding the hifi3 version and some hifi4 & 5 logistics.

singalsu (changes requested):

singalsu:
@ShriramShastry Any updates? I think only commit messages updates are needed, the rest look OK.

lgirdwood:
> @ShriramShastry Any updates? I think only commit messages updates are needed, the rest look OK.

@ShriramShastry ping - need to close v2.7 merge window soon.

src/math/exp_fcn_hifi4.c
```diff
@@ -6,6 +6,7 @@
  *
  */
 
+#include <sof/math/exp_fcn.h>
```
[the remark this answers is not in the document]
ShriramShastry: Done

src/math/exp_fcn_hifi.c
```diff
@@ -9,19 +9,30 @@
...
+#include <xtensa/config/core-isa.h>
+#if XCHAL_HAVE_HIFI5
+#include <xtensa/tie/xt_hifi5.h>
+#elif XCHAL_HAVE_HIFI4
 #include <xtensa/tie/xt_hifi4.h>
+#else
+#include <xtensa/tie/xt_hifi3.h>
+#endif
 
+#define SOFM_CONVERG_ERROR 28823037624320LL /* error smaller than 1e-4,1/2 ^ -44.7122876209085 */
```
btian1, line 30: add comments to show its Q format? like Q31?

src/include/sof/math/exp_fcn.h
ShriramShastry: Done

lgirdwood:
@singalsu @andrula-song good for you now ?

src/math/exp_fcn_hifi.c
```diff
@@ -110,8 +125,52 @@ static void mul_s64(ae_int64 in_0, ae_int64 in_1, ae_int64 *ptroutbitshi, ae_int
...
 	producthi = AE_ADD64(producthi, product_hl_lh_h);
-
 	*ptroutbitslo = productlolo;
+#elif SOFM_EXPONENTIAL_HIFI3 == 1
+
+	ae_int64 producthi_1c;
+	ae_int64 producthi_2c;
+	ae_int64 productlo_2c;
+
+	ae_int64 s0 = AE_SRLI64(in_0, 63);
```
andrula-song, line 135: AE_SRAI64() should be better?

src/math/exp_fcn_hifi.c
```diff
@@ -110,8 +125,52 @@ static void mul_s64(ae_int64 in_0, ae_int64 in_1, ae_int64 *ptroutbitshi, ae_int
...
+				  ((AE_MOVINT64_FROMINT32X2(in1_32) >> 32)), in0_32);
+	productlolo = AE_MUL32U_LL(in0_32, in1_32);
+
+	product_hl_lh_h = AE_SRAI64(producthilo, 32);
+	product_hl_lh_l = AE_SLAI64(producthilo, 32);
+
+	productlo = AE_ADD64(productlolo, product_hl_lh_l);
+	producthi = AE_ADD64(producthihi, product_hl_lh_h);
+
+	carry = AE_ADD64(AE_SRLI64(productlolo, 1), AE_SRLI64(product_hl_lh_l, 1));
```
andrula-song, line 155: here too.

src/math/exp_fcn_hifi.c
ShriramShastry: Why would sign extension be preferred over zero? I saw that several test samples were failing.

test/cmocka/src/math/arithmetic/exponential.c
```diff
@@ -68,6 +68,7 @@ static void test_math_arithmetic_exponential_fixed(void **state)
 	for (i = 0; i < 256; i++) {
```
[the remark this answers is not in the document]
ShriramShastry: There is no filter here; `256` is a number that represents the number of random number test samples that are tested each time.

src/math/exp_fcn_hifi.c
ShriramShastry: There is no need to continue the entering sign, implying that the requirement is to take absolute value.

According to the documentation, there is a distinction between `AE_SRLI64` and `AE_SRAI64`.
AE_SRLI64 is a 64-bit logical (`zero-extending`) right shift of AE_DR register d0 by immediate, with result placed in d.
```
proto AE_SRLI64 {out ae_int64 d, in ae_int64 d0, in immediate sa} {} {
    AE_SRLI64 d, d0, sa;
}
```

AE_SRAI64 is a 64-bit arithmetic (`sign-extending`) right shift of AE_DR register d0 by immediate, with result placed in d.

```
proto AE_SRAI64 {out ae_int64 d, in ae_int64 d0, in immediate sa} {} {
    AE_SRAI64 d, d0, sa;
}
```

src/include/sof/math/exp_fcn.h
cujomalainey: Why is this ifdef needed? you can just check the HiFi directly in the file

src/math/exp_fcn_hifi.c
```diff
@@ -97,6 +118,49 @@ static void mul_s64(ae_int64 in_0, ae_int64 in_1, ae_int64 *ptroutbitshi, ae_int
...
+	carry = AE_SRLI64(carry, 63);
+	producthi = AE_ADD64(producthi, carry);
+
+	producthi_1c = AE_NOT64(producthi);
+	producthi_2c = AE_NEG64(producthi);
+	productlo_2c = AE_NEG64(productlo);
+
+	if (x_r) {
+		if (productlo == (ae_int64)0) {
```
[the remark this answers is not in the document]
cujomalainey: does `if (!productlo)` not work?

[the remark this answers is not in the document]
ShriramShastry: `if (!productlo)` with I see following error

`error: no operator overload defined for unary operator '!' (operand type 'ae_int64' (aka '_TIE_xt_hifi2_ae_int64'))`

src/include/sof/math/exp_fcn.h
ShriramShastry: Then check is then required in all C code and HiFi intrinsic.

Depending on performance, build support ranges from 3 to 5 with a single common placed `#ifdef` that divides Generic C code and HiFi intrinsic that spans over all active settings. The test is being run by cmocka test using the standard function `sofm_exp_int32()`. will also need to bring along more #ifdef gear.

Consequently, I believe #ifdef are necessary.

test/cmocka/src/math/arithmetic/exponential.c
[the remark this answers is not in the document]
marc-hb: > use a macro to define 256 and describe what it is

src/math/exp_fcn_hifi.c
```diff
@@ -6,21 +6,33 @@
...
+#include <stdint.h>
 #include <stdint.h>
 #include <stddef.h>
+#include <errno.h>
+
+#if defined(SOFM_EXPONENTIAL_HIFI3) || defined(SOFM_EXPONENTIAL_HIFI4) || \
+	    defined(SOFM_EXPONENTIAL_HIFI5)
 
-#if defined(SOFM_EXPONENTIAL_HIFI4)
+#include <xtensa/config/core-isa.h>
```
marc-hb, line 21: It makes no sense to include this file again. `SOFM_` are defined in `exp_fcn.h` which already includes `core-isa.h`.

marc-hb (changes requested):
Clarify the purpose of the `SOFM_` macros

src/math/exp_fcn_hifi.c
lgirdwood: @singalsu @andrula-song dont we have a high level Kconfig that sets the HIFI version that can be reused ?
@ShriramShastry are we using this header only for detection of HiFi version, or do we also need it for other reasons ?

marc-hb: I don't understand your answer, I think you missed my (very simple) point and that we're talking across each other. Trying to rephrase my simple point:

If any `SOFM_EXPx` is defined, then `<xtensa/config/core-isa.h>` was already included by `exp_fcn.h`. So, just don't `#include` that same file  again.

More generally speaking: testing and debugging `#include` files is extremely time-consuming and difficult. So all `#ifdef` and `#include` logic must be as simple as possible and very well commented in the source code. If some pre-processor stuff is not obvious then it must be either simplified or explained in the source.

ShriramShastry: Yup, including `isa.h` is a mistake in `exp_fcn_hifi.c`. I will include it.

singalsu (approved):

src/include/sof/math/exp_fcn.h
singalsu: It's useful to have the added layer of generic/hifiN macros to e.g. build generic version with hifi toolchain. It could be simplified a bit but since it's only macros and actual code I'd not require perfection.

lgirdwood:
@andrula-song good for you now ?

src/include/sof/math/exp_fcn.h
lgirdwood: @singalsu we have a bit of a mess around HIFI definitions atm e.g.
```
grep -rn HIFI4 src/audio/
src/audio/src/src_hifi4.c:11:#if SRC_HIFI4
src/audio/module_adapter/module/volume/volume_hifi4.c:24:#ifdef VOLUME_HIFI4
src/audio/module_adapter/module/volume/volume.c:1211:#ifdef VOLUME_HIFI4
src/audio/module_adapter/module/volume/volume.c:1246:#if XCHAL_HAVE_HIFI3 || XCHAL_HAVE_HIFI4
src/audio/module_adapter/module/volume/volume_hifi4_with_peakvol.c:9: * \brief Volume HIFI4 processing implementation with peak volume detection
src/audio/module_adapter/module/volume/volume_hifi4_with_peakvol.c:24:#ifdef VOLUME_HIFI4
src/audio/mixer/mixer_hifi3.c:10:#if __XCC__ && (XCHAL_HAVE_HIFI3 || XCHAL_HAVE_HIFI4)
src/audio/mixer/mixer.c:199:#if XCHAL_HAVE_HIFI3 || XCHAL_HAVE_HIFI4
src/audio/dcblock/dcblock_hifi4.c:12:#ifdef DCBLOCK_HIFI4
src/audio/copier/copier_hifi.c:8:#if __XCC__ && (XCHAL_HAVE_HIFI3 || XCHAL_HAVE_HIFI4)
src/audio/component.c:26:#if XCHAL_HAVE_HIFI3 || XCHAL_HAVE_HIFI4
```
This is not the PR to fix those, but we do need a high level SIMD level Kconfig that could be used by all modules and could also be used to downgrade if needed (for debug).

lgirdwood (approved):
Lets fix have a top level SIMD level Kconfig as an improvement for v2.8.

src/include/sof/math/exp_fcn.h
marc-hb: > It's useful to have the added layer of generic/hifiN macros to e.g. build generic version with hifi toolchain.

Code speaks louder than words, here's an example from @singalsu (thanks!)
- #8108

btian1 (approved):
personally, I don't suggest spend time on hifi4 and hifi5 specific version, this is due to there is no actual usage for this exp function, and even there is a usage, as long as it is not in timing critical path, HIFI3 is fine for use.

andrula-song (approved):

PR #7038, opened 2023-02-03
kpb: add micselector function

fkwasowi:
Add micsel function to create a topology
with KD without a separate micselector module

### DIFF
--- a/src/audio/kpb.c
+++ b/src/audio/kpb.c
@@ -228,7 +228,7 @@ static void kpb_set_params(struct comp_dev *dev,
 	params->sample_valid_bytes =
 		kpb->ipc4_cfg.base_cfg.audio_fmt.valid_bit_depth / 8;
 	params->buffer_fmt = kpb->ipc4_cfg.base_cfg.audio_fmt.interleaving_style;
-	params->buffer.size = kpb->ipc4_cfg.base_cfg.obs * KPB_MAX_BUFF_TIME * params->channels;
+	params->buffer.size = kpb->ipc4_cfg.base_cfg.ibs * KPB_MAX_BUFF_TIME * params->channels;
 
 	params->host_period_bytes = params->channels *
 				    params->sample_container_bytes *
@@ -808,6 +808,33 @@ static int kpb_prepare(struct comp_dev *dev)
 			break;
 		default:
 			break;
+		}
+	}
+#else
+	/* Update number of sel_sink channels.
+	 * If OBS is not equal to IBS it means that KPB will work in micselector mode.
+	 */
+	if (kpb->ipc4_cfg.base_cfg.ibs != kpb->ipc4_cfg.base_cfg.obs) {
+		struct list_item *sink_list;
+		const uint32_t byte_align = 1;
+		const uint32_t frame_align_req = 1;
+		uint32_t sink_id;
+
+		list_for_item(sink_list, &dev->bsink_list) {
+			struct comp_buffer *sink =
+				container_of(sink_list, struct comp_buffer, source_list);
+			struct comp_buffer __sparse_cache *sink_c = buffer_acquire(sink);
+
+			audio_stream_init_alignment_constants(byte_align, frame_align_req,
+							      &sink_c->stream);
+			sink_id = sink_c->id;
+
+			if (sink_id == 0)
+				sink_c->stream.channels = kpb->num_of_sel_mic;
+			else
+				sink_c->stream.channels = kpb->config.channels;
+
+			buffer_release(sink_c);
 		}
 	}
 #endif /* CONFIG_IPC_MAJOR_4 */
--- a/src/audio/kpb.c
+++ b/src/audio/kpb.c
@@ -91,6 +91,10 @@ struct comp_data {
 #ifdef CONFIG_IPC_MAJOR_4
 	struct ipc4_kpb_module_cfg ipc4_cfg;
 #endif /* CONFIG_IPC_MAJOR_4 */
+	uint32_t num_of_sel_mic;
+	uint32_t num_of_in_channels;
+	uint32_t offsets[KPB_MAX_MICSEL_CHANNELS];
+	struct kpb_micselector_config mic_sel;
 };
 
 /*! KPB private functions */
@@ -808,9 +812,9 @@ static int kpb_prepare(struct comp_dev *dev)
 	}
 #endif /* CONFIG_IPC_MAJOR_4 */
 
-	if (!kpb->sel_sink || !kpb->host_sink) {
-		comp_info(dev, "kpb_prepare(): could not find sinks: sel_sink %p host_sink %p",
-			  kpb->sel_sink, kpb->host_sink);
+	if (!kpb->sel_sink) {
+		comp_err(dev, "kpb_prepare(): could not find sink: sel_sink %p",
+			 kpb->sel_sink);
 		ret = -EIO;
 	}
 
@@ -2018,9 +2022,9 @@ static inline bool validate_host_params(struct comp_dev *dev,
 		/* Host buffer size is too small - history data
 		 * may get overwritten.
 		 */
-		comp_err(dev, "kpb: host_buffer_size (%d) must be at least %d",
-			 host_buffer_size, HOST_BUFFER_MIN_SIZE(hb_size_req, kpb->config.channels));
-		return false;
+		comp_warn(dev, "kpb: host_buffer_size (%d) must be at least %d",
+			  host_buffer_size,
+			  HOST_BUFFER_MIN_SIZE(hb_size_req, kpb->config.channels));
 	} else if (kpb->sync_draining_mode) {
 		/* Sync draining allowed. Check if we can perform draining
 		 * with current settings.
@@ -2060,6 +2064,50 @@ static inline void kpb_change_state(struct comp_data *kpb,
 	kpb->state_log = (kpb->state_log << 4) | state;
 }
 
+static int kpb_set_micselect(struct comp_dev *dev, const void *data,
+			     int max_data_size)
+{
+	const struct kpb_micselector_config *mic_sel = data;
+	struct comp_data *kpb = comp_get_drvdata(dev);
+	const size_t mic_cnt = kpb->config.channels - KPB_REFERENCE_SUPPORT_CHANNELS;
+	const uint8_t valid_mask = KPB_COUNT_TO_BITMASK(mic_cnt);
+	size_t i;
+
+	if ((valid_mask & mic_sel->mask) == 0) {
+		comp_err(dev, "error: invalid micselector bit mask");
+		return -EINVAL;
+	}
+	 /* selected mics counter */
+	size_t num_of_sel_mic = 0;
+
+	for (i = 0; i < mic_cnt; i++) {
+		if (KPB_IS_BIT_SET(mic_sel->mask, i)) {
+			kpb->offsets[num_of_sel_mic] = i;
+			num_of_sel_mic++;
+		}
+	}
+	kpb->num_of_sel_mic = num_of_sel_mic;
+	kpb->num_of_in_channels = kpb->config.channels;
+	kpb->mic_sel.mask = mic_sel->mask;
+	return 0;
+}
+
+static int kpb_set_large_config(struct comp_dev *dev, uint32_t param_id,
+				bool first_block,
+				bool last_block,
+				uint32_t data_offset,
+				const char *data)
+{
+	comp_info(dev, "kpb_set_large_config()");
+
+	switch (param_id) {
+	case KP_BUF_CLIENT_MIC_SELECT:
+		return kpb_set_micselect(dev, data, data_offset);
+	default:
+		return -EINVAL;
+	}
+}
+
 static const struct comp_driver comp_kpb = {
 	.type = SOF_COMP_KPB,
 	.uid = SOF_RT_UUID(kpb_uuid),
@@ -2073,6 +2121,7 @@ static const struct comp_driver comp_kpb = {
 		.prepare	= kpb_prepare,
 		.reset		= kpb_reset,
 		.params		= kpb_params,
+		.set_large_config = kpb_set_large_config,
 #ifdef CONFIG_IPC_MAJOR_4
 		.get_attribute	= kpb_get_attribute,
 		.bind		= kpb_bind,
--- a/src/audio/kpb.c
+++ b/src/audio/kpb.c
@@ -924,6 +924,168 @@ static int kpb_reset(struct comp_dev *dev)
 	return ret;
 }
 
+#ifdef KPB_HIFI3
+#if CONFIG_FORMAT_S16LE
+static void kpb_micselect_copy16(struct comp_buffer __sparse_cache *sink,
+				 struct comp_buffer __sparse_cache *source, size_t size,
+				 uint32_t in_channels, uint32_t micsel_channels, uint32_t *offsets)
+{
+	struct audio_stream __sparse_cache *istream = &source->stream;
+	struct audio_stream __sparse_cache *ostream = &sink->stream;
+	uint16_t ch;
+	size_t i;
+
+	AE_SETCBEGIN0(ostream->addr);
+	AE_SETCEND0(ostream->end_addr);
+
+	buffer_stream_invalidate(source, size);
+	const ae_int16 *in_ptr = (const ae_int16 *)istream->r_ptr;
+	ae_int16x4 d16 = AE_ZERO16();
+	const size_t in_offset = in_channels * sizeof(ae_int16);
+	const size_t out_offset = micsel_channels * sizeof(ae_int16);
+	const size_t samples_per_chan = size / (sizeof(uint16_t) * micsel_channels);
+	ae_int16 *out_ptr;
+
+	for (ch = 0; ch < micsel_channels; ch++) {
+		const ae_int16 *input_data = (const ae_int16 *)(in_ptr) + offsets[ch];
+
+		out_ptr = (ae_int16 *)ostream->w_ptr;
+		out_ptr += ch;
+		for (i = 0; i < samples_per_chan; i++) {
+			AE_L16_XP(d16, input_data, in_offset);
+			AE_S16_0_XC(d16, out_ptr, out_offset);
+		}
+	}
+}
+#endif
+#if CONFIG_FORMAT_S24LE || CONFIG_FORMAT_S32LE
+static void kpb_micselect_copy32(struct comp_buffer __sparse_cache *sink,
+				 struct comp_buffer __sparse_cache *source, size_t size,
+				 uint32_t in_channels, uint32_t micsel_channels, uint32_t *offsets)
+{
+	struct audio_stream __sparse_cache *istream = &source->stream;
+	struct audio_stream __sparse_cache *ostream = &sink->stream;
+	uint16_t ch;
+	size_t i;
+
+	AE_SETCBEGIN0(ostream->addr);
+	AE_SETCEND0(ostream->end_addr);
+
+	buffer_stream_invalidate(source, size);
+
+	const ae_int32 *in_ptr = (const ae_int32 *)istream->r_ptr;
+	ae_int32x2 d32 = AE_ZERO32();
+	const size_t in_offset = in_channels * sizeof(ae_int32);
+	const size_t out_offset = micsel_channels * sizeof(ae_int32);
+	const size_t samples_per_chan = size / (sizeof(uint32_t) * micsel_channels);
+	ae_int32 *out_ptr;
+
+	for (ch = 0; ch < micsel_channels; ch++) {
+		const ae_int32 *input_data = (const ae_int32 *)(in_ptr) + offsets[ch];
+
+		out_ptr = (ae_int32 *)ostream->w_ptr;
+		out_ptr += ch;
+		for (i = 0; i < samples_per_chan; i++) {
+			AE_L32_XP(d32, input_data, in_offset);
+			AE_S32_L_XC(d32, out_ptr, out_offset);
+		}
+	}
+}
+#endif
+#else
+static void kpb_micselect_copy16(struct comp_buffer __sparse_cache *sink,
+				 struct comp_buffer __sparse_cache *source, size_t size,
+				 uint32_t in_channels,  uint32_t micsel_channels, uint32_t *offsets)
+{
+	struct audio_stream __sparse_cache *istream = &source->stream;
+	struct audio_stream __sparse_cache *ostream = &sink->stream;
+
+	buffer_stream_invalidate(source, size);
+	size_t out_samples;
+	uint16_t ch;
+
+	const int16_t *in_data;
+	int16_t *out_data;
+	const uint32_t samples_per_chan = size / (sizeof(uint16_t) * micsel_channels);
+
+	for (ch = 0; ch < micsel_channels; ch++) {
+		out_samples = 0;
+		in_data = (int16_t *)istream->r_ptr;
+		out_data = (int16_t *)ostream->w_ptr;
+
+		for (size_t i = 0; i < samples_per_chan * in_channels; i += in_channels) {
+			if (&out_data[out_samples + ch]
+					>= (int16_t *)ostream->end_addr) {
+				out_data = (int16_t *)ostream->addr;
+				out_samples = 0;
+			}
+			out_data[out_samples + ch] = in_data[i + offsets[ch]];
+			out_samples += micsel_channels;
+		}
+	}
+}
+
+static void kpb_micselect_copy32(struct comp_buffer __sparse_cache *sink,
+				 struct comp_buffer __sparse_cache *source, size_t size,
+				 uint32_t in_channels, uint32_t micsel_channels, uint32_t *offsets)
+{
+	struct audio_stream __sparse_cache *istream = &source->stream;
+	struct audio_stream __sparse_cache *ostream = &sink->stream;
+
+	buffer_stream_invalidate(source, size);
+	size_t out_samples;
+	uint16_t ch;
+	const int32_t *in_data;
+	int32_t *out_data;
+	const uint32_t samples_per_chan = size / (sizeof(uint32_t) * micsel_channels);
+
+	for (ch = 0; ch < micsel_channels; ch++) {
+		out_samples = 0;
+		in_data = (int32_t *)istream->r_ptr;
+		out_data = (int32_t *)ostream->w_ptr;
+
+		for (size_t i = 0; i < samples_per_chan * in_channels; i += in_channels) {
+			if (&out_data[out_samples + ch]
+					>= (int32_t *)ostream->end_addr) {
+				out_data = (int32_t *)ostream->addr;
+				out_samples = 0;
+			}
+			out_data[out_samples + ch] = in_data[i + offsets[ch]];
+			out_samples += micsel_channels;
+		}
+	}
+}
+#endif
+static void kpb_micselect_copy(struct comp_dev *dev, struct comp_buffer __sparse_cache *sink_c,
+			       struct comp_buffer __sparse_cache *source_c, size_t copy_bytes,
+			       uint32_t channels)
+{
+	struct comp_data *kpb = comp_get_drvdata(dev);
+	size_t sample_width = kpb->config.sampling_width;
+	uint32_t *offsets = kpb->offsets;
+
+	switch (sample_width) {
+#if CONFIG_FORMAT_S16LE
+	case 16:
+		kpb_micselect_copy16(sink_c, source_c, copy_bytes,
+				     channels, kpb->num_of_sel_mic, offsets);
+		break;
+#endif /* CONFIG_FORMAT_S16LE */
+#if CONFIG_FORMAT_S24LE || CONFIG_FORMAT_S32LE
+	case 24:
+		kpb_micselect_copy32(sink_c, source_c, copy_bytes,
+				     channels, kpb->num_of_sel_mic, offsets);
+		break;
+	case 32:
+		kpb_micselect_copy32(sink_c, source_c, copy_bytes,
+				     channels, kpb->num_of_sel_mic, offsets);
+		break;
+#endif /* CONFIG_FORMAT_S24LE || CONFIG_FORMAT_S32LE */
+	default:
+		comp_cl_err(&comp_kpb, "KPB: An attempt to copy not supported format!");
+		return;
+	}
+}
 /**
  * \brief Copy real time input stream into sink buffer,
  *	and in the same time buffers that input for
@@ -941,7 +1103,7 @@ static int kpb_copy(struct comp_dev *dev)
 	struct comp_data *kpb = comp_get_drvdata(dev);
 	struct comp_buffer *source, *sink;
 	struct comp_buffer __sparse_cache *source_c, *sink_c = NULL;
-	size_t copy_bytes = 0;
+	size_t copy_bytes = 0, produced_bytes = 0;
 	size_t sample_width = kpb->config.sampling_width;
 	struct draining_data *dd = &kpb->draining_task_data;
 	uint32_t avail_bytes;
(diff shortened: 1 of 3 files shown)


src/audio/kpb.c
```diff
@@ -91,6 +91,10 @@ struct comp_data {
 #ifdef CONFIG_IPC_MAJOR_4
 	struct ipc4_kpb_module_cfg ipc4_cfg;
 #endif /* CONFIG_IPC_MAJOR_4 */
+	uint32_t mic_sel_cnt_;
+	size_t num_of_in_channels_;
+	uint16_t offsets_[4];
+	struct kpb_micselector_config *mic_sel_;
```
lyakh, line 97: IMHO ending variable / structure names with an underscore is a part of the style and I don't think it's very popular in SOF in general and in this file in particular, I'd rather keep this uniform

src/audio/kpb.c
```diff
@@ -2060,6 +2068,60 @@ static inline void kpb_change_state(struct comp_data *kpb,
...
+	struct comp_data *kpb = comp_get_drvdata(dev);
+	struct ipc4_audio_format fmt_in = kpb->ipc4_cfg.base_cfg.audio_fmt;
+	const size_t mic_cnt = fmt_in.channels_count - KPB_REFERENCE_SUPPORT_CHANNELS;
+	const uint8_t valid_mask = COUNT_TO_BITMASK(mic_cnt);
+
+	if ((valid_mask & mic_sel->mask_) == 0) {
+		comp_err(dev, "error: invalid micselector bit mask");
+		return -EINVAL;
+	}
+	 /*selected mics counter*/
```
lyakh, line 2213: spaces around text, please

src/audio/kpb.c
```diff
@@ -91,6 +91,10 @@ struct comp_data {
 #ifdef CONFIG_IPC_MAJOR_4
 	struct ipc4_kpb_module_cfg ipc4_cfg;
 #endif /* CONFIG_IPC_MAJOR_4 */
+	uint32_t mic_sel_cnt_;
+	size_t num_of_in_channels_;
+	uint16_t offsets_[4];
```
lyakh, line 96: this array is assigned but I don't see it being used in this commit

src/audio/kpb.c
```diff
@@ -453,6 +457,10 @@ static struct comp_dev *kpb_new(const struct comp_driver *drv,
 	kpb->hd.c_hb = NULL;
 	kpb->kpb_no_of_clients = 0;
 	kpb->state_log = 0;
+	/* mask is set to zero to allow legacy coping for default config */
```
lyakh, line 460: "coping" or "copying?"

src/audio/kpb.c
```diff
@@ -453,6 +457,10 @@ static struct comp_dev *kpb_new(const struct comp_driver *drv,
...
+	kpb->mic_sel_->mask_ = 0x0;
+	kpb->mic_sel_cnt_ = 0;
+	kpb->num_of_in_channels_ = 0;
```
lyakh, line 463: in fact all these 0 assignments are redundant - `kpb` is allocated using `rzalloc()`. Using `rzalloc()` means we *want* it to be 0 and we rely on it and we use it. Not just do it "in case we forget something." I'd rather remove the above assignments instead of adding new ones.

src/audio/kpb.c
```diff
@@ -808,7 +816,7 @@ static int kpb_prepare(struct comp_dev *dev)
 	}
 #endif /* CONFIG_IPC_MAJOR_4 */
 
-	if (!kpb->sel_sink || !kpb->host_sink) {
+	if (!kpb->sel_sink) {
 		comp_info(dev, "kpb_prepare(): could not find sinks: sel_sink %p host_sink %p",
 			  kpb->sel_sink, kpb->host_sink);
```
lyakh, line 821: the text should be changed too. And BTW it probably should be at least `comp_warn()` if not outright `comp_err()`

src/include/sof/audio/kpb.h
```diff
@@ -58,6 +58,11 @@ struct comp_buffer;
 #define KPB_BYTES_TO_S16_SAMPLES(s)	((s) >> 1)
 #define KPB_BYTES_TO_S32_SAMPLES(s)	((s) >> 2)
 
+/* Macro that returns mask with selected bit set */
```
lyakh, line 61: "bits"

src/include/sof/audio/kpb.h
```diff
@@ -58,6 +58,11 @@ struct comp_buffer;
...
+#define COUNT_TO_BITMASK(cnt) (((0x1 << (cnt)) - 1))
+#define IS_BIT_SET(value, idx) ((value) & (1 << (idx)))
```
lyakh, line 63: `BIT(idx)`

src/audio/kpb.c
lyakh: Also, I tried to find logic in type selection but I couldn't. Why is number of microphones `uint32_t` and number of channels `size_t`? Why are offsets 16-bit? Do we expect more than 255 microphones?

src/audio/kpb.c
```diff
@@ -901,6 +901,130 @@ static int kpb_reset(struct comp_dev *dev)
...
+				out_samples = 0;
+				in_data = (uint16_t *)istream->r_ptr;
+				out_data = (uint16_t *)ostream->w_ptr;
+
+				for (size_t i = 0; i < samples_per_chan; i++) {
+					if (&out_data[out_samples * kpb->mic_sel_cnt_ + out_channel]
+							>= (uint16_t *)ostream->end_addr) {
+						out_data = (uint16_t *)ostream->addr;
+						out_samples = 0;
+					}
```
lyakh, line 992: I think you could optimise this too using our new strategy: pre-calculate how many samples fit until the end of the buffer, copy that, then wrap and copy the rest? You could make the copying loop into a small function to not duplicate the code

src/audio/kpb.c
```diff
@@ -973,34 +1097,39 @@ static int kpb_copy(struct comp_dev *dev)
 			ret = PPL_STATUS_PATH_STOP;
 			break;
 		}
+		if (kpb->mic_sel_cnt_ == 0) {
```
lyakh, line 1100: what does it mean if `mic_sel_cnt_ == 0`?

fkwasowi: The IPC large config set for microphone selection was not sent and kpb works as before - it simply copies the data.

src/audio/kpb.c
```diff
@@ -2060,6 +2068,50 @@ static inline void kpb_change_state(struct comp_data *kpb,
 	kpb->state_log = (kpb->state_log << 4) | state;
 }
 
+static int kpb_set_micselect(struct comp_dev *dev, const void *data,
+			     int max_data_size)
+{
+	comp_info(dev, "kpb_set_micselect");
```
kv2019i, line 2199: Style note: in SOF code we variables declared before any code, so comp_info() should be after the variable declarations. And this probably should be "comp_dbg()" as this just indicates the function is entered and provides no other information.

src/audio/kpb.c
```diff
@@ -2060,6 +2068,50 @@ static inline void kpb_change_state(struct comp_data *kpb,
...
+	const struct kpb_micselector_config *mic_sel = data;
+	struct comp_data *kpb = comp_get_drvdata(dev);
+	const size_t mic_cnt = kpb->config.channels - KPB_REFERENCE_SUPPORT_CHANNELS;
```
kv2019i, line 2202: Is this guaranteed config.channels is at least KBP_REFERENCE_SUPPORT_CHANNELS so that this doesn't become negative?

src/audio/kpb.c
fkwasowi: ,,offsets" is used in kpb_set_micselect function

kv2019i: Hmm @lyakh array is used in set_large_config ?

I wonder about the "[4]". Maybe this should be a define and there probably should be some comment why it is 4.

kv2019i (commented):
The code was refreshed in middle of my review, so some comments might be stale. Nothing really major. Minor cleanups. I'd also expand the commit a bit and explain in plain English waht "set the micselector" is. I mean this "allow to configure which mics are used by the kpb, using large_config_set interface". Or something like that.

src/audio/kpb.c
fkwasowi: Yes. KPB gets a minimum of 3 channels.

src/audio/kpb.c
fkwasowi: Done

src/include/sof/audio/kpb.h
lgirdwood: Also need to prefix new KPB macros with ```KPB_``` to avoid collisions with other header files.

src/include/sof/audio/kpb.h
```diff
@@ -150,6 +159,16 @@ struct history_data {
 	struct history_buffer *c_hb; /**< current buffer used for writing */
 };
 
+enum ipc4_kpb_module_config_params {
+	    /* Mic selector for client - sets microphone id for real time sink mic selector */
+	    KP_BUF_CLIENT_MIC_SELECT = 11,
```
lgirdwood, line 164: Why 11 ? if its for Windows compatibility we should say so that no one changes it later.

src/audio/kpb.c
```diff
@@ -897,6 +897,130 @@ static int kpb_reset(struct comp_dev *dev)
...
+		const ae_int16 *in_ptr = (const ae_int16 *)istream->r_ptr;
+		ae_int16x4 d16 = AE_ZERO16();
+		const size_t in_offset = kpb->num_of_in_channels * sizeof(ae_int16);
+		const size_t out_offset = kpb->mic_sel_cnt * sizeof(ae_int16);
+		const size_t samples_per_chan = size / (sizeof(uint16_t) * kpb->mic_sel_cnt);
+		ae_int16 *buf = (ae_int16 *)(ostream->addr);
+		ae_int16 *buf_end = (ae_int16 *)(ostream->end_addr);
+		ae_int16 *out_ptr = (ae_int16 *)buf;
+
+		AE_SETCBEGIN0(buf);
```
lgirdwood, line 923: @fkwasowi good to have another Hifi developer, will add you to other HiFi reviews.

src/audio/kpb.c
```diff
@@ -969,34 +1093,39 @@ static int kpb_copy(struct comp_dev *dev)
...
+				comp_err(dev, "kpb_copy(): too much data to buffer.");
 			}
 
-			/* Update buffered size. NOTE! We only record buffered
-			 * data up to the size of history buffer.
-			 */
-			kpb->hd.buffered += MIN(kpb->hd.buffer_size -
-						kpb->hd.buffered,
-						copy_bytes);
+			comp_update_buffer_produce(sink_c, copy_bytes);
```
pblaszko, line 1121: it is called twice for non-micselect option now

src/audio/kpb.c
```diff
@@ -969,34 +1093,39 @@ static int kpb_copy(struct comp_dev *dev)
...
+		if (kpb->mic_sel_cnt == 0) {
+			kpb_copy_samples(sink_c, source_c, copy_bytes, sample_width, channels);
+			avail_bytes = audio_stream_get_avail_bytes(&source_c->stream);
 
-		kpb_copy_samples(sink_c, source_c, copy_bytes, sample_width, channels);
+			/* Buffer source data internally in history buffer for future
+			 * use by clients.
+			 */
+			if (avail_bytes <= kpb->hd.buffer_size) {
+				ret = kpb_buffer_data(dev, source_c, copy_bytes);
```
pblaszko, line 1104: now the data is buffered in history-buffer only in non-micselect case
I think if/else should only select between kpb_copy_samples() / kpb_micselect_copy() and leave history-buffer part unchanged.

pblaszko (changes requested):

src/audio/kpb.c
```diff
@@ -91,6 +91,10 @@ struct comp_data {
 #ifdef CONFIG_IPC_MAJOR_4
 	struct ipc4_kpb_module_cfg ipc4_cfg;
 #endif /* CONFIG_IPC_MAJOR_4 */
+	uint16_t mic_sel_cnt;
```
andrula-song, line 94: can you add a comment on this or change it to num_of_sel_mic?

src/audio/kpb.c
```diff
@@ -893,6 +897,130 @@ static int kpb_reset(struct comp_dev *dev)
 	return ret;
 }
 
+#ifdef KPB_HIFI3
+static void kpb_micselect_copy(struct comp_dev *dev, struct comp_buffer __sparse_cache *sink,
+			       struct comp_buffer __sparse_cache *source, size_t size,
+			       uint32_t channels)
```
andrula-song, line 903: since this function is do the audio stream copy, there is a function audio_stream_copy in audio_stream_copy.h, we can just call that function to replace this function.

src/audio/kpb.c
```diff
@@ -893,6 +897,130 @@ static int kpb_reset(struct comp_dev *dev)
...
+		ae_int32x2 d32 = AE_ZERO32();
+		const size_t in_offset = kpb->num_of_in_channels * sizeof(ae_int32);
+		const size_t out_offset = kpb->mic_sel_cnt * sizeof(ae_int32);
+		const size_t samples_per_chan = size / (sizeof(uint32_t) * kpb->mic_sel_cnt);
+		ae_int32 *buf = (ae_int32 *)(ostream->addr);
+		ae_int32 *buf_end = (ae_int32 *)(ostream->end_addr);
+		ae_int32 *out_ptr = (ae_int32 *)buf;
+
+		AE_SETCBEGIN0(buf);
+		AE_SETCEND0(buf_end);
```
andrula-song, line 946: can we just put "AE_SETCBEGIN0(ostream->addr); AE_SETCEND0(ostream->end_addr);" out of the if statement? and then the buf and buf_end is unused.

src/audio/kpb.c
andrula-song, line 903: seems the input parameter "channels" is unused.

src/audio/kpb.c
```diff
@@ -893,6 +897,130 @@ static int kpb_reset(struct comp_dev *dev)
...
+				AE_S32_L_XC(d32, out_ptr, out_offset);
+			}
+		}
+	}
+}
+#else
+static void kpb_micselect_copy(struct comp_dev *dev, struct comp_buffer __sparse_cache *sink,
+			       struct comp_buffer __sparse_cache *source, size_t size,
+			       uint32_t in_channels)
+{
```
andrula-song, line 963: why not keep the implementation same as the HiFi version? Then we can reduce the first for loop from in_channels times to  kpb->mic_sel_cnt times and many if statement.

src/audio/kpb.c
lyakh: > Hmm @lyakh array is used in set_large_config ?
> 
> I wonder about the "[4]". Maybe this should be a define and there probably should be some comment why it is 4.

@kv2019i I said "in this commit" - yes, it's used in the second commit. It's ok but just a bit confusing. If possible it's usually good to add data when it's needed, but it isn't critical of course. Within one PR it's still ok.

lyakh: > ,,offsets" is used in kpb_set_micselect function. What size would be appropriate?

I think you could use one of two approached in this and similar cases: (1) tight optimisation for space, you really use types that are enough for values they might need to contain. Then you could use `unsigned char` for offsets, etc. Or (2) you use "natural" types that should be the easiest for the DSP to handle, that are best suitable for integer operations - then I'd just use `int` or `unsigned int` or `size_t` or `uintptr_t` as appropriate. Using `uint16_t` immediately creates an impression, that it's important for this field to have that type - either for compatibility with other types, e.g. if a pointer to it is used as a function argument, or because 8 bits aren't enough and 16 bits are guaranteed to be enough and we are tight on space, none of which seem to hold.

src/include/sof/audio/kpb.h
fkwasowi: For compatibility with IPC4 and tests:
enum KpbModuleConfigParams
{
    //! Configure the buffer size for Keyphrase,onset and buffer to skip sizes.
    KP_BUF_CFG_MODE   = 0,
    //! Configure the module ID's which would be part of the Fast mode tasks
    KP_BUF_CFG_FM_MODULE  = 1,
    //! Parameter to retrieve KP position in buffer
    /*!
      Use LARGE_CONFIG_GET to retrieve KP buffer position in bytes of samples
      for kpb module inout format.

      \see KeyPhraseBufferPosition
    */
    KP_BUF_POS_IN_BUFFER = 2,
    //! Parameter to request Draining On Demand (DOD).
    /*!
      Start draining given pin without KPB.
    */
    KP_BUF_PIN_DOD = 3,
    //! Parameter to retrieve KP position in buffer
    /*!
      Use LARGE_CONFIG_GET to retrieve KP buffer position in bytes of samples
      for kpb module inout format.

      \see KeyPhraseBufferPositionExtended

      Sample code to retrieve KeyPhraseBufferPositionExtended:
      \code
      Message::LargeConfigOp message(true, KPB_MODULE_ID, KPB_INSTANCE_ID);
      message.GetBits().large_param_id = Message::LargeConfigOp::VENDOR_CONFIG_PARAM;
      message.GetBits().init_block = true;
      message.GetBits().final_block = true;
      message.GetBits().data_off_size = sizeof(TL); // type/length struct

      Message::LargeConfigOp::ExtendedParamId param = {0};
      param.part.parameter_type = KP_BUF_POS_IN_BUFFER_EXTENDED;
      param.part.parameter_instance = MODEL_ID; // KPB_CLIENT_ID
      TL input_mailbox(param.full, sizeof(KeyPhraseBufferPositionExtended));

      KeyPhraseBufferPositionExtended* output_mailbox = NULL;
      send_ipc(message, input_mailbox, (uint8_t**)&output_mailbox);

      \endcode
    */
    KP_BUF_POS_IN_BUFFER_EXTENDED = 4,
    KP_BUF_GET_DATA = 7,
    //! Configure the module to enable ULP support
    /*!
      The command enables ULP support / ULP Audio Gateway support in
      History Buffer.
    */
    KP_BUF_ULP_SUPPORT = 8,
    //! Specifies all data consumers attached to History Buffer.
    /*!
        It is also needed when a module wants to request data to be streamed
        via DATA_CONSUMER_REGISTERED and DATA_CONSUMER_UNREGISTERED messages.
    */
    KP_BUF_DATA_CONSUMER_INFO = 9,
    /*! Sent when a reference signal from playback is connected/disconnected to/from
        the capture path
     */
    KP_BUF_REFERENCE_DATA_STATE = 10,
    //! Mic selector for client - sets microphone id for real time sink mic selector
    KP_BUF_CLIENT_MIC_SELECT = 11,
};

src/audio/kpb.c
fkwasowi: Done

src/include/sof/audio/kpb.h
fkwasowi: I have added the KPB_

src/include/sof/audio/kpb.h
fkwasowi: Comment added to code

src/audio/kpb.c
```diff
@@ -897,6 +897,124 @@ static int kpb_reset(struct comp_dev *dev)
 	return ret;
 }
 
+#ifdef KPB_HIFI3
+static void kpb_micselect_copy(struct comp_dev *dev, struct comp_buffer __sparse_cache *sink,
```
singalsu, line 901: It would be nicer to split the HiFiX function versions to separate files like kpb_hifiX.c and kpb_generic.c. The processing modes become too cluttered if with all the versions under #if macros.

.xxproject
```diff
@@ -0,0 +1,2 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<xxProperties/>
```
lyakh, line 2: please remove

src/audio/kpb.c
```diff
@@ -2060,6 +2063,50 @@ static inline void kpb_change_state(struct comp_data *kpb,
...
+				bool first_block,
+				bool last_block,
+				uint32_t data_offset,
+				const char *data)
+{
+	comp_info(dev, "kpb_set_large_config()");
+
+	switch (param_id) {
+	case KP_BUF_CLIENT_MIC_SELECT:
+			return kpb_set_micselect(dev, data, data_offset);
```
lyakh, line 2224: indentation

.xxproject
lgirdwood: @fkwasowi feel free to add a ```.gitignore``` update and include this file type in another PR. This then wont bother you again.

.xxproject
```diff
@@ -0,0 +1,2 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
lgirdwood, line 1: @fkwasowi btw, your editor has done something funny with the commit message.

src/audio/kpb.c
fkwasowi: Done

src/include/sof/audio/kpb.h
```diff
@@ -150,6 +159,17 @@ struct history_data {
 	struct history_buffer *c_hb; /**< current buffer used for writing */
 };
 
+enum ipc4_kpb_module_config_params {
+	    /* Mic selector for client - sets microphone id for real time sink mic selector
+	     * IPC4-compatible ID - please do not change the number */
+	    KP_BUF_CLIENT_MIC_SELECT = 11,
```
lyakh, line 165: indentation: please remove spaces

src/audio/kpb.c
```diff
@@ -897,6 +897,119 @@ static int kpb_reset(struct comp_dev *dev)
...
+
+			for (size_t i = 0; i < samples_per_chan; i++) {
+				if (&out_data[out_samples * kpb->num_of_sel_mic + ch]
+						>= (uint16_t *)ostream->end_addr) {
+					out_data = (uint16_t *)ostream->addr;
+					out_samples = 0;
+				}
+				out_data[out_samples * kpb->num_of_sel_mic + ch] =
+						in_data[i * in_channels + kpb->offsets[ch]];
+				out_samples++;
```
lyakh, line 985: you could optimise this a bit by replacing multiplications with additions:
```
for (size_t i = 0; i < samples_per_chan; i += in_channels) {
	...
	out_data[out_samples + ch] = in_data[i + kpb->offsets[ch]];
	out_samples += kpb->num_of_sel_mic;
```

src/audio/kpb.c
```diff
@@ -897,6 +897,119 @@ static int kpb_reset(struct comp_dev *dev)
...
+
+			for (size_t i = 0; i < samples_per_chan; i++) {
+				if (&out_data[out_samples * kpb->num_of_sel_mic + ch]
+						>= (uint32_t *)ostream->end_addr) {
+					out_data = (uint32_t *)ostream->addr;
+					out_samples = 0;
+				}
+				out_data[out_samples * kpb->num_of_sel_mic + ch] =
+						in_data[i * in_channels + kpb->offsets[ch]];
+				out_samples++;
```
lyakh, line 1006: here as well

src/audio/kpb.c
```diff
@@ -2060,6 +2063,50 @@ static inline void kpb_change_state(struct comp_data *kpb,
...
+static int kpb_set_large_config(struct comp_dev *dev, uint32_t param_id,
+				bool first_block,
+				bool last_block,
+				uint32_t data_offset,
+				const char *data)
+{
+	comp_info(dev, "kpb_set_large_config()");
+
+	switch (param_id) {
+	case KP_BUF_CLIENT_MIC_SELECT:
```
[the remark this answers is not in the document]
singalsu: We don't at the moment support for other than byte control type in IPC4, so e.g. enum is not possible. But please align with @ujfalusi how to utilize the param_id bits space in this set large config IPC so we are later able to add the other control types, and several controls of same type for a component.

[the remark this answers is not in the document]
ujfalusi: @singalsu, @fkwasowi, I think this is not an enum type of control but a binary. It is expecting some mask sent down, no?

So, I think LARGE_CONFIG, param_id and the payload is u32 is fine here.

We are mixing up things with `enum` and **enum**. The param_id has to be unique to target a certain configuration, defining that with enum is fine, could be #define as well, but we like enums.

The enum type or for the matter: multiple, same type controls/parameters for a same module is the issue at the end of the debate.

But if we look a bit further: if a module have more than one parameter of the same type (volume, switch, enum, bytes), how that should be handled? If we need to add another mic select  for this module for example. I guess we will have a new param_id?
Since the binary is already assuming this (we have param_id + payalod), I believe this is going to be the case going forward?
So, each parameter (like multiple volume for a single module) is differentiated based on the param_id?

src/audio/kpb.c
andrula-song: and one more with
for (ch = 0; ch < kpb->num_of_sel_mic; ch++) {
			out_samples = 0;
			in_data = (uint16_t *)istream->r_ptr + kpb->offsets[ch];
			out_data = (uint16_t *)ostream->w_ptr + ch;

			for (size_t i = 0; i < samples_per_chan; i += in_channels) {
				...
				out_data[out_samples * kpb->num_of_sel_mic] = in_data[i];

src/audio/kpb.c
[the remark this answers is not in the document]
singalsu: It seems like value -1 means all channels, while values 0, 1, .. refer to channels. So setting left and right channel to a different gain value would need two set large config messages, right?

lgirdwood:
@pblaszko updates OK ?

pblaszko:
> @pblaszko updates OK ?

Changes go in a good direction, but I know there are still problems with passing basic KPB tests locally, so I cannot approve.
@fkwasowi maybe it is a good idea to change this PR to draft until confirmation that functionality works properly?

src/audio/kpb.c
```diff
@@ -91,6 +91,10 @@ struct comp_data {
 #ifdef CONFIG_IPC_MAJOR_4
 	struct ipc4_kpb_module_cfg ipc4_cfg;
 #endif /* CONFIG_IPC_MAJOR_4 */
+	uint32_t num_of_sel_mic;
```
lgirdwood, line 94: Please dont include the rimage update in this patch, if its needed for this feature then it needs to be a west update.

src/audio/kpb.c
```diff
@@ -897,6 +897,130 @@ static int kpb_reset(struct comp_dev *dev)
 	return ret;
 }
 
+#ifdef KPB_HIFI3
```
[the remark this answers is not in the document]
lgirdwood: @fkwasowi can you reach out to @singalsu on this, we do need a high level Kconfig for Hifi flavor that can be used by all modules.

src/include/sof/math/numbers.h
```diff
@@ -29,7 +29,13 @@
 	__a < 0 ? -1 :		\
 	__a > 0 ? 1 : 0;	\
 })
-
+#ifndef ROUND_DOWN
+#define ROUND_DOWN(size, alignment) ({			\
```
lgirdwood, line 33: @singalsu whats the guidance here now for math macro naming to avoid collisions with Zephyr or C libraries ?

src/audio/kpb.c
lyakh: @fkwasowi oh, sorry, the loop in my change suggestion wasn't correct, I think it should be
```
for (size_t i = 0; i < samples_per_chan * in_channels; i += in_channels) {
```

src/audio/kpb.c
fkwasowi: Fixed

src/audio/kpb.c
fkwasowi: Done

andrula-song: sorry, my second for loop is not right too, it should be 
           for (ch = 0; ch < kpb->num_of_sel_mic; ch++) {
		out_samples = 0;
		in_data = (uint16_t *)istream->r_ptr + kpb->offsets[ch];
		out_data = (uint16_t *)ostream->w_ptr + ch;
		for (size_t i = 0; i < samples_per_chan * in_channels; i += in_channels) {
			...
			out_data[out_samples * kpb->num_of_sel_mic] = in_data[i];

src/audio/kpb.c
```diff
@@ -893,6 +927,165 @@ static int kpb_reset(struct comp_dev *dev)
...
+
+	for (ch = 0; ch < micsel_channels; ch++) {
+		out_samples = 0;
+		in_data = (uint16_t *)istream->r_ptr;
+		out_data = (uint16_t *)ostream->w_ptr;
+
+		for (size_t i = 0; i < samples_per_chan * in_channels; i += in_channels) {
+			if (&out_data[out_samples + ch]
+					>= (uint16_t *)ostream->end_addr) {
+				out_data = (uint16_t *)ostream->addr;
```
andrula-song, line 1019: actually, our the type of our signal is signed, it's better to set out_data &  in_data  as int16_t.

src/audio/kpb.c
```diff
@@ -893,6 +927,165 @@ static int kpb_reset(struct comp_dev *dev)
...
+				 uint32_t in_channels, uint32_t micsel_channels, uint32_t *offsets)
+{
+	struct audio_stream __sparse_cache *istream = &source->stream;
+	struct audio_stream __sparse_cache *ostream = &sink->stream;
+
+	buffer_stream_invalidate(source, size);
+	size_t out_samples;
+	uint16_t ch;
+	const uint32_t *in_data;
+	uint32_t *out_data;
```
andrula-song, line 1039: here too, int32_t

src/audio/kpb.c
```diff
@@ -810,6 +810,36 @@ static int kpb_prepare(struct comp_dev *dev)
...
+	/* Update number of sel_sink channels.
+	 * If OBS is not equal to IBS it means that KPB will work in micselector mode.
+	 */
+	if (kpb->ipc4_cfg.base_cfg.ibs != kpb->ipc4_cfg.base_cfg.obs) {
+		struct list_item *sink_list;
+		const uint32_t byte_align = 1;
+		const uint32_t frame_align_req = 1;
+		uint32_t sink_id;
+
+		if (!list_is_empty(&dev->bsink_list)) {
```
lyakh, line 823: this isn't needed, please, remove. If the list is empty, the loop won't execute, so the effect is exactly the same

src/audio/kpb.c
```diff
@@ -810,6 +810,36 @@ static int kpb_prepare(struct comp_dev *dev)
...
+					container_of(sink_list, struct comp_buffer, source_list);
+				struct comp_buffer __sparse_cache *sink_c = buffer_acquire(sink);
+
+				audio_stream_init_alignment_constants(byte_align, frame_align_req,
+								      &sink_c->stream);
+				sink_id = sink_c->id;
+
+				if (sink_id == 0)
+					sink_c->stream.channels =
+							KPB_MICSELECT_SUPPORT_CHANNELS;
```
lyakh, line 835: then you'll also be able to put this on a single line

src/audio/kpb.c
```diff
@@ -1000,19 +1160,38 @@ static int kpb_copy(struct comp_dev *dev)
...
+			kpb_copy_samples(sink_c, source_c, copy_bytes, sample_width, channels);
+		} else {
+			uint32_t avail = audio_stream_get_avail_bytes(&source_c->stream);
+			uint32_t free = audio_stream_get_free_bytes(&sink_c->stream);
+
+			copy_bytes = MIN(avail, free * channels / kpb->num_of_sel_mic);
+			copy_bytes = ROUND_DOWN(copy_bytes, (sample_width >> 3) * channels);
+			produced_bytes = copy_bytes * kpb->num_of_sel_mic / channels;
+			produced_bytes =
+				ROUND_DOWN(produced_bytes,
```
lyakh, line 1173: just a cosmetic thing - maybe putting at least these two lines on the same one would look a bit better

src/audio/kpb.c
```diff
@@ -991,7 +1150,8 @@ static int kpb_copy(struct comp_dev *dev)
 			break;
 		}
 
-		copy_bytes = audio_stream_get_copy_bytes(&source_c->stream, &sink_c->stream);
+		copy_bytes =
+			audio_stream_get_copy_bytes(&source_c->stream, &sink_c->stream);
```
lyakh, line 1154: I think it looked better on a single line

src/audio/kpb.c
```diff
@@ -1000,19 +1160,38 @@ static int kpb_copy(struct comp_dev *dev)
...
+		} else {
+			uint32_t avail = audio_stream_get_avail_bytes(&source_c->stream);
+			uint32_t free = audio_stream_get_free_bytes(&sink_c->stream);
+
+			copy_bytes = MIN(avail, free * channels / kpb->num_of_sel_mic);
+			copy_bytes = ROUND_DOWN(copy_bytes, (sample_width >> 3) * channels);
+			produced_bytes = copy_bytes * kpb->num_of_sel_mic / channels;
+			produced_bytes =
+				ROUND_DOWN(produced_bytes,
+					   (sample_width >> 3) * kpb->num_of_sel_mic);
```
lyakh, line 1174: I'm wondering if the compiler is smart enough to calculate this only once, maybe we could help it a bit by doing `unsigned int total_bytes_per_sample = (sample_width >> 3) * kpb->num_of_sel_mic;`

lyakh (commented):
nice, I think gcc copying should run faster now with no multiplications. Just a couple of cosmetic suggestions left

src/audio/kpb.c
fkwasowi: Done

src/audio/kpb.c
fkwasowi: fixed

lyakh (approved):

src/audio/kpb.c
```diff
@@ -228,7 +228,7 @@ static void kpb_set_params(struct comp_dev *dev,
 	params->sample_valid_bytes =
 		kpb->ipc4_cfg.base_cfg.audio_fmt.valid_bit_depth / 8;
 	params->buffer_fmt = kpb->ipc4_cfg.base_cfg.audio_fmt.interleaving_style;
-	params->buffer.size = kpb->ipc4_cfg.base_cfg.obs * KPB_MAX_BUFF_TIME * params->channels;
+	params->buffer.size = kpb->ipc4_cfg.base_cfg.ibs * KPB_MAX_BUFF_TIME * params->channels;
```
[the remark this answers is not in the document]
fkwasowi: Fixed

src/audio/kpb.c
```diff
@@ -924,6 +924,165 @@ static int kpb_reset(struct comp_dev *dev)
...
+static void kpb_micselect_copy(struct comp_dev *dev, struct comp_buffer __sparse_cache *sink_c,
+			       struct comp_buffer __sparse_cache *source_c, size_t copy_bytes,
+			       uint32_t channels)
+{
+	struct comp_data *kpb = comp_get_drvdata(dev);
+	size_t sample_width = kpb->config.sampling_width;
+	uint32_t *offsets = kpb->offsets;
+
+	switch (sample_width) {
+#if CONFIG_FORMAT_S16LE
```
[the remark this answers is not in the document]
fkwasowi: fixed

pblaszko (approved):

PR #7041, opened 2023-02-03
ipc: save pipeline attributes

dnikodem:
During create a new pipeline, the FW is not store information about pipeline attributes. The information is lost and we can not verify later if pipeline has ULP or Autonomous_reset attributes.

This PR added a attributes field in pipeline structure and save the information during pipeline create.

### DIFF
--- a/src/ipc/ipc4/helper.c
+++ b/src/ipc/ipc4/helper.c
@@ -147,8 +147,7 @@ struct ipc_comp_dev *ipc_get_comp_by_ppl_id(struct ipc *ipc, uint16_t type, uint
 	return NULL;
 }
 
-static int ipc4_create_pipeline(uint32_t pipeline_id, uint32_t priority,
-				uint32_t memory_size, uint32_t core_id)
+static int ipc4_create_pipeline(struct ipc4_pipeline_create *pipe_desc)
 {
 	struct ipc_comp_dev *ipc_pipe;
 	struct pipeline *pipe;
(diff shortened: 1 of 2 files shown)


src/ipc/ipc4/helper.c
```diff
@@ -147,24 +147,23 @@ struct ipc_comp_dev *ipc_get_comp_by_ppl_id(struct ipc *ipc, uint16_t type, uint
 	return NULL;
 }
 
-static int ipc4_create_pipeline(uint32_t pipeline_id, uint32_t priority,
-				uint32_t memory_size, uint32_t core_id)
+static int ipc4_create_pipeline(struct ipc4_pipeline_create *pipe_desc)
```
lyakh, line 150: looks like it can be `const`?

dnikodem: good catch - I will change it to const in next patch. Thanks

lgirdwood (commented):
LGTM, lets wait for the const update before merging.

pblaszko (commented):
The purpose of this PR is to prepare to use pipeline attributes. I would add in this PR also enum with possible attribute values in IPC protocol (ULP, Autonomous Reset) for future use.

dnikodem:
> The purpose of this PR is to prepare to use pipeline attributes. I would add in this PR also enum with possible attribute values in IPC protocol (ULP, Autonomous Reset) for future use.

Pipeline attributes are defined in base_fw.h for ipc4.

pblaszko:
> > The purpose of this PR is to prepare to use pipeline attributes. I would add in this PR also enum with possible attribute values in IPC protocol (ULP, Autonomous Reset) for future use.
> 
> Pipeline attributes are defined in base_fw.h for ipc4.

OK, I missed that. LGTM then.

pblaszko (approved):

PR #6996, opened 2023-01-27
Add Smart amp feedback pipeline in nocodec topology

ranj063:
This is a replacement for #6799 and depends on the kernel PR https://github.com/thesofproject/linux/pull/4160

PR #7042, opened 2023-02-03
samples: convert smart amp test module to not expect the blob during init with IPC4

ujfalusi:
Hi,

the module initialization for IPC4 does not contain binary (configuration) blob, only base_cfg and optional bae_cfg_ext.
The configuration blob is sent as a large_config message separately.

### DIFF
--- a/src/samples/audio/smart_amp_test.c
+++ b/src/samples/audio/smart_amp_test.c
@@ -32,6 +32,9 @@ DECLARE_TR_CTX(smart_amp_comp_tr, SOF_UUID(smart_amp_comp_uuid),
 	       LOG_LEVEL_INFO);
 
 struct smart_amp_data {
+#if CONFIG_IPC_MAJOR_4
+	struct sof_smart_amp_ipc4_config ipc4_cfg;
+#endif
 	struct sof_smart_amp_config config;
 	struct comp_data_blob_handler *model_handler;
 	void *data_blob;
@@ -55,8 +58,10 @@ static struct comp_dev *smart_amp_new(const struct comp_driver *drv,
 {
 #if CONFIG_IPC_MAJOR_3
 	const struct ipc_config_process *ipc_sa = spec;
-#endif
 	const struct sof_smart_amp_config *cfg;
+#else
+	const struct ipc4_base_module_extended_cfg *base_cfg = spec;
+#endif
 	struct smart_amp_data *sad;
 	struct comp_dev *dev;
 	size_t bs;
@@ -80,12 +85,23 @@ static struct comp_dev *smart_amp_new(const struct comp_driver *drv,
 	k_mutex_init(&sad->lock);
 
 #if CONFIG_IPC_MAJOR_4
-	cfg = spec;
-	bs = sizeof(struct sof_smart_amp_config);
+	if (base_cfg->base_cfg_ext.nb_input_pins != SMART_AMP_NUM_IN_PINS ||
+	    base_cfg->base_cfg_ext.nb_output_pins != SMART_AMP_NUM_OUT_PINS) {
+		comp_err(dev, "smart_amp_new(): Invalid pin configuration");
+		goto sad_fail;
+	}
+
+	/* Copy the base_cfg */
+	memcpy_s(&sad->ipc4_cfg.base, sizeof(sad->ipc4_cfg.base),
+		 &base_cfg->base_cfg, sizeof(base_cfg->base_cfg));
+
+	/* Copy the pin formats */
+	bs = sizeof(sad->ipc4_cfg.input_pins) + sizeof(sad->ipc4_cfg.output_pin);
+	memcpy_s(sad->ipc4_cfg.input_pins, bs,
+		 base_cfg->base_cfg_ext.pin_formats, bs);
 #else
 	cfg = (struct sof_smart_amp_config *)ipc_sa->data;
 	bs = ipc_sa->size;
-#endif
 
 	if ((bs > 0) && (bs < sizeof(struct sof_smart_amp_config))) {
 		comp_err(dev, "smart_amp_new(): failed to apply config");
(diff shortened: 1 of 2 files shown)


kv2019i (approved):

lgirdwood (approved):

lgirdwood:
@wszypelt @lrudyX good to go here ? Dont think this flow is tested in CI yet ?

wszypelt:
@lgirdwood 3 tests on KD was failed, I think it's not good to merge

ujfalusi:
> @lgirdwood 3 tests on KD was failed, I think it's not good to merge

To be fair, most of the PRs are failing in a same way, if I read the logs right, no?

wszypelt:
@ujfalusi a lot of PR has a problem with one test and right now I'm trying to understand why, but here the whole module fails.
so I'm not sure if it's the same problem as others ;/

wszypelt:
@ujfalusi @lgirdwood I'm taking FW and will check these tests on another machine, in a moment I will write how it went.

ujfalusi:
> @ujfalusi @lgirdwood I'm taking FW and will check these tests on another machine, in a moment I will write how it went.

Right, so, with this patch the configuration blob is _not_ expected within MOD_INIT_INSTANCE message, a separate LARGE_CONFIG should be use to send the configuration (param_id = 2) and the blob itself should be the exact same payload as it was with IPC3, So in the abi header you need to change the magic to 0x34464F53, set they type to 2 and send that with Linux.

ujfalusi:
I was not aware that we had any tests for detect_test at all, like we did not have for the smart_amp_test neither.

mwasko:
@iganakov please review proposed changes

wszypelt:
@ujfalusi @lgirdwood 
As for the current tests
Unfortunately, all three KD tests fail on this PR. I checked on the same settings, compared it to FW from main. those tests pass there, I also confirmed that the 16bit16bit test is unstable, so I will exclude it from the scope and rerun other PR. The other two KD tests remain.

mszleszy:
@lgirdwood 
I'll update tests asap but I need to clarify few things.
KdCfgBlobHeader is being removed from init instance message and it should be send in separate LargeConfigSet.
Should I use IPC4_DETECT_TEST_SET_CONFIG param? 
If so the message will contain base config + KdCfgBlobHeader - do we want to keep it this way, change IPC4_DETECT_TEST_SET_CONFIG  requirements or use different kd param ?

softwarecki (changes requested):

iganakov (changes requested):
Looks good to me except unused field `load_memory_size` in `sof_detect_test_config`. Probably, should be removed.

src/samples/audio/smart_amp_test.c
```diff
@@ -80,19 +85,31 @@ static struct comp_dev *smart_amp_new(const struct comp_driver *drv,
 	k_mutex_init(&sad->lock);
 
 #if CONFIG_IPC_MAJOR_4
-	cfg = spec;
-	bs = sizeof(struct sof_smart_amp_config);
+	if (base_cfg->base_cfg_ext.nb_input_pins != 2 ||
+	    base_cfg->base_cfg_ext.nb_output_pins != 1) {
```
aiChaoSONG, line 89: We have the macros defined, use them here?
```
#define SMART_AMP_NUM_IN_PINS		2
#define SMART_AMP_NUM_OUT_PINS		1
```

aiChaoSONG:
The smart amp change in this PR is verified with #6996 and https://github.com/thesofproject/linux/pull/4160, need more time to test the WoV change.

ujfalusi:
I will drop the detect_test change from the PR for now and open a new one for it to get the smart amp feature enabled faster.

ujfalusi:
Changes since v1:
- drop the detect_test patch 
- use the defines in place for the number of pins number
- remove duplicated 'ipc4' from the IPC4 config struct name

ujfalusi:
> @lgirdwood I'll update tests asap but I need to clarify few things. KdCfgBlobHeader is being removed from init instance message and it should be send in separate LargeConfigSet. Should I use IPC4_DETECT_TEST_SET_CONFIG param? If so the message will contain base config + KdCfgBlobHeader - do we want to keep it this way, change IPC4_DETECT_TEST_SET_CONFIG requirements or use different kd param ?

@mszleszy, the detect_test at init will receive the `struct ipc4_base_module_cfg` only. The configuration is going to arrive via `set_large_config` in a form of (an unmodified, IPC3 compatible) `struct sof_detect_test_config` and the param ID for this config_large is `IPC4_DETECT_TEST_SET_CONFIG` (2).

Fwiw, this is how the blob will look like in topology:
```
CONTROLBYTES_PRIV(DETECTOR_priv,
`       bytes "0x53,0x4f,0x46,0x34,0x02,0x00,0x00,0x00,' # magic + param_id
`       0x20,0x00,0x00,0x00,0x00,0x00,0x00,0x00,'  # size + abi version
`       0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,' # unused
`       0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,' # unused 

`       0x20,0x00,0x00,0x00,0x00,0x00,0x00,0x00,' # the blob itself (struct sof_detect_test_config)
`       0x00,0x00,0x00,0x00,0x34,0x08,0x00,0x00,'
`       0x03,0x00,0x18,0x00,0x00,0x00,0x00,0x00,'
`	0xD0,0x07,0x00,0x00,0x00,0x00,0x00,0x00"'
)

```

ujfalusi:
> Looks good to me except unused field `load_memory_size` in `sof_detect_test_config`. Probably, should be removed.

@iganakov, I have intentionally kept the original `struct sof_detect_test_config` intact in order to be able to re-use the same blob from IPC3 in IPC4 with only needing to update the sof_abi_hdr in topology and in user space.
We can not drop members from the struct  from the middle without breaking existing blobs.

ranj063 (approved):

aiChaoSONG (approved):

libinyang (approved):

softwarecki (approved):

PR #6965, opened 2023-01-18
[UPDATE ZEPHYR] zephyr: dai: Check error code before xrun report

softwarecki:
The dma_get_status function returns -EINVAL if it detects an invalid parameter, -EPIPE if it detects an xrun. Added error code check before calling xrun handler.

Updated zephyr revision to 5d902ea621e43ec9a3916eed8dea5ab3647d81c3.

PR #7044, opened 2023-02-06
dai: fix xrun handling

makarukp:
dai copy shuld continue flow if status return xrun warning


softwarecki:
dw dma driver also reports -EPIPE when it detect xrun (https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/dma/dma_dw_common.c#L804). Shouldn't we call dai_report_xrun for this driver? In my opinion, in default case should be a general status read error handling, not xrun handling.

softwarecki:
Similar PR: https://github.com/thesofproject/sof/pull/6965

kv2019i (changes requested):
Some comments inline, added @juimonen to review.

kv2019i:
@makarup @softwarecki I think I now understand why we have native dai-zephyr.c. 

If I run test to a non-connected HDMI PCM node, with non-native HD-DMA driver I get:
```
[   47.628023] <wrn> dai_comp: comp:1 0x40001 dai_copy(): nothing to copy
[   47.629000] <err> hda_dma: hda_dma_link_check_xrun(): underrun detected
[   47.729056] <err> hda_dma: hda_dma_link_check_xrun(): underrun detected
[   48.648050] <inf> ll_schedule: ll timer avg 2251, max 5396, overruns 0
```

With native Zephyr driver I get:
```
[    4.125471] <wrn> dai_comp: comp:1 0x40001 dai_copy(): nothing to copy
[    4.225553] <err> dai_comp: comp:1 0x40001 dai_report_xrun(): underrun due to no data available
[    4.225565] <err> component: comp:1 0x40001 comp_underrun(): dev->comp.id = 262145, source->avail = 384, copy_bytes = 0
[    4.225573] <inf> pipe: pipe:1 0x0 pipe trigger cmd 6
[    4.225588] <err> pipe: pipe:1 0x0 pipeline_copy(): ret = -32, start->comp.id = 262145, dir = 1
```

So in both cases the xrun is observed (makes sense as display side is not connected in this case), but in non-native case the pipeline is not stopped. Took a while but then a realized hda_dma_link_check_xrun() in the old src/drivers/intel/hda/hda-dma.c
 always returns 0 for success, even if xrun is found.

I traced this back to this old commit from 2020:
```
commit 293dfe2469cb2f29895847cb3dd2bc831e3ea7c3
Author: Tomasz Lauda <tomasz.lauda@linux.intel.com>
Date:   Tue Mar 10 10:09:30 2020 +0100

    hda-dma: refactor xrun handling
    
    Changes behaviour of HDA Link overruns and underruns handling.
    Let's no longer stop the stream, but just report an error.
    It might happen that just after the release buffer is still
    full/not yet empty after the previous run, but it shouldn't
    affect the data, since stream has been paused anyway.
```

The Linux stack is still depending on this behaviour (the Linux audio servers open all the PCM nodes and for the HDMI nodes, the expectation is that PCM streaming can be started without an error even if no display is connected at the time).

So I think we need to modify xrun handling so that for the HDA link, we just log a warning, but do not stop the pipeline. Other xrun handling should be as before.

makarukp:
@kv2019i @softwarecki 
I have similar conclusion, however, now I am not 100% sure if dw_dma_driver xrun meaning is the same as hda dma DGCS_BUR and DGCS_BOR. Is it really xrun or error ???
'
#if CONFIG_DMA_DW_HW_LLI
	if (!(dw_read(dev_cfg->base, DW_DMA_CHAN_EN) & DW_CHAN(channel))) {
		LOG_ERR("xrun detected");
		return -EPIPE;
	}
#endif
'

I see that in legacy it was only warning for hda but error for dw_dma. Maybe in cases different than standard xrun resulting from delay it should be reported as an error.

What do you think ?

src/audio/dai-zephyr.c
```diff
@@ -1224,12 +1224,26 @@ static int dai_copy(struct comp_dev *dev)
...
+	case -EPIPE:
+		if (dev->direction == SOF_IPC_STREAM_PLAYBACK)
+			comp_warn(dev, "dai_copy(): dma_get_status() underrun occurred, ret = %u",
+				  ret);
+		else
+			comp_warn(dev, "dai_copy(): dma_get_status() overrun occurred, ret = %u",
+				  ret);
+		break;
+	default:
 		dai_report_xrun(dev, 0);
```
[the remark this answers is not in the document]
lgirdwood: @kv2019i IIRC calling ```dai_report_xrun(dev, 0);``` would cause the XRUN handler on SOF Linux driver to run (and stop/restart the stream) ? Should this not be something to enable/disable (default) at runtime ?

kv2019i:
@makarukp Ack, I can confirm that. With non-native DW-DMA driver, xruns are reported and will stop the pipe. So this special handling is only for the HD-DMA link DMA. Now for Linux, we do need to keep the special handling, and the move of drivers to Zephyr does pose a new challenge how to maintain. Hmm, maybe we put the xrun-raises-an-error logic behind a Kconfig option and otherwise just emit a warning to FW log...?

src/audio/dai-zephyr.c
```diff
@@ -1227,9 +1227,23 @@ static int dai_copy(struct comp_dev *dev)
 	/* get data sizes from DMA */
 	ret = dma_get_status(dd->chan->dma->z_dev, dd->chan->index, &stat);
 	if (ret < 0) {
-		dai_report_xrun(dev, 0);
+		/* DMA status can return -EPIPE and current status content if xrun occurs */
+		if (ret == -EPIPE)
+#if XRUN_RAISES_AN_ERROR
+			dai_report_xrun(dev, 0);
```
lgirdwood, line 1233: and this will eventually call the IPC3/4 version of the notification in subsequent PRs.

src/audio/dai-zephyr.c
```diff
@@ -1227,9 +1227,23 @@ static int dai_copy(struct comp_dev *dev)
 	/* get data sizes from DMA */
 	ret = dma_get_status(dd->chan->dma->z_dev, dd->chan->index, &stat);
 	if (ret < 0) {
-		dai_report_xrun(dev, 0);
+		/* DMA status can return -EPIPE and current status content if xrun occurs */
+		if (ret == -EPIPE)
+#if XRUN_RAISES_AN_ERROR
```
lgirdwood, line 1232: Needs the CONFIG_ prefix

src/audio/Kconfig
```diff
@@ -72,6 +72,13 @@ config COMP_SRC
 	help
 	  Select for SRC component
 
+config XRUN_RAISES_AN_ERROR
+	bool "XRUN is handled and raises an error"
```
lgirdwood, line 76: Best to call this NOTIFY_HOST_ON_XRUN with "Notify the host via IPC that a XRUN/Glitch has occurred"

lgirdwood (commented):
Just some minor comments

src/audio/dai-zephyr.c
lgirdwood: no action needed here, btw, just making an observation.

softwarecki: Some { } would be nice

kv2019i: Hmm. @makarukp @softwarecki @lgirdwood I wonder if we should start by just making the xrun handling here emit warnings and not call dai_report_xrun() yet. That would allow to unblock Zephyr upstream updated.

The CONFIG_NOTIFY_HOST_ON_XRUN is very misleading as no notifies are sent to host in current code. What dai_report_xrun() is to escalate the problem to pipeline level, which will try to restart the pipeline. There is also another layer of conditional logic after dai_report_xrun(). E.g. "NO_XRUN_RECOVERY" option in src/audio/pipeline/pipeline-xrun.c at the pipeline level.

So seems more time is needed to sort out how we introduce the IPC4 notification mechanisms (which are new) and how we tackle the problem that HDA link xruns happen in common use-cases and in the past we have ignored them.

We still have xrun logic at the pipeline level driven by buffering status, so it's not like we are disabling xrun handling completely.

kv2019i (changes requested):
I propose we simplify even further and just emit warnings and do not call dai_xrun_report() yet. There seems to be more opens (see inline).

lgirdwood:
> I propose we simplify even further and just emit warnings and do not call dai_xrun_report() yet. There seems to be more opens (see inline).

Fine by me.

src/audio/dai-zephyr.c
```diff
@@ -1227,9 +1227,23 @@ static int dai_copy(struct comp_dev *dev)
...
+#if CONFIG_NOTIFY_HOST_ON_XRUN
+			dai_report_xrun(dev, 0);
 		return ret;
+#else
+			if (dev->direction == SOF_IPC_STREAM_PLAYBACK)
+				comp_warn(dev, "dai_copy(): dma_get_status() underrun occurred, "
+					  "ret = %u", ret);
+			else
+				comp_warn(dev, "dai_copy(): dma_get_status() overrun occurred, "
+					  "ret = %u", ret);
```
lyakh, line 1241: I think you need braces here. GCC used to complain about constructs like
```
if (a)
	if (b)
		do_x();
	else
		do_y();
```
because of the potential ambiguity of the `else` (this ain't no python). This probably wasn't compiled with `CONFIG_NOTIFY_HOST_ON_XRUN` disabled

src/audio/Kconfig
```diff
@@ -72,6 +72,13 @@ config COMP_SRC
 	help
 	  Select for SRC component
 
+config NOTIFY_HOST_ON_XRUN
+	bool "Notify the host via IPC that a XRUN/Glitch has occurred"
+	default n
+	help
+	  Select for handling DMA xruns and reporting them as errors.
+	  Otherwise, only warnings are emitted.
+
```
kv2019i, line 81: I won't block on this, but if we have the Kconfig, I'd do:
- put it at the end of audio/Kconfig after WRAP_ACTUAL_POSITION (now it's in middle of SRC related options)
- name it DAI_REPORT_XRUNS
- description: "Report DAI overrun/underruns to pipeline"

This would better reflect what this does.

kv2019i (approved):
Ok with the fix now, but if you are refreshing, a few style comments to the Kconfig option to make it more understandable.

makarukp:
> I propose we simplify even further and just emit warnings and do not call dai_xrun_report() yet. There seems to be more opens (see inline).

Does it mean i could remove Kconfig dependency  and simplify it to below example ?
`
switch (ret) {
case 0:
	break;
case -EPIPE:
        tr_warn()
default:
        return ret
`

kv2019i (commented):
@makarukp You have unrelated rimage changes in the updated commit.

Otherwise looks good to go, thanks!

softwarecki (approved):

dnikodem (approved):

lgirdwood (approved):

lgirdwood:
@makarukp can you check the CI. Thanks !

tmleman (approved):

src/audio/dai-zephyr.c
```diff
@@ -1226,10 +1226,21 @@ static int dai_copy(struct comp_dev *dev)
...
+	case 0:
+		break;
+	case -EPIPE:
+		/* DMA status can return -EPIPE and current status content if xrun occurs */
+		if (dev->direction == SOF_IPC_STREAM_PLAYBACK)
+			comp_warn(dev, "dai_copy(): dma_get_status() underrun occurred, ret = %u",
+				  ret);
+		else
+			comp_warn(dev, "dai_copy(): dma_get_status() overrun occurred, ret = %u",
+				  ret);
```
lyakh, line 1239: could you confirm that it's your intent to return the `-EPIPE` error here? If so you need a `COMPILER_FALLTHROUGH` here otherwise you need a `break`

lyakh (dismissed):

src/audio/dai-zephyr.c
makarukp: Good remark! I missed `break`...

lgirdwood (approved):

lgirdwood:
@makarukp can you check CI, thanks !

kv2019i:
@wszypelt This one seems to have hit unexpected errors as well.

src/audio/dai-zephyr.c
```diff
@@ -1226,10 +1226,23 @@ static int dai_copy(struct comp_dev *dev)
...
+		if (dev->direction == SOF_IPC_STREAM_PLAYBACK)
+			comp_dbg(dev, "dai_copy(): dma_get_status() underrun occurred, ret = %u",
+				  ret);
+		else
+			comp_dbg(dev, "dai_copy(): dma_get_status() overrun occurred, ret = %u",
+				  ret);
+		break;
+	default:
 		return ret;
+		break;
```
lyakh, line 1243: please remove the `break`

lyakh (changes requested):

lyakh (approved):

kv2019i:
@wszypelt @makarukp This time it seems an unrelated fail running "git clone" , can you check and agree how to best handle (repush or rerun of CI).

kv2019i:
Now merged! Thank you @wszypelt @makarukp @lyakh for helping to get this across the finish line.

PR #7047, opened 2023-02-07
[Test Only] Dummy wov 4 mtl hda

libinyang:
Don't need to review and Don't merge. Only for test.


lgirdwood (commented):
@libinyang we need a test topology for WoV upstream so it can be used as an example and also used by CI.

marc-hb:
> Don't need to review and Don't merge. Only for test.

Github drafts are meant for that: they don't spam reviewers.

PR #7052, opened 2023-02-07
module-adapter: allow different statuses

lyakh:
The following problem is observed when testing pause functionality:

in a case of a playback stream host -> mixin -> mixout -> DAI where host and mixin belong to the pipeline A and mixout and DAI belong to pipeline B, pipeline B is scheduled before A. So at start we get: scheduler start
B: no data
A: read 1 period of data
scheduler sleep
(1 period of data pending in pipeline A)
scheduler start
B: send 1 period of data out
A: read 1 period of data
scheduler sleep
(1 period of data pending in pipeline A)
PAUSE
RELEASE
scheduler start
B: old data before sleep isn't sent because A is PAUSED A: read 1 period of data
scheduler sleep
(2 periods of data pending in pipeline A)
...

With repeated pause-release cycles this leads to buffer overrun. To fix this allow sending out data by pipeline B when A is still paused.

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -622,12 +622,6 @@ module_single_sink_setup(struct comp_dev *dev,
 	int i = 0;
 
 	list_for_item(blist, &dev->bsource_list) {
-		/* check if the source dev is in the same state as the dev */
-		if (source_c[i]->source->state != dev->state) {
-			i++;
-			continue;
-		}
-
 		comp_get_copy_limits_frame_aligned(source_c[i], sinks_c[0], &c);
 
 		if (!mod->skip_src_buffer_invalidate)
@@ -665,12 +659,6 @@ module_single_source_setup(struct comp_dev *dev,
 	int i = 0;
 
 	list_for_item(blist, &dev->bsink_list) {
-		/* check if the sink dev is in the same state as the dev */
-		if (sinks_c[i]->sink->state != dev->state) {
-			i++;
-			continue;
-		}
-
 		comp_get_copy_limits_frame_aligned(source_c[0], sinks_c[i], &c);
 
 		min_frames = MIN(min_frames, c.frames);


lgirdwood (commented):
@ranj063 fyi

src/audio/module_adapter/module_adapter.c
```diff
@@ -622,12 +622,6 @@ module_single_sink_setup(struct comp_dev *dev,
 	int i = 0;
 
 	list_for_item(blist, &dev->bsource_list) {
-		/* check if the source dev is in the same state as the dev */
```
ranj063, line 625: @lyakh this is a special case for the mixin/mixout alone right or is it widely applicable for any connected sink pipeline because of the order in which trigger pipelines with IPC4?

lyakh: @ranj063 I'm not sure. Definitely it's only possibly with module-adapter components. In our "popular" topologies I only see mixin-out, copier and volume being on the border between two pipelines. Copier isn't using module-adapter, but we could check some volume pipelines. I so far haven't figured out where in volume consumed and produced bytes are updated, will try to figure that out tomorrow. Then I'll add some debugging there and check.

lyakh:
The only CI failure is a timed-out suspend-resume test https://sof-ci.01.org/sofpr/PR7052/build3776/devicetest/index.html?model=TGLU_RVP_NOCODEC_IPC4ZPH&testcase=check-suspend-resume-5

src/audio/module_adapter/module_adapter.c
lyakh: @ranj063 I tested with a capture pipeline like DMIC -> gain.A -> copier.A -> gain.B -> PCM and I don't see any difference with or without this PR so at least it isn't breaking that use-case either

lyakh: @ranj063 I know why there was no difference. The difference is that although in one case `module_process()` is called from `module_adapter_simple_copy()` with `num_input_buffers` of `num_output_buffers` equal to 0, both those parameters are ignored by `volume_process()`... If I add there something like
```
	if (!num_output_buffers || !num_input_buffers)
		return 0;
```
then of course the behaviour changes. But still both possibilities seem to work. Maybe it's because with capture usually you don't have to send data at a constant rate. And also the effect of this PR seems to depend strongly on individual processing modules - how they use those parameters.

ranj063:
> The only CI failure is a timed-out suspend-resume test https://sof-ci.01.org/sofpr/PR7052/build3776/devicetest/index.html?model=TGLU_RVP_NOCODEC_IPC4ZPH&testcase=check-suspend-resume-5

@lyakh the internal CI tests have failed too

lyakh:
> > The only CI failure is a timed-out suspend-resume test https://sof-ci.01.org/sofpr/PR7052/build3776/devicetest/index.html?model=TGLU_RVP_NOCODEC_IPC4ZPH&testcase=check-suspend-resume-5
> 
> @lyakh the internal CI tests have failed too

@ranj063 I cannot open the results, does it work for you?

src/audio/module_adapter/module_adapter.c
kv2019i: Ack, I don't think this can work if the adapter makes this call. I think it should be component that handles this logic, and indeed, if it has data in its buffers, it is ACTIVE and there's room in outputs, processing should happen.

kv2019i (approved):
I think this is the correct thing to do.

lgirdwood (approved):
ok, so this now aligns the buffer period state at start and after pause.

lgirdwood:
CI fully passing. @ranj063 will merge, we can deal with speed bumps after we see daily test.

PR #7055, opened 2023-02-07
topology2: Use input/output pin consistently

ranj063:
Use the input/putput pin terminology consistently to align with the kernel driver and avoid any confusion.


tools/topology/topology2/cavs-nocodec.conf
```diff
@@ -36,7 +36,6 @@
 <dai.conf>
```
kv2019i, line 36: "putput pin" ? :)

kv2019i (approved):
Looks good. One typo in commit, but that's it.

I did a double-check of all usages of "pin" in FW code and basicly 99% of usage is with "input pin" and "output pin". There is one usage of of "sink->id" in copier code, but even there the "sink->id" refers to copier's output, so this is not referring to "sink_pin".

So this seems to be a welcome alignment of terms.

tools/topology/topology2/cavs-nocodec.conf
ranj063: fixed

tools/topology/topology2/include/common/output_audio_format.conf
```diff
@@ -1,7 +1,7 @@
-# Class definition for spirce pin audio format object
```
plbossart, line 1: I like the spirce pin. That sounds like a good concept that's not overloaded just yet.

plbossart (approved):

aiChaoSONG (approved):

PR #4466, opened 2021-07-08
pm_runtime: longer hp sram power gating delay

slawblauciak:
Signed-off-by: Slawomir Blauciak <slawomir.blauciak@linux.intel.com>

PR #7056, opened 2023-02-07
cavs: increase exit delay on memory bank powerup

cujomalainey:
For some reason the memory banks can still be inaccessible despite registers saying they are ready. If they are accessed prematurely this will result in a crash, typically when memory is being initialized. Adding this delay allows a device to pass over 1k reboots without a crash that previously could pass 5 cycles.

### DIFF
--- a/src/platform/intel/cavs/lib/pm_runtime.c
+++ b/src/platform/intel/cavs/lib/pm_runtime.c
@@ -367,6 +367,7 @@ static inline void cavs_pm_runtime_core_en_memory(uint32_t index)
 
 	cavs_pm_memory_hp_sram_power_gate(core_memory_ptr, SOF_CORE_S_SIZE,
 					  true);
+	idelay(MEMORY_POWER_CHANGE_DELAY);
 
 #endif
 }


cujomalainey:
Note this was tested on rpl-001 but the diff between RPL-001 and main appears to be unrelated to the code flow for this issue.

cujomalainey:
Note this regression is likely caused by the fact that https://github.com/thesofproject/sof/pull/4466 (also on adl-004-drop-stable) managed to completely skip landing on main.

We can pick that up instead if preferred, but I figured an exit delay is more deterministic since we know the registers signal availability before the bank is actually ready.

andyross:
Drive by zephyrization comment: AFAICT the equivalent driver[1] in a zephyr build does seem to have (what I think is!) the equivalent delay: https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/xtensa/intel_adsp/cavs/sram.c#L44

Might be a good idea to audit to be sure.  Are there known rules we could enforce more simply?  Maybe something like "Must wait 3us after a write to the LDOCTL register"?  Might be easier than sprinkling delay calls everywhere, and PM isn't likely to be a microsecond-scale performance path.

[1] For the rather different Zephyr PM framework, but the low level code looks more or less directly translated.

cujomalainey:
> Drive by zephyrization comment: AFAICT the equivalent driver[1] in a zephyr build does seem to have (what I think is!) the equivalent delay: https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/xtensa/intel_adsp/cavs/sram.c#L44
> 
> Might be a good idea to audit to be sure. Are there known rules we could enforce more simply? Maybe something like "Must wait 3us after a write to the LDOCTL register"? Might be easier than sprinkling delay calls everywhere, and PM isn't likely to be a microsecond-scale performance path.
> 
> [1] For the rather different Zephyr PM framework, but the low level code looks more or less directly translated.

Actually I think its missing unfortunately. This new delay is the last step so it would be after this [line](https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/xtensa/intel_adsp/cavs/sram.c#L101)

mwasko:
> Note this regression is likely caused by the fact that #4466 (also on adl-004-drop-stable) managed to completely skip landing on main.

It looks like the PR was left hanging and eventually closed when @slawblauciak left the SOF team. 

> 
> We can pick that up instead if preferred, but I figured an exit delay is more deterministic since we know the registers signal availability before the bank is actually ready.

@pblaszko did some research and it looks like this behavior might be related to a known issue. If we confirm that then the recommended fix will be to read EBB status twice and if different values is received then read again.

pblaszko:
> @pblaszko did some research and it looks like this behavior might be related to a known issue. If we confirm that then the recommended fix will be to read EBB status twice and if different values is received then read again.

Some more details about known HW issue:

Issue is root caused to a known limitation in the design. FW WA will be the solution and current WA is robust enough to enable PRQ.

**Title:**
Audio DSP: Value of HSPGISTS0 toggles after clearing bit in HSPGCTL0

**Issue Details:**
FW is sampling wrong SRAM status and accessing uninitialized L2 SRAMs for operation.

**Simulation Observations:**
SRAMPGE controller is taking longer time to reflect input PGE change to output staggered PGE to Memory EBBs and Memory Fabric
Memory fabric is directly qualifying PGE bits to PGE status to enable Immediate EBB OFF status but resulting in one pulse false status for EBB power ON case

**Root Cause:**  
Known Limitation of SRAMPGE logic that whenever PGE bits are toggled, staggering counter always start from 0-bit and traverse through whole PGE array cycle by cycle based on SPGS delay. SPGS default value is 1-XOSC clock cycle. In FPGA image EBB-29 is powered ON. It took 58 xosc clock cycles from PGE input change to the staggered output change. 58-clock cycles are 1.5us that enables FW read polling on EBB status and memory fabric glitch is sampled.

**Fix Proposal:**
cAVS1.8/2.x products (CNP/ICP/TGP/LKF/MCC/ADP-P/SUECreek/WHL/CML/RKL): FW WA  -- Read the EBB status twice and if different values is received then read again. HPDSP Cache is still using old srampge logic that need to be analyzed for possible staggering issues.
ACE (JLF/CVF) – L2SRAM staggering uArch is redefined to fix the multiple past issue around SRAMPGE block. HPDSP Cache is still using old srampge logic that need to be analyzed for possible staggering issues. Memory fabric glitch issue is fixed through new uArch definition.
ACE (MTP and future) -- L2SRAM staggering uArch is redefined to fix the multiple past issue around SRAMPGE block. HPDSP Caches need to implement new uArch proposal to avoid any possible failure for the program. Memory fabric glitch issue is fixed through new uArch definition.

cujomalainey:
> > We can pick that up instead if preferred, but I figured an exit delay is more deterministic since we know the registers signal availability before the bank is actually ready.
> 
> @pblaszko did some research and it looks like this behavior might be related to a known issue. If we confirm that then the recommended fix will be to read EBB status twice and if different values is received then read again.

Please make this a priority, we currently have devices in the hands of dogfooders that are unstable and are running out of time on our release calendar

mbukowsk:
I have already created PR with recommended change
#7065

PR #7057, opened 2023-02-07
topology2: speaker-echo-ref: fix hardcoded dai_index

yongzhi1:
Do not hardcode dai_index to 1.

### DIFF
--- a/tools/topology/topology2/platform/intel/speaker-echo-ref.conf
+++ b/tools/topology/topology2/platform/intel/speaker-echo-ref.conf
@@ -15,7 +15,7 @@ Object.Pipeline {
 		index	$ECHO_REF_DAI_PIPELINE_ID
 
 		Object.Widget.copier."1" {
-			dai_index	1
+			dai_index	$SPEAKER_SSP_DAI_INDEX
 			dai_type	"SSP"
 			type		dai_out
 			copier_type	"SSP"


tools/topology/topology2/platform/intel/speaker-echo-ref.conf
```diff
@@ -15,7 +15,7 @@ Object.Pipeline {
 		index	$ECHO_REF_DAI_PIPELINE_ID
 
 		Object.Widget.copier."1" {
-			dai_index	1
+			dai_index	$SPEAKER_SSP_DAI_INDEX
 			dai_type	"SSP"
 			type		dai_out
```
plbossart, line 20: @ranj063 out of curiosity is there a reason why we never use dai_in? It's implied or the default?

plbossart (approved):

kv2019i (approved):

lgirdwood (approved):

tools/topology/topology2/platform/intel/speaker-echo-ref.conf
lgirdwood: possibly dai_in enumerates to 0 and becomes default. @ranj063 do we need a followup for the copier base to make this a mandatory setting ?

PR #7060, opened 2023-02-08
topology2: ssp: add link clock aux data for mtl

juimonen:
Add conditional link clock settings for mtl ssp using auxiliary data in the blob generation.

### DIFF
--- a/tools/topology/topology2/cavs-nocodec-bt.conf
+++ b/tools/topology/topology2/cavs-nocodec-bt.conf
@@ -2,6 +2,7 @@
 <searchdir:include/common>
 <searchdir:include/components>
 <searchdir:include/dais>
+<searchdir:include/dais/intel>
 <searchdir:include/pipelines/cavs>
 <searchdir:platform/intel>
 
@@ -29,9 +30,11 @@
 <dai.conf>
 <host.conf>
 <dmic-default.conf>
+<ssp-default.conf>
 
 <bt-default.conf>
 <bt-generic.conf>
+<intel_ssp_aux_config.conf>
 
 Define {
 	MCLK 				24576000
@@ -78,12 +81,21 @@ Object.Dai.SSP [
 		sample_bits		32
 		quirks			"lbm_mode"
 		io_clk		$MCLK
+		version         $SSP_BLOB_VERSION
+
 		Object.Base.hw_config.1 {
 			name	"SSP0"
 			id      0
 			mclk_freq       $MCLK
 			bclk_freq       3072000
 			tdm_slot_width  32
+			IncludeByKey.PLATFORM {
+				"mtl" {
+					Object.Base.link_config.1 {
+						clock_source	1
+					}
+				}
+			}
 		}
 	}
 	{
@@ -95,6 +107,7 @@ Object.Dai.SSP [
 		sample_bits		32
 		quirks			"lbm_mode"
 		io_clk		$MCLK
+		version         $SSP_BLOB_VERSION
 
 		Object.Base.hw_config.1 {
 			name	"SSP1"
@@ -102,6 +115,13 @@ Object.Dai.SSP [
 			mclk_freq	$MCLK
 			bclk_freq	3072000
 			tdm_slot_width	32
+			IncludeByKey.PLATFORM {
+				"mtl" {
+					Object.Base.link_config.1 {
+						clock_source	1
+					}
+				}
+			}
 		}
 	}
 ]
--- a/tools/topology/topology2/cavs-nocodec.conf
+++ b/tools/topology/topology2/cavs-nocodec.conf
@@ -2,6 +2,7 @@
 <searchdir:include/common>
 <searchdir:include/components>
 <searchdir:include/dais>
+<searchdir:include/dais/intel>
 <searchdir:include/pipelines/cavs>
 <searchdir:platform/intel>
 
@@ -42,6 +43,9 @@
 <input_audio_format.conf>
 <output_audio_format.conf>
 <dmic-default.conf>
+<ssp-default.conf>
+<intel_ssp_aux_config.conf>
+
 
 Define {
 	MCLK 				24576000
@@ -94,6 +98,7 @@ Object.Dai.SSP [
 		sample_bits		32
 		quirks			"lbm_mode"
 		io_clk		$MCLK
+		version         $SSP_BLOB_VERSION
 
 		Object.Base.hw_config.1 {
 			name	"SSP0"
@@ -101,6 +106,13 @@ Object.Dai.SSP [
 			mclk_freq	$MCLK
 			bclk_freq	3072000
 			tdm_slot_width	32
+			IncludeByKey.PLATFORM {
+				"mtl" {
+					Object.Base.link_config.1 {
+						clock_source	1
+					}
+				}
+			}
 		}
 	}
 	{
@@ -112,6 +124,7 @@ Object.Dai.SSP [
 		sample_bits		32
 		quirks			"lbm_mode"
 		io_clk		$MCLK
+		version         $SSP_BLOB_VERSION
 
 		Object.Base.hw_config.1 {
 			name	"SSP1"
@@ -119,6 +132,13 @@ Object.Dai.SSP [
 			mclk_freq	$MCLK
 			bclk_freq	3072000
 			tdm_slot_width	32
+			IncludeByKey.PLATFORM {
+				"mtl" {
+					Object.Base.link_config.1 {
+						clock_source	1
+					}
+				}
+			}
 		}
 	}
 	{
@@ -130,6 +150,7 @@ Object.Dai.SSP [
 		sample_bits		32
 		quirks			"lbm_mode"
 		io_clk		$MCLK
+		version         $SSP_BLOB_VERSION
 
 		Object.Base.hw_config.1 {
 			name	"SSP2"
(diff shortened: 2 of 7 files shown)


tools/topology/topology2/platform/intel/ssp-default.conf
```diff
@@ -0,0 +1,4 @@
+# Default SSP variable definitions
+Define {
+       SSP_BLOB_VERSION		0x100
+}
```
lgirdwood, line 4: missing new line

lgirdwood (commented):
@ranj063 @jsarha pls review

tools/topology/topology2/cavs-nocodec-bt.conf
```diff
@@ -77,12 +80,21 @@ Object.Dai {
...
 		Object.Base.hw_config.1 {
 			name	"SSP0"
 			id      0
 			mclk_freq       $MCLK
 			bclk_freq       3072000
 			tdm_slot_width  32
+			IncludeByKey.PLATFORM {
+				"mtl" {
+					Object.Base.link_config.1 {
+						clock_source	2
```
plbossart, line 94: can we not have some sort of enum that is self explanatory or a comment. This is very hard to review.

tools/topology/topology2/cavs-nocodec-bt.conf
```diff
@@ -77,12 +80,21 @@ Object.Dai {
 		sample_bits		32
 		quirks			"lbm_mode"
 		io_clk		$MCLK
+		version         $SSP_BLOB_VERSION
```
plbossart, line 83: can this blob version be added automagically since we know which platform is used?

plbossart (commented):
seems a bit intrusive an repetitive @juimonen, can we not have a mtl default with the values added instead of changing all topologies.

juimonen:
> seems a bit intrusive an repetitive @juimonen, can we not have a mtl default with the values added instead of changing all topologies.

yes agreed. @ranj063 any idea how to add this only once based on platform?

tools/topology/topology2/cavs-nocodec-bt.conf
ranj063: I'd not pollute the hw_config class with NHLT-specific stuff at all. The clock_source is a SSP-specific value that can be set in ssp.conf based on PLATFORM definition

ranj063: what i'd think you need is something like this: 
```
diff --git a/tools/topology/topology2/include/dais/ssp.conf b/tools/topology/topology2/include/dais/ssp.conf
index 9c73e7562..99fc9024e 100644
--- a/tools/topology/topology2/include/dais/ssp.conf
+++ b/tools/topology/topology2/include/dais/ssp.conf
@@ -111,6 +111,25 @@ Class.Dai."SSP" {
        # platform clock frequency
        DefineAttribute.io_clk {}
 
+       DefineAttribute."clock_source" {
+               constraints {
+                       !valid_values [
+                               "XTAL"
+                               "CARDINAL"
+                               "PLL"
+                               "MCLK"
+                               "WOV_CRO"
+                       ]
+                       !tuple_values [
+                               0
+                               1
+                               2
+                               3
+                               4
+                       ]
+               }
+       }
+
        attributes {
                !constructor [
                        "name"
@@ -140,4 +159,10 @@ Class.Dai."SSP" {
        frame_pulse_width       0
        tdm_padding_per_slot    false
        version                 $SSP_BLOB_VERSION
+
+       IncludeByKey.PLATFORM {
+               "mtl"   {
+                       clock_source    PLL
+               }
+       }
 }
```

plbossart: > +       IncludeByKey.PLATFORM {
> +               "mtl"   {
> +                       clock_source    PLL

I hope we are using the 24.576MHz cardinal clock that comes for free and not the PLL? That index 2 seems completely wrong...

ranj063: @plbossart I couldnt be wrong with the indices, I was just giving an example. @juimonen please correct me

lgirdwood:
> > seems a bit intrusive an repetitive @juimonen, can we not have a mtl default with the values added instead of changing all topologies.
> 
> yes agreed. @ranj063 any idea how to add this only once based on platform?

Dont we have a platform file that can be included like in topology1 ? if not, please add one.

ranj063:
> Dont we have a platform file that can be included like in topology1 ? if not, please add one

@lgirdwood yes we do. There's mtl.conf that we can add this to

kv2019i:
@juimonen  array-support merged in #7113 so rebase will be needed

lgirdwood:
@juimonen still RFC ?

juimonen:
> @juimonen still RFC ?

I want to see the CI after rebase (zephyr update) and there's still topology "cleanup" issues to be discussed...

juimonen:
@ranj063 @plbossart we need to figure this out, related code is merged and used in alsa-utils and zephyr ssp driver. This is a working solution, but if you want this cleaner, I need some guidance how to proceed. I would really like to get this for 2.5...

tools/topology/topology2/cavs-nocodec.conf
```diff
@@ -126,13 +146,21 @@ Object.Dai {
...
 			mclk_freq	$MCLK
 			bclk_freq	3072000
 			tdm_slot_width	32
+			IncludeByKey.PLATFORM {
+				"mtl" {
+					Object.Base.link_config.1 {
+						clock_source	2
+					}
+				}
+			}
```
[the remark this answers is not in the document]
lgirdwood: @juimonen can we take this from platform.conf i.e. using the same method as ```SSP_BLOB_VERSION``` above ? This could be an attribute in teh base class that would default to ```PLATFORM_CLOCK_blah_DEFAULT``` ?

tools/topology/topology2/platform/intel/ssp-default.conf
lgirdwood: ack, mtl.conf

lgirdwood:
> @ranj063 @plbossart we need to figure this out, related code is merged and used in alsa-utils and zephyr ssp driver. This is a working solution, but if you want this cleaner, I need some guidance how to proceed. I would really like to get this for 2.5...

Looks like we need the clock source enum and the mtl.conf

kv2019i:
@juimonen  Can you check how this could be moved to mtl.conf and evaluate the effort? I understand the concern that this bit now is needed in every topology used with mtl, and would b ebetter to have this in one place.

@plbossart @lgirdwood  Ok to merge this now if Jaska commits to making a follow-up change. We really do need this change, we are getting corruption in audio signal without and this could mask all kinds of other things.

tools/topology/topology2/platform/intel/ssp-default.conf
juimonen: no this is the default config for blob ver 1 (legacy). mtl.conf has the blob ver 1.5 def in this PR...

tools/topology/topology2/cavs-nocodec.conf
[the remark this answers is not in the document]
juimonen: I don't currently know a way to add this to all hw_configs in a certain platform in some other way than this

tools/topology/topology2/cavs-nocodec-bt.conf
juimonen: all this is good and dandy, but it would require me to rewrite/move the alsa-utils auxdata parser

juimonen:
probably people would like something like this: https://github.com/juimonen/sof/tree/aux_test

problem is that it just doesn't work with alsa-utils and I kind of don't know why.

tools/topology/topology2/cavs-nocodec-bt.conf
juimonen: We can debate about the sw-philosophical issues for a long time... So ssp blob is created per hw_config, like in BT we have 3 different hw_configs, we create 3 different blobs into nhlt. Aux data is embedded by the spec after the blob. So if the aux data is moved into ssp level, it would mean that all blobs would need a copy of the _same_ aux data. I understand that in the link clock case this could be fine, but link clock is not the only aux data item. So if someone can guarantee me that any aux data items will not be different in any hw_configs per ssp, the move to ssp level would be justified. Otherwise I see the hw_config quite logical place for it.

plbossart: I will repeat my suggestion to ALWAYS provide the AUX information with the clock source, and have the firmware discard what it cannot understand.

The point is that if you select a bitclock that divided from 24.576 MHz then you *have* to select the clock_source 1. If you derive the bitclock from 19.2 then clock_source 0 it is.

There's no differences between platforms, the mapping between clock sources is always the same, and we only have to select between 0 and 1 for now.

In other words, rather than keep this optional make it a mandatory one-size-fits-all.

ranj063 (approved):
im OK to merge this to unblock certain tests. We can work on beautifying it later

lgirdwood (approved):

ranj063:
@lgirdwood @kv2019i lets not merge this one just yet. I have alternative PRs for this

sys-pt1s:

kv2019i:
@juimonen @ranj063 Close this one as a different solution got merged...?

PR #7061, opened 2023-02-08
topology2: deep buffer: use 32 bits to align with other pipelines

RanderWang:
Now deep buffer uses S24_LE but other pipelines use S32_LE internal format . Chrome team found a bug that deep buffer can't work well with I2S codec since fw only support S32_LE for I2S now.

This patch will use S32_LE to align with current requirement

Signed-off-by: Rander Wang <rander.wang@intel.com>


kv2019i (approved):
We'll need to add support for S16/S24 pipelines as well, but that can be done in follow-ups.

lgirdwood (approved):

PR #7065, opened 2023-02-09
cavs: memory bank powerup flow adjustment

mbukowsk:
Memory banks enablement flow is adjusted to reccommended approach. Power status should  be read twice to ensure ebb readiness

Signed-off-by: Michal Bukowski <michal.bukowski@intel.com>

### DIFF
--- a/src/platform/intel/cavs/include/cavs/lib/pm_memory.h
+++ b/src/platform/intel/cavs/include/cavs/lib/pm_memory.h
@@ -72,17 +72,22 @@ static inline void cavs_pm_memory_hp_sram_mask_set(uint32_t mask, int segment,
 {
 	uint32_t expected = enabled ? 0 : mask;
 	uint32_t delay = 0;
+	uint32_t i;
 
 	io_reg_update_bits(SHIM_HSPGCTL(segment), mask, enabled ? 0 : mask);
 	io_reg_update_bits(SHIM_HSRMCTL(segment), mask, enabled ? 0 : mask);
 
-	idelay(MEMORY_POWER_CHANGE_DELAY);
-
-	while ((io_reg_read(SHIM_HSPGISTS(segment)) & mask) != expected) {
+	/* Double check of PG status needed to confirm EBB readiness */
+	for (i = 0; i < 2; i++) {
 		idelay(MEMORY_POWER_CHANGE_DELAY);
-		delay += MEMORY_POWER_CHANGE_DELAY;
-		if (delay >= MEMORY_POWER_CHANGE_TIMEOUT)
-			platform_panic(SOF_IPC_PANIC_MEM);
+
+		while ((io_reg_read(SHIM_HSPGISTS(segment)) & mask) != expected) {
+			idelay(MEMORY_POWER_CHANGE_DELAY);
+			delay += MEMORY_POWER_CHANGE_DELAY;
+			if (delay >= MEMORY_POWER_CHANGE_TIMEOUT)
+				platform_panic(SOF_IPC_PANIC_MEM);
+		}
+		delay = 0;
 	}
 }
 


src/platform/intel/cavs/include/cavs/lib/pm_memory.h
```diff
@@ -84,6 +84,19 @@ static inline void cavs_pm_memory_hp_sram_mask_set(uint32_t mask, int segment,
 		if (delay >= MEMORY_POWER_CHANGE_TIMEOUT)
 			platform_panic(SOF_IPC_PANIC_MEM);
 	}
+
+#if CAVS_VERSION == CAVS_VERSION_2_5
```
pblaszko, line 88: This code is recommended for all cAVS platforms, not only 2_5. as this is "cavs_pm_" function, this check needs to be removed.

lgirdwood (approved):
@mwasko @cujomalainey LGTM.

pblaszko (approved):

dnikodem (approved):

mbukowsk:
@cujomalainey Could you please verify this fix on your setup?

abonislawski (approved):
@wszypelt can you check CI issue?

lgirdwood:
@mbukowsk still draft ?

plbossart:
@lgirdwood this PR looks like a prime candidate for a stable-2.2 update?

mbukowsk:
> @mbukowsk still draft ?

Still fix is not verified and CI not working

mbukowsk:
I've done cosmetic change. for loop is added instead of code duplication. Primarily code was duplicated due to ifdef for cavs 2_5. Fix is applicable for all cavs versions, define is removed so same code is executed in simple loop.

cujomalainey:
Got it cycling, I'll check it tonight, probably will hit about 2k cycles by then

src/platform/intel/cavs/include/cavs/lib/pm_memory.h
```diff
@@ -72,17 +72,22 @@ static inline void cavs_pm_memory_hp_sram_mask_set(uint32_t mask, int segment,
...
-		if (delay >= MEMORY_POWER_CHANGE_TIMEOUT)
-			platform_panic(SOF_IPC_PANIC_MEM);
+
+		while ((io_reg_read(SHIM_HSPGISTS(segment)) & mask) != expected) {
+			idelay(MEMORY_POWER_CHANGE_DELAY);
+			delay += MEMORY_POWER_CHANGE_DELAY;
+			if (delay >= MEMORY_POWER_CHANGE_TIMEOUT)
+				platform_panic(SOF_IPC_PANIC_MEM);
+		}
+	delay = 0;
```
cujomalainey, line 90: indent is incorrect here

cujomalainey:
Probably at just shy of 400 cycles right now. Probably will hit 1k tonight. If its still going by then I will give my +1 so we can submit, but will still let it run through the night

cujomalainey (approved):
935 cycles passed so far

mwasko:
SOFCI TEST

src/platform/intel/cavs/include/cavs/lib/pm_memory.h
```diff
@@ -72,17 +72,22 @@ static inline void cavs_pm_memory_hp_sram_mask_set(uint32_t mask, int segment,
...
-		if (delay >= MEMORY_POWER_CHANGE_TIMEOUT)
-			platform_panic(SOF_IPC_PANIC_MEM);
+
+		while ((io_reg_read(SHIM_HSPGISTS(segment)) & mask) != expected) {
+			idelay(MEMORY_POWER_CHANGE_DELAY);
+			delay += MEMORY_POWER_CHANGE_DELAY;
+			if (delay >= MEMORY_POWER_CHANGE_TIMEOUT)
+				platform_panic(SOF_IPC_PANIC_MEM);
+		}
+		delay = 0;
```
lyakh, line 90: not sure I understand. Does this mean, that once we successfully identified the register value to match what we expect, it can change and we have to run the loop again to catch a second such change? Or do we just have to
```
idelay();
while (io_reg_read() != expected)
	idelay();
idelay();
io_reg_read();
```
? If this is the case, I wouldn't use a loop there but just add that additional register read (possibly with a delay) after the loop.

kv2019i:
https://github.com/thesofproject/sof/issues/6864 hit once in https://sof-ci.01.org/sofpr/PR7065/build3847/devicetest/index.html 

The qemu-failures are more complicated, the changed code affects CNL and ICL and the qemu check in CI now fails, so this seems to be related to the code.

OTOH, we have https://github.com/thesofproject/sof/pull/7000 review which will remove CNL and ICL from mainline, making the failures moot.

@lgirdwood @mwasko guidance how to proceed? I can't immediately see how the change of this PR can break the qemu test, but apparently it does.

src/platform/intel/cavs/include/cavs/lib/pm_memory.h
cujomalainey: regarding your recommendation, is that not what the outer loop enforces?

cujomalainey (commented):
1930 cycles passed

cujomalainey:
stopped my test at 4600 reboots, i think we are good

cujomalainey:
This needs to be cherry picked to Rpl-001 asap

PR #7068, opened 2023-02-09
hda: Maintain dma buffer alignment

makarukp:
Hda chain dma buffer require aligning after its size recalculation.

### DIFF
--- a/src/audio/chain_dma.c
+++ b/src/audio/chain_dma.c
@@ -522,8 +522,6 @@ static int chain_task_init(struct comp_dev *dev, uint8_t host_dma_id, uint8_t li
 		goto error;
 	}
 
-	fifo_size = ALIGN_UP_INTERNAL(fifo_size, addr_align);
-
 	switch (cd->link_connector_node_id.f.dma_type) {
 	case ipc4_hda_link_input_class:
 		/* Increasing buffer size for capture path as L1SEN exit takes sometimes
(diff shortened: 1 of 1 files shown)


src/audio/chain_dma.c
```diff
@@ -544,6 +542,8 @@ static int chain_task_init(struct comp_dev *dev, uint8_t host_dma_id, uint8_t li
 		break;
 	}
 
+	fifo_size = ALIGN_UP_INTERNAL(fifo_size, addr_align);
+
```
plbossart, line 546: the change looks good, but the fifo_size computation above looks quite questionable. There is NOTHING that says the host buffer size is 20ms. That might be an assumption of the OS that shall not be named, but for Linux the host buffer size can be modified at will with a modified hw_params.

This seems very very odd to me.

@jsarha @ujfalusi @kv2019i am I the only one thinking something's not quite right here?

jsarha (approved):
Makes sense and fixes the problem. Approved.

src/audio/chain_dma.c
kv2019i: @plbossart For the L1 exit latency part, this should be independent of host size, so having the calculation here seems ok, but the second comment on position reporting seems indeed odd (without input on actual host buffer size). Of course on Linux side, we could limit buffer sizes if chain-dma is used, but that is definitely not done now (and might get come suprising app behaviour as otherwise SOF is quite relaxed w.r.t. buffer sizes towards the apps).

kv2019i (approved):
The fifo-size calculation does indeed look a bit suspect, but that is not added by this PR, so as such this looks like good to go.

dnikodem (approved):

PR #7071, opened 2023-02-10
[DO NOT MERGE] cause a deliberate panic in sof-test CI runs

kv2019i:


sys-pt1s:

PR #7077, opened 2023-02-13
mux: consider sinks under same sched_comp only for state propogation

johnylin76:
#6882 has introduced the special case for DEMUX in capture pipeline which cannot apply the prepare/reset propagation rule for multiple endpoints. The sink stream direction was used for identifying the special case.

However, it comes another special case needed to be identified for escaping from the multi-endpoint rules. For example:
```
 PCM0P ----> smart_amp ----> SSP0.OUT  # playback pipeline(1)
              ^
              |
 PCM0C <---- demux <----- SSP0.IN      # echo reference pipeline(2)
              |
              v
 PCM1C <---- rtc_audio <---- DMIC      # capture pipeline(3)
```

Consider DEMUX in pipeline(2), it has 3 sinks:
1. feedback reference to smart_amp in pipeline(1)
2. inhabitant sink in the same pipeline(2)
3. echo reference to rtc_audio in pipeline(3)

The 1st and the 3rd sinks should be irrelevant to state propagation through pipeline(2). Rather than checking sink stream direction (playback or capture), the more accurate way for identification is to check pipeline scheduling comp. If it differs from source pipeline the sink should be treated as the special case. Like the above example, the scheduling comp for pipeline(1):SSP0.OUT and pipeline(3):DMIC is not same as pipeline(2):SSP0.IN, while in multiple-endpoint case the sink pipelines will have the same scheduling comp as source pipeline.

### DIFF
--- a/src/audio/mux/mux.c
+++ b/src/audio/mux/mux.c
@@ -650,31 +650,32 @@ static int mux_reset(struct comp_dev *dev)
 	} else {
 		/*
 		 * On capture path, stop propagating reset signal to upstream if any of the sink is
-		 * active. The condition is only applicable to comp DEMUX (single-source to
-		 * multi-sink).
+		 * active and belongs to the pipeline driven by the same scheduling comp. The
+		 * condition is only applicable to comp DEMUX (single-source to multi-sink).
 		 *
-		 * One exception is that the active sink of DEMUX is on playback path, e.g. DEMUX
-		 * on Echo Reference path (capture) linked to the sink SMART_AMP on Speakers path
-		 * (playback). In practice, the playback sink state should not affect the reset
-		 * process through the capture path DEMUX locates on.
+		 * The sink pipeline may be driven by other scheduling comp for DEMUX in particular
+		 * use cases, i.e. the feedback reference path to SMART_AMP in playback pipeline or
+		 * RTC_AUDIO_PROCESSING in capture pipeline. In such cases the sink state should be
+		 * irrelevant to the state change in current pipeline.
 		 */
 		list_for_item(blist, &dev->bsink_list) {
 			struct comp_buffer *sink = container_of(blist, struct comp_buffer,
 								source_list);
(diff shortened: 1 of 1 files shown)


lyakh:
ever more special cases... What about a simple case like
```
              /-----------SSP0.out
PCM0 ---> demux
              \-----------SSP1.out
```
also two different scheduling components but the status is relevant?..

johnylin76:
> ever more special cases... What about a simple case like
> 
> ```
>               /-----------SSP0.out
> PCM0 ---> demux
>               \-----------SSP1.out
> ```
> 
> also two different scheduling components but the status is relevant?..

Does the current multiple endpoint have the plan to handle this case? AFAIK, it now only targets for multi-PCM to single-DAI topologies, i.e.
```
 PCM0P --(1)--\
               MIXER --(3)--> SSP0.OUT
 PCM1P --(2)--/

 PCM0C <--(1)-\
               DEMUX <--(3)-- DMIC.IN
 PCM1C <--(2)-/
```
So basically the DEMUX sink state checking logic is only applicable on capture paths.

lgirdwood:
@lyakh pls review.

lgirdwood:
@ranj063 pls also review.

lyakh:
> @lyakh pls review.

@lrgirdwo I don't know which review criteria to apply here. If it were a PR to an actively developed branch I'd disagree with adding support for a specific class of topologies while breaking another (supposedly unused) class of topologies even more. But since this PR is to a "stable" branch I'm not sure how to grade this. I think a "correct" solution to this problem would be to handle it generically, e.g. we could decide to differentiate between "streaming" and "reference" links and only propagate any pipeline triggers across streaming and not across reference connections, but that could be too big of a change for a "stable" branch.

johnylin76:
Thanks for the review. We planned to close this for now due to advantages for the alternative solution.

In spite of that, I would still like to know about https://github.com/thesofproject/sof/pull/7077#issuecomment-1427625728
@plbossart perhaps?

PR #7078, opened 2023-02-13
Audio: Fix the peak volume calculation error in volume component

andrula-song:
Fix the peak volume calculation error in volume component, calculate the maximum absolute value of input as peak volume as close source firmware did.

Signed-off-by: Andrula Song <andrula.song@intel.com>

PR #7087, opened 2023-02-14
Audio: Volume: Add HiFi4 implementation of volume

andrula-song:
Add peak_vol to calculate peak volume with continuous sample processing. Since HiFi4 supports 2 circular buffers, so set both vol and peak_vol as circular buffer to store gain and temp peak value 4 times. And recaculate the peak volume after all samples are processed is faster than processing samples one by one.

Signed-off-by: Andrula Song <andrula.song@intel.com>

### DIFF
--- a/src/audio/module_adapter/module/volume/volume.c
+++ b/src/audio/module_adapter/module/volume/volume.c
@@ -391,7 +391,6 @@ static int volume_init(struct processing_module *mod)
 		comp_err(dev, "volume_init(): Failed to allocate %d", vol_size);
 		return -ENOMEM;
 	}
-
 	md->private = cd;
 
 	/* Set the default volumes. If IPC sets min_value or max_value to
@@ -572,6 +571,16 @@ static int volume_init(struct processing_module *mod)
 		comp_err(dev, "volume_init(): Failed to allocate %d", vol_size);
 		return -ENOMEM;
 	}
+	/*
+	 * malloc memory to store temp peak volume 4 times to ensure the address
+	 * is 8-byte aligned for multi-way xtensa intrinsic operations.
+	 */
+	cd->peak_vol = rmalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, vol_size);
+	if (!cd->peak_vol) {
+		rfree(cd);
+		comp_err(dev, "volume_init(): Failed to allocate peak_vol %d", vol_size);
+		return -ENOMEM;
+	}
 
 	md->private = cd;
 
@@ -650,6 +659,7 @@ static int volume_free(struct processing_module *mod)
 	/* clear mailbox */
 	memset_s(&regs, sizeof(regs), 0, sizeof(regs));
 	mailbox_sw_regs_write(cd->mailbox_offset, &regs, sizeof(regs));
+	rfree(cd->peak_vol);
 #endif
 
 	comp_dbg(mod->dev, "volume_free()");
--- a/src/audio/module_adapter/module/volume/volume_hifi4.c
+++ b/src/audio/module_adapter/module/volume/volume_hifi4.c
@@ -17,21 +17,15 @@
 #include <stddef.h>
 #include <stdint.h>
 
-LOG_MODULE_DECLARE(volume_hifi3, CONFIG_SOF_LOG_LEVEL);
+LOG_MODULE_DECLARE(volume_hifi4, CONFIG_SOF_LOG_LEVEL);
 
 #include <sof/audio/volume.h>
 
-#if VOLUME_HIFI3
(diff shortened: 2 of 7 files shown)


src/include/sof/audio/volume.h
```diff
@@ -34,15 +34,30 @@
...
 struct sof_ipc_ctrl_value_chan;
 
-#define CONFIG_GENERIC
-
 #if defined(__XCC__)
 #include <xtensa/config/core-isa.h>
 
-#if XCHAL_HAVE_HIFI3
-#undef CONFIG_GENERIC
+#if XCHAL_HAVE_HIFI4  /* HiFi 4 */
```
lgirdwood, line 40: Btw, XCHAL is xtos specific, we need a high level CONFIG_ARCH_XTENSA_HIFI4 etc and use them instead. @singalsu fyi.

src/audio/module_adapter/module/volume/volume_hifi4.c
```diff
@@ -0,0 +1,394 @@
...
+#include <sof/common.h>
+#include <ipc/stream.h>
+#include <stddef.h>
+#include <stdint.h>
+
+LOG_MODULE_DECLARE(volume_hifi4, CONFIG_SOF_LOG_LEVEL);
+
+#include <sof/audio/volume.h>
+
+#if VOLUME_HIFI4
```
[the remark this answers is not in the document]
singalsu: Generally yes, but for testing it could be practical to change build of only one component e.g. to avoid everything to slow down too much. But that can be patched too when debugging so I don't have strong opinions for this.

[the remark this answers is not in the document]
lgirdwood: We can use a local macro for this e.g.
```c
#define VOLUME_HIFI4 (HIFI4 && !VOLUME_FORCE_GENERIC)
```

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -572,6 +571,16 @@ static int volume_init(struct processing_module *mod)
 		comp_err(dev, "volume_init(): Failed to allocate %d", vol_size);
 		return -ENOMEM;
 	}
+	/*
+	 * malloc memory to store temp peak volume 4 times to ensure the address
+	 * is 8-byte aligned for multi-way xtensa intrinsic operations.
+	 */
+	cd->peak_vol = rmalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, vol_size);
```
lyakh, line 578: have to free `peak_vol` in the error case at line 607 below.

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -572,6 +571,16 @@ static int volume_init(struct processing_module *mod)
...
+	if (!cd->peak_vol) {
+		rfree(cd);
```
lyakh, line 580: don't you need to free `cd->vol` here too? Looks like this function's error handling has to be fixed - the error path at new line number 607 below isn't freeing any allocations here. And in general maybe that whole `instance_id` check can be moved to the very top of the function before all the allocations?

src/audio/module_adapter/module/volume/volume_hifi4.c
```diff
@@ -0,0 +1,394 @@
...
+	/** to ensure the address is 8-byte aligned and avoid risk of
+	 * error loading of volume gain while the cd->vol would be set
+	 * as circular buffer
+	 */
+	if (cd->copy_gain)
+		vol_store_gain(cd, channels_count);
+
+	buf = (ae_f32x2 *)cd->vol;
+	buf_end = (ae_f32x2 *)(cd->vol + channels_count * 2);
+	vol = (ae_f32x2 *)buf;
```
lyakh, line 192: `vol` and `buf` have the same type, no need to cast

lyakh (changes requested):

src/audio/module_adapter/module/volume/volume.c
andrula-song: thanks, done.

src/include/sof/audio/volume.h
aiChaoSONG: XCHAL_HAVE_xxx is provided by Xtensa toolchain (processror config), we have a copy in src/platform/core_code_name/include/arch just for gcc complication.  it is toolchain specific indeed.

andrula-song:
Since this PR is the hifi4 version implementation of peak volume detection , So combine into the PR https://github.com/thesofproject/sof/pull/7078

andrula-song:
Close this PR since https://github.com/thesofproject/sof/pull/7078 has been merged.

PR #7089, opened 2023-02-14
Data processing scheduler - SOF tasks as Zephyr preemptive threads

marcinszkudlinski:
DP scheduler is a scheduler that creates a separate preemptible Zephyr thread for each SOF task
There's only one instance of DP in the system, however, the threads may run on any core
Threads are pinned to specific core, there's no SMP processing.

The task execution may be delayed and task may be re-scheduled periodically

**NOTE:** 
- task - means a SOF task
- thread - means a Zephyr preemptible thread

**USECASES:**
- to run modules that have longer processing time than 1 LL cycle. I.e. speech recognition, that may run for seconds
- to create pipelines that have some of modules executed on other cores that main LL thread for a pipeline
- to general run zephyr thread using SOF task abstraction (like IPC thread)

TODOs

**EDF:**
Threads run on the same priority, lower than thread running LL tasks. Zephyr EDF mechanism
is used for decision which thread/task is to be scheduled next. The DP scheduler calculates
the task deadline and set it in Zephyr thread properties, the final scheduling decision is made
by Zephyr.

**task with budgets:**
task start at high priority, after given execution time the priority is dropped


iuliana-prodan:
@marcinszkudlinski what will be the difference between DP scheduling and DMA domain? 
They are both using zephyr's threads, but one will be preempted by DMA IRQ (DMA domain) and the other by timer (DP scheduling)?

marcinszkudlinski:
@iuliana-prodan DMA domain uses 1 zephyr thread, DP uses separate thread for each component/module
in DP can have different tick for each module 
i.e. you can create a pipeline like:

DMIC (1ms, core 0) --> processing (10ms, core 1) --> HDA (1ms, core 0)
of course last module would need to wait 10ms before 1st processing (or processing silence till than) 

and next step: when there are more DP modules running on single core, i.e. one 1ms tick, second 2 seconds tick (like speech recognition) - DP will allow processing using EDF scheduling with preemtion and avoid starvation of 1ms module

zephyr/Kconfig
```diff
@@ -37,4 +37,13 @@ config DMA_DOMAIN
 	  and dma_single_chan in Zephyr once it becomes more
 	  stable.
 
+config ZEPHYR_DP_SCHEDULER
+	bool "use Zephyr thread based DP scheduler"
+	default y if ACE
+	default n
+	depends on IPC_MAJOR_4
```
kv2019i, line 44: You can also "depends on ZEPHYR_SOF_MODULE" to hide it from builds where Zephyr is not used.

src/include/sof/schedule/dp_schedule.h
```diff
@@ -0,0 +1,99 @@
...
+#include <sof/schedule/task.h>
+#include <sof/trace/trace.h>
+#include <user/trace.h>
+#include <stdint.h>
+
+/**
+ *
+ * DP scheduler is a scheduler that creates a separate preemptible Zephyr thread for each SOF task
+ *   There's only one instance of DP in the system, however, the threads may run on any core
+ *   Threads are pinned to specific core, there's no SMP processing.
```
kv2019i, line 20: Maybe easier to express "threads can be assigned to any core in the system, but each thread is pinned to a single core for its execution." Or maybe not, but you get the picture. So you can pick any core, but tasks stay on the core they were started on.

src/include/sof/schedule/dp_schedule.h
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * The task execution may be delayed and task may be re-scheduler periodically
```
kv2019i, line 22: typo: re-scheduled

src/include/sof/schedule/dp_schedule.h
```diff
@@ -0,0 +1,99 @@
...
+ * NOTE: delayed start and rescheduling takes place in sync with LL scheduler, meaning the
+ *   DP scheduler is triggered as the last task of LL. That implicates a limitation: LL scheduler
```
kv2019i, line 24: implies a limitation

src/include/sof/schedule/dp_schedule.h
```diff
@@ -0,0 +1,99 @@
...
+ *   MUST be running on primary core in order to have this feature working.
```
kv2019i, line 25: given we currently run a LL scheduler on each core, might be easier to express this that the DP scheduler is triggered by the last task of the LL running on primary core..?

src/include/sof/schedule/dp_schedule.h
```diff
@@ -0,0 +1,99 @@
...
+ *   It is fine, because rescheduling is a feature used for data processing when a pipeline is
+ *   running.
+ *
+ * Other possible usage of DP scheduler is to schedule task with DP_SCHEDULER_RUN_TASK_IMMEDIATELY
+ *   as start parameter. It will force the task to work without any delays and asyn to LL.
```
kv2019i, line 30: async to LL

src/include/sof/schedule/dp_schedule.h
```diff
@@ -0,0 +1,99 @@
...
+ *   This kind of scheduling may be used for tasks like IPC.
```
kv2019i, line 31: This is a welcome addition to current usage of the EDF for this purpose.

src/include/sof/schedule/dp_schedule.h
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * TODO  - EDF:
+ * Threads run on the same priority, lower than thread running LL tasks. Zephyr EDF mechanism
+ * is used for decision which thread/task is to be scheduled next. The DP scheduler calculates
+ * the task deadline and set it in Zephyr thread properties, the final scheduling decision is made
+ * by Zephyr.
+ *
+ * Each time tick the scheduler iterates through the list of all active tasks and calculates
+ * a deadline based on
+ *  - knowledge how the modules are binded
```
kv2019i, line 54: grammar: are bound

src/schedule/zephyr_dp_schedule.c
```diff
@@ -0,0 +1,376 @@
...
+				/* mark the task as cancelled */
+				scheduler_dp_task_cancel(&dp_sch, task);
+				break;
+
+			case SOF_TASK_STATE_COMPLETED:
+				break;
+
+			default:
+				/* illegal state */
+				assert(false);
```
kv2019i, line 187: I can see why you wanted to raise the assert discussion. ;) But agreed, this is definitely good example of assert that would be good to leave in product builds. OTOH, one could argue along the lines of https://www.kernel.org/doc/Documentation/process/coding-style.rst "Do not crash the kernel" , that even with state corruption, system should not be crashes if not absolutely necessary. We do have sof_panic() and k_panic() available for those cases where system should absolutely panic and that's something you could use here as well and it'd always be in all builds.

src/schedule/zephyr_dp_schedule.c
```diff
@@ -0,0 +1,376 @@
...
+{
+	struct dp_schedule_data *sch = data;
+	struct dp_task_pdata *pdata = task->priv_data;
+	uint32_t lock_flags;
+
+	zephyr_dp_lock(&lock_flags);
+
+	if (task->state != SOF_TASK_STATE_INIT &&
+	    task->state != SOF_TASK_STATE_COMPLETED)
+		return -EINVAL;
```
kv2019i, line 217: Oops, unlock missing.

src/schedule/zephyr_dp_schedule.c
```diff
@@ -0,0 +1,376 @@
...
+	/* start the thread - it will immediately stop at a semaphore */
+	k_thread_start(thread_id);
+
+	return 0;
+err:
+	/* cleanup - free all allocated resources */
+	if (p_stack)
+		rfree(p_stack);
+	if (thread_id)
+		k_thread_abort(thread_id);
```
kv2019i, line 370: Should k_thread_abort be done before p_stack is freed? I guess this doens't matter us thread is not started yet, but looks a bit weird.

src/include/sof/schedule/dp_schedule.h
```diff
@@ -0,0 +1,99 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
```
kv2019i, line 1: drop the "MTL: " prefix in git commit, not really needed

src/audio/component.c
```diff
@@ -302,8 +302,9 @@ int comp_copy(struct comp_dev *dev)
```
kv2019i, line 302: And here as well, drop the "MTL: " prefix.

kv2019i (changes requested):
Excellent V1 @marcinszkudlinski ! I really only have one or two bigger issues in the code. I added a lot of typo/grammer comments (as it was just convenient to add them while reading the documentation parts), but these are really minor, I could understand the concepts.

src/schedule/zephyr_dp_schedule.c
```diff
@@ -0,0 +1,376 @@
...
+			state = task_run(task);
+
+			/*
+			 * handle task state - the task may be canceled by external call
+			 * or request termination itself
+			 */
+			zephyr_dp_lock(&lock_flags);
+
+			/*
+			 * check if task is still running, may have be canceled by external call
```
lyakh, line 165: typo: "may have been"

src/audio/pipeline/pipeline-schedule.c
```diff
@@ -355,6 +375,49 @@ int pipeline_comp_task_init(struct pipeline *p)
...
+					     comp->ipc_config.core,
+					     TASK_DP_STACK_SIZE,
+					     ZEPHYR_DP_THREAD_PRIORITY);
+		if (ret < 0) {
+			rfree(component_task);
+			return ret;
+		}
+	}
+
+	comp->task = component_task;
```
lyakh, line 415: so if in the beginning `comp->task` wasn't `NULL`, you make it `NULL` now?

src/include/sof/audio/component_ext.h
```diff
@@ -147,13 +148,46 @@ static inline int comp_trigger_remote(struct comp_dev *dev, int cmd)
...
+{
+	int ret;
+
+	ret = dev->drv->ops.trigger(dev, cmd);
+
+	/* start a thread in case of shared component or DP scheduling */
+	if (dev->task) {
+		/* schedule or cancel task, component must be shared or in DP domain */
+		assert((dev->is_shared) ||
+		       (dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_DP));
```
lyakh, line 161: both internal parentheses are superfluous and at least the former ones really cannot even cause any confusion

src/include/sof/schedule/dp_schedule.h
marcinszkudlinski: "ask of the LL running on primary core" sounds ok,
I just need to make it more clear - we CAN have LL running on every core, but we DON'T have one till a pipeline is running on the core in question. Theoretically, one can start an LL pipeline on core 1, leaving core 0 idle, can't he? In this case DP won't work properly.

lyakh (changes requested):
ATM this cannot work with a component on a core other than the pipeline. The problem is accessing buffer object list elements without locking in places like `pipeline_for_each_comp()`. We should block attempts to specify a core different from the pipeline core for DP components until such support is implemented.

src/include/sof/schedule/dp_schedule.h
marcinszkudlinski: > ATM this cannot work with a component on a core other than the pipeline. The problem is accessing buffer object list elements without locking in places like `pipeline_for_each_comp()`. We should block attempts to specify a core different from the pipeline core for DP components until such support is implemented.

I assure it can, at least works in tests now. The whole point of DP is to mix 1) diffrent ticks 2) diffrent cores

marcinszkudlinski:
> ATM this cannot work with a component on a core other than the pipeline. The problem is accessing buffer object list elements without locking in places like `pipeline_for_each_comp()`. We should block attempts to specify a core different from the pipeline core for DP components until such support is implemented.

I assure it can, at least works in tests now. The whole point of DP is to mix 1) diffrent ticks 2) diffrent cores

src/include/sof/schedule/dp_schedule.h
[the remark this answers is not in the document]
marcinszkudlinski: 1) Won't simplify much, couple of lines
2) in future there's a concept of "taksks with budgets", not supported by zephyr directly. We discuss it with Zephyr and they only provide some callback to make it possible to implement
3) I would keep an abstraction for a) porting b) common API with schedulers like LL working in completely diffrent way

src/include/sof/schedule/dp_schedule.h
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * \brief Set the Data Processing scheduler to be accessible at secondary cores
+ */
+int scheduler_dp_init_secondary_core(void);
+
+/**
+ * dp_scheduler_run should be run in LL scheduler loop, after all LL task have been executed
+ */
+void dp_scheduler_ll_tick(void);
```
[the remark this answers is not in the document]
marcinszkudlinski: true, should be unified

src/schedule/zephyr_dp_schedule.c
```diff
@@ -0,0 +1,376 @@
...
+ * function called event LL tick, always on primary core
+ *
+ * TODO:
+ * the scheduler should here calculate deadlines of all task and tell Zephyr about them
+ * Currently there's an assumption that the task is always ready to run
+ *
+ * TODO: tasks with budgets
+ */
+void dp_scheduler_ll_tick(void)
+{
```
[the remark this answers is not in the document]
marcinszkudlinski: Yes, we do. The while idea is to check the state of all DP tasks immediately after LL is finished. 
Currently it does not do much - just trigger of periodic triggering, but much more sophisticated EDF algorithm is coming where deadlines need to be estimated every tick

src/schedule/zephyr_dp_schedule.c
```diff
@@ -0,0 +1,376 @@
...
+		k_sem_give(&pdata->sem);
+	} else {
+		/* wait for tick */
+		pdata->ticks_to_trigger	=
+				ROUND_DOWN(start, LL_TIMER_PERIOD_US) / LL_TIMER_PERIOD_US;
+		task->state = SOF_TASK_STATE_QUEUED;
+	}
+
+	/* add a task to DP scheduler list */
+	list_item_prepend(&task->list, &sch->tasks);
```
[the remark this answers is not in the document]
marcinszkudlinski: ok, would look better, changing, but is in lock so not a big deal

marcinszkudlinski:
@lyakh maybe i'm missing something how the buffers between componenets are managed and cache is managed. Lets discuss it in meeting. As I said, the main reason for DP is to allow offloading time consuming operation with long buffering / long processing to secondary cores. If it require some changes in pipeline_for_each_comp() - we need to implement it asap and add to this PR

src/include/sof/schedule/dp_schedule.h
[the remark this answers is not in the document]
lyakh: I meant tasks like IPC. IPC with Zephyr is currently using the EDF scheduler, which is just a thin API wrapper around Zephyr work queues, not much more than that.

lyakh:
> > ATM this cannot work with a component on a core other than the pipeline. The problem is accessing buffer object list elements without locking in places like `pipeline_for_each_comp()`. We should block attempts to specify a core different from the pipeline core for DP components until such support is implemented.
> 
> I assure it can, at least works in tests now. The whole point of DP is to mix 1) diffrent ticks 2) diffrent cores

@marcinszkudlinski passed tests aren't a proof of correctness. We need to design and implement a robust approach to running linked components or pipelines on different cores instead of enabling it and then fixing inevitable bugs. This PR also only disables local interrupt as protection while aiming to support multiple cores. It seems like testing wasn't excessive.

marcinszkudlinski:
> > > ATM this cannot work with a component on a core other than the pipeline. The problem is accessing buffer object list elements without locking in places like `pipeline_for_each_comp()`. We should block attempts to specify a core different from the pipeline core for DP components until such support is implemented.
> > 
> > 
> > I assure it can, at least works in tests now. The whole point of DP is to mix 1) diffrent ticks 2) diffrent cores
> 
> @marcinszkudlinski passed tests aren't a proof of correctness. We need to design and implement a robust approach to running linked components or pipelines on different cores instead of enabling it and then fixing inevitable bugs. This PR also only disables local interrupt as protection while aiming to support multiple cores. It seems like testing wasn't excessive.

ok, so we need to fix this issue ASAP, secondary cores support is a must

marcinszkudlinski:
Missing commit added - all "resolved" conversation should be fixed in the code now

src/include/sof/schedule/dp_schedule.h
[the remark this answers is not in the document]
marcinszkudlinski: I know about current zephyr_EDF, but we need to have IPC to be preemptive also, as a "task with budget" - IPC should be processed very quickly, but if processing takes to long it must yield to LL. Current zephyr implementation of EDF, as you said, does not do much and is not a really EDF, just "start the earliest deadline task and hope it finish processing in time".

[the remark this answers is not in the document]
kv2019i: @marcinszkudlinski @lyakh Let's not detour to IPC too much here. The current IPC with EDF is clearly not optimal and should be either written on top of native work queues, or something else if we need to have preemption after budget has expired (has not implemented so far in SOF, so this itself is a new feature).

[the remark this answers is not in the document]
marcinszkudlinski: ok, I know is not implemented, but is scheduled (sic!) for implementation :)

kv2019i (approved):
Good for me. The open issues w.r.t. multicore safety in IPC4 pipeline code, are not specific to DP, so I think these can be addressed in separate PRs.

src/include/sof/schedule/dp_schedule.h
```diff
@@ -0,0 +1,100 @@
...
+ *   as start parameter. It will force the task to work without any delays and async to LL.
+ *   This kind of scheduling may be used for tasks like IPC.
+ *
+ * Task run() may return:
+ *  SOF_TASK_STATE_RESCHEDULE - the task will be rescheduled as specified in scheduler period
+ *				note that task won't ever be rescheduled if LL is not running
+ *  SOF_TASK_STATE_COMPLETED  - the task will be removed from scheduling, but not terminated
+ *				calling schedule_task will add the task to processing again
+ *				task_complete() will be called
+ *  SOF_TASK_STATE_CANCEL     - the task will be cancelled permanently,
```
marcinszkudlinski, line 40: self review: DP should be able to reschedule the task after canceling. 
To remove a task from periodic scheduling task_cancel must be called (of SOF_TASK_STATE_CANCEL/ SOF_TASK_STATE_COMPLETED returned trom task_run). Anyway, there's no reason why the task is terminated when cancelled and can't be scheduled again. To be fixed.

lyakh:
> @lyakh maybe i'm missing something how the buffers between componenets are managed and cache is managed. Lets discuss it in meeting. As I said, the main reason for DP is to allow offloading time consuming operation with long buffering / long processing to secondary cores. If it require some changes in pipeline_for_each_comp() - we need to implement it asap and add to this PR

@marcinszkudlinski here's an example: consider a topology like
```
	LL			DP
	copier.1 -> buffer.1 -> volume.1
```
Then you can have a race like
```
LL / core 0:						DP / core 1

pipeline_copy(copier.1)
pipeline_comp_copy(start = copier.1, dir=UPSTREAM)
pipeline_for_each_comp(copier.1, dir=UPSTREAM)
buffer = buffer.1
buffer_c->walking = true;
<core 0 writes cache back to SRAM,
 SRAM has buffer->walking = true>			access buffer, read SRAM into core 1 cache
							module_adapter_simple_copy()
							sinks_c[i++] = buffer_acquire(sink);
							comp_get_copy_limits_frame_aligned()
							<core 1 reads SRAM into cache
							 sink_c[0]->walking = true>
buffer_c->walking = false;
<core 0 writes cache back to SRAM,
 SRAM has buffer->walking = false>
							<core 0 writes cache back to SRAM,
							 SRAM has buffer->walking = true>
					BUG
```

src/include/sof/schedule/dp_schedule.h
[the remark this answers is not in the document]
aborisovich: @lyakh, @kv2019i we detour to IPC for a reason. Works on entering D3 power state with Zephyr Power Manager had been suspended as there is a bug in zephyr k_waitable abstraction that handles scheduling ipc task for future (after wake from D3). No progress had been made on the subject so we are trying hard with @marcinszkudlinski to replace the scheduling system.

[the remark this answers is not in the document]
lyakh: @marcinszkudlinski @aborisovich @kv2019i why are you saying that the current IPC thread isn't preemptible? As long as it has a lower priority than the LL scheduler it will be preempted by it, regardless of any budgets. When the next timer interrupt triggers, interrupting the IPC thread, on exit the Zephyr scheduler will select a runnable task with the highest priority and that will be the LL scheduler thread.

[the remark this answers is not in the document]
marcinszkudlinski: All tasks in current "edf" implementation is based on Zephyr work queue and run by a single Zephyr thread. Right, this thread can be preempted. BUT the tasks themselves are cooperative to each other. All the current edf can do is to decide which task from the queue starts first

marcinszkudlinski:
@lyakh 
I'll implement consumer-producer save buffering to fix this

src/include/sof/schedule/dp_schedule.h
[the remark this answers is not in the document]
andyross: Indeed, the Zephyr work queue is an in-order list of items to run, it doesn't handle preemption between tasks (which aren't threads, so they can't **be** preempted, they're just callbacks that run on the same thread).   This isn't really a bug, it's just the way the data structure works.

If you want to preempt something, it has to be running in a thread with its own stack that can be suspended.

Note that there's an alternative, the "p4wq", which is a little more complicated, but based on the idea of running queue items in pooled threads such that they all have their own priorities and can preempt each other, but without the overhead of having to spawn one thread per task.  This was actually written with SOF in mind originally, as it happens, though it will likely require some work to enable.  See https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/sys/p4wq.h

lyakh:
> @lyakh I'll implement consumer-producer save buffering to fix this

@marcinszkudlinski what I think could be a good solution, which we'll also need to support multicore streams, is splitting `struct comp_buffer` into two parts: one with a list head and `.walking` and a link to another one with the rest. The former would be accessed uncached and the latter cached

src/include/sof/schedule/dp_schedule.h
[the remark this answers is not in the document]
lyakh: @andyross maybe I misunderstood your "it will likely require some work to enable" but it is already used by SOF for IDC

zephyr/Kconfig
```diff
@@ -37,4 +37,14 @@ config DMA_DOMAIN
...
+	depends on ZEPHYR_SOF_MODULE
+	help
+	  Enable Data Processing preemptive scheduler based on
```
lgirdwood, line 47: Can we advertise the features here so it's easy to understand the benefits.

src/schedule/zephyr_dp_schedule.c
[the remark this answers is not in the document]
lgirdwood: @marcinszkudlinski I assume if we have a DP task that takes longer than the LL tick period of 1ms (e.g  a 2ms duration media decode for 10ms PCM audio) that the next LL will just prempt the DP task and return to it again after the next LL ?

src/include/sof/audio/component_ext.h
```diff
@@ -147,13 +148,46 @@ static inline int comp_trigger_remote(struct comp_dev *dev, int cmd)
 	return idc_send_msg(&msg, IDC_BLOCKING);
 }
 
+static inline int comp_trigger_local(struct comp_dev *dev, int cmd)
+{
+	int ret;
+
+	ret = dev->drv->ops.trigger(dev, cmd);
```
lgirdwood, line 155: Dont we need to use the wrapper API here ?

marcinszkudlinski: > hat the next LL will just prempt the DP task and return to it again after the next LL

Yes, DP tasks (not a single task, multiple tasks) are to be run as threads, lower priority than LL. 

Of course if a DP module is triggered less frequently than LL, the buffers between them must be big enough

marcinszkudlinski: > Dont we need to use the wrapper API here ?

you mean **static inline int comp_trigger** ?
No, comp_trigger calls comp_trigger_local or comp_trigger_remote, so we're in api wrapper in fact

marcinszkudlinski:
NOTES:
this is a rebase to newest main
> marcinszkudlinski force-pushed the dp_scheduler_public branch from 627a937 to c56a499 1 minute ago

this is new version after review changes
> marcinszkudlinski force-pushed the dp_scheduler_public branch from c56a499 to be1823f 1 minute ago

To speed up the merging I introduced a lock - DP must be on the same core as LL pipleline. 
Main fix is on the way

aborisovich (approved):
Working on refactoring IPC to use DP scheduler.
Right now IPC tasks are scheduled & executed so I'm happy with this PR.

src/schedule/zephyr_dp_schedule.c
marc-hb: > this is definitely good example of assert that would be good to leave in product builds.

Let's stick to most common industry terminology and call that a `panic()` or `BUG_ON()`,  `error()`, etc. By definition `assert()`s are turned off to save cycles or memory.

src/include/sof/schedule/dp_schedule.h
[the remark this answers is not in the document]
lyakh: So, yes, as a further optimisation we could in the future consider moving this to p4wq, that would allow us to use threads more optimally - use a common pool of threads instead of a thread per task, right @andyross ?

src/schedule/zephyr_dp_schedule.c
```diff
@@ -0,0 +1,378 @@
...
+DECLARE_TR_CTX(dp_tr, SOF_UUID(dp_sched_uuid), LOG_LEVEL_INFO);
+
+struct scheduler_dp_data {
+	struct list_item tasks;		/* list of active dp tasks */
+	struct k_spinlock lock;		/* synchronization between cores */
+};
+
+struct task_dp_pdata {
+	k_tid_t thread_id;		/* zephyr thread ID */
+	k_thread_stack_t *p_stack;	/* pointer to thread stack */
```
lyakh, line 35: also as a further enhancement - we could try explicitly declare the stack pointer as `__sparse_cache` since stack should be cached.

src/schedule/zephyr_ll.c
```diff
@@ -240,6 +241,12 @@ static void zephyr_ll_run(void *data)
 
 	zephyr_ll_unlock(sch, &flags);
 
+#if CONFIG_ZEPHYR_DP_SCHEDULER
+	/* trigger calculation of DP tasks as the last action of LL scheduler. Primary Core only*/
+	if (cpu_get_id() == PLATFORM_PRIMARY_CORE_ID)
+		scheduler_dp_ll_tick();
```
lyakh, line 247: hmm, now this makes me a bit uncomfortable... I think it would be better to make this just a normal LL task... Would that work? Just give it a priority 1 lower than all LL pipeline tasks

src/schedule/zephyr_dp_schedule.c
```diff
@@ -0,0 +1,378 @@
...
+
+	task->state = SOF_TASK_STATE_CANCEL;
+	list_item_del(&task->list);
+
+	scheduler_dp_unlock(lock_key);
+
+	return 0;
+}
+
+static int scheduler_dp_task_complete(void *data, struct task *task)
```
lyakh, line 114: AFAICS `schedule_task_complete()` is never used in our code base. Do you plan to use it? If not, maybe you could just leave it NULL. And maybe we can remove it.

src/schedule/zephyr_dp_schedule.c
```diff
@@ -0,0 +1,378 @@
...
+	/* this is asyn cancel - mark the task as canceled and remove it from scheduling */
+	lock_key = scheduler_dp_lock();
+
+	task->state = SOF_TASK_STATE_COMPLETED;
+	list_item_del(&task->list);
+
+	scheduler_dp_unlock(lock_key);
+
+	/* let the thread run once again to call task_complete in a thread context */
+	k_sem_give(&pdata->sem);
```
lyakh, line 129: ...if you do need `schedule_task_complete()` I'll have to ask why here you do wake up the task and in `.cancel()` you don't

src/schedule/zephyr_dp_schedule.c
```diff
@@ -0,0 +1,378 @@
...
+	void *p_stack = NULL;
+	int ret;
+
+	if (task->priv_data)
+		return -EEXIST;
+
+	ret = schedule_task_init(task, uid, SOF_SCHEDULE_DP, 0, ops->run, data, core, 0);
+	if (ret < 0) {
+		tr_err(&dp_tr, "zephyr_dp_task_init(): schedule_task_init failed");
+		goto err;
```
lyakh, line 312: you can still just return immediately here

andyross:
> So, yes, as a further optimisation we could in the future consider moving this to p4wq, that would allow us to use threads more optimally - use a common pool of threads instead of a thread per task, right @andyross ?

That's exactly what it's intended to do, yes.  If your tasks are all separate threads, you need to allocate a full stack for every one and when you switch tasks you need to do a full context switch through the kernel (but you retain the ability to preempt tasks with other higher priority tasks).  With p4wq, in the uncontended/unpreempted case there's no context switch and no extra thread needed, the worker thread just reaches the end of a callback and then pulls the next one out of the list.  And you only need enough threads in the pool to handle the maximum preemption depth desired by the app (e.g. one for each distinct priority of task).

I haven't had time to look carefully here, but yes: I'd look at that planning sooner rather than later.  Threads are extremely expensive.

src/schedule/zephyr_dp_schedule.c
andyross: FWIW, the relevant Zephyr APIs are `k_panic()` (intended for fatal system-wide errors) and `k_oops()` (thread-terminating errors).  In fact both do the same thing, and arrive in `k_sys_fatal_error_handler()` with a "reason" code of `K_ERR_KERNEL_{OOPS,PANIC}`.   Despite the k_ prefix, that function is actually application-defined and can do whatever you want.  The default is a weak function in kernel/fatal.c that just logs the panic (but doesn't dump the stack, we should add that) and halts the system.

src/schedule/zephyr_dp_schedule.c
lyakh: FYI #7200

src/schedule/zephyr_dp_schedule.c
```diff
@@ -0,0 +1,378 @@
...
+
+		}
+
+		/* call task_complete  */
+		if (task->state == SOF_TASK_STATE_COMPLETED) {
+			/*
+			 * clear semaphore count to avoid executing task_complete twice
+			 * in case the task_complete have been called during task execution
+			 */
+			k_sem_reset(&task_pdata->sem);
```
andyross, line 210: This looks racy.  The k_sem_give() call does not seem to be done with the spinlock held, so if you reach this at the same time the "tick" code is trying to wake you up for the next event, you'll lose the update and sleep forever.

Basically: the counting behavior of semaphores is your friend and prevents this kind of race.  If you need to reset it in a context that's not some kind of error handling, it's probably because there's a bug with the state management (e.g. why was "task_complete() called during task execution", isn't that a bug? If not, shouldn't task_complete() be responsible for squashing the duplicates?)

andyross:
One nitpick left in a comment.  But... really reading through this I have to say that this code is almost 1:1 equivalent with the existing Zephyr work queue (not the p4wq, just the regular one), except that each queue gets only one work item ("task" in SOFese) associated with it, and nothing here cares about the "delayed work" features that complicate the Zephyr implementation.

Even absent the discussion about whether to move to p4wq[1], maybe it might be a good idea to refactor this to wrap a work queue?  That would remove all the responsibility for thread lifecycle management and probably simplify things.

[1] Which, it should be pointed out, can also share threads with the EDF scheduler too.  Threads are threads.

marc-hb:
> maybe it might be a good idea to refactor this to wrap a work queue? 

Even if that's not done now for obvious time constraints, please make sure this refactoring suggestion is at least present as a comment in the source code so no one in the future spends more time maintaining parts of this code than they would replacing them with the equivalent code in Zephyr.

Please add this analysis too in the source code, it seems extremely valuable:

> this code is almost 1:1 equivalent with the existing Zephyr work queue (not the p4wq, just the regular one), except that each queue gets only one work item ("task" in SOFese) associated with it, and nothing here cares about the "delayed work" features that complicate the Zephyr implementation.

Duplication has a maintenance cost and debt but _unknown_ duplication costs **much** more.

None of these comments implies that the equivalent Zephyr code is perfect or bug-free. It's only a reminder that it has many more users, developers, reviewers and validation.

marcinszkudlinski:
Guys, IT IS NOT A WORKQUEUE
Yes, currently it just triggers threads at certain times so it may be done using a WQ, even single threaded

BUT It is a base for sophisticated EDF algorithm, which need to calculate component readiness and deadlines after every LL tick. 
In current closed src implementation is even more detailed - recalculation of all deadlines takes place at every single task switch, including return form interrupt handlers etc. 

In future it will go through all pipelines and components, looking which modules should go first to meet the last data output deadline etc. 

The main idea of the EDF scheduling is to have ALL modules in separate threads and recalcuate conditions as often as possible 

So 
 - regular single threaded WQ does not work at all 
 - P4 is better, may be used, yet its pointless because we'll need to have number of threads = number of tasks

For tasks like IPC - we can consider P4 to save some memory. Just keep in mind that all of those tasks must be a "task with budget" - should work on higer prioroty than LL, but be limited to certain cycles budget than must yeld to LL

lyakh:
> * P4 is better, may be used, yet its pointless because we'll need to have number of threads = number of tasks

why would we need one thread per task? Cannot we have one thread per priority? Wouldn't it be enough to handle all tasks with the same priority in a single thread - they don't need to preempt each other

marcinszkudlinski:
> Wouldn't it be enough to handle all tasks with the same priority in a single thread - they don't need to preempt each other

https://en.wikipedia.org/wiki/Earliest_deadline_first_scheduling

_Whenever a scheduling event occurs (task finishes, new task released, etc.) the queue will be searched for the process closest to its deadline. This process is the next to be scheduled for execution._

scheduling event is LL tick or EDF task processing complete. 
Lets consider we do have 
 - chain of LL running every 1ms 
 - DP module 1 running every 10ms with processing time 5ms.  
 - DP module 2 running every 300ms with processing time 100ms. 

after LL TICK_1: assuming (regadless of reasons) that DP1 is not ready, DP2 is ready for processing. 
DP2 gets CPU for rest of the time tick (0.8ms ? whatever)

after LL TICK_2: DP1 is ready now DP2 is processing (still has 99.2ms to go)
**DP2 MUST be preempted and DP1 need to get CPU to meet its deadline - have data ready before its next tick** 
DP1 gets CPU for rest of the time tick (0.8ms)

marcinszkudlinski:
One more detail
When creating a DP module, the module itself requests a stack size for processing. For some modules (dolby atmos, speech recognition etc.) it may be scientifically larger than for simple operations.  
Size of DP processing stack is described in module manifest

    /*! size (in bytes) of stack required by module instance for its DP processing task.
        NOTE: First entry in the main binary loaded by the ASDP ROM requires this field to be set
        to the number of bss pages used by the main module (Base FW module).
    */
    uint16_t    instance_stack_size;

That makes usage of P4 impossible - in P4 any worker may pick up any task for processing

src/schedule/zephyr_dp_schedule.c
marcinszkudlinski: Ok, I can remove complete and all logic regarding calling/not calling complete
it will simplify a lot in fact.

src/schedule/zephyr_dp_schedule.c
marcinszkudlinski: *  SOF_TASK_STATE_COMPLETED  - the task will be removed from scheduling,
     *				calling schedule_task will add the task to processing again
     *				task_complete() will be called
     *  SOF_TASK_STATE_CANCEL     - the task will be removed from scheduling,
     *				calling schedule_task will add the task to processing again
     *				task_complete() won't be called

calling complete will call task_complete() **in the context of task thread**

but looking at #7200 - maybe it should be simpified

marcinszkudlinski:
> translate

marcinszkudlinski:
> Presumably with EDF you wouldn't be waking up all tasks but only the one with the next deadline, or you'd translate deadlines to priorities, but that would be ugly.

All tasks that are ready for processing will be woken up. 
I don't need to translate to priorites - all I should do is to set a deadline in task descriptor and let Zephyr do the rest:

https://docs.zephyrproject.org/3.2.0/kernel/services/scheduling/index.html
_A thread’s relative priority is primarily determined by its static priority. However, when both earliest-deadline-first scheduling is enabled ([CONFIG_SCHED_DEADLINE](https://docs.zephyrproject.org/3.2.0/kconfig.html#CONFIG_SCHED_DEADLINE)) and a choice of threads have equal static priority, then the thread with the earlier deadline is considered to have the higher priority._

https://docs.zephyrproject.org/3.2.0/kernel/services/threads/index.html#c.k_thread_deadline_set

andyross:
FWIW: Zephyr threads have an EDF priority already.  You don't have to do any queue management; just set the value on the thread when it becomes runnable via k_thread_deadline_set().  The p4wq work items have this as a data field.

andyross:
Actually to be stronger: given that this implementation doesn't have the EDF features yet, that the OS already provides it via a largely equivalent tool, I think it seems like maybe the time to do that port is now?

src/schedule/zephyr_dp_schedule.c
marcinszkudlinski: Not quite - I DO want to sleep forever when task is COMPLETED. Time tick won't ever wake it up. The only way to wake the task is to call task_schedule again

the situation I want to **prevent is double call of task_complete** - 
 - task is running (semaphore locked)
 - call to task complete arrive (async) => unlock the semaphore to ensure task_complete is called in task context
 - task_complete is called
 - the loop goes to start - semaphore is unlocked
 **- task_complete is called again :(**
 - task sleeps forever (as designed) on semaphore

ANYWAY - task completed is to be removed. So this prevention won't be needed

marcinszkudlinski: task complete removed, semaphore reset not needed anymore

marcinszkudlinski:
@andyross will Zephyr calulate deadlines according to pipeline structure? 
No
As I said: " all I need to do is to **set a deadline in task descriptor and let Zephyr do the rest**"

aborisovich:
We had to drop the idea of using this scheduler for IPC.
As indicated in the dp scheduler desciption - new thread is created for each SOF task.
Unfortunately we have a interface mismatch since "SOF task" in IPC terms is a single IPC send action (of single message from IPC queue), not the management of the whole queue. So it would be too expensive to create a thread each time we have to send a IPC message to Host. I'll create the RFC on how we would like to proceed with IPC refactor. We would like to use zephyr p4 work queue abstraction just like IDC already does.

src/include/sof/schedule/dp_schedule.h
```diff
@@ -0,0 +1,101 @@
...
+ * NOTE: delayed start and rescheduling takes place in sync with LL scheduler, meaning the
+ *   DP scheduler is triggered as the last task of LL running on a primary core.
+ *   That implies a limitation: LL scheduler MUST be running on primary core in order to have
+ *   this feature working.
+ *   It is fine, because rescheduling is a feature used for data processing when a pipeline is
+ *   running.
+ *
+ * Other possible usage of DP scheduler is to schedule task with DP_SCHEDULER_RUN_TASK_IMMEDIATELY
+ *   as start parameter. It will force the task to work without any delays and async to LL.
+ *   This kind of scheduling may be used for tasks like IPC.
```
aborisovich, line 32: > This kind of scheduling may be used for tasks like IPC.

Please drop this line from description.

aborisovich (changes requested):

src/schedule/zephyr_dp_schedule.c
```diff
@@ -0,0 +1,358 @@
...
+	return 0;
+}
+
+static int scheduler_dp_task_free(void *data, struct task *task)
+{
+	k_spinlock_key_t lock_key;
+	struct task_dp_pdata *pdata = task->priv_data;
+
+	/* abort the execution of the thread */
+	k_thread_abort(pdata->thread_id);
```
andyross, line 125: Thread abort is a little dangerous given the API here, where you have user callbacks that might hold locks or other resources at the point where they're aborted.  Ideally you'd want to do e.g. k_thread_join() here after having flagged a "terminate" command to the thread's main loop somehow.

src/schedule/zephyr_dp_schedule.c
```diff
@@ -0,0 +1,358 @@
...
+				break;
+
+			case SOF_TASK_STATE_CANCEL:
+			case SOF_TASK_STATE_COMPLETED:
+				/* remove from scheduling */
+				list_item_del(&task->list);
+				break;
+
+			default:
+				/* illegal state, serious defect, won't happen */
```
andyross, line 183: If it's a "will not happen" case, might be preferable to use an __ASSERT() here instead.

src/schedule/zephyr_dp_schedule.c
```diff
@@ -0,0 +1,358 @@
...
+	p_stack = rballoc_align(0, SOF_MEM_CAPS_RAM, stack_size, Z_KERNEL_STACK_OBJ_ALIGN);
+	if (!p_stack) {
+		tr_err(&dp_tr, "zephyr_dp_task_init(): stack alloc failed");
+		ret = -ENOMEM;
+		goto err;
+	}
+
+	/* allocate zephyr thread  */
+	p_thread = rzalloc(SOF_MEM_ZONE_SYS_SHARED, 0, SOF_MEM_CAPS_RAM,
+			   sizeof(*p_thread));
```
andyross, line 318: Minor, but you don't need to zero the thread struct.  k_thread_create() below will fully initialize it.

src/schedule/zephyr_dp_schedule.c
```diff
@@ -0,0 +1,358 @@
...
+	thread_id = k_thread_create(p_thread, p_stack, stack_size, dp_thread_fn, task,
+				    NULL, NULL, task_priority, K_USER, K_FOREVER);
+	if (!thread_id)	{
+		ret = -EFAULT;
+		tr_err(&dp_tr, "zephyr_dp_task_init(): zephyr thread create failed");
+		goto err;
+	}
+	/* pin the thread to specific core */
+	ret = k_thread_cpu_pin(thread_id, task->core);
+	if (ret < 0) {
```
andyross, line 335: Ditto, this won't fail unless something specifies a bad CPU or tries to pin a running thread.

src/audio/pipeline/pipeline-schedule.c
```diff
@@ -355,6 +375,49 @@ int pipeline_comp_task_init(struct pipeline *p)
...
+	int ret;
+	struct task_ops ops  = {
+		.run		= dp_task_run,
+		.get_deadline	= NULL,
+		.complete	= NULL
+	};
+
+	if (!comp->task) {
+		component_task = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
+					 sizeof(*component_task));
```
andyross, line 399: Design point: there are a lot of independent heap blocks being allocated all at once and then all freed together.  It wouldn't be too hard to block them all together in a struct or something instead.

andyross (commented):
Sorry, I can by this one again and realized I had some unposted comments.   Cleaned them up and submitting late.  Nothing major.

src/include/sof/audio/component_ext.h
```diff
@@ -147,13 +148,46 @@ static inline int comp_trigger_remote(struct comp_dev *dev, int cmd)
...
+
+	/* start a thread in case of shared component or DP scheduling */
+	if (dev->task) {
+		/* schedule or cancel task, component must be shared or in DP domain */
+		assert(dev->is_shared ||
+		       dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_DP);
```
lyakh, line 161: sorry, don't understand: why does it have to be shared?

lyakh (commented):
Just one comment to clarify

src/include/sof/audio/component_ext.h
marcinszkudlinski: its folllwing (in fact verifying checking) the logic that was here before
Note that IDC on second core will finally call comp_trigger (and comp_trigger_local on second core)

amd the condition was:
return (**dev->is_shared && !cpu_is_me(dev->ipc_config.core))** ? comp_trigger_remote(dev, cmd) : dev->drv->ops.trigger(dev, cmd);

And according to description (before/after)
```
        struct task *task; /**< component's processing task used only
                                   * for components running on different core
                                 * than the rest of the pipeline

```
```
        struct task *task;	/**< component's processing task used
			  *  1) for components running on different core
			  *    than the rest of the pipeline
			  *  2) for all DP tasks
			  *
```
may be removed, it is just double check

kv2019i (approved):

aborisovich (approved):

src/schedule/zephyr_dp_schedule.c
marcinszkudlinski: anyway, worth to check

src/schedule/zephyr_dp_schedule.c
marcinszkudlinski: I just changed from assert ;)

src/schedule/zephyr_dp_schedule.c
marcinszkudlinski: there's **scheduler_dp_task_cancel** for this - it will remove the thread from EDF scheduling and let it stop on main semaphore

scheduler_dp_task_free is a unconditional, async remove, it frees all the resources. We MUST be sure the thread is dead for good at this point. So -> abort is a good way to do it

andyross (approved):
Out of nitpicks, the zephyr-side usage here all seems fine.

src/include/sof/audio/component_ext.h
lyakh: @marcinszkudlinski correct me if I'm wrong: this function is entered from `comp_trigger()` under the condition `if (!dev->is_shared || cpu_is_me(dev->ipc_config.core))` i.e. one of the conditions when this function is entered is a normal local component trigger with the LL scheduler, which now will trigger this your assert. It really looks both unneeded and wrong to me.

src/include/sof/audio/component_ext.h
```diff
@@ -147,13 +148,46 @@ static inline int comp_trigger_remote(struct comp_dev *dev, int cmd)
...
 	assert(dev->drv->ops.trigger);
 
-	return (dev->is_shared && !cpu_is_me(dev->ipc_config.core)) ?
-		comp_trigger_remote(dev, cmd) : dev->drv->ops.trigger(dev, cmd);
+	if (dev->is_shared && !cpu_is_me(dev->ipc_config.core))
+		ret = comp_trigger_remote(dev, cmd);
+	else
+		ret = comp_trigger_local(dev, cmd);
+
+	return ret;
```
lyakh, line 190: btw, there's no reason not to keep the above pattern and just change it to
```
	return (dev->is_shared && !cpu_is_me(dev->ipc_config.core)) ?
		comp_trigger_remote(dev, cmd) : comp_trigger_local(dev, cmd);
```

marcinszkudlinski: I did it on purpose
It WAY more clear to read and WAY more easy to debug step by step or when setting a breakpoint.

And in result the machine code is 1:1 same.
Couple of more bytes in git repo does not bother me at all

src/include/sof/audio/component_ext.h
marcinszkudlinski: the assert won't be met **only** in case there's **a task** assigned to a component that is
 - not DP **AND** is running on the same core that the rest of pipeline

it won't be hit for any component that have no task assigned ( task == NULL) (line 158 above)

It is following the description of *task field I stated above, is the description incorrect?
Agreed it is jus ta double check, I can remove it, no problem

lyakh (approved):

src/include/sof/audio/component_ext.h
lyakh: ah, now I understand, I completely missed the point with `dev->task != NULL`. So in fast - yes, the indication, that we can only land here either with a shared component or with the DP scheduler would be useful here - either as a comment, or an `if` with a `comp_err(); return;` or as an assertion, but not a blocker of course.

marcinszkudlinski:
Cosmetic changes to satisfy checkpatch

lgirdwood (approved):

lgirdwood:
@keqiaozhang can you check the CI queue, both cavs and ace tests have run, but the parent? job is still showing queuing. Thanks.

marc-hb:
This was merged with clear failure: https://github.com/thesofproject/sof/actions/runs/4413659612/jobs/7734391466

```
/zep_workspace/sof/src/schedule/zephyr_dp_schedule.c:134:20: warning: incorrect type in argument 1 (different address spaces)
/zep_workspace/sof/src/schedule/zephyr_dp_schedule.c:348:36: warning: incorrect type in assignment (different address spaces)
```
cc: @lyakh

keqiaozhang:
> @keqiaozhang can you check the CI queue, both cavs and ace tests have run, but the parent? job is still showing queuing. Thanks.

Fixed. Please check it in new PRs.

marc-hb:
"(different address spaces)" fix submitted by @aiChaoSONG in #7286. thanks!

PR #7095, opened 2023-02-14
lib: fix compiler macro for vec_mem*

cujomalainey:
using incorrect names for compiler, missing ISA header for HiFi check, also should add inlines if we don't have the HiFi implementations

### DIFF
--- a/src/lib/lib.c
+++ b/src/lib/lib.c
@@ -79,7 +79,7 @@ int memset_s(void *dest, size_t dest_size, int data, size_t count)
 	return arch_memset_s(dest, dest_size, data, count);
 }
 
-#if !__XCC || !XCHAL_HAVE_HIFI3 || !CONFIG_LIBRARY
+#if !__XCC__ || !XCHAL_HAVE_HIFI3 || CONFIG_LIBRARY
 void *__vec_memcpy(void *dst, const void *src, size_t len)
 {
 	return memcpy(dst, src, len);
--- a/src/arch/xtensa/include/arch/compiler_info.h
+++ b/src/arch/xtensa/include/arch/compiler_info.h
@@ -15,6 +15,7 @@
 #define __ARCH_COMPILER_INFO_H__
 
 #include <xtensa/hal.h>
+#include <xtensa/config/core-isa.h>
 
 /* read used compilator name and version */
 /* CC_NAME must consist of 3 characters with null termination */


cujomalainey:
Bleh this doesn't work for host builds as core-isa is not available. Should we fork this into a arch specific file?

src/lib/lib.c
```diff
@@ -7,6 +7,7 @@
 #include <sof/common.h>
 #include <rtos/string.h>
 #include <sof/compiler_info.h>
+#include <xtensa/config/core-isa.h>
```
lgirdwood, line 10: Since this is arch specific include in a generic file, can it be put behind an ifdef for xtos xtnesa configuration ?

kv2019i: @cujomalainey Hmm, this should come via compiler_info.h . E.g. in XTOS build this is pulled in for xtensa in
src/arch/xtensa/include/arch/compiler_info.h

In Zephyr, "__XCC__" is passed directly via DEFINES , so we don't need to pull core-isa.h

cujomalainey: @kv2019i I traced the headers and didnt find the include core-isa in there, but it seems to work without the direct include. Hmm weird

cujomalainey:
```
. /usr/local/google/home/cujomalainey/sof/src/include/sof/common.h                                                                                       
.. /usr/local/google/home/cujomalainey/sof/src/include/sof/debug/panic.h
... /usr/local/google/home/cujomalainey/sof/src/arch/xtensa/include/arch/debug/panic.h
.... /usr/local/google/home/cujomalainey/sof/xtos/include/rtos/cache.h                                                                                         
..... /usr/local/google/home/cujomalainey/sof/src/arch/xtensa/include/arch/lib/cache.h                                                                         
...... /usr/local/google/home/cujomalainey/xtensa/XtDevTools/install/builds/RG-2017.8-linux/cavs2x_LX6HiFi3_2017_8/xtensa-elf/arch/include/xtensa/config/core-isa.h
```

Lol I don't think that is the include path we intended

src/lib/lib.c
```diff
@@ -79,7 +79,7 @@ int memset_s(void *dest, size_t dest_size, int data, size_t count)
 	return arch_memset_s(dest, dest_size, data, count);
 }
 
-#if !__XCC || !XCHAL_HAVE_HIFI3 || !CONFIG_LIBRARY
+#if !__XCC__ || !XCHAL_HAVE_HIFI3 || CONFIG_LIBRARY
 void *__vec_memcpy(void *dst, const void *src, size_t len)
```
andyross, line 83: Probably easiest just to declare this weak, no?  That allows this fallback to be declared unconditionally without needing to know a-priori which targets define it.  Doesn't look like SOF has a standard token for that, though Zephyr will.  So something like:

```c
// Maybe put this in a SOF header somewhere...
#ifndef __ZEPHYR__
#define __weak __attribute__((__weak__))
#endif

__weak void *__vec_memcpy(void *dst, const void *src, size_t len)
{
    return memcpy(dst, src, len);
}
```

andyross (approved):
Nitpick on robust handling of fallback definitions.  Logic change to the macros looks obviously corret

kv2019i (approved):

src/lib/lib.c
kv2019i: @cujomalainey Heh, I admit that somewhat suprising include path, so indeed better to have this explicitly.

kv2019i:
The qemu-boot for icl/cnl and mtl-sparse are known issues, others passs, so proceeding with merge.

src/lib/lib.c
cujomalainey: I will try this out, but its dependent on xcc actually handling it properly.

PR #7097, opened 2023-02-15
topology2: cavs-rt5682: Update BT offload SSP for MTL chromebook

udaymb:
For MTL chromebook, SSP 1 is used for BT offload with cavs-rt5682 configuration


kv2019i (approved):
Thanks @udaymb !

juimonen (approved):

lgirdwood (approved):

PR #7099, opened 2023-02-15
Fix pipeline source/sink component discovery and smart_amp_test 

ujfalusi:
…omponent

It's not enough to look at the first component in the module's buffer list. Check all of them to make sure that they belong to different pipelines.

### DIFF
--- a/src/ipc/ipc-common.c
+++ b/src/ipc/ipc-common.c
@@ -106,7 +106,7 @@ struct ipc_comp_dev *ipc_get_ppl_comp(struct ipc *ipc, uint32_t pipeline_id, int
 	struct ipc_comp_dev *icd;
 	struct comp_buffer *buffer;
 	struct comp_dev *buff_comp;
-	struct list_item *clist;
+	struct list_item *clist, *blist;
 	struct ipc_comp_dev *next_ppl_icd = NULL;
 
 	list_for_item(clist, &ipc->comp_list) {
@@ -117,17 +117,22 @@ struct ipc_comp_dev *ipc_get_ppl_comp(struct ipc *ipc, uint32_t pipeline_id, int
 		/* first try to find the module in the pipeline */
 		if (dev_comp_pipe_id(icd->cd) == pipeline_id) {
 			struct list_item *buffer_list = comp_buffer_list(icd->cd, dir);
+			bool last_in_pipeline = true;
 
 			/* The component has no buffer in the given direction */
 			if (list_is_empty(buffer_list))
 				return icd;
 
-			/* it's connected pipeline, so find the connected module */
-			buffer = buffer_from_list(buffer_list->next, struct comp_buffer, dir);
-			buff_comp = buffer_get_comp(buffer, dir);
+			/* check all connected modules to see if they are on different pipelines */
+			list_for_item(blist, buffer_list) {
+				buffer = buffer_from_list(blist, struct comp_buffer, dir);
+				buff_comp = buffer_get_comp(buffer, dir);
 
-			/* Next component is placed on another pipeline */
-			if (buff_comp && dev_comp_pipe_id(buff_comp) != pipeline_id)
+				if (buff_comp && dev_comp_pipe_id(buff_comp) == pipeline_id)
+					last_in_pipeline = false;
+			}
+			/* all connected components placed on another pipeline */
+			if (last_in_pipeline)
 				next_ppl_icd = icd;
 		}
 	}


ujfalusi:
#6996 uncovered the issue that if we start aplay first followed by arecord then the smart_amp paths works as expected, however if we start arecord first then aplay, we will have a firmware panic (division by 0).

lgirdwood (approved):

kv2019i (approved):
@RanderWang Has made the same fix as part of https://github.com/thesofproject/sof/pull/6927/commits/ec1aeca427e8aa46e3f77aec9877fbeb9d0979f7  can you @RanderWang check this version?

ujfalusi:
> @RanderWang Has made the same fix as part of [ec1aeca](https://github.com/thesofproject/sof/commit/ec1aeca427e8aa46e3f77aec9877fbeb9d0979f7) can you @RanderWang check this version?

I think @ranj063 tested that and it was not fixing the issue we were seeing with the smart amp tplg.

ujfalusi:
There is a difference in implementation which might be just enough that #6927 is not addressing this correctly (6927 returns when the b_endpoint remains true, so it might skip components in the list to be checked?

aiChaoSONG (approved):

src/ipc/ipc-common.c
```diff
@@ -117,17 +117,22 @@ struct ipc_comp_dev *ipc_get_ppl_comp(struct ipc *ipc, uint32_t pipeline_id, int
 		/* first try to find the module in the pipeline */
 		if (dev_comp_pipe_id(icd->cd) == pipeline_id) {
 			struct list_item *buffer_list = comp_buffer_list(icd->cd, dir);
+			bool last_in_pipeline = true;
```
RanderWang, line 120: it is used to find source or sink, so last_in_pipeline is not very good

RanderWang:
> There is a difference in implementation which might be just enough that #6927 is not addressing this correctly (6927 returns when the b_endpoint remains true, so it might skip components in the list to be checked?

Please share the module graph.  My PR will found the first endpoint, this PR will find all the endpoint but return the last one.
I want to study the graph whether there are two or more endpoint.

ujfalusi:
@RanderWang, it is the tplg from #6996

RanderWang:
> @RanderWang, it is the tplg from #6996
> 
> you need to run arecord on 0:0 first, then start aplay on 0:0 to reproduce the issue (the other way around is working, but if you start aplay, start arecord, kill aplay, start aplay: you will also crash the firmware).

@aiChaoSONG  do you use tplgpaser to dump a png file ? I want to study it.

RanderWang:
> @RanderWang, it is the tplg from #6996
> 
> you need to run arecord on 0:0 first, then start aplay on 0:0 to reproduce the issue (the other way around is working, but if you start aplay, start arecord, kill aplay, start aplay: you will also crash the firmware).

there are two capture streams in AEC. Also arecord first, then next to reproduce it. I don't get the difference

ujfalusi:
@RanderWang: https://sof-ci.01.org/sofpr/PR6996/build3930/devicetest/TGLU_RVP_NOCODEC_IPC4ZPH/sof-tgl-nocodec.png

ranj063:
> > There is a difference in implementation which might be just enough that #6927 is not addressing this correctly (6927 returns when the b_endpoint remains true, so it might skip components in the list to be checked?
> 
> Please share the module graph. My PR will found the first endpoint, this PR will find all the endpoint but return the last one. I want to study the graph whether there are two or more endpoint. In my opinion, only one valid endpoint should be found

@RanderWang with the smart amp tplg, the current solution ends.up declaring the smart amp module as an endpoint and it is wrong.

ranj063:
SOFCI TEST

ranj063:
> …omponent
> 
> It's not enough to look at the first component in the module's buffer list. Check all of them to make sure that they belong to different pipelines.

and oh on the plus side this also fixes #6864

kv2019i (approved):
Looks good. @RanderWang if there are remaining issues, please follow-up with an incremental PR to address them.

RanderWang:
> Looks good. @RanderWang if there are remaining issues, please follow-up with an incremental PR to address them.

NP.  My pr is still waiting for another month review. We need to speed up.

aiChaoSONG:
> @aiChaoSONG do you use tplgpaser to dump a png file ? I want to study it.

https://github.com/thesofproject/sof-test/blob/main/tools/tplgtool2.py

./tplgtool2.py -d graph path-to-topology [-o path-to-png]

maruthi-machani:
@ranj063 
Are these changes verified for topologies with Mixer or Mux? On AMD platform, we are observing the IPC timeout for pipeline complete with these changes for the topology which involves mux component. Passthrough topologies are working fine.

PR #7100, opened 2023-02-15
smart_amp_test: Correct bytes calculations for processing

ujfalusi:
The avail_frames should not be used for other than the feedback buffer. Source/sink should be using the avail_passthrough_frames when calculating for the processing.

If we have feedback buffer then the avail_frames can be different than avail_passthrough_frames which will cause underrun reports.

### DIFF
--- a/src/samples/audio/smart_amp_test.c
+++ b/src/samples/audio/smart_amp_test.c
@@ -664,8 +664,6 @@ static int smart_amp_copy(struct comp_dev *dev)
 		audio_stream_avail_frames(&source_buf->stream,
 					  &sink_buf->stream);
 
-	avail_frames = avail_passthrough_frames;
-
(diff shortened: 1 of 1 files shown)


ujfalusi:
#6996 uncovered the issue that if we start aplay first followed by arecord then the smart_amp paths works as expected, however if we start arecord first then aplay, we will have a firmware panic (division by 0).
This issue is fixed by #7099 but it uncovers a bug in the smart_amp_test module which would result underruns because it is using the wrong *_frames for calculations for processing.

Applying this PR and #7099 will unblock the #6996.

lgirdwood (approved):

kv2019i (approved):

aiChaoSONG (approved):

PR #7101, opened 2023-02-15
Fixes for smart amp test module

ranj063:


fredoh9:
For alsabat test failure, captured wave file shows empty/silence. But I can generate sinewave and find frequency response right. Trying to find why we have empty wave file.

fredoh9:
Too early updated above.
Left channel is fine, 821Hz detected but only right channel is empty/silence.

> Channel 1 - Checking for target frequency 821.00 Hz
> Amplitude: 31127.0; Percentage: [94]
> Detected peak at 821.04 Hz of 41.83 dB
>  Total 42.7 dB from 818.85 to 823.24 Hz
>  PASS: Peak detected at target frequency
> Detected at least 1 signal(s) in total
> 
> Channel 2 - Checking for target frequency 821.00 Hz
> Amplitude: 0.0; Percentage: [0]
> Detected at least 0 signal(s) in total
>

ranj063:
> Left channel is fine, 821Hz detected but only right channel is empty/silence.
> 
> >

Thanks @fredoh9. fixed now.

aiChaoSONG:
@fredoh9 Maybe there is an issue or configuration issue, I remembered that when I was in lab, I only hear front left, and front left.

src/samples/audio/smart_amp_test.c
```diff
@@ -665,6 +617,31 @@ static int smart_amp_copy(struct comp_dev *dev)
...
+					sad->ipc4_cfg.base.audio_fmt.sampling_frequency;
+
+				buffer_release(buffer_c);
+				break;
+			}
+
+			buffer_release(buffer_c);
+		}
+	}
+#endif
```
RanderWang, line 641: one risk that : smart_amp_copy and ipc4 unbind run in different threads. It is possible that playback is using the feedback buffer and capture is stopped and unbind then the buffer is removed. 

The original code use mutex to sync two threads.

RanderWang: Kai suggested that we should add ref count to free buffer to avoid such issue, but now we just free buffer

ranj063: this is a valid point. The only way then would be to check at every copy(). I wonder if that's acceptable

RanderWang: Can we use notifier method ? currently we have
````c
enum notify_id {
	NOTIFIER_ID_CPU_FREQ = 0,		/* struct clock_notify_data * */
	NOTIFIER_ID_SSP_FREQ,			/* struct clock_notify_data * */
	NOTIFIER_ID_KPB_CLIENT_EVT,		/* struct kpb_event_data * */
	NOTIFIER_ID_DMA_DOMAIN_CHANGE,		/* struct dma_chan_data * */
	NOTIFIER_ID_BUFFER_PRODUCE,		/* struct buffer_cb_transact* */
	NOTIFIER_ID_BUFFER_CONSUME,		/* struct buffer_cb_transact* */
	NOTIFIER_ID_BUFFER_FREE,		/* struct buffer_cb_free* */
	NOTIFIER_ID_DMA_COPY,			/* struct dma_cb_data* */
	NOTIFIER_ID_LL_POST_RUN,		/* NULL */
	NOTIFIER_ID_DMA_IRQ,			/* struct dma_chan_data * */
	NOTIFIER_ID_DAI_TRIGGER,		/* struct dai_group * */
	NOTIFIER_ID_COUNT
};
````

Comment (ranj063) [reply src/samples/audio/smart_amp_test.c:641]: I have modified to check for feedback buf in every copy. Its not that bad for the smart amp because it really only has 2 source buffers.

Comment (RanderWang) [reply src/samples/audio/smart_amp_test.c:641]: yes, but still not address the multi-thread risk

Comment (RanderWang) [reply src/samples/audio/smart_amp_test.c:641]:
````c
diff --git a/src/include/sof/lib/notifier.h b/src/include/sof/lib/notifier.h
index 06762b071..39e17cfd4 100644
--- a/src/include/sof/lib/notifier.h
+++ b/src/include/sof/lib/notifier.h
@@ -34,6 +34,8 @@ enum notify_id {
        NOTIFIER_ID_LL_POST_RUN,                /* NULL */
        NOTIFIER_ID_DMA_IRQ,                    /* struct dma_chan_data * */
        NOTIFIER_ID_DAI_TRIGGER,                /* struct dai_group * */
+       NOTIFIER_ID_MODULE_BIND,                /* struct ipc4_module_bind_unbind */
+       NOTIFIER_ID_MODULE_UNBIND,              /* struct ipc4_module_bind_unbind */
        NOTIFIER_ID_COUNT
 };

diff --git a/src/ipc/ipc4/helper.c b/src/ipc/ipc4/helper.c
index 7ebe12449..5c6525af3 100644
--- a/src/ipc/ipc4/helper.c
+++ b/src/ipc/ipc4/helper.c
@@ -16,6 +16,7 @@
 #include <ipc/dai.h>
 #include <sof/ipc/msg.h>
 #include <sof/lib/mailbox.h>
+#include <sof/lib/notifier.h>
 #include <sof/list.h>
 #include <sof/platform.h>
 #include <rtos/wait.h>
@@ -375,6 +376,9 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
        if (ret < 0)
                goto e_src_bind;

+       notifier_event(NULL, NOTIFIER_ID_MODULE_BIND,
+                      NOTIFIER_TARGET_CORE_ALL_MASK, bu, sizeof(*bu));
+
        ret = comp_buffer_connect(sink, sink->ipc_config.core, buffer,
                                  PPL_CONN_DIR_BUFFER_TO_COMP);
        if (ret < 0) {
@@ -386,6 +390,9 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
        if (ret < 0)
                goto e_sink_bind;

+       notifier_event(NULL, NOTIFIER_ID_MODULE_BIND,
+                      NOTIFIER_TARGET_CORE_ALL_MASK, bu, sizeof(*bu));
+
        /* update direction for sink component if it is not set already */
        if (!sink->direction_set && source->direction_set) {
                sink->direction = source->direction;
@@ -407,10 +414,14 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
        return IPC4_SUCCESS;
 err:
        comp_unbind(sink, bu);
+       notifier_event(NULL, NOTIFIER_ID_MODULE_UNBIND,
+                  NOTIFIER_TARGET_CORE_ALL_MASK, bu, sizeof(*bu));
 e_sink_bind:
 e_sink_connect:
        comp_unbind(source, bu);
+       notifier_event(NULL, NOTIFIER_ID_MODULE_UNBIND,
+                  NOTIFIER_TARGET_CORE_ALL_MASK, bu, sizeof(*bu));
 e_sink_bind:
 e_sink_connect:
        comp_unbind(source, bu);
+       notifier_event(NULL, NOTIFIER_ID_MODULE_UNBIND,
+          NOTIFIER_TARGET_CORE_ALL_MASK, bu, sizeof(*bu));
 e_src_bind:
        pipeline_disconnect(source, buffer, PPL_CONN_DIR_COMP_TO_BUFFER);
 free:
@@ -485,6 +496,12 @@ int ipc_comp_disconnect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
        if (sink->pipeline->status < COMP_STATE_ACTIVE)

+       notifier_event(NULL, NOTIFIER_ID_MODULE_UNBIND,
+          NOTIFIER_TARGET_CORE_ALL_MASK, bu, sizeof(*bu));
+       notifier_event(NULL, NOTIFIER_ID_MODULE_UNBIND,
+          NOTIFIER_TARGET_CORE_ALL_MASK, bu, sizeof(*bu));
+
+
        ret = comp_unbind(src, bu);
        ret1 = comp_unbind(sink, bu);
        irq_local_enable(flags);
diff --git a/src/samples/audio/smart_amp_test.c b/src/samples/audio/smart_amp_test.c
index 1b84a393d..ef759c53d 100644
--- a/src/samples/audio/smart_amp_test.c
+++ b/src/samples/audio/smart_amp_test.c
@@ -7,7 +7,8 @@
 #include <sof/compiler_attributes.h>
 #include <sof/samples/audio/smart_amp_test.h>
-#include <sof/trace/trace.h>
+#include <sof/trace/trace.h>   
+#include <sof/lib/notifier.h>
 #include <sof/ipc/msg.h>
 #include <sof/ut.h>
@@ -44,6 +45,51 @@ struct smart_amp_data {
        uint32_t out_channels;
 };

+static void smart_amp_notifier_bind(void *arg, enum notify_id type, void *data)
+{
+       struct comp_dev *dev = arg;
+       struct smart_amp_data *sad = comp_get_drvdata(dev);
+       struct comp_buffer __sparse_cache *buffer_c;
+       struct comp_buffer *source_buffer;
+       struct list_item *blist;
+
+       comp_err(dev, "=========== smart_amp_notifier_bind()");
+
+       /* searching for feedback source buffers */
+       list_for_item(blist, &dev->bsource_list) {
+               source_buffer = container_of(blist, struct comp_buffer, sink_list);
+               buffer_c = buffer_acquire(source_buffer);
+               if (IPC4_SINK_QUEUE_ID(buffer_c->id) == SOF_SMART_AMP_FEEDBACK_QUEUE_ID) {
+                       k_mutex_lock(&sad->mutex, K_FOREVER);
+                       sad->feedback_buf = source_buffer;
+
+                       buffer_c->stream.channels = sad->config.feedback_channels;
+                       buffer_c->stream.rate = sad->config.base.audio_fmt.sampling_frequency;
+                       buffer_release(buffer_c);
+                       k_mutex_unlock(&sad->mutex);
+
+                       break;
+               }
+
+               buffer_release(buffer_c);
+       }
+}
+
+static void smart_amp_notifier_unbind(void *arg, enum notify_id type, void *data)
+{
+       struct comp_dev *dev = arg;
+       struct smart_amp_data *sad = comp_get_drvdata(dev);
+       struct ipc4_module_bind_unbind *bu = data;
+
+       comp_err(dev, "=========== smart_amp_notifier_unbind()");
+
+       if (bu->extension.r.dst_instance_id == SOF_SMART_AMP_FEEDBACK_QUEUE_ID) {
+               k_mutex_lock(&sad->mutex, K_FOREVER);
+               sad->feedback_buf = NULL;
+               k_mutex_unlock(&sad->mutex);
+       }
+}
 static struct comp_dev *smart_amp_new(const struct comp_driver *drv,
                                      const struct comp_ipc_config *config,
                                      const void *spec)
@@ -75,6 +121,12 @@ static struct comp_dev *smart_amp_new(const struct comp_driver *drv,
 #if CONFIG_IPC_MAJOR_4
        k_mutex_init(&sad->mutex);

+       notifier_register(dev, NULL, NOTIFIER_ID_MODULE_BIND,
+                         smart_amp_notifier_bind, 0);
+
+       notifier_register(dev, NULL, NOTIFIER_ID_MODULE_UNBIND,
+                         smart_amp_notifier_unbind, 0);
+
        cfg = spec;
        bs = sizeof(struct sof_smart_amp_config);
 #else

````

Comment (RanderWang) [reply src/samples/audio/smart_amp_test.c:641]: I want to totally remove bind & unbind ops for ipc4 bind msg and use notifier to replace them

Comment (ranj063) [reply src/samples/audio/smart_amp_test.c:641]: > yes, but still not address the multi-thread risk

why not? it's not like an unbind will happen during a copy right?

Comment (ranj063) [reply src/samples/audio/smart_amp_test.c:641]: @RanderWang ok i will close this PR. Why dont you submit your version to remove bind/unbind?

Comment (ranj063) [reply src/samples/audio/smart_amp_test.c:641]: BTW how does this notifier work? lets say a comp bind/unbind happens, which is in the IPC context, will the notifier be scheduled as a new task to be executed after the IPC task is done? what I am asking is if the notifier callback will be guaranteed to execute before the next copy.

Comment (RanderWang) [reply src/samples/audio/smart_amp_test.c:641]: > @RanderWang ok i will close this PR. Why dont you submit your version to remove bind/unbind?

I plan to but I am busy with multi core feature and AEC. It is in my OKR

Comment (RanderWang) [reply src/samples/audio/smart_amp_test.c:641]: > BTW how does this notifier work? lets say a comp bind/unbind happens, which is in the IPC context, will the notifier be scheduled as a new task to be executed after the IPC task is done? what I am asking is if the notifier callback will be guaranteed to execute before the next copy.

Sure. There is also a multi-thread risk. We still need to add ref count to avoid issue.

Comment (lyakh) [reply src/samples/audio/smart_amp_test.c:641]: concerning the multithreading race: this is again largely a question of whether we allow running connected pipelines on different cores. So far we don't. Then, if both these events - copying and unbinding - are happening on the same core, what contexts they're running in and can they preempt each other? IIRC unbinding would be happening in an IPC thread on core 0 or in the IDC thread (p4wq work with Zephyr) on other cores. Pipeline threads have the highest priority, so it's a question of whether they can preempt IPC and IDC contexts

Comment (RanderWang) [reply src/samples/audio/smart_amp_test.c:641]: is it for module adapter ?  You should add some comment in TODO.  We need add refcounting first to avoid multi-thread risk.

Comment (lyakh) [reply src/samples/audio/smart_amp_test.c:641]: > is it for module adapter ? You should add some comment in TODO. We need add refcounting first to avoid multi-thread risk.

@RanderWang could you please describe the scenario that you find problematic and that would need an addition of a reference counter?

Comment (RanderWang) [reply src/samples/audio/smart_amp_test.c:641]: @lyakh  I meant removing bind & unbind ops in smart_test which we use mutex to sync with ipc thread and smart_copy thread.  Smart AMP saves a pointer of feedback buffer which is from another stream.  smart AMP use it for copy and a risk that ipc unbind will delete it since the feedback stream is stopped when it is used by smart AMP. We can't delete it since it is used by smart AMP. With reference counter we can address it.

Comment (RanderWang) [reply src/samples/audio/smart_amp_test.c:641]: sad->feedback_buf is not synchronized with ipc unbind thread. The mutex is not used by ipc thread now and only used by smart thread, so it can't protect feedback_buf now.

File: src/samples/audio/smart_amp_test.c [hunk 2]
```diff
@@ -692,6 +669,9 @@ static int smart_amp_copy(struct comp_dev *dev)
 
 		buffer_release(buf);
 	}
+#if CONFIG_IPC_MAJOR_4
+	sad->feedback_buf = NULL;
```
lyakh, line 666: so on every copy you have to re-assign the feedback buffer... TBH I'm not sure what smart-amp-test is actually for (it's "just" a test, right? So it isn't used in any real audio streaming?) but this seems rather expensive

src/samples/audio/smart_amp_test.c
ranj063: but i assign the feedback buf before processing and unset it right after in every copy. its not optimal but there's no risk with multithreading here

RanderWang: yes, but it is not protected when it is using by copy unless using some mutex to protect feedback buffer to avoiding free or reused.

RanderWang: in original code both smart amp and ipc thread use mutex to access feedback buffer, but now ipc thread doesn't use it

lyakh: @RanderWang could you sketch a sequence of events when you think protection will be required and without one data will be corrupted?

src/samples/audio/smart_amp_test.c
lgirdwood: Yeah, this will need a comment to explain the specific case if needed other wise it could be copy and pasted into other samples.

src/samples/audio/smart_amp_test.c
RanderWang: Two usage of feedback buffer,  smart copy  and ipc bind/unbind message for capture stream.  We need to take care of feedback buffer is being used by smart copy when ipc unbind msg is processed to delete the feedback buffer for capture stream. I will add refer count for buffer to avoid competition

lyakh: @RanderWang sorry, maybe you could do something like https://github.com/thesofproject/sof/pull/7089#issuecomment-1433117586 - just list the sequence of steps / function calls that you think can cause a problem

RanderWang: Two streams : smart amp playback and ssp capture
(1) set up two streams
(2) bind smart_amp.2.1 to copier.ssp.8.1, a feedback buffer will be created to connect two streams
(3) start playback stream one core 0, capture stream on core 1
(4) smart amp playback is copying data from feedback buffer
(5) capture stream is stopped and smart_amp.2.1 is unbind with copier.ssp.8.1, feedback buffer is disconnected and free
     at this time step (4) is still using the feedback buffer

smart

lyakh: @RanderWang running connected pipelines on different cores is still unsupported.

lyakh: ...which also incidentally means, that this is none of smart-amp specific issue, this would be the case with any connected pipelines on different cores - with copier, mixer or anything

RanderWang: @lyakh  The major issue is that unbind will disconnect buffer and free buffer, how about we don't do it for shared buffer case in unbind function ?

lyakh: @RanderWang why is this an issue? What does it break? Running on a single core or on different cores?

gkbldcig:

lgirdwood (approved):
Lots of code removal !

kv2019i:
Good now @RanderWang ?

kv2019i (approved):
Code seems good, but I don't fully understand the conclusion on the long discussion in this PR on the multi-thread case. It is marked as resolved, so I guess this PR is now good to go, but would be good for @RanderWang @lyakh  to ack with a +1 that this is ok to proceed.

RanderWang:
@lgirdwood  @kv2019i.  I also discussed my method (https://github.com/thesofproject/sof/pull/7259) to remove bind & unbind with @ranj063  @lyakh  and got a conclusion that we will do bind & unbind job in module adapter. @andrula-song will adapt smart_amp to module interface and bind & unbind will be removed in her PR. We don't need to feedback buffer check in smart_amp_copy

andrula-song:
> 
hi @RanderWang 
since the there is a limit of  avail_frames = MIN(avail_passthrough_frames, avail_feedback_frames); and the avail_frames  and avail_passthrough_frames could be different(https://github.com/thesofproject/sof/pull/7100), I am afraid we have to check feedback buffer in smart_amp_copy function.

RanderWang:
> > 
> 
> hi @RanderWang since the there is a limit of avail_frames = MIN(avail_passthrough_frames, avail_feedback_frames); and the avail_frames and avail_passthrough_frames could be different(#7100), I am afraid we have to check feedback buffer in smart_amp_copy function.

you can discuss with Ranjani

andrula-song:
> @andrula-song good for you now, if so pls approve or reject
apart of this PR already exist in our code, and I am using module adapter to remove bind & unbind. So I don't think merge this PR is necessary.

lgirdwood:
@ranj063 still needed ?

kv2019i:
Turning into draft for now...

PR #7104, opened 2023-02-15
Restore old chain dma implementation

jsarha:
Just curious if the old version would still pass the whole CI. May also be useful in comparing the logic to find out what goes wrong with the new implementation. But I have no intention of proposing to merge this.


jsarha:
Oh, but I forgot to enable chain dma from the topology.

gkbldcig:

PR #7105, opened 2023-02-16
ipc4: copier: set sink buffer format when sink format is set

RanderWang:
The original code only set sink format in copier but forget to set format of sink buffer. This will result to error when sink components use it.

### DIFF
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -1297,6 +1297,27 @@ static void update_internal_comp(struct comp_dev *parent, struct comp_dev *child
 	child->direction = parent->direction;
 }
 
+static void update_buffer_format(struct comp_buffer __sparse_cache *buf_c,
+				 const struct ipc4_audio_format *fmt)
+{
+	int i;
+
+	buf_c->stream.channels = fmt->channels_count;
+	buf_c->stream.rate = fmt->sampling_frequency;
+	audio_stream_fmt_conversion(fmt->depth,
+				    fmt->valid_bit_depth,
+				    &buf_c->stream.frame_fmt,
+				    &buf_c->stream.valid_sample_fmt,
+				    fmt->s_type);
+
+	buf_c->buffer_fmt = fmt->interleaving_style;
+
+	for (i = 0; i < SOF_IPC_MAX_CHANNELS; i++)
+		buf_c->chmap[i] = (fmt->ch_map >> i * 4) & 0xf;
+
+	buf_c->hw_params_configured = true;
+}
+
 /* configure the DMA params */
 static int copier_params(struct comp_dev *dev, struct sof_ipc_stream_params *params)
 {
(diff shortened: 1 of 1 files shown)


src/audio/copier/copier.c
```diff
@@ -1446,6 +1449,38 @@ static int copier_set_sink_fmt(struct comp_dev *dev, const void *data,
...
+		sink_c = buffer_acquire(sink);
+
+		/* only updated current used sink buffer */
+		sink_id = IPC4_SINK_QUEUE_ID(sink_c->id);
+		if (sink_id != sink_fmt->sink_id) {
+			buffer_release(sink_c);
+			continue;
+		}
+
+		sink_c->stream.channels = cd->out_fmt[sink_id].channels_count;
```
ranj063, line 1466: why not use sink_fmt-> instead of cd->out_fmt[sink_id].?

src/audio/copier/copier.c
```diff
@@ -1446,6 +1449,38 @@ static int copier_set_sink_fmt(struct comp_dev *dev, const void *data,
 							      &sink_fmt->sink_fmt, ipc4_gtw_none,
 							      ipc4_bidirection);
 
+	/* update corresponding sink format */
+	list_for_item(sink_list, &dev->bsink_list) {
```
ranj063, line 1453: looks like this piece of code in already there in copier_params() on line 1360. Can you create a helper and reuse?

RanderWang: updated with a help funciton

src/audio/copier/copier.c
```diff
@@ -1442,6 +1445,24 @@ static int copier_set_sink_fmt(struct comp_dev *dev, const void *data,
...
+		int sink_id;
+
+		sink = container_of(sink_list, struct comp_buffer, source_list);
+		sink_c = buffer_acquire(sink);
+
+		/* only updated current used sink buffer */
```
ranj063, line 1455: nit-pick, this comment is not needed. the first comment to "update corresponding sink format" is enough

src/audio/copier/copier.c
```diff
@@ -1297,6 +1297,27 @@ static void update_internal_comp(struct comp_dev *parent, struct comp_dev *child
 	child->direction = parent->direction;
 }
 
+static void update_buffer_format(struct comp_buffer __sparse_cache *buf_c,
+				 const struct ipc4_audio_format *in_fmt)
```
ranj063, line 1301: another nit-pick, can we call this arg just fmt? in one case its input format and in another its output, so better to just keep this simple.

src/audio/copier/copier.c
```diff
@@ -1297,6 +1297,27 @@ static void update_internal_comp(struct comp_dev *parent, struct comp_dev *child
 	child->direction = parent->direction;
 }
 
+static void update_buffer_format(struct comp_buffer __sparse_cache *buf_c,
```
ranj063, line 1300: and you have a typo in the commit message. Maybe its better to say "Use helper function to update buffer format and avoid code duplication"

src/audio/copier/copier.c
RanderWang: sure, thanks!

src/audio/copier/copier.c
RanderWang: sure, thanks

src/audio/copier/copier.c
RanderWang: thanks

ranj063 (approved):

lgirdwood (approved):

PR #7107, opened 2023-02-16
xtensa-build-zephyr.py: apply debug_overlay after ipc4_overlay

aiChaoSONG:
The config overlays are applied in the order we specify them, the debug_overlay contains the debug configurations, and should be applied last to make sure those debug options take effect.

`CONFIG_SOF_LOG_LEVEL_INF=y` is defined in ipc4_overlay.conf, I personally in my branch added `CONFIG_SOF_LOG_LEVEL_DBG=y`, but no effect due to the order we apply ipc4_overlay.conf and debug_overlay.conf.  I think finally `CONFIG_SOF_LOG_LEVEL_DBG=y` will be added in debug_overlay.conf once mtrace is ready.

### DIFF
--- a/scripts/xtensa-build-zephyr.py
+++ b/scripts/xtensa-build-zephyr.py
@@ -590,10 +590,6 @@ def build_platforms():
 			build_cmd += args.cmake_args
 
 		overlays = [str(item.resolve(True)) for item in args.overlay]
-		# The '-d' option is a shortcut for '-o path_to_debug_overlay', we are good
-		# if both are provided, because it's no harm to merge the same overlay twice.
-		if args.debug:
-			overlays.append(str(pathlib.Path(SOF_TOP, "app", "debug_overlay.conf")))
 
 		# The '-i IPC4' is a shortcut for '-o path_to_ipc4_overlay' (and more), we
 		# are good if both are provided, because it's no harm to merge the same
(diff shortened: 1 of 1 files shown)


marc-hb (changes requested):
> CONFIG_SOF_LOG_LEVEL_INF=y is defined in ipc4_overlay.conf, 

While I understand the connection between CONFIG_LOG_BACKEND_ADSP_MTRACE=y / CONFIG_ZEPHYR_LOG=y and IPC4, the default log _level_ has nothing to do with IPC4 and it shouldn't be in this file. Can you please (re)move it?

The solution to this problem is to not have conflicting values in overlays that seem compatible with each other. We should play ordering games at all; they're very time-consuming to debug.

aiChaoSONG:
@marc-hb I will remove `CONFIG_SOF_LOG_LEVEL_INF=y` (as it is default value) from ipc4_overlay.conf for tgl and tgl-h, and close this PR, is this good for you?

marc-hb:
>  I will remove CONFIG_SOF_LOG_LEVEL_INF=y (as it is default value) from ipc4_overlay.conf for tgl and tgl-h

If it works then it sounds good to me. The question is: what will be the default level when not using the debug overlay?

aiChaoSONG:
> If it works then it sounds good to me. The question is: what will be the default level when not using the debug overlay?

@marc-hb  info is the default level

Issue #7081, opened 2023-02-13
[BUG] Large (multiple parts) blobs might be handled incorrectly


paulstelian97:
**Describe the bug**

> A clear and concise description of what the bug is.

Whenever attempting to switch out the FIR EQ configuration at runtime (while playing), an incorrect error message (Filter length 252 exceeds limitation of build) shows up in the logs and playback stops. On the i.MX platform it also leads to a DSP hang (possibly due to unrelated improper handling of XRUN conditions). The firmware logs were taken via etrace (because of the hang)

> What have you tried to diagnose or workaround this issue?
Please also read https://thesofproject.github.io/latest/contribute/process/bug-tracking.html for further information on submitting bugs.

I have attempted to change the configuration `COMP_LEGACY_INTERFACE` to `y` in order to attempt to bypass the Module Adapter (the logs are with this configuration set to `y`, which tells me I didn't actually manage to remove it from the causes).

**To Reproduce**
> Steps to reproduce the behavior: (e.g. list commands or actions used to reproduce the bug)

In one terminal, run

    aplay -Dhw:1,0 -t raw -f S32_LE -c 2 -r 48000 /dev/urandom

In another terminal, run various sof-ctl commands with the flat/mid/pass (small, single part blobs) and loudness (multi-part blob, which is the problem). After the execution of the multi-part blob, which succeeds, the next one leads to the incorrect message and XRUN (combined with DSP hang on i.MX).

    # These three commands succeed no matter how much you repeat or randomize them
    sof-ctl -Dhw:1 -n 44 -s /unit_tests/sof/tools/ctl/eq_fir_flat.txt
    sof-ctl -Dhw:1 -n 44 -s /unit_tests/sof/tools/ctl/eq_fir_mid.txt
    sof-ctl -Dhw:1 -n 44 -s /unit_tests/sof/tools/ctl/eq_fir_pass.txt
    # This command succeeds once, but breaks something internally
    sof-ctl -Dhw:1 -n 44 -s /unit_tests/sof/tools/ctl/eq_fir_loudness.txt
    # This command fails, no matter which of the four available files is given
    sof-ctl -Dhw:1 -n 44 -s /unit_tests/sof/tools/ctl/eq_fir_flat.txt

**Reproduction Rate**
> How often does the issue happen ? i.e. 1/10 (once in ten attempts), 1/1000 or all the time.
Does the reproduction rate vary with any other configuration or user action, if so please describe and show the new reproduction rate.

[...]

PR #7109, opened 2023-02-16
sof: audio: eq_fir: Fix GCC-specific check

paulstelian97:
Currently, the check to prevent too large filters from being loaded on GCC builds of SOF does not work correctly. This commit makes sure the error occurs immediately, instead of at the next coefficients change.

### DIFF
--- a/src/audio/eq_fir/eq_fir.c
+++ b/src/audio/eq_fir/eq_fir.c
@@ -313,7 +313,7 @@ static int eq_fir_init_coef(struct comp_dev *dev, struct sof_eq_fir_config *conf
 		}
 
 #if defined FIR_MAX_LENGTH_BUILD_SPECIFIC
-		if (fir[i].taps * nch > FIR_MAX_LENGTH_BUILD_SPECIFIC) {
+		if (eq->length * nch > FIR_MAX_LENGTH_BUILD_SPECIFIC) {
 			comp_err(dev, "Filter length %d exceeds limitation for build.",
 				 fir[i].taps);
 			return -EINVAL;


paulstelian97:
@dbaluta @singalsu this is the smallest possible change that "fixes" issue #7081 . The error shows up immediately as you load the big blob as opposed to one attempt later.

A change that I'm thinking about, which will be more complex, is to verify the new configuration inside set_config function and reject invalid configurations there.

paulstelian97:
Changed: the check is proper and doesn't break the original commit's intent either. This can be merged as-is, but I am working on better handling of these issues that builds upon this.

src/audio/eq_fir/eq_fir.c
```diff
@@ -313,7 +313,7 @@ static int eq_fir_init_coef(struct comp_dev *dev, struct sof_eq_fir_config *conf
 		}
 
 #if defined FIR_MAX_LENGTH_BUILD_SPECIFIC
-		if (fir[i].taps * nch > FIR_MAX_LENGTH_BUILD_SPECIFIC) {
+		if (eq->length * nch > FIR_MAX_LENGTH_BUILD_SPECIFIC) {
 			comp_err(dev, "Filter length %d exceeds limitation for build.",
 				 fir[i].taps);
```
dbaluta, line 318: I agree with this change. This is indeed an improvement although it doesn't fix the problem on i.MX.

@singalsu  I think we can merge it.

@paulstelian97 Please enhance the commit message by specifying the commit subject + hash that did the reording.

singalsu (approved):
Thanks, yes there's a mistake with order of things and this is the correct check.

dbaluta (approved):

lgirdwood (approved):

PR #7102, opened 2023-02-15
[DNM] west.yml: upgrade Zephyr to 4d4e25292faa

kv2019i:
Total of 1116 commits, including following related to intel_adsp/sparse/dmic/xtensa:

8f5bcb2e76c3 intel_adsp: ace: fix linker script for xcc-clang compiler
18ce85c20130 tests: intel_adsp: ssp: fix dma data sizes
60a20471b561 intel_adsp: ace: enable interrupts for secondary core
6045eed2f361 intel_adsp: ace: enable core power gating
e1dbc2efef50 intel_adsp: ace: add core power off step
a99b073392fc intel_adsp: ace: d3 exit update
156c7cd21759 intel_adsp: bbzero/bmemcpy with picolibc fix
60196ca1126a cmake: sparse: deprecate old sparse support
91902c5fd4db cmake: add sparse support to the new SCA infrastructure
a684714d5c82 soc: intel_adsp: Correct HDA parameter docstrings
db495a5ebee3 xtensa: stop execution under simulator for double exception
8ff88346955b xtensa: sparse: fix address space mismatch
7965fd2b4a8a samples/boards/intel_adsp: Make sample work with twister
422250d3b183 mm: intel_adsp_mtl_tlb: suppress sparse address space warnings
618a478ded70 xtensa: fix sparse warning when converting to uncache pointer
8b391dc43841 drivers: audio: dmic_nrfx_pdm: Fix a race condition in the driver
8794de2934f7 intel_adsp: soc: ace: Add communication widget driver
a9b3d935500c intel_adsp: dai: Add support for ALH up to 16 nodes
837432506269 drivers: dai: intel: dmic: don't use assert for error handling

PR #7110, opened 2023-02-16
dai-zephyr: fixes to dma_stop() usage

kv2019i:
To fix SOF SSP tests with latest Zephyr main, see failures without this fix in #7102

For all tests to pass, https://github.com/zephyrproject-rtos/zephyr/pull/54900 is also needed.

### DIFF
--- a/src/audio/dai-zephyr.c
+++ b/src/audio/dai-zephyr.c
@@ -1014,6 +1014,7 @@ static void dai_update_start_position(struct comp_dev *dev)
 static int dai_comp_trigger_internal(struct comp_dev *dev, int cmd)
 {
 	struct dai_data *dd = comp_get_drvdata(dev);
+	int prev_state = dev->state;
 	int ret;
 
 	comp_dbg(dev, "dai_comp_trigger_internal(), command = %u", cmd);
@@ -1101,7 +1102,7 @@ static int dai_comp_trigger_internal(struct comp_dev *dev, int cmd)
 		dai_trigger_op(dd->dai, cmd, dev->direction);
 #else
 		dai_trigger_op(dd->dai, cmd, dev->direction);
-		if (dev->state == COMP_STATE_ACTIVE) {
+		if (prev_state == COMP_STATE_ACTIVE) {
 			ret = dma_stop(dd->chan->dma->z_dev, dd->chan->index);
 		} else {
 			comp_warn(dev, "dma was stopped earlier");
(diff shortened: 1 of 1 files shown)


kv2019i:
Something wrong with the changes and/or these cannot be merged until we update Zephyr. Marking as DNM until I have rootcaused the problem.

kv2019i:
V2:
- one bug in the second patch related to RELEASE trigger
- this version tested with latest Zephyr in https://github.com/thesofproject/sof/pull/7111

lyakh:
> To fix SOF SSP tests with latest Zephyr main, see failures without this fix in #7102
> 
> For all tests to pass, [zephyrproject-rtos/zephyr#54900](https://github.com/zephyrproject-rtos/zephyr/pull/54900) is also needed.

arrrgh, looks like we cannot see failures in #7102 any more :-( It would've been good to have a direct link to those failures here... Do we have those failures anywhere else?

paulstelian97 (approved):

kv2019i:
@lyakh wrote:
> arrrgh, looks like we cannot see failures in #7102 any more :-( It would've been good to have a direct link to those failures here... Do we have those failures anywhere else?

Sorry for non-Intel folks, the log is there in public but I know of no easy to find the link given the PR is now closed and I did not add the link  manually to the bug comments. I do nowadays almost always add a link (just for this reason), but we are human and make mistakes. @lyakh you can see the log in Intel internal database for test plan 21013

juimonen (commented):
me wondering do we need this prev_state at all... I understood that dma should handle now "double" stops and starts....

src/audio/dai-zephyr.c
```diff
@@ -1111,7 +1112,7 @@ static int dai_comp_trigger_internal(struct comp_dev *dev, int cmd)
 		break;
 	case COMP_TRIGGER_PAUSE:
 		comp_dbg(dev, "dai_comp_trigger_internal(), PAUSE");
-		if (dev->state == COMP_STATE_ACTIVE) {
+		if (prev_state == COMP_STATE_ACTIVE) {
```
kv2019i, line 1114: @juimonen Yup, I originally had three places to replace with "prev_state", but one change was not correct. I removed the first check in this PR already., so that only leaves this one check before dma_suspend(). Reading the Zephyr dma.h docs again, it seems also dma_suspend() is safe to call (even if channel is not active. So probably we can make the change and simplify this series.

tmleman:
I think that now when it's safe from power management perspective to call multiple times DMA start and stop, we can just revert this commit https://github.com/thesofproject/sof/commit/fa0456e5a8996dcdfee97f1e7142891cc3dd90ae

tmleman (approved):

kv2019i:
This needs to be merged together with Zephyr baseline update. Replaced with https://github.com/thesofproject/sof/pull/7125

PR #7112, opened 2023-02-16
Tools: Topology2: Change PGA widget curve_duration to 20 ms

singalsu:
The previous value 10 is only 0.1 us that disables the gain ramp totally and causes harsh sounding volume transitions.

### DIFF
--- a/tools/topology/topology2/include/components/gain.conf
+++ b/tools/topology/topology2/include/components/gain.conf
@@ -167,7 +167,7 @@ Class.Widget."gain" {
 	period_sink_count	2
 	period_source_count	2
 	curve_type		"fade"
-	curve_duration		10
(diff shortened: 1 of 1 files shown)


singalsu:
Note: This exposes a bug in gain ramp handling so this should not be merged yet.

RanderWang:
@singalsu  why not submit the patch without draft lable ?   I planed to but forget.

singalsu:
> @singalsu why not submit the patch without draft lable ? I planed to but forget.

Thanks, now non-draft. This exposes a current bug but this change needs to be done to avoid the volume jumps.

RanderWang (approved):

lgirdwood (approved):

PR #7113, opened 2023-02-16
Use arrays for objects in top-level conf files and included conf files

ranj063:
This looks like a lot of changes for one patch but in reality the changes are quite monotonous to use arrays in object declarations. This fixes the problem of the topology writer having to come up with obscenely large instance ID's for objects just to make them exclusive when including conf files programatically.


ranj063:
> This looks like a lot of changes for one patch but in reality the changes are quite monotonous to use arrays in object declarations. This makes the problem os the topology writer having to come up with obscenely large instance ID's for objects just to make them exclusive when including confi files programatically.

I have verified that the generated binaries with this change are identical to the ones we have today except for these three files and I think these are because of some module names being different due to the change in instance IDs.
sof-tgl-rt711-rt1308-4ch.tplg
sof-tgl-rt711-rt1316-rt714.tplg
sof-tgl-rt715-rt711-rt1308-mono.tplg

kv2019i (approved):
Seems much cleaner without the additional ids (that needed to be unique), thanks!

tools/topology/topology2/cavs-nocodec-bt.conf
```diff
@@ -113,130 +113,134 @@ Object.Dai {
...
-	passthrough-playback.1 {
-		index	1
-		Object.Widget.pipeline.1 {
-			stream_name 'NoCodec-0'
-		}
-		Object.Widget.copier.1 {
-			stream_name 'SSP0 Playback'
-		}
-	}
+	passthrough-be [
```
jsarha, line 116: The changes to these new pipeline object instantiations with number of pipeline arrays were hardest to follow. The new jumping indexes do not look the most logical choices, but then again I can see that they come from history. In the end the indexes do not matter for anything else but they are needed when referring to these objects from outside, do they?

jsarha (approved):
All in all I think I understand the logic behind these changes, and yes the result look much cleaner.

tools/topology/topology2/cavs-nocodec-bt.conf
ranj063: yes, the indices have no meaning in most cases except for the widget names. But I havent touched those obects within pipeline class definitions yet. That will be the next target after this one is merged.

bardliao (approved):

kv2019i:
qemu-boot for icl/cnl fail and the mtl-zephyr-sparse are known issues, other checks pass so proceeding with merge.

PR #7124, opened 2023-02-20
Audio: Fix the pointer update errors while volume ramp enabled

andrula-song:
while enable volume ramp, we may need to run the while loop in volume_process function several times, but we only update the read/write pointers of in/output streams in module_adapter copy function after the whole volume_process function finished. Then we would always get and store data from wrong address and lead to discontinuity waveform.

### DIFF
--- a/src/audio/module_adapter/module/volume/volume_generic.c
+++ b/src/audio/module_adapter/module/volume/volume_generic.c
@@ -68,8 +68,8 @@ static void vol_s24_to_s24(struct processing_module *mod, struct input_stream_bu
 	const int nch = source->channels;
 	int remaining_samples = frames * nch;
 
-	x = source->r_ptr;
-	y = sink->w_ptr;
+	x = audio_stream_wrap(source, (char *)source->r_ptr + bsource->consumed);
+	y = audio_stream_wrap(sink, (char *)sink->w_ptr + bsink->size);
 
 	bsource->consumed += VOL_S32_SAMPLES_TO_BYTES(remaining_samples);
 	bsink->size += VOL_S32_SAMPLES_TO_BYTES(remaining_samples);
@@ -117,8 +117,8 @@ static void vol_s32_to_s32(struct processing_module *mod, struct input_stream_bu
 	const int nch = source->channels;
 	int remaining_samples = frames * nch;
 
-	x = source->r_ptr;
-	y = sink->w_ptr;
+	x = audio_stream_wrap(source, (char *)source->r_ptr + bsource->consumed);
+	y = audio_stream_wrap(sink, (char *)sink->w_ptr + bsink->size);
 	bsource->consumed += VOL_S32_SAMPLES_TO_BYTES(remaining_samples);
 	bsink->size += VOL_S32_SAMPLES_TO_BYTES(remaining_samples);
 	while (remaining_samples) {
@@ -169,8 +169,8 @@ static void vol_s16_to_s16(struct processing_module *mod, struct input_stream_bu
 	const int nch = source->channels;
 	int remaining_samples = frames * nch;
 
-	x = source->r_ptr;
-	y = sink->w_ptr;
+	x = audio_stream_wrap(source, (char *)source->r_ptr + bsource->consumed);
+	y = audio_stream_wrap(sink, (char *)sink->w_ptr + bsink->size);
 	bsource->consumed += VOL_S16_SAMPLES_TO_BYTES(remaining_samples);
 	bsink->size += VOL_S16_SAMPLES_TO_BYTES(remaining_samples);
 	while (remaining_samples) {
--- a/src/audio/module_adapter/module/volume/volume_generic_with_peakvol.c
+++ b/src/audio/module_adapter/module/volume/volume_generic_with_peakvol.c
@@ -66,8 +66,8 @@ static void vol_s24_to_s24(struct processing_module *mod, struct input_stream_bu
 	int32_t tmp;
 
 	memset(cd->peak_regs.peak_meter, 0, sizeof(uint32_t) * cd->channels);
-	x = source->r_ptr;
-	y = sink->w_ptr;
+	x = audio_stream_wrap(source, (char *)source->r_ptr + bsource->consumed);
+	y = audio_stream_wrap(sink, (char *)sink->w_ptr + bsink->size);
 
 	bsource->consumed += VOL_S32_SAMPLES_TO_BYTES(remaining_samples);
 	bsink->size += VOL_S32_SAMPLES_TO_BYTES(remaining_samples);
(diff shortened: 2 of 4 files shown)


andrula-song:
if we don't fix this bug, the fade in pattern would be like this
7_TestPeakVolBat test_01_09_peakvol_bat_scope 16000Hz_16in16bit_1ch-0 1 _2023_02_09_03_21_41_curve_part_channel_1
never update the read/write pointer of in/output stream in volume_process function, and each while loop always fetch data from the same address

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -523,7 +526,7 @@ static inline void init_ramp(struct vol_data *cd, uint32_t curve_duration, uint3
 	/* In IPC4 driver sends curve_duration in hundred of ns - it should be
 	 * converted into ms value required by firmware
 	 */
-	cd->initial_ramp = Q_MULTSR_32X32((int64_t)curve_duration,
+	cd->initial_ramp = Q_MULTSR_32X32((int64_t)curve_duration ,
```
lyakh, line 529: spurious change

lyakh (commented):
Sorry, could you explain a bit more? Why aren't calculations like
```
	bsource->consumed += VOL_S32_SAMPLES_TO_BYTES(samples);
	bsink->size += VOL_S32_SAMPLES_TO_BYTES(samples);
```
in all scaling functions doing the right thing? They seem to be accumulating scaled data so the sum should be the same? You could compare in `module_adapter_simple_copy()` to check if and why consumed and produced byte counts are different?

lgirdwood:
@singalsu pls review.

src/audio/module_adapter/module_adapter.c
```diff
@@ -736,7 +736,8 @@ static int module_adapter_simple_copy(struct comp_dev *dev)
 		src_c = attr_container_of(mod->input_buffers[i].data,
 					  struct comp_buffer __sparse_cache,
 					  stream, __sparse_cache);
-		comp_update_buffer_consume(src_c, mod->input_buffers[i].consumed);
+		if (!mod->skip_src_read_pointer_update)
```
lgirdwood, line 739: can you inline comment here and below on what condition would activate the update ?

src/audio/module_adapter/module_adapter.c
```diff
@@ -758,7 +759,8 @@ static int module_adapter_simple_copy(struct comp_dev *dev)
 
 		if (!mod->skip_sink_buffer_writeback)
 			buffer_stream_writeback(sink_c, mod->output_buffers[i].size);
-		comp_update_buffer_produce(sink_c, mod->output_buffers[i].size);
+		if (!mod->skip_dst_write_pointer_update)
```
lgirdwood, line 762: ditto - inline comment.

lgirdwood (commented):
@andrula-song whats the MCPS impact ?

src/audio/module_adapter/module_adapter.c
```diff
@@ -736,7 +736,8 @@ static int module_adapter_simple_copy(struct comp_dev *dev)
 		src_c = attr_container_of(mod->input_buffers[i].data,
 					  struct comp_buffer __sparse_cache,
 					  stream, __sparse_cache);
-		comp_update_buffer_consume(src_c, mod->input_buffers[i].consumed);
+		if (!mod->skip_src_read_pointer_update)
+			comp_update_buffer_consume(src_c, mod->input_buffers[i].consumed);
```
singalsu, line 740: Is volume component not updating consumed correctly or not producing output_buffers[i].size when executing the ramp? Could this be fixed there instead?

andrula-song: > Is volume component not updating consumed correctly or not producing output_buffers[i].size when executing the ramp? 

In our volume_process function, we just calculate the "size", but never update the read/write pointer of in/output stream. while executing ramp, we will execute the while loop several times, for example, each volume_process would process 48 frames, and each while loop( for ramp) process 6 frames, if we do not update the pointer in that while loop, we will get wrong data.

>Could this be fixed there instead?
I used to work around (record the initial pointers before while loop, and assign them to read/write pointers after while loop) and it works. I can figure out which performance better and use the best one?

lyakh: @andrula-song The description of this PR says: "we may need to run the while loop in volume_process function several times" whereas my understanding is that those loops shouldn't be run more than twice: normally they would run only once as long as there's no wrapping, and if data wrapped, the loop will run a second time to process the remaining part at the beginning of the buffer. Are you saying, that in your tests the loop ran more than twice? Was it with IPC3 or IPC4? That would only be possible if the input buffer were refilled while we were doing volume processing and that would only be possible if input were generated on a different core, but we don't support linked components on different cores. Could you explain your test case where this problem is observed?

singalsu: Volume as module adapter component should process input_buffers[0].size amount of samples and update ramp at the determined rate. If the block is 6 frames it would loop calculate 48 frames with 8 different ramp gains. The period # of frames is not necessarily an integer multiple of ramp gain block, so the last one can be different length.

Did you see issues with the fact that peakvolume process is called with input_buffers[0].size as zero a few times in the beginning of stream. All the ramp time counts should not start until the component receives data.

singalsu: It should be also possible to continue e.g. the ramp block of e.g. 6 across process periods to keep an even ramp update rate always. I'd go to that direction instead of adjusting the process amount to match exactly N * ramp block.

andrula-song: "we may need to run the while loop in volume_process function several times"  for **ramp**. For example we process 48 frames every time we call the volume_process, while executing ramp(fade in),  we update gain per 6 frames(volume_ramp), we need to execute the while 8 times to finish the processing of 48 frames. But if the ramp is disabled, then the while loop run only once.

andrula-song:
> @andrula-song whats the MCPS impact ?

actually if add 2 flags in struct processing_module, would reduce about 11% cycles than the original one, but would increase other components' MCPS, So finally I move the modification to volume_process itself.

src/audio/module_adapter/module_adapter.c
singalsu: Yes, and it's also very likely that ramp can end in the middle of process period. I think we've missed this regression when converting volume to module adapter (no tests done for ramp). I'd check first doing the changes to volume/peak volume process and keeping the component as normal simple copy client instead of adding new client mode to module adapter.

andrula-song: > It should be also possible to continue e.g. the ramp block of e.g. 6 across process periods to keep an even ramp update rate always. I'd go to that direction instead of adjusting the process amount to match exactly N * ramp block.

yep, the exact number in my comment is just a example.

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -1112,6 +1117,9 @@ static int volume_process(struct processing_module *mod,
 
 		/* copy and scale volume */
 		cd->scale_vol(mod, &input_buffers[0], &output_buffers[0], frames);
+		processed_bytes = frames * frame_byte;
+		audio_stream_consume(source, processed_bytes);
+		audio_stream_produce(sink, processed_bytes);
```
singalsu, line 1122: Are the above consume/produce operations expensive? I think it in addition to wrap also calculates avail and free. Maybe the module adapterization could be reverted for cd->scale_vol() and us the buffer pointers from called function instead. The other components could be confused from buffer usage going back and forth.

singalsu (approved):
Thanks for fix! This fixes the obvious pointers mistake that caused choppy ramp. But there's still more bugs in the IPC4 volume ramp it seems when adjusting the volume up and down repeatedly.

tmleman (approved):

andrula-song:
> @andrula-song The ramp always starts from zero gain, not from previous gain. The fix could be other PR as well.
> 

with PR https://github.com/thesofproject/sof/pull/7197, theramp start is from the previous gain now

Screenshot from 2023-03-02 12-50-43

PR #7127, opened 2023-02-20
remove cAVS-2.0

lyakh:


kv2019i (approved):
Typo in commit (platformsm), but other than that, looks good and tests are passing.

PR #7128, opened 2023-02-20
Tools: Topology2: Add high-pass IIR to HDA analog capture

### LINKED ISSUE #7116: [BUG][MTL][HDA]  A volume drop wave at the beginning of headset recording
**Describe the bug**
when checking headset capture with zero data, we observed there's a volume drop wave at the beginning. Not sure if it's a power up pop noise or a real gain issue. Need help from developer to clarify.

image

image

**To Reproduce**
arecord -Dhw:0,0 -f dat -c 2 test.wav -vvv

**Reproduction Rate**
100%

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * SOF: main branch
2) Name of the topology file
    * Topology: sof-hda-generic-4ch.tplg
3) Name of the platform(s) on which the bug is observed.
    * Platform: MTLP-RVP-HDA with Realtek AIOC31 + Transducer board

singalsu:
This patch includes to sof-hda-generic.conf the new
    highpass-capture-be.conf as passthrough-capture-be.conf replacement
    that adds to DAI copier pipeline the IIR high-pass filter. The
    response is hard coded to 0 dB amplification and 40 Hz cut-off.
    
    The analog microphone headsets creates a strong DC input when the bias
    voltage settles. It can be even full-scale and sound very loud. The
    high-pass filter suppresses such input signal.
    
    Fixes #7116

### DIFF
--- a/tools/topology/topology2/include/pipelines/cavs/highpass-capture-be.conf
+++ b/tools/topology/topology2/include/pipelines/cavs/highpass-capture-be.conf
@@ -1,0 +1,87 @@
+#
+# BE DAI highpass pipeline for capture
+#
+# All attributes defined herein are namespaced by alsatplg to
+# "Object.Pipeline.highpass-be-capture.N.attribute_name"
+#
+# Usage: highpass-capture-be pipeline object can be instantiated as:
+#
+# Object.Pipeline.highpass-capture-be."N" {
+#	direction	"capture"
+# 	period		1000
+# 	time_domain	"timer"
+# 	channels	2
+# 	rate		48000
+# }
+#
+# Where N is the unique pipeline ID within the same alsaconf node.
+#
+
+<include/common/audio_format.conf>
+<include/components/copier.conf>
+<include/components/pipeline.conf>
+<include/controls/bytes.conf>
+<include/components/eqiir.conf>
+
+Class.Pipeline."highpass-capture-be" {
+
+	DefineAttribute."index" {}
+
+	<include/pipelines/pipeline-common.conf>
+
+	attributes {
+		!constructor [
+			"index"
+		]
+
+		unique	"instance"
+	}
+
+	Object.Widget {
+		copier."1" {
+			type dai_in
+			num_audio_formats 1
+			num_input_audio_formats 1
+			num_output_audio_formats 1
+
+			Object.Base.audio_format.1 {
+				in_bit_depth		32
+				in_valid_bit_depth	32
+				out_bit_depth		32
+				out_valid_bit_depth	32
+				dma_buffer_size "$[$obs * 2]"
+			}
+		}
+
+		eqiir."1" {
+			num_audio_formats 1
+			num_input_audio_formats 1
+			num_output_audio_formats 1
+
+			Object.Base.audio_format.1 {
+				in_bit_depth		32
+				in_valid_bit_depth	32
+				out_bit_depth		32
+				out_valid_bit_depth	32
+			}
+
+			Object.Control.bytes."1" {
+				<include/components/eqiir/highpass_40hz_0db_48khz.conf>
+			}
+		}
+
+		pipeline."1" {
+			priority		0
+			lp_mode		0
+		}
+	}
+
+	time_domain	"timer"
+	dynamic_pipeline 1
+	channels	2
+	channels_min	2
+	channels_max	2
+	rate		48000
+	rate_min	48000
+	rate_max	48000
+}
--- a/tools/topology/topology2/cavs-mixin-mixout-hda.conf
+++ b/tools/topology/topology2/cavs-mixin-mixout-hda.conf
@@ -70,7 +70,7 @@ Object.Pipeline {
 		}
 	]
 
-	passthrough-capture-be [
+	highpass-capture-be [
 		{
 			index		4
 			direction	capture
@@ -87,6 +87,11 @@ Object.Pipeline {
 					out_bit_depth		32
 					out_valid_bit_depth	32
 					dma_buffer_size "$[$ibs * 2]"
+				}
+			}
+			Object.Widget.eqiir.1 {
+				Object.Control.bytes."1" {
+					name '4 Main capture Iir Eq'
 				}
 			}
 		}
@@ -125,6 +130,10 @@ Object.Base.route [
 	}
 	{
 		source 'copier.HDA.4.1'
+		sink 'eqiir.4.1'
+	}
+	{
+		source 'eqiir.4.1'
 		sink 'copier.host.3.1'
 	}
 ]
--- a/tools/topology/topology2/sof-hda-generic.conf
+++ b/tools/topology/topology2/sof-hda-generic.conf
@@ -21,6 +21,7 @@
 <deepbuffer-playback.conf>
 <passthrough-be.conf>
 <passthrough-capture-be.conf>
+<highpass-capture-be.conf>
 <data.conf>
 <pcm.conf>
 <pcm_caps.conf>


singalsu:
Here's a picture of the topology. The EQ is added copier.HDA.4.1 pipeline:
 
sof-hda-generic

singalsu:
I actually intended to put 0 dB gain version to this patch. The +20 dB gain version may be too loud. If it would be suitable there would be at least need to add gain component to pipeline to be able to attenuate to suitable level. Comments about desired level are appreciated.

singalsu:
I found that headset capture level is quite suitable with HDA codec setting 0 dB gain and +20 dB boost, so I switched the IIR to 0 dB gain version.

singalsu:
I'm not getting with UPX as strong DC in the beginning but here the before and after waveform pictures look like this:

Screenshot from 2023-02-20 17-20-05
Screenshot from 2023-02-20 17-20-23

tools/topology/topology2/include/pipelines/cavs/highpass-capture-be.conf
```diff
@@ -0,0 +1,93 @@
...
+				dma_buffer_size "$[$obs * 2]"
+			}
+		}
+
+		eqiir."1" {
+			num_audio_formats 2
+			num_input_audio_formats 2
+			num_output_audio_formats 2
+
+			# 32-bit 48KHz 2ch
```
singalsu, line 59: The 24 bit format seems to be unnecessary since the above copier is 32 bits only.

tools/topology/topology2/include/pipelines/cavs/highpass-capture-be.conf
```diff
@@ -0,0 +1,93 @@
...
+# 	channels	2
+# 	rate		48000
+# }
+#
+# Where N is the unique pipeline ID within the same alsaconf node.
+#
+
+<include/common/audio_format.conf>
+<include/components/copier.conf>
+<include/components/pipeline.conf>
```
singalsu, line 22: <include/components/eqiir.conf> and <include/controls/bytes.conf> are missing

plbossart:
maybe a silly question @singalsu but why not use the same processing and parameters as with topology1?

Put differently, why would we need to innovate and have different solutions for the same type of platforms?

lgirdwood:
> maybe a silly question @singalsu but why not use the same processing and parameters as with topology1?
> 
> Put differently, why would we need to innovate and have different solutions for the same type of platforms?

I dont think the platforms are same wrt the HDA codec used on the audio card. @singalsu ?

plbossart:
> > maybe a silly question @singalsu but why not use the same processing and parameters as with topology1?
> > Put differently, why would we need to innovate and have different solutions for the same type of platforms?
> 
> I dont think the platforms are same wrt the HDA codec used on the audio card. @singalsu ?

The topology is exactly the same no matter what HDaudio is used on the board @lgirdwood

lgirdwood:
> > > maybe a silly question @singalsu but why not use the same processing and parameters as with topology1?
> > > Put differently, why would we need to innovate and have different solutions for the same type of platforms?
> > 
> > 
> > I dont think the platforms are same wrt the HDA codec used on the audio card. @singalsu ?
> 
> The topology is exactly the same no matter what HDaudio is used on the board @lgirdwood

The fix should is only needed for MTL IIUC from the bug. @singalsu can you make sure this is targeted at the correct HDA files.

singalsu:
> > > > maybe a silly question @singalsu but why not use the same processing and parameters as with topology1?
> > > > Put differently, why would we need to innovate and have different solutions for the same type of platforms?
> > > 
> > > 
> > > I dont think the platforms are same wrt the HDA codec used on the audio card. @singalsu ?
> > 
> > 
> > The topology is exactly the same no matter what HDaudio is used on the board @lgirdwood
> 
> The fix should is only needed for MTL IIUC from the bug. @singalsu can you make sure this is targeted at the correct HDA files.

We have this same fix in every Topology1 sof-hda-generic topology. See HDA1.IN to PCM0C capture pipeline:

sof-hda-generic

In the tools/build_tools/topology/sof-hda-generic.conf line 646 the IIR definition is this same filter:

```
# 40 Hz second order high-pass, gain 0 dB, created with example_iir_eq.m 27-Nov-2019
SectionData."eq_iir_coef_highpass_40hz_0db_48khz.m4" {
	       bytes "0x53,0x4f,0x46,0x00,0x00,0x00,0x00,0x00,
       0x58,0x00,0x00,0x00,0x00,0xc0,0x00,0x03,
       0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,
       0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,
       0x58,0x00,0x00,0x00,0x02,0x00,0x00,0x00,
       0x01,0x00,0x00,0x00,0x00,0x00,0x00,0x00,
       0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,
       0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,
       0x00,0x00,0x00,0x00,0x01,0x00,0x00,0x00,
       0x01,0x00,0x00,0x00,0x00,0x00,0x00,0x00,
       0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,
       0x00,0x00,0x00,0x00,0x4c,0xdf,0x78,0xc0,
       0x23,0xae,0x86,0x7f,0x12,0x1b,0xf5,0x1f,
       0xdb,0xc9,0x15,0xc0,0x12,0x1b,0xf5,0x1f,
       0x00,0x00,0x00,0x00,0xb2,0x7f,0x00,0x00"

}

```

@plbossart This is is not a new solution. It just has been missing from all topology2 work so far.

singalsu:
@lgirdwood @plbossart The DC in the beginning happens with every single-ended (not enough pins in 3.5 mm stereo headset for differential) connected analog microphone when the bias voltage stabililizes so it's needed in every analog input. Apparently the codecs save this digital logic of the high-pass filter since this has been seen in (nearly) all HDA and SDW codec systems.

tools/topology/topology2/include/pipelines/cavs/highpass-capture-be.conf
singalsu: Fixed now

lgirdwood:
> @plbossart This is is not a new solution. It just has been missing from all topology2 work so far.

ok, so you mean this is already in topology1 but missing in topology2 ?

singalsu:
> > @plbossart This is is not a new solution. It just has been missing from all topology2 work so far.
> 
> ok, so you mean this is already in topology1 but missing in topology2 ?

Yep.

lgirdwood (approved):

btian1:
One concern is：IIR have potential precision issue, if use IIR for long time capture, there may have audio distortion after some time, do we have FIR HP filter for this case?

singalsu:
> The DC noise also have negative impact on speaker, do we implement same on playback?

We've added topology to development sof-hda-efx-generic*.tplg that contains but IIR and FIR for playback path. They are currently set for pass-through but it's simple to activate a filter on one or both with sof-ctl runtime or embed to topology the configuration.

> 
> Another concern is：IIR have potential precision(error accumulation) issue, if use IIR for long time playback/capture, there may have audio distortion after some time, do we have FIR HP filter for this case?

It's possible to do but we don't have such example. True 24 or 32 bit quality (e.g. something like -140 dB THD+N and better) would likely need a FIR. The FIR filter can be designed with tools/tune/eq as parametric filter or directly with Matlab or Octave firpm() or remez()  and use the tools/tune/eq blob packing functions.

lgirdwood:
SOFCI TEST

PR #7131, opened 2023-02-20
mixin-mixout: fix 2 corner cases

lyakh:
Fix a case of empty or inactive sources and `audio_stream_set_zero()` error handling

### DIFF
--- a/src/audio/mixin_mixout/mixin_mixout.c
+++ b/src/audio/mixin_mixout/mixin_mixout.c
@@ -553,6 +553,8 @@ static int mixout_process(struct processing_module *mod,
 		sink_bytes = dev->frames * audio_stream_frame_bytes(mod->output_buffers[0].data);
 		if (!audio_stream_set_zero(mod->output_buffers[0].data, sink_bytes))
 			mod->output_buffers[0].size = sink_bytes;
+		else
+			mod->output_buffers[0].size = 0;
 	}
 
 	module_source_info_release(mod_source_info);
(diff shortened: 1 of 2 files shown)


src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -512,7 +512,9 @@ static int mixout_process(struct processing_module *mod,
 			continue;
 
 		pending_frames = md->pending_frames[source_index];
-		frames_to_produce = MIN(frames_to_produce, pending_frames);
+
+		if (source->state == COMP_STATE_ACTIVE || pending_frames)
+			frames_to_produce = MIN(frames_to_produce, pending_frames);
```
kv2019i, line 517: I can understand the !COMP_STATE_ACTIVE case, but how come we can ignore zero pending frames case? Wouldn't this cause audio quality issue if the active source comes back with data on next 1ms tick, but we already mixed out zeros to the output?

kv2019i (commented):
Other patches ok, but not sure about the behaviour change in the second patch. Please see question inline.

lgirdwood (approved):

src/audio/mixin_mixout/mixin_mixout.c
lyakh: @kv2019i you mean when that zero-pending source is active? Then we **will** fo the calculation and `frames_to_produce` will be 0

kv2019i: Ack @lyakh , so logic is correct.

kv2019i (approved):
Thanks @lyakh , clear now.

PR #6939, opened 2023-01-11
dai: change to use new version of dai_config_get

juimonen:
Start using new version of dai_config_get where config struct is given as pointer argument from outside.

Signed-off-by: Jaska Uimonen <jaska.uimonen@intel.com>

PR #7133, opened 2023-02-20
[DNM] west.yml: upgrade Zephyr to 9af2789cad

kv2019i:
#9125 and #6939 failed to different errors. This PR combines both, let's see if we get a test pass (which would be a logical outcome).


kv2019i:
Baseline update done in #7135 , closing this one.

PR #7136, opened 2023-02-20
include: sof: debug: Split panic.h into Zephyr and XTOS-specific headers

LaurentiuM1234:
The purpose of this commit is to separate Zephyr-specific definitions from XTOS-specific definitions. Based on the build, <rtos/panic.h> will contain the required definitions.

If the commit from this PR is acceptable, I believe the next step should be to remove `sof/debug/panic.h` altogether and replace it by `rtos/panic.h`.

Tested by doing XTOS and Zephyr builds for imx8 and imx8x.

EDIT: also included the removal of `sof/debug/panic.h` and its replacement with `rtos/panic.h`


dbaluta:
FYI: @lgirdwood  our plan is to support  Zephyr + SOF on arm64.

kv2019i (commented):
Excellent, thanks @LaurentiuM1234 . This seems to be indeed the correct path and will help to clean the rtos split further. I have some pending work to send a IPC at panic, but on both OS'es, this needs to be plugged into lower levels of panic processing, so no conflict with this PR.

paulstelian97 (approved):

iuliana-prodan (approved):

LaurentiuM1234:
> This looks good to me. Did you check that applying each patch incrementally doesn't cause any build issue? Asking just to be sure.
> 
> This is helps users doing bisection in the future.

Just did a re-check and it seems all good. Also, the commits were added to the PR incrementally and I'm assuming if there were any problems arising from applying only a subset of the patches they would have been detected by the CI.

Note: all the build checks I've done locally were for imx8 and imx8x

dbaluta:
@LaurentiuM1234 AFAIK, CI only compiles the entire patchset, it doesn't do incremental compilation.

LaurentiuM1234:
> @LaurentiuM1234 AFAIK, CI only compiles the entire patchset, it doesn't do incremental compilation.

What I meant was basically the following flow:

1. Push first commit
2. Wait for CI results
3. Push second commit
4. Wait for CI results
5. Push third commit
6. Wait for CI results

Although the purpose of adding the commits this way wasn't to test them incrementally using the CI, I was assuming they were still tested this way. Please correct me if I'm wrong.

dbaluta:
@LaurentiuM1234 I see. We are OK, then.

kv2019i (approved):
Thanks @LaurentiuM1234 , looks good now!

lgirdwood (approved):

marc-hb:
I think it's impossible to make any header change without breaking something unexpected. This PR broke `scripts/fuzz.sh` which is not in CI yet (working on it). Fix submitted in:
- #7318

PR #7137, opened 2023-02-21
[RFC]module_adapter: introduce SYS_COMP_MODULE_INIT macro

aiChaoSONG:
This patch introduces the SYS_COMP_MODULE_INIT helper macro for module initialization function, and uses it globally, so that for modules use module adapter, we have a unified way to declare and use the module initialization function.


aiChaoSONG:
build error, close until I can fix

Issue #7084, opened 2023-02-13
[BUG] If ChainDMA is enabled FW crashes during multiple-pause-resume test


jsarha:
**Describe the bug**
If ChainDMA is enabled in topology the multiple-pause-resume test fails in CI:
https://sof-ci.01.org/sofpr/PR7073/build3851/devicetest/index.html

**To Reproduce**
Enable ChainDMA for TGL by reverting c3807ae4a76 and run on TGL upx-i11:
TPLG=/lib/firmware/intel/avs-tplg/sof-hda-generic.tplg MODEL=TGLU_UP_HDA_IPC4ZPH ~/work/sof-test/test-case/multiple-pause-resume.sh -r 50

**Reproduction Rate**
Does not always fail immediately, but sometimes takes over ten seconds. However, the FW always crashes well before the test completes.

**Expected behavior**
multiple-pause-resume test would pass.

**Impact**
Can not enable ChainDMA when its broken and no multichannel audio without ChainDMA.

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel: {aa9bd450439db90f28c438641d159a3dd8c41297}
    * SOF: {7b8da69176bf2c5247701b9d409923954362deb4} with c3807ae4a7617314b753ae7713200e3ae918e039 reverted.
2) Name of the topology file
    * Topology: {sof-hda-generic.tplg}
3) Name of the platform(s) on which the bug is observed.
    * Platform: {TGL upx-i11}

**Screenshots or console output**
Please refer to: https://sof-ci.01.org/sofpr/PR7073/build3851/devicetest/index.html

I have executed the test several times and there is no clear pattern to be seen on what happened on FW logs before the crash. However, from the Linux side logs its visible that the crash always happens while handling ChainDMA messages:
Feb 13 15:35:58 jhary kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc timed out for 0xe040000|0x0
...
Feb 13 16:19:46 jhary kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc timed out for 0xe050101|0x0
...
Feb 13 16:26:02 jhary kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc timed out for 0xe070101|0x0
...
Feb 13 22:13:53 jhary kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc timed out for 0xe050000|0x0

[...]

PR #7138, opened 2023-02-21
ll-schedule: fix tasks removed during execution

lyakh:
Zephyr LL-scheduler is supposed to be able to handle tasks, removed or added while the scheduler is executing. However, there is a bug in that implementation. If the task, that is currently executing, is removed, its list head, that links it into the list of tasks, is initialised. So when trying to get a pointer to the next task we obtain a pointer to the same task.

BugLink: https://github.com/thesofproject/sof/issues/7084

### DIFF
--- a/src/audio/chain_dma.c
+++ b/src/audio/chain_dma.c
@@ -303,7 +303,6 @@ static int chain_task_start(struct comp_dev *dev)
 	}
 
 	pm_policy_state_lock_get(PM_STATE_RUNTIME_IDLE, PM_ALL_SUBSTATES);
-	cd->chain_task.state = SOF_TASK_STATE_INIT;
 	k_spin_unlock(&drivers->lock, key);
 
 	return 0;
--- a/src/schedule/zephyr_ll.c
+++ b/src/schedule/zephyr_ll.c
@@ -162,19 +162,20 @@ static void zephyr_ll_run(void *data)
 {
 	struct zephyr_ll *sch = data;
 	struct task *task;
-	struct list_item *list;
+	struct list_item *list, *tmp, task_head = LIST_INIT(task_head);
 	uint32_t flags;
 
 	zephyr_ll_lock(sch, &flags);
 
 	/*
-	 * We have to traverse the list manually, because we drop the lock while
-	 * executing tasks, at that time tasks can be removed from or added to
-	 * the list.
+	 * We drop the lock while executing tasks, at that time tasks can be
+	 * removed from or added to the list, including the task that was
+	 * executed. Use a temporary list to make sure that the main list is
+	 * always consistent and contains the tasks, that we haven't run in this
+	 * cycle yet.
 	 */
-	list = sch->tasks.next;
 
-	while (list != &sch->tasks) {
+	for (list = sch->tasks.next; !list_is_empty(&sch->tasks); list = sch->tasks.next) {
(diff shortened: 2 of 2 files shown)


lgirdwood (approved):

iuliana-prodan (approved):

marc-hb:
Could we merge these two commits over two different days? So we have one run of daily test with just the first commit, which looks like a very big change.

Also, compatibility with upstream Zephyr was just broken, some daily tests are already broken: https://github.com/thesofproject/sof/actions/runs/4228168969/jobs/7343348093
```
/zep_workspace/sof/src/lib/dai.c: In function 'dai_get_zephyr_device':
/zep_workspace/sof/src/lib/dai.c:159:53: error: passing argument 2 of 'dai_config_get' makes pointer from integer without a cast [-Werror=int-conversion]
  159 |                 cfg = dai_config_get(zephyr_dev[i], dir);
      |                                                     ^~~
      |                                                     |
      |                                                     int
In file included from /zep_workspace/sof/src/lib/dai.c:23:
/zep_workspace/zephyr/include/zephyr/drivers/dai.h:336:53: note: expected 'struct dai_config *' but argument is of type 'int'
  336 |                                  struct dai_config *cfg,
      |                                  ~~~~~~~~~~~~~~~~~~~^~~
/zep_workspace/sof/src/lib/dai.c:159:23: error: too few arguments to function 'dai_config_get'
  159 |                 cfg = dai_config_get(zephyr_dev[i], dir);
      |                       ^~~~~~~~~~~~~~
/zep_workspace/zephyr/include/zephyr/drivers/dai.h:335:19: note: declared here
  335 | static inline int dai_config_get(const struct device *dev,
      |                   ^~~~~~~~~~~~~~
```

lyakh:
converted back to draft: tests are still failing (real tests, not this timing out CI run), and I cannot for now figure out why. I'll need to debug later.

lyakh:
SOFCI TEST

lyakh:
The single failure is a suspend-resume test https://sof-ci.01.org/sofpr/PR7138/build4166/devicetest/index.html?model=ADLP_RVP_SDW_IPC4ZPH&testcase=check-suspend-resume-with-playback

src/audio/chain_dma.c
```diff
@@ -303,7 +303,6 @@ static int chain_task_start(struct comp_dev *dev)
 	}
 
 	pm_policy_state_lock_get(PM_STATE_RUNTIME_IDLE, PM_ALL_SUBSTATES);
-	cd->chain_task.state = SOF_TASK_STATE_INIT;
```
makarukp, line 306: Can we also remove 'schedule_task_cancel(&cd->chain_task)' in 'chain_task_pause' ? I think after this change it will not be necessary.

jsarha (approved):
I run multiple-suspend-resume test number of times on my  upx-i11 with this PR and "hda: chain dma: cancel task before freeing it" reverted. I got one transient failure (but no crash), but on three other times the test was successful. So I guess this is a proper fix for the ll scheduler crashing problem.

src/audio/chain_dma.c
lyakh: @makarukp this should be a separate PR, reverting that change - after this is merged

makarukp (approved):

PR #7141, opened 2023-02-21
Tools: Topology2: Add IIR high-pass with gain to DMIC capture

singalsu:
This patch replaces in DMIC0 capture the gain.N.1 component with IIR high-pass. The gain has been a placeholder
dai-copier-gain-module-copier-capture.conf while the EQ component has not been available. The pipeline is replaced by
otherwise similar dai-copier-eqiir-module-copier-capture.conf.

The dmic-default.conf sets the default IIR response to 40 Hz high-pass with 0 dB gain. The topologies will get only the high-pass filter but no amplification.  The default is changed in sof-hda-generic.conf to 20 dB gain to address the too silent capture because these topologies for generic Linux end users do not contain any other capture enhancements.


singalsu:
Example with sof-hda-generic-4ch.tplg, the gain.12.1 is changed to  eqiir.12.1:

sof-hda-generic-4ch

RanderWang (commented):
Only 2ch stream is supported ? We also use 4ch for DMIC0

tools/topology/topology2/include/pipelines/cavs/dai-copier-eqiir-module-copier-capture.conf
```diff
@@ -0,0 +1,126 @@
...
+			sink	copier.module.$index.2
+		}
+	}
+
+	direction	"capture"
+	dynamic_pipeline 1
+	time_domain	"timer"
+	channels	2
+	channels_min	2
+	channels_max	2
```
singalsu, line 122: @RanderWang Did you mean this? I found yesterday that 4 channels capture worked despite of these set to 2. But it makes sense, this should be 4. Is changing only channels_max enough?

RanderWang: I am not sure, but it is very possible.

tools/topology/topology2/cavs-sdw.conf
```diff
@@ -11,7 +11,7 @@
 <virtual.conf>
 <host-copier-gain-mixin-playback.conf>
 <mixout-gain-dai-copier-playback.conf>
-<dai-copier-gain-module-copier-capture.conf>
+<dai-copier-eqiir-module-copier-capture.conf>
```
lgirdwood, line 14: Would this be needed for SDW DMICs too ?

singalsu: I'd assume the SDW codec can be set for suitable gain and it hopefully also handles the mic start transient for ALH DAI.

This change is impacting the cAVS/ACE direct attach DMIC DAI usage. It seems to be enabled in many SDW platform topologies. E.g. sof-mtl-rt711-4ch.tplg:

sof-mtl-rt711-4ch

sof-tgl-rt711-rt1308-4ch.tplg:

sof-tgl-rt711-rt1308-4ch

sof-adl-rt711-4ch.tplg:

sof-adl-rt711-4ch

tools/topology/topology2/include/pipelines/cavs/dai-copier-eqiir-module-copier-capture.conf
singalsu: OK, I just fixed the channels_max

tools/topology/topology2/cavs-sdw.conf
lgirdwood: Can we do direct DMICs first and then see if this is needed for SDW ?

> I'd assume the SDW codec can be set for suitable gain and it hopefully also handles the mic start transient for ALH DAI.

You don't sound sure, probably better to add direct DMIC first and then evaluate SDW DMICs as another PR ?

plbossart:
@singalsu @lgirdwood we've enabled the high-pass for the SoundWire mic codec path before, see

````
ifdef(`NO_LOCAL_MIC', `',
`
# Low Latency capture pipeline 5 on PCM 4 using max 2 channels of s32le.
# Schedule 48 frames per 1000us deadline with priority 0 on core 0
PIPELINE_PCM_ADD(sof/pipe-highpass-switch-capture.m4, <<<<< HERE!
	5, 4, 2, s32le,
	1000, 0, 0,
	48000, 48000, 48000)
')
````

I'd like to see the topology2 case just follow what was enabled before....Let's not be creative or re-invent the wheel.

lgirdwood:
> I'd like to see the topology2 case just follow what was enabled before....Let's not be creative or re-invent the wheel.

Ack - we need same config as topology1. @singalsu can you make this clear if we deviate with any new topology2 updates vs topology1. Thanks.

lgirdwood (approved):

singalsu:
> @singalsu @lgirdwood we've enabled the high-pass for the SoundWire mic codec path before, see
> I'd like to see the topology2 case just follow what was enabled before....Let's not be creative or re-invent the wheel.

Yep true, I've forgotten that case, so it should be then added too. Should it be part of this PR or separate work?

tools/topology/topology2/cavs-sdw.conf
singalsu: > Can we do direct DMICs first and then see if this is needed for SDW ?
> 
> > I'd assume the SDW codec can be set for suitable gain and it hopefully also handles the mic start transient for ALH DAI.
> 
> You don't sound sure, probably better to add direct DMIC first and then evaluate SDW DMICs as another PR ?

Yes, I prefer other PR. I just got access with help from @keqiaozhang to a suitable device with SDW and I will test with it.

RanderWang (approved):

lgirdwood:
> Yep true, I've forgotten that case, so it should be then added too. Should it be part of this PR or separate work? I'd prefer separate because it's more tricky for me to test.

Do a new PR for it. Thanks

marc-hb:
As found by @keqiaozhang this broke MTL:
- #7204

PR #7142, opened 2023-02-21
Switch to using rtos/sof.h instead of sof/sof.h

LaurentiuM1234:
The purpose of this commit is to separate the XTOS-specifc code from the Zephyr-specifc code found in sof/sof.h.

### DIFF
--- a/zephyr/include/rtos/sof.h
+++ b/zephyr/include/rtos/sof.h
@@ -1,0 +1,110 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright(c) 2016 Intel Corporation. All rights reserved.
+ *
+ * Author: Liam Girdwood <liam.r.girdwood@linux.intel.com>
+ */
+
+#ifndef __ZEPHYR_RTOS_SOF_H__
+#define __ZEPHYR_RTOS_SOF_H__
+
+#include <sof/common.h>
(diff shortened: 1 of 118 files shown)


zephyr/include/rtos/sof.h
```diff
@@ -0,0 +1,110 @@
...
+	struct notify_data *notify_data;
+
+	/* platform dai information */
+	const struct dai_info *dai_info;
+
+	/* platform DMA information */
+	const struct dma_info *dma_info;
+
+	/* cascading interrupt controller root */
+	struct cascade_root *cascade_root;
```
lgirdwood, line 81: Fwiw, I bet we can start removing some of these now in Zephyr config as a lot of HW related code has been moved to Zephyr APIs. @abonislawski @lyakh fyi.

lgirdwood (approved):

iuliana-prodan (approved):

bhiregoudar (approved):

zephyr/include/rtos/sof.h
LaurentiuM1234: This will probably require more cleanup patches as more and more stuff gets moved to Zephyr. For now, will try to remove the obvious unused function headers and structure fields when splitting the headers.

dbaluta (approved):

PR #7145, opened 2023-02-22
Host: limit bytes per copy to one period for the whole pipeline

btian1:
Previously, this limiation only implemented on playback pipeline. Now move it both playback and capture pipeline, since capture also need this limitation to avoid high load spike.


ranj063 (approved):

RanderWang (approved):

src/audio/host-zephyr.c
```diff
@@ -484,22 +484,18 @@ static uint32_t host_get_copy_bytes_normal(struct comp_dev *dev)
...
-				  avail_bytes, free_bytes);
-	}
+
+	/* limit bytes per copy to one period for the whole pipeline
+	 * in order to avoid high load spike
+	 */
+	copy_bytes = MIN(hd->period_bytes, MIN(avail_bytes, free_bytes));
+	if (!copy_bytes)
+		comp_info(dev, "no bytes to copy, %d in buffer, %d in DMA",
+			  avail_bytes, free_bytes);
```
aiChaoSONG, line 498: for playback, avail_bytes is the bytes available in dma, free_bytes is the free bytes in downstream buffer; but for capture, avail_bytes is the bytes available in buffer, and free bytes is the free bytes in downstream DMA. so the log info here probably is not correct.

aiChaoSONG (changes requested):
the log is not correct, avail_bytes and free bytes are direction dependent.

src/audio/host-zephyr.c
btian1: it is just print some information help developer debug when there is a bug exists,  to say wrong is not proper, afterall, deleoper always add/remove logs when debug, just print avail_bytes and free_bytes, developer have the sense to justify printed information.

btian1:
@kv2019i ， please review this PR

src/audio/host-zephyr.c
```diff
@@ -484,22 +484,18 @@ static uint32_t host_get_copy_bytes_normal(struct comp_dev *dev)
...
-	} else {
+	else
 		avail_bytes = audio_stream_get_avail_bytes(&buffer_c->stream);
-		copy_bytes = MIN(avail_bytes, free_bytes);
-		if (!copy_bytes)
-			comp_info(dev, "no bytes to copy, %d avail in buffer, %d free in DMA",
-				  avail_bytes, free_bytes);
-	}
+
+	/* limit bytes per copy to one period for the whole pipeline
```
juimonen, line 492: I don't understand how this can work... you are combining playback and capture direction avail and  free bytes with MIN? so depending on direction isn't the other value like random? So it can be like always 0 or whatever the "balance" between free and avail happens to be... or I just don't understand how free_bytes and avail_bytes works.

btian1: @juimonen, the real change is: change capture copy_bytes from: 
copy_bytes = MIN(avail_bytes, free_bytes);
to:
copy_bytes = MIN(hd->period_bytes, MIN(avail_bytes, free_bytes));

Since both playback and capture have same limitation for copy_bytes, then I moved out for save code space.
Please check again.

aiChaoSONG (approved):

juimonen (approved):

abonislawski (approved):

PR #7146, opened 2023-02-22
buffer: add ref_count for buffer sharing

RanderWang:
For smart amp or aec case, one buffer is shared between two streams which may work on different cores. When one of the stream is stopped, the buffer is unbind and removed. Now use ref_count to make sure that it will be deleted when the last user unbind it.

### DIFF
--- a/src/audio/buffer.c
+++ b/src/audio/buffer.c
@@ -180,7 +180,7 @@ void buffer_free(struct comp_buffer *buffer)
 		.buffer = buffer,
 	};
 
-	if (!buffer)
+	if (!buffer || atomic_get(&buffer->ref_count))
 		return;
 
 	buf_dbg(buffer, "buffer_free()");
--- a/src/audio/pipeline/pipeline-graph.c
+++ b/src/audio/pipeline/pipeline-graph.c
@@ -198,6 +198,7 @@ int pipeline_connect(struct comp_dev *comp, struct comp_buffer *buffer,
 	comp_list = comp_buffer_list(comp, dir);
 	buffer_attach(buffer, comp_list, dir);
 	buffer_set_comp(buffer, comp, dir);
+	atomic_inc(&buffer->ref_count);
 
 	irq_local_enable(flags);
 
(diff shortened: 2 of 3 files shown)


lyakh:
I still don't understand why this is needed. `ipc_pipeline_module_free()` seems to be doing the right thing - when the last of the two users of a buffer is disconnected, the buffer is freed too. I don't see why we also need a refcount. You raised this point in a smart amp discussion, does it mean, that that buffer, but also the smart amp's feedback buffer is on its `bsource_list` so it should be freed the same way as the data buffer

RanderWang:
> I still don't understand why this is needed. `ipc_pipeline_module_free()` seems to be doing the right thing - when the last of the two users of a buffer is disconnected, the buffer is freed too. I don't see why we also need a refcount. You raised this point in a smart amp discussion, does it mean, that that buffer, but also the smart amp's feedback buffer is on its `bsource_list` so it should be freed the same way as the data buffer

The feedback buffer is shared by two different streams. When capture is stopped, the feedback buffer is unbind and then free, but it is still used by playback stream. we need to avoid it been free when it is used by playback stream

RanderWang:
Take AEC for example,  one capture copier.ssp.8.1 and aec.13.1 are connected with one buffer. When capture stream (Copier.SSP.8.1) is stopped, the copier.ssp.8.1 and aec.13.1 are unbind and the buffer will be free but it is used by stream (copier.DMIC.14.1)

Untitled

src/audio/buffer.c
```diff
@@ -180,7 +180,7 @@ void buffer_free(struct comp_buffer *buffer)
 		.buffer = buffer,
 	};
 
-	if (!buffer)
+	if (!buffer || atomic_get(&buffer->ref_count))
```
lgirdwood, line 183: How do we guarantee buffer R/W positions are valid if buffer is shared by 2 streams and other multiples cores ?

RanderWang:
I don't like this idea now

Issue #7148, opened 2023-02-22
[BUG][MTL] volume change doesn't take effect during playback


keqiaozhang:
**Describe the bug**
There're something wrong when updating gain.  The volume will not change with the adjustment of gain. The ramp duration should be wrong. 
This issue only happens on all MTL platforms(I2S, HDA and SDW), no such issue on TGL-IPC4 platforms.

**To Reproduce**
1. Terminal A: aplay -Dhw:0,0 -f s16_le -c 2 -r 48000 test.wav
2. Terminal B: change the gain in alsamixer.

**Reproduction Rate**
100%

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * SOF: main
2) Name of the platform(s) on which the bug is observed.
    * Platform: MTL

andrula-song:
I draft a PR https://github.com/thesofproject/sof/pull/7152  to fix it

lgirdwood:
@keqiaozhang can we close ?

PR #7152, opened 2023-02-22
Audio: Fix the volume change doesn't take effect issue

andrula-song:
If the initial_ramp is zero, the ramp_finished would always be true, that means we will never copy gain to circular buffer except we reset state.

### DIFF
--- a/src/audio/module_adapter/module/volume/volume.c
+++ b/src/audio/module_adapter/module/volume/volume.c
@@ -536,6 +536,7 @@ static inline void init_ramp(struct vol_data *cd, uint32_t curve_duration, uint3
 		cd->vol_min = VOL_MIN;
 		cd->vol_max = VOL_MAX;
 	}
+	cd->copy_gain = true;
 }
 
 static int volume_init(struct processing_module *mod)


keqiaozhang:
Verified, this PR can fix issue https://github.com/thesofproject/sof/issues/7148

btian1 (approved):

singalsu (approved):

andrula-song:
SOFCI TEST

tmleman (approved):

lyakh:
was this a regression? This seems to be changing generic code, not MTL-specific, how was it working before on other platforms?

andrula-song:
> was this a regression? This seems to be changing generic code, not MTL-specific, how was it working before on other platforms?

Sorry, not. Actually it didn't work for all HIFI3 and HIFI4 platforms with xcc build before yesterday, because the implementation of scale_vol function will use a circular buffer to store gain 4 times for multi-way xtensa intrinsic. But worked on HIFI3 platforms(with the peak volume detection enabled) after https://github.com/thesofproject/sof/pull/7078 was merged, because I added a new implementation which processing data by channel(do not need to copy gain 4 times) for better performance on HIFI3 version while peak volume detection is enabled.

kv2019i:
A few issues in:
https://sof-ci.01.org/sofpr/PR7152/build4091/devicetest/index.html

- HDA failures are caused by https://github.com/thesofproject/sof/issues/7156 -> not related to this PR
- the TGL/ADL SDW suspend/resume fail is related to upstream kernel rebase -> not related to this PR

Proceeding with merge.

PR #7154, opened 2023-02-22
IDC: set IDC task priority equal to that of the IPC task

lyakh:
One of the functions of the IDC thread is to execute IPCs on secondary cores, those have to be executed with the same priority as when they're run on the primary core, i.e. with the priority of the IPC thread. Set IDC thread priority equal to the IPC thread.

### DIFF
--- a/src/idc/zephyr_idc.c
+++ b/src/idc/zephyr_idc.c
@@ -25,6 +25,7 @@
 #include <sof/drivers/idc.h>
 #include <sof/init.h>
 #include <sof/ipc/common.h>
+#include <sof/schedule/edf_schedule.h>
 #include <rtos/alloc.h>
 #include <rtos/spinlock.h>
 #include <ipc/topology.h>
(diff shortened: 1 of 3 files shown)


lgirdwood (approved):
Good find, they should not preempt LL on secondary cores. @mwasko @marcinszkudlinski fyi.

src/idc/zephyr_idc.c
```diff
@@ -110,8 +110,11 @@ int idc_send_msg(struct idc_msg *msg, uint32_t mode)
 
 	idc_send_memcpy_err = memcpy_s(msg_cp, sizeof(*msg_cp), msg, sizeof(*msg));
 	assert(!idc_send_memcpy_err);
-	/* TODO: verify .priority and .deadline */
-	work->priority = K_HIGHEST_THREAD_PRIO + 1;
+	/*
+	 * Same priority as the IPC thread which is an EDF task and under Zephyr
+	 * its priority is set in scheduler_init_edf() to 1
+	 */
+	work->priority = 1;
```
marcinszkudlinski, line 117: OK, just pls use a meaningful macro - same here and in scheduler_init_edf

lyakh: @marcinszkudlinski done

marcinszkudlinski (approved):

dnikodem (approved):

PR #7158, opened 2023-02-22
ams: add support for ams to kd and kpb modules

iganakov:
Add support for AMS event notifications to KD and KPB modules.
Add 'Key-phrase detected' message to AMS service.  Please review the following commits only:

32c454612: ams: add KPD notification message
bbd073f00: kpb: add ams support
573bf0ec0: kd: add ams support


src/include/sof/lib/ams.h
```diff
@@ -24,6 +24,9 @@
 /* Wildcard for module_id and instance_id values */
 #define AMS_ANY_ID 0xFFFF
 
+/* AMS messages */
+extern const uint8_t LP_KEY_PHRASE_DETECTED_UUID[UUID_SIZE];
```
lyakh, line 28: why should an array be called in all capital letters?

src/lib/ams.c
```diff
@@ -32,6 +32,10 @@ DECLARE_TR_CTX(ams_tr, SOF_UUID(ams_uuid), LOG_LEVEL_INFO);
 
 static struct ams_context ctx[CONFIG_CORE_COUNT];
 
+/* AMS messages */
+const uint8_t LP_KEY_PHRASE_DETECTED_UUID[UUID_SIZE] =
+	"LP_KEYPHDET    ";
```
lyakh, line 37: is this supposed to be a UUID? This doesn't look like one? Although I don't know whether this would be a valid UUID, just usually they look differently

src/audio/kpb.c
```diff
@@ -126,6 +133,52 @@ static uint64_t kpb_task_deadline(void *data)
...
+	inst_id = IPC4_INST_ID(dev_comp_id(dev));
+
+	ret = ams_get_message_type_id(LP_KEY_PHRASE_DETECTED_UUID,
+				      &kpb->am_uuid_id);
+	if (ret)
+		return ret;
+
+	ret = ams_register_consumer(kpb->am_uuid_id, mod_id, inst_id,
+				    kpb_on_ams_kpd_notification, dev);
+	return ret;
```
lyakh, line 166: `return ams_register_consumer(...);`

src/audio/kpb.c
```diff
@@ -763,9 +829,15 @@ static int kpb_prepare(struct comp_dev *dev)
...
 	}
 
+#if CONFIG_AMS
+	/* AMS Register KPB for notification */
+	ret = kpb_register_ams_consumer(dev);
+#else
 	/* Register KPB for notification */
 	ret = notifier_register(dev, NULL, NOTIFIER_ID_KPB_CLIENT_EVT,
 				kpb_event_handler, 0);
+#endif /* CONFIG_AMS */
```
lyakh, line 839: presumably you could make `kpb_event_handler()` depend on `!CONFIG_AMS`. In fact it probably should be done to avoid a compiler warning about an unused static function

src/samples/audio/detect_test.c
```diff
@@ -902,6 +986,13 @@ static int test_keyword_prepare(struct comp_dev *dev)
 					   &cd->data_blob_size,
 					   &cd->data_blob_crc);
 
+#if CONFIG_AMS
+	/* Register KD as AMS producer */
+	ret = test_keyword_register_ams_producer(dev);
+	if (ret)
+		return ret;
+#endif
+
```
lyakh, line 995: also disable everything notifier-related `#if !CONFIG_AMS`?

src/lib/ams.c
iganakov: I wanted to be aligned with AMS on ACE platform implementation. But I agree it doesn't look like valid UUID. I'll use `DECLARE_SOF_RT_UUID` instead

src/include/sof/lib/ams.h
iganakov: Please, see comment below

src/audio/kpb.c
iganakov: Fixed

src/audio/kpb.c
iganakov: Makes sense. Fixed.

src/samples/audio/detect_test.c
iganakov: Fixed

src/include/sof/lib/ams.h
```diff
@@ -24,6 +24,9 @@
 /* Wildcard for module_id and instance_id values */
 #define AMS_ANY_ID 0xFFFF
 
+/* AMS messages */
+extern const struct sof_uuid ams_kpd_msg_uuid; /* Key-phrase detected message*/
```
plbossart, line 28: this does not scale, the AMS service should not have to know about 3rd-party IP messages.

Put differently, we are not going to add a new variable every time someone makes use of AMS.

src/lib/ams.c
```diff
@@ -30,6 +30,10 @@ DECLARE_SOF_UUID("ams", ams_uuid, 0xea9c4bca, 0x5b7d, 0x48c6,
 
 DECLARE_TR_CTX(ams_tr, SOF_UUID(ams_uuid), LOG_LEVEL_INFO);
 
+/* Key-phrase detected AMS message uuid: 80a11122-b36c-11ed-afa1-0242ac120002*/
+DECLARE_SOF_RT_UUID("ams_kpd_msg", ams_kpd_msg_uuid, 0x80a11122, 0xb36c, 0x11ed,
+		    0xaf, 0xa1, 0x02, 0x42, 0xac, 0x12, 0x00, 0x02);
+
```
plbossart, line 36: AMS should be a service that uses UUIDs blindly. Definitions of 3rd party module GUIDs should not be there.

Even if it was the case, this should be conditionally added on KPB being compiled in.

src/audio/kpb.c
```diff
@@ -126,6 +136,51 @@ static uint64_t kpb_task_deadline(void *data)
...
+	struct comp_data *kpb = comp_get_drvdata(dev);
+	uint16_t mod_id = IPC4_MOD_ID(dev_comp_id(dev));
+	uint16_t inst_id = IPC4_INST_ID(dev_comp_id(dev));
+
+	comp_dbg(dev, "kpb_unregister_ams_consumer()");
+
+	return ams_unregister_consumer(kpb->am_uuid_id, mod_id, inst_id,
+				       kpb_on_ams_kpd_notification);
+}
+#endif /* CONFIG_AMS */
```
plbossart, line 182: I don't get how this can be conditional..

Either KPB depends on AMS for functionality or it doesn't depend on AMS at all.

Something's not right here.

plbossart (changes requested):
Sustained objection from me on the way the GUIDs are declared as part of the ams.c code. This cannot work if you want applications to make use of this service, we clearly do not want an endless set of patches to add new GUIDs when two modules need to communicate.

@lgirdwood @mwasko FYI

src/idc/idc.c
```diff
@@ -335,6 +347,9 @@ void idc_cmd(struct idc_msg *msg)
 	case iTS(IDC_MSG_SECONDARY_CORE_CRASHED):
 		idc_secondary_core_crashed(msg->header);
 		break;
+	case iTS(IDC_MSG_AMS):
+		idc_process_async_msg(msg->header & 0xFFFF);
```
lgirdwood, line 351: We probably need a mask macro here to align all users.

lgirdwood (commented):
I wonder if storing basefw, library and/or module GUIDs in an elf linker section with metadata is enough to scale here so that we don't need to change any SW or FW when new external AMS users are added.

Could topology addditionally be used to help establish relationships between AMS users with metadata ?

src/lib/ams.c
lgirdwood: @iganakov we do need a solution where the UUIDs are discovered at runtime, i.e. we have no hard coding in the AMS core. The AMS core should be usable by any module without AMS knowing about that module. This will mean extending the AMS API so that new clients can register their UUIDs (and any other meta data) at runtime (needs to support FW boot and post boot registrations).

src/audio/kpb.c
```diff
@@ -126,6 +136,51 @@ static uint64_t kpb_task_deadline(void *data)
 	return SOF_TASK_DEADLINE_ALMOST_IDLE;
 }
 
+#if CONFIG_AMS
+void kpb_on_ams_kpd_notification(const struct ams_message_payload *const ams_message_payload,
```
lyakh, line 140: `static`? Maybe drop "on?"

src/audio/kpb.c
```diff
@@ -91,10 +95,16 @@ struct comp_data {
 #ifdef CONFIG_IPC_MAJOR_4
 	struct ipc4_kpb_module_cfg ipc4_cfg;
 #endif /* CONFIG_IPC_MAJOR_4 */
+
+#if CONFIG_AMS
+	uint32_t am_uuid_id;
```
lyakh, line 100: `ams_uuid_id`?

src/samples/audio/detect_test.c
```diff
@@ -85,19 +89,25 @@ struct comp_data {
...
-
 	struct ipc_msg *msg;	/**< host notification */
 
 	void (*detect_func)(struct comp_dev *dev,
 			    const struct audio_stream __sparse_cache *source, uint32_t frames);
 	struct sof_ipc_comp_event event;
+
+#if CONFIG_AMS
+	/* AMS related data */
+	uint32_t am_uuid_id;
```
lyakh, line 106: `ams_uuid_id`?

src/lib/ams.c
iganakov: @plbossart @lgirdwood Sure, I agree that messages shouldn't be hardcoded in core AMS implementation. What I'd like to do is to put ams messages definitions for SOF built-in modules into one common place, e.g. `src\include\sof\lib\ams_msg.h`, thus it will be easier to manage them. 3rd party modules can define their own messages independently.

src/audio/kpb.c
iganakov: Fixed

src/audio/kpb.c
iganakov: Ok, fixed

iganakov:
Add "Key-phrase detected" ams message definition in ams_msg.h/ams_msg.c files which will be used by built-in modules

src/lib/ams.c
lyakh: I'm not sure that would make it easier to manage them. Both the consumer and the producer need access to the UUID and they'd also usually need access to a type declaration for the data, transferred with the AMS. So they'll have a common header. Then why don't they declare that UUID in that header too? That way when adding new AMS users they won't need to touch a global header with all UUIDs.

src/lib/ams_msg.c
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) 2023 Intel Corporation
+ *
+ * Author: Ievgen Ganakov <ievgen.ganakov@intel.com>
+ */
+
+#include <sof/lib/ams_msg.h>
+
+/* Key-phrase detected AMS message uuid: 80a11122-b36c-11ed-afa1-0242ac120002*/
+DECLARE_SOF_RT_UUID("ams_kpd_msg", ams_kpd_msg_uuid, 0x80a11122, 0xb36c, 0x11ed,
+		    0xaf, 0xa1, 0x02, 0x42, 0xac, 0x12, 0x00, 0x02);
```
lyakh, line 12: we're still discussing whether we want to collect UUIDs, right? We probably won't get too many AMS users but still I don't see a value in having to change a common file every time you add a new user.

src/samples/audio/detect_test.c
```diff
@@ -140,6 +151,62 @@ static void notify_host(const struct comp_dev *dev)
...
+
+	ret = ams_register_producer(cd->ams_uuid_id, mod_id, inst_id);
+	if (ret)
+		return ret;
+
+	cd->ams_payload.message_type_id = cd->ams_uuid_id;
+	cd->ams_payload.producer_module_id = mod_id;
+	cd->ams_payload.producer_instance_id = inst_id;
+	cd->ams_payload.message_length = sizeof(struct kpb_client);
+	cd->ams_payload.message = (uint8_t *)&cd->client_data;
```
lyakh, line 179: would it make sense to make an exported `kpb_register_ams_producer(const struct comp_dev *dev, struct ams_message_payload *payload)` and only let the caller add `payload->message` and `payload->message_length` after calling it?

src/samples/audio/detect_test.c
```diff
@@ -140,6 +151,62 @@ static void notify_host(const struct comp_dev *dev)
...
+
+	return 0;
+}
+
+static int test_keyword_unregister_ams_producer(const struct comp_dev *dev)
```
lyakh, line 184: this could be a helper too, just add a `uuid` parameter

iganakov:
In new revision I put AMS messages to KPB module. Added "Register KPB client message". Also added helpers for both AMS consumer and producer.

src/samples/audio/detect_test.c
iganakov: Implemented in new revision

mwasko (approved):

sys-pt1s:

src/audio/kpb.c
```diff
@@ -130,6 +137,179 @@ static uint64_t kpb_task_deadline(void *data)
...
+const uint8_t ams_kpd_msg_uuid[UUID_SIZE] = {0x80, 0xa1, 0x11, 0x22, 0xb3, 0x6c,
+					     0x11, 0xed, 0xaf, 0xa1, 0x02, 0x42,
+					     0xac, 0x12, 0x00, 0x02};
+
+/* Register KPB client uuid: c5d9850f-cbdd-4bf4-9c6a-43dfc5faee9f*/
+const uint8_t ams_reg_cli_msg_uuid[UUID_SIZE] = {0xcd, 0x59, 0x85, 0x0f, 0xdd,
+						 0xcb, 0xf4, 0x4b, 0x9c, 0x6a,
+						 0x43, 0xdf, 0xc5, 0xfa, 0xee,
+						 0x9f};
+
```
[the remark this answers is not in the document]
lgirdwood: @iganakov ping - need to unblock. Can we have UUIDs in a global header as a macro definition of the UUID values. Then C files can then include the header and use the named macro with the UUID value ?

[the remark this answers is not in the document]
iganakov: @lgirdwood, sorry for the late reply. Yes, it seems this is the most reasonable way to do it. Thanks!
I'm working on a new revision of this PR

src/ipc/ipc4/ams_helpers.c
```diff
@@ -0,0 +1,79 @@
...
+int ams_helper_register_consumer(struct comp_dev *dev,
+				 uint32_t ams_uuid_id,
+				 const uint8_t *msg_uuid,
+				 ams_msg_callback_fn callback)
+{
+	uint16_t mod_id = IPC4_MOD_ID(dev_comp_id(dev));
+	uint16_t inst_id = IPC4_INST_ID(dev_comp_id(dev));
+	int ret;
+
+	ret = ams_get_message_type_id(msg_uuid, &ams_uuid_id);
```
lyakh, line 46: I think you want the caller of this function to get the UUID ID back? Then the second parameter of this function should be a pointer. Same in `ams_helper_register_producer()`. Has this been runtime-tested?..

iganakov: @lyakh Yes, it has been tested locally and ams is working. But I hit another issue related to IPC4 notification in KD which is under investigation now. Definitely it is a mistake, and it's already fixed. Thanks!

src/audio/kpb.c
lyakh: @plbossart I think this means, that KPB can now use either AMS or the notifier

src/include/sof/lib/ams_msg.h
```diff
@@ -0,0 +1,27 @@
...
+#define DEFINE_UUID_(name, l1, s1, s2, b1, b2, b3, b4, b5, b6, b7, b8) \
+	static const uint32_t name[] = \
+	{ \
+		l1, \
+		s2 << 16 | s1, \
+		b4 << 24 | b3 << 16 | b2 << 8 | b1, \
+		b8 << 24 | b7 << 16 | b6 << 8 | b5, \
+	}
+
+#define DEFINE_UUID(name, ...) DEFINE_UUID_(name, __VA_ARGS__)
```
lyakh, line 25: is this macro AMS-specific? If it is it would be less confusing if you called it `AMS_DEFINE_UUID`

src/audio/kpb.c
```diff
@@ -135,6 +147,24 @@ static uint64_t kpb_task_deadline(void *data)
 	return SOF_TASK_DEADLINE_ALMOST_IDLE;
 }
 
+#if CONFIG_AMS
+
+/* KPD message*/
+DEFINE_UUID(ams_kpd_msg_uuid, AMS_KPD_MSG_UUID);
+
+static void kpb_ams_kpd_notification(const struct ams_message_payload *const ams_message_payload,
```
plbossart, line 155: naming nitpick: what is the meaning of the repeated "kpb" in the helper? is this intentional? Or can we just call this "kpb_ams_notification"?

src/audio/kpb.c
```diff
@@ -347,8 +377,8 @@ static int kpb_unbind(struct comp_dev *dev, void *data)
 
 	return 0;
 }
-
 #else /* CONFIG_IPC_MAJOR_4 */
+
```
plbossart, line 381: spurious change?

src/audio/kpb.c
```diff
@@ -772,9 +816,17 @@ static int kpb_prepare(struct comp_dev *dev)
 		kpb->clients[i].r_ptr = NULL;
 	}
 
+#if CONFIG_AMS
+	/* AMS Register KPB for notification */
+	ret = ams_helper_register_consumer(dev, &kpb->kpd_uuid_id,
+					   (const uint8_t *)ams_kpd_msg_uuid,
```
plbossart, line 822: is the cast necessary? It's a classic with UUIDs to keep cast pointers, can we not use a 'native' uuid_t or something that everyone can agree on?

src/samples/audio/detect_test.c
```diff
@@ -140,6 +150,31 @@ static void notify_host(const struct comp_dev *dev)
...
+
+static int ams_notify_kpb(const struct comp_dev *dev)
+{
+	struct comp_data *cd = comp_get_drvdata(dev);
+	struct ams_message_payload ams_payload;
+
+	cd->client_data.r_ptr = NULL;
+	cd->client_data.sink = NULL;
+	cd->client_data.id = 0; /**< TODO: acquire proper id from kpb */
+	/* time in milliseconds */
```
plbossart, line 166: I don't see a variable name that looks like a time or duration, is this comment intentional?

````c
uint32_t drain_req; /** defines draining size in bytes. */
````

File: src/samples/audio/detect_test.c [hunk 23]
```diff
@@ -85,19 +91,23 @@ struct comp_data {
 	uint32_t keyphrase_samples; /**< keyphrase length in samples */
 	uint32_t drain_req; /** defines draining size in bytes. */
 	uint16_t sample_valid_bytes;
-	struct kpb_event_data event_data;
```
plbossart, line 88: ideally small code moves would be done in a small separate patch.

src/samples/audio/detect_test.c
```diff
@@ -85,19 +91,23 @@ struct comp_data {
...
 	struct kpb_client client_data;
 
 #if CONFIG_KWD_NN_SAMPLE_KEYPHRASE
 	int16_t *input;
 	size_t input_size;
 #endif
-
 	struct ipc_msg *msg;	/**< host notification */
 
 	void (*detect_func)(struct comp_dev *dev,
```
plbossart, line 102: ideally small code moves would be done in a small separate patch.

src/samples/audio/detect_test.c
```diff
@@ -85,19 +91,23 @@ struct comp_data {
...
 			    const struct audio_stream __sparse_cache *source, uint32_t frames);
 	struct sof_ipc_comp_event event;
+
+#if CONFIG_AMS
+	uint32_t kpd_uuid_id;
+#else
+	struct kpb_event_data event_data;
```
plbossart, line 109: ideally small code moves would be done in a small separate patch.

src/samples/audio/detect_test.c
```diff
@@ -887,7 +941,7 @@ static int test_keyword_prepare(struct comp_dev *dev)
 		/* Default threshold value has to be changed
 		 * according to host new format.
 		 */
-		int ret = test_keyword_get_threshold(dev, valid_bits);
+		ret = test_keyword_get_threshold(dev, valid_bits);
```
plbossart, line 944: ideally small code moves would be done in a small separate patch.

plbossart (dismissed):
only minor style changes, looks much better than the initial version. Thanks @iganakov

src/audio/kpb.c
iganakov: @plbossart it's not a repetition of `kpb`, it is another acronym `kpd` which stands for "key phrase detected" message.
But I agree it is a little bit confusing.

src/include/sof/lib/ams_msg.h
iganakov: Yes, it is AMS-specific. I'll change the name to `AMS_DEFINE_UUID`

src/samples/audio/detect_test.c
iganakov: @plbossart yes, this comment was intentional, `drain_req `should be in milliseconds. Maybe with previous implementations  of KPB module draining size was in bytes but now KPB accepts draining size in ms. In `kpb_init_drainig()` it then calculates the number of bytes based on `drain_req ` value received in notification payload.
`size_t drain_req = cli->drain_req * kpb->config.channels *
			       (kpb->config.sampling_freq / 1000) *
			       (KPB_SAMPLE_CONTAINER_SIZE(sample_width) / 8);`

kv2019i (commented):
A few clarification/cleanups, but this starts to seem ready to go. For the one checkpath error, I think this is something we can live with.

src/include/sof/lib/ams_msg.h
```diff
@@ -0,0 +1,27 @@
...
+#ifndef __SOF_LIB_AMS_MSG_H__
+#define __SOF_LIB_AMS_MSG_H__
+
+/* AMS messages */
+
+/* Key-phrase detected AMS message uuid: 80a11122-b36c-11ed-afa1-0242ac120002*/
+#define AMS_KPD_MSG_UUID 0x80a11122, 0xb36c, 0x11ed, 0xaf, 0xa1, 0x02, 0x42, 0xac, 0x12, 0x00, 0x02
+
+#define DEFINE_UUID_(name, l1, s1, s2, b1, b2, b3, b4, b5, b6, b7, b8) \
+	static const uint32_t name[] = \
```
marc-hb, line 17: Why `uint32_t` and the endianness and VA_ARGS complexity? This seems to be always cast to `uint8t_t *` in the end anyway.

src/audio/kpb.c
marc-hb: See below.

src/include/sof/lib/ams_msg.h
```diff
@@ -0,0 +1,27 @@
...
+ * Author: Ievgen Ganakov <ievgen.ganakov@intel.com>
+ */
+
+#ifndef __SOF_LIB_AMS_MSG_H__
+#define __SOF_LIB_AMS_MSG_H__
+
+/* AMS messages */
+
+/* Key-phrase detected AMS message uuid: 80a11122-b36c-11ed-afa1-0242ac120002*/
+#define AMS_KPD_MSG_UUID 0x80a11122, 0xb36c, 0x11ed, 0xaf, 0xa1, 0x02, 0x42, 0xac, 0x12, 0x00, 0x02
```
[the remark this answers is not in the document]
marc-hb: I don't understand what problem DEFINE_UUID() solves. How about something much simpler like this:

```C
typedef uint8_t simple_uuid_t[16];
#define AMS_KPD_MSG_UUID { 0x80, 0xa1, 0x11, 0x22, 0xb3, 0x6c, 0x11, 0xed, 0xaf, 0xa1, 0x02, 0x42, 0xac, 0x12, 0x00, 0x02 }

static const simple_uuid_t ams_kpd_msg_uuid = AMS_KPD_MSG_UUID;
```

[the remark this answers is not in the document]
marc-hb: In case alignement was a concern:
```C
typedef union {
  uint8_t bytes[16];
  uint32_t _; // just for alignment
} simple_uuid_t;

#define AMS_KPD_MSG_UUID { 0x80, 0xa1, 0x11, 0x22, 0xb3, 0x6c, 0x11, 0xed, 0xaf, 0xa1, 0x02, 0x42, 0xac, 0x12, 0x00, 0x02 }

static const simple_uuid_t ams_kpd_msg_uuid = {
  .bytes = AMS_KPD_MSG_UUID
};

```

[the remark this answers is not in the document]
lyakh: > This gives a checkpatch error "ERROR: Macros with complex values should be enclosed in parentheses" , but I think it's ok. At least I can't come up with a better way to wrap this. @lyakh ideas on this?

@kv2019i I actually think including `static` in `DECLARE_SOF_UUID` isn't a very good idea. I'd remove `static` from it and add it in every specific use like `static DECLARE_SOF_UUID(...)`, and move the `__section()` attribute to a more traditional place, e.g. after the name. Would that allow AMS use the same macro then? And in fact `DEFINE_UUID_` also has `static` in it, which shouldn't be done probably either. So, if at all possible, would be best to use the common SOF UUID tooling. But then the problem remains if we really still need a macro with just the values. I think yes, it's ok to ignore the checkpatch error in this case.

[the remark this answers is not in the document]
iganakov: I can do it the way @marc-hb suggested. Can we agree on that @lyakh @kv2019i?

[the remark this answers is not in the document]
lyakh: is it impossible to re-use `DECLARE_SOF_UUID()`? What would be the problem with it - apart from `static`?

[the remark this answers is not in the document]
iganakov: @lyakh AMS message uuid should be defined in common header. There are at least 4 modules which will use AMS - KPB, WHM, WoV, AEC, more are coming. There will be more then 10 messages defined and 3rd party modules should be able to import them. If I'll use a macro like @lgirdwood suggested I won't be able to path its value to `DECLARE_SOF_UUID()` without splitting it to proper args. So I suppose additional macro would be needed.

src/audio/kpb.c
plbossart: oh wow, even on my 24' monitor I didn't see the kbp/kpd difference. Do you mind adding a comment so that this is not lost?

src/include/sof/lib/ams_msg.h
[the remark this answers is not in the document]
plbossart: why not have an extern GUID_BLAH in the header and the value stored in a variable? Macros are evil.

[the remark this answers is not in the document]
marc-hb: > So, if at all possible, would be best to use the common SOF UUID tooling.

The entire idea behind UUIDs is that they can be universally exchanged across all projects/files/modules/languages/etc. without having to share or re-use anything: no header file, no API, no macro, no naming convention, no central registration,... and not tooling. Yet somehow over-engineering manages to creep in anyway :-(

I agree we shouldn't copy/paste ` __section(".static_uuids")` or solutions to endianness issues BUT if some files somewhere only want to share some universal 16 bytes constant (either macro or extern) and nothing else then these files shouldn't have to worry about re-using some "tooling"; that would be overkill.

src/samples/audio/detect_test.c
marc-hb: +1, git commits move around.

https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes

src/include/sof/lib/ams_msg.h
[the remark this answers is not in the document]
marc-hb: > without having to share or re-use anything: no header file, no API,...

Another key feature of UUIDs is IMMUTABILITY. I hate duplication more than anyone else but immutability means UUID values _can_ be and in fact are copied if/when needed. Not just for fun of course but if/when it makes things easier.

“A little copying is better than a little dependency”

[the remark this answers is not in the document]
lyakh: > If I'll use a macro like @lgirdwood suggested I won't be able to path its value to `DECLARE_SOF_UUID()` without splitting it to proper args. So I suppose additional macro would be needed.

@iganakov I've just tried this and it seems to have worked. Whether or not we decide to define UUIDs in headers, it seems like it's also possible with the existing API:
```
diff --git a/src/lib/notifier.c b/src/lib/notifier.c
index a601538d7..ccb524ff6 100644
--- a/src/lib/notifier.c
+++ b/src/lib/notifier.c
@@ -22,8 +22,9 @@
 LOG_MODULE_REGISTER(notifier, CONFIG_SOF_LOG_LEVEL);
 
 /* 1fb15a7a-83cd-4c2e-8b32-4da1b2adeeaf */
-DECLARE_SOF_UUID("notifier", notifier_uuid, 0x1fb15a7a, 0x83cd, 0x4c2e,
-		 0x8b, 0x32, 0x4d, 0xa1, 0xb2, 0xad, 0xee, 0xaf);
+#define uuid_notify 0x1fb15a7a, 0x83cd, 0x4c2e, 0x8b, 0x32, 0x4d, 0xa1, 0xb2, 0xad, 0xee, 0xaf
+#define __DECLARE_SOF_UUID(...) DECLARE_SOF_UUID(__VA_ARGS__)
+__DECLARE_SOF_UUID("notifier", notifier_uuid, uuid_notify);
 
 DECLARE_TR_CTX(nt_tr, SOF_UUID(notifier_uuid), LOG_LEVEL_INFO);
 
```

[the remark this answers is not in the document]
iganakov: @lyakh I've tried your approach and I'm unable to access uuid value in runtime using `DECLARE_SOF_UUID()`. Then I noticed the macro description which tells there is need of uuid dictionary from ldc file to do that. Is there a way to avoid this using SOF uuid infrastructure?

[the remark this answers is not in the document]
marc-hb: Indeed `uuid.h` seems to have grown some coupling with SMEX and .ldc files which the most advanced Zephyr configurations... don't even use anymore.

Again the entire purpose and design of UUIDs is be "universal" and simple by NOT requiring any API/framework/tooling. If the `DECLARE_SOF_UUID()` "tooling" does not work in this particular case then don't use any tooling at all because none is required yet everything will STILL be compatible with anything else using the same UUIDs. That's the magic of UUIDs: compatibility with zero API.

iganakov:
In new revision I defined AMS message uuid the way @marc-hb suggested. 
@lgirdwood @kv2019i @lyakh @plbossart please review again

marc-hb:
I spotted a few genuine compilation issues "hidden" between the unrelated http://archive.ubuntu.com/ubuntu  timeouts

https://github.com/thesofproject/sof/actions/runs/4993519332/jobs/8942765968?pr=7158
```
/home/runner/work/sof/sof/workspace/sof/src/include/sof/coherent.h:70:3: error: use of undeclared identifier 'DCACHE_LINE_SIZE'
} __coherent;
  ^
/home/runner/work/sof/sof/workspace/sof/src/include/sof/coherent.h:20:51: note: expanded from macro '__coherent'
#define __coherent __attribute__((packed, aligned(DCACHE_LINE_SIZE)))
                                                  ^
/home/runner/work/sof/sof/workspace/sof/src/include/sof/coherent.h:321:3: warning: implicit declaration of function 'dcache_invalidate_region' is invalid in C99 [-Wimplicit-function-declaration]
                dcache_invalidate_region(cc, size);
                ^
/home/runner/work/sof/sof/workspace/sof/src/include/sof/coherent.h:334:3: warning: implicit declaration of function 'dcache_writeback_invalidate_region' is invalid in C99 [-Wimplicit-function-declaration]
                dcache_writeback_invalidate_region(c, size);
                ^
/home/runner/work/sof/sof/workspace/sof/src/include/sof/coherent.h:386:3: warning: implicit declaration of function 'dcache_invalidate_region' is invalid in C99 [-Wimplicit-function-declaration]
                dcache_invalidate_region(cc, size);
                ^
/home/runner/work/sof/sof/workspace/sof/src/include/sof/coherent.h:399:3: warning: implicit declaration of function 'dcache_writeback_invalidate_region' is invalid in C99 [-Wimplicit-function-declaration]
                dcache_writeback_invalidate_region(c, size);
                ^
```

src/audio/kpb.c
```diff
@@ -917,8 +970,10 @@ static int kpb_reset(struct comp_dev *dev)
 			kpb_reset_history_buffer(kpb->hd.c_hb);
 		}
 
+#ifndef CONFIG_AMS
 		/* Unregister KPB from notifications */
 		notifier_unregister(dev, NULL, NOTIFIER_ID_KPB_CLIENT_EVT);
+#endif
```
lyakh, line 976: no need to unregister AMS here?

iganakov: No, there is no need to unregister AMS here. `kpb_reset()` is called during `PIPELINE_RESET ` state. RESET on pipeline could be called in two cases:
1. Init -> Reset -> Pause -> Running
2. Pause -> Reset -> Delete

So calling `ams_helper_unregister_consumer()` in kpb_free() will be sufficient.

lyakh: then why is `notifier_unregister()` called both in reset and free? Is that superfluous?

kv2019i (commented):
Looks good to me now. One odditity in first commit, the Kconfig change should be a separate commit.

lgirdwood (approved):

kv2019i:
@plbossart Can you check? 
@iganakov Can you address the Kconfig change in first patch?

iganakov:
@kv2019i Sure, I'll update it today

marc-hb:
Compilation still failing in https://github.com/thesofproject/sof/actions/runs/5048359483/jobs/9056458593?pr=7158

```
 -c /home/runner/work/sof/sof/workspace/sof/src/ipc/ipc4/ams_helpers.c
In file included from /home/runner/work/sof/sof/workspace/sof/src/ipc/ipc4/ams_helpers.c:8:
In file included from /home/runner/work/sof/sof/workspace/sof/src/include/ipc4/ams_helpers.h:11:
In file included from /home/runner/work/sof/sof/workspace/sof/src/include/sof/lib/ams.h:15:
/home/runner/work/sof/sof/workspace/sof/src/include/sof/coherent.h:70:3: error: use of undeclared identifier 'DCACHE_LINE_SIZE'
} __coherent;
  ^
/home/runner/work/sof/sof/workspace/sof/src/include/sof/coherent.h:20:51: note: expanded from macro '__coherent'
#define __coherent __attribute__((packed, aligned(DCACHE_LINE_SIZE)))
                                                  ^
/home/runner/work/sof/sof/workspace/sof/src/include/sof/coherent.h:321:3: warning: implicit declaration of function 'dcache_invalidate_region' is invalid in C99 [-Wimplicit-function-declaration]
                dcache_invalidate_region(cc, size);
                ^
/home/runner/work/sof/sof/workspace/sof/src/include/sof/coherent.h:334:3: warning: implicit declaration of function 'dcache_writeback_invalidate_region' is invalid in C99 [-Wimplicit-function-declaration]
                dcache_writeback_invalidate_region(c, size);
                ^
/home/runner/work/sof/sof/workspace/sof/src/include/sof/coherent.h:386:3: warning: implicit declaration of function 'dcache_invalidate_region' is invalid in C99 [-Wimplicit-function-declaration]
                dcache_invalidate_region(cc, size);
                ^
/home/runner/work/sof/sof/workspace/sof/src/include/sof/coherent.h:399:3: warning: implicit declaration of function 'dcache_writeback_invalidate_region' is invalid in C99 [-Wimplicit-function-declaration]
                dcache_writeback_invalidate_region(c, size);
                ^
In file included from /home/runner/work/sof/sof/workspace/sof/src/ipc/ipc4/ams_helpers.c:9:
In file included from /home/runner/work/sof/sof/workspace/sof/src/include/sof/audio/component.h:19:
In file included from /home/runner/work/sof/sof/workspace/sof/src/include/sof/audio/buffer.h:11:
In file included from /home/runner/work/sof/sof/workspace/sof/src/include/sof/audio/audio_stream.h:22:
In file included from /home/runner/work/sof/sof/workspace/sof/zephyr/include/rtos/cache.h:11:
/home/runner/work/sof/sof/workspace/sof/src/arch/host/include/arch/lib/cache.h:16:20: error: static declaration of 'dcache_invalidate_region' follows non-static declaration
static inline void dcache_invalidate_region(void *addr, size_t size) {}
                   ^
/home/runner/work/sof/sof/workspace/sof/src/include/sof/coherent.h:321:3: note: previous implicit declaration is here
                dcache_invalidate_region(cc, size);
                ^
In file included from /home/runner/work/sof/sof/workspace/sof/src/ipc/ipc4/ams_helpers.c:9:
In file included from /home/runner/work/sof/sof/workspace/sof/src/include/sof/audio/component.h:19:
In file included from /home/runner/work/sof/sof/workspace/sof/src/include/sof/audio/buffer.h:11:
In file included from /home/runner/work/sof/sof/workspace/sof/src/include/sof/audio/audio_stream.h:22:
In file included from /home/runner/work/sof/sof/workspace/sof/zephyr/include/rtos/cache.h:11:
/home/runner/work/sof/sof/workspace/sof/src/arch/host/include/arch/lib/cache.h:18:20: error: static declaration of 'dcache_writeback_invalidate_region' follows non-static declaration
static inline void dcache_writeback_invalidate_region(void *addr,
                   ^
/home/runner/work/sof/sof/workspace/sof/src/include/sof/coherent.h:334:3: note: previous implicit declaration is here
                dcache_writeback_invalidate_region(c, size);
                ^
4 warnings and 3 errors generated.
```

zephyr/CMakeLists.txt
```diff
@@ -443,6 +443,7 @@ zephyr_library_sources_ifdef(CONFIG_IPC_MAJOR_4
 	${SOF_IPC_PATH}/ipc4/dai.c
 	${SOF_IPC_PATH}/ipc4/logging.c
 	${SOF_IPC_PATH}/ipc4/notification.c
+	${SOF_IPC_PATH}/ipc4/ams_helpers.c
```
marc-hb, line 446: You probably need this:
```
zephyr_library_sources_ifdef(CONFIG_IPC_MAJOR_4
      ${SOF_IPC_PATH}/ipc4/ams_helpers.c
)

```

kv2019i (commented):
Looks good now @iganakov . Can you check the ipc4 fuzzer build error in CI? @marc-hb 's idea probably will be enough to fix it. 

@plbossart ping again, this would be ready to go

zephyr/CMakeLists.txt
iganakov: Thanks @marc-hb! Fixed

iganakov:
@kv2019i ipc4 fuzzer error is fixed, thanks

marc-hb:
And now of course the old fuzzer started failing! However this time it's unrelated and can be ignored
https://github.com/thesofproject/sof/actions/runs/5065883155/jobs/9094978916?pr=7158
- #7675

kv2019i:
Known fails in https://sof-ci.01.org/sofpr/PR7158/build8402/devicetest/index.html and https://sof-ci.01.org/sofpr/PR7158/build8403/devicetest/index.html , proceeding with merge.

PR #7160, opened 2023-02-22
topology2: Add kcps pipeline attribute

jsarha:
Adds kcps attribute to pipeline class with 400000 default value, and overrides the value in deepbuffer-playback pipeline with 50000.

### DIFF
--- a/tools/topology/topology2/include/common/tokens.conf
+++ b/tools/topology/topology2/include/common/tokens.conf
@@ -80,6 +80,7 @@ Object.Base.VendorToken {
 		# ABI 4.0 onwards
 		lp_mode		207
 		use_chain_dma		209
+		kcps			210
 	}
 
 	"9" {
--- a/tools/topology/topology2/include/components/pipeline.conf
+++ b/tools/topology/topology2/include/components/pipeline.conf
@@ -13,6 +13,12 @@
 #
 # Where N is the unique instance number for pipeline widget in the same alsaconf node.
 
+Define {
+       # If KCPS value for the pipeline is 0 then the driver does not send any
+       # KCPS adjustment message to the firmware. This is the default.
+       KCPS_PIPELINE_DEFAULT	0
+}
+
 Class.Widget."pipeline" {
 	# pipeline_id for the pipeline widget
 	DefineAttribute."index" {}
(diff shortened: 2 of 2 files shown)


jsarha:
Not sure if I should add come constraints etc., but AFAIK this should be enough get started, and to be able to test Linux driver side.

lgirdwood:
@yongzhi1 which pipelines do you expect to use LF clock, pls let @jsarha know so we can update this PR.

jsarha:
Added {
		# Token reference and type
		token_ref	"sof_tkn_scheduler.word"
} to attribute definition. Without it nothing is added to the topology.

tools/topology/topology2/include/pipelines/cavs/deepbuffer-playback.conf
```diff
@@ -107,6 +107,7 @@ Class.Pipeline."deepbuffer-playback" {
 			priority	0
 			# enable lp mode
 			lp_mode		1
+			kcps		50000
```
lgirdwood, line 110: Both this value and the 400k value need to come from a platform file. e.g.
```
include(platform.conf)
kcps PLATFORM_DEEPBUFFER_KCPS
```

lgirdwood (commented):
@jsarha we also need the real deepbuffer number on MTL too with current code.

tools/topology/topology2/include/pipelines/cavs/deepbuffer-playback.conf
plbossart: Not sure why @lgirdwood, if you have processing for the deep buffer pipeline, that can be entirely dependent on the module that's added on this pipeline. Different algorithms might have different requirements, there's no good default value that work for a given platform across all algorithms or configurations.

lgirdwood (approved):

tools/topology/topology2/include/pipelines/cavs/deepbuffer-playback.conf
lgirdwood: It should only be overridden in the machine topology file where it can be used to adjust machine specific pipelines (aligns with Windows use case now).

PR #7166, opened 2023-02-23
module_adapter: support modules with input only

iganakov:
Current module adapter implementation doesn't support modules with input pin only, such as KD or other detectors.
This type of module just consumes data so there is no need to configure output processing buffer. 
Add parameter to `module_data` to configure input buffer but omit output buffer configuration.

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -12,6 +12,7 @@
 
 #include <sof/audio/buffer.h>
 #include <sof/audio/component.h>
+#include <sof/audio/component_ext.h>
 #include <sof/audio/ipc-config.h>
 #include <sof/audio/module_adapter/module/generic.h>
 #include <sof/audio/pipeline.h>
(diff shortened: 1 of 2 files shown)


src/audio/module_adapter/module_adapter.c
```diff
@@ -144,6 +145,91 @@ struct comp_dev *module_adapter_new(const struct comp_driver *drv,
 	return NULL;
 }
 
+/**
+ * \brief Function to prepare module with input only
+ * @param dev component device pointer.
+ * @return 0 if succeeded, error code otherwise.
+ */
+static int only_input_module_prepare(struct comp_dev *dev)
```
ranj063, line 153: @iganakov would it be possible to use the simple_copy flag for the components of interest and modify the check for mod->num_input_buffers and mod->num_output_buffers to if (!mod->num_input_buffers && mod->num_output_buffers)?

src/audio/module_adapter/module_adapter.c
```diff
@@ -168,8 +254,8 @@ int module_adapter_prepare(struct comp_dev *dev)
 	comp_dbg(dev, "module_adapter_prepare() start");
 
 	/*
-	 * check if the component is already active. This could happen in the case of mixer when
-	 * one of the sources is already active
+	 * Check if the component is already active. This could happen
+	 * in the case of mixer when one of the sources is already active
```
ranj063, line 258: these re unrelated changes

src/audio/module_adapter/module_adapter.c
```diff
@@ -258,19 +354,20 @@ int module_adapter_prepare(struct comp_dev *dev)
...
 	/* Module is prepared, now we need to configure processing settings.
 	 * If module internal buffer is not equal to natural multiple of pipeline
-	 * buffer we have a situation where module adapter have to deep buffer certain amount
-	 * of samples on its start (typically few periods) in order to regularly
-	 * generate output once started (same situation happens for compress streams
-	 * as well).
+	 * buffer we have a situation where module adapter have to deep buffer
+	 * certain amount of samples on its start (typically few periods)
+	 * in order to regularly generate output once started
+	 * (same situation happens for compress streams as well).
```
ranj063, line 368: same here. these are unrelated and unnecessary changes

src/audio/module_adapter/module_adapter.c
```diff
@@ -258,19 +354,20 @@ int module_adapter_prepare(struct comp_dev *dev)
...
 	 */
+
```
ranj063, line 370: this one too

src/audio/module_adapter/module_adapter.c
```diff
@@ -357,15 +458,16 @@ int module_adapter_prepare(struct comp_dev *dev)
 		}
 	} else {
 		list_for_item(blist, &mod->sink_buffer_list) {
-			struct comp_buffer *buffer = container_of(blist, struct comp_buffer,
+			struct comp_buffer *buffer = container_of(blist,
```
ranj063, line 461: can you please remove all the unnecessary changes like this one?

src/audio/module_adapter/module_adapter.c
```diff
@@ -560,10 +670,15 @@ static void module_adapter_process_output(struct comp_dev *dev)
 	struct list_item *blist;
 	int i = 0;
 
+	/* If module has input only just return */
```
ranj063, line 673: I think what you're missing in your component implementations is the simple_copy flag. This function would really be used only component do not produce period bytes in every period

src/include/sof/audio/module_adapter/module/generic.h
```diff
@@ -153,6 +153,7 @@ struct module_data {
 	struct module_processing_data mpd; /**< shared data comp <-> module */
 	void *module_adapter; /**<loadable module interface handle */
 	uint32_t module_entry_point; /**<loadable module entry point address */
+	bool has_input_only; /**< flag to indicate that module has only input pin*/
```
ranj063, line 156: this flag seems redundant, a module with no output would have no sink buffers. So, if a call to if (list_is_empty()) would tell you whether it has an output or not

src/audio/module_adapter/module_adapter.c
iganakov: @ranj063 Ok, `simple_copy ` looks promising. I'm closing this PR and will open another one for simple copy implementation.

PR #7169, opened 2023-02-23
[SKIP CI] docker: upgrade to ubuntu 22.04 as a base image

fredoh9:
Upgrade Ubuntu to latest LTS version, 22.04.

This one line change is supposed to be enough but there is permission issue when UID is not matching with local UID.

```diff
-FROM ubuntu:20.04
+FROM ubuntu:22.04
```

All other lines are the fix for the permission issue.

### DIFF
--- a/scripts/docker-run.sh
+++ b/scripts/docker-run.sh
@@ -33,6 +33,8 @@ test "$(id -u)" = 1000 ||
   >&2 printf "Warning: this script should be run as user ID 1000 to match the container's account\n"
 
 set -x
+# FIXME: During the transition to sudo-cwd.sh, the tag will be "latest_ubuntu22.04".
+#       Later it will be back to latest
 docker run -i -v "${SOF_TOP}":/home/sof/work/sof.git \
 	-v "${SOF_TOP}":/home/sof/work/sof-bind-mount-DO-NOT-DELETE \
 	--env CMAKE_BUILD_TYPE \
@@ -42,6 +44,5 @@ docker run -i -v "${SOF_TOP}":/home/sof/work/sof.git \
 	--env VERBOSE \
 	--env http_proxy="$http_proxy" \
 	--env https_proxy="$https_proxy" \
-	--user "$(id -u)" \
 	$SOF_DOCKER_RUN \
-	thesofproject/sof "$@"
+	thesofproject/sof:latest_ubuntu22.04 ./scripts/sudo-cwd.sh "$@"
--- a/scripts/docker_build/sof_builder/Dockerfile
+++ b/scripts/docker_build/sof_builder/Dockerfile
@@ -14,8 +14,9 @@
 # docker run -it  -v <insert sof dir here>:/home/sof/work/sof.git --user `id -u` sof ./incremental.sh
 #
 
-FROM ubuntu:20.04
-ARG UID=1000
+FROM ubuntu:22.04
+# pick random high value to intentionally mismatch local UID
+ARG UID=1999
 
 ARG host_http_proxy
 ARG host_https_proxy
@@ -62,7 +63,11 @@ ARG CLONE_DEFAULTS="--depth 5"
 
 # Set up sof user
 RUN useradd --create-home -d /home/sof -u $UID -G sudo sof && \
+	echo 'sof ALL = NOPASSWD: ALL' > /etc/sudoers.d/sof && \
+	chmod 0440 /etc/sudoers.d/sof && \
 	echo "sof:test0000" | chpasswd && adduser sof sudo
+# allow others to read/traverse sof home
+RUN chmod o+rx /home/sof
 ENV HOME /home/sof
 
 # Use ToT alsa utils for the latest topology patches.


fredoh9:
There is initial permission issue. ID sof is created with UID 1000, but it actually has different UID. Looking at the problem.

`
sof:x:1011:1011::/home/sof:/bin/sh
`

marc-hb:
Try my new sudo-cwd.sh script: https://github.com/thesofproject/sof/commit/c28400baaa383030f0e6860e922365a79e51c31a

sys-pt1s:

lgirdwood:
@fredoh9 btw why are we using LTS ? We need latest and greatest Ubuntu for current tooling that users and devs will use.

marc-hb:
In this particular case I think it does not really matter because this image is just a container for the latest and greatest ALSA

In general, if it works with the latest LTS then it works with more recent Ubuntu versions. The converse is not always true.

Also, we don't really have the time to debug Ubuntu.

fredoh9:
yes, sof docker image is mainly being used for tools and topologies. And alsa-lib and alsa-utils are updated with building a docker image. I think it doesn't matter of Ubuntu version.

The other use cases are building GCC toolchains and QEMU test, this might need greatest and latest Ubuntu for build tools. Like Marc said every Ubuntu upgrade introduces new problems to debug.

lgirdwood:
Sorry @fredoh9 @marc-hb disagree - LTS will just mean we are out of date re compiler and any build/run time dependecies for ALSA. Developers will be working on the leading edge release and our environment should align. Distro's are also not going to be deploying our new FW/drivers and tooling to old/current LTS versions.

marc-hb:
> Developers will be working on the leading edge release and our environment should align

https://arstechnica.com/information-technology/2020/05/ubuntu-20-04-welcome-to-the-future-linux-lts-disciples/ "Perhaps not surprisingly, Canonical estimates that 95 percent of all Ubuntu installations are LTS releases."

I never use non-LTS releases, the vast majority of my time is already spent with audio-related bugs and even some LTS bugs ( https://github.com/thesofproject/sof-test/discussions ) so I really don't have time for more non-audio bugs. I know some SOF developers upgrade more frequently but that difference never caused any issue so far.

> Distro's are also not going to be deploying our new FW/drivers and tooling to old/current LTS versions.

Yes and that's why @fredoh9 maintains this docker image: because ALSA is always too old even in latest, non-LTS which means upgrading more frequently gives nothing useful.

> LTS will just mean we are out of date re compiler and any build/run time dependecies for ALSA.

Could you please be more specific here, what sort of ALSA dependency? We've always been using LTS releases here so far and I don't remember any issue so I don't understand what problem(s) this extra work will solve.

marc-hb:
https://ubuntu.com/about/release-cycle

> **Interim** releases will introduce new capabilities from Canonical and upstream open source projects, they serve as a **proving ground** for these new capabilities.

It's fantatistic to have some SOF developers testing and finding the latest Ubuntu bugs and incompatibilities and reporting them to Ubuntu but this particular docker image does not need to do that too.

Interim releases are maintained for 9 months. After 9 months, Ubuntu deletes all the binaries! https://www.releases.ubuntu.com/ Then interim releases become literally unusable overnight. This means an urgent ALSA upgrade can be delayed because now @fredoh9 or @juimonen has to debug some Ubuntu issue instead of releasing new ALSA binaries.

fredoh9:
tested with thesofproject/sof:20230413_ubuntu22.04

@lgirdwood @marc-hb
In this PR, I'm still based on Ubuntu 22.04 LTS. Ubuntu 23.04 will come out soon. I will evaluate more when new Ubuntu is available. I will submit a another PR with newer Ubuntu.

marc-hb (approved):
Nice job, this is an area where each line changed tends to represent a LOT of work.

fredoh9:
For my curiosity, I check build tools version

* Ubuntu 20.04 image
- gcc/g++ 9.4.0
- ld 2.34
- make 4.2.1
- cmake 3.16.3

* Ubuntu 22.04 image
- gcc/g++ 11.3.0
- ld 2.38
- make 4.3
- cmake 3.22.1

fredoh9:
@lgirdwood can you approve this? I have a plan to upgrade to the latest Ubuntu.

marc-hb:
> @fredoh9 I think we should stop pointing at the latest and moving tag in the docker shell script, we need to start pointing at a fixed version that you update in PRs like this one.

At least during this backwards-incompatible `sudo-cwd.sh` transition. Then maybe switch back to `latest` later.

fredoh9:
A bit trickier than I thought. Good point. I will point a specific tag rather than latest for now

fredoh9:
Looks good.

https://github.com/thesofproject/sof/actions/runs/4789961866/jobs/8518460013?pr=7169

Digest: sha256:6e65eb281ae7560c6c89f34ce73c77fbec60ca1b22a6ff922d3db51d8651edf5

Errors are due missing 'xtensa-cnl-elf-gcc', which is not part of the docker image any more.

lgirdwood (approved):

lgirdwood:
@wszypelt @lrudyX docker not tested by internal CI. Can we merge ?

wszypelt:
@lgirdwood Internal Intel CI System/merge/build tests green, ready to merge on my part

PR #7171, opened 2023-02-24
module_adapter: use type in abi header as config_id for IPC3

aiChaoSONG:
The type member in struct sof_abi_hdr is used for component's specific blob type for IPC3, some module count on it to know the blob is for module config or module processing algorithm model. We should pass it to set_configuration ops.


aiChaoSONG:
reopen later

PR #7172, opened 2023-02-24
module_adapter: use type in abi header as config_id for IPC3

aiChaoSONG:
The type member in struct sof_abi_hdr is used for component's specific blob type for IPC3, some module count on it to know the blob is for module config or module processing algorithm model. We should pass it to set_configuration ops.

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -920,9 +920,12 @@ static int module_adapter_get_set_params(struct comp_dev *dev, struct sof_ipc_ct
 			pos = MODULE_CFG_FRAGMENT_LAST;
 	}
 
-	/* IPC3 does not use config_id, so pass 0 for config ID as it will be ignored anyway */
+	/*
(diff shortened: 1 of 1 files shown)


aiChaoSONG:
The implementation of the get_configuration ops for ipc3 in module adapter is probably wrong,  will fix it later.

RanderWang (approved):

juimonen (approved):

ranj063:
@aiChaoSONG what issue does this fix?

aiChaoSONG:
> @aiChaoSONG what issue does this fix?

I am trying to convert detect_test module to use module adapter, and in the set_configuration ops, the blob sent by kernel could be SOF_DETECT_TEST_CONFIG or SOF_DETECT_TEST_MODOL for IPC3.  the type member in abi header will tell us which is setting, just like param_id in IPC4. or should we only do conversion for IPC4? 

Link: https://github.com/thesofproject/sof/blob/main/src/samples/audio/detect_test.c#L463~L490

ranj063 (approved):

lgirdwood (approved):

PR #7173, opened 2023-02-24
ll-schedule: fix initial state handling on task free

makarukp:
LL scheduler initial state require treatment in special case when task is already scheduled but not yet in running state. In that specific case it is on tasks list and it has to be deleted from it. Problem could arise when ipc stress tests are applied.

### DIFF
--- a/src/schedule/zephyr_ll.c
+++ b/src/schedule/zephyr_ll.c
@@ -385,6 +385,8 @@ static int zephyr_ll_task_free(void *data, struct task *task)
 	 */
 	switch (task->state) {
 	case SOF_TASK_STATE_INIT:
+		must_wait = false;
+		break;
 	case SOF_TASK_STATE_FREE:
 		on_list = false;
 		/* fall through */


src/schedule/zephyr_ll.c
```diff
@@ -385,6 +385,8 @@ static int zephyr_ll_task_free(void *data, struct task *task)
 	 */
 	switch (task->state) {
 	case SOF_TASK_STATE_INIT:
+		must_wait = false;
+		break;
```
lyakh, line 389: Are you trying to fix the same problem as https://github.com/thesofproject/sof/pull/7138/commits/b3a808afa8c767d8ff0a925eddd902e371ca5d6e ? Do you know why the DMA chain task is doing that? I think it's just wrong and in that case this PR shouldn't be needed.

lyakh (changes requested):

src/schedule/zephyr_ll.c
makarukp: Yes. From some reason I was convinced that calling 'schedule_task' changes state for 'SOF_TASK_STATE_INIT', but now I see it is 'SOF_TASK_STATE_QUEUED'. I agree that states changes should be managed by scheduler and your change is right direction. In such case, I am going close that PR.

PR #7174, opened 2023-02-25
platform: ace: notifying about idle thread readiness

tmleman:
Informing the primary core that the Idle thread on secondary core is ready. During the D3 exit flow thread is not initialize again, but restored from previously saved context.

Requires:

- [x] https://github.com/zephyrproject-rtos/zephyr/pull/55182

### DIFF
--- a/src/platform/intel/ace/platform.c
+++ b/src/platform/intel/ace/platform.c
@@ -24,6 +24,7 @@
 #include <ipc/info.h>
 #include <kernel/abi.h>
 #include <rtos/clk.h>
+#include <sof/lib/cpu.h>
 
 #include <sof_versions.h>
 #include <stdint.h>
@@ -74,6 +75,11 @@ int platform_boot_complete(uint32_t boot_message)
 	return ipc_platform_send_msg(&msg);
 }
 
+static struct pm_notifier pm_state_notifier = {
+	.state_entry = NULL,
+	.state_exit = cpu_notify_state_exit,
+};
+
 /* Runs on the primary core only */
 int platform_init(struct sof *sof)
 {
@@ -113,6 +119,9 @@ int platform_init(struct sof *sof)
 	ret = dmac_init(sof);
 	if (ret < 0)
 		return ret;
+
+	/* register power states entry / exit notifiers */
+	pm_notifier_register(&pm_state_notifier);
 
 	/* initialize the host IPC mechanisms */
 	trace_point(TRACE_BOOT_PLATFORM_IPC);
--- a/zephyr/lib/cpu.c
+++ b/zephyr/lib/cpu.c
@@ -59,11 +59,25 @@ static FUNC_NORETURN void secondary_init(void *arg)
 #if CONFIG_ZEPHYR_NATIVE_DRIVERS
 #include <sof/trace/trace.h>
 #include <rtos/wait.h>
-#include <zephyr/pm/pm.h>
 
 LOG_MODULE_DECLARE(zephyr, CONFIG_SOF_LOG_LEVEL);
 
 extern struct tr_ctx zephyr_tr;
+
+/* notifier called after every power state transition */
+void cpu_notify_state_exit(enum pm_state state)
+{
+	if (state == PM_STATE_SOFT_OFF)	{
+#if CONFIG_MULTICORE
+		if (!cpu_is_primary(arch_proc_id())) {
+			/* Notifying primary core that secondary core successfully exit the D3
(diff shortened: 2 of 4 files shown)


src/platform/intel/ace/platform.c
```diff
@@ -98,6 +115,9 @@ int platform_init(struct sof *sof)
 	if (ret < 0)
 		return ret;
 
+	/* register power states entry / exit notifiers */
+	pm_notifier_register(&pm_state_notifier);
```
lgirdwood, line 119: Is long term switch over to AMS here ?

aborisovich (changes requested):

src/platform/intel/ace/platform.c
tmleman: No, it will stay as it is (if nothing changes).

zephyr/lib/cpu.c
```diff
@@ -27,6 +27,11 @@ extern K_KERNEL_STACK_ARRAY_DEFINE(z_interrupt_stacks, CONFIG_MP_NUM_CPUS,
 static atomic_t start_flag;
 static atomic_t ready_flag;
 
+void ace_mark_thread_as_ready(void)
+{
+	atomic_set(&ready_flag, 1);
+}
```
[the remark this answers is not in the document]
lgirdwood: We could save a function call and static inline this public API too.

sys-pt1s:

zephyr/lib/cpu.c
```diff
@@ -80,7 +85,13 @@ int cpu_enable_core(int id)
 		return 0;
 
 #if ZEPHYR_VERSION(3, 0, 99) <= ZEPHYR_VERSION_CODE
```
marcinszkudlinski, line 87: why we need to check zephyr version? We won't ever go back to older zephyr releases.

paulstelian97: Are there plans to integrate some code from here into Zephyr itself as opposed to it staying in SOF? If so, then these checks are still worth it.

lgirdwood: @paulstelian97 yes, that the direction, IIUC there are still some dependencies to upstream.
I agree though, we dont need the version check as west manifest should allow us to specify supported Zephyr commit.
@tmleman does this PR need a west update ?

tmleman: > does this PR need a west update ?

@lgirdwood Yes, at first I need to merge zephyr PR and after that this can be merged with zephyr update. But regardless of that, everyone can review and approve.

aborisovich (approved):

lgirdwood:
@tmleman any update ?

tmleman:
> any update ?

@lgirdwood I'm still waiting for the merge on the zephyr side

zephyr/lib/cpu.c
```diff
@@ -79,7 +79,13 @@ int cpu_enable_core(int id)
...
-	z_init_cpu(id);
+	/* During kernel initialization, the next pm state is set to ACTIVE. By checking this
+	 * value, we determine if this is the first core boot, if not, we need to skip idle thread
+	 * initialization. By reinitializing the idle thread, we would overwrite the kernel structs
+	 * and the idle thread stack.
+	 */
+	if (pm_state_next_get(id)->state == PM_STATE_ACTIVE)
+		z_init_cpu(id);
```
kv2019i, line 102: Given this is an open-coded  of z_smp_start_cpu() (see comment above), I don't fully understand why we have a different implementation here, and if this is correct, why this is not needed in the Zephyr kernel/smp.c version. Wouldn't the same problem be there? Or is this specific to a configuration where state is saved to persistant memory? If yes, this would need to be ifdef'ed out so this applies only to applicable platforms.

kv2019i (changes requested):
Now I see in second commit this is related to state-save. This should be somehow conditional on the feature being used, given the generic code is used for all platform/vendors.

zephyr/lib/cpu.c
kv2019i: @nashif @teburd @dcpleung @andyross  ^^ trying to understand why z_smp_start_cpu() does not need this?

kv2019i:
Usual PM fail in https://sof-ci.01.org/sofpr/PR7174/build6662/devicetest/index.html (but passes on at least one device) and https://sof-ci.01.org/sofpr/PR7174/build6663/devicetest/index.html has one fail on system PM.

Proceeding with the fix. Thanks @tmleman for filing the issue.

PR #7106, opened 2023-02-16
topology2: enable WoV on mtl hda platform

libinyang:
enable WoV on mtl hda platform

PR #7182, opened 2023-02-28
Fix wov topology issue

aiChaoSONG:
Minor wov topology issues, fix them in this patch, or the wov feature doesn't work

### DIFF
--- a/tools/topology/topology2/include/pipelines/cavs/wov-detect.conf
+++ b/tools/topology/topology2/include/pipelines/cavs/wov-detect.conf
@@ -67,9 +67,9 @@ Class.Pipeline."wov-detect" {
 				out_rate	16000
 				in_channels	2
 				out_channels	1
-				in_ch_map	CHANNEL_MAP_STEREO
+				in_ch_map	$CHANNEL_MAP_STEREO
 				in_ch_cfg	1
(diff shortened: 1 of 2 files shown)


aiChaoSONG:
@ranj063 I'd like to add a test topology sof-hda-generic-2ch-kwd.tplg after this PR, but quote your comments in #7106
> why only MTL, why not TGL/ADL too?

For MTL, we have to set `DMIC_DRIVER_VERSION == 3` for MTL, but it is not needed for TGL/ADL, that means we cannot have a single hda-generic topology for all three MTL/TGL/ADL platforms, but can we have two? for example, sof-mtl-hda-generic-2ch-kwd.tplg/sof-tgl-hda-generic-2ch-kwd.tplg

juimonen:
@aiChaoSONG not sure could we have the default driver_version behind platform conditionals in include/dais/dmic.conf?

aiChaoSONG:
> @aiChaoSONG not sure could we have the default driver_version behind platform conditionals in include/dais/dmic.conf?

The problem is that in a single topology binary, we can only have a single driver_version, which can not support MTL/TGL simultaneously.

lgirdwood (approved):

lgirdwood:
@RanderWang pls review

ranj063 (approved):

PR #7186, opened 2023-02-28
module_adapter: support modules with only sink or source connected

iganakov:
Extend module adapter simple_copy flow to support modules with only one sink or one source connected such as KD or Intel WoV.


src/audio/module_adapter/module_adapter.c
```diff
@@ -230,7 +230,7 @@ int module_adapter_prepare(struct comp_dev *dev)
 	list_for_item(blist, &dev->bsink_list)
 		mod->num_output_buffers++;
 
-	if (!mod->num_input_buffers || !mod->num_output_buffers) {
+	if (!mod->num_input_buffers && !mod->num_output_buffers) {
```
ranj063, line 233: can we change the err message below to say that both input and output buffer counts are 0?

src/audio/module_adapter/module_adapter.c
```diff
@@ -659,17 +663,21 @@ module_single_source_setup(struct comp_dev *dev,
...
 
-		mod->output_buffers[i].size = 0;
-		mod->output_buffers[i].data = &sinks_c[i]->stream;
-		i++;
+			mod->output_buffers[i].size = 0;
+			mod->output_buffers[i].data = &sinks_c[i]->stream;
+			i++;
+		}
 	}
-
```
ranj063, line 672: unrelated change

ranj063 (commented):
looks good, just a couple of nimor nit-picks

src/audio/module_adapter/module_adapter.c
iganakov: Sure

src/audio/module_adapter/module_adapter.c
iganakov: Fixed

ranj063 (approved):

makarukp (approved):

dnikodem (approved):

src/audio/module_adapter/module_adapter.c
```diff
@@ -241,20 +241,24 @@ int module_adapter_prepare(struct comp_dev *dev)
...
-		comp_err(dev, "module_adapter_prepare(): failed to allocate output buffers");
-		ret = -ENOMEM;
-		goto in_free;
+	if (mod->num_output_buffers) {
+		mod->output_buffers = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
+					sizeof(*mod->output_buffers) * mod->num_output_buffers);
+		if (!mod->output_buffers) {
+			comp_err(dev, "module_adapter_prepare(): failed to allocate output buffers");
+			ret = -ENOMEM;
+			goto in_free;
```
lyakh, line 260: this isn't something that this patch is breaking, but clean up labels aren't correct here. Look at `goto in_free;` at line 321 / 325 below: it's leaking `mod->output_buffers`. The order of labels should be (choose whatever names you prefer):
```
free4:
	for (i = 0; i < mod->num_output_buffers; i++)
		rfree((__sparse_force void *)mod->output_buffers[i].data);
free3:
	for (i = 0; i < mod->num_input_buffers; i++)
		rfree((__sparse_force void *)mod->input_buffers[i].data);
free2:
	rfree(mod->output_buffers);
free1:
	rfree(mod->input_buffers);
	return ret;
```
Also if later more initialisation steps are added the respective clean up will just need to be added, the existing steps won't need to be changed

lyakh (commented):
there's a bug that was there before this PR, but since you're touching this code, would be good to fix it too

src/audio/module_adapter/module_adapter.c
iganakov: Good point, thanks. Fixed.

src/audio/module_adapter/module_adapter.c
```diff
@@ -387,16 +396,15 @@ int module_adapter_prepare(struct comp_dev *dev)
...
-in_free:
+in_data_free:
 	for (i = 0; i < mod->num_input_buffers; i++)
 		rfree((__sparse_force void *)mod->input_buffers[i].data);
-
+out_free:
+	rfree(mod->output_buffers);
+in_free:
 	rfree(mod->input_buffers);
 	return ret;
```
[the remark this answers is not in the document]
iganakov: @lyakh Ok, I've added an assignment to NULL for both `input_buffers `and `output_buffers` if their number = 0.

lgirdwood (approved):

src/audio/module_adapter/module_adapter.c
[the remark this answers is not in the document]
lyakh: Ok, yes, this fixes my concern, thanks. Now that's safe against that potential bug. Sorry, *but* - there's a "but": (1) generally if I do an if-else block with both branches present, I'd rather do
```
if (a)
	do_a();
else
	do_b();
```
than
```
if (!a)
	do_b();
else
	do_a();
```
and (2) you added code to cover that potential problem to the main code path. That means it will now always be executed - just to protect an error case. Whereas alternatively you could just place checks for the number of input and output buffers on the error path - below the `in_free` and `out_free` labels. That way your "good" path wouldn't be punished - it wouldn't need to execute any extra code, and only the error path would be "punished" - but the error path is anyway "never" executed and if it is - it's ok for it to run slightly slower. Of course - both remarks perfectly optional. My actual concern is addressed - thanks. Just some ideas to potentially improve it a bit.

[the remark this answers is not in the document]
iganakov: @lyakh I agree with (1).

(2): My intention was to not proceed with any memory allocation ops in case of `num_input_buffers ` or `num_output_buffers ` counts to 0. Please correct me if I'm wrong, in case of module with input or output only this part of the code will always try to allocate memory for `mod->input_buffers` or `mod->output_buffers`. Also, I don't see any checks for the number of `bytes` provided to `rzalloc()` and consecutive `malloc` operations. It looks like we will go deeper and deeper knowing we're going to allocate size = 0. I suppose penalty could be higher without those checks before allocation.

src/audio/module_adapter/module_adapter.c
```diff
@@ -241,20 +241,33 @@ int module_adapter_prepare(struct comp_dev *dev)
...
+		mod->input_buffers = NULL;
+	} else {
+		mod->input_buffers =
+			rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
+				sizeof(*mod->input_buffers) * mod->num_input_buffers);
+
+		if (!mod->input_buffers) {
+			comp_err(dev, "module_adapter_prepare(): failed to allocate input buffers");
+			ret = -ENOMEM;
+			goto in_free;
```
kv2019i, line 254: Hmm, "in_free" will rfree input_buffers, but that alloc already failed, so this does not make sense. 
Instead "return -ENOMEM;" directly here?

src/audio/module_adapter/module_adapter.c
```diff
@@ -241,20 +241,33 @@ int module_adapter_prepare(struct comp_dev *dev)
...
+		mod->output_buffers = NULL;
+	} else {
+		mod->output_buffers =
+			rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
+				sizeof(*mod->output_buffers) * mod->num_output_buffers);
+
+		if (!mod->output_buffers) {
+			comp_err(dev, "module_adapter_prepare(): failed to allocate output buffers");
+			ret = -ENOMEM;
+			goto out_free;
```
kv2019i, line 269: Same here, "out_free" seems wrong, we only need to free the input buffers. This is harmless as output_buffers will be NULL, but the code is still confusing and may lead to bugs later on when other people modify this code and add new cases on top.

kv2019i (changes requested):
The goto labels seem a bit misleading, can you check.

src/audio/module_adapter/module_adapter.c
iganakov: Fixed

src/audio/module_adapter/module_adapter.c
iganakov: @kv2019i I rewrote this flow, should be cleaner now.

src/audio/module_adapter/module_adapter.c
```diff
@@ -241,20 +241,31 @@ int module_adapter_prepare(struct comp_dev *dev)
...
-		goto in_free;
+	if (mod->num_output_buffers) {
+		mod->output_buffers =
+			rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
+				sizeof(*mod->output_buffers) * mod->num_output_buffers);
+		if (!mod->output_buffers) {
+			comp_err(dev, "module_adapter_prepare(): failed to allocate output buffers");
+			if (mod->input_buffers)
+				rfree(mod->input_buffers);
+			return -ENOMEM;
```
lyakh, line 265: `rfree(NULL)` is ok. Please use `goto` and add a label above line 410 below

src/audio/module_adapter/module_adapter.c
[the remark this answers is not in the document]
lyakh: to (2) in principle `sys_heap_aligned_alloc()` has a check for 0-byte allocation, but ok, we can keep it here too.

src/audio/module_adapter/module_adapter.c
iganakov: @lyakh But isn't it easier here to just `rfree(mod->input_buffers)` even without a NULL check and return an error? At that  point only `mod->input_buffers` could be allocated and there is no need to free `mod->output_buffers` and jump to label.

kv2019i: I'm ok with both. Traditionally, if you have a goto-based rollback of resources, you can return immediately on first error when there is no resources to free yet (so goto is not needed), but once you an hit error, you should use the rollbacks and jump to to a goto label. Of course, if you don't have have a rollback and you just free resources at location of the error and return, that's different, but you should not be mixing approaches. Either you have a rollback and you use it, or you don't. But as said, this is minor stuff, not blocking PR because of this.

kv2019i (approved):
Thanks, looks good now! One comment on L265, but I'm good with the current implementation as well.

src/audio/module_adapter/module_adapter.c
iganakov: @kv2019i Thanks for explanation. Add a label to be aligned with the "goto label" approach.

Issue #7184, opened 2023-02-24
[ADL/TGL/MTL NOCODEC] Frequency response is different between left and right channel


fredoh9:
Found alsabat failure with ADLP_RVP_NOCODEC_IPC4ZPH.
https://sof-ci.01.org/linuxpr/PR4212/build3371/devicetest/index.html?model=ADLP_RVP_NOCODEC_IPC4ZPH&testcase=check-alsabat-headset-playback

- Kernel PR: 4212
- SOF Branch: main
- SOF Commit: a74ad932700c
-  Zephyr Commit: zephyr-v3.3.0-200-g9af2789cad17
- tplg: sof-adl-nocodec.tplg

![sof-adl-nocodec](https://user-images.githubusercontent.com/45636982/221083553-bfabff52-f449-4f88-8975-1805df63882f.png)

alsabat test only complains about right channel.

```
BAT analysis: signal has 65536 frames at 48000 Hz, 2 channels, 2 bytes per sample.

Channel 1 - Checking for target frequency 821.00 Hz
Amplitude: 31128.9; Percentage: [95]
Detected peak at 821.04 Hz of 41.56 dB
 Total 42.4 dB from 818.85 to 823.24 Hz
 PASS: Peak detected at target frequency
Detected at least 1 signal(s) in total

Channel 2 - Checking for target frequency 821.00 Hz
Amplitude: 30990.5; Percentage: [94]
Detected peak at 815.92 Hz of 20.72 dB
 Total 20.7 dB from 815.92 to 815.92 Hz
Detected peak at 821.04 Hz of 38.11 dB
 Total 39.5 dB from 817.38 to 822.51 Hz
 PASS: Peak detected at target frequency
Detected peak at 823.97 Hz of 20.52 dB
 Total 39.5 dB from 823.97 to 823.97 Hz
 PASS: Peak detected at target frequency
Detected at least 3 signal(s) in total

Return value is -1003
```

Attached: [ADLP_RVP_NOCODEC_alsabat_wav_captured.zip](https://github.com/thesofproject/linux/files/10820349/ADLP_RVP_NOCODEC_alsabat_wav_captured.zip)

- bat.wav.9LAB5b ; alsabat test output
- left.wav ; only left channel
- right.wav ; only right channel- 

Left and Right channels are separated. Did frequency analysis.
[...]

PR #7187, opened 2023-02-28
clock: replace per-clock locks with a global one

lyakh:
select_cpu_clock() on cAVS currently takes all the clock locks while adjusting them. And that function is called from clock_set_freq() which also takes one of those locks, which leads to a recursive spin-lock. To fix that remove per-clock spin-locks and use a global clock lock instead.

### DIFF
--- a/src/lib/clk.c
+++ b/src/lib/clk.c
@@ -26,6 +26,8 @@ DECLARE_SOF_UUID("clock", clock_uuid, 0x8890ea76, 0x0df9, 0x44ae,
 
 DECLARE_TR_CTX(clock_tr, SOF_UUID(clock_uuid), LOG_LEVEL_INFO);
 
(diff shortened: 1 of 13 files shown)


juimonen (approved):
lgtm. no idea why we have had separate locks if we are not even using them.

lgirdwood (approved):

lyakh:
an alsabat device test failed https://sof-ci.01.org/sofpr/PR7187/build4253/devicetest/index.html?model=ADLP_RVP_NOCODEC_IPC4ZPH&testcase=check-alsabat-headset-playback - not new: #7184

lyakh:
@wszypelt could you check the QB failure, please?

tmleman (approved):

kv2019i (approved):

lgirdwood:
SOFCI TEST

PR #7200, opened 2023-03-02
schedule: remove schedule_task_complete()

lyakh:
schedule_task_complete() is never used, remove it.

### DIFF
--- a/src/include/sof/schedule/schedule.h
+++ b/src/include/sof/schedule/schedule.h
@@ -100,17 +100,6 @@ struct scheduler_ops {
 	int (*schedule_task_running)(void *data, struct task *task);
 
 	/**
-	 * Sets task into completed state along with executing additional
-	 * scheduler specific operations.
-	 * @param data Private data of selected scheduler.
-	 * @param task Task to be set into completed state.
-	 * @return 0 if succeeded, error code otherwise.
-	 *
-	 * This operation is optional.
-	 */
-	int (*schedule_task_complete)(void *data, struct task *task);
-
-	/**
 	 * Reschedules already scheduled task with new start time.
 	 * @param data Private data of selected scheduler.
 	 * @param task Task to be rescheduled.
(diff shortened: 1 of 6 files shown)


src/include/sof/schedule/schedule.h
```diff
@@ -99,17 +99,6 @@ struct scheduler_ops {
...
-	 * Sets task into completed state along with executing additional
-	 * scheduler specific operations.
-	 * @param data Private data of selected scheduler.
-	 * @param task Task to be set into completed state.
-	 * @return 0 if succeeded, error code otherwise.
-	 *
-	 * This operation is optional.
-	 */
-	int (*schedule_task_complete)(void *data, struct task *task);
-
```
aborisovich, line 112: I need it in the IPC that is currently refactored to use this DP scheduler.
Please note that `src\ipc\ipc-common.c` has `ipc_task_ops` struct that calls `ipc_complete_task` on task completion.
Please don't delete.

aborisovich (changes requested):

marcinszkudlinski:
/** \brief Task operations. */
    struct task_ops {
	    enum task_state (*run)(void *data);	/**< task's main operation */
	    void (*complete)(void *data);		/**< executed on completion */
	    uint64_t (*get_deadline)(void *data);	/**< returns current deadline */
    };


void (*complete)(void *data);		/**< executed on completion */
will it be removed as well?

aborisovich:
Hmm, it works for existing IPC and on my IPC with DP scheduler. It looks like it is not related to the `struct task_ops.complete` member. @lyakh can you please Provide more description in the PR why it is never used/what is not exactly happening to tasks?

src/include/sof/schedule/schedule.h
aborisovich: ok, false alarm. Task completion is executed though I do not understand why.
Can you provide some more context @lyakh to the PR description?

lgirdwood:
@marcinszkudlinski pls let @lyakh know what you need, we can close if this API needed ?

lyakh:
@aborisovich @marcinszkudlinski I'm a bit confused now - are there plans to begin to use this and we should keep it or are there no such plans? Confusingly we have two similarly names flows in SOF scheduling today:
1. `schedule_task_complete()` defined in src/include/sof/schedule/schedule.h and calling `.schedule_task_complete()` from `struct scheduler_ops`. As a comment in that structure says, it "Sets task into completed state." The `schedule_task_complete()` function is never called anywhere in present SOF. So, unless you plan to add calls to it, `.schedule_task_complete()` isn't needed and can be deleted in existing schedulers and in DP too
2. `task_complete()` defined src/include/sof/schedule/task.h, calling `.complete()` from `struct task_ops` - that is a task callback, called when a task is completed. It is used and this PR doesn't touch it.

lyakh:
> @marcinszkudlinski pls let @lyakh know what you need, we can close if this API needed ?

@lgirdwood looks like it was a mistake and they don't really need it

marcinszkudlinski (approved):

PR #7203, opened 2023-03-02
[SKIP CI] .github/daily-tests.yml: add sparse-zephyr.yml

marc-hb:
Because why not.

### DIFF
--- a/.github/workflows/daily-tests.yml
+++ b/.github/workflows/daily-tests.yml
@@ -18,3 +18,5 @@ jobs:
     uses: ./.github/workflows/pull-request.yml
   run-zephyr-builds:
     uses: ./.github/workflows/zephyr.yml
+  sparse-zephyr:
+    uses: ./.github/workflows/sparse-zephyr.yml


marc-hb:
In https://github.com/thesofproject/sof/actions/runs/4318983392/jobs/7537920586 checkpatch is clueless and wrong

.github/workflows/daily-tests.yml
```diff
@@ -18,3 +18,5 @@ jobs:
     uses: ./.github/workflows/pull-request.yml
   run-zephyr-builds:
     uses: ./.github/workflows/zephyr.yml
+  sparse-zephyr:
```
kv2019i, line 21: I'd argue some better rationale could be added to the git commit that "why not".

kv2019i (approved):
Makes sense to add this as well, but I'd still add a bit of background to the git commit message.

.github/workflows/daily-tests.yml
marc-hb: Fair enough, I should probably have said something better like "there is no reason to exclude sparse from daily tests" :-)

Thanks for not blocking this.

lgirdwood (approved):

aborisovich:
Because that's why:
image

Please hotfix.

marc-hb:
Thanks for spotting this, tentative fix submitted in:
- #7211

PR #7206, opened 2023-03-03
dai-zephyr: remove no use functions in dai zephyr

btian1:
Function: dai_comp_hw_params does not do anything and remove it.


lyakh (approved):

kv2019i (approved):
Adding @juimonen in case something was planned for dai_comp_hw_params().

juimonen (approved):
No plans from my side. If it is not used, it is not used. I think this was lifted from older ipc3 code base. We can add it back if needed.

lgirdwood (approved):

Issue #6734, opened 2022-12-07
[BUG] FW reported error: 7 for 0x13000004|0x1 when multiple-pipeline or multiple-pause-resume on MTLP_RVP_NOCODEC


miRoox:
**Describe the bug**
CI observed an IPC error "FW reported error: 7 - Unsupported operation requested" for 0x13000004|0x1 when testing multiple-pipeline or multiple-pause-resume on MTLP_RVP_NOCODEC.

runtime PM and IMR was disabled during the test.

**To Reproduce**
```
TPLG=/lib/firmware/intel/sof-ace-tplg/sof-mtl-nocodec.tplg MODEL=MTLP_RVP_NOCODEC ~/sof-test/test-case/multiple-pause-resume.sh -r 50
```
or
```
TPLG=/lib/firmware/intel/sof-ace-tplg/sof-mtl-nocodec.tplg MODEL=MTLP_RVP_NOCODEC ~/sof-test/test-case/multiple-pipeline.sh -f p -c 20 -l 50
```

**Reproduction Rate**
About 50%

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel: https://github.com/thesofproject/linux/commit/c78e584e7704
    * SOF: https://github.com/thesofproject/sof/commit/4a9a99aeb24a
2) Name of the topology file
    * Topology: sof-ace-tplg/sof-mtl-nocodec.tplg
3) Name of the platform(s) on which the bug is observed.
    * Platform: MTLP_RVP_NOCODEC 

**Screenshots or console output**

**dmesg**
```
[ 1029.026338] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx      : 0x13000004|0x1: GLB_SET_PIPELINE_STATE [data size: 12]
[ 1029.026868] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx reply: 0x33000007|0x1: GLB_SET_PIPELINE_STATE
[ 1029.026872] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: FW reported error: 7 - Unsupported operation requested
[ 1029.027039] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc error for msg 0x13000004|0x1
[...]

PR #7207, opened 2023-03-03
ipc4: copier: Clarify the log and return value when the dai_copier is…

ujfalusi:
… not found

We are always returning 0 if the dai_copier is not found and never an error code. Make it clear by returning 0.

With that out of the way, it is clear that this is not an error, it is something which can happen.
Firmware log is a proof of that since it is flooded with <err> copier: comp:X 0xSOMEHEX failed to find dai comp or sink pipeline not running.

prints. Drop the log level to warn (but I think it could be even dbg


marc-hb:
Thanks Peter. This `<err>` has been spamming test logs with hundreds and even thousands of lines in long, daily test runs which makes it impossible to spot other, real errors.

- https://github.com/thesofproject/sof/issues/6734#issuecomment-1452731165

RanderWang (approved):

abonislawski (approved):

lyakh (commented):
Let me try to understand why that path is hit, hopefully today

src/audio/copier/copier.c
```diff
@@ -950,8 +950,8 @@ static int copier_comp_trigger(struct comp_dev *dev, int cmd)
 
 	dai_copier = pipeline_get_dai_comp_latency(dev->pipeline->pipeline_id, &latency);
 	if (!dai_copier) {
-		comp_err(dev, "failed to find dai comp or sink pipeline not running.");
-		return ret;
+		comp_warn(dev, "failed to find dai comp or sink pipeline not running.");
+		return 0;
```
lgirdwood, line 954: Ok, so this looks suspect without an inline comment. Can you add one.

lgirdwood (commented):
Need a comment on wht we return 0 on a potential error.

marc-hb:
Superseded by #7210?

ujfalusi:
> Superseded by this?
> 
>     * [Fix "failed to find dai comp or sink pipeline not running." #7210](https://github.com/thesofproject/sof/pull/7210)

This PR is still valid with that PR as we would still return 0 and print with err level.

src/audio/copier/copier.c
ujfalusi: @lyakh, would you add a patch to your #7210 to change the return value to some error code as  a last step? This should not happen now afaik, so if there is no copier found it is an error?
Then I can close this PR.

ujfalusi:
Changes since v1:
- drop to info log level
- reword the printed information and add additional comment for clarity
- with #7210 this is not going to happen

lgirdwood (approved):

PR #7209, opened 2023-03-03
copier: fix format check for set_attenuation

pblaszko:
Fix format validation in set_attenuation function.

In previus implementation, bitdepth was compared to sink->buffer_fmt value, which is an enum sof_ipc_buffer_format (interleaved or not).

Format bitdepth is represented by stream.frame_fmt (enum sof_ipc_frame), so it should be compared to this value.


lgirdwood (approved):

RanderWang (approved):

tmleman (approved):

pblaszko:
SOFCI TEST

kv2019i (approved):

PR #7210, opened 2023-03-03
Fix "failed to find dai comp or sink pipeline not running."

lyakh:
The "failed to find dai comp or sink pipeline not running." error message has been bugging SOF developers since a long time without causing any apparent problems. Finally root-cause it and fix the underlying problem.

### DIFF
--- a/src/audio/pipeline/pipeline-graph.c
+++ b/src/audio/pipeline/pipeline-graph.c
@@ -486,7 +486,7 @@ struct comp_dev *pipeline_get_dai_comp(uint32_t pipeline_id, int dir)
 }
 
 #if CONFIG_IPC_MAJOR_4
-/* visit connected pipeline to find the dai comp and latency.
+/* Playback only: visit connected pipeline to find the dai comp and latency.
  * This function walks down through a pipelines chain looking for the target dai component.
  * Calculates the delay of each pipeline by determining the number of buffered blocks.
  */
@@ -521,8 +521,6 @@ struct comp_dev *pipeline_get_dai_comp_latency(uint32_t pipeline_id, uint32_t *l
 		/* Calculate pipeline latency */
 		input_data = comp_get_total_data_processed(source, 0, true);
 		output_data = comp_get_total_data_processed(ipc_sink->cd, 0, false);
-		if (!input_data || !output_data)
-			return NULL;
 
 		ret = comp_get_attribute(source, COMP_ATTR_BASE_CONFIG, &input_base_cfg);
 		if (ret < 0)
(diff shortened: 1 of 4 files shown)


marc-hb:
Only one device ran tests in https://sof-ci.01.org/sofpr/PR7210/build4295/devicetest/index.html

There is one failure but no logs in QB build/11533585. @wszypelt any idea why there is no error message?

marc-hb:
SOFCI TEST

marc-hb:
https://sof-ci.01.org/sofpr/PR7210/build4298/build/tools.txt failed because of some docker issue, @fredoh9 will look into it

```
+ docker run -i -v /srv/home/jenkins/workspace/sof_generic_build/sof:/home/sof/work/sof.git -v /srv/home/jenkins/workspace/sof_generic_build/sof:/home/sof/work/sof-bind-mount-DO-NOT-DELETE --env CMAKE_BUILD_TYPE --env PRIVATE_KEY_OPTION --env USE_XARGS --env NO_PROCESSORS --env VERBOSE --env http_proxy=http://proxy.jf.intel.com:911 --env https_proxy=http://proxy.jf.intel.com:912 --user 1015 thesofproject/sof ./sof/scripts/build-tools.sh
docker: Error response from daemon: failed to create shim: OCI runtime create failed: runc create failed: unable to start container process: exec: "./sof/scripts/build-tools.sh": stat ./sof/scripts/build-tools.sh: no such file or directory: unknown.
time="2023-03-03T16:17:04-08:00" level=error msg="error waiting for container: context canceled"
```

marc-hb:
SOFCI TEST

src/audio/pipeline/pipeline-graph.c
```diff
@@ -532,7 +530,9 @@ struct comp_dev *pipeline_get_dai_comp_latency(uint32_t pipeline_id, uint32_t *l
 		if (ret < 0)
 			return NULL;
 
-		*latency += input_data / input_base_cfg.ibs - output_data / output_base_cfg.obs;
+		if (input_data && output_data)
```
RanderWang, line 533: if  comp_get_attribute is not support by source or sink, the input_base_cfg & output_base_cfg are not initialized. There will be zero-division exception when input_base_cfg & output_base_cfg  are zero filled

ujfalusi: if the `comp_get_attribute` is not supported then the base_cfg will be not touched, it will remain un-initialized, likely 0, but by definition it is not safe to touch it.
Which means that if the `comp_get_attribute()` returns 0, the value which is passed uninitialized is not safe to touch, 0 means **success** or **untouched**.

Therefore, `input_base_cfg` and `output_base_cfg` must be initialized to either 0 and have a check for ibs and obs.

lyakh: agree, good catch. This is a bug that already exists on "main" independent of this PR. A fix added to the PR

lgirdwood (approved):

lgirdwood:
@lyakh can you check CI thanks !

lyakh:
@wszypelt do I understand it correctly that the QB failed a keyword-detect test? Does it have playback pipelines? This PR only touches playback code.

RanderWang (approved):
there are at least two pipelines for normal playback stream. Sometime the pipeline with the host module will be created first, then dai pipeline, so we can't find dai for this case since dai pipeline is not  created.

RanderWang:
> 



> > there are at least two pipelines for normal playback stream. Sometime the pipeline with the host module will be created first, then dai pipeline, so we can't find dai for this case since dai pipeline is not created.
> 
> @RanderWang do you have such current examples with our pipelines? I think pipelines are now created by driver IPCs always in the same deterministic order, so it cannot happen that the order varies between boots.

@lyakh Yes, Linux driver changed logic then this type of error disappears.  I don't have such example now.

lgirdwood:
SOFCI TEST

lgirdwood:
@lyakh still seeing a CI failure. Do we need to split this into 3 PRs ?

PR #7217, opened 2023-03-06
[DNM] zephyr: start a 1000Hz k_timer at DSP boot

kv2019i:
Emits a line of trace every 1ms even when no audio pipelines are created.

This is intended for debugging purposes.


sys-pt1s:

PR #7227, opened 2023-03-07
Gain no name prefix

jsarha:
"topology2: gain: Add "name_no_prefix" attribute to gain class" the main commit here. "topology2: cavs-mixin-mixout-hda: Rename mixin-mixout volumes, no prefix"  is an example how to use the new feature (and the old variable value expansion), but in it self its a serious attempt to fix the mixer naming problems found in cavs-mixin-mixout-hda.conf.

The associated Linux driver PR is needed this PR to work.

### DIFF
--- a/tools/topology/topology2/include/common/tokens.conf
+++ b/tools/topology/topology2/include/common/tokens.conf
@@ -52,6 +52,7 @@ Object.Base.VendorToken {
 		curve_type		260
 		curve_duration		261
 		init_value		262
+		no_wname_in_kcontrol_name	263
 	}
 
 	"6" {
--- a/tools/topology/topology2/include/components/gain.conf
+++ b/tools/topology/topology2/include/components/gain.conf
@@ -68,6 +68,18 @@ Class.Widget."gain" {
                 token_ref       "gain.word"
 	}
 
+	DefineAttribute."no_wname_in_kcontrol_name" {
+		type	"string"
+		# Token set reference name
+		token_ref	"gain.bool"
+		constraints {
+			!valid_values [
+				"true"
+				"false"
+			]
+		}
+	}
+
 	# Attribute categories
 	attributes {
 		#
@@ -156,4 +168,5 @@ Class.Widget."gain" {
 	init_value		0x7fffffff
 	num_input_pins		1
 	num_output_pins		1
+	no_wname_in_kcontrol_name	"true"
 }
--- a/tools/topology/topology2/include/components/eqiir.conf
+++ b/tools/topology/topology2/include/components/eqiir.conf
@@ -61,4 +61,5 @@ Class.Widget."eqiir" {
 	no_pm			"true"
 	num_input_pins		1
 	num_output_pins		1
+	no_wname_in_kcontrol_name	"true"
 }


tools/topology/topology2/include/components/gain.conf
```diff
@@ -82,6 +82,20 @@ Class.Widget."gain" {
 		token_ref	"sof_tkn_comp.word"
 	}
 
+	# 'normal' = add widget name as prefix to gain name
+	# 'no_prefix' = use gain name as it is
+	DefineAttribute."name_no_prefix" {
```
lgirdwood, line 87: I guess this wont work in the manifest but has to become per component ?

lgirdwood (commented):
@plbossart fyi.

tools/topology/topology2/include/components/gain.conf
jsarha: Oh, I forgot the old comments there. The property was firs an enum, but the I decided boolean is better. I'll remove those.

jsarha: > I guess this wont work in the manifest but has to become per component ?

Yes, but its easy to make it default behavior. This is what I was trying to accomplish all the time. Then when I started to study the manifest approach, I bumped into this way of doing it.

tools/topology/topology2/include/common/tokens.conf
```diff
@@ -52,6 +52,7 @@ Object.Base.VendorToken {
 		curve_type		260
 		curve_duration		261
 		init_value		262
+		use_topology_name	263
```
lgirdwood, line 55: Which other modules do we need to add this attribute too ?

lgirdwood (commented):
@plbossart pls review.

tools/topology/topology2/include/common/tokens.conf
jsarha: Are there any other modules creating alsa mixers? The PCM names can be independently named already so I do not see a need for anything else.

jsarha:
Set ready for review to get more attention.

jsarha:
This PR [1] helped me notice that I actually had the pre and post mixer gain names the wrong way around. Now this is fixed and here is how the topology looks now in tplgtool2.py produced picture when the mentioned PR is in use:
sof-hda-generic

And here is how "amixer -c 0 controls" output looks like:
numid=11,iface=CARD,name='HDMI/DP,pcm=3 Jack'
numid=17,iface=CARD,name='HDMI/DP,pcm=4 Jack'
numid=23,iface=CARD,name='HDMI/DP,pcm=5 Jack'
numid=10,iface=CARD,name='Line Out Jack'
numid=9,iface=CARD,name='Mic Jack'
numid=2,iface=MIXER,name='Master Playback Switch'
numid=1,iface=MIXER,name='Master Playback Volume'
numid=8,iface=MIXER,name='Mic Boost Volume'
numid=5,iface=MIXER,name='Mic Playback Switch'
numid=4,iface=MIXER,name='Mic Playback Volume'
numid=7,iface=MIXER,name='Capture Switch'
numid=6,iface=MIXER,name='Capture Volume'
numid=3,iface=MIXER,name='Loopback Mixing'
numid=12,iface=MIXER,name='IEC958 Playback Con Mask'
numid=18,iface=MIXER,name='IEC958 Playback Con Mask',index=1
numid=24,iface=MIXER,name='IEC958 Playback Con Mask',index=2
numid=13,iface=MIXER,name='IEC958 Playback Pro Mask'
numid=19,iface=MIXER,name='IEC958 Playback Pro Mask',index=1
numid=25,iface=MIXER,name='IEC958 Playback Pro Mask',index=2
numid=14,iface=MIXER,name='IEC958 Playback Default'
numid=20,iface=MIXER,name='IEC958 Playback Default',index=1
numid=26,iface=MIXER,name='IEC958 Playback Default',index=2
numid=15,iface=MIXER,name='IEC958 Playback Switch'
numid=21,iface=MIXER,name='IEC958 Playback Switch',index=1
numid=27,iface=MIXER,name='IEC958 Playback Switch',index=2
numid=33,iface=MIXER,name='Post Mixer Analog Playback Volume'
numid=32,iface=MIXER,name='Pre Mixer Analog Playback Volume'
numid=34,iface=MIXER,name='eqiir.4.1 4 Main capture Iir Eq'
numid=35,iface=MIXER,name='gain.15.1 Deepbuffer Volume'
numid=16,iface=PCM,name='ELD',device=3
numid=29,iface=PCM,name='Playback Channel Map',device=3
numid=22,iface=PCM,name='ELD',device=4
numid=30,iface=PCM,name='Playback Channel Map',device=4
numid=28,iface=PCM,name='ELD',device=5
numid=31,iface=PCM,name='Playback Channel Map',device=5

[1] https://github.com/thesofproject/sof-test/pull/1014

tools/topology/topology2/include/components/gain.conf
```diff
@@ -171,4 +183,5 @@ Class.Widget."gain" {
 	init_value		0x7fffffff
 	num_input_pins		1
 	num_output_pins		1
+	name_no_prefix		"false"
```
plbossart, line 186: not following, the first part of this patch adds 'use_topology_name' but this sets 'name_no_prefix'

I don't see the link between the two.

tools/topology/topology2/cavs-mixin-mixout-hda.conf
```diff
@@ -54,8 +55,9 @@ Object.Pipeline {
 			}
 
 			Object.Widget.gain.1 {
+				use_topology_name "true"
 				Object.Control.mixer.1 {
-					name '1 2nd Playback Volume'
+					name 'Pre Mixer $ANALOG_PLAYBACK_PCM Volume'
```
plbossart, line 60: pre- and post mix is fine, but I don't get what ANALOG means here?

sys-pt1s:

tools/topology/topology2/cavs-mixin-mixout-hda.conf
jsarha: That refers to the PCM this pipeline is going to.

tools/topology/topology2/include/components/gain.conf
jsarha: The both mean "do not add the widget name as a prefix to the kcontrol name that is visible in the alsa mixer, e.g. use the topology name only". However, on the Linux side referring to topology is not very accurate as the widget may not be coming from a topology file at all, but from some ASoC component driver for instance.

lgirdwood:
@ranj063 any inputs ?

lgirdwood:
@jsarha @ranj063 any updates on alignment here ?

jsarha:
> @jsarha @ranj063 any updates on alignment here ?

I am just hoping someone would review this or say that I am on a wrong track. @ranj063, @plbossart , @juimonen what do you say?

jsarha:
Oh, just noticed that there was a conflict. Rebased the code and fixed it so all should be good for reviewing again.

tools/topology/topology2/include/common/tokens.conf
plbossart: "use_topology_name" seems to mean "do_not_add_prefix_use_topology_name_as_is" 

It's not a replacement of a name by a topology-defined one, more an enforcement to use as is.

Did I get this right? if yes, the naming is questionable.

tools/topology/topology2/include/components/gain.conf
plbossart: Now I am also lost, we have both name_no_prefix and use_topology_name, but they are the same concept?

Why can't we have just one token added?

plbossart (changes requested):
Just as lost as the matching kernel PR @jsarha, sorry.

tools/topology/topology2/include/common/tokens.conf
jsarha: Yes, you got it. I try to come up with a better name.

tools/topology/topology2/include/components/gain.conf
jsarha: Ok, I go back to name_no_prefix everywhere.

jsarha:
@plbossart and @ranj063 , new version pushed. Tried to address Pierre's comments.

tools/topology/topology2/include/components/gain.conf
ranj063: @jsarha shouldn't this be a mixer control attribute instead of a gain? WHat happens if different component has a mixer control?

jsarha:
Ranjani pointed out quite correctly that the name_no_prefix should come from topoplogy mixer object not from gain object. After banging my head against  the wall for two days and still not getting the mixer attribute to work, I decided that its time to push the no_prefix part back on proceed with the mixer renaming separately. There is a new PR here: https://github.com/thesofproject/sof/pull/7641

I will continue with this one on lower priority, in parallel with the mixer renaming part on higher priority.

jsarha:
@plbossart and @ranj063 there is now a new version of this PR. Not much has changed on the topology side. Just dropped the mixer name changes, changed the gain attribute name to more specific, and changed its value to default to true.

However, now that I am this far I think that we probably need this property for IirEQ, FirEQ, and SmartAmp widgets too (are there more modules with associated kcontrols), to make the kcontrol names uniform. Should probably make another PR for changing the names of those entities to follow the scheme we now use in the mixer names.

PR #7233, opened 2023-03-08
platform: mtl: lowest clock as default

tmleman:
This patch sets lowest clock as default for Meteorlake platform. In current version FW can do unnecessary clocks switching operations. FW should start from low value and scale-up if necessary.


marcinszkudlinski (approved):

kv2019i:
We can't merge this until the KCPS_REGISTER stuff is merged to kernel and topologies and we have sane values. @jsarha @ujfalusi are we ready?

jsarha:
> We can't merge this until the KCPS_REGISTER stuff is merged to kernel and topologies and we have sane values. @jsarha @ujfalusi are we ready?
> 
> E.g. [thesofproject/linux#4213](https://github.com/thesofproject/linux/pull/4213) and friends.

https://github.com/thesofproject/sof/pull/7160 is merged, but it currently only has the 0 kcps defined as the default value for all pipelines.

abonislawski:
This PR does not need any KCPS stuff. Our default clock is the lowest one, thats why we have to force HP clock in init (in current code).


> #7160 is merged, but it currently only has the 0 kcps defined as the default value for all pipelines.

And we should keep it at 0. This is calculated by FW based on manifest data (currently only on MTL-004 branch).
Driver should use kcps_register to adjust this **if necessary**.

kv2019i (changes requested):
@abonislawski wrote:
> And we should keep it at 0. This is calculated by FW based on manifest data (currently only on MTL-004 branch).

Right, so it's 0 in tplg and fw doesn't calculate so this patch will break MTL in main as DSP clock will be kept at lowest level? -1 until this is clarified. Shouldn't the calculation basd on FW data (or higher default in tplg) be merged first, before we do this PR.

> Driver should use kcps_register to adjust this if necessary.

Hmm, how will topology authors know when it is necessary? This seems very fuzzy to me and we need to have this documented somewhere else then review comments..

abonislawski (changes requested):
yeap, actually we cant merge it now because temporary HP switch in init will not update clk info so its better to keep it as is for now

lgirdwood:
Assigning milestone so this can be tracked and merged once dependencies are all in place.

tmleman:
Since it will not gonna be merge in the near future I'm closing this PR.

PR #7224, opened 2023-03-07
dai: host: add fast mode

makarukp:
Add fast mode which enables to support unlimited transfer size (more than
ibs/obs) between a dma buffer and the copier. Enable it to satisfy
specific module requirements to guarantee enough data size to decoders or
from encoders since the pcm audio period size does not apply to variable
size frame.

PR #7234, opened 2023-03-08
Topo2 fast mode

jsarha:
This PR is waiting for Linux side PR to be accepted and merged: https://github.com/thesofproject/linux/pull/4229

### DIFF
--- a/tools/topology/topology2/include/components/copier.conf
+++ b/tools/topology/topology2/include/components/copier.conf
@@ -147,6 +147,15 @@ Class.Widget."copier" {
 		token_ref	"sof_tkn_copier.word"
 	}
 
+	#
+	# In fast mode copier does unlimited data transfers not limited to LL scheduling
+	# tick (typically 1ms).
+	#
+	DefineAttribute."fast_mode" {
+		# Token reference and type
+		token_ref	"sof_tkn_copier.bool"
+	}
+
 	attributes {
 		#
 		# The copier widget name would be constructed using the copier type, index and
--- a/tools/topology/topology2/include/common/tokens.conf
+++ b/tools/topology/topology2/include/common/tokens.conf
@@ -178,5 +178,6 @@ Object.Base.VendorToken {
 		name "sof_tkn_copier"
 		node_type		1980
 		deep_buffer_dma_ms	1981
+		fast_mode		1982
 	}
 }
--- a/tools/topology/topology2/include/pipelines/cavs/deepbuffer-playback.conf
+++ b/tools/topology/topology2/include/pipelines/cavs/deepbuffer-playback.conf
@@ -53,6 +53,7 @@ Class.Pipeline."deepbuffer-playback" {
 			num_input_audio_formats 3
 			num_output_audio_formats 3
 			deep_buffer_dma_ms	$DEEPBUFFER_FW_DMA_MS
+			fast_mode		true
 			# 16-bit 48KHz 2ch
 			Object.Base.audio_format.1 {
 				out_bit_depth		32


jsarha:
I do not follow why this works the way it does, but I noticed that "fast_mode" attribute was almost symmetrical to "deep_buffer_dma_ms" attribute. There were only two difference, "deep_buffer_dma_ms" is word and "fast_mode" is boolean, "deep_buffer_dma_ms" has no default value, but "fast_mode" has. Changing "fast_mode" to word did not help, but removing the "false" default value from copier.conf did. Do we have a bug somewhere? Anyway, the two PRs now work together as I intended them.

jsarha:
So this and https://github.com/thesofproject/linux/pull/4229 are mean to be used with https://github.com/thesofproject/sof/pull/7224/ . AFAIK, there is still some technical problems with firmware changes, so better wait until all three work correctly together.

tools/topology/topology2/include/components/copier.conf
```diff
@@ -147,6 +147,18 @@ Class.Widget."copier" {
 		token_ref	"sof_tkn_copier.word"
 	}
 
+	DefineAttribute."fast_mode" {
+		type	"string"
+		# Token set reference name
+		token_ref	"sof_tkn_copier.bool"
+		constraints {
+			!valid_values [
```
aiChaoSONG, line 155: The bool value is a subset of integer in C lang and topology, assign "true" or "false" string to it is not going to work. the below code will allow us to write something like `mix_type    mix_in`, but in topology generation, mix_type get the value 0.
```
	DefineAttribute."mix_type" {
		type "string"
                token_ref       "sof_tkn_mixinout.word"
                constraints {
                        !valid_values [
                                "mix_in"
                                "mix_out"
                        ]
                        !tuple_values [
                                0
                                1
                        ]
                }
	}
```

lgirdwood (commented):
@jsarha any update ?

tools/topology/topology2/include/components/copier.conf
```diff
@@ -147,6 +147,18 @@ Class.Widget."copier" {
 		token_ref	"sof_tkn_copier.word"
 	}
 
+	DefineAttribute."fast_mode" {
```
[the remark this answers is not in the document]
jsarha: Updated with your suggestion.

tools/topology/topology2/include/components/copier.conf
aiChaoSONG: @jsarha Are you sure the fast_mode can be assigned correctly with 
```
!valid_values [
    "true"
    "false"
]
```
instead of
```
!valid_values [
    "true"
    "false"
]
!tuple_values [
   1
   0
]
```

jsarha: The original is not the first instance of notation I used: https://github.com/thesofproject/sof/blob/1319163ea2c6d8bea6fd7189dea6275ac5e34bf5/tools/topology/topology2/include/dais/ssp.conf#L78

However, I think the constraint section is not really needed. There are other attributes declared without it.

lgirdwood (changes requested):

sys-pt1s:

tools/topology/topology2/sof-hda-generic.conf
```diff
@@ -1,3 +1,10 @@
+Object.Base.manifest.1 {
+	name "sof_manifest"
+        Object.Base.data.1 {
```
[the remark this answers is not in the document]
jsarha: Hack dropped.

lgirdwood (approved):
@jsarha @ranj063 this looks fine to me for topology, is this now good for FW/kernel or are there still opens ?

kv2019i:
@jsarha This seems stuck, converting to draft. Please change back to ready when this is resumed.

jsarha:
Close an obsolete PR.

PR #7237, opened 2023-03-09
topology2: add google rtc aec support

RanderWang:


tools/topology/topology2/include/components/aec.conf
```diff
@@ -0,0 +1,160 @@
+#
+# Common widget aec
+#
+# A generic aec widget. All attributes defined herein are namespaced
+# by alsatplg to "Object.Widget.aec.N.attribute_name"
```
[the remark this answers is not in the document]
plbossart: indeed, we should NEVER talk about AEC in general. This has to be a specific type of AEC....

tools/topology/topology2/include/components/aec.conf
```diff
@@ -0,0 +1,160 @@
...
+#	}
+#
+# Where N is the unique instance number for the aec object within the same alsaconf node.
+Define {
+	AEC_UUID	"a6:a0:80:b7:9f:26:6f:46:b4:77:23:df:a0:5a:f7:58"
+}
+
+Class.Widget."aec" {
+	#
+	# Pipeline ID for the aec widget object
```
plbossart, line 21: the AEC uses two sources (reference + capture), and has one sink pin. What would the pipeline look like here. Is this a pipeline with a single module?

tools/topology/topology2/include/components/aec.conf
```diff
@@ -0,0 +1,160 @@
...
+	DefineAttribute."is_pages" {
+		# Token set reference name and type
+		token_ref	"sof_tkn_comp.word"
+	}
+
+	# Attribute categories
+	attributes {
+		#
+		# The aec widget name would be constructed using the index and instance attributes.
+		# For ex: "aec.1.1" or "aec.10.2" etc.
```
plbossart, line 61: no. we absolutely want to avoid generic terms such as 'aec'. It has to be clearly linked with the 3rd party or library used.

tools/topology/topology2/include/components/aec.conf
```diff
@@ -0,0 +1,160 @@
...
+			out_ch_map      $CHANNEL_MAP_3_POINT_1
+		}
+		4 {
+			pin_index               1
+			in_channels             4
+			out_channels            4
+			in_ch_cfg       $CHANNEL_CONFIG_QUATRO
+			in_ch_map       $CHANNEL_MAP_3_POINT_1
+			out_ch_cfg      $CHANNEL_CONFIG_QUATRO
+			out_ch_map      $CHANNEL_MAP_3_POINT_1
```
plbossart, line 125: 4 channels for echo cancellation? What does this mean? Are we talking about 4 channels for the reference, for the captured data, both, none?

tools/topology/topology2/include/components/aec.conf
```diff
@@ -0,0 +1,160 @@
...
+		}
+	}
+
+	# Default attribute values for aec widget
+	# UUID: B780A0A6-269F-466F-B477-23DFA05AF758
+	uuid			$AEC_UUID
+	type			"effect"
+	no_pm 			"true"
+	core_id			0
+	cpc 			5000
```
plbossart, line 156: ??? how is this defined?

tools/topology/topology2/include/components/aec.conf
```diff
@@ -0,0 +1,160 @@
...
+			out_bit_depth           16
+			out_valid_bit_depth     16
+			in_ch_cfg       $CHANNEL_CONFIG_QUATRO
+			in_ch_map       $CHANNEL_MAP_3_POINT_1
+			out_ch_cfg      $CHANNEL_CONFIG_QUATRO
+			out_ch_map      $CHANNEL_MAP_3_POINT_1
+		}
+	}
+
+	# Default attribute values for aec widget
```
plbossart, line 150: which type?

plbossart (changes requested):
AEC is not a generic solution, there's just too many variants.

The other concern is that this is starting to be too much for a nocodec topology, which has become a dumping ground. We are not going to add all possible variants of aec, smart-amp, wov in nocodec.conf. We need to revisit the goals and how tests will be handled.

NAK NAK NAK.

tools/topology/topology2/include/components/aec.conf
RanderWang: It is a common value in widget definition. As I know cSOF doesn't have a complete solution for cpc used in fw for schedule.  Take ref fw for example,  the cpc is zero in mod_cfg and is built from CPS (_Number of DSP cycles consumed by the module instance to process one second of data_)  in windows driver.  CPS should be generated from statistic in FW for different stream format. We can define a accurate cpc when cSOF has a solution in fw for cpc based schedule.

        [[module.entry]]
        name = "GAIN"
        uuid = "61BCA9A8-18D0-4A18-8E7B-2639219804B7"
        affinity_mask = "0x1"
        instance_count = "40"
        domain_types = "0"
        load_type = "0"
        module_type = "9"
        auto_start = "0"
        sched_caps = [1, 0x00008000]

        # pin = [dir, type, sample rate, size, container, channel-cfg]
        pin = [0, 0, 0xfeef, 0xf, 0xf, 0x45ff,
                        1, 0, 0xfeef, 0xf, 0xf, 0x45ff]

        mod_cfg = [0, 0, 0, 0, 416, 914000, 48, 64, 0,                      0, 0,
                   1, 0, 0, 0, 416, 1321600, 192, 256, 0,                   0, 0,
                   2, 0, 0, 0, 416, 1786000, 192, 256, 0,                   0, 0,
                   3, 0, 0, 0, 416, 2333000, 48, 64, 0,                     0, 0,

tools/topology/topology2/include/components/aec.conf
RanderWang: It is for future. If we don't have such requirement, I will remove it.

tools/topology/topology2/include/components/aec.conf
RanderWang: We can define different UUID for different solution. Currently it is for google rtc aec.

plbossart: it's not a question of UUIDs only, all the formats and capabilities will likely be different between AlgorithmA and AlgorithmZ. I don't see what kind of generic solution you can expect. This is really google-rtc-aec.conf.

tools/topology/topology2/include/components/aec.conf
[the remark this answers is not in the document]
RanderWang: sure. I first used rtc aec but later I though we can use single aec parameters definition for different variants.

tools/topology/topology2/include/components/aec.conf
RanderWang: sure, thanks

tools/topology/topology2/include/components/aec.conf
RanderWang: Defined two pins in this file. And the binding info is defined in pipeline

tools/topology/topology2/include/components/aec.conf
RanderWang: sure,thanks

tools/topology/topology2/cavs-nocodec.conf
```diff
@@ -463,131 +408,6 @@ Object.Pipeline.passthrough-capture-be [
 	}
 ]
 
-Object.Pipeline.gain-capture [
-	{
-		format		$FORMAT
-		index		18
-		Object.Widget.pipeline.1 {
-			stream_name "DMIC0 Raw"
-		}
```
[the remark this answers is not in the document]
RanderWang: This was introduced to prepare for AEC. I will not change nocodec.conf

tools/topology/topology2/include/components/google-rtc-aec.conf
```diff
@@ -0,0 +1,126 @@
...
+			pin_index               0
+		}
+		2 {
+			pin_index               1
+		}
+	}
+
+	Object.Base.output_audio_format {
+		1 {
+		}
```
plbossart, line 113: can someone educate me on the purpose of declaring empty formats? @RanderWang @ranj063 ?

tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf
```diff
@@ -0,0 +1,112 @@
+#
```
plbossart, line 1: we should never assume how an element is connected.

Specifically in this case, it's possible that there is a noise-suppression inserted before the host copier, or that there are two input paths, one with a direct host copier and one with the noise suppression.

I think we need to leave the choice of how an algorithm is inserted to the top-level topology. Don't introduce restrictions when we already know they don't always hold.

plbossart (changes requested):
not ok with the assumption that the AEC is directly connected to a copier. This may or may not be true if noise suppression comes into play. That's why we introduced the 3 input paths (raw, AEC, AEC+NS).

In other words, we need to defer the connections to the very top level and not create artificial restrictions we'll have to undo later.

Adding @cujomalainey and @singalsu for additional comments.

tools/topology/topology2/include/components/google-rtc-aec.conf
RanderWang: empty declaring means it will use default setting, like 48k, 16bits, 2ch

tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf
RanderWang: @plbossart  In current plan, we have three pipelines for output:  (1) raw (2) AEC (3) AEC+NS. This PR focus on pipeline (2) and if we have NS module, we will change one of AEC pipeline to AEC+NS. 

Do you means you prefer to merage (2) and (3) with some flag ?

singalsu:
@RanderWang Can you share a tplgtool.py created picture of this topology?

cujomalainey:
> Adding @cujomalainey and @singalsu for additional comments.

@plbossart thanks. @thesofproject/google should be CCed on all things related to our AEC and related components. Thanks

tools/topology/topology2/include/components/google-rtc-aec.conf
```diff
@@ -0,0 +1,126 @@
...
+# Common widget google-rtc-aec
+#
+# A generic google-rtc-aec widget. All attributes defined herein are namespaced
+# by alsatplg to "Object.Widget.google-rtc-aec.N.attribute_name"
+#
+# Usage: this component can be used by instantiating it in the parent object. i.e.
+#
+# 	Object.Widget.google-rtc-aec."N" {
+# 		index		1
+# 		type		"effect"
```
cujomalainey, line 11: just curious why `type` is needed at the top level? Isn't it implied by the class?

tools/topology/topology2/include/components/google-rtc-aec.conf
```diff
@@ -0,0 +1,126 @@
...
+	# google-rtc-aec object instance
+	#
+	DefineAttribute."instance" {}
+
+	#include common component definition
+	<include/components/widget-common.conf>
+
+	DefineAttribute."num_audio_formats" {
+		# Token set reference name and type
+		token_ref	"sof_tkn_comp.word"
```
cujomalainey, line 35: what does this mean?

tools/topology/topology2/include/components/google-rtc-aec.conf
```diff
@@ -0,0 +1,126 @@
...
+	#
+	# google-rtc-aec widget UUID
+	#
+	DefineAttribute."uuid" {
+		type	"string"
+		# Token set reference name and type
+		token_ref	"sof_tkn_comp.uuid"
+	}
+
+	DefineAttribute."cpc" {
```
cujomalainey, line 47: what is cpc? Can we really get some docs on these new formats? its really frustrating trying to validate this against out code without context. There is nothing in the docs website or the existing usages.

tools/topology/topology2/include/components/google-rtc-aec.conf
```diff
@@ -0,0 +1,126 @@
...
+		# Token set reference name and type
+		token_ref	"sof_tkn_comp.word"
+	}
+
+	DefineAttribute."is_pages" {
```
cujomalainey, line 52: same as above

tools/topology/topology2/include/components/google-rtc-aec.conf
cujomalainey: Is there a way to allow ranges rather than distinct tuples? Also how does this work when pins may have different formats?

tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf
plbossart: @RanderWang not following what you meant by "we will change one of AEC pipeline to AEC+NS".

It's the same AEC in both cases, no? This is a hierarchical set of paths.

plbossart: This is my understanding of the ask:

````
DMIC ---> copier ---> host
                   |
                   +------> AEC -----> copier -----> host
                                 |
                                 +----------> NS -------> copier -----------> host
````
````

File: tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf [hunk 14]
```diff
@@ -0,0 +1,112 @@
...
+					tlv_callback
+				]
+
+				Object.Base.extops."extctl" {
+					get 258
+					put 0
+				}
+				max 4096
+				Object.Base.data."google-rtc-aec_blob" {
+					bytes "
```
cujomalainey, line 76: any chance this can be included for cleanliness?

tools/topology/topology2/platform/intel/google-rtc-aec-generic.conf
```diff
@@ -0,0 +1,156 @@
+Object.Pipeline.gain-module-copier [
```
cujomalainey, line 1: context on what `generic` means?

cujomalainey (changes requested):
I'll reinforce this again as in my other PR. I will not approve commits that relate to google code if I don't have the documentation to approve what I am reviewing. **please prioritize these documents**

tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf
RanderWang: thanks.  My understanding is.   The advantage is that it will skip 3 copiers to improve performance.
````
Copier -------> DMIC -----> host
   |----------------> AEC ------> host
   |------> AEC ------> NS---> host
````

**Comment by RanderWang** [comment]
> I'll reinforce this again as in my other PR. I will not approve commits that relate to google code if I don't have the documentation to approve what I am reviewing. **please prioritize these documents**

According to the question, a documents of topology2 is needed. @lgirdwood  this is a big task

File: tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf (hunk 9 above)
Comment (RanderWang) [reply tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf:1]: do we need gain for AEC pipeline ?

Comment (RanderWang) [reply tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf:1]: > @RanderWang not following what you meant by "we will change one of AEC pipeline to AEC+NS".
> 
> It's the same AEC in both cases, no? This is a hierarchical set of paths.

Currently we don't have NS support so I create two same pipelines with AEC and will add NS support in one of them if NS is ready.

**Comment by RanderWang** [comment]
> @RanderWang Can you share a tplgtool.py created picture of this topology?

@singalsu  Do we need gain for each pipeline ? if not we can remove them and one copier module so that we have better performance
sof-mtl-max98357a-rt5682-rtc-aec

File: tools/topology/topology2/include/components/google-rtc-aec.conf (hunk 11 above)
Comment (RanderWang) [reply tools/topology/topology2/include/components/google-rtc-aec.conf:35]: it means num_audio_formats is a type of word

File: tools/topology/topology2/include/components/google-rtc-aec.conf (hunk 12 above)
Comment (RanderWang) [reply tools/topology/topology2/include/components/google-rtc-aec.conf:47]: It is described in my rimage documents. CPC : cycle per chunk data (1ms in most case)

File: tools/topology/topology2/include/components/google-rtc-aec.conf (hunk 13 above)
Comment (RanderWang) [reply tools/topology/topology2/include/components/google-rtc-aec.conf:52]: It is a concept of ipc4 module.  
````c
struct ipc4_base_module_cfg {
	uint32_t cpc;		/**< the max count of Cycles Per Chunk processing */
	uint32_t ibs;		/**< input Buffer Size (in bytes)  */
	uint32_t obs;		/**< output Buffer Size (in bytes) */
	uint32_t is_pages;	/**< number of physical pages used */
	struct ipc4_audio_format audio_fmt;
````

File: tools/topology/topology2/platform/intel/google-rtc-aec-generic.conf (hunk 15 above)
Comment (RanderWang) [reply tools/topology/topology2/platform/intel/google-rtc-aec-generic.conf:1]: it means it will fit for all google rtc aec usage

File: tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf (hunk 14 above)
Comment (RanderWang) [reply tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf:76]: sure.

File: tools/topology/topology2/include/components/google-rtc-aec.conf (hunk 10 above)
Comment (cujomalainey) [reply tools/topology/topology2/include/components/google-rtc-aec.conf:11]: Yes I get that by why is it recommended at the pipeline level, isn't every aec a effect? That doesn't explain why

File: tools/topology/topology2/include/components/google-rtc-aec.conf (hunk 12 above)
Comment (cujomalainey) [reply tools/topology/topology2/include/components/google-rtc-aec.conf:47]: Are those docs published anywhere?

File: tools/topology/topology2/include/components/google-rtc-aec.conf (hunk 13 above)
Comment (cujomalainey) [reply tools/topology/topology2/include/components/google-rtc-aec.conf:52]: Is this pages as in memory pages?

File: tools/topology/topology2/platform/intel/google-rtc-aec-generic.conf (hunk 15 above)
Comment (cujomalainey) [reply tools/topology/topology2/platform/intel/google-rtc-aec-generic.conf:1]: That still isn't clear how it differs from the capture

File: tools/topology/topology2/include/components/google-rtc-aec.conf (hunk 13 above)
Comment (RanderWang) [reply tools/topology/topology2/include/components/google-rtc-aec.conf:52]: yes, it is the memory pages used in audio dsp

File: tools/topology/topology2/include/components/google-rtc-aec.conf (hunk 12 above)
Comment (RanderWang) [reply tools/topology/topology2/include/components/google-rtc-aec.conf:47]: yes,  In rimage doc https://github.com/thesofproject/sof-docs/pull/451

File: tools/topology/topology2/platform/intel/google-rtc-aec-generic.conf (hunk 15 above)
Comment (RanderWang) [reply tools/topology/topology2/platform/intel/google-rtc-aec-generic.conf:1]: A pcm stream is composed of at least two pipelines,  one for input processing and the other is for output processing. google-rtc-aec-capture.conf is for output processing.  generic.conf includes pcm definition, how to connect to input processing pipeline and route info.

**Comment by singalsu** [comment]
> > @RanderWang Can you share a tplgtool.py created picture of this topology?
> 
> @singalsu Do we need gain for each pipeline ? if not we can remove them and one copier module so that we have better performance 

Thanks for the picture. I would place AEC up one layer and have and leave gains there after AEC as placeholder for other processing. Gain may be needed for mic mute feature anyway with some device button & led. But it depends on what the google rtc aec topology requirement is. Can you please comment @cujomalainey @johnylin76 .

File: tools/topology/topology2/platform/intel/google-rtc-aec-generic.conf (hunk 15 above)
Comment (cujomalainey) [reply tools/topology/topology2/platform/intel/google-rtc-aec-generic.conf:1]: > google-rtc-aec-capture.conf is for output processing

That is very confusing

Why no call them capture (actually capture) and reference?

File: tools/topology/topology2/include/components/google-rtc-aec.conf (hunk 13 above)
Comment (cujomalainey) [reply tools/topology/topology2/include/components/google-rtc-aec.conf:52]: what if its variable based on format?

Comment (RanderWang) [reply tools/topology/topology2/include/components/google-rtc-aec.conf:52]: is_pages is used for bss.  Different format will has different ibs & obs which will be added to calculated the full memory size.
````c
static void
sof_ipc4_update_pipeline_mem_usage(struct snd_sof_dev *sdev, struct snd_sof_widget *swidget,
				   struct sof_ipc4_base_module_cfg *base_config)
{
	struct sof_ipc4_fw_module *fw_module = swidget->module_info;
	struct snd_sof_widget *pipe_widget;
	struct sof_ipc4_pipeline *pipeline;
	int task_mem, queue_mem;
	int ibs, bss, total;

	ibs = base_config->ibs;
	bss = base_config->is_pages;

	task_mem = SOF_IPC4_PIPELINE_OBJECT_SIZE;
	task_mem += SOF_IPC4_MODULE_INSTANCE_LIST_ITEM_SIZE + bss;

	if (fw_module->man4_module_entry.type & SOF_IPC4_MODULE_LL) {
		task_mem += SOF_IPC4_FW_ROUNDUP(SOF_IPC4_LL_TASK_OBJECT_SIZE);
		task_mem += SOF_IPC4_FW_MAX_QUEUE_COUNT * SOF_IPC4_MODULE_INSTANCE_LIST_ITEM_SIZE;
		task_mem += SOF_IPC4_LL_TASK_LIST_ITEM_SIZE;
	} else {
		task_mem += SOF_IPC4_FW_ROUNDUP(SOF_IPC4_DP_TASK_OBJECT_SIZE);
		task_mem += SOF_IPC4_DP_TASK_LIST_SIZE;
	}

	ibs = SOF_IPC4_FW_ROUNDUP(ibs);
	queue_mem = SOF_IPC4_FW_MAX_QUEUE_COUNT * (SOF_IPC4_DATA_QUEUE_OBJECT_SIZE +  ibs);

	total = SOF_IPC4_FW_PAGE(task_mem + queue_mem);

	pipe_widget = swidget->spipe->pipe_widget;
	pipeline = pipe_widget->private;
	pipeline->mem_usage += total;
}
````

File: tools/topology/topology2/platform/intel/google-rtc-aec-generic.conf (hunk 15 above)
Comment (RanderWang) [reply tools/topology/topology2/platform/intel/google-rtc-aec-generic.conf:1]: good idea. I will change it.

File: tools/topology/topology2/include/components/google-rtc-aec.conf (hunk 8 above)
Comment (RanderWang) [reply tools/topology/topology2/include/components/google-rtc-aec.conf:113]: sure, you can define different channel counts on different input pins, please check audio_format in audio_format.conf
````c
        # output channel count
        DefineAttribute."out_channels" {
                # Token set reference name
                token_ref       "sof_tkn_cavs_audio_format.word"
                constraints {
                        min 1
                        max 8
                }
        }

        # output channel map
        DefineAttribute."out_ch_map" {
                # Token set reference name
                token_ref       "sof_tkn_cavs_audio_format.word"
        }

        # output channel config
        DefineAttribute."out_ch_cfg" {
                # Token set reference name
                token_ref       "sof_tkn_cavs_audio_format.word"
        }

````

File: tools/topology/topology2/include/components/google-rtc-aec.conf (hunk 10 above)
Comment (RanderWang) [reply tools/topology/topology2/include/components/google-rtc-aec.conf:11]: why do we need to set this in pipeline level ? It should be in widget level. Here is at widget level

File: tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf (hunk 9 above)
Comment (plbossart) [reply tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf:1]: @RanderWang you added TWO instances of AEC, which is considerably more complicated that a copier?

This makes no sense to me sorry. @cujomalainey please chime in and clarify the ask.

**Comment by cujomalainey** [comment]
I don't think we need gain on capture, I believe it includes an AGC @perahgren to correct me if I am wrong

File: tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf (hunk 9 above)
Comment (RanderWang) [reply tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf:1]: @plbossart  No.  we don't use two pipelines at the same time.  If FW supports NS, we only use AEC + NS pipeline and the another AEC pipeline is not used. but for your pipeline it will be AEC + copier + NS, more than one module : copier

Comment (RanderWang) [reply tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf:1]: In statically analysis,  one more AEC module, but in real usage, only one AEC is used since another pipeline is not active.
if FW doesn't support NS, only AEC pipeline is used, or  AEC + NS is used.

**Comment by RanderWang** [comment]
@cujomalainey  could you please help to answer https://github.com/thesofproject/sof/pull/7237#discussion_r1137230464. What is your picture of pipeline ?

File: tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf (hunk 9 above)
Comment (plbossart) [reply tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf:1]: that's too complicated, it means more connections for the echo reference needing to go to two AEC module instances.

And in terms of use case transitions it's absolutely awful. Not to mention tuning where the SAME set of parameters will have to be duplicated for two instances of the same thing.

I maintain this is not the right direction. 

@thesofproject/google please speak-up.

Comment (RanderWang) [reply tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf:1]: sure, I will change the topology. @cujomalainey  Do you have any idea about Pierre's recommendation.

Comment (cujomalainey) [reply tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf:1]: @plbossart we are planning to roll AEC out on spk+dmic, it would be good to leave a HS+HS-mic option for experimentation in the config for us to test with

File: tools/topology/topology2/include/components/google-rtc-aec.conf (hunk 10 above)
Comment (cujomalainey) [reply tools/topology/topology2/include/components/google-rtc-aec.conf:11]: I am not sure I understand the levels...

File: tools/topology/topology2/include/components/google-rtc-aec.conf (hunk 13 above)
Comment (cujomalainey) [reply tools/topology/topology2/include/components/google-rtc-aec.conf:52]: What is the consequence if the number is under?

File: tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf (hunk 9 above)
Comment (plbossart) [reply tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf:1]: Sorry, I don't get your point @cujomalainey. It might be easier to provide Rander with a picture of the desired topology...

Comment (cujomalainey) [reply tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf:1]: image
image

Ideally we would have a copy of each of the above images for SPK+DMIC, and a copy of the second for HS + HS Mic on top of the standard volume only passthrough options

File: tools/topology/topology2/include/components/google-rtc-aec.conf (hunk 13 above)
Comment (RanderWang) [reply tools/topology/topology2/include/components/google-rtc-aec.conf:52]: the memory allocation may be failed, but since the memory size is aligned to page, a little more memory is prepared for allocation.

File: tools/topology/topology2/include/components/google-rtc-aec.conf (hunk 10 above)
Comment (RanderWang) [reply tools/topology/topology2/include/components/google-rtc-aec.conf:11]: I am also not sure I understand the levels...  Here type is a example in comments for example

File: tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf (hunk 9 above)
Comment (cujomalainey) [reply tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf:1]: @plbossart ah thanks I missed that, hierarchy is better, less instances.

**Comment by RanderWang** [comment]
@plbossart  @cujomalainey  this diagram is for the pipeline. There is no NS support now, so I temporarily gain for NS. And Copier.SSP.8.1 provides  loop back data of speaker playback stream.

````
DMIC ------->     copier ---> host
                    |
                    +------> AEC -----> copier -----> host
                                           |
                                           +----------> NS -------> copier host

````
aec1

**Comment by lgirdwood** [comment]
> > I'll reinforce this again as in my other PR. I will not approve commits that relate to google code if I don't have the documentation to approve what I am reviewing. **please prioritize these documents**
> 
> According to the question, a documents of topology2 and ipc4 is needed. @lgirdwood this is a big task

Ack - @mmaka1 and @mwasko have some sof-docs text and diagrams now in progress to help explain the new features being upstreamed into sof-docs.
@jsarha can you have a look at topology2 and make sure we have the correct docs for the topology features being added here.
@cujomalainey are you able to add a FEATURE issue on sof-docs listing the areas that need more docs with some priority. This will help us prioritize new documentation.

**Review by kv2019i — APPROVED** [review]
Thanks @RanderWang , all my concerns now addressed.

**Comment by sys-pt1s** [comment]

**Comment by kv2019i** [comment]
@RanderWang Status of this?

**Comment by RanderWang** [comment]
> @RanderWang Status of this?
@kv2019i  rebased to latest code.  @cujomalainey  do you have any comments ? If not, let's merge it.

**Comment by marc-hb** [comment]
https://sof-ci.01.org/sofpr/PR7237/build7147/devicetest/index.html failed with some `aplay` errors.

PS: maybe a direct link to the new sof-docs would be useful?

**Comment by RanderWang** [comment]
> https://sof-ci.01.org/sofpr/PR7237/build7147/devicetest/index.html failed with some `aplay` errors.
> 
> PS: maybe a direct link to the new sof-docs would be useful?

This PR only affects sof-mtl-max98357a-rt5682.conf, not the nocodec topology.  sof-docs for google-rtc-aec ? Actually this PR follows the topology1 for google-rtc-aec and it is better for google to provide how their aec work.

**Comment by marc-hb** [comment]
SOFCI TEST

File: tools/topology/topology2/cavs-rt5682.conf [hunk 16]
```diff
@@ -91,6 +97,8 @@ Define {
 	BT_PB_HOST_PIPELINE_SINK "copier.SSP.10.1"
 	BT_PB_DAI_PIPELINE_SRC "copier.host.9.1"
 	BT_PB_PIPELINE_STREAM_NAME "copier.SSP.10.1"
+	AEC				0
```
cujomalainey, line 100: What is this? A lot of the other defines are a lot more verbose.

tools/topology/topology2/cavs-rt5682.conf
```diff
@@ -91,6 +97,8 @@ Define {
...
+	REF_SOURCE			'module-copier.8.2'
```
cujomalainey, line 101: AEC ref source?

tools/topology/topology2/include/components/google-rtc-aec.conf
```diff
@@ -0,0 +1,138 @@
...
+		# instance attribute
+		#
+		unique	"instance"
+	}
+
+	num_audio_formats 2
+	num_input_audio_formats 2
+	num_output_audio_formats 1
+	Object.Base.input_audio_format {
+		1 {
```
cujomalainey, line 103: any chance we give names or at least a comment so we know what endpoint is which?

cujomalainey (commented):
getting close, thanks

tools/topology/topology2/cavs-rt5682.conf
RanderWang: sure, updated to GOOGL_RTC_AEC_SUPPORT

tools/topology/topology2/cavs-rt5682.conf
RanderWang: updated to GOOGLE_RTC_AEC_REF_SOURCE

tools/topology/topology2/include/components/google-rtc-aec.conf
RanderWang: sure, added

RanderWang:
@cujomalainey  updated, thanks

cujomalainey (approved):

Issue #195, opened 2018-08-06
GCC compilation broken again


keqiaozhang:
This is a regression and caused by thesofproject/sof/pull/189 - ae793d26e
Error info：
dw-dma.c: In function 'dw_dma_stop':

dw-dma.c:489:18: error: unused variable 'lli' [-Werror=unused-variable]
  struct dw_lli2 *lli;
                  ^
dw-dma.c:488:6: error: unused variable 'i' [-Werror=unused-variable]
  int i = 0;
      ^
cc1: all warnings being treated as errors

[config.log](https://github.com/thesofproject/sof/files/2261213/config.log)

keqiaozhang:
This issue happens on BYT BDW and HSW.

keqiaozhang:
Confirmed that this issue is fixed by thesofproject/sof/pull/196.

PR #7239, opened 2023-03-10
topology2: cavs-nocodec: include SSP1 pipelines conditionally

aiChaoSONG:
There is pinmux conflict between SSP1 and DMIC on MTL, as there are other SSPs can be used, we don't include SSP1 pipelines only for mtl to unlock DMIC test.

### DIFF
--- a/tools/topology/topology2/cavs-nocodec.conf
+++ b/tools/topology/topology2/cavs-nocodec.conf
@@ -76,6 +76,7 @@ Define {
 	SSP0_MIXER_SOURCE_2		'mixin.16.1'
 	SSP0_MIXER_SINK_2		'mixout.16.1'
 	SSP0_MIXER_SOURCE_3		'mixin.21.1'
+	SSP1_ENABLED			"true"
 }
 
 # override defaults with platform-specific config
@@ -109,24 +110,6 @@ Object.Dai.SSP [
 
 		Object.Base.hw_config.1 {
 			name	"SSP0"
-			id	0
-			mclk_freq	$MCLK
-			bclk_freq	3072000
-			tdm_slot_width	32
-		}
-	}
-	{
-		id 		1
-		dai_index	1
-		direction	"duplex"
-		name		NoCodec-1
-		default_hw_conf_id	0
-		sample_bits		32
-		quirks			"lbm_mode"
-		io_clk		$MCLK
-
-		Object.Base.hw_config.1 {
-			name	"SSP1"
 			id	0
 			mclk_freq	$MCLK
 			bclk_freq	3072000
@@ -182,18 +165,6 @@ Object.Pipeline.host-copier-gain-mixin-playback [
 		}
 	}
 	{
-		index	3
-
-		Object.Widget.copier.1 {
-			stream_name 'SSP1 Playback'
-		}
-		Object.Widget.gain.1 {
-			Object.Control.mixer.1 {
-				name 'Playback Volume 3'
-			}
-		}
-	}
-	{
 		index	5
 
 		Object.Widget.copier.1 {
@@ -243,23 +214,6 @@ Object.Pipeline.mixout-gain-smart-amp-dai-copier-playback [
 ]
 
 Object.Pipeline.mixout-gain-dai-copier-playback [
-	{
-		index	4
-
-		Object.Widget.copier.1 {
-			dai_index 1
-			dai_type "SSP"
-			copier_type "SSP"
-			stream_name "NoCodec-1"
-			node_type $I2S_LINK_OUTPUT_CLASS
-		}
-
-		Object.Widget.gain.1 {
-			Object.Control.mixer.1 {
-				name 'Main Playback Volume 4'
-			}
-		}
-	}
 	{
 		index	6
 
@@ -417,13 +371,6 @@ Object.Pipeline.dai-copier-gain-module-copier-capture [
 ]
 
 Object.Pipeline.passthrough-capture [
-	{
-		index 	9
-
-		Object.Widget.copier.1 {
-			stream_name 'SSP1 Capture'
-		}
-	}
 	{
 		index 	11
 
@@ -635,26 +582,6 @@ Object.PCM.pcm [
 		}
 	}
 	{
-		name	"Port1"
-		id 1
-		direction	"duplex"
-		Object.Base.fe_dai.1 {
-			name	"Port1"
-		}
-
-		Object.PCM.pcm_caps.1 {
-			direction	"playback"
-			name "SSP1 Playback"
-			formats 'S16_LE,S24_LE,S32_LE'
-		}
-
-		Object.PCM.pcm_caps.2 {
-			direction	"capture"
-			name "SSP1 Capture"
-			formats 'S16_LE,S24_LE,S32_LE'
-		}
-	}
-	{
 		name	"Port2"
 		id 2
 		direction	"duplex"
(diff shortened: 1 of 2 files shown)


tools/topology/topology2/cavs-nocodec.conf
```diff
@@ -65,6 +65,7 @@ Define {
 	DEEP_BUFFER_PIPELINE_SINK	'mixout.2.1'
 	DEEP_BUFFER_PCM_NAME		'Deepbuffer Port0'
 	SMART_AMP_UUID			"1e:96:7a:16:e4:8a:ea:11:89:f1:00:0c:29:ce:16:35"
+	SSP1_ENABLED			"true"
```
plbossart, line 68: that's a nice change.

I would flip the logic and make all SSPs disabled by default, so that we can select which ones we want for each platform. This would be very handy when we want to test e.g. on Up Extreme where this only one SSP available on the connector.

tools/topology/topology2/cavs-nocodec.conf
```diff
@@ -753,3 +668,113 @@ Object.Base.route [
...
+				source	"mixin.3.1"
+				sink	"mixout.4.1"
+			}
+			{
+				source	"gain.4.1"
+				sink	"copier.SSP.4.1"
+			}
+			{
+				source	"copier.SSP.10.1"
+				sink	"copier.host.9.1"
```
plbossart, line 776: this is far from self-explanatory, @ranj063's cleanups should really be applied, There's nothing in there that looks like SSP1....

aiChaoSONG: agree, The pipelines are group by pipeline instead by DAI, a clean up would be better.

aiChaoSONG:
This patch will require alsatplg change in https://github.com/alsa-project/alsa-utils/pull/195

aiChaoSONG:
Updates:
- rebased on latest main
- remove dependency on https://github.com/alsa-project/alsa-utils/pull/195, because we have no idea when it will get merged

lyakh:
@marcinszkudlinski could you please fix these two: https://github.com/thesofproject/sof/actions/runs/4423320751/jobs/7755939911#step:7:1572 and https://github.com/thesofproject/sof/actions/runs/4423320751/jobs/7755939911#step:7:1590? thanks

tools/topology/topology2/cavs-nocodec.conf
```diff
@@ -774,3 +691,114 @@ IncludeByKey.SSP0_MIXER_2LEVEL {
...
+
+
+# There is pinmux conflict between SSP1 and DMIC on MTL,
+# so include SSP1 pipelines conditionally.
+IncludeByKey.SSP1_ENABLED {
+	"true" {
+		Object.Dai.SSP [
+			{
+				id 		1
+				dai_index	1
```
lgirdwood, line 703: Can we not do this via a macro defining the SSP to use in the platform file i.e.
```
# mtl.conf only define the SSPs that are available
undefine NOCODEC_TEST_SSP1
define NOCODEC_TEST_SSP2
define NOCODEC_TEST_SSP3
define NOCODEC_TEST_SSP4
# and so on
```

aiChaoSONG: I don't think topology2 support 'undefine' keyword. The only thing we can do is to first set the default values for each DAI and override default values in PLATFORM.conf

aiChaoSONG:
@ranj063 @ujfalusi @kv2019i could you please help to give this PR a review?

plbossart:
> @ranj063 @ujfalusi @kv2019i could you please help to give this PR a review?

the dependency on https://github.com/alsa-project/alsa-utils/pull/195 prevents any merge from happening. Why are you asking for approvals @aiChaoSONG if there is no agreement on the dependencies and plan to update the docker image?

kv2019i:
@plbossart wrote:
> the dependency on [alsa-project/alsa-utils#195](https://github.com/alsa-project/alsa-utils/pull/195) prevents any merge from happening. Why are you asking for approvals @aiChaoSONG if there is no agreement on the dependencies and plan to update the docker image?

I think the alsa-utils #195 dependency is now removed. In this https://github.com/thesofproject/sof/compare/b0b1be1251048b57ce1488085181048b1cdd3aca..e2cf2b866c925859e5db9f827a947c2f425d65c6

kv2019i (approved):

tools/topology/topology2/cavs-nocodec.conf
```diff
@@ -76,10 +76,13 @@ Define {
 	SSP0_MIXER_SOURCE_2		'mixin.16.1'
 	SSP0_MIXER_SINK_2		'mixout.16.1'
 	SSP0_MIXER_SOURCE_3		'mixin.21.1'
+	SSP1_ENABLED			"false"
 }
 
 # override defaults with platform-specific config
 IncludeByKey.PLATFORM {
+	"tgl"	"platform/intel/tgl.conf"
+	"adl"	"platform/intel/tgl.conf"
```
plbossart, line 85: I am having a bit of a heartburn with this PR.

Is the pinmux is a problem on one instance of the MTL RVP, or the MTL SOC in general? It's got to be the former, there is not general incompatibility between SSP1 and DMIC, it's just that the configuration of ONE board that has a restriction.

So rather than having a blanket removal for SSP1, I only only remove SSP1 on specific targets, and use udev rules or a tplg_name kernel option to work-around this limitation.

tools/topology/topology2/platform/intel/tgl.conf
```diff
@@ -0,0 +1,4 @@
+# TGL-specific variable definitions
+Define {
+	SSP1_ENABLED			"true"
```
plbossart, line 3: so we are going to have to add SSP1_ENABLED for LNL as well? It'd be smarter to disable SSP1 when needed, fewer configurations to maintain in the end.

tools/topology/topology2/cavs-nocodec.conf
aiChaoSONG: @plbossart Let's extinguish the fire that burns your heart, how about generating two topology targets, see updated code.
- sof-mtl-nocodec-ssp0-ssp2.tplg, use this one for MTL RVP, only SSP0 and SSP2 are enabled
- sof-mtl-nocodec.tplg, use this for other MTL device on which all SSPs are enabled.

I think this should address your concern that there are boards have/haven't the pin-mux issue. And configurations are all in the cmake, and only needed if we want to disable SSP1.

tools/topology/topology2/platform/intel/tgl.conf
mengdonglin: > so we are going to have to add SSP1_ENABLED for LNL as well? It'd be smarter to disable SSP1 when needed, fewer configurations to maintain in the end.

@plbossart  @aiChaoSONG  We're not sure if LNL will have the same pin mux conflict between SSP1 and MIC as MTL, but probably it will. So it's helpful if we can disable SSP1 when needed.

tools/topology/topology2/cavs-nocodec.conf
lgirdwood: @plbossart good for you now ?

plbossart (approved):

kv2019i:
One multiple-pause fail in https://sof-ci.01.org/sofpr/PR7239/build4940/devicetest/index.html , but seems to be a case of https://github.com/thesofproject/sof/issues/7324 so not related to this PR. Proceeding with merge.

Issue #7230, opened 2023-03-08
[BUG][MTL-004][Chrome] Xrun on capture pipeline


aiChaoSONG:
**Describe the bug**
use the firmware compiled without -d option (main or releases/mtl/v4.0), we see capture  xrun on mtl chromebook.  if firmware is compiled with -d option, see issue #7228.
**To Reproduce**
do capture on any capture pipeline (headset, DMIC or BT), we observe I/O errror and xrun in mtrace from the begining, no obvious error in dmesg.

**Reproduction Rate**
100%

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel: {8d5df0f5a from chrome kernel}
    * SOF: {MTL-004 release, tag: releases/mtl/v4.0, or sof main. compiled **without** -d option}
2) Name of the topology file
    * Topology: {sof-mtl-max98360a-rt5682.tplg}
3) Name of the platform(s) on which the bug is observed.
    * Platform: {{MTL Chromebook}

```
[00:00:00.132,541] <inf> ll_schedule: zephyr_ll_task_schedule_common: task add 0xa00a8fc0 0x20180U priority 0 flags 0x0
[00:00:00.133,143] <inf> host_comp: host_get_copy_bytes_normal: comp:1 0x40001 no bytes to copy, available bytes: 0, free_bytes: 1536
[00:00:00.133,151] <inf> dai_intel_ssp: dai_ssp_get_properties: dai_ssp_get_properties dai_index 2
[00:00:00.133,156] <inf> dai_intel_ssp: dai_ssp_get_properties: dai_ssp_get_properties fifo 172048
[00:00:00.133,161] <inf> dai_intel_ssp: dai_ssp_get_properties: dai_ssp_get_properties handshake 6
[00:00:00.133,166] <inf> dai_intel_ssp: dai_ssp_get_properties: dai_ssp_get_properties init delay 0
[00:00:00.133,175] <inf> dai_intel_ssp: dai_ssp_pre_start: dai_ssp_pre_start
[00:00:00.133,206] <inf> dai_intel_ssp: dai_ssp_start: dai_ssp_start
[00:00:00.133,211] <inf> dai_intel_ssp: dai_ssp_update_bits: dai_ssp_update_bits base 2a000, reg 34, mask 100, value 100
[00:00:00.133,263] <wrn> dai_comp: comp:0 0x40000 dai_copy(): nothing to copy
[...]

Issue #7228, opened 2023-03-08
[BUG][MTL-Chromebook][MTL-004] IPC timeout when set pipeline state to running


aiChaoSONG:
**Describe the bug**
The DMIC capture pipeline on MTL chromebook cannot work, we see ipc timeout for setting pipeline state to running IPC.

**To Reproduce**
do recording through DMIC pipeline.

**Reproduction Rate**
100%

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel: {8d5df0f5a from chrome kernel}
    * SOF: {MTL-004 release, tag: releases/mtl/v4.0}
2) Name of the topology file
    * Topology: {sof-mtl-max98360a-rt5682.tplg}
3) Name of the platform(s) on which the bug is observed.
    * Platform: {MTL Chromebook}

```
[   89.493528] sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx      : 0x13000004|0x0: GLB_SET_PIPELINE_STATE
[   89.998215] sof-audio-pci-intel-mtl 0000:00:1f.3: ipc timed out for 0x13000004|0x0
[   90.006716] sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ IPC dump start ]------------
[   90.016161] sof-audio-pci-intel-mtl 0000:00:1f.3: Host IPC initiator: 0x93000004|0x0|0x0, target: 0x33000000|0x0|0x0, ctl: 0x3
[   90.028894] sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ IPC dump end ]------------
[   90.038132] sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ DSP dump start ]------------
[   90.047562] sof-audio-pci-intel-mtl 0000:00:1f.3: IPC timeout
[   90.053983] sof-audio-pci-intel-mtl 0000:00:1f.3: fw_state: SOF_FW_BOOT_COMPLETE (6)
[   90.062642] sof-audio-pci-intel-mtl 0000:00:1f.3: ROM status: 0x5, ROM error: 0x0
[   90.071002] sof-audio-pci-intel-mtl 0000:00:1f.3: ROM debug status: 0x50000005, ROM debug error: 0x0
[   90.081211] sof-audio-pci-intel-mtl 0000:00:1f.3: ROM feature bit not enabled
[...]

PR #7242, opened 2023-03-10
topology2: add cpc value for gain

aiChaoSONG:
The missing cpc in gain component results in wrong calculation for clock selection, and causes capture issues.

Link: https://github.com/thesofproject/sof/issues/7228
Link: https://github.com/thesofproject/sof/issues/7230

### DIFF
--- a/tools/topology/topology2/include/components/gain.conf
+++ b/tools/topology/topology2/include/components/gain.conf
@@ -164,6 +164,7 @@ Class.Widget."gain" {
 	type 			"pga"
 	uuid 			"A8:A9:BC:61:D0:18:18:4A:8E:7B:26:39:21:98:04:B7"
 	no_pm			"true"
+	cpc 			10183
 	period_sink_count	2
 	period_source_count	2
 	curve_type		"fade"


RanderWang (approved):

abonislawski (approved):
Thanks @aiChaoSONG  can you also update other modules to match the manifest data?

aiChaoSONG:
@RanderWang @abonislawski CPC for all modules in manifest are zero, but we also have CPS there, is it ok for you to calculate  CPC from the maximum CPS for the module, to make CPC work for all supported formats.

@abonislawski I see several formula, could you please clarify which one is correct?
```
cps = cpc * sampling frequency
cps = (cpc / (frame_size * frame_count)) * 1000

From the cps and cpc values for SRC moudle in mtl.toml, the formula look like
cps = cpc * 1000
```

abonislawski:
@aiChaoSONG please use mtl-004 manifest as a reference:
https://github.com/thesofproject/rimage/commits/mtl-004-drop-stable
it contains updated manifest with measured values

@kfrydryx could you please clarify cps/cpc calculations?

kfrydryx:
@abonislawski @aiChaoSONG CPS means cycles per second and CPC - cycles per chunk. And since we have 1ms scheduling period, `CPS = CPC*(1s/scheduling_period) = CPC*1000`. You can calculate the value in multiple ways, but the result will be the same.

kv2019i (approved):

kv2019i:
FYI @mwasko @lgirdwood  -- the relation to topology (and how to calculate) needs to be covered in sof-docs (related to https://github.com/thesofproject/sof-docs/pull/451  )

PR #6513, opened 2022-11-01
Tools: Testbench: xt-testbench

singalsu:
Now ready. see commit texts for description.

PR #7244, opened 2023-03-10
Tools: Testbench: Avoid need linux kernel headers in build

singalsu:
This patch adds a simplified linux/types.h into testbench headers to enable include of ALSA asoc.h to build without other difficult kernel headers content.

The include path for asoc.h is changed in fuzzer, tplg_parser, and testbench to be without prefix alsa/sound/uapi to allow the Makefile headers path to specify only the uapi directory.

This change avoids build fail with xcc toolchain that can't use the gcc toolchain headers from the system where the ALSA headers are located.

### DIFF
--- a/tools/testbench/CMakeLists.txt
+++ b/tools/testbench/CMakeLists.txt
@@ -7,6 +7,8 @@ project(SOF_TESTBENCH C)
 include(../../scripts/cmake/misc.cmake)
 include(CheckCCompilerFlag)
 
+set(default_asoc_h "/usr/include/alsa/sound/uapi/asoc.h")
+
 add_executable(testbench
 	testbench.c
 	common_test.c
@@ -30,6 +32,12 @@ target_include_directories(testbench PRIVATE
 	"${PROJECT_SOURCE_DIR}/xtos/include"
 )
 endif()
+
+# Configuration time, make copy
+configure_file(${default_asoc_h} ${CMAKE_CURRENT_BINARY_DIR}/include/alsa/sound/asoc.h)
+
+# Build time
+target_include_directories(testbench PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/include")
 
 # -Wimplicit-fallthrough is preferred, check if it's supported
 check_c_compiler_flag(-Wimplicit-fallthrough supports_implicit_fallthrough)
--- a/tools/testbench/include/linux/types.h
+++ b/tools/testbench/include/linux/types.h
@@ -1,0 +1,22 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+
+#ifndef __TESTBENCH_LINUX_TYPES_H__
+#define __TESTBENCH_LINUX_TYPES_H__
+
+#include <stdint.h>
+
+/*
+ * This header files allows to include asoc.h for topology parsing to
+ * non-gcc builds with other C library, e.g. the one that is used by
+ * xt-xcc compiler. The kernel linux/types.h cannot be used because
+ * the other definitions there are not compatible with the toolchain.
+ */
+
+/* There are minimum types needed for alsa/sound/uapi/asoc.h */
+
+typedef int64_t __le64;
+typedef int32_t __le32;
+typedef int16_t __le16;
+typedef uint8_t __u8;
+
+#endif /* __TESTBENCH_LINUX_TYPES_H__ */
--- a/tools/tplg_parser/CMakeLists.txt
+++ b/tools/tplg_parser/CMakeLists.txt
@@ -7,6 +7,8 @@ project(SOF_TPLG_PARSER C)
 include(../../scripts/cmake/misc.cmake)
 include(CheckCCompilerFlag)
 
+set(default_asoc_h "/usr/include/alsa/sound/uapi/asoc.h")
+
 set(sof_source_directory "${PROJECT_SOURCE_DIR}/../..")
 
 if (CONFIG_LIBRARY_STATIC)
@@ -39,6 +41,12 @@ target_include_directories(sof_tplg_parser PRIVATE ${sof_source_directory}/src/p
 
 # TODO: The topology parser should NOT need to include RTOS header. FIX.
 target_include_directories(sof_tplg_parser PRIVATE ${sof_source_directory}/xtos/include)
+
+# Configuration time, make copy
+configure_file(${default_asoc_h} ${CMAKE_CURRENT_BINARY_DIR}/include/alsa/sound/asoc.h)
+
+# Build time
+target_include_directories(sof_tplg_parser PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/include")
 
 # -Wimplicit-fallthrough is preferred, check if it's supported
 check_c_compiler_flag(-Wimplicit-fallthrough supports_implicit_fallthrough)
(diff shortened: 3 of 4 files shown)


marc-hb (approved):
Note this was successfully tested as part of larger
- #6513

tools/fuzzer/CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ target_include_directories(sof-fuzzer PRIVATE
 	"${SOF_ROOT_SOURCE_DIRECTORY}/src/arch/host/include"
 	"${SOF_ROOT_SOURCE_DIRECTORY}/src/platform/library/include"
 	"${SOF_ROOT_SOURCE_DIRECTORY}"
+        "/usr/include/alsa/sound/uapi"
```
lgirdwood, line 43: This needs to be a CMake macro (with this as the default) as the ALSA lib prefix can be different.

marc-hb (changes requested):
> This change avoids build fail with xcc toolchain that can't use the gcc toolchain headers from the system where the ALSA headers are located.

On second thought this is way too much of a hack.

Instead ask CMake to copy `/usr/include/alsa/sound/uapi/asoc.h` into something like `build_dir/alsa/sound/uapi/asoc.h`

Something like this UNTESTED:

```
configure_file ( /usr/include/alsa/sound/uapi/asoc.h ${CMAKE_CURRENT_BINARY_DIR}/alsa/sound/uapi/asoc.h


target_include_directories(sometarget ${CMAKE_CURRENT_BINARY_DIR}alsa/sound/uapi/asoc.h)
```

https://www.google.com/search?q=cmake+generated+header


PS: I ran `pkg-config --cflags-only-I alsa` and made sure it's empty on Ubuntu, Fedora and Arch Linux. In other words the location is standard.

lgirdwood:
> PS: I ran `pkg-config --cflags-only-I alsa` and made sure it's empty on Ubuntu, Fedora and Arch Linux. In other words the location is standard.

Thanks for checking @marc-hb - @singalsu I would make sure the comment about running ```pkg-config --cflags-only-I alsa``` is next to where you define the location. That way easy to check and update.

marc-hb:
No this is very far from what I suggested, sorry for the confusion.

\- I mentioned `pkg-config --cflags-only-I alsa` only to demonstrate it was always empty hence NOT needed. To test your latest CMake code invoking `pkg-config` you would need to find a really weird Linux distribution where it is not empty (hence required), I really doubt you found any such distro? So your latest CMake code is not tested and most likely does not work :-) Also, if we ever wanted to use `pkg-config` (I think we do **not**) then it should not be invoked from CMake directly, there's a CMake API for that: https://cmake.org/cmake/help/latest/module/FindPkgConfig.html

\- There are 3 different `asoc.h` files in `/usr/include/` + maybe others installed manually so I think we should NOT `#include <asoc.h>` because no one wants to debug the order of `-I` include flags in very long and usually hidden command lines generated by 2 or 3 layers of build system indirections - multiplied by the number of supported toolchains. Top up that mess with whatever "creative" modifications developers have done to their system; like installing custom ALSA versions in random locations. So I think we should keep `#include <alsa/sound/asoc.h>`. It is what a non-empty `pkg-config` would require anyway. To make `#include <alsa/sound/asoc.h>` work with XCC, simply copy the hardcoded `/usr/include/alsa/sound/asoc.h` to some `${CMAKE_BINARY_DIR}/alsa/sound/` directory and add the latter to `target_include_directories()`. See above.

marc-hb:
> simply copy the hardcoded /usr/include/alsa/sound/asoc.h to some ${CMAKE_BINARY_DIR}/alsa/sound/ directory and add the latter to target_include_directories()

This is actually the example given in https://cmake.org/cmake/help/latest/command/configure_file.html !

Two differences:
- zero substitution needed, pure copy
- You may need to create the `${CMAKE_BINARY_DIR}/alsa/sound/` directory yourself. This should do it: https://cmake.org/cmake/help/latest/command/file.html#make-directory

singalsu:
I tried to manually set the path that pkg-config to /foo and checked that the build tried to find headers from /foo/alsa/sound/uapi

> No this is very far from what I suggested, sorry for the confusion.
> 
> - I mentioned `pkg-config --cflags-only-I alsa` only to demonstrate it was always empty hence NOT needed. To test your latest CMake code invoking `pkg-config` you would need to find a really weird Linux distribution where it is not empty (hence required), I really doubt you found any such distro? 

I could test it only by setting other_alsa_include_path to /foo and check that code build failed with -I/foo/alsa/sound/uapi seen in the command line. But not a real test, agree.

> So your latest CMake code is not tested and most likely does not work :-) Also, if we ever wanted to use `pkg-config` (I think we do **not**) then it should not be invoked from CMake directly, there's a CMake API for that: https://cmake.org/cmake/help/latest/module/FindPkgConfig.html

OK, I'll learn about that

> 
> - There are 3 different `asoc.h` files in `/usr/include/` + maybe others installed manually so I think we should NOT `#include <asoc.h>` because no one wants to debug the order of `-I` include flags in very long and usually hidden command lines generated by 2 or 3 layers of build system indirections - multiplied by the number of supported toolchains. Top up that mess with whatever "creative" modifications developers have done to their system; like installing custom ALSA versions in random locations. So I think we should keep `#include <alsa/sound/asoc.h>`. It is what a non-empty `pkg-config` would require anyway. To make `#include <alsa/sound/asoc.h>` work with XCC, simply copy the hardcoded `/usr/include/alsa/sound/asoc.h` to some `${CMAKE_BINARY_DIR}/alsa/sound/` directory and add the latter to `target_include_directories()`. See above.

Yep, will try :)

marc-hb:
> copy the hardcoded /usr/include/alsa/sound/asoc.h to some ${CMAKE_BINARY_DIR}/alsa/sound/ directory and add ${CMAKE_BINARY_DIR} to target_include_directories(  ). See above and below.

I just realized you may need the `SYSTEM` keyword in `target_include_directories( ... SYSTEM ... )` because we have `#include <alsa/sound/asoc.h>` with `< >`.

Unless we switch to `#include "alsa/sound/asoc.h"`? Hopefully not, this would affect the regular host toolchain.

I'm not sure. However I'm sure that we would have EVEN MORE problems like this if we switch to `#include "asoc.h"` :-)

singalsu:
> I just realized you may need the `SYSTEM` keyword in `target_include_directories( ... SYSTEM ... )` because we have `#include <alsa/sound/asoc.h>` with `< >`.

I tried to replace PRIVATE with SYSTEM. The cmake errors with that. Changing to SYSTEM PRIVATE works. But I have no idea what I'm doing. Is this better?

marc-hb (approved):

lgirdwood (approved):

tools/testbench/CMakeLists.txt
```diff
@@ -31,6 +33,12 @@ target_include_directories(testbench PRIVATE
 )
 endif()
 
+# Configuration time, make copy
+configure_file(${default_asoc_h} ${CMAKE_CURRENT_BINARY_DIR}/include/alsa/sound/asoc.h)
+
+# Build time
+target_include_directories(testbench SYSTEM PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/include")
```
marc-hb, line 40: off-line @singalsu said SYSTEM is not needed. Should be removed then.

singalsu: Yep, I just changed it back to PRIVATE.

kv2019i (approved):

marc-hb:
Only suspend/resume and PM failures in MTL https://sof-ci.01.org/sofpr/PR7244/build4801/devicetest

Everything else green.

PR #7251, opened 2023-03-10
Tools: Tplg_parser: Fix void pointer arithmetic

singalsu:
Pointer arithmetic with void pointer is not allowed. Functions get_token_uint32_t() and get_token_comp_format() are updated to handle correctly the bytes offset.

### DIFF
--- a/tools/tplg_parser/tokens.c
+++ b/tools/tplg_parser/tokens.c
@@ -49,7 +49,8 @@ int get_token_uint32_t(void *elem, void *object, uint32_t offset,
 		       uint32_t size)
 {
 	struct snd_soc_tplg_vendor_value_elem *velem = elem;
-	uint32_t *val = object + offset;
+	char *vobject = object;
+	uint32_t *val = (uint32_t *)(vobject + offset);
 
 	*val = velem->value;
 	return 0;
@@ -70,7 +71,8 @@ int get_token_comp_format(void *elem, void *object, uint32_t offset,
 			  uint32_t size)
 {
 	struct snd_soc_tplg_vendor_string_elem *velem = elem;
-	uint32_t *val = object + offset;
+	char *vobject = object;
+	uint32_t *val = (uint32_t *)(vobject + offset);
 
 	*val = find_format(velem->string);
 	return 0;
@@ -79,7 +81,8 @@ int get_token_comp_format(void *elem, void *object, uint32_t offset,
 int get_token_dai_type(void *elem, void *object, uint32_t offset, uint32_t size)
 {
 	struct snd_soc_tplg_vendor_string_elem *velem = elem;
-	uint32_t *val = (uint32_t *)((uint8_t *)object + offset);
+	char *vobject = object;
+	uint32_t *val = (uint32_t *)(vobject + offset);
 
 	*val = find_dai(velem->string);
 	return 0;
@@ -89,7 +92,8 @@ int get_token_process_type(void *elem, void *object, uint32_t offset,
 			   uint32_t size)
 {
 	struct snd_soc_tplg_vendor_string_elem *velem = elem;
-	uint32_t *val = (uint32_t *)((uint8_t *)object + offset);
+	char *vobject = object;
+	uint32_t *val = (uint32_t *)(vobject + offset);
 
 	*val = tplg_get_process_name(velem->string);
 	return 0;


tools/tplg_parser/tokens.c
```diff
@@ -49,7 +49,7 @@ int get_token_uint32_t(void *elem, void *object, uint32_t offset,
 		       uint32_t size)
 {
 	struct snd_soc_tplg_vendor_value_elem *velem = elem;
-	uint32_t *val = object + offset;
+	uint32_t *val = (uint32_t *)((uint8_t *)object + offset);
```
marc-hb, line 52: In the larger #6513 , @lyakh noted that `(uint32_t *)((uint8_t *)object + offset)` can cause unaligned access.

How about `(uint32_t *)object + offset/4` instead to cause a crash if offset is not aligned?

Or maybe an `assert()`

lyakh: @marc-hb yes, but I didn't imply which way was preferred. Maybe the unaligned access was a feature and not a bug. Although I *think* Xtensa doesn't like them, but I might be wrong.

singalsu: The code in kernel is like the patch above. See sound/soc/sof/topology.c line 346. The usage in tplg_parser is similar.

Some assume aligned macro could be used?

lgirdwood: Lets keep it simple and move on. Non alignment needs to be supported due to packing.
```diff
diff --git a/tools/tplg_parser/tokens.c b/tools/tplg_parser/tokens.c
index b85256d8f..c728b50e9 100644
--- a/tools/tplg_parser/tokens.c
+++ b/tools/tplg_parser/tokens.c
@@ -49,7 +49,8 @@ int get_token_uint32_t(void *elem, void *object, uint32_t offset,
                       uint32_t size)
 {
        struct snd_soc_tplg_vendor_value_elem *velem = elem;
-       uint32_t *val = object + offset;
+       char *vobject = object;
+       uint32_t *val = (uint32_t *)(vobject + offset);
 
        *val = velem->value;
        return 0;
```

lyakh: @lgirdwood but does Xtensa even support unaligned access? I see this in the list of exception causes:
```
#define EXCCAUSE_UNALIGNED		9	/* Unaligned Load or Store */
```
and I think when we pack data for IPC or similar ABIs we make sure ourselves that all fields have suitable alignment

lgirdwood: This is topology and not parsed by running FW, simulator can trap or enable this option (but we've never hit so I suspect CC is not directly loading the cast, but doing a shift load sequence).

singalsu: I followed Liam's proposal and updated all `uint32_t *val = (uint32_t *)((uint8_t *)object + offset);` occurrences. Hope it's OK now.

lyakh: > This is topology and not parsed by running FW, simulator can trap or enable this option (but we've never hit so I suspect CC is not directly loading the cast, but doing a shift load sequence).

@lgirdwood aah, right, this is never sent to the firmware as is, sorry for confusion

kv2019i (approved):

lgirdwood (approved):

tools/tplg_parser/tokens.c
```diff
@@ -49,7 +49,8 @@ int get_token_uint32_t(void *elem, void *object, uint32_t offset,
 		       uint32_t size)
 {
 	struct snd_soc_tplg_vendor_value_elem *velem = elem;
-	uint32_t *val = object + offset;
+	char *vobject = object;
+	uint32_t *val = (uint32_t *)(vobject + offset);
```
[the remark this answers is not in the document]
marc-hb: So `sizeof(char)` is always 1. If a `char` had say 16 bits, then `sizeof(uint8_t)` would be... 0.5? 1? Undefined?

`sizeof(uint32_t)` would likely be 2 and I bet a lot of other code would break :-)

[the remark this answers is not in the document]
kv2019i: @marc-hb I think on platforms where char is 16bit, there is no uint8_t, so sizeof(uint8_t) would not be a problem. But yeah, I think POSIX guarantees char to be 8bits, so this is definitely beyond academic for testbench.

marc-hb (approved):
Forgot to approve sorry. Just in  case it matters ;-)

singalsu:
SOCFI TEST

marc-hb:
suspend/resume failures in https://sof-ci.01.org/sofpr/PR7251/build4658/devicetest/index.html

alsabat failures in https://sof-ci.01.org/sofpr/PR7251/build4659/devicetest/index.html, hopefully gone with relaxed `-k sigma` https://github.com/thesofproject/sof-test/pull/1011 ?

marc-hb:
SOFCI TEST

marc-hb:
One suspend/resume failure in https://sof-ci.01.org/sofpr/PR7251/build4749/devicetest/index.html
 
https://sof-ci.01.org/sofpr/PR7251/build4748/devicetest is all green

PR #7256, opened 2023-03-12
topology2: extract new "tplg-targets.cmake" files from CMakeList.txt

marc-hb:
The CMakeList.txt files in tools/topology/ have always confused two
different things:

- the list of .tplg files to generate, and
- the CMake logic used to invoke alsatplg and generate them.

Separate the two at last thanks a dead-simple `include()` command.  If I
had knew it was so easy I would have done this much sooner.

The `tools/build_tools/` directory is identical before versus after this
commit.

This will make it much faster to identify what actually changes in
future commits and pull requests.  This will also make CODEOWNERS
smarter and reduce Github notification noise.

Don't do it for topology1 to minimize churn and git disruption in
backports. Topology1 is not very active any more whereas topology2 has
never been released yet.

### DIFF
--- a/tools/topology/topology2/avs-tplg/CMakeLists.txt
+++ b/tools/topology/topology2/avs-tplg/CMakeLists.txt
@@ -1,50 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
-#  Array of "input-file-name;output-file-name;comma separated pre-processor variables"
-set(TPLGS
-# CAVS HDMI only topology with passthrough pipelines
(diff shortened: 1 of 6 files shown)


marc-hb:
This also makes the complete duplication across CMakeLists.txt files more obvious and easier to fix in the future.

marc-hb:
SOFCI TEST

marc-hb:
Unrelated alsabat and suspend/resume failures in https://sof-ci.01.org/sofpr/PR7256/build4531/devicetest/index.html

RanderWang (approved):

lgirdwood (approved):

PR #7259, opened 2023-03-13
ipc4: smart_amp_test: use notifier to replace bind & unbind ops

RanderWang:
To adapt smart_amp to module adapter interface we need to avoid using
    bind & unbind ops. Smart_amp uses bind & unbind ops to deal with
    feedback buffer. Now use notifier to achieve this feature.

### DIFF
--- a/src/include/sof/lib/notifier.h
+++ b/src/include/sof/lib/notifier.h
@@ -34,6 +34,8 @@ enum notify_id {
 	NOTIFIER_ID_LL_POST_RUN,		/* NULL */
 	NOTIFIER_ID_DMA_IRQ,			/* struct dma_chan_data * */
 	NOTIFIER_ID_DAI_TRIGGER,		/* struct dai_group * */
+	NOTIFIER_ID_MODULE_BIND,		/* struct ipc4_module_bind_unbind */
+	NOTIFIER_ID_MODULE_UNBIND,		/* struct ipc4_module_bind_unbind */
 	NOTIFIER_ID_COUNT
 };
 
--- a/src/ipc/ipc4/helper.c
+++ b/src/ipc/ipc4/helper.c
@@ -16,6 +16,7 @@
 #include <ipc/dai.h>
 #include <sof/ipc/msg.h>
 #include <sof/lib/mailbox.h>
+#include <sof/lib/notifier.h>
 #include <sof/list.h>
 #include <sof/platform.h>
 #include <rtos/wait.h>
@@ -372,7 +373,6 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 		goto free;
 	}
 
-
 	ret = comp_buffer_connect(sink, sink->ipc_config.core, buffer,
 				  PPL_CONN_DIR_BUFFER_TO_COMP);
 	if (ret < 0) {
@@ -387,6 +387,10 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 	ret = comp_bind(sink, bu);
 	if (ret < 0)
 		goto e_sink_bind;
+
+	notifier_event(NULL, NOTIFIER_ID_MODULE_BIND,
+		       NOTIFIER_TARGET_CORE_MASK(source->ipc_config.core) |
+		       NOTIFIER_TARGET_CORE_MASK(sink->ipc_config.core), bu, sizeof(*bu));
 
 	/* update direction for sink component if it is not set already */
 	if (!sink->direction_set && source->direction_set) {
@@ -417,6 +421,10 @@ e_sink_connect:
 	pipeline_disconnect(source, buffer, PPL_CONN_DIR_COMP_TO_BUFFER);
 free:
 	irq_local_enable(flags);
+	notifier_event(NULL, NOTIFIER_ID_MODULE_UNBIND,
+		       NOTIFIER_TARGET_CORE_MASK(source->ipc_config.core) |
+		       NOTIFIER_TARGET_CORE_MASK(sink->ipc_config.core), bu, sizeof(*bu));
+
 	buffer_free(buffer);
 	return IPC4_INVALID_RESOURCE_STATE;
 }
@@ -471,6 +479,10 @@ int ipc_comp_disconnect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 
 	if (!buffer)
 		return IPC4_INVALID_RESOURCE_ID;
+
+	notifier_event(NULL, NOTIFIER_ID_MODULE_UNBIND,
+		       NOTIFIER_TARGET_CORE_MASK(src->ipc_config.core) |
+		       NOTIFIER_TARGET_CORE_MASK(sink->ipc_config.core), bu, sizeof(*bu));
 
 	/*
 	 * Disconnect and unbind buffer from source/sink components and continue to free the buffer
--- a/src/samples/audio/smart_amp_test.c
+++ b/src/samples/audio/smart_amp_test.c
@@ -8,6 +8,7 @@
 #include <sof/samples/audio/smart_amp_test.h>
 #include <sof/audio/ipc-config.h>
 #include <sof/trace/trace.h>
+#include <sof/lib/notifier.h>
 #include <sof/ipc/msg.h>
 #include <rtos/init.h>
 #include <sof/ut.h>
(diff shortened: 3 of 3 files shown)


ranj063:
@RanderWang do you really need this? If you convert the smart amp to use the module adapter, you've got the bind;unbind ops there and you can handle it there already isnt it?

RanderWang:
> 



> would it be possible to instead add a bind / unbind method pair to `struct module_interface` to be called from `module_adapter_bind()` and `module_adapter_unbind()` respectively? That seems like a better option to me than a notifier indirection

@lyakh  The problem is we don't want to add bind & unbind for each module and this PR is going to remove them. Another problem is that module may be in another DSP core, so we need to switch to that core. This can be done by notifier.

lyakh:
> > 
> 
> > would it be possible to instead add a bind / unbind method pair to `struct module_interface` to be called from `module_adapter_bind()` and `module_adapter_unbind()` respectively? That seems like a better option to me than a notifier indirection
> 
> @lyakh The problem is we don't want to add bind & unbind for each module and this PR is going to remove them. Another problem is that module may be in another DSP core, so we need to switch to that core. This can be done by notifier.

@RanderWang why don't we want to add them? We wouldn't have to add them to all, if NULL just skip. And going to a different core shouldn't be a problem either - we do it often enough everywhere.

ranj063:
> @ranj063 It is a issue of design pattern. Now we process each module dependent ops in each module. And you will find that module dependent data structure is defined in each *.c file not in *.h file which means only this module can access it. If we process all the even in module adapter, we need switch or if to check which module to do what process. And one question is that : how to access module data structure ? Move module data structure from *.c file to *.h file ?

@RanderWang the mixin/mixout and smart_amp have exactly the same needs in terms of knowing which source/sink modules are connected at the time of copy. And for this the bind/unbin in the module_adapter and the module_source_info is enough. The smart_amp once converted to use the module_adapter interface, can use module_source_info_acquire() to query the connected sources.

lyakh:
> @RanderWang the mixin/mixout and smart_amp have exactly the same needs in terms of knowing which source/sink modules are connected at the time of copy.

We discussed this a bit with @ranj063 and we came to the conclusion, that where we have branching streams (mix, mux, smart amp, etc.) in general we have to protect the branching component against possible races between the pipeline task and the IPC. E.g. if one stream is already performing streaming, then an IPC arrives to add a branch to a mixer. So far we're protected by disabling local interrupts in `ipc_comp_connect()` when calling `comp_bind()` but that's only enough as long as we don't run cross-core streams. When we enable that, we'll need to add proper locking.
Then we have a consistent state always inside `.copy()` / `.process()` but we also have to check in which state the other pipeline currently is - has it been started or only prepared. So the `.process()` method of such modules always has to check all connected buffers and the state of any connected pipelines.
Then with the smart amp we have an additional complexity, that the streams aren't symmetric. So its `.copy()` method also has to check the type of the connected pipeline to see if it's the feedback connection. That's it, no need for `.bind()` in this case, no need for notifications.
Note, that in other cases we might indeed need a custom `.bind()` implementation, in that case we should consider adding `.bind()` / `.unbind()` callbacks to the module adapter API.

RanderWang:
> > @RanderWang the mixin/mixout and smart_amp have exactly the same needs in terms of knowing which source/sink modules are connected at the time of copy.
> 
> We discussed this a bit with @ranj063 and we came to the conclusion, that where we have branching streams (mix, mux, smart amp, etc.) in general we have to protect the branching component against possible races between the pipeline task and the IPC. E.g. if one stream is already performing streaming, then an IPC arrives to add a branch to a mixer. So far we're protected by disabling local interrupts in `ipc_comp_connect()` when calling `comp_bind()` but that's only enough as long as we don't run cross-core streams. When we enable that, we'll need to add proper locking. 

Why we need proper locking ?  If the module doesn't work on the ipc thread core,  ipc4_process_on_core will be called and this message will be processed on the pipeline core, so disabling local interrupts should work.  I am not sure current ipc4_process_on_core is used correctly, but the idea is correct.

> Then we have a consistent state always inside `.copy()` / `.process()` but we also have to check in which state the other pipeline currently is - has it been started or only prepared. So the `.process()` method of such modules always has to check all connected buffers and the state of any connected pipelines. Then with the smart amp we have an additional complexity, that the streams aren't symmetric. So its `.copy()` method also has to check the type of the connected pipeline to see if it's the feedback connection. 

Like above comments, if the buffer state change happen on the same core, you don't need to check buffer state since pipeline copy has the highest priority and disabling local interrupt should work

> That's it, no need for `.bind()` in this case, no need for notifications. Note, that in other cases we might indeed need a custom `.bind()` implementation, in that case we should consider adding `.bind()` / `.unbind()` callbacks to the module adapter API.

RanderWang:
> > @ranj063 It is a issue of design pattern. Now we process each module dependent ops in each module. And you will find that module dependent data structure is defined in each *.c file not in *.h file which means only this module can access it. If we process all the even in module adapter, we need switch or if to check which module to do what process. And one question is that : how to access module data structure ? Move module data structure from *.c file to *.h file ?
> 
> @RanderWang the mixin/mixout and smart_amp have exactly the same needs in terms of knowing which source/sink modules are connected at the time of copy. And for this the bind/unbin in the module_adapter and the module_source_info is enough. The smart_amp once converted to use the module_adapter interface, can use module_source_info_acquire() to query the connected sources.

@ranj063  we have different logic here. I want to totally remove bind & unbind even in module adapter, and let each module to process its individual event.  We can refine smart amp to use the logic of module adapter, what I concern is that we can do other event processing in bind & unbind in future, not just checking source & sink.

RanderWang:
hi @ranj063,  @andrula-song was assigned the task to adapt smart amp to module adapter interface. She first thought the task was blocked by bind & unbind, so wait for us. Now in my opinion,  the bind & unbind should be included in this task and she can start now. @aiChaoSONG also have a question about module adapter that how to find buffer with input stream.  He said he had discussed with you and you can check with him.

RanderWang:
@lyakh  please check the following code simplified from fw code . If means we need to create all the connected pipeline on the same core, or the result is unexpected. If we follow this rule in topology, can we make everything simple since all event happen on single core ?

````c
int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
{
          .....

	/* Pass IPC to target core if both modules has the same target core */
	if (!cpu_is_me(source->ipc_config.core) && source->ipc_config.core == sink->ipc_config.core)
		return ipc4_process_on_core(source->ipc_config.core, false);

	buffer = ipc4_create_buffer(source, sink, bu->extension.r.src_queue,
				    bu->extension.r.dst_queue);
	if (!buffer) {
		tr_err(&ipc_tr, "failed to allocate buffer to bind %d to %d", src_id, sink_id);
		return IPC4_OUT_OF_MEMORY;
	}
         .....
}
````

lyakh:
> Why we need proper locking ? If the module doesn't work on the ipc thread core, ipc4_process_on_core will be called and this message will be processed on the pipeline core, so disabling local interrupts should work.

@RanderWang I wrote that we'd need that for "cross-core streams," i.e. for cases when connected pipelines run on different cores

lyakh:
> I want to totally remove bind & unbind even in module adapter

@RanderWang can you explain why?

lyakh:
> @lyakh please check the following code simplified from fw code . It means we need to create all the connected pipeline on the same core, or the result is unexpected. If we follow this rule in topology, can we make everything simple since all event happen on single core ?

@RanderWang sure, life would remain "simple" if we could keep the current restriction that all connected pipelines must run on the same core. However we're told that in the future we'll have to support different cores too

RanderWang:
> > Why we need proper locking ? If the module doesn't work on the ipc thread core, ipc4_process_on_core will be called and this message will be processed on the pipeline core, so disabling local interrupts should work.
> 
> @RanderWang I wrote that we'd need that for "cross-core streams," i.e. for cases when connected pipelines run on different cores

@lyakh  I meant that the bind & unbind function can be dispatched to the core pipeline is on?  For two pipelines on different core, the bind & unbind event can be dispatched to different core, any issue ?

RanderWang:
> > I want to totally remove bind & unbind even in module adapter
> 
> @RanderWang can you explain why?

After first we didn't have bind & unbind ops in module adapter and at that time we discussed how to deal with these ops, then I planned to remove bind & unbind ops.

lyakh:
> > > Why we need proper locking ? If the module doesn't work on the ipc thread core, ipc4_process_on_core will be called and this message will be processed on the pipeline core, so disabling local interrupts should work.
> > 
> > 
> > @RanderWang I wrote that we'd need that for "cross-core streams," i.e. for cases when connected pipelines run on different cores
> 
> @lyakh I meant that the bind & unbind function can be dispatched to the core pipeline is on? For two pipelines on different core, the bind & unbind event can be dispatched to different core, any issue ?

@RanderWang yes - if your copying pipeline A runs on core 0 and your other pipeline B, that should be connected to pipeline A runs on core 1, you run `.bind()` on core 1 and you have a race

lyakh:
> > > I want to totally remove bind & unbind even in module adapter
> > 
> > 
> > @RanderWang can you explain why?
> 
> At first we didn't have bind & unbind ops in module adapter and at that time we discussed how to deal with these ops, then I planned to remove bind & unbind ops.

but now we do have it, so why remove it again?

RanderWang:
@andrula-song  please check the comments in this PR, it will be helpful

RanderWang:
> > > > I want to totally remove bind & unbind even in module adapter
> > > 
> > > 
> > > @RanderWang can you explain why?
> > 
> > 
> > At first we didn't have bind & unbind ops in module adapter and at that time we discussed how to deal with these ops, then I planned to remove bind & unbind ops.
> 
> but now we do have it, so why remove it again?

I don't want o remove it , just talk about my another idea.

PR #7260, opened 2023-03-13
Tools: Testbench/tplg_parser: Check -Wimplicit-fallthrough support

singalsu:
The xt-xcc compiler RG-2017.8 does not support -Wimplicit-fallthrough option. Add to CMakeLists.txt files check for the option and use it if it is supported.


src/arch/host/CMakeLists.txt
```diff
@@ -11,9 +12,16 @@ target_include_directories(sof_public_headers INTERFACE
 )
 endif()
 
+# -Wimplicit-fallthrough is preferred, check if it's supported
+check_c_compiler_flag(-Wimplicit-fallthrough supports_flag)
+if (supports_flag)
```
marc-hb, line 17:
```suggestion
if (supports_implicit_fallthrough)
```

singalsu: This is fixed now

marc-hb (approved):

lgirdwood (approved):

PR #7262, opened 2023-03-13
Clean up flexible arrays

lyakh:
Make sure flexible arrays are only used in the end of structures and use them instead of 0-size arrays where possible.


marc-hb:
Nice! Is there some new compiler flag we can now use to avoid future regressions?

https://people.kernel.org/kees/bounded-flexible-arrays-in-c

cujomalainey (approved):

lgirdwood (approved):

PR #7263, opened 2023-03-13
[SKIP CI] ci: add cherry-pick check for release branches

cujomalainey:
Add a non mandatory check for release branches that help reviewers verify all patches have landed on main before the patch to release branch.

The current configuration requires the cherry-pick command to be run with -x. This can be configured to do subject matching if that is too much for developers.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,25 @@
+---
+# SPDX-License-Identifier: BSD-3-Clause
+
+# yamllint disable-line rule:truthy
+on:
+  pull_request:
+    branches-ignore:
+      - main
+
+name: Validate
+
+# this job runs the following script
+# https://github.com/cujomalainey/cherry-pick-check/blob/main/entrypoint.sh
+
+jobs:
+  check:
+    name: Check Cherry Picks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Check all commits exist in main
+        uses: cujomalainey/cherry-pick-check@v1
+        with:
+          require_ref: true
+          main_branch: main


.github/workflows/release.yml
```diff
@@ -0,0 +1,18 @@
...
+name: Validate
+
+jobs:
+  check:
+    name: Check Cherry Picks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Check all commits exist in main
+        uses: cujomalainey/cherry-pick-check@v1
```
marc-hb, line 15: Can you add a convenience link to https://github.com/cujomalainey/cherry-pick-check/blob/main/entrypoint.sh ?

marc-hb (commented):
I like the idea.

Could you make https://github.com/cujomalainey/cherry-pick-check/blob/main/entrypoint.sh a bit more shellcheck-compliant?

Can you also add `set -e`?
- https://github.com/thesofproject/sof-test/issues/312

A `main` function would also be nice:
- https://github.com/thesofproject/sof-test/issues/740

.github/workflows/release.yml
cujomalainey: > Could you make https://github.com/cujomalainey/cherry-pick-check/blob/main/entrypoint.sh a bit more shellcheck-compliant?

Please criticize till your heart is content, this is the first serious shell script I have ever written

marc-hb: I think the "native"/preferred language for Github Action is Typescript (superset of Javascript) if you prefer that.

One advantage of TypeScript is I won't be able to review it :-D

cujomalainey: @marc-hb if I write it typescript, neither of us will be able to read it

marc-hb:
I have barely looked at https://github.com/cujomalainey/cherry-pick-check/blob/main/entrypoint.sh yet but based on the desired feature I just remembered it must be doing something like this old script of mine which is very well tested:
https://github.com/thesofproject/sof-test/blob/main/tools/CI/check-gitrange.bash

A while back I searched for a Github Action that would return the list of submitted commits but all the ones I found surprisingly sucked. I guess it's because most Github users use commits only for "fixups" https://github.com/zephyrproject-rtos/zephyr/issues/53566



There's also the shallow cloning problem: https://github.com/thesofproject/linux/issues/2556, fortunately sof.git is small enough.

Also on my way too long reading list: https://docs.github.com/en/actions/using-workflows/reusing-workflows

cujomalainey:
> I have barely looked at https://github.com/cujomalainey/cherry-pick-check/blob/main/entrypoint.sh yet but based on the desired feature I just remembered it must be doing something like this old script of mine which is very well tested: https://github.com/thesofproject/sof-test/blob/main/tools/CI/check-gitrange.bash
> 

That appears to be missing the parent branch check part though. That being said, I admit that is half the script.

cujomalainey:
@marc-hb shellcheck passes now, I also fixed yaml, if it looks good to you on action side I will push a new release and update the PR.

mwasko (approved):

lgirdwood (approved):

marc-hb (changes requested):
Script review in https://github.com/cujomalainey/cherry-pick-check/pull/1

PR #7271, opened 2023-03-14
ams: handle alloc failures in ams_init()

kv2019i:
Current code may end up dereferencing null pointers in case memory allocations fail in ams_init(). Handle this with proper error handling.

### DIFF
--- a/src/lib/ams.c
+++ b/src/lib/ams.c
@@ -587,6 +587,8 @@ int ams_init(void)
 
 	*ams = rzalloc(SOF_MEM_ZONE_SYS, SOF_MEM_FLAG_COHERENT, SOF_MEM_CAPS_RAM,
 		       sizeof(**ams));
+	if (!*ams)
+		return -ENOMEM;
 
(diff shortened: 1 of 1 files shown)


abonislawski:
> SOF_MEM_ZONE_SYS Fixed size heap where alloc always succeeds and is never freed. Used by any init code that will never give up the memory.

I think we are not checking SYS_ZONE allocs in sof or something changed?

lgirdwood:
> > SOF_MEM_ZONE_SYS Fixed size heap where alloc always succeeds and is never freed. Used by any init code that will never give up the memory.
> 
> I think we are not checking SYS_ZONE allocs in sof or something changed?

This checks need to be mandatory now as Zephyr allocator can change underneath us.

abonislawski (approved):

aiChaoSONG (approved):

kv2019i:
A few failures in https://sof-ci.01.org/sofpr/PR7271/build4571/devicetest/index.html but not related to AMS, so proceeding with merge.

PR #7272, opened 2023-03-14
volume: add validity check on number of channels

kv2019i:
Volume has multiple data arrays sized to SOF_IPC_MAX_CHANNELS and code that uses these arrays without bounds checks (like volume_ramp()).

Add a sanity check to volume_prepare() to ensure channels count is never set to a higher value.

### DIFF
--- a/src/audio/module_adapter/module/volume/volume.c
+++ b/src/audio/module_adapter/module/volume/volume.c
@@ -1300,7 +1300,13 @@ static int volume_prepare(struct processing_module *mod)
 	 * for entire topology specified time.
 	 */
 	cd->ramp_finished = false;
+
 	cd->channels = sink_c->stream.channels;
+	if (cd->channels > SOF_IPC_MAX_CHANNELS) {
+		ret = -EINVAL;
+		goto err;
+	}
+
 	cd->sample_rate_inv = (int32_t)(1000LL * INT32_MAX / sink_c->stream.rate);
 
 	buffer_release(sink_c);


src/audio/module_adapter/module/volume/volume.c
```diff
@@ -1300,7 +1300,13 @@ static int volume_prepare(struct processing_module *mod)
 	 * for entire topology specified time.
 	 */
 	cd->ramp_finished = false;
+
 	cd->channels = sink_c->stream.channels;
+	if (cd->channels > SOF_IPC_MAX_CHANNELS) {
```
lgirdwood, line 1305: if this is an array, won't we need to check >= here ?

lgirdwood:
@abonislawski pls review.

src/audio/module_adapter/module/volume/volume.c
kv2019i: @lgirdwood  The code will operate on the arrays with "for (i = 0; i < cd->channels; i++)" loops so it's ok to have ch->channels==SOF_IPC_MAX_CHANNELS.

singalsu (approved):

aiChaoSONG (approved):

kv2019i:
One alsabat failure in https://sof-ci.01.org/sofpr/PR7272/build4576/devicetest/index.html but not related to this PR. One not-run case in https://sof-ci.01.org/sofpr/PR7272/build4575/devicetest/index.html but otherwise good. Proceeding with merge.

Issue #7114, opened 2023-02-17
XCC RG-2017.8-linux: generated object code is affected by -g2 level (the default level) + longer source paths


marc-hb:
tl;dr: if you're failing to reproduce a build, try a longer or shorter `west topdir` source directory. That's the workaround. There seems to be a threshold somewhere around 27 characters.

Summary
-------

A very strange and complex toolchain issue where the `-g` level and debug symbols affect the _object code_ on some Intel CI build system(s) for a totally unknown reason.

Unexpected variations like this break build reproducibility = how this issue was discovered in the first place.

Variations like these could also make it harder to reproduce race conditions and other difficult bugs.

I focused on `.text.k_ticks_to_us_floor64` and RG-2017.8-linux below because it just happened to be the first difference I found but I noticed that the -g level also affects `.text.z_swap_irqlock` in the same conditions and there could be unexpected differences in other files.

   => Don't trust RG-2017.8-linux to produce consistent object code

This is not a problem with the Zephyr SDK which produces identical object code across Linux and Windows, as routinely verified in CI across all targets: https://github.com/thesofproject/sof/actions/runs/4189511798/jobs/7262044804

I haven't observed any problem with RI-2020.5-linux (and MTL) **yet** which Zephyr still identifies  as GNU 4.2.2. Note there is on-going work to switch to `xt-clang`.

Longer story
------------

Reproduced on SOF commit 64fe2ea7d2895 + `west update` + `sof/scripts/xtensa-build-zephyr.py -p tgl`

Reproduced with a couple other SOF commits too.

The code generated for the k_ticks_to_us_floor64() normally looks like this:

```
xt-objdump -s --section=.text.k_ticks_to_us_floor64 ../build-tgl/zephyr/kernel/CMakeFiles/kernel.dir/sched.c.obj

Contents of section .text.k_ticks_to_us_floor64:
 0000 3661000c 05b10000 61000041 00000c07  6a......a..A....
 0010 b08682b0 b6a2a2c8 fe87ba01 1bbb0c0d  ................
 0020 c2afff81 0000e008 00a0d4a2 a0e582a0  ................
 0030 c482b0f4 827cfb0c 0afaddea dd16fd07  .....|..........
 0040 9cea40a2 8250d282 40b2a240 c382dabb  ..@..P..@..@....
 0050 cabbdd07 cd068100 00e00800 3d0b2d0a  ............=.-.
 0060 1df0bd03 ad02dd07 cd068100 00e00800  ................
[...]

PR #7277, opened 2023-03-14
xtensa-build-xephyr: fix xt-objcopy failure when default-params missing

marc-hb:
Fixes recent commit 8aab18351fd6 ("xtensa-build-zephyr: fix DEFAULT_TOOLCHAIN_VARIANT spill on next platf")

Some XtensaTools installation are missing this `default-params` symbolic link:

```
XtensaTools/config/
|-- X4H3I16w2D48w3a_2017_8-params
|-- X6H3CNL_2017_8-params
|-- cavs2x_LX6HiFi3_2017_8-params
`-- default-params -> cavs2x_LX6HiFi3_2017_8-params
```

Maybe it's missing when installing with the graphical interface?

This symbolic link is surprisingly enough to make `xt-objcopy` work _without_ the XTENSA_ variables. But when the variables _and_ the link are both missing, then `xt-objcopy` fails with the usual error:

```
in current dir: work/current/sof; running command:
 XtDevTools/install/tools/RG-2017.8-linux/XtensaTools/bin/xt-objcopy
 --remove-section .comment sof/build-tgl/zephyr/zephyr.strip
       build-sof-staging/sof-info/tgl/stripped-zephyr.elf
Error: there is no Xtensa core registered as the default.

You need to either specify the name of a registered Xtensa core (with
the --xtensa-core option or the XTENSA_CORE environment variable) or
specify a different registry of Xtensa cores (with the --xtensa-system
option or the XTENSA_SYSTEM environment variable).

The following Xtensa cores are available:
	hifiep_bd5
	cavs2x_LX6HiFi3_2017_8
	sample_config
	sample_flix
	...
```

Fix this failure by simply passing the XTENSA_ variables to xt-objcopy.

Kudos to Seppo Ingalsuo for the interactive debugging session that allowed root-causing this problem extremely quickly.

### DIFF
--- a/scripts/xtensa-build-zephyr.py
+++ b/scripts/xtensa-build-zephyr.py
@@ -674,7 +674,7 @@ def build_platforms():
 		if platform not in RI_INFO_UNSUPPORTED:
 			reproducible_checksum(platform, west_top / platform_build_dir_name / "zephyr" / "zephyr.ri")
 
-		install_platform(platform, sof_platform_output_dir)
+		install_platform(platform, sof_platform_output_dir, platf_build_environ)
 
 	src_dest_list = []
 
@@ -706,7 +706,7 @@ def build_platforms():
 			symlinks=True, ignore_dangling_symlinks=True, dirs_exist_ok=True)
 
 
-def install_platform(platform, sof_platform_output_dir):
+def install_platform(platform, sof_platform_output_dir, platf_build_environ):
 
 	# Keep in sync with caller
 	platform_build_dir_name = f"build-{platform}"
@@ -797,7 +797,12 @@ def install_platform(platform, sof_platform_output_dir):
 		# Remove it like some gcc test scripts do:
 		# https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=c7046906c3ae
 		if "strip" in str(dstname):
-			execute_command([str(x) for x in [objcopy, "--remove-section", ".comment", src, dst]])
+			execute_command(
+				[str(x) for x in [objcopy, "--remove-section", ".comment", src, dst]],
+				# Some xtensa installs don't have a
+				# XtensaTools/config/default-params symbolic link
+				env=platf_build_environ,
+			)
 		elif f.txt:
 			dos2unix(src, dst)
 		else:


marc-hb:
alsabat failures in cavs2.5 https://sof-ci.01.org/sofpr/PR7277/build4610/devicetest/index.html as usual these days, everything else green.

The other device-test is stuck however.

scripts/xtensa-build-zephyr.py
```diff
@@ -797,7 +797,12 @@ class InstFile:
 		# Remove it like some gcc test scripts do:
 		# https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=c7046906c3ae
 		if "strip" in str(dstname):
-			execute_command([str(x) for x in [objcopy, "--remove-section", ".comment", src, dst]])
+			execute_command(
+				[str(x) for x in [objcopy, "--remove-section", ".comment", src, dst]],
+				# Some xtensa installs don't have a
+				# XtensaTools/config/default-params symbolic link
```
andyross, line 803: Not sure the comment is really needed, the vendor toolchain passes junk in the environment.  No reason to expect the binutils equivalents to work any differently from the compiler.

andyross (approved):
Ouch.  I've bumped into similar nonsense playing with "hand-installed" toolchains.  The toolchain install process in xtensa is really fragile.  It's *almost* enough to just tar up and copy the directory around, but some things break.  IIRC, in some versions of the toolchain that path is a local symlink, and sometimes it contains the full path of the original install tree.

scripts/xtensa-build-zephyr.py
marc-hb: Thanks for the review!

> The toolchain install process in xtensa is really fragile. It's almost enough to just tar up and copy the directory around, but some things break.

While investigating #7114 I had the... "opportunity" to compare installation across different machines. All the binaries are (fortunately!) identical and the only differences are absolute paths in a number of text config files (set by the installer script). It's much simpler than what I feared initially.

>  Not sure the comment is really needed, the vendor toolchain passes junk in the environment. No reason to expect the binutils equivalents to work any differently from the compiler.

I've successfully invoked `xt-objdump` many times without any `XTENSA_` variable thanks to this  `default-params` symlink - without realizing it until today! I think that symlink is worth knowing. Here for a lack of a better place.

singalsu (approved):
This change works for my Xplorer installation, while previous didn't, thanks!

lgirdwood (approved):

Issue #7270, opened 2023-03-14
[FEATURE] Move rimage toml config files to SOF repository


kv2019i:
**Is your feature request related to a problem? Please describe.**
The toml files in sof/rimage/config/*.toml have recently started covered properties of SOF modules that are no longer static properties of a platform, but rather are describiting attributes that can change when code is changed in SOF main. 

This highly problematic if a code to optimize a module's performance is merged to SOF main, a matching update is needed to rimage toml files (for all affected platforms), plus a separate sof/west.yml update to bring that in to SOF builds (with Zephyr where west is used).

**Describe the solution you'd like**
Rimage repository should only host the tool, not the actual platform-specific content.

**Describe alternatives you've considered**
Moving only IPC4 specific data is one option, but due to design of rimage, this seems rather complex and in the end the easier solution would be to move all of toml config files.

**Additional context**
- Recent rimage pull request history -> a sharp uptick in activity with IPC4 platforms:
https://github.com/thesofproject/rimage/pulls?q=is%3Apr+is%3Aclosed+sort%3Aupdated-desc
- https://github.com/thesofproject/sof/issues/8073

lgirdwood:
@mwasko @marcinszkudlinski fyi - obvious effort reduction here to keep toml manifest configs alongside FW code.

marc-hb:
> The toml files in sof/rimage/config/*.toml have recently started covered properties of SOF modules that are no longer static properties of a platform, but rather are describiting attributes that can change when code is changed in SOF main.

Wait: the static properties still exist AFAIK. So shouldn't rimage be able to read from more than one .toml file / from _both_ places? Would simply concatenating the files in the order given be enough?

kv2019i:
@marc-hb wrote:
> Wait: the static properties still exist AFAIK. So shouldn't rimage be able to read from more than one .toml file / from _both_ places? Would simply concatenating the files in the order given be enough? This would mean moving fewer files / less content 

But why should rimage (the tool) repo host even the static details? I agree we could split, but I'm not sure if there is any benefit to hosting in multiple repos. These are essentially the platforms we maintain in SOF repository. I.e. shouldn't rimage repo be more like a repo for gcc/clang in that it provides a tool that takes X in and spews Y out, but the input data X is hosted in other repos.

marc-hb:
After taking a look at https://github.com/thesofproject/sof-docs/pull/451 I wonder why we're even building the extended manifest using C code in 2023?

[...]

PR #7280, opened 2023-03-15
Copy rimage toml config files to SOF repository

aiChaoSONG:
This patch moves rimage/config to sof/config.
Firmware manifest configuration files should stay
with SOF instead of rimage tool.

Link: https://github.com/thesofproject/sof/issues/7270

If you don't think sof/config is a good place, please comment the place you think is good.

### DIFF
--- a/config/bdw.toml
+++ b/config/bdw.toml
@@ -1,0 +1,15 @@
+version = [1, 0]  # use simple file write
+
+[adsp]
+name = "bdw"
+
+[[adsp.mem_zone]]
+type = "IRAM"
+base = "0"
+size = "0x50000"
+host_offset = "0x000A0000"
+[[adsp.mem_zone]]
+type = "DRAM"
+base = "0x00400000"
+size = "0xA0000"
+host_offset = "0x0"
--- a/config/apl.toml
+++ b/config/apl.toml
@@ -1,0 +1,49 @@
+version = [1, 8]
+
+[adsp]
+name = "apl"
+image_size = "0x0A0000"        # (8 + 2) bank * 64KB
+
+[[adsp.mem_zone]]
+type = "ROM"
+base = "0xBEFE0000"
+size = "0x00002000"
+[[adsp.mem_zone]]
+type = "SRAM"
+base = "0xA000A000"
+size = "0x100000"
+
+[cse]
+partition_name = "ADSP"
+[[cse.entry]]
+name = "ADSP.man"
+offset = "0x58"
+length = "0x378"
+[[cse.entry]]
+name = "cavs0015.met"
+offset = "0x400"
+length = "0x60"
+[[cse.entry]]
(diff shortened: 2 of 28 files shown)


marc-hb:
I'd like to finish https://github.com/zephyrproject-rtos/zephyr/pull/54700 first.

There's also the question about whether it's good idea to keep growing the role of a user-space tool written in a low-level, unsafe and unreliable language.

aiChaoSONG:
I think this patch will impact the commit https://github.com/zephyrproject-rtos/zephyr/pull/54700/commits/810b0f273b8586bba99446228bc9e8364b468a38, but I think this one should go first, because in your commit, you use sof/rimage/config as the default dir, which is going to change very soon. this patch continues using `west sign` which I don't think you changed any option for it.

kv2019i:
@aiChaoSONG Thanks for doing this! Not sure why System/merge/build failed, but otherwise CI looks very promising.

@marc-hb Ack on #54700  but given it didn't make it to SOF2.5 and this PR is going to be very labour-intensive to keep up-to-date (as we have a continuous stream of rimage PRs to these toml files), I'd like to timebox it so that if #54700 is not done this week, we'll switch the priorities and do the rimage move first. This will have immediate benefits as the wave of rimage PRs is jus increasing.

aiChaoSONG:
SOFCI TEST

marc-hb:
```diff
- [RFC] Move rimage toml config files to SOF repository
+ [RFC] Copy rimage toml config files to SOF repository
```
I guess the removal will come later to ease the transition?

What is the transition strategy?

aiChaoSONG:
Update: 
- commit message revision, move -> copy because we don't delete rimage/config yet
- sync with rimage main, add newly merged rimage changes.

aiChaoSONG:
> What is the transition strategy? Don't forget there are two use cases:

- I think zephyr/west.yml doesn't use rimage.
- sof/west.yml, once the PR merged, we are using tomls in sof/config. we can keep the rimage commit currently used as long as there are no changes to rimage itself, or we can update to later rimage main.

marc-hb:
> I think zephyr/west.yml doesn't use rimage.

Internal zephyr CI does use rimage, it's required to test on hardware.

gkbldcig:

kv2019i:
@aiChaoSONG This can proceed now, but we need to keep the toml files in rimage as well to allow for a smooth transition.

config/bdw.toml
```diff
@@ -0,0 +1,15 @@
+version = [1, 0]  # use simple file write
```
lgirdwood, line 1: Intel legacy platforms are no longer in main branch - i.e. cavs2.5 and above is for main.

aiChaoSONG:
Close as we don't want a simple toml file moving, instead, we want to split platform config and algorithm config

PR #7281, opened 2023-03-15
aria: configure stream params

pblaszko:
Configure in/out buffer stream params.
In Windows scenarios, unconfigured values were causing zero-division exception in audio_stream_get_avail_frames.

Correct sample group size calculation for proper buffers management.

### DIFF
--- a/src/audio/aria/aria.c
+++ b/src/audio/aria/aria.c
@@ -151,7 +151,7 @@ static int init_aria(struct comp_dev *dev, const struct comp_ipc_config *config,
 
 	ibs = cd->base.ibs;
 	chc = cd->base.audio_fmt.channels_count;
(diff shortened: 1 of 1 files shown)


lgirdwood (commented):
This looks like 2 patches ? i.e. for regression bisectability

pblaszko:
> This looks like 2 patches ? i.e. for regression bisectability

Separated into two commits.

mwasko (approved):

softwarecki (approved):

iganakov (commented):
I think it would be better to add `.params` operation for Aria and set and verify stream params here, e.g using `comp_verify_params()` function or do it manually. Also I don't see component period frames were set for Aria

tmleman (approved):

lgirdwood (approved):

PR #7286, opened 2023-03-16
schedule: fix address space mismatches

aiChaoSONG:
Fix below sparse warnings:

zephyr_dp_schedule.c:134:20: warning: incorrect type in argument 1 (different address spaces)

zephyr_dp_schedule.c:348:36: warning: incorrect type in assignment (different address spaces)

introduced by 3ee1d787

### DIFF
--- a/src/schedule/zephyr_dp_schedule.c
+++ b/src/schedule/zephyr_dp_schedule.c
@@ -131,7 +131,7 @@ static int scheduler_dp_task_free(void *data, struct task *task)
 	scheduler_dp_unlock(lock_key);
 
 	/* free task stack */
-	rfree(pdata->p_stack);
+	rfree((__sparse_force void *)pdata->p_stack);
 
 	/* all other memory has been allocated as a single malloc, will be freed later by caller */
 	return 0;
@@ -276,7 +276,7 @@ int scheduler_dp_task_init(struct task **task,
 			   size_t stack_size,
 			   uint32_t task_priority)
 {
-	void *p_stack = NULL;
+	void __sparse_cache *p_stack = NULL;
 
 	/* memory allocation helper structure */
 	struct {
@@ -301,7 +301,8 @@ int scheduler_dp_task_init(struct task **task,
 
 	/* allocate stack - must be aligned so a separate alloc */
 	stack_size = Z_KERNEL_STACK_SIZE_ADJUST(stack_size);
-	p_stack = rballoc_align(0, SOF_MEM_CAPS_RAM, stack_size, Z_KERNEL_STACK_OBJ_ALIGN);
+	p_stack = (__sparse_force void __sparse_cache *)
+		rballoc_align(0, SOF_MEM_CAPS_RAM, stack_size, Z_KERNEL_STACK_OBJ_ALIGN);
 	if (!p_stack) {
 		tr_err(&dp_tr, "zephyr_dp_task_init(): stack alloc failed");
 		ret = -ENOMEM;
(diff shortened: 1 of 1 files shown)


aiChaoSONG:
SOFCI TEST

dbaluta (approved):

src/schedule/zephyr_dp_schedule.c
```diff
@@ -131,7 +131,7 @@ static int scheduler_dp_task_free(void *data, struct task *task)
 	scheduler_dp_unlock(lock_key);
 
 	/* free task stack */
-	rfree(pdata->p_stack);
+	rfree(z_soc_uncached_ptr(pdata->p_stack));
```
marcinszkudlinski, line 134: not needed, heap implementation works on both cached and uncached aliases. Changing to uncache here is just a waste of CPU cycles

src/schedule/zephyr_dp_schedule.c
```diff
@@ -345,7 +345,7 @@ int scheduler_dp_task_init(struct task **task,
 	/* success, fill the structures */
 	task_memory->task.priv_data = &task_memory->pdata;
 	task_memory->pdata.thread_id = thread_id;
-	task_memory->pdata.p_stack = p_stack;
+	task_memory->pdata.p_stack = z_soc_cached_ptr(p_stack);
```
marcinszkudlinski, line 348: here too, alloc will return cached pointer. no need to change
to be 100% accurate - a flag SOF_MEM_CAPS_CACHE indeed may be added to allocation
It won't change a bit in flow, just be more clear to read

task_memory = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM **| SOF_MEM_CAPS_CACHE**, sizeof(*task_memory));


p_stack = rballoc_align(0, SOF_MEM_CAPS_RAM **| SOF_MEM_CAPS_CACHE**, stack_size, Z_KERNEL_STACK_OBJ_ALIGN);

marcinszkudlinski (changes requested):
Good change to Zephyr kernel types no need however to play with cached/uncached aliases. Wasn't __sparse_cache macros dedicated to avoid warnings?

src/schedule/zephyr_dp_schedule.c
lyakh: the same pointer should be used for freeing, that was allocated. If it's a cached pointer, so far our solution is to type-case it to uncached, e.g. like in https://github.com/thesofproject/sof/blob/1319163ea2c6d8bea6fd7189dea6275ac5e34bf5/src/audio/module_adapter/module_adapter.c#L403

src/schedule/zephyr_dp_schedule.c
lyakh: Use this for `rballoc()` https://github.com/thesofproject/sof/blob/1319163ea2c6d8bea6fd7189dea6275ac5e34bf5/src/audio/module_adapter/module_adapter.c#L326-L327 and add `__sparse_cache` to `p_stack` definition

lyakh (changes requested):

marc-hb:
This makes https://github.com/thesofproject/sof/actions/runs/4433211392/jobs/7778032254?pr=7286  pass.

kv2019i (approved):

marcinszkudlinski (approved):

lyakh (approved):

lgirdwood (approved):

Issue #7240, opened 2023-03-10
[BUG] Headset capture does not work with S16_LE format on IPC4-SDW platforms


keqiaozhang:
**Describe the bug**
I was trying to enable alsabat test for headset capture on MTL platforms, I found that if I test capture with s16_le on MTL-SDW, the max peak always 99%-100%. The recorded wav is full of noise. But with same amixer settings, S32_LE works well. it can record the real data without noise. This issue should be a regression, but not sure how long it has existed. 

**Console output**
```
arecord -Dhw:0,1 -f s16_le -r 48000  -c 2 test.wav -vvv -d 5
Recording WAVE 'test.wav' : Signed 16 bit Little Endian, Rate 48000 Hz, Stereo
Hardware PCM card 0 'sof-soundwire' device 1 subdevice 0
Its setup is:
  stream       : CAPTURE
  access       : RW_INTERLEAVED
  format       : S16_LE
  subformat    : STD
  channels     : 2
  rate         : 48000
  exact rate   : 48000 (48000/1)
  msbits       : 16
  buffer_size  : 16384
  period_size  : 4096
  period_time  : 85333
  tstamp_mode  : NONE
  tstamp_type  : MONOTONIC
  period_step  : 1
  avail_min    : 4096
  period_event : 0
  start_threshold  : 1
  stop_threshold   : 16384
  silence_threshold: 0
  silence_size : 0
  boundary     : 4611686018427387904
  appl_ptr     : 0
  hw_ptr       : 0
Max peak (8192 samples): 0x00007ff6 #################### 99%
Max peak (8192 samples): 0x00007ffc #################### 99%
Max peak (8192 samples): 0x00007ffe #################### 99%
Max peak (8192 samples): 0x00008000 #################### 100%
Max peak (8192 samples): 0x00007ffc #################### 99%
Max peak (8192 samples): 0x00007ffe #################### 99%
Max peak (8192 samples): 0x00007ffe #################### 99%
Max peak (8192 samples): 0x00007ff0 #################### 99%
Max peak (8192 samples): 0x00007ff2 #################### 99%
Max peak (8192 samples): 0x00008000 #################### 100%
Max peak (8192 samples): 0x00007ff8 #################### 99%
Max peak (8192 samples): 0x00007ff8 #################### 99%
Max peak (8192 samples): 0x00008000 #################### 100%
Max peak (8192 samples): 0x00008000 #################### 100%
....
```
**To Reproduce**
[...]

PR #7289, opened 2023-03-16
topology2: cavs-sdw: set capture host in_valid_bit_depth = 32

bardliao:
For some reason the input valid bit depth was 16. But the output valid bit depth of the previous widget is 32. Fix it to 32 now.

Fixes: #7240


kv2019i (approved):
Looks correct.

PR #7291, opened 2023-03-16
Arch: Host: Update atomic functions for gcc and xt-xcc

singalsu:
This patch adds for host gcc build the updated atomic operations. These are intended to replace the previous __sync built-ins. For more information see
https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html .

The atomic functions for xcc are fake but sufficient for host arch build for static single-thread testbench.

Suggested-by: Guennadi Liakhovetski <guennadi.liakhovetski@linux.intel.com>

### DIFF
--- a/src/arch/host/include/arch/atomic.h
+++ b/src/arch/host/include/arch/atomic.h
@@ -17,9 +17,44 @@ typedef struct {
 	volatile int32_t value;
 } atomic_t;
 
+#if defined __GNUC__ && !defined __XCC__
+
 static inline int32_t arch_atomic_read(const atomic_t *a)
 {
-	return (*(volatile int32_t *)&a->value);
+	return __atomic_load_n(&a->value, __ATOMIC_SEQ_CST);
+}
+
+static inline void arch_atomic_set(atomic_t *a, int32_t value)
+{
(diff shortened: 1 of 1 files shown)


lyakh (commented):
it is a bit confusing, that the "host" arch is using xt-xcc, but I guess it has its use, probably to run testbench on the DSP? I guess we could eventually use https://github.com/thesofproject/sof/blob/main/src/arch/xtensa/hal/atomics.c for proper locking in that configuration - if we ever need it...

singalsu:
> it is a bit confusing, that the "host" arch is using xt-xcc, but I guess it has its use, probably to run testbench on the DSP? I guess we could eventually use https://github.com/thesofproject/sof/blob/main/src/arch/xtensa/hal/atomics.c for proper locking in that configuration - if we ever need it...

The testbench xcc build is only for xr-run simulation environment on a developer's development computer. It's very convenient for debugging the code, profiling and checking performance. Running it on a DSP native would not bring value.

The "host" could be renamed but I don't know what would be better.

lgirdwood (approved):

PR #7293, opened 2023-03-16
platform: mt8195: Support C++ linkage with native/vendor xcc

andyross:
[Minor build fix for C++ linkage with the existing toolchains.  This was developed against the mt8195/v0.4 branch obviously, but is being submitted to main for propriety.  Someone (@marc-hb ?) will have to clue me in about the SOF process for cross-patching and backports.]

At least some versions of the Cadence toolchain for this platform are built with a slightly variant libc (vs. the newlib build which seems more common for other SOF devices).  That's fine in principle, SOF doesn't call anything out of the libc anyway.  But when linking with the C++ standard libraries, it pulls symbols out that need to go in otherwise-unrecognized linker sections.  And it wants to see an _unlink_r() function (the reentrant worker underneath unlink()) defined for reasons that completely escape me.

### DIFF
--- a/src/platform/mt8195/mt8195.x.in
+++ b/src/platform/mt8195/mt8195.x.in
@@ -264,6 +264,7 @@ SECTIONS
     *(.rodata.*)
     *(.gnu.linkonce.r.*)
     *(.rodata1)
+    *(.clib.rodata)
     __XT_EXCEPTION_TABLE__ = ABSOLUTE(.);
     KEEP (*(.xt_except_table))
     KEEP (*(.gcc_except_table))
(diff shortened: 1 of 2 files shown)


marc-hb:
> Someone (@marc-hb ?) will have to clue me in about the SOF process for cross-patching and backports.]

I haven't seen anything codified. What I've seen and done is the usual:
- submit to main
- wait for something to break
- if not, cherry-pick -x to branch

@lgirdwood , @kv2019i ?

kv2019i:
@marc-hb wrote:
> @andyross > I haven't seen anything codified. What I've seen and done is the usual:
> 
> * submit to main
> * wait for something to break
> * if not, cherry-pick -x to branch

Ack, submit to main and then we have basicly owners for all older branches and it's their task to backport (or not to backport).

kv2019i (approved):

lgirdwood (approved):

Issue #7191, opened 2023-03-01
[BUG] ZEPHYR FATAL ERROR 4: ASSERTION FAIL [aligned_addr == addr] intel_adsp_hda.h:163


keqiaozhang:
**Describe the bug**
Observed this issue in CI daily test. From the kernel message, there's a IPC timed out `0x13020004`, but mtrace recorded a fatal exception and caused zephyr kernel panic on CPU0. 

**DMESG**
```
[ 3556.753931] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx      : 0x13020004|0x0: GLB_SET_PIPELINE_STATE
[ 3556.788076] kernel: snd_sof:sof_pcm_open: sof-audio-pci-intel-tgl 0000:00:1f.3: pcm: open stream 0 dir 1
[ 3556.788082] kernel: snd_sof:sof_pcm_open: sof-audio-pci-intel-tgl 0000:00:1f.3: period min 192 max 16384 bytes
[ 3556.788084] kernel: snd_sof:sof_pcm_open: sof-audio-pci-intel-tgl 0000:00:1f.3: period count 2 max 16
[ 3556.788086] kernel: snd_sof:sof_pcm_open: sof-audio-pci-intel-tgl 0000:00:1f.3: buffer max 65536 bytes
[ 3556.788185] kernel: snd_sof_intel_hda_common:hda_link_dma_hw_params: sof-audio-pci-intel-tgl 0000:00:1f.3: format_val=0x11, rate=48000, ch=2, format=2
[ 3556.788191] kernel: snd_sof:sof_pcm_hw_params: sof-audio-pci-intel-tgl 0000:00:1f.3: pcm: hw params stream 0 dir 1
[ 3556.788198] kernel: snd_sof_intel_hda_common:hda_dsp_stream_hw_params: sof-audio-pci-intel-tgl 0000:00:1f.3: FW Poll Status: reg[0x80]=0x40000 successful
[ 3556.788212] kernel: snd_sof_intel_hda_common:hda_dsp_stream_hw_params: sof-audio-pci-intel-tgl 0000:00:1f.3: FW Poll Status: reg[0x80]=0x40000 successful
[ 3556.788216] kernel: snd_sof_intel_hda_common:hda_dsp_stream_setup_bdl: sof-audio-pci-intel-tgl 0000:00:1f.3: period_bytes:0x4000
[...]

PR #7299, opened 2023-03-17
coherent: fix cache flush logic if coherent is not the first member

kv2019i:
If coherent object is at non-zero offset, the init call ends up invalidating data cache outside the allocated buffer.

All current users of "struct coherent" have the object as the first member, so nothing is broken with current code.

### DIFF
--- a/src/include/sof/coherent.h
+++ b/src/include/sof/coherent.h
@@ -171,7 +171,7 @@ static inline void *__coherent_init(size_t offset, const size_t size)
 	c->core = cpu_get_id();
 	list_init(&c->list);
 	/* inv local data to coherent object */
-	dcache_invalidate_region(uncache_to_cache(c), size);
+	dcache_invalidate_region(uncache_to_cache(object), size);
 
 	return object;
 }
@@ -260,7 +260,7 @@ static inline void *__coherent_init_thread(size_t offset, const size_t size)
 	c->core = cpu_get_id();
 	list_init(&c->list);
 	/* inv local data to coherent object */
-	dcache_invalidate_region(uncache_to_cache(c), size);
+	dcache_invalidate_region(uncache_to_cache(object), size);
 
 	return object;
 }


lyakh (approved):
yeah, fixing the same assumption in other locations like https://github.com/thesofproject/sof/blob/da9344a6ad725f1e41225b025345a742579bf0f4/src/include/sof/coherent.h#L191 https://github.com/thesofproject/sof/blob/da9344a6ad725f1e41225b025345a742579bf0f4/src/include/sof/coherent.h#L217 https://github.com/thesofproject/sof/blob/da9344a6ad725f1e41225b025345a742579bf0f4/src/include/sof/coherent.h#L239 (and more) would require some (internal) API changes

lgirdwood (approved):

ranj063 (approved):

lyakh:
> @kv2019i @lyakh I assume we need followups for
> 
>     1. coherent zalloc() alignment to cache line size ?
> 
>     2. remove the offset, i..e always at offset 0 ?

@lgirdwood I don't think we need those because we so far aren't aware of any bugs that those changes would fix. In fact I think we could add support for the coherent header anywhere within the managed object, apart from type-casting convenience I don't see any reason why it has to be in the beginning.

kv2019i:
SOFCI TEST

kv2019i:
A few failes in https://sof-ci.01.org/sofpr/PR7299/build4829/devicetest/index.html (PM fail, also seen on other PRs) an dhttps://sof-ci.01.org/sofpr/PR7299/build4830/devicetest/index.html . In the latter, once case of https://github.com/thesofproject/sof/issues/7191 was hit (https://sof-ci.01.org/sofpr/PR7299/build4830/devicetest/index.html?model=TGLU_RVP_NOCODEC_IPC4ZPH&testcase=multiple-pause-resume-5 ). Neither related to this PR, so proceeding with merge.

PR #7303, opened 2023-03-20
copier: correct host processed data in copier

btian1:
This patch followed patch #7210, copier also need change to get correct processed data.

### DIFF
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -1720,11 +1720,21 @@ static uint64_t copier_get_processed_data(struct comp_dev *dev, uint32_t stream_
 {
 	struct copier_data *cd = comp_get_drvdata(dev);
 	uint64_t ret = 0;
-	bool source = dev->direction == SOF_IPC_STREAM_CAPTURE;
 
-	if (cd->endpoint_num && cd->bsource_buffer != input) {
+	/*
+	 * total data processed is calculated as: accumulate all input data in bytes
+	 * from host, then store in host_data structure, this function intend to retrieve
+	 * correct processed data number.
+	 * Due to host already integrated as part of copier, so for host specific case,
+	 * as long as direction is correct, return correct processed data.
+	 * Dai still use ops driver to get data, later, dai will also be integrated
+	 * into copier, this part will be changed again for dai specific.
+	 */
+	if (cd->endpoint_num) {
 		if (stream_no < cd->endpoint_num) {
 			if (dev->ipc_config.type == SOF_COMP_HOST && !cd->ipc_gtw) {
+				bool source = dev->direction == SOF_IPC_STREAM_PLAYBACK;
+
 				if (source == input)
 					ret = cd->hd->total_data_processed;
 			} else {


btian1:
@lyakh , please check, this part also need change.

src/audio/copier/copier.c
```diff
@@ -1716,7 +1716,7 @@ static uint64_t copier_get_processed_data(struct comp_dev *dev, uint32_t stream_
 {
 	struct copier_data *cd = comp_get_drvdata(dev);
 	uint64_t ret = 0;
-	bool source = dev->direction == SOF_IPC_STREAM_CAPTURE;
+	bool source = dev->direction == SOF_IPC_STREAM_PLAYBACK;
```
lyakh, line 1719: I think you're right, yes. But please also move this variable definition to between lines 1723 and 1724 below - to where it's needed and where it's clear, that it's only used for the host type copier.
TBH, all the `if`s on lines 1721-1724 look suspicious to me, I wouldn't be surprised if they had errors. E.g. IIUC `endpoint_num` is always equal to 1 for DAI, host and gateway copier types and 0 otherwise. The check in line 1721 checks, that it's one of those copier types AND that the direction is correct. Now if it is one of those types but the direction is wrong, we take the `else` path on line 1731, is that correct? In fact that direction test should never fail AFAICS, so if it does fail it should be an error condition?

btian1: this is part of code from coe, can we use another patch to change it? or you want change it in this patch?

lyakh: > 2\. regarding move 1719 to host part, it is fine for this patch, however, check below, later dai will also have similar implementation, please check below:
>         [a899b9c](https://github.com/thesofproject/sof/commit/a899b9caa61f932368d0936b15c34e89b9553bc8)
>         also need move?

Yes, I'd move it there too. I find it usually better to limit the scope of variables to where they're used

lyakh: @softwarecki could you please double-check that the conditional logic there is correct and that my suspicions in https://github.com/thesofproject/sof/pull/7303/commits/24171f66946486d247dcd60e557cb361a8574e78#r1141840437 are wrong?

btian1: @lyakh , how about just merge this first and use another PR to track the logic changes?

lgirdwood (commented):
@marcinszkudlinski pls review.

src/audio/copier/copier.c
btian1: if (cd->endpoint_num && cd->bsource_buffer != input) {
		if (stream_no < cd->endpoint_num) {

change to:
if (cd->endpoint_num)
?

then else part will make more sense.
in dai, I already chagned to switch for this part:

			case  SOF_COMP_HOST:
				/* need handle gtw case separately,
				 *since cd->hd not assigned in this case.
				 */
				if (cd->ipc_gtw) {
					struct comp_dev *host_dev;

					host_dev = cd->endpoint[stream_no];
					ret = comp_get_total_data_processed(host_dev,
									    0, input);
				} else if (source == input) {
					ret = cd->hd->total_data_processed;
				}
				break;

			case  SOF_COMP_DAI:
				if (source == input)
					ret = cd->dd->total_data_processed;
				break;

			default:
				ret = comp_get_total_data_processed(cd->endpoint[stream_no],
							    0, input);
				break;

			}

lgirdwood (commented):
@btian1 Since this logic is complex, can you add some inline comments so everyone can align and understand the copier flow.

btian1:
> you add some

added, please check and review again.

src/audio/copier/copier.c
```diff
@@ -1716,18 +1716,25 @@ static uint64_t copier_get_processed_data(struct comp_dev *dev, uint32_t stream_
...
-								    0, input);
-			}
+	/*
+	 * bsource_buffer is assigned in create host/dai/gtw, however, host/dai
+	 * part are removing this flag, so this condition is not valid, remove
+	 * cd->bsource_buffer != input.
+	 * Also, cd->endpoint_num currently are always 1, and stream_no always 0
+	 * no need to check this part.
+	 */
+	if (cd->endpoint_num) {
```
lgirdwood, line 1727: Ok, so this is a good commit message and describes the patch code change.
But, the inline comment needs to describe the code and not the change. e.g.
```c
/* total data processed is calculated as follows ... describe calculation logic */
```

src/audio/copier/copier.c
```diff
@@ -1716,18 +1716,25 @@ static uint64_t copier_get_processed_data(struct comp_dev *dev, uint32_t stream_
...
+		if (dev->ipc_config.type == SOF_COMP_HOST && !cd->ipc_gtw) {
+			bool source = dev->direction == SOF_IPC_STREAM_PLAYBACK;
+
+			if (source == input)
+				ret = cd->hd->total_data_processed;
```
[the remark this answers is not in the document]
btian1: else part should be covered the ret's initialize, it is 0, see line 1718

src/audio/copier/copier.c
```diff
@@ -1720,18 +1720,23 @@ static uint64_t copier_get_processed_data(struct comp_dev *dev, uint32_t stream_
...
-			if (dev->ipc_config.type == SOF_COMP_HOST && !cd->ipc_gtw) {
-				if (source == input)
-					ret = cd->hd->total_data_processed;
-			} else {
-				ret = comp_get_total_data_processed(cd->endpoint[stream_no],
-								    0, input);
-			}
+	/*
+	 * total data processed is calculated as: accumulate all input data in bytes
+	 * from host, then store in host_data structure, this function intend to retrieve
```
lgirdwood, line 1726: ok, but we need to say why we have to chose between the two methods below.  why is it important that playback, type == host is used to select teh calculation.

btian1: ok， adding more comments.

src/audio/copier/copier.c
```diff
@@ -1720,18 +1720,27 @@ static uint64_t copier_get_processed_data(struct comp_dev *dev, uint32_t stream_
...
+	 * into copier, this part will be changed again for dai specific.
+	 */
+	if (cd->endpoint_num) {
+		if (dev->ipc_config.type == SOF_COMP_HOST && !cd->ipc_gtw) {
+			bool source = dev->direction == SOF_IPC_STREAM_PLAYBACK;
+
+			if (source == input)
+				ret = cd->hd->total_data_processed;
+		} else {
+			ret = comp_get_total_data_processed(cd->endpoint[stream_no],
```
lyakh, line 1740: we're now not evaluating `stream_no` at all in this function and here we're using it as an index into an array. I don't know what the original intention of this parameter was. I see that it's currently always 0. But maybe we should put something like
```
if (stream_no >= cd->endpoint_num) {
	comp_err();
	return 0;
}
```
between lines 1733 and 1734 above?

src/audio/copier/copier.c
```diff
@@ -1720,18 +1720,27 @@ static uint64_t copier_get_processed_data(struct comp_dev *dev, uint32_t stream_
...
+							    0, input);
 		}
+
```
lyakh, line 1743: we don't need this empty line

src/audio/copier/copier.c
```diff
@@ -1720,18 +1720,27 @@ static uint64_t copier_get_processed_data(struct comp_dev *dev, uint32_t stream_
...
+	/*
+	 * total data processed is calculated as: accumulate all input data in bytes
+	 * from host, then store in host_data structure, this function intend to retrieve
+	 * correct processed data number.
+	 * Due to host already integrated as part of copier, so for host specific case,
+	 * as long as direction is correct, return correct processed data.
+	 * Dai still use ops driver to get data, later, dai will also be integrated
+	 * into copier, this part will be changed again for dai specific.
+	 */
+	if (cd->endpoint_num) {
```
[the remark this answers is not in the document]
plbossart: 3 types @lyakh: host-gateway, dai-gateway and module.

src/audio/copier/copier.c
btian1: same result with:
if (stream_no < cd->endpoint_num) 

Since we are not able to see stream_no >= cd->endpoint_num, I removed this condition.

if need add, then we can add above code back, right? add above code back also have return 0, since ret initialized as 0.

RanderWang (approved):

src/audio/copier/copier.c
lyakh: yes, this achieves the same except it isn't printing an error

lyakh (approved):

src/audio/copier/copier.c
btian1: they are all done.
@kv2019i , please let me know anything left for merge?

kv2019i:
Timeout system suspend tests in one system in https://sof-ci.01.org/sofpr/PR7303/build5724/devicetest/index.html . Otherwise looks clean, so proceeding with merge.

PR #7306, opened 2023-03-20
Reduce DMA reload frequency on host DMA

lyakh:


src/audio/host-zephyr.c
```diff
@@ -454,9 +455,20 @@ static int host_copy_normal(struct host_data *hd, struct comp_dev *dev)
...
 		       NOTIFIER_TARGET_CORE_LOCAL, &next, sizeof(next));
-	ret = dma_reload(hd->chan->dma->z_dev, hd->chan->index, 0, 0, copy_bytes);
-	if (ret < 0)
-		comp_err(dev, "host_copy_normal(): dma_copy() failed, ret = %u", ret);
+
+	hd->partial_size += copy_bytes;
+	buffer_c = buffer_acquire(hd->dma_buffer);
+
+	/* Reload DMA when half of the buffer has been used */
+	if (hd->partial_size >= buffer_c->stream.size >> 1) {
```
ranj063, line 463: @lyakh seems like a good idea to reduce DMA traffic but do you think it makes sense to limit this to playback only?

lyakh: @ranj063 what effect would this have for capture? If the host buffer is large, it would delay triggering the host until more data is available, right? So, that means some additional latency. Do we use such larger buffers for capture? If we don't - that isn't a problem, if we do - isn't that the reason why we use them - to buffer more data in them before waking up the consumer?

src/audio/host-zephyr.c
```diff
@@ -454,9 +455,20 @@ static int host_copy_normal(struct host_data *hd, struct comp_dev *dev)
...
 	notifier_event(hd->chan, NOTIFIER_ID_DMA_COPY,
 		       NOTIFIER_TARGET_CORE_LOCAL, &next, sizeof(next));
-	ret = dma_reload(hd->chan->dma->z_dev, hd->chan->index, 0, 0, copy_bytes);
-	if (ret < 0)
-		comp_err(dev, "host_copy_normal(): dma_copy() failed, ret = %u", ret);
+
+	hd->partial_size += copy_bytes;
+	buffer_c = buffer_acquire(hd->dma_buffer);
+
+	/* Reload DMA when half of the buffer has been used */
```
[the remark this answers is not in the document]
lyakh: 50% because I had no idea and wanted to get some feedback - thank you! :-) Then why 20%? Should it maybe be a minimum of 4 periods and the whole buffer?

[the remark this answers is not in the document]
kv2019i: @mwasko @mmaka1 can you chime in on how this is supposed to work?

src/audio/host-zephyr.c
ranj063: I don't think we use large buffers for capture yet but I think there may be a plan to use 10ms buffers if not larger. But yes in that case, maybe it makes sense to do the same thing there as well.

src/audio/host-zephyr.c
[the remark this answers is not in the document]
ranj063: @lyakh , @yongzhi1 graciously agreed to do a quick experiment with deep buffer playback with this PR and reported a good increase in the lower pc-states residencies.

src/audio/host-zephyr.c
ranj063: BTW @lyakh do we know that the DMA transfer block size is equal to the size of the DMA buffer size in the case of deep buffer? It should be but just wondering if it really is or not.

lyakh: I think in my tests I saw the deep buffer size equal to 38400 bytes, i.e. 100 * 384, and 384 which is 32 bits * 2 channels * 48 and corresponds to 1ms at 48000 samples per second, i.e. it's a 100ms buffer, which seems correct. And I hope that isn't the size that the DMA transfers in one go? Or have I misunderstood?

src/audio/host-zephyr.c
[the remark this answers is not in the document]
lyakh: @mmaka1 If we do it in the DMA driver, the threshold should be configurable. Configuration parameters are passed in a `struct dma_config` structure. It doesn't have anything like a "threshold" but it has source and destination burst lengths. Should we use that?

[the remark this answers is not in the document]
kv2019i: @mmaka1 @lyakh What if we have this as a platform setting for now (maybe as a Kconfig)? We could use perhaps DMA attributes, but I wonder if the DMA driver can provide this info. The DMA driver will will likely not know much about how it's integrated to the system (i.e. what's the max latency to access host memory) .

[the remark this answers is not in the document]
ranj063: @lyakh this looks good to me!

src/audio/host-zephyr.c
```diff
@@ -459,8 +464,10 @@ static int host_copy_normal(struct host_data *hd, struct comp_dev *dev)
 	hd->partial_size += copy_bytes;
 	buffer_c = buffer_acquire(hd->dma_buffer);
 
-	/* Reload DMA when half of the buffer has been used */
-	if (hd->partial_size >= buffer_c->stream.size >> 1) {
+	/* Reload DMA when the threshold is reached, add a 2ms safety margin */
+	if (!IS_ENABLED(CONFIG_HOST_DMA_RELOAD_DELAY_ENABLE) ||
+	    buffer_c->stream.size < hd->period_bytes << 3 ||
+	    buffer_c->stream.size - hd->partial_size <= (2 + threshold) * hd->period_bytes) {
```
kv2019i, line 470: Some comment on "period_bytes << 3" would be nice as well.

kv2019i (approved):
I think there are lot of ways this could be parameterized, but I think as a starting point, this covers the basic need, is easy to understand and can be demonstrated to bring immediate savings, so I seems like a good addition.

lyakh:
The CI reported a failed test: https://sof-ci.01.org/sofpr/PR7306/build4998/devicetest/index.html?model=TGLU_RVP_NOCODEC_IPC4ZPH&testcase=check-playback-10sec which is strange. It's on a non-deep buffer stream, so this PR shouldn't affect that test. But I also don't see any open issues similar to that failure.

src/audio/host-zephyr.c
lyakh: @kv2019i comment updated

lyakh:
CI: no relevant failures this time

src/audio/host-zephyr.c
plbossart: why do we take the half-size into account. Not getting the idea.

plbossart (approved):
Nice work @lyakh

ranj063 (approved):

PR #7309, opened 2023-03-21
[DNM] kpb: add support for more sinks

fkwasowi:
kpb should support up to 4 clients

NOTE: The change cannot be tested at this time. Dual models such as Mslexa and Corlexa for MTL are only on gna, and converged does not yet support gna. The change will only be tested when gna is supported.

### DIFF
--- a/src/audio/kpb.c
+++ b/src/audio/kpb.c
@@ -77,8 +77,10 @@ struct comp_data {
 	struct task draining_task;
 	struct draining_data draining_task_data;
 	struct kpb_client clients[KPB_MAX_NO_OF_CLIENTS];
-	struct comp_buffer *sel_sink; /**< real time sink (channel selector)*/
-	struct comp_buffer *host_sink; /**< draining sink (client) */
+	/* Number of sinks depends on number of real time sink (channel selector)
+	 * and draining sinks (clients)
+	 */
+	struct comp_buffer *sinks[KPB_MAX_NO_OF_SINKS];
 	uint32_t kpb_no_of_clients; /**< number of registered clients */
 	uint32_t source_period_bytes; /**< source number of period bytes */
 	uint32_t sink_period_bytes; /**< sink number of period bytes */
@@ -186,8 +188,10 @@ static int kpb_set_verify_ipc_params(struct comp_dev *dev,
 	kpb->ipc4_cfg.base_cfg = ipc_config->base_cfg;
 
 	/* Initialize sinks */
-	kpb->sel_sink = NULL;
-	kpb->host_sink = NULL;
+	int sink_idx;
+
+	for (sink_idx = 0; sink_idx < KPB_MAX_NO_OF_SINKS; sink_idx++)
+		kpb->sinks[sink_idx] = NULL;
 
 	if (!kpb_is_sample_width_supported(kpb->config.sampling_width)) {
 		comp_err(dev, "kpb_set_verify_ipc_params(): requested sampling width not supported");
@@ -285,34 +289,36 @@ static int kpb_bind(struct comp_dev *dev, void *data)
 	bu = (struct ipc4_module_bind_unbind *)data;
 	buf_id = IPC4_COMP_ID(bu->extension.r.src_queue, bu->extension.r.dst_queue);
 
-	/* We're assuming here that KPB Real Time sink (kpb->sel_sink) is
+	/* We're assuming here that KPB Real Time sink (kpb->sinks[0]) is
 	 * always connected to input pin of Detector pipeline so during IPC4
 	 * Bind operation both src_queue and dst_queue will have id = 0
 	 * (Detector/MicSel has one input pin). To properly connect KPB sink
 	 * with Detector source we're looking for buffer with id=0.
 	 */
+	if (IPC4_MOD_ID(dev->ipc_config.id) == bu->primary.r.module_id)
+		list_for_item(blist, &dev->bsink_list) {
+			struct comp_buffer *sink = container_of(blist, struct comp_buffer,
+								source_list);
+			struct comp_buffer __sparse_cache *sink_c = buffer_acquire(sink);
+			int sink_buf_id;
 
-	list_for_item(blist, &dev->bsink_list) {
-		struct comp_buffer *sink = container_of(blist, struct comp_buffer, source_list);
-		struct comp_buffer __sparse_cache *sink_c = buffer_acquire(sink);
-		int sink_buf_id;
+			if (!sink_c->sink) {
+				ret = -EINVAL;
+				buffer_release(sink_c);
+				break;
+			}
 
-		if (!sink_c->sink) {
-			ret = -EINVAL;
+			sink_buf_id = sink_c->id;
 			buffer_release(sink_c);
-			break;
-		}
 
-		sink_buf_id = sink_c->id;
-		buffer_release(sink_c);
-
-		if (sink_buf_id == buf_id) {
-			if (sink_buf_id == 0)
-				kpb->sel_sink = sink;
-			else
-				kpb->host_sink = sink;
+			if (sink_buf_id == buf_id) {
+				if (sink_buf_id == 0)
+					kpb->sinks[0] = sink;
+				else
+					if (bu->extension.r.src_queue < KPB_MAX_NO_OF_SINKS)
+						kpb->sinks[bu->extension.r.src_queue] = sink;
+			}
 		}
-	}
 
 	return ret;
 }
@@ -327,18 +333,15 @@ static int kpb_unbind(struct comp_dev *dev, void *data)
 {
 	struct comp_data *kpb = comp_get_drvdata(dev);
 	struct ipc4_module_bind_unbind *bu;
-	int buf_id;
 
 	comp_dbg(dev, "kpb_bind()");
 
 	bu = (struct ipc4_module_bind_unbind *)data;
-	buf_id = IPC4_COMP_ID(bu->extension.r.src_queue, bu->extension.r.dst_queue);
 
 	/* Reset sinks when unbinding */
-	if (buf_id == 0)
-		kpb->sel_sink = NULL;
-	else
-		kpb->host_sink = NULL;
+	if (IPC4_MOD_ID(dev->ipc_config.id) == bu->primary.r.module_id)
+		if (bu->extension.r.src_queue < KPB_MAX_NO_OF_SINKS)
+			kpb->sinks[bu->extension.r.src_queue] = NULL;
 
 	return 0;
 }
@@ -361,8 +364,10 @@ static int kpb_set_verify_ipc_params(struct comp_dev *dev,
 	assert(!ret);
 
 	/* Initialize sinks */
-	kpb->sel_sink = NULL;
-	kpb->host_sink = NULL;
+	int sink_idx;
+
+	for (sink_idx = 0; sink_idx < KPB_MAX_NO_OF_SINKS; sink_idx++)
+		kpb->sinks[sink_idx] = NULL;
 
 	if (!kpb_is_sample_width_supported(kpb->config.sampling_width)) {
 		comp_err(dev, "kpb_set_verify_ipc_params(): requested sampling width not supported");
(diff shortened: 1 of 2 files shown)


src/audio/kpb.c
```diff
@@ -308,9 +312,9 @@ static int kpb_bind(struct comp_dev *dev, void *data)
 
 		if (sink_buf_id == buf_id) {
 			if (sink_buf_id == 0)
-				kpb->sel_sink = sink;
+				kpb->sinks[0] = sink;
 			else
-				kpb->host_sink = sink;
+				kpb->sinks[bu->extension.r.src_queue] = sink;
```
lyakh, line 317: should we check `bu->extension.r.src_queue` to avoid bad array access?

src/audio/kpb.c
```diff
@@ -336,9 +340,9 @@ static int kpb_unbind(struct comp_dev *dev, void *data)
 
 	/* Reset sinks when unbinding */
 	if (buf_id == 0)
-		kpb->sel_sink = NULL;
+		kpb->sinks[0] = NULL;
 	else
-		kpb->host_sink = NULL;
+		kpb->sinks[bu->extension.r.src_queue] = NULL;
```
lyakh, line 345: same here

src/audio/kpb.c
```diff
@@ -1417,6 +1434,47 @@ static int kpb_buffer_data(struct comp_dev *dev,
...
+{
+	int ret = 0;
+
+	comp_cl_info(&comp_kpb, "kpb_unregister_client()");
+
+	if (!cli) {
+		comp_cl_err(&comp_kpb, "kpb_unregister_client(): no client data");
+		return -EINVAL;
+	}
+	if (kpb->kpb_no_of_clients >= KPB_MAX_NO_OF_CLIENTS ||
```
lyakh, line 1457: this shouldn't happen, right? Such a kpb couldn't have been created? But I see that the same is checked in `kpb_register_client(()`

src/audio/kpb.c
```diff
@@ -1626,7 +1685,7 @@ static void kpb_init_draining(struct comp_dev *dev, struct kpb_client *cli)
 		comp_info(dev, "kpb_init_draining(), schedule draining task");
 
 		/* Add one-time draining task into the scheduler. */
-		kpb->draining_task_data.sink = kpb->host_sink;
+		kpb->draining_task_data.sink = kpb->sinks[cli->id + KPB_NO_OF_SELECTOR_SINK];
```
lyakh, line 1688: `sink_idx`

src/audio/kpb.c
```diff
@@ -1795,7 +1854,14 @@ static enum task_state kpb_draining_task(void *arg)
 	buffer_release(sink);
 
 	/* Reset host-sink copy mode back to its pre-draining value */
-	sink = buffer_acquire(kpb->host_sink);
+	int sink_idx;
+
+	for (sink_idx = 1; sink_idx < KPB_MAX_NO_OF_SINKS; sink_idx++) {
+		if (draining_data->sink->id == kpb->sinks[sink_idx]->id) {
```
lyakh, line 1860: in fact when you register clients you do
```
	kpb->clients[cli->id].id  = cli->id;
```
so the ID is always equal to the index. So you don't need to search, you can just use `kpb->sinks[draining_data->sink->id]`. Maybe there are more locations like this, please, check, e.g. in `kpb_copy()`

lyakh (changes requested):

src/audio/kpb.c
```diff
@@ -285,7 +289,7 @@ static int kpb_bind(struct comp_dev *dev, void *data)
 	bu = (struct ipc4_module_bind_unbind *)data;
 	buf_id = IPC4_COMP_ID(bu->extension.r.src_queue, bu->extension.r.dst_queue);
```
iganakov, line 291: Taking into consideration that we assigning kpb sinks here, code below should be executed only when kpb module is a source module during bind operation. E.g. you can add a check:                                                                                   `if (IPC4_MOD_ID(dev->ipc_config.id) = bu->primary.r.module_id)

src/audio/kpb.c
```diff
@@ -308,9 +312,9 @@ static int kpb_bind(struct comp_dev *dev, void *data)
 
 		if (sink_buf_id == buf_id) {
```
iganakov, line 313: It seems this could be simplified to 
```suggestion
		if (sink_buf_id == buf_id)
		        kpb->sinks[bu->extension.r.src_queue] = sink;
```

src/audio/kpb.c
```diff
@@ -336,9 +340,9 @@ static int kpb_unbind(struct comp_dev *dev, void *data)
```
iganakov, line 340: Also check if KPB is a source module when unbinding

src/audio/kpb.c
```diff
@@ -336,9 +340,9 @@ static int kpb_unbind(struct comp_dev *dev, void *data)
 
 	/* Reset sinks when unbinding */
 	if (buf_id == 0)
```
iganakov, line 342: Probably no need to check for `buf_id == 0 `, just assign NULL based on `src_queue`?

src/audio/kpb.c
```diff
@@ -1101,13 +1112,14 @@ static int kpb_copy(struct comp_dev *dev)
 {
 	int ret = 0;
 	struct comp_data *kpb = comp_get_drvdata(dev);
-	struct comp_buffer *source, *sink;
+	struct comp_buffer *source, *sink = NULL;
```
iganakov, line 1115: no need to assign NULL here, you have a check `if (!sink)` after every sink assignment below

lyakh: @iganakov it *is* now needed with this version of the code (at least formally) because of the `if` at the new line number 1229 - if that `if` is never entered, `sink` remains unassigned and is checked for `NULL` after the loop. But after the code is changed to take into account the fact, that `kpb->sinks[id]->id == id` always, this initialisation probably won't be needed again.

fkwasowi: Unfortunately this NULL is necessary, otherwise the PR could not pass all steps due to an error that an uninitialised variable could be used.

src/audio/kpb.c
lgirdwood: ack, if this comes from host we need to validate.

src/audio/kpb.c
lgirdwood: @fkwasowi best to share the compiler error so that @iganakov and @lyakh can resolve.

src/audio/kpb.c
```diff
@@ -1417,6 +1434,47 @@ static int kpb_buffer_data(struct comp_dev *dev,
...
+			    cli->id);
+		ret = -EINVAL;
+	} else {
+		kpb->clients[cli->id].drain_req = 0;
+		kpb->clients[cli->id].sink = NULL;
+		kpb->clients[cli->id].r_ptr = NULL;
+		kpb->clients[cli->id].state = KPB_CLIENT_UNREGISTERED;
+		kpb->kpb_no_of_clients--;
+		ret = 0;
+		kpb->sinks[cli->id + KPB_NO_OF_SELECTOR_SINK] = NULL;
```
lgirdwood, line 1473: Where do we vaidate ```cli->id``` is < array size + ```KPB_NO_OF_SELECTOR_SINK``` ?

src/audio/kpb.c
fkwasowi: Such a kpb should not be created

src/audio/kpb.c
fkwasowi: /home/sof/work/sof.git/src/audio/kpb.c: In function 'kpb_copy':
/home/sof/work/sof.git/src/include/sof/coherent.h:[318](https://github.com/thesofproject/sof/actions/runs/4596130915/jobs/8117209535?pr=7309#step:4:319):10: error: 'sink' may be used uninitialized in this function [-Werror=maybe-uninitialized]
  318 |   c->key = k_spin_lock(&c->lock);
/home/sof/work/sof.git/src/audio/kpb.c:1115:31: note: 'sink' was declared here
 1115 |  struct comp_buffer *source, *sink;
      |                               ^~~~
make[3]: *** [CMakeFiles/sof.dir/build.make:772: CMakeFiles/sof.dir/src/audio/kpb.c.o] Error 1
make[3]: *** Waiting for unfinished jobs....
make[2]: *** [CMakeFiles/Makefile2:1453: CMakeFiles/sof.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:2390: src/arch/xtensa/CMakeFiles/bin.dir/rule] Error 2
make: *** [Makefile:736: bin] Error 2
Error: Process completed with exit code 2.

sys-pt1s:

plbossart (changes requested):

src/audio/kpb.c
```diff
@@ -77,8 +77,10 @@ struct comp_data {
 	struct task draining_task;
 	struct draining_data draining_task_data;
 	struct kpb_client clients[KPB_MAX_NO_OF_CLIENTS];
-	struct comp_buffer *sel_sink; /**< real time sink (channel selector)*/
-	struct comp_buffer *host_sink; /**< draining sink (client) */
+	/* Number of sinks depends on number of real time sink (channel selector)
+	 * and draining sinks (clients)
```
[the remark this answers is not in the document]
fkwasowi: Depending on whether and which phrase is detected, a corresponding phrase will be drained on the corresponding sink.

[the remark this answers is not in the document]
plbossart: so is it possible that more than one client drains the buffer at the same time? Draining has historically been a problem, it needs to be very fast but avoid overloading the system, it's far from simple and the existing code for draining is questionable at best.

fkwasowi:
The change cannot be tested. Dual models such as Mslexa and Corlexa for MTL are only on gna, and converged does not yet support gna. The change will only be tested when gna is supported.

src/audio/kpb.c
```diff
@@ -1795,7 +1855,12 @@ static enum task_state kpb_draining_task(void *arg)
 	buffer_release(sink);
 
 	/* Reset host-sink copy mode back to its pre-draining value */
-	sink = buffer_acquire(kpb->host_sink);
+	for (sink_idx = 1; sink_idx < KPB_MAX_NO_OF_SINKS; sink_idx++) {
```
lyakh, line 1858: `for (sink_idx = KPB_NO_OF_SELECTOR_SINK;...`

src/audio/kpb.c
```diff
@@ -1213,7 +1224,12 @@ static int kpb_copy(struct comp_dev *dev)
 		break;
 	case KPB_STATE_HOST_COPY:
 		/* In host copy state we only copy to host buffer. */
-		sink = kpb->host_sink;
+		for (sink_idx = 1; sink_idx < KPB_MAX_NO_OF_SINKS; sink_idx++) {
```
lyakh, line 1227: `for (sink_idx = KPB_NO_OF_SELECTOR_SINK;...`

lgirdwood:
@fkwasowi any update ?

fkwasowi:
> @fkwasowi any update ?

Dual models such as mslexa and corlexa for MTL are only on gna, and converged does not yet support gna. The change will only be tested when gna is supported.

src/audio/kpb.c
```diff
@@ -285,34 +289,36 @@ static int kpb_bind(struct comp_dev *dev, void *data)
 	bu = (struct ipc4_module_bind_unbind *)data;
 	buf_id = IPC4_COMP_ID(bu->extension.r.src_queue, bu->extension.r.dst_queue);
 
-	/* We're assuming here that KPB Real Time sink (kpb->sel_sink) is
+	/* We're assuming here that KPB Real Time sink (kpb->sinks[0]) is
```
lyakh, line 292: does this assumption depend on the topology? So it's possible to construct a "wrong" topology where this won't be true but that otherwise will be valid?

src/audio/kpb.c
```diff
@@ -285,34 +289,36 @@ static int kpb_bind(struct comp_dev *dev, void *data)
...
-		if (sink_buf_id == buf_id) {
-			if (sink_buf_id == 0)
-				kpb->sel_sink = sink;
-			else
-				kpb->host_sink = sink;
+			if (sink_buf_id == buf_id) {
+				if (sink_buf_id == 0)
+					kpb->sinks[0] = sink;
+				else
+					if (bu->extension.r.src_queue < KPB_MAX_NO_OF_SINKS)
```
lyakh, line 318: just merge these two lines
```
else if (bu->extension.r.src_queue < KPB_MAX_NO_OF_SINKS)
```

lgirdwood:
@fkwasowi any update ?

fkwasowi:
> @fkwasowi any update ?

We are still waiting for the GNA, because then it will be possible to load a double model into wov

PR #7310, opened 2023-03-21
module_adapter: add queue ID to IO stream buffer for simple copy

aiChaoSONG:
Some components that have multiple inputs or outputs need to identify each input/output stream buffer, because the buffer may be put to special use. For example, the main input buffer and feedback input buffer for smart amp component must be identified, and can not be used interchangeably.

This patch adds queue ID to input/output stream buffer, so the component will know which buffer is special.

The buffer ID is composed by source/sink queue id in terms of route connection, so for input buffer of a widget, we need to get the sink queue ID of source buffer, vice versa for output buffer queue ID assignment.

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -20,6 +20,7 @@
 #include <sof/ut.h>
 #include <limits.h>
 #include <stdint.h>
+#include <ipc4/module.h>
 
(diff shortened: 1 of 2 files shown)


aiChaoSONG:
@ranj063 introduced module_source_info struct to find input queue id, which is safe in SMP case, close this.

PR #7313, opened 2023-03-21
Amd driver update for latest sof fw

maruthi-machani:


src/arch/xtensa/configs/rembrandt_defconfig
```diff
@@ -11,7 +11,7 @@ CONFIG_FORMAT_CONVERT_HIFI3=n
 CONFIG_LP_SRAM=n
 CONFIG_HAVE_AGENT=n
 CONFIG_COMP_VOLUME=n
-CONFIG_COMP_BLOB=n
+##CONFIG_COMP_BLOB=n
```
kv2019i, line 14: A few things:

- the commit messages would benefit from a bit more description on what is the intended change and why it is done (virtual dai is removed)

- keeping bisect working -- I think this breaks bisect as first commit removes definitions that are still used in the codebase. the tree should remain buildable after each commit in the series. this will ensure git bisect will work.

- minor style: the common style is to have space after prefixes in git titles, e.g. "drivers: amd: mux update"

src/platform/amd/rembrandt/include/platform/lib/memory.h
```diff
@@ -21,11 +21,11 @@
...
 
 #define DRAM0_BASE	0xE0000000
-#define DRAM0_SIZE	0x7000
+#define DRAM0_SIZE	0x10000
 #define SRAM0_BASE	0x9FF00000
 
-#define DRAM1_BASE      0xE0007000
-#define DRAM1_SIZE      0x19000
+#define DRAM1_BASE      0xE0010000
+#define DRAM1_SIZE      0x10000
```
kv2019i, line 28: Maybe have the memory size changes in a separate commit?

kv2019i (commented):
Code itself looks good, some notes on split to git commits and their descriptions.

src/arch/xtensa/configs/rembrandt_defconfig
```diff
@@ -11,7 +11,7 @@ CONFIG_FORMAT_CONVERT_HIFI3=n
 CONFIG_LP_SRAM=n
```
dbaluta, line 11: update for latest sof FW is something very generic and wont be very useful as a commit message in few months.

Please add something more specific. 

e.g drivers: amd: Fix AMD drivers

... give here reason for the fix pointing to commits that caused the change.

lgirdwood (approved):

PR #7315, opened 2023-03-21
host: remove unused code

lyakh:
After host functionality has been integrated into the copier, the host component interface became unused, remove it.


lyakh:
Also tested with IPC3 / legacy.
CI failures: https://sof-ci.01.org/sofpr/PR7315/build4912/devicetest/index.html?model=MTLP_RVP_NOCODEC&testcase=check-suspend-resume-with-capture-5 is a suspend-resume failure, https://sof-ci.01.org/sofpr/PR7315/build4913/devicetest/index.html?model=TGLU_RVP_NOCODEC_IPC4ZPH&testcase=multiple-pause-resume-5 is a multiple-pause-release. The latter might indeed indicate a bug in SOF, but cannot be related to this PR, which only removes complete functions and a complete host component driver, so any failure caused by it would show as a failure to build a pipeline.

ranj063:
@lyakh this would mean IPC3+zephyr would not be an option at all right?

lgirdwood:
> @lyakh this would mean IPC3+zephyr would not be an option at all right?

Its used on NXP platforms atm.

btian1:
better to remove it after some time? let the bullet fly for a while? in case, regression or other things happen.

lyakh:
> > @lyakh this would mean IPC3+zephyr would not be an option at all right?
> 
> Its used on NXP platforms atm.

@lgirdwood @ranj063 is IPC3+Zephyr ever used with native Zephyr drivers?..

juimonen:
@lyakh only some native drivers support ipc3 configs, so it would work only partially.

lyakh:
> @lyakh only some native drivers support ipc3 configs, so it would work only partially.

@juimonen and is it planned to support that configuration?

juimonen:
@lyakh my humble understanding was that ipc3 + zephyr native drivers was kind of transition phase, and official support is for ipc4 only. OTH it would not be too big task to add the code to dmic, alh and hda. ssp supports ipc3. However, authors for these drivers are different, so there is no 1 single entity to do this or send bug reports. So I would not hold my breath...

dbaluta:
> @lyakh this would mean IPC3+zephyr would not be an option at all right?

IPC3 + zephyr is used on NXP i.MX8 platforms. We are planning to transition to IPC4 but I don't see this happening in in Q2 or Q3.

@lgirdwood I think it is time for a new TSC .

lgirdwood:
> > @lyakh this would mean IPC3+zephyr would not be an option at all right?
> 
> IPC3 + zephyr is used on NXP i.MX8 platforms. We are planning to transition to IPC4 but I don't see this happening in in Q2 or Q3.
> 
> @lgirdwood I think it is time for a new TSC .

Ack, was planning this once v2.5 is tagged

RanderWang (commented):
It is good for me for ipc4 side.

src/audio/host-zephyr.c
```diff
@@ -1027,119 +852,8 @@ void host_zephyr_reset(struct host_data *hd, uint16_t state)
...
-	if (stream_no == 0 && source == input)
-		ret = hd->total_data_processed;
-
-	return ret;
-}
-
-static const struct comp_driver comp_host = {
-	.type	= SOF_COMP_HOST,
-	.uid	= SOF_RT_UUID(host_uuid),
-	.tctx	= &host_tr,
```
btian1, line 1118: You are making host-zephyr from a device to a small lib, so at least, title may need change, it is not simply remove code, with your changes, host-zephyr.c may change to host-zephyr-lib.c

ranj063: I don't think we need to change the name of the file here. lib could indicate that it is built as part of a library.

ranj063 (approved):
As long as we've agreed that we dont care about ipc3+zephyr anymore, this change is good.

lyakh:
> As long as we've agreed that we dont care about ipc3+zephyr anymore, this change is good.

@ranj063 I think it isn't that much about ipc3+Zephyr as about IPC3+Zephyr+native drivers. Whereas "native drivers" also seems not a very well defined condition. Does it mean all drivers? I.e. we cannot mix some drivers Zephyr native and some "legacy" SOF? I think that should be possible, right @juimonen? If it's possible, then in this case the condition is probably, that the DAI driver is a Zephyr driver?

juimonen:
@lyakh mixing sof xtos audio drivers and zephyr native audio drivers is possible in theory, but not in practice. In a sense that you either use native "glue" code or xtos apis on sof. We don't have so refined separation as per driver you could choose which one to use. Terms have been heavily convoluted, but as said, trying to compile ipc3 sof with some audio drivers in zephyr side and some in xtos would not just work (unless you heavily modify makefiles). And as a consequence, zephyr native drivers with ipc3 will not be fully functional.

dbaluta:
> As long as we've agreed that we dont care about ipc3+zephyr anymore, this change is good.

@ranj063 at least this year for i.MX we care about it. We are planning to completely switch to Zephyr but we are still using IPC3

ranj063:
> > As long as we've agreed that we dont care about ipc3+zephyr anymore, this change is good.
> 
> @ranj063 at least this year for i.MX we care about it. We are planning to completely switch to Zephyr but we are still using IPC3

@dbaluta did you see @juimonen 's comment? How are you testing this?

lgirdwood:
> > As long as we've agreed that we dont care about ipc3+zephyr anymore, this change is good.
> 
> @ranj063 at least this year for i.MX we care about it. We are planning to completely switch to Zephyr but we are still using IPC3

@dbaluta pls block this PR - we can do the cleanup after NXP is ready. @lyakh we can remove code that is unused on Intel platforms only for IPC3 + Zephyr.

dbaluta (changes requested):
blocking this until we get a chance to test it

Cc @paulstelian97 @iuliana-prodan

sys-pt1s:

btian1:
@lyakh , this is also a memory optimization, if possible, please provide fw size reduction for this PR, I believe should > 1K

lyakh:
> @lyakh , this is also a memory optimization, if possible, please provide fw size reduction for this PR, I believe should > 1K

We can also `#ifdef` that code to achieve the same memory saving without potentially breaking any IPC3+Zephyr+native_driver non-Intel configurations. But that would of course be uglier

lgirdwood:
@dbaluta any update ?

dbaluta:
Looking at this:

```
if(CONFIG_ZEPHYR_NATIVE_DRIVERS)
»       zephyr_library_sources(
»       »       ${SOF_AUDIO_PATH}/host-zephyr.c                                                                           
»       )
else()
»       zephyr_library_sources(
»       »       ${SOF_AUDIO_PATH}/host-legacy.c
»       )
endif()
```

It looks like this doesn't affect us right now because `NATIVE_DRIVERS` is not enabled for IMX. 

Anyhow, does this mean that when we will need to switch to native drivers will also need to switch to IPC4?

dbaluta (approved):

dbaluta:
This should be OK to merge after resolving the conflicts.

kv2019i:
@lyakh Status? Not marked as draft, has conflicts, "RFC" title, but then has a full set approvals.

lyakh:
> @lyakh Status? Not marked as draft, has conflicts, "RFC" title, but then has a full set approvals.

@kv2019i this PR breaks IPC3 + native Zephyr drivers. We can remove IPC3 support from "main" for Intel platforms, but from the discussion above it looks like other platforms might want to use that configuration at least temporarily for driver porting to Zephyr. In that case we probably shouldn't break it. I'll convert this to draft

PR #7319, opened 2023-03-22
topology2: cleanup basic definitions

plbossart:
I don't know how we managed to have so many blatant logical inversions and inconsistencies.

dai_in/out
ibs/obs
aif_in/aif_out
capture/playback direction and names

and all of the above.

If it looks bad it's because it is. We need stronger class definitions to avoid such issues.

### DIFF
--- a/tools/topology/topology2/cavs-src-mixin-mixout-hda.conf
+++ b/tools/topology/topology2/cavs-src-mixin-mixout-hda.conf
@@ -73,7 +73,7 @@ Object.Pipeline {
 		}
 	]
 
-	passthrough-be [
+	passthrough-playback-be [
 		{
 			index		4
 			direction	capture
--- a/tools/topology/topology2/cavs-src-mixin-mixout-hda.conf
+++ b/tools/topology/topology2/cavs-src-mixin-mixout-hda.conf
@@ -73,10 +73,9 @@ Object.Pipeline {
 		}
 	]
 
-	passthrough-playback-be [
+	passthrough-capture-be [
 		{
 			index		4
-			direction	capture
 
 			Object.Widget.copier."1" {
 				dai_type 	"HDA"
--- a/tools/topology/topology2/include/pipelines/cavs/deepbuffer-playback.conf
+++ b/tools/topology/topology2/include/pipelines/cavs/deepbuffer-playback.conf
@@ -49,6 +49,7 @@ Class.Pipeline."deepbuffer-playback" {
 			copier_type	"host"
 			type	"aif_in"
 			node_type $HDA_HOST_OUTPUT_CLASS
+
 			num_audio_formats 3
 			num_input_audio_formats 3
 			num_output_audio_formats 3
--- a/tools/topology/topology2/include/pipelines/cavs/deepbuffer-playback.conf
+++ b/tools/topology/topology2/include/pipelines/cavs/deepbuffer-playback.conf
@@ -67,7 +67,7 @@ Class.Pipeline."deepbuffer-playback" {
 				out_bit_depth		32
 				out_valid_bit_depth	32
 				# 100ms buffer
-				dma_buffer_size "$[$obs * $DEEPBUFFER_FW_DMA_MS]"
+				dma_buffer_size "$[$ibs * $DEEPBUFFER_FW_DMA_MS]"
 			}
 			# 32-bit 48KHz 2ch
 			Object.Base.audio_format.3 {
--- a/tools/topology/topology2/include/pipelines/cavs/mixout-gain-dai-copier-playback.conf
+++ b/tools/topology/topology2/include/pipelines/cavs/mixout-gain-dai-copier-playback.conf
@@ -47,7 +47,7 @@ Class.Pipeline."mixout-gain-dai-copier-playback" {
 	Object.Widget {
 		mixout."1" {}
 		copier."1" {
-			type dai_in
+			type dai_out
 			node_type $HDA_LINK_OUTPUT_CLASS
 			period_sink_count 0
 			period_source_count 2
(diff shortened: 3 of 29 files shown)


marc-hb:
>  so many blatant logical inversions and inconsistencies.

Do we have no test coverage for these? If not, why and how were they submitted?

tools/topology/topology2/platform/intel/sdw-amp-generic.conf
```diff
@@ -163,7 +163,7 @@ IncludeByKey.SDW_AMP_FEEDBACK {
 				}
 			]
 
-			passthrough-be [
+			passthrough-capture-be [
```
ranj063, line 164: @plbossart this is not the wrong direction. The original passthrough-be is applicable for both directions but the passthrough-capture-be limits to 32-bit support only

tools/topology/topology2/include/pipelines/cavs/passthrough-capture.conf
```diff
@@ -16,9 +16,9 @@
 # Where N is the unique pipeline ID within the same alsaconf node.
 #
 
-<include/common/audio_format.conf>
 <include/components/copier.conf>
 <include/components/pipeline.conf>
+<include/common/audio_format.conf>
```
ranj063, line 21: is this an unrelated change? the original order is in alphabetical order no?

tools/topology/topology2/cavs-src-mixin-mixout-hda.conf
```diff
@@ -73,7 +73,7 @@ Object.Pipeline {
 		}
 	]
 
-	passthrough-be [
+	passthrough-playback-be [
```
ranj063: this is wrong @plbossart. Like I mentioned passthrough-be is not only for playback. If you look at the direction below, it is meant for capture. So changing this to playback is wrong

tools/topology/topology2/include/pipelines/cavs/passthrough-playback-be.conf
```diff
@@ -32,17 +32,12 @@ Class.Pipeline."passthrough-playback-be" {
...
-		# mixin-playback objects instantiated within the same alsaconf node must have
-		# unique pipeline_id attribute
-		#
 		unique	"instance"
 	}
 
 	Object.Widget {
 		copier."1" {
-			type dai_in
-			node_type $HDA_LINK_OUTPUT_CLASS
```
ranj063, line 45: if you remove the default node_type, I think we should make sure it is set in all top-level conf files first. I think we avoid setting it for HDA

tools/topology/topology2/include/pipelines/cavs/deepbuffer-playback.conf
```diff
@@ -49,6 +49,7 @@ Class.Pipeline."deepbuffer-playback" {
 			copier_type	"host"
 			type	"aif_in"
 			node_type $HDA_HOST_OUTPUT_CLASS
+
```
ranj063, line 52: spurious change

tools/topology/topology2/include/pipelines/cavs/mixout-gain-dai-copier-playback.conf
```diff
@@ -47,7 +47,7 @@ Class.Pipeline."mixout-gain-dai-copier-playback" {
 	Object.Widget {
 		mixout."1" {}
 		copier."1" {
-			type dai_in
+			type dai_out
```
ranj063, line 50: I am not sure about this one @plbossart . I think you made too many changes in this PR that are not correct and you've got failures left, right & center. It might be a good idea to go step by step and separate the clean ups one by one

tools/topology/topology2/include/pipelines/cavs/passthrough-playback.conf
```diff
@@ -62,7 +62,7 @@ Class.Pipeline."passthrough-playback" {
 				in_valid_bit_depth	24
 				out_bit_depth		32
 				out_valid_bit_depth	32
-				dma_buffer_size "$[$obs * 2]"
+				dma_buffer_size "$[$ibs * 2]"
```
ranj063, line 66: and oh I forgot t mention, we need to remove dma_buffer_size altogether. This is no longer used by the kernel and it computed based on ibs/obs and stream direction

tools/topology/topology2/include/pipelines/cavs/passthrough-capture.conf
plbossart: it's to align passthrough-capture and passthrough-playback. One of them doesn't follow the convention. I don't care what the convention is, I just want the two files to be aligned.

tools/topology/topology2/cavs-src-mixin-mixout-hda.conf
plbossart: It's all needlessly complicated. No wonder no one can figure this out.

tools/topology/topology2/include/pipelines/cavs/passthrough-playback-be.conf
plbossart: I don't know how to progress. This is wrong in general. It should never have been set this way, if there is an attribute that is not valid by default across all uses, then it cannot be initialized by default in the base class.

tools/topology/topology2/include/pipelines/cavs/deepbuffer-playback.conf
plbossart: no, this is intentional. 

this was to make sure all uses of the copier_type, type and node_type are well isolated to make sure we can detect any differences.

tools/topology/topology2/platform/intel/sdw-amp-generic.conf
plbossart: And btw why is this named passthrough in the first place. This should be 'copier-be' or 'generic-be'. The passthrough property depends on how copiers are connected. A copier does not have in itself any passthrough attribute.

gkbldcig:

plbossart:

broken CI @lgirdwood

plbossart:
> @plbossart any chance you can split this PR up as it could be only 1 or 2 patches that are problematic. i.e. we can apply the fixes that are passing CI and revisiste the failure patches ?

I'll close this PR and let @ranj063 take over. I broke too many eggs with my changes.

PR #7250, opened 2023-03-10
Update rimage to a version, that doesn't need Zephyr ELF mangling

lyakh:
Upgrade rimage but keep Zephyr for now

PR #7320, opened 2023-03-22
.github/zephyr.yml: add a temporary submodules/west consistency check

marc-hb:
Real-world experience has proved again that big READMEs are not enough, not even when they're only a couple lines away as the one added in commit 8fd351ea9aa6 ("west.yml: add warning to keep git submodules in sync"). Only some failure / red color stand a chance.

It also seems some people rarely use "git status". This was discovered in commit d9eb16aa6655 ("cmake: add warning when git submodule changes are found") but is still surprising.


marc-hb:
`--filter=tree:0` in the first commit made `west update` in https://github.com/thesofproject/sof/actions/runs/4486669431/jobs/7889366353?pr=7320 twice faster than in (for example) https://github.com/thesofproject/sof/actions/runs/4485522295/jobs/7887155283?pr=7319

marc-hb:
The "git submodules" consistency check added by the 2nd commit failed in https://github.com/thesofproject/sof/actions/runs/4486669431/jobs/7889366353?pr=7320 which is exactly its purpose. This was just broken by PR #7250

abonislawski (approved):

softwarecki (approved):

.github/workflows/zephyr.yml
```diff
@@ -39,6 +39,24 @@ jobs:
...
+        run: |
+
+          cd workspace/sof
+          git submodule update --init --recursive
+
+          west update
+          if git status --porcelain=v2 | grep ^ ; then
+             git status
+             echo 'FAIL: inconsistency between git submodules and west.yml!'
+             echo 'Always use "git status"'
```
lgirdwood, line 56: What are you telling people to do here ? By itself ```git status``` wont fix the problem.
Btw, why not just convert rimage to west and avoid temp W/A ?

marc-hb: > What are you telling people to do here ? By itself git status wont fix the problem.

The error message says "inconsistency between git submodules and west.yml" and the check name is "git submodule consistency". There's also a comment in west.yml above the line that people changed.

> Btw, why not just convert rimage to west and avoid temp W/A ?

That's explained in the comment above. rimage is already west, the longer-term fix is to change its location so it does coincide with the XTOS submodule. However this will (temporarily) break CI as every rimage change as we just saw in https://github.com/zephyrproject-rtos/zephyr/pull/54700 and https://github.com/zephyrproject-rtos/zephyr/pull/56099

marc-hb: @lgirdwood I added a direct reference to the detailed rimage submodule instructions in sof/west.yml, I should have done that from the start.

PR #7325, opened 2023-03-24
MTL enters D3 power state using Zephyr Power Manager

aborisovich:
This PR enables Zephyr IPC driver to enter/leave D3 power state using Zephyr Power Manager.

Commit 98f8cf214844700585fa277fc66245f212e20a27 adds pm notifiers that are triggered every time power state is changed.
Commit 8e4eeab67ebb8baf9e3ceb5db01f5301e889a34d adds IPC4 driver resume/suspend capability.
Commit a6e531b81d0b0ecea7881941e046dc8969bd3d1d enables IMR context save/restore.
Commit e17fd43847da9f61836406b85990ef4d97c8f3c1 provides error handling mechanism for IPC response to Host when errors during IPC power transition are encountered.

Dependecies:

- [x] https://github.com/zephyrproject-rtos/zephyr/pull/56168
- [x] https://github.com/thesofproject/sof/pull/7857


src/platform/intel/ace/platform.c
```diff
@@ -73,6 +77,59 @@ int platform_boot_complete(uint32_t boot_message)
...
+		storage_buffer_size = tlb_api->get_storage_size();
+
+		/* add space for LPSRAM */
+		storage_buffer_size += LP_SRAM_SIZE;
+
+		/* allocate IMR buffer and store it in the global pointer */
+		global_imr_ram_storage = rmalloc(SOF_MEM_ZONE_SYS,
+						 0,
+						 SOF_MEM_CAPS_L3,
+						 storage_buffer_size);
```
tmleman, line 118: Should be done only for the primary core.

src/platform/intel/ace/platform.c
```diff
@@ -73,6 +77,59 @@ int platform_boot_complete(uint32_t boot_message)
...
+		rfree(global_imr_ram_storage);
+		global_imr_ram_storage = NULL;
+
+		/* zero the mask before D3 context restore */
+		struct ipc *ipc = ipc_get();
+		ipc->task_mask = 0;
+		ipc->pm_prepare_D3 = false;
+
+		/* send FW Ready message */
+		platform_boot_complete(0);
```
tmleman: dito

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -14,6 +14,7 @@ CONFIG_COMP_SRC_IPC4_FULL_MATRIX=y
 CONFIG_PM=y
 CONFIG_PM_DEVICE=y
 CONFIG_PM_DEVICE_RUNTIME=y
+CONFIG_PM_DEVICE_RUNTIME_EXCLUSIVE=y
```
tmleman, line 17: do we need this?

src/platform/intel/ace/platform.c
```diff
@@ -115,12 +133,18 @@ static void notify_pm_state_exit(enum pm_state state)
 		rfree(global_imr_ram_storage);
 		global_imr_ram_storage = NULL;
 
-		/* zero the mask before D3 context restore */
-		struct ipc *ipc = ipc_get();
-		ipc->task_mask = 0;
-		ipc->pm_prepare_D3 = false;
+		mem_window_idle_exit();
+		sys_clock_idle_exit();
```
tmleman, line 137: I think this should go here: https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/xtensa/intel_adsp/ace/power.c#L293-L317
It's best to keep all things related to HW on the kernel side.

tmleman (changes requested):

src/ipc/ipc-zephyr.c
```diff
@@ -79,6 +83,62 @@ static bool message_handler(const struct device *dev, void *arg, uint32_t data,
...
+
+	// we are not entering D3 - return error code bad message
+	if (!(ipc->task_mask & IPC_TASK_POWERDOWN)) {
+		return -EBADMSG;
+	}
+
+	// we are not entering D3 - power state change is not caused by IPC message
+	// from host that would disable the lock.
+	if(pm_policy_state_lock_is_active(PM_STATE_SOFT_OFF, PM_ALL_SUBSTATES))
+		return -EACCES;
```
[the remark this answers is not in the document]
aborisovich: It is unlocked in the line 195:
```
#if defined(CONFIG_PM)
		// unlock policy locks set during firmware init
		pm_policy_state_lock_put(PM_STATE_SOFT_OFF, PM_ALL_SUBSTATES);
```
And locked back in the body of `ipc_device_resume_handler`:
`pm_policy_state_lock_get(PM_STATE_SOFT_OFF, PM_ALL_SUBSTATES);`

src/platform/intel/ace/platform.c
aborisovich: And it is.
Callbacks `notify_pm_state_entry` and `notify_pm_state_exit` are registered in the body of the:
```
/* Runs on the primary core only */
int platform_init(struct sof *sof)
```
function. I'll write some docs.

app/boards/intel_adsp_ace15_mtpm.conf
aborisovich: We don't, we do need to trigger pm hooks.

gkbldcig:

src/platform/intel/ace/platform.c
aborisovich: is this ok now?

tmleman: I don't see any changes. If you mean your comment then no.

This comment in the code is about the _**platform_init**_ function, and it's true. But the notifications you register are global and will be called from every core.

aborisovich: Sorry, I added comments and forgot to push. Now the docs should be visible.

src/ipc/ipc-zephyr.c
```diff
@@ -79,6 +83,62 @@ static bool message_handler(const struct device *dev, void *arg, uint32_t data,
...
+{
+	struct ipc *ipc = (struct ipc *)arg;
+
+	ipc_set_drvdata(ipc, NULL);
+	ipc->task_mask = 0;
+	ipc->pm_prepare_D3 = false;
+
+	// re-enable D3 power state lock
+	pm_policy_state_lock_get(PM_STATE_SOFT_OFF, PM_ALL_SUBSTATES);
+
```
tmleman, line 127: This state is is entered by force. That means the all locks are ignored. I don't see why we want to disable lock after pm_state_force.

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -20,6 +20,8 @@ CONFIG_PM_POLICY_CUSTOM=y
 CONFIG_POWER_DOMAIN=y
 CONFIG_POWER_DOMAIN_INTEL_ADSP=y
 
+CONFIG_ADSP_IMR_CONTEXT_SAVE=y
```
kv2019i, line 23: @marcinszkudlinski Is there's a way to avoid context save at runtime? Boot with purge-bit set? FYI to @ujfalusi @plbossart

src/platform/intel/ace/platform.c
```diff
@@ -23,6 +23,10 @@
 #include <ipc/info.h>
```
kv2019i, line 23: minor nit: typo in git summary "trasition" -> "transition"

src/platform/intel/ace/platform.c
```diff
@@ -73,6 +77,62 @@ int platform_boot_complete(uint32_t boot_message)
...
+				(struct intel_adsp_tlb_api *)tlb_dev->api;
+
+		/* get HPSRAM storage buffer size */
+		storage_buffer_size = tlb_api->get_storage_size();
+
+		/* add space for LPSRAM */
+		storage_buffer_size += LP_SRAM_SIZE;
+
+		/* allocate IMR buffer and store it in the global pointer */
+		global_imr_ram_storage = rmalloc(SOF_MEM_ZONE_SYS,
```
kv2019i, line 118: This is a bit funky to have Zephyr store state into address that is allocated elsewhere. But I guess as long as the L3 heap is implemented on SOF side, we have to do it like this.

src/ipc/ipc-zephyr.c
```diff
@@ -79,6 +83,62 @@ static bool message_handler(const struct device *dev, void *arg, uint32_t data,
...
+ * Checks whether device power state should be actually changed.
+ *
+ * @param dev IPC device.
+ * @param arg IPC struct pointer.
+ */
+static int ipc_device_suspend_handler(const struct device *dev, void *arg)
+{
+	struct ipc *ipc = (struct ipc *)arg;
+
+	// we are not entering D3 - return error code bad message
```
kv2019i, line 98: C-comments please.

tmleman (changes requested):

src/ipc/ipc-zephyr.c
```diff
@@ -79,6 +83,62 @@ static bool message_handler(const struct device *dev, void *arg, uint32_t data,
...
+
+	ipc_set_drvdata(ipc, NULL);
+	ipc->task_mask = 0;
+	ipc->pm_prepare_D3 = false;
+
+	// re-enable D3 power state lock
+	pm_policy_state_lock_get(PM_STATE_SOFT_OFF, PM_ALL_SUBSTATES);
+
+	/* attach handlers */
+	intel_adsp_ipc_set_message_handler(INTEL_ADSP_IPC_HOST_DEV, message_handler, ipc);
```
kv2019i, line 129: Not related to this patch, but realized this interface is not really well named, this should be used for all SOF platforms, not just Intel DSPs, so "intel_adsp_" prefix is not good here. But alas this is not added by this PR. FYI to @juimonen

src/ipc/ipc-zephyr.c
```diff
@@ -188,6 +250,12 @@ int platform_ipc_init(struct ipc *ipc)
 
 	/* attach handlers */
 	intel_adsp_ipc_set_message_handler(INTEL_ADSP_IPC_HOST_DEV, message_handler, ipc);
+#ifdef CONFIG_PM
+	intel_adsp_ipc_set_suspend_handler(INTEL_ADSP_IPC_HOST_DEV,
+					   ipc_device_suspend_handler, ipc);
+	intel_adsp_ipc_set_resume_handler(INTEL_ADSP_IPC_HOST_DEV,
+					  ipc_device_resume_handler, ipc);
```
kv2019i, line 257: I guess these are no-ops on platforms where RAM context is not maintained in D3? I.e. it's ok to still set them.

kv2019i (changes requested):
A few minor nits (see comments inline), but nothing major.

src/platform/intel/ace/platform.c
```diff
@@ -108,6 +119,12 @@ static void notify_pm_state_entry(enum pm_state state)
 						 0,
 						 SOF_MEM_CAPS_L3,
 						 storage_buffer_size);
+
+		/* IPC driver may not allow to change power state */
+		int ret = pm_device_action_run(INTEL_ADSP_IPC_HOST_DEV, PM_DEVICE_ACTION_SUSPEND);
```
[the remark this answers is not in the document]
aborisovich: So it looks like it can't "do it himself".
I've spend some time trying to figure this out and the registered in the IPC device [PM control function](https://docs.zephyrproject.org/2.6.0/reference/power_management/index.html#device-model-with-power-management-support) (here `ipc_pm_action` registered with the `PM_DEVICE_DT_DEFINE` macro) is not called "automatically" in any way.
This is also indicated in the Zephyr docs [Device Power Management API](https://docs.zephyrproject.org/2.6.0/reference/power_management/index.html#device-power-management-api).
Partial quote:

> Device Power Management API[¶](https://docs.zephyrproject.org/2.6.0/reference/power_management/index.html#device-power-management-api)
> The SOC interface and application use these APIs to perform power management operations on the devices.
(...)

> Device Set Power State[¶](https://docs.zephyrproject.org/2.6.0/reference/power_management/index.html#device-set-power-state)
> int pm_device_state_set(const struct device *dev, uint32_t device_power_state, pm_device_cb cb, void *arg);
> Calls the pm_control() handler function implemented by the device driver with PM_DEVICE_STATE_SET command.

So there is not magic here for the Device PM model. Device PM control functions are registered with PM_DEVICE_DT_DEFINE and added to the internal list, but this list is not invoked automatically in any way.
Documentation clearly states that we should call power state transition for the Device manually using `pm_device_state_set()` from the Zephyr application level. Besides, even if that list would be called automatically, there would be a question of the ordering (which pm control functions for which devices to call in what order).
I'm open for suggestions on how to do it automatically.

src/ipc/ipc-zephyr.c
aborisovich: I've verified this once again using debugger - this `pm_policy_state_lock_put(PM_STATE_SOFT_OFF, PM_ALL_SUBSTATES);`
in the `ipc_device_suspend_handler()` function works despite entering power state with `pm_state_force`.
If you comment out line where the lock is unlocked (so the IPC unlocks the lock to allow entering D3):
`// pm_policy_state_lock_put(PM_STATE_SOFT_OFF, PM_ALL_SUBSTATES);`
then, pm_policy_state_lock_is_active() returns `true` value.
In my opinion this lock system should stay as it is a validation for the IPC driver API and the Zephyr PM.
As the IPC message is the only source of signal for the whole system to enter D3 power state,
we should make sure that `ipc->task_mask` is set to `IPC_TASK_POWERDOWN` during driver suspension.

lgirdwood (commented):
LGTM, I think we need to revisist some API naming to be more generic as a subsequent PR and synchronise the west ID after dependencues are merged..

src/platform/intel/ace/platform.c
tmleman: This code will be called on other cores than primary. You need to add _**if (cpu_is_primary(arch_proc_id()))**_

src/platform/intel/ace/platform.c
tmleman, line 1: The code should compile and work when CONFIG_ADSP_IMR_CONTEXT_SAVE is disabled.

tmleman (changes requested):

src/ipc/ipc-zephyr.c
tmleman: The lock over PM_STATE_SOFT_OFF was to be permanent. It has no sense to me that we enter this state by force and remove lock during this.

src/platform/intel/ace/platform.c
aborisovich: is this ok now?

app/boards/intel_adsp_ace15_mtpm.conf
marcinszkudlinski: What do you mean? IMR context save is a process to keep the system state during D3 - when RAM power is turned off. We don't want to avoid this, but use it

west.yml
```diff
@@ -42,10 +42,8 @@ manifest:
       revision: e3a03f5ec7d8d33be705c5ce8a632d998ce9b4d1
 
     - name: zephyr
-      repo-path: zephyr
-      revision: 1851950977d5c89d8727e1058a5732f71ce4cd2a
-      remote: zephyrproject
-
+      url: https://github.com/aborisovich/zephyr
```
marcinszkudlinski, line 45: honestly, you don't want to use your zephyr fork on main 
Adding "req change" - to prevent somebody from accidently merging this

marcinszkudlinski (changes requested):

aborisovich:
Adding:
```
if (!cpu_is_primary(arch_proc_id()))
		return;
```
in `notify_pm_state_entry/exit` functions is breaking all tests.
Debugging why...

lgirdwood:
@aborisovich do the host drivers need to do anything different to enable IMR context save ?

aborisovich:
> @aborisovich do the host drivers need to do anything different to enable IMR context save ?

No, as far as I know they don't.
Please confirm @marcinszkudlinski .

src/ipc/ipc-zephyr.c
aborisovich: > The lock over PM_STATE_SOFT_OFF was to be permanent. It has no sense to me that we enter this state by force and remove lock during this.

Well, it does to me. Whatever state we force into we need to make sure that internal DSP communication is intact.

aborisovich: Fixed, removed the lock that was conceptually wrong and replaced with check whether the IPC queue is empty.

aborisovich:
Revision update - addressed @tmleman remarks:
* removed `pm_policy_state_lock_put`, `pm_policy_state_lock_get` locks
* in `intel_adsp_ipc_set_suspend_handler `added check whether ipc message queue is empty

Additionally:
* moved assertion on `irq_is_enabled` check from the `intel_adsp_ipc_set_resume_handler` to Zephyr IPC API power transition logic in the `ipc_pm_action`.

src/platform/intel/ace/platform.c
```diff
@@ -116,6 +128,12 @@ static void notify_pm_state_entry(enum pm_state state)
 						 SOF_MEM_CAPS_L3,
 						 storage_buffer_size);
 #endif /* CONFIG_ADSP_IMR_CONTEXT_SAVE */
+
+		/* change powe state and check if IPC subsystem is prepared to enter D3 */
+		int ret = pm_device_action_run(INTEL_ADSP_IPC_HOST_DEV, PM_DEVICE_ACTION_SUSPEND);
+
+		if (!ret)
+			power_state_failure_report(ret, true, PM_STATE_SOFT_OFF);
```
lyakh, line 136: maybe this is correct, but at least it *looks* suspicious - usually `ret == 0` means success, and here a failure is reported?

aborisovich: good catch

src/platform/intel/ace/platform.c
```diff
@@ -136,7 +154,17 @@ static void notify_pm_state_exit(enum pm_state state)
...
+
+		if (!ret)
+			power_state_failure_report(ret, false, PM_STATE_SOFT_OFF);
+
+		enum pm_device_state ipc_state = PM_DEVICE_STATE_SUSPENDING;
+
+		pm_device_state_get(INTEL_ADSP_IPC_HOST_DEV, &ipc_state);
+		__ASSERT_NO_MSG(ipc_state == PM_DEVICE_STATE_ACTIVE);
+
+		/* sends fw-ready message signalling successful exit from D3 state */
```
tmleman: This should also be under _**#ifdef**_

```c
#ifdef CONFIG_ADSP_IMR_CONTEXT_SAVE
		/*your code*/
#endif
```

tmleman (changes requested):

aborisovich:
Regarding checkpatch there is 1 warning...
Basically, it is very hard to co-develop anything with @tmleman 🤣 
image

aborisovich:
Revision update:
* put under #ifdef CONFIG_IMR_CONTEXT_SAVE what should be there
* fixed error handling `if (ret)`

aborisovich:
> @aborisovich wrote:
> 
> > Regarding checkpatch there is 1 warning... Basically, it is very hard to co-develop anything with @tmleman 🤣 image
> 
> The guidance is that you need to have a signed-off for the patch author and for every co-developed-by person as well. I.e. you need to have signed-off by @tmleman to make this correct.

I've added Tomasz Leman as Signed-off-by also, however this does not fix the obvious problem:

> WARNING: Co-developed-by and Signed-off-by: name/email do not match

paulstelian97:
I wonder if you need a second Co-developed-by line for yourself, so that it could match _that_ line?

kv2019i:
@paulstelian97 @aborisovich Checkpatch is not that clever. You need to use following form to keep checkpatch happy. This is pretty quick to check lcoally:

OK
``` 
    foo: bar
    
    Co-developed-by: Foo Bar <foo.bar@foo.com>
    Signed-off-by: Foo Bar <foo.bar@foo.com>
```

That makes checkpatch happy.

NOT OK - This is correct per policy but checkpatch still complains:
```
    foo: bar
    
    Co-developed-by: Foo Bar <foo.bar@foo.com>
    Signed-off-by: Foo Bar <foo.bar@foo.com>
```

lyakh:
> Regarding checkpatch there is 1 warning... Basically, it is very hard to co-develop anything with @tmleman rofl 

maybe you just need both SOBs

src/platform/intel/ace/platform.c
```diff
@@ -74,6 +78,74 @@ int platform_boot_complete(uint32_t boot_message)
...
 }
 
+/* address where zephyr PM will save memory during D3 transition */
+#ifdef CONFIG_ADSP_IMR_CONTEXT_SAVE
+extern void *global_imr_ram_storage;
+#endif
+
+/**
+ * @brief Notifier called before every power state transition.
+ * Works on Primary Core only.
```
tmleman, line 100: This code is not doing anything for secondary cores but it will be executed on them. This comment need to be changed or removed.

src/platform/intel/ace/platform.c
```diff
@@ -74,6 +78,74 @@ int platform_boot_complete(uint32_t boot_message)
...
+						 0,
+						 SOF_MEM_CAPS_L3,
+						 storage_buffer_size);
+#endif /* CONFIG_ADSP_IMR_CONTEXT_SAVE */
+	}
+}
+
+/**
+ * @brief Notifier called after every power state transition.
+ * Works on Primary Core only.
```
tmleman, line 143: dito

tmleman (changes requested):

west.yml
marc-hb: This is obviously for testing. I hope no maintainer would merge this but if you want to be extra safe making this a "draft" is better than [DNM]. Drafts have no "merge" button at all.

src/platform/intel/ace/platform.c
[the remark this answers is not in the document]
aborisovich: Fixed, there was a bug in Zephyr PM, now it can do it by himself.

src/platform/intel/ace/platform.c
aborisovich: It does now.

src/ipc/ipc-zephyr.c
```diff
@@ -79,6 +83,51 @@ static bool message_handler(const struct device *dev, void *arg, uint32_t data,
...
+ * @brief IPC device suspend handler callback function.
+ * Checks whether device power state should be actually changed.
+ *
+ * @param dev IPC device.
+ * @param arg IPC struct pointer.
+ */
+static int ipc_device_suspend_handler(const struct device *dev, void *arg)
+{
+	struct ipc *ipc = (struct ipc *)arg;
+
```
tmleman, line 97: Why not just call ipc_get()?

tmleman (changes requested):
If suspending the IPC device fails, we will return to the idle state (PM_STATE_ACTIVE). The host will not receive any message from the DSP about what happened. We need to send a failure message and add logs of the problems encountered.

aborisovich:
[Full scope tests](http://popychaczka-ace.igk.intel.com/history/85346#diff) on the existing state of PRs (this and Zephyr part) proved regression in :

- multiple versions of test_01_08_gain_quality
- multiple versions of test_01_09_peakvol_quality
- multiple versions of aria driver
- i2s_with_simultaneous_hda [16000Hz_16in16bit_2ch_16000Hz_16in16bit_2ch-i2s_internal]

Failures seem to be unrelated to this PR, rebasing the the newest SOF main and restarting.

tmleman:
@aborisovich  SOF CI looks good https://github.com/thesofproject/sof/pull/7325. How did our internal full scope tests go?

aborisovich:
> Full scope tests on the existing state of PRs (this and Zephyr part) proved regression in :
> 
> * multiple versions of test_gain_quality
> * multiple versions of test_peakvol_quality
> * multiple versions of aria driver
> * i2s_with_simultaneous_hda
> 
> Failures seem to be unrelated to this PR, rebasing the the newest SOF main and restarting.

New full scope test results fix those tests mentioned above (rebase to newest SOF main helped).
Currently the only regression we have is single stress test with graceful shutdown.

src/ipc/ipc-zephyr.c
aborisovich: Well, this IPC Zephyr API was supposed to be generic to be used not only by Intel.
Thus we left some freedom for data passing in handler args.
Same is done in the IPC message handler. So if we pass anyway the IPC struct as function `arg`, why calling `ipc_get()`?

aborisovich:
@marc-hb what's going on with the `sof-ci/jenkins/build`? It tells me `unknown error`...

marc-hb:
We had a jenkins outage today, apologies. Please resubmit tomorrow.

marc-hb:
Different topic: I noticed the first commit uprevs Zephyr in sof/west.yml. There are two possibilities:
- Either this uprev commit can be submitted separately. Then it should be submitted separately first in a different PR merged first so we get at least one day of separate validation for _Continuous_ Integration.
- Or this Zephyr uprev _cannot_ be submitted separately because of some non-backwards compatible change in Zephyr or some other reason. Then it should _not_ be a separate commit not to break `git -C sof bisect`.

src/ipc/ipc-zephyr.c
tmleman: The `generic  use` argument is enough for me.

Just passing it around seems weird when we have access to it locally. It looked as if it was done that way so that the `arg` parameter had some use.

tmleman (changes requested):

west.yml
```diff
@@ -45,9 +45,8 @@ manifest:
 
     - name: zephyr
       repo-path: zephyr
-      revision: e2e3dc0771188f2a01e3aaa792f81e6c6a611eb6
+      revision: 2bf335864cac67c6001d8f7c7ad7e903b20ee080
```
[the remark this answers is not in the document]
aborisovich: I'll move this change to separate PR. As @marc-hb indicated, we are also dealing with signature change of the `intel_adsp_ipc_send_message()` function witch is used without PM too obviously. It would work better in terms of integration and bisecting.

aborisovich:
Dropped Zephyr revision update from this PR, created a new one that just updates function reference: https://github.com/thesofproject/sof/pull/7857

aborisovich:
Push update:

- Added new commit to handle IPC power transition failures and send proper response to the Host

aborisovich:
Fixing checkpatch errors one by one...
This one got my attention (an annoying one):
image
I wonder what was author thinking when was creating this rule...
@marc I suggest removing it since an explicit comparison between bolean and the variable is actually **LESS** error prone
than implicit cast of the variable to true or false inside if statement braces.
That is also why, in more developed languages like C++, types of variables that are implicitly castable to bool were reduced.

aborisovich:
I do not know how to address this checkpatch issue:
image
When this is not split across the lines, checkpatch complains that I've exceeded the 100 characters in width...

marc-hb:
> Ok, maybe not in this case particularly, because the pm_prepare_D3 was defined (why?) with type int.

So this warning seemed valid because you were comparing an `int` with a `bool`.

I have no idea why, maybe someone thought this could have more than 2 states in the future? If that makes no sense then feel free to convert it to a boolean.

> That is also why, in more developed languages like C++, types of variables that are implicitly castable to bool were reduced.

Looks checkpatch and C++ agree!

I see the current version has `if (!ipc->pm_prepare_D3)` instead which checkpatch seems happy with.

> > quoted string split across lines

> When this is not split across the lines, checkpatch complains that I've exceeded the 100 characters in width...

Yes this type of catch-22 tend to happen. I don't know a way out; split it and ignore the warning.

marc-hb:
> I wonder what was author thinking when was creating this rule...

Enjoy the 2013 discussion: https://lore.kernel.org/lkml/1365563834.27174.12.camel@joe-AO722/#r

lyakh:
> I do not know how to address this checkpatch issue: image When this is not split across the lines, checkpatch complains that I've exceeded the 100 characters in width... Moreover, I've seen similar usages of `tr_err` in code done in the similar way.

long strings in log prints are tolerated and preferred over split strings. The reasoning is that split strings are difficult to grep for after seeing them in a log. But would be good to at least take that string on the next line - splitting after `&ipc_tr,`

lyakh:
> I wonder what was author thinking when was creating this rule...

this comes from the Linux kernel coding style, which SOF largely follows. E.g. in the kernel it's preferred to write `if (addr)` over `if (addr != NULL)` and when you explicitly need a boolean value you'd rather write like `bool flag = !!calc();` instead of `bool flag = calc() != 0;`

tmleman (approved):

aborisovich:
> > I do not know how to address this checkpatch issue: image When this is not split across the lines, checkpatch complains that I've exceeded the 100 characters in width... Moreover, I've seen similar usages of `tr_err` in code done in the similar way.
> 
> long strings in log prints are tolerated and preferred over split strings. The reasoning is that split strings are difficult to grep for after seeing them in a log. But would be good to at least take that string on the next line - splitting after `&ipc_tr,`

Resolved issue with the too long log by removing function name prefix. Turned out Zephyr logging provides this name prefix by default so it is not needed, thanks @tmleman .

aborisovich:
> > I wonder what was author thinking when was creating this rule...
> 
> Enjoy the 2013 discussion: https://lore.kernel.org/lkml/1365563834.27174.12.camel@joe-AO722/#r

It had been proved over and over again on many technical and non-technical subjects, that Linux Kernel project leadership is not an entity listening to anyone's reasoning so, no thank you. I will restrain myself from providing references with links because it can be easily found in the internet.
The question is why SOF project follows something so helpless as Linux Kernel. I do not think this is a requirement but the choice we made. And now the project undergoes convergence with Windows and Chrome OS, so it is far from linux kernel practices.

pjdobrowolski (approved):

zephyr/lib/cpu.c
```diff
@@ -64,6 +68,42 @@ LOG_MODULE_DECLARE(zephyr, CONFIG_SOF_LOG_LEVEL);
...
+
+		__ASSERT_NO_MSG(tlb_dev);
+		const struct intel_adsp_tlb_api *tlb_api =
+				(struct intel_adsp_tlb_api *)tlb_dev->api;
+
+		/* get HPSRAM storage buffer size */
+		storage_buffer_size = tlb_api->get_storage_size();
+
+		/* add space for LPSRAM */
+		storage_buffer_size += LP_SRAM_SIZE;
```
kv2019i, line 96: Minor: it's not so obvious this is coming from Zephyr adsp_memory.h . Might be more readable to just directly use "DT_REG_SIZE(DT_NODELABEL(sram1)" as this a Zephyr-only file.

zephyr/lib/cpu.c
```diff
@@ -64,6 +68,42 @@ LOG_MODULE_DECLARE(zephyr, CONFIG_SOF_LOG_LEVEL);
...
+
+		/* allocate IMR buffer and store it in the global pointer */
+		global_imr_ram_storage = rmalloc(SOF_MEM_ZONE_SYS,
```
kv2019i, line 99: By alloc.h  "* 1) System Zone. Fixed size heap where alloc always succeeds and is never freed. Used by any init code that will never give up the memory." 
This memory is freed at restore so SYS_RUNTIME would be more appropriate zone. Functionally this should not matter as SOS_MEM_CAPS_L3 is passed so actual allocation is done from IMR.

src/ipc/ipc-zephyr.c
```diff
@@ -17,11 +17,15 @@
 #include <sof/ipc/schedule.h>
```
kv2019i, line 17: typo in commit: Zepyr IPC driver

src/ipc/ipc4/handler.c
```diff
@@ -1068,6 +1068,25 @@ void ipc_boot_complete_msg(struct ipc_cmd_hdr *header, uint32_t data)
 	header->ext = 0;
 }
 
+#if defined(CONFIG_PM_DEVICE) && defined(CONFIG_INTEL_ADSP_IPC)
+void ipc_send_failed_power_transition_response(void)
+{
+	struct ipc4_message_request *request = ipc_from_hdr(&msg_data.msg_in);
+
```
kv2019i, line 1075: minor: L1075 undeeded empty line

kv2019i (approved):
Some comments but nothing major, so I'll put my approval on here. Let's see how the tests look like.

plbossart:
> > > I wonder what was author thinking when was creating this rule...
> > 
> > 
> > Enjoy the 2013 discussion: https://lore.kernel.org/lkml/1365563834.27174.12.camel@joe-AO722/#r
> 
> It had been proved over and over again on many technical and non-technical subjects, that Linux Kernel project leadership is not an entity listening to anyone's reasoning so, no thank you. I will restrain myself from providing references with links because it can be easily found in the internet. The question is why SOF project follows Linux Kernel. I do not think this is a requirement but the choice we made. And now the project undergoes convergence with Windows and Chrome OS.

The Linux kernel has been a rather successful project and likely the most successful open-source project. Please keep your rants to yourself, it's not helpful really.

aborisovich:
Push update:

- addressed @kv2019i review - fixed whitespaces, typos and changed `global_imr_ram_storage` pointer to be allocated from the `SYS_RUNTIME` memory pool.

aborisovich:
> > > > I wonder what was author thinking when was creating this rule...
> > > 
> > > 
> > > Enjoy the 2013 discussion: https://lore.kernel.org/lkml/1365563834.27174.12.camel@joe-AO722/#r
> > 
> > 
> > It had been proved over and over again on many technical and non-technical subjects, that Linux Kernel project leadership is not an entity listening to anyone's reasoning so, no thank you. I will restrain myself from providing references with links because it can be easily found in the internet. The question is why SOF project follows Linux Kernel. I do not think this is a requirement but the choice we made. And now the project undergoes convergence with Windows and Chrome OS.
> 
> The Linux kernel has been a rather successful project and likely the most successful open-source project. Please keep your rants to yourself, it's not helpful really.

My rants are my rants. My argumentation is my argumentation. The former is not needed, the latter is valid.

aborisovich:
Is the test failure `check-suspend-resume-with-capture-5.sh` on MTL nocodec topology known issue?

zephyr/lib/cpu.c
aborisovich: Well, I would leave this matter to the author @marcinszkudlinski since it is his commit.
He is currently on vacation. Can we address this later? I will let him know.

kv2019i:
@aborisovich wrote:
> Is the test failure `check-suspend-resume-with-capture-5.sh` on MTL nocodec topology known issue?

Yes, it's a known sporadic fail we see. So https://sof-ci.01.org/sofpr/PR7325/build10138/devicetest/index.html looks good. Of course given your PR changes the D3 flow, we'd ideally have a reliable pass as the baseline to compare against, but alas we don't have that here. Let's see the "System/merge/build" results.

softwarecki (approved):

kv2019i:
Known fail in https://sof-ci.01.org/sofpr/PR7325/build10139/devicetest/index.html . https://sof-ci.01.org/sofpr/PR7325/build10138/devicetest/index.html actually looks better than other PRs today (no hits of https://github.com/thesofproject/sof/issues/7482 ), so this might actually improve the situation. Proceeding with merge.

marc-hb:
Suspected regression
- #7969

PR #7328, opened 2023-03-24
src: src should not use buffers when using module API

marcinszkudlinski:
Src can use module API, but is still getting data from comp_buffer 
What more dangerous, it is calling buffer_acquire on source and sink.
Both buffers have already been taken by module adapter, so it may lead to deadlocks.

### DIFF
--- a/src/audio/src/src.c
+++ b/src/audio/src/src.c
@@ -628,8 +628,8 @@ static int src_verify_params(struct comp_dev *dev, struct comp_data *cd,
 }
 
 static int src_get_copy_limits(struct comp_data *cd,
-			       const struct comp_buffer __sparse_cache *source,
-			       const struct comp_buffer __sparse_cache *sink)
+			       const struct audio_stream __sparse_cache *source,
+			       const struct audio_stream __sparse_cache *sink)
 {
 	struct src_param *sp;
 	struct src_stage *s1;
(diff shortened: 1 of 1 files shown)


marcinszkudlinski:
similar problem (buffer_aquire when using moduel API) is probably in mixin_mixout.c and mux.c

all Buffers are already aquired at this point in module adapter:

	/* acquire all sink and source buffers */
	list_for_item(blist, &dev->bsink_list) {
		struct comp_buffer *sink;

		sink = container_of(blist, struct comp_buffer, source_list);
		sinks_c[i++] = buffer_acquire(sink);
	}
	i = 0;
		struct comp_buffer *source;

		source = container_of(blist, struct comp_buffer, sink_list);
		source_c[i++] = buffer_acquire(source);
	}

paulstelian97 (approved):
This needs to be checked for all components. Approved for src fix.

lgirdwood (approved):
Good find @marcinszkudlinski !
@singalsu pls review.

singalsu (approved):
Good finding, thanks!

abonislawski (approved):

softwarecki (approved):

marc-hb:
This commit introduced a `sparse` regression that is affecting every PR now, example:
https://github.com/thesofproject/sof/actions/runs/4746254605/jobs/8429649357?pr=7473
```
sof/src/audio/src/src.c:977:21: warning: cast removes address space '__cache' of expression
sof/src/audio/src/src.c:977:18: warning: incorrect type in assignment (different address spaces)
sof/src/audio/src/src.c:978:19: warning: cast removes address space '__cache' of expression
sof/src/audio/src/src.c:978:16: warning: incorrect type in assignment (different address spaces)
```

```C
       source_c = (struct audio_stream *)input_buffers[0].data;
       sink_c = (struct audio_stream *)output_buffers[0].data;
```

We didn't catch it at the time and are seeing it only now because we had a "sparse saturation" problem that @lyakh just started fixing with much more recent commit f772b39e4d5de2e1e

PR #7336, opened 2023-03-27
set_attenuation: adding checker based on config data

fkwasowi:
Adding attenuation checking based on the frame_fmt from the config of the module instead of calling
sink->stream.frame_fmt, because sending IPC_SET_ATTENUATION before calling module_prepare results in an error, because at this stage the sink values are not yet set (sink->stream.frame_fmt is zero)

### DIFF
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -1576,9 +1576,8 @@ static int copier_set_sink_fmt(struct comp_dev *dev, const void *data,
 static int set_attenuation(struct comp_dev *dev, uint32_t data_offset, const char *data)
 {
 	struct copier_data *cd = comp_get_drvdata(dev);
-	struct comp_buffer *sink;
-	struct list_item *sink_list;
 	uint32_t attenuation;
+	uint32_t __sparse_cache valid_fmt, frame_fmt;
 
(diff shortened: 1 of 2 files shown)


pblaszko (approved):

dnikodem (approved):

dabekjakub (approved):

src/audio/copier/copier.c
```diff
@@ -1591,14 +1589,18 @@ static int set_attenuation(struct comp_dev *dev, uint32_t data_offset, const cha
 		comp_err(dev, "attenuation %d is out of range", attenuation);
 		return -EINVAL;
 	}
+	uint32_t __sparse_cache valid_fmt, frame_fmt;
```
RanderWang, line 1592: can we move the declaration to the beginning of this function ?

RanderWang (approved):
one concern

src/audio/copier/copier.c
fkwasowi: SOF CI

fkwasowi:
SOF CI

src/audio/copier/copier.c
fkwasowi: Done

PR #7341, opened 2023-03-28
idc: call cache function z_xtensa_cache_flush directly

nashif:
Do not use intermediate macro SOC_DCACHE_FLUSH from Zephyr, use cache
function instead.
SOC_DCACHE_FLUSH will be removed from zephyr.

Signed-off-by: Anas Nashif <anas.nashif@intel.com>

### DIFF
--- a/src/idc/zephyr_idc.c
+++ b/src/idc/zephyr_idc.c
@@ -29,6 +29,9 @@
 #include <rtos/alloc.h>
 #include <rtos/spinlock.h>
 #include <ipc/topology.h>
+#if defined(CONFIG_ARCH_XTENSA) && !defined(CONFIG_KERNEL_COHERENCE)
+#include <zephyr/arch/xtensa/cache.h>
+#endif
 
 /*
  * Inter-CPU communication is only used in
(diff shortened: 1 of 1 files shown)


src/idc/zephyr_idc.c
```diff
@@ -29,6 +29,9 @@
 #include <rtos/alloc.h>
 #include <rtos/spinlock.h>
 #include <ipc/topology.h>
+#if defined(CONFIG_ARCH_XTENSA) && !defined(CONFIG_KERNEL_COHERENCE)
+#include <zephyr/arch/xtensa/cache.h>
+#endif
```
kv2019i, line 34: Hmm, if these are not defined, wouldn't the build anyways fail when z_xtensa_cache_flush() is not defined?

kv2019i (changes requested):
One question on include guards, see inline.

sys-pt1s:

src/idc/zephyr_idc.c
nashif: do not remember why I did this, afaik this is the same in `zephyr/lib/alloc.c`

lgirdwood: @kv2019i do we need to fix ```zephyr/lib/alloc.c``` or is this correct ?

kv2019i:
Code migrated to use sys_cache_, can close this one.

PR #7342, opened 2023-03-29
dai-zephyr: use direct function calling to replace callback

btian1:
With callback register/unregister, make code not so easy to read. Also, there is no direct benefit for callback usage, direct call also works well.

### DIFF
--- a/src/audio/dai-zephyr.c
+++ b/src/audio/dai-zephyr.c
@@ -88,10 +88,6 @@ int dai_assign_group(struct comp_dev *dev, uint32_t group_id)
 	comp_dbg(dev, "dai_assign_group(), group %d num %d",
 		 group_id, dd->group->num_dais);
 
-	/* Register for the atomic trigger event */
-	notifier_register(dev, dd->group, NOTIFIER_ID_DAI_TRIGGER,
-			  dai_atomic_trigger, 0);
-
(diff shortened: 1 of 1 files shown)


src/audio/dai-zephyr.c
```diff
@@ -88,10 +88,6 @@ int dai_assign_group(struct comp_dev *dev, uint32_t group_id)
 	comp_dbg(dev, "dai_assign_group(), group %d num %d",
 		 group_id, dd->group->num_dais);
 
-	/* Register for the atomic trigger event */
```
ranj063, line 91: @btian1 im not sure about this change. THis is meant for triggering aggregated DAIs together. How are you testing aggregation to say this works fine?

btian1: I can't 100% sure, it is in my road for dai_config, so want to have a try to replace, first tried with local playback and capture, it works fine, then check CI, seems also good.

Regarding Aggregated trigger, if you confirmed this PR is not applicable, could you give more comments here? I don't know this part.

plbossart: this PR doesn't look good to me at all. Either we remove the notion of dai group or we keep it as is. Changing parts of the solution without context or test results makes no sense to me.

PR #7344, opened 2023-03-29
peakvol: prevent division by zero

fkwasowi:
When calculating the time_ratio, a situation may occur where cd->initial_ramp is equal to zero. To avoid dividing by zero it is necessary to insert an additional condition

### DIFF
--- a/src/audio/module_adapter/module/volume/volume.c
+++ b/src/audio/module_adapter/module/volume/volume.c
@@ -252,6 +252,9 @@ static inline int32_t volume_windows_fade_ramp(struct vol_data *cd, int32_t ramp
 	int32_t pow_value; /* Q2.30 */
 	int32_t volume_delta = cd->tvolume[channel] - cd->rvolume[channel]; /* Q16.16 */
 
+	if (!cd->initial_ramp)
+		return cd->tvolume[channel];
+
 	time_ratio = (((int64_t)ramp_time) << 30) / (cd->initial_ramp << 3);
 	pow_value = volume_pow_175(time_ratio);
 	return cd->rvolume[channel] + Q_MULTSR_32X32((int64_t)volume_delta, pow_value, 16, 30, 16);


src/audio/module_adapter/module/volume/volume.c
```diff
@@ -252,7 +252,10 @@ static inline int32_t volume_windows_fade_ramp(struct vol_data *cd, int32_t ramp
 	int32_t pow_value; /* Q2.30 */
 	int32_t volume_delta = cd->tvolume[channel] - cd->rvolume[channel]; /* Q16.16 */
 
-	time_ratio = (((int64_t)ramp_time) << 30) / (cd->initial_ramp << 3);
+	if (cd->initial_ramp != 0)
+		time_ratio = (((int64_t)ramp_time) << 30) / (cd->initial_ramp << 3);
+	else
+		time_ratio = (((int64_t)ramp_time) << 30);
```
pblaszko, line 258: in case driver configures ramp time to 0, it does not make sense to enter these functions at all. its hard to predict what will be the result of continuing these volume calculation with non-natural time_ratio.
I think cd->initial_ramp should be checked to be non-zero in volume_process(), otherwise do not enter volume_ramp()

pblaszko (changes requested):

src/audio/module_adapter/module/volume/volume.c
pblaszko: or maybe even better solution: in init_ramp(): if (!cd->initial_ramp): cd->ramp_finished = true
Then volume_process will not enter ramp function

fkwasowi: ,,or maybe even better solution: in init_ramp(): if (!cd->initial_ramp): cd->ramp_finished = true
Then volume_process will not enter ramp function" - After this correction, we enter to the ramp function and division by zero continues to occur.

in case driver configures ramp time to 0, it does not make sense to enter these functions at all. its hard to predict what will be the result of continuing these volume calculation with non-natural time_ratio.
I think cd->initial_ramp should be checked to be non-zero in volume_process(), otherwise do not enter volume_ramp()- signal verification does not pass after this amendment.

fkwasowi: Done

tmleman (approved):

pblaszko (approved):

dnikodem (approved):

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -252,6 +252,9 @@ static inline int32_t volume_windows_fade_ramp(struct vol_data *cd, int32_t ramp
 	int32_t pow_value; /* Q2.30 */
 	int32_t volume_delta = cd->tvolume[channel] - cd->rvolume[channel]; /* Q16.16 */
 
+	if (!cd->initial_ramp)
```
lgirdwood, line 255: Btw, does it makse sense to always initialse this to a sane non 0 value during new() and reset() so that value is always good and a ramp is always perfomed i.e. no audible artifacts ? 
@singalsu pls comment here.

pblaszko: at least for IPC4, this value always comes in module init_instance IPC (curve_duration), so initializing it with non-zero value does not change anything, it is overwritten before use anyway.

lgirdwood: ok, so this means we need are not validating this value from the IPC ? i.e. host is sending an illegal value ?

fkwasowi:
SOFCI TEST

src/audio/module_adapter/module/volume/volume.c
pblaszko: It is not illegal, it just means "change volume immediately (time=0)"

lgirdwood: ok, that's fine.

lgirdwood (approved):

src/audio/module_adapter/module/volume/volume.c
btian1: another way is:
cd->vol_min and cd->vol_max, if same also means set to target volume immediately.
return cd->vol_min may also works.

PR #7346, opened 2023-03-29
topology1: set lower priority for DMIC task in GOOGLE_RTC_AUDIO case

johnylin76:
When GOOGLE_RTC_AUDIO is applied, Speaker and DMIC pipeline tasks are scheduled in the same domain. The AEC function in GOOGLE_RTC_AUDIO has the hard requirement for input sample synchronization between DMIC input and Echo reference (sourced from Speaker).

At present, both pipelines are pri-0 so the task order in schedule list depends on which pipeline starts first. DMIC pipeline task needs the most computation cycles due to AEC and NR processings, even more than the entire schedule period in peak. If DMIC is scheduled former than Speaker, the synchronization may be broken by the peak of DMIC task which punts Speaker task to the next period.

To avoid the above case, this commit sets DMIC as pri-1 on GOOGLE_RTC_AUDIO applied. It guarantees that DMIC task will be always put after Speaker task.

### DIFF
--- a/tools/topology/topology1/platform/intel/intel-generic-dmic.m4
+++ b/tools/topology/topology1/platform/intel/intel-generic-dmic.m4
@@ -10,6 +10,9 @@ ifdef(`DMIC_48k_PCM_NAME',`',
 ifdef(`DMIC_16k_PCM_NAME',`',
 `define(DMIC_16k_PCM_NAME, `DMIC16kHz')')
 
+ifdef(`DMIC_48k_PRIORITY', `',
+`define(DMIC_48k_PRIORITY, 0)')
+
 # variable that need to be defined in upper m4
 ifdef(`DMICPROC',`',`fatal_error(note: Need to define dmic processing for intel-generic-dmic
 )')
@@ -127,7 +130,7 @@ dnl     deadline, priority, core, time_domain)
 DAI_ADD(sof/pipe-dai-capture.m4,
 	DMIC_PIPELINE_48k_ID, DMIC, 0, DMIC_DAI_LINK_48k_NAME,
 	concat(`PIPELINE_SINK_', DMIC_PIPELINE_48k_ID), 2, s32le,
-	DMIC_48k_PERIOD_US, 0, DMIC_48k_CORE_ID, SCHEDULE_TIME_DOMAIN_TIMER)
+	DMIC_48k_PERIOD_US, DMIC_48k_PRIORITY, DMIC_48k_CORE_ID, SCHEDULE_TIME_DOMAIN_TIMER)
 
 # capture DAI is DMIC 1 using 2 periods
 # Buffers use s32le format, with 16 frame per 1000us on core 0 with priority 0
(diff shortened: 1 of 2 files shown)


kv2019i (approved):
Thanks! Once reviewed and merged, needs to be backported to stable-v2.2

singalsu (approved):

lkoenig (approved):

plbossart:
> If DMIC is scheduled former than Speaker, the synchronization may be broken by the peak of DMIC task which punts Speaker task to the next period.

I don't get what changing the priority would fix. If you deal with the speaker first, then it's the DMIC that will not complete its task before the next firmware tick.

When the total cycles required exceed what's available, no amount of priority changes will solve the problem that the DSP is oversubscribed.

What am I missing @johnylin76 ?

lgirdwood:
> > If DMIC is scheduled former than Speaker, the synchronization may be broken by the peak of DMIC task which punts Speaker task to the next period.
> 
> I don't get what changing the priority would fix. If you deal with the speaker first, then it's the DMIC that will not complete its task before the next firmware tick.
> 
> When the total cycles required exceed what's available, no amount of priority changes will solve the problem that the DSP is oversubscribed.
> 
> What am I missing @johnylin76 ?

Agree, I don't think we are seeing over budget fix here (as not doable without doing less work), but more a task reordeing only fix to ensure a better data flow (which may also improve cache performance). LGTM.

plbossart:
> Agree, I don't think we are seeing over budget fix here (as not doable without doing less work), but more a task reordeing only fix to ensure a better data flow (which may also improve cache performance). LGTM.

The main issue I have is that xruns on the playback path will be easy to detect - there will be an audible glitch over multiple ms. On capture it's less obvious to detect, it could very well be that we end-up with missing or overwritten data, who knows.

cujomalainey:
> > Agree, I don't think we are seeing over budget fix here (as not doable without doing less work), but more a task reordeing only fix to ensure a better data flow (which may also improve cache performance). LGTM.
> 
> The main issue I have is that xruns on the playback path will be easy to detect - there will be an audible glitch over multiple ms. On capture it's less obvious to detect, it could very well be that we end-up with missing or overwritten data, who knows.

We are working to fix echo issues on our AEC, not glitch problems. The AEC can sometimes peak over budget but on average is below. Making sure speaker runs first makes sure we always get the reference in time to avoid leaking.

plbossart:
> We are working to fix echo issues on our AEC, not glitch problems. The AEC can sometimes peak over budget but on average is below. Making sure speaker runs first makes sure we always get the reference in time to avoid leaking.

It's even less clear to me @cujomalainey. You get the echo reference from the SSP capture stream, don't you? How would running the speaker task first make the reference available earlier - if it's in another pipeline anyways?

cujomalainey:
@plbossart reference comes from the speaker path directly. Capture pulls raw dmic. Making sure speaker is first makes sure we don't xrun on the demux cross link of the pipes.

cujomalainey (approved):

cujomalainey:
clearing my approval as I have some new questions on our internal analysis

plbossart:
> @plbossart reference comes from the speaker path directly. Capture pulls raw dmic. Making sure speaker is first makes sure we don't xrun on the demux cross link of the pipes.

humm ok, if the echo reference is generated with a demux, then it's not really a question of priority but rather of doing things in the right order with dependencies handled. Playing with priorities seem to hide a graph walk issue IMHO.

johnylin76:
Thanks all for the review. Yep after digging the original issue deeper I think the reason why this fix works is due to the dependency between DMIC and Speaker pipe tasks, rather than the over-budget.

I have revised the commit message https://github.com/thesofproject/sof/pull/7346/commits/d1eb23a16aaeb625732aa9c52f2e1eaa7131b610

sys-pt1s:

lgirdwood (approved):

PR #7348, opened 2023-03-29
topology2: Mod cfg ibs obs cpc addition

jsarha:
I am feeling quite clueless here, so even if this does not work, I push this out in hope someone can spot the problem.

For some reason the widget tuple array only contains the first instance of mod_cfg objects in the topology2 conf file. I am trying to read the array out with this code: https://github.com/jsarha/linux-sof/tree/peter-cpc-test

### DIFF
--- a/tools/topology/topology2/include/common/mod_cfg.conf
+++ b/tools/topology/topology2/include/common/mod_cfg.conf
@@ -1,0 +1,45 @@
+# Class definition for module configuration object
+# mod_cfg objects can be instantiated as:
+#
+# Object.Base.mod_cfg."0" {
+#	ibs	384
+#	obs	384
+#	cpc	4600
+# }
+#
+
+Class.Base."mod_cfg" {
+
+    DefineAttribute."instance" {
+    }
+
+    # input buffer size
+    DefineAttribute."ibs" {
+	# Token set reference name
+	token_ref    "sof_tkn_mod_cfg.word"
+    }
+
+    # output buffer size
+    DefineAttribute."obs" {
+	# Token set reference name
+	token_ref    "sof_tkn_mod_cfg.word"
+    }
+
+    # Cycles Per Chunk
+    DefineAttribute."cpc" {
+	# Token set reference name
+	token_ref    "sof_tkn_mod_cfg.word"
+    }
+
+    attributes {
+	!constructor [
+	    "instance"
+	]
+
+	#
+	# mod_cfg objects instantiated within the same alsaconf node
+	# must have unique instance attribute values
+	#
+	unique	  "instance"
+    }
+}
--- a/tools/topology/topology2/include/common/tokens.conf
+++ b/tools/topology/topology2/include/common/tokens.conf
@@ -24,6 +24,7 @@ Object.Base.VendorToken {
 		output_pin_binding_name	414
 		num_input_audio_formats	415
 		num_output_audio_formats 416
+		num_mod_cfgs		417
 	}
 
 	"2" {
@@ -178,5 +179,13 @@ Object.Base.VendorToken {
 		name "sof_tkn_copier"
 		node_type		1980
 		deep_buffer_dma_ms	1981
+	}
+
+	# ABI 4.0 onwards
+	"19" {
+		name "sof_tkn_mod_cfg"
+		ibs			1990
+		obs			1991
+		cpc			1992
 	}
 }
--- a/tools/topology/topology2/include/components/widget-common.conf
+++ b/tools/topology/topology2/include/components/widget-common.conf
@@ -109,3 +109,9 @@ DefineAttribute."num_output_audio_formats" {
 	# Token set reference name and type
 	token_ref	"sof_tkn_comp.word"
 }
+
+# Number defined module configurations
+DefineAttribute."num_mod_cfgs" {
+	# Token set reference name and type
+	token_ref	"sof_tkn_comp.word"
+}
--- a/tools/topology/topology2/include/pipelines/cavs/dai-copier-be.conf
+++ b/tools/topology/topology2/include/pipelines/cavs/dai-copier-be.conf
@@ -17,6 +17,7 @@
 #
 
 <include/common/audio_format.conf>
+<include/common/mod_cfg.conf>
 <include/components/copier.conf>
 <include/components/pipeline.conf>
 
--- a/tools/topology/topology2/include/pipelines/cavs/dai-copier-eqiir-module-copier-capture.conf
+++ b/tools/topology/topology2/include/pipelines/cavs/dai-copier-eqiir-module-copier-capture.conf
@@ -17,6 +17,7 @@
 #
 
 <include/common/audio_format.conf>
+<include/common/mod_cfg.conf>
 <include/components/copier.conf>
 <include/controls/bytes.conf>
 <include/components/eqiir.conf>
--- a/tools/topology/topology2/include/pipelines/cavs/dai-copier-gain-mixin-capture.conf
+++ b/tools/topology/topology2/include/pipelines/cavs/dai-copier-gain-mixin-capture.conf
@@ -17,6 +17,7 @@
 #
 
 <include/common/audio_format.conf>
+<include/common/mod_cfg.conf>
 <include/components/copier.conf>
 <include/components/gain.conf>
 <include/components/mixin.conf>
--- a/tools/topology/topology2/include/pipelines/cavs/dai-copier-gain-module-copier-capture.conf
+++ b/tools/topology/topology2/include/pipelines/cavs/dai-copier-gain-module-copier-capture.conf
@@ -17,6 +17,7 @@
 #
 
 <include/common/audio_format.conf>
+<include/common/mod_cfg.conf>
 <include/components/copier.conf>
 <include/components/gain.conf>
 <include/components/pipeline.conf>
--- a/tools/topology/topology2/include/pipelines/cavs/dai-kpb-be.conf
+++ b/tools/topology/topology2/include/pipelines/cavs/dai-kpb-be.conf
@@ -18,6 +18,7 @@
 #
 
 <include/common/audio_format.conf>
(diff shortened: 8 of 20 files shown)


tools/topology/topology2/cavs-mixin-mixout-hda.conf
```diff
@@ -36,11 +36,34 @@ Object.Pipeline {
...
+					ibs 2112
+					obs 2145
+					cpc 2178
+				}
+				Object.Base.mod_cfg.3 {
+					ibs 3112
+					obs 3145
+					cpc 3178
+				}
+				num_mod_cfgs 3
```
lgirdwood, line 66: Do we need this, the CC should count this.

jsarha:
@ranj063 can help me out here?

tools/topology/topology2/cavs-mixin-mixout-hda.conf
jsarha: Ok. I was actually wondering why some widgets have explicitly set num_audio_formats and some doesn't.

ranj063:
> I am feeling quite clueless here, so even if this does not work, I push this out in hope someone can spot the problem.
> 
> For some reason the widget tuple array only contains the first instance of mod_cfg objects in the topology2 conf file. I am trying to read the array out with this code: https://github.com/jsarha/linux-sof/tree/peter-cpc-test

@jsarha the tplg is fine. but what you are missing is the parsing bits in the kernel. Take a look at how we parset multiple audio formats here: https://github.com/thesofproject/linux/blob/bef9852fe3228ce83717dada3ebcb7ad07c6099b/sound/soc/sof/topology.c#L1238

You need to allocate memory for "n" number of token sets first. And then parse the number of sets you're expecting like https://github.com/thesofproject/linux/blob/bef9852fe3228ce83717dada3ebcb7ad07c6099b/sound/soc/sof/ipc4-topology.c#L221 and then finally parse the tokens like https://github.com/thesofproject/linux/blob/bef9852fe3228ce83717dada3ebcb7ad07c6099b/sound/soc/sof/ipc4-topology.c#L257. Hope this helps

tools/topology/topology2/cavs-mixin-mixout-hda.conf
ranj063: some widgets support only 1 format, so they're set in teh class definition. Others support variable number of format based on tplg, so they're set in the top-level file

lgirdwood:
@jsarha @ujfalusi @ranj063 we need to have a per HW target per module file for each module i.e.
```
cavs/tgl/copier.conf
ace/mtl/copier.conf
```
and so on that we would include in teh high level generic pipeline files.

jsarha:
>

sys-pt1s:

kv2019i:
@jsarha Status? Please mark as draft if this is not intended for merging...

jsarha:
> @jsarha Status? Please mark as draft if this is not intended for merging...

This PR is good as such, but it does not contain the structures requested here: https://github.com/thesofproject/sof/pull/7348#issuecomment-1497387922

They could be implemented as separate PR. But maybe I'll mark this as draft as you requested and try to put all the pieces together. Then again I have no idea of any mod_cfg settings for any platform, so in the end there will not be any settings in the later PR either, just places for different platforms to put them.

jsarha:
Actually, I am not so sure about the second commit in this series. Maybe its better to include the mod_cfg.conf already in the component files (e.g. in include/component/*.conf), and not in the pipeline definitions. I already have an alternative commit: https://github.com/thesofproject/sof/commit/ab8538cb37559fb477d1e5c8878d634b71cbf68f  
@ranj063 , this is better, isn't it?

ranj063:
> Actually, I am not so sure about the second commit in this series. Maybe its better to include the mod_cfg.conf already in the component files (e.g. in include/component/*.conf), and not in the pipeline definitions. I already have an alternative commit: [ab8538c](https://github.com/thesofproject/sof/commit/ab8538cb37559fb477d1e5c8878d634b71cbf68f) @ranj063 , this is better, isn't it?

@jsarha yes adding it to the widget class definitions where they will be instantaited and used is probably better

jsarha:
New version pushed. This version should have what was requested in https://github.com/thesofproject/sof/pull/7348#issuecomment-1497387922 . Please check that I have added the placeholder and included it in actual component widgets. I already removed pipeline from them, but for many of them I have no idea what those entities are.

jsarha:
> @jsarha conflicts?

Solved now.

tools/topology/topology2/include/components/asrc.conf
```diff
@@ -29,6 +29,12 @@ Class.Widget."asrc" {
 	#include common component definition
 	<include/components/widget-common.conf>
 
+	#include platform specific module configs for the asrc module
```
ranj063, line 32: @jsarha is there value in this commit just yet? Can we not add the files as and when needed?

kv2019i: @lgirdwood Can you ack as this was an ask from you? I.e. to have the placeholders or not.

kv2019i (approved):
Not sure what is more likely, need to override values per platform, or per specific product, but in any case, this PR provides the necessary building blocks for both needs, so seems good.

jsarha:
Pushed forwards, and changes requested in mail, so turning to draft again.

Girdwood, Liam R:
> Jyri, a topology override must copy the manifest CPC schema exactly.
> Even the unused fields must be copied and populated with 0 today.

lgirdwood:
@jsarha now that the manifest and kernel CPC parts are working we should add the overide option for v2.7.
@ujfalusi @plbossart fyi.

plbossart:
> @jsarha now that the manifest and kernel CPC parts are working we should add the overide option for v2.7. @ujfalusi @plbossart fyi.

We need to agree on the definition of 'working'. I still routinely see messages telling me that CPC is not set ?

lgirdwood:
> > @jsarha now that the manifest and kernel CPC parts are working we should add the overide option for v2.7. @ujfalusi @plbossart fyi.
> 
> We need to agree on the definition of 'working'. I still routinely see messages telling me that CPC is not set ?

There are still a lot of holes in the manifest CPC data for each module and (format, rate, channels) configuration. @aiChaoSONG will be working on populating this prior to v2.7. This should fix the kernel warning messages for all tested configs.

The toplogy part will use the same CPC schema and kernel logic/IPC, but will allow 
1) Adding a missing CPC data element if one is missing in manifest for a new configuration.
2) Overide an existing manifest CPC data element if needed.

plbossart:
I'd like to reach the point where all CPC data is valid in the manifest and used for something meaningful, before we plan work on overriding stuff.

lgirdwood:
> I'd like to reach the point where all CPC data is valid in the manifest and used for something meaningful, before we plan work on overriding stuff.

Need to do both in parallel, to be ready for mistakes in manifest CPC.

plbossart:
> > I'd like to reach the point where all CPC data is valid in the manifest and used for something meaningful, before we plan work on overriding stuff.
> 
> Need to do both in parallel, to be ready for mistakes in manifest CPC.

we can't afford to hedge our bets for every single feature. The firmware team is responsible for providing correct CPC values, if they don't it's a bug and they fix it. 
 We have lots of other things to do really with SoundWire-related enablement.

lgirdwood:
> > > I'd like to reach the point where all CPC data is valid in the manifest and used for something meaningful, before we plan work on overriding stuff.
> > 
> > 
> > Need to do both in parallel, to be ready for mistakes in manifest CPC.
> 
> we can't afford to hedge our bets for every single feature. The firmware team is responsible for providing correct CPC values, if they don't it's a bug and they fix it. We have lots of other things to do really with SoundWire-related enablement.

CPC architecture is based on a single module unit test with no upstream/downstream module I$/D$ impact taken into consideration for module CPC performance delta when different pipeline topologies are used. This is not a big item and will be painful to modify/fix in manifest.

kv2019i:
In draft status, pushing to v2.9.

tools/topology/topology2/include/common/mod_cfg.conf
```diff
@@ -0,0 +1,45 @@
+# Class definition for module configuration object
+# mod_cfg objects can be instantiated as:
+#
+# Object.Base.mod_cfg."0" {
+#	ibs	384
+#	obs	384
+#	cpc	4600
```
btian1, line 7: can we add comments here:
for ibs/obs, it is initialized as: 48 * 2 * 4 = 384, it is 48k LL with 2 channel 32 bit format.
for cpc, I am not sure how CPC figured out as instantiated value?

tools/topology/topology2/include/common/mod_cfg.conf
```diff
@@ -0,0 +1,45 @@
...
+# }
+#
+
+Class.Base."mod_cfg" {
+
+    DefineAttribute."instance" {
+    }
+
+    # input buffer size
```
btian1, line 16: let's give comments like: 
ibs = (sample rate / (1000000/period)) * channels * valid_sample_width

tools/topology/topology2/include/components/widget-common-inc.conf
```diff
@@ -0,0 +1,5 @@
+#
+# This file is for include files needed for widget-common.conf
+#
+<include/common/mod_cfg.conf>
+
```
btian1, line 5: why not directly include mod_cfg.conf?

tools/topology/topology2/include/components/asrc.conf
btian1: 1. I agree add on demand.
2. if want to add in this patch, I would suggest there is a folder to cover all component instead of directly under platform/intel/mod_cfg/

kv2019i:
Stable-v2.9 branched, this didn't make the cut, bumping forward. @jsarha is this still valid (given bumps over multiple releases)?

kv2019i:
No activity in 2.7...2.9, clearing the milestone.

PR #7350, opened 2023-03-30
[CI ONLY] Test rimage PR 151

softwarecki:


sys-pt1s:

PR #7357, opened 2023-03-31
Fix handling of the mixer-module adapter combo on IPC3 platforms

paulstelian97:
[audio: mixer: Improve handling of inactive input streams](https://github.com/thesofproject/sof/commit/cdd2a88cb7dd3b969b22ae210079866578714a87)

This helps the mixer dynamically add/remove input streams when needed. Without this commit I would have the issue where the mixer would read from inactive inputs and outside the used portion of the corresponding input buffers.

[audio: module_adapter: Ensure prepare always runs on modules](https://github.com/thesofproject/sof/commit/bb5052355eb74fc59aa1ed8441cfb7ed4336be7e)

This commit makes sure I can actually start two streams with a delay on different mixer inputs (as opposed to in quick succession, as if given via the & symbol in a shell)

CC: @dbaluta

### DIFF
--- a/src/audio/mixer/mixer.c
+++ b/src/audio/mixer/mixer.c
@@ -82,11 +82,12 @@ static int mixer_process(struct processing_module *mod,
 	struct mixer_data *md = module_get_private_data(mod);
 	struct comp_dev *dev = mod->dev;
 	const struct audio_stream __sparse_cache *sources_stream[PLATFORM_MAX_STREAMS];
-	int32_t i = 0;
+	int32_t i = 0, j = 0;
 	uint32_t frames = INT32_MAX;
 	/* Redundant, but helps the compiler */
 	uint32_t source_bytes = 0;
 	uint32_t sink_bytes;
+	int active_input_buffers = 0;
 
 	comp_dbg(dev, "mixer_process() %d", num_input_buffers);
 
@@ -105,10 +106,11 @@ static int mixer_process(struct processing_module *mod,
 		if (avail_frames == 0)
 			continue;
 
+		active_input_buffers++;
 		frames = MIN(frames, avail_frames);
 	}
 
-	if (!num_input_buffers || (frames == 0 && md->sources_inactive)) {
+	if (!active_input_buffers) {
 		/*
 		 * Generate silence when sources are inactive. When
 		 * sources change to active, additionally keep
@@ -119,18 +121,11 @@ static int mixer_process(struct processing_module *mod,
 		if (!audio_stream_set_zero(mod->output_buffers[0].data, sink_bytes))
 			mod->output_buffers[0].size = sink_bytes;
 
-		md->sources_inactive = true;
-
 		return 0;
 	}
 
 	/* Every source has the same format, so calculate bytes based on the first one */
 	source_bytes = frames * audio_stream_frame_bytes(mod->input_buffers[0].data);
-
-	if (md->sources_inactive) {
-		md->sources_inactive = false;
-		comp_dbg(dev, "mixer_copy exit sources_inactive state");
-	}
 
 	sink_bytes = frames * audio_stream_frame_bytes(mod->output_buffers[0].data);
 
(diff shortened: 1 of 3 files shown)


src/audio/mixer/mixer.c
```diff
@@ -119,33 +121,39 @@ static int mixer_process(struct processing_module *mod,
 		if (!audio_stream_set_zero(mod->output_buffers[0].data, sink_bytes))
 			mod->output_buffers[0].size = sink_bytes;
 
-		md->sources_inactive = true;
-
```
paulstelian97, line 123: TODO: Since this field is no longer used, should I try to remove it?

src/audio/mixer/mixer.c
```diff
@@ -119,33 +121,39 @@ static int mixer_process(struct processing_module *mod,
...
 	sink_bytes = frames * audio_stream_frame_bytes(mod->output_buffers[0].data);
 
 	comp_dbg(dev, "mixer_process(), source_bytes = 0x%x, sink_bytes = 0x%x",
 		 source_bytes, sink_bytes);
 
 	/* mix streams */
-	for (i = 0; i < num_input_buffers; i++)
-		sources_stream[i] = mod->input_buffers[i].data;
+	int j = 0;
```
paulstelian97, line 136: Do I move this to the top?

src/audio/mixer/mixer.c
```diff
@@ -119,33 +121,39 @@ static int mixer_process(struct processing_module *mod,
...
+		/* if one source is inactive, skip it */
+		if (avail_frames == 0)
+			continue;
+
+		sources_stream[j++] = mod->input_buffers[i].data;
+	}
 
-	md->mix_func(dev, mod->output_buffers[0].data, sources_stream, num_input_buffers, frames);
+	if (j)
+		md->mix_func(dev, mod->output_buffers[0].data, sources_stream, j, frames);
```
paulstelian97, line 152: Perhaps I should use active_input_buffers here and below?

src/audio/mixer/mixer.c
dbaluta: yes, remove it if no longer used.

src/audio/mixer/mixer.c
iuliana-prodan: You need j anyway to have the exact active buffers, so you can continue using j here and below

src/audio/mixer/mixer.c
iuliana-prodan: yes, you can add it near i :)

dbaluta (approved):

src/audio/mixer/mixer.c
kv2019i: Hmm, wouldn't j==active_input_buffers ? If j can be differnent, then I think it deserves a longer more descriptive name as it's used across this long function (i.e. not just a loop index variable anymore).

kv2019i (approved):
Thanks! One minor readibility/style comment, but nothing to preventing merge.

src/audio/mixer/mixer.c
paulstelian97: They can differ if there's a race changing which buffers are active between the first check and this loop. I can write an assert stating that they don't differ I suppose.

andrula-song (approved):

lyakh (commented):
The second patch in this PR is touching some very central code and not only for IPC3

src/audio/module_adapter/module_adapter.c
```diff
@@ -167,21 +167,16 @@ int module_adapter_prepare(struct comp_dev *dev)
...
-	ret = comp_set_state(dev, COMP_TRIGGER_PREPARE);
-	if (ret < 0)
-		return ret;
+		comp_err(dev, "module_adapter_prepare() error %x: module prepare failed",
+			 ret);
 
-	if (ret == COMP_STATUS_STATE_ALREADY_SET) {
-		comp_warn(dev, "module_adapter_prepare(): module has already been prepared");
-		return PPL_STATUS_PATH_STOP;
+		return -EIO;
```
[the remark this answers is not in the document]
dbaluta: This is addressed with https://github.com/thesofproject/sof/pull/7422/commits/72da532185b627450af522e703edc9cac9827b93

I think @paulstelian97  tried to keep the original semantic of the code

PR #7359, opened 2023-03-31
Audio: Component: Update to component.h module adapter conversions

singalsu:
In component.h the eq_iir and mixer have still the old component API function prototypes. They can be removed as not needed. These components do not support legacy mode.

Similar update is done to testbench common_test.h. Remove eq_iir and mixer old init function prototypes and add to module adapter style init functions mfcc and mixer those were missing.


kv2019i (approved):

dbaluta (approved):

lgirdwood:
SOFCI TEST

sys-pt1s:

lgirdwood:
SOFCI TEST

tools/testbench/include/testbench/common_test.h
```diff
@@ -118,6 +116,8 @@ void sys_comp_tdfb_init(void);
 void sys_comp_module_demux_interface_init(void);
 void sys_comp_module_eq_fir_interface_init(void);
 void sys_comp_module_eq_iir_interface_init(void);
+void sys_comp_module_mfcc_interface_init(void);
+void sys_comp_module_mixer_interface_init(void);
```
lyakh, line 120: if it worked until now without these - does it mean, that this isn't compile-tested by the CI or maybe that these prototypes aren't needed and those functions can be static?

singalsu: I wonder if recent testbench patches have forgotten these? But these will definitely be needed to load the components (unless they get deprecate - e.g. mixer in favor of mixin/mixout?) and run them.

singalsu: @lyakh Sorry, I was wrong. These protypes are no more necessary. The component.h is used, need to delete all these in the patch.

lyakh (approved):

kv2019i:
SOFCI TEST

lgirdwood (approved):

PR #7360, opened 2023-04-03
Some more topology cleanups

ranj063:
Rename passthrough-capture.conf to be-dai.conf
And remove setting the default node_type in the BE DAI pipeline class definitions and set them during pipeline object instantiation based on the DAI type.

### DIFF
--- a/tools/topology/topology2/cavs-mixin-mixout-efx-hda.conf
+++ b/tools/topology/topology2/cavs-mixin-mixout-efx-hda.conf
@@ -80,7 +80,7 @@ Object.Pipeline {
 		}
 	]
 
-	passthrough-be [
+	io-gateway [
 		{
 			index		4
 			direction	capture
--- a/tools/topology/topology2/cavs-mixin-mixout-efx-hda.conf
+++ b/tools/topology/topology2/cavs-mixin-mixout-efx-hda.conf
@@ -33,6 +33,7 @@ Object.Pipeline {
 			index 2
 
 			Object.Widget.copier.1 {
+				node_type $HDA_LINK_OUTPUT_CLASS
 				stream_name $HDA_ANALOG_DAI_NAME
 				dai_type "HDA"
 				copier_type "HDA"
--- a/tools/topology/topology2/cavs-src-mixin-mixout-hda.conf
+++ b/tools/topology/topology2/cavs-src-mixin-mixout-hda.conf
@@ -33,6 +33,7 @@ Object.Pipeline {
 			index 2
 
 			Object.Widget.copier.1 {
+				node_type $HDA_LINK_OUTPUT_CLASS
 				stream_name $HDA_ANALOG_DAI_NAME
 				dai_type "HDA"
 				copier_type "HDA"
--- a/tools/topology/topology2/cavs-src-mixin-mixout-hda.conf
+++ b/tools/topology/topology2/cavs-src-mixin-mixout-hda.conf
@@ -73,7 +73,7 @@ Object.Pipeline {
 		}
 	]
 
-	passthrough-be [
+	io-gateway [
 		{
 			index		4
 			direction	capture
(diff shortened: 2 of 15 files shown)


tools/topology/topology2/cavs-mixin-mixout-efx-hda.conf
```diff
@@ -80,7 +80,7 @@ Object.Pipeline {
 		}
 	]
 
-	passthrough-be [
+	be-dai [
```
plbossart, line 84: Sorry @ranj063 I find the naming confusing

a) a DAI can be bidirectional - this isn't
b) This is really a BE copier, there's nothing related to the DAI configuration, is there?

tools/topology/topology2/include/pipelines/cavs/be-dai.conf
```diff
@@ -33,8 +33,8 @@ Class.Pipeline."be-dai" {
 		]
 
 		#
-		# mixin-playback objects instantiated within the same alsaconf node must have
-		# unique pipeline_id attribute
+		# be-dai objects instantiated within the same alsaconf node must have
+		# unique instance attribute
```
plbossart, line 37: that's even more confusing for the case of an SSP, we need two instance attributes for the same SSP.

tools/topology/topology2/platform/intel/hdmi-generic.conf
```diff
@@ -64,6 +64,7 @@ Object.Pipeline {
 			index $HDMI1_DAI_PIPELINE_ID
 
 			Object.Widget.copier.1 {
+				node_type $HDA_LINK_OUTPUT_CLASS
```
plbossart, line 67: can we make the node_type a required attributes in the class definition? That would catch misses in the definition of the pipelines, no?

tools/topology/topology2/cavs-src-mixin-mixout-hda.conf
```diff
@@ -33,6 +33,7 @@ Object.Pipeline {
 			index 2
 
 			Object.Widget.copier.1 {
+				node_type $HDA_LINK_OUTPUT_CLASS
```
plbossart, line 36: same comment, if node_type is a required attribute let's declare it as such in all the base pipeline classes.

tools/topology/topology2/sof-hda-generic.conf
```diff
@@ -14,7 +14,6 @@
 <host-copier-gain-src-mixin-playback.conf>
 <mixout-gain-dai-copier-playback.conf>
 <mixout-gain-efx-dai-copier-playback.conf>
-<dai-copier-gain-mixin-capture.conf>
```
plbossart, line 17: question: is there a way to detect when an included file is not used at all?

plbossart (commented):
couple of comments @ranj063

tools/topology/topology2/cavs-mixin-mixout-efx-hda.conf
ranj063: a) yes this is birectional. its simply a BE DAI copier. The reason why I call this a be-dai instead of a be-copier is because it has a pipeline widget as well, not just a copier.
b) It's a copier alright and the DAI config for the copier comes from the type of DAI and is added at the time of instantiation.

But having said this, I really dont have a preference. I'll go with be-copier if you prefer that. I just dont think that's accurate either

tools/topology/topology2/platform/intel/hdmi-generic.conf
ranj063: Yes, that my eventual goal but I can't do that just yet. I need to clean a lot more pipelines for it

tools/topology/topology2/cavs-src-mixin-mixout-hda.conf
ranj063: has to be part of copier class not pipeline class.

kv2019i (approved):
No strong preference w.r.t. copier-or-dai, but anyways seem better to not call this passthrough.

ranj063:
> Can I suggest IO_Gateway then, it seems to be the term that's closest to the IPC4 documentation.
yes, fixed now.

plbossart (approved):
LGTM

gkbldcig:

PR #7362, opened 2023-04-03
Compile DMA domain by default for i.MX with Zephyr & enable tracing on i.MX platforms

dbaluta:
DMA domain with Zephyr works fine for i.MX platforms. So, remove "experimental" from description and enable it by default for i.MX platforms.

Second patch enables tracing for i.MX platforms with zephyr.

### DIFF
--- a/zephyr/Kconfig
+++ b/zephyr/Kconfig
@@ -29,13 +29,10 @@ config SOF_ZEPHYR_STRICT_HEADERS
 	  If unsure, say n.
 
 config DMA_DOMAIN
-	bool "Experimental: Enable the usage of DMA domain."
-	default n
+	bool "Enable the usage of DMA domain."
+	default y if IMX
 	help
 	  This enables the usage of the DMA domain in scheduling.
-	  This will most likely come to replace dma_multi_chan
-	  and dma_single_chan in Zephyr once it becomes more
-	  stable.
 
 config ZEPHYR_DP_SCHEDULER
 	bool "use Zephyr thread based DP scheduler"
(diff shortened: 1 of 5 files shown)


paulstelian97:
Code itself looks good, I just need to be convinced it's properly tested.

iuliana-prodan (approved):

paulstelian97 (approved):
Conditional approve: compilation must not break on platforms like TGL like the previous revision. If that is true, then this looks good. (and from a cursory look that doesn't seem to be a problem anymore)

dbaluta:
@paulstelian97  Fixed compilation problems. Also, manually tested on i.MX8MP. Will start a nightly test run to see if there are any issues reported.

zephyr/Kconfig
```diff
@@ -29,13 +29,11 @@ config SOF_ZEPHYR_STRICT_HEADERS
 	  If unsure, say n.
 
 config DMA_DOMAIN
-	bool "Experimental: Enable the usage of DMA domain."
+	bool "Enable the usage of DMA domain."
+	default y if IMX
 	default n
```
LaurentiuM1234, line 34: nitpick: is `default n` really required? It should be set to `n` if CONFIG_IMX is not enabled. Otherwise, LGTM.

dbaluta: Yes, I think you are right. It is set to `n`. Will update.

paulstelian97 (approved):
I will accept this, though I still need to understand how the default value works in kconfig if no "default" applies.

LaurentiuM1234 (approved):

dbaluta:
> I will accept this, though I still need to understand how the default value works in kconfig if no "default" applies.

When kconfig entry doesn't have a default clause it goes directly to `N`.

lgirdwood (approved):

Issue #6837, opened 2022-12-19
[BUG] distortion during playback on MTL-NOCODEC


keqiaozhang:
**Describe the bug**
we observed the waveform distortion on MTL-NOCODEC, we have a similar issue [SOF#6727](https://github.com/thesofproject/sof/issues/6727) on mtl-stable-002, but it has been fixed on 002 branch, so let's use this bug to track the distortion issue on main branch.

![image](https://user-images.githubusercontent.com/22468760/208354033-59e7c892-1dbb-4155-b033-2ebaea69d8ac.png)

**To Reproduce**
~/sof-test/test-case/check-alsabat.sh -p hw:sofnocodec,0 -c hw:sofnocodec,0 -C 2

**Reproduction Rate**
100%

**Environment**
Kernel Branch: topic/sof-dev
Kernel Commit: 
[c41b1735d3da](https://github.com/thesofproject/linux/commit/c41b1735d3da)
SOF Branch: main
SOF Commit: 
[e6f72dee8289](https://github.com/thesofproject/sof/commit/e6f72dee8289)
Platform: MTL-NOCODEC

[bat.wav.3Vbz0k.zip](https://github.com/thesofproject/sof/files/10256182/bat.wav.3Vbz0k.zip)

juimonen:
This is because of missing link clock setting in mtl/ace platforms in ssp driver.

mtl 003 branch has a hard coded quick fix for "link clock", but that is not in sof main.

"non hard coded" fix is coming from:
https://github.com/zephyrproject-rtos/zephyr/pull/53297

after this, there also needs to be a topology change for mtl to configure/set the link clock.

wszypelt:
Reproduction test under implementation

keqiaozhang:
This issue is still exists in CI.
![image](https://user-images.githubusercontent.com/22468760/220502218-8601db82-f23e-4276-a58c-7c0601a6e114.png)

lgirdwood:
ok, looks like we could still be missing a MTL branch patch in main @mwasko fyi.

juimonen:
So we seem to have 2 overlapping bugs here:
1) link clock setting in mtl (this should not cause choppiness, but will amplify it)
- there is a fix pending for it https://github.com/thesofproject/sof/pull/7060
2) even without anything link clock related we have the choppiness:
- https://github.com/thesofproject/sof/issues/7184
- something is fishy in the feedback stream

fredoh9:
[...]

PR #7363, opened 2023-04-03
topology2: ssp: Add blob version

ranj063:
Add the SSP blob version and set it based on the platform.

This is a replacement for https://github.com/thesofproject/sof/pull/7060

Fixes: https://github.com/thesofproject/sof/issues/6837


aiChaoSONG (approved):
Changes look good and simpler than #7060.

In the SSP context, the word `version` maybe too general, I think `blob_version` maybe better, but no big deal.

tools/topology/topology2/platform/intel/tgl.conf
```diff
@@ -0,0 +1,4 @@
+# TGL/ADL-specific variable definitions
+Define {
+	SSP_BLOB_VERSION		0x100
```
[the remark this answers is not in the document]
ranj063: @plbossart I think blob version 1.00 for TGL/ADL is what makes it possible for the alsatplg compiler to ignore the aux parameters that include the link clock_source

tools/topology/topology2/common_definitions.conf
```diff
@@ -58,4 +58,7 @@ Define {
 	SPI_INPUT_CLASS				26 # SPI Input (DSP <-)
 
 	DEEPBUFFER_FW_DMA_MS			100 # 100 ms copier dma size
+
+	SSP_BLOB_VERSION_1_0			0x100
+	SSP_BLOB_VERSION_1_5			0x105
```
lgirdwood, line 63: 0x150 ?

kv2019i (approved):
Not entirely sure how well this aligns with https://github.com/thesofproject/sof/pull/7370 , but put these two together seems to be a clean solution. Approving at least this one.

tools/topology/topology2/common_definitions.conf
plbossart: ack, this is 105 - as strange as it might look

tools/topology/topology2/platform/intel/tgl.conf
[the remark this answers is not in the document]
plbossart: @kv2019i I don't think that's true any longer with the change suggested by @ranj063 to make zephyr deal with aux in all cases, and do nothing for TGL.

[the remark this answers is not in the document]
plbossart: @kv2019i see https://github.com/zephyrproject-rtos/zephyr/pull/56497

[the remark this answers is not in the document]
plbossart: I think the SSP blob 1.05 was added before TGL. @mmaka1 @mwasko can you please confirm?

ranj063:
@plbossart @kv2019i @lgirdwood I got to the root of the problem:
1. There's a strict restriction wrt the alsatplg compiler which requires the "Object.Base.link_config.1" to be defined right within the Object.Base.hw_config.N{} node. This is wrong and should be fixed in the compiler so that the Object definition in hw_config_cardinal_clock.conf will work as expected.
2. blob version 0x100 is required for TGL with the above change so that the clock_source setting will be ignored.
3. Alternatively, if we set _c105 for TGL, we need the zephyr patch https://github.com/zephyrproject-rtos/zephyr/pull/56497

In conclusion, I think this is what we should do. Merge this PR so that we can unblock the alsabat tests in CI. Then work on fixing the alsatplg to allow link_config to come from the class definition instead of the object instance. Finally after the zephyr patch is merged, set the blob type to 0x105 for all platforms.

Let me know what you think.

tools/topology/topology2/platform/intel/bt-ssp-config.conf
```diff
@@ -50,6 +56,9 @@ Object.Dai.SSP [
 			tdm_slots	2
 			tx_slots	3
 			rx_slots	0
+			Object.Base.link_config.1 {
+				clock_source	0
```
kv2019i, line 60: Should this be the cardinal clock (i.e. "1"), so we can use integer divider for 1.536Mhz?

tools/topology/topology2/platform/intel/bt-ssp-config-lbm.conf
```diff
@@ -51,6 +57,9 @@ Object.Dai.SSP [
 			tdm_slots	2
 			tx_slots	3
 			rx_slots	3
+			Object.Base.link_config.1 {
+				clock_source	0
```
kv2019i, line 61: See comment for non-loopback version.

kv2019i (commented):
One comment inline about BT loopback topologies. Otherwise this looks like a clean approach.

tools/topology/topology2/platform/intel/bt-ssp-config.conf
ranj063: no should 0 for xtal as xtal is 38.4 and 38.4M/1.536M = 25

fredoh9:
I check this PR with MTL nocodec device, it generates solid waveform, alsabat test consistently passes!

tools/topology/topology2/platform/intel/bt-ssp-config.conf
ranj063: @kv2019i I will let you answer this question and follow up with another PR if we need to change the clock source for BT. For now, it actually makes no difference since we dont build the topology for MTL

PR #7369, opened 2023-04-04
platform: ace: register clock using core_kcps_adjust api

pblaszko:
Fixes related to cpu-clock-manager.
Fix for Windows OED scenarios, where driver sends LargeConfigSet::RegisterKcps. Such call was causing unexpected behavior, because cpu-clock-manage API is not used in platform_init when registering clock.

### DIFF
--- a/src/platform/intel/ace/platform.c
+++ b/src/platform/intel/ace/platform.c
@@ -12,6 +12,7 @@
 #include <sof/ipc/driver.h>
 #include <sof/ipc/msg.h>
 #include <sof/lib/agent.h>
+#include <sof/lib/cpu-clk-manager.h>
 #include <sof/lib/mm_heap.h>
 #include <sof/lib/watchdog.h>
 #include <sof/schedule/edf_schedule.h>
@@ -80,7 +81,13 @@ int platform_init(struct sof *sof)
 
 	trace_point(TRACE_BOOT_PLATFORM_CLOCK);
 	platform_clock_init(sof);
-	clock_set_freq(CLK_CPU(cpu_get_id()), CLK_MAX_CPU_HZ);
+
+	/* Set DSP clock to MAX using KCPS API. Value should be lowered when KCPS API
+	 * for modules is implemented
+	 */
+	ret = core_kcps_adjust(cpu_get_id(), CLK_MAX_CPU_HZ / 1000);
+	if (ret < 0)
+		return ret;
 
 	trace_point(TRACE_BOOT_PLATFORM_SCHED);
 	scheduler_init_edf();


pblaszko:
SOFCI TEST

dnikodem (approved):

tmleman (approved):

pblaszko:
SOFCI TEST

src/platform/intel/ace/platform.c
```diff
@@ -80,7 +81,9 @@ int platform_init(struct sof *sof)
 
 	trace_point(TRACE_BOOT_PLATFORM_CLOCK);
 	platform_clock_init(sof);
-	clock_set_freq(CLK_CPU(cpu_get_id()), CLK_MAX_CPU_HZ);
+	ret = core_kcps_adjust(cpu_get_id(), CLK_MAX_CPU_HZ / 1000);
```
lgirdwood, line 84: so IIUC correctly, this sets max clock at init using kcps API now ? Maybe an inline comment above to state the reason and any expectation to have driver lower this when its ready.

pblaszko: @lgirdwood yes, you're correct. Inline comment added.

pblaszko:
SOFCI TEST

sys-pt1s:

src/lib/cpu-clk-manager.c
```diff
@@ -62,14 +61,9 @@ int core_kcps_adjust(int adjusted_core_id, int kcps_delta)
...
 	freq = max_core_consumption();
 
-	for (core_id = 0; core_id < CONFIG_CORE_COUNT; core_id++) {
-		/* Convert kcps to cps */
-		ret = request_freq_change(core_id, freq * 1000);
-		if (ret < 0)
-			goto out;
-	}
+	/* Convert kcps to cps */
+	ret = request_freq_change(adjusted_core_id, freq * 1000);
```
KFrydryk, line 65: That is correct in the sense, that there is only one register setting clocks for all of the cores (for cavs and ace at least), but please double check if this call updates all of the `clock_info` structures. `clock_set_freq` does not explicitly do that. For sofs CAVS implementation it works fine, as the platform specific `clk_info->set_freq` function iterates over cores. When using zephyrs implementation though, there is a separate structure holding clock info (see zephyr/soc/xtensa/intel_adsp/common/clk.c). This used to be unaligned with sofs structures and might be worth checking.

pblaszko: Good catch, thank you @KFrydryk !
Indeed, Zephyr stores clock info per Core:
`static struct adsp_clock_info platform_clocks[CONFIG_MP_MAX_NUM_CPUS];`

It does not make too much sense and it is confusing. Also potentially may lead to some issues with clock management.
I think this should be investigated separately then, with another PR's for SOF and Zephyr.

I am going to remove this commit from this PR, as it is not critical for the moment.

lgirdwood (approved):

KFrydryk (approved):

lgirdwood:
@wszypelt @lrudyX CI seems blocked, good to merge ?

wszypelt:
@lgirdwood good to merge ;)

kv2019i (approved):

PR #7370, opened 2023-04-04
topology2: Move the hw_config class definition

ranj063:
Move and rename the hw_config class to
platform/intel/hw_config_cardinal_clk.conf.
This hw_config is specific to Intel SSP DAI that use the cardinal clock for the mclk frequency.

Ideally, this class should have immutable mclk_frequency and link clock source. But because the alsatplg compiler expects the name of the hw config class to be "hw_config" without any extensions, it is left as modifiable for now. Once the topology compiler is modified, this will be made immutable in a follow up PR.

Also, introduce a new hw_config_simple.conf file that contains the hw_config definition for the HDA/DMIC/SDW type DAIs with only the ID and name attributes.

This is the replacement for the second part of #7060 to set the link clock source


tools/topology/topology2/platform/intel/hw_config_cardinal_clk.conf
```diff
@@ -44,7 +45,20 @@ Class.Base."hw_config" {
...
-	DefineAttribute."bclk_freq" {}
+	DefineAttribute."bclk_freq" {
+		constraints {
+			# Allowed values of bclk derived from mclk_freq of 24.576MHz
+			!valid_values [
+				"12288000"
+				"6144000"
+				"3072000"
+				"1536000"
+				"768000"
```
plbossart, line 56: missing 384000?

tools/topology/topology2/cavs-nocodec.conf
```diff
@@ -30,7 +30,7 @@
 <fe_dai.conf>
 <ssp.conf>
 <dmic.conf>
-<hw_config.conf>
+<hw_config_cardinal_clk.conf>
```
plbossart, line 33: we probably need to introduce intel prefixes?

tools/topology/topology2/platform/intel/hw_config_simple.conf
```diff
@@ -0,0 +1,35 @@
...
+#
+# Object.Base.hw_config.1 {
+# 		id		0
+#		name		"HDA0"
+# 	}
+#
+# where 1 is the unique instance ID for the hw_config object within the same alsaconf
+# node.
+
+Class.Base."hw_config" {
```
plbossart, line 12: I am not following how we can have the same class name for two different definitions, here and https://github.com/thesofproject/sof/pull/7370/commits/07ffe5d2e55e6f245e639eb6d2ecacfe5e47090b#diff-6097d83c177264e03795b3cb083915a0153a69fe445ea6046004160e75209628R15

tools/topology/topology2/cavs-nocodec.conf
ranj063: this file is in the platform/intel folder already, do you still want to add the intel prefix?

plbossart: we should have something like <intel/hw_config_cardinal_clk.conf> then to avoid conflicts. the 'flat' naming and search path guarantees someone will screw up one day.

tools/topology/topology2/platform/intel/hw_config_simple.conf
ranj063: This is alsatplg compiler restriction for now. We can only include one or the other so we have only one class definition included. so the class name is the same "hw_config" because thats what the compiler expects

gkbldcig:

tools/topology/topology2/cavs-nocodec.conf
ranj063: ack fixed now

kv2019i (commented):
Still not following how this handles the differences in SSP blob version,  With v1, we want to specify mclk and let driver pick the source, with v1.5, we want to specify clock source in tplg and pass that. But we need to cover both options.

Could we have separate versions of hw_config_cardinal_clk.conf for these two cases? link-config definitions would always be in e.g. bt-ssp-config.*conf, but ignored on v1.5 blob platforms. 

UPDATE: or hmm, or should this go together with https://github.com/thesofproject/sof/pull/7363 ? that would explain how the blob verssion is set.

tools/topology/topology2/platform/intel/hw_config_cardinal_clk.conf
```diff
@@ -1,15 +1,16 @@
...
-# 		mclk_freq	24000000
-# 		bclk_freq	4800000
-# 		tdm_slot_width	25
+# 		bclk_freq	3072000
+#		tdm_slot_width  32
 # 	}
 #
-# where NAME is the unique instance name for the hw_config object
-# within the same alsaconf node.
+# where 1 is the unique instance ID for the hw_config object within the same alsaconf
```
lgirdwood, line 10: Why use a number when a name is more descriptive ?

plbossart:
@kv2019i the main point is that the value of the MCLK and the link source have to be selected together. You cannot select the audio cardinal clock of 24.576 MHz and use link_source 0 (xtal). That's broken by design. The main idea is to bundle together fields that need to be handled consistently.

tools/topology/topology2/platform/intel/hw_config_cardinal_clk.conf
ranj063: @lgirdwood all objects need to have instance so that we can use arrays.

ranj063:
clubbed with #7363. Closing this one

PR #7405, opened 2023-04-05
topology: imx: updates and add new codec

iuliana-prodan:
Combine some topology files with different codecs and rate.
Also, add new tplg for wm8962 codec, for i.MX8QM and i.MX8MP.


LaurentiuM1234:
A few suggestions:

1. Maybe use a more generic name for the topologies? For example: `sof-imx8mp-wm-generic` or even `sof-imx8mp-generic` if this can also be used for CS codecs. Seems kinda odd that `sof-imx8mp-wm8962` is built from `sof-imx8mp-wm8960`
2. Could try to introduce macros such as:
```
define(`STREAM_NAME',
`ifelse(`$1', `wm8960', `wm8960-hifi',
`$1', `wm8962', `wm8962',
`fatal_error(`Codec $1 not supported.')')'
)
```
This would get rid of the need to pass the stream name as the codec name when building the .tplg file.

3. Could also try to introduce some utility macros such as the one above for MCLK, BCLK or even the for the pipeline rate. This would get rid of the 'RATE' variable and should make the code a bit easier to read as it would get rid of the DAI_CONFIG ifelse cluster.

Otherwise, LGTM.

iuliana-prodan:
> A few suggestions:
> 

@LaurentiuM1234  thanks for all the suggestion, I agree with almost all of them :)
I tried to make small changes in order to have everything ready for next week, when we have a patch release for i.MX8MP and we need the support for wm8962.
I agree we can improve the topologies we have now, but I'm not sure if we should invest too much time on this. 
Maybe it will be better to look into topology2.

> 1. Maybe use a more generic name for the topologies? For example: `sof-imx8mp-wm-generic` or even `sof-imx8mp-generic` if this can also be used for CS codecs. Seems kinda odd that `sof-imx8mp-wm8962` is built from `sof-imx8mp-wm8960`

Yes, I could use a sof-imx8mp-wm-generic - just for WM codecs.
I don't want to mix wm and cs codecs - they have different DAIs, clocks - I don't want to complicate things too much. 
It's already hard to follow (at least for me) the m4 files. So, adding too many variables it will be worst.

> 2. Could try to introduce macros such as:
> 
> ```
> define(`STREAM_NAME',
> `ifelse(`$1', `wm8960', `wm8960-hifi',
> `$1', `wm8962', `wm8962',
> `fatal_error(`Codec $1 not supported.')')'
> )
> ```
> 
> This would get rid of the need to pass the stream name as the codec name when building the .tplg file.
> 
Yes, I can add the STREAM_NAME.

> 3. Could also try to introduce some utility macros such as the one above for MCLK, BCLK or even the for the pipeline rate. This would get rid of the 'RATE' variable and should make the code a bit easier to read as it would get rid of the DAI_CONFIG ifelse cluster.
> 
MLCK and BLCK I believe they are computed somehow. I wouldn't complicate too much the m4 file.
I will think about this, but, at least for now, I would like to keep it as is.

I already used RATE variable for 8MP, since there are multiple rates support for different wm codec.
For 8QM I didn't change it since we only have 48000 on both wm8960 and wm8962. We should add it when necessary.

Fow now, the DAI_CONFIG ifelse cluster will remain. Will have to think of a simple and elegant solution for MCLK, BCLK, codec type (master or provider).

> Otherwise, LGTM.
> 
> EDIT: If trying to go for a more generic approach, it might be worth trying to also pass MCLK, BCLK, FSYNC and SAI_INDEX as arguments.

I didn't add SAI_INDEX since we have the same index for all codecs. I thinks that some variables worth adding when necessary, now it will just make the m4 more difficult to read.

Maybe the MCLK/BCLK can be define as you suggested with the STREAM_NAME example above 🤔

I'll update this PR with some of these suggestions, others will be addressed in another PR.
Hope this will be ok for you!

Thanks!

paulstelian97 (approved):

sys-pt1s:

iuliana-prodan:
Added a new version - v2:
- rearrange the topologies based on platform name in CMakeLists.txt;
- use STREAM_NAME based on CODEC name;
- define DAI_BE_NAME from STREAM_NAME.

I've kept the original names for topology files (sof-imx8mp-wm8960 and sof-imx8-wm8960). 
For other codecs are generated from these.

LaurentiuM1234 (approved):

dbaluta (approved):

lgirdwood (approved):

paulstelian97 (approved):
This is even cleaner than the previously approved version, re-approving.

dbaluta:
SOFCI TEST

dbaluta:
@lgirdwood can you help merge this? Looks like CI is blocked. Would like to have this merged to start tests over weekend.

marc-hb:
>  Looks like CI is blocked

"Internal Intel CI System/merge" ignores topology changes completely, always has. So no need to wait for or restart anything.

Other results are either ok or unrelated failures.

plbossart (approved):

plbossart:
merging to unlock tests

PR #7413, opened 2023-04-06
[CI ONLY] Test rimage PR 155

softwarecki:
changed rimage revision in west and git submodule


sys-pt1s:

lgirdwood:
@softwarecki I would say this is a pass.

PR #7422, opened 2023-04-07
module_adapter: Propagate error from module specific callback

dbaluta:
When module_prepare() fails we need to return to calling module the exact specific error not just -EIO.

There is no real use for Host AP to receive an error code that doesn't reflect the actual error.

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -170,13 +170,10 @@ int module_adapter_prepare(struct comp_dev *dev)
 	/* Prepare module */
 	ret = module_prepare(mod);
 	if (ret) {
-		if (ret == PPL_STATUS_PATH_STOP)
-			return ret;
-
-		comp_err(dev, "module_adapter_prepare() error %x: module prepare failed",
-			 ret);
-
-		return -EIO;
+		if (ret != PPL_STATUS_PATH_STOP)
+			comp_err(dev, "module_adapter_prepare() error %x: module prepare failed",
+				 ret);
+		return ret;
 	}
 
 	/* Get period_bytes first on prepare(). At this point it is guaranteed that the stream


paulstelian97 (approved):
I'm curious why it was set as -EIO in the first place

iuliana-prodan (approved):

dbaluta:
SOFCI TEST

kv2019i (approved):
Only reason that comes to mind if module_prepare() has different signature (e.g. different type of error values) than module_adapter_prepare(), but this doesn't seem to be the case.

dbaluta:
> I'm curious why it was set as -EIO in the first place

You wrote the code :) so only God and you might know the answer.

paulstelian97:
Git blame points me out because of a few reshuffling commits where I failed to notice/understand its purpose and I preserved it as-is. The -EIO return seems to be there from before the rename to module_adapter, I can keep looking if it truly matters.

dbaluta:
@lgirdwood @lyakh  can you please help merging this.

src/audio/module_adapter/module_adapter.c
```diff
@@ -176,7 +176,7 @@ int module_adapter_prepare(struct comp_dev *dev)
 		comp_err(dev, "module_adapter_prepare() error %x: module prepare failed",
 			 ret);
 
-		return -EIO;
+		return ret;
```
lyakh, line 179: Nice, I actually commented on this exact code earlier today. If you want you can make it even more compact as
```
if (ret != PPL_STATUS_PATH_STOP)
	comp_err(dev, "module_adapter_prepare() error %x: module prepare failed", ret);
return ret;
```
but that's up to you

lyakh (approved):

dbaluta:
SOFCI TEST

PR #7424, opened 2023-04-10
Smart_amp: Convert dummy smart_amp component to use the module adapter

andrula-song:
Convert the dummy smart_amp(smart_amp_test) module to use the module adapter interface. Add "id" into input_stream_buffer to record buffer id for feedback buffer identification.

### DIFF
--- a/src/samples/audio/CMakeLists.txt
+++ b/src/samples/audio/CMakeLists.txt
@@ -1,9 +1,14 @@
 # SPDX-License-Identifier: BSD-3-Clause
-
 if(CONFIG_SAMPLE_SMART_AMP)
-	add_local_sources(sof
-		smart_amp_test.c
-	)
+	if(CONFIG_IPC_MAJOR_3)
+		add_local_sources(sof
+			smart_amp_test_ipc3.c
+		)
+	elseif(CONFIG_IPC_MAJOR_4)
+		add_local_sources(sof
+			smart_amp_test_ipc4.c
+		)
+	endif()
 endif()
 
 if(CONFIG_SAMPLE_KEYPHRASE)
--- a/src/samples/audio/smart_amp_test_ipc3.c
+++ b/src/samples/audio/smart_amp_test_ipc3.c
@@ -42,8 +42,6 @@ struct smart_amp_data {
 	struct comp_buffer *feedback_buf; /**< feedback source buffer */
 	struct comp_buffer *sink_buf; /**< sink buffer */
 
-	struct k_mutex lock; /**< protect feedback_buf updated */
-
 	smart_amp_proc process;
 
 	uint32_t in_channels;
@@ -76,18 +74,14 @@ static struct comp_dev *smart_amp_new(const struct comp_driver *drv,
 	if (!sad->model_handler)
 		goto sad_fail;
 
-	k_mutex_init(&sad->lock);
 
 	cfg = (struct sof_smart_amp_config *)ipc_sa->data;
 	bs = ipc_sa->size;
-
 	if (bs > 0 && bs < sizeof(struct sof_smart_amp_config)) {
 		comp_err(dev, "smart_amp_new(): failed to apply config");
 		goto sad_fail;
 	}
-
 	memcpy_s(&sad->config, sizeof(struct sof_smart_amp_config), cfg, bs);
-
 	dev->state = COMP_STATE_READY;
 
 	return dev;
@@ -321,14 +315,11 @@ static int smart_amp_trigger(struct comp_dev *dev, int cmd)
 	switch (cmd) {
 	case COMP_TRIGGER_START:
 	case COMP_TRIGGER_RELEASE:
-		k_mutex_lock(&sad->lock, K_FOREVER);
 		if (sad->feedback_buf) {
 			struct comp_buffer __sparse_cache *buf = buffer_acquire(sad->feedback_buf);
-
 			buffer_zero(buf);
 			buffer_release(buf);
 		}
-		k_mutex_unlock(&sad->lock);
 		break;
 	case COMP_TRIGGER_PAUSE:
 	case COMP_TRIGGER_STOP:
@@ -439,7 +430,6 @@ static int smart_amp_copy(struct comp_dev *dev)
 		audio_stream_avail_frames(&source_buf->stream,
 					  &sink_buf->stream);
 
-	k_mutex_lock(&sad->lock, K_FOREVER);
 	if (sad->feedback_buf) {
 		struct comp_buffer __sparse_cache *buf = buffer_acquire(sad->feedback_buf);
 
@@ -467,7 +457,6 @@ static int smart_amp_copy(struct comp_dev *dev)
 
 		buffer_release(buf);
 	}
-	k_mutex_unlock(&sad->lock);
 
 	if (!avail_frames)
 		avail_frames = avail_passthrough_frames;
@@ -545,7 +534,6 @@ static int smart_amp_prepare(struct comp_dev *dev)
 	buffer_c = buffer_acquire(sad->source_buf);
 	sad->in_channels = audio_stream_get_channels(&buffer_c->stream);
 
-	k_mutex_lock(&sad->lock, K_FOREVER);
 	if (sad->feedback_buf) {
 		struct comp_buffer __sparse_cache *buf = buffer_acquire(sad->feedback_buf);
 
@@ -553,7 +541,6 @@ static int smart_amp_prepare(struct comp_dev *dev)
 		buf->stream.rate = audio_stream_get_rate(&buffer_c->stream);
 		buffer_release(buf);
 	}
-	k_mutex_unlock(&sad->lock);
 
 	sad->process = get_smart_amp_process(dev, buffer_c);
 	if (!sad->process) {
--- a/src/samples/audio/smart_amp_test_ipc3.c
+++ b/src/samples/audio/smart_amp_test_ipc3.c
@@ -16,10 +16,6 @@
 #include <rtos/mutex.h>
 #endif
 
-#if CONFIG_IPC_MAJOR_4
-#include <ipc4/module.h>
-#endif
-
 static const struct comp_driver comp_smart_amp;
 
 LOG_MODULE_REGISTER(smart_amp_test, CONFIG_SOF_LOG_LEVEL);
@@ -31,10 +27,12 @@ DECLARE_SOF_RT_UUID("smart_amp-test", smart_amp_comp_uuid, 0x167a961e, 0x8ae4,
 DECLARE_TR_CTX(smart_amp_comp_tr, SOF_UUID(smart_amp_comp_uuid),
 	       LOG_LEVEL_INFO);
 
+typedef int(*smart_amp_proc)(struct comp_dev *dev,
+			     const struct audio_stream __sparse_cache *source,
+			     const struct audio_stream __sparse_cache *sink, uint32_t frames,
+			     int8_t *chan_map);
+
 struct smart_amp_data {
-#if CONFIG_IPC_MAJOR_4
-	struct sof_smart_amp_ipc4_config ipc4_cfg;
-#endif
 	struct sof_smart_amp_config config;
 	struct comp_data_blob_handler *model_handler;
 	void *data_blob;
@@ -56,12 +54,8 @@ static struct comp_dev *smart_amp_new(const struct comp_driver *drv,
 				      const struct comp_ipc_config *config,
 				      const void *spec)
 {
-#if CONFIG_IPC_MAJOR_3
 	const struct ipc_config_process *ipc_sa = spec;
 	const struct sof_smart_amp_config *cfg;
-#else
-	const struct ipc4_base_module_extended_cfg *base_cfg = spec;
-#endif
 	struct smart_amp_data *sad;
 	struct comp_dev *dev;
 	size_t bs;
@@ -84,32 +78,15 @@ static struct comp_dev *smart_amp_new(const struct comp_driver *drv,
 
 	k_mutex_init(&sad->lock);
 
-#if CONFIG_IPC_MAJOR_4
-	if (base_cfg->base_cfg_ext.nb_input_pins != SMART_AMP_NUM_IN_PINS ||
-	    base_cfg->base_cfg_ext.nb_output_pins != SMART_AMP_NUM_OUT_PINS) {
-		comp_err(dev, "smart_amp_new(): Invalid pin configuration");
-		goto sad_fail;
-	}
-
-	/* Copy the base_cfg */
-	memcpy_s(&sad->ipc4_cfg.base, sizeof(sad->ipc4_cfg.base),
-		 &base_cfg->base_cfg, sizeof(base_cfg->base_cfg));
-
-	/* Copy the pin formats */
-	bs = sizeof(sad->ipc4_cfg.input_pins) + sizeof(sad->ipc4_cfg.output_pin);
-	memcpy_s(sad->ipc4_cfg.input_pins, bs,
-		 base_cfg->base_cfg_ext.pin_formats, bs);
-#else
 	cfg = (struct sof_smart_amp_config *)ipc_sa->data;
 	bs = ipc_sa->size;
 
-	if ((bs > 0) && (bs < sizeof(struct sof_smart_amp_config))) {
+	if (bs > 0 && bs < sizeof(struct sof_smart_amp_config)) {
 		comp_err(dev, "smart_amp_new(): failed to apply config");
 		goto sad_fail;
 	}
 
 	memcpy_s(&sad->config, sizeof(struct sof_smart_amp_config), cfg, bs);
-#endif
 
 	dev->state = COMP_STATE_READY;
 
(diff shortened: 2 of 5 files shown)


RanderWang (commented):
I am not happy with feedback check in every process function.

src/include/sof/audio/buffer.h
```diff
@@ -49,6 +49,10 @@ extern struct tr_ctx buffer_tr;
 /** \brief Retrieves subid (comp id) from the buffer */
 #define trace_buf_get_subid(buf_ptr) ((buf_ptr)->id)
 
+#if CONFIG_IPC_MAJOR_4
+#define SOF_SMART_AMP_FEEDBACK_QUEUE_ID        1
```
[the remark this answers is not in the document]
andrula-song: ah, once I wanted to move the feedback buffer clear operate in smart_amp trigger to module_adpater triger ops, but since you explained it is not necessary in your comment below, so I will move it back to smart_amp header file.

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -0,0 +1,430 @@
...
+		return -EINVAL;
+	}
+}
+
+static int smart_amp_free(struct processing_module *mod)
+{
+	struct smart_amp_data *sad = module_get_private_data(mod);
+	struct comp_dev *dev = mod->dev;
+
+	comp_info(dev, "smart_amp_free()");
```
RanderWang, line 189: switch to comp_dbg to reduce log

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -0,0 +1,430 @@
...
+	output_buffers[0].size += sink_bytes;
+
+	return 0;
+}
+
+static int smart_amp_reset(struct processing_module *mod)
+{
+	struct comp_dev *dev = mod->dev;
+
+	comp_info(dev, "smart_amp_reset()");
```
RanderWang, line 369: comp_dbg

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -0,0 +1,430 @@
...
+
+static inline int smart_amp_get_config(struct processing_module *mod,
+				       uint32_t config_id, uint32_t *data_offset_size,
+				       uint8_t *fragment, size_t fragment_size)
+{
+	struct smart_amp_data *sad = module_get_private_data(mod);
+	struct comp_dev *dev = mod->dev;
+	uint32_t cfg_size = sizeof(struct sof_smart_amp_config);
+
+	comp_info(dev, "smart_amp_get_config()");
```
RanderWang, line 168: comp_dbg

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -0,0 +1,430 @@
...
+
+static int smart_amp_init(struct processing_module *mod)
+{
+	struct smart_amp_data *sad;
+	struct comp_dev *dev = mod->dev;
+	struct module_data *mod_data = &mod->priv;
+	int bs;
+	const struct ipc4_base_module_extended_cfg *base_cfg = mod_data->cfg.init_data;
+
+	comp_info(dev, "smart_amp_init()");
```
RanderWang, line 55: dbg

RanderWang (commented):
Looks pretty good!  some minor changes are recommaned

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -0,0 +1,441 @@
...
+				if (IPC4_SINK_QUEUE_ID(inputs[i].id) ==
+				   SOF_SMART_AMP_FEEDBACK_QUEUE_ID) {
+					sad->feedback_input = &inputs[i];
+
+					/* Set the frame_align info for feedback stream. if we
+					 * execute this part first time, we will get 0 feedback
+					 * frame (by sad->feedback_input->size) in process
+					 * function since the frame_align info were zero while
+					 * calculating in module_adapter aspect.
+					 */
```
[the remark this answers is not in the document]
andrula-song: And I'm thinking is it suitable to do audio_stream_init_alignment_constants(1, 1, stream)(with no limit with both byte and frame) to get default frame_align and frame_align_shift in audio_stream_init().  And we can set the specific  byte_align(as 
 8 for xtensa intrinsic) and frame_align in components individual ops. Then we can reduce the twice available frame calculation in mixin component, one in module_single_sink_setup and the other in mixin_process. And the calculation in mixin_process even use the division one.

andrula-song:
hi @wszypelt 
can you help to check the internal CI  TestCopierIpc::test_01_21_simple_loopback_ipc_ipc --fwlogs memory_window ?
some other PRs failed this case too.
Thanks!

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -137,247 +122,110 @@ static void smart_amp_set_params(struct comp_dev *dev,
...
+	struct comp_dev *dev = mod->dev;
+	struct smart_amp_data *sad = module_get_private_data(mod);
 
-	cfg_size = sizeof(struct sof_smart_amp_config);
-
-	if (cfg_size > *data_size) {
-		comp_err(dev, "smart_amp_get_config(): wrong config size %d",
-			 *data_size);
-	}
+	comp_info(dev, "smart_amp_set_config()");
```
RanderWang, line 138: comp_dbg

wszypelt:
@andrula-song  sure, will look at it

wszypelt:
@andrula-song I tested a test that fails on the main branch, all iterations were successful (100 out of 100), 
in addition, this PR has already been rerun several times and fails in the same place.

andrula-song:
> @andrula-song I tested a test that fails on the main branch, all iterations were successful (100 out of 100), in addition, this PR has already been rerun several times and fails in the same place.

Since before  this time 11794571 pull-7424-merge_1271156 ptlpmssmain03:8845	2023-04-11 15:29:36 I did not see this error,  and I don't think my code would lead to such error. Could you provide the kernel log as well? it seems to be a IPC timeout error.

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -0,0 +1,449 @@
...
+					   avail_frames);
+			fb_bytes = avail_frames *
+				audio_stream_frame_bytes(sad->fb_input->data);
+			buffer_stream_invalidate(fb_buf, fb_bytes);
+
+			sad->process(mod, sad->fb_input, &output_buffers[0],
+				     avail_frames, sad->config.feedback_ch_map);
+			sad->fb_input->consumed = fb_bytes;
+		}
+	}
```
aiChaoSONG, line 358: This could be put in an else clause, to save a true/false test.

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -0,0 +1,449 @@
...
+
+	if (!avail_frames)
+		avail_frames = avail_passthrough_frames;
```
aiChaoSONG, line 361: This could be put in an else clause, to save a boolean test.

src/include/sof/audio/module_adapter/module/module_interface.h
```diff
@@ -51,7 +51,8 @@ struct input_stream_buffer {
 	void __sparse_cache *data; /* data stream buffer */
 	uint32_t size; /* size of data in the buffer */
 	uint32_t consumed; /* number of bytes consumed by the module */
-
+	/* comp_buffer id, add to identify buffer of smart_amp*/
+	uint32_t id;
```
aiChaoSONG, line 55: It is strange that we have an ID field in the structure, but the id is something else's ID. I think it could be documented as input_stream_buffer ID, and it is assigned from comp_buffer, or you can rename it to comp_buff_id.

you can also add queue id, like https://github.com/thesofproject/sof/pull/7310

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -0,0 +1,449 @@
...
+	struct smart_amp_data *sad = module_get_private_data(mod);
+
+	comp_dbg(dev, "smart_amp_set_config()");
+
+	switch (config_id) {
+	case SMART_AMP_SET_MODEL:
+		return comp_data_blob_set(sad->model_handler, pos,
+					 data_offset_size, fragment, fragment_size);
+	case SMART_AMP_SET_CONFIG:
+		{
```
aiChaoSONG, line 146: the braces are redundant?  try to run cppcheck on this file.

src/samples/audio/smart_amp_test_ipc4.c
andrula-song: but add a assignment? Or it seems very strange.

andrula-song:
SOFCI TEST

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -443,178 +286,157 @@ static smart_amp_proc get_smart_amp_process(struct comp_dev *dev,
...
 
-	k_mutex_lock(&sad->lock, K_FOREVER);
-	if (sad->feedback_buf) {
-		struct comp_buffer __sparse_cache *buf = buffer_acquire(sad->feedback_buf);
+	smart_amp_input_setup(mod, input_buffers, num_input_buffers);
+	avail_passthrough_frames = sad->src_input->size;
+	if (sad->fb_input) {
+		fb_buf = attr_container_of(sad->fb_input->data,
+					   struct comp_buffer __sparse_cache,
+					   stream, __sparse_cache);
```
lyakh, line 336: no. Take a normal container, then acquire it.

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -443,178 +286,157 @@ static smart_amp_proc get_smart_amp_process(struct comp_dev *dev,
...
-	comp_dbg(dev, "smart_amp_copy()");
-
-	/* available bytes and samples calculation */
-	avail_passthrough_frames =
-		audio_stream_avail_frames(&source_buf->stream,
-					  &sink_buf->stream);
+	uint32_t fb_bytes;
+	uint32_t source_bytes;
 
-	k_mutex_lock(&sad->lock, K_FOREVER);
```
lyakh: why do you think you don't need locking any more? The module adapter API doesn't lock internally

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -443,178 +286,157 @@ static smart_amp_proc get_smart_amp_process(struct comp_dev *dev,
...
-		buffer_release(buf);
 	}
-	k_mutex_unlock(&sad->lock);
 
 	if (!avail_frames)
 		avail_frames = avail_passthrough_frames;
+
+	src_buf = attr_container_of(sad->src_input->data,
+				    struct comp_buffer __sparse_cache,
+				    stream, __sparse_cache);
```
lyakh, line 359: same

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -443,178 +286,157 @@ static smart_amp_proc get_smart_amp_process(struct comp_dev *dev,
...
-
-		buf->stream.channels = sad->config.feedback_channels;
-		buf->stream.rate = buffer_c->stream.rate;
-		buffer_release(buf);
-	}
-	k_mutex_unlock(&sad->lock);
-
+	sad->input_stream_num = 0;
+	sad->fb_input = NULL;
 	sad->process = get_smart_amp_process(dev, buffer_c);
```
lyakh, line 423: you cannot use `buffer_c` after releasing it

lyakh (changes requested):

src/samples/audio/smart_amp_test_ipc4.c
andrula-song: I used to, but this would lead to internal CI test failed.

src/samples/audio/smart_amp_test_ipc4.c
andrula-song: Rander told me the sad->lock is just for bind/unbind, and suggest me remove it .
https://github.com/thesofproject/sof/pull/7424#discussion_r1161513266.

src/samples/audio/smart_amp_test_ipc4.c
andrula-song: we do acquire in module_adapter_simple_copy, is it suitable to acquire again here?

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -443,178 +286,157 @@ static smart_amp_proc get_smart_amp_process(struct comp_dev *dev,
...
+ * (get from module adapter)has changed.
+ */
+static void smart_amp_input_setup(struct processing_module *mod,
+				  struct input_stream_buffer *inputs,
+				  int32_t num_input_buffers)
+{
+	struct smart_amp_data *sad = module_get_private_data(mod);
+	int i;
+
+	if (sad->input_stream_num != num_input_buffers) {
```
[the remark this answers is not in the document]
andrula-song: there will be warning when committing, so I return to the != judging condition.

src/samples/audio/smart_amp_test_ipc4.c
RanderWang: yes, this is lock is used for bind/unbind thread and process thread. Now bind/unbind are removed, all lock reference is used in the same thread so we don't need it now.

keqiaozhang:
SOFCI TEST

src/samples/audio/smart_amp_test_ipc4.c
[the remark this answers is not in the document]
lyakh: are you sure? I don't see why there should be one. Could you please double-check and quote the warning here?

src/samples/audio/smart_amp_test_ipc4.c
lyakh: it's also used in `.prepare()`, isn't it called from the IPC context?

src/samples/audio/smart_amp_test_ipc4.c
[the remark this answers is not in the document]
andrula-song: Sorry, used to with "else", then there is the warning. I was so foolish at that time.

src/include/sof/audio/module_adapter/module/module_interface.h
```diff
@@ -51,7 +51,8 @@ struct input_stream_buffer {
 	void __sparse_cache *data; /* data stream buffer */
 	uint32_t size; /* size of data in the buffer */
 	uint32_t consumed; /* number of bytes consumed by the module */
-
+	/* comp_buffer id, add to identify buffer of smart_amp*/
+	uint32_t comp_buf_id;
```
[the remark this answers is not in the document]
andrula-song: thanks, used to check feedback input every processing.

src/include/sof/audio/module_adapter/module/module_interface.h
```diff
@@ -51,7 +51,6 @@ struct input_stream_buffer {
 	void __sparse_cache *data; /* data stream buffer */
 	uint32_t size; /* size of data in the buffer */
 	uint32_t consumed; /* number of bytes consumed by the module */
-
```
ranj063, line 54: spurious change

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -1,11 +1,12 @@
 // SPDX-License-Identifier: BSD-3-Clause
 //
-// Copyright(c) 2020 Intel Corporation. All rights reserved.
+// Copyright(c) 2023 Intel Corporation. All rights reserved.
```
ranj063, line 3: i dont think we need to modify this

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -447,178 +290,171 @@ static smart_amp_proc get_smart_amp_process(struct comp_dev *dev,
...
+	if (sad->input_stream_num == num_input_buffers)
+		return;
+
+	if (num_input_buffers == 1) {
+		sad->src_input = &inputs[0];
+		sad->src_buf = attr_container_of(inputs[0].data,
+						 struct comp_buffer __sparse_cache,
+						 stream, __sparse_cache);
+		sad->fb_input = NULL;
+		sad->fb_buf = NULL;
```
ranj063, line 314: can we set the sad->input_stream_num at the beginning and return here? Then we can reduce the indentation after removing the else

andrula-song: done.

src/samples/audio/smart_amp_test_ipc4.c
lyakh: @RanderWang ok, so we're relying on IPC serialisation. But then that also holds for `.bind()` and `.unbind()`, including for IPC3, doesn't it? So we can remove it from the original smart-amp test, before this PR? Can we make that a separate commit in this PR at least? That seems important enough to me to be just done as a side effect in one of commits.

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -447,178 +290,171 @@ static smart_amp_proc get_smart_amp_process(struct comp_dev *dev,
...
+	sad->input_stream_num = num_input_buffers;
 
-	k_mutex_lock(&sad->lock, K_FOREVER);
-	if (sad->feedback_buf) {
-		struct comp_buffer __sparse_cache *buf = buffer_acquire(sad->feedback_buf);
+	if (num_input_buffers == 1) {
+		sad->src_input = &inputs[0];
+		sad->src_buf = attr_container_of(inputs[0].data,
+						 struct comp_buffer __sparse_cache,
+						 stream, __sparse_cache);
```
lyakh, line 314: no, we shouldn't even store acquired buffer pointers in any persistent objects. You shouldn't keep and reuse pointers to acquired buffers between calls to `.process()`. `.process()` is called every time with pointers to buffers, they should be used, not something stored from previous calls. This really should be redesigned.

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -447,178 +290,171 @@ static smart_amp_proc get_smart_amp_process(struct comp_dev *dev,
...
+	uint32_t sink_bytes;
+	uint32_t fb_bytes;
+	uint32_t source_bytes;
 
-			comp_dbg(dev, "smart_amp_copy(): processing %d feedback frames (avail_passthrough_frames: %d)",
-				 avail_frames, avail_passthrough_frames);
+	smart_amp_input_setup(mod, input_buffers, num_input_buffers);
+	avail_passthrough_frames = sad->src_input->size;
+	if (sad->fb_input) {
+		fb_buf_c = buffer_acquire(sad->fb_buf);
```
lyakh, line 354: some confusion around buffer handling, please see comment above.

src/samples/audio/smart_amp_test_ipc4.c
ranj063: I see no need to store these buffer pointers at all anymore if we're going to look them up during every process invocation

andrula-song: once I looked up this buffers every process, but @RanderWang suggested me to have optimization to check only when input stream numbers changed. And this can improve the performance of the hot code.

ranj063: I think its best to just look it up every time @andrula-song. The acquired pointer will be released by the module adapter after the process callback is done. It does not makes sense to hold to the pointer.

src/samples/audio/smart_amp_test_ipc4.c
RanderWang: @lyakh  No, .bind() and .unbind() is not dependent on IPC serialisation. Since .bind & .unbind are triggered by another stream. No such issue on IPC3 since it requires that feedback stream should be started before smart amp stream please check smart_amp_prepare

src/samples/audio/smart_amp_test_ipc4.c
andrula-song: Yes your are right, save these pointers here is not right. But look up every time really costs many cycles and make the performance worse, maybe I can find a better way to  avoid this check later.

lyakh: you can store unacquired pointers and re-acquire them. Acquiring them is cheap when there's no sharing - it's just an uncached-to-cached conversion, and that is basically just an OR or an AND and an OR

andrula-song:
SOFCI TEST

andrula-song:
SOFCI TEST

keqiaozhang:
SOFCI TEST

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -141,247 +124,105 @@ static void smart_amp_set_params(struct comp_dev *dev,
...
-	switch (param_id) {
-	case SMART_AMP_GET_CONFIG:
-		return smart_amp_get_config(dev, data, data_offset);
+		if (fragment_size != sizeof(struct sof_smart_amp_config)) {
+			comp_err(dev, "smart_amp_set_config(): invalid config size %u, expect %u",
+				 fragment_size, sizeof(struct sof_smart_amp_config));
+			return -EINVAL;
+		}
+		comp_dbg(dev, "smart_amp_set_config(): config size = %u", fragment_size);
+		memcpy_s(&sad->config, fragment_size, fragment, fragment_size);
```
lyakh, line 153: the first size should be the size of the destination, i.e. `memcpy_s(&sad->config, sizeof(sad->config), fragment, fragment_size);`

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -447,178 +288,163 @@ static smart_amp_proc get_smart_amp_process(struct comp_dev *dev,
...
+		buf = attr_container_of(inputs[i].data,
+					struct comp_buffer __sparse_cache,
+					stream, __sparse_cache);
 
-			avail_frames = MIN(avail_passthrough_frames,
-					   avail_feedback_frames);
+		if (IPC4_SINK_QUEUE_ID(buf->id) == SOF_SMART_AMP_FEEDBACK_QUEUE_ID)
+			sad->fb_input = &inputs[i];
+		else
+			sad->src_input = &inputs[i];
```
lyakh, line 321: We still disagree on this: I think it's better not to store pointers to acquired buffers in persistent objects, instead it isn't too large of an overhead to run the identification on each copy. @andrula-song thinks it's too much of an overhead and since we do own buffers where we use them, it's ok to also store them persistently. I agree, that in practice there's no difference - as long as we don't "accidentally" access those pointers outside of the `.process()` method, the damage is "only theoretical." Also this is "just a sample," not the real production code. So, if others agree, that this is ok, we can keep it this way. At least I'd put a comment here.

ranj063: @lyakh while this is sample code, the real smart amp implementation will have to follow this lead as well. I am also worried that the production code might even involve multiple cores. Not sure if it will make the situation worse or not but I tihnk it's best to avoid it. I have an alternate suggestion. We do have the bind/unbind ops in the module adapter where we save the sources in a list. In a way this list in ordered in terms of the sources getting added as the modules are bound. So in the case of the smart amp, the source component for the playback path will always be at the 0th index and the feedback comp if available will always be at 1. Can we make use of this info to retrieve the source buffer from the 0th index and feedback buffer from the first index during process? And if w used the module_source_info_acquire/module_source_info_release() in the smart amp module, we'd also avoid coherency issues. what do you think?

src/samples/audio/smart_amp_test_ipc4.c
andrula-song: since we treat it as an error if fragment_size !=sizeof(sad->config) , so actually if we execute to here, the fragment_size should equal sizeof(sad->config). But it is okay to keep your format for standardization.

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -27,101 +26,85 @@ DECLARE_SOF_RT_UUID("smart_amp-test", smart_amp_comp_uuid, 0x167a961e, 0x8ae4,
...
 	struct sof_smart_amp_config config;
 	struct comp_data_blob_handler *model_handler;
 	void *data_blob;
 	size_t data_blob_size;
-
-	struct comp_buffer *source_buf; /**< stream source buffer */
-	struct comp_buffer *feedback_buf; /**< feedback source buffer */
-	struct comp_buffer *sink_buf; /**< sink buffer */
-
-	struct k_mutex lock; /**< protect feedback_buf updated */
```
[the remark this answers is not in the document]
andrula-song: I'd like to add commit message to explain why I drop it.

keqiaozhang:
SOFCI TEST

andrula-song:
SOFCI TEST

src/samples/audio/smart_amp_test_ipc4.c
lyakh: @ranj063 I would be cautious using the module binding order. Why is it fixed? Can it change? We were thinking about a softer approach - during initialisation (binding or whatever step is suitable) check which index corresponds to the source (or to the feedback) buffer and record the index. But even about that I cannot say how reliable it would be. Can one of them (e.g. index 0) be unbound at run-time and then re-bound this time as index 0?

andrula-song:
SOFCI TEST

src/samples/audio/smart_amp_test_ipc4.c
lyakh: @ranj063 ok, so you're saying, that that's already used there. Ok, then we won't break it further if we use that information again. But let's make it explicit somewhere - put a check that that's indeed the case somewhere to be run where buffers can potentially be reconfigured and a comment, can that be done?

andrula-song: @ranj063  @lyakh 
@lyakh if the order is not fixed why is it that when num_input_buffers == 0, that it is always the src buffer and not the feedback buffer?
 >> in my opinion, if num_input_buffers == 0, that it is always the src  because the smart_amp is in playback pipeline. if there is only feedback (capture pipeline), we even never create smart_amp this component.

andrula-song: @ranj063 Actually the last push of my code already used module_source_info_acquire/release to protect the bind/unbind interrupt during smart_amp processing(though I met a exception, but I agree to use the acquire/release to keep it safer). But what's your thought about getting the source and feedback buffers from the module source info array? Is it to use the ipc_config.type? Why don't we use the buffer id since we have this info in IPC4. It seems very tricky to determinate the input by info outside of smart_amp itself.

ranj063: @lyakh i just had a chat with @andrula-song  and what she's storing in sad src_input and fb_input is not a pointer to the acquired buffer or the streams but rather the pointer to the input_buffers in struct processing_module. So this part is actually fine.

ranj063 (approved):

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -0,0 +1,438 @@
...
+
+	sink_buffer = list_first_item(&dev->bsink_list, struct comp_buffer, source_list);
+	buffer_c = buffer_acquire(sink_buffer);
+	sad->out_channels = audio_stream_get_channels(&buffer_c->stream);
+	audio_stream_init_alignment_constants(1, 1, &buffer_c->stream);
+	sad->process = get_smart_amp_process(dev, buffer_c);
+	buffer_release(buffer_c);
+
+	sad->input_stream_num = 0;
+	sad->fb_input = NULL;
```
RanderWang, line 419: in prepare function input_stream_num & fb_input can be setup (1) if feedback stream is started before smart amp, input_stream_num = 2 , fb_input = feedback  or (2) input_stream_num = 1, fb_input = NULL. Please check IPC3 path, they are set up here.

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -0,0 +1,438 @@
...
+	uint32_t cfg_size = sizeof(struct sof_smart_amp_config);
+
+	comp_dbg(dev, "smart_amp_get_config()");
+
+	switch (config_id) {
+	case SMART_AMP_GET_CONFIG:
+		ret = memcpy_s(fragment, fragment_size, &sad->config, cfg_size);
+		if (ret)
+			comp_err(dev, "smart_amp_get_config(): wrong config size %d",
+				 fragment_size);
```
RanderWang, line 176: if error happen,  no need to do "*data_offset_size = cfg_size;"

RanderWang (approved):
LGTM.  two minor issues and you can refine it in another PR if you like my idea

src/samples/audio/smart_amp_test_ipc4.c
lyakh: > Actually the last push of my code already used module_source_info_acquire/release to protect the bind/unbind interrupt during smart_amp processing

just a note: `module_source_info_acquire()` doesn't actually protect anything unless multiple cores are used for some of the modules inputs or outputs, i.e. in most cases it simply converts an uncached address to a cached one. Also note, that objects, used with the coherent API should use that API consistently for all accesses to the data. Mixing cached and uncached access to the data (i.e. with and without `module_source_info_acquire()`) will cause data corruption

lyakh: > @lyakh i just had a chat with @andrula-song and what she's storing in sad src_input and fb_input is not a pointer to the acquired buffer or the streams but rather the pointer to the input_buffers in struct processing_module. So this part is actually fine.

@ranj063 yes, currently those `input_buffers[]` are persistent but once we fix and get something like #7504 in, those will be on-stack objects, so storing them persistently will be a bug.

src/samples/audio/smart_amp_test_ipc4.c
andrula-song: from the original code and my test, seems while prepare we always only get the source stream. I 'd like to keep my code now to avoid the accident cases if we keep playback path and start-> stop -> start capture path.

src/samples/audio/smart_amp_test_ipc4.c
andrula-song: ok, let's look up every processing. Now I know why you ask for such change, thanks.

src/samples/audio/smart_amp_test_ipc4.c
andrula-song: now we should look up every process since the input buffer will not be persistent later soon.

kv2019i:
@lyakh can you check?

andrula-song:
SOFCI TEST

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -28,108 +25,94 @@ DECLARE_SOF_RT_UUID("smart_amp-test", smart_amp_comp_uuid, 0x167a961e, 0x8ae4,
...
-
-	struct k_mutex lock; /**< protect feedback_buf updated */
-
+	struct input_stream_buffer *src_input; /* source input stream buffer*/
+	struct input_stream_buffer *fb_input; /* feedback input stream buffer*/
 	smart_amp_proc process;
-
-	uint32_t in_channels;
 	uint32_t out_channels;
+	uint32_t input_stream_num;
```
lyakh, line 42: also not needed?

lyakh (commented):
Looks like we're getting very close, let's just remove these left-overs and maybe that will be the last revision

andrula-song:
SOFCI TEST

lyakh (approved):

kv2019i:
So ready to go, but needs to pass the CI checks.

src/samples/audio/CMakeLists.txt
```diff
@@ -1,9 +1,17 @@
...
-		smart_amp_test.c
-	)
+if(CONFIG_IPC_MAJOR_3)
+	if(CONFIG_SAMPLE_SMART_AMP)
+		add_local_sources(sof
+			smart_amp_test_ipc3.c
+		)
+	endif()
+elseif(CONFIG_IPC_MAJOR_4)
+	if(CONFIG_SAMPLE_SMART_AMP)
```
marc-hb, line 10: Single `if(CONFIG_SAMPLE_SMART_AMP)` on top.

src/samples/audio/smart_amp_test_ipc3.c
```diff
@@ -545,18 +310,17 @@ static int smart_amp_trigger(struct comp_dev *dev, int cmd)
...
-		ret = PPL_STATUS_PATH_STOP;
+		return PPL_STATUS_PATH_STOP;
 
 	switch (cmd) {
 	case COMP_TRIGGER_START:
 	case COMP_TRIGGER_RELEASE:
-		k_mutex_lock(&sad->lock, K_FOREVER);
 		if (sad->feedback_buf) {
 			struct comp_buffer __sparse_cache *buf = buffer_acquire(sad->feedback_buf);
+
```
marc-hb, line 320: Avoid unneeded git conflicts

kv2019i:
@andrula-song If you can refresh this PR to make the Intel QB tests pass, we should be good to go.

andrula-song:
hi @wszypelt  can you help to trigger again? TGL  TestLoadFwBasic failed.

kv2019i:
PM fail in https://sof-ci.01.org/sofpr/PR7424/build7101/devicetest/index.html -> not related.
And PM fail on one device in https://sof-ci.01.org/sofpr/PR7424/build7102/devicetest/index.html -> not related
Proceeding with merge.

Issue #7192, opened 2023-03-01
[FEATURE] Make SOF with Zephyr run on i.MX93's A55 core with Jailhouse


LaurentiuM1234:
**PURPOSE**
I've decided to open this issue in order to track the current progress and interact with the community. Our plan is to make SOF with Zephyr run on i.MX93's cortex-A55 core with the help of Jailhouse. Jailhouse will be used to split the hardware resources between Linux and Zephyr which will be running independently on different cores.

Feedback and suggestions would be very much appreciated.

**ISSUES**

1. Currently, using `scripts/xtensa-build-zephyr`, I have managed to obtain the `zephyr.elf`. The script fails when trying to run `smex` in order to build the `.ldc` file. My current understanding is that `smex` isn't able to properly parse the `zephyr.elf` file because of the fact that there's currently no support for parsing `ELF-64` files. My question regarding this would be: would it be a good idea to add support in `smex` for parsing such files? Also, would it be necessary since, for now, my plan for now is to use Zephyr's logging system (no trace, the logs will be printed via the UART)?
2. Regarding `zephyr.elf`'s sections: are there any restrictions imposed by SOF (e.g: alignment, order)? For now, I was able to get away by simply adding the `.static_uuids`, `.static_log` and `.trace_ctx` to the already present sections added for ARM64 platforms on Zephyr.
3. Would going through `smex` and the signing process still be required if the plan is to load the firmware image through Jailhouse? Please correct me if I'm wrong but my current understanding is that the .ri file resulting from the SOF build process contains some sort of header which is "stripped" during the load process in the linux kernel. Because of this, since Jailhouse loads the firmware image as-is, the A55 core would probably not be able to work properly. Would it be acceptable to bypass the .ri file creation process and simply utilize the generated `zephyr.bin`? (which can be generated by enabling `CONFIG_BUILD_OUTPUT_BIN`)

**RELATED PRs**
TODO

CC: @dbaluta @iuliana-prodan @paulstelian97 @lgirdwood

lgirdwood:
@LaurentiuM1234 what is jailhouse ? any links ?
Adding @nashif @mwasko @andyross

LaurentiuM1234:
@lgirdwood Sure, for a more detailed overview please see [1] and [2]. For a more general overview please see [3]. Also, please see [4] for github repo.

[1]: https://lwn.net/Articles/578295/
[2]: https://lwn.net/Articles/578852/
[...]

Issue #7248, opened 2023-03-10
[FEATURE][Zephyr] Clarify expectations of platform.h for Zephyr SOF builds


kv2019i:
Levelling up from https://github.com/thesofproject/sof/issues/5602

**Is your feature request related to a problem? Please describe.**
The platform.h has three interfaces:

int platform_boot_complete(uint32_t boot_message);
int platform_boot_complete(uint32_t boot_message);
int platform_context_save(struct sof *sof);

Plus a set of capability definitions used by application code (like PLATFORM_MAX_CHANNELS). These need to be defined by each platform via platform.h.

Documented here: https://github.com/thesofproject/sof/blob/main/src/include/sof/platform.h

Now we have three types of platforms implemented in SOF main (2023/March situation):
  - native Zephyr, can only be built with Zephyr (e.g. platform/intel/ace)
  - native XTOS, can be only built with XTOS (e.g. platform/mtk8186)
  - hybrid that can be built with either Zephyr or XTOS (e.g. platform/intel/cavs and platform/imx8)

**Describe the solution you'd like**
Clear definition of semantics of platform.h expected for native Zephyr platforms. Based on experience with intel/ace , there is still need for a thin platform layer. 

As XTOS does not have its own platform layer, SOF platform.h will need to continue to serve as the platform layer as long as SOF main supports XTOS builds.

**Describe alternatives you've considered**
- removal of platform layer completely and make it XTOS only interface
       - this is still a possibility but hard to do unless all platforms move to native Zephyr 
       - a new solution would be needed for e.g. intel/ace/platform.c 
- split out "platfrom-capability.h" that only exports platform.h interfaces needed by src/audio/
       - on Zephyr this can be a thin mapping to DTS information (and possibly shared across platforms)
       - this would allow to clean up code for hybrid platforms that continue to support both Zephyr and XTOS builds (versus the current ifdef ZEPHYR mess we have now in some of the platform files)
        
**Additional context**

kv2019i:
FYI @thesofproject/maintainers @andyross @lenghonglin @marcinszkudlinski @mwasko @andrey @dbaluta @LaurentiuM1234 @paulstelian97

dbaluta:
> int platform_boot_complete(uint32_t boot_message);
> int platform_boot_complete(uint32_t boot_message);

Do you mean `int platform_init(struct sof *sof)` instead of one of `platform_boot_complete`?

kv2019i:
[...]

PR #7425, opened 2023-04-10
Introduce i.MX93 platform support

LaurentiuM1234:
This PR introduces all the necessary changes for running SOF using Zephyr and Jailhouse on i.MX93's A55 core.

Will update  https://github.com/thesofproject/sof/issues/7192 to contain a short description of our solution and all of its quirks. (**EDIT**: this is now up)

**EDIT**: Depends on Zephyr PR: https://github.com/zephyrproject-rtos/zephyr/pull/56713

### DIFF
--- a/src/platform/Kconfig
+++ b/src/platform/Kconfig
@@ -125,6 +125,15 @@ config IMX8ULP
 	 Select if your target platform is imx8ulp-compatible.
 	 imx.8ulp support dsp.
 
+config IMX93_A55
+	bool "Build for NXP i.MX93 arm64 architecture (Zephyr-only)"
+	select BUILD_OUTPUT_BIN
+	select ZEPHYR_LOG
+	select HOST_PTABLE
+	select DMA_DOMAIN
+	help
+	 Select if your target platform is imx93-compatible.
+
 config RENOIR
 	bool "Build for Renoir"
 	select XT_INTERRUPT_LEVEL_5
--- a/src/platform/imx93_a55/include/platform/drivers/idc.h
+++ b/src/platform/imx93_a55/include/platform/drivers/idc.h
@@ -1,0 +1,28 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright 2023 NXP
+ */
+
+#ifdef __ZEPHYR_RTOS_IDC_H__
+
+#ifndef __PLATFORM_DRIVERS_IDC_H__
+#define __PLATFORM_DRIVERS_IDC_H__
+
+#include <stdint.h>
+
+/* TODO: remove me if possible */
+
+struct idc_msg;
+
+static inline int idc_send_msg(struct idc_msg *msg, uint32_t mode)
+{
+	return 0;
+}
+
+#endif /* __PLATFORM_DRIVERS_IDC_H__ */
+
+#else
+
+#error "This file shouldn't be included from outside of Zephyr's rtos/idc.h"
+
+#endif /* __ZEPHYR_RTOS_IDC_H__ */
--- a/src/platform/imx93_a55/include/platform/lib/memory.h
+++ b/src/platform/imx93_a55/include/platform/lib/memory.h
@@ -1,0 +1,72 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright 2023 NXP
+ */
+
+#ifdef __SOF_LIB_MEMORY_H__
+
+#ifndef __PLATFORM_LIB_MEMORY_H__
+#define __PLATFORM_LIB_MEMORY_H__
+
+#include <rtos/cache.h>
+
+#define PLATFORM_DCACHE_ALIGN DCACHE_LINE_SIZE
+
+/* seems like in xtensa architecture cached data
+ * may be placed at some special addresses in the
+ * SRAM?
+ *
+ * since we're running on A55 cores, there's no such
+ * thing so all the below cache/shared data "management"
+ * functions aren't necessary.
+ */
+#define SHARED_DATA
+
+#define uncache_to_cache(address) address
+#define cache_to_uncache(address) address
+#define cache_to_uncache_init(address) address
+#define is_uncached(address) 0
+
+/* number of heaps to be used.
+ *
+ * i.MX93 is non-SMP so we're only going to use 1 heap.
+ */
+#define PLATFORM_HEAP_SYSTEM 1
+#define PLATFORM_HEAP_SYSTEM_RUNTIME 1
+#define PLATFORM_HEAP_RUNTIME 1
+#define PLATFORM_HEAP_BUFFER 1
+
+/* host and firmware use shared memory */
+#define host_to_local(addr) (addr)
+#define local_to_host(addr) (addr)
+
+/* if this gets too big then it will be
+ * necessary to increase the SRAM size from
+ * the Zephyr DTS. This should be fine as
+ * we're going to reserve 800MB for Jailhouse
+ * usage.
+ */
+#define HEAPMEM_SIZE 0x00010000
+
+/* SOF uses A side of the WAKEUPMIX MU */
+#define MU_BASE 0x42430000
+
+/* SOF uses EDMA2 (a.k.a EDMA4 in the TRM) */
+#define EDMA2_BASE 0x42010000
+#define EDMA2_CHAN_SIZE 0x8000
+
+/* WM8962 is connected to SAI3 */
+#define SAI3_BASE 0x42660000
+
+static inline void *platform_shared_get(void *ptr, int bytes)
+{
+	return ptr;
+}
+
+#endif /* __PLATFORM_LIB_MEMORY_H__ */
+
+#else
+
+#error "This file shouldn't be included from outside of sof/lib/memory.h"
+
+#endif /* __SOF_LIB_MEMORY_H__*/
--- a/src/platform/imx93_a55/lib/clk.c
+++ b/src/platform/imx93_a55/lib/clk.c
@@ -1,0 +1,46 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/*
+ * Copyright 2023 NXP
+ */
+
+#include <rtos/clk.h>
+#include <sof/lib/notifier.h>
+
+static const struct freq_table platform_cpu_freq[] = {
+	{ CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC,
+		CONFIG_SYS_CLOCK_TICKS_PER_SEC * 1000 },
+};
+
+static struct clock_info platform_clocks_info[NUM_CLOCKS];
+
+void platform_clock_init(struct sof *sof)
+{
+	int i;
+
+	sof->clocks = platform_clocks_info;
+
+	/* The CCM doesn't seem to allow setting a core's
+	 * frequency. It probably sets the whole cluster's
+	 * frequency to some value (not relevant). Since
+	 * we're running on top of Jailhouse we don't want
+	 * to allow SOF to change the cluster's frequency
+	 * since that would also affect Linux.
+	 *
+	 * Also, as a consequence to this, on SMP systems,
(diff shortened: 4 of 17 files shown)


src/platform/Kconfig
```diff
@@ -125,6 +125,16 @@ config IMX8ULP
...
 
+config IMX93_A55
+	bool "Build for NXP i.MX93 arm64 architecture (Zephyr-only)"
+	select BUILD_OUTPUT_BIN
+	select ZEPHYR_LOG
+	select HOST_PTABLE
+	select DMA_DOMAIN
+	select IMX_ARM64
+	help
+	 Select if your target platform is imx93-compatible.
```
paulstelian97, line 136: Commit order -- this config should be introduced _after_ the code it depends on.

src/platform/imx93_a55/include/platform/lib/memory.h
```diff
@@ -0,0 +1,72 @@
...
+#ifndef __PLATFORM_LIB_MEMORY_H__
+#define __PLATFORM_LIB_MEMORY_H__
+
+#include <rtos/cache.h>
+
+#define PLATFORM_DCACHE_ALIGN DCACHE_LINE_SIZE
+
+/* seems like in xtensa architecture cached data
+ * may be placed at some special addresses in the
+ * SRAM?
```
paulstelian97, line 17: This thing is specific to Intel platforms where the memory map can have distinct regions pointing to the same memory, which can have different caching settings. None of the i.MX platforms have this ability.

Thankfully the current code works correctly despite the hack, but ONLY because SOF runs on a single core. If it ever ends up running on multiple cores this must be fixed somehow.

src/platform/imx93_a55/lib/dai.c
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+#include <rtos/sof.h>
+#include <sof/drivers/sai.h>
+#include <sof/drivers/edma.h>
+#include <sof/lib/memory.h>
+
+static struct dai sai[] = {
+	{
+		.index = 3, /* SOF uses SAI3 */
```
paulstelian97, line 13: This index is slightly weird -- other i.MX platforms just use indices 0 and 1 despite it being like ESAI1 and SAI2 or similar.

You _can_ set it as 3 if you believe this is better. What's important is this index must match a value in the topology (so once this is merged you cannot really ever change it).

paulstelian97 (commented):
Nice work, there's a few comments (the commit order one you need to deal with, the others are discussions/opinions).

And while it does look sketchy in some places, some of these things can be fixed later on too. I tried to point out what needs the attention right now.

Do we have a proof of concept (this actually working) as of this time?

src/platform/Kconfig
LaurentiuM1234: Shouldn't it be the other way around though? Seems kinda odd to use a config which has yet to be introduced. Either way is fine to me as the build will fail anyways without all the commits in this PR.

src/platform/imx93_a55/lib/dai.c
LaurentiuM1234: The only reason I set it to 3 is because at the time I believed it needs to match the SAI index which, according to what you just said, is not the case. I'd say leave this as-if for now as it makes more sense (at least to me) for it to match the SAI index. There's no problem with the topology so it should be fine.

LaurentiuM1234:
@paulstelian97 Thanks for the feedback !

Regarding this:
> Do we have a proof of concept (this actually working) as of this time?

I've made a status update here: https://github.com/thesofproject/sof/issues/7192. Please check out my last comment, section **STATUS**. TL; DR: Yup this works.

src/platform/Kconfig
paulstelian97: Ah fair enough. What truly matters is that we have no commit where the build fails (eventually commits where a config can never be enabled are good). So this may be fine to dismiss.

paulstelian97:
The other comment where you should never have a commit where build breaks with any of the defconfigs still applies. You don't want to break bisection!

LaurentiuM1234:
> The other comment where you should never have a commit where build breaks with any of the defconfigs still applies. You don't want to break bisection!

Thanks for pointing this out. Just tried building SOF with Zephyr for imx8 and tgl while incrementally adding all the patches here. Seems fine.

zephyr/include/rtos/cache.h
```diff
@@ -8,12 +8,59 @@
 
 /* TODO: align with Zephyr generic cache API when ready */
 #define __SOF_LIB_CACHE_H__
+
+#ifndef CONFIG_IMX_ARM64
 #include <arch/lib/cache.h>
+#else
```
dbaluta, line 14: This is not imx specific but arm64 specific. So, I wonder if we could only introduce CONFIG_ARM64 and somehow figure out that we are on imx arm64 by checking for CONFIG_IMX && CONFIG_ARM64?

lgirdwood: Fyi @nashif has a Zephyr Cache API in the works that will abstact the arch cache ops with a generic API..

lgirdwood:
@LaurentiuM1234 might be worths splitting this PR out a little, per feature for easier review.

src/platform/imx93_a55/include/platform/drivers/idc.h
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+#ifdef __ZEPHYR_RTOS_IDC_H__
+
+#ifndef __PLATFORM_DRIVERS_IDC_H__
+#define __PLATFORM_DRIVERS_IDC_H__
+
+#include <stdint.h>
+
+/* TODO: remove me if possible */
```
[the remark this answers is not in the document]
LaurentiuM1234: It's referring to the whole file. Same goes for `interrupt.h`, `dai.h` and `trace.h`. Apart from the fact that they help silence compilation errors they serve no purpose in i.MX93's case ATM.

LaurentiuM1234:
> @LaurentiuM1234 might be worths splitting this PR out a little, per feature for easier review.

Agreed. With the removal of `CONFIG_IMX_ARM64`, I can definitely see some changes worth submitting in other PRs (e.g: the cache API-related one).

src/ipc/ipc3/handler.c
```diff
@@ -1663,6 +1663,32 @@ void ipc_cmd(struct ipc_cmd_hdr *_hdr)
...
+		 *	message, the window regions and the reply
+		 *	header in the following order:
+		 *		1) reply structure
+		 *		2) sof_ipc_fw_ready structure
+		 *		3) windows structure
+		 *	(all of the above information is written
+		 *	contiguously in the hostbox)
+		 */
+
+		ret = platform_boot_complete(0);
```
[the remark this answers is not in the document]
LaurentiuM1234: ACK. Should also probably introduce a new function in `wrapper.c`. The ifdefs will still have to be used in said functions.

LaurentiuM1234:
**UPDATE**
Now that I've broken down this PR into several PRs and the Zephyr PR has already been merged will mark this PR as ready for review. Also, this PR now only introduces platform support for i.MX93.

src/platform/imx93_a55/include/platform/drivers/idc.h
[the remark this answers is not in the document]
paulstelian97: The idc file specifically shouldn't really have been needed for _any_ of the i.MX platforms, as IDC is Intel specific, and future multi-core DSPs from others will likely use different inter-core communication standards.

A future refactor should make sure to remove this file being needed from all non-CAVS platforms.

src/platform/imx93_a55/include/platform/lib/memory.h
```diff
@@ -0,0 +1,72 @@
...
+#define is_uncached(address) 0
+
+/* number of heaps to be used.
+ *
+ * i.MX93 is non-SMP so we're only going to use 1 heap.
+ */
+#define PLATFORM_HEAP_SYSTEM 1
+#define PLATFORM_HEAP_SYSTEM_RUNTIME 1
+#define PLATFORM_HEAP_RUNTIME 1
+#define PLATFORM_HEAP_BUFFER 1
```
paulstelian97, line 37: I don't think you can share system vs runtime/buffer heaps can you? Or is my knowledge of them being structurally different (and thus mandatory to separate) only applicable on XTOS builds?

On XTOS, for reference, the system heap works very differently from the others (allocation is a pointer increment, as opposed to using a structure to keep track of allocated vs freed heap slots)

src/platform/imx93_a55/lib/clk.c
```diff
@@ -0,0 +1,46 @@
...
+	int i;
+
+	sof->clocks = platform_clocks_info;
+
+	/* The CCM doesn't seem to allow setting a core's
+	 * frequency. It probably sets the whole cluster's
+	 * frequency to some value (not relevant). Since
+	 * we're running on top of Jailhouse we don't want
+	 * to allow SOF to change the cluster's frequency
+	 * since that would also affect Linux.
```
paulstelian97, line 27: This is an interesting discussion to have -- how will Linux's own core frequency change affect Zephyr? Can it possibly mess with the playback in audible ways? Make it harder for the core to keep up if e.g. compress playback is happening?

paulstelian97 (approved):
No complaints (thus approve), but a little bit of discussion.

src/platform/imx93_a55/include/platform/lib/memory.h
kv2019i: Currently these are mapped to a single heap in Zephyr SOF builds and the PLATFORM_HEAP_FOO definitions are implemented. Instead, we have a single heap implemented with sys_heap and the SOF heap type definitions are only used to control alloc time behaviour (and we maintain existing semantics here, e.g. SYSTEM allocs can never fail). 

There is lot of activity in this area though. On Intel side, we recently added support for a separately managed persistant memory heap (SOF_MEM_CAPS_L3), and there's been active discussions about adding infra to support per-core heaps. But no PRs yet how to expose this via memory.h, so for now, we stick to implementing existing semantics to SOF app code. Platform code needs to provide necessary definitions both for XTOS and Zephyr builds.

kv2019i (approved):
Looks good! I think some of this code could be removed if/when we get further with platform interface cleanup -> https://github.com/thesofproject/sof/issues/7248 
... but alas, it's not complete yet, so now there's indeed some muddy waters in terms of what a platform needs to define. 

If you want to minimize this code and only support Zephyr SOF build, you can take example of sof/src/platform/intel/ace/include/ace/ . It's not perfect yet, but it drops a lot of definitions that are not needed in Zephyr builds, resulting in much smaller set of code needed to add a new platform (on SOF side that is, of course the definitions need to be in Zephyr).

src/platform/imx93_a55/lib/clk.c
LaurentiuM1234: Very good question, thanks for posing it. I'm not sure how altering the cluster's frequency from Linux's side would affect SOF's flow as I haven't tested it out but this is definitely something we need to watch out for.

LaurentiuM1234:
> If you want to minimize this code and only support Zephyr SOF build, you can take example of sof/src/platform/intel/ace/include/ace/ . It's not perfect yet, but it drops a lot of definitions that are not needed in Zephyr builds, resulting in much smaller set of code needed to add a new platform (on SOF side that is, of course the definitions need to be in Zephyr).

Was thinking of introducing something similar for i.MX based on used architecture (or maybe even for all used architectures if there's a lot in common) but I think I'll leave this for after the drivers are ported to Zephyr as things may clear out.

lyakh (approved):

iuliana-prodan (approved):

src/platform/imx93_a55/include/platform/drivers/idc.h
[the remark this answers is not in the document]
lgirdwood: ack - cleanup will remove this incrementally.

lgirdwood (approved):

PR #7429, opened 2023-04-11
Audio: module adapter: Fix module_set_configuration

dbaluta:
After commit 7d0a0c673f94 ("Audio: Module adapter: Pass entire cdata to clients for all controls") md->ops->set_configuration gets the entire cdata but module_set_configuration expects only cdata>data[0].data.

So, restore the pointer fragment to cdata->data[0].data.

With IPC4 in comp_data_blob_set() the use of
struct sof_ipc_ctrl_data pointer step is not done but the fragment is treated as raw data payload without the header.


paulstelian97 (approved):
Approve for NXP usage, but I want an approval from other IPC3 platforms (AMD?) and from Intel (IPC4) before feeling fine about this merge.

src/audio/module_adapter/module/generic.c
```diff
@@ -347,15 +347,27 @@ int module_free(struct processing_module *mod)
...
 	size_t offset = 0;
 	uint8_t *dst;
 	int ret;
 
+#if CONFIG_IPC_MAJOR_3
+	if (cdata->cmd != SOF_CTRL_CMD_BINARY) {
+		comp_err(dev, "module_set_configuration(), illegal control command");
+		return -EINVAL;
+	}
+#endif
```
lyakh, line 370: It is confusing indeed. In `module_adapter_cmd()` either
`module_adapter_ctrl_get_set_data(dev, cdata, ...);` or
`md->ops->set_configuration(..., cdata,...);` or
`md->ops->get_configuration(..., cdata,...);`
is called. In `module_adapter_ctrl_get_set_data()` then `module_adapter_get_set_params(dev, cdata, set);` and from it
`md->ops->set_configuration(..., (const uint8_t *)cdata->data[0].data,...);` or
`md->ops->get_configuration(..., (const uint8_t *)cdata->data[0].data,...);`
are called, i.e. indeed `.set_configuration()` and `.get_configuration()` are sometimes called with `cdata` and  sometimes with `cdata->data[0].data` as their argument. Isn't the problem already there?

paulstelian97: Wow indeed, this seems much worse than what comes from a first look. Perhaps a different fix is in order to make everything consistent.

dbaluta: @lyakh wait a second

In `module_adapter_ctrl_get_set_data()` then `module_adapter_get_set_params(dev, cdata, set)`; and from it
`md->ops->set_configuration(..., (const uint8_t *)cdata->data[0].data,...);` or
`md->ops->get_configuration(..., (const uint8_t *)cdata->data[0].data,...);`

this doesn't look true to be on my machine.

I see that in `module_adapter_get_set_params(dev, cdata, set)` we call `md->ops->set_configuration(mod, cdata->data[0].type, pos, data_offset_size, (const uint8_t *)cdata, cdata->num_elems,  NULL, 0);` or 

` md->ops->get_configuration(mod, pos, &data_offset_size, (uint8_t *)cdata, cdata->num_elems);`

lyakh: > @lyakh wait a second

@dbaluta ah, you're right, my local tree wasn't updated yet after my vacation, it was still missing 7d0a0c673f94fe59fbe9b7586f841434ac2569b4 which you're actually referencing in your commit description. So, you're right, my mistake!

lyakh (approved):

src/audio/module_adapter/module/generic.c
```diff
@@ -347,15 +347,27 @@ int module_free(struct processing_module *mod)
...
+	const uint8_t *fragment = fragment_in;
+#endif
 	struct module_data *md = &mod->priv;
 	struct comp_dev *dev = mod->dev;
 	size_t offset = 0;
 	uint8_t *dst;
 	int ret;
 
+#if CONFIG_IPC_MAJOR_3
+	if (cdata->cmd != SOF_CTRL_CMD_BINARY) {
```
lyakh, line 366: but I'm uncertain about this now. 7d0a0c673f94fe59fbe9b7586f841434ac2569b4 says, that data_blob.c only handles binary controls, that's why that check is added there. But here any control type can be handled?

dbaluta: I think you are right. Will remove this.

dbaluta:
@singalsu yes, our tests for Cadence were failing. Thanks for review!

iuliana-prodan (approved):

lyakh (approved):

PR #7432, opened 2023-04-11
 topology2: dmic-generic: add macro for DMIC_CORE_ID

yongzhi1:


juimonen (approved):
fine with me.

kv2019i (approved):

kv2019i:
@yongzhi1 can you refresh to get the CI pass?

@ranj063 @RanderWang good to go?

tools/topology/topology2/platform/intel/dmic-default.conf
```diff
@@ -33,5 +33,5 @@ Define {
 	WOV_UUID			'1f:d5:a8:eb:27:78:b5:47:82:ee:de:6e:77:43:af:67'
 	WOV_CPC				'360000'
 	DMIC1_PCM_CAPS			'Passthrough Capture 18'
-
+	DMIC_CORE_ID			0
```
ranj063, line 36: @yongzhi1 can we simplify this a  bit by adding a core_id attribute in pipeline_common.conf. Then all you need to do is set the core_id for each pipeline. instead of repeating this for all widgets

RanderWang (approved):

tools/topology/topology2/platform/intel/dmic-default.conf
RanderWang: @ranj063 currently we need to set "core_id" for each widget and set "core" for pipeline in Linux kernel driver. widget core id is built from topology except pipeline widget.
````c
/* Component extended tokens */
static const struct sof_topology_token comp_ext_tokens[] = {
	{SOF_TKN_COMP_UUID, SND_SOC_TPLG_TUPLE_TYPE_UUID, get_token_uuid,
		offsetof(struct snd_sof_widget, uuid)},
	{SOF_TKN_COMP_CORE_ID, SND_SOC_TPLG_TUPLE_TYPE_WORD, get_token_u32,
		offsetof(struct snd_sof_widget, core)},
};
````

ranj063: @RanderWang what does that have to do with the simplification that I am asking?

RanderWang: does widgets include pipeline_common.conf ?

ranj063: No they dont. But there's inheritance in topology2. You se it in pipeline-common.conf, it will automatically be passed to all widgets in the pipeline

yongzhi1: Good idea, let me think about it.

ranj063: @yongzhi1 what im aasking for should take less than 5 mins to do. Would you like me to give you a patch? I dont want to merge this only to redo it

ranj063 (approved):

ranj063 (approved):

tools/topology/topology2/platform/intel/dmic-default.conf
ranj063: this needs a cthis likely needs a compiler update and should be done later.

RanderWang: > No they dont. But there's inheritance in topology2. You set it in pipeline-common.conf, it will automatically be passed to all widgets in the pipeline

thanks, a good feature

ranj063:
the usual timeouts in device test failures. merging

PR #7439, opened 2023-04-13
dma: disable hda when pipeline is reset

fkwasowi:
DGCS_GEN bit should be set to 0 on a pipeline state change to reset


src/audio/host-zephyr.c
```diff
@@ -1027,8 +1027,7 @@ static int host_position(struct comp_dev *dev,
 void host_zephyr_reset(struct host_data *hd, uint16_t state)
 {
 	if (hd->chan) {
-		if (state == COMP_STATE_ACTIVE)
-			dma_stop(hd->chan->dma->z_dev, hd->chan->index);
+		dma_stop(hd->chan->dma->z_dev, hd->chan->index);
```
paulstelian97, line 1030: Does this work in all states, including if reset is called twice in a row with no intervening calls? (just a question since I'm not fully familiar with this)

kv2019i: @paulstelian97 I think this is ok, Zephyr dma.h has for stop that "stop is allowed on channels that have already been stopped and must report success. " ... the documentation was recently clarified on Zephyr side as SOF had multiple places where we could potentially have redundant calls. But we save a few bytes and cycles if we don't need to track state in both SOF and Zephyr...

src/audio/host-zephyr.c
```diff
@@ -1027,8 +1027,7 @@ static int host_position(struct comp_dev *dev,
 void host_zephyr_reset(struct host_data *hd, uint16_t state)
```
kv2019i, line 1027: The description is a bit misleading as it assume this code is always used with HD-DMA, while this code is in fact generic and could be used with a wide variety of DMA hardware and drivers.

But this does explain why this is important step to do (and explains what e.g. happens on HD-DMA hardware when not done), so not a blocker.

src/audio/host-zephyr.c
paulstelian97: The question here is that does stop work even after the prior dma_release_channel?

But I think I can answer that myself (yes, because hd->chan will be null and this code path won't run anyway)

paulstelian97 (approved):

fkwasowi:
SOFCI TEST

src/audio/host-zephyr.c
ranj063, line 1030: @fkwasowi what issue does this change fix? AFAICT we also do dma_stop() in host_zephyr_trigger() in the STOP case with the same check for state. While the state may not be active during reset, i'd assume that it must have been active during stop and the dma should be stopped there

fkwasowi: There is a dma_start in host_zephyr_trigger that sets DGCS_GEN to 1, and a dma_stop that sets it to 0. Pipeline reset usually occurs after PAUSE (there is also a COMP_STATE_PREPARE state in between, and it is on this state that we enter the host_zephyr_reset function). I have not seen such a scenario that the RESET pipeline state is executed immediately after the RUN pipeline state (COMP_STATE_ACTIVE is only when we are in the RUN state). Such a sequence is incorrect. This fixes the scenario: https://github.com/intel-innersource/drivers.audio.firmware.converged/issues/505

tmleman (approved):

fkwasowi:
SOFCI TEST

fkwasowi:
SOFCI TEST

makarukp (approved):

dnikodem (approved):

src/audio/host-zephyr.c
lyakh: does this mean, that we're effectively reverting #6849 ? Was it wrong?

plbossart: I am also not following the change nor the failing scenario.

We do clear DGCS_GEN in hda_hda_stop_common, what causes it not to be cleared?

I'll add a "request change" to mark my concern, this DGCS_GEN bit is very very touchy and we have to be super careful with any changes to these sequences.

plbossart (changes requested):

src/audio/host-zephyr.c
fkwasowi: IF was added because in sof the same channel was stopped several times in some scenarios. Because of this, the ref count in the dev pm was going off. But the problem was fixed later on the hda driver side, so now the IF can be removed.

fkwasowi: HDA DMA programming sequence in pipeline flows, coming from closed-source FW (obligatory to work properly with closed-source driver):

Reset (XRUN):

IPC: Set Pipe State(PAUSE): FW: if coming from RUN state, then no DMA programming
SW: Set DMA RUN = 0 and Poll untill it reflects 0
IPC: Set Pipe State(RESET): FW:
a. GEN=0
b. FIFORDY=0
SW: HDA Stream RESET
SW: HDA Stream Configure(BDL, CTL etc.)
SW: start Host DMA
IPC: Set Pipe State(PAUSE): FW: if coming from RESET state then FIFORDY=1
IPC: Set Pipe State (RUN): FW: Set GEN=1

fkwasowi:
SOFCI TEST

fkwasowi:
SOFCI TEST

src/audio/host-zephyr.c
tmleman: > Is it necessary, to undo all your changes [fa0456e](https://github.com/thesofproject/sof/commit/fa0456e5a8996dcdfee97f1e7142891cc3dd90ae) ?

I don't know if it's necessary but it should be possible to do. This has already been partially done in this commit 
 https://github.com/thesofproject/sof/commit/a942f10731b1a72101a6ac4be74de5a33d5c7acd

kv2019i:
@fkwasowi I see you asked for re-review from @plbossart . Can you confirm his concern was addressed (w.r.t. "what causes it not to be cleared?"..?

fkwasowi:
> @fkwasowi I see you asked for re-review from @plbossart . Can you confirm his concern was addressed (w.r.t. "what causes it not to be cleared?"..?

In the reported scenario, we did not enter the hda_dma_stop_common function.

PR #7440, opened 2023-04-13
Audio: SRC: Audio: SRC: Use only module API and fix IPC4 capture direction

singalsu:
This patch cleans up SRC component. There is no need to keep
    IPC3 version build with legacy component interface. The patch
    mainly removes code under CONFIG_IPC_MAJOR_3 and makes some
    functions available for all builds from CONFIG_IPC_MAJOR_4.
    
    The IPC4 capture issue is fixed by using the src_init()
    IPC rates for source and sink rate initialize. It needs a
    similar kernel update to always set those values. In practical
    topologies in prepare the buffers may not contain the correct
    stream information because the order of host copier and
    dai copier pipeline prepare can be host first where capture
    SRC is usually placed.

### DIFF
--- a/src/audio/src/src.c
+++ b/src/audio/src/src.c
@@ -69,8 +69,6 @@ struct ipc4_config_src {
 DECLARE_SOF_RT_UUID("src", src_uuid, 0xe61bb28d, 0x149a, 0x4c1f,
 		    0xb7, 0x09, 0x46, 0x82, 0x3e, 0xf5, 0xf5, 0xae);
 #elif CONFIG_IPC_MAJOR_3
-static const struct comp_driver comp_src;
-
 /* c1c5326d-8390-46b4-aa47-95c3beca6550 */
 DECLARE_SOF_RT_UUID("src", src_uuid, 0xc1c5326d, 0x8390, 0x46b4,
 		    0xaa, 0x47, 0x95, 0xc3, 0xbe, 0xca, 0x65, 0x50);
@@ -98,6 +96,7 @@ struct comp_data {
 	int source_frames;
 	int sink_frames;
 	int sample_container_bytes;
+	int channels_count;
 	void (*src_func)(struct comp_dev *dev,
 			 struct comp_data *cd,
 			 const struct audio_stream __sparse_cache *source,
@@ -487,7 +486,7 @@ static int src_rate_check(const void *spec)
 {
 	const struct ipc4_config_src *ipc_src = spec;
 
-	if (ipc_src->base.audio_fmt.sampling_frequency == 0 && ipc_src->sink_rate == 0)
+	if (ipc_src->base.audio_fmt.sampling_frequency == 0 || ipc_src->sink_rate == 0)
 		return -EINVAL;
 
 	return 0;
@@ -496,8 +495,14 @@ static int src_rate_check(const void *spec)
 static int src_stream_pcm_source_rate_check(struct ipc4_config_src cfg,
 					    struct sof_ipc_stream_params *params)
 {
-	if (cfg.base.audio_fmt.sampling_frequency &&
-	    params->rate != cfg.base.audio_fmt.sampling_frequency)
+	/* Nothing to check */
+	return 0;
+}
+
+static int src_stream_pcm_sink_rate_check(struct ipc4_config_src cfg,
+					  struct sof_ipc_stream_params *params)
+{
+	if (cfg.sink_rate && params->rate != cfg.sink_rate)
 		return -EINVAL;
 
 	return 0;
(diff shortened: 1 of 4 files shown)


lgirdwood:
@singalsu any ETA ?

src/audio/src/src.c
```diff
@@ -592,6 +591,27 @@ static void src_set_sink_params(struct comp_dev *dev, struct comp_buffer __spars
...
+}
+
+static void src_set_alignment(struct audio_stream __sparse_cache *source,
+			      struct audio_stream __sparse_cache *sink)
+{
+	const uint32_t byte_align = 1;
+	const uint32_t frame_align_req = 1;
+
+	audio_stream_init_alignment_constants(byte_align, frame_align_req, source);
+	audio_stream_init_alignment_constants(byte_align, frame_align_req, sink);
```
paulstelian97, line 612: Does this work with XCC builds? IPC3 isn't dead just yet. Do those implementations not have stricter alignment requirements?

singalsu:
> @singalsu any ETA ?

Should be few days after finding solution to current DMIC issue. I need to make IPC3 and IPC4 test topologies to avoid breaking things with this, we've removed from SOF platforms those previously had test topologies for SRC and we can't use v2.2 to test this change.

src/audio/src/src.c
singalsu: Yep, I'm testing with IPC3 too and will do for SRC after I can get back to this topic. I'm still using IPC3 builds for tests since we don't have everything yet converted to module adapter and IPC4. I prefer to develop in smaller steps like legacy -> module_adapter -> IPC3 to IPC4. This is my personal work preference.

paulstelian97: I have tested this patch on i.MX8QM (platform "imx8") and it seems to be fine on GCC (XTOS and Zephyr) and XCC (Zephyr, support not yet upstream).

singalsu: > I have tested this patch on i.MX8QM (platform "imx8") and it seems to be fine on GCC (XTOS and Zephyr) and XCC (Zephyr, support not yet upstream).

Great! I just fixed some IPC4 issues and will test a bit more before update. Did you try also capture direction?

paulstelian97: I don't think we _have_ any SRC + capture topologies on i.MX so nothing to test. So only playback tested.

singalsu: I got a capture direction SRC fail in testbench that is IPC3 but I haven't replicated it in a real device, so I'm not sure it's a real issue.

paulstelian97: Yeah, thankfully @iuliana-prodan really was quick to point out that we do in fact have the capture direction as well. Just tested it, seems fine on my end. Maybe it takes multiple iterations to reproduce? (I only tried once on two rates where SRC has to do work)

singalsu: That's very good news! The pipelines are not fully accurate in testbench with file instead of host/dai. In my case fail happened immediately in prepare() where one of the requested rates was zero.

src/audio/src/src.c
```diff
@@ -503,30 +502,58 @@ static int src_stream_pcm_source_rate_check(struct ipc4_config_src cfg,
...
+	ret = buffer_set_params(sink_c, &src_params, true);
 	buffer_release(sink_c);
+
+	/* To pass source and sink rate check functions, alternatively skip the
+	 * checks for IPC4.
+	 */
+	if (dev->direction == SOF_IPC_STREAM_PLAYBACK)
+		params->rate = cd->ipc_config.base.audio_fmt.sampling_frequency;
+	else
+		params->rate = cd->ipc_config.sink_rate;
```
ranj063, line 554: @singalsu I think we should use params->rate = cd->ipc_config.sink_rate; for both directions

src/audio/src/src.c
```diff
@@ -731,32 +777,39 @@ static int src_params_general(struct comp_dev *dev, struct comp_data *cd,
 
 	comp_info(dev, "src_params()");
 
-	src_set_params(dev, params);
-	err = src_verify_params(dev, cd, params);
+	err = src_set_params(mod);
+	if (err < 0) {
+		comp_err(dev, "src_params(): set params failed.");
+		return -EINVAL;
```
ranj063, line 783: return err?

src/audio/src/src.c
```diff
@@ -731,32 +777,39 @@ static int src_params_general(struct comp_dev *dev, struct comp_data *cd,
...
+	}
+
+	err = src_verify_params(mod);
 	if (err < 0) {
 		comp_err(dev, "src_params(): pcm params verification failed.");
 		return -EINVAL;
```
ranj063, line 789: same here. it is existing code but better to fix it

ranj063, line 789: same here. it is existing code but better to fix it

src/audio/src/src.c
```diff
@@ -930,39 +1006,59 @@ static int src_init(struct processing_module *mod)
...
+			 cd->ipc_config.base.audio_fmt.depth);
+		rfree(cd);
+		return -EINVAL;
+	}
+#elif CONFIG_IPC_MAJOR_3
+	mod->verify_params_flags = BUFF_PARAMS_RATE;
+#endif
 
 	mod->simple_copy = true;
-
```
ranj063, line 948: spurious change

ranj063 (commented):
looks good @singalsu. Just a few minor comments to address from my side

ranj063 (approved):

singalsu:
No changes, new push to restart CI.

btian1 (approved):
@singalsu , there are two patches in this PR, however, the comments part is totally same with the second patch, may need change PR title and comments slightly to reflect the first patch?

singalsu:
> @singalsu , there are two patches in this PR, however, the comments part is totally same with the second patch, may need change PR title and comments slightly to reflect the first patch?

Could it be a problem in your browser or github? The commits should be:

```
Audio: Module adapter: Pass entire new() IPC to SRC

This patch differentiates SRC component from normal processing
component IPC. The corrected size allows the client component to
receive all of init() IPC.

```

and 

```
Audio: SRC: Use only module API and fix IPC4 capture direction

This patch cleans up SRC component. There is no need to keep
IPC3 version build with legacy component interface. The patch
mainly removes code under CONFIG_IPC_MAJOR_3 and makes some
functions available for all builds from CONFIG_IPC_MAJOR_4.

The IPC4 capture issue is fixed by using the src_init()
IPC rates for source and sink rate initialize. It needs a
similar kernel update to always set those values. In practical
topologies in prepare the buffers may not contain the correct
stream information because the order of host copier and
dai copier pipeline prepare can be host first where capture
SRC is usually placed.

```

kv2019i:
One case of "ASSERTION FAIL [aligned_addr == addr]" on one DUT in https://sof-ci.01.org/sofpr/PR7440/build7322/devicetest/index.html . Not related to this PR.

Otherwise looks good, proceeding with merge.

PR #7441, opened 2023-04-13
Use accessors to read all struct audio_stream members

lyakh:
users shouldn't access `struct audio_stream` internals directly, add accessors for that. This is part 1 - adding read accessors. We keep direct access in the header itself and in buffer.c - for those it's logical to have direct knowledge about internal structure of the buffer class. After part 1 we will add write accessors too.

### DIFF
--- a/src/audio/aria/aria.c
+++ b/src/audio/aria/aria.c
@@ -218,6 +218,7 @@ static void aria_free(struct comp_dev *dev)
 static void aria_set_stream_params(struct comp_buffer *buffer, struct aria_data *cd)
 {
 	struct comp_buffer __sparse_cache *buffer_c;
+	enum sof_ipc_frame valid_fmt, frame_fmt;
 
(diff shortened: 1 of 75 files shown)


cujomalainey (approved):
approval for google_*, drc and crossover

kv2019i (approved):
I do hope this is aligned with @marcinszkudlinski so we don't have to do this again soon. Looks good though.

lgirdwood:
> I do hope this is aligned with @marcinszkudlinski so we don't have to do this again soon. Looks good though.

Intent is to be able to change core internal with minimal impact to clients (as clients will use the accessor APIs). This is likely to conflict easily so will merge. We need to add more accesssors around pipelines and buffers now.

marc-hb:
This was merged but it does not compile:
https://github.com/thesofproject/sof/actions/runs/4785404105/jobs/8508074030

Issue #6316, opened 2022-09-20
[FEATURE] add get_total_data_processed in mixin &mixout for new pipeline latency algorithm


RanderWang:
In patch https://github.com/thesofproject/sof/pull/5985 update the algorithm of pipeline latency calculation. It is based on the number of buffered data blocks between the first and last component. It add ops get_total_data_processed in copier module, but didn't add it in mixin & mixout module.  The algorithm can't work for multiple-pipeline stream with mixin & mixout.

RanderWang:
@softwarecki

softwarecki:
@RanderWang yeah, that's right! I didn't do it right away because another colleague was expanding a functionality od the mixin/mixout at that time. At the moment I'm on the turbo critical path related to the upstream mtl platform so adding that functionality got a lower priority. I have this functionality on my list and will add it in the future.

PR #7442, opened 2023-04-14
module_adapter: add get_total_process_data_process support

RanderWang:
Get_total_process_data_process is used to get the processed data of a module when stream is started. It is used to calculate pipeline latency with the algorithm of substracting processed data count between source pipeline module and sink pipeline module. This patch will calculate consumed & produced data count based on input & output buffer setting and return produced data count for source module and consumed data for sink module.

Currently only copier supports this function, now add it to module_adapter framework to support other modules.

Fix https://github.com/thesofproject/sof/issues/6316


RanderWang:
comp_get_total_data_processed usage:

````c
struct comp_dev *pipeline_get_dai_comp_latency(uint32_t pipeline_id, uint32_t *latency)
{
       ....

	/* Walks through the ipc component list and get sink endpoint component of the given
	 * pipeline. This function returns the first sink. We assume that dai is connected to pin 0.
	 */
	ipc_sink = ipc_get_ppl_sink_comp(ipc, pipeline_id);
	while (ipc_sink) {
		struct comp_buffer *buffer;
		uint64_t input_data, output_data;
		struct ipc4_base_module_cfg input_base_cfg = {.ibs = 0};
		struct ipc4_base_module_cfg output_base_cfg = {.obs = 0};
		int ret;

		/* Calculate pipeline latency */
		input_data = **comp_get_total_data_processed**(source, 0, true);
		output_data = **comp_get_total_data_processed**(ipc_sink->cd, 0, false);

		ret = comp_get_attribute(source, COMP_ATTR_BASE_CONFIG, &input_base_cfg);
		if (ret < 0)
			return NULL;

		ret = comp_get_attribute(ipc_sink->cd, COMP_ATTR_BASE_CONFIG, &output_base_cfg);
		if (ret < 0)
			return NULL;

		if (input_data && output_data && input_base_cfg.ibs && output_base_cfg.obs)
			*latency += input_data / input_base_cfg.ibs -
				output_data / output_base_cfg.obs;
````

File: src/include/sof/audio/module_adapter/module/generic.h [hunk 1]
```diff
@@ -209,6 +210,10 @@ struct processing_module {
 
 	/* table containing the list of connected sources */
 	struct module_source_info *source_info;
+
+	/* total processed data after stream started */
+	uint64_t total_data_consumed;
+	uint64_t total_data_produced;
```
kv2019i, line 216: Not blocking: just wondering whether 64bit is really needed and worth the memory/cycles spent. I guess the any code using get_total_process_data anyways needs to handle possibility of wrap-arounds, so not sure 64bit is needed (latency will never exceed 32bit range).

But alas, this is not really related to this PR as the data type was picked already for copier.

kv2019i (approved):
No blockers.

lgirdwood (approved):

PR #6951, opened 2023-01-13
Optimize host copier with one less buffer copy

btian1:
Currently, copier actually created 2 devices for data input, it is
host and copier, hence there are two times copy from DMA buffer
to local buffer then to component buffer.

This PR intend to remove host device and one more time copy
through use host data to handle host interface calling.

Basically, there are several steps to go there:

1. Add host data to copier data structure.
2. Share host data structure to copier by add a host copier common header file.
3. In copier, all host related ops calling should be replaced with new host exposed interface.
4. Remove one more time copy and host endpoint buffer.
5. Remove host device.

Performance data:
1. mcps, saved 1.9mcps, 5.9 --> 4.0, 33% saved for copier.
    Before: comp:0 0x40000 perf comp_copy samples 48 period 1000 cpu avg 586 peak 598
    After:   comp:0 0x40000 perf comp_copy samples 48 period 1000 cpu avg 408 peak 422

2. memory, since host endpoint buffer and host device got removed, for each copier,
   16bit/32bit 384/768 bytes saved for buffer, 112 saved for host device.
   As a summary, 16/32 bit playback/capture totally saved 496/884 bytes for each copier usage.
   For the cases that have more than 10 copier usage, memory saving is quite remarkable.

Note：multiple sink host copier is not tested and will be fixed in a follow up dai optimization PR.

Signed-off-by: Baofeng Tian <baofeng.tian@intel.com>

PR #7443, opened 2023-04-14
Fix legacy Zephyr builds with non-Zephyr native drivers

lyakh:
This fixes breakage introduced by #6951 to legacy driver builds

### DIFF
--- a/src/audio/host-legacy.c
+++ b/src/audio/host-legacy.c
@@ -5,6 +5,7 @@
 // Author: Liam Girdwood <liam.r.girdwood@linux.intel.com>
 //         Keyon Jie <yang.jie@linux.intel.com>
 
+#include <sof/audio/host_copier.h>
 #include <sof/audio/buffer.h>
 #include <sof/audio/component_ext.h>
 #include <sof/audio/pcm_converter.h>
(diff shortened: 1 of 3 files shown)


lyakh:
CI failures: both on-device tests https://sof-ci.01.org/sofpr/PR7443/build5936/devicetest/index.html and https://sof-ci.01.org/sofpr/PR7443/build5935/devicetest/index.html are suspend-resume failures. Moreover, this PR only introduces functional changes to non-native builds `CONFIG_ZEPHYR_NATIVE_DRIVERS=n` and such builds aren't tested by CI ATM.

kv2019i (approved):

kv2019i:
Needs a second review, @btian1 @juimonen @paulstelian97 @ranj063 ?

paulstelian97:
Question: how does this affect xtos builds? Is this file common to both?

btian1:
can you specify which build will have copier + host_legacy combination? does CONFIG_ZEPHYR_NATIVE_DRIVERS=n + zephyr build cover this case?

lyakh:
> can you specify which build will have copier + host_legacy combination? does CONFIG_ZEPHYR_NATIVE_DRIVERS=n + zephyr build cover this case?

@btian1 IPC4 with Zephyr with `CONFIG_ZEPHYR_NATIVE_DRIVERS=n`

lyakh:
> Question: how does this affect xtos builds? Is this file common to both?

@paulstelian97 directly, obviously, with XTOS builds you cannot have native Zephyr drivers, so if you use IPC4 with XTOS - you'd be using this code and presumably it's broken without this PR

btian1:
> > can you specify which build will have copier + host_legacy combination? does CONFIG_ZEPHYR_NATIVE_DRIVERS=n + zephyr build cover this case?
> 
> @btian1 IPC4 with Zephyr with `CONFIG_ZEPHYR_NATIVE_DRIVERS=n`

With zephyr and CONFIG_ZEPHYR_NATIVE_DRIVERS=n?

if CI does not cover this test, how can we know it is an issue and it is fixed by this PR?
Also, if this config is not POR, then we should not support this, if need support, we need ask CI to add this configuration.

when I work on #6951, I deliberately avoid to change host-legacy code, and only change copier and host-zephyr to avoid introduce more regression.

lyakh:
> > > can you specify which build will have copier + host_legacy combination? does CONFIG_ZEPHYR_NATIVE_DRIVERS=n + zephyr build cover this case?
> > 
> > 
> > @btian1 IPC4 with Zephyr with `CONFIG_ZEPHYR_NATIVE_DRIVERS=n`
> 
> With zephyr and CONFIG_ZEPHYR_NATIVE_DRIVERS=n?

yes

> if CI does not cover this test, how can we know it is an issue and it is fixed by this PR? Also, if this config is not POR, then we should not support this, if need support, we need ask CI to add this configuration.

The CI doesn't cover all possible configurations and I'd say it wouldn't be possible or feasible to try that. We don't even have analogs of Linux kernel's "allyesconfig" and "randconfig." Still it doesn't mean that we're free to break those untested configurations.  Particularly since the CI only tests on Intel hardware, we still shouldn't break non-Intel configurations. And this kind of a configuration with Zephyr, legacy drivers and IPC4 might well be used by somebody at least as a transitional phase. Still we can rather easily test this PR - try that configuration on Intel hardware before that breaking PR, the current "main" and after this PR.

> when I work on #6951, I deliberately avoid to change host-legacy code, and only change copier and host-zephyr to avoid introduce more regression.

Yes, I remember, we even discussed those legacy stubs for new functions and we briefly discussed the possibility of breakage. But it is indeed difficult to analyse all possibilities "in your head" and without CI testing no wonder we broke it. Still I think it would be good to fix it and to try to avoid and future unintended breakage.

paulstelian97:
> > Question: how does this affect xtos builds? Is this file common to both?
> 
> @paulstelian97 directly, obviously, with XTOS builds you cannot have native Zephyr drivers, so if you use IPC4 with XTOS - you'd be using this code and presumably it's broken without this PR

Understood, so when we migrate i.MX8 platforms to IPC4 we're going to have to only migrate the Zephyr port as opposed to both of them (currently Zephyr is functional and XTOS is more stable/more tested). @dbaluta fyi

lyakh:
> > > Question: how does this affect xtos builds? Is this file common to both?
> > 
> > 
> > @paulstelian97 directly, obviously, with XTOS builds you cannot have native Zephyr drivers, so if you use IPC4 with XTOS - you'd be using this code and presumably it's broken without this PR
> 
> Understood, so when we migrate i.MX8 platforms to IPC4 we're going to have to only migrate the Zephyr port as opposed to both of them (currently Zephyr is functional and XTOS is more stable/more tested). @dbaluta fyi

@paulstelian97 well theoretically you can have IPC4+XTOS but only with "legacy" (current SOF tree) drivers. Since you're talking about an IPC4 migration I assume that's your goal. Then yes, I'd say it would make more sense to go for Zephyr+IPC4. And also for Zephyr-native drivers. So, you'd have to migrate both drivers and the IPC implementation. Then you can choose which one you migrate first. Maybe the easiest would be from
1. Zephyr + IPC3 + SOF drivers
2. to Zephyr + IPC3 + Zephyr native drivers
3. to Zephyr + IPC4 + Zephyr native drivers

Then you avoid that awkward place with IPC4 + SOF drivers where you'd need to use host-legacy with compier.

dbaluta:
@lyakh question about Intel state with native zephyr drives. How does one enable one native driver instead of a SOF driver? What native drivers Intel already has enabled?

We are looking on enabling PDM microphone for IMX and we noticed that Zephyr already supports this via MCU SDK and we want to avoid rewriting it.

Is Intel DMIC supported via native driver for Zephyr?

kv2019i:
@dbaluta  wrote: 
> @lyakh question about Intel state with native zephyr drives. How does one enable one native driver instead of a SOF driver? > What native drivers Intel already has enabled?

We used CONFIG_ZEPHYR_NATIVE_DRIVERS for the transition, but now all drivers are moved over (hd-dma and dw for DMA; dmic,/ssp/hda/sdw DAI drivers), clock/interrupts were moved already earlier.

PR #7444, opened 2023-04-14
west: test rimage with only ssl 3 support

juimonen:
Test rimage version that supports only ssl 3 upwards.

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -36,7 +36,7 @@ manifest:
     - name: rimage
       repo-path: rimage
       path: sof/rimage
-      revision: 48777207f5fdc49314dd4709d660e891c48d5ba9
+      revision: pull/157/head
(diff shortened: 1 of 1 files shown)


marc-hb:
https://github.com/thesofproject/sof/actions/runs/4701641680/jobs/8337938948?pr=7444 fails because (among others?) https://github.com/zephyrproject-rtos/docker-image/pkgs/container/zephyr-build is still based on Ubuntu 20.04

marc-hb:
https://sof-ci.01.org/sofpr/PR7444/build5899/build/mtl_zph_ipc4_error.txt should be more recent but it fails the same

juimonen:
SOFCI TEST

gkbldcig:

PR #7446, opened 2023-04-14
[SKIP CI] .github/zephyr: move zephyr revision in first results matrix position

marc-hb:
Recent experience has shown that upstream Zephyr regressions are much more common than platform-specific failures. So move the Zephyr version in first position, so this will group zephyr regressions and make them much more obvious.

Use very short keys for the revisions because the left column in Github's results "checks" tab is very narrow and cannot be resized.

### DIFF
--- a/.github/workflows/zephyr.yml
+++ b/.github/workflows/zephyr.yml
@@ -63,6 +63,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # These keys are kept short because Github's left column is not resizable.
+        # Search "zephyr_revision" and see below what they expand to.
+        zephyr_revision: [
+          mnfst,  # special value: don't override sof/west.yml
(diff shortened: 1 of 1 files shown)


marc-hb:
Seems to work!

https://github.com/thesofproject/sof/actions/runs/4703435326/jobs/8341931809?pr=7446

Screenshot 2023-04-14 at 13 33 04

.github/workflows/zephyr.yml
```diff
@@ -63,6 +63,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        zephyr_revision: [
```
aborisovich, line 66: Please add comments to what is "mnfst" and "zmain".
It is obvious for us, but somebody new to CI yamls may take time to decrypt.

aborisovich (approved):

PR #7454, opened 2023-04-18
Topology2: Add mtl-rt711-l0-rt1316-l23-rt714-l1 support

aiChaoSONG:
This topology could be used by MTL SDW RVP with 3-in-1 SDCA codec board, or other hardware configuration with the same codec layout:
    SDW0: RT711 Headphone
    SDW1: RT714 DMIC
    SDW2: RT1316 Speaker
    SDW3: RT1316 Speaker


tools/topology/topology2/sof-ace-tplg/tplg-targets.cmake
```diff
@@ -30,4 +30,8 @@ PDM1_MIC_B_ENABLE=1,DMIC0_PCM_ID=99,PREPROCESS_PLUGINS=nhlt,\
 NHLT_BIN=nhlt-sof-mtl-max98357a-rt5682.bin,DEEPBUFFER_FW_DMA_MS=10,HEADSET_SSP_DAI_INDEX=2,\
 SPEAKER_SSP_DAI_INDEX=0,HEADSET_CODEC_NAME=SSP2-Codec,SPEAKER_CODEC_NAME=SSP0-Codec,\
 BT_NAME=SSP1-BT,BT_INDEX=1,BT_ID=8,BT_PCM_NAME=Bluetooth,INCLUDE_ECHO_REF=true"
+
+"cavs-sdw\;sof-mtl-rt711-l0-rt1316-l23-rt714-l1\;PLATFORM=mtl,NUM_SDW_AMPS=2,SDW_DMIC=1,\
+NUM_HDMIS=0,SDW_SPK_STREAM=SDW2-Playback,SDW_SPK_IN_STREAM=SDW2-Capture,SDW_DMIC_STREAM=SDW1-Capture,\
+PREPROCESS_PLUGINS=nhlt,NHLT_BIN=sof-mtl-rt711-l0-rt1316-l23-rt714-l1.bin"
```
bardliao, line 36: nhlt is only required when we use PCH DMIC

tools/topology/topology2/sof-ace-tplg/tplg-targets.cmake
```diff
@@ -30,4 +30,8 @@ PDM1_MIC_B_ENABLE=1,DMIC0_PCM_ID=99,PREPROCESS_PLUGINS=nhlt,\
 NHLT_BIN=nhlt-sof-mtl-max98357a-rt5682.bin,DEEPBUFFER_FW_DMA_MS=10,HEADSET_SSP_DAI_INDEX=2,\
 SPEAKER_SSP_DAI_INDEX=0,HEADSET_CODEC_NAME=SSP2-Codec,SPEAKER_CODEC_NAME=SSP0-Codec,\
 BT_NAME=SSP1-BT,BT_INDEX=1,BT_ID=8,BT_PCM_NAME=Bluetooth,INCLUDE_ECHO_REF=true"
+
+"cavs-sdw\;sof-mtl-rt711-l0-rt1316-l23-rt714-l1\;PLATFORM=mtl,NUM_SDW_AMPS=2,SDW_DMIC=1,\
```
plbossart, line 34: I am also not very happy about NUM_SDW_AMPS=2 because it does not provide any indication that different links are used. This could also be set in the case of two amplifiers on the same link.

This seems very confusing @aiChaoSONG @bardliao

bardliao: > I am also not very happy about NUM_SDW_AMPS=2 because it does not provide any indication that different links are used. This could also be set in the case of two amplifiers on the same link.
> 
> This seems very confusing @aiChaoSONG @bardliao

Maybe rename it to **NUM_SDW_AMP_LINKS**?

plbossart: Ah, it's about the number of links or the number of amps? That's a nice confusion indeed :-)

bardliao: > Ah, it's about the number of links or the number of amps? That's a nice confusion indeed :-)

It is about number of links. If the 2 amps are on the same sdw link, NUM_SDW_AMPS should be 1 because we have only one cpu dai.

plbossart: ok, let's rename this then, thanks for the explanation @bardliao

aiChaoSONG:
SOFCI TEST

kv2019i (approved):

bardliao (approved):

plbossart:
@aiChaoSONG can you fix the conflicts and better isolate the SoundWire topologies from Chromebooks? We want to have enough separation with markers/comments that we can merge independent patches without tripping on each other.

plbossart (approved):

ranj063 (approved):

gkbldcig:

PR #7458, opened 2023-04-19
zephyr: include: rtos: Use native Zephyr cache management functions o…

LaurentiuM1234:
Since Zephyr offers cache-management operations for the arm64 architecture we can make use of them on all ARM64-based platforms. In time, if the architecture supports it, more platforms could end up using them instead of using the ones in arch/lib/cache.h.

### DIFF
--- a/zephyr/include/rtos/cache.h
+++ b/zephyr/include/rtos/cache.h
@@ -24,12 +24,6 @@
 
 #endif /* defined(CONFIG_XTENSA) && defined(CONFIG_INTEL) */
 
-/* writeback and invalidate data */
-#define CACHE_WRITEBACK_INV	0
-
-/* invalidate data */
-#define CACHE_INVALIDATE	1
-
 /* sanity check - make sure CONFIG_DCACHE_LINE_SIZE is valid */
 #if !defined(CONFIG_DCACHE_LINE_SIZE_DETECT) && (CONFIG_DCACHE_LINE_SIZE > 0)
 #define DCACHE_LINE_SIZE CONFIG_DCACHE_LINE_SIZE
--- a/zephyr/include/rtos/cache.h
+++ b/zephyr/include/rtos/cache.h
@@ -6,14 +6,65 @@
 #ifndef __ZEPHYR_RTOS_CACHE_H__
 #define __ZEPHYR_RTOS_CACHE_H__
 
-/* TODO: align with Zephyr generic cache API when ready */
 #define __SOF_LIB_CACHE_H__
-#include <arch/lib/cache.h>
+
+#if !defined(__ASSEMBLER__) && !defined(LINKER)
+
+#include <zephyr/cache.h>
+#include <zephyr/debug/sparse.h>
+
+#if defined(CONFIG_XTENSA) && defined(CONFIG_INTEL)
+
+/* definitions required by xtensa-based Intel platforms.
+ *
+ * TODO: if possible, move these to Zephyr.
+ */
+#define SRAM_UNCACHED_ALIAS 0x20000000
+#define is_cached(address) (!!((uintptr_t)(address) & SRAM_UNCACHED_ALIAS))
+
+#endif /* defined(CONFIG_XTENSA) && defined(CONFIG_INTEL) */
 
 /* writeback and invalidate data */
 #define CACHE_WRITEBACK_INV	0
 
 /* invalidate data */
 #define CACHE_INVALIDATE	1
+
+/* sanity check - make sure CONFIG_DCACHE_LINE_SIZE is valid */
+#if !defined(CONFIG_DCACHE_LINE_SIZE_DETECT) && (CONFIG_DCACHE_LINE_SIZE > 0)
+#define DCACHE_LINE_SIZE CONFIG_DCACHE_LINE_SIZE
+#else
+#if defined(CONFIG_LIBRARY) || defined(CONFIG_ZEPHYR_POSIX)
+#define DCACHE_LINE_SIZE 64
+#else
+#error "Invalid cache configuration."
+#endif /* defined(CONFIG_LIBRARY) || defined(CONFIG_ZEPHYR_POSIX) */
+#endif /* !defined(CONFIG_DCACHE_LINE_SIZE_DETECT) && (CONFIG_DCACHE_LINE_SIZE > 0) */
+
+static inline void dcache_writeback_region(void __sparse_cache *addr, size_t size)
+{
+	/* TODO: return value should probably be checked here */
+	sys_cache_data_flush_range((__sparse_force void *)addr, size);
+}
+
+static inline void dcache_invalidate_region(void __sparse_cache *addr, size_t size)
+{
+	/* TODO: return value should probably be checked here */
+	sys_cache_data_invd_range((__sparse_force void *)addr, size);
+}
+
+static inline void icache_invalidate_region(void __sparse_cache *addr, size_t size)
+{
+	/* TODO: return value should probably be checked here */
+	sys_cache_instr_invd_range((__sparse_force void *)addr, size);
+}
+
+static inline void dcache_writeback_invalidate_region(void __sparse_cache *addr, size_t size)
+{
+	/* TODO: return value should probably be checked here */
+	sys_cache_data_flush_and_invd_range((__sparse_force void *)addr, size);
+}
+
+#endif /* !defined(__ASSEMBLER__) && !defined(LINKER) */
 
 #endif /* __ZEPHYR_RTOS_CACHE_H__ */


paulstelian97 (approved):

zephyr/include/rtos/cache.h
```diff
@@ -8,12 +8,59 @@
 
 /* TODO: align with Zephyr generic cache API when ready */
 #define __SOF_LIB_CACHE_H__
+
+#if !CONFIG_ARM64
 #include <arch/lib/cache.h>
+#else
+#include <zephyr/cache.h>
+#endif /* CONFIG_ARM64 */
```
kv2019i, line 16: This is a very welcome change. FYI  to @nashif  -- there's a change in progress on Zephyr side to allow us to move to sys_cache also on xtensa -> https://github.com/zephyrproject-rtos/zephyr/pull/50136

zephyr/include/rtos/cache.h
```diff
@@ -8,12 +8,59 @@
...
+#if !defined(CONFIG_DCACHE_LINE_SIZE_DETECT) && \
+	(CONFIG_DCACHE_LINE_SIZE > 0)
+
+#define DCACHE_LINE_SIZE CONFIG_DCACHE_LINE_SIZE
+
+static inline void dcache_writeback_region(void *addr, size_t size)
+{
+	/* TODO: return value should probably be checked here */
+	sys_cache_data_flush_range(addr, size);
+}
```
kv2019i, line 40: Hmm, did you consider to put this to arch/arm64/lib/cache.h ? I think that would be cleaner still as long as we need some wrapping on SOF side? 

Given our intention is to remove rtos/cache.h and use sys_cache_.* directly, this is probably not a big issue to have some ifdefs in this file, but if it's not a big task, why not use the existing infra and add arm64 arch files to SOF. I think you'll need that for other stuff as well (like the compiler_version.h thing).

kv2019i (commented):
In principle ok, but some questions, please see inline

zephyr/include/rtos/cache.h
nashif: yes, this is almost ready to go, we just fixed the inlining issue, ideally we shall be able to just use the zephyr cache API for everyone avoid indirection.

zephyr/include/rtos/cache.h
LaurentiuM1234: Moving all this stuff to `arch/arm64/lib/cache.h` or keeping it in here should be the same thing as it's only supposed to be temporary anyways (thanks to @nashif's patches which hopefully will take care of things on xtensa side). Also, shouldn't we be trying to get rid of the `arch/` hierarchy entirely instead of adding more stuff to it? (although only temporarily in this case)

kv2019i: @LaurentiuM1234  Right, I change my mind on the compiler_info and given this will go away as well, no need to add arch if we already know now it will be removed soon.

lgirdwood:
@LaurentiuM1234 any update, I think the Zephyr cache changes are now merged ?

LaurentiuM1234:
> @LaurentiuM1234 any update, I think the Zephyr cache changes are now merged ?

Checked https://github.com/zephyrproject-rtos/zephyr/pull/50136 and it has been tagged with DNM for now. Seems like it will be merged later on.

Might be worth keeping the wrappers over the Zephyr cache-management API at least on ARM64 where it seems like there's an option to `device_map` (`K_MEM_CACHE_NONE`) which marks the region as non-cacheable. In this case there would probably be no need to invalidate the data cache (at least for the window regions: stream, hostbox, outbox etc...). We'd have to pass something like `region_is_cached` to the wrappers and based on its value decide to call the Zephyr cache-management API. But at this point this is just an idea at the top of my head. Removing the wrappers for now and re-adding them later on could also be a solution I guess since I'm not 100% confident that my idea will work as expected. Let me know what you think @lgirdwood @kv2019i @nashif.

kv2019i:
@LaurentiuM1234 wrote:
> Checked [zephyrproject-rtos/zephyr#50136](https://github.com/zephyrproject-rtos/zephyr/pull/50136) and it has been tagged with DNM for now. Seems like it will be merged later on.

This was now merged and pulled into SOF main.

> Might be worth keeping the wrappers over the Zephyr cache-management API 
> at least on ARM64 where it seems like there's an option to `device_map` (`K_MEM_CACHE_NONE`) 

I think we must keep the wrappers as long as we have any non-Zephyr targets in SOF main (although at some point we can turn these into reverse wrappers, i.e. change app code to call Zephyr interface, and provide wrappers for XTOS builds).

But for Zephyr builds, we should map to native Zephyr calls for all targets and not include arch/lib/cache.h (this should be reserved for XTOS).

Can you update the PR so that the mapping to native Zephyr calls is done for all targets and then you can have some additional defines for ARM64 on top?

LaurentiuM1234:
@kv2019i Sorry for taking such a long time to update this. It should be good to go RN. Tested on NXP's side and seems to work fine.

LaurentiuM1234:
Fixed CI errors for sparse Zephyr and included a little macro cleanup.

zephyr/include/rtos/cache.h
```diff
@@ -6,14 +6,72 @@
...
+#else
+
+/* If CONFIG_DCACHE_LINE_SIZE is not set appropriately then
+ * fallback to a default value. This ought to be alright
+ * since all platforms that support Zephyr use this value.
+ * This is better than forcing all vendors to set
+ * CONFIG_DCACHE_LINE_SIZE in order to avoid compilation
+ * errors.
+ */
+#define DCACHE_LINE_SIZE 128
```
lyakh, line 41: I think it's 64 at least on Intel platforms

lyakh (changes requested):

zephyr/include/rtos/cache.h
LaurentiuM1234: yup, rn I'm working on adding a new configuration (some sort of wrapper over CONFIG_DCACHE_LINE_SIZE from Zephyr) which will be the equivalent of all of those messy ifdefs

LaurentiuM1234:
Updates:

1) Introduced `CONFIG_SOF_DCACHE_LINE_SIZE` which replaces all the messy ifdefs from `zephyr/include/rtos/cache.h`. This config can be removed when all vendors define `CONFIG_DCACHE_LINE_SIZE` in their Zephyr board configurations. For now, this configuration will try to use the value of `CONFIG_DCACHE_LINE_SIZE` if it's set properly, otherwise it will fallback to some default value based on platform.

zephyr/include/rtos/cache.h
```diff
@@ -6,14 +6,63 @@
...
 #define CACHE_INVALIDATE	1
 
+/* use SOF's wrapper over CONFIG_DCACHE_LINE_SIZE from Zephyr.
+ * This is because we don't want to force vendors to define
+ * CONFIG_DCACHE_LINE_SIZE inside their board configurations
+ * in order to avoid compilation errors.
+ *
+ * TODO: in time, switch to simply using CONFIG_DCACHE_LINE_SIZE
+ */
+#define DCACHE_LINE_SIZE CONFIG_SOF_DCACHE_LINE_SIZE
```
lyakh, line 34: I'm wondering how much more work would it be to go with `CONFIG_DCACHE_LINE_SIZE` directly?

lyakh (approved):

zephyr/include/rtos/cache.h
LaurentiuM1234: Hmm, shouldn't be much trouble. Will give it a shot and submit a Zephyr PR if I can get something working.

btian1 (approved):

zephyr/include/rtos/cache.h
dbaluta: @lyakh so should we merge this?

LaurentiuM1234:
Marking this as a draft until I can figure out if we can easily get rid of CONFIG_SOF_DCACHE_LINE_SIZE by using CONFIG_DCACHE_LINE_SIZE from Zephyr.

LaurentiuM1234:
Zephyr PR in which CONFIG_DCACHE_LINE_SIZE is set for all SOF-supported boards is ready: https://github.com/zephyrproject-rtos/zephyr/pull/57926. Will update this PR once the Zephyr one is merged.

LaurentiuM1234:
https://github.com/zephyrproject-rtos/zephyr/pull/57926 has been merged. Now waiting on https://github.com/thesofproject/sof/pull/7642 and this can be merged IMO.

paulstelian97 (approved):

kv2019i:
Known error https://sof-ci.01.org/sofpr/PR7458/build8506/devicetest/index.html , fuzz check failure unrelated to this PR. Proceeding with merge.

PR #7459, opened 2023-04-19
Add interrupt-related changes for arm64-based platforms

LaurentiuM1234:
This PR introduces all required interrupt-related changes for ARM64 platforms.

### DIFF
--- a/zephyr/include/rtos/interrupt.h
+++ b/zephyr/include/rtos/interrupt.h
@@ -52,7 +52,8 @@ static inline void interrupt_unregister(uint32_t irq, const void *arg)
  */
 static inline int interrupt_get_irq(unsigned int irq, const char *cascade)
 {
-#if defined(CONFIG_LIBRARY) || defined(CONFIG_ACE) || defined(CONFIG_ZEPHYR_POSIX)
+#if defined(CONFIG_LIBRARY) || defined(CONFIG_ACE) || \
+	defined(CONFIG_ZEPHYR_POSIX) || defined(CONFIG_ARM64)
 	return irq;
 #else
 	if (cascade == irq_name_level2)


paulstelian97 (approved):
For the first commit, things are getting unwieldy so they might need to be changed in the future. Approve overall though, that's the subject of a future refactor.

zephyr/include/rtos/interrupt.h
```diff
@@ -83,10 +83,24 @@ static inline uint32_t interrupt_disable(uint32_t irq, void *arg)
...
+ * done after the ISR handler is executed. Also
+ * said interrupts are never masked so there's
+ * also no need for interrupt_clear_mask().
+ */
+#define interrupt_clear(irq)
+#define interrupt_clear_mask(irq, mask)
+#else
+void interrupt_clear_mask(uint32_t irq, uint32_t mask);
+#endif /* CONFIG_ARM64 */
+
```
kv2019i, line 100: And this is really only needed with XTOS. With native Zephyr SOF build, the interrupt interface for drivers is all on Zephyr side. SOF app side only needs primitives to disable interrupts, but that can be done with generic OS interface.

But alas, if you want to maintain XTOS support with ARM64, then you need this.

kv2019i (approved):

zephyr/include/rtos/interrupt.h
paulstelian97: This _looks_ like there's not going to be any XTOS support on ARM64, but correct me if I'm wrong Laurentiu.

We do want to eventually phase out XTOS for the older i.MX platforms as well, but that's not happening yet.

LaurentiuM1234: This was required because of some compilation errors caused by these symbols not existing (e.g: for zephyr_dma_domain which is also used for Zephyr+XTOS).

LaurentiuM1234: Aforementioned issue should be fixed by https://github.com/thesofproject/sof/pull/7470 and  https://github.com/thesofproject/sof/pull/7462. Although not necessarily related to i.MX93 driver support I've added the last commit over there to avoid adding an extra PR dependency as it's already somewhat hard to keep track of them.

zephyr/include/rtos/interrupt.h
```diff
@@ -52,7 +52,8 @@ static inline void interrupt_unregister(uint32_t irq, const void *arg)
  */
 static inline int interrupt_get_irq(unsigned int irq, const char *cascade)
 {
-#if defined(CONFIG_LIBRARY) || defined(CONFIG_ACE) || defined(CONFIG_ZEPHYR_POSIX)
+#if defined(CONFIG_LIBRARY) || defined(CONFIG_ACE) || \
+	defined(CONFIG_ZEPHYR_POSIX) || CONFIG_ARM64
```
paulstelian97, line 56: Lacks consistency. Probably should do defined(CONFIG_ARM64) if there's no reason not to do that.

iuliana-prodan (approved):

Issue #403, opened 2018-09-21
panic dump contains line number and filename


xiulipan:
Dump the filename and line number if we are using software panic.

lgirdwood:
@xiulipan and also preserve the IRQ level prior to switching IRQs OFF.

xiulipan:
close via #423

PR #7461, opened 2023-04-19
zephyr: include: sof: Add Zephyr version of compiler_info.h

LaurentiuM1234:
This is required by ARM64 platforms as they don't have an arch/compiler_info.h so the compilation would fail.

### DIFF
--- a/zephyr/include/sof/compiler_info.h
+++ b/zephyr/include/sof/compiler_info.h
@@ -1,0 +1,37 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright 2023 NXP
+ */
+
+#ifndef __SOF_COMPILER_INFO_H__
+#define __SOF_COMPILER_INFO_H__
+
+#if defined(__XCC__)
+
+#include <xtensa/hal.h>
+
+#define CC_MAJOR (XTHAL_RELEASE_MAJOR / 1000)
+#define CC_MINOR ((XTHAL_RELEASE_MAJOR % 1000) / 10)
+#define CC_MICRO XTHAL_RELEASE_MINOR
+#define CC_NAME "XCC"
+#define CC_DESC " " XCC_TOOLS_VERSION
+
+#elif defined(__GNUC__)
+
+#define CC_MAJOR __GNUC__
+#define CC_MINOR __GNUC_MINOR__
+#define CC_MICRO __GNUC_PATCHLEVEL__
+#define CC_NAME "GCC"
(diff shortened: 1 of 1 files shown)


paulstelian97 (approved):
Approve for now, this should be removed in the future whenever possible as it is a big hack in the end.

zephyr/include/sof/compiler_info.h
```diff
@@ -0,0 +1,24 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright 2023 NXP
+ */
+
+#ifndef __SOF_COMPILER_INFO_H__
+#define __SOF_COMPILER_INFO_H__
```
kv2019i, line 7: This is completely a SOF level construct, so I wonder if you could just add arm64 arch directory and add compiler_info.h there (e.g. sof/src/arch/arm64/include/arch/compiler_info.h).

I also looked at Zephyr. but they don't seem to export any common toolchain version string in toolchain.h. This just exports compiler capabilities in terms of preprocessor defines. So I don't think we can implement the SOF compiler_info interface using general Zephyr interface. Or did I miss something @dcpleung @nashif @marc-hb ?

kv2019i (changes requested):
I don't think this is special to ARM64, this is SOF side abstraction that needs to be added for each new arch.

zephyr/include/sof/compiler_info.h
LaurentiuM1234: You're right, this is not arm64 specific. I have no problem with using the `arch/` hierarchy  (same goes for this one https://github.com/thesofproject/sof/pull/7458) but isn't our goal to get rid of it instead of adding more to it? Unless there's a better alternative to this (using Zephyr support) the only solution (apart from the `arch/` one) which I'm seeing ATM is to make this file a combination of all architecture-specific `compiler_info.h`. If this is not okay either then I guess there's no choice but to introduce `sof/src/arch/arm64/include/arch/compiler_info.h`.

LaurentiuM1234: Should be a bit better now.

kv2019i (approved):
Excellent, thank you!

kv2019i:
Tests look ok. Filed an internal Intel ticket (#403) for the one FW file lookup check fail in https://sof-ci.01.org/sofpr/PR7461/build6235/devicetest/index.html 

Sparse is a known bug and fix is in progress. Likewise for Zephyr mainline build fails.

PR #7463, opened 2023-04-19
Add support for SOF_IPC_FW_READY message initiated by host

LaurentiuM1234:
In the case of i.MX93 Linux will send an IPC_FW_READY message to SOF and will expect to receive the following information in the hostbox:
	1) reply structure
	2) sof_ipc_fw_ready structure
	3) sof_ipc_window structure

This flow is required because Jailhouse is currently controlled from user space. This means that SOF will be up before the platform driver from Linux so the IPC_FW_READY sequence can't be initiated by SOF.

### DIFF
--- a/src/ipc/ipc3/handler.c
+++ b/src/ipc/ipc3/handler.c
@@ -1603,6 +1603,35 @@ void ipc_boot_complete_msg(struct ipc_cmd_hdr *header, uint32_t data)
 	header->dat[1] = data;
 }
 
+static int ipc_fw_ready(void)
+{
+#ifdef CONFIG_IMX93_A55
+	/* VERY IMPORTANT:
+	 *	* due to how the FW is started, i.MX93 has the
+	 *	following flow (please note that the host driver
+	 *	is blacklisted and inserted later on):
+	 *
+	 *	1) Linux kernel boots and user space becomes
+	 *	available.
+	 *	2) FW is started (through Jailhouse) => SOF is running
+	 *	4) Host driver module is inserted.
+	 *	5) Host platform driver sends SOF_IPC_FW_READY and
+	 *	expects SOF to send the SOF_IPC_FW_READY
+	 *	message, the window regions and the reply
+	 *	header in the following order:
+	 *		1) reply structure
+	 *		2) sof_ipc_fw_ready structure
+	 *		3) windows structure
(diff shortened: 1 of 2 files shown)


paulstelian97 (approved):

kv2019i (approved):
I couldn't first make my mind whether this is an ABI change or not, but given this uses existing ABI definitions and is limited to one platform, my end conclusion is that this is not ABI change. Thanks for the good inline documentation, that helped in review.

iuliana-prodan (approved):

kv2019i:
SOFCI TEST

zephyr/wrapper.c
```diff
@@ -215,9 +229,7 @@ int start_complete(void)
 	pm_policy_state_lock_get(PM_STATE_RUNTIME_IDLE, PM_ALL_SUBSTATES);
 	pm_policy_state_lock_get(PM_STATE_SOFT_OFF, PM_ALL_SUBSTATES);
 #endif
-
-	/* let host know DSP boot is complete */
-	return platform_boot_complete(0);
```
dbaluta, line 220: there is no reason to remove the comment from here. so please keep it.

src/ipc/ipc3/handler.c
```diff
@@ -1603,6 +1603,36 @@ void ipc_boot_complete_msg(struct ipc_cmd_hdr *header, uint32_t data)
 	header->dat[1] = data;
 }
```
dbaluta, line 1605: Please use `ipc` prefix to commit subject. e.g `ipc:  Add support for SOF_IPC_FW_READY message initiated by host `

Please don't abruptly mention about jailhouse. 

I would rephrase the paragraph starting at `This flow is required because with ...`.


```This flow is required because on i.MX93, SOF runs on a A55 core and with current design
the FW is started independently and will be up before SOF driver from Linux.

We use Jailhouse to separate to isolate the cores. 
```

LaurentiuM1234:
Fixed problems pointed out by @dbaluta. Please let me know if there's anything else you'd like me to address. Otherwise, IMO this is ready for merge.

paulstelian97 (approved):

dbaluta (approved):

dbaluta:
thanks @LaurentiuM1234! will merge this once CI is ready.

PR #7467, opened 2023-04-19
Remove unused soc.h header inclusions

LaurentiuM1234:
Since i.MX93 doesn't have a soc.h file in Zephyr nor does it need one right now, instead of adding an empty soc.h to Zephyr just encase the '#include <soc.h>' statement in an ifdef block.

### DIFF
--- a/zephyr/lib/alloc.c
+++ b/zephyr/lib/alloc.c
@@ -25,7 +25,6 @@
 #include <zephyr/pm/policy.h>
 #include <version.h>
 #include <zephyr/sys/__assert.h>
-#include <soc.h>
 
 #if defined(CONFIG_ARCH_XTENSA) && !defined(CONFIG_KERNEL_COHERENCE)
 #include <zephyr/arch/xtensa/cache.h>
--- a/zephyr/lib/cpu.c
+++ b/zephyr/lib/cpu.c
@@ -15,7 +15,6 @@
 #include <sof/lib/pm_runtime.h>
 
 /* Zephyr includes */
-#include <soc.h>
 #include <version.h>
 #include <zephyr/kernel.h>
 
--- a/zephyr/wrapper.c
+++ b/zephyr/wrapper.c
@@ -28,7 +28,6 @@
 #include <zephyr/pm/policy.h>
 #include <version.h>
 #include <zephyr/sys/__assert.h>
-#include <soc.h>
 
 LOG_MODULE_REGISTER(zephyr, CONFIG_SOF_LOG_LEVEL);
 


paulstelian97 (approved):

zephyr/lib/alloc.c
```diff
@@ -25,7 +25,9 @@
 #include <zephyr/pm/policy.h>
 #include <version.h>
 #include <zephyr/sys/__assert.h>
+#ifndef CONFIG_IMX93_A55
```
kv2019i, line 28: Hmm, this is a bit iffy, but I don't see existing include guards to check whether soc.h exists for a platform or not, so maybe this is the pragmatic option now. We we'll have more platforms, we need to come up with something cleaner.

kv2019i (approved):
A bit iffy. but ok with this.

tmleman (approved):

zephyr/lib/alloc.c
LaurentiuM1234: Agreed. This doesn't scale very well. This should only be temporary until `soc.h` is introduced (probably when drivers are ported to Zephyr).

LaurentiuM1234: ACK, since there's already 3 people who believe introducing `soc.h` would be a better idea I will submit a Zephyr PR and mark this as a draft until that one gets merged.

LaurentiuM1234:
@kv2019i @lyakh @iuliana-prodan Seems like the usage of `soc.h` is discouraged on Zephyr ARM64 side of things (please see https://github.com/zephyrproject-rtos/zephyr/pull/57071 and https://github.com/zephyrproject-rtos/zephyr/pull/46147). As such, since I can't make a case that `soc.h` is something needed (apart from it solving these build issues) I'd say we go with the ifdef guards unless there's a third option which I'm not seeing ATM.

zephyr/lib/alloc.c
nashif: we should be able to remove this include, not sure why is it needed here, did you try to remove it altogether?

dbaluta:
As @nashif said I think would worth taking a look of why is it necessary. How bad things break if we completely remove it.

zephyr/lib/alloc.c
LaurentiuM1234: > we should be able to remove this include, not sure why is it needed here, did you try to remove it altogether?

Thanks for the suggestion ! I've tried removing all `soc.h` occurrences and it seems fine. Updated PR.

paulstelian97 (approved):
This looks ready, I'd say update pull request title and we're golden on this one.

dbaluta (approved):

zephyr/lib/alloc.c
kv2019i: Even better this way, looks good.

PR #6480, opened 2022-10-25
zephyr: update SYS_INIT calls

gmarull:
Use the new call signature: int (*init_fn)(void);

**NOTE**: This should not be merged before https://github.com/zephyrproject-rtos/zephyr/pull/51217 reaches Zephyr main and this project updates Zephyr revision.

PR #7468, opened 2023-04-19
src: init: init.c: Remove unused sof_init parameter

LaurentiuM1234:
This commit fixes compilation error:

"error: initialization of 'int (*)(void)' from incompatible pointer type 'int (*)(const struct device *)' [-Werror=incompatible-pointer-types]"

from SOF CI Zephyr builds.

### DIFF
--- a/src/init/init.c
+++ b/src/init/init.c
@@ -340,10 +340,8 @@ int sof_main(int argc, char *argv[])
 
 struct device;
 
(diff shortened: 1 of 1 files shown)


paulstelian97:
Dependent on #6480

LaurentiuM1234:
> Dependent on #6480

Thanks a lot! Seems like that PR already fixes the issue so this can be closed.

PR #7470, opened 2023-04-19
schedule: zephyr_dma_domain: Conditionally call interrupt_clear_mask()

LaurentiuM1234:
There's no need to clear IRQ mask on ARM64 architecture as there's currently no way to mask them. This commit aims to fix the compilation issues that would be caused by calling interrupt_clear_mask() on ARM64-based platforms.

### DIFF
--- a/src/schedule/zephyr_dma_domain.c
+++ b/src/schedule/zephyr_dma_domain.c
@@ -32,6 +32,10 @@ LOG_MODULE_DECLARE(ll_schedule, CONFIG_SOF_LOG_LEVEL);
 #define interrupt_disable mux_interrupt_disable
 #endif
 
+#ifdef CONFIG_ARM64
+#define interrupt_clear_mask(irq, bit)
+#endif /* CONFIG_ARM64 */
+
 #define SEM_LIMIT 1
 #define ZEPHYR_PDOMAIN_STACK_SIZE 8192
 


paulstelian97 (approved):

dbaluta (approved):

src/schedule/zephyr_dma_domain.c
```diff
@@ -137,7 +137,9 @@ static void dma_irq_handler(void *data)
 
 	/* clear IRQ */
 	dma_interrupt_legacy(channel, DMA_IRQ_CLEAR);
+#ifndef CONFIG_ARM64
```
iuliana-prodan, line 140: What kind of interrupts are these - edge, level? Are you sure they are not triggered again?

From what I see interrupt_clear_mask goes in wrapper.c in platform_interrupt_clear() which says "handled by zephyr" - so, should these be handled by zephyr and they are not yet?

Is kind of weird to me to not be cleared.

LaurentiuM1234: In the case of ARM64 with GIC the interrupts are cleared after the ISR is executed (as a reference please see `arm_gic_eoi()` from Zephyr and `arch/arm64/core/isr_wrapper.S` ) so there's no need to clear them explicitly.


> From what I see interrupt_clear_mask goes in wrapper.c in platform_interrupt_clear() which says "handled by zephyr" - so, should these be handled by zephyr and they are not yet?

This is not handled by Zephyr as far as I can see:
```
100 #if !CONFIG_IMX                                                                                                                                                                                                
101 
102 void interrupt_mask(uint32_t irq, unsigned int cpu)
103 {
104 »       /* TODO: how do we mask on other cores with Zephyr APIs */
105 }
106 
107 void interrupt_unmask(uint32_t irq, unsigned int cpu)
108 {
109 »       /* TODO: how do we unmask on other cores with Zephyr APIs */
110 }
111 
112 void platform_interrupt_set(uint32_t irq)
113 {
114 »       /* handled by zephyr - needed for linkage */
115 }
116 
117 void platform_interrupt_clear(uint32_t irq, uint32_t mask)
118 {
119 »       /* handled by zephyr - needed for linkage */
120 }
121 #endif

```
Most likely platform_interrupt_clear() is taken from `interrupt-generic.c` or `interrupt-irqsteer.c`. In the case of irqsteer it kinda makes sense that they need to be cleared as irqsteer is not an IRQ controller supported by Zephyr but that's not the case for ARM64's GIC.

> What kind of interrupts are these - edge, level? Are you sure they are not triggered again?

They're level triggered it seems. If they were to be triggered again then we'd have a problem with Zephyr's ISR handling.

iuliana-prodan: Ok, I looked in Zephyr, in suggested files/functions and is clear now. Thanks!

Now, this goes the same as #7467. 
Instead of these ifdefs, can you implement interrupt_clear_mask() for imx93 - empty function or a wrapper? 
As stated [here](https://github.com/thesofproject/sof/pull/7467#discussion_r1172302181), I would keep the consistency for all platforms.

PS: I agree with ifdefs if there are some hw limitations :)

LaurentiuM1234:
Removed the need to use the ifdefs by adding a macro for `interrupt_clear_mask()` for ARM64.

paulstelian97 (approved):
Re-approve, this is much better than the version I previously approved. Good work!

dbaluta (approved):

kv2019i:
Feel free to merge, I'll give @iuliana-prodan to re-review as well.

lgirdwood (approved):

PR #7472, opened 2023-04-19
topology2: cavs-rt5682: remove 24-bit override in mixout-gain-dai-cop…

yongzhi1:
…ier-playback

With 208c2049f ("topology2: deep buffer: use 32 bits to align with other pipeline"), we can remove this override now.


yongzhi1:
Thanks @ranj063 for questioning the override.

plbossart (approved):

ranj063 (approved):

PR #7473, opened 2023-04-19
topology2: tokens: Split the tokens for input/output pin index

ranj063:
Separate the pin index token ID for input/output pin index to avoid collisions between the two and incorrect pin index parsing in the kernel.

Corresponding kernel PR: https://github.com/thesofproject/linux/pull/4311

There is no dependency between the two in terms of merging.

### DIFF
--- a/tools/topology/topology2/include/common/input_audio_format.conf
+++ b/tools/topology/topology2/include/common/input_audio_format.conf
@@ -15,7 +15,7 @@ Class.Base."input_audio_format" {
 	DefineAttribute."instance" {
 	}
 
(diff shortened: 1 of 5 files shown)


bardliao (approved):

paulstelian97 (approved):
Will assume the indentation issues come from Github/the diff format itself, as opposed to being genuine.

ranj063:
> Will assume the indentation issues come from Github/the diff format itself, as opposed to being genuine.

Thanks @paulstelian97. fixed now

kv2019i (approved):

PR #7479, opened 2023-04-20
intel_adsp20: Enable LNL (ACE 2.0) platform in sof

jxstelter:
This patch contains changes required to enable LNL (ACE 2.0) platform in SOF
and bring-up basic interfaces and features.

Required zephyr change:

- [x] https://github.com/zephyrproject-rtos/zephyr/pull/57066 


src/include/sof/lib/dai-zephyr.h
```diff
@@ -76,6 +76,16 @@ struct timestamp_ops {
...
+union hdalink_cfg {
+	uint16_t full;
+	struct {
+		uint16_t lchan	:4;
+		uint16_t hchan	:4;
+		uint16_t stream	:6;
+		uint16_t rsvd	:1;
+		uint16_t dir	:1;
+	} part;
+};
```
lgirdwood, line 88: Does this need to be packed or aligned ?

src/ipc/ipc4/dai.c
```diff
@@ -30,6 +30,49 @@
 
 LOG_MODULE_DECLARE(ipc, CONFIG_SOF_LOG_LEVEL);
 
+int dai_set_link_hda_config(uint16_t *link_config,
+							uint32_t dai_type,
```
lgirdwood, line 34: looks tabs/spaces mixed ?

src/ipc/ipc4/dai.c
```diff
@@ -30,6 +30,49 @@
...
+			 * group size
+			 */
+			link_cfg.part.hchan = (out_fmt->channels_count - 1) / 2;
+		} else {
+			link_cfg.part.hchan = out_fmt->channels_count - 1;
+		}
+		link_cfg.part.stream = common_config->host_dma_config->stream_id;
+		break;
+	default:
+		/* other types of DAIs not handled for now */
```
lgirdwood, line 67: Best to log some failure here since config is coming from host SW.

src/ipc/ipc4/helper.c
```diff
@@ -755,3 +755,22 @@ int ipc4_add_comp_dev(struct comp_dev *dev)
...
 	return IPC4_SUCCESS;
 };
+
+int ipc4_find_dma_config(struct ipc_config_dai *dai, uint8_t *data_buffer, uint32_t size)
+{
+#if defined(CONFIG_ACE_VERSION_2_0)
+	struct ipc_dma_config *dma_config =
+	    (struct ipc_dma_config *)(data_buffer + (size * 4) - sizeof(struct ipc_dma_config));
+	uint32_t *dma_config_id =
+	    (uint32_t *)(data_buffer + (size * 4) - sizeof(struct ipc_dma_config) - 8);
```
lgirdwood, line 765: looks like we are getting some structures from a data blob. Its probably better to have these as static inlines or macros so that they can be reused by others (and mean less chance of others introducing errors)

src/audio/copier/copier.c
```diff
@@ -416,50 +416,69 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
...
+			 *       dma_channel_id;
+			 *       stream_id;
+			 *       dma_stream_channel_map;
+			 *   }
+			 * }
+			 */
+			/* get dma_config if exists */
+			int ret = ipc4_find_dma_config(&dai,
+							(uint8_t *)copier->gtw_cfg.config_data,
+							copier->gtw_cfg.config_length);
```
plbossart, line 453: we still have the problem that the OFLEN bit will not be set for SoundWire/ALH, so the firmware *cannot* program the  channel map in PCMSyCM registers.

It's only been 1.5 months that we talk about this.

plbossart (changes requested):
ok for SSP/DMIC, NAK on SoundWire/ALH

src/audio/copier/copier.c
jxstelter: Yes, You are right. Actually reading of this field is left but data not utilized yet. It could change when EVAD will be supported, but for now I will remove that change.

plbossart: I think you can use the DMA stream_id but the device/channel mapping not really.

jxstelter: Removed

src/ipc/ipc4/helper.c
jxstelter: Actually this is behind the interface specific blob. It was added at the end probably to keep backward compatibility and use the same IPC for all generations. 
Changed to macros.

src/ipc/ipc4/dai.c
jxstelter: Log added

src/ipc/ipc4/dai.c
jxstelter: fixed

src/include/sof/lib/dai-zephyr.h
jxstelter: I think it is not needed.

app/boards/intel_adsp_ace20_lnl.conf
```diff
@@ -0,0 +1,71 @@
...
+
+CONFIG_POWER_DOMAIN=y
+CONFIG_POWER_DOMAIN_INTEL_ADSP=y
+
+# enable Zephyr drivers
+CONFIG_ZEPHYR_NATIVE_DRIVERS=y
+CONFIG_DAI=y
+CONFIG_DAI_INIT_PRIORITY=70
+CONFIG_DAI_INTEL_DMIC=y
+CONFIG_DAI_DMIC_HW_IOCLK=19200000
```
plbossart, line 28: this is only true on development systems.

app/boards/intel_adsp_ace20_lnl.conf
```diff
@@ -0,0 +1,71 @@
...
+CONFIG_PCM_CONVERTER_FORMAT_S16_C32_AND_S24_C32=y
+CONFIG_PCM_CONVERTER_FORMAT_S24_C24_AND_S24_C32=y
+CONFIG_PCM_CONVERTER_FORMAT_S24_C32_AND_S24_C24=y
+CONFIG_PCM_CONVERTER_FORMAT_S16_C32_AND_S16_C32=y
+CONFIG_LOG=y
+CONFIG_LOG_MODE_DEFERRED=y
+CONFIG_LOG_FUNC_NAME_PREFIX_INF=y
+CONFIG_COMP_VOLUME_WINDOWS_FADE=y
+CONFIG_COMP_UP_DOWN_MIXER=y
+CONFIG_SYS_CLOCK_TICKS_PER_SEC=12000
```
plbossart, line 57: not sure what this is?

src/platform/Kconfig
```diff
@@ -204,7 +211,7 @@ endchoice
 
 config MAX_CORE_COUNT
 	int
-	default 4 if TIGERLAKE
+	default 4 if TIGERLAKE || LUNARLAKE
```
plbossart, line 214: that's not correct, please check the numbers of cores in the HAS.

src/platform/lunarlake/include/platform/lib/dai.h
```diff
@@ -0,0 +1,46 @@
...
+/* SSP */
+
+/*
+ * Number of base and extended SSP ports must be defined separately
+ * since some HW registers are in two groups, one for base and one
+ * for extended.
+ */
+
+/** \brief Number of 'base' SSP ports available */
+#define DAI_NUM_SSP_BASE	6
```
plbossart, line 22: nak, this is 3.

src/platform/lunarlake/include/platform/lib/dma.h
```diff
@@ -0,0 +1,23 @@
...
+#ifdef __SOF_LIB_DMA_H__
+
+#ifndef __PLATFORM_LIB_DMA_H__
+#define __PLATFORM_LIB_DMA_H__
+
+/* max number of supported DMA channels */
+#define PLATFORM_MAX_DMA_CHAN	9
+
+/* number of supported DMACs */
+#define PLATFORM_NUM_DMACS	6
```
plbossart, line 15: very odd, we have 9 and 11 above.

src/platform/lunarlake/lib/clk.c
```diff
@@ -0,0 +1,19 @@
...
+// Author: Tomasz Lauda <tomasz.lauda@linux.intel.com>
+//         Janusz Jankowski <janusz.jankowski@linux.intel.com>
+
+#include <sof/common.h>
+#include <rtos/clk.h>
+
+static const struct freq_table platform_cpu_freq[] = {
+	{ 38400000, 38400 },
+	{ 120000000, 120000 },
+	{ CLK_MAX_CPU_HZ, 400000 },
```
plbossart, line 14: I don't think this is true either, or it's a 'notional/nominal' value.

plbossart (changes requested):
couple of misses in the LNL definitions @jxstelter

src/include/sof/audio/ipc-config.h
```diff
@@ -61,22 +101,22 @@ struct ipc_config_src {
...
-					/**< the rate from stream */
-	uint32_t sink_rate;             /**< Define fixed sink rate or */
-					/**< use 0 to indicate need to get */
-					/**< the rate from stream */
+	uint32_t source_rate;	/**< Define fixed source rate or */
+				/**< use 0 to indicate need to get */
+				/**< the rate from stream */
+	uint32_t sink_rate;	/**< Define fixed sink rate or */
+				/**< use 0 to indicate need to get */
+				/**< the rate from stream */
```
lyakh, line 109: can we use this occasion to also fix these to be proper multi-line doxygen?

src/include/sof/audio/ipc-config.h
```diff
@@ -61,22 +101,22 @@ struct ipc_config_src {
...
-					/**< In pull mode the ASRC outputs */
-					/**< a defined number of frames while */
-					/**< number of input frames varies. */
+	uint32_t operation_mode;/**< push 0, pull 1, In push mode the */
+				/**< ASRC consumes a defined number */
+				/**< of frames at input, with varying */
+				/**< number of frames at output. */
+				/**< In pull mode the ASRC outputs */
+				/**< a defined number of frames while */
+				/**< number of input frames varies. */
```
lyakh, line 119: ditto, but please leave at least one space after ';'

kv2019i (changes requested):
I think this looks good to go (no blocking issues aside the ones already pointed out by other reviewers). In general, the src/platform part could be slimmed down further, but I'd not start doing this work in this LNL PR as the actions need to be done on MTL as well (relates to https://github.com/thesofproject/sof/issues/7248 ).

zephyr/CMakeLists.txt
```diff
@@ -757,7 +757,10 @@ find_package(Python3 COMPONENTS Interpreter)
 set(PYTHON3 "${Python3_EXECUTABLE}")
 
 # We don't know why we have this https://github.com/thesofproject/sof/issues/5212
-target_compile_options(SOF INTERFACE -fno-inline-functions)
+# For clang this option disables inlineing
+if(${ZEPHYR_TOOLCHAIN_VARIANT} STREQUAL "xcc")
+   target_compile_options(SOF INTERFACE -fno-inline-functions)
+endif()
```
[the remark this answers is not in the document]
paulstelian97: This is even more interesting, do we need it with the older xcc-based toolchains? Is it needed on the new clang based toolchains that have ZEPHYR_TOOLCHAIN_VARIANT==xt-clang?

But I agree that this PR is not the place to discuss this.

src/include/sof/audio/ipc-config.h
jxstelter: fixed

src/platform/lunarlake/include/platform/lib/memory.h
```diff
@@ -0,0 +1,67 @@
...
+
+/* window 2 */
+#define SRAM_DEBUG_BASE			((uint32_t)(WIN_BASE(2) + WIN2_OFFSET))
+#define SRAM_DEBUG_SIZE			0x800
+
+#define SRAM_EXCEPT_BASE		(SRAM_DEBUG_BASE + SRAM_DEBUG_SIZE)
+#define SRAM_EXCEPT_SIZE		0x800
+
+#define SRAM_STREAM_BASE		(SRAM_EXCEPT_BASE + SRAM_EXCEPT_SIZE)
+#define SRAM_STREAM_SIZE		0x1000
```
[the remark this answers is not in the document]
jxstelter: Yes, most of the definitions from /platform/... directory is not used for IPC4, however lack of this definition causes errors. This should be cleaned up for LNL and MTL as well. In separate PR I think.

src/platform/lunarlake/include/platform/lib/dma.h
jxstelter: Most of the definitions from /platform/... directory is not used for IPC4, however lack of this definition causes errors. This should be cleaned up for LNL and MTL as well. In separate PR I think.
But fixed for now.

src/platform/lunarlake/include/platform/lib/dai.h
jxstelter: fixed

src/platform/Kconfig
jxstelter: There are 5 cores in LNL, however Zephyr do not support more than 4 yet. 
Zephyr PR with fix WIP.

app/boards/intel_adsp_ace20_lnl.conf
```diff
@@ -0,0 +1,71 @@
...
+CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC=19200000
```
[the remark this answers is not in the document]
jxstelter: Added fpga_overlay.

app/boards/intel_adsp_ace20_lnl.conf
jxstelter: Added fpga_overlay.

app/boards/intel_adsp_ace20_lnl.conf
```diff
@@ -53,6 +53,7 @@ CONFIG_LOG_MODE_DEFERRED=y
 CONFIG_LOG_FUNC_NAME_PREFIX_INF=y
 CONFIG_COMP_VOLUME_WINDOWS_FADE=y
 CONFIG_COMP_UP_DOWN_MIXER=y
+CONFIG_COMP_CHAIN_DMA=y
```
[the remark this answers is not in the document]
jxstelter: fixed

zephyr/CMakeLists.txt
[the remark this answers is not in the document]
jxstelter: removed from this PR.

app/boards/intel_adsp_ace20_lnl.conf
plbossart: I don't know how we can generate a 12kHz clock with the existing hardware?

scripts/xtensa-build-zephyr.py
```diff
@@ -70,7 +70,6 @@
 	xtensa_tools_version_postfix = "-unsupportedOS"
 	warnings.warn(f"Your operating system: {py_platform.system()} is not supported")
 
-
```
plbossart, line 73: spurious change?

src/platform/Kconfig
plbossart: that's worthy of a comment...

src/platform/lunarlake/include/platform/lib/dma.h
```diff
@@ -0,0 +1,23 @@
...
+#ifdef __SOF_LIB_DMA_H__
+
+#ifndef __PLATFORM_LIB_DMA_H__
+#define __PLATFORM_LIB_DMA_H__
+
+/* max number of supported DMA channels */
+#define PLATFORM_MAX_DMA_CHAN	11
+
+/* number of supported DMACs */
+#define PLATFORM_NUM_DMACS	4
```
plbossart, line 15: is this one needed?

src/platform/lunarlake/include/platform/platform.h
```diff
@@ -0,0 +1,46 @@
...
+ *  The choice depends on HW features on different platform
+ */
+#define PLATFORM_DEFAULT_CLOCK CLK_SSP
+
+/* Host page size */
+#define HOST_PAGE_SIZE		4096
+
+/* Platform stream capabilities */
+#define PLATFORM_MAX_CHANNELS	8
+#define PLATFORM_MAX_STREAMS	16
```
plbossart, line 32: we have 9+11 HDAudio DMAs, how does this align with 16 here?

src/audio/copier/copier.c
```diff
@@ -404,10 +404,17 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
...
-		dai_index[dai_count - 1] = (dai_index[dai_count - 1] >> 4) & 0xF;
-		dai.type = SOF_DAI_INTEL_SSP;
-		dai.is_config_blob = true;
-		type = ipc4_gtw_ssp;
+		{
+			dai_index[dai_count - 1] = (dai_index[dai_count - 1] >> 4) & 0xF;
+			dai.type = SOF_DAI_INTEL_SSP;
+			dai.is_config_blob = true;
+			type = ipc4_gtw_ssp;
+			int ret = ipc4_find_dma_config(&dai, (uint8_t *)copier->gtw_cfg.config_data,
```
plbossart, line 412: nit-pick: I would declare 'ret' at the function scope level and remove the braces, this would be nicer with no indentation change.

src/audio/copier/copier.c
```diff
@@ -457,9 +464,17 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
...
+			dai.type = SOF_DAI_INTEL_DMIC;
+			dai.is_config_blob = true;
+			type = ipc4_gtw_dmic;
+
+			int ret = ipc4_find_dma_config(&dai, (uint8_t *)copier->gtw_cfg.config_data,
+							copier->gtw_cfg.config_length);
+			if (ret != 0)
+				return -EINVAL;
+			dai.out_fmt = (void *)&(copier->out_fmt);
+		}
```
plbossart, line 477: same here, declare 'ret' at the top level and avoid indentation change.

src/include/sof/audio/ipc-config.h
```diff
@@ -61,22 +101,26 @@ struct ipc_config_src {
...
+					  *  compensates for drift.
+					  */
+	uint32_t operation_mode; /**< push 0, pull 1, In push mode the
+				   *  ASRC consumes a defined number
+				   *  of frames at input, with varying
+				   *  number of frames at output.
+				   *  In pull mode the ASRC outputs
+				   *  a defined number of frames while
+				   *  number of input frames varies.
+				   */
```
plbossart, line 123: remove this change, this is just an indentation change unrelated to this PR, or push it to a separate patch.

plbossart (changes requested):
looks mostly good, couple of nit-picks below

app/boards/intel_adsp_ace20_lnl.conf
kv2019i: @plbossart This is ok, this is just used for Zephyr internal bookkeeping.

src/platform/Kconfig
kv2019i: Ack, please add a comment and ideally a link to a public PR or bug for the issue.

I understood the limitation to 4 cores only affects spinlock debugging, so this is news to be that we are limited.

kv2019i (commented):
Only the core count issue left (please add a comment and/or link to a public bug/pr). Otherwise good for me.

scripts/xtensa-build-zephyr.py
marc-hb: No whitespace changes in this file please, it's fairly active and does not need extra git conflicts.

src/audio/copier/copier.c
jxstelter: fixed

src/audio/copier/copier.c
jxstelter: fixed

scripts/xtensa-build-zephyr.py
jxstelter: fixed

plbossart (changes requested):
minor comments below @jxstelter

marc-hb:
There are two things missing for this to compile in Github Actions:

```
modules/hal/xtensa/zephyr/soc/intel_ace20_lnl/
```

Similar to @aborisovich 's previous https://github.com/zephyrproject-rtos/hal_xtensa/commit/c7fe0872bc3c910068527f0fe841bd4030874e0e for `intel_ace15_mtpm`

This warning:
```
[290/339] Building C object SOF/sof/src/ipc/ipc4/dai.c.obj
SOF/sof/src/ipc/ipc4/dai.c: In function 'dai_set_link_hda_config':
SOF/sof/src/ipc/ipc4/dai.c:39:40: warning: unused variable 'copier_config' [-Wunused-variable]
   39 |         struct ipc4_copier_module_cfg *copier_config =
      |                                        ^~~~~~~~~~~~~
```

marc-hb (changes requested):
This `unused variable 'copier_config'` warning looks trivial after all. I suspected some hard to fix, platform-dependent macrobatics but I found none, it's just unused always.

jxstelter:
> There are two things missing for this to compile in Github Actions:
> 
> ```
> modules/hal/xtensa/zephyr/soc/intel_ace20_lnl/
> ```
> 
> Similar to @aborisovich 's previous [zephyrproject-rtos/hal_xtensa@c7fe087](https://github.com/zephyrproject-rtos/hal_xtensa/commit/c7fe0872bc3c910068527f0fe841bd4030874e0e) for `intel_ace15_mtpm`

PR added: https://github.com/zephyrproject-rtos/hal_xtensa/pull/19

jxstelter:
> minor comments below @jxstelter



> This `unused variable 'copier_config'` warning looks trivial after all. I suspected some hard to fix, platform-dependent macrobatics but I found none, this local variable is just unused always.

fixed

src/ipc/ipc4/dai.c
```diff
@@ -31,6 +31,44 @@
 
 LOG_MODULE_DECLARE(ipc, CONFIG_SOF_LOG_LEVEL);
 
+int dai_set_link_hda_config(uint16_t *link_config,
+			struct ipc_config_dai *common_config,
+			const void *spec_config)
+{
+#if defined(CONFIG_ACE_VERSION_2_0)
+	struct ipc4_audio_format *out_fmt = (struct ipc4_audio_format *) common_config->out_fmt;
```
lyakh, line 39: no need to type-cast to and from `void *`, please, remove everywhere

src/ipc/ipc4/dai.c
```diff
@@ -31,6 +31,44 @@
...
+	struct sof_ipc_dai_config *sof_cfg;
+	union hdalink_cfg link_cfg;
+
+	switch (common_config->type) {
+	case SOF_DAI_INTEL_SSP:
+		sof_cfg = (struct sof_ipc_dai_config *)spec_config;
```
lyakh, line 45: this also drops the `const` qualifier - another reason to avoid type-casts wherever possible

src/ipc/ipc4/dai.c
```diff
@@ -31,6 +31,44 @@
 
 LOG_MODULE_DECLARE(ipc, CONFIG_SOF_LOG_LEVEL);
 
+int dai_set_link_hda_config(uint16_t *link_config,
+			struct ipc_config_dai *common_config,
+			const void *spec_config)
+{
```
lyakh, line 37: isn't this too HDA specific? I'd say it should go in the driver

src/platform/lunarlake/include/platform/drivers/alh.h
```diff
@@ -0,0 +1,19 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright(c) 2022 Intel Corporation. All rights reserved.
+ */
+
+#ifdef __SOF_DRIVERS_ALH__
+
+#ifndef __PLATFORM_DRIVERS_ALH__
+#define __PLATFORM_DRIVERS_ALH__
```
btian1, line 9: if this is platform specific, suggest add LNL in the definition in case of other platform duplicated definition.
__LNL_PLATFORM_DRIVERS_ALH__

src/platform/lunarlake/include/platform/lib/memory.h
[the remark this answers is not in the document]
btian1: When I start SOF, I was confused by those definitions also, IPC4 should not depend on those errors, would you specify the error type? at least, may need a macro to cover this part, those are not used for IPC4 already.

src/platform/lunarlake/include/platform/lib/pm_runtime.h
```diff
@@ -0,0 +1,72 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright(c) 2022 Intel Corporation. All rights reserved.
```
btian1, line 3: new added file, should be 2023

src/platform/lunarlake/include/platform/lib/memory.h
```diff
@@ -0,0 +1,67 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright(c) 2022 Intel Corporation. All rights reserved.
```
btian1, line 3: should be 2023

src/audio/copier/copier.c
```diff
@@ -492,6 +492,11 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
 		dai.type = SOF_DAI_INTEL_SSP;
 		dai.is_config_blob = true;
 		type = ipc4_gtw_ssp;
+		ret = ipc4_find_dma_config(&dai, (uint8_t *)copier->gtw_cfg.config_data,
+				copier->gtw_cfg.config_length);
+		if (ret != 0)
+			return -EINVAL;
+		dai.out_fmt = (void *)&(copier->out_fmt);
```
lyakh, line 499: no need to type-cast

src/audio/copier/copier.c
```diff
@@ -544,14 +549,18 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
 		dai.type = SOF_DAI_INTEL_DMIC;
 		dai.is_config_blob = true;
 		type = ipc4_gtw_dmic;
+
+		ret = ipc4_find_dma_config(&dai, (uint8_t *)copier->gtw_cfg.config_data,
+					copier->gtw_cfg.config_length);
+		if (ret != 0)
+			return -EINVAL;
+		dai.out_fmt = (void *)&(copier->out_fmt);
```
lyakh, line 557: ditto

src/lib/dai.c
```diff
@@ -186,7 +186,11 @@ static void dai_set_device_params(struct dai *d)
 		break;
 	case SOF_DAI_INTEL_ALH:
 		d->dma_dev = DMA_DEV_ALH;
+#ifdef CONFIG_DMA_INTEL_ADSP_GPDMA
 		d->dma_caps = DMA_CAP_GP_LP | DMA_CAP_GP_HP;
+#else
+		d->dma_caps = DMA_CAP_HDA;
+#endif
```
ujfalusi, line 193: Is there a reason to special case ALH and not merge this to the previous patch?

btian1:
code style and CI still have errors, please check and fix.

marc-hb:
> code style and CI still have errors, please check and fix.

Are you referring to checkpatch? If yes can you be more specific? There I see false positives for the most part.

src/include/sof/audio/ipc-config.h
```diff
@@ -9,6 +9,42 @@
...
+					  */
+	uint32_t stream_id;
+	/* DmaStreamChannelMap */
+	struct dma_stream_channel_map channel_map;
+	uint32_t dma_priv_config_size;
+	uint8_t  dma_priv_config[0];
+} __attribute__((packed, aligned(4)));
+
+#define GET_IPC_DMA_CONFIG(buffer, size) \
+	((uint32_t *)((buffer) + ((size) * 4) - sizeof(struct ipc_dma_config)))
```
lyakh, line 45: this definition apparently *requires* that `buffer` is a pointer to 8-bit data, e.g. a `uint8_t *` pointer. Maybe it's better to make it explicit: either typecast to such a type, or make this a `static inline`. Also worth adding, that `size` is in 32-bit dwords.

src/ipc/ipc4/dai.c
```diff
@@ -42,6 +79,12 @@ int dai_config_dma_channel(struct dai_data *dd, struct comp_dev *dev, const void
 		COMPILER_FALLTHROUGH;
 	case SOF_DAI_INTEL_DMIC:
 		channel = 0;
+#if defined(CONFIG_ACE_VERSION_2_0)
+		if (dai->host_dma_config->pre_allocated_by_host)
+			channel = dai->host_dma_config->dma_channel_id;
+		else
+			channel = 0;
```
lyakh, line 86: this is already set above, please, remove

lyakh (commented):
the only remaining request from me is that `dai_set_link_hda_config()` should be implemented in the HDA driver

jxstelter:
> 

The problem with this is that depending on interface (DMIC vs SSP) different fields of IPC4 (Init Instance) needs to be used to calculate the link config value.  If moved to zephyr driver, we need to pass there several parameters instead of single uint16_t value. Therefore I prepare it here and pass to driver which programs HW register.

src/ipc/ipc4/dai.c
kv2019i: @lyakh @jxstelter This is a bit iffy but we do have DAI type specific code for all vendors here and we need to convert from the IPC structs into language of the Zephyr driver interface (e.g. a single link_config field).  At least in my initial review I checked that we need IPC4 structs the driver does not have access to.

@jxstelter Can you double-check could we calculate this in the driver? Direction and stream_id should be available already in driver code -- I guess depth and channel count should be available as well...?

src/ipc/ipc4/helper.c
```diff
@@ -789,3 +789,19 @@ int ipc4_add_comp_dev(struct comp_dev *dev)
...
+	    (struct ipc_dma_config *)GET_IPC_DMA_CONFIG(data_buffer, size);
+	uint32_t *dma_config_id = GET_IPC_DMA_CONFIG_ID(data_buffer, size);
+
+	if (*dma_config_id != GTW_DMA_CONFIG_ID)
+		return IPC4_INVALID_REQUEST;
+
+	dai->host_dma_config = dma_config;
+#endif
+	return IPC4_SUCCESS;
+}
```
aborisovich, line 806: Are we going to re-use ipc4_find_dma_config for MTL or TGL?
If there are no plans for it and this is purely LNL function I would wrap whole function definition with #ifdef (so we do not produce more dummy functions) and also calls to it.

aborisovich (approved):

src/ipc/ipc4/helper.c
plbossart: it's my understanding that the DMA_CONFIG_TLV is only relevant for ACE2+. It's useless for MTL and TGL.

jxstelter:
@plbossart  Pierre, is your change request resolved?

kv2019i:
@lyakh Ok with the response on link_config? This is anyways the current DAI interface in Zephyr, so I'd say even if we'd change it, this would have to be in different pr.

@plbossart Ping?

src/audio/copier/copier.c
```diff
@@ -563,7 +572,7 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
 
 	/* create multi_endpoint_buffer for ALH multi-gateway case */
 	if (dai_count > 1) {
-		int ret = create_endpoint_buffer(parent_dev, cd, config, copier, type, true, 0);
+		ret = create_endpoint_buffer(parent_dev, cd, config, copier, type, true, 0);
 
 		if (ret < 0)
```
ujfalusi, line 577: The blank line is now not needed.

src/ipc/ipc4/dai.c
```diff
@@ -31,6 +31,43 @@
 
 LOG_MODULE_DECLARE(ipc, CONFIG_SOF_LOG_LEVEL);
 
+int dai_set_link_hda_config(uint16_t *link_config,
```
ujfalusi, line 34: this never fails, can be `void`

src/audio/dai-zephyr.c
```diff
@@ -143,6 +143,9 @@ int dai_set_config(struct dai *dai, struct ipc_config_dai *common_config,
 	case SOF_DAI_INTEL_SSP:
 		cfg.type = is_blob ? DAI_INTEL_SSP_NHLT : DAI_INTEL_SSP;
 		cfg_params = is_blob ? spec_config : &sof_cfg->ssp;
+		if (dai_set_link_hda_config(&cfg.link_config,
+					    common_config, cfg_params) < 0)
+			return -EINVAL;
```
ujfalusi, line 148: since `dai_set_link_hda_config()` cannot fail, not much reason to have it return an int and not much point of checking the return value?

src/include/sof/audio/ipc-config.h
```diff
@@ -9,6 +9,42 @@
...
+					  */
+	uint32_t stream_id;
+	/* DmaStreamChannelMap */
+	struct dma_stream_channel_map channel_map;
+	uint32_t dma_priv_config_size;
+	uint8_t  dma_priv_config[0];
+} __attribute__((packed, aligned(4)));
+
+#define GET_IPC_DMA_CONFIG(buffer, size) \
+	((uint32_t *)((uint8_t *)(buffer) + (size) - sizeof(struct ipc_dma_config)))
```
ujfalusi, line 45: why don't you cast it directly to `struct ipc_dma_config *` or to `void *` here?

src/ipc/ipc4/helper.c
```diff
@@ -789,3 +789,19 @@ int ipc4_add_comp_dev(struct comp_dev *dev)
 
 	return IPC4_SUCCESS;
 };
+
+int ipc4_find_dma_config(struct ipc_config_dai *dai, uint8_t *data_buffer, uint32_t size)
+{
+#if defined(CONFIG_ACE_VERSION_2_0)
+	struct ipc_dma_config *dma_config =
+	    (struct ipc_dma_config *)GET_IPC_DMA_CONFIG(data_buffer, size);
```
ujfalusi, line 797: so, this after the preprocessor looks like:
```c
struct ipc_dma_config *dma_config = (struct ipc_dma_config *)((uint32_t *)((uint8_t *)(data_buffer) + (size) - sizeof(struct ipc_dma_config)));
```

src/ipc/ipc4/helper.c
```diff
@@ -789,3 +789,19 @@ int ipc4_add_comp_dev(struct comp_dev *dev)
...
+	uint32_t *dma_config_id = GET_IPC_DMA_CONFIG_ID(data_buffer, size);
+
+	if (*dma_config_id != GTW_DMA_CONFIG_ID)
+		return IPC4_INVALID_REQUEST;
+
+	dai->host_dma_config = dma_config;
```
ujfalusi, line 803:
```c
dai->host_dma_config = GET_IPC_DMA_CONFIG(data_buffer, size);
```
and no need to temporary storage?

lyakh:
> @lyakh Ok with the response on link_config? This is anyways the current DAI interface in Zephyr, so I'd say even if we'd change it, this would have to be in different pr.

@kv2019i yep, just let's address comments from @ujfalusi

marcinszkudlinski (approved):

src/audio/copier/copier.c
```diff
@@ -492,6 +492,13 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
 		dai.type = SOF_DAI_INTEL_SSP;
 		dai.is_config_blob = true;
 		type = ipc4_gtw_ssp;
+		ret = ipc4_find_dma_config(&dai, (uint8_t *)copier->gtw_cfg.config_data,
+					   copier->gtw_cfg.config_length * 4);
+		if (ret != 0) {
+			comp_err(parent_dev, "No host_dma_config found in blob!");
```
ujfalusi, line 498: Is it really host_dma_config and not ssp_dma_config?

src/audio/copier/copier.c
```diff
@@ -544,14 +551,20 @@ static int create_dai(struct comp_dev *parent_dev, struct copier_data *cd,
 		dai.type = SOF_DAI_INTEL_DMIC;
 		dai.is_config_blob = true;
 		type = ipc4_gtw_dmic;
+
+		ret = ipc4_find_dma_config(&dai, (uint8_t *)copier->gtw_cfg.config_data,
+					   copier->gtw_cfg.config_length * 4);
+		if (ret != 0) {
+			comp_err(parent_dev, "No host_dma_config found in blob!");
```
ujfalusi, line 558: Is it really host_dma_config and not dmic_dma_config?

ujfalusi (approved):
Only couple of 'I prefer to write code in a different way' type of things, but those are subjective I guess as no one pointed out those.

I don't have further comments.

kv2019i:
Known error in https://sof-ci.01.org/sofpr/PR7479/build8442/devicetest/index.html and https://sof-ci.01.org/sofpr/PR7479/build8443/devicetest/index.html , proceeding with merge.

marc-hb:
> > This unused variable 'copier_config' warning looks trivial after all. I suspected some hard to fix, platform-dependent macrobatics but I found none, this local variable is just unused always.

> fixed

Somehow it came back. Fixed (again?) in
- #7685

PR #7480, opened 2023-04-20
buffer: fix access without acquisition

lyakh:
In most of the user code buffer control structures should only be accessed after acquisition, not via their raw pointers.

### DIFF
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -962,6 +962,7 @@ static int copier_comp_trigger(struct comp_dev *dev, int cmd)
 	struct comp_dev *dai_copier;
 	struct copier_data *dai_cd;
 	struct comp_buffer *buffer;
+	struct comp_buffer __sparse_cache *buffer_c;
 	uint32_t latency;
 	int ret, i;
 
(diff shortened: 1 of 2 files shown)


paulstelian97 (approved):
This code relies on the buffer/stream sizes not changing when the buffer is not acquired. Right now that should be true (a buffer never changes its size, but new buffers can be created and some can be deleted if topology IPCs occur).

ranj063 (approved):

kv2019i (approved):
Would be nice to have compiler help us catch these things...

lyakh:
> Would be nice to have compiler help us catch these things...

@kv2019i not the compiler, we could make sparse complain about that - it has a "noderef" annotation, but that would be a significant effort to apply those consistently...

lyakh:
> This code relies on the buffer/stream sizes not changing when the buffer is not acquired. Right now that should be true (a buffer never changes its size, but new buffers can be created and some can be deleted if topology IPCs occur).

@paulstelian97 the last set value of `.size` might still be only in cache and we're trying to read from RAM there, which might not contain that value yet. Also, how do you know that it isn't acquired - by another context?

paulstelian97:
> > This code relies on the buffer/stream sizes not changing when the buffer is not acquired. Right now that should be true (a buffer never changes its size, but new buffers can be created and some can be deleted if topology IPCs occur).
> 
> @paulstelian97 the last set value of `.size` might still be only in cache and we're trying to read from RAM there, which might not contain that value yet. Also, how do you know that it isn't acquired - by another context?

Does 'size' change dynamically to the point where this is a genuine issue? You do read size with the lock (I have no qualms with that) and my assertion is that it shouldn't just change on its own once you unlock the buffer. Is that wrong?

src/audio/copier/copier.c
```diff
@@ -962,6 +962,7 @@ static int copier_comp_trigger(struct comp_dev *dev, int cmd)
 	struct comp_dev *dai_copier;
 	struct copier_data *dai_cd;
 	struct comp_buffer *buffer;
+	struct comp_buffer __sparse_cache *buffer_c;
```
RanderWang, line 965: I prefer to move this definition after "struct copier_data *cd = comp_get_drvdata(dev);"

RanderWang (approved):

Issue #7476, opened 2023-04-20
[BUG] toplogy1 : error: tuple SOF_TKN_BUF_SIZE: Numerical result out of range when pipeline period size increases beyond 40ms


Vamshigopal:
**Describe the bug**
when we increase the period size beyond 40ms in PIPELINE_PCM_ADD and DAI_ADD in *.m4 , it fails to build the topology with below error.

ALSA lib data.c:854:(parse_tuple_set) tuple SOF_TKN_BUF_SIZE: Numerical result out of range
ALSA lib parser.c:212:(tplg_load_config) failed to parse topology
Unable to load configuration: Invalid argument

Branch: main
alsatplg version 1.2.8
libasound version 1.2.8
libatopology version 1.2.8

**To Reproduce**

"sof-tgl-max98357a-rt5682\;sof-adl-max98360a-rt5682\;-DCODEC=MAX98360A\;-DFMT=s32le\;-DPLATFORM=adl\;-DLINUX_MACHINE_DRIVER=sof_rt5682\;-DAMP_SSP=1\;-DBT_OFFLOAD\;-**DSPK_MIC_PERIOD_US=500000**"

./scripts/build-tools.sh

**Reproduction Rate**
10/10

**Expected behavior**
Build should work without any errors.

**Impact**
What impact does this issue have on your progress (e.g., annoyance, showstopper)

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel: {SHA}
    * SOF: main
2) Name of the topology file
    * Topology: tools/topology/topology1/sof-tgl-max98357a-rt5682.m4
3) Name of the platform(s) on which the bug is observed.
    * Platform: RPL

**Screenshots or console output**
[tplg-build.log](https://github.com/thesofproject/sof/files/11284640/tplg-build.log)

sathyap-chrome:
@kv2019i  is there any change in RPL-001 or cavs-2.5 release ? I recollect we could change it in older releases

kv2019i:
@Vamshigopal @sathyap-chrome No change that at least I know about. I also tried this locally with our SOF docker image and I could not trigger this error. Will continue to look at this... maybe something in my test procedure.

kv2019i:
[...]

PR #7481, opened 2023-04-20
topology1: fix buffer size calculation if period-size >44ms

kv2019i:
Calculation of SOF_TKN_BUF_SIZE in COMP_PERIOD_FRAMES() macro led to incorrect results with large period size values. For example at 48000Hz sampling rate, period size larger than 44739us would be incorrectly calculated.

This happens as m4 eval does arithmetic in 32bit signed values and multiplication of period size and sampling rate can easily exceed 2^31.

Fix the issue by splitting the arithmetic in steps that fit available value range.

Link: https://github.com/thesofproject/sof/issues/7476

### DIFF
--- a/tools/topology/topology1/m4/buffer.m4
+++ b/tools/topology/topology1/m4/buffer.m4
@@ -44,6 +44,8 @@ dnl COMP_BUFFER_SIZE( num_periods, sample_size, channels, fmames)
 define(`COMP_BUFFER_SIZE', `eval(`$1 * $2 * $3 * $4')')
 
 dnl COMP_PERIOD_FRAMES( sample_rate, period_us)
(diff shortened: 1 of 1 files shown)


kv2019i:
Will backport to v2.2 when merged.

iuliana-prodan (approved):

Vamshigopal (approved):
LGTM

Vamshigopal:
Tested the fix with larger buffer size, build passed without errors.

kv2019i:
V2: 
- minor adjustment to handle 44100 better
- it's still a bit kludgy (e.g. with 11025 and 22050 rates), but best I can come with m4 eval for now

PR #7484, opened 2023-04-20
zephyr: de-static-ify memcpy_s()

marc-hb:
TODO: memset_s(). Others?

memcpy_s() used to be `static` since the dawn of SOF+Zephyr in commit c90055f2f51e ("header: rtos: use rtos specific version of string.h")

This is unlike pre-Zephyr XTOS SOF; I don't know why.

So far it's been working fine but it's now causing problems when trying to link with external .a libraries that assume `memcpy_s` is not static.

Note Zephyr rejected the `memcpy_s()` annex K due to security concerns like https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1969.htm

So this problem may seem superficially related to Zephyr but it's actually not, it is a pure SOF issue.

### DIFF
--- a/zephyr/include/rtos/string.h
+++ b/zephyr/include/rtos/string.h
@@ -34,24 +34,10 @@ static inline int rstrcmp(const char *s1, const char *s2)
 }
 
 /* C library does not have the "_s" versions used by Windows */
-static inline int memcpy_s(void *dest, size_t dest_size,
-			   const void *src, size_t count)
-{
-	if (!dest || !src)
-		return -EINVAL;
+int memcpy_s(void *dest, size_t dest_size,
+	     const void *src, size_t count);
 
-	if ((dest >= src && (char *)dest < ((char *)src + count)) ||
-	    (src >= dest && (char *)src < ((char *)dest + dest_size)))
-		return -EINVAL;
-
-	if (count > dest_size)
-		return -EINVAL;
-
-	memcpy(dest, src, count);
-
-	return 0;
-}
-
+// TODO
 static inline int memset_s(void *dest, size_t dest_size, int data, size_t count)
 {
 	if (!dest)
--- a/zephyr/lib.c
+++ b/zephyr/lib.c
@@ -9,6 +9,25 @@
 #include <stddef.h>
 #include <stdint.h>
 
+/* Duplicate of src/arch/xtensa/include/arch/string.h */
+int memcpy_s(void *dest, size_t dest_size,
+	     const void *src, size_t count)
+{
+	if (!dest || !src)
+		return -EINVAL;
+
+	if ((dest >= src && (char *)dest < ((char *)src + count)) ||
+	    (src >= dest && (char *)src < ((char *)dest + dest_size)))
+		return -EINVAL;
+
+	if (count > dest_size)
+		return -EINVAL;
+
+	memcpy(dest, src, count);
+
+	return 0;
+}
+
 void *__vec_memcpy(void *dst, const void *src, size_t len)
 {
 	return memcpy(dst, src, len);


andyross (approved):
FWIW: the purpose behind the static is that without it (i.e. with a function that has a bare "inline") the function will emit the definition[1] in a regular .text section and not a linkonce one, so you get multiple definition errors if it exists more than once (i.e. lives in a header file that gets included in more than one C file in the build).

Basically it's a mistake in the C standard; you have to use "static inline" if you want to have a function definition in a header.  And yes, this will then collide with competing declarations of the same function with normal linkage.  I have a vague sense that this got fixed in some recent C standard, maybe?  @stephanosio on the Zephyr discord would be the first person I'd ask.

Regardless, this looks harmless and reasonable.  Though the root problem is that we're likely mixing up memcpy_s() implementations somehow and we should probably find those other libc headers as they'll likely cause problems elsewhere too.

[1] Even inline functions get callable assembly emitted for them, in the expectation that the linker will drop it.

marc-hb:
@iganakov I submitted this only as a "demo" only and as space to discuss these problems. I didn't even bother changing `memset_s`.

I don't need this change (unlike you?) so I will abandon this draft PR once the discussion reaches some sort of conclusion.

zephyr/include/rtos/string.h
```diff
@@ -34,24 +34,10 @@ static inline int rstrcmp(const char *s1, const char *s2)
 }
 
 /* C library does not have the "_s" versions used by Windows */
```
lyakh, line 36: to the commit message - sorry, which security concerns do you mean? The link is to a document about the Annex K itself, they do mention several potential security implications, did you mean any specific ones of them?

lyakh (approved):
This would probably also save some memory - I'm currently counting more than 180 calls, some of them also embedded in further inline functions. Performance might suffer though. But maybe we should carefully use `memcpy()` in performance critical sections.

zephyr/include/rtos/string.h
marc-hb: First: this is off-topic. The only point relevant here is: "Zephyr does not have memcpy_s, most likely never will. SOF is on its own".

Now back to your question. https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1969.htm has list of concerns about `memcpy_s()`, including some security concerns. I don't know which concerns in particular the Zephyr developers agree with, this is a question for the Zephyr project. I just heard some of them are listed in this document but I don't know which ones in particular.

abonislawski (approved):

zephyr/include/rtos/string.h
paulstelian97: Without the _exact_ concerns pointed out I cannot in good faith offer a proper approve to this PR. I'm having mixed thoughts on whether memcpy_s is good or bad so I need more convincing to fully approve its removal and replacement with plain memcpy.

softwarecki (approved):

zephyr/include/rtos/string.h
paulstelian97: Then this is too confusing for me to properly review. Why say that memcpy_s is a security concern, yet _don't_ remove it? The commit description and the actual action clash, so at least a different commit message is required.

marc-hb: @paulstelian97 if you set aside (the bad parts of) the commit message for a moment, can you confirm you're OK with the code change?

paulstelian97: The change itself I'm fine if it builds.

marc-hb:
I think there is consensus, @iganakov please resubmit a complete PR with `memset_s` and anything else required if you need this.

PR #7489, opened 2023-04-21
Audio_stream: Add default byte and frame alignment set in init function

andrula-song:
Set byte_align as 1 and frame_align as 1 means no limit on both byte and frame for this stream. Call audio_stream_init_alignment_constants in audio_stream_init, set default align info while buffer created. And every specific component can set its own byte and frame align info in its prepare function later.


dbaluta (approved):

paulstelian97 (approved):
Approving, am curious about the behaviour of the old code. What would the alignment values be set to in the old code? 0 (or some other invalid value)? Uninitialized/left over from previous settings?

andrula-song:
> Approving, am curious about the behaviour of the old code. What would the alignment values be set to in the old code? 0 (or some other invalid value)? Uninitialized/left over from previous settings?

actually the byte align and frame align info are only used for comp_get_copy_limits_frame_aligned API. If we don,t set them in specific component, then the value of them would be zero and get 0 frames available by comp_get_copy_limits_frame_aligned(use shift).
But it is okay if you use comp_get_copy_limits_frame to get the available frames(use division).

andrula-song:
SOFCI TEST

andrula-song:
SOFCI TEST

RanderWang (approved):
LGTM

kv2019i:
PM tests fail on one device in https://sof-ci.01.org/sofpr/PR7489/build6294/devicetest/index.html , but not related to this PR. Upstream Zephyr build fails are a known thing (SYS_INIT interface chance), so all is good, proceeding with merge.

PR #7490, opened 2023-04-21
Tools: Tune: Add to DMIC modes generate alsa-utils export

singalsu:
This patch adds export format for header file that alsa-utils uses. Automatic generation is much more convenient than hand editing when there is update need.


tools/tune/dmic/dmic_batch.m
```diff
@@ -27,25 +27,31 @@
...
 prm.pdm(2).clk_edge = 0;
 prm.pdm(2).skew = 0;
+prm.vfn = {};
 
 %% Design
 prm.pdmclk_min = 2.4e6;
 prm.pdmclk_max = 4.8e6;
 
+%% alsa-utils
+if exist('include_alsa_utils') == 7
```
ShriramShastry, line 37: I believe there is a second argument missing!

tools/tune/dmic/dmic_fir_export_alsa.m
```diff
@@ -0,0 +1,104 @@
...
+rpi = round(fir.rp * 1e2);
+rsi = round(fir.rs);
+
+hfn = sprintf('%s/pdm-decim-fir.h', hdir);
+vfn = sprintf('fir_int32_%02d_%04d_%04d_%03d_%03d', fir.m, pbi, sbi, rpi, rsi);
+sfn = sprintf('pdm_decim_int32_%02d_%04d_%04d_%03d_%03d', fir.m, pbi, sbi, rpi, rsi);
+
+fprintf('Exporting %s ...\n', hfn);
+
+if exist(hfn) == 2
```
ShriramShastry, line 31: I believe there is a second argument missing!

tools/tune/dmic/dmic_init.m
```diff
@@ -55,17 +54,17 @@
 fprintf('Selected fifo_a_fs=%d, fifo_b_fs=%d: ', prm.fifo_a_fs, prm.fifo_b_fs);
 cfg = select_mode(common_mcic_list, a_mfir_list, b_mfir_list, common_clkdiv_list, cfg);
 cfg = get_cic_config(prm, cfg, hw);
-cfg = get_fir_config(prm, cfg, hw, spec);
+[cfg, prm] = get_fir_config(prm, cfg, hw, spec);
```
ShriramShastry, line 57: `cfg` argument is not in use; if you need to return the `cfg` for any reason, please include a note. Thanks

tools/tune/dmic/dmic_init.m
```diff
@@ -137,6 +136,7 @@
 fir.rs = rs;
 fir.m = mfir;
 dmic_fir_export(fir, 'include');
+[~, prm.vfn] = dmic_fir_export_alsa(fir, 'include_alsa_utils', prm.vfn);
 
 end
```
ShriramShastry, line 142: Please think about substituting the phrase `if length(common_mcic_list) > 0` at line 159 with the phrase `isempty(common_mcic_list)`.

ShriramShastry: Please consider changing line 182 with` cfg = get_cic_config(prm, cfg, hw) ` `cfg = get_cic_config( cfg, hw)` as the `prm` variable is not being used.

ShriramShastry: Take into account eliminating the superfluous` semicolon` at lines 193 and 198.

ShriramShastry (commented):
I've included my code syntax review comments. Please consider making a correction.
Many thanks

tools/tune/dmic/dmic_batch.m
singalsu: yes, using 'dir' should be faster so good to do.

tools/tune/dmic/dmic_fir_export_alsa.m
singalsu: yep, useful here too.

tools/tune/dmic/dmic_init.m
singalsu: It remains from Matlab version of dmic_computed.c (in drivers directory). I think I can clean up the unused parts to avoid them to confuse.

tools/tune/dmic/dmic_init.m
singalsu: Thanks, I did that cleanup too. I've checked that none of the exported coefficients change due to these changes, so they were safe to do.

juimonen (approved):
thanks @singalsu! makes me do less errors...

kv2019i:
Good to go @ShriramShastry ?

ShriramShastry (approved):
I like the changes, thanks for making them.

kv2019i:
@singalsu Can you refresh to get a clean CI run and then we can merge...?

singalsu:
> @singalsu Can you refresh to get a clean CI run and then we can merge...?

Done, there's two failures but they can't be due to this patch.

PR #7488, opened 2023-04-21
[SKIP CI] parse_sparse_output.sh: add "removes address space" and "too many"

marc-hb:
Catch the additional two warnings below:
```
warning: too many warnings
warning: cast removes address space
```

PR #7492, opened 2023-04-21
src: drop wrong type-casts to match cache annotations

lyakh:
No need for type-casting: void * is compatible with any pointers and both pointers already have the __sparse_cache annotation. Dropping it causes sparse warnings.


kv2019i (approved):

marc-hb (approved):

marc-hb:
This makes all sparse checks green again (until #7488)

ranj063 (approved):

PR #7477, opened 2023-04-20
host-zephyr: Fix glitches

serhiy-katsyuba-intel:
Fixes glitches when host copier has different container size on source and sink.

PR #7493, opened 2023-04-21
ipc4: Fix host capture audio distortion

serhiy-katsyuba-intel:
Fixes audio distortion when host gateway is used in capture mode with different container size for its source and sink.


src/audio/copier/copier.c
```diff
@@ -1491,6 +1491,13 @@ static int copier_params(struct comp_dev *dev, struct sof_ipc_stream_params *par
 			if (dev->ipc_config.type == SOF_COMP_HOST && !cd->ipc_gtw) {
```
lyakh, line 1491: sorry, could you explain? I thought this condition above meant "not an IPC gateway" and the commit description says "host gateway is used?"

serhiy-katsyuba-intel: That "if" condition means "if Host gateway". Both Host and IPC gateway use SOF_COMP_HOST as a type. Maybe that is wrong and some specific type should be added for IPC gateway, but this is how it is now.

serhiy-katsyuba-intel:
SOF CI TEST

wszypelt:
SOFCI TEST

lgirdwood (commented):
Needs an inline comment.

src/audio/copier/copier.c
lyakh: clarified: in this terminology host and DAI components are called gateways too!

btian1: please check ipc_gtw, if it is true, this part can't be entered with gateway case, it only cover host but not gateway case.

src/audio/copier/copier.c
```diff
@@ -1491,6 +1491,13 @@ static int copier_params(struct comp_dev *dev, struct sof_ipc_stream_params *par
...
 				component_set_nearest_period_frames(dev, params->rate);
 				if (params->direction == SOF_IPC_STREAM_CAPTURE) {
+					enum sof_ipc_frame __sparse_cache dummy, out_ipc_fmt;
+
+					audio_stream_fmt_conversion(cd->out_fmt->depth,
+								    cd->out_fmt->valid_bit_depth,
+								    &dummy, &out_ipc_fmt,
+								    cd->out_fmt->s_type);
+					params->frame_fmt = out_ipc_fmt;
```
RanderWang, line 1500: This change is good but why we have "if (params->direction == SOF_IPC_STREAM_CAPTURE) {" here. It was introduced by host copier optimization.  The params only affect host or dai, why not move it to the entry of copier_params like
````c
static int copier_params(struct comp_dev *dev, struct sof_ipc_stream_params *params)
{
	struct comp_buffer *sink, *source;
	struct comp_buffer __sparse_cache *sink_c, *source_c;
	struct list_item *sink_list;
	int ret = 0;
	int i;

	comp_dbg(dev, "copier_params()");

	memset(params, 0, sizeof(*params));
	params->direction = cd->direction;
	params->channels = cd->config.base.audio_fmt.channels_count;
	params->rate = cd->config.base.audio_fmt.sampling_frequency;

+ if (params->direction == SOF_IPC_STREAM_PLAYBACK) {
+	params->sample_container_bytes = cd->config.base.audio_fmt.depth / 8;
+	params->sample_valid_bytes = cd->config.base.audio_fmt.valid_bit_depth / 8;
+      	params->buffer.size = cd->config.base.ibs;
+ } else {
+      params->buffer.size = cd->config.base.obs;
+     params->sample_container_bytes = cd->out_fmt->depth / 8;
+     params->sample_valid_bytes =
+		cd->out_fmt->valid_bit_depth / 8;
+}
	params->stream_tag = cd->config.gtw_cfg.node_id.f.v_index + 1;
	params->frame_fmt = dev->ipc_config.frame_fmt;
	params->buffer_fmt = cd->config.base.audio_fmt.interleaving_style;


````

**Review by RanderWang — COMMENTED** [review]
more idea

File: src/audio/copier/copier.c (hunk 1 above)
Comment (btian1) [inline src/audio/copier/copier.c:1491]: ”Fixes audio distortion when host gateway is used in capture mode with different container size for its source and sink.”

Again, host gateway case should not enter this condition, it should go to line 1508:

				ret = cd->endpoint[i]->drv->ops.params(cd->endpoint[i],
								       params);

File: src/audio/copier/copier.c [hunk 3]
```diff
@@ -1502,8 +1502,6 @@ static int copier_params(struct comp_dev *dev, struct sof_ipc_stream_params *par
 					/* set up callback */
 					notifier_register(dev, cd->hd->chan,
 							  NOTIFIER_ID_DMA_COPY, copier_dma_cb, 0);
-
-				cd->hd->process = cd->converter[IPC4_COPIER_GATEWAY_PIN];
```
btian1, line 1506: preivous hd->process is based on input and endpoint buffer, after remove endpoint buffer, we have to choose converter[0], since it is based on audio input format and output format.

serhiy-katsyuba-intel: This is the code in host-zephyr.c
```
/* set processing function */
if (params->direction == SOF_IPC_STREAM_CAPTURE)
	hd->process = pcm_get_conversion_function(host_buf_c->stream.frame_fmt,
						  dma_buf_c->stream.frame_fmt);
else
	hd->process = pcm_get_conversion_function(dma_buf_c->stream.frame_fmt,
						  host_buf_c->stream.frame_fmt);
```

host_buf_c is copier sink buffer for playback and copier source for capture. Just container sizes of source and sink formats are used here to select conversion function. And that is a correct behavior for host component.

For 24/32 to 16/16 conversion the code above correctly selects pcm_convert_s32_to_s16(), as 24/32 coming from driver has sample data in upper 24 bits of container, so should be treaded kind of 32/32.

Without fix, that correct pcm_convert_s32_to_s16() is replaced by pcm_convert_s24_to_s16() which is incorrect for host component. That would be correct for non-host component buffer to buffer conversion, but not for audio coming from/going to driver.

ranj063: @RanderWang can you please comment? The converter functions set up in host-zephyr.c is meant for IPC3 isnt it?

ranj063: @serhiy-katsyuba-intel I think this will break a bunch of QB tests. This particular piece of code was specifically added because the default converter function does not work as intended. Maybe the ipc4 converter function has a bug that needs to be fixed instead of removing this completely

serhiy-katsyuba-intel: @ranj063 , seems indeed pcm_func_vc_map[] (used by copier pcm_get_conversion_vc_function()) is incorrect for ipc4_gtw_host and 16 to/from 24 bit conversion.
However, pcm_func_map (used by pcm_get_conversion_function() in host-zephyr) is correct and sufficient for all possible Host gateway formats.
Are those QB tests in CI? Maybe the failures you previously observed were caused by another bug, e.g. https://github.com/thesofproject/sof/pull/7477.

greg-intel:
SOFCI TEST

src/audio/copier/copier.c
ranj063: @serhiy-katsyuba-intel no you can check the QB failures in this PR too

RanderWang: > @RanderWang can you please comment? The converter function set up in host-zephyr.c is meant for IPC3 isnt it?

yes,  the converter function in host-zephyr.c is for IPC3 and it only supports c16/s16 (container size/ sample valid size), c32/s24,  c32/s32. For host gtw we need to support c24/s24 (supported by windows)

serhiy-katsyuba-intel:
Closing this PR in favor of https://github.com/thesofproject/sof/pull/7539

PR #7499, opened 2023-04-24
Audio: EQ-FIR: Optimize the eq_fir_2x_sx functions

andrula-song:
Optimize the eq_fir_2x_sx functions by remove the duplicated usage of circular buffer.


andrula-song:
SOFCI TEST

lgirdwood (commented):
Whats the before/after MCPS ?

andrula-song:
the result from my xtensa testbench is below:
Screenshot from 2023-04-20 14-25-38
the main calculation is in fir_32x16_2x_hifi3( did not find any where to optimize for this function), and this is depend on the coefficient. in this test, I used the mid type, which use 40 taps. Compared with the original version:
for 16 bit format save about 14% cycles for the whole calculation of eq_fir_2x_s16  and save about 37% cycles for eq_fir_2x_s16 without fir_32x16_2x_hifi3

for 24 bit format save about 8% cycles for the whole calculation of eq_fir_2x_s24  and save about 21% cycles for eq_fir_2x_s16 without fir_32x16_2x_hifi3

for 32 bit format save about 9% cycles for the whole calculation of eq_fir_2x_s32  and save about 26% cycles for eq_fir_2x_s16 without fir_32x16_2x_hifi3

src/audio/eq_fir/eq_fir_hifi3.c
```diff
@@ -98,64 +92,62 @@ void eq_fir_2x_s24(struct fir_state_32x16 fir[], struct input_stream_buffer *bso
...
-			d0 = AE_SRAI32R(z0, 8);
-			d0 = AE_SLAI32S(d0, 8);
-			d0 = AE_SRAI32(d0, 8);
+			in = src + ch;
+			out = dst + ch;
+
+			for (i = 0; i < (n >> 1); i += nch) {
+				/* Load two input samples via input pointer x */
+				AE_L32_XP(d0, in, inc_nch_s);
+				AE_L32_XP(d1, in, inc_nch_s);
```
btian1, line 126: this processing is still based on channel data, i.e, try to handle one channel data in this line 123 for loop, then handle another channel, so:

1. Possible handle data based on sample? take 2 channel as example.
    handle L1R1L2R2
    current L1 L2 --> R1 R2
    proposed: L1R1L2R2 --> L3R3L4R4.
you can still handle 2 samples for each channel, this will make load/store fully aligned with 64 bit operation.

another disadvantage with current handle is: once the buffer is big, take deep buffer for example.
the cache miss will cost much more cycles than real, image big buffer have to write back and load new data in cache.

src/audio/eq_fir/eq_fir_hifi3.c
```diff
@@ -33,55 +33,49 @@ void eq_fir_2x_s32(struct fir_state_32x16 fir[], struct input_stream_buffer *bso
...
-		for (i = 0; i < (frames >> 1); i++) {
-			/* Load two input samples via input pointer x */
-			fir_comp_setup_circular(source);
-			AE_L32_XC(d0, x, inc_nch_s);
-			AE_L32_XC(d1, x, inc_nch_s);
-
-			/* Compute FIR */
+	int samples = nch * frames;
+
+	while (samples) {
```
singalsu, line 50: I'm worrying about change of processing order. A run in real device with long FIR would give more confidence that this is better. Maybe also xt testbench could be enough because it has pipelines and scheduler framework from SOF. You could add a test response where left and right channel have different coefficients and make them long. See the FIR EQ examples generator how it's done.

src/include/sof/math/fir_config.h
```diff
@@ -44,7 +44,7 @@
 #if XCHAL_HAVE_HIFI2EP == 1
 #define FIR_HIFIEP	1
 #define FIR_HIFI3	0
-#elif XCHAL_HAVE_HIFI3 == 1
+#elif XCHAL_HAVE_HIFI3 == 1 || XCHAL_HAVE_HIFI4 == 1
```
singalsu, line 47: Good catch! I didn't realize that HIFI4 build does not have HIFI3 set while they are mostly compatible. I  wonder if there's other similar issues in code?

singalsu:
I got with TGL HiFi3 build of xt testbench these MCPS numbers. It includes overhead of scheduler and pipeline, but excludes file components. Command is "process_test('eq-fir', 32, 32, 48000, 0, 'xt-run');".

```
	git main	pr7499
16	14.10		13.59
24	14.43		13.86
32	14.15		13.64

```

src/audio/eq_fir/eq_fir_hifi3.c
andrula-song: each channel can have its own coefficient, and its delay depends on the statistic in this channel, processing by sample will cost much more cycles.

src/include/sof/math/fir_config.h
andrula-song: actually in our config, we enabled HIFI3 while HIFI4 build. Add || XCHAL_HAVE_HIFI4 == 1 is because I'm not sure if the HIFI3 will still enabled while we have HIFI5 or other newer version(but might HIFI4 is enabled). I can  check the other code.

kv2019i (approved):

kv2019i:
@andrula-song Can you check the CI fails? Otherwise seems ready to go.

kv2019i:
Mandatory CIs pass, known fails in https://sof-ci.01.org/sofpr/PR7499/build8482/devicetest/index.html and https://sof-ci.01.org/sofpr/PR7499/build8481/devicetest/index.html. Fuzzer has issues now, please do @marc-hb  @cujomalainey @andyross ping me when I should again start gating items for the fuzzer.

marc-hb:
> Fuzzer has issues now, please do @marc-hb @cujomalainey @andyross ping me when I should again start gating items for the fuzzer.

Main branch is clean, still not clear (to me) what to do about stable-v2.2, see discussion in #7675

sof-docs also needs an update.

cujomalainey:
@marc-hb I made #7709 so we can stop hijacking this PR :)

PR #7501, opened 2023-04-24
ams: fix sparse address space change

lyakh:
When changing pointer type between different sparse address spaces, the change has to be forced to let sparse know, that this is intended and to avoid a warning. Fix such type cases in ams.c.


paulstelian97 (approved):
Looks nice. Does it cover all the warnings or only part of them?

lyakh:
> Looks nice. Does it cover all the warnings or only part of them?

@paulstelian97 all that we see in CI ATM

kv2019i (approved):

kv2019i:
Checkpatch doesn't like tools to be discussed in summary line:
""WARNING: A patch subject line should describe the change not the tool that found it""

I think the summary is ok as it is.

lgirdwood (approved):

PR #7503, opened 2023-04-24
topology1: fix buffer size calculation if period-size >44ms

kv2019i:
Calculation of SOF_TKN_BUF_SIZE in COMP_PERIOD_FRAMES() macro led to incorrect results with large period size values. For example at 48000Hz sampling rate, period size larger than 44739us would be incorrectly calculated.

This happens as m4 eval does arithmetic in 32bit signed values and multiplication of period size and sampling rate can easily exceed 2^31.

Fix the issue by splitting the arithmetic in steps that fit available value range.

Link: https://github.com/thesofproject/sof/issues/7476
(cherry picked from commit be43b4e4ebb9e774a05318c4b54c0445c7938db3)


lgirdwood (approved):
@mwasko @cujomalainey best to cherry-pick for cavs stable branch.

PR #7504, opened 2023-04-24
Remove module adapter persistent buffers

lyakh:
This is a take-over for #7449


src/include/sof/audio/module_adapter/module/generic.h
```diff
@@ -179,13 +179,19 @@ struct processing_module {
 	uint32_t period_bytes; /** pipeline period bytes */
 	uint32_t deep_buff_bytes; /**< copy start threshold */
 	uint32_t output_buffer_size; /**< size of local buffer to save produced samples */
-	struct input_stream_buffer *input_buffers;
-	struct output_stream_buffer *output_buffers;
-	uint32_t num_input_buffers; /**< number of input buffers */
-	uint32_t num_output_buffers; /**< number of output buffers */
+		/**< pointers local input buffers in case of raw data mode */
+	void __sparse_cache *local_input_buffer[PLATFORM_MAX_STREAMS];
+		/**< pointers local output buffer in case of raw data mode */
```
ranj063, line 184: nit-pick, a couple of indentation issues here

ranj063 (approved):
just a minor nit but looks good to go otherwise

jxstelter (approved):

src/include/sof/audio/module_adapter/module/generic.h
lyakh: @ranj063 you mean something like
```
	void __sparse_cache *local_input_buffer[PLATFORM_MAX_STREAMS]; /**< pointers local input
									* buffers in case of raw
									* data mode */
	void __sparse_cache *local_output_buffer[PLATFORM_MAX_STREAMS]; /**< pointers local output
									 * buffer in case of raw
									 * data mode */
```
?

lyakh:
device CI: a single suspend-resume failure https://sof-ci.01.org/sofpr/PR7504/build6391/devicetest/index.html?model=ADLP_RVP_SDW_IPC4ZPH&testcase=check-suspend-resume-with-capture

src/audio/mixer/mixer.c
```diff
@@ -100,8 +100,8 @@ static int mixer_process(struct processing_module *mod,
 	for (i = 0; i < num_input_buffers; i++) {
```
kv2019i, line 100: What's with the "modAPI: " prefix ? Probably not worth changing now as it's in the original comment, but this is fails the "follow existing codebase style" sanity check quite badly.

src/audio/module_adapter/module_adapter.c
```diff
@@ -159,6 +159,8 @@ int module_adapter_prepare(struct comp_dev *dev)
 	struct module_data *md = &mod->priv;
```
kv2019i, line 159: And now "Mod:" ...?

src/audio/module_adapter/module_adapter.c
```diff
@@ -792,19 +778,15 @@ static int module_adapter_simple_copy(struct comp_dev *dev)
...
-		mod->input_buffers[i].consumed = 0;
-	}
-
 	return ret;
 }
 
 int module_adapter_copy(struct comp_dev *dev)
 {
+	struct input_stream_buffer input_buffers[PLATFORM_MAX_STREAMS];
+	struct output_stream_buffer output_buffers[PLATFORM_MAX_STREAMS];
```
kv2019i, line 787: This was noted in the original review, but this could be a stack size issue given stack sizes are separately configurable. At least we have good stack-overrun tools now.

kv2019i (approved):
I dislike the git commit prefixes, but otherwise looks ok.

src/audio/module_adapter/module_adapter.c
lyakh: @kv2019i at least I removed identical arrays from the original version https://github.com/thesofproject/sof/pull/7449/commits/32911d654d8e85719f5121a0b584ab77c3210a1f#diff-8065c18eec2f5b971af80cc5aa42ec413fd9432024dfb0040c676c9dee278f31L700-R683 in `module_adapter_simple_copy()` where that function is called from this one so both are allocated on stack at the same time... But yes, even a single set makes up for 384 bytes...

lyakh:
Let me convert this to a draft to avoid accidental merge. This PR has an inexplicable problem with QB - just like the original #7449 had. They both fail an IPC gateway test. And it's inexplicable because the test doesn't use the module adapter - it builds and tests a minimal copier->copier topology. And both these PRs only modify the module adapter and its users. The failure is a firmware exception in `free_list_add()` called from `sys_heap_free()` from `rfree()` from `zephyr_ll_task_free()`. Since none of the code, modified by this PR, runs in that test, it seems like it uncovers a latent bug elsewhere. Moreover, as verified by @wszypelt (thanks again), this bug doesn't trigger if the same test is run directly after a reset.

lgirdwood:
@lyakh ping - will this be needed after recent copier work from @btian1 and @ranj063  ? (or can be rebased and retested ?)

lyakh:
> @lyakh ping - will this be needed after recent copier work from @btian1 and @ranj063 ? (or can be rebased and retested ?)

@lgirdwood I'd like to get this merged, yes. As you see in the last comments here I stopped working on this because this was triggering some unrelated QB errors. I then asked @marcinszkudlinski if he could take this back (as this was originally his work) but I don't think he had much progress with this either... Then I think some conflicting work was submitted (and possibly merged) that actually moved in the opposite direction - tried to increase the use of persistent storage of buffer pointers. So, yes, I'd like to revisit this at some point... Unless @marcinszkudlinski has now different goals with this?

PR #7507, opened 2023-04-24
platform: move tgl tgl-h to supported platforms

fredoh9:
scripts/xtensa-build-all.sh -a throws error when tgl tgl-h are built with gcc tool chain. New SOF docker image don't have xtensa-cnl-elf-gcc any more. Related change is dc9ba281d76519dbed3670fe3691d0a9aedee85a.

Move tgl tgl-h from DEFAULT_PLATFORMS to SUPPORTED_PLATFORMS. Those platforms can be built if xcc/gcc toolchains are available.

### DIFF
--- a/scripts/xtensa-build-all.sh
+++ b/scripts/xtensa-build-all.sh
@@ -8,7 +8,6 @@ set -e
 # Platforms built and tested by default in CI using the `-a` option.
 # They must have a toolchain available in the latest Docker image.
 DEFAULT_PLATFORMS=(
-    tgl tgl-h
     imx8 imx8x imx8m imx8ulp
     rn rmb
     mt8186 mt8195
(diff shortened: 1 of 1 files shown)


fredoh9:
New SOF docker image is coming https://github.com/thesofproject/sof/pull/7169

marc-hb (approved):

marc-hb:
Strange how MTL is missing from zephyr builds in https://sof-ci.01.org/sofpr/PR7507/build6398/build/index.html
@greg-intel any idea?

greg-intel:
@marc-hb I have to assume it's related to the large number of issues Jenkins experienced today.  Also could be due to a lack of availability.  But that's just a guess, I'm not sure what generates the piece that's missing, but if it's at the end...

marc-hb:
> @marc-hb I have to assume it's related to the large number of issues Jenkins experienced today.

Whatever problems happen, we should never get a build of the same branch with half the platforms

marc-hb:
SOFCI TEST

marc-hb (changes requested):
Change of mind, let's keep the TGL toolchain instead: #7515

Simpler and faster.

PR #6530, opened 2022-11-03
app: zephyr: add CONFIG_ASSERT debug overlay

kv2019i:
Enable CONFIG_ASSERT=y in SOF debug overlay.

Also add a commented out list of additional useful Zephyr debugging
tools to the overlay. These incur a higher runtime cost, so are left
disabled by default. These are all debug tools that have been tested to
work with SOF and found useful.

Dependencies for merging (SOF):

- [x] https://github.com/thesofproject/sof/pull/6823
- [x] https://github.com/thesofproject/sof/pull/6897
- [x] https://github.com/thesofproject/sof/pull/6899
- [x] https://github.com/thesofproject/sof/pull/6900
- [x] https://github.com/thesofproject/sof/pull/6836 (OR https://github.com/thesofproject/sof/pull/6909 )
- [x] https://github.com/thesofproject/sof/pull/7110 (needed for Zephyr baseline update)
- [x] Zephyr baseline update (to bring in all fixes in below fix) -> https://github.com/thesofproject/sof/pull/7125
- [x] rootcause assert hit in mixin_mixout/mixin_mixout.c:368 (https://sof-ci.01.org/sofpr/PR6530/build4058/devicetest/index.html)

Dependencies for merging (Zephyr):
- [x] https://github.com/zephyrproject-rtos/zephyr/pull/53394
- [x] https://github.com/zephyrproject-rtos/zephyr/pull/53395
- [x] https://github.com/zephyrproject-rtos/zephyr/pull/54278
- [x] to merge PR54278, solution is needed for regression observed in https://github.com/thesofproject/sof/pull/6965 that is now blocking Zephyr updates, fix in https://github.com/thesofproject/sof/pull/7044
- [x] https://github.com/zephyrproject-rtos/zephyr/pull/54900

PR #7508, opened 2023-04-24
[SKIP CI] .github/zephyr: build with the debug overlay and CONFIG_ASSERT

marc-hb:
This makes sure Zephyr's -fmacro-prefix-map is working and keeps the
builds reproducible when using a recent enough toolchain.

As found in the CONFIG_ASSERT PR
https://github.com/thesofproject/sof/pull/6530#issuecomment-1482330214
this is not true for old Xtensa toolchains.

### DIFF
--- a/.github/workflows/zephyr.yml
+++ b/.github/workflows/zephyr.yml
@@ -80,6 +80,14 @@ jobs:
           # Temporary testbed for Zephyr development.
           -i IPC4 tgl tgl-h,
         ]
+        build_opts: [""]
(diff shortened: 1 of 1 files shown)


marc-hb:
`-d` builds successfully compared between Windows and Linux in https://github.com/thesofproject/sof/actions/runs/4792504901/jobs/8525051740

paulstelian97 (approved):
This is extremely interesting, if it also works then way to go!

kv2019i (approved):

PR #7516, opened 2023-04-26
lib-manager: fix compilation breakage

lyakh:
Fix breakage introduced by 32e4f7e27a57 ("buffer: use an accessor to read struct audio_stream::end_ptr") which accidentally wrongly used audio_stream_get_end_addr() to access the .end_addr member of struct lib_manager_dma_buf.


marc-hb (approved):

lyakh:
The breakage had no effect on run-time: it's "only" a debug print, so worst case it would print a wrong value, but in fact those `tr_dbg()` debug prints are disabled in all our builds, so this wouldn't even be visible at all.

paulstelian97 (approved):

ujfalusi (approved):

dbaluta (approved):

tmleman (approved):

Issue #7478, opened 2023-04-20
[BUG] Audio playback gets heavily distorted after rapid pause/release on HDA analog output 


ujfalusi:
**Describe the bug**
While testing audio playback on hw:0,0 on UPX-i11 board I have checked pause/resume with aplay -i (holding <SPACE> key to rapid pause/resume) and after  few seconds the audio playback got corrupted.
The playback continued normally, firmware reported no issues, but the audio is completely distorted, high pitched and garbled.

**To Reproduce**

aplay -iv -Dhw:0,0 file.wav (48K 16bit)

Sounds good, no distortion.
Now, press and HOLD the ``<SPACE>`` key to rapid pause/resume aplay.
After few seconds you can hear that the sound starts to be up-shifted pitched, now release <SPACE> and if it is paused, press it again to resume.
Audio is pitched up, distorted, garbled
Not affecting HDMI (ChainDMA).

The same issue present on DeepBuffer PCM (-Dhw:0,31)

**Reproduction Rate**
10/10

**Expected behavior**
audio is not distorted

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel: topic/sof-dev or upstream linux-next
    * SOF: ebaa1429f9d40dc17a28438193789a771cbab2cf
2) Name of the topology file
    * Topology: sof-hda-generic.tplg
3) Name of the platform(s) on which the bug is observed.
    * Platform: TGL (UPX-i11)

**Screenshots or console output**
nothing out of ordinary in logs, all looks error free.

plbossart:
Someone was talking about finding something to keep themselves busy yesterday @ujfalusi.

Humm, wish granted, eh? wow, nice bug, this will be a tough one to crack.

ujfalusi:
Update: it is not the rapid pause/resume, it can be a slow, pasue/resume, small wait, pause/resume.
After the 10th or so iteration the audio starts to be distorted.

ujfalusi:
Update: it is not reproducible with the reference firmware.

ujfalusi:
this is the original tplg path:
![image](https://user-images.githubusercontent.com/12640614/233984194-7df0a93c-9781-4ba5-ab38-2e82bf9a3828.png)


I have tried with these variations:
[...]

PR #7517, opened 2023-04-26
audio: mixin_mixout_generic: normal_mix: Update src/dst after each copy

ujfalusi:
The copying part of the functions need to update the src/dst buffer pointers after each iteration to correctly account for the possible wrap of the buffer.

This was discovered while doing pause - 1sec_wait - resume - 3sec_wait cycles and around the 10th iteration audio became 'metallic' sounding.

Fixes: #7478 

Thanks @ranj063 to test with XCC which is not affected and pointing me to the correct path.

### DIFF
--- a/src/audio/mixin_mixout/mixin_mixout_generic.c
+++ b/src/audio/mixin_mixout/mixin_mixout_generic.c
@@ -56,6 +56,8 @@ static void normal_mix_channel_s16(struct audio_stream __sparse_cache *sink, int
 		nmax = audio_stream_samples_without_wrap_s16(sink, dst);
 		n = MIN(n, nmax);
 		memcpy_s(dst, n * sizeof(int16_t), src, n * sizeof(int16_t));
+		dst += n;
+		src += n;
 	}
 }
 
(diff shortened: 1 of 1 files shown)


kv2019i (approved):

juimonen (approved):

paulstelian97 (approved):

lyakh (approved):
Interesting! If I understand the affected algorithm here correctly, those `normal_mix_*()` functions do mixing and possibly additionally copying. Only if after mixing there are samples for copying (this always happens at least with the first source being mixed in) and when those samples wrap - then we get this bug. So apparently this only happens when pausing

ujfalusi:
> Interesting! If I understand the affected algorithm here correctly, those `normal_mix_*()` functions do mixing and possibly additionally copying. Only if after mixing there are samples for copying (this always happens at least with the first source being mixed in) and when those samples wrap - then we get this bug. So apparently this only happens when pausing

In a simple  case when you have a single path and a single stream then there is  only copying, the mixing is skipped as there is nothing to mix.

andrula-song (approved):

kv2019i:
The testbench test fails to apt fetch issue, Zephyr mtl builds fail due to https://github.com/thesofproject/sof/pull/7516 (now fixed), so that leaves us with https://sof-ci.01.org/sofpr/PR7517/build6472/devicetest/index.html (usual fail) and https://sof-ci.01.org/sofpr/PR7517/build6473/devicetest/index.html (one suspend-resume fail). No fails pointing to this PR, so proceeding with merge.

Issue #3871, opened 2021-02-26
[FEATURE] Allow topologies to manifest and sanity check component IDs


cujomalainey:
**Is your feature request related to a problem? Please describe.**
Speaking with @plbossart he suggested an idea where, with the rise of closed source and a large number of processing components, we could sanity check the topology against the firmware rather than having a cryptic IPC failure when the topology requests a component not in the firmware.

**Describe the solution you'd like**
If we provide an ID manifest we can do simple checks early rather than failing mid load.

**Describe alternatives you've considered**
Keep using what we have

**Additional context**
I think this may become more important as other closed source adopters possibly start generating IDs we will need tools to help identify contents of various blobs when we start possibly getting reports (i might be speculating here)

lgirdwood:
IIRC, the UUIDs for each components are built into an ELF section today. Rimage currently parses all the ELF sections so we could add a command line flag to rimage and tell it to dump UUIDs (including from pre-built images).
Note: this could also be done in a python tool if better, we do have the manifest tool and it may be simpler to add the UUID dump here.

lgirdwood:
Partially implemented in v1.8 as manifest now includes the list of UUIDs that are built into the FW. What's missing now is automatic generation of this data by rimage and use by kernel. Slated for v1.9

lgirdwood:
Looks like this is part of IPC4 work.

RanderWang:
@marc-hb @fredoh9 as I we discussed it in above PR7523, I decided to integrate this feature into tplgtool2.py. As I only study python less than a week, I prefer to get advice from you.  I plan to add a command option like -fw xxx.fw and move the fw parsing code into tplgytool2.py or create a module library for fw parsing ?  Thanks for your feedback

marc-hb:
There are basically 2 options:
1. either you create a brand new file `mynewtool.py` that `import`s `tplgtool2.py`, treated as a library. See demo below. Two files, one new file.
2. or extend `tplgtool2.py` with the new feature(s) you want, adding some new --options. One file, no new file.

Demo of 1., already working today:
```
$ cd sof-test/tools
$ apt-install ipython # more user-friendly

$ ipython3

In [3]: import tplgtool2 # cwd is in the defaut sys.path

In [3]: tplgtool2.TplgBinaryFormat().parse_file('/lib/firmware/intel/sof-tplg/sof-whl-rt5682.tplg')

```

RanderWang:
done

PR #7523, opened 2023-04-27
tools: add a tool to do sanity check for topology and firmware

RanderWang:
With the rise of closed source and a large number of processing components, we could sanity check the topology against the firmware rather than having a cryptic IPC failure when the topology requests a component not in the firmware.

usage:  **tplgfw_check.py [-h] [--version] -t T [T ...] -f F [F ...] [-v]** 
_tplgfw_check.py -t sof-tgl-nocodec.tplg -f basefw.bin_

It is for **[FEATURE] Allow topologies to manifest and sanity check component IDs** https://github.com/thesofproject/sof/issues/3871

Validation case:
**case 1:  Ref FW + ipc4 topology without smart amp and eqiir pass**
rander@rander-i9:~/sof/sof/tools/sof_tplg_fw_sanity_check$ ./tplgfw_check.py -f /home/rander/dsp_fw_mtl_release.bin -t /home/rander/sof-mtl-nocodec.tplg 
FW header Container: 
    id = u'$AE1' (total 4)
    len = 2260
    version_major = 1
    version_minor = 0
    module_entries = 27 

IPC4 supported FW

This FW can support with this topology

**case 2:  Ref FW + ipc4 topology with smart amp and eqiir failed**
rander@rander-i9:~/sof/sof/tools/sof_tplg_fw_sanity_check$ ./tplgfw_check.py -f /home/rander/dsp_fw_mtl_release.bin -t /home/rander/sof/sof/tools/build_tools/topology/topology2/development/sof-mtl-nocodec.tplg 
FW header Container: 
    id = u'$AE1' (total 4)
    len = 2260
    version_major = 1
    version_minor = 0
    module_entries = 27 

IPC4 supported FW

Error: widget: smart_amp.2.1 b'\x1e\x96z\x16\xe4\x8a\xea\x11\x89\xf1\x00\x0c)\xce\x165' is not supported by this FW
Error: widget: eqiir.14.1 b"\xe6\xc0PQ\xf9'\xc8N\x83Q\xc7\x05\xb6B\xd1/" is not supported by this FW

**case 3:  IPC4 cSOF FW + ipc4 topology pass**
rander@rander-i9:~/sof/sof/tools/sof_tplg_fw_sanity_check$ ./tplgfw_check.py -f /home/rander/zephyr/build-sof-staging/sof/community/sof-mtl.ri -t /home/rander/sof/sof/tools/build_tools/topology/topology2/development/sof-mtl-nocodec.tplg 
FW header Container: 
    id = u'$AE1' (total 4)
    len = 1948
    version_major = 1
    version_minor = 0
    module_entries = 19 

IPC4 supported FW

This FW can support with this topology

rander@rander-i9:~/sof/sof/tools/sof_tplg_fw_sanity_check$ ./tplgfw_check.py -f /home/rander/zephyr/build-sof-staging/sof/dbgkey/sof-tgl.ri -t /home/rander/sof/sof/tools/build_tools/topology/topology2/development/sof-tgl-nocodec.tplg 
FW header Container: 
    id = u'$AE1' (total 4)
    len = 1580
    version_major = 1
    version_minor = 0
    module_entries = 15 

IPC4 supported FW

This FW can support with this topology

**case 4:  IPC4 cSOF FW + ipc3 topology failed**
rander@rander-i9:~/sof/sof/tools/sof_tplg_fw_sanity_check$ ./tplgfw_check.py -f /home/rander/zephyr/build-sof-staging/sof/dbgkey/sof-tgl.ri -t /home/rander/sof/sof_tmp/tools/build_tools/topology/topology1/development/sof-tgl-nocodec-ci.tplg 
FW header Container: 
    id = u'$AE1' (total 4)
    len = 1580
    version_major = 1
    version_minor = 0
    module_entries = 15 

IPC4 supported FW

Error: widget: PCM0P b'\x0c\x10\x9d\x8bxm\x8fA\x90\xa3\xe0\xe8\x05\xd0\x85+' is not supported by this FW
Error: widget: PGA1.1 b'~g~\xb7\xf4_\x88A\xaf\x14\xfb\xa8\xbd\xbf\x86\x82' is not supported by this FW
Error: widget: MUXDEMUX1.0 b'hh\xb2\xc40\x14\x0eG\xa0\x89\x15\xd1\xc7\x7f\x85\x1a' is not supported by this FW
......
Error: widget: DMIC0.IN b"'\r\xb0\xc2\xbc\xffPA\xa5\x1a$\\y\xc5\xe5K" is not supported by this FW
Error: widget: PCM5C b'\x0c\x10\x9d\x8bxm\x8fA\x90\xa3\xe0\xe8\x05\xd0\x85+' is not supported by this FW
Error: widget: DMIC1.IN b"'\r\xb0\xc2\xbc\xffPA\xa5\x1a$\\y\xc5\xe5K" is not supported by this FW

**case 5:  IPC3 SOF FW + ipc4 topology failed**
rander@rander-i9:~/sof/sof/tools/sof_tplg_fw_sanity_check$ ./tplgfw_check.py -f /home/rander/Downloads/sof-tgl.ri -t /home/rander/sof/sof/tools/build_tools/topology/topology2/development/sof-tgl-nocodec.tplg 
FW header Container: 
    id = u'XMan' (total 4)
    len = 736
    version_major = 16
    version_minor = 0
    module_entries = 16777216 

IPC3 supported FW

Error: widget: copier.host.1.1 b'\x83\x0c\xa0\x9b\x12\xca\x83J\x94<\x1f\xa2\xe8/\x9d\xda' is not supported by this FW
Error: widget: gain.1.1 b'\xa8\xa9\xbca\xd0\x18\x18J\x8e{&9!\x98\x04\xb7' is not supported by this FW
Error: widget: mixin.1.1 b'\xb2ne9q;I@\x8d?\xf9,\xd5\xc4<\t' is not supported by this FW
......
Error: widget: copier.host.21.1 b'\x83\x0c\xa0\x9b\x12\xca\x83J\x94<\x1f\xa2\xe8/\x9d\xda' is not supported by this FW
Error: widget: gain.21.1 b'\xa8\xa9\xbca\xd0\x18\x18J\x8e{&9!\x98\x04\xb7' is not supported by this FW
Error: widget: mixin.21.1 b'\xb2ne9q;I@\x8d?\xf9,\xd5\xc4<\t' is not supported by this FW

### DIFF
--- a/tools/sof_tplg_fw_sanity_check/tplgfw_check.py
+++ b/tools/sof_tplg_fw_sanity_check/tplgfw_check.py
@@ -1,0 +1,739 @@
+#!/usr/bin/env python3
+
+import argparse
+import construct
+import enum
+import os
+import re
+import sys
+import typing
+from collections import defaultdict
+from construct import this, Container, ListContainer, Struct, Switch, Array, Bytes, GreedyBytes, GreedyRange, FocusedSeq, Pass, Padded, Padding, Prefixed, Flag, Byte, Int16ul, Int32ul, Int64ul, Terminated
+from dataclasses import dataclass
+from functools import cached_property, partial
+
+# Pylint complain about the missing names even in conditional import.
+# pylint: disable=E0611
+if construct.version < (2, 9, 0):
+    from construct import Const, String
+
+    def __enum_dict(enumtype):
+        return dict((e.name, e.value) for e in enumtype)
+    # here is a workaround to make Enum/FlagsEnum accept enum type, like construct 2.9
+    def Enum(subcon, enumtype):
+        return construct.Enum(subcon, default=Pass, **__enum_dict(enumtype))
+    def FlagsEnum(subcon, enumtype):
+        return construct.FlagsEnum(subcon, **__enum_dict(enumtype))
+
+if construct.version >= (2, 9, 0):
+    # https://github.com/construct/construct/blob/master/docs/transition29.rst
+    # Pylint complain about the argument reodering
+    # pylint: disable=E0102,W1114
+    from construct import PaddedString as String
+    from construct import Enum, FlagsEnum
+
+    def Const(subcon, value):
+        return construct.Const(value, subcon)
+
+def CompleteRange(subcon):
+    "like GreedyRange, but will report error if anything left in stream."
+    return FocusedSeq("range",
+            "range" / GreedyRange(subcon),
+            Terminated
+        )
+
+def get_flags(flagsenum: Container) -> "list[str]":
+    "Get flags for FlagsEnum container."
+    return [name for (name, value) in flagsenum.items() if value is True and not name.startswith("_")]
+
+class TplgType(enum.IntEnum):
+    r"""File and Block header data types.
+
+    `SND_SOC_TPLG_TYPE_`
+    """
+    MIXER = 1
+    BYTES = 2
+    ENUM = 3
+    DAPM_GRAPH = 4
+    DAPM_WIDGET = 5
+    DAI_LINK = 6
+    PCM = 7
+    MANIFEST = 8
+    CODEC_LINK = 9
+    BACKEND_LINK = 10
+    PDATA = 11
+    DAI = 12
+
+class VendorTupleType(enum.IntEnum):
+    r"""vendor tuple types.
+
+    `SND_SOC_TPLG_TUPLE_TYPE_`
+    """
+    UUID = 0
+    STRING = 1
+    BOOL = 2
+    BYTE = 3
+    WORD = 4
+    SHORT = 5
+
+class DaiFormat(enum.IntEnum):
+    r"""DAI physical PCM data formats.
+
+    `SND_SOC_DAI_FORMAT_`
+    """
+    I2S = 1
+    RIGHT_J = 2
+    LEFT_J = 3
+    DSP_A = 4
+    DSP_B = 5
+    AC97 = 6
+    PDM = 7
+
+class DaiLnkFlag(enum.IntFlag):
+    r"""DAI link flags.
+
+    `SND_SOC_TPLG_LNK_FLGBIT_`
+    """
+    SYMMETRIC_RATES = 0b0001
+    SYMMETRIC_CHANNELS = 0b0010
+    SYMMETRIC_SAMPLEBITS = 0b0100
+    VOICE_WAKEUP = 0b1000
+
+class DaiMClk(enum.IntEnum):
+    r"""DAI mclk_direction.
+
+    `SND_SOC_TPLG_MCLK_`
+    """
+    CO = 0
+    CI = 1
+
+class DaiBClk(enum.IntEnum):
+    r"""DAI topology BCLK parameter.
+
+    `SND_SOC_TPLG_BCLK_`
+    """
+    CP = 0
+    CC = 1
+
+class DaiFSync(enum.IntEnum):
+    r"""DAI topology FSYNC parameter.
+
+    `SND_SOC_TPLG_FSYNC_`
+    """
+    CP = 0
+    CC = 1
+
+class DaiClockGate(enum.IntEnum):
+    r"""DAI clock gating.
+
+    `SND_SOC_TPLG_DAI_CLK_GATE_`
+    """
+    UNDEFINED = 0
+    GATED = 1
+    CONT = 2
+
+class CtlTlvt(enum.IntEnum):
+    "`SNDRV_CTL_TLVT_`"
+    CONTAINER = 0
+    DB_SCALE = 1
+    DB_LINEAR = 2
+    DB_RANGE = 3
+    DB_MINMAX = 4
+    DB_MINMAX_MUTE = 5
+
+    CHMAP_FIXED = 0x101
+    CHMAP_VAR = 0x102
+    CHMAP_PAIRED = 0x103
+
+class DapmType(enum.IntEnum):
+    "`SND_SOC_TPLG_DAPM_`"
+    INPUT = 0
+    OUTPUT = 1
+    MUX = 2
+    MIXER = 3
+    PGA = 4
+    OUT_DRV = 5
+    ADC = 6
+    DAC = 7
+    SWITCH = 8
+    PRE = 9
+    POST = 10
+    AIF_IN = 11
+    AIF_OUT = 12
+    DAI_IN = 13
+    DAI_OUT = 14
+    DAI_LINK = 15
+    BUFFER = 16
+    SCHEDULER = 17
+    EFFECT = 18
+    SIGGEN = 19
+    SRC = 20
+    ASRC = 21
+    ENCODER = 22
+    DECODER = 23
+
+class PcmFormatsFlag(enum.IntFlag):
+    r"""PCM sample formats.
+
+    `SND_PCM_FMTBIT_LINEAR` and `SND_PCM_FMTBIT_FLOAT`
+    """
+    S8 = 1 << 0
+    U8 = 1 << 1
+    S16_LE = 1 << 2
+    S16_BE = 1 << 3
+    U16_LE = 1 << 4
+    U16_BE = 1 << 5
+    S24_LE = 1 << 6
+    S24_BE = 1 << 7
+    U24_LE = 1 << 8
+    U24_BE = 1 << 9
+    S32_LE = 1 << 10
+    S32_BE = 1 << 11
+    U32_LE = 1 << 12
+    U32_BE = 1 << 13
+
+    FLOAT_LE = 1 << 14
+    FLOAT_BE = 1 << 15
+
+class SofVendorToken(enum.IntEnum):
+    r"""SOF vendor tokens
+
+    See `tools/topology1/sof/tokens.m4` in SOF.
+    """
+    # sof_buffer_tokens
+    SOF_TKN_BUF_SIZE = 100
+    SOF_TKN_BUF_CAPS = 101
+    # sof_dai_tokens
+    SOF_TKN_DAI_TYPE = 154
+    SOF_TKN_DAI_INDEX = 155
+    SOF_TKN_DAI_DIRECTION = 156
+    # sof_sched_tokens
+    SOF_TKN_SCHED_PERIOD = 200
+    SOF_TKN_SCHED_PRIORITY = 201
+    SOF_TKN_SCHED_MIPS = 202
+    SOF_TKN_SCHED_CORE = 203
+    SOF_TKN_SCHED_FRAMES = 204
+    SOF_TKN_SCHED_TIME_DOMAIN = 205
+    SOF_TKN_SCHED_DYNAMIC_PIPELINE = 206
+    # sof_volume_tokens
+    SOF_TKN_VOLUME_RAMP_STEP_TYPE = 250
+    SOF_TKN_VOLUME_RAMP_STEP_MS = 251
+    # sof_src_tokens
+    SOF_TKN_SRC_RATE_IN = 300
+    SOF_TKN_SRC_RATE_OUT = 301
+    # sof_asrc_tokens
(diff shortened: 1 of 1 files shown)


tools/sof_tplg_fw_sanity_check/tplgfw_check.py
```diff
@@ -0,0 +1,739 @@
+#!/usr/bin/env python3
```
kv2019i, line 1: I'd maybe reword "rise of closed source" and just have "With increasing number of processing components...".

tools/sof_tplg_fw_sanity_check/tplgfw_check.py
```diff
@@ -0,0 +1,739 @@
...
+                        if cmd_args.verbose:
+                            print("Widget: ", name, "is supported by ", m.name)
+                        break;
+
+                if found == False:
+                    not_found += 1
+                    print("Error: widget:", name, uuid, "is not supported by this FW")
+
+        if not_found == 0:
+            print("This FW can support with this topology")
```
kv2019i, line 737: maybe: "FW is compatible with the topology."

please also return 0 in case of success and a negative error in case the test fails. then this script can be used as tool by other app/scripts.

ujfalusi:
@RanderWang, I'm not sure what is cryptic about this message:
```
Direct firmware load for intel/avs-lib/tgl/community/5150C0E6-27F9-4EC8-8351-C705B642D12F.bin failed with error -2
Library file 'intel/avs-lib/tgl/community/5150C0E6-27F9-4EC8-8351-C705B642D12F.bin' is missing
failed to find module info for widget eqiir.4.1 with UUID 5150C0E6-27F9-4EC8-8351-C705B642D12F
```
We cannot find the module (not in basefw and not available as loadable) for `eqiir.4.1` which has UUID `5150C0E6-27F9-4EC8-8351-C705B642D12F`

On the other hand:
```
Error: widget: eqiir.14.1 b"\xe6\xc0PQ\xf9'\xc8N\x83Q\xc7\x05\xb6B\xd1/" is not supported by this FW
```
One would have to decipher what that `"\xe6\xc0PQ\xf9'\xc8N\x83Q\xc7\x05\xb6B\xd1/" ` is, how that is an UUID.

Does this tool checks for external libraries? Closed source might rise outside of open source ;)

kv2019i (approved):
That's a cool tool and a welcome addition! I'd run this through pylint (if you haven't already). 

I'm guessing we don't have other Python tools yet to parse the tplg (we have the tplg_parser in C). But I could imagine this could be used as basis for other tools as well.

tools/sof_tplg_fw_sanity_check/tplgfw_check.py
cujomalainey: To be honest though it sounds like a great start to a horror tale for the linux mailing list

cujomalainey (commented):
great tool, only thing is python in my opinion is not great at handling raw bytes. Might be worth checking out this which lets you just build it into the type then pretty print the debug and it does all of its range checks on enums for you https://github.com/cujomalainey/ant-rs/blob/development/ant/src/messages/config.rs#L43

marc-hb (changes requested):
This is a MASSIVE copy/paste/diverge of https://github.com/thesofproject/sof-test/blob/main/tools/tplgtool2.py without even pointing it.

This is python and 2023, there is no reason for so much copy/paste/diverge. Don't do this.

The obvious fix is to convert tplgtool2.py to a library that you can `import` and re-use
https://realpython.com/python-main-function/#a-basic-python-main

> great tool, only thing is python in my opinion is not great at handling raw bytes.

Python alone is not but `construct` used here is pretty good. Also, Python is familiar to many more people (at least for now) and I don't think Rust's low-level performance and memory management are needed here.

fredoh9 (changes requested):
Instead of copying tplgtool2.py here, can you submit a patch in sof-test and use it from sof-test?

marc-hb:
> Instead of copying tplgtool2.py here, can you submit a patch in sof-test and use it from sof-test?

I was about to suggest more or less the same thing. Why can't `sof-test/tplgtool2.py` do both? 1. What it already does now, and 2. this new thing.

lgirdwood:
> > Instead of copying tplgtool2.py here, can you submit a patch in sof-test and use it from sof-test?
> 
> I was about to suggest more or less the same thing. Why can't `sof-test/tplgtool2.py` do both? 1. What it already does now, and 2. this new thing.

Lets move all sof-test stress/smoke/validation tests that are useful for developers to the main sof repo. e.g. to ```sof/test``` directory in the main repo. Having a separate test repo today gives the impression that its for testers only, testing is for developers so lets make it more obvious. CI/validation only tests should remain in sof-test if we have any.

plbossart:
> > > Instead of copying tplgtool2.py here, can you submit a patch in sof-test and use it from sof-test?
> > 
> > 
> > I was about to suggest more or less the same thing. Why can't `sof-test/tplgtool2.py` do both? 1. What it already does now, and 2. this new thing.
> 
> Lets move all sof-test stress/smoke/validation tests that are useful for developers to the main sof repo. e.g. to `sof/test` directory in the main repo. Having a separate test repo today gives the impression that its for testers only, testing is for developers so lets make it more obvious. CI/validation only tests should remain in sof-test if we have any.

@lgirdwood it's not really practical to move everything under the sof/ umbrella. sof-test has different reviewers, should not be tied to releases, has very strong ties to CI, and you would create mutual dependencies between PRs for sof and sof-test within sof.

marc-hb:
> sof-test [...] should not be tied to releases, has very strong ties to CI, and you would create mutual dependencies between PRs for sof and sof-test within sof.

This would indeed make testing and reproducing failures in SOF branches much more complicated. Imagine if SOF git bisect moved the test code at the same time, good luck with that! The only workaround would probably be to git clone sof twice!! Once for the firmware and another clone for the tests. Then most users including CI would probably name the second SOF clone... "sof-test" :-)

A more accurate name for `sof-test` would be `ALSA-and-topologies-test-suite` and ALSA lifecycle's is independent from SOF. Unlike SOF unit tests, the coupling between sof and sof-test is very loose: the Linux kernel and ALSA sit between them.

sof-ci/jenkins has a strong dependency on sof-test but it's only one-way: sof-test has zero dependency on Intel CI. In fact sof-test had contributions from non-Intel Yocto users.

> Lets move all sof-test stress/smoke/validation tests that are useful for developers [...] CI/validation only tests should remain in sof-test if we have any.

I don't know what is a "validation-only" test. Did you mean Intel test _plans_ maybe? There is no test plan in sof-test right now; 100% of the content in sof-test can (and must) be used by all developers using ALSA right now. If you're not already using sof-test then you're not an SOF developer, I mean you're not an SOF developer using Linux and ALSA

> Having a separate test repo today gives the impression that its for testers only, testing is for developers so lets make it more obvious.

If one additional `git clone` command or a mere repo name is all it takes to discourage some developer from reproducing a test failure then... I'm not impressed. But maybe we can help even such test-allergic developer(s) with this simple change:

```diff
--- sof/west.yml
+++ sof/west.yml
@@ -59,6 +59,11 @@ manifest:
           - lz4
           - tinycrypt
 
+    # Linux ALSA and topologies test suite
+    - name: sof-test
+      repo-path: sof-test
+      revision: main
+
   self:
     # Changes to submanifests/*.yml files _are_ effective; these have no
     # specified 'revision:'
```

That's probably where most Linux people have sof-test already cloned anyway.

It would be pointless for Windows developers but it's small.

It is typical to find `west update` in an SOF `git bisect` script, this would have to change to something like `west update zephyr rimage tomlc99` to avoid moving sof-test, maybe not a bad thing.

ranj063:
> @RanderWang, I'm not sure what is cryptic about this message:
> 
> ```
> Direct firmware load for intel/avs-lib/tgl/community/5150C0E6-27F9-4EC8-8351-C705B642D12F.bin failed with error -2
> Library file 'intel/avs-lib/tgl/community/5150C0E6-27F9-4EC8-8351-C705B642D12F.bin' is missing
> failed to find module info for widget eqiir.4.1 with UUID 5150C0E6-27F9-4EC8-8351-C705B642D12F
> ```
> 
> We cannot find the module (not in basefw and not available as loadable) for `eqiir.4.1` which has UUID `5150C0E6-27F9-4EC8-8351-C705B642D12F`
> 
> On the other hand:
> 
> ```
> Error: widget: eqiir.14.1 b"\xe6\xc0PQ\xf9'\xc8N\x83Q\xc7\x05\xb6B\xd1/" is not supported by this FW
> ```
> 
> One would have to decipher what that `b"\xe6\xc0PQ\xf9'\xc8N\x83Q\xc7\x05\xb6B\xd1/"` is, how that is an UUID.
> 
> Does this tool checks for external libraries? Closed source might rise outside of open source ;)

yeah seriously why do we need a new tool for this? The kernel error message is quite self-explanatory that the base FW does not support the module or there's no external library available for it either. I don't see what additional value this tool brings on top what the kernel dmesg already tells us.

marc-hb:
>  I'm not sure what is cryptic about this message:

Error messages can be fixed, this is a digression.

The relevant question is: "Is it valuable to parse a topology file without having to 1. find the corresponding hardware 2. boot a linux kernel or restart the firmware". I don't have the answer but I imagine it could be useful for some larger scale validation?

Another cool feature of `construct` is the ability to edit binaries interactively from the Python shell:
https://construct.readthedocs.io/en/latest/intro.html#example

PS: if this had been submitted as a 10 times smaller patch in sof-test then most people here would have not noticed. Maybe it's not too late :-)

plbossart:
I think there's value in checking that the topology makes references to valid UUIDs.

But this is something that the kernel should do prior to initiating any IPCs.
The fact that we get a cryptic IPC fails is not really problematic IMHO, what's problematic is that we do not detect this problem earlier by cross-checking what the topology wants to use and what the firmware + modules can provide.

btw, 'with the rise of modules', it's impossible to statically check if an UUID is supported or not. The UUID may well well be supported by a module that is different from the base firmware. 

"tplgfw_check.py -t sof-tgl-nocodec.tplg -f basefw.bin" is completely useless IMHO, one would need the list of all additional modules as well.

ranj063:
> I think there's value in checking that the topology makes references to valid UUIDs.
> 
> But this is something that the kernel should do prior to initiating any IPCs. The fact that we get a cryptic IPC fails is not really problematic IMHO, what's problematic is that we do not detect this problem earlier by cross-checking what the topology wants to use and what the firmware + modules can provide.
@plbossart this is exactly what we do. The kernel checks for all necessary module info during topology parsing and if not found, the audio card wont even be registered.

plbossart:
> > I think there's value in checking that the topology makes references to valid UUIDs.
> > But this is something that the kernel should do prior to initiating any IPCs. The fact that we get a cryptic IPC fails is not really problematic IMHO, what's problematic is that we do not detect this problem earlier by cross-checking what the topology wants to use and what the firmware + modules can provide.
> @plbossart this is exactly what we do. The kernel checks for all necessary module info during topology parsing and if not found, the audio card wont even be registered.

that's not really what the log above shows, we attempt to load a module and then fail.

ranj063:
> that's not really what the log above shows, we attempt to load a module and then fail.

No @plbossart . The error message is from https://github.com/thesofproject/linux/blob/1a7de96f85e37b4f879f8ef6d15c337959336c93/sound/soc/sof/ipc4-topology.c#L329.

This function is called during sof_topology_complete() when we make sure that the IPC structures for all widgets in topology are setup

RanderWang:
@plbossart  @ranj063 I may misunderstand this feature : [FEATURE] Allow topologies to manifest and sanity check component IDs https://github.com/thesofproject/sof/issues/3871. In my opinion we need a tool to do static check without driver. If our driver can check these UUID before IPC communication, then we don't need this tool and close the feature requirement ?

RanderWang:
Thanks, I will add this feature to tplgtool2.py directly and close this PR.

PR #7526, opened 2023-04-27
buffer: add and switch to final read and write accessors for struct audio_stream

lyakh:
This makes `struct audio_stream` completely opaque to users, no direct access outside of audio_stream.h and buffer.c should happen to its fields after this.

### DIFF
--- a/src/audio/rtnr/rtnr.c
+++ b/src/audio/rtnr/rtnr.c
@@ -384,10 +384,10 @@ static int rtnr_params(struct comp_dev *dev, struct sof_ipc_stream_params *param
 	cd->sink_stream.channels = audio_stream_get_channels(&sink_c->stream);
 
 	/* set source/sink stream overrun/underrun permitted */
-	cd->sources_stream[0].overrun_permitted = source_c->stream.overrun_permitted;
-	cd->sink_stream.overrun_permitted = sink_c->stream.overrun_permitted;
-	cd->sources_stream[0].underrun_permitted = source_c->stream.underrun_permitted;
-	cd->sink_stream.underrun_permitted = sink_c->stream.underrun_permitted;
+	cd->sources_stream[0].overrun_permitted = audio_stream_get_overrun(&source_c->stream);
+	cd->sink_stream.overrun_permitted = audio_stream_get_overrun(&sink_c->stream);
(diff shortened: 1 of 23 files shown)


src/audio/rtnr/rtnr.c
```diff
@@ -384,10 +384,10 @@ static int rtnr_params(struct comp_dev *dev, struct sof_ipc_stream_params *param
 	cd->sink_stream.channels = audio_stream_get_channels(&sink_c->stream);
 
 	/* set source/sink stream overrun/underrun permitted */
-	cd->sources_stream[0].overrun_permitted = source_c->stream.overrun_permitted;
-	cd->sink_stream.overrun_permitted = sink_c->stream.overrun_permitted;
-	cd->sources_stream[0].underrun_permitted = source_c->stream.underrun_permitted;
-	cd->sink_stream.underrun_permitted = sink_c->stream.underrun_permitted;
+	cd->sources_stream[0].overrun_permitted = audio_stream_get_overrun(&source_c->stream);
```
kv2019i, line 387: I wonder if it would make sense to rename the struct fields (with e.g. "_" prefix) at the same time as you make these changes. This would prevent any future accidental direct use. Or at least one would really have to try hard to use the struct fields directly.

kv2019i (approved):
One improvement idea inline, but that's a more generic note to how to remove the direct accesses (not specific to this PR).

paulstelian97 (approved):

lyakh:
That's a curious failure https://sof-ci.01.org/sofpr/PR7526/build6566/devicetest/index.html?model=ADLP_RVP_SDW_IPC4ZPH&testcase=multiple-pipeline-all

RanderWang (approved):
LGTM

kv2019i:
One PM fail in https://sof-ci.01.org/sofpr/PR7526/build6573/devicetest/index.html 
On PM fail in https://sof-ci.01.org/sofpr/PR7526/build6574/devicetest/index.html
Both known cases, so we can proceed with merging.

src/audio/rtnr/rtnr.c
kv2019i: @lyakh But these will now only be used in the inlined functions in, so the ugliness is very well isolated. Nobody should be using or looking at these fields directly anymore, that's what  we want. But not blocking this PR on this.

PR #7527, opened 2023-04-27
Intel: remove XTOS support

lyakh:
XTOS isn't supported any more for Intel platforms, remove support for it.
If anybody knows what additional code is Intel-XTOS-specific and can be removed, please let me know.


paulstelian97 (changes requested):
Blocking until we can test that nothing breaks on non-Intel XTOS builds (it is not obvious that the changes are Intel-only)

kv2019i (changes requested):
I think the github flows need to be updated here as well, or otherwise there will be many CI failures

lyakh:
> I think the github flows need to be updated here as well, or otherwise there will be many CI failures

@kv2019i yes, apparently PR CI is still attempting TGL XTOS builds, @lgirdwood @marc-hb do we need them? If yes, then obviously this PR cannot be merged

src/arch/xtensa/CMakeLists.txt
```diff
@@ -255,66 +241,11 @@ add_custom_target(sof_post_process
...
-	add_custom_target(process_base_module
-		COMMAND ${CMAKE_COMMAND} -E copy sof-pre sof-${fw_name}
-		DEPENDS prepare_sof_post_process
-	)
-endif()
+# Do nothing / pass-through
+add_custom_target(process_base_module
+	COMMAND ${CMAKE_COMMAND} -E copy sof-pre sof-${fw_name}
+	DEPENDS prepare_sof_post_process
+)
```
marc-hb, line 248: I don't see the point of removing this part of the CMake code, it has absolutely nothing platform specific and we just don't know how companies outside Intel use rimage. I believe the point of this PR is to block everyone from building TGL+XTOS in this branch. This is already a large and significant change, please don't top it up with risky, unnecessary and **platform-independent** CMake code cleanups. Even less so considering the entire project has 0.6 CMake maintainers including myself.

installer/GNUmakefile
```diff
@@ -14,14 +14,11 @@
...
 
-# See rimage/config/*.toml
-SIGNED_list  ?= tgl tgl-h
-
 # To find aliases, try in a Linux kernel git clone:
 #
 #   git grep 'sof-.*\.ri' -- sound/soc/
 
-ALIAS_OTHER_KEY_list += adl adl-s rpl rpl-s
+ALIAS_OTHER_KEY_list += imx8
```
marc-hb, line 21: This line should probably be dropped.

installer/GNUmakefile
```diff
@@ -39,11 +36,7 @@ $(info UNSIGNED_list = ${UNSIGNED_list} )
 $(info SIGNED_list   = ${SIGNED_list}   )
 $(info ALIAS_list    = ${ALIAS_list}    )
 
-target_of_ehl := tgl
-target_of_adl   := tgl
-target_of_adl-s := tgl-h
-target_of_rpl   := tgl
-target_of_rpl-s := tgl-h
+target_of_imx := imx8
```
marc-hb, line 39: Could you try adding one AMD product?

.github/workflows/installer.yml
```diff
@@ -23,8 +23,7 @@ jobs:
       fail-fast: false
       matrix:  # just a vector in this case
         make_env: [
-          IPC_VERSION=,  # default version
-          IPC_VERSION=IPC4 UNSIGNED_list= SIGNED_list='tgl tgl-h',
+          IPC_VERSION= UNSIGNED_list='imx8' SIGNED_list=,  # default version
```
marc-hb, line 26: @dbaluta, @andyross , is anyone outside Intel looking at IPC4? With or without Zephyr?

marc-hb (changes requested):

.github/workflows/installer.yml
cujomalainey: @bhiregou @bhiregoudar @yaochunhung

installer/GNUmakefile
lgirdwood: > > I think the github flows need to be updated here as well, or otherwise there will be many CI failures
> 
> @kv2019i yes, apparently PR CI is still attempting TGL XTOS builds, @lgirdwood @marc-hb do we need them? If yes, then obviously this PR cannot be merged

yep - lets make sure we update all flows. CI should not be doing any xtos builds for Intel platforms as Intel only uses Zephyr now..

.github/workflows/installer.yml
paulstelian97: Right now i.MX8 IPC4 support is a discussion topic. I am not committing to doing it soon, BUT it is in my backlog, I am just lacking some information related to the firmware side of this conversion (for the kernel I have some patches ready, minus exact filenames)

I do expect Zephyr support to be done before the IPC4 migration.

installer/GNUmakefile
lyakh: > Could you try adding one AMD product?

@marc-hb can we make it a follow-up? I'd be happy to get this right with a minimum changes

.github/workflows/installer.yml
marc-hb: > I do expect Zephyr support to be done before the IPC4 migration.

Thanks, this is the information relevant to this line. It means this Github Action will never need to build IPC4+XTOS ever again so this line can be removed as @lyakh currently does.

src/arch/xtensa/CMakeLists.txt
marc-hb: > nobody else uses the bootloader - check CONFIG_XT_BOOT_LOADER

Sure but what does removing this particular CMake code achieves? When the Linux kernel code removes support for old hardware it's neither for the purpose of saving disk space nor for aesthetics nor for "ideology". It's because old code gets in the way of updating APIs and testing changes. In other words Linux kernel code is removed _when it solves an actual, specific maintenance problem_. I don't see what specific problem is being solved by removing this CMake code here now.

I really want to minimize CMake churn as much as possible and considering I'm the only one (barely) maintaining sof/cmake code I think I deserve this. As a coincidence I've just performed some CMake archeology for #7546 and deleted CMake code there would have make that archeology even more painful that it already was. By the way we have many more rimage-related changes in flight right now.

Feel free to a "this is dead code now since..." comment, that would be useful.

Quoting from #7552 that I just submitted, #7552 and 34d7baa80061 similarly revert some massive search/replace that backfired:

> If we want to stop supporting something then let's keep things simple and remove it only where it actually gets in the way and where it actually becomes a problem. No need to search and scrub every reference to it from the face of the planet and accidentally make our life harder on release branches, create pointless git conflicts and increase maintenance costs when trying to reduce them!

marc-hb:
https://github.com/thesofproject/sof/actions/runs/4831781226/jobs/8609674416?pr=7527 just needs some adjustments to the expected output

src/arch/xtensa/CMakeLists.txt
lyakh: @marc-hb sorry, TBH I don't understand your objection here. We are removing the only use case for this block of code. If we keep it we'll keep stumbling over it trying to identify what it's there for, until someone removes it. Why - simple - to reduce maintenance costs. I might miss some further dependencies, breakage possibilities or whatever. But here it seems pretty clear to me - that `if` branch will never be entered. So it can be removed.

marc-hb: >  Why - simple - to reduce maintenance costs.

I'm the only one doing CMake maintenance right now and I gave you examples. I'm sorry you don't understand them.

marc-hb: > If we keep it we'll keep stumbling over it trying to identify what it's there for, until someone removes it. Why - simple - to reduce maintenance costs.

"we" has been only myself for the last few years since Janusz left. So please leave the estimation of maintenance costs to me and believe me when I say that I don't do it for fun.

lyakh:
SOFCI TEST

lyakh:
CI: two suspend-resume failures in https://sof-ci.01.org/sofpr/PR7527/build6996/devicetest/index.html - unrelated

kv2019i (approved):
Thanks @lyakh , passing CI now.

juimonen (approved):

kv2019i:
Thanks @lyakh , merging!

PR #7520, opened 2023-04-26
zephyr: use system cache API

kv2019i:
Use zephyr cache APIs instead of xtensa specific ones.

PR #7528, opened 2023-04-27
ipc4: Handle typing errors in component retrieval

andyross:
In IPC4, the component ID is entirely under the control of the input protocol, and as components of multiple types exist in the list, it's possible for protocol messages to fetch an incorrectly typed object, leading to crashes at runtime.

Force all usage to go through a get_comp() utility that does type checking.  There was already a similar utility for the special case of pipelines (which have a separate ID to query).

### DIFF
--- a/src/ipc/ipc4/helper.c
+++ b/src/ipc/ipc4/helper.c
@@ -135,6 +135,16 @@ struct comp_dev *comp_new_ipc4(struct ipc4_module_init_instance *module_init)
 	return dev;
 }
 
+static struct ipc_comp_dev *get_comp(struct ipc *ipc, uint16_t type, uint32_t id)
+{
+	struct ipc_comp_dev *c = ipc_get_comp_by_id(ipc, id);
+
+	if (c && c->type == type)
+		return c;
+
+	return NULL;
+}
+
 struct ipc_comp_dev *ipc_get_comp_by_ppl_id(struct ipc *ipc, uint16_t type, uint32_t ppl_id)
 {
 	struct ipc_comp_dev *icd;
@@ -283,7 +293,7 @@ int ipc_pipeline_free(struct ipc *ipc, uint32_t comp_id)
 	int ret;
 
 	/* check whether pipeline exists */
-	ipc_pipe = ipc_get_comp_by_id(ipc, comp_id);
+	ipc_pipe = get_comp(ipc, COMP_TYPE_PIPELINE, comp_id);
 	if (!ipc_pipe)
 		return IPC4_INVALID_RESOURCE_ID;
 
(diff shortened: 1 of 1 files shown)


andyross:
crash case found via fuzzing, IPC4 fuzz updating coming in an hour or two once I'm sure it's shaken out

cujomalainey (approved):

andyross:
Seems like the build failures are continuing fallout of the cache API mishap fixed in https://github.com/zephyrproject-rtos/zephyr/pull/57298 I think?  That's merged now in Zephyr, not sure how it propagates to the SOF CI.

src/ipc/ipc4/helper.c
```diff
@@ -135,6 +135,16 @@ struct comp_dev *comp_new_ipc4(struct ipc4_module_init_instance *module_init)
 	return dev;
 }
 
+static struct ipc_comp_dev *get_comp(struct ipc *ipc, uint16_t type, uint32_t id)
+{
+	struct ipc_comp_dev *c = ipc_get_comp_by_id(ipc, id);
+
+	if (c && c->type == type)
```
ranj063, line 142: @andyross assuming there are 2 different types of components with the same ID, wouldnt this always fail if the first component in the list is not for the specified type even if  the other one exists in the list?

andyross: It would, but by my reading that would be a bug in the design?  The list as currently used is already populated with polymorphic components, and already assuming that the first entry with a matching ID is the one desired.  Are there circumstances where it would be possible to have duplicate IDs?

andyross: Ah...  OK, so if the component ID is input to the protocol, this actually is something subject to injected bugs that we'll need to address, I think.  I'll take a look, could be as simple as traversing the list and erroring out at creation time.

andyross: Indeed, I see it now.  You're absolutely right, the initial component ID is set by the kernel, not the firmware.  So there needs to be protection in the firmware against duplicates, I'll see what I can work up and attach it to this PR, as it's related.  The fix here is protecting us against the results of the duplication (and even if there aren't duplicates, this remains an externally triggerable crash which we have to fix), but there is a rooter-cause that needs to be addressed too.

marc-hb:
> Seems like the build failures are continuing fallout of the cache API mishap fixed in https://github.com/zephyrproject-rtos/zephyr/pull/57298 I think?

Yes https://github.com/thesofproject/sof/actions/runs/4822123309/jobs/8588936882?pr=7528

> That's merged now in Zephyr, not sure how it propagates to the SOF CI.

Please review Zephyr uprev in sof/west.yml:
- #7520

kv2019i (approved):
While should not happen with the actual kernel driver, agreed this needs to be checked.

ranj063:
> While should not happen with the actual kernel driver, agreed this needs to be checked.

yeah we can deal with the check in a follow up PR.

andyross:
Update as described to add a fix for the creation of duplicate components too.  IPC3 didn't need it, it already had checking.

(Also: sorry for the churn -- pushed the whole fuzz branch on top of this one by mistake)

PR #7529, opened 2023-04-27
Tools: Topology2: SRC format fixes

singalsu:
This patch avoids with development topology sof-hda-src-generic.tplg the playback start failure and error seen in trace "buffer: buffer_alloc(): new size = 0 is invalid".

The kernel selects in case of multiple output formats defined the format that matches the input. It then caused the obs to become zero. When the SRC is used in the mixin-based playback pipeline, its output format is limited to 32-bits format only. So replace the multiple output formats with a single 48K, 2ch 32-bit format.

### DIFF
--- a/tools/topology/topology2/include/components/src_format.conf
+++ b/tools/topology/topology2/include/components/src_format.conf
@@ -1,369 +1,238 @@
 #src format array
 			num_input_audio_formats 42
-			num_output_audio_formats 42
 
-			# 8khz input
-			Object.Base.audio_format.1 {
-				in_rate                 8000
-				in_bit_depth            16
-				in_valid_bit_depth      16
-				out_bit_depth           32
-				out_valid_bit_depth     24
+			Object.Base.input_audio_format [
+				# 8khz input
+				{
+					in_rate                 8000
+					in_bit_depth            16
+					in_valid_bit_depth      16
 				}
+				{
+					in_rate                 8000
+					in_bit_depth            32
+					in_valid_bit_depth      24
+				}
+				{
+					in_bit_depth            32
+					in_valid_bit_depth      32
+				}
+				# 11.025 khz input
+				{
+					in_rate                 11025
+					in_bit_depth            16
+					in_valid_bit_depth      16
+				}
+				{
+					in_rate                 11025
+					in_bit_depth            32
+					in_valid_bit_depth      24
+				}
+				{
+					in_rate                 11025
+					in_bit_depth            32
+					in_valid_bit_depth      32
+				}
+				# 12khz input
+				{
+					in_rate                 12000
+					in_bit_depth            16
+					in_valid_bit_depth      16
+				}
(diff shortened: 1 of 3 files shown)


ranj063 (approved):

tools/topology/topology2/include/components/src_format.conf
```diff
@@ -1,369 +1,240 @@
...
+				}
+				{
+					in_rate                 8000
+					in_bit_depth            32
+					in_valid_bit_depth      24
+				}
+				{
+					in_bit_depth            32
+					in_valid_bit_depth      32
+				}
```
plbossart, line 19: is there any value in having all these 3 formats? In what case would the first and second be used? Usually there's a copier that will convert to S32_LE, no?

RanderWang (approved):

tools/topology/topology2/include/components/src_format.conf
singalsu: Yes, we could narrow the scope of formats. I was also thinking to rename this to have "playback" in .conf file name. With formats reduced maybe src_fomat_host_mixin_playback.conf, or something. I will need other version for capture. What do you think?

tools/topology/topology2/include/components/src_format.conf
```diff
@@ -1,369 +1,240 @@
...
+
+			num_output_audio_formats 1
+
+			Object.Base.output_audio_format [
+				{
+					out_rate                 48000
+					out_bit_depth            32
+					out_valid_bit_depth      32
+					# for 11.025k, 22.05k, 44.1k, 88.2k and 176.4k, extra 4 sample size is needed
+					obs "$[($out_channels * (($[($out_rate + 999)] / 1000) + 4)) * ($out_bit_depth / 8)]"
```
ujfalusi, line 238: but this is 48k?

kv2019i: right, so this can be simplified now

kv2019i (approved):
The obs calc can be simplified now, but otherwise looks good.

tools/topology/topology2/include/components/src_format.conf
singalsu: In conversion from some 44.1 -> 48 kHz the internal block sizes are not multiple of 1ms so I kept this. But in a quick test it ran also OK with with +0 instead of +4. If it looks safe to omit with more tests I'll do it. It also helps to get more 2^N like buffers sizes than might help with allocation.

singalsu: I'm removing this like, seems it's not needed.

kv2019i:
PM fails in https://sof-ci.01.org/sofpr/PR7529/build6861/devicetest/index.html and in https://sof-ci.01.org/sofpr/PR7529/build6862/devicetest/index.html . Not blocking merge.

PR #7531, opened 2023-04-27
IPC4 fuzzing support

andyross:
Support for IPC4 protocol fuzzing with the existing rig.  Just select CONFIG_IPC_MAJOR_4=y and it will adjust as needed.  It seems to have pretty decent coverage as far as I've been able to check manually, has run on my system for about two hours now (running 64 instances in parallel), and has uncovered one bug (in #7528 -- that will obviously want to merge before this does).

@marc-hb I didn't adjust the existing fuzz.sh script in this PR, it's still doing IPC3 by default.  There are lots of ways to handle this: you could run the fuzzer twice in CI for each protocol, we could add new "posix" platform variants for each, etc...  Let me know what you'd prefer.

### DIFF
--- a/src/audio/base_fw.c
+++ b/src/audio/base_fw.c
@@ -18,6 +18,11 @@
 #include <zephyr/drivers/counter.h>
 #endif
 
+/* TODO: Remove platform-specific code, see https://github.com/thesofproject/sof/issues/7549 */
+#if defined(CONFIG_SOC_SERIES_INTEL_ACE) || defined(CONFIG_INTEL_ADSP_CAVS)
+# define INTEL_ADSP 1
+#endif
+
 LOG_MODULE_REGISTER(basefw, CONFIG_SOF_LOG_LEVEL);
 
 /* 0e398c32-5ade-ba4b-93b1-c50432280ee4 */
@@ -197,8 +202,13 @@ static int basefw_mem_state_info(uint32_t *data_offset, char *data)
 	index = 0;
 	tuple_data[index++] = info.free_phys_mem_pages;
 	tuple_data[index++] = info.ebb_state_dword_count;
-	for (i = 0; i < info.ebb_state_dword_count; i++)
+	for (i = 0; i < info.ebb_state_dword_count; i++) {
+#ifdef INTEL_ADSP
 		tuple_data[index + i] = io_reg_read(SHIM_HSPGCTL(i));
+#else
+		tuple_data[index + i] = 0;
+#endif
+	}
 	index += info.ebb_state_dword_count;
 
 	tuple_data[index++] = info.page_alloc_struct.page_alloc_count;
@@ -222,7 +232,11 @@ static int basefw_mem_state_info(uint32_t *data_offset, char *data)
 	index = 0;
 	tuple_data[index++] = info.free_phys_mem_pages;
 	tuple_data[index++] = info.ebb_state_dword_count;
+#ifdef INTEL_ADSP
 	tuple_data[index++] = io_reg_read(LSPGCTL);
+#else
+	tuple_data[index++] = 0;
+#endif
 	tuple_data[index++] = info.page_alloc_struct.page_alloc_count;
 	ptr = (uint16_t *)(tuple_data + index);
 	for (i = 0; i < info.page_alloc_struct.page_alloc_count; i++)
--- a/src/audio/base_fw.c
+++ b/src/audio/base_fw.c
@@ -11,6 +11,7 @@
 #include <sof/lib/cpu-clk-manager.h>
 #include <sof/lib/cpu.h>
 #include <rtos/init.h>
+#include <platform/lib/clk.h>
 
 #if CONFIG_ACE_V1X_ART_COUNTER || CONFIG_ACE_V1X_RTC_COUNTER
 #include <zephyr/device.h>
--- a/src/ipc/ipc4/helper.c
+++ b/src/ipc/ipc4/helper.c
@@ -20,8 +20,14 @@
 #include <sof/platform.h>
 #include <sof/schedule/ll_schedule_domain.h>
 #include <rtos/wait.h>
-#ifdef __ZEPHYR__
-#include <adsp_memory.h> /* for IMR_BOOT_LDR_MANIFEST_BASE */
+
+/* TODO: Remove platform-specific code, see https://github.com/thesofproject/sof/issues/7549 */
+#if defined(CONFIG_SOC_SERIES_INTEL_ACE) || defined(CONFIG_INTEL_ADSP_CAVS)
+#define RIMAGE_MANIFEST 1
+#endif
+
+#ifdef RIMAGE_MANIFEST
+#include <adsp_memory.h>
 #endif
 
 #include <rtos/sof.h>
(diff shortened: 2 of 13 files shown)


src/ipc/ipc4/helper.c
```diff
@@ -21,8 +21,9 @@
 #include <sof/schedule/ll_schedule_domain.h>
 #include <rtos/wait.h>
 
-#define RIMAGE_MANIFEST (defined(__ZEPHYR__) && \
-	(defined(CONFIG_SOC_SERIES_INTEL_ACE) || defined(CONFIG_INTEL_ADSP_CAVS)))
+#if defined(CONFIG_SOC_SERIES_INTEL_ACE) || defined(CONFIG_INTEL_ADSP_CAVS)
+#define RIMAGE_MANIFEST 1
```
cujomalainey, line 25: collapse this into the commit that adds this?

cujomalainey (commented):
mostly pretty good, just mostly patch cleanup

cujomalainey:
Something must be weird with the state, you pushed the PR 1h ago, the commits 2 days ago, and my review was 1min ago but yet is marked as "outdated"....

marc-hb:
> it's still doing IPC3 by default. There are lots of ways to handle this: you could run the fuzzer twice in CI for each protocol, we could add new "posix" platform variants for each, etc... Let me know what you'd prefer.

Can you please try something like this? (UNTESTED)
`yamllint` is your friend! Cause force-pushing every small indentation fix is NOT fun :-)

I didn't expect any shell script change, pretty sure I anticipated this PR. Is there any?


```diff
--- a/.github/workflows/ipc_fuzzer.yml
+++ b/.github/workflows/ipc_fuzzer.yml
@@ -42,15 +42,14 @@ jobs:
           fuzz-seconds: 5
 
 
-  # TODO, to add IPC4 support fix compilation of:
-  #
-  #    ./scripts/fuzz.sh -t 1 -- -DCONFIG_IPC_MAJOR_4=y
-  #
-  # ... or of some other _IPC4_ -DOVERLAY_CONFIG=
-  #
-  # Then use a simple IPC3/IPC4 matrix like the one in zephyr.yml
   simple-IPC3-fuzz_sh:
     runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        # Keep these names short due to questionable Github UI choices
+        IPC: [ IPC3, IPC4 ]
+
     steps:
       - name: add i386 arch
         run: |
@@ -78,4 +77,8 @@ jobs:
           cd workspace
           clang --verbose
           set -x
-          sof/scripts/fuzz.sh -o _.log -t 300
+          case '${{ matrix.IPC }}' in
+            IPC3) cmake_arg='-DCONFIG_IPC_MAJOR_3=y'
+            IPC4) cmake_arg='-DCONFIG_IPC_MAJOR_4=y'
+          esac
+          sof/scripts/fuzz.sh -o _.log -t 300 -- "$cmake_arg"
```

andyross:
Update, merge two patches which were vestigial versions of the same fix (this sat around in my tree for a while).  Clean up the checkpatch warnings.  Also filed bug #7549 to track the platform-dependent code discussion.

I'm too timid to muck with CI files in the same PR, so no change to the workflow yaml here.  But I did validate that fuzz.sh works in place if you just pass the kconfig (sorry, didn't realize you'd left the "`--`" syntax trick in the command line there)

andyross:
Remaining checkpatch issue is a false positive.  It's not possible to non-trivially use memcpy_s in that circumstance.  The `comp_data` destination buffer is just a void* on the main IPC object, it doesn't have a build-time (or runtime) checkable size, we just assume it can hold one message.

src/audio/base_fw.c
```diff
@@ -197,8 +201,13 @@ static int basefw_mem_state_info(uint32_t *data_offset, char *data)
 	index = 0;
 	tuple_data[index++] = info.free_phys_mem_pages;
 	tuple_data[index++] = info.ebb_state_dword_count;
-	for (i = 0; i < info.ebb_state_dword_count; i++)
+	for (i = 0; i < info.ebb_state_dword_count; i++) {
+#ifdef INTEL_ADSP
 		tuple_data[index + i] = io_reg_read(SHIM_HSPGCTL(i));
```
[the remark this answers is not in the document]
andyross: I filed the bug in #7549, does that not capture the concern?  Or you're saying you want the note in the source code?  That's fair, I'll admit ignorance of SOF policy on that sort of thing, but I do see quite a few sitting around in the tree so it's surely acceptable.

marc-hb:
> I'm too timid to muck with CI files in the same PR, so no change to the workflow yaml here. But I did validate that fuzz.sh works in place if you just pass the kconfig

Thanks for manual testing this, good enough for now.

src/audio/base_fw.c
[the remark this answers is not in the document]
cujomalainey: Yes just presence in the code is what I was looking for, thanks!

cujomalainey (approved):

src/ipc/ipc4/helper.c
```diff
@@ -20,8 +20,14 @@
...
 #include <rtos/wait.h>
-#ifdef __ZEPHYR__
-#include <adsp_memory.h> /* for IMR_BOOT_LDR_MANIFEST_BASE */
+
+/* TODO: Remove platform-specific code, see https://github.com/thesofproject/sof/issues/7549 */
+#if defined(CONFIG_SOC_SERIES_INTEL_ACE) || defined(CONFIG_INTEL_ADSP_CAVS)
+#define RIMAGE_MANIFEST 1
+#endif
+
+#if RIMAGE_MANIFEST
```
lyakh, line 29: `#ifdef`? Also you still need to check for `__ZEPHYR__` I think - as of today IPC4 can still be done with XTOS and we aren't removing that yet

src/ipc/ipc4/helper.c
```diff
@@ -685,11 +691,20 @@ const struct comp_driver *ipc4_get_drv(uint8_t *uuid)
 
 const struct comp_driver *ipc4_get_comp_drv(int module_id)
 {
-	struct sof_man_fw_desc *desc = (struct sof_man_fw_desc *)IMR_BOOT_LDR_MANIFEST_BASE;
+	struct sof_man_fw_desc *desc = NULL;
```
lyakh, line 694: superfluous initialisation

src/platform/posix/ipc.c
```diff
@@ -69,6 +69,7 @@ static void fuzz_isr(const void *arg)
 	memmove(&fuzz_in[0], &fuzz_in[n + 1], rem);
 	fuzz_in_sz = rem;
 
+#ifndef CONFIG_IPC_MAJOR_4
```
lyakh, line 72: shouldn't this rather check for `CONFIG_IPC_MAJOR_3`?

src/ipc/ipc4/helper.c
andyross: No, it's to handle the case below where it may or may not be initialized to IMR depending on RIMAGE_MANIFEST.

src/ipc/ipc4/helper.c
andyross: Good catch, though in this context #if works the same (the original scheme expanded it to some defined() expressions, which I then had to revert because it turns out that's a compiler warning on some toolchain or another).

But __ZEPHYR__ is implicit on the test: RIMAGE_MANIFEST is set based on detection of Zephyr SOC kconfigs.

src/ipc/ipc4/helper.c
lyakh: Yes, if `RIMAGE_MANIFEST` is defined, then `desc` is assigned immediately and there's no need for initialisation. If it isn't defined, then the function returns immediately and `desc` is never used, so there's again no need to initialise. BTW, if it isn't defined and there's a `return` statement right at the beginning, then compilers / analysers might complain about a whole bunch of unused variables and unreachable code.

src/ipc/ipc4/helper.c
lyakh: At least in some cases trying to do an `#if` on a symbol, that isn't defined at all, generated an error (I think it was a hard error, not a warning) - cannot check value of something, that isn't defined.
`__ZEPHYR__` - implicitly, yes... I guess it's good enough for now.

RanderWang (approved):

kv2019i:
One known fail in https://sof-ci.01.org/sofpr/PR7531/build6911/devicetest/index.html , otherwise looks good.

In review comments, @andyross @lyakh there's the one superfluous variable init that is not addressed. I do think this is valid, but I'm inclined to proceed with merge as the rest of the PR is good and we have tests passing.

kv2019i (approved):

marc-hb:
Small #7579 adds IPC4 fuzzing to Github Actions, please review.

PR #7513, opened 2023-04-25
Revert "scripts/docker-run.sh: run with sudo-cwd.sh"

marc-hb:
This reverts commit 80e9c3454a0a048704f863a52443b536a27a2410.

This was merged too early, there are still build issues and failures to solve first, notably issues caused by the now missing xtensa-cnl-elf-gcc (for TGL)

https://github.com/thesofproject/sof/actions/runs/4789961860/jobs/8518459642 https://github.com/thesofproject/sof/actions/runs/4789961859/jobs/8518459155 etc.

Issue #7066, opened 2023-02-09
SOF 2.5 Release Checklist


kv2019i:
Estimated Timeline (in ISO 8601 weeks)
- stable-v2.5 branch 2023-W10
- RC1 2023-W11
- final tag 2023-W12 
- release with product binaries 2023-W13-

Pre RC1 Steps:
- [x] tag any of the issues and feature lists with milestone v2.X  https://github.com/thesofproject/sof/issues?q=is%3Aopen+is%3Aissue+milestone%3Av2.5
- [x] Choose a SOF commit on main branch as candidate for branch out stable-v2.X (done, 528c7ebce0c1bcb50bcffeb30fc3be8cf549a98d )
- [x] Testing & sign off this commit
- [x] branch out stable-v2.5 , target WW10 

RC1 Steps: 
- [x] Cherry pick critical bug fixings and features to stable-v2.X for RC1
- [x] Perform FW security scan and fix opens -> done, fixed integrated in https://github.com/thesofproject/sof/pull/7283
- [x] Testing & sign off
- [x] Tag v2.5-rc1  -> https://github.com/thesofproject/sof/releases/tag/v2.5-rc1
- [x] Make rc1 release on https://github.com/thesofproject/sof/releases

RC1 Vendor specific Steps / Intel:
- [ ] signing of the FW, upload FW and topology binaries to internal release folder -> skipped for v2.5
- [ ] Update binary release script -> skipped for v2.5
- [ ] Make rc1 release on https://github.com/thesofproject/sof-bin -> skipped for v2.5

Final release steps:
- [x] Cherry pick critical bug fixings to stable-v2.X
- [x] Testing & sign off
- [x] Tag v2.X -> https://github.com/thesofproject/sof/releases/tag/v2.5
- [x] Make release on https://github.com/thesofproject/sof/releases

Final release, vendor specific steps / Intel:
- [x] signing of the FW, upload FW and topology binaries to internal release folder
- [x] Make release on https://github.com/thesofproject/sof-bin

Final steps:
- [x] Declare the release on https://www.sofproject.org/blog/

Cleanup:
- [x] Move any unfinished issues tagged to next milestone (the vX.Y milestone tags)

kv2019i:
SOF2.5 stable branch created https://github.com/thesofproject/sof/tree/stable-v2.5 
Target to tag rc1 next week.

marc-hb:
> stable-v2.5 branch WW10

[...]

PR #7532, opened 2023-04-28
version.cmake: switch SOF_MAJOR etc. to a new, static `versions.json`

marc-hb:
As discussed in #6952, relying on `git describe` for defining SOF_MAJOR and friends is very brittle and unreliable. Switch to a static `versions.json` file instead.

Note the full `git describe --tags` output is _still_ present in the binary, this is useful and left unchanged. It's just not used any more to guess SOF_MAJOR, SOF_MINOR and SOF_MICRO.

Use JSON because CMake and pretty much every other piece of software has a JSON parser out of the box.

This aligns with linux/Makefile, Zephyr/VERSIONS and probably many others. - except we use JSON because we're in 2023 so we don't spend time having fun re-implementing parsers any more.


marc-hb:
This is ready for review but merge is blocked on the upgrade of the Docker image (#7513)

dbaluta (approved):
I think this is a good start from relying on git in order to get the version.

paulstelian97 (approved):
This is a great change. While it does create issues with redundant information between tags (which won't disappear) and the actual version, I believe we can take enough care to keep it in sync. So I'm more on the "yes this is good" side of the fence, approving.

kv2019i (approved):
Thanks @marc-hb ! We need to add this to the release checklist for 2.6 once that is created. FYI @mengdonglin @lgirdwood

kv2019i:
One fail in https://sof-ci.01.org/sofpr/PR7532/build6883/devicetest/index.html , not related, proceeding with merge.

marc-hb:
> Thanks @marc-hb ! We need to add this to the release checklist for 2.6 once that is created. FYI @mengdonglin @lgirdwood

Submitted:
- https://github.com/thesofproject/sof-bin/pull/123

Edited:
- https://github.com/thesofproject/sof/issues/7066

PR #7538, opened 2023-04-28
Fixes for i.MX xt-clang build

paulstelian97:
These two commits, together with upstream patch https://github.com/zephyrproject-rtos/zephyr/pull/57374 fix the build of SOF for i.MX platforms with the new xt-clang Cadence 2023 toolchain.

Keeping as draft until the Zephyr patch is merged. (edit: Zephyr patch merged, west manifest updated)

### DIFF
--- a/src/arch/xtensa/include/xtensa/config/core.h
+++ b/src/arch/xtensa/include/xtensa/config/core.h
@@ -36,6 +36,24 @@
 #ifndef XTENSA_CONFIG_CORE_H
 #define XTENSA_CONFIG_CORE_H
 
+/*
+ * This define is used by the new 2023 xt-clang toolchain and, while there are a few definitions
+ * (identical to this one) in various implementations such as newlib, none of them is in use when
+ * building SOF with Zephyr and XtensaTools.
+ */
+
+#if defined(__ZEPHYR__)
+
+#ifndef __UINT32_C
+#define __UINT32_C(x) x ## U
+#endif
+
+#ifndef UINT32_C
+#define UINT32_C(x) __UINT32_C(x)
+#endif
+
+#endif
+
 /*  CONFIGURATION INDEPENDENT DEFINITIONS:  */
 #ifdef __XTENSA__
 #include <xtensa/hal.h>
--- a/src/drivers/imx/interrupt-generic.c
+++ b/src/drivers/imx/interrupt-generic.c
@@ -36,6 +36,7 @@ int irqstr_get_sof_int(int irqstr_int)
 
 void platform_interrupt_init(void) {}
 
+#ifndef __ZEPHYR__
 void platform_interrupt_set(uint32_t irq)
 {
 	arch_interrupt_set(irq);
@@ -45,6 +46,7 @@ void platform_interrupt_clear(uint32_t irq, uint32_t mask)
 {
 	arch_interrupt_clear(irq);
 }
+#endif /* __ZEPHYR__ */
 
 uint32_t platform_interrupt_get_enabled(void)
 {
--- a/src/drivers/imx/interrupt-irqsteer.c
+++ b/src/drivers/imx/interrupt-irqsteer.c
@@ -451,6 +451,7 @@ void platform_interrupt_init(void)
 		interrupt_cascade_register(dsp_irq + i);
 }
 
+#ifndef __ZEPHYR__
 void platform_interrupt_set(uint32_t irq)
 {
 	if (interrupt_is_dsp_direct(irq))
@@ -462,6 +463,7 @@ void platform_interrupt_clear(uint32_t irq, uint32_t mask)
 	if (interrupt_is_dsp_direct(irq))
 		arch_interrupt_clear(irq);
 }
+#endif /* __ZEPHYR */
 
 uint32_t platform_interrupt_get_enabled(void)
 {
--- a/west.yml
+++ b/west.yml
@@ -43,7 +43,7 @@ manifest:
 
     - name: zephyr
       repo-path: zephyr
-      revision: aba3b12e311b4779338406fe3e7c4c54f655d0cd
+      revision: fa5117225af9c2167449d79cccd5f0e81bb10696
       remote: zephyrproject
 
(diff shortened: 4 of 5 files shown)


src/arch/xtensa/include/xtensa/config/core.h
```diff
@@ -36,6 +36,19 @@
...
+/* Zephyr/XCC specific definitions */
+#if defined(__ZEPHYR__)
+
+#ifndef __UINT32_C
+#define __UINT32_C(x) x ## U
+#endif
+
+#ifndef UINT32_C
+#define UINT32_C(x) __UINT32_C(x)
+#endif
```
marc-hb, line 48: This `#ifndef #define` pattern is at best a necessary evil because the definition can change depending the `#include` order (recent example: 2512698098ed) and the direction the wind blows.

So can you transfer some of the information and justification from the commit message to the comment here?

paulstelian97: The reason I had to put it in here in the first place, especially with the __ZEPHYR__ guard, is because of the include order. On XTOS builds (including i.MX8) without that guard this is included before newlib and newlib does _not_ do this, which leads to conflicts.

I will copy the comment and resubmit, maybe today but if not on Tuesday.

dbaluta:
Now because Zephyr PR was merged I think we are ready to get this out of draft state. 

@iuliana-prodan should we update the west yaml to pull in the new zephyr code?

kv2019i:
@dbaluta wrote:
>     * are we using https://github.com/zephyrproject-rtos/zephyr/ with sof? or https://github.com/thesofproject/zephyr
>     * should we create a separate PR to upgrade the zephyr hash?

We are using https://github.com/zephyrproject-rtos/zephyr/ and the version defined in sof/west.yml.

So you can add a commit to this PR to update the Zephyr version (see git log of sof/west.yml for examples). If we have a breaking interface change (like the SYS_INIT last week), then the SOF change and west.yml update need to be squashed into a single git commit, in order not to break git bisect.

dbaluta:
@lyakh  great! then we are fine. A commit to update zephyr has was already added to this PR.

kv2019i:
@dbaluta wrote:
> @lyakh great! then we are fine. A commit to update zephyr has was already added to this PR. Also, how are you doing for stable-v2.x version?

stable-v2.x branches have their own west.yml and there we've either used upstream Zephyr repo (if no backports are needed to Zephyr), or SOF zephyr clone (if Zephyr backports are needed for the stable series).

The west.yml specifies both the repo and commit, so we can easily switch. Any any SOF branch/commit will always have a clear mapping to a Zephyr version, so there's no ambiguity on the version used.

PS west.yml can also point to an unmerged pull request. This is a nice tool to test a Zephyr change with SOF CI (before it gets merged).

marc-hb:
> are we using https://github.com/zephyrproject-rtos/zephyr/ [upstream] with sof? or https://github.com/thesofproject/zephyr [fork]

Short answer: upstream every time it's possible, fork when it's not.

Long answer: see @kv2019i's above :-)

kv2019i (approved):
Looks good, let's see the CI for the Zephyr update (should be good).

kv2019i:
@tmleman Do we need https://github.com/thesofproject/sof/pull/7174 to make the INtel MTL test pass, or is this something else (this PR moves to a newer Zephyr).

tmleman:
> Do we need #7174 to make the INtel MTL test pass, or is this something else (this PR moves to a newer Zephyr).

@kv2019i If we do a zephyr update without my patch, the secondary cores will stop working on MTL. Currently, there are probably no tests for this in CI, so the fail's are rather unrelated.

paulstelian97:
I don't have any qualms with which specific version of Zephyr to update to, my pull request just holds the earliest commit that solves the problem, but I wouldn't mind anything newer (or an update on the main branch that I can rebase on top of, which is even better IMO)

dbaluta:
@tmleman so should we merge https://github.com/thesofproject/sof/pull/7174 first? Or should we merge current PR now? 

Is #7174 ready? We would like to have this patch merged asap because we have an internal release coming.

tmleman:
> @tmleman so should we merge #7174 first? Or should we merge current PR now?
> 
> Is #7174 ready? We would like to have this patch merged asap because we have an internal release coming.

@dbaluta https://github.com/thesofproject/sof/pull/7174 is ready to be merged and I think it would be better if it gets merged first. It would make bisecting easier in case of any problems.

kv2019i:
@wszypelt Ok to proceed, this is not touching Intel platforms.?

kv2019i:
@paulstelian97 #7174 merged, can you rebase and let's see if the CI passes now.

paulstelian97:
Main branch still behind the commit I need, do I keep the update?

dbaluta:
@paulstelian97 we still need to zephyr update, right? otherwise, the build will fail for SOF imx8 with xt-clang.

src/arch/xtensa/include/xtensa/config/core.h
```diff
@@ -36,6 +36,24 @@
...
+
+#endif
```
lyakh, line 55: is this really a suitable header for this? This seems some compiler compatibility support, how about xtos/include/sof/compiler_attributes.h?

paulstelian97: I'm pretty sure the xtos/ subfolder isn't used in Zephyr builds; I want the opposite (something that only applies to Zephyr)

src/arch/xtensa/include/xtensa/config/core.h
marc-hb: > On XTOS builds (including i.MX8) without that guard this is included before newlib and newlib does not do this, which leads to conflicts.

What sort of conflicts? This is coming up again in non-Zephyr build https://github.com/thesofproject/sof/pull/9351#discussion_r1707489595 ...

Maybe @keith-packard or @stephanosio could help us get out of this libc / toolchain nightmare?

PR #7544, opened 2023-05-01
app: Switch main return type to 'int'

keith-packard:
With Zephyr adopting a different type for 'main', adapt this application to match.

This is required when building without the `-ffreestanding` compiler flag as the C standard for "hosted" applications requires one of a small set of specific signatures for the `main` function.

### DIFF
--- a/app/src/main.c
+++ b/app/src/main.c
@@ -20,7 +20,7 @@ int sof_main(int argc, char *argv[]);
  * TODO: Here comes SOF initialization
  */
 
-void main(void)
+int main(void)
 {
 	int ret;
 
@@ -47,4 +47,5 @@ void main(void)
 	 */
 	k_thread_suspend(k_current_get());
 #endif
+	return 0;
 }


sofci:

gkbldcig:

marc-hb:
test this please

app/src/main.c
```diff
@@ -47,4 +47,5 @@ void main(void)
 	 */
 	k_thread_suspend(k_current_get());
 #endif
+	return 0;
```
ranj063, line 50: dont we need to return ret after line 32 as well?

keith-packard: There wasn't one there before, and I didn't think this patch should change the control flow. If you think it should do something different here, that seems like a separate issue that should be applied as a separate PR.

ranj063: @keith-packard the function was returning a void before so of course it wasnt there before but now that it return an int, i think it should fix this part. I do not think it is a separate issue

keith-packard: it wasn't returning from there though, there were no return statements in the function at all.

ranj063: OK so can we at least return ret at the end then?

keith-packard: Yeah, you'd think that would be the right thing to do so that the error would get propagated back into the system. However, Zephyr doesn't currently do anything with the return value, nor is there anything obvious that could be done with the current design. As a result of that, the Zephyr TSC decided to simply require that `main` return 0 and make all other values reserved until someone figures out a better plan.

You can check out the Zephyr spec for main here:

https://github.com/zephyrproject-rtos/zephyr/blob/main/doc/develop/languages/c/index.rst

ranj063: Thanks for the explanation @keith-packard

ranj063 (approved):

app/src/main.c
keith-packard: no worries; it's a bit unexpected give experience with other operating systems, so understanding how Zephyr got to the current spec is helpful in reviewing patches like this. Thanks much for your review!

marc-hb: No strong opinion but I kinda like the unusual `return 0;`: it draws attention to the fact that the value is **ignored**.

`42` would of course have been better but `0` is OK.

> the Zephyr TSC decided to simply require that main return 0

The corresponding documentation PR  has 133 comments so I will personally trust that decision blindly :-)
https://github.com/zephyrproject-rtos/zephyr/pull/54628



Fun fact: rename `main()` to `main_whatever()` and the SOF firmware still compiles and links.

paulstelian97 (approved):

stephanosio (approved):

dbaluta (approved):

kv2019i:
Ready to go, pending CI results for Intel quickbuild -> @wszypelt

softwarecki (approved):

PR #7545, opened 2023-05-02
topology2: cavs-sdw: make jack optional

bardliao:
We assume sdw jack is always there in a sdw topology. 
This commit make it be optional to deal with the sdw amp only case.

### DIFF
--- a/tools/topology/topology2/cavs-sdw.conf
+++ b/tools/topology/topology2/cavs-sdw.conf
@@ -69,6 +69,7 @@ Define {
 	SDW_JACK_IN_BE_ID	1
 	NUM_SDW_AMP_LINKS 0
 	SDW_DMIC 0
+	SDW_JACK true
 }
 
 # override defaults with platform-specific config
@@ -90,11 +91,6 @@ IncludeByKey.NUM_HDMIS {
 "[3-4]" "platform/intel/hdmi-generic.conf"
 }
 
-# include deep buffer config if buffer size is in 1 - 1000 ms.
-IncludeByKey.DEEPBUFFER_FW_DMA_MS{
-        "[1-1000]" "platform/intel/deep-buffer.conf"
-}
-
 IncludeByKey.NUM_SDW_AMP_LINKS {
 "[1-2]" "platform/intel/sdw-amp-generic.conf"
 }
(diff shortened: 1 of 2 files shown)


tools/topology/topology2/platform/intel/sdw-jack-generic.conf
```diff
@@ -0,0 +1,156 @@
+#
+# List of all DAIs
+#
+#ALH Index: 0, Direction: duplex
```
ranj063, line 4: that comment about duplex is wrong

tools/topology/topology2/platform/intel/sdw-jack-generic.conf
```diff
@@ -0,0 +1,156 @@
...
+			name	"ALH3"
+		}
+	}
+]
+
+#
+# Pipeline definitions
+#
+
+# Pipeline ID:1 PCM ID: 0
```
ranj063, line 40: this comment makes no sense. There are 3 pipelines below and no PCMs unlit later

tools/topology/topology2/platform/intel/sdw-jack-generic.conf
```diff
@@ -0,0 +1,156 @@
...
+			id	0
+			name	"ALH2"
+		}
+	}
+	{
+		dai_index 10
+		id 		$SDW_JACK_IN_BE_ID
+		direction	"capture"
+		name		$SDW_JACK_IN_STREAM
+		default_hw_conf_id	1
```
ranj063, line 25: there's only 1 hw_config. Why should the default_hw_config_id be 1 here?

tools/topology/topology2/cavs-sdw.conf
```diff
@@ -103,159 +104,7 @@ IncludeByKey.SDW_DMIC {
...
-				node_type $ALH_LINK_INPUT_CLASS
-			}
-			Object.Widget.eqiir.1 {
-				Object.Control.bytes."1" {
-					name '4 Main capture Iir Eq'
-				}
-			}
-		}
-	]
+IncludeByKey.SDW_JACK {
```
ranj063, line 107: assuming there's no Jack, do we need the deep buffer? or does it need to move into sdw-jack.conf as well?

bardliao: > assuming there's no Jack, do we need the deep buffer? or does it need to move into sdw-jack.conf as well?

I thought about this. But consider that we may add deep buffer support for amp someday, and it won't compile if we include deep-buffer.conf in both sdw-jack-generic.conf and sdw-amp-generic.conf. So, I decide to include deep-buffer.conf in cavs-sdw.conf and set DEEPBUFFER_FW_DMA_MS in the cmake file.
BTW, setting DEEPBUFFER_FW_DMA_MS=0 will still include deep-buffer.conf. We need to set DEEPBUFFER_FW_DMA_MS to an alphabet, like "N" to remove deep buffer support.

tools/topology/topology2/platform/intel/sdw-jack-generic.conf
bardliao: > that comment about duplex is wrong

Removed the comment

tools/topology/topology2/platform/intel/sdw-jack-generic.conf
bardliao: > this comment makes no sense. There are 3 pipelines below and no PCMs unlit later

Removed it, too. I have no idea why we need those comments. It was copied from the original csvs-sdw.conf.

tools/topology/topology2/platform/intel/sdw-jack-generic.conf
bardliao: > there's only 1 hw_config. Why should the default_hw_config_id be 1 here?

Changed it to 0 now.

bardliao:
changes:  [v2.txt](https://github.com/thesofproject/sof/files/11371634/v2.txt)

tools/topology/topology2/cavs-sdw.conf
RanderWang: Discussed by Pierre that it makes no sense for amp to use deep buffer since the power saved by deep buffer is insignificant compare to power consumption by amp.

bardliao: Thanks @RanderWang @ranj063, moved deep-buffer.conf to sdw-jack-generic.conf now

keqiaozhang:
SOFCI TEST

RanderWang (approved):
LGTM

plbossart (approved):

ranj063 (approved):

PR #6276, opened 2022-09-08
xtensa-build-zephyr: pass sof build version to rimage

dnikodem:
Rimage expects to receive fw_ver_build to prepare the manifest.

Signed-off-by: Damian Nikodem <damian.nikodem@intel.com>

PR #7546, opened 2023-05-02
xtensa-build-zephyr.py: don't extract SOF_BUILD from generated .h file

marc-hb:
This replaces and simplifies commit a823958a8d2d3 ("xtensa-build-zephyr: pass sof build version to rimage") with a constant.

EDIT: That commit was submitted to avoid a test failure in a broken, internal
test looking for an SOF_BUILD value... hardcoded to 1! (internal FW issue
257, test TestLoadFwExtended::()::test_00_01_load_fw_and_check_version")

The final goal is for this script to stop extracting anything from `generated/sof_versions.h` so rimage can be configured _before_ the build has started. Other commits will deal with other parts of sof_versions.h

SOF_BUILD can be replaced by a constant because it has always been one when building with Zephyr. The optional BLD_COUNTERS feature - disabled by default in XTOS since commit 9f8cce152240 ("Disable __TIME__ and the non-reproducible build counter by default") for build reproducibility reasons - has never worked with Zephyr for the following reasons:

- The sof_add_build_counter_rule() invocation is located in the top-level sof/CMakeLists.txt file which has never been used by the Zephyr build.

- sof_add_build_counter_rule() registers a POST_BUILD command for the top-level "sof" CMake target but there is no "sof" build target in the Zephyr build.

- Variables like VERSION_BUILD_COUNTER_CMAKE_PATH are not defined in the Zephyr build for some unknown reason.

- Probably others.

### DIFF
--- a/scripts/cmake/version-build-counter.cmake
+++ b/scripts/cmake/version-build-counter.cmake
@@ -8,6 +8,10 @@ set(VERSION_BUILD_COUNTER_CMAKE_PATH ${CMAKE_CURRENT_LIST_DIR}/version-build-cou
 
 set(BUILD_COUNTER_PATH "${SOF_ROOT_BINARY_DIRECTORY}/.build")
 
+# Among many other differences, the Zephyr build does not invoke
+# sof_add_build_counter_rule() at build time. In other words, Zephyr has
+# never supported incrementing the BLD_COUNTERS: SOF_BUILD is always
+# equal to the starting value below when building with Zephyr.
 if(NOT EXISTS "${BUILD_COUNTER_PATH}")
 	file(WRITE "${BUILD_COUNTER_PATH}" "1")
 endif()
--- a/scripts/xtensa-build-zephyr.py
+++ b/scripts/xtensa-build-zephyr.py
@@ -60,7 +60,6 @@ west_top = pathlib.Path(SOF_TOP, "..").resolve()
 default_rimage_key = pathlib.Path(SOF_TOP, "keys", "otc_private_key.pem")
 
 sof_fw_version = None
-sof_build_version = None
 
 if py_platform.system() == "Windows":
 	xtensa_tools_version_postfix = "-win32"
@@ -444,17 +443,16 @@ def west_update():
 	execute_command(["west", "update"], check=True, timeout=3000, cwd=west_top)
 
 
-def get_build_and_sof_version(abs_build_dir):
-	"""[summary] Get version string major.minor.micro and build of SOF
+def get_sof_version(abs_build_dir):
+	"""[summary] Get version string major.minor.micro of SOF
 	firmware file. When building multiple platforms from the same SOF
 	commit, all platforms share the same version. So for the 1st platform,
 	generate the version string from sof_version.h and later platforms will
 	reuse it.
 	"""
 	global sof_fw_version
-	global sof_build_version
-	if sof_fw_version and sof_build_version:
-		return sof_fw_version, sof_build_version
+	if sof_fw_version:
+		return sof_fw_version
 
 	versions = {}
 	with open(pathlib.Path(abs_build_dir,
@@ -465,9 +463,8 @@ def get_build_and_sof_version(abs_build_dir):
 				versions[words[1]] = words[2]
 	sof_fw_version = versions['SOF_MAJOR'] + '.' + versions['SOF_MINOR'] + '.' + \
 		      versions['SOF_MICRO']
-	sof_build_version = versions['SOF_BUILD']
 
-	return sof_fw_version, sof_build_version
+	return sof_fw_version
 
 def rmtree_if_exists(directory):
 	"This is different from ignore_errors=False because it deletes everything or nothing"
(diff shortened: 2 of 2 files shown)


marc-hb:
https://sof-ci.01.org/sofpr/PR7546/build6851/build/mtl_zph_ipc4.txt is still invoking `rimage` with `-b 1` as it always has.

marc-hb:
https://sof-ci.01.org/sof-pr-viewer/#/build/PR7546/build11924100 has the usual error below. 

Everything else is green including https://sof-ci.01.org/sofpr/PR7546/build6853/devicetest and https://sof-ci.01.org/sofpr/PR7546/build6852/devicetest

@wszypelt  a green report would be so unusual that it would be almost worrying!

cc: 
```
22:33:53,387 INFO  - Executing pre-execute action...
22:33:53,387 INFO  - Running step...
22:42:34,319 INFO  - Executing post-execute action...
22:42:34,319 ERROR - Step 'master>SCANS>TESTS>TRIGGER PYTHON SMOKE_IPC4 TESTS' is failed: Step is failed since the triggered build is failed, cancelled, or timed out.
```

marc-hb:
Someone re-run https://sof-ci.01.org/sof-pr-viewer/#/build/PR7546/build11933012 and even that is green now. Everything is.

kv2019i (approved):

aborisovich (changes requested):
Well, the PR looks OK but I am not fan of the solution you took for
`sign_cmd += ["-b", "1" ]` . It makes no sense in my opinion because it is just hardcoded value in the script without any specific reason. Rest of the PR is ok but I would like to push you for a bit more, because you wrote that

> The final goal is for this script to stop extracting anything from generated/sof_versions.h so rimage can be configured before the build has started. Other commits will deal with other parts of sof_versions.h

So, if this statement is a motivation for this PR, I think you can go further and refactor a bit how the version is generated. What I mean is:

1. From SOF configure rimage using [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) and pass SOF FW version to rimage CMake at configure time.
2. Change rimage - if no command line arguments provided with version number, use the one from CMake cache.
3. Remove completely `get_sof_version(abs_build_dir)` function from `xtensa-build-zephyr.py`.
4. Generate sof_versions.h file using [configure_file](https://cmake.org/cmake/help/latest/command/configure_file.html) function.

This approach would finally simplify version handling and make it more robust.
We would pass if from CMake down, not from Cmake to python to west to rimage command args.
SOF versioning is not bount in any way to platform we build for, so there should be no problem with this approach. I think it matches also the idea of having rimage producing .ri images as part of "build" step not "sign" step.

marc-hb:
Thanks for the  review! Very hard to find reviewers in this area.


> but I am not fan of the solution you took for `sign_cmd += ["-b", "1" ]` . It makes no sense in my opinion because it is just hardcoded value in the script without any specific reason

So the rationale for hardcoding `sign_cmd += ["-b", "1" ]` was to keep the binary 100% unchanged. Thanks for pointing this out; it's not a very good reason. In fact I'm now leaning toward dropping this line. Then rimage will default to `-b 0` when there is no `-b`. Also, if I drop this line then users can more easily override the value using `west config` or whatever other way CMake way they find without wondering what happens when there are **two** `-b` options on the same rimage command line; always confusing.

What dropping `sign_cmd += ["-b", "1" ]` work for you?

> (and eventually consider removing this build counter from rimage if it has no use cases).

Maybe. SOF_BUILD is disabled by default but it still works in XTOS builds though. Maybe some SOF_BUILD fan could also find a way to finally make it work when building Zephyr? Dunno.


> pass SOF FW version to rimage CMake at configure time.

There seems to be a misunderstanding here? -b is an rimage parameter passed when _running_ rimage.

> Generate sof_versions.h file using configure_file function.

But this would make sof_versions.h  more static that it is now? See commit 23d2a017773c. Either way it looks  off-topic to me in this PR but maybe I'm missing something.



 ---------

FetchContent() suggestion.

Adopting FetchContent() would be a much bigger change than this small, "no-op" PR! I mean SOF_BUILD never worked before this small PR and it still does not work after this PR and that's pretty much it...

I see where you're coming from because FetchContent() clearly looks like a tool designed to solve dependencies like SOF=>rimage . However I see a number of reasons why FetchContent would not fit here. Main reason: I think it would go against the spirit and the letter of the Zephyr build.


- Some Zephyr tests have a DIRECT dependency rimage to run on hardware WITHOUT depending on any SOF code! So FetchContent() in SOF would not help them, they would have to FetchContent in Zephyr too...
- ... but Zephyr requires all dependencies to be listed in a centralized place: either in a west manifest: https://docs.zephyrproject.org/latest/contribute/external.html#contributing-external-tooling (or in module.yml for binaries https://docs.zephyrproject.org/latest/contribute/bin_blobs.html#bin-blobs)
- Tracking dependencies in a single place makes https://www.cisa.gov/sbom easier. I think SBOM is one of the reasons Zephyr wants it done like this.
- Downloading at configuration/build time goes against "Hermetic" builds https://github.com/tektoncd/pipeline/blob/main/docs/hermetic.md ("This means they do not have network access,...)

Please start a new issue if you want to keep recommending FetchContent()?

aborisovich:
> Would dropping sign_cmd += ["-b", "1" ] work for you?

Sure, it's my only concern.
Maybe I'll submit a feature request about my idea. Because you see, FetchContent may but does not have to clone. We may still stick to west in terms of cloning, just use functions from fetch content to populate the CMake cache of rimage and run its configure step as part of SOF configure step. Then, built rimage executable would have 'hardcoded version' of SOF inside of it. And we could use that version when no rimage input arguments related to version are provided (aka 'default'). So rimage would use not only information provided in the imput args but also information given from the build. Well, anyway that was just an idea.

marc-hb:
> Would dropping sign_cmd += ["-b", "1" ] work for you?

I was dropping this line and then I had second thoughts and I took a second look at #6276. I found internal issue 257. So the whole reason SOF_BUILD was added to xtensa-build-zephyr.py in the first place is apparently because some broken test is checking for SOF_BUILD to be hardcoded to 1!! So I kept the hardcoding for now and added a FIXME :-(

aborisovich (approved):
> So the whole reason SOF_BUILD was added to xtensa-build-zephyr.py in the first place is apparently because some broken test is checking for SOF_BUILD to be hardcoded to 1!!

This is just sad...

kv2019i:
Seems ready to go, but need a clean CI pass for required checks.

marc-hb:
> Seems ready to go, but need a clean CI pass for required checks.

The Ubuntu mirrors were very unreliable yesterday. Today they worked!

https://sof-ci.01.org/sofpr/PR7546/build7928/devicetest has a couple failures as usual, nothing that could be caused by this.

The zmain branch needs the Zephyr SDK 0.16.1, known issue.

Everything else is green now.

PR #7552, opened 2023-05-03
[SKIP CI] sof_builder/Dockerfile: restore deleted toolchains

marc-hb:
Building a Docker image takes several hours. To reduce costs and save time we don't want to maintain a different Docker image for the stable-v2.2 branch. Restore toolchains that were accidentally removed from the Dockerfile because this will unnecessarily paint red every PR submitted to the stable-v2.2 branch, example:

https://github.com/thesofproject/sof/actions/runs/4867162042/jobs/8679468096

If we want to stop supporting something then let's keep things simple and remove it only where it actually gets in the way and where it actually becomes a problem. No need to search and scrub every reference to it from the face of the planet and accidentally make our life harder on release branches, create pointless git conflicts and _increase_ maintenance costs when trying to reduce them!

### DIFF
--- a/scripts/docker_build/sof_builder/Dockerfile
+++ b/scripts/docker_build/sof_builder/Dockerfile
@@ -93,7 +93,7 @@ RUN cd "$HOME" && \
 	git clone $CLONE_DEFAULTS --branch sof-gcc10x $CT_NG_REPO && \
 	cd crosstool-ng && \
 	./bootstrap && ./configure --prefix=`pwd` && make && make install && \
-	for arch in cnl imx imx8m imx8ulp rn rmb mt8186 mt8195; do \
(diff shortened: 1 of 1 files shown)


paulstelian97:
Deleting them is still needed at some point, but you make a point on how that deletion should probably be batched and not happen all the time.

Is stable-v2.2 still an actively-supported branch? If so then yeah you probably should keep the toolchains all the way until those platforms go out of support even in stable branches.

marc-hb:
> Is stable-v2.2 still an actively-supported branch?

It has been mainly for topologies so far if you check the git log. For the rest not much has changed.

The main idea here is not to branch the docker image unless/until necessary. Imagine having to compile and install some very old ALSA version to make a minor topology fix or addition.

paulstelian97 (approved):
I suspect a removal will still be useful in the future, but not now.

PR #7554, opened 2023-05-03
Add multi-sink support for the optimized DAI copiers

ranj063:


src/audio/dai-zephyr.c
```diff
@@ -250,13 +252,54 @@ static enum dma_cb_status dai_dma_cb(struct dai_data *dd, struct comp_dev *dev,
...
+			sink = container_of(sink_list, struct comp_buffer, source_list);
+
+			/* this has been handled above already */
+			if (sink == dd->local_buffer)
+				continue;
+
+			sink_c = buffer_acquire(sink);
+			sink_dev = sink_c->sink;
+
+			j = IPC4_SINK_QUEUE_ID(sink_c->id);
```
btian1, line 281: this j may have bugs, I remember I observe 65537 value.

src/audio/dai-zephyr.c
```diff
@@ -250,13 +252,54 @@ static enum dma_cb_status dai_dma_cb(struct dai_data *dd, struct comp_dev *dev,
...
+
+			if (!converter[j]) {
+				buffer_release(sink_c);
+				buffer_release(dma_buf);
+				return -EINVAL;
```
btian1, line 286: before return, I want to add err comment here with print j

src/audio/dai-zephyr.c
ranj063: I don't think so @btian1. This logic is already used in the copier. You have to tell me exactly why it would be invalid.

btian1: if (i >= IPC4_COPIER_MODULE_OUTPUT_PINS_COUNT)
		return -EINVAL;

src/audio/dai-zephyr.c
```diff
@@ -1283,17 +1340,42 @@ int dai_zephyr_copy(struct dai_data *dd, struct comp_dev *dev)
...
 
 	/* calculate minimum size to copy */
 	if (dev->direction == SOF_IPC_STREAM_PLAYBACK) {
+		buf_c = buffer_acquire(dd->local_buffer);
 		src_samples = audio_stream_get_avail_samples(&buf_c->stream);
 		sink_samples = free_bytes / sampling;
 		samples = MIN(src_samples, sink_samples);
+		buffer_release(buf_c);
 	} else {
+		struct list_item *sink_list;
```
btian1, line 1352: suggest add comments for the reason of calculate min samples at else beginning

src/audio/dai-zephyr.c
```diff
@@ -1283,17 +1340,42 @@ int dai_zephyr_copy(struct dai_data *dd, struct comp_dev *dev)
...
+			struct comp_dev *sink_dev;
+			struct comp_buffer *sink;
+
+			sink = container_of(sink_list, struct comp_buffer, source_list);
+			buf_c = buffer_acquire(sink);
+			sink_dev = buf_c->sink;
+
+			/* this happens when the DAI device is an endpoint created by the copier */
+			if (!sink_dev) {
+				sink_samples = audio_stream_get_free_samples(&buf_c->stream);
```
btian1, line 1365: can combine 1365-1366 with 1372-1373?

ranj063: I dont think so, there's no easy way to combine them. This is easier to read I think

src/audio/dai-zephyr.c
```diff
@@ -1283,17 +1342,47 @@ int dai_zephyr_copy(struct dai_data *dd, struct comp_dev *dev)
...
+		buf_c = buffer_acquire(dd->local_buffer);
 		src_samples = audio_stream_get_avail_samples(&buf_c->stream);
 		sink_samples = free_bytes / sampling;
 		samples = MIN(src_samples, sink_samples);
+		buffer_release(buf_c);
 	} else {
+		struct list_item *sink_list;
+
+		/*
+		 * In the case of capture DAI's with multiple sinks, compute the minimum number
```
lyakh, line 1357: Took me some time to check what you mean by a "sink" here - looks like an output, right? I think we call opposite things by a "sink pin" and a "sink buffer," when you say "a DAI with multiple sinks" I imaging sink pins, i.e. inputs, which isn't the case here?

src/audio/dai-zephyr.c
```diff
@@ -1283,17 +1342,47 @@ int dai_zephyr_copy(struct dai_data *dd, struct comp_dev *dev)
...
+			sink = container_of(sink_list, struct comp_buffer, source_list);
+			buf_c = buffer_acquire(sink);
+			sink_dev = buf_c->sink;
+
+			/*
+			 * There's no sink device when the DAI device is an endpoint created by the
+			 * copier. So just get the sink_samples and break out of the loop as there
+			 * can only be 1 sink buffer in this case.
+			 */
+			if (!sink_dev || (sink_dev && sink_dev->state == COMP_STATE_ACTIVE)) {
```
lyakh, line 1375: just `if (!sink_dev || sink_dev->state == COMP_STATE_ACTIVE)` is enough

src/audio/dai-legacy.c
```diff
@@ -1047,7 +1048,11 @@ static int dai_copy(struct comp_dev *dev)
 
 	comp_dbg(dev, "dai_copy()");
 
-	return dai_zephyr_copy(dd, dev);
+	/*
+	 * DAI devices will only ever have 1 sink, so no need to pass an array of PCM converter
+	 * functions. The default one to use is set in dd->process.
+	 */
+	return dai_zephyr_copy(dd, dev, NULL);
```
lyakh, line 1055: see - also here - `NULL` isn't an array, it's a pointer

src/audio/dai-zephyr.c
```diff
@@ -250,13 +252,69 @@ static enum dma_cb_status dai_dma_cb(struct dai_data *dd, struct comp_dev *dev,
...
+		/* copy from DMA to all sink buffers using the right PCM converter function */
+		list_for_item(sink_list, &dev->bsink_list) {
+			struct comp_buffer __sparse_cache *sink_c;
+			struct comp_dev *sink_dev;
+			struct comp_buffer *sink;
+			int j;
+
+			sink = container_of(sink_list, struct comp_buffer, source_list);
+
+			/* this has been handled above already */
```
lyakh, line 275: would it be possible to remove that special case and just handle it in the loop like all others?

src/audio/dai-zephyr.c
ranj063: right I mean sink buffers. Will modify the wording to say sink buffers.

src/audio/dai-zephyr.c
ranj063: I dont want to modify the special case because I think there's something off with the converter function. The dd-> process is set during dai_zephyr_params() for the single sink buffer case and I want to keep it that way until I figure out whats wrong with the converter function.

ranj063:
SOFCI TEST

btian1 (approved):

src/audio/dai-zephyr.c
```diff
@@ -250,13 +252,69 @@ static enum dma_cb_status dai_dma_cb(struct dai_data *dd, struct comp_dev *dev,
...
 	local_buf = buffer_acquire(dd->local_buffer);
 
-	if (dev->direction == SOF_IPC_STREAM_PLAYBACK)
+	if (dev->direction == SOF_IPC_STREAM_PLAYBACK) {
 		ret = dma_buffer_copy_to(local_buf, dma_buf,
 					 dd->process, bytes);
-	else
-		ret = dma_buffer_copy_from(dma_buf, local_buf,
-					   dd->process, bytes);
+		buffer_release(local_buf);
```
RanderWang, line 258: playback also supports multi-sink feature, one is sent to dai gateway and another one is sent back to some modules for feedback.

ranj063: @RanderWang we have no tests for this today. So, I'm going to table it for when we need it.

src/audio/copier/copier.c
```diff
@@ -1387,7 +1387,7 @@ static int do_endpoint_copy(struct comp_dev *dev)
 			break;
 		case SOF_COMP_DAI:
 			if (cd->endpoint_num == 1)
```
RanderWang, line 1389: BTW if (cd->endpoint_num == 1) should be remove here since it is not a multiple-endpoint case, or it will confuse reader that what is else ?

RanderWang (approved):
This PR reveals a issue that we miss host copier part for multi-sink case like:
````
host copier----> module X   (playback case)
      |---> moudle Y
````

**Comment by ranj063** [comment]
> This PR reveals a issue that we miss host copier part for multi-sink case like:
> 
> ```
> host copier----> module X   (playback case)
>       |---> moudle Y
> ```

@RanderWang yes we're aware. @btian1 will be working on adding that

**Comment by btian1** [comment]
> that we miss host copier part for multi-sink case lik

if there is no CI test to cover this, then new topology change is expected for new code development.

File: src/lib/dma.c [hunk 11]
```diff
@@ -378,3 +378,28 @@ int dma_buffer_copy_to(struct comp_buffer __sparse_cache *source,
...
+	uint32_t sink_bytes = audio_stream_sample_bytes(&sink->stream) *
+			      samples;
+	int ret;
+
+	/* process data */
+	ret = process(istream, 0, &sink->stream, 0, samples);
+
+	buffer_stream_writeback(sink, sink_bytes);
+
+	comp_update_buffer_produce(sink, sink_bytes);
```
lyakh, line 402: writeback and produce even if `process()` returned an error?

xtos/include/sof/lib/dma.h
```diff
@@ -532,6 +532,15 @@ int dma_buffer_copy_from(struct comp_buffer __sparse_cache *source,
 			 struct comp_buffer __sparse_cache *sink,
 			 dma_process_func process, uint32_t source_bytes);
 
+/*
+ * Used when copying DMA buffer bytes into multiple sink buffers, one at a time using the provided
+ * conversion function. DMA buffer consume should be performed after the data has been copied
+ * to all sinks.
+ */
```
lyakh, line 539: don't think we need this same comment in both the header and the source - when documenting with Doxygen, I think we only comment in the header? But please check, not sure

src/audio/dai-zephyr.c
```diff
@@ -250,13 +252,69 @@ static enum dma_cb_status dai_dma_cb(struct dai_data *dd, struct comp_dev *dev,
...
+	} else {
+		struct list_item *sink_list;
+
+		audio_stream_invalidate(&dma_buf->stream, bytes);
+		ret = dma_buffer_copy_from_no_consume(dma_buf, local_buf, dd->process, bytes);
```
lyakh, line 263: ignoring errors?

ranj063: @lyakh the pcm converter functions can never fail. I've added a comment now that says that. So its OK to ignore and move on.

src/lib/dma.c
ranj063: @lyakh ret can never return an error because the process callbacks are all PCM converter functions. So yes, ignoring is fine.

src/audio/dai-zephyr.c
lyakh: ok, this might really be minor - but I really wouldn't even assign `ret` if I know I'll ignore it. There are many functions returning non-void that get called with no return value assignment. I find that preferable over assigning a value and never using it...

src/audio/dai-zephyr.c
```diff
@@ -250,13 +252,69 @@ static enum dma_cb_status dai_dma_cb(struct dai_data *dd, struct comp_dev *dev,
...
+					goto err;
+				}
+
+				if (!converter[j]) {
+					comp_err(dev, "No PCM converter for sink queue %d\n", j);
+					ret = -EINVAL;
+					goto err;
+				}
+
+				if (sink_dev && sink_dev->state == COMP_STATE_ACTIVE)
```
lyakh, line 306: ehm... `if (ret >= 0 && sink_dev && sink_dev->state == COMP_STATE_ACTIVE)` and remove those `goto`s...

lyakh (approved):
really minor stuff, feel free to ignore

ranj063:
SOFCI TEST

ranj063:
SOFCI TEST

kv2019i:
Known fails in https://sof-ci.01.org/sofpr/PR7554/build7499/devicetest/index.html . The IMX build issue in zmain is knonw as well. So looks good, proceeding.

PR #7566, opened 2023-05-05
.gitmodules: add comments about relative URL. No change.

marc-hb:
Answering a question from @ceolin and others before him.


paulstelian97 (approved):
This legit clarifies a few things about submodules for me as well. They felt like accepted magic ever since they were introduced in this project!

PR #7568, opened 2023-05-07
host-zephyr: No need to call dma_sg_free() twice

ranj063:
It is called during reset already. No need to invoke it again during free.

### DIFF
--- a/src/audio/host-zephyr.c
+++ b/src/audio/host-zephyr.c
@@ -666,7 +666,6 @@ void host_zephyr_free(struct host_data *hd)
 	dma_put(hd->dma);
 
 	ipc_msg_free(hd->msg);
-	dma_sg_free(&hd->config.elem_array);
 }
 
 static void host_free(struct comp_dev *dev)


ranj063:
SOFCI TEST

paulstelian97:
Do we have the guarantee that reset is _always_ called before free?

ranj063:
> Do we have the guarantee that reset is _always_ called before free?

@paulstelian97 reset must always be called before free otherwise the comp states will be messed up

paulstelian97:
Well guess that's one more thing to check whether it's actually happening in all cases. It isn't clear from the code that reset() must be called before free().

ranj063:
> Well guess that's one more thing to check whether it's actually happening in all cases. It isn't clear from the code that reset() must be called before free().

@paulstelian97 with the Linux driver, this will always be the case. A pipeline will be deleted after it is reset during hw_free. And the reason I want to fix this is because of a double free seen during host pipeline deletion. It is a bug that's hard to reproduce. Let me try to find an instance.

btian1:
if reset always ahead of free for calling, it is must to fix this, not sure whether windows part have same calling sequence or not.

paulstelian97:
I would avoid the double free by:
* Either not actually freeing in reset, OR
* Setting the pointer to NULL in reset, so that in free the no-op dma_sg_free(NULL) is called. (TODO: is it a no-op?)

lyakh:
> * Setting the pointer to NULL in reset, so that in free the no-op dma_sg_free(NULL) is called. (TODO: is it a no-op?)

@paulstelian97 This is a good idea but actually this is already done:

https://github.com/thesofproject/sof/blob/aed1788bbea8d8b7d44d983b86ee580a31638718/src/lib/dma.c#L318-L322

and

https://github.com/thesofproject/sof/blob/aed1788bbea8d8b7d44d983b86ee580a31638718/xtos/include/sof/lib/dma.h#L495-L499

so calling `dma_sg_free()` repeatedly doesn't hurt. So, @ranj063 I understand, that on the "normal" path this is called a second time after a `.reset()`, but are we sure it's never called without a `reset()`? E.g. on the error path in `create_host()`? I'd rather not merge this unless we've verified all the paths.

lyakh (changes requested):
marking it -1 out of abundance of caution - might cause leaks

paulstelian97:
So the original code doesn't have a double free condition here anyway. I'd suggest keep the original code then.

ranj063:
> E.g. on the error path in `create_host()`? I'd rather not merge this unless we've verified all the paths.

@lyakh even if create_host failed, what would calling dma_sg_free() be needed for? The memory is actually not allocated until params anyway. But if params failed, reset will be called because we do params during trigger.

lyakh:
> > E.g. on the error path in `create_host()`? I'd rather not merge this unless we've verified all the paths.
> 
> @lyakh even if create_host failed, what would calling dma_sg_free() be needed for? The memory is actually not allocated until params anyway. But if params failed, reset will be called because we do params during trigger.

@ranj063 So, you're claiming, that when we reach this location `dma_sg_free()` has already been called once and therefore `elem_array->elems == NULL`, but since (1) this isn't fixing anything, (2) we don't have memory leak testing in CI, I'd do a check: add a `panic()` or an error print there if the pointer isn't NULL. Let's run CI with that then we'll be more certain, and maybe keep it in the code too.

PR #7574, opened 2023-05-08
cpc: make CPC data "opt-in" with fallback

lgirdwood:
CPC is used to select the correct clock for the desired pipeline workload.

In the case where the CPC data is not available from the manifest and the kernel driver sends a CPC of 0 then SOF should pick a platform safe level to ensure operational glitch free audio at the expense of power consumption. i.e. it's better to prioritize and preserve audio quality over audio power when CPC data is missing.

### DIFF
--- a/src/audio/pipeline/pipeline-stream.c
+++ b/src/audio/pipeline/pipeline-stream.c
@@ -296,13 +296,15 @@ static int add_pipeline_cps_consumption(struct comp_dev *current,
 		cd = &md->cfg.base_cfg;
 	}
 
-	int kcps = cd->cpc * 1000 / ppl_data->p->period;
-	if (kcps == 0) {
-		tr_warn(pipe, "0 KCPS for module: %#x, core: %d", current->ipc_config.id, ppl_data->p->core);
-	} else {
-		core_kcps_adjust(ppl_data->p->core, kcps);
-		tr_info(pipe, "Registering KCPS consumption: %d, core: %d", kcps, ppl_data->p->core);
+	if (cd->cpc == 0) {
+		cd->cpc = PLATFORM_IPC4_MAX_CPC;
+		tr_warn(pipe, "0 CPS requested for module: %#x, core: %d using safe max CPC: %d",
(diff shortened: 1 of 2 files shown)


lgirdwood:
@kv2019i @ujfalusi @plbossart @mwasko @mmaka1 fyi

src/audio/pipeline/pipeline-stream.c
```diff
@@ -296,13 +296,15 @@ static int add_pipeline_cps_consumption(struct comp_dev *current,
...
 	}
 
-	int kcps = cd->cpc * 1000 / ppl_data->p->period;
-	if (kcps == 0) {
-		tr_warn(pipe, "0 KCPS for module: %#x, core: %d", current->ipc_config.id, ppl_data->p->core);
-	} else {
-		core_kcps_adjust(ppl_data->p->core, kcps);
-		tr_info(pipe, "Registering KCPS consumption: %d, core: %d", kcps, ppl_data->p->core);
+	if (cd->cpc == 0) {
+		cd->cpc = PLATFORM_IPC4_MAX_CPC;
```
ujfalusi, line 300: Is this compiles on no ACE platforms?

It might be better to
```c
/* Use maximum clock budget, assume 1ms chunk size */
cd->cpc = CLK_MAX_CPU_HZ / 1000;
```

src/audio/pipeline/pipeline-stream.c
```diff
@@ -296,13 +296,15 @@ static int add_pipeline_cps_consumption(struct comp_dev *current,
...
+		tr_warn(pipe, "0 CPS requested for module: %#x, core: %d using safe max CPC: %d",
+			current->ipc_config.id, ppl_data->p->core, cd->cpc);
 	}
+	int kcps = cd->cpc * 1000 / ppl_data->p->period;
```
ujfalusi, line 304: Hrm, so the CPC is expected as `KCPS` and than it is converted to `CPS` (cps = cd->cpc * 1000) then to `CPC` (cps / ppl_data->p->period)?

PR #7575, opened 2023-05-09
[mtl-005] pipeline: make CPC data "opt-in" with fallback

ujfalusi:
CPC is used to select the correct clock for the desired pipeline workload.

In the case where the CPC data is not available from the manifest and the kernel driver sends a CPC of 0 then SOF should pick a platform safe level to ensure operational glitch free audio at the expense of power consumption. i.e. it's better to prioritize and preserve audio quality over audio power when CPC data is missing.

Replacing #7574

### DIFF
--- a/src/audio/pipeline/pipeline-stream.c
+++ b/src/audio/pipeline/pipeline-stream.c
@@ -296,13 +296,17 @@ static int add_pipeline_cps_consumption(struct comp_dev *current,
 		cd = &md->cfg.base_cfg;
 	}
 
-	int kcps = cd->cpc * 1000 / ppl_data->p->period;
-	if (kcps == 0) {
-		tr_warn(pipe, "0 KCPS for module: %#x, core: %d", current->ipc_config.id, ppl_data->p->core);
-	} else {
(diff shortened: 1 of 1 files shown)


src/audio/pipeline/pipeline-stream.c
```diff
@@ -296,14 +296,17 @@ static int add_pipeline_cps_consumption(struct comp_dev *current,
...
+	if (cd->cpc == 0) {
+		/* Use maximum clock budget, assume 1ms chunk size */
+		cd->cpc = CLK_MAX_CPU_HZ / 1000;
+		tr_warn(pipe,
+			"0 CPS requested for module: %#x, core: %d using safe max CPC: %u",
+			current->ipc_config.id, ppl_data->p->core, cd->cpc);
 	}
 
+	int kcps = cd->cpc * 1000 / ppl_data->p->period;
+	core_kcps_adjust(ppl_data->p->core, kcps);
```
ujfalusi, line 308: I have missed the `tr_info()` print, I will submit an update.

ujfalusi:
Changes since v1:
- added the tr_info() print back

PR #7576, opened 2023-05-09
cadence: enable processing with ipc4 protocol

makarukp:
Maintain codec configuration initialization for ipc4 protocol in cadence
module by separation critical functions into ipc3/ipc4 versions

Changes affect:
- different size of init/runtime config allocation
- resolving codec id from configuration from configuration params for ipc4

### DIFF
--- a/src/audio/module_adapter/module/cadence.c
+++ b/src/audio/module_adapter/module/cadence.c
@@ -113,6 +113,49 @@ static int cadence_code_get_api_id(uint32_t compress_id)
 	}
 }
 
+#if CONFIG_IPC_MAJOR_4
+static int cadence_codec_resolve_api(struct processing_module *mod)
+{
+	int ret;
+	struct snd_codec codec_params;
+	struct comp_dev *dev = mod->dev;
+	struct cadence_codec_data *cd = module_get_private_data(mod);
+	uint32_t api_id = CODEC_GET_API_ID(DEFAULT_CODEC_ID);
+	uint32_t n_apis = ARRAY_SIZE(cadence_api_table);
+	struct module_data *codec = &mod->priv;
+	struct module_param *param;
+	int i;
+	xa_codec_func_t *api = NULL;
+
+	/* For ipc4 protocol codec parameters has to be retrieved from configuration */
+	if (!codec->cfg.data) {
+		comp_err(dev, "cadence_codec_resolve_api(): could not find cadence config");
+		return -EINVAL;
+	}
+	param = codec->cfg.data;
+	api_id = param->id >> 16;
+
+	/* Find and assign API function */
+	for (i = 0; i < n_apis; i++) {
+		if (cadence_api_table[i].id == api_id) {
+			api = cadence_api_table[i].api;
+			break;
+		}
+	}
+
+	/* Verify API assignment */
+	if (!api) {
+		comp_err(dev, "cadence_codec_resolve_api(): could not find API function for id %x",
+			 api_id);
+		return -EINVAL;
+	}
+	cd->api = api;
+	cd->api_id = api_id;
+
+	return 0;
+}
+
+#elif CONFIG_IPC_MAJOR_3
 static int cadence_codec_resolve_api(struct processing_module *mod)
 {
 	int ret;
@@ -122,6 +165,7 @@ static int cadence_codec_resolve_api(struct processing_module *mod)
 	uint32_t api_id = CODEC_GET_API_ID(DEFAULT_CODEC_ID);
 	uint32_t n_apis = ARRAY_SIZE(cadence_api_table);
 	int i;
+	xa_codec_func_t *api = NULL;
 
 	if (mod->stream_params->ext_data_length) {
 		ret = memcpy_s(&codec_params, mod->stream_params->ext_data_length,
@@ -140,22 +184,25 @@ static int cadence_codec_resolve_api(struct processing_module *mod)
 	/* Find and assign API function */
 	for (i = 0; i < n_apis; i++) {
 		if (cadence_api_table[i].id == api_id) {
-			cd->api = cadence_api_table[i].api;
+			api = cadence_api_table[i].api;
 			break;
 		}
 	}
 
 	/* Verify API assignment */
-	if (!cd->api) {
+	if (!api) {
 		comp_err(dev, "cadence_codec_resolve_api(): could not find API function for id %x",
 			 api_id);
 		return -EINVAL;
 	}
-
+	cd->api = api;
 	cd->api_id = api_id;
 
 	return 0;
 }
+#else
+#error Unknown IPC major version
+#endif
 
 static int cadence_codec_post_init(struct processing_module *mod)
 {
@@ -207,16 +254,19 @@ static int cadence_codec_post_init(struct processing_module *mod)
 	return 0;
 }
 
+#if CONFIG_IPC_MAJOR_4
 static int cadence_codec_init(struct processing_module *mod)
 {
-	int ret;
-	struct comp_dev *dev = mod->dev;
+	const struct ipc4_cadence_module_cfg *cfg;
 	struct module_data *codec = &mod->priv;
-	struct cadence_codec_data *cd = NULL;
+	struct cadence_codec_data *cd;
+	struct module_config *setup_cfg;
+	struct comp_dev *dev = mod->dev;
+	int ret;
 
 	comp_dbg(dev, "cadence_codec_init() start");
 
-	cd = rballoc(0, SOF_MEM_CAPS_RAM, sizeof(struct cadence_codec_data));
+	cd = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(struct cadence_codec_data));
 	if (!cd) {
 		comp_err(dev, "cadence_codec_init(): failed to allocate memory for cadence codec data");
 		return -ENOMEM;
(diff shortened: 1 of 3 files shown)


src/audio/module_adapter/module/cadence.c
```diff
@@ -121,8 +121,20 @@ static int cadence_codec_resolve_api(struct processing_module *mod)
 	struct cadence_codec_data *cd = module_get_private_data(mod);
 	uint32_t api_id = CODEC_GET_API_ID(DEFAULT_CODEC_ID);
 	uint32_t n_apis = ARRAY_SIZE(cadence_api_table);
+#if CONFIG_IPC_MAJOR_4
```
dbaluta, line 124: Can you explain a little bit the context and why this change is needed?

src/audio/module_adapter/module/cadence.c
```diff
@@ -209,11 +222,15 @@ static int cadence_codec_post_init(struct processing_module *mod)
...
+	const struct ipc4_cadence_module_cfg *cfg;
+#endif
 	int ret;
 	struct comp_dev *dev = mod->dev;
 	struct module_data *codec = &mod->priv;
 	struct cadence_codec_data *cd = NULL;
 
+
 	comp_dbg(dev, "cadence_codec_init() start");
```
dbaluta, line 234: Remove new line here.

src/audio/module_adapter/module/cadence.c
makarukp: Current implementation does not work with ipc4. In order to enable ipc4 flow api id has to be retrieved from initial configuration.

dbaluta: I don't like this. You are taking the easy route now, sprinkling ifdefs all over the place :) . This can hurt us on the long term. 

Can you find a solution and make it work in the same way it works for IPC3? Otherwise, give a thorough explanation in the commit message why the current solution doesn't work for IPC4 and why we need to change it.

:)

marcinszkudlinski:
if the differences between IPC3/4 are too significant to unify, maybe use a separate functions (or flies cadence_ipc3.c / cadence_ipc4.c)

src/audio/module_adapter/module/cadence.c
makarukp: I am trying to figure out how to obtain any better looking solution at the moment. Initial configuration payload has different content and is serviced in different way for ipc3 and ipc4. I could not find better place than init function to allocate memory for all necessary configuration. There are more protocol principal differences which makes it more difficult to substitute 1:1 ipc3/ipc4 . For example I am not sure if it is right direction  to resolve codec id from extended stream parameters which are not available in ipc4.
By now I could propose rewriting two separate functions for ipc3 and ipc4 which could make it more readable. Would it look better for you ? :-)

I see that similar approach is used in EQ_FIR module.

makarukp:
SOFCI TEST

src/audio/module_adapter/module/cadence.c
```diff
@@ -113,6 +113,49 @@ static int cadence_code_get_api_id(uint32_t compress_id)
...
+		return -EINVAL;
+	}
+
+	cd->api_id = api_id;
+
+	return 0;
+}
+#endif
+
+#if CONFIG_IPC_MAJOR_3
```
lyakh, line 158: `#elif`

src/audio/module_adapter/module/cadence.c
```diff
@@ -207,12 +251,95 @@ static int cadence_codec_post_init(struct processing_module *mod)
...
+	const struct ipc4_cadence_module_cfg *cfg;
+	struct module_data *codec = &mod->priv;
+	struct cadence_codec_data *cd = NULL;
+	struct module_config *setup_cfg;
 	struct comp_dev *dev = mod->dev;
+	int ret;
+
+	comp_dbg(dev, "cadence_codec_init() start");
+
+	cd = rballoc(0, SOF_MEM_CAPS_RAM, sizeof(struct cadence_codec_data));
```
lyakh, line 266: why `rballoc()` for a structure?

src/audio/module_adapter/module/cadence.c
```diff
@@ -207,12 +251,95 @@ static int cadence_codec_post_init(struct processing_module *mod)
...
+		/* allocate memory for set up config */
+		setup_cfg->data = rballoc(0, SOF_MEM_CAPS_RAM, cfg->param_size);
+		if (!setup_cfg->data) {
+			comp_err(dev, "cadence_codec_init(): failed to alloc setup config");
+			ret = -ENOMEM;
+			goto free;
+		}
+
+		/* allocate memory for runtime set up config */
+		codec->cfg.data = rballoc(0, SOF_MEM_CAPS_RAM, cfg->param_size);
```
lyakh, line 294: why `rballoc()` here and in line 286 above? Please review allocation in all the code

src/audio/module_adapter/module/cadence.c
```diff
@@ -113,6 +113,49 @@ static int cadence_code_get_api_id(uint32_t compress_id)
 	}
 }
 
+#if CONFIG_IPC_MAJOR_4
```
btian1, line 116: to avoid this occur multiple times in one source file, normally use:

#if CONFIG_IPC_MAJOR_3
....
#elif #if CONFIG_IPC_MAJOR_4
....
#else
#error Unknown IPC major version
#endif

btian1 (commented):
one more extra question: what kind of codec planned to introduce from cadence?

src/audio/module_adapter/module/cadence.c
```diff
@@ -207,16 +254,100 @@ static int cadence_codec_post_init(struct processing_module *mod)
...
+		comp_err(dev, "cadence_codec_init(): failed to allocate memory for cadence codec data");
+		return -ENOMEM;
+	}
+
+	codec->private = cd;
+	codec->mpd.init_done = 0;
+	cd->self = NULL;
+	cd->mem_tabs = NULL;
+	cd->api = NULL;
+	cd->setup_cfg.avail = false;
```
lyakh, line 280: why did you use `rzalloc()` then and now you're initialising values to zeros again?

src/audio/module_adapter/module/cadence.c
```diff
@@ -207,16 +254,100 @@ static int cadence_codec_post_init(struct processing_module *mod)
...
+					  cfg->param_size);
+		if (!codec->cfg.data) {
+			comp_err(dev, "cadence_codec_init(): failed to alloc runtime setup config");
+			ret = -ENOMEM;
+			goto free_cfg;
+		}
+
+		codec->cfg.size = cfg->param_size;
+		ret = memcpy_s(codec->cfg.data, codec->cfg.size,
+			       cfg->param, cfg->param_size);
```
lyakh, line 308: you're overwriting the whole just allocated `codec->cfg.data`, why did you allocate it with `rzalloc()` then? Also, using different sizes for source and destination in `memcpy_s()` above is more misleading here than logical. You allocated `codec->cfg.data` with size `cfg->param_size` so, **that** is its size. I think it would look more logical to me as
```
tmp_size = cfg->param_size;
codec->cfg.data = rmalloc(..., tmp_size);
...
ret = memcpy_s(codec->cfg.data, tmp_size, cfg->param, cfg->param_size);
codec->cfg.size = tmp_size;
```

src/audio/module_adapter/module/cadence.c
```diff
@@ -207,16 +254,100 @@ static int cadence_codec_post_init(struct processing_module *mod)
...
+		if (ret) {
+			comp_err(dev, "cadence_codec_init(): failed to init runtime config %d",
+				 ret);
+			goto free_cfg2;
+		}
+		codec->cfg.avail = true;
+
+		setup_cfg->size = cfg->param_size;
+		ret = memcpy_s(setup_cfg->data, setup_cfg->size,
+			       cfg->param, cfg->param_size);
```
lyakh, line 318: same here - it's overwritten, no need for `rzalloc()`

src/audio/module_adapter/module/cadence.c
```diff
@@ -207,16 +254,100 @@ static int cadence_codec_post_init(struct processing_module *mod)
...
-	struct cadence_codec_data *cd = NULL;
+	struct cadence_codec_data *cd;
+	struct comp_dev *dev = mod->dev;
+	struct module_config *setup_cfg;
+	int ret;
 
 	comp_dbg(dev, "cadence_codec_init() start");
 
-	cd = rballoc(0, SOF_MEM_CAPS_RAM, sizeof(struct cadence_codec_data));
+	cd = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(struct cadence_codec_data));
```
lyakh, line 350: many fields are initialised to 0 below. Either remove those initialisations or use `rmalloc()`

src/audio/module_adapter/module/cadence.c
```diff
@@ -231,10 +362,11 @@ static int cadence_codec_init(struct processing_module *mod)
 
 	/* copy the setup config only for the first init */
 	if (codec->state == MODULE_DISABLED && codec->cfg.avail) {
-		struct module_config *setup_cfg = &cd->setup_cfg;
+		setup_cfg = &cd->setup_cfg;
 
 		/* allocate memory for set up config */
-		setup_cfg->data = rballoc(0, SOF_MEM_CAPS_RAM, codec->cfg.size);
+		setup_cfg->data = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
+					  codec->cfg.size);
```
lyakh, line 369: same - it's overwritten. In general - usually `data` means some unstructured data and as such initialising it to 0 rarely makes sense.

src/audio/module_adapter/module/cadence.c
mbukowsk: There is no sense to use memcpy_s  with same argument passed for dest_size and count. For that purpose we have basic memcpy(). Safe version additionaly checks destination buffer size if data to be copied fits the buffer. In my opinion this is correct usage of _s function.

lyakh: @mbukowsk I disagree. Firstly just using an additional variable to *hide* the fact that you're using the same value doesn't suddenly make the code meaningful. Secondly `memcpy_s()` *still* makes sense when used with the same size - it verifies that the source and the destination don't overlap. Thirdly your version uses the same value just as much as mine. The difference is that in my version the same variable is at least used both for the allocation and copying, that way it is clear that that is the size of the buffer. Even though I'm personally not a fan of those "safe" versions, if we do use them, let's use them in the best possible way.

mwasko (approved):

iganakov (approved):

Issue #7548, opened 2023-04-26
[ADL/RPL] [Bug] BT HFP offload capture stops working from second time


Vamshigopal:
[ transferred - previously filed to SOF Linux driver as https://github.com/thesofproject/linux/issues/4330 ]

**Description**
with BToffload enabled , when we connect HFP-only supported bt headset, mic stops working from second try,  from sof trace logs we see 
[    85209608.124491] (        1869.218628) c0 dw-dma                 src/drivers/dw/dma.c:1062 INFO dw_dma_avail_data_size() size is 0!
[    85209633.593240] (          25.468748) c0 dw-dma                 src/drivers/dw/dma.c:1128 ERROR dw_dma_get_data_size(): xrun detected
[dmesg.log](https://github.com/thesofproject/linux/files/11333182/dmesg.log)

[sof-logger.log](https://github.com/thesofproject/linux/files/11333132/sof-logger.log)

[btsnoop.log](https://github.com/thesofproject/linux/files/11333133/btsnoop.log)

**steps to reproduce**
1. Connect to an HFP-only bt headset
2. Start a youtube clip via BT output
3. Start mic recording via BT input
4. Stop mic recording via BT input
5. Start mic recording via BT input

**Expected**
 Mic should start at step 5

**Result**
 Mic fails to start 

**repro rate**
10/10

**SW Configuration**
kernel:  https://chromium.googlesource.com/chromiumos/third_party/kernel/+/refs/heads/chromeos-5.15
sof FW: https://github.com/thesofproject/sof/tree/cavs2.5-001-drop-stable
sof topology:  sof-adl-max98360a-rt5682.tplg

kv2019i:
Tested the same FW with nocodec tplg and the problem does not occur. Stopping and restarting SSP record while playback is kept running, works without issues with the nocodec tplg. Only delta between working and non-working configuration is the data format (48kHz/2ch/32bit versus 8kHz/1ch/16bit), which makes this very curious.

With the BT HFP configuration, I can also see the SSP RX FIFOs is getting filled, so the problem would appear to be with DW-DMA not moving data from SSP FIFO to SRAM (on second capture start).

kv2019i:
Narrowed it now down to:.
 - 8kHz 1ch 16bit -> fails the test sequence
 - 8kHz 2ch 16bit -> passes the sequence

Dumps of hw state in passing/failing cases have not revealed anything yet, so still no explanation why this happens.

kv2019i:
[...]

PR #7577, opened 2023-05-09
intel: ssp: drain RX fifo when starting

kv2019i:
Mainline version of https://github.com/thesofproject/sof/pull/7564 

Link: https://github.com/thesofproject/sof/issues/7548

### DIFF
--- a/src/drivers/intel/ssp/ssp.c
+++ b/src/drivers/intel/ssp/ssp.c
@@ -975,6 +975,10 @@ static void ssp_early_start(struct dai *dai, int direction)
 
 	key = k_spin_lock(&dai->lock);
 
+	/* RX fifo must be cleared before start */
+	if (direction == DAI_DIR_CAPTURE)
+		ssp_empty_rx_fifo(dai);
+
 	/* request mclk/bclk */
 	ssp_pre_start(dai);
 
--- a/src/drivers/intel/ssp/ssp.c
+++ b/src/drivers/intel/ssp/ssp.c
@@ -74,8 +74,17 @@ static void ssp_empty_rx_fifo(struct dai *dai)
 	uint64_t sample_ticks = clock_ticks_per_sample(PLATFORM_DEFAULT_CLOCK,
 						       ssp->params.fsync_rate);
 	uint32_t retry = SSP_RX_FLUSH_RETRY_MAX;
+	bool direct_reads = ssp->state[DAI_DIR_CAPTURE] <= COMP_STATE_PREPARE;
 	uint32_t entries;
 	uint32_t i;
+
+#if CONFIG_DMA_SUSPEND_DRAIN
+	/*
+	 * In drain mode, DMA is stopped before DAI, so flush must be
+	 * always done with direct register read.
+	 */
+	direct_reads = true;
+#endif
 
 	/*
 	 * To make sure all the RX FIFO entries are read out for the flushing,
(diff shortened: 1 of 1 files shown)


kv2019i:
FYI @bardliao who worked on this area with the ES_BCLK feature.

src/drivers/intel/ssp/ssp.c
```diff
@@ -975,6 +975,11 @@ static void ssp_early_start(struct dai *dai, int direction)
 
 	key = k_spin_lock(&dai->lock);
 
+	if (direction == DAI_DIR_CAPTURE) {
+		/* RX fifo must be cleared before start */
+		ssp_empty_rx_fifo(dai);
+	}
```
ujfalusi, line 981: niptick:
```c
/* Before starting a capture the RX FIFO must be cleared */
if (direction == DAI_DIR_CAPTURE)
	ssp_empty_rx_fifo(dai);
```

src/drivers/intel/ssp/ssp.c
```diff
@@ -87,9 +96,12 @@ static void ssp_empty_rx_fifo(struct dai *dai)
...
 		entries = SSCR3_RFL_VAL(ssp_read(dai, SSCR3));
 		dai_dbg(dai, "ssp_empty_rx_fifo(), before flushing, entries %d", entries);
-		for (i = 0; i < entries + 1; i++)
-			/* read to try empty fifo */
-			ssp_read(dai, SSDR);
+
+		if (read_ssdr) {
+			for (i = 0; i < entries + 1; i++)
+				/* read to try empty fifo */
+				ssp_read(dai, SSDR);
```
ujfalusi, line 103: nitpick: I would move the comment up, before the for() loop and reword it to `Empty the FIFO by reading out the entries`

kv2019i: Done

src/drivers/intel/ssp/ssp.c
kv2019i: Changed (although noting that in Zephyr, braces are mandatory here, trying to learn this style as this gets us rejects on Zephyr side all the time).

ujfalusi: Well, if you have the comment inside of the if, then yes, the braces help, but I suggested to move it _outside_ of the if:
```c
/* Before starting a capture the RX FIFO must be cleared */
if (direction == DAI_DIR_CAPTURE)
	ssp_empty_rx_fifo(dai);
```
Does Zephyr mandates that braces are mandatory in all instance?

ujfalusi (approved):

src/drivers/intel/ssp/ssp.c
```diff
@@ -74,9 +74,18 @@ static void ssp_empty_rx_fifo(struct dai *dai)
 	uint64_t sample_ticks = clock_ticks_per_sample(PLATFORM_DEFAULT_CLOCK,
 						       ssp->params.fsync_rate);
 	uint32_t retry = SSP_RX_FLUSH_RETRY_MAX;
+	bool direct_reads = ssp->state[DAI_DIR_CAPTURE] <= COMP_STATE_PREPARE;
```
btian1, line 77: how about COMP_STATE_PAUSED? it is not active, and seems will not read with this state.

kv2019i: @btian1 That's a good question. Earlier version of the patch broke PAUSE, so I did spend some time looking at this. Curiously, ssp_pause does not stop the hardware and it based on tests, this is not needed after pause. I can't fully explain why this is the case (and DMA is stopped on pause).

ujfalusi: the other patch fixes the pause by yanking the FIFO before resume (which is a start operation), it is really strange to be honest, but w/o the FIFO drain patch I had broken pause/resume and I think @kv2019i had a working one on his (identical) setup.

PR #7588, opened 2023-05-11
Dummy smart_amp: copy input_pins and output_pin separately

andrula-song:
The original code confuses static code analyzers since it was
relying on the fact that we have the 2x input and 1x output pin
config in adjacent position in smart_amp_data struct similarly
to the extended module configuration.

While it works, it is not a good practice.

Split the copy of input and output pin formats to make the code
obvious and less error prone.

Fixes: https://github.com/andrula-song/sof/commit/bcc14074cd2b80d2e2a4803932daf6a961b3cb0a ("smart_amp_test: Split the module config and blob receiving for IPC4")

Signed-off-by: Andrula Song <andrula.song@intel.com>

### DIFF
--- a/src/samples/audio/smart_amp_test_ipc4.c
+++ b/src/samples/audio/smart_amp_test_ipc4.c
@@ -44,7 +44,8 @@ static int smart_amp_init(struct processing_module *mod)
 	struct smart_amp_data *sad;
 	struct comp_dev *dev = mod->dev;
 	struct module_data *mod_data = &mod->priv;
-	int bs;
+	const size_t in_size = sizeof(struct ipc4_input_pin_format) * SMART_AMP_NUM_IN_PINS;
+	const size_t out_size = sizeof(struct ipc4_output_pin_format) * SMART_AMP_NUM_OUT_PINS;
 	int ret;
 	const struct ipc4_base_module_extended_cfg *base_cfg = mod_data->cfg.init_data;
 
@@ -70,9 +71,10 @@ static int smart_amp_init(struct processing_module *mod)
 	}
 
 	/* Copy the pin formats */
-	bs = sizeof(sad->ipc4_cfg.input_pins) + sizeof(sad->ipc4_cfg.output_pin);
-	memcpy_s(sad->ipc4_cfg.input_pins, bs,
-		 base_cfg->base_cfg_ext.pin_formats, bs);
+	memcpy_s(sad->ipc4_cfg.input_pins, in_size,
+		 base_cfg->base_cfg_ext.pin_formats, in_size);
+	memcpy_s(&sad->ipc4_cfg.output_pin, out_size,
+		 &base_cfg->base_cfg_ext.pin_formats[in_size], out_size);
 
 	mod->simple_copy = true;
 


src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -71,7 +71,7 @@ static int smart_amp_init(struct processing_module *mod)
 
 	/* Copy the pin formats */
 	bs = sizeof(sad->ipc4_cfg.input_pins) + sizeof(sad->ipc4_cfg.output_pin);
-	memcpy_s(sad->ipc4_cfg.input_pins, bs,
+	memcpy_s(&sad->ipc4_cfg.input_pins[0], bs,
```
marc-hb, line 74: Leaving the specific analyzer aside, can you explain in the commit message what difference this change makes with respect to buffer overflows?

marc-hb: Just realized this is copying two contiguous fields at once. No surprise the static analyzer is not happy!

It's also a bit strange to use `memcpy_s()` whose only purpose is to force the developer to be extra careful and then use it to "overflow" the copy from one field to the other.

I would either do two copies or use `memcpy`, not `memcpy_s`.

ranj063: I agree there should be separate copies for the input and out pin formats.

andrula-song: Done, thanks.

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -44,7 +44,7 @@ static int smart_amp_init(struct processing_module *mod)
 	struct smart_amp_data *sad;
 	struct comp_dev *dev = mod->dev;
 	struct module_data *mod_data = &mod->priv;
-	int bs;
+	int in_size, out_size;
```
marc-hb, line 47: These are compile-time constants, not variables.
```suggestion
	const int in_size = sizeof(sad->ipc4_cfg.input_pins);
```

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -70,9 +70,12 @@ static int smart_amp_init(struct processing_module *mod)
...
 	/* Copy the pin formats */
-	bs = sizeof(sad->ipc4_cfg.input_pins) + sizeof(sad->ipc4_cfg.output_pin);
-	memcpy_s(sad->ipc4_cfg.input_pins, bs,
-		 base_cfg->base_cfg_ext.pin_formats, bs);
+	in_size = sizeof(sad->ipc4_cfg.input_pins);
+	out_size = sizeof(sad->ipc4_cfg.output_pin);
+	memcpy_s(sad->ipc4_cfg.input_pins, in_size,
+		 base_cfg->base_cfg_ext.pin_formats, in_size);
+	memcpy_s(&sad->ipc4_cfg.output_pin, out_size,
+		 base_cfg->base_cfg_ext.pin_formats + in_size, out_size);
```
marc-hb, line 78: Pointer arithmetic on a flexible array member, wow! I wonder what that means. It looks suspicious... maybe this instead?

```suggestion
		 &base_cfg->base_cfg_ext.pin_formats[0] + in_size, out_size);
```

marc-hb: > Pointer arithmetic on a flexible array member, wow! I wonder what that means.

Most likely the same thing as a regular array, sorry the noise. In that case:

```suggestion
		 &base_cfg->base_cfg_ext.pin_formats[in_size], out_size);
```

btian1: this may cause memory leak, it may add pointer based on pin_format size instead of byte, better continue add (char *) at beginning in case xtensa compiler is not so strong.
(char *)&base_cfg->base_cfg_ext.pin_formats[0] + in_size

or 
(char *)&base_cfg->base_cfg_ext.pin_formats[1] ?

marc-hb: >  in case xtensa compiler is not so strong.

The xtensa compiler is based on either GCC or Clang so there's no standard compliance concern.

paulstelian97 (commented):
Code change is good, commit message does have a tiny issue...

This change makes the code _less_ error prone, not more.

andrula-song:
> Code change is good, commit message does have a tiny issue...
> 
> This change makes the code _less_ error prone, not more.

Thanks.

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -70,9 +71,10 @@ static int smart_amp_init(struct processing_module *mod)
 	}
 
 	/* Copy the pin formats */
-	bs = sizeof(sad->ipc4_cfg.input_pins) + sizeof(sad->ipc4_cfg.output_pin);
-	memcpy_s(sad->ipc4_cfg.input_pins, bs,
-		 base_cfg->base_cfg_ext.pin_formats, bs);
+	memcpy_s(sad->ipc4_cfg.input_pins, in_size,
+		 base_cfg->base_cfg_ext.pin_formats, in_size);
+	memcpy_s(&sad->ipc4_cfg.output_pin, out_size,
+		 &base_cfg->base_cfg_ext.pin_formats[in_size], out_size);
```
ranj063, line 77: is this correct? Shouldnt this be 
```memcpy_s(&sad->ipc4_cfg.output_pin, out_size,
		 &base_cfg->base_cfg_ext.pin_formats[SMART_AMP_NUM_IN_PINS], out_size);
```
or more simply:
``` 
sad->ipc4_cfg.output_pin = base_cfg->base_cfg_ext.pin_formats[SMART_AMP_NUM_IN_PINS];
```

andrula-song: but the type of base_cfg_ext.pin_formats is ### **uint8_t**
struct ipc4_base_module_cfg_ext {
	/* specifies number of items in input_pins array. Maximum size is 8 */
	uint16_t nb_input_pins;
	/* specifies number of items in output_pins array. Maximum size is 8 */
	uint16_t nb_output_pins;
	uint8_t reserved[12];
	/* Specifies format of input pins followed by output pins.
	 * Pin format arrays may be non-continuous i.e. may contain pin #0 format
	 * followed by pin #2 format in case pin #1 will not be in use.
	 * FW assigned format of the pin based on pin_index, not on a position of
	 * the item in the array. Applies to both input and output pins.
	 */
	**uint8_t pin_formats[];**

ranj063: @andrula-song I'm not sure what you're trying to say. what you really want to do is this: 
```
diff --git a/src/samples/audio/smart_amp_test_ipc4.c b/src/samples/audio/smart_amp_test_ipc4.c
index e37fe3f74..e45c35d2e 100644
--- a/src/samples/audio/smart_amp_test_ipc4.c
+++ b/src/samples/audio/smart_amp_test_ipc4.c
@@ -47,6 +47,7 @@ static int smart_amp_init(struct processing_module *mod)
        int bs;
        int ret;
        const struct ipc4_base_module_extended_cfg *base_cfg = mod_data->cfg.init_data;
+       const size_t in_size, out_size;
 
        comp_dbg(dev, "smart_amp_init()");
        sad = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*sad));
@@ -70,9 +71,11 @@ static int smart_amp_init(struct processing_module *mod)
        }
 
        /* Copy the pin formats */
-       bs = sizeof(sad->ipc4_cfg.input_pins) + sizeof(sad->ipc4_cfg.output_pin);
-       memcpy_s(sad->ipc4_cfg.input_pins, bs,
-                base_cfg->base_cfg_ext.pin_formats, bs);
+       in_size = sizeof(struct ipc4_input_pin_format) * SMART_AMP_NUM_IN_PINS;
+       memcpy_s(sad->ipc4_cfg.input_pins, in_size, base_cfg->base_cfg_ext.pin_formats, in_size);
+       out_size = sizeof(struct ipc4_output_pin_format) * SMART_AMP_NUM_OUT_PINS;
+       memcpy_s(sad->ipc4_cfg.input_pins, in_size, base_cfg->base_cfg_ext.pin_formats + in_size,
+                in_size);
 
        mod->simple_copy = true;
 
```

andrula-song: aligned, let's use sizeof(struct x) * num to calculate the size to make code more clear.

ranj063 (approved):

andrula-song:
SOFCI TEST

src/samples/audio/smart_amp_test_ipc4.c
lyakh:
```
	memcpy_s(sad->ipc4_cfg.input_pins, sizeof(sad->ipc4_cfg.input_pins),
		 base_cfg->base_cfg_ext.pin_formats, in_size);
	memcpy_s(&sad->ipc4_cfg.output_pin, sizeof(sad->ipc4_cfg.output_pin),
		 &base_cfg->base_cfg_ext.pin_formats[in_size], out_size);
```

andrula-song:
SPFCI TEST

kv2019i (approved):

ranj063:
SOFCI TEST

andrula-song:
SOFCI TEST

kv2019i:
One fail hit in https://sof-ci.01.org/sofpr/PR7588/build7818/devicetest/index.html -- a known bug.

PR #7589, opened 2023-05-11
samples: smart_amp_test_ipc4: Split the pin configuration copy in init

ujfalusi:
The original code confuses static code analyzers since it was relying on the fact that we have the 2x input and 1x output pin config in adjacent position in smart_amp_data struct similarly to the extended module configuration.

While it works, it is not a good practice.

Split the copy of input and output pin formats to make the code obvious and error prone.

Fixes: bcc14074cd2b ("smart_amp_test: Split the module config and blob receiving for IPC4")
Reported-by: Andrey Borisovich <andrey.borisovich@intel.com>

### DIFF
--- a/src/samples/audio/smart_amp_test_ipc4.c
+++ b/src/samples/audio/smart_amp_test_ipc4.c
@@ -44,7 +44,8 @@ static int smart_amp_init(struct processing_module *mod)
 	struct smart_amp_data *sad;
 	struct comp_dev *dev = mod->dev;
 	struct module_data *mod_data = &mod->priv;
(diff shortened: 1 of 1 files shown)


andrula-song:
I have a PR https://github.com/thesofproject/sof/pull/7588 about this too, which one do you like to keep?

lyakh (approved):

ranj063:
> I have a PR #7588 about this too, which one do you like to keep?

@ujfalusi Andrula beat you to this already. Lets go with #7588

PR #7598, opened 2023-05-12
copier: preparation for copier module adapter interface

btian1:
no change for functionality, changes list below:

1. Removed dependency in dai-zephyr.c and dai-legacy.c for comp_dai_get_hw_params with direct calling.
2. Removed one case for COMP_ATTR_VDMA_INDEX in function: copier_get_attribute, since it only used inside copier.
3. Renamed function name with better aligned with name convention.
4. Removed copier get hw parameters function.

This is for preparation of copier module interface.

### DIFF
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -1801,7 +1801,7 @@ static int copier_get_hw_params(struct comp_dev *dev, struct sof_ipc_stream_para
 	if (dev->ipc_config.type != SOF_COMP_DAI)
 		return -EINVAL;
 
-	return dai_zephyr_get_hw_params(dd, dev, params, dir);
+	return dai_common_get_hw_params(dd, dev, params, dir);
 }
 
 static int copier_unbind(struct comp_dev *dev, void *data)
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -522,12 +522,8 @@ static int init_pipeline_reg(struct comp_dev *dev)
 	struct copier_data *cd = comp_get_drvdata(dev);
 	struct ipc4_pipeline_registers pipe_reg;
 	uint32_t gateway_id;
-	int ret;
 
-	ret = comp_get_attribute(dev, COMP_ATTR_VDMA_INDEX, &gateway_id);
-	if (ret)
-		return ret;
-
+	gateway_id = cd->config.gtw_cfg.node_id.f.v_index;
 	if (gateway_id >= IPC4_MAX_PIPELINE_REG_SLOTS) {
 		comp_cl_err(&comp_copier, "gateway_id %u out of array bounds.", gateway_id);
 		return -EINVAL;
@@ -1721,9 +1717,6 @@ static int copier_get_attribute(struct comp_dev *dev, uint32_t type, void *value
 	struct copier_data *cd = comp_get_drvdata(dev);
 
 	switch (type) {
-	case COMP_ATTR_VDMA_INDEX:
-		*(uint32_t *)value = cd->config.gtw_cfg.node_id.f.v_index;
-		break;
 	case COMP_ATTR_BASE_CONFIG:
 		*(struct ipc4_base_module_cfg *)value = cd->config.base;
 		break;
--- a/src/audio/dai-legacy.c
+++ b/src/audio/dai-legacy.c
@@ -264,7 +264,7 @@ static void dai_free(struct comp_dev *dev)
 	rfree(dev);
 }
 
-int dai_zephyr_get_hw_params(struct dai_data *dd, struct comp_dev *dev,
+int dai_common_get_hw_params(struct dai_data *dd, struct comp_dev *dev,
 			     struct sof_ipc_stream_params *params, int dir)
 {
 	int ret;
@@ -295,7 +295,7 @@ static int dai_comp_get_hw_params(struct comp_dev *dev,
 {
 	struct dai_data *dd = comp_get_drvdata(dev);
 
-	return dai_zephyr_get_hw_params(dd, dev, params, dir);
+	return dai_common_get_hw_params(dd, dev, params, dir);
 }
 
 static int dai_comp_hw_params(struct comp_dev *dev,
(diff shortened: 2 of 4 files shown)


src/audio/pipeline/pipeline-params.c
```diff
@@ -162,8 +165,18 @@ static int pipeline_comp_hw_params(struct comp_dev *current,
...
+		dd = comp_get_drvdata(current);
+#elif CONFIG_IPC_MAJOR_4
+		struct copier_data *cd = comp_get_drvdata(current);
+
+		dd = cd->dd[0];
+#else
+#error Unknown IPC major version
+#endif
+		/* dai device already removed, currently it is copier */
+		ret = dai_zephyr_get_hw_params(dd, current, &ppl_data->params->params, dir);
```
ranj063, line 179: not true for multi endpoint DAI yet

btian1: ok, seems I put too much in one patch again, let me split and upload with new patches.

src/audio/dai-legacy.c
```diff
@@ -1164,7 +1159,6 @@ static const struct comp_driver comp_dai = {
 		.create				= dai_new,
 		.free				= dai_free,
 		.params				= dai_params,
-		.dai_get_hw_params		= dai_comp_get_hw_params,
```
ranj063, line 1167: I think you need to keep this for now.

src/audio/copier/copier.c
```diff
@@ -646,11 +646,8 @@ static int init_pipeline_reg(struct comp_dev *dev)
 	struct copier_data *cd = comp_get_drvdata(dev);
 	struct ipc4_pipeline_registers pipe_reg;
 	uint32_t gateway_id;
-	int ret;
 
-	ret = comp_get_attribute(dev, COMP_ATTR_VDMA_INDEX, &gateway_id);
-	if (ret)
-		return ret;
+	gateway_id = cd->config.gtw_cfg.node_id.f.v_index;
```
RanderWang, line 650: remove blank line after this line

RanderWang (commented):
LGTM

ranj063 (approved):

ranj063:
@btian1 ADL nocodec has failures on playback. Can you please check?

RanderWang (approved):

src/audio/dai-legacy.c
```diff
@@ -317,12 +317,15 @@ static int dai_comp_hw_params(struct comp_dev *dev,
...
 
-static int dai_verify_params(struct comp_dev *dev,
+static int dai_verify_params(struct dai_data *dd, struct comp_dev *dev,
 			     struct sof_ipc_stream_params *params)
 {
 	struct sof_ipc_stream_params hw_params;
+	int ret;
 
-	comp_dai_get_hw_params(dev, &hw_params, params->direction);
+	ret = dai_zephyr_get_hw_params(dd, dev, &hw_params, params->direction);
```
kv2019i, line 326: @btian1 @ranj063 Hmm. This is getting very confusing to read. Why do we use "dai_zephyr_" name for both legacy and Zephyr? This seems to have started in 79c43394892e05bb4778c7b30c87a007dd41d3be . I mean, dai-legacy.c is specifically NOT using Zephyr DAI interface, so it makes the code very hard to read. FYI @juimonen

kv2019i (approved):
This PR is good, but @btian1 @ranj063 please see comment about the confusing use of dai_zephyr_foo() function naming.

src/audio/dai-legacy.c
lyakh: @kv2019i yeah, it is confusing, but it's what we have both in host-{zephyr,legacy}.c and in dai-{zephyr,legacy}.c.

ranj063: @kv2019i it is confusing but the copier could in theory use either dai-legacy or dai-zephyr and this common function name allows us to unify the copier for both cases. Maybe dai_zephyr_ isnt the best prefix. Should we just remove the zephyr from the name and replace it with dai_common instad of dai_zephyr?

btian1: let me remove zephyr prefix and submit with a new patchset

btian1:
> 

this is done with name change, please review and check again.

btian1:
this PR should be ready for merge, @kv2019i @ranj063 , please check and merge it.

src/audio/pipeline/pipeline-params.c
btian1: since multi-endpoint PR merged, seems this part of change need involve again, let me try to submit a new patch to cover this.

src/audio/pipeline/pipeline-params.c
```diff
@@ -162,8 +165,27 @@ static int pipeline_comp_hw_params(struct comp_dev *current,
...
+		/* dai-legacy case */
+		if (current->drv->ops.dai_get_hw_params)
+			ret = current->drv->ops.dai_get_hw_params(current,
+								  &ppl_data->params->params, dir);
+#endif
+#elif CONFIG_IPC_MAJOR_4
+		struct copier_data *cd = comp_get_drvdata(current);
+		struct dai_data *dd = cd->dd[0];
+
+		ret = dai_common_get_hw_params(dd, current, &ppl_data->params->params, dir);
```
lyakh, line 185: Not specific to this commit, but with the recent copier / DAI integration work, I'm a bit concerned about mixing up of the layers. Now we have copier and DAI layers mixing in with the pipeline code. Is that an intermediate step or is it what we want eventually?

btian1: I understand.

Before this change, the get hw params operation still looks very strange, it calls it by an inline ops driver interface, and this interface is not an standard for module adapter, so this is why this patch exists.

Regarding the change, due to we have to support multiple cases(otherwise, CI will report error)
looks a little redundant, maybe one day, ipc3 support was totally removed from SOF, then we may figure out a better way to implement this get hw params.

currently, seems this is the only way to remove this interface from copier.

lyakh (approved):

RanderWang (approved):

src/audio/dai-legacy.c
kv2019i: Let's handle in a follow-up.

kv2019i (approved):

kv2019i:
Known fails in https://sof-ci.01.org/sofpr/PR7598/build8568/devicetest/index.html and https://sof-ci.01.org/sofpr/PR7598/build8567/devicetest/index.html . Proceeding with merge.

PR #7606, opened 2023-05-12
Multiple endpoint DAI copier optimization

ranj063:
Removes the endpoint DAI devices and endpoint buffers for multiple endpoint DAI copiers and saves an extra copy.

### DIFF
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -304,79 +304,6 @@ static enum sof_ipc_stream_direction
 	}
 }
 
-static int init_dai_single(struct comp_dev *parent_dev,
-			   const struct comp_driver *drv,
-			   struct comp_ipc_config *config,
-			   const struct ipc4_copier_module_cfg *copier,
-			   struct pipeline *pipeline,
-			   struct ipc_config_dai *dai,
-			   enum ipc4_gateway_type type)
-{
-	struct copier_data *cd;
-	struct dai_data *dd;
-	int ret;
-
-	cd = comp_get_drvdata(parent_dev);
-
-	if (cd->direction == SOF_IPC_STREAM_PLAYBACK) {
-		enum sof_ipc_frame out_frame_fmt, out_valid_fmt;
-
-		audio_stream_fmt_conversion(copier->out_fmt.depth,
-					    copier->out_fmt.valid_bit_depth,
-					    &out_frame_fmt,
-					    &out_valid_fmt,
-					    copier->out_fmt.s_type);
-		config->frame_fmt = out_frame_fmt;
-		pipeline->sink_comp = parent_dev;
-	} else {
-		enum sof_ipc_frame in_frame_fmt, in_valid_fmt;
-
-		audio_stream_fmt_conversion(copier->base.audio_fmt.depth,
-					    copier->base.audio_fmt.valid_bit_depth,
-					    &in_frame_fmt, &in_valid_fmt,
-					    copier->base.audio_fmt.s_type);
-		config->frame_fmt = in_frame_fmt;
-		pipeline->source_comp = parent_dev;
-	}
-
-	parent_dev->ipc_config.frame_fmt = config->frame_fmt;
-
-	dd = rzalloc(SOF_MEM_ZONE_RUNTIME_SHARED, 0, SOF_MEM_CAPS_RAM, sizeof(*dd));
-	if (!dd)
-		return -ENOMEM;
-
-	ret = dai_zephyr_new(dd, parent_dev, dai);
-	if (ret < 0)
-		goto e_dd;
-
-	pipeline->sched_id = config->id;
-
-	ret = comp_dai_config(dd, parent_dev, dai, copier);
-	if (ret < 0)
-		goto e_zephyr;
-
-	cd->converter[IPC4_COPIER_GATEWAY_PIN] =
-			get_converter_func(&copier->base.audio_fmt, &copier->out_fmt, type,
-					   IPC4_DIRECTION(dai->direction));
-	if (!cd->converter[IPC4_COPIER_GATEWAY_PIN]) {
-		comp_err(parent_dev, "failed to get converter type %d, dir %d",
-			 type, dai->direction);
-		ret = -EINVAL;
-		goto e_zephyr;
-	}
-
-	cd->endpoint_num++;
-	cd->dd[0] = dd;
-
-	return 0;
-
-e_zephyr:
-	dai_zephyr_free(dd);
-e_dd:
-	rfree(dd);
-	return ret;
-}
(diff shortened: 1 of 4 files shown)


src/include/ipc4/copier.h
```diff
@@ -246,7 +240,7 @@ struct copier_data {
 
 	/* buffer to mux/demux data from/to multiple endpoint buffers for ALH multi-gateway case */
 	struct comp_buffer *multi_endpoint_buffer;
-	channel_copy_func copy_single_channel;
+	pcm_converter_func copy_single_channel;
```
btian1, line 243: simply change the function type? can we further assign pcm converter function instead of use above 2 function?
i.e, use audio_stream_copy in pcm converter to cover this?

RanderWang (approved):
nice work !

src/audio/dai-zephyr.c
```diff
@@ -1280,6 +1346,128 @@ static void dai_report_xrun(struct dai_data *dd, struct comp_dev *dev, uint32_t
...
+			break;
+		case -EPIPE:
+			/* DMA status can return -EPIPE and current status content if xrun occurs */
+			if (direction == SOF_IPC_STREAM_PLAYBACK)
+				comp_dbg(dev, "dai_zephyr_multi_endpoint_copy(): dma_get_status() underrun occurred, endpoint: %d ret = %u",
+					 i, ret);
+			else
+				comp_dbg(dev, "dai_zephyr_multi_endpoint_copy(): dma_get_status() overrun occurred, enpdoint: %d ret = %u",
+					 i, ret);
+			break;
```
kv2019i, line 1382: If we break on L1389, free_bytes and avail_bytes can in theory be INT32_MAX when we exit the loop?

kv2019i (commented):
One question inline, otherwise looks good.

src/audio/copier/copier.c
```diff
@@ -419,6 +399,11 @@ static int init_dai(struct comp_dev *parent_dev,
 	if (ret < 0)
 		return ret;
 
+	/* save the channel map and count for all endpoints */
+	ret = copier_set_channel_map(parent_dev, copier, type, index);
+	if (ret < 0)
```
lyakh: I think you need to destroy the buffer in case of error.

src/audio/copier/copier.c
```diff
@@ -363,16 +363,6 @@ static int init_dai(struct comp_dev *parent_dev,
 	if (ret < 0)
 		goto e_zephyr_free;
 
-	cd->converter[IPC4_COPIER_GATEWAY_PIN] =
-			get_converter_func(&copier->base.audio_fmt, &copier->out_fmt, type,
-					   IPC4_DIRECTION(dai->direction));
```
lyakh, line 446: formats cannot change between multiple calls to `init_dai()`?

lyakh (commented):
Nice! Just some cosmetic suggestions

btian1 (approved):

src/audio/dai-zephyr.c
```diff
@@ -340,6 +340,72 @@ dai_dma_cb(struct dai_data *dd, struct comp_dev *dev, uint32_t bytes,
...
+			  struct comp_buffer *multi_endpoint_buffer)
+{
+	struct comp_buffer __sparse_cache *multi_buf_c, *dma_buf;
+	enum dma_cb_status dma_status = DMA_CB_STATUS_RELOAD;
+	uint32_t i, bytes;
+
+	comp_dbg(dev, "dai_dma_multi_endpoint_cb()");
+
+	/* stop dma copy for pause/stop/xrun */
+	if (dev->state != COMP_STATE_ACTIVE || dd->xrun) {
```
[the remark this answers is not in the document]
ranj063: I was stumped here as well but I think thats the reason. Even in the single endpoint case, thats what we do. So I left it as is. Maybe something to think about and likely change in the future.

src/audio/copier/copier.c
ranj063: correct, so the last patch moves it out to create_dai() instead

ranj063:
SOFCI TEST

src/audio/dai-zephyr.c
```diff
@@ -1280,6 +1346,128 @@ static void dai_report_xrun(struct dai_data *dd, struct comp_dev *dev, uint32_t
...
+		case 0:
+			break;
+		case -EPIPE:
+			/* DMA status can return -EPIPE and current status content if xrun occurs */
+			if (direction == SOF_IPC_STREAM_PLAYBACK)
+				comp_dbg(dev, "dai_zephyr_multi_endpoint_copy(): dma_get_status() underrun occurred, endpoint: %d ret = %u",
+					 i, ret);
+			else
+				comp_dbg(dev, "dai_zephyr_multi_endpoint_copy(): dma_get_status() overrun occurred, enpdoint: %d ret = %u",
+					 i, ret);
```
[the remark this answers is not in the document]
lyakh: well, I dare suggest - this is `comp_dbg()` and I'm *pretty* sure nobody depends on this exact behaviour and I don't think we ever guaranteed that our debug prints never change

src/audio/dai-zephyr.c
kv2019i: @ranj063 But we are not doing 
````
	avail_bytes = MIN(avail_bytes, stat.pending_length);
		free_bytes = MIN(free_bytes, stat.free); 
``` 
... if we break. Shouldn't these be modified in case of xrun?

lyakh (commented):
Just one request to invert the error case handling in `create_endpoint_buffer()`

src/audio/copier/copier.c
```diff
@@ -201,18 +195,7 @@ static int create_endpoint_buffer(struct comp_dev *parent_dev,
...
-	    (type == ipc4_gtw_alh && is_multi_gateway(copier_cfg->gtw_cfg.node_id))) {
-		ret = copier_set_alh_multi_gtw_channel_map(parent_dev, copier_cfg, buffer,
-							   index, &chan_map);
-		if (ret < 0)
-			return ret;
-	}
 
-	buffer_c = buffer_acquire(buffer);
 	for (i = 0; i < SOF_IPC_MAX_CHANNELS; i++)
 		buffer_c->chmap[i] = (chan_map >> i * 4) & 0xf;
```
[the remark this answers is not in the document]
ranj063: But thats what I mentioned in the commit message that endpoint bufffers are not used in the multi-gateway case anyway. So there's no need to set the chan_map at all

kv2019i (approved):

ranj063:
SOFCI TEST

ranj063:
just usual failures in device tests. merging

PR #7354, opened 2023-03-30
LMDK: Loadable Modules Dev Kit build scripts

serhiy-katsyuba-intel:
Initial commit with set of cmake and linker scripts to build loadable modules binary.

Things yet TODO:
  add all necessary Dev Kit header files to include dir,
  replace dummy example with somewhat useful module example.

PR #7595, opened 2023-05-12
[TEST IGNORE] Big sign switch

marc-hb:

PR #7608, opened 2023-05-15
LMDK: fixed rimage and private key paths discovery

aborisovich:
Path to rimage and private key had been failing to be used while provided in form of relative path using dots.
Changed how rimage is discovered in the scripts:
* rimage is searched for using find_program with paths and hints provided
* changed `RIMAGE_COMMAND` to `RIMAGE_INSTALL_DIR` so end-user may specify directory where rimage executable is placed. It will be used by `find_program`.
* If `RIMAGE_INSTALL_DIR` is not provided, `find_program` will try to search for rimage in the directory where SOF project installs it (`west_dir/build-rimage`).
* Updated README.md with instructions on how to build:
  - simplified configure and build commands
  - updated explanations on how to use RIMAGE_INSTALL_DIR
  - added information on how to use toolchain file
* private key path was parsed incorrectly when provided windows path style using backslashes. It was passed to rimage as is resulting in the invalid path as the last slash was made by rimage as forward slash. Added path normalization what fixed the problem.

This addresses issues I've described during code review in https://github.com/thesofproject/sof/pull/7354#pullrequestreview-1411703077 .


pjdobrowolski (approved):

serhiy-katsyuba-intel (approved):

tmleman (approved):

dnikodem (approved):

aborisovich:
> LDMK is new to me so I unfortunately don't know enough to review this.
> 
> I've been working on removing the explicit `west sign` invocation from `xtensa-build-zephyr.py` and let `west build` do it instead, still a draft PR tested OK, @aborisovich please take a quick look: #7595
> 
> I hope this is independent from LMDK and does not conflict in any way?
> 
> The first step is #7546 which is NOT a draft, ready to review thanks!

> 

Currently LMDK is completely separate instance in terms of a build system.
We have no dependencies to west, zephyr or xtensa-build-zephyr.py script.
I'll have a look in a moment.

PR #7612, opened 2023-05-15
Audio: Volume: Change volume internal gain to Q8.23 from Q1.23

singalsu:
This patch changes for IPC4 configuration the DSP internal gain format from Q1.23 to Q8.23. Minimum needed would be Q2.23 but the integer part 8 bits is same as in IPC3 and would allow signal amplification in volume component in some later phase. The code is prepared for amplification due to IPC3.

This change allows the Q1.31 max gain 0x7fffffff to round to exactly one in Q8.23 notation. The gain value one passes effectively input to output unmodified. The change prevents lose about 1 dB of signal-to-noise ratio from peak quality of PCM formats.

The comment from convert_volume_ipc4_to_ipc3() is removed because it is unrelated and some remain from old code version.

The hard coded 23 for conversion to/from Qx.23 is changed to macro VOL_QXY_Y because it's used elsewhere in the component. The saturation (to 32 bit signed) with 64 bit intermediate is kept in the function convert_volume_ipc4_to_ipc3() in case of VOL_QXY_Y would be 31. The code requires int32_t or ae_f32 compatible signed 32 bit compatible gain value instead of full uint32_t range.

### DIFF
--- a/src/audio/module_adapter/module/volume/volume.c
+++ b/src/audio/module_adapter/module/volume/volume.c
@@ -499,13 +499,7 @@ static int set_volume_ipc4(struct vol_data *cd, uint32_t const channel,
  */
 static inline uint32_t convert_volume_ipc4_to_ipc3(struct comp_dev *dev, uint32_t volume)
 {
-	/* Limit received volume gain to MIN..MAX range before applying it.
-	 * MAX is needed for now for the generic C gain arithmetics to prevent
-	 * multiplication overflow with the 32 bit value. Non-zero MIN option
-	 * can be useful to prevent totally muted small volume gain.
-	 */
-
-	return sat_int24(Q_SHIFT_RND(volume, 31, 23));
+	return sat_int32(Q_SHIFT_RND((int64_t)volume, 31, VOL_QXY_Y));
 }
 
 static inline uint32_t convert_volume_ipc3_to_ipc4(uint32_t volume)
@@ -513,7 +507,7 @@ static inline uint32_t convert_volume_ipc3_to_ipc4(uint32_t volume)
 	/* In IPC4 volume is converted into Q1.23 format to be processed by firmware.
 	 * Now convert it back to Q1.31
 	 */
-	return sat_int32(Q_SHIFT_LEFT((int64_t)volume, 23, 31));
+	return sat_int32(Q_SHIFT_LEFT((int64_t)volume, VOL_QXY_Y, 31));
 }
 
 static inline void init_ramp(struct vol_data *cd, uint32_t curve_duration, uint32_t target_volume)
--- a/src/include/sof/audio/volume.h
+++ b/src/include/sof/audio/volume.h
@@ -65,8 +65,10 @@ struct sof_ipc_ctrl_value_chan;
 //** \brief Volume gain Qx.y */
 #define COMP_VOLUME_Q1_23 1
 
-//** \brief Volume gain Qx.y integer x number of bits including sign bit. */
-#define VOL_QXY_X 1
+/** \brief Volume gain Qx.y integer x number of bits including sign bit.
+ * With Q8.23 format the gain range is -138.47 to +42.14 dB.
+ */
+#define VOL_QXY_X 8
 
 //** \brief Volume gain Qx.y fractional y number of bits. */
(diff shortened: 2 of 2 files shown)


marc-hb:
Github's http://azure.archive.ubuntu.com/ubuntu mirror seems to be flaky in https://github.com/thesofproject/sof/actions/runs/4981366391/jobs/8915525889, seems to affect other PRs too :-(

singalsu:
Note, this is a fix for a failing python test that is currently disabled for SOF CI. @wszypelt

src/include/sof/audio/volume.h
```diff
@@ -66,7 +66,7 @@ struct sof_ipc_ctrl_value_chan;
 #define COMP_VOLUME_Q1_23 1
 
 //** \brief Volume gain Qx.y integer x number of bits including sign bit. */
-#define VOL_QXY_X 1
+#define VOL_QXY_X 8
```
btian1, line 69: Please add each dynamic range before and after your convert, in theory, if input is Q1.31, why output need Q8.23 or Q2.23?

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -499,21 +499,15 @@ static int set_volume_ipc4(struct vol_data *cd, uint32_t const channel,
...
 static inline uint32_t convert_volume_ipc4_to_ipc3(struct comp_dev *dev, uint32_t volume)
 {
-	/* Limit received volume gain to MIN..MAX range before applying it.
-	 * MAX is needed for now for the generic C gain arithmetics to prevent
-	 * multiplication overflow with the 32 bit value. Non-zero MIN option
-	 * can be useful to prevent totally muted small volume gain.
-	 */
-
-	return sat_int24(Q_SHIFT_RND(volume, 31, 23));
+	return sat_int32(Q_SHIFT_RND((int64_t)volume, 31, VOL_QXY_Y));
```
btian1, line 502: this looks strange, if use macro, 31 and 23 both need use macro, if not, we can all use num, if one is macro,, and one is num, looks strange, isn't it?

singalsu: The original code should have used the macro and not hard code the 23. We don't have a macro for IPC4 volume format, so I used just 31.

src/include/sof/audio/volume.h
singalsu: Yep, I'll add to commit text and code comment to header. With Q8.23 the gain can be up to 42.144 dB.  Lowest non-zero gain is -138.47 dB.

src/audio/module_adapter/module/volume/volume.c
lrudyX: The PR has been tested with extended scope. PASSED in 196 from 202 cases and 51 from 51. Looks great. One case which is failing is 44100hz and I think it is not related to peakvol.

lrudyX:
> Note, this is a fix for a failing python test that is currently disabled for SOF CI. @wszypelt

@singalsu The PR has been tested. Almost all passed:
image
Except one case with 44100hz which looks like another issue, not related to peak vol

juimonen (approved):
lgtm, after @btian1 comments addressed that is...

src/include/sof/audio/volume.h
singalsu: done!

andrula-song (approved):

src/audio/module_adapter/module/volume/volume.c
btian1: we can change and unify based on sof code convention instead of keep as of now, I changed a lot requested by Guennadi in my PR.

singalsu: Y refers in VOL_QXY_Y to the fractional  bits count, in that case 23. It would not make sense to "hardcode" it for use of value 23 as you propose. I like compact macro names due to our line lenght limits so Q_FRACTIONAL_BITS is not as compact.

Changing macro names those are in many locations in code would need another patch or even PR.

kv2019i:
Mandatory checks pass, known fail in https://sof-ci.01.org/sofpr/PR7612/build8188/devicetest/index.html .

PR #7618, opened 2023-05-16
copier-ipcgtw:  move out ipcgtw related code from copier to copier_ipcgtw

btian1:


src/audio/copier/copier.c
```diff
@@ -54,10 +54,45 @@ DECLARE_SOF_RT_UUID("copier", copier_comp_uuid, 0x9ba00c83, 0xca12, 0x4a83,
...
+				    in_fmt->s_type);
+	audio_stream_fmt_conversion(out_fmt->depth, out_fmt->valid_bit_depth, &out, &out_valid,
+				    out_fmt->s_type);
+
+	/* check container & sample size */
+	if (use_no_container_convert_function(in, in_valid, out, out_valid))
+		return pcm_get_conversion_function(in, out);
+	else
+		return pcm_get_conversion_vc_function(in, in_valid, out, out_valid, type, dir);
+}
```
ranj063, line 95: why are you moving this part of the code? It makes it harder to review the actual change which is simply to expose those 3 functions

src/audio/copier/copier_ipcgtw.c
```diff
@@ -250,3 +250,81 @@ void ipcgtw_zephyr_reset(struct comp_dev *dev)
 		comp_warn(dev, "ipcgtw_reset(): no buffer found");
 	}
 }
+
+#if CONFIG_IPC4_GATEWAY
```
ranj063, line 254: copier_ipcgtw.c wont be built if the config CONFIG_IPC4_GATEWAY is not selected. So no need to check this again.

btian1: ok， I just move with same and keep origin, let me remove it.

src/include/sof/audio/ipcgtw_copier.h
```diff
@@ -34,16 +34,16 @@ struct ipcgtw_data {
...
 int ipcgtw_zephyr_params(struct ipcgtw_data *ipcgtw_data, struct comp_dev *dev,
 			 struct sof_ipc_stream_params *params);
 
 void ipcgtw_zephyr_reset(struct comp_dev *dev);
 
+int create_ipcgtw(struct comp_dev *parent_dev, struct copier_data *cd,
+		  struct comp_ipc_config *config,
+		  const struct ipc4_copier_module_cfg *copier);
+
+void free_ipcgtw(struct copier_data *cd);
```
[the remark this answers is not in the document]
btian1: ok, can I rename all api in a new patch?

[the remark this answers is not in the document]
ranj063: please rename existing ones first and then as and when you move new ones please add with the correct name

lyakh:
is this replacing #7606 or building on top, temporarily including it?

btian1:
> 

not replacing, just on top of it, need waiting for 7606 merge first, then rebase this one.
following have 3: ipcgtw, host, and dai PRs.

src/include/sof/audio/ipcgtw_copier.h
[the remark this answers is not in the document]
btian1: ipcgtw only have create and free, reset and params also need rename, I renamed with a new patch, I can do rename for create and free in this patch, and later only cover reset and params.

src/audio/copier/copier_ipcgtw.c
```diff
@@ -250,3 +250,79 @@ void ipcgtw_zephyr_reset(struct comp_dev *dev)
...
+	ipcgtw_zephyr_free(ipcgtw_data);
+e_buf:
+	buffer_free(cd->endpoint_buffer[cd->endpoint_num]);
+	return ret;
+}
+
+void copier_ipcgtw_free(struct copier_data *cd)
+{
+	ipcgtw_zephyr_free(cd->ipcgtw_data);
+	buffer_free(cd->endpoint_buffer[0]);
```
ranj063, line 327: can you move the buffer_free inside ipcgtw_zephyr_free() and then rename that to copier_ipcgtw_free()?

src/audio/copier/copier_ipcgtw.c
```diff
@@ -250,3 +250,79 @@ void ipcgtw_zephyr_reset(struct comp_dev *dev)
...
+		goto e_buf;
+	}
+
+	ipcgtw_data = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*ipcgtw_data));
+	if (!ipcgtw_data) {
+		ret = -ENOMEM;
+		goto e_buf;
+	}
+
+	ipcgtw_zephyr_new(ipcgtw_data, gtw_cfg, parent_dev);
```
ranj063, line 287: maybe you can remove this function and move the code right here.

src/audio/copier/copier_ipcgtw.c
```diff
@@ -284,7 +258,19 @@ int copier_ipcgtw_create(struct comp_dev *parent_dev, struct copier_data *cd,
 		goto e_buf;
 	}
 
-	ipcgtw_zephyr_new(ipcgtw_data, gtw_cfg, parent_dev);
+	ipcgtw_data->node_id = gtw_cfg->node_id;
```
ranj063, line 261: @btian1 please squash this with the third patch in the series

src/include/sof/audio/ipcgtw_copier.h
```diff
@@ -34,6 +35,66 @@ struct ipcgtw_data {
...
+	IPC4_IPCGWCMD_FLUSH_DATA = 3
+};
+
+/* Incoming IPC gateway message */
+struct ipc4_ipc_gateway_cmd_data {
+	/* node_id of the target gateway */
+	union ipc4_connector_node_id node_id;
+	/* Payload (actual size is in the header extension.r.data_size) */
+	uint8_t payload[];
+};
```
RanderWang, line 81: if this structure is transmitted from memory window, we need to use "__packed __aligned(4);", also include "struct ipc4_ipc_gateway_cmd_data_reply"

RanderWang (commented):
LGTM

src/include/sof/audio/ipcgtw_copier.h
btian1: these structure directly copied from ipcgtw.h, no other changes:
in = (const struct ipc4_ipc_gateway_cmd_data *)MAILBOX_HOSTBOX_BASE;
out = (struct ipc4_ipc_gateway_cmd_data_reply *)reply_payload;

itself does not allocate memory, it just get memory from driver.

src/audio/copier/CMakeLists.txt
```diff
@@ -1,2 +1,2 @@
-add_local_sources(sof copier.c copier_hifi.c copier_generic.c)
+add_local_sources(sof copier.c copier_hifi.c copier_generic.c copier_ipcgtw.c)
```
ranj063, line 1: copier_ipc_gtw.c should be added with a dependency on CONFIG_IPC4_GATEWAY just like above right? Why are we removing it?

src/include/sof/audio/ipcgtw_copier.h
btian1: thanks, @RanderWang @ranj063 , I added with a new patch, a practice is compiler will take care of alignment when structure memory allocation, but may not all compiler can cover this.

ranj063 (approved):

RanderWang (approved):

src/audio/copier/copier_ipcgtw.c
```diff
@@ -250,3 +223,94 @@ void ipcgtw_zephyr_reset(struct comp_dev *dev)
...
+		comp_buffer_connect(parent_dev, config->core,
+				    cd->endpoint_buffer[cd->endpoint_num],
+				    PPL_CONN_DIR_COMP_TO_BUFFER);
+		cd->bsource_buffer = false;
+	} else {
+		comp_buffer_connect(parent_dev, config->core,
+				    cd->endpoint_buffer[cd->endpoint_num],
+				    PPL_CONN_DIR_BUFFER_TO_COMP);
+		cd->bsource_buffer = true;
+	}
```
lyakh, line 284: this too - move it below converter function

src/audio/copier/copier_ipcgtw.c
```diff
@@ -250,3 +223,94 @@ void ipcgtw_zephyr_reset(struct comp_dev *dev)
...
+	ret = create_endpoint_buffer(parent_dev, cd, config, copier, ipc4_gtw_none, false, 0);
+	if (ret < 0)
+		return ret;
+
+	gtw_cfg = &copier->gtw_cfg;
+	if (!gtw_cfg->config_length) {
+		comp_err(parent_dev, "ipcgtw_create(): empty ipc4_gateway_config_data");
+		ret = -EINVAL;
+		goto e_buf;
+	}
```
lyakh, line 252: can we check this before creating the buffer?

src/include/sof/audio/ipcgtw_copier.h
```diff
@@ -34,6 +35,66 @@ struct ipcgtw_data {
...
+	 * This payload size if 4 bytes smaller (size of the union above).
+	 */
+	uint8_t payload[];
+};
+
+int ipcgtw_process_cmd(const struct ipc4_ipcgtw_cmd *cmd,
+		       void *reply_payload, uint32_t *reply_payload_size);
+
 int copier_ipcgtw_params(struct ipcgtw_data *ipcgtw_data, struct comp_dev *dev,
 			 struct sof_ipc_stream_params *params);
```
lyakh, line 99: why are some functions exported as `copier_ipcgtw_*()` and others just as `ipcgtw_*()`?

serhiy-katsyuba-intel (approved):

src/audio/copier/copier.c
marc-hb: which code convention?

src/include/ipc4/copier.h
```diff
@@ -32,6 +32,7 @@
 #include <sof/compiler_attributes.h>
 #include <sof/audio/buffer.h>
 #include <sof/audio/pcm_converter.h>
+#include <sof/audio/component.h>
```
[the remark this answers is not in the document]
btian1: yes, it can resolve, just curious is it a standard when can use struct declaration and header file usage is just a structure pointer, then we can try to declare it instead of include a header file?

src/audio/copier/copier_ipcgtw.c
btian1: ok, this is origin logic, I did not change it but if ahead of allocate endpoint buffer, we may need change error handing part.

src/audio/copier/copier_ipcgtw.c
```diff
@@ -250,3 +223,94 @@ void ipcgtw_zephyr_reset(struct comp_dev *dev)
...
+	blob = (const struct ipc4_ipc_gateway_config_blob *)
+		((const struct ipc4_gateway_config_data *)gtw_cfg->config_data)->config_blob;
+
+	/* Endpoint buffer is created in copier with size specified in copier config. That buffer
+	 * will be resized to size specified in IPC gateway blob later in ipcgtw_params().
+	 */
+	comp_dbg(parent_dev, "ipcgtw_create(): buffer_size: %u", blob->buffer_size);
+	ipcgtw_data->buf_size = blob->buffer_size;
+
+	list_item_append(&ipcgtw_data->item, &ipcgtw_list_head);
```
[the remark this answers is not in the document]
btian1: main benefit is just make eror handling more simple?

src/audio/copier/copier_ipcgtw.c
btian1: more simple error handling?

src/include/sof/audio/ipcgtw_copier.h
btian1: this part are all aligned, all are started with copier_opcgtw_*

ranj063: ipcgtw_process_cmd() isnt

btian1: shall we change this as well? it is simply ipcgtw process, not related with copier.

src/audio/copier/copier_ipcgtw.c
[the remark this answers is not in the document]
lyakh: yes, it's better to first fully prepare an object before making it globally accessible by e.g. adding it to a global list. Unless there are also other ways in which that object can be found. E.g. if you enable a notifier with this object as its argument and also add it to a list - then if there's a race, if there's no locking, you might need to add to the list first before registering the notifier. But better to lock to avoid such races.

src/audio/copier/copier_ipcgtw.c
lyakh: > you wanna handle all possible errors first, then handle the left things.

exactly

src/audio/copier/copier_ipcgtw.c
lyakh: I'd also move lines 236 and 242 below this check. Why modify any external data if we conclude, that the configuration is invalid and error out?

lyakh (commented):
thanks for addressing comments! Just one more minor reordering request

lyakh (approved):

src/audio/copier/copier.c
btian1: haha, normally, sub-function should be in upper part, and it will be called in lower in the source file.
I understand this is not must, since header file already declared this function prototype, and it can be used earlier before the real code, @marc-hb

marc-hb: > normally, sub-function should be in upper part, and it will be called in lower in the source file.

Even with `static` functions you can use a forward declaration. I'm not aware of any code convention about this. It certainly does not deserve this git noise:

> why are you moving this part of the code? It makes it harder to review the actual change which is simply to expose those 3 functions

PR #7619, opened 2023-05-16
topology2: move bt nocodecs to development

juimonen:
Move bluetooth nocodec topology to development and add target for mtl.


kv2019i (approved):
Thanks @juimonen !

ranj063 (approved):

ranj063 (approved):

PR #7621, opened 2023-05-16
Audio: DRC: Convert component to module adapter API

singalsu:
This patch replaces legacy API with new module adapter API.


src/audio/drc/drc_generic.c
```diff
@@ -526,7 +527,7 @@ static void drc_delay_input_sample_s16(struct drc_state *state,
 	*y = y0;
 }
 
-static void drc_s16_default(const struct comp_dev *dev,
+static void drc_s16_default(struct processing_module *mod,
```
singalsu, line 530: Note: `struct drc_comp_data *cd` would have been sufficient but I used mod to avoid circular dependency with drc_func typedef.

src/audio/drc/drc.c
```diff
@@ -34,7 +35,9 @@
 #include <stddef.h>
 #include <stdint.h>
 
-static const struct comp_driver comp_drc;
+#define DISABLED_CODE 0
```
ranj063, line 38: this is weird? Why not just remove that code is not used?

src/audio/drc/drc.c
```diff
@@ -121,7 +124,7 @@ inline int drc_set_pre_delay_time(struct drc_state *state,
 
 static int drc_setup(struct drc_comp_data *cd, uint16_t channels, uint32_t rate)
 {
-	uint32_t sample_bytes = get_sample_bytes(cd->source_format);
+	uint32_t sample_bytes = get_sample_bytes(cd->sink_format);
```
ranj063, line 127: is this a fix in the current code? Should it be a pre-requisite separate patch then?

src/audio/drc/drc.c
```diff
@@ -140,78 +143,70 @@ static int drc_setup(struct drc_comp_data *cd, uint16_t channels, uint32_t rate)
...
-static struct comp_dev *drc_new(const struct comp_driver *drv,
-				const struct comp_ipc_config *config,
-				const void *spec)
+static int drc_init(struct processing_module *mod)
 {
-	struct comp_dev *dev = NULL;
+	struct module_data *md = &mod->priv;
+	struct comp_dev *dev = mod->dev;
+	struct module_config *cfg = &md->cfg;
 	struct drc_comp_data *cd = NULL;
```
ranj063, line 151: init not necessary

src/audio/drc/drc.c
singalsu: Yep, it will go away, forgot to remove.

src/audio/drc/drc.c
singalsu: In IPC4 components the sink buffer params are set with init() IPC parameters so better to get the stream information from sink instead of source. I could of course make this change a separate IPC4 changes and additions commit.

src/audio/drc/drc.c
singalsu: It was in original. I think it is to ensure a free is done for null pointer.

src/audio/drc/drc.c
```diff
@@ -5,6 +5,7 @@
 // Author: Pin-chih Lin <johnylin@google.com>
 
 #include <sof/audio/component.h>
+#include <sof/audio/module_adapter/module/generic.h>
```
btian1, line 8: any reason to insert a header file instead of add it in the last?

src/audio/drc/drc.c
```diff
@@ -304,189 +301,190 @@ static int drc_trigger(struct comp_dev *dev, int cmd)
...
-	struct comp_copy_limits cl;
-	struct drc_comp_data *cd = comp_get_drvdata(dev);
-	struct comp_buffer *sourceb, *sinkb;
-	struct comp_buffer __sparse_cache *source_c, *sink_c;
-	int ret = 0;
+	const uint32_t byte_align = 1;
+	const uint32_t frame_align_req = 1;
 
-	comp_dbg(dev, "drc_copy()");
+	audio_stream_init_alignment_constants(byte_align, frame_align_req, source);
```
btian1, line 337: possible just pass 1 as parameter and add comments to say the reason for set alignment to 1?

btian1 (commented):
I  have two more questions:
1. I did not work on module adapter yet, question is: where create mod and assign accordingly?
2. possible to split with multiple patches? I understand this is one functionality, but still looks too long for review.

src/audio/drc/drc.c
singalsu: fixed now

src/audio/drc/drc.c
```diff
@@ -370,7 +370,7 @@ static int drc_reset(struct processing_module *mod)
 
 	drc_reset_state(&cd->state);
 
-	cd->source_format = 0;
+	cd->sink_format = 0;
```
lyakh, line 373: this is `SOF_IPC_FRAME_S16_LE` - is this what we want?

singalsu: Component private data is rzalloc'ed so this way the component is reset to init state. I think use of zero here instead of enum is correct. Do you agree?

src/audio/drc/drc.c
```diff
@@ -140,353 +141,206 @@ static int drc_setup(struct drc_comp_data *cd, uint16_t channels, uint32_t rate)
...
-
 	cd = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*cd));
 	if (!cd)
-		goto fail;
-
-	comp_set_drvdata(dev, cd);
+		return -ENOMEM;
 
+	md->private = cd;
 	cd->drc_func = NULL;
```
cujomalainey, line 169: you can also delete this line since `rzalloc` makes this null anyways

src/audio/drc/drc.c
```diff
@@ -140,353 +141,206 @@ static int drc_setup(struct drc_comp_data *cd, uint16_t channels, uint32_t rate)
...
+	mod->simple_copy = true;
+	return 0;
 
 cd_fail:
 	comp_data_blob_handler_free(cd->model_handler);
 	rfree(cd);
-fail:
-	rfree(dev);
-	return NULL;
+	return ret;
```
cujomalainey, line 193: do we have any concerns about the dangling pointer left in `mod` here?

src/audio/drc/drc.c
```diff
@@ -140,353 +141,206 @@ static int drc_setup(struct drc_comp_data *cd, uint16_t channels, uint32_t rate)
...
-	source_c = buffer_acquire(sourceb);
-	sink_c = buffer_acquire(sinkb);
+	comp_dbg(dev, "drc_process()");
 
 	/* Check for changed configuration */
 	if (comp_is_new_data_blob_available(cd->model_handler)) {
 		cd->config = comp_get_data_blob(cd->model_handler, NULL, NULL);
-		ret = drc_setup(cd, audio_stream_get_channels(&source_c->stream),
-				audio_stream_get_rate(&source_c->stream));
+		ret = drc_setup(cd, audio_stream_get_channels(sink), audio_stream_get_rate(sink));
```
cujomalainey, line 258: you flip from source to sink here, why?

src/audio/drc/drc.c
```diff
@@ -269,6 +269,43 @@ static int drc_process(struct processing_module *mod,
...
+	comp_dbg(dev, "drc_params()");
+
+	comp_params = *params;
+	comp_params.channels = mod->priv.cfg.base_cfg.audio_fmt.channels_count;
+	comp_params.rate = mod->priv.cfg.base_cfg.audio_fmt.sampling_frequency;
+	comp_params.buffer_fmt = mod->priv.cfg.base_cfg.audio_fmt.interleaving_style;
+
+	audio_stream_fmt_conversion(mod->priv.cfg.base_cfg.audio_fmt.depth,
+				    mod->priv.cfg.base_cfg.audio_fmt.valid_bit_depth,
+				    &frame_fmt, &valid_fmt,
```
cujomalainey, line 292: don't know what this function does, but seems weird we are throwing away `valid_fmt`

src/audio/drc/drc.c
cujomalainey: I would just delete the line or use the enum, leaving as it is not helpful

cujomalainey (changes requested):

src/audio/drc/drc.c
lyakh: > Component private data is rzalloc'ed so this way the component is reset to init state. I think use of zero here instead of enum is correct. Do you agree?

@singalsu no, actually I don't. Then I'd say the initialisation is also wrong - if that field is explicitly initialised to 0 there. It is ok to leave objects uninitialised *until* they're first used - then of course they must be initialised. I think you have two possibilities:
1. you initialise an object when it's first instantiated, then whenever you get to use it you know, that it is in some consistent state
2. you initialise data any time *before* the first use. In this case it's find to leave that data in some random undefined state until then, but of course then you must know exactly *when* your data is needed for the first time.

Both ways are valid. The second one can be more optimal in terms of performance - you don't "waste" cycles on initialisation, if it's anyway going to be overwritten later, but it might also be more complex - you risk missing the point when the data is used and you end up consuming random data.

Either way when you do set your data to something, it should be something *meaningful*, and not 0 just because it's some magic universal initialiser :-)

src/audio/drc/drc.c
johnylin76: I think the flip should be contained by the following commit https://github.com/thesofproject/sof/pull/7621/commits/90997e3af60d4b29a8dfe224c74ccae8636610a8 instead

src/audio/drc/drc.c
singalsu: Yep, in function module_adapter_new() https://github.com/thesofproject/sof/blob/ead25e82a2ca5d04dccc13b3930f0e9ee95e4c35/src/audio/module_adapter/module_adapter.c#L135 mod and dev are freed if there was error.

src/audio/drc/drc.c
singalsu: Yep, fixed.

src/audio/drc/drc.c
singalsu: Fixed now.

src/audio/drc/drc.c
singalsu: Fixed now, deleted the line.

src/audio/drc/drc.c
```diff
@@ -140,353 +141,206 @@ static int drc_setup(struct drc_comp_data *cd, uint16_t channels, uint32_t rate)
...
-		goto fail;
-
-	comp_set_drvdata(dev, cd);
+		return -ENOMEM;
 
+	md->private = cd;
 	cd->drc_func = NULL;
 
 	/* Handler for configuration data */
 	cd->model_handler = comp_data_blob_handler_new(dev);
```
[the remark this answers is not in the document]
johnylin76: IMHO (as a module developer) I would prefer that the adapter layer can handle the operation for config blobs, e.g. blob (de-)fragmentation, while the module API s(g)et_configuration simply provides the whole config at once. I am not sure about other module types like IADK but I think the common-typed modules should be independent of data blob handlng.

singalsu:
@cujomalainey @johnylin76 Does this look OK now?

singalsu:
@kv2019i @btian1 Proposed merge order for DRC patches

1. https://github.com/thesofproject/sof/pull/7621
2. https://github.com/thesofproject/rimage/pull/161
3. https://github.com/thesofproject/sof/pull/7627

lyakh (approved):

src/audio/drc/drc.c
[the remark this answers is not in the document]
cujomalainey: We can do this in a follow up

src/audio/drc/drc.c
cujomalainey: please test before submitting

cujomalainey (commented):
my only concern left is frame format comment, otherwise it looks good, let me know once you validate that then I am ok to sign off

singalsu:
Updated PR to build and work with changes from #7622 .

src/audio/drc/drc.c
cujomalainey: @singalsu any word on this?

singalsu: Sorry, not yet. We had heavy focus in code optimization in last weeks. I Hope I can do this next week.

singalsu: I made this topology as sof-tgl-nocodec.tplg replacement for UPX device. Drc.0.1 is s16 input/output, similarly drc.1.1 is s24, drc2.1 is s32. The playback PCMs accept all s16/s24/s32 formats. The I2S DAIs are s32. The capture PCMs get SSP loopback audio.

sof-tgl-nocodec-drc-formats

I'll make tomorrow script to test all formats combinations.

singalsu: @cujomalainey Now confirmed, the next line `comp_params.frame_fmt = frame_fmt;` needs to be `comp_params.frame_fmt = valid_fmt;`. Good catch! And need to fix this from other components too. It causes s24_le format to be treated as s32_le.

cujomalainey: Glad it was caught, thanks for testing :)

src/audio/drc/drc.c
```diff
@@ -269,6 +269,43 @@ static int drc_process(struct processing_module *mod,
...
+
+	comp_params.frame_fmt = frame_fmt;
+
+	for (i = 0; i < SOF_IPC_MAX_CHANNELS; i++)
+		comp_params.chmap[i] = (mod->priv.cfg.base_cfg.audio_fmt.ch_map >> i * 4) & 0xf;
+
+	component_set_nearest_period_frames(dev, comp_params.rate);
+	sinkb = list_first_item(&dev->bsink_list, struct comp_buffer, source_list);
+	sink_c = buffer_acquire(sinkb);
+	ret = buffer_set_params(sink_c, &comp_params, true);
```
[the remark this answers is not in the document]
cujomalainey: @johnylin76 sorry im not certain what the issue is here. Can you clarify?

src/audio/drc/drc.c
```diff
@@ -140,353 +141,206 @@ static int drc_setup(struct drc_comp_data *cd, uint16_t channels, uint32_t rate)
...
-		break;
-	default:
-		comp_err(dev, "drc_cmd_set_data() error: invalid cdata->cmd");
-		ret = -EINVAL;
-		break;
-	}
+	comp_info(mod->dev, "drc_set_config()");
 
-	return ret;
+	return comp_data_blob_set(cd->model_handler, pos, data_offset_size, fragment,
```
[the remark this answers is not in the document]
cujomalainey: ping @ranj063

src/audio/drc/drc.c
```diff
@@ -140,353 +141,248 @@ static int drc_setup(struct drc_comp_data *cd, uint16_t channels, uint32_t rate)
...
-
-	/* Get source, sink, number of frames etc. to process. */
-	comp_get_copy_limits(source_c, sink_c, &cl);
+	comp_params.frame_fmt = valid_fmt;
 
-	/* Run DRC function */
-	drc_process(dev, source_c, sink_c, cl.frames, cl.source_bytes,
-		    cl.sink_bytes);
+	for (i = 0; i < SOF_IPC_MAX_CHANNELS; i++)
+		comp_params.chmap[i] = (mod->priv.cfg.base_cfg.audio_fmt.ch_map >> i * 4) & 0xf;
```
cujomalainey, line 297: Can we get a helper for this? e.g. `get_chan_from_map(map, chan)`?

cujomalainey:
only one more comment from me, thanks!

src/audio/drc/drc.c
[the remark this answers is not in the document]
ranj063: @johnylin76 @singalsu comp_data_blob_set() was introduced before IPC4 which supports module_set_configuration. set_configuration is part of the module interface ops and will not be replaced.

lgirdwood (approved):

gkbldcig:

src/audio/drc/drc.c
singalsu: I think I should convert this to use ipc4_update_buffer_format() from ipc/ipc4/helper.c, I've missed that such exists. It updates the channels map and everything else here.

singalsu: I've now updated it similarly as volume.c. The changed version updates both source and sink buffer parameters. I'm not sure about it so I'll check if that's the policy generally for simple 1 source 1 sink components. The patch to use sink parameters for DRC setup instead of source would be then unnecessary.

singalsu:
Changed drc.h function prototype drc_default_pass() that was missing the __sparse_cache for source and sink. CI check with sof/scripts/parse_sparse_output.sh found the issue.

src/audio/drc/drc.c
```diff
@@ -139,353 +141,234 @@ static int drc_setup(struct drc_comp_data *cd, uint16_t channels, uint32_t rate)
...
+	sinkb = list_first_item(&dev->bsink_list, struct comp_buffer, source_list);
+	sink_c = buffer_acquire(sinkb);
+	ipc4_update_buffer_format(sink_c, &mod->priv.cfg.base_cfg.audio_fmt);
 	buffer_release(sink_c);
-	buffer_release(source_c);
 
-	return ret;
+	sourceb = list_first_item(&dev->bsource_list, struct comp_buffer, sink_list);
+	source_c = buffer_acquire(sourceb);
+	ipc4_update_buffer_format(source_c, &mod->priv.cfg.base_cfg.audio_fmt);
```
cujomalainey, line 291: oh gosh, grepping for `base_cfg.audio_fmt`  is all over the place. Mind adding a macro for `mod` to `audio_fmt`? Can be done in another PR.

src/include/sof/audio/drc/drc.h
```diff
@@ -70,7 +70,7 @@ struct drc_comp_data {
 	struct comp_data_blob_handler *model_handler;
 	struct sof_drc_config *config;      /**< pointer to setup blob */
 	bool config_ready;                  /**< set when fully received */
-	enum sof_ipc_frame source_format;   /**< source frame format */
+	enum sof_ipc_frame sink_format;     /**< source frame format */
```
cujomalainey, line 73: comment update

also how do we know what we will get if the format is not set for source?

cujomalainey (approved):
two minor issues

src/include/sof/audio/drc/drc.h
singalsu: I got confirm from @ranj063 that update both sink and source should be done as it is now. I can then remove the source -> sink commit that has this comment mistake.

src/audio/drc/drc.c
lgirdwood: ack, I think there will be a lot of low hanging fruit for convenience macros to take care of cleanup and readability after all the modules are converted.

lgirdwood (approved):

Issue #7261, opened 2023-03-13
[FEATURE] Pipeline 2.0 with native IPC4, Module API and required windows features 


marcinszkudlinski:
**Is your feature request related to a problem? Please describe.**
 
We're heading fast for firmware fully compatible with current windows driver

I order to achieve this: 
 - IPC4 was introduced, IPC3 was set as depreciated
 - SOF component API was set as depreciated, and "Module API" was introduced 
 - New way of scheduling (Data processing) was introduced (PR still open)
 
To fit the requirements into current SOF code:
 - the pipeline itself is still controlled in IPC3 way, using a wrapper/translator to be compatible with IPC4
 - the pipeline is still using component API only, a heavy and slow module adapter must be used
 
Till now such approach was fine but is getting out of hands. Is consuming a lot of memory, is slow and with high risk of regression in legacy platforms
 
**Describe the solution you'd like**
 
I think there's a right moment to introduce **PIPELINE 2.0**
 - using IPC4 natively
 - using module API as main API 
 - without any buffering between LL modules
 - with no cache invalidation (both for audio data and metadata) unless necessary
- with producer-consumer safe manageable DP Queues instead of simple buffers between chunks of LL 
- in case when modules are on separate cores - cache coherent shared DP-queue
 - and more - the list is not 100% complete
 
 
**Describe alternatives you've considered**
Keep modifying current pipeline implementation. 
I my opinion 
-	technical debt is now too high 
-	high risk of introducing regression in legacy platforms
 
**Additional context**

plbossart:
@marcinszkudlinski The list of suggested improvements seem appealing, we should always try to do better, but allow me to put on my 'bad cop' hat to help clarify the directions.

a) What exactly does "Till now such approach was fine but is getting out of hands. Is consuming a lot of memory, is slow"
You will have to bring a lot more data to make a case for a change. Most of the memory issues we'd had were because of very large 3rd party modules, it's not clear to me what the pipeline infrastructure costs. The slowness is not clear either.

b) The wording also suggests that it's not possible to support DP scheduling without changing the pipeline level, which is not something I've heard before.

c) "The right moment" looks to me as the "wrong time and wrong place". We are half-way through the IPC4 introduction, there's no way any engineering manager in their right mind would accept to change the plumbing and have to redo all the validation, unless there's a clear showstopper preventing us from reaching the committed goals.

[...]

PR #7622, opened 2023-05-16
sink / source API

marcinszkudlinski:
This introduces sink and source API 

in summary: audio_stream is bind to data buffer - it contains write/read pointers etc. Has no encapsulation, so anybody can access buffer directly

Sink/source **provide an API** to get to the data. It does not assume the data are in any buffer etc. Usage of API is enforced. API may do anything with the data - move in memory for linearization, take care of coherency, etc. 

THE POINT IS: any module that use the API may get data from many sources and send to many sources, no matter if it is a circular buffer (audio_stream), or a coherent DP queue, shared LL buffer, DAI, data linearization layer - the transition from one source/sink to another is transparent and may be done in runtime

To make transition smooth, an implementation of sink/source has been added to audio_stream. 
There's also a POC - src module, with some TODOs

Sink/source API is a part of pipeline 2.0 will be required for
 - DP processing on secondary cores
 - Memory optimalization in LL modules chains

### DIFF
--- a/src/audio/CMakeLists.txt
+++ b/src/audio/CMakeLists.txt
@@ -13,6 +13,7 @@ if((NOT CONFIG_LIBRARY) OR CONFIG_LIBRARY_STATIC)
 		buffer.c
 		source_api_helper.c
 		sink_api_helper.c
+		sink_source_utils.c
 		channel_map.c
 	)
 	if(CONFIG_COMP_BLOB)
--- a/src/audio/CMakeLists.txt
+++ b/src/audio/CMakeLists.txt
@@ -12,6 +12,7 @@ if((NOT CONFIG_LIBRARY) OR CONFIG_LIBRARY_STATIC)
 		component.c
 		buffer.c
 		source_api_helper.c
+		sink_api_helper.c
 		channel_map.c
 	)
 	if(CONFIG_COMP_BLOB)
--- a/src/audio/CMakeLists.txt
+++ b/src/audio/CMakeLists.txt
@@ -11,6 +11,7 @@ if((NOT CONFIG_LIBRARY) OR CONFIG_LIBRARY_STATIC)
 		host-legacy.c
 		component.c
 		buffer.c
+		source_api_helper.c
 		channel_map.c
 	)
 	if(CONFIG_COMP_BLOB)
--- a/src/audio/CMakeLists.txt
+++ b/src/audio/CMakeLists.txt
@@ -14,6 +14,7 @@ if((NOT CONFIG_LIBRARY) OR CONFIG_LIBRARY_STATIC)
 		source_api_helper.c
 		sink_api_helper.c
 		sink_source_utils.c
+		audio_stream.c
 		channel_map.c
 	)
 	if(CONFIG_COMP_BLOB)
--- a/src/audio/asrc/asrc.c
+++ b/src/audio/asrc/asrc.c
@@ -555,7 +555,7 @@ static int asrc_params(struct comp_dev *dev,
 
 	/* Don't change sink rate if value from IPC is 0 (auto detect) */
 	if (asrc_get_sink_rate(&cd->ipc_config))
-		sink_c->stream.rate = asrc_get_sink_rate(&cd->ipc_config);
+		audio_stream_set_rate(&sink_c->stream, asrc_get_sink_rate(&cd->ipc_config));
 
 	/* set source/sink_frames/rate */
 	cd->source_rate = audio_stream_get_rate(&source_c->stream);
--- a/src/audio/audio_stream.c
+++ b/src/audio/audio_stream.c
@@ -1,0 +1,175 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// Copyright(c) 2023 Intel Corporation. All rights reserved.
+//
+
+#include <sof/audio/audio_stream.h>
+#include <sof/audio/buffer.h>
+
+static size_t audio_stream_get_free_size(struct sof_sink __sparse_cache *sink)
+{
+	struct audio_stream __sparse_cache *audio_stream =
+			attr_container_of(sink, struct audio_stream __sparse_cache,
+					  sink_api, __sparse_cache);
+
+	return audio_stream_get_free_bytes(audio_stream);
+}
+
+static int audio_stream_get_buffer(struct sof_sink __sparse_cache *sink, size_t req_size,
+				   void **data_ptr, void **buffer_start, size_t *buffer_size)
+{
+	struct audio_stream __sparse_cache *audio_stream =
+			attr_container_of(sink, struct audio_stream __sparse_cache,
+					  sink_api, __sparse_cache);
+
+	if (req_size > audio_stream_get_free_size(sink))
+		return -ENODATA;
+
+	/* get circular buffer parameters */
+	*data_ptr = audio_stream->w_ptr;
+	*buffer_start = audio_stream->addr;
+	*buffer_size = audio_stream->size;
+	return 0;
+}
+
+static int audio_stream_commit_buffer(struct sof_sink __sparse_cache *sink, size_t commit_size)
+{
+	struct audio_stream __sparse_cache *audio_stream =
+			attr_container_of(sink, struct audio_stream __sparse_cache,
+					  sink_api, __sparse_cache);
+
+	if (commit_size)
+		audio_stream_produce(audio_stream, commit_size);
+
+	return 0;
+}
+
+static size_t audio_stream_get_data_available(struct sof_source __sparse_cache *source)
+{
+	struct audio_stream __sparse_cache *audio_stream =
+			attr_container_of(source, struct audio_stream __sparse_cache,
+					  source_api, __sparse_cache);
+
+	return audio_stream_get_avail_bytes(audio_stream);
+}
+
+static int audio_stream_get_data(struct sof_source __sparse_cache *source, size_t req_size,
+				 void  **data_ptr, void **buffer_start, size_t *buffer_size)
+{
+	struct audio_stream __sparse_cache *audio_stream =
+			attr_container_of(source, struct audio_stream __sparse_cache,
+					  source_api, __sparse_cache);
+
+	if (req_size > audio_stream_get_data_available(source))
+		return -ENODATA;
+
+	/* get circular buffer parameters */
+	*data_ptr = audio_stream->r_ptr;
+	*buffer_start = audio_stream->addr;
+	*buffer_size = audio_stream->size;
+	return 0;
+}
+
+static int audio_stream_release_data(struct sof_source __sparse_cache *source, size_t free_size)
+{
+	struct audio_stream __sparse_cache *audio_stream =
+			attr_container_of(source, struct audio_stream __sparse_cache,
+					  source_api, __sparse_cache);
+
+	if (free_size)
+		audio_stream_consume(audio_stream, free_size);
+
+	return 0;
+}
+
+static int audio_stream_set_ipc_params_source(struct sof_source __sparse_cache *source,
+					      struct sof_ipc_stream_params *params,
+					      bool force_update)
+{
+	struct audio_stream __sparse_cache *audio_stream =
+			attr_container_of(source, struct audio_stream __sparse_cache,
+					  source_api, __sparse_cache);
+	struct comp_buffer __sparse_cache *buffer =
+			attr_container_of(audio_stream, struct comp_buffer __sparse_cache,
+					  stream, __sparse_cache);
+
+	return buffer_set_params(buffer, params, force_update);
+}
+
+static int audio_stream_set_ipc_params_sink(struct sof_sink __sparse_cache *sink,
+					    struct sof_ipc_stream_params *params,
+					    bool force_update)
+{
+	struct audio_stream __sparse_cache *audio_stream =
+			attr_container_of(sink, struct audio_stream __sparse_cache,
+					  sink_api, __sparse_cache);
+	struct comp_buffer __sparse_cache *buffer =
+			attr_container_of(audio_stream, struct comp_buffer __sparse_cache,
+					  stream, __sparse_cache);
+
+	return buffer_set_params(buffer, params, force_update);
+}
+
+static int audio_stream_source_set_alignment_constants(struct sof_source __sparse_cache *source,
+						       const uint32_t byte_align,
+						       const uint32_t frame_align_req)
+{
+	struct audio_stream __sparse_cache *audio_stream =
+			attr_container_of(source, struct audio_stream __sparse_cache,
+					  source_api, __sparse_cache);
+
+	audio_stream_init_alignment_constants(byte_align, frame_align_req, audio_stream);
+
+	return 0;
+}
+
+static int audio_stream_sink_set_alignment_constants(struct sof_sink __sparse_cache *sink,
+						     const uint32_t byte_align,
+						     const uint32_t frame_align_req)
+{
+	struct audio_stream __sparse_cache *audio_stream =
+			attr_container_of(sink, struct audio_stream __sparse_cache,
+					  sink_api, __sparse_cache);
+
+	audio_stream_init_alignment_constants(byte_align, frame_align_req, audio_stream);
+
+	return 0;
+}
+
+static const struct source_ops audio_stream_source_ops = {
+	.get_data_available = audio_stream_get_data_available,
+	.get_data = audio_stream_get_data,
+	.release_data = audio_stream_release_data,
+	.audio_set_ipc_params = audio_stream_set_ipc_params_source,
+	.set_alignment_constants = audio_stream_source_set_alignment_constants
+};
+
+static const struct sink_ops audio_stream_sink_ops = {
+	.get_free_size = audio_stream_get_free_size,
+	.get_buffer = audio_stream_get_buffer,
+	.commit_buffer = audio_stream_commit_buffer,
+	.audio_set_ipc_params = audio_stream_set_ipc_params_sink,
+	.set_alignment_constants = audio_stream_sink_set_alignment_constants
+};
+
+void audio_stream_init(struct audio_stream __sparse_cache *audio_stream,
+		       void *buff_addr, uint32_t size)
+{
+	audio_stream->size = size;
+	audio_stream->addr = buff_addr;
+	audio_stream->end_addr = (char *)audio_stream->addr + size;
+
+	/* set the default alignment info.
+	 * set byte_align as 1 means no alignment limit on byte.
+	 * set frame_align as 1 means no alignment limit on frame.
+	 */
+	audio_stream_init_alignment_constants(1, 1, audio_stream);
+
+	source_init(audio_stream_get_source(audio_stream), &audio_stream_source_ops,
+		    (__sparse_force struct sof_audio_stream_params *)
+		    &audio_stream->runtime_stream_params);
+	sink_init(audio_stream_get_sink(audio_stream), &audio_stream_sink_ops,
+		  (__sparse_force struct sof_audio_stream_params *)
+		  &audio_stream->runtime_stream_params);
+	audio_stream_reset(audio_stream);
+}
--- a/src/audio/copier/copier_generic.c
+++ b/src/audio/copier/copier_generic.c
@@ -71,8 +71,7 @@ void update_buffer_format(struct comp_buffer __sparse_cache *buf_c,
 
 	audio_stream_set_frm_fmt(&buf_c->stream, frame_fmt);
 	audio_stream_set_valid_fmt(&buf_c->stream, valid_fmt);
-
-	buf_c->buffer_fmt = fmt->interleaving_style;
+	audio_stream_set_buffer_fmt(&buf_c->stream, fmt->interleaving_style);
 
 	for (i = 0; i < SOF_IPC_MAX_CHANNELS; i++)
 		buf_c->chmap[i] = (fmt->ch_map >> i * 4) & 0xf;
@@ -238,7 +237,8 @@ int create_endpoint_buffer(struct comp_dev *parent_dev,
 	audio_stream_set_rate(&buffer_c->stream, copier_cfg->base.audio_fmt.sampling_frequency);
 	audio_stream_set_frm_fmt(&buffer_c->stream, config->frame_fmt);
 	audio_stream_set_valid_fmt(&buffer_c->stream, valid_fmt);
-	buffer_c->buffer_fmt = copier_cfg->base.audio_fmt.interleaving_style;
+	audio_stream_set_buffer_fmt(&buffer_c->stream,
+				    copier_cfg->base.audio_fmt.interleaving_style);
 
 	for (i = 0; i < SOF_IPC_MAX_CHANNELS; i++)
 		buffer_c->chmap[i] = (chan_map >> i * 4) & 0xf;
--- a/src/audio/copier/copier_generic.c
+++ b/src/audio/copier/copier_generic.c
@@ -62,15 +62,15 @@ void update_buffer_format(struct comp_buffer __sparse_cache *buf_c,
 	enum sof_ipc_frame valid_fmt, frame_fmt;
 	int i;
 
-	buf_c->stream.channels = fmt->channels_count;
-	buf_c->stream.rate = fmt->sampling_frequency;
+	audio_stream_set_channels(&buf_c->stream, fmt->channels_count);
+	audio_stream_set_rate(&buf_c->stream, fmt->sampling_frequency);
 	audio_stream_fmt_conversion(fmt->depth,
 				    fmt->valid_bit_depth,
 				    &frame_fmt, &valid_fmt,
 				    fmt->s_type);
 
-	buf_c->stream.frame_fmt = frame_fmt;
-	buf_c->stream.valid_sample_fmt = valid_fmt;
+	audio_stream_set_frm_fmt(&buf_c->stream, frame_fmt);
+	audio_stream_set_valid_fmt(&buf_c->stream, valid_fmt);
 
 	buf_c->buffer_fmt = fmt->interleaving_style;
 
--- a/src/audio/module_adapter/module/generic.c
+++ b/src/audio/module_adapter/module/generic.c
@@ -94,8 +94,11 @@ int module_init(struct processing_module *mod, struct module_interface *interfac
 		return -EIO;
 	}
 
-	if (!interface->init || !interface->prepare || !interface->process ||
-	    !interface->reset || !interface->free) {
+	/*check interface, there must be one and only one of processing procedure */
+	if (!interface->init || !interface->prepare ||
+	    !interface->reset || !interface->free ||
+	    (!!interface->process + !!interface->process_audio_stream +
+	     !!interface->process_raw_data != 1)) {
 		comp_err(dev, "module_init(): comp %d is missing mandatory interfaces",
 			 dev_comp_id(dev));
 		return -EIO;
@@ -225,16 +228,17 @@ int module_prepare(struct processing_module *mod)
 	return ret;
 }
 
-int module_process(struct processing_module *mod, struct input_stream_buffer *input_buffers,
-		   int num_input_buffers, struct output_stream_buffer *output_buffers,
-		   int num_output_buffers)
+int module_process_legacy(struct processing_module *mod,
+			  struct input_stream_buffer *input_buffers, int num_input_buffers,
+			  struct output_stream_buffer *output_buffers,
+			  int num_output_buffers)
 {
 	struct comp_dev *dev = mod->dev;
 	int ret;
 
 	struct module_data *md = &mod->priv;
 
-	comp_dbg(dev, "module_process() start");
+	comp_dbg(dev, "module_process_legacy() start");
 
 	if (md->state != MODULE_IDLE) {
 		comp_err(dev, "module_process(): wrong state of comp_id %x, state %d",
@@ -245,15 +249,59 @@ int module_process(struct processing_module *mod, struct input_stream_buffer *in
 	/* set state to processing */
 	md->state = MODULE_PROCESSING;
 
-	ret = md->ops->process(mod, input_buffers, num_input_buffers, output_buffers,
-			       num_output_buffers);
+	if (md->ops->process_audio_stream)
+		ret = md->ops->process_audio_stream(mod, input_buffers, num_input_buffers,
+						    output_buffers, num_output_buffers);
+	else if (md->ops->process_raw_data)
+		ret = md->ops->process_raw_data(mod, input_buffers, num_input_buffers,
+						output_buffers, num_output_buffers);
+	else
+		ret = -EOPNOTSUPP;
+
 	if (ret && ret != -ENOSPC && ret != -ENODATA) {
 		comp_err(dev, "module_process() error %d: for comp %d",
 			 ret, dev_comp_id(dev));
 		return ret;
 	}
 
-	comp_dbg(dev, "module_process() done");
+	comp_dbg(dev, "module_process_legacy() done");
+
+	/* reset state to idle */
+	md->state = MODULE_IDLE;
+	return 0;
+}
+
+int module_process_sink_src(struct processing_module *mod,
+			    struct sof_source __sparse_cache **sources, int num_of_sources,
+			    struct sof_sink __sparse_cache **sinks, int num_of_sinks)
+
+{
+	struct comp_dev *dev = mod->dev;
+	int ret;
+
+	struct module_data *md = &mod->priv;
+
+	comp_dbg(dev, "module_process sink src() start");
+
+	if (md->state != MODULE_IDLE) {
+		comp_err(dev, "module_process(): wrong state of comp_id %x, state %d",
+			 dev_comp_id(dev), md->state);
+		return -EPERM;
+	}
+
+	/* set state to processing */
+	md->state = MODULE_PROCESSING;
+
+	assert(md->ops->process);
+	ret = md->ops->process(mod, sources, num_of_sources, sinks, num_of_sinks);
+
+	if (ret && ret != -ENOSPC && ret != -ENODATA) {
+		comp_err(dev, "module_process() error %d: for comp %d",
+			 ret, dev_comp_id(dev));
+		return ret;
+	}
+
+	comp_dbg(dev, "module_process sink src() done");
 
 	/* reset state to idle */
 	md->state = MODULE_IDLE;
--- a/src/audio/module_adapter/module/generic.c
+++ b/src/audio/module_adapter/module/generic.c
@@ -192,7 +192,9 @@ static int validate_config(struct module_config *cfg)
 	return 0;
 }
 
-int module_prepare(struct processing_module *mod)
+int module_prepare(struct processing_module *mod,
+		   struct sof_source __sparse_cache **sources, int num_of_sources,
+		   struct sof_sink __sparse_cache **sinks, int num_of_sinks)
 {
 	int ret;
 	struct module_data *md = &mod->priv;
@@ -205,7 +207,7 @@ int module_prepare(struct processing_module *mod)
 	if (mod->priv.state < MODULE_INITIALIZED)
 		return -EPERM;
 
-	ret = md->ops->prepare(mod);
+	ret = md->ops->prepare(mod, sources, num_of_sources, sinks, num_of_sinks);
 	if (ret) {
 		comp_err(dev, "module_prepare() error %d: module specific prepare failed, comp_id %d",
 			 ret, dev_comp_id(dev));
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -165,6 +165,55 @@ err:
 	return NULL;
 }
 
+static int module_adapter_sink_src_prepare(struct comp_dev *dev)
+{
+	struct comp_buffer __sparse_cache *source_buffers_c[PLATFORM_MAX_STREAMS];
+	struct comp_buffer __sparse_cache *sinks_buffers_c[PLATFORM_MAX_STREAMS];
+	struct sof_sink __sparse_cache *audio_sink[PLATFORM_MAX_STREAMS];
+	struct sof_source __sparse_cache *audio_src[PLATFORM_MAX_STREAMS];
+	struct processing_module *mod = comp_get_drvdata(dev);
+	struct list_item *blist;
+	uint32_t num_of_sources = 0;
+	uint32_t num_of_sinks = 0;
+	int ret;
+
+	/* acquire all sink and source buffers, get handlers to sink/source API */
+	list_for_item(blist, &dev->bsink_list) {
+		struct comp_buffer *sink_buffer_uc;
+
+		sink_buffer_uc = container_of(blist, struct comp_buffer, source_list);
+		sinks_buffers_c[num_of_sinks] = buffer_acquire(sink_buffer_uc);
+		audio_sink[num_of_sinks] =
+				audio_stream_get_sink(&sinks_buffers_c[num_of_sinks]->stream);
+		sink_reset_num_of_processed_bytes(audio_sink[num_of_sinks]);
+		num_of_sinks++;
+	}
+
+	list_for_item(blist, &dev->bsource_list) {
+		struct comp_buffer *source_buffer_uc;
+
+		source_buffer_uc = container_of(blist, struct comp_buffer, sink_list);
+		source_buffers_c[num_of_sources] = buffer_acquire(source_buffer_uc);
+		audio_src[num_of_sources] =
+				audio_stream_get_source(&source_buffers_c[num_of_sources]->stream);
+		source_reset_num_of_processed_bytes(audio_src[num_of_sources]);
+		num_of_sources++;
+	}
+
+	/* Prepare module */
+	ret = module_prepare(mod, audio_src, num_of_sources, audio_sink, num_of_sinks);
+
+	/* release all source buffers in reverse order */
+	for (int i = num_of_sources - 1; i >= 0; i--)
+		buffer_release(source_buffers_c[i]);
+
+	/* release all sink buffers in reverse order */
+	for  (int i = num_of_sinks - 1; i >= 0 ; i--)
+		buffer_release(sinks_buffers_c[i]);
+
+	return ret;
+}
+
 /*
  * \brief Prepare the module
  * \param[in] dev - component device pointer.
@@ -189,7 +238,11 @@ int module_adapter_prepare(struct comp_dev *dev)
 	comp_dbg(dev, "module_adapter_prepare() start");
 
 	/* Prepare module */
-	ret = module_prepare(mod);
+	if (IS_PROCESSING_MODE_SINK_SOURCE(mod))
+		ret = module_adapter_sink_src_prepare(dev);
+	else
+		ret = module_prepare(mod, NULL, 0, NULL, 0);
+
 	if (ret) {
 		if (ret != PPL_STATUS_PATH_STOP)
 			comp_err(dev, "module_adapter_prepare() error %x: module prepare failed",
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -24,6 +24,19 @@
 LOG_MODULE_REGISTER(module_adapter, CONFIG_SOF_LOG_LEVEL);
 
 /*
+ * helpers to determine processing type
+ * Needed till all the modules use PROCESSING_MODE_SINK_SOURCE
+ */
+#define IS_PROCESSING_MODE_AUDIO_STREAM(mod) \
+		(!!((struct module_data *)&(mod)->priv)->ops->process_audio_stream)
+
+#define IS_PROCESSING_MODE_RAW_DATA(mod) \
+		(!!((struct module_data *)&(mod)->priv)->ops->process_raw_data)
+
+#define IS_PROCESSING_MODE_SINK_SOURCE(mod) \
+		(!!((struct module_data *)&(mod)->priv)->ops->process)
+
+/*
  * \brief Create a module adapter component.
  * \param[in] drv - component driver pointer.
  * \param[in] config - component ipc descriptor pointer.
@@ -240,7 +253,9 @@ int module_adapter_prepare(struct comp_dev *dev)
 		return -EINVAL;
 	}
 
-	if (mod->simple_copy && mod->num_input_buffers > 1 && mod->num_output_buffers > 1) {
+	/* check processing mode */
+	if (IS_PROCESSING_MODE_AUDIO_STREAM(mod) &&
+	    mod->num_input_buffers > 1 && mod->num_output_buffers > 1) {
 		comp_err(dev, "module_adapter_prepare(): Invalid use of simple_copy");
 		return -EINVAL;
 	}
@@ -276,7 +291,7 @@ int module_adapter_prepare(struct comp_dev *dev)
 	 * no need to allocate intermediate sink buffers if the module produces only period bytes
 	 * every period and has only 1 input and 1 output buffer
 	 */
-	if (mod->simple_copy)
+	if (!IS_PROCESSING_MODE_RAW_DATA(mod))
 		return 0;
 
 	/* Module is prepared, now we need to configure processing settings.
@@ -710,7 +725,7 @@ module_single_source_setup(struct comp_dev *dev,
 	return num_output_buffers;
 }
 
-static int module_adapter_simple_copy(struct comp_dev *dev)
+static int module_adapter_audio_stream_type_copy(struct comp_dev *dev)
 {
 	struct comp_buffer __sparse_cache *source_c[PLATFORM_MAX_STREAMS];
 	struct comp_buffer __sparse_cache *sinks_c[PLATFORM_MAX_STREAMS];
@@ -746,11 +761,12 @@ static int module_adapter_simple_copy(struct comp_dev *dev)
 			num_input_buffers = 1;
 	}
 
-	ret = module_process(mod, mod->input_buffers, num_input_buffers,
-			     mod->output_buffers, num_output_buffers);
+	ret = module_process_legacy(mod, mod->input_buffers, num_input_buffers,
+				    mod->output_buffers, num_output_buffers);
 	if (ret) {
 		if (ret != -ENOSPC && ret != -ENODATA) {
-			comp_err(dev, "module_adapter_simple_copy() process failed with error: %x",
+			comp_err(dev,
+				 "module_adapter_audio_stream_type_copy() failed with error: %x",
 				 ret);
 			goto out;
 		}
@@ -825,7 +841,68 @@ out:
 	return ret;
 }
 
-int module_adapter_copy(struct comp_dev *dev)
+static int module_adapter_sink_source_copy(struct comp_dev *dev)
+{
+	struct comp_buffer __sparse_cache *source_buffers_c[PLATFORM_MAX_STREAMS];
+	struct comp_buffer __sparse_cache *sinks_buffers_c[PLATFORM_MAX_STREAMS];
+	struct sof_sink __sparse_cache *audio_sink[PLATFORM_MAX_STREAMS];
+	struct sof_source __sparse_cache *audio_src[PLATFORM_MAX_STREAMS];
+	struct processing_module *mod = comp_get_drvdata(dev);
+	struct list_item *blist;
+	uint32_t num_of_sources = 0;
+	uint32_t num_of_sinks = 0;
+	int ret;
+
+	comp_dbg(dev, "module_adapter_sink_source_copy(): start");
+
+	/* acquire all sink and source buffers, get handlers to sink/source API */
+	list_for_item(blist, &dev->bsink_list) {
+		struct comp_buffer *sink_buffer;
+
+		sink_buffer = container_of(blist, struct comp_buffer, source_list);
+		sinks_buffers_c[num_of_sinks] = buffer_acquire(sink_buffer);
+		audio_sink[num_of_sinks] =
+				audio_stream_get_sink(&sinks_buffers_c[num_of_sinks]->stream);
+		sink_reset_num_of_processed_bytes(audio_sink[num_of_sinks]);
+		num_of_sinks++;
+	}
+
+	list_for_item(blist, &dev->bsource_list) {
+		struct comp_buffer *source_buffer;
+
+		source_buffer = container_of(blist, struct comp_buffer, sink_list);
+		source_buffers_c[num_of_sources] = buffer_acquire(source_buffer);
+		audio_src[num_of_sources] =
+				audio_stream_get_source(&source_buffers_c[num_of_sources]->stream);
+		source_reset_num_of_processed_bytes(audio_src[num_of_sources]);
+		num_of_sources++;
+	}
+
+	ret = module_process_sink_src(mod, audio_src, num_of_sources, audio_sink, num_of_sinks);
+
+	if (ret != -ENOSPC && ret != -ENODATA && ret) {
+		comp_err(dev, "module_adapter_sink_source_copy() process failed with error: %x",
+			 ret);
+	}
+
+	/* release all source buffers in reverse order */
+	for (int i = num_of_sources - 1; i >= 0; i--) {
+		mod->total_data_consumed += source_get_num_of_processed_bytes(audio_src[i]);
+		buffer_release(source_buffers_c[i]);
+	}
+
+	/* release all sink buffers in reverse order */
+	for  (int i = num_of_sinks - 1; i >= 0 ; i--) {
+		mod->total_data_produced += sink_get_num_of_processed_bytes(audio_sink[i]);
+		buffer_release(sinks_buffers_c[i]);
+	}
+
+	comp_dbg(dev, "module_adapter_sink_source_copy(): done");
+
+	return ret;
+}
+
+static int module_adapter_raw_data_type_copy(struct comp_dev *dev)
 {
 	struct processing_module *mod = comp_get_drvdata(dev);
 	struct module_data *md = &mod->priv;
@@ -836,10 +913,7 @@ int module_adapter_copy(struct comp_dev *dev)
 	uint32_t min_free_frames = UINT_MAX;
 	int ret, i = 0;
 
-	comp_dbg(dev, "module_adapter_copy(): start");
-
-	if (mod->simple_copy)
-		return module_adapter_simple_copy(dev);
+	comp_dbg(dev, "module_adapter_raw_data_type_copy(): start");
 
 	list_for_item(blist, &mod->sink_buffer_list) {
 		sink = container_of(blist, struct comp_buffer, sink_list);
@@ -881,11 +955,12 @@ int module_adapter_copy(struct comp_dev *dev)
 		i++;
 	}
 
-	ret = module_process(mod, mod->input_buffers, mod->num_input_buffers,
-			     mod->output_buffers, mod->num_output_buffers);
+	ret = module_process_legacy(mod, mod->input_buffers, mod->num_input_buffers,
+				    mod->output_buffers, mod->num_output_buffers);
 	if (ret) {
 		if (ret != -ENOSPC && ret != -ENODATA) {
-			comp_err(dev, "module_adapter_copy() error %x: module processing failed",
+			comp_err(dev,
+				 "module_adapter_raw_data_type_copy() %x: module processing failed",
 				 ret);
 			goto out;
 		}
@@ -915,6 +990,8 @@ int module_adapter_copy(struct comp_dev *dev)
 
 	module_adapter_process_output(dev);
 
+	comp_dbg(dev, "module_adapter_raw_data_type_copy(): done");
+
 	return 0;
 
 out:
@@ -926,8 +1003,27 @@ out:
 		mod->input_buffers[i].size = 0;
 		mod->input_buffers[i].consumed = 0;
 	}
-
+	comp_dbg(dev, "module_adapter_raw_data_type_copy(): error %x", ret);
 	return ret;
+}
+
+int module_adapter_copy(struct comp_dev *dev)
+{
+	comp_dbg(dev, "module_adapter_copy(): start");
+
+	struct processing_module *mod = comp_get_drvdata(dev);
+
+	if (IS_PROCESSING_MODE_AUDIO_STREAM(mod))
+		return module_adapter_audio_stream_type_copy(dev);
+
+	if (IS_PROCESSING_MODE_RAW_DATA(mod))
+		return module_adapter_raw_data_type_copy(dev);
+
+	if (IS_PROCESSING_MODE_SINK_SOURCE(mod))
+		return module_adapter_sink_source_copy(dev);
+
+	comp_err(dev, "module_adapter_copy(): unknown processing_data_type");
+	return -EINVAL;
 }
 
 static int module_adapter_get_set_params(struct comp_dev *dev, struct sof_ipc_ctrl_data *cdata,
@@ -1135,16 +1231,13 @@ int module_adapter_reset(struct comp_dev *dev)
 		return ret;
 	}
 
-	if (!mod->simple_copy)
+	if (IS_PROCESSING_MODE_RAW_DATA(mod)) {
 		for (i = 0; i < mod->num_output_buffers; i++)
 			rfree((__sparse_force void *)mod->output_buffers[i].data);
-
-	rfree(mod->output_buffers);
-
-	if (!mod->simple_copy)
 		for (i = 0; i < mod->num_input_buffers; i++)
 			rfree((__sparse_force void *)mod->input_buffers[i].data);
-
+	}
+	rfree(mod->output_buffers);
 	rfree(mod->input_buffers);
 
 	mod->num_input_buffers = 0;
--- a/src/audio/rtnr/rtnr.c
+++ b/src/audio/rtnr/rtnr.c
@@ -381,7 +381,7 @@ static int rtnr_params(struct comp_dev *dev, struct sof_ipc_stream_params *param
 
 	/* set source/sink stream channels */
 	cd->sources_stream[0].channels = audio_stream_get_channels(&source_c->stream);
-	cd->sink_stream.channels = audio_stream_get_channels(&sink_c->stream);
+	audio_stream_set_channels(&cd->sink_stream, audio_stream_get_channels(&sink_c->stream));
 
 	/* set source/sink stream overrun/underrun permitted */
 	cd->sources_stream[0].overrun_permitted = audio_stream_get_overrun(&source_c->stream);
--- a/src/audio/sink_api_helper.c
+++ b/src/audio/sink_api_helper.c
@@ -1,0 +1,177 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// Copyright(c) 2023 Intel Corporation. All rights reserved.
+//
+
+#include <sof/audio/sink_api.h>
+#include <sof/audio/sink_api_implementation.h>
+#include <sof/audio/audio_stream.h>
+
+void sink_init(struct sof_sink __sparse_cache *sink, const struct sink_ops *ops,
+	       struct sof_audio_stream_params *audio_stream_params)
+{
+	sink->ops = ops;
+	sink->audio_stream_params = audio_stream_params;
+}
+
+size_t sink_get_free_size(struct sof_sink __sparse_cache *sink)
+{
+	return sink->ops->get_free_size(sink);
+}
+
+int sink_get_buffer(struct sof_sink __sparse_cache *sink, size_t req_size,
+		    void **data_ptr, void **buffer_start, size_t *buffer_size)
+{
+	int ret;
+
+	if (sink->requested_write_frag_size)
+		return -EBUSY;
+
+	ret = sink->ops->get_buffer(sink, req_size, data_ptr,
+						buffer_start, buffer_size);
+
+	if (!ret)
+		sink->requested_write_frag_size = req_size;
+	return ret;
+}
+
+int sink_commit_buffer(struct sof_sink __sparse_cache *sink, size_t commit_size)
+{
+	int ret;
+
+	/* check if there was a buffer obtained for writing by sink_get_buffer */
+	if (!sink->requested_write_frag_size)
+		return -ENODATA;
+
+	/* limit size of data to be committed to previously obtained size */
+	if (commit_size > sink->requested_write_frag_size)
+		commit_size = sink->requested_write_frag_size;
+
+	ret = sink->ops->commit_buffer(sink, commit_size);
+
+	if (!ret)
+		sink->requested_write_frag_size = 0;
+
+	sink->num_of_bytes_processed += commit_size;
+	return ret;
+}
+
+size_t sink_get_num_of_processed_bytes(struct sof_sink __sparse_cache *sink)
+{
+	return sink->num_of_bytes_processed;
+}
+
+void sink_reset_num_of_processed_bytes(struct sof_sink __sparse_cache *sink)
+{
+	sink->num_of_bytes_processed = 0;
+}
+
+enum sof_ipc_frame sink_get_frm_fmt(struct sof_sink __sparse_cache *sink)
+{
+	return sink->audio_stream_params->frame_fmt;
+}
+
+enum sof_ipc_frame sink_get_valid_fmt(struct sof_sink __sparse_cache *sink)
+{
+	return sink->audio_stream_params->valid_sample_fmt;
+}
+
+uint32_t sink_get_rate(struct sof_sink __sparse_cache *sink)
+{
+	return sink->audio_stream_params->rate;
+}
+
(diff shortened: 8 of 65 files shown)


src/include/sof/audio/source_api.h
```diff
@@ -0,0 +1,267 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
```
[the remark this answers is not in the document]
marcinszkudlinski: maybe, but separation is done for purpose 
1)there are entities that are source or sink only. They need to implement the corresponding part of function - i.e. DMIC is source only

2) yes, most common implementation will be a buffer (in this PR there's an audio stream, more types like DP Queue is coming) - and it has to implement both sink and source. The modules, however, will receive only handler to sink or to a source respectively

marcinszkudlinski:
Still fighting with some compilations checks. Anyway - ready for review

marc-hb:
> Still fighting with some compilations checks.

It means this PR contains significant changes. Have you considered spreading the commits in this PR across multiple PRs?
https://docs.zephyrproject.org/latest/contribute/contributor_expectations.html

Less "big-bang" and more "continuous" integration. Not just for compilation checks but for daily tests too.

src/samples/audio/smart_amp_test_ipc3.c
```diff
@@ -537,8 +537,8 @@ static int smart_amp_prepare(struct comp_dev *dev)
 	if (sad->feedback_buf) {
 		struct comp_buffer __sparse_cache *buf = buffer_acquire(sad->feedback_buf);
 
-		buf->stream.channels = sad->config.feedback_channels;
-		buf->stream.rate = audio_stream_get_rate(&buffer_c->stream);
+		audio_stream_set_channels(&buf->stream, sad->config.feedback_channels);
+		audio_stream_set_rate(&buf->stream, audio_stream_get_rate(&buffer_c->stream));
```
aiChaoSONG, line 541: this belongs to the first commit

src/include/sof/audio/source_api.h
```diff
@@ -0,0 +1,267 @@
...
+	/**
+	 * OPTIONAL: Notification to the source implementation about changes in audio format
+	 */
+	void (*on_audio_format_set)(audio_src_hnd src_handler);
+};
+
+struct source {
+	const struct source_ops *ops;
+	uint32_t locked_read_frag_size;
+	uint32_t num_of_bytes_processed;		     /** processed bytes counter */
```
lyakh, line 111: `size_t` for both?

src/include/sof/audio/source_api.h
```diff
@@ -0,0 +1,267 @@
...
+	struct sof_audio_stream_params *audio_stream_params; /** pointer to audio params */
+};
+
+static inline void source_init(audio_src_hnd src_handler, const struct source_ops *ops,
+			       struct sof_audio_stream_params *audio_stream_params)
+{
+	struct source *source = (struct source *)src_handler;
```
lyakh, line 118: let's not do that. Please define a proper opaque type instead of a `void *` macro

src/include/sof/audio/source_api.h
```diff
@@ -0,0 +1,267 @@
...
+
+	/**
+	 * OPTIONAL: Notification to the source implementation about changes in audio format
+	 */
+	void (*on_audio_format_set)(audio_src_hnd src_handler);
+};
+
+struct source {
+	const struct source_ops *ops;
+	uint32_t locked_read_frag_size;
```
lyakh, line 110: this field seems the least clear so it needs a comment the most

src/include/sof/audio/source_api.h
```diff
@@ -0,0 +1,267 @@
...
+}
+
+static inline int source_get_data(
+		audio_src_hnd src_handler, uint32_t req_size,
+		void **data_ptr, void **buffer_start, void **buffer_end, uint32_t *buffer_size)
+{
+	struct source *source = (struct source *)src_handler;
+	int ret;
+
+	if (source->locked_read_frag_size)
```
lyakh, line 139: once we know what `locked_read_frag_size` should mean we'll understand the usage better too

src/include/sof/audio/source_api.h
```diff
@@ -0,0 +1,267 @@
...
+		source->locked_read_frag_size = req_size;
+	return ret;
+}
+
+static inline int source_put_data(audio_src_hnd src_handler, uint32_t free_size)
+{
+	struct source *source = (struct source *)src_handler;
+	int ret;
+
+	if (!source->locked_read_frag_size)
```
lyakh, line 155: reading data prepares it for writing? Not very clear...

src/include/sof/audio/sink_api.h
```diff
@@ -0,0 +1,263 @@
...
+ * Examples of data sinks:
+ * - /dev/null
+ *     all the data stored in sink buffer are just simply discarded
+ * - I2S sender
+ *     Data stored in sink buffer will be sent to the external world
+ * - a memory ring buffer
+ *     data stored in the buffer will be sent to another module (usually using source API, but it
+ *     does not matter in fact).
+ *
+ *  The main advantage of using source API instead of just taking pointers to the buffers is that
```
lyakh, line 29: typo "sink" - would be good to double-check for copy-paste typos

src/include/sof/audio/sink_api.h
```diff
@@ -0,0 +1,263 @@
...
+ *   - part of the buffer may be locked to prevent reading
+ *   etc.etc. it depends on implementation of the data sink
+ *
+ * NOTE: the module should get a complete portion of space it needs for processing, fill it
+ *       than release. The reason is - the depending on the implementation, the calls may be
+ *       expensive - may involve some data moving in memory, cache writebacks, etc.
+ */
+
+/* sink structure obsfucator */
+#define audio_sink_hnd void*
```
lyakh, line 42: ditto

src/include/sof/audio/sink_api.h
```diff
@@ -0,0 +1,263 @@
...
+
+struct sink_ops {
+	/**
+	 * Retrieves size of free space available in sink (in bytes)
+	 * return number of free bytes in buffer available to immediate filling
+	 */
+	uint32_t (*get_free_size)(audio_sink_hnd sink_handler);
```
lyakh, line 49: `size_t` - for all sizes

src/include/sof/audio/sink_api.h
```diff
@@ -0,0 +1,263 @@
...
+
+	/**
+	 * OPTIONAL: Notification to the source implementation about changes in audio format
+	 */
+	void (*on_audio_format_set)(audio_sink_hnd sink_handler);
+};
+
+struct sink {
+	const struct sink_ops *ops;	/** operations interface */
+	uint32_t locked_write_frag_size;
```
lyakh, line 94: comment

src/include/sof/audio/sink_api.h
```diff
@@ -0,0 +1,263 @@
...
+	 * @param [out] data_ptr a pointer to the space will be provided there
+	 * @param [out] buffer_start pointer to circular buffer start
+	 * @param [out] buffer_end pointer to circular buffer end
+	 * @param [out] buffer_size size of circular buffer
+	 *
+	 * @retval -ENODATA if req_size is bigger than free space
+	 *
+	 */
+	int (*get_buffer)(audio_sink_hnd src_handler, uint32_t req_size, void **data_ptr,
+			  void **buffer_start, void **buffer_end, uint32_t *buffer_size);
```
lyakh, line 69: I don't think these should be exposed. This should be implementation internal. You should only export things like read data, write data, get available data amount, get free space etc. No pointers at all! I think our existing `audio_stream_produce()` / `audio_stream_consume()` are a good direction to follow.

src/include/sof/audio/module_adapter/module/module_interface.h
```diff
@@ -81,15 +83,53 @@ struct module_interface {
...
-	 * component copy in .copy(). This procedure is responsible to consume
+	 * Module specific processing procedure
+	 * This procedure is responsible to consume
 	 * samples provided by the module_adapter and produce/output the processed
 	 * ones back to module_adapter.
+	 *
+	 * there are 3 versions of the procedure, the difference is the format of
+	 * input/output data
+	 *
+	 * the module MUST implement one and ONLY one of them
```
lyakh, line 96: use a `union`?

lyakh (changes requested):
I think the API can and should be improved

src/include/sof/audio/source_api.h
```diff
@@ -0,0 +1,267 @@
...
+	 *
+	 *  and the data receiver may use it as usual, rollover will simple never occur
+	 *  NOTE! the caller MUST NOT assume that pointers to start/end of the circular buffer
+	 *	  are constant. They may change between calls
+	 *
+	 * @param [in] req_size requested size of data.
+	 * @param [out] data_ptr a pointer to data will be provided there
+	 * @param [out] buffer_start pointer to circular buffer start
+	 * @param [out] buffer_end pointer to circular buffer end
+	 * @param [out] buffer_size size of circular buffer
```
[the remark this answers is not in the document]
marcinszkudlinski: yes, it is. Just making it easier for the user

src/include/sof/audio/module_adapter/module/module_interface.h
marcinszkudlinski: We talked about it already ;) and it has been rejected

3 function pointers are to show
1) there are 3 APIs here (raw/stream/snk_src)
2) make module_adapter aware what API the module uses

alternatively I need to introduce an enum, like simple_copy,  and make the module set it.

src/include/sof/audio/source_api.h
marcinszkudlinski: comment added

src/audio/CMakeLists.txt
```diff
@@ -11,6 +11,8 @@ if((NOT CONFIG_LIBRARY) OR CONFIG_LIBRARY_STATIC)
 		host-legacy.c
 		component.c
 		buffer.c
+		source_api_helper.c
+
```
btian1: one more space line not needed

src/audio/source_api_helper.c
```diff
@@ -0,0 +1,135 @@
+// SPDX-License-Identifier: BSD-3-Clause
```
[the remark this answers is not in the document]
marcinszkudlinski: The main benefit is that the API is flexible. Is not connected with any buffer etc.
Please see here - the pipeline 2.0 idea - https://github.com/thesofproject/sof/issues/7261

The target is to get rid of module_adapter and other layers and let the modules work without knowing how the data are processed by the pipeline. Currently module - no matter if connected using module adapter or directly to the pipeline - must use audio stream which is tightly bind with buffer.c. Module must do many operation itself - like buffer_aquire. In 2.0 approach a module does not need to know a bit about buffering, caching etc. All it needs to do is request data, process, release / request buffer, fill with result, commit. 
All underlaying operations should be performed by the sink/source implementation

i.e. lets assume a module needs a linear data at input - can't deal with circular buffers
All the pipeline needs to do is use a sink/source implementation that is able  to linearize data and connect use it to connect the module in question with its predecessor. The predecessor itself won't even notice
(such implementation does not exist yet - currently only a very simple wrapper for audio_stream is implemented in this PR)


second example - DP queue. It is currently under development, a kind of buffer that connects 2 modules with different processing periods, possibly on separate cores, taking care of all cache coherency etc. 

MAIN IDEA - the module itself does not notice if it talks to a simple buffer, a DP queue, a DMIC (which in fact is a source!), or whatever. That requires an API with strict internal data encapsulation

src/include/sof/audio/source_api.h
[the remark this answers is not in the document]
lyakh: personally to me it makes this more confusing - it gives an impression of something non-trivial happening there

src/include/sof/audio/sink_api.h
lyakh: That my comment was wrong. We do need pointers because an alternative would be copying data every time for module processing and that would be too expensive. Currently those pointers are obtained by `audio_stream_get_rptr()` and `audio_stream_get_wptr()`. But I'd still remove the last parameter

src/include/sof/audio/source_api.h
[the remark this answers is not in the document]
marcinszkudlinski: Maybe you're right, 
I'll remove buffer_end than

src/audio/sink_api_helper.c
```diff
@@ -0,0 +1,144 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// Copyright(c) 2023 Intel Corporation. All rights reserved.
+//
+
+#include <sof/audio/sink_api.h>
+#include <sof/audio/sink_api_internal.h>
+
+inline void sink_init(struct sink *sink, const struct sink_ops *ops,
+		      struct sof_audio_stream_params *audio_stream_params)
```
lyakh, line 10: if possible `const`

src/audio/sink_api_helper.c
```diff
@@ -0,0 +1,144 @@
...
+	ret = sink->ops->commit_buffer(sink, commit_size);
+
+	if (!ret)
+		sink->locked_write_frag_size = 0;
+
+	sink->num_of_bytes_processed += commit_size;
+	return ret;
+}
+
+inline uint32_t sink_get_num_of_processed_bytes(audio_sink_hnd sink)
```
lyakh, line 59: same types as in source part, e.g. `size_t`

src/audio/audio_stream.c
```diff
@@ -0,0 +1,118 @@
...
+		audio_stream_consume(audio_stream, free_size);
+
+	return 0;
+}
+
+static const struct source_ops audio_stream_source_ops = {
+	.get_data_available = audio_stream_get_data_available,
+	.get_data = audio_stream_get_data,
+	.put_data = audio_stream_put_data,
+	.on_audio_format_set = NULL
```
lyakh, line 90: superfluous `NULL` initialisation

src/include/sof/audio/audio_stream.h
```diff
@@ -43,8 +47,13 @@
  * consumption/production and update the buffer state by calling
  * audio_stream_consume()/audio_stream_produce() (just a single call following
  * series of reads/writes).
+ *
+ * Audio stream implements pipeline2.0 sink and source API
  */
 struct audio_stream {
+	struct source source_api;	/**< source API */
+	struct sink sink_api;		/**< sink API */
```
lyakh, line 55: this doesn't seem right to me. These structures should be opaque. And you include them and the "internal" header into a public header, that "everybody" includes.

src/audio/module_adapter/module/generic.c
```diff
@@ -245,8 +251,21 @@ int module_process(struct processing_module *mod, struct input_stream_buffer *in
...
+	else if (md->ops->process_audio_stream)
+		ret = md->ops->process_audio_stream(mod,
+				(struct input_stream_buffer *)sources, num_of_sources,
+				(struct output_stream_buffer *)sinks, num_of_sinks);
+	else if (md->ops->process_raw_data)
+		ret = md->ops->process_audio_stream(mod,
+				(struct input_stream_buffer *)sources, num_of_sources,
+				(struct output_stream_buffer *)sinks, num_of_sinks);
+	else
+		ret = -EINVAL;
```
lyakh, line 267: this cannot happen, you've checked for it already, maybe an assert?

src/include/sof/audio/source_api.h
```diff
@@ -0,0 +1,137 @@
...
+ * implemented and used as a mid-layer for modules needing it.
+ *
+ * NOTE: the module should get a complete portion of data it needs for processing, process it
+ *       than release. The reason is - the depending on the implementation, the calls may be
+ *       expensive - may involve some data moving in memory, cache writebacks, etc.
+ */
+
+/** definition of obfsfucated handler of source API */
+struct source;
+typedef struct source *audio_src_hnd;
```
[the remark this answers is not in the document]
kv2019i: Hmm, we've not used typedeffed struct pointers in SOF so far, so why not just use "struct source*"... ?

src/audio/module_adapter/module_adapter.c
```diff
@@ -240,7 +253,9 @@ int module_adapter_prepare(struct comp_dev *dev)
 		return -EINVAL;
 	}
 
-	if (mod->simple_copy && mod->num_input_buffers > 1 && mod->num_output_buffers > 1) {
+	/* check processing mode */
+	if (IS_PROCESSING_MODE_AUDIO_STREAM(mod) &&
+	    mod->num_input_buffers > 1 && mod->num_output_buffers > 1) {
```
kv2019i, line 258: This is nice and helps with code readability as well.

marcinszkudlinski:
(rebase to newest main only)

src/audio/source_api_helper.c
```diff
@@ -0,0 +1,135 @@
...
+
+	ret = source->ops->get_data(source, req_size, data_ptr,
+						 buffer_start, buffer_end, buffer_size);
+
+	if (!ret)
+		source->locked_read_frag_size = req_size;
+	return ret;
+}
+
+inline int source_put_data(audio_src_hnd source, size_t free_size)
```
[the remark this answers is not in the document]
marcinszkudlinski: For me - yes, release is much better. 
Yet there were discussions and finally PUT won
@plbossart  ?

src/include/sof/audio/source_api.h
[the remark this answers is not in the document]
marcinszkudlinski: Up to you, personally I prefer using a handler instead of a pointer. Much less typing

src/include/sof/audio/audio_stream.h
marcinszkudlinski: Not everybody, just implementations of the sink/source
and I don't see any way to encapsulate it more because
1) implementations must implement all function pointers 
2) in most cases struct sink / struct source will be a static member of the implementation (as it is in audio_stream now). To make it static - at least a size of the structure must be known. It is no way in C to encapsulate it effectively.

In C++ is simple - implementations inherit from sink/source and implement protected virtual functions, having no access to private members. Unfortunately it is C :(
All I could do is to put struct source/sink into .c file and export create_() that would malloc space for structures. I really don't want to do this for many reasons.

I change sink_api_internals.h to sink_api_implementation.h and add some comments about it

src/audio/asrc/asrc.c
```diff
@@ -555,7 +555,7 @@ static int asrc_params(struct comp_dev *dev,
 
 	/* Don't change sink rate if value from IPC is 0 (auto detect) */
 	if (asrc_get_sink_rate(&cd->ipc_config))
-		sink_c->stream.rate = asrc_get_sink_rate(&cd->ipc_config);
+		audio_stream_set_rate(&sink_c->stream, asrc_get_sink_rate(&cd->ipc_config));
```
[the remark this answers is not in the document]
marcinszkudlinski: needed all functions to extract struct sof_audio_stream_params

src/audio/sink_api_helper.c
marcinszkudlinski: no, const is not an option as there are SET functions in API

src/audio/source_api_helper.c
```diff
@@ -0,0 +1,135 @@
...
+{
+	return source->audio_stream_params->underrun_permitted;
+}
+
+inline void source_set_valid_fmt(audio_src_hnd source,
+				      enum sof_ipc_frame valid_sample_fmt)
+{
+	source->audio_stream_params->valid_sample_fmt = valid_sample_fmt;
+	if (source->ops->on_audio_format_set)
+		source->ops->on_audio_format_set(source);
```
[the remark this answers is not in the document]
marcinszkudlinski: Good point, 
To avoid creating a function pointer for every parameter I thought about change notification only. But yes, some validation should be there. Will think about it

src/audio/module_adapter/module/generic.c
```diff
@@ -245,8 +251,21 @@ int module_process(struct processing_module *mod, struct input_stream_buffer *in
...
+				(struct source **)sources, num_of_sources,
+				(struct sink **)sinks, num_of_sinks);
+	else if (md->ops->process_audio_stream)
+		ret = md->ops->process_audio_stream(mod,
+				(struct input_stream_buffer *)sources, num_of_sources,
+				(struct output_stream_buffer *)sinks, num_of_sinks);
+	else if (md->ops->process_raw_data)
+		ret = md->ops->process_audio_stream(mod,
+				(struct input_stream_buffer *)sources, num_of_sources,
+				(struct output_stream_buffer *)sinks, num_of_sinks);
```
lyakh, line 265: this doesn't look right either. Maybe those structures need `.data` or reuse `struct input_stream_buffer` / `struct output_stream_buffer`

src/audio/module_adapter/module/generic.c
```diff
@@ -225,9 +228,12 @@ int module_prepare(struct processing_module *mod)
...
 
-int module_process(struct processing_module *mod, struct input_stream_buffer *input_buffers,
-		   int num_input_buffers, struct output_stream_buffer *output_buffers,
-		   int num_output_buffers)
+/* for simplification, module process gets a void pointers as data types
+ * and calls proper handler at the end
+ */
+int module_process(struct processing_module *mod,
+		   void *sources, int num_of_sources,
+		   void *sinks, int num_of_sinks)
```
lyakh, line 236: `sources` and `sinks` should have their proper `struct source **` and `struct sink **` types with proper sparse annotations

src/audio/module_adapter/module/generic.c
```diff
@@ -245,8 +251,21 @@ int module_process(struct processing_module *mod, struct input_stream_buffer *in
...
+	if (md->ops->process)
+		ret = md->ops->process(mod,
+				(struct source **)sources, num_of_sources,
+				(struct sink **)sinks, num_of_sinks);
+	else if (md->ops->process_audio_stream)
+		ret = md->ops->process_audio_stream(mod,
+				(struct input_stream_buffer *)sources, num_of_sources,
+				(struct output_stream_buffer *)sinks, num_of_sinks);
+	else if (md->ops->process_raw_data)
+		ret = md->ops->process_audio_stream(mod,
```
lyakh, line 263: `->process_raw_data()`?

src/include/sof/audio/audio_stream.h
```diff
@@ -15,6 +15,10 @@
 #define __SOF_AUDIO_AUDIO_STREAM_H__
 
 #include <sof/audio/format.h>
+#include <sof/audio/sink_api.h>
+#include <sof/audio/sink_api_implementation.h>
+#include <sof/audio/source_api.h>
+#include <sof/audio/source_api_implementation.h>
```
lyakh, line 21: again both implementers and users will get these headers. So, yes, they are in effect public.

src/audio/module_adapter/module/generic.c
```diff
@@ -245,8 +251,21 @@ int module_process(struct processing_module *mod, struct input_stream_buffer *in
...
-	ret = md->ops->process(mod, input_buffers, num_input_buffers, output_buffers,
-			       num_output_buffers);
+	if (md->ops->process)
+		ret = md->ops->process(mod,
+				(struct source **)sources, num_of_sources,
+				(struct sink **)sinks, num_of_sinks);
+	else if (md->ops->process_audio_stream)
+		ret = md->ops->process_audio_stream(mod,
+				(struct input_stream_buffer *)sources, num_of_sources,
+				(struct output_stream_buffer *)sinks, num_of_sinks);
```
lyakh, line 261: Don't think this can work - especially for `sinks` - they aren't even in the beginning of the structure. I think you need temporary arrays

src/audio/sink_source_utils.c
```diff
@@ -0,0 +1,52 @@
...
+
+#include <sof/audio/sink_source_utils.h>
+#include <sof/audio/sink_api.h>
+#include <sof/audio/source_api.h>
+#include <sof/common.h>
+#include <rtos/string.h>
+#include <sof/math/numbers.h>
+#include <assert.h>
+
+int source_to_sink_copy(struct source *source, struct sink *sink, bool free, size_t size)
```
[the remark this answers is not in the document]
marcinszkudlinski: OMG, much worse. I forgot to release data / put buffer.... 😁

src/audio/module_adapter/module/generic.c
marcinszkudlinski: i'd like to get rid of input_stream_buffer / output_stream_buffer - those are taken from IADK modules and in fact have no use here.

src/include/sof/audio/audio_stream.h
marcinszkudlinski: here everything is "private"
What do you think about:

public:
`  struct * source;  -  will never be defined, intended to use as a pointer only`

implementation:
```
  struct source_private {
   int a, b ,c; 
   }
 
```
and in .c fie:
```

  define get_source_priv(_src) (struct source_private *)_src 
  
  int source_something(struct *source _source) 
  {
      struct source_private *source = get_source_priv(_source);
    
      source->a =.... 
  }
```

lyakh: I don't see what this brings - implementers, users, etc. anyway have to know the size to allocate / include in their structures, and if they use `struct source_private` for that - what have we won?

marcinszkudlinski: you're right
Looks like there's no way to block access at compiler level without mallocs

kv2019i: @marcinszkudlinski @lyakh  I guess this depends on how much encapsulation we want. I don't think blocking access is a huge deal, but I if we want to have sink/source implementations that have a vastly more complex implementation, it's very likely they require some of their own state. So if we need this, using a private pointer plus a type-specific init, would seem like a good, well-known solution. Should be familiar to C++ folks as well (as the d-pointer pattern -> https://en.wikipedia.org/wiki/Opaque_pointer ). 
Simple types (like audio_stream.h) could just leave the private pointer unused and no heap/malloc would be needed (in our current uses).

We can surely do this incrementally. We only have one implmentation now, so there's quite the risk of overengineering here without actual users and as long as client code just accesses the public functions, were are free to modify the implementation in follow-up PRs.

src/audio/module_adapter/module/generic.c
marcinszkudlinski: redesigned,

marcinszkudlinski:
Sparse cache should be cleared now
Still setting audio params API needs to be fixed: 
1) check in the params are valid
2) cache - current approach is fast but cannot work in case of cross core queue

src/audio/audio_stream.c
```diff
@@ -0,0 +1,110 @@
...
+{
+	struct audio_stream __sparse_cache *audio_stream =
+			container_of(sink, struct audio_stream, sink_api);
+
+	return audio_stream_get_free_bytes(audio_stream);
+}
+
+static int audio_stream_get_buffer(struct sink __sparse_cache *sink, size_t req_size,
+				   void __sparse_cache **data_ptr,
+				   void __sparse_cache **buffer_start, size_t *buffer_size)
```
lyakh, line 19: ah, this is actually interesting now. To which memory does the attribute apply and can we control that? E.g. `data_ptr` - does the memory, to which it points now reside in the `__cache` address space, or the memory, to which `*data_ptr` points?.. It's similar to the `const` attribute: `const void **ptr` and `void * const *ptr` are two different things. But I'm not sure if moving a sparse attribute around has the same effect...

src/audio/sink_source_utils.c
```diff
@@ -0,0 +1,71 @@
...
+
+int source_to_sink_copy(struct source __sparse_cache *source, struct sink __sparse_cache *sink,
+			bool free, size_t size)
+{
+	uint8_t __sparse_cache *src_ptr;
+	uint8_t __sparse_cache *src_begin;
+	uint8_t __sparse_cache *src_end;
+	uint8_t __sparse_cache *dst_ptr;
+	uint8_t __sparse_cache *dst_begin;
+	uint8_t __sparse_cache *dst_end;
```
lyakh, line 22: I think that's too much now... Currently we use `__sparse_cache` for `struct comp_buffer` and since `struct audio_stream` since it's embedded into it. But we don't mark buffer data as `__sparse_cache` even though we allocate it as cached (`rballoc()` always returns cached addresses). So I think most of your added annotations aren't needed.

btian1 (commented):
1. shall we create a new folder under src/audio/ to store new added source files? like pipe2.0
    I guess this is just a beginning, and later will have more source files adding.

2. with new files add in, possible to have a check on sof fw size? before and after? it may add 20K code size, just a guess.

marcinszkudlinski:
(last push - rebase and CI checks cleanup)

cujomalainey (commented):
Only got part way through, will finish tomorrow

src/audio/sink_api_helper.c
```diff
@@ -0,0 +1,142 @@
...
+	ret = sink->ops->commit_buffer(sink, commit_size);
+
+	if (!ret)
+		sink->locked_write_frag_size = 0;
+
+	sink->num_of_bytes_processed += commit_size;
+	return ret;
+}
+
+size_t sink_get_num_of_processed_bytes(struct sink __sparse_cache *sink)
```
cujomalainey, line 59: i feel like all of these should be annotate with `inline`

src/audio/audio_stream.c
```diff
@@ -0,0 +1,116 @@
...
+	.get_buffer = audio_stream_get_buffer,
+	.commit_buffer = audio_stream_commit_buffer
+};
+
+inline void audio_stream_init(struct audio_stream __sparse_cache *buffer,
+			      void *buff_addr, uint32_t size)
+{
+	buffer->size = size;
+	buffer->addr = buff_addr;
+	buffer->end_addr = (char *)buffer->addr + size;
```
cujomalainey, line 103: i don't see this being used anywhere

src/audio/src/src.c
```diff
@@ -335,57 +335,64 @@ int src_polyphase_init(struct polyphase_src *src, struct src_param *p,
...
 	int s2_blk_in;
 	int s2_blk_out;
-	void *sbuf_addr = cd->delay_lines;
+	uint32_t n_read = 0, n_written = 0;
+	int ret;
+	void __sparse_cache *buffer_start;
 	void *sbuf_end_addr = &cd->delay_lines[cd->param.sbuf_length];
 	size_t sbuf_size = cd->param.sbuf_length * sizeof(int32_t);
-	int nch = audio_stream_get_channels(source);
+	int nch = source_get_channels(source);	/* src channels must == sink channels */
```
cujomalainey, line 359: this comment looks like it belongs in prepare

src/audio/src/src.c
```diff
@@ -558,11 +583,11 @@ static void src_set_sink_params(struct comp_dev *dev, struct comp_buffer __spars
...
 				    cd->ipc_config.base.audio_fmt.s_type);
+	sink_set_frm_fmt(sink, frame_fmt);
+	sink_set_valid_fmt(sink, valid_fmt);
+	sink_set_channels(sink, cd->ipc_config.base.audio_fmt.channels_count);
 
-	audio_stream_set_frm_fmt(&sinkb->stream, frame_fmt);
-	audio_stream_set_valid_fmt(&sinkb->stream, valid_fmt);
-	audio_stream_set_channels(&sinkb->stream, cd->ipc_config.base.audio_fmt.channels_count);
-
+	/*TODO: change to sink/src interface */
```
cujomalainey, line 590: for POC we should probably close these out

src/audio/src/src.c
```diff
@@ -602,28 +627,22 @@ static int src_set_params(struct processing_module *mod)
 	return 0;
 }
 
-static void src_set_sink_params(struct comp_dev *dev, struct comp_buffer __sparse_cache *sinkb) {}
+static void src_set_sink_params(struct comp_dev *dev, struct sink __sparse_cache *sink,
+				struct comp_buffer __sparse_cache *sinkb)
+{
+	/* empty */
```
cujomalainey, line 633: duplicate of above signature

cujomalainey (changes requested):

src/audio/sink_api_helper.c
lyakh: this seems to be an exported function, so it cannot be `inline`

marcinszkudlinski: It may be, it won't hurt. Theoretically is possible to inline exported function, however, I don't think clang compiler/linker is smart enough to do it.

src/audio/audio_stream.c
marcinszkudlinski: It is copy&paste with what was here before. 
I don't like the idea of fixing audio_stream in the same PR

src/audio/src/src.c
marcinszkudlinski: Disagree. It is usage of new interface, the very last commit

src/audio/src/src.c
marcinszkudlinski: Can you explain?

src/audio/audio_stream.c
marcinszkudlinski: audio_stream_get_end_addr() is used 36 times
audio_stream_get_size() 46 times
nothing to fix, it is being used

src/audio/sink_api_helper.c
andyross: Yeah, that's a LTO feature, and nothing has wired it up for Zephyr builds yet (and in any case the Xtensa clang is based on an LLVM too old to support it anyway).

There's lots of aesthetic argument over inline in C.  FWIW, conventions within the Zephyr tree itself (whether SOF wants to honor that is yet another argument, heh) have largely fallen out as:

* Use "static inline" liberally as a trick to allow function bodies to appear in header files and achieve cheap "LTO-lite" advantages for simple APIs (at the cost of the occasional horrifying knot of header dependency hell, to be fair).

* Generally don't try to use inline for performance choices within a single C file.  Modern compilers are probably smarter about this than you are.  If you are tempted, at least look at the generated code first to be sure you know what you want and what you are getting.

* If you are at the point where you're looking at generated code and hyperoptimizing, prefer the `ALWAYS_INLINE` macro instead, which is a firmer promise that Zephyr provides for all toolchains[1].

[1] There are a tiny handful of spots where we rely on this for correctness reasons, e.g. a macro-like "C function" in the arch layer that needs to inspect the stack pointer and will behave differently if it's in a different frame.  There are situations in Linux where this is needed for memory barrier control too, IIRC.

src/audio/src/src.c
cujomalainey: Ah nvm, I missed "Sink" somehow in the name of above function

src/audio/sink_api_helper.c
andyross: Oh, and to jump in on the export bit: there's no particular limitation there.  Obviously a function **called** from an external translation unit can only be a true function, but within the file the compiler is free to inline if it wants to, subject to whatever the heuristics tell it is a good idea (being a visible symbol, it's going to be required to be present in the output assembly anyway, so -Os might say "no inlining" for that reason, etc...)

src/audio/src/src.c
cujomalainey: yes but we should not be making comments about our assumed state in the processing function, we should be validating that in prepare. It just feels weird the way its written like a check is assumed here. Whereas I see its used to explain the logic and why there is an optimization. Maybe "chan sink == chan src therefore we only need to use one"

src/audio/audio_stream.c
cujomalainey: I think I confused inline with static in my review, apologies

src/audio/rtnr/rtnr.c
```diff
@@ -381,7 +381,7 @@ static int rtnr_params(struct comp_dev *dev, struct sof_ipc_stream_params *param
 
 	/* set source/sink stream channels */
 	cd->sources_stream[0].channels = audio_stream_get_channels(&source_c->stream);
-	cd->sink_stream.channels = audio_stream_get_channels(&sink_c->stream);
+	audio_stream_get_channels(cd->sink_stream) = audio_stream_get_channels(&sink_c->stream);
```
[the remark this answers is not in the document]
marcinszkudlinski: yeah, looks like it is never compiled as CI did not complain.,

marcinszkudlinski:
sink/src interface extended to fulfill prepare needs
module_prepare api extended to (optional) use sink/source 
SRC module use exclusively sink/src now, no access to audio_stream nor to comp_buffer left

please re-review

src/include/sof/audio/format.h
```diff
@@ -115,6 +116,29 @@
 #include "format_hifi3.h"
 #endif
 
+struct sof_audio_stream_params {
```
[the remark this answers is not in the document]
marcinszkudlinski: moving to audio_stream by now

marcinszkudlinski:
rebased

src/audio/audio_stream.c
```diff
@@ -0,0 +1,175 @@
...
+	struct audio_stream __sparse_cache *audio_stream =
+			attr_container_of(sink, struct audio_stream __sparse_cache,
+					  sink_api, __sparse_cache);
+
+	if (req_size > audio_stream_get_free_size(sink))
+		return -ENODATA;
+
+	/* get circular buffer parameters */
+	*data_ptr = (__sparse_force void __sparse_cache *)audio_stream->w_ptr;
+	*buffer_start = (__sparse_force void __sparse_cache *)audio_stream->addr;
```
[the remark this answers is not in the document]
marcinszkudlinski: I think enforcing audio processing as cached is the only possible way to keep performance.
I see no reason not do do it, especially in sink/source approach, where modules are not required to perform any special actions to keep data coherent. 
Anyway, it is subject to talk.

marcinszkudlinski:
>>**What would a "Bytes" format look like for compressed data?**

@cujomalainey  This commit does not modify audio formats - just simple change interface to cover the same functionality in diffrent way. Compressed format is another subject, but not part of this PR

src/audio/audio_stream.c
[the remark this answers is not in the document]
lyakh: @lgirdwood would adding `__sparse_cache` annotations to audio buffer pointers be ok with you?

src/audio/module_adapter/module_adapter.c
```diff
@@ -165,6 +165,55 @@ struct comp_dev *module_adapter_new(const struct comp_driver *drv,
...
+		audio_src[num_of_sources] =
+				audio_stream_get_source(&source_buffers_c[num_of_sources]->stream);
+		source_reset_num_of_processed_bytes(audio_src[num_of_sources]);
+		num_of_sources++;
+	}
+
+	/* Prepare module */
+	ret = module_prepare(mod, audio_src, num_of_sources, audio_sink, num_of_sinks);
+
+	/* release all source buffers in reverse order */
```
kv2019i, line 206: why reverse order?

kv2019i (approved):
Some nitpicks and places to improve, but no showstoppers.

src/audio/module_adapter/module_adapter.c
marcinszkudlinski: According to @lyakh  it has to be this way. I don't fully understand why,

lyakh: @kv2019i In general you don't do
```
lock(a);
lock(b);
...
unlock(a);
unlock(b);
```
but in the Zephyr spinlock case the reason is also rather simple: `k_spin_unlock()` restores the CPU state to the one saved by the respective `k_spin_lock()`. So typically you take the first spinlock with interrupts enabled, the second one, however, already with interrupts disabled. So if you then unlock in the wrong order, you first restore to the state with interrupts enabled, which is too early at that point, but then again to interrupts disabled, and you leave your context that way - with interrupts still disabled.

marcinszkudlinski:
SOFCI TEST

src/audio/audio_stream.c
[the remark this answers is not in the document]
marcinszkudlinski: to speed up - this PR is open long enough ;) - I removed by now sparse from pointers, with some comments

marcinszkudlinski:
Please re-review and merge if possible
sparse_cache from buffer pointers removed

cujomalainey:
please rebase you have conflicts

marcinszkudlinski:
rebased

btian1 (commented):
+1917， -468 line code changes,  should have a data to cover FW size before and after this PR?

marcinszkudlinski:
@btian1 
size of .text is 1988bytes (decimal) more
size of .rodata is 84 bytes more

before: 
	No	Start		End		Size	Type	Name
	1	0xa1048000	0xa10484e4	0x4e4	TEXT	.imr
	2	0xa0030000	0xa003016a	0x16a	TEXT	.WindowVectors.text
	3	0xa0030180	0xa00301a9	0x29	TEXT	.Level2InterruptVector.text
	4	0xa00301c0	0xa00301e9	0x29	TEXT	.Level3InterruptVector.text
	5	0xa0030200	0xa0030229	0x29	TEXT	.DebugExceptionVector.text
	6	0xa00302c0	0xa00302e9	0x29	TEXT	.NMIExceptionVector.text
	7	0xa0030300	0xa0030303	0x3	TEXT	.KernelExceptionVector.text
	8	0xa0030340	0xa0030352	0x12	TEXT	.UserExceptionVector.text
	9	0xa00303c0	0xa00303c6	0x6	TEXT	.DoubleExceptionVector.text
	10	0xa0030400	0xa0087254	0x56e54	TEXT	.text
	11	0xa0088000	0xa00dfbb0	0x57bb0	DATA	.rodata
	12	0xa00dfbb0	0xa00dff10	0x360	DATA	initlevel
	13	0xa00dff10	0xa00e070c	0x7fc	DATA	device_area
	14	0xa00e070c	0xa00e0724	0x18	DATA	k_p4wq_initparam_area
	15	0xa00e0730	0xa00e0a06	0x2d6	DATA	device_handles
	16	0xa00e0a08	0xa00e0c48	0x240	DATA	log_const_area
	17	0xa00e0c48	0xa00e0c68	0x20	DATA	log_backend_area
	18	0xa00e0c68	0xa00e0cd4	0x6c	DATA	.fw_ready
	19	0x400e0d00	0x400e2d00	0x2000	HEAP	.noinit
	20	0x400e2d00	0x400f05e8	0xd8e8	DATA	.data
	21	0x400f05f0	0x400f0720	0x130	DATA	sw_isr_table
	22	0x400f0720	0x400f07b2	0x92	DATA	device_states
	23	0x400f07b4	0x400f0a54	0x2a0	DATA	pm_device_slots_area
	24	0x400f0a54	0x400f0a90	0x3c	DATA	log_mpsc_pbuf_area
	25	0x400f0a90	0x400f0a94	0x4	DATA	log_msg_ptr_area
	26	0x400f0a94	0x400f0aac	0x18	DATA	k_heap_area
	27	0x400f0aac	0x400f0abc	0x10	DATA	k_sem_area
	28	0xa00f0ac0	0xa00ffac0	0xf000	DATA	.cached
	29	0x40100000	0x401029d8	0x29d8	BSS	.bss
	30	0x40102a00	0x401d2a00	0xd0000	HEAP	.heap_mem
	31	0xa01d2a00	0xa01d3004	0x604	HEAP	.unused_ram_start_marker
	32	0x00020000	0x00020078	0x78	DATA	.module.boot
	33	0x00020078	0x000200f0	0x78	DATA	.module.main
	34	0x000200f0	0x00020750	0x660	DATA	.static_uuid_entries
	35	0x00020750	0x000207b0	0x60	DATA	.static_log_entries
	36	0x000207b0	0x000208e0	0x130	DATA	.fw_metadata
	
	
	
after:
 Found 52 sections, listing valid sections......
	No	Start		End		Size	Type	Name
	1	0xa1048000	0xa10484e4	0x4e4	TEXT	.imr
	2	0xa0030000	0xa003016a	0x16a	TEXT	.WindowVectors.text
	3	0xa0030180	0xa00301a9	0x29	TEXT	.Level2InterruptVector.text
	4	0xa00301c0	0xa00301e9	0x29	TEXT	.Level3InterruptVector.text
	5	0xa0030200	0xa0030229	0x29	TEXT	.DebugExceptionVector.text
	6	0xa00302c0	0xa00302e9	0x29	TEXT	.NMIExceptionVector.text
	7	0xa0030300	0xa0030303	0x3	TEXT	.KernelExceptionVector.text
	8	0xa0030340	0xa0030352	0x12	TEXT	.UserExceptionVector.text
	9	0xa00303c0	0xa00303c6	0x6	TEXT	.DoubleExceptionVector.text
	10	0xa0030400	0xa0087a18	0x57618	TEXT	.text
	11	0xa0088000	0xa00dfc04	0x57c04	DATA	.rodata
	12	0xa00dfc04	0xa00dff64	0x360	DATA	initlevel
	13	0xa00dff64	0xa00e0760	0x7fc	DATA	device_area
	14	0xa00e0760	0xa00e0778	0x18	DATA	k_p4wq_initparam_area
	15	0xa00e0780	0xa00e0a56	0x2d6	DATA	device_handles
	16	0xa00e0a58	0xa00e0c98	0x240	DATA	log_const_area
	17	0xa00e0c98	0xa00e0cb8	0x20	DATA	log_backend_area
	18	0xa00e0cb8	0xa00e0d24	0x6c	DATA	.fw_ready
	19	0x400e0d40	0x400e2d40	0x2000	HEAP	.noinit
	20	0x400e2d40	0x400f0698	0xd958	DATA	.data
	21	0x400f06a0	0x400f07d0	0x130	DATA	sw_isr_table
	22	0x400f07d0	0x400f0862	0x92	DATA	device_states
	23	0x400f0864	0x400f0b04	0x2a0	DATA	pm_device_slots_area
	24	0x400f0b04	0x400f0b40	0x3c	DATA	log_mpsc_pbuf_area
	25	0x400f0b40	0x400f0b44	0x4	DATA	log_msg_ptr_area
	26	0x400f0b44	0x400f0b5c	0x18	DATA	k_heap_area
	27	0x400f0b5c	0x400f0b6c	0x10	DATA	k_sem_area
	28	0xa00f0b80	0xa00ffb80	0xf000	DATA	.cached
	29	0x40100000	0x401029d8	0x29d8	BSS	.bss
	30	0x40102a00	0x401d2a00	0xd0000	HEAP	.heap_mem
	31	0xa01d2a00	0xa01d3004	0x604	HEAP	.unused_ram_start_marker
	32	0x00020000	0x00020078	0x78	DATA	.module.boot
	33	0x00020078	0x000200f0	0x78	DATA	.module.main
	34	0x000200f0	0x00020750	0x660	DATA	.static_uuid_entries
	35	0x00020750	0x000207b0	0x60	DATA	.static_log_entries
	36	0x000207b0	0x000208e0	0x130	DATA	.fw_metadata

src/audio/audio_stream.c
```diff
@@ -0,0 +1,173 @@
...
+
+static const struct sink_ops audio_stream_sink_ops = {
+	.get_free_size = audio_stream_get_free_size,
+	.get_buffer = audio_stream_get_buffer,
+	.commit_buffer = audio_stream_commit_buffer,
+	.audio_set_ipc_params = audio_stream_set_ipc_params_sink,
+	.set_alignment_constants = audio_stream_sink_set_alignment_constants
+};
+
+inline void audio_stream_init(struct audio_stream __sparse_cache *audio_stream,
```
lyakh, line 155: exported `inline` isn't meaningful AFAIK

src/audio/module_adapter/module_adapter.c
```diff
@@ -23,6 +23,19 @@
...
+ * Needed till all the modules use PROCESSING_MODE_SINK_SOURCE
+ */
+#define IS_PROCESSING_MODE_AUDIO_STREAM(mod) \
+		(!!((struct module_data *)&(mod)->priv)->ops->process_audio_stream)
+
+#define IS_PROCESSING_MODE_RAW_DATA(mod) \
+		(!!((struct module_data *)&(mod)->priv)->ops->process_raw_data)
+
+#define IS_PROCESSING_MODE_SINK_SOURCE(mod) \
+		(!!((struct module_data *)&(mod)->priv)->ops->process)
```
lyakh, line 37: sorry, I'm a bit confused. An earlier patch in this PR added sink and source APIs to the "audio stream API" (i.e. `struct audio_stream`). That means that all module adapter users, setting the `simple_copy` flag can now already directly be switched to using the sink / source API. So you shouldn't need separate `module_adapter_audio_stream_type_copy()` and `IS_PROCESSING_MODE_AUDIO_STREAM()`?

kv2019i:
I marked this as P1. Let's try to get this to mergeable condition so we can take this in early in SOF2.7 cycle and have good runway for testing.

src/audio/module_adapter/module/generic.c
```diff
@@ -225,12 +228,13 @@ int module_prepare(struct processing_module *mod)
 	return ret;
 }
 
-int module_process(struct processing_module *mod, struct input_stream_buffer *input_buffers,
-		   int num_input_buffers, struct output_stream_buffer *output_buffers,
-		   int num_output_buffers)
+int module_process_stream_buffer(struct processing_module *mod,
```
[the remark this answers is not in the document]
marcinszkudlinski: I decode methods that use input_stream_buffer and output_stream_buffer here
Inside of those there are 2 ways of processing - audio stream and raw data

**I don't like it at all** - those are structures from IADK and should never be used here. Anyway - backward compatibilty 

Anyway - you're right that it may sound too close to audio_stream or something
Changing to module_process_legacy

jxstelter (approved):

btian1:
> @btian1 size of .text is 1988bytes (decimal) more size of .rodata is 84 bytes more
> 
> before: No Start End Size Type Name 1 0xa1048000 0xa10484e4 0x4e4 TEXT .imr 2 0xa0030000 0xa003016a 0x16a TEXT .WindowVectors.text 3 0xa0030180 0xa00301a9 0x29 TEXT .Level2InterruptVector.text 4 0xa00301c0 0xa00301e9 0x29 TEXT .Level3InterruptVector.text 5 0xa0030200 0xa0030229 0x29 TEXT .DebugExceptionVector.text 6 0xa00302c0 0xa00302e9 0x29 TEXT .NMIExceptionVector.text 7 0xa0030300 0xa0030303 0x3 TEXT .KernelExceptionVector.text 8 0xa0030340 0xa0030352 0x12 TEXT .UserExceptionVector.text 9 0xa00303c0 0xa00303c6 0x6 TEXT .DoubleExceptionVector.text 10 0xa0030400 0xa0087254 0x56e54 TEXT .text 11 0xa0088000 0xa00dfbb0 0x57bb0 DATA .rodata 12 0xa00dfbb0 0xa00dff10 0x360 DATA initlevel 13 0xa00dff10 0xa00e070c 0x7fc DATA device_area 14 0xa00e070c 0xa00e0724 0x18 DATA k_p4wq_initparam_area 15 0xa00e0730 0xa00e0a06 0x2d6 DATA device_handles 16 0xa00e0a08 0xa00e0c48 0x240 DATA log_const_area 17 0xa00e0c48 0xa00e0c68 0x20 DATA log_backend_area 18 0xa00e0c68 0xa00e0cd4 0x6c DATA .fw_ready 19 0x400e0d00 0x400e2d00 0x2000 HEAP .noinit 20 0x400e2d00 0x400f05e8 0xd8e8 DATA .data 21 0x400f05f0 0x400f0720 0x130 DATA sw_isr_table 22 0x400f0720 0x400f07b2 0x92 DATA device_states 23 0x400f07b4 0x400f0a54 0x2a0 DATA pm_device_slots_area 24 0x400f0a54 0x400f0a90 0x3c DATA log_mpsc_pbuf_area 25 0x400f0a90 0x400f0a94 0x4 DATA log_msg_ptr_area 26 0x400f0a94 0x400f0aac 0x18 DATA k_heap_area 27 0x400f0aac 0x400f0abc 0x10 DATA k_sem_area 28 0xa00f0ac0 0xa00ffac0 0xf000 DATA .cached 29 0x40100000 0x401029d8 0x29d8 BSS .bss 30 0x40102a00 0x401d2a00 0xd0000 HEAP .heap_mem 31 0xa01d2a00 0xa01d3004 0x604 HEAP .unused_ram_start_marker 32 0x00020000 0x00020078 0x78 DATA .module.boot 33 0x00020078 0x000200f0 0x78 DATA .module.main 34 0x000200f0 0x00020750 0x660 DATA .static_uuid_entries 35 0x00020750 0x000207b0 0x60 DATA .static_log_entries 36 0x000207b0 0x000208e0 0x130 DATA .fw_metadata
> 
> after: Found 52 sections, listing valid sections...... No Start End Size Type Name 1 0xa1048000 0xa10484e4 0x4e4 TEXT .imr 2 0xa0030000 0xa003016a 0x16a TEXT .WindowVectors.text 3 0xa0030180 0xa00301a9 0x29 TEXT .Level2InterruptVector.text 4 0xa00301c0 0xa00301e9 0x29 TEXT .Level3InterruptVector.text 5 0xa0030200 0xa0030229 0x29 TEXT .DebugExceptionVector.text 6 0xa00302c0 0xa00302e9 0x29 TEXT .NMIExceptionVector.text 7 0xa0030300 0xa0030303 0x3 TEXT .KernelExceptionVector.text 8 0xa0030340 0xa0030352 0x12 TEXT .UserExceptionVector.text 9 0xa00303c0 0xa00303c6 0x6 TEXT .DoubleExceptionVector.text 10 0xa0030400 0xa0087a18 0x57618 TEXT .text 11 0xa0088000 0xa00dfc04 0x57c04 DATA .rodata 12 0xa00dfc04 0xa00dff64 0x360 DATA initlevel 13 0xa00dff64 0xa00e0760 0x7fc DATA device_area 14 0xa00e0760 0xa00e0778 0x18 DATA k_p4wq_initparam_area 15 0xa00e0780 0xa00e0a56 0x2d6 DATA device_handles 16 0xa00e0a58 0xa00e0c98 0x240 DATA log_const_area 17 0xa00e0c98 0xa00e0cb8 0x20 DATA log_backend_area 18 0xa00e0cb8 0xa00e0d24 0x6c DATA .fw_ready 19 0x400e0d40 0x400e2d40 0x2000 HEAP .noinit 20 0x400e2d40 0x400f0698 0xd958 DATA .data 21 0x400f06a0 0x400f07d0 0x130 DATA sw_isr_table 22 0x400f07d0 0x400f0862 0x92 DATA device_states 23 0x400f0864 0x400f0b04 0x2a0 DATA pm_device_slots_area 24 0x400f0b04 0x400f0b40 0x3c DATA log_mpsc_pbuf_area 25 0x400f0b40 0x400f0b44 0x4 DATA log_msg_ptr_area 26 0x400f0b44 0x400f0b5c 0x18 DATA k_heap_area 27 0x400f0b5c 0x400f0b6c 0x10 DATA k_sem_area 28 0xa00f0b80 0xa00ffb80 0xf000 DATA .cached 29 0x40100000 0x401029d8 0x29d8 BSS .bss 30 0x40102a00 0x401d2a00 0xd0000 HEAP .heap_mem 31 0xa01d2a00 0xa01d3004 0x604 HEAP .unused_ram_start_marker 32 0x00020000 0x00020078 0x78 DATA .module.boot 33 0x00020078 0x000200f0 0x78 DATA .module.main 34 0x000200f0 0x00020750 0x660 DATA .static_uuid_entries 35 0x00020750 0x000207b0 0x60 DATA .static_log_entries 36 0x000207b0 0x000208e0 0x130 DATA .fw_metadata

only 2k bytes adding, looks great.

btian1 (approved):

src/include/sof/audio/audio_stream.h
```diff
@@ -30,6 +30,34 @@
...
+struct sof_audio_stream_params {
+	enum sof_ipc_frame frame_fmt;	/**< Sample data format */
+	enum sof_ipc_frame valid_sample_fmt;
+
+	uint32_t rate;		/**< Number of data frames per second [Hz] */
+	uint16_t channels;	/**< Number of samples in each frame */
+
+	/** alignment limit of stream copy, this value indicates how many
+	 * integer frames which can meet both byte align and frame align
+	 * requirement. it should be set in component prepare or param functions
```
plbossart, line 51: this comment is very very unclear. What does 'byte align' mean? What does 'frame align' mean? 

I know the text is the same as before but I still can't figure out what the intent of this flag is.

Consider rewording or clarifying this field.

src/audio/source_api_helper.c
```diff
@@ -0,0 +1,167 @@
...
+
+#include <sof/audio/source_api.h>
+#include <sof/audio/source_api_implementation.h>
+#include <sof/audio/audio_stream.h>
+
+void source_init(struct source __sparse_cache *source, const struct source_ops *ops,
+		 struct sof_audio_stream_params *audio_stream_params)
+{
+	source->ops = ops;
+	source->locked_read_frag_size = 0;
```
plbossart, line 14: usually when a function or a member uses the term 'locked' there's a lock involved?

This is confusion to me, what part is locked here? This is worthy of a comment, you can't just mention a lock and not explain what the design is.

src/include/sof/audio/source_api.h
```diff
@@ -0,0 +1,180 @@
...
+ * implemented and used as a mid-layer for modules needing it.
+ *
+ * NOTE: the module should get a complete portion of data it needs for processing, process it
+ *       than release. The reason is - the depending on the implementation, the calls may be
+ *       expensive - may involve some data moving in memory, cache writebacks, etc.
+ *
+ * Multicore:
+ *  single source is intended to be used by a single component only, so all calls to it
+ *  may safely be performed using a cached pointer alias. All data modified by source helper are
+ *  local and specific to a single source only
```
plbossart, line 49: this is very very unclear. What does single source mean really? as compared to 'multi sources'?
And no word on the 'locked attribute' ?

src/include/sof/audio/sink_api_implementation.h
```diff
@@ -0,0 +1,91 @@
...
+	 */
+	int (*set_alignment_constants)(struct sink __sparse_cache *sink,
+				       const uint32_t byte_align,
+				       const uint32_t frame_align_req);
+};
+
+/** internals of sink API. NOT TO BE MODIFIED OUTSIDE OF sink_api_helper.h */
+struct sink {
+	const struct sink_ops *ops;	/** operations interface */
+	size_t locked_write_frag_size; /** keeps number of bytes locked by get_buffer() */
```
plbossart, line 73: how does the locking work?

src/audio/source_api_helper.c
```diff
@@ -0,0 +1,167 @@
...
+//
+// Copyright(c) 2023 Intel Corporation. All rights reserved.
+//
+
+#include <sof/audio/source_api.h>
+#include <sof/audio/source_api_implementation.h>
+#include <sof/audio/audio_stream.h>
+
+void source_init(struct source __sparse_cache *source, const struct source_ops *ops,
+		 struct sof_audio_stream_params *audio_stream_params)
```
lyakh, line 11: maybe annotate `audio_stream_params` with `__sparse_cache` too instead of casting it out? Same for `sink_init()`

plbossart (changes requested):
The intent of this PR sounds good.

I am still going to request changes for
a) naming. using just 'struct sink' is madness, you need to prefix this API to avoid conflicts
b) locking is absolutely unclear and undocumented.

src/audio/source_api_helper.c
marcinszkudlinski: no, it cannot be sparse. It is up to the implementation how to ensure coherency - This pointer may be uncached - if the queue is shared between cores like dp_queue

src/include/sof/audio/audio_stream.h
marcinszkudlinski: This is NOT my comment, I just moved frame_align from buffer.h to audio_stream.h
I don't want to "fix" audio_stream in this PR, even if in comments

src/audio/copier/copier_generic.c
```diff
@@ -238,7 +237,8 @@ int create_endpoint_buffer(struct comp_dev *parent_dev,
 	audio_stream_set_rate(&buffer_c->stream, copier_cfg->base.audio_fmt.sampling_frequency);
 	audio_stream_set_frm_fmt(&buffer_c->stream, config->frame_fmt);
 	audio_stream_set_valid_fmt(&buffer_c->stream, valid_fmt);
-	buffer_c->buffer_fmt = copier_cfg->base.audio_fmt.interleaving_style;
+	audio_stream_set_buffer_fmt(&buffer_c->stream,
+				    copier_cfg->base.audio_fmt.interleaving_style);
```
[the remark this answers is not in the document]
marcinszkudlinski: again - not my code, just API change
This PR is not intended to fix copier

src/audio/source_api_helper.c
marcinszkudlinski: there's no locking at this level. Name indeed misleading, changed

src/include/sof/audio/audio_stream.h
plbossart: @andrula-song can you please take the AR to fix those comments? 
they are confusing and make little sense.
Adding @singalsu as reviewer :-)

src/audio/copier/copier_generic.c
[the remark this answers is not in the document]
plbossart: So, I beg to disagree. YOU are adding a new set_buffer_fmt L153 which only sets the interleaving style.

Naming matters and we'd miss an opportunity to clean-up the APIs and later their implementation.

src/audio/source_api_helper.c
```diff
@@ -0,0 +1,167 @@
...
+
+int source_set_params(struct sof_source __sparse_cache *source,
+		      struct sof_ipc_stream_params *params, bool force_update)
+{
+	if (source->ops->audio_set_ipc_params)
+		return source->ops->audio_set_ipc_params(source, params, force_update);
+	return 0;
+}
+
+int source_set_alignment_constants(struct sof_source __sparse_cache *source,
```
plbossart, line 160: I would also prefix with sof_ all the functions, just like you did for the struct sof_source. It'd be more future- and conflict-proof.

src/audio/sink_api_helper.c
```diff
@@ -0,0 +1,177 @@
...
+{
+	return sink->audio_stream_params->buffer_fmt;
+}
+
+bool sink_get_overrun(struct sof_sink __sparse_cache *sink)
+{
+	return sink->audio_stream_params->overrun_permitted;
+}
+
+int sink_set_frm_fmt(struct sof_sink __sparse_cache *sink, enum sof_ipc_frame frame_fmt)
```
plbossart, line 99: set_frame_fmt?

frm_fmt looks like a typo...

src/audio/sink_api_helper.c
```diff
@@ -0,0 +1,177 @@
...
+
+int sink_set_params(struct sof_sink __sparse_cache *sink,
+		    struct sof_ipc_stream_params *params, bool force_update)
+{
+	if (sink->ops->audio_set_ipc_params)
+		return sink->ops->audio_set_ipc_params(sink, params, force_update);
+	return 0;
+}
+
+int sink_set_alignment_constants(struct sof_sink __sparse_cache *sink,
```
plbossart, line 170: prefix functions with sof_ ?

src/include/sof/audio/sink_api.h
```diff
@@ -0,0 +1,163 @@
...
+ * - /dev/null
+ *     all the data stored in sink buffer are just simply discarded
+ * - I2S sender
+ *     Data stored in sink buffer will be sent to the external world
+ * - a memory ring buffer
+ *     data stored in the buffer will be sent to another module (usually using source API, but it
+ *     does not matter in fact).
+ *
+ *  The main advantage of using sink API instead of just taking pointers to the buffers is that
+ *  the buffer may be prepared at the moment the data producer is requesting it. i.e.
```
plbossart, line 35: 'may be prepared' is ambiguous, it could mean

a) may be in prepared state ('prepared' is an adjective in your sentence)
b) may be transitioned to prepared state ('prepared is a verb in your sentence)

Which is it?

src/include/sof/audio/sink_api_implementation.h
```diff
@@ -0,0 +1,91 @@
...
+	int (*set_alignment_constants)(struct sof_sink __sparse_cache *sink,
+				       const uint32_t byte_align,
+				       const uint32_t frame_align_req);
+};
+
+/** internals of sink API. NOT TO BE MODIFIED OUTSIDE OF sink_api_helper.h */
+struct sof_sink {
+	const struct sink_ops *ops;	/** operations interface */
+	size_t requested_write_frag_size; /** keeps number of bytes requested by get_buffer() */
+	size_t num_of_bytes_processed; /** processed bytes counter */
```
plbossart, line 74: since the start?

src/include/sof/audio/source_api_implementation.h
```diff
@@ -0,0 +1,91 @@
...
+	int (*set_alignment_constants)(struct sof_source __sparse_cache *source,
+				       const uint32_t byte_align,
+				       const uint32_t frame_align_req);
+};
+
+/** internals of source API. NOT TO BE MODIFIED OUTSIDE OF source_api_helper.h */
+struct sof_source {
+	const struct source_ops *ops;
+	size_t requested_read_frag_size;	/** keeps size of data obtained by get_data() */
+	size_t num_of_bytes_processed;	/** processed bytes counter */
```
plbossart, line 74: since the start?

What happens if there are multiple readers from the same source (if that's allowed?)

src/audio/audio_stream.c
```diff
@@ -0,0 +1,175 @@
...
+void audio_stream_init(struct audio_stream __sparse_cache *audio_stream,
+		       void *buff_addr, uint32_t size)
+{
+	audio_stream->size = size;
+	audio_stream->addr = buff_addr;
+	audio_stream->end_addr = (char *)audio_stream->addr + size;
+
+	/* set the default alignment info.
+	 * set byte_align as 1 means no alignment limit on byte.
+	 * set frame_align as 1 means no alignment limit on frame.
```
plbossart, line 164: no idea what this means either.

src/include/sof/audio/module_adapter/module/module_interface.h
```diff
@@ -81,15 +83,57 @@ struct module_interface {
...
+	 * It can be used by modules that support 1:1, 1:N, N:1 sources:sinks configuration.
+	 */
+	int (*process_audio_stream)(struct processing_module *mod,
+				    struct input_stream_buffer *input_buffers,
+				    int num_input_buffers,
+				    struct output_stream_buffer *output_buffers,
+				    int num_output_buffers);
+
+	/**
+	 * process_raw_data (depreciated)
```
plbossart, line 128: deprecated

src/include/sof/audio/module_adapter/module/module_interface.h
```diff
@@ -81,15 +83,57 @@ struct module_interface {
...
+	 * process
+	 *	- sources are handlers to source API struct source*[]
+	 *	- sinks are handlers to sink API struct sink*[]
+	 */
+	int (*process)(struct processing_module *mod,
+		       struct sof_source __sparse_cache **sources, int num_of_sources,
+		       struct sof_sink __sparse_cache **sinks, int num_of_sinks);
+
+	/**
+	 * process_audio_stream (depreciated)
```
plbossart, line 113: deprecated

src/include/sof/audio/module_adapter/module/module_interface.h
```diff
@@ -81,15 +83,57 @@ struct module_interface {
...
+	 *	    - sinks[].data is a pointer to raw audio data
 	 */
-	int (*process)(struct processing_module *mod, struct input_stream_buffer *input_buffers,
-		       int num_input_buffers, struct output_stream_buffer *output_buffers,
-		       int num_output_buffers);
+	int (*process_raw_data)(struct processing_module *mod,
+				struct input_stream_buffer *input_buffers,
+				int num_input_buffers,
+				struct output_stream_buffer *output_buffers,
+				int num_output_buffers);
```
plbossart, line 138: not fully understanding the concept of this PR, there was one process_ and now there are two process_, both deprecated. What's the direction then?

test/cmocka/src/audio/eq_fir/eq_fir_process.c
```diff
@@ -171,7 +171,7 @@ static int setup(void **state)
 	mod->stream_params->channels = params->channels;
 	mod->period_bytes = get_frame_bytes(params->source_format, params->channels) * 48000 / 1000;
 
-	ret = module_prepare(mod);
+	ret = module_prepare(mod, NULL, 0, NULL, 0);
```
plbossart, line 174: so an EQ has no sources and no sinks?

That's difficult to follow.... What is this supposed to represent?

plbossart (changes requested):
Some improvements since the last version, more suggested below.
I don't quite get the direction of travel with all these process_ functions and zero sinks/sources, the code ends-up quite confusing to the reader.

plbossart:
> Since I don't see any major issues left in the comments I am going to proceed with merge.
> 
> @plbossart the PR is hanging till May so I would like to kindly ask you next time to raise questions about implementation earlier and not after the PR gets approved and is ready to move on. We have already very bad statistics on how long SOF PRs are waiting in review and last minute comments with questions are not helping to improve the situation.

It doesn't make my comments go away. There were a couple of bad misses, such as not having a prefix for library functions is borderline really, and blaming the reviewer is extremely bad form in open source.

src/audio/audio_stream.c
andrula-song: there is a limited calculation frame number function audio_stream_avail_frames_aligned, it use shift to replace division, the byte_align and frame_align is for shift index calculation.

src/include/sof/audio/audio_stream.h
andrula-song: here is the link https://github.com/thesofproject/sof/pull/7833

src/audio/source_api_helper.c
```diff
@@ -0,0 +1,167 @@
...
+#include <sof/audio/source_api.h>
+#include <sof/audio/source_api_implementation.h>
+#include <sof/audio/audio_stream.h>
+
+void source_init(struct sof_source __sparse_cache *source, const struct source_ops *ops,
+		 struct sof_audio_stream_params *audio_stream_params)
+{
+	source->ops = ops;
+	source->requested_read_frag_size = 0;
+	source->audio_stream_params = audio_stream_params;
```
lyakh, line 15: @marcinszkudlinski in fact this is a bug. You call this function with an *acquired* object and you store a pointer to that acquired object permanently and then you use it outside of the object ownership scope. This will cause failures. Please fix in a follow up.

mwasko:
> It doesn't make my comments go away. There were a couple of bad misses, such as not having a prefix for library functions which is borderline really, and blaming the reviewer is extremely bad form in open source.
> 
My intention is not to ignore your comments, they can be addressed in a follow up PRs. The problem is 
situation where **new** review feedback is dropped weeks (sometimes months) after the PR has already been discussed and improved. I am aware that each one of us has their own work to do and response time may differ, but waiting with review for last minute has negative impact on SOF project velocity and usually is discouraging for code submitters. Of course, if there are some critical problems found in a late review then we should address them immediately but otherwise my recommendation would be to stage the improvements in a follow up PRs.

src/audio/source_api_helper.c
marcinszkudlinski: In case of audio_stream handlers to sink/sources are __sparse
```
static inline struct sof_sink __sparse_cache *
audio_stream_get_sink(struct audio_stream __sparse_cache *audio_stream)
{
    return &audio_stream->sink_api;
}

```

That implies that the handlers may be used when buffer is aquired, and only there
**This is enforced by module_adapter and there's no danger here**

The only concern is academic - we keep something that should not be used without lock. 
At the moment we may think about shadowing audio_stream_params  in sink/source structure as a temporary sollution (till comp_buffer/audio stream is still in use), this will cost some RAM. 

Next implementation (dp_queue) will keep audio_stream_params in shared, never cached memory - no shadowing needed. Other upcoming implementations - like shared LL buffer - will be used exclusively by single core (so may be safely always cached - no shadowing needed).

marc-hb:
Due to a complex sequence of events described in fix #7846 (please review), this PR surfaced a latent `assert()` override warning.

@marcinszkudlinski which toolchain do you typically use and how come it did not report this?

https://github.com/thesofproject/sof/actions/runs/5311578359/jobs/9614975368
```
[274/366] Building CXX object modules/sof/CMakeFiles/modules_sof.dir/D_/a/sof/sof/workspace/sof/src/audio/module_adapter/iadk/system_agent.cpp.obj
In file included from D:/a/sof/sof/workspace/sof/src/include/sof/common.h:30,
                 from D:/a/sof/sof/workspace/sof/src/include/sof/audio/sink_api.h:10,
                 from D:/a/sof/sof/workspace/sof/src/include/sof/audio/module_adapter/module/module_interface.h:17,
                 from D:/a/sof/sof/workspace/sof/src/include/sof/audio/module_adapter/iadk/iadk_module_adapter.h:19,
                 from D:/a/sof/sof/workspace/sof/src/audio/module_adapter/iadk/system_agent.cpp:22:
D:/a/sof/sof/workspace/sof/zephyr/include/rtos/panic.h:19: warning: "assert" redefined
   19 | #define assert(x) __ASSERT_NO_MSG(x)
      | 
In file included from D:/a/sof/sof/workspace/sof/src/audio/module_adapter/iadk/system_agent.cpp:16:
D:/a/sof/sof/workspace/sof/src/include/sof/audio/module_adapter/iadk/utilities/array.h:60: note: this is the location of the previous definition
   60 | #define assert(cond)
      | 
```

PR #7623, opened 2023-05-16
[DNM] pm: policy: d0i3: wait for ipc ack before pg

tmleman:
The DSP cannot enter the power gating state if it has not yet received an ack from host after sending an ipc message.


juimonen (commented):
not totally sure about the ipc flows, but looks ok to me.

marcinszkudlinski (approved):

west.yml
```diff
@@ -43,7 +43,7 @@ manifest:
 
     - name: zephyr
       repo-path: zephyr
-      revision: 9fc99928caf04c3ae78871f803ce8cee83541ae0
+      revision: 7d54586751cfe405d2cdcfd850f8100bc9844f63
```
kv2019i, line 46: Thanks for the detailed git commit message! I'm afraid you may need to udpate this now that we merged https://github.com/thesofproject/sof/pull/7642 but that PR was older and had the approvals plus tests pass, so I merged that first.

src/ipc/ipc-zephyr.c
```diff
@@ -184,6 +184,11 @@ void ipc_platform_send_msg_direct(const struct ipc_msg *msg)
 	intel_adsp_ipc_send_message_emergency(INTEL_ADSP_IPC_HOST_DEV, hdr->pri, hdr->ext);
 }
 
+int ipc_platform_poll_is_host_ready(void)
+{
+	return intel_adsp_ipc_is_complete(INTEL_ADSP_IPC_HOST_DEV);
+}
```
kv2019i, line 190: I'm too late to the review of the Zephyr side change, so I'm not going to block this here, but, but .... this does feel like a very common problem and problem that is typically solved without polling. I.e.
  - get PM resource (prevent power gaing)
  - send message
  - wait for ack
  - when ack received, release PM resource

But yeah, this kind of implements the same thing with a SOF side API, so this is probably good addition at this point.

kv2019i (approved):

west.yml
tmleman: No problem, I think that it's better if we do zephyr update in smaller jumps.

kv2019i:
Mandatory checks pass, known fails in https://sof-ci.01.org/sofpr/PR7623/build8445/devicetest/index.html and https://sof-ci.01.org/sofpr/PR7623/build8446/devicetest/index.html

PR #7626, opened 2023-05-16
copier: pipeline period recalculation

tmleman:
In case of a deep buffering pipelines don't require scheduling on every millisecond. We can calculate period based on the gateway buffer size.

At pipeline creation we assume that it can work on longer periods, period value is set to zero to be later adjust base on the pipeline components.

### DIFF
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -40,6 +40,7 @@
 #include <sof/audio/dai_copier.h>
 #include <sof/audio/ipcgtw_copier.h>
 #include <sof/audio/module_adapter/module/generic.h>
+#include <sof/schedule/ll_schedule_domain.h>
 
 #if CONFIG_ZEPHYR_NATIVE_DRIVERS
 #include <zephyr/drivers/dai.h>
@@ -92,6 +93,41 @@ static int copier_init(struct processing_module *mod)
 	}
 
 	dev->pipeline = ipc_pipe->pipeline;
+	/* Calculation of the period in which the component should be scheduled based on
+	 * the input buffer size.
+	 */
+	if (copier->gtw_cfg.dma_buffer_size > 0 && dev->pipeline->deep_buffering) {
+		/* Size of the single frame in bytes. */
+		uint32_t frame_size = copier->base.audio_fmt.channels_count *
+				       (copier->base.audio_fmt.depth >> 3);
+		/* Size of the one second of audio data in bytes. */
+		uint32_t one_s = frame_size * copier->base.audio_fmt.sampling_frequency;
+		/* Size of the ten millisecond of audio data in bytes. */
+		uint32_t ten_ms = DIV_ROUND_UP(one_s, 100);
+		/* The number of ten milliseconds chunks of data that will fit into the buffer. */
+		uint32_t chunk_count = copier->gtw_cfg.dma_buffer_size / ten_ms;
+
+		/* If buffer can fit at more than ten milliseconds of data we can try to schedule
+		 * pipe on periods bigger than one millisecond.
+		 */
+		if (chunk_count) {
+			dev->deep_buffering = true;
+			if (chunk_count == 1)
+				dev->period = (5 * LL_TIMER_PERIOD_US);
+			else
+				dev->period = (10 * LL_TIMER_PERIOD_US);
+
+			/* If a component that may run on bigger periods has already been added to
+			 * the pipeline, the pipe must run at the lowest possible value.
+			 */
+			if (!dev->pipeline->period || dev->pipeline->period < dev->period) {
+				dev->pipeline->period = dev->period;
+				comp_cl_warn(&comp_copier,
+					     "changing the period for the pipe (new value %u)",
+					     dev->pipeline->period);
+			}
+		}
+	}
 
 	node_id = copier->gtw_cfg.node_id;
 	/* copier is linked to gateway */
--- a/src/include/sof/audio/component.h
+++ b/src/include/sof/audio/component.h
@@ -591,6 +591,9 @@ struct comp_dev {
 	bool is_shared;		/**< indicates whether component is shared
 				  *  across cores
 				  */
+	bool deep_buffering;	/**< indicates whether component is able to work
+				  *  deep buffering
+				  */
 	struct comp_ipc_config ipc_config;	/**< Component IPC configuration */
 	struct tr_ctx tctx;	/**< trace settings */
 
--- a/src/ipc/ipc4/helper.c
+++ b/src/ipc/ipc4/helper.c
@@ -18,7 +18,6 @@
 #include <sof/lib/mailbox.h>
 #include <sof/list.h>
 #include <sof/platform.h>
-#include <sof/schedule/ll_schedule_domain.h>
 #include <rtos/wait.h>
 
 /* TODO: Remove platform-specific code, see https://github.com/thesofproject/sof/issues/7549 */
@@ -206,7 +205,8 @@ static int ipc4_create_pipeline(struct ipc4_pipeline_create *pipe_desc)
 	}
 
 	pipe->time_domain = SOF_TIME_DOMAIN_TIMER;
-	pipe->period = LL_TIMER_PERIOD_US;
+	pipe->period = 0;
+	pipe->deep_buffering = true;
 
 	/* sched_id is set in FW so initialize it to a invalid value */
 	pipe->sched_id = 0xFFFFFFFF;
@@ -868,6 +868,23 @@ int ipc4_add_comp_dev(struct comp_dev *dev)
 	tr_dbg(&ipc_tr, "ipc4_add_comp_dev add comp %x", icd->id);
 	/* add new component to the list */
 	list_item_append(&icd->list, &ipc->comp_list);
+
+	/* If new component added to the pipeline is not fitted for deep buffering we need to
+	 * inform pipeline it cannot work on long periods.
+	 */
+	if (!dev->deep_buffering) {
+		if (!dev->pipeline) {
+			const uint32_t pipe_id = dev->ipc_config.pipeline_id;
+			struct ipc_comp_dev *ipc_comp =
+				ipc_get_comp_by_ppl_id(ipc,
+						       COMP_TYPE_PIPELINE,
+						       pipe_id);
+
+			dev->pipeline = ipc_comp->pipeline;
+		}
+
+		dev->pipeline->deep_buffering = false;
+	}
 
 	return IPC4_SUCCESS;
 };
--- a/src/audio/pipeline/pipeline-graph.c
+++ b/src/audio/pipeline/pipeline-graph.c
@@ -14,6 +14,7 @@
 #include <sof/lib/uuid.h>
 #include <sof/compiler_attributes.h>
 #include <sof/list.h>
+#include <sof/schedule/ll_schedule_domain.h>
 #include <rtos/spinlock.h>
 #include <rtos/string.h>
 #include <rtos/clk.h>
@@ -301,6 +302,14 @@ int pipeline_complete(struct pipeline *p, struct comp_dev *source,
 
 	data.start = source;
 	data.p = p;
+
+	/* If a component has been added to the pipeline that cannot run on longer periods
+	 * we need to set pipe period to one ms.
+	 */
+	if (!p->deep_buffering) {
+		p->period = LL_TIMER_PERIOD_US;
+		pipe_info(p, "setting pipe period to %llu us", LL_TIMER_PERIOD_US);
+	}
 
 	/* now walk downstream from source component and
 	 * complete component task and pipeline initialization
--- a/src/include/sof/audio/pipeline.h
+++ b/src/include/sof/audio/pipeline.h
@@ -94,6 +94,8 @@ struct pipeline {
 		bool aborted;		/* STOP or PAUSE failed, stay active */
 		bool pending;		/* trigger scheduled but not executed yet */
 	} trigger;
+	/* pipe can use long periods for scheduling in case of a deep buffering */
+	bool deep_buffering;
 };
 
 struct pipeline_walk_context {


src/audio/copier/copier.c
```diff
@@ -740,6 +740,29 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
 	}
 
 	dev->pipeline = ipc_pipe->pipeline;
+	/* Calculation of the period in which the pipeline should be scheduled based on
+	 * the input buffer size.
+	 */
+	if (copier->gtw_cfg.dma_buffer_size > 0) {
+		/* Size of the single sample in bytes. */
+		uint32_t sample_size = copier->base.audio_fmt.channels_count *
+				       (copier->base.audio_fmt.depth / 8);
```
ranj063, line 749: doesnt this depend on direction? And maybe even restrict this to the host gateway case?

src/audio/copier/copier.c
```diff
@@ -740,6 +740,29 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
...
+
+
+		__ASSERT_NO_MSG(ms_count > 0);
+		if (ms_count == 0)
+			ms_count = 1;
+
+		/* By default, the period was set to one millisecond, so now just multiply it
+		 * to get a new value.
+		 */
+		dev->pipeline->period = dev->pipeline->period * ms_count;
```
ranj063, line 763: what about the buffer sizes between modules? All the intermediate buffers are set up to handle only period size data right? Wouldnt this end up causing failures?

mmaka1, line 763: What if there are two copiers, each one attached to a gateway? The second one overrides the period no matter what are the constraints of the first one?

src/audio/copier/copier.c
```diff
@@ -740,6 +740,29 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
...
 	dev->pipeline = ipc_pipe->pipeline;
+	/* Calculation of the period in which the pipeline should be scheduled based on
+	 * the input buffer size.
+	 */
+	if (copier->gtw_cfg.dma_buffer_size > 0) {
+		/* Size of the single sample in bytes. */
+		uint32_t sample_size = copier->base.audio_fmt.channels_count *
+				       (copier->base.audio_fmt.depth / 8);
+		/* Size of the one millisecond of audio data in bytes. */
+		uint32_t one_ms = (sample_size * copier->base.audio_fmt.sampling_frequency) / 1000;
```
btian1, line 751: possible input is 44100? maybe a switch is better in case of remainder happened

src/audio/copier/copier.c
```diff
@@ -740,6 +740,29 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
...
+		/* The number of milliseconds of data that will fit into the buffer. */
+		uint32_t ms_count = copier->gtw_cfg.dma_buffer_size / one_ms;
```
btian1, line 753: if there is reminder, then ms_count is not accurate, possible add this?

if (copier->gtw_cfg.dma_buffer_size % one_ms)
           ms_count++;

src/audio/copier/copier.c
```diff
@@ -740,6 +740,29 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
...
+
+
+		__ASSERT_NO_MSG(ms_count > 0);
+		if (ms_count == 0)
+			ms_count = 1;
```
btian1, line 758: put 757-758 ahead of 756?

src/audio/copier/copier.c
lyakh: if it's needed, then `SOF_DIV_ROUND_UP()` should be used

src/audio/copier/copier.c
lyakh: since `ms_count` is unsigned, then if it's == 0, you need only one - either an assertion failure or a default value. Also note, that this will behave differently depending on whether asserts are enabled or not.

lyakh: I think this is a wrong use of an assertion then. I think asserts are used if ever during debugging to make sure that our assumptions are correct. That's why they're disabled at run-time. Because once we've checked that no cosmic rays develop an unhealthy affinity to our code, we can forget about them. OTOH, if that can indeed happen at run-time, then an assert is useless there and we need proper error checking.

src/audio/copier/copier.c
```diff
@@ -100,6 +101,38 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
...
+	 * the input buffer size.
+	 */
+	if (copier->gtw_cfg.dma_buffer_size > 0) {
+		/* Size of the single sample in bytes. */
+		uint32_t sample_size = copier->base.audio_fmt.channels_count *
+				       (copier->base.audio_fmt.depth / 8);
+		/* Size of the one second of audio data in bytes. */
+		uint32_t one_s = sample_size * copier->base.audio_fmt.sampling_frequency;
+		/* Size of the one millisecond of audio data in bytes. */
+		uint32_t one_ms = SOF_DIV_ROUND_UP(one_s, 1000);
```
lyakh, line 114: this should be rounded down for a correct rounding up below

src/audio/copier/copier.c
```diff
@@ -100,6 +101,38 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
...
+		/* The number of milliseconds of data that will fit into the buffer. */
+		uint32_t ms_count = SOF_DIV_ROUND_UP(copier->gtw_cfg.dma_buffer_size, one_ms);
+
+		/* If buffer can feet at least four milliseconds of data we can try to schedule
```
lyakh, line 118: typo "fit"

src/audio/copier/copier.c
tmleman: No, right now we aim to use it on simple pipelines. Only when we have two connected gtw's.

src/audio/copier/copier.c
```diff
@@ -100,6 +101,41 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
...
+	if (copier->gtw_cfg.dma_buffer_size > 0 && dev->pipeline->long_period) {
+		/* Size of the single sample in bytes. */
+		uint32_t sample_size = copier->base.audio_fmt.channels_count *
+				       (copier->base.audio_fmt.depth / 8);
+		/* Size of the one second of audio data in bytes. */
+		uint32_t one_s = sample_size * copier->base.audio_fmt.sampling_frequency;
+		/* Size of the ten millisecond of audio data in bytes. */
+		uint32_t ten_ms = one_s / 100;
+		/* The number of ten milliseconds chunks of data that will fit into the buffer. */
+		uint32_t chunk_count = copier->gtw_cfg.dma_buffer_size / ten_ms;
```
lyakh, line 116: just as an example to clarify why this won't quite calculate correct results and why you need to round up line 114:
```
one_s = 8090
ten_ms = 80
dma_buffer_size = 8000
chunk_count = 100
```
that will tell you, that you can fit 100 10ms long chunks in the buffer, so you could conclude that a full second of audio fits into the buffer, whereas it doesn't - 90 bytes don't fit. Whereas if you round up the first division in line 114, you get
```
one_s = 8090
ten_ms = 81
dma_buffer_size = 8000
chunk_count = 98
```
so you conclude, that 980ms of audio will fit into the buffer, which is correct - they will fit there and will occupy 0.98*8090 < 7983 < 8000 bytes.

src/audio/copier/copier.c
```diff
@@ -100,6 +101,41 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
...
+
+			/* If a component that may run on bigger periods has already been added to
+			 * the pipeline, the pipe must run at the lowest possible value.
+			 */
+			if (dev->pipeline->period != 0) {
+				if (dev->pipeline->period < dev->period)
+					dev->pipeline->period = dev->period;
+			} else {
+				dev->pipeline->period = dev->period;
+			}
```
lyakh, line 136: you wanted to write:
```
if (!dev->pipeline->period || dev->pipeline->period < dev->period)
	dev->pipeline->period = dev->period;
```

src/ipc/ipc4/helper.c
```diff
@@ -787,6 +787,23 @@ int ipc4_add_comp_dev(struct comp_dev *dev)
...
+			const uint32_t pipe_id = dev->ipc_config.pipeline_id;
+			struct ipc_comp_dev *ipc_comp =
+				ipc_get_comp_by_ppl_id(ipc,
+						       COMP_TYPE_PIPELINE,
+						       pipe_id);
+
+			dev->pipeline = ipc_comp->pipeline;
+		}
+
+		dev->pipeline->long_period = false;
```
kv2019i, line 804: I was thinking it would be ncie to have a log entry to catch if topology has an incorrect configuration. But as this is called also in normal case, addign a INF level log is probably not a good idea here. It would be good to have a note/warning if host is asking for a combiation with copier configured with deep buffers for DMA, but the pipeline is not compatible.

src/audio/copier/copier.c
```diff
@@ -100,6 +101,41 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
...
+		/* If buffer can fit at more than ten milliseconds of data we can try to schedule
+		 * pipe on periods bigger than one millisecond.
+		 */
+		if (chunk_count) {
+			dev->deep_buffering = true;
+			if (chunk_count == 1)
+				dev->period = (5 * LL_TIMER_PERIOD_US);
+			else
+				dev->period = (10 * LL_TIMER_PERIOD_US);
+
```
kv2019i, line 127: With so many inputs (both from runtime PCM config as well as topology), I think this warrants printing out the modified period size to FW log.

src/audio/copier/copier.c
```diff
@@ -100,6 +101,41 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
 	}
 
 	dev->pipeline = ipc_pipe->pipeline;
+	/* Calculation of the period in which the component should be scheduled based on
+	 * the input buffer size.
+	 */
+	if (copier->gtw_cfg.dma_buffer_size > 0 && dev->pipeline->long_period) {
+		/* Size of the single sample in bytes. */
+		uint32_t sample_size = copier->base.audio_fmt.channels_count *
+				       (copier->base.audio_fmt.depth / 8);
```
btian1, line 110: "/ 8" --> ">> 3"

src/audio/copier/copier.c
```diff
@@ -100,6 +101,41 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
 	}
 
 	dev->pipeline = ipc_pipe->pipeline;
+	/* Calculation of the period in which the component should be scheduled based on
+	 * the input buffer size.
+	 */
+	if (copier->gtw_cfg.dma_buffer_size > 0 && dev->pipeline->long_period) {
+		/* Size of the single sample in bytes. */
+		uint32_t sample_size = copier->base.audio_fmt.channels_count *
```
btian1, line 109: sample_size --> frame_size?

src/audio/copier/copier.c
btian1: may need enumerate all the possible freqs and then divide can be avoided by look up table.
like is only support 44100 and 48000 input, then two divide can be avoided.

src/audio/copier/copier.c
```diff
@@ -100,6 +101,41 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
...
+		/* Size of the ten millisecond of audio data in bytes. */
+		uint32_t ten_ms = one_s / 100;
+		/* The number of ten milliseconds chunks of data that will fit into the buffer. */
+		uint32_t chunk_count = copier->gtw_cfg.dma_buffer_size / ten_ms;
+
+		/* If buffer can fit at more than ten milliseconds of data we can try to schedule
+		 * pipe on periods bigger than one millisecond.
+		 */
+		if (chunk_count) {
+			dev->deep_buffering = true;
```
btian1, line 122: dev->pipeline->long_period same meaning with dev->deep_buffering

src/include/sof/audio/component.h
```diff
@@ -586,6 +586,9 @@ struct comp_dev {
 	bool is_shared;		/**< indicates whether component is shared
 				  *  across cores
 				  */
+	bool deep_buffering;	/**< indicates whether component is able to work
```
btian1, line 589: difference with long_period?

lgirdwood (commented):
@tmleman This should be a pipeline 2.0 API i.e.
```c
int pipeline_sleep(pipeline, ms);
```
And this could be called by copiers during process() when conditions allow the pipe to skip schedling ticks. It should also be a feature that can be enabled/disabled by topology flag (i. assume there is an IPC4 flag to enable it ?)

src/audio/copier/copier.c
tmleman: I assumed that rounding is not needed because none of the formats we use will result in a value that is not divisible by 100. I added rounding anyway.

src/include/sof/audio/component.h
tmleman: No, but I wanted them to be easily distinguishable in the code.

src/audio/copier/copier.c
lgirdwood: Its best to use the >> as not all compilers will do the right thing. e.g. could be optimal on xtensa, but on ARM it could divide using the same CC flags.

lgirdwood (commented):
@tmleman @ranj063 I assume we have a CI test or topology that can test 10ms pipelines ? If so, we should make sure its run since we are optimizing non 1ms pipelines now.

src/include/sof/audio/component.h
btian1: unless code can't referenced, if can keep one flag, I would suggest keep one flag.

lgirdwood:
@tmleman sorry to ask, are you able to rebase again on HEAD as a build failure was introduced with VMH and was fixed late yesterday. Thanks !

tmleman:
23ww31.4 Update(2/2): rebase

src/include/sof/audio/component.h
btian1: seems all modified to deep_buffering, what I mean is: possible only use one bool, instead of use two.

btian1 (commented):
@tmleman , I would suggest split this PR to two patches if you need use two deep_buffering, one is for dev->deep_buffering adding, another is for pipeline->deep_buffering adding, this will make the PR more easy to read and understand.

src/include/sof/audio/component.h
tmleman: I misunderstood original comment. I don't see way to replace those two flags with one right now.

src/audio/copier/copier.c
RanderWang: so we will not support 11.025 khz and 22.05 khz ? 
````c
enum ipc4_sampling_frequency {
	IPC4_FS_8000HZ = 8000,
	IPC4_FS_11025HZ = 11025,
	IPC4_FS_12000HZ = 12000, /**< Mp3, AAC, SRC only */
	IPC4_FS_16000HZ = 16000,
	IPC4_FS_18900HZ = 18900, /**< SRC only for 44100 */
	IPC4_FS_22050HZ = 22050,
	IPC4_FS_24000HZ = 24000, /**< Mp3, AAC, SRC only */
	IPC4_FS_32000HZ = 32000,
	IPC4_FS_37800HZ = 37800, /**< SRC only for 44100 */
	IPC4_FS_44100HZ = 44100,
	IPC4_FS_48000HZ = 48000, /**< Default */
	IPC4_FS_64000HZ = 64000, /**< AAC, SRC only */
	IPC4_FS_88200HZ = 88200, /**< AAC, SRC only */
	IPC4_FS_96000HZ = 96000, /**< AAC, SRC only */
	IPC4_FS_176400HZ = 176400, /**< SRC only */
	IPC4_FS_192000HZ = 192000, /**< SRC only */
	IPC4_FS_INVALID
};
````

**Comment by lgirdwood** [comment]
@tmleman does the latest update resolve the review comments ?

**Comment by tmleman** [comment]
@abonislawski @marcinszkudlinski request for a review.

File: src/ipc/ipc4/helper.c [hunk 16]
```diff
@@ -787,6 +787,23 @@ int ipc4_add_comp_dev(struct comp_dev *dev)
...
+	/* If new component added to the pipeline is not fitted for deep buffering we need to
+	 * inform pipeline it cannot work on long periods.
+	 */
+	if (!dev->deep_buffering) {
+		if (!dev->pipeline) {
+			const uint32_t pipe_id = dev->ipc_config.pipeline_id;
+			struct ipc_comp_dev *ipc_comp =
+				ipc_get_comp_by_ppl_id(ipc,
+						       COMP_TYPE_PIPELINE,
+						       pipe_id);
```
lyakh, line 799: are we sure, that `ipc_comp` cannot be `NULL` here? `pipe_id` comes from an IPC... BTW, looks like `ipc4_add_comp_dev()` can be `static`

lgirdwood:
> 23ww40.3 Update: rebase

@tmleman CI showing some issues. We probably want to rebase and retest to give confidence the calculations are correct.

tmleman:
I'm closing this PR. I don't have time to work on it at the moment.

PR #7627, opened 2023-05-16
Tools: Topology2: Add DRC to sof-hda-efx-generic topologies

singalsu:


singalsu:
Note: This can't be merged before #7621 .

cujomalainey:
Is the difference between 1 and 15 simply the buffer depth?

singalsu:
> Is the difference between 1 and 15 simply the buffer depth?

Yes, the deep buffer pipeline 15 is defined in https://github.com/thesofproject/sof/blob/main/tools/topology/topology2/platform/intel/deep-buffer.conf and the processing in it looks similar to 1. DEEPBUFFER_FW_DMA_MS sets the copier DMA size but I wonder if it currently has impact.

cujomalainey:
Is this topology targeted for use by google? if yes we can drop FIR component as we won't use it

singalsu:
> Is this topology targeted for use by google? if yes we can drop FIR component as we won't use it

The "EFX" topology includes all to enable our internal testing. The processing is by default bypassed at component level but I understand it's not sufficient for a product, there's still an overhead for MCPS and latency.

But we should make then a minimized pipeline without FIR for topologies for Chrome OS. Would Chrome also use these sof-hda-generic topologies or same pipeline macros like mixout-gain-efx-dai-copier-playback?

juimonen (approved):
"The 32 not format" probably typo in commit message.... otherwise as this is development topology - so lgtm

cujomalainey:
I'm not familiar enough with the `mixout-gain-efx-dai-copier-playback` layout to say yes or no, but it does look like something we are trying to move towards (assuming it is the depicted above) @johnylin76 FYI

johnylin76:
In the current use case for Google, DRC is applied in 3-band manner. That is, a 1-to-3 CROSSOVER connected with DRC on each band (3 DRCs in total). In SOF practice now we wrap them all into MULTIBAND_DRC module [source](https://github.com/thesofproject/sof/blob/main/src/audio/multiband_drc/multiband_drc.c) rather than topology.

To map to EFX I think CROSSOVER can be formed by IIR but DRC will have the difficulty. Although the current DRC implementation supports any channel count, only one parameter set can be configured.

kv2019i:
@singalsu Can you remove the "DNM" prefix from PR title when the dependencies are merged?

singalsu:
No changes, just rebase of patch to re-trigger CI. FW part of DRC is now merged. To use this topology update the rimage patch needs merge.

lgirdwood (approved):

lgirdwood:
@singalsu conflicts.

singalsu:
> @singalsu conflicts.

Thanks, fixed now (the crossover tplg2 changes were merged).

lgirdwood (approved):

PR #7631, opened 2023-05-16
Modify the naming convertion for the host copier in topology

ranj063:
The names for the host copier has been changed from the current convention of copier.host.<PIPELINE_ID>.<INSTANCE_ID> to host-copier.<PCM_ID>.<DIRECTION>


ranj063:
SOFCI TEST

tools/topology/topology2/include/components/host-copier.conf
```diff
@@ -0,0 +1,140 @@
...
+		]
+
+		#
+		# mandatory attributes that must be provided when the buffer class is instantiated
+		#
+		!mandatory [
+			"num_input_pins"
+			"num_output_pins"
+			"num_input_audio_formats"
+			"num_output_audio_formats"
```
plbossart, line 112: same here, can both be non zero?

tools/topology/topology2/platform/intel/dmic-wov.conf
```diff
@@ -108,15 +109,15 @@ Object.Pipeline {
 Object.Base.route [
 	{
 		source $DMIC_WOV_DAI_PIPELINE_KPB
-		sink $DMIC1_HOST_PIPELINE_SINK
+		sink "host-copier.11.capture"
```
plbossart, line 112: same, why not use the macro?

plbossart (commented):
I guess I am not fully getting the idea for the copier input/output pins, and I also wonder about replacing the macros with open-coding, see below @ranj063.

The replacement looks mostly good otherwise, I didn't spot any major issues.

tools/topology/topology2/platform/intel/dmic-wov.conf
ranj063: same argument here

tools/topology/topology2/platform/intel/bt-generic.conf
```diff
@@ -198,7 +199,7 @@ Object.Base.route [
 	}
 	{
 		source	$BT_CP_DAI_PIPELINE_SRC
-		sink	$BT_CP_HOST_PIPELINE_SINK
+		sink	"host-copier.$BT_PCM_ID.capture"
```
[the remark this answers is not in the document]
ranj063: @plbossart I dont want to use the macro because for 1 if the PCM ID changes, the macro will also have to change. Its not reasy to keep track of the macro based on another macro. This way, the route sink is tied only to the PCM ID only not on the PCM ID and another macro that uses it.

tools/topology/topology2/include/components/host-copier.conf
```diff
@@ -0,0 +1,140 @@
...
+			"pcm_id"
+			"direction"
+		]
+
+		#
+		# mandatory attributes that must be provided when the buffer class is instantiated
+		#
+		!mandatory [
+			"num_input_pins"
+			"num_output_pins"
```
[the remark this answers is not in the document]
ranj063: i've removed the default values and mandatory attributes for both. num_input_pins and num_output_pins come into the picture only where there're more than 1. Otherwise, they're just ignored by the kernel driver and in all our topologies today we dont have more than 1 of these for playback/capture.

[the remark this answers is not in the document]
ranj063: ANd I was wrong about this. So I've fixed all host-copiers to have the right number of input/output pins based on direction

tools/topology/topology2/cavs-mixin-mixout-efx-hda.conf
```diff
@@ -137,4 +138,8 @@ Object.Base.route [
 		source 'copier.HDA.4.1'
 		sink 'copier.host.3.1'
 	}
+	{
+		source 'host-copier.0.playback'
+		sink 'gain.1.1'
+	}
```
bardliao, line 145: Can we pass `pcm_id` to host-copier-gain-mixin-playback.conf, and set route there?

ranj063: @bardliao thats what I first considered. But moving the route from/to the host-copier makes it easier for me to passthrough pipelines later on.

tools/topology/topology2/include/components/host-copier.conf
plbossart: still not clear on what to do with eg. a host-copier for playback. What is the meaning of num_input_audio_formats here?

tools/topology/topology2/include/pipelines/cavs/host-copier-gain-mixin-playback.conf
```diff
@@ -45,12 +45,12 @@ Class.Pipeline."host-copier-gain-mixin-playback" {
...
 
 	Object.Widget {
-		copier."1" {
-			copier_type	"host"
+		host-copier."1" {
 			type	"aif_in"
 			node_type $HDA_HOST_OUTPUT_CLASS
 			num_input_audio_formats 3
 			num_output_audio_formats 3
+			num_output_pins 1
```
plbossart, line 53: I am lost on this one, so we have 3 input formats but no input pins?

tools/topology/topology2/include/pipelines/cavs/mixout-gain-host-copier-capture.conf
```diff
@@ -47,12 +47,12 @@ Class.Pipeline."mixout-gain-host-copier-capture" {
...
 		mixout."1" {}
 
-		copier."1" {
-			copier_type "host"
+		host-copier."1" {
 			type	"aif_out"
 			node_type $HDA_HOST_INPUT_CLASS
 			num_input_audio_formats 4
 			num_output_audio_formats 4
+			num_input_pins	1
```
plbossart, line 55: here as well

plbossart (commented):
Still confused on pin definitions...

tools/topology/topology2/include/components/host-copier.conf
plbossart: for the record: even if for playback there is no input pin we still need the input_audio_formats that are used at the DMA level. It's a layering violation in the design but it is what it is.

PR #7551, opened 2023-05-02
[stable-v2.2] cherry-pick docker and other upgrade from main branch

marc-hb:
95%, mostly clean cherry-picks from main branch

This avoids branching the Docker image and turns all checks green.

Issue #7614, opened 2023-05-15
Zephyr main branch requires 0.16.x SDK (Intel toolchain change)


marc-hb:
Since zephyr commit https://github.com/zephyrproject-rtos/zephyr/commit/9fc99928caf04c3ae78 and https://github.com/zephyrproject-rtos/zephyr/commit/49d0ad55206aa, the Zephyr main branch is not compatible with the Zephyr SDK 0.15.x anymore, it now requires 0.16.x. The Zephyr main branch now depends on different (and correct) toolchain names which are not present in 0.15.x

 -> this requires some "infrastructure" updates

Every PR trying to build the Zephyr main branch (a.k.a. "zmain") started to fail like this, example in https://github.com/thesofproject/sof/actions/runs/4984592530

```
CMake Error at /zep_workspace/zephyr/cmake/compiler/gcc/target.cmake:12 (message):
  C compiler
  /home/cwd_user/zephyr-sdk-0.15.2/xtensa-intel_ace15_mtpm_zephyr-elf/bin/xtensa-intel_ace15_mtpm_zephyr-elf-gcc
  not found - Please check your toolchain installation
```

- [ ] Fixing this for Linux compilation should be more or less a one-line change in the `zephyr/docker-run.sh` script (example in previous commit 11441d85409)

- [ ] Change ubuntu-20.04 -> ubuntu-22.04 in the 2 files `.github/workflows/*zephyr.yml` to stay in sync with https://github.com/zephyrproject-rtos/docker-image/pkgs/container/zephyr-build

- [ ] The Windows compilation that we've enjoyed thanks to @aborisovich's heroic work (commit df1ba222306d) is unfortunately a bit more complicated.

Note upgrading will also help with OpenSSL 1.x removal where this has already been discussed:
- https://github.com/thesofproject/rimage/pull/157

marc-hb:
P1 because this will block any Zephyr uprev past the commits listed above.

iuliana-prodan:
I know about this commit https://github.com/zephyrproject-rtos/hal_xtensa/commit/e636ce8fdf334addaae37fb2417dc2cb8bf7b50a which I reverted because it required an update in zephyr sdk - see [comment](https://github.com/zephyrproject-rtos/zephyr/pull/57084#issuecomment-1518092809)
[...]

PR #7632, opened 2023-05-17
[SKIP SOF-TEST] .github/codestyle: drop spurious *.yml pattern

marc-hb:
This suddenly failed in for PR #7551 in
https://github.com/thesofproject/sof/actions/runs/4998109337/jobs/8953192254

I have no idea why it never failed earlier (new yamllint version maybe?) but it's always been wrong and should have never been there anyway.

Fixes commit 3d69a7f69eb63e (".github: extend yamllint line-length to 100")

### DIFF
--- a/.github/workflows/codestyle.yml
+++ b/.github/workflows/codestyle.yml
@@ -64,4 +64,4 @@ jobs:
           yamllint_config: '{extends: default, rules: {line-length: {max: 100}}}'
         run: yamllint -f parsable
           -d "$yamllint_config"
-          --strict .github/workflows/*.yml *.yml
+          --strict .github/workflows/*.yml


marc-hb:
zmain failure is known and unrelated: #7614

dbaluta (approved):

paulstelian97:
The original commit says that it added that pattern specifically so west.yml is also covered. Do we not want that?

.github/workflows/codestyle.yml
```diff
@@ -64,4 +64,4 @@ jobs:
           yamllint_config: '{extends: default, rules: {line-length: {max: 100}}}'
         run: yamllint -f parsable
           -d "$yamllint_config"
-          --strict .github/workflows/*.yml *.yml
+          --strict .github/workflows/*.yml
```
kv2019i, line 67: Is it spurious, mianline has sof/west.yml (but 2.2 stable does not and thus the fail in 7551)?

kv2019i (changes requested):

marc-hb:
Thanks, I need more sleep!

PR #7634, opened 2023-05-17
topology2: add tgl rt712 topology

bardliao:
Add sof-tgl-rt712 support. rt712 is a multi function codec which shpports headset, amp, and dmic functions.


plbossart (approved):

ranj063:
all known failures in device tests. merging

PR #7640, opened 2023-05-17
Change the naming convention for the ALH DAI copiers in topology

ranj063:
For ALH DAI, the new name is changed to reflect the link ID and index in an aggregated group like: dai-copier-ALH.SDW1-Playback.0 or alh-copier.SDW1-Playback.1

Also, today we do  not show the aggregated DAI copiers in the topology graph. So add a route via a virtual widget to show its existence as below
sof-mtl-rt711-l0-rt1316-l23-rt714-l1

### DIFF
--- a/tools/topology/topology2/cavs-sdw-src-gain-mixin.conf
+++ b/tools/topology/topology2/cavs-sdw-src-gain-mixin.conf
@@ -11,7 +11,7 @@
 <tokens.conf>
 <virtual.conf>
 <src-gain-mixin-playback.conf>
-<mixout-gain-dai-copier-playback.conf>
+<mixout-gain-alh-dai-copier-playback.conf>
 <host-gateway-capture.conf>
 <io-gateway-capture.conf>
 <data.conf>
@@ -93,14 +93,12 @@ Object.Pipeline {
 		}
 	]
 
-	mixout-gain-dai-copier-playback [
+	mixout-gain-alh-dai-copier-playback [
 			{
 			index 2
 
-			Object.Widget.copier.1 {
+			Object.Widget.alh-copier.1 {
 				stream_name	$SDW_PLAYBACK_PCM
-				dai_type	"ALH"
-				copier_type	"ALH"
 				type		"dai_in"
 				node_type $ALH_LINK_OUTPUT_CLASS
 			}
@@ -128,19 +126,42 @@ Object.Pipeline {
 			}
 		}
 	]
+}
 
-	io-gateway-capture [
+# Capture pipeline widgets
+Object.Widget {
+	alh-copier [
 		{
+			stream_name	$SDW_CAPTURE_PCM
 			direction	"capture"
-			index 3
-			copier_type "ALH"
-			Object.Widget.copier.1 {
-				stream_name	$SDW_CAPTURE_PCM
-				dai_type	"ALH"
-				copier_type	"ALH"
-				type		"dai_out"
-				node_type $ALH_LINK_INPUT_CLASS
-			}
+			type		"dai_out"
+			index		3
+			node_type $ALH_LINK_INPUT_CLASS
+			num_input_audio_formats 1
+			num_output_audio_formats 1
+			num_output_pins 1
+
+			Object.Base.input_audio_format [
+				{
+					in_bit_depth		32
+					in_valid_bit_depth	32
+				}
+			]
+			Object.Base.output_audio_format [
+				{
+					out_bit_depth		32
+					out_valid_bit_depth	32
+				}
+			]
+		}
+	]
+
+	pipeline [
+		{
+			index		3
+			priority	0
+			lp_mode		0
+			dynamic_pipeline 1
 		}
 	]
 }
@@ -183,14 +204,14 @@ Object.Base.route [
 	}
         {
                 source 'gain.2.1'
-                sink 'copier.ALH.2.1'
+                sink 'alh-copier.$SDW_PLAYBACK_PCM.0'
         }
 	{
-		source	"copier.ALH.3.1"
+		source	"alh-copier.$SDW_CAPTURE_PCM.0"
 		sink	"host-copier.1.capture"
 	}
 	{
 		source 'host-copier.0.playback'
-		sink 'gain.1.1'
+		sink 'src.1.1'
 	}
 ]
--- a/tools/topology/topology2/include/components/alh-dai-copier.conf
+++ b/tools/topology/topology2/include/components/alh-dai-copier.conf
@@ -1,0 +1,143 @@
+#
+# ALH DAI gateway copier
+#
+# ALH DAI copier widget. All attributes defined herein are namespaced
+# by alsatplg to "Object.Widget.alh-copier.N.attribute_name"
+#
+# Usage: this component can be used by instantiating it in the parent object. i.e.
+#
+# 	Object.Widget.alh-copier."N" {
+#		dai_index 0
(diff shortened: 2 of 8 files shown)


tools/topology/topology2/cavs-nocodec.conf
```diff
@@ -1051,7 +1051,7 @@ IncludeByKey.SSP1_ENABLED {
 						direction	playback
 
 						Object.Widget.dai-copier.1 {
-							dai_index 0
+							dai_index 1
```
plbossart, line 1058: same here for the SSP.

tools/topology/topology2/cavs-sdw-src-gain-mixin.conf
```diff
@@ -137,6 +138,7 @@ Object.Pipeline {
 			Object.Widget.dai-copier.1 {
 				stream_name	$SDW_CAPTURE_PCM
 				dai_type	"ALH"
+				dai_index	10
```
plbossart, line 141: is this one right? I don't recall us using a SoundWire DAI '10'?

tools/topology/topology2/platform/intel/sdw-amp-generic.conf
```diff
@@ -52,6 +52,7 @@ Object.Pipeline {
 			Object.Widget.dai-copier.1 {
 				stream_name $SDW_SPK_STREAM
 				dai_type "ALH"
+				dai_index	20
```
plbossart, line 55: here too. I must admit I don't follow what this dai_index is. It clearly makes no sense if this is used to name a DAI.

tools/topology/topology2/cavs-sdw-src-gain-mixin.conf
bardliao: > is this one right? I don't recall us using a SoundWire DAI '10'? 

Yes, this follows the rule from https://github.com/thesofproject/sof/commit/b407a1aab9e40fc060a6b6ec826df9c61512dc8b
Sdw dai index is not used by node_id. The only rule of sdw dai index is don't be duplicated. 
For example, we can't have 2 dai_index 10.

tools/topology/topology2/platform/intel/sdw-amp-generic.conf
bardliao: > here too. I must admit I don't follow what this dai_index is. It clearly makes no sense if this is used to name a DAI.

`20` means pcm `2`, index `0`.
We include pcm id in the dai_index because the dais are defined in different files and sdw-xxx-generic.conf can be included by different sdw topologys. The dai index could be conflict if they are 0,1,2,3... We don't know what index can we use when we add a new sdw topology unless we check all included .confs.
For the reason above, I think use different dai index group for differnt pcm could be easier to decide a new dai index.

tools/topology/topology2/cavs-sdw-src-gain-mixin.conf
ranj063: @plbossart @bardliao please have a look at this PR again. I have removed those artificial DAI index that we proviously used and replaced it with the actual index of the DAI in an aggregated group. If there is no aggregation the dai_index will always be 0

tools/topology/topology2/include/pipelines/cavs/mixout-gain-alh-dai-copier-playback.conf
```diff
@@ -0,0 +1,101 @@
...
+
+			# copier only supports one format based on mixin/mixout requirements:
+			# 32-bit 48KHz 2ch
+			Object.Base.audio_format [
+				{
+					in_bit_depth		32
+					in_valid_bit_depth	32
+					out_bit_depth		32
+					out_valid_bit_depth	32
+				}
```
plbossart, line 61: can we split the formats?

tools/topology/topology2/cavs-sdw-src-gain-mixin.conf
```diff
@@ -93,14 +93,13 @@ Object.Pipeline {
...
+	mixout-gain-alh-dai-copier-playback [
 			{
 			index 2
 
-			Object.Widget.copier.1 {
+			Object.Widget.alh-copier.1 {
 				stream_name	$SDW_PLAYBACK_PCM
-				dai_type	"ALH"
-				copier_type	"ALH"
+				dai_index	0
```
plbossart, line 102: do we need the dai_index?

tools/topology/topology2/cavs-sdw-src-gain-mixin.conf
```diff
@@ -128,21 +127,39 @@ Object.Pipeline {
...
 	]
+}
 
-	io-gateway-capture [
+# Capture pipeline widgets
+Object.Widget {
+	alh-copier [
 		{
+			stream_name	$SDW_CAPTURE_PCM
+			dai_index	0
```
plbossart, line 137: do we need the dai_index and what would it mean here?

tools/topology/topology2/cavs-sdw-src-gain-mixin.conf
ranj063: Yes, we do need the dai_index. When there's aggregation, the dai_index would be set based on the index of the DAI in the aggregated group. When there' sno aggregation, it will just be 0

tools/topology/topology2/cavs-sdw-src-gain-mixin.conf
ranj063: same thing for all non-aggregated ALH DAIs the dai_index must be set to 0

plbossart: can it instead be omitted and default to zero, and be added only when it's useful?

It's not really a dai_index as much as a copier instance, is it?

ranj063: yes, done!

tools/topology/topology2/include/components/alh-dai-copier.conf
```diff
@@ -0,0 +1,143 @@
+#
```
plbossart, line 1: commit title: topology2: componenets:

tools/topology/topology2/include/components/alh-dai-copier.conf
```diff
@@ -0,0 +1,143 @@
...
+
+		#
+		# immutable attributes cannot be modified in the object instance
+		# num_input_audio_formats for capture and num_output_formats for playback are
+		# mandatory even though the num_input_pins/num_output_pins are 0 because
+		# these are used to set the DMA buffer format.
+		#
+		!immutable [
+			"uuid"
+			"dai_type"
```
plbossart, line 123: isn't the dai_type restricted to ALH only?

plbossart (approved):
LGTM @ranj063, to the extent of my ability to review topology files.

Only a couple of comments below

tools/topology/topology2/include/components/alh-dai-copier.conf
ranj063: yes, its set to ALH below and set to immutable here.

tools/topology/topology2/include/components/alh-dai-copier.conf
ranj063: fixed now

bardliao (approved):

ranj063:
Usual failures with nocodec. Merging.

marc-hb:
Regression reported in
- #7677

PR #7648, opened 2023-05-19
topology2: remove cavs-sdw.tplg from avs-tplg

bardliao:
There is no device that uses the topology.

### DIFF
--- a/tools/topology/topology2/avs-tplg/tplg-targets.cmake
+++ b/tools/topology/topology2/avs-tplg/tplg-targets.cmake
@@ -14,9 +14,6 @@ DEEPBUFFER_FW_DMA_MS=100"
 "sof-hda-generic\;sof-hda-generic-4ch\;\
 HDA_CONFIG=mix,NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,\
 PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-hda-generic-4ch.bin,USE_CHAIN_DMA=true,DEEPBUFFER_FW_DMA_MS=100"
-# CAVS SDW topology with passthrough pipelines
-"cavs-sdw\;cavs-sdw\;DEEPBUFFER_FW_DMA_MS=100,\
-PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-cavs-sdw.bin"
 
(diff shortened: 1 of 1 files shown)


juimonen (approved):

ranj063 (approved):

ranj063:
SOFCI TEST

plbossart:
We do have devices with IPC4 on TGL @bardliao. 
Why wouldn't we allow for these devices to use a SoundWire link? This would even restrict our ability to test IPC4 on UpExtreme.

ranj063:
> We do have devices with IPC4 on TGL @bardliao. Why wouldn't we allow for these devices to use a SoundWire link? This would even restrict our ability to test IPC4 on UpExtreme.

yeah looks like the TGL RVP uses this topology. The question then is if the naming for this topology correct?

fredoh9 (commented):
TGLU_RVP_SDW_IPC4ZPH use cavs-sdw.tplg after renaming to sof-tgl-rt711-rt1308-4ch.tplg. If you can rename to sof-tgl-rt711-rt1308-4ch that would be easier to deploy this topology.

bardliao:
> TGLU_RVP_SDW_IPC4ZPH use cavs-sdw.tplg after renaming to sof-tgl-rt711-rt1308-4ch.tplg. If you can rename to sof-tgl-rt711-rt1308-4ch that would be easier to deploy this topology.

@fredoh9 We have [sof-tgl-rt711-rt1308-4ch](https://github.com/thesofproject/sof/blob/main/tools/topology/topology2/avs-tplg/tplg-targets.cmake#L34) already. No need to use cavs-sdw.tplg and rename to sof-tgl-rt711-rt1308-4ch.tplg.

plbossart:
Ah yes, this PR removes cavs-sdw.tplg, indeed that file should not be used by anyone.

plbossart (approved):

fredoh9:
Looks like SOF CI pretend to test the topology but it has been running the test without speaker pipeline. I fixed the problem and using [sof-tgl-rt711-rt1308-4ch](https://github.com/thesofproject/sof/blob/main/tools/topology/topology2/avs-tplg/tplg-targets.cmake#L34). I'm running the test, will let you know shortly

fredoh9:
SOFCI TEST

fredoh9 (approved):
thank you for clearing this up

PR #7654, opened 2023-05-19
dai: always reload dma ll chain

softwarecki:
A problem was observed with launching capture with dmic. Sometimes the host starts reading data too late, so that the capture buffer becomes full. In this case, the dma descriptors are not reloaded. This leads to stopping the dma. When the host starts reading and reads the accumulated data the stream is already stopped. With this patch, the descriptors are always reloaded and the stream will not be stopped. A glitch may occur but it is better than no data at all.

### DIFF
--- a/src/audio/dai-zephyr.c
+++ b/src/audio/dai-zephyr.c
@@ -1577,6 +1577,7 @@ int dai_zephyr_copy(struct dai_data *dd, struct comp_dev *dev, pcm_converter_fun
 	/* return if nothing to copy */
 	if (!copy_bytes) {
 		comp_warn(dev, "dai_zephyr_copy(): nothing to copy");
+		dma_reload(dd->chan->dma->z_dev, dd->chan->index, 0, 0, 0);
 		return 0;
 	}
 


serhiy-katsyuba-intel (approved):

tmleman (commented):
Shouldn't this be handled during overrun/underrun?

kv2019i (commented):
This definitely fixes a real problem, but I think we should look at other options to get the same effect or at least expand the git commit explanation (the description in pull request is already good).

src/audio/dai-zephyr.c
```diff
@@ -1577,6 +1577,7 @@ int dai_zephyr_copy(struct dai_data *dd, struct comp_dev *dev, pcm_converter_fun
 	/* return if nothing to copy */
 	if (!copy_bytes) {
 		comp_warn(dev, "dai_zephyr_copy(): nothing to copy");
+		dma_reload(dd->chan->dma->z_dev, dd->chan->index, 0, 0, 0);
```
ranj063, line 1580: @softwarecki the multi-endpoint case has been split out into dai_zephyr_multi_endpoint_copy() now. DOesn't the same thing need to be done there as well?

ranj063:
@softwarecki I'm curious to know why this problem occurs in the first place ie "Sometimes the host starts reading data too late". We always start the sink pipeline first followed by the source pipeline. So it's quite surprising that the host copy would not precede a DAI copy

src/audio/dai-zephyr.c
btian1: multi part even no check for copy_bytes == 0 case, may need add also.

kv2019i (approved):
Hmm, I think this makes sense and the expanded commit description helps. At least in capture direction, it makes sense to keep DMA going on if there is still room in the DMA buffer (and if not, there's going to be an xrun with -EPIPE a bit later on -> won't go unnoticed).

And I think this also applies in playback direction. It's a bit odd for pipeline not to provide data (to DAI), but it would seem ok to keep reloading. If the DMA buffer can cover the hickup, then great, if not, we'll get an xrun as DAI DMA runs out of data/free-space.
In playback direction this is a bit weird

tmleman (approved):

kv2019i:
https://sof-ci.01.org/sofpr/PR7654/build8449/devicetest/index.html has multiple failures, but the multiple-pause is known bug that was fixed later last week, the kmod errors are sof-test bugs (due to new modules getting added on kernel side with upstream merge) and then there's an unrealted PM error. I'll proceed with merge.

Issue #7652, opened 2023-05-19
[BUG] topology1: FW pcm_params fail with large period size >44739


kv2019i:
**Describe the bug**
Continuation of 
https://github.com/thesofproject/sof/issues/7476

See log in https://github.com/thesofproject/sof/issues/7476#issuecomment-1550896125

Vamshigopal:
@kv2019i  i took the conf file and dmesg at different instances, so im not sure, let me share the logs and conf file with 50ms period size.
i do agree with your assumption, as i see same error for 50ms, 80ms.. i.e SOF_TKN_SCHED_PERIOD > 44739

kv2019i:
@Vamshigopal Bugfix candidate at https://github.com/thesofproject/sof/pull/7655
I didn't realize we had similar signed 32bit arithmetic in FW to calculate the frame size. :( So this obviously fails in the same conditions as the m4 calculations failed.

kv2019i:
Fix merged, closing.

PR #7655, opened 2023-05-19
audio: component: fix frame size calculation with large period sizes

kv2019i:
If the product of sampling rate and period size (in microseconds)
exceeds INT_MAX (signed 32bit), an invalid component->frame is
set. This happens for example with sampling rate of 48000Hz and
period size of 50000usec.

While the 'rate' and 'period' fields are uint32_t values, ceil_divide()
in math/numbers.h does calculation with signed 32bit values, leading to
invalid results with smaller input values.

Fix the issue by doing the frame size calculation with 64bit
arithmetic.

Link: https://github.com/thesofproject/sof/issues/7652
Reported-by: Vamshi Krishna Gopal <vamshi.krishna.gopal@intel.com>

### DIFF
--- a/src/include/sof/audio/component.h
+++ b/src/include/sof/audio/component.h
@@ -21,6 +21,7 @@
 #include <sof/audio/pipeline.h>
 #include <rtos/panic.h>
 #include <rtos/idc.h>
+#include <sof/common.h>
 #include <sof/list.h>
 #include <rtos/alloc.h>
 #include <sof/lib/cpu.h>
@@ -41,6 +42,7 @@
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
+#include <limits.h>
 
 struct comp_dev;
 struct sof_ipc_stream_posn;
@@ -779,6 +781,8 @@ int comp_set_state(struct comp_dev *dev, int cmd);
 static inline void component_set_nearest_period_frames(struct comp_dev *current,
 						       uint32_t rate)
 {
+	uint64_t frames;
+
 	/* Sample rate is in Hz and period in microseconds.
 	 * As we don't have floats use scale divider 1000000.
 	 * Also integer round up the result.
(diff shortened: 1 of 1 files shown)


Vamshigopal:
Thanks @kv2019i for figuring out the issue. I've verified the fix, it works fine.

src/include/sof/audio/component.h
```diff
@@ -779,27 +780,37 @@ int comp_set_state(struct comp_dev *dev, int cmd);
...
-		current->frames = 96000 * current->period / 1000000;
+		frames = 96000 * period / 1000000;
 		break;
 	case 176400:
-		current->frames = 192000 * current->period / 1000000;
+		frames = 192000 * period / 1000000;
 		break;
 	default:
-		current->frames = ceil_divide(rate * current->period, 1000000);
+		frames = rate * period / 1000000;
```
btian1, line 805: seems duplicated code, can we try to change as below:
	switch (rate) {
	case 44100:
		rate = 48000;
		break;
	case 88200:
		rate = 96000;
		break;
	case 176400:
		rate = 192000;
		break;
	default:
		break;

frames = rate * period / 1000000;
if(rate*period % 1000000)
    frames++;

tmleman (approved):

src/include/sof/audio/component.h
kv2019i: Now updated in V2, @btian1 please check.

src/include/sof/audio/component.h
```diff
@@ -779,27 +781,35 @@ int comp_set_state(struct comp_dev *dev, int cmd);
...
-		current->frames = 96000 * current->period / 1000000;
+		frames = 96000;
 		break;
 	case 176400:
-		current->frames = 192000 * current->period / 1000000;
+		frames = 192000;
 		break;
 	default:
-		current->frames = ceil_divide(rate * current->period, 1000000);
+		frames = rate;
```
btian1, line 805: after use rate, then line 805 can be deleted.

src/include/sof/audio/component.h
```diff
@@ -779,27 +781,35 @@ int comp_set_state(struct comp_dev *dev, int cmd);
...
 		break;
 	}
+
+	frames = SOF_DIV_ROUND_UP(frames * current->period, 1000000);
```
btian1, line 809: frames = SOF_DIV_ROUND_UP(rate * current->period, 1000000);

kv2019i: so this would have to be
frames = SOF_DIV_ROUND_UP((uint64_t)rate * current->period, 1000000);

src/include/sof/audio/component.h
```diff
@@ -779,27 +781,35 @@ int comp_set_state(struct comp_dev *dev, int cmd);
...
 	 * Also integer round up the result.
 	 * dma buffer size should align with 32bytes which can't be
 	 * compatible with current 45K adjustment. 48K is a suitable
 	 * adjustment.
 	 */
+
 	switch (rate) {
 	case 44100:
-		current->frames = 48000 * current->period / 1000000;
+		frames = 48000;
```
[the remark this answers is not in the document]
kv2019i: @btian1 rate is a 32bit variable, so we can't use that or the logic on L809 goes wrong. 

Hmmm, I can of course add type-casting on L805

kv2019i:
V3:
- with reusing "rate" and casting to uint64_t in calculation

lyakh (approved):

btian1 (approved):

kv2019i:
Known fails in https://sof-ci.01.org/sofpr/PR7655/build8405/devicetest/index.html and https://sof-ci.01.org/sofpr/PR7655/build8406/devicetest/index.html , proceeding with merge. Thanks to all reviewers!

PR #7656, opened 2023-05-19
Tools: Test: Audio: Add test for level dependent gain

singalsu:
This patch adds to process_test.m test with amplitude sweep to determine level dependent logarithmic gain.

Currently there is no test pass/fail criteria. But the test result plot can be examined from directory plots. E.g. first full test for DRC with "process_test('drc', 32, 32, 48000, 1);".

The view the plots from shell with "eog plots/ldlg_drc_*.png". Or examine the plots directory with a file maneger GUI.


singalsu:
Here's example of process test for DRC to show the compression characteristic. 

ldlg_drc_32_32_48000_1
ldlg_drc_32_32_48000_1

btian1 (approved):

kv2019i (approved):

PR #7660, opened 2023-05-19
dai-zephyr: fix use-after-free in dai copy

kv2019i:
The dd->local_buffer cannot be assumed to be stable across calls to component copy function. Specifically in case DAI is connected to multiple sinks, it is possible that one of the sinks is disconnected while the DAI is still running. E.g. this happens with SOF sof-cavs-nocodec topology with a capture DAI feeding both a smart-amp (in playback pipeline) and a normal capture pipeline. If the playback pipeline with smart-amp is stopped, the dd->local_buffer will point to a freed buffer, leading to heap corruption.

Fix the issue by updating the local_buffer pointer at start of dai_copy().

Link: https://github.com/thesofproject/sof/issues/7191


src/audio/dai-zephyr.c
```diff
@@ -896,21 +914,14 @@ int dai_zephyr_params(struct dai_data *dd, struct comp_dev *dev,
...
-	else
-		dd->local_buffer = list_first_item(&dev->bsink_list,
-						   struct comp_buffer,
-						   source_list);
-
 	/* check if already configured */
 	if (dev->state == COMP_STATE_PREPARE) {
 		comp_info(dev, "dai_zephyr_params() component has been already configured.");
 		return 0;
 	}
```
ranj063, line 922: @kv2019i this is an excellent find. In the case of the DaI  copier with multiple sinks, we connect all the sinks during hw_params and then during trigger we trigger all the pipelines. And depending on which order the sinks are connected, it could either be the smart_amp or the gain module that goes first. But I dont think the right solution is to set the local_buffer during copy every time. Instead we need to identify the buffer in the capture path to the gain and set that as the local_buffer during params. This is needed because we use the local_buffer frame_fmt to choose the processing function for the DAI copy.

ranj063: @kv2019i I thought about this a bit more and I think a more reasonable solution would be to remove this altogether https://github.com/thesofproject/sof/blob/38f3f5d4119e382c8cee99bbc6ec1d3fff407f93/src/audio/dai-zephyr.c#L267.

And let the loop below handle all sinks. This way all the sinks that are not connected won't be used also.

btian1: should also need remove line L269? totally use loop?

kv2019i: @ranj063 @btian1 But L267 is outside IPC4 ifdef, so this would seem to break IPC, right? 

I don't think setting local_buf at every copy is very expensive as we are taking the first item on the list, so this should be easy case for the compiler to handle. But I do agree with @lyakh this feels brittle. We should have some explicit action when modules are disconnected, both for functional correctness and to make the code easier to follow. Now storing a local copy of bsource_list/bsink_list sounds like a very, very, very bad idea when ipc_comp_disconnect() may be called at any time.

But should we have a simple fix first, or should I send a version that adds bind/unbind to copier right away?


@lyakh yes, this feels

kv2019i: @ranj063 Hmm, I tried a simple variant where I pick the local_buffer from the sink that is on the same pipeline (as components of the same pipeline are guaranteed to be persistent while pipeline is playing). But I seem to be hitting errors that need more debug. Plus, not sure whether we can assume copier at least one sink connected to a component in the same pipeline. Probably always the case in practise, but this configuration is not really forbidden either.

kv2019i: Addressed dd->process reconfig in V2 and implemented unbind. I'll close this conversation, please comment on the new code.

kv2019i:
CI results for V2. Looking at the Intel internal CI system fail at https://sof-ci.01.org/sof-pr-viewer/#/build/PR7660/build12047489 -> not sure of rootcause yet. cAVS2.5 Intel tests look good (all pass), but one fail in https://sof-ci.01.org/sofpr/PR7660/build8330/devicetest . Latter seems a bit odd as there is no error in either dmesg or FW mtrace. arecord returns a non-zero status and multiple-pause-resume test fails, but this would seem like a different case.

src/audio/copier/copier.c
```diff
@@ -1820,6 +1831,7 @@ static const struct comp_driver comp_copier = {
 		.dai_ts_stop			= copier_dai_ts_stop_op,
 		.dai_ts_get			= copier_dai_ts_get_op,
 		.dai_get_hw_params		= copier_get_hw_params,
+		.unbind			= copier_unbind,
```
btian1, line 1834: 1. I am trying to remove those not compatible interface with module, this is a newly added one.
2. where to call copier_unbind? I did not find in current PR.

src/audio/dai-zephyr.c
```diff
@@ -1461,6 +1461,37 @@ int dai_zephyr_multi_endpoint_copy(struct dai_data **dd, struct comp_dev *dev,
...
+	local_buf = buffer_acquire(dd->local_buffer);
+	local_fmt = audio_stream_get_frm_fmt(&local_buf->stream);
+	buffer_release(local_buf);
+
+	dd->process = pcm_get_conversion_function(local_fmt, dma_fmt);
+
+	if (!dd->process) {
+		comp_err(dev, "converter function NULL: local fmt %d dma fmt %d\n",
+			 local_fmt, dma_fmt);
+		dd->local_buffer = NULL;
```
lyakh, line 1491: can we use a temporary variable in lines 1473-1480 above and only assign `dd->local_buffer` after this check?

lyakh (approved):
just one minor potential improvement

src/audio/dai-zephyr.c
```diff
@@ -1504,6 +1535,15 @@ int dai_zephyr_copy(struct dai_data *dd, struct comp_dev *dev, pcm_converter_fun
 
 	buffer_release(buf_c);
 
+	/* handle module runtime unbind */
+	if (!dd->local_buffer) {
+		set_new_local_buffer(dd, dev);
+
+		if (!dd->local_buffer) {
+			comp_warn(dev, "dai_zephyr_copy(): local buffer unbound, cannot copy");
```
btian1, line 1543: unbound --> unbind

src/include/sof/audio/dai_copier.h
```diff
@@ -52,4 +52,7 @@ int dai_zephyr_get_hw_params(struct dai_data *dd, struct comp_dev *dev,
 int dai_zephyr_multi_endpoint_copy(struct dai_data **dd, struct comp_dev *dev,
 				   struct comp_buffer *multi_endpoint_buffer,
 				   int num_endpoints);
+
+int dai_zephyr_unbind(struct dai_data *dd, struct comp_dev *dev, void *data);
+
```
btian1, line 57: extra blank line add

RanderWang (approved):
LGTM

src/audio/copier/copier.c
lyakh: I think we discussed this and decided to keep them

src/audio/dai-zephyr.c
lyakh: @btian1 I think this is a common "short hand" for a passive voice construct, the full version would be "buffer is unbound" or "buffer was unbound" or similar or maybe it should be interpreted as "buffer departed" but in any case "unbind" would be even more wrong

src/audio/copier/copier.c
kv2019i: @btian1 1. You probably know better with @ranj063 how the disconnect is handled with module interface (not adapted but native module_interface.h implementation). I think this could be easier as the list of input/outputs is passed to every call to process() so the module anyways has to check for changes on every call. So the converted interface might not need this at all. But as long as we use component, the possibility of disconnect must be handled and here I do it with unbind.

For example module_adapter.c does handle unbind and it modifies the list of buffers passed to module process(). I.e. it covers this scenario.

2. copier_unbind is called from ipc4_unbind_module_instance() -> ipc_comp_disconnect() -> comp_unbind()

src/audio/dai-zephyr.c
kv2019i: @lyakh Took a look at this, but actually I'd prefer not to. We know at this point already that the old value of "dd->local_buffer" is invalid and will be sooner freed from the heap, so we want to erase in any case. I.e. even if no converter is found, we want to reset dd->local_buffer.

kv2019i:
V3:
- rebase only, no changes to code

src/include/sof/audio/dai_copier.h
kv2019i: @btian1 Oops, forgot to check this. But hmm, there's traditionally blank link between last line of header declaration and "#endif" (this is used in Zephyr and Linux so I'd say let's follow same policy). We seem to have a few exceptions in SOF, but these should be fixed.

src/audio/dai-zephyr.c
kv2019i: hmm, native English speakers can correct, but I think unbound is correct. it's here as adjective (unbound buffer), but could be indeed read as shorthand for past tense verb ("buffer freed", "buffer allocated").

src/audio/copier/copier.c
ranj063: yes, lets keep it for now. the module adapter might need extensions to handle this

ranj063 (approved):

btian1 (approved):

src/audio/dai-zephyr.c
lyakh: sorry, maybe I didn't explain well. This function is only called `if (!dd->local_buffer)` right? Then it is assigned a nun-NULL value either in line 1474 or in line 1478 above. But then we check if we can find a conversion function and if not - set that pointer to NULL *again*. Whereas we could simply avoid setting it to something we cannot use anyway completely. That way at no point of time would it be set to something useless. I.e.
```
	struct comp_buffer *buf;
	...
	if (dev->direction == SOF_IPC_STREAM_PLAYBACK)
		buf = list_first_item(&dev->bsource_list,
						   struct comp_buffer,
						   sink_list);
	else
		buf = list_first_item(&dev->bsink_list,
						   struct comp_buffer,
						   source_list);

	local_buf = buffer_acquire(buf);
	local_fmt = audio_stream_get_frm_fmt(&local_buf->stream);
	buffer_release(local_buf);

	dd->process = pcm_get_conversion_function(local_fmt, dma_fmt);

	if (!dd->process)
		comp_err(dev, "converter function NULL: local fmt %d dma fmt %d\n",
			 local_fmt, dma_fmt);

	dd->local_buffer = buf;
```

kv2019i: Right, now I get your point @lyakh . Let me do a small fixup to this in separate PR given I got a green run on the CI and ok from reviewers and we can't of really need to have this fix asap in the tree.

kv2019i:
CI good, only one known case fails in https://sof-ci.01.org/sofpr/PR7660/build8330/devicetest/index.html .

I'll proceed with merge but promise to fixup the local_buffer double-setting as per @lyakh 's comment.

PR #7664, opened 2023-05-22
dts: handle the zero config size in apply_config instead of set_configuration

joechengxperi:
To handle zero config size case in set_configuration would let dts_codec_apply_config() never be called due to early return when checking if md->new_config_size is zero. This causes the problem that DTS tuning parameter would never be set. We've found this issue on the latest Chrome OS version on RPL. So, instead, check the config size in dts_codec_apply_config() could resolve the zero config case and let the tuning parameter could be set successfully.


joechengxperi:
Close this PR due to include an additional commit not belong to DTS. Will create a new one.

PR #7665, opened 2023-05-22
dts: handle zero config size case in apply_config instead of set_configuration

joechengxperi:
To handle zero config size case in set_configuration would let dts_codec_apply_config never be called due to early return when checking if md->new_config_size is zero. This causes the problem that DTS tuning parameter would never be set. We've found the issue on the latest Chrome OS version on RPL. So, instead, check the config size in dts_codec_apply_config() could resolve the zero config case and let the tuning parameter could be set successfully.

### DIFF
--- a/src/audio/module_adapter/module/dts.c
+++ b/src/audio/module_adapter/module/dts.c
@@ -315,11 +315,11 @@ static int dts_codec_apply_config(struct processing_module *mod)
 	/* Check that config->data isn't invalid and has size greater than 0 */
 	config_header_size = sizeof(config->size) + sizeof(config->avail);
 	if (config->size < config_header_size) {
-		comp_err(dev, "dts_codec_apply_config() config->data is invalid");
-		return -EINVAL;
+		comp_warn(dev, "dts_codec_apply_config() config->data is invalid");
+		return 0;
 	} else if (config->size == config_header_size) {
-		comp_err(dev, "dts_codec_apply_config() size of config->data is 0");
-		return -EINVAL;
+		comp_warn(dev, "dts_codec_apply_config() size of config->data is 0");
+		return 0;
 	}
 
 	/* Calculate size of config->data */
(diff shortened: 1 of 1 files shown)


johnylin76 (approved):
The bug in the recent code results from the calling of module_set_configuration() may reset md->new_cfg_size. The recent checker will be wrong under such cases.

markbartonxperi (approved):

src/audio/module_adapter/module/dts.c
```diff
@@ -316,10 +316,10 @@ static int dts_codec_apply_config(struct processing_module *mod)
 	config_header_size = sizeof(config->size) + sizeof(config->avail);
 	if (config->size < config_header_size) {
 		comp_err(dev, "dts_codec_apply_config() config->data is invalid");
-		return -EINVAL;
+		return 0;
```
kv2019i, line 319: Should the comp_err() be changed to comp_warn() (or lower) if this is no longer an error? Do we lose the ability to flag real errors when we do not return -EINVAL?

kv2019i (commented):
Change looks valid, but a notes inline

src/audio/module_adapter/module/dts.c
johnylin76: > Should the comp_err() be changed to comp_warn() (or lower) if this is no longer an error?

It sounds good. @joechengxperi please change to use comp_warn instead.

> Do we lose the ability to flag real errors when we do not return -EINVAL?

It should be fine to keep it silent. In my view the real error should not be met in common use cases (where the constant config is set by script on audio system server initializing). The checker after modification can still make sure that the invalid config won't be applied in any cases.

joechengxperi: @kv2019i @johnylin76 
Updated to comp_warn().

cujomalainey (approved):

cujomalainey:
@andyross looks like Fuzzer CI is borked? Not sure if this is stale

PR #7670, opened 2023-05-23
Change the naming scheme for the SSP/HDA/DMIC DAI copiers

ranj063:
Change the convention from 
`copier.<copier_type>.<pipeline ID>.<instance ID>` 
to
`dai-copier.<copier_type>.<stream_name>.<direction>`

Depends on https://github.com/thesofproject/linux/pull/4375


tools/topology/topology2/platform/intel/speaker-echo-ref.conf
```diff
@@ -17,6 +17,7 @@ Object.Pipeline {
 	dai-copier-be [
 		{
 			index	$ECHO_REF_DAI_PIPELINE_ID
+			direction	"capture"
```
bardliao, line 20: Why didn't we need `direction` before?

ranj063: its never used before. direction is only meant for IPC3. But now I need it for the name

tools/topology/topology2/cavs-mixin-mixout-hda.conf
```diff
@@ -123,15 +123,15 @@ Object.PCM.pcm [
...
+		sink 'dai-copier.HDA.$HDA_ANALOG_DAI_NAME.playback'
 		source 'gain.2.1'
 	}
 	{
 		source 'mixin.1.1'
 		sink 'mixout.2.1'
 	}
 	{
-		source 'dai-copier.HDA.4.1'
+		source 'dai-copier.HDA.$HDA_ANALOG_DAI_NAME.capture'
```
plbossart, line 134: if we need to reduce the widget name length, "playback" and "capture" are obvious targets.

We could use something like "p"/"c" (as done for ALSA in /proc pcm0c/   pcm0p/   pcm10p/  pcm1p/). Or 'pb' and 'cp'.

Just a suggestion.

plbossart (approved):
one suggestion below @ranj063

bardliao (approved):

ranj063:
SOFCI TEST

tools/topology/topology2/include/components/dai-copier.conf
```diff
@@ -137,6 +137,9 @@ Class.Widget."dai-copier" {
```
kv2019i, line 137: "These are" or "They are"

kv2019i (approved):
The dependency to kernel change is unfortunate, but I guess a price we have to pay. Otherwise, this is a massive improvement to readability/maintainability of the tplg files, so a very welcome change.

tools/topology/topology2/include/components/dai-copier.conf
ranj063: fixed and now the kernel dependency is merged as well. So lets retest

kv2019i:
Nothing related to this PR in https://sof-ci.01.org/sofpr/PR7670/build8583/devicetest/index.html , proceeding with merge.

PR #7672, opened 2023-05-23
schedule: zephyr_dma_domain: Fix various issues

LaurentiuM1234:
Issues fixed by this series of patches:

1. Zephyr DMA domain doesn't support DMACs with aggregated IRQs. This lead to the DMA IRQ handler giving multiple semaphore resources (since the handler is registered multiple times for the same IRQ but with different data)
2. Each time a DMA IRQ was triggered all pipeline tasks were executed which, in the case of i.MX8MP, would lead to warnings such as "no bytes to copy". To fix this, a pipeline task is now executed only if a DMA IRQ was triggered for the channel used by said task.

### DIFF
--- a/src/include/sof/audio/pipeline.h
+++ b/src/include/sof/audio/pipeline.h
@@ -129,6 +129,10 @@ struct pipeline_task {
 	struct task task;		/**< parent structure */
 	bool registrable;		/**< should task be registered on irq */
 	struct comp_dev *sched_comp;	/**< pipeline scheduling component */
+	/* private data used by Zephyr DMA domain for internal
+	 * housekeeping - don't touch.
+	 */
+	void *priv;
 };
 
 #define pipeline_task_get(t) container_of(t, struct pipeline_task, task)
--- a/src/include/sof/schedule/ll_schedule_domain.h
+++ b/src/include/sof/schedule/ll_schedule_domain.h
@@ -195,13 +195,10 @@ static inline void domain_disable(struct ll_schedule_domain *domain, int core)
 static inline bool domain_is_pending(struct ll_schedule_domain *domain,
 				     struct task *task, struct comp_dev **comp)
 {
-	bool ret;
-
-	assert(domain->ops->domain_is_pending);
-
-	ret = domain->ops->domain_is_pending(domain, task, comp);
-
-	return ret;
+	if (domain->ops->domain_is_pending)
+		return domain->ops->domain_is_pending(domain, task, comp);
+	else
+		return true;
 }
 
 
--- a/src/schedule/zephyr_dma_domain.c
+++ b/src/schedule/zephyr_dma_domain.c
@@ -143,13 +143,16 @@ static void dma_irq_handler(void *data)
 	irq = zephyr_dma_domain_data->irq;
 	dt = zephyr_dma_domain_data->dt;
 
-	/* clear IRQ */
-	dma_interrupt_legacy(channel, DMA_IRQ_CLEAR);
-	interrupt_clear_mask(irq, BIT(channel_index));
+	/* was the IRQ triggered by this channel? */
+	if (dma_interrupt_legacy(channel, DMA_IRQ_STATUS_GET)) {
+		/* clear IRQ */
+		dma_interrupt_legacy(channel, DMA_IRQ_CLEAR);
+		interrupt_clear_mask(irq, BIT(channel_index));
 
-	/* give resources to thread semaphore */
-	if (dt->handler)
-		k_sem_give(sem);
+		/* give resources to thread semaphore */
+		if (dt->handler)
+			k_sem_give(sem);
+	}
 }
 
 static int enable_dma_irq(struct zephyr_dma_domain_data *data)
--- a/src/schedule/zephyr_dma_domain.c
+++ b/src/schedule/zephyr_dma_domain.c
@@ -56,6 +56,27 @@ struct zephyr_dma_domain_data {
 	 * associated channel was triggered.
 	 */
 	bool triggered;
+	/* used to keep track of how many pipeline tasks point
+	 * to the same data structure.
+	 */
+	int usage_count;
+	/* used to keep track of how many times data->triggered
+	 * can be left set to true before resetting it to false.
+	 *
+	 * this is needed for mixer topologies because the
+	 * non-registrable pipeline tasks need to be executed
+	 * when an IRQ comes for the registrable pipeline task.
+	 */
+	int pending_count;
+	/* used when linking a data structure to a non-registrable
+	 * pipeline task.
+	 *
+	 * in the case of mixer, the non-registrable pipeline tasks
+	 * have the same scheduling component as the registrable
+	 * pipeline task so this can be used as a criteria to attach
+	 * the data structure to the non-registrable pipeline task.
+	 */
+	struct comp_dev *sched_comp;
 };
 
 struct zephyr_dma_domain_thread {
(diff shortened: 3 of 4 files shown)


src/schedule/zephyr_dma_domain.c
```diff
@@ -453,15 +482,20 @@ static int zephyr_dma_domain_unregister(struct ll_schedule_domain *domain,
...
 
 	zephyr_dma_domain = ll_sch_get_pdata(domain);
 	pipe_task = pipeline_task_get(task);
 	core = cpu_get_id();
 	dt = zephyr_dma_domain->domain_thread + core;
+	data = pipe_task->priv;
 
 	tr_info(&ll_tr, "zephyr_dma_domain_unregister()");
 
+	data->pending_count--;
```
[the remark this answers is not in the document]
LaurentiuM1234: In theory this shouldn't happen since `domain_is_pending()` always resets its value when 0. Since cancelling a task is synchronous that means `domain_is_pending()` will always be called.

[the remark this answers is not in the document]
LaurentiuM1234: Still need to stress test and if this does happen then we have a really big problem.

src/include/sof/audio/pipeline.h
```diff
@@ -129,6 +129,10 @@ struct pipeline_task {
 	struct task task;		/**< parent structure */
 	bool registrable;		/**< should task be registered on irq */
 	struct comp_dev *sched_comp;	/**< pipeline scheduling component */
+	/* private data used by Zephyr DMA domain for internal
+	 * housekeeping - don't touch.
+	 */
+	void *priv;
```
dbaluta, line 135: I wouldn't mention about Zephyr DMA domain here. Other's might use this also in the future. This is common enough inside structures so a generic comment like `private data for internal housekeeping` would be enough.

src/include/sof/schedule/ll_schedule_domain.h
```diff
@@ -195,13 +195,10 @@ static inline void domain_disable(struct ll_schedule_domain *domain, int core)
...
-
-	assert(domain->ops->domain_is_pending);
-
-	ret = domain->ops->domain_is_pending(domain, task, comp);
-
-	return ret;
+	if (domain->ops->domain_is_pending)
+		return domain->ops->domain_is_pending(domain, task, comp);
+	else
+		return true;
```
dbaluta, line 201: Usual pattern here is to remove the `else` and say somthing like this.

```
if (foo)
     return foo();

return true;
```

src/schedule/zephyr_dma_domain.c
```diff
@@ -219,6 +231,15 @@ static int register_dma_irq(struct zephyr_dma_domain *domain,
 			crt_data->channel = crt_chan;
 			crt_data->dt = dt;
 
+			/* attach crt_data to pipeline task.
+			 *
+			 * this will be used in
+			 * zephyr_dma_domain_is_pending to decide
+			 * whether there was a DMA IRQ for the channel
+			 * associated with this task.
+			 */
```
dbaluta: chars per line limit is at 100. So we can compact the compact the comments here on fewer lines.

src/schedule/zephyr_dma_domain.c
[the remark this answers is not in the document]
paulstelian97: Yeah I feel like this very decrement can get it below zero, before `domain_is_pending()` has a chance to reset it.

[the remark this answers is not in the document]
paulstelian97: Say, if you're somehow stopping two pipelines quickly which leads to unregistering things inbetween them. Say if it's playback and capture (which work with different scheduling component). The following order may cause issues:

1. IRQ arrives for playback; it is handled, pending = 1
2. Playback pipeline is torn down, pending = 0
3. Capture pipeline is torn down before IRQ ever arrives for capture this period, pending = -1

It's pretty hard to actually have this happen I might say, but having "pending" not equate an actual pipeline walk pending simply doesn't sit well with me (variable name hinting one thing while it technically _isn't_ that thing)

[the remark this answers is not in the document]
paulstelian97: Okay then my only remaining complaint is that the name can send you down the wrong train of thought. I can accept it if nobody else agrees with me though.

src/include/sof/audio/pipeline.h
lyakh: cannot you just add your `ctrl_data` pointer to `struct zephyr_ll_pdata` stored in `task.priv_data`?

src/schedule/zephyr_dma_domain.c
```diff
@@ -56,6 +56,27 @@ struct zephyr_dma_domain_data {
 	 * associated channel was triggered.
 	 */
 	bool triggered;
+	/* used to keep track of how many pipeline tasks point
+	 * to the same data structure.
```
lyakh, line 60: is it the same as just "number of tasks, sharing the same data structure?"

src/schedule/zephyr_dma_domain.c
```diff
@@ -56,6 +56,27 @@ struct zephyr_dma_domain_data {
...
+	 */
+	int usage_count;
+	/* used to keep track of how many times data->triggered
```
lyakh, line 63: can we similarly simplify this and remove "used to keep track of?"

src/schedule/zephyr_dma_domain.c
```diff
@@ -219,6 +240,24 @@ static int register_dma_irq(struct zephyr_dma_domain *domain,
 			if (core != crt_chan->core)
 				continue;
 
+			if ((crt_data->sched_comp == pipe_task->sched_comp ||
+			     !crt_data->sched_comp) && !pipe_task->registrable) {
+				/* attach crt_data to current pipeline
+				 * task only if there's a match between
+				 * the scheduling components or i
```
lyakh, line 247: "if"

src/schedule/zephyr_dma_domain.c
```diff
@@ -219,6 +240,24 @@ static int register_dma_irq(struct zephyr_dma_domain *domain,
...
+				 * set yet.
+				 */
+				pipe_task->priv = crt_data;
+
+				crt_data->usage_count++;
+				crt_data->pending_count++;
+
+				return 0;
+			} else if (!pipe_task->registrable) {
+				return 0;
```
lyakh, line 258: can we do this:
```
if (!pipe_task->registrable) {
	if (crt_data->sched_comp && crt_data->sched_comp != pipe_task->sched_comp)
		return 0;

	/* attach crt_data to current pipeline
	...
}

File: src/schedule/zephyr_dma_domain.c [hunk 9]
```diff
@@ -309,12 +334,16 @@ static int zephyr_dma_domain_register(struct ll_schedule_domain *domain,
...
 		tr_err(&ll_tr,
-		       "failed to register DMA IRQ for pipe task %p on core %d",
-		       pipe_task, core);
+		       "failed to register DMA IRQ for pipe task %p on core %d reg: %d",
+		       pipe_task, core, pipe_task->registrable);
 
 		return ret;
 	}
 
+	/* don't let non-registrable tasks go further */
```
lyakh, line 343: this comment doesn't add any information - it simply restates the next statement. If a comment is needed here, it should explain something that isn't obvious from the code

src/schedule/zephyr_dma_domain.c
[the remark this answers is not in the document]
lyakh: maybe add an `assert()` at least

src/schedule/zephyr_dma_domain.c
LaurentiuM1234: Yep.

src/schedule/zephyr_dma_domain.c
LaurentiuM1234: Sure

src/schedule/zephyr_dma_domain.c
LaurentiuM1234: You're right. Will remove or rework it.

LaurentiuM1234:
Closing this as it will come in conflict with https://github.com/thesofproject/sof/pull/7923. Will eventually re-open.

PR #7673, opened 2023-05-23
tools: Remove older fuzz framework

andyross:
Fuzzing via the new framework is now integrated at oss-fuzz upstream, so there's no point to keep this in SOF anymore.  The github workflow has bitrot vs. the newer build.sh, and that docker build is very heavyweight vs. the newer fuzz.sh smoke test that runs in the regular build container anyway.


andyross:
Note that this removes a github workflow, which mildly terrifies me as I never know how those work or if the CI build reflects the old or updated configuration.

cujomalainey (approved):

cujomalainey:
No major workflows broken and I am the sole owner, merging

PR #7674, opened 2023-05-23
Revert "tools: Remove older fuzz framework"

andyross:
This got merged too fast.  Turns out it broke the newer fuzz integration that was in the same YAML file.  Also there are some evolving review comments.  Will resubmit.

This reverts commit 11e57f50309dbaec3a665dde951b38e219e9f762.


marc-hb (approved):
> @marc-hb do we want to do a full revert and bring back a known CI system

Yes please because this will make the code review clearer and simpler.

It's only one isolated CI check failing without any other side-effect and it shouldn't last for longer than a day or two. We've survived much worse for much longer.

cujomalainey (approved):

PR #7678, opened 2023-05-24
Copier-dai: copier and dai part code split and refactor

btian1:


src/audio/copier/copier_dai.c
```diff
@@ -281,3 +281,29 @@ void copier_dai_free(struct copier_data *cd)
...
+			ret = dai_zephyr_config_prepare(cd->dd[i], dev);
+			if (ret < 0)
+				return ret;
+
+			ret = dai_zephyr_prepare(cd->dd[i], dev);
+			if (ret < 0)
+				return ret;
+		}
+	}
+	return ret;
```
lyakh, line 308: is this function expected to return positive values? If not, just `return 0` here and remove initialisation

src/audio/copier/copier_dai.c
```diff
@@ -300,3 +300,156 @@ int copier_dai_prepare(struct comp_dev *dev, struct copier_data *cd)
...
+		} while (r_ptr < r_end_ptr && w_ptr < (int32_t *)audio_stream_get_end_addr(dst));
+
+		src_stream_sample_count -= r_ptr - r_ptr_before_loop;
+	}
+
+	return 0;
+}
+
+int copier_dai_params(struct sof_ipc_stream_params *params,
+		      struct comp_dev *dev, struct copier_data *cd, int dai_index)
```
lyakh, line 387: don't all these exported copier functions have `dev` as their first parameter probably followed by `cd`? Would be good to follow some such convention

btian1: yes, I like put cd first, dev as second, since dev is only for log output, let me change.

src/audio/copier/copier_dai.c
```diff
@@ -286,3 +286,29 @@ void copier_dai_free(struct copier_data *cd)
...
+	} else {
+		for (int i = 0; i < cd->endpoint_num; i++) {
+			ret = dai_zephyr_config_prepare(cd->dd[i], dev);
+			if (ret < 0)
+				return ret;
+
+			ret = dai_zephyr_prepare(cd->dd[i], dev);
+			if (ret < 0)
+				return ret;
+		}
```
lyakh, line 319: it seems to me like you don't need the `if`? Just use the loop and `if (cd->endpoint_num == 1)` the loop will just run once. Am I missing anything?

src/audio/copier/copier.c
```diff
@@ -165,26 +127,16 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
...
+			if (copier_ipcgtw_create(dev, cd, &dev->ipc_config,
+						 copier, ipc_pipe->pipeline)) {
 				comp_cl_err(&comp_copier, "unable to create IPC gateway");
 				goto error_cd;
 			}
 
-			if (cd->direction == SOF_IPC_STREAM_PLAYBACK)
-				ipc_pipe->pipeline->source_comp = dev;
-			else
-				ipc_pipe->pipeline->sink_comp = dev;
```
lyakh, line 575: this doesn't need to be added to `copier_ipcgtw_create()`?

src/audio/copier/copier.c
```diff
@@ -255,8 +254,7 @@ static int copier_reset(struct comp_dev *dev)
 			copier_ipcgtw_reset(dev);
 		break;
 	case SOF_COMP_DAI:
-		for (i = 0; i < cd->endpoint_num; i++)
-			dai_zephyr_reset(cd->dd[i], dev);
+			copier_dai_reset(cd, dev);
```
lyakh, line 257: indentation

src/audio/copier/copier.c
btian1: any reason that this part can't move to copier_ipcgtw_create?
move to create function can make copier_new looks as simple as possible.
Also, inside can combine with other PLAYBACK and CAPTURE case.

src/audio/copier/copier.c
btian1: ok, wrong indentation.

marc-hb:
There was no MTL device available for https://sof-ci.01.org/sofpr/PR7678/build8683/devicetest

src/include/sof/audio/dai_copier.h
```diff
@@ -56,12 +56,9 @@ int dai_zephyr_multi_endpoint_copy(struct dai_data **dd, struct comp_dev *dev,
...
-int copier_dai_init(struct comp_dev *parent_dev,
-		    const struct comp_driver *drv,
-		    struct comp_ipc_config *config,
-		    const struct ipc4_copier_module_cfg *copier,
-		    struct pipeline *pipeline,
-		    struct ipc_config_dai *dai,
-		    enum ipc4_gateway_type type,
-		    int index, int dai_count);
+struct copier_data;
+int copier_dai_create(struct comp_dev *parent_dev, struct copier_data *cd,
```
ranj063, line 60: I would squash this patch with the previous patch, so we dont have to first add copie_dai_init() and then remove it here.

src/audio/copier/copier.c
```diff
@@ -571,6 +544,39 @@ static int copier_copy(struct comp_dev *dev)
...
+		return 0;
+	case SOF_COMP_DAI:
+		if (cd->endpoint_num == 1)
+			return dai_common_copy(cd->dd[0], dev, cd->converter);
+		break;
+	default:
+		break;
+	}
+	/* handle multi-endpoint and maybe type == none copy */
+	return do_multi_endpoint_copy(cd, dev);
```
ranj063, line 560: this is a misnomer. This case handles both the multi-endpoint case for DAI's and the copy for module copiers.

src/include/sof/audio/dai_copier.h
btian1: ok, it will just make the patch looks very big.

src/audio/copier/copier.c
btian1: I agree on this, when I tried to change for this, I found it is not only for multiendpoint, so there are two ways:

1. simply change the function name, like: do_multi_endpoint_module_copy
2. split this function further to two, do_multi_endpoint_copy and do_module_copy

I like 2, what's your opinion?

lyakh (approved):

RanderWang (approved):
LGTM

ranj063 (approved):

kv2019i:
@wszypelt Can you check? It seems new platforms are triggering errors not related to the PR tested.

ranj063:
All tests passing now except for a timeout. merging

PR #7680, opened 2023-05-24
topology2: add sdw format

bardliao:
Some sdw codecs may not support all formats. We can use 'format' to inform to use the specified format.

### DIFF
--- a/tools/topology/topology2/cavs-sdw.conf
+++ b/tools/topology/topology2/cavs-sdw.conf
@@ -73,6 +73,8 @@ Define {
 	NUM_SDW_AMP_LINKS 0
 	SDW_DMIC 0
 	SDW_JACK true
+	SDW_JACK_FORMAT	''
+	SDW_AMP_FORMAT	''
 }
 
 # override defaults with platform-specific config
--- a/tools/topology/topology2/platform/intel/sdw-amp-generic.conf
+++ b/tools/topology/topology2/platform/intel/sdw-amp-generic.conf
@@ -82,6 +82,11 @@ IncludeByKey.NUM_SDW_AMP_LINKS {
 					num_input_audio_formats 1
 					num_output_audio_formats 1
 					num_input_pins 1
+IncludeByKey.SDW_AMP_FORMAT {
+"[s*le]"	{
+					format		$SDW_AMP_FORMAT
+	}
+}
 
 					# 32-bit 48KHz 2ch
 					Object.Base.input_audio_format [
@@ -108,6 +113,11 @@ IncludeByKey.NUM_SDW_AMP_LINKS {
 					num_input_audio_formats 1
 					num_output_audio_formats 1
 					num_output_pins 1
+IncludeByKey.SDW_AMP_FORMAT {
+"[s*le]"	{
+					format		$SDW_AMP_FORMAT
+	}
+}
 
 					# 32-bit 48KHz 2ch
 					Object.Base.input_audio_format [
@@ -243,6 +253,11 @@ IncludeByKey.SDW_AMP_FEEDBACK {
 					num_input_audio_formats 1
 					num_output_audio_formats 1
 					num_output_pins 1
+IncludeByKey.SDW_AMP_FORMAT {
+"[s*le]"	{
+					format		$SDW_AMP_FORMAT
+	}
+}
 
 					IncludeByKey.NUM_SDW_AMP_LINKS {
 					"2"	{
--- a/tools/topology/topology2/platform/intel/sdw-jack-generic.conf
+++ b/tools/topology/topology2/platform/intel/sdw-jack-generic.conf
@@ -65,6 +65,11 @@ Object.Pipeline {
 			Object.Widget.alh-copier.1 {
 				stream_name $SDW_JACK_OUT_STREAM
 				node_type $ALH_LINK_OUTPUT_CLASS
+IncludeByKey.SDW_JACK_FORMAT {
+"[s*le]"	{
+				format		$SDW_JACK_FORMAT
+	}
+}
 			}
 			Object.Widget.gain.1 {
 				Object.Control.mixer.1 {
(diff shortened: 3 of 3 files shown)


tools/topology/topology2/platform/intel/sdw-jack-generic.conf
```diff
@@ -103,6 +108,11 @@ Object.Widget {
 			num_input_audio_formats 1
 			num_output_audio_formats 1
 			num_output_pins 1
+IncludeByKey.SDW_JACK_FORMAT {
+"[s*le]"	{
+					format		$SDW_JACK_FORMAT
+	}
+}
```
plbossart, line 115: I don't think it's a good idea. There is no reason to propagate restrictions on the Data Port WordLength register all the way to the gains

plbossart (changes requested):
not convinced @bardliao, same objections as for https://github.com/thesofproject/linux/pull/4374

tools/topology/topology2/platform/intel/sdw-amp-generic.conf
```diff
@@ -108,6 +113,11 @@ IncludeByKey.NUM_SDW_AMP_LINKS {
 					num_input_audio_formats 1
 					num_output_audio_formats 1
 					num_output_pins 1
+IncludeByKey.SDW_AMP_FORMAT {
+"[s*le]"	{
```
ranj063, line 117: @bardliao I dont think you need the IncludeByKey at all. The format attribute is already resitricted to the s*le and float types

tools/topology/topology2/platform/intel/sdw-jack-generic.conf
ranj063: @plbossart I dont think it propagated to the gain here. Still just the DAI copier

tools/topology/topology2/cavs-sdw.conf
```diff
@@ -73,6 +73,8 @@ Define {
 	NUM_SDW_AMP_LINKS 0
 	SDW_DMIC 0
 	SDW_JACK true
+	SDW_JACK_FORMAT	''
+	SDW_AMP_FORMAT	''
```
ranj063, line 77: @bardliao I have an alternate suggestion instead of empty formats. Why not make the format token a bit mask ie. SDW_AMP_FORMAT "BIT(0) | BIT (1) | BIT(2)" to say that all 3 formats are supported. 

Then in the kernel you can test if all formats are support or not and do a fixup based on what is supported.

tools/topology/topology2/platform/intel/sdw-jack-generic.conf
plbossart: right but still we have the valid_bit_depth so why do we need a format to represent what can be sent on the bus?

ranj063: Ahh so are you suggesting using the valid_bit_depth/container size in the output formats of the DAI copier? If it really only has support for 1 format, all of the output formats will have the same valid bit depth and container size and the kernel can infer the supported format from there?

plbossart: I am not sure if the kernel can do everything magically, my question is really whether there is a conceptual difference between valid_bit_depth and this new "format" suggestion.

it's also interesting because it's a limitation from the 3rd party chip that trickles into the topology, we've never had to deal with this before.

ranj063: I think using the valid_bit_depth can be used as a proxy for doing the sample format fixup. @bardliao what do you think?

tools/topology/topology2/platform/intel/sdw-amp-generic.conf
bardliao: > @bardliao I dont think you need the IncludeByKey at all. The format attribute is already resitricted to the s*le and float types

I added it because otherwise we will set format = '' by default which is invalid.

tools/topology/topology2/cavs-sdw.conf
bardliao: > @bardliao I have an alternate suggestion instead of empty formats. Why not make the format token a bit mask ie. SDW_AMP_FORMAT "BIT(0) | BIT (1) | BIT(2)" to say that all 3 formats are supported.
> 
> Then in the kernel you can test if all formats are support or not and do a fixup based on what is supported.

2 opens if we use mask.
1. What should kernel select if only s24_le and s16_le is supported, but user play with s32_le? 
2. The format token is a string in IPC3/topology1, if we use a mask in topology2, then topology2 will never be used in IPC3.

tools/topology/topology2/platform/intel/sdw-jack-generic.conf
bardliao: @plbossart @ranj063 valid_bit_depth  depends on hw_params. It will be the format that user play/capture. For example, it will be 16 when user play a s16_le wav.
The PR and the corresponding kernel PR is trying to specify the format of hw_params.

bardliao: > not following @bardliao, the hw_params depend on what the BE can support, not what applications request.

Right, but if the BE can support all formats, hw_params will be the format that applications request.
This comment let me think the issue is actually that ASoC doesn't change the hw_params if the format is not supported by the codec. For example, rt711-sdca supports `(SNDRV_PCM_FMTBIT_S16_LE | SNDRV_PCM_FMTBIT_S20_3LE | SNDRV_PCM_FMTBIT_S24_LE)`, but the format of hw_params is still s32_le if we play s32_le format file.

> 
> The point is that we have two concepts that seem completely identical, so why not just use one.

We set `valid_bit_depth` in the available formats and we can have multiple `valid_bit_depth` in a host widget.
And we can't use the `valid_bit_depth` in base format to specify the BE format in `sof_ipc4_pcm_dai_link_fixup` because base format will be decided in `sof_pcm_hw_params` -> `.ipc_prepare`.
And the `format` token is used to specify the format in IPC3. That's why I reuse the `format` token to specify the format in IPC4, too.

lgirdwood:
@bardliao any update ?

plbossart:
sorry @lgirdwood I may be the problematic reviewer on this one. I'll work with @bardliao and @ranj063 to get this right.

bardliao:
We don't need this PR to specify BE format.

PR #7681, opened 2023-05-24
Topology2: add 1 channel capture support

bardliao:
Adding 1 channel capture support

### DIFF
--- a/tools/topology/topology2/cavs-sdw.conf
+++ b/tools/topology/topology2/cavs-sdw.conf
@@ -74,6 +74,7 @@ Define {
 	SDW_DMIC 0
 	SDW_JACK true
 	PASSTHROUGH	false
+	SDW_JACK_CAPTURE_CH 2
 }
 
 # override defaults with platform-specific config
--- a/tools/topology/topology2/platform/intel/sdw-jack-generic.conf
+++ b/tools/topology/topology2/platform/intel/sdw-jack-generic.conf
@@ -314,6 +314,8 @@ Object.PCM.pcm [
 		Object.PCM.pcm_caps.1 {
 			name "Passthrough Capture 0"
 			formats 'S16_LE,S32_LE'
+			channels_min	$SDW_JACK_CAPTURE_CH
+			channels_max	$SDW_JACK_CAPTURE_CH
 		}
 	}
 ]
--- a/tools/topology/topology2/platform/intel/sdw-jack-generic.conf
+++ b/tools/topology/topology2/platform/intel/sdw-jack-generic.conf
@@ -180,10 +180,55 @@ Object.Pipeline.host-gateway-capture [
 		Object.Widget.host-copier.1 {
 			stream_name	"Passthrough Capture 0"
 			pcm_id 1
+			num_input_audio_formats 9
+			num_output_audio_formats 9
 			Object.Base.audio_format.1 {
 				# 32/32 -> 16/16 bits conversion is done here
 				in_bit_depth	32
 				in_valid_bit_depth	32
+			}
+			# 32-bit 48KHz 1ch
+			Object.Base.input_audio_format.7 {
+				in_channels		1
+				in_bit_depth		32
+				in_valid_bit_depth	32
+				in_ch_cfg	$CHANNEL_CONFIG_MONO
+				in_ch_map	$CHANNEL_MAP_MONO
+			}
+			Object.Base.output_audio_format.7 {
+				out_bit_depth		32
+				out_valid_bit_depth	32
+				out_channels		1
+				out_ch_cfg	$CHANNEL_CONFIG_MONO
+				out_ch_map	$CHANNEL_MAP_MONO
+			}
+			# 24-bit 48KHz 1ch
+			Object.Base.input_audio_format.8 {
+				in_channels		1
+				in_bit_depth		32
+				in_valid_bit_depth	32
+				in_ch_cfg	$CHANNEL_CONFIG_MONO
+				in_ch_map	$CHANNEL_MAP_MONO
+			}
+			Object.Base.output_audio_format.8 {
+				out_bit_depth		32
+				out_valid_bit_depth	24
+				out_channels		1
+				out_ch_cfg	$CHANNEL_CONFIG_MONO
+				out_ch_map	$CHANNEL_MAP_MONO
+			}
+			# 16-bit 48KHz 1ch
+			Object.Base.input_audio_format.9 {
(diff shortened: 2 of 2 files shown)


tools/topology/topology2/include/pipelines/cavs/host-gateway-capture.conf
```diff
@@ -105,6 +105,43 @@ Class.Pipeline."host-gateway-capture" {
...
+			Object.Base.audio_format.9 {
+				in_channels		1
+				in_bit_depth		32
+				in_valid_bit_depth	32
+				out_channels		1
+				in_ch_cfg	$CHANNEL_CONFIG_MONO
+				in_ch_map	$CHANNEL_MAP_MONO
+				out_ch_cfg	$CHANNEL_CONFIG_MONO
+				out_ch_map	$CHANNEL_MAP_MONO
+			}
```
plbossart, line 144: This is a generic change that impacts all solutions. Does this really work with e.g. HDaudio and DMIC?

I don't follow the logic where we would have to propagate all restrictions from a 3rd party codec all the way to the host-gateway and create a giant union of all requirements.

plbossart (changes requested):
not following this one either @bardliao

tools/topology/topology2/include/pipelines/cavs/host-gateway-capture.conf
ranj063: agree that this can't go in the class definition. Will need to go in the specific instantiation for the pipeline that needs this

bardliao: @plbossart @ranj063 But this is about extending the format capacity. The 1 ch format will be selected only if we do 1 channel capture. I.e `arecord -c 1`. Nothing will be different if we do `arecord -c 2`.

plbossart: some applications such as PulseAudio will try all possible hw_params.

if you expose support for 1ch, and it does not work on a specific platform, there's an issue.

We only want to expose 1ch for the platforms where it actually works.

bardliao: > some applications such as PulseAudio will try all possible hw_params.
> 
> if you expose support for 1ch, and it does not work on a specific platform, there's an issue.
> 
> We only want to expose 1ch for the platforms where it actually works.

@plbossart That is exactly the second commit does. We will set SDW_JACK_CAPTURE_CH = 1 in Cmakefile if the codec support 1 channel capture only. SDW_JACK_CAPTURE_CH is 2 by default which means the existing topologies support 2 channel format only. In other words, the supported formats doesn't change after the 1st commit extending the available formats. It is more like to add 1 channel format support in NHLT.

jairaj-arava: @udaymb @RDharageswari @yongzhi1

bardliao: > The first commit changes host-gateway-capture.conf, so it's not a SoundWire-only change at all @bardliao
> 
> This is changing EVERY host DMA.

Yes, but only one of the formats in host-gateway-capture.conf will be selected and the 1 channel formats will be selected only when we do 1 channel capture. And this will not happen unless we set channels_min/channels_max to 1 in PCM caps.

kv2019i: Ack, so as long as the pcm_caps is not change to by default expose 1-2 channel range to ALSA, this will have limited effect to other scenarios.

kv2019i (approved):

tools/topology/topology2/include/pipelines/cavs/host-gateway-capture.conf
bardliao: > agree that this can't go in the class definition. Will need to go in the specific instantiation for the pipeline that needs this

@ranj063 Sorry, I didn't get your point. Adding available formats in the class definition can benefit all pipelines that use the class. I didn't get why we need to add available formats in specific pipelines.

ranj063: @bardliao but the point is not all pipeline need to support 1ch capture

ranj063 (approved):

tools/topology/topology2/platform/intel/sdw-jack-generic.conf
```diff
@@ -81,11 +81,50 @@ Object.Pipeline {
...
 				stream_name	"Passthrough Capture 0"
 				pcm_id 1
+				num_input_audio_formats 9
+				num_output_audio_formats 9
 				Object.Base.audio_format.1 {
 					# 32/32 -> 16/16 bits conversion is done here
 					in_bit_depth	32
 					in_valid_bit_depth	32
 				}
+				# 32-bit 48KHz 1ch
```
ranj063, line 91: sorry clicked on approve before I realized this. We should not use the audio_format class anymore. Please split this into input_audio_format and output_audio_format objects please

bardliao: > sorry clicked on approve before I realized this. We should not use the audio_format class anymore. Please split this into input_audio_format and output_audio_format objects please

Done

kv2019i:
@bardliao rebase needed, it's probably the sdw passthrough PR that conflicts with this PR now.

bardliao:
> @bardliao rebase needed, it's probably the sdw passthrough PR that conflicts with this PR now.

done

tools/topology/topology2/cavs-sdw.conf
```diff
@@ -74,6 +74,7 @@ Define {
 	SDW_DMIC 0
 	SDW_JACK true
 	PASSTHROUGH	false
+	SDW_JACK_CAPTURE_CH 2
```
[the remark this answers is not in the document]
bardliao: > so to be clear, this PR does NOT enable a 1-ch capture. Will this be added later only for CS42L42 topologies?

Correct, this PR doesn't enable 1-ch capture. We will enable 1-ch capture ONLY for cs42l42 topologies.

plbossart (approved):

kv2019i:
One multiple-pause-resume fail in https://sof-ci.01.org/sofpr/PR7681/build9163/devicetest/index.html but not related to this PR. Proceeding with merge.

Issue #6576, opened 2022-11-10
[FEATURE] Clean-up topology2/cavs to make it user-friendly and maintainable


plbossart:
The topology2 framework allows for code reuse, modularity, macros and arithmetic to determine settings.

Sadly, our existing cavs topologies are 'developer' level and a number of ground rules were forgotten.

- [ ]  all controls and PCM names SHALL have a meaningful name. 

"volume playback 1' is not acceptable, each PGA and control shall be clearly identified with a pipeline information and position in the pipeline.
The use of 'master' SHALL be revisited. There SHALL not be more than one 'master' control on a path to an endpoint.

when possible, all PCM and control names SHALL be backwards compatible with existing UCM files. Don't be creative.

- [x] topology2/cavs files SHALL have the same level of modularity as topology1. We know e.g. for HDaudio/SoundWire/I2S that we'll have tons of variations (jack, no jack, echo reference, one amp, 2 amps, BT, HDMI with 3/4 streams). We have to track and make sure all variability in topology1 is supported in topology2

- [x]  the indices used for pipelines and components matter. in topology1 we could easily identify when a pipeline was connected to SSP0. This information is lost in topology2. so the rule is that just by using at the topology viewer, anyone SHALL be able to know which PCM device to use to play on a specific DAI.

Example of poor index management can be seen on this topology:
![cavs-tgl-nocodec](https://user-images.githubusercontent.com/1149035/201227404-28b07a39-0870-479e-9aef-90f7e0d51677.png)

copier.host.1.1 is not related to the PCM index.
copier.SSP.12.1 is not related to the DAI index.
All modules use '.1' at the end except the copier.module.8.2. It's unclear what all these 1s mean, and which we need a different convention for a copier?

- [x]  all generated topology2 files SHALL be named with the conventions used in the kernel. A topology file that is not known to the kernel SHALL be moved to a developer directory.  (Done)

Examples of bad names: cavs-mtl-sdw-dmic cavs-sdw-src-gain-mixin  cavs-mixin-mixout-hda-2ch-tgl cavs-tgl-nocodec cavs-passthrough-hdmi
Examples of good names: sof-mtl-nocodec
That's actually the only good name in the entire set of topology2 files....

Common sense really.

plbossart:
@lgirdwood @ranj063 @bardliao @marc-hb @yongzhi1  FYI

marc-hb:
I think examples would be great. Pointers to PRs fixing this sort of issues even better.

[...]

PR #7682, opened 2023-05-24
Rename cavs nocodec mixers

jsarha:
This tries to fix the first tap in https://github.com/thesofproject/sof/issues/6576 for cavs-nocodec topology.

### DIFF
--- a/tools/topology/topology2/cavs-nocodec.conf
+++ b/tools/topology/topology2/cavs-nocodec.conf
@@ -82,6 +82,9 @@ Define {
 	SSP0_MIXER_SOURCE_3		'mixin.21.1'
 	SSP1_ENABLED			"true"
 	PASSTHROUGH			"false"
+	SSP0_PCM_NAME			"Port0"
+	SSP1_PCM_NAME			"Port1"
+	SSP2_PCM_NAME			"Port2"
 }
 
 # override defaults with platform-specific config
@@ -180,7 +183,7 @@ IncludeByKey.PASSTHROUGH {
 				}
 				Object.Widget.gain.1 {
 					Object.Control.mixer.1 {
-						name 'Playback Volume 1'
+						name 'Pre Mixer $SSP0_PCM_NAME Playback Volume'
 					}
 				}
 			}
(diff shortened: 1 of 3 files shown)


tools/topology/topology2/cavs-nocodec.conf
```diff
@@ -267,7 +268,7 @@ IncludeByKey.PASSTHROUGH {
 				}
 				Object.Widget.gain.1 {
 					Object.Control.mixer.1 {
-						name 'Main Capture Volume 1'
+						name 'Post Mux $SSP0_PCM_NAME Capture Volume'
```
ranj063, line 273: @jsarha you're hitting the 44 char limit with this one right? Maybe it will be fixed when you removed the widget name from the name?

ranj063 (approved):
looks good to me

tools/topology/topology2/cavs-nocodec.conf
jsarha: Yes. I decided to ignore those cases that that that would be Ok without the widget name prefix. However, I am open for other suggestions for naming standard that would fit into smaller space, still be informative enough, and would fulfill this: https://docs.kernel.org/sound/designs/control-names.html

kv2019i:
@jsarha Btw, please always ensure there are enough reviewers in the PR. Add people who've edited the files recently, ask collegues, and what not.

jsarha:
> Welcome improvement! And thanks for the good examples in git commti messages. These helped a lot when reviewing.

Coming up with the examples will be harder in the sdw and rt5682 topologies (@plbossart , they are in scope of #6576, are't they?) as I do not have access to the HW in question?

tools/topology/topology2/cavs-nocodec.conf
```diff
@@ -340,7 +342,7 @@ IncludeByKey.PASSTHROUGH {
 						out_ch_map	$CHANNEL_MAP_3_POINT_1
 					}
 					Object.Control.mixer.1 {
-						name 'Main Capture Volume 3'
+						name 'Pre Mux $DMIC0_PCM_0_NAME Capture Volume'
```
plbossart, line 345: again where is there a MUX on DMIC capture?

plbossart: that's demux...

jsarha: ... Ok, will fix.

plbossart (approved):
thanks @jsarha

kv2019i:
Known fails only, proceeding with merge.

Issue #7677, opened 2023-05-24
[BUG] FW reported error: 104 - Other failure of module instance initialization request


keqiaozhang:
**Describe the bug**
This is a tplg regression when testing aplay on smart amp pipeline. This issue only happens on ADLP_SKU0B00_SDCA_IPC4ZPH.

Daily test run 26232?model=ADLP_SKU0B00_SDCA_IPC4ZPH&testcase=check-playback-100sec + most other tests on ADLP_SKU0B00_SDCA_IPC4ZPH

**dmesg**
```
[52479.099417] snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx      : 0x40020004|0x10036: MOD_INIT_INSTANCE [data size: 216]
[52479.099912] snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx reply: 0x60000068|0x10036: MOD_INIT_INSTANCE
[52479.099914] sof-audio-pci-intel-tgl 0000:00:1f.3: FW reported error: 104 - Other failure of module instance initialization request
[52479.099926] sof-audio-pci-intel-tgl 0000:00:1f.3: ipc error for msg 0x40020004|0x10036
[52479.099931] sof-audio-pci-intel-tgl 0000:00:1f.3: failed to create module alh-copier.SDW1-Playback.0
[52479.099982] snd_sof:sof_widget_free_unlocked: sof-audio-pci-intel-tgl 0000:00:1f.3: widget gain.21.1 freed
[52479.099987] snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx      : 0x12010000|0x0: GLB_DELETE_PIPELINE
[52479.100494] snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx reply: 0x32000000|0x0: GLB_DELETE_PIPELINE
[52479.100504] snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx done : 0x12010000|0x0: GLB_DELETE_PIPELINE
[52479.100508] snd_sof:sof_widget_free_unlocked: sof-audio-pci-intel-tgl 0000:00:1f.3: widget pipeline.21 freed
[...]

PR #7686, opened 2023-05-25
topology2: sdw-amp-generic: set virtual widget type output

bardliao:
Widget type snd_soc_dapm_out_drv and snd_soc_dapm_output will be treat as virtual widget and be ignored in sof topology in kernel. We need to set type = output or out_drv in topology when it is a virtual widget.

Fixes: 3835846836d4 ("topology2: sdw-amp-generic: Show all aggregated DAIs on the graph")

Fixes: #7677


ranj063 (approved):

aiChaoSONG (approved):

marc-hb:
Still not working?
- #7701

PR #7690, opened 2023-05-25
dai-zephyr: do not set local buffer if no process func found

kv2019i:
Rework set_new_local_buffer() to not set dd->local_buffer multiple times.

### DIFF
--- a/src/audio/dai-zephyr.c
+++ b/src/audio/dai-zephyr.c
@@ -1469,32 +1469,33 @@ int dai_zephyr_multi_endpoint_copy(struct dai_data **dd, struct comp_dev *dev,
 static void set_new_local_buffer(struct dai_data *dd, struct comp_dev *dev)
 {
 	struct comp_buffer __sparse_cache *dma_buf = buffer_acquire(dd->dma_buffer);
-	struct comp_buffer __sparse_cache *local_buf;
+	struct comp_buffer __sparse_cache *local_buf_c;
+	struct comp_buffer *local_buf;
 	uint32_t dma_fmt = audio_stream_get_frm_fmt(&dma_buf->stream);
 	uint32_t local_fmt;
 
 	buffer_release(dma_buf);
 
 	if (dev->direction == SOF_IPC_STREAM_PLAYBACK)
-		dd->local_buffer = list_first_item(&dev->bsource_list,
-						   struct comp_buffer,
-						   sink_list);
+		local_buf = list_first_item(&dev->bsource_list,
+					    struct comp_buffer,
+					    sink_list);
 	else
-		dd->local_buffer = list_first_item(&dev->bsink_list,
-						   struct comp_buffer,
-						   source_list);
+		local_buf = list_first_item(&dev->bsink_list,
+					    struct comp_buffer,
+					    source_list);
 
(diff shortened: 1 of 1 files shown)


kv2019i:
Not sure @lyakh this helps much, but we do avoid one extra dd->local_buffer write.

juimonen (approved):
not sure I understand the optimization, but looks innocent enough to me

src/audio/dai-zephyr.c
```diff
@@ -1468,32 +1468,33 @@ int dai_zephyr_multi_endpoint_copy(struct dai_data **dd, struct comp_dev *dev,
 static void set_new_local_buffer(struct dai_data *dd, struct comp_dev *dev)
 {
 	struct comp_buffer __sparse_cache *dma_buf = buffer_acquire(dd->dma_buffer);
+	struct comp_buffer __sparse_cache *local_buf_c;
 	struct comp_buffer __sparse_cache *local_buf;
```
lyakh, line 1472: ouch, this one isn't `__sparse_cache`, there must be an error in sparse test.

lyakh (changes requested):
looks good, thanks for fixing it! But need to fix one hick up.

lyakh:
@kv2019i @juimonen Given two possibilities:
```
	never_fails();
	ret = might_fail();
	if (ret < 0) {
		undo_never_fails();
		return ret;
	}
```
and
```
	ret = might_fail();
	if (ret < 0)
		return ret;

	never_fails();
```
as long as the order indeed isn't important - often both will work correctly, but I think one is clearly preferable over the other and should be used whenever possible.

src/audio/dai-zephyr.c
```diff
@@ -1468,32 +1468,33 @@ int dai_zephyr_multi_endpoint_copy(struct dai_data **dd, struct comp_dev *dev,
...
 
 	dd->process = pcm_get_conversion_function(local_fmt, dma_fmt);
 
-	if (!dd->process) {
+	if (dd->process)
+		dd->local_buffer = local_buf;
+	else
 		comp_err(dev, "converter function NULL: local fmt %d dma fmt %d\n",
 			 local_fmt, dma_fmt);
-		dd->local_buffer = NULL;
```
btian1, line 1495: even line 1478, 1482 set this, line 1495 still can clear this setting, right?
seems this is a code refine, are you finding this with bug or code inspection?

src/audio/dai-zephyr.c
kv2019i: Fixed in V2.

lyakh (approved):

dnikodem (approved):

kv2019i:
The Intel CI System/merge/build fails to one run of "TestSrcFullScope::test_01_04_src" indicating a glitch on one platform. I don't think it can be related to this PR, but need to do another run to be sure.

btian1 (approved):

tmleman (approved):

lgirdwood (approved):

lgirdwood:
SOFCI TEST

lgirdwood:
@kv2019i still failing CI, can you check. Thanks

PR #7691, opened 2023-05-25
aria: optimize in/out buffer copy

pblaszko:
Current Aria implementation use double for loop, copying separately every sample between circular and linear buffers. This solution is extremely slow and for some formats consumes even more than 10 MCPS just for data preparation.

Replacing for-loops with generic wrapped copy functions.

image

### DIFF
--- a/src/audio/aria/aria.c
+++ b/src/audio/aria/aria.c
@@ -304,9 +304,6 @@ static int aria_copy(struct comp_dev *dev)
 	struct aria_data *cd;
 	uint32_t source_bytes;
 	uint32_t sink_bytes;
-	uint32_t frag = 0;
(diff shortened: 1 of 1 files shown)


abonislawski (approved):

btian1:
quite straightforward, please provide before/after mcps data for this optimization.

pblaszko:
> quite straightforward, please provide before/after mcps data for this optimization.

added to description

tmleman (approved):

andrula-song (approved):

btian1 (approved):

kv2019i:
Known fails in sof-ci device tests, proceeding with merge.

PR #7693, opened 2023-05-25
src: mixin_mixout: Set channels during mixout_params

ranj063:
comp_verify_params() doesn't set the channels. So set it explicitly during params.


btian1 (approved):

juimonen (approved):
tests look ok so lgtm

kv2019i:
Known fails in sof-ci PR device tests, mandatory checks pass.

PR #7698, opened 2023-05-26
Audio: Multiband-DRC: Convert to module adapter

singalsu:


marc-hb:
Without firmware (due to the build failure) and combination of other serious CI issues,  https://sof-ci.01.org/sofpr/PR7698/build9948/devicetest uselessly hogged 3 MTL devices for more than 200 minutes :-(

singalsu:
> Without firmware (due to the build failure) and combination of other serious CI issues, https://sof-ci.01.org/sofpr/PR7698/build9948/devicetest uselessly hogged 3 MTL devices for more than 200 minutes :-(

Sorry for that, would [SKIP CI] in this PR title or in commit message title prevent CI to run? I've been thinking sharing early and often is good - had no idea the tests continue after failed build. There's a problem with component dependencies to crossover and drc those don't build correctly in current git main.

marc-hb:
> Sorry for that, would [SKIP CI] in this PR title or in commit message title prevent CI to run?

Yes, or the newer and clearer [SKIP SOF-TEST]. Both work the same.

> I've been thinking sharing early and often is good 

Yes and drafts are great for that.

> had no idea the tests continue after failed build. 

This is a CI design issue that is being worked on.

singalsu:
First working code version for IPC3, and IPC4.

singalsu:
> > First working code version for IPC3, and IPC4.
> 
> Still draft ?

Need to wait for crossover merge to remove the duplicated patch. Also will split this into three PRs with topology and tools separately.

Topology part needs DRC topology PR merge. I think I will add a configuration macro to select for EFX topology DRC vs. multiband-DRC. Multiband-DRC is more versatile but also it has a much higher load. Optimizations would be useful for this component.

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -6,9 +6,13 @@ CONFIG_INTEL_DMIC=n
 CONFIG_INTEL_DMIC_NHLT=n
 CONFIG_INTEL_SSP=n
 CONFIG_INTEL_ALH=n
+
+# enable additinal audio processing
 CONFIG_COMP_SRC=y
 CONFIG_COMP_SRC_IPC4_FULL_MATRIX=y
 CONFIG_COMP_DRC=y
+CONFIG_COMP_CROSSOVER=y
+CONFIG_COMP_MULTIBAND_DRC=y
```
singalsu, line 15: @lgirdwood @kv2019i Do you have opinion about this? If we are not having issues with .text size this might be acceptable. But if we should be more compact these could be default no. DRC is essential as are the EQs, these are more like optional for more advanced speaker enhance. What do you think?

lgirdwood: Lets have these in default Linux Kconfig.

src/audio/crossover/crossover.c
```diff
@@ -618,7 +618,7 @@ static int crossover_process_audio_stream(struct processing_module *mod,
 					  struct output_stream_buffer *output_buffers,
```
singalsu, line 618: Add Fixes: 5bcbcf1 ("Audio: Crossover: Convert to module adapter") to commit message.

singalsu: Still not good,

/zep_workspace/sof/src/audio/crossover/crossover.c:656:58: warning: incorrect type in argument 2 (different address spaces)

Seems avoiding sparse errors needs a different approach

lgirdwood (commented):
Does this need the ALSA kcontrol updates ?

singalsu:
> Does this need the ALSA kcontrol updates ?

IPC3 version works as before, with IPC4 the control is not available. It now works with IPC4 specific addition in module to force internally the processing switch on. The processing can be controlled on/off with user space sent bytes control blob (in #8049). Currently there is no other way.

I plan to submit another PR to make the blob controlled on/off switch as MCPS consumption efficient as in IPC3. Currently even in bypass the band split filters remain running that is not happening with switch control in IPC3. The idea is to turn the processing switch off if all band filters and emphasis/de-emphasis are switched off in the blob. What do you think of this @google ?

src/audio/crossover/crossover.c
```diff
@@ -170,14 +170,14 @@ static int crossover_assign_sinks(struct processing_module *mod,
...
-				 pipeline_id);
+				 sink_id);
 			break;
 		}
 
 		if (sinks[i]) {
 			comp_err(dev,
 				 "crossover_assign_sinks(), multiple sinks from pipeline %d are assigned",
-				 pipeline_id);
+				 sink_id);
```
lgirdwood, line 180: same string fix needed as previous PR ?

lgirdwood (commented):
Are we trending for making v2.7 code freeze ?

src/audio/crossover/crossover.c
singalsu: Yep, I need to rebase this and move the crossover fixes to other PR since they are no more critical for this component.

singalsu:
The new error with sparse cache in crossover.c was caused by regression in my other PR. The error appears with this PR since the required crossover filter is enabled in kconfig by this coponent. The fix for sparse cache is addressed in #8088.

lgirdwood (approved):

cujomalainey:
@singalsu i landed your other PR, please rebase this one.

lgirdwood (approved):

src/audio/multiband_drc/multiband_drc.c
```diff
@@ -358,291 +363,203 @@ static int multiband_drc_cmd_get_value(struct comp_dev *dev,
...
-		ret = comp_data_blob_set_cmd(cd->model_handler, cdata);
-		break;
-	default:
-		comp_err(dev, "multiband_drc_cmd_set_data() error: invalid cdata->cmd");
-		ret = -EINVAL;
-		break;
-	}
-
-	return ret;
+	comp_err(dev, "tdfb_cmd_get_value() error: invalid cdata->cmd");
```
kv2019i, line 368: tdfb, copy-paste?

kv2019i (approved):
One typoe in log print, maybe you can fix this up in a follow-up. Otherwise looks good.

PR #7699, opened 2023-05-28
[SKIP SOF-TEST] [TEST] [stable-v2.2] try pointing oss-fuzz at some older branch

marc-hb:
For context:
- #7675
- #7709

### DIFF
--- a/.github/workflows/ipc_fuzzer.yml
+++ b/.github/workflows/ipc_fuzzer.yml
@@ -30,12 +30,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Build Fuzzers
-        uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+        # Try some random, 2 months old branch
+        uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@links
         with:
           oss-fuzz-project-name: 'sound-open-firmware'
 
       - name: Run Fuzzers
-        uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+        uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@links
         with:
           oss-fuzz-project-name: 'sound-open-firmware'
           language: c


marc-hb:
@jonathanmetzman, @andyross, @cujomalainey : is there a way to use some older branch of oss-fuzz? Or is only the `master` branch usable?

In this test PR  I semi-randomly picked the existing branch https://github.com/google/oss-fuzz/commits/links. It's old enough for the purpose of this test.

In https://github.com/thesofproject/sof/actions/runs/5101622354/jobs/9170621114?pr=7699
I can see that the build seems to point at this older `links` branch in a number of places. HOWEVER:


```
2023-05-28 01:03:18,439 - root - INFO - Running: docker build -t gcr.io/oss-fuzz/sound-open-firmware 
   --file /opt/oss-fuzz/projects/sound-open-firmware/Dockerfile /opt/oss-fuzz/projects/sound-open-firmware.
```


The `/opt/oss-fuzz/projects/sound-open-firmware/Dockerfile` file used here is clearly NOT from the `links` branch (currently at commit 761128ec4121) because it starts by running  `apt-get install -y       gettext git libc6-dev-i386 libglib2.0-dev libncurses5-dev       libtool ninja-build python3-pip` which is from  much later with https://github.com/google/oss-fuzz/pull/10342/files. So it seems to be picked from `master` anyway. How come?

Thanks in advance!

cujomalainey:
@marc-hb I think that is a github-actions question

I still don't see the benefit here. This branch and IPC3 are mostly stable/archived. Hence anything we fuzz on main we should be able to cherry-pick down. Nothing new should be landing on this branch so I don't see the benefit of continued fuzzing past support on main.

marc-hb:
Thanks @cujomalainey.

> Hence anything we fuzz on main we should be able to cherry-pick down

But main is likely to become IPC type 4 only while stable-v2.2 is IPC type 3 only. Then cherry-picking won't work; too different.

> Nothing new should be landing on this branch so I don't see the benefit of continued fuzzing past support on main.

Fair enough! Could you please copy your last comment in a commit message, delete `.github/workflows/ipc_fuzzer.yml` in the stable-v2.2 branch and submit a PR! This should be enough to make CI on that branch green again; much appreciated.

PS: I'm still curious whether oss-fuzz branches can be used or not. A similar situation could happen in other projects; any project can branch.

marc-hb:
stable-v2.2 removal submitted in
- #7795

Issue #7249, opened 2023-03-10
[BUG] cavs platform_init should not write to DSP shim registers in Zephyr build


kv2019i:
**Is your feature request related to a problem? Please describe.**
Fix invalid hw access in platform_init() of intel/cavs platform_init in Zephyr builds.

See https://github.com/thesofproject/sof/issues/5602 and https://github.com/thesofproject/sof/issues/7248

This is not correct
```
»       io_reg_write(DSP_INIT_GENO,
»       »       GENO_MDIVOSEL | GENO_DIOPTOSEL);

»       io_reg_write(DSP_INIT_IOPO,
»       »       IOPO_DMIC_FLAG | IOPO_I2S_FLAG);

»       io_reg_write(DSP_INIT_ALHO,
»       »       ALHO_ASO_FLAG | ALHO_CSO_FLAG);

»       io_reg_write(DSP_INIT_LPGPDMA(0),
»       »       LPGPDMA_CHOSEL_FLAG | LPGPDMA_CTLOSEL_FLAG);
»       io_reg_write(DSP_INIT_LPGPDMA(1),
»       »       LPGPDMA_CHOSEL_FLAG | LPGPDMA_CTLOSEL_FLAG);
```

kv2019i:
Started working on this. Some of the registers are written in Zephyr platfom boot code, but not all, so we cannot simply remove this yet. I also didn't want to add ifdef ZEPHYR on individual register writes, but rather wait until we have this init done correctly in Zephyr main and then remove this. 

Also consider how to do this cleanly for intel/cavs. Once  @juimonen complete the runtime_pm refactoring, we are starting to get close to the point where we could add a thin "platform/cavs-zephyr" that gets rid of all platform specific code, or drop XTOS support altogether for intel/cavs in SOF main.

@lgirdwood @juimonen any thoughts?

lgirdwood:
@nashif fyi
@kv2019i I assume we have this flow in the Zephyr init and are executing it twice today ?

kv2019i:
@lgirdwood Nopes, we are missing bits on Zephyr side -> https://github.com/zephyrproject-rtos/zephyr/pull/55738

lgirdwood:
@kv2019i @juimonen any update ?

kv2019i:
Need to move to v2.7. I have a PR in Zephyr side, but I can't complete it in SOF2.6 timeline (Zephyr 3.4 feature freeze starts tomorrow).

PR #7702, opened 2023-05-29
platform: cavs: do not program GPDMA/ALH ownership in Zephyr builds [MANIFEST UDPATE!]

kv2019i:
If SOF is built with Zephyr, the GPDMA/ALH ownership programming should happen in Zephyr platform code, not on SOF side.

Link: https://github.com/zephyrproject-rtos/zephyr/pull/55738
Closes: https://github.com/thesofproject/sof/issues/7249

### DIFF
--- a/src/platform/intel/cavs/platform.c
+++ b/src/platform/intel/cavs/platform.c
@@ -336,6 +336,7 @@ static void platform_init_hw(void)
 	io_reg_write(DSP_INIT_IOPO,
 		IOPO_DMIC_FLAG | IOPO_I2S_FLAG);
 
+#ifndef __ZEPHYR__
 	io_reg_write(DSP_INIT_ALHO,
 		ALHO_ASO_FLAG | ALHO_CSO_FLAG);
 
@@ -343,6 +344,7 @@ static void platform_init_hw(void)
 		LPGPDMA_CHOSEL_FLAG | LPGPDMA_CTLOSEL_FLAG);
 	io_reg_write(DSP_INIT_LPGPDMA(1),
 		LPGPDMA_CHOSEL_FLAG | LPGPDMA_CTLOSEL_FLAG);
+#endif
 }
 
 /* Runs on the primary core only */
--- a/west.yml
+++ b/west.yml
@@ -45,7 +45,7 @@ manifest:
 
     - name: zephyr
       repo-path: zephyr
-      revision: 7d54586751cfe405d2cdcfd850f8100bc9844f63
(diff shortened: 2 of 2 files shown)


kv2019i:
Marking as DNM until Zephyr-side dependency is merged (to Zephyr).

juimonen (approved):
I think zephyr is merged, though we need west update for zephyr

kv2019i:
@juimonen wrote:
> I think zephyr is merged, though we need west update for zephyr

Yup, only a west update needed now, I could update this PR to do that.

I'm a bit tempted by your PR https://github.com/thesofproject/sof/pull/7706 to remove the platform for cavs altogether. The results are amazingly good! :) I wonder if the IMR state save is working correctly, maybe that's something we need to double-check, but that's one option. Or we merge this and then #7706 and keep more options open in case we need a a quick revert.

juimonen:
yeah also personally little bit suspicious about the #7706 results... I would still do some more testing with, exactly with pm/sleep/contextsave area...

kv2019i:
https://github.com/thesofproject/sof/pull/7706 is likely to make this PR no longer needed at all, but given that requires a bit more time to get the IMR boot implemetned, let's go ahead and proceed with this in the mean time.

kv2019i:
One PM fail in https://sof-ci.01.org/sofpr/PR7702/build8959/devicetest/index.html , otherwise looks good. Proceeding with merge.

marc-hb:
@keqiaozhang did this affect #7191?

src/platform/intel/cavs/platform.c
```diff
@@ -336,13 +336,15 @@ static void platform_init_hw(void)
...
+#ifndef __ZEPHYR__
 	io_reg_write(DSP_INIT_ALHO,
 		ALHO_ASO_FLAG | ALHO_CSO_FLAG);
 
 	io_reg_write(DSP_INIT_LPGPDMA(0),
 		LPGPDMA_CHOSEL_FLAG | LPGPDMA_CTLOSEL_FLAG);
 	io_reg_write(DSP_INIT_LPGPDMA(1),
 		LPGPDMA_CHOSEL_FLAG | LPGPDMA_CTLOSEL_FLAG);
+#endif
 }
```
dbaluta, line 349: @kv2019i did you guys moved the DMA drivers to zephyr? Care to let me know the path in zephyr where this happens?

kv2019i: @dbaluta Ack, all DMA drivers (HD-DMA and DW-DMA) moved now. They are in work/zephyr/drivers/dma/

PR #7707, opened 2023-05-30
ipc4: handler: chain_dma: Handle chain free when the chain_dma is not…

ujfalusi:
… created

Do not treat as error if the chain dma is asked to be destroyed if it has not been created yet.

The kernel might send such a request under special circumstances: the PCM is opened, configured but never started/triggered then just closed, freed.

This is the same way how pipelines are handled, but in case of pipelines we use the pipe state.


paulstelian97 (approved):

RanderWang (approved):

kv2019i:
Only one PM faill in https://sof-ci.01.org/sofpr/PR7707/build8645/devicetest/index.html , so looking good. Proceed with merge.

PR #7710, opened 2023-05-30
host: zephyr: Subtract partial_size from dma avail/free size.

softwarecki:
When the host component detects use of deep buffering it decreases frequency of dma reloading. The number of processed data is summed up in the partial_size variable and they are acknowledged in larger packets. However, the size of processed data was not subtracted from the number of available data, which could lead to a buffer underrun/overrun.


ranj063 (approved):
nice catch @softwarecki

btian1 (approved):

tmleman (approved):

dnikodem (approved):

abonislawski (approved):

PR #7711, opened 2023-05-30
intel: ssp: fix RX fifo drain logic

kv2019i:
Separate the RX drain functions for start and stop cases. For startup, do not add any delays but read the full fifo out and ensure the dma request line (SSSR.RFS) is cleared before new capture stream is started.

Link: https://github.com/thesofproject/sof/issues/7548
Fixes: 4a4d8d261907 ("intel: ssp: drain RX fifo when starting")

### DIFF
--- a/src/drivers/intel/ssp/ssp.c
+++ b/src/drivers/intel/ssp/ssp.c
@@ -67,6 +67,34 @@ static void ssp_empty_tx_fifo(struct dai *dai)
 		ssp_write(dai, SSSR, sssr);
 }
 
+/*
+ * Empty SSP receive FIFO before DMA startup and
+ * ensure the DMA request is not already raised when
+ * DMA is started. If TX is not active, this is a no-op.
+ */
+static void ssp_empty_rx_fifo_for_start(struct dai *dai)
+{
+	uint32_t retry = SSP_RX_FLUSH_RETRY_MAX;
+	uint32_t entries;
+	uint32_t i;
+
+	while ((ssp_read(dai, SSSR) & SSSR_RNE) && retry--) {
+		entries = SSCR3_RFL_VAL(ssp_read(dai, SSCR3));
+
+		/* let DMA consume data or read RX FIFO directly */
+		for (i = 0; i < entries + 1; i++)
+			ssp_read(dai, SSDR);
+
+		entries = SSCR3_RFL_VAL(ssp_read(dai, SSCR3));
+	}
+
+	/* clear interrupt */
+	ssp_update_bits(dai, SSSR, SSSR_ROR, SSSR_ROR);
+
+	if (ssp_read(dai, SSSR) & SSSR_RFS)
+		dai_warn(dai, "ssp_empty_rx_fifo_for_startup(), RFS=1 after flush!");
+}
+
 /* empty SSP receive FIFO */
 static void ssp_empty_rx_fifo(struct dai *dai)
 {
@@ -74,16 +102,9 @@ static void ssp_empty_rx_fifo(struct dai *dai)
 	uint64_t sample_ticks = clock_ticks_per_sample(PLATFORM_DEFAULT_CLOCK,
 						       ssp->params.fsync_rate);
 	uint32_t retry = SSP_RX_FLUSH_RETRY_MAX;
-	bool direct_reads = ssp->state[DAI_DIR_CAPTURE] <= COMP_STATE_PREPARE;
 	uint32_t entries;
-	uint32_t i;
-
 #if CONFIG_DMA_SUSPEND_DRAIN
-	/*
-	 * In drain mode, DMA is stopped before DAI, so flush must be
-	 * always done with direct register read.
-	 */
-	direct_reads = true;
+	uint32_t i;
 #endif
 
 	/*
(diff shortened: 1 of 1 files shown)


src/drivers/intel/ssp/ssp.c
```diff
@@ -98,10 +119,10 @@ static void ssp_empty_rx_fifo(struct dai *dai)
 		dai_dbg(dai, "ssp_empty_rx_fifo(), before flushing, entries %d", entries);
 
 		/* let DMA consume data or read RX FIFO directly */
-		if (direct_reads) {
-			for (i = 0; i < entries + 1; i++)
-				ssp_read(dai, SSDR);
-		}
+#if CONFIG_DMA_SUSPEND_DRAIN
+		for (i = 0; i < entries + 1; i++)
```
btian1, line 123: can remove 106 - 108, and use:
for ( int i = 0; i < entries + 1; i++)

btian1 (commented):
this is a follow up #7577, one more question is: the change only have impact on BT or all SSP related operation? should be all, right?

src/drivers/intel/ssp/ssp.c
```diff
@@ -67,23 +67,44 @@ static void ssp_empty_tx_fifo(struct dai *dai)
...
+static void ssp_empty_rx_fifo_for_start(struct dai *dai)
+{
+	uint32_t retry = SSP_RX_FLUSH_RETRY_MAX;
+	uint32_t entries;
+	uint32_t i;
+
+	while ((ssp_read(dai, SSSR) & SSSR_RNE) && retry--) {
+		entries = SSCR3_RFL_VAL(ssp_read(dai, SSCR3));
+
+		/* let DMA consume data or read RX FIFO directly */
```
ujfalusi, line 84: the comment is not correct.

src/drivers/intel/ssp/ssp.c
```diff
@@ -67,23 +67,44 @@ static void ssp_empty_tx_fifo(struct dai *dai)
...
+		for (i = 0; i < entries + 1; i++)
+			ssp_read(dai, SSDR);
+
+		entries = SSCR3_RFL_VAL(ssp_read(dai, SSCR3));
```
ujfalusi, line 88: no need to re-read the entries here.

src/drivers/intel/ssp/ssp.c
```diff
@@ -67,23 +67,44 @@ static void ssp_empty_tx_fifo(struct dai *dai)
...
+	}
+
+	/* clear interrupt */
+	ssp_update_bits(dai, SSSR, SSSR_ROR, SSSR_ROR);
```
ujfalusi, line 92: Clear Receive FIFO overrun

src/drivers/intel/ssp/ssp.c
```diff
@@ -67,23 +67,44 @@ static void ssp_empty_tx_fifo(struct dai *dai)
...
+
+	if (ssp_read(dai, SSSR) & SSSR_RFS)
+		dai_warn(dai, "ssp_empty_rx_fifo_for_startup(), RFS=1 after flush!");
+}
+
 /* empty SSP receive FIFO */
 static void ssp_empty_rx_fifo(struct dai *dai)
```
ujfalusi, line 99: if the other incarnation is `ssp_empty_rx_fifo_for_start()` then this is `ssp_empty_rx_fifo_for_stop()`?

src/drivers/intel/ssp/ssp.c
```diff
@@ -98,10 +119,10 @@ static void ssp_empty_rx_fifo(struct dai *dai)
...
 
 		/* let DMA consume data or read RX FIFO directly */
-		if (direct_reads) {
-			for (i = 0; i < entries + 1; i++)
-				ssp_read(dai, SSDR);
-		}
+#if CONFIG_DMA_SUSPEND_DRAIN
+		for (i = 0; i < entries + 1; i++)
+			ssp_read(dai, SSDR);
+#endif
```
ujfalusi, line 125: I'm not certain about this. Even if the DMA is active (not disabled) but the RX FIFO does not have trigger amount of data then the DMA will not clear the FIFO, so we just spin w/o any reason.

src/drivers/intel/ssp/ssp.c
kv2019i: Yeah, that would be nicer, but for now, we've stick to Linux kernel style and declare variable at start. I know we start to have some C++ inspired code that is not following this anymore...

src/drivers/intel/ssp/ssp.c
kv2019i: @ujfalusi Ack, so this will not drain the samples that fall below the DMA trigger threshold and the function will timeout (which at 8kHz is a long timeout). OTOH, in old code fifo is emptied directly and DMA is left without trailing data, so not great either.

Vamshigopal:
@kv2019i  i've tested the patch with noraml scenario and with no-codec tplg with 8k rate. Issue is not reproducible.

kv2019i:
Excellent. thanks @Vamshigopal ! @ujfalusi is working on a revised patch we are currently stress testing on multiple devices. This builds on this same fix, but addresses some of the issues raised in this review. We'll share this later for review.

src/drivers/intel/ssp/ssp.c
```diff
@@ -67,23 +67,44 @@ static void ssp_empty_tx_fifo(struct dai *dai)
 		ssp_write(dai, SSSR, sssr);
 }
 
+/*
+ * Empty SSP receive FIFO before DMA startup and
+ * ensure the DMA request is not already raised when
+ * DMA is started. If TX is not active, this is a no-op.
```
lyakh, line 73: I'm confused a bit: isn't receive FIFO usually the FIFO where data arrives from the interface, i.e. for capture? Why is it talking about TX then?

Vamshigopal:
I've encountered a strange fimding while testing this patch.

Issue: No sound from playback (BT output ) 
Issue scenario:

steps:
1. start yt playback from Btoutput 
2. start recording from Btinput
3. stop recording from btinput
4. repeat 2,3 for couple of times
5. reboot
6. start yt playback from btoutput

In step 6 we dont see any sound from btoutput.

kv2019i:
Closing, let's go with #7717  instead.

PR #7712, opened 2023-05-30
topology2: intel: dmic-generic: Simplify the core ID setting

ranj063:
When all components of a pipeline are scheduled on the same core, use the pipeline's core_id attribute to set the core_id for all the components in it. Also, remove the default core_id in the module-copier and eqiir class definitions so that it doesn't take priority over the value passed from the pipeline object.

### DIFF
--- a/tools/topology/topology2/platform/intel/dmic-generic.conf
+++ b/tools/topology/topology2/platform/intel/dmic-generic.conf
@@ -76,10 +76,9 @@ IncludeByKey.PASSTHROUGH {
 			{
 				format		$FORMAT
 				index 		$DMIC0_HOST_PIPELINE_ID
-
+				core_id		$DMIC_CORE_ID
 				Object.Widget.host-copier.1 {
 					stream_name $DMIC0_PCM_CAPS
-					core_id $DMIC_CORE_ID
 					pcm_id	$DMIC0_PCM_ID
 					num_input_audio_formats 2
 					num_output_audio_formats 2
@@ -103,7 +102,6 @@ IncludeByKey.PASSTHROUGH {
 					}
 				}
 				Object.Widget.gain.1 {
-					core_id $DMIC_CORE_ID
 					num_input_audio_formats 2
 					num_output_audio_formats 2
 					Object.Base.audio_format.1 {
@@ -130,7 +128,6 @@ IncludeByKey.PASSTHROUGH {
 				}
 
 				Object.Widget.module-copier."2" {
-					core_id $DMIC_CORE_ID
 					num_input_audio_formats 2
 					num_output_audio_formats 2
 					Object.Base.audio_format.1 {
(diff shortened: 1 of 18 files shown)


tools/topology/topology2/platform/intel/dmic-generic.conf
```diff
@@ -76,10 +76,9 @@ IncludeByKey.PASSTHROUGH {
 			{
 				format		$FORMAT
 				index 		$DMIC0_HOST_PIPELINE_ID
-
+				core_id		$DMIC_CORE_ID
 				Object.Widget.host-copier.1 {
 					stream_name $DMIC0_PCM_CAPS
-					core_id $DMIC_CORE_ID
```
plbossart, line 82: @ranj063 for my education, how does the pipeline-level core_id get mirrored in each module instance? is this something done by the topology parser or something done by the driver+firmware?

ranj063: @plbossart it done by the topology compiler. All children objects automatically inherit their parent object's attribute values unless they have them set explicitly too.

plbossart: what happens if the base class for an object has a different value to that of the object parent? Who wins? The parent or the base class?

ranj063: Base class

ranj063: The topology2 documentation has the information on the order of priority for where the value is inferred from.

plbossart: I am sure everyone has read that documentation, right? I wouldn't even know where to find a published version.

juimonen (approved):
lgtm, although touching many other files than dmic-generic (in the commit title)

yongzhi1 (approved):

plbossart (approved):

kv2019i (changes requested):
All dmic capture tests fail in CI, please check.

ranj063:
> All dmic capture tests fail in CI, please check.

@kv2019i i was testing with core ID 222 just be able to spot it easily in the conf file and lef it in by mistake. FIxed now. tests should pass

tools/topology/topology2/platform/intel/dmic-generic.conf
ranj063: @plbossart there's preliminary documentation for topology2but we're going to enhance it to include all recently introduced features. Please stay tuned.

kv2019i:
Apparently we did not have 200+ cores. ;) That was a good smoke test. This version now passes tests. Only known PM fail in https://sof-ci.01.org/sofpr/PR7712/build8776/devicetest/index.html , proceeding with merge.

kv2019i (approved):

PR #7716, opened 2023-05-31
[cavs2.5-001-drop-stable] intel: ssp: TX and RX FIFO drain fixes

ujfalusi:
Hi,

both the TX and RX FIFO drain handling have issues currently, both fixed by this PR.

On TX side we regularly see:
```
c2 wait                         src/lib/wait.c:52   ERROR poll timeout reg 487496 mask 63 val 0 us 937
c2 ssp-dai      1.0   /drivers/intel/ssp/ssp.c:71   WARN ssp_empty_tx_fifo() warning: timeout
```
The reason is that the 'calculated' timeout is incorrect for low sampling  rates.

On RX side the FIFO draining needs to follow different strategies depending on when the draining is done to avoid starting and stopping issues.

This PR can replace #7711, so far it is holding up to the local stress tests.


ujfalusi:
created against wrong branch....

PR #7717, opened 2023-05-31
[stable-v2.2] intel: ssp: RX FIFO drain fix (try 2)

ujfalusi:
Hi,

both the TX and RX FIFO drain handling have issues currently, both fixed by this PR.

On TX side we regularly see:
```
c2 wait                         src/lib/wait.c:52   ERROR poll timeout reg 487496 mask 63 val 0 us 937
c2 ssp-dai      1.0   /drivers/intel/ssp/ssp.c:71   WARN ssp_empty_tx_fifo() warning: timeout
```
The reason is that the 'calculated' timeout is incorrect for low sampling rates.

On RX side the FIFO draining needs to follow different strategies depending on when the draining is done to avoid starting and stopping issues.

**Note:** the TX side is not touched in this PR

This PR can replace https://github.com/thesofproject/sof/pull/7711, so far it is holding up to the local stress tests.

### DIFF
--- a/src/drivers/intel/ssp/ssp.c
+++ b/src/drivers/intel/ssp/ssp.c
@@ -67,49 +67,74 @@ static void ssp_empty_tx_fifo(struct dai *dai)
 		ssp_write(dai, SSSR, sssr);
 }
 
-/* empty SSP receive FIFO */
-static void ssp_empty_rx_fifo(struct dai *dai)
+static void ssp_empty_rx_fifo_on_start(struct dai *dai)
+{
+	uint32_t retry = SSP_RX_FLUSH_RETRY_MAX;
+	uint32_t i, sssr;
+
+	sssr = ssp_read(dai, SSSR);
+
+	if (sssr & SSSR_ROR) {
+		/* The RX FIFO is in overflow condition, empty it */
+		for (i = 0; i < SSP_FIFO_DEPTH; i++)
+			ssp_read(dai, SSDR);
+
+		/* Clear the overflow status */
+		ssp_update_bits(dai, SSSR, SSSR_ROR, SSSR_ROR);
+		/* Re-read the SSSR register */
+		sssr = ssp_read(dai, SSSR);
+	}
+
+	while ((sssr & SSSR_RNE) && retry--) {
+		uint32_t entries = SSCR3_RFL_VAL(ssp_read(dai, SSCR3));
+
+		/* Empty the RX FIFO (the DMA is not running at this point) */
+		for (i = 0; i < entries + 1; i++)
+			ssp_read(dai, SSDR);
+
+		sssr = ssp_read(dai, SSSR);
+	}
+}
+
+static void ssp_empty_rx_fifo_on_stop(struct dai *dai)
 {
 	struct ssp_pdata *ssp = dai_get_drvdata(dai);
 	uint64_t sample_ticks = clock_ticks_per_sample(PLATFORM_DEFAULT_CLOCK,
 						       ssp->params.fsync_rate);
 	uint32_t retry = SSP_RX_FLUSH_RETRY_MAX;
-	bool direct_reads = ssp->state[DAI_DIR_CAPTURE] <= COMP_STATE_PREPARE;
-	uint32_t entries;
-	uint32_t i;
+	uint32_t entries[2];
+	uint32_t i, sssr;
 
-#if CONFIG_DMA_SUSPEND_DRAIN
-	/*
-	 * In drain mode, DMA is stopped before DAI, so flush must be
-	 * always done with direct register read.
-	 */
-	direct_reads = true;
-#endif
+	sssr = ssp_read(dai, SSSR);
+	entries[0] = SSCR3_RFL_VAL(ssp_read(dai, SSCR3));
 
-	/*
-	 * To make sure all the RX FIFO entries are read out for the flushing,
-	 * we need to wait a minimal SSP port delay after entries are all read,
-	 * and then re-check to see if there is any subsequent entries written
-	 * to the FIFO. This will help to make sure there is no sample mismatched
-	 * issue for the next run with the SSP RX.
-	 */
-	while ((ssp_read(dai, SSSR) & SSSR_RNE) && retry--) {
-		entries = SSCR3_RFL_VAL(ssp_read(dai, SSCR3));
-		dai_dbg(dai, "ssp_empty_rx_fifo(), before flushing, entries %d", entries);
+	while ((sssr & SSSR_RNE) && retry--) {
+		/* Wait one sample time */
+		wait_delay(sample_ticks);
 
-		/* let DMA consume data or read RX FIFO directly */
-		if (direct_reads) {
-			for (i = 0; i < entries + 1; i++)
+		entries[1] = SSCR3_RFL_VAL(ssp_read(dai, SSCR3));
+		sssr = ssp_read(dai, SSSR);
+
+		if (entries[0] > entries[1]) {
+			/*
+			 * The DMA is reading the FIFO, check the status in the
+			 * next loop
+			 */
+			entries[0] = entries[1];
+		} else if (!(sssr & SSSR_RFS)) {
+			/*
+			 * The DMA request is not asserted, read the FIFO
+			 * directly, otherwise let the next loop iteration to
+			 * check the status
+			 */
+			for (i = 0; i < entries[1] + 1; i++)
 				ssp_read(dai, SSDR);
 		}
 
-		/* wait to get valid fifo status and re-check */
-		wait_delay(sample_ticks);
-		entries = SSCR3_RFL_VAL(ssp_read(dai, SSCR3));
-		dai_dbg(dai, "ssp_empty_rx_fifo(), after flushing, entries %d", entries);
+		sssr = ssp_read(dai, SSSR);
 	}
 
-	/* clear interrupt */
+	/* Just in case clear the overflow status */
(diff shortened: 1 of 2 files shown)


src/drivers/intel/ssp/ssp.c
```diff
@@ -74,49 +74,74 @@ static void ssp_empty_tx_fifo(struct dai *dai)
...
+{
+	uint32_t retry = SSP_RX_FLUSH_RETRY_MAX;
+	uint32_t i, sssr;
+
+	sssr = ssp_read(dai, SSSR);
+
+	if (sssr & SSSR_ROR) {
+		/* The RX FIFO is in overflow condition, empty it */
+		for (i = 0; i < 32; i++)
+			ssp_read(dai, SSDR);
```
kv2019i, line 87: Is this safe. The maximum FIFO length is 16 right?

src/drivers/intel/ssp/ssp.c
```diff
@@ -74,49 +74,74 @@ static void ssp_empty_tx_fifo(struct dai *dai)
...
+			 * next loop
+			 */
+			entries[0] = entries[1];
+		} else if (!(sssr & SSSR_RFS)) {
+			/*
+			 * The DMA request is not asserted, read the FIFO
+			 * directly, otherwise let the next loop iteration to
+			 * check the status
+			 */
+			for (i = 0; i < 32; i++)
```
kv2019i, line 137: Same here, I believe 32 is the maximum but on hw covered by this driver, the fifo depth is 16.

src/drivers/intel/ssp/ssp.c
```diff
@@ -1154,8 +1179,6 @@ static int ssp_probe(struct dai *dai)
 	/* Disable dynamic clock gating before touching any register */
 	pm_runtime_get_sync(SSP_CLK, dai->index);
 
-	ssp_empty_rx_fifo(dai);
-
 	return 0;
```
kv2019i, line 1182: This is a bit of mystery, but seems safe and reasonable thing to remove.

kv2019i (commented):
Looks great, thanks @ujfalusi . One comment on the fifo length (inline).

kv2019i:
I ran the stress tests I used with #7711 and this PR also passes it (1500+ rounds). Plus no tx timeout warnings now with this set.

src/drivers/intel/ssp/ssp.c
ujfalusi: It appears to be 32 deep on TGL at least (upx-i11).
```diff
diff --git a/src/drivers/intel/ssp/ssp.c b/src/drivers/intel/ssp/ssp.c
index a75f8752ed0a..cb0b076ebc37 100644
--- a/src/drivers/intel/ssp/ssp.c
+++ b/src/drivers/intel/ssp/ssp.c
@@ -81,7 +81,15 @@ static void ssp_empty_rx_fifo_on_start(struct dai *dai)
 
 	sssr = ssp_read(dai, SSSR);
 
+	dai_info(dai, "ssp_empty_rx_fifo_on_start(): initial SSRSA = 0x%08x",
+		 ssp_read(dai, SSRSA));
+	dai_info(dai, "ssp_empty_rx_fifo_on_start(): initial SSSR = 0x%08x",
+		 ssp_read(dai, SSSR));
+	dai_info(dai, "ssp_empty_rx_fifo_on_start(): initial entries = %u",
+		 (uint32_t)SSCR3_RFL_VAL(ssp_read(dai, SSCR3)));
+
 	if (sssr & SSSR_ROR) {
+		dai_info(dai, "ssp_empty_rx_fifo_on_start(): yanking FIFO");
 		/* The RX FIFO is in overflow condition, empty it */
 		for (i = 0; i < 32; i++)
 			ssp_read(dai, SSDR);
@@ -90,17 +98,30 @@ static void ssp_empty_rx_fifo_on_start(struct dai *dai)
 		ssp_update_bits(dai, SSSR, SSSR_ROR, SSSR_ROR);
 		/* Re-read the SSSR register */
 		sssr = ssp_read(dai, SSSR);
+		dai_info(dai, "ssp_empty_rx_fifo_on_start(): entries after yank = %u",
+			 (uint32_t)SSCR3_RFL_VAL(ssp_read(dai, SSCR3)));
 	}
 
+	dai_info(dai, "ssp_empty_rx_fifo_on_start(): before loop: SSSR = 0x%08x", sssr);
 	while ((sssr & SSSR_RNE) && retry--) {
 		uint32_t entries = SSCR3_RFL_VAL(ssp_read(dai, SSCR3));
 
+		dai_info(dai,
+			 "ssp_empty_rx_fifo_on_start(): retry round: %d: entries: %u",
+			 SSP_RX_FLUSH_RETRY_MAX - retry, entries);
 		/* Empty the RX FIFO (the DMA is not running at this point) */
 		for (i = 0; i < entries + 1; i++)
 			ssp_read(dai, SSDR);
 
 		sssr = ssp_read(dai, SSSR);
+		dai_info(dai,
+			 "ssp_empty_rx_fifo_on_start(): end of retry round: %d: ssr: 0x%08x",
+			 SSP_RX_FLUSH_RETRY_MAX - retry, sssr);
 	}
+
+	sssr = ssp_read(dai, SSSR);
+	dai_info(dai,
+		 "ssp_empty_rx_fifo_on_start(): after all of this: SSSR = 0x%08x", sssr);
 }
 
 static void ssp_empty_rx_fifo_on_stop(struct dai *dai)
```
With this PR (FIFO depth is 32):
```
/drivers/intel/ssp/ssp.c:84   INFO ssp_empty_rx_fifo_on_start(): initial SSRSA = 0x00000001
/drivers/intel/ssp/ssp.c:86   INFO ssp_empty_rx_fifo_on_start(): initial SSSR = 0x000000dc
/drivers/intel/ssp/ssp.c:88   INFO ssp_empty_rx_fifo_on_start(): initial entries = 0
/drivers/intel/ssp/ssp.c:92   INFO ssp_empty_rx_fifo_on_start(): yanking FIFO
/drivers/intel/ssp/ssp.c:101  INFO ssp_empty_rx_fifo_on_start(): entries after yank = 0
/drivers/intel/ssp/ssp.c:105  INFO ssp_empty_rx_fifo_on_start(): before loop: SSSR = 0x00000014
/drivers/intel/ssp/ssp.c:123  INFO ssp_empty_rx_fifo_on_start(): after all of this: SSSR = 0x00000014
```
Using `SSP_FIFO_DEPTH`, which is defined to 16 (so FIFO depth of 16):
```
/drivers/intel/ssp/ssp.c:84   INFO ssp_empty_rx_fifo_on_start(): initial SSRSA = 0x00000001
/drivers/intel/ssp/ssp.c:86   INFO ssp_empty_rx_fifo_on_start(): initial SSSR = 0x000000dc
/drivers/intel/ssp/ssp.c:88   INFO ssp_empty_rx_fifo_on_start(): initial entries = 0
/drivers/intel/ssp/ssp.c:92   INFO ssp_empty_rx_fifo_on_start(): yanking FIFO
/drivers/intel/ssp/ssp.c:101  INFO ssp_empty_rx_fifo_on_start(): entries after yank = 16
/drivers/intel/ssp/ssp.c:105  INFO ssp_empty_rx_fifo_on_start(): before loop: SSSR = 0x0000005c
/drivers/intel/ssp/ssp.c:109  INFO ssp_empty_rx_fifo_on_start(): retry round: 1: entries: 16
/drivers/intel/ssp/ssp.c:117  INFO ssp_empty_rx_fifo_on_start(): end of retry round: 1: ssr: 0x00000014
/drivers/intel/ssp/ssp.c:123  INFO ssp_empty_rx_fifo_on_start(): after all of this: SSSR = 0x0000001c
```
In a later case (assuming depth of 16 for the FIFO) after the yanking of the FIFO we still have 16 entries left (which is an odd result as the number of entries in FIFO is +1, so 17?).
It looks awfully like that the FIFO is 32 deep by this experiment. I'm fine to use the two step yanking assuming a 16 word deep FIFO, at the end it should not matter much.

src/drivers/intel/ssp/ssp.c
ujfalusi: Yes, `entries[1] + 1` is the correct counter

Vamshigopal:
I've encountered a strange fimding while testing this patch.

Issue: No sound from playback (BT output ) 
Issue scenario:

steps:
1. start yt playback from Btoutput 
2. start recording from Btinput
3. stop recording from btinput
4. repeat 2,3 for couple of times
5. reboot
6. start yt playback from btoutput

In step 6 we dont see any sound from btoutput.

ujfalusi:
Changes since v1:
- Drop the TX fifo wait fix patch due to sideffect
- Correct the FIFO depth for Tiger Lake platforms (it is 32 instead of 16)
- Use the `SSP_FIFO_DEPTH` for yanking loop
- Use the remaining slots in stop yanking loop

The FIFO depth has been verified with a full FIFO and reading out with 4 entry steps, it is really 32 deep.

The TX FIFO wait patch did a wrong calculation for the timeout  (too long time to wait) but it did made sure that the TX FIFO is drained. This caused a side effect on capture side also because the wait might be too long for the scheduling and we were missing ticks. It is still not clear _why_ we are waiting for the TX FIFO to drain, it has not benefits at all but it will lead to TX underflow right away.

kv2019i (approved):

kv2019i:
Thi sis ready to go, just need a second review. @lyakh @juimonen @plbossart @ranj063

juimonen (approved):

kv2019i:
@Vamshigopal Ready to go with this version? The others PRs will likely require a longer testing/review cycle as they might impact more use-cases.

PR #6056, opened 2022-07-28
Audio: Enhancing audio quality tests for device testing

ShriramShastry:
For the purpose of remote device testing, this check-in adds a correction to the scripts that evaluate gain,
frequency responsiveness, total harmonic distortion plus noise, and dynamic range.

The threshold scale has been defined and the decibel full scale computation has been made easier using new scripts.

You require Octave or Matlab. The scripts now operate properly in both Matlab and octave.

The user account, audio device, audio format, and capture sound card audio device and audio format for the remote device must be modified in the configuration file sof test perf config.m.

Signed-off-by: shastry <malladi.sastry@intel.com>

PR #7719, opened 2023-05-31
Tools: Test: Audio: Valgrind run related fixes

singalsu:
The changes to scripts/host-testbench.sh have unintentionally dropped valgrind run from test.

This patch enables valgrind for process_test.m runs and fixes issue in test run octave side function to silently ignore error about failed shell command. Normally test fail when there is no data or incorrect data, but a valgrind failure with correct test output was passed.

Valgrind output becomes visible if testbench run trace stderr redirection to file is disabled. If not done valgrind error would stop the test but the analysis would not be printed to console.

### DIFF
--- a/tools/test/audio/process_test.m
+++ b/tools/test/audio/process_test.m
@@ -409,6 +409,9 @@ function test = test_defaults(t)
 	test.thdnf_max = [];            % Set per component
 	test.dr_db_min = 80;            % Min. DR
 	test.fr_rp_max_db = 0.5;        % Allow 0.5 dB frequency response ripple
+
+	% No need to collect trace
+	test.trace = '';
 end
 
 function test = test_run_process(test)
--- a/tools/test/audio/test_utils/test_run.m
+++ b/tools/test/audio/test_utils/test_run.m
@@ -92,14 +92,16 @@ else
 	fprintf(fh, 'XTRUN=\n');
 end
 
-% Override defaults in comp_run.sh
-fprintf(fh, 'VALGRIND=false\n');
 fclose(fh);
 
 arg = sprintf('-t %s', fn_config);
 rcmd = sprintf('%s %s', ex, arg);
 fprintf('Running ''%s''...\n', rcmd);
-system(rcmd);
+ret = system(rcmd);
+if ret
+	error('''%s'' returned status %d\n', rcmd, ret);
+end
+
 delete(fn_config);
 
 end
--- a/tools/test/audio/test_utils/test_run.m
+++ b/tools/test/audio/test_utils/test_run.m
@@ -55,7 +55,7 @@ if isfield(test, 'nch_in') == 0
 	test.nch_in = test.nch;
 end
 
-fprintf(fh, '#!/bin/sh\n', test.comp);
+fprintf(fh, '#!/bin/sh\n');
 fprintf(fh, 'COMP=\"%s\"\n', test.comp);
 fprintf(fh, 'DIRECTION=playback\n');
 fprintf(fh, 'BITS_IN=%d\n', test.bits_in);


tools/test/audio/test_utils/test_run.m
```diff
@@ -92,14 +92,17 @@
...
-fprintf(fh, 'VALGRIND=false\n');
 fclose(fh);
 
 arg = sprintf('-t %s', fn_config);
 rcmd = sprintf('%s %s', ex, arg);
 fprintf('Running ''%s''...\n', rcmd);
-system(rcmd);
+ret = system(rcmd);
+if ret
+	fprintf('Returned status %d\n', ret);
```
marc-hb, line 102: Because rcmd may print stuff too:

```suggestion
printf('%s eturned status %d\n', rcmd, ret);
```
Can't you combine `fprintf` and `error` into one? I hope one does not print on stdout while the other on stderr because buffering could order them randomly.

marc-hb (commented):
Should there be a `Fixes commit abcdedf` in the commit message?

tools/test/audio/test_utils/test_run.m
singalsu: Re-read the manual, today error in octave can do same as printf... and so does Matlab. It was an old limitation that no more bothers. Good that you asked!

btian1 (approved):

btian1 (approved):

tools/test/audio/test_utils/test_run.m
```diff
@@ -92,14 +92,16 @@
...
 
 arg = sprintf('-t %s', fn_config);
 rcmd = sprintf('%s %s', ex, arg);
 fprintf('Running ''%s''...\n', rcmd);
-system(rcmd);
+ret = system(rcmd);
+if ret
+	error('''%s'' returned status %d\n', rcmd, ret);
+end
+
```
ShriramShastry, line 104: At line number 58, `fprintf(fh, '#!/bin/sh\n', test.comp);` Isn't the output format strange?

tools/test/audio/process_test.m
```diff
@@ -409,6 +409,9 @@
 	test.thdnf_max = [];            % Set per component
 	test.dr_db_min = 80;            % Min. DR
 	test.fr_rp_max_db = 0.5;        % Allow 0.5 dB frequency response ripple
+
+	% No need to collect trace
+	test.trace = '';
```
ShriramShastry, line 414: At line number 101, `tn = 1;`, tn is not used; please consider using it or removing it.

At line number 321, `min_bits = min(test.bits_in, test.bits_out);`, do you intend to use minimum bits elsewhere ? Otherwise, it's unnecessary code that should be removed.

ShriramShastry (commented):
Please have a look at my comments, which I've added.

Thank you very much.

tools/test/audio/process_test.m
```diff
@@ -409,6 +409,9 @@
 	test.thdnf_max = [];            % Set per component
```
kv2019i, line 409: Minor nit: the "Fixes commit-id" tag not quite following Linux style. 

```
If your patch fixes a bug in a specific commit, e.g. you found an issue using
``git bisect``, please use the 'Fixes:' tag with the first 12 characters of
the SHA-1 ID, and the one line summary.  Do not split the tag across multiple
lines, tags are exempt from the "wrap at 75 columns" rule in order to simplify
parsing scripts.  For example::

»       Fixes: 54a4f0239f2e ("KVM: MMU: make kvm_mmu_zap_page() return the number of pages it actually freed")                                          
```

kv2019i (approved):

tools/test/audio/process_test.m
marc-hb: There are git scripts that scan branches for that specific `Fixes:` syntax.

tools/test/audio/process_test.m
marc-hb: @ShriramShastry how is this related to this PR?

tools/test/audio/test_utils/test_run.m
marc-hb: Good catch: it does look like a bug but how is it related to this PR?

tools/test/audio/process_test.m
ShriramShastry: It has nothing to do with PR and does not rely on PR to succeed. A little correction can help greatly improve code quality, and I believe Seppo won't mind (because he wrote the majority of the code) considering the proposal.

marc-hb: These look like interesting suggestions but they're still off-topic here.

ShriramShastry: Hmm - sure it's off the  topic, what are your thoughts if you notice improvement and believe that rectification would be beneficial without adding extra work?  I am not sure if the chosen approach is OK or NOT, and for the PR, I request Seppo to consider but leave the decision to him.

singalsu: Thanks for suggestions. They are better to leave to another PR later. If you like you are welcome to make a PR since you found them. Or I can do later when time, as you prefer.

tools/test/audio/process_test.m
singalsu: The commit message is now fixed. You may want to re-review because of 2nd commit added to fix the syntax error found by Sriram. And strange that it didn't error.

ShriramShastry (approved):
Thanks for making the change; they now appear good to me.

tools/test/audio/process_test.m
ShriramShastry: I would want to propose that you incorporate it as well as as you can take over [https://github.com/thesofproject/sof/pull/6056](url).
Thank you!!

marc-hb: https://docs.zephyrproject.org/latest/contribute/contributor_expectations.html

PR #7726, opened 2023-06-01
[IPC4] Enable core dump

RanderWang:
Enable core dump on TGL & MTL. When FW panic happen, zephyr will dump panic information in debug window, then Linux kernel can get this dump info


RanderWang:
@kv2019i Very sorry that I missed "arch_system_halt" in your original patch. I can't recall the reason.

juimonen (approved):
lgtm, just probable typo in "copules time"

dnikodem (approved):

kv2019i:
@RanderWang I think you need to declare the shutdown somehow:

https://github.com/thesofproject/sof/actions/runs/5142711830/jobs/9257018685?pr=7726
```
 D:/a/sof/sof/workspace/sof/zephyr/wrapper.c: In function 'k_sys_fatal_error_handler':
D:/a/sof/sof/workspace/sof/zephyr/wrapper.c:339:9: error: implicit declaration of function 'arch_system_halt' [-Werror=implicit-function-declaration]
  339 |         arch_system_halt(reason);
```

I wonder if there's some more public function to do the same.

Otherwise looks we are ready to go.

zephyr/wrapper.c
```diff
@@ -336,4 +336,5 @@ void k_sys_fatal_error_handler(unsigned int reason,
 	ipc_send_panic_notification();
 
 	LOG_ERR("Halting system");
+	arch_system_halt(reason);
```
btian1, line 339: possible set to debug only?

app/boards/intel_adsp_cavs25.conf
```diff
@@ -8,3 +8,7 @@ CONFIG_INTEL_ALH=y
 CONFIG_LP_MEMORY_BANKS=1
 CONFIG_HP_MEMORY_BANKS=30
 CONFIG_RIMAGE_SIGNING_SCHEMA="tgl"
+
+CONFIG_DEBUG_COREDUMP=y
+CONFIG_DEBUG_COREDUMP_BACKEND_INTEL_ADSP_MEM_WINDOW=y
+CONFIG_DEBUG_COREDUMP_MEMORY_DUMP_MIN=y
```
btian1, line 14: debug feature should set to debug only build, release build should not have.
maybe we should have another set of config for debug build only.

zephyr/wrapper.c
RanderWang: maybe it is used for waiting GDB attach

app/boards/intel_adsp_cavs25.conf
RanderWang: Coredump only happens with FW panic and it doesn't affect performance. It is used for debugging FW panic not for debug build

RanderWang:
@kv2019i  fixed. Thanks!

app/boards/intel_adsp_cavs25.conf
btian1: can I understand that all the work will be done from zephyr side, and sof itself does not have code related with core dump?

zephyr/wrapper.c
kv2019i: We are overriding the Zephyr panic handlers here, so we do need to halt the system here, whether it's debug or not.

app/boards/intel_adsp_cavs25.conf
kv2019i: This is intented for release builds. We will not have logs enabled in release builds, so this coredump capability is the only way to get coredumps into kernel logs. Similar functionality has been enabled in all past (IPC3 based) SOF releases.

kv2019i (approved):

app/boards/intel_adsp_cavs25.conf
RanderWang: > can I understand that all the work will be done from zephyr side, and sof itself does not have code related with core dump?

Most part of work is done in zephyr. SOF will only send a notification IPC message to host.

tmleman (approved):

kv2019i:
Known fails in https://sof-ci.01.org/sofpr/PR7726/build8950/devicetest/index.html and https://sof-ci.01.org/sofpr/PR7726/build8951/devicetest/index.html

marc-hb:
mtrace still incomplete in https://sof-ci.01.org/softestpr/PR1057/build444/devicetest/index.html?model=TGLU_RVP_SDW_IPC4ZPH&testcase=multiple-pause-resume-5

PR #7731, opened 2023-06-02
audio: avoid division when calculating samples and frames

lyakh:
Bytes per sample and bytes per frame are usually powers of 2, so it bitshift can be used instead of division to calculate frame and sample counts. Note, that this doesn't necessarily improve performance, our measurements haven't showed any significant difference to the original code, but rather this is a clean up, achieved by replacing open-coded divisions spread around the code by calls to one of `audio_stream_bytes_to_samples()` and `audio_stream_bytes_to_frames()`

### DIFF
--- a/src/audio/dai-zephyr.c
+++ b/src/audio/dai-zephyr.c
@@ -1500,7 +1500,6 @@ static void set_new_local_buffer(struct dai_data *dd, struct comp_dev *dev)
 /* copy and process stream data from source to sink buffers */
 int dai_common_copy(struct dai_data *dd, struct comp_dev *dev, pcm_converter_func *converter)
 {
-	uint32_t dma_fmt;
 	uint32_t sampling;
 	struct comp_buffer __sparse_cache *buf_c;
 	struct dma_status stat;
@@ -1509,6 +1508,7 @@ int dai_common_copy(struct dai_data *dd, struct comp_dev *dev, pcm_converter_fun
 	uint32_t copy_bytes;
 	uint32_t src_samples;
 	uint32_t sink_samples;
+	uint32_t period_samples;
 	uint32_t samples = UINT32_MAX;
 	int ret;
 
@@ -1535,16 +1535,15 @@ int dai_common_copy(struct dai_data *dd, struct comp_dev *dev, pcm_converter_fun
 
 	buf_c = buffer_acquire(dd->dma_buffer);
 
-	dma_fmt = audio_stream_get_frm_fmt(&buf_c->stream);
-	sampling = get_sample_bytes(dma_fmt);
-
-	buffer_release(buf_c);
+	sampling = audio_stream_sample_bytes(&buf_c->stream);
+	period_samples = audio_stream_bytes_to_samples(&buf_c->stream, dd->period_bytes);
 
 	/* handle module runtime unbind */
 	if (!dd->local_buffer) {
 		set_new_local_buffer(dd, dev);
 
 		if (!dd->local_buffer) {
+			buffer_release(buf_c);
 			comp_warn(dev, "dai_zephyr_copy(): local buffer unbound, cannot copy");
 			return 0;
 		}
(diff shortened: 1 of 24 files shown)


src/include/sof/audio/audio_stream.h
```diff
@@ -118,6 +121,18 @@ static inline enum sof_ipc_frame audio_stream_get_frm_fmt(
 	return buf->frame_fmt;
 }
 
+static inline enum sof_ipc_frame audio_stream_get_fmt_shift(
```
serhiy-katsyuba-intel, line 124: Why does this return enum sof_ipc_frame? Probably, it should return int8_t or signed char.

src/audio/host-zephyr.c
```diff
@@ -398,18 +398,20 @@ static uint32_t host_get_copy_bytes_normal(struct host_data *hd, struct comp_dev
...
-	dma_sample_bytes = get_sample_bytes(dma_buf_c->stream.frame_fmt);
+	dma_sample_bytes = get_sample_bytes(audio_stream_get_frm_fmt(&dma_buf_c->stream));
 	buffer_release(dma_buf_c);
 
 	buffer_c = buffer_acquire(buffer);
 
 	/* calculate minimum size to copy */
 	if (dev->direction == SOF_IPC_STREAM_PLAYBACK) {
-		avail_samples = (dma_stat.pending_length - hd->partial_size) / dma_sample_bytes;
+		avail_samples = audio_stream_bytes_to_samples(&dma_buf_c->stream,
```
serhiy-katsyuba-intel, line 408: dma_buf_c was released on line 402, so we can't access its ->stream data here.

src/include/sof/audio/audio_stream.h
```diff
@@ -73,6 +73,9 @@ struct audio_stream {
 	 */
 	uint16_t frame_align_shift;
 
+	signed char fmt_shift;	/**< Use to calculate samples */
```
lgirdwood, line 76: there is already a frame_shift_align above, can we reuse ? I would also include in the coment to "avoid divisions"

src/include/sof/audio/audio_stream.h
lyakh: copy-paste, will fix, thanks!

src/audio/dai-zephyr.c
```diff
@@ -1554,13 +1554,13 @@ int dai_zephyr_copy(struct dai_data *dd, struct comp_dev *dev, pcm_converter_fun
...
 		src_samples = audio_stream_get_avail_samples(&buf_c->stream);
-		sink_samples = free_bytes / sampling;
+		sink_samples = audio_stream_bytes_to_samples(&buf_c->stream, free_bytes);
 		samples = MIN(src_samples, sink_samples);
 		buffer_release(buf_c);
 	} else {
 		struct list_item *sink_list;
 
-		src_samples = avail_bytes / sampling;
+		src_samples = audio_stream_bytes_to_samples(&buf_c->stream, avail_bytes);
```
serhiy-katsyuba-intel, line 1563: Same here, dd->dma_buffer has been already released.

src/audio/dai-zephyr.c
```diff
@@ -1602,7 +1602,8 @@ int dai_zephyr_copy(struct dai_data *dd, struct comp_dev *dev, pcm_converter_fun
 	 * if FAST_MODE is enabled, then one period limitation is omitted
 	 */
 	if (!(dd->ipc_config.feature_mask & BIT(IPC4_COPIER_FAST_MODE)))
-		samples = MIN(samples, dd->period_bytes / sampling);
+		samples = MIN(samples,
+			      audio_stream_bytes_to_samples(&buf_c->stream, dd->period_bytes));
```
serhiy-katsyuba-intel, line 1606: sampling in original code is DMA buffer sample size. Here buf_c is just a temporary variable used to acquire different buffers: DMA buffer or local buffer depending of direction. At this line on code we do not know to what buffer buf_c is pointing. Another thing, is that buf_c was already relesed.

singalsu:
On TGL UPX, git main https://github.com/thesofproject/sof/commit/6fb22d645ef673034efdcdc1dd7bc94197db5d59 is 20.6 MCPS, with this patch 21.0 MCPS.

Is someone able to test with MTL and HiFi4?

lyakh:
> On TGL UPX, git main [6fb22d6](https://github.com/thesofproject/sof/commit/6fb22d645ef673034efdcdc1dd7bc94197db5d59) is 20.6 MCPS, with this patch 21.0 MCPS.

@singalsu kind of strance, also because CI almost all tests failed https://sof-ci.01.org/sofpr/PR7731/build8859/devicetest/index.html But I have a (supposedly) better version now, let's see what CI says about it

src/audio/dai-zephyr.c
```diff
@@ -1554,13 +1554,13 @@ int dai_zephyr_copy(struct dai_data *dd, struct comp_dev *dev, pcm_converter_fun
 	if (dev->direction == SOF_IPC_STREAM_PLAYBACK) {
 		buf_c = buffer_acquire(dd->local_buffer);
 		src_samples = audio_stream_get_avail_samples(&buf_c->stream);
-		sink_samples = free_bytes / sampling;
+		sink_samples = audio_stream_bytes_to_samples(&buf_c->stream, free_bytes);
```
btian1, line 1557: personally, I don't suggest have such code, for integer divide, in compiler and its inside lib, there is also a shift operation(its own fast algorithm), it is not complicated, and don't waste too much cycles, however, with new added function and inside if, it do need more cycles.

singalsu:
@lyakh Please fix the conflict

lyakh:
as verified by @singalsu (thanks!) this provides a very marginal performance improvement, not far outside the error margin, while replacing simple and easy to understand divisions with (inline) function calls. So, I'm not convinced we want this? @lgirdwood ?

btian1:
> 

check my comments two weeks ago:

personally, I don't suggest have such code, for integer divide, in compiler and its inside lib, there is also a shift operation(its own fast algorithm), it is not complicated, and don't waste too much cycles, however, with new added function and inside if, it do need more cycles.

Once you have switch like code, normal cycle cost > 5, and a fast divide normally take similar, no need to do this further, I also dropped other PRs, like: #7766

andrula-song:
Did you test the performance? I'm not sure the "function call +  judging condition + shift" would be better than a division

lyakh:
> Did you test the performance? I'm not sure the "function call + judging condition + shift" would be better than a division

@andrula-song performance has been tested by @singalsu and he came up with a *very* small improvement, practically negligible. So, it isn't really a performance improvement, but rather a clean up. All functions are inline, so shouldn't really have the overhead. Yes, there is one `if` now in each of the conversions. We could eliminate them by using function pointers, but then of course we'd have real function calls instead of inlines. So, no, it isn't really a performance improvement, but just a replacement of all the open-coded divisions by one of the two new functions `audio_stream_bytes_to_samples()` and `audio_stream_bytes_to_frames()`

PR #7732, opened 2023-06-02
[stable-v2.2 - DNM] intel: ssp: DMA request management

ujfalusi:
Hi,

the first two patch is #7717 due to dependency.

This PR changes how the DMA request is managed (or not managed currently):
instead of keeping both TX and RX DMA request active, only keep the one enabled which is in use.

This PR has been verified on Chrome and it is running an aplay + arecrod start/stop cycle for 1200 cycles at the time when the PR is created


ujfalusi:
Changes since v1:
- drop the TX FIFO maximum timeout correction patch, it looks like that waiting for the FIFO to really drain can have side-effects.

kv2019i:
#7717 merged, this can be rebased

PR #7733, opened 2023-06-02
Coherent: Do not calculate cache_to_uncache for non-shared

singalsu:


singalsu:
On TGL UPX, git main https://github.com/thesofproject/sof/commit/6fb22d645ef673034efdcdc1dd7bc94197db5d59 is 20.6 MCPS, with this patch 20.5 MCPS. Tiny 0.1 MCPS saving with this.

src/audio/module_adapter/module_adapter.c
```diff
@@ -828,19 +828,21 @@ static int module_adapter_simple_copy(struct comp_dev *dev)
...
-	uint32_t min_free_frames = UINT_MAX;
-	int ret, i = 0;
+	size_t size;
+	uint32_t min_free_frames;
+	int ret, i;
 
 	comp_dbg(dev, "module_adapter_copy(): start");
 
 	if (mod->simple_copy)
 		return module_adapter_simple_copy(dev);
```
btian1, line 842: then module_adapter_simple_copy can be static inline, this can save one more 0.1

src/audio/module_adapter/module_adapter.c
```diff
@@ -828,19 +828,21 @@ static int module_adapter_simple_copy(struct comp_dev *dev)
...
 	struct comp_buffer *source, *sink;
-	struct comp_buffer __sparse_cache *sink_c = NULL;
+	struct comp_buffer __sparse_cache *sink_c;
 	struct list_item *blist;
-	size_t size = MAX(mod->deep_buff_bytes, mod->period_bytes);
-	uint32_t min_free_frames = UINT_MAX;
-	int ret, i = 0;
+	size_t size;
+	uint32_t min_free_frames;
+	int ret, i;
```
kv2019i, line 837: Hmm, not sure about this. First, I wonder if the compiler can already optimize this out? And if not, wouldn't it be easier to split module_adapter_copy into module_adapter_copy_complex() which is called if simple_copy is not used.

kv2019i (commented):
The changes are functionally ok, but I'm not sure about the first patch.

src/audio/module_adapter/module_adapter.c
btian1: this change make code looks strange and divided, although may save few cycles for stack init.
if really want, can use two functions with if else.

singalsu: Yes, the saving is tiny, need to re-think this, or close since saving is so small.

singalsu:
Changing module_adapter_simple_copy() to static inline did not improve performance at all. So I dropped the module adapter related commit since the initializes moving for not simple copy was not approved. This change to coherent functions is not providing measurable improvement for average MCPS, but in theory should be better.

lyakh (approved):

btian1 (approved):

PR #7735, opened 2023-06-02
[stable-v2.2 DNM] intel: ssp: Do not wait for TX FIFO drain on stop

ujfalusi:
Hi,

this PR is on top of #7732 (which is on top of #7717).

In my opinion there is no reason to wait for the TX FIFO to drain on stop:
- if it was only playback, the SSP is going to be disabled
- if there is an ongoing capture then the FIFO will drain
- if the codec is clock provider then there could be a case when the I2S clocks are missing and we wait for the drain -> timout

Local and product testing looks good with the PR.


ujfalusi:
Closing, replaced by https://github.com/thesofproject/sof/pull/7741

PR #7738, opened 2023-06-02
[STABLE-v2.6] versions.json: bump to 2.6.0

marc-hb:
First 2.6.0 version on the stable-v2.6 branch


kv2019i (approved):
I was looking at git history to see where this should be done, but couldn't find this on Friday :)

juimonen (approved):

PR #7740, opened 2023-06-05
Copier: preparation work for convert copier to module interface

btian1:


src/audio/copier/copier.c
```diff
@@ -575,20 +555,69 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
...
+	if (!dev)
+		goto error_dev;
+
+	dev->ipc_config = *config;
+
+	cd = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*cd));
+	if (!cd)
+		goto error_cd;
+
+	comp_set_drvdata(dev, mod);
```
ranj063, line 598: @btian1 if you do this here, you have to change all the comp_get_drv_data() in the code to use 
struct copier_data *cd = module_get_private_data(mod);

btian1: I know, will modify all of them later.

src/audio/copier/copier.c
```diff
@@ -122,46 +121,82 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
...
 	dev->direction = cd->direction;
 	dev->state = COMP_STATE_READY;
-	return dev;
-
-error_cd:
-	rfree(cd);
+	return 0;
 error:
-	rfree(dev);
+	return ret;
```
lyakh, line 156: `rfree(cd);`

src/audio/copier/copier.c
```diff
@@ -122,46 +121,82 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
...
+
+	comp_set_drvdata(dev, mod);
+	md = &mod->priv;
+	mod->dev = dev;
+	dst = &md->cfg;
+	dst->init_data = spec;
+
+	ret = copier_init(mod);
+	if (!ret)
+		return dev;
```
lyakh, line 189: `rfree(dev);`

src/audio/copier/copier.c
```diff
@@ -567,6 +566,33 @@ static int do_multi_endpoint_module_copy(struct copier_data *cd, struct comp_dev
 			}
 		}
 	}
+	if (!ret) {
+		comp_update_buffer_consume(src_c, processed_data.source_bytes);
+		if (!cd->endpoint_num || cd->bsource_buffer)
+			cd->input_total_data_processed += processed_data.source_bytes;
+	}
```
lyakh, line 573: these 5 lines are now executed twice - here and in `do_module_copy()` below?

src/audio/copier/copier.c
```diff
@@ -54,49 +55,46 @@ DECLARE_SOF_RT_UUID("copier", copier_comp_uuid, 0x9ba00c83, 0xca12, 0x4a83,
...
+	struct module_data *md = &mod->priv;
+	struct module_config *dst = &md->cfg;
+	struct ipc4_copier_module_cfg *copier = (struct ipc4_copier_module_cfg *)dst->init_data;
+	struct comp_ipc_config *config = &dev->ipc_config;
+	int i, ret = 0;
 
 	cd = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*cd));
-	if (!cd)
+	if (!cd) {
+		ret = -ENOMEM;
```
ranj063, line 73: you could simply return -ENOMEM here instead of goto?

src/audio/copier/copier.c
```diff
@@ -771,22 +771,53 @@ static int set_attenuation(struct comp_dev *dev, uint32_t data_offset, const cha
...
+				    uint32_t data_offset_size,
+				    const uint8_t *fragment, size_t fragment_size,
+				    uint8_t *response,
+				    size_t response_size)
+{
+	struct module_data *md = &mod->priv;
+	struct comp_dev *dev = mod->dev;
+	int ret;
+
+	comp_dbg(dev, "volume_set_config()");
```
ranj063, line 904: volume?

src/audio/copier/copier.c
```diff
@@ -771,22 +771,53 @@ static int set_attenuation(struct comp_dev *dev, uint32_t data_offset, const cha
...
+
+	dcache_invalidate_region((__sparse_force void __sparse_cache *)fragment, fragment_size);
+
+	ret = module_set_configuration(mod, config_id, pos, data_offset_size, fragment,
```
ranj063, line 908: why do you need this?

src/audio/copier/copier.c
```diff
@@ -871,18 +900,30 @@ static int copier_get_large_config(struct comp_dev *dev, uint32_t param_id,
...
+static int copier_get_large_config(struct comp_dev *dev, uint32_t param_id,
+				   bool first_block,
+				   bool last_block,
+				   uint32_t *data_offset,
+				   char *data)
+{
+	struct processing_module *mod = comp_get_drvdata(dev);
+
+	return copier_get_configuration(mod, param_id, data_offset,
+					data, 0);
```
ranj063, line 1042: will it all fit in one line?

src/audio/copier/copier.c
```diff
@@ -630,7 +657,11 @@ static int copier_copy(struct comp_dev *dev)
 		break;
 	}
 	/* handle multi-endpoint and module copy */
-	return do_multi_endpoint_module_copy(cd, dev);
+	ret = do_multi_endpoint_copy(cd, dev);
```
ranj063, line 747: can we move this inside case SOF_COMP_DAI: for the multiendpoint case?

src/audio/copier/copier.c
```diff
@@ -567,6 +566,33 @@ static int do_multi_endpoint_module_copy(struct copier_data *cd, struct comp_dev
...
+{
+	struct comp_buffer *src, *sink;
+	struct comp_buffer __sparse_cache *src_c, *sink_c;
+	struct comp_copy_limits processed_data;
+	struct list_item *sink_list;
+	int ret  = 0;
+
+	processed_data.source_bytes = 0;
+
+	if (cd->endpoint_num && !cd->bsource_buffer) {
```
ranj063, line 590: for the module copier, will cd->endpoint_num ever be > 0?

src/audio/copier/copier.c
```diff
@@ -625,6 +625,59 @@ static int do_module_copy(struct copier_data *cd, struct comp_dev *dev)
...
+		 */
+		do_conversion_adapter_copy(cd, source_s, sink_s, &processed_data,
+					   i, mod->input_buffers[0].size);
+
+		mod->output_buffers[i].size = processed_data.sink_bytes;
+		cd->output_total_data_processed += processed_data.sink_bytes;
+	}
+
+	mod->input_buffers[0].consumed = processed_data.source_bytes;
+	if (!cd->endpoint_num || cd->bsource_buffer)
```
ranj063, line 675: endpoint_num will always be 0 for module copier

ranj063: and instead of checking for bsource_buffer, I'd check the cd->direction

src/audio/copier/copier.c
```diff
@@ -636,6 +689,40 @@ static int do_module_copy(struct copier_data *cd, struct comp_dev *dev)
  * from single stream (for output gateways) so such gateways work kind of like a single
  * gateway, i.e., produce/consume single stream.
  */
+
+static int copier_process(struct processing_module *mod,
```
ranj063, line 693: this will likely run into compilation issues because this function is unused. Either you need to call this from copier_copy() or you need convert to module_interface in this patch

src/audio/copier/copier.c
```diff
@@ -772,22 +772,47 @@ static int set_attenuation(struct comp_dev *dev, uint32_t data_offset, const cha
...
+				    const uint8_t *fragment, size_t fragment_size,
+				    uint8_t *response,
+				    size_t response_size)
+{
+	struct module_data *md = &mod->priv;
+	struct comp_dev *dev = mod->dev;
+
+	comp_dbg(dev, "copier_set_config()");
+
+	dcache_invalidate_region((__sparse_force void __sparse_cache *)fragment, fragment_size);
```
lyakh, line 812: I'm wondering why some `.set_configuration()` methods invalidate cache and others don't? If it's needed, maybe it should be done centrally in `module_set_large_config()`? @ranj063

src/audio/copier/copier.c
lyakh: sorry, not sure I understand your comment correctly, maybe you could rephrase? And in general - is this commit incomplete? I don't understand which function is converted to the module adapter API here?

src/audio/copier/copier.c
```diff
@@ -578,45 +523,39 @@ static int do_multi_endpoint_copy(struct copier_data *cd, struct comp_dev *dev)
...
+static int do_module_process_copy(struct copier_data *cd, struct comp_dev *dev,
+				  struct input_stream_buffer *input_buffers,
+				  int num_input_buffers,
+				  struct output_stream_buffer *output_buffers,
+				  int num_output_buffers)
 
-	processed_data.source_bytes = 0;
+{
+	struct audio_stream __sparse_cache *source_s;
+	struct audio_stream __sparse_cache *sink_s;
```
lyakh, line 534: what does the `_s` suffix stand for? We usually use `_c` for "cached," what do you mean here?

btian1: When I made the change, source_c and sink_c was occupied, so use _s, means stream cached.
I can change back, this is not big deal, current blocker is how to call copier trigger, may need a new interface or add it into copier, it may cause extra cycles.

ranj063 (approved):

RanderWang (approved):

src/audio/copier/copier.c
lgirdwood: any update here ?

lgirdwood (commented):
@btian1 can you check CI thanks.

src/audio/copier/copier.c
```diff
@@ -643,7 +643,7 @@ static int copier_copy(struct comp_dev *dev)
 	switch (dev->ipc_config.type) {
 	case SOF_COMP_HOST:
 		if (!cd->ipc_gtw)
-			return do_endpoint_copy(dev);
```
ranj063, line 560: @btian1 if you're going to do this, I think we can go further and remove do_enpoint_copy() altogether isnt it?

src/audio/copier/copier.c
ranj063: @lgirdwood this piece of code is not repeated twice anymore.

src/audio/copier/copier.c
```diff
@@ -122,46 +120,83 @@ static struct comp_dev *copier_new(const struct comp_driver *drv,
...
+	if (!dev)
+		goto error_dev;
+
+	dev->ipc_config = *config;
+
+	comp_set_drvdata(dev, mod);
+	md = &mod->priv;
+	mod->dev = dev;
+	dst = &md->cfg;
+	dst->init_data = spec;
```
lyakh, line 184: how about just `md->cfg.init_data = spec` and remove `dst`. Or even `mod->priv.cfg.init_data = spec` and remove both `md` and `dst`

src/audio/copier/copier.c
```diff
@@ -629,12 +639,13 @@ static int copier_copy(struct comp_dev *dev)
...
-static int copier_params(struct comp_dev *dev, struct sof_ipc_stream_params *params)
+static int copier_mod_params(struct processing_module *mod)
 {
-	struct processing_module *mod = comp_get_drvdata(dev);
+	struct sof_ipc_stream_params *params = mod->stream_params;
 	struct copier_data *cd = module_get_private_data(mod);
-	int i, ret = 0;
+	struct comp_dev *dev = mod->dev;
+	int ret = 0;
+	int i;
```
lyakh, line 649: what was wrong with the original line 637?

src/audio/copier/copier.c
```diff
@@ -639,16 +650,20 @@ static int copier_copy(struct comp_dev *dev)
 	case SOF_COMP_HOST:
 		if (!cd->ipc_gtw)
 			return do_endpoint_copy(dev);
-		break;
+
+		/* do nothing in the gateway copier case */
```
lyakh, line 628: is this a change in behaviour? It used to call `do_multi_endpoint_module_copy()` here before?

lyakh (commented):
mostly looks good to me, just one doubt about a potential change in behaviour and a couple of minor comments

src/audio/copier/copier.c
btian1: In assembly level, there is no difference at all for both implementation.

src/audio/copier/copier.c
btian1: logic changed.
host copy is neither multi_endpoint nor module copy, right? so directly return for host and do nothing for gtw case.

src/audio/copier/copier.c
btian1: you are right, nothing changed, let me change back, don't remember why have this change.

src/audio/copier/copier.c
```diff
@@ -528,13 +528,49 @@ static int do_conversion_copy(struct comp_dev *dev,
...
+static int copier_copy_to_sinks(struct copier_data *cd, struct comp_dev *dev,
+				struct comp_buffer __sparse_cache *src_c,
+				struct comp_copy_limits *processed_data)
+{
+	struct comp_buffer __sparse_cache *sink_c;
+	struct list_item *sink_list;
+	struct comp_buffer *sink;
+	int ret = 0;
+
+	list_for_item(sink_list, &dev->bsink_list) {
```
lgirdwood, line 512: Large code block - can we have an inline comment on what we are doing for each list item.

src/audio/copier/copier.c
ranj063: @lyakh its not really a change in behaviour. It does the same thing in all cases, just in a simplified way

src/audio/copier/copier.c
```diff
@@ -54,49 +55,45 @@ DECLARE_SOF_RT_UUID("copier", copier_comp_uuid, 0x9ba00c83, 0xca12, 0x4a83,
...
-	comp_cl_dbg(&comp_copier, "copier_new()");
-
-	dev = comp_alloc(drv, sizeof(*dev));
-	if (!dev)
-		return NULL;
-
-	dev->ipc_config = *config;
+	struct comp_dev *dev = mod->dev;
+	struct module_data *md = &mod->priv;
+	struct module_config *dst = &md->cfg;
```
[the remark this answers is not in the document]
btian1: ok, line 68 already have config named, for avoid confusing, I removed this variable.

src/audio/copier/copier.c
btian1: @lyakh @lgirdwood , use md->cfg.init_data = spec to remove dst usage, mod->priv.cfg.init_data = spec this is a little long, I prefer use md->cfg.init_data = spec.

src/audio/copier/copier.c
btian1: this code is actually moved out from do_multi_endpoint_module_copy, it is module copy part, I am adding simple comments for explain it.

src/audio/copier/copier.c
lyakh: > In assembly level, there is no difference at all for both implementation.

Sure, but we aren't programming in assembly. There are many ways to write code in C to generate the same assembly output. You can name your variables `a1`, `a2` etc. but that's not how we're programming

btian1: fully agree, please review the changes, and let me know if you still have further comments.

src/audio/copier/copier.c
lyakh: @ranj063 I'm not very familiar with the IPC gateway flow, so I might be wrong. Before this change for the IPC gateway case `do_multi_endpoint_module_copy()` was called, then in it the `if (cd->endpoint_num && !cd->bsource_buffer)` condition would be satisfied and `do_endpoint_copy()` would be called? In it it would then detect that `.multi_endpoint_buffer` isn't set so it would call
```
		return cd->endpoint[0]->drv->ops.copy(cd->endpoint[0]);
```
?

ranj063: > @ranj063 I'm not very familiar with the IPC gateway flow, so I might be wrong. Before this change for the IPC gateway case `do_multi_endpoint_module_copy()` was called, then in it the `if (cd->endpoint_num && !cd->bsource_buffer)` condition would be satisfied and `do_endpoint_copy()` would be called? In it it would then detect that `.multi_endpoint_buffer` isn't set so it would call
> 
> ```
> 		return cd->endpoint[0]->drv->ops.copy(cd->endpoint[0]);
> ```
> 
> ?

but @lyakh in do_endpoint)copy() for the ipc_gtw case, all we do is return 0. whch is effectively what this PR does. Do avoid a roundabout way to return 0 for the IPC gateway case.

btian1: I think previous logic should have something wrong for gtw case, it does go into multi_endpoint_copy and perform one time do_endpoint_copy(althrough it will return 0 immediately), this version will directly return 0 with gtw case.

So, conclusion is:
1. previous copier_copy code logic is redundant for gtw case.
2. both CI and QB are passed with this changes.

lyakh: > in do_endpoint)copy() for the ipc_gtw case, all we do is return 0. whch is effectively what this PR does.

@ranj063 oh, sorry, I was looking at my slightly outdated copy of copier.c, sorry

src/audio/copier/copier.c
```diff
@@ -469,28 +469,6 @@ static int copier_comp_trigger(struct comp_dev *dev, int cmd)
...
-	struct copier_data *cd = module_get_private_data(mod);
-
-	switch (dev->ipc_config.type) {
-	case SOF_COMP_HOST:
-		if (!cd->ipc_gtw)
-			return host_common_copy(cd->hd, dev, copier_host_dma_cb);
-		break;
-	case SOF_COMP_DAI:
-		if (cd->endpoint_num == 1)
-			return dai_common_copy(cd->dd[0], dev, cd->converter);
```
lyakh, line 425: this case is already handled somewhere, right?

lyakh (approved):
we can address the comments in a follow-up, and the problem with the `.init_data` is larger than just the copier

lgirdwood (approved):
@ranj063 good for you ?

ranj063:
> @ranj063 good for you ?

@lgirdwood yes already approved

PR #7741, opened 2023-06-05
[stable-v2.2] intel: ssp: DMA request management and TX stop simplification

ujfalusi:
Hi,

This PR contains two patches:

This first one (verified on product: #7732) changes how the DMA request is managed (or not managed currently):
instead of keeping both TX and RX DMA request active, only keep the one enabled which is in use.

The second one is an improved version of #7735 to make sure that the DMA is not running to fill the TX FIFO when we disable the SSP (in case of playback only stop).
#7735 has been verified on production, this change just makes it a bit safer.

Replaces #7732 and #7735

### DIFF
--- a/src/drivers/intel/ssp/ssp.c
+++ b/src/drivers/intel/ssp/ssp.c
@@ -835,18 +835,19 @@ static int ssp_set_config_blob(struct dai *dai, struct ipc_config_dai *common_co
 {
 	struct ipc4_ssp_configuration_blob *blob = spec_config;
 	struct ssp_pdata *ssp = dai_get_drvdata(dai);
-	uint32_t ssc0, sstsa, ssrsa;
+	uint32_t ssc0, sstsa, ssrsa, sscr1;
 
 	/* set config only once for playback or capture */
 	if (dai->sref > 1)
 		return 0;
 
 	ssc0 = blob->i2s_driver_config.i2s_config.ssc0;
-	sstsa = blob->i2s_driver_config.i2s_config.sstsa;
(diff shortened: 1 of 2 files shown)


ujfalusi:
Changes since v1:
- New patch inserted as first to handle the SSP clock consumer cases.

The set passes validation, but broader testing would be appreciated.

lgirdwood (commented):
@ujfalusi whats the status with this update in Zephyr ?

ujfalusi:
Changes since v2:
- rebased on stable-v2.2 to make sure that there are no conflicts and the PR can be tested with other mainline fixes

ujfalusi:
Changes since v3:
- rebased on stable-v2.2 after the ssp mclk always on patch

kv2019i:
This is a really good cleanup. Can you @Vamshigopal and @brentlu review this as well?

Vamshigopal:
@kv2019i Sure, I was checking on Btoffload usecase where we have diff mclk
https://github.com/thesofproject/sof/blob/a951579bb27e7f8556824ffe4e88705ccee0d024/tools/topology/topology1/platform/intel/intel-generic-bt.m4#L11

checking on switching usecases between spp1(speaker) and ssp2(btoffload).

ujfalusi:
I will respin this PR with some learnings from Zephyr debugging

kv2019i (approved):

Vamshigopal:
Tested basic audio usecases and BT offload related usecases , Test results are clean and good. No regressions seen.

Vamshigopal (approved):
Changes looks good to me.

PR #7751, opened 2023-06-06
[Not For Main] Audio: Volume: Add bypass processing functions for volume 

andrula-song:
Add bypass functions for volume. If gain of all channels equals 0dB, then we use bypass functions to transport the stream data.


src/audio/module_adapter/module/volume/volume_generic.c
```diff
@@ -93,6 +93,47 @@ static void vol_s24_to_s24(struct processing_module *mod, struct input_stream_bu
...
+{
+	struct audio_stream __sparse_cache *source = bsource->data;
+	struct audio_stream __sparse_cache *sink = bsink->data;
+	int32_t *x;
+	int32_t *y;
+	int nmax, n;
+	const int nch = audio_stream_get_channels(source);
+	int remaining_samples = frames * nch;
+
+	x = audio_stream_wrap(source, (char *)audio_stream_get_rptr(source) + bsource->consumed);
```
singalsu, line 121: Use audio_stream_copy()

andrula-song: not so suitable since when ramp, we won't update the read/ write pointer inside the while loop of volume_process.

andrula-song:
close this PR and use https://github.com/thesofproject/sof/pull/7753 to replace it.

PR #7756, opened 2023-06-06
Cloud-detected fuzz fixups

andyross:
A collection of minor fixes discovered by oss-fuzz over the last week.  There's some guesswork involved here, as most of the crashes are lower down as other code follows e.g. bad pointers.  I **think** the conditions here cover most (maybe all) of the failures I'm seeing reported.  We'll see how that evolves, no doubt there will be more to come.

### DIFF
--- a/src/audio/kpb.c
+++ b/src/audio/kpb.c
@@ -394,7 +394,12 @@ static int kpb_set_verify_ipc_params(struct comp_dev *dev,
 
 	ret = memcpy_s(&kpb->config, sizeof(kpb->config), ipc_config->data,
 		       ipc_config->size);
-	assert(!ret);
+
+	if (ret) {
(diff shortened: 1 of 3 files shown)


cujomalainey (approved):

cujomalainey:
@mwasko fixes for cavs25-drop-stable as well as other stable branches

kv2019i:
@andyross @mwasko This introduced an rimage regression. Fixed in https://github.com/thesofproject/sof/pull/7782

marc-hb:
The accidental and unrelated rimage change was caught by Github check https://github.com/thesofproject/sof/actions/runs/5190604343/jobs/9357300136
(it's also clearly visible in the "Files changed" tab)

Please don't merge PRs with unexplained red crosses.

PR #7761, opened 2023-06-07
copier: fix attenuation stream frames calculation

iganakov:
When calculating number of frames to be attenuated in HOST copier source stream frame bytes should be used instead of sink frame bytes.

E.g. in case of 24/24 bit to 24/32 bit PCM conversion using sink container size will reduce frames number by 1/4 which will lead to multiple glitches.

Also attenuation has to be applied in HOST copier in playback scenario only, thus remove from do_conversion_copy()

### DIFF
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -439,7 +439,6 @@ static int do_conversion_copy(struct comp_dev *dev,
 			      struct comp_copy_limits *processed_data)
 {
 	int i;
-	int ret;
 
 	/* buffer params might be not yet configured by component on another pipeline */
 	if (!src->hw_params_configured || !sink->hw_params_configured)
@@ -454,12 +453,6 @@ static int do_conversion_copy(struct comp_dev *dev,
 
 	cd->converter[i](&src->stream, 0, &sink->stream, 0,
 			 processed_data->frames * audio_stream_get_channels(&sink->stream));
-
-	if (cd->attenuation) {
-		ret = apply_attenuation(dev, cd, sink, processed_data->frames);
-		if (ret < 0)
-			return ret;
-	}
 
 	buffer_stream_writeback(sink, processed_data->sink_bytes);
 	comp_update_buffer_produce(sink, processed_data->sink_bytes);
--- a/src/audio/copier/copier_host.c
+++ b/src/audio/copier/copier_host.c
@@ -126,7 +126,7 @@ void copier_host_free(struct copier_data *cd)
 void copier_host_dma_cb(struct comp_dev *dev, size_t bytes)
 {
 	struct copier_data *cd = comp_get_drvdata(dev);
-	struct comp_buffer __sparse_cache *sink;
+	struct comp_buffer __sparse_cache *sink, *source;
 	int ret, frames;
 
 	comp_dbg(dev, "copier_host_dma_cb() %p", dev);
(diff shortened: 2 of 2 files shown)


src/audio/copier/copier_host.c
```diff
@@ -107,15 +107,14 @@ void copier_host_dma_cb(struct comp_dev *dev, size_t bytes)
...
-		frames = bytes / get_sample_bytes(audio_stream_get_frm_fmt(&sink->stream));
-		frames = frames / audio_stream_get_channels(&sink->stream);
+	/* Apply attenuation since copier copy missed this with host device
+	 * remove. Attenuation has to be applied in HOST Copier only with
+	 * playback scenario.
+	 */
+	if (cd->attenuation && dev->direction == SOF_IPC_STREAM_PLAYBACK) {
+		source = buffer_acquire(cd->hd->dma_buffer);
+		sink = buffer_acquire(cd->hd->local_buffer);
+		frames = bytes / audio_stream_frame_bytes(&source->stream);
```
btian1, line 117: seems there is no issue with previous implementation, at this moment, copy already done, all the data is already in sink buffer, attenuation only happens on sink buffer, so use sink->stream to calculate frame is more suitable?

src/audio/copier/copier.c
```diff
@@ -753,12 +752,6 @@ static int do_conversion_copy(struct comp_dev *dev,
 	cd->converter[i](&src->stream, 0, &sink->stream, 0,
 			 processed_data->frames * audio_stream_get_channels(&sink->stream));
 
-	if (cd->attenuation) {
```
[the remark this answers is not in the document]
iganakov: @btian1 Yes, only for playback and host copier

src/audio/copier/copier_host.c
iganakov: @btian1 We're getting number of bytes from DMA here. So if e.g. host format is 24/24bit/48000kHz/2ch there will be PCM conversion to 24/32bit applied on host copier. Using `sink->stream`  frame bytes  will result in 36 frames                 (288 bytes/4 * 2ch) to be attenuated instead of 48. Below is an example of distorted signal:
 
image

dnikodem (approved):

src/audio/copier/copier_host.c
```diff
@@ -107,15 +107,14 @@ void copier_host_dma_cb(struct comp_dev *dev, size_t bytes)
...
-			sink = buffer_acquire(cd->hd->dma_buffer);
-
-		frames = bytes / get_sample_bytes(audio_stream_get_frm_fmt(&sink->stream));
-		frames = frames / audio_stream_get_channels(&sink->stream);
+	/* Apply attenuation since copier copy missed this with host device
+	 * remove. Attenuation has to be applied in HOST Copier only with
+	 * playback scenario.
+	 */
+	if (cd->attenuation && dev->direction == SOF_IPC_STREAM_PLAYBACK) {
+		source = buffer_acquire(cd->hd->dma_buffer);
```
dnikodem, line 115: I missed it last time - what about releasing the source buffer?

dnikodem (changes requested):

src/audio/copier/copier_host.c
btian1: ok, I see, is 24/24bit possible? or do you see this case? in CI or QB? I wonder it is still 24/32bit format.

iganakov: Yes, it is possible. It is one of the main formats used by Windows driver.

kv2019i (approved):

kv2019i:
#7191 hit on MTL nocodec in https://sof-ci.01.org/sofpr/PR7761/build9362/devicetest/index.html, not related to this PR . Rest of checks good.

PR #7762, opened 2023-06-07
[NOT FOR MAIN] module_adapter: skip invalidate/writeback with single_core

ranj063:
Add a new flag, is_multi_core, and set it if a module to connected to another module on a different core.


ranj063:
SOFCI TEST

PR #7763, opened 2023-06-08
Kconfig: don't fall back on CONFIG_TIGERLAKE

marc-hb:
There's really no reason to fall back on a default "platform" and pick up a bunch of other, totally random settings.

Now the following command fails immediately with a useful `platform not defined` error message instead of stopping with a cryptic `platform/trace/trace.h: No such file or directory` after successfully compiling half the .c files in a Frankenstein CONFIG_uration.

```
  west build -p --board mimx8mm_evk_a53 modules/audio/sof/app
```

More context in #7192.

### DIFF
--- a/src/arch/xtensa/CMakeLists.txt
+++ b/src/arch/xtensa/CMakeLists.txt
@@ -22,6 +22,8 @@ elseif(CONFIG_MT8188)
         set(platform_folder mt8188)
 elseif(CONFIG_MT8195)
 	set(platform_folder mt8195)
+else()
+	message(FATAL_ERROR "Platform not defined, check your Kconfiguration?")
 endif()
 
 set(fw_name ${CONFIG_RIMAGE_SIGNING_SCHEMA})
--- a/src/arch/xtensa/CMakeLists.txt
+++ b/src/arch/xtensa/CMakeLists.txt
@@ -12,6 +12,10 @@ elseif(CONFIG_IMX8M)
 	set(platform_folder imx8m)
 elseif(CONFIG_IMX8ULP)
 	set(platform_folder imx8ulp)
+elseif(CONFIG_RENOIR)
+	set(platform_folder amd/renoir)
+elseif(CONFIG_REMBRANDT)
+	set(platform_folder amd/rembrandt)
 elseif(CONFIG_MT8186)
         set(platform_folder mt8186)
 elseif(CONFIG_MT8188)
@@ -22,18 +26,11 @@ endif()
 
 set(fw_name ${CONFIG_RIMAGE_SIGNING_SCHEMA})
 
-set(platform_ld_script ${platform_folder}.x)
+# File name without directory
+get_filename_component(_plf_ld_script ${platform_folder} NAME)
+set(platform_ld_script ${_plf_ld_script}.x)
 set(platform_rom_ld_script rom.x)
 
-if(CONFIG_RENOIR)
-	set(platform_folder amd/renoir)
-	set(platform_ld_script renoir.x)
-endif()
-
-if(CONFIG_REMBRANDT)
-	set(platform_folder amd/rembrandt)
-	set(platform_ld_script rembrandt.x)
-endif()
 # includes
 
 # None of these should be included if Zephyr strict headers are used.
--- a/scripts/fuzz.sh
+++ b/scripts/fuzz.sh
@@ -67,13 +67,17 @@ main()
   : "${TEST_DURATION:=3}"
   : "${FUZZER_STDOUT:=/dev/stdout}" # bashism
 
-  west build -d build-fuzz -b native_posix "$SOF_TOP"/app/ -- \
-    -DCONFIG_ASSERT=y \
-    -DCONFIG_SYS_HEAP_BIG_ONLY=y \
-    -DCONFIG_ZEPHYR_NATIVE_DRIVERS=y \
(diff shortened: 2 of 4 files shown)


dbaluta (dismissed):

marc-hb:
The fuzzer used to be configured with CONFIG_TIGERLAKE, that can probably be just added back
https://github.com/thesofproject/sof/actions/runs/5206226654/jobs/9392562028?pr=7763

More surprisingly, AMD Renoir was also defaulting to CONFIG_TIGERLAKE?!
https://github.com/thesofproject/sof/actions/runs/5206226652/jobs/9392562617?pr=7763

andyross (dismissed):
Makes sense.  FWIW in the long term it's probably best to to some work to unify the SOF "platform" choice with the Zephyr-side "board"/"soc"/"arch" tuple.

As regards the fuzzer: there were definitely some leaky kconfigs in there.  But for that application specifically, it seemed worthwhile to make sure as much of the firmware source base gets included as it would for the real device, even if the configuration is a little non-orthogonal.

src/arch/xtensa/CMakeLists.txt
```diff
@@ -18,6 +18,8 @@ elseif(CONFIG_MT8188)
         set(platform_folder mt8188)
 elseif(CONFIG_MT8195)
 	set(platform_folder mt8195)
+else()
+	message(FATAL_ERROR "Platform not defined, check your Kconfiguration?")
```
andyross, line 22: FWIW: you could move this to a build-time #error in a C file somewhere to avoid having to do this in two different cmake schemes.

marc-hb: I respectfully but strongly disagree. I think it's better to check and fail as early as possible. Also, C compilation is parallelized and typically unreadable before adding `-j1` whereas CMake configuration is single-threaded. I think it's also better to always have a "default" error clause in a case/switch, and/or to check a variable is not empty before using it (the `-I platform/${EMPTY}/include` wasted a large amount of my time!)

I avoid CMake as much as possible but the duplication here is only one line. It's only because Zephyr versus non-Zephyr so it won't grow larger.

Finally, in which C file?

All this being said, if there is some C file where this check could be useful TOO then why not; "defence in depth".

marc-hb:
All LNL tests failed in https://sof-ci.01.org/sof-pr-viewer/#/build/PR7763/build12144879, just like they did in totally different #7772, so this is clearly unrelated.

andyross (approved):

marc-hb:
Thanks to whoever (@wszypelt ?) restored the LNL device. Now the newer https://sof-ci.01.org/sof-pr-viewer/#/build/PR7763/build12147803 is mostly passing except for this `teardown` failure below that occurred in only two tests. @wszypelt this feels like a bug in the test framework or infrastructure; you may want to let the validation team know about it.

In any case it's totally unrelated to this PR.

```
# SAME test is both PASSED and ERROR, that's funny
06:58:34,859 INFO  - [gw0] [100%] PASSED tests/ace/fw_116_sdw_low_level/test_04_sdw_transfer_loopback_aggregation.py::TestSdwTransferLoopbackAggregation::test_116_04_sdw_loopback_aggregation_capture[0-source_PDI_4_to_sink_PDI_5_and_3-16000Hz_24in32bit_2ch]
06:58:35,126 INFO  - [gw0] [100%] ERROR  tests/ace/fw_116_sdw_low_level/test_04_sdw_transfer_loopback_aggregation.py::TestSdwTransferLoopbackAggregation::test_116_04_sdw_loopback_aggregation_capture[0-source_PDI_4_to_sink_PDI_5_and_3-16000Hz_24in32bit_2ch]

06:58:35,127 INFO  - _ ERROR at teardown of TestSdwTransferLoopbackAggregation.test_116_04_sdw_loopback_aggregation_capture[0-source_PDI_4_to_sink_PDI_5_and_3-16000Hz_24in32bit_2ch] _

06:58:35,127 INFO  - cavs_scripts-py\utilities\test_logger.py:404: in iteration_end
06:58:35,127 INFO  - l_handler.logger.close_current_iteration()
06:58:35,127 INFO  - cavs_scripts-py\utilities\loggers\weblogger_lite\weblogger_lite.py:253: in close_current_iteration
06:58:35,127 INFO  - self.iterations_files.append(self.current_iteration.close(passed=self.current_iteration_passed(),
06:58:35,127 INFO  - cavs_scripts-py\utilities\loggers\weblogger_lite\weblogger_lite.py:114: in close
06:58:35,127 INFO  - with open(iter_path_abs, "w") as file:
06:58:35,127 INFO  - E FileNotFoundError: [Errno 2] No such file or directory: 'C:\\history\\cAVS\\LABELED\\SMOKE_IPC4_pull-7763-merge_1405642_7421_7431\\116_04_TestSdwTransferLoopbackAggregation\\iters\\1_TestSdwTransferLoopbackAggregation_test_116_04_sdw_loopback_aggregation_capture[0-source_PDI_4_to_sink_PDI_5_and_3-16000Hz_24in32bit_2ch].html'
```

wszypelt:
@marc-hb Thanks for the info, I'll debug it today

juimonen (approved):

marc-hb:
> Thanks for the info, I'll debug it today

Glad I could report a actionable bug! Thanks for fixing the issue, https://sof-ci.01.org/sof-pr-viewer/#/build/PR7763/build12159888 is all green now.

PR #7764, opened 2023-06-08
Topology2: Remove usage of audio_format class

ranj063:
First PR to remove it from HDA and HDMI definitions. This will be followed up with several others to do the same thing until the audio_format class is deprecated


juimonen (approved):
not sure of the end goal here, but changes look ok and consistent

ranj063:
> not sure of the end goal here, but changes look ok and consistent

@juimonen the end goal is to stop using the audio_format class. We cannot defie input/output audio formats in pairs always.

kv2019i (approved):

PR #7768, opened 2023-06-08
[cavs2.5-001-drop-stable] intel: ssp: Revise receive FIFO draining

Vamshigopal:
The receive FIFO needs to be drained in a different way depending when it is done.
- before start If the RX FIFO is in overflow state then we must read all the entries out to empty it (it was after all full).

- before stop The DMA might be already running to read out data. Check the FIFO level change in one sample time which gives us the needed information to decide to wait for another loop for the DMA burst to finish, wait for the DMA to start it's burst (DMA request was asserted) or drain the FIFO directly.

No need to drain the RX fifo at probe time.

(cherry picked from commit 69d56a757740274a6f2ccf17fc77c4188f17fc01)

### DIFF
--- a/src/drivers/intel/ssp/ssp.c
+++ b/src/drivers/intel/ssp/ssp.c
@@ -67,49 +67,74 @@ static void ssp_empty_tx_fifo(struct dai *dai)
 		ssp_write(dai, SSSR, sssr);
 }
 
-/* empty SSP receive FIFO */
-static void ssp_empty_rx_fifo(struct dai *dai)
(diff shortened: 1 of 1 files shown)


Vamshigopal:
@mwasko @kv2019i @ujfalusi We need this fix in cavs2.5-001-drop-stable branch.

btian1 (approved):

ujfalusi (approved):

kv2019i:
@mwasko ping?

PR #7773, opened 2023-06-08
src: mixin_mixout: Don't invalidate or writeback source/bink buffers

ranj063:
Handle the source/sink buffer invalidate/writeback in the module adapter instead. Remove the 2 flags in struct processing_module as these are not needed anymore.

### DIFF
--- a/src/audio/mixin_mixout/mixin_mixout.c
+++ b/src/audio/mixin_mixout/mixin_mixout.c
@@ -122,7 +122,6 @@ static int mixin_init(struct processing_module *mod)
 
 	dev->ipc_config.frame_fmt = frame_fmt;
 	mod->simple_copy = true;
-	mod->skip_src_buffer_invalidate = true;
 
 	return 0;
 }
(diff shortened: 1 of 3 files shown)


src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -378,18 +376,11 @@ static int mixin_process(struct processing_module *mod,
...
 		frames_to_copy = MIN(source_avail_frames, sinks_free_frames);
 		bytes_to_consume_from_source_buf =
 			audio_stream_period_bytes(input_buffers[0].data, frames_to_copy);
-		if (bytes_to_consume_from_source_buf > 0) {
+		if (bytes_to_consume_from_source_buf > 0)
 			input_buffers[0].consumed = bytes_to_consume_from_source_buf;
-			source_c = attr_container_of(input_buffers[0].data,
-						     struct comp_buffer __sparse_cache,
-						     stream, __sparse_cache);
-			buffer_stream_invalidate(source_c, bytes_to_consume_from_source_buf);
```
btian1, line 391: seems this is the only difference, is it caused CI error?

PR #7775, opened 2023-06-09
topology2: add development tplg to test playback with 2 cores

mengdonglin:
This is to allow CI to test a simple multicore playback case: SSP0 playback on core 0 and SSP2 playback on core 1.

Capture uses core 0 only atm. More multicore usage will be introduced later.

Tested on MTL RVP.


tools/topology/topology2/development/tplg-targets.cmake
```diff
@@ -37,6 +37,9 @@ DEEPBUFFER_D0I3_COMPATIBLE=true"
 "cavs-nocodec-multicore\;sof-mtl-nocodec-multicore\;PLATFORM=mtl,SSP1_ENABLED=true,SSP0_CORE_ID=0,\
 SSP1_CORE_ID=1,SSP2_CORE_ID=2,PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-mtl-nocodec.bin"
 
+"cavs-nocodec-multicore\;sof-mtl-nocodec-multicore-ssp0-ssp2\;PLATFORM=mtl,SSP1_ENABLED=false,\
+SSP0_CORE_ID=0,SSP2_CORE_ID=1,PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-mtl-nocodec.bin"
```
ranj063, line 41: nit-pick: SSP0_CORE_ID is 0 by default. so no need to set it again here

ranj063 (approved):

RanderWang (approved):

tools/topology/topology2/development/tplg-targets.cmake
mengdonglin: > nit-pick: SSP0_CORE_ID is 0 by default. so no need to set it again here
@ranj063 Thank you! I've removed this.

mengdonglin:
@kv2019i Can you help review this PR and merge it if it looks good to you?  The CI failures are not caused by this topology PR as we haven't enabled multi-core test yet. This PR is to allow regular CI testing from a  simple multi-core case.

mengdonglin:
@ranj063 @RanderWang @plbossart @kv2019i  I also add the same 2-core nocodec topology for TGL. So we can compare test results on MTL and TGL.

kv2019i:
Manifest-check failing due to regression, fix in https://github.com/thesofproject/sof/pull/7782

Otherwise looks normal, proceeding with merge.

PR #7786, opened 2023-06-12
audio: buffer: do not mix cached and non-cached access in buffer_init

kv2019i:
No reason to mix cached and noncached operations when initializing the buffer in buffer_ini(). We take a cached reference, so let's use that to initialize the object.

### DIFF
--- a/src/audio/buffer.c
+++ b/src/audio/buffer.c
@@ -31,6 +31,7 @@ struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t flags, u
 {
 	struct comp_buffer *buffer;
 	struct comp_buffer __sparse_cache *buffer_c;
+	void *stream_addr;
 
 	tr_dbg(&buffer_tr, "buffer_alloc()");
 
@@ -51,23 +52,22 @@ struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t flags, u
 		return NULL;
 	}
 
-	buffer->stream.addr = rballoc_align(0, caps, size, align);
-	if (!buffer->stream.addr) {
+	stream_addr = rballoc_align(0, caps, size, align);
+	if (!stream_addr) {
 		rfree(buffer);
 		tr_err(&buffer_tr, "buffer_alloc(): could not alloc size = %u bytes of type = %u",
 		       size, caps);
 		return NULL;
 	}
 
-	buffer->stream.underrun_permitted = !!(flags & SOF_BUF_UNDERRUN_PERMITTED);
-	buffer->stream.overrun_permitted = !!(flags & SOF_BUF_OVERRUN_PERMITTED);
-
-	list_init(&buffer->source_list);
-	list_init(&buffer->sink_list);
-
 	/* From here no more uncached access to the buffer object, except its list headers */
 	buffer_c = buffer_acquire(buffer);
+	buffer_c->stream.addr = stream_addr;
 	buffer_init(buffer_c, size, caps);
+
+	buffer_c->stream.underrun_permitted = !!(flags & SOF_BUF_UNDERRUN_PERMITTED);
+	buffer_c->stream.overrun_permitted = !!(flags & SOF_BUF_OVERRUN_PERMITTED);
+
 	buffer_release(buffer_c);
 
 	/*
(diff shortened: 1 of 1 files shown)


juimonen (approved):

src/audio/buffer.c
```diff
@@ -31,6 +31,7 @@ struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t flags, u
 {
 	struct comp_buffer *buffer;
 	struct comp_buffer __sparse_cache *buffer_c;
+	void *stream_addr;
```
lyakh, line 34: commit message: `buffer_alloc()` not `buffer_init()`

src/audio/buffer.c
```diff
@@ -51,23 +52,25 @@ struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t flags, u
...
+	/* From here no more uncached access to the buffer object, except its list headers */
+	buffer_c = buffer_acquire(buffer);
+	buffer_c->stream.addr = stream_addr;
+	buffer_init(buffer_c, size, caps);
+
 	buffer->stream.underrun_permitted = !!(flags & SOF_BUF_UNDERRUN_PERMITTED);
 	buffer->stream.overrun_permitted = !!(flags & SOF_BUF_OVERRUN_PERMITTED);
 
 	list_init(&buffer->source_list);
 	list_init(&buffer->sink_list);
```
lyakh, line 72: these buffer allocation and initialisation code is extremely sensitive. You say, that you don't want to mix cached and uncached access, but here you keep uncached access in 4 assignments above. Further, this change should actually corrupt buffer memory: in your new line 64 you obtain a cached address, then you modify the object via that address, i.e. it gets read into cache and the cache copy gets modified. Then the above 4 lines perform modifications on uncached memory, now cache is outdated. Then in line 86 below cache is written back, overwriting your modifications to uncached memory.

lyakh (changes requested):

src/audio/buffer.c
kv2019i: fixed in V2

src/audio/buffer.c
```diff
@@ -51,23 +52,25 @@ struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t flags, u
...
 	/* From here no more uncached access to the buffer object, except its list headers */
 	buffer_c = buffer_acquire(buffer);
+	buffer_c->stream.addr = stream_addr;
 	buffer_init(buffer_c, size, caps);
+
+	buffer_c->stream.underrun_permitted = !!(flags & SOF_BUF_UNDERRUN_PERMITTED);
+	buffer_c->stream.overrun_permitted = !!(flags & SOF_BUF_OVERRUN_PERMITTED);
+
+	list_init(&buffer_c->source_list);
+	list_init(&buffer_c->sink_list);
```
lyakh, line 72: now these two lines kind of break the concept. On the one hand they're useless, because those list objects are *members* in lists, not list *heads*, OTOH we always do that and in some places we actually use the fact that list members are "empty heads" when they aren't on any list. And here it breaks the concept because it assigns
```
head->next = head;
```
where `head` now is a cached address and those lists are always linked with uncached addresses

btian1 (approved):

iuliana-prodan (approved):

src/audio/buffer.c
kv2019i: @lyakh pushed these out to uncached ops at the end in V3 version

marc-hb:
This fix seems super-generic. Should it be backported? I mean after a while without noticing any regression.

marc-hb:
In unrelated https://sof-ci.01.org/softestpr/PR1058/build447/devicetest/index.html (internal run ID 27676), MTLP_RVP_NOCODEC is unfortunately still failing as usual as before this PR.


Still no useful info in mtrace, still looking like a flushing problem.

marc-hb:
My bad, https://sof-ci.01.org/softestpr/PR1058/build447/devicetest/index.html is very recent YET it ran with some old SOF commit  8fb011df6c95 that predates this PR?!?

This PR was merged on June 14 16:19:49  UTC and that test run was started on June 14 23:06:05 UTC, yet the test run used June 13 SOF commit 8fb011df6c95. I have no idea why; will discuss and investigate in https://github.com/thesofproject/sof-test/pull/1058.


Other recent test runs look better! Crossing fingers.

PR #7791, opened 2023-06-13
[Not for main]Audio: Aria: Optimization of Aria

andrula-song:
Optimize the Aria, test on tgl device, the bypass mode can reduce about 51.1% cycles and the normal mode can reduce about 24.0% cycles.

### DIFF
--- a/src/audio/aria/aria_generic.c
+++ b/src/audio/aria/aria_generic.c
@@ -7,50 +7,97 @@
 
 #ifdef ARIA_GENERIC
 
+/**
+ * \brief Aria gain index mapping table
+ */
+const uint8_t INDEX_TAB[] = {
+		0,    1,    2,    3,
+		4,    5,    6,    7,
+		8,    9,    0,    1,
+		2,    3,    4,    5,
+		6,    7,    8,    9,
+		0,    1,    2,    3
+};
+
 inline void aria_algo_calc_gain(struct comp_dev *dev, size_t gain_idx,
-				int32_t *__restrict data, const size_t src_size)
+				struct audio_stream __sparse_cache *source, int frames)
 {
 	struct aria_data *cd = comp_get_drvdata(dev);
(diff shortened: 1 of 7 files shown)


src/audio/aria/aria_generic.c
```diff
@@ -7,50 +7,93 @@
...
+	int32_t *src = audio_stream_get_rptr(source);
+	int i, n;
 
-	for (size_t i = 0; i < src_size; ++i)
-		max_data = MAX(max_data, ABS(data[i]));
+	while (samples) {
+		n = audio_stream_samples_without_wrap_s32(source, src);
+		n = MIN(samples, n);
+		for (i = 0; i < n; i++)
+			max_data = MAX(max_data, ABS(src[i]));
```
singalsu, line 37: I'm not sure but is there risk for ABS() to be evaluated twice due to implementation of MAX() macro?

src/audio/aria/aria_generic.c
```diff
@@ -7,50 +7,93 @@
...
+			}
+			gain += step;
+			out += ch_n;
 		}
+		samples -= n;
+		in = cir_buf_wrap(in, cd->data_addr, cd->data_end);
+		out = audio_stream_wrap(sink, out);
 	}
-	cd->gain_state = (cd->gain_state + 1) % ARIA_MAX_GAIN_STATES;
+	cd->gain_state = INDEX_TAB[cd->gain_state + 1];
```
singalsu, line 92: Nice!

singalsu (approved):

src/audio/aria/aria_generic.c
andrula-song: from the xtensa test result, it is evaluated twice, thanks!

andrula-song:
the failed CI cases are unrelated issues and failed QB cases is fw load failed, unrelated too. Merge it.

Issue #7482, opened 2023-04-20
[BUG] ipc timed out MOD_SET_DX in suspend-resume in TGL/MTL


fredoh9:
**Describe the bug**

Found in Intel internal daily test, test case was check-suspend-resume-with-playback 
planresultdetail/23903?model=TGLU_UP_HDA_IPC4ZPH&testcase=check-suspend-resume-with-playback-5

```
[ 6168.932762] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx      : 0x47000000|0x0: MOD_SET_DX [data size: 8]
[ 6168.971415] kernel: e1000e: EEE TX LPI TIMER: 00000011
[ 6169.437427] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: ipc timed out for 0x47000000|0x0
[ 6169.437432] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: Attempting to prevent DSP from entering D3 state to preserve context
[ 6169.437435] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: ------------[ IPC dump start ]------------
[ 6169.437453] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: hda irq intsts 0x00000000 intlctl 0xc0000000 rirb 00
[ 6169.437457] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: dsp irq ppsts 0x00000000 adspis 0x00000000
[ 6169.437479] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: Host IPC initiator: 0xc7000000|0x0|0x0, target: 0x0|0x0|0x0, ctl: 0x3
[ 6169.437483] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: ------------[ IPC dump end ]------------
[ 6169.437486] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: ------------[ DSP dump start ]------------
[ 6169.437488] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: IPC timeout
[ 6169.437490] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: fw_state: SOF_FW_BOOT_COMPLETE (7)
[ 6169.437498] kernel: sof-audio-pci-intel-tgl 0000:00:1f.3: 0x00000005: module: ROM, state: FW_ENTERED, running
[...]

PR #7792, opened 2023-06-13
Audio: Module adapter: Add a new simple copy function

singalsu:
This patch adds a simplified version of simple copy for module adapter clients with one source and one sink. The overhead of discovering multiple source and sink buffers is avoided.

The module_adapter_bind() and module_adapter_unbind() functions are changed to detect multiple sources or sinks condition to control the used simple copy function.

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -314,6 +314,19 @@ int module_adapter_prepare(struct comp_dev *dev)
 		return -EINVAL;
 	}
 
+#if CONFIG_IPC_MAJOR_3
+	/* Check if audio stream client has only one source and one sink buffer to use a
+	 * simplified copy function.
+	 */
+	if (IS_PROCESSING_MODE_AUDIO_STREAM(mod) && mod->num_input_buffers == 1 &&
+	    mod->num_output_buffers == 1) {
+		mod->source_comp_buffer = list_first_item(&dev->bsource_list,
+							  struct comp_buffer, sink_list);
+		mod->sink_comp_buffer = sink;
+		mod->stream_copy_single_to_single = true;
+	}
+#endif
+
 	/* allocate memory for input buffers */
 	if (mod->num_input_buffers) {
 		mod->input_buffers =
@@ -785,6 +798,55 @@ module_single_source_setup(struct comp_dev *dev,
 	return num_output_buffers;
 }
 
+static int module_adapter_audio_stream_copy_1to1(struct comp_dev *dev)
+{
+	struct processing_module *mod = comp_get_drvdata(dev);
+	struct comp_buffer __sparse_cache *source_c;
+	struct comp_buffer __sparse_cache *sink_c;
+	uint32_t num_output_buffers = 0;
+	uint32_t frames;
+	int ret;
+
+	source_c = buffer_acquire(mod->source_comp_buffer);
+	sink_c = buffer_acquire(mod->sink_comp_buffer);
+	frames = audio_stream_avail_frames_aligned(&source_c->stream, &sink_c->stream);
+	mod->input_buffers[0].size = frames;
+	mod->input_buffers[0].consumed = 0;
+	mod->input_buffers[0].data = &source_c->stream;
+	mod->output_buffers[0].size = 0;
+	mod->output_buffers[0].data = &sink_c->stream;
+	if (!mod->skip_src_buffer_invalidate) /* TODO: add mod->is_multi_core && optimization */
+		buffer_stream_invalidate(source_c,
+					 frames * audio_stream_frame_bytes(&source_c->stream));
+
+	/* Note: Source buffer state is not checked to enable mixout to generate zero
+	 * PCM codes when source is not active.
+	 */
+	if (sink_c->sink->state == dev->state)
+		num_output_buffers = 1;
+
+	ret = module_process_legacy(mod, mod->input_buffers, 1,
+				    mod->output_buffers, num_output_buffers);
+
+	/* consume from the input buffer */
+	mod->total_data_consumed += mod->input_buffers[0].consumed;
+	if (mod->input_buffers[0].consumed)
+		audio_stream_consume(&source_c->stream, mod->input_buffers[0].consumed);
+
+	/* produce data into the output buffer */
+	mod->total_data_produced += mod->output_buffers[0].size;
+	if (!mod->skip_sink_buffer_writeback) /* TODO: add mod->is_multi_core && optimization */
+		buffer_stream_writeback(sink_c, mod->output_buffers[0].size);
+
+	if (mod->output_buffers[0].size)
+		audio_stream_produce(&sink_c->stream, mod->output_buffers[0].size);
+
+	/* release all buffers */
+	buffer_release(sink_c);
+	buffer_release(source_c);
+	return ret;
+}
+
 static int module_adapter_audio_stream_type_copy(struct comp_dev *dev)
 {
 	struct comp_buffer __sparse_cache *source_c[PLATFORM_MAX_STREAMS];
@@ -794,6 +856,9 @@ static int module_adapter_audio_stream_type_copy(struct comp_dev *dev)
 	uint32_t num_input_buffers = 0;
 	uint32_t num_output_buffers = 0;
 	int ret, i = 0;
+
+	if (mod->stream_copy_single_to_single)
+		return module_adapter_audio_stream_copy_1to1(dev);
 
 	/* acquire all sink and source buffers */
 	list_for_item(blist, &dev->bsink_list) {
@@ -1429,6 +1494,32 @@ int module_adapter_get_attribute(struct comp_dev *dev, uint32_t type, void *valu
 	return 0;
 }
 
+static bool module_adapter_multi_sink_source_check(struct comp_dev *dev)
+{
+	struct processing_module *mod = comp_get_drvdata(dev);
+	struct list_item *blist;
+	int num_sources = 0;
+	int num_sinks = 0;
+
+	list_for_item(blist, &dev->bsource_list)
+		num_sources++;
+
+	list_for_item(blist, &dev->bsink_list)
+		num_sinks++;
+
+	comp_dbg(dev, "num_sources=%d num_sinks=%d", num_sources, num_sinks);
+
+	if (num_sources != 1 || num_sinks != 1)
+		return true;
+
+	/* re-assign the source/sink modules */
+	mod->sink_comp_buffer = list_first_item(&dev->bsink_list, struct comp_buffer, source_list);
+	mod->source_comp_buffer = list_first_item(&dev->bsource_list,
+						  struct comp_buffer, sink_list);
+
+	return false;
+}
+
 int module_adapter_bind(struct comp_dev *dev, void *data)
 {
 	struct module_source_info __sparse_cache *mod_source_info;
@@ -1440,6 +1531,8 @@ int module_adapter_bind(struct comp_dev *dev, void *data)
 
 	bu = (struct ipc4_module_bind_unbind *)data;
 	src_id = IPC4_COMP_ID(bu->primary.r.module_id, bu->primary.r.instance_id);
+
+	mod->stream_copy_single_to_single = !module_adapter_multi_sink_source_check(dev);
 
 	/* nothing to do if this module is the source during bind */
 	if (dev->ipc_config.id == src_id)
@@ -1489,6 +1582,8 @@ int module_adapter_unbind(struct comp_dev *dev, void *data)
 
 	bu = (struct ipc4_module_bind_unbind *)data;
 	src_id = IPC4_COMP_ID(bu->primary.r.module_id, bu->primary.r.instance_id);
+
+	mod->stream_copy_single_to_single = !module_adapter_multi_sink_source_check(dev);
 
 	/* nothing to do if this module is the source during unbind */
 	if (dev->ipc_config.id == src_id)
--- a/src/include/sof/audio/module_adapter/module/generic.h
+++ b/src/include/sof/audio/module_adapter/module/generic.h
@@ -184,6 +184,8 @@ struct processing_module {
 	struct output_stream_buffer *output_buffers;
 	uint32_t num_input_buffers; /**< number of input buffers */
 	uint32_t num_output_buffers; /**< number of output buffers */
+	struct comp_buffer *source_comp_buffer; /**< single source component buffer */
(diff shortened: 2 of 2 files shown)


src/audio/module_adapter/module_adapter.c
```diff
@@ -825,6 +840,57 @@ static int module_adapter_simple_copy(struct comp_dev *dev)
...
+			goto out;
+		}
+
+		ret = 0;
+	}
+
+	/* consume from input buffer */
+	mod->total_data_consumed += mod->input_buffers[0].consumed;
+	if (mod->input_buffers[0].consumed)
+		audio_stream_consume(&source_c->stream, mod->input_buffers[0].consumed);
```
lyakh, line 877: the original version uses `comp_update_buffer_{consume,produce}()` which also calls the notifier. Is this change on purpose?

src/audio/module_adapter/module_adapter.c
```diff
@@ -1319,6 +1388,12 @@ int module_adapter_bind(struct comp_dev *dev, void *data)
 	/* set the source dev pointer */
 	mod_source_info->sources[source_index] = source_dev;
 
+	/* Disable 1 source 1 sink simple copy mode if multiple source buffers */
+	if (count_ready_or_higher_sources(mod_source_info) > 1) {
+		comp_info(dev, "module_adapter_bind(), multi_sink_or_source = true");
+		mod->multi_source_or_sink = true;
```
lyakh, line 1394: this doesn't seem right. You shouldn't need to count those sources in certain states. You can do this perfectly deterministically on paths where that state is set or cleared. If `.bind()` adds a source in one of those states - then you know that the number has changed. If it adds one in a lower state - it doesn't change here, so no need to re-count. Same with `.unbind()` - do we allow unbinding ready sources?

src/audio/module_adapter/module_adapter.c
```diff
@@ -825,6 +840,57 @@ static int module_adapter_simple_copy(struct comp_dev *dev)
 	return ret;
 }
 
+static inline int module_adapter_simple_copy_one_source_one_sink(struct processing_module *mod)
```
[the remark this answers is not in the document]
singalsu: Not using same (dev -> mod) saves one operation. Could maybe change module_adapter_simple_copy() instead? Dev is simply mod->dev when needed.

src/audio/module_adapter/module_adapter.c
```diff
@@ -710,7 +725,7 @@ module_single_source_setup(struct comp_dev *dev,
 	return num_output_buffers;
 }
 
-static int module_adapter_simple_copy(struct comp_dev *dev)
+static inline int module_adapter_simple_copy(struct comp_dev *dev)
```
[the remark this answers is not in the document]
singalsu: Yes, I'm pretty sure I got same performance without, so the compiler in-lined. I thought making it explicit is better, but agree if that's the practise.

src/audio/module_adapter/module_adapter.c
singalsu: It's using part of Liam's patch. Copied accidentally, but this is better, right?

src/audio/module_adapter/module_adapter.c
singalsu: I likely have a bug in unbind since there's fails in repeated suspend-resume test. Need to rethink this, thanks for commenting!

src/audio/module_adapter/module_adapter.c
```diff
@@ -1319,6 +1405,12 @@ int module_adapter_bind(struct comp_dev *dev, void *data)
 	/* set the source dev pointer */
 	mod_source_info->sources[source_index] = source_dev;
 
+	/* Disable 1 source 1 sink simple copy mode if multiple source buffers */
+	if (module_adapter_multi_sink_source_check(dev)) {
+		comp_dbg(dev, "module_adapter_bind(), multi_sink_or_source = true");
+		mod->multi_source_or_sink = true;
```
ranj063, line 1411: you can simplify this to mod->multi_source_or_sink = module_adapter_multi_sink_source_check(dev);

src/audio/module_adapter/module_adapter.c
```diff
@@ -1353,6 +1445,12 @@ int module_adapter_unbind(struct comp_dev *dev, void *data)
 	if (source_index >= 0)
 		mod_source_info->sources[source_index] = NULL;
 
+	/* Re-enable 1 source 1 sink simple copy mode if single source buffer */
+	if (mod->simple_copy && !module_adapter_multi_sink_source_check(dev)) {
```
ranj063, line 1449: do we need this simple_copy check here? if all we care about is whether a module has multi sources/sinks, it shouldnt matter right?

src/audio/module_adapter/module_adapter.c
```diff
@@ -1275,6 +1342,25 @@ int module_adapter_get_attribute(struct comp_dev *dev, uint32_t type, void *valu
...
+	list_for_item(blist, &dev->bsource_list)
+		num_sources++;
+
+	list_for_item(blist, &dev->bsink_list)
+		num_sinks++;
+
+	comp_dbg(dev, "num_sources=%d num_sinks=%d", num_sources, num_sinks);
+	if (num_sources > 1 || num_sinks > 1)
+		return true;
+	else
```
ranj063, line 1360: redundant else

src/audio/module_adapter/module_adapter.c
ranj063: @singalsu please take this patch https://github.com/thesofproject/sof/pull/7805/commits/1e8b25299f2c9760c11abe4ffe6256c9ea36d99b and I think your IPC timeouts with unbind will go away

src/audio/module_adapter/module_adapter.c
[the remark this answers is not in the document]
lyakh: I'm fine either way as long as it's consistent :-) Also worth checking with other functions in the same file. As for saving an operation - not sure where? You need both in both anyway. So it's either
```
```
or
```
```
looks like both are just a single pointer dereference, but you don't know what the compiler makes of them, so, not easy to say with certainty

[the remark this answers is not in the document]
singalsu: Seem to be the same, it's a macro despite that it looks like a function. Unfortunately trace still needs dev handle, all other stuff needs only the more narrow mod scope.

src/audio/module_adapter/module_adapter.c
singalsu: Thanks, that solves it! I totally missed that scenario.

src/audio/module_adapter/module_adapter.c
singalsu: I removed this but, I wonder if it now causes an issue. If a module adapter client has not set simple copy the processing should never be handled by simple copy function. To speed up code, the module_adapter_copy() when calling module_adapter_simple_copy_1to1() is not checking mod->simple_copy.

I think I need to add to module_adapter_multi_sink_source_check() test for mod->simple_copy, and return true if simple_copy is false.

ranj063:
SOFCI TEST

src/audio/module_adapter/module_adapter.c
```diff
@@ -245,6 +245,20 @@ int module_adapter_prepare(struct comp_dev *dev)
...
+	if (mod->simple_copy && mod->num_input_buffers == 1 && mod->num_output_buffers == 1) {
+		mod->source = list_first_item(&dev->bsource_list, struct comp_buffer, sink_list);
+		mod->sink = sink;
+		mod->multi_source_or_sink = false;
+	} else {
+		mod->source = NULL;
+		mod->sink = NULL;
+		mod->multi_source_or_sink = true;
+	}
+
```
btian1, line 261: which part called first? prepare or bind? seems above code is duplicated with bind code.

singalsu: I don't think there's guaranteed order for those events. Do you see risk or ways to mitigate it?

singalsu:
The saving with this rebased pipeline2.0 adapted patch is 3.4 MCPS from sum of 10s playback average numbers

```
                            Name,        Mtrace id, avg MCPS, max MCPS
            host-copier.0.playback,   comp:0 0x40000,     3.39,    11.42
                          gain.1.1,   comp:0 0x60000,     3.28,     9.11
                         mixin.1.1,   comp:0 0x20000,     3.27,     3.45
                        mixout.2.1,   comp:1 0x30000,     2.59,     3.31
                          gain.2.1,   comp:1 0x60001,     3.37,    16.81
    dai-copier.HDA.Analog.playback,   comp:1 0x40001,     4.04,     4.28
                                  ,                 ,    19.95,    48.39

                              Name,        Mtrace id, avg MCPS, max MCPS
            host-copier.0.playback,   comp:0 0x40000,     3.40,     6.89
                          gain.1.1,   comp:0 0x60000,     2.34,     8.22
                         mixin.1.1,   comp:0 0x20000,     2.54,     2.64
                        mixout.2.1,   comp:1 0x30000,     1.74,     2.33
                          gain.2.1,   comp:1 0x60001,     2.44,     8.48
    dai-copier.HDA.Analog.playback,   comp:1 0x40001,     4.13,     4.51
                                  ,                 ,    16.60,    33.06
```

src/include/sof/audio/module_adapter/module/generic.h
```diff
@@ -184,6 +184,8 @@ struct processing_module {
 	struct output_stream_buffer *output_buffers;
 	uint32_t num_input_buffers; /**< number of input buffers */
 	uint32_t num_output_buffers; /**< number of output buffers */
+	struct comp_buffer *source; /**< single source buffer for multi_source_or_sink = false */
```
marcinszkudlinski, line 187: Please don't call it sink/source - may be confusions with struct sof_source / struct sof_sink interface

`struct comp_buffer source_comp_buffer;` 

or simiar

src/include/sof/audio/module_adapter/module/generic.h
```diff
@@ -203,6 +205,11 @@ struct processing_module {
 	 */
 	bool skip_src_buffer_invalidate;
 
+	/* False for module with one sink and one source to reduce module processing
+	 * overhead. True if multiple source or sink buffers.
+	 */
+	bool multi_source_or_sink;
```
marcinszkudlinski, line 211: bool simple_1_1_copy ?
to be consistent with module_adapter_audio_stream_copy_1to1 that is called when true

marcinszkudlinski (changes requested):
Rather minor - just propositions for new names, no pushing. Just please don't use sink/source :)

singalsu:
The PR is updated according to @marcinszkudlinski 's proposals. The boolean is inverted with new name `mod->stream_copy_single_to_single` .

src/audio/module_adapter/module_adapter.c
```diff
@@ -313,6 +313,21 @@ int module_adapter_prepare(struct comp_dev *dev)
...
+	 */
+	if (mod->num_input_buffers == 1 && mod->num_output_buffers == 1) {
+		mod->source_comp_buffer = list_first_item(&dev->bsource_list,
+							  struct comp_buffer, sink_list);
+		mod->sink_comp_buffer = sink;
+		mod->stream_copy_single_to_single = true;
+	} else {
+		mod->source_comp_buffer = NULL;
+		mod->sink_comp_buffer = NULL;
+		mod->stream_copy_single_to_single = false;
```
lyakh, line 328: `.stream_copy_1to1`?

lyakh (commented):
just a couple of very minor comments

src/audio/module_adapter/module_adapter.c
singalsu: I was thinking 1to1 looks too much like 1:1 that would give wrong impression about doing just memcpy() from buffer to buffer.

singalsu:
Previous CI round 43 successful and 2 failing checks, failed Internal Intel CI System/merge/build with West, failed 
sof-ci/jenkins/pr-device-test/main-cavs2.5  with multiple suspend resume. New try...

btian1 (approved):
seems Ranjani trying other method with your PR for #7850 #7858 
This one should be fine for merge.

Just one concern, all our optimization is trying to figure out special case and use specific function to handle it, this have benefit for this special case, however, it does increase performance for worst case and code size, I am not clear why COE team  must get those module performance low for this, if there is no low power(38.4mhz）mode, it is not must for handle module with multiple routines.

singalsu:
> seems Ranjani trying other method with your PR for #7850 #7858 This one should be fine for merge.
> 
> Just one concern, all our optimization is trying to figure out special case and use specific function to handle it, this have benefit for this special case, however, it does increase performance for worst case and code size, I am not clear why COE team must get those module performance low for this, if there is no low power(38.4mhz）mode, it is not must for handle module with multiple routines.

The previous version did not not enable the optimization, now it's using fix by @ranj063 to get the bind/unbind code to work correctly.

Yes, this is a best case optimization. But it's quite common that components have only one source and and one sink so I think this is worth it. It will be even better when pipeline 2.x let's components have linear shared buffers. Since this is small add there is not much code size overhead and it can be cleaned up when the shared buffers work is ready.

ranj063 (approved):

singalsu:
42 successful and 3 failing checks
- Zephyr / build-linux (zmain, imx8 imx8x imx8m) (pull_request)
- sof-ci/jenkins/pr-device-test/main-ace --> verify-kernel-boot-log.sh, check-pause-resume-capture-100.sh
- sof-ci/jenkins/pr-device-test/main-cavs2.5 --> check-suspend-resume-5.sh

singalsu:
SOFCI TEST

singalsu:
Now 43 successful and 2 failing checks
- Zephyr / build-linux (zmain, imx8 imx8x imx8m) 
- sof-ci/jenkins/pr-device-test/main-cavs2.5 --> check-suspend-resume-with-playback.sh timeout

singalsu:
SOFCI TEST

ranj063:
SOFCI TEST

singalsu:
SOFCI TEST

kv2019i (approved):
Thanks @singalsu , looks good now!

kv2019i:
https://sof-ci.01.org/sofpr/PR7792/build10174/devicetest/index.html looks a bit worrisome. I went though all the failures (verify-kernel-bootlog, check-runtime-pm-status , multiple-pipeline-capture ) and all are failures of https://github.com/thesofproject/sof/issues/7482 , IOW with MOD_SET_DX failing. I'm going to proceed and merge this, but we need to raise attention to #7482 .

Issue #7709, opened 2023-05-30
[CLEANUP] Legacy Fuzzer


cujomalainey:
Decide what to do about the fuzzer on  stable-v2.2 

EDIT: The beginning of the discussion is in
- #7675

marc-hb:
These should probably be updated:

https://thesofproject.github.io/latest/developer_guides/fuzzing/fuzzing_in_docker.html
https://github.com/thesofproject/sof-docs/tree/master/developer_guides/fuzzing

marc-hb:
stable-v2.2 removal submitted in
- #7795

lgirdwood:
@lyakh @juimonen did the legacy fuzzer get fully removed form main ?

marc-hb:
Yes however I'm afraid sof-docs still describes it, see above.

lgirdwood:
> Yes however I'm afraid sof-docs still describes it, see above.

ok, can you send a PR to remove from sof-docs. Thanks. I will close here.

PR #7795, opened 2023-06-13
[SKIP SOF-TEST] [STABLE-v2.2] .github: disable fuzzing

marc-hb:
This is a partial fix for #7709; doc updates still to be done.

Quoting @cujomalainey in
- #7699
> I still don't see the benefit here. This branch and IPC3 are mostly
> stable/archived. Hence anything we fuzz on main we should be able to
> cherry-pick down. Nothing new should be landing on this branch so I
> don't see the benefit of continued fuzzing past support on main.

Quoting @andyross in
- #7675
> I guess my gut says that fuzzing is a technique for validation of
> new code. It has minimal (but sure, not zero) value on maintenance
> branches that are expected to be protocol-compatible in perpetuity.


marc-hb:
Unrelated Github glitches in https://github.com/thesofproject/sof/actions/runs/5260133903/jobs/9506630809?pr=7795

Everything else green.

cujomalainey (approved):

kv2019i (approved):

PR #7808, opened 2023-06-15
buffer: prevent cache corruption

lyakh:
Recent research discovered a set of potential issues related with cache prefetch. Specifically it seems like uncached access to memory can cause cache prefetch. This can cause problems in buffer_attach() and buffer_detach() where buffers are added to or removed from lists respectively via uncached addresses, after which they can be used via cached addresses. Add proper cache synchronisation and interrupt locking to protect against such memory corruption.

### DIFF
--- a/src/audio/buffer.c
+++ b/src/audio/buffer.c
@@ -286,8 +286,15 @@ void comp_update_buffer_consume(struct comp_buffer __sparse_cache *buffer, uint3
 		 (char *)audio_stream_get_addr(&buffer->stream)));
 }
 
+/*
+ * Locking: must be called with interrupts disabled! Serialized IPCs protect us
+ * from racing attach / detach calls, but the scheduler can interrupt the IPC
+ * thread and begin using the buffer for streaming. FIXME: this is still a
+ * problem with different cores.
+ */
 void buffer_attach(struct comp_buffer *buffer, struct list_item *head, int dir)
 {
+	struct list_item *list = buffer_comp_list(buffer, dir);
 	struct list_item __sparse_cache *needs_sync;
 	bool further_buffers_exist;
 
@@ -302,11 +309,17 @@ void buffer_attach(struct comp_buffer *buffer, struct list_item *head, int dir)
 	if (further_buffers_exist)
 		dcache_writeback_region(needs_sync, sizeof(struct list_item));
 	/* The cache line can be prefetched here, invalidate it after prepending */
-	list_item_prepend(buffer_comp_list(buffer, dir), head);
+	list_item_prepend(list, head);
 	if (further_buffers_exist)
 		dcache_invalidate_region(needs_sync, sizeof(struct list_item));
+	/* no dirty cache lines exist for this buffer yet, no need to write back */
+	dcache_invalidate_region(uncache_to_cache(list), sizeof(*list));
 }
 
+/*
+ * Locking: must be called with interrupts disabled! See buffer_attach() above
+ * for details
+ */
 void buffer_detach(struct comp_buffer *buffer, struct list_item *head, int dir)
 {
 	struct list_item __sparse_cache *needs_sync_prev, *needs_sync_next;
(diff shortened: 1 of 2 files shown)


btian1:
"Recent research discovered a set of potential issues related with cache prefetch. Specifically it seems like uncached access to memory can cause cache prefetch. " 

is this reasonable? if access memory in uncached, then dsp should perform cache prefetch, right?

lyakh:
> is this reasonable? if access memory in uncached, then dsp should perform cache prefetch, right?

@btian1 https://github.com/thesofproject/sof/issues/7191#issuecomment-1591608474

btian1:
I would suggest wait for some time for #7191 , if still reproducible, we can ask keqiao help for reproduce with this PR.

src/audio/buffer.c
```diff
@@ -302,11 +309,17 @@ void buffer_attach(struct comp_buffer *buffer, struct list_item *head, int dir)
 	if (further_buffers_exist)
 		dcache_writeback_region(needs_sync, sizeof(struct list_item));
 	/* The cache line can be prefetched here, invalidate it after prepending */
-	list_item_prepend(buffer_comp_list(buffer, dir), head);
+	list_item_prepend(list, head);
 	if (further_buffers_exist)
 		dcache_invalidate_region(needs_sync, sizeof(struct list_item));
+	/* no dirty cache lines exist for this buffer yet, no need to write back */
+	dcache_invalidate_region(uncache_to_cache(list), sizeof(*list));
```
btian1, line 316: really need this invalidate? should be automatic? if need, maybe a lot of other places need as well.

src/audio/module_adapter/module_adapter.c
```diff
@@ -357,13 +358,17 @@ int module_adapter_prepare(struct comp_dev *dev)
...
 								  0, PLATFORM_DCACHE_ALIGN);
+			uint32_t flags;
+
 			if (!buffer) {
 				comp_err(dev, "module_adapter_prepare(): failed to allocate local buffer");
 				ret = -ENOMEM;
 				goto free;
 			}
 
+			irq_local_disable(flags);
```
btian1, line 369: when turn off irq, means some or one of irq have negative impact with buffer_attach.
this part of change can make buffer_attach more secure, just not sure it is must or not.

src/audio/buffer.c
lyakh: > really need this invalidate? should be automatic? if need, maybe a lot of other places need as well.

@btian1 it's our current understanding that if memory is accessed via its uncached address, it still can be prefetched. Which isn't a problem if we continue accessing that memory only via that uncached address as is usually the case for us. But if you at some point do access that memory via cache - then you get a problem. This is the case here, therefore I add cache invalidation.

src/audio/module_adapter/module_adapter.c
kv2019i: This does feel out-of-place in module_adapter.c level code. Given both IPC and pipeline are run in regular threads, we should be able to use plain mutex'es for exclusive access here. This would make it also userspace compatible for future-proofing this code.

But maybe beyond this PR, especially given pipeline_connect() already does the IRQ enable/disable dance, so the same should be done here to be safe.

kv2019i (approved):
Sorry, took a long time to review, but this is a complex piece of code to review. I'm leaning towards approving as this does problems in current codebase. See more comments inline.

src/audio/buffer.c
btian1: ok, I am not 100% understand SOF's invalidate and writeback, normally, those should be done automatically by DSP itself.
Regarding the case you mentioned previously, when need send data from memory to DMA, then it need a writeback definitely,  or in the case memory is shared between different cores.

Just a little worry, with more and more invalidate and writeback, performance will get impacted.

btian1 (approved):

kv2019i:
Known fails in https://sof-ci.01.org/sofpr/PR7808/build9654/devicetest/index.html and https://sof-ci.01.org/sofpr/PR7808/build9655/devicetest/index.html .

src/audio/module_adapter/module_adapter.c
kv2019i: @lyakh Ack. Let's revisit this when we introduce user-space, this is anyways not the only place where we have explicit irq control in the SOF audio code.

Issue #7308, opened 2023-03-21
simultaneous playback of multiple streams with MUX component not working with latest sof FW (20-march-2023)


maruthi-machani:
With latest sof FW (20-march-2023), simultaneous playbacks of 2 input streams using mux component only single stream is working. no audio playback observed on the second stream. Earlier sof FW (02-March-2023) simultaneous playback was working properly.

plbossart:
"simultaneous playbacks of 2 input streams"

You already lost me here @maruthi-machani, no idea how you managed to have 'playback' and 'input streams' in the same sentend.

Please provide clearer descriptions of 
- what platform
- what topology
- what test
you are running.

maruthi-machani:
with AMD platform, when simultaneous two input streams are played using mux component only single input stream is played. this issue is observed with the latest sof FW.
Topology uses mux component with two input streams. Attaching topology for reference.
[sof-acp-rmb-tdm8ch-dmic4ch_m4.txt](https://github.com/thesofproject/sof/files/11046953/sof-acp-rmb-tdm8ch-dmic4ch_m4.txt)
[pipe-echo-ref-dai-playback_m4.txt](https://github.com/thesofproject/sof/files/11046954/pipe-echo-ref-dai-playback_m4.txt)

dbaluta:
@maruthi-machani I bet the problem start at this commit:

commit c399624fcfb7dfaf75d69671bc2a91ec9812a6ad
Author: Seppo Ingalsuo <seppo.ingalsuo@linux.intel.com>
Date:   Tue Dec 20 18:48:10 2022 +0200

    Audio: Mux: Convert mux and demux to module adapter



We have similar issue but with MIXER.

plbossart:
> **Usecase**: simultaneous playback of two input streams on headset and speaker.

Did  you mean: playback of one stream on both headset and speaker? 

It's very hard to parse your explanations @maruthi-machani

maruthi-machani:
simultaneously playing two different streams one on headset and other on speaker.

plbossart:
Never mind, it's described in the topology where a TDM interface is used.

````
# PCM0 ----> buffer ------+
#                         |----> muxdemux ----> buffer ---->  ACPSP_play
# PCM1 ----> buffer ----->
````

plbossart:
@maruthi-machani this isn't a topology that's used on the Intel devices, headset and amplifiers use separate I2S links.

maruthi-machani:
[...]

PR #7811, opened 2023-06-15
Audio: Mux: Fix mistake in frames count handling

singalsu:
The source frames count need to be checked from every active input_buffers[n].size). Similarly in consuming the source buffers it needs to be updated to every
mod->input_buffers[n].consumed.

Fixes: #7308

### DIFF
--- a/src/audio/mux/mux.c
+++ b/src/audio/mux/mux.c
@@ -527,18 +527,24 @@ static int mux_process(struct processing_module *mod,
 	struct comp_buffer __sparse_cache *source_c;
 	struct list_item *clist;
 	const struct audio_stream __sparse_cache *sources_stream[MUX_MAX_STREAMS] = { NULL };
-	int frames;
+	int frames = 0;
 	int sink_bytes;
 	int source_bytes;
-	int i;
+	int i, j;
 
 	comp_dbg(dev, "mux_process()");
 
 	/* align source streams with their respective configurations */
+	j = 0;
 	list_for_item(clist, &dev->bsource_list) {
 		source = container_of(clist, struct comp_buffer, sink_list);
 		source_c = buffer_acquire(source);
 		if (source_c->source->state == dev->state) {
+			if (frames)
+				frames = MIN(frames, input_buffers[j].size);
+			else
+				frames = input_buffers[j].size;
+
 			i = get_stream_index(dev, cd, source_c->pipeline_id);
 			/* return if index wrong */
 			if (i < 0) {
@@ -549,13 +555,13 @@ static int mux_process(struct processing_module *mod,
 			sources_stream[i] = &source_c->stream;
 		}
 		buffer_release(source_c);
+		j++;
 	}
 
 	/* check if there are any sources active */
 	if (num_input_buffers == 0)
 		return 0;
 
-	frames = input_buffers[0].size;
 	source_bytes = frames * audio_stream_frame_bytes(mod->input_buffers[0].data);
 	sink_bytes = frames * audio_stream_frame_bytes(mod->output_buffers[0].data);
 	mux_prepare_active_look_up(cd, output_buffers[0].data, &sources_stream[0]);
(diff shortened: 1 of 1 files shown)


singalsu:
Special thanks to @juimonen for creating https://github.com/thesofproject/sof/pull/7812 and @lyakh for corrections suggestions.

src/audio/mux/mux.c
```diff
@@ -530,15 +530,18 @@ static int mux_process(struct processing_module *mod,
...
 	comp_dbg(dev, "mux_process()");
 
 	/* align source streams with their respective configurations */
+	j = 0;
+	frames = INT32_MAX;
 	list_for_item(clist, &dev->bsource_list) {
 		source = container_of(clist, struct comp_buffer, sink_list);
 		source_c = buffer_acquire(source);
 		if (source_c->source->state == dev->state) {
+			frames = MIN(frames, input_buffers[j].size);
```
ranj063, line 544: @singalsu do you think we should move this check for active state to the module adapter copy and only have active source buffers in the input_buffers? then we can simply iterate over the list of input_buffers[i].size to find the MIN and set the size for all input buffers after processing as well

src/audio/mux/mux.c
```diff
@@ -530,15 +530,18 @@ static int mux_process(struct processing_module *mod,
...
 	int sink_bytes;
 	int source_bytes;
-	int i;
+	int i, j;
 
 	comp_dbg(dev, "mux_process()");
 
 	/* align source streams with their respective configurations */
+	j = 0;
+	frames = INT32_MAX;
```
ranj063, line 539: if we do this, then we'd some sanity check for frames before processing

btian1: in case dev->bsource_list is empty

src/audio/mux/mux.c
btian1: intention was combine line 570 - 579?

singalsu: I think it's a good idea and would help the clients. Could we do such fix later since it impacts many components and use this way as quick fix?

src/audio/mux/mux.c
singalsu: Yep, need to handle that.

src/audio/mux/mux.c
```diff
@@ -530,15 +530,18 @@ static int mux_process(struct processing_module *mod,
 	int frames;
 	int sink_bytes;
 	int source_bytes;
-	int i;
```
singalsu, line 533: Should add to commit message also Fixes: c399624 ("Audio: Mux: Convert mux and demux to module adapter")

kv2019i (approved):
Minor nit we should use "Closes: #abc" to refer to bugs and issues in the commit.

But this is critical for v2.6, so let's not change for that. Looks good for me.

kv2019i:
P1 as this is needed for SOF2.6 . Please review.

src/audio/mux/mux.c
```diff
@@ -527,18 +527,24 @@ static int mux_process(struct processing_module *mod,
...
 	/* align source streams with their respective configurations */
+	j = 0;
 	list_for_item(clist, &dev->bsource_list) {
 		source = container_of(clist, struct comp_buffer, sink_list);
 		source_c = buffer_acquire(source);
 		if (source_c->source->state == dev->state) {
+			if (frames)
+				frames = MIN(frames, input_buffers[j].size);
+			else
+				frames = input_buffers[j].size;
```
plbossart, line 546: this is rather odd, why not initialize frame to MAX_INT and do the line 544 unconditionally?

kv2019i: @plbossart This was in previous version of the patch, but bsource_list may be empty in which case we'd end up with frames==MAX_INT

lgirdwood (approved):

lgirdwood:
@wszypelt @lrudyX this PR is for IPC3 and not tested on Intel CI. Merging.

PR #7812, opened 2023-06-15
topology1: test topology for tgl nocodec mux

juimonen:
Add test topology for muxing 2 pipelines into 1 with SSP dai and capturing the result from loopback.


singalsu (approved):
Thanks! Works, and very useful for testing a case that we missed.

lgirdwood (approved):

PR #7273, opened 2023-03-14
dai-zephyr: possible division-by-zero if max_block_count is zero

kv2019i:
If the max_block_count attribute is zero, this may lead to division by zero error. Handle this explicitly but reporting error if max_block_count is zero.

PR #7815, opened 2023-06-16
dai-zephyr: check if max_block_count is zero

abonislawski:
Just a copy of #7273 which added this check only for dai_capture_params

### DIFF
--- a/src/audio/dai-zephyr.c
+++ b/src/audio/dai-zephyr.c
@@ -639,6 +639,11 @@ static int dai_playback_params(struct dai_data *dd, struct comp_dev *dev, uint32
 			goto out;
 		}
 
+		if (!max_block_count) {
+			comp_err(dev, "dai_playback_params(): invalid max-block-count of zero");
+			goto out;
+		}
+
 		if (max_block_count < period_count) {
 			comp_dbg(dev, "dai_playback_params(): block count = %d not supported by DMA", period_count);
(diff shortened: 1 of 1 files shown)


src/audio/dai-zephyr.c
```diff
@@ -639,6 +639,11 @@ static int dai_playback_params(struct dai_data *dd, struct comp_dev *dev, uint32
 			goto out;
 		}
 
+		if (!max_block_count) {
```
btian1, line 642: better to combine this with line 637 with:
if (err < 0 || !max_block_count) {

since they have same cause from dma_get_attribute

lyakh: two different conditions though, deserve separate handling

lyakh (approved):

dnikodem (approved):

kv2019i:
Hmm, there are two failures in Intel 12210252 quickbuild. 124_20_TestTwoHdaStreams and 116_00_TestSdwTransferLoopback . 
A bit confusingly, the test_results json file has "passed" as result, so it's not immediately apparent why the test is marked as failed. @wszypelt ?

tmleman (approved):

kv2019i:
DUT config issue in  https://sof-ci.01.org/sofpr/PR7815/build9693/devicetest/index.html , not related to this PR. Proceeding with merge.

PR #7816, opened 2023-06-16
[sof-2.6] audio: buffer: do not mix cached and non-cached access in buffer_alloc

kv2019i:
No reason to mix cached and noncached operations when initializing the buffer in buffer_alloc(). We take a cached reference, so let's use that to initialize the object.

(cherry picked from commit a209cbd31f9a1a85c21568d12566c4220d573c00)


kv2019i:
One PM fail in https://sof-ci.01.org/sofpr/PR7816/build9704/devicetest , otherwise good. Proceeding with merge.

PR #7819, opened 2023-06-16
[DNM] pipeline 2.0: Introducing Data Processing Queue 

marcinszkudlinski:
DP queue is a circular buffer providing safe consumer/producer cached operations cross cores

DP Queue is the only buffering that modules scheduled as DP can use 

Modules can access Dp Queue using sink/source API

Advantages:
 - DP queue does not require any locking - shared part is kept in shared non cached memory (faster than writebacks)
 - DP queue provides cached buffers to modules for faster processing
 - data consumer and producer may safely use it independently at the same time, even from diffrent cores
 - performance: there's only one cache writeback and no additional copying inside

prerequisites:
 1) incoming and outgoing data rate MUST be the same
 2) Both data consumer and data producer declare max chunk sizes
    they want to use (IBS/OBS)
    any requests of bigger data than IBS/OBS will be rejected

 The queue may work in several modes
 1) simple mode
    in case both receiver and sender are located on the
    same core and cache coherency does not matter.
    In this case DP Queue is a simple ring buffer, with IBS and OBS

 2) shared mode
   In this case we need to writeback cache when new
   data arrive and invalidate cache on secondary core.
   That means the whole cacheline must be used exclusively
   by sink or by source, incoming data will be available to use when a
   cacheline is filled completely

TODO - not in this PR - data linearization mode
In this mode the buffers provided to the modules will be linear. 
That will require one extra memcopy operation

### DIFF
--- a/src/audio/dp_queue.c
+++ b/src/audio/dp_queue.c
@@ -1,0 +1,328 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// Copyright(c) 2021 Intel Corporation. All rights reserved.
+//
+
+#include <sof/audio/dp_queue.h>
+#include <sof/audio/sink_api_implementation.h>
+#include <sof/audio/source_api_implementation.h>
+#include <sof/audio/audio_stream.h>
+
+#include <rtos/alloc.h>
+#include <ipc/topology.h>
+
+/**
+ * sink and source structures are cached and belong exclusively to a single core
+ * they must not be a part of coherent dp_queue structure
+ */
+struct dp_queue_sink_source_c {
+	struct dp_queue *owner; /**< pointer to uncached dp_queue structure */
+	struct sof_source source;	/**< src api handler */
+	struct sof_sink sink;	/**< sink api handler */
+};
+
+struct dp_queue {
+	/* sink/src structures, located in cached memory */
+	struct dp_queue_sink_source_c __sparse_cache *sink_src_c;
+
+	uint32_t ibs;
+	uint32_t obs;
+	uint32_t flags;
+
+	size_t data_buffer_size;
+	uint8_t *data_buffer;
+
+	uint32_t write_offset;
+	uint32_t read_offset;
+	uint32_t available_data; /* amount of data ready for immediate reading */
+	/* free buffer space
+	 * NOTE!
+	 *  - when dp queue is shared, available_data + free_space DOES NOT eq data_buffer_size
+	 *  - when dp queue is not shared available_data + free_space always == data_buffer_size
+	 */
+	uint32_t free_space;
+
+	struct sof_audio_stream_params audio_stream_params;
+
+	struct k_spinlock lock;
+};
+
+static inline bool dp_queue_is_shared(struct dp_queue *dp_queue)
+{
+	return !!(dp_queue->flags & DP_QUEUE_MODE_SHARED);
+}
+
+static inline uint8_t *dp_queue_buffer_end(struct dp_queue *dp_queue)
+{
+	return dp_queue->data_buffer + dp_queue->data_buffer_size;
+}
+
+struct dp_queue *dp_queue_get_from_sink(struct sof_sink __sparse_cache *sink)
+{
+	struct dp_queue_sink_source_c __sparse_cache *sink_src_c =
+		attr_container_of(sink, struct dp_queue_sink_source_c __sparse_cache,
+				  sink, __sparse_cache);
+
+	return sink_src_c->owner;
+}
+
+struct dp_queue *dp_queue_get_from_source(struct sof_source __sparse_cache *source)
+{
+	struct dp_queue_sink_source_c __sparse_cache *sink_src_c =
+		attr_container_of(source, struct dp_queue_sink_source_c __sparse_cache,
+				  source, __sparse_cache);
+	return sink_src_c->owner;
+}
+
+static inline void dp_queue_invalidate_shared(struct dp_queue *dp_queue,
+					      void __sparse_cache *ptr, uint32_t size)
+{
+	/* rollback? */
+	if ((uintptr_t)ptr + size > (uintptr_t)dp_queue_buffer_end(dp_queue)) {
+		/* writeback till the end of circular buffer */
+		dcache_invalidate_region
+			(ptr, (uintptr_t)dp_queue_buffer_end(dp_queue) - (uintptr_t)ptr);
+		size -= (uintptr_t)dp_queue_buffer_end(dp_queue) - (uintptr_t)ptr;
+	}
+	/* invalidate rest of data - from begin of buffer, remaining size */
+	dcache_invalidate_region((__sparse_force void __sparse_cache *)dp_queue->data_buffer, size);
+}
+
+static inline k_spinlock_key_t dp_queue_lock(struct dp_queue *dp_queue)
+{
+	return k_spin_lock(&dp_queue->lock);
+}
+
+static inline void dp_queue_unlock(struct dp_queue *dp_queue, k_spinlock_key_t key)
+{
+	k_spin_unlock(&dp_queue->lock, key);
+}
+
+struct sof_sink __sparse_cache *dp_queue_get_sink(struct dp_queue *dp_queue)
+{
+	return &dp_queue->sink_src_c->sink;
+}
+
+struct sof_source __sparse_cache *dp_queue_get_source(struct dp_queue *dp_queue)
+{
+	return &dp_queue->sink_src_c->source;
+}
+
+static size_t dp_queue_get_free_size(struct sof_sink __sparse_cache *sink)
+{
+	struct dp_queue *dp_queue = dp_queue_get_from_sink(sink);
+
+	return dp_queue->free_space;
+}
+
+static int dp_queue_get_buffer(struct sof_sink __sparse_cache *sink, size_t req_size,
+			       void **data_ptr, void **buffer_start, size_t *buffer_size)
+{
+	struct dp_queue *dp_queue = dp_queue_get_from_sink(sink);
+
+	/* dp_queue_get_free_size will return free size with adjustment for cacheline if needed */
+	if (req_size > dp_queue->ibs || req_size > dp_queue_get_free_size(sink))
+		return -ENODATA;
+
+	/* no need to lock, just reading data that may be modified by commit_buffer only */
+	*data_ptr = dp_queue->data_buffer + dp_queue->write_offset;
+	*buffer_start = dp_queue->data_buffer;
+	*buffer_size = dp_queue->data_buffer_size;
+
+	/* provided buffer is an empty space, the requestor will perform write operations only
+	 * no need to invalidate cache - will be overwritten anyway
+	 */
+	return 0;
+}
+
+static int dp_queue_commit_buffer(struct sof_sink __sparse_cache *sink, size_t commit_size)
+{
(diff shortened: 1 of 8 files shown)


cujomalainey:
@marcinszkudlinski can you leave this as draft until all base PRs have landed?

marcinszkudlinski:
> @marcinszkudlinski can you leave this as draft until all base PRs have landed?

I'd rather mark it as "do not merge" and squash all commit from prev PR to one
Nobody looks at drafts :( and I hope to have some feedback quickly

marcinszkudlinski:
https://github.com/thesofproject/sof/pull/7622 has been merged
rebased on newest main

marc-hb:
> Nobody looks at drafts :( and I hope to have some feedback quickly

I think you're confusing two different things.

The only way to ignore drafts is to explicitly filter them out. Maintainers like @cujomalainey explicitly filter out drafts because they have a LOT of PRs to go through. It's easier for them and their time is rarer and more precious than the time of submitters.

Other people don't filter out drafts. It's true that no one gets notified if you CREATE the PR as a draft the first time (useful feature). BUT if you demote a non-draft to a draft then everyone is _still_ notified. Demoting does NOT unsubscribe.

kv2019i (commented):
The need to create queues in adapter in "dp: add dp_queue to module adapter processing " is a bit funky, but the commit describe the intent well and the there is ample inline commentary, so I don't really see blockers with this.

lyakh:
> > @marcinszkudlinski can you leave this as draft until all base PRs have landed?
> 
> I'd rather mark it as "do not merge" and squash all commit from prev PR to one Nobody looks at drafts :( and I hope to have some feedback quickly

@marcinszkudlinski I do look at drafts

lyakh:
@marcinszkudlinski thanks for a nice explanatory text! Could you please explain one more thing: why is our current buffer implementation unsuitable? It's also a ring buffer, so AFAICS it's rather similar to what you're proposing. It *currently* doesn't have all the features that you're describing, but wouldn't it be better to see if those features can be added? Then, also, you're saying that the DP queue can only be used by DP modules. But DP modules are usually linked to "normal" components. What buffers should be used between a DP and a "normal" module? So these buffers are only used if you have at least 2 DP modules linked?

kv2019i:
@lyakh wrote:
> @marcinszkudlinski I do look at drafts. Let's really make it unmergeable somehow - either by making it a draft or I can gladly give it a "-1" just for that ;-) No, not out of any ill intent but we can mark it as -1 if you prefer - just to make accidental merging impossible and let everybody see, that this PR isn't going to be merged tomorrow.

DNM is sufficient to prevent merge. We've used it already to tag PRs that are ready to go review/content wise, but need to wait for some dependency (e.g. Zephyr) to go in. Maybe this warrants a wider discussion (not in this PR), but I do not usually look at drafts and I consider draft-status an indication the PR is not for general consumption. 

This is clearly a PR that is ready for review so I think non-draft is correct status for this.

marcinszkudlinski:
Current implementation may be modified, but I don't see a point to do it - it is already complicated and adding more features may be difficult. The hardest problem I see is buffer aquiring, which is really hard to get rid of because all modules (and other instances) do aquire/release by themselves. 

Sink/src has been introduced to make buffering internals independent from modules, with no additional steps required from the data producers/consumers. That makes possible to create a few specialized buffering implementation instead of a huge unversal one, and some kind of pipeline buffer factory (speaking in C++ language :) ) - modules may provide vector of requirements, factory would chose and configure an implementation that suits both sides - and receive a prepared and ready to use handlers to sink/src. Or similar sollution - to be decided 


As to DP/LL modules:

DP modules must have DP queue on both sides (in/out data), because only DpQueue currently can provide all features required 
for dp processing (cross core, no locking, diffrent IBS/OBS)
 
BUT there are no objections to connect LLs using DpQueue as well. Its just waste of resources as LLs don't need most of the features. In fact a single LL produce data that must be 100% consumed in the same cycle by next LL in pipeline, so shared buffer should be used here (another implementation of sink/source - very very simple one)

The example of pipeline in "2.0 way" may look like

DAI (source) --> **LL1** -->(sink) LL Shared Buffer1(source) -->**LL2** ->(sink)DpQueue(source) --> **DP1** ---> (sink) DpQueue (source)  --> **LL3** -->(sink) DAI 

DpQueue is also a smooth way to connect 2 pipelines running on separate cores

DAI (source) --> **LL1 prim core** -->(sink) LL Shared Buffer1(source) -->**LL2  prim core** ->(sink)DpQueue(source) -->**LL1secCore** ->(sink) LL Shared Buffer2(source) --> **LL2secCore** ->(sink) LL Shared Buffer2(source) --> **LL3secCore** ->(sink) DAI

src/audio/dp_queue.c
```diff
@@ -0,0 +1,328 @@
...
+	return dp_queue->free_space;
+}
+
+static int dp_queue_get_buffer(struct sof_sink __sparse_cache *sink, size_t req_size,
+			       void **data_ptr, void **buffer_start, size_t *buffer_size)
+{
+	struct dp_queue *dp_queue = dp_queue_get_from_sink(sink);
+
+	/* dp_queue_get_free_size will return free size with adjustment for cacheline if needed */
+	if (req_size > dp_queue->ibs || req_size > dp_queue_get_free_size(sink))
```
lyakh, line 124: isn't free size never greater than `.ibs`? If so, just the second half of the check is enough

src/audio/dp_queue.c
```diff
@@ -0,0 +1,328 @@
...
+		return -ENODATA;
+
+	/* no need to lock, just reading data that may be modified by commit_buffer only */
+	*data_ptr = dp_queue->data_buffer + dp_queue->write_offset;
+	*buffer_start = dp_queue->data_buffer;
+	*buffer_size = dp_queue->data_buffer_size;
+
+	/* provided buffer is an empty space, the requestor will perform write operations only
+	 * no need to invalidate cache - will be overwritten anyway
+	 */
```
lyakh, line 134: so is the buffer accessed via cached addresses? Then you need `__sparse_cache` for all those addresses. I think both for locking and cache synchronisation the reader will have to do the hard part. But this is the same situation here now as with the current buffer: currently your DP write and read pointers are uncached. So, frequently reading and writing them by both the reader and the writer will be slow. And if you make it cached you'll have to synchronise

src/audio/dp_queue.c
```diff
@@ -0,0 +1,328 @@
...
+				dcache_writeback_invalidate_region
+					((__sparse_force void __sparse_cache *)ptr,
+					PLATFORM_DCACHE_ALIGN);
+
+				/* mark data as available to read */
+				dp_queue->available_data += PLATFORM_DCACHE_ALIGN;
+				/* get next cacheline */
+				current_cacheline = (current_cacheline + 1) %
+						    total_num_of_cachelines;
+			}
```
lyakh, line 182: this whole loop should be a single call to `dcache_writeback_invalidate_region()` or 2 of them at most, if wrapping

src/audio/dp_queue.c
```diff
@@ -0,0 +1,328 @@
...
+			/* overlap? */
+			last_cacheline %= total_num_of_cachelines;
+
+			/* now go one by one.
+			 * if current_cacheline == last_full_cacheline - nothing to do
+			 */
+			while (current_cacheline != last_cacheline) {
+				/* writeback / invalidate */
+				uint8_t *ptr = dp_queue->data_buffer +
+						(current_cacheline * PLATFORM_DCACHE_ALIGN);
```
lyakh, line 172: so, looks like `.data_buffer` does have to be cached, so a lot of the code has to be changed.

lyakh (changes requested):
I've stopped after the first 2 commits for now, addressing those comments will force changes in the last 2 commits too presumably. If you like you can make a separate PR with the first 2 commits only

src/audio/module_adapter/module_adapter.c
```diff
@@ -169,183 +169,63 @@ static int module_adapter_sink_src_prepare(struct comp_dev *dev)
...
 		struct comp_buffer *source_buffer_uc;
 
 		source_buffer_uc = container_of(blist, struct comp_buffer, sink_list);
-		source_buffers_c[num_of_sources] = buffer_acquire(source_buffer_uc);
-		audio_src[num_of_sources] =
-				audio_stream_get_source(&source_buffers_c[num_of_sources]->stream);
-		source_reset_num_of_processed_bytes(audio_src[num_of_sources]);
-		num_of_sources++;
+		source_buffers_c[i] = buffer_acquire(source_buffer_uc);
+		mod->sources[i] = audio_stream_get_source(&source_buffers_c[i]->stream);
```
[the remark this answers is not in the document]
ranj063: what happens if a modules are disconnected during unbind?

ranj063 (commented):
nice PR @marcinszkudlinski. I have the same conern as @lyakh about storing the source/sink acquired pointers and how we deal with module unbind at runtime

src/audio/module_adapter/module_adapter.c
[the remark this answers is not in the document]
marcinszkudlinski: in this case I can agree with this.. Will clear all mod->sources[] mod->sinks[] after use. Just using couple of cycles and may sometimes be safer
This is a limitation of audio_stream - as long as it need to be aquired/released. Alternative is to aquire/release each time th halder is used (inside of audio_stream itself). In my opition - too expensive. 

in case of DpQueue the handlers may be kept safely.

src/audio/dp_queue.c
```diff
@@ -0,0 +1,328 @@
...
+#include <sof/audio/audio_stream.h>
+
+#include <rtos/alloc.h>
+#include <ipc/topology.h>
+
+/**
+ * sink and source structures are cached and belong exclusively to a single core
+ * they must not be a part of coherent dp_queue structure
+ */
+struct dp_queue_sink_source_c {
```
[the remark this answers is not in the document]
marcinszkudlinski: I added it to make more clear that this structure contains all dp_q runtime that may safely be cached. Ok, will change

src/audio/module_adapter/module_adapter.c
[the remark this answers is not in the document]
lyakh: > This is a limitation of audio_stream - as long as it need to be aquired/released. Alternative is to aquire/release each time th halder is used (inside of audio_stream itself). In my opition - too expensive.

If running on a single core it's cheap - as long as asserts are disabled it's a single pointer conversion when acquiring

> in case of DpQueue the handlers may be kept safely.

Sorry, kept for how long? Not over multiple `.copy()` runs? And in fact DP must be preemptive, so actually they shouldn't be kept for too long anyway

src/audio/module_adapter/module_adapter.c
```diff
@@ -386,6 +390,107 @@ static int module_adapter_legacy_prepare(struct comp_dev *dev)
...
+
+err:
+	for (int i = 0; i < mod->num_of_sources; i++)
+		if (mod->sources[i]) {
+			struct dp_queue *dp_queue = dp_queue_get_from_source(mod->sources[i]);
+
+			dp_queue_free(dp_queue);
+			mod->sources[i] = NULL;
+		}
+	mod->num_of_sinks = 0;
```
RanderWang, line 475: "mod->num_of_sources = 0;" ?

marcinszkudlinski:
I'm closing this PR - will split it to smaller parts

PR #7824, opened 2023-06-16
Mixin/Mixout Optimization for simple pipelines

ranj063:
Includes the patch from #7792 

For a pipeline like: host-copier->gain->mixin->mixout->gain->dai-copier
The original MCPS (Measured on TGL chromebook) is:
```
            host-copier.2.playback,   comp:0 0x40000,     3.46,     6.89
                          gain.5.1,   comp:0 0x60000,     2.17,     9.22
                         mixin.5.1,   comp:0 0x20000,     2.62,     5.94
                        mixout.6.1,   comp:1 0x30000,     1.89,     2.42
                          gain.6.1,   comp:1 0x60001,     2.22,    16.57
 dai-copier.SSP.NoCodec-2.playback,   comp:1 0x40001,     6.05,     6.41
                                  ,                 ,    18.42,    47.44

```
And the optimized performance is:
```
                              Name,        Mtrace id, avg MCPS, max MCPS
            host-copier.2.playback,   comp:0 0x40000,     3.42,     6.86
                          gain.5.1,   comp:0 0x60000,     2.19,     9.24
                         mixin.5.1,   comp:0 0x20000,     0.80,     0.82
                        mixout.6.1,   comp:1 0x30000,     2.14,     2.44
                          gain.6.1,   comp:1 0x60001,     2.19,     9.34
 dai-copier.SSP.NoCodec-2.playback,   comp:1 0x40001,     5.90,    13.75
                                  ,                 ,    16.64,    42.46

```

### DIFF
--- a/src/audio/mixin_mixout/mixin_mixout.c
+++ b/src/audio/mixin_mixout/mixin_mixout.c
@@ -125,7 +125,6 @@ static int mixin_init(struct processing_module *mod)
 				    mod->priv.cfg.base_cfg.audio_fmt.s_type);
 
 	dev->ipc_config.frame_fmt = frame_fmt;
-	dev->bypass_capable = true;
 
 	return 0;
 }
--- a/src/audio/mixin_mixout/mixin_mixout.c
+++ b/src/audio/mixin_mixout/mixin_mixout.c
@@ -92,6 +92,10 @@ struct mixout_data {
 	 * by mixin but not yet produced in mixout.
 	 */
 	uint32_t pending_frames[MIXOUT_MAX_SOURCES];
+
+	void (*mix_func)(struct comp_dev *dev, struct audio_stream __sparse_cache *sink,
+			 const struct audio_stream __sparse_cache **sources, uint32_t count,
+			 uint32_t frames);
 };
 
 static int mixin_init(struct processing_module *mod)
@@ -121,14 +125,14 @@ static int mixin_init(struct processing_module *mod)
 				    mod->priv.cfg.base_cfg.audio_fmt.s_type);
 
 	dev->ipc_config.frame_fmt = frame_fmt;
-	mod->skip_src_buffer_invalidate = true;
+	dev->bypass_capable = true;
 
 	return 0;
 }
 
 static int mixout_init(struct processing_module *mod)
 {
-	struct module_source_info __sparse_cache *mod_source_info;
+	struct module_data *mod_data = &mod->priv;
 	struct comp_dev *dev = mod->dev;
 	struct mixout_data *mo_data;
 	enum sof_ipc_frame frame_fmt, valid_fmt;
@@ -139,9 +143,7 @@ static int mixout_init(struct processing_module *mod)
 	if (!mo_data)
 		return -ENOMEM;
 
-	mod_source_info = module_source_info_acquire(mod->source_info);
-	mod_source_info->private = mo_data;
-	module_source_info_release(mod_source_info);
+	mod_data->private = mo_data;
 
 	audio_stream_fmt_conversion(mod->priv.cfg.base_cfg.audio_fmt.depth,
 				    mod->priv.cfg.base_cfg.audio_fmt.valid_bit_depth,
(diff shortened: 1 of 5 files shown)


src/audio/module_adapter/module_adapter.c
```diff
@@ -825,6 +871,67 @@ static int module_adapter_simple_copy(struct comp_dev *dev)
...
+	int ret;
+
+	source_c = buffer_acquire(mod->source);
+	source_dev = source_c->source;
+	if (source_dev->source_bypass_capable &&
+	    !module_adapter_is_multi_sink_source(source_dev)) {
+		struct comp_buffer *source;
+
+		source = list_first_item(&source_dev->bsource_list, struct comp_buffer,
+					 sink_list);
```
andrula-song, line 890: can we add a mod->source = source here. In case we may have pipeline like copier-> volume(bypass) -> mixin(bypass) ->mixout. If always take the last component's source as source may lead to unexpected error.

src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -482,14 +494,53 @@ static int mixout_process(struct processing_module *mod,
...
+		 */
+		if (source_buffer->sink != mod->dev) {
+			uint32_t channels = audio_stream_get_channels(output_buffers[0].data);
+
+			audio_stream_copy(input_buffers[0].data, 0, output_buffers[0].data, 0,
+					  frames * channels);
+
+			bytes = frames * audio_stream_frame_bytes(output_buffers[0].data);
+			input_buffers[0].consumed = bytes;
+			output_buffers[0].size = bytes;
```
andrula-song, line 540: can we set mod->dev->source_bypass_capable = true; and store mod->source & mod->sink  here? transfer pointer is faster than audio stream copy.

src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -482,14 +494,53 @@ static int mixout_process(struct processing_module *mod,
...
+		source_buffer = attr_container_of(input_buffers[0].data,
+						  struct comp_buffer __sparse_cache,
+						  stream, __sparse_cache);
+		/*
+		 * When the mixin module is bypassed, the source buffer points to the mixin's
+		 * source buffer. In this case, simply copy the data from the mixin's source
+		 * buffer into the sink buffer. Otherwise, continue on to the normal mixout
+		 * processing below.
+		 */
+		if (source_buffer->sink != mod->dev) {
```
lyakh, line 532: this looks suspicious. If we add generic bypass mechanism, then individual modules shouldn't need to identify such cases, it should be transparent to them. Or maybe we should always call module's `.process()` method and those modules, that can and should be bypassed can indicate a short-circuit condition in which case the module adapter calls the next `.process()` with the old input?

lyakh (changes requested):

src/audio/module_adapter/module_adapter.c
andrula-song: actually, I think if there is only one mixin and one mixout, then the mixin and mixout both can be "bypass". and since we trigger processing from pipeline start to end, I don't think we need a loop here, if we always update the mod->source is its last component is bypass component.

src/audio/mixin_mixout/mixin_mixout.c
ranj063: same problem with not all modules not using the module adapter.

src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -482,14 +494,53 @@ static int mixout_process(struct processing_module *mod,
...
+		struct comp_buffer __sparse_cache *source_buffer;
+		uint32_t frames, bytes;
+
+		if (num_input_buffers == 0) {
+			mixout_generate_silence(mod, &output_buffers[0]);
+			return 0;
+		}
+
+		frames = audio_stream_avail_frames_aligned(input_buffers[0].data,
+							   output_buffers[0].data);
```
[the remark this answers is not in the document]
andrula-song: from your code line 814:
	frames = audio_stream_avail_frames_aligned(&source_c->stream, &sink_c->stream);
	mod->input_buffers[0].size = frames;
audio_stream_avail_frames_aligned return the MIN(avail, free).

lyakh (commented):
not a real review (yet), just a couple of questions, that occurred to me

PR #7753, opened 2023-06-06
[Not For Main]Audio: Volume: Add bypass functions

andrula-song:
Add bypass functions for volume. If gain of all channels equals 0dB, then we use bypass functions to transport the stream data.

PR #7828, opened 2023-06-19
Audio: Volume: Add bypass functions

andrula-song:
cherry-pick from topic/performance branch https://github.com/thesofproject/sof/pull/7753
Add bypass functions for volume. If gain of all channels equals 0dB, then we use bypass functions to transport the stream data.


andrula-song:
hi @wszypelt , can you help to trigger CI again? thanks!

andrula-song:
SOFCI TEST

src/audio/module_adapter/module/volume/volume_hifi4.c
```diff
@@ -0,0 +1,343 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// Copyright(c) 2018 Intel Corporation. All rights reserved.
```
btian1, line 3: change to 2023

src/audio/module_adapter/module/volume/volume_hifi4.c
```diff
@@ -0,0 +1,343 @@
...
+//
+// Author: Tomasz Lauda <tomasz.lauda@linux.intel.com>
```
btian1, line 5: author also need change

src/audio/module_adapter/module/volume/volume_hifi4.c
```diff
@@ -0,0 +1,343 @@
...
+
+/**
+ * \file
+ * \brief Volume HiFi3 processing implementation without peak volume detection
```
btian1, line 9: HIFI4

src/audio/module_adapter/module/volume/volume_hifi4_with_peakvol.c
```diff
@@ -0,0 +1,379 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// Copyright(c) 2022 Intel Corporation. All rights reserved.
```
btian1, line 3: 2022 --> 2023

src/audio/module_adapter/module/volume/volume_generic.c
```diff
@@ -93,6 +93,47 @@ static void vol_s24_to_s24(struct processing_module *mod, struct input_stream_bu
...
+	y = audio_stream_wrap(sink, (char *)audio_stream_get_wptr(sink) + bsink->size);
+
+	bsource->consumed += VOL_S32_SAMPLES_TO_BYTES(remaining_samples);
+	bsink->size += VOL_S32_SAMPLES_TO_BYTES(remaining_samples);
+	while (remaining_samples) {
+		nmax = audio_stream_samples_without_wrap_s24(source, x);
+		n = MIN(remaining_samples, nmax);
+		nmax = audio_stream_samples_without_wrap_s24(sink, y);
+		n = MIN(n, nmax);
+		memcpy_s(y, n * sizeof(int32_t), x, n * sizeof(int32_t));
```
btian1, line 131: size is same, use memcpy to save one more parameter calculation and passing?

andrula-song: memcpy_s is recommended since it is more safety.

src/audio/module_adapter/module/volume/volume_hifi4_with_peakvol.c
andrula-song: I didn't rewrite this document, so I don't think need to change like this.

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -971,6 +1002,20 @@ static int volume_set_volume(struct processing_module *mod, const uint8_t *data,
...
+	if (!cd->ramp_finished) {
+		cd->is_bypass = false;
+	} else {
+		for (i = 0; i < channels_count; i++) {
+			if (cd->volume[i] != VOL_ZERO_DB) {
+				cd->is_bypass = false;
+				break;
+			}
+		}
+	}
```
lyakh, line 1015: could be
```
cd->is_bypass = cd->ramp_finished;

for (i = 0; i < channels_count; i++) {
	if (cd->volume[i] != VOL_ZERO_DB) {
		cd->is_bypass = false;
		break;
	}
}
```

lyakh (commented):
how is this PR aligned with #7824 ? Both PRs use the word "bypass" but AFAICS the meaning is quite different: here it means copying data without processing, in #7824 it means passing a pointer with no copying at all. If we do take the approach in that PR and implement that kind of optimisation, supposedly it would be better to use it for volume too? Then this copying PR shouldn't be needed at all?

andrula-song:
> how is this PR aligned with #7824 ? Both PRs use the word "bypass" but AFAICS the meaning is quite different: here it means copying data without processing, in #7824 it means passing a pointer with no copying at all. If we do take the approach in that PR and implement that kind of optimisation, supposedly it would be better to use it for volume too? Then this copying PR shouldn't be needed at all?

but for volume with peak meter, we have to find the max value from the input, this "bypass" means do not do the gain multiplication . But if we don't enable peak meter, we can use the function of https://github.com/thesofproject/sof/pull/7824.

src/audio/module_adapter/module/volume/volume_generic.c
```diff
@@ -147,6 +188,46 @@ static void vol_s32_to_s32(struct processing_module *mod, struct input_stream_bu
...
+	int remaining_samples = frames * nch;
+
+	x = audio_stream_wrap(source, (char *)audio_stream_get_rptr(source) + bsource->consumed);
+	y = audio_stream_wrap(sink, (char *)audio_stream_get_wptr(sink) + bsink->size);
+	bsource->consumed += VOL_S32_SAMPLES_TO_BYTES(remaining_samples);
+	bsink->size += VOL_S32_SAMPLES_TO_BYTES(remaining_samples);
+	while (remaining_samples) {
+		nmax = audio_stream_samples_without_wrap_s32(source, x);
+		n = MIN(remaining_samples, nmax);
+		nmax = audio_stream_samples_without_wrap_s32(sink, y);
```
kv2019i, line 223: @andrula-song @singalsu 
Hmm,  audio_stream_samples_without_wrap_s24() is identical to audio_stream_samples_without_wrap_s32(), so effectively vol_passthrough_s32_to_s32() and vol_passthrough_s24_to_s24 are duplicated code. Can we just use one fucntion instead? 

And it seems generic copy seems like something we use in other audio functions, so do we have this function already somewhere else we could just call?

kv2019i (approved):
One case of duplicated code, but as that is in the generic reference version, not blocking the PR for that. But please do check.

lyakh:
> > how is this PR aligned with #7824 ? Both PRs use the word "bypass" but AFAICS the meaning is quite different: here it means copying data without processing, in #7824 it means passing a pointer with no copying at all. If we do take the approach in that PR and implement that kind of optimisation, supposedly it would be better to use it for volume too? Then this copying PR shouldn't be needed at all?
> 
> @lyakh the problem is that not all components use the module adapter yet, esp the copier. So if the gain is just before the DAI copier, then we cannot use the bypass option in the module adapter. As for using the term here, I'd replace it with passthrough as we;re not really bypassing the module here

@ranj063 arrrgh... So module-adapter bypass is only possible when the downstream component is using module-adapter?.. Maybe we should bite the bullet and do this at the component level. This will presumably require changes to all components, because currently they all grab buffers from their lists directly. Then for all module adapter components the module adapter layer will handle it transparently

src/audio/module_adapter/module/volume/volume_hifi3.c
```diff
@@ -133,6 +133,55 @@ static void vol_s24_to_s24_s32(struct processing_module *mod, struct input_strea
...
+
+/**
+ * \brief HiFi3 enabled volume passthrough from 24/32 bit to 24/32 or 32 bit.
+ * \param[in,out] dev Volume base component device.
+ * \param[in,out] sink Destination buffer.
+ * \param[in,out] source Input buffer.
+ * \param[in] frames Number of frames to process.
+ * \param[in] attenuation factor for peakmeter adjustment (unused)
+ */
+static void vol_passthrough_s24_to_s24_s32(struct processing_module *mod,
```
singalsu, line 145: Could these volume processing functions names be cleaned up to simple s16/s24/s32. The old names are from days when volume was used for format conversion. Now with same input and output forward always we could make the names simpler.

singalsu:
> > > how is this PR aligned with #7824 ? Both PRs use the word "bypass" but AFAICS the meaning is quite different: here it means copying data without processing, in #7824 it means passing a pointer with no copying at all. If we do take the approach in that PR and implement that kind of optimisation, supposedly it would be better to use it for volume too? Then this copying PR shouldn't be needed at all?
> > 
> > 
> > @lyakh the problem is that not all components use the module adapter yet, esp the copier. So if the gain is just before the DAI copier, then we cannot use the bypass option in the module adapter. As for using the term here, I'd replace it with passthrough as we;re not really bypassing the module here
> 
> @ranj063 arrrgh... So module-adapter bypass is only possible when the downstream component is using module-adapter?.. Maybe we should bite the bullet and do this at the component level. This will presumably require changes to all components, because currently they all grab buffers from their lists directly. Then for all module adapter components the module adapter layer will handle it transparently

The bypass with volume is more complicated. It's enabled on the fly when all channels have gain of exactly 1. E.g. in start of stream gain is -90 dB and after ramp to 0 dB the mode is switched. If any of gains is touched the is changed to process the non-1 gain.

singalsu (approved):

src/audio/module_adapter/module/volume/volume_hifi3.c
singalsu: Also  all non-module adapter mode code should be removed from volume, but not a task for this PR.

src/audio/module_adapter/module/volume/volume_hifi3_with_peakvol.c
```diff
@@ -21,348 +21,9 @@ LOG_MODULE_DECLARE(volume_hifi3, CONFIG_SOF_LOG_LEVEL);
...
-		m = MAX(cd->peak_vol[i], cd->peak_vol[i + channels_count]);
-		m = MAX(m, cd->peak_vol[i + channels_count * 2]);
-		m = MAX(m, cd->peak_vol[i + channels_count * 3]);
-		cd->peak_regs.peak_meter[i] = m;
-	}
-}
-#endif /* CONFIG_FORMAT_S16LE */
-
-#else /* HIFI3 VERSION */
+#if CONFIG_COMP_PEAK_VOL
```
lyakh, line 26: at least we could select peak-volume specific files in CMakeLists.txt?

kv2019i:
Known fail with dmic rec case in one DUT in https://sof-ci.01.org/sofpr/PR7828/build9915/devicetest/index.html and one known fail in https://sof-ci.01.org/sofpr/PR7828/build9916/devicetest/index.html . Proceeding with merge.

lyakh (commented):
after this PR the difference between volume_hifi3.c and volume_hifi4.c *essentially* is this:
```
--- module/volume/volume_hifi3.c	2023-06-22 09:46:10.864869868 +0200
+++ module/volume/volume_hifi4.c	2023-06-22 09:46:10.868869832 +0200
@@ -157,8 +157,7 @@
 						     + bsource->consumed);
 	ae_f32x2 *out = (ae_f32x2 *)audio_stream_wrap(sink, (char *)audio_stream_get_wptr(sink)
 						      + bsink->size);
-	const int channels_count = audio_stream_get_channels(sink);
-	int samples = channels_count * frames;
+	int samples = audio_stream_get_channels(sink) * frames;
 
 	bsource->consumed += VOL_S32_SAMPLES_TO_BYTES(samples);
 	bsink->size += VOL_S32_SAMPLES_TO_BYTES(samples);
@@ -449,6 +448,8 @@
 	const int channels_count = audio_stream_get_channels(sink);
 	int samples = channels_count * frames;
 
+	bsource->consumed += VOL_S16_SAMPLES_TO_BYTES(samples);
+	bsink->size += VOL_S16_SAMPLES_TO_BYTES(samples);
 	while (samples) {
 		m = audio_stream_samples_without_wrap_s16(source, in);
 		n = MIN(m, samples);
@@ -462,8 +463,6 @@
 		}
 		AE_SA64POS_FP(outu, out);
 		samples -= n;
-		bsource->consumed += VOL_S16_SAMPLES_TO_BYTES(n);
-		bsink->size += VOL_S16_SAMPLES_TO_BYTES(n);
 		in = audio_stream_wrap(source, in);
 		out = audio_stream_wrap(sink, out);
 	}
```
is that a real difference or something improved in one file and forgotten in the other? Additionally there are a couple of white space / empty line differences that already create random noise...

src/audio/module_adapter/module/volume/volume_hifi3_with_peakvol.c
andrula-song: Sorry i was on vacation that time, and this PR is merged. Then i will fix these in the next PR.

lyakh: @andrula-song no worries, doing that in a separate PR would be good!

PR #7772, opened 2023-06-08
[SKIP SOF-TEST] .github: add first LP64 compilation

marc-hb:
A ton of (valid!) warnings but this will make sure the code at least
compiles.

For more context see
- #7192

PR #7829, opened 2023-06-19
Use mimx93_evk_a55 board for SOF also

dbaluta:
We previously created a synthetic board named mimx93_evk_a55_sof but this is not a good direction to take. 

We need to use the default board mimx93_evk_a55 and adapt it to SOF using dts overlays and Kconfig fragments. 

After these patches we can build SOF firmware image for A55 using:

```
west build -p --board mimx93_evk_a55  modules/audio/sof/app
```


LaurentiuM1234 (approved):
Looks good. Assuming `mimx93_evk_a55_sof` will be removed from Zephyr don't forget to also update LP64 CI which uses it. Currently, things are a bit weird IMO with the mmu_regions.c using the regions from the overlay but these regions will be removed anyways when the support for native drivers is added. Out of curiosity: did you also try building SOF from Zephyr side using these changes?

iuliana-prodan (approved):

dbaluta:
> Out of curiosity: did you also try building SOF from Zephyr side using these changes?

I only build SOF from Zephyr side like this:

```
$ /* clone zephyr, run west init, west update */
# build zephyr.bin
$ west build -p --board mimx93_evk_a55  modules/audio/sof/app
```

LaurentiuM1234:
> Do you remember where was the `L64 CI` PR that added this support.

https://github.com/thesofproject/sof/pull/7772


> Also, thinking about this. I wonder if we shouldn't somehow used 32bits for SOF FW with a55. As an experiment using 64 bits is fine.
> 
> But I think for a more practical usecases when imx95 will arrive we should compile SOF firmware for a55 core with 32 bits support rather than 64 bits.
> 
> We don't really need 64 bits and we also save some space.

This will require some experimentation. Zephyr seems to compile arm64 binaries using the `lp64` flag atm so we'll have to watch out for this too.

kv2019i (approved):

kv2019i:
@dbaluta Feel free to merge.

PR #7833, opened 2023-06-20
Audio: Fix the aligned feature comments

andrula-song:
Fix the comments of frame_align and frame_align_shift to make them more clear and understandable.

### DIFF
--- a/src/include/sof/audio/audio_stream.h
+++ b/src/include/sof/audio/audio_stream.h
@@ -46,17 +46,24 @@ struct sof_audio_stream_params {
 	uint32_t rate;		/**< Number of data frames per second [Hz] */
 	uint16_t channels;	/**< Number of samples in each frame */
 
-	/** alignment limit of stream copy, this value indicates how many
-	 * integer frames which can meet both byte align and frame align
-	 * requirement. it should be set in component prepare or param functions
+	/**
+	 * align_frame_cnt indicates minimum number of frames that satisfies both byte
+	 * align and frame align requirements. E.g: Consider an algorithm that processes
+	 * in blocks of 3 frames configured to process 16-bit stereo using xtensa HiFi3
+	 * SIMD. Therefore with 16-bit stereo we have a frame size of 4 bytes, and
+	 * SIMD intrinsic requirement of 8 bytes(2 frames) for HiFi3 and an algorithim
+	 * requirement of 3 frames. Hence the common processing block size has to align
+	 * with frame(1), intrinsic(2) and algorithm (3) giving us an optimum processing
+	 * block size of 6 frames.
 	 */
-	uint16_t frame_align;
+	uint16_t align_frame_cnt;
 
 	/**
-	 * alignment limit of stream copy, alignment is the frame_align_shift-th
-	 * power of 2 bytes. it should be set in component prepare or param functions
+	 * the free/available bytes of sink/source right shift align_shift_idx, the result
+	 * multiplied by align_frame_cnt is the frame count free/available that can meet
+	 * the align requirement.
 	 */
-	uint16_t frame_align_shift;
+	uint16_t align_shift_idx;
 
 	bool overrun_permitted; /**< indicates whether overrun is permitted */
 	bool underrun_permitted; /**< indicates whether underrun is permitted */
@@ -403,9 +410,9 @@ static inline uint32_t audio_stream_frame_align_get(const uint32_t byte_align,
 }
 
 /**
- * Set frame_align_shift and frame_align of stream according to byte_align and
+ * Set align_shift_idx and align_frame_cnt of stream according to byte_align and
  * frame_align_req alignment requirement. Once the channel number,frame size
- * are determined，the frame_align and frame_align_shift are determined too.
+ * are determined，the align_frame_cnt and align_shift_idx are determined too.
  * these two feature will be used in audio_stream_get_avail_frames_aligned
  * to calculate the available frames. it should be called in component prepare
  * or param functions only once before stream copy. if someone forgets to call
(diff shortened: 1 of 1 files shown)


src/include/sof/audio/audio_stream.h
```diff
@@ -46,15 +46,23 @@ struct sof_audio_stream_params {
...
 	uint16_t channels;	/**< Number of samples in each frame */
 
-	/** alignment limit of stream copy, this value indicates how many
-	 * integer frames which can meet both byte align and frame align
-	 * requirement. it should be set in component prepare or param functions
+	/**
+	 * Alignment limit of stream copy, used by audio_stream_avail_frames_aligned.
+	 * We can set byte_align and frame_align in the init or prepare functions to
+	 * limit the processing data size is byte_align bytes aligned and frame_align
+	 * frames aligned. This feature indicates how many integer frames can meet both
```
btian1, line 53: feature or number or data or value is better?

btian1 (commented):
Normally, one variable does not need add so much comments, since you are adding, this means this variable is hard to understand directly,  how about add one example to show what exactly it means.

src/include/sof/audio/audio_stream.h
plbossart: that comment is not helpful @andrula-song. It's not clear if 'byte_align' is a boolean or a number of bytes to align on. Likewise 'frame_aligned' could be a boolean but you seem to use it as a counter. 

Consider renaming the structure members to describe exactly what they do in self-explanatory ways.

src/include/sof/audio/audio_stream.h
```diff
@@ -46,15 +46,23 @@ struct sof_audio_stream_params {
...
 	/**
-	 * alignment limit of stream copy, alignment is the frame_align_shift-th
-	 * power of 2 bytes. it should be set in component prepare or param functions
+	 * Alignment limit of stream copy, used by audio_stream_avail_frames_aligned.
+	 * We use right shifting frame_align_shift bits to replace divising framesize to
+	 * avoid runtime division. That means, if the frame size and the frame_align
+	 * we set both are power of 2, we will get the same results by shifting and
+	 * division. Otherwise we will get smaller available frame number than division,
+	 * and processing data size is the frame_align_shift-th power of 2 bytes aligned.
+	 * It should be set in component prepare or param functions
```
plbossart, line 65: same, you are making the explanations more verbose without explaining the concepts.

plbossart (changes requested):

src/include/sof/audio/audio_stream.h
lgirdwood: @andrula-song adding some example context is also helpful to explain. e.g. 
```
....using xtensa HiFi3 then its' best we align frame sizes of X bytes, e.g. a 16 bit stereo stream with Hifi3 would be Y frames and Z bytes....using Hifi4 then ......
```

src/include/sof/audio/audio_stream.h
```diff
@@ -46,17 +46,20 @@ struct sof_audio_stream_params {
...
 	uint16_t channels;	/**< Number of samples in each frame */
 
-	/** alignment limit of stream copy, this value indicates how many
-	 * integer frames which can meet both byte align and frame align
-	 * requirement. it should be set in component prepare or param functions
+	/**
+	 * align_frame_cnt indicates minimum frames can meet both bytes align and
+	 * frames align requirement. Eg: for a stereo 16-bit stream, the processing
+	 * data must be 8 bytes align as well as 3 frames align to meet the xtensa
+	 * intrinsic and algorithm requirement, then the align_frame_cnt is 6.
```
lgirdwood, line 53: Would 16bit stereo not be 4 bytes aligned instead of 8 ?

andrula-song: the multi-way load/store xtensa intrinsic needs to be 8-byte aligned.

btian1 (approved):

src/include/sof/audio/audio_stream.h
lgirdwood: > the multi-way load/store xtensa intrinsic needs to be 8-byte aligned.

ok, so the instrinsic requirement is 8 bytes (2 frames) and the algo requiremnet is 3 frames and hence the common factor processing block size is 6 frames.. You need to break it down as follows so it's easier to follow.

e.g. Consider an algorithm that processes in blocks of 3 frames configured to prcess 16bit stereo using xtensa HiFi3 SIMD. Therefore with 16 bit stereo we have a frame size of 4 bytes, and SIMD intrinsic requirment of 2 frames for HiFi X and a an algorithim requirment of 3 frames. Hence the common processing block size has to align with frame(1), intrinsic(2) and algorithm (3) giving us an optimum processing block size of 6 frames.

lgirdwood (approved):

lgirdwood:
@plbossart good now ?

singalsu (approved):

src/include/sof/audio/audio_stream.h
```diff
@@ -46,17 +46,24 @@ struct sof_audio_stream_params {
 	uint32_t rate;		/**< Number of data frames per second [Hz] */
 	uint16_t channels;	/**< Number of samples in each frame */
 
-	/** alignment limit of stream copy, this value indicates how many
-	 * integer frames which can meet both byte align and frame align
-	 * requirement. it should be set in component prepare or param functions
+	/**
+	 * align_frame_cnt indicates minimum frames can meet both bytes align and
```
lyakh, line 50: not a native English speaker, but maybe "minimum number of frames, that satisfies both byte and frame alignment requirements."

lgirdwood (approved):

lgirdwood:
SOFCI TEST

dbaluta (approved):

PR #7835, opened 2023-06-20
Fixes for multi-core support

serhiy-katsyuba-intel:
(1)
module_adapter: Fix for multi-core usage
ipc4_create_buffer() to get ibs/obs size uses comp_get_attribute(, COMP_ATTR_BASE_CONFIG, ) which returns module's struct processing_module::priv.cfg.base_cfg. Since buffer and module could be created on different cores, this fix makes sure module's struct processing_module is allocated in non-cached memory and can be read from any core.

(2)
Fix pipeline allocation for multi-core usage
pipeline_get_dai_comp_latency() iterates over all pipelines and reaches to struct pipeline::pipeline_id. If pipeline was created on core different from the core making pipeline_get_dai_comp_latency() call, reading pipeline_id returns wrong value. This fix ensures that struct pipeline can be read from any core.

WARNING: these fixes could potentially affect performance. I did not do any performance measurements with and without these modifications.

It might be possible to fix same problems without simply using non-cached memory but with some more complex solution. So the ideas are welcomed whether we can merge these fixes as is or we could come out with better solutions.

These are backports of fixes we already have on mtl-005-drop-stable.

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -55,7 +55,7 @@ struct comp_dev *module_adapter_new(const struct comp_driver *drv,
 	dev->ipc_config = *config;
 	dev->drv = drv;
 
-	mod = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*mod));
+	mod = rzalloc(SOF_MEM_ZONE_RUNTIME_SHARED, 0, SOF_MEM_CAPS_RAM, sizeof(*mod));
 	if (!mod) {
 		comp_err(dev, "module_adapter_new(), failed to allocate memory for module");
 		rfree(dev);
--- a/src/audio/pipeline/pipeline-graph.c
+++ b/src/audio/pipeline/pipeline-graph.c
@@ -122,7 +122,7 @@ struct pipeline *pipeline_new(uint32_t pipeline_id, uint32_t priority, uint32_t 
 	heap_trace_all(0);
 
 	/* allocate new pipeline */
-	p = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*p));
+	p = rzalloc(SOF_MEM_ZONE_RUNTIME_SHARED, 0, SOF_MEM_CAPS_RAM, sizeof(*p));
 	if (!p) {
 		pipe_cl_err("pipeline_new(): Out of Memory");
 		return NULL;


src/audio/module_adapter/module_adapter.c
```diff
@@ -55,7 +55,7 @@ struct comp_dev *module_adapter_new(const struct comp_driver *drv,
 	dev->ipc_config = *config;
 	dev->drv = drv;
 
-	mod = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*mod));
+	mod = rzalloc(SOF_MEM_ZONE_RUNTIME_SHARED, 0, SOF_MEM_CAPS_RAM, sizeof(*mod));
```
marcinszkudlinski, line 58: Please use kconfig option - non_coherent_pipeline_workaround or similar
this is not a target sollution, not a fix at all

marcinszkudlinski (changes requested):

src/audio/module_adapter/module_adapter.c
lyakh: how can a buffer and a module, attached to it, be allocated on different cores? Shouldn't both be done on the target core, where the pipeline will run? However, if this is for the case when parts of a stream are running on different cores, then this is still not supported and we still have bigger issues to fix for that. It is good to gradually prepare for that and fix parts that we come across, but I'd agree with @marcinszkudlinski that this needs to be considered carefully: ATM we don't have such use cases so for now it doesn't fix anything, but it probably will slow things down somewhat.

src/audio/pipeline/pipeline-graph.c
```diff
@@ -122,7 +122,7 @@ struct pipeline *pipeline_new(uint32_t pipeline_id, uint32_t priority, uint32_t
 	heap_trace_all(0);
 
 	/* allocate new pipeline */
-	p = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*p));
+	p = rzalloc(SOF_MEM_ZONE_RUNTIME_SHARED, 0, SOF_MEM_CAPS_RAM, sizeof(*p));
```
lyakh, line 125: mostly the same for this: no, let's not do this now. We'd need to make almost all pipeline / component / module allocations uncached now.

lyakh (changes requested):

marcinszkudlinski:
This workaround is also required to make DP modules working on secondary cores

src/audio/module_adapter/module_adapter.c
serhiy-katsyuba-intel: That ipc4_create_buffer() was called for buffer between two components on different pipelines and pipelines are on different cores. To get obs size ipc4_create_buffer() reaches component's base_cfg. The problem is that ipc4_create_buffer() was called from one core but that componet's base_cfg was allocated and written on another core.

Seems such problem only happens for buffers between pipelines on different cores, not sure.

So, yes, probably, this problem should be fixed by ensuring ipc4_create_buffer() is called from correct core.

lyakh: @serhiy-katsyuba-intel connecting components or pipelines on different cores is unsupported ATM

marcinszkudlinski: we need to fix it, cross core bind LL-DP is something we DO need and we need it soon

lyakh: @marcinszkudlinski sure, so somebody has to get to work on this and get it done ASAP

marcinszkudlinski:
Closing, properly fix in https://github.com/thesofproject/sof/pull/7951

PR #7837, opened 2023-06-20
LL scheduler: fix a slight bug and add a warning

lyakh:
Fix an off-by-one error in timer statistics calculation and add a warning for scheduling overruns

### DIFF
--- a/src/schedule/zephyr_domain.c
+++ b/src/schedule/zephyr_domain.c
@@ -72,10 +72,8 @@ static void zephyr_domain_thread_fn(void *p1, void *p2, void *p3)
 		dt->handler(dt->arg);
 		cycles1 = k_cycle_get_32();
 
-		if (cycles1 > cycles0)
-			diff = cycles1 - cycles0;
-		else
-			diff = UINT32_MAX - cycles0 + cycles1;
+		/* This handles wrapping correctly too */
+		diff = cycles1 - cycles0;
 
 		timer_fired = k_timer_status_get(&zephyr_domain->timer);
 		if (timer_fired > 1)


src/schedule/zephyr_domain.c
```diff
@@ -78,8 +78,11 @@ static void zephyr_domain_thread_fn(void *p1, void *p2, void *p3)
 			diff = UINT32_MAX - cycles0 + cycles1;
 
 		timer_fired = k_timer_status_get(&zephyr_domain->timer);
-		if (timer_fired > 1)
+		if (timer_fired > 1) {
 			overruns++;
+			tr_warn(&ll_tr, "LL timer overrun: took %u cycles, missed %u interrupts",
+				diff, timer_fired);
+		}
```
kv2019i, line 82: Not sure about this.  The system is under a load spike, so adding an immediate warning to the mix is likely to make the situation worse.

Especially now as the IPC handling can cause spikes to LL processing time (there is work ongoing to address this, e.g. what @aborisovich is doing), but until then, I'd rather report overruns via the periodic report.

src/schedule/zephyr_domain.c
```diff
@@ -72,10 +72,7 @@ static void zephyr_domain_thread_fn(void *p1, void *p2, void *p3)
 		dt->handler(dt->arg);
 		cycles1 = k_cycle_get_32();
 
-		if (cycles1 > cycles0)
-			diff = cycles1 - cycles0;
-		else
-			diff = UINT32_MAX - cycles0 + cycles1;
+		diff = cycles1 - cycles0;
```
kv2019i, line 75: I think this is less obvious to the reader, but alas this is simpler (plus there's the of-by-one as the module should be UINT32_MAX+1

kv2019i (changes requested):
Not sure about the first one...

kv2019i (approved):

dnikodem (approved):

kv2019i:
Unrelated PM fails in https://sof-ci.01.org/sofpr/PR7837/build9919/devicetest/index.html and https://sof-ci.01.org/sofpr/PR7837/build9918/devicetest/index.html . Proceeding with merge.

src/schedule/zephyr_domain.c
btian1: k_cycle_get_32 returns a value with max32, it is an accumulation of each read based on HW clock(max is 38400000), based on these knowledge, we know that after around 111 seconds, wrap will happens, below logs showed this:
[  111.847735] <inf> ll_schedule: zephyr_domain_thread_fn: ll timer 4294952856
[    0.000626] <inf> ll_schedule: zephyr_domain_thread_fn: ll timer 23858

if multiply 111.847735 or 0.000626 with 38400000, then you will get a similar data with later printed, actually they should same, just because the system debug for timing.

So previous code is correct, simply "-" will have issues when wrap happened.

Issue #7703, opened 2023-05-29
[BUG] Issues with using Zephyr CONFIG_LOG_MIPI_SYST_USE_CATALOG with SOF


kv2019i:
**Describe the bug**
Currently SOF Zephyr builds use "TEXT" logging output by default.

One alternative to text logging is the MIPI SYS-T Catalog logging backend which is found in Zephyr upstream:
https://github.com/zephyrproject-rtos/zephyr/issues/19841
https://docs.zephyrproject.org/latest/samples/subsys/logging/syst/README.html

There are currently some issues enabling this build with SOF:

- [x] Builds with Cadence xtensa tool chain fail with MIPI_SYST_USE_CATALOG (work/zephyr/lib/os/cbprintf_packaged.c:22:2: error: "CONFIG_CBPRINTF_PACKAGE_SUPPORT_TAGGED_ARGUMENT) -> https://github.com/zephyrproject-rtos/zephyr/pull/58499
- [x] No logs from SOF emitted to backend (is_in_log_strings_section() check always returns false for SOF prints, potential issue in linker scripts) -> https://github.com/zephyrproject-rtos/zephyr/issues/58476 , fix in https://github.com/zephyrproject-rtos/zephyr/pull/58719
- [x] Builds MIPI SYST USE_CAT and xt-clang will fail due to missing libc functionality (memset, strlen, ... ) -> https://github.com/thesofproject/sof/pull/7836 and https://github.com/zephyrproject-rtos/zephyr/pull/59287
- [x] FW panic if USE_CATALOG=y on MTL -> https://github.com/thesofproject/sof/issues/7839

**To Reproduce**
Add following to work/sof/app/prj.conf

```
CONFIG_LOG_BACKEND_ADSP_MTRACE=y
CONFIG_LOG_BACKEND_ADSP_MTRACE_OUTPUT_SYST=y
CONFIG_LOG_MIPI_SYST_ENABLE=y
CONFIG_LOG_MIPI_SYST_USE_CATALOG=n
```

And rebuild SOF. Capture log with "mtrace-reader.py" to a file "dsplog.bin" and post-process with upstream mipi-syst tools like:

systprint -c ../../work/build-tgl/zephyr/mipi_syst_collateral.xml dsplog.bin

**Reproduction Rate**
100%

**Expected behavior**
Sys-T Catalog logging can be enabled for all support logging backends.

**Impact**
Inability to use more effiecient standard dictionary format.

**Environment**
[...]

PR #7838, opened 2023-06-20
app: add logging_mipisystcat_overlay.conf overlay config

kv2019i:
Add an overlay config to build SOF with MIPI Sys-T Catalog logging enabled.

This adds MIPI Sys-T Catalog as an overlay option. There are still issues (tracked in https://github.com/thesofproject/sof/issues/7703 ) with using this output in SOF, and some of the SOF CI infra is not prepare to handle binary-encoded logs, so this option cannot yet be enabled across the board.

### DIFF
--- a/app/logging_mipisystcat_overlay.conf
+++ b/app/logging_mipisystcat_overlay.conf
@@ -1,0 +1,9 @@
+# Enable MIPI Sys-T Catalog FW logging
+# ref: https://docs.zephyrproject.org/latest/samples/subsys/logging/syst/README.html
+CONFIG_LOG_MIPI_SYST_ENABLE=y
+CONFIG_LOG_MIPI_SYST_USE_CATALOG=y
+
+# set as output type of all SOF backends
+CONFIG_LOG_BACKEND_ADSP_OUTPUT_SYST=y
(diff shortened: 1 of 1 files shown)


btian1:
Shall we expect with this PR and MIPI device, SOF logs can be routed out to host like mtrace?

kv2019i:
@btian1 wrote:
> Shall we expect with this PR and MIPI device, SOF logs can be routed out to host like mtrace?

This PR only changes en encoding of the logs. E.g. one can still use "mtrace-reader.py" to capture logs with SOF Linux driver. The only change is that one needs to use additional tools to parse the captured (binary) logs.

btian1 (approved):

btian1:
> 

please send out a mail(once #7703 resolved) for how to use tool to decode mipi logs, one more ask is: what's the benefit to add one more log format? windows required?

lgirdwood (approved):

marc-hb:
> please send out a mail(once https://github.com/thesofproject/sof/issues/7703 resolved)

There is a "subscribe" button.

> one more ask is: what's the benefit to add one more log format? 

https://docs.zephyrproject.org/latest/services/logging/index.html

Issue #7803, opened 2023-06-14
[ADL/RPL] [Bug] Invalid MCLK configuration with BT HFP configuration


kv2019i:
**Describe the bug**
Follow-up to https://github.com/thesofproject/sof/issues/7548

Description form @ujfalusi in the above bug
I think there is another angle to this issue which is brought to light by the TX FIFO drain simplifications.
The setup is
**SSP1** - **clock provider**, playback only to speaker, want MCLK as **19200000**
**SSP2** - **clock receiver**, playback/capture to BT, want MCLK as  **38400000**

We see this sequence in case of a failure:
SSP2 start (sets MCLK to 38400000)
```
ssp-dai      1.2   /drivers/intel/ssp/ssp.c:1111 INFO ssp_trigger() cmd 7
ssp-dai      1.2   /drivers/intel/ssp/ssp.c:900  INFO ssp_pre_start()
mn                 ./drivers/intel/ssp/mn.c:259  INFO mclk_rate 38400000, mclk_source_clock 1
mn                 ./drivers/intel/ssp/mn.c:220  INFO mclk_id 0 mdivr_val 1
mn                 ./drivers/intel/ssp/mn.c:330  INFO find_mn for freq 38400000 bclk 128000
mn                 ./drivers/intel/ssp/mn.c:628  INFO bclk_rate 128000, *out_scr_div 300, m 1, n 1
ssp-dai      1.2   /drivers/intel/ssp/ssp.c:234  INFO ssp_bclk_prepare_enable(): sscr0 = 0xc0c12b7f
ssp-dai      1.2   /drivers/intel/ssp/ssp.c:993  INFO ssp_early_start(): SSE set for SSP2
ssp-dai      1.2   /drivers/intel/ssp/ssp.c:1111 INFO ssp_trigger() cmd 1
ssp-dai      1.2   /drivers/intel/ssp/ssp.c:1008 INFO ssp_start()
```
SSP2 start (clocks have been configured for SSP2 already)
```
ssp-dai      1.2   /drivers/intel/ssp/ssp.c:1111 INFO ssp_trigger() cmd 7
ssp-dai      1.2   /drivers/intel/ssp/ssp.c:900  INFO ssp_pre_start()
ssp-dai      1.2   /drivers/intel/ssp/ssp.c:993  INFO ssp_early_start(): SSE set for SSP2
ssp-dai      1.2   /drivers/intel/ssp/ssp.c:1111 INFO ssp_trigger() cmd 1
[...]

PR #7840, opened 2023-06-21
intel: ssp: support zero mclk_rate in hw_config

brentlu:
Add a feature to set mclk_rate to zero so it won't interfere with other ssp port.

### DIFF
--- a/src/drivers/intel/ssp/ssp.c
+++ b/src/drivers/intel/ssp/ssp.c
@@ -119,6 +119,11 @@ static int ssp_mclk_prepare_enable(struct dai *dai)
 	struct sof_ipc_dai_config *config = &ssp->config;
 	int ret;
 
+	if (!config->ssp.mclk_rate) {
+		dai_info(dai, "ssp_mclk_prepare_enable(): SSP%d do not need mclk", dai->index);
+		return 0;
+	}
+
 	if (ssp->clk_active & SSP_CLK_MCLK_ACTIVE)
 		return 0;
 
@@ -136,6 +141,9 @@ static int ssp_mclk_prepare_enable(struct dai *dai)
 static void ssp_mclk_disable_unprepare(struct dai *dai)
 {
 	struct ssp_pdata *ssp = dai_get_drvdata(dai);
+
+	if (!ssp->config.ssp.mclk_rate)
+		return;
 
 	if (!(ssp->clk_active & SSP_CLK_MCLK_ACTIVE))
 		return;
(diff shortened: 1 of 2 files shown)


tools/topology/topology1/platform/intel/intel-generic-bt.m4
```diff
@@ -61,17 +60,17 @@ define(`data_names', DAI_DATA_NAMES(LIST(`     ', "ssp_data1", "ssp_data2", "ssp
...
 		`SSP_TDM(1, 16, 1, 1),'
 		`SSP_MULTI_CONFIG_DATA(ssp_data1, 16))',
-	`MULTI_SSP_CONFIG(hw_config2, eval(HW_CONFIG_ID + 1), DSP_A, SSP_CLOCK(mclk, BT_MCLK, codec_mclk_in),'
+	`MULTI_SSP_CONFIG(hw_config2, eval(HW_CONFIG_ID + 1), DSP_A, SSP_CLOCK(mclk, 0, codec_mclk_in),'
 		`SSP_CLOCK(bclk, 128000, codec_master, inverted),'
 		`SSP_CLOCK(fsync, 8000, codec_master),'
 		`SSP_TDM(1, 16, 1, 1),'
 		`SSP_MULTI_CONFIG_DATA(ssp_data2, 16))',
-	`MULTI_SSP_CONFIG(hw_config3, eval(HW_CONFIG_ID + 2), DSP_A, SSP_CLOCK(mclk, BT_MCLK, codec_mclk_in),'
+	`MULTI_SSP_CONFIG(hw_config3, eval(HW_CONFIG_ID + 2), DSP_A, SSP_CLOCK(mclk, 0, codec_mclk_in),'
```
plbossart, line 73: Humm, can you clarify what configurations conflict?

We've seen the same type of problems recently with @kv2019i and @CBala21, it's a generic problem that the MCLK should not be tied to specific SSPs.

I am not a big fan of this zero-rate solution.

brentlu: For this case: https://github.com/thesofproject/sof/issues/7803

For chrome books, we always choose same mclk rate for both headphone and speaker so there is no problem even most of our speaker amplifiers do not need mclk at all.

I am still struggling running bt offload on my chrome book but I believe it should work.

It we don't want to set the rate zero, we could set the mclk_id to a specific number. But it's weird that an invalid mclk id with a valid rate specified in topology.

kv2019i: @brentlu This is a bit problematic as ADSP is the clock consuming only in HFP BT profile. With A2DP BT offload, ADSP needs to provide BCLK, so MCLK is needed. So I think a better solution is something like this https://github.com/thesofproject/sof/pull/7741 that ignores MCLK programming if the ADSP is the clock consumer.

kv2019i (commented):
Commented on the BT parts, see inline

tools/topology/topology1/platform/intel/intel-generic-bt.m4
plbossart: The problem is that the MCLK_IN is a valid setting since TGL. In theory the DSP can take the MCLK from the external world and generate BCLK/FSYNC from that external clock.

brentlu: @plbossart yes MCLK_IN could be one source for M/N divider but I can't find the implementation in m/n driver. Did I miss something?

src/drivers/intel/ssp/ssp.c
```diff
@@ -410,21 +418,23 @@ static int ssp_set_config_tplg(struct dai *dai, struct ipc_config_dai *common_co
...
-	if (!config->ssp.mclk_rate ||
-	    config->ssp.mclk_rate > ssp_freq[MAX_SSP_FREQ_INDEX].freq) {
-		dai_err(dai, "ssp_set_config(): invalid MCLK = %d Hz (valid < %d)",
-			config->ssp.mclk_rate,
-			ssp_freq[MAX_SSP_FREQ_INDEX].freq);
-		ret = -EINVAL;
-		goto out;
-	}
+	/* not every codec reqire mclk */
+	if (config->ssp.mclk_rate) {
```
btian1, line 422: config->ssp.mclk_rate
if it is zero, previous logic will goto out, current does not, is this expected?

brentlu: Yes, in previous logic, you need to specify a mclk id and rate even the codec does not need it. Here we use the zero to specify this situation.

tools/topology/topology1/platform/intel/intel-generic-bt.m4
kv2019i: @brentlu  I stand corrected. It seems we can indeed omit MCLK output for BT offload (in all profiles). This needs to be double-checked on hardware, but would seem to be the ase. And indeed, this is not needed to generate BCLK.
So looks good for BT.

PR #7844, opened 2023-06-21
topology2: cavs-sdw : Update BT offload SSP for MTL chromebook

udaymb:
For MTL chromebook, SSP 1 is used for BT offload with cavs-sdw configuration


tools/topology/topology2/cavs-sdw.conf
```diff
@@ -42,6 +42,10 @@
 <output_audio_format.conf>
 <dmic-default.conf>
 <hdmi-default.conf>
+<platform/intel/ssp_aux_config.conf>
+<ssp.conf>
+<bt-default.conf>
+<bt-generic.conf>
```
ranj063, line 48: @udaymb can't do this. This will add BT to every sdw topology

udaymb: Updated the conf file accordingly

ranj063 (approved):

PR #7852, opened 2023-06-22
ipc: Fix typo in error checking

andyross:
Found a fuzz error I thought I'd fixed being reported again.  Turns out the fix had a typo and only fixed half the problem space.  Need to check source and sink, not source and source!


andyross:
@cujomalainey points out that I clearly have a stale rimage submodule (this is an old tree) that is polluting the PRs.  Will get that fixed up; I hadn't even noticed it.

andyross:
Hm... cleaning up and pushing does nothing to that submodule, even though it's from a separately cloned repo that can't possibly have it.  Will delete this and repush.

PR #7853, opened 2023-06-22
audio/selector: Don't assert on input-controllable memcpy_s failure

andyross:
The size argument to the memcpy_s() here is under the control of external data and can fail on garbage.  That needs to be a runtime failure, not an assertion.

### DIFF
--- a/src/audio/selector/selector.c
+++ b/src/audio/selector/selector.c
@@ -193,7 +193,11 @@ static struct comp_dev *selector_new(const struct comp_driver *drv,
 	comp_set_drvdata(dev, cd);
 
 	ret = memcpy_s(&cd->config, sizeof(cd->config), ipc_process->data, bs);
-	assert(!ret);
+	if (ret) {
+		rfree(cd);
+		rfree(dev);
+		return NULL;
+	}
 
 	dev->state = COMP_STATE_READY;
 	return dev;


andyross:
Another fuzz find

src/audio/selector/selector.c
```diff
@@ -193,7 +193,8 @@ static struct comp_dev *selector_new(const struct comp_driver *drv,
 	comp_set_drvdata(dev, cd);
 
 	ret = memcpy_s(&cd->config, sizeof(cd->config), ipc_process->data, bs);
-	assert(!ret);
+	if (ret)
+		return NULL;
```
tlissows, line 197: It looks like losing reference to allocated memory. Suggested change:
```
	if (ret) {
		rfree(cd);
		rfree(dev);
		return NULL;
	}
```

tlissows (changes requested):

andyross:
Rebase, include the suggestion as it seems safe and obviously correct.

FWIW: this is a fatal error situation, where the host has sent a garbage command.  I'd left the recover out as it's essentially impossible for me to test locally (this failure just doesn't come up often enough in fuzzing outside the cloud rig).  But I can't see why it would hurt.

lgirdwood (approved):

tlissows (approved):

lyakh (approved):

lgirdwood:
@wszypelt @lrudyX good to merge ? @andyross has no way to check CI.

wszypelt:
@lgirdwood after rerun in internal intel CI System everything went fine, so it's green on my part

PR #7854, opened 2023-06-22
ipc3: Don't trust size values in protocol input

andyross:
The handling for SOF_COMP_NONE was extracting a "size" field from a
struct sof_ipc_comp_process command and passing it down into the
module_adapter layer where it was being used as the range for a
memcpy_s().

But that struct is IPC input data from the host!  The size can
trivially be wrong and cause an overflow.  And needless to say,
fuzzing discovered the hole and blew it up.

Note that the previous behavior when LIBRARY||UNIT_TEST is left
unchanged (fuzz builds as "real" firmware and doesn't use those
features).  It turns out that this is still in production use, though
the details of why the code is different are unclear.  Drop a note in
to fix this later.

### DIFF
--- a/src/ipc/ipc3/helper.c
+++ b/src/ipc/ipc3/helper.c
@@ -277,10 +277,14 @@ static void comp_specific_builder(struct sof_ipc_comp *comp,
 	case SOF_COMP_MODULE_ADAPTER:
 	case SOF_COMP_NONE:
 		config->process.type = proc->type;
-		config->process.size = proc->size;
 #if CONFIG_LIBRARY || UNIT_TEST
+		/* FIXME: protocol code paths shouldn't be different for testing */
+		config->process.size = proc->size;
 		config->process.data = proc->data + comp->ext_data_length;
 #else
+		/* Clamp size, don't trust the protocol-supplied value! */
+		config->process.size = MIN(proc->size,
+					   SOF_IPC_MSG_MAX_SIZE - sizeof(*proc));
 		config->process.data = proc->data;
 #endif
 		break;


src/ipc/ipc3/helper.c
```diff
@@ -278,11 +278,7 @@ static void comp_specific_builder(struct sof_ipc_comp *comp,
 	case SOF_COMP_NONE:
 		config->process.type = proc->type;
 		config->process.size = proc->size;
-#if CONFIG_LIBRARY || UNIT_TEST
-		config->process.data = proc->data + comp->ext_data_length;
-#else
-		config->process.data = proc->data;
-#endif
+		config->process.data = NULL;
```
lyakh, line 281: no, `proc->data` is an array, so this is perfectly valid

lyakh (changes requested):

src/ipc/ipc3/helper.c
btian1, line 281: it is not only for NONE, it is for ten component cases, check line :268-278.
Are you sure all above cases all does not need this?

lgirdwood: Why do we handle this differently on host ? Expectation is that ABI should be the same, I ssuspect the testbench is doing something different @singalsu @andrula-song ?

lgirdwood (commented):
I suspect UT or testbench is doing something different to runtime FW here and should be fixed to align (alongside this).

andyross:
So what's the consensus here?  As @lyakh points out, my explanation is wrong.  It's not the buffer pointer that is at fault, it's that the resulting transfer is unbounded.  Nothing in the current scheme passes a valid size for the buffer, so we don't know how much we can copy.

I can try to fix this (by extracting a size from somewhere up higher in the IPC layer I guess), but it seems like there's confusion about whether or not the unit test stuff needs it?

My assumption had been that this code path would never be hit in practice and was just a fuzzable edge.  What's the actual circumstance where you'd get data passed in any testing rig for a COMP_NONE component?

Broadly: what happens if we just turn the COMP_NONE handling into a runtime error instead?  That would be even less code.

lgirdwood:
@andrula-song what happens if we make the host handling the same as FW handling here ? Does testbench still work as expected ? We need to align handling here.

andrula-song:
> @andrula-song what happens if we make the host handling the same as FW handling here ? Does testbench still work as expected ? We need to align handling here.

if we make the host handling the same as FW handling here would lead to those sof_ipc_comp_process components(eq,mux/demux) unit test failed. Checked the code, the comp->ext_data_length is UUID_SIZE. It makes sense if there is multiple components in test topology.

andrula-song:
hi @lgirdwood 
yes, it will break the UT for eq-iir, eq-fir,mux and demux. I tried to remove the 'ext_data_length' segment yesterday, but will get error " get_drv(): driver not found,". Seems needs more effort to remove this segment.

lgirdwood:
@andyross we need to fix the length issue first. Seppo is best to look at this when he's back. Not sure why testbench is doing something else.

andyross:
OK.  This version leaves the UNIT_TEST/LIBRARY path unchanged (with a FIXME note).  It will have the same bug on bad input, but that isn't involved in either production firmware nor fuzzing, so nothing is getting worse at least.

Also, instead of nulling the pointer out with the bad explanation, this does a slightly more sophisticated trick where it clamps the protocol-provided garbage size value (which was the real root cause) to the known size of the IPC command buffer.  I still argue we should remove this whole switch case and replace it with an error, but I guess that will need to wait for the test updates?

kv2019i (approved):

lyakh (commented):
This is now handling the same fuzzing detected problem as #7830 and as I commented there in https://github.com/thesofproject/sof/pull/7830#issuecomment-1632068982 I think this function should return an error in case of invalid input data

src/ipc/ipc3/helper.c
```diff
@@ -277,10 +277,14 @@ static void comp_specific_builder(struct sof_ipc_comp *comp,
...
 		config->process.type = proc->type;
-		config->process.size = proc->size;
 #if CONFIG_LIBRARY || UNIT_TEST
+		/* FIXME: protocol code paths shouldn't be different for testing */
+		config->process.size = proc->size;
 		config->process.data = proc->data + comp->ext_data_length;
 #else
+		/* Clamp size, don't trust the protocol-supplied value! */
+		config->process.size = MIN(proc->size,
+					   SOF_IPC_MSG_MAX_SIZE - sizeof(*proc));
```
lgirdwood, line 287: why subtract the sizeof(*proc) here, wouldn't we just want to clamp to max message size. TBH, we should just reject anything bigger than ```SOF_IPC_MSG_MAX_SIZE``` with an error.

lyakh: I think it's because `proc->size` is the size of the `.data[]` array at the end of `struct sof_ipc_comp_process` so the total size should fit, i.e. `sizeof(*proc) + proc->size <= SOF_IPC_MSG_MAX_SIZE`. But yes, making too large `.size` values an error would be good!

lgirdwood: Thanks for the reminder @lyakh - maybe data->size would be a better name.

dbaluta (approved):

src/ipc/ipc3/helper.c
singalsu: I think the data advance to past possible UUID can be done in every build, not just UT and testbench.

singalsu:
I removed my reject to not block this while I'm one week OOO. But I don't agree with this (no error, cut end of too long IPC) fix approach.

lgirdwood:
> I removed my reject to not block this while I'm one week OOO. But I don't agree with this (no error, cut end of too long IPC) fix approach.

ack, it seems the best course is to return an error up the stack and add some error message in the logs.

btian1 (approved):

andyross:
Sorry, this got left stale.  I'm going to close this, as AFAICT the error it addresses isn't repeating post- #7830   The side question about why unit testing is using a variant protocol can be addressed elsewhere.

PR #7855, opened 2023-06-22
ipc: Fix typo in error checking

andyross:
Found a fuzz error I thought I'd fixed being reported again.  Turns out the fix had a typo and only fixed half the problem space.  Need to check source and sink, not source and source!


andyross:
Unbelievable.  The rimage delta is still there!   I'm guessing it's because GH cached the commit ID?  Closing again.  Will resubmit a slightly changed commit message I guess....

PR #7861, opened 2023-06-26
[DNM][Do Not Review] loadable module PoC

lyakh:
this is a continuation of #7018 and https://github.com/aiChaoSONG/sof/commits/sma_loadable to demonstrate building loadable modules, taking smart-amp-test as an example


lmdk/libraries/module_example/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+
+cmake_minimum_required(VERSION 3.20)
```
marc-hb, line 2: I see nothing specific in this short file that requires 3.20, is there?

We already have a (too many?) `cmake_minimum_required()` elsewhere, I don't see why we need another one here.

gkbldcig:

PR #7867, opened 2023-06-27
[DNM][Do Not Review] loadable module PoC: mtl-005 fork

lyakh:
This is a version for the mtl-005-drop-stable branch. Both the smart-amp-test module and a firmware image can be built from this state.


lgirdwood:
@abonislawski can you take a quick review and add any other folks who need to check. Thanks.

PR #7868, opened 2023-06-27
Audio: Fix IPC4 sink component buffer parameters set for S24_LE

singalsu:
The frame_fmt from
audio_stream_fmt_conversion(32, 24, &frame_fmt, &valid_fmt) is SOF_IPC_FRAME_S32_LE (2). valid_fmt is SOF_IPC_FRAME_S24_4LE (1).

Set of member frame_fmt from struct sof_ipc_stream_params should be done with valid_fmt instead to get the components to handle the S24_LE stream correctly.

### DIFF
--- a/src/audio/eq_fir/eq_fir.c
+++ b/src/audio/eq_fir/eq_fir.c
@@ -190,7 +190,7 @@ static int eq_fir_params(struct processing_module *mod)
 				    &frame_fmt, &valid_fmt,
 				    mod->priv.cfg.base_cfg.audio_fmt.s_type);
 
-	comp_params.frame_fmt = frame_fmt;
+	comp_params.frame_fmt = valid_fmt;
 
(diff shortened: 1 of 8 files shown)


singalsu:
SOFCI TEST

ranj063 (approved):

cujomalainey (approved):
Thanks for following up on my request and testing this

RanderWang:
> Seems valid to me. In current topologies, we use S32_LE for all host pipes, so this has not been hit, but needs to be fixed.

we could use S24_LE and used it a few months ago but it is still valid to use valid_fmt done in this PR.
S24_LE pipeline will have performance advantage and it is used by ref FW.
This PR is valid since we convert different groups of valid sample size & container size in gateway copier to IPC3 format (S16_LE, S24_LE and S32_LE), so internal module will only use IPC3 format

PR #7871, opened 2023-06-28
Audio: Crossover: Convert to module adapter

singalsu:
This patch contains basic changes to run crossover component as module adapter client in IPC3 system. Additional changes are needed for IPC4.

### DIFF
--- a/src/audio/crossover/crossover.c
+++ b/src/audio/crossover/crossover.c
@@ -4,6 +4,7 @@
 //
 // Author: Sebastiano Carlucci <scarlucci@google.com>
 
+#include <sof/audio/module_adapter/module/generic.h>
 #include <sof/audio/buffer.h>
 #include <sof/audio/component.h>
 #include <sof/audio/data_blob.h>
@@ -36,7 +37,7 @@
 #include <stdint.h>
 #include <limits.h>
 
-static const struct comp_driver comp_crossover;
+LOG_MODULE_REGISTER(crossover, CONFIG_SOF_LOG_LEVEL);
 
 /* 948c9ad1-806a-4131-ad6c-b2bda9e35a9f */
 DECLARE_SOF_RT_UUID("crossover", crossover_uuid, 0x948c9ad1, 0x806a, 0x4131,
@@ -96,8 +97,8 @@ static inline void crossover_reset_state(struct comp_data *cd)
  * \return the position at which pipe_id is found in config->assign_sink.
  *	   -EINVAL if not found.
  */
-static int crossover_get_stream_index(struct sof_crossover_config *config,
-					  uint32_t pipe_id)
+static int crossover_get_stream_index(struct processing_module *mod,
+				      struct sof_crossover_config *config, uint32_t pipe_id)
 {
 	int i;
 	uint32_t *assign_sink = config->assign_sink;
@@ -106,8 +107,8 @@ static int crossover_get_stream_index(struct sof_crossover_config *config,
 		if (assign_sink[i] == pipe_id)
 			return i;
 
-	comp_cl_err(&comp_crossover, "crossover_get_stream_index() error: couldn't find any assignment for connected pipeline %u",
-		    pipe_id);
+	comp_err(mod->dev, "crossover_get_stream_index() error: couldn't find any assignment for connected pipeline %u",
+		 pipe_id);
 
 	return -EINVAL;
 }
@@ -123,56 +124,65 @@ static int crossover_get_stream_index(struct sof_crossover_config *config,
  * \return number of sinks assigned. This number should be equal to
  *	   config->num_sinks if no errors were found.
  */
-static int crossover_assign_sinks(struct comp_dev *dev,
-				  struct sof_crossover_config *config,
-				  struct comp_buffer **sinks)
+static int crossover_assign_sinks(struct processing_module *mod,
+				  struct comp_buffer **sinks,
+				  bool *enabled)
 {
+	struct comp_data *cd = module_get_private_data(mod);
+	struct sof_crossover_config *config = cd->config;
+	struct comp_dev *dev = mod->dev;
 	struct comp_buffer *sink;
 	struct comp_buffer __sparse_cache *sink_c;
 	struct list_item *sink_list;
 	int num_sinks = 0;
 	int i;
+	int j = 0;
 
-	/* Align sink streams with their respective configurations */
 	list_for_item(sink_list, &dev->bsink_list) {
-		unsigned int pipeline_id, state;
+		unsigned int sink_id, state;
 
 		sink = container_of(sink_list, struct comp_buffer, source_list);
 		sink_c = buffer_acquire(sink);
-
-		pipeline_id = sink_c->pipeline_id;
+#if CONFIG_IPC_MAJOR_4
+		sink_id = cd->output_pin_index[j];
+#else
+		sink_id = sink_c->pipeline_id;
+#endif
 		state = sink_c->sink->state;
 		buffer_release(sink_c);
-
-		if (state != dev->state)
+		if (state != dev->state) {
+			j++;
 			continue;
+		}
 
 		/* If no config is set, then assign the sinks in order */
 		if (!config) {
 			sinks[num_sinks++] = sink;
+			enabled[j++] = true;
 			continue;
 		}
 
-		i = crossover_get_stream_index(config, pipeline_id);
+		i = crossover_get_stream_index(mod, config, sink_id);
 
 		/* If this sink buffer is not assigned
 		 * in the configuration.
 		 */
 		if (i < 0) {
 			comp_err(dev,
-				 "crossover_acquire_sinks(), could not find sink %d in config",
+				 "crossover_assign_sinks(), could not find sink %d in config",
 				 pipeline_id);
 			break;
 		}
 
 		if (sinks[i]) {
 			comp_err(dev,
-				 "crossover_acquire_sinks(), multiple sinks from pipeline %d are assigned",
+				 "crossover_assign_sinks(), multiple sinks from pipeline %d are assigned",
 				 pipeline_id);
 			break;
 		}
 
 		sinks[i] = sink;
+		enabled[j++] = true;
 		num_sinks++;
 	}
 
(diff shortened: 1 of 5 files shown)


src/audio/crossover/crossover.c
```diff
@@ -123,56 +124,61 @@ static int crossover_get_stream_index(struct sof_crossover_config *config,
...
 		sink = container_of(sink_list, struct comp_buffer, source_list);
 		sink_c = buffer_acquire(sink);
-
 		pipeline_id = sink_c->pipeline_id;
 		state = sink_c->sink->state;
 		buffer_release(sink_c);
-
-		if (state != dev->state)
+		if (state != dev->state) {
+			j++;
```
cujomalainey, line 150: nice catch

src/audio/crossover/crossover.c
```diff
@@ -123,56 +124,61 @@ static int crossover_get_stream_index(struct sof_crossover_config *config,
...
 		if (sinks[i]) {
 			comp_err(dev,
-				 "crossover_acquire_sinks(), multiple sinks from pipeline %d are assigned",
+				 "crossover_assign_sinks(), multiple sinks from pipeline %d are assigned",
 				 pipeline_id);
 			break;
 		}
 
 		sinks[i] = sink;
+		enabled[j++] = true;
```
cujomalainey, line 181: these changes seem more like a bugfix rather than a IPC4 change, can you split into a different commit?

src/audio/crossover/crossover.c
```diff
@@ -469,167 +461,85 @@ static int crossover_validate_config(struct comp_dev *dev,
...
+	struct comp_dev *dev = mod->dev;
+	struct audio_stream __sparse_cache *source = input_buffers[0].data;
 	uint32_t num_sinks;
 	uint32_t num_assigned_sinks = 0;
-	uint32_t frames = UINT_MAX;
-	uint32_t source_bytes, avail;
-	uint32_t sinks_bytes[SOF_CROSSOVER_MAX_STREAMS] = { 0 };
-
-	comp_dbg(dev, "crossover_copy()");
+	uint32_t frames = input_buffers[0].size;
```
cujomalainey, line 519: this assumes all buffers have the same space, won't we overrun?

src/include/sof/audio/component.h
```diff
@@ -715,6 +714,7 @@ void sys_comp_kpb_init(void);
 void sys_comp_multiband_drc_init(void);
 void sys_comp_selector_init(void);
 
+void sys_comp_module_crossover_interface_init(void);
```
btian1, line 717: seems this is not must, check up_down_mixer.c, it can be initialized with:
SOF_MODULE_INIT(up_down_mixer, sys_comp_module_up_down_mixer_interface_init);

src/audio/crossover/crossover.c
```diff
@@ -642,215 +562,180 @@ static int crossover_copy(struct comp_dev *dev)
...
-		if (!sinks[i])
-			continue;
-		/*
-		 * WARNING: if a different thread happens to lock the same
-		 * buffers in different order, they can deadlock
-		 */
-		sinks_c[i] = buffer_acquire(sinks[i]);
-		avail = audio_stream_avail_frames(&source_c->stream,
-						  &sinks_c[i]->stream);
-		frames = MIN(frames, avail);
```
johnylin76, line 656: This logic seems to be removed by the commit, is it intentional?
It looks important that determines the processed frames with regard to sink capacities in case of overrun.

src/audio/crossover/crossover.c
```diff
@@ -642,215 +562,180 @@ static int crossover_copy(struct comp_dev *dev)
...
+		crossover_set_stream_params(sink, mod);
+	}
+#endif
 
 	/* Validate frame format and buffer size of sinks */
 	list_for_item(sink_list, &dev->bsink_list) {
 		sink = container_of(sink_list, struct comp_buffer, source_list);
 		sink_c = buffer_acquire(sink);
 
+		/* Note: Printing pipeline IDs for debug, these are with IPC4 zeros, probably due
```
johnylin76, line 646: Does the bug you mentioned still exist under IPC4? If yes, is it worked-around by the commit, or is it non-critical?

src/include/sof/audio/component.h
singalsu: Statically linked testbench needs this header, it's not using the advanced component introduce and initialize system.

src/audio/crossover/crossover.c
singalsu: I can't see the bug, is there one? The enabled vector is for correctly updating mod->output_buffers[i].size. I'm not sure why the multiple buffers are provided by module API like they are now with possibly different state than component. It would be nice if the slow list search could be avoided totally and only the buffers vector from module api could be used.

src/audio/crossover/crossover.c
singalsu: It should not happen. In this case case the function module_single_source_setup() checks the frames avail/free from all source and sink combinations: https://github.com/thesofproject/sof/blob/a38458dd2276fa24e37f5e60ef0a5613b03cbe00/src/audio/module_adapter/module_adapter.c#L778

src/audio/crossover/crossover.c
singalsu: Same as Curtis asked? The module adapter does this for clients. Though there's also the sinks[i] NULL case if states are not equal. I have not been able to create such test situation, could a NULL sink impact the avail/free. I'm not sure.

src/audio/crossover/crossover.c
singalsu: I'm now after vacation starting to study this again -- I hope I'll gain understanding of IPC4 to be able to address this.

src/audio/crossover/crossover.c
cujomalainey: Definitely not intuitive, mind leaving a comment explaining that?

src/audio/crossover/crossover.c
```diff
@@ -106,12 +107,24 @@ static int crossover_get_stream_index(struct sof_crossover_config *config,
...
 	return -EINVAL;
 }
 
+static uint32_t crossover_get_sink_pipeline_id(struct comp_buffer __sparse_cache *sink_c)
+{
+#if CONFIG_IPC_MAJOR_4
+	// TODO: IPC4 renumbers pipeline_id:s due to pipelines count limitation, so can't use it.
+	// Use instead output pin index, cd->ipc4_cfg->output_pin.pin_index. The crossover
+	// topology blobs for IPC4 are now built with those indices.
+	return sink_c->sink->pipeline->pipeline_id;
```
singalsu, line 122: This change is stll needed. The crossover configuration blobs for IPC4 need to be populated with pin indexes 0, 1, ... instead of pipeline_id:s.

singalsu:
Crossover now runs with this PR correctly with IPC4 and IPC3, so changing this to non-draft. I'm able to run 2-way configuration with UPX board in nocodec mode. I play one stream and capture back from SSP loopback paths the band filtered filters versions those appear to be OK.

The other patches those belong to this work are:
- https://github.com/thesofproject/sof/pull/7883
- https://github.com/thesofproject/sof/pull/7884
- https://github.com/thesofproject/rimage/pull/170

src/audio/crossover/crossover.c
kv2019i: @singalsu can you add a comment?

src/include/sof/audio/component.h
kv2019i: @btian1 I think the SOF_MODULE_INIT is in crossover.c -- looks correct o me.

kv2019i (commented):
Thanks @singalsu ! I think there is one req-for-an-inline-comment outstanding, but otherwise looks good to go.

src/audio/crossover/crossover.c
singalsu: Thanks, yep, the comment is now added!

kv2019i (approved):

cujomalainey (approved):

src/audio/crossover/crossover.c
```diff
@@ -642,215 +665,179 @@ static int crossover_copy(struct comp_dev *dev)
...
-		/*
-		 * WARNING: if a different thread happens to lock the same
-		 * buffers in different order, they can deadlock
-		 */
-		sinks_c[i] = buffer_acquire(sinks[i]);
-		avail = audio_stream_avail_frames(&source_c->stream,
-						  &sinks_c[i]->stream);
-		frames = MIN(frames, avail);
+	/* Process crossover */
+	if (frames) {
```
lyakh, line 669: nitpick: an easier form would be
```
if (!frames)
    return -ENODATA;
```

src/audio/crossover/crossover.c
```diff
@@ -642,215 +665,179 @@ static int crossover_copy(struct comp_dev *dev)
...
-								      dev->frames);
-			if (audio_stream_get_size(&sink_c->stream) < sink_period_bytes) {
-				comp_err(dev,
-					 "crossover_prepare(), sink %d buffer size %d is insufficient",
-					 sink_c->pipeline_id, audio_stream_get_size(&sink_c->stream));
-				ret = -ENOMEM;
-			}
 		}
 
+		audio_stream_init_alignment_constants(1, 1, &sink_c->stream);
```
lyakh, line 760: this needs to be done even if the formats didn't match and we're going to error out?

src/audio/crossover/crossover.c
singalsu: Yep, done

src/audio/crossover/crossover.c
singalsu: Yep, ordered this way for more compact appear with my style but done now.

src/audio/crossover/crossover.c
```diff
@@ -642,215 +666,179 @@ static int crossover_copy(struct comp_dev *dev)
...
-			}
 		}
 
 		buffer_release(sink_c);
-
 		if (ret < 0)
-			goto out;
+			return ret;
+
+		audio_stream_init_alignment_constants(1, 1, &sink_c->stream);
```
lyakh, line 764: no, this cannot be done here - cannot use `sink_c` after releasing

btian1 (approved):

lyakh (approved):

lgirdwood:
@wszypelt @lrudyX good to merge ? IIUC, crossover not tested by internal CI today.

lrudyX:
> @wszypelt @lrudyX good to merge ? IIUC, crossover not tested by internal CI today.

It is waiting in queue. I have moved it to the top. We had some problems with DUTs.

lrudyX:
@lgirdwood Internal CI result is pass :)

PR #7872, opened 2023-06-28
Topology2: Remove usage of audio_format class - Part 3

ranj063:


lgirdwood (approved):
@jsarha pls review

jsarha (approved):
The changes appear symmetric so AFAICT they should be correct. The topology concept  
 for channel map and channel config is quite new to me, but it feels a bit odd to use 3.1 channel-map for 4ch dmic... but that is not a new thing in this PR.

PR #7874, opened 2023-06-28
Comp no wname in kcontrol name

jsarha:
Ok, this PR has evolved enough so that the branch name does not anymore match the content, so I guess its time to turn the page and create a new one. The predecessor can be found as draft here: https://github.com/thesofproject/sof/pull/7227

### DIFF
--- a/tools/topology/topology2/include/common/tokens.conf
+++ b/tools/topology/topology2/include/common/tokens.conf
@@ -24,6 +24,7 @@ Object.Base.VendorToken {
 		output_pin_binding_name	414
 		num_input_audio_formats	415
 		num_output_audio_formats 416
+		no_wname_in_kcontrol_name 417
 	}
 
 	"2" {
--- a/tools/topology/topology2/include/components/widget-common.conf
+++ b/tools/topology/topology2/include/components/widget-common.conf
@@ -104,3 +104,15 @@ DefineAttribute.uuid {
 	# Token set reference name and type
 	token_ref	"comp.uuid"
 }
+
+DefineAttribute."no_wname_in_kcontrol_name" {
+	type	"string"
+	# Token set reference name
+	token_ref	"comp.bool"
+	constraints {
+		!valid_values [
+			"true"
+			"false"
+		]
+	}
+}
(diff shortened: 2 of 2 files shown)


ranj063:
@jsarha I was discussing the volume test in sof-ci with @marc-hb yesterday and we realized that the IPC4 volume tests do nothing today because there are no mixer names with "PGA" in the name. Now, if we remove the widget name in the mixer names altogether, the tests will also no nothing for IPC3 as well. So we first need to think about what the volume tests do before we move ahead with this change

tools/topology/topology2/include/components/eqiir.conf
```diff
@@ -61,4 +61,5 @@ Class.Widget."eqiir" {
 	no_pm			"true"
 	num_input_pins		1
 	num_output_pins		1
+	no_wname_in_kcontrol_name	"true"
```
ranj063, line 64: why not set this to true in widget-common.conf?

marc-hb:
Context:
- https://github.com/thesofproject/sof-test/issues/1068
- https://github.com/thesofproject/sof-test/issues/1069

tools/topology/topology2/include/components/eqiir.conf
lgirdwood: ack, seems that base class is best place.

jsarha:
Moved the default value change to widget-common.conf.

jsarha:
> @jsarha I was discussing the volume test in sof-ci with @marc-hb yesterday and we realized that the IPC4 volume tests do nothing today because there are no mixer names with "PGA" in the name. Now, if we remove the widget name in the mixer names altogether, the tests will also no nothing for IPC3 as well. So we first need to think about what the volume tests do before we move ahead with this change

The simplest way would be using amixer numid notation to identify the mixers, but maybe a regex pattern to match a particular mixer would be easier to follow.

jsarha:
> Moved the default value change to widget-common.conf.

tools/topology/topology2/include/components/eqiir.conf
jsarha: done

lgirdwood:
@ranj063 good for you ?

ranj063 (approved):

lgirdwood (approved):

PR #7878, opened 2023-06-29
volume: fix log message to match caller

cujomalainey:
Likely artifact from module adapter conversion

### DIFF
--- a/src/audio/module_adapter/module/volume/volume.c
+++ b/src/audio/module_adapter/module/volume/volume.c
@@ -1432,7 +1432,7 @@ static struct comp_dev *volume_new(const struct comp_driver *drv,
 
 	mod = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*mod));
 	if (!mod) {
-		comp_err(dev, "module_adapter_new(), failed to allocate memory for module");
+		comp_err(dev, "volume_new(), failed to allocate memory for module");
 		goto fail;
 	}
 


marc-hb (approved):

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -1432,7 +1432,7 @@ static struct comp_dev *volume_new(const struct comp_driver *drv,
 
 	mod = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*mod));
 	if (!mod) {
-		comp_err(dev, "module_adapter_new(), failed to allocate memory for module");
+		comp_err(dev, "volume_new(), failed to allocate memory for module");
```
btian1, line 1435: May need futher change to : volume_new(), failed to allocate memory for volume

btian1 (approved):

kv2019i (approved):
Thanks @cujomalainey !

In Zephyr builds. we could actually drop these as we have CONFIG_LOG_FUNC_NAME_PREFIX_ in Zephyr, but for now as we support both XTOS and Zephyr in SOF, we need to keep the function names in log messages in common code.

src/audio/module_adapter/module/volume/volume.c
cujomalainey: I'm inclined to leave it as is since its allocating for `processing_module` which means its technically correct

cujomalainey:
Unrelated HW failures

PR #7879, opened 2023-06-30
perf_cnt: simplify the difference calculation for cycles

btian1:
Previously, due to cycle get max is UINT32_MAX, so have to check and wrap the code, however, unsigned int can wrap automatically with minus, remove the check and make perf calculation more simple.

### DIFF
--- a/xtos/include/sof/lib/perf_cnt.h
+++ b/xtos/include/sof/lib/perf_cnt.h
@@ -127,16 +127,8 @@ struct perf_cnt_data {
 			(uint32_t)sof_cycle_get_64();				\
 		uint32_t cpu_ts =						\
 			(uint32_t)perf_cnt_get_cpu_ts();			\
-		if (plat_ts > (pcd)->plat_ts)					\
-			(pcd)->plat_delta_last = plat_ts - (pcd)->plat_ts;	\
-		else                                             \
-			(pcd)->plat_delta_last = UINT32_MAX - (pcd)->plat_ts   \
-									+ plat_ts; \
-		if (cpu_ts > (pcd)->cpu_ts)			\
-			(pcd)->cpu_delta_last = cpu_ts - (pcd)->cpu_ts; \
-		else								\
-			(pcd)->cpu_delta_last = UINT32_MAX - (pcd)->cpu_ts	\
-									+ cpu_ts;\
+		(pcd)->plat_delta_last = plat_ts - (pcd)->plat_ts;		\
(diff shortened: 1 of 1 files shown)


xtos/include/sof/lib/perf_cnt.h
```diff
@@ -123,22 +120,8 @@ struct perf_cnt_data {
...
-									+ plat_ts; \
-		if (cpu_ts > (pcd)->cpu_ts)			\
-			(pcd)->cpu_delta_last = cpu_ts - (pcd)->cpu_ts; \
-		else								\
-			(pcd)->cpu_delta_last = UINT32_MAX - (pcd)->cpu_ts	\
-									+ cpu_ts;\
-		if ((pcd)->plat_delta_last > (pcd)->plat_delta_peak)		\
-			(pcd)->plat_delta_peak = (pcd)->plat_delta_last;	\
+		uint32_t cpu_ts = k_cycle_get_32();				\
+		(pcd)->cpu_delta_last = cpu_ts - (pcd)->cpu_ts;			\
```
kv2019i, line 124: Hmm, this is now doing a lot more than simplifying the wrap-around calculation. You change the clock source for both cpu_ts and plat_ts and I'm not sure this is correct. perf_cnt_get_cpu_ts() has been defined just because sof_cycle_get_64() is not the correct function on all platforms. Moving this to k_cycle_get_32() might be ok, but requires a wider review.

Could you make a PR that only simplifies the delta calculations but does not modify the clock acquire calls?

kv2019i (commented):
Not sure about the cpu/plat_ts clock source changes. Might be ok, but I'd prefer to have this change in separate PR/commit.

xtos/include/sof/lib/perf_cnt.h
btian1: ok, however, we don't use plat_ts for performance calculation.
may be two commit is better, let me submit with a simple one.

lyakh: @kv2019i I think the reason for removing the use of the platform clock source is that it isn't used under Zephyr, under Zephyr the same clock is used for both

lgirdwood: This is also xtos related header. We need both clocks availabe via Zephyr APIs.

btian1: yes, I only changed the difference part, and keep both clock in perf_cnt header.

lgirdwood (approved):

ranj063 (approved):

PR #7884, opened 2023-06-30
Tools: Tune: Update crossover configurations export for tplg2

singalsu:
The script example_crossover.m is converted to export 2, 3, and 4 way crossover configurations for tplg1 and tplg2.

The crossover parameters like number of sinks, pipeline ids of sinks, sample rate, and band limits are added to generate blob filenames.

The changes include fixes for running the script with Matlab in addition to Octave.

### DIFF
--- a/tools/tune/crossover/crossover_coef_quant.m
+++ b/tools/tune/crossover/crossover_coef_quant.m
@@ -5,7 +5,7 @@ qf_iir = 30;
 
 addpath ./../eq
 
-if length(lowpass) != length(highpass)
+if length(lowpass) ~= length(highpass)
 	error("length of lowpass and highpass array do not match");
 end
 
--- a/tools/tune/crossover/crossover_gen_coefs.m
+++ b/tools/tune/crossover/crossover_gen_coefs.m
@@ -1,4 +1,4 @@
-function crossover = crossover_gen_coefs(fs, fc_low, fc_mid, fc_high);
+function crossover = crossover_gen_coefs(fs, fc_low, fc_mid, fc_high)
 
 addpath ./../eq/
 switch nargin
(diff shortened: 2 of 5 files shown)


tools/tune/crossover/crossover_coef_quant.m
```diff
@@ -5,7 +5,7 @@
 
 addpath ./../eq
 
-if length(lowpass) != length(highpass)
+if length(lowpass) ~= length(highpass)
```
btian1, line 8: seems "~=" is correct, does "!=" previously wrong?

andrula-song: The symbol for not equal to in MATLAB is "~="

ShriramShastry: It appears that the author mixed up the C [ `!=`]and Matlab [ `~=`] syntaxes

tools/tune/crossover/crossover_gen_coefs.m
```diff
@@ -54,7 +54,7 @@
 		b = [1 - cutoff, 0, 0];
 		a = [1, 0, 0];
 		return;
-	endif
```
ShriramShastry, line 57: Extra semicolon is superfluous at line number `1`, `13`, `18`, `25` and `gain_db` as input argument might be unused , please consider removing or replacing with `~`

tools/tune/crossover/crossover_plot_freq.m
```diff
@@ -18,6 +18,8 @@
 % Then to plot the transferfn for y1 we would create a filter such as:
 %  x(n) ---> LR4 LO-PASS --> LR4 LO-PASS --> y1(n)
 
+figure;
+
```
ShriramShastry, line 22: At the line number 71 and 72, `w` , please consider renaming as variable is over written and unused. at the same time , please consider removing semicolon at line `68` end .

tools/tune/crossover/example_crossover.m
```diff
@@ -1,54 +1,117 @@
 function example_crossover();
```
ShriramShastry, line 1: please consider removing semicolon at the end, it is unnecessary

ShriramShastry (commented):
I believe that further code syntax corrections are required! Thank you.

gkbldcig:

tools/tune/crossover/crossover_gen_coefs.m
singalsu: Yep, good to remove the extra semicolons. I'll leave for now the gain_db in case there's plan to add support for it later. Could multiply lp.b with 10^(gain_db/20) but I'll leave to later since it would need additional checking. Is it OK @thesofproject/google ?

singalsu (commented):
Thanks a lot for review @ShriramShastry !

lgirdwood:
@cujomalainey good for you ?

cujomalainey (approved):

kv2019i (approved):

PR #7886, opened 2023-07-03
Topology2: hdmi: use macro to set HDMI BE name

bardliao:
HDMI BE name may not be iDispx.


plbossart (approved):

bardliao:
We may need a different .conf for different display audio type. Closing it for mow.

bardliao:
Reopen it as we can reuse the macros in hdmi-default.conf when we add a new display audio type.

ranj063 (approved):

lgirdwood (approved):

PR #7887, opened 2023-07-04
Audio: MFCC: Add HiFi4 & HiFi3 implementation

andrula-song:
Add HiFi4 & HiFi3 implementation of MFCC functions.

### DIFF
--- a/src/audio/mfcc/mfcc_common.c
+++ b/src/audio/mfcc/mfcc_common.c
@@ -1,0 +1,183 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// Copyright(c) 2023 Intel Corporation. All rights reserved.
+//
+// Author: Andrula Song <andrula.song@intel.com>
+
+#include <sof/audio/mfcc/mfcc_comp.h>
+
+#include <sof/audio/component.h>
+#include <sof/audio/audio_stream.h>
+#include <sof/math/auditory.h>
+#include <sof/math/matrix.h>
+#include <sof/math/sqrt.h>
+#include <sof/math/trig.h>
+#include <sof/math/window.h>
+#include <sof/trace/trace.h>
+#include <user/mfcc.h>
+#include <errno.h>
+#include <stddef.h>
+#include <stdint.h>
+
+LOG_MODULE_REGISTER(mfcc_common, CONFIG_SOF_LOG_LEVEL);
+/* MFCC with 16 bit FFT benefits from data normalize, for 32 bits there's no
+ * significant impact. The amount of left shifts for FFT input is limited to
+ * 10 that equals about 60 dB boost. The boost is compensated in Mel energy
+ * calculation.
+ */
+#if MFCC_FFT_BITS == 16
+#define MFCC_NORMALIZE_FFT
+#else
+#undef MFCC_NORMALIZE_FFT
+#endif
+#define MFCC_NORMALIZE_MAX_SHIFT	10
+
+/*
+ * The main processing function for MFCC
+ */
+
+static int mfcc_stft_process(const struct comp_dev *dev, struct mfcc_state *state)
+{
+	struct mfcc_buffer *buf = &state->buf;
+	struct mfcc_fft *fft = &state->fft;
+	int mel_scale_shift;
+	int input_shift;
+	int i;
+	int m;
+	int cc_count = 0;
+
+	/* Phase 1, wait until whole fft_size is filled with valid data. This way
+	 * first output cepstral coefficients originate from streamed data and not
+	 * from buffers with zero data.
+	 */
+	comp_dbg(dev, "mfcc_stft_process(), avail = %d", buf->s_avail);
+	if (state->waiting_fill) {
+		if (buf->s_avail < fft->fft_size)
+			return 0;
+
+		state->waiting_fill = false;
+	}
+
+	/* Phase 2, move first prev_size data to previous data buffer, remove
+	 * samples from input buffer.
+	 */
+	if (!state->prev_samples_valid) {
+		mfcc_fill_prev_samples(buf, state->prev_data, state->prev_data_size);
+		state->prev_samples_valid = true;
+	}
+
+	/* Check if enough samples in buffer for FFT hop */
+	m = buf->s_avail / fft->fft_hop_size;
+	for (i = 0; i < m; i++) {
+		/* Clear FFT input buffer because it has been used as scratch */
+		bzero(fft->fft_buf, fft->fft_buffer_size);
+
+		/* Copy data to FFT input buffer from overlap buffer and from new samples buffer */
+		mfcc_fill_fft_buffer(state);
+
+		/* TODO: remove_dc_offset */
+
+		/* TODO: use_energy & raw_energy */
+
+#ifdef MFCC_NORMALIZE_FFT
+		/* Find block scale left shift for FFT input */
+		input_shift = mfcc_normalize_fft_buffer(state);
+#else
+		input_shift = 0;
+#endif
+
+		/* Window function */
+		mfcc_apply_window(state, input_shift);
+
+		/* TODO: use_energy & !raw_energy */
+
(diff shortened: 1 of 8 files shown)


andrula-song:
here is the Xtensa profile result:
Screenshot from 2023-07-04 10-24-27

compared with the original C version, the saving cycles percentages are:
HiFi3 version mfcc_fill_prev_samples() saves about 40.8% cycles;
HiFi3 version mfcc_sink_copy_zero_s16() saves about 82.5% cycles;
HiFi3 version mfcc_sink_copy_data_s16() saves about 60.1% cycles;
HiFi3 version mfcc_source_copy_s16() saves about 50.6% cycles;
HiFi4 version mfcc_source_copy_s16() saves about 67.0% cycles;
HiFi3 version mfcc_fill_fft_buffer() saves about 5.9% cycles;
HiFi3 version mfcc_normalize_fft_buffer() saves about 66.8% cycles;
HiFi3 version mfcc_apply_window() saves about 33.5% cycles;

btian1:
can we just split it to multiple patches? 1, 2, 3 instead of use 1, since at least hifi3 and hifi4 should be in different patches.

src/audio/mfcc/mfcc_hifi4.c
```diff
@@ -0,0 +1,549 @@
...
+
+void mfcc_generic_debug_close(void)
+{
+#ifdef DEBUGFILES_READ_FFT
+	fclose(fh_fft_out_read);
+#endif
+#ifdef DEBUGFILES_READ_MEL
+	fclose(fh_mel_read);
+#endif
+	fclose(fh_fft_in);
```
lgirdwood, line 82: Where does fclose() come from ? Is this only used for testing on the host ? Reason I ask is that this will collide with Zephyr C library.

lgirdwood (commented):
Good performance upgrades !
I'm seeing a lot of Posix file IO test/debug code in a FW module. Can the testbench do this instead ?

src/audio/mfcc/mfcc_hifi4.c
andrula-song: yeah, it is only for testing on the host. Actually I did not finish all function level optimization, the matrix functions can be optimized too. I only use the simulator statistics to make sure the output of the function is the same, but can not see the performance of the whole algorithm, so keep the file testing to sync with Seppo after he come back.

lgirdwood: ok, we need to remove the posix file APIs here since primamry module target is DSP. Lets wait for Seppo.

src/audio/mfcc/mfcc_hifi3.c
```diff
@@ -0,0 +1,554 @@
...
+		AE_S32_L_XP(temp, fft_in, fft_inc);
+	}
+#endif
+}
+
+/*
+ * The main processing function for MFCC
+ */
+
+static int mfcc_stft_process(const struct comp_dev *dev, struct mfcc_state *state)
```
singalsu, line 302: Would a mfcc_common.c be useful to keep common functions for implementations? I didn't check this function but looks like it's generic.

src/audio/mfcc/mfcc_common.c
```diff
@@ -0,0 +1,284 @@
...
+	fh_fft_out_read = fopen("ref_fft_out.txt", "r");
+#endif
+#ifdef DEBUGFILES_READ_MEL
+	fh_mel_read = fopen("ref_fft_mel.txt", "r");
+#endif
+	fh_fft_in = fopen("fft_in.txt", "w");
+	fh_fft_out = fopen("fft_out.txt", "w");
+	fh_pow = fopen("fft_pow.txt", "w");
+	fh_mel = fopen("fft_mel.txt", "w");
+	fh_ceps = fopen("ceps.txt", "w");
```
lgirdwood, line 69: Can we make this all part of the UT or do it via testbench ?

andrula-song: of course, I am trying.

singalsu: It could be a different PR though, since it's not about optimizing for HiFi. Some generic framework to write unlimited test data s16 and s32 numbers to trace and to parse the the data back to ascii text numbers would be useful.

andrula-song:
actually, I marked the definition of macro MFCC_DEBUG_WITH_TESTBENCH, that means we will never run those test code while runtime in SOF, it is the codepatch check who is not intelligent enough to figure out which line would run. But still want to keep this code for algorithm development statistic check instead of mtrace, since the output data is too much and mtrace would lose information.

lgirdwood:
> actually, I marked the definition of macro MFCC_DEBUG_WITH_TESTBENCH, that means we will never run those test code while runtime in SOF, it is the codepatch check who is not intelligent enough to figure out which line would run. But still want to keep this code for algorithm development statistic check instead of mtrace, since the output data is too much and mtrace would lose information.

Ok, but testbench should be able do this for you. The ALSA plugin can do file IO too, so these would be preffered over coding file IO in each module.

andrula-song:
> > actually, I marked the definition of macro MFCC_DEBUG_WITH_TESTBENCH, that means we will never run those test code while runtime in SOF, it is the codepatch check who is not intelligent enough to figure out which line would run. But still want to keep this code for algorithm development statistic check instead of mtrace, since the output data is too much and mtrace would lose information.
> 
> Ok, but testbench should be able do this for you. The ALSA plugin can do file IO too, so these would be preffered over coding file IO in each module.

ok, I will remove the file operations first, and this PR only contain optimization and use another PR to add script for store those statistics.

andrula-song:
is there anything need to change? @lgirdwood  @singalsu

src/audio/mfcc/mfcc_hifi4.c
```diff
@@ -0,0 +1,289 @@
...
+	struct mfcc_fft *fft = &state->fft;
+	const int fft_inc = sizeof(fft->fft_buf[0]);
+	ae_int16 *win_in = (ae_int16 *)state->window;
+	const int win_inc = sizeof(ae_int16);
+	ae_int32x2 temp;
+	ae_int16x4 win;
+	int j;
+
+#if MFCC_FFT_BITS == 16
+	/* TODO: Use proper multiply and saturate function to make sure no overflows */
```
singalsu, line 200: You can drop these comments about saturate, it's now handled with AE multiply instructions.

src/audio/mfcc/mfcc_hifi4.c
```diff
@@ -0,0 +1,289 @@
...
+	for (j = 0; j < fft->fft_size; j++) {
+		AE_L16_IP(sample, fft_in, 0);
+		AE_L16_XP(win, win_in, win_inc);
+		temp = AE_MULF16SS_00(sample, win);
+		temp = AE_SLAA32S(temp, input_shift);
+		sample =  AE_ROUND16X4F32SASYM(temp, temp);
+		AE_S16_0_XP(sample, fft_in, fft_inc);
+	}
+#else
+	/* TODO: Use proper multiply and saturate function to make sure no overflows */
```
singalsu, line 213: and this

src/audio/mfcc/mfcc_hifi3.c
```diff
@@ -0,0 +1,293 @@
...
+	for (j = 0; j < fft->fft_size; j++) {
+		AE_L16_IP(sample, fft_in, 0);
+		AE_L16_XP(win, win_in, win_inc);
+		temp = AE_MULF16SS_00(sample, win);
+		temp = AE_SLAA32S(temp, input_shift);
+		sample = AE_ROUND16X4F32SASYM(temp, temp);
+		AE_S16_0_XP(sample, fft_in, fft_inc);
+	}
+#else
+	/* TODO: Use proper multiply and saturate function to make sure no overflows */
```
singalsu, line 217: and this

lgirdwood (approved):

singalsu (approved):

PR #7889, opened 2023-07-04
Topology2: sdw-amp-generic: add 2nd amp feedback when SDW_AMP_FEEDBACK is true

bardliao:
Currently, we add the 2nd sdw amp feedback unconditionally. It will lead
to an issue when aggregated apms don't support feedback function.

Also, adding sof-tgl-rt711-rt1308-rt715.tplg in this PR.


plbossart (approved):
thanks @bardliao !

lgirdwood (approved):

PR #7890, opened 2023-07-04
Convert copier to use the module interface

ranj063:
Includes patches from #7740

### DIFF
--- a/src/audio/asrc/asrc.c
+++ b/src/audio/asrc/asrc.c
@@ -541,7 +541,7 @@ static int asrc_params(struct comp_dev *dev,
 
 	err = asrc_verify_params(dev, pcm_params);
 	if (err < 0) {
-		comp_err(dev, "src_params(): pcm params verification failed.");
+		comp_err(dev, "asrc_params(): pcm params verification failed.");
 		return -EINVAL;
 	}
 
--- a/src/audio/asrc/asrc.c
+++ b/src/audio/asrc/asrc.c
@@ -539,6 +539,10 @@ static int asrc_params(struct comp_dev *dev,
 
 	comp_info(dev, "asrc_params()");
 
+#if CONFIG_IPC_MAJOR_4
+	ipc4_base_module_cfg_to_stream_params(&cd->ipc_config.base, pcm_params);
+#endif
+
 	err = asrc_verify_params(dev, pcm_params);
 	if (err < 0) {
 		comp_err(dev, "asrc_params(): pcm params verification failed.");
@@ -552,6 +556,12 @@ static int asrc_params(struct comp_dev *dev,
 
 	source_c = buffer_acquire(sourceb);
 	sink_c = buffer_acquire(sinkb);
+
+#if CONFIG_IPC_MAJOR_4
+	/* update the source/sink buffer formats. Sink rate will be modified below */
+	ipc4_update_buffer_format(source_c, &cd->ipc_config.base.audio_fmt);
+	ipc4_update_buffer_format(sink_c, &cd->ipc_config.base.audio_fmt);
+#endif
 
 	/* Don't change sink rate if value from IPC is 0 (auto detect) */
 	if (asrc_get_sink_rate(&cd->ipc_config))
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -44,8 +44,6 @@
 #include <zephyr/drivers/dai.h>
 #endif
 
-static const struct comp_driver comp_copier;
-
 LOG_MODULE_REGISTER(copier, CONFIG_SOF_LOG_LEVEL);
 
 /* this id aligns windows driver requirement to support windows driver */
@@ -84,9 +82,6 @@ static int copier_init(struct processing_module *mod)
 
 	for (i = 0; i < IPC4_COPIER_MODULE_OUTPUT_PINS_COUNT; i++)
 		cd->out_fmt[i] = cd->config.out_fmt;
-
-	list_init(&dev->bsource_list);
-	list_init(&dev->bsink_list);
 
 	ipc_pipe = ipc_get_comp_by_ppl_id(ipc, COMP_TYPE_PIPELINE, config->pipeline_id);
 	if (!ipc_pipe) {
@@ -144,6 +139,9 @@ static int copier_init(struct processing_module *mod)
 		};
 
 		dev->direction_set = true;
+	} else {
+		/* set max sink count for module copier */
+		mod->max_sinks = IPC4_COPIER_MODULE_OUTPUT_PINS_COUNT;
 	}
 
 	dev->direction = cd->direction;
@@ -154,43 +152,7 @@ error:
 	return ret;
 }
 
-static struct comp_dev *copier_new(const struct comp_driver *drv,
-				   const struct comp_ipc_config *config,
-				   const void *spec)
-{
-	struct comp_dev *dev;
-	struct processing_module *mod;
-	struct module_data *md;
-	int ret;
-
-	comp_cl_dbg(&comp_copier, "copier_new()");
-
-	mod = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*mod));
-	if (!mod)
-		return NULL;
-
-	dev = comp_alloc(drv, sizeof(*dev));
-	if (!dev)
-		goto error_dev;
-
-	dev->ipc_config = *config;
-
-	comp_set_drvdata(dev, mod);
-	md = &mod->priv;
-	mod->dev = dev;
-	md->cfg.init_data = spec;
-
-	ret = copier_init(mod);
-	if (!ret)
-		return dev;
-
-	rfree(dev);
-error_dev:
-	rfree(mod);
-	return NULL;
-}
-
-static void copier_mod_free(struct processing_module *mod)
+static int copier_free(struct processing_module *mod)
 {
 	struct copier_data *cd = module_get_private_data(mod);
 	struct comp_dev *dev = mod->dev;
@@ -211,27 +173,25 @@ static void copier_mod_free(struct processing_module *mod)
 	}
 
 	rfree(cd);
+
+	return 0;
 }
 
-static void copier_free(struct comp_dev *dev)
-{
-	struct processing_module *mod = comp_get_drvdata(dev);
+static int copier_params(struct processing_module *mod);
 
-	copier_mod_free(mod);
-
-	rfree(mod);
-	rfree(dev);
-}
-
-static int copier_mod_prepare(struct processing_module *mod,
-			      struct sof_source __sparse_cache **sources, int num_of_sources,
-			      struct sof_sink __sparse_cache **sinks, int num_of_sinks)
+static int copier_prepare(struct processing_module *mod,
+			  struct sof_source __sparse_cache **sources, int num_of_sources,
+			  struct sof_sink __sparse_cache **sinks, int num_of_sinks)
 {
 	struct copier_data *cd = module_get_private_data(mod);
 	struct comp_dev *dev = mod->dev;
 	int ret;
 
-	comp_dbg(dev, "copier_prepare()");
+	ret = copier_params(mod);
+	if (ret < 0)
+		return ret;
+
+	comp_info(dev, "copier_prepare()");
 
 	switch (dev->ipc_config.type) {
 	case SOF_COMP_HOST:
@@ -267,28 +227,7 @@ static int copier_mod_prepare(struct processing_module *mod,
 	return 0;
 }
 
-static int copier_prepare(struct comp_dev *dev)
-{
-	struct processing_module *mod = comp_get_drvdata(dev);
-	int ret;
-
-	/* cannot configure DAI while active */
-	if (dev->state == COMP_STATE_ACTIVE) {
-		comp_info(dev, "copier_config_prepare(): Component is in active state.");
-		return 0;
-	}
-
-	ret = comp_set_state(dev, COMP_TRIGGER_PREPARE);
-	if (ret < 0)
-		return ret;
-
-	if (ret == COMP_STATUS_STATE_ALREADY_SET)
-		return PPL_STATUS_PATH_STOP;
-
-	return copier_mod_prepare(mod, NULL, 0, NULL, 0);
-}
-
-static int copier_mod_reset(struct processing_module *mod)
+static int copier_reset(struct processing_module *mod)
 {
 	struct copier_data *cd = module_get_private_data(mod);
 	struct ipc4_pipeline_registers pipe_reg;
@@ -320,18 +259,6 @@ static int copier_mod_reset(struct processing_module *mod)
 	}
 
 	return 0;
-}
-
-static int copier_reset(struct comp_dev *dev)
-{
-	struct processing_module *mod = comp_get_drvdata(dev);
-	int ret;
-
-	ret = copier_mod_reset(mod);
-
-	comp_set_state(dev, COMP_TRIGGER_RESET);
-
-	return ret;
 }
 
 static int copier_comp_trigger(struct comp_dev *dev, int cmd)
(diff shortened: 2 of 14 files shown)


src/audio/module_adapter/module_adapter.c
```diff
@@ -1317,6 +1325,13 @@ int module_adapter_trigger(struct comp_dev *dev, int cmd)
 
 	comp_dbg(dev, "module_adapter_trigger(): cmd %d", cmd);
 
+	/* handle host/DAI gateway modules separately */
+	if (dev->ipc_config.type == SOF_COMP_HOST || dev->ipc_config.type == SOF_COMP_DAI) {
+		struct module_data *md = &mod->priv;
+
+		return md->ops->endpoint_ops->trigger(dev, cmd);
```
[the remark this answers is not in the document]
ranj063: it is all handled in the module adapter. Common code doesnt need to do anything

src/include/sof/audio/module_adapter/module/module_interface.h
```diff
@@ -67,6 +67,79 @@ struct output_stream_buffer {
...
+	 */
+	int (*dai_get_hw_params)(struct comp_dev *dev,
+				 struct sof_ipc_stream_params *params, int dir);
+
+	/**
+	 * Triggers device state.
+	 * @param dev Component device.
+	 * @param cmd Trigger command.
+	 */
+	int (*trigger)(struct comp_dev *dev, int cmd);
```
RanderWang, line 140: how about get_attribute ops ? It is used to get base_module_cfg in each module. Do you plan to support it by get_configuration ?

ranj063: it is already done in the module adapter @RanderWang

src/audio/copier/copier.c
```diff
@@ -528,13 +528,49 @@ static int do_conversion_copy(struct comp_dev *dev,
...
+static int copier_copy_to_sinks(struct copier_data *cd, struct comp_dev *dev,
+				struct comp_buffer __sparse_cache *src_c,
+				struct comp_copy_limits *processed_data)
+{
+	struct comp_buffer __sparse_cache *sink_c;
+	struct list_item *sink_list;
+	struct comp_buffer *sink;
+	int ret = 0;
+
+	list_for_item(sink_list, &dev->bsink_list) {
```
lgirdwood, line 456: large code block so best to have an inline comment to say what we are doing for each item i.e. do conversion and copy for each active sink.

src/audio/copier/copier.c
```diff
@@ -528,13 +528,49 @@ static int do_conversion_copy(struct comp_dev *dev,
...
+		if (ret < 0) {
+			comp_err(dev, "failed to copy buffer for comp %x",
+				 dev->ipc_config.id);
+			break;
+		}
+	}
+
+	if (!ret) {
+		comp_update_buffer_consume(src_c, processed_data->source_bytes);
+		if (!cd->endpoint_num || cd->bsource_buffer)
```
lgirdwood, line 477: I would explain this condition with an inline comment too.

src/audio/module_adapter/module_adapter.c
```diff
@@ -1425,6 +1440,73 @@ void module_adapter_free(struct comp_dev *dev)
 	rfree(dev);
 }
 
+int module_adapter_get_hw_params(struct comp_dev *dev, struct sof_ipc_stream_params *params,
```
lgirdwood, line 1443: We should really have doxygen comments on all module adapter public APIs.

lgirdwood (commented):
LGTM, just minor comments.

src/audio/module_adapter/module_adapter.c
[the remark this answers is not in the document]
btian1: So
https://github.com/thesofproject/sof/blob/main/src/audio/pipeline/pipeline-stream.c#L370 
pipeline_comp_trigger
will be replaced by: module_adapter_trigger after convert, right?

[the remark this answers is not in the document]
ranj063: yes

RanderWang (approved):
LGTM

src/audio/copier/copier.c
ranj063: this comes from PR 7740. I will let @btian1 add the comment

btian1: adding

ranj063:
SOFCI TEST

src/audio/copier/copier.c
```diff
@@ -536,21 +463,59 @@ static int copier_copy_to_sinks(struct copier_data *cd, struct comp_dev *dev,
...
+	struct copier_data *cd = module_get_private_data(mod);
 	struct comp_buffer __sparse_cache *src_c;
 	struct comp_copy_limits processed_data;
-	struct comp_buffer *src;
-	int ret;
+	int i;
 
-	src = list_first_item(&dev->bsource_list, struct comp_buffer, sink_list);
-	src_c = buffer_acquire(src);
+	if (!num_input_buffers || !num_output_buffers)
```
btian1, line 476: this is also CI or QB fix?
for module case:
dev->ipc_config.type == SOF_COMP_NONE

can we return at:

	if (num_output_buffers == 1) {
		module_single_sink_setup(dev, source_c, sinks_c);
		if (sinks_c[0]->sink->state != dev->state) {
			num_output_buffers = 0;
			buffer_release(sinks_c[0]);
                       goto out;
		}
	} else if (num_input_buffers == 1) {
		module_single_source_setup(dev, source_c, sinks_c);
		if (source_c[0]->source->state != dev->state) {
			num_input_buffers = 0;
			buffer_release(source_c[0]);
                        goto out;
		}
	} else {
		ret = -EINVAL;
		goto out;
	}

src/audio/copier/copier_host.c
```diff
@@ -100,6 +101,9 @@ int copier_host_create(struct comp_dev *parent_dev, struct copier_data *cd,
 		ret = init_pipeline_reg(parent_dev);
 		if (ret)
 			goto e_conv;
+
+		/* set max sink count for playback */
+		mod->max_sinks = IPC4_COPIER_MODULE_OUTPUT_PINS_COUNT;
```
btian1, line 106: max_sinks need set for different cases, it depend on direction, can we move to module_adapter and set it together?

src/audio/copier/copier.c
```diff
@@ -144,6 +139,9 @@ static int copier_init(struct processing_module *mod)
 		};
 
 		dev->direction_set = true;
+	} else {
+		/* set max sink count for module copier */
+		mod->max_sinks = IPC4_COPIER_MODULE_OUTPUT_PINS_COUNT;
```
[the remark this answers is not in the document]
ranj063: Max sources and sinks is initialized to 1 in the module adapter new. Every module must update it if needed like here

src/audio/copier/copier_host.c
ranj063: Cannot be moved. This is module specific

src/audio/copier/copier.c
ranj063: Type will be NONE for every component.not just the module copier except for host or dai copier

src/audio/module_adapter/module/generic.c
```diff
@@ -83,11 +83,12 @@ int module_init(struct processing_module *mod, struct module_interface *interfac
 
 	comp_dbg(dev, "module_init() start");
 
+#if CONFIG_IPC_MAJOR_3
```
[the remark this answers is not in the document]
ranj063: We should have used the dev state for ipc3 as well but it's best left untouched at this point.

src/audio/copier/copier.c
btian1: possible to remove line 475-476 with above code? I added "goto out" for each zero case in module_adapter_audio_stream_type_copy

src/audio/module_adapter/module_adapter.c
```diff
@@ -867,6 +871,10 @@ static int module_adapter_audio_stream_type_copy(struct comp_dev *dev)
 	uint32_t num_input_buffers, num_output_buffers;
 	int ret, i = 0;
 
+	/* handle special case of HOST/DAI type components */
+	if (dev->ipc_config.type == SOF_COMP_HOST || dev->ipc_config.type == SOF_COMP_DAI)
+		return module_process_legacy(mod, NULL, 0, NULL, 0);
```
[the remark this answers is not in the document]
ranj063: @lyakh its not that the host/dai will process data with no src/sink buffers/ Its just that they're special in terms of what the source/sink buffers are. For the host playback case, the source buffer is the DMA buffer and similarly for the capture case, the sink buffer is the DMA buffer. So I completely skip the part with source/sink buffers and let the module itself handle it.

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -1118,39 +1118,28 @@ static int volume_get_config(struct processing_module *mod,
...
-	for (i = 0; i < SOF_IPC_MAX_CHANNELS; i++)
-		vol_params.chmap[i] = (mod->priv.cfg.base_cfg.audio_fmt.ch_map >> i * 4) & 0xf;
-
-	component_set_nearest_period_frames(dev, vol_params.rate);
+	component_set_nearest_period_frames(dev, params->rate);
 
 	/* volume component will only ever have 1 sink buffer */
 	sinkb = list_first_item(&dev->bsink_list, struct comp_buffer, source_list);
 	sink_c = buffer_acquire(sinkb);
-	ret = buffer_set_params(sink_c, &vol_params, true);
```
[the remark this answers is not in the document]
ranj063: it isnt dropped, it is handled by ipc4_update_buffer_format() now

src/audio/copier/copier.c
ranj063: @btian1 no unfortunately not. It is up to the module to decide how it wants to handle this case. For example, the mixin will choose to skip processing altogether whereas the mixout will need to produce silence.

src/audio/module_adapter/module_adapter.c
```diff
@@ -284,8 +273,23 @@ int module_adapter_prepare(struct comp_dev *dev)
...
+	/* Get period_bytes first on prepare(). At this point it is guaranteed that the stream
+	 * parameter from sink buffer is settled, and still prior to all references to period_bytes.
+	 */
+	sink = list_first_item(&dev->bsink_list, struct comp_buffer, source_list);
+	sink_c = buffer_acquire(sink);
+
+	mod->period_bytes = audio_stream_period_bytes(&sink_c->stream, dev->frames);
+	comp_dbg(dev, "module_adapter_prepare(): got period_bytes = %u", mod->period_bytes);
+
+	buffer_release(sink_c);
```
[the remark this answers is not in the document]
ranj063: done

src/audio/module_adapter/module/generic.c
```diff
@@ -83,11 +83,12 @@ int module_init(struct processing_module *mod, struct module_interface *interfac
...
 	if (mod->priv.state == MODULE_INITIALIZED)
 		return 0;
 	if (mod->priv.state > MODULE_INITIALIZED)
 		return -EPERM;
-
+#endif
```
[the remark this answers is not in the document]
ranj063: done

src/audio/module_adapter/module_adapter.c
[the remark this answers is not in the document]
btian1: hope in future, we can have a better way to handle all the cases instead of use "if", two disadvantages for this:
1. put more burden to last part of code execution.
2. code is not easy to understand.


like copier_copy(copier_process), we finally make it looks much more clear than initial state, I hope we can do similar on this too.

src/audio/copier/copier.c
btian1: then mixin and mixout should call: module_adapter_audio_stream_type_copy separately right? still different num_input_buffers and num_output_buffers? I just want make module process as simple as it can.

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -1070,8 +1069,7 @@ static int volume_set_config(struct processing_module *mod, uint32_t config_id,
 		return ret;
 
 	/* return if more fragments are expected or if the module is not prepared */
-	if ((pos != MODULE_CFG_FRAGMENT_LAST && pos != MODULE_CFG_FRAGMENT_SINGLE) ||
-	    md->state < MODULE_INITIALIZED)
+	if (pos != MODULE_CFG_FRAGMENT_LAST && pos != MODULE_CFG_FRAGMENT_SINGLE)
```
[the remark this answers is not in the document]
lyakh: @ranj063 right, I got it the other way round :-/ But now in a separate commit it's very clear, thanks!

btian1 (approved):

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -1118,39 +1118,28 @@ static int volume_get_config(struct processing_module *mod,
...
 
 	/* volume component will only ever have 1 sink buffer */
 	sinkb = list_first_item(&dev->bsink_list, struct comp_buffer, source_list);
 	sink_c = buffer_acquire(sinkb);
-	ret = buffer_set_params(sink_c, &vol_params, true);
+	ipc4_update_buffer_format(sink_c, &mod->priv.cfg.base_cfg.audio_fmt);
 	buffer_release(sink_c);
 
-	return ret;
+	sourceb = list_first_item(&dev->bsource_list, struct comp_buffer, sink_list);
```
btian1, line 1137: previous code only have sink buffer update,  does not have source buffer update, just want to avoid some parameter missing caused issues? do we need add source update to each module?

ranj063: for the volume component, the source buffer and sink buffer formats are exactly the same. So yes we should update both buffers. And should we do it for each module individually? Probably not, something we should work on unifying for all modules in IPC4 in a follow up PR

btian1: so, any audio processing module, like iir, fir they also need assume source and sink buffer have same format, right?

lgirdwood (approved):

kv2019i (approved):

src/audio/copier/copier.c
```diff
@@ -211,27 +173,25 @@ static void copier_mod_free(struct processing_module *mod)
...
 	struct copier_data *cd = module_get_private_data(mod);
 	struct comp_dev *dev = mod->dev;
 	int ret;
 
-	comp_dbg(dev, "copier_prepare()");
+	ret = copier_params(mod);
+	if (ret < 0)
+		return ret;
+
+	comp_info(dev, "copier_prepare()");
```
lyakh, line 195: can we keep it `comp_dbg()`?

src/audio/module_adapter/module_adapter.c
```diff
@@ -548,6 +548,10 @@ int module_adapter_params(struct comp_dev *dev, struct sof_ipc_stream_params *pa
 			return ret;
 	}
 
+#if CONFIG_IPC_MAJOR_4
+	ipc4_base_module_cfg_to_stream_params(&mod->priv.cfg.base_cfg, params);
+	ipc4_base_module_cfg_to_stream_params(&mod->priv.cfg.base_cfg, mod->stream_params);
```
lyakh, line 553: for my understanding: why is this needed? Is this fixing a bug?

src/audio/asrc/asrc.c
```diff
@@ -553,6 +557,12 @@ static int asrc_params(struct comp_dev *dev,
 	source_c = buffer_acquire(sourceb);
 	sink_c = buffer_acquire(sinkb);
 
+#if CONFIG_IPC_MAJOR_4
+	/* update the source/sink buffer formats. Sink rate will be modified below */
+	ipc4_update_buffer_format(source_c, &cd->ipc_config.base.audio_fmt);
+	ipc4_update_buffer_format(sink_c, &cd->ipc_config.base.audio_fmt);
+#endif
```
lyakh, line 564: same here - what's the effect of these changes?

lyakh (approved):
minor comments / questions only

src/audio/module_adapter/module_adapter.c
ranj063: Yes, now that the copier is not using the comp_driver, its params wont get called until prepare. So the params if we wont do this and components that do not use the module adapter yet will fail

src/audio/asrc/asrc.c
ranj063: same explanation, copier doesnt set the source/sink buffer until prepare. So asrc_params( will fail without it

lgirdwood:
SOFCI TEST

lgirdwood:
@ranj063 pls merge after CAVS CI finishes, I've restarted it as it was pending for 1 day.

ranj063:
> @ranj063 pls merge after CAVS CI finishes, I've restarted it as it was pending for 1 day.

@lgirdwood even though it shows as pending, if you click on the link, you can see the results. I have notified @fredoh9 @marc-hb about the problem

src/audio/module_adapter/module_adapter.c
lyakh: does that mean that that before this commit copier.c is broken, which breaks bisection? Should this commit be merged with https://github.com/thesofproject/sof/pull/7890/commits/6bcaf899cde497db312603408877401d24f35768 which comes before it?

Issue #7629, opened 2023-05-16
google/oss-fuzz/infra/cifuzz/actions downloads about 200 extra packages from http://archive.ubuntu.com/ubuntu 


marc-hb:
Question for @cujomalainey and @andyross 

The (older) IPC fuzzer check added by commit 39551b08ce4d uses the Github Action `google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master`

This action runs in Docker image `gcr.io/oss-fuzz-base/base-builder:latest`

Then, the action uses `apt-get` to download ~ 200 extra Ubuntu packages from http://archive.ubuntu.com/ubuntu on top of `oss-fuzz-base/base-builder`

One of the main purposes of Docker is to prepackage a complete, specialized image for the job. So it's strange to see that an image named "oss-fuzz-base" is short of 200 Ubuntu packages to... perform fuzzing.

This is not just inefficient; it may also hit some rate limits. If I were http://archive.ubuntu.com/ubuntu , I would not accept that much traffic from Github Actions. In fact this run timed out,  http://archive.ubuntu.com/ubuntu 
 stopped responding: https://github.com/thesofproject/sof/actions/runs/4993519332/jobs/8942765323

To avoid rate limits, most Github Actions download images and packages from internal mirrors like ghcr.io and http://azure.archive.ubuntu.com/

Again it would be even better if the gcr.io/oss-fuzz-base/base-builder:latest image was better self-contained and didn't need that many extra packages in the first place.

EDIT: timing out again in https://github.com/thesofproject/sof/actions/runs/5000127070/jobs/8957163851?pr=7546 / PR #7546

andyross:
This code is still in use in the oss-fuzz integration, so needs to stay there until it's replaced with the new version.  I should have that submitted today.  Once that merges in upstream, this becomes dead code and can be removed.

cujomalainey:
@marc-hb this sounds related https://github.com/google/oss-fuzz/issues/7130

andyross:
New integration submitted at https://github.com/google/oss-fuzz/pull/10342

It's somewhat lighter-weight, relying on the built-in clang and tooling more.  Though it now has to pull the Zephyr SDK for the build image (but not every time it builds the fuzzers!), which is 1G+ and constantly growing.  (Actually there is a variant with only the host tools somewhere, which is all we need for x86 fuzzing, which would be a big win.)

andyross:
[...]

PR #7891, opened 2023-07-04
ipc3: get_drv: Check comp_type also when driver is matched with UUID

jsarha:
Add a safe guard against malformed SOF_IPC_TPLG_COMP_NEW messages where the UUID matched driver type does not match the component type in the message.

Reported-by: Curtis Malainey <curtis@malainey.com>

related discussions can be found here: https://github.com/thesofproject/sof/pull/7830

### DIFF
--- a/src/include/sof/audio/module_adapter/module/generic.h
+++ b/src/include/sof/audio/module_adapter/module/generic.h
@@ -33,7 +33,7 @@
 				(value)); \
 	} while (0)
 
-#define DECLARE_MODULE_ADAPTER(adapter, uuid, tr) \
+#define DECLARE_MODULE_ADAPTER_WITH_TYPE(adapter, uuid, tr, comp_type)		\
 static struct comp_dev *module_##adapter##_shim_new(const struct comp_driver *drv, \
 					 const struct comp_ipc_config *config, \
 					 const void *spec) \
@@ -42,7 +42,7 @@ static struct comp_dev *module_##adapter##_shim_new(const struct comp_driver *dr
 } \
 \
 static const struct comp_driver comp_##adapter##_module = { \
-	.type = SOF_COMP_MODULE_ADAPTER, \
+	.type = comp_type, \
 	.uid = SOF_RT_UUID(uuid), \
 	.tctx = &(tr), \
 	.ops = { \
@@ -74,6 +74,9 @@ UT_STATIC void sys_comp_module_##adapter##_init(void) \
 } \
 \
 DECLARE_MODULE(sys_comp_module_##adapter##_init)
+
+#define DECLARE_MODULE_ADAPTER(adapter, uuid, tr) \
+	DECLARE_MODULE_ADAPTER_WITH_TYPE(adapter, uuid, tr, SOF_COMP_MODULE_ADAPTER)
 
 /**
  * \enum module_state
--- a/src/ipc/ipc3/helper.c
+++ b/src/ipc/ipc3/helper.c
@@ -121,6 +121,11 @@ static const struct comp_driver *get_drv(struct sof_ipc_comp *comp)
 		goto out;
 	}
 
+	if (comp->type > SOF_COMP_MODULE_ADAPTER) {
+		tr_err(&comp_tr, "Bad component type: %d\n", comp->type);
+		goto out;
+	}
+
 	/* search driver list with UUID */
 	key = k_spin_lock(&drivers->lock);
 
@@ -134,16 +139,23 @@ static const struct comp_driver *get_drv(struct sof_ipc_comp *comp)
 		}
 	}
 
-	if (!drv)
+	k_spin_unlock(&drivers->lock, key);
+
+	if (!drv) {
 		tr_err(&comp_tr,
 		       "get_drv(): the provided UUID (%8x%8x%8x%8x) doesn't match to any driver!",
 		       *(uint32_t *)(&comp_ext->uuid[0]),
 		       *(uint32_t *)(&comp_ext->uuid[4]),
 		       *(uint32_t *)(&comp_ext->uuid[8]),
 		       *(uint32_t *)(&comp_ext->uuid[12]));
+		goto out;
+	}
 
-	k_spin_unlock(&drivers->lock, key);
-
+	if (drv->type != comp->type && drv->type != SOF_COMP_MODULE_ADAPTER) {
+		tr_err(&comp_tr, "get_drv(): Found %pU driver and IPC type differ %d != %d",
+		       drv->tctx->uuid_p, drv->type, comp->type);
+		drv = NULL;
+	}
 out:
 	if (drv)
 		tr_dbg(&comp_tr, "get_drv(), found driver type %d, uuid %pU",
--- a/src/audio/module_adapter/module/volume/volume.c
+++ b/src/audio/module_adapter/module/volume/volume.c
@@ -1671,7 +1671,7 @@ static struct module_interface gain_interface = {
 	.free = volume_free
 };
 
-DECLARE_MODULE_ADAPTER(gain_interface, gain_uuid, gain_tr);
+DECLARE_MODULE_ADAPTER_WITH_TYPE(gain_interface, gain_uuid, gain_tr, SOF_COMP_VOLUME);
 SOF_MODULE_INIT(gain, sys_comp_module_gain_interface_init);
 #endif
 #endif
--- a/src/audio/src/src.c
+++ b/src/audio/src/src.c
@@ -1058,5 +1058,5 @@ static struct module_interface src_interface = {
 	.free = src_free,
 };
 
-DECLARE_MODULE_ADAPTER(src_interface, src_uuid, src_tr);
+DECLARE_MODULE_ADAPTER_WITH_TYPE(src_interface, src_uuid, src_tr, SOF_COMP_SRC);
 SOF_MODULE_INIT(src, sys_comp_module_src_interface_init);


marc-hb:
The fuzzing failure seems relevant.

cc: @andyross your work in action!

This is funny: I just clicked on the most recent PR looking only for fuzzer _startup times_ in order to close #7629 and I find what looks like an actual catch.

```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==5247==ERROR: AddressSanitizer: SEGV on unknown address 0x00000000 (pc 0x081e7924 bp 0xf20fed98 sp 0xf20fed50 T7)
==5247==The signal is caused by a READ memory access.
==5247==Hint: address points to the zero page.
    #0 0x81e7924 in get_drv /home/runner/work/sof/sof/workspace/sof/src/ipc/ipc3/helper.c:147:11
    #1 0x81e6ea7 in comp_new /home/runner/work/sof/sof/workspace/sof/src/ipc/ipc3/helper.c:322:8
    #2 0x81eb55a in ipc_comp_new /home/runner/work/sof/sof/workspace/sof/src/ipc/ipc3/helper.c:659:7
    #3 0x81df7d6 in ipc_glb_tplg_comp_new /home/runner/work/sof/sof/workspace/sof/src/ipc/ipc3/handler.c:1298:8
    #4 0x81dede6 in ipc_cmd /home/runner/work/sof/sof/workspace/sof/src/ipc/ipc3/handler.c:1680:9
    #5 0x81a0319 in ipc_platform_do_cmd /home/runner/work/sof/sof/workspace/sof/src/platform/posix/ipc.c:162:2
    #6 0x81a433c in ipc_do_cmd /home/runner/work/sof/sof/workspace/sof/src/ipc/ipc-common.c:330:9
    #7 0x81ce96b in task_run /home/runner/work/sof/sof/workspace/sof/zephyr/include/rtos/task.h:94:9
    #8 0x81ce469 in edf_work_handler /home/runner/work/sof/sof/workspace/sof/zephyr/edf_schedule.c:31:16
    #9 0x828b533 in work_queue_main /home/runner/work/sof/sof/workspace/zephyr/kernel/work.c:668:3
    #10 0x8168102 in z_thread_entry /home/runner/work/sof/sof/workspace/zephyr/lib/os/thread_entry.c:36:2
    #11 0x8134e49 in __asan::AsanThread::ThreadStart(unsigned long long) (/home/runner/work/sof/sof/workspace/build-fuzz/zephyr/zephyr.exe+0x8134e49)

```

jsarha:
> The fuzzing failure seems relevant.
> 
> cc: @andyross your work in action!

This was excellent. My quick test if the this kind of fix would cause other problems (if the comp type check would be too strict in some cases) had a weakness that would not show in the normal tests, but the fuzzer found it immediately.

jsarha:
I think this is now ready for review even if CI tests are still running.

lgirdwood (approved):

src/ipc/ipc3/helper.c
```diff
@@ -134,16 +139,23 @@ static const struct comp_driver *get_drv(struct sof_ipc_comp *comp)
...
 		       *(uint32_t *)(&comp_ext->uuid[0]),
 		       *(uint32_t *)(&comp_ext->uuid[4]),
 		       *(uint32_t *)(&comp_ext->uuid[8]),
 		       *(uint32_t *)(&comp_ext->uuid[12]));
+		goto out;
+	}
 
-	k_spin_unlock(&drivers->lock, key);
-
+	if (drv->type != comp->type && drv->type != SOF_COMP_MODULE_ADAPTER) {
```
lyakh, line 154: what exactly is this condition: `drv->type == SOF_COMP_MODULE_ADAPTER`? So it's possible to have a component in topology with UUID, matching `SOF_COMP_MODULE_ADAPTER` but with a different type? Or how is this possible?

cujomalainey: I see what you are doing here, but it doesn't validate that the comp type is the right type. E.g. this still leaves the possibility for a SRC UUID with a VOL ENUM.

@lyakh how terrible would it be if we just moved that busted switch case out to each component and just `container_of` our way back to the correct pointer for vol and src?

cujomalainey (changes requested):
patch doesn't provide a way to make sure enum is valid.

Also please use my work email in the tag, thanks

cujomalainey@chromium.org

src/ipc/ipc3/helper.c
```diff
@@ -121,6 +121,11 @@ static const struct comp_driver *get_drv(struct sof_ipc_comp *comp)
 		goto out;
 	}
 
+	if (comp->type > SOF_COMP_MODULE_ADAPTER) {
+		tr_err(&comp_tr, "Bad component type: %d\n", comp->type);
+		goto out;
```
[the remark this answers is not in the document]
jsarha: Yes... maybe in the history lock have been kept trough out the function. No point for the label really now. I can drop it, unless we change the approach again...

src/ipc/ipc3/helper.c
jsarha: For instance MUX components use module adapter. I could not figure out how dig out what is hidden behind the module adapter framework. The nature of IPC3 is still quite strange to me.

jsarha:
This fix is not really good for anything, but now that I am trying to follow the original root cause[1] I am little by little leaning back to my original fix [2] or a variation of it. The actual root cause and the fundamental problem with the init message interpretation is the fact that the sender can send what ever payload to what ever module.  Fixing the switch-case in module_adapter_new() is not enough. The comp_common_builder() and comp_specific_builder() will anyway create the 'union ipc_config_specific spec' according to the found ipc type copy stuff from the actual payload according to it. 

So about generic fix, I do not think there is any low hanging fruit. One way could be adding the accepted message types to struct comp_driver and give it as an extra parameter to DECLARE_MODULE_ADAPTER(), and checking the type in  module_adapter_new().

The above would fix the issue we are working on, but if that really a generic fix? The fuzzer does not obey any message structures and we really cannot make any assumptions of the passed configuration data in the module. I do not see there is really any way around that the modules just have to be made robust enough to not to crash on bad or missing configuration data.

[1] https://github.com/thesofproject/sof/pull/7830#issuecomment-1615007899
[2] https://github.com/thesofproject/sof/pull/7830

src/ipc/ipc3/helper.c
cujomalainey: bleh, my `container_of` approach won't work since the data is parsed and copied, not aliased. 

Honestly I am more and more inclined to just suggest we go back to #7830 (with an additional guard for vol) and move that check into a IPC3 only block. There is not easy way fix the protocol without making a major revision (deprecating the enum field) and fixing legacy component init payloads. Also if we ifdef it around IPC3 it will be very clear it can be deleted when IPC3 is removed from the repo.

cujomalainey:
> What do you say @lyakh & @cujomalainey ?

Looks like we were typing at the same time :) see https://github.com/thesofproject/sof/pull/7891#discussion_r1254868606

src/ipc/ipc3/helper.c
btian1: we can try to modify module_adapter.c to resolve this issue.

lyakh: @cujomalainey @jsarha it seems to me that we just have to go with the "first principles" here: if we don't trust what we get over IPC, then we need to check everything. E.g. `.type` is checked in `comp_specific_builder()` but unrecognised type is just silently ignored and that doesn't seem right to me. Because if we do happen to have `comp->ext_data_length != 0` then `get_drv()` might use UUID to get a matching driver and we'll continue with an invalid type.
Then we also get size from those topology IPCs. And in general it's valid for module-adapter components to have a size of 0. In that case no initialisation data is allocated. So individual drivers must check for that, yes.

gkbldcig:

jsarha:
This fix is not very useful since it cannot check message types through module_adapter.

jsarha:
... let's test the concept.

lgirdwood:
@jsarha some build issues ?

src/include/sof/audio/module_adapter/module/generic.h
```diff
@@ -42,7 +42,7 @@ static struct comp_dev *module_##adapter##_shim_new(const struct comp_driver *dr
 } \
 \
 static const struct comp_driver comp_##adapter##_module = { \
-	.type = SOF_COMP_MODULE_ADAPTER, \
+	.type = comp_type, \
```
lyakh, line 45: this is a rather significant change, and I think it will break at least this check: https://github.com/thesofproject/sof/blob/70272282dbfeba2e0df7653cb104647968d2f30b/src/audio/module_adapter/module_adapter.c#L135 and possibly more. Whereas my proposal to change `comp_specific_builder()` would only affect IPC3. @ranj063 do you think this is a good idea?

jsarha: @lyakh , but how would I know in comp_specific_builder() what kind of messages the module - found with uuid - actually accepts, unless its somehow built in to the module? I can of course add an extra member to struct comp_driver, but I do not think this issue can not be solved without adding the accepted componen message type to each module.

lyakh: No, don't think it's possible to check the type generically in common code for all components with the current module-adapter API. That's why I'm not proposing that. Instead I'm just proposing to (1) check that the type value in IPC is meaningful - one of the values, that the firmware recognises, and error out otherwise, instead of silently ignoring it, and (2) in each individual component driver check that the type from the IPC is correct

jsarha: Ok, simple enough.

PR #7892, opened 2023-07-05
Fix memory allocation for input/output buffers in the module adapter

ranj063:


src/audio/module_adapter/module_adapter.c
```diff
@@ -865,8 +864,7 @@ static int module_adapter_audio_stream_type_copy(struct comp_dev *dev)
 	struct comp_buffer __sparse_cache *sinks_c[PLATFORM_MAX_STREAMS];
 	struct processing_module *mod = comp_get_drvdata(dev);
 	struct list_item *blist;
-	uint32_t num_input_buffers = 0;
-	uint32_t num_output_buffers = 0;
+	uint32_t num_input_buffers, num_output_buffers;
 	int ret, i = 0;
 
 	if (mod->stream_copy_single_to_single)
```
btian1, line 870: there is a similar code in pipe 2.0 DP PR2, how about move i = 0; to before line above line 873? this will make line 873 - 880 and 882 - 889 looks symmetry。

src/audio/module_adapter/module_adapter.c
```diff
@@ -879,23 +877,30 @@ static int module_adapter_audio_stream_type_copy(struct comp_dev *dev)
...
+	num_input_buffers = i;
 
 	/* setup active input/output buffers for processing */
-	if (mod->num_output_buffers == 1) {
+	if (num_output_buffers == 1) {
 		num_input_buffers = module_single_sink_setup(dev, source_c, sinks_c);
 		if (sinks_c[0]->sink->state == dev->state)
 			num_output_buffers = 1;
+		else
+			num_output_buffers = 0;
```
btian1, line 897: if (sinks_c[0]->sink->state ！= dev->state)
        num_output_buffers = 0;

4 lines to 2 lines.

src/audio/module_adapter/module_adapter.c
```diff
@@ -879,23 +877,30 @@ static int module_adapter_audio_stream_type_copy(struct comp_dev *dev)
...
 	} else {
 		num_output_buffers = module_single_source_setup(dev, source_c, sinks_c);
 		if (source_c[0]->source->state == dev->state)
 			num_input_buffers = 1;
+		else
+			num_input_buffers = 0;
```
btian1, line 903: same with above.

src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -123,6 +123,7 @@ static int mixin_init(struct processing_module *mod)
 	dev->ipc_config.frame_fmt = valid_fmt;
 	mod->skip_src_buffer_invalidate = true;
 
+	mod->max_sinks = MIXIN_MAX_SINKS;
```
lgirdwood, line 126: Would this be a Kconfig or is it set in the IPC4 ABI ?

ranj063: @lgirdwood it is set in the mixin/mixout header and doesn't seem to be bound by ABI. But the mixin_mixout module only works with IPC4 so I guess indirectly it is

btian1 (approved):

lgirdwood (approved):

PR #7894, opened 2023-07-05
sof: new header with tlv

tmleman:
Adding a header file containing definitions of the TLV structure and related functions.

Most of the code is a copy of the functionality added for the Base_Fw module in the 3efd78048d commit. The only difference are the names change of the base structure, functions and field in the structure. A new feature is a function that allows to find TLV value of the indicated type.

The reason for these changes is the fact that data in the form of TLV will also be transferred to other modules.

### DIFF
--- a/src/audio/base_fw.c
+++ b/src/audio/base_fw.c
@@ -4,6 +4,7 @@
 #include <sof/audio/component.h>
 #include <sof/lib/memory.h>
 #include <sof/ut.h>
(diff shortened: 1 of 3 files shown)


RanderWang (approved):
LGTM. Thanks!

btian1 (approved):

lgirdwood (approved):

Issue #7899, opened 2023-07-06
[BUG] Topology2: sdw amp doesn't work with passthrough topology (invalid input num_pins: 0)


bardliao:
**Describe the bug**
sdw amp doesn't work with `PASSTHROUGH=true` topologies.

**To Reproduce**
Add `PASSTHROUGH=true` in tplg-targets.cmake. Like
```patch
diff --git a/tools/topology/topology2/avs-tplg/tplg-targets.cmake b/tools/topology/topology2/avs-tplg/tplg-targets.cmake
index 5aeb446793ba..7a4dc85b9ac1 100644
--- a/tools/topology/topology2/avs-tplg/tplg-targets.cmake
+++ b/tools/topology/topology2/avs-tplg/tplg-targets.cmake
@@ -17,7 +17,7 @@ PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-hda-generic-4ch.bin,USE_CHAIN_DMA=true

 # IPC4 topology for TGL rt711 Headset + rt1316 Amplifier + rt714 DMIC
 "cavs-sdw\;sof-tgl-rt711-rt1316-rt714\;NUM_SDW_AMP_LINKS=2,SDW_DMIC=1,\
-PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-tgl-rt711-rt1316-rt714.bin"
+PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-tgl-rt711-rt1316-rt714.bin,PASSTHROUGH=true"

 "cavs-sdw\;sof-adl-rt711-l0-rt1316-l12-rt714-l3\;NUM_SDW_AMP_LINKS=2,SDW_DMIC=1"

```

Run `aplay -Dhw:0,2 test.wav -vv`

**Reproduction Rate**
100%

**Expected behavior**
aplay works without errors.

**Impact**
We can't test sdw amp with passthrough topologies.

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel: {43f9252ea5a497a7952cb1efc9a94d10d28c6dae}
    * SOF: {8adbb850448a7a6154cfa222e5cb44184416354c}
2) Name of the topology file
    * Topology: {sof-tgl-rt711-rt1316-rt714.tplg} and set PASSTHROUGH=true
3) Name of the platform(s) on which the bug is observed.
    * Platform: {TGL}

[...]

PR #7900, opened 2023-07-06
Topology2: sdw-amp-generic: add missing num_input_pins

bardliao:
num_input_pins was missing in passthrough mode.

Fixes: #7899

### DIFF
--- a/tools/topology/topology2/platform/intel/sdw-amp-generic.conf
+++ b/tools/topology/topology2/platform/intel/sdw-amp-generic.conf
@@ -111,6 +111,7 @@ IncludeByKey.PASSTHROUGH {
 			{
 				stream_name $SDW_SPK_STREAM
 				node_type $ALH_LINK_OUTPUT_CLASS
+				num_input_pins 1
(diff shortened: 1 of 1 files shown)


plbossart (approved):
Thanks @bardliao 

@ranj063 can't we flag such errors at build time with a mandatory property?

ranj063:
> Thanks @bardliao
> 
> @ranj063 can't we flag such errors at build time with a mandatory property?

@plbossart unfortunately not with this one because num_input_pins is not mandatory for the copier.

ranj063 (approved):

plbossart:
> @plbossart unfortunately not with this one because num_input_pins is not mandatory for the copier.

What prevents us from making it mandatory? I thought the two sides are needed for DMA config and buffer config.

ranj063:
> > @plbossart unfortunately not with this one because num_input_pins is not mandatory for the copier.
> 
> What prevents us from making it mandatory? I thought the two sides are needed for DMA config and buffer config.

I can look into it in a follow up PR

Issue #5212, opened 2022-01-14
Why do we use -fno-inline-functions?


marc-hb:
-fno-inline-functions was added in 2016 by giant commit c0dfb4e62ab6ef7 `core: initial import of open source DSP firmware`

Then it was transferred from automake to CMake in 2019 by commit e0ba98d21ae8 `cmake: add CMakeLists for firmware build`

It was copied to the Zephyr build by initial Zephyr commit de41202f8fea

None of these explains why.

marc-hb:
From @andyross about Zephyr:
> FWIW: within Zephyr, almost all the spots where we're inlining for performance (or, very occasionally, correctness) are tagged ALWAYS_INLINE (which expands to the always_inline gcc attribute), and won't be affected.  Still, seems like something that should be on absent weird compiler bugs.

marc-hb:
#5213 test removes the flag and shows "typical" failures, not worse than usual.

lgirdwood:
This was based on historical XCC settings so will need more analysis for code size, perf and stability.

marc-hb:
cc: @aborisovich , @mmaka1 , @mwasko , @abonislawski

aborisovich:
Ok so looking at [standards](http://www.open-std.org/JTC1/SC22/WG14/www/standards) and citating latest C99 documentation [WG14 N1256](http://www.open-std.org/JTC1/SC22/WG14/www/docs/n1256.pdf):
6.7.4 § 3:
> An inline definition of a function with external linkage shall not contain a definition of a
modifiable object with static storage duration, and shall not contain a reference to an
identifier with internal linkage

Well maybe we have a lot of definitions of modifiable objects with static storage duration inside a functions that have a external linkage? It seems true to me as there are tons of macros defining some static structs in the function bodies?

Next thing its probably so called "i need to control everything C programmer habits". 
Standard says 6.7.4 § 5:

> (...) Making a function an inline function suggests that calls to the function be as
fast as possible. The extent to which such suggestions are effective is
implementation-defined.

As far as I know there had been tons of effort invested into compilers so they can effectively make decisions on what is more effective - inlining a function or creating a new stack frame. That being said, I am not sure how much of this effort had been put into the "old standards" (which C99 certainly is).

We could probably try to approach it practically by simple performance comparisons.
Let's just try to turn this off (-fno-inline-functions flag), remove those ALWAYS_INLINE (replace then with stadard "inline" keyword hint) and see what a result will be.

[...]

PR #3251, opened 2020-07-29
Pcm converter: return error code

ktrzcinx:
Assertion may be replaced with error code return, then error won't be so fatal and error log may be traced in client component context.

PR #7909, opened 2023-07-10
module adapter and copier performance optimization

btian1:
this PR did following things:
    optimize module adapter and copier, reduce host and dai performance increase caused by #7890.
    below is the rough result:
image

### DIFF
--- a/src/audio/dai-zephyr.c
+++ b/src/audio/dai-zephyr.c
@@ -259,7 +259,6 @@ dai_dma_cb(struct dai_data *dd, struct comp_dev *dev, uint32_t bytes,
 	if (dev->direction == SOF_IPC_STREAM_PLAYBACK) {
 		ret = dma_buffer_copy_to(local_buf, dma_buf,
 					 dd->process, bytes);
-		buffer_release(local_buf);
 	} else {
 		struct list_item *sink_list;
 
@@ -269,7 +268,6 @@ dai_dma_cb(struct dai_data *dd, struct comp_dev *dev, uint32_t bytes,
 		 * so no need to check the return value of dma_buffer_copy_from_no_consume().
 		 */
 		ret = dma_buffer_copy_from_no_consume(dma_buf, local_buf, dd->process, bytes);
-		buffer_release(local_buf);
 #if CONFIG_IPC_MAJOR_4
 		/* Skip in case of endpoint DAI devices created by the copier */
 		if (converter) {
@@ -318,7 +316,6 @@ err:
 		audio_stream_consume(&dma_buf->stream, bytes);
 	}
 
-	local_buf = buffer_acquire(dd->local_buffer);
 	/* assert dma_buffer_copy succeed */
 	if (ret < 0) {
 		struct comp_buffer __sparse_cache *source_c, *sink_c;
@@ -997,8 +994,11 @@ int dai_common_params(struct dai_data *dd, struct comp_dev *dev,
 		buffer_c = buffer_acquire(dd->dma_buffer);
 		buffer_set_params(buffer_c, &hw_params,
 				  BUFFER_UPDATE_FORCE);
+		dd->sampling = get_sample_bytes(hw_params.frame_fmt);
 		buffer_release(buffer_c);
 	}
+
+	dd->fast_mode = dd->ipc_config.feature_mask & BIT(IPC4_COPIER_FAST_MODE);
 
 	return dev->direction == SOF_IPC_STREAM_PLAYBACK ?
 		dai_playback_params(dd, dev, period_bytes, period_count) :
@@ -1507,8 +1507,7 @@ static void set_new_local_buffer(struct dai_data *dd, struct comp_dev *dev)
 /* copy and process stream data from source to sink buffers */
 int dai_common_copy(struct dai_data *dd, struct comp_dev *dev, pcm_converter_func *converter)
 {
-	uint32_t dma_fmt;
-	uint32_t sampling;
+	uint32_t sampling = dd->sampling;
 	struct comp_buffer __sparse_cache *buf_c;
 	struct dma_status stat;
 	uint32_t avail_bytes;
@@ -1539,13 +1538,6 @@ int dai_common_copy(struct dai_data *dd, struct comp_dev *dev, pcm_converter_fun
 
 	avail_bytes = stat.pending_length;
 	free_bytes = stat.free;
-
-	buf_c = buffer_acquire(dd->dma_buffer);
-
-	dma_fmt = audio_stream_get_frm_fmt(&buf_c->stream);
-	sampling = get_sample_bytes(dma_fmt);
-
-	buffer_release(buf_c);
 
 	/* handle module runtime unbind */
 	if (!dd->local_buffer) {
@@ -1608,7 +1600,7 @@ int dai_common_copy(struct dai_data *dd, struct comp_dev *dev, pcm_converter_fun
 	 * in order to avoid high load spike
 	 * if FAST_MODE is enabled, then one period limitation is omitted
 	 */
-	if (!(dd->ipc_config.feature_mask & BIT(IPC4_COPIER_FAST_MODE)))
+	if (!dd->fast_mode)
 		samples = MIN(samples, dd->period_bytes / sampling);
 
 	copy_bytes = samples * sampling;
--- a/src/audio/host-legacy.c
+++ b/src/audio/host-legacy.c
@@ -970,12 +970,6 @@ static int host_reset(struct comp_dev *dev)
 	return 0;
 }
 
-/* copy and process stream data from source to sink buffers */
-int host_common_copy(struct host_data *hd, struct comp_dev *dev, copy_callback_t cb)
-{
-	return hd->copy(hd, dev, cb);
-}
-
 static int host_copy(struct comp_dev *dev)
 {
 	struct host_data *hd = comp_get_drvdata(dev);
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -880,7 +880,11 @@ static int module_adapter_audio_stream_type_copy(struct comp_dev *dev)
 
 	/* handle special case of HOST/DAI type components */
 	if (dev->ipc_config.type == SOF_COMP_HOST || dev->ipc_config.type == SOF_COMP_DAI)
+#if CONFIG_IPC_MAJOR_3
 		return module_process_legacy(mod, NULL, 0, NULL, 0);
+#else
+		return module_process_stream(mod, NULL, 0, NULL, 0);
+#endif
 
 	if (mod->stream_copy_single_to_single)
 		return module_adapter_audio_stream_copy_1to1(dev);
--- a/src/include/sof/lib/dai-zephyr.h
+++ b/src/include/sof/lib/dai-zephyr.h
@@ -168,6 +168,10 @@ struct dai_data {
 
 	/* llp slot info in memory windows */
 	struct llp_slot_info slot_info;
+	/* save current sampling for current dai device */
+	uint32_t sampling;
+	/* fast mode, use one byte memory to save repreated cycles */
+	bool fast_mode;
 };
 
 /* these 3 are here to satisfy clk.c and ssp.h interconnection, will be removed leter */
--- a/src/audio/host-zephyr.c
+++ b/src/audio/host-zephyr.c
@@ -1077,12 +1077,6 @@ static int host_reset(struct comp_dev *dev)
 	return 0;
 }
 
-/* copy and process stream data from source to sink buffers */
-int host_common_copy(struct host_data *hd, struct comp_dev *dev, copy_callback_t cb)
-{
-	return hd->copy(hd, dev, cb);
-}
-
 static int host_copy(struct comp_dev *dev)
 {
 	struct host_data *hd = comp_get_drvdata(dev);
--- a/src/include/sof/audio/host_copier.h
+++ b/src/include/sof/audio/host_copier.h
@@ -112,7 +112,11 @@ void host_common_reset(struct host_data *hd, uint16_t state);
 int host_common_trigger(struct host_data *hd, struct comp_dev *dev, int cmd);
 int host_common_params(struct host_data *hd, struct comp_dev *dev,
 		       struct sof_ipc_stream_params *params, notifier_callback_t cb);
-int host_common_copy(struct host_data *hd, struct comp_dev *dev, copy_callback_t cb);
+/* copy and process stream data from source to sink buffers */
+static inline int host_common_copy(struct host_data *hd, struct comp_dev *dev, copy_callback_t cb)
+{
+	return hd->copy(hd, dev, cb);
+}
 void host_common_update(struct host_data *hd, struct comp_dev *dev, uint32_t bytes);
 void host_common_one_shot(struct host_data *hd, uint32_t bytes);
 int copier_host_create(struct comp_dev *dev, struct copier_data *cd,
--- a/src/include/sof/audio/module_adapter/module/generic.h
+++ b/src/include/sof/audio/module_adapter/module/generic.h
@@ -356,4 +356,16 @@ int find_module_source_index(struct module_source_info __sparse_cache *msi,
 	return -EINVAL;
 }
 
+static inline int module_process_stream(struct processing_module *mod,
+					struct input_stream_buffer *input_buffers,
+					int num_input_buffers,
+					struct output_stream_buffer *output_buffers,
+					int num_output_buffers)
+{
+	struct module_data *md = &mod->priv;
+
+	return md->ops->process_audio_stream(mod, input_buffers, num_input_buffers,
+					     output_buffers, num_output_buffers);
+}
+
 #endif /* __SOF_AUDIO_MODULE_GENERIC__ */


src/audio/module_adapter/module_adapter.c
```diff
@@ -866,7 +866,7 @@ static int module_adapter_audio_stream_copy_1to1(struct comp_dev *dev)
 	return ret;
 }
 
-static int module_adapter_audio_stream_type_copy(struct comp_dev *dev)
+static inline int module_adapter_audio_stream_type_copy(struct comp_dev *dev)
```
lyakh, line 869: I'm suspicious about this. We usually assume that the compiler mostly ignores `inline` in .c files and decides by itself which functions to inline. And typically it does inline static functions that are only called once. Have you checked the resulting object? With which compiler does this make a difference?

src/audio/module_adapter/module_adapter.c
```diff
@@ -871,13 +871,14 @@ static inline int module_adapter_audio_stream_type_copy(struct comp_dev *dev)
...
 	struct processing_module *mod = comp_get_drvdata(dev);
+	struct module_data *md = &mod->priv;
 	struct list_item *blist;
 	uint32_t num_input_buffers, num_output_buffers;
 	int ret, i = 0;
 
 	/* handle special case of HOST/DAI type components */
 	if (dev->ipc_config.type == SOF_COMP_HOST || dev->ipc_config.type == SOF_COMP_DAI)
-		return module_process_legacy(mod, NULL, 0, NULL, 0);
+		return md->ops->process_audio_stream(mod, NULL, 0, NULL, 0);
```
lyakh, line 881: what about setting `md->state`?

src/audio/pcm_converter/pcm_converter.c
```diff
@@ -33,9 +33,9 @@ int pcm_convert_as_linear(const struct audio_stream __sparse_cache *source, uint
 
 	/* assert enough avail/free samples in source and sink buffer */
 	if (audio_stream_get_avail_samples(source) < samples + ioffset)
-		return -EINVAL;
+		return;
 	if (audio_stream_get_free_samples(sink) < samples + ooffset)
-		return -EINVAL;
+		return;
```
lyakh, line 38: this loses these potential errors, but yes, I can see, that possibly they aren't needed

src/audio/dai-zephyr.c
```diff
@@ -222,7 +222,7 @@ static int dai_get_fifo(struct dai *dai, int direction, int stream_id)
 }
 
 /* this is called by DMA driver every time descriptor has completed */
-static enum dma_cb_status
+static inline enum dma_cb_status
```
lyakh, line 225: same here - this shouldn't matter

src/audio/dai-zephyr.c
```diff
@@ -308,8 +307,8 @@ dai_dma_cb(struct dai_data *dd, struct comp_dev *dev, uint32_t bytes,
 				}
 
 				if (sink_dev && sink_dev->state == COMP_STATE_ACTIVE)
-					ret = dma_buffer_copy_from_no_consume(dma_buf, sink_c,
-									      converter[j], bytes);
+					dma_buffer_copy_from_no_consume(dma_buf, sink_c,
+									converter[j], bytes);
 err:
```
lyakh, line 310: not related to this your PR and probably not relevant for performance, but could we make the code above more readable and less error prone by removing those mid-function jumps and rewriting it as
```
	if (j >= IPC4_COPIER_MODULE_OUTPUT_PINS_COUNT) {
		comp_err(dev, "Sink queue ID: %d >= max output pin count: %d\n",
			 j, IPC4_COPIER_MODULE_OUTPUT_PINS_COUNT);
		ret = -EINVAL;
	} else if (!converter[j]) {
		comp_err(dev, "No PCM converter for sink queue %d\n", j);
		ret = -EINVAL;
	} else if (sink_dev && sink_dev->state == COMP_STATE_ACTIVE) {
		ret = dma_buffer_copy_from_no_consume(dma_buf, sink_c,
						      converter[j], bytes);
	}
```

lyakh (commented):
how about single-line wrappers in src/audio/source_api_helper.c and src/audio/sink_api_helper.c - they're in .c files so they're never inlined and I think some of them might be called on hot paths.

src/audio/module_adapter/module_adapter.c
btian1: seems md->state is for ipc3 special, and copier_process does not depend on this, so there is no error reported in CI and QB.

src/audio/pcm_converter/pcm_converter.c
btian1: yeah, I did not remove it totally, we can try to remove those checks later.
SOF worry too much about each possible error and this cause much more overhead in component.

src/audio/dai-zephyr.c
btian1: Those code are added may be can resolve some CI or QB failures, @ranj063 may comment on this.

Actually not only this, there is a lot of code, like below in dai_common_copy:

	/* Check possibility of glitch occurrence */
	if (dev->direction == SOF_IPC_STREAM_PLAYBACK &&
	    copy_bytes + avail_bytes < dd->period_bytes)
		comp_warn(dev, "dai_common_copy(): Copy_bytes %d + avail bytes %d < period bytes %d, possible glitch",
			  copy_bytes, avail_bytes, dd->period_bytes);
	else if (dev->direction == SOF_IPC_STREAM_CAPTURE &&
		 copy_bytes + free_bytes < dd->period_bytes)
		comp_warn(dev, "dai_common_copy(): Copy_bytes %d + free bytes %d < period bytes %d, possible glitch",
			  copy_bytes, free_bytes, dd->period_bytes);

	if (!copy_bytes) {
		comp_warn(dev, "dai_zephyr_copy(): nothing to copy");
		dma_reload(dd->chan->dma->z_dev, dd->chan->index, 0, 0, 0);
		return 0;
	}

I will try to remove or optimize those in next round of copier optimization.

src/audio/dai-zephyr.c
btian1: since only one time calling, add inline at least not wrong, right? :)

src/audio/module_adapter/module_adapter.c
lyakh: I wouldn't add `inline`s in these cases unless I knew that they actually improve the compiler output

src/audio/dai-zephyr.c
lyakh: no, as I said, every change should be justified, "it shouldn't hurt" doesn't look like a valid jastification to me

src/audio/module_adapter/module_adapter.c
btian1: I proved inline is necessary for static function, I converted module_adapter.c obj file to asm file, then compare it with/without inline, the different is obvious, let me added somewhere to show this.

src/audio/dai-zephyr.c
btian1: please check previous comments, compiler itself can't do inline operation, add line can reduce the entry/exit function cost.

btian1 (commented):
inline is only used for time critical function, otherwise, it will bring extra cost in code size.

src/audio/module_adapter/module_adapter.c
lyakh: hm, that's a pity. Which compiler was this?

src/audio/module_adapter/module_adapter.c
lyakh: This is a change in behaviour. In other cases functions, that you inlined, really did nothing else apart from calling a method. And if it's really only needed for IPC3, removing those lines is more difficult to verify, because CI doesn't test IPC3 any more. It seems to me that `md->state` is something like a debugging flag, used to verify that we aren't running various operations in a wrong state, nothing else. Hitting such a wrong state would be an error, and I don't think we're hitting those errors ever, so maybe that `.state` could be removed completely.

ranj063:
@btian1 can you please rebase

src/audio/module_adapter/module_adapter.c
kv2019i: @lyakh @btian1 we still have this https://github.com/thesofproject/sof/issues/5212

src/audio/dai-legacy.c
```diff
@@ -128,32 +127,13 @@ static void dai_dma_cb(void *arg, enum notify_id type, void *data)
...
-			 audio_stream_get_avail_samples(&source_c->stream) *
-				audio_stream_frame_bytes(&source_c->stream),
-			 audio_stream_get_free_samples(&sink_c->stream) *
-				audio_stream_frame_bytes(&sink_c->stream));
-	} else {
-		/* update host position (in bytes offset) for drivers */
-		dd->total_data_processed += bytes;
-	}
+	/* update host position (in bytes offset) for drivers */
+	dd->total_data_processed += bytes;
```
kv2019i, line 136: I'm not sure we can remove this, the DMA copy might fail and we'd need to handle this.

src/audio/dai-zephyr.c
```diff
@@ -317,7 +315,6 @@ dai_dma_cb(struct dai_data *dd, struct comp_dev *dev, uint32_t bytes,
 		audio_stream_consume(&dma_buf->stream, bytes);
 	}
 
-	local_buf = buffer_acquire(dd->local_buffer);
 	/* assert dma_buffer_copy succeed */
```
kv2019i, line 318: Ok, this looked suspicious at first, but seeems we indeed unncessarily release and acquire local_buf here.

src/audio/dai-zephyr.c
```diff
@@ -1537,13 +1536,6 @@ int dai_common_copy(struct dai_data *dd, struct comp_dev *dev, pcm_converter_fun
...
 	free_bytes = stat.free;
 
-	buf_c = buffer_acquire(dd->dma_buffer);
-
-	dma_fmt = audio_stream_get_frm_fmt(&buf_c->stream);
-	sampling = get_sample_bytes(dma_fmt);
-
-	buffer_release(buf_c);
-
 	/* handle module runtime unbind */
```
kv2019i, line 1539: Is it guaranteed dd->sampling matches with the "dma_buffer" fmt?
UPDATE: ack, so you add dd->samping in this PR, right, so this is good.

src/audio/dai-zephyr.c
```diff
@@ -1605,7 +1597,7 @@ int dai_common_copy(struct dai_data *dd, struct comp_dev *dev, pcm_converter_fun
 	 * in order to avoid high load spike
 	 * if FAST_MODE is enabled, then one period limitation is omitted
 	 */
-	if (!(dd->ipc_config.feature_mask & BIT(IPC4_COPIER_FAST_MODE)))
+	if (dd->non_fast_mode)
```
kv2019i, line 1600: This looks a bit surprising if this helps in performance. You need a mem dereference and check for a bit. Not sure if the modified version is faster, but I guess you've checked the generated assembly?

src/audio/eq_fir/eq_fir.c
```diff
@@ -171,36 +171,27 @@ static inline int set_fir_func(struct processing_module *mod, enum sof_ipc_frame
...
+	ipc4_update_buffer_format(sink_c, &mod->priv.cfg.base_cfg.audio_fmt);
 	buffer_release(sink_c);
-	return ret;
+
+	sourceb = list_first_item(&dev->bsource_list, struct comp_buffer, sink_list);
+	source_c = buffer_acquire(sourceb);
+	ipc4_update_buffer_format(source_c, &mod->priv.cfg.base_cfg.audio_fmt);
+	buffer_release(source_c);
+
+	return 0;
```
kv2019i, line 194: I'm losing track of what the module params should do, but I guess addition of the source fmt update is needed here. This is starting to go outside the original topic of this PR.

src/audio/module_adapter/module/cadence.c
```diff
@@ -862,8 +862,7 @@ cadence_codec_set_configuration(struct processing_module *mod, uint32_t config_i
 		return ret;
 
 	/* return if more fragments are expected or if the module is not prepared */
-	if ((pos != MODULE_CFG_FRAGMENT_LAST && pos != MODULE_CFG_FRAGMENT_SINGLE) ||
-	    md->state < MODULE_IDLE)
+	if (pos != MODULE_CFG_FRAGMENT_LAST && pos != MODULE_CFG_FRAGMENT_SINGLE)
```
kv2019i, line 865: @ranj063 @lgirdwood comment on here? I remember seeing discussions where the module common code should be checking the module state machine is followed, and removing these checks seems to go counter to that.

src/include/sof/audio/component.h
```diff
@@ -602,6 +602,11 @@ struct comp_dev {
 	/* private data - core does not touch this */
 	void *priv_data;	/**< private data */
 
+	/* store frame_fmt to avoid repeated calculation */
+	enum sof_ipc_frame in_frame_fmt;
+	enum sof_ipc_frame in_valid_fmt;
+	enum sof_ipc_frame out_frame_fmt;
+	enum sof_ipc_frame out_valid_fmt;
```
kv2019i, line 609: I don't see the calculations done on actual hot paths (in the processing calls). Does it makes sense to change SRAM to saved MCPS when the impact is only on the prepare/init paths?

kv2019i (changes requested):
Thanks @btian1 , good perf optimizations. I'm afraid the PR as a whole is a bit hard to review. It contains many straighforward commits that could go in right away, but then it also has more controversial optimizations (either because of the implementation or the gains realized) and lot of unrelated cleanups (PR title is about copier performance but PR also has changes to EQs and various other components -> another PR for these?).

src/include/sof/audio/component.h
btian1: this patch purpose is not for save mcps, it target for avoid repeated calculation in copier and mixin mixout for format, only these two modules, other modules does not have format conversion, so I did not change other modules.

src/audio/eq_fir/eq_fir.c
btian1: do you mean lose update fmt? it does update everything with ipc4_update_buffer_format and ipc4_update_buffer_format

src/audio/dai-zephyr.c
btian1: yes, here should have a bug before the change.

src/audio/dai-zephyr.c
btian1: yes, dd->sampling should keep unchanged with same dd.

src/audio/dai-zephyr.c
btian1: it definitely can speed up, new implementation is 2x faster than previous, and this is an un-changed value during running, no need calculate it per 1ms.

src/audio/dai-legacy.c
btian1: it is just perform pcm converter function instead of dma operation, dma operation happend in other place, search "dma_reload", there is error handling.

src/lib/dma.c
```diff
@@ -388,14 +385,13 @@ int dma_buffer_copy_from_no_consume(struct comp_buffer __sparse_cache *source,
...
 	/* process data */
-	ret = process(istream, 0, &sink->stream, 0, samples);
+	process(istream, 0, &sink->stream, 0, samples);
 
 	buffer_stream_writeback(sink, sink_bytes);
 
 	comp_update_buffer_produce(sink, sink_bytes);
 
-	return ret;
+	return 0;
```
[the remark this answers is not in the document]
btian1: 1. pcm converter function type is not unified, some are returning samples, some are void. 
    so, this patch mainly cover unify the return type for converter function.

2. regarding improve part, it is not big, since only saved two ret assignment.

main purpose is 1, 2 is extra benefit, there is no need to return a useless samples for a processing function.

src/audio/module_adapter/module_adapter.c
btian1: inline should only be added with:
1. only called once or very limited times.
2. timing critical calling.

for example, audio_stream_fmt_conversion this should not inline, it does not make sense, I will change it later.
it is neither timing critical calling, also called multiple times in different place.

src/audio/dai-zephyr.c
ranj063: @btian1 this is code carried over from the previous implementation when we did the optimization or conversion to module interface. If there's scope for clean up/further optimization lets do it.

cujomalainey (approved):

cujomalainey:
approval for google owned code only

src/audio/module_adapter/module_adapter.c
lyakh: > @lyakh @btian1 we still have this #5212

@kv2019i ok, that explains it, thanks! That's why xcc doesn't make decisions to inline use-once static functions...

lyakh: > inline should only be added with:
> 
>     1. only called once or very limited times.
> 
>     2. timing critical calling.

3. when inlining the function hardly makes the caller larger, when the function to be inlined is very small, e.g. like with API wrappers

```
static inline int call_init(struct context *ctx)
{
  return ctx->ops.init(ctx);
}
```

src/audio/dai-zephyr.c
kv2019i: @btian1 But IPC4_COPIER_FAST_MODE is a constant so I'd expect the compiler generate pretty efficient code also for original code. In both variants we need to dereference 'dd' and check one bit.

But if you've checked the compiler output, good for me.

src/audio/eq_fir/eq_fir.c
kv2019i: I mean setting format for "bsource_list". Original code only sets format for bsink_list, but your new code bsource_list and bsink_list. This may be the right thing to do, but a bit confusing to have this in a commit "to simplify parameters".

src/include/sof/audio/component.h
kv2019i: @btian1  Not following. By adding new fields to component.h, we increase the SRAM use for all components in the system, so if this does not save MCPS, why get rid of the calculations?

Saving in code size can help reduce SRAM consumption as well, but when augmenting widely used structs (like in component.h), we need to ensure we have a good compromise.

btian1:
> Thanks @btian1 , good perf optimizations. I'm afraid the PR as a whole is a bit hard to review. It contains many straighforward commits that could go in right away, but then it also has more controversial optimizations (either because of the implementation or the gains realized) and lot of unrelated cleanups (PR title is about copier performance but PR also has changes to EQs and various other components -> another PR for these?).

ok, let me split it to two PRs separately. this one only target for optimization.

btian1 (commented):
@lyakh @ranj063 @kv2019i  could you help continue your review?

lyakh:
> @lyakh @ranj063 @kv2019i could you help continue your review?

I think it would be good now to have #7958 merged first, e.g. because it removes the `.state` field from the module adapter. And if for some reason that PR doesn't get merged, we'll have to reconsider this one for dropping those `.state` checks

src/audio/pcm_converter/pcm_converter.c
btian1: @lyakh , I intend to remove this check, as other converter function does not have this check, what do you think？

btian1 (commented):
@kv2019i @lyakh @ranj063 @RanderWang , please continue review this PR.

src/audio/module_adapter/module_adapter.c
```diff
@@ -879,7 +879,11 @@ static inline int module_adapter_audio_stream_type_copy(struct comp_dev *dev)
 
 	/* handle special case of HOST/DAI type components */
 	if (dev->ipc_config.type == SOF_COMP_HOST || dev->ipc_config.type == SOF_COMP_DAI)
+#if CONFIG_IPC_MAJOR_3
 		return module_process_legacy(mod, NULL, 0, NULL, 0);
+#else
+		return mod->priv.ops->process_audio_stream(mod, NULL, 0, NULL, 0);
```
lgirdwood, line 885: This should be 0 cost with the staic inline in the header - I know we do check the ops != NULL, so that will add cost, but we shouldcheck all mandarory ops != NULL at init and check all optional ops != NULL at every usage.

src/audio/component.c
```diff
@@ -182,8 +182,9 @@ void comp_get_copy_limits_frame_aligned(const struct comp_buffer __sparse_cache
 
 #ifdef STREAMCOPY_HIFI3
 
-int audio_stream_copy(const struct audio_stream __sparse_cache *source, uint32_t ioffset,
-		      struct audio_stream __sparse_cache *sink, uint32_t ooffset, uint32_t samples)
+void audio_stream_copy(const struct audio_stream __sparse_cache *source, uint32_t ioffset,
```
lgirdwood, line 185: what about failures here ? i.e. some error occurs in a module during copy. modules should also be allowed to return 0 bytes.

src/audio/copier/copier.c
```diff
@@ -591,7 +591,7 @@ static int copier_process(struct processing_module *mod,
 	switch (dev->ipc_config.type) {
 	case SOF_COMP_HOST:
 		if (!cd->ipc_gtw)
-			return host_common_copy(cd->hd, dev, copier_host_dma_cb);
+			return cd->hd->copy(cd->hd, dev, copier_host_dma_cb);
```
lgirdwood, line 594: nak, we cant kill wrappers.

lgirdwood (changes requested):

src/audio/copier/copier.c
btian1: if can't kill, how about inline? inline have same effect with this change.

lgirdwood: yes - we can inline and should inline.

src/audio/module_adapter/module_adapter.c
lgirdwood: This can be optimised, the md->ops will be known at init time, we dont need to check this at process time to know which ops function to call.

The return value check can also be improved by only checking (ret < 0), we should not be filtering error numbers. Likewise modules should only return -ENODATA or -ENOSPC iff these are real errors that stop the module from processing any further data (without a reset).

```c
int module_process_legacy(struct processing_module *mod,
			  struct input_stream_buffer *input_buffers, int num_input_buffers,
			  int num_output_buffers)
{
	int ret;

	comp_dbg(dev, "module_process_legacy() start");

#if CONFIG_IPC_MAJOR_3
	if (md->state != MODULE_IDLE) {
		comp_err(dev, "module_process(): wrong state of comp_id %x, state %d",
			 dev_comp_id(dev), md->state);
		return -EPERM;
	}

	/* set state to processing */
	md->state = MODULE_PROCESSING;
#endif
	if (md->ops->process_audio_stream)
		ret = md->ops->process_audio_stream(mod, input_buffers, num_input_buffers,
						    output_buffers, num_output_buffers);
	else if (md->ops->process_raw_data)
		ret = md->ops->process_raw_data(mod, input_buffers, num_input_buffers,
						output_buffers, num_output_buffers);
	else
		ret = -EOPNOTSUPP;

	if (ret && ret != -ENOSPC && ret != -ENODATA) {
		comp_err(dev, "module_process() error %d: for comp %d",
			 ret, dev_comp_id(dev));
		return ret;
	}

	comp_dbg(dev, "module_process_legacy() done");

#if CONFIG_IPC_MAJOR_3
	/* reset state to idle */
	md->state = MODULE_IDLE;
#endif
	return 0;
}
```

and for

```c
int module_adapter_copy(struct comp_dev *dev)
{

	if (IS_PROCESSING_MODE_AUDIO_STREAM(mod))
		return module_adapter_audio_stream_type_copy(dev);

	if (IS_PROCESSING_MODE_RAW_DATA(mod))
		return module_adapter_raw_data_type_copy(dev);

	if (IS_PROCESSING_MODE_SINK_SOURCE(mod))
		return module_adapter_sink_source_copy(dev);

	comp_err(dev, "module_adapter_copy(): unknown processing_data_type");
	return -EINVAL;
}
```
This can be improved by moving the mode check to init time rather than process time.

btian1: module_init already checked:

	/*check interface, there must be one and only one of processing procedure */
	if (!interface->init || !interface->prepare ||
	    !interface->reset || !interface->free ||
	    (!!interface->process + !!interface->process_audio_stream +
	     !!interface->process_raw_data != 1)) {
		comp_err(dev, "module_init(): comp %d is missing mandatory interfaces",
			 dev_comp_id(dev));
		return -EIO;
	}

However, we still need classify which process is valid during each processing:
process
process_raw_data
process_audio_stream

to avoid IS_PROCESSING_MODE_AUDIO_STREAM(mod), I need add a flag in mod to indicate which process is valid.
then use switch: module_adapter_copy, like below:

switch (mod->process_module) {
case 1:
   process_audio_stream;
case 2:
   process_raw_data;
case 3: process;
}

something like this, any more good suggestions?

do you think it is not allowed or don't want to add more #if IPC3?
Later, I want to submit a PR to separate module_adapter.c to ipc3/ipc4, then inside #if will be totally disappear.

src/audio/host-legacy.c
```diff
@@ -234,35 +234,22 @@ void host_common_update(struct host_data *hd, struct comp_dev *dev, uint32_t byt
...
-			 audio_stream_get_avail_samples(&source->stream) *
-				audio_stream_frame_bytes(&source->stream),
-			 audio_stream_get_free_samples(&sink->stream) *
-				audio_stream_frame_bytes(&sink->stream));
-
 	buffer_release(sink);
 	buffer_release(source);
 
-	if (ret < 0)
-		return;
```
RanderWang, line 264: if ret < 0 we will not update stream position for host, but now it is not correct

src/include/sof/lib/dai-zephyr.h
```diff
@@ -168,6 +168,10 @@ struct dai_data {
 
 	/* llp slot info in memory windows */
 	struct llp_slot_info slot_info;
+	/* save current sampling for current dai device */
+	uint32_t sampling;
+	/* non fast mode, use memory to save mcps */
+	bool non_fast_mode;
```
RanderWang, line 174: I prefer to use if (!dd->fast_mode).    Rarely use non_xxxx_xxxx.

src/include/sof/audio/component.h
btian1: yes, performance/memory/code size always like a triangle, this adding is reasonable and acceptable(one int), since it is not changing at runtime.

src/include/sof/lib/dai-zephyr.h
btian1: it will add extra cost from performance perspective, since ! need at least one more instruction operation.

src/audio/host-legacy.c
btian1: there is no chance to get ret < 0 case, so I removed the return.

src/audio/pcm_converter/pcm_converter.c
```diff
@@ -31,12 +31,6 @@ int pcm_convert_as_linear(const struct audio_stream __sparse_cache *source, uint
 	int chunk;
 	int N1, N2;
 
-	/* assert enough avail/free samples in source and sink buffer */
-	if (audio_stream_get_avail_samples(source) < samples + ioffset)
-		return -EINVAL;
-	if (audio_stream_get_free_samples(sink) < samples + ooffset)
-		return -EINVAL;
```
lyakh, line 38: an example of where these PCM processing functions are called in https://github.com/thesofproject/sof/blob/8f6018057b64441f26de6d509ac16514d7347f12/src/lib/dma.c#L339 and it seems like there the check isn't complete - it gets available data from the source, but doesn't check whether the sink has sufficient free space. Not sure if this can ever become a problem (e.g. with multiple sources writing data to a single sink), but if that does have to be checked, maybe it has to be checked uniformly for all such calls.

src/include/sof/audio/component.h
kv2019i: @btian1 Ack but I'm arguing that this tradeoff is not good if the code saving is not on the hot path (in a processing function) but the memory cost affects everything.

src/include/sof/lib/dai-zephyr.h
kv2019i: @btian1 compiler should produce equal-speed code for both variants

src/audio/pcm_converter/pcm_converter.c
kv2019i: @btian1 Can you comment?

lgirdwood (commented):
@btian1 lets merge inline Kconfig 1st then refactor this on top with ops accessors APIs and resubmit.

src/include/sof/lib/dai-zephyr.h
btian1: that's not possible, it definitely need a ! operation to get the desired results.

src/audio/pcm_converter/pcm_converter.c
btian1: I like "uniform", sof did a lot of check in many functions, hard to say those check are all must, especially those on critical timing path.

based on current sof architecture(2X period buffer size),  source is checked by samples( as long as it is not zero).

for sink, I am not sure whether there is a case with multi src --> single sink, if there have, single sink buffer size still need 2X of total src, I mean, if sink have no space, there is pipeline overrun.

we can add sink check if you still feel not safe enough.

src/audio/dai-zephyr.c
btian1: this will not be needed after #8033 merge

src/include/sof/lib/dai-zephyr.h
lyakh: > @btian1 compiler should produce equal-speed code for both variants

@kv2019i yeah, right, in principle. Only if one of the branches is taken much more often, but Zephyr has `likely()` and `unlikely()` for that

src/audio/pcm_converter/pcm_converter.c
btian1: seems I reverted https://github.com/thesofproject/sof/pull/3251 in this patch, @ktrzcinx , may I know any reason to add those checkers in patch #3251

btian1: @lgirdwood @lyakh @mmaka1 , if you think keep error handling is must for converter, I can remove this patch, and find other way to do optimization.

ktrzcinx: @btian1 before #3251 there already was this check as an `assert()`. The change was related with some bug which ends up with SOF crash so I've changed `assert()` to `if() return` to gain human readable log about root cause of the issue in a similar cases in the future.

btian1: @ktrzcinx ， this part can be added back, I mean the whole #3251, in this patch, I almost reverted all.

src/audio/module_adapter/module_adapter.c
lgirdwood: Since we check for ops != NULL in init then I would expect

```c

// header
static inline int module_process(mod, blah, blah) {
     return mod->priv.ops->process_audio_stream(mod, NULL, 0, NULL, 0);
}

// C file
#if CONFIG_IPC_MAJOR_3
		return module_process_legacy(mod, NULL, 0, NULL, 0);
#else
		return module_process(mod, NULL, 0, NULL, 0);
#endif
```

lgirdwood (commented):
@btian1 some of these patches can be merged now,  you could split the PR so the patches with no comments can be merged.
I need to see the inline usage for merging the 1st patch as per my comment.

keqiaozhang:
SOFCI TEST

src/include/sof/audio/component.h
```diff
@@ -559,6 +559,16 @@ struct comp_ipc_config {
 #endif
 };
 
+/**
+ * \enum module_process_type
+ * \brief Module-proces type
+ */
+enum module_process_type {
+	MODULE_PROCESS_STREAM,
```
lgirdwood, line 567: Can we have some inline comments describing each type with examples and how to categorize. Can be in another PR .

lgirdwood (approved):

src/include/sof/audio/component.h
btian1: yes, I removed this patch, only keep 3 patches for this PR, and will submit new PR later.

lyakh (approved):

PR #7910, opened 2023-07-10
ace_version: change to correct version

fkwasowi:
For ACE we expect HW_CFG_VERSION 0x0.

### DIFF
--- a/src/platform/intel/ace/include/ace/version.h
+++ b/src/platform/intel/ace/include/ace/version.h
@@ -8,16 +8,6 @@
 #ifndef __ACE_VERSION_H__
 #define __ACE_VERSION_H__
 
-#define ACE_VERSION_1_5		0x10500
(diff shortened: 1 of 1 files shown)


pblaszko (approved):

dnikodem (approved):

abonislawski (changes requested):
please change only HW_CFG_VERSION and do not remove ACE_VERSION which is useful (like for check CAVS_VERSION > CAVS_VERSION_2_0 instead of several _if version_)

lgirdwood (commented):
@fkwasowi I think the long term direction is correct here, we dont need these in "application code". These macros should come from Zephyr and I'm not sure if they are today ?

abonislawski (approved):
Good point, I forgot about zephyr where we already have ACE 1_5/2_0 kconfigs, just not a generic ACE_VERSION if I remember correctly

PR #7912, opened 2023-07-10
audio: mixer: Set correct number of sources

dbaluta:
After commit 867ada28f we only allocate 1 buffer for mixer even if there could be 2 input sources.

So, set correct number of sources for mixer.

Fixes 867ada28f ("audio: module_adapter: Allocate max number of input/output buffers")
Signed-off-by: Daniel Baluta <daniel.baluta@nxp.com>

### DIFF
--- a/src/audio/mixer/mixer.c
+++ b/src/audio/mixer/mixer.c
@@ -56,6 +56,8 @@ static int mixer_init(struct processing_module *mod)
 	mod_data->private = md;
 	mod->verify_params_flags = BUFF_PARAMS_CHANNELS;
 	mod->no_pause = true;
+	mod->max_sources = MIXER_MAX_SOURCES;
+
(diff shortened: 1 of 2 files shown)


lgirdwood (commented):
@dbaluta any reason to make 2 the max ?
@ranj063 any comment ?

ranj063:
> @dbaluta any reason to make 2 the max ? @ranj063 any comment ?

2 is th most we've tested with IPC3, Liam. I'd keep it at 2 for now and if needed increase it later.

ranj063 (approved):

btian1 (approved):

dbaluta:
> @dbaluta any reason to make 2 the max ?

we never used more than 2 sources.

iuliana-prodan (approved):

PR #7915, opened 2023-07-10
volume: Bail out if there is no configuration

jsarha:
It is possible to fabricate a message for FW that initializes volume module without a configuration. Make a safe guard again crashing from that.

Fixes a similar problem with volume module, as this PR fixes for SRC module: https://github.com/thesofproject/sof/pull/7830

### DIFF
--- a/src/audio/module_adapter/module/volume/volume.c
+++ b/src/audio/module_adapter/module/volume/volume.c
@@ -407,6 +407,12 @@ static int volume_init(struct processing_module *mod)
 	const size_t vol_size = sizeof(int32_t) * SOF_IPC_MAX_CHANNELS * 4;
 	int i;
 
+	if (!vol || cfg->size != sizeof(*vol)) {
+		comp_err(dev, "volume_init(): No configuration data or bad data size %u",
+			 cfg->size);
+		return -EINVAL;
+	}
+
 	cd = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(struct vol_data));
 	if (!cd)
 		return -ENOMEM;


src/ipc/ipc3/helper.c
```diff
@@ -146,8 +146,8 @@ static const struct comp_driver *get_drv(struct sof_ipc_comp *comp)
 
 out:
 	if (drv)
-		tr_dbg(&comp_tr, "get_drv(), found driver type %d, uuid %pU",
-		       drv->type, drv->tctx->uuid_p);
+		tr_warn(&comp_tr, "get_drv(), found driver type %d msg %d, uuid %pU",
```
btian1, line 149: this is the right execution code, drv is expected to be true, so this print will always happen, I would not suggest to print this or change it warning, keep as tr_dbg is better.

If delete this, may also make sense, due to if error happens, line 137 - 143 would print out message.

jsarha: Oh, that an unrelated debugging change that slipped in. I'll get rid of it.

lgirdwood (approved):

cujomalainey (approved):

cujomalainey (commented):
Actually can you add that `cfg->size == sizeof(*vol)`?

kv2019i (approved):

lgirdwood:
@jsarha ping, can you adress comment and check CI. Lot of red. Thanks

src/audio/module_adapter/module/volume/volume.c
```diff
@@ -407,6 +407,11 @@ static int volume_init(struct processing_module *mod)
 	const size_t vol_size = sizeof(int32_t) * SOF_IPC_MAX_CHANNELS * 4;
 	int i;
 
+	if (!vol) {
+		comp_err(dev, "volume_init(): No configuration data");
+		return -EINVAL
```
cujomalainey, line 412: missing `;`

jsarha: Oops, now fixed.

jsarha:
> Actually can you add that `cfg->size == sizeof(*vol)`?

Added.

lgirdwood (approved):

Issue #7801, opened 2023-06-14
[FEATURE][SMART_AMP] Revamp smart_amp component for adaptive solutions


johnylin76:
**Is your feature request related to a problem? Please describe.**
The current smart_amp component implementation [source code](https://github.com/thesofproject/sof/tree/main/src/audio/smart_amp) is named in a generic matter. It contains smart_amp.c as the role of common adapter (handling component API), and smart_amp_[generic|maxim_dsm].c for generic and Maxim DSM solution respectively. However, the implementation inside doesn't perform in the same manner. The design is deeply bound to Maxim DSM solution, which is not surprising since Maxim is the initial author and there is no other solution supported for a long time.

**Describe the solution you'd like**
Revamp smart_amp to be in a generic matter, and adaptive to other solutions.
The expected framework would be alike module_adapter.

**Describe alternatives you've considered**
1. ask the new solution provider to create another smart_amp component and keep the present one untouched.
     - cons: will produce code duplication
2. implement the new smart_amp solution by module_adapter. Given that there is the on-going project to convert existing components into modules with module_adapter
     - cons: not sure about the feasibility since amart_amp is multi-input (2 source, 1 sink) component

johnylin76:
Hi @lgirdwood ,
This issue needs your input especially for the 2nd point of alternatives.

If the on-going plan in mainstream is to migrate the processing components to module_adapter form, would you suggest to do smart_amp revamp task for adapting new smart_amp solution?
Does module_adapter support multiple source or sink already?

johnylin76:
+ @cujomalainey for viz

cujomalainey:
@ranj063 I believe I heard your name come up last time I talked about this, feel free to unsubscribe if I am having a mental cache miss.

lgirdwood:
@johnylin76 sorry for delay, yeah I agree - a generic smart amp layer would be the best option. We already have the "dummy" smart amp logic which could be modified to become a generic layer. Today the dummy smart amp just copies data, so its probably a good starting point. Also adding @lyakh @aiChaoSONG @mengdonglin. Think it may be too tight to squeeze this in for v2.7, v2.8 probably more resonable.

johnylin76:
@lgirdwood Thanks for the advice.
As for the dummy smart amp mentioned above, you mean [smart_amp_test](https://github.com/thesofproject/sof/blob/main/src/samples/audio/smart_amp_test_ipc4.c) implementation, is that correct?
[...]

PR #7920, opened 2023-07-11
aw_mt8188 v0.01

awqyh:
Add SKTune code


dbaluta:
Can you please add a short introduction about what is aW and what is supposed to do. Like a cover letter.

gkbldcig:

lgirdwood:
test this please

src/audio/smart_amp_aw/aw_smart_amp.c
```diff
@@ -0,0 +1,557 @@
...
+}
+
+static int smart_amp_comp_reset(struct comp_dev *dev)
+{
+	struct smart_amp_data *smartpa_data = comp_get_drvdata(dev);
+
+	comp_dbg(dev, "[Awinic] smart_amp_reset()");
+
+	smartpa_data->in_channels = 0;
+	smartpa_data->out_channels = 0;
```
cujomalainey, line 428: Does the fb buffer need to be reset even on pause? if not then the reset should go here

src/audio/smart_amp_aw/aw_smart_amp.c
```diff
@@ -0,0 +1,557 @@
...
+	buf_c = buffer_acquire(smartpa_data->sink_buf); 
+	smartpa_data->out_channels = audio_stream_get_channels(&buf_c->stream);
+	buffer_release(buf_c);
+
+	source_c = buffer_acquire(smartpa_data->source_buf);
+
+	if (smartpa_data->feedback_buf) {
+		buf_c = buffer_acquire(smartpa_data->feedback_buf);
+
+		audio_stream_set_channels(&buf_c->stream, smartpa_data->config.feedback_channels);
```
cujomalainey, line 477: @lgirdwood won't this get reset by someone requesting after this event?

src/audio/smart_amp_aw/aw_smart_amp.c
```diff
@@ -0,0 +1,557 @@
...
+			/* copy IV data frome stream to IV buf */
+			smart_amp_fb_data_prepare(smartpa_data->algo_handle, dev, &feedback_buf->stream,
+				     avail_feedback_frames);
+
+			comp_update_buffer_consume(feedback_buf, feedback_bytes);
+		}
+
+		buffer_release(feedback_buf);
+	} else {
+		/*IV buffer clear*/
```
cujomalainey, line 391: delete the `else` if you are not going to use it

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
```diff
@@ -0,0 +1,758 @@
...
+	return NULL;
+
+}
+
+
+int smart_amp_init(sktune_t *      sktune, struct comp_dev *dev)
+{
+	int ret;
+
+	ret = sktune_api_init(sktune->handle);
```
cujomalainey, line 538: do you return unix style error codes?

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
```diff
@@ -0,0 +1,758 @@
...
+
+	sum_mem_size += (int)algo_size;
+
+	comp_dbg(dev, "[Awinic] module:%p (%d bytes used)", sktune, sum_mem_size);
+
+	return sktune;
+
+err:
+	smart_amp_sktune_free(sktune);
+	rfree(sktune);
```
cujomalainey, line 528: double free here

src/audio/smart_amp_aw/aw_smart_amp.c
```diff
@@ -0,0 +1,557 @@
...
+
+static void smart_amp_comp_free(struct comp_dev *dev)
+{
+	struct smart_amp_data *smartpa_data = comp_get_drvdata(dev);
+
+	comp_dbg(dev, "[Awinic] smart_amp_free()");
+
+	smart_amp_deinit(smartpa_data->algo_handle, dev);
+
+	smartpa_data->algo_handle = NULL;
```
cujomalainey, line 290: unneeded since this freed on the next line

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
```diff
@@ -0,0 +1,758 @@
...
+#include <rtos/bit.h>
+#include <sof/common.h>
+#include <ipc/channel_map.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <sof/audio/smart_amp/aw_smart_amp.h>
+#include "aw_api_public.h"
+
+
+#define SOF_SMART_AMP_RX_ENABLE   	(0x10013D11)
```
cujomalainey, line 25: these are not SOF IDs, these are AWINIC IDs

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
```diff
@@ -0,0 +1,758 @@
...
+
+static int smartpa_amp_get_msg_parser(sktune_t *sktune, struct comp_dev *dev, char *data_buf, uint32_t len, int msg_id)
+{
+	uint32_t actual_data_len = len;
+	char *data_ptr = data_buf;
+	int ret;
+
+	if (sktune->sub_msg_info[msg_id].status != AW_MESG_READY) {
+		comp_err(dev, "[Awinic] set msg cmd no read, please write cmd first");
+		return -1;
```
cujomalainey, line 54: use a proper error code please, same goes for the rest of the file

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
```diff
@@ -0,0 +1,758 @@
...
+			 buf_size = sizeof(uint32_t);
+		} else {
+			comp_err(dev, "[Awinic] bad params size  %d ", max_size);
+			ret = -1;
+		}
+	} break;
+	case SOF_SMART_AMP_RX_SKT_PARAMS : {
+		ret = sktune_api_get_params(sktune->handle, (char *)data_buf, max_size);
+		if (ret < 0) {
+			comp_err(dev, "[Awinic] get params failed ");
```
cujomalainey, line 103: return the error? this whole function seems to fail silently to userspace on api failures

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
```diff
@@ -0,0 +1,758 @@
...
+			comp_err(dev, "[Awinic] get unsupport params ID  %d ", params_id);
+			ret = -1;
+	} break;
+
+	}
+
+	if (ret < 0) {
+		cdata->data->abi = SOF_ABI_VERSION;
+		cdata->data->size = 0;
+		return -1;
```
cujomalainey, line 249: `return ret;`

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
```diff
@@ -0,0 +1,758 @@
...
+		sktune->sub_msg_info[msg_id].status = AW_MESG_READY;
+		sktune->sub_msg_info[msg_id].opcode_id = msg_hdr->opcode_id;
+		comp_info(dev, "[Awinic] set msg opcode[0x%x] done", msg_hdr->opcode_id);
+		return 0;
+	}
+
+	if (msg_hdr->type == DSP_MSG_TYPE_DATA) {
+		actual_data_len = len - sizeof(dsp_msg_hdr_t);
+		data_ptr = (char *)(data_buf + sizeof(dsp_msg_hdr_t));
+		ret = sktune_api_set_data(sktune->handle, msg_hdr->opcode_id, data_ptr, actual_data_len);
```
cujomalainey, line 286: make sure this call is hardened against user abuse

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
```diff
@@ -0,0 +1,758 @@
...
+
+	input16 = audio_stream_get_rptr(stream);
+	input32 = audio_stream_get_rptr(stream);
+	output16 = (int16_t *)buf;
+	output32 = (int32_t *)buf;
+
+	switch (audio_stream_get_frm_fmt(stream)) {
+	case SOF_IPC_FRAME_S16_LE:
+		for (idx = 0 ; idx < sample_num ; idx++) {
+				input16 = audio_stream_read_frag_s16(stream, idx);
```
cujomalainey, line 616: @singalsu can you comment on the proper way to read the buffer? this looks like the legacy method

src/include/sof/audio/smart_amp/aw_smart_amp.h
```diff
@@ -0,0 +1,128 @@
...
+#include <sof/audio/component.h>
+
+/* Awinic Maximum number of channels for algorithm in */
+#define SMART_AMP_FF_MAX_CH_NUM		2
+/* Awinic Maximum number of channels for algorithm out */
+#define SMART_AMP_FF_OUT_MAX_CH_NUM	4
+/* Awinic Maximum number of channels for feedback  */
+#define SMART_AMP_FB_MAX_CH_NUM		4
+
+#define SMART_AMP_FRM_SZ	48 /* samples per 1ms */
```
cujomalainey, line 21: 48 is samples per ms, not frame size

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
```diff
@@ -0,0 +1,758 @@
...
+		comp_warn(dev, "[Awinic] ff frame size zero warning.");
+		return 0;
+	}
+
+	if (frames > SMART_AMP_FF_BUF_DB_SZ) {
+		comp_err(dev, "[Awinic] ff frame size overflow  : %d", frames);
+		return -1;
+	}
+
+	num_ch = MIN(audio_stream_get_channels(source), SMART_AMP_FB_MAX_CH_NUM);
```
cujomalainey, line 685: `SMART_AMP_FF_MAX_CH_NUM`

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
```diff
@@ -0,0 +1,758 @@
...
+
+	ret = smart_amp_get_buffer(sktune->frame_in.data_ptr, frames, source, num_ch);
```
cujomalainey, line 687: if the only thing that varies here is the data_ptr then i feel like this and the above function can be generalized together.

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
```diff
@@ -0,0 +1,758 @@
...
+	int32_t *input32, *output32;
+
+	input16 = (int16_t *)buf;
+	input32 = (int32_t *)buf;
+	output16 = audio_stream_get_wptr(stream);
+	output32 = audio_stream_get_wptr(stream);
+
+	switch (audio_stream_get_frm_fmt(stream)) {
+	case SOF_IPC_FRAME_S16_LE:
+		for (idx = 0 ; idx < sample_num ; idx++) {
```
cujomalainey, line 716: @singalsu also here

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
```diff
@@ -0,0 +1,758 @@
...
+			     uint32_t frames)
+{
+	int num_ch, ret;
+	
+	if (frames == 0) {
+		comp_warn(dev, "[Awinic] ff frame size zero warning.");
+		return 0;
+	}
+
+	if (frames > SMART_AMP_FF_BUF_DB_SZ) {
```
cujomalainey, line 680: some boards are doing up to 10ms, i recommend you consider making your code more versatile.

cujomalainey (changes requested):

installer/config.mk
```diff
@@ -0,0 +1,37 @@
+# To customize the installation, copy this file to config.mk and edit
```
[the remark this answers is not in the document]
awqyh: image
We configured it according to the step4 "Copy the commented installer/sample-config.mk to installer/config.mk, then select a list of platforms and provide an optional target hostname in the latter file. "

src/audio/CMakeLists.txt
```diff
@@ -79,6 +79,10 @@ if((NOT CONFIG_LIBRARY) OR CONFIG_LIBRARY_STATIC)
 		sof_add_static_library(dsm ./smart_amp/dsm_lib/libdsm.a)
 		add_subdirectory(smart_amp)
 	endif()
+	if(CONFIG_AW_SKTUNE)
+		sof_add_static_library(awinic ./smart_amp_aw/skt_lib/libawinic.a)
```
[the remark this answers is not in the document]
awqyh: Do you mean the "sof/third_party" folder?

src/audio/Kconfig
```diff
@@ -301,6 +301,15 @@ config MAXIM_DSM
 	  feedback processing block running on capture thread with
 	  amp current and voltage feedback.
 
+config AW_SKTUNE
+	bool "Awinic SKTune component"
+	select COMP_BLOB
+	default n
+	help
+	  Select for Awinic SKTune component.
+	  This component protect the speaker from overheating and
```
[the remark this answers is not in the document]
awqyh: What do you mean, please? Should we change "Awinic SKTune component" to "on awinic supported chips"

src/audio/Kconfig
```diff
@@ -301,6 +301,15 @@ config MAXIM_DSM
...
+	  excursion violation.1
```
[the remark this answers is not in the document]
awqyh: Do you mean we wrote an extra 1?

src/audio/smart_amp_aw/aw_api/inc/aw_api_public.h
```diff
@@ -0,0 +1,40 @@
...
+// int32_t aw_mtk_dsp_set_b_interleave(void *env_ptr,uint8_t b_interleave);
+int32_t sktune_api_set_params(void *env_ptr, char *params, uint32_t params_size);
+int32_t sktune_api_get_params(void *env_ptr, char *params, uint32_t params_size);
+// int32_t aw_mtk_dsp_set_module_param(int32_t module_id, void *env_ptr, void *params, uint32_t params_size);
+// int32_t aw_mtk_dsp_get_module_param(int32_t module_id, void *env_ptr, void *params, uint32_t params_size);
+
+int32_t sktune_api_get_vmax(void *env_ptr, int32_t* vmax, int32_t channel);
+int32_t sktune_api_set_vmax(void *env_ptr, int32_t vmax, int32_t channel);
+// int32_t aw_mtk_dsp_set_xmax(void *env_ptr, int32_t xmax, int32_t channel);
+// int32_t aw_mtk_dsp_get_xmax(void *env_ptr, int32_t* xmax, int32_t channel);
```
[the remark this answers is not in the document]
awqyh: These are functions that other projects use, we will delete them.

src/audio/smart_amp_aw/aw_api/inc/aw_api_public.h
```diff
@@ -0,0 +1,40 @@
+#ifndef __AW_API_PUBLIC_H__
```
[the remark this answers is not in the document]
awqyh: Ok, we will add "# SPDX-License-Identifier: BSD-3-Clause" in aw_api_public.h

src/include/sof/audio/smart_amp/aw_smart_amp.h
awqyh: Ok. Is there anything we need to change?

src/audio/CMakeLists.txt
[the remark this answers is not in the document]
cujomalainey: yes that is what I meant, thanks

installer/config.mk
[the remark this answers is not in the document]
cujomalainey: I don't think it needs to be included int he commit.

src/audio/Kconfig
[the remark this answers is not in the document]
cujomalainey: Just want that added to description so that its clear this doesn't protect all speakers from overheating

src/audio/Kconfig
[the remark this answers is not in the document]
cujomalainey: yes

src/include/sof/audio/smart_amp/aw_smart_amp.h
cujomalainey: Should probably rename your define so it makes sense. Also that config is only true in 48k which the system may not always use. Recommend you use some of the audio_stream helpers for this math

lyakh:
I'm wondering if we should try and make this use the module adapter API from the beginning?

src/audio/smart_amp_aw/aw_smart_amp.c
```diff
@@ -0,0 +1,557 @@
...
+	bs = sad->config.size;
+
+	comp_dbg(dev, "smart_amp_get_config(), actual blob size = %u, expected blob size = %u",
+		 bs, sizeof(struct sof_smart_amp_config));
+
+	if (bs == 0 || bs > size)
+		return -EINVAL;
+
+	ret = memcpy_s(cdata->data->data, size, &sad->config, bs);
+	assert(!ret);
```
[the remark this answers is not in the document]
cujomalainey: agreed so its probably actually not likely to be hit at all and probably is not needed

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
```diff
@@ -0,0 +1,758 @@
...
+sktune_t* smart_amp_sktune_alloc(struct comp_dev *dev)
+{
+	uint32_t sum_mem_size = 0;
+	unsigned int algo_size;
+	uint32_t size;
+	sktune_t* sktune = NULL;
+
+
+	/*memory allocation for sktune*/
+	sktune = rballoc(0, SOF_MEM_CAPS_RAM, sizeof(sktune_t));
```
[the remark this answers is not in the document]
cujomalainey: if this structure has large buffers embedded in it then rballoc is fine

src/audio/smart_amp_aw/aw_smart_amp.c
```diff
@@ -0,0 +1,557 @@
...
+		comp_err(dev, "[Awinic] samrt amp init failed!");
+		goto error;
+	}
+
+	dev->state = COMP_STATE_READY;
+
+	return dev;
+
+error:
+	rfree(amp_data->algo_handle);
```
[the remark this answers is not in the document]
johnylin76: Hi Awinic folks,

An important note for dynamic memory allocation in SOF 3P mpdule design: you should always keep the transparency for dynamic memory occupancy. That is, NEVER allocate dynamic memory inside the library directly through stdio functions e.g. malloc. There are two alternative ways:

1. explicitly define the buffer you'll need inside the library, and do the allocation in external domain on module initialization (same for de-allocation on module release)

    Example code for Maxim smart_amp, which defines all struct objects in external domain, and do memory allocation/de-allocation at once:
    `smart_amp_alloc_memory()`
    https://github.com/thesofproject/sof/blob/main/src/audio/smart_amp/smart_amp.c#L76

    `smart_amp_free_memory()`
    https://github.com/thesofproject/sof/blob/main/src/audio/smart_amp/smart_amp.c#L47

    Example code for Waves in module_adapter, it preserves the struct definition, an API function is built to 
query the memory size needed inside from external domain `MaxxEffect_GetMessageMaxSize()`
https://github.com/thesofproject/sof/blob/main/src/include/sof/audio/MaxxEffect/Control/RPC/MaxxEffect_RPC_Server.h#L26

    Obtain the required size and allocate the memory  in external domain,
https://github.com/thesofproject/sof/blob/main/src/audio/module_adapter/module/waves.c#L194

3. if you really need the inline memory allocation (allocate the memory in internal domain), you will need to create the specified API to expose allocate/free to external domain, and use them to allocate memory in internal

    Example code for DTS in module_adapter:
    `dts_effect_allocate_codec_memory()`
https://github.com/thesofproject/sof/blob/main/src/audio/module_adapter/module/dts.c#L19

    `dts_effect_allocate_free_memory()`
https://github.com/thesofproject/sof/blob/main/src/audio/module_adapter/module/dts.c#L38

    The implementation requires an external structure for monitoring, e.g. `struct module_memory`. By this method the external domain will have counterpart of the set of allocated buffer pointers
https://github.com/thesofproject/sof/blob/main/src/include/sof/audio/module_adapter/module/generic.h#L131

Last but not least, for data buffer allocation, always use the following function to make sure the correct memory region is applied:  
```
memory = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, size);
```

Thanks

johnylin76 (commented):
@lgirdwood @cujomalainey 

The context about this draft commit:

Based on our current work for https://github.com/thesofproject/sof/issues/7801, Awinic, a new 3P smart_amp solution, will eventually be implemented as an adaptive solution behind the smart_amp generic layer after revamp, and put to mainstream.

That is, the current implementation won't be published this way. Instead, it was introduced to make the development work not blocked by  https://github.com/thesofproject/sof/issues/7801

src/audio/smart_amp_aw/aw_smart_amp.c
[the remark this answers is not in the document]
lyakh: > for data buffer allocation, always use the following function to make sure the correct memory region is applied:
> 
> ```
> memory = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, size);
> ```

I think for buffer allocations `rballoc()` or `rballoc_align()` should be used

src/include/sof/audio/smart_amp/aw_smart_amp.h
awqyh: Ok, we'll modify the algorithms once they're stable

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
awqyh: Once the algorithm is stable, we'll optimize the code

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
awqyh: We will change the buffer size to 10ms for the time being

src/audio/smart_amp_aw/aw_api/inc/aw_api_public.h
```diff
@@ -0,0 +1,40 @@
...
+#define __AW_API_PUBLIC_H__
+
+// int32_t aw_mtk_dsp_get_version(char* buf);
+uint64_t sktune_api_get_size(void);
+// uint64_t aw_mtk_dsp_get_params_size(void);
+int32_t sktune_api_init(void *env_ptr);
```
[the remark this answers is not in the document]
awqyh: Once the algorithm is stable, we'll optimize the code

src/audio/smart_amp_aw/aw_smart_amp.c
```diff
@@ -0,0 +1,557 @@
...
+				   struct sof_ipc_ctrl_data *cdata, int size)
+{
+	int ret = 0;
+
+	comp_dbg(dev, "[Awinic] smart_amp_ctrl_get_data() size: %d", size);
+
+	switch (cdata->cmd) {
+	case SOF_CTRL_CMD_BINARY:
+		ret = smart_amp_ctrl_get_bin_data(dev, cdata, size);
+		break;
```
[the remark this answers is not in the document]
awqyh: We're just trying to avoid size errors

src/audio/smart_amp_aw/aw_smart_amp.c
```diff
@@ -0,0 +1,557 @@
...
+}
+
+static int smart_amp_comp_trigger(struct comp_dev *dev, int cmd)
+{
+	struct smart_amp_data *smartpa_data = comp_get_drvdata(dev);
+	int ret = 0;
+
+	comp_dbg(dev, "[Awinic] smart_amp_trigger(), command = %u", cmd);
+
+	ret = comp_set_state(dev, cmd);
```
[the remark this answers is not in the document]
awqyh: When we allocate memory, the data is already cleared to zero, so can we not use COMP_STATE_INIT?

src/audio/smart_amp_aw/aw_smart_amp.c
```diff
@@ -0,0 +1,557 @@
...
+		break;
+	default:
+		break;
+	}
+
+	return ret;
+}
+
+
+
```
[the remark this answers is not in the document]
awqyh: OK

src/arch/xtensa/configs/mt8188_defconfig
```diff
@@ -13,3 +13,4 @@ CONFIG_COMP_SWITCH=n
 CONFIG_COMP_KPB=n
 CONFIG_COMP_SEL=n
 CONFIG_COMP_ASRC=n
+CONFIG_AW_SKTUNE=y
```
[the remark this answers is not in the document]
awqyh: Now our modification can run normally. Do we not need to change it any more?

src/audio/Kconfig
```diff
@@ -301,6 +301,15 @@ config MAXIM_DSM
 	  feedback processing block running on capture thread with
 	  amp current and voltage feedback.
 
+config AW_SKTUNE
```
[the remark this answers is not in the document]
awqyh: OK, THKS

src/audio/smart_amp_aw/aw_smart_amp.c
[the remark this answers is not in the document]
lyakh: sorry, not sure I understand your question. This is your `.trigger()` method, it's called for various events - start, stop, pause, etc. It can happen that an invalid state transition would be requested here, so you probably want to check that?

johnylin76:
> I'm wondering if we should try and make this use the module adapter API from the beginning?

@lyakh That's also one of my concerns, which had been mentioned in https://github.com/thesofproject/sof/issues/7801#issue-1756658279

First thing to mention is the context in https://github.com/thesofproject/sof/pull/7920#pullrequestreview-1531640161. That is to say, the upstream version won't be as is, a new component in src/audio/. Instead, it will be a part in the modular smart amp design, or a module in module_adapter.

The latter is the ultimate goal with no doubt. But in consideration of project timeline, the former should be more practicable for now since more common logic can be shared among smart_amp solutions. We are working towards it

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
singalsu: Is this copy from circular to linear buffer? There's a function available for that.

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
singalsu: audio_stream_copy_from_linear() should work here

tools/topology/CMakeLists.txt
```diff
@@ -61,4 +61,4 @@ add_custom_target(topologies ALL)
 add_dependencies(topologies topologies1)
 
 add_subdirectory(topology1)
-add_subdirectory(topology2)
+#add_subdirectory(topology2)
```
lyakh, line 64: a left-over

src/audio/smart_amp_aw/aw_smart_amp_wapper.c
```diff
@@ -749,10 +736,8 @@ int smart_amp_process(sktune_t *       sktune, struct comp_dev *dev,
 
 
 	ret = smart_amp_put_buffer(sktune->frame_in.data_ptr,
-			   frames, sink,
-			   MIN(num_ch_out, SMART_AMP_FF_OUT_MAX_CH_NUM));
+				frames, sink,
+				MIN(num_ch_out, SMART_AMP_FF_OUT_MAX_CH_NUM));
```
lyakh, line 740: I understand you'll merge the two commits for the final version?

src/audio/smart_amp_aw/aw_smart_amp.c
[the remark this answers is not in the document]
johnylin76: ah yes @lyakh is correct, please use the following command instead (for allocating data buffers):
```
memory = rballoc(0, SOF_MEM_CAPS_RAM, size);
```
Please note the memory allocated from rballoc() won't be preset to all-zeros since most use cases for data buffers don't need that. If needed, please use bzero() after that.

PR #7921, opened 2023-07-11
drivers: mtk: Add few fixes

wenliangwu:
This pull request contains two commits: one adds a missing parameter in the debug log and another updates the MEMIF configuration format. 

Initially, MEMIF format was configured based on DAI configuration. However, DAI DMA(MEMIF) can support a different bit width than that of BE DAI for MediaTek AFE. Hence, it is better to obtain BE frame format from DAI_CONFIG and MEMIF format from the configuration of DAI component which should be identical to the bitwidth in pipeline.

### DIFF
--- a/src/drivers/mediatek/afe/afe-memif.c
+++ b/src/drivers/mediatek/afe/afe-memif.c
@@ -294,6 +294,20 @@ static int memif_set_config(struct dma_chan_data *channel, struct dma_sg_config 
 	ret = afe_dai_get_config(memif->afe, dai_id, &memif->channel, &memif->rate, &memif->format);
 	if (ret < 0)
 		return ret;
+
+	/* memif format should follow DAI component, not dai hw configuration */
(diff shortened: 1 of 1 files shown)


dbaluta:
` Afe patch 0703` is not really a good descriptive comment for a PR.

For the future please add something that could use reviewers to better understand what's all about :)

lgirdwood (approved):

yaochunhung (approved):

wenliangwu:
> ` Afe patch 0703` is not really a good descriptive comment for a PR.
> 
> For next PR add something that could use reviewers to better understand what's all about :)

At first, I thought the name was not important. However, I will try to select a more appropriate and specific name in the future.
Thank you.

kv2019i:
Checkpatch fail is a non-issue, sof-docs build issue is present in mainline. Proceeding with merge.

PR #7925, opened 2023-07-12
buffer: fix DSP panic caused by buffer_attach

wenliangwu:
This is the PR for Issues#7916.
Cache operation at **_buffer_attach_** results in DSP panic problem on MediaTek platform, but cache invalidation is required for INTEL platform to solve the prefetch issue.
Add **#if CONFIG_INTEL** to make the line specific to INTEL platform.

### DIFF
--- a/src/audio/buffer.c
+++ b/src/audio/buffer.c
@@ -312,8 +312,13 @@ void buffer_attach(struct comp_buffer *buffer, struct list_item *head, int dir)
 	list_item_prepend(list, head);
 	if (further_buffers_exist)
 		dcache_invalidate_region(needs_sync, sizeof(struct list_item));
-	/* no dirty cache lines exist for this buffer yet, no need to write back */
+#if CONFIG_INTEL
+	/*
+	 * Until now the buffer object wasn't in cache, but uncached access to it could have
+	 * triggered a cache prefetch. Drop that cache line to avoid using stale data in it.
+	 */
 	dcache_invalidate_region(uncache_to_cache(list), sizeof(*list));
+#endif
 }
 
 /*


src/audio/buffer.c
```diff
@@ -313,7 +313,7 @@ void buffer_attach(struct comp_buffer *buffer, struct list_item *head, int dir)
 	if (further_buffers_exist)
 		dcache_invalidate_region(needs_sync, sizeof(struct list_item));
 	/* no dirty cache lines exist for this buffer yet, no need to write back */
-	dcache_invalidate_region(uncache_to_cache(list), sizeof(*list));
+	dcache_writeback_invalidate_region(uncache_to_cache(list), sizeof(*list));
```
kv2019i, line 316: Hmm, this is now in conflict with the comment on L315 . So either we need to change the comment or there is something else going on.

@lyakh I believe you added the comment?

kv2019i (commented):
Thanks for the patch! We do need to understand why this happens and at minimum change the comment (see inline note).

src/audio/buffer.c
lyakh, line 316: this shouldn't be needed and is actually potentially dangerous. `list` here is an uncached address, so `list_item_prepend(list, head);` above modifies SRAM directly bypassing cache. If we now write back cache contents on top of our newest changes, we'll lose those our changes.

lyakh (changes requested):

src/audio/buffer.c
lyakh: I see that on mt8188 `cache_to_uncache()` is an identity conversion, so you don't have cached and uncached aliases. Is the memory always accessed via cache in your case?

wenliangwu: Currently, in MTK platform, a physical memory won't be simultaneously mapped to both cached and uncached memory. If a memory is mapped to cached memory, it will only be accessed through the cache.

In the case, I checked the memory address of the list and found that it is located in the cache memory. So yes, the memory is always accessed via cache.

btian1 (commented):
since PR status is ready for review, please change the comments accordingly: "This is a test PR for Issues#7916."

wenliangwu:
> since PR status is ready for review, please change the comments accordingly: "This is a test PR for Issues#7916."

@btian1 Sorry, I'm not quite familiar with github. My intention in creating the PR was to initiate a discussion, as I am uncertain if my solution is correct.  Is it possible to change PR status?

src/audio/buffer.c
lyakh: @wenliangwu aha, that's the difference, causing the problem then. If I understand you correctly, the memory, used by the DSP on MTK platforms is currently configured to use or not to use cache. The address will remain the same. But you said "currently," does that mean that in principle it would be possible to map memory for both access modes?
And with SOF you do configure some memory to use and the rest not to use cache. Can you tell me which memory does and which doesn't use cache? .data, .bss, .rodata, stack, heap. And you're using XTOS, right? AFAICS from reading the code under XTOS `rmalloc()` always allocates memory in cached mode, ignoring flags... That seems like a bug to me but I'd rather not try to fix it myself because (1) on Intel hardware it's enough to use a different address to switch between cached and uncached access modes, so our design uses that simply because we weren't sure how other architectures worked. (2) I'm not sure if we still can even build SOF with XTOS for Intel hardware, let alone test it. (3) The fix for Intel hardware would just use a different address for uncached allocations, whereas for MTK you need to remap that memory differently.
That being said I think a fix for this specific issue could be simple: just remove that invalidation line and update the comment above it as
```
-/* no dirty cache lines exist for this buffer yet, no need to write back */
+/* no dirty cache lines exist for this buffer yet, no need to write back or invalidate */
```

wenliangwu: > But you said "currently," does that mean that in principle it would be possible to map memory for both access modes? 
>
We don't have plan to do this. If the mapping is required, I will need to discuss it with other team members.

> which memory does and which doesn't use cache? .data, .bss, .rodata, stack, heap. 
>
Now, these sections all use cache.

> And you're using XTOS, right?
>
Yes, we still use XTOS. 

> That being said I think a fix for this specific issue could be simple: just remove that invalidation line and update the comment above it as
> 
> ```
> -/* no dirty cache lines exist for this buffer yet, no need to write back */
> +/* no dirty cache lines exist for this buffer yet, no need to write back or invalidate */
> ```
I have tested it before and it was able to fix the problem as well.  Furthermore, I have discovered that other companies also use an identity conversion in cache_to_uncache(), indicating that they may be facing similar problems. Therefore, I prefer your simple solution over dealing with memory remapping.

@lyakh If you believe that the root cause has been identified and your proposed solution can resolve the issue based on my test, would you be able to submit a pull request to fix it? As I do not fully understand the problem, it may not be wise for me to make any changes. If this leads to another issue, I may not be able to analyze it properly.

lgirdwood:
@wenliangwu fwiw, cache maangement is one of the reasons Intel are moving to Zephyr as the cache complexity is an "opt-in" for code sections/objects where high performance is needed vs an "opt-out" for code/objects where stability is needed. i.e. all Zephyr code by default uses the safe stability cache mappings/config by default and only the smaller high performance "hot" areas need to opt-in. I would recommend a similar direction with Zephyr.

src/audio/buffer.c
lyakh: > @lyakh If you believe that the root cause has been identified and your proposed solution can resolve the issue based on my test, would you be able to submit a pull request to fix it? As I do not fully understand the problem, it may not be wise for me to make any changes. If this leads to another issue, I may not be able to analyze it properly.

@wenliangwu sure, I could do that in principle, but it's your PR - you've identified a problem on your hardware and you've tracked it back to this cache operation, I'm just proposing a slightly modified version of your initial commit. But on Intel hardware we need that cache invalidation to solve the prefetch issue. So we need to make that line architecture specific... So you can just do this:
```
+#if CONFIG_INTEL
-	/* no dirty cache lines exist for this buffer yet, no need to write back */
+	/*
+	 * Until now the buffer object wasn't in cache, but uncached access to it could have
+	 * triggered a cache prefetch. Drop that cache line to avoid using stale data in it.
+	 */
 	dcache_invalidate_region(uncache_to_cache(list), sizeof(*list));
+#endif
```

wenliangwu: If we use #if CONFIG_INTEL to make the line specific for INTEL platform, will it cause issues when other company adopts  Zephyr?

lyakh: @wenliangwu I don't think it depends on the RTOS, otherwise I'd propose to use `#ifdef __ZEPHYR__`, I think it's the way how cached and uncached access is performed on Intel DSP implementations and on how cache prefetch works there.

wenliangwu: Got it. Thanks for your explanation.
Let's go with 
```
+#if CONFIG_INTEL
-	/* no dirty cache lines exist for this buffer yet, no need to write back */'
+	/*
+	 * Until now the buffer object wasn't in cache, but uncached access to it could have
+	 * triggered a cache prefetch. Drop that cache line to avoid using stale data in it.
+	 */
 	dcache_invalidate_region(uncache_to_cache(list), sizeof(*list));
+#endif
```

I will update the commit and the comment of pull request.

src/audio/buffer.c
```diff
@@ -312,8 +312,14 @@ void buffer_attach(struct comp_buffer *buffer, struct list_item *head, int dir)
 	list_item_prepend(list, head);
 	if (further_buffers_exist)
 		dcache_invalidate_region(needs_sync, sizeof(struct list_item));
-	/* no dirty cache lines exist for this buffer yet, no need to write back */
+#if CONFIG_INTEL
+	/* no dirty cache lines exist for this buffer yet, no need to write back
```
lyakh, line 316: I think this comment line can be dropped now, you probably forgot to remove it

wenliangwu: @lyakh I just copied the code snippet you pasted. (https://github.com/thesofproject/sof/pull/7925#discussion_r1263343320)
 
Do you mean I should drop the line "no dirty cache lines exist for this buffer yet, no need to write back"?

lyakh: @wenliangwu yes, the diff I posted had a `-` in front of that line, that meant that that line had to be dropped

btian1:
> > since PR status is ready for review, please change the comments accordingly: "This is a test PR for Issues#7916."
> 
> @btian1 Sorry, I'm not quite familiar with github. My intention in creating the PR was to initiate a discussion, as I am uncertain if my solution is correct. Is it possible to change PR status?

yes, you can change PR from ready for review to DRAFT by click right middle side: "convert to draft", before you convinced for review, we can keep it as DRAFT.,

btian1 (approved):

lgirdwood (approved):

lgirdwood:
SOFCI TEST

PR #7929, opened 2023-07-12
format: stream: pcm_converter: Add support for 8-bit formats

softwarecki:
This PR adds basic support for unsigned 8 bit format. 8 bit samples in 8 and 16 bit container and conversion between 32 bit signed format is supported.

### DIFF
--- a/src/audio/pcm_converter/pcm_converter_generic.c
+++ b/src/audio/pcm_converter/pcm_converter_generic.c
@@ -1,14 +1,17 @@
 // SPDX-License-Identifier: BSD-3-Clause
-//
-// Copyright(c) 2019 Intel Corporation. All rights reserved.
-//
-// Author: Tomasz Lauda <tomasz.lauda@linux.intel.com>
+/*
+ * Copyright(c) 2023 Intel Corporation. All rights reserved.
+ *
+ * Author: Tomasz Lauda <tomasz.lauda@linux.intel.com>
+ *	   Adrian Warecki <adrian.warecki@intel.com>
+ */
 
 /**
  * \file audio/pcm_converter/pcm_converter_generic.c
  * \brief PCM converter generic processing implementation
  * \authors Tomasz Lauda <tomasz.lauda@linux.intel.com>
  * \authors Karol Trzcinski <karolx.trzcinski@linux.intel.com>
+ * \authors Adrian Warecki <adrian.warecki@intel.com>
  */
 
 #include <sof/audio/pcm_converter.h>
@@ -484,6 +487,9 @@ static int pcm_convert_f_to_s32(const struct audio_stream __sparse_cache *source
 #endif /* CONFIG_PCM_CONVERTER_FORMAT_FLOAT && CONFIG_PCM_CONVERTER_FORMAT_S32LE */
 
 const struct pcm_func_map pcm_func_map[] = {
+#if CONFIG_PCM_CONVERTER_FORMAT_U8
+	{ SOF_IPC_FRAME_U8, SOF_IPC_FRAME_U8, audio_stream_copy },
+#endif /* CONFIG_PCM_CONVERTER_FORMAT_U8 */
 #if CONFIG_PCM_CONVERTER_FORMAT_S16LE
 	{ SOF_IPC_FRAME_S16_LE, SOF_IPC_FRAME_S16_LE, audio_stream_copy },
 #endif /* CONFIG_PCM_CONVERTER_FORMAT_S16LE */
@@ -493,7 +499,7 @@ const struct pcm_func_map pcm_func_map[] = {
 #if CONFIG_PCM_CONVERTER_FORMAT_S24_3LE
 	{ SOF_IPC_FRAME_S24_3LE, SOF_IPC_FRAME_S24_3LE, audio_stream_copy },
 #endif /* CONFIG_PCM_CONVERTER_FORMAT_S24_3LE */
-#if  CONFIG_PCM_CONVERTER_FORMAT_S24LE && CONFIG_PCM_CONVERTER_FORMAT_S16LE
+#if CONFIG_PCM_CONVERTER_FORMAT_S24LE && CONFIG_PCM_CONVERTER_FORMAT_S16LE
 	{ SOF_IPC_FRAME_S16_LE, SOF_IPC_FRAME_S24_4LE, pcm_convert_s16_to_s24 },
 	{ SOF_IPC_FRAME_S24_4LE, SOF_IPC_FRAME_S16_LE, pcm_convert_s24_to_s16 },
 #endif /* CONFIG_PCM_CONVERTER_FORMAT_S24LE && CONFIG_PCM_CONVERTER_FORMAT_S16LE */
@@ -841,6 +847,10 @@ const struct pcm_func_vc_map pcm_func_vc_map[] = {
 	{ SOF_IPC_FRAME_S32_LE, SOF_IPC_FRAME_S24_4LE, SOF_IPC_FRAME_S32_LE, SOF_IPC_FRAME_S16_LE,
 		ipc4_gtw_alh, ipc4_playback, pcm_convert_s24_to_s32 },
 #endif
+#if CONFIG_PCM_CONVERTER_FORMAT_U8 && CONFIG_PCM_CONVERTER_FORMAT_S16LE
+	{ SOF_IPC_FRAME_S16_LE, SOF_IPC_FRAME_U8, SOF_IPC_FRAME_S16_LE, SOF_IPC_FRAME_U8,
+		ipc4_gtw_all, ipc4_bidirection, audio_stream_copy },
+#endif /* CONFIG_PCM_CONVERTER_FORMAT_U8 && CONFIG_PCM_CONVERTER_FORMAT_S16LE */
 #if CONFIG_PCM_CONVERTER_FORMAT_S32LE && CONFIG_PCM_CONVERTER_FORMAT_S24LE
 	{ SOF_IPC_FRAME_S32_LE, SOF_IPC_FRAME_S24_4LE, SOF_IPC_FRAME_S32_LE, SOF_IPC_FRAME_S24_4LE,
 		ipc4_gtw_all & ~(ipc4_gtw_link | ipc4_gtw_alh | ipc4_gtw_host | ipc4_gtw_dmic),
--- a/src/audio/pcm_converter/pcm_converter_generic.c
+++ b/src/audio/pcm_converter/pcm_converter_generic.c
@@ -29,8 +29,69 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#define BYTES_TO_U8_SAMPLES	0
 #define BYTES_TO_S16_SAMPLES	1
 #define BYTES_TO_S32_SAMPLES	2
+
+#if CONFIG_PCM_CONVERTER_FORMAT_U8 && CONFIG_PCM_CONVERTER_FORMAT_S32LE
+static int pcm_convert_u8_to_s32(const struct audio_stream __sparse_cache *source,
+				uint32_t ioffset, struct audio_stream __sparse_cache *sink,
+				uint32_t ooffset, uint32_t samples)
+{
+	uint8_t *src = audio_stream_get_rptr(source);
+	int32_t *dst = audio_stream_get_wptr(sink);
+	uint32_t processed;
+	uint32_t nmax, i, n;
+
+	src += ioffset;
+	dst += ooffset;
+	for (processed = 0; processed < samples; processed += n) {
+		src = audio_stream_wrap(source, src);
+		dst = audio_stream_wrap(sink, dst);
+		n = samples - processed;
+		nmax = audio_stream_bytes_without_wrap(source, src) >> BYTES_TO_U8_SAMPLES;
+		n = MIN(n, nmax);
+		nmax = audio_stream_bytes_without_wrap(sink, dst) >> BYTES_TO_S32_SAMPLES;
+		n = MIN(n, nmax);
+		for (i = 0; i < n; i++) {
+			*dst = (INT8_MIN + *src) << 24;
+			src++;
+			dst++;
+		}
+	}
+
+	return samples;
+}
+
+static int pcm_convert_s32_to_u8(const struct audio_stream __sparse_cache *source,
+				uint32_t ioffset, struct audio_stream __sparse_cache *sink,
+				uint32_t ooffset, uint32_t samples)
+{
+	int32_t *src = audio_stream_get_rptr(source);
+	uint8_t *dst = audio_stream_get_wptr(sink);
+	uint32_t processed;
+	uint32_t nmax, i, n;
+
+	src += ioffset;
+	dst += ooffset;
+	for (processed = 0; processed < samples; processed += n) {
+		src = audio_stream_wrap(source, src);
+		dst = audio_stream_wrap(sink, dst);
+		n = samples - processed;
+		nmax = audio_stream_bytes_without_wrap(source, src) >> BYTES_TO_S32_SAMPLES;
+		n = MIN(n, nmax);
+		nmax = audio_stream_bytes_without_wrap(sink, dst) >> BYTES_TO_U8_SAMPLES;
+		n = MIN(n, nmax);
+		for (i = 0; i < n; i++) {
+			*dst = sat_int8(Q_SHIFT_RND(*src, 24, 0)) - INT8_MIN;
+			src++;
+			dst++;
+		}
+	}
+
+	return samples;
+}
+#endif /* CONFIG_PCM_CONVERTER_FORMAT_U8 && CONFIG_PCM_CONVERTER_FORMAT_S32LE */
 
 #if CONFIG_PCM_CONVERTER_FORMAT_S16LE && CONFIG_PCM_CONVERTER_FORMAT_S24LE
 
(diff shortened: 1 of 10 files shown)


src/include/sof/audio/audio_stream.h
```diff
@@ -1040,6 +1040,14 @@ static inline void audio_stream_fmt_conversion(enum ipc4_bit_depth depth,
...
 	*valid_fmt = (valid >> 3) - 2;
 
+	if (depth == 8) {
+		*frame_fmt = SOF_IPC_FRAME_U8;
+	}
+
+	if (valid == 8) {
+		*valid_fmt = SOF_IPC_FRAME_U8;
+	}
+
```
lgirdwood, line 1050: Best to have the KCONFIG around this to save space.

src/audio/pcm_converter/pcm_converter_generic.c
```diff
@@ -1,14 +1,17 @@
 // SPDX-License-Identifier: BSD-3-Clause
-//
-// Copyright(c) 2019 Intel Corporation. All rights reserved.
-//
-// Author: Tomasz Lauda <tomasz.lauda@linux.intel.com>
+/*
```
lgirdwood, line 2: IIRC, the SPDX spec uses //

lgirdwood (commented):
@softwarecki is MCPS critical here ? Do we need Hifi3/4 support ?
@singalsu @andrula-song fyi.

src/audio/pcm_converter/pcm_converter_generic.c
softwarecki: We seem to have adopted the following following rules in this project:

Start .c files like this:
```
// SPDX-License-Identifier: ...
/*
 * Copyright(c) ...
 *
 * Author: ...
 */

File: src/include/sof/audio/audio_stream.h (hunk 1 above)
Comment (btian1) [reply src/include/sof/audio/audio_stream.h:1050]: line 1043 - 1050 should ahead of line 1040, otherwise, "-2" operation will overflow.

File: src/include/sof/audio/audio_stream.h [hunk 3]
```diff
@@ -1040,6 +1040,14 @@ static inline void audio_stream_fmt_conversion(enum ipc4_bit_depth depth,
 	*frame_fmt = (depth >> 3) - 2;
 	*valid_fmt = (valid >> 3) - 2;
 
+	if (depth == 8) {
+		*frame_fmt = SOF_IPC_FRAME_U8;
+	}
+
+	if (valid == 8) {
```
RanderWang, line 1047: do you need to check type == IPC4_TYPE_UNSIGNED_INTEGER ?

src/audio/pcm_converter/pcm_converter_generic.c
```diff
@@ -493,7 +499,7 @@ const struct pcm_func_map pcm_func_map[] = {
 #if CONFIG_PCM_CONVERTER_FORMAT_S24_3LE
 	{ SOF_IPC_FRAME_S24_3LE, SOF_IPC_FRAME_S24_3LE, audio_stream_copy },
 #endif /* CONFIG_PCM_CONVERTER_FORMAT_S24_3LE */
-#if  CONFIG_PCM_CONVERTER_FORMAT_S24LE && CONFIG_PCM_CONVERTER_FORMAT_S16LE
+#if CONFIG_PCM_CONVERTER_FORMAT_S24LE && CONFIG_PCM_CONVERTER_FORMAT_S16LE
```
btian1, line 567: understand it is minor change, but separate patch will more clear.

src/audio/pcm_converter/pcm_converter_generic.c
```diff
@@ -841,6 +847,10 @@ const struct pcm_func_vc_map pcm_func_vc_map[] = {
 	{ SOF_IPC_FRAME_S32_LE, SOF_IPC_FRAME_S24_4LE, SOF_IPC_FRAME_S32_LE, SOF_IPC_FRAME_S16_LE,
 		ipc4_gtw_alh, ipc4_playback, pcm_convert_s24_to_s32 },
 #endif
+#if CONFIG_PCM_CONVERTER_FORMAT_U8 && CONFIG_PCM_CONVERTER_FORMAT_S16LE
+	{ SOF_IPC_FRAME_S16_LE, SOF_IPC_FRAME_U8, SOF_IPC_FRAME_S16_LE, SOF_IPC_FRAME_U8,
+		ipc4_gtw_all, ipc4_bidirection, audio_stream_copy },
```
btian1, line 917: why S16LE and U8 convert still use audio_stream_copy? it should use new converter function, if new function in next patch, we need squash them into one.

src/include/sof/audio/format_hifi3.h
```diff
@@ -27,4 +27,14 @@ static inline int16_t sat_int16(int32_t x)
...
 }
 
+static inline int8_t sat_int8(int32_t x)
+{
+	if (x > INT8_MAX)
+		return INT8_MAX;
+	else if (x < INT8_MIN)
+		return INT8_MIN;
+	else
+		return (int8_t)x;
```
btian1, line 37: and intrinsic seems does not have special instruction to handle it, no sure about clamp instruction, @singalsu @andrula-song , do you have ideas to handle it? use sat16, then shift right 8?

anyway, in hifi3, use directly c seems not an efficient way.

src/audio/pcm_converter/pcm_converter_generic.c
```diff
@@ -29,9 +29,70 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#define BYTES_TO_U8_SAMPLES	0
 #define BYTES_TO_S16_SAMPLES	1
 #define BYTES_TO_S32_SAMPLES	2
 
+#if CONFIG_PCM_CONVERTER_FORMAT_U8 && CONFIG_PCM_CONVERTER_FORMAT_S32LE
+static int pcm_convert_u8_to_s32(const struct audio_stream __sparse_cache *source,
```
btian1, line 37: may need add another patch to add it to hifi version as well, since current build enabled hifi3 or hifi4.

src/include/sof/audio/format_hifi3.h
andrula-song: The HiFi3 & HiFi4 all contains a 16-entry 64-bit register file, so I am afraid there would not be a direct intrinsic for this.
"use sat16, then shift right 8?"
--> if for C code, I think this would be better than use sat16, and for HiFi version maybe use a 16-bit container left shift 8 bits with saturation and then right shift 8 bits.

src/audio/pcm_converter/pcm_converter_generic.c
softwarecki: There is no conversion here. It is a simple copy of 8in16 frames to 8in16.

src/audio/pcm_converter/pcm_converter_generic.c
lgirdwood: Fair enough, I must be thinking of something else.

src/include/sof/audio/format_hifi3.h
lgirdwood: Ok, it seems Intrinsics dont make sense here. I think we are good then with generic C only.

src/include/ipc/stream.h
```diff
@@ -56,6 +56,7 @@ enum sof_ipc_frame {
 	SOF_IPC_FRAME_FLOAT,
 	/* other formats here */
 	SOF_IPC_FRAME_S24_3LE,
+	SOF_IPC_FRAME_U8,
```
kv2019i, line 59: This needs an ABI bump as this is adding an item to the IPC3 interface. I.e. just bump the MINOR version in this PR.

src/include/sof/audio/audio_stream.h
```diff
@@ -1040,6 +1040,14 @@ static inline void audio_stream_fmt_conversion(enum ipc4_bit_depth depth,
 	*frame_fmt = (depth >> 3) - 2;
 	*valid_fmt = (valid >> 3) - 2;
 
+	if (depth == 8) {
```
kv2019i, line 1043: "if (depth == 8 && valid == 8) {"

Wouldn't this be more accurate? It seems any other case would be an invalid configuration.

src/include/sof/audio/audio_stream.h
kv2019i: Ack @btian1 . If we can ensure both valid_fmt/frame_fmt are set (like in 24_3LE case), this is less of an issue as the invalid value gets overridden (and then we match the existing code layout for 24_3LE).

kv2019i (changes requested):
Only minor issues, please see inline.

btian1 (commented):
@softwarecki , I think you may need seach sof with frame_fmt or other key words to get more possible place to change, like in ssp.c, it currently only handle:

	switch (ssp->params.sample_valid_bits) {
	case 16:
		params->frame_fmt = SOF_IPC_FRAME_S16_LE;
		break;
	case 24:
		params->frame_fmt = SOF_IPC_FRAME_S24_4LE;
		break;
	case 32:
		params->frame_fmt = SOF_IPC_FRAME_S32_LE;
		break;
	default:
		dai_err(dai, "ssp_get_hw_params(): not supported format");
		return -EINVAL;
	}

src/include/sof/audio/format_generic.h
```diff
@@ -40,4 +40,14 @@ static inline int16_t sat_int16(int32_t x)
...
 }
 
+static inline int8_t sat_int8(int32_t x)
+{
+	if (x > INT8_MAX)
+		return INT8_MAX;
+	else if (x < INT8_MIN)
+		return INT8_MIN;
+	else
+		return (int8_t)x;
```
lyakh, line 50: all the `else`s can be dropped

src/audio/pcm_converter/pcm_converter_generic.c
```diff
@@ -29,9 +29,70 @@
...
+				  uint32_t ooffset, uint32_t samples)
+{
+	uint8_t *src = audio_stream_get_rptr(source);
+	int32_t *dst = audio_stream_get_wptr(sink);
+	int processed;
+	int nmax, i, n;
+
+	src += ioffset;
+	dst += ooffset;
+	for (processed = 0; processed < samples; processed += n) {
```
lyakh, line 48: this might be an example of why I usually use `unsigned` for non-negative loop iterators: `samples` is unsigned, `processed` is incremented by a variable `n`, which is calculated by subtraction, so *purely in theory* if `n` somehow becomes negative and `processed` could become negative too. I'm pretty sure the below logic makes this impossible, but just to be on the safe side in such cases...

src/audio/pcm_converter/pcm_converter_generic.c
marc-hb: It's a bit complicated but I tried to explain it in commit 1db950e0780cf7

src/include/sof/audio/audio_stream.h
softwarecki: There are many possible values of depth and valid (0..7, 9..15) for which this function returns a negative value and moving if upper will not correct it. The depth and valid variables is a enum and apparently the author assumed that there would be no illegal values. In the end, returning an invalid value won't break anything. It just won't find any conversion function.

src/include/sof/audio/audio_stream.h
softwarecki: Unfortunately, in my case, the type variable is 0. This enum seems invalid because two values at the same time can be used to describe a type, eg. IPC4_TYPE_MSB_INTEGER and IPC4_TYPE_UNSIGNED_INTEGER.

src/audio/pcm_converter/pcm_converter_generic.c
softwarecki: Let's do it in a separate PR.

src/include/sof/audio/audio_stream.h
softwarecki: I separated it to support also a 8 bit samples in 16 and 32 bit containers

src/include/sof/audio/format_generic.h
softwarecki: It returns int8 not uint8

softwarecki:
> @softwarecki , I think you may need seach sof with frame_fmt or other key words to get more possible place to change, like in ssp.c, it currently only handle:
@btian1 This PR adds basic support for the new format. If I wanted to handle everything in one PR, there would be a lot of changes, which would make the review process more difficult.

src/audio/pcm_converter/pcm_converter_generic.c
softwarecki: ok, I changed `int` to `uint32_t` because `samples` is defined as `uint32_t`. However, `audio_stream_bytes_without_wrap` returns `int`, which doesn't fit the puzzle.

src/include/ipc/stream.h
lgirdwood: ack, a minor update is all :)

lgirdwood (approved):

src/audio/pcm_converter/pcm_converter_generic.c
marc-hb: > audio_stream_bytes_without_wrap returns int, which doesn't fit the puzzle.

In commit a892e0f66cb74 you added an `assert()` to make sure the returned value is always positive. That function should just be fixed to return an unsigned type.

BTW I think at least one of these asserts is buggy because you've used `intptr_t`, see 64bits demo below. Why not just `uintptr_t`?

All this proves once again the point that C integers are an absolute nightmare full of pitfalls (and security vulnerabilities) and that they require extreme care.

```C
#include <assert.h>
#include <stdint.h>
#include <stdio.h>

int main()
{
  void * const ptr1 = (void *)  INT64_MAX;
  void * const ptr2 = (void *) UINT64_MAX ; // -1

  printf("%p, %p\n", ptr1, ptr2);
  printf("%ld, %ld\n", (intptr_t) ptr1, (intptr_t) ptr2);

  printf("ptr1 < ptr2 is %s\n",  (intptr_t) ptr1 <  (intptr_t) ptr2 ? "true" : "false");
  printf("ptr1 < ptr2 is %s\n", (uintptr_t) ptr1 < (uintptr_t) ptr2 ? "true" : "false");

  return 0;
}

ptr1 < ptr2 is false
ptr1 < ptr2 is true
```

marc-hb: https://stefansf.de/c-quiz/#integer-constants + many others

marc-hb (commented):
> @btian1 This PR adds basic support for the new format. If I wanted to handle everything in one PR, there would be a lot of changes, which would make the review process more difficult.

RanderWang (approved):
LGTM. And I am also working on S24_BE and will follow this PR.

test/cmocka/src/audio/selector/selector_test.c
```diff
@@ -438,8 +438,9 @@ static void test_audio_sel(void **state)
 		fill_source_s32(sel_state);
 		break;
 #endif /* CONFIG_FORMAT_S24LE || CONFIG_FORMAT_S32LE */
-/* TODO: add S24_3LE support */
-	case SOF_IPC_FRAME_S24_3LE:
```
btian1, line 442: I understand this means no u8 and s24_3le support in cmocka, does s24_3le support in SOF for windows?  just want to double confirm.

src/include/sof/audio/audio_stream.h
```diff
@@ -1040,6 +1040,14 @@ static inline void audio_stream_fmt_conversion(enum ipc4_bit_depth depth,
 	*frame_fmt = (depth >> 3) - 2;
 	*valid_fmt = (valid >> 3) - 2;
 
+#ifdef CONFIG_FORMAT_U8
+	if (depth == 8)
+		*frame_fmt = SOF_IPC_FRAME_U8;
+
+	if (valid == 8)
+		*valid_fmt = SOF_IPC_FRAME_U8;
+#endif /* CONFIG_FORMAT_U8 */
```
btian1, line 1049: maybe it is time to use switch to cover all possible frame format and valid format convert.

src/include/sof/audio/format_generic.h
btian1: simply remove line 49, and keep line 50, @lyakh , do you mean this?

src/audio/pcm_converter/pcm_converter_generic.c
btian1: you may need add it in this PR, otherwise, U8 support is not enabled fully for S32, because this part is not built with default,  just simply copy this generic C code to hifi3 or hifi4 like max function is fine.

kv2019i (approved):
Thanks, looks good now. I'll leave it to @softwarecki whether you want to add the hifi placeholder in this PR or a follow-up. One will get a build warning when it's missing, so I think this can be done separately as well.

src/include/sof/audio/format_generic.h
lyakh:
```
	if (x > INT8_MAX)
		return INT8_MAX;
	if (x < INT8_MIN)
		return INT8_MIN;
	return (int8_t)x;
```

tmleman (approved):

test/cmocka/src/audio/selector/selector_test.c
RanderWang: yes, it is supported by windows. SOF can support it but needs a lot of work.

kv2019i:
@softwarecki Can you comment on the last opens from @btian1 and indicate should we merge this and you follow-up with more PRs, or should we hold off ?

lgirdwood:
> @softwarecki Can you comment on the last opens from @btian1 and indicate should we merge this and you follow-up with more PRs, or should we hold off ?

Lets follow up, as it looks like comments are minor and PR is needed by other PRs.

lgirdwood (approved):

PR #7933, opened 2023-07-13
audio: drc: fix missing header

cujomalainey:


src/include/sof/audio/drc/drc.h
```diff
@@ -9,6 +9,7 @@
 
 #include <stdint.h>
 #include <sof/audio/drc/drc_plat_conf.h>
+#include <sof/audio/buffer.h>
```
btian1, line 12: https://github.com/thesofproject/sof/blob/main/src/audio/drc/drc.c#L9C30-L9C30

drc.c already include buffer.h, what's the purpose to add here? if needed add here, should above be deleted?

cujomalainey: Line 61 doesn't compile without this

lgirdwood (approved):

src/include/sof/audio/drc/drc.h
cujomalainey: Removed from the c file

btian1: const struct audio_stream __sparse_cache *source, ? this line, if this is the case, maybe simply add:
struct audio_stream;
before line 60 is fine.

cujomalainey: I would rather just include the proper header unless you have a reason not to

src/audio/drc/drc.c
```diff
@@ -6,7 +6,6 @@
 
 #include <sof/audio/component.h>
 #include <sof/audio/data_blob.h>
-#include <sof/audio/buffer.h>
```
lyakh, line 9: why are you removing it from here? Are you not using types from in in *this* file? Recursive header inclusion isn't enough, headers should be explicitly included where used. Including in both the .c and the .h is ok

src/include/sof/audio/drc/drc.h
lyakh: here actually just forward declarations *might* be enough

src/audio/drc/drc.c
kv2019i: @lyakh isn't a bit too much to include both in foo.h and foo.c ? we are not requiring this systematically.

src/include/sof/audio/drc/drc.h
kv2019i: I think including buffer.h here is ok. Forward declarations make sense in interface headers that are included in a large number of files, but I don't think we need to go overboard with this in files like drc.h that are only included in a handful of places.

kv2019i (approved):

PR #7937, opened 2023-07-13
mixin_out: remove frame_fmt in mixin mixout init function

btian1:
1. remove frame_fmt conversion in mixin mixout init function, it will be converted and assigned in parameter function.
2. change frame format type.

### DIFF
--- a/src/audio/mixin_mixout/mixin_mixout.c
+++ b/src/audio/mixin_mixout/mixin_mixout.c
@@ -100,7 +100,6 @@ static int mixin_init(struct processing_module *mod)
 	struct comp_dev *dev = mod->dev;
 	struct mixin_data *md;
 	int i;
-	enum sof_ipc_frame frame_fmt, valid_fmt;
 
 	comp_dbg(dev, "mixin_init()");
 
@@ -115,12 +114,6 @@ static int mixin_init(struct processing_module *mod)
 		md->sink_config[i].gain = IPC4_MIXIN_UNITY_GAIN;
 	}
 
-	audio_stream_fmt_conversion(mod->priv.cfg.base_cfg.audio_fmt.depth,
-				    mod->priv.cfg.base_cfg.audio_fmt.valid_bit_depth,
-				    &frame_fmt, &valid_fmt,
-				    mod->priv.cfg.base_cfg.audio_fmt.s_type);
-
-	dev->ipc_config.frame_fmt = valid_fmt;
 	mod->skip_src_buffer_invalidate = true;
 
 	mod->max_sinks = MIXIN_MAX_SINKS;
@@ -132,7 +125,6 @@ static int mixout_init(struct processing_module *mod)
 	struct module_source_info __sparse_cache *mod_source_info;
 	struct comp_dev *dev = mod->dev;
 	struct mixout_data *mo_data;
-	enum sof_ipc_frame frame_fmt, valid_fmt;
 
 	comp_dbg(dev, "mixout_new()");
 
@@ -144,12 +136,6 @@ static int mixout_init(struct processing_module *mod)
 	mod_source_info->private = mo_data;
 	module_source_info_release(mod_source_info);
 
-	audio_stream_fmt_conversion(mod->priv.cfg.base_cfg.audio_fmt.depth,
-				    mod->priv.cfg.base_cfg.audio_fmt.valid_bit_depth,
-				    &frame_fmt, &valid_fmt,
-				    mod->priv.cfg.base_cfg.audio_fmt.s_type);
-
-	dev->ipc_config.frame_fmt = valid_fmt;
 	mod->skip_sink_buffer_writeback = true;
 
 	mod->max_sources = MIXOUT_MAX_SOURCES;
--- a/src/include/sof/audio/mixer.h
+++ b/src/include/sof/audio/mixer.h
@@ -48,7 +48,7 @@ typedef void (*mixer_func)(struct comp_dev *dev, struct audio_stream __sparse_ca
 
 /** \brief Volume processing functions map. */
 struct mixer_func_map {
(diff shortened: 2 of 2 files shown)


ranj063:
@btian1 what issue does this fix?

btian1:
> @btian1 what issue does this fix?

this is the reason why it is draft, I am not sure about this yesterday before go to home, try to use CI to check,

This morning, with below code, I think there is no error with current implementation, and one more issue is fixed by you when you move function: ipc4_base_module_cfg_to_stream_params to helper.c

	switch (ssp->params.sample_valid_bits) {
	case 16:
		params->frame_fmt = SOF_IPC_FRAME_S16_LE;
		break;
	case 24:
		params->frame_fmt = SOF_IPC_FRAME_S24_4LE;
		break;
	case 32:
		params->frame_fmt = SOF_IPC_FRAME_S32_LE;
		break;
	default:
		dai_err(dai, "ssp_get_hw_params(): not supported format");
		return -EINVAL;
	}

ranj063:
> Q is: which is the right one for params->frame_fmt? frame_fmt or valid_fmt? we are using two different. @ranj063

@btian1 I think valid_fmt is the right one to use but I'd like @singalsu to confirm.

btian1:
> > Q is: which is the right one for params->frame_fmt? frame_fmt or valid_fmt? we are using two different. @ranj063
> 
> @btian1 I think valid_fmt is the right one to use but I'd like @singalsu to confirm.

yes, valid_fmt give more information compared with frame_fmt, from my search, only get_frame_bytes use this information, so both valid_fmt and frame_fmt both return same, this is the reason there is no failed case so far.
@ranj063 @singalsu 

anyway, I think we need align the usage, let's wait for seppo's comment and decide how to continue, and which part need the change.

kv2019i (commented):
Seems correct, but can you explain a bit why valid_fmt is wrong here? The code looks fairly confusing and it would help to have a git commit explaining the difference in this context. I can see mixin/mixout is calculating valid_fmt itself, but is this done later, so valid_fmt is not actually valid yet here. Or something else?

lgirdwood:
> Seems correct, but can you explain a bit why valid_fmt is wrong here? The code looks fairly confusing and it would help to have a git commit explaining the difference in this context. I can see mixin/mixout is calculating valid_fmt itself, but is this done later, so valid_fmt is not actually valid yet here. Or something else?

@btian1 ping.

btian1:
> 

SOF have two type of format classify: container size and valid size, normally, container size --> frame_fmt, and valid size -->valid_fmt, this is the normal usage, we should not mix them, if there is a need to use frame_fmt like valid format, we should add one more variable in dev to indicate this.

like for audio stream, there are two variables to indicate different case in sof_audio_stream_params:

	enum sof_ipc_frame frame_fmt;	/**< Sample data format */
	enum sof_ipc_frame valid_sample_fmt;

so, we should align the usage from begin to end to avoid mis-usage and mis-understanding, agree? @ranj063 @kv2019i @singalsu @lgirdwood

singalsu:
> > > Q is: which is the right one for params->frame_fmt? frame_fmt or valid_fmt? we are using two different. @ranj063
> > 
> > 
> > @btian1 I think valid_fmt is the right one to use but I'd like @singalsu to confirm.
> 
> yes, valid_fmt give more information compared with frame_fmt, from my search, only get_frame_bytes use this information, so both valid_fmt and frame_fmt both return same, this is the reason there is no failed case so far. @ranj063 @singalsu
> 
> anyway, I think we need align the usage, let's wait for seppo's comment and decide how to continue, and which part need the change.

Yep, valid_fmt is correct for all 16/24/32 bit formats including S24_LE. The frame_fmt is set to S32_LE for the non-packed 24 bit format that SOF uses.

lgirdwood (approved):

ranj063:
> Yep, valid_fmt is correct for all 16/24/32 bit formats including S24_LE. The frame_fmt is set to S32_LE for the non-packed 24 bit format that SOF uses.

@btian1 then the first patch is not correct is it?

btian1:
> 
@ranj063 @singalsu 
valid_fmt is converted through valid_bit_size, it does cover S24_4LE compared with frame_fmt, however, as I said, it we need valid_fmt, we should add it or change the variable name instead of assign valid_fmt to frame_fmt, this will cause mis-understanding.

RanderWang (commented):
why we need to build ipc_config.frame_fmt again ? but don't do it for IPC3 path ?
The idea is that: IPC3 driver will build ipc_config in kernel driver and send it for FW, but nothing for IPC4 driver. So FW has to build it by itself based on module init data. Currently the ipc_config is transferred from source to sink (.e.g. host to dai). If some module doesn't build it with audio_stream_fmt_conversion then it will use the ipc_config from its source module. If it uses a different format compared to source, there will be a error.  .e.g. 
input 32/32 ---> copier1 convert to 16/16 --->  copier2.   Copier2 will use the S32_LE for Copier1 if it doesn't build frame_fmt.

Now our CI can't detect the issue since we always has 32bit format for internal pipeline modules
So we need to create some test pipeline with format conversion modules.

lgirdwood (approved):
@ranj063 good for you ?

ranj063:
SOFCI TEST

src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -115,12 +114,6 @@ static int mixin_init(struct processing_module *mod)
 		md->sink_config[i].gain = IPC4_MIXIN_UNITY_GAIN;
 	}
 
-	audio_stream_fmt_conversion(mod->priv.cfg.base_cfg.audio_fmt.depth,
```
ranj063, line 118: @btian1 you're removing the assignment of the frame_fmt in the ipc_config but the commit message talks about something else totally. Can you please fix it?

PR #7944, opened 2023-07-13
Fix intelligo adapter

cujomalainey:
API modification breakages
- fix create arg attributes
- fix buffer access helpers

### DIFF
--- a/src/audio/igo_nr/igo_nr.c
+++ b/src/audio/igo_nr/igo_nr.c
@@ -402,7 +402,7 @@ static int32_t igo_nr_params(struct comp_dev *dev,
 	cd->source_rate = audio_stream_get_rate(&source_c->stream);
 	cd->sink_rate = audio_stream_get_rate(&sink_c->stream);
 
-	if (audio_stream_get_channels(source_c->stream) !=
+	if (audio_stream_get_channels(&source_c->stream) !=
 	    audio_stream_get_channels(&sink_c->stream)) {
 		comp_err(dev, "igo_nr_params(), mismatch source/sink stream channels");
 		cd->invalid_param = true;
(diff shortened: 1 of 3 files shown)


src/audio/igo_nr/igo_nr.c
```diff
@@ -402,7 +402,7 @@ static int32_t igo_nr_params(struct comp_dev *dev,
 	cd->source_rate = audio_stream_get_rate(&source_c->stream);
 	cd->sink_rate = audio_stream_get_rate(&sink_c->stream);
 
-	if (audio_stream_get_channels(source_c->stream) !=
+	if (audio_stream_get_channels(&source_c->stream) !=
```
btian1, line 405: interesting, this part of code should not be compiled with current build system, compiler will report error with this cse.

btian1 (approved):
Are we try to re-integrate this part of code into SOF?

cujomalainey:
> Are we try to re-integrate this part of code into SOF?

The code depends on a 3P library that hasn't been integrated past the CAVS2.5 branch. I'm working on adding stub libraries so we can at least build test and fuzz adapter code. These fixes are my findings from this work.

src/audio/igo_nr/igo_nr.c
cujomalainey: Yes, it's because it's not test in CI currently, trying to fix that :)

lyakh (approved):

kv2019i (approved):
Thanks @cujomalainey ! A stub that would allow at least a compile test in CI would be nice to catch these indeed.

kv2019i:
Lot of failing cases. We have a mismatch with upstream Zephyr main (not related to this PR), and we also have a regression in SOF main causing compiler warnings (again not related to this PR -> https://github.com/thesofproject/sof/pull/7949 ). So good to go for this PR.

PR #7932, opened 2023-07-13
audio: delete switch component

cujomalainey:
mostly not implemented and hasn't been meaningfully touched in > 3 yr.

PR #7945, opened 2023-07-14
module_interface.h: add forward struct comp_dev and timestamp_data

marc-hb:
Fixes commit 9cc6454df6d2 ("module_adapter: Extend the module interface for endpoint devices")

As the endpoint_ops methods use only pointers to these two objects, it's not necessary to make their entire object structure visible. However a forward declaration is required to make them known to endpoint_ops users and fix the following warning:

```
error: 'struct comp_dev' declared inside parameter list will not be
visible outside of this definition or declaration [-Werror]
```

### DIFF
--- a/src/include/sof/audio/module_adapter/module/module_interface.h
+++ b/src/include/sof/audio/module_adapter/module/module_interface.h
@@ -67,6 +67,8 @@ struct output_stream_buffer {
 	uint32_t size; /* size of data in the buffer */
 };
 
+struct comp_dev;
+struct timestamp_data;
 /**
  * \struct module_endpoint_ops
  * \brief Ops relevant only for the endpoint devices such as the host copier or DAI copier.


marc-hb:
It's still a mystery why Github Actions did not catch this in the original PR #7890 https://github.com/thesofproject/sof/actions/runs/5514181363/jobs/10053140087

Even weirder: it passed with the zephyr manifest in but not with the zephyr main branch in https://github.com/thesofproject/sof/actions/runs/5545805201/jobs/10125245211?pr=7932 for #7932?!? This misled me into wasting a bit of time with an irrelevant zephyr git bisect...

Sometimes PR A and PR B can both pass and not together but I don't see how this could have happened here Of course it depends on whether this .h file is actually used or not but how could it have been used before being merged?

btian1 (commented):
one question is: why CI does not report error before merge, and report error after merge?

line 70-71 add is needed, then how about below structure?

sof_ipc_stream_posn
sof_ipc_stream_params
input_stream_buffer
module_cfg_fragment_position

marc-hb:
> one question is: why CI does not report error before merge, and report error after merge?

See the first comment. Will look into it.

> then how about below structure?

> `sof_ipc_stream_posn`
> ...

Simply because `struct  sof_ipc_stream_posn` is already defined. For instance, this is how `pipeline-stream.c` gets `sof/src/include/ipc/stream.h`

```
In file included from sof/src/include/sof/audio/format.h:29,
                 from sof/src/include/sof/audio/audio_stream.h:17,
                 from sof/src/include/sof/audio/buffer.h:11,
                 from sof/src/audio/pipeline/pipeline-stream.c:8:
```

Other .c files get other .h files in other ways, otherwise the compiler would complain the same.

marc-hb:
Today's daily builds failed consistently: MTL and LNL all failed whatever the Zephyr version is while the others never failed.

src/include/sof/audio/module_adapter/module/module_interface.h
```diff
@@ -67,6 +67,8 @@ struct output_stream_buffer {
 	uint32_t size; /* size of data in the buffer */
 };
 
+struct comp_dev;
+struct timestamp_data;
```
lyakh, line 71: actually this file also should include stdint.h and stdbool.h

lyakh (approved):

lyakh:
> It's still a mystery why Github Actions did not catch this in the original PR #7890 https://github.com/thesofproject/sof/actions/runs/5514181363/jobs/10053140087

I think it worked originally because everywhere where this header was included originally also other headers were already available and now something changed somewhere - either the order of includes, or a new one was added without that accidental fix, or a header got removed somewhere.

marc-hb:
> and now something changed somewhere - either the order of includes, or a new one was added without that accidental fix, or a header got removed somewhere.

... which means it could break again, so maybe @btian1 has a point?

Should we try to compile this .h file with a empty.c file and see what happens, would that help?

This looks very relevant:
- https://github.com/zephyrproject-rtos/zephyr/issues/41543

lgirdwood (approved):

src/include/sof/audio/module_adapter/module/module_interface.h
lgirdwood: can be fixed incrementally.

lgirdwood:
Will take this now as it fixes build with some configs, can be refined further if needed.

PR #7951, opened 2023-07-14
Fixes for cross core connected pipelines

serhiy-katsyuba-intel:
Set of fixes to support connected pipelines created on different cores. We need cross core connected pipelines support as part of DP scheduling implementation.

Additionally, this PR https://github.com/thesofproject/sof/pull/7962 is required to support multi pipeline set state IPC when pipelines are created on different cores.


marcinszkudlinski:
Good direction: a module **MUST** be used exclusively on one core - from creation to destruction. That includes data copying and all setup callbacks etc.

cross core / cache operations should be executed by pipeline/sink/src/etc. code

lyakh (commented):
Both `comp_bind()` and `comp_get_attribute()`, called from `ipc4_create_buffer()` are called from `ipc_comp_connect()` which has https://github.com/thesofproject/sof/blob/95874fd8e2e12219cb49c0504427d8e9d1eff974/src/ipc/ipc4/helper.c#L375-L376 so the whole that IPC will already run on the target core, why do you need to redirect single functions there?

serhiy-katsyuba-intel:
> so the whole that IPC will already run on the target core, why do you need to redirect individual functions there?

Consider the case of two connected pipelines, one on core 1 and another on core 2. In such case we cannot simply pass IPC to core 1 or core 2, instead IPC is still processed on core 0 and bind() is called on core 1 for one of the components and on core 2 for another.

Current implementation (before this PR) only supports a case when both connected components are on same core, but, we, seems, have to support a cross core connection on pipelines.

lyakh:
> > so the whole that IPC will already run on the target core, why do you need to redirect individual functions there?
> 
> Consider the case of two connected pipelines, one on core 1 and another on core 2. In such case we cannot simply pass IPC to core 1 or core 2, instead IPC is still processed on core 0 and bind() is called on core 1 for one of the components and on core 2 for another.
> 
> Current implementation (before this PR) only supports a case when both connected components are on same core, but, we, seems, have to support a cross core connection of pipelines.

Ah, the `SOF_IPC4_MOD_BIND` IPC actually handles 2 components - a source and a destination, which can run on different cores (when this is actually supported). Got it, you're right, thanks.

src/idc/idc.c
```diff
@@ -126,6 +126,40 @@ static void idc_ipc(void)
...
+{
+	struct ipc_comp_dev *ipc_dev;
+	struct idc_payload *payload;
+	struct ipc4_module_bind_unbind *bu;
+
+	ipc_dev = ipc_get_comp_by_id(ipc_get(), comp_id);
+	if (!ipc_dev)
+		return -ENODEV;
+
+	payload = idc_payload_get(*idc_get(), cpu_get_id());
```
lgirdwood, line 140: what heppens if payload is NULL ? best to check.

src/include/sof/audio/component_ext.h
```diff
@@ -208,15 +208,61 @@ static inline int comp_prepare(struct comp_dev *dev)
...
+	rfree(base_cfg);
+	return ret;
+}
+#endif	/* CONFIG_IPC_MAJOR_4 */
 
 /** See comp_ops::get_attribute */
 static inline int comp_get_attribute(struct comp_dev *dev, uint32_t type,
 				     void *value)
 {
+#if CONFIG_IPC_MAJOR_4
```
lgirdwood, line 253: better to have two copies of ```comp_get_attribute```, 1 for ipc3 and the other for ipc4 (and this way we can put the ipc4 version with all the other ipc4 code in teh header under a single ifdef).

lgirdwood (commented):
Looks good @serhiy-katsyuba-intel, I guess also needs more validation before merge ready ?

src/idc/idc.c
serhiy-katsyuba-intel: idc_payload_get() returns pointer to static_payload array (+ offset depending on core). That static_payload is declared statically in idc.c and so never is NULL (unless someone modify the code).

marcinszkudlinski (approved):

lgirdwood (approved):

Issue #7931, opened 2023-07-12
[BUG][Multicore] topology loading failed with source module-copier.21.2 not found


fredoh9:
**Describe the bug**

MTLP_RVP_NOCODEC_MULTICORE platform failed to load multicore topology.

```
[    5.092484] snd_sof:sof_route_load: sof-audio-pci-intel-mtl 0000:00:1f.3: sink host-copier.0.capture control none source dai-copier.SSP.NoCodec-0.capture
[    5.092488] snd_sof:sof_route_load: sof-audio-pci-intel-mtl 0000:00:1f.3: sink host-copier.2.capture control none source dai-copier.SSP.NoCodec-2.capture
[    5.092491] snd_sof:sof_route_load: sof-audio-pci-intel-mtl 0000:00:1f.3: sink gain.1.1 control none source host-copier.0.playback
[    5.092494] snd_sof:sof_route_load: sof-audio-pci-intel-mtl 0000:00:1f.3: sink gain.5.1 control none source host-copier.2.playback
[    5.092496] snd_sof:sof_route_load: sof-audio-pci-intel-mtl 0000:00:1f.3: sink gain.19.1 control none source module-copier.21.2
[    5.092498] sof-audio-pci-intel-mtl 0000:00:1f.3: error: source module-copier.21.2 not found
[    5.092533] sof-nocodec sof-nocodec: ASoC: topology: add_route failed: -22
[    5.092556] sof-nocodec sof-nocodec: ASoC: topology: could not load header: -22
[    5.092726] sof-audio-pci-intel-mtl 0000:00:1f.3: error: tplg component load failed -22
[    5.092758] sof-audio-pci-intel-mtl 0000:00:1f.3: error: failed to load DSP topology -22
[    5.092784] sof-audio-pci-intel-mtl 0000:00:1f.3: ASoC: error at snd_soc_component_probe on 0000:00:1f.3: -22
[    5.092839] sof-nocodec sof-nocodec: ASoC: failed to instantiate card -22
[    5.093102] sof-nocodec: probe of sof-nocodec failed with error -22
```

**To Reproduce**
Just loading the latest sof-mtl-nocodec-multicore.tplg

**Reproduction Rate**
100%

**Expected behavior**

**Impact**
What impact does this issue have on your progress (e.g., annoyance, showstopper)

**Environment**

- Kernel Branch: topic/sof-dev
[...]

PR #7847, opened 2023-06-21
topology2: nocodec-multicore: Add connected pipelines

yongzhi1:
Tested multi-stream capture on core1.

PR #7954, opened 2023-07-17
topology2: use pass-through pipeline for multicore test on cavs

RanderWang:
platforms

Fix daily CI issuse: failed to load multicore topology

It is caused by : https://github.com/thesofproject/sof/pull/7847. Currently DMIC is not enabled by daily CI test on CAVS platforms, so we need to set pass-though = true.

fix https://github.com/thesofproject/sof/issues/7931

### DIFF
--- a/tools/topology/topology2/development/tplg-targets.cmake
+++ b/tools/topology/topology2/development/tplg-targets.cmake
@@ -43,10 +43,12 @@ PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-mtl-nocodec-ssp0-ssp2.bin,DEEPBUFFER_F
 DEEPBUFFER_D0I3_COMPATIBLE=true"
 
 "cavs-nocodec-multicore\;sof-mtl-nocodec-multicore\;PLATFORM=mtl,SSP1_ENABLED=true,SSP0_CORE_ID=0,\
-SSP1_CORE_ID=1,SSP2_CORE_ID=2,PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-mtl-nocodec-multicore.bin"
+SSP1_CORE_ID=1,SSP2_CORE_ID=2,PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-mtl-nocodec-multicore.bin,\
+PASSTHROUGH=true"
 
 "cavs-nocodec-multicore\;sof-mtl-nocodec-multicore-ssp0-ssp2\;PLATFORM=mtl,SSP1_ENABLED=false,\
-SSP2_CORE_ID=1,PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-mtl-nocodec-multicore-ssp0-ssp2.bin"
+SSP2_CORE_ID=1,PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-mtl-nocodec-multicore-ssp0-ssp2.bin,\
+PASSTHROUGH=true"
 
 "cavs-nocodec-multicore\;sof-mtl-nocodec-multicore-4ch\;PLATFORM=mtl,SSP1_ENABLED=false,\
 SSP0_CORE_ID=0,DMIC_CORE_ID=1,SSP2_CORE_ID=2,NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,\
--- a/tools/topology/topology2/development/tplg-targets.cmake
+++ b/tools/topology/topology2/development/tplg-targets.cmake
@@ -21,10 +21,12 @@ PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-adl-nocodec.bin,DEEPBUFFER_FW_DMA_MS=1
 PLATFORM=adl"
 
 "cavs-nocodec-multicore\;sof-tgl-nocodec-multicore-ssp0-ssp2\;PLATFORM=tgl,SSP1_ENABLED=false,\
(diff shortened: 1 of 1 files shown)


kv2019i (approved):

fredoh9:
Daily CI broker is for MTL, how will this fix MTL issue?

RanderWang:
> Daily CI broker is for MTL, how will this fix MTL issue?

Thanks! updated for MTL.

aiChaoSONG:
I don't think this will work, because define PASSTHROUGH to true is the same as not defining it or define it to any value other than false. 

Possible fix of issue #7931

RanderWang:
> I don't think this will work, because define PASSTHROUGH to true is the same as not defining it or define it to any value other than false.
> 
> Possible fix of issue #7931 is #7960

No, the default PASSTHROUGH is false.

RanderWang:
Attach topology build warning: why CI can't detect it ?
````
cmake --build /home/rander/Workspace/sof/tools/build_tools -- -j 20
[0/1] Re-running CMake...
-- Configuring done (0.1s)
-- Generating done (0.0s)
-- Build files have been written to: /home/rander/Workspace/sof/tools/build_tools
[0/3] Generating sof-mtl-nocodec-multicore.tplg
saving nhlt as binary in nhlt-sof-mtl-nocodec-multicore.bin
ALSA lib dapm.c:247:(tplg_build_routes) undefined source widget/stream 'module-copier.21.2'
[2/3] Generating sof-mtl-nocodec-multicore-ssp0-ssp2.tplg
saving nhlt as binary in nhlt-sof-mtl-nocodec-multicore-ssp0-ssp2.bin
ALSA lib dapm.c:247:(tplg_build_routes) undefined source widget/stream 'module-copier.21.2'

WARNING: building tools/ is now incremental by default!
         To build from scratch delete: /home/rander/Workspace/sof/tools/build_tools
         or use the -C option.
````

RanderWang:
the original PR:https://github.com/thesofproject/sof/pull/7847 change PASSTHROUGH=true to false but forgot to add it to other MTL topology build, so It is unified to to set PASSTHROUGH=true for other MTL build.

aiChaoSONG:
@RanderWang But do we really want use PASSTHROUGH=true, this will cause every pipeline to  host_copier->dai_copier connection, everything in the middle is removed. 

Of course, there is a gap between cavs-nocodec.conf and cavs-nocodec-multicore.conf, for cavs-nocodec.conf PASSTHROUGH=true makes every host->dai connection passthrough, but for cavs-nocodec-multicore.conf, PASSTHROUGH=true only makes DMIC passthrough.

RanderWang:
> > I don't think this will work, because define PASSTHROUGH to true is the same as not defining it or define it to any value other than false.
> > Possible fix of issue #7931 is #7960
> 
> No, the default PASSTHROUGH is false.



> @RanderWang But do we really want use PASSTHROUGH=true, this will cause every pipeline to host_copier->dai_copier connection, everything in the middle is removed.
> 
> Of course, there is a gap between cavs-nocodec.conf and cavs-nocodec-multicore.conf, for cavs-nocodec.conf PASSTHROUGH=true makes every host->dai connection passthrough, but for cavs-nocodec-multicore.conf, PASSTHROUGH=true only makes DMIC passthrough.
> 
> @ranj063 I think we need get cavs-nocodec.conf and cavs-nocodec-multicore.conf aligned?

@aiChaoSONG  the cavs-nocodec-multicore.conf is for CI to validate multicore in simplest case, so we just use passthrough.

kv2019i:
@RanderWang @aiChaoSONG @ranj063 It of course really depends what this tplg is supposed to do. It's a fair point from @aiChaoSONG  that if this is not aligned with the baseline cavs-nocodec, it's going to be harder and harder to maintain this when the baseline tplg is developed further.

RanderWang:
> @RanderWang @aiChaoSONG @ranj063 It of course really depends what this tplg is supposed to do. It's a fair point from @aiChaoSONG that if this is not aligned with the baseline cavs-nocodec, it's going to be harder and harder to maintain this when the baseline tplg is developed further.

@kv2019i We don't need to align cavs-nocodec.tplg and cavs-nocodec-multicore.tplg. A simple and better method is to add core setting in cavs-nocodec.tplg then we can discard cavs-nocodec-multicore.tplg and avoid any more maintain effort.

The cavs-nocodec-multicore.tplg is a simple pass-through pipeline when we validated multicore feature at first. At that time we had many bugs include hw hang. Now we can drop it since everything is stable now. If we still want to keep cavs-nocodec-multicore.tplg for feature validation on LNL or later chip,  I will want to make it a simple pass-through one to make debug easier.

RanderWang:
@kv2019i  In my opinion, developer should add core setting to the really used topology like hda or sdw topology, not the case that port some pipelines to cavs-nocodec-multicore.tplg and make it complex more and more.

aiChaoSONG:
>We don't need to align cavs-nocodec.tplg and cavs-nocodec-multicore.tplg. 

@RanderWang we are aligning the behavior of PASSTHROUGH, it is designed to do one-key passthrough for all host->dai connections, not only for DMIC. it should behave the same in whatever topology conf.

RanderWang:
> > We don't need to align cavs-nocodec.tplg and cavs-nocodec-multicore.tplg.
> 
> @RanderWang we are aligning the behavior of PASSTHROUGH, it is designed to do one-key passthrough for all host->dai connections, not only for DMIC. it should behave the same in whatever topology conf.

I don't want any PASSTHROUGH code in cavs-nocodec-multicore.tplg since It is designed to use pass-through pipeline.

RanderWang:
I suggest two options: (1) add multicore setting in cavs-nocodec.conf and remove cavs-nocodec-multicore.conf
(2) make cavs-nocodec-multicore.conf simple and remove the added PASSTHROUGH code

kv2019i:
Thanks @RanderWang and @aiChaoSONG . Rander's proposal seems good to me.

RanderWang:
thanks!

PR #7958, opened 2023-07-18
copier code refine and module parameter simplify

btian1:
@kv2019i note: this is split from https://github.com/thesofproject/sof/pull/7909

### DIFF
--- a/src/audio/aria/aria.c
+++ b/src/audio/aria/aria.c
@@ -154,23 +154,11 @@ static void aria_set_stream_params(struct comp_buffer *buffer,
 				   struct processing_module *mod)
 {
 	struct comp_buffer __sparse_cache *buffer_c;
-	enum sof_ipc_frame valid_fmt, frame_fmt;
 	const struct ipc4_audio_format *audio_fmt = &mod->priv.cfg.base_cfg.audio_fmt;
-	struct aria_data *cd = module_get_private_data(mod);
 
 	buffer_c = buffer_acquire(buffer);
 
-	audio_stream_fmt_conversion(audio_fmt->depth,
-				    audio_fmt->valid_bit_depth,
-				    &frame_fmt, &valid_fmt,
-				    audio_fmt->s_type);
-	audio_stream_set_buffer_fmt(&buffer_c->stream,
-				   audio_fmt->interleaving_style);
-	audio_stream_set_frm_fmt(&buffer_c->stream, frame_fmt);
-	audio_stream_set_valid_fmt(&buffer_c->stream, valid_fmt);
-	audio_stream_set_channels(&buffer_c->stream, cd->chan_cnt);
-	audio_stream_set_rate(&buffer_c->stream, audio_fmt->sampling_frequency);
-
+	ipc4_update_buffer_format(buffer_c, audio_fmt);
 #ifdef ARIA_GENERIC
 	audio_stream_init_alignment_constants(1, 1, &buffer_c->stream);
 #else
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -123,8 +123,7 @@ static int copier_init(struct processing_module *mod)
 #if CONFIG_IPC4_GATEWAY
 		case ipc4_ipc_output_class:
 		case ipc4_ipc_input_class:
-			ret = copier_ipcgtw_create(dev, cd, &dev->ipc_config,
-						   copier, ipc_pipe->pipeline);
+			ret = copier_ipcgtw_create(dev, cd, copier, ipc_pipe->pipeline);
 			if (ret < 0) {
 				comp_err(dev, "unable to create IPC gateway");
 				goto error;
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -101,8 +101,7 @@ static int copier_init(struct processing_module *mod)
 		switch (node_id.f.dma_type) {
 		case ipc4_hda_host_output_class:
 		case ipc4_hda_host_input_class:
-			ret = copier_host_create(dev, cd, &dev->ipc_config,
-						 copier, cd->direction, ipc_pipe->pipeline);
+			ret = copier_host_create(dev, cd, copier, ipc_pipe->pipeline);
 			if (ret < 0) {
 				comp_err(dev, "unable to create host");
 				goto error;
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -114,8 +114,7 @@ static int copier_init(struct processing_module *mod)
 		case ipc4_i2s_link_input_class:
 		case ipc4_alh_link_output_class:
 		case ipc4_alh_link_input_class:
-			ret = copier_dai_create(dev, cd, &dev->ipc_config,
-						copier, ipc_pipe->pipeline);
+			ret = copier_dai_create(dev, cd, copier, ipc_pipe->pipeline);
 			if (ret < 0) {
 				comp_err(dev, "unable to create dai");
 				goto error;
--- a/src/audio/copier/copier_dai.c
+++ b/src/audio/copier/copier_dai.c
@@ -99,8 +99,6 @@ static int copier_dai_init(struct comp_dev *dev,
 		pipeline->source_comp = dev;
 	}
 
-	dev->ipc_config.frame_fmt = config->frame_fmt;
-
 	/* save the channel map and count for ALH multi-gateway */
 	if (type == ipc4_gtw_alh && is_multi_gateway(copier->gtw_cfg.node_id)) {
 		ret = copier_set_alh_multi_gtw_channel_map(dev, copier, index);
--- a/src/audio/copier/copier_dai.c
+++ b/src/audio/copier/copier_dai.c
@@ -31,18 +31,18 @@ static uint32_t bitmask_to_nibble_channel_map(uint8_t bitmask)
 	return nibble_map;
 }
 
-static int copier_set_alh_multi_gtw_channel_map(struct comp_dev *parent_dev,
+static int copier_set_alh_multi_gtw_channel_map(struct comp_dev *dev,
 						const struct ipc4_copier_module_cfg *copier_cfg,
 						int index)
 {
-	struct processing_module *mod = comp_get_drvdata(parent_dev);
+	struct processing_module *mod = comp_get_drvdata(dev);
 	struct copier_data *cd = module_get_private_data(mod);
 	const struct sof_alh_configuration_blob *alh_blob;
 	uint8_t chan_bitmask;
 	int channels;
 
 	if (!copier_cfg->gtw_cfg.config_length) {
-		comp_err(parent_dev, "No ipc4_alh_multi_gtw_cfg found in blob!");
+		comp_err(dev, "No ipc4_alh_multi_gtw_cfg found in blob!");
 		return -EINVAL;
 	}
 
@@ -54,7 +54,7 @@ static int copier_set_alh_multi_gtw_channel_map(struct comp_dev *parent_dev,
 
 	channels = popcount(chan_bitmask);
 	if (channels < 1 || channels > SOF_IPC_MAX_CHANNELS) {
-		comp_err(parent_dev, "Invalid channels mask: 0x%x", chan_bitmask);
+		comp_err(dev, "Invalid channels mask: 0x%x", chan_bitmask);
 		return -EINVAL;
 	}
 
@@ -64,7 +64,7 @@ static int copier_set_alh_multi_gtw_channel_map(struct comp_dev *parent_dev,
 	return 0;
 }
 
-static int copier_dai_init(struct comp_dev *parent_dev,
+static int copier_dai_init(struct comp_dev *dev,
 			   struct comp_ipc_config *config,
 			   const struct ipc4_copier_module_cfg *copier,
 			   struct pipeline *pipeline,
@@ -72,7 +72,7 @@ static int copier_dai_init(struct comp_dev *parent_dev,
 			   enum ipc4_gateway_type type,
 			   int index, int dai_count)
 {
-	struct processing_module *mod = comp_get_drvdata(parent_dev);
+	struct processing_module *mod = comp_get_drvdata(dev);
 	struct copier_data *cd = module_get_private_data(mod);
 	struct dai_data *dd;
 	int ret;
@@ -86,7 +86,7 @@ static int copier_dai_init(struct comp_dev *parent_dev,
 					    &out_valid_fmt,
 					    copier->out_fmt.s_type);
 		config->frame_fmt = out_frame_fmt;
-		pipeline->sink_comp = parent_dev;
+		pipeline->sink_comp = dev;
 		cd->bsource_buffer = true;
 	} else {
 		enum sof_ipc_frame in_frame_fmt, in_valid_fmt;
@@ -96,14 +96,14 @@ static int copier_dai_init(struct comp_dev *parent_dev,
 					    &in_frame_fmt, &in_valid_fmt,
 					    copier->base.audio_fmt.s_type);
 		config->frame_fmt = in_frame_fmt;
-		pipeline->source_comp = parent_dev;
+		pipeline->source_comp = dev;
 	}
 
-	parent_dev->ipc_config.frame_fmt = config->frame_fmt;
+	dev->ipc_config.frame_fmt = config->frame_fmt;
 
 	/* save the channel map and count for ALH multi-gateway */
 	if (type == ipc4_gtw_alh && is_multi_gateway(copier->gtw_cfg.node_id)) {
-		ret = copier_set_alh_multi_gtw_channel_map(parent_dev, copier, index);
+		ret = copier_set_alh_multi_gtw_channel_map(dev, copier, index);
 		if (ret < 0)
 			return ret;
 	}
@@ -112,14 +112,14 @@ static int copier_dai_init(struct comp_dev *parent_dev,
 	if (!dd)
 		return -ENOMEM;
 
-	ret = dai_common_new(dd, parent_dev, dai);
+	ret = dai_common_new(dd, dev, dai);
 	if (ret < 0)
 		goto free_dd;
 
 	pipeline->sched_id = config->id;
 
 	cd->dd[index] = dd;
-	ret = comp_dai_config(cd->dd[index], parent_dev, dai, copier);
+	ret = comp_dai_config(cd->dd[index], dev, dai, copier);
 	if (ret < 0)
 		goto e_zephyr_free;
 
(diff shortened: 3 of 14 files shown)


src/audio/eq_fir/eq_fir.c
```diff
@@ -171,36 +171,27 @@ static inline int set_fir_func(struct processing_module *mod, enum sof_ipc_frame
...
+	struct comp_buffer *sinkb, *sourceb;
+	struct comp_buffer __sparse_cache *sink_c, *source_c;
 
 	comp_dbg(dev, "eq_fir_params()");
 
-	comp_params = *params;
-	comp_params.channels = mod->priv.cfg.base_cfg.audio_fmt.channels_count;
-	comp_params.rate = mod->priv.cfg.base_cfg.audio_fmt.sampling_frequency;
-	comp_params.buffer_fmt = mod->priv.cfg.base_cfg.audio_fmt.interleaving_style;
+	ipc4_base_module_cfg_to_stream_params(&mod->priv.cfg.base_cfg, params);
```
lyakh, line 180: before this change `mod->stream_params` wasn't modified, it was used as a template, now it's modified. Is this correct?

src/audio/eq_iir/eq_iir.c
```diff
@@ -835,35 +835,27 @@ static void eq_iir_set_alignment(struct audio_stream __sparse_cache *source,
...
-	comp_params.rate = mod->priv.cfg.base_cfg.audio_fmt.sampling_frequency;
-	comp_params.buffer_fmt = mod->priv.cfg.base_cfg.audio_fmt.interleaving_style;
-
-	audio_stream_fmt_conversion(mod->priv.cfg.base_cfg.audio_fmt.depth,
-				    mod->priv.cfg.base_cfg.audio_fmt.valid_bit_depth,
-				    &frame_fmt, &valid_fmt,
-				    mod->priv.cfg.base_cfg.audio_fmt.s_type);
 
-	comp_params.frame_fmt = valid_fmt;
+	ipc4_base_module_cfg_to_stream_params(&mod->priv.cfg.base_cfg, params);
```
lyakh, line 844: same here

src/audio/src/src.c
```diff
@@ -533,28 +533,14 @@ static int src_stream_pcm_sink_rate_check(struct ipc4_config_src cfg,
...
-	audio_stream_fmt_conversion(mod->priv.cfg.base_cfg.audio_fmt.depth,
-				    mod->priv.cfg.base_cfg.audio_fmt.valid_bit_depth,
-				    &frame_fmt, &valid_fmt,
-				    mod->priv.cfg.base_cfg.audio_fmt.s_type);
-
-	src_params.frame_fmt = valid_fmt;
-	component_set_nearest_period_frames(dev, src_params.rate);
-
-	ret = sink_set_params(sink, &src_params, true);
+	ipc4_base_module_cfg_to_stream_params(&mod->priv.cfg.base_cfg, params);
```
lyakh, line 541: `mod->stream_params` is modified

kv2019i (commented):
Thanks for splitting this out @btian1 ! I think this is very good cleanup for the most part. There are a few smaller things were we change functionality and it's not a plain refactoring. I'm ok with this but please @ranj063 ack these are correct.

And a generic rant and reminder that please let's remember to add rationale to git commit message if we a change is not a simpler refactor.

src/audio/module_adapter/module/generic.c
```diff
@@ -83,12 +83,6 @@ int module_init(struct processing_module *mod, struct module_interface *interfac
 
 	comp_dbg(dev, "module_init() start");
 
-#if CONFIG_IPC_MAJOR_3
-	if (mod->priv.state == MODULE_INITIALIZED)
-		return 0;
-	if (mod->priv.state > MODULE_INITIALIZED)
-		return -EPERM;
-#endif
```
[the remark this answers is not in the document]
ranj063: @kv2019i this is safe to remove for IPC3 but as for IPC4, I think i'd like @dbaluta to confirm is it is OK to remove for the cadence codecs

src/audio/copier/copier_dai.c
```diff
@@ -265,7 +265,7 @@ int copier_dai_create(struct comp_dev *dev, struct copier_data *cd,
 
 	/* create multi_endpoint_buffer for ALH multi-gateway case */
 	if (dai_count > 1) {
-		ret = create_endpoint_buffer(dev, cd, config, copier, type, true, 0);
+		ret = create_endpoint_buffer(dev, cd, copier, true);
```
RanderWang, line 268: please drop this commit. It will be used by https://github.com/thesofproject/sof/pull/7959/commits/78a8eec9cb9a010065c3bd7294f699986fbe7015

RanderWang (commented):
Good job!  Please check my comments about create_endpoint_buffer

src/audio/eq_fir/eq_fir.c
lyakh: I'm confused so I'm going to give this PR a -1 for now, I think we have to clarify something. I tried to follow where parameters come from and how they're used, so I went back to https://github.com/thesofproject/sof/commit/a25cdbb90b67be1f08c81e8de1641b7cefdf5fd5 and I now don't think it's quite correct. At the very least it doesn't seem to do what its description is saying. It isn't updating parameters, it's degrading them. Parameters are allocated in `module_adapter_params()` and copied from the IPC, but then that commit overwrites it with original data from `.base_cfg`. Now, I think with IPC4 data "from the IPC" is actually fake and all 0 - it comes from https://github.com/thesofproject/sof/blob/214f98378f12ed179bd22f0c4feb09b0b3aaea81/src/ipc/ipc4/handler.c#L117 but then some of it is overwritten in https://github.com/thesofproject/sof/blob/214f98378f12ed179bd22f0c4feb09b0b3aaea81/src/audio/module_adapter/module_adapter.c#L522 and also some of it is used to configure buffers e.g. in https://github.com/thesofproject/sof/blob/214f98378f12ed179bd22f0c4feb09b0b3aaea81/src/ipc/ipc-helper.c#L140 or in https://github.com/thesofproject/sof/blob/214f98378f12ed179bd22f0c4feb09b0b3aaea81/src/ipc/ipc-helper.c#L156 . Then parameters are overwritten, but buffer parameters remain as they were set in `comp_verify_params()`... At least should make it obvious in `module_adapter_params()` and only use that data for IPC3? Then, those two added lines inhttps://github.com/thesofproject/sof/commit/a25cdbb90b67be1f08c81e8de1641b7cefdf5fd5 - they fill the same data into two locations - can we just use `memcpy()` for the second one? And do we actually need to write data to `params` - it's a local variable in https://github.com/thesofproject/sof/blob/214f98378f12ed179bd22f0c4feb09b0b3aaea81/src/ipc/ipc4/handler.c#L117 and it will be used for all components in the pipeline while recursing in https://github.com/thesofproject/sof/blob/214f98378f12ed179bd22f0c4feb09b0b3aaea81/src/ipc/ipc4/handler.c#L130 so every component (if it's using the module adapter API) will just overwrite it with its own copy...
Sorry for a long and not very consistent comment, but I think we need to clean up and clarify a bit before we add more complexity to parameter update and propagation flow

lyakh (changes requested):
Can we hold on with this a bit until we clarify my doubts about parameter setting. I'll give this -1 for now just to make sure that we don't forget to answer these questions before merging

src/audio/aria/aria.c
```diff
@@ -188,21 +188,9 @@ static void aria_free(struct comp_dev *dev)
...
-				    &frame_fmt, &valid_fmt,
-				    cd->base.audio_fmt.s_type);
-	audio_stream_set_buffer_fmt(&buffer_c->stream,
-				    cd->base.audio_fmt.interleaving_style);
-	audio_stream_set_frm_fmt(&buffer_c->stream, frame_fmt);
-	audio_stream_set_valid_fmt(&buffer_c->stream, valid_fmt);
-	audio_stream_set_channels(&buffer_c->stream, cd->chan_cnt);
-	audio_stream_set_rate(&buffer_c->stream, cd->base.audio_fmt.sampling_frequency);
-
+	ipc4_update_buffer_format(buffer_c, &cd->base.audio_fmt);
```
singalsu, line 193: Nice, I didn't know about this function! It helps to simplify other module adapter conversion patches.

src/audio/eq_fir/eq_fir.c
btian1: @lyakh , @kv2019i  really thanks you bring so much for this topic, how to say, like I may have similar feelings like you, just can't change too much a time(CI need pass), I have below things want to clarify:

1. Nobody said the logic can't change, yes, params are changing compared with previous.
2. there are also other functions to change params, like comp_verify_params, comp_update_params.

so, I will change first when there is a reason and check CI and then your comments, somehow, CI is first gold criteria for me.

I can remove these part of changes first, and then discuss further for the actual changes to unblock the PR merge, let me know what do you think.

src/audio/module_adapter/module/generic.c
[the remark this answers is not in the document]
btian1: @dbaluta , will IPC3 also support cadence codecs?

src/audio/eq_fir/eq_fir.c
lyakh: @btian1 in fact I'd even try to make a PR to clean up and clarify my above comments and to add any of your fixes to that parameter handling code, if needed. And then update this PR on top of that one. I think that order would be better because fixes should come before optimisations. Alternatively you could still remove those changes from this PR and add a PR after this one, but that would probably (1) first add more confusion to the current situation in this PR, and (2) make it then more difficult to fix it in a follow-up.

btian1: ok, let me remove params related patches as first step to make this PR can be reviewed and merged, and we can discuss later for params related changes.

lyakh (approved):

src/audio/module_adapter/module/generic.c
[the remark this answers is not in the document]
dbaluta: @btian1  yes!

dbaluta:
@btian1 will give it a test today!

src/audio/module_adapter/module/generic.c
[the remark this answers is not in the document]
btian1: so we can't remove those ipc3 specific code? @dbaluta @lyakh @ranj063, since ipc3 need it? from my observation, I did not see those code is must even for IPC3

dbaluta (changes requested):
This PR doesn't compile with IPC3

```
/home/nxf25322/work/sof_dir/sof/src/audio/module_adapter/module/cadence.c: In function 'cadence_codec_init':
/home/nxf25322/work/sof_dir/sof/src/audio/module_adapter/module/cadence.c:356:11: error: 'struct module_data' has no member named 'state'
  if (codec->state == MODULE_DISABLED && codec->cfg.avail) {
           ^~
/home/nxf25322/work/sof_dir/sof/src/audio/module_adapter/module/cadence.c:356:22: error: 'MODULE_DISABLED' undeclared (first use in this function); did you mean 'LOG_DISABLE'?
  if (codec->state == MODULE_DISABLED && codec->cfg.avail) {
                      ^~~~~~~~~~~~~~~
                      LOG_DISABLE
/home/nxf25322/work/sof_dir/sof/src/audio/module_adapter/module/cadence.c:356:22: note: each undeclared identifier is reported only once for each function it appears in
/home/nxf25322/work/sof_dir/sof/src/audio/module_adapter/module/cadence.c: In function 'cadence_codec_set_configuration':
/home/nxf25322/work/sof_dir/sof/src/audio/module_adapter/module/cadence.c:855:22: error: unused variable 'md' [-Werror=unused-variable]
                      ^~
make[2]: *** [CMakeFiles/sof.dir/build.make:1123: CMakeFiles/sof.dir/src/audio/module_adapter/module/cadence.c.o] Error 1

```

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -109,22 +109,7 @@ static int google_rtc_audio_processing_params(
...
-		audio_stream_set_channels(&sink_c->stream, cd->config.output_fmt.channels_count);
-		audio_stream_set_rate(&sink_c->stream, cd->config.output_fmt.sampling_frequency);
-
-		sink_c->buffer_fmt = out_fmt->interleaving_style;
-		params->frame_fmt = valid_fmt;
-
-		sink_c->hw_params_configured = true;
-
+		ipc4_update_buffer_format(sink_c, out_fmt);
 		buffer_release(sink_c);
```
dbaluta, line 113: Here you assume that google rtc audio always uses IPC4 right?

src/audio/module_adapter/module/generic.c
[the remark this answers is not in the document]
dbaluta: @btian1 we can't remove IPC3 specific code as IMX uses it. 

This patch works fine for imx cadence tests. You just need to do the following changes:

```
diff --git a/src/audio/module_adapter/module/cadence.c b/src/audio/module_adapter/module/cadence.c
index ef912d7e2..dd0654228 100644
--- a/src/audio/module_adapter/module/cadence.c
+++ b/src/audio/module_adapter/module/cadence.c
@@ -353,7 +353,7 @@ static int cadence_codec_init(struct processing_module *mod)
        codec->mpd.init_done = 0;
 
        /* copy the setup config only for the first init */
-       if (codec->state == MODULE_DISABLED && codec->cfg.avail) {
+       if (codec->cfg.avail) {
                setup_cfg = &cd->setup_cfg;
 
                /* allocate memory for set up config */
@@ -852,7 +852,6 @@ cadence_codec_set_configuration(struct processing_module *mod, uint32_t config_i
                                const uint8_t *fragment, size_t fragment_size, uint8_t *response,
                                size_t response_size)
 {
-       struct module_data *md = &mod->priv;
        int ret;
```

lgirdwood (commented):
@btian1 it may be simpler here for any IPC changes to split between files i.e.
we could have file-generic.c (core and generic feature code) with file-ipc3.c and file-ipc4.c (for IPC specifc feature adaptions).

src/audio/module_adapter/module/generic.c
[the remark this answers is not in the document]
btian1: @dbaluta yes, I did it last night, if this patch works fine for imx ipc3 cadence codec test, means md->state can be removed for both ipc4 and ipc3, if so, could you add your approval to this PR?

@lgirdwood ack, due to CI does not check ipc3 build, some place change are ignored, I added last night.

This change is not a irrelevant change, it is critical for host performance in #7909

btian1:
> 

ok, if can't remove it, we have to split it to two files with ipc3 / ipc4 prefix file.

btian1:
I will split one patch with ipc3 and ipc4, please stay tuned for next round review.

lyakh:
> ok, if can't remove it, we have to split it to two files with ipc3 / ipc4 prefix file.

@btian1 sorry, could you explain? Do I understand it correctly, that you want to split a file into two parts because you cannot remove the `mod->priv.state` field? Isn't that already mostly encapsulated in `#if CONFIG_...IPC...3` so you can just add a couple of missing ones and the effect of handling that `.state` field should anyway be negligible?

lgirdwood:
> > ok, if can't remove it, we have to split it to two files with ipc3 / ipc4 prefix file.
> 
> @btian1 sorry, could you explain? Do I understand it correctly, that you want to split a file into two parts because you cannot remove the `mod->priv.state` field? Isn't that already mostly encapsulated in `#if CONFIG_...IPC...3` so you can just add a couple of missing ones and the effect of handling that `.state` field should anyway be negligible?

You need to do whatever you think will be best here for simplicity, readability, velocity - it could be separate files for ipc OR could be using ifdefs. The main point is you cant break IPC3 and the code must make sense, be maintainable, readable etc.

Reviewers can then help align and fine tune the PR so it's acceptable to everyone.

btian1:
@lgirdwood @dbaluta @ranj063  I removed the md->state patch, and I plan submit another pr to cover audio/src/module_adapter/module related ipc3/ipc4 split, please review again, this pr should be ready for merge.

lgirdwood (approved):

lgirdwood:
@dbaluta good for you now ?

juimonen (approved):
changes look fine with me if CI is happy...

src/audio/copier/copier_dai.c
```diff
@@ -31,18 +31,18 @@ static uint32_t bitmask_to_nibble_channel_map(uint8_t bitmask)
 	return nibble_map;
 }
 
-static int copier_set_alh_multi_gtw_channel_map(struct comp_dev *parent_dev,
```
ranj063, line 34: a part of this is in the previous commit. Can you please reorganize to move all of the same change to the same patch?

src/audio/copier/copier_dai.c
```diff
@@ -99,8 +99,6 @@ static int copier_dai_init(struct comp_dev *dev,
 		pipeline->source_comp = dev;
 	}
 
-	dev->ipc_config.frame_fmt = config->frame_fmt;
```
ranj063: this needs to be a separate patch with an explanation for why it is not needed anymore

src/audio/copier/copier_dai.c
```diff
@@ -265,7 +265,7 @@ int copier_dai_create(struct comp_dev *dev, struct copier_data *cd,
 
 	/* create multi_endpoint_buffer for ALH multi-gateway case */
 	if (dai_count > 1) {
-		ret = create_endpoint_buffer(dev, cd, config, copier, type, true, 0);
```
ranj063: @btian it will be better if you split this PR into 2 , one with the copier clean up and the other for using the update_buffer_format function for the other modules

src/audio/copier/copier.c
```diff
@@ -101,8 +101,7 @@ static int copier_init(struct processing_module *mod)
 		switch (node_id.f.dma_type) {
 		case ipc4_hda_host_output_class:
 		case ipc4_hda_host_input_class:
-			ret = copier_host_create(dev, cd, &dev->ipc_config,
```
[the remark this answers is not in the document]
btian1: done

src/audio/copier/copier_dai.c
btian1: done

src/audio/copier/copier_dai.c
btian1: done

src/audio/copier/copier_dai.c
btian1: outdated, all changes should be done now.

ranj063 (approved):

iuliana-prodan (approved):

lgirdwood:
@dbaluta good for you - assuming so as @iuliana-prodan has approved ?

cujomalainey (approved):

btian1:
> @dbaluta good for you - assuming so as @iuliana-prodan has approved ?

yes, Prodan helped verify and approved, since Baluta is taking annual leave.

kv2019i:
https://sof-ci.01.org/sofpr/PR7958/build11365/devicetest/index.html and https://sof-ci.01.org/sofpr/PR7958/build11366/devicetest/index.html are ok, with known fails only. With @iuliana-prodan 's ack in place of @dbaluta , I'll proceed with merge.

PR #7960, opened 2023-07-18
topology2: fix conditional inclusion of DMIC snippets for nocodec topology

aiChaoSONG:
DMIC now is controlled by two variables, PASSHTHROUGH and NUM_DMICS,
we should consider the values for both variables to decide the inclusion of DMIC
pipelines, routes, widgets, etc.

Fixes: #7931

Without this PR, there are DMIC components declared in the topology even with NUM_DMICS=0, and our tplgtool.py will fail to parse the topology because of a widget apprears in SectionRoute, but not in SectionWidget. Or there is an incomplete and isolated route in the topology graph.

### DIFF
--- a/tools/topology/topology2/cavs-nocodec-multicore.conf
+++ b/tools/topology/topology2/cavs-nocodec-multicore.conf
@@ -300,184 +300,189 @@ Object.Pipeline.io-gateway-capture [
 
 IncludeByKey.PASSTHROUGH {
 "false" {
-		Object.Pipeline.gain-module-copier [
-			{
-				index		19
-				core_id $DMIC_CORE_ID
-				direction	"capture"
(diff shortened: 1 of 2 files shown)


ranj063 (approved):

marc-hb:
> our tplgtool.py will fail to parse the topology because of a widget apprears in SectionRoute, but not in SectionWidget. Or there is an incomplete and isolated route in the topology graph.

I have plans to run this tool in CI; would it have helped here?

RanderWang (approved):

RanderWang:
@kv2019i  Please help to merge it so we can get latest status from CI test. Then I will merge cavs-nocodec-multicore.conf setting to cavs-nocodec.conf. Thanks!

PR #7962, opened 2023-07-18
ipc4: multi pipeline set state on different cores

abonislawski:
This will allow to correctly set multiple pipelines state even if they are allocated on different cores.

ipc4_set_pipeline_state will check if several cores are involved
- set ppl state if only current core requested
- process IPC on another core if only single secondary core requested
- send IDC messages if several secondary cores involved

Cherry-pick from mtl-005 branch


src/ipc/ipc4/handler.c
```diff
@@ -435,15 +452,23 @@ static int ipc4_set_pipeline_state(struct ipc4_message_request *ipc4)
...
-		ipc_compound_pre_start(state.primary.r.type);
-		ret = set_pipeline_state(ppl_icd, cmd, &delayed);
-		ipc_compound_post_start(state.primary.r.type, ret, delayed);
+		if (!cpu_is_me(ppl_icd->core)) {
+			if (use_idc) {
+				struct idc_msg msg = { IDC_MSG_PPL_STATE,
+					IDC_MSG_PPL_STATE_EXT(ppl_id[i]), ppl_icd->core, sizeof(cmd),
+					&cmd, };
+
+				ret = idc_send_msg(&msg, IDC_BLOCKING);
```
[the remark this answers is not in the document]
btian1: line 473-474, althrough have a long distance to reach.

src/ipc/ipc4/handler.c
```diff
@@ -404,6 +404,8 @@ static int ipc4_set_pipeline_state(struct ipc4_message_request *ipc4)
 	struct ipc *ipc = ipc_get();
 	uint32_t cmd, ppl_count, id;
 	const uint32_t *ppl_id;
+	bool use_idc = false;
```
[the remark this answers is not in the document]
kv2019i: ... it's also about IDC usage as that's what in the end happens. Cause-versus-effect , I'm happy with both namings.

kv2019i (approved):

src/idc/idc.c
```diff
@@ -265,6 +265,33 @@ static int idc_reset(uint32_t comp_id)
...
 }
 
+/**
+ * \brief Executes IDC pipeline set state message.
+ * \param[in] ppl_id Pipeline id to be triggered.
+ * \return Error code.
+ */
+static int idc_ppl_state(uint32_t ppl_id)
+{
+#if CONFIG_IPC_MAJOR_4
```
abonislawski, line 275: zephyr + tgl IPC**3** failed so I added this check

src/ipc/ipc4/handler.c
```diff
@@ -425,6 +427,21 @@ static int ipc4_set_pipeline_state(struct ipc4_message_request *ipc4)
...
+		}
+
+		if (i) {
+			if (ppl_icd->core != idx)
+				use_idc = true;
+		} else {
+			idx = ppl_icd->core;
+		}
+	}
+
```
ranj063, line 444: can we simplify the loop below to do this right after the first one? 
```
	/* execute IPC on the secondary core if all pipelines are on the same core */
	if (!use_idc && !cpu_is_me(idx))
		return ipc4_process_on_core(idx, false);
```

lyakh:
> ... it's also about IDC usage as that's what in the end happens. Cause-versus-effect , I'm happy with both namings.

@kv2019i the thing is, that in the other case it's also using IDC - when all of the pipelines in a set are processed on a different core and `ipc_process_on_core()` is used, so the flag isn't doing what its name is suggesting - it isn't distinguishing "use IDC" against "don't use IDC" cases

src/ipc/ipc4/handler.c
```diff
@@ -425,6 +427,21 @@ static int ipc4_set_pipeline_state(struct ipc4_message_request *ipc4)
...
+			tr_err(&ipc_tr, "ipc: comp %d not found", ppl_id[i]);
+			return IPC4_INVALID_RESOURCE_ID;
+		}
+
+		if (i) {
+			if (ppl_icd->core != idx)
+				use_idc = true;
+		} else {
+			idx = ppl_icd->core;
+		}
```
btian1, line 442: if else parentheses should not need

lgirdwood: I think this is good here for readability.

lgirdwood (approved):

src/ipc/ipc4/handler.c
lyakh: it is not only for readability, without them the `else` would become ambiguous, IIRC the compiler issues a warning in such cases, and in fact it would be misinterpreted as referring to the innermost `if`, i.e. removing braces here would actually mean
```
	if (i)
		if (ppl_icd->core != idx) {
			use_idc = true;
		} else {
			idx = ppl_icd->core;
		}
```
which would be very different to what is intended.

marcinszkudlinski (approved):

pjdobrowolski (approved):

serhiy-katsyuba-intel (approved):

lgirdwood:
@abonislawski conflicts

serhiy-katsyuba-intel:
Adrian is on vacation this week. The conflict was with my commit merged half an hour ago. I resolved the conflict here on GitHub, however, this resulted in addition merge commit, which, probably, will not look nice in history. So let's wait for Adrian to rebase its commit and update the PR.

src/ipc/ipc4/handler.c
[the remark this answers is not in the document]
lyakh: @kv2019i sorry, I meant, that IDCs are used in both cases, just that in one case the use is explicit and in the other case it's hidden under Zephyr API calls. So, using IDCs isn't a differentiating factor between the cases

src/ipc/ipc4/handler.c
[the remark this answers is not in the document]
lyakh: still `idc_send_msg()` returns POSIX error codes, so they cannot be passed up to the caller here.

lgirdwood (commented):
@abonislawski can you rebase. Thanks !

juimonen:
closing as merged in #7995

PR #7964, opened 2023-07-18
ace: copier: hda: synchronized FPI updates of HD-A gateways

tmleman:
This PR adds synchronized FPI updates of HD-A gateways. Driver can define group of such gateways via module init IPC. The driver may also specify an update period for each group, different than the default one determined by the system tick frequency.

In addition, a change has been introduced that makes it possible to read config size in module init function.  Each module receive its own configuration at creation (MODULE_INIT IPC). In case of a gateway those configurations can contain additional value (aux_conf). The only way to check if such config was received is to compare size of the data received via ipc and standard configuration size.

### DIFF
--- a/src/audio/Kconfig
+++ b/src/audio/Kconfig
@@ -32,6 +32,16 @@ config HOST_DMA_RELOAD_THRESHOLD
 	  The DMA buffer threshold in milliseconds to trigger host DMA
 	  reloading.
 
+config HOST_DMA_STREAM_SYNCHRONIZATION
+	bool "Stream DMA Transfers Synchronization"
+	default y if ACE
+	default n
+	help
+	  Enable synchronized Firmware Pointer Increment (FPI) register updates of HD-A gateways
+	  belonging to a group defined by the driver. The driver may also specify an update period
+	  for each group, different than the default one determined by the system tick frequency.
+	  This feature will allow host lower power consumption in scenarios with deep buffering.
+
 config COMP_DAI
 	bool "DAI component"
 	default y
--- a/src/audio/copier/copier_host.c
+++ b/src/audio/copier/copier_host.c
@@ -6,11 +6,90 @@
 
 #include <sof/audio/component_ext.h>
 #include <sof/audio/host_copier.h>
+#include <sof/tlv.h>
 #include <sof/trace/trace.h>
 #include <ipc4/copier.h>
 #include <sof/audio/module_adapter/module/generic.h>
 
 LOG_MODULE_DECLARE(copier, CONFIG_SOF_LOG_LEVEL);
+
+#if CONFIG_HOST_DMA_STREAM_SYNCHRONIZATION
+
+/* NOTE: The code that modifies and checks the contents of the list can be executed on different
+ * cores. But since adding and removing modules is done exclusively via IPC, and we never add or
+ * remove more than one module, these functions do not require synchronization.
+ */
+struct fpi_sync_group {
+	uint32_t id;
+	uint32_t period;
+	uint32_t ref_count;
+	struct list_item item;
+};
+
+static struct list_item group_list_head = LIST_INIT(group_list_head);
+
+static struct fpi_sync_group *find_group_by_id(uint32_t id)
+{
+	struct list_item *item;
+
+	list_for_item(item, &group_list_head) {
+		struct fpi_sync_group *group = list_item(item, struct fpi_sync_group, item);
+
+		if (group->id == id)
+			return group;
+	}
+
+	return NULL;
+}
+
+int add_to_fpi_sync_group(struct comp_dev *parent_dev,
+			  struct host_data *hd,
+			  struct ipc4_copier_sync_group *sync_group)
+{
+	struct fpi_sync_group *group = find_group_by_id(sync_group->group_id);
+
+	if (group) {
+		if (group->period != sync_group->fpi_update_period_usec) {
+			comp_err(parent_dev, "incorrect period %u for group %u (currently %u)",
+				 group->id, group->period, sync_group->fpi_update_period_usec);
+			return -EINVAL;
+		}
+
+		group->ref_count++;
+	} else {
+		group = rzalloc(SOF_MEM_ZONE_RUNTIME_SHARED, 0, SOF_MEM_CAPS_RAM, sizeof(*group));
+		if (!group) {
+			comp_err(parent_dev, "Failed to alloc memory for new group");
+			return -ENOMEM;
+		}
+
+		group->id = sync_group->group_id;
+		group->period = sync_group->fpi_update_period_usec;
+		group->ref_count = 1;
+		list_item_append(&group->item, &group_list_head);
+	}
+
+	hd->is_grouped = true;
+	hd->group_id = group->id;
+	hd->period_in_cycles = k_us_to_cyc_ceil64(group->period);
+	comp_dbg(parent_dev, "gtw added to group %u with period %u", group->id, group->period);
+	return 0;
+}
+
+void delete_from_fpi_sync_group(struct host_data *hd)
+{
+	struct fpi_sync_group *group = find_group_by_id(hd->group_id);
+
+	if (!group)
+		return;
+
+	group->ref_count--;
+	if (group->ref_count == 0) {
+		list_item_del(&group->item);
+		rfree(group);
+	}
+}
+#endif
 
 /* Playback only */
 static int init_pipeline_reg(struct comp_dev *dev)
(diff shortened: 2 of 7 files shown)


src/include/sof/audio/component.h
```diff
@@ -544,15 +544,16 @@ struct comp_driver_info {
  * Audio component base configuration from IPC at creation.
  */
 struct comp_ipc_config {
-	uint32_t core;		/**< core we run on */
```
btian1, line 547: please double check the patch comments, there may have some spell mistakes, recive --> receive, thos --> those?

lyakh (changes requested):

src/audio/host-zephyr.c
```diff
@@ -433,6 +433,29 @@ static uint32_t host_get_copy_bytes_normal(struct host_data *hd, struct comp_dev
...
 
+#if CONFIG_HOST_DMA_STREAM_SYNCHRONIZATION
+static inline bool stream_sync(struct host_data *hd, struct comp_dev *dev)
+{
+	if (!hd->is_grouped)
+		return true;
+
+	uint64_t current_time = k_cycle_get_64();
+
+	if (current_time >= hd->next_sync) {
```
kv2019i, line 444: Having timer logic without wrap-around handling is always a bit iffy, but given we have explicit 64bit counter, it would seem this is safe. Even with a 10x faster system clock, this would still take thousands of years to wrap around. I think current uptime record holder is still the one Cisco switch that is around 26 years now, so we have some safety margin -> https://en.wikipedia.org/wiki/Uptime

kv2019i (approved):
Copier is getting complicated, but nothing really to complain about in the patch. It seems copier is the right place for this, and the extension handling looks correct. I have one minor nit on whether the 64bit wrap-around should be handled or not, but that's a minor thing.

src/audio/host-zephyr.c
lgirdwood: ack, worth an inline comment here so that no one tries to fix later on.

src/audio/copier/copier_host.c
```diff
@@ -78,6 +152,34 @@ int copier_host_create(struct comp_dev *parent_dev, struct copier_data *cd,
...
+		const uint32_t tlv_addr = (uint32_t)copier_cfg + basic_size;
+		const uint32_t tlv_size = config->ipc_config_size - basic_size;
+		uint32_t value_size = 0;
+		void *value_ptr = NULL;
+
+		tlv_value_get((void *)tlv_addr, tlv_size, HDA_SYNC_FPI_UPDATE_GROUP,
+			      &value_ptr, &value_size);
+
+		if (value_ptr) {
+			struct ipc4_copier_sync_group *sync_group = NULL;
```
lyakh, line 170: superfluous initialisation

src/audio/copier/copier_host.c
```diff
@@ -78,6 +152,34 @@ int copier_host_create(struct comp_dev *parent_dev, struct copier_data *cd,
...
+
+			if (value_size != sizeof(struct ipc4_copier_sync_group))
+				return -EINVAL;
+
+			sync_group = (struct ipc4_copier_sync_group *)((void *)value_ptr);
+
+			ret = add_to_fpi_sync_group(parent_dev, hd, sync_group);
+			if (ret != 0)
```
btian1, line 178: if (ret < 0)?

lgirdwood (commented):
@tmleman only minor things here and we are good to merge.

lyakh:
> @tmleman only minor things here and we are good to merge.

@lgirdwood I think https://github.com/thesofproject/sof/pull/7964#discussion_r1267912877 is rather significant

src/audio/Kconfig
```diff
@@ -32,6 +32,16 @@ config HOST_DMA_RELOAD_THRESHOLD
 	  The DMA buffer threshold in milliseconds to trigger host DMA
 	  reloading.
 
+config HOST_DMA_STREAM_SYNCHRONIZATION
+	bool "Stream DMA Transfers Synchronization"
+	default y if ACE
+	default n
+	help
+	  Enable synchronized FPI updates of HD-A gateways belonging to a group defined by the
```
lyakh, line 40: Maybe worth putting "Buffer Position Increment (FPI)" here once explicitly

src/audio/copier/copier_host.c
```diff
@@ -6,12 +6,86 @@
...
+struct fpi_sync_group {
+	uint32_t id;
+	uint32_t period;
+	uint32_t ref_count;
+	struct list_item item;
+};
+
+static struct list_item group_list_head = LIST_INIT(group_list_head);
+
+struct fpi_sync_group *find_group_by_id(uint32_t id)
```
lyakh, line 26: since it isn't in any header, please make `static`, this would also explain, that it's only used when adding and deleting groups, i.e. only in IPC thread context, so no locking is needed... supposedly

lgirdwood:
@tmleman can you check the internal CI results. Thanks !

src/include/sof/audio/component.h
```diff
@@ -544,15 +544,16 @@ struct comp_driver_info {
...
 	enum sof_comp_type type;	/**< component type */
-	uint32_t periods_sink;	/**< 0 means variable */
-	uint32_t periods_source;/**< 0 means variable */
-	uint32_t frame_fmt;	/**< SOF_IPC_FRAME_ */
-	uint32_t xrun_action;	/**< action we should take on XRUN */
+	uint32_t periods_sink;		/**< 0 means variable */
+	uint32_t periods_source;	/**< 0 means variable */
+	uint32_t frame_fmt;		/**< SOF_IPC_FRAME_ */
+	uint32_t xrun_action;		/**< action we should take on XRUN */
+	uint32_t ipc_config_size;	/**< size of a config received by ipc */
```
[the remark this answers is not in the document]
btian1: can comp_ipc_config be changed? not sure why not left some redundancy space for later adding, like uint32_t valid_fmt to align internal usage.

src/audio/copier/copier_host.c
```diff
@@ -6,12 +6,86 @@
...
 
 LOG_MODULE_DECLARE(copier, CONFIG_SOF_LOG_LEVEL);
 
+#if CONFIG_HOST_DMA_STREAM_SYNCHRONIZATION
+struct fpi_sync_group {
+	uint32_t id;
+	uint32_t period;
+	uint32_t ref_count;
+	struct list_item item;
+};
```
btian1, line 22: move to host_copier.h?

src/audio/copier/copier_host.c
btian1: should include all new added functions.

src/include/sof/audio/component.h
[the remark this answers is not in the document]
tmleman: @btian1 can you explain exactly what you mean? Sorry, but I don't quite understand what your comment means.

src/audio/Kconfig
tmleman: > is this "FPI" - the same?

Yes.

> But shouldn't it be BPI or BFPI

**BPI** No, **BFPI** more likely

**DGBFPI** stands for **D**SP **G**ateway **B**uffer **F**irmware **P**ointer **I**ncrement. Which is probably the generic name for the registers of Link Input, Link Output, Host Input and Host Output.

This three-letter acronym (FPI) appears in many places in the documentation. I think just expanding the acronym will be enough.

tmleman:
23ww31.3 Update: rebase

src/include/ipc4/copier.h
```diff
@@ -119,6 +119,14 @@ enum ipc4_copier_features {
 	IPC4_COPIER_FAST_MODE = 0
 };
 
+#if CONFIG_HOST_DMA_STREAM_SYNCHRONIZATION
+#define HDA_SYNC_FPI_UPDATE_GROUP 0
+struct ipc4_copier_sync_group {
+	uint32_t group_id;
+	uint32_t fpi_update_period_usec;
+} __packed __aligned(4);
```
btian1, line 127: for unint32 started structure, compiler should always allocated a 4 aligned address, right?
I mean __aligned(4) is redundant.

tmleman:
23ww31.5 Update: removing the ugly workaround and replacing it with an implementation resistant to incorrect values in IPC payload.

src/include/sof/audio/component.h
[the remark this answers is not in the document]
tmleman: From what I saw in code `struct comp_ipc_config` is internal SOF struct and its not expose to driver. It's created on FW side during module initialization and therefore I assume that is safe to add new value to it.

https://github.com/thesofproject/sof/blob/main/src/ipc/ipc4/helper.c#L81-L105

src/include/ipc4/copier.h
lgirdwood: @btian1 its best to be explicit as we will be compiling on various 32/64 bit compilers.

lgirdwood (approved):

PR #7966, opened 2023-07-18
Dsm adapter fixes

cujomalainey:
Fixes unneeded depends and include pathing for zephyr build

### DIFF
--- a/src/audio/smart_amp/Kconfig
+++ b/src/audio/smart_amp/Kconfig
@@ -4,7 +4,6 @@ menu "Smart amplifier solutions"
 	visible if COMP_SMART_AMP
 
 	config MAXIM_DSM
-		depends on CAVS
 		bool "Maxim DSM solution"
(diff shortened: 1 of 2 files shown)


btian1 (approved):
please check QB result, need perform again, seems test board lost connect.

cujomalainey:
> please check QB result, need perform again, seems test board lost connect.

This change should be a no-op for CI without the dependent change so CI can be ignored except for contextual checks

marc-hb (approved):
> This change should be a no-op for CI without the dependent change so CI can be ignored except for contextual checks

Only suspend/resume failures in https://sof-ci.01.org/sofpr/PR7966/build10937/devicetest/index.html

https://sof-ci.01.org/sof-pr-viewer/#/build/PR7966/build12449243 is green.

Everything else is green

kv2019i (approved):

PR #7972, opened 2023-07-20
zephyr: s/MP_NUM_CPUS/MP_MAX_NUM_CPUS/

ceolin:
Zephyr uses MP_MAX_NUM_CPUS internally to represent the number of cores available and consequently to allocate resources. It is even expected, and checked through assert, that these two symbols have the same value. Use different value can lead to an undesired behavior, so lets use MP_MAX_NUM_CPUS.

### DIFF
--- a/src/platform/Kconfig
+++ b/src/platform/Kconfig
@@ -221,7 +221,7 @@ config MAX_CORE_COUNT
 
 config CORE_COUNT
 	int "Number of cores"
-	default MP_NUM_CPUS if KERNEL_BIN_NAME = "zephyr"
+	default MP_MAX_NUM_CPUS if KERNEL_BIN_NAME = "zephyr"
 	default MAX_CORE_COUNT
 	range 1 MAX_CORE_COUNT
 	help
--- a/zephyr/lib/cpu.c
+++ b/zephyr/lib/cpu.c
@@ -24,7 +24,7 @@
 
 #if CONFIG_MULTICORE && CONFIG_SMP
 
-extern K_KERNEL_STACK_ARRAY_DEFINE(z_interrupt_stacks, CONFIG_MP_NUM_CPUS,
+extern K_KERNEL_STACK_ARRAY_DEFINE(z_interrupt_stacks, CONFIG_MP_MAX_NUM_CPUS,
 				   CONFIG_ISR_STACK_SIZE);
 
 static atomic_t start_flag;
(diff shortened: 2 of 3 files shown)


sofci:

gkbldcig:

lyakh:
test this please

lyakh (commented):
@ceolin could you explain a bit more, please? Is it planned to deprecate one of the config options? If not - I could never really understand, why would one care about the maximum number of cores if one wants to build Zephyr to only support a certain core number? If it's about core IDs, then we probably should explain this. Currently they're defined as
```
config MP_NUM_CPUS
	int "Number of CPUs/cores"
	default MP_MAX_NUM_CPUS
	range 1 4
	help
	  Number of multiprocessing-capable cores available to the
	  multicpu API and SMP features.

config MP_MAX_NUM_CPUS
	int "Maximum number of CPUs/cores"
	default 1
	range 1 4
	help
	  Maximum number of multiprocessing-capable cores available to the
	  multicpu API and SMP features.
```
which doesn't explain a lot

kv2019i (approved):
Thanks @ceolin !

kv2019i:
I agree with @lyakh the split is difficult to understand, but that's a Zephyr matter in the end. I think it'd be a bad idea to use a different define to size e.g. structs that need X copies for every core, SOF versus Zephyr.

tmleman (approved):

ceolin:
> @ceolin could you explain a bit more, please? Is it planned to deprecate one of the config options? If not - I could never really understand, why would one care about the maximum number of cores if one wants to build Zephyr to only support a certain core number? If it's about core IDs, then we probably should explain this. Currently they're defined as
> 
> ```
> config MP_NUM_CPUS
> 	int "Number of CPUs/cores"
> 	default MP_MAX_NUM_CPUS
> 	range 1 4
> 	help
> 	  Number of multiprocessing-capable cores available to the
> 	  multicpu API and SMP features.
> 
> config MP_MAX_NUM_CPUS
> 	int "Maximum number of CPUs/cores"
> 	default 1
> 	range 1 4
> 	help
> 	  Maximum number of multiprocessing-capable cores available to the
> 	  multicpu API and SMP features.
> ```
> 
> which doesn't explain a lot

That is really confuse indeed. `MP_NUM_CPUS` has no used inside Zephyr and it is planned to be removed (I already have a pr doing it). There is even an assert there checking if these symbols have the same value. So having different values and using MP_NUM_CPUS in your application may result in something that is not really intended.

ceolin:
> I agree with @lyakh the split is difficult to understand, but that's a Zephyr matter in the end. I think it'd be a bad idea to use a different define to size e.g. structs that need X copies for every core, SOF versus Zephyr.

yep, using different values can really cause problems.

ceolin:
@kv2019i @lyakh checkpatch complains is not due these changes, but I can fix it if needed. Just let me know.

lyakh (approved):
of `CONFIG_MP_MAX_NUM_CPUS` is what we're going to be using throughout then sure.

lgirdwood (approved):

PR #7976, opened 2023-07-24
src: drivers: imx: ipc.c: Don't assert GIR twice

LaurentiuM1234:
According to MU's documentation from the i.MX TRM, setting the GIRn bit to 1 if already 1 may lead to issuing a second interrupt.

This leads to kernel errors such as:

"reply size (16) exceeds the buffer size (12)"

because of the fact that the MU IRQ is triggered prematurely and the firmware doesn't have enough time to write the reply. As such, the host will read the message it has sent to the firmware instead of its reply.

To fix this, make sure that GIRn is set to 0 before setting it 1.

For context, please see the following logs:

**dmesg_output**

**[33614.486537] sof-audio-of-imx8 556e8000.dsp: reply size: 16, cmd: 0x30030000, err: 20 (after read**)
[33614.516442] sof-audio-of-imx8 556e8000.dsp: reply size (16) exceeds the buffer size (12)
[33614.516465] sof-audio-of-imx8 556e8000.dsp: ipc tx error for 0x30030000 (msg/reply size: 16/12): -22
[33614.516468] sof-audio-of-imx8 556e8000.dsp: reply size: 16, cmd: 0x30030000, err: 20 (before read)
[33614.516478] sof-audio-of-imx8 556e8000.dsp: reply size: 12, cmd: 0x10000000, err: 0 (after read)
[33614.583866] sof-audio-of-imx8 556e8000.dsp: sof_ipc3_route_setup: route BUF4.1 -> MIXER1.0 failed
[33614.592784] sof-audio-of-imx8 556e8000.dsp: sof_ipc3_set_up_all_pipelines: route set up failed
[33614.601429] sof-audio-of-imx8 556e8000.dsp: Failed to restore pipeline after resume -22
[33614.610015] sof-audio-of-imx8 556e8000.dsp: ASoC: error at snd_soc_pcm_component_pm_runtime_get on 556e8000.dsp: -22
[33614.620572]  sai1-wm8960-hifi: ASoC: error at __soc_pcm_open on sai1-wm8960-hifi: -22
[33614.628431]  PCM: ASoC: error at dpcm_be_dai_startup on PCM: -22
[33614.634469]  PCM: ASoC: error at dpcm_fe_dai_startup on PCM: -22
[33614.640673] sof-audio-of-imx8 556e8000.dsp: ASoC: error at snd_soc_pcm_component_pm_runtime_get on 556e8000.dsp: -22
[33614.651432]  sai1-wm8960-hifi: ASoC: error at __soc_pcm_open on sai1-wm8960-hifi: -22
[33614.659543]  PCM Deep Buffer: ASoC: error at dpcm_be_dai_startup on PCM Deep Buffer: -22
[33614.667680]  PCM Deep Buffer: ASoC: error at dpcm_fe_dai_startup on PCM Deep Buffer: -22

From the first line we can see that the host has read the command issued to the firmware instead of the reply.

Since the issue is very hard to reproduce (takes a few thousand inner test iterations) it will take a bit more time to fully test this fix.

Another way to fix this may be to use imx_mu_write when setting GIRn, but since testing takes so much time, this can be addressed later on as the only benefit to this approach would be its cosmetic value.

### DIFF
--- a/src/drivers/imx/ipc.c
+++ b/src/drivers/imx/ipc.c
@@ -63,9 +63,6 @@ static void irq_handler(void *arg)
 
 	/* reply message(done) from host */
 	if (status & IMX_MU_xSR_GIPn(IMX_MU_VERSION, 1)) {
-		/* Disable GP interrupt #1 */
-		imx_mu_xcr_rmw(IMX_MU_VERSION, IMX_MU_GIER, 0, IMX_MU_xCR_GIEn(IMX_MU_VERSION, 1));
-
 		/* Clear GP pending interrupt #1 */
 		imx_mu_write(IMX_MU_xSR_GIPn(IMX_MU_VERSION, 1),
 			     IMX_MU_xSR(IMX_MU_VERSION, IMX_MU_GSR));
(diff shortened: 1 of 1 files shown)


LaurentiuM1234:
Fixed a typo in `poll_for_register_delay()`'s call which would result in the function reading the value not from the CR's address but from the address given by CR's value. This would result in a translation fault on i.MX93.

LaurentiuM1234:
Updates:

- test is done, everything seems to work as expected.

dbaluta (approved):

dbaluta:
@lgirdwood this is good to merge. We need one more approval and @iuliana-prodan is out of office. So, can you please help?

lgirdwood (approved):

lgirdwood:
@wszypelt this PR is not tested by Intel CI so it looks like a false positive. Merging.

PR #7977, opened 2023-07-24
ipc3: Deprecate invalid enums

cujomalainey:
Mark enums that are unused or never implemented as deprecated since they will never be implemented as IPC3 is legacy.


src/include/kernel/abi.h
```diff
@@ -30,7 +30,7 @@
 /** \brief SOF ABI version major, minor and patch numbers */
 #define SOF_ABI_MAJOR 3
 #define SOF_ABI_MINOR 27
-#define SOF_ABI_PATCH 0
+#define SOF_ABI_PATCH 1
```
RanderWang, line 33: Is this a bug fix ?  I prefer a minor change

RanderWang (commented):
I am not certain about the SOF_API_PATCH change.

src/include/ipc/topology.h
```diff
@@ -34,9 +34,9 @@ enum sof_comp_type {
 	SOF_COMP_MIXER,
 	SOF_COMP_MUX,
 	SOF_COMP_SRC,
-	SOF_COMP_SPLITTER,
+	SOF_COMP_DEPRECATED0, /* Formerly SOF_COMP_SPLITTER */
 	SOF_COMP_TONE,
-	SOF_COMP_SWITCH,
+	SOF_COMP_DEPRECATED1, /* Formerly SOF_COMP_SWITCH */
```
btian1, line 39: these code are totally no harmful, it neither have memory occupation, nor have code size, delete or keep both ok.

src/include/kernel/abi.h
lyakh: As commented in https://github.com/thesofproject/linux/pull/4475/commits/0876dffe3d39b9b252cf3e1ba072e63e4edbc2f6#r1273099107 - in principle this change is purely in the software, it doesn't actually change any run-time behaviour, so actually no ABI version change should be needed

src/include/ipc/topology.h
lgirdwood: we cant delete as that changes the IDs, but this change is good for me.

src/include/kernel/abi.h
lgirdwood: ack, these IDs were never used so I would be safe to drop the ABI update.

src/include/ipc/topology.h
cujomalainey: Yes, if we deleted it would change the enum and break the ABI, hence deprecation.

src/include/kernel/abi.h
cujomalainey: Got it, i will force skip the patch change check. @plbossart FYI for when I revert this on the kernel PR as well.

btian1 (approved):
seems can't, CI is failing and request you for ABI changes.

RanderWang (approved):

cujomalainey:
> seems can't, CI is failing and request you for ABI changes.

It's a patch check that can be overridden by the admins :)

plbossart (approved):

plbossart:
there's no way this PR can cause CI failures, the failures are a pre-existing condition that should be investigated separately.

lgirdwood (approved):

lgirdwood:
@lrudyX @wszypelt good to merge, not expecting CI failure here ?

wszypelt:
@lgirdwood after rebuild, good to merge :)

cujomalainey:
rebase to trigger rebuild

lgirdwood (approved):

PR #7978, opened 2023-07-25
Audio: Aria: Convert aria to module adapter interface

andrula-song:
Convert the component aria to use module adapter interface.

### DIFF
--- a/src/audio/aria/aria.c
+++ b/src/audio/aria/aria.c
@@ -28,8 +28,6 @@
 #include <stddef.h>
 #include <stdint.h>
 
-static const struct comp_driver comp_aria;
-
 LOG_MODULE_REGISTER(aria, CONFIG_SOF_LOG_LEVEL);
 
 /* these ids aligns windows driver requirement to support windows driver */
@@ -48,10 +46,9 @@ static size_t get_required_emory(size_t chan_cnt, size_t smpl_group_cnt)
 	return ALIGN_UP(num_of_ms * chan_cnt * smpl_group_cnt, 2) * sizeof(int32_t);
 }
 
-static int aria_algo_init(struct comp_dev *dev, void *buffer_desc,
+static int aria_algo_init(struct aria_data *cd, void *buffer_desc,
 			  size_t att, size_t chan_cnt, size_t smpl_group_cnt)
 {
-	struct aria_data *cd = comp_get_drvdata(dev);
 	size_t idx;
 
 	cd->chan_cnt = chan_cnt;
@@ -74,18 +71,18 @@ static int aria_algo_init(struct comp_dev *dev, void *buffer_desc,
 	return 0;
 }
 
-void aria_process_data(struct comp_dev *dev,
-		       struct audio_stream __sparse_cache *source,
-		       struct audio_stream __sparse_cache *sink,
-		       size_t frames)
+static inline void aria_process_data(struct processing_module *mod,
+				     struct audio_stream __sparse_cache *source,
+				     struct audio_stream __sparse_cache *sink,
+				     size_t frames)
 {
-	struct aria_data *cd = comp_get_drvdata(dev);
+	struct aria_data *cd = module_get_private_data(mod);
 	size_t data_size = audio_stream_frame_bytes(source) * frames;
 	size_t sample_size = audio_stream_get_channels(source) * frames;
 
 	if (cd->att) {
-		aria_algo_calc_gain(dev, INDEX_TAB[cd->gain_state + 1], source, frames);
-		cd->aria_get_data(dev, sink, frames);
+		aria_algo_calc_gain(cd, INDEX_TAB[cd->gain_state + 1], source, frames);
+		cd->aria_get_data(mod, sink, frames);
 	} else {
 		/* bypass processing gets unprocessed data from buffer */
 		cir_buf_copy(cd->data_ptr, cd->data_addr, cd->data_end,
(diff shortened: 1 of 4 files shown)


src/audio/aria/aria.c
```diff
@@ -99,35 +96,31 @@ void aria_process_data(struct comp_dev *dev,
...
 
-	dev->ipc_config = *config;
+	comp_info(dev, "aria_init()");
+
 	list_init(&dev->bsource_list);
 	list_init(&dev->bsink_list);
 
-	dcache_invalidate_region((__sparse_force void __sparse_cache *)spec, sizeof(*aria));
+	dcache_invalidate_region((__sparse_force void __sparse_cache *)mod_data->cfg.init_data,
+				 sizeof(*aria));
```
lyakh, line 115: this actually doesn't look correct. For IPC4 the invalidation is already done in `comp_new_ipc4()`, for IPC3 `.init_data` points to a buffer, allocated as cached and filled in with data in `module_load_config()`, so, by invalidating cache here we might drop data, written there?

src/include/sof/audio/aria/aria.h
```diff
@@ -92,4 +85,10 @@ struct aria_data {
 
 extern const uint8_t INDEX_TAB[];
 
+/**
+ * \brief Aria gain processing function
+ */
+void aria_algo_calc_gain(struct aria_data *cd, size_t gain_idx,
+			 struct audio_stream __sparse_cache *source, int frames);
```
lyakh, line 92: this could just stay where it was to reduce the diff

lyakh (approved):
the first comment isn't actually related to this PR, the second one is minor

src/include/sof/audio/aria/aria.h
andrula-song: I'd like keep this change since it can reduce one pointer get operation in hot code.

src/audio/aria/aria.c
andrula-song: thanks, so better to remove this line than written?

src/include/sof/audio/aria/aria.h
btian1: maybe there is mis-understanding, from code change, looks like you are adding a new api, however, it is just refresh, keep it at the origin place will make it more clear.

I am not sure one thing: inline removed in this prototype, is it possible move this function to header file totally? since it is inline function, and use OPT macro to distinguish generic and hifi version.

lyakh: > I'd like keep this change since it can reduce one pointer get operation in hot code.

Sorry, I didn't mean any factual changes, I meant only the function location - by keeping it where it is now while doing all the changes that you need to do, those changes will be easier to identify and review

src/audio/aria/aria.c
lyakh: I think yes, it should be removed, but please make it a separate PR. Is aria used anywhere with IPC3?

andrula-song: or a separate commit? No, as I know the aria only supports 32-bit container and we don't have aria before mtl.

src/include/sof/audio/aria/aria.h
andrula-song: ah, since I want to use the parameter struct aria_data *cd, and it is defined after the original location.

src/audio/aria/aria.c
```diff
@@ -137,71 +130,48 @@ static int init_aria(struct comp_dev *dev, const struct comp_ipc_config *config,
...
-				    cd->base.audio_fmt.s_type);
+				    audio_fmt->s_type);
 	audio_stream_set_buffer_fmt(&buffer_c->stream,
-				    cd->base.audio_fmt.interleaving_style);
+				   audio_fmt->interleaving_style);
 	audio_stream_set_frm_fmt(&buffer_c->stream, frame_fmt);
 	audio_stream_set_valid_fmt(&buffer_c->stream, valid_fmt);
 	audio_stream_set_channels(&buffer_c->stream, cd->chan_cnt);
-	audio_stream_set_rate(&buffer_c->stream, cd->base.audio_fmt.sampling_frequency);
+	audio_stream_set_rate(&buffer_c->stream, audio_fmt->sampling_frequency);
```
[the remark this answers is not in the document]
andrula-song: thanks, but I don't need set the buffer_c->chmap

src/include/sof/audio/aria/aria.h
lyakh: the structure is defined? Well, maybe moving the structure definition up will create a smaller diff. Or maybe just forward-declaring it would be enough, but that's unlikely, since you actually access its internals

src/audio/aria/aria.c
lyakh: if it's IPC4 only, then yes, it's a smaller change and just a separate commit should be enough

app/overlays/tgl-h/ipc4_overlay.conf
```diff
@@ -12,6 +12,7 @@ CONFIG_PCM_CONVERTER_FORMAT_S16_C32_AND_S24_C32=y
 CONFIG_PCM_CONVERTER_FORMAT_S16_C32_AND_S16_C32=y
 CONFIG_COMP_VOLUME_WINDOWS_FADE=y
 CONFIG_COMP_DRC=y
+CONFIG_COMP_ARIA=y
```
btian1, line 15: is this for chrome and linux only？ if there is no real usage for aria in chrome and linux, keep as N is reasonable.

andrula-song: I would like to enable it since we have already add support in .tmol file, and we will need this component in the next generation device

lgirdwood (approved):

lgirdwood:
@btian1 good for you ?
@andrula-song can you check internal CI result. Thanks !

andrula-song:
hi @wszypelt , could you please trigger the internal CI again? seems the fw load failed.

app/overlays/tgl-h/ipc4_overlay.conf
btian1: commit message is for tgl device, and tgl device does not need aria, right?
which device need which feature is pre-defined, if mtl need aria, we can enable it in mtl config, not for tgl.

I guess enable it in tgl just for convenient test purpose, maybe @kv2019i @lgirdwood  can have some input.

no harmful, just add more code to tgl fw.

src/include/sof/audio/aria/aria.h
btian1: add one more line with:
struct aria_data;
in front of this function declaration is fine.

this can reduce the code diff.

wszypelt:
@andrula-song done, all green :)

app/overlays/tgl-h/ipc4_overlay.conf
andrula-song: okay, remove from this PR, we can enable it later in another PR.

btian1 (approved):

andrula-song:
the failed cases are from zephyr lib, seems not related with my PR.

PR #7982, opened 2023-07-25
imx: memory: Fix mixer going out of memory

dbaluta:
Commit 05871a16de0d ("pipeline2.0: add source/sink api to audio_stream") added two new fields to `struct audio_stream`:
```
 struct audio_stream {
+       struct sof_source source_api;
+       struct sof_sink sink_api;
```
and because `struct audio_stream` is part of `struct comp_buffer` we see an increase of `struct comp_buffer` size from 256 to 384.

With this modification, i.MX mixer use case goes out of memory:
```
c0 dma-trace src/lib/alloc.c:765    ERROR failed to alloc 0x180 bytes zone 0x4 caps 0x1 flags 0x0
c0 dma-trace src/audio/buffer.c:51  ERROR buffer_alloc(): could not alloc structure
```
This is not ideal. Anyhow, it was necessary for new source/sink API.

Fix this by increasing the number of 512-byte heap blocks to 32.


lgirdwood (approved):
@dbaluta fwiw, we still have some structures that have both xtos and zephyr specific data that are being optimized. Also this type of fix wont be needed for Zephyr :)

dbaluta:
@lgirdwood yes, this is not needed for Zephyr!

LaurentiuM1234 (approved):

btian1 (approved):
I noticed intel platform this number is 8, and mixer size increase is 128 bytes, possible to change it to 16?, anyway, it is imx platform, depend on you!

iuliana-prodan (approved):

PR #7984, opened 2023-07-26
topology2: ssp: add quirks for NHLT table endpoints

brentlu:
Two quirk flags are added. The quirk "bt_sideband" will set the device type of corresponding NHLT endpoints to BT(4) and others to Codec(0).

The other quirk "render_feedback" is for speaker path with echo reference or IV sense feedback. The endpoint direction will use Render with loopback(3) and Feedback for render(4) instead of Render(0) and Capture(1).

### DIFF
--- a/tools/topology/topology2/include/dais/ssp.conf
+++ b/tools/topology/topology2/include/dais/ssp.conf
@@ -91,9 +91,17 @@ Class.Dai."SSP" {
 		constraints {
 			!valid_values [
 				"lbm_mode"
+				"bt_sideband"
+				"render_feedback"
+				"lbm_mode,bt_sideband"
+				"lbm_mode,render_feedback"
 			]
 			!tuple_values [
 				64
+				128
+				256
+				192
+				320
 			]
 
 		}
--- a/tools/topology/topology2/cavs-rt5682.conf
+++ b/tools/topology/topology2/cavs-rt5682.conf
@@ -164,7 +164,7 @@ Object.Dai.SSP [
 			}
 			"true" {
 				direction	"duplex"
-				quirks		"lbm_mode"
+				quirks		"lbm_mode,render_feedback"
 			}
 		}
 		name		$SPEAKER_CODEC_NAME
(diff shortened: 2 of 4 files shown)


brentlu:
PR:
linux: https://github.com/thesofproject/linux/pull/4492
alsa-utils: https://github.com/alsa-project/alsa-utils/pull/226
topology: https://github.com/thesofproject/sof/pull/7984

tools/topology/topology2/include/dais/ssp.conf
```diff
@@ -91,9 +91,17 @@ Class.Dai."SSP" {
 		constraints {
 			!valid_values [
 				"lbm_mode"
+				"bt_sideband"
+				"render_feedback"
+				"lbm_mode,bt_sideband"
+				"lbm_mode,render_feedback"
```
plbossart, line 97: this doesn't scale, we don't want to create a new definition for each possible combination. This needs to be a mask or something.

tools/topology/topology2/include/dais/ssp.conf
```diff
@@ -91,9 +91,17 @@ Class.Dai."SSP" {
...
 			]
 			!tuple_values [
 				64
+				128
+				256
+				192
+				320
```
plbossart, line 104: this doesn't seem related to NHLT?

plbossart (changes requested):

tools/topology/topology2/include/dais/ssp.conf
brentlu: I know this change looks stupid... Another way is to add new property like 'nhlt_quirk' to handle this.

plbossart: even if we add a new property, what we have today is 'global' and has no notion of direction, but the NHLT tags are not-only endpoint specific but direction specific. I am not sure how to reconcile the two, we have no placeholder for direction-specific information.

lgirdwood: I think here we could have the NHLT object and only assign direction to it when we instanciate it ? @ranj063 ?

PR #7986, opened 2023-07-26
lib-manager: Cleanups and fix the DMA handling to start the channel only one

ujfalusi:
Hi,

in order to be able to follow the hardware programming sequence (set the GEN bit first on DSP side followed by EN bit on the host side for the HDA DMA).

The firmware should only start the DMA once to copy the library data, the kernel side PR will take care of the correct sequencing:
https://github.com/thesofproject/linux/pull/4496


ujfalusi:
@all, please CC people whom I have missed.

ujfalusi:
The kernel can be extracted from the following PR: https://github.com/thesofproject/linux/pull/4494

src/library_manager/lib_manager.c
```diff
@@ -563,69 +567,110 @@ int lib_manager_load_library(uint32_t dma_id, uint32_t lib_id)
...
+
+	dma_ext = _ext_lib->runtime_data;
+
+	/* allocate temporary manifest buffer */
+	man_tmp_buffer = (__sparse_force void __sparse_cache *)
+			rballoc_align(0, SOF_MEM_CAPS_DMA,
+				      MAN_MAX_SIZE_V1_8, dma_ext->addr_align);
+	if (!man_tmp_buffer) {
+		ret = -ENOMEM;
+		goto e_init;
```
ranj063, line 661: dont we have to free dma_ext and kcps_adjust in this case too?

src/library_manager/lib_manager.c
```diff
@@ -604,36 +588,57 @@ int lib_manager_load_library(uint32_t dma_id, uint32_t lib_id)
...
+	}
 
 cleanup:
 	core_kcps_adjust(cpu_get_id(), -(CLK_MAX_CPU_HZ / 1000));
 	rfree((void *)dma_ext.dma_addr);
-	rfree((__sparse_force void *)man_tmp_buffer);
 
+e_buf:
+	rfree((__sparse_force void *)man_tmp_buffer);
+e_init:
```
plbossart, line 685: I can't follow (maybe because of the diff format) how the transfers were started multiple times and why, which makes the change even less self-explanatory.

src/include/sof/lib_manager.h
```diff
@@ -84,6 +84,8 @@ struct ext_library {
 	uint32_t mods_exec_load_cnt;
 	struct ipc_lib_msg *lib_notif_pool;
 	uint32_t lib_notif_count;
+
+	void *runtime_data;
```
plbossart, line 88: the commit message leaves me scratching my head

a)  "We cannot assume that the GEN bit is already enabled": who is 'we'? is this firmware or kernel?

b)  "The first LOAD_LIBRARY with lib_id = 0 and valid dma_id 0". what is the definition of 'first'? Is this for the basefw?

c) "Second LOAD_LIBRARY with valid lib_id and dma_id". What is the definition of 'second'? Is this for the first library that's not the basefw? Is this that the kernel is *REQUIRED* to send two LOAD_LIBRARY IPCs with the lib_id being zero first and something non-zero next?

c) Single stage flow (currently supported)
What is the point of having two options, what is the recommended solution and why? This mixes current situation and suggested improvement, it's very hard to follow your line of thought

src/ipc/ipc4/handler.c
```diff
@@ -456,15 +456,11 @@ static int ipc4_set_pipeline_state(struct ipc4_message_request *ipc4)
 static int ipc4_load_library(struct ipc4_message_request *ipc4)
 {
 	struct ipc4_module_load_library library;
-	int ret = IPC4_UNAVAILABLE;
 
 	library.header.dat = ipc4->primary.dat;
-	library.data.dat = ipc4->extension.dat;
 
-	ret = lib_manager_load_library(library.header.r.dma_id, library.header.r.lib_id);
-	tr_info(&ipc_tr, "ipc: ipc4_load_library %d", ret);
```
plbossart, line 465: why not keep a trace of the load library success? What are we trying to optimize?

src/library_manager/lib_manager.c
```diff
@@ -563,69 +567,110 @@ int lib_manager_load_library(uint32_t dma_id, uint32_t lib_id)
...
-	uint32_t addr_align;
-	int ret, ret2;
-	void __sparse_cache *man_tmp_buffer = NULL;
+	int ret;
 
-	if (!lib_id || lib_id >= LIB_MANAGER_MAX_LIBS) {
-		tr_err(&lib_manager_tr,
-		       "lib_manager_load_library(): invalid lib_id: %u", lib_id);
-		return -EINVAL;
-	}
```
plbossart: this seems to revert the first patch "lib-manager: Check validity of lib_id" ??

src/library_manager/lib_manager.c
```diff
@@ -563,69 +567,110 @@ int lib_manager_load_library(uint32_t dma_id, uint32_t lib_id)
...
 	if (ret < 0)
-		goto cleanup;
+		goto err_dma;
+
+	_ext_lib->runtime_data = dma_ext;
+
+	return 0;
+
+err_dma:
+	core_kcps_adjust(cpu_get_id(), -(CLK_MAX_CPU_HZ / 1000));
```
plbossart, line 617: that would put the clock into negative or zero territory?

plbossart (commented):
This PR isn't for the faint-hearted. It'd need quite a bit of love to make it self-explanatory and within reach of 'the rest of us'.

src/library_manager/lib_manager.c
pjdobrowolski: adjust kcps is done because of recreating flow from windows driver in FW. Please use IPC4_RAT_DSP_KCPS instead.

src/library_manager/lib_manager.c
```diff
@@ -656,6 +657,10 @@ int lib_manager_load_library(uint32_t dma_id, uint32_t lib_id)
 	if (ret < 0)
 		goto cleanup;
 
+	ret = core_kcps_adjust(cpu_get_id(), CLK_MAX_CPU_HZ / 1000);
+	if (ret < 0)
+		goto cleanup;
+
```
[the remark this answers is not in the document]
pjdobrowolski: Setting high kcps is done because in legacy FW that operation switches oscillator to HPROSC and prevents copying errors.

pjdobrowolski (changes requested):

src/library_manager/lib_manager.c
```diff
@@ -332,26 +333,26 @@ static void lib_manager_update_sof_ctx(struct sof_man_fw_desc *desc, uint32_t li
...
 	int ret;
 
-	uint32_t align = 4;
-	uint32_t entry_index = LIB_MANAGER_GET_MODULE_INDEX(module_id);
-
-	new_drv_info = rballoc_align(0, SOF_MEM_CAPS_RAM,
-				     sizeof(struct comp_driver_info), align);
+	new_drv_info = rmalloc(SOF_MEM_ZONE_RUNTIME_SHARED, 0,
+			       SOF_MEM_CAPS_RAM | SOF_MEM_FLAG_COHERENT,
+			       sizeof(struct comp_driver_info));
```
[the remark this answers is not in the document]
ujfalusi: @lyakh, can you help me out with this?

src/library_manager/lib_manager.c
ujfalusi: Before this patch the DMA got re-configured for each transfer block which must be avoided to be able to follow the programming sequence of the DMA.
With this change we will configure and start the DMA once, receive all the chunks and then stop the DMA.

src/include/sof/lib_manager.h
ujfalusi: I will drop this patch

src/library_manager/lib_manager.c
ujfalusi: I would not require another message (two actually) to be sent when the same thing can be achieved without that said message.

src/library_manager/lib_manager.c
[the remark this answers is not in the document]
lyakh: this change does several things:
1. no need to use `rballoc()` - this isn't a data buffer, this is just a normal structure object
2. allocating as "shared" means - uncached. I think that's appropriate because that's what seems to be used for `comp_register()` in other places.

app/boards/intel_adsp_cavs25.conf
```diff
@@ -7,6 +7,10 @@ CONFIG_AMS=y
 CONFIG_INTEL_ALH=y
 CONFIG_LP_MEMORY_BANKS=1
 CONFIG_HP_MEMORY_BANKS=30
+CONFIG_MM_DRV=y
+CONFIG_MM_DRV_INTEL_ADSP_TLB_REMAP_UNUSED_RAM=y
+CONFIG_INTEL_MODULES=y
+CONFIG_LIBRARY_MANAGER=y
```
lyakh, line 13: I guess we can load libraries on TGL. But can we use them there? AFAICS to use a library we need to map it and I'm not sure this is supported on TGL? We do seem to have a non-MTL driver drivers/mm/mm_drv_intel_adsp_tlb.c for that but I'm not actually sure if it's for TGL and whether it was ever tested with library loading

ujfalusi:
Changes since v2:
- add stop_dma label to the "don't restart DMA while loading a library" as it was missed as error case in the original patch
- add new patch for timeout handling while waiting for the host DMA to transfer the library
- Add cleanups for the error, info and debug prints in the lib_manager.

app/boards/intel_adsp_cavs25.conf
ujfalusi: Hrm, yes, I don't know, these are needed to actually make the loading to work on TGL family. I guess we need to start from somewhere to get the loading to work if we want to?

src/library_manager/lib_manager.c
```diff
@@ -445,21 +445,28 @@ static int lib_manager_dma_deinit(struct lib_manager_dma_ext *dma_ext, uint32_t
...
+	int loop_count = 3000;
 	int ret;
 
 	ret = dma_get_status(dma_ext->chan->dma->z_dev, dma_ext->chan->index, &stat);
 	if (ret < 0)
 		return ret;
 
 	/* Wait till whole data acquired */
-	while (stat.pending_length < size) {
+	while (stat.pending_length < size && loop_count--) {
```
ujfalusi, line 465: This is not correct, should be:
```c
	/* Wait till whole data acquired */
	while (stat.pending_length < size && loop_count) {
		k_usleep(100);

		ret = dma_get_status(dma_ext->chan->dma->z_dev, dma_ext->chan->index, &stat);
		if (ret < 0)
			return ret;

		loop_count--;
	}
```

lgirdwood (commented):
@ujfalusi can we split this PR up into another PR for all the "can be merged today" fixes, clean ups and improvements that are not pending alignment with reviewers. This should allow further testing and focus discussion on the alignment areas.

src/library_manager/lib_manager.c
```diff
@@ -469,62 +469,53 @@ static int lib_manager_dma_deinit(struct lib_manager_dma_ext *dma_ext, uint32_t
...
 
 	ret = dma_start(dma_ext->chan->dma->z_dev, dma_ext->chan->index);
 	if (ret < 0)
-		goto return_on_fail;
+		return ret;
 
 	/* Wait till whole data acquired */
 	while (avail_bytes < size) {
 		/* get data sizes from DMA */
 		ret = dma_get_status(dma_ext->chan->dma->z_dev, dma_ext->chan->index, &stat);
```
btian1, line 467: shall we remove those space line?

src/library_manager/lib_manager.c
```diff
@@ -37,22 +37,10 @@ DECLARE_SOF_UUID("lib_manager", lib_manager_uuid, 0x54cf5598, 0x8b29, 0x11ec,
...
-	uintptr_t end_addr;	/**< buffer end pointer */
-	uint32_t size;		/**< buffer size */
-	uint32_t avail;		/**< buffer avail data */
-};
-
 struct lib_manager_dma_ext {
 	struct dma *dma;
 	struct dma_chan_data *chan;
-	struct lib_manager_dma_buf dmabp;
+	uintptr_t dma_addr;		/**< buffer start pointer */
```
btian1, line 43: if use addr, then less code change.

src/library_manager/lib_manager.c
```diff
@@ -445,19 +445,28 @@ static int lib_manager_dma_deinit(struct lib_manager_dma_ext *dma_ext, uint32_t
...
+	int loop_count = 1000;
 	int ret;
 
 	ret = dma_get_status(dma_ext->chan->dma->z_dev, dma_ext->chan->index, &stat);
 	if (ret < 0)
 		return ret;
 
 	/* Wait till whole data acquired */
-	while (stat.pending_length < size) {
+	while (stat.pending_length < size && loop_count) {
```
btian1, line 465: better to adjust the sleep time instead of add new count, if add new count, how to determine the times?
100 us * 1000 = 100ms, so you are expect load library within 100ms?

src/library_manager/lib_manager.c
```diff
@@ -664,8 +669,8 @@ int lib_manager_load_library(uint32_t dma_id, uint32_t lib_id)
 	ret = lib_manager_store_library(&dma_ext, man_tmp_buffer, lib_id, addr_align);
 
 cleanup:
+	core_kcps_adjust(cpu_get_id(), -(CLK_MAX_CPU_HZ / 1000));
```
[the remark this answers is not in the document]
ujfalusi: Oh, you are right, I need another label to be used to rewind the kcps,

src/library_manager/lib_manager.c
ujfalusi: the file would use some cleanup for coding style and consistency, but it is not a subject of this patch.

src/library_manager/lib_manager.c
ujfalusi: I thought that it is a bit of over-engineering to have time check, so I used a simple counter.
Let me see if I can find a good template for the time counter.

jxstelter (approved):

src/library_manager/lib_manager.c
```diff
@@ -468,16 +467,17 @@ static int lib_manager_load_data_from_host(struct lib_manager_dma_ext *dma_ext,
 	if (ret < 0)
 		return ret;
 
+	ret = dma_get_status(dma_ext->chan->dma->z_dev, dma_ext->chan->index, &stat);
```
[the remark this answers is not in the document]
btian1: then you can use:
do
{

}while()

still can have same effect.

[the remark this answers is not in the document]
ujfalusi:
```
do {
	gather condition
	if (condition met)
		break;

	k_usleep(100);
} while (test for timeout);
```
would work, yes.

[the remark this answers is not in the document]
ujfalusi: and to iterate on that:
```
do {
	gather condition
	if (ret < 0 || condition met)
		return ret;

	k_usleep(100);
} while (test for timeout);
```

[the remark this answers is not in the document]
ujfalusi: But this is not good either. I don't want needles sleep after the first get_dma_status() and I want the get_dma_status() as the last call at the very end of the loop (at timeout).
this would:
get_dma_status()
sleep
report timout

while in _theory_ the DMA could have finished during the last sleep and we are not checking that, unless we add a get_dma_status() before the report  timeout, which is the same thing as I have right now, but on the other end.

I will not touch this patch.

ujfalusi:
Changes since v5:
- Change the timeout loop to a more compact implementation

app/boards/intel_adsp_cavs25.conf
kv2019i: I'm ok to go with incremental change, but it's surprising to see this works as the TLB_REMAP driver in Zephyr is not available to cAVS2.5.

kv2019i (approved):
I'd recommend update at least one git commit to explain the change better, but other than that, series looks good.

app/boards/intel_adsp_cavs25.conf
ujfalusi: Right, I will drop the CONFIG_MM_DRV_INTEL_ADSP_TLB_REMAP_UNUSED_RAM and add two more patches to handle the Kconfig variations better.

ujfalusi:
Changes since v6:
- Drop the CONFIG_MM_DRV_INTEL_ADSP_TLB_REMAP_UNUSED_RAM from the cavs25 config
- fix the handling of the case when INTEL_MODULES is not enabled
- Add MM_DRV as dependency for the LIBRARY_LOADER as it is needed for compilation
- Commit message updated as per @kv2019i's suggestion

kv2019i (approved):
Thanks @ujfalusi , looks good now!

kv2019i:
@pjdobrowolski can you check?

lyakh:
> I'm ok to go with incremental change, but it's surprising to see this works as the TLB_REMAP driver in Zephyr is not available to cAVS2.5.

@kv2019i I think the reason is, that the remapping is only done in step 2 - when the loaded module is instantiated to be used in a pipeline. As long as that isn't done, remapping isn't needed.

ujfalusi:
> > I'm ok to go with incremental change, but it's surprising to see this works as the TLB_REMAP driver in Zephyr is not available to cAVS2.5.
> 
> @kv2019i I think the reason is, that the remapping is only done in step 2 - when the loaded module is instantiated to be used in a pipeline. As long as that isn't done, remapping isn't needed.

and I have dropped the TLB_REMAP, I think it is a good thing that we can test at least the loading to catch regressions in other places?

src/library_manager/lib_manager.c
```diff
@@ -444,23 +444,28 @@ static int lib_manager_dma_deinit(struct lib_manager_dma_ext *dma_ext, uint32_t
...
-
-		ret = dma_get_status(dma_ext->chan->dma->z_dev, dma_ext->chan->index, &stat);
-		if (ret < 0)
+	for (;;) {
+		ret = dma_get_status(dma_ext->chan->dma->z_dev,
+				     dma_ext->chan->index, &stat);
+		if (ret < 0 || stat.pending_length >= size)
 			return ret;
+
+		if (sof_cycle_get_64() > timeout)
```
lyakh, line 473: on the first iteration this *really* shouldn't trigger - we shouldn't be spending more than 200ms between line 452 and here, so I think the previous version with
```
    k_usleep(100);
} while (sof_cycle_get_64() <= timeout);
```
was good too :-) But not too important

lyakh (approved):

lgirdwood (approved):

lgirdwood:
@pjdobrowolski good for you now ?

pjdobrowolski (approved):

app/boards/intel_adsp_cavs25_tgph.conf
```diff
@@ -6,6 +6,9 @@ CONFIG_INTEL_SSP=y
 CONFIG_INTEL_ALH=y
 CONFIG_LP_MEMORY_BANKS=1
 CONFIG_HP_MEMORY_BANKS=30
+CONFIG_MM_DRV=y
+CONFIG_INTEL_MODULES=y
```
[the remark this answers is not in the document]
ujfalusi: The MM_DRV is compile time dependency (as per https://github.com/thesofproject/sof/pull/7986/commits/14e4b6c1bd869965afa15c9293b4d8a31044f547), I can drop the INTEL_MODULES as the code will compile now (https://github.com/thesofproject/sof/pull/7986/commits/78eeaa0c50b50dc988f37e7ba73552331659475a)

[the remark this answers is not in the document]
jxstelter: > memory manager and tlb was prepared for mtl. Are you sure that it is valid configuration? @jxstelter we've talk about it ealier, any thoughts?

For library loading the generic MM driver will work. However instantiating library will fail since memory remapping is required. Currently only MTL specific code (mm_drv_intel_adsp_mtl_tlb.c) has remapping implemented.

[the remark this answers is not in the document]
ujfalusi: I guess it is fine if instantiating fails, but the loading works. For now. If we keep this part working than 'only' the instantiation needs to be worked on for TGL, if we decide to do so.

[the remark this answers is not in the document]
lgirdwood: ok, let revisit the MM if needed later. We can merge for now.

Issue #7979, opened 2023-07-25
[BUG][MTL] Speaker playback IO error


aiChaoSONG:
**Describe the bug**
Playback a wave to speaker with aplay, see io error in terminal immediately. no obviously error in kernel log, and in mtrace, xrun happened.

```
[    1.999690] <inf> host_comp: host_get_copy_bytes_normal: comp:0 0x40000 no bytes to copy, available samples: 384, free_samples: 0
[    2.000706] <inf> host_comp: host_get_copy_bytes_normal: comp:0 0x40000 no bytes to copy, available samples: 384, free_samples: 0

```

**To Reproduce**
simple do aplay to speaker pipeline: 
aplay -Dhw:0,2 -f dat -c 2 -i -vvv /dev/zero

**Reproduction Rate**
100%

**Reproduced with**
Latest daily sof-dev kernel
Latest SOF main
topology:  sof-mtl-rt711-l0-rt1316-l23-rt714-l1.tplg
Platform: {MTL SDW} with external codecs on GEN4 AIOC (sh-mtlp-rvp-sdw-09, not in CI farm)
-	Headset: ALC711 (headset) on link 0, 
-	Speaker: ALC1316 x2 on link 2 & 3 
-	DMIC: ALC714 on link1.  

kernel options snd_soc_sof_sdw quirk=2

The first bad commit is: e424b874a477617d080f368ae67b46d19b7e6d4d
![image](https://github.com/thesofproject/sof/assets/8703742/4224a916-a5e2-46e5-b910-71d538ccbf14)

mengdonglin:
@ranj063 Can you help check this regression?

ranj063:
> @ranj063 Can you help check this regression?

@mengdonglin I think I know the issue. We're missing the src buffer update in the case of multie endpoint playback. I've given @btian1 the change to try out and check if it fixes the problem

lgirdwood:
> it fixed this issue, please check with below PR, @aiChaoSONG @bardliao @ranj063 #7987

@aiChaoSONG Now merged #7987.

aiChaoSONG:
Verified fix #7987, close.

PR #7987, opened 2023-07-27
copier: fix no consume for multi endpoint copy

btian1:
In last time copier code split, it split multi-endpoint copy and module copy with commit: e424b87, source buffer consume was missed, added it back with this PR.

Fixes:#7979

### DIFF
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -561,6 +561,10 @@ static int copier_multi_endpoint_dai_copy(struct copier_data *cd, struct comp_de
 
 	ret = dai_zephyr_multi_endpoint_copy(cd->dd, dev, cd->multi_endpoint_buffer,
 					     cd->endpoint_num);
+	if (!ret) {
+		comp_update_buffer_consume(src_c, processed_data.source_bytes);
+		cd->input_total_data_processed += processed_data.source_bytes;
+	}
 err:
 	buffer_release(src_c);
 


ranj063 (approved):

src/audio/copier/copier.c
```diff
@@ -561,6 +561,8 @@ static int copier_multi_endpoint_dai_copy(struct copier_data *cd, struct comp_de
 
 	ret = dai_zephyr_multi_endpoint_copy(cd->dd, dev, cd->multi_endpoint_buffer,
 					     cd->endpoint_num);
+
+	comp_update_buffer_consume(src_c, processed_data.source_bytes);
```
RanderWang, line 565:
````c
if (!ret)
     comp_update_buffer_consume(src_c, processed_data.source_bytes);
````

File: src/audio/copier/copier.c [hunk 2]
```diff
@@ -561,6 +561,8 @@ static int copier_multi_endpoint_dai_copy(struct copier_data *cd, struct comp_de
 
 	ret = dai_zephyr_multi_endpoint_copy(cd->dd, dev, cd->multi_endpoint_buffer,
 					     cd->endpoint_num);
+	if (!ret)
+		comp_update_buffer_consume(src_c, processed_data.source_bytes);
```
lyakh, line 565: does `.input_total_data_processed` need to be updated too?

btian1: I need to say:  "I think so". according to origin code logic, we need update this as well without extra conditions.
since it is multi dai case, endpoint_num != 0 and bsource_buffer == true.

These part of code is still in critical path, we'd better to find a way to remove some checks.

aiChaoSONG:
Small improvement suggestion:
It is important to mention what bug a commit fixes, or the previously commit that bring the bug. This makes the git history much more trackable.

In the commit message, `with last time change` is descriptive and not specific, it didn't mention the commit id, people will spend time to find what is the last time change, let's save their time.

Please also add `Fixes: #7979` in commit message.

lgirdwood (approved):

PR #7988, opened 2023-07-27
zephyr_ll: Fix panic upon cancelling not scheduled task

serhiy-katsyuba-intel:
Calling zephyr_ll_task_cancel() for a task which was actually never scheduled results later in panic in zephyr_ll_task_done().

The fix makes zephyr_ll_task_cancel() do nothing for tasks which were only initialized but never scheduled.


lyakh (approved):

juimonen (approved):
looks ok to me (and I trust @lyakh in this)

PR #7989, opened 2023-07-28
Enable probes on MTL

serhiy-katsyuba-intel:
First commit is borrowed from Adrian Bonislawski https://github.com/abonislawski/sof/commit/e5b45c632104e3726748b12f8cdda4ebd4e17444 -- this is to port probe component to use Zephyr DMA stuff.

Rest are fixes required for probes to work properly.

Additionally, this zephyr_ll scheduler fix is required for injection probes not to cause FW panic: https://github.com/thesofproject/sof/pull/7988


src/probe/probe.c
```diff
@@ -200,14 +261,20 @@ static int probe_dma_init(struct probe_dma_ext *dma, uint32_t direction)
...
 	if (err < 0) {
 		tr_err(&pr_tr, "probe_dma_deinit(): dma_stop() failed");
 		return err;
 	}
-
+#if CONFIG_ZEPHYR_NATIVE_DRIVERS
+	dma_release_channel(dma->dc.dmac->z_dev, dma->dc.chan->index);
+#else
 	dma_channel_put_legacy(dma->dc.chan);
+#endif
```
btian1, line 277: just a suggestion, use one #if from line 264 - 277, this can  remove one more "#if #else #endif", overhead is duplicated err message.

btian1 (commented):
QB still have failed case, please check.

serhiy-katsyuba-intel:
SOFCI TEST

src/probe/probe.c
```diff
@@ -121,6 +121,7 @@ static int probe_dma_buffer_init(struct probe_dma_buf *buffer, uint32_t size,
 	return 0;
 }
 
+#if !CONFIG_ZEPHYR_NATIVE_DRIVERS
```
lgirdwood, line 124: Is the [DEBUG] title still needed here ?

src/audio/module_adapter/module_adapter.c
```diff
@@ -858,7 +858,7 @@ static int module_adapter_audio_stream_copy_1to1(struct comp_dev *dev)
 		buffer_stream_writeback(sink_c, mod->output_buffers[0].size);
 
 	if (mod->output_buffers[0].size)
-		audio_stream_produce(&sink_c->stream, mod->output_buffers[0].size);
+		comp_update_buffer_produce(sink_c, mod->output_buffers[0].size);
```
lgirdwood, line 861: @serhiy-katsyuba-intel @btian1 @ranj063 IIUC, this code was changed (in reverse to this patch) to improve MCPS. I think the comp_ API was calling some notifiers (that I think were only used for legacy WOV and L1 exit and no longer used). I would say here we need a light touch when invoking a probe or injector. We could remove the notifier and directly check for probe/injector.

src/probe/probe.c
serhiy-katsyuba-intel: Removed [DEBUG] from title.

src/audio/module_adapter/module_adapter.c
serhiy-katsyuba-intel: I did some performance measurements in comp_update_buffer_produce(). When CONFIG_SOF_LOG_LEVEL_DBG is enabled, debug logs at the bottom of that func consumes o lot of CPU. Entire function consumes 11233 ccounts of which two buf_dbg() consume 10419 ccounts. Removing buf_dbg() reducess CPU consumtion more then 10 times.

Without buf_dbg() and when no listeners registered, comp_update_buffer_produce() seems does not do anything which consumes lot's of CPU. If needed, it could be potentially optimized more: there is a loop there over all cores to check if IDC is needed to send notification to another core -- that loop could be eliminated with prior "if" as usually (always?) notification sender and receiver is on same core. But such optimization seems will not give us much.

I modified "Enable buffer produce notifications required by probe" commit by commenting out those expensive buf_dbg() calles. Hope now it's OK to return comp_update_buffer_produce() back to module_adapter.

High CPU consumption only happens if CONFIG_SOF_LOG_LEVEL_DBG is enabled. Do you recall if DBG level was enabled when performance measurements were made previously?

src/audio/buffer.c
```diff
@@ -251,6 +258,7 @@ void comp_update_buffer_produce(struct comp_buffer __sparse_cache *buffer, uint3
 		 (char *)audio_stream_get_addr(&buffer->stream)) << 16 |
 		((char *)audio_stream_get_wptr(&buffer->stream) -
 		 (char *)audio_stream_get_addr(&buffer->stream)));
+*/
```
lgirdwood, line 261: Should this be commented out ?

lgirdwood: ok, so it would seem the best way is to KCONFIG_SOF_LOG_DBG_BUFFER with a boolean that just #ifdefs these logs. This would allow a Kconfig to enable if anyone needed to debug a buffer.

lgirdwood (approved):

jxstelter (approved):

abonislawski (approved):

tmleman (approved):

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -88,3 +88,6 @@ CONFIG_CLOCK_CONTROL=y
 CONFIG_DEBUG_COREDUMP=y
```
kv2019i, line 88: The commit prefix should be fixed. We haven't been very consistent in "sof/app/overlays" but "app: overlay: Enable probe MTL" would do.

src/audio/module_adapter/module_adapter.c
kv2019i, line 861: "audio: module_adapter: Enable buffer produce notifications"

kv2019i (commented):
Thanks @serhiy-katsyuba-intel , looks good. Can you update the commit prefixes to align with existing style?

src/audio/buffer.c
```diff
@@ -228,11 +228,13 @@ void comp_update_buffer_produce(struct comp_buffer __sparse_cache *buffer, uint3
 
 	/* return if no bytes */
 	if (!bytes) {
+#if CONFIG_SOF_LOG_DBG_BUFFER
```
btian1, line 231: it build nothing as default with buf_dbg, if change build log level, then it will print lot of log messages,  I just mean default there is no cycle cost, and enable it indeed cost too much, not sure need a config to just cover is expensive or not.

jsarha (approved):
I have just a minor nitpick, that can be ignored, but there were some other valid comments, however with those fixed, approved by me.

PR #7993, opened 2023-07-31
module_adapter: inline some time critical functions

btian1:
Those functions are only called once and in critical timing path, Inline would save the function parameter push/pull stack operation.

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -562,7 +562,7 @@ int module_adapter_params(struct comp_dev *dev, struct sof_ipc_stream_params *pa
  * @buff_size: size of the module input buffer
  * @bytes: number of bytes available in the source buffer
  */
-static void
+static inline void
 ca_copy_from_source_to_module(const struct audio_stream __sparse_cache *source,
 			      void __sparse_cache *buff, uint32_t buff_size, size_t bytes)
 {
@@ -590,7 +590,7 @@ ca_copy_from_source_to_module(const struct audio_stream __sparse_cache *source,
  * @buff: pointer to the module output buffer
  * @bytes: number of bytes available in the module output buffer
  */
-static void
+static inline void
 ca_copy_from_module_to_sink(const struct audio_stream __sparse_cache *sink,
 			    void __sparse_cache *buff, size_t bytes)
 {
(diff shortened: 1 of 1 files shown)


dbaluta:
Do, you have any data for this? This kind of optimizations need to be documented.

btian1:
> is kind of optimizations need to be documented.

I do have, but it is in audio stream path, this is raw data path, it normally can save 5% or upper cycles with same test.

lgirdwood (commented):
@btian1 can you update the commit message to show perf improvement for some tested pipelines and any memory increase. Thanks

btian1:
> @btian1 can you update the commit message to show perf improvement for some tested pipelines and any memory increase. Thanks

these are IADK modules, used by AMD, we have no way to perform a test, so can't collect data, I collected data from 
audio stream copy， after removed function shell, around 3%-5% improve, check below link:
https://github.com/thesofproject/sof/pull/7909/commits/320b54ba2a1b2503156dd1f9e863a595cb74fc1c
@marcinszkudlinski , please also comment on this patch.

btian1:
> > > @btian1 can you update the commit message to show perf improvement for some tested pipelines and any memory increase. Thanks
> > 
> > 
> > these are IADK modules, used by AMD, we have no way to perform a test, so can't collect data, I collected data from audio stream copy， after removed function shell, around 3%-5% improve, check below link: [320b54b](https://github.com/thesofproject/sof/commit/320b54ba2a1b2503156dd1f9e863a595cb74fc1c) @marcinszkudlinski , please also comment on this patch.
> 
> ok, but what about code size impact ?

I checked multiple times, seems there is no difference for inline and not inline, anything changed in compiler build options?

They are same with below for tgl build：
 module: input size 668231 (0xa3247) bytes 35 sections
 module: text 319959 (0x4e1d7) bytes
    data 348272 (0x55070) bytes
    bss  9312 (0x2460) bytes

Now, I don't have a justification to keep this PR.

lgirdwood:
> I checked multiple times, seems there is no difference for inline and not inline, anything changed in compiler build options?
> 
> They are same with below for tgl build： module: input size 668231 (0xa3247) bytes 35 sections module: text 319959 (0x4e1d7) bytes data 348272 (0x55070) bytes bss 9312 (0x2460) bytes
> 
> Now, I don't have a justification to keep this PR.

Yeah, it looks like the compiler is doing the correct thing, not sure why you could see an perf difference, maybe rerun the test ?

btian1:
> 

the result is from #7909 ， #7909 need wait #7958 merge first, then I will have a check again.

PR #7995, opened 2023-07-31
ipc4: multi pipeline set state on different cores

juimonen:
This will allow to correctly set multiple pipelines state even if they are allocated on different cores.

ipc4_set_pipeline_state will check if several cores are involved
- set ppl state if only current core requested
- process IPC on another core if only single secondary core requested
- send IDC messages if several secondary cores involved


lgirdwood:
This is an update of #7962 as @abonislawski is OOO.

mwasko (approved):

PR #8002, opened 2023-08-01
DP: changes in sink/source

marcinszkudlinski:
1) add min_free / min_available 
Its obvious that to process, a module must have enough data on all inputs (sources) and enough space on all outputs (sinks). 
For each input/output the amount of required data/space may be different. 

This PR adds min_free/min_available) to sink/source

This information will be used by DP Scheduler to determine if the module may be scheduled.

2) add data invalidation to audio_stream sink/src implementation

Audio_stream is not making cache operations on data
buffer itself, it requires that the module to
call data invalidation/writeback

Sink/source interface works differently, it removes any
responsibility from the module providing a ready to
use pointers in cached memory

this PR adds cache operations to audio_stream sink/src implementation

### DIFF
--- a/src/audio/audio_stream.c
+++ b/src/audio/audio_stream.c
@@ -37,9 +37,14 @@ static int audio_stream_commit_buffer(struct sof_sink __sparse_cache *sink, size
 	struct audio_stream __sparse_cache *audio_stream =
 			attr_container_of(sink, struct audio_stream __sparse_cache,
 					  sink_api, __sparse_cache);
+	struct comp_buffer __sparse_cache *buffer_c =
+			attr_container_of(audio_stream, struct comp_buffer __sparse_cache,
+					  stream, __sparse_cache);
 
-	if (commit_size)
+	if (commit_size) {
+		buffer_stream_writeback(buffer_c, commit_size);
 		audio_stream_produce(audio_stream, commit_size);
+	}
 
 	return 0;
 }
@@ -60,8 +65,14 @@ static int audio_stream_get_data(struct sof_source __sparse_cache *source, size_
 			attr_container_of(source, struct audio_stream __sparse_cache,
 					  source_api, __sparse_cache);
 
+	struct comp_buffer __sparse_cache *buffer_c =
+			attr_container_of(audio_stream, struct comp_buffer __sparse_cache,
+					  stream, __sparse_cache);
+
 	if (req_size > audio_stream_get_data_available(source))
 		return -ENODATA;
+
+	buffer_stream_invalidate(buffer_c, req_size);
 
 	/* get circular buffer parameters */
 	*data_ptr = audio_stream->r_ptr;
--- a/src/audio/sink_api_helper.c
+++ b/src/audio/sink_api_helper.c
@@ -175,3 +175,13 @@ int sink_set_alignment_constants(struct sof_sink __sparse_cache *sink,
 		return sink->ops->set_alignment_constants(sink, byte_align, frame_align_req);
 	return 0;
 }
+
+void sink_set_obs(struct sof_sink __sparse_cache *sink, size_t obs)
+{
+	sink->obs = obs;
+}
+
+size_t sink_get_obs(struct sof_sink __sparse_cache *sink)
+{
+	return sink->obs;
+}
--- a/src/audio/source_api_helper.c
+++ b/src/audio/source_api_helper.c
@@ -165,3 +165,13 @@ int source_set_alignment_constants(struct sof_source __sparse_cache *source,
 		return source->ops->set_alignment_constants(source, byte_align, frame_align_req);
 	return 0;
 }
+
+void source_set_ibs(struct sof_source __sparse_cache *source, size_t ibs)
+{
+	source->ibs = ibs;
+}
+
+size_t source_get_ibs(struct sof_source __sparse_cache *source)
+{
+	return source->ibs;
+}
--- a/src/include/sof/audio/sink_api.h
+++ b/src/include/sof/audio/sink_api.h
@@ -129,6 +129,8 @@ int sink_set_rate(struct sof_sink __sparse_cache *sink, unsigned int rate);
 int sink_set_channels(struct sof_sink __sparse_cache *sink, unsigned int channels);
 int sink_set_overrun(struct sof_sink __sparse_cache *sink, bool overrun_permitted);
 int sink_set_buffer_fmt(struct sof_sink __sparse_cache *sink, uint32_t buffer_fmt);
+void sink_set_obs(struct sof_sink __sparse_cache *sink, size_t obs);
+size_t sink_get_obs(struct sof_sink __sparse_cache *sink);
 
 /**
  * initial set of audio parameters, provided in sof_ipc_stream_params
--- a/src/include/sof/audio/source_api.h
+++ b/src/include/sof/audio/source_api.h
@@ -140,6 +140,8 @@ int source_set_rate(struct sof_source __sparse_cache *source, unsigned int rate)
 int source_set_channels(struct sof_source __sparse_cache *source, unsigned int channels);
 int source_set_underrun(struct sof_source __sparse_cache *source, bool underrun_permitted);
 int source_set_buffer_fmt(struct sof_source __sparse_cache *source, uint32_t buffer_fmt);
+void source_set_ibs(struct sof_source __sparse_cache *source, size_t ibs);
+size_t source_get_ibs(struct sof_source __sparse_cache *source);
 
 /**
  * initial set of audio parameters, provided in sof_ipc_stream_params
--- a/src/ipc/ipc4/helper.c
+++ b/src/ipc/ipc4/helper.c
@@ -327,20 +327,14 @@ int ipc_pipeline_free(struct ipc *ipc, uint32_t comp_id)
 }
 
 static struct comp_buffer *ipc4_create_buffer(struct comp_dev *src, struct comp_dev *sink,
-					      uint32_t src_queue, uint32_t dst_queue)
+					      uint32_t src_obs, uint32_t src_queue,
+					      uint32_t dst_queue)
 {
-	struct ipc4_base_module_cfg src_cfg;
 	struct sof_ipc_buffer ipc_buf;
-	int buf_size, ret;
-
-	ret = comp_get_attribute(src, COMP_ATTR_BASE_CONFIG, &src_cfg);
-	if (ret < 0) {
-		tr_err(&ipc_tr, "failed to get base config for src %#x", dev_comp_id(src));
-		return NULL;
-	}
+	int buf_size;
 
 	/* double it since obs is single buffer size */
-	buf_size = src_cfg.obs * 2;
+	buf_size = src_obs * 2;
 
 	memset(&ipc_buf, 0, sizeof(ipc_buf));
 	ipc_buf.size = buf_size;
@@ -356,6 +350,8 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 	struct comp_buffer *buffer;
 	struct comp_dev *source;
 	struct comp_dev *sink;
+	struct ipc4_base_module_cfg source_src_cfg;
+	struct ipc4_base_module_cfg sink_src_cfg;
 	uint32_t flags;
 	int src_id, sink_id;
 	int ret;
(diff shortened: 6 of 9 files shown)


src/include/sof/audio/sink_api.h
```diff
@@ -160,4 +160,8 @@ int sink_set_alignment_constants(struct sof_sink __sparse_cache *sink,
 				 const uint32_t byte_align,
 				 const uint32_t frame_align_req);
 
+void sink_set_min_free(struct sof_sink __sparse_cache *sink, size_t min_free);
+size_t sink_get_min_free(struct sof_sink __sparse_cache *sink);
+bool sink_get_is_enough_free(struct sof_sink __sparse_cache *sink);
+
```
btian1, line 166: these one line functions can be directly put to header file and with inline in declaration, see component_ext.h:
https://github.com/thesofproject/sof/blob/main/src/include/sof/audio/component_ext.h#L69

src/include/sof/audio/source_api.h
```diff
@@ -171,4 +171,8 @@ int source_set_alignment_constants(struct sof_source __sparse_cache *source,
 				   const uint32_t byte_align,
 				   const uint32_t frame_align_req);
 
+void source_set_min_available(struct sof_source __sparse_cache *source, size_t min_available);
+size_t source_get_min_available(struct sof_source __sparse_cache *source);
+bool source_get_is_enough_available(struct sof_source __sparse_cache *source);
```
btian1, line 176: same with second patch.

src/ipc/ipc4/helper.c
```diff
@@ -327,27 +327,24 @@ int ipc_pipeline_free(struct ipc *ipc, uint32_t comp_id)
...
 	memset(&ipc_buf, 0, sizeof(ipc_buf));
 	ipc_buf.size = buf_size;
 	ipc_buf.comp.id = IPC4_COMP_ID(src_queue, dst_queue);
 	ipc_buf.comp.pipeline_id = src->ipc_config.pipeline_id;
 	ipc_buf.comp.core = src->ipc_config.core;
-	return buffer_new(&ipc_buf);
+
+	struct comp_buffer *buffer = buffer_new(&ipc_buf);
+
+	return buffer;
```
btian1, line 347: any reason for this change? previous looks more simple

src/include/sof/audio/sink_api.h
marcinszkudlinski: No, I can't do it like this
Sink/source is a structure that **must not** be accessible directly (like private in C++). Therefore in sink_api.h the structure is not defined, all the user have is a forward declaration. 
Operations on the structure must be encapsulated in separate C file.

src/include/sof/audio/source_api.h
marcinszkudlinski: as above

marcinszkudlinski:
rebased on newest main to fix build errors

src/audio/audio_stream.c
```diff
@@ -21,10 +21,15 @@ static int audio_stream_get_buffer(struct sof_sink __sparse_cache *sink, size_t
...
 			attr_container_of(sink, struct audio_stream __sparse_cache,
 					  sink_api, __sparse_cache);
+	struct comp_buffer __sparse_cache *buffer_c =
+			attr_container_of(audio_stream, struct comp_buffer __sparse_cache,
+					  stream, __sparse_cache);
 
 	if (req_size > audio_stream_get_free_size(sink))
 		return -ENODATA;
 
+	buffer_stream_invalidate(buffer_c, req_size);
```
lgirdwood, line 31: Will this impact MCPS of current LL code today ? or will this PR move the invalidate from one place to another ?

src/audio/sink_api_helper.c
```diff
@@ -175,3 +175,18 @@ int sink_set_alignment_constants(struct sof_sink __sparse_cache *sink,
...
+
+size_t sink_get_min_free(struct sof_sink __sparse_cache *sink)
+{
+	return sink->min_free;
+}
+
+bool sink_get_is_enough_free(struct sof_sink __sparse_cache *sink)
+{
+	return (sink->min_free <= sink_get_free_size(sink));
+}
```
lgirdwood, line 192: These are better as static inline in the header as it will save a call() in most cases.

src/audio/source_api_helper.c
```diff
@@ -165,3 +165,19 @@ int source_set_alignment_constants(struct sof_source __sparse_cache *source,
...
+
+bool source_get_is_enough_available(struct sof_source __sparse_cache *source)
+{
+	return (source->min_available <= source_get_data_available(source));
+}
+
+size_t source_get_min_available(struct sof_source __sparse_cache *source)
+{
+	return source->min_available;
+}
```
lgirdwood, line 182: ditto for static inline in header.

src/audio/audio_stream.c
marcinszkudlinski: Just move from module to base code

src/audio/sink_api_helper.c
marcinszkudlinski: see comments for btian:

> No, I can't do it like this
Sink/source is a structure that must not be accessible directly (like private in C++). Therefore in sink_api.h the structure is not defined here, all the user has is a forward declaration.
Operations on the structure must be encapsulated in separate C file.

src/audio/source_api_helper.c
marcinszkudlinski: ...same

src/audio/audio_stream.c
lgirdwood: ok, can we remove from any modules in same patch to avoid any increase in MCPS e.g. for double invalidates. I'm guessing we have some module still doing this ?

src/audio/sink_api_helper.c
lgirdwood: ok, so these are private for core use only. Zephyr takes an approach based on API prefix, but we could do that OR we can
1) Use existing ```include/sof``` as private API directory
2) Use new ```include/public``` as public API directory

We could start moving APIs to public quite quickly and this would force modules on the public API ?

src/audio/audio_stream.c
btian1: I double checked, seems only one place use this:
https://github.com/thesofproject/sof/blob/main/src/audio/sink_api_helper.c#L30
no repeated invalidate operation.

it will have minor mcps impact, since it is in copy path.

src/audio/sink_api_helper.c
btian1: as long as these functions are not called in timing critical path, it is fine for the price, if in copy/process function, then we may need move it further.

lgirdwood: I'm not in favour of having extra calls as the cost of partitioning public/private APIs. We should be able to provide an optimal and clear client API using doxygen, comment partionint and get/set accessors.

kv2019i: @lgirdwood @marcinszkudlinski If these are truly general (usable for any sink/source implementation), then we can indeed inline them.

For source/sink methods that may be overridden by the specific type of source/sink, we have to incur the function call cost, as this is only way to abstract the interface.

But yeah. we can mix the two. True generic helpers can be inlined, while the "ops" methods incur a call overhead.

marcinszkudlinski: @kv2019i @lgirdwood  we'll need to expose sink / source structure
there was a lot effort NOT to do this
all sink/src functions are already implemented in this way, please don't change it for those 2

lgirdwood: So everything is exposed anyway (open code), but stating manadory usage of accessors in the header would be helpful to eliminate out of scope usage.

kv2019i (approved):
Some of the commits are a bit hard to read, but code itself looks ok.

src/audio/audio_stream.c
```diff
@@ -21,10 +21,15 @@ static int audio_stream_get_buffer(struct sof_sink __sparse_cache *sink, size_t
...
+
```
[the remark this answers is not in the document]
marcinszkudlinski: @kv2019i it is already addressed, it is not unconditional
```
static inline void buffer_stream_invalidate(struct comp_buffer __sparse_cache *buffer,
					    uint32_t bytes)
{
	if (!is_coherent_shared(buffer, c))
		return;

	audio_stream_invalidate(&buffer->stream, bytes);
}

static inline void buffer_stream_writeback(struct comp_buffer __sparse_cache *buffer,
					   uint32_t bytes)
{
	if (!is_coherent_shared(buffer, c))
		return;

	audio_stream_writeback(&buffer->stream, bytes);
}
```

so those have effect in case of shared buffers only. Shared buffers are always connecting 2 cores.

src/audio/audio_stream.c
```diff
@@ -21,10 +21,15 @@ static int audio_stream_get_buffer(struct sof_sink __sparse_cache *sink, size_t
 	struct audio_stream __sparse_cache *audio_stream =
```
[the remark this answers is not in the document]
marcinszkudlinski: yes. To be more specific - in sink/source implementation. In case of audio stream implementation of sink/source - it is the place.

btian1 (commented):
MTLP_RVP_NOCODEC
there is still a lot of failure for mtlp nocodec.

lyakh (commented):
don't understand why those minimum data and space are needed

src/audio/audio_stream.c
```diff
@@ -37,9 +42,14 @@ static int audio_stream_commit_buffer(struct sof_sink __sparse_cache *sink, size
...
 			attr_container_of(sink, struct audio_stream __sparse_cache,
 					  sink_api, __sparse_cache);
+	struct comp_buffer __sparse_cache *buffer_c =
+			attr_container_of(audio_stream, struct comp_buffer __sparse_cache,
+					  stream, __sparse_cache);
 
-	if (commit_size)
+	if (commit_size) {
+		buffer_stream_writeback(buffer_c, commit_size);
```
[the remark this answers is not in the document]
marcinszkudlinski: this is not writting back the buffer_c stucture, but **the audio data from the stream**, it must be done on acquired buffer. Buffer acquire/release, however, **does not touch the audio data at all**. 
```
__must_check static inline struct comp_buffer __sparse_cache *buffer_acquire(
	struct comp_buffer *buffer)
{
	struct coherent __sparse_cache *c = coherent_acquire_thread(&buffer->c, sizeof(*buffer));

	return attr_container_of(c, struct comp_buffer __sparse_cache, c, __sparse_cache);
}

static inline void buffer_release(struct comp_buffer __sparse_cache *buffer)
{
	coherent_release_thread(&buffer->c, sizeof(*buffer));
}
```

This is performed by buffer_stream_writeback, with verification if the buffer is shared:

```
static inline void buffer_stream_writeback(struct comp_buffer __sparse_cache *buffer,  uint32_t bytes)
{
	if (!is_coherent_shared(buffer, c))
		return;

	audio_stream_writeback(&buffer->stream, bytes);
}

static inline void audio_stream_writeback(struct audio_stream __sparse_cache *buffer,
					  uint32_t bytes)
{
	uint32_t head_size = bytes;
	uint32_t tail_size = 0;

	/* check for potential wrap */
	if ((char *)buffer->w_ptr + bytes > (char *)buffer->end_addr) {
		head_size = (char *)buffer->end_addr - (char *)buffer->w_ptr;
		tail_size = bytes - head_size;
	}

	dcache_writeback_region((__sparse_force void __sparse_cache *)buffer->w_ptr, head_size);
	if (tail_size)
		dcache_writeback_region((__sparse_force void __sparse_cache *)buffer->addr,
					tail_size);
}

```

in sink/source API I want to remove all responsibility of caches, etc. from module code. Currently buffer_stream_writeback  is called in modules.... in loads of places. That makes the code extremely vulnerable to problems - it is enough that somebody forgets to call it, and everyting is working... till the module is the one "at edge" in multicore pipeline. 
In sink/source the pointers provided to modules must be ready-to-use, no additional operations

image

src/audio/sink_api_helper.c
```diff
@@ -175,3 +175,18 @@ int sink_set_alignment_constants(struct sof_sink __sparse_cache *sink,
 		return sink->ops->set_alignment_constants(sink, byte_align, frame_align_req);
 	return 0;
 }
+
+void sink_set_min_free(struct sof_sink __sparse_cache *sink, size_t min_free)
+{
+	sink->min_free = min_free;
```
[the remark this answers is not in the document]
marcinszkudlinski: you think that is better to call module and ask "do you have enough data" or "are you ready to process"? Maybe you're right... i.e. in variable rate formats like MPEG. And it is working this way in a reference FW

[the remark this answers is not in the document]
lyakh: yeah, maybe convert this to draft while we're uncertain which way to go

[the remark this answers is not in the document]
marcinszkudlinski: lets follow reference design, I've just checked it and its working fine (with glithes to be fixed ;) ). 
Anyway, keeping IBS/OBS in sink/src seems to be a good idea. Those are declared in module_bind and are specific to each connection sink-module, source-module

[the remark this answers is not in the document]
marcinszkudlinski: https://github.com/thesofproject/sof/pull/8044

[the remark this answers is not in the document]
lyakh: do I understand it correctly that #8044 now supersedes this PR? Can we close this one then?

src/audio/audio_stream.c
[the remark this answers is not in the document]
marcinszkudlinski: Ok, got your point
You're 100% right, I put this invalidation in wrong place
Protection against dirty cacheline should be in audio_stream_get_data - providing data for reading using source_api.

src/audio/sink_api_helper.c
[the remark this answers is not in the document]
marcinszkudlinski: No, cache operations are still needed, 
I also want to keep IBS/OBS in sink/source structures, not in component. I know that currently all innputs/outputs for component have same IBS/OBS (as declared in init_instance IPC), but in reference FW there's a way to have components with individual IBSes/OBSes per pin.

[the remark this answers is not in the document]
lyakh: > No, cache operations are still needed, I also want to keep IBS/OBS in sink/source structures, not in component. I know that currently all innputs/outputs for component have same IBS/OBS (as declared in init_instance IPC), but in reference FW there's a way to have components with individual IBSes/OBSes per pin.

could you explain why that would be needed and how that can be configured?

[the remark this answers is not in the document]
marcinszkudlinski: It is through module specific calls - unfortunately :(  - IPC4 init_instance has one pair of IBS/OBS 
Example - maybe a little imaginary, but reasonable, for performance and memory usage (buffers!): an echo cancelation module receiving  a 48/24/2 on main input and 16/16/1 from DMIC

lgirdwood (approved):

PR #8003, opened 2023-08-02
dp: introduce per-core DP scheduler

marcinszkudlinski:
At start point an incorrect assumption has been taken that it is enough to have one single instance of
DP scheduler located on a primary core 

This commit introduces a DP instance per each core

### DIFF
--- a/src/include/sof/schedule/dp_schedule.h
+++ b/src/include/sof/schedule/dp_schedule.h
@@ -16,20 +16,10 @@
 /**
  *
  * DP scheduler is a scheduler that creates a separate preemptible Zephyr thread for each SOF task
- *   There's only one instance of DP in the system, however, threads can be assigned and pinned
- *   to any core in the system for its execution, there's no SMP processing.
  *
  * The task execution may be delayed and task may be re-scheduled periodically
  * NOTE: delayed start and rescheduling takes place in sync with LL scheduler, meaning the
- *   DP scheduler is triggered as the last task of LL running on a primary core.
- *   That implies a limitation: LL scheduler MUST be running on primary core in order to have
- *   this feature working.
- *   It is fine, because rescheduling is a feature used for data processing when a pipeline is
- *   running.
- *
- * Other possible usage of DP scheduler is to schedule task with DP_SCHEDULER_RUN_TASK_IMMEDIATELY
- *   as start parameter. It will force the task to work without any delays and async to LL.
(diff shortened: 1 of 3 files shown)


lgirdwood (approved):

lgirdwood:
@marcinszkudlinski do we have a DP multicore test topology in CI yet ? I'm guessing this is a TODO ? @mengdonglin do you know ?

tmleman (approved):

src/include/sof/schedule/dp_schedule.h
```diff
@@ -67,13 +67,9 @@
...
 int scheduler_dp_init(void);
 
-/**
- * \brief Set the Data Processing scheduler to be accessible at secondary cores
- */
-int scheduler_dp_init_secondary_core(void);
-
 /**
  * \brief initialize a DP task and add it to scheduling
+ * This may be called at any core, not necessarily on the one the task will be scheduled on
```
btian1, line 72: seems have some grammar mistakes?
"not necessarily on the one the task will be scheduled on"

src/schedule/zephyr_dp_schedule.c
```diff
@@ -102,8 +102,8 @@ void scheduler_dp_ll_tick(void *receiver_data, enum notify_id event_type, void *
 
 static int scheduler_dp_task_cancel(void *data, struct task *task)
 {
-	(void)(data);
-	k_spinlock_key_t lock_key;
+	struct scheduler_dp_data *dp_sch = (struct scheduler_dp_data *)data;
```
btian1, line 105: I don't see the usage of dp_sch, just for consume *data?

src/schedule/zephyr_dp_schedule.c
```diff
@@ -118,7 +118,8 @@ static int scheduler_dp_task_cancel(void *data, struct task *task)
 
 static int scheduler_dp_task_free(void *data, struct task *task)
 {
-	k_spinlock_key_t lock_key;
+	unsigned int lock_key;
+	struct scheduler_dp_data *dp_sch = (struct scheduler_dp_data *)data;
```
btian1, line 122: same as above.

src/include/sof/schedule/dp_schedule.h
marcinszkudlinski: Even worse, comments have not been changes and are not true anymore
Now the threat must be created on the same core it will run on. 
Comments changed, assert added

PR #7985, opened 2023-07-26
Revert "Audio: SRC: Add safeguard against missing, wrong size or type…

marcinszkudlinski:
… init data"

This reverts commit 4849a71ada60085542722cfc6ca7e82c2b428ecf.

The commit has been merged with CI test fail
The test is now failing constantly making CI unusabe

Either CI test or the code requires changes before merging

FW logs:

[    0.122168] <inf> init: print_version_banner: FW ABI 0x301b000 DBG ABI 0x5003000 tags SOF:v2.5-stable-branch-733-g8683d2d265d1 zephyr:zephyr-v3.4.0-496-g431108d89e17 src hash 0x4b68d789 (ref hash 0x4b68d789)
[    0.122183] <inf> clock: clock_set_freq: clock 0 set freq 400000000Hz freq_idx 2
[    0.122190] <inf> clock: clock_set_freq: clock 1 set freq 400000000Hz freq_idx 2
[    0.122196] <inf> clock: clock_set_freq: clock 2 set freq 400000000Hz freq_idx 2
*** Booting Zephyr OS build zephyr-v3.4.0-496-g431108d89e17 ***
[    0.122378] <inf> main: main: SOF on intel_adsp_ace15_mtpm
[    0.122386] <inf> main: main: SOF initialized
[    0.157836] <inf> ipc: ipc_cmd: rx    : 0x43000000|0x30701000
[    0.160618] <inf> ipc: ipc_cmd: rx    : 0x43000000|0x30801000
[    0.171581] <inf> ipc: ipc_cmd: rx    : 0x44000000|0x3070000c
[    0.171590] <wrn> basefw: basefw_set_large_config: returning success for Set FW_CONFIG without handling it
[    0.174833] <inf> ipc: ipc_cmd: rx    : 0x44000000|0x3070000c
[    0.174846] <wrn> basefw: basefw_set_large_config: returning success for Set FW_CONFIG without handling it
[    0.179795] <inf> ipc: ipc_cmd: rx    : 0x44000000|0x31400008
[    0.181086] <inf> ipc: ipc_cmd: rx    : 0x44000000|0x30600064
[    0.241513] <inf> ipc: ipc_cmd: rx    : 0x11000006|0x0
[    0.241518] <inf> pipe: pipeline_new: pipeline new pipe_id 0 priority 0
[    0.242786] <inf> ipc: ipc_cmd: rx    : 0x40000004|0x15
[...]

PR #8004, opened 2023-08-02
Audio: SRC: Add safeguard against missing, wrong size or type init data

jsarha:
In some error situations the configuration init_data may be NULL, and in such a situations we should fail gracefully and not crash. Also adds check that the IPC message is of correct type and for IPC3 only that it is of correct type.

This is a reincarnation of https://github.com/thesofproject/sof/pull/7830 that was reverted due to CI problems: https://github.com/thesofproject/sof/pull/7985

The message type check is now done for IPC3 only.

### DIFF
--- a/src/audio/src/src.c
+++ b/src/audio/src/src.c
@@ -986,6 +986,18 @@ static int src_init(struct processing_module *mod)
 	struct comp_data *cd = NULL;
 
 	comp_dbg(dev, "src_init()");
+#if CONFIG_IPC_MAJOR_3
+	if (dev->ipc_config.type != SOF_COMP_SRC) {
+		comp_err(dev, "src_init(): Wrong IPC config type %u",
+			 dev->ipc_config.type);
+		return -EINVAL;
+	}
+#endif
+	if (!cfg->init_data || cfg->size != sizeof(cd->ipc_config)) {
+		comp_err(dev, "src_init(): Missing or bad size (%u) init data",
+			 cfg->size);
+		return -EINVAL;
+	}
 
 	/* validate init data - either SRC sink or source rate must be set */
 	if (src_rate_check(cfg->init_data) < 0) {


lgirdwood:
> @lgirdwood what Intel platforms are still using IPC3?
> 

None on main. Only older platforms using stable-v2.2 branch for fixes only.

> What Intel platform completely switched to Zephyr? Do you support IPC3 + Zephyr?
> 
> We plan to deprecate XTOS starting with Q4 on i.MX platforms. But we will still be using IPC3 + Zephyr for a while ( I guess)

All Intel platforms on main. i.e. everything from Tigerlake onwards. All older platforms are on stable v2.2 branch with xtos + IPC3.

lgirdwood (approved):

src/audio/src/src.c
```diff
@@ -987,6 +987,13 @@ static int src_init(struct processing_module *mod)
 
 	comp_dbg(dev, "src_init()");
 
+	if ((IS_ENABLED(CONFIG_IPC_MAJOR_3) && dev->ipc_config.type != SOF_COMP_SRC) ||
+	    !cfg->init_data || cfg->size != sizeof(cd->ipc_config)) {
+		comp_err(dev, "src_init(): Missing or bad size (%u) init data",
```
btian1, line 992: this is complex condition, and the err message can't cover so much cases.
if for bugfix, we can simply add:
#if CONFIG_IPC_MAJOR_3
if (!cfg->init_data) {
     comp_err(dev, "src_init(): Missing or bad size (%u) init data",cfg->size);
    return -EINVAL;
}
#endif

for handle other cases, how about use separate "if"?

jsarha: The IPC config type check is now separate, and inside #if IPC3. Functionally this should be equivalent to the previous version.

btian1 (approved):

singalsu (approved):

kv2019i:
SOFCI TEST

kv2019i:
Looks ready to go, just needs a clean CI run to merge.

cujomalainey (approved):

iuliana-prodan (approved):

dnikodem (approved):

PR #8005, opened 2023-08-02
Fast mode task config + vendor config task

tobonex:
First PR here, open to heavy criticism.

Adding vendor config set and fmt(fast mode task) config as first steps to fmt implementation. Fixes IPC error on fmt config due to no fmt implementation.

### DIFF
--- a/src/audio/kpb.c
+++ b/src/audio/kpb.c
@@ -2384,9 +2384,19 @@ static int kpb_set_large_config(struct comp_dev *dev, uint32_t param_id,
 				uint32_t data_offset,
 				const char *data)
 {
+	/* We can use extended param id for both extended
+	 * and standard param id
+	 */
+	const struct byte_array_simple *ba = (struct byte_array_simple *)data;
+	union ipc4_extended_param_id extended_param_id;
+
 	comp_info(dev, "kpb_set_large_config()");
 
-	switch (param_id) {
+	extended_param_id.full = param_id;
+
+	switch (extended_param_id.part.parameter_type) {
+//	case 1:
+//		return 0;
 	case KP_BUF_CLIENT_MIC_SELECT:
 		return kpb_set_micselect(dev, data, data_offset);
 	default:
--- a/src/audio/kpb.c
+++ b/src/audio/kpb.c
@@ -106,6 +106,8 @@ struct comp_data {
 	uint32_t num_of_in_channels;
 	uint32_t offsets[KPB_MAX_MICSEL_CHANNELS];
 	struct kpb_micselector_config mic_sel;
+	struct kpb_fmt_dev_list fmt_device_list;
+	struct fast_mode_task fmt;
 
 #if CONFIG_AMS
 	uint32_t kpd_uuid_id;
@@ -141,6 +143,19 @@ static inline bool validate_host_params(struct comp_dev *dev,
 					size_t hb_size_req);
 static inline void kpb_change_state(struct comp_data *kpb,
 				    enum kpb_state state);
+/* KpbFastModeTaskModulesList Namespace */
+static inline int AllocFmtModuleListItem(struct kpb_fmt_dev_list *fmt_device_list,
+					 struct comp_dev *mi_ptr, struct comp_dev ***item);
+static void ClearFmtModulesList(struct kpb_fmt_dev_list *fmt_device_list,
+				uint32_t outpin_idx);
+static int PrepareFmtModulesList(struct comp_dev *kpb_dev, uint32_t outpin_idx,
+				 const struct kpb_task_params *modules_to_prepare,
+				 struct comp_dev **last_copier_mi);
+/* FMT Namespace */
+static int RegisterModulesList(struct fast_mode_task *fmt,
+			       struct device_list *new_list, size_t list_idx);
+static int UnregisterModulesList(struct fast_mode_task *fmt,
+				 struct device_list *list_to_remove, size_t list_idx);
 
 static uint64_t kpb_task_deadline(void *data)
 {
@@ -376,6 +391,8 @@ static int kpb_unbind(struct comp_dev *dev, void *data)
 	else
 		kpb->host_sink = NULL;
 
+	/* Clear fmt config */
+	ClearFmtModulesList(&kpb->fmt_device_list, bu->extension.r.src_queue);
 	return 0;
 }
 
@@ -2378,6 +2395,188 @@ static int kpb_set_micselect(struct comp_dev *dev, const void *data,
 	return 0;
 }
 
+int devicelist_push(struct device_list *devlist, struct comp_dev **dev)
+{
+	if (devlist->count != DEVICE_LIST_SIZE) {
+		devlist->devs[devlist->count] = dev;
+		devlist->count++;
+		return 0;
+	}
+	return -EINVAL;
+}
+
+void devicelist_reset(struct device_list *devlist, bool remove_items)
+{
+	/* deallocate items */
+	if (remove_items) {
+		for (int i = 0; i < DEVICE_LIST_SIZE; i++)
+			*devlist->devs[i] = NULL;
+	}
+	/* zero the pointers */
+	for (int i = 0; i < DEVICE_LIST_SIZE; i++)
+		devlist->devs[i] = NULL;
+
+	devlist->count = 0;
+}
+
+static struct comp_dev *get_dev_from_mi_id(uint32_t module_id, uint32_t instance_id)
+{
+	struct comp_dev *dev = NULL;
+	uint32_t comp_id = IPC4_COMP_ID(module_id, instance_id);
+
+	dev = ipc4_get_comp_dev(comp_id);
+	return dev;
+}
+
+static inline int AllocFmtModuleListItem(struct kpb_fmt_dev_list *fmt_device_list,
+					 struct comp_dev *mi_ptr, struct comp_dev ***item)
+{
+	//check if module already added?
+	for (size_t module_slot_idx = 0; module_slot_idx < FAST_MODE_TASK_MAX_MODULES_COUNT;
+	     ++module_slot_idx){
+		if (fmt_device_list->modules_list_item_[module_slot_idx] == mi_ptr)
+			return -EINVAL; // was ADSP_INVALID_REQUEST;
+	}
+	//add module to first available field
+	for (size_t module_slot_idx = 0; module_slot_idx < FAST_MODE_TASK_MAX_MODULES_COUNT;
+	     ++module_slot_idx){
+		if (!fmt_device_list->modules_list_item_[module_slot_idx]) {
+			fmt_device_list->modules_list_item_[module_slot_idx] = mi_ptr;
+			*item = &fmt_device_list->modules_list_item_[module_slot_idx];
+			return 0;
+		}
+	}
+	return -ENOMEM;
+}
+
+static int PrepareFmtModulesList(struct comp_dev *kpb_dev,
+				 uint32_t outpin_idx,
+				 const struct kpb_task_params *modules_to_prepare,
+				 struct comp_dev **last_copier_mi)
+{
+	if (!kpb_dev)
+		return -EINVAL;
+	if (!modules_to_prepare)
+		return -EINVAL;
+	if (!last_copier_mi)
+		return -EINVAL;
+	if (outpin_idx >= KPB_MAX_SINK_CNT)
+		return -EINVAL;
+
+	int ret = 0;
+	struct comp_dev *dev = NULL;
+	struct kpb_fmt_dev_list *fmt_device_list =
+			&((struct comp_data *)comp_get_drvdata(kpb_dev))->fmt_device_list;
+
+	if (modules_to_prepare->number_of_modules != 0) {
+		fmt_device_list->kpb_list_item_[outpin_idx] = kpb_dev;
+		ret = devicelist_push(&fmt_device_list->device_list_[outpin_idx],
+				      &fmt_device_list->kpb_list_item_[outpin_idx]);
+		if (ret < 0)
+			return ret;
+	}
+	for (size_t mod_idx = 0; mod_idx < modules_to_prepare->number_of_modules; ++mod_idx) {
+		dev = get_dev_from_mi_id(modules_to_prepare->dev_ids[mod_idx].module_id,
+					 modules_to_prepare->dev_ids[mod_idx].instance_id);
+
+		if (!dev)
+			return -EINVAL;
+
+		struct comp_dev **new_list_item_ptr;
+
+		ret = AllocFmtModuleListItem(fmt_device_list, dev, &new_list_item_ptr);
+		if (ret < 0)
+			return ret;
+		*new_list_item_ptr = dev;
+		ret = devicelist_push(&fmt_device_list->device_list_[outpin_idx],
+				      new_list_item_ptr);
+		if (ret < 0)
+			return ret;
+	}
+
+	*last_copier_mi = dev;
+
+	return ret;
+}
+
+static void ClearFmtModulesList(struct kpb_fmt_dev_list *fmt_device_list,
+				uint32_t outpin_idx)
+{
+	/* Note: this should be validated in layer above. */
+	assert(outpin_idx < KPB_MAX_SINK_CNT);
+	//ACE
+	devicelist_reset(&fmt_device_list->device_list_[outpin_idx], true);
+}
+
+//struct fast_mode_task fmt;
+
+static int UnregisterModulesList(struct fast_mode_task *fmt,
+				 struct device_list *list_to_remove, size_t list_idx)
+{
+	if (list_to_remove == fmt->device_list_[list_idx]) {
+		fmt->device_list_[list_idx] = NULL;
+		return 0;
+	}
+	if (!fmt->device_list_[list_idx]) {
+		/* Nothing to do here */
+		return 0;
+	}
+	return -EINVAL;
+}
+
+/* Comment from ACE, may be outdated:
+ * Important: function below should be called only from within critical section
+ * (Goto KPB for more details)
+ */
+static int RegisterModulesList(struct fast_mode_task *fmt,
+			       struct device_list *new_list, size_t list_idx)
+{
+	if (list_idx >= ARRAY_SIZE(fmt->device_list_))
+		return -EINVAL;
+
+	/* Check if slot is free */
+	if (!fmt->device_list_[list_idx]) {
+		fmt->device_list_[list_idx] = new_list;
+		return 0;
+	}
+	if (new_list == fmt->device_list_[list_idx]) {
+		/* Already registered. */
+		return 0;
+	}
+	return -EINVAL;//was ADSP_ALREADY_IN_USE;
+}
+
+static int ConfigureFastModeTask(struct comp_dev *kpb_dev, const struct kpb_task_params *cfg,
+				 size_t pin)
+{
+	if (!cfg && pin >= KPB_MAX_SINK_CNT && pin == REALTIME_PIN_ID &&
+	    cfg->dev_ids <= 0)
+		return -EINVAL;
+
+	int ret = 0;
+	/* not sure if this var is needed for anything */
+	struct comp_dev *last_copier_ptr = NULL;
+	struct comp_data *priv_data = (struct comp_data *)comp_get_drvdata(kpb_dev);
+
+	ret = UnregisterModulesList(&priv_data->fmt,
+				    &priv_data->fmt_device_list.device_list_[pin],
+				    pin);
+	assert(ret == 0);
+
+	ClearFmtModulesList(&priv_data->fmt_device_list, pin);
+
+	/* When modules count IS 0 we only need to remove modules from Fast Mode. */
+	if (cfg && cfg->number_of_modules > 0) {
+		if (ret == 0)
+			ret = PrepareFmtModulesList(kpb_dev, pin, cfg, &last_copier_ptr);
+		if (ret == 0)
+			ret = RegisterModulesList(&priv_data->fmt,
+						  &priv_data->fmt_device_list.device_list_[pin],
+						  pin);
+	}
+	return ret;
+}
+
 static int kpb_set_large_config(struct comp_dev *dev, uint32_t param_id,
 				bool first_block,
 				bool last_block,
@@ -2386,17 +2585,41 @@ static int kpb_set_large_config(struct comp_dev *dev, uint32_t param_id,
 {
 	/* We can use extended param id for both extended
 	 * and standard param id
+	 * Also we have to smuggle a struct through the char pointer
 	 */
 	const struct byte_array_simple *ba = (struct byte_array_simple *)data;
 	union ipc4_extended_param_id extended_param_id;
+	int ret = 0;
 
 	comp_info(dev, "kpb_set_large_config()");
 
 	extended_param_id.full = param_id;
 
 	switch (extended_param_id.part.parameter_type) {
-//	case 1:
-//		return 0;
+	case KP_BUF_CFG_FM_MODULE:
+	/* ACE comment:
+	 * Modules count equals 0 is a special case in which we want to clear list for given pin.
+	 * In case of that however dataAs<KpBufferingTaskParams> will return NULL.
+	 * To avoid this we first checking only modules count first and full config after that.
+	 * Other solution would be for driver to allways send 12 bytes even if there is no module
+	 * on list - which is also not very elegant.
+	 */
+	const struct kpb_task_params *cfg = NULL;
+	/* get first dword from payload, dword size field */
+	uint32_t nr_of_modules = *(uint32_t *)ba->data;
+	uint32_t outpin_id = extended_param_id.part.parameter_instance;
+
+	if (nr_of_modules != 0) {
+		cfg = (struct kpb_task_params *)ba->data;
+		if (ret  < 0)
+			return -EINVAL;
+	}
+
+	if (outpin_id >= KPB_MAX_SINK_CNT)
+		return -EINVAL;
+
+	return ConfigureFastModeTask(dev, cfg, outpin_id);
+
 	case KP_BUF_CLIENT_MIC_SELECT:
 		return kpb_set_micselect(dev, data, data_offset);
 	default:
--- a/src/include/ipc4/kpb.h
+++ b/src/include/ipc4/kpb.h
@@ -19,4 +19,15 @@
 struct ipc4_kpb_module_cfg {
 	struct ipc4_base_module_cfg base_cfg;
 } __packed __aligned(4);
+
+enum ipc4_kpb_module_config_params {
+	//! Configure the module ID's which would be part of the Fast mode tasks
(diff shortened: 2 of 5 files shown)


sofci:

gkbldcig:

btian1 (commented):
Please fix CI issues first

src/include/sof/audio/kpb.h
```diff
@@ -159,18 +163,89 @@ struct history_data {
...
+struct kpb_task_params {
+	/* If largeconfigset is set to KP_POS_IN_BUFFER then number of modules must
+	 * correspond to number of modules between kpb and copier attached to hostdma.
+	 * Once draining path is configured, cannot be reinitialized/changed.
+	 */
+	uint32_t        number_of_modules;
+	struct {
+		uint16_t module_id;
+		uint16_t instance_id;
+	}		    dev_ids[1];
```
lgirdwood, line 184: indentation

lgirdwood (commented):
Needs more decription in the commit messages to explain feature and changes.

src/include/ipc4/module.h
```diff
@@ -203,6 +206,35 @@ struct ipc4_module_bind_unbind {
...
+ * the beginning of the structure.
+ */
+struct ipc4_mod_conf_param {
+	//! parameter id
+	uint32_t param_id;
+	//! size of value (in bytes)
+	uint32_t param_size;
+	//! value (padded w/ zeroes to align to dword)
+	uint32_t param_data[1];
+};
```
tmleman: We have struct that represent TLV, use it instead.

https://github.com/thesofproject/sof/blob/main/src/include/sof/tlv.h

tmleman (changes requested):

src/audio/kpb.c
```diff
@@ -2384,9 +2384,17 @@ static int kpb_set_large_config(struct comp_dev *dev, uint32_t param_id,
 				uint32_t data_offset,
 				const char *data)
 {
+	/* We can use extended param id for both extended
+	 * and standard param id
+	 */
+	const struct byte_array_simple *array = (struct byte_array_simple *)data;
```
btian1, line 2596: redundant line 2390?

src/ipc/ipc4/handler.c
```diff
@@ -828,6 +828,100 @@ static int ipc4_get_large_config_module_instance(struct ipc4_message_request *ip
...
+
+			tlv = (struct ipc4_mod_conf_param *)((const uint8_t *)tlv +
+					TLV_DATA_OFFSET + tlv->param_size + padding);
+		}
+		return IPC4_SUCCESS;
+	}
+	/* else, !(init_block && final_block) */
+	const struct ipc4_mod_conf_param *tlv = (struct ipc4_mod_conf_param *)data;
+	struct byte_array_simple array = {
+			(uint8_t *)&data, data_off_size };
```
btian1: should no need use two lines.

src/include/sof/audio/kpb.h
```diff
@@ -159,18 +163,89 @@ struct history_data {
...
+		uint16_t instance_id;
+	} dev_ids[1];
+};
+
+/* fmt namespace: */
+#define FAST_MODE_TASK_MAX_LIST_COUNT 5
+
+struct fast_mode_task {
+	/*! Array of pointers to all module lists to be processed. */
+	struct device_list *device_list_[FAST_MODE_TASK_MAX_LIST_COUNT];
```
lyakh, line 192: can we avoid ending names with an underscore?

src/audio/kpb.c
```diff
@@ -2386,14 +2591,42 @@ static int kpb_set_large_config(struct comp_dev *dev, uint32_t param_id,
...
+		 * very elegant.
+		 */
+		const struct kpb_task_params *cfg = NULL;
+		/* get first dword from payload, dword size field */
+		uint32_t nr_of_modules = *(uint32_t *)array->data;
+		uint32_t outpin_id = extended_param_id.part.parameter_instance;
+
+		if (nr_of_modules != 0) {
+			cfg = (struct kpb_task_params *)array->data;
+			if (ret  < 0)
```
lyakh, line 2621: wrong test

src/audio/kpb.c
```diff
@@ -2378,6 +2399,190 @@ static int kpb_set_micselect(struct comp_dev *dev, const void *data,
...
+	return -EINVAL;//was ADSP_ALREADY_IN_USE;
+}
+
+static int ConfigureFastModeTask(struct comp_dev *kpb_dev, const struct kpb_task_params *cfg,
+				 size_t pin)
+{
+	assert(cfg && pin < KPB_MAX_SINK_CNT && pin != REALTIME_PIN_ID &&
+	       cfg->dev_ids > 0);
+
+	int ret = 0;
```
lyakh, line 2562: superfluous initialisation

src/audio/kpb.c
```diff
@@ -2378,6 +2399,190 @@ static int kpb_set_micselect(struct comp_dev *dev, const void *data,
...
+		return 0;
+	}
+	return -EINVAL;//was ADSP_ALREADY_IN_USE;
+}
+
+static int ConfigureFastModeTask(struct comp_dev *kpb_dev, const struct kpb_task_params *cfg,
+				 size_t pin)
+{
+	assert(cfg && pin < KPB_MAX_SINK_CNT && pin != REALTIME_PIN_ID &&
+	       cfg->dev_ids > 0);
```
lyakh, line 2560: this is IPC data, right? Wrong IPC data shouldn't crash the firmware. Alco `cfg` can be `NULL` here for removing the module, we shouldn't crash in that case

src/audio/kpb.c
```diff
@@ -2378,6 +2399,190 @@ static int kpb_set_micselect(struct comp_dev *dev, const void *data,
...
+
+	int ret = 0;
+	/* not sure if this var is needed for anything */
+	struct comp_dev *last_copier_ptr = NULL;
+	struct comp_data *priv_data = (struct comp_data *)comp_get_drvdata(kpb_dev);
+
+	ret = unregister_modules_list(&priv_data->fmt,
+				      &priv_data->fmt_device_list.device_list_[pin],
+				      pin);
+	assert(ret == 0);
```
lyakh, line 2570: same here, just supplying the same pin number twice or similar would cause an error here?

lyakh (changes requested):
not a complete review, let's fix these issues first. Maybe this should be a draft PR for now

src/ipc/ipc4/handler.c
```diff
@@ -828,6 +828,99 @@ static int ipc4_get_large_config_module_instance(struct ipc4_message_request *ip
...
+		if ((const uint8_t *)tlv >= end_offset) {
+			/* This means that there is indeed nothing to be done here */
+			return IPC4_INVALID_CONFIG_DATA_STRUCT;
+		}
+
+		while ((const uint8_t *)tlv < end_offset) {
+			const struct byte_array_simple array = {
+					(uint8_t *)&tlv->param_data[0], tlv->param_size };
+
+			ret = drv->ops.set_large_config(dev, tlv->param_id, init_block,
```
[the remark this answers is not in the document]
tobonex: Good catch. From what I can see here, set_large_config implementations do not check for `dev == NULL` either , and first thing those do is trying to access  `dev->priv_data`.

The problem is, I based this function on ipc4_get_large_config_module_instance which does the same thing. Here we also write to dev variable only if module_id is not 0 (because base_fw has no dev apparently). 

After searching a bit I saw that base_fw also implements large config set in `basefw_set_large_config` function. It takes comp_dev argument just to be compatible with the interface, but assumes it is NULL as it is not used. 

So, it seems it really can be called with `dev == NULL` in this one case.

src/audio/kpb.c
tobonex: This mistake, and some of the other ones, comes from the fact that I tried to make the code as similar to ACE as I could, assuming that original logic was correct. Looking back, it really seems pretty inconsistent.

ACE used `debug_assert` here, and it seems it did actual assert when debug was false. Not sure what the idea was here. I will just change it to return.

src/ipc/ipc4/handler.c
```diff
@@ -828,6 +831,89 @@ static int ipc4_get_large_config_module_instance(struct ipc4_message_request *ip
...
+	}
+	/* else, !(init_block && final_block) */
+	const struct sof_tlv *tlv = (struct sof_tlv *)data;
+	struct byte_array_simple array = { (uint8_t *)data, data_off_size };
+	uint32_t param_id = 0;
+
+	if (init_block) {
+		// for initial block use param_id from tlv, otherwise use built-in
+		// module instance handling mechanism (there is no default impl of
+		// caching large_param_id - @see LargeConfigSet)
```
kv2019i, line 906: C-style comments here as well

src/audio/kpb.c
```diff
@@ -2378,6 +2399,190 @@ static int kpb_set_micselect(struct comp_dev *dev, const void *data,
...
+static inline int alloc_fmt_module_list_item(struct kpb_fmt_dev_list *fmt_device_list,
+					     struct comp_dev *mi_ptr, struct comp_dev ***item)
+{
+	//check if module already added?
+	for (size_t module_slot_idx = 0; module_slot_idx < FAST_MODE_TASK_MAX_MODULES_COUNT;
+	     ++module_slot_idx){
+		if (fmt_device_list->modules_list_item_[module_slot_idx] == mi_ptr)
+			return -EINVAL; // was ADSP_INVALID_REQUEST;
+	}
+	//add module to first available field
```
kv2019i, line 2447: C++ comment

src/audio/kpb.c
```diff
@@ -2378,6 +2399,190 @@ static int kpb_set_micselect(struct comp_dev *dev, const void *data,
...
+
+	dev = ipc4_get_comp_dev(comp_id);
+#endif
+	return dev;
+}
+
+static inline int alloc_fmt_module_list_item(struct kpb_fmt_dev_list *fmt_device_list,
+					     struct comp_dev *mi_ptr, struct comp_dev ***item)
+{
+	//check if module already added?
```
kv2019i, line 2441: C++ comment

kv2019i (commented):
Mostly minor style issue. I understand this is based on code with C++ background, so some of the conversion artifacts can be omitted, but I did find at least one odd case of error handling.

As for functionality, I don't know kpb well enough to really do meaningful commentary and would welcome a review round by someone more familiar with kpb.

src/include/ipc4/module.h
```diff
@@ -51,6 +54,23 @@ enum sof_ipc4_module_type {
...
+	struct{
+		uint32_t parameter_type     : 8;
+		uint32_t parameter_instance : 24;
+	} part;
+} __packed __aligned(4);
+
+struct byte_array_simple {
+	uint8_t *data;
+	uint32_t size;
+};
```
lgirdwood, line 72: Is byte_array_simple part of the IPC ABI ?

src/audio/kpb.c
```diff
@@ -2386,14 +2591,41 @@ static int kpb_set_large_config(struct comp_dev *dev, uint32_t param_id,
...
+		 * very elegant.
+		 */
+		const struct kpb_task_params *cfg = NULL;
+		/* get first dword from payload, dword size field */
+		uint32_t nr_of_modules = *(uint32_t *)array->data;
+		uint32_t outpin_id = extended_param_id.part.parameter_instance;
+
+		if (nr_of_modules != 0) {
+			cfg = (struct kpb_task_params *)array->data;
+			if (!cfg)
```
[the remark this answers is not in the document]
kv2019i: @tobonex how about this? I see you addressed rest of my concerns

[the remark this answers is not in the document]
tobonex: I will take care of it next time I push

src/include/sof/audio/kpb.h
```diff
@@ -171,6 +171,7 @@ struct kpb_micselector_config {
 	/* channel bit set to 1 implies channel selection */
 	uint32_t mask;
 };
+
```
btian1, line 174: extra no need change

src/include/ipc4/kpb.h
```diff
@@ -19,4 +19,16 @@
...
 } __packed __aligned(4);
+
+/* For the sake of compatibility, do not change IDs, only add new ones.*/
+enum ipc4_kpb_module_config_params {
+	/*! Configure the module ID's which would be part of the Fast mode tasks */
+	KP_BUF_CFG_FM_MODULE = 1,
+	/* Mic selector for client - sets microphone id for real time sink mic selector
+	 * IPC4-compatible ID - please do not change the number
+	 */
+	KP_BUF_CLIENT_MIC_SELECT = 11,
```
btian1, line 30: how about 101? left more extend for this?

src/audio/kpb.c
```diff
@@ -2378,6 +2399,192 @@ static int kpb_set_micselect(struct comp_dev *dev, const void *data,
...
+		return 0;
+	}
+	/* was ADSP_ALREADY_IN_USE */
+	return -EINVAL;
+}
+
+static int ConfigureFastModeTask(struct comp_dev *kpb_dev, const struct kpb_task_params *cfg,
+				 size_t pin)
+{
+	if (pin >= KPB_MAX_SINK_CNT || pin == REALTIME_PIN_ID || cfg->dev_ids <= 0)
```
tobonex, line 2562: In reference FW, there was this check:
`cfg->dev_ids <= 0`
(checking if flexible array member is 0, or null as its value is a memory address in debug)
Can flexible array member be NULL at all? Removing this one.

lyakh: in practice that condition will never be satisfied, and it doesn't really make sense. Currently `dev_ids` is at offset 4 inside `struct kpb_task_params` so, e.g. if `cfg == -4` then `cfg->dev_ids == 0`. If `cfg == -8` **and** the comparison is performed as `signed` then `cfg->dev_ids < 0`. But actually the compiler can even complain about that comparison between a pointer and an integer. You can only compare *compatible pointers* to each other. You can compare two `int *` pointers but IIRC you'll get a warning if you compare `int *` and `char *`

kv2019i:
@tobonex CI is failing on build, can you check:

https://github.com/thesofproject/sof/actions/runs/6151677351/job/16692249787?pr=8005
```
zep_workspace/sof/src/ipc/ipc4/handler.c: In function 'ipc4_set_large_config_module_instance':
/zep_workspace/sof/src/ipc/ipc4/handler.c:1032:67: error: passing argument 8 of 'ipc4_set_vendor_config_module_instance' discards 'const' qualifier from pointer target type [-Werror=discarded-qualifiers]
```

kv2019i (commented):
A few C++'ism, but I can live with those. The CI fail needs to be fixed before merge, and I'd appreciate some comment on why FMT is needed by KPB. Otherwise looks good to go, thanks @tobonex !

src/ipc/ipc4/handler.c
```diff
@@ -823,6 +826,79 @@ static int ipc4_get_large_config_module_instance(struct ipc4_message_request *ip
...
+		 * caching large_param_id - @see LargeConfigSet)
+		 */
+		param_id = tlv->type;
+		array.data += TLV_DATA_OFFSET;
+		array.size = array.size - TLV_DATA_OFFSET;
+		data_off_size -= TLV_DATA_OFFSET;
+	}
+	ret = drv->ops.set_large_config(dev, param_id, init_block,
+		final_block, data_off_size, (uint8_t *)(&array));
+	return ret;
```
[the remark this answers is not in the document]
tobonex: Fixed this, but ret is still used elsewhere.

src/audio/kpb.c
```diff
@@ -2378,22 +2398,202 @@ static int kpb_set_micselect(struct comp_dev *dev, const void *data,
...
+	for (size_t mod_idx = 0; mod_idx < modules_to_prepare->number_of_modules; ++mod_idx) {
+#ifdef CONFIG_IPC_MAJOR_4
+		uint32_t comp_id = IPC4_COMP_ID(modules_to_prepare->dev_ids[mod_idx].module_id,
+						modules_to_prepare->dev_ids[mod_idx].instance_id);
+
+		dev = ipc4_get_comp_dev(comp_id);
+#endif
+
+		if (!dev)
+			return -EINVAL;
```
lyakh, line 2482: so for IPC3 this function will return `-EINVAL`, can we just make that explicit somehow?

src/audio/kpb.c
```diff
@@ -2378,22 +2398,202 @@ static int kpb_set_micselect(struct comp_dev *dev, const void *data,
...
+		ret = alloc_fmt_module_list_item(fmt_device_list, dev, &new_list_item_ptr);
+		if (ret < 0)
+			return ret;
+		*new_list_item_ptr = dev;
+		ret = devicelist_push(&fmt_device_list->device_list[outpin_idx],
+				      new_list_item_ptr);
+		if (ret < 0)
+			return ret;
+	}
+	return ret;
```
lyakh, line 2495: could do `return 0` here and remove initialisation above

src/audio/kpb.c
```diff
@@ -2378,22 +2398,202 @@ static int kpb_set_micselect(struct comp_dev *dev, const void *data,
...
+	case KP_BUF_CFG_FM_MODULE: {
+		/* Modules count equals 0 is a special case in which we want to clear list for
+		 * given pin. Reference FW also allowed for cfg/data to be NULL, but this is no
+		 * longer the case.
+		 */
+		const struct kpb_task_params *cfg = (struct kpb_task_params *)data;
+		uint32_t outpin_id = extended_param_id.part.parameter_instance;
+
+		if (outpin_id >= KPB_MAX_SINK_CNT)
+			return -EINVAL;
```
lyakh, line 2593: this is checked twice - here and in the function. I'd remove this check.

tobonex:
Refactored some of the code so that it makes more sense:
- removed `struct byte_array_simple` that I previously added - managed without it
- removed some duplicated checks
- `configure_fast_mode_task` function no longer accepts null in cfg argument as it made error checking messy

src/audio/kpb.c
tobonex: Yes, I do not support IPC3 here yet.  About initialisation of dev and the check - I could put those into the #ifdef, and if IPC4 is not defined just return in #else clause . Also I could add comment saying that  IPC3 support is not here yet. Does it sound good?

src/audio/kpb.c
```diff
@@ -2378,6 +2399,192 @@ static int kpb_set_micselect(struct comp_dev *dev, const void *data,
...
+{
+	struct comp_dev *dev = NULL;
+	/* for now considering only IPC4 */
+#ifdef CONFIG_IPC_MAJOR_4
+	uint32_t comp_id = IPC4_COMP_ID(module_id, instance_id);
+
+	dev = ipc4_get_comp_dev(comp_id);
+#endif
+	return dev;
+}
```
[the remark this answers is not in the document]
tobonex: Yeah, used it only once anyway, removed it.

src/ipc/ipc4/handler.c
```diff
@@ -823,6 +826,79 @@ static int ipc4_get_large_config_module_instance(struct ipc4_message_request *ip
...
+			ret = drv->ops.set_large_config(dev, tlv->type, init_block,
+				final_block, data_off_size, (uint8_t *)(&array));
+			if (ret < 0) {
+				ipc_cmd_err(&ipc_tr, "failed to set large_config_module_instance %x : %x",
+					    (uint32_t)module_id, (uint32_t)instance_id);
+				return IPC4_INVALID_RESOURCE_ID;
+			}
+			/* Align up to a multiple of 4 */
+			tlv = (struct sof_tlv *)((const uint8_t *)tlv +
+					TLV_DATA_OFFSET + ALIGN_UP(tlv->length, 4));
```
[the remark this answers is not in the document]
tobonex: this `tlv_next` looks convenient, but...
is does not have the `ALIGN_UP`. Reference FW aligned it up, so I guess that we also need it here. I do not know all of the possible IPC cases so I'm not sure if it's actually necessary or not.

src/include/ipc4/kpb.h
tobonex: It is defined as 11 in reference FW. There are some IPC commands in-between those two that are not defined in sof yet.

src/audio/kpb.c
```diff
@@ -2378,22 +2398,202 @@ static int kpb_set_micselect(struct comp_dev *dev, const void *data,
...
+	int ret = unregister_modules_list(&priv_data->fmt,
+					  &priv_data->fmt_device_list.device_list[pin],
+					  pin);
+	if (ret)
+		return -EINVAL;
+
+	clear_fmt_modules_list(&priv_data->fmt_device_list, pin);
+
+	/* When modules count IS 0 we only need to remove modules from Fast Mode. */
+	if (cfg->number_of_modules > 0) {
```
lyakh, line 2560: the function itself also has checks for `cfg->number_of_modules > 0`  and does nothing if that isn't satisfied. I'm remove this and other already known to hold checks from the function - see comment there

src/audio/kpb.c
```diff
@@ -2378,22 +2398,202 @@ static int kpb_set_micselect(struct comp_dev *dev, const void *data,
...
+static int prepare_fmt_modules_list(struct comp_dev *kpb_dev,
+				    uint32_t outpin_idx,
+				    const struct kpb_task_params *modules_to_prepare)
+{
+	if (!kpb_dev)
+		return -EINVAL;
+	if (!modules_to_prepare)
+		return -EINVAL;
+	if (outpin_idx >= KPB_MAX_SINK_CNT)
+		return -EINVAL;
```
lyakh, line 2459: All of the above checks and the below `modules_to_prepare->number_of_modules != 0` are checked by the caller. I think they can be removed here, or you can turn them into assert()s with a comment

src/audio/kpb.c
lyakh: IPC3 isn't supported at all by this, right? So I'd really put something like
```
if (IS_ENABLED(CONFIG_IPC_MAJOR_3))
    return -EINVAL;
```
somewhere very early, into some preparation function without which the whole fast mode task won't work

src/ipc/ipc4/handler.c
```diff
@@ -823,6 +826,79 @@ static int ipc4_get_large_config_module_instance(struct ipc4_message_request *ip
 	return ret;
 }
 
+static int ipc4_set_vendor_config_module_instance(uint32_t module_id,
+						  uint32_t instance_id,
+						  bool init_block,
+						  bool final_block,
+						  uint32_t data_off_size,
+						  const char *data,
```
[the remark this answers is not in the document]
lyakh: I'd say we have mostly 3 options for this:
1. `void *` - because functions like `write(2)` use it, but arithmetics is difficult with it
2. `char *` - because `char` is simple, everybody knows, that it's 1 byte...
3. `uint8_t *` - because it's explicitly a pointer to a byte array

SOF uses all 3 options in different cases. In general I'd probably use (1), but since we already use them all... Feel free to make your choice, my comment was a proposal, not a hard requirement.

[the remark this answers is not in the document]
tobonex: Leaving this one as it is for now.

tobonex:
> A few C++'ism, but I can live with those. The CI fail needs to be fixed before merge, and I'd appreciate some comment on why FMT is needed by KPB. Otherwise looks good to go, thanks @tobonex !

Originally, in case if WoV event, KPB used FMT to drain history buffer and send this data to Host ASAP. Normal tasks were too slow. We still need it for the same reason.

src/audio/kpb.c
tobonex: Left this one, removed other ones.

src/audio/kpb.c
```diff
@@ -2378,6 +2399,192 @@ static int kpb_set_micselect(struct comp_dev *dev, const void *data,
...
+
+		ret = alloc_fmt_module_list_item(fmt_device_list, dev, &new_list_item_ptr);
+		if (ret < 0)
+			return ret;
+		*new_list_item_ptr = dev;
+		ret = devicelist_push(&fmt_device_list->device_list[outpin_idx],
+				      new_list_item_ptr);
+		if (ret < 0)
+			return ret;
+	}
```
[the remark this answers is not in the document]
tobonex: I didn't read that much into the original FMT functionality code yet, but I do have some info about it though. In reference FW, executing the FMT also triggered processing of some LL modules whenever there are available DSP cycles. I'm guessing those modules are the ones provided in the IPC here.

I agree that the way the lists were (and are) handled was kind of confusing, but I'm not sure yet what can I refactor here. It's either some important logic, or just some C++'ism.

The idea for now was to implement it up to this point so that the tests related to the FMT config IPC pass. After FMT functionality actually gets implemented it will be easier to see if we can simplify the config.

src/audio/kpb.c
tobonex: Added this.
I'm also considering putting my new kpb.c functions in `#ifdef CONFIG_IPC_MAJOR_4` so that it's more explicit (instead of just this part).

lyakh: > I'm also considering putting my new kpb.c functions in `#ifdef CONFIG_IPC_MAJOR_4` so that it's more explicit (instead of just this part).

I think yes, that would be the best for now. You do need a preprocessor check there because of calling an IPC4 specific function, so just using `IS_ENABLED(CONFIG_IPC_MAJOR_3)` unfortunately isn't enough - that wasn't a very good idea, sorry. So, yes, I think putting everything under `#if IPC4` would be the best option ATM

kv2019i:
@tobonex wrote:
> Originally, in case if WoV event, KPB used FMT to drain history buffer and send this data to Host ASAP. Normal tasks were too slow. We still need it for the same reason.

Ack, thanks, this is a sufficient description.

src/include/sof/audio/kpb.h
```diff
@@ -159,18 +163,88 @@ struct history_data {
...
+ *		  |
+ *		dev_list.modules_list_item(comp_dev* )
+ *		  ^
+ *		  |
+ *		dev_list.device_list(comp_dev**)
+ *		  ^
+ *		  |
+ *		fmt.device_list(device_list*)
+ *
+ *
```
lyakh, line 235: 2 empty lines above and below the description should be removed IMHO

src/include/sof/audio/kpb.h
```diff
@@ -159,18 +163,88 @@ struct history_data {
...
+ *		COMP_DEVS
+ *		  ^
+ *		  |
+ *		dev_list.modules_list_item(comp_dev* )
+ *		  ^
+ *		  |
+ *		dev_list.device_list(comp_dev**)
+ *		  ^
+ *		  |
+ *		fmt.device_list(device_list*)
```
lyakh, line 233: not sure what this structure means. Are these elements of the `struct device_list::devs[]` array above? Maybe we can find some easier (at least for me) representation. If this is an array, maybe arrange entries in lines
```
array[]
    +-------> element 1
    +-------> element 2
    +...
    +-------> element N
```
or something similar. The above looks to me like things pointing to each other, or calling each other, or similar

lyakh (approved):
comments have been addressed, thanks! Haven't checked what exactly changed from previous versions, but at least locations, that I commented on, look better now

PR #8008, opened 2023-08-04
perf-cnt: add peak mcps count

btian1:
add peak mcps counter to indicate which period have max mcps.

### DIFF
--- a/src/include/sof/audio/component.h
+++ b/src/include/sof/audio/component.h
@@ -251,11 +251,12 @@ enum {
 		  (uint32_t)((pcd)->cpu_delta_peak))
 
 #define comp_perf_avg_info(pcd, comp_p)					\
-	comp_info(comp_p, "perf comp_copy samples %u period %u cpu avg %u peak %u",\
+	comp_info(comp_p, "perf comp_copy samples %u period %u cpu avg %u peak %u %u",\
 		  (uint32_t)((comp_p)->frames),            \
 		  (uint32_t)((comp_p)->period),			    \
 		  (uint32_t)((pcd)->cpu_delta_sum),			\
-		  (uint32_t)((pcd)->cpu_delta_peak))
+		  (uint32_t)((pcd)->cpu_delta_peak),			\
+		  (uint32_t)((pcd)->peak_mcps_period_cnt))
 
 /** @}*/
 
--- a/xtos/include/sof/lib/perf_cnt.h
+++ b/xtos/include/sof/lib/perf_cnt.h
@@ -17,14 +17,24 @@
 #include <rtos/timer.h>
 
 struct perf_cnt_data {
+	/* current platform clock */
 	uint32_t plat_ts;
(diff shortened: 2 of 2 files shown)


xtos/include/sof/lib/perf_cnt.h
```diff
@@ -25,6 +25,8 @@ struct perf_cnt_data {
 	uint32_t cpu_delta_peak;
 	uint32_t cpu_delta_sum;
 	uint32_t sample_cnt;
+	/* indicate which period have peak mcps in current module */
```
mwasko, line 28: Indicate which period or sample when the peak occurred? To keep it aligned with existing perf_cnt code I would suggest  to use 'sample' word or rename sample_cnt to period_cnt.

xtos/include/sof/lib/perf_cnt.h
```diff
@@ -25,6 +25,8 @@ struct perf_cnt_data {
...
+	uint32_t peak_mcps_cnt;
```
mwasko, line 29: This variable name suggests that we count peak_mcps occurrence. While there is a comment that explain actual meaning the name is still a bit confusing. Maybe you could change it to something like 'peak_mcps_sample' ?

xtos/include/sof/lib/perf_cnt.h
lgirdwood: @btian1 its probably worth adding a comment to each member in this structure so that its easy to follow (since the structure is now getting busier).

src/include/sof/audio/component.h
```diff
@@ -251,10 +251,11 @@ enum {
 		  (uint32_t)((pcd)->cpu_delta_peak))
 
 #define comp_perf_avg_info(pcd, comp_p)					\
-	comp_info(comp_p, "perf comp_copy samples %u period %u cpu avg %u peak %u",\
+	comp_info(comp_p, "perf comp_copy samples %u period %u cpu avg %u peak %u %u",\
 		  (uint32_t)((comp_p)->frames),            \
 		  (uint32_t)((comp_p)->period),			    \
 		  (uint32_t)((pcd)->cpu_delta_sum),			\
+		  (uint32_t)((pcd)->peak_mcps_cnt),			\
```
aiChaoSONG, line 258: let's move this peak_mcps_cnt to the end if you really want to print it. with an analysis tool, this looks not very necessary.

btian1: Seppo's script is based on this, :).
move to end is reasonable, it will not break your current script.

xtos/include/sof/lib/perf_cnt.h
btian1: ok

singalsu (approved):

kv2019i:
@aiChaoSONG good to go?

lgirdwood (approved):

PR #8009, opened 2023-08-04
Update topology parser with IPC4 support

ranj063:


lgirdwood (approved):
Thanks @ranj063 not be long now until testbench can use IPC4. @singalsu pls review.

tools/tplg_parser/include/tplg_parser/topology.h
```diff
@@ -38,11 +39,52 @@ struct sof;
...
+ * @audio_fmt: audio format for the pin
+ *
+ * This structure can be used for both output or input pins and the pin_index is relative to the
+ * pin type i.e output/input pin
+ */
+struct sof_ipc4_pin_format {
+	uint32_t pin_index;
+	uint32_t buffer_size;
+	struct sof_ipc4_audio_format audio_fmt;
+};
```
kv2019i, line 64: @ranj063 audio_format is packed but not this one, is this safe?

kv2019i (approved):
Looks good, one minor question, but nothing preventing a merge.

singalsu (approved):

tools/tplg_parser/include/tplg_parser/topology.h
```diff
@@ -206,5 +206,7 @@ bool tplg_is_valid_priv_size(size_t size_read, size_t priv_size,
 int tplg_create_object(struct tplg_context *ctx,
 		       const struct sof_topology_module_desc *desc, int num_desc,
 		       const char *name, void *object, size_t max_object_size);
-
+int sof_parse_token_sets(void *object, const struct sof_topology_token *tokens,
+			 int count, struct snd_soc_tplg_vendor_array *array,
+			 int priv_size, int num_sets, int object_size);
```
plbossart, line 254: I must admit I don't even know what the tplg_parser is or does, so stopping the review here. Way beyond my understanding.

ranj063: @plbossart the tplg_parser is a user-space library for parsing the topologies and it is used by the testbench today. And it will also be utilized for the alsa plugin that we are current working on.

tools/tplg_parser/include/tplg_parser/topology.h
ranj063: @kv2019i yes it is OK.  this struct is only used to parse and save the topology audio formats info. It is never sent in the IPC payload so it doesnt have to be packed.

kv2019i:
@ranj063 Test results look pretty ok. Only pm fail in https://sof-ci.01.org/sofpr/PR8009/build11406/devicetest/index.html and known pause-resume fail https://sof-ci.01.org/sofpr/PR8009/build11407/devicetest/index.html . So that leaves with the system/merge/build fail in https://sof-ci.01.org/sof-pr-viewer/#/build/PR8009/build12565732 . Can you check that?

wszypelt:
@ranj063 @lgirdwood there was a problem with building the FW, we are investigating what is going on

PR #8012, opened 2023-08-08
Mixin_out: remove remap mode from sof mixin out module

btian1:
there is no user case to support remapping in SOF, remove it to save code size and benefit performance.

### DIFF
--- a/src/audio/mixin_mixout/mixin_mixout.c
+++ b/src/audio/mixin_mixout/mixin_mixout.c
@@ -75,7 +75,6 @@ struct mixin_sink_config {
 /* mixin component private data */
 struct mixin_data {
 	normal_mix_func normal_mix_channel;
-	remap_mix_func remap_mix_channel;
 	mute_func mute_channel;
 	struct mixin_sink_config sink_config[MIXIN_MAX_SINKS];
 };
@@ -532,7 +531,6 @@ static int mixin_reset(struct processing_module *mod)
 	comp_dbg(dev, "mixin_reset()");
 
 	mixin_data->normal_mix_channel = NULL;
-	mixin_data->remap_mix_channel = NULL;
 	mixin_data->mute_channel = NULL;
 
 	return 0;
@@ -670,7 +668,6 @@ static int mixin_prepare(struct processing_module *mod,
 	case SOF_IPC_FRAME_S24_4LE:
 	case SOF_IPC_FRAME_S32_LE:
 		md->normal_mix_channel = normal_mix_get_processing_function(fmt);
-		md->remap_mix_channel = remap_mix_get_processing_function(fmt);
 		md->mute_channel = mute_mix_get_processing_function(fmt);
 		break;
 	default:
@@ -678,7 +675,7 @@ static int mixin_prepare(struct processing_module *mod,
 		return -EINVAL;
 	}
 
-	if (!md->normal_mix_channel || !md->remap_mix_channel || !md->mute_channel) {
+	if (!md->normal_mix_channel || !md->mute_channel) {
 		comp_err(dev, "have not found the suitable processing function");
 		return -EINVAL;
 	}
--- a/src/audio/mixin_mixout/mixin_mixout.c
+++ b/src/audio/mixin_mixout/mixin_mixout.c
@@ -182,46 +182,17 @@ static int mix_and_remap(struct comp_dev *dev, const struct mixin_data *mixin_da
 
 	sink_config = &mixin_data->sink_config[sink_index];
 
-	if (sink_config->mixer_mode == IPC4_MIXER_NORMAL_MODE) {
-		/* Mix streams. mix_channel() is reused here to mix streams, not individual
-		 * channels. To do so, (multichannel) stream is treated as single channel:
-		 * channel count is passed as 1, channel index is 0, frame indices (start_frame
-		 * and mixed_frame) and frame count are multiplied by real stream channel count.
-		 */
-		mixin_data->normal_mix_channel(sink, start_frame * audio_stream_get_channels(sink),
-					       mixed_frames * audio_stream_get_channels(sink),
-					       source,
-					       frame_count * audio_stream_get_channels(sink),
-					       sink_config->gain);
-	} else if (sink_config->mixer_mode == IPC4_MIXER_CHANNEL_REMAPPING_MODE) {
-		int i;
+	/* Mix streams. mix_channel() is reused here to mix streams, not individual
+	 * channels. To do so, (multichannel) stream is treated as single channel:
+	 * channel count is passed as 1, channel index is 0, frame indices (start_frame
+	 * and mixed_frame) and frame count are multiplied by real stream channel count.
+	 */
+	mixin_data->normal_mix_channel(sink, start_frame * audio_stream_get_channels(sink),
+				       mixed_frames * audio_stream_get_channels(sink),
+				       source,
+				       frame_count * audio_stream_get_channels(sink),
+				       sink_config->gain);
 
-		for (i = 0; i < audio_stream_get_channels(sink); i++) {
-			uint8_t source_channel =
-				(sink_config->output_channel_map >> (i * 4)) & 0xf;
-
-			if (source_channel == 0xf) {
-				mixin_data->mute_channel(sink, i, start_frame, mixed_frames,
-							 frame_count);
-			} else {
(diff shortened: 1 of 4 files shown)


kv2019i (approved):
Good for me, but I'd like @mwasko @marcinszkudlinski  @mmaka1 to ack this removal.

lgirdwood (commented):
@mwasko do we need remap ?

mwasko:
@btian1 I understand that today we do not have any usecases for remapping at Mixin_out, but since it has already been implemented and enabled with internal tests then maybe it would be good to keep it under some kconfig flag, just in case we will need it in future?

marcinszkudlinski:
agree with @mwasko, 
I suggest put ```} else if (sink_config->mixer_mode == IPC4_MIXER_CHANNEL_REMAPPING_MODE)``` and following lines under ifdef, leave all ```remap_mix_channel_s16``` etc. functions in compilation. If there's no references to them, the linker wont include them in final binary, but they still will be compiled (and verified if compilation still works)

btw. what is the purpose of removal?

btian1:
> agree with @mwasko, I suggest put `} else if (sink_config->mixer_mode == IPC4_MIXER_CHANNEL_REMAPPING_MODE)` and following lines under ifdef, leave all `remap_mix_channel_s16` etc. functions in compilation. If there's no references to them, the linker wont include them in final binary, but they still will be compiled (and verified if compilation still works)
> 
> btw. what is the purpose of removal?

there is no user case or requirement for remapping in current SOF, remove it would save code size and benefit for mixin_mixout performance.

If use Kconfig, will make code looks not so integrity, we can add it back in future if there is a need.

andrula-song:
> agree with @mwasko, I suggest put `} else if (sink_config->mixer_mode == IPC4_MIXER_CHANNEL_REMAPPING_MODE)` and following lines under ifdef, leave all `remap_mix_channel_s16` etc. functions in compilation. If there's no references to them, the linker wont include them in final binary, but they still will be compiled (and verified if compilation still works)
> 
> btw. what is the purpose of removal?

i remember we used to decide to remove the remap mode code while our last performance optimization one month ago.

marcinszkudlinski (approved):

src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -75,7 +75,6 @@ struct mixin_sink_config {
 /* mixin component private data */
 struct mixin_data {
 	normal_mix_func normal_mix_channel;
-	remap_mix_func remap_mix_channel;
```
lyakh, line 78: should we remove the `typedef` for `remap_mix_func` too?

lyakh (approved):

btian1:
SOFCI TEST

src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -182,46 +182,17 @@ static int mix_and_remap(struct comp_dev *dev, const struct mixin_data *mixin_da
...
-	} else {
-		comp_err(dev, "Unexpected mixer mode: %d", sink_config->mixer_mode);
-		return -EINVAL;
-	}
+	/* Mix streams. mix_channel() is reused here to mix streams, not individual
+	 * channels. To do so, (multichannel) stream is treated as single channel:
+	 * channel count is passed as 1, channel index is 0, frame indices (start_frame
+	 * and mixed_frame) and frame count are multiplied by real stream channel count.
+	 */
+	mixin_data->normal_mix_channel(sink, start_frame * audio_stream_get_channels(sink),
```
serhiy-katsyuba-intel, line 189: normal_mix_channel() originally was implemented to be used in both normal mode and channel remapping mode. It mixed one channel from one stream with another channel from another stream. In normal mode (without remapping, streams have same number of channels) fake channel count was used equal to 1: so streams were treated as single channel streams and were mixed entirely at one call. Hence that comment describing this trick.
However, I see, some time ago that channel mixing feature was removed and current version of normal_mix_channel() only mixes entire streams.
So that comment above can be removed (should be removed previously). That function could also be renamed, for example, to just "mix()" as now it has nothing to do with channels (it iterates by samples and does not care about channels) and now "normal" is the only supported mode.

src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -182,46 +182,17 @@ static int mix_and_remap(struct comp_dev *dev, const struct mixin_data *mixin_da
...
+				       mixed_frames * audio_stream_get_channels(sink),
+				       source,
+				       frame_count * audio_stream_get_channels(sink),
```
serhiy-katsyuba-intel, line 192: I see a problem here introduced earlier not by this PR. The declaration of
normal_mix_channel_s16() in mixin_mixout_generic.c is:

static void normal_mix_channel_s16(struct audio_stream __sparse_cache *sink, int32_t start_frame,
				   int32_t mixed_frames,
				   const struct audio_stream __sparse_cache *source,
				   int32_t frame_count, uint16_t gain)

However, for frame_count parameter frame_count * audio_stream_get_channels(sink) is passed by the caller. So that is a sample count, not frame_count. The code only works properly because normal_mix_channel_s16() treats frame_count as sample_count. Anyway, that is very confusing and all normal_mix_channel_sXX() should be fixed with renaming frame_count with sample_count (same problem with start_frame and mixed_frames parameters which are also treated as start_sample and mixed_samples). In addition, since after some previous modification normal_mix_channel_s16() has nothing to do with channels (it iterates by samples and mixes entire streams) that function could be renamed to just mix_s16().

This is not a problem introduced by this PR and could be addressed in a separate PR.

src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -182,46 +182,17 @@ static int mix_and_remap(struct comp_dev *dev, const struct mixin_data *mixin_da
...
-							      source, source_channel,
-							      audio_stream_get_channels(source),
-							      frame_count, sink_config->gain);
-			}
-		}
-	} else {
-		comp_err(dev, "Unexpected mixer mode: %d", sink_config->mixer_mode);
-		return -EINVAL;
-	}
+	/* Mix streams. mix_channel() is reused here to mix streams, not individual
```
serhiy-katsyuba-intel, line 184: This function "mix_and_remap()" now only mixes, so could be also renamed to just mix().

serhiy-katsyuba-intel:
@mwasko , @marcinszkudlinski , I see remapping feature was already broken some time ago. For example, normal_mix_channel_s16() no longer mixes individual channels but mixes entire streams (despite what name suggests). So remapping feature currently is broken anyway.

serhiy-katsyuba-intel:
> i remember we used to decide to remove the remap mode code while our last performance optimization one month ago.

Remapping feature forced to iterate over frames not samples. Iterating over frames uses additional divisions (by channel count) which I thought might take many CPU cycles.

However, (1) we came to a conclusion that division is not that terribly heavy and (2) this PR does not eliminate iteration by frames.

But anyway, I would suggest to proceed with with PR just to make code simple. Remapping feature was already broken some time ago and, probably, the complexity of mixin/mixout led to this.

src/audio/mixin_mixout/mixin_mixout.c
lgirdwood: adding @singalsu and @andrula-song for this new remapping problem. Do we need a new issue to track ? If so, lets add.

lgirdwood:
> > i remember we used to decide to remove the remap mode code while our last performance optimization one month ago.
> 
> Remapping feature forced to iterate over frames not samples. Iterating over frames uses additional divisions (by channel count) which I thought might take many CPU cycles.
> 
> However, (1) we came to a conclusion that division is not that terribly heavy and (2) this PR does not eliminate iteration by frames.
> 
> But anyway, I would suggest to proceed with with PR just to make code simple. Remapping feature was already broken some time ago and, probably, the complexity of mixin/mixout led to this.

@serhiy-katsyuba-intel if gopod for you, pls approve and we can proceed and fix the remmaping as a new PR.

serhiy-katsyuba-intel (approved):
Approved the PR. However, it would be nice to also remove mute_channel() implementations as this was only used in channel remapping mode and now this functionality is left in code unused.

lgirdwood:
> Approved the PR. However, it would be nice to also remove mute_channel() implementations as this was only used in channel remapping mode and now this functionality is left in code unused.

@singalsu @andrula-song pls factor this into remapping work.

PR #8019, opened 2023-08-09
pipeline: register KCPS consumption and enable dynamic clock control

abonislawski:
This will enable dynamic clock control based on KCPS budget.

Mostly cherry-picked from mtl-005-drop-stable + kconfig

pipeline: Register and unregister pipelines CPS consumption on run/pause
+Squashed commits:
https://github.com/thesofproject/sof/commit/924f0d26db2cc68c819185dd42cd491de72f8b5d pipeline: print warning if 0 KCPS received

https://github.com/thesofproject/sof/commit/7bb6e7ef3ba93d6406226a5fbe8476f0dc1098c7 pipeline: register CPS consumption for a proper core

https://github.com/thesofproject/sof/commit/5f0c4e4c3160252d3794feec7d1ac1303f5f9cd1 pipeline: make CPC data "opt-in" with fallback

platform: register basefw CPS consumption on boot

+Squashed commit:
https://github.com/thesofproject/sof/commit/38b4cb1475fb3f3828cbd7a117ec020ae2f57a12 platform: mtl: lowest clock as default

boards: mtl: enable KCPS dynamic clock


src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+#endif
 	pipe_info(p, "pipe trigger cmd %d", cmd);
 
 	p->trigger.aborted = false;
 
 	switch (cmd) {
 	case COMP_TRIGGER_PAUSE:
+#if CONFIG_KCPS_DYNAMIC_CLOCK_CONTROL
+		/* dirty - dont add kcps on fallthrough */
+		clocks_handled = true;
```
kv2019i, line 376: This indeed pops out a bit, but good to have the comment.

src/platform/Kconfig
```diff
@@ -385,6 +385,13 @@ config CAVS_USE_LPRO_IN_WAITI
 	  After waiti exit clock source will be restored.
 	  Choose n if unclear.
 
+config KCPS_DYNAMIC_CLOCK_CONTROL
+	bool "Use KCPS budget to determine DSP clock"
```
kv2019i, line 389: I think we should explain KCPS at least once in Kconfig, either in summary or in the help text. I'd go with "Use compute budget expressed in Kilo Cycles Per Second (KCPS) to determine DSP clock".

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+
+	if (!comp_is_single_pipeline(current, ppl_data->start)) {
+		pipe_dbg(ppl_data->p, "add_pipeline_cps_consumption(), current is from another pipeline");
+		return 0;
+	}
+
+	/* complete component init */
+	current->pipeline = ppl_data->p;
+
+	/* modules created throug module adapter have different priv_data */
```
kv2019i, line 292: typo: through

kv2019i (commented):
Thanks @abonislawski ! Code looks good, a few minor nits for the documentation.

tmleman (approved):

src/platform/intel/ace/platform.c
```diff
@@ -80,18 +81,22 @@ static struct pm_notifier pm_state_notifier = {
 	.state_exit = cpu_notify_state_exit,
 };
 
+#if CONFIG_KCPS_DYNAMIC_CLOCK_CONTROL
+/* Value to be determined experimentaly */
+#define BASE_CPS_USAGE 10000
+#else
+#define BASE_CPS_USAGE (CLK_MAX_CPU_HZ / 1000)
+#endif
```
lgirdwood, line 89: Would this be platform specific or does it make sense to use Kconfig here ?

lgirdwood (commented):
Good stuff - just a few opens.

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
 	} while (true);
 }
 
+#if CONFIG_KCPS_DYNAMIC_CLOCK_CONTROL
+static int add_pipeline_cps_consumption(struct comp_dev *current,
```
btian1, line 274: how to deal with peak pipeline consumption? it normally 3 times than average.
avg 1275 peak 3891

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+		cd = comp_get_drvdata(current);
+	} else {
+		struct processing_module *mod = comp_get_drvdata(current);
+		struct module_data *md = &mod->priv;
+		cd = &md->cfg.base_cfg;
+	}
+
+	if (cd->cpc == 0) {
+		/* Use maximum clock budget, assume 1ms chunk size */
+		cd->cpc = CLK_MAX_CPU_HZ / 1000;
```
btian1, line 303: there is a fix to ignore cpc and set to default max always, I would support that solution due to the peak mcps for pipeline still not resolved：#8007

src/platform/intel/ace/platform.c
btian1: CLK_MAX_CPU_HZ this should be platform specific

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -88,3 +88,5 @@ CONFIG_CLOCK_CONTROL=y
 CONFIG_DEBUG_COREDUMP=y
 CONFIG_DEBUG_COREDUMP_BACKEND_INTEL_ADSP_MEM_WINDOW=y
 CONFIG_DEBUG_COREDUMP_MEMORY_DUMP_MIN=y
+
+CONFIG_KCPS_DYNAMIC_CLOCK_CONTROL=y
```
btian1, line 92: I would suggest N this here for now, until peak mcps rootcaused, it is 3 times larger than avg.

src/platform/intel/ace/platform.c
abonislawski: it could be platform specific but probably we can move this to more appropriate place

src/audio/pipeline/pipeline-stream.c
abonislawski: This is a w/a, not a fix. On main we can simply disable CONFIG_KCPS_DYNAMIC_CLOCK_CONTROL to ignore CPC so we dont need any additional second layer of ignoring cpc

app/boards/intel_adsp_ace15_mtpm.conf
abonislawski: Yes, we can drop this commit before merge but for now its good to turn it on and test in CI

src/audio/pipeline/pipeline-stream.c
kv2019i: @btian1 It's a good question, but I think the only reasonably solution is to optimize the peaks out (both the absolute peak value, and the frequency they occur). We do have buffering in SRAM to survive some spikes in scheduling. But obviously if the peaks are too high and/or too frequent, this won't work. This is a bug really then. You need to either fix the code that caues the peak, or adjust the CPS data in manifest.

app/boards/intel_adsp_ace15_mtpm.conf
kv2019i: @btian1 Let's turn this on, so we can test this. The peak is not necessarily a problem if the SRAM buffering we have can cover it. If not, then we need to anyways address this, either fix the rootcause for peak and/or increase the budgets in manifest.

btian1: ok

btian1 (approved):

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+		tr_warn(pipe,
+			"0 CPS requested for module: %#x, core: %d using safe max CPC: %u",
+			current->ipc_config.id, ppl_data->p->core, cd->cpc);
+	}
+
+	int kcps = cd->cpc * 1000 / ppl_data->p->period;
+	tr_info(pipe, "Registering KCPS consumption: %d, core: %d", kcps, ppl_data->p->core);
+	core_kcps_adjust(ppl_data->p->core, kcps);
```
ujfalusi, line 311: If none of the module instances have recorded CPC value and the we have 6 components active then the kcps will overflow to negative as the `kcps_consumption` is an `int` array...

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+	if (cd->cpc == 0) {
+		/* Use maximum clock budget, assume 1ms chunk size */
+		cd->cpc = CLK_MAX_CPU_HZ / 1000;
+		tr_warn(pipe,
+			"0 CPS requested for module: %#x, core: %d using safe max CPC: %u",
+			current->ipc_config.id, ppl_data->p->core, cd->cpc);
+	}
+
+	int kcps = cd->cpc * 1000 / ppl_data->p->period;
+	tr_info(pipe, "Registering KCPS consumption: %d, core: %d", kcps, ppl_data->p->core);
```
ujfalusi, line 310: It might help to print out the module that is bringing the CPC change

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+	core_kcps_adjust(ppl_data->p->core, kcps);
+
+	int summary_cps = core_kcps_get(ppl_data->p->core);
+	tr_info(pipe, "Sum of KCPS consumption: %d, core: %d", summary_cps, ppl_data->p->core);
+
+	return pipeline_for_each_comp(current, ctx, dir);
```
ujfalusi, line 316: can this fail? if it does are we removing the kcps we just added?

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+	/* complete component init */
+	current->pipeline = ppl_data->p;
+
+	/* modules created throug module adapter have different priv_data */
+	if (current->drv->type != SOF_COMP_MODULE_ADAPTER) {
+		cd = comp_get_drvdata(current);
+	} else {
+		struct processing_module *mod = comp_get_drvdata(current);
+		struct module_data *md = &mod->priv;
+		cd = &md->cfg.base_cfg;
```
ujfalusi, line 298: missing blank line above

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
 	} while (true);
 }
 
+#if CONFIG_KCPS_DYNAMIC_CLOCK_CONTROL
+static int add_pipeline_cps_consumption(struct comp_dev *current,
+					struct comp_buffer *calling_buf,
+					struct pipeline_walk_context *ctx, int dir)
+{
+	struct pipeline_data *ppl_data = ctx->comp_data;
+	struct ipc4_base_module_cfg *cd  = NULL;
```
ujfalusi, line 279: no need to initialize it to NULL

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+
+	return pipeline_for_each_comp(current, ctx, dir);
+}
+
+static int remove_pipeline_cps_consumption(struct comp_dev *current,
+					   struct comp_buffer *calling_buf,
+					   struct pipeline_walk_context *ctx, int dir)
+{
+	struct pipeline_data *ppl_data = ctx->comp_data;
+	struct ipc4_base_module_cfg *cd  = NULL;
```
ujfalusi, line 324: no need to initialize it to NULL

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+	/* complete component init */
+	current->pipeline = ppl_data->p;
+
+	/* modules created throug module adapter have different priv_data */
+	if (current->drv->type != SOF_COMP_MODULE_ADAPTER) {
+		cd = comp_get_drvdata(current);
+	} else {
+		struct processing_module *mod = comp_get_drvdata(current);
+		struct module_data *md = &mod->priv;
+		cd = &md->cfg.base_cfg;
```
ujfalusi, line 343: blank line.

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+	}
+
+	int kcps = cd->cpc * 1000/ppl_data->p->period;
```
ujfalusi, line 346: spaces around `/`

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+	core_kcps_adjust(ppl_data->p->core, -kcps); /* 1000 chunks per second, so cpc value matches kcps? */
+	tr_info(pipe, "Unregistering KCPS consumption: %d, core: %d", kcps, ppl_data->p->core);
```
ujfalusi, line 348: probably mention the component which is dropping the kcps?

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+	int summary_cps = core_kcps_get(ppl_data->p->core);
```
ujfalusi, line 349: it is a bit dense like this, a blank line between 348 and 349?

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+	tr_info(pipe, "Sum of KCPS consumption: %d, core: %d", summary_cps, ppl_data->p->core);
```
ujfalusi, line 350: can this print be moved within `core_kcps_adjust()` with some adjustment?

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
 	case COMP_TRIGGER_PAUSE:
+#if CONFIG_KCPS_DYNAMIC_CLOCK_CONTROL
+		/* dirty - dont add kcps on fallthrough */
+		clocks_handled = true;
+		/* setup walking ctx for adding consumption */
+		data.start = p->source_comp;
+		data.p = p;
+		walk_ctx.comp_func = remove_pipeline_cps_consumption;
+
+		ret = walk_ctx.comp_func(p->source_comp, NULL, &walk_ctx, PPL_DIR_DOWNSTREAM);
```
ujfalusi, line 382: COMPILER_FALLTHROUGH;
?

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+#endif
 	case COMP_TRIGGER_STOP:
 		if (p->status == COMP_STATE_PAUSED || p->xrun_bytes) {
 			/* The task isn't running, trigger inline */
 			ret = pipeline_trigger_run(p, host, cmd);
 			return ret < 0 ? ret : 0;
 		}
-
```
ujfalusi, line 288: spurious change?

src/audio/pipeline/pipeline-stream.c
```diff
@@ -295,6 +396,15 @@ int pipeline_trigger(struct pipeline *p, struct comp_dev *host, int cmd)
 	case COMP_TRIGGER_PRE_RELEASE:
 	case COMP_TRIGGER_PRE_START:
 		/* Add all connected pipelines to the list and trigger them all */
+#if CONFIG_KCPS_DYNAMIC_CLOCK_CONTROL
+		/* setup walking ctx for removing consumption */
+		data.start = p->source_comp;
+		data.p = p;
+		walk_ctx.comp_func = add_pipeline_cps_consumption;
```
ujfalusi, line 403: this whole section can be under the `if(!clocks_handled)` ?

src/audio/pipeline/pipeline-stream.c
```diff
@@ -295,6 +396,15 @@ int pipeline_trigger(struct pipeline *p, struct comp_dev *host, int cmd)
...
+
+		if(!clocks_handled)
+			ret = walk_ctx.comp_func(p->source_comp, NULL, &walk_ctx, PPL_DIR_DOWNSTREAM);
```
ujfalusi, line 406: if fails then should we do something?

src/audio/pipeline/pipeline-stream.c
ujfalusi, line 382: if fails then should we do something?

src/platform/intel/ace/platform.c
```diff
@@ -80,18 +81,22 @@ static struct pm_notifier pm_state_notifier = {
...
 
 	trace_point(TRACE_BOOT_PLATFORM_CLOCK);
 	platform_clock_init(sof);
-
-	/* Set DSP clock to MAX using KCPS API. Value should be lowered when KCPS API
-	 * for modules is implemented
-	 */
-	ret = core_kcps_adjust(cpu_get_id(), CLK_MAX_CPU_HZ / 1000);
+	kcps_budget_init();
+	ret = core_kcps_adjust(cpu_get_id(), BASE_CPS_USAGE);
```
ujfalusi, line 99: The commit message does not match with the change..
If I'm not mistaken we already registered a 'base FW consumption' which happened to be resulting full speed, right?

The commit is more like to change the base FW consumption to a lower, yet to be determined number?

src/platform/intel/ace/platform.c
```diff
@@ -80,18 +81,22 @@ static struct pm_notifier pm_state_notifier = {
...
 	int ret;
 
 	trace_point(TRACE_BOOT_PLATFORM_CLOCK);
 	platform_clock_init(sof);
-
-	/* Set DSP clock to MAX using KCPS API. Value should be lowered when KCPS API
-	 * for modules is implemented
-	 */
-	ret = core_kcps_adjust(cpu_get_id(), CLK_MAX_CPU_HZ / 1000);
+	kcps_budget_init();
```
ujfalusi, line 98: and this should be a separate change and be done earlier, before any of the kcps handling. The lock must be initialized

lgirdwood (commented):
@abonislawski ping - I think this fixes some reported issues.

abonislawski:
> I think this fixes some reported issues.

Unfortunately not, I will post more info in issue thread

src/platform/Kconfig
```diff
@@ -385,6 +385,13 @@ config CAVS_USE_LPRO_IN_WAITI
 	  After waiti exit clock source will be restored.
 	  Choose n if unclear.
 
+config KCPS_DYNAMIC_CLOCK_CONTROL
```
lyakh, line 388: should it depend on IPC4? Because, e.g., it seems to assume that a PAUSE always precedes a STOP, which isn't the case with IPC3?

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+	pipe_dbg(ppl_data->p, "remove_pipeline_cps_consumption(), current->comp.id = %u, dir = %u",
+		 dev_comp_id(current), dir);
+
+	if (!comp_is_single_pipeline(current, ppl_data->start)) {
+		pipe_dbg(ppl_data->p, "remove_pipeline_cps_consumption(), current is from another pipeline");
+		return 0;
+	}
+
+	/* complete component init */
+	current->pipeline = ppl_data->p;
```
lyakh, line 335: why is this needed here? Very strange.

src/platform/Kconfig
abonislawski: Yes, it should

src/platform/Kconfig
abonislawski: Sure, added in "help"

src/platform/intel/ace/platform.c
abonislawski: yeap, moved to separate commit

src/platform/intel/ace/platform.c
abonislawski: Well, for me its fine because actually we are now registering proper **basefw** consumption. Previously it wasnt **basefw** consumption, it was just the highest possible value.
Anyway updated commit a little to note about proper basefw value

src/audio/pipeline/pipeline-stream.c
abonislawski: yeah, thats true. Previously it was intended to fail when 0 cpc received but later it was allowed so we ended in possible negative value. Probably just an int64?

src/audio/pipeline/pipeline-stream.c
abonislawski: EDIT: reverted and moved back to pipeline stream because of never ending compile errors

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -86,3 +86,4 @@ CONFIG_DEBUG_COREDUMP_MEMORY_DUMP_MIN=y
 
 CONFIG_PROBE=y
 CONFIG_PROBE_DMA_MAX=2
+CONFIG_KCPS_DYNAMIC_CLOCK_CONTROL=y
```
lgirdwood, line 89: @abonislawski I'm happy to merge this if this defaults to "n" at the moment until we can resolve all the opens. We also need to add and align all the CPCs on manifest (and that's still WIP, but > 50% done). CI tests for CPC also landing soon, so safe to make default "y" whan all pieces are in place.

btian1:
@abonislawski , could you list here how you get each component cpc data? by below ways?
1. cpc = avg.
2. cpc = avg * 110%
3. cpc = avg * 150%?
4. cpc = (avg+peak) / 2

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,17 +270,128 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+
+	if (cd->cpc == 0) {
+		/* Use maximum clock budget, assume 1ms chunk size */
+		cd->cpc = CLK_MAX_CPU_HZ / 1000;
+		tr_warn(pipe,
+			"0 CPS requested for module: %#x, core: %d using safe max CPC: %u",
+			current->ipc_config.id, ppl_data->p->core, cd->cpc);
+	}
+
+	kcps = cd->cpc * 1000 / ppl_data->p->period;
```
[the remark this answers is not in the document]
btian1: p->trigger.delay = (data.delay_ms * 1000 + p->period - 1) / p->period;
similar code, should checked.

btian1 (commented):
current clock adjust happens on pipeline trigger? how about adjust clock during runtime?

lgirdwood:
> > I think this fixes some reported issues.
> 
> Unfortunately not, I will post more info in issue thread

Good now ?

src/audio/pipeline/pipeline-stream.c
kv2019i: @abonislawski @ujfalusi Is this addressed? Does core_kcps_adjust() handle this case and not overflow?

app/boards/intel_adsp_ace15_mtpm.conf
kv2019i: What's the latest status, do we have any known opens in mainline left? @abonislawski

kv2019i (changes requested):
Let's try to get this in now in good time for v2.8. Only real blocker I see is the kcps overflow -- can you check (see inline)?

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,17 +270,128 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+		struct module_data *md = &mod->priv;
+
+		cd = &md->cfg.base_cfg;
+	}
+
+	kcps = cd->cpc * 1000 / ppl_data->p->period;
+	tr_info(pipe, "Unregistering module: %#x KCPS consumption: %d, core: %d",
+		current->ipc_config.id, kcps, ppl_data->p->core);
+
+	core_kcps_adjust(ppl_data->p->core, -kcps);
```
lyakh, line 356: Wouldn't it be more efficient to collect all kcps adjustments from all pipeline components and adjust the clock only once?

lgirdwood:
@abonislawski should we upstream with ```Kconfig = n``` until we can give full validation cycle. i.e. that would unblock the code merging and would allow more testing without impacting other users.

src/platform/intel/ace/platform.c
tmleman, line 1: Fix commit description. There's nothing left from my changes:

```txt
+Squashed commit:
https://github.com/thesofproject/sof/commit/38b4cb1475fb3f3828cbd7a117ec020ae2f57a12 platform: mtl: lowest clock as default
```

This should be removed.

tmleman (changes requested):

app/boards/intel_adsp_ace15_mtpm.conf
lgirdwood: @abonislawski can you confirm this is now working as expected ? I suppose CI will show us if any CPC numbers are worng, but only for tested modules today.

src/audio/pipeline/pipeline-stream.c
kv2019i: @abonislawski ping?

btian1 (commented):
@lgirdwood @abonislawski , I have a comment on clock adjust:
it should only need be enabled on release build, and keep debug build as always MAX clock.

Debug build no need to perform clock adjust!

Also, Our CI performance and CPC data is based on release build, please take care on this.

lgirdwood:
> @lgirdwood @abonislawski , I have a comment on clock adjust: it should only need be enabled on release build, and keep debug build as always MAX clock.
> 
> Debug build no need to perform clock adjust!
> 
> Also, Our CI performance and CPC data is based on release build, please take care on this.

No, debug build should also use CPC if selected in Kconfig (but we may need a Kconfig debug option to increase teh CPC calculation since we will be printing more logs). i.e. Kconfig CONFIG_DEBUG_CPC_ADJUST which will add X to each CPC to factor in logging costs.

btian1:
> > @lgirdwood @abonislawski , I have a comment on clock adjust: it should only need be enabled on release build, and keep debug build as always MAX clock.
> > Debug build no need to perform clock adjust!
> > Also, Our CI performance and CPC data is based on release build, please take care on this.
> 
> No, debug build should also use CPC if selected in Kconfig (but we may need a Kconfig debug option to increase teh CPC calculation since we will be printing more logs). i.e. Kconfig CONFIG_DEBUG_CPC_ADJUST which will add X to each CPC to factor in logging costs.

debug normally = release * 120%, question is: what's the point to perform clock adjust in debug build?

src/audio/pipeline/pipeline-stream.c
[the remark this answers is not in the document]
abonislawski: I saw this is a fixed value in ipc4_create_pipeline pipe->period = LL_TIMER_PERIOD_US;
anyway what 0 period means for pipeline? Just use module cpc value in such case?

abonislawski:
Im back to this and fixing review comments now

src/audio/pipeline/pipeline-stream.c
abonislawski: It could but this is how it works in reference fw and anyway we will perform maximum 1 clock switch based on current clock configuration in MTL

src/audio/pipeline/pipeline-stream.c
abonislawski: it cant

src/audio/pipeline/pipeline-stream.c
abonislawski: no fail here, any stop conditions means we are done with kcps calculations
(in new approach this walk_ctx func will only calculate kcps, clock change is done outside of this func)

src/audio/pipeline/pipeline-stream.c
```diff
@@ -268,24 +270,123 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+	pipe_dbg(ppl_data->p, "add_pipeline_cps_consumption(), current->comp.id = %u, dir = %u",
+		 dev_comp_id(current), dir);
+
+	if (!comp_is_single_pipeline(current, ppl_data->start)) {
+		pipe_dbg(ppl_data->p, "add_pipeline_cps_consumption(), current is from another pipeline");
+		return 0;
+	}
+
+	/* complete component init */
+	current->pipeline = ppl_data->p;
```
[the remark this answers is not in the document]
abonislawski: removed

src/audio/pipeline/pipeline-stream.c
abonislawski: removed

src/audio/pipeline/pipeline-stream.c
abonislawski: Update: changed to your idea and adjusting clock only once now for one pipeline, this also simplifies things for freq variable overflow issue

src/audio/pipeline/pipeline-stream.c
abonislawski: **fixed this** but actually freq inside cpu-clk-manager could overflow, not kcps_consumption
also changed add_pipeline_cps_consumption flow so core_kcps_adjust will be called only once for one pipeline

src/audio/pipeline/pipeline-stream.c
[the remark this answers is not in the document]
abonislawski: @lgirdwood this is only one unresolved issue after today's update. Im not sure what 0 period could mean?

app/boards/intel_adsp_ace15_mtpm.conf
abonislawski: It should work correctly, already used in a few MTL releases and we have cpc data for most (or all) of our modules.
Updated today with rebase & review fixes

tmleman (approved):

btian1:
Another benefit for remove clock control for debug build: whenever glitch or discontinue issue happened, we can have a try on debug build directly, if no issue, it is CPC caused clock underrun issue, no need to manually disable clock control for check.

src/audio/pipeline/pipeline-stream.c
```diff
@@ -299,17 +301,129 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+	} else {
+		struct processing_module *mod = comp_get_drvdata(current);
+		struct module_data *md = &mod->priv;
+
+		cd = &md->cfg.base_cfg;
+	}
+
+	kcps = cd->cpc * 1000 / ppl_data->p->period;
+	tr_info(pipe, "Unregistering module: %#x KCPS consumption: %d, core: %d",
+		current->ipc_config.id, kcps, ppl_data->p->core);
```
lyakh, line 378: maybe consider making this and a similar print in `add_pipeline_cps_consumption()` above a `tr_dbg()`

src/audio/pipeline/pipeline-stream.c
```diff
@@ -299,17 +301,129 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+		/* setup walking ctx for adding consumption */
+		data.start = p->source_comp;
+		data.p = p;
+		data.kcps = 0;
+		walk_ctx.comp_func = remove_pipeline_cps_consumption;
+
+		ret = walk_ctx.comp_func(p->source_comp, NULL, &walk_ctx, PPL_DIR_DOWNSTREAM);
+
+		core_kcps_adjust(p->core, -data.kcps);
+		tr_info(pipe, "Sum of KCPS consumption: %d, core: %d", core_kcps_get(p->core), p->core);
```
lyakh, line 423: right, I think this single `tr_info()` level message should be enough for most cases. If we notice some illogical values, we'll debug further

btian1:
> My comments have been addressed, so marking this as good to go.
> 
> @btian1 comment w.r.t. how we handle debug build with asserts is an interesting one. There is no question asserts will slow down the code, so the CPC budget should be adjusted. I think one option would be to disable budget based DSP speed adjustements in debug builds (or some new mechanism to add a +10% to budgets if doing a debug build). Both options have their own pro/cons.

The reason we do clock adjust is we want to save some power for customer build, now customer build only use release build, if debug build also need apply clock adjust, it is hard to control the CPC, 10% margin 20% margin? we don't know this exactly, and for some corner cases, it maybe 50%, it may bring extra issues that we can't expect before.

kv2019i:
@btian1 wrote:
> @kv2019i wrote:
>> @btian1 comment w.r.t. how we handle debug build with asserts is an interesting one. There is no question asserts will slow down the code, so the CPC budget should be adjusted. I think one option would be to disable budget based DSP speed adjustements in debug builds (or some new mechanism to add a +10% to budgets if doing a debug build). Both options have their own pro/cons.
> The reason we do clock adjust is we want to save some power for customer build, now customer build only use release build, if debug build also need apply clock adjust, it is hard to control the CPC, 10% margin 20% margin? we don't know this exactly, and for some corner cases, it maybe 50%, it may bring extra issues that we can't expect before.

That leaves then only possibility to disable CPC based adjustments in debug builds. This is not ideal as the delta between debug and release builds grow -- we want to use debug builds in CI as e.g. asserts are very useful to find programming mistakes in submitted PRs. Not sure if we have capacity to test both release and debug builds (FYI @keqiaozhang @mengdonglin @marc-hb @fredoh9 and @lyakh @ujfalusi with whom this was discussed recently ).

kv2019i:
@abonislawski I do think there is something else wrong still with this PR. We have a lot of fails in the MTL tests. The logs are not published to 01.org now, but errors look like:

````
[  202.698774] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx      : 0x13000003|0x1: GLB_SET_PIPELINE_STATE [data size: 12]
[  203.201276] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc timed out for 0x13000003|0x1
```` 



FW side:
```
[   10.146323] <inf> ipc: ipc_cmd: rx	: 0x13000003|0x1
[   10.146340] <inf> pipe: pipeline_trigger: pipe:0 0x0 pipe trigger cmd 2
[   10.146346] <inf> pipe: remove_pipeline_cps_consumption: Unregistering module: 0x4 KCPS consumption: 7116, core: 0
[   10.146351] <inf> pipe: remove_pipeline_cps_consumption: Unregistering module: 0x10 KCPS consumption: 20663, core: 0
[   10.146358] <inf> pipe: remove_pipeline_cps_consumption: Unregistering module: 0x10004 KCPS consumption: 7116, core: 0
[   10.146381] <inf> clock: clock_set_freq: clock 0 set freq 23024000Hz freq_idx 0
[   10.146426] <inf> clock: clock_set_freq: clock 1 set freq 23024000Hz freq_idx 0
[   10.146510] <inf> clock: clock_set_freq: clock 2 set freq 23024000Hz freq_idx 0
[   10.373946] <inf> ll_schedule: zephyr_domain_thread_fn: ll core 0 timer avg 9796, max 42745, overruns 39```
```

So clocks are reduced quite a lot and this leads to serious amount of overruns (39 above -> during one 1sec window).

One option to try is to disable the CPC based adjustments in app/debug_overlay.conf . This at least tells whether the problem is CPC budgets not matching actual performance when debug is enabled (and it is enabled in PR tests!) or is this caused by something else.

**Comment by marc-hb** [comment]
> if debug build also need apply clock adjust, it is hard to control the CPC, 10% margin 20% margin? we don't know this exactly, and for some corner cases, it maybe 50%, it may bring extra issues that we can't expect before.

Good point but I think running the _same code paths_ by default is much more important. The ONLY difference between debug and release builds should be asserts, logging, timings and performance.

If 50% is not enough then there is too much logging or you just found a performance bug in logging. Both are good to find.

(Of course you should be able to easily turn off clock control when debugging interactively; this discussion is only about the **defaults** used in validation) 

You could argue that the dynamic clock control code makes only timing differences and does not affect the rest of the code more than logging. I disagree for two reasons:
- Clock control is much more complex than logging hence much more likely to have bugs that affect the rest of the code
- Clock control code needs to be TESTED in debug builds too!

Signing is a similar problem: every "developer mode" in every product I have ever seen is implemented using a special insecure keys. All of them keep running the SAME CODE in developer mode. Because you do not want to wait release time to start exercising signing and authentication code; that's too late (and even when running the same signing code we STILL find last minute bugs like #8281 and https://github.com/thesofproject/rimage/pull/123)

**Comment by marc-hb** [comment]
> @abonislawski should we upstream with Kconfig = n until we can give full validation cycle. i.e. that would unblock the code merging and would allow more testing without impacting other users.

For the usual _continuous_ integration reasons, @abonislawski please turn the feature OFF at the last minute once you get a mostly clean run, all review comments addressed and the PR approved.

After a few days submit another, one-line PR that just turns the feature ON by default.

These two separate steps will give us the necessary perspective in daily test results (which stress the code much more than PR tests)

**Comment by btian1** [comment]
> @btian1 wrote:
> 
> > @kv2019i wrote:
> > > @btian1 comment w.r.t. how we handle debug build with asserts is an interesting one. There is no question asserts will slow down the code, so the CPC budget should be adjusted. I think one option would be to disable budget based DSP speed adjustements in debug builds (or some new mechanism to add a +10% to budgets if doing a debug build). Both options have their own pro/cons.
> > > The reason we do clock adjust is we want to save some power for customer build, now customer build only use release build, if debug build also need apply clock adjust, it is hard to control the CPC, 10% margin 20% margin? we don't know this exactly, and for some corner cases, it maybe 50%, it may bring extra issues that we can't expect before.
> 
> That leaves then only possibility to disable CPC based adjustments in debug builds. This is not ideal as the delta between debug and release builds grow -- we want to use debug builds in CI as e.g. asserts are very useful to find programming mistakes in submitted PRs. Not sure if we have capacity to test both release and debug builds (FYI @keqiaozhang @mengdonglin @marc-hb @fredoh9 and @lyakh @ujfalusi with whom this was discussed recently ).

yes, this is a middle way, @abonislawski , could you add a patch to disable clock control in performance build?
Add
CONFIG_KCPS_DYNAMIC_CLOCK_CONTROL=n
to app/perf_overlay.conf

**Comment by lgirdwood** [comment]
> > @abonislawski should we upstream with Kconfig = n until we can give full validation cycle. i.e. that would unblock the code merging and would allow more testing without impacting other users.
> 
> For the usual _continuous_ integration reasons, @abonislawski please turn the feature OFF at the last minute once you get a mostly clean run, all review comments addressed and the PR approved.
> 
> After a few days submit another, one-line PR that just turns the feature ON by default.
> 
> These two separate steps will give us the necessary perspective in daily test results (which stress the code much more than PR tests)

Ack - this will help mimize impact or help bisect.

**Review by marcinszkudlinski — CHANGES_REQUESTED** [review]
This PR won't work for DP
In typical DP a pipeline is on core 0 and DP modules - that belong to the pipeline - are located on core 1 or 2. 
MCPS for a pipeline should be calculated separately for each core

**Comment by lgirdwood** [comment]
@marcinszkudlinski btw, any thoughts on measuring DP workloads in terms of MCPS ? It seems like we need to add timer logic to schedule() and yield() in Zephyr.

**Comment by lgirdwood** [comment]
@abonislawski will the latest update align with @marcinszkudlinski DP work ?

**Comment by abonislawski** [comment]
@lgirdwood aligned with DP work in the latest push

File: src/audio/pipeline/pipeline-stream.c [hunk 33]
```diff
@@ -311,17 +313,135 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
 	} while (true);
 }
 
+#if CONFIG_KCPS_DYNAMIC_CLOCK_CONTROL
+static int add_pipeline_cps_consumption(struct comp_dev *current,
```
lyakh, line 317: I'd call it `pipeline_add_cps_consumption()`

src/audio/pipeline/pipeline-stream.c
```diff
@@ -311,17 +313,135 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+		tr_dbg(pipe, "Registering module: %#x KCPS consumption: %d, core: %d",
+		       current->ipc_config.id, kcps, comp_core);
+
+		ppl_data->kcps[comp_core] += kcps;
+	}
+
+	return pipeline_for_each_comp(current, ctx, dir);
+}
+
+static int remove_pipeline_cps_consumption(struct comp_dev *current,
```
lyakh, line 363: same: would be good to keep using the `pipeline_*` namespace

src/audio/pipeline/pipeline-stream.c
```diff
@@ -311,17 +313,135 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
 	switch (cmd) {
 	case COMP_TRIGGER_PAUSE:
+#if CONFIG_KCPS_DYNAMIC_CLOCK_CONTROL
+		/* dirty - dont add kcps on fallthrough */
+		clocks_handled = true;
+		/* setup walking ctx for adding consumption */
+		data.start = p->source_comp;
+		data.p = p;
+		memset(data.kcps, 0, CONFIG_CORE_COUNT * sizeof(data.kcps[0]));
+		walk_ctx.comp_func = remove_pipeline_cps_consumption;
```
lyakh, line 432: it is already set to that function. You can remove line 415 if you want to keep these assignments symmetric

src/audio/pipeline/pipeline-stream.c
```diff
@@ -311,17 +313,135 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
 
 	switch (cmd) {
 	case COMP_TRIGGER_PAUSE:
+#if CONFIG_KCPS_DYNAMIC_CLOCK_CONTROL
+		/* dirty - dont add kcps on fallthrough */
+		clocks_handled = true;
+		/* setup walking ctx for adding consumption */
+		data.start = p->source_comp;
+		data.p = p;
+		memset(data.kcps, 0, CONFIG_CORE_COUNT * sizeof(data.kcps[0]));
```
lyakh, line 431: do
```
struct pipeline_data data = {
    .start = p->source_comp,
    .p = p,
};
```
and remove these 3 lines here and also in the `RELEASE` case below

app/perf_overlay.conf
```diff
@@ -1,3 +1,4 @@
 CONFIG_PERFORMANCE_COUNTERS=y
 CONFIG_SYS_HEAP_RUNTIME_STATS=y
-CONFIG_TIMING_FUNCTIONS=y
\ No newline at end of file
+CONFIG_TIMING_FUNCTIONS=y
+CONFIG_KCPS_DYNAMIC_CLOCK_CONTROL=n
```
lyakh, line 4: please add a new line at the end

src/audio/pipeline/pipeline-stream.c
lgirdwood: ack, lets keep with the ```pipeline_``` name space. I think we now have most base APIs aligning to this but probably some that still need to be updated.

kv2019i:
Build is failing:
```
/srv/home/jenkins/workspace/sof_config_build@2/sof/src/audio/pipeline/pipeline-stream.c:415:17: error: use of undeclared identifier 'data'
```
https://sof-ci.01.org/sofpr/PR8019/build1715/build/firmware_community/mtl.log

lgirdwood:
@marcinszkudlinski can you re-review and approve if good now for you . Thanks !

marc-hb:
Build failures in https://github.com/thesofproject/sof/actions/runs/7667629748/job/20897861756?pr=8019 and https://github.com/thesofproject/sof/actions/runs/7667629757/job/20897863643?pr=8019


I suspect you are just missing some `#ifdef`.


The errors are difficult to find until .github fixes #8807 and #8799 are merged: please help review.

marc-hb:
MTL hardware tests are almost all failed in https://sof-ci.01.org/sofpr/PR8019/build2360/devicetest/index.html

While test results have not been great recently. I've never seen so much red recently so it is most likely caused by this PR (currently at 989bda8bcbaa9de66c60ea82be8ceabf1b1bbe03)

src/audio/pipeline/pipeline-stream.c
```diff
@@ -311,17 +313,136 @@ static void pipeline_trigger_xrun(struct pipeline *p, struct comp_dev **host)
...
+		/* Use maximum clock budget, assume 1ms chunk size */
+		ppl_data->kcps[comp_core] = CLK_MAX_CPU_HZ / 1000;
+		tr_warn(pipe,
+			"0 CPS requested for module: %#x, core: %d using safe max KCPS: %u",
+			current->ipc_config.id, comp_core, ppl_data->kcps[comp_core]);
+
+		return PPL_STATUS_PATH_STOP;
+	} else {
+		kcps = cd->cpc * 1000 / current->period;
+		//tr_dbg(pipe, "Registering module: %#x KCPS consumption: %d, core: %d",
```
marcinszkudlinski, line 355: style

marcinszkudlinski (approved):

marc-hb:
All MTL tests still fail: https://sof-ci.01.org/sofpr/PR8019/build2392/devicetest/index.html

There also a lot of build failures:
https://github.com/thesofproject/sof/actions/runs/7693205483/job/20961497200?pr=8019
https://github.com/thesofproject/sof/actions/runs/7693205482/job/20961494915?pr=8019
https://github.com/thesofproject/sof/actions/runs/7693205487/job/20961494936?pr=8019

lgirdwood:
@abonislawski re the CI test failures, are we missing anything on the kernel side. i.e. does kernel need to do any IP programming flow around DSP clock change ?
Fwiw, I know we have some loops that spin around a clock register bit change, could some be timing out or spinning forever ?
```
[  407.529254] kernel: snd_sof:sof_pcm_trigger: sof-audio-pci-intel-mtl 0000:00:1f.3: pcm: trigger stream 6 dir 1 cmd 0
[  407.529276] kernel: snd_sof:sof_ipc4_trigger_pipelines: sof-audio-pci-intel-mtl 0000:00:1f.3: trigger cmd: 0 state: 3
[  407.529297] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx      : 0x13000003|0x1: GLB_SET_PIPELINE_STATE [data size: 12]
[  408.031809] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc timed out for 0x13000003|0x1
[  408.031838] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: Attempting to prevent DSP from entering D3 state to preserve context
[  408.031847] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ IPC dump start ]------------
[  408.031879] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: Host IPC initiator: 0x93000003|0x1|0x0, target: 0x1b060000|0x0|0x0, ctl: 0x3
[  408.031887] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ IPC dump end ]------------
[  408.031894] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ DSP dump start ]------------
[  408.031899] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: IPC timeout
[  408.031904] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: fw_state: SOF_FW_BOOT_COMPLETE (7)
[  408.031930] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: 0x50000005: module: ROM_EXT, state: FW_ENTERED, running
[  408.031942] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: Firmware state: 0x0, status/error code: 0x0
[  408.031970] kernel: snd_sof:sof_ipc4_find_debug_slot_offset_by_type: sof-audio-pci-intel-mtl 0000:00:1f.3: Slot type 0x4c455400 is not available in debug window
[  408.031975] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ DSP dump end ]------------
[  408.032005] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: failed to pause all pipelines
[  408.032017] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ASoC: error at soc_component_trigger on 0000:00:1f.3: -110
[  408.032032] kernel:  DMIC Raw: ASoC: trigger FE cmd: 0 failed: -110
```

abonislawski:
@lgirdwood  We shouldnt need anything from kernel side as this is already used on MTL branches.
It seems that everything crashes on debug build (default for PR testing in linux CI) and works on release build (default for PR testing on second internal CI), I need to investigate why

> Fwiw, I know we have some loops that spin around a clock register bit change, could some be timing out or spinning forever ?
I will check that part

kv2019i:
Still failure here on MTL nocodec https://sof-ci.01.org/sofpr/PR8019/build3645/devicetest/index.html?model=MTLP_RVP_NOCODEC&testcase=check-capture-10sec

marc-hb:
This compilation failure seems related:

```
/zep_workspace/sof/src/lib/cpu-clk-manager.c: In function 'core_kcps_adjust':
/zep_workspace/sof/src/lib/cpu-clk-manager.c:67:69: error: 'CLK_MAX_CPU_HZ' undeclared (first use in this function)
   67 |                 ret = request_freq_change(core_id, MIN(freq * 1000, CLK_MAX_CPU_HZ));
      |                                                                     ^~~~~~~~~~~~~~
```

kv2019i (changes requested):
Let me mark with a -1 so this is not accidentally merged. Code looks good, but we still have CI tests failing.

kv2019i (approved):
Tests seem better now, removing my -1

abonislawski:
Internal and Linux CI pass

I have also merged two ppl consumption functions into single pipeline_calc_cps_consumption
divided base cps for primary and secondary, by default pretty high values but we can optimize this later if required

```
#define PRIMARY_CORE_BASE_CPS_USAGE 20000
#define SECONDARY_CORE_BASE_CPS_USAGE 10000
```

lgirdwood (approved):

Issue #8006, opened 2023-08-03
[RFC] Direct use of cached / not-cached shared memory alias instead of coherent.h in case of buffers


marcinszkudlinski:
As everybody knows, not coherent cache is a serious complication we must deal with. 

There are 2 ways of share data between cores:
1) Using non-cached alias - read/write directly from L2 RAM 
2) Using a cached alias and a mechanism of writeback/invalidation to ensure coherent data between cores

Currently Zephyr uses solution 1, SOF uses solution 2

Shared memory in general has a lot of advantages - it is easy to use, the code is "obvious correct" instead of "correct". The structures may be used by several cores at the same time (standard multithread rules for sharing data apply of course) 

The question is - **how much it is slower?** 

I made some measurements on MTL RVP - access to the data using several methods:
1) not shared data. 2 ways were measured:
 - cached memory ensuring cache hit 
 - using coherent.h in not-shared mode

2) shared data - 3 ways were measured:
- not cached memory alias
- using coherent.h in shared mode
- manual invalidation before operations / writeback after operations

all results in CPU cycles

NOT SHARED MEMORY

|ID|desc|Cached (hit)|coherent.h (not shared)|
|--|--|--|--|
|1|read single dword|30|45|
|2|read double dword|31|44|
|3|read cacheline (16 dword)|80|98|
|4|read 64dwords|226|257|
|5|read 1024dword|3132|3449|
|6|read twice 1024|6234|6852|
|7|write single dword|30|44|
|8|write double dword|31|43|
|9|write cacheline (16dword)|84|43|
|10|write 64 dwords|195|211|
|11|wrtite 1024 dwords|2632|2748|
|12|twice write 1024 dwords|5535|5749|

no surprise here - using of direct access to memory is faster

SHARED MEMORY

|ID|desc|uncahced|cached (with inv/wb)|coherent.h (shared)|
|--|--|--|--|--|
|1|read single dword|35|38|119|
|2|read double dword|35|39|119|
|3|read cacheline (16 dword)|115|87|175|
|4|read 64dwords|344|234|336|
|5|read 1024dword|5034|3177|3575|
|6|read twice 1024|10036|6279|6972|
|7|write single dword|32|37|121|
|8|write double dword|32|40|119|
|9|write cacheline (16dword)|76|81|172|
|10|write 64 dwords|200|208|302|
|11|wrtite 1024 dwords|2636|2749|2955|

[...]

PR #8020, opened 2023-08-09
pipeline 2.0: Introducing Data Processing Queue

marcinszkudlinski:
re-inroducing DP queue. This PR contains DpQueue code only

dependencies:
- [x] github.com/thesofproject/sof/pull/8002
- [x] https://github.com/thesofproject/sof/pull/8167
- [x] PR https://github.com/thesofproject/sof/pull/8166

some sparse warning are reported now till RFC changes are implemented

DP queue is a lockless circular buffer
providing safe consumer/producer cached operations cross cores

prerequisites:
 1) incoming and outgoing data rate MUST be the same
 2) Both data consumer and data producer declare
    max chunk sizes they want to use (IBS/OBS)

required Buffer size:
	- 2*MAX(IBS,OBS) if the larger of IBS/OBS
          is multiplication of smaller
	- 3*MAX(IBS,OBS) otherwise

The queue may work in 2 modes
1) local mode
   in case both receiver and sender are located on
   the same core and cache coherency
   does not matter. dp_queue structure is located in cached memory
   In this case DP Queue is a simple ring buffer

2) shared mode
   In this case we need to writeback cache when new data
   arrive and invalidate cache on secondary core.
   dp_queue structure is located in shared memory

dpQueue is a lockless consumer/producer safe buffer.
It is achieved by having only 2 shared variables:

 write_offset - can be modified by data producer only
 read_offset - can be modified by data consumer only

 as 32 bit operations are atomic, it is multi-thread and multi-core save

There some explanation needed how free_space and
available_data are calculated

number of avail data in circular buffer may be calculated as:
	data_avail = write_offset - read_offset
  and check for wrap around
	if (data_avail < 0) data_avail = buffer_size + data_avail

The problem is when write_offset == read_offset,
!!! it may mean either that the buffer is empty
    or the buffer is completely filled !!!

To solve the above issue having only 2 variables mentioned before:
 - allow both offsets to point from 0 to DOUBLE buffer_size
 - when calculating pointers to data, use:
         data_bufer[offset % buffer_size]
 - use double buffer size in wrap around check when calculating
   available data

And now:
  - write_offset == read_offset
		always means "buffer empty"
  - write_offset == read_offset + buffer_size
		always means "buffer full"

  - data_avail = write_offset - read_offset
	if (data_avail < 0) data_avail = 2 * buffer_size + data_avail

### DIFF
--- a/src/audio/audio_stream.c
+++ b/src/audio/audio_stream.c
@@ -49,7 +49,8 @@ static size_t audio_stream_get_data_available(struct sof_source *source)
 }
 
 static int audio_stream_get_data(struct sof_source *source, size_t req_size,
-				 void  **data_ptr, void **buffer_start, size_t *buffer_size)
+				 void const **data_ptr, void const **buffer_start,
+				 size_t *buffer_size)
 {
 	struct audio_stream *audio_stream = container_of(source, struct audio_stream, source_api);
 	struct comp_buffer *buffer = container_of(audio_stream, struct comp_buffer, stream);
--- a/src/audio/dp_queue.c
+++ b/src/audio/dp_queue.c
@@ -1,0 +1,300 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// Copyright(c) 2023 Intel Corporation. All rights reserved.
+//
+
+#include <sof/common.h>
+#include <sof/trace/trace.h>
+#include <sof/lib/uuid.h>
+
+#include <sof/audio/dp_queue.h>
+
+#include <rtos/alloc.h>
+#include <ipc/topology.h>
+
+LOG_MODULE_REGISTER(dp_queue, CONFIG_SOF_LOG_LEVEL);
+
+/* 393608d8-4188-11ee-be56-0242ac120002 */
+DECLARE_SOF_RT_UUID("dp_queue", dp_queue_uuid, 0x393608d8, 0x4188, 0x11ee,
+		    0xbe, 0x56, 0x02, 0x42, 0xac, 0x12, 0x20, 0x02);
+DECLARE_TR_CTX(dp_queue_tr, SOF_UUID(dp_queue_uuid), LOG_LEVEL_INFO);
+
+static inline uint8_t __sparse_cache *dp_queue_buffer_end(struct dp_queue *dp_queue)
+{
+	return dp_queue->_data_buffer + dp_queue->data_buffer_size;
+}
+
+static inline struct dp_queue *dp_queue_from_sink(struct sof_sink *sink)
+{
+	return container_of(sink, struct dp_queue, _sink_api);
+}
+
+static inline struct dp_queue *dp_queue_from_source(struct sof_source *source)
+{
+	return container_of(source, struct dp_queue, _source_api);
+}
+
+static inline void dp_queue_invalidate_shared(struct dp_queue *dp_queue,
+					      void __sparse_cache *ptr, size_t size)
+{
+	/* no cache required in case of not shared queue */
+	if (!dp_queue_is_shared(dp_queue))
+		return;
+
+	/* wrap-around? */
+	if ((uintptr_t)ptr + size > (uintptr_t)dp_queue_buffer_end(dp_queue)) {
+		/* writeback till the end of circular buffer */
+		dcache_invalidate_region
+			(ptr, (uintptr_t)dp_queue_buffer_end(dp_queue) - (uintptr_t)ptr);
+		size -= (uintptr_t)dp_queue_buffer_end(dp_queue) - (uintptr_t)ptr;
+		ptr = dp_queue->_data_buffer;
+	}
+	/* invalidate rest of data */
+	dcache_invalidate_region(ptr, size);
+}
+
+static inline void dp_queue_writeback_shared(struct dp_queue *dp_queue,
+					     void __sparse_cache *ptr, size_t size)
+{
+	/* no cache required in case of not shared queue */
+	if (!dp_queue_is_shared(dp_queue))
+		return;
+
+	/* wrap-around? */
+	if ((uintptr_t)ptr + size > (uintptr_t)dp_queue_buffer_end(dp_queue)) {
+		/* writeback till the end of circular buffer */
+		dcache_writeback_region
+			(ptr, (uintptr_t)dp_queue_buffer_end(dp_queue) - (uintptr_t)ptr);
+		size -= (uintptr_t)dp_queue_buffer_end(dp_queue) - (uintptr_t)ptr;
+		ptr = dp_queue->_data_buffer;
+	}
+	/* writeback rest of data */
+	dcache_writeback_region(ptr, size);
+}
+
+static inline
+uint8_t __sparse_cache *dp_queue_get_pointer(struct dp_queue *dp_queue, uint32_t offset)
+{
+	/* check if offset is not in "double area"
+	 * lines below do a quicker version of offset %= dp_queue->data_buffer_size;
+	 */
+	if (offset >= dp_queue->data_buffer_size)
+		offset -= dp_queue->data_buffer_size;
+	return dp_queue->_data_buffer + offset;
+}
+
+static inline
+uint32_t dp_queue_inc_offset(struct dp_queue *dp_queue, uint32_t offset, uint32_t inc)
+{
+	assert(inc <= dp_queue->data_buffer_size);
+	offset += inc;
+	/* wrap around ? 2*size because of "double area" */
+	if (offset >= 2 * dp_queue->data_buffer_size)
+		offset -= 2 * dp_queue->data_buffer_size;
+	return offset;
+}
+
+static inline
+size_t _dp_queue_get_data_available(struct dp_queue *dp_queue)
+{
+	int32_t avail_data =  dp_queue->_write_offset - dp_queue->_read_offset;
+	/* wrap around ? 2*size because of "double area" */
+	if (avail_data < 0)
+		avail_data = 2 * dp_queue->data_buffer_size + avail_data;
+
+	return avail_data;
+}
+
+static size_t dp_queue_get_data_available(struct sof_source *source)
+{
+	struct dp_queue *dp_queue = dp_queue_from_source(source);
+
+	CORE_CHECK_STRUCT(dp_queue);
+	return _dp_queue_get_data_available(dp_queue);
+}
+
+static size_t dp_queue_get_free_size(struct sof_sink *sink)
+{
+	struct dp_queue *dp_queue = dp_queue_from_sink(sink);
+
+	CORE_CHECK_STRUCT(dp_queue);
+	return dp_queue->data_buffer_size - _dp_queue_get_data_available(dp_queue);
+}
+
+static int dp_queue_get_buffer(struct sof_sink *sink, size_t req_size,
+			       void **data_ptr, void **buffer_start, size_t *buffer_size)
+{
+	struct dp_queue *dp_queue = dp_queue_from_sink(sink);
+
+	CORE_CHECK_STRUCT(dp_queue);
+	if (req_size > dp_queue_get_free_size(sink))
+		return -ENODATA;
+
+	/* note, __sparse_force is to be removed once sink/src use __sparse_cache for data ptrs */
+	*data_ptr = (__sparse_force void *)dp_queue_get_pointer(dp_queue, dp_queue->_write_offset);
+	*buffer_start = (__sparse_force void *)dp_queue->_data_buffer;
+	*buffer_size = dp_queue->data_buffer_size;
+
+	/* no need to invalidate cache - buffer is to be written only */
+	return 0;
+}
+
+static int dp_queue_commit_buffer(struct sof_sink *sink, size_t commit_size)
+{
+	struct dp_queue *dp_queue = dp_queue_from_sink(sink);
+
+	CORE_CHECK_STRUCT(dp_queue);
+	if (commit_size) {
+		dp_queue_writeback_shared(dp_queue,
+					  dp_queue_get_pointer(dp_queue, dp_queue->_write_offset),
+					  commit_size);
+
+		/* move write pointer */
+		dp_queue->_write_offset =
+				dp_queue_inc_offset(dp_queue, dp_queue->_write_offset, commit_size);
+	}
+
+	return 0;
+}
+
+static int dp_queue_get_data(struct sof_source *source, size_t req_size,
+			     void const **data_ptr, void const **buffer_start, size_t *buffer_size)
+{
+	struct dp_queue *dp_queue = dp_queue_from_source(source);
+	__sparse_cache void *_data_ptr;
+
+	CORE_CHECK_STRUCT(dp_queue);
+	if (req_size > dp_queue_get_data_available(source))
+		return -ENODATA;
+
+	_data_ptr = dp_queue_get_pointer(dp_queue, dp_queue->_read_offset);
+
+	/* clean cache in provided data range */
+	dp_queue_invalidate_shared(dp_queue, _data_ptr, req_size);
+
+	*buffer_start = (__sparse_force void *)dp_queue->_data_buffer;
+	*buffer_size = dp_queue->data_buffer_size;
+	*data_ptr = (__sparse_force void *)_data_ptr;
+
+	return 0;
+}
+
+static int dp_queue_release_data(struct sof_source *source, size_t free_size)
+{
+	struct dp_queue *dp_queue = dp_queue_from_source(source);
+
+	CORE_CHECK_STRUCT(dp_queue);
+	if (free_size) {
+		/* data consumed, free buffer space, no need for any special cache operations */
+		dp_queue->_read_offset =
+				dp_queue_inc_offset(dp_queue, dp_queue->_read_offset, free_size);
+	}
+
+	return 0;
+}
+
+static int dp_queue_set_ipc_params(struct dp_queue *dp_queue,
+				   struct sof_ipc_stream_params *params,
+				   bool force_update)
+{
+	CORE_CHECK_STRUCT(dp_queue);
+	if (dp_queue->_hw_params_configured && !force_update)
+		return 0;
+
+	dp_queue->audio_stream_params.frame_fmt = params->frame_fmt;
+	dp_queue->audio_stream_params.rate = params->rate;
+	dp_queue->audio_stream_params.channels = params->channels;
+	dp_queue->audio_stream_params.buffer_fmt = params->buffer_fmt;
+
+	dp_queue->_hw_params_configured = true;
+
+	return 0;
+}
+
+static int dp_queue_set_ipc_params_source(struct sof_source *source,
+					  struct sof_ipc_stream_params *params,
+					  bool force_update)
+{
+	struct dp_queue *dp_queue = dp_queue_from_source(source);
+
+	CORE_CHECK_STRUCT(dp_queue);
+	return dp_queue_set_ipc_params(dp_queue, params, force_update);
+}
+
+static int dp_queue_set_ipc_params_sink(struct sof_sink *sink,
+					struct sof_ipc_stream_params *params,
+					bool force_update)
+{
+	struct dp_queue *dp_queue = dp_queue_from_sink(sink);
+
+	CORE_CHECK_STRUCT(dp_queue);
+	return dp_queue_set_ipc_params(dp_queue, params, force_update);
+}
+
+static const struct source_ops dp_queue_source_ops = {
+	.get_data_available = dp_queue_get_data_available,
+	.get_data = dp_queue_get_data,
+	.release_data = dp_queue_release_data,
+	.audio_set_ipc_params = dp_queue_set_ipc_params_source,
+};
+
+static const struct sink_ops dp_queue_sink_ops = {
+	.get_free_size = dp_queue_get_free_size,
+	.get_buffer = dp_queue_get_buffer,
+	.commit_buffer = dp_queue_commit_buffer,
+	.audio_set_ipc_params = dp_queue_set_ipc_params_sink,
+};
+
+struct dp_queue *dp_queue_create(size_t ibs, size_t obs, uint32_t flags)
+{
+	struct dp_queue *dp_queue;
+
+	/* allocate DP structure */
+	if (flags & DP_QUEUE_MODE_SHARED)
+		dp_queue = rzalloc(SOF_MEM_ZONE_RUNTIME_SHARED, 0, SOF_MEM_CAPS_RAM,
+				   sizeof(*dp_queue));
+	else
+		dp_queue = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*dp_queue));
+	if (!dp_queue)
+		return NULL;
+
+	dp_queue->_flags = flags;
+
+	CORE_CHECK_STRUCT_INIT(dp_queue, flags & DP_QUEUE_MODE_SHARED);
+
+	/* initiate sink/source */
+	source_init(dp_queue_get_source(dp_queue), &dp_queue_source_ops,
+		    &dp_queue->audio_stream_params);
+	sink_init(dp_queue_get_sink(dp_queue), &dp_queue_sink_ops,
+		  &dp_queue->audio_stream_params);
+
+	/* set obs/ibs in sink/source interfaces */
+	sink_set_obs(&dp_queue->_sink_api, obs);
+	source_set_ibs(&dp_queue->_source_api, ibs);
+
+	uint32_t max_ibs_obs = MAX(ibs, obs);
+	uint32_t min_ibs_obs = MIN(ibs, obs);
+
+	/* calculate required buffer size */
+	if (max_ibs_obs % min_ibs_obs == 0)
+		dp_queue->data_buffer_size = 2 * max_ibs_obs;
+	else
+		dp_queue->data_buffer_size = 3 * max_ibs_obs;
+
+	/* allocate data buffer - always in cached memory alias */
+	dp_queue->data_buffer_size = ALIGN_UP(dp_queue->data_buffer_size, PLATFORM_DCACHE_ALIGN);
+	dp_queue->_data_buffer = (__sparse_force __sparse_cache void *)
+			rballoc_align(0, 0, dp_queue->data_buffer_size, PLATFORM_DCACHE_ALIGN);
+	if (!dp_queue->_data_buffer)
+		goto err;
+
+	tr_info(&dp_queue_tr, "DpQueue created, shared: %u ibs: %u obs %u, size %u",
+		dp_queue_is_shared(dp_queue), ibs, obs, dp_queue->data_buffer_size);
+
+	/* return a pointer to allocated structure */
+	return dp_queue;
+err:
+	tr_err(&dp_queue_tr, "DpQueue creation failure");
+	rfree(dp_queue);
+	return NULL;
+}
--- a/src/include/sof/audio/dp_queue.h
+++ b/src/include/sof/audio/dp_queue.h
@@ -1,0 +1,169 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright(c) 2023 Intel Corporation. All rights reserved.
+ *
+ */
+
+#ifndef __SOF_DP_QUEUE_H__
+#define __SOF_DP_QUEUE_H__
+
+#include <sof/audio/sink_api.h>
+#include <sof/audio/source_api.h>
+#include <sof/audio/sink_api_implementation.h>
+#include <sof/audio/source_api_implementation.h>
+#include <sof/audio/audio_stream.h>
+#include <rtos/bit.h>
+#include <sof/common.h>
+#include <ipc/topology.h>
+#include <sof/coherent.h>
+
+/**
+ * DP queue is a lockless circular buffer
+ * providing safe consumer/producer cached operations cross cores
+ *
+ * prerequisites:
+ *  1) incoming and outgoing data rate MUST be the same
+ *  2) Both data consumer and data producer declare max chunk sizes they want to use (IBS/OBS)
+ *
+ * required Buffer size:
+ *	- 2*MAX(IBS,OBS) if the larger of IBS/OBS is multiplication of smaller
+ *	- 3*MAX(IBS,OBS) otherwise
+ *
+ * The queue may work in 2 modes
+ * 1) local mode
+ *    in case both receiver and sender are located on the same core and cache coherency
+ *    does not matter. dp_queue structure is located in cached memory
+ *    In this case DP Queue is a simple ring buffer
+ *
+ * 2) shared mode
+ *    In this case we need to writeback cache when new data arrive and invalidate cache on
+ *    secondary core. dp_queue structure is located in shared memory
+ *
+ *
+ * dpQueue is a lockless consumer/producer safe buffer. It is achieved by having only 2 shared
+ * variables:
+ *  _write_offset - can be modified by data producer only
+ *  _read_offset - can be modified by data consumer only
+ *
+ *  as 32 bit operations are atomic, it is multi-thread and multi-core save
+ *
+ * There some explanation needed how free_space and available_data are calculated
+ *
+ * number of avail data in circular buffer may be calculated as:
+ *	data_avail = _write_offset - _read_offset
+ *   and check for wrap around
+ *	if (data_avail < 0) data_avail = buffer_size - data_avail
+ *
+ * The problem is when _write_offset == _read_offset,
+ * !!! it may mean either that the buffer is empty or the buffer is completely filled !!!
+ *
+ * To solve the above issue having only 2 variables mentioned before:
+ *  - allow both offsets to point from 0 to DOUBLE buffer_size
+ *  - when calculating pointers to data, use: data_bufer[offset % buffer_size]
+ *  - use double buffer size in wrap around check when calculating available data
+ *
+ * And now:
+ *   - _write_offset == _read_offset
+ *		always means "buffer empty"
+ *   - _write_offset == _read_offset + buffer_size
+ *		always means "buffer full"
+ */
+
+struct dp_queue;
+struct sof_audio_stream_params;
+
+/* DP flags */
+#define DP_QUEUE_MODE_LOCAL 0
+#define DP_QUEUE_MODE_SHARED BIT(1)
+
+/* the dpQueue structure */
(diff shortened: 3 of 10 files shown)


src/audio/dp_queue.c
```diff
@@ -0,0 +1,367 @@
...
+
+struct dp_queue *dp_queue_get_from_sink(struct sof_sink *sink)
+{
+	return container_of(sink, struct dp_queue, sink_api);
+}
+
+struct dp_queue *dp_queue_get_from_source(struct sof_source *source)
+{
+	return container_of(source, struct dp_queue, source_api);
+}
```
lgirdwood, line 59: These need to be static inline in the header - I dont think we can say API partitioning (i.e. what is public/private) is higher priority than MCPS.

lgirdwood (commented):
Overall I think this looks good, but we do need to prioritze MCPS over partitioning of code between private and public API calls.

src/audio/dp_queue.c
```diff
@@ -0,0 +1,367 @@
...
+	 *  - when dp queue is shared, available_data + free_space DOES NOT eq data_buffer_size
+	 *  - when dp queue is not shared available_data + free_space always == data_buffer_size
+	 */
+	uint32_t free_space;
+
+	bool hw_params_configured;
+	struct sof_audio_stream_params audio_stream_params;
+
+	struct k_spinlock lock;
+};
```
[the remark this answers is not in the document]
marcinszkudlinski: in fact we do need more...
I did not catch it before, but struct k_spinlock must be in uncached memory, but struct dp_queue may be either shared/not shared (as discussed lately). 
(somehow I missed it because I focused mostly on multicore)

FIX: use spinlock for shared queue, irq_locks for not shared. To be fixed

src/audio/dp_queue.c
marcinszkudlinski: We need to double check if clang linker is able to inline such functions (I know there are "out there" linkers smart enough). We'll be able to eat a cake and have it
OR we can simply go to C++ ;)

src/audio/dp_queue.c
```diff
@@ -0,0 +1,367 @@
...
+{
+	uint32_t align;
+	struct dp_queue *dp_queue;
+
+	/* allocate DP structure */
+	if (flags & DP_QUEUE_MODE_SHARED)
+		dp_queue = rzalloc(SOF_MEM_ZONE_RUNTIME_SHARED, 0, SOF_MEM_CAPS_RAM,
+				   sizeof(*dp_queue));
+	else
+		dp_queue = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*dp_queue));
```
kv2019i, line 310: Just logging a comment that this is not compatible with spinlock use, so needs to be modified. You pointed this yourself already, but just to keep track.

kv2019i (changes requested):
Looks good to me. The issue with spinlock with non-shared use needs to fixed, but otherwise seems good to go.

src/audio/dp_queue.c
marcinszkudlinski: unfortunately they're needed outside, till Pipeline2.0 is ready. Now there's a need for some tricks in module adapter to make connection to existing design

marcinszkudlinski:
@lgirdwood 
Probably a lock like below will need to be used in the system several times. Maybe it is a good idea to export it to a library?

```
union lock_key {
	k_spinlock_key_t spinlock;
	int irq_lock;
};

static inline union lock_key dp_queue_lock(struct dp_queue *dp_queue)
{
	/* use cross-core spinlock in case of shared queue
	 * When shared, dp_queue structure is located in not cached memory
	 * as required by spinlock
	 */
	union lock_key key;

	if (dp_queue_is_shared(dp_queue))
		key.spinlock = k_spin_lock(&dp_queue->lock);
	else
		/* use faster irq_lock in case of not shared queue (located in cached mem) */
		key.irq_lock = irq_lock();

	return key;
}

static inline void dp_queue_unlock(struct dp_queue *dp_queue, union lock_key key)
{
	if (dp_queue_is_shared(dp_queue))
		k_spin_unlock(&dp_queue->lock, key.spinlock);
	else
		irq_unlock(key.irq_lock);
}

```

src/audio/dp_queue.c
```diff
@@ -0,0 +1,386 @@
...
+	bool hw_params_configured;
+	struct sof_audio_stream_params audio_stream_params;
+
+	struct k_spinlock lock;
+};
+
+union lock_key {
+	k_spinlock_key_t spinlock;
+	int irq_lock;
+};
```
kv2019i, line 44: This goes a bit out of the scope of this PR, but given we have very little use of locking in sof/src/audio (mostly in the DAI code, chain-dma and kpb), this is worth a discussion. 

Given we might want to run some of the audio modules in unprivileged code, having code that uses spinlocks or irqlocks is not really future proof. So I wonder if we should just move to mutexes in places like this, which would be available both for user-space and kernel threads.

This does incur a perf penalty, and might cause problems as we still have non-Zephyr SOF users who run audio components in IRQ context, so this is probably not something we can do now, but I'd say this should be the longterm solution for audio code.

kv2019i (approved):
A comment w.r.t. locks inline, but not blocking this PR. My immediate concerns addressed, so +1 to this.

src/audio/dp_queue.c
```diff
@@ -0,0 +1,386 @@
...
+
+	return 0;
+}
+
+static size_t dp_queue_get_data_available(struct sof_source *source)
+{
+	struct dp_queue *dp_queue = dp_queue_get_from_source(source);
+
+	/* access is read only, using uncached alias, no need to lock */
+
```
btian1, line 211: redundant empty line

btian1 (approved):

src/audio/dp_queue.c
```diff
@@ -0,0 +1,391 @@
...
+#include <rtos/alloc.h>
+#include <ipc/topology.h>
+
+struct dp_queue {
+	struct sof_source source_api;  /**< src api handler */
+	struct sof_sink sink_api;      /**< sink api handler */
+
+	uint32_t ibs;
+	uint32_t obs;
+	uint32_t flags;
```
lgirdwood, line 20: nitpick: I would comment which flags are used here.

lgirdwood (approved):

lgirdwood:
@marcinszkudlinski I'm seeing a CI build error with Zephyr
```
ccache /home/cwd_user/zephyr-sdk-0.16.1/xtensa-intel_ace15_mtpm_zephyr-elf/bin/xtensa-intel_ace15_mtpm_zephyr-elf-gcc -DCC_OPTIMIZE_FLAGS=\"-O2\" -DKERNEL -DXCC_TOOLS_VERSION=\"zephyr\" -D__ZEPHYR__=1 -DRELATIVE_FILE=\"../sof/src/audio/dp_queue.c\" -I/zep_workspace/zephyr/include -I/zep_workspace/build-mtl/zephyr/include/generated -I/zep_workspace/zephyr/soc/xtensa/intel_adsp/ace -I/zep_workspace/zephyr/soc/xtensa/intel_adsp/ace/include -I/zep_workspace/zephyr/soc/xtensa/intel_adsp/ace/include/intel_ace15_mtpm -I/zep_workspace/zephyr/soc/xtensa/intel_adsp/common/include -I/zep_workspace/zephyr/drivers -I/zep_workspace/sof/zephyr/include -I/zep_workspace/sof/src/platform/intel/ace/include -I/zep_workspace/sof/src/platform/meteorlake/include -I/zep_workspace/sof/src/include/sof/audio/module_adapter/iadk -I/zep_workspace/sof/src/include/sof/audio/module_adapter/library -I/zep_workspace/modules/hal/xtensa/include -I/zep_workspace/modules/hal/xtensa/XTENSA_HAL -I/zep_workspace/modules/hal/xtensa/zephyr/soc/intel_ace15_mtpm -I/zep_workspace/sof/rimage/src/include -I/zep_workspace/sof/src/include -I/zep_workspace/sof/src/arch/xtensa/include -I/zep_workspace/sof/third_party/include -I/zep_workspace/sof/xtos/include -isystem /zep_workspace/zephyr/lib/libc/minimal/include -isystem /opt/toolchains/zephyr-sdk-0.16.1/xtensa-intel_ace15_mtpm_zephyr-elf/bin/../lib/gcc/xtensa-intel_ace15_mtpm_zephyr-elf/12.2.0/include -isystem /opt/toolchains/zephyr-sdk-0.16.1/xtensa-intel_ace15_mtpm_zephyr-elf/bin/../lib/gcc/xtensa-intel_ace15_mtpm_zephyr-elf/12.2.0/include-fixed -fno-strict-aliasing -O2 -imacros /zep_workspace/build-mtl/zephyr/include/generated/autoconf.h -ffreestanding -fno-common -g -gdwarf-4 -fdiagnostics-color=always --sysroot=/home/cwd_user/zephyr-sdk-0.16.1/xtensa-intel_ace15_mtpm_zephyr-elf/xtensa-intel_ace15_mtpm_zephyr-elf -imacros /zep_workspace/zephyr/include/zephyr/toolchain/zephyr_stdint.h -Wall -Wformat -Wformat-security -Wno-format-zero-length -Wno-pointer-sign -Wpointer-arith -Wexpansion-to-defined -Wno-unused-but-set-variable -Werror=implicit-int -fno-pic -fno-pie -Werror -fno-asynchronous-unwind-tables -fstrict-overflow -fno-reorder-functions --param=min-pagesize=0 -fno-defer-pop -fmacro-prefix-map=/zep_workspace/sof/app=CMAKE_SOURCE_DIR -fmacro-prefix-map=/zep_workspace/zephyr=ZEPHYR_BASE -fmacro-prefix-map=/zep_workspace=WEST_TOPDIR -ffunction-sections -fdata-sections -mlongcalls -DXTENSA_TOOLCHAIN_VARIANT=1 -std=c99 -nostdinc -fno-inline-functions -std=gnu99 -MD -MT modules/sof/CMakeFiles/modules_sof.dir/zep_workspace/sof/src/audio/dp_queue.c.obj -MF modules/sof/CMakeFiles/modules_sof.dir/zep_workspace/sof/src/audio/dp_queue.c.obj.d -o modules/sof/CMakeFiles/modules_sof.dir/zep_workspace/sof/src/audio/dp_queue.c.obj -c /zep_workspace/sof/src/audio/dp_queue.c
/zep_workspace/sof/src/audio/dp_queue.c: In function 'dp_queue_create':
/zep_workspace/sof/src/audio/dp_queue.c:344:9: error: implicit declaration of function 'sink_set_obs'; did you mean 'sink_set_rate'? [-Werror=implicit-function-declaration]
  344 |         sink_set_obs(&dp_queue->sink_api, ibs);
      |         ^~~~~~~~~~~~
      |         sink_set_rate
/zep_workspace/sof/src/audio/dp_queue.c:345:9: error: implicit declaration of function 'source_set_ibs'; did you mean 'source_set_rate'? [-Werror=implicit-function-declaration]
  345 |         source_set_ibs(&dp_queue->source_api, obs);
      |         ^~~~~~~~~~~~~~
      |         source_set_rate
[227/378] Building C object modules/sof/CMakeFiles/modules_sof.dir/lib/alloc.c.obj
[228/378] Building C object modules/sof/CMakeFiles/modules_sof.dir/lib/cpu.c.obj
```

marcinszkudlinski:
github.com/https://github.com/thesofproject/sof/pull/8002
dependency

marcinszkudlinski:
rebased after https://github.com/thesofproject/sof/pull/8002 merge

kv2019i:
@marcinszkudlinski  Is https://github.com/thesofproject/sof/issues/8006 still a hard dependency (like in description), or is this ok to merge once reviews-and-testing is ok?

src/include/sof/audio/dp_queue.h
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * DP queue is a circular buffer providing safe consumer/producer cached operations cross cores
+ *
+ * prerequisites:
+ *  1) incoming and outgoing data rate MUST be the same
+ *  2) Both data consumer and data producer declare max chunk sizes they want to use (IBS/OBS)
+ *
+ * The queue may work in 2 modes
+ * 1) simple mode
```
lyakh, line 22: "local mode?"

src/include/sof/audio/dp_queue.h
```diff
@@ -0,0 +1,88 @@
...
+ *    in case both receiver and sender are located on the same core and cache coherency
+ *    does not matter. dp_queue structure is located in cached memory
+ *    In this case DP Queue is a simple ring buffer
+ *    Buffer size must be:
+ *	- 2*MAX(IBS,OBS) if IBS(obs) is a multiplication of OBS(IBS)
+ *	- 3*MAX(IBS,OBS) otherwise
```
lyakh, line 28: same as for the commit text

src/include/sof/audio/dp_queue.h
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * 2) shared mode
+ *    In this case we need to writeback cache when new data arrive and invalidate cache on
+ *    secondary core. That means the whole cacheline must be used exclusively by sink or by source
+ *    incoming data will be available to use when a cacheline is filled completely
+ *    dp_queue structure is located in shared memory
+ *
+ *    buffer size is always 3*MAX(IBS,OBS,CACHELINE) + CACHELINE
```
lyakh, line 36: also like in commit text

src/include/sof/audio/dp_queue.h
```diff
@@ -0,0 +1,88 @@
...
+ *
+ */
+
+struct dp_queue;
+struct sof_audio_stream_params;
+
+/* DP flags */
+#define DP_QUEUE_MODE_SIMPLE 0
```
lyakh, line 44: "LOCAL?"

lyakh (dismissed):

src/audio/dp_queue.c
```diff
@@ -0,0 +1,391 @@
...
+	 * When shared, dp_queue structure is located in not cached memory
+	 * as required by spinlock
+	 */
+	union lock_key key;
+
+	if (dp_queue_is_shared(dp_queue))
+		key.spinlock = k_spin_lock(&dp_queue->lock);
+	else
+		/* use faster irq_lock in case of not shared queue (located in cached mem) */
+		key.irq_lock = irq_lock();
```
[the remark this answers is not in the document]
marcinszkudlinski: No, 
1) k_pinlock must be located in uncached alias - very problematic here.
2) in any case irq_lock is faster than spinlock

src/audio/dp_queue.c
```diff
@@ -0,0 +1,391 @@
...
+
+	return key;
```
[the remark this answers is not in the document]
marcinszkudlinski: in fact it is single dword (uinon of 2 dwords)
```
union lock_key {
	k_spinlock_key_t spinlock;
	int irq_lock;
};
```
I can cast it to dword witch compilation time assert (sizeof (k_spinlock_key_t) == sizeof(int))
But I don't like it, better return a union

marcinszkudlinski:
> @marcinszkudlinski Is #8006 still a hard dependency (like in description), or is this ok to merge once reviews-and-testing is ok?

There will be a lot of sparse warnings if not implemented (as marked in comments)

src/audio/dp_queue.c
```diff
@@ -0,0 +1,391 @@
...
+	struct dp_queue *dp_queue = dp_queue_get_from_sink(sink);
+
+	/* dp_queue_get_free_size will return free size with adjustment for cacheline if needed */
+	if (req_size > dp_queue_get_free_size(sink))
+		return -ENODATA;
+
+	/* no need to lock, just reading data that may be modified by commit_buffer only
+	 *
+	 * note! a sparse warning will be generated here till
+	 * https://github.com/thesofproject/sof/issues/8006 is implemented
```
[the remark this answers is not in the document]
marcinszkudlinski: can't help it now, it is a soft dependency and I hope we can implement it soon enough

src/audio/dp_queue.c
```diff
@@ -0,0 +1,391 @@
...
+{
+	struct dp_queue *dp_queue = dp_queue_get_from_sink(sink);
+
+	if (commit_size) {
+		union lock_key key = dp_queue_lock(dp_queue);
+
+		if (dp_queue_is_shared(dp_queue)) {
+			/* a shared queue. We need to go through committed cachelines one-by-one
+			 * and if the whole cacheline is committed - writeback cache
+			 * and mark data as available for reading
```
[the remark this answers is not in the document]
marcinszkudlinski: >>What if the reader tries to read it on another core before the writer finishes the cache line?
Can't happen, the reader won't get this data - from reader point of view this data has not been produced yet. dp_queue->available_data won't be incremented till the cacheline is written back

As to using of a partially filled cacheline, as long as the reader won't try to modify data (it may result in cache writeback triggered by CPU and overwriting data being produced) it may be done, but for a price. The price will be one extra invalidation on consumer side and one extra writeback on producer (some of cache already written will have to be written back again, same applies to reading from cache).  I wanted to avoid it, but 1) according to my latest cache measurements and 2) the frame size is usually larger than cacheline and 3) dpQueue code will be simpler - it won't hurt, . Will give a try
also will add a clear comment to sink/source API that data from reading buffer can't be modified by the reader in any way

src/audio/dp_queue.c
```diff
@@ -0,0 +1,391 @@
...
+	 */
+	*buffer_start = dp_queue->data_buffer;
+	*buffer_size = dp_queue->data_buffer_size;
+	*data_ptr = dp_queue->data_buffer + dp_queue->read_offset;
+
+	/* clean cache in provided data range */
+	if (dp_queue_is_shared(dp_queue))
+		dp_queue_invalidate_shared(dp_queue,
+					   (__sparse_force void __sparse_cache *)*data_ptr,
+					   req_size);
```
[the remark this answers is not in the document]
marcinszkudlinski: dp_queue_invalidate_shared takes care for wrapping

src/audio/dp_queue.c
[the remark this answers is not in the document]
lyakh: for anything larger than 32 bits I'd pass a pointer as a parameter and let the function modify it there

src/audio/dp_queue.c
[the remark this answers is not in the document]
lyakh: > will add a clear comment to sink/source API that data from reading buffer can't be modified by the reader in any way

can we make all reader pointers `const`?

src/audio/dp_queue.c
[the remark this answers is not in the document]
marcinszkudlinski: lockless queue introduced, this code is removed

src/audio/dp_queue.c
```diff
@@ -0,0 +1,391 @@
...
+
+static inline void dp_queue_invalidate_shared(struct dp_queue *dp_queue,
+					      void __sparse_cache *ptr, uint32_t size)
+{
+	/* wrap-around? */
+	if ((uintptr_t)ptr + size > (uintptr_t)dp_queue_buffer_end(dp_queue)) {
+		/* writeback till the end of circular buffer */
+		dcache_invalidate_region
+			(ptr, (uintptr_t)dp_queue_buffer_end(dp_queue) - (uintptr_t)ptr);
+		size -= (uintptr_t)dp_queue_buffer_end(dp_queue) - (uintptr_t)ptr;
```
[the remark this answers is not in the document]
marcinszkudlinski: no, definitely not
invalidate to the end of the buffer, **decrease** size so size now says "how many bytes left for invalidation"

marcinszkudlinski:
Added const to pointers provided by source API 
I'm putting it in this PR because in new implementation of dpQueue any modification of source data may be harmful and extremely hard to track.

src/audio/dp_queue.c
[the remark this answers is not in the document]
lyakh: haven't looked at the code yet, but can DP modules have more than one source or sink?

src/audio/dp_queue.c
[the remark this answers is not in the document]
lyakh: ehm, yes, I think you're right

src/audio/dp_queue.c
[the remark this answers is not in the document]
lyakh: it can be avoided by adding type-casts

src/audio/dp_queue.c
```diff
@@ -0,0 +1,351 @@
...
+	*buffer_size = dp_queue->data_buffer_size;
+	*data_ptr = dp_queue_get_pointer(dp_queue, dp_queue->read_offset);
+
+	/* clean cache in provided data range */
+	dp_queue_invalidate_shared(dp_queue, *data_ptr, req_size);
+
+	return 0;
+}
+
+static int dp_queue_release_data(struct sof_source *source, size_t free_size)
```
lyakh, line 218: a static function, so not too important, but maybe a name like "update_read" would be clearer - we don't really release data, we release buffer space

src/include/sof/audio/dp_queue.h
```diff
@@ -0,0 +1,114 @@
...
+ *  - allow both offsets to point from 0 to DOUBLE buffer_size
+ *  - when calculating pointers to data, use: data_bufer[offset % buffer_size]
+ *  - use double buffer size in wrap around check when calculating available data
+ *
+ * And now:
+ *   - write_offset == read_offset
+ *		always means "buffer empty"
+ *   - write_offset == read_offset + buffer_size
+ *		always means "buffer full"
+ */
```
lyakh, line 64: the main differences to the current audio_stream.h implementation seem to be:
1. DP Queue doesn't overwrite data - if the buffer is full, writing will fail
2. DP Queue uses the "double buffer size" trick to avoid a shared "buffer full" flag, which allows lockless use
3. The use of cached or uncached pointers for the control structure is tightly connected to the buffer being shared or not
4. It's using the sink / source API

Is that correct? So in principle this buffer implementation could be made generic and used for all audio buffers, but since our current audio-stream buffers have to support overwriting, the producer has to be able to update the read-pointer, so it needs locking

src/audio/dp_queue.c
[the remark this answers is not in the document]
marcinszkudlinski: yes, it can. Yet the PR implementing 8006 is on its way, probably early next week. Either a comment (to be removed) or cast (to be removed) 
Probably none of them will be merged

src/audio/dp_queue.c
```diff
@@ -0,0 +1,351 @@
...
+
+struct sof_sink *dp_queue_get_sink(struct dp_queue *dp_queue)
+{
+	return &dp_queue->sink_api;
+}
+
+struct sof_source *dp_queue_get_source(struct dp_queue *dp_queue)
+{
+	return &dp_queue->source_api;
+}
```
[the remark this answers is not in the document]
marcinszkudlinski: Not unless whole struct dp_queue is exported too
I think that it is worth some discussion - how to make something really private without harming performance...

src/include/sof/audio/dp_queue.h
marcinszkudlinski: _but since our current audio-stream buffers have to support overwriting_ 

I see audio_stream_produce() does support overwriting, but 
 - for LL it makes no point at all - LL must produce a complete portion of data from input and send a complete portion to output. If any of LL fails - it is a glitch anyway
 -   for DP - a module will be run only if it is ready to process, in common case it means "have enough data on all inputs and enough space on all outputs" 

so yes _:DP Queue doesn't overwrite data - if the buffer is full, writing will fail_. In each case - overwrite or drop - data stream is corrupted

_"So in principle this buffer implementation could be made generic and used for all audio buffers,"_
No, it is to heavy to be at each place

There's a need for a couple more queues:

1) simple reusable LL buffer
 For ALL single-core LLs in chain we do need a **single** memory space for sink and a **single** **source** 
 LL produces and consumes **a complete portion of data**, leaving a source buffer empty. Then a next LL is run. So the empty memory space may be easily re-used as a sink for next LL

 2) cross core LL buffer
  for connected cross core LL pipelines. It may be dpQueue, it will just need a smaller buffer size, 1*MAX(ibs,obs) in that case + some tiny details. May be done i,e, by another DP_QUEUE_MODE_ flag

3) linearization dpQueue
the heaviest of all, to be used if a module required a linear data instead of circular, providing internal linearization of data. My idea is that the module may request this somehow (TBD how) and a proper queue is created for it. Note that is applies to DP only. LLs will have linear data in natural way (always start from buffer beginning)

src/audio/dp_queue.c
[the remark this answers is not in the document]
lyakh: I'm certainly also for limiting the scope of types, but unfortunately that means reduced performance because of inability to use inline functions. If we were programming Firefox or LibreOffice or Gnome, I'd sacrifice performance for the sake of correctness. But for DSP firmware maybe we should do the opposite? Especially for code, used on hot paths during audio data processing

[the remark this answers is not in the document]
marcinszkudlinski: a discussion opened https://github.com/thesofproject/sof/issues/8100

marcinszkudlinski:
TODOs: 
1) wait for sparse changes
2) performance vs enacpsualtion discussion https://github.com/thesofproject/sof/issues/8100

lgirdwood:
@marcinszkudlinski btw, sparse fixes merged, you should be able to rebase and fix any local issues.

marc-hb:
As long as there is no git conflict you don't need to rebase to get sparse fixes, force push in place is enough. CI always merges with the latest.

This avoids rebase noise.

lgirdwood:
> As long as there is no git conflict you don't need to rebase to get sparse fixes, force push in place is enough. CI always merges with the latest.

Are you sure ? i.e. rerunning some GH actions like "Zephyr / build-linux (*)" manually will still reproduce build errors after build fixes have been merged to HEAD. Not a blocker for anyone, but an annoyance only.

marc-hb:
> >  As long as there is no git conflict you don't need to rebase to get sparse fixes, force push in place is enough. CI always merges with the latest.

>  i.e. rerunning some GH actions like "Zephyr / build-linux (*)" manually will still reproduce build errors after build fixes have been merged to HEAD.

Correct, Github's "re-run" button re-tests the same merge commit. You need to force push a new SHA1.

I think SOFCI TEST is different.

tl;dr: just increase the commit date and force-push. No need to rebase when there is no git conflict.

marc-hb:
> @marcinszkudlinski btw, sparse fixes merged

The cross-CI whack-a-mole continues:
- https://github.com/zephyrproject-rtos/zephyr/pull/62101#issuecomment-1703391086

marcinszkudlinski:
All requested changes addressed
1) inline functions for performance
2) rebased on https://github.com/thesofproject/sof/pull/8106 to avoid rebase problems 

PLEASE SKIP 1st COMMIT IN REVIEW

src/audio/dp_queue.c
```diff
@@ -0,0 +1,298 @@
...
+	/* writeback rest of data */
+	dcache_writeback_region(ptr, size);
+}
+
+static inline
+uint8_t __sparse_cache *dp_queue_get_pointer(struct dp_queue *dp_queue, uint32_t offset)
+{
+	/* check if offset is not in "double area" */
+	if (offset >= dp_queue->data_buffer_size)
+		offset -= dp_queue->data_buffer_size;
```
RanderWang, line 80: I am confused that dp_queue_inc_offset using if (offset >= 2 * dp_queue->data_buffer_size) but here not * 2

src/audio/dp_queue.c
```diff
@@ -0,0 +1,298 @@
...
+	return dp_queue->_data_buffer + offset;
+}
+
+static inline
+uint32_t dp_queue_inc_offset(struct dp_queue *dp_queue, uint32_t offset, uint32_t inc)
+{
+	offset += inc;
+	/* wrap around ? 2*size because of "double area" */
+	if (offset >= 2 * dp_queue->data_buffer_size)
+		offset -= 2 * dp_queue->data_buffer_size;
```
RanderWang, line 90: how about if  offset  is still > 2 * dp_queue->data_buffer_size ?

RanderWang (commented):
LGTM

src/audio/dp_queue.c
marcinszkudlinski: that means - somebody tries to increase offset more than twice size of a ring buffer. A defect.
you're right, assert(inc <= data_buffer_size) might be useful in this place

src/audio/dp_queue.c
marcinszkudlinski: ring buffer range is from **0** to **(data_buffer_size-1)**
offset may get values from **0** to **(2*data_buffer_size-1)** as explained before - to distinguish between "buffer full" and "buffer empty" in lockless code, but should be taken modulo data_buffer_size 

lines 78-80 are doing a faster version of operation: ``offset %= dp_queue->data_buffer_size;``
will add a comment

iganakov (approved):

lgirdwood:
> all deps resolved, rebased on newest main

@marcinszkudlinski Can you check in the CI results. Thanks !

marcinszkudlinski:
> > all deps resolved, rebased on newest main
> 
> @marcinszkudlinski Can you check in the CI results. Thanks !

looks like there's no results:
https://sof-ci.01.org/sof-pr-viewer/#/build/PR8020/build12764971

pushing an empty commit (rebase to newest main) to retrigger

src/audio/audio_stream.c
```diff
@@ -49,7 +49,8 @@ static size_t audio_stream_get_data_available(struct sof_source *source)
 }
```
kv2019i, line 49: @marcinszkudlinski Maybe remove the "Conflicts: .." bits from commit message...?

src/audio/dp_queue.c
```diff
@@ -0,0 +1,300 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// Copyright(c) 2021 Intel Corporation. All rights reserved.
```
kv2019i, line 3: 2023 for a new file?

kv2019i (commented):
One extra merge-conflict bit in the commit message, can you clean up @marcinszkudlinski . Otherwise I can re-confirm my +1 for this.

marcinszkudlinski:
all green

jxstelter (approved):

src/include/sof/audio/dp_queue.h
```diff
@@ -0,0 +1,169 @@
...
+
+	/* private: */
+	struct sof_source _source_api;  /**< src api handler */
+	struct sof_sink _sink_api;      /**< sink api handler */
+
+	uint32_t _flags;		/* DP_QUEUE_MODE_* */
+
+	uint8_t __sparse_cache *_data_buffer;
+	uint32_t _write_offset;		/* private: to be modified by data producer using API */
+	uint32_t _read_offset;		/* private: to be modified by data consumer using API */
```
lyakh, line 95: both `size_t`?

src/audio/dp_queue.c
```diff
@@ -0,0 +1,300 @@
...
+	/* wrap around ? 2*size because of "double area" */
+	if (offset >= 2 * dp_queue->data_buffer_size)
+		offset -= 2 * dp_queue->data_buffer_size;
+	return offset;
+}
+
+static inline
+size_t _dp_queue_get_data_available(struct dp_queue *dp_queue)
+{
+	int32_t avail_data =  dp_queue->_write_offset - dp_queue->_read_offset;
```
lyakh, line 100: `ssize_t`

src/audio/dp_queue.c
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	return 0;
+}
+
+static int dp_queue_get_data(struct sof_source *source, size_t req_size,
+			     void const **data_ptr, void const **buffer_start, size_t *buffer_size)
+{
+	struct dp_queue *dp_queue = dp_queue_from_source(source);
+	__sparse_cache void *_data_ptr;
```
lyakh, line 164: most SOF code doesn't use variables / structure fields beginning or ending with underscores

src/audio/dp_queue.c
```diff
@@ -0,0 +1,300 @@
...
+		    &dp_queue->audio_stream_params);
+	sink_init(dp_queue_get_sink(dp_queue), &dp_queue_sink_ops,
+		  &dp_queue->audio_stream_params);
+
+	/* set obs/ibs in sink/source interfaces */
+	sink_set_obs(&dp_queue->_sink_api, obs);
+	source_set_ibs(&dp_queue->_source_api, ibs);
+
+	uint32_t max_ibs_obs = MAX(ibs, obs);
+	uint32_t min_ibs_obs = MIN(ibs, obs);
```
lyakh, line 276: `size_t`

src/audio/dp_queue.c
```diff
@@ -0,0 +1,300 @@
...
+
+	/* calculate required buffer size */
+	if (max_ibs_obs % min_ibs_obs == 0)
+		dp_queue->data_buffer_size = 2 * max_ibs_obs;
+	else
+		dp_queue->data_buffer_size = 3 * max_ibs_obs;
```
lyakh, line 282: I still don't understand this calculation...

tmleman (approved):

kv2019i:
CI is good. Zephyr mainline CIs fail due to https://github.com/zephyrproject-rtos/zephyr/issues/62464

@marcinszkudlinski @lyakh let's address the size_t/int variables, and if needed, clarify the rationale for 2/3x buffer sizing, in a follow-up PR. Thanks all for reviews!

src/audio/dp_queue.c
marcinszkudlinski: lest go with examples:

EXAMPLE 1:
 Consume 2 pieces each cycle
 produce 4 pieces every 2 cycles

buf size: 8

CYCLE0 prodice 4 pieces
buffer contains: 1111

CYCLE1 consume 2 pieces
buf: 11

CYCLE2 - we DO NOT know who's gonna be first - consumer or producer
  consumer first: empty -> 2222
  producer first: 112222 -> 2222
  
top buffer occupation is 6 ( 2*MAX(IBS,OBS)  = 8, operation safe)

EXAMPLE 2:
 Consume 3 pieces each 3 cycles
 produce 5 pieces every 5 cycles

(this cannot happen now, as currently there's no DP-DP connection, but may in the future)

cycle0 produce 3 pieces
111

cycle3 produce 3 pieces
111222

cycle5 - consumer must wait, not enough data
111222

cycle6 produce 3 pieces
111222333

cycle9 produce 3 pieces
111222333444 

cycle 10 - consume 5 pieces
2333444 

**top buffer occupation is 12.** 
2*MAX(IBS,OBS) = 10 - DROP!!!! no buffer space
3*MAX(IBS,OBS) = 15 operation safe

PR #8026, opened 2023-08-11
Volume module code refine and re-position

btian1:
1. remove legacy interface from volume, since it is only used for CAVS 1.X and 2.0 platform.
2. create new ipc3 and ipc4 specific files to cover ipc3 and ipc4 specific code.
3. move volume from module_adapter/modlue/volume to src/audio folder.


btian1:
@lgirdwood , current ipc3/ip4 folder only have volume.c, all other code still keep as common, do you prefer to copy those common code to ipc3/ipc4 separately? 

During our meeting， Seppo mentioned want to move out volume folder from current to sof/src/audio folder, what's your opinion?

src/audio/module_adapter/module/volume/ipc3/volume.c
```diff
@@ -71,7 +71,7 @@ DECLARE_TR_CTX(volume_tr, SOF_UUID(volume_uuid), LOG_LEVEL_INFO);
 #if CONFIG_FORMAT_S16LE
 /**
  * \brief Used to find nearest zero crossing frame for 16 bit format.
- * \param[in,out] source Source buffer.
+ * \param[in,out] Source buffer.
```
lyakh, line 61: I think the original was correct: name of the parameter, followed by a description. Same below in this file and in ipc4/volume.c

lyakh (commented):
probably a matter of taste, but if we aren't adding more files to these ipc3 and ipc4 directories, maybe just add a suffix to the file name like we did for src/samples/audio/smart_amp_test_ipc[34].c

src/audio/module_adapter/module/volume/ipc3/volume.c
btian1: yes, I know there is nothing wrong here, however, patch checker is reporting error for this, so have to made this change.

btian1 (commented):
this is not done yet, if simply add postfix with ipc3 and ipc4, then future volume related work still need check ipc3/ipc4 both.

Since volume is an very important module, I prefer to copy all files to each folder to get a totally separate module, and then move it to src/audio/volume, what do you think? current position seems moved due to module adapter purpose.
@lyakh

lyakh:
> this is not done yet, if simply add postfix with ipc3 and ipc4, then future volume related work still need check ipc3/ipc4 both.
> 
> Since volume is an very important module, I prefer to copy all files to each folder to get a totally separate module, and then move it to src/audio/volume, what do you think? current position seems moved due to module adapter purpose. @lyakh

Sorry, don't understand. What do you mean "future work needs to check?" Isn't it the same whether you have volume/ipc4/volume.c or volume/volume_ipc4.c? I think in both cases it's obvious what the file is for. It's just getting pretty deep already with src/audio/module_adapter/module/volume/ipc4/ and creating 2 directories for 1 file each seems a bit of an overkill.
However, actually maybe we shouldn't duplicate the file completely. I see that the difference is rather large, but I assume there's also a lot of common code. Can we have
volume/volume_common.c
volume/volume_ipc4.c
volume/volume_ipc3.c
?

src/audio/module_adapter/module/volume/ipc3/volume.c
btian1: git commit failure, you can't submit if not change, :(

btian1:
> 

your statement is my initial thought, yes, this will looks code more reasonable, and this is also the reason I stopped to submit more patches for volume,  however, within this change, I want to have below purpose: 
totally separate ipc3/ipc4 volume, since our CI does not cover ipc3 anymore, if we still have common files, then any volume related changes have to be tested with ipc3.

If copy each file to ipc3/ipc4 accordingly, then any future change in ipc4 folder will not impact ipc3 anymore, we will have no concern for merge speed will be much faster.

src/audio/module_adapter/module/volume/ipc3/volume.c
lyakh: `git commit -n`

lyakh:
> > 
> 
> your statement is my initial thought, yes, this will looks code more reasonable, and this is also the reason I stopped to submit more patches for volume, however, within this change, I want to have below purpose: totally separate ipc3/ipc4 volume, since our CI does not cover ipc3 anymore, if we still have common files, then any volume related changes have to be tested with ipc3.
> 
> If copy each file to ipc3/ipc4 accordingly, then any future change in ipc4 folder will not impact ipc3 anymore, we will have no concern for merge speed will be much faster.

That's true, yes, but most of SOF is IPC version independent - used both for IPC3 and IPC4. And we do have to touch and modify that common code. We also touch and modify code, also used on other platforms and still with XTOS - and none of that is tested by our CI. So, no, I think we have to keep updating common code and for that we have to keep it shared instead of duplicating it.

btian1:
> > > 
> > 
> > 
> > your statement is my initial thought, yes, this will looks code more reasonable, and this is also the reason I stopped to submit more patches for volume, however, within this change, I want to have below purpose: totally separate ipc3/ipc4 volume, since our CI does not cover ipc3 anymore, if we still have common files, then any volume related changes have to be tested with ipc3.
> > If copy each file to ipc3/ipc4 accordingly, then any future change in ipc4 folder will not impact ipc3 anymore, we will have no concern for merge speed will be much faster.
> 
> That's true, yes, but most of SOF is IPC version independent - used both for IPC3 and IPC4. And we do have to touch and modify that common code. We also touch and modify code, also used on other platforms and still with XTOS - and none of that is tested by our CI. So, no, I think we have to keep updating common code and for that we have to keep it shared instead of duplicating it.

Thanks, Guennadi, I am both ok for two directions, let's get more feedback from @lgirdwood @kv2019i 

I am refining volume part code to separate ipc3/ipc4, there are two directions:
1.  use volume_ipc3.c/volume_ipc4.c to separate, and other code keep as common, may need add a volume_common.c as well.
2. Totally separate ipc3/ipc4 with different folders, common files are copied two different directories.

Let me know.

src/audio/CMakeLists.txt
```diff
@@ -163,7 +163,12 @@ check_optimization(hifi3 -mhifi3 "" -DOPS_HIFI3)
 set(sof_audio_modules mixer volume src asrc eq-fir eq-iir dcblock crossover tdfb drc multiband_drc mfcc mux)
 
 # sources for each module
-set(volume_sources module_adapter/module/volume/volume.c module_adapter/module/volume/volume_generic.c)
+if(CONFIG_IPC_MAJOR_3)
+	set(volume_sources module_adapter/module/volume/volume_ipc3.c module_adapter/module/volume/volume_generic.c)
+elseif(CONFIG_IPC_MAJOR_4)
+	set(volume_sources module_adapter/module/volume/volume_ipc4.c module_adapter/module/volume/volume_generic.c)
+endif()
+
```
lgirdwood, line 171: Looks like sources are the same except the ipc file. Does cmake support creating a ```volume_sources``` with common files and then only *add* ipc3/4 depending on the Kconfig ?

src/audio/module_adapter/CMakeLists.txt
```diff
@@ -11,8 +11,13 @@ if((NOT CONFIG_LIBRARY) OR CONFIG_LIBRARY_STATIC)
 		module/volume/volume_hifi4.c
 		module/volume/volume_generic_with_peakvol.c
 		module/volume/volume_hifi3_with_peakvol.c
-		module/volume/volume_hifi4_with_peakvol.c
-		module/volume/volume.c)
+		module/volume/volume_hifi4_with_peakvol.c)
+
+	if(CONFIG_IPC_MAJOR_3)
+	add_local_sources(sof module/volume/volume_ipc3.c)
```
lgirdwood, line 17: indentation ?

lgirdwood (commented):
I think this is a good IPC abstraction simplification as we are cleanly separating IPC specific code away from common coder.

lgirdwood:
@btian1 @lyakh its going to be cleaner if we take volume (and then all otehr modules) and move them to their own directories. e.g.
```
src/audio/volume
    volume.c
    ipc3.c
    ipc4.c
    hifi3.c
    volume.h
```

lyakh:
> @btian1 @lyakh its going to be cleaner if we take volume (and then all otehr modules) and move them to their own directories. e.g.
> 
> ```
> src/audio/volume
>     volume.c
>     ipc3.c
>     ipc4.c
>     hifi3.c
>     volume.h
> ```

@lgirdwood that's exactly what I'm proposing: volume already has its own directory sof/src/audio/module_adapter/module/volume/ so I'm proposing to put IPC version dependent files in the same directory as sof/src/audio/module_adapter/module/volume/volume-ipc3.c and sof/src/audio/module_adapter/module/volume/volume-ipc4.c and keep the common part as sof/src/audio/module_adapter/module/volume/volume.c

lgirdwood:
> > @btian1 @lyakh its going to be cleaner if we take volume (and then all otehr modules) and move them to their own directories. e.g.
> > ```
> > src/audio/volume
> >     volume.c
> >     ipc3.c
> >     ipc4.c
> >     hifi3.c
> >     volume.h
> > ```
> 
> @lgirdwood that's exactly what I'm proposing: volume already has its own directory sof/src/audio/module_adapter/module/volume/ so I'm proposing to put IPC version dependent files in the same directory as sof/src/audio/module_adapter/module/volume/**volume-ipc3.c** and sof/src/audio/module_adapter/module/volume/**volume-ipc4.c** and keep the common part as sof/src/audio/module_adapter/module/volume/**volume.c**

Ok, but we should have them in src/audio/modules/ as it's self explanitory.

src/audio/CMakeLists.txt
btian1: source will not be same in following patches, this patch just create ipc3 and ipc4 source file, following patches will remove each unused code, and common code out, please stay tuned.

src/audio/module_adapter/CMakeLists.txt
btian1: I checked following code, no indentation used, so I did not add indentation.

btian1:
> 

@lgirdwood, do you mean you want to keep volume in src/audio/module_adapter/module folder?
Personally, I want to put it to src/audio/volume, any reason keep volume inside module adapter?

lyakh (commented):
I just checked: removed all IPC-version-specific functions from volume.c and replaced IPC-version-specific code in functions with one-line function calls. Took me about 5 minutes and I was left with more than 1000 lines of common code. I'd keep that common code in volume.c and only move IPC-version-specific code into volume-ipc3.c and volume-ipc4.c

kv2019i (commented):
Sorry, a bit late to the review. Splitting up the commits with first duplicating the code in the series, and then removing out the common bits in latter patches in the series, is a bit unconventional way, but works this way as well I guess. In general, I think it'd be really important we try to maximize the common code and not create any incentives to add to IPC3/4 specific code. Ifdefs are not great for readibility though, so I do like the net outcome of this series.

src/audio/module_adapter/CMakeLists.txt
lgirdwood: ok, but the other code should use conditional indentation, so that can be fixed.

src/audio/module_adapter/CMakeLists.txt
```diff
@@ -6,18 +6,18 @@ add_local_sources(sof module_adapter.c module/generic.c)
...
-		module/volume/volume_hifi3_with_peakvol.c
-		module/volume/volume_hifi4_with_peakvol.c
-		module/volume/volume_common.c)
+		../volume/volume_generic.c
+		../volume/volume_hifi3.c
+		../volume/volume_hifi4.c
+		../volume/volume_generic_with_peakvol.c
+		../volume/volume_hifi3_with_peakvol.c
+		../volume/volume_hifi4_with_peakvol.c
+		../volume/volume_common.c)
```
lgirdwood, line 15: This need to be in a Cmakefile in the module directory i.e. 
```src/audio/volume/CMakeList.txt```

**Review by lgirdwood — COMMENTED** [review]
Almost good to merge, but I agree with @kv2019i we do need to make this PR bisectable.

File: src/audio/module_adapter/CMakeLists.txt (hunk 4 above)
Comment (lgirdwood) [reply src/audio/module_adapter/CMakeLists.txt:15]: also no need to use ```_with_``` in file names, I would just call the files (using above order)
```
volume/generic.c
volume/hifi3.c
volume/hifi4.c
volume/peakvol.c
volume/hifi3_peakvol.c
volume/hifi4_peakvol.c
volume/common.c
```

**Review by btian1 — COMMENTED** [review]
@kv2019i @lgirdwood  @lyakh , no worry about bisect,
I always keep each patch independent.
each patch will pass CI then add another new patch.

@dbaluta @iuliana-prodan , could you check from ipc3 side as well? since ipc3 volume also got changed.

File: src/audio/module_adapter/module/volume/volume_common.c [hunk 5]
```diff
@@ -0,0 +1,326 @@
...
+		/* Invert sign for volume down ramp step */
+		if (delta < 0)
+			coef = -coef;
+
+		cd->ramp_coef[chan] = coef;
+		comp_dbg(mod->dev, "cd->ramp_coef[%d] = %d", chan, cd->ramp_coef[chan]);
+	}
+
+	return 0;
+}
```
lyakh: it should be larger, about 3 times as large. Please see attached example. Note, it isn't complete - I just replaced some IPC[34]-specific code with dummy function calls, you'd need to implement those properly and add respective parameters
[volume.c.zip](https://github.com/thesofproject/sof/files/12415809/volume.c.zip)

btian1: do you mean move more code to common source file?  from my opinion, common file just used to save exactly same code, not all possible common code with more modification on both side, as long as ipc3 and ipc4 are totally separated, and ipc3 ipc4 source file is in right size(<1000), it should be ok?

src/audio/module_adapter/CMakeLists.txt
kv2019i: @lgirdwood @btian1  both are fine to me, but we have still kept the module/component prefix in filenames (e.g. we have pipeline/pipeline-foobar.c so to be consistent, we should have volume/volume_generic.c:

```
├── src
│   ├── CMakeLists.txt
│   ├── src.c
│   ├── src_generic.c
│   ├── src_hifi2ep.c
│   ├── src_hifi3.c
│   └── src_hifi4.c
├── tdfb
│   ├── CMakeLists.txt
│   ├── tdfb.c
│   ├── tdfb_direction.c
│   ├── tdfb_generic.c
│   ├── tdfb_hifi3.c
│   └── tdfb_hifiep.c
....
```

Personally I'd rather have a module prefix in filename. It's a bit confusing to have "generic.c" open in editor and not immediately see what instance of generic.c this one is (maybe just an old grumpy emacs user here). But oh well, just my personal preference.

src/audio/module_adapter/module/volume/volume_common.c
kv2019i: Hmm, looking at @lyakh 's code, so I understand right the proposal that we'd define function pointers for things like "vol_get_processing_function()" (with some optional parameters in the function template). So then we could have common code for e.g. prepare() (like in Guennadi's example), and only have the ops setup in ipc3/ipc4.c. I think this sounds reasonable, but devil may be in the details.

src/audio/module_adapter/CMakeLists.txt
lgirdwood: @kv2019i btw, what does mtrace show wrt filenames ?

kv2019i: @kv2019i logging uses module names and does not use the filename (e.g. we init with "LOG_MODULE_REGISTER(volume, CONFIG_SOF_LOG_LEVEL);").

lgirdwood: ok, I'm happy that logging shows the correct detail and ```volume_``` prefix for file names to help some editors.

src/audio/module_adapter/module/volume/volume_common.c
btian1: Let's me recall the initial purpose of this PR:
volume.c is too large with around 1700 lines code mixed with IPC3 and IPC4

So, this PR main purpose is: 
1. separate IPC3 and IPC4 for volume.
2. shorten source file length.

What's the difference between Guennadi and Mine:

Mine: ipc3.c and ipc4.c is still the main source(separate driver interface), common is lib code.
Guennadi: ipc3.c and ipc4.c is lib ipc specific code, common is main source(ipc3/4 shared driver interface).

with this PR, common source have 300+ lines, and ipc3/ipc4 source have 600+ lines, compared with origin,  
it is much shorten and easy to read and maintain.

So I think this change is good, common is big or less actually is not the critical part of this PR.
Any future code or driver interface change will not impact another ipc, of course, you can change it as well if you want.

src/audio/volume/CMakeLists.txt
```diff
@@ -2,17 +2,17 @@
...
-		volume_hifi3_with_peakvol.c
-		volume_hifi4_with_peakvol.c
-		volume_common.c)
+		generic.c
+		hifi3.c
+		hifi4.c
+		generic_peakvol.c
+		hifi3_peakvol.c
+		hifi4_peakvol.c
+		common.c)
```
lgirdwood, line 11: Lets go with @kv2019i name suggestion here.

lgirdwood (commented):
Good stuff @btian1 - I think we do need to squash some patches though so that moves are atomic in the commit log. i.e. most of these patches could be squashed.

marc-hb:
>  I think we do need to squash some patches though so that moves are atomic in the commit log. i.e. most of these patches could be squashed.

From: https://docs.zephyrproject.org/latest/contribute/contributor_expectations.html#multiple-commits-on-a-single-pr

> Keep in mind each commit in the PR must still build cleanly and pass all the CI tests. [...] The CI system does not enforce this policy, so it is the PR author’s responsibility to verify.


(for obvious bisectability reasons)

In other words, commits must be "as small as possible, but not smaller" (c) :-)

PS: I did not look at this particular PR.

btian1:
> 

Fully agree and aligned with this guideline, each patch have its functionality change and in atomic level, squash is not always good.

src/audio/module_adapter/CMakeLists.txt
lyakh: > Personally I'd rather have a module prefix in filename. It's a bit confusing to have "generic.c" open in editor and not immediately see what instance of generic.c this one is

A weak +1 on this. In general I can live with either solution, that's why I didn't comment on this. And (as many know) I'm all against redundancy so - if the folder is already called "volume" why should files in it have a "volume_" prefix? Same for function and type names: should static functions use a "volume_" prefix or not? But since some time my preference shifted towards this redundancy - to use a prefix in all those cases for pretty much the same reason - for editor, shell and tag use. Imagine if all drivers and modules had functions called `init()`, `configure()`, `copy()`, etc.

src/audio/module_adapter/module/volume/volume_common.c
lyakh: > Hmm, looking at @lyakh 's code, so I understand right the proposal that we'd define function pointers for things like "vol_get_processing_function()" (with some optional parameters in the function template). So then we could have common code for e.g. prepare() (like in Guennadi's example), and only have the ops setup in ipc3/ipc4.c. I think this sounds reasonable, but devil may be in the details.

@kv2019i actually I meant it even simpler - just have two implementations of `vol_get_processing_function()` - one in volume_ipc3.c and one in volume_ipc4.c

src/include/sof/audio/volume.h
```diff
@@ -277,4 +280,110 @@ static inline void peak_vol_update(struct vol_data *cd)
...
+
+	time_ratio = (((int64_t)ramp_time) << 30) / (cd->initial_ramp << 3);
+	pow_value = volume_pow_175(time_ratio);
+	return cd->rvolume[channel] + Q_MULTSR_32X32((int64_t)volume_delta, pow_value, 16, 30, 16);
+}
+#endif
+
+void reset_state(struct vol_data *cd);
+
+static inline void prepare_ramp(struct comp_dev *dev, struct vol_data *cd)
```
lyakh, line 324: `volume_prepare_ramp()`

src/include/sof/audio/volume.h
```diff
@@ -277,4 +280,110 @@ static inline void peak_vol_update(struct vol_data *cd)
...
+	if (!cd->initial_ramp)
+		return cd->tvolume[channel];
+
+	time_ratio = (((int64_t)ramp_time) << 30) / (cd->initial_ramp << 3);
+	pow_value = volume_pow_175(time_ratio);
+	return cd->rvolume[channel] + Q_MULTSR_32X32((int64_t)volume_delta, pow_value, 16, 30, 16);
+}
+#endif
+
+void reset_state(struct vol_data *cd);
```
lyakh, line 322: `volume_reset_state()`

src/audio/module_adapter/module/volume/volume_common.c
btian1: @kv2019i , @lyakh, no devils, it is not a difficult change, I just want to mix ipc3 and ipc4 code anymore.
If go with this way, any further ipc4 changes still need ipc3 verify, since they are still share prepare/params/proces, etc.

src/include/sof/audio/volume.h
btian1: I did want to change like add volume_ prefix, but later find there is no extra errors, so did not make it, let me change.

src/audio/module_adapter/CMakeLists.txt
marc-hb: > if the folder is already called "volume" why should files in it have a "volume_" prefix?

Because most editors suck? :-) They don't display the end of the path.

It also depends on the basename, for instance we have multiple `interrupt.c` but at least "interrupt" means _something_ whereas "generic" and "common" carry _zero_ information. So the latter needs a prefix more than the former.

singalsu:
It's hard to comment the changes of patch but by look of what is left to ipc3 and ipc4 versions, it's still too much. I'd expect more into common module. e.g.

- volume_ramp()
- Module ops like init(), prepare(), process(), reset()

The prepare() differs most in ipc3 vs. ipc4. Could module adapter call before prepare() some separate ipc4 params hook if it is defined, otherwise do same sink and source params update that is duplicated to most components.

Remove of legacy dev api could be a separate PR that would be quick to review and approve?

src/audio/module_adapter/module/volume/volume_ipc4.c
```diff
@@ -0,0 +1,1433 @@
...
+#endif /* CONFIG_FORMAT_S16LE */
+
+#if CONFIG_FORMAT_S24LE
+/**
+ * \brief Used to find nearest zero crossing frame for 24 in 32 bit format.
+ * \param[in,out] source audio stream buffer.
+ * \param[in] frames Number of frames.
+ * \param[in,out] prev_sum Previous sum of channel samples.
+ */
+static uint32_t vol_zc_get_s24(const struct audio_stream __sparse_cache *source,
```
lgirdwood: These dont look ipc4 specific ? aren't the volume operations all IPC generic code ?

src/audio/module_adapter/module/volume/volume_ipc4.c
```diff
@@ -0,0 +1,1433 @@
...
+	cd->channels = 0;
+	cd->ramp_finished = true;
+	cd->vol_ramp_frames = 0;
+	cd->vol_ramp_elapsed_frames = 0;
+	cd->sample_rate_inv = 0;
+	cd->copy_gain = true;
+	cd->is_passthrough = false;
+}
+
+#if CONFIG_IPC_MAJOR_3
```
lgirdwood: IPC3 logic in ipc4 file ?

lgirdwood (commented):
There is a lot of churn going on showing the file movement which is distracting from the end result. @btian1 can you squash into atomic and bisectable commits ? This may even be 1-2 commits

src/include/sof/audio/volume_uuid.h
```diff
@@ -0,0 +1,33 @@
...
+		    0xaf, 0xe9, 0xfe, 0x5d, 0xba, 0xa4, 0xc3, 0x93);
+
+/* 61bca9a8-18d0-4a18-8e7b-2639219804b7 */
+DECLARE_SOF_RT_UUID("gain", gain_uuid, 0x61bca9a8, 0x18d0, 0x4a18,
+		    0x8e, 0x7b, 0x26, 0x39, 0x21, 0x98, 0x04, 0xb7);
+
+DECLARE_TR_CTX(gain_tr, SOF_UUID(gain_uuid), LOG_LEVEL_INFO);
+#endif
+
+DECLARE_TR_CTX(volume_tr, SOF_UUID(volume_uuid), LOG_LEVEL_INFO);
```
kv2019i, line 31: This is a bit iffy to use use these DECLARE_ macros in a header file. If this is ever included twice, you'll get errors.

Did you consider having these in volume_ipc3.c and volume_ipc4.c ? You'd have to duplicate the "volume_tr" duplication, but otherwise this should work. Any other downsides?

kv2019i (approved):
Thanks, looks good now @btian1 ! One minor note on the uuids , please take a look at my inline comment.

lgirdwood (approved):

src/include/sof/audio/volume_uuid.h
lgirdwood: Ack - @btian1 can we look at moving the volume headers to the volume src directory (in a new PR) as AFAICT they all should be private for volume only (except for any IPC structures).

btian1: @kv2019i , move it to ipc3.c and ipc4.c will be like my previous implementation, volume.c will refer these declaration, it is hard to refer.
 I have to create new header file to  accommodate this and make sure it only be included once.

@lgirdwood , I also want to move each header file to each folder, however:

1. all the component directory does not include headers.
2. how to resolve cmocka include? like in: 
    sof\test\cmocka\src\audio\volume\volume_process.c
    #include <sof/audio/volume.h>

if we want to move, we need take care each component, not only volume and resolve issue 2.

src/audio/module_adapter/module/volume/volume_ipc4.c
```diff
@@ -0,0 +1,419 @@
...
+	cd->peak_report_cnt = CONFIG_PEAK_METER_UPDATE_PERIOD * 1000 / mod->dev->period;
+	if (cd->peak_report_cnt == 0)
+		cd->peak_report_cnt = 1;
+#endif
+	ret = volume_params(mod);
+	if (ret < 0)
+		return ret;
+
+	return 0;
+}
```
lyakh: @btian1 just over 400 lines - looks much better to me now! Thanks for doing the work!

PR #8027, opened 2023-08-11
zephyr: cpu: init cpu if context save is not support

RanderWang:
Fix multicore test issue on cavs platforms which don't support context save now, so need to init cpu when the core boot up.


kv2019i (approved):
Good catch @RanderWang !

tmleman (approved):

btian1 (approved):

zephyr/lib/cpu.c
```diff
@@ -148,7 +148,9 @@ int cpu_enable_core(int id)
 	 * initialization. By reinitializing the idle thread, we would overwrite the kernel structs
 	 * and the idle thread stack.
 	 */
+#ifdef CONFIG_ADSP_IMR_CONTEXT_SAVE
 	if (pm_state_next_get(id)->state == PM_STATE_ACTIVE)
+#endif
```
lyakh, line 153: nitpick: would
```
if (IS_ENABLED(CONFIG_ADSP_IMR_CONTEXT_SAVE) && pm_state_next_get(id)->state == PM_STATE_ACTIVE)
```
also work?

RanderWang: NP.  change it.

aiChaoSONG:
SOFCI TEST

Issue #7759, opened 2023-06-07
[BUG] I/O error when testing multiple pause/resume, XRUN nothing to copy


keqiaozhang:
**Describe the bug**
Observed this issue in CI tests, it happens on ADL/TGL IPC4 platform. But there're no errors in dmesg or mtrace. The reproduce rate is about 50%.

**To Reproduce**
~/sof-test/test-case/multiple-pause-resume.sh -r 50

**Reproduction Rate**
50%

**mtrace**
```
[  111.337111] <err> dma_dw_common: dw_dma_get_status: xrun detected
[  111.337125] <wrn> dai_comp: dai_common_copy: comp:1 0x40001 dai_zephyr_copy(): nothing to copy
[  111.338111] <err> dma_dw_common: dw_dma_get_status: xrun detected
[  111.338146] <wrn> dai_comp: dai_common_copy: comp:1 0x40001 dai_zephyr_copy(): nothing to copy
[  111.339111] <err> dma_dw_common: dw_dma_get_status: xrun detected
[  111.339133] <wrn> dai_comp: dai_common_copy: comp:1 0x40001 dai_zephyr_copy(): nothing to copy
[  111.340113] <err> dma_dw_common: dw_dma_get_status: xrun detected
[  111.340128] <wrn> dai_comp: dai_common_copy: comp:1 0x40001 dai_zephyr_copy(): nothing to copy
[  111.341111] <err> dma_dw_common: dw_dma_get_status: xrun detected
[  111.341146] <wrn> dai_comp: dai_common_copy: comp:1 0x40001 dai_zephyr_copy(): nothing to copy
[  111.342111] <err> dma_dw_common: dw_dma_get_status: xrun detected
[  111.342133] <wrn> dai_comp: dai_common_copy: comp:1 0x40001 dai_zephyr_copy(): nothing to copy
[  111.343111] <err> dma_dw_common: dw_dma_get_status: xrun detected
[  111.343125] <wrn> dai_comp: dai_common_copy: comp:1 0x40001 dai_zephyr_copy(): nothing to copy
```

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel: {[17b5d3eee2c6](https://github.com/thesofproject/linux/commit/17b5d3eee2c6)}
    * SOF: {[1dd1939f05ef](https://github.com/thesofproject/sof/commit/1dd1939f05ef)}

[...]

PR #8029, opened 2023-08-11
ipc4: set_dx: ignoring logger status notifications

tmleman:
Before HOST can request FW to enter D3 state, has to stop all pipelines. If it's done properly all possible source of IPC messages from FW to HOST are disabled.

The only exception from this is logger. If logs are enabled FW will continue to produce new traces and eventually trigger threshold notification.

This notification is not critical so we can skip it during D3 entry. The message will remain in the internal list and will be sent when the DSP wakes up.

### DIFF
--- a/src/ipc/ipc-zephyr.c
+++ b/src/ipc/ipc-zephyr.c
@@ -35,6 +35,8 @@
 #include <rtos/task.h>
 #include <rtos/spinlock.h>
 #include <ipc/header.h>
+#include <sof/ipc/msg.h>
+#include <ipc4/notification.h>
 
 #include <stdbool.h>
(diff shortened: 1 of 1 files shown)


kv2019i (approved):
Thanks @tmleman !

btian1 (approved):

RanderWang (approved):

kv2019i:
We seem to have a another build issue with Zephyr main that needs to be followed-up -> not related to this PR though, so can ignore the Zephyr main fails in CI. We hit the known #7759 (under debug, unrelated) and alsabat fail on the main-ace plan, but this looks to be a DUT issue. Proceeding with merge.

PR #7547, opened 2023-05-02
Tools: Topology2: Add SRC component to cavs-nocodec.conf topologies

singalsu:
This patch adds 8 - 192 kHz playback and capture SRC to nocodec topologies such sof-tgl-nocodec.tplg. It enables testing of SRC component with IPC4.

PR #8032, opened 2023-08-14
src: bugfix: rate for sink not set

marcinszkudlinski:
Rate for sink was not set properly in src.c

### DIFF
--- a/src/audio/src/src.c
+++ b/src/audio/src/src.c
@@ -576,6 +576,7 @@ static void src_set_sink_params(struct comp_dev *dev, struct sof_sink __sparse_c
 	sink_set_valid_fmt(sink, valid_fmt);
 	sink_set_channels(sink, cd->ipc_config.base.audio_fmt.channels_count);
 	sink_set_buffer_fmt(sink, cd->ipc_config.base.audio_fmt.interleaving_style);
+	sink_set_rate(sink, cd->ipc_config.sink_rate);
 }
 
 #elif CONFIG_IPC_MAJOR_3


singalsu:
Thanks, to be sure I'll give this a try with #7547 . We don't have in CI test for this.

kv2019i (approved):
Looks good. I'll wait @singalsu for your ack on this.

src/audio/src/src.c
```diff
@@ -576,6 +576,7 @@ static void src_set_sink_params(struct comp_dev *dev, struct sof_sink __sparse_c
 	sink_set_valid_fmt(sink, valid_fmt);
 	sink_set_channels(sink, cd->ipc_config.base.audio_fmt.channels_count);
 	sink_set_buffer_fmt(sink, cd->ipc_config.base.audio_fmt.interleaving_style);
+	sink_set_rate(sink, cd->ipc_config.sink_rate);
```
btian1, line 579: possible to have a common function like: ipc4_update_buffer_format to cover all set in one function then can be used for all modules?

singalsu:
Unfortunately this is not fixing the regression that has happened.  @marcinszkudlinski Is there a bug that you see that this fixes? 

Probably OK to merge this but some other fixes are needed too.

singalsu:
I wonder if every component should set both source and sink params from init IPC? E.g. Aria sets both in https://github.com/thesofproject/sof/blob/8f6018057b64441f26de6d509ac16514d7347f12/src/audio/aria/aria.c#L182

What is the recommendation for IPC4 modules @marcinszkudlinski  ?

I'm getting with this patch a divide by zero exception in source_get_data_frames_available(). Apparently source_get_frame_bytes(source) returns zero.

src/audio/src/src.c
lgirdwood: that would need to be done at the module layer. 1st step needs to be aliging all module to use the module API, 2nd step would be to then inpect each module to see where we can move repeating logic from modules to core.

lgirdwood (approved):

lgirdwood:
@wszypelt @lrudyX not expecting this to fail unless CI is making a different assumption to SRC for rates ?

lrudyX:
> @wszypelt @lrudyX not expecting this to fail unless CI is making a different assumption to SRC for rates ?

We have fail with LNL FPGA. Fails only with this platform.

btian1 (approved):

lgirdwood:
@marcinszkudlinski rerunning CI to rule out a false positive result.

kv2019i:
Proceeding with merge, only known fails in CI report.

PR #8033, opened 2023-08-14
audio: xtensa: add Kconfig for inline functions

btian1:
Enable inline function for cavs 2.5 and above platform, this is a follow up with issue:#5212

### DIFF
--- a/src/arch/xtensa/CMakeLists.txt
+++ b/src/arch/xtensa/CMakeLists.txt
@@ -98,7 +98,11 @@ else()
 endif()
 
 # C & ASM flags
-target_compile_options(sof_options INTERFACE ${stdlib_flag} -fno-inline-functions ${XTENSA_C_ASM_FLAGS})
+if(CONFIG_COMPILER_INLINE_FUNCTION_OPTION)
+	target_compile_options(sof_options INTERFACE ${stdlib_flag} ${XTENSA_C_ASM_FLAGS})
+else()
+	target_compile_options(sof_options INTERFACE ${stdlib_flag} -fno-inline-functions ${XTENSA_C_ASM_FLAGS})
+endif()
 
 # C flags
 # TODO: Generator expressions are supported only with Make and Ninja,
--- a/Kconfig.sof
+++ b/Kconfig.sof
@@ -114,6 +114,13 @@ config COMPILER_WORKAROUND_CACHE_ATTR
 	  There is a weird optimization bug with gcc10x and gcc8.1
 	  (with -O2 flags) on IMX platforms. See PR #4605.
 
+config COMPILER_INLINE_FUNCTION_OPTION
+        bool
+	default y if CAVS_VERSION_2_5 || ACE
+        default n
+        help
+          When enabled, -fno-inline-function option is not passed to compiler
+
 rsource "src/Kconfig"
 
 # See zephyr/modules/Kconfig
--- a/src/audio/dai-zephyr.c
+++ b/src/audio/dai-zephyr.c
@@ -555,6 +555,8 @@ static int dai_verify_params(struct dai_data *dd, struct comp_dev *dev,
 	struct sof_ipc_stream_params hw_params;
 	int ret;
 
+	memset(&hw_params, 0, sizeof(hw_params));
+
 	ret = dai_common_get_hw_params(dd, dev, &hw_params, params->direction);
 	if (ret < 0) {
 		comp_err(dev, "dai_verify_params(): dai_verify_params failed ret %d", ret);
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -799,8 +799,9 @@ set(VERSION_H_PATH ${GENERATED_DIRECTORY}/sof_versions.h)
 find_package(Python3 COMPONENTS Interpreter)
 set(PYTHON3 "${Python3_EXECUTABLE}")
 
-# We don't know why we have this https://github.com/thesofproject/sof/issues/5212
+if (NOT CONFIG_COMPILER_INLINE_FUNCTION_OPTION)
 target_compile_options(SOF INTERFACE -fno-inline-functions)
+endif()
 
 # SOF needs `typeof`, `__VA_ARGS__` and maybe other GNU C99
 # extensions. TODO other flags required ?


zephyr/Kconfig
```diff
@@ -56,4 +56,13 @@ config ZEPHYR_DP_SCHEDULER
 	  DP modules can be located in dieffrent cores than LL pipeline modules, may have
 	  different tick (i.e. 300ms for speech reccognition, etc.)
 
+config INLINE_FUNCTIONS
+	bool
+	default y if CAVS_VERSION_2_5 || ACE
+	default n
+	depends on CAVS_VERSION_2_5
+	depends on ACE
```
lyakh, line 64: I don't think you need these two dependencies

lyakh (commented):
Nice! Any size / performance comparison data yet?

zephyr/Kconfig
btian1: I added based on previous config, like ZEPHYR_DP_SCHEDULER, so not 100% sure, maybe you are right.

btian1 (commented):
please check #5212 , I pasted data for the compare.

zephyr/Kconfig
lyakh: `ZEPHYR_DP_SCHEDULER` is user-selectable - it has a comment after the type field, and its "default y if X" and "depends on Y" are different:
```
config ZEPHYR_DP_SCHEDULER
	bool "use Zephyr thread based DP scheduler"
	default y if ACE
	default n
	depends on IPC_MAJOR_4
	depends on ZEPHYR_SOF_MODULE
```
Whereas here it's strange that the same config options are used in "default y if ..." and in "depends on ..." statements. In principle it makes sense: (1) it sets a default, and (2) it prevents overrides to "y" by other platforms, but is that needed? Also, don't those two dependencies mean, that it depends on both enabled at the same time, which is never the case? I would just remove those dependencies - if somebody wants it, I don't see why they shouldn't be able to try it.

lyakh:
> please check #5212 , I pasted data for the compare.

interesting... So mixout performance improves by 10.5% with inlining, mixin by 3.5%, whereas gain becomes 3.4% slower (is 0x60000 vs. 0x60001 - playback vs. capture?) While size increased by about 0.2%. Looks like a clear win to me! :-)

lyakh (approved):

zephyr/Kconfig
```diff
@@ -60,8 +60,6 @@ config INLINE_FUNCTIONS
 	bool
 	default y if CAVS_VERSION_2_5 || ACE
 	default n
-	depends on CAVS_VERSION_2_5
```
iuliana-prodan: This change should be squashed in previous commit (no 2 commit).

btian1: sorry, should rebase instead of adding.

lgirdwood (approved):

zephyr/Kconfig
lgirdwood: @iuliana-prodan probably worth checking your platforms and enabling when tests pass, IIRC this was an old XCC issue so probably good for all modern targets.

src/arch/xtensa/Kconfig
```diff
@@ -17,3 +17,9 @@ config WAKEUP_HOOK
 	  This config should be selected by other platform-level configs.
 	  Platforms that use it, have to implement hook function
 	  platform_interrupt_on_wakeup.
+
+config INLINE_FUNCTIONS
+        bool
+        default n
+        help
+          Indicate whether no_inine_functions are added into build option.
```
kv2019i, line 25: typo: s/no_ininie/no_inline/

I think both the config option name and description should be improved. We are not enabling/disable inline functions here, but rather a specific compiler option "-finline-functions".

I'd suggest INLINE_CONSIDER_ALL_FUNCTIONS as the config option name and explain in description this is related to "-finline-functions" compiler option and whether it is enabled/disabled.

src/arch/xtensa/CMakeLists.txt
```diff
@@ -98,8 +98,11 @@ else()
 endif()
 
 # C & ASM flags
-target_compile_options(sof_options INTERFACE ${stdlib_flag} -fno-inline-functions ${XTENSA_C_ASM_FLAGS})
-
+if(CONFIG_INLINE_FUNCTIONS)
+	target_compile_options(sof_options INTERFACE ${stdlib_flag} ${XTENSA_C_ASM_FLAGS})
```
kv2019i, line 102: Minor: when -finline-functions is not passed (like we do here), then this optimization is controlled by the optimization level (-O), so this is a bit misleading. But I'm ok to go like this as this is aligned with previous practise in SOF.

kv2019i (changes requested):
Excellent data on the impact. I have some comments on the naming and documentation, see inline.

zephyr/Kconfig
iuliana-prodan: yes, definitely - I'll keep an eye on this :)

lgirdwood:
> @btian1 re-running CI, not expecting a logic error here but could be timing. https://sof-ci.01.org/sofpr/PR8033/build11621/devicetest/index.html?model=TGLU_RVP_NOCODEC_IPC4ZPH&testcase=check-capture-10sec

ok, looks good now.

src/arch/xtensa/CMakeLists.txt
btian1: exactly, we did not add -finline_functions option,  since our current target is only for xcc, and sof will use xcc with -o3, it will enable inline_functions automatically, in future, if sof used for other platforms(like IA), this logic still works right? compiler should have this sense.

src/arch/xtensa/Kconfig
lyakh: "consider" sounds a bit like we'd be using an AI for this decision making and even then it would be an exaggeration :-) Maybe something along the lines "automatic inlining?" Or "compiler inline" or "compiler inline optimization" etc. Maybe indeed verbatim `COMPILER_INLINE_FUNCTIONS`.

btian1:
SOFCI TEST

iuliana-prodan (approved):

src/arch/xtensa/Kconfig
lgirdwood: Lets just say exactly what this is. i.e.
This options add -no_inline_functions to the compiler command line. Some old compilers may not emit correct code
when enabled.

dbaluta:
> interesting... So mixout performance improves by 10.5% with inlining, mixin by 3.5%, whereas gain becomes 3.4% slower (is 0x60000 vs. 0x60001 - playback vs. capture?) While size increased by about 0.2%. Looks like a clear win to me! :-)

Can you share the tests with us? How did you measured the impact of the patch at runtime. 

We would like to check this on imx. Also, what toolchain have you used? GCC or XCC. Zephyr or XTOS?

src/arch/xtensa/CMakeLists.txt
```diff
@@ -98,8 +98,11 @@ else()
 endif()
```
kv2019i, line 98: typo still in commit: "option noline_functions"

src/arch/xtensa/Kconfig
```diff
@@ -17,3 +17,9 @@ config WAKEUP_HOOK
 	  This config should be selected by other platform-level configs.
 	  Platforms that use it, have to implement hook function
 	  platform_interrupt_on_wakeup.
+
+config INLINE_CONSIDER_ALL_FUNCTIONS
+        bool
+        default n
+        help
+          Indicate whether noinine_functions are added into build option.
```
kv2019i, line 25: I'd describe exactly what this does: "When enabled, -fno-inline-function option is not passed to compiler"

zephyr/Kconfig
```diff
@@ -56,4 +56,11 @@ config ZEPHYR_DP_SCHEDULER
 	  DP modules can be located in dieffrent cores than LL pipeline modules, may have
 	  different tick (i.e. 300ms for speech reccognition, etc.)
 
+config INLINE_CONSIDER_ALL_FUNCTIONS
+	bool
+	default y if CAVS_VERSION_2_5 || ACE
+	default n
+	help
+	  Indicate whether noinine_functions are added into build option.
+
```
kv2019i, line 65: Sorry, missed this in first review, but please don't define this twice if this is used both for XTOS and Zephyr, but rather put at top-level Kconfig.sof. See e.g. option COMPILER_WORKAROUND_CACHE_ATTR . You could put this next to that one.

kv2019i (changes requested):
Few more nits still, the descsription could be even more accurate/concrete and I'd like the Kconfig definition to be done only once in the tree.

btian1:
> > interesting... So mixout performance improves by 10.5% with inlining, mixin by 3.5%, whereas gain becomes 3.4% slower (is 0x60000 vs. 0x60001 - playback vs. capture?) While size increased by about 0.2%. Looks like a clear win to me! :-)
> 
> Can you share the tests with us? How did you measured the impact of the patch at runtime.
> 
> We would like to check this on imx. Also, what toolchain have you used? GCC or XCC. Zephyr or XTOS?

please add -o app/perf_overlay.conf into fw build, then you will get trace logs with component cycle performance.

btian1:
SOFCI TEST

kv2019i (approved):
Thanks @btian1, looks good now! Great to have this long-open issue closed.

PR #8034, opened 2023-08-14
app: enable OUTPUT_DISASSEMBLY for Zephyr builds

kv2019i:
With Zephyr commit cc5763344709 ("Build system: disable `OUTPUT_DISASSEMBLY` by default"), generation of zephyr.lst is disabled by default.

Enable disassembly generation in SOF configuration as it is expected by SOF builds rules for reproducible builds.


app/prj.conf
```diff
@@ -17,6 +17,9 @@ CONFIG_BUILD_OUTPUT_STRIPPED=y
 # default value.
 CONFIG_CLEANUP_INTERMEDIATE_FILES=n
 
+# Generate zephyr.lst file
+CONFIG_OUTPUT_DISASSEMBLY=y
```
btian1, line 21: what's the benefit of output lst file during build? who will use it? it will somehow make build slower.

btian1 (approved):

aiChaoSONG (approved):

app/prj.conf
aiChaoSONG: xtensa-build-zephyr.py will package this file to a compressed file for debug, this file is very useful for debug.  It seems without this file, xtensa-build-zephyr.py will complain.

kv2019i:
Zephyr mainline tests pass again, proceeding with merge.

greg-intel:
I ran the zephyr build job, it had been failing in two to three minutes, and has now passed the five minute mark, so it looks like this PR has fixed the issue, thanks!

btian1:
image

will it fix above CI issue?

PR #8036, opened 2023-08-15
smart_amp: revamp to two-layer modular design structure

johnylin76:
The motivation is depicted in https://github.com/thesofproject/sof/issues/7801

This PR revamps smart_amp component design to two-layer structure, i.e. generic layer and inner model layer.

Generic layer is the common part of smart amp process which can be regarded as the glue code interacting between SOF component ops and inner model. While inner model may have various implementations respectively for solution suppliers in a modular way.

This PR revamps Smart Amplifier component design to a two-layer structured design, i.e. generic layer and inner model layer. The latter can have various implementations respectively for Amplifier solution suppliers, while the former is the common part of Smart Amp process adaptable for all solutions.

In structural aspect, one can regard generic layer as the glue code that wraps inner model in a SOF component. Ops are defined for interaction between two layers. Inner model is the solution-specific modular code, which may have static libraries linked as needed. The structure is figured below:
 
```
                                  SRC(FF)   SINK(OUT)  +-SRC(FB)  bytectl
  +- SMART_AMP         |^ comp ops    |         ^      |           ^|
  | +------------------v|-------------v---------|------v-----------|v---------+
  | | Generic Layer                 (chan remap/fmt conv)          ||         |
  | |    (memory mgr)--+------> :::::::::BUFFERS:::::::::::::      |+> CONFIG |
  | +------------------|-|^-----------|---------^------|-----------^|---------+
  |                    | || mod ops   |         |      |           ||
  | +------------------v-v|-----------v---------|------v-----------|v---------+
  | | Inner Model   :::::::::::::::::::::BUFFERS:::::::::::::::::::::::       |
  | |                  (solution-specific impl/wrapper)            |+> MODEL  |
  | +------------------------------|^------------------------------^----------+
  +---                             v| lib ops                      | CALDATA
                             Static Libs (as needed)     ----------+
  Note:
  - FF(Feed-Forward): un-processed playback frame source
  - FB(Feedback): feedback reference frame source (from the capture pipeline)
```
As illustrated above, generic layer handles the cross-communication for inner model and SOF pipeline flow, as well as the smart-amp common tasks including:
1. Channel remapping for input/output frames
2. Frame format conversion for input/output frames. It allows inner model to work with different format from SOF audio stream. (Now it only allows the precision of inner model format >= SOF stream, e.g. inner model: S32_LE, SOF stream: S16_LE)
3. Runtime memory full-management. The dynamic memory required by inner model will be allocated/owned/released by generic layer.


src/audio/smart_amp/smart_amp.c
```diff
@@ -646,25 +629,95 @@ static int smart_amp_copy(struct comp_dev *dev)
...
+			break;
+		}
+	}
+	if (i == num_fb_fmts) {
+		comp_err(dev, "smart_amp_set_frame_fmts(): no supported feedback fmt");
+		return -EINVAL;
+	}
+
+	/* set frame formats to inner model */
+	ret = mod->mod_ops->set_fmts(mod, sad->ff_mod.frame_fmt, sad->fb_mod.frame_fmt);
```
lgirdwood, line 698: I'm guessing we are missing a ```module_set_fmt()``` wrapper fo this in module core ?

src/audio/smart_amp/smart_amp_generic.c
```diff
@@ -9,194 +9,292 @@
...
+				*mod_ptr = 0;
+			} else {
+				src_idx = src_frag + chan_map[ch];
+				src_ptr = audio_stream_read_frag_s32(src, src_idx);
+				*mod_ptr = *src_ptr;
+			}
+			mod_ptr++;
+		}
+		src_frag += src_ch;
+	}
```
lgirdwood, line 35: @singalsu @andrula-song is this using the optimal method now ?

singalsu: It would be better to write with frames without wrap helper to avoid audio_stream_read_frag_s32() usage for every sample read.

src/audio/smart_amp/smart_amp_generic.c
```diff
@@ -9,194 +9,292 @@
...
+	int32_t *mod_ptr = (int32_t *)src_mod->buf.data;
+
+	for (fr = 0 ; fr < frames ; fr++) {
+		for (ch = 0 ; ch < src_mod->channels; ch++) {
+			if (chan_map[ch] == -1) {
+				*mod_ptr = 0;
+			} else {
+				src_idx = src_frag + chan_map[ch];
+				src_ptr = audio_stream_read_frag_s32(src, src_idx);
+				*mod_ptr = *src_ptr << 8;
```
andrula-song, line 63: here too, use audio_stream_read_frag_s32(), there is an example in aria_algo_get_data() in aria_generic.c

src/audio/smart_amp/smart_amp_generic.c
btian1: the most waste part is if else, take 48 frame, 2 channel as example, if need 96 check for if/else, this is big cost.
better to use channel in first loop, and then check, chan_map[ch], if == -1, with one loop, else with another loop, like below:

	for (ch = 0 ; ch < src_mod->channels; ch++) {
		if (chan_map[ch] == -1) {
			for (fr = 0 ; fr < frames ; fr++) {
				*mod_ptr = 0;
				mod_ptr += src_mod->channels;
			}
		} else {
			for (fr = 0 ; fr < frames ; fr++) {
				src_idx = src_frag + chan_map[ch];
				src_ptr = audio_stream_read_frag_s16(src, src_idx);
				*mod_ptr = *src_ptr;
                                mod_ptr += src_mod->channels;
			}
			src_frag += src_ch;
		}
	}


this will have only 2 if/else, will save cycles a lot.
Note: this is just reference code, please have more investigation to get correct offset calculation.

lyakh: @btian1 I don't think that's equivalent to as the

johnylin76: Thanks for the advice. I will do the follow-up optimization

btian1: it is just for demo purpose for save if/else, as I said in the note, "need more investigation to get correct offset"

cujomalainey:
@lgirdwood how much power (and would it be even worth it) if userspace opened the PCM in non-blocking so we can just never yield samples to userspace allowing both CRAS and the DMA to sit idle?

src/audio/smart_amp/smart_amp.c
johnylin76: I'm not sure I get the idea. Do you mean we need the ops `set_fmt()` to set the format of core?

The design now has ops `set_fmts()` to set the format of feed-forward (FF) data path and feedback (FB) path respectively. Indeed the core format can be independent of input/ouput data format, however I wonder if that has the real need (in fact I think FF and FB formats could align each other (as well as core format) then `set_fmts()` could be replaced with `set_fmt()`). Since the design now is proposed to handle input/output data format conversion in generic layer, it should be fair to require the format alignment.

lgirdwood: Sorry, I meant an inline static wrapper func around the API ops dereference.

lgirdwood:
> @lgirdwood how much power (and would it be even worth it) if userspace opened the PCM in non-blocking so we can just never yield samples to userspace allowing both CRAS and the DMA to sit idle?

@cujomalainey not sure - ideally if we do nothing then we should not consume power. However, this proposal may enable DMA channels and maybe other IPs in preparation for stream propagation (and consume some power, but I would think less than the full stream). 

I assume you want to minimize the number of PCMs for smart amp when not needed ?I think @ranj063 has recently been looking at a way to get rid of the second PCM when its not needed (IIRC for AEC?).

src/audio/smart_amp/smart_amp_passthru.c
```diff
@@ -0,0 +1,156 @@
...
+
+/* self-declared inner model data struct */
+struct passthru_mod_data {
+	struct smart_amp_mod_data_base base;
+	uint16_t ff_fmt;
+	uint16_t fb_fmt;
+};
+
+const int ff_fmt_count = 3;
+const uint16_t supported_ff_fmts[] = {
```
lgirdwood, line 27: static ?

src/include/sof/audio/smart_amp/smart_amp.h
```diff
@@ -11,180 +11,300 @@
...
+ * Refer to "struct smart_amp_mod_data_base" for details of declaration.
+ * Args:
+ *    dev - the comp_dev object for logger tracing use only.
+ * Returns:
+ *    The pointer of "smart_amp_mod_data_base" which is the leading member in
+ *    the created model data struct.
+ */
+struct smart_amp_mod_data_base *mod_data_create(const struct comp_dev *dev);
+
+struct inner_model_ops {
```
lgirdwood, line 174: I assume all ops are mandatory ? We probably want to say in comments so developers know if any are optional.

lgirdwood (commented):
@johnylin76 This looks ready to land, I think we just need the optimal copy() code and a few minor comments and we are good.

src/audio/smart_amp/Kconfig
```diff
@@ -1,19 +1,39 @@
...
+if COMP_SMART_AMP
+
+choice
+        prompt "Smart Amplifier solution applied"
+        default PASSTHRU_AMP
+	help
+	  The selection for Smart Amplifier component implementation
+	  will depend on the Amplifier solution supplier. It is fair
+	  to treat the supported solutions as mutually exclusive ones.
+	  There should be no more than one solution selected per build
```
cujomalainey, line 12: Why can't we setup a system similar to modules and have support for all devices at once? You could also do composition and just have this by a stand in for default implementation in a module?

src/audio/smart_amp/smart_amp_generic.c
```diff
@@ -9,194 +9,292 @@
...
+{
+	int fr, ch;
+	int src_frag = 0;
+	int src_ch = audio_stream_get_channels(src);
+	int src_idx;
+	int32_t *src_ptr;
+	int32_t *mod_ptr = (int32_t *)src_mod->buf.data;
+
+	for (fr = 0 ; fr < frames ; fr++) {
+		for (ch = 0 ; ch < src_mod->channels; ch++) {
```
lyakh, line 24: no space before `;` - everywhere, please

src/include/sof/audio/smart_amp/smart_amp.h
johnylin76: Yes they are all mandatory. I will add to the comment.

src/audio/smart_amp/Kconfig
johnylin76: We are able to do it for sure. The code implementation will be a bit more complicated (as how module_adapter is implemented https://github.com/thesofproject/sof/blob/main/src/include/sof/audio/module_adapter/module/generic.h

I didn't implement that way due to the proposed plan in #7801 said the long-term goal is to use module API for all components, so I would defer that to the future task for converting smart_amp to use module API.

src/audio/smart_amp/smart_amp.c
```diff
@@ -646,25 +629,95 @@ static int smart_amp_copy(struct comp_dev *dev)
...
+	ret = mod->mod_ops->get_supported_fmts(mod, &ff_fmts, &num_ff_fmts,
+					       &fb_fmts, &num_fb_fmts);
+	if (ret) {
+		comp_err(dev, "smart_amp_set_frame_fmts(): failed to get supported formats");
+		return ret;
+	}
+
+	/* available formats: SOF_IPC_FRAME_S16_LE, SOF_IPC_FRAME_S24_4LE, SOF_IPC_FRAME_S32_LE */
+	for (i = 0; i < num_ff_fmts; i++) {
+		if (ff_fmts[i] >= ff_src_fmt && ff_fmts[i] <= SOF_IPC_FRAME_S32_LE) {
```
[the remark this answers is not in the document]
johnylin76: yep I agree with that. The comparison will be easy for audio_format in IPC4 by comparing the depth attribute.
The design now (in IPC3 format) relies on the background knowledge that the enum value is the same order as the depth among {S16_LE=0, S24_4LE=1, S32_LE=2}, which is implicit.

To elaborate the current design, I would propose to wrap the implicit comparison into helper functions with comments for the background. Or maybe add get_depth(fmt) helper and compare by get_depth(fmt). get_depth(fmt) can also help smart_amp_generic implementation where we need to specify the shift bit for each conversion pair, e.g. S16-to-S24, S16-to-S32.

src/audio/smart_amp/smart_amp.c
johnylin76: I see. Anyway it triggered my re-consideration of the format for module core. I revised to unify the module formats which include core, ff-input and fb-input previously. Now the module ops will only set one format .

src/audio/smart_amp/smart_amp_generic.c
johnylin76: just updated the commit for better buffer read/write strategy.

src/audio/smart_amp/smart_amp.c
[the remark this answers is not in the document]
johnylin76: updated the commit. I added a helper function get_sample_bitdepth() in include/sof/audio/format.h. I'm not sure if it's appropriate, we can also put it inside of smaty_amp if that is the case. WDYT?

src/audio/smart_amp/smart_amp.c
```diff
@@ -676,91 +729,64 @@ static int smart_amp_prepare(struct comp_dev *dev)
...
-		bitwidth = 32;
-		break;
-	default:
-		comp_err(dev, "[DSM] smart_amp_process() error: not supported frame format %d",
-			 audio_stream_get_frm_fmt(&source_c->stream));
-		goto error;
+		buffer_release(buf_c);
+	} else {
+		/* should not be used anyway */
+		fb_src_fmt = 0;
```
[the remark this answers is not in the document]
johnylin76: just revised the code to prevent the redundant code like this.

kv2019i (commented):
My first-pass review, a few comments inline but this seems solid. With 2.8 opened for development, this would be a good time to get bigger changes like this in.  We do have a couple of big changes ongoing (one is the move to module adapter) and now with 2.8 we will revamp the buffer structure (https://github.com/thesofproject/sof/pull/8106 ), so some coordination is needed to avoid too many rebase hassles.

src/audio/smart_amp/smart_amp.c
```diff
@@ -1,8 +1,10 @@
 // SPDX-License-Identifier: BSD-3-Clause
 //
-// Copyright(c) 2020 Maxim Integrated All rights reserved.
+// Copyright(c) 2023 Google LLC.
 //
-// Author: Ryan Lee <ryans.lee@maximintegrated.com>
+// Author: Pin-chih Lin <johnylin@google.com>
```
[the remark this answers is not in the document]
johnylin76: The strategy for this revamp is to re-design the smart amp component while moving all the model-specific codes (of Maxim DSM) into smart_amp_maxim_dsm.c.

On that account I preserved the copyright of Maxim in smart_amp_maxim_dsm.c, and replaced with mine in smart_amp.c.

Either way is fine to me I just wasn't sure how to accurately make the statement. I cannot find any example of having more than one copyright declaimer in one file.

Does it look good to you with the disclaimer rephrased like the following?

```
//
// Copyright(c) 2020 Maxim Integrated All rights reserved.
//
// Author: Ryan Lee <ryans.lee@maximintegrated.com>
//
// Copyright(c) 2023 Google LLC.
//
// Author: Pin-chih Lin <johnylin@google.com>
```

Thanks

src/audio/smart_amp/smart_amp.c
```diff
@@ -533,45 +484,72 @@ static int smart_amp_trigger(struct comp_dev *dev, int cmd)
...
-	case SOF_IPC_FRAME_S16_LE:
-	case SOF_IPC_FRAME_S24_4LE:
-	case SOF_IPC_FRAME_S32_LE:
-		return smart_amp_process;
-	default:
-		comp_err(dev, "smart_amp_process() error: not supported frame format");
-		return NULL;
+	if (frames == 0) {
+		comp_warn(dev, "smart_amp_copy(): feedback frame size zero warning.");
+		return 0;
```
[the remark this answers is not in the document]
johnylin76: this warning message is as it was actually (just moved here due to the revamp.

Based on my previous debugging experience, this can be reproduced every now and then with no harm (well... at least as far as I've met), but is not an expected behavior.

src/audio/smart_amp/smart_amp_maxim_dsm.c
```diff
@@ -60,8 +176,62 @@ static int maxim_dsm_init(struct smart_amp_mod_struct_t *hspk, struct comp_dev *
...
+		return 0;
+
+	return MIN(DSM_DEFAULT_MAX_NUM_PARAM, cmdblock[DSM_GET_CH1_IDX]);
+}
+
+static int maxim_dsm_get_handle_size(struct smart_amp_mod_struct_t *hspk)
+{
+	enum DSM_API_MESSAGE retcode;
+	struct dsm_api_memory_size_ext_t memsize;
+	int *circularbuffersize = hspk->circularbuffersize;
```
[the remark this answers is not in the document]
johnylin76: this is the same as the original code (authored by Maxim) before the revamp (just moved from elsewhere). So I would keep it as is. And I think it is probably named that way on purposes to be aligned with the pre-built libraries for Maxim

src/audio/smart_amp/smart_amp.c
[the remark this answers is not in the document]
kv2019i: You are right @johnylin76 , strangely enough we don't really seem to have any files with multiple copyright statements. So most contributors are fine to track changes with git log. But right, multiple copyright lines is fine (refer to Linux and Zephyr sourcebases where this is more common), just add it on the side like in your example and we are good.

I'm ok with the original as well, but removing a copyright line is quite unique (also in Linux, Zephyr), so we'd need Ryan from Maxim to ack the patch. Easier if you just add on the side.

kv2019i:
@johnylin76 if you have time now to do a rebase after https://github.com/thesofproject/sof/pull/8217 is merged, we could now push this in.

lgirdwood:
> @johnylin76 if you have time now to do a rebase after #8217 is merged, we could now push this in.

@johnylin76 #827 now merged, good to go ?

src/audio/smart_amp/smart_amp.c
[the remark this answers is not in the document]
johnylin76: Thanks. I should also cc this PR to Ryan

src/audio/smart_amp/smart_amp.c
```diff
@@ -14,200 +16,129 @@
...
 
-	if (!caldata->data) {
-		comp_err(dev, "smart_amp_alloc_caldata(): model->data rballoc failed");
-		return -ENOMEM;
-	}
+	/* sof -> mod feedback data re-mapping and format conversion */
+	size = SMART_AMP_FB_BUF_DB_SZ * sizeof(int32_t);
+	ret = smart_amp_buf_alloc(&sad->fb_mod.buf, size);
+	if (ret < 0)
+		return ret;
```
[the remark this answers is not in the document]
johnylin76: Thanks. I updated the commit as the suggested by @lyakh 
https://github.com/thesofproject/sof/pull/8036/files#diff-4a0050312896c64976b6e80b0d7fc01348b8457d831698766e49900c05ee85ffR68

johnylin76:
> > @johnylin76 if you have time now to do a rebase after #8217 is merged, we could now push this in.
> 
> @johnylin76 #827 now merged, good to go ?

Thanks. I just addressed the 2nd-round comments and rebased the commit.

kv2019i (approved):
Thanks @johnylin76 , looks good now!

lgirdwood (approved):

PR #8037, opened 2023-08-16
audio: volume: Fix compilation error when COM_PEAK_VOL is not selected

ranj063:
CONFIG_PEAK_METER_UPDATE_PERIOD depends on CONFIG_COMP_PEAK_VOL

### DIFF
--- a/src/audio/module_adapter/module/volume/volume.c
+++ b/src/audio/module_adapter/module/volume/volume.c
@@ -1294,10 +1294,11 @@ static int volume_prepare(struct processing_module *mod,
 
 #if CONFIG_IPC_MAJOR_4
 	cd->peak_cnt = 0;
+#if CONFIG_COMP_PEAK_VOL
 	cd->peak_report_cnt = CONFIG_PEAK_METER_UPDATE_PERIOD * 1000 / mod->dev->period;
 	if (cd->peak_report_cnt == 0)
 		cd->peak_report_cnt = 1;
-
+#endif
 	ret = volume_params(mod);
 	if (ret < 0)
 		return ret;


src/audio/module_adapter/module/volume/volume.c
```diff
@@ -1294,7 +1294,9 @@ static int volume_prepare(struct processing_module *mod,
 
 #if CONFIG_IPC_MAJOR_4
 	cd->peak_cnt = 0;
+#if CONFIG_COMP_PEAK_VOL
 	cd->peak_report_cnt = CONFIG_PEAK_METER_UPDATE_PERIOD * 1000 / mod->dev->period;
+#endif
```
btian1, line 1299: #endif should goto line 1302?

ranj063: I'm not sure about this. Maybe all of peak_cnt reporting should go within the #if. But this is enough for now to fix the compilation error when the config isn't enabled

kv2019i: I'd say 1302 needs to be included as well.

s/COM/COMP/ typo also in cmmit title

andrula-song:
then why not fix the definition of struct vol_data in volume.h as well?

ranj063:
> then why not fix the definition of struct vol_data in volume.h as well? line 145 should be "#if CONFIG_COMP_PEAK_VOL" too.

@andrula-song I think those are fine to be left in. THey will be unused if PEAK_VOL isnt selected

btian1 (approved):

lgirdwood (approved):

lgirdwood:
@wszypelt @lrudyX think this may be a false positove as this is fixing a compilation error with volume. Good to merge ?

wszypelt:
@lgirdwood good to merge :)

PR #5047, opened 2021-12-02
Ipc4: container size support in format conversion

RanderWang:
Ipc4 stream have container size and valid sample size.
    In windows system, container size is normally 32bit
    and valid sample size is 16bit or 24bit rely on system
    setting. Currently host and dai component are included
    by copier module to manage gateway. These two components
    doesn't support container size and all the code is designed
    for the case that container size is always equal to valid
    sample size.

With this patch HDA playback works with default windows audio player in format of 16bit/48000HZ/2CH.  24bit will be added later and it is a little tricky since it is really S24_3LE

PR #8039, opened 2023-08-16
Multicore DP processing using DpQueue

marcinszkudlinski:
dependencies:
- [ ] github.com/thesofproject/sof/pull/8002
- [ ] changes described in this RFC: https://github.com/thesofproject/sof/issues/8006
- [ ] https://github.com/thesofproject/sof/pull/8020

DP component need to work flawlessly on different cores
than the main pipeline processing data independently
from the modules before and after them in the pipeline

To achieve this, a cross core producer-consumer safe
dp_queue should be used

DP queue can only be connected to modules that use
sink/src interface, so DP modules need to use it
obligatory.

To connect dp_queue into modules chain, double buffering
method is used:
 - in LL task DP module is processed as an LL module, but the
   copy method is copying data from/to audio_streams
   to/from dp_queues
 - the main DP module processing takes place in DP task
   (in separate Zephyr thread). The tread may be bind to
   separate core


src/include/sof/audio/module_adapter/module/generic.h
```diff
@@ -199,10 +200,21 @@ struct processing_module {
...
+			struct output_stream_buffer *output_buffers;
+			uint32_t num_input_buffers; /**< number of inputs */
+			uint32_t num_output_buffers; /**< number of outputs */
+		};
+		struct {
+			/* this is used in case of DP processing */
+			struct dp_queue *dp_queue_ll_to_dp[MODULE_MAX_SOURCES];
+			struct dp_queue *dp_queue_dp_to_ll[MODULE_MAX_SOURCES];
+		};
+	};
```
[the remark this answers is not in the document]
marcinszkudlinski: sure, 
```dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_DP)```
set in create instance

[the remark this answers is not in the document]
Comment (lgirdwood) [reply src/include/sof/audio/module_adapter/module/generic.h:217]: ok, probably worth a comment in the structure stating this to help other developers

File: src/audio/component.c [hunk 2]
```diff
@@ -328,9 +328,16 @@ int comp_copy(struct comp_dev *dev)
 
 	assert(dev->drv->ops.copy);
```
btian1, line 329: diffrent cores --> different cores in commit message

src/audio/module_adapter/module_adapter.c
```diff
@@ -1024,6 +1163,66 @@ static int module_adapter_audio_stream_type_copy(struct comp_dev *dev)
...
+static int module_adapter_copy_dp_queues(struct comp_dev *dev)
+{
+	/*
+	 * copy data from component audio streams to dp_queue
+	 * DP module processing itself will take place in DP thread
+	 * This is an adapter, to be removed when pipeline2.0 is ready
+	 */
+	struct processing_module *mod = comp_get_drvdata(dev);
+	int i = 0;
+	struct list_item *blist;
```
btian1, line 1176: int i = 0;
	struct list_item *blist;
-->
	struct list_item *blist;
	int i = 0;

src/audio/module_adapter/module_adapter.c
```diff
@@ -1024,6 +1163,66 @@ static int module_adapter_audio_stream_type_copy(struct comp_dev *dev)
...
+		struct sof_source *data_src = audio_stream_get_source(&buffer_c->stream);
+
+		struct sof_sink *data_sink = dp_queue_get_sink(mod->dp_queue_ll_to_dp[i]);
+		uint32_t to_copy = MIN(sink_get_free_size(data_sink),
+				       source_get_data_available(data_src));
+
+		source_to_sink_copy(data_src, data_sink, true, to_copy);
+
+		buffer_release(buffer_c);
+		i++;
```
btian1, line 1194: remove 1194 and change 1187 to 
struct sof_sink *data_sink = dp_queue_get_sink(mod->dp_queue_ll_to_dp[i++]);
?
also for align, i = 0; put to 1177?

src/audio/module_adapter/module_adapter.c
```diff
@@ -1436,6 +1637,24 @@ int module_adapter_reset(struct comp_dev *dev)
 		mod->num_output_buffers = 0;
 	}
 
+#if CONFIG_ZEPHYR_DP_SCHEDULER
+	if (IS_PROCESSING_MODE_SINK_SOURCE(mod) &&
+	    mod->dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_DP) {
+		for (int i = 0; i < mod->num_of_sources; i++) {
```
btian1, line 1643: remove int, i already defined in 1613

src/ipc/ipc4/helper.c
```diff
@@ -374,6 +370,19 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
+	 *	size = 2*obs of source module (obs is single buffer size)
+	 * in case of DP -> LL
+	 *	size = 2*ibs of destination (LL) module. DP queue will handle obs of DP module
+	 */
+	uint32_t buf_size;
+
+	if (source->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_LL)
+		buf_size = source_src_cfg.obs * 2;
+	else
+		buf_size = sink_src_cfg.ibs * 2;
```
btian1, line 384: does ibs and obs already 2 times of actual need? if so, do we still need *2 here?

marcinszkudlinski:
covnverted to draft as some changes required for LL connected pipelines

src/ipc/ipc4/helper.c
marcinszkudlinski: IBS and OBS are strict values that come from the IPC call
I supposed to be **exact** value that is 
 - needed for module to process 1 cycle (IBS)
 - amount of data produced by the module in 1 cycle (OBS)

2* comes from this https://github.com/thesofproject/sof/pull/5047

marcinszkudlinski:
ready to review, contains 8206  as squash commit

src/audio/module_adapter/module_adapter.c
```diff
@@ -1346,11 +1328,20 @@ int module_adapter_reset(struct comp_dev *dev)
...
+		rfree(mod->output_buffers);
+		rfree(mod->input_buffers);
 
-	mod->num_input_buffers = 0;
-	mod->num_output_buffers = 0;
+		mod->num_input_buffers = 0;
+		mod->num_output_buffers = 0;
+	}
+
+	if (IS_PROCESSING_MODE_SINK_SOURCE(mod)) {
```
kv2019i: Can this be concurrent or should this be "else if"?

src/audio/module_adapter/module_adapter.c
```diff
@@ -71,7 +73,16 @@ struct comp_dev *module_adapter_new(const struct comp_driver *drv,
 	dev->ipc_config = *config;
 	dev->drv = drv;
 
-	mod = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*mod));
+	/* allocate module information.
+	 * for DP shared modules this struct must be accessible from all cores
+	 * Unfortunalely at this point there's no information of components the module
+	 * will be binded to. So we need to allocate shared memory for each DP module
```
kv2019i, line 79: language: binded to -> bound to

src/audio/module_adapter/module_adapter.c
```diff
@@ -208,6 +225,118 @@ static int module_adapter_sink_src_prepare(struct comp_dev *dev)
...
 
+#if CONFIG_ZEPHYR_DP_SCHEDULER
+static int module_adapter_dp_queue_prepare(struct comp_dev *dev)
+{
+	struct list_item *blist;
+	struct processing_module *mod = comp_get_drvdata(dev);
+	int i;
+	int ret;
+	struct dp_queue *dp_queue;
+	int dp_mode = dev->is_shared ? DP_QUEUE_MODE_SHARED : DP_QUEUE_MODE_LOCAL;
```
kv2019i, line 236: reverse-xmas tree for declarations please

src/audio/module_adapter/module_adapter.c
```diff
@@ -208,6 +225,118 @@ static int module_adapter_sink_src_prepare(struct comp_dev *dev)
...
+	return 0;
+
+err:
+	for (int i = 0; i < mod->num_of_sources; i++)
+		if (mod->dp_queue_ll_to_dp[i]) {
+			dp_queue_free(mod->dp_queue_ll_to_dp[i]);
+			mod->dp_queue_ll_to_dp[i] = NULL;
+			mod->sources[i] = NULL;
+		}
+	mod->num_of_sources = 0;
```
kv2019i, line 322: While legal, please add {} for the for statement

src/audio/module_adapter/module_adapter.c
```diff
@@ -208,6 +225,118 @@ static int module_adapter_sink_src_prepare(struct comp_dev *dev)
...
+
+	for (int i = 0; i < mod->num_of_sinks; i++)
+		if (mod->dp_queue_dp_to_ll[i]) {
+			dp_queue_free(mod->dp_queue_dp_to_ll[i]);
+			mod->dp_queue_dp_to_ll[i] = NULL;
+			mod->sinks[i] = NULL;
+		}
```
kv2019i, line 329: Ditto here

src/include/sof/audio/component_ext.h
```diff
@@ -55,6 +55,7 @@ static inline void comp_free(struct comp_dev *dev)
 	    dev->task) {
 		schedule_task_free(dev->task);
 		rfree(dev->task);
+		dev->task = NULL;
```
kv2019i, line 58: This seems a bit out of place. Is this just a generic fix, or somehow related to the DP change?

src/ipc/ipc4/helper.c
kv2019i: Hmm, I don't get the link to #5047 ? So some entities don't understand difference between valid and container sizes, and for that we double memory size for all existing LL->LL usages?
Please tell me I'm missing something here.

src/include/sof/audio/dp_queue.h
```diff
@@ -25,9 +25,36 @@
...
+ * as we do not know who goes first - consumer or producer, especially when both are located on
+ * separate cores and EDF scheduling is used
+ *
+ *  Consumer reads 5 bytes every cycle (IBS = 5)
+ *  producer writes 5 bytes every cycle (OBS = 5)
+ *    - cycle0 consumer goes first - must wait (buf occupation = 0)
+ *		producer produce 5 bytes (buf occupation = 5)
+ *    - cycle1 producer goes first - produce 5 bytes (buf occupation = 10)
+ *		consumer consumes 5 bytes (buf occupation = 5)
+ * ===> max buf occupation = 10
```
kv2019i, line 56: Thanks. excellent documentation here!

kv2019i (commented):
Code looks good and very easy to follow. A few very minor comments to code style, and really only one bigger question on the buf_size calculation.

src/include/sof/audio/dp_queue.h
```diff
@@ -25,9 +25,36 @@
...
  *  2) Both data consumer and data producer declare max chunk sizes they want to use (IBS/OBS)
  *
- * required Buffer size:
- *	- 2*MAX(IBS,OBS) if the larger of IBS/OBS is multiplication of smaller
- *	- 3*MAX(IBS,OBS) otherwise
+ * required Buffer size is 2*MAX(IBS,OBS) to allow free read/write in various data chunk sizes
+ * and execution periods (of course in/out data rates must be same)
+ * example:
+ *  Consumer reads 3bytes each 3 cycles (IBS = 3)
+ *  producer writes 5bytes every 5 cycles (OBS = 5)
```
lyakh, line 32: in your example producer writes 3 and consumer reads 5, not as you've written. Then I think it's also more logical to start counting with 0 at cycle 0. So the first production happens after 3 (or 5) cycles

marcinszkudlinski:
(just a rebase push)

src/ipc/ipc4/helper.c
marcinszkudlinski: I simply did not want to make any changes to way that LL pipeline is working in this PR.
Double buffer size LL was introduced as a bugfix and is already in current code:

```
static struct comp_buffer *ipc4_create_buffer(struct comp_dev *src, bool is_shared,
					      uint32_t src_obs, uint32_t src_queue,
					      uint32_t dst_queue)
{
	struct sof_ipc_buffer ipc_buf;
	int buf_size;

	/* double it since obs is single buffer size */
	buf_size = src_obs * 2;
```

After my changes LL it is working in the very same way. Regardless if it is right or not - I really don't think changing it in this PR is a good idea. It may be in separate PR

For DP - I have just followed the same pattern, maybe a little too fast. I'll double check if 2* it brings any value here

src/include/sof/audio/component_ext.h
marcinszkudlinski: Just a fix, leaving dangling pointers is always a bad idea. 
You right, moving to separate commit

src/audio/module_adapter/module_adapter.c
```diff
@@ -208,6 +225,118 @@ static int module_adapter_sink_src_prepare(struct comp_dev *dev)
...
+	list_for_item(blist, &dev->bsource_list) {
+		struct comp_buffer *source_buffer =
+			container_of(blist, struct comp_buffer, sink_list);
+
+		/* copy IBS & OBS from buffer to be shadowed */
+		size_t ibs = source_get_ibs(audio_stream_get_source(&source_buffer->stream));
+		size_t obs = sink_get_obs(audio_stream_get_sink(&source_buffer->stream));
+
+		/* create a shadow dp queue */
+		dp_queue = dp_queue_create(ibs, obs, dp_mode);
```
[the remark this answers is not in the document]
marcinszkudlinski: I hope it will be possible to stop using comp_buffers soon enough. But one step at the time

1) at the moment I can't connect dpQueue to anything else than src module and... comp_buffer ;) 

2) whole pipeline mechanisms use comp_buffer as a base - iteration, binding, etc. etc. 

So as 99% of the code does have to use comp_buffer, I don't see better way to use dpQueue than to create this intermediate step - lets everybody around work in the old fasion (using comp buffer / audio stream) and use double buffering for DP modules. 

I know it is not perfect and finally I would like to keep a table of sinks/sources only

src/audio/module_adapter/module_adapter.c
```diff
@@ -175,37 +177,33 @@ struct comp_dev *module_adapter_new(const struct comp_driver *drv,
...
-
-		sink_buffer = container_of(blist, struct comp_buffer, source_list);
-		audio_sink[num_of_sinks] =
-				audio_stream_get_sink(&sink_buffer->stream);
-		sink_reset_num_of_processed_bytes(audio_sink[num_of_sinks]);
-		num_of_sinks++;
+		struct comp_buffer *sink_buffer =
+				container_of(blist, struct comp_buffer, source_list);
+		mod->sinks[i] = audio_stream_get_sink(&sink_buffer->stream);
+		i++;
```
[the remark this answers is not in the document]
marcinszkudlinski: It won't always point to buffers - it may be to dpQueue, LL shadred queue, DAI (dai does not need to use any buffer - it may just expose source API), etc. etc.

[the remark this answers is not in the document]
marcinszkudlinski: another reason - performance
call to the module looks like: 

```
	int (*process)(struct processing_module *mod,
		       struct sof_sink **sinks, int num_of_sinks);
```

just a simple pass of pointers

src/audio/module_adapter/module_adapter.c
```diff
@@ -787,62 +755,64 @@ module_single_source_setup(struct comp_dev *dev,
...
-		buffer_stream_invalidate(source_c,
-					 frames * audio_stream_frame_bytes(&source_c->stream));
+	mod->output_buffers[0].data = &mod->sink_comp_buffer->stream;
+
+	if (!mod->skip_src_buffer_invalidate) { /* TODO: add mod->is_multi_core && optimization */
+		/* moved bytes to its own variable to fix checkpatch */
+		uint32_t bytes =
+			frames * audio_stream_frame_bytes(&mod->source_comp_buffer->stream);
+		buffer_stream_invalidate(mod->source_comp_buffer,
+					 bytes);
```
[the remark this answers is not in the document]
marcinszkudlinski: ok, this is a fix to previous PR => separate commit

src/audio/component.c
```diff
@@ -328,9 +328,16 @@ int comp_copy(struct comp_dev *dev)
...
 
-	/* copy only if we are the owner of the LL component */
-	if (dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_LL &&
-	    cpu_is_me(dev->ipc_config.core)) {
+	/* copy only if we are the owner of component OR this is DP component
+	 *
+	 * DP components (modules) require two stage processing:
+	 *  - in first step the pipeline (this) must copy data to/from module DP queues
```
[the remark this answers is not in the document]
marcinszkudlinski: comment extended:
```
	/* copy only if we are the owner of component OR this is DP component
	 *
	 * DP components (modules) require two stage processing:
	 *
	 *   LL_mod -> [comp_buffer -> dp_queue] -> dp_mod -> [dp_queue -> comp_buffer] -> LL_mod
	 *
	 *  - in first step (it means - now) the pipeline must copy source data from comp_buffer
	 *    to dp_queue and result data from dp_queue to comp_buffer
	 *
	 *  - second step will be performed by a thread specific to the DP module - DP module
	 *    will take data from input dpQueue (using source API) , process it
	 *    and put in output DP queue (using sink API)
	 *
	 * this allows the current pipeline structure to see a DP module as a "normal" LL
	 *
	 * to be removed when pipeline 2.0 is ready
	 */
```

src/audio/module_adapter/module_adapter.c
```diff
@@ -208,6 +225,118 @@ static int module_adapter_sink_src_prepare(struct comp_dev *dev)
...
+		dp_queue = dp_queue_create(ibs, obs, dp_mode);
+
+		if (!dp_queue)
+			goto err;
+
+		mod->dp_queue_ll_to_dp[i] = dp_queue;
+		/* it will override source pointers set by module_adapter_sink_src_prepare
+		 * module will use shadow dpQueue for processing
+		 */
+		mod->sources[i] = dp_queue_get_source(dp_queue);
```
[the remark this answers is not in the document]
marcinszkudlinski: CONS: lists are 1) slower 2) harder to debug (no direct access to structs in the lists from debugger)

PROS: List are more memory friendly

I don't like lists personally, especially in the "C way", C++ is a different story

But at this point you're right, since there are tables of sinks/sources and there's no need to iterate through the whole list in every cycle - I'll change a table to a list

src/include/sof/audio/dp_queue.h
marcinszkudlinski: Oops, right
I also change the example to more DP-like, consumer should start immediately when it becomes ready - meaning when it has at least 5 bytes to process

src/include/sof/audio/dp_queue.h
btian1: I am still confusing about cycle meanings.

src/audio/module_adapter/module_adapter.c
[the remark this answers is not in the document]
lyakh: arrays are certainly faster when accessing random elements - it's faster to find the 17th element in an array than to iterate 17 list entries. But (1) we hardly ever have more than 2 buffers in those arrays / lists, and (2) we usually scan them sequentially. In that case I'm not sure arrays are (much) faster. And maintaining multiple access ways to the same data is often a significant burden.

src/include/sof/audio/dp_queue.h
marcinszkudlinski: All audio processing is performed periodically, i.e. LL executes every 1ms
"a cycle" is just an abstract unit of time

src/include/sof/audio/dp_queue.h
marcinszkudlinski: whatever you imagine ;) may be an LL pipeline cycle - it is an abstract time unit for this example

src/ipc/ipc4/helper.c
```diff
@@ -368,21 +366,38 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
 
 	/*
-	 * set ibs and obs in sink/src api of created buffer
-	 *	IBS of a buffer is OBS of source component
-	 *	OBS of a buffer is IBS of destination component
+	 * set ibs and obs in sink/src api of created buffer. Thats may be tricky:
+	 * buffer is connected like:
+	 *	source_src_cfg -> (sink_ifc) BUFFER (source_ifc) -> sink_src_cfg
+	 *
+	 *	source_src_cfg needs to set its OBS (out buffer size) as OBS in buffer's sink ifc
```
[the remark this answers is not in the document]
lyakh: @marcinszkudlinski I agree with your explanation, but just that phrase "OBS in buffer's sink ifc" - I still don't understand it. You describes **module**'s source and sink and input and output buffer sizes (IBS and OBS). But that phrase seems to be describing connections from the buffer's PoV. So, a buffer has a sink and a source, but just one size - the buffer size, is doesn't have OBS and IBS, and even less does it have IBS and OBS on its sink and source independently. Am I wrong?

src/ipc/ipc4/helper.c
```diff
@@ -310,14 +310,10 @@ int ipc_pipeline_free(struct ipc *ipc, uint32_t comp_id)
 }
 
 static struct comp_buffer *ipc4_create_buffer(struct comp_dev *src, bool is_shared,
-					      uint32_t src_obs, uint32_t src_queue,
+					      uint32_t buf_size, uint32_t src_queue,
```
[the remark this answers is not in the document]
marcinszkudlinski: TGL is compiled without this flag

src/audio/module_adapter/module_adapter.c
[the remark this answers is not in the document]
marcinszkudlinski: I'll change it here but it goes a little bit deeper
image

If it really matters (well, as for now I don't think anybody would rub SOF on 64bit arch) there's a need to do it at every such place. May be a lot of them

src/audio/module_adapter/module_adapter.c
```diff
@@ -1098,6 +1294,8 @@ int module_adapter_copy(struct comp_dev *dev)
 	comp_dbg(dev, "module_adapter_copy(): start");
 
 	struct processing_module *mod = comp_get_drvdata(dev);
+	if (mod->dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_DP)
+		return module_adapter_copy_dp_queues(dev);
```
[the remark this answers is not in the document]
marcinszkudlinski: no, a module that is to be run as DP must use sink/src interface

src/audio/module_adapter/module_adapter.c
```diff
@@ -275,6 +273,9 @@ int module_adapter_prepare(struct comp_dev *dev)
 	mod->period_bytes = audio_stream_period_bytes(&sink->stream, dev->frames);
 	comp_dbg(dev, "module_adapter_prepare(): got period_bytes = %u", mod->period_bytes);
 
+	mod->num_input_buffers = 0;
```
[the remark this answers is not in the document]
marcinszkudlinski: I just don't like increasing a field that has not been initialized explicitly

> Also, num_input_buffers compared with num_of_sources?
Right, more less the same, need to change it somehow

marcinszkudlinski:
TODO: 
1) num_input_buffers compared / num_of_sources, do not need 2 for the same 
2) consusing  IBS/OBS in sink/source
3) size_t ??? i think it requires separate change for cleanup

marcinszkudlinski:
>confusing IBS/OBS in sink/source

I made this change, pls check and review

lgirdwood (approved):

src/audio/component.c
```diff
@@ -328,9 +328,25 @@ int comp_copy(struct comp_dev *dev)
 
 	assert(dev->drv->ops.copy);
```
kv2019i, line 329: I still get a compiler error on the commit message. Maybe instead:
```
DP components need to work flawlessly when run on a different
core than that of its connected peer modules.

To achieve this...
```

kv2019i (approved):
Good for me now, thanks @marcinszkudlinski !

src/audio/module_adapter/module_adapter.c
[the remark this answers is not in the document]
lyakh: > ok, got it, sof always treat all un-initialized as 0, init it also acceptable for me.

@btian1 I think this is incorrect and a dangerous misunderstanding. Uninitialised is uninitialised and as such is undefined. It is important to know what has been initialised and then one can rely on that initialisation. OTOH objects that don't need to be initialised at the time of their instantiation can and IMHO usually should be left uninitialised and it's up to us to realise that and to handle those objects respectively. Usually there's no need to initialise objects when it's clear that they will be deterministically completely overwritten before the first evaluation.

marcinszkudlinski:
> num_input_buffers / num_of_sources, do not need 2 for the same

fixed

src/audio/module_adapter/module_adapter.c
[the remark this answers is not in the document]
marcinszkudlinski: it may be in almost any order in module_adapter_copy
just one condition: DP should be before sink_source
in fact maybe it will be better to change it a bit: 

```
	if (IS_PROCESSING_MODE_AUDIO_STREAM(mod))
		return module_adapter_audio_stream_type_copy(dev);

	if (IS_PROCESSING_MODE_RAW_DATA(mod))
		return module_adapter_raw_data_type_copy(dev);

		if (mod->dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_DP) {
			return module_adapter_copy_dp_queues(dev);
		} else {
			return module_adapter_sink_source_copy(dev);
		}
	}

	comp_err(dev, "module_adapter_copy(): unknown processing_data_type");
	return -EINVAL;
```

lgirdwood:
> > num_input_buffers / num_of_sources, do not need 2 for the same
> 
> fixed
> 
> all issues addressed, pls review and merge if possible

@marcinszkudlinski can you check internal CI. Thanks !

marcinszkudlinski:
@lgirdwood internal CI is green  now

kv2019i:
Windows CI fails known (and fix is merged in mainline). nocodec errors in pr-device-test also known, fix in mainline. Proceeding with merge.

PR #8041, opened 2023-08-16
TigerLake: disable memory windows setting on sof side

RanderWang:
Memory windows is set on zephyr side, so remove it and also remove xsram_window for memory windows to save memory since it is not used by ipc4.

Currently we find bug that when we disable CONFIG_TRACE, memory windows can't work well since the CONFIG_TRACE affects memory windows size. It is designed for IPC3 but it is not fit for IPC4.

### DIFF
--- a/src/platform/Kconfig
+++ b/src/platform/Kconfig
@@ -15,7 +15,6 @@ config TIGERLAKE
 	select DMA_GW
 	select DW
 	select DW_DMA if !ZEPHYR_NATIVE_DRIVERS
-	select MEM_WND
 	select DMA_HW_LLI
 	select DW_DMA_AGGREGATED_IRQ
 	select DMA_FIFO_PARTITION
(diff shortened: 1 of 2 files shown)


ujfalusi:
@RanderWang, if you do this then the whole mem_window.c and .h should go, they are no longer used..

I have a PR which achieves the same without deleting code: https://github.com/thesofproject/sof/pull/8043

ujfalusi:
isn't it a problem that `zephyr/CMakeLists.txt` still refers to the mem_window even if you don't build it now?
I think we should be a bit more cautious and do things throughout?

lgirdwood (approved):
Ack no longer needed, @kv2019i any comment.

kv2019i (approved):
Actually, let's go with this. @ujfalusi I'll follow-up with a patch to remove the mem_window from Zephyr makefile.

PR #8044, opened 2023-08-16
Add "is_ready_to_process" call to module API

marcinszkudlinski:
dependencies:
- [x] github.com/thesofproject/sof/pull/8002

This is a modification of module API - add is_ready_to_process call

In case of LL modules all modules have "copy" or "process" method
called periodically, regardless if they're ready to process or not

In case of DP processing the module must have enough data on all
its inputs, enough space for result storage on all outputs, etc.
Unfortunately simply check of number of bytes at all sources/sinks
may not be sufficient (i.e. in case of variable rate audio formats)

The is_ready_to_process call will be called when DP scheduler makes
decision if the module is to be scheduled at this LL tick or not.
If the module wants its thread to start (and process called) it should
return TRUE

### DIFF
--- a/src/audio/src/src.c
+++ b/src/audio/src/src.c
@@ -670,9 +670,9 @@ static int src_verify_params(struct processing_module *mod)
 	return ret;
 }
 
-static int src_get_copy_limits(struct comp_data *cd,
-			       struct sof_source __sparse_cache *source,
-			       struct sof_sink __sparse_cache *sink)
+static bool src_get_copy_limits(struct comp_data *cd,
+				struct sof_source __sparse_cache *source,
+				struct sof_sink __sparse_cache *sink)
 {
 	struct src_param *sp;
 	struct src_stage *s1;
@@ -709,9 +709,9 @@ static int src_get_copy_limits(struct comp_data *cd,
 	}
 
 	if (sp->blk_in == 0 && sp->blk_out == 0)
-		return -EIO;
+		return false;
 
-	return 0;
+	return true;
 }
 
 static int src_params_general(struct processing_module *mod,
@@ -1004,21 +1004,16 @@ static int src_process(struct processing_module *mod,
 		       struct sof_sink __sparse_cache **sinks, int num_of_sinks)
 {
 	struct comp_data *cd = module_get_private_data(mod);
-	struct comp_dev *dev = mod->dev;
-	int ret;
 
-	comp_dbg(dev, "src_process()");
+	comp_dbg(mod->dev, "src_process()");
 
 	/* src component needs 1 source and 1 sink */
-	ret = src_get_copy_limits(cd, sources[0], sinks[0]);
-	if (ret) {
-		comp_dbg(dev, "No data to process.");
+	if (!src_get_copy_limits(cd, sources[0], sinks[0])) {
+		comp_dbg(mod->dev, "No data to process.");
 		return 0;
 	}
 
-	ret = cd->src_func(cd, sources[0], sinks[0]);
-
-	return ret;
+	return cd->src_func(cd, sources[0], sinks[0]);
 }
 
 static int src_set_config(struct processing_module *mod, uint32_t config_id,
--- a/src/audio/src/src.c
+++ b/src/audio/src/src.c
@@ -999,6 +999,16 @@ static int src_prepare(struct processing_module *mod,
 	return src_prepare_general(mod, sources[0], sinks[0]);
 }
 
+
+static bool src_is_ready_to_process(struct processing_module *mod,
+				    struct sof_source __sparse_cache **sources, int num_of_sources,
+				    struct sof_sink __sparse_cache **sinks, int num_of_sinks)
+{
+	struct comp_data *cd = module_get_private_data(mod);
+
+	return src_get_copy_limits(cd, sources[0], sinks[0]);
+}
+
 static int src_process(struct processing_module *mod,
 		       struct sof_source __sparse_cache **sources, int num_of_sources,
 		       struct sof_sink __sparse_cache **sinks, int num_of_sinks)
@@ -1059,6 +1069,7 @@ static struct module_interface src_interface = {
 	.init  = src_init,
 	.prepare = src_prepare,
 	.process = src_process,
+	.is_ready_to_process = src_is_ready_to_process,
 	.set_configuration = src_set_config,
 	.get_configuration = src_get_config,
 	.reset = src_reset,
--- a/src/include/sof/audio/module_adapter/module/module_interface.h
+++ b/src/include/sof/audio/module_adapter/module/module_interface.h
@@ -162,6 +162,25 @@ struct module_interface {
 		       struct sof_sink __sparse_cache **sinks, int num_of_sinks);
 
 	/**
+	 * (optional) return true if the module is ready to process
+	 * This procedure should check if the module is ready for immediate
+	 * processing.
+	 *
+	 * NOTE! the call MUST NOT perform any time consuming operations
+	 *
+	 * this procedure will always return true for LL module
+	 *
+	 * For DP there's a default implementation that will do a simple check if there's
+	 * at least IBS bytes of data on first source and at least OBS free space on first sink
+	 *
+	 * In case more sophisticated check is needed the method should be implemented in
+	 * the module
+	 */
+	bool (*is_ready_to_process)(struct processing_module *mod,
+				    struct sof_source __sparse_cache **sources, int num_of_sources,
+				    struct sof_sink __sparse_cache **sinks, int num_of_sinks);
+
+	/**
 	 * Module specific processing procedure
 	 * This procedure is responsible to consume
 	 * samples provided by the module_adapter and produce/output the processed
--- a/src/include/sof/audio/module_adapter/module/generic.h
+++ b/src/include/sof/audio/module_adapter/module/generic.h
@@ -247,6 +247,30 @@ void module_free_all_memory(struct processing_module *mod);
 int module_prepare(struct processing_module *mod,
 		   struct sof_source __sparse_cache **sources, int num_of_sources,
 		   struct sof_sink __sparse_cache **sinks, int num_of_sinks);
+
+static inline
+bool module_is_ready_to_process(struct processing_module *mod,
+				struct sof_source __sparse_cache **sources,
+				int num_of_sources,
+				struct sof_sink __sparse_cache **sinks,
+				int num_of_sinks)
+{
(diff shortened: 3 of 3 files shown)


src/audio/module_adapter/module/generic.c
```diff
@@ -282,6 +282,20 @@ int module_process_legacy(struct processing_module *mod,
...
+bool module_is_ready_to_process(struct processing_module *mod,
+				struct sof_source __sparse_cache **sources,
+				int num_of_sources,
+				struct sof_sink __sparse_cache **sinks,
+				int num_of_sinks)
+{
+	struct module_data *md = &mod->priv;
+
+	if (!md->ops->is_ready_to_process)
+		return false;
```
lyakh, line 294: this should be `true` IMHO - if no method is implemented, by default always run. Returning `false` here effectively makes this method compulsory, because if it isn't implemented, it will never run

btian1 (commented):
"The is_ready_to_process call will be called when DP scheduler makes
decision if the module is to be scheduled at this LL tick or not.
If the module wants its thread to start (and process called) it should
return TRUE"

will "is_ready_to_process" be called at each period processing?

marcinszkudlinski:
>will "is_ready_to_process" be called at each period processing?

yes, for every DP module
Simple check of buffers usage is not enough

src/audio/src/src.c
```diff
@@ -999,6 +999,18 @@ static int src_prepare(struct processing_module *mod,
...
+
+static bool src_is_ready_to_process(struct processing_module *mod,
+	       struct sof_source __sparse_cache **sources, int num_of_sources,
+	       struct sof_sink __sparse_cache **sinks, int num_of_sinks)
+{
+	struct comp_data *cd = module_get_private_data(mod);
+	int ret;
+
+	ret = src_get_copy_limits(cd, sources[0], sinks[0]);
+	return (ret == 0);
```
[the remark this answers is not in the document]
marcinszkudlinski: really? Much less readable, assembler code probably 100% identical

src/audio/module_adapter/module/generic.c
marcinszkudlinski: and that is the intention
for LLs it does not matter, LLs must be ready at every cycle. For DP modules this must be implemented or the module will start every LL cycle, sometimes pointless. 
So I prefer "false" - it will force to implement the method. Or a runtime assert != NULL

lyakh: if it is compulsory, please check once at initialisation. And only for DP modules. And we have to make sure not to call this for LL modules, or replace this check with an `if (LL) return true;`

src/audio/src/src.c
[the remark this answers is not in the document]
lyakh: readability is subjective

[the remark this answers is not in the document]
btian1: how about directly change src_get_copy_limits return type to bool? it is a static function, only called once in src.c.
and ret < 0 also happen in one place.

btian1 (commented):
if called in each period processing, then it is a burden for processing, it is not light.

src/audio/src/src.c
```diff
@@ -709,9 +709,9 @@ static int src_get_copy_limits(struct comp_data *cd,
 	}
 
 	if (sp->blk_in == 0 && sp->blk_out == 0)
-		return -EIO;
+		return false;
 
-	return 0;
+	return true;
```
btian1, line 714: Thanks, it looks nice, would further suggest use a separate patch for this change.

src/audio/module_adapter/module/generic.c
```diff
@@ -96,13 +96,24 @@ int module_init(struct processing_module *mod, struct module_interface *interfac
...
+		     !!interface->process_raw_data != 1)) {
+			comp_err(dev, "module_init(): LL comp %d is missing mandatory interfaces",
+				 dev_comp_id(dev));
+			return -EIO;
+		}
+	} else {
+		/* DP processing */
+		if (!interface->init || !interface->prepare ||
+		    !interface->reset || !interface->free ||
+		    !interface->process || !interface->is_ready_to_process) {
```
kv2019i, line 112: Minor nit: This could be more readable if the common checks for "init", "prepare", "reset" and "free" would be done first on L98 and then additional checks for LL and DP specifically. 
Not blocking, nice-to-have comment.

src/audio/module_adapter/module/generic.c
```diff
@@ -282,6 +293,21 @@ int module_process_legacy(struct processing_module *mod,
...
+{
+	struct module_data *md = &mod->priv;
+
+	/* LL module has to be always ready for processing */
+	if (mod->dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_LL)
+		return true;
+
+	return md->ops->is_ready_to_process(mod, sources, num_of_sources, sinks, num_of_sinks);
+}
+
```
kv2019i, line 310: I do share a bit of concern @btian1 raised that is the overhead of this function worth it. In many media frameworks, the condition "is ready to process" is coupled to the process() function. 

But I guess if this only is used by DP, and you have sensible usage DP (i.e. you actually save more by not scheduling the processs if it's not ready), then this makes more sense.

kv2019i (commented):
Coide itself looks ok. Not 100% sure about the concept, but if this has concrete benefits for DP implementation, I'm ok to go ahead with this. Please see the comments.

src/audio/module_adapter/module/generic.c
marcinszkudlinski: in DP there's a need to remove it from process and put in separate function.
DP module should not be scheduled till it is ready. It usually is **not** every LL cycle. It's more "on demand" - if the module is ready it will get CPU

Checking it in process() would require schedule DP process, context switch to DP thread, calling process() and having result "nothing to process" 
is much more CPU consuming that this call performed in last LL task context.

I double checked reference design and will do a modification to be consistent: a default implementation of the method:
_return true when there is at least IBS on all inputs and at least OBS free space at sinks_
It may make both implementation easier and sometimes a bit better performance

lyakh:
> Coide itself looks ok. Not 100% sure about the concept, but if this has concrete benefits for DP implementation, I'm ok to go ahead with this. Please see the comments.

@kv2019i yes, this is also my understanding, that this avoids all the overhead of scheduling a Zephyr thread for it only to decide, that it has nothing to do, whereas the scheduler could just call this quick method inline to save that scheduling. Maybe we could have a comment confirming that somewhere in the code.

src/include/sof/audio/module_adapter/module/module_interface.h
```diff
@@ -161,6 +161,23 @@ struct module_interface {
...
+	 * In common case that means quick check if a module is properly
+	 * configured and there's enough data for processing on all sources
+	 * and enough space for result storage in all sinks
+	 *
+	 * NOTE! the call MUST NOT perform any time consuming operations
+	 *
+	 * this procedure is obligatory for DP processing modules
+	 * if not implemented, the wrapper will simply return false
+	 */
+	bool (*is_ready_to_process)(struct processing_module *mod,
```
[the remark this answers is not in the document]
marcinszkudlinski: IADK interface is fixed, no changes in class ProcessingModuleInterface

marcinszkudlinski:
Default is_ready_to_process added
Default implementation has dependency on https://github.com/thesofproject/sof/pull/8002

lgirdwood:
CI shows a build error
```
[ 50%] Building C object CMakeFiles/sof.dir/src/audio/data_blob.c.o
In file included from /home/sof/work/sof.git/src/audio/data_blob.c:7:
/home/sof/work/sof.git/src/include/sof/audio/module_adapter/module/generic.h: In function 'module_is_ready_to_process':
/home/sof/work/sof.git/src/include/sof/audio/module_adapter/module/generic.h:269:51: error: implicit declaration of function 'source_get_ibs'; did you mean 'source_get_rate'? [-Werror=implicit-function-declaration]
  269 |  return (source_get_data_available(sources[0]) >= source_get_ibs(sources[0]) &&
      |                                                   ^~~~~~~~~~~~~~
      |                                                   source_get_rate
/home/sof/work/sof.git/src/include/sof/audio/module_adapter/module/generic.h:[270](https://github.com/thesofproject/sof/actions/runs/5901774504/job/16008545146?pr=8044#step:4:271):35: error: implicit declaration of function 'sink_get_obs'; did you mean 'sink_get_rate'? [-Werror=implicit-function-declaration]
  270 |   sink_get_free_size(sinks[0]) >= sink_get_obs(sinks[0]));
      |                                   ^~~~~~~~~~~~
      |                                   sink_get_rate
[ 51%] Building C object CMakeFiles/sof.dir/src/audio/src/src_generic.c.o
```

marcinszkudlinski:
rebased after https://github.com/thesofproject/sof/pull/8002 merge

btian1 (approved):

abonislawski (approved):

kv2019i:
SOFCI TEST

src/audio/module_adapter/module/generic.c
```diff
@@ -95,16 +95,29 @@ int module_init(struct processing_module *mod, struct module_interface *interfac
...
 	}
 
+	/* for LL there must be one and only one of three possible processing method */
+	if (dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_LL &&
+	    (!!interface->process + !!interface->process_audio_stream +
+	    !!interface->process_raw_data != 1)) {
+		comp_err(dev, "module_init(): LL comp %d is missing mandatory process interface",
+			 dev_comp_id(dev));
+		return -EIO;
+	} else if (dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_DP &&
```
lyakh, line 113: no need for an `else`

lyakh (approved):

lgirdwood:
@marcinszkudlinski can you check internal CI. Thanks !

marcinszkudlinski:
Problem was that in internal CI there's a test that runs smartamp module as a DP. 
This commit introduced a check - DP module must have sink/src interface, which in fact is not required at the moment. The change was there because in initial PR version I wanted to make is_ready_to_process obligatory. After review changes it is optional, so no changes in module interface check is needed at all. **Removed**

**Sink/src  however, will be required for all DP when dpQueue is introduced.** 
(technically it will be required for all DPs on secondary cores - subject to discussion if we want to allow legacy interfaces in case of DPs on same cores as a pipeline)

lgirdwood (approved):

PR #8049, opened 2023-08-16
Tools: Topology2: Add Multiband-DRC to sof-hda-efx-generic.tplg

singalsu:
This patch adds the topology2 files for multiband-DRC. The EFX
    processing chain is gain - IIR - FIR - DRC or Multiband-DRC. The
    DRC selection is done with macro EFX_DRC_COMPONENT and values
    "singleband" and "multiband".
    
    The multiband-drc has two ALSA controls, bytes and switch. The
    switch control switches the processing on and off.
    
    The generated topology files with multiband-drc are:
    
    sof-hda-efx-mbdrc-generic.tplg
    sof-hda-efx-mbdrc-generic-2ch.tplg
    sof-hda-efx-mbdrc-generic-4cg.tplg
    
    The topology files with DRC are as before:
    
    sof-hda-efx-mbdrc-generic.tplg
    sof-hda-efx-mbdrc-generic-2ch.tplg
    sof-hda-efx-mbdrc-generic-4cg.tplg
    
    The configuration blob default is updated to what is generated
    by the current example_multiband_drc.m configuration script. The
    pass-through blob is set as default for tplg2. The topology 1
    blobs are updated similarly.

Picture of sof-hda-efx-mbdrc-generic.tplg with multi-band DRC:
sof-hda-efx-mbdrc-generic

PIcture of sof-hda-efx-mbdrc-generic.tplg with DRC:
sof-hda-efx-generic


tools/topology/topology2/cavs-mixin-mixout-efx-hda.conf
```diff
@@ -55,7 +55,12 @@ Object.Pipeline {
 			}
 			Object.Widget.drc.1 {
 				Object.Control.bytes."1" {
-					name '2 Main playback DRC'
+					name 'Post Mixer $ANALOG_PLAYBACK_PCM DRC'
```
singalsu, line 58: This small fix could be separate commit.

tools/topology/topology2/development/tplg-targets.cmake
```diff
@@ -68,18 +68,19 @@ NHLT_BIN=nhlt-sof-lnl-nocodec-fpga-4ch.bin,PASSTHROUGH=true,DMIC_IO_CLK=19200000
...
 USE_CHAIN_DMA=true,DEEPBUFFER_FW_DMA_MS=100,EFX_FIR_PARAMS=passthrough,EFX_IIR_PARAMS=passthrough,\
-EFX_DRC_PARAMS=passthrough"
+EFX_DRC_PARAMS=passthrough,EFX_MULTIBAND_DRC_PARAMS=passthrough"
 
 "sof-hda-generic\;sof-hda-efx-generic-4ch\;\
 HDA_CONFIG=efx,NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,\
 PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-hda-efx-generic-4ch.bin,USE_CHAIN_DMA=true,\
 DEEPBUFFER_FW_DMA_MS=100,EFX_FIR_PARAMS=passthrough,EFX_IIR_PARAMS=passthrough,\
-EFX_DRC_PARAMS=passthrough"
+EFX_DRC_PARAMS=passthrough,EFX_MULTIBAND_DRC_PARAMS=passthrough"
```
singalsu, line 83: Will try to add some overall EFX configuration to control add of DRC vs. multiband-DRC, both are not needed in same topology for same endpoint. And also simplified versions, maybe with just IIR, just IIR & FIR. Need to find out how to allow the customization without doing a lot of duplication.

tools/topology/topology2/cavs-mixin-mixout-efx-hda.conf
```diff
@@ -59,8 +59,14 @@ Object.Pipeline {
...
 			Object.Widget.multiband_drc.1 {
-				Object.Control.bytes."1" {
-					name 'Post Mixer $ANALOG_PLAYBACK_PCM Multiband-DRC'
+				Object.Control {
+					bytes."1" {
+						name 'Post Mixer $ANALOG_PLAYBACK_PCM MBDRC bytes'
+					}
+					mixer."1" {
+						name 'Post Mixer $ANALOG_PLAYBACK_PCM MBDRC switch'
+						max 1
```
singalsu, line 68: For some reason I needed to duplicate max 1 here

tools/topology/topology2/include/pipelines/cavs/mixout-gain-efx-dai-copier-playback.conf
```diff
@@ -155,10 +156,27 @@ Class.Pipeline."mixout-gain-efx-dai-copier-playback" {
...
+						shift	0
+					}
+					Object.Base.ops.1 {
+						name	"ctl"
+						info	"volsw"
+						#259 binds the mixer control to switch get/put handlers
+						get	259
+						put	259
+					}
+					max 1
```
singalsu, line 179: and here

tools/topology/topology1/m4/multiband_drc_coef_passthrough.m4
```diff
@@ -0,0 +1,70 @@
+# Exported Control Bytes 16-Aug-2023
+CONTROLBYTES_PRIV(MULTIBAND_DRC_priv,
```
lgirdwood, line 2: This is topology1 seperate PR ?

lgirdwood (commented):
PR may need split

tools/topology/topology1/m4/multiband_drc_coef_passthrough.m4
singalsu: I've split the tplg1 blobs update to #8079. Next update to this PR won't contain those.

tools/topology/topology2/cavs-mixin-mixout-efx-hda.conf
lgirdwood: yep.

lgirdwood (commented):
I think you have slef reviewed and identified the changes needed.

tools/topology/topology2/include/components/multiband_drc/passthrough.conf
```diff
@@ -0,0 +1,71 @@
+# Exported Control Bytes 16-Aug-2023
```
lgirdwood, line 1: We really need to have a comment explaining how these are generated.

tools/topology/topology2/include/components/multiband_drc.conf
```diff
@@ -0,0 +1,85 @@
...
+	#
+	Object.Control {
+                mixer."1" {
+                        Object.Base.channel.1 {
+                                name    "fc"
+                                shift	0
+                        }
+                        Object.Base.ops.1 {
+                                name	"ctl"
+                                info	"volsw"
```
cujomalainey, line 67: volsw? why not just switch?

tools/topology/topology2/include/components/multiband_drc.conf
```diff
@@ -0,0 +1,85 @@
...
+		]
+		unique	"instance"
+	}
+
+	#
+	# multiband_drc widget switch control
+	#
+	Object.Control {
+                mixer."1" {
+                        Object.Base.channel.1 {
```
cujomalainey, line 61: why is only one channel defined?

singalsu: There's a single processing control for all channels in the component, so front center single channel was chosen here. It's also same as in this tplg1 pipeline https://github.com/thesofproject/sof/blob/7c013b27e541130bc6b8f3ac6a4a8b22b07cf91a/tools/topology/topology1/sof/pipe-drc-eq-volume-demux-playback.m4#L56.

tools/topology/topology2/include/components/multiband_drc.conf
singalsu: I'm not sure if switch would work, I haven't tried. This is also the same as in tplg1 https://github.com/thesofproject/sof/blob/7c013b27e541130bc6b8f3ac6a4a8b22b07cf91a/tools/topology/topology1/sof/pipe-drc-eq-volume-demux-playback.m4#L51.

Switch control is a mixercontrol (volume) with max defined as 1, so it likely needs to be this. I can try what happens with switch.

tools/topology/topology2/include/components/multiband_drc/passthrough.conf
singalsu: Yep, I'll do a PR to change the text in the tool.

singalsu: Now fixed in this PR, re-exported the blobs. Tool PR is #8142.

cujomalainey (approved):

kv2019i (approved):
@ranj063 @jsarha for tplg conventions, ok?

tools/topology/topology2/include/components/multiband_drc.conf
johnylin76: I'm not sure about topology2. But for topology1 there are 3 control tytes: bytes, volsw, enum. Just as Seppo said, there is no specific type for switch, we use volsw type with max value 1 instead

johnylin76 (approved):
LGTM, thanks.

johnylin76:
Hi @singalsu I have a question about multi-stream mix-out playback. (not related to this PR, just come into my mind while reviewing).

For GOOGLE_AEC topology, the AEC module has one input for echo reference, which is sourced from a DEMUX on playback pipeline. The expectation for compound topology graph of both multi-stream and AEC may be formed as follows:
 ```
 PCM0P (pipeline 2) ---> Volume ------|
                                      V
                                    Mixer (pipeline 1) -----> DEMUX ---> SSP0
                                      ^                         |
 PCM1P (pipeline 3) ---> EQIIR  ------|                         |
                                              |-----------------|
                                              V
 PCM2C (pipeline 5) <--------- NR <--------- AEC <--|
                                                    |
                                                  Mixer (pipeline 4) <--- DMIC1
                                                    |
 PCM3C (pipeline 6) <-------------------------------|
```
Do you think this is technically feasible?
FYI, AEC can be run without active echo reference, the playback and capture path are independent of each other.

singalsu:
> Do you think this is technically feasible?
> FYI, AEC can be run without active echo reference, the playback and capture path are independent of each other.

@johnylin76 The capture direction looks good, except typo with mixer, you must have meant demux there.

In playback direction the IIR location would be typically for all playback just before demux but if you need the other stream un-equalized this would be the way. I don't see why such topology would not work. I once tried 48 -> 16 kHz SRC with a mock-up AEC a reference path that wasn't working due to a scheduler limitation but here the reference path is without processing.

@kv2019i @mengdonglin @plbossart please chime in if this something not committed. Also SOF releases have differences in scheduler & pipelines capability, v2.2 may be more limited.

lgirdwood (approved):

PR #8052, opened 2023-08-17
tplg_parser: Introduce debug print helper

ranj063:
Useful to turn on/off the topology parsing logs at run-time.


lgirdwood (approved):

lgirdwood:
> Intentional to change the default to "off" for the prints...? Good to go if this was the intention.

yep, should be off by default. on by developer action.

Issue #8051, opened 2023-08-17
[BUG] Inconsistent  component ID assignment


aiChaoSONG:
The primary message for init module instance structure is listed below:
```
		struct {
			/**< module id */
			uint32_t module_id          : 16;
			/**< instance id */
			uint32_t instance_id        : 8;
			/**< ModuleMsg::INIT_INSTANCE */
			uint32_t type               : 5;
			/**< Msg::MSG_REQUEST */
			uint32_t rsp                : 1;
			/**< Msg::MODULE_MSG */
			uint32_t msg_tgt            : 1;
			uint32_t _reserved_0        : 1;
		} r;
```
module id takes 16 bits, and instance_id takes next 8 bits.

Here is how we assign component ID;
```
	comp_id = IPC4_COMP_ID(module_init->primary.r.module_id,
			       module_init->primary.r.instance_id);
```
and the ID macros:
```
#define IPC4_COMP_ID(x, y)	((x) << 16 | (y))
#define IPC4_MOD_ID(x) ((x) >> 16)
#define IPC4_INST_ID(x)	((x) & 0xffff)
#define IPC4_SRC_QUEUE_ID(x)	(((x) >> 16) & 0xffff)
#define IPC4_SINK_QUEUE_ID(x)	((x) & 0xffff)
```

`module_id<<16|instance_id` doesn't comply to the ipc structure, which make it hard to do message cross check between mtrace and linux kernel message, and also leave a 8 bit hole in the 32bit comp_id, bad.

aiChaoSONG:
@lgirdwood When the code for IPC4 is added, this issue appears. As ipc_get_comp_by_id() assumes the id is unique among component and pipeline, the author use `module_id<<16|instance_id` to make it unique (a trick), this is based on the module_id is never zero, because module_id==0 is taken by BringUp and BaseFw.

ipc_get_comp_by_id() is also used by IPC3, it is not that easy to fix this in a short time.

lyakh:
In fact in general using bitfields in ABIs seems to be a bad idea, e.g. https://stackoverflow.com/questions/6043483/why-bit-endianness-is-an-issue-in-bitfields . So eventually we might want to replace all such bitfields structures with well-defined for each endianness bitwise access operations

aiChaoSONG:
fixed by #8057, close

PR #8053, opened 2023-08-17
include: ipc4: module: fix component ID macros

aiChaoSONG:
The existing component ID is composed with
'module_id << 16 | instance_id', it doesn't
comply to the initial instance IPC structure:

struct {
    uint32_t module_id   : 16;
    uint32_t instance_id : 8;
    ...
};

When the ID is logged through mtrace, it is
hard to use the ID to find the related linux
kernel message.

This patch fixes component ID composition macros,
thus helps to eliminate component ID inconsistency
 between firmware and linux kernel.

Link: https://github.com/thesofproject/sof/issues/8051

### DIFF
--- a/src/include/ipc4/module.h
+++ b/src/include/ipc4/module.h
@@ -407,10 +407,10 @@ struct ipc4_module_load_library {
 	} data;
 } __packed __aligned(4);
 
-#define IPC4_COMP_ID(x, y)	((x) << 16 | (y))
-#define IPC4_MOD_ID(x) ((x) >> 16)
-#define IPC4_INST_ID(x)	((x) & 0xffff)
-#define IPC4_SRC_QUEUE_ID(x)	(((x) >> 16) & 0xffff)
-#define IPC4_SINK_QUEUE_ID(x)	((x) & 0xffff)
+#define IPC4_COMP_ID(x, y)	((y) << 16 | (x))
+#define IPC4_MOD_ID(x)	((x) & 0xffff)
+#define IPC4_INST_ID(x)	((x) >> 16)
+#define IPC4_SRC_QUEUE_ID(x)	((x) & 0xffff)
+#define IPC4_SINK_QUEUE_ID(x)	(((x) >> 16) & 0xffff)
 
 #endif
--- a/src/ipc/ipc4/helper.c
+++ b/src/ipc/ipc4/helper.c
@@ -50,8 +50,6 @@
 
 LOG_MODULE_DECLARE(ipc, CONFIG_SOF_LOG_LEVEL);
 
-#define IPC4_MOD_ID(x) ((x) >> 16)
-
 extern struct tr_ctx comp_tr;
 
 void ipc_build_stream_posn(struct sof_ipc_stream_posn *posn, uint32_t type,


RanderWang (approved):
LGTM

btian1 (approved):

RanderWang:
> There are test failures, just change the macro is not enough, need more debug
> 
> For IPC4, after correcting the component ID composition macros, the component ID now is calculated with: 'instance_id << 16 | module_id', which complies with the IPC message structure.
> 
> It is possible that an ipc_comp_dev structure for pipeline has the same ID as the ipc_comp_dev structure for component now, for example, a pipeline with pipeline ID 2 has the same ID with a component with 'module_id == 2 & instance_id == 0', under this circumstance, the ipc_get_comp_by_id() won't work.
> 
> This will need repo level cleanup for the use of ipc_get_comp_by_id(), because it's possible to get a pipeline_dev when you want a component_dev, and vice versa. big work.
> 
> @kv2019i @lrgirdwo any idea? do we still want this change

Thanks ! @aiChaoSONG .   I finally recall your finding. I did also find this issue when I define it

btian1 (commented):
shall squash or use separate patch? since the second is fixing CI errors, I prefer squash.

aiChaoSONG:
This PR depends on https://github.com/thesofproject/sof/pull/8057

aiChaoSONG:
The log from firmware and kernel are aligned now:
`[    2.473491] <inf> component: comp_copy: comp:5 0x10010 perf comp_copy samples 48 period 1000 cpu avg 1436 peak 1449 992`

```
[  218.869399] kernel: snd_sof:sof_ipc4_widget_setup: sof-audio-pci-intel-mtl 0000:00:1f.3: Create widget eqiir.12.1 instance 1 - pipe 12 - core 0
[  218.869403] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx      : 0x40010010|0x5000a: MOD_INIT_INSTANCE [data size: 40]
```

Note the last 5 digits of `0x40010010` is component id, wihch is the same in the firmware log.

kv2019i (approved):
This does make log readability better, so why not.

lyakh:
This appears to work - tests pass, but I don't quite understand how :-) These macros are used not only in logging, they're also used in the actual code logic. E.g. in dai-zephyr.c
```
j = IPC4_SINK_QUEUE_ID(sink_c->id);
```
and there are a lot of such places. And that `sink_c->id` comes from
```
buffer->id = desc->comp.id;
```
and the right-hand side is from an IPC itself. So, after this PR that `j` in dai-zephyr.c and similar places will be different, right? And it all still works? Interesting...

aiChaoSONG:
@lyakh In IPC4, for buffer, the buffer ID is composed with source widget's queue id, and sink widget queue ID:
`ipc_buf.comp.id = IPC4_COMP_ID(src_queue, dst_queue)`, if src_widget and sink_widget of a buffer only have one input/output, the buffer in between should have ID=0x00000000.

so IPC4's buffer ID is not unique, but it doesn't matter, because buffers are found and accessed by iterating component's  bsource_list and bsink_list (see struct comp_dev).

This patch does change how src_queue and dst_queue is put in a 32bit value, low 16bits or high 16 bits, but as we also use macro for queue ID access, so it won't cause issue with this patch.

Next, what we can improve is to also make buffer ID unique while keep src/dst_queue info and maybe source/sink widget info.  What's more, the max queue id is 8, so 3 bits is enough for queue ID.

src/include/ipc4/module.h
```diff
@@ -407,10 +407,10 @@ struct ipc4_module_load_library {
...
-#define IPC4_COMP_ID(x, y)	((x) << 16 | (y))
-#define IPC4_MOD_ID(x) ((x) >> 16)
-#define IPC4_INST_ID(x)	((x) & 0xffff)
-#define IPC4_SRC_QUEUE_ID(x)	(((x) >> 16) & 0xffff)
-#define IPC4_SINK_QUEUE_ID(x)	((x) & 0xffff)
+#define IPC4_COMP_ID(x, y)	((y) << 16 | (x))
+#define IPC4_MOD_ID(x) ((x) & 0xffff)
+#define IPC4_INST_ID(x)	((x) >> 16)
+#define IPC4_SRC_QUEUE_ID(x)	IPC4_MOD_ID(x)
+#define IPC4_SINK_QUEUE_ID(x)	IPC4_INST_ID(x)
```
lgirdwood, line 414: any reason why we are not acessing the bit fields directly here ? I know IPC3 had use bit masks and this is probably why we inherited these conveniene macros.

aiChaoSONG: No special reason, just because the bit fields access are the same with IPC4_MOD_ID and IPC4_INST_ID, I will update and keep bit filed access as it was.

lyakh:
> This patch does change how src_queue and dst_queue are put in a 32bit value, (src_q in high 16bits, sink_q in low 16 bits previously, but swapped now), but as we also use macro for queue ID access, so it won't cause issue with this patch.

Ah, I think this is the key - since you change both composition and extraction macros, the internal representation changes, but individual fields keep their values. As long as we never hard-code access to them. Should work then, thanks!

src/include/ipc4/module.h
lgirdwood: no, I mean - we dont need the macro for IPC4, we can remove it and access the bit filed directly and let the CC take core of the shifting/masking for us.

aiChaoSONG:
@lgirdwood ping?

lgirdwood (approved):

PR #8058, opened 2023-08-18
lib_manager: Fix sparse warnings

ujfalusi:
PR #7986 introduced 3 sparse warnings which got overlooked, fix them with a followup patch

There are sparse warnings on zephyr side as well, I will send a PR to fix that.

### DIFF
--- a/src/library_manager/lib_manager.c
+++ b/src/library_manager/lib_manager.c
@@ -529,8 +529,8 @@ static void __sparse_cache *lib_manager_allocate_store_mem(uint32_t size,
 		return NULL;
 	}
 
-	sys_cache_data_invd_range(local_add, size);
-	sys_cache_instr_invd_range(local_add, size);
+	dcache_invalidate_region(local_add, size);
+	icache_invalidate_region(local_add, size);
 
 	return local_add;
 }
--- a/src/library_manager/lib_manager.c
+++ b/src/library_manager/lib_manager.c
@@ -561,7 +561,7 @@ static int lib_manager_store_library(struct lib_manager_dma_ext *dma_ext,
 	ret = lib_manager_store_data(dma_ext, (uint8_t __sparse_cache *)library_base_address +
 				     MAN_MAX_SIZE_V1_8, preload_size - MAN_MAX_SIZE_V1_8);
 	if (ret < 0) {
-		rfree(library_base_address);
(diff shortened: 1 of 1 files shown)


src/library_manager/lib_manager.c
```diff
@@ -529,8 +530,8 @@ static void __sparse_cache *lib_manager_allocate_store_mem(uint32_t size,
 		return NULL;
 	}
 
-	sys_cache_data_invd_range(local_add, size);
-	sys_cache_instr_invd_range(local_add, size);
+	dcache_invalidate_region(local_add, size);
+	icache_invalidate_region(local_add, size);
```
kv2019i, line 534: Why we'd use XTOS wrappers in code that is already Zephyr native (this is including <zephyr/cache.h>)...?

kv2019i (changes requested):
First patch ok, not ok with second.. @lyakh why wrap in code that does not need wrapping?

src/library_manager/lib_manager.c
lyakh: @kv2019i because our wrappers take correct sparse address space as an argument. @ujfalusi initially proposed adding casts there too, but I think it's better to use our wrappers (we also use them a lot everywhere) without type-casting - where possible

kv2019i: right, that's murky but I'll buy the argument. We shouldn't then be including izephyr/cache.h anymore right?

lyakh: @kv2019i yeah, I think we can remove that header now. And BTW there are already multiple calls to `dcache_*_region()` in this file, so, yes, I should've stuck with them, sorry.

lgirdwood (approved):

src/library_manager/lib_manager.c
```diff
@@ -529,8 +530,8 @@ static void __sparse_cache *lib_manager_allocate_store_mem(uint32_t size,
 		return NULL;
```
kv2019i, line 530: @ujfalusi Please remove cache.h inclusion on L24

ujfalusi:
Changes since v1:
- removed cache.h inclusion

kv2019i (approved):

btian1 (approved):

kv2019i:
SOFCI TEST

PR #8059, opened 2023-08-18
audio: module_adapter: Solve the build error using xtensa toolchain

hailongfan-mtk:
Build error info is as follows when compiling with xt-xx toolchain: 
1.redefinition of ‘i’
2.‘for’ loop initial declaration used outside C99 mode
Fix the above error.

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -184,6 +184,7 @@ static int module_adapter_sink_src_prepare(struct comp_dev *dev)
 	uint32_t num_of_sources = 0;
 	uint32_t num_of_sinks = 0;
 	int ret;
+	int i = 0;
 
 	/* acquire all sink and source buffers, get handlers to sink/source API */
 	list_for_item(blist, &dev->bsink_list) {
@@ -212,11 +213,11 @@ static int module_adapter_sink_src_prepare(struct comp_dev *dev)
 	ret = module_prepare(mod, audio_src, num_of_sources, audio_sink, num_of_sinks);
 
 	/* release all source buffers in reverse order */
-	for (int i = num_of_sources - 1; i >= 0; i--)
+	for (i = num_of_sources - 1; i >= 0; i--)
 		buffer_release(source_buffers_c[i]);
 
 	/* release all sink buffers in reverse order */
-	for  (int i = num_of_sinks - 1; i >= 0 ; i--)
+	for  (i = num_of_sinks - 1; i >= 0 ; i--)
 		buffer_release(sinks_buffers_c[i]);
 
 	return ret;
@@ -1013,6 +1014,7 @@ static int module_adapter_sink_source_copy(struct comp_dev *dev)
 	uint32_t num_of_sources = 0;
 	uint32_t num_of_sinks = 0;
 	int ret;
+	int i = 0;
 
 	comp_dbg(dev, "module_adapter_sink_source_copy(): start");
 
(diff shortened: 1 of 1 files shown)


sofci:

src/audio/module_adapter/module_adapter.c
```diff
@@ -184,6 +184,7 @@ static int module_adapter_sink_src_prepare(struct comp_dev *dev)
 	uint32_t num_of_sources = 0;
 	uint32_t num_of_sinks = 0;
 	int ret;
+	int i = 0;
```
lyakh, line 187: superfluous initialisation in both cases. And I think we're now building for C99 by default...

gkbldcig:

src/audio/module_adapter/module_adapter.c
hailongfan-mtk: > According to the following documentation, xt-xcc does not support C99.
> 
> > > > > https://thesofproject.github.io/latest/introduction/index.html
> > > > > Cadence xt-xcc or Cadence xt-clang?
> 
> It depends on the platform. As of June 2020, most platforms supported by SOF rely on xt-xcc. Going forward, all newer platforms require xt-clang. The gcc frontend doesn’t support unusually large registers, hence the move to xt-clang.
> 
> Note that xt-xcc does not fully support C99. xt-clang does.
> 
> Local test result: 1.Add the "-std=c99" C_FLAG , we can support C99, but more build errors will be generated. 2. use xt-clang could fix current issue, but need remove unused function.

According to the following documentation, xt-xcc does not support C99.

https://thesofproject.github.io/latest/introduction/index.html
Cadence xt-xcc or Cadence xt-clang?

It depends on the platform. As of June 2020, most platforms supported by SOF rely on xt-xcc. Going forward, all newer platforms require xt-clang. The gcc frontend doesn’t support unusually large registers, hence the move to xt-clang.

Note that xt-xcc does not fully support C99. xt-clang does.

Local test result:
1.Add the "-std=c99" C_FLAG , we can support C99, but more build errors will be generated.
2. use xt-clang could fix current issue, but need remove unused function.

Test code: [https://gist.github.com/hailongfan-mtk/c474e9b043e1a7df5f24e336369f46a5](url)

lgirdwood:
test this please

src/audio/module_adapter/module_adapter.c
hailongfan-mtk: > superfluous initialisation in both cases. And I think we're now building for C99 by default...

According to the following documentation, xt-xcc does not support C99.

https://thesofproject.github.io/latest/introduction/index.html
Cadence xt-xcc or Cadence xt-clang?

It depends on the platform. As of June 2020, most platforms supported by SOF rely on xt-xcc. Going forward, all newer platforms require xt-clang. The gcc frontend doesn’t support unusually large registers, hence the move to xt-clang.

Note that xt-xcc does not fully support C99. xt-clang does.

Local test result:
1.Add the "-std=c99" C_FLAG , we can support C99, but more build errors will be generated.
2. use xt-clang could fix current issue, but need remove unused function.

Test code: [https://gist.github.com/hailongfan-mtk/c474e9b043e1a7df5f24e336369f46a5](https://github.com/thesofproject/sof/pull/url)

lgirdwood: @hailongfan-mtk so Intel deprecated a lot of the older legacy platforms that used older xt-xcc in CI and this is probably why new PRs are passing with C99 changes. It sounds like we need to enforce the minimum agreed C version on the CI .

lgirdwood (approved):

btian1 (approved):

lgirdwood:
SOFCI TEST

lgirdwood:
@wszypelt @lrudyX CI been pending a while. I've restarted via GH above, but it may be blocked and need a local job start. Can someone check. Thanks !

yaochunhung (approved):

lgirdwood:
@wszypelt @lrudyX can you share the internal CI test failure results with @hailongfan-mtk

wszypelt:
@lgirdwood @hailongfan-mtk I traced the history of tests, I see that there is no single scheme, I will try to perform the tests again and wait for the results

hailongfan-mtk:
@lgirdwood  sir, could you help add me as a SOF member? thanks.

PR #8065, opened 2023-08-21
platform: tigerlake: remove unused definitions

kv2019i:
platform: tigerlake: cleanup unused code for IPC3, SPI, GPIO, XTOS
platform: tigerlake: remove unused definitions

### DIFF
--- a/src/platform/intel/cavs/platform.c
+++ b/src/platform/intel/cavs/platform.c
@@ -11,13 +11,10 @@
 #if (CONFIG_CAVS_LPS)
 #include <cavs/lps_wait.h>
 #endif
-#include <cavs/mem_window.h>
 #include <sof/common.h>
 #include <sof/compiler_info.h>
 #include <sof/debug/debug.h>
-#include <sof/drivers/dw-dma.h>
 #include <rtos/idc.h>
-#include <rtos/interrupt.h>
 #include <sof/ipc/common.h>
 #include <rtos/timer.h>
 #include <sof/fw-ready-metadata.h>
@@ -38,7 +35,6 @@
 #include <sof/schedule/edf_schedule.h>
 #include <sof/schedule/ll_schedule.h>
 #include <sof/schedule/ll_schedule_domain.h>
-#include <sof/trace/dma-trace.h>
 #include <sof/trace/trace.h>
 #include <ipc/header.h>
 #include <ipc/info.h>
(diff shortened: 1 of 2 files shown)


src/platform/intel/cavs/platform.c
```diff
@@ -11,13 +11,10 @@
 #if (CONFIG_CAVS_LPS)
```
btian1, line 11: will CONFIG_CAVS_LPS be set in one platform?

btian1 (approved):

src/platform/intel/cavs/platform.c
```diff
@@ -322,14 +153,6 @@ int platform_init(struct sof *sof)
 	if (ret < 0)
 		return ret;
 
-	/* init low latency single channel DW-DMA domain and scheduler */
-	sof->platform_dma_domain =
-		dma_single_chan_domain_init
-			(&sof->dma_info->dma_array[PLATFORM_DW_DMA_INDEX],
-			 PLATFORM_NUM_DW_DMACS,
-			 PLATFORM_DEFAULT_CLOCK);
-	scheduler_init_ll(sof->platform_dma_domain);
```
lyakh, line 331: this is removing functionality that isn't used by default and has a high, but not 100% chance of being broken. Still it's also possible at least in theory that someone is using it now, isn't it? So, it's probably a decision call - whether we want to deprecate it first and maybe disable by default or just remove it as you do it here

lyakh (approved):

src/platform/intel/cavs/platform.c
kv2019i: @btian1 I'll follow-up with separate PR to remove CAVS_LPS stuff. This needs a bit more surgery so I'll do that in a separate patch.

src/platform/intel/cavs/platform.c
kv2019i: @lyakh Good catch. I did go a bit back and forth on this. But in the end, cAVS2.5 IPC4 config is only made available as a opt-in release (with e.g. SOF2.6), so the usage at this point is very limited, so chances of somebody using it for something exotic like DMA scheduler domain, would seem much smaller than odds of winning in lottery. So I went ahead and removed.

kv2019i:
Let me rebase and upload. One Intel internal test fails, but the fails are on platforms not touched in this PR, so seems like issues with the DUTs. Let me rebase.

PR #8069, opened 2023-08-21
[RFC] remove post-mixout gain for playback

yongzhi1:
By removing the gain, we can save 10183 CPC, thus allows LF clock (38.4M) to be used for deep-buf and regular spk/hs playback use cases.

### DIFF
--- a/tools/topology/topology2/cavs-rt5682.conf
+++ b/tools/topology/topology2/cavs-rt5682.conf
@@ -16,7 +16,7 @@
 <io-gateway.conf>
 <io-gateway-capture.conf>
 <host-copier-gain-mixin-playback.conf>
-<mixout-gain-dai-copier-playback.conf>
+<mixout-dai-copier-playback.conf>
 <deepbuffer-playback.conf>
 <dai-copier-be.conf>
 <dai-copier-eqiir-module-copier-capture.conf>
@@ -232,7 +232,7 @@ Object.Pipeline {
 		}
 	]
 
-	mixout-gain-dai-copier-playback [
+	mixout-dai-copier-playback [
 		{
 			index	2
 
@@ -242,12 +242,6 @@ Object.Pipeline {
 				copier_type	"SSP"
 				stream_name	"$HEADSET_CODEC_NAME"
 				node_type $I2S_LINK_OUTPUT_CLASS
-			}
-
-			Object.Widget.gain.1 {
-				Object.Control.mixer.1 {
-					name	'Post Mixer $HEADSET_PCM_NAME Playback Volume'
-				}
 			}
 		}
 		{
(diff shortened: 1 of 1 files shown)


plbossart:
@yongzhi1 what about actual products which will select a 3rd party processing? Optimizing power for the simple case where we have no processing seems to have limited ROI.

yongzhi1:
> @yongzhi1 what about actual products which will select a 3rd party processing? Optimizing power for the simple case where we have no processing seems to have limited ROI.

I agree it's worth it only if the current tplg is in its final form and removing pre-mixer gain does not hurt anything.

btian1:
Note： gain only have around 3 avg mcps, and if you can achieve 38.4mhz playback with remove the gain, means you are still around 38.4mhz, and each module peak mcps is 3-5 times than avg, you still have chance to get under-run for the pipeline.
it is not safe.

lgirdwood (commented):
Shouldn't we change the deep buffer topology only ?

yongzhi1:
> Note： gain only have around 3 avg mcps, and if you can achieve 38.4mhz playback with remove the gain, means you are still around 38.4mhz, and each module peak mcps is 3-5 times than avg, you still have chance to get under-run for the pipeline. it is not safe.

Thanks @btian1 for the input, with this change, "Sum of KCPS consumption: 29119", looks the performance redundancy is on the low side with 38.4Mhz. I originally thought that the margin is already built into the formula:  
https://github.com/thesofproject/sof/blob/mtl-005-hotfix3/src/lib/cpu-clk-manager.c#L62

plbossart:
> Shouldn't we change the deep buffer topology only ?

endpoint processing happens post-mix.

lgirdwood:
> > Shouldn't we change the deep buffer topology only ?
> 
> With both gain and mixin removed from deep buffer topology, the minimalist stream now only needs "Sum of KCPS consumption: 28851":
> 
> sof-mtl-max98357a-rt5682-ssp2-ssp0-db

Thanks @yongzhi1 the diagram answers my question. Next steps
1) Have we measured peak/avg MCPS ?
2) Do we have have WIP patches for CPC manifest updates for copier, mixin, mixout and gain for this format, channels, rate ?

yongzhi1:
> 1. Have we measured peak/avg MCPS ?
> 2. Do we have have WIP patches for CPC manifest updates for copier, mixin, mixout and gain for this format, channels, rate ?

I used the CPC from https://github.com/thesofproject/rimage/blob/mtl-005-drop-stable/config/mtl.toml, will check overruns with more tests.

btian1:
> With

28851 corresponding mcps is? also the accumulation is based on component, right? it may not reflect the actual pipeline real cost.
there is pipeline cost in perf log, I hope later can be added into CI as well:

[    4.987705] <inf> ll_schedule: do_task_run: perf_cycle task 0xbe0c2340, 0x20180U cpu avg 1616 peak 4802
[    5.999661] <inf> ll_schedule: do_task_run: perf_cycle task 0xbe0c15c0, 0x20180U cpu avg 1655 peak 3806

you can see, peak pipe cycle > 2 times compared with avg.

btian1:
> pared with avg.

As long as you need adjust volume for current pipeline, the current volume ramp performance is quite big, > 20 mcps for worst case, it is almost impossible to achieve 38.4mhz LP playback, please watch progress on issue #8162.

yongzhi1:
Thanks @btian1 for the info, seems only passthrough can benefit from the 38.4Mhz low freq clock.
cc @tmleman for possible ways to optimize individual component and close this one then.

PR #8070, opened 2023-08-21
topology2: fix SoundWire files

plbossart:
The topology2 seems to be in a constant state of poorly maintained stuff piled of top of mistakes with no agreed indentation style.

It's really scary. This PR is a simple attempt to at least cleanup the SoundWire parts. I could be wrong on some of the changes, but the existing code has a number of issues as well, e.g. a number of formats and configurations that make absolutely no sense.

@ranj063 @bardliao @jsarha please review. This is both urgent and important at this point.

### DIFF
--- a/tools/topology/topology2/platform/intel/bt-generic.conf
+++ b/tools/topology/topology2/platform/intel/bt-generic.conf
@@ -25,35 +25,44 @@ Object.Pipeline {
 				num_input_audio_formats 3
 				num_output_audio_formats 3
 				num_input_pins 1
-
-				Object.Base.audio_format.0 {
-					in_bit_depth		16
-					in_valid_bit_depth	16
-					out_bit_depth		16
-					out_valid_bit_depth	16
-					in_channels		2
-					out_channels		2
-				}
-				Object.Base.audio_format.2 {
-					in_bit_depth		16
-					in_valid_bit_depth	16
-					out_bit_depth		16
-					out_valid_bit_depth	16
-					in_channels		1
-					out_channels		1
-					in_rate		8000
-					out_rate		8000
-				}
-				Object.Base.audio_format.3 {
-					in_bit_depth		16
-					in_valid_bit_depth	16
-					out_bit_depth		16
-					out_valid_bit_depth	16
-					in_channels		1
-					out_channels		1
-					in_rate		16000
-					out_rate		16000
-				}
+				Object.Base.input_audio_format [
+					{
+						in_bit_depth		16
+						in_valid_bit_depth	16
+						in_channels		2
+					}
+					{
+						in_bit_depth		16
+						in_valid_bit_depth	16
+						in_channels		1
+						in_rate			8000
+					}
+					{
+						in_bit_depth		16
+						in_valid_bit_depth	16
+						in_channels		1
+						in_rate			16000
+					}
+				]
+				Object.Base.output_audio_format [
+					{
+						out_bit_depth		16
+						out_valid_bit_depth	16
+						out_channels		2
+					}
+					{
+						out_bit_depth		16
+						out_valid_bit_depth	16
+						out_channels		1
+						out_rate		8000
+					}
+					{
+						out_bit_depth		16
+						out_valid_bit_depth	16
+						out_channels		1
+						out_rate		16000
+					}
+				]
 			}
 		}
 	]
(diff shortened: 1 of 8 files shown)


bardliao (approved):

lgirdwood (approved):

tools/topology/topology2/platform/intel/sdw-amp-generic.conf
```diff
@@ -349,11 +349,21 @@ IncludeByKey.SDW_AMP_FEEDBACK {
 					Object.Widget.host-copier.1 {
 						stream_name	"amp feedback"
 						pcm_id 3
-						Object.Base.audio_format.1 {
-							# 32 -> 16 bits conversion is done here,
-							# so in_bit_depth is 32 (and out_bit_depth is 16).
-							in_bit_depth	32
-						}
+						num_input_audio_formats 1
```
ranj063, line 352: @plbossart you've got a lot of failures in the SDW device tests and it is because you have removed a bunch of supported formats here

tools/topology/topology2/platform/intel/bt-generic.conf
```diff
@@ -160,36 +190,47 @@ Object.Pipeline {
 				copier_type	"SSP"
 				stream_name	$BT_NAME
 				node_type	$I2S_LINK_INPUT_CLASS
-				num_input_audio_formats 4
-				num_output_audio_formats 4
```
plbossart, line 164: @ranj063 may I removed too many things but can we agree that this is completely wrong? There are 3 formats supported

tools/topology/topology2/platform/intel/sdw-dmic-generic.conf
```diff
@@ -30,11 +30,21 @@ Object.Pipeline {
 			Object.Widget.host-copier.1 {
 				stream_name	"sdw dmic"
 				pcm_id 4
-				Object.Base.audio_format.1 {
-					# 32 -> 16 bits conversion is done here,
-					# so in_bit_depth is 32 (and out_bit_depth is 16).
-					in_bit_depth	32
-				}
```
plbossart, line 37: someone will also have to explain to me what this means?

tools/topology/topology2/platform/intel/sdw-jack-generic.conf
```diff
@@ -186,56 +186,58 @@ Object.Pipeline.host-gateway-capture [
 		Object.Widget.host-copier.1 {
 			stream_name	"Passthrough Capture 0"
 			pcm_id 1
-			num_input_audio_formats 9
-			num_output_audio_formats 9
```
plbossart, line 190: and that's also clearly wrong @ranj063. I don't know how we identified 9 formats below.

tools/topology/topology2/platform/intel/sdw-amp-generic.conf
plbossart: I am not sure what you are referring to @ranj063, I just split input and output definitions. There are a couple of points where the starting point seems wrong, I don't have a clear picture of how these topologies were reviewed and verified.

bardliao: @plbossart [num_input_audio_formats 6](https://github.com/thesofproject/sof/blob/main/tools/topology/topology2/include/pipelines/cavs/host-gateway-capture.conf#L50) and the supported formats are defined in the host-gateway-capture pipeline. We just overwrite the first supported format here and don't touch other supported formats. When you overwrite `num_input_audio_formats` to 1, you will lose the other 5 supported formats. Same as `num_output_audio_formats`

tools/topology/topology2/platform/intel/sdw-dmic-generic.conf
bardliao: > someone will also have to explain to me what this means?

We want to do 32 -> 16 bits conversion in the copier. So, set in_bit_depth	to 32, and the default value of out_bit_depth is 16.

tools/topology/topology2/platform/intel/sdw-jack-generic.conf
bardliao: > and that's also clearly wrong @ranj063. I don't know how we identified 9 formats below.

The first 6 supported formats are already defined in [host-gateway-capture.conf](https://github.com/thesofproject/sof/blob/main/tools/topology/topology2/include/pipelines/cavs/host-gateway-capture.conf). So, we just define the 7th to 9th supported formats here.

tools/topology/topology2/platform/intel/sdw-amp-generic.conf
plbossart: Goodness. We should never ever do this sort of things. 

Unless I am gravely mistaken @bardliao @ranj063 the premise of this topology2 is that this looks object-oriented, so there's a presumption of inheritance of properties. Overriding stuff without even a comment to that effect is really pushing it, no? 

Why do we need to convert anyways?

tools/topology/topology2/platform/intel/sdw-jack-generic.conf
ranj063: @plbossart that was a mistake. We cannot maintain this for sure. We should have defined all 9 formats here. Anyway, the reason why the audio_format class has been left into these files is because I wanted to use the new DefineArrays feature for these because we have quite a few. Do you mind if I take over this PR?

plbossart: > Do you mind if I take over this PR?

Be my guest and have at it! I just started this out of frustration when trying to add the use_chain_dma stuff, I am all for doing the 'right thing'(tm) here.

plbossart:
I'll let someone smarter than me work on this :-)

PR #8079, opened 2023-08-23
Tools: Topology1: Multiband-DRC: Update setup blobs

singalsu:
The multiband_drc_coef_default.m4 is updated to what the current setup tool example_multiband_drc() exports.

Another blob for pass-through is added with emphasis and all bands processing disabled. Such blob is useful when headphone and speaker paths are shared. Headphone mode should use pass-through.


lgirdwood (approved):

kv2019i:
@singalsu Please backport to stable-v2.2 for IPC3 use.

PR #8083, opened 2023-08-23
Audio: Crossover: Adjust error message for pin index usage in config

singalsu:
In IPC4 system the output sinks of crossover are not defined by pipeline identifiers but by pin indices. The error message is updated to be IPC version agnostic to avoid confusion.


singalsu:
Changing as result from discussion in #8078.

lgirdwood (approved):

kv2019i (approved):

PR #8094, opened 2023-08-25
Tools: Topology: Add ASRC capture to cavs-nocodec.conf topologies

andrula-song:
This patch adds ASRC capture from Dmic such as sof-tgl-nocodec.tplg. Set ASRC_CAPTURE_ENABLE as true enables test of ASRC component with IPC4.

### DIFF
--- a/tools/topology/topology2/development/cavs-nocodec-asrc.conf
+++ b/tools/topology/topology2/development/cavs-nocodec-asrc.conf
@@ -1,0 +1,694 @@
+<searchdir:include>
+<searchdir:include/common>
+<searchdir:include/components>
+<searchdir:include/dais>
+<searchdir:include/pipelines/cavs>
+<searchdir:platform>
+<searchdir:platform/intel>
+<vendor-token.conf>
+<manifest.conf>
+<pdm_config.conf>
+<tokens.conf>
+<virtual.conf>
+<host-gateway-capture.conf>
+<host-gateway-playback.conf>
+<io-gateway-capture.conf>
+<io-gateway.conf>
+<dai-copier-asrc-capture.conf>
+<asrc-dai-copier-playback.conf>
+<gain-playback.conf>
+<data.conf>
+<pcm.conf>
+<pcm_caps.conf>
+<fe_dai.conf>
+<ssp.conf>
+<dmic.conf>
+<intel/hw_config_cardinal_clk.conf>
+<manifest.conf>
+<route.conf>
+<intel/common_definitions.conf>
+<dai-copier.conf>
+<pipeline.conf>
+<dai.conf>
+<host.conf>
+<input_pin_binding.conf>
+<output_pin_binding.conf>
+<input_audio_format.conf>
+<output_audio_format.conf>
+<dmic-default.conf>
+
+Define {
+	MCLK 				24576000
+	NUM_DMICS			0
+	DMIC0_DAI_COPIER		'dai-copier.DMIC.NoCodec-6.capture'
+	DMIC0_NAME			'NoCodec-6'
+	PLATFORM 			"none"
+	SSP0_PCM_NAME			"Port0"
+	SSP1_PCM_NAME			"Port1"
+	SSP2_PCM_NAME			"Port2"
+	SSP0_PCM_ID			0
+	SSP1_PCM_ID			1
+	SSP2_PCM_ID			2
+}
+
+# override defaults with platform-specific config
+IncludeByKey.PLATFORM {
+	"tgl"	"platform/intel/tgl.conf"
+	"adl"	"platform/intel/tgl.conf"
+	"mtl"	"platform/intel/mtl.conf"
+	"lnl"	"platform/intel/mtl.conf"
+}
+
+# Dmic-->ASRC-->host copier
+
+Object.Dai.SSP [
+	{
+		id 		0
+		dai_index	0
+		direction	"duplex"
+		name		NoCodec-0
+		default_hw_conf_id	0
+		sample_bits		32
+		quirks			"lbm_mode"
+		io_clk		$MCLK
+
+		Object.Base.hw_config.1 {
+			name	"SSP0"
+			id	0
+			bclk_freq	3072000
+			tdm_slot_width	32
+			# TODO: remove this. Needs alsaptlg change.
+			Object.Base.link_config.1 {
+				clock_source	1
+			}
+		}
+	}
+
+	{
+		id 		1
+		dai_index	1
+		direction	"duplex"
+		name		NoCodec-1
+		default_hw_conf_id	0
+		sample_bits		32
+		quirks			"lbm_mode"
+		io_clk		$MCLK
+
+		Object.Base.hw_config.1 {
+			name	"SSP1"
+			id	0
+			bclk_freq	3072000
+			tdm_slot_width	32
+			# TODO: remove this. Needs alsaptlg change.
+			Object.Base.link_config.1 {
+				clock_source	1
+			}
+		}
+	}
+
+	{
+		id 		2
+		dai_index	2
+		direction	"duplex"
+		name		NoCodec-2
+		default_hw_conf_id	0
+		sample_bits		32
+		quirks			"lbm_mode"
+		io_clk		$MCLK
+
+		Object.Base.hw_config.1 {
+			name	"SSP2"
+			id	0
+			bclk_freq	3072000
+			tdm_slot_width	32
+			# TODO: remove this. Needs alsaptlg change.
+			Object.Base.link_config.1 {
+				clock_source	1
+			}
+		}
+	}
+]
+
+Object.Dai.DMIC [
+	{
+		dai_index 0
+		name			$DMIC0_NAME
+		id 			$DMIC0_ID
+		driver_version		$DMIC_DRIVER_VERSION
+		io_clk			$DMIC_IO_CLK
+		clk_min		500000
+		clk_max		4800000
+		unmute_ramp_time_ms	200
+		# num_pdm_active should always set to 2 but depending on the number of DMIC's
+		# the mic's are enabled or disabled in each PDM.
+		num_pdm_active 	2
+
+		Object.Base.hw_config.1 {
+			id	0
+			name	"DMIC0"
+		}
+
+		# PDM controller config
+		Object.Base.pdm_config.1 {
+			mic_a_enable	$PDM0_MIC_A_ENABLE
+			mic_b_enable	$PDM0_MIC_B_ENABLE
+			ctrl_id	0
+		}
+
+		Object.Base.pdm_config.2 {
+			ctrl_id	1
+			mic_a_enable	$PDM1_MIC_A_ENABLE
+			mic_b_enable	$PDM1_MIC_B_ENABLE
+		}
+	}
+]
+
+# Dmic-->ASRC-->host copier PCM ASRC Capture3
+# SSP0-->host copier PCM Port0
+# SSP1-->ASRC-->host copier PCM Port1
+# SSP2-->ASRC-->host copier PCM Port2
+Object.Pipeline.dai-copier-asrc-capture [
+	{
+		index		1
+		core_id		$DMIC_CORE_ID
+		Object.Widget.dai-copier.1 {
+			dai_index	0
+			dai_type	"DMIC"
+			copier_type	"DMIC"
+			type		dai_out
+			stream_name	$DMIC0_NAME
+			node_type $DMIC_LINK_INPUT_CLASS
+			num_input_audio_formats	2
+			Object.Base.input_audio_format [
+				{
+					in_bit_depth		32
+					in_valid_bit_depth	32
+				}
+				{
+					in_channels		4
+					in_bit_depth		32
+					in_valid_bit_depth	32
+					in_ch_cfg	$CHANNEL_CONFIG_3_POINT_1
+					in_ch_map	$CHANNEL_MAP_3_POINT_1
+				}
+			]
+			num_output_audio_formats	2
+			Object.Base.output_audio_format [
+				{
+					out_bit_depth		32
+					out_valid_bit_depth	32
+				}
+				{
+					out_channels		4
+					out_bit_depth		32
+					out_valid_bit_depth	32
+					out_ch_cfg	$CHANNEL_CONFIG_3_POINT_1
+					out_ch_map	$CHANNEL_MAP_3_POINT_1
+				}
+			]
+		}
+
+		Object.Widget.asrc."1" {
+			rate_out 48000
+			num_input_audio_formats	2
+			Object.Base.input_audio_format [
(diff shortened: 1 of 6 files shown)


andrula-song:
add pipeline 23, capture pipeline from dmic0-> asrc->host-copier
sof-tgl-nocodec

tools/topology/topology2/cavs-nocodec.conf
```diff
@@ -354,6 +356,112 @@ IncludeByKey.PASSTHROUGH {
 			}
 		}
 
+		IncludeByKey.NUM_DMICS {
+			"[1-4]" {
+						IncludeByKey.ASRC_CAPTURE_ENABLE {
```
lgirdwood, line 361: indentation ?

lgirdwood (commented):
@singalsu @jsarha @ranj063 pls review

tools/topology/topology2/cavs-nocodec.conf
jsarha: Do I read this correctly, we have?
```
IncludeByKey.PASSTHROUGH {
"false" {
	...
	IncludeByKey.NUM_DMICS {
	"[1-4]"	{
		...
		}
	}

+	IncludeByKey.NUM_DMICS {
+	"[1-4]" {
+		IncludeByKey.ASRC_CAPTURE_ENABLE {
+		"true"	{
+			...
+			}
+		}
+	}
	...
```

Is there some reason for having two NUM_DMIC [1-4] blocks following each other, or is this use of big nested IncludebyKey blocks just getting out of hand?

Not in the scope of this PR, but it could be a time to split nocodec file into smaller pieces included from within IncludeByKey blocks.

plbossart:
I still have concerns about 
a) why we think ASRC is a good idea for a nocodec topology where all clocks are derived from the same master clock...
b) how ASRC is actually tested and what the value of this integration is.

NAK for now.

tools/topology/topology2/include/components/asrc_format_s32_convert_from_24k.conf
```diff
@@ -0,0 +1,26 @@
+#asrc format array
```
singalsu, line 1: Are these the same as with SRC, could the same .conf files be used for both?

singalsu:
> I still have concerns about a) why we think ASRC is a good idea for a nocodec topology where all clocks are derived from the same master clock... b) how ASRC is actually tested and what the value of this integration is.
> 
> NAK for now.

Could the topology for nocodec be like this:

```
host copier <--- ASRC <--- DAI copier DMIC

host copier ---> ASRC ---> DAI copier SSP0
host copier <------------- DAI copier SSP0

host copier -------------> DAI copier SSP2
host copier <--- ASRC <--- DAI copier SSP2

host copier ---> ASRC ---> DAI copier SSP1
host copier <--- ASRC <--- DAI copier SSP1
```
It would allow to capture back playback ASRC output in SSP loopback mode (LBM). And feed data for capture ASRC.

The SSP1 duplex ASRC is to test shared DAI tracking. SSP1 is not working in UPX so some other DUT needs to be used or switch SSPs to test such after the two simpler cases are tested to work.

The default SSP configuration in nocodec is clock producer while we would need follower role. But this would be a start until we figure out how to get follower mode to work and how to hook a clock producer sound card to UPX audio HAT.

I tested ASRC originally with UP2 and Digiberry S/PDIF sound card and Pierre's SSP follower setup. I used S/PDIF connection to a USB sound card. It drifted in rate slightly vs. timer scheduled SOF.

It would be a fair effort to try to achieve similar setup with UPX. Do you think we should do it or just stick in tracking a synchronous DAI and believing that it would track a clock follower DAI?

andrula-song:
the latest topology:
sof-tgl-nocodec-asrc

tools/topology/topology2/include/pipelines/cavs/dai-copier-asrc-capture.conf
```diff
@@ -0,0 +1,72 @@
...
+	Object.Widget {
+		dai-copier."1" {
+			type dai_out
+			<include/components/asrc_format_sxx_to_s32_convert.conf>
+			num_output_pins 1
+		}
+
+		asrc."1" {
+			asynchronous_mode	1
+			operation_mode		1
```
aiChaoSONG, line 52: Those two attributes could be transformed to string type to make it human readable.

tools/topology/topology2/include/pipelines/cavs/dai-copier-asrc-capture.conf
```diff
@@ -0,0 +1,72 @@
...
+#
+
+<include/common/audio_format.conf>
+<include/components/dai-copier.conf>
+<include/components/asrc.conf>
+<include/components/pipeline.conf>
+
+Class.Pipeline."dai-copier-asrc-capture" {
+
+	DefineAttribute."index" {}
```
aiChaoSONG, line 27: duplicated definition with pipeline-common.conf

tools/topology/topology2/development/cavs-nocodec-asrc.conf
```diff
@@ -0,0 +1,758 @@
...
+			dai_type	"SSP"
+			type		dai_out
+			copier_type	"SSP"
+			stream_name	"NoCodec-1"
+			node_type	$I2S_LINK_INPUT_CLASS
+		}
+
+		Object.Widget.asrc."1" {
+			rate_in 48000
+			rate_out 44100
```
singalsu, line 264: Is this a capture ASRC with 44.1 kHz rate out to PCM?

tools/topology/topology2/development/cavs-nocodec-asrc.conf
```diff
@@ -0,0 +1,758 @@
...
+			dai_type "SSP"
+			copier_type "SSP"
+			stream_name "NoCodec-0"
+			node_type $I2S_LINK_OUTPUT_CLASS
+			num_input_pins 1
+			<include/components/asrc_format_sxx_to_s32_convert.conf>
+		}
+		Object.Widget.asrc."1" {
+			rate_in 48000
+			rate_out 44100
```
singalsu, line 541: Is they playback DAI really 44.1 kHz?

tools/topology/topology2/include/components/asrc_format_s32_to_sxx_convert.conf
```diff
@@ -0,0 +1,346 @@
+#src format array
+			num_input_audio_formats 14
+
+			Object.Base.input_audio_format [
+				# 8khz input
+				{
```
singalsu, line 6: These could use the new array definition in alsatplg. Can these be shared, same include files, with SRC?

tools/topology/topology2/include/pipelines/cavs/asrc-dai-copier-playback.conf
```diff
@@ -0,0 +1,104 @@
...
+					in_valid_bit_depth      32
+				}
+			]
+			num_output_audio_formats 1
+			Object.Base.output_audio_format [
+				{
+					out_bit_depth           32
+					out_valid_bit_depth     32
+				}
+			]
```
singalsu, line 86: Why 3 input audio formats and one output? The asrc is not supporting format conversion, but same input and output format for s16/s24/s32.

tools/topology/topology2/development/cavs-nocodec-asrc.conf
andrula-song: sorry, used don't know that SSP can not do sample convert.

tools/topology/topology2/development/cavs-nocodec-asrc.conf
andrula-song: No, it is 48kHz

tools/topology/topology2/development/cavs-nocodec-asrc.conf
```diff
@@ -0,0 +1,694 @@
...
+	}
+]
+
+
+
+# PCM0 --> ASRC--> SSP0
+
+# PCM1 --> ASRC--> SSP1
+
+# PCM2--> SSP2
```
singalsu, line 531: In my UPX SSP0 and SSP2 work, SSP1 doesn't, I have used used SSP0 and SSP2 in topologies for tgl  but I guess your test device is different.

singalsu (approved):
Looks good to me

tools/topology/topology2/development/cavs-nocodec-asrc.conf
andrula-song: yes, I can get the right curve

andrula-song:
hello @ranj063 , can you help to review this PR?

lgirdwood:
@andrula-song conflicts.

lgirdwood:
@andrula-song @singalsu @plbossart my preference is to maximise code under regular test, even if we have to fake the remote time domain here in this mode.

Thinking aloud we could have a test Kconfig for ASRC that fakes the remote time domain. i.e. we can read CCOUNT (via Zephyr API) and convert some of it's LSBs to a random drift value.

tools/topology/topology2/include/components/asrc.conf
```diff
@@ -94,6 +119,8 @@ Class.Widget."asrc" {
 	# Default attributes for asrc
 	#
 	type		"asrc"
-	uuid 		"f6:72:ec:c8:26:85:af:4f:9d:39:a2:3d:0b:54:1d:e2"
+	uuid 		"2d:40:b4:66:68:b4:f2:42:81:a7:b3:71:21:86:3d:d4"
```
kv2019i, line 122: This is a bit odd, why does the UUID change?

tools/topology/topology2/include/components/asrc_format_s32_convert_from_24k.conf
```diff
@@ -0,0 +1,56 @@
...
+					out_bit_depth            32
+					out_valid_bit_depth      24
+				}
+				{
+					out_rate                 16000
+					out_bit_depth            16
+					out_valid_bit_depth      16
+				}
+
+			num_input_audio_formats 1
```
kv2019i, line 38: Ditto here, 1 format, but below there are 3 formats listed.

kv2019i (changes requested):
Few minor notes on the tplg code

plbossart:
> @andrula-song @singalsu @plbossart my preference is to maximise code under regular test, even if we have to fake the remote time domain here in this mode.

Of course we want to test as many things as possible, but we are both resource-constrained and time-constrained. 

The fundamental issue is IMHO a confusion between component testing and CI (Continuous Integration). We have exactly zero products that rely on ASRC, so while it's an important capability it has by construction a lower priority. I will assert that ASRC does not belong in the nocodec topology that is the basic for our quick smoke tests for hardware (DMA, DAIs, etc). Exhibit A for this assertion is what we do for each power-on: we fall-back to the traditional passthrough pipelines and gradually add complexity.

And even if this ASRC component is added as an opt-in, it's not clear to me if/when this component would be tested, by what/whom and at which frequency.

I would like the test methodology to be addressed before we merge anything. "Merge first, think later about test" will lead to bitrot or be dependent on developer bandwidth, both of which are not desirable directions.

singalsu:
@plbossart The name of this PR is somewhat confusing, the usual nocodec topology we use is not impacted, this adds another nocodec type topology to use alternatively. It can be seen from the CMakeLists.txt part or PR `development/cavs-nocodec-asrc\;sof-tgl-nocodec-asrc`. The source .conf is different and the produced topology name is different.

singalsu:
Even if this would not be used in regular test there should be a way to store a topology as evidence that the modifications to code (this time ipc4 & module adapter client) at least with that git hash worked. It's a lot of effort to create a new topology from scratch, and these topologies for test are very helpful when making the real production topologies. All the small details in getting things to work are quickly forgotten.

plbossart:
@singalsu I can only concur with the notion that writing a new topology is hard, but so is maintaining a topology that has a gazillion options. It took us so much time to restore basic passthrough functionality which was broken by gradual additions. @ranj063 is still working on the format stuff, and we'll have even more changes coming.

I don't really follow the point of using a hash as a proof of the last working setup. We really need these files to be maintained, and that means agreeing on what gets tested and when.

tools/topology/topology2/include/components/asrc.conf
andrula-song: Done, thanks!

singalsu:
> It took us so much time to restore basic passthrough functionality which was broken by gradual additions.

It could be solved by splitting the pass-through topology into a separate file and build, like this PR. I've done it myself to understand what is needed for a bare bones topology and made some new topologies from the minimum. The normal cavs-nocodec.conf has grown beyond my comprehension. Changing it a bit too easily breaks something, at least when I'm trying.

kv2019i:
@plbossart I think this matches e.g. situation with "sof-nocodec-bt-*.tplg". I'm in practise the only one who runs tests on this and tests run purely on a per-need basis. 

I'd still this is more useful than not having any topology at all. E.g. this is a reference of how to use a particular component in end-to-end context and with a topology than can be run on widely available hardware. Plus git will give you some indication on who to contact if you want run in to issue.

Would it help if we mandate a dedicated owner for each such topology? We could simply use CODEOWNERS to track owner.

kv2019i (approved):
Thanks @andrula-song , my concerns addressed.

lgirdwood:
@plbossart good now ?

plbossart:
@lgirdwood I am still opposed to the directions. This adds a new tools/topology/topology2/development/cavs-nocodec-asrc.conf file, and that brings up again the question on who is going to maintain yet another file.

I started talking with @ranj063 on a different approach where the host- and dai- copiers are included and only a simpler component added. This should be a pre-canned set of classes with a "PUT YOUR CODE HERE" when adding a new component, be it ASRC, DRC, and whatever is under development.

There is no reason to e.g. create yet another PCM device or replicate a complicated topology. We are again confusing component level testing and integration tests.

lgirdwood:
> @lgirdwood I am still opposed to the directions. This adds a new tools/topology/topology2/development/cavs-nocodec-asrc.conf file, and that brings up again the question on who is going to maintain yet another file.
> 
> I started talking with @ranj063 on a different approach where the host- and dai- copiers are included and only a simpler component added. This should be a pre-canned set of classes with a "PUT YOUR CODE HERE" when adding a new component, be it ASRC, DRC, and whatever is under development.
> 
> There is no reason to e.g. create yet another PCM device or replicate a complicated topology. We are again confusing component level testing and integration tests.

Ok, I'm fine as long as we can have upstream topologies that can easily enable developers

1) Usage of any module under testbench with IPC4.
2) Usage of direct A->B IPC4 pipelines under plugin.

singalsu:
@andrula-song There's a conflict

andrula-song:
> @andrula-song There's a conflict

thanks, but since I have added the benchmark ASRC topologies, I would like to close this one.

PR #8096, opened 2023-08-25
Kconfig.zephyr: check priorities off

pjdobrowolski:
Feature checking build time priority in zephyr is causing building failure. Untile in SOF it won't be adjust that feature should be turned off.

### DIFF
--- a/zephyr/Kconfig
+++ b/zephyr/Kconfig
@@ -56,4 +56,14 @@ config ZEPHYR_DP_SCHEDULER
 	  DP modules can be located in dieffrent cores than LL pipeline modules, may have
 	  different tick (i.e. 300ms for speech reccognition, etc.)
 
+config CHECK_INIT_PRIORITIES
+	bool "Build time initialization priorities check"
+	default n
+	help
+	  Check the build for initialization priority issues by comparing the
+	  initialization priority in the build with the device dependency
+	  derived from the devicetree definition.
+	  Fails the build on priority errors (dependent devices, inverted
+	  priority), see CHECK_INIT_PRIORITIES_FAIL_ON_WARNING to fail on
+	  warnings (dependent devices, same priority) as well.
 endif


pjdobrowolski:
Feature: https://github.com/zephyrproject-rtos/zephyr/pull/58388

zephyr/Kconfig
```diff
@@ -56,4 +56,14 @@ config ZEPHYR_DP_SCHEDULER
 	  DP modules can be located in dieffrent cores than LL pipeline modules, may have
 	  different tick (i.e. 300ms for speech reccognition, etc.)
 
+config CHECK_INIT_PRIORITIES
```
kv2019i, line 59: This is not the right way. We can just set this to 'n' in the in sof/app/prj.conf 

But we can't do this until the we update to latest Zephyr mainline and take in this patch
```
commit 058f256ea140026e46e458dd51036ea2c1f81b2a
Author: Fabio Baltieri <fabiobaltieri@google.com>
Date:   Fri Mar 10 17:26:35 2023 +0000

    Kconfig.zephyr: check priorities by default 
```

Then you'll get errors like this:
https://github.com/thesofproject/sof/actions/runs/5973406935/job/16205619385?pr=8095

kv2019i (changes requested):
Thanks @pjdobrowolski for catching this! I think we have to solve this with a prj.conf change when updating zephyr revision next time. We can't do this before or after.

zephyr/Kconfig
pjdobrowolski: I thought about that and we should do it as soon as possible. Is there any arguments against update?

kv2019i: @pjdobrowolski no obstacles, we can do one immediately. I just merged a 2200+ commit update yesterday, so we are very close to Zephyr mainline already so I don't expect many (new) issues beside this priority level check.

We need to investigate the priority check warnings as well and why those fail for us.

marc-hb: > I thought about that and we should do it as soon as possible.

We must do it in the _same commit_ where we update the Zephyr manifest and get this new feature. Neither before nor after to make sure every SOF commit builds.

If you need a local workaround in the mean time then `west config build.cmake-args -- -DCONFIG_CHECK_INIT_PRIORITIES=n` is probably the most convenient?

> I just merged a 2200+ commit update yesterday,

This sort of number can be deceiving; these could be almost all in drivers that we'll never use.

marc-hb (changes requested):
> We need to investigate the priority check warnings as well and why those fail for us.

I think we cannot disable this check without at least some vague understanding of why it fails. First of all, is the problem in SOF or in Zephyr?

Please also add a sample error message in the commit message so people can relate.

pjdobrowolski:
We have a problem in SOF with ssp, dmic and dma.
```

20:10:32,457 INFO  - ERROR: /soc/ssp@28000 POST_KERNEL 42 83 < /soc/dfpmccu@71b00/io0_domain POST_KERNEL 75 70
20:10:32,457 INFO  - ERROR: /soc/ssp@29000 POST_KERNEL 42 84 < /soc/dfpmccu@71b00/io0_domain POST_KERNEL 75 70
20:10:32,457 INFO  - ERROR: /soc/ssp@2a000 POST_KERNEL 42 85 < /soc/dfpmccu@71b00/io0_domain POST_KERNEL 75 70
20:10:32,457 INFO  - ERROR: /soc/ssp@2b000 POST_KERNEL 42 86 < /soc/dfpmccu@71b00/io0_domain POST_KERNEL 75 70
20:10:32,457 INFO  - ERROR: /soc/ssp@2c000 POST_KERNEL 42 87 < /soc/dfpmccu@71b00/io0_domain POST_KERNEL 75 70
20:10:32,458 INFO  - ERROR: /soc/ssp@2d000 POST_KERNEL 42 88 < /soc/dfpmccu@71b00/io0_domain POST_KERNEL 75 70
20:10:32,458 INFO  - ERROR: /soc/dmic0@10000 POST_KERNEL 70 74 < /soc/dfpmccu@71b00/hub_ulp_domain POST_KERNEL 75 73
20:10:32,458 INFO  - ERROR: /soc/dmic1@10000 POST_KERNEL 70 75 < /soc/dfpmccu@71b00/hub_ulp_domain POST_KERNEL 75 73
20:10:32,458 INFO  - ERROR: /soc/dma@72c00 POST_KERNEL 40 69 < /soc/dfpmccu@71b00/hst_domain POST_KERNEL 75 65
20:10:32,458 INFO  - ERROR: /soc/dma@7c000 POST_KERNEL 40 82 < /soc/dfpmccu@71b00/hub_ulp_domain POST_KERNEL 75 73
20:10:32,458 INFO  - ERROR: /soc/dma@7d000 POST_KERNEL 40 71 < /soc/dfpmccu@71b00/io0_domain POST_KERNEL 75 70
20:10:32,458 INFO  - ERROR: /soc/dma@7e000 POST_KERNEL 40 72 < /soc/dfpmccu@71b00/io0_domain POST_KERNEL 75 70
20:10:32,458 INFO  - ERROR: /soc/dma@72600 POST_KERNEL 40 67 < /soc/dfpmccu@71b00/hst_domain POST_KERNEL 75 65
20:10:32,458 INFO  - ERROR: /soc/dma@72400 POST_KERNEL 40 66 < /soc/dfpmccu@71b00/hst_domain POST_KERNEL 75 65
20:10:32,458 INFO  - ERROR: /soc/dma@72800 POST_KERNEL 40 68 < /soc/dfpmccu@71b00/hst_domain POST_KERNEL 75 65
```

kv2019i:
@marc-hb We do need to disable these checks blindly first as this breakage is already in Zephyr mainline and we cannot update Zephyr unless we fix this. The root issue of course needs to be addressed as well, but that would need to be with Zephyr PRs and this is anyways too late for SOF as the problematic commit was already merged to Zephyr main.

kv2019i:
Filed https://github.com/zephyrproject-rtos/zephyr/issues/61961 to track this, so we can unblock Zephyr/SOF CI and do the fix in parallel if needed. I could not assign you @pjdobrowolski  to the bug, but you are welcome to look.

kv2019i:
FYI @abonislawski @marcinszkudlinski @tmleman - seems we have some conflict in the Intel driver definitions that is now uncovered by this Kconfig check.

tmleman:
> seems we have some conflict in the Intel driver definitions that is now uncovered by this Kconfig check.

@kv2019i there is already a fix in the review https://github.com/zephyrproject-rtos/zephyr/pull/61955

marc-hb:
Interesting this issue affects only for MTL and LNL but neither IMX nor TGL
https://github.com/thesofproject/sof/actions/runs/6000180937/job/16271747199?pr=8107

marc-hb:
> there is already a fix in the review https://github.com/zephyrproject-rtos/zephyr/pull/61955

In other words this new check is finding real issues? Quick, let's disable it! :-)

marc-hb:
This does not seem to fail anymore:
https://github.com/thesofproject/sof/actions/runs/6006456965

@pjdobrowolski  close?

pjdobrowolski:
Fix confirmed.

PR #8097, opened 2023-08-25
host_common_params: adjust the values to the format 24/24

fkwasowi:
Change resolves timeout problem during offload of playback with 44.1 format

Currently DMA buffer size is aligned to 32. It should be:
rec_buff_size = ROUND_UP(rec_buff_size, min_buffer_size);
rec_buff_size = ROUND_UP(rec_buff_size, round_up_size);

Period_bytes should also not be aligned to 32 for 24/24 format

### DIFF
--- a/src/audio/host-zephyr.c
+++ b/src/audio/host-zephyr.c
@@ -768,6 +768,7 @@ int host_common_params(struct host_data *hd, struct comp_dev *dev,
 	uint32_t align;
 	int i, channel, err;
 	bool is_scheduling_source = dev == dev->pipeline->sched_comp;
+	uint32_t round_up_size;
 
 	/* host params always installed by pipeline IPC */
 	hd->host_size = params->buffer.size;
@@ -836,8 +837,10 @@ int host_common_params(struct host_data *hd, struct comp_dev *dev,
 	}
 
 	/* calculate DMA buffer size */
-	buffer_size = ALIGN_UP(period_bytes, align) * period_count;
-	buffer_size = MAX(buffer_size, ALIGN_UP(hd->ipc_host.dma_buffer_size, align));
+	round_up_size = (params->frame_fmt == SOF_IPC_FRAME_S24_3LE) ? (3 * align) : align;
+	buffer_size = ROUND_UP(period_bytes, round_up_size) * period_count;
+	if (hd->ipc_host.dma_buffer_size != 0)
+		buffer_size = ROUND_UP(hd->ipc_host.dma_buffer_size, buffer_size);
 
 	/* alloc DMA buffer or change its size if exists */
 	/*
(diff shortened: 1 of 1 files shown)


src/include/sof/math/numbers.h
```diff
@@ -24,6 +24,12 @@
 	typeof(alignment) __alignment = (alignment);	\
 	__size - (__size % __alignment);		\
 })
+
+#define ROUND_UP(size, alignment)				\
+	(((size) % (alignment) == 0)				\
+	? (size)						\
+	: ((size) - ((size) % (alignment)) + (alignment)))
+
```
lgirdwood, line 32: Need to be careful here and reuse the Zephyr maths macros as we've had some naming collisions in the past. @singalsu @kv2019i any reason why we are not using Zephyr C library maths macros today (as I expect we have some duplication).

lgirdwood (commented):
Good catch, just an open on which rounding we have to use.

marcinszkudlinski (changes requested):

src/include/sof/math/numbers.h
fkwasowi: The only difference is that zephyr's ROUND_UP modifies the value each time (I don't know if it's intentional?), while the ROUND_UP I added doesn't modify the value when it is divisible by the given argument. This difference caused me to get incorrect results. So I can modify ROUND_UP in zephyr?

marcinszkudlinski (approved):

iganakov (approved):

src/audio/host-zephyr.c
```diff
@@ -836,8 +837,9 @@ int host_common_params(struct host_data *hd, struct comp_dev *dev,
 	}
 
 	/* calculate DMA buffer size */
-	buffer_size = ALIGN_UP(period_bytes, align) * period_count;
-	buffer_size = MAX(buffer_size, ALIGN_UP(hd->ipc_host.dma_buffer_size, align));
+	round_up_size = (params->frame_fmt == SOF_IPC_FRAME_S24_3LE) ? (3 * align) : align;
+	buffer_size = ROUND_UP(period_bytes, round_up_size) * period_count;
+	buffer_size = MAX(buffer_size, ROUND_UP(hd->ipc_host.dma_buffer_size, buffer_size));
```
lyakh, line 842: the last line doesn't look right. Comparing to the current code it should be
```
buffer_size = MAX(buffer_size, ROUND_UP(hd->ipc_host.dma_buffer_size, round_up_size));
```
but if you really want to round up to a multiple of `buffer_size` then you don't need a `MAX()` - the rounding result won't be smaller, as long as `dma_buffer_size != 0`

lyakh (changes requested):

mwasko (approved):

src/audio/host-zephyr.c
```diff
@@ -836,8 +837,10 @@ int host_common_params(struct host_data *hd, struct comp_dev *dev,
 	}
 
 	/* calculate DMA buffer size */
-	buffer_size = ALIGN_UP(period_bytes, align) * period_count;
-	buffer_size = MAX(buffer_size, ALIGN_UP(hd->ipc_host.dma_buffer_size, align));
+	round_up_size = (params->frame_fmt == SOF_IPC_FRAME_S24_3LE) ? (3 * align) : align;
+	buffer_size = ROUND_UP(period_bytes, round_up_size) * period_count;
+	if (hd->ipc_host.dma_buffer_size != 0)
+		buffer_size = ROUND_UP(hd->ipc_host.dma_buffer_size, buffer_size);
```
lyakh, line 843: @fkwasowi it still looks strange to me - why we potentially almost double the buffer size. Say, if `buffer_size` is 4096 bytes and `dma_buffer_size` is just a bit larger, like 4100, this will set the buffer size to double the original size - to 8192 bytes, is this what is needed?

PR #8098, opened 2023-08-25
Scripts: run-mocks.sh: Change default config to avoid error

singalsu:
The script runs the unit tests natively as gcc compiled.  The tgph_defconfig is no more available, so use another configuration that is available in src/arch/xtensa/configs.


kv2019i (approved):
Aa, right, sorry, I missed that.

lgirdwood (approved):

PR #8101, opened 2023-08-25
Modify the echo reference capture pipeline

ranj063:
The new topology looks like this (Note that there is no capture PCM for echo ref anymore and SSP0-Codec DAI feeds into the google-rtc-ace module directly)
sof-mtl-max98357a-rt5682-ssp2-ssp0

This PR depends on the kernel PR https://github.com/thesofproject/linux/pull/4550


lgirdwood (commented):
More removals than additions which is always good.
@RDharageswari @yongzhi1 have you been able to validate with AEC ?
@cujomalainey fyi.

RDharageswari:
Hi Liam,
I am seeing some issues w.r.t third capture..Will continue to test and update

tools/topology/topology2/platform/intel/speaker-echo-ref.conf
```diff
@@ -33,41 +13,24 @@ Object.Pipeline {
...
-Object.Base.route [
-	{
-		source	"dai-copier.SSP.$SPEAKER_CODEC_NAME.capture"
-		sink	$ECHO_REF_COPIER_MODULE
-	}
-	{
-		source	$ECHO_REF_COPIER_MODULE
-		sink	"host-copier.$ECHO_REF_PCM_ID.capture"
-	}
-]
```
kv2019i, line 73: @ranj063  Not necessarily for this PR, but should we still have templates for both cases, echo ref to an AEC running on the DSP (like this case), and echo ref to user-space. I mean a given product will use only one option, not both, but in the past we've enabled both.

cujomalainey:
Also not directly related to this PR, but if this is targeted for ChromeOS, can we remove the EQIIR 14.1, we won't use it for now.

tools/topology/topology2/platform/intel/speaker-echo-ref.conf
```diff
@@ -1,25 +1,5 @@
-Define {
-	ECHO_REF_STREAM_NAME	"Echo reference"
```
cujomalainey, line 2: @perahgren are you ok if we fallback to AP AEC in this context?

cujomalainey (approved):
~cached comment~

plbossart:
> Also not directly related to this PR, but if this is targeted for ChromeOS, can we remove the EQIIR 14.1, we won't use it for now.

why? it's there on purpose to remove the DC component. You absolutely want this in Chrome on DMIC pipelines, it's an endpoint effect here, not something that the user should tweak.
Adding @singalsu for more comments.

cujomalainey:
> > Also not directly related to this PR, but if this is targeted for ChromeOS, can we remove the EQIIR 14.1, we won't use it for now.
> 
> why? it's there on purpose to remove the DC component. You absolutely want this in Chrome on DMIC pipelines, it's an endpoint effect here, not something that the user should tweak. Adding @singalsu for more comments.

If that is what it is for we would rather use the dcblock which is a exact duplicate of what we deploy in CRAS.

Also we have 0 users of it on intel platforms in the past 9 programs

plbossart:
> > > Also not directly related to this PR, but if this is targeted for ChromeOS, can we remove the EQIIR 14.1, we won't use it for now.
> > 
> > 
> > why? it's there on purpose to remove the DC component. You absolutely want this in Chrome on DMIC pipelines, it's an endpoint effect here, not something that the user should tweak. Adding @singalsu for more comments.
> 
> If that is what it is for we would rather use the dcblock which is a exact duplicate of what we deploy in CRAS.
> 
> Also we have 0 users of it on intel platforms in the past 9 programs

We really need @singalsu to comment, but I can tell you 100% of DMIC-based non-Chrome platforms use the EQIIR on the DMIC path. IIRC the EQIIR also boost the gain. It's a big surprise that Chrome topologies skipped it.

tools/topology/topology2/platform/intel/speaker-echo-ref.conf
ranj063: @kv2019i for aec in userspace, we would need the host copier and the PCM but we can't tie it to this topology. So I'd say we should leave it until.theres a need for it.

lgirdwood:
> Hi Liam, I am seeing some issues w.r.t third capture..Will continue to test and update

@RDharageswari ok, pls approve when you are happy with test results.

singalsu:
> > > > Also not directly related to this PR, but if this is targeted for ChromeOS, can we remove the EQIIR 14.1, we won't use it for now.
> > > 
> > > 
> > > why? it's there on purpose to remove the DC component. You absolutely want this in Chrome on DMIC pipelines, it's an endpoint effect here, not something that the user should tweak. Adding @singalsu for more comments.
> > 
> > 
> > If that is what it is for we would rather use the dcblock which is a exact duplicate of what we deploy in CRAS.
> > Also we have 0 users of it on intel platforms in the past 9 programs
> 
> We really need @singalsu to comment, but I can tell you 100% of DMIC-based non-Chrome platforms use the EQIIR on the DMIC path. IIRC the EQIIR also boosts the gain. It's a big surprise that Chrome topologies skipped it.

The generic Linux user space does not have enhancement for the microphone capture as CRAS in ChromeOS has. Without IIR and gain boost the capture would be very silent. We have got complains about it when it was lacking. The high-pass IIR is better than just gain for boosting the signal since suppressing the lowest frequencies gives more headroom for gain. Also without the IIR we see with various quality microphones components in PCs, that  the DC levels that the HW DCCOMP in DMIC is not capable to suppress so it's useful to have IIR or dcblock there to ensure capture is without DC offset. We've used for convenience the IIR EQ instead of dcblock + EQ.

I'm adding to tplg2 dcblock and a topology for hda-generic so using it will be possible in IPC4 and tplg2. We could consider it as default for HDA analog capture if the MCPS is lower than IIR (we should be able to optimize). The HDA headset codec also produces a long DC pulse in the capture start.

This nocodec topology is just for testing, it's not a reference, so the audio quality aspect with it is not critical. But we should have a good set of components included in my opinion since this undergoes a lot of testing. The hda-generic topologies are a reference for PCs so there we should have a good generic example.

singalsu:
@ranj063 I recommend to keep the IIR there and proceed. We can soon after switch it to dcblock when the conversion work is tested and merged.

I missed in my previous comment that this wasn't nocodec where we have similar looking capture pipelines. Also this PR is not related to IIR so it would be anyway other PR.

yongzhi1 (approved):

yongzhi1:
Think module-copier.18.1 is redundant, but the patch works fine.

ranj063:
> Think module-copier.18.1 is redundant, but the patch works fine.

@yongzhi1 yes lets fix it in a follow up

RDharageswari:
I am still seeing errors reported by FW. with kernel+topology and firmware for AEC. Working with Ranjani

ranj063:
> I am still seeing errors reported by FW. with kernel+topology and firmware for AEC. Working with Ranjani

@RDharageswari can you please check if the PR works now. I have removed the AEC blob for setting audio formats

RDharageswari (approved):
works good for me.

kv2019i (approved):
@ranj063 Feel free to merge if this is ready and kernel status is ok as well.

lgirdwood:
CI has had network issues. Rerun.

lgirdwood:
SOFCI TEST

lgirdwood:
@wszypelt @lrudyX CI seems to be in pending state, any chance someone could manually restart. Thanks !

wszypelt:
@lgirdwood in Internal Intel CI system, all green :)

tools/topology/topology2/include/pipelines/cavs/google-rtc-aec-capture.conf
```diff
@@ -89,26 +89,6 @@ Class.Pipeline."google-rtc-aec-capture" {
 			]
                 }
 
-		google-rtc-aec."1" {
-			Object.Control.bytes."1" {
```
singalsu, line 70: Does the real AEC version without mockup use the bytes control. Was this added just for the mockup? If the real AEC has bytes control shouldn't we have it in topology to be able to exercise it with about right ballpark size random bytes blob?

PR #8106, opened 2023-08-28
[DNM] Direct use of cached / not-cached shared memory alias instead of coherent.h in case of buffers

tobonex:
DO NOT MERGE - the PR will  be  spit to several PRs

- [x]  https://github.com/thesofproject/sof/pull/8166
- [x] https://github.com/thesofproject/sof/pull/8167

All info and context: https://github.com/thesofproject/sof/issues/8006

### DIFF
--- a/src/audio/aria/aria.c
+++ b/src/audio/aria/aria.c
@@ -169,8 +169,8 @@ static void aria_set_stream_params(struct comp_buffer *buffer,
 }
 
 static int aria_prepare(struct processing_module *mod,
-			struct sof_source __sparse_cache **sources, int num_of_sources,
-			struct sof_sink __sparse_cache **sinks, int num_of_sinks)
+			struct sof_source **sources, int num_of_sources,
+			struct sof_sink **sinks, int num_of_sinks)
 {
 	int ret;
 	struct comp_buffer *source, *sink;
--- a/src/audio/aria/aria.c
+++ b/src/audio/aria/aria.c
@@ -72,8 +72,8 @@ static int aria_algo_init(struct aria_data *cd, void *buffer_desc,
 }
 
 static inline void aria_process_data(struct processing_module *mod,
-				     struct audio_stream __sparse_cache *source,
-				     struct audio_stream __sparse_cache *sink,
+				     struct audio_stream *source,
+				     struct audio_stream *sink,
 				     size_t frames)
 {
 	struct aria_data *cd = module_get_private_data(mod);
@@ -153,19 +153,14 @@ static int aria_free(struct processing_module *mod)
 static void aria_set_stream_params(struct comp_buffer *buffer,
 				   struct processing_module *mod)
 {
-	struct comp_buffer __sparse_cache *buffer_c;
 	const struct ipc4_audio_format *audio_fmt = &mod->priv.cfg.base_cfg.audio_fmt;
 
-	buffer_c = buffer_acquire(buffer);
-
-	ipc4_update_buffer_format(buffer_c, audio_fmt);
+	ipc4_update_buffer_format(buffer, audio_fmt);
 #ifdef ARIA_GENERIC
-	audio_stream_init_alignment_constants(1, 1, &buffer_c->stream);
+	audio_stream_init_alignment_constants(1, 1, &buffer->stream);
 #else
-	audio_stream_init_alignment_constants(8, 1, &buffer_c->stream);
+	audio_stream_init_alignment_constants(8, 1, &buffer->stream);
 #endif
-
-	buffer_release(buffer_c);
 }
 
 static int aria_prepare(struct processing_module *mod,
--- a/src/audio/buffer.c
+++ b/src/audio/buffer.c
@@ -31,7 +31,6 @@ struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t flags, u
 				 bool is_shared)
 {
 	struct comp_buffer *buffer;
-	struct comp_buffer __sparse_cache *buffer_c;
 	void *stream_addr;
 
 	tr_dbg(&buffer_tr, "buffer_alloc()");
@@ -62,14 +61,11 @@ struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t flags, u
 	}
 
 	/* From here no more uncached access to the buffer object, except its list headers */
-	buffer_c = buffer_acquire(buffer);
-	audio_stream_set_addr(&buffer_c->stream, stream_addr);
-	buffer_init(buffer_c, size, caps);
+	audio_stream_set_addr(&buffer->stream, stream_addr);
+	buffer_init(buffer, size, caps);
 
-	audio_stream_set_underrun(&buffer_c->stream, !!(flags & SOF_BUF_UNDERRUN_PERMITTED));
-	audio_stream_set_overrun(&buffer_c->stream, !!(flags & SOF_BUF_OVERRUN_PERMITTED));
-
-	buffer_release(buffer_c);
+	audio_stream_set_underrun(&buffer->stream, !!(flags & SOF_BUF_UNDERRUN_PERMITTED));
+	audio_stream_set_overrun(&buffer->stream, !!(flags & SOF_BUF_OVERRUN_PERMITTED));
 
 	list_init(&buffer->source_list);
 	list_init(&buffer->sink_list);
@@ -77,7 +73,7 @@ struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t flags, u
 	return buffer;
 }
 
-void buffer_zero(struct comp_buffer __sparse_cache *buffer)
+void buffer_zero(struct comp_buffer *buffer)
 {
 	buf_dbg(buffer, "stream_zero()");
 
@@ -88,7 +84,7 @@ void buffer_zero(struct comp_buffer __sparse_cache *buffer)
 					audio_stream_get_size(&buffer->stream));
 }
 
-int buffer_set_size(struct comp_buffer __sparse_cache *buffer, uint32_t size, uint32_t alignment)
+int buffer_set_size(struct comp_buffer *buffer, uint32_t size, uint32_t alignment)
 {
 	void *new_ptr = NULL;
 
@@ -124,7 +120,7 @@ int buffer_set_size(struct comp_buffer __sparse_cache *buffer, uint32_t size, ui
 	return 0;
 }
 
-int buffer_set_params(struct comp_buffer __sparse_cache *buffer,
+int buffer_set_params(struct comp_buffer *buffer,
 		      struct sof_ipc_stream_params *params, bool force_update)
 {
 	int ret;
@@ -153,7 +149,7 @@ int buffer_set_params(struct comp_buffer __sparse_cache *buffer,
 	return 0;
 }
 
-bool buffer_params_match(struct comp_buffer __sparse_cache *buffer,
+bool buffer_params_match(struct comp_buffer *buffer,
 			 struct sof_ipc_stream_params *params, uint32_t flag)
 {
 	assert(params);
@@ -195,19 +191,7 @@ void buffer_free(struct comp_buffer *buffer)
 	coherent_free_thread(buffer, c);
 }
 
-/*
- * comp_update_buffer_produce() and comp_update_buffer_consume() send
- * NOTIFIER_ID_BUFFER_PRODUCE and NOTIFIER_ID_BUFFER_CONSUME notifier events
- * respectively. The only recipient of those notifications is probes. The
- * target for those notifications is always the current core, therefore notifier
- * callbacks will be called synchronously from notifier_event() calls. Therefore
- * we cannot pass unlocked buffer pointers to probes, because if they try to
- * acquire the buffer, that can cause a deadlock. In general locked objects
- * shouldn't be passed to potentially asynchronous contexts, but here we have no
- * choice but to use our knowledge of the local notifier behaviour and pass
- * locked buffers to notification recipients.
- */
-void comp_update_buffer_produce(struct comp_buffer __sparse_cache *buffer, uint32_t bytes)
+void comp_update_buffer_produce(struct comp_buffer *buffer, uint32_t bytes)
 {
 	struct buffer_cb_transact cb_data = {
 		.buffer = buffer,
@@ -229,8 +213,7 @@ void comp_update_buffer_produce(struct comp_buffer __sparse_cache *buffer, uint3
 
 	audio_stream_produce(&buffer->stream, bytes);
 
-	/* Notifier looks for the pointer value to match it against registration */
-	notifier_event(cache_to_uncache(buffer), NOTIFIER_ID_BUFFER_PRODUCE,
+	notifier_event(buffer, NOTIFIER_ID_BUFFER_PRODUCE,
 		       NOTIFIER_TARGET_CORE_LOCAL, &cb_data, sizeof(cb_data));
 
 #if CONFIG_SOF_LOG_DBG_BUFFER
@@ -246,7 +229,7 @@ void comp_update_buffer_produce(struct comp_buffer __sparse_cache *buffer, uint3
 #endif
 }
 
-void comp_update_buffer_consume(struct comp_buffer __sparse_cache *buffer, uint32_t bytes)
+void comp_update_buffer_consume(struct comp_buffer *buffer, uint32_t bytes)
 {
 	struct buffer_cb_transact cb_data = {
 		.buffer = buffer,
@@ -268,7 +251,7 @@ void comp_update_buffer_consume(struct comp_buffer __sparse_cache *buffer, uint3
 
 	audio_stream_consume(&buffer->stream, bytes);
 
-	notifier_event(cache_to_uncache(buffer), NOTIFIER_ID_BUFFER_CONSUME,
+	notifier_event(buffer, NOTIFIER_ID_BUFFER_CONSUME,
 		       NOTIFIER_TARGET_CORE_LOCAL, &cb_data, sizeof(cb_data));
 
 #if CONFIG_SOF_LOG_DBG_BUFFER
--- a/src/audio/buffer.c
+++ b/src/audio/buffer.c
@@ -27,7 +27,8 @@ DECLARE_SOF_RT_UUID("buffer", buffer_uuid, 0x42544c92, 0x8e92, 0x4e41,
 		 0xb6, 0x79, 0x34, 0x51, 0x9f, 0x1c, 0x1d, 0x28);
 DECLARE_TR_CTX(buffer_tr, SOF_UUID(buffer_uuid), LOG_LEVEL_INFO);
 
-struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t flags, uint32_t align)
+struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t flags, uint32_t align,
+				 bool is_shared)
 {
 	struct comp_buffer *buffer;
 	struct comp_buffer __sparse_cache *buffer_c;
@@ -42,11 +43,11 @@ struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t flags, u
 		return NULL;
 	}
 
-	/*
-	 * allocate new buffer, align the allocation size to a cache line for
-	 * the coherent API
-	 */
-	buffer = coherent_init_thread(struct comp_buffer, c);
+	/* allocate new buffer	 */
+	enum mem_zone zone = is_shared ? SOF_MEM_ZONE_RUNTIME_SHARED : SOF_MEM_ZONE_RUNTIME;
+
+	buffer = rzalloc(zone, 0, SOF_MEM_CAPS_RAM, sizeof(*buffer));
+
 	if (!buffer) {
 		tr_err(&buffer_tr, "buffer_alloc(): could not alloc structure");
 		return NULL;
@@ -69,18 +70,6 @@ struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t flags, u
 	audio_stream_set_overrun(&buffer_c->stream, !!(flags & SOF_BUF_OVERRUN_PERMITTED));
 
 	buffer_release(buffer_c);
-
-	/*
-	 * The buffer hasn't yet been marked as shared, hence buffer_release()
-	 * hasn't written back and invalidated the cache. Therefore we have to
-	 * do this manually now before adding to the lists. Buffer list
-	 * structures are always accessed uncached and they're never modified at
-	 * run-time, i.e. buffers are never relinked. So we have to make sure,
-	 * that what we have written into buffer's cache is in RAM before
-	 * modifying that RAM bypassing cache, and that after this cache is
-	 * re-loaded again.
-	 */
-	dcache_writeback_invalidate_region(uncache_to_cache(buffer), sizeof(*buffer));
 
 	list_init(&buffer->source_list);
 	list_init(&buffer->sink_list);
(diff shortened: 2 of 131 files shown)


lgirdwood:
@tobonex fyi - some conflicts (will block CI).

src/include/sof/audio/buffer.h
```diff
@@ -209,32 +209,27 @@ bool buffer_params_match(struct comp_buffer __sparse_cache *buffer,
 static inline void buffer_stream_invalidate(struct comp_buffer __sparse_cache *buffer,
 					    uint32_t bytes)
 {
-	if (!is_coherent_shared(buffer, c))
-		return;
-
-	audio_stream_invalidate(&buffer->stream, bytes);
+	if (buffer->is_shared)
+		audio_stream_invalidate(&buffer->stream, bytes);
```
lyakh, line 215: what about DMA buffers - on the host or DAI edge? Are they also handled by this class, if yes - their caches have to be managed, don't they?

marcinszkudlinski:
CI is green :) except for and code style checker (is it really a problem?):

```
CHECK: Macro argument reuse '_c' - possible side-effects?
#125: FILE: src/include/sof/coherent.h:87:
+#define CORE_CHECK_STRUCT(_c) { assert(!!(_c)->__is_shared == !!is_uncached(_c)); \
+				assert(cpu_get_id() == (_c)->__core || (_c)->__is_shared); }
```


and MTL sparse (not from this change):
```
+ ./sof/scripts/parse_sparse_output.sh
Reminder: to see ALL warnings you must as usual build _from scratch_
/zep_workspace/zephyr/drivers/dma/dma_intel_adsp_hda.c:190:17: error: typename in expression
/zep_workspace/zephyr/drivers/dma/dma_intel_adsp_hda.c:190:26: error: Expected ; at end of statement
/zep_workspace/zephyr/drivers/dma/dma_intel_adsp_hda.c:190:26: error: got rp
/zep_workspace/zephyr/drivers/dma/dma_intel_adsp_hda.c:199:17: error: typename in expression
/zep_workspace/zephyr/drivers/dma/dma_intel_adsp_hda.c:199:26: error: Expected ; at end of statement
/zep_workspace/zephyr/drivers/dma/dma_intel_adsp_hda.c:199:26: error: got wp
/zep_workspace/zephyr/drivers/dma/dma_intel_adsp_hda.c:190:17: error: undefined identifier 'uint32_t'
/zep_workspace/zephyr/drivers/dma/dma_intel_adsp_hda.c:191:37: error: undefined identifier 'rp'
/zep_workspace/zephyr/drivers/dma/dma_intel_adsp_hda.c:199:17: error: undefined identifier 'uint32_t'
/zep_workspace/zephyr/drivers/dma/dma_intel_adsp_hda.c:200:37: error: undefined identifier 'wp'
Error: Process completed with exit code 1.
```

src/include/sof/coherent.h
```diff
@@ -80,9 +80,23 @@ struct coherent {
 
 /* debug sharing amongst cores */
 #ifdef COHERENT_CHECK_NONSHARED_CORES
+
+#define CORE_CHECK_STRUCT_FIELD uint32_t __core; bool __is_shared
+#define CORE_CHECK_STRUCT_INIT(_c, is_shared) { (_c)->__core = cpu_get_id(); \
+						(_c)->__is_shared = is_shared; }
+#define CORE_CHECK_STRUCT(_c) { assert(!!(_c)->__is_shared == !!is_uncached(_c)); \
+				assert(cpu_get_id() == (_c)->__core || (_c)->__is_shared); }
```
lyakh, line 88: I'd put all the conditions in a single `assert()`

src/include/sof/audio/buffer.h
marcinszkudlinski: and they are in the very same way as before
This PR does not touch the audio data, just control structures
and in 100% of situations where it was "is_coherent_shared" now is "is_shared". +

For DAIs and DMA buffers the structures were **not** marked as shared:  coherent_shared_thread for buffers were called only once in the whole code - in comp_buffer_connect, called from ipc helpers (possible to mark buffer as shared)  and from IPC probes (not possible to mark as shared)

so cache operations were not performed at this point. And as CI is green - it is still working as DMA won't work at all using cached alias

It is, however, a very good question and I'll investigate it

src/ipc/ipc3/helper.c
```diff
@@ -572,6 +572,10 @@ static int ipc_comp_to_buffer_connect(struct ipc_comp_dev *comp,
 	tr_dbg(&ipc_tr, "ipc: comp sink %d, source %d -> connect", buffer->id,
 	       comp->id);
 
+	if (comp->core != buffer->cb->core) {
+		tr_err(&ipc_tr, "ipc: shared buffers are not supported for IPC3");
+		return -ENOTSUP;
+	}
```
kv2019i, line 573: This is interesting. I think this might be ok given I don't think we have any mainline users who need this feature. And if in future somebody needs this, one option to create all buffers as shared (as a build-time option to support this type of usage).

src/include/sof/coherent.h
kv2019i: @lyakh Given this is a debug option, woudln't separate asserts be nicer? You'd see which condition failed right away.

src/include/sof/audio/buffer.h
```diff
@@ -134,6 +134,8 @@ extern struct tr_ctx buffer_tr;
  */
 struct comp_buffer {
 	/* data buffer */
+	CORE_CHECK_STRUCT_FIELD;
+
```
kv2019i, line 138: This is very nice approach.

src/audio/dai-zephyr.c
```diff
@@ -977,8 +977,9 @@ int dai_common_params(struct dai_data *dd, struct comp_dev *dev,
 			return err;
 		}
 	} else {
+		/* allocate not shared buffer */
 		dd->dma_buffer = buffer_alloc(buffer_size, SOF_MEM_CAPS_DMA, 0,
-					      addr_align);
+					      addr_align, false);
```
kv2019i, line 949: We actually never used coherent.h to mark the handovers between DSP cores and DMA engines (like is done e.g. with DMA API in Linux -> https://www.kernel.org/doc/Documentation/DMA-API-HOWTO.txt ).

So while it's a bit confusing, I think this is good. I wonder we should reflect this in the code comments and other documentation that "not shared buffer" is actually "not shared buffer descriptor". The logic driven by whether object is shared or not, deals with buffer metadata and not the actual payload buffers. For casual observer, "shared buffer" certainly implies the statements here apply to the buffer itself (while that really is not the case here).

kv2019i (commented):
I think this looks good to go. Given the perf measurements back the approach, I'm good to go with this. The one ENOTSUP on IPC3 is an interesting, but AFAIK it won't impact any user at this point.

src/ipc/ipc3/helper.c
marcinszkudlinski:
```
static int ipc_comp_to_buffer_connect(struct ipc_comp_dev *comp,
				      struct ipc_comp_dev *buffer)
{
	tr_dbg(&ipc_tr, "ipc: comp sink %d, source %d -> connect", buffer->id,
	       comp->id);
#if CONFIG_INCOHERENT
	if (comp->core != buffer->cb->core) {
		tr_err(&ipc_tr, "ipc: shared buffers are not supported for IPC3 on incoherent architectures");
		return -ENOTSUP;
	}
#endif
	return comp_buffer_connect(comp->cd, comp->core, buffer->cb,
				   PPL_CONN_DIR_COMP_TO_BUFFER);
}

```

I'm not very familiar with other architectures, should I make this change?

src/audio/buffer.c
```diff
@@ -27,7 +27,8 @@ DECLARE_SOF_RT_UUID("buffer", buffer_uuid, 0x42544c92, 0x8e92, 0x4e41,
 		 0xb6, 0x79, 0x34, 0x51, 0x9f, 0x1c, 0x1d, 0x28);
 DECLARE_TR_CTX(buffer_tr, SOF_UUID(buffer_uuid), LOG_LEVEL_INFO);
 
-struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t flags, uint32_t align)
+struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t flags, uint32_t align,
```
btian1, line 30: this is for commit message, it is not POC now, could you change the commit message?

src/audio/buffer.c
```diff
@@ -303,30 +292,7 @@ void comp_update_buffer_consume(struct comp_buffer __sparse_cache *buffer, uint3
 void buffer_attach(struct comp_buffer *buffer, struct list_item *head, int dir)
 {
 	struct list_item *list = buffer_comp_list(buffer, dir);
-	struct list_item __sparse_cache *needs_sync;
```
btian1, line 306: FIXME: this is still a
 * problem with different cores.

after change, is there still a problem?

src/include/sof/audio/buffer.h
btian1: both ask and answer are clear, see below mails information in June:

On Tue, 6 Jun 2023, Tian, Baofeng wrote:

> Thanks, Guennadi, this is aligned.
>
> I am not sure whether close source have similar invalidate/writeback implementation or not, I don't remember old FW have this implementation.
> I don't know too much history about SOF, any particular reason to add invalidate and writeback for each buffer operation?
> Normally, it should be automatic(on ARM or old FW), only need add in some special case.

You have to synchronise cache for DMA.

Macin's implementation is quite clear and easy to understand, Guennadi's concern is dma buffer.
it need writeback or invalidate before sending/get DMA, question is: is it really required?

src/ipc/ipc-helper.c
```diff
@@ -173,13 +174,9 @@ int comp_buffer_connect(struct comp_dev *comp, uint32_t comp_core,
 			struct comp_buffer *buffer, uint32_t dir)
 {
 	/* check if it's a connection between cores */
-	if (buffer->core != comp_core) {
-		/* set the buffer as a coherent object */
-		coherent_shared_thread(buffer, c);
-
+	if (buffer->core != comp_core)
```
btian1, line 168: do we need add:
buffer->is_shared = true;
or check it here?

src/ipc/ipc3/helper.c
btian1: line 457 are aligned with line 575-578, buffer is not shared and if core is not same, then return with error. 

if add CONFIG_INCOHERENT limitation, then line 457 need check also to determine allocate shared or not shared buffer.

src/ipc/ipc4/helper.c
```diff
@@ -360,6 +361,7 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 	/* Pass IPC to target core if both modules has the same target core */
 	if (!cpu_is_me(source->ipc_config.core) && source->ipc_config.core == sink->ipc_config.core)
 		return ipc4_process_on_core(source->ipc_config.core, false);
+	is_shared = (source->ipc_config.core != sink->ipc_config.core);
```
btian1, line 357: shouldn't more formal?
is_shared = (source->ipc_config.core != sink->ipc_config.core）? true: false;

src/audio/aria/aria.c
```diff
@@ -170,7 +170,7 @@ static void aria_set_stream_params(struct comp_buffer *buffer,
 
 static int aria_prepare(struct processing_module *mod,
 			struct sof_source **sources, int num_of_sources,
-			struct sof_sink __sparse_cache **sinks, int num_of_sinks)
```
btian1, line 173: I would suggest squash this with previous one, since they do same thing.

src/audio/aria/aria.c
```diff
@@ -72,8 +72,8 @@ static int aria_algo_init(struct aria_data *cd, void *buffer_desc,
 }
 
 static inline void aria_process_data(struct processing_module *mod,
-				     struct audio_stream __sparse_cache *source,
-				     struct audio_stream __sparse_cache *sink,
+				     struct audio_stream *source,
+				     struct audio_stream *sink,
```
btian1, line 76: yes, you remove both source and sink here.

src/include/sof/coherent.h
btian1: @kv2019i , should add COHERENT_CHECK_NONSHARED_CORES to debug build?

src/include/sof/audio/buffer.h
lyakh: > and they are in the very same way as before
> This PR does not touch the audio data, just control structures

Not sure this is correct. Look at `audio_stream_invalidate()` - it synchronises audio data, not control structures. And before it was done always, now you move it to only be done for shared buffers

src/audio/dai-zephyr.c
lyakh: > The logic driven by whether object is shared or not, deals with buffer metadata and not the actual payload buffers. For casual observer, "shared buffer" certainly implies the statements here apply to the buffer itself (while that really is not the case here).

@kv2019i really? I actually think it's both. A buffer is marked as "shared" if it's between two components, executed on different cores. So, if the producer runs on core A, that core A will both access the buffer control structure and actually write data into the buffer. Then consumer on core B will access the control structure and read data on core B.

src/audio/buffer.c
marcinszkudlinski: I don't know what was the problem and if it disappeared or not, so I left the comment just in case

src/ipc/ipc4/helper.c
marcinszkudlinski: I believe it is quite clear, I would leave as is

src/audio/aria/aria.c
marcinszkudlinski: all prev sink/source commits were separate, but at this point you're right, separation of commits does not bring any value here.

src/include/sof/coherent.h
marcinszkudlinski: agree with @lyakh, 2 asserts are also more clear to read
As it is a debug option it does make any changes in production performance or code size

src/audio/buffer.c
btian1: after change, there is almost no code left in this function, ok, let's keep it, :)

src/ipc/ipc3/helper.c
kv2019i: I think @marcinszkudlinski this would be fine. We don't have any INCOHERENT, multi-core boards supported in upstream other than the Intel platforms we've moved to IPC4, so this does not effect anyone at the moment, but codifiying this limitation in code like this would be good.

src/include/sof/audio/buffer.h
marcinszkudlinski: @lyakh 
no. it was not, look like removed code:

```
static inline void buffer_stream_invalidate(struct comp_buffer __sparse_cache *buffer,
					    uint32_t bytes)
{
	if (!is_coherent_shared(buffer, c))
		return;

	audio_stream_invalidate(&buffer->stream, bytes);
}
```

src/ipc/ipc3/helper.c
marcinszkudlinski: @btian1 
>  then line 457 need check also to determine allocate shared or not shared buffer.

That's the whole point, at this moment there's no information about the modules the buffer will be bind to 
So we have we can either use all buffers as shared or limit usage

marcinszkudlinski:
```
	/* register buffer with pipeline */
#if CONFIG_USE_SHARED_BUFFERS_FOR_IPC3
	buffer = buffer_new(desc, true);
#else
	buffer = buffer_new(desc, false);
#endif
```
```
static int ipc_comp_to_buffer_connect(struct ipc_comp_dev *comp,
				      struct ipc_comp_dev *buffer)
{
	tr_dbg(&ipc_tr, "ipc: comp sink %d, source %d -> connect", buffer->id,
	       comp->id);

#if !CONFIG_USE_SHARED_BUFFERS_FOR_IPC3
	if (comp->core != buffer->cb->core) {
		tr_err(&ipc_tr, "ipc: shared buffers are not supported for IPC3");
		return -ENOTSUP;
	}
#endif
	return comp_buffer_connect(comp->cd, comp->core, buffer->cb,
				   PPL_CONN_DIR_COMP_TO_BUFFER);
}
```

src/ipc/ipc-helper.c
marcinszkudlinski: checking makes sense
adding buffer->is_shared = true;  is the very bad idea as it won't change memory alias the buffer is accessed by

src/ipc/ipc3/helper.c
marcinszkudlinski: as discussed - it is an overkill, just stay with a check with if coherent

src/ipc/ipc4/helper.c
lyakh: > shouldn't more formal? is_shared = (source->ipc_config.core != sink->ipc_config.core）? true: false;

@btian1 no. `!=` is a boolean operation, its result is `true` or `false`. What you're proposing is similar to
```
bool fn();

bool x = fn();
return (x == true) ? true : false;
```
instead of just
```
return fn();
```
I'd actually go a step further and remove parentheses above, but, well...

src/include/sof/coherent.h
lyakh: > @lyakh Given this is a debug option, woudln't separate asserts be nicer? You'd see which condition failed right away.

@kv2019i if it was spelled out in the code directly, then yes, separate asserts would be better. Here it's a single macro, so I thought using a single assert would fit better. But yes, you're right, with separate ones you get precise diagnostics immediately.

src/include/sof/audio/buffer.h
lyakh: indeed, you're right, sorry for the noise!

kv2019i (approved):
Marking with approve. If oithers agree, we can have another round (rebase if needed), hold of other PRs and get this merged (as this is very invasive in the codebase and affects a lot of in-flight PRs).

lgirdwood (approved):

src/include/sof/coherent.h
btian1: @marcinszkudlinski , please check code style, seems _c is not a good name.

btian1 (approved):

RanderWang (approved):

kv2019i:
@tobonex @marcinszkudlinski Seems we are good to go, can you rebase, then we'll halt PR merged for a time and do a timeboxed review effort to go through the gazillion small changes and merge the PR.

marcinszkudlinski:
rebase done

marcinszkudlinski:
the most important is to check [sparse: remove buffer_acquire / buffer_release](https://github.com/thesofproject/sof/pull/8106/commits/92079c6dbbe7d6dd599dc5e655c6d4cedcd7ec3b)

second huge commit - 
[sparse: remove all sparse_cache from struct sink/source api](https://github.com/thesofproject/sof/pull/8106/commits/3c61ff1670291a678f720ce8237cda4f0bc45272) is created 100% by editor "replace all" so is much less likely be incorrect

other commits have already been discussed and approved

tobonex:
While working on it I noted some significant changes, some may be outdated:

copier_ipcgtw_process() in copier_ipcgtw.c

mux_get_processing_function in mux_generic.c - buffer was not released here for some reason

dai_dma_cb in dai-zephyr.c err label is hanging after deleting buffer_release here, should we just break the list 
loop? If yes what is the best way to do that?

audio_steram.h:1074 above audio_stream_get_source - do we want to delete comment above it now?
also do we want to delete buffer_acquire and buffer_release definitions now?

host_legacy.c: host_common_update() - had to leave the logic here 
host_zephyr.c: host_common_update() - had to leave the logic here

module_adapter.c : check if I refactored those correctly: module_adapter_sink_src_prepare, 
module_adapter_sink_source_copy

module_adapter_audio_stream_type_copy - leaving variables as they are, just removing sparse_cache. I want to avoid changing multiple functions for this one.

generic.h module_source_info_acquire() : is removing it from coherent ok?

some rebased files to check(all commits):
multiband_drc.c
volume.c

marcinszkudlinski:
_mux_get_processing_function in mux_generic.c - buffer was not released here for some reason_
?????
image

tobonex:
> _mux_get_processing_function in mux_generic.c - buffer was not released here for some reason_ ????? image

this is dmux_get_processing, i meant the one above, mux_get_processing, sink_c was not released for some reason

marcinszkudlinski:
> this is dmux_get_processing, i meant the one above, mux_get_processing, sink_c was not released for some reason

right, there was no release in mux_get_processing_function
looking as those 2 were called I bet is just a forgotten release
image

as long as the buffer is was not shared such lack of release have had no negative effects

lgirdwood:
> 

ok, this is a problem. We need to play safer here. Can we

1) Create a new PR that will merge 1st and make coherent.h APIs a NOP i.e. turns then into empty static inline functions for all configurations.
2) Update this PR on top of PR 1 above and split patch per file. This could even mean breaking this PR into more PRs and incrementally removing coherent APIs.  i.e. its no longer a big bang.

This approach means we have remove the performance penalty from coherent and break the work down into more manageable chunk that we can bisect and individually revert patches if needed with no impact to other modules.

kv2019i:
Checked (potentially less coverage in upstream CI):
- src/audio/module_adapter/module/cadence.c
- src/audio/module_adapter/module/dts/dts.c
- src/audio/module_adapter/module/waves/waves.c

src/audio/copier/copier_ipcgtw.c
```diff
@@ -110,29 +109,26 @@ int copier_ipcgtw_process(const struct ipc4_ipcgtw_cmd *cmd,
...
-	} else {
+	if (!buf) {
 		/* NOTE: this func is called from IPC processing task and can be potentially
 		 * called before pipeline start even before buffer has been attached. In such
 		 * case do not report error but return 0 bytes available for GET_DATA and
 		 * 0 bytes free for SET_DATA.
 		 */
-		buf_c = NULL;
 		comp_warn(dev, "copier_ipcgtw_process(): no buffer found");
 	}
```
tobonex, line 119: about " copier_ipcgtw_process() in copier_ipcgtw.c" I just meant this if block,  
I did not substitute buf for buf_c here, as we do not need to NULL it now

tobonex:
> @tobonex Can you add the comments inline so it's easier to discuss specific points in the context of the code in question. The comment on dai-zephyr.c sounds a bit scary, but I guess you mean with "hanging" that the error label is left to code where it really is not needed anymore, right? Code is not actually hanging...

Yes, I'm making the comments now. And yes I meant the label is pointing at the end of block, not execution stopping.

marcinszkudlinski:
> @tobonex Can you add the comments inline so it's easier to discuss specific points in the context of the code in question. The comment on dai-zephyr.c sounds a bit scary, but I guess you mean with "hanging" that the error label is left to code where it really is not needed anymore, right? Code is not actually hanging...

I'm double checking dai-zephyr right now

src/audio/dai-zephyr.c
```diff
@@ -306,24 +299,24 @@ dai_dma_cb(struct dai_data *dd, struct comp_dev *dev, uint32_t bytes,
...
-					ret = dma_buffer_copy_from_no_consume(dma_buf, sink_c,
-									      converter[j], bytes);
-err:
-				buffer_release(sink_c);
+					ret = dma_buffer_copy_from_no_consume(dd->dma_buffer,
+									      sink, converter[j],
+									      bytes);
 			}
 		}
+err:
```
tobonex, line 307: Looking at it again I think I made a mistake here, I moved the label out of the list loop.  It seems the loop does not stop after an error, and just continues on to the next iteration, as error label was there to just release the buffer. Will it jump correctly if it points to end of a loop block?

jxstelter (approved):
Module_adapter, ipc

marcinszkudlinski:
closing, this PR has been split to several smaller parts

PR #8112, opened 2023-08-28
zephyr: edf-schedule: fix CONTAINER_OF type

fabiobaltieri:
The z_delayed_work field in struct task is a delayed work, so to go from k_work to it safely there's an extra step. Add a
k_work_delayable_from_work to fix that.

This is harmless right now because k_work is the first field in k_work_delayable but the extra step is the safe way of doing it.

### DIFF
--- a/zephyr/edf_schedule.c
+++ b/zephyr/edf_schedule.c
@@ -24,7 +24,8 @@ static K_THREAD_STACK_DEFINE(edf_workq_stack, 8192);
 
 static void edf_work_handler(struct k_work *work)
 {
-	struct task *task = CONTAINER_OF(work, struct task, z_delayed_work);
+	struct k_work_delayable *dwork = k_work_delayable_from_work(work);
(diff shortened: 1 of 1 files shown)


sofci:

gkbldcig:

marc-hb:
test this please

lyakh (approved):
currently `work` is at the top of `struct k_work_delayable`, so this works, but definitely let's do it correctly

kv2019i (approved):
Thanks @fabiobaltieri !

fabiobaltieri:
Thanks folks!

> currently `work` is at the top of `struct k_work_delayable`, so this works, but definitely let's do it correctly

Yeah that's correct and that'll likely never change, but there's load of other cases in the code base where structures may not be as stable and if it changes stuff would break in real bad way, so I'm working on adding type checking to `CONTAINER_OF` to avoid those situations but needs to sort all of these first.

kv2019i:
@wszypelt the internal Intel CI seems to have got stuck with this one, can you check?

PR #8120, opened 2023-08-29
imx: sdma: Add support for loading RAM script

dbaluta:
SDMA support loading firmware RAM scripts in order to enhance functionality. We will need this for supporting multi-fifo with upcoming MICFIL PDM driver.

Allow SDMA to load such scripts selected via CONFIG_HAVE_SDMA_FIRMWARE config.

A RAM script interface guarantees that there will be the following information provided:
- sdma_code, an array of bytes containting the actual script binary
- RAM_CODE_START_ADDR, location in RAM where to load the script
- RAM_CODE_SIZE, size of the sdma_code array.

Because RAM scripts are NXP proprietary binaries we won't make them public but offer the interface to integrate them with SOF in private builds.

We use for now sdma_script_code_imx7d_4_5.h header file which must be present in location indicated by CONFIG_SDMA_SCRIPT_CODE.

### DIFF
--- a/src/drivers/imx/sdma.c
+++ b/src/drivers/imx/sdma.c
@@ -265,6 +265,14 @@ static int sdma_upload_context(struct dma_chan_data *chan)
 			   sizeof(*pdata->ctx) / 4);
 }
 
+#if CONFIG_HAVE_SDMA_FIRMWARE
+static int sdma_load_firmware(struct dma *dma, void *buf, int addr, int size)
+{
+	return sdma_run_c0(dma->chan->dma, SDMA_CMD_C0_SET_PM,
+			   (uint32_t)buf, addr, size / 2);
+}
+#endif
+
 /* Below SOF related functions will be placed */
 
 static int sdma_probe(struct dma *dma)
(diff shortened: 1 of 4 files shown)


kv2019i (approved):

src/drivers/imx/sdma.c
```diff
@@ -265,6 +265,14 @@ static int sdma_upload_context(struct dma_chan_data *chan)
 			   sizeof(*pdata->ctx) / 4);
 }
 
+#if CONFIG_HAVE_SDMA_FIRMWARE
+static int sdma_load_firmware(struct dma *dma, void *buf, int addr, int size)
+{
+	return sdma_run_c0(dma->chan->dma, SDMA_CMD_C0_SET_PM,
+			   (uint32_t)buf, addr, size / 2);
```
btian1, line 272: Maybe compiler finally change /2 to >>1, maybe not, I suggest SOF eliminate /2 /4 /8 operation and use >> instead.

btian1 (approved):

lgirdwood (approved):

lgirdwood:
@wszypelt @lrudyX not expecting this NXP only patch to run on Intel config for CI. Good to merge ?

kv2019i:
Ignoring Intel CI fails as this code is not related in any way. FYI @wszypelt

PR #8124, opened 2023-08-30
Audio: ASRC: Split asrc.c to 2 files.

andrula-song:
Split asrc.c file to asrc_ipc3.c for IPC3 version and asrc_ipc4.c for IPC4 version.


src/audio/asrc/asrc_ipc3.c
```diff
@@ -0,0 +1,1048 @@
...
+/* Simple count value to prevent first delta timestamp
+ * from being input to low-pass filter.
+ */
+#define TS_STABLE_DIFF_COUNT	2
+
+/* Low pass filter coefficient for measured drift factor,
+ * The low pass function is y(n) = c1 * x(n) + c2 * y(n -1)
+ * coefficient c2 needs to be 1 - c1.
+ */
+#define COEF_C1		Q_CONVERT_FLOAT(0.01, 30)
```
singalsu, line 42: A lot of code looks generic and common. Why not keep asrc.c as the common module api client and split ipc3 and ipc4 specifics to separate source files. It's not sustainable to duplicate the code for every new IPC major version.

singalsu: It is also very possible that IPC3 operation is permanently broken due to other changes to SOF. Could consider erroring for IPC3 build and remove all specifics if we are not able to validate it.

lgirdwood (commented):
@andrula-song my expectation is that we will also have a common/core file that will be used by both IPC3/IPC4 and all SIMD varients.

btian1:
@andrula-song ， please refer #8026， it have all common in volume.c and ipc3 and ipc4.c only have ipc3 and ipc4 specific code.

src/audio/asrc/asrc_ipc3.c
andrula-song: since we will convert to module interface finally, and it would be much different. so I plan to split it first. I just compile it by docker for IPC3 and did not see any error, did not verify it on IPC3 yet

andrula-song:
> @andrula-song my expectation is that we will also have a common/core file that will be used by both IPC3/IPC4 and all SIMD varients.

okay, I can keep in one file.

src/audio/asrc/CMakeLists.txt
```diff
@@ -3,3 +3,8 @@
 add_local_sources(sof asrc.c asrc_farrow.c asrc_farrow_generic.c
 	asrc_farrow_hifi3.c)
 
+if(CONFIG_IPC_MAJOR_3)
+	add_local_sources(sof asrc_ipc3.c asrc_farrow.c asrc_farrow_generic.c asrc_farrow_hifi3.c)
+elseif(CONFIG_IPC_MAJOR_4)
+	add_local_sources(sof asrc_ipc4.c asrc_farrow.c asrc_farrow_generic.c asrc_farrow_hifi3.c)
+endif()
```
lyakh, line 10: can we do the split similar to #8026 to avoid code duplication?

lgirdwood:
> > @andrula-song my expectation is that we will also have a common/core file that will be used by both IPC3/IPC4 and all SIMD varients.
> 
> okay, I can keep in one file.

yes, but we should also have files for ipc3, ipc4 and hifi versions. Pls see the volume module in HEAD by @btian1 as this is what we want to do.

andrula-song:
> > > @andrula-song my expectation is that we will also have a common/core file that will be used by both IPC3/IPC4 and all SIMD varients.
> > 
> > 
> > okay, I can keep in one file.
> 
> yes, but we should also have files for ipc3, ipc4 and hifi versions. Pls see the volume module in HEAD by @btian1 as this is what we want to do.

okay, I will do this later, but my higher priority is to make asrc work, I'd like to close this PR temporary, and reopen this after it works.

PR #8130, opened 2023-08-30
topology2: cavs-rt5682: use macros for all PCM IDs

yongzhi1:
Make it less error-prone for maintenance.


jsarha (approved):
Always for a common definition on any value used in multiple places. Makes the topology sources more readable.

PR #8131, opened 2023-08-30
topology2: Modify the HDMI number to support the Rex variant

udaymb:
Modify NUM_HDMIS to 3 to support the Rex variant. Based on this change, SPK_ID and BT_ID is modified.


tools/topology/topology2/sof-ace-tplg/tplg-targets.cmake
```diff
@@ -53,9 +53,9 @@ DEEPBUFFER_D0I3_COMPATIBLE=true"
 "cavs-rt5682\;sof-mtl-max98357a-rt5682-ssp2-ssp0-2ch-pdm1\;PLATFORM=mtl,NUM_DMICS=2,\
 PDM0_MIC_A_ENABLE=0,PDM0_MIC_B_ENABLE=0,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,\
 DMIC0_PCM_ID=99,PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-mtl-max98357a-rt5682-2ch.bin,\
-HEADSET_SSP_DAI_INDEX=2,SPEAKER_SSP_DAI_INDEX=0,HEADSET_CODEC_NAME=SSP2-Codec,SPEAKER_CODEC_NAME=SSP0-Codec,\
-BT_NAME=SSP1-BT,BT_INDEX=1,BT_ID=8,BT_PCM_NAME=Bluetooth,INCLUDE_ECHO_REF=true,USE_CHAIN_DMA=true,\
-DEEPBUFFER_FW_DMA_MS=10,DEEPBUFFER_D0I3_COMPATIBLE=true"
+NUM_HDMI=3,HEADSET_SSP_DAI_INDEX=2,SPEAKER_SSP_DAI_INDEX=0,SPK_ID=6,HEADSET_CODEC_NAME=SSP2-Codec,\
```
plbossart, line 56: it's NUM_HDMIS=3 and this should be set for all devices.

@ranj063 @bardliao can we make this platform-dependent so that we don't have this sort of issues?

plbossart (changes requested):

tools/topology/topology2/sof-ace-tplg/tplg-targets.cmake
bardliao: > it's NUM_HDMIS=3 and this should be set for all devices.
> 
> @ranj063 @bardliao can we make this platform-dependent so that we don't have this sort of issues?

Yes, we can define NUM_HDMI in platform/intel/mtl.conf

kv2019i (approved):

tools/topology/topology2/sof-ace-tplg/tplg-targets.cmake
udaymb: @bardliao @plbossart - If this is included in the mtl.conf, other topologies would affect as below information still exists for one configuration
image

https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/intel/boards/sof_rt5682.c#L188

Should we need to update with NUM_DMICS=4 for these topologies?

lgirdwood:
@ranj063 @jsarha any comments ?

lgirdwood:
@plbossart good for you ? and do you want a followup PR for NUM_DMICS=4 ?

tools/topology/topology2/sof-ace-tplg/tplg-targets.cmake
ranj063: @udaymb if there's 1 topology with NUM_HDMIS=4, then thats the one that needs to override it at build time. Please make 3 the default and set it to 4 where needed

jsarha (approved):
So due to one less HDMI PCMs the BT ans speaker IDs are shifted down? SPK_ID = 7 default coming from some file and overridden here?

I guess this is Ok. However I'd urge to put all possible explanations into the commit message. These topo2 makefile changes are hardly self explanatory.

bardliao:
The code looks good. However, we need to revisit the machine driver. We have SOF_RT5682_NUM_HDMIDEV(4) for some Rex variants. 
For example,
https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/intel/boards/sof_rt5682.c#L188

We need to make sure machine driver and topology are aligned.

lgirdwood:
> The code looks good. However, we need to revisit the machine driver. We have SOF_RT5682_NUM_HDMIDEV(4) for some Rex variants. For example, https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/intel/boards/sof_rt5682.c#L188
> 
> We need to make sure machine driver and topology are aligned.

@bardliao is this something you do in the machine driver ?

tools/topology/topology2/sof-ace-tplg/tplg-targets.cmake
udaymb: Updated the commit

udaymb:
> So due to one less HDMI PCMs the BT ans speaker IDs are shifted down? SPK_ID = 7 default coming from some file and overridden here?
> 
> I guess this is Ok. However I'd urge to put all possible explanations into the commit message. These topo2 makefile changes are hardly self explanatory.

@jsarha - Updated the cmake file and commit message accordingly

bardliao:
> > The code looks good. However, we need to revisit the machine driver. We have SOF_RT5682_NUM_HDMIDEV(4) for some Rex variants. For example, https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/intel/boards/sof_rt5682.c#L188
> > We need to make sure machine driver and topology are aligned.
> 
> @bardliao is this something you do in the machine driver ?

No, that was not me.

bardliao (approved):

plbossart (approved):

ranj063 (approved):

PR #8132, opened 2023-08-30
Preparation for introducing the SOF ALSA plugin

ranj063:
The SOF ALSA plugin will be used for running SOF topology pipelines to run on the host for development/debugging purposes.
The changes in this PR include preparatory patches for:
 - Building shared libraries for SOF modules
 - Posix headers for host build
 - Addition of missing platform definitions for library build

### DIFF
--- a/posix/include/rtos/alloc.h
+++ b/posix/include/rtos/alloc.h
@@ -1,0 +1,172 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright(c) 2016 Intel Corporation. All rights reserved.
+ *
+ * Author: Liam Girdwood <liam.r.girdwood@linux.intel.com>
+ *         Keyon Jie <yang.jie@linux.intel.com>
+ */
+
+/**
+ * \file xtos/include/rtos/alloc.h
+ * \brief Memory Allocation API definition
+ * \author Liam Girdwood <liam.r.girdwood@linux.intel.com>
+ * \author Keyon Jie <yang.jie@linux.intel.com>
+ */
+
+#ifndef __SOF_LIB_ALLOC_H__
+#define __SOF_LIB_ALLOC_H__
+
+#include <rtos/bit.h>
+#include <rtos/string.h>
+#include <sof/trace/trace.h>
+#include <user/trace.h>
+
+#include <stddef.h>
+#include <stdint.h>
+
+/** \addtogroup alloc_api Memory Allocation API
+ *  @{
+ */
+
+/**
+ * \brief Heap Memory Zones
+ *
+ * The heap has three different zones from where memory can be allocated :-
+ *
+ * 1) System Zone. Fixed size heap where alloc always succeeds and is never
+ * freed. Used by any init code that will never give up the memory.
+ *
+ * 2) System Runtime Zone. Heap zone intended for runtime objects allocated
+ * by the kernel part of the code.
+ *
+ * 3) Runtime Zone. Main and larger heap zone where allocs are not guaranteed to
+ * succeed. Memory can be freed here.
+ *
+ * 4) Buffer Zone. Largest heap zone intended for audio buffers.
+ *
+ * 5) Runtime Shared Zone. Similar to Runtime Zone, but content may be used and
+ * fred from any enabled core.
+ *
+ * 6) System Shared Zone. Similar to System Zone, but content may be used from
+ * any enabled core.
+ *
+ * See platform/memory.h for heap size configuration and mappings.
+ */
+enum mem_zone {
+	SOF_MEM_ZONE_SYS = 0,		/**< System zone */
+	SOF_MEM_ZONE_SYS_RUNTIME,	/**< System-runtime zone */
+	SOF_MEM_ZONE_RUNTIME,		/**< Runtime zone */
+	SOF_MEM_ZONE_BUFFER,		/**< Buffer zone */
+	SOF_MEM_ZONE_RUNTIME_SHARED,	/**< Runtime shared zone */
+	SOF_MEM_ZONE_SYS_SHARED,	/**< System shared zone */
+};
+
+/** \name Heap zone flags
+ *  @{
+ */
+
+/** \brief Indicates that original content should not be copied by realloc. */
+#define SOF_MEM_FLAG_NO_COPY	BIT(1)
+/** \brief Indicates that if we should return uncached address. */
+#define SOF_MEM_FLAG_COHERENT  BIT(2)
+
+/** @} */
+
+/**
+ * Allocates memory block.
+ * @param zone Zone to allocate memory from, see enum mem_zone.
+ * @param flags Flags, see SOF_MEM_FLAG_...
+ * @param caps Capabilities, see SOF_MEM_CAPS_...
+ * @param bytes Size in bytes.
+ * @return Pointer to the allocated memory or NULL if failed.
+ *
+ * @note Do not use for buffers (SOF_MEM_ZONE_BUFFER zone).
+ *       Use rballoc(), rballoc_align() to allocate memory for buffers.
+ */
+void *rmalloc(enum mem_zone zone, uint32_t flags, uint32_t caps, size_t bytes);
+
+/**
+ * Similar to rmalloc(), guarantees that returned block is zeroed.
+ *
+ * @note Do not use  for buffers (SOF_MEM_ZONE_BUFFER zone).
+ *       rballoc(), rballoc_align() to allocate memory for buffers.
+ */
+void *rzalloc(enum mem_zone zone, uint32_t flags, uint32_t caps, size_t bytes);
+
+/**
+ * Allocates memory block from SOF_MEM_ZONE_BUFFER.
+ * @param flags Flags, see SOF_MEM_FLAG_...
+ * @param caps Capabilities, see SOF_MEM_CAPS_...
+ * @param bytes Size in bytes.
+ * @param alignment Alignment in bytes.
+ * @return Pointer to the allocated memory or NULL if failed.
+ */
+void *rballoc_align(uint32_t flags, uint32_t caps, size_t bytes,
+		    uint32_t alignment);
+
+/**
+ * Similar to rballoc_align(), returns buffer aligned to PLATFORM_DCACHE_ALIGN.
+ */
+static inline void *rballoc(uint32_t flags, uint32_t caps, size_t bytes)
+{
+	return rballoc_align(flags, caps, bytes, PLATFORM_DCACHE_ALIGN);
+}
+
+/**
+ * Changes size of the memory block allocated from SOF_MEM_ZONE_BUFFER.
+ * @param ptr Address of the block to resize.
+ * @param flags Flags, see SOF_MEM_FLAG_...
+ * @param caps Capabilities, see SOF_MEM_CAPS_...
+ * @param bytes New size in bytes.
+ * @param old_bytes Old size in bytes.
+ * @param alignment Alignment in bytes.
+ * @return Pointer to the resized memory of NULL if failed.
+ */
+void *rbrealloc_align(void *ptr, uint32_t flags, uint32_t caps, size_t bytes,
+		      size_t old_bytes, uint32_t alignment);
+
+/**
+ * Similar to rballoc_align(), returns resized buffer aligned to
+ * PLATFORM_DCACHE_ALIGN.
+ */
+static inline void *rbrealloc(void *ptr, uint32_t flags, uint32_t caps,
+			      size_t bytes, size_t old_bytes)
+{
+	return rbrealloc_align(ptr, flags, caps, bytes, old_bytes,
+			       PLATFORM_DCACHE_ALIGN);
+}
+
+/**
+ * Frees the memory block.
+ * @param ptr Pointer to the memory block.
+ *
+ * @note Blocks from SOF_MEM_ZONE_SYS cannot be freed, such a call causes
+ *       panic.
+ */
+void rfree(void *ptr);
+
+/**
+ * Allocates memory block from the system heap reserved for the specified core.
+ * @param core Core id.
+ * @param bytes Size in bytes.
+ */
+void *rzalloc_core_sys(int core, size_t bytes);
+
+/**
+ * Calculates length of the null-terminated string.
+ * @param s String.
+ * @return Length of the string in bytes.
+ */
+int rstrlen(const char *s);
+
+/**
+ * Compares two strings, see man strcmp.
+ * @param s1 First string to compare.
+ * @param s2 Second string to compare.
+ * @return See man strcmp.
+ */
+int rstrcmp(const char *s1, const char *s2);
+
+/** @}*/
+
+#endif /* __SOF_LIB_ALLOC_H__ */
--- a/posix/include/rtos/atomic.h
+++ b/posix/include/rtos/atomic.h
@@ -1,0 +1,39 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright(c) 2018 Intel Corporation. All rights reserved.
+ *
+ * Author: Liam Girdwood <liam.r.girdwood@linux.intel.com>
+ */
+
+#ifndef __SOF_ATOMIC_H__
+#define __SOF_ATOMIC_H__
+
+#include <arch/atomic.h>
+#include <stdint.h>
+
+static inline void atomic_init(atomic_t *a, int32_t value)
+{
+	arch_atomic_init(a, value);
+}
+
+static inline int32_t atomic_read(const atomic_t *a)
+{
+	return arch_atomic_read(a);
+}
+
+static inline void atomic_set(atomic_t *a, int32_t value)
+{
+	arch_atomic_set(a, value);
+}
+
+static inline int32_t atomic_add(atomic_t *a, int32_t value)
+{
+	return arch_atomic_add(a, value);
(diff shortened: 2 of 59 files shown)


posix/include/rtos/alloc.h
```diff
@@ -0,0 +1,172 @@
+/* SPDX-License-Identifier: BSD-3-Clause
```
btian1, line 1: not able to include those headers? duplicate is not a good option.

ranj063: @btian1 this is for host build for testbench/tplg parser/plugin etc where zephyr/xtos doesnt feature

btian1: so you mean plugin build does not have xtos/zephyr included?

tools/testbench/common_test.c
```diff
@@ -264,14 +264,14 @@ int tb_pipeline_params(struct testbench_prm *tp, struct ipc *ipc, struct pipelin
...
-	if (debug)
+	if (host_trace_level >= LOG_LEVEL_DEBUG)
 		printf("debug: %s", message);
 }
 
 /* enable trace in testbench */
 void tb_enable_trace(bool enable)
 {
-	test_bench_trace = enable;
+	host_trace_level = enable;
```
lyakh, line 274: confused by this commit: (1) where `host_trace_level` is defined (apparently, it is defined somewhere, since this all works, or maybe it doesn't need to be, and (2) doesn't the above mean, that we set `host_trace_level` to either `LOG_LEVEL_ERROR` or to 0, which isn't even a valid error level?

src/platform/library/lib/trace.c
```diff
@@ -12,8 +12,7 @@
 #include <sof/trace/dma-trace.h>
 
 /* enable trace by default in testbench */
-int test_bench_trace = 1;
-int debug;
+int host_trace_level = 1;
```
lyakh, line 15: should be merged with the previous commit to support bisection. Also wouldn't it be better to use `LOG_LEVEL_ERROR`?

posix/include/rtos/alloc.h
lyakh: we do already use SOF headers for library (testbench, cmocka) builds, cannot we also use them for this? Really should avoid duplicating - a sure path to breakage and confusion...

src/audio/Kconfig
```diff
@@ -684,4 +684,11 @@ config WRAP_ACTUAL_POSITION
 	  It is not necessary that on wrap, the buffer position would be zero.At wrap,
 	  in some cases based on the period size, the frame may not exactly be at the
 	  end of the buffer and roll over for some bytes from the beginning of the buffer.
+
+config COMP_MODULES_SO
```
lyakh, line 688: we might soon need a similar option to build SOF modules for loading to the DSP, can we maybe add "HOST" somewhere to this option to distinguish between the two?

posix/include/rtos/alloc.h
kv2019i: But which set? I think this approach has merits as well. I think we have three options:
1) XTOS build with CONFIG_LIBRARY (use sof/xtos/rtos)
2) Zephyr build with posix as arch (use sof/zephyr/rtos)
3) Native library build (use sof/posix/rtos)

I think (3) involves code duplication, but for a pure library build (that can be linked to any Linux application), I think this approach is cleanest. Especially if we longterm remove the the XTOS build. Using Zephyr posix build is one option, but I don't think that is really ideal for the library use (like an ALSA plugin, or exposing audio modules via some other framerwork like LV2).

posix/include/rtos/atomic.h
```diff
@@ -0,0 +1,39 @@
...
+
+#ifndef __SOF_ATOMIC_H__
+#define __SOF_ATOMIC_H__
+
+#include <arch/atomic.h>
+#include <stdint.h>
+
+static inline void atomic_init(atomic_t *a, int32_t value)
+{
+	arch_atomic_init(a, value);
```
kv2019i, line 16: I don't think this arch_foo() layer makes sense. If we pull the rtos layer out from XTOS (or Zephyr), then we don't need any dependency to a separate arch layer. I.e. this should use POSIX primitives (or Linux extensions) directly.

src/arch/host/include/arch/debug/panic.h
```diff
@@ -5,7 +5,7 @@
  * Author: Tomasz Lauda <tomasz.lauda@linux.intel.com>
  */
 
-#ifdef __XTOS_RTOS_PANIC_H__
+#ifdef __POSIX_RTOS_PANIC_H__
```
kv2019i, line 8: Hmm, continuing with previous, @lrgirdwo @ranj063 , I don't see a point in arch layer if we go for indepdent "rtos" implementation ("XTOS", "Zephyr" and "posix").

src/platform/library/include/platform/lib/memory.h
```diff
@@ -20,6 +20,11 @@
 
 struct sof;
 
+#define EBB_BANKS_IN_SEGMENT		32
+
+#define PLATFORM_LPSRAM_EBB_COUNT	1
+#define PLATFORM_HPSRAM_EBB_COUNT	1
+
```
kv2019i, line 27: We really shouldn't need these anymore.

src/audio/Kconfig
kv2019i: Or LIBRARY something, that's what this is in the end, right?

kv2019i (commented):
Looks pretty good. One bigger issue is the arch layer. If we decide to go with a top-level rtos abstraction (i.e. library is an alternative to a XTOS or Zephyr build), I don't see a point in using the arch layer. We can proceed to merge this version and remove it later, but I'd like to get a consensus before we proceed. Zephyr posix build is of course one option, but not sure whether it works as a plugin library (probably less work to have a native library build for Linux).

tools/testbench/common_test.c
lgirdwood: ```debug``` colides easily and ```test_bench_trace``` tightly coupled any user code to testbench.

src/platform/library/lib/trace.c
lgirdwood: ack fo bisect.

posix/include/rtos/alloc.h
lgirdwood: I would move list.h and other generic (non RTOS api) level headers to a sof/lib (but this could be another patch).

src/arch/host/include/arch/debug/panic.h
lgirdwood: We are stuck a bit today whilst xtos is still being used, i.e. difficult to extract until more xtos related code can be removed (and this depends on progress with Zephyr integration for all platforms).

posix/include/rtos/alloc.h
kv2019i: @lgirdwood Right, but we don't need XTOS arch code if plugin uses its own rtos implementation.  sof/posix/rtos should be independent of sof/src/arch . cmocka uses it, but it was done before we had a rtos abstraction.

marc-hb (commented):
I don't think "host" is a good name for what is really just SIMULation. 

tl;dr: "host" = best avoided. Use "simulation" instead.

**Simulation** is the term used by Zephyr, XTensa and the testbench:
- https://docs.zephyrproject.org/latest/boards/posix/index.html
- https://docs.zephyrproject.org/latest/boards/xtensa/xt-sim/doc/index.html
- https://www.cadence.com/en_US/home/tools/ip/tensilica-ip/technologies.html
- https://thesofproject.github.io/latest/developer_guides/testbench/index.html

"Host" is already used for something different: Host is where the Linux / Windows driver run:
- https://thesofproject.github.io/latest/introduction/index.html

Of course "host" is typically the same system for many developers because why would you inconveniently use two different systems when the same workstation can do everything. Yet it absolutely does not have to be the same system because these two "hosts" are conceptually two totally different things. The testbench and the fuzzer have many "driver-free" users. Conversely, production "hosts" use drivers but never any testbench or fuzzer. For Yocto users these two "hosts" are different systems too.

Using the same word "host" for everything just because some people use the same workstation for everything maintains a bad confusion.

Making things even worse, "host" and "target" are also widely misunderstood toolchain concepts:
https://mesonbuild.com/Cross-compilation.html
> In a nutshell, the typical mistake assumes that the terms build, host and target refer to some fixed positions whereas they're actually relative to where the current compiler is running. Think of host as a child of the current compiler and target as an optional grand-child. Compilers don't change their terminology when they're creating another compiler.

lgirdwood:
> I don't think "host" is a good name for what is really just SIMULation.
> 
> tl;dr: "host" = best avoided. Use "simulation" instead.
> 
> **Simulation** is the term used by Zephyr and XTensa:

@marc-hb Not following why you think this is a simulation ? This is an ALSA plugin that can run SOF pipeline topologies and modules in userspace.

marc-hb:
> @marc-hb Not following why you think this is a simulation ? This is an ALSA plugin that can run SOF pipeline topologies and modules in userspace.

Because it does not run on the DSP like all the other simulations I listed above.

posix/include/rtos/atomic.h
ranj063: @kv2019i but the arch layer used here would use the gcc atomic builtins as defined in the host arch atomic.h. Isnt that good enough?

ranj063:
> > @marc-hb Not following why you think this is a simulation ? This is an ALSA plugin that can run SOF pipeline topologies and modules in userspace.
> 
> Because it does not run on the DSP like all the other simulations I listed above.

This is not sumlation @marc-hb and host matches exactly with your quoted definition. These modules can be used to run pipelines on the host just like the linux/windows driver.

lgirdwood:
> > @marc-hb Not following why you think this is a simulation ? This is an ALSA plugin that can run SOF pipeline topologies and modules in userspace.
> 
> Because it does not run on the DSP like all the other simulations I listed above.

@marc-hb You are probably thinking of the POSIX Zephyr mode where SOF runs under Zephyr as a POSIX application (it could be considered a simulation) that we use for fuzzing. This plugin is a userspace executable that runs generic SOF pipeline and module code.

marc-hb:
> @marc-hb You are probably thinking of ...

There are different cases and I'm thinking "host" suits none of them.

In some places in the code we need a name for "not running on DSP". But negations don't make good names; I don't think anyone wants something like CONFIG_NOT_ON_THE_DSP.

So instead we tend to see things like HOST_MODULES, `host_trace_level` and `arch/host/include`. But which host!?! There are _many different hosts_, so "host" is very confusing. 

=> Specific "hosts" should be named with their specific name, e.g.: fuzzer, testbench, zephyr POSIX, ALSA plugin, etc. For Linux/Windows "hosts" it's probably too late and I don't have any better name.

=> CONFIG_NOT_ON_THE_DSP should be called CONFIG_SIMULATION because it's the term already used everywhere, see all the examples above.

lyakh:
> CONFIG_NOT_ON_THE_DSP should be called CONFIG_SIMULATION because it's the term already used everywhere, see all the examples above.

@marc-hb I disagree too. All the above examples are ok with "simulation" because they are simulations. Whereas actually running in a user configuration "on the host," being part of the audio processing pipeline and performing useful work - that isn't a simulation any more. Historically maybe "host" used to mean "native" to the "build host." I.e. not cross-compiled, but now it certainly doesn't have to be the case. E.g. in case of ARM if we build ALSA plugins, we'll build them to run on the ARM "host" and we might well cross-compile them. So in case of ARM we have 3 architectures: build / test / simulate, DSP, and OS "host." Then, yes, we could use "simulation" for the test system, like the CI, which is typically x86*, the DSP can be Xtensa or ARM, and the "OS host" can be x86* or ARM.

posix/include/rtos/atomic.h
kv2019i: @ranj063  This surely works and to be clear, I'm ok to merge this as an intermedia step, but I want to get aligned conceptually that there is no need to depend on the XTOS arch layer.

src/include/sof/trace/trace.h
```diff
@@ -233,19 +233,26 @@ void _log_sofdict(log_func_t sofdict_logf, bool atomic, const void *log_entry,
 
 #ifdef CONFIG_LIBRARY
 
-extern int test_bench_trace;
+#include <sys/time.h>
+
+/* trace level used on host configurations */
+extern int host_trace_level;
```
kv2019i, line 239: I don't think we need to get caught up with the naming (so not a showstopper for me), but as a option, we could call this "extern int sof_library_trace_level".

I think this is logical. We have a build target to build SOF a library, not an application, and when built this way, certain extra interfaces are exposed to do things normally done internally by "SOF-the-application". Like setting trace levels.

src/audio/Kconfig
```diff
@@ -684,4 +684,12 @@ config WRAP_ACTUAL_POSITION
 	  It is not necessary that on wrap, the buffer position would be zero.At wrap,
 	  in some cases based on the period size, the frame may not exactly be at the
 	  end of the buffer and roll over for some bytes from the beginning of the buffer.
+
+config COMP_HOST_MODULES_SO
+	bool "Build modules as shared objects"
```
kv2019i, line 689: COMP_MODULE_SHARED_LIBRARY_BUILD

"Build SOF modules as shared libraries"

ranj063: updated now @kv2019i

marc-hb:
> Whereas actually running in a user configuration "on the host," being part of the audio processing pipeline and performing useful work - that isn't a simulation any more. 

You could call it "hybrid emulation" if you want to make a difference. Note the difference between emulation and simulation depends on whom you ask. Either term makes is clear that the code does not run on the DSP.

"Host" is clearly not on the DSP either but it makes you wonder "which host(s)?" which may or may not be relevant. Super confusing.

> E.g. in case of ARM if we build ALSA plugins, we'll build them to run on the ARM "host" and we might well cross-compile them. So in case of ARM we have 3 architectures: build / test / simulate, DSP, and OS "host." Then, yes, we could use "simulation" for the test system, like the CI, which is typically x86*, the DSP can be Xtensa or ARM, and the "OS host" can be x86* or ARM.

Toolchains matter of course but I'm not sure architecture examples really help clarifying... Anyway good demonstration that the word "host" is now void of any meaning, thanks this is my main point! I suspect "host" is appealing because it's.... short.

RanderWang (approved):
LGTM

src/include/sof/trace/trace.h
```diff
@@ -233,19 +233,26 @@ void _log_sofdict(log_func_t sofdict_logf, bool atomic, const void *log_entry,
...
-		fprintf(stderr, msg, strrchr(__FILE__, '/') + 1,\
-			__LINE__, ##__VA_ARGS__);	\
-		fprintf(stderr, "\n");							\
-	}										\
+do {								\
+	(void)ctx;						\
+	(void)id_1;						\
+	(void)id_2;						\
+	struct timeval tv;					\
+	char *msg = "(%s:%d) " format;				\
```
lyakh, line 248: Why not put both these variables under the `if`? In practical terms the difference isn't huge - also this way their scope is limited to the "pseudo-loop," but having them under the `if` would be more logical

tools/testbench/common_test.c
lyakh: @lgirdwood sorry, I mean it's a wrong assignment IMHO. Before `test_bench_trace` was effectively a boolean variable - it only took values of 0 and 1. Now `host_trace_level` is a printing verbosity **level** taking a value between `LOG_LEVEL_CRITICAL` (1) and `LOG_LEVEL_VERBOSE` (4). So this assignment should be fixed.

posix/include/rtos/alloc.h
lyakh: not hugely important, but calling these "posix headers" is a bit confusing - don't think all these headers are POSIX

lyakh (changes requested):

kv2019i (approved):
I'm good to merge. We probably need to refactor rtos layer more, but this is easier done in a follow-up. Having this merged, allows more people to contribute and this doesn't interfere with other builds of SOF, so I don't see much to gain by keeping this in review for much logner.

lyakh:
> I'm good to merge. We probably need to refactor rtos layer more, but this is easier done in a follow-up. Having this merged, allows more people to contribute and this doesn't interfere with other builds of SOF, so I don't see much to gain by keeping this in review for much logner.

@kv2019i I think https://github.com/thesofproject/sof/pull/8132#discussion_r1316793944 is a bug and should be fixed

lgirdwood (approved):

tools/testbench/common_test.c
lgirdwood: @ranj063 can you update testbench to take a log level instead of a bool on the command line which can be passed above. Thanks!

ranj063: fixed now

posix/include/sof/lib/mailbox.h
```diff
@@ -0,0 +1,118 @@
...
+}
+
+static inline
+void mailbox_dspbox_read(void *dest, size_t dest_size,
+			 size_t offset, size_t bytes)
+{
+	int dsp_read_err __unused;
+
+	dcache_invalidate_region((__sparse_force void __sparse_cache *)(MAILBOX_DSPBOX_BASE +
+									offset), bytes);
```
lyakh, line 64: do we really need all these dcache calls? None of the hosts are cache-incoherent, are they?

PR #8134, opened 2023-08-31
audio: host: continue optimization for  host path

btian1:
before copier move to module interface, with same condition, its one time cycle is 420, now, it is 455.

After this PR, copier host playback path is lower 3% compared with before module interface.

### DIFF
--- a/src/audio/host-zephyr.c
+++ b/src/audio/host-zephyr.c
@@ -247,7 +247,7 @@ void host_common_update(struct host_data *hd, struct comp_dev *dev, uint32_t byt
 	bool update_mailbox = false;
 	bool send_ipc = false;
 
-	if (hd->ipc_host.direction == SOF_IPC_STREAM_PLAYBACK) {
+	if (dev->direction == SOF_IPC_STREAM_PLAYBACK) {
 		source = buffer_acquire(hd->dma_buffer);
 		sink = buffer_acquire(hd->local_buffer);
 		ret = dma_buffer_copy_from(source, sink, hd->process, bytes);
@@ -257,20 +257,14 @@ void host_common_update(struct host_data *hd, struct comp_dev *dev, uint32_t byt
 		ret = dma_buffer_copy_to(source, sink, hd->process, bytes);
 	}
 
-	/* assert dma_buffer_copy succeed */
-	if (ret < 0)
-		comp_err(dev, "host_common_update() dma buffer copy failed, dir %d bytes %d avail %d free %d",
-			 hd->ipc_host.direction, bytes,
-			 audio_stream_get_avail_samples(&source->stream) *
-				audio_stream_frame_bytes(&source->stream),
-			 audio_stream_get_free_samples(&sink->stream) *
-				audio_stream_frame_bytes(&sink->stream));
-
 	buffer_release(sink);
 	buffer_release(source);
 
-	if (ret < 0)
+	if (ret < 0) {
+		comp_err(dev, "host_common_update() copy failed, dir %d bytes %d  return: %d",
+			 dev->direction, bytes, ret);
 		return;
+	}
 
 	hd->total_data_processed += bytes;
 
--- a/src/audio/host-zephyr.c
+++ b/src/audio/host-zephyr.c
@@ -374,7 +374,6 @@ static uint32_t host_get_copy_bytes_normal(struct host_data *hd, struct comp_dev
 {
 	struct comp_buffer *buffer = hd->local_buffer;
 	struct comp_buffer *buffer_c;
-	struct comp_buffer *dma_buf_c;
 	struct dma_status dma_stat;
 	uint32_t avail_samples;
 	uint32_t free_samples;
@@ -391,9 +390,7 @@ static uint32_t host_get_copy_bytes_normal(struct host_data *hd, struct comp_dev
 		return 0;
 	}
 
-	dma_buf_c = buffer_acquire(hd->dma_buffer);
-	dma_sample_bytes = get_sample_bytes(audio_stream_get_frm_fmt(&dma_buf_c->stream));
-	buffer_release(dma_buf_c);
+	dma_sample_bytes = hd->config.src_width;
 
 	buffer_c = buffer_acquire(buffer);
 
--- a/src/audio/host-zephyr.c
+++ b/src/audio/host-zephyr.c
@@ -455,7 +455,6 @@ static inline bool stream_sync(struct host_data *hd, struct comp_dev *dev)
  */
 static int host_copy_normal(struct host_data *hd, struct comp_dev *dev, copy_callback_t cb)
 {
-	struct comp_buffer *buffer_c;
 	uint32_t copy_bytes;
 	const unsigned int threshold =
 #if CONFIG_HOST_DMA_RELOAD_DELAY_ENABLE
@@ -473,7 +472,6 @@ static int host_copy_normal(struct host_data *hd, struct comp_dev *dev, copy_cal
 	cb(dev, copy_bytes);
 
 	hd->partial_size += copy_bytes;
-	buffer_c = buffer_acquire(hd->dma_buffer);
 
 	/*
 	 * On large buffers we don't need to reload DMA on every period. When
@@ -482,8 +480,8 @@ static int host_copy_normal(struct host_data *hd, struct comp_dev *dev, copy_cal
 	 * also adding a 2ms safety margin.
 	 */
 	if (!IS_ENABLED(CONFIG_HOST_DMA_RELOAD_DELAY_ENABLE) ||
-	    audio_stream_get_size(&buffer_c->stream) < hd->period_bytes << 3 ||
-	    audio_stream_get_size(&buffer_c->stream) - hd->partial_size <=
+	    hd->dma_buffer_size < hd->period_bytes << 3 ||
+	    hd->dma_buffer_size - hd->partial_size <=
 	    (2 + threshold) * hd->period_bytes) {
 		if (stream_sync(hd, dev)) {
 			ret = dma_reload(hd->chan->dma->z_dev, hd->chan->index, 0, 0,
@@ -494,9 +492,6 @@ static int host_copy_normal(struct host_data *hd, struct comp_dev *dev, copy_cal
 			hd->partial_size = 0;
 		}
 	}
-
-	buffer_release(buffer_c);
-
 	return ret;
 }
 
@@ -872,6 +867,7 @@ int host_common_params(struct host_data *hd, struct comp_dev *dev,
 
 		config->src_width = audio_stream_sample_bytes(&dma_buf_c->stream);
 		config->dest_width = config->src_width;
+		hd->dma_buffer_size = audio_stream_get_size(&dma_buf_c->stream);
 		buffer_release(dma_buf_c);
 	}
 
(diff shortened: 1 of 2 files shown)


lgirdwood (commented):
@btian1 can you include what simplification changes were made in the commit message as it difficult to tell.

btian1:
> 

yes, please check again, it is still in draft, I need submit more patches to make host run as fast as it can.

src/include/sof/audio/host_copier.h
```diff
@@ -117,7 +118,67 @@ static inline int host_common_copy(struct host_data *hd, struct comp_dev *dev, c
 {
 	return hd->copy(hd, dev, cb);
 }
-void host_common_update(struct host_data *hd, struct comp_dev *dev, uint32_t bytes);
+
+static inline void host_common_update(struct host_data *hd, struct comp_dev *dev, uint32_t bytes)
```
lyakh, line 123: this one seems too big for a header to me. And it is called from multiple locations: e.g. from host-zephyr.c and from copier_host.c, so it will increase code size. We cannot inline everything, done in `.copy()` - we will have function calls and we should have them. We could try and inline "everything" effectively creating a single huge function, but that doesn't seem like a good idea to me...

src/audio/host-zephyr.c
```diff
@@ -243,35 +243,19 @@ void host_common_update(struct host_data *hd, struct comp_dev *dev, uint32_t byt
 {
 	struct comp_buffer __sparse_cache *source;
 	struct comp_buffer __sparse_cache *sink;
-	int ret;
```
lgirdwood, line 246: Looks like from commit message this should be 2 patches ? i.e. stage 1 and 2 above seem to be independent ?

src/audio/host-zephyr.c
```diff
@@ -378,9 +377,7 @@ static uint32_t host_get_copy_bytes_normal(struct host_data *hd, struct comp_dev
 		return 0;
 	}
 
-	dma_buf_c = buffer_acquire(hd->dma_buffer);
-	dma_sample_bytes = get_sample_bytes(audio_stream_get_frm_fmt(&dma_buf_c->stream));
-	buffer_release(dma_buf_c);
+	dma_sample_bytes = hd->config.src_width;
```
lgirdwood, line 380: ok, good - this is pre-calculated so can be reused on every iteration instead of looked up.

src/audio/host-zephyr.c
```diff
@@ -442,7 +442,6 @@ static inline bool stream_sync(struct host_data *hd, struct comp_dev *dev)
  */
 static int host_copy_normal(struct host_data *hd, struct comp_dev *dev, copy_callback_t cb)
 {
-	struct comp_buffer __sparse_cache *buffer_c;
```
lgirdwood, line 467: Ok, this is good as host_copy_normal() should always run on the same core as the buffer copier (and the copier would be the point where we would copy to another core). @marcinszkudlinski @lyakh fyi.

lgirdwood (commented):
Good savings here as we can assume that we only cross core with copier and up/down mix. Lets get inputs from @marcinszkudlinski and @lyakh as next step.

src/audio/host-zephyr.c
```diff
@@ -257,20 +257,18 @@ void host_common_update(struct host_data *hd, struct comp_dev *dev, uint32_t byt
...
+	buffer_release(sink);
+	buffer_release(source);
+
+	if (ret < 0) {
+		comp_err(dev, "host_common_update() copy failed, dir %d bytes %d avail %d free %d",
+			 dev->direction, bytes,
 			 audio_stream_get_avail_samples(&source->stream) *
 				audio_stream_frame_bytes(&source->stream),
 			 audio_stream_get_free_samples(&sink->stream) *
 				audio_stream_frame_bytes(&sink->stream));
```
lyakh, line 267: no, using `source` and `sink` after releasing them is wrong. We can postpone this until after the coherent API is removed from `struct comp_buffer`

src/audio/host-zephyr.c
```diff
@@ -302,17 +303,11 @@ void host_common_update(struct host_data *hd, struct comp_dev *dev, uint32_t byt
...
-			send_ipc = true;
-		}
-	}
-
-	if (update_mailbox) {
-		pipeline_get_timestamp(dev->pipeline, dev, &hd->posn);
-		mailbox_stream_write(dev->pipeline->posn_offset,
-				     &hd->posn, sizeof(hd->posn));
-		if (send_ipc)
+			pipeline_get_timestamp(dev->pipeline, dev, &hd->posn);
```
lyakh, line 306: can it be that we arrive here with `hd->cont_update_posn == true`? Then we would call `pipeline_get_timestamp()` and `mailbox_stream_write()` twice?

lyakh (changes requested):
Commits 3 and 4 look good, but the first two commits don't look convincing to me.

src/audio/host-zephyr.c
lyakh: in principle I think it's good to have this change, but in single core configurations impact should be minimal. In such configurations `buffer_acquire()` is a conversion from an uncached to a cached alias, `buffer_release()` is a NOP, and `get_sample_bytes()` is a primitive switch. So I really wouldn't expect this change to bring 5% of performance. However - if asserts are enabled - then yes, buffer acquisition will perform several more checks. Are you measuring performance with asserts enabled?

src/audio/host-zephyr.c
btian1: that's really tricky case, in theory, it is possible....
they are both assigned by params, if driver pass both as true, then this is the case.

src/audio/host-zephyr.c
btian1: it is debug build with -d option, so I would assume assert enabled.
BTW, I checked release build performance, it is much better than debug
                  host          dai
debug         396          455
release        340          405

we are seeing a nearly 15% difference between release and debug, it maybe caused by the assert.

kv2019i: That is still surprisingly high savings. Looking at assembly, the buffer_acquire() variant does not do this much more work. There is more code that is jumped over in single core case, but this should affect runtime cost this much. Perhaps this is leading to additional instruction cache miss. It would be good to understand this better, so we could pick good/bad usages better in code review.

kv2019i (commented):
Great findings @btian1  ! One generic comment about the commit messages, these should have a bit more context to explain the context on numbers. Although, it's great to have the numbers, this is much better than "optimize the copy function".

I am puzzled about some of the optimizations. They seem to have out-sized impact to measured performance when compared to compiler assembly output (see comments inline). OTOH, this is not really a reason to not merge this, just trying to learn from these to address good patterns in code reviews in the future.

Some of this will conflict with https://github.com/thesofproject/sof/pull/8106 which more generally removes buffer_acquire/release calls.

src/audio/host-zephyr.c
btian1: for debug, I did not check the assembly code for buffer_acquire().
There is so much macro inside.
after patch: #8106 merge, I guess there is a performance improve for each component and whole pipeline.
hard to give a number, maybe 10%

btian1:
move back to DRAFT, let's wait for #8106 merge first.

src/audio/host-zephyr.c
lyakh: > release 340 405

this means, that with a release build (no debugging overlay, asserts disabled) before your patch(es) DSP load way 405 and after your patch(es) 340? Is it with all of them? I'm particularly wondering about commits 1 and 2, I really don't think we should go down to that level of optimisation. I think we can talk about 3 code "classes"
1. configuration. All the code outside the LL scheduler threads, i.e. for components everything outside `.copy()` / `.process()` calls. That isn't very performance critical, emphasis on correctness and clarity
2. inside .copy()` / `.process()` but outside "byte shuffling" loops - that should be reasonably optimised but I wouldn't go crazy there. Avoid wasteful code, but keep it clean and readable
3. the hottest spots - actual byte processing loops. These should be optimised to the maximum. Here we can sacrifice readability to squeeze out an additional instruction saving

So, I think in some of your commits you might be overdoing with "class 2" optimisations. Maybe we could put a threshold - if a specific "class 2" optimisation brings less than 5% "overall" (i.e. measured over the entire `.copy()` call) performance saving and reduces clarity, readability, introduces code reuse etc. - maybe we shouldn't take it.

btian1: host      dai
debug build:  396      455
release build: 340      405

above table does not related this patch, it is just with same code base build with release and debug, different cycle cost for release and debug, it is obvious that release is faster 15-20% percent compared with debug.

I did not over-eng code, all the change are in critical timing path(1ms calling),  this function is mainly used perform host copy.

src/audio/host-zephyr.c
```diff
@@ -244,8 +244,6 @@ void host_common_update(struct host_data *hd, struct comp_dev *dev, uint32_t byt
 	struct comp_buffer __sparse_cache *source;
```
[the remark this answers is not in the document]
kv2019i: @btian1 git commit is best as that is stored in history, and anyone looking at the code later on, can find the reason why the change was made from git history. This is good to spread good coding practises, avoid people from removing useful code (if they don't understand why it is there) and give context in which conditions the savings are realized. This can be very short though, just key bits (what are the measure units, what platform and what was measured).

src/audio/host-zephyr.c
```diff
@@ -298,17 +299,13 @@ void host_common_update(struct host_data *hd, struct comp_dev *dev, uint32_t byt
...
-	if (update_mailbox) {
-		pipeline_get_timestamp(dev->pipeline, dev, &hd->posn);
-		mailbox_stream_write(dev->pipeline->posn_offset,
-				     &hd->posn, sizeof(hd->posn));
-		if (send_ipc)
+			if (!hd->cont_update_posn) {
+				pipeline_get_timestamp(dev->pipeline, dev, &hd->posn);
+				mailbox_stream_write(dev->pipeline->posn_offset,
+						     &hd->posn, sizeof(hd->posn));
+			}
```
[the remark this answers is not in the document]
lyakh: I really think this goes into the C optimisation territory. Both versions should be equivalent to the compiler and we have no idea how it chooses to compile each of them. Even if this version happens to produce faster code with the current version of the compiler and our current compilation flags, I don't think there are any guarantees that every other version will benefit from this rearrangement too. Even if this does indeed generate faster code now, I think it's more by luck than by science. As such I wouldn't rely on this kind of reorganisations.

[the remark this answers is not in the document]
btian1: it is not luck, it simply save two flags, and also saved some jump operation like beq based on two bool flags, after those are removed, it is a little faster, althrough logic is not changing, I agree before change code is more clear from logic.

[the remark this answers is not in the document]
lyakh: @btian1 you're correctly describing what the generated code would be like **if** the compiler didn't use any optimisations. With `-O0` it should allocate that boolean variable on stack, assign it, check it, use jumps etc. (but even then I'm not sure your code has fewer jumps, I'd have to check better) **BUT** - we do use compiler optimisations. And with those you cannot any more be sure, that the compiler doesn't optimise that boolean variable away. In fact I very much suspect that it does. You can check by looking at the disasm output, but even then - you'll only see what your specific compiler generated today. You won't know what another compiler (e.g. the one used to build release binaries) will generate - at the time of the next release, or whether we'll change compiler flags by then, or whether the release after that a new version of the compiler will be used, that will optimise the code differently. That's why I said "by luck"

lyakh (approved):

kv2019i (approved):
Thanks @btian1 !

kv2019i:
One system-PM fail in https://sof-ci.01.org/sofpr/PR8134/build12859/devicetest/index.htm (had to look up manually, result reporting has still issues), rest is ok. Proceeding with merge.

PR #8135, opened 2023-08-31
CODEOWNERS: add kv2019i to the list of default owners

kv2019i:
Add myself to the list of default owners that get added to all pull requests if no other entries match.


lgirdwood (approved):
Thank you @kv2019i

btian1:
image
Do we need fix above errors?

lgirdwood:
> Do we need fix above errors?

Yes - lets do in a followup PR.

PR #8140, opened 2023-09-01
Audio: Multiband-DRC: Add switch control for IPC4

singalsu:


ujfalusi (approved):
@singalsu, looks good to my eyes, thanks,

src/audio/multiband_drc/multiband_drc.c
```diff
@@ -321,18 +311,42 @@ static int multiband_drc_cmd_set_value(struct processing_module *mod,
...
 				    enum module_cfg_fragment_position pos,
 				    uint32_t data_offset_size, const uint8_t *fragment,
 				    size_t fragment_size, uint8_t *response,
 				    size_t response_size)
 {
 	struct multiband_drc_comp_data *cd = module_get_private_data(mod);
+	struct comp_dev *dev = mod->dev;
 
-	comp_dbg(mod->dev, "multiband_drc_set_config()");
+	comp_info(dev, "multiband_drc_set_config()");
```
singalsu, line 323: Hmm, no need to change this.

src/audio/multiband_drc/multiband_drc.c
```diff
@@ -250,17 +250,7 @@ static int multiband_drc_init(struct processing_module *mod)
 	md->private = cd;
 	cd->multiband_drc_func = NULL;
 	cd->crossover_split = NULL;
-#if CONFIG_IPC_MAJOR_4
```
singalsu, line 253: I think I need to drop this change for IPC4. The default should be enabled true, since an old kernel would never enable this switch control.

With new kernel the switch control is set to cached control state right after init, so the processing would be off if requested so. Also in IPC4 this control is not read back to kernel so this won't impact control state in kernel.


--> So I'd just change the comment and explain new and old kernel different and keep enabled = true.

src/include/ipc4/header.h
```diff
@@ -171,4 +171,31 @@ struct ipc4_message_reply {
...
+ * @id: unique id of the control
+ * @num_elems: Number of elememnts in the chanv array
+ * @reserved: reserved for future use, must be set to 0
+ * @chanv: channel ID and value array
+ */
+struct sof_ipc4_control_msg_payload {
+	uint16_t id;
+	uint16_t num_elems;
+	uint32_t reserved[4];
+	struct sof_ipc4_ctrl_value_chan chanv[0];
```
kv2019i, line 198: Let's use "chanv[]" for new code.

kv2019i (commented):
Looks good with the change to last patch.

src/include/ipc4/header.h
singalsu: Yep, it builds and works!

src/audio/multiband_drc/multiband_drc.c
```diff
@@ -251,11 +251,10 @@ static int multiband_drc_init(struct processing_module *mod)
 	cd->multiband_drc_func = NULL;
 	cd->crossover_split = NULL;
 #if CONFIG_IPC_MAJOR_4
-	/* Note: Currently there is no ALSA switch control in IPC4 to control
-	 * processing on/off. This workaround can be removed after is available.
-	 * Binary control for configuration blob can be used instead. There is
-	 * no processing bypass control in the blob but all processing can be
-	 * switched to neutral with supported min. 2-band mode.
+	/* Initialize to enabled is a workaround for IPC4 kernel version 6.5 and
```
ujfalusi, line 254: I would say that it is going to be true for <= 6.6

src/include/ipc4/header.h
```diff
@@ -171,4 +171,31 @@ struct ipc4_message_reply {
...
+
+/**
+ * struct sof_ipc4_ctrl_value_chan: generic channel mapped value data
+ * @channel: Channel ID
+ * @value: control value
+ */
+struct sof_ipc4_ctrl_value_chan {
+	uint32_t channel;
+	uint32_t value;
+};
```
lgirdwood, line 185: needs to be packed, aligned

lgirdwood (commented):
Thanks @singalsu , one comment from me.

lgirdwood (approved):

cujomalainey (approved):

src/audio/multiband_drc/multiband_drc.c
```diff
@@ -321,18 +321,42 @@ static int multiband_drc_cmd_set_value(struct processing_module *mod,
...
+		ctl = (struct sof_ipc4_control_msg_payload *)fragment;
+		comp_dbg(dev, "SOF_IPC4_SWITCH_CONTROL_PARAM_ID id = %d, num_elems = %d",
+			 ctl->id, ctl->num_elems);
+
+		if (ctl->id == 0 && ctl->num_elems == 1) {
+			cd->process_enabled = ctl->chanv[0].value;
+			comp_info(dev, "process_enabled = %d", cd->process_enabled);
+		} else {
+			comp_err(dev, "Illegal control id = %d, num_elems = %d",
+				 ctl->id, ctl->num_elems);
```
lyakh, line 348: but still returning 0 - ignoring and continuing?

src/include/ipc4/header.h
```diff
@@ -171,4 +171,31 @@ struct ipc4_message_reply {
...
+
+/**
+ * struct sof_ipc4_control_msg_payload - IPC payload for kcontrol parameters
+ * @id: unique id of the control
+ * @num_elems: Number of elememnts in the chanv array
+ * @reserved: reserved for future use, must be set to 0
+ * @chanv: channel ID and value array
+ */
+struct sof_ipc4_control_msg_payload {
+	uint16_t id;
```
btian1, line 195: here use uint16_t, how about line 183 and 184?

src/audio/multiband_drc/multiband_drc.c
```diff
@@ -321,18 +321,42 @@ static int multiband_drc_cmd_set_value(struct processing_module *mod,
...
 	struct multiband_drc_comp_data *cd = module_get_private_data(mod);
+	struct comp_dev *dev = mod->dev;
 
-	comp_dbg(mod->dev, "multiband_drc_set_config()");
+	comp_dbg(dev, "multiband_drc_set_config()");
 
-#if CONFIG_IPC_MAJOR_3
-	/* Other control types are possible only with IPC3 */
+#if CONFIG_IPC_MAJOR_4
+	struct sof_ipc4_control_msg_payload *ctl;
```
btian1, line 335: line 336 only used under line 339, can move to line 339

src/audio/multiband_drc/multiband_drc.c
btian1: should not, need return a negative value.

src/include/ipc4/header.h
ujfalusi: @btian1, the `id` is the identification (usually index) of a control for the module instance, u16 should be enough for a while until it is exhausted.
line 184 is the value stored for the channel, atm we only use this for enum and switch, but in theory it can be used for volume if needed in the future. There u16 would not be right.

src/audio/multiband_drc/multiband_drc.c
singalsu: I'll test what happens if I return error, I didn't want to stop streaming so I left it out, but maybe it doesn't. It's good if only amixer errors.

src/include/ipc4/header.h
singalsu: Resolving this, not an issue.

src/audio/multiband_drc/multiband_drc.c
singalsu: The control handler now errors as requested. The playback doesn't start at all since the control is sent to FW after init. I guess it's good to error incorrect topologies instead of just logging the errors, so proposing this now.

src/audio/multiband_drc/multiband_drc.c
singalsu: I combined the pointer declare and set. OK now?

lgirdwood (approved):

cujomalainey:
is cavs2.5 CI down?

singalsu:
SOFCI TEST

singalsu:
@cujomalainey Some good news, we have now successfully tested also enum control. Switch and enum controls update notification from firmware to user space works too. I'll update the example later to TDFB component.

btian1 (approved):

PR #8142, opened 2023-09-01
Tools: Tune: Multiband-DRC: Add comment about generator script

singalsu:
This patch helps to find the right tool to create the setup blobs.

### DIFF
--- a/tools/tune/multiband_drc/example_multiband_drc.m
+++ b/tools/tune/multiband_drc/example_multiband_drc.m
@@ -139,7 +139,7 @@ blob8_ipc4 = multiband_drc_build_blob(num_bands, enable_emp_deemp, emp_coefs, ..
 				      endian, 4);
 
 tplg_write(tplg1_fn, blob8, "MULTIBAND_DRC");
-tplg2_write(tplg2_fn, blob8_ipc4, "multiband_drc_config", 'Exported Control Bytes');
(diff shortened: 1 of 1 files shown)


ShriramShastry (approved):

lgirdwood (approved):

tools/tune/multiband_drc/example_multiband_drc.m
```diff
@@ -139,7 +139,7 @@ function export_multiband_drc(prm)
 				      endian, 4);
 
 tplg_write(tplg1_fn, blob8, "MULTIBAND_DRC");
-tplg2_write(tplg2_fn, blob8_ipc4, "multiband_drc_config", 'Exported Control Bytes');
+tplg2_write(tplg2_fn, blob8_ipc4, "multiband_drc_config", "Exported with script example_multiband_drc.m");
```
lgirdwood, line 142: btw, next steps would be to include the command line and tool version used to generate the data so that any developer can reproduce.

lgirdwood (approved):

tools/tune/multiband_drc/example_multiband_drc.m
singalsu: Yep, we can use more comment estate than this with a bit more changes.

Issue #7774, opened 2023-06-09
[BUG] Various IPC timed out with multi-core run simultaneously


keqiaozhang:
**Describe the bug**
We enabled multi-core tplg in CI and observed that it doesn't support 3 cores run at the same time. It failed with IPC timed out 0x40030004|MOD_INIT_INSTANCE.

**dmesg**
```
[ 2398.696058] kernel: snd_sof:sof_ipc4_widget_setup: sof-audio-pci-intel-mtl 0000:00:1f.3: Create widget dai-copier.SSP.NoCodec-2.playback instance 3 - pipe 6 - core 2
[ 2398.696064] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx      : 0x40030004|0x203002f: MOD_INIT_INSTANCE [data size: 188]
[ 2399.201595] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc timed out for 0x40030004|0x203002f
[ 2399.201612] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: Attempting to prevent DSP from entering D3 state to preserve context
[ 2399.201618] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ IPC dump start ]------------
[ 2399.201643] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: Host IPC initiator: 0xc0030004|0x203002f|0x0, target: 0x40030004|0x203002f|0x0, ctl: 0x3
[ 2399.201650] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ IPC dump end ]------------
[ 2399.201655] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ DSP dump start ]------------
[ 2399.201659] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: IPC timeout
[ 2399.201663] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: fw_state: SOF_FW_BOOT_COMPLETE (7)
[ 2399.201677] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ROM status: 0x5, ROM error: 0x0
[ 2399.201682] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ROM debug status: 0x50000005, ROM debug error: 0x0
[...]

PR #8143, opened 2023-09-04
Ipc4 multicore issue fixes

RanderWang:
BUG: Various IPC timed out with multi-core run simultaneously https://github.com/thesofproject/sof/issues/7774.

This issue is very random and can be reproduced after many cycles test. The direct reason is that : there is a self loop
in the ipc message list like:   ipc msg -> M1 -> M2 ->..... > Mn ->Mn , this result to a infinite loop with ipc spin lock hold for the logging function. This is one major reason that the issue can't be reproduced with logging disabled.
````c
-static bool is_notification_queued(void)
 {
        struct ipc *ipc = ipc_get();
-       struct list_item *slist;
-       struct ipc_msg *msg;
        k_spinlock_key_t key;
        bool queued = false;
 
        key = k_spin_lock(&ipc->lock);
-       if (list_is_empty(&ipc->msg_list))
-               goto out;
-
-       list_for_item(slist, &ipc->msg_list) {
-               msg = container_of(slist, struct ipc_msg, list);
-
-               if (msg == &msg_notify) {
-                       queued = true;
-                       break;
-               }
-       }
-
-out:
        k_spin_unlock(&ipc->lock, key);
 
        return queued;

````
Why the single ipc message is double added ? The reason is that In multicore case, IPC message is dispatched from primary core to secondary core which send reply message to host. Primary core will do nothing if IPC_TASK_SECONDARY_CORE is set. But in rare case, the secondary code finish the reply message and clear this flag before the ipc thread in primary core check this flag because the primary core was busy with logging thread and check it after awhile, then primary core also send reply message again. This results to the reply message being inserted two times in ipc message list and result to infinite loop when visiting the list for the logging after awhile.

We can find some hint in kernel log before ipc timeout.
````
[58167.022730] snd_sof:sof_ipc4_route_setup: sof-audio-pci-intel-mtl 0000:00:1f.3: bind gain.4.1:0 -> dai-copier.SSP.NoCodec-1.playback:0
[58167.022743] snd_sof:: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx      : 0x45050006|0x50004: MOD_BIND
[58167.023200] snd_sof:: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx reply: 0x65000000|0x50004: MOD_BIND
[58167.023374] snd_sof:: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx reply: 0x65000000|0x50004: MOD_BIND
[58167.023417] snd_sof:snd_sof_ipc_reply: 0000:00:1f.3: no reply expected, received 0x65000000, will be ignored
````
For the case in bug:https://github.com/thesofproject/sof/issues/7774
````
[ 2398.695476] : snd_sof:: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx done : 0x40030006|0x2030010: MOD_INIT_INSTANCE [data size: 64]
[ 2398.695496] : sof-audio-pci-intel-mtl ipc tx 0x44030006|0x30000018: MOD_LARGE_CONFIG_SET [data size: 24]
[ 2398.695718] : sof-audio-pci-intel-mtl ipc tx reply: 0x64000000|0x30000018: MOD_LARGE_CONFIG_SET
[ 2398.695995] : sof-audio-pci-intel-mtl ipc tx reply: 0x64000000|0x30000018: MOD_LARGE_CONFIG_SET
[ 2398.696006] : sof-audio-pci-intel-mtl  ipc tx done : 0x44030006|0x30000018: MOD_LARGE_CONFIG_SET [data size: 24]
[ 2398.696018] : snd_sof:snd_sof_ipc_reply: 0000:00:1f.3: no reply expected, received 0x64000000, will be ignored
[ 2398.696043] : snd_sof:sof_widget_setup_unlocked: 0000:00:1f.3: widget gain.6.1 setup complete
[ 2398.696058] : snd_sof:sof_ipc4_widget_setup:  Create widget dai-copier.SSP.NoCodec-2.playback instance 3 - pipe 6 - core 2
[ 2398.696064] : snd_sof:sof_ipc4_log_header: : ipc tx      : 0x40030004|0x203002f: MOD_INIT_INSTANCE [data size: 188]
[ 2399.201595] : sof-audio-pci-intel-mtl 0000:00:1f.3: ipc timed out for 0x40030004|0x203002f
[ 2399.201612] : sof-audio-pci-intel-mtl 0000:00:1f.3: Attempting to prevent DSP from entering D3 state to preserve context
[ 2399.201618] : sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ IPC dump start ]------------
[ 2399.201643] : sof-audio-pci-intel-mtl 0000:00:1f.3: Host IPC initiator: 0xc0030004|0x203002f|0x0, target: 0x40030004|0x203002f|0x0, ctl: 0x3
[ 2399.201650] : sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ IPC dump end ]------------
[ 2399.201655] : sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ DSP dump start ]------------
[ 2399.201659] : sof-audio-pci-intel-mtl 0000:00:1f.3: IPC timeout
````
This PR checks whether the msg has been done by secondary core and do nothing for primary core. Also simply the check for logging.

### DIFF
--- a/src/ipc/ipc4/handler.c
+++ b/src/ipc/ipc4/handler.c
@@ -65,7 +65,7 @@ struct ipc4_msg_data {
 static struct ipc4_msg_data msg_data;
 
 /* fw sends a fw ipc message to send the status of the last host ipc message */
-static struct ipc_msg msg_reply;
+static struct ipc_msg msg_reply = {0, 0, 0, 0, LIST_INIT(msg_reply.list)};
 
 static struct ipc_msg msg_notify;
 
@@ -1303,7 +1303,6 @@ void ipc_cmd(struct ipc_cmd_hdr *_hdr)
 	msg_reply.tx_size = 0;
 	msg_reply.header = in->primary.dat;
 	msg_reply.extension = in->extension.dat;
-	list_init(&msg_reply.list);
 
 	target = in->primary.r.msg_tgt;
 
@@ -1330,8 +1329,62 @@ void ipc_cmd(struct ipc_cmd_hdr *_hdr)
 		char *data = ipc->comp_data;
 		struct ipc4_message_reply reply;
 
+		/* Process flow and time stamp for IPC4 msg processed on secondary core :
+		 * core 0 (primary core)				core x (secondary core)
+		 * # IPC msg thread		#IPC delayed worker     #core x idc thread
+		 * ipc_task_ops.run()
+		 * ipc_do_cmd()
+		 * msg_reply.header = in->primary.dat
+		 * ipc4_process_on_core(x)
+		 * mask |= SECONDARY_CORE
+		 * idc_send_message()
+		 * Case 1:
+		 * // Ipc msg processed by secondary core		idc_ipc()
+		 * if ((mask & SECONDARY_CORE))				ipc_cmd()
+		 *	return;						ipc_msg_send()
+		 *							mask &= ~SECONDARY_CORE
+		 *
+		 *				ipc_platform_send_msg
+		 * ----------------------------------------------------------------------------
+		 * Case 2:
+		 *                                                      idc_ipc()
+		 *                                                      ipc_cmd()
+		 *							//Prepare reply msg
+		 *                                                      msg_reply.header =
+		 *                                                      reply.primary.dat;
+		 *                                                      ipc_msg_send()
+		 *                                                      mask &= ~SECONDARY_CORE
+		 *
+		 * if ((mask & IPC_TASK_SECONDARY_CORE))
+		 *	return;
+		 * // Ipc reply msg was prepared, so return
+		 * if (msg_reply.header != in->primary.dat)
+		 *	return;
+		 *				ipc_platform_send_msg
+		 * ----------------------------------------------------------------------------
+		 * Case 3:
+		 *                                                      idc_ipc()
+		 *                                                      ipc_cmd()
+		 *							//Prepare reply msg
+		 *                                                      msg_reply.header =
+		 *                                                      reply.primary.dat;
+		 *                                                      ipc_msg_send()
+		 *                                                      mask &= ~SECONDARY_CORE
+		 *
+		 *                              ipc_platform_send_msg
+		 *
+		 * if ((mask & IPC_TASK_SECONDARY_CORE))
+		 *      return;
+		 * // Ipc reply msg was prepared, so return
+		 * if (msg_reply.header != in->primary.dat)
+		 *	return;
+		 */
+
 		/* Reply prepared by secondary core */
 		if ((ipc->task_mask & IPC_TASK_SECONDARY_CORE) && cpu_is_primary(cpu_get_id()))
+			return;
+		/* Reply has been prepared by secondary core */
+		if (msg_reply.header != in->primary.dat)
 			return;
 
 		/* Do not send reply for SET_DX if we are going to enter D3
--- a/src/ipc/ipc4/handler.c
+++ b/src/ipc/ipc4/handler.c
@@ -67,7 +67,7 @@ static struct ipc4_msg_data msg_data;
 /* fw sends a fw ipc message to send the status of the last host ipc message */
 static struct ipc_msg msg_reply = {0, 0, 0, 0, LIST_INIT(msg_reply.list)};
 
-static struct ipc_msg msg_notify;
+static struct ipc_msg msg_notify = {0, 0, 0, 0, LIST_INIT(msg_notify.list)};
 
 /*
  * Global IPC Operations.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/ipc/ipc4/handler.c [hunk 1]
```diff
@@ -65,7 +65,7 @@ struct ipc4_msg_data {
 static struct ipc4_msg_data msg_data;
 
 /* fw sends a fw ipc message to send the status of the last host ipc message */
-static struct ipc_msg msg_reply;
+static struct ipc_msg msg_reply = {0, 0, 0, 0, {&msg_reply.list, &msg_reply.list}};
```
kv2019i, line 68: maybe:
static struct ipc_msg msg_reply = {0, 0, 0, 0, LIST_INIT(&msg_reply.list)};

src/ipc/ipc4/handler.c
```diff
@@ -67,7 +67,7 @@ static struct ipc4_msg_data msg_data;
 /* fw sends a fw ipc message to send the status of the last host ipc message */
 static struct ipc_msg msg_reply = {0, 0, 0, 0, {&msg_reply.list, &msg_reply.list}};
 
-static struct ipc_msg msg_notify;
+static struct ipc_msg msg_notify = {0, 0, 0, 0, {&msg_notify.list, &msg_notify.list}};
```
kv2019i, line 70: If LIST_INIT works, that would be cleaner here as well.

kv2019i (commented):
Very good catch @RanderWang , the SECONDARY_CORE checking indeed does not seem. One minor code style proposal inline, but otherwise looks good.

src/ipc/ipc4/handler.c
```diff
@@ -65,7 +65,7 @@ struct ipc4_msg_data {
 static struct ipc4_msg_data msg_data;
```
lyakh, line 65: This is what's happening:
```
IPC4:

core 0					core 1

IPC
ipc_task_ops.run()
ipc_do_cmd()
ipc_platform_do_cmd()
ipc_cmd()
mask |= INLINE
mask |= SECONDARY
IDC
					idc_ipc()
					ipc_cmd()
					ipc_msg_send()
					list_item_append()
					mask &= ~SECONDARY
if (!(mask & SECONDARY))
	ipc_msg_send()
	list_item_append()

ipc_task_ops.complete()
ipc_complete_task()
```
and it isn't happening on IPC3 because there we always check return value on the primary core and it's set to 1 when delegating to a secondary core. In principle I don't think you need to take the spinlock below, because in `idc_handler()` first the message is added to the list and only then the `IPC_TASK_SECONDARY_CORE` flag is cleared, so at least one of the indicators is always there, but it actually does "seem to add correctness" using a spinlock here too, so, I wouldn't object it.
The only request here - I think we need a comment to explain, that just checking the flag isn't enough, because the secondary core could've outrun us and already cleared the flag, so we need to check the list too.
But - can it not happen that the IPC worker also preempts us, sends the message out and removes it from the list, so we re-add it again? Maybe we need a more robust flag like that return code in IPC3.

RanderWang: @lyakh  In term of lock, I first didn't add it but added it later for some reviewer may advise to add it.  And the issue that IPC worker also preempts ipc thread, I first thought the IPC thread has higher priority than worker thread, so it is no problem. But later I still got the ipc time out issue in long time test. I added some log and found it was caused by your predict. Does zephyr schedule thread based not only on priority ? Consider starve thread ?  Thanks!

src/ipc/ipc4/handler.c
RanderWang: thanks, refined,  it should be static struct ipc_msg msg_reply = {0, 0, 0, 0, LIST_INIT(msg_reply.list)};

RanderWang:
@kv2019i  @lyakh  Update for your comments. Thanks!

RanderWang:
Validated this PR with 3 different MTL RVPs, all passed a 3 hours multicore stress test.

src/ipc/ipc4/handler.c
lyakh: Sorry, I shouldn't have used "preempt" - I meant when running on 2 cores if the worker could run on core 1 before core 0 checks the mask as in pseudo-code above.
And BTW, it just occurred to me, that our IPC sending work can run on any core and that is probably not what we want?

softwarecki (approved):
Good catch!

src/ipc/ipc4/handler.c
```diff
@@ -1330,9 +1329,63 @@ void ipc_cmd(struct ipc_cmd_hdr *_hdr)
...
+		 *                                                      ipc_cmd()
+		 *							//Prepare reply msg
+		 *                                                      msg_reply.header =
+		 *                                                      reply.primary.dat;
+		 *                                                      ipc_msg_send()
+		 *                                                      mask &= ~SECONDARY_CORE
+		 *
+		 *                              ipc_platform_send_msg
+		 *
+		 * if ((mask & IPC_TASK_SECONDARY_CORE))
```
kv2019i, line 1362: @RanderWang  Case 2 and 3 are identical, I can't spot any difference?

kv2019i (commented):
Code is good, but there's an issue with the long inline comment.

src/ipc/ipc4/handler.c
lyakh: @kv2019i I think the difference is when `ipc_platform_send_msg()` is run by the IPC work handler - before core 0 resumes IPC processing or after?

lgirdwood:
Lets get this tested in daily test tomorrow. @mwasko fyi.

PR #8144, opened 2023-09-04
audio: module adapter: split out module adapter ipc3 and ipc4 code

btian1:
Create ipc3 and ipc4 specific source file, these files will only be used to store specific code.
Then move code from module_adapter.c to module_adapter_ipc3.c and module_adapter_ipc4.c.

### DIFF
--- a/src/audio/module_adapter/CMakeLists.txt
+++ b/src/audio/module_adapter/CMakeLists.txt
@@ -1,7 +1,11 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 
-add_local_sources(sof module_adapter.c module/generic.c)
+if(CONFIG_IPC_MAJOR_3)
+	add_local_sources(sof module_adapter.c module_adapter_ipc3.c module/generic.c)
+elseif(CONFIG_IPC_MAJOR_4)
+	add_local_sources(sof module_adapter.c module_adapter_ipc4.c module/generic.c)
+endif()
 
 if(NOT CONFIG_COMP_MODULE_SHARED_LIBRARY_BUILD)
 	if(CONFIG_CADENCE_CODEC)
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -29,19 +29,6 @@
 LOG_MODULE_REGISTER(module_adapter, CONFIG_SOF_LOG_LEVEL);
 
 /*
- * helpers to determine processing type
- * Needed till all the modules use PROCESSING_MODE_SINK_SOURCE
- */
-#define IS_PROCESSING_MODE_AUDIO_STREAM(mod) \
-		(!!((struct module_data *)&(mod)->priv)->ops->process_audio_stream)
-
-#define IS_PROCESSING_MODE_RAW_DATA(mod) \
-		(!!((struct module_data *)&(mod)->priv)->ops->process_raw_data)
-
-#define IS_PROCESSING_MODE_SINK_SOURCE(mod) \
-		(!!((struct module_data *)&(mod)->priv)->ops->process)
-
-/*
  * \brief Create a module adapter component.
  * \param[in] drv - component driver pointer.
  * \param[in] config - component ipc descriptor pointer.
@@ -95,62 +82,12 @@ struct comp_dev *module_adapter_new(const struct comp_driver *drv,
 	comp_set_drvdata(dev, mod);
 	list_init(&mod->sink_buffer_list);
 
-#if CONFIG_IPC_MAJOR_3
-	const unsigned char *data;
-	uint32_t size;
-
-	switch (config->type) {
-	case SOF_COMP_VOLUME:
-	{
-		const struct ipc_config_volume *ipc_volume = spec;
-
-		size = sizeof(*ipc_volume);
-		data = spec;
-		break;
-	}
-	case SOF_COMP_SRC:
-	{
-		const struct ipc_config_src *ipc_src = spec;
-
-		size = sizeof(*ipc_src);
-		data = spec;
-		break;
-	}
-	default:
-	{
-		const struct ipc_config_process *ipc_module_adapter = spec;
-
-		size = ipc_module_adapter->size;
-		data = ipc_module_adapter->data;
-		break;
-	}
-	}
-
-	/* Copy initial config */
-	if (size) {
-		ret = module_load_config(dev, data, size);
-		if (ret) {
-			comp_err(dev, "module_adapter_new() error %d: config loading has failed.",
-				 ret);
-			goto err;
-		}
-		dst->init_data = dst->data;
-	} else {
+	ret = module_adapter_init_data(dev, dst, config, spec);
+	if (ret) {
+		comp_err(dev, "module_adapter_new() %d: module init data failed",
+			 ret);
 		goto err;
 	}
-#else
-	if (drv->type == SOF_COMP_MODULE_ADAPTER) {
-		const struct ipc_config_process *ipc_module_adapter = spec;
-
-		dst->init_data = ipc_module_adapter->data;
-		dst->size = ipc_module_adapter->size;
-		dst->avail = true;
-
-		memcpy(&dst->base_cfg, ipc_module_adapter->data, sizeof(dst->base_cfg));
-	} else {
-		dst->init_data = spec;
-	}
-#endif
 
 	/* Modules must modify them if they support more than 1 source/sink */
 	mod->max_sources = 1;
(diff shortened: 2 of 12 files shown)


src/audio/module_adapter/module_adapter_ipc3.c
```diff
@@ -0,0 +1,330 @@
...
+{
+	return 0;
+}
+
+uint64_t module_adapter_get_total_data_processed(struct comp_dev *dev,
+						 uint32_t stream_no, bool input)
+{
+	return 0;
+}
+
```
lgirdwood, line 330: These are IPC4 only APIs, I would move them to the common header and make then static inline if using IPC3. That way they all become NOPs for all IPC3 users (and CC will eliminate).

lgirdwood (approved):
@kv2019i any comments. @btian1 ipc4 only static inline APIs can be fixed incrementally.

src/audio/module_adapter/module_adapter_ipc3.c
btian1: yes, however, if move to generic header file:
1. need add #if ipc3
2. ipc4  implementation have more than 200 lines, not suitable to move to header. 
    still need to keep in ipc4 source file.
3. somehow, with current implementation, code looks symmetry.

Finally, those should be empty for ipc3 build as long as with -o3, each compiler should eliminate those empty function.

src/audio/module_adapter/module_adapter.c
```diff
@@ -869,144 +790,6 @@ static int module_adapter_audio_stream_copy_1to1(struct comp_dev *dev)
...
-	uint32_t num_input_buffers, num_output_buffers;
-	int ret, i = 0;
-
-	/* handle special case of HOST/DAI type components */
-	if (dev->ipc_config.type == SOF_COMP_HOST || dev->ipc_config.type == SOF_COMP_DAI)
-#if CONFIG_IPC_MAJOR_3
-		return module_process_legacy(mod, NULL, 0, NULL, 0);
-#else
-		return module_process_stream(mod, NULL, 0, NULL, 0);
-#endif
```
lyakh, line 887: just because of this 1-line difference the whole function is duplicated?

src/audio/module_adapter/CMakeLists.txt
```diff
@@ -2,9 +2,9 @@
 
 
 if(CONFIG_IPC_MAJOR_3)
-	add_local_sources(sof module_adapter.c module_adapter_ipc3.c module/generic.c)
+	add_local_sources(sof module_adapter.c module_adapter_ipc3.c module/generic_ipc3.c)
 elseif(CONFIG_IPC_MAJOR_4)
-	add_local_sources(sof module_adapter.c module_adapter_ipc4.c module/generic.c)
+	add_local_sources(sof module_adapter.c module_adapter_ipc4.c module/generic_ipc4.c)
```
lyakh, line 7: after this commit generic_ipc3.c and generic_ipc4.c are 90% identical. I'm not a fan of such duplication.

kv2019i (commented):
This does bring to light issues that ifdefs easily hide. I think this looks pretty good, but I'd take another look at module_adapter_audio_stream_type_copy(). Handling setup and parameters will be IPC specific, but the actually processing code should be IPC agnotic. OTOH, not sure what is behind the logic for the 1-line different.

src/audio/module_adapter/CMakeLists.txt
btian1: generic.c is a little special, it does not like other common files(extract small function to cover).
it difference are quite a lot(>10), and not easy extract small functions to cover.
Do you have any better ideas?

src/audio/module_adapter/module_adapter.c
btian1: yes, this is deserved, since process is time critical, code looks duplicated, however, it does not, due to ipc3 and ipc4 build separately.

btian1:
> This does bring to light issues that ifdefs easily hide. I think this looks pretty good, but I'd take another look at module_adapter_audio_stream_type_copy(). Handling setup and parameters will be IPC specific, but the actually processing code should be IPC agnotic. OTOH, not sure what is behind the logic for the 1-line different.

"should be IPC agnotic" this is because we are using separate function to handle ipc3 and ipc4, still same way with setup and parameters, developer will goto module_adapter_ipc3 for ipc3 related, and go to module_adapter_ipc4 for ipc4 related.

src/audio/module_adapter/module_adapter_ipc3.c
lgirdwood: > 

Thats fine having a single ```#if IPC3``` in the header, this also make sure we are not building any code that is not needed even in debug. Fwiw, this is what Linux does too.

src/audio/module_adapter/module_adapter.c
kv2019i: @btian1 The code is is duplicated in the codebase and if we find a bug in this long function, it now needs to be fixed in two places.

But right, it is a compromise to make, we don't want any added runtime cost in functions like copy/process.

I'm looking at @ranj063 's 5f6f22d25e3a74eca84ca01ba79949d3ebbce960 and your commit 4cd928007edf10ed5ac6fa4022a2eb0a348e651a . So this is a anyways a special case for DAI.

Can we call "module_process_for_dai()" in generic.c and then have a ifdef in header to call (inline) either legacy or stream (IPC3 or IPC4)? Then we can share this function as-is and have no runtime penalty.

lyakh: I still don't think we should optimise even the hot path to that level. At least I don't think we're currently at that stage. As I said again - I do think, that we have to optimise to the maximum internal processing loops. There - yes, every instruction, every cycle counts. But anything before or after that loop - I don't think we're saving function calls there. Those paths should avoid wasting cycles meaninglessly of course, but I don't think we're getting rid of all function calls at that level.

src/audio/module_adapter/CMakeLists.txt
lyakh: you might need to handle each of them individually. In fact I'm not sure I understand the approach: is it our rule now that we must not have a single `#if CONFIG_IPC_MAJOR_3` in our C files, and if we find one, that file should be split? We absolutely must get rid of all of them? Why are they so harmful?

src/audio/module_adapter/module_adapter.c
lyakh: > Can we call "module_process_for_dai()" in generic.c

@kv2019i maybe "edge" since it is for both DAI and host?

src/audio/module_adapter/CMakeLists.txt
btian1: there are two reasons to do this:

1. separate ipc3 and ipc4 code.
2. easier to optimize ipc4 specific code, no need have concern on impact on ipc3.

src/audio/module_adapter/module_adapter.c
lyakh: > module_process_legacy was called in multiple place, so change name is not suitable.

@btian1 I think what @kv2019i meant was
```
static inline int module_process_for_dai(...)
{
#if CONFIG_IPC_MAJOR_3
    return module_process_legacy(....);
#else
    return module_process_stream(...);
#endif
}
```
that way you remove the `#if` from the function body, incur no run-time overhead, and keep both functions to be called elsewhere as needed.

> also as this is the common file for ipc3 and ipc4, expectation is zero #if IPC3/IPC4.

I'm not aware of this expectation. We're certainly rather far from that. I wouldn't be against it - I'd be happy to have zero `#if`s in .c code altogether, like the Linux kernel does, but SOF is more tolerant to that. So, in general, getting rid of them is good, but not at any cost, I would say.

lgirdwood:
@ranj063 any comments ?

src/audio/module_adapter/CMakeLists.txt
lgirdwood: @btian1 we still need to have a common file where any code that is teh same for ipc3/4 resides.

lgirdwood (commented):
Needs a common file if generic_ipc3/4 are 90% identical.

btian1:
> Needs a common file if generic_ipc3/4 are 90% identical.

@lgirdwood @lyakh , for generic.c, if duplicated 90% code is not allowed, then I would suggest don't change it, because any changes will add cost to ipc4 module adapter(since only ipc3 have code, and common function will cover ipc4 as well), this is not good, so I would suggest keep generic as for now, it is quite simple and easy to find/understand ipc3 only code.

I removed patch 2,  and only keep 1.

kv2019i (approved):

PR #8122, opened 2023-08-29
Tools: Tune: Add tplg2 support to dcblock setup tool

singalsu:

PR #8145, opened 2023-09-04
Tools: Topology2: Add DC block component

singalsu:
Updated description - this is more a collection of patches to create in simple way topologies to test all processing components by e.g. loading and unloading the component & format specific topologies.

- first patch adds dcblock widget class and example configuration blob
- second patch adds build of minimalistic topology to test dccomp. The include files needed are generated by shell script bench_comp_generate.sh <component>. They may need small hand edit for bytes control presence, add of other control like switch, or name or blob file name.
- in same style test topologies are added for DRC, IIR, FIR in patches 3 - 5.

Example of topology to test s32 format dcblock:

sof-hda-benchmark-dcblock32-tgl


singalsu:
@ranj063 @kv2019i @plbossart I wonder if this is a good way to create simple topologies to allow test run for any processing component that can fit the one sourcec one sink template? Can it be done more efficiently with tplg2 syntax? I'd like to run these same topologies in DUT, SOF ALSA plugin, and testbench environments.

tools/topology/topology2/include/components/dcblock/default.conf
```diff
@@ -0,0 +1,12 @@
+# Exported Control Bytes 29-Aug-2023
```
lgirdwood, line 1: Can we comment on how these are generated i.e. so that users know what tool to run and any command line args used to generate this data (and it can be reproduced by them).

lgirdwood (commented):
@singalsu I'm good with a bench tplg2 dir that can be invoked later by CI using test bench and plugin to test quality, perf on host.

tools/topology/topology2/include/components/dcblock/default.conf
singalsu: Good that you noticed, this was generated with an earlier version of #8122. I need to update this new generated file that shows the text.

singalsu: It's fixed now

tools/topology/topology2/include/bench/bench_comp_generate.sh
```diff
@@ -0,0 +1,104 @@
...
+    generate_playback_controls
+    generate_capture_controls
+}
+
+generate_comp ()
+{
+    format=$1
+    fn=${comp}_${format}.conf
+    echo Creating file $fn
+    cat > $fn <<EOF_COMP
```
singalsu, line 28: Seems there's some issues found by shellcheck.

tools/topology/topology2/include/components/dcblock/default.conf
```diff
@@ -0,0 +1,12 @@
+# Exported with script example_dcblock.m 27-Sep-2023
```
lgirdwood, line 1: but what about the cmd line args used to generate this and other blobs ? i.e. how would someone else reproduce this ?

tools/topology/topology2/include/bench/bench_comp_generate.sh
```diff
@@ -0,0 +1,104 @@
...
+
+set -e
+
+main () {
+    if [ "$#" -ne 1 ]; then
+	echo "Usage: $0 <component>"
+	exit 1
+    fi
+    comp=$1
+
```
marc-hb, line 14:
```suggestion
FULL_CMD=( "$0" "$@" )
```

tools/topology/topology2/include/bench/bench_comp_generate.sh
```diff
@@ -0,0 +1,104 @@
...
+    generate_capture_controls
+}
+
+generate_comp ()
+{
+    format=$1
+    fn=${comp}_${format}.conf
+    echo Creating file "$fn"
+    cat > "$fn" <<EOF_COMP
+		# Created with script "$0 $comp"
```
marc-hb, line 29:
```suggestion
		# Created with script ${FULL_CMD[@]}
```

This addresses @lgirdwood 's concern.

It does not "re-quote" whitespace but good enough for now.

singalsu: Thanks, yes that's better, though it's not same as what Liam asked.

tools/topology/topology2/include/components/dcblock/default.conf
singalsu: Though I guess you meant how to launch the blobs generator from shell command line. I can add that as well.

singalsu: It would be then `cd tools/tune/dcblock; octave --no-window-system example_dcblock.m` .

tools/topology/topology2/include/bench/bench_comp_generate.sh
```diff
@@ -0,0 +1,104 @@
...
+# Copyright(c) 2023 Intel Corporation. All rights reserved.
+
+set -e
+
+main () {
+    if [ "$#" -ne 1 ]; then
+	echo "Usage: $0 <component>"
+	exit 1
+    fi
+    comp=$1
```
marc-hb, line 13: Nit
```suggestion
    COMP=$1
```

kv2019i:
@ranj063 Good to go?

PR #8146, opened 2023-09-04
zephyr: ll-schedule: stop using zephyr/timeout_q.h

kv2019i:
Replace the call to z_abort_timeout() to be able to drop dependency to timeout_q.h. This header is intended to be Zephyr kernel private and plan is to remove this.

Unblock https://github.com/zephyrproject-rtos/zephyr/pull/62051

### DIFF
--- a/src/schedule/zephyr_domain.c
+++ b/src/schedule/zephyr_domain.c
@@ -23,7 +23,6 @@
 
 #include <zephyr/kernel.h>
 #include <zephyr/sys_clock.h>
-#include <zephyr/timeout_q.h>
 
 LOG_MODULE_DECLARE(ll_schedule, CONFIG_SOF_LOG_LEVEL);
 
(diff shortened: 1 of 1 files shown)


nashif (approved):

src/schedule/zephyr_domain.c
```diff
@@ -112,7 +111,7 @@ static void zephyr_domain_timer_fn(struct k_timer *timer)
 	 * registered again next time.
 	 */
 	if (!zephyr_domain) {
-		z_abort_timeout(&timer->timeout);
+		k_timer_stop(&timer);
```
btian1, line 114: k_timer_stop(&timer); or k_timer_stop(timer); ?

kv2019i: Thanks @btian1  , fixed!

lyakh:
> @lyakh can you check? I was checking k_timer_stop() implementation and it would seem to do the trick, but wondering why we did not use this to begin with?

You have seen this, right? https://github.com/thesofproject/sof/commit/89ca17b0f0dda6e3618be1a0f20e11fcb93a0dac  But I think `k_timer_stop()` should work too, not sure why I didn't use it back then. The only difference (taking into account that we don't use a timer stop function) between them is an attempt to run pending tasks, but there should be no tasks pending on that timer since we never use 'k_timer_status_sync()' So, looks good...

lyakh (approved):

Issue #8115, opened 2023-08-29
[BUG] IPC timeout check mismatch between kernel and fw


RanderWang:
Linux driver set ipc timeout duration to 500ms, but FW wait 5000ms in cpu_disable_core for SET_DX ipc message. This makes driver stop but fw is still waiting.  And why we need 5000ms to disable core 1 ~ n ? Can we adopt 400 or 300 ms ?

````c
void cpu_disable_core(int id) {
       .....
	uint64_t timeout = k_cycle_get_64() +
		k_ms_to_cyc_ceil64(CONFIG_SECONDARY_CORE_DISABLING_TIMEOUT); // 5000ms

	/* Waiting for secondary core to enter idle state */
	while (arch_cpu_active(id) && (k_cycle_get_64() < timeout))
		idelay(PLATFORM_DEFAULT_DELAY);

	if (arch_cpu_active(id)) {
		tr_err(&zephyr_tr, "core %d did not enter idle state", id);
		return;
	}
     .....
}
````

**Comment by kv2019i** [comment]
Labeling P2 for investigation, need to check with git blame on what is history of this long timeout.

**Comment by RanderWang** [comment]
normally only needs less than 1ms
````
[    9.832556] <err> zephyr: cpu_disable_core: ~~~~~~~~ core[2] entry
[    9.832568] <inf> os: pm_state_set: --------------- power off core[2] 
[    9.832585] <err> zephyr: cpu_disable_core: ^^^^^^^^^^ core[2] exit

[    9.835976] <err> zephyr: cpu_disable_core: ~~~~~~~~ core[1] entry
[    9.835986] <inf> os: pm_state_set: --------------- power off core[1] 
[    9.836003] <err> zephyr: cpu_disable_core: ^^^^^^^^^^ core[1] exit
````

**Comment by kv2019i** [comment]
@RanderWang git blame is your friend.. looking at history, the 5000 comes from this commit from @tmleman 
```
commit b6a4bb682cade0f4eb7a9ffe872f6fe46e099a2e
Author: Tomasz Leman <tomasz.m.leman@intel.com>
Date:   Fri Apr 29 16:37:07 2022 +0200

    platform: Kconfig: add timeout for disabling secondary core
    
    Count of loops used when waiting for secondary core to enter OFF state
    in transition to D3 state.
```

This is a generic setting applying to all platforms, so it's probably conservative because of that. Do you @tmleman any specific reason to have such a long time.

[...]

PR #8149, opened 2023-09-05
core: set timeout to 400ms for secondary core disabling

RanderWang:
Linux driver set ipc timeout duration to 500ms, but FW wait 5000ms in cpu_disable_core for SET_DX ipc message. This makes driver stop but fw is still waiting. Actually the disabling can be successful in far less than 5000ms and it was first designed for FPGA case. Now change it to 400ms to make debugging easier.

This patch changes cavs and ace15 platforms and doesn't change ace20 for its current state.

fix https://github.com/thesofproject/sof/issues/8115

### DIFF
--- a/src/platform/Kconfig
+++ b/src/platform/Kconfig
@@ -468,7 +468,7 @@ config SOF_STACK_SIZE
 
 config SECONDARY_CORE_DISABLING_TIMEOUT
 	int
-	default 5000
+	default 400
 	depends on MULTICORE
 	help
 	  Timeout value (in ms) for secondary core to enter D3 state.


btian1 (approved):
assumed 100ms used to do communication and disabling is enough?

src/platform/Kconfig
```diff
@@ -467,7 +467,7 @@ config SOF_STACK_SIZE
 	  components.
 
 config SECONDARY_CORE_DISABLING_TIMEOUT
-	int
+	int "Secondary core disabling timeout"
 	default 5000
```
kv2019i, line 471: @RanderWang Can you make a patch that changes the default to 400ms instead? SOF driver and firmware are hosted in same public project, so it makes sense to adapt any defaults between the two main entities. Special configs (like for FPGA) should be the exception, not the norm. Otherwise we need to keep doing the same change for all new platforms..

kv2019i (changes requested):
Let's just change the default.

src/platform/Kconfig
RanderWang: @kv2019i  updated, thanks

kv2019i (approved):
Thanks @RanderWang !

src/platform/Kconfig
btian1: FPGA does need 5000ms, in future, if SOF need run on FPGA again, we'd better to have a mechanism to change it back.

PR #8153, opened 2023-09-05
[DNM / TEST] ipc4: idc: Print error message if trigger delay is ignored

ujfalusi:
Are we ignoring the delayed trigger on a remote core? Is it happening?
Print out an error if it does.

### DIFF
--- a/src/idc/idc.c
+++ b/src/idc/idc.c
@@ -349,8 +349,15 @@ static int idc_ppl_state(uint32_t ppl_id, uint32_t phase)
 
 	if (phase & IDC_PPL_STATE_PHASE_TRIGGER) {
 		bool delayed = false;
+		int ret;
 
-		return ipc4_pipeline_trigger(ppl_icd, cmd, &delayed);
+		ret = ipc4_pipeline_trigger(ppl_icd, cmd, &delayed);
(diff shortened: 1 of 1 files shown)


src/idc/idc.c
```diff
@@ -349,8 +349,15 @@ static int idc_ppl_state(uint32_t ppl_id, uint32_t phase)
 
 	if (phase & IDC_PPL_STATE_PHASE_TRIGGER) {
 		bool delayed = false;
+		int ret;
+
+		ret = ipc4_pipeline_trigger(ppl_icd, cmd, &delayed);
+		if (delayed)
+			tr_err(&idc_tr,
+			       "[NOT_RIGHT] idc: comp %d: cmd %d is delayed!!",
```
lgirdwood, line 357: I would change to "error"

lgirdwood (commented):
Good check.

src/idc/idc.c
ujfalusi: It has not trigger ed in the first test plan run, I will trigger a new one to see if it ever happens.
In theory the likeliness of this check to trigger should be the same as we have delayed trigger in the primary core.

Issue #8148, opened 2023-09-05
[BUG][MTL] Firmware boot failure due to timeout during suspend/resume stress test (ROM status 0x50000005, ROM error 0x0)


keqiaozhang:
**Describe the bug**
This issue happened when testing suspend/resume with audio. It only happened once with stress test.

**dmesg**
```
[51369.325401] kernel: snd_sof_intel_hda_common:mtl_dsp_cl_init: sof-audio-pci-intel-mtl 0000:00:1f.3: Primary core power up successful
[51369.325408] kernel: snd_sof_intel_hda_common:mtl_dsp_cl_init: sof-audio-pci-intel-mtl 0000:00:1f.3: FW Poll Status: reg[0x73214]=0x80000000 successful
[51369.325420] kernel: snd_sof_intel_hda_common:mtl_enable_interrupts: sof-audio-pci-intel-mtl 0000:00:1f.3: FW Poll Status: reg[0x1800]=0x41 successful
[51369.325429] kernel: snd_sof_intel_hda_common:mtl_enable_interrupts: sof-audio-pci-intel-mtl 0000:00:1f.3: FW Poll Status: reg[0x1140]=0x1 successful
[51369.337199] kernel: nvme nvme0: 8/0/0 default/read/poll queues
[51371.343143] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ DSP dump start ]------------
[51371.343150] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: Firmware boot failure due to timeout
[51371.343153] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: fw_state: SOF_FW_BOOT_IN_PROGRESS (3)
[51371.343197] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ROM status: 0x0, ROM error: 0x0
[51371.343200] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ROM debug status: 0x50000005, ROM debug error: 0x0
[51371.343205] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ROM feature bit enabled
[51371.343207] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ DSP dump end ]------------
[51371.343209] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: error: failed to boot DSP firmware after resume -5
[...]

Issue #7866, opened 2023-06-27
[BUG][MTL] S0ix long run failure, Firmware boot failure due to timeout (ROM status: 0x50000005, ROM error: 0x0)


RDharageswari:
EDIT: the kernel part of this issue has been transferred to newer:
- https://github.com/thesofproject/linux/issues/4608

------------

We see this error reported in multiple scenario:
One was long run S0ix testing.
```
[  352.971033] sof-audio-pci-intel-mtl 0000:00:1f.3: ipc timed out for 0x47000000|0x0
[  352.979537] sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ IPC dump start ]------------
[  352.988991] sof-audio-pci-intel-mtl 0000:00:1f.3: Host IPC initiator: 0xc7000000|0x0|0x0, target: 0x67000000|0x0|0x0, ctl: 0x3
[  353.001735] sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ IPC dump end ]------------
[  353.010979] sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ DSP dump start ]------------
[  353.020423] sof-audio-pci-intel-mtl 0000:00:1f.3: IPC timeout
[  353.026868] sof-audio-pci-intel-mtl 0000:00:1f.3: fw_state: SOF_FW_BOOT_COMPLETE (7)
[  353.035552] sof-audio-pci-intel-mtl 0000:00:1f.3: ROM status: 0xffffffff, ROM error: 0xffffffff
[  353.045287] sof-audio-pci-intel-mtl 0000:00:1f.3: ROM debug status: 0x50000005, ROM debug error: 0x0
[  353.055497] sof-audio-pci-intel-mtl 0000:00:1f.3: ROM feature bit not enabled
[  353.063477] sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ DSP dump end ]------------
[  353.072721] sof-audio-pci-intel-mtl 0000:00:1f.3: ctx_save IPC error: -110, proceeding with suspend
[  689.764439] sof-audio-pci-intel-mtl 0000:00:1f.3: GAIN (UUID: 61BCA9A8-18D0-4A18-8E7B-2639219804B7): No CPC match in the firmware file's manifest (ibs/obs: 768/768)
[  692.922594] sof-audio-pci-intel-mtl 0000:00:1f.3: ipc timed out for 0x47000000|0x0
[...]

PR #8156, opened 2023-09-05
[v2.7] Disable IMR context save for MTL 

mengdonglin:
This is a follow up of PR #7994. We want to make sure it's safe to disable IMR context save for MTL and other platforms in sof v2.7

Please find more background info in https://github.com/thesofproject/sof/pull/7994#issuecomment-1706131833

We also want to check if disabling IMR context save can avoid the FW boot failure:
- https://github.com/thesofproject/sof/issues/7866#issuecomment-1700880432
- https://github.com/thesofproject/sof/issues/8148


tmleman (approved):

zephyr/lib/cpu.c
```diff
@@ -148,7 +148,7 @@ int cpu_enable_core(int id)
 	 * initialization. By reinitializing the idle thread, we would overwrite the kernel structs
 	 * and the idle thread stack.
 	 */
-	if (!IS_ENABLED(CONFIG_ADSP_IMR_CONTEXT_SAVE) ||
+	if (!defined(CONFIG_ACE) ||
```
ranj063, line 151: @mengdonglin looks good. Can we also update the comment above to call out the difference between cavs and ACE?

mengdonglin: @ranj063 Thanks for your review. I updated the comments in the code.

RanderWang (approved):
It is a temporary solution for v2.7 and it is not fit to check hw platform in common framework. I can add dsp restore support on CAVS if I am free

zephyr/lib/cpu.c
```diff
@@ -143,14 +143,25 @@ int cpu_enable_core(int id)
 		return 0;
 
 #if ZEPHYR_VERSION(3, 0, 99) <= ZEPHYR_VERSION_CODE
-	/* During kernel initialization, the next pm state is set to ACTIVE. By checking this
+	/* Zephyr implementations on secondary cores initialzation are
+	 * different between MTL+ and other platforms.
+	 */
+#if defined(CONFIG_ACE)
+	/* On MTL+ platforms, only run z_init_cpu() on boot, not when waking up from D3,
+	 * regardless whether saving to IMR is enabled or not.
```
lyakh, line 151: this is not how I understood the flow in a discussion yesterday. If on MTL+ IMR context saving isn't enabled, then waking up from D3 will be equivalent to the first boot, so the `pm_state_next_get(id)->state == PM_STATE_ACTIVE` condition will be true and `z_init_cpu()` will be run.

lyakh (changes requested):

zephyr/lib/cpu.c
mengdonglin: > this is not how I understood the flow in a discussion yesterday. If on MTL+ IMR context saving isn't enabled, then waking up from D3 will be equivalent to the first boot, so the `pm_state_next_get(id)->state == PM_STATE_ACTIVE` condition will be true and `z_init_cpu()` will be run.

@tmleman Can you help clarify?
I also updated the PR according to your https://github.com/thesofproject/sof/commit/fa1572c4d5668bda0867d46c687808bcdd800458 for mtl-006.

tmleman: On MTL `z_init_cpu(id)` is called for secondary cores only for the first init. As I wrote in the comment in the code:

> During kernel initialization, the next pm state is set to ACTIVE.

If core goes from D0 to D3, next state is set to `PM_STATE_SOFT_OFF`. Thanks to this we know if the core is enabled for the first time.

`CONFIG_ADSP_IMR_CONTEXT_SAVE` changes the behavior of the primary core during D3. If disabled, every exit from D3 for primary core is like first boot. Therefore, if we re-enable the secondary core after that, the function `z_init_cpu(id)` will be called.

zephyr/lib/cpu.c
```diff
@@ -143,12 +143,19 @@ int cpu_enable_core(int id)
 		return 0;
 
 #if ZEPHYR_VERSION(3, 0, 99) <= ZEPHYR_VERSION_CODE
-	/* During kernel initialization, the next pm state is set to ACTIVE. By checking this
+	/* Zephyr implementations on secondary cores initialzation are
+	 * different between MTL+ and other platforms.
+	 *
+	 * On MTL+ platforms, only run z_init_cpu() on boot, not when waking up from D3,
+	 * regardless whether saving to IMR is enabled or not.
```
plbossart, line 150: that seems wrong to me @mengdonglin 

What is the difference between 'boot'  and 'waking up from D3' if there is no context saved in IMR?

plbossart (changes requested):
I don't think the comments are quite right.

kv2019i:
So seems we are only discussng about the comments now. We do really need to get this in to keep to the SOF2.7 timeline. 

Let me propose alternative wording based on comments so far.
https://github.com/thesofproject/sof/pull/8170

marc-hb:
> So seems we are only discussing about the comments now.

Ignoring... test failures?
- https://github.com/thesofproject/sof/pull/8170#issuecomment-1709010997

kv2019i:
Merged via https://github.com/thesofproject/sof/pull/8170 

We'll follow-up to clarify the z_init_cpu() code (and comments) in mainline to get rid of platform specific checks on SOF side.

PR #8161, opened 2023-09-06
audio: src: refactor src component with more common code in src source file

btian1:
move out all ipc3 and ipc4 specifc code, data structure and functions.


iuliana-prodan:
@btian1 why are you splitting IPC3 vs IPC4 on component level?
Do you (meaning Intel) plan of doing this for all audio components?
What is the purpose of this separation?

btian1:
> @btian1 why are you splitting IPC3 vs IPC4 on component level? Do you (meaning Intel) plan of doing this for all audio components? What is the purpose of this separation?

@iuliana-prodan , you can see that current module become more and more larger, and inside, ipc3 and ipc4 code are mixed together, make code hard to read/understand/maintain.

these changes have two purpose:
1. move ipc specific code to separate source file.
2. move back header file to source directory.

currently, volume was done, module adapter is in: #8144, this is the third one.

src/audio/src/src_uuid.h
```diff
@@ -0,0 +1,32 @@
...
+#define __SOF_AUDIO_SRC_UUID_H__
+
+#include <sof/lib/uuid.h>
+#include <sof/trace/trace.h>
+#include <sof/audio/src/src.h>
+
+#if CONFIG_IPC_MAJOR_4
+
+/* e61bb28d-149a-4c1f-b709-46823ef5f5a3 */
+DECLARE_SOF_RT_UUID("src", src_uuid, 0xe61bb28d, 0x149a, 0x4c1f,
```
lyakh, line 19: the name is confusing, this isn't a declaration, it's a definition - it actually defines objects. So, this shouldn't be in a header. Besides - this can be perfectly placed in src_ipc4.c and src_ipc3.c, they divide very well.

src/audio/src/src_uuid.h
```diff
@@ -0,0 +1,32 @@
...
+#elif CONFIG_IPC_MAJOR_3
+/* c1c5326d-8390-46b4-aa47-95c3beca6550 */
+DECLARE_SOF_RT_UUID("src", src_uuid, 0xc1c5326d, 0x8390, 0x46b4,
+		    0xaa, 0x47, 0x95, 0xc3, 0xbe, 0xca, 0x65, 0x50);
+
+#else
+#error "No or invalid IPC MAJOR version selected."
+#endif /* CONFIG_IPC_MAJOR_4 */
+
+DECLARE_TR_CTX(src_tr, SOF_UUID(src_uuid), LOG_LEVEL_INFO);
```
lyakh, line 31: this also defines an object, please, keep it in .c

src/audio/src/src_ipc3.c
```diff
@@ -0,0 +1,193 @@
...
+}
+
+int src_set_params(struct processing_module *mod, struct sof_sink __sparse_cache *sink)
+{
+	return 0;
+}
+
+void src_set_sink_params(struct comp_dev *dev, struct sof_source __sparse_cache *source,
+			 struct sof_sink __sparse_cache *sink)
+{
```
lyakh, line 83: well, this function now isn't really setting sink parameters? Please choose a different name

src/audio/src/src_uuid.h
btian1: I also want to put it ipc3 and ipc4 separately, however, if you check DECLARE_SOF_RT_UUID inside, you will find static definition, how it can be referred by src.c?

#define DECLARE_SOF_UUID(entity_name, uuid_name,			\
			 va, vb, vc,					\
			 vd0, vd1, vd2, vd3, vd4, vd5, vd6, vd7)	\
	__section(".static_uuids")					\
	static const struct sof_uuid_entry uuid_name ## _ldc = {	\
		{.a = va, .b = vb, .c = vc,				\
		 .d = {vd0, vd1, vd2, vd3, vd4, vd5, vd6, vd7}},	\
		entity_name "\0"					\
	}

lyakh: right, that defined an object called `src_uuid_ldc`, right? And that one is then used in `DECLARE_TR_CTX(src_tr, SOF_UUID(src_uuid), LOG_LEVEL_INFO);` So I would either duplicate that single line `DECLARE_TR_CTX(...);` in both ipc3 and ipc4 files or leave them all in src.c with `#if`s but still please don't include them in a header. By placing them in that header and **only** including it in src.c you effectively placing them there, so you can just do that explicitly instead.

btian1: there is another way to use src_uuid:
DECLARE_MODULE_ADAPTER(src_interface, src_uuid, src_tr);
hard to resolve this one.

src/audio/src/src/src_ipc4_int32_10_21_2500_5000.h
```diff
@@ -5,7 +5,7 @@
  */
 
 /** \cond GENERATED_BY_TOOLS_TUNE_SRC */
-#include <sof/audio/src/src.h>
+#include "../src.h"
```
singalsu, line 8: Why is the location changed? The coefficients are generic polyphase FIR coefficients for a fractional ratio and could be in theory used by some other SRC like component. It's not a big deal to move and move back if need but was this causing issues?

If needed please make this change also to SRC coefficients generator. It can be a separate PR. These files must be directly generated without any manual edits done.

btian1: @singalsu , intention is move all related header to one folder, you can check volume part, all volume related moved to src/audio/volume folder.

src/audio/src/src_uuid.h
btian1: then have to add extern part, I added, please check again, not sure this looks better or not.

lyakh: Yes, that now looks correct C. Someone might complain that macros like `DECLARE_SOF_RT_UUID()` and `DECLARE_TR_CTX()` are supposed to hide their implementation and you aren't supposed to access internals of the code, that they generate, but I'd say that isn't a major concern.

lyakh (approved):
The actual blocking comment has been addressed - thanks. It still would be good to rename `src_set_source_sink_params()` and of course this needs an approval from @singalsu when all his concerns have been addressed.

src/audio/src/src/src_ipc4_int32_10_21_2500_5000.h
singalsu: OK!

Can you please make another PR to change this similarly https://github.com/thesofproject/sof/blob/2c6c1b4d4fb636ea78dfe1136737c4b6baf5b21c/tools/tune/src/src_export_coef.m#L53

Try to run e.g. https://github.com/thesofproject/sof/blob/main/tools/tune/src/src_ipc4_int32.m in octave

and check that you get similar h-files as these edited.

singalsu: Also I think directory name "coefficients" would be better than "src" for the moved header files.

src/audio/src/src_ipc3.c
lgirdwood: @singalsu any suggestions for appropriate name.

lgirdwood (approved):
LGTM, name can be fixed incrementally.

src/audio/src/src/src_small_int32_table.h
```diff
@@ -8,27 +8,27 @@
...
-#include <sof/audio/coefficients/src/src_small_int32_4_5_4535_5000.h>
-#include <sof/audio/coefficients/src/src_small_int32_5_4_4535_5000.h>
-#include <sof/audio/coefficients/src/src_small_int32_5_6_4354_5000.h>
-#include <sof/audio/coefficients/src/src_small_int32_6_5_4354_5000.h>
-#include <sof/audio/coefficients/src/src_small_int32_7_8_4535_5000.h>
-#include <sof/audio/coefficients/src/src_small_int32_8_7_4535_5000.h>
-#include <sof/audio/coefficients/src/src_small_int32_20_21_4167_5000.h>
-#include <sof/audio/coefficients/src/src_small_int32_21_20_4167_5000.h>
-#include <sof/audio/src/src.h>
+#include "src_small_int32_1_2_2268_5000.h"
```
singalsu, line 11: This change needs to be updated (other PR) here https://github.com/thesofproject/sof/blob/2c6c1b4d4fb636ea78dfe1136737c4b6baf5b21c/tools/tune/src/src_export_table_2s.m#L85

src/audio/src/src/src_small_int32_table.h
```diff
@@ -8,27 +8,27 @@
...
+#include "src_small_int32_4_3_4535_5000.h"
+#include "src_small_int32_4_5_4535_5000.h"
+#include "src_small_int32_5_4_4535_5000.h"
+#include "src_small_int32_5_6_4354_5000.h"
+#include "src_small_int32_6_5_4354_5000.h"
+#include "src_small_int32_7_8_4535_5000.h"
+#include "src_small_int32_8_7_4535_5000.h"
+#include "src_small_int32_20_21_4167_5000.h"
+#include "src_small_int32_21_20_4167_5000.h"
+#include "../src.h"
```
singalsu, line 31: And this change here https://github.com/thesofproject/sof/blob/2c6c1b4d4fb636ea78dfe1136737c4b6baf5b21c/tools/tune/src/src_export_table_2s.m#L94

src/audio/src/src_ipc3.c
singalsu: In IPC4 build the same name function sets source and sink parameters and these component data members are updated in init() where directly received from kernel.

If IPC3 function name is changed, then it could be src_get_source_sink_params(), since in IPC3 these are retrieved from params walk result for buffers.

kv2019i (approved):
Looks good. https://github.com/thesofproject/sof/pull/8167 will conflict though, and needs to go first due to its dependencies (if it is ready today).

marcinszkudlinski (approved):

kv2019i:
D'oh, my bad, I checked @singalsu you had approved, but it seemed you had NOT. Please double-check the issues where addressed correctly. @btian1 please follow up

btian1:
> D'oh, my bad, I checked @singalsu you had approved, but it seemed you had NOT. Please double-check the issues where addressed correctly. @btian1 please follow up

@kv2019i , @singalsu , I think what Seppo mentioned are addressed, only matlab file change is not yet covered, it need use another PR to continue, no worry no this.

PR #8163, opened 2023-09-06
loging: add logging through probes

dabekjakub:
Logging with probes was not implemented. This implements ipc that enables logging with probe configuration.


src/ipc/ipc4/logging.c
```diff
@@ -150,27 +154,47 @@ int ipc4_logging_enable_logs(bool first_block,
...
+
+	if (data_offset_or_size < sizeof(struct ipc4_log_state_info))
+		return -EINVAL;
+
+	dcache_invalidate_region((__sparse_force void __sparse_cache *)data, data_offset_or_size);
+
+	log_state = (const struct ipc4_log_state_info *)data;
+
+	if (log_state->enable) {
+		probe_logging_init(probe_logging_hook);
```
kv2019i, line 179: Hmm, I think we have a bit of a conlict. In current code, probe loggign is enabled by adding a probe point for the logs:

sof/src/probe/probe.c:probe_point_add()

@abonislawski added this (see git history).

So not sure how probe add and IPC4 log enable should relate to each other. I guess we could support both ways to enable probe log. Need to investigate more how this works.

src/ipc/ipc4/logging.c
```diff
@@ -10,12 +10,16 @@
 #include <ipc4/base_fw.h>
 #include <ipc4/error_status.h>
 #include <ipc4/logging.h>
+#include <zephyr/logging/log_backend.h>
+#include <zephyr/logging/log.h>
```
lyakh, line 14: we group headers alphabetically and by directories, not by function. Please leave them together with other Zephyr headers

kv2019i: I think is to avoid the CONFIG_LOG_BACKEND_ADSP_MTRACE ifdef.

src/ipc/ipc4/logging.c
```diff
@@ -150,27 +154,47 @@ int ipc4_logging_enable_logs(bool first_block,
...
+		return -EINVAL;
+
+	if (data_offset_or_size < sizeof(struct ipc4_log_state_info))
+		return -EINVAL;
+
+	dcache_invalidate_region((__sparse_force void __sparse_cache *)data, data_offset_or_size);
+
+	log_state = (const struct ipc4_log_state_info *)data;
+
+	if (log_state->enable) {
```
kv2019i, line 178: mtrace and probes has not been exclusive before, so we'd have to somehow make them work side-by-side after this PR as well. Could you rework this to have  ashared enable_logs handling and then have a helper function to actually implement "enable" and "disable" and these functions could have the ifdefs for mtrace and probe. Then you could at least build both, and in worst case, we can log bugs if the two cannot be enabled at the same time.

kv2019i (changes requested):
Thanks, now I understand better. I think we should unify the IPC handling and at least put the mtrace/probe enable/dsiable calls in some common place so we can reuse the IPC code.

I think we can leave the policy call whether the two can be enabled to build time, but the code should at least allow to build with both.

src/ipc/ipc4/logging.c
dabekjakub: Logging through probes at the same time as mtrace ? Technically possible I think but it is useless since You only ever use one method. Windows will use probes and linux will use mtrace hence the distinction.
Logging by probes only happens only  if the logging is configured in Enable Logs function. Probe logging is always ready to go, but only starts on enable logs witch will set probes as active.

The very idea of how it is done now with ifdefs rubs me the wrong way - this should not look like that. This area needs an overhaul. So if possible I would like to keep theme separate for now. This may in future help in a way (I had an idea to save some func pointers and overwrite theme case by case - but it needs more time in "head oven").
That's why I would like to push it as is since it enables logging for windows and does not impact anyone else.

kv2019i: @dabekjakub We don't need to refactor the code now, agreed, but even if we do the minimal change, we can't break existing stuff. We have probes support in Linux as well, so we need to keep that working. I'm ok to make mtrace and probe exclusive  (at build time) but this needs to added as a ifdef check that raises a (clear) build error if both are selected.

Second, I don't still understand how this matches with "sof/src/probe.c:probe_point_add()" that also adds probe_logging_init(). So now you have two places in code where probe_logging_init() is called. I thought also Windows will use this API (this code was added by @abonislawski ). I mean somehow Windows also needs to set the DMA channel for probe, I'd assume it's a IPC4_PROBE_MODULE_PROBE_POINTS_ADD. 

If the two are exclusive, then I'm ok with the current code, but it looks strange we have two IPC4 handlers doing the same thing and with no code to handle concurrent use. Also, I don't see how this handles the case where ENABLE_LOGS is sent (with probes enabled), but PROBE_POINTS_ADD is not sent. You'd start calling the probe hook, but the DMA channel is not established, so it would seem the FW would crash in probe.c:probe_logging_hook() in this case. I know this the expected flow, but FW must not crash if sent IPCs in invalid sequence.

dabekjakub: I will test scenarios with probe adding on Monday - and will remove the duplicate calls.
Will add some assert for both enabled - this would be up for change in future.
However I disagree with FW being a flow controller. What would You check for ?  Crash should avoided if possible that is true but how would You handle it here ?

dabekjakub: Only then we have to follow the state of either DMA or flow. And FW should not ever handle the flow itself. The same situation would occur if You demanded playback at wrongly configurated gateway i think. Only way to do it semi correctly would be checking if hook was connected.

lgirdwood: all - can we make it best practice now to inline comment all flows around cache usage.  i.e. to make it obvious to the user why we are doing the cache operation and on what.

dabekjakub:
@kv2019i updated build to error if both are enabled - it would anyway from redefinition but there is an error for now.

Added function to check if backend is configured and result in error when not. 

Should be good to go now.

src/ipc/ipc4/logging.c
```diff
@@ -150,27 +158,49 @@ int ipc4_logging_enable_logs(bool first_block,
...
+		if (!probe_is_backend_configured())
+			return -EINVAL;
+
+		log_backend_activate(log_backend, probe_logging_hook);
+
+	} else  {
+		probe_logging_init(NULL);
```
kv2019i, line 190: Remove the init(NULL) here? Otherwise if you send ENABLE_LOGS with enable=1, enable=0 and enable=1, it won't work anymore.

kv2019i (changes requested):
Thanks @dabekjakub , looks good now! One small thing still in patch, can you take a look?

src/ipc/ipc4/logging.c
```diff
@@ -150,27 +158,49 @@ int ipc4_logging_enable_logs(bool first_block,
...
+	const struct ipc4_log_state_info *log_state;
+
+	if (!(first_block && last_block))
+		return -EINVAL;
+
+	if (data_offset_or_size < sizeof(struct ipc4_log_state_info))
+		return -EINVAL;
+
+	/* Make sure we work on correct ipc data by invalidating cache */
+	dcache_invalidate_region((__sparse_force void __sparse_cache *)data, data_offset_or_size);
```
lyakh, line 179: is this needed because data could've been written by a different core? Maybe write that explicitly in the comment

src/logging/log_backend_probe.c
```diff
@@ -109,3 +109,16 @@ void probe_logging_init(probe_logging_hook_t fn)
...
+{
+	return &log_backend_adsp_probe;
+}
+
+bool probe_is_backend_configured(void)
+{
+	if (probe_hook)
+		return true;
+	else
+		return false;
```
lyakh, line 123: how about one of
```
return !!probe_hook;
return probe_hook != NULL;
```

kv2019i (approved):
Thanks, looks good now!

kv2019i:
One fail in pause-resume in https://sof-ci.01.org/sofpr/PR8163/build12942/devicetest , but does not seem to be related to this PR (and no errors in either dmesg or mtrace, aplay just returns non-zero on one iteration, rest are fine). Proceeding with merge.

marc-hb:
https://github.com/thesofproject/sof/actions/runs/6187233612/job/16796644911 was failing. Fixed in #8222

PR #8166, opened 2023-09-06
Remove sparse step1

marcinszkudlinski:
this is a step 1 of https://github.com/thesofproject/sof/pull/8106

### DIFF
--- a/src/audio/buffer.c
+++ b/src/audio/buffer.c
@@ -53,6 +53,8 @@ struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t flags, u
 		return NULL;
 	}
 
+	CORE_CHECK_STRUCT_INIT(buffer, is_shared);
+
 	buffer->is_shared = is_shared;
 	stream_addr = rballoc_align(0, caps, size, align);
 	if (!stream_addr) {
@@ -81,6 +83,7 @@ struct comp_buffer *buffer_alloc(uint32_t size, uint32_t caps, uint32_t flags, u
 void buffer_zero(struct comp_buffer __sparse_cache *buffer)
 {
 	buf_dbg(buffer, "stream_zero()");
+	CORE_CHECK_STRUCT(buffer);
 
 	bzero(audio_stream_get_addr(&buffer->stream), audio_stream_get_size(&buffer->stream));
 	if (buffer->caps & SOF_MEM_CAPS_DMA)
@@ -92,6 +95,8 @@ void buffer_zero(struct comp_buffer __sparse_cache *buffer)
 int buffer_set_size(struct comp_buffer __sparse_cache *buffer, uint32_t size, uint32_t alignment)
 {
 	void *new_ptr = NULL;
+
+	CORE_CHECK_STRUCT(buffer);
 
 	/* validate request */
 	if (size == 0) {
@@ -131,6 +136,8 @@ int buffer_set_params(struct comp_buffer __sparse_cache *buffer,
 	int ret;
 	int i;
 
+	CORE_CHECK_STRUCT(buffer);
+
 	if (!params) {
 		buf_err(buffer, "buffer_set_params(): !params");
 		return -EINVAL;
@@ -158,6 +165,7 @@ bool buffer_params_match(struct comp_buffer __sparse_cache *buffer,
 			 struct sof_ipc_stream_params *params, uint32_t flag)
 {
 	assert(params);
+	CORE_CHECK_STRUCT(buffer);
 
 	if ((flag & BUFF_PARAMS_FRAME_FMT) &&
 	     audio_stream_get_frm_fmt(&buffer->stream) != params->frame_fmt)
@@ -180,6 +188,8 @@ void buffer_free(struct comp_buffer *buffer)
 	struct buffer_cb_free cb_data = {
 		.buffer = buffer,
 	};
+
+	CORE_CHECK_STRUCT(buffer);
 
 	if (!buffer)
 		return;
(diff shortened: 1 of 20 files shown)


btian1 (approved):

RanderWang (approved):

src/ipc/ipc-helper.c
```diff
@@ -174,12 +175,12 @@ int comp_buffer_connect(struct comp_dev *comp, uint32_t comp_core,
...
 	if (buffer->core != comp_core) {
-		/* set the buffer as a coherent object */
-		coherent_shared_thread(buffer, c);
+		/* buffer must be shared */
+		assert(buffer->is_shared);
 
 		if (!comp->is_shared)
 			comp_make_shared(comp);
-	}
+		}
```
kv2019i, line 183: Indent off on L183

kv2019i (approved):

kv2019i:
System PM test fails in https://sof-ci.01.org/sofpr/PR8166/build12642/devicetest/index.html and https://sof-ci.01.org/sofpr/PR8166/build12641/devicetest/index.html but occur with baseline. Main problem is the System/merge/build mandatory check now fails. @marcinszkudlinski can you check?

marcinszkudlinski:
> System PM test fails in https://sof-ci.01.org/sofpr/PR8166/build12642/devicetest/index.html and https://sof-ci.01.org/sofpr/PR8166/build12641/devicetest/index.html but occur with baseline. Main problem is the System/merge/build mandatory check now fails. @marcinszkudlinski can you check?

Problem is with IPC Probes

marcinszkudlinski:
Fixed - required commit moved from "step2" to step1

kv2019i:
Waiting for @lyakh to ack and for mandatory CI to pass.

src/audio/buffer.c
```diff
@@ -335,32 +301,6 @@ void buffer_attach(struct comp_buffer *buffer, struct list_item *head, int dir)
  */
 void buffer_detach(struct comp_buffer *buffer, struct list_item *head, int dir)
```
lyakh, line 302: `head` is now unneeded, can be removed later

src/include/sof/audio/buffer.h
```diff
@@ -144,6 +142,7 @@ struct comp_buffer {
 	uint32_t caps;
 	uint32_t core;
 	struct tr_ctx tctx;			/* trace settings */
+	bool is_shared;			/* buffer structure is shared between 2 cores */
```
lyakh, line 147: I see this checked in a couple of places, but I don't see it assigned anywhere?

src/include/sof/coherent.h
```diff
@@ -80,9 +80,23 @@ struct coherent {
 
 /* debug sharing amongst cores */
 #ifdef COHERENT_CHECK_NONSHARED_CORES
+
+#define CORE_CHECK_STRUCT_FIELD uint32_t __core; bool __is_shared
+#define CORE_CHECK_STRUCT_INIT(_c, is_shared) { (_c)->__core = cpu_get_id(); \
+						(_c)->__is_shared = is_shared; }
+#define CORE_CHECK_STRUCT(_c) { assert(!!(_c)->__is_shared == !!is_uncached(_c)); \
+				assert(cpu_get_id() == (_c)->__core || (_c)->__is_shared); }
+
```
lyakh, line 89: these fields are now not in `struct coherent` so I don't think they're needed in this header. Besides, `struct comp_buffer` already has `bool is_shared` and presumably it will be assigned close enough to where you now assign `__is_shared` so we can just reuse the former?

lyakh (changes requested):

src/include/sof/audio/buffer.h
marcinszkudlinski: good catch, fixed

src/ipc/ipc-helper.c
```diff
@@ -174,8 +175,8 @@ int comp_buffer_connect(struct comp_dev *comp, uint32_t comp_core,
 {
 	/* check if it's a connection between cores */
 	if (buffer->core != comp_core) {
-		/* set the buffer as a coherent object */
-		coherent_shared_thread(buffer, c);
+		/* buffer must be shared */
+		assert(buffer->is_shared);
```
lgirdwood, line 179: Is this recoverable ? I assume this data is derived from topology so could a bad topology cause an assert here ?

lgirdwood (approved):
We can fix the assert later if needed, but need to get this in CI.

src/ipc/ipc-helper.c
marcinszkudlinski: Not recoverable, must not happen. 
If happens, a cached alias will be used for shared memory and system will crash later

src/include/sof/coherent.h
lyakh: I think I'd do something like
```
struct multicore {
    bool shared;
#if CONFIG_MULTIPLE_DEBUG
    int core:
#endif
};
```
somewhere in src/include/sof, maybe in src/include/sof/common.h or a new src/include/sof/multicore.h, then pass a pointer to that structure to `CORE_CHECK_STRUCT()` and other your checking macros, and use `.shared` for both debugging and run-time

src/ipc/ipc-helper.c
lyakh: but here at least you have a perfect chance to error out, don't you?

kv2019i:
@lyakh @marcinszkudlinski Is the only blocker now the debug feature implementation? Can this be handled in separate PR (given multiple PRs are anyways planned to for this effort).

kv2019i:
Let's proceed now, and address the debug related issues later if needed (FYI @marcinszkudlinski ).

kv2019i:
Zephyr mainlin builds are failing because of https://github.com/zephyrproject-rtos/zephyr/issues/62464

PR #8169, opened 2023-09-06
Merge PR8131 to mtl-006-drop-stable branch

udaymb:
Merge the patch from [pr#8131](https://github.com/thesofproject/sof/pull/8131) to mtl-006-drop-stable branch

NUM_HDMIS is set to 3 as default value for MTL platform. SPK_ID and BT_ID are modified based on this change in the target cmake file.


lgirdwood:
@mwasko @marcinszkudlinski fyi

PR #8172, opened 2023-09-06
audio: google_rtc: Remove set_large_config op for IPC4

ranj063:
The input and output audio formats are sent during module init already. So no need to overwrite it with topology data that could be incorrect.

### DIFF
--- a/src/audio/google/google_rtc_audio_processing.c
+++ b/src/audio/google/google_rtc_audio_processing.c
@@ -704,30 +704,6 @@ static int google_rtc_audio_processing_copy(struct comp_dev *dev)
 }
 
 #if CONFIG_IPC_MAJOR_4
-static int google_rtc_audio_processing_set_large_config(struct comp_dev *dev,
-							uint32_t param_id,
-							bool first_block,
-							bool last_block,
-							uint32_t data_offset,
-							const char *data)
-{
-	struct google_rtc_audio_processing_comp_data *cd = comp_get_drvdata(dev);
-	size_t bcfg_size, ext_size;
-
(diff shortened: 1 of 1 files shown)


yongzhi1 (approved):

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -704,30 +704,6 @@ static int google_rtc_audio_processing_copy(struct comp_dev *dev)
...
-							const char *data)
-{
-	struct google_rtc_audio_processing_comp_data *cd = comp_get_drvdata(dev);
-	size_t bcfg_size, ext_size;
-
-	switch (param_id) {
-	case IPC4_AEC_SET_EXT_FMT:
-		bcfg_size = sizeof(struct ipc4_base_module_cfg);
-		ext_size = sizeof(cd->config) - bcfg_size;
-		memcpy_s((void *)&cd->config + bcfg_size, ext_size, data, ext_size);
```
btian1, line 721: so frame format is fixed, and can't be changed after module_init, if there is new stream with RTC, then module_init will be called again.

ranj063: no it wont. the module is initialized only once

kv2019i (approved):

kv2019i:
Code looks good. Unexpected fails in https://sof-ci.01.org/sofpr/PR8172/build12671/devicetest/index.html -- can't immediately see how these can be connected to this PR, but needs to be investigated before merge.

kv2019i:
SOFCI TEST

PR #8173, opened 2023-09-07
[TEST IGNORE] sof-glk-nocodec: disable pipelines when disabling SSPs

marc-hb:
When we added the flags to disable SSP0 and SSP1 on the UP2, we took the shortcut of just removing the PCMs in topology but left the pipelines and widgets in the topology in. While this works in practice to prevent us from testing those SSPs, the right way is to also remove those pipelines also when the SSPs are disabled.

This stops tplgtool2.py that complained constantly about this inconsistency since https://github.com/thesofproject/sof-test/pull/1079 and made the sof-test verify-tplg-binary.sh fail every time:

```
tplgtool2.py sof-glk-nocodec.tplg

ERROR: No pcm id=0 for widget=PCM0C
ERROR: No pcm id=1 for widget=PCM1C
ERROR: No pcm id=0 for widget=PCM0P
ERROR: No pcm id=1 for widget=PCM1P
ERROR: tplgtool2.py returned 4
```

This change affects only sof-apl-nocodec and sof-glk-nocodec.

PR #8175, opened 2023-09-07
sof-glk-nocodec: disable pipelines when disabling SSPs

marc-hb:
When we added the flags to disable SSP0 and SSP1 on the UP2, we took the shortcut of just removing the PCMs in topology but left the pipelines and widgets in the topology in. While this works in practice to prevent us from testing those SSPs, the right way is to also remove those pipelines also when the SSPs are disabled.

This stops tplgtool2.py from complaining constantly about this inconsistency since https://github.com/thesofproject/sof-test/pull/1079 which made the sof-test verify-tplg-binary.sh fail every time:

```
tplgtool2.py sof-glk-nocodec.tplg

ERROR: No pcm id=0 for widget=PCM0C
ERROR: No pcm id=1 for widget=PCM1C
ERROR: No pcm id=0 for widget=PCM0P
ERROR: No pcm id=1 for widget=PCM1P
ERROR: tplgtool2.py returned 4
```

This change affects only sof-apl-nocodec and sof-glk-nocodec.

### DIFF
--- a/tools/topology/topology1/sof-cavs-nocodec.m4
+++ b/tools/topology/topology1/sof-cavs-nocodec.m4
@@ -101,17 +101,19 @@ dnl     time_domain, sched_comp)
 
 # Volume switch capture pipeline 2 on PCM 0 using max 2 channels of PIPE_BITS.
 # Set 1000us deadline on core SSP0_CORE_ID with priority 0
-PIPELINE_PCM_ADD(sof/pipe-volume-switch-capture.m4,
+ifdef(`DISABLE_SSP0',,
+	`PIPELINE_PCM_ADD(sof/pipe-volume-switch-capture.m4,
 	2, 0, 2, PIPE_BITS,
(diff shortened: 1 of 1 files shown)


plbossart (approved):
looks ok but @ranj063 needs to review this

ranj063:
> looks ok but @ranj063 needs to review this

approved :). @plbossart this is my patch

bardliao (approved):

lgirdwood (approved):

marc-hb:
Wait, we looked at the graph and found a problem. @ranj063 gave me a new patch.

PS: please don't promote others drafts.

marc-hb:
New version from @ranj063 force-pushed, this time it's ready. See new, smaller graph and test results in stable-v2.2 TEST PR #8173.

This patch has no effect on the main branch, I will cherry-pick -x after merge to stable-v2.2 (sticking to the normal flow)

PR #8176, opened 2023-09-07
[stable-v2.2] Tools: Topology1: Multiband-DRC: Update setup blobs

singalsu:
The multiband_drc_coef_default.m4 is updated to what the current setup tool example_multiband_drc() exports.

Another blob for pass-through is added with emphasis and all bands processing disabled. Such blob is useful when headphone and speaker paths are shared. Headphone mode should use pass-through.


singalsu:
"21 successful and 1 failing checks"

There is no multiband-drc instance in CI tests. The verify-tplg-binary.sh fail with sof-apl-nocodec.tplg is not related, this happens also with stable-2.2 as such.

tools/topology/topology1/m4/multiband_drc_coef_default.m4
```diff
@@ -1,7 +1,7 @@
-# Exported Control Bytes 11-May-2021
+# Exported Control Bytes 16-Aug-2023
```
lgirdwood, line 1: Whats the command line to reproduce ?

lgirdwood (commented):
Topology2 versions higher priority than topology1

tools/topology/topology1/m4/multiband_drc_coef_default.m4
singalsu: Seems I need to update #8142 to update this comment for tplg1 too. This change is already merged to main so I'd need to update it there as well. But I don't think the setup tool update is worth to do for v2.2. It would need to be a new middle version of v2.2 and git main without tplg2 part.

The tplg2 is already in order. This tplg1 part was split out from the tplg2 patch.

kv2019i:
@singalsu what's the direction with this?

singalsu:
The tool is still missing the howto add for tplg1. I'll close this PR and update later. @johnylin76 provided a better performing blob that we can take into use. I'll return to this topic with it.

PR #8177, opened 2023-09-07
Audio: Kconfig: Deprecate tone component

singalsu:
Tone component has been without maintenance and validation, and there is no interface to control sound generation from user space so it will be deprecated and removed from successive release.


singalsu:
@kv2019i This should get into SOF v2.7

kv2019i (approved):
Thanks, let's get to main first. Might be a bit tight for rc1, let's see if we can get this in still (depends on whether there will be a rc2 or not).

lgirdwood (approved):

PR #8180, opened 2023-09-08
[WIP] loadable libraries and smart-amp-test as an example

lyakh:
This implements support for loadable libraries in SOF and adds the necessary interface to the smart-amp-test to be built and used as such a module. Still WiP


lmdk/pic/smart_amp_test/smart_amp_test.toml
```diff
@@ -0,0 +1,80 @@
+version = [3, 0]
```
lgirdwood, line 1: We should really run the pre-processor on the toml files like we do with linker scripts. @kv2019i fyi.

lmdk/pic/smart_amp_test/smart_amp_test.toml
```diff
@@ -0,0 +1,80 @@
...
+type = "SRAM"
+base = "0xa00f0000"
+size = "0x100000"
+
+[[adsp.mem_alias]]
+type = "uncached"
+base = "0x40000000"
+[[adsp.mem_alias]]
+type = "cached"
+base = "0xA0000000"
```
lgirdwood, line 26: We can the #include <platform.tml> for above

lmdk/pic/smart_amp_test/smart_amp_test.toml
```diff
@@ -0,0 +1,80 @@
...
+[[signed_pkg.module]]
+name = "ADSP.met"
+
+[adsp_file]
+[[adsp_file.comp]]
+base_offset = "0x2000"
+
+[fw_desc.header]
+name = "ADSPFW"
+load_offset = "0x40000"
```
lgirdwood, line 57: We can the #include <vendor.tml>

src/library_manager/lib_manager.c
```diff
@@ -423,15 +423,13 @@ static int lib_manager_dma_buffer_alloc(struct lib_manager_dma_ext *dma_ext,
 	 * allocate new buffer: this is the actual DMA buffer but we
 	 * traditionally allocate a cached address for it
 	 */
-	dma_ext->dma_addr = (uintptr_t)rballoc_align(0, SOF_MEM_CAPS_DMA, size,
```
lgirdwood, line 426: @lyakh is this patch neededd for mtl-06 i.e. is it fixing a bug

src/idc/zephyr_idc.c
```diff
@@ -63,7 +63,7 @@ static void idc_handler(struct k_p4wq_work *work)
 	int payload = -1;
 	k_spinlock_key_t key;
 
-	sys_cache_data_flush_range(msg, sizeof(*msg));
+	sys_cache_data_invd_range(msg, sizeof(*msg));
```
lgirdwood, line 66: Fix needed for mtl-06 ? and main as seperate PR ?

src/audio/eq_fir/eq_fir.c
```diff
@@ -636,7 +636,7 @@ static int eq_fir_reset(struct processing_module *mod)
 	return 0;
 }
 
-static struct module_interface eq_fir_interface = {
+static const struct module_interface eq_fir_interface = {
```
lgirdwood, line 639: seperate patch ?

src/include/sof/audio/module_adapter/module/generic.h
```diff
@@ -172,6 +172,12 @@ struct module_source_info {
 	void *private; /**< additional module-specific private info */
 };
 
+enum module_processing_type {
+	MODULE_PROCESS_TYPE_SOURCE_SINK,
+	MODULE_PROCESS_TYPE_STREAM,
+	MODULE_PROCESS_TYPE_RAW,
+};
```
lgirdwood, line 179: Can we doxygen comment this enaum and what each means.

src/audio/module_adapter/library/native_system_agent.c
```diff
@@ -30,9 +30,6 @@ void *native_system_agent_start(uint32_t *sys_service,
 	native_sys_agent.log_handle = log_handle;
 
 	void *system_agent_p = &native_sys_agent;
-	uint32_t **sys_service_p = &sys_service;
-
-	*sys_service_p = (uint32_t *)(&native_sys_agent.system_service);
```
lgirdwood, line 35: seperate fix ?

src/audio/module_adapter/module/generic.c
```diff
@@ -222,8 +222,7 @@ int module_prepare(struct processing_module *mod,
 	 * as it has been applied during the procedure - it is safe to
 	 * free it.
 	 */
-	if (md->cfg.data)
-		rfree(md->cfg.data);
+	rfree(md->cfg.data);
```
lgirdwood, line 225: seperate patch

src/ipc/ipc4/handler.c
```diff
@@ -782,7 +782,7 @@ static int ipc4_process_glb_message(struct ipc4_message_request *ipc4)
 
 static int ipc4_init_module_instance(struct ipc4_message_request *ipc4)
 {
-	struct ipc4_module_init_instance module_init = {};
+	struct ipc4_module_init_instance module_init;
```
lgirdwood, line 785: seperate fix.

lmdk/cmake/build.cmake
```diff
@@ -68,10 +71,18 @@ foreach(MODULE ${MODULES_LIST})
 		#"-Wl,--no-undefined" "-Wl,--unresolved-symbols=report-all" "-Wl,--error-unresolved-symbols"
 		#"-Wl,--gc-sections"	# may remove .bss and that will result in rimage error, do not use for now
 		"-Wl,-Map,$<TARGET_FILE:${MODULE}>.map"	# optional: just for debug
-		"-T" "${MODULE}_ldscripts/elf32xtensa.x"
+		#"-T" "${MODULE}_ldscripts/elf32xtensa.x"
+		# FIXME: use a script to calculate values
+		"-Wl,-Ttext=0xa06ca000"
+		"-Wl,-Tdata=0xa06cb000"
+		"-Wl,--section-start=.rodata=0xa06cb100"
```
lgirdwood, line 78: magic - are we copied here ?

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -39,6 +39,8 @@ struct smart_amp_data {
 	uint32_t out_channels;
 };
 
+static int keep_bss;
```
lgirdwood, line 42: needs a comment.

lgirdwood (commented):
Looks like a lot of patches here are ready to be sent as seperate PRs.
Additionally, we need to use the same module build and linking methodolody as Linux i.e. .o file instead of shared libraries.

lmdk/pic/smart_amp_test/smart_amp_test.toml
kv2019i: @lgirdwood @aiChaoSONG Hmm, given we always have one platform and one algo bit, not sure we need to go that far. But for sure this is one option for https://github.com/thesofproject/rimage/pull/176 One benefit is that we'll have cpp on all platforms, so this would cover the portability part.

lgirdwood: so upgrading to use cpp here would also be easy enough to do and also allow us to pick UUIDs and memory addreses and sizes from headers (instead of coding them more than once).

lmdk/cmake/build.cmake
lyakh: there's a comment there - need a script to calculate these. The approach is rather simple: .text is fixed (per module? to make them unique?) the .data should then be page-aligned. Haven't been able to achieve that in a better way so far

src/library_manager/lib_manager.c
lyakh: @lgirdwood I haven't seen any run-time failures that this patch would fix. I think "strangely" DMA on ADSP can work both with cached and uncached pointers, although it actually never uses cache. So I just find using cached pointers for DMA incorrect and confusing - it gives an impression of DMA using cache. So I think this is a correctness patch, but I don't know if it actually fixes anything. Besides this is lib-manager which isn't used yet anyway.

lmdk/cmake/build.cmake
```diff
@@ -9,14 +9,43 @@ endif()
...
-		"${LMDK_BASE}/include"
+		"${SOF_BASE}/src/include"
+		"${SOF_BASE}/src/include/sof/audio/module_adapter/iadk"
+		"${SOF_BASE}/src/include/ipc"
+		"${SOF_BASE}/src/include/ipc4"
+		"${SOF_BASE}/src/include/math"
+		"${SOF_BASE}/src/include/lib"
+		"${SOF_BASE}/src/arch/xtensa/include"
+		"${SOF_BASE}/zephyr/include"
+		"${SOF_BASE}/xtos/include"
```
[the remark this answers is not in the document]
lyakh: yes, needed for compiler_attributes.h, although might be possible to get rid of it - builds without it at least as a module too

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -65,7 +74,9 @@ static int smart_amp_init(struct processing_module *mod)
 
 	if (base_cfg->base_cfg_ext.nb_input_pins != SMART_AMP_NUM_IN_PINS ||
 	    base_cfg->base_cfg_ext.nb_output_pins != SMART_AMP_NUM_OUT_PINS) {
-		comp_err(dev, "smart_amp_init(): Invalid pin configuration");
+		comp_err(dev, "smart_amp_init(): Invalid pin configuration: %d %d",
```
lgirdwood, line 77: I would keep changes to bare minimum needed in any module. Chnages to log messages should be another PR.

src/audio/module_adapter/module/modules.c
```diff
@@ -295,10 +303,16 @@ static int modules_free(struct processing_module *mod)
 	rfree(md->mpd.in_buff);
 	rfree(md->mpd.out_buff);
 
+#if 0
+	/*
+	 * Module unloading isn't implemented yet, it has to be done from a
+	 * dedicated IPC, not when stopping streams.
+	 */
```
lgirdwood, line 310: Module unloading come from ref counting users based on running pipelines.

src/audio/module_adapter/module/modules.c
```diff
@@ -57,64 +57,73 @@ DECLARE_TR_CTX(intel_codec_tr, SOF_UUID(intel_uuid), LOG_LEVEL_INFO);
...
-	/* Check if module is FDK*/
-	if (mod_buildinfo->api_version_number.fields.major < SOF_MODULE_API_MAJOR_VERSION) {
-		mod_adp = system_agent_start(md->module_entry_point, module_id,
-					     instance_id, 0, log_handle, &mod_cfg);
-	} else {
-		/* If not start agent for sof loadable */
-		mod->is_native_sof = true;
-		mod_adp = native_system_agent_start(mod->sys_service, md->module_entry_point,
-						    module_id, instance_id, 0, log_handle,
-						    &mod_cfg);
```
lgirdwood: Can you go into more detail what is being modified here.

lgirdwood (commented):
It looks like we have some seperate fixes and patches in the PR and it seems like they could upstream now as separate PRs.
@pblaszko pls see if any of @lyakh makefile changes will help your PR to reuse headers.

lgirdwood (commented):
See its been updated to newer Zephyr. One other thing we need is to have the module specific infra in the module directory, I dont want a duplication of stuff in module/ and lmdk/

lyakh:
Not all of it should be merged (e.g. no need for Zephyr PR and we need to decide whether we set the smart-amp sample as module hard), but it's now close to the final version, I think, I'd say, almost no hacks.

lyakh:
and yes, I'll resolve conflicts tomorrow

src/audio/module_adapter/module/modules.c
lyakh: yes, forgot to update the comment, fixing, thanks

lgirdwood (commented):
@lyakh there are a lot of cosmetic, fixes, small updates that can be sent in different PRs today and should be applied without blocking on the larger features. Then this PR can be rebased and show the min changes for features.

scripts/calc_elf_addresses.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# We need to take addresses of .text, .data, .rodata and .bss and calculate
+# offsets, because the linker will keep everything between .data and .rodata and
+# between .rodata and .bss in the final image too, when we specify section
+# addresses. So just taking sizes of .text, .data and .rodata isn't enough.
+# We use file offset from the objdump command output to avoid unused sections.
+
+source <(
+objdump -h $file | awk '/ .text / {print "text="toupper($6)} / .rodata / {print "rodata="toupper($6)} / .data / {print "data="toupper($6)}'
+)
```
marc-hb, line 14: Please no parsing and arithmetic in shell script... this is really trivial in Python! Demo, copy/paste this in your terminal:

```
python -c "from elftools.elf.elffile import ELFFile
elf=ELFFile(open('zephyr.elf','rb'))
print(hex(elf.get_section_by_name('.text').header.sh_offset))"
```
=> 0xe400

Plenty more samples in:
- git -C zephyr/ grep ELFFile
- https://github.com/eliben/pyelftools/tree/master/examples

marc-hb: > git -C zephyr/ grep ELFFile


https://docs.zephyrproject.org/latest/build/cmake/index.html#supporting-scripts-and-tools

lyakh:
This is the first time that this PR works on top of the current "main" with no additional *essential* Zephyr changes. The only Zephyr addition that I am usint with this is https://github.com/zephyrproject-rtos/zephyr/pull/66573 but that is just `west sign` support for loadable modules. In principle it isn't needed to get this to run, one can use rimage directly. With this PR my testing sequence looks like:
1. build
```
west build -d build-mtl -b intel_adsp_ace15_mtpm modules/audio/sof/app/ -- -DOVERLAY_CONFIG="debug_overlay.conf"
west sign -d build-mtl -t rimage -p build-rimage/rimage --tool-data modules/audio/sof/rimage/config -- -k $key_path/key.pem
west sign -d build-mtl -t rimage -p build-rimage/rimage -- -k $key_path/key.pem -c modules/audio/sof/src/samples/audio/smart_amp_llext/smart_amp_test.toml -r build-mtl/zephyr/smart_amp_llext/libsmart_amp_test_out.so
```
2. to run just copy zephyr.ri and llext.ri to the usual locations

lyakh:
next: replace the bash script with python

scripts/calc_elf_addresses.sh
lyakh: @marc-hb https://github.com/thesofproject/sof/pull/8180/commits/072875bf3f8170d6795ea92239befc32d81025c0#diff-737b4acbcc477942bbd9b68cf553004b518f9d9d7ed8938d08f839ec3a240da3

src/library_manager/lib_manager.c
```diff
@@ -155,26 +155,13 @@ static void __sparse_cache *lib_manager_get_instance_bss_address(uint32_t module
...
-	void __sparse_cache *va_base =
-		(void __sparse_cache *)(mod->segment[SOF_MAN_SEGMENT_BSS].v_base_addr +
-					inst_offset);
-
-	tr_dbg(&lib_manager_tr,
-	       "lib_manager_get_instance_bss_address(): instance_bss_size: %#x, pointer: %p",
-	       instance_bss_size, (__sparse_force void *)va_base);
-
-	return va_base;
+	return (void __sparse_cache *)mod->segment[SOF_MAN_SEGMENT_BSS].v_base_addr;
```
jxstelter, line 197: This change will break IADK originated modules compatibility. IADK modules will not work in multiple instances.

src/samples/audio/smart_amp_llext/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
...
+target_link_options(${MODULE} PRIVATE
+	${MODULE_LINKER_PARAMS}
+)
+
+add_custom_target(${MODULE}_llext ${CMAKE_C_COMPILER}
+	${MODULE_LINKER_PARAMS}
+	"$$(${SOF_BASE}scripts/calc_elf_addresses.py" -f lib${MODULE}.so -t "A06CA000)"
+	-shared -fPIC
+	-o lib${MODULE}_llext.so $<TARGET_OBJECTS:${MODULE}>
+	COMMAND ${CMAKE_STRIP} -R .xt.* -o lib${MODULE}_out.so lib${MODULE}_llext.so
```
marc-hb, line 66: `tests/subsys/llext/hello_world/` uses an `.llext` suffix, why does this need to be called `lib{MODULE}_llext.so` like a UNIX library?

This is not just name bikeshedding: if we want the build system to smoothly scale to building and signing multiple modules then we're going to need some consistency.

scripts/calc_elf_addresses.py
marc-hb, line 1: `calc_elf_addresses` sounds very generic and focused more on the input of the script than on its output. Can you find a more specific name, maybe something like `llext_elf_linker_params.py`?

scripts/calc_elf_addresses.py
```diff
@@ -0,0 +1,43 @@
...
+						help='.text section address')
+
+	args = parser.parse_args()
+
+def main():
+	parse_args()
+
+	elf = ELFFile(open(args.file, 'rb'))
+
+	text_addr = int(args.text_addr.replace("0x", ""), 16)
```
marc-hb, line 26: Don't restrict the input to base 16, use auto-base 0:

```suggestion
	text_addr = int(args.text_addr, 0)
```

```python
>>> int("0x10",0)
16
>>> int("10",0)
10
>>> 
```

marc-hb (dismissed):

marc-hb:
```
west build -d build-mtl -b intel_adsp_ace15_mtpm modules/audio/sof/app/ -- -DOVERLAY_CONFIG="debug_overlay.conf"
west sign -d build-mtl -t rimage -p build-rimage/rimage --tool-data modules/audio/sof/rimage/config -- -k $key_path/key.pem
west sign -d build-mtl -t rimage -p build-rimage/rimage -- -k $key_path/key.pem -c modules/audio/sof/src/samples/audio/smart_amp_llext/smart_amp_test.toml -r build-mtl/zephyr/smart_amp_llext/libsmart_amp_test_out.so
```

The first, `west build` command already invokes `west sign` to sign `zephyr.elf` so something is wrong if you need to invoke `west sign` again in the second command.

> to run just copy zephyr.ri and llext.ri to the usual locations

So we have `hello_world.llext` in Zephyr, `lib$module_llext.elf` here and now `llext.ri`... the names and suffixes seem all over the place, zero consistency. Why not a simple `smart_amp.llext` -> `smart_amp.ri`?

lyakh:
> The first, `west build` command already invokes `west sign` to sign `zephyr.elf` so something is wrong if you need to invoke `west sign` again in the second command.

That's just because I don't have `rimage` installed system-wide

src/samples/audio/smart_amp_llext/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
...
+	-imacros${PROJECT_BINARY_DIR}/../include/generated/autoconf.h
+	-save-temps -O2
+)
+
+set(MODULE_LINKER_PARAMS -nostdlib -nodefaultlibs)
+target_link_options(${MODULE} PRIVATE
+	${MODULE_LINKER_PARAMS}
+)
+
+add_custom_target(${MODULE}_llext ${CMAKE_C_COMPILER}
```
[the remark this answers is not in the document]
lyakh: > Every `add_custom_target()` needs a corresponding `add_custom_command() and vice-versa, this is tedious, painful and error-prone but this CMake for you: https://samthursfield.wordpress.com/2015/11/21/cmake-dependencies-between-targets-and-files-and-custom-commands/
> 
> ``

@marc-hb sorry, not getting it why "every `add_custom_target()` needs a corresponding `add_custom_command()`?" `add_custom_target()` can include commands to build the target already, AFAIU `add_custom_command()` is only needed if one needs to *add more* commands (to that target)?

src/samples/audio/smart_amp_llext/CMakeLists.txt
lyakh: it doesn't need to, I just kept `.so` because that's what this actually is - it's a shared library with just some section addresses set explicitly and some sections stripped.

src/samples/audio/smart_amp_llext/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
...
+	${MODULE_LINKER_PARAMS}
+	"$$(${SOF_BASE}scripts/calc_elf_addresses.py" -f lib${MODULE}.so -t "A06CA000)"
```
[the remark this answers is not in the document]
lyakh: > and then read the file here

@marc-hb how do I portably read a file in cmake *at build time*? E.g. into a variable or an environment variable

src/samples/audio/smart_amp_llext/CMakeLists.txt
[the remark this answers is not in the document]
marc-hb: The best long story is https://samthursfield.wordpress.com/2015/11/21/cmake-dependencies-between-targets-and-files-and-custom-commands/

The short story is: because CMake sucks.

src/samples/audio/smart_amp_llext/CMakeLists.txt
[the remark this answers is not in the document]
marc-hb: https://cmake.org/cmake/help/latest/command/file.html#read

Also check the next one: `file(STRINGS ...)`

src/samples/audio/smart_amp_llext/CMakeLists.txt
marc-hb: OK but that's not a native library. I think `smart_amp.llext` provides more information while being simpler and more consistent.

src/samples/audio/smart_amp_llext/CMakeLists.txt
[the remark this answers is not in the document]
tejlmand: > @marc-hb how do I portably read a file in cmake at build time? E.g. into a variable or an environment variable

why do you need to read it into a variable or environment variable ?
If the file exists in the file system, then pass the information about the file to the script / command which needs it.

scripts/calc_elf_addresses.py
```diff
@@ -0,0 +1,43 @@
...
+
+	upper = rodata_offset - text_offset + int(args.text_addr, 16) + 0xfff
+	rodata_addr = upper - (upper % 0x1000)
+
+	upper = data_offset - rodata_offset + rodata_addr + 0xf
+	data_addr = upper - (upper % 0x10)
+
+	print(f'-Wl,-Ttext=0x{text_addr:x} '
+              f'-Wl,--section-start=.rodata=0x{rodata_addr:x} '
+              f'-Wl,-Tdata=0x{data_addr:x}')
```
tejlmand, line 40: going through a print statement like this is something that I believe is error prone and hard to debug.
What if warnings or other info is printed during python invocation ?

src/samples/audio/smart_amp_llext/CMakeLists.txt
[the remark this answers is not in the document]
tejlmand: > @teburd, @tejlmand is every llext module expected to implement this usual but ugly add_custom_ dance by itself?

initial look tells me it should be possible to create a common function which defines the needed target, but would need to know a bit more on the exact use-case for this to be able to provide some guidelines on doing so, if there is a need.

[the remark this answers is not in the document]
tejlmand: > Every add_custom_target() needs a corresponding `add_custom_command() and vice-versa.

Not true, example from Zephyr, `menuconfig` is a custom target not needed anything else.
But when multiple other targets depends on output created by a single custom command, then such command must be driven by a custom target. Something which often true in complex build system, like Zephyr.

tejlmand (commented):
couple of drive-by comments.

marc-hb (commented):
Thanks a lot @tejlmand for your time, much appreciated.
> initial look tells me it should be possible to create a common function which defines the needed target, but would need to know a bit more on the exact use-case for this to be able to provide some guidelines on doing so, if there is a need.

So the long story short is: Zephyr has this brand new and wonderful `llext` run-time library thing but no corresponding CMake, build or integration story. Yet? The only documentation I found is at https://docs.zephyrproject.org/latest/samples/subsys/llext/shell_loader/README.html and it tells the reader to compile `helloworld.llext` and convert it to hex by directly invoking the compiler themselves!

More confusing: there is a `zephyr/tests/subsys/llext/CMakeLists.txt` sample file (from @teburd) but it is a standalone Zephyr project! Wild guess: does this mean llext users should use Zephyr's sysbuild multi-build?

In other words the sign.py PR https://github.com/zephyrproject-rtos/zephyr/pull/66573 (which is how I landed here) seems to be relying on an llext build system  that... does not exist yet - not even as a clear and well-defined example that can be replicated.

src/samples/audio/smart_amp_llext/CMakeLists.txt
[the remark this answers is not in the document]
lyakh: > The best long story is https://samthursfield.wordpress.com/2015/11/21/cmake-dependencies-between-targets-and-files-and-custom-commands/
> 
> The short story is: because CMake sucks.

@marc-hb sorry, the short story is too short and the long one is too long. I looked through the whole article (sorry again, didn't take the time to read it completely), and didn't find the explanation to why this is needed. Could you point out to a specific section in it, please?

src/samples/audio/smart_amp_llext/CMakeLists.txt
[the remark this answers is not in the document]
lyakh: > https://cmake.org/cmake/help/latest/command/file.html#read
> 
> Also check the next one: `file(STRINGS ...)`

@marc-hb maybe what I asked wasn't put correctly, but by **build time** I meant - as opposed to, say, configuration time. AFAICS `file()` reads a file in at configuration time, whereas I need to read it after generating it at build time.

src/samples/audio/smart_amp_llext/CMakeLists.txt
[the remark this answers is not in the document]
lgirdwood: I am fine with a custom rule today. The important thing is that we have modules and can re-visit cmake improvements when they become priority.

src/samples/audio/smart_amp_llext/CMakeLists.txt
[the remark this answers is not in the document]
lgirdwood: So Windows build support for this feature is a secondary consideration, yes its needed, but not on day 1. We need to dog food this and re-visit improvements as needed.

lgirdwood (commented):
Nothing major blocking this from me.
@lyakh pls rebase since IIUC a lot of the patches already upstream.

teburd:
The cmake I wrote for the llext hello world sample is a working proof of concept not a final solution, as most things in software there's plenty of room for improvement. CMake doesn't have a target function for partially linked/custom elfs like I have for the arm build in this case and so I had to setup a custom command. The intent should be to do as @tejlmand suggested and eventually create a zephyr friendly llext target function much like the typical library/executable target functions in cmake work.

I have some on going work that might tie back into this actually as I need the compiler flags and include paths for something else.

src/samples/audio/smart_amp_llext/CMakeLists.txt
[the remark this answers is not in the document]
marc-hb: > Could you point out to a specific section in it, please?

Not really sorry. If this could be made shorter then:
- it wouldn't have taken me years of CMake practice to finally realize why you always must use both `add_custom_command()` + `add_custom_target()` at the same time
- I wouldn't keep forgetting the detailed reasons why and I wouldn't need a bookmark to this 1700 words long article.

Here's 2 sentences from the introduction:

> If you’ve tried do anything non-trivial in CMake using the add_custom_command() command, you may have got stuck in this horrible swamp of confusion [...] What makes it so hard is that there’s not one limitation, but several. 


The good news is you don't actually need to read and understand all the issues in this article, instead you can just trust me and this article and just use both `add_custom_command()` + `add_custom_target()`. That's exactly what I do myself; as I just wrote I don't remember precisely all the bad things that happen when you use only one. We're getting close to the point where doing it would take less time than discussing it.

The very slightly less short story is: **dependencies are not expressed correctly** if you don't use both which means things are not always built when you expect them to be (below the surface, CMake is a way to express dependencies like any other build system). So either you run out of date code or you're faced with a super cryptic error message, buried by the build concurrency. Either wastes an incredible amount of time.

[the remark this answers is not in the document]
lyakh: > The very slightly less short story is: **dependencies are not expressed correctly** if you don't use both which means things are not always built when you expect them to be

@marc-hb aha, ok, that's already something, not an explanation but at least a hint at what might happen otherwise... Since this is supposedly a bug in cmake (i.e. it isn't a documented feature, it's an empirically derived work-around), this might also behave differently with different versions of cmake, and hopefully some newer ones might even actually fix it, or it might not trigger in this specific case at all, or it might be something else, but since it doesn't cost much, we can add it here.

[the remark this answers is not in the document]
marc-hb: > Since this is supposedly a bug in cmake (i.e. it isn't a documented feature, it's an empirically derived work-around), this might also behave differently with different versions of cmake,

No, this is a _design_ issue with respect to custom rules. Different CMake versions should behave mostly the same. What does not behave the same is... users: different users have this weird tendency to modify different files! To which a half-broken graph of dependencies will react differently. Then, users will ALL be consistently confused no matter what happened and they will reach the "CMake hotline" :-)

lyakh:
This now needs 2 Zephyr PRs to work:
* https://github.com/zephyrproject-rtos/zephyr/pull/67105
* https://github.com/zephyrproject-rtos/zephyr/pull/66573

With those and if additionally smart-amp-test is set to "m" the following commands
```
west build -d mtl-build -b intel_adsp_ace15_mtpm modules/audio/sof/app/
west sign -d mtl-buikd -t rimage -- -k key.pem
west sign -d mtl-build -t rimage -D $PWD/modules/audio/sof/src/samples/audio/smart_amp_llext -- -k key.pem -r mtl-build/zephyr/smart_amp_llext/libsmart_amp_test_out.so
```
SOF and a smart-amp-test module can be built and signed

lyakh:
This should now be portable - build under Windows too.

lyakh:
Removing "DNM" since the `CONFIG_MODULES` option has been merged into Zephyr and this PR no longer depends on https://github.com/zephyrproject-rtos/zephyr/pull/67105 but CI failures still have to be fixed

scripts/llext_link_helper.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: BSD-3-Clause
+
+# We need to calculate ELF section addresses of an LLEXT module and use them to
+# run the linker. We could just use Python to calculate addresses and pass them
+# back to cmake to have it call the linker. However, there doesn't seem to be a
+# portable way to do that. Therefore we pass the linker path and all the command
```
marc-hb, line 7: @tejlmand , any idea?

I think we want this to be done in Zephyr for any `llext` eventually, not just SOF. @lyakh this problem is not SOF specific or is it? Either way state it in this comment so we know where the solution for this should live eventually.

marc-hb (commented):
A lot of kernel panics in MTL https://sof-ci.01.org/sofpr/PR8180/build1803/devicetest/index.html ...

scripts/llext_link_helper.py
lyakh: @marc-hb it is SOF-specific. It isn't needed for generic llext loading on Xtensa under Zephyr. It's only needed to match SOF's loadable module / manifest design

lyakh:
CI failure analysis:
1.  verify-kernel-boot-log.sh  failures on all 3 ACE platforms in https://sof-ci.01.org/sofpr/PR8180/build1820/devicetest/index.html is the same as https://github.com/thesofproject/sof/pull/8732#issuecomment-1888936506
2. https://sof-ci.01.org/sofpr/PR8180/build1820/devicetest/index.html?model=MTLP_RVP_HDA&testcase=check-capture-3times is with high certainty the same problem: it's a kernel error in the *very beginning* of the test at timestamp 289.26:
```
[  289.261770] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc timed out for 0x47000000|0x0
[  289.261829] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: Attempting to prevent DSP from entering D3 state to preserve context
[  289.261837] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ IPC dump start ]------------
[  289.261918] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: Host IPC initiator: 0xc7000000|0x0|0x0, target: 0x1b0a0000|0x0|0x0, ctl: 0x3
[  289.261962] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ IPC dump end ]------------
[  289.261995] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: IPC timeout
[  289.262024] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ctx_save IPC error: -110, proceeding with suspend
```
which is 2.5 seconds after the end of the previous test at 286.63:
```
[  286.630950] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx      : 0x12010000|0x0: GLB_DELETE_PIPELINE
[  286.631411] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx reply: 0x32000000|0x0: GLB_DELETE_PIPELINE
[  286.631484] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx done : 0x12010000|0x0: GLB_DELETE_PIPELINE
[  286.631503] kernel: snd_sof:sof_widget_free_unlocked: sof-audio-pci-intel-mtl 0000:00:1f.3: widget pipeline.2 freed
[  286.631511] kernel: snd_sof:sof_widget_free_unlocked: sof-audio-pci-intel-mtl 0000:00:1f.3: widget dai-copier.HDA.Analog.playback freed
[  286.632079] kernel: snd_sof:sof_pcm_close: sof-audio-pci-intel-mtl 0000:00:1f.3: pcm: close stream 31 dir 0
```
so it's likely that between these two timestamps the DSP was sent into a suspend and hit the same `panic()` as above
3. https://sof-ci.01.org/sofpr/PR8180/build1820/devicetest/index.html?model=MTLP_RVP_HDA&testcase=check-playback-3rounds is exactly the same kind of failure between seconds 317 and 320.
4.  check-runtime-pm-status.sh again hits the same bug and prints DSP panic dumps verbatim
5. https://sof-ci.01.org/sofpr/PR8180/build1820/devicetest/index.html?model=MTLP_RVP_NOCODEC&testcase=multiple-pause-resume-50 is a still not fully fixed pause-resume failure
6. all module loading-unloading, suspend-resume and sof-logger tests hit the same bug too

marc-hb (commented):
Brand new 67431 (still in review) is finally adding  new, much desired, generic CMake code to help with generating all `.llext` files. Please try to leverage as much of that as possible as opposed to duplicating it from scratch (this is not about who was first.) We can't afford to be "creative" and re-invent Zephyr code. This is even more important for CMake code than for C code because we have approximately 0.2 CMake maintainer in total in the SOF project, even less skill and exactly zero people interested in working on CMake. De-duplicating or more generally deleting redundant code is especially tedious and time-consuming because it's extremely hard to preserve "test coverage" and make sure nothing is getting lost or damaged.

lyakh:
> Brand new 67431 (still in review) is finally adding new, much desired, generic CMake code to help with generating all `.llext` files. Please try to leverage as much of that as possible as opposed to duplicating it from scratch (this is not about who was first.) We can't afford to be "creative" and re-invent Zephyr code. This is even more important for CMake code than for C code because we have approximately 0.2 CMake maintainer in total in the SOF project, even less skill and exactly zero people interested in working on CMake. De-duplicating or more generally deleting redundant code is especially tedious and time-consuming because it's extremely hard to preserve "test coverage" and make sure nothing is getting lost or damaged.

@marc-hb Thanks for the pointer, once that PR is merged, we shall move to use its cmake infrastructure, but I don't think we should block on waiting for it to be merged. If this is the only remaining reason for your request for change, I'd kindly request you to reconsider it. I can change the name to .llext of course, that's very minor because we're anyway interested in the final signed version, which will have a suffix of .ri

lgirdwood:
> > Brand new 67431 (still in review) is finally adding new, much desired, generic CMake code to help with generating all `.llext` files. Please try to leverage as much of that as possible as opposed to duplicating it from scratch (this is not about who was first.) We can't afford to be "creative" and re-invent Zephyr code. This is even more important for CMake code than for C code because we have approximately 0.2 CMake maintainer in total in the SOF project, even less skill and exactly zero people interested in working on CMake. De-duplicating or more generally deleting redundant code is especially tedious and time-consuming because it's extremely hard to preserve "test coverage" and make sure nothing is getting lost or damaged.
> 
> @marc-hb Thanks for the pointer, once that PR is merged, we shall move to use its cmake infrastructure, but I don't think we should block on waiting for it to be merged. If this is the only remaining reason for your request for change, I'd kindly request you to reconsider it. I can change the name to .llext of course, that's very minor because we're anyway interested in the final signed version, which will have a suffix of .ri

Ack - I'm not going to wait on a Zephyr PR that is WIP. We can come back and align with the Zephyr PR once its stabilized and aligned upstream in Zephyr.

lyakh:
All the CI TGL failures in https://sof-ci.01.org/sofpr/PR8180/build1908/devicetest/index.html are the same as in #8747 which means, that we cannot move to that Zephyr commit before we fix that regression

src/library_manager/lib_manager.c
lgirdwood: @lyakh this now resolved  ?

lgirdwood (commented):
I'm good to go providing the IADK comment from @jxstelter is resolved alongside any CI issues are resolved. We can come back and align with Zephyr cmake once its merged.

src/library_manager/lib_manager.c
lyakh: @jxstelter fixed by keeping the system-service version unchanged

lyakh:
> I'm good to go providing the IADK comment from @jxstelter is resolved alongside any CI issues are resolved. We can come back and align with Zephyr cmake once its merged.

@lgirdwood yes: https://github.com/thesofproject/sof/pull/8180#discussion_r1455619438

marc-hb:
The review of the llext+CMake work in Zephyr is going extremely well, multiple approvals including from the maintainer:
- https://github.com/zephyrproject-rtos/zephyr/pull/67431

@lyakh please download it and give it at least a try before merging this.

At the very least we don't want this PR #8180 to immediately introduce a (logical) conflict with this Zephyr PR about to be merged.

lyakh:
> The review of the llext+CMake work in Zephyr is going extremely well, multiple approvals including from the maintainer:
> 
>     * [cmake: add llext compilation module zephyrproject-rtos/zephyr#67431](https://github.com/zephyrproject-rtos/zephyr/pull/67431)
> 
> 
> @lyakh please download it and give it at least a try before merging this.
> 
> At the very least we don't want this PR #8180 to immediately introduce a (logical) conflict with this Zephyr PR 67431 about to be merged.

@marc-hb that PR is now merged, but it doesn't help us here a whole lot. It doesn't implement address calculations for us. So at best we can use the new `add_llext_target()` for the first stage - compiling sources and generating an intermediate binary, but then we anyway need to add a custom target to rewrite addresses. So I still think this can be done as a second step after this PR is merged

marc-hb:
> but then we anyway need to add a custom target to rewrite addresses. 

Thanks for looking at it! Any idea how we can solve this in the longer term? (= never?). I remember you said it's SOF-specific but SOF is one of the very first llext users so it's still a pity. @pillo79 any idea, even vague?

One thing this PR seems to miss compared to the way Zephyr does it is CFLAGS management...

pillo79:
I discussed this briefly on Discord with @ly

pillo79:
@marc-hb I talked about this briefly yesterday with @lyakh - I am preparing a tiny followup to that PR with a custom post processing step. That should help!

app/prj.conf
```diff
@@ -56,3 +56,7 @@ CONFIG_SCHED_CPU_MASK_PIN_ONLY=y
 CONFIG_SYS_CLOCK_TICKS_PER_SEC=15000
 CONFIG_DAI=y
 CONFIG_HEAP_MEM_POOL_SIZE=2048
+
+CONFIG_LLEXT=y
+CONFIG_LLEXT_STORAGE_WRITABLE=y
+CONFIG_MODULES=y
```
pjdobrowolski, line 62: If we have 3 different approaches to loading modules, I wouldn't use such generic name. It cause misunderstanding.

scripts/llext_link_helper.py
pjdobrowolski: Why can't we do it in linker scripts?

pjdobrowolski (dismissed):

app/prj.conf
lyakh: @pjdobrowolski it isn't a name that we choose, it's hard-coded in kconfig tools. We want to use `tristate` in module Kconfig and set it to "m" to build as a module. And for that you need `CONFIG_MODULES`. And I'd propose to use `tristate` for any modules - whichever implementation they use, but that's up to respective authors / maintainers.

scripts/llext_link_helper.py
lyakh: > Why can't we do it in linker scripts?

@pjdobrowolski why should we? If you use a linker script you need to define everything, a complete module linking layout. I see no need for that. The only thing that we need is a fixed memory location where we want to map respective modules, that's the only parameter I'm using. The rest is calculated automatically. In fact ideally even the load address shouldn't be needed. Think about OS executables, libraries, Linux kernel modules, etc. The OS automatically finds a location to copy the code and the data to and automatically selects addresses to map it to.

pjdobrowolski: Because it keeps all memory locations and operations in one place. I understand that it is easier to do it but FW is a bit more complex that software and adding python to it doesn't help.

app/prj.conf
pjdobrowolski: >  modules:
>        The Symbol instance for the modules symbol. Currently hardcoded to
>        MODULES, which is backwards compatible. Kconfiglib will warn if
>        'option modules' is set on some other symbol. Tell me if you need proper
>        'option modules' support.
> 
Can't we add another specific symbol for LLEXT?

scripts/llext_link_helper.py
marc-hb: > FW is a bit more complex that software  and adding python to it doesn't help.

This is quite vague and cryptic, can you elaborate? Python has great support for ELF. It is portable, popular, interactive and high-level. So it's hard to find another tool that checks so many boxes for such a task. It's also consistent with Zephyr which uses all over the place for similar jobs.

Maybe your comment wasn't about the choice of the Python language specifically but more generally about computing addresses versus hardcoding them?

> Because it keeps all memory locations and operations in one place

This is a bit confusing too: addresses dynamically computed are not kept in a different place... would it help to make sure they get printed and saved?

marc-hb: BTW there is on-going work to add a post-processing step in CMake:
- https://github.com/zephyrproject-rtos/zephyr/pull/67997

marc-hb: > FW is a bit more complex that software...

Depends how you look at it
https://en.wikipedia.org/wiki/Address_space_layout_randomization

app/prj.conf
marc-hb: > Tell me if you need proper 'option modules' support.

Please source quotations. This one comes from  here:
https://github.com/ulfalizer/Kconfiglib/blame/061e71f7d78cb0/kconfiglib.py#L671

This comment is 7 years old. Zephyr modules are months old and still in the making. I guess this text should be updated.



The reason @teburd and @lyakh could implement Zephyr modules without making any `kconfiglib.py` change is because `kconfiglib.py` has stuck to bug-for-bug compatibility with the original Kconfig implementation in C. https://www.kernel.org/doc/html/latest/kbuild/kconfig-language.html
I suspect the `kconfiglib.py` test suite covered modules even when Zephyr didn't support them yet.

Diverging `kconfiglib.py` away from the original Linux kernel implementation could break the `kconfiglib.py` test suite and make them slightly incompatible with each other, require slightly different documentations, disrupt other Zephyr users besides SOF,...A potentially huge amount of hassle for a small naming gain.

BTW `kconfiglib.py` is also used outside Zephyr.

I agree calling one of SOF's "3 different approaches" with a generic "CONFIG_MODULES" sucks, but Zephyr seems to have only one approach so it has no reason to disambiguate. This is hopefully just an implementation detail that most users will not pay much attention to: Kconfig has a user interface that shows the `help` text.

Note we have "_HOST_" abbreviated to "_HST_" and register sets called `DfPMCCH`. This is not an excuse not to care about names but it put things in a bit of perspective :-)



In any case this sounds like a Zephyr topic, not an SOF topic?

lyakh: Thanks for the context, @marc-hb 

> > Tell me if you need proper 'option modules' support.
> 
> Please source quotations. This one comes from here: https://github.com/ulfalizer/Kconfiglib/blame/061e71f7d78cb0/kconfiglib.py#L671
> 
> This comment is 7 years old. Zephyr modules are months old and still in the making. I guess this text should be updated.

Just to clarify a bit - it isn't just about the text (documentation), it's in the kconfiglib implementation. And yes, your text below actually alludes to that too.

[snip]

> I agree calling one of SOF's "3 different approaches" with a generic "CONFIG_MODULES" sucks, but Zephyr seems to have only one approach so it has no reason to disambiguate. This is hopefully just an implementation detail that most users will not pay much attention to: Kconfig has a user interface that shows the `help` text.

This isn't an option for **one** of the 3 different approaches. Other approaches are welcome and encouraged to use it too! IMHO it's a simple and natural way to switch a module between a built-in and modular build, so, any Kconfig options, that want this functionality, are free to use this! E.g. if we modularise our Maths libraries. As for which of the "three approaches" is used - I'd propose to use other means to make those decisions, if we ever have modules, that can be built with more than one option. E.g. indeed with this PR the smart-amp-test module would be buildable in three ways - built-in, as a module using system services or as an LLEXT. Currently The "y" Kconfig option selects whether it's built-in, selecting "m" build it as an LLEXT **immediately** during the base-firmware build, and to build it as a system-services module you run a separate cmake command, which ignores this Kconfig option completely, so you can build such a module regardless of this Kconfig value.

> Note we have "_HOST_" abbreviated to "_HST_" and register sets called `DfPMCCH`. This is not an excuse not to care about names but it put things in a bit of perspective :-)
> 
> In any case this sounds like a Zephyr topic, not an SOF topic?

marc-hb: Thanks @lyakh . I didn't expect system-services modules to ignore the Kconfig of the base firmware (sounds quite risky) but if they do then this entire discussion seems indeed irrelevant.

lyakh:
@pjdobrowolski have we answered all your concerns? If so please consider removing your request for change

lyakh:
Now that we upgraded to a Zephyr version with `CONFIG_MODULES` support, this is mergeable (again). CI failures: https://sof-ci.01.org/sofpr/PR8180/build3168/devicetest/index.html?model=MTLP_RVP_NOCODEC&testcase=multiple-pause-resume-50 I think is known and unrelated (@kv2019i @marc-hb am I right?), https://sof-ci.01.org/sofpr/PR8180/build3169/devicetest/index.html is suspend-resume failures, style - one warning is a moved code line, another one is `__used` vs. `__attribute__((used))` - this one could be addressed I think, Zephyr does have a `__used` macro, SOF should be able to use it too.

src/samples/audio/smart_amp_llext/CMakeLists.txt
lyakh: @kv2019i @marc-hb so you want .llext as the output before signing? And after signing .ri again? With or without "llext?"

lyakh: smart_amp.llext before signing and smart_amp.llext.ri after signing?

lgirdwood (approved):

PR #8182, opened 2023-09-11
intel_adsp: ace: cpu clock management

tmleman:
Updated clock definitions for ACE family platforms.

- [x] https://github.com/zephyrproject-rtos/zephyr/pull/62452


app/overlays/mtl/fpga_overlay.conf
```diff
@@ -1,2 +1,3 @@
 CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC=19200000
 CONFIG_DAI_DMIC_HW_IOCLK=19200000
+CONFIG_XTENSA_CCOUNT_HZ=24000000
```
plbossart, line 3: is this correct in all cases? I clearly remember an FPGA image running faster for DMIC tests.

src/platform/lunarlake/include/platform/lib/clk.h
```diff
@@ -14,7 +14,7 @@
 
 #include <ace/lib/clk.h>
 
-#define CLK_MAX_CPU_HZ		400000000
+#define CLK_MAX_CPU_HZ		CONFIG_XTENSA_CCOUNT_HZ
```
plbossart, line 17: Are you sure about the 393.2 MHz value?

It should be an integer multiple of 24.576 MHz and it's off a bit.

should be 393.216 MHz

plbossart (commented):
I can only agree with this change, but the comments and values seem off?

src/platform/lunarlake/include/platform/lib/clk.h
tmleman: Yes, you are right. The exact value is `393216000` and this is given as the default for `CONFIG_XTENSA_CCOUNT_HZ`.
`393.2 MHz` is taken from documentation.

app/overlays/mtl/fpga_overlay.conf
tmleman: My mistake, I copied the wrong value, it should be 40 MHz.

src/platform/lunarlake/include/platform/lib/clk.h
```diff
@@ -14,21 +14,19 @@
...
 
-#define CPU_LPRO_FREQ_IDX	1
-
-#define CPU_HPRO_FREQ_IDX	2
+#define CPU_IPLL_FREQ_IDX	1
 
 #define CPU_LOWEST_FREQ_IDX	CPU_WOVCRO_FREQ_IDX
 
-#define CPU_DEFAULT_IDX		CPU_HPRO_FREQ_IDX
+#define CPU_DEFAULT_IDX		CPU_IPLL_FREQ_IDX
```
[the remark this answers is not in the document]
tmleman: Yes

src/audio/base_fw.c
```diff
@@ -51,7 +51,7 @@ static int basefw_config(uint32_t *data_offset, char *data)
 	tuple = tlv_next(tuple);
 	tlv_value_uint32_set(tuple,
 			     IPC4_SLOW_CLOCK_FREQ_HZ_FW_CFG,
-			     clock_get_freq(CPU_LPRO_FREQ_IDX));
+			     clock_get_freq(CPU_LOWEST_FREQ_IDX));
```
kv2019i, line 54: Hmm, will this break bisect with the previous git commit? If yes, please merge this change to the Zephyr baseline update. 

Note: we should really avoid these kind of changes and provide transition period with Zephyr (i.e. add new defines in Zephyr, merge to SOF, then clean up old from Zephyr once done). E.g. now this will block @dbaluta 's PR to update Zephyr as there a interface change in Intel code.
If there's no dependency, then this please copy this note paragraphi to /dev/null.

kv2019i (commented):
Looks good except for two things: 1) we need a clean CI pass, and 2) please confirm we don't break git bisect.

src/audio/base_fw.c
tmleman: I have corrected the order of commits to make bisection possible.

You're right about changes in zephyr. In this case it was possible to introduce them there that way so the kernel could be updated without any changes in SOF. I will remember about this next time.

jxstelter (approved):

kv2019i:
Right, so now the remaining hurdle is this new build error on Intel MTL:
```
In file included from /zep_workspace/zephyr/include/zephyr/sys/mem_manage.h:104,
                 from /zep_workspace/zephyr/include/zephyr/sys/device_mmio.h:47,
                 from /zep_workspace/zephyr/include/zephyr/device.h:15,
                 from /zep_workspace/zephyr/drivers/mm/mm_drv_intel_adsp.h:17,
                 from /zep_workspace/zephyr/drivers/mm/mm_drv_intel_adsp_mtl_tlb.c:23:
/zep_workspace/zephyr/drivers/mm/mm_drv_intel_adsp_mtl_tlb.c: In function 'sys_mm_drv_mm_init':
/zep_workspace/zephyr/include/zephyr/sys/__assert.h:44:52: error: format '%p' expects argument of type 'void *', but argument 2 has type 'long unsigned int' [-Werror=format=] 
```

Affects #8235 as well. So we'd need a fix for this to be merged at the same time.

dbaluta:
> Right, so now the remaining hurdle is this new build error on Intel MTL:
> Affects #8235 as well. So we'd need a fix for this to be merged at the same time.

@kv2019i  sent a patch here: https://github.com/zephyrproject-rtos/zephyr/pull/62854

dnikodem (approved):

btian1 (approved):

kv2019i:
Oh man, @tmleman @dbaluta so SOF CI build is ok and pr-device-test tests even seem to pass, but:
- we have one mandatory Intel CI failing
- all Windows builds fail (lost ninja, no idea what is causing this)
- the sparse build is broken after the picolibc change

Considering I could merge PRs on Monday with all-green in the PR results page, this is pretty disappointing, but it is what it is.

It seems there are now multiple fails not related to actual SOF functionality, so I'd be inclined to merge this once mandatory CI checks passs, and file P1 bugs to track the sparse and windows build failures.

kv2019i:
@marc-hb @lyakh Any insights on the sparse build fail? I couldn't make it work locally even without this PR, so I have no immediate insight how to go about to fix this...

marc-hb:
> all Windows builds fail (lost ninja, no idea what is causing this)

Fix submitted:
- #8250

marc-hb:
> I couldn't make it work locally even without this PR

First you need a special version of sparse with @lyakh's fix. As always, check the build logs. In this case the "git clone sparse analyzer" step.

Then:

```shell
PATH=$PATH:$HOME/sparse

sof/scripts/xtensa-build-zephyr.py -p mtl --cmake-args=-DZEPHYR_SCA_VARIANT=sparse   --cmake-args=-DCONFIG_LOG_USE_VLA=n |& tee sparse.log

scripts/parse_sparse_output.sh < sparse.log

```

Same as what CI does, no CI secret.

marc-hb:
> the sparse build is broken after the picolibc change

Confirmed:
- https://github.com/zephyrproject-rtos/zephyr/issues/63003

`CONFIG_MINIMAL_LIBC=y` in an overlay should be enough to avoid it.

pblaszko (approved):

tmleman:
Maybe we should wait for that too https://github.com/zephyrproject-rtos/zephyr/pull/63004

kv2019i:
Nice, https://sof-ci.01.org/sofpr/PR8182/build13439/devicetest/index.html 

I guess https://github.com/zephyrproject-rtos/zephyr/pull/63004 was not tested on hardware...? It seems none of the resulting binaries boot anymore.

tmleman:
I'm closing this PR. Those changes and zephyr version  update are made here https://github.com/thesofproject/sof/pull/8268.

PR #8183, opened 2023-09-11
audio: volume: optimize windows volume ramp

btian1:
windows volume ramp peak optimization, currently, with these patches, nearly 20% are reduced, measured on TGL with 48k 2 channel stream. if there is 4 channel or more, the reduction will be higher due to no matter channels you have, as long as target volume are same, ramp calculation only take once.

### DIFF
--- a/src/audio/volume/volume.c
+++ b/src/audio/volume/volume.c
@@ -243,6 +243,24 @@ static inline int32_t volume_windows_fade_ramp(struct vol_data *cd, int32_t ramp
 }
 #endif
 
+void volume_set_ramp_channel_counter(struct vol_data *cd, uint32_t channels_count)
+{
+	int i;
+	bool is_same_volume = true;
+
+	for (i = 1; i < channels_count; i++) {
+		if (cd->tvolume[0] != cd->tvolume[i]) {
+			is_same_volume = false;
+			break;
+		}
+	}
+
+	if (is_same_volume)
+		cd->ramp_channel_counter = 1;
+	else
+		cd->ramp_channel_counter = channels_count;
+}
+
 /**
  * \brief Ramps volume changes over time.
  * \param[in,out] vol_data Volume component data
@@ -272,7 +290,7 @@ static inline void volume_ramp(struct processing_module *mod)
 #endif
 
 	/* Update each volume if it's not at target for active channels */
-	for (i = 0; i < cd->channels; i++) {
+	for (i = 0; i < cd->ramp_channel_counter; i++) {
 		/* skip if target reached */
 		volume = cd->volume[i];
 		tvolume = cd->tvolume[i];
@@ -314,6 +332,9 @@ static inline void volume_ramp(struct processing_module *mod)
 		}
 		cd->volume[i] = new_vol;
 	}
+
+	for (i = cd->ramp_channel_counter; i < cd->channels; i++)
+		cd->volume[i] = cd->volume[0];
 
 	cd->is_passthrough = cd->ramp_finished;
 	for (i = 0; i < cd->channels; i++) {
--- a/src/audio/volume/volume.c
+++ b/src/audio/volume/volume.c
@@ -332,7 +332,7 @@ static inline void volume_ramp(struct processing_module *mod)
 		}
 		cd->volume[i] = new_vol;
 	}
-
+	/* assign other channel volume as the first calculated volume with same volume case */
 	for (i = cd->ramp_channel_counter; i < cd->channels; i++)
 		cd->volume[i] = cd->volume[0];
 
@@ -557,8 +557,9 @@ static int volume_process(struct processing_module *mod,
 	comp_dbg(mod->dev, "volume_process()");
 
 	while (avail_frames) {
+#if CONFIG_COMP_PEAK_VOL
 		volume_update_current_vol_ipc4(cd);
-
+#endif
 		if (cd->ramp_finished || cd->vol_ramp_frames > avail_frames) {
 			/* without ramping process all at once */
 			frames = avail_frames;
--- a/src/audio/volume/volume_ipc3.c
+++ b/src/audio/volume/volume_ipc3.c
@@ -145,6 +145,9 @@ int volume_init(struct processing_module *mod)
 		cd->muted[i] = false;
 	}
 
+	/* all target volume are same */
+	cd->ramp_channel_counter = 1;
+
 	switch (cd->ramp_type) {
 #if CONFIG_COMP_VOLUME_LINEAR_RAMP
 	case SOF_VOLUME_LINEAR:
@@ -217,6 +220,7 @@ int volume_set_config(struct processing_module *mod, uint32_t config_id,
 					return ret;
 			}
 		}
+		volume_set_ramp_channel_counter(cd, cd->channels);
 
 		volume_ramp_check(mod);
 		break;
--- a/src/audio/volume/volume_ipc4.c
+++ b/src/audio/volume/volume_ipc4.c
@@ -159,7 +159,7 @@ int volume_init(struct processing_module *mod)
 
 	md->private = cd;
 
-	for (channel = 0; channel < channels_count ; channel++) {
+	for (channel = 0; channel < channels_count; channel++) {
 		if (vol->config[0].channel_id == IPC4_ALL_CHANNELS_MASK)
 			channel_cfg = 0;
 		else
--- a/src/audio/volume/volume_ipc4.c
+++ b/src/audio/volume/volume_ipc4.c
@@ -176,6 +176,8 @@ int volume_init(struct processing_module *mod)
 
 	init_ramp(cd, vol->config[0].curve_duration, target_volume[0]);
 
+	volume_set_ramp_channel_counter(cd, channels_count);
+
 	cd->mailbox_offset = offsetof(struct ipc4_fw_registers, peak_vol_regs);
 	cd->mailbox_offset += instance_id * sizeof(struct ipc4_peak_volume_regs);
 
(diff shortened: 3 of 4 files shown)


singalsu (changes requested):
This is not acceptable. Need to have subjective and objective evidence to demonstrate that volume transitions still sound nice.

When discussed this today @lrgirdwo proposed a kconfig setting for volume fade-in and transitions quality. The worst-quality setting could be something where majority of subjects are not noticing transitions roughness vs. ideal every sample ramped solution.

This is further complicated by the fact that ramp update rate need (as seen in this removed code) depends on ramp duration. We should test different durations in range e,g, 10 ms - 200 ms to see that is needed.

This also depends on ramp type. The thresholds and update rates in the code were created for linear ramp type.

btian1:
> This is not acceptable. Need to have subjective and objective evidence to demonstrate that volume transitions still sound nice.
> 
> When discussed this today @lrgirdwo proposed a kconfig setting for volume fade-in and transitions quality. The worst-quality setting could be something where majority of subjects are not noticing transitions roughness vs. ideal every sample ramped solution.
> 
> This is further complicated by the fact that ramp update rate need (as seen in this removed code) depends on ramp duration. We should test different durations in range e,g, 10 ms - 200 ms to see that is needed.
> 
> This also depends on ramp type. The thresholds and update rates in the code were created for linear ramp type.

Question is: we need find a way to validate it, and how 128/256/500/1000 was set when ramp algorithm was designed?

singalsu:
> Question is: we need find a way to validate it, and how 128/256/500/1000 was set when ramp algorithm was designed?

Subjective test work always but arranging it is a lot of work. This is easy to simulate with Matlab. With some time allowed I could create a wav data set for people to listen and score.

src/audio/volume/volume.c
```diff
@@ -352,10 +352,6 @@ void volume_prepare_ramp(struct comp_dev *dev, struct vol_data *cd)
 	 * four presets are used.
 	 */
 	if (cd->initial_ramp < VOL_RAMP_UPDATE_THRESHOLD_FASTEST_MS)
-		ramp_update_us = VOL_RAMP_UPDATE_FASTEST_US;
```
lgirdwood, line 355: @singalsu would it be better here making ```ramp_update_us``` a Kconfig build time constant. he Kconfig help should explain the benefits and trade off for selecting the value.

marcinszkudlinski (changes requested):
1) Not sure about audible effect of this change - why do you think 128us is too fast?

2) if it is OK, better adjust VOL_RAMP_UPDATE_* macros, 

3) it you really want to fo the change, remove VOL_RAMP_UPDATE_FASTEST_US from .h as is now orphaned, never used definition

btian1:
> 

Not sure about audible effect of this change - why do you think 128us is too fast?
 6 samples need a new ramping volume.

if slower ramp is OK, better adjust VOL_RAMP_UPDATE_* macros to get proper times
this is also applicable, can change its range from 125-1000 to 300 - 1200, just an example.

it you really want to do the change this way, remove VOL_RAMP_UPDATE_FASTEST_US from .h as is now orphaned, never used definition
this is just draft, like throw away a brick in order to get a gem.

src/audio/volume/volume.c
singalsu: There could be choice for smallest update step 125 us / 250 us / 500 us / 1000 us, and the thresholds for selecting the larger steps could be replaced with an equation that is roughly like current if - else code. But the problem is that in my opinion anything but the fastest updated ramp sound quite annoying. So I'd not change the default 125 us.

Should I make such PR?

src/audio/volume/volume.c
```diff
@@ -311,6 +329,11 @@ static inline void volume_ramp(struct processing_module *mod)
 		cd->volume[i] = new_vol;
 	}
 
+	if (cd->is_same_volume) {
+		for (i = 1; i < cd->channels; i++)
```
lyakh, line 333: I'd just do `for (i = cd->ramp_channel_counter; i < cd->channels; i++)` and remove the `if` in line 332

keqiaozhang:
SOFCI TEST

lgirdwood:
@btian1 @singalsu I assume we are going for Kconfig based approach now as discussed yesterday call ? if so, do we need all patches from this or can we drop some ?

singalsu:
> @lgirdwood , @singalsu , I think these are all needed, it just did optimization on windows fade ramp with 20% reduction, it always take effect.
> 
> I don't like the config PR(#8227), it is adding complex to volume module. after linux/chrome use linear ramp, there is no peak mcps, the ratio < 2, for windows, it they don't want fade, just send NONE curve when set volume.
> 
> anyway, this patch does not have relation with PR:#8227.

#8227 has the problem that it fails CI that expects the playback fade-in to happen. I think I'll mark it as don't merge until we know better if it's acceptable. It's like cleaning up a room with a bulldozer.

src/audio/volume/volume.c
```diff
@@ -243,6 +243,24 @@ static inline int32_t volume_windows_fade_ramp(struct vol_data *cd, int32_t ramp
 }
 #endif
 
+void volume_set_ramp_channel(struct vol_data *cd, uint32_t channels_count)
```
singalsu, line 246: Maybe function name should be instead `volume_set_ramp_channel_counter() `to avoid confusion what it does.

src/audio/volume/volume_ipc3.c
```diff
@@ -149,6 +149,10 @@ int volume_init(struct processing_module *mod)
 		cd->muted[i] = false;
 	}
 
+	cd->is_same_volume = true;
+	/* all target volume are same */
+	cd->ramp_channel_counter = 1;
+
```
singalsu, line 155: I wonder if false and 0 would be more safe initialize?

src/audio/volume/volume_ipc4.c
```diff
@@ -156,6 +156,7 @@ int volume_init(struct processing_module *mod)
 	}
 
 	md->private = cd;
+	cd->is_same_volume = true;
```
singalsu, line 159: This initialize is different vs. proposed for ipc3. Also I wonder if false, 0 would be better default here.

src/audio/volume/volume.c
```diff
@@ -232,11 +232,11 @@ static inline int32_t volume_windows_fade_ramp(struct vol_data *cd, int32_t ramp
 {
 	int32_t time_ratio; /* Q2.30 */
```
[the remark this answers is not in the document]
btian1: can't image how fast this user case? since ramp is ms based,  if this really happened, current code can handle it well？

src/audio/volume/volume_ipc3.c
btian1: 0 may cause the first startup volume failure, set to 1 means for ipc3, all target volume are same, so only ramp one channel is ok.

lgirdwood (commented):
@marcinszkudlinski @singalsu does this update work for you guys ?

src/audio/volume/volume.c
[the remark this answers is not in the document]
singalsu: I'm testing it with topology that has e.g. 400 ms ramp and with script that applies amixer cset volume controls very rapidly same and different "random" muted and unmuted values out of my head for channels. Playback sine wave and record it and see that signal envelope follows commands.

singalsu:
> @marcinszkudlinski @singalsu does this update work for you guys ?

I haven't yet - lack of time but I can do later. I'd encourage all developers to have a way to test themselves changes when it's known that CI is limited. Be able to apply controls with script and playback and capture and analyze the result.

singalsu:
I changed curve duration default to 200 ms in gain.conf and played sine wave while abusing alsamixer volume control. Adjusting the volume down at cursor key repeat rate made glitches to first channel:

Screenshot from 2023-09-27 12-07-17

There are also glitches when adjusting volume continuously up but they are not as well visible.

With random 0 - 9 numbers keys presses I was able to get 1 ms long gain glitch;

Screenshot from 2023-09-27 12-13-44

btian1:
@singalsu @lgirdwood , I removed last patch, please check whether you still have comments on left 3 patches.

singalsu (approved):
Looks good to me, thanks!

kv2019i (approved):
Thanks, looks good @btian1 !

btian1:
> 1. Not sure about audible effect of this change - why do you think 128us is too fast?
> 2. if slower ramp is OK, better adjust VOL_RAMP_UPDATE_* macros to get proper times
> 3. it you really want to do the change this way, remove VOL_RAMP_UPDATE_FASTEST_US from .h as is now orphaned, never used definition

Please check again, I removed last patch, @marcinszkudlinski @lgirdwood

src/audio/volume/volume.c
```diff
@@ -315,6 +333,9 @@ static inline void volume_ramp(struct processing_module *mod)
 		cd->volume[i] = new_vol;
 	}
 
+	for (i = cd->ramp_channel_counter; i < cd->channels; i++)
+		cd->volume[i] = cd->volume[0];
```
andrula-song, line 337: can we add some explanatory note here? since it seems only for the volume of every channels is the same.

btian1: @andrula-song , new comments added.

lgirdwood:
@marcinszkudlinski good for you ?

singalsu:
Another confirm this PR version is OK. A script that applies rapid volume controls with amixer cset is passed without glitches. The low level spectral images during ramp are normal. The red spectral line with all frequeinces in the end is due to sudden stop of playback while this analog recording of line out continued.

Screenshot from 2023-10-18 10-49-07

PR #8170, opened 2023-09-06
[v2.7] Disable IMR context save for MTL [alternative comment wording]

kv2019i:
We really need this for v2.7 and rc1 cut-off is planned on Friday, so let me propose an alternative comment wording for
https://github.com/thesofproject/sof/pull/8156

We can merge either that PR or this, but we really need to get this forward (as we agree on the code and this is now about the comment).

PR #8136, opened 2023-08-31
[DNM] west.yml: test Zephyr PR 62180

kv2019i:
https://github.com/zephyrproject-rtos/zephyr/pull/62180

PR #8185, opened 2023-09-12
[Depends on 62522] zephyr: cavs: add secondary core context save support

RanderWang:
Register pm_state_notifier to set ready_flag for secondary core when it is powered up for second time after first fw boot. We can remove CONFIG_ADSP_IMR_CONTEXT_SAVE check for cavs platform for this feature.

zephyr side pr: https://github.com/zephyrproject-rtos/zephyr/pull/62522

It has been discussed in https://github.com/thesofproject/sof/pull/8156 and https://github.com/thesofproject/sof/pull/8170

The IMR_CONTEXT_SAVE check was first introduced by https://github.com/thesofproject/sof/pull/8027

### DIFF
--- a/zephyr/lib/cpu.c
+++ b/zephyr/lib/cpu.c
@@ -148,8 +148,7 @@ int cpu_enable_core(int id)
 	 * initialization. By reinitializing the idle thread, we would overwrite the kernel structs
 	 * and the idle thread stack.
 	 */
-	if (!IS_ENABLED(CONFIG_ADSP_IMR_CONTEXT_SAVE) ||
-	    pm_state_next_get(id)->state == PM_STATE_ACTIVE)
+	if (pm_state_next_get(id)->state == PM_STATE_ACTIVE)
 		z_init_cpu(id);
 #endif
 
--- a/src/platform/intel/cavs/platform.c
+++ b/src/platform/intel/cavs/platform.c
@@ -102,6 +102,10 @@ int platform_boot_complete(uint32_t boot_message)
 	return 0;
 }
 
+static struct pm_notifier pm_state_notifier = {
+	.state_exit = cpu_notify_state_exit,
+};
+
 /* Runs on the primary core only */
 int platform_init(struct sof *sof)
 {
@@ -149,6 +153,9 @@ int platform_init(struct sof *sof)
 	ret = dmac_init(sof);
 	if (ret < 0)
 		return ret;
+
+	/* register power states exit notifiers */
+	pm_notifier_register(&pm_state_notifier);
 
 	/* initialize the host IPC mechanisms */
 	trace_point(TRACE_BOOT_PLATFORM_IPC);


kv2019i (commented):
Thanks @RanderWang ! Have you made a test PR that combines this PR with a Zephyr update pointing to your Zephyr patch (like I did https://github.com/thesofproject/sof/pull/8136 ). This would good step to make to ensure the combination passes all SOF CI tests.

The patch itself looks good. I won't mark approve yet to ensure this isn't accidentally merged too soon, but yes, looks good.

lgirdwood (approved):

RanderWang:
> Thanks @RanderWang ! Have you made a test PR that combines this PR with a Zephyr update pointing to your Zephyr patch (like I did #8136 ). This would good step to make to ensure the combination passes all SOF CI tests.
> 
> The patch itself looks good. I won't mark approve yet to ensure this isn't accidentally merged too soon, but yes, looks good.

@kv2019i I tested it in local environment. Current Github CI test doesn't use multi-core topology so we can't depend on Github CI.
My next plan is to add multi-core support in cavs-nocodec and remove cavs-nocodec-multicore topology as we have discussed

zephyr/lib/cpu.c
```diff
@@ -148,8 +148,7 @@ int cpu_enable_core(int id)
 	 * initialization. By reinitializing the idle thread, we would overwrite the kernel structs
 	 * and the idle thread stack.
 	 */
-	if (!IS_ENABLED(CONFIG_ADSP_IMR_CONTEXT_SAVE) ||
-	    pm_state_next_get(id)->state == PM_STATE_ACTIVE)
+	if (pm_state_next_get(id)->state == PM_STATE_ACTIVE)
```
lyakh, line 151: this means, that now if IMR context saving is **not** enabled we potentially call `z_init_cpu()` less frequently, or will it not happen because now the `pm_state_next_get(id)->state == PM_STATE_ACTIVE` will cover all those cases?

RanderWang: sure,  we only need to call z_init_cpu() for dsp core init with above condition for both CAVS  and ACE platforms since CAVS dsp core context save is being supported now.  IMR_CONTEXT_SAVE has nothing to do with z_init_cpu because
IMR_CONTEXT_SAVE is for the whole audio context but z_init_cpu is only for dsp core. IMR_CONTEXT_SAVE is not fit for this check but it can differentiate CAVS & ACE, so it was used.

tmleman (approved):

lgirdwood:
@andyross need your re-review on the Zephyr dependecy to unblock this target for 3rd party SDV usage. Thanks !
@cujomalainey fyi.

RanderWang:
@lgirdwood @kv2019i    I tested this PR with multi-core tplg https://github.com/thesofproject/sof/pull/8240 (include recent zephyr cavs multicore changes). All tests passed.  Next I will update the west for zephyr and try to merge the topology pr.  Thanks!

kv2019i (approved):
Looks good! I guess we can merge this already now, and merge the Zephyr part independently @RanderWang  ?

RanderWang:
> Looks good! I guess we can merge this already now, and merge the Zephyr part independently @RanderWang ?

sure. Zephyr part has been merged. Thanks!

kv2019i:
@RanderWang I now pushed https://github.com/thesofproject/sof/pull/8463 . The Zephyr change does not become visible to SOF until we bring it in via west.yml. Not sure whether we can merged this and #8463 independently, do we need to merge in particular order, or even at the same time...?

RanderWang:
> @RanderWang I now pushed #8463 . The Zephyr change does not become visible to SOF until we bring it in via west.yml. Not sure whether we can merged this and #8463 independently, do we need to merge in particular order, or even at the same time...?

@kv2019i This PR is for multi-core feature and depends on topology support. Now our topology is not merged and waiting for this PR so CI doesn't get any error. (topology PR: https://github.com/thesofproject/sof/pull/8240)

PR #8187, opened 2023-09-12
module-adapter: make interface const

lyakh:
Module adapter interfaces are a collection of methods, implementing the API, they never change. Make them const.


lgirdwood (approved):

RanderWang (approved):
LGTM

PR #8188, opened 2023-09-12
system-agent: remove an unused variable

lyakh:
sys_service_p in
src/audio/module_adapter/library/native_system_agent.c is unused, remove it.


lgirdwood (approved):
@marcinszkudlinski used in the future ?

btian1 (approved):

pjdobrowolski:
That pointer will be used in future for passing system service which provides layer of communication between FW and loadable modules. We can remove that lines but they will come back.

lyakh:
> That pointer will be used in future for passing system service which provides layer of communication between FW and loadable modules. We can remove that lines but they will come back.

@pjdobrowolski I think it would be better to remove them now and see if and when we really need them.

pjdobrowolski (approved):

PR #8189, opened 2023-09-12
IDC: fix message cache handling

lyakh:
When an IDC message is received from another core, the receiver core doesn't have any useful data for it in cache. Writing back the cache can corrupt the message data. Cache has to be invalidated instead.


lgirdwood (commented):
Can we spell out this flow with inline comments in sender and reciver. i.e. it needs to be obvious to the reader when we WB and when we INV. What about message payload ?

lgirdwood:
@mwasko @marcinszkudlinski fyi - for mtl-06

lyakh:
> Can we spell out this flow with inline comments in sender and reciver. i.e. it needs to be obvious to the reader when we WB and when we INV. What about message payload ?

@lgirdwood comments added

tmleman (approved):

btian1 (approved):

lgirdwood:
Known ZRESTRICT Zephyr issue reported. Unrelated.

PR #8192, opened 2023-09-13
Audio: Fix the errors of dai timestamp ops

andrula-song:
use the ts_config of struct dai_data as input/output parameter instead a temporary variable while calling the dai timestamp functions.


src/audio/dai-zephyr.c
```diff
@@ -1669,7 +1669,7 @@ static int dai_copy(struct comp_dev *dev)
 int dai_common_ts_config_op(struct dai_data *dd, struct comp_dev *dev)
 {
 	struct ipc_config_dai *dai = &dd->ipc_config;
-	struct dai_ts_cfg cfg;
+	struct dai_ts_cfg *cfg = (struct dai_ts_cfg *)&dd->ts_config;
```
lyakh, line 1672: why do we actually have `struct dai_ts_cfg` identical to `struct timestamp_cfg` and `struct dai_ts_data` identical to `struct timestamp_data`?

andrula-song: The content of both struct is the same, and the final function in zephyr/driver/dai/intel/ needs such input parameter.

btian1 (commented):
Can we say this is a bug fix?

src/audio/dai-zephyr.c
lyakh: right, so would it be possible to drop `struct timestamp_cfg` and `struct timestamp_data` completely and use `struct dai_ts_cfg` and `struct dai_ts_data` instead everywhere?

andrula-song: okay, then I would like to redefine the struct struct dai_data.

RanderWang (approved):

src/audio/dai-zephyr.c
lyakh: yes, I think that would be good - and define copies of those structures in dai-legacy.h under `#ifndef __ZEPHYR__`

jxstelter (approved):

src/audio/dai-zephyr.c
andrula-song: but if we are using dai_legacy.c, we will use the original struct timestamp_cfg and struct timestamp_data(because the original interface without zephyr needs such input parameters). So I don't think it is good to redefine struct dai_data in dai-legacy.h in this PR.

lyakh: sure, we can rename them for legacy, right? And no, it doesn't have to be fixed in this PR

andrula-song: then I'd like use another PR to redefine for legacy, because I can not test it before I push code to PR right now.

PR #8194, opened 2023-09-13
Cross core fixes for mtl-006-drop-stable

serhiy-katsyuba-intel:


src/ipc/ipc4/helper.c
```diff
@@ -412,6 +406,8 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 		goto e_sink_connect;
 	}
 
+	/* comp_bind() may call comp_bind_remote() and so IDC interrupt should be enabled */
```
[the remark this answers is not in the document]
serhiy-katsyuba-intel: If we assume IPC is working correctly -- there is no need in that critical section. Probably, it was added to workaround some IPC bug(s) we had.
Pipeline should not start if binding is not yet completed. This is guarantied by serial nature of IPC sequence. Start pipeline IPC may only come after FW send response to bind IPC. During bind IPC processing, IPC register has BUSY flag set indicating FW is not ready for next IPC. Only after response to bind IPC, that busy flag is cleared and driver can send next IPC to FW.

src/ipc/ipc4/helper.c
```diff
@@ -433,13 +429,12 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 		source->direction_set = true;
 	}
 
-	irq_local_enable(flags);
-
```
lyakh, line 437: you wrote that it cannot be preempted - how about the scheduler? Cannot a timer interrupt preempt it and then run the LL-scheduler and access these buffers in an inconsistent state? Doesn't seem safe to me

lyakh (commented):
I think as it stands we have a few locations, e.g. in pipeline scheduling core, that aren't cross-core safe. You can find multiple "FIXME" and other comments around the code... I wouldn't be surprised if attempts to run cross-core with these changes result in crashes.

src/ipc/ipc4/helper.c
marcinszkudlinski: agree with @lyakh 
@lyakh problem is that we cannot proceed IDC when interrupts are locked, problem is that we DO need to have connected LL pipelines this week on branch 006, there's no time to do it better as we discussed before

@serhiy-katsyuba-intel maybe we can try to locate all modules structures in shared memory and execute bind on master core?  Just a suggestion without deep analysis

marcinszkudlinski: @serhiy-katsyuba-intel so you suggest that the fix:
> prevent the IPC task getting preempted which could result in buffers being only half connected when a pipeline task gets executed.	

may be actually caused by driver or IPC driver problem - races that lead to binding modules on running pipeline?
the issue in question: https://github.com/thesofproject/sof/issues/6653

I checked that the pipeline state is not verified before bind operation, if the above is true it is worth to reject bind when the pipeline is in incorrect state

serhiy-katsyuba-intel: > I checked that the pipeline state is not verified before bind operation, if the above is true it is worth to reject bind when the pipeline is in incorrect state

Yes, that might be the good option.

Another option is to block preempting without blocking interrupts. Zephyr might have some API for this, if, not, than it could be done on LL scheduler level -- there are zephyr_ll_lock()/unlock(), however, they are implemented now just by calling irq_local_disable()/enable() :(

lyakh: > I assume driver should send bind IPC only when pipelines (with binding components) are not running.

@serhiy-katsyuba-intel @marcinszkudlinski the problem is with forked streams: when a stream is running and then you bind or unbind another stream to or from it.

serhiy-katsyuba-intel: @lyakh, added k_sched_lock() in a separate PR https://github.com/thesofproject/sof/pull/8205 to guard bind/unbind.

However, if bind allowed to be called on running pipeline then we have more problems: .params() and .prepare() will not be called after bind for component on a running pipeline. Buffer stream params (like format, channel number) will be not initialized if capture pipeline is binded to some other running pipeline -- comp_verify_params() only initializes source buffers on capture pipeline.
Skipped .prepare() call may also create problems, e.g. mixin and mixout use .prepare() to select proper gain and mix functions depending of format.
So we need to address those problems if we allow bind to be called for running pipeline.

PR #8196, opened 2023-09-14
imx: sai: Fix sof run fail issue on imx8ulp

TangleZ:
Fix sof run fail issue:
Don't need to enable packed mode for sai on 8ulp.
Also correct frame size for 8ulp.

### DIFF
--- a/src/drivers/imx/sai.c
+++ b/src/drivers/imx/sai.c
@@ -355,6 +355,7 @@ static inline int sai_set_config(struct dai *dai, struct ipc_config_dai *common_
 		return -EINVAL;
 	}
 
+#ifndef CONFIG_IMX8ULP
 	switch (sai->params.tdm_slot_width) {
 	case 8:
 		val_cr4 |= REG_SAI_CR4_FPACK_8;
@@ -365,6 +366,7 @@ static inline int sai_set_config(struct dai *dai, struct ipc_config_dai *common_
 	default:
 		break;
 	}
+#endif
 
 	val_cr4 |= REG_SAI_CR4_FRSZ(sai->params.tdm_slots);
 	val_cr4 |= REG_SAI_CR4_CHMOD;


dbaluta:
This works for me as it is. In the future we might add a parameter in the topology to enable/disable pack mode.

dbaluta (approved):

src/drivers/imx/sai.c
```diff
@@ -365,8 +366,13 @@ static inline int sai_set_config(struct dai *dai, struct ipc_config_dai *common_
 	default:
 		break;
 	}
+#endif
 
+#ifndef CONFIG_IMX8ULP
 	val_cr4 |= REG_SAI_CR4_FRSZ(sai->params.tdm_slots);
+#else
+	val_cr4 |= REG_SAI_CR4_FRSZ((sai->params.tdm_slots == 1) ? 2 : sai->params.tdm_slots);
```
iuliana-prodan, line 374: Why do you add this for 8ULP?
I know that for 8ULP the TDM slots is 16, so you'll have:
`val_cr4 |= REG_SAI_CR4_FRSZ(sai->params.tdm_slots);`
the same as for non-8ULP.

iuliana-prodan (changes requested):

src/drivers/imx/sai.c
```diff
@@ -355,6 +355,7 @@ static inline int sai_set_config(struct dai *dai, struct ipc_config_dai *common_
 		return -EINVAL;
 	}
 
+#ifndef CONFIG_IMX8ULP
 	switch (sai->params.tdm_slot_width) {
```
LaurentiuM1234, line 359: If not supported by the current way the DMAs are configured the packing feature should just be removed altogether as it might break things if the slot width is not 32 (e.g: i.MX8ULP's case). IMO the right way would be to add it together with the support for compatible DMA configurations.

src/drivers/imx/sai.c
TangleZ: TDM slots is different slot width, slot width is 16bit, and slot is 1 for 8ulp. And when slot == 1 means only enable 1 channel but we still need two words in one frame.

iuliana-prodan: Oh, you're right, I confused `tdm_slot_width` with `tdm_slots`.

> And when slot == 1 means only enable 1 channel but we still need two words in one frame.
So, you have 2 words in frame, but only one is useful, the rest is zeroed? Or what?

dbaluta:
@TangleZ have a look at this fix for FRSZ setup:

https://github.com/thesofproject/sof/pull/8200

src/drivers/imx/sai.c
TangleZ: As Daniel said, we would add a parameter for packing mode, then this can be controlled by the user.

iuliana-prodan (approved):

dbaluta (approved):

PR #8197, opened 2023-09-14
helper: add a condition before updating the component direction

fkwasowi:
At the pipeline reset stage, the direction of the components is not always known (example for the pipeline below), resulting in a call to a field that is null (src_buf->source->direction_set/src_buf->source->direction) and consequently a crash/timeout.

image

### DIFF
--- a/src/ipc/ipc4/handler.c
+++ b/src/ipc/ipc4/handler.c
@@ -274,7 +274,7 @@ int ipc4_pipeline_prepare(struct ipc_comp_dev *ppl_icd, uint32_t cmd)
 		switch (status) {
 		case COMP_STATE_INIT:
 			tr_dbg(&ipc_tr, "pipeline %d: reset from init", ppl_icd->id);
-			ret = ipc4_pipeline_complete(ipc, ppl_icd->id);
+			ret = ipc4_pipeline_complete(ipc, ppl_icd->id, cmd);
 			if (ret < 0)
 				ret = IPC4_INVALID_REQUEST;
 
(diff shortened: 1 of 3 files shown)


src/ipc/ipc4/handler.c
```diff
@@ -274,7 +274,7 @@ int ipc4_pipeline_prepare(struct ipc_comp_dev *ppl_icd, uint32_t cmd)
 		switch (status) {
 		case COMP_STATE_INIT:
 			tr_dbg(&ipc_tr, "pipeline %d: reset from init", ppl_icd->id);
-			ret = ipc4_pipeline_complete(ipc, ppl_icd->id);
+			ret = ipc4_pipeline_complete(ipc, ppl_icd->id, cmd);
```
ranj063, line 277: @fkwasowi may I ask why we even need to invoke ipc4_pipeline_complete() during pipeline reset?

fkwasowi:
> Is this addressing a crash, that's caught by CI? Or only in your testing? Should this go to any stable branches too?

Due to this problem, the 09_oed_dmic_mute_unmute test is blocked. The issue is reported in the following HSD: https://hsdes.intel.com/appstore/article/#/18031254461

tmleman (approved):

dabekjakub (approved):

fkwasowi:
SOFCI TEST

fkwasowi:
SOFCI TEST

src/ipc/ipc4/handler.c
kv2019i: This is still unanswered, but I assume this is a more complex. git blame shows this code has been here for a long, long time...

kv2019i (approved):
Some unanswered questions, but given this fixes a DSP crash, seems we can proceed with merge.

kv2019i:
SOFCI TEST

kv2019i:
Uh, sorry @fkwasowi I didn't notice you had pinged the CI already. But let's see that if this now completes...

src/ipc/ipc4/handler.c
fkwasowi: I think this is the answer to the question: https://github.com/thesofproject/sof/commit/8541d68b507ef28b68a79713634e3b434d9f53bb

fkwasowi:
SOFCI TEST

kv2019i:
The one error in https://sof-ci.01.org/sofpr/PR8197/build13064/devicetest is a DUT issue. error in log:
```
2023-09-18 10:40:04 UTC [REMOTE_ERROR] ktime=115 sof-test PID=3493: unexpected value in /tmp/sof-test-card0.lock! Concurrent testing?
```

Proceeding with merge.

Issue #8178, opened 2023-09-07
[FEATURE] remove SOF build time dependency to rimage headers


kv2019i:
**Is your feature request related to a problem? Please describe.**
rimage was originally moved to a different repository as it is an independent tool from SOF. 

With changes since SOF2.4, we have more and more direct dependencies from SOF to rimage. IPC4 protocol requires metadata that is defined in firmware manifest (data maintained in toml files, rimage encodes this to firmware binary file).
Most recently, loadable module support was added to SOF firmware and this code depends on firmware file structure definitions that are needed both by rimage and the new SOF module loader.

The build dependencies are causing issues to downstream SOF users (e.g. https://github.com/zephyrproject-rtos/zephyr/pull/62262  ).

Having the rimage manifest changes (like https://github.com/thesofproject/rimage/pull/177) not tested by full SOF CI, is causing operational issues (problems caused by rimage changes are detected too late).

**Describe the solution you'd like**
Merge rimage back to sof repository as it is now more integrate part of SOF. 

**Describe alternatives you've considered**
- alternative: duplicate manifest headers in SOF (these are part of the loadable modules ABI)
      - cons: does not help with SOF CI woes with rimage PRs

kv2019i:
FYI @nashif @lgirdwood @marc-hb @softwarecki @aiChaoSONG

marc-hb:
cc: @andyross , @dbaluta , @iuliana-prodan , @dcpleung 

> rimage was originally moved to a different repository as it is an independent tool from SOF.

Anyone remembers the rationale for a separate rimage repo? We don't want to go back and forth and back. I know booting Zephyr tests on Intel/NXP hardware is possible *with* rimage and *without* SOF. Could have this been a reason?

Having more git repos for no good reason makes everything more complicated. Among others and as mentioned above, it makes CI especially more complicated: how do you submit an rimage change (or Zephyr) to SOF testing? You can:
- https://github.com/thesofproject/rimage/blob/main/README.md
- https://docs.zephyrproject.org/latest/develop/modules.html#process-for-submitting-changes-to-existing-modules

... BUT it's tedious and experience shows most people don't take the time.

[...]

PR #8198, opened 2023-09-14
topology1: add sof-rpl-rt711-l0-rt1316-l12.tplg support

bardliao:
Adding sof-rpl-rt711-l0-rt1316-l12.tplg. The topology is the same as sof-rpl-rt711-l0-rt1316-l12-rt714-l3, but without local mic (rt714).


mengdonglin:
@kv2019 This is for customer RPL device. Should such topology updates go to v2.2.x or v2.7?

plbossart (approved):

lgirdwood (approved):

kv2019i:
@mengdonglin so this needs to be backported to stable-v2.2 and then include it to a sof-bin release that will include the latest v2.2-tplg package

@lgirdwood we probably should remove all these Intel tplg1 files from main, to make it clear tplg1 updates (for Intel platforms) need to be submitted to stable-v2.2

plbossart:
> @lgirdwood we probably should remove all these Intel tplg1 files from main, to make it clear tplg1 updates (for Intel platforms) need to be submitted to stable-v2.2

Or move topologies out of the sof repos....

marc-hb:
> Or move topologies out of the sof repos....

And then move them back in a couple years later :-)
- https://github.com/thesofproject/sof/issues/8178#issuecomment-1710534303

Quoting that issue:
> https://en.wikipedia.org/wiki/Conway%27s_law aside (does not apply here), the ability to make arbitrary combinations is probably the only good reason to have multiple, independent repos. For instance we want the latest SOF-TEST branch to be able to test most SOF branches out of the box. We don't want to branch SOF-TEST every time that we branch SOF, that would be much more work. Obviously, you also don't want SOF-TEST code to change while bisecting SOF. So: a different repo for SOF-TEST. It's different for lower level CMocka unit tests which are in the same repo.

So is there a frequent need to test arbitrary combinations of sof.git version X with sof.git version Y? I haven't heard of any so far. If not then a separate repo is a lot of overhead for no value.

PR #8200, opened 2023-09-14
drivers: imx: sai: Fix frame size computation

dbaluta:
When I2S-like interface is enabled we must have
2 words in a frame because this is how I2S works (it has 2 channels)!

So, even if the user plays a mono file we must configure 2 words per slot and the second slot will be masked so the users will get silence on the second channel.

### DIFF
--- a/src/drivers/imx/sai.c
+++ b/src/drivers/imx/sai.c
@@ -229,6 +229,7 @@ static inline int sai_set_config(struct dai *dai, struct ipc_config_dai *common_
 	uint32_t val_cr2 = 0, val_cr4 = 0, val_cr5 = 0;
 	uint32_t mask_cr2 = 0, mask_cr4 = 0, mask_cr5 = 0;
 	uint32_t clk_div;
+	bool tdm_enable;
 	struct sai_pdata *sai = dai_get_drvdata(dai);
 
 	sai->config = *config;
@@ -250,6 +251,14 @@ static inline int sai_set_config(struct dai *dai, struct ipc_config_dai *common_
 	} else {
 		clk_div = (config->sai.mclk_rate / config->sai.bclk_rate / 2) - 1;
 	}
+
+	/* TDM mode is enabled only when fmt is dsp_a or dsp_b and
+	 * we can have from 1 to 32 channels.
+	 * for any other formats we assume I2S like interface where
+	 * audio frames have 2 channels, even for mono scenario. The
+	 * second channel will be masked out.
+	 */
+	tdm_enable = false;
 
 	switch (config->format & SOF_DAI_FMT_FORMAT_MASK) {
 	case SOF_DAI_FMT_I2S:
@@ -285,6 +294,7 @@ static inline int sai_set_config(struct dai *dai, struct ipc_config_dai *common_
 		val_cr2 |= REG_SAI_CR2_BCP;
 		val_cr4 |= REG_SAI_CR4_FSE;
 		val_cr4 |= REG_SAI_CR4_SYWD(1U);
+		tdm_enable = true;
 		break;
 	case SOF_DAI_FMT_DSP_B:
 		/*
@@ -293,6 +303,7 @@ static inline int sai_set_config(struct dai *dai, struct ipc_config_dai *common_
 		 */
 		val_cr2 |= REG_SAI_CR2_BCP;
 		val_cr4 |= REG_SAI_CR4_SYWD(1U);
+		tdm_enable = true;
 		break;
 	case SOF_DAI_FMT_PDM:
 		val_cr2 |= REG_SAI_CR2_BCP;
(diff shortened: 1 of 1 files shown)


src/drivers/imx/sai.c
```diff
@@ -229,6 +232,7 @@ static inline int sai_set_config(struct dai *dai, struct ipc_config_dai *common_
 	uint32_t val_cr2 = 0, val_cr4 = 0, val_cr5 = 0;
 	uint32_t mask_cr2 = 0, mask_cr4 = 0, mask_cr5 = 0;
 	uint32_t clk_div;
+	int tdm_mode;
```
btian1, line 235: how about directly use bool tdm_mode? false naturally means disable, and true is enable.

btian1 (approved):

src/drivers/imx/sai.c
```diff
@@ -366,7 +375,12 @@ static inline int sai_set_config(struct dai *dai, struct ipc_config_dai *common_
...
 	}
 
-	val_cr4 |= REG_SAI_CR4_FRSZ(sai->params.tdm_slots);
+	if (tdm_mode == TDM_ENABLE)
+		val_cr4 |= REG_SAI_CR4_FRSZ(sai->params.tdm_slots);
+	else
+		val_cr4 |= REG_SAI_CR4_FRSZ(
+		    (sai->params.tdm_slots == 1) ? 2 : sai->params.tdm_slots);
+
 	val_cr4 |= REG_SAI_CR4_CHMOD;
```
TangleZ, line 384: It seems conflict with the variable tdm_mode, because we always set mode OUTPUT mode here?

src/drivers/imx/sai.c
```diff
@@ -251,6 +255,9 @@ static inline int sai_set_config(struct dai *dai, struct ipc_config_dai *common_
 		clk_div = (config->sai.mclk_rate / config->sai.bclk_rate / 2) - 1;
 	}
 
+	/* TDM mode is enabled only when fmt is dsp_a or dsp_b */
+	tdm_mode = TDM_DISABLE;
+
 	switch (config->format & SOF_DAI_FMT_FORMAT_MASK) {
 	case SOF_DAI_FMT_I2S:
```
TangleZ, line 262: Here no need to enable tdm_mode?

src/drivers/imx/sai.c
dbaluta: @btian1 yes, this makes sense.

src/drivers/imx/sai.c
dbaluta: I don't think TDM mode from my patch is the same as TDM mode configuration of output mode.

TDM mode says that transmit data pins are tri-stated when slots are masked or channels disabled. You can have I2S mode enabled and output mode set to tdm.

src/drivers/imx/sai.c
dbaluta: The reasoning here is like that: when user wants to use TDM mode (e.g up to 32 channels) they will set config->format to dsp_a and dsp_b. 

This way DSP will now that we will have more than 2 channels.

I think this is a little bit confusing but it is a convention between Linux kernel and dsp firmware. 

We can refine this code in the future.

src/drivers/imx/sai.c
```diff
@@ -366,7 +375,12 @@ static inline int sai_set_config(struct dai *dai, struct ipc_config_dai *common_
 		break;
 	}
 
-	val_cr4 |= REG_SAI_CR4_FRSZ(sai->params.tdm_slots);
+	if (tdm_mode == TDM_ENABLE)
+		val_cr4 |= REG_SAI_CR4_FRSZ(sai->params.tdm_slots);
+	else
+		val_cr4 |= REG_SAI_CR4_FRSZ(
+		    (sai->params.tdm_slots == 1) ? 2 : sai->params.tdm_slots);
```
iuliana-prodan, line 382: This is more restricted than the "old" configuration.
Meaning, before these changes we were setting 
`val_cr4 |= REG_SAI_CR4_FRSZ(sai->params.tdm_slots);`
for all formats and platforms.

There is no possibility to have 1 channel (or 1 word per frame) in other case than DSP-A or DSP-B format?

src/drivers/imx/sai.c
TangleZ: Got it, then ok for me.

src/drivers/imx/sai.c
TangleZ: I think the reason is as Daniel said, we need to use 2 words even on mono, because the "wire" from DAI to codec uses I2S protocol! which has a fixed number of 2 channels. @dbaluta ,Please correct me if I was wrong.

dbaluta: this is a convention we need to make between SAI implementation  and setting the format in topology.

We sai like this: we will implement true TDM with dsp_a/dsp_b and true I2S with the rest of the formats. This is not ideal but is best we can do with little effort.

iuliana-prodan: Please add a comment with this explanation, in code, and make tdm_mode bool as suggested here https://github.com/thesofproject/sof/pull/8200/files#r1326674287

dbaluta:
@btian1 @TangleZ @iuliana-prodan addressed all your comments.

iuliana-prodan (approved):

TangleZ (approved):

dbaluta:
SOFCI TEST

PR #8205, opened 2023-09-15
ipc4: Add preeption guard to bind/unbind

serhiy-katsyuba-intel:
This is a workaround to protect bind/unbind operations to be preempted by LL scheduler.

### DIFF
--- a/src/ipc/ipc4/helper.c
+++ b/src/ipc/ipc4/helper.c
@@ -335,6 +335,15 @@ static struct comp_buffer *ipc4_create_buffer(struct comp_dev *src, struct comp_
 	return buffer_new(&ipc_buf);
 }
 
+/* FIXME: avoid Zephyr internal platform specific defines */
+#ifdef CONFIG_SOC_SERIES_INTEL_ACE
+	#define lock_ll_sched(flags)    irq_disable(ACE_IRQ_TO_ZEPHYR(ACE_INTL_TTS))
+	#define unlock_ll_sched(flags)  irq_enable(ACE_IRQ_TO_ZEPHYR(ACE_INTL_TTS))
+#else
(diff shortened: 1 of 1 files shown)


src/ipc/ipc4/helper.c
```diff
@@ -389,6 +389,14 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
 
+	/* k_sched_lock() does not prevent current thread to be preempted if the thread
+	 * calls something like sleep(). So just in case let's also use irq_local_disable()
+	 * to guard two com_buffer_connect() calls. Interrupts should be re-enabled again for
+	 * comp_bind_remote() to work. Let's pray nobody calls sleep() in .bind() handler.
+	 * This is intendent to be a temporary workaround and should be replaced with some
+	 * proper robust guard solution.
+	 */
+	k_sched_lock();
 	irq_local_disable(flags);
```
marcinszkudlinski, line 400: do we still need irq_disable than?

Even if - we need to use irq_lock instead of "local disable"

serhiy-katsyuba-intel: Without irq_local_disable(), timer interrupt may come and increase LL scheduler semaphore. That will not start LL scheduler because of k_shed_lock(), however, if inside comp_buffer_connect() somebody calls sleep(), then LL scheduler will start. Having irq_local_disable() will not allow timer interrupt to increase semaphore. So, its better to have irq_local_disable() at least for duration of two com_buffer_connect(). However, we cannot have interrupts disabled during comp_bind_remote() as we rely of k_shed_lock() and that nobody uses sleep() inside .bind() handler.
So k_shed_lock() is not 100% replacement for irq_local_disable() but just a good enough workaround we can quickly introduce.

keqiaozhang:
SOFCI TEST

PR #8207, opened 2023-09-15
Audio: Volume: Handle all volume ramp types

singalsu:
The first patch adds a conversion from IPC4 to SOF volume ramp types.

The second patch changes further volume to be able to handle the supported ramp types.

The third patch adds to gain.conf the other supported ramp types. The default is kept as windows fade.

### DIFF
--- a/src/audio/volume/peak_volume.h
+++ b/src/audio/volume/peak_volume.h
@@ -24,6 +24,7 @@
 #ifndef __SOF_IPC4_PEAK_VOL_H__
 #define __SOF_IPC4_PEAK_VOL_H__
 
+#include <ipc/topology.h>
 #include <ipc4/base-config.h>
 
 enum ipc4_vol_mode {
@@ -43,7 +44,11 @@ enum ipc4_peak_volume_param {
 
 enum ipc4_curve_type {
 	IPC4_AUDIO_CURVE_TYPE_NONE = 0,
-	IPC4_AUDIO_CURVE_TYPE_WINDOWS_FADE
+	IPC4_AUDIO_CURVE_TYPE_WINDOWS_FADE,
+	IPC4_AUDIO_CURVE_TYPE_LINEAR,
+	IPC4_AUDIO_CURVE_TYPE_LOG,
+	IPC4_AUDIO_CURVE_TYPE_LINEAR_ZC,
+	IPC4_AUDIO_CURVE_TYPE_LOG_ZC,
 };
 
 static const uint32_t IPC4_ALL_CHANNELS_MASK = 0xffffffff;
@@ -74,4 +79,28 @@ struct ipc4_peak_volume_module_cfg {
 	struct ipc4_base_module_cfg base_cfg;
 	struct ipc4_peak_volume_config config[];
 } __packed __aligned(8);
+
+static inline enum sof_volume_ramp ipc4_curve_type_convert(enum ipc4_curve_type ipc4_type)
+{
+	switch (ipc4_type) {
+	case IPC4_AUDIO_CURVE_TYPE_WINDOWS_FADE:
+		return SOF_VOLUME_WINDOWS_FADE;
+
+	case IPC4_AUDIO_CURVE_TYPE_LINEAR:
+		return SOF_VOLUME_LINEAR;
+
+	case IPC4_AUDIO_CURVE_TYPE_LOG:
+		return SOF_VOLUME_LOG;
+
+	case IPC4_AUDIO_CURVE_TYPE_LINEAR_ZC:
+		return SOF_VOLUME_LINEAR_ZC;
+
+	case IPC4_AUDIO_CURVE_TYPE_LOG_ZC:
+		return SOF_VOLUME_LOG_ZC;
+
+	case IPC4_AUDIO_CURVE_TYPE_NONE:
+	default:
+		return SOF_VOLUME_WINDOWS_NO_FADE;
+	}
+}
 #endif
--- a/src/audio/volume/volume.c
+++ b/src/audio/volume/volume.c
@@ -283,18 +283,22 @@ static inline void volume_ramp(struct processing_module *mod)
 		 * calculated from previous gain and ramp time. The slope
 		 * coefficient is calculated in volume_set_chan().
 		 */
-#if defined CONFIG_COMP_VOLUME_WINDOWS_FADE && defined CONFIG_COMP_VOLUME_LINEAR_RAMP
-		if (cd->ramp_type == SOF_VOLUME_WINDOWS_FADE)
+		switch (cd->ramp_type) {
+#if CONFIG_COMP_VOLUME_WINDOWS_FADE
+		case SOF_VOLUME_WINDOWS_FADE:
 			new_vol = volume_windows_fade_ramp(cd, ramp_time, i);
-		else
-			new_vol = volume_linear_ramp(cd, ramp_time, i);
-#elif defined CONFIG_COMP_VOLUME_WINDOWS_FADE
-		new_vol = volume_windows_fade_ramp(cd, ramp_time, i);
-#elif defined CONFIG_COMP_VOLUME_LINEAR_RAMP
-		new_vol = volume_linear_ramp(cd, ramp_time, i);
-#else
-		new_vol = tvolume;
+			break;
 #endif
+#if CONFIG_COMP_VOLUME_LINEAR_RAMP
+		case SOF_VOLUME_LINEAR:
+		case SOF_VOLUME_LINEAR_ZC:
+			new_vol = volume_linear_ramp(cd, ramp_time, i);
+			break;
+#endif
+		default:
+			new_vol = tvolume;
+		}
+
 		if (volume < tvolume) {
 			/* ramp up, check if ramp completed */
 			if (new_vol < tvolume)
--- a/src/audio/volume/volume_ipc4.c
+++ b/src/audio/volume/volume_ipc4.c
@@ -58,13 +58,10 @@ static int set_volume_ipc4(struct vol_data *cd, uint32_t const channel,
 	cd->mvolume[channel] = 0;
 	/* set muted as false*/
 	cd->muted[channel] = false;
-#if CONFIG_COMP_VOLUME_WINDOWS_FADE
+
 	/* ATM there is support for the same ramp for all channels */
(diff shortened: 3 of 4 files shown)


singalsu:
@ranj063 @kv2019i Does this change prevent topology to work with IPC4 reference firmware - and does it matter? If that is the case, then we need to also re-order for IPC4 the table https://github.com/thesofproject/sof/blob/036da2ac437c02b838e273c70aee83a2fafb6c59/src/include/ipc/topology.h#L151 or create new ipc4 specific table with windows_no_fade and windows_fade as first entries.

For performance reasons we want to enable topology to select also lighter to compute other ramp types.

tools/topology/topology2/include/components/gain.conf
```diff
@@ -151,7 +159,7 @@ Class.Widget."gain" {
 	uuid 			"A8:A9:BC:61:D0:18:18:4A:8E:7B:26:39:21:98:04:B7"
 	no_pm			"true"
 	cpc 			10183
-	curve_type		"fade"
+	curve_type		"windows_fade"
```
btian1, line 162: is it still default as fade, right? how to make it configurable?

singalsu: Yep, but it should be another PR (change default to "linear" and have e.g. in nocodec topologies this set into "windows_fade" to keep it tested).

The intention with this patch is to fix wrong IPC sent to firmware.

ranj063:
> @ranj063 @kv2019i Does this change prevent topology to work with IPC4 reference firmware - and does it matter? If that is the case, then we need to also re-order for IPC4 the table
> 
> https://github.com/thesofproject/sof/blob/036da2ac437c02b838e273c70aee83a2fafb6c59/src/include/ipc/topology.h#L151
> 
> or create new ipc4 specific table with windows_no_fade and windows_fade as first entries.
> For performance reasons we want to enable topology to select also other (lighter) ramp types.

@singalsu im not sure if this will affect the ref fw. But AFAIK this wouldnt be the first thing that breaks the ref fw. Maybe we should limit to testing just the passthrough pipelines with the ref fw to avoid any issues.

tools/topology/topology2/include/components/gain.conf
btian1: ok, @singalsu  @ranj063 @kv2019i @lgirdwood, do we have a way to config this curve_type to make it:
1. windows  sof build, set to windoes_fade.
2. linux/chrome build, set to linear.

we should have a delta config to do this, since windows is not always same with chrome.

btian1 (approved):

singalsu:
> > @ranj063 @kv2019i Does this change prevent topology to work with IPC4 reference firmware - and does it matter? If that is the case, then we need to also re-order for IPC4 the table
> > https://github.com/thesofproject/sof/blob/036da2ac437c02b838e273c70aee83a2fafb6c59/src/include/ipc/topology.h#L151
> > 
> > or create new ipc4 specific table with windows_no_fade and windows_fade as first entries.
> > For performance reasons we want to enable topology to select also other (lighter) ramp types.
> 
> @singalsu im not sure if this will affect the ref fw. But AFAIK this wouldnt be the first thing that breaks the ref fw. Maybe we should limit to testing just the passthrough pipelines with the ref fw to avoid any issues.

We could 
- Decide to change the ramp enum types for IPC4 so that SOF volume would remain compatible with Windows driver. In this case the binary would be compatible.
- Or make the enums reference FW compatible with a kconfig build option
- Ignore the difference and test with pass-through topologies only the reference FW

What should we do?

marcinszkudlinski (approved):

tools/topology/topology2/include/components/gain.conf
singalsu: This PR keeps operation unchanged with windows fade as default, just re-arranged the enum values.

I'd like to make a separate PR to change the default type to linear and change nocodec topology to use the windows fade curve.

src/include/ipc/topology.h
```diff
@@ -156,6 +166,7 @@ enum sof_volume_ramp {
 	SOF_VOLUME_WINDOWS_FADE,
 	SOF_VOLUME_WINDOWS_NO_FADE,
 };
+#endif
```
kv2019i, line 169: Ouch, this is a problem. We have these definitions also in Linux kernel and we can't have #ifdef for IPC version. So if we need separate definitions, we need to use a different name.

kv2019i (changes requested):
Hmm, the condition enum def is a problem. Can we have a separate enum for IPC4?

RanderWang (approved):

src/include/ipc/topology.h
btian1: @singalsu , @kv2019i  maybe no need to add #if, simply use new enum replace old ones, the reason is: sof use these as it is defined, I mean the sequence does not have impact on volume for both ipc3 and ipc4.

singalsu: @kv2019i The kernel treats this as generic u32

https://github.com/thesofproject/linux/blob/cffcac365eac0d60dc38959cadaa5078bc924b8e/sound/soc/sof/ipc3-topology.c#L119

https://github.com/thesofproject/linux/blob/cffcac365eac0d60dc38959cadaa5078bc924b8e/sound/soc/sof/ipc4-topology.c#L130

There is this header but it's not used. Could we remove it instead? 

https://github.com/thesofproject/linux/blob/cffcac365eac0d60dc38959cadaa5078bc924b8e/include/sound/sof/topology.h#L138

singalsu: @kv2019i As other fix approach could duplicate all ramp types to https://github.com/thesofproject/sof/blob/b608ec840c80b3e27bc8cc4d4d3a401f3929b396/src/audio/volume/peak_volume.h#L44 and add a conversion function from `enum ipc4_peak_volume_param` to `enum sof_volume_ramp`. What do you prefer?

kv2019i: @singalsu  Can we have separate enums for IPC3 and IPC4 and also remove the unused enum in kernel? I think that would be the cleanest (no ifdefs in headers, no mismatch between definitions between kernel and FW).

singalsu:
Re-request review from everyone since the patch is now very different. Hopefully like @kv2019i suggested. Since the code now touches a part with known issue I included also fix for code that prevented other than windows fade to work in ipc4 build.

src/audio/volume/peak_volume.h
```diff
@@ -43,7 +44,11 @@ enum ipc4_peak_volume_param {
 
 enum ipc4_curve_type {
 	IPC4_AUDIO_CURVE_TYPE_NONE = 0,
-	IPC4_AUDIO_CURVE_TYPE_WINDOWS_FADE
+	IPC4_AUDIO_CURVE_TYPE_WINDOWS_FADE,
+	IPC4_AUDIO_CURVE_TYPE_LINEAR,
+	IPC4_AUDIO_CURVE_TYPE_LOG,
+	IPC4_AUDIO_CURVE_TYPE_LINEAR_ZC,
+	IPC4_AUDIO_CURVE_TYPE_LOG_ZC,
```
btian1, line 51: I like this sequence, 0, with windows fade, and then linear.
it brings:
1. compatible with windows.
2. easy to remove later curve type.

src/audio/volume/volume.c
```diff
@@ -283,18 +283,22 @@ static inline void volume_ramp(struct processing_module *mod)
...
+		switch (cd->ramp_type) {
+#if CONFIG_COMP_VOLUME_WINDOWS_FADE
+		case SOF_VOLUME_WINDOWS_FADE:
 			new_vol = volume_windows_fade_ramp(cd, ramp_time, i);
-		else
+			break;
+#endif
+#if CONFIG_COMP_VOLUME_LINEAR_RAMP
+		case SOF_VOLUME_LINEAR:
+		case SOF_VOLUME_LINEAR_ZC:
```
btian1, line 294: does LINEAR_ZC also use same ramp?

src/audio/volume/volume_ipc4.c
```diff
@@ -86,8 +86,11 @@ static void init_ramp(struct vol_data *cd, uint32_t curve_duration, uint32_t tar
 	/* In IPC4 driver sends curve_duration in hundred of ns - it should be
 	 * converted into ms value required by firmware
 	 */
-	cd->initial_ramp = Q_MULTSR_32X32((int64_t)curve_duration,
-					  Q_CONVERT_FLOAT(1.0 / 10000, 31), 0, 31, 0);
+	if (cd->ramp_type == SOF_VOLUME_WINDOWS_NO_FADE)
+		cd->initial_ramp = 0;
```
btian1, line 90: I'd like to add:
cd->ramp_finished = true; 
to fast skip the ramp in process.

andrula-song (approved):

RanderWang:
> @ranj063 @kv2019i Does this change prevent topology to work with IPC4 reference firmware - and does it matter? If that is the case, then we need to also re-order for IPC4 the table
> 
> https://github.com/thesofproject/sof/blob/036da2ac437c02b838e273c70aee83a2fafb6c59/src/include/ipc/topology.h#L151
> or create new ipc4 specific table with windows_no_fade and windows_fade as first entries.
> 
> For performance reasons we want to enable topology to select also other (lighter) ramp types.

Actually the old SOF definition is conflicted with windows usage. This PR makes it aligned with windows
````c
enum CurveType
{
    AUDIO_CURVE_TYPE_NONE = 0,
    AUDIO_CURVE_TYPE_WINDOWS_FADE = 1
};
````

src/audio/volume/volume.c
singalsu: Yes both use the linear equation.

src/audio/volume/volume_ipc4.c
singalsu: Yes, good catch, I'll add it and test it.

lgirdwood (approved):

lgirdwood:
@kv2019i did enum get fixed to your approval ?

singalsu:
"46 successful and 1 failing checks" , rebase and retry with same patch.

lgirdwood (approved):

lgirdwood:
> "46 successful and 1 failing checks" , rebase and retry with same patch.

Unrelated Zephyr build issue.

kv2019i (approved):
Thanks @singalsu , looks good now!

PR #8208, opened 2023-09-15
Tools: Topology2: Change in capture gain curve_duration to 50 ms

singalsu:
This change increases the ramp duration from 20 ms to 50 ms. It lowers the peak load of peak volume component due to longer same gain value blocks. The internal update rate for gain becomes 250 us instead of 125 us. The longer fade-in ramp also conceals better possible analog capture start transients.

This changes for 4ch capture for gain.11.1 in sof-hda-generic-4ch.tplg from

CPU_PEAK(MAX) = 21.95
PEAK(MAX)/AVG(AVG) = 7.51

to

CPU_PEAK(MAX) = 9.07
PEAK(MAX)/AVG(AVG) = 3.12

### DIFF
--- a/tools/topology/topology2/platform/intel/dmic-generic.conf
+++ b/tools/topology/topology2/platform/intel/dmic-generic.conf
@@ -102,6 +102,7 @@ IncludeByKey.PASSTHROUGH {
 					}
 				}
 				Object.Widget.gain.1 {
+					curve_duration 500000
 					num_input_audio_formats 2
 					num_output_audio_formats 2
 					Object.Base.audio_format.1 {
--- a/tools/topology/topology2/cavs-nocodec-multicore.conf
+++ b/tools/topology/topology2/cavs-nocodec-multicore.conf
@@ -335,6 +335,7 @@ IncludeByKey.PASSTHROUGH {
 							}
 						}
 						Object.Widget.gain.1 {
+							curve_duration 500000
 							num_input_audio_formats 2
 							num_output_audio_formats 2
 							Object.Base.audio_format.1 {
@@ -396,6 +397,7 @@ IncludeByKey.PASSTHROUGH {
 							}
 						}
 						Object.Widget.gain.1 {
+							curve_duration 500000
 							num_input_audio_formats 2
 							num_output_audio_formats 2
 							Object.Base.audio_format.1 {
@@ -454,6 +456,7 @@ IncludeByKey.PASSTHROUGH {
 							}
 						}
 						Object.Widget.gain.1 {
+							curve_duration 500000
 							num_input_audio_formats 2
 							num_output_audio_formats 2
 							Object.Base.audio_format.1 {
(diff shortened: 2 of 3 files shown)


btian1 (commented):
change to 40/50/60ms have same result for peak ramp results, better to have a check for the result.

tools/topology/topology2/platform/intel/dmic-generic.conf
```diff
@@ -102,6 +102,7 @@ IncludeByKey.PASSTHROUGH {
 					}
 				}
 				Object.Widget.gain.1 {
+					curve_duration 500000
```
btian1, line 105: shall we set longer ramp for dmic capture? like 100ms? capture is something different with playback.

singalsu: The time need to conceal start transient in MTL and LNL for DMIC 48 kHz is about this length. It has been longer for cAVS platforms but since they use SOF v2.2 and tplg1 there is no need to address separately. So I'd keep this as 50 ms, same for all capture.

btian1 (approved):

lgirdwood (approved):

lgirdwood:
@marcinszkudlinski @mwasko fyi - good improvement on MCPS

lgirdwood:
@wszypelt @lrudyX fyi - not sure what volume gain curve duration you guys are using in CI today.

btian1:
@singalsu @lgirdwood 
https://sof-ci.ostc.intel.com/#/result/planresultdetail/32024

seems CI still failed, either change CI pass reference or revert this patch.

mszleszy:
@lgirdwood tests uses different curve duration, and according to my knowledge it should not be static as curve length is being defined in InitInstance ipc or LargeConfigSet message of Gain module 
example for 0.1s of fade in:
image

singalsu:
> @singalsu @lgirdwood https://sof-ci.ostc.intel.com/#/result/planresultdetail/32024
> 
> seems CI still failed, either change CI pass reference or revert this patch.

@btian1 If you mean issue #8238 the bug is in CI, not in topology or gain component. Alsabat or using alsabat from sof-test should have a way to ignore from start the ramp time.

singalsu:
> @lgirdwood tests uses different curve duration, and according to my knowledge it should not be static as curve length is being defined in InitInstance ipc or LargeConfigSet message of Gain module example for 0.1s of fade in: image

@mszleszy Yes, the gain component follows the curve_duration and curve_type request from init() IPC. But currently the curve duration and curve type from successive volume commands is ignored. Also we can't have different curve per channel.

mszleszy:
@singalsu as far as i know setting up different configs per channel was supported up to cavs2.5 It's not required any longer but the structure stayed the same for back compatibility. 
I can speak only form fw QA perspective that PeakVol and Gain configuration worked pretty well do far, so any new fail on those modules looks like a regression. But if necessary we can do a deep dive for this case

singalsu:
> @singalsu as far as i know setting up different configs per channel was supported up to cavs2.5 It's not required any longer but the structure stayed the same for back compatibility. I can speak only form fw QA perspective that PeakVol and Gain configuration worked pretty well so far, so any new fail on those modules looks like a regression. But if necessary we can do a deep dive for this case

Yep, we should not in attempt to tame the peak MCPS cause any regressions.

PR #8211, opened 2023-09-18
audio: volume: move volume related config to component config

btian1:
volume related config are missed to move during volume move out from module adapter, move volume kconfig from module adapter folder to component folder.


btian1:
@kv2019i , do I need fix the code style check patch problem?

kv2019i (approved):
No need to change code due to checkpatch warnings, better to move as-is.

src/audio/Kconfig
```diff
@@ -711,4 +711,85 @@ config COMP_MODULE_SHARED_LIBRARY_BUILD
 	  Select if you want to build modules as shared objects that can be used to run
 	  pipelines on the host with the testbench or the ALSA plugin.
 
+config COMP_VOLUME
```
lgirdwood, line 714: Can we create a Kconfig for each Module i.e. ```src/audio/volume/Kconfig``` and include it from here.

lgirdwood (commented):
Content is good, we jsut need to have the Kconfig in the volume directory.

singalsu (approved):

kv2019i (changes requested):
Wait a sec, @btian1 commit says we move files but patch diff shows only adding lines. So are the kconfig defs still left to module_adapter?

lgirdwood (approved):

lgirdwood:
@btian1 can you check internal CI. Thanks

dnikodem (approved):

dbaluta (approved):

Issue #8224, opened 2023-09-19
[BUG][mtl-006] I/O error when running aplay on speaker pipeline


keqiaozhang:
**Describe the bug**
It seems that there's a new regression on MTL-SDW(single core) platforms. there's an I/O error when checking aplay on speaker pipeline at every first time. The reproduce rate is 100%. No critical errors in kernel log, but there're many `dai_zephyr_multi_endpoint_copy(): nothing to copy` and XRUNs detected in mtrace.

```
[    0.103653] <inf> pipe: add_pipeline_cps_consumption: Sum of KCPS consumption: 44306, core: 0
[    0.103656] <inf> pipe: add_pipeline_cps_consumption: Registering module: 0x20000 KCPS consumption: 4996, core: 0
[    0.103661] <inf> pipe: add_pipeline_cps_consumption: Sum of KCPS consumption: 49302, core: 0
[    0.103671] <inf> ll_schedule: zephyr_ll_task_schedule_common: task add 0xa010b800 0x201e0U priority 0 flags 0x0
[    0.104108] <wrn> dai_comp: dai_zephyr_multi_endpoint_copy: comp:1 0x40001 dai_zephyr_multi_endpoint_copy(): nothing to copy
[    0.104171] <inf> host_comp: host_get_copy_bytes_normal: comp:0 0x40000 no bytes to copy, available samples: 0, free_samples: 192
[    0.105105] <err> dma_dw_common: dw_dma_get_status: xrun detected
[    0.105111] <wrn> dai_comp: dai_zephyr_multi_endpoint_copy: comp:1 0x40001 dai_zephyr_multi_endpoint_copy(): nothing to copy
[    0.105120] <inf> host_comp: host_get_copy_bytes_normal: comp:0 0x40000 no bytes to copy, available samples: 0, free_samples: 192
[    0.106116] <err> dma_dw_common: dw_dma_get_status: xrun detected
[    0.106121] <err> dma_dw_common: dw_dma_get_status: xrun detected
[    0.106125] <wrn> dai_comp: dai_zephyr_multi_endpoint_copy: comp:1 0x40001 dai_zephyr_multi_endpoint_copy(): nothing to copy
[    0.106135] <inf> host_comp: host_get_copy_bytes_normal: comp:0 0x40000 no bytes to copy, available samples: 0, free_samples: 192
[    0.107111] <err> dma_dw_common: dw_dma_get_status: xrun detected
[...]

PR #8226, opened 2023-09-19
dai-zephyr: Do not skip dma_reload() when sink if full

serhiy-katsyuba-intel:
If sink is full and so no bytes were copied, reload DMA anyway as otherwise xrun may happen if reload is skipped for a long time.


serhiy-katsyuba-intel:
Fix for https://github.com/thesofproject/sof/issues/8224.

PR #8228, opened 2023-09-19
ipc4: handler: Add fallthrough comment to prevent compilation error

ranj063:
Add the fallthrough comment to prevent the error seen with shared library build:
"handler.c:312:20: error: this statement may fall through [-Werror=implicit-fallthrough=]"

### DIFF
--- a/src/ipc/ipc4/handler.c
+++ b/src/ipc/ipc4/handler.c
@@ -311,6 +311,7 @@ int ipc4_pipeline_prepare(struct ipc_comp_dev *ppl_icd, uint32_t cmd)
 	case SOF_IPC4_PIPELINE_STATE_EOS:
 		if (status != COMP_STATE_ACTIVE)
 			return IPC4_INVALID_REQUEST;
+		/* fallthrough */
 	case SOF_IPC4_PIPELINE_STATE_SAVED:
 	case SOF_IPC4_PIPELINE_STATE_ERROR_STOP:
 	default:


btian1 (approved):
which build will show this error? I did not see it in my local build.

src/ipc/ipc4/handler.c
```diff
@@ -311,6 +311,7 @@ int ipc4_pipeline_prepare(struct ipc_comp_dev *ppl_icd, uint32_t cmd)
 	case SOF_IPC4_PIPELINE_STATE_EOS:
 		if (status != COMP_STATE_ACTIVE)
 			return IPC4_INVALID_REQUEST;
+		/* fallthrough */
```
lyakh, line 314: `COMPILER_FALLTHROUGH`

ranj063: @lyakh compiler_fallthrough doesn't seem to help with the host build with ipc4. What am I missing?

lyakh: Ok, looking at the definition: https://github.com/thesofproject/sof/blob/992dd676aebc69a97defb7d5361e853919b57d14/xtos/include/sof/compiler_attributes.h#L45-L54 - the only possibility I see for that not to work for you is if in your case the `__attribute__((fallthrough))` definition is used, but that attribute isn't known to your compiler and it's ignoring an unknown attribute. Can you try `-Wignored-attributes` (note, that when building with sparse enabled, the `-Wno-attributes` option is used, which suppresses such warnings)

ranj063:
pulled into #8232

src/ipc/ipc4/handler.c
marc-hb: Zephyr has a universal "fallthrough" that we could copy/re-use https://github.com/thesofproject/sof/pull/8383#discussion_r1373810500

cc: @singalsu

PR #8237, opened 2023-09-20
S32 support for AEC

cujomalainey:


lgirdwood (commented):

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -497,6 +513,109 @@ static int google_rtc_audio_processing_trigger(struct comp_dev *dev, int cmd)
 	return comp_set_state(dev, cmd);
 }
 
+#if CONFIG_FORMAT_S16LE
+static void render_s16(struct google_rtc_audio_processing_comp_data *cd)
+{
+	struct comp_buffer *buffer_c = buffer_acquire(cd->aec_reference);
```
singalsu, line 519: Buffer_acquire() removal applies also for this usage, right?

singalsu (approved):

src/audio/google/google_rtc_audio_processing.c
lgirdwood: yep, we can remove after if this merges first.

lgirdwood (commented):
LGTM. @andrula-song pls also review.

lgirdwood (approved):
@cujomalainey good to merge. Some conflicts though.

cujomalainey:
> @cujomalainey good to merge. Some conflicts though.

Negative, tests reported broken audio, working on debugging, just had to fight my tooling super hard this week

btian1 (commented):
any performance difference with float api?

cujomalainey:
> any performance difference with float api?

From what I understand, none, since the s16 is essentially a frontend for the f32 api

cujomalainey:
This is blocked by #8258. The AEC xruns the pipe on ll

marcinszkudlinski (changes requested):
Please do not merge, i'll include those changes in AEC DP processing PR

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -616,6 +616,109 @@ static void render_s16(struct google_rtc_audio_processing_comp_data *cd)
...
+
+	while (num_frames_remaining) {
+		nmax = audio_stream_frames_without_wrap(&mic_buf->stream, src);
+		n = MIN(num_frames_remaining, nmax);
+		nmax = audio_stream_frames_without_wrap(&output_buf->stream, dst);
+		n = MIN(n, nmax);
+		for (i = 0; i < n; i++) {
+			j = cd->raw_mic_buffer_frame_index * cd->num_capture_channels;
+			for (channel = 0; channel < cd->num_capture_channels; channel++) {
+				cd->raw_mic_buffer[j] = Q_CONVERT_QTOF(src[channel], 31);
```
btian1, line 685: those Q_CONVERT_FLOAT(QTOF) macros normally used with fixed numbers, if you use here, you will have very negative performance impact, this is not proposed to use this macro, please find other way to convert to fix point data.

cujomalainey: If there is a problem with the standard macros then they should be fixed or an alternative provided. The problem should not be addressed here if there is no other alternative currently.

cujomalainey:
Test ping @thesofproject/google

cujomalainey:
Sorry one more test, debugging team configs @thesofproject/google

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -424,15 +429,23 @@ static struct comp_dev *google_rtc_audio_processing_create(
 		goto fail;
 	bzero(cd->raw_mic_buffer, cd->num_frames * cd->num_capture_channels * sizeof(cd->raw_mic_buffer[0]));
 	cd->raw_mic_buffer_frame_index = 0;
+	for (channel = 0; channel < cd->num_capture_channels; channel++)
+		cd->raw_mic_buffer_ptrs[channel] = &cd->raw_mic_buffer[channel * cd->num_frames];
```
andyross, line 433: Looks like there's a `!aec_reference_buffer` early exit case just below this, probably best to put this after that out of general "do the cheap checks first" hygiene.  It also allows the compiler (maybe) or you to roll this into the other num_capture_channels loop you're adding below.  I don't think I see any other branching here that would prevent that, though it's hard to tell in a review UI.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -424,15 +429,23 @@ static struct comp_dev *google_rtc_audio_processing_create(
...
 		0, SOF_MEM_CAPS_RAM,
 		cd->num_frames * sizeof(cd->aec_reference_buffer[0]) *
 		cd->num_aec_reference_channels);
 	if (!cd->aec_reference_buffer)
 		goto fail;
-	bzero(cd->aec_reference_buffer, cd->num_frames * cd->num_aec_reference_channels * sizeof(cd->aec_reference_buffer[0]));
+	bzero(cd->aec_reference_buffer,
+	      cd->num_frames *
+	      cd->num_aec_reference_channels *
+	      sizeof(cd->aec_reference_buffer[0]));
```
andyross, line 444: Style nitpick: this is a whitespace-only change, and it faked me out because I thought for a second because I was sure that this was changing the bzero() region without updating the size of the allocation.

IMHO the whole "write the size expression in the alloc and bzero steps" pattern is fragile.  I guess my aesthetics would say that if we're going to change the source code bytes that compute this number, we should fix it so the size is computed exactly once and used as an immutable value in both the alloc/clear.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -441,6 +454,9 @@ static struct comp_dev *google_rtc_audio_processing_create(
 		goto fail;
 	bzero(cd->output_buffer, cd->num_frames * sizeof(cd->output_buffer[0]));
 	cd->output_buffer_frame_index = 0;
+	for (channel = 0; channel < cd->num_capture_channels; channel++)
+		cd->aec_reference_buffer_ptrs[channel] =
+			&cd->aec_reference_buffer[channel * cd->num_frames];
```
andyross, line 459: Confused: this is looping over the same array as the num_aec_reference_channels loop above and initializing the same indexes to the same values.   Looks like semantically it's identical to looping once over the maximum of the two counts?  Or maybe it's supposed to be initializing a different array?

src/audio/google/google_rtc_audio_processing.c
andyross: Yeah, unless I'm looking at the wrong code, it looks to me like this macro is a little pessimal, dividing (!) a float by a constructed 64 bit (!!) quantity (so at runtime this is going to emit a double precision division instead of the single precision multiply we want).  From format.h:

```
#define Q_CONVERT_QTOF(x, ny) ((float)(x) / ((int64_t)1 << (ny)))
```

Swapping that to something like this should generate better code for sure, though edge cases at the boundaries are always hard to reason about: `#define S32_TO_F(x)  ((x) * (1.0f / BIT(31)))`

src/audio/google/google_rtc_audio_processing.c
cujomalainey: very likely a copy paste error

cujomalainey:
> Notes and nitpicks. I guess one thing to note somewhere is that this is going to be highly dependent on platforms with a working FPU/HiFi engine and attendant toolchain and OS.

I don't think this makes a difference since this is essentially a no-op as the conversion to FP was already done inside the lib, this just makes its easily handed for multiple formats outside.

cujomalainey:
No idea why github won't let me reply to https://github.com/thesofproject/sof/pull/8237#discussion_r1383779719 but for some reason there is no reply box.

this is a question for @ShriramShastry and @singalsu

marcinszkudlinski:
@andyross @cujomalainey @singalsu  @nashif 
If its true that Zephyr does not save HiFi registers in preemption, it is a huge gap and must be fixed, DP processing is designed as a set of threads with deadlines, preempting each other (and being preempted by LL tasks)

marcinszkudlinski:
As to slow float->int conversion, the answer is:
```
int16_t CONVERT(float data)
{
	const xtfloat ratio = 2 << 14;
	xtfloat x0 = data;
	xtfloat x1;
	int16_t x;

	x1 = XT_MUL_S(x0, ratio);
	x = XT_TRUNC_S(x1, 0);

	return x;
}
```

converting of 10ms of 4channel data from float to int16 took 1.1ms (!!!!) using Q_CONVERT_FLOAT, 120us using CONVERT as above
Q_CONVERT_QTOF, however, is quite fast, conversion from 10ms/4chanels fom int16 to float takes 140ms

andyross:
Regarding HiFi registers: I'm guessing @dcpleung might have some WIP on saving the SIMD registers, if not I can take a crack at it.  It's not particularly difficult, though it does have some stack/performance tradeoffs to worry about (it's a ton of data to push on the stacks, but then stacks are cached where e.g. the thread struct can't be).  And IIRC (i.e. I might be wrong, would have to look up) Xtensa doesn't have a way to lazy-evaluate (e.g. trap on use) the context switch, we'd likely pay the cost for every interrupt/switch.

And as far as how to implement it: my recommendation would be to stay away from the explicit TIE macros as they defeat the optimizer (and obviously aren't portable anyway).  Checking quickly, the loop below (very roughly what the code in question is doing) already optimizes to a pleasingly unrolled loop over the scalar FPU with xcc 2021.6 on TGL.  With xt-clang and the ACE config, it actually emits some pretty scarily optimized (but quite fast looking) well-bundled HiFi SIMD loops.  The TIE macros are definitely in third place by my accounting (though still ahead of gcc: the Zephyr SDK for TGL won't emit FPU code at all and calls into libgcc -- pretty sure FPU support is supposed to work, might be a compiler build goof).

andyross:
> Checking quickly, the loop below [...]

Oops, forgot loop:

```c
void f_to_s32(float *in, int *out, int n)
{
    for (int i = 0; i < n; i++)
        out[i] = in[i] * (1.0f / INT_MAX);
}
```

andyross:
> And IIRC (i.e. I might be wrong, would have to look up) Xtensa doesn't have a way to lazy-evaluate (e.g. trap on use) the context switch

I looked it up.  It does.  It gets complicated with SMP in the presence of SCHED_PIN_ONLY=n because you need to shoot down the state on another core when threads migrate, but that's not a SOF configuration so we can defer optimizing that case.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -498,6 +513,212 @@ static int google_rtc_audio_processing_trigger(struct comp_dev *dev, int cmd)
...
+			if (cd->raw_mic_buffer_frame_index == cd->num_frames) {
+				GoogleRtcAudioProcessingProcessCapture_float32(cd->state,
+									       (const float **)cd->raw_mic_buffer_ptrs,
+									       cd->output_buffer_ptrs);
+				cd->output_buffer_frame_index = 0;
+				cd->raw_mic_buffer_frame_index = 0;
+			}
+
+			src += audio_stream_get_channels(&mic_buf->stream);
+			dst += audio_stream_get_channels(&output_buf->stream);
```
andrula-song, line 603: same here.

ShriramShastry:
> No idea why github won't let me reply to [#8237 (comment)](https://github.com/thesofproject/sof/pull/8237#discussion_r1383779719) but for some reason there is no reply box.
> 
> this is a question for @ShriramShastry and @singalsu


> Q_CONVERT_QTOF

Hi @cujomalainey ,

I think you can use `'Q_CONVERT_QTOF'` without any problems for the time being. In `cmocka` wrappers, `Q_CONVERT_QTOF` is frequently utilized, and there is no appreciable loss of accuracy in terms of Total Harmonic Distortion or Unit in last place.



If performance is an issue, I'm not sure what the expectations are—what the cycle and benefit are expected to be, how long they will last, and so on. 

Yes, it is true that the `#define Q_CONVERT_QTOF(x, ny) ((float)(x) / ((int64_t)1 << (ny)))` division operator may be an argument. In that case, the implementation might be made quickly for any compiler by introducing the inverse `#define Q_CONVERT_QTOF(x, ny) ((float)(x) * inv((int64_t)1 << (ny)))` and employing a leading-to-zeros technique with a LUT to prevent loops.

Thank you!

marcinszkudlinski:
I'm finishing with DP venison of AEC and the loop looks like:

```
	for (int i = 0; i < cd->num_frames; i++) {
		for (channel = 0; channel < cd->num_capture_channels; channel++)
			cd->process_buffer_ptrs[channel][i] =
					Q_CONVERT_QTOF(src[channel], 15);

		src += cd->num_capture_channels;
		if (src >= src_buf_end)
			src = src_buf_start;
	}
```
execution time for the loop above - for 10ms 16bit 4 channels - 140us


**please do not merge this commit - once DP version is stable - hopefully bvvery soon - i'll put a new PR for this**

marcinszkudlinski:
I';m closing this PR, 
all changes for 32bit are here: https://github.com/thesofproject/sof/pull/8469

Issue #8238, opened 2023-09-21
[BUG] ramp time extended by 0.03s after capture gain curve_duration change


keqiaozhang:
**Describe the bug**
Observed this issue on all IPC4-nocodec(single core) platforms.  it caused alsabat test failures.
The ramp time extended by 0.03s after capture gain curve_duration change(https://github.com/thesofproject/sof/commit/d0d74a477f64a83654bccb3905175584d05d9257):
![image](https://github.com/thesofproject/sof/assets/22468760/ea78614a-071c-4aaa-b658-def3dabd1d06)

**To Reproduce**
check the alsabat wav file

**Reproduction Rate**
100%

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel: {[cffcac365eac](https://github.com/thesofproject/linux/commit/cffcac365eac0d60dc38959cadaa5078bc924b8e)}
    * SOF: {[capture gain curve_duration change](https://github.com/thesofproject/sof/commit/d0d74a477f64a83654bccb3905175584d05d9257)}
2) Name of the topology file
    * Topology: {nocodec tplg}
3) Name of the platform(s) on which the bug is observed.
    * Platform: {NOCODEC-IPC4 platforms}

marc-hb:
The commit message says "This change increases the ramp duration from 20 ms to 50 ms"

So it's not a bug, it's a feature? :-)

keqiaozhang:
> So it's not a bug, it's a feature? :-)

Not sure about this, but alsabat cannot survive from such long ramp time.

singalsu:
Ouch, but there's justification for this change. 1) capture artefacts mitigation accidentally dropped from tplg2 vs. tplg1 2) we can avoid a big MCPS peak with this lower rate ramp.

Is there a way to ask alsabat to ignore a bit long audio from beginning?

singalsu:
If not possible to change alsabat we can have a test topology with short ramp?

singalsu:
I could not find from alsabat command line options to ignore audio from very beginning of test. I don't like at all to change something because of limitation in our test procedure but this might help https://github.com/thesofproject/sof/pull/8242 .

lgirdwood:
@singalsu ack, the topology used for the test can have its ramp shortened, providing its not a end user topology. i.e. testing only topology.

marc-hb:
Let's wait for validation in the next daily test run.

[...]

PR #8242, opened 2023-09-21
Tools: Topology2: Remove 50 ms curve duration set from nocodec

singalsu:
We need to care about audio user experience and peak MCPS usage in production topologies.

The alsabat test is disturbed by longer ramp so it can be removed from nocodec topologies since the codec is never used by end users and also the peak MCPS mitigation is not relevant for a test topology, as long as higher MCPS is not triggering error reports. The curve duration becomes without explicit set the default 20 ms.

Fixes: #8238
Fixes: d0d74a477f64a83654bccb3905175584d05d9257
       ("Tools: Topology2: Change in capture gain
       curve_duration to 50 m")

### DIFF
--- a/tools/topology/topology2/cavs-nocodec-multicore.conf
+++ b/tools/topology/topology2/cavs-nocodec-multicore.conf
@@ -335,7 +335,6 @@ IncludeByKey.PASSTHROUGH {
 							}
 						}
 						Object.Widget.gain.1 {
-							curve_duration 500000
 							num_input_audio_formats 2
 							num_output_audio_formats 2
 							Object.Base.audio_format.1 {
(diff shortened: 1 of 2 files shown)


btian1 (commented):
if CI can't make it pass with 50ms, then we have to revert, however, I would suggest tile and message use below:

Revert "Tools: Topology2: Change in capture gain curve_duration to 50 ms"

This reverts commit d0d74a477f64a83654bccb3905175584d05d9257.

singalsu:
> if CI can't make it pass with 50ms, then we have to revert, however, I would suggest tile and message use below:
> 
> Revert "Tools: Topology2: Change in capture gain curve_duration to 50 ms"
> 
> This reverts commit [d0d74a4](https://github.com/thesofproject/sof/commit/d0d74a477f64a83654bccb3905175584d05d9257).

No, it's not fully reverting the patch. Only the nocodec topologies those are tested in CI are changed back.

And really need to start to develop fix for alsabat. We can't continue changing default behavior to follow shortcomings of test tool.

kv2019i (approved):

lgirdwood (approved):

Issue #6917, opened 2023-01-05
[FEATURE] pass vendor specific FW signing options to west sign


kv2019i:
**Is your feature request related to a problem? Please describe.**
On Intel platforms, we need to sign production releases with west sign options "-i" and  "-p". Currently SOF's xtensa-build-zephyr.py does not allow passing these options.

**Describe the solution you'd like**
Option to pass a list of options that suits multivendor usage with "west sign".

@marc-hb May 2022 EDIT: this almost complete. It has NOT been implemented by "passing a list of options to xtensa-build-zephyr.py". It's using `west config` for `west sign` instead.

kv2019i:
FYI @abonislawski @marc-hb @lgirdwood @juimonen

lgirdwood:
@nashif fyi

marc-hb:
So as I suspected, https://github.com/zephyrproject-rtos/zephyr/pull/52942 is trying to solve this problem (among others). I just finished a (detailed) review of it.

If upstream Zephyr solves this problem with some new configuration file or something and we add some hack in `xtensa-build-zephyr.py` at _the same time_ then we're very likely to end-up in some forward/backward compatibility nightmare with upstream.

So temporarily for this 2.4 release would it be possible to (re-)run `west sign` manually with the modified `rimage` parameters? AFAIK production releases are NOT automated anyway. This will overwrite the `.ri` file in the `build-tgl` directory, then it only needs to be copied to `build-sof-staging/`.

After running `xtensa-build-zephyr.py`, scrolling up 1 or 2 screens shows the exact `west sign` command it has run.

kv2019i:
@marc-hb Let's leave the 2.4 release aside and consider just the specific question of ability to pass signing options. I think a generic ability to pass FW signing options seems like a generic thing we will need now and future. I don't see Zephyr PR52942 changing this.

If I look up Zephyr docs on west sign:
https://docs.zephyrproject.org/latest/develop/west/sign.html
```
west sign -t your_tool -- ARGS_FOR_YOUR_TOOL
```

... so generic ability to pass extra  options from xtensa-build-zephyr.py to "ARGS_FOR_YOUR_TOOL" seems like a good capability to have as it's something "west sign" tool has.

If signing is merged to build step on Zephyr side, then it would seem the extra options would need to be passed to build step, but otherwise still applicable.

[...]

PR #8243, opened 2023-09-22
copy zephyr based sof image toml files from rimage to sof and split to platform and module toml

aiChaoSONG:
This patch copies Intel IPC4 toml files from rimage to sof repo, and split toml files
to platform related toml and module related toml. 
Also update build scripts. 

cc:
- #8178 
- https://github.com/thesofproject/rimage/pull/179


scripts/xtensa-build-zephyr.py
```diff
@@ -585,8 +582,22 @@ def build_rimage():
 			rimage_build_cmd.append("-v")
 	execute_command(rimage_build_cmd, cwd=west_top)
 
+def prepare_rimage_desc(platform_name, dest_dir):
+	toml_name = platform_name + '.toml'
+	platform_desc = SOF_TOP / "config" / "platform" / toml_name
+	module_desc = SOF_TOP / "config" / "module" / toml_name
```
marc-hb, line 588:
```suggestion
	module_config = SOF_TOP / "config" / "module" / toml_name
```

marc-hb (changes requested):
- In the relevant commit message, please add the exact rimage commit you copied the .toml files from.
- Rename the copied files in a separate commit.

This will make it easy to compare and follow the git history of the copied files across multiple repos.

PS: it is possible to merge the git histories of several repos but that would a bit overkill for so few files.

marc-hb:
Looks like we have a problem with IPC3:

https://github.com/thesofproject/sof/actions/runs/6269606713/job/17026293267?pr=8243
```
FileNotFoundError: [Errno 2] No such file or directory: '/zep_workspace/sof/config/platform/imx8.toml'
```

marc-hb:
The "ninja not found" failures are worrying but unrelated to this PR, I've seen them elsewhere. Re-running the job in the Github interface usually gets rid of them.

aiChaoSONG:
@marc-hb Didn't expect the failures on non-intel platforms, I need to add fallback for non-intel platforms to still read toml file from rimage.
@kv2019i  Or should I also copy imx toml files to sof?

marc-hb:
> Or should I also copy imx toml files to sof?

Yes I was wondering what makes this move IPC4-specific so far - if anything? Couldn't it be for everything Zephyr? Would be much simpler.

BTW there is a LOT of context here and so far this PR links to none. @aiChaoSONG please add a link to at least #7270 and maybe to  #8178 as well

cc: @dbaluta, @andyross

aiChaoSONG:
@marc-hb Ok, I am thinking(and working) to move toml for zephyr base sof, instead of only moving toml for intel IPC4.

marc-hb (commented):
Rather than:
```
config/
├── module
│   └── lnl.toml
└── platform
    └── lnl.toml
```
... could we have this instead?
```
config/
└── lnl
    ├── module.toml
    └── platform.toml
```
Then the script can concatenate `config/$platf/*.toml` without any hardcoded name. More flexible and future proof.

Also, `git log -- config/lnl/` is easier and more logical.

kv2019i:
@marc-hb @aiChaoSONG I'm wondering why not move everyone over. If conceptually rimage is made just a tool with configuration data elsewhere, then all platforms (supported in mailine) move. Limiting to Zephyr doesn't make too much sense as the toml contents is just the same for Zephyr and non-Zephyr.

@thesofproject/amd @thesofproject/mediatek @thesofproject/nxp inputs? This probably doesn't affect you much as long as builds continue to work as before.

aiChaoSONG:
@kv2019i We have deprecated XTOS and IPC3 on sof mainline, so I only move those with zephyr RTOS. As we won't touch xtensa-build-all.sh anymore, can we count on AMD/NXP/MTK to move their own toml and do script update? The main problem for me is that we don't have their xtensa toolchain and could never build their firmware successfully, let along test signing correctness.

kv2019i:
@aiChaoSONG Aa right. We do have "gcc-build-default-platforms" that covers all platforms and if I checked right, only place where the toml path is defined is work/sof/src/arch/xtensa/CMakeLists.txt (in two make rules plus one comment). So if you change that, that should work for all XTOS builds.

But ack. This gets iffier if we want to use a different layout for the toml files in SOF repo. E.g. for IMX targets that support both XTOS and Zephyr still, this gets really complicated.

So maybe in this PR:
 - copy all Zephyr platforms to SOF
 - keep IPC3 only targets intact
 - once ready, remove the IPC4 algorithm sections from rimage repository
        - removing Intel platform definitions is potentially a problem as that would break Zephyr unit tests now (FYI @nashif ), so this can be done only once Zephyr is ready for removal
        - toml files used by both Zephyr/XTOS (like IMX) cannot be removed as that would break XTOS IMX builds
        
FYI @nashif - I think in Zephyr, the platform toml file should be generated from DTS data at build time, but that has to be done first, before we can remove any of the current toml data from the repository.

aiChaoSONG:
@kv2019i  After looking at cmake files, firmware using xtos is signed with cmake steps. As there is only platform related toml contents for XTOS, I could file another PR(or maybe add more commits in this PR), which copies xtos based sof toml files to sof, and update below cmake, which is quite smooth transition.

Talking about toml files shared between Zephyr and XTOS, they are imx8.toml, imx8m.toml and imx8x.toml, they are still shared after copying the rest of toml files (because only intel platform tomls have those module related configuration).

`-c "${PROJECT_SOURCE_DIR}/rimage/config/${fw_name}.toml"` will be changed to `-c "${PROJECT_SOURCE_DIR}/config/${fw_name}/10-platform.toml"`

```
		run_rimage
		COMMAND ${PROJECT_BINARY_DIR}/rimage_ep/build/rimage
			-o sof-${fw_name}.ri
			-c "${PROJECT_SOURCE_DIR}/rimage/config/${fw_name}.toml"
			-s ${MEU_OFFSET}
			-k ${RIMAGE_PRIVATE_KEY}
			-i ${RIMAGE_IMR_TYPE}
			-f ${SOF_MAJOR}.${SOF_MINOR}.${SOF_MICRO}
			-b ${SOF_BUILD}
			-e
			sof-${fw_name}
```

config/lnl/10-platform.toml
```diff
@@ -0,0 +1,57 @@
+version = [3, 0]
```
lgirdwood, line 1: Ehm, not following why we have a 10 prefix in our filenames. I want to #include these via the pre-processor.

scripts/xtensa-build-zephyr.py
```diff
@@ -585,8 +582,32 @@ def build_rimage():
 			rimage_build_cmd.append("-v")
 	execute_command(rimage_build_cmd, cwd=west_top)
 
-
-def rimage_options(platform_dict):
+def concat_rimage_configs(platform_name, dest_dir):
```
lgirdwood, line 585: nak, lets use the pre-processor. We also need to add Kconfig detail into the toml too.

lgirdwood (changes requested):

lgirdwood:
@aiChaoSONG we are going to pre-process our toml files in teh same way we pre-process linker scripts. I would split this PR into 2 PRs.
PR 1 ) just moves the existing tomls files.
PR 2 ) adds pre-processing support to toml. i.e.
```
/* Intel platform.toml.in */
#include <zephyr/meteorlake/memory-defs.h>
// define all memory regions using macros from C header
```
Then this would be included by the product toml.in
```
/* meteorlake toml.in */
#include <platform.toml>
#include <cse.toml>
#if CONFIG_VOLUME 
    #include <volume.toml>
#endif
// and so on
```

marc-hb:
>  The main problem for me is that we don't have their xtensa toolchain and could never build their firmware successfully, 

Our docker container has most toolchains, this is what CI uses everyday:

https://github.com/thesofproject/sof/actions/runs/6271755640/job/17031962306?pr=8243

marc-hb:
> removing Intel platform definitions is potentially a problem as that would break Zephyr unit tests now (FYI @nashif ), so this can be done only once Zephyr is ready for removal

Yes this will break that. BTW this is why `rimage` was moved out of SOF: to make it possible to run Zephyr tests on hardware _independently from sof.git. History in:
- #8178

With hindsight we know that extracting rimage.git from sof.git was a mistake and we'd like to go back and put everything in a single sof.git repo like it was. But this is not happening right now. So this is where we are right now: moving ALL .toml files out of rimage will break Zephyr tests big time.

I think this PR is losing sight of the main objective. The main and most important objective is: _support multiple .toml files_. So this generic feature must be the first PR, _before_ splitting and moving files around. Once the ability and main feature is merged in the main branch(es) and available for everyone, then it will become easier in a SECOND step for every developer, test suite and CI to try and see what breaks when .toml files get sliced and diced.

Now jumping the gun to this second requirement: the location of .toml files. #7270 considered an interesting possibility: moving _only_ the modules / IPC4 specific .toml to sof.git. Why? Simply because locating _modules_ .toml files out of rimage.git is the only thing _actually_ required! This is the only actual problem to solve. Having the _platform_ .toml files in rimage.git never caused any real problem. Then #7270 **assumed** that moving ALL .toml files would be "an easier solution". But now we now this assumption was wrong: it's actually much more complex to move all platform .toml files - and it's NOT a requirement. Moving everything just "sounds cool" like extracting rimage.git sounded cool but it's just a lot of extra work and breakages with zero value.

marc-hb (changes requested):
I've been at last spending some time reading, thinking about the issues here.

   ==>  tl;dr: do NOT implement any multi-toml logic in `xtensa-build-zephyr.py`. It's just the wrong place.

"All problems in computer science can be solved by another level of indirection"
 
"...except for the problem of too many layers of indirection/obfuscation".

`xtensa-build-zephyr.py` is a very good demonstration of the latter.

Every time there are two many layers of indirection, debugging and fixing issues becomes a nightmare because every layer obscures everything below it.

The number of build indirection layers involved when using `xtensa-build-zephyr.py` is insane: https://docs.zephyrproject.org/latest/develop/west/sign.html#rimage
This page does not even list `xtensa-build-zephyr.py`, Github Actions and Docker!


Debugging build issues is always like "peeling an onion": trying to reproduce with fewer and fewer layers until you can finally see where the problem is. I know because I'm one of the very rare persons fixing the build (just last week: https://github.com/zephyrproject-rtos/zephyr/issues/63003, https://github.com/thesofproject/sof/issues/8250, https://github.com/thesofproject/sof/pull/8253, ...)

The more logic there is a layer of indirection, the harder it is to peel. For that reason,  everyone who has ever worked on `xtensa-build-zephyr.py` always agreed it **should be kept as simple and thin as possible**.

As a matter of fact I just spent a number of weeks moving rimage-related logic OUT of `xtensa-build-zephyr.py`:
- #6917

Ideally, configuring rimage with multiple configuration files belongs to rimage (or maybe even better: in https://github.com/thesofproject/tomlc99) So I don't understand why you abandoned https://github.com/thesofproject/rimage/pull/179.

In 179 @lgirdwood requested C pre-processing of .toml files. Then you closed 179 and replaced it with this PR which still does NOT use the C pre-processor...? I'm not following. Instead you moved file concatenation and rimage-related logic to xtensa-build-zephyr.py = as far away from rimage as possible :-(

If multi-toml  can't be done directly in rimage for whatever reason, then the second best choice is a place that is much closer to rimage: either Zephyr's CMake or `west sign`. 

BTW using the C pre-processor feels overly complex and somewhat overkill to me but if we go that route then you definitely want that to be either part of or below the build system The build system knows about the C pre-processor unlike `xtensa-build-zephyr.py` which is quite a distance away from it.

lgirdwood:
@aiChaoSONG pls work with @lyakh on pre-processor for toml modules

marc-hb:
I'm making very good progress moving the rimage git repo back into sof:
- #8178

config/imx8.toml
```diff
@@ -0,0 +1,20 @@
+version = [1, 0]  # use simple file write
```
btian1: imx platform also support zephyr build, right?

kv2019i:
Converted back to draft. Please mark again for review when/if you update this after the recent rimage change. If we go for a different PR (to introduce preprocessor for toml), then this can be closed.

aiChaoSONG:
close, rimage movement is done by Marc,  I think Guennadi will do toml preprocess. this is not the right direction to fix.

PR #8245, opened 2023-09-22
audio: perf: replace performance measure clock to dsp real clock

btian1:
Previously, sof performance measurement use a platform clock from zephyr, it is based on 38.4mhz, now with this patch, sof can use dsp real clock to get exactly cycles, cycle calculation and convert can be ignored.


xtos/include/sof/lib/perf_cnt.h
```diff
@@ -58,7 +76,7 @@ struct perf_cnt_data {
 	#ifdef CONFIG_TIMING_FUNCTIONS
 		#define perf_cnt_get_cpu_ts arch_timing_counter_get
 	#else
-		#define perf_cnt_get_cpu_ts sof_cycle_get_64
+		#define perf_cnt_get_cpu_ts arch_timing_counter_get_DSP
 	#endif	/* CONFIG_TIMING_FUNCTIONS */
```
kv2019i, line 80: @btian1 I think would be easier to enable CONFIG_TIMING_FUNCTIONS=y and then arch_timing_counter_get() will return CCOUNT on xtensa. No need to have hw specific code on SOF side.

tmleman:
I have two questions:
Why it's in xtos header if its for zephyr?
How and where will it be used? It may turn out that it will be necessary to add a lock for D0i3 due to the clock gating enabled with it.

kv2019i (commented):
Hmm, only concern I have is that this is a subtle change of semantics. The old code (with platform clock) allows to measure DSP load versus real time and this is accurate even if the DSP frequency is changed during the measurement window. If we move to measure with CCOUNT, the number have to understood differently (you can't calculate DSP load % in the same way anyway, unless you have data about the DSP frequency changes during the use-case).

I think both are useful. Former is useful for system debug to understand whether the DSP is behaving as expected, while latter is more useful to measure algorithm performance.

We can't have both enabled, and it's true the DSP clock changes are only done at very specific times with current SOF code (at stream start/stop, not during steady state of audio streaming), so probably this has little impact in practise. Just need everybody who is doing system debug to be aware of this change so they know what they are looking at.

@marcinszkudlinski any thoughts? shall we go with CCOUNT measurements?

btian1:
> I have two questions: 
Why it's in xtos header if its for zephyr? 
we can move it, this is legacy part, when change to zephyr, this file is not moved.

How and where will it be used? It may turn out that it will be necessary to add a lock for D0i3 due to the clock gating enabled with it.
please check again, only enable one config is fine.

tmleman:
@btian1 the change itself looks good. When are these measurements used? You build FW with this overlay and then run normal tests?

kv2019i (changes requested):
I'd prefer to use the option for all builds -> CONFIG_TIMING_FUNCTIONS=y

It seems Zephry has separate Kconfigs more for costly usages of timing functrions and baseline timing function support is lightweight.

btian1:
> @btian1 the change itself looks good. When are these measurements used? You build FW with this overlay and then run normal tests?

this is for sof mcps test, add -o app/perf_overlay.conf can get a performance build, later we can use this to get CPC data.

app/perf_overlay.conf
```diff
@@ -1,2 +1,3 @@
 CONFIG_PERFORMANCE_COUNTERS=y
 CONFIG_SYS_HEAP_RUNTIME_STATS=y
+CONFIG_TIMING_FUNCTIONS=y
```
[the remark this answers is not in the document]
btian1: I searched sof code repo, currently, only perf_cnt.h use this config, so put for perf only build should be fine, and I noticed there is zephyr code use this config as well, if defined to sof, not sure whether this cause other issue or not.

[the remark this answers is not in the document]
kv2019i: Ack @btian1.  Can you explain this in git commit that this only impacts behaviour of perf_cnt_get_cpu_ts() in SOF. When CONFIG_TIMING_FUNCTION=y, arch_timing_counter_get() is used to sample CPU time. On Xtensa, this maps to CCOUNT.

btian1:
performance script need change as well, let me change perf script and submit together.

tmleman:
> this is for sof mcps test, add -o app/perf_overlay.conf can get a performance build, later we can use this to get CPC data.

How this will address the fact that `ccount` frequency can change in runtime?

btian1:
> > this is for sof mcps test, add -o app/perf_overlay.conf can get a performance build, later we can use this to get CPC data.
> 
> How this will work in situation when `ccount` frequency will change in runtime?
> 
> Look at #8019

this is not for clock control, this is for mcps calculation with performance build, to enable performance build, you simply add -o app/perf_overlay.conf to build command line is fine.

tmleman:
> this is not for clock control, this is for mcps calculation with performance build, to enable performance build, you simply add -o app/perf_overlay.conf to build command line is fine.

Ok, but doesn't the result you get depend on the clock frequency?

btian1:
> > this is not for clock control, this is for mcps calculation with performance build, to enable performance build, you simply add -o app/perf_overlay.conf to build command line is fine.
> 
> Ok, but doesn't the result you get depend on the clock frequency?

no dependent anymore, please check:https://github.com/thesofproject/sof-test/pull/1114, they will work together to get CPC data in CI build and running.

xtos/include/sof/lib/perf_cnt.h
pblaszko: @kv2019i @btian1 "DSP clock changes are only done at very specific times with current SOF code (at stream start/stop, not during steady state of audio streaming)"

In reference FW, IPC task and DP processing task can preempt each other (IPC task has budget, for some time has high priority, then switch to low priority). So potentially DP task can start with lower DSP clock, then it can be preempted by IPC task which starts another pipeline and switch DSP clock. Then DP task finish chunk processing with changed clock.
I'm not sure how is it working in SOF?

Moreover, we still have plan (don't we?) to implement DP modules which switch DSP clock in runtime, like WoV or AEC.

Also, in the past platforms, FW had to do HW-specific implementation that switches clocks in runtime, e.g. every time when accessing IMR. Fortunately current platforms do not require this, but who knows...

So in my opinion keeping WLCK as a reference is a safer solution.

btian1: 1. this CONFIG_TIMING_FUNCTIONS only enabled in performance build(with -o app/perf_overlay.conf option).
2. clock measure only happens on module start and module end in function comp_copy(https://github.com/thesofproject/sof/blob/main/src/audio/component.c)
    is it possible clock will switch inside module?

pblaszko: OK, it it is only enabled in performance build, then I'm completely fine with this.

kv2019i (approved):
Thanks @btian1 , looks good now!

tmleman (approved):

PR #8254, opened 2023-09-22
[SKIP SOF-TEST] .github: build-windows: choco install ninja

marc-hb:
Ninja seems to have just mysteriously disappeared from https://github.com/actions/runner-images/blob/win22/20230918.1/images/win/Windows2022-Readme.md See #8250 for more details.

Let's try to install it explicitly.

PR #8255, opened 2023-09-23
app/prj.conf: disable PICOLIBC with CONFIG_MINIMAL_LIBC=y

marc-hb:
Starting from Zephyr commit f0daf904bb02, CONFIG_PICOLIBC is on by default.

PICOLIBC does not seem compatible with sparse yet: https://github.com/zephyrproject-rtos/zephyr/issues/63003

Even if it were compatible with sparse, it seems like a pretty big change that we should not immediately and blindly accept.

### DIFF
--- a/app/prj.conf
+++ b/app/prj.conf
@@ -14,6 +14,11 @@ CONFIG_OUTPUT_DISASSEMBLY=y
 
 CONFIG_HAVE_AGENT=n
 
+# PICOLIBC does not seem compatible with the "sparse" static analyzer
+# yet, see https://github.com/zephyrproject-rtos/zephyr/issues/63003
+# Plus the paint on it is still wet.
+CONFIG_MINIMAL_LIBC=y
+
 CONFIG_LOG=y
 CONFIG_LOG_PRINTK=y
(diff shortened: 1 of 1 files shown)


marc-hb:
The fix for the unrelated build-windows failures is in:
- #8254

lyakh (approved):

tmleman (approved):

kv2019i:
Excellent debug @marc-hb on the sparse problem. On one hand, now would be a good time to change to picolibc (as we are very early in 2.8 cycle), but agreed, we need to solve the sparse issue first (and not block other Zephyr updated due to this).

kv2019i:
@wszypelt ok to merge despite failures? this is gating Zephyr updates now and this PR alone does not have any change to build  -- it explicitly sets an option that newer Zephyr versions would change to something else by default.

kv2019i (approved):

wszypelt:
@kv2019i We had a problem with the network on the machines, the LAB team has already fixed it, correct results should be available within an hour

iuliana-prodan (approved):

tmleman:
@kv2019i I think we can merge it

PR #8256, opened 2023-09-25
Zephyr: Fix path to google_hotword_detect.c

singalsu:
The file is found from subdirectory src/audio/google. Without fix if CONFIG_COMP_GOOGLE_HOTWORD_DETECT is enabled build fails:

CMake Error at sof/zephyr/cmake/modules/extensions.cmake:424 (add_library):
Cannot find source file:
/src/audio/google_hotword_detect.c


cujomalainey (approved):
Thanks for spotting this, it clearly shows also that I don't have the stub build configured properly. I will have to go ahead and fix that.

marc-hb:
> ... which is quite a long time ago.

I don't know anything about this code but I do know that "if it's not tested then it does not work" :-)

PR #8246, opened 2023-09-22
topology2: Add 4ch audio formats for gain-capture pipeline

RanderWang:
Add support in gain-capture pipeline to support 4ch DMIC stream.

PR #8258, opened 2023-09-25
Audio: Convert Google RTC audio processing to module API

singalsu:
Fist commit is module adapter conversion

Second commit is improvement to mockup to mix and saturate all available matching channels.

Third commit adds build and load to testbench.

Related to, no more dependent of:

- [ ] https://github.com/thesofproject/sof/pull/8237


lgirdwood (commented):
@singalsu good stuff !
@johnylin76 @cujomalainey fyi - pls review.

singalsu:
@cujomalainey This is the first version that works with created test topology in sof-nocodec style, sharing it soon as well. I think it's best to get #8237 merged first, then I'll rebase this.

singalsu:
Fixed conflict with 7112246d4fef43e3a5caffbbd0c1d58f14c02303 ("google_rtc_audio_processing: remove buffer_acquire")

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -294,85 +277,121 @@ static int google_rtc_audio_processing_cmd_set_data(
...
+		return ret;
+
+	/* Accept the new blob immediately so that userspace can write
+	 * the control in quick succession without error.
+	 * This ensures the last successful control write from userspace
+	 * before prepare/copy is applied.
+	 * The config blob is not referenced after reconfigure() returns
+	 * so it is safe to call comp_get_data_blob here which frees the
+	 * old blob. This assumes cmd() and prepare()/copy() cannot run
+	 * concurrently which is the case when there is no preemption.
```
lyakh, line 337: in principle `.copy()` can run concurrently with IPCs, in which case the timer interrupt can interrupt IPC processing, then the scheduler will choose between the IPC thread and the LL scheduler thread and will pick up the latter, since it has a higher priority. But if the component has only one sink and one source, mostly it's only the "stop" IPC that can arrive while the stream is already running, or can some configuration be sent to a running stream too?

RanderWang:
@singalsu  How do you test it ? it is used by chrome so it is better to sync with chrome team.

singalsu:
> @singalsu How do you test it ? it is used by chrome so it is better to sync with chrome team.

I tested this in UPX nocodec mode with new very topology in PR #8266. Testing with Chrome sounds like a good idea but I can't do it myself. I'll see if I can get help from them to test this.

yongzhi1:
Sanity checked the patch along with https://github.com/thesofproject/linux/pull/4550 and [#8101](https://github.com/thesofproject/sof/pull/8101), [#8246](https://github.com/thesofproject/sof/pull/8246), [#8204](https://github.com/thesofproject/sof/pull/8204), no regression found with AEC mock module.

src/audio/google/google_rtc_audio_processing.c
singalsu: I preserved this part as close to the original so this should not change the operation. @lkoenig What do you think of this?

lkoenig: Configurations can be sent when a stream is running.

singalsu: @lyakh @lkoenig I think it depends on the (unknown) internals of the processing whether new config data blob during copy() is a hazard  or not. The open-source mock-up version ignores the blob content totally.

The reconfigure set here should ensure that the next copy() correctly re-initializes for the blob. Please let me know if this needs to be addressed now, I could propose a fix. Or (preferably) let @thesofproject/google  address this in successive PR.

lkoenig: The data should be whole and not altered during the call of `GoogleRtcAudioProcessingReconfigure` and `GoogleRtcAudioProcessingParseSofConfigMessage`
As of today we do not have concurrent access protection enabled in the `GoogleRtcAudioProcessing`.

lyakh: @singalsu at least I would note in the comment, that a race is possible and should be fixed or at least further investigated

lgirdwood (approved):

cujomalainey:
Apologies for the delay here, I will look on monday, been a crazy week

lgirdwood:
@singalsu still draft ?

singalsu:
> @singalsu still draft ?

I think it's best to wait for merge of #8237 that is a larger change.

src/audio/google/google_rtc_audio_processing_mock.c
```diff
@@ -145,11 +147,17 @@ int GoogleRtcAudioProcessingProcessCapture_int16(GoogleRtcAudioProcessingState *
 	int16_t *ref = state->aec_reference;
 	int16_t *mic = (int16_t *) src;
 	int16_t *out = dest;
-	int n;
+	int channels;
+	int n, m;
+
+	channels = MIN(state->num_capture_channels, state->num_aec_reference_channels);
+	channels = MIN(state->num_output_channels, channels);
```
ranj063, line 154: @singalsu this works for the 2ch case where the PDM 0/1 are configured on the board but if only PDM 2/3 are connected, this won't work correctly right?

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -8,6 +8,7 @@
 #include <ipc/stream.h>
 #include <ipc/topology.h>
 #include <ipc4/aec.h>
+#include <sof/audio/module_adapter/module/generic.h>
```
btian1, line 11: if this patch have multiple purpose(from commit message), I would suggest split out 1 or 2 patches.

src/audio/google/google_rtc_audio_processing_mock.c
```diff
@@ -145,11 +147,17 @@ int GoogleRtcAudioProcessingProcessCapture_int16(GoogleRtcAudioProcessingState *
...
 
 	memset(dest, 0, sizeof(int16_t) * state->num_output_channels * state->num_frames);
 	for (n = 0; n < state->num_frames; ++n) {
-		*out = *mic + *ref;
+		for (m = 0; m < channels; m++)
```
btian1, line 158: is it a bug? since previous code does not have channels loop.

singalsu: The original code mixed only to first channel, with this loop we can mix to other output channels. But according to Ranjani's comment we should decide what to do with non-matching channels counts.

src/audio/google/google_rtc_audio_processing.c
cujomalainey: yes please

src/audio/google/google_rtc_audio_processing_mock.c
cujomalainey: in an ideal world, we would reduce our channels to only the ones we care about so we don't need these silly configs and only have real channels in the data.

src/arch/host/configs/library_defconfig
```diff
@@ -16,6 +17,7 @@ CONFIG_COMP_VOLUME=y
 CONFIG_COMP_VOLUME_LINEAR_RAMP=y
 CONFIG_COMP_VOLUME_WINDOWS_FADE=y
 CONFIG_DEBUG_MEMORY_USAGE_SCAN=n
+CONFIG_GOOGLE_RTC_AUDIO_PROCESSING_MOCK=y
```
cujomalainey, line 20: you can just enable the stub config

singalsu: Thanks, helps with two other conversions also, done!

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -484,44 +512,55 @@ static void google_rtc_audio_processing_free(struct comp_dev *dev)
...
-{
-	comp_dbg(dev, "google_rtc_audio_processing_trigger(): %d", cmd);
-
-	return comp_set_state(dev, cmd);
+	return 0;
 }
 
-static int google_rtc_audio_processing_prepare(struct comp_dev *dev)
+static int google_rtc_audio_processing_prepare(struct processing_module *mod,
+					       struct sof_source __sparse_cache **sources,
```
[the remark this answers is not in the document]
singalsu: Good catch, thanks!

src/audio/google/google_rtc_audio_processing_mock.c
singalsu: I changed it to version where every output channel has a mix of mic and reference. Hope it's good for testing now.

src/arch/host/configs/library_defconfig
```diff
@@ -3,6 +3,7 @@ CONFIG_COMP_CROSSOVER=y
 CONFIG_COMP_DCBLOCK=y
 CONFIG_COMP_DRC=y
 CONFIG_COMP_FIR=y
+CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING=y
```
btian1, line 6: is google RTC processing is a common feature or platform specific feature? 
if it is platform specific, I  would suggest put it to platform config.

singalsu: The name of this config file is perhaps misleading, though there's "host" in the path. This is for testbench and it's safe to have every component included with it. Testbench build is a good catcher in CI for build errors since we have many components with "default n" in kconfig those might be missed in normal platform builds.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -484,44 +515,55 @@ static void google_rtc_audio_processing_free(struct comp_dev *dev)
...
 	list_for_item(source_buffer_list_item, &dev->bsource_list) {
 		struct comp_buffer *source = container_of(source_buffer_list_item,
 							  struct comp_buffer, sink_list);
-
 #if CONFIG_IPC_MAJOR_4
 		if (IPC4_SINK_QUEUE_ID(source->id) == SOF_AEC_FEEDBACK_QUEUE_ID) {
 #else
 		if (source->source->pipeline->pipeline_id != dev->pipeline->pipeline_id) {
 #endif
-			cd->aec_reference = source;
```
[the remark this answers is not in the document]
singalsu: It's because module adapter provides the input and ouput streams as arrays of pointers to process(). I use the index to access the right stream. This also eases the transition to source/sink api in pipeline2.0 with better abstraction where process() receives array of sof_source and sof_sink pointers. It let's us optimize further the pipeline buffers. I'm not sure keeping the old way would prevent that but I felt this is cleaner approach.

What do you think @marcinszkudlinski , was this the right thing to do?

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -70,6 +68,8 @@ struct google_rtc_audio_processing_comp_data {
 	uint8_t *memory_buffer;
 	struct comp_data_blob_handler *tuning_handler;
 	bool reconfigure;
+	int aec_reference_source;
```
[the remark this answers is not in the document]
singalsu: Sorry, which pointer? I'm removing above the comp_buffet pointers and I think the rest of pointers in google_rtc_audio_processing_comp_data are still used. But I may well miss here something.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -629,21 +678,25 @@ static int google_rtc_audio_processing_copy(struct comp_dev *dev)
 			}
 		}
 		num_samples_remaining -= n;
-		ref = audio_stream_wrap(&buffer->stream, ref);
+		ref = audio_stream_wrap(ref_stream, ref);
 	}
-	comp_update_buffer_consume(buffer, num_aec_reference_bytes);
+	input_buffers[cd->aec_reference_source].consumed = num_aec_reference_bytes;
```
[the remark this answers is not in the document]
singalsu: Yes, not very clean API -- frames as input and bytes as output. But as far as I understand the design is from reference firmware. Did you mean a similar helper function but for this multiple sources/sinks case? This component will be converted to source/sink API after this PR for DP mode so we could review and possibly make improvements to api there.

FYI @marcinszkudlinski

singalsu:
Marked this as DNM - I'm not sure if this PR step is necessary in converting this to DP. But please continue reviewing and sharing your thoughts so the next step can be better.

src/audio/google/google_rtc_audio_processing.c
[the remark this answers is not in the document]
cujomalainey: nvm i got my the buffers confused between the alloced ones versus the SOF ones, apologies

src/audio/google/google_rtc_audio_processing.c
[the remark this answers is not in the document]
cujomalainey: Yes, i think breaking into this struct is probably not the safest thing for API purposes on the buffer, we should fix the api or make a new call.

lgirdwood:
> Marked this as DNM - I'm not sure if this PR step is necessary in converting this to DP. But please continue reviewing and sharing your thoughts so the next step can be better.

All modules should use the module API as component API is deprecated and will gradually be removed/refactored.i.e. we only have 1 API for modules.

marcinszkudlinski:
Please proceed with merging
I'm finishing with changes to sink/src interface and DP processing of the module, all my changes have been based on top of this PR
So pls proceed and merge it once ready

singalsu:
@cujomalainey Is this version good for merge? It will be a short time version only. Work by @marcinszkudlinski for DP schduling is now done and needs this patch as base.

lkoenig (approved):
The change looks good: I haven't look into too much details.
Did you run it with the mock  an d got good audio?

src/arch/host/configs/library_defconfig
lkoenig: Don't you want to have the mock config as well here ?

lkoenig (approved):
Nice change so we can some at least build coverage.

src/audio/google/google_rtc_audio_processing_mock.c
```diff
@@ -145,11 +147,28 @@ int GoogleRtcAudioProcessingProcessCapture_int16(GoogleRtcAudioProcessingState *
...
+	 * scheme to mix to every channel for testing.
+	 */
 	memset(dest, 0, sizeof(int16_t) * state->num_output_channels * state->num_frames);
 	for (n = 0; n < state->num_frames; ++n) {
-		*out = *mic + *ref;
+		im = 0;
+		ir = 0;
+		for (io = 0; io < state->num_output_channels; io++) {
+			out[io] = sat_int16((int32_t)mic[im++] + ref[ir++]);
+			if (im == state->num_capture_channels)
```
lkoenig, line 165: You might mix several time same channel like is output is 2 and input is one channel you'll get 2 times the first channel.
```
 out[io] = sat_int16(
            (im < state->num_capture_channels ? (int32_t)mic[im++] : 0) + 
            (ir < state->num_reference_channles ? ref[ir++] : 0)
            );
```
Might be more appropriate (sorry for the vey poor codestyle).

lkoenig (changes requested):

src/audio/google/google_rtc_audio_processing_mock.c
singalsu: We don't really know what's best for testing with mock-up. Sure, I can update it to be like this.

singalsu: @lkoenig Thanks, I included your suggestion as such, it works well!

lkoenig (approved):

cujomalainey:
Sorry been busy internally, I trust @lkoenig 's review

lyakh (approved):

lgirdwood (approved):

cujomalainey (approved):

PR #8048, opened 2023-08-16
Intel: remove Intel XTOS platform code and drivers

kv2019i:
To complete the transition to Zephyr for Intel hardware targets, remove the bulk of platform code and drivers only used by Intel cAVS2.5 target in the non-Zephyr XTOS build. This is no longer supported (and hasn't been tested at all in recent times), so time to remove this.

Note: there is still some remaining bits in the codetree that will be removed in follow-up PRs. The platform driver headers need a bit more work to remove (due to depencies from rtos headers), the cAVS2.5 LPS code and a few other bits.

PR #8261, opened 2023-09-26
[SKIP CI]github action: remove duplicate build for TGL/TGL-H

aiChaoSONG:
At the beginning, we have two build jobs in github action to build IPC3 and IPC4 firmware for TGL/TGL-H.

The PR https://github.com/thesofproject/sof/pull/8048 switches cAVS2.5 configs to use IPC4 by default and empties the cAVS2.5 overlay files. After the change, the xtensa-build-zephyr.py script is building the
same IPC4 firmware with or without '-i IPC4' option. So we have two jobs running different build command but build the same IPC4 firmware.

Recently, commit 5004d0fe1e74 ("zephyr.yml: remove ipc option for zephyr build") removes '-i IPC4' option in github action for TGL/TGL-H IPC4 build. So we have duplicated jobs to build firmware for TGL and TGL-H now.

This patch removes the duplicated build job which previously is used to build IPC3 firmware for TGL/TGL-H.

### DIFF
--- a/.github/workflows/zephyr.yml
+++ b/.github/workflows/zephyr.yml
@@ -107,7 +107,6 @@ jobs:
         IPC_platforms: [
           # - IPC3 default
           imx8 imx8x imx8m,
-          tgl tgl-h,  # UNSUPPORTED! Will be removed
           # - IPC4 default
           mtl, lnl,
           # Temporary testbed for Zephyr development.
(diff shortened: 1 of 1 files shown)


lgirdwood (approved):
@keqiaozhang @marc-hb can we proceed and remove the duplication now ?

marc-hb (changes requested):
Thanks @aiChaoSONG for the cleanup and very detailed commit message.

There are still some obsolete comments like "# Very few IPC3 platforms support IPC4 too.". That's not true any more.

Almost there!

keqiaozhang (approved):
LGTM

marc-hb (approved):

tmleman (approved):

PR #8267, opened 2023-09-26
.github/sparse-zephyr: add -DCONFIG_MINIMAL_LIBC to avoid picolibc

marc-hb:
Add a sparse-specific workaround for the incompatibility with picolibc (the new Zephyr default)
- https://github.com/zephyrproject-rtos/zephyr/issues/63003

Also fix comment in commit 2a9473a17be4 ("app/prj.conf: disable PICOLIBC with CONFIG_MINIMAL_LIBC=y"): we don't need to disable PICOLIBC _everywhere_; we only need to disable it when using sparse.


lgirdwood (approved):

tmleman (approved):

lyakh (approved):
would a sparse overlay be possible and desirable?

PR #8274, opened 2023-09-29
ipc4: mixin/mixout: Remove redundant source_info stuff

serhiy-katsyuba-intel:
Mutex protected module_source_info was an old failed attempt to support cross-core connection of mixin and mixout. That functionality was never properly tested as FW did not support cross-core connections there.

For connected mixin and mixout located on same core, that functionality is not needed. For cross-core connection, that functionality creates dead-lock.

### DIFF
--- a/src/audio/mixin_mixout/mixin_mixout.c
+++ b/src/audio/mixin_mixout/mixin_mixout.c
@@ -79,10 +79,22 @@ struct mixin_data {
 	struct mixin_sink_config sink_config[MIXIN_MAX_SINKS];
 };
 
-/* mixout component private data. This can be accessed from different cores. */
+/*
+ * Mixin calls "consume" on its source data but never calls "produce" -- that one is called
+ * by mixout for its sink data. So between mixin_process() and mixout_process() a number of
+ * consumed (in mixin) yet not produced (in mixout) frames should be stored for each mixin
+ * and mixout pair.
+ */
+struct pending_frames {
+	struct comp_dev *mixin;
+	uint32_t frames;
+};
+
+/* mixout component private data */
 struct mixout_data {
 	/* number of currently mixed frames in mixout sink buffer */
 	uint32_t mixed_frames;
+
 	/*
 	 * Source data is consumed by mixins in mixin_process() but sink data cannot be
 	 * immediately produced. Sink data is produced by mixout in mixout_process() after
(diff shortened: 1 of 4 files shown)


src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -727,13 +718,95 @@ static int mixout_prepare(struct processing_module *mod,
...
+	/* find an empty slot in the pending_frames array */
+	pending_frames = get_mixin_pending_frames(mixout_data, NULL);
+	if (!pending_frames) {
+		/* no free slot in pending_frames array */
+		comp_err(mod->dev, "Too many inputs!");
+		return -ENOMEM;
+	}
+
+	pending_frames->mixin = mixin;
+	pending_frames->frames = 0;
```
lyakh, line 774: you need to swap these: if between them the copying thread finds this mixin with uninitialised `.frames` - it will try to copy data from it...

lyakh (commented):
this not only removes `source_info` from mixin-mixout, but also removes generic module binding and unbinding and moves it to mixin-mixout only. This might be valid, but it's certainly more than what the commit description is saying

src/audio/mixin_mixout/mixin_mixout.c
```diff
@@ -79,20 +79,45 @@ struct mixin_data {
 	struct mixin_sink_config sink_config[MIXIN_MAX_SINKS];
 };
 
-/* mixout component private data. This can be accessed from different cores. */
+/*
+ * Mixin calls "consume" on its source data but never calls "produce" -- that one is called
+ * by mixout for its sink data. So between mixin_process() and mixout_process() a number of
+ * consumed (in mixin) yet not produced (in mixout) frames should be stored for each mixin
+ * and mixout pair.
```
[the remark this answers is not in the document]
serhiy-katsyuba-intel: Yes, mixin can be connected to multiple mixouts and mixout can be connected to multiple mixins. A pair here means a connection pair (think of it as an unused buffer between mixin and mixout). Amount of pending frames should be stored for each mixin <-> mixout connection pair. Such pair info is stored in mixout_data as an array of MIXOUT_MAX_SOURCES. Each entry besides pending frame also contains a mixin, So, basically, this is std::map<mixin, frames> stored in each mixout. No problem here with many-to-many topologies.

serhiy-katsyuba-intel:
> this not only removes `source_info` from mixin-mixout, but also removes generic module binding and unbinding and moves it to mixin-mixout only. This might be valid, but it's certainly more than what the commit description is saying

No, module_adpter_bind() and unbind() are still there. They are just stripped from the stuff that was only needed for mixin/mixout. Previously that was implemented in mixout_bind() and unbind() and were moved to module_adapter after mixin/mixout was modified to use it. That module_adapter_source_info is (1) not needed at all, (2) does not work: does nothing on single core and creates dead-lock on cross-core.

src/audio/mixin_mixout/mixin_mixout.c
serhiy-katsyuba-intel: Swapped. However, this was not really necessary:
(1) frames are cleared in unbind() and on initialization,
(2) bind() is executed on same core as LL and it blocks interrupts (blocks LL to preempt IPC or IDC thread). So bind cannot be preempted by LL (at least this is true for same core binding). This version of mixin and mixout only supports same core connection. For cross-core connection (we are still uncertain if we need it) a separate version of mixin and mixout should be implemented.

lgirdwood:
@serhiy-katsyuba-intel there were also some recent MCPS optimisations for local core 1:1 copying. Can you confirm there is no MCPS impact for this use case. Thanks

serhiy-katsyuba-intel:
> @serhiy-katsyuba-intel there were also some recent MCPS optimisations for local core 1:1 copying. Can you confirm there is no MCPS impact for this use case. Thanks

The PR does not add new code to execution flow, it just removes module_source_info stuff which is not needed. The performance might become a little bit better or at least the same, not worse.

marcinszkudlinski (approved):

lgirdwood (approved):

tmleman (approved):

lgirdwood:
Rerun the actions and Zephyr issues are known and unrelated..

PR #8279, opened 2023-10-02
Audio: TDFB: Update component for IPC4

singalsu:
First patch adds the needed IPC4 header

Second patch updates the TDFB component

### DIFF
--- a/src/audio/tdfb/tdfb.c
+++ b/src/audio/tdfb/tdfb.c
@@ -40,6 +40,12 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#if CONFIG_IPC_MAJOR_4
+#include <ipc4/header.h>
+#include <ipc4/module.h>
+#include <ipc4/notification.h>
+#endif
+
 /* The driver assigns running numbers for control index. If there's single control of
  * type switch, enum, binary they all have index 0.
  */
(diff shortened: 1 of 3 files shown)


src/audio/tdfb/tdfb.c
```diff
@@ -57,10 +63,58 @@ DECLARE_SOF_RT_UUID("tdfb", tdfb_uuid,  0xdd511749, 0xd9fa, 0x455c, 0xb3, 0xa7,
...
+				 sizeof(struct tdfb_notification_payload));
+	if (!msg)
+		return NULL;
+
+	payload = (struct tdfb_notification_payload *)msg->tx_data;
+	payload->module_data.instance_id = IPC4_INST_ID(ipc_config->id);
+	payload->module_data.module_id = IPC4_MOD_ID(ipc_config->id);
+	payload->module_data.event_id = SOF_IPC4_NOTIFY_MODULE_EVENTID_ALSA_MAGIC_VAL |
+		control_type_param_id;
+	payload->module_data.event_data_size = sizeof(struct sof_ipc4_control_msg_payload) +
```
lgirdwood, line 96: where do we validate size ?

lgirdwood (commented):
@singalsu is there an upper limit on what size should be here for TDFB ? if so, we should check our payload is above min and below max sizes.

src/audio/tdfb/tdfb.c
singalsu: Sorry, what do you mean? The above  = ipc_msg_w_ext_init() has accepted the entire message size at this point of code. The used notification payload is fixed size, there is only one num_elems in used enum control. The kernel would error if the channel would not exist in topology.

lgirdwood: The FW should not trust the sender, i.e. it has to check the data is a valid size, so ```uint32_t event_data_size;``` has to be checked by FW.

singalsu: FW is in this case the sender and creates in this function a message template that is used for entire life time of it. When enum control is updated, only  the `ipc_payload->control_value.value` is updated in the next function tdfb_send_notification(). I think I'm missing something but I don't know how to make this more secure than this.

lgirdwood: This sounds fine is FW locally creates this data, the point is that remotely generated data is untrsuted and should always be validated.

src/audio/tdfb/tdfb.c
```diff
@@ -633,17 +712,82 @@ static int tdfb_cmd_set_value(struct processing_module *mod, struct sof_ipc_ctrl
...
+		return tdfb_cmd_switch_set(ctl, cd);
+
+	case SOF_IPC4_ENUM_CONTROL_PARAM_ID:
+		comp_info(mod->dev, "SOF_IPC4_ENUM_CONTROL_PARAM_ID id = %d, num_elems = %d",
+			  ctl->id, ctl->num_elems);
+		return tdfb_cmd_enum_set(ctl, cd);
+	}
+#endif /* CONFIG_IPC_MAJOR_4 */
+
+#if CONFIG_IPC_MAJOR_3
```
lyakh, line 785: `#elif`

src/audio/tdfb/tdfb.c
```diff
@@ -633,17 +712,82 @@ static int tdfb_cmd_set_value(struct processing_module *mod, struct sof_ipc_ctrl
 	comp_err(mod->dev, "tdfb_cmd_set_value() error: invalid cdata->cmd");
 	return -EINVAL;
 }
+#endif /* CONFIG_IPC_MAJOR_3 */
+
+#if CONFIG_IPC_MAJOR_4
```
lyakh, line 717: `#elif`

src/audio/tdfb/tdfb.c
```diff
@@ -557,11 +627,18 @@ static int tdfb_cmd_get_value(struct processing_module *mod, struct sof_ipc_ctrl
...
-			   uint32_t config_id, uint32_t *data_offset_size,
+			   uint32_t param_id, uint32_t *data_offset_size,
 			   uint8_t *fragment, size_t fragment_size)
 {
+#if CONFIG_IPC_MAJOR_4
+	comp_err(mod->dev, "tdfb_get_config(), Not supported, should not happen");
+	return -EINVAL;
+#endif
+
+#if CONFIG_IPC_MAJOR_3
```
lyakh, line 641: `#elif`

src/audio/tdfb/tdfb.c
```diff
@@ -57,10 +63,58 @@ DECLARE_SOF_RT_UUID("tdfb", tdfb_uuid,  0xdd511749, 0xd9fa, 0x455c, 0xb3, 0xa7,
...
+		(union ipc4_notification_header *)&msg_proto.header;
+	struct ipc_msg *msg;
+	struct tdfb_notification_payload *payload;
+
+	memset_s(&msg_proto, sizeof(msg_proto), 0, sizeof(msg_proto));
+	primary->r.notif_type = SOF_IPC4_MODULE_NOTIFICATION;
+	primary->r.type = SOF_IPC4_GLB_NOTIFICATION;
+	primary->r.rsp = SOF_IPC4_MESSAGE_DIR_MSG_REQUEST;
+	primary->r.msg_tgt = SOF_IPC4_MESSAGE_TARGET_FW_GEN_MSG;
+	msg = ipc_msg_w_ext_init(msg_proto.header, msg_proto.extension,
```
lyakh, line 86: `msg_proto.extension` hasn't been set, so it's 0, right? Could use it explicitly to make it clearer

src/audio/tdfb/tdfb.c
```diff
@@ -57,10 +63,58 @@ DECLARE_SOF_RT_UUID("tdfb", tdfb_uuid,  0xdd511749, 0xd9fa, 0x455c, 0xb3, 0xa7,
...
+{
+	struct tdfb_notification_payload *ipc_payload;
+
+	ipc_payload = (struct tdfb_notification_payload *)msg->tx_data;
+	ipc_payload->control_value.value = val;
+	ipc_msg_send(msg, NULL, false);
+}
+#endif /* CONFIG_IPC_MAJOR_4 */
 
+#if CONFIG_IPC_MAJOR_3
```
lyakh, line 117: `#elif`

src/audio/tdfb/tdfb.c
```diff
@@ -447,9 +502,21 @@ static int tdfb_init(struct processing_module *mod)
...
+	cd->msg = tdfb_notification_init(mod, SOF_IPC4_ENUM_CONTROL_PARAM_ID,
+					 CTRL_INDEX_AZIMUTH_ESTIMATE);
+	if (!cd->msg) {
+		comp_err(dev, "Failed to initialize control notification.");
+		ret = -EINVAL;
+		goto err_free_cd;
+	}
+#endif /* CONFIG_IPC_MAJOR_4 */
+
+#if CONFIG_IPC_MAJOR_3
```
lyakh, line 515: `#elif`

lgirdwood (approved):

src/include/ipc4/header.h
```diff
@@ -173,6 +173,7 @@ struct ipc4_message_reply {
 
 #define SOF_IPC4_SWITCH_CONTROL_PARAM_ID 200
 #define SOF_IPC4_ENUM_CONTROL_PARAM_ID  201
+#define SOF_IPC4_NOTIFY_MODULE_EVENTID_ALSA_MAGIC_VAL ((uint32_t)(0xA15A << 16))
```
btian1, line 176: why not directly with 0xA15A0000

RanderWang (approved):
LGTM

kv2019i (approved):

PR #8280, opened 2023-10-02
Tools: Topology2: Add widget class TDFB

singalsu:
First patch adds enum control class.

Second patch adds TDFB component widget class and example configuration blobs.


tools/topology/topology2/include/components/tdfb/line2_50mm_azm90_90_13_16khz.conf
```diff
@@ -0,0 +1,361 @@
+# Exported with script example_line_array.m 02-Oct-2023
```
lgirdwood, line 1: Is this the full command line to generate the data ?

lgirdwood (commented):
The data needs to say something like
```
Generated by running "octave example.m -a 50 -b 100 -c 1 out.file"
```

tools/topology/topology2/include/components/tdfb.conf
```diff
@@ -0,0 +1,174 @@
...
+		# Enum controls
+		enum."1" {
+			Object.Base {
+				channel.1 {
+					name	"fc"
+					reg	3
+					shift	0
+				}
+				text.0 {
+					name "mytext"
```
ranj063, line 102: @singalsu is this intentional to add 2 enum controls here with the same set of values? Also, can we rename this to something more meaningful like tdfb_enum_text ?

singalsu: I think I should add comment to explain all controls. The first enum control is for beam angle set, the second enum is for estimated sound arrival angle. I was wondering if I should get the values from an include file to avoid duplication. What do you think?

tools/topology/topology2/include/components/tdfb/line2_50mm_azm90_90_13_16khz.conf
singalsu: That's all for Octave or Matlab shell, but I can add second comment line how to build it from SOF tree top. I just added in another PR additional comment print line to these.

lgirdwood: Thanks, my aim is to make it easy for anyone navigating topology that they can easily work out how to re-tune any module that has coefficients. Long term we need to go a little further and install all the octave/matlab tooling (with a ```sof-``` prefix) so that thay can run 
```
sof-tdfb.m --help
```
and get all the cmd line args or instructions to tune TDFB (and likewise for all modules).

lgirdwood:
@singalsu any update ?

singalsu:
> @singalsu any update ?

It's updated now, new version of Matlab script is in #8374 that I used to generate the blobs.

tools/topology/topology2/include/components/tdfb.conf
```diff
@@ -0,0 +1,174 @@
+#
+#
+# A generic TDFB component. All attributes defined herein are namespaced
+# by alsatplg to "Object.Widget.tdfb.attribute_name"
```
kv2019i, line 4: minor: one explanation of what TDFB means would be nice in tdfb.conf (it is in the git commit and other places so not a big issue)

kv2019i (approved):
Thanks @singalsu , looks good!

tools/topology/topology2/include/components/tdfb.conf
singalsu: I just updated it, better to have it in the class definition .conf and not just in the commit message.

lgirdwood (approved):

PR #8282, opened 2023-10-03
topology1: sof-adl-cs35l41: remove pop noise

brentlu:
Enable bclk early start to remove speaker pop noise when resuming from S3 or switching audio path.

[copy from issue tracker]
What steps will reproduce the problem?
1. Updated Rev8 image via recovery media,went to OOBE and boot into OS.
2. Play audio file with Gallery app.
3. Plug Typec headphone to unit then unplug it.
4. Find Po noise output from speaker.

What is the expected output?
There is no" Po" noise output from speaker when unplug Typec headphone/Rusume unit from S3 with audio playing.

What do you see instead?
Po noise output from speaker when unplug Typec headphone/Rusume unit from S3 with audio playing.

### DIFF
--- a/tools/topology/topology1/sof-tgl-max98357a-rt5682.m4
+++ b/tools/topology/topology1/sof-tgl-max98357a-rt5682.m4
@@ -475,7 +475,7 @@ ifelse(
 		SSP_CLOCK(bclk, 6144000, codec_slave),
 		SSP_CLOCK(fsync, 48000, codec_slave),
 		SSP_TDM(4, 32, 3, 15),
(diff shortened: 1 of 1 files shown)


cujomalainey:
@brentlu I don't think the partner issue link is visible to most people so probably best not to post it

brentlu:
> @brentlu I don't think the partner issue link is visible to most people so probably best not to post it

ok, let me post info from issue description instead.

lgirdwood (approved):

kv2019i (approved):

PR #8283, opened 2023-10-03
Tools: Tune: dcblock: Create set of cut-off frequencies

singalsu:
This patch modifies script example_dblock.m to create in addition to existing default coefficients blob with parameter R=0.98 a set of blobs those provide (-3 dB) cut-off frequencies 20, 30, 40, 50, 100, 200 Hz for 16 kHz and 48 kHz for various usages. Human understandable parameters for blobs helps to select the configuration for the needed dcblock usage.

The parameter value R for given frequencies is calculated with iterative function dcblock_rval_calculate(). A closed form equation might be possible to derive from the transfer function. If such is found this function can be replaced with quicker equation -- or with more advanced faster converging iteration.

The topology blob export functions are also modified to add comment line for exact build command.

### DIFF
--- a/tools/tune/dcblock/example_dcblock.m
+++ b/tools/tune/dcblock/example_dcblock.m
@@ -1,16 +1,53 @@
 function example_dcblock()
 
+% Default blob, about 150 Hz cut-off @ 48 kHz
+prm.fc = [];
+prm.fs = [];
+prm.R_coeffs = [0.98, 0.98, 0.98, 0.98, 0.98, 0.98, 0.98, 0.98];
+prm.id = "default";
+dcblock_blob_calculate(prm)
+
+% Generate a set configuration blobs for 16 and 48 kHz rate
+% The selected high-pass cut-off frequencies are in range 20 - 200 Hz:
+% 20, 30, 40, 50, 80, 100, and 200 Hz
+%
+% Select for applications one that is best tradeoff between DC level
+% settle time and the lowest frequency that should pass without
+% much attenuation. Cut-off frequency by definition is the point where
+% the attenuation is 3.01 dB.
+
+prm.R_coeffs = [];
+for fs = [16e3 48e3]
+	for fc = [20 30 40 50 80 100 200]
+		prm.id = sprintf("%dhz_%dkhz", round(fc), round(fs / 1000));
+		prm.fs = fs;
+		prm.fc = fc;
+		dcblock_blob_calculate(prm)
+	end
+end
+
+end
+
+function dcblock_blob_calculate(prm)
+
 % Set the parameters here
-tplg1_fn = "../../topology/topology1/m4/dcblock_coef_default.m4"; % Control Bytes File
-tplg2_fn = "../../topology/topology2/include/components/dcblock/default.conf";
+tplg1_fn = sprintf("../../topology/topology1/m4/dcblock_coef_%s.m4", prm.id); % Control Bytes File
+tplg2_fn = sprintf("../../topology/topology2/include/components/dcblock/%s.conf", prm.id);
 % Use those files with sof-ctl to update the component's configuration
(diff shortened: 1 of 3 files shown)


tools/tune/dcblock/example_dcblock.m
```diff
@@ -1,28 +1,61 @@
...
+prm.fc = [];
+prm.fs = [];
+prm.R_coeffs = [0.98, 0.98, 0.98, 0.98, 0.98, 0.98, 0.98, 0.98];
+prm.id = "default";
+dcblock_blob_calculate(prm)
+
+% Generate a set of cut-off frequencies for 16 and 48 kHz rate
+prm.R_coeffs = [];
+for fs = [16e3 48e3]
+	for fc = [20 30 40 50 80 100 200]
```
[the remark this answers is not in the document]
singalsu: Fixed now!

lgirdwood (approved):

tools/tune/dcblock/example_dcblock.m
```diff
@@ -36,3 +77,54 @@ function example_dcblock()
...
+
+if (fc_target / fs < 10 / 48e3 || fc_target / fs > 1000 / 48e3)
+	error("Illegal fc_target");
+end
+
+h_target = 1 / sqrt(2); % -3.01 dB
+R = 0.5;
+R_step = 0.005;
+sign = 1;
+err_prev = inf;
```
ShriramShastry, line 92: The value assigned to `err_prev` could be unnecessary.

tools/tune/dcblock/example_dcblock.m
```diff
@@ -36,3 +77,54 @@ function example_dcblock()
...
+w = 2 * pi * fc_target / fs;
+j = sqrt(-1);
+z = exp(j * w);
+
+% Iteration 0
+h = (1 - z^-1) / (1 - R * z^-1);
+a = abs(h);
```
ShriramShastry, line 99: The value assigned to `a` could be unnecessary.

tools/tune/dcblock/example_dcblock.m
```diff
@@ -36,3 +77,54 @@ function example_dcblock()
...
+		sign = -sign;
+		R_step = R_step / 2;
+	end
+	R = R + sign * R_step;
+	err_prev = err;
+end
+
+% Sane result?
+if R < eps || R > 1 - eps
+	error("Calculate of R iteration failed")
```
ShriramShastry, line 117: Perhaps there is a missing semicolon at the end.

ShriramShastry (approved):
Overall, as a first drafting for changes to cut-off frequencies, it looks good to me, with only a few minor suggestions caught.

Thank you for making the necessary modifications.

cujomalainey (approved):

tools/tune/dcblock/example_dcblock.m
kv2019i: @singalsu issue or not?

kv2019i:
@singalsu  Mandatory tests need to pass, otherwise this seems ready to go.

tools/tune/dcblock/example_dcblock.m
singalsu: I think it's good to have the semicolon there, Matlab syntax does not require it, but possibly it could display unnecessary return status from error(). I'll do Sriram's suggested fixes, stay tuned.

tools/tune/dcblock/example_dcblock.m
singalsu: Yep, remain from previous iteration code version, both can be deleted.

singalsu:
@ShriramShastry Thanks for spotting the unnecessary code lines, those are fixed now!

lgirdwood (approved):

PR #8288, opened 2023-10-05
[TEST ONLY] Use SRC DP module

yongzhi1:

### DIFF
--- a/tools/topology/topology2/cavs-sdw.conf
+++ b/tools/topology/topology2/cavs-sdw.conf
@@ -11,6 +11,7 @@
 <tokens.conf>
 <virtual.conf>
 <host-copier-gain-mixin-playback.conf>
+<src-gain-mixin-playback.conf>
 <mixout-gain-alh-dai-copier-playback.conf>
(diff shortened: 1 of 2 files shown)


lgirdwood (commented):
@singalsu pls review.
@yongzhi1 good stuff, should we keep the volume too ?

yongzhi1:
> @yongzhi1 good stuff, should we keep the volume too ?
The two gain module are kept, since SRC is the only DP-compatible module right now, I am using it as a ref for DP multi-core support.

lgirdwood:
@marcinszkudlinski I assume the topology usage above is good for you ?

marcinszkudlinski:
yes, topology looks good, but some issues found, working with @yongzhi1 to resolve.

singalsu (approved):

yongzhi1:
Thanks @marcinszkudlinski for the great support, SRC as DP scheduling module tested working on core 2 now, so this PR served its purpose. Let me close and move the focus to https://github.com/thesofproject/sof/pull/8308, thanks!!

PR #8289, opened 2023-10-05
Fix testbench xt build fail

singalsu:


lgirdwood (approved):

src/audio/mfcc/mfcc_hifi3.c
```diff
@@ -121,7 +121,7 @@ void mfcc_fill_prev_samples(struct mfcc_buffer *buf, int16_t *prev_data,
 
 	buf->s_avail -= prev_data_length;
 	buf->s_free += prev_data_length;
-	buf->r_ptr = in;
+	buf->r_ptr = (void *)in; /* int16_t pointer but direct cast is not possible */
```
lgirdwood, line 124: is ```in``` the wrong ptr type, do we need to look at other changes here ?

src/audio/module_adapter/module_adapter.c
```diff
@@ -1578,10 +1578,12 @@ int module_adapter_reset(struct comp_dev *dev)
 	}
 #endif /* CONFIG_ZEPHYR_DP_SCHEDULER */
 	if (IS_PROCESSING_MODE_SINK_SOURCE(mod)) {
+		int i;
```
kv2019i, line 1581: I wonder if we could build testbench in C99 mode. When we build SOF currently (at least Zephyr builds), we are using C99 mode:
```
-std=c99 -std=gnu99
```

... so basicly C99'ism can pass through our CI checks and testbench will always be in catch-up mode. Zephyr is C99+ -> https://docs.zephyrproject.org/latest/develop/languages/c/index.html

That's a harder question where to set this, but would be a longer term solution.

src/audio/mfcc/mfcc_hifi3.c
singalsu: It's up to @andrula-song to confirm but in the patch she used ae_int32 type to load two ae_int16 types, that's in comments a few lines above this typecast. I was wondering too if this can be correct but it appears so.

src/audio/module_adapter/module_adapter.c
singalsu: Need to check how to add this, this commit could be possible to drop with that change.

src/audio/mfcc/mfcc_hifi3.c
lgirdwood: ok, intrinsic types are non portable. Lets revisist as we do need to support AVX here (and the API would need to become more generic).

lgirdwood (approved):

kv2019i (approved):
Thanks @singalsu !

src/audio/mfcc/mfcc_hifi3.c
andrula-song: @singalsu  @lgirdwood , I used to use AE_LA16X4_XC to load a 16bit * 4 data, but I found the result is not right, I think it is a bug of xtensa. So I use AE_L32_XC to load a 32bit data(16bit * 2) to replace twice AE_L16_XC, and force the datatype to ae_int16 to copy the last 16 bit if the input data size is odd.

PR #8292, opened 2023-10-05
[SKIP SOF-TEST] Clean up .gitignore now that we're using CMake

marc-hb:
"Out of source" builds have been used since the build was switched to CMake in 2019. So there's no need to ignore specific .o, .la and a gazillion of other files extensions anymore. It's now enough to ignore `build*` directories, no need to ignore files by extension any more.

Also remove personal ignore preferences like *.mbox, cscope* etc. They belong to a centralized file in the HOME directory, not to a massive copy/paste/diverge across all projects.


marc-hb:
@wszypelt the test `tests/avs/fw_02_base_fw_api/test_01_fw_large_config_get.py::TestFwLargeConfigGet::test_02_01_fw_config --platform=TGL` started timing out. The same test timed out the same in both  https://sof-ci.01.org/sof-pr-viewer/#/build/PR8292/build12911196 and totally unrelated https://sof-ci.01.org/sof-pr-viewer/#/build/PR8286/build12911195 

All other tests passed.

.gitignore
```diff
@@ -1,50 +1,16 @@
...
+# cscope*, TAGS, *~,...
+#
+# Instead, set your non-project specific, personal git ignore preference
+# only once for all git projects in your home directory:
+#     https://git-scm.com/docs/gitignore
+# So this does not interfere other people's "git status" preferences.
+#
+# Also, don't use `git add *` or `git commit -a` and git push blindly
+# without a look at which files are getting pushed. This is bad no
+# matter what is in git ignore.
```
kv2019i, line 12: Hmm, not sure about this. Is there any harm to the overblown gitignore file? 

I mean we frequently have contributors not so used to open-source tools, so misuse of "git commit -a" is a non-zero risk and large gitignore may provide some protection against this (and bring value to the project).

kv2019i (commented):
Neutral, I can go with this but is there are harm in the big file. At least I've not hit with gitignore blocking relevant contributions ever in SOF.

(And personal note: I don't like to use gitignore much, I prefer to see a list of all files that are not tracked by git, so personally I'm all for minimizing this. but given we had a large gitignore, going back to minimal seems to cause some risk)

lgirdwood (commented):
Lets deal with the CMake changes 1st. The other tooling changes can come later, we also have to deal with tools.

.gitignore
marc-hb: > Is there any harm to the overblown gitignore file?

Yes :-)

> may provide "some protection" against this

"Some" protection is worse than none.

The only way for _both_ authors and reviewers to miss spurious files is to NOT review. "Some" protection teaches contributors and reviewers that not reviewing is fine and that `git commit -a` is fine - because it works most of the time thanks to "some" protection! Until a file not "protected" by .gitignore gets added and missed.

Moreover, even if some people "know better" and really want to use `git commit -a` then it's much "better" even _for them_ to configure it once for all in the $HOME/gitignore file and avoid an "inconsistent git commit -a experience" across repos.

kv2019i:
@dbaluta @marcinszkudlinski others, thoughts? I'm really neutral on this.

marc-hb:
SOFCI TEST

marc-hb:
I pushed it to storage and it does not seem to break anything. Can you please take a quick look at this test run  and make sure the cleanup happens as expected?
https://sof-jenkins.ostc.intel.com/job/sof_plan_build/5466/

kv2019i (approved):

PR #8296, opened 2023-10-06
[DNM][Do Not Review][Demo Only] LMDK 006 demo

lyakh:
This is only a partial demonstration of the "development kit" loadable module approach, ported on top of mtl-007-drop-stable. The example smart-amp-test, built from the lmdk/ directory can be loaded, it initialises, but further behaviour breaks either during preparation or when copying. A possible reason would be some further headers that need to be updated. Frequently changing headers make maintaining the "development kit" difficult and absence of logging from the modules make debugging complicated


lgirdwood:
@RDharageswari @mengdonglin @aiChaoSONG  fyi.
@pjdobrowolski fyi - we are not yet sure why we have breakage here at runtime. Can you confirm mixin is working correctly with the copied headers or are we missing some recent patches ?

lgirdwood:
@mwasko fyi

PR #8298, opened 2023-10-10
LL tick source start/stop fix

marcinszkudlinski:
as DP uses LL as a tick source, it starts an empty LL task to ensure that an instance of LL is running
This applies to mostly to secondary cores, but is may happen that there's no LL instance started on primary

This commit fixes 2 problems
 - LL tick source was stated once for every DP module
 - LL tick source wasn't stopped when there were no DP modules running anymore

### DIFF
--- a/src/schedule/zephyr_dp_schedule.c
+++ b/src/schedule/zephyr_dp_schedule.c
@@ -30,7 +30,7 @@ DECLARE_TR_CTX(dp_tr, SOF_UUID(dp_sched_uuid), LOG_LEVEL_INFO);
 
 struct scheduler_dp_data {
 	struct list_item tasks;		/* list of active dp tasks */
-	struct task task;		/* LL task - source of DP tick */
+	struct task ll_tick_src;	/* LL task - source of DP tick */
 };
 
 struct task_dp_pdata {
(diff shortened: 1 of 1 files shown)


src/schedule/zephyr_dp_schedule.c
```diff
@@ -403,7 +403,7 @@ int scheduler_dp_init(void)
 	scheduler_init(SOF_SCHEDULE_DP, &schedule_dp_ops, dp_sch);
 
 	/* init src of DP tick */
-	ret = schedule_task_init_ll(&dp_sch->task,
+	ret = schedule_task_init_ll(&dp_sch->ll_tick_src,
```
kv2019i, line 406: Not related to this PR, but revisiting this code, one possibility would be to just use k_timer to run scheduler_dp_ll_tick() every 1ms. But, but , this would be from IRQ context, so reusing the LL thread should work as well.

kv2019i (approved):

src/schedule/zephyr_dp_schedule.c
marcinszkudlinski: We need to run it when LL finishes, using of LL task is intentional

tmleman (approved):

PR #8299, opened 2023-10-10
[MTL 006] intel: ace: optimization of DSP clock usage

tmleman:
After this PR aDSP will dynamically switch clock for power savings.

### DIFF
--- a/src/ipc/ipc4/handler.c
+++ b/src/ipc/ipc4/handler.c
@@ -42,6 +42,8 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#include <zephyr/pm/policy.h>
+
 /* Command format errors during fuzzing are reported for virtually all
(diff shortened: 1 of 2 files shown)


src/ipc/ipc4/handler.c
```diff
@@ -402,13 +404,17 @@ static int ipc_wait_for_compound_msg(void)
 	int try_count = 30; /* timeout out is 30 x 10ms so 300ms for IPC */
 
 	while (atomic_read(&msg_data.delayed_reply)) {
+		pm_policy_state_lock_get(PM_STATE_ACTIVE, 1);
```
btian1, line 407: please specify the meaning of "1", other sof code use macro like: PM_ALL_SUBSTATES instead of direct numbers.

tmleman: Will be done better on upstream.

PR #8301, opened 2023-10-10
Audio: IGO_NR: Convert to module adapter

singalsu:
This is first "working" draft IPC4 version, but processing with stub library is forced to 48 frames for 1 ms LL scheduling. It obviously can't be used with the real processing library. For 16 ms need to change this further to support DP mode in IPC4.


src/audio/igo_nr/igo_nr.c
```diff
@@ -750,95 +670,104 @@ static void igo_nr_lib_init(struct comp_dev *dev)
...
+	/* Default NR on
+	 *
+	 * Note: There is a race condition with this switch control set and kernel set
+	 * ALSA switch control if such is defined in topology. This set overrides kernel
+	 * SOF driver set for control since it happens just after component init. The
+	 * user needs to re-apply the control to get expected operation. The owner of
+	 * this component should check if this is desired operation. A possible fix would
+	 * be set here only if kernel has not applied the switch control.
+	 */
 	cd->process_enable[cd->config.active_channel_idx] = true;
```
singalsu, line 742: @fuyuntsuo Can you please check if this needs a fix.

src/audio/igo_nr/igo_nr.c
```diff
@@ -440,67 +401,55 @@ static int32_t igo_nr_params(struct comp_dev *dev,
...
-		comp_err(dev, "igo_nr_cmd_get_value() error: invalid cdata->cmd %d", cdata->cmd);
-		ret = -EINVAL;
-		break;
+		comp_err(dev, "igo_nr_cmd_get_config() error: invalid cdata->cmd %d", cdata->cmd);
+		return -EINVAL;
 	}
-	return ret;
+
+#elif CONFIG_IPC_MAJOR_4
+	comp_err(dev, "igo_nr_cmd_get_config() error: not supported");
```
singalsu, line 450: @ranj063 How should we handle get_config() with IPC4, error it like here, or add code to populate the struct with response as in IPC3? If a control is volatile a get_config() to firmware might happen. I tried it today, but kernel doesn't allow volatile controls for dynamic pipelines. So, is it best to just return error if this would get called?

lgirdwood: If not supported **today** then return an error, we can always come back when kernel/topology ready.

lgirdwood (commented):
Looks clean.

src/audio/igo_nr/igo_nr.c
lgirdwood: Lets keep any existing assumptions, we can always come back and address if newer versions of IGO need the fix.

src/audio/igo_nr/igo_nr.c
singalsu: Yep, we discussed this with @ujfalusi , the control get to firmware is not happening unless we update the kernel to handle volatile type. The only case is when FW sends an empty control notification to which kernel responds with control get. But IGO NR code is not sending notifications.

lgirdwood: Thats fine, please just proceed with what we have and we can update for IPC4 when kernel ready.

src/audio/igo_nr/igo_nr.c
```diff
@@ -275,124 +277,97 @@ static inline int32_t set_capture_func(struct comp_dev *dev)
...
-		comp_err(dev, "igo_nr_verify_params(): comp_verify_params() failed.");
-
-	return ret;
+	return 0;
 }
 
-/* set component audio stream parameters */
-static int32_t igo_nr_params(struct comp_dev *dev,
-			     struct sof_ipc_stream_params *pcm_params)
+#if DISABLED_CODE
```
lyakh: is `DISABLED_CODE` defined anywhere? Checking with an `#if` means that the macro is defined and you're checking for it to be equal to 0, I've seen builds failing when an undefined macro is used like this, possibly it depends on the compiler version or on compilation flags. Just use `#ifdef` unless you're sure that it's defined

src/audio/igo_nr/igo_nr.c
```diff
@@ -275,124 +277,97 @@ static inline int32_t set_capture_func(struct comp_dev *dev)
...
+/* check component audio stream parameters */
+static int32_t igo_nr_check_params(struct processing_module *mod)
```
lyakh: Why is it changed if it's disabled code now? Or will it be re-enabled for the final version of the PR?

src/audio/igo_nr/igo_nr.c
```diff
@@ -360,48 +497,42 @@ static int igo_nr_free(struct processing_module *mod)
 	return 0;
 }
 
-#if DISABLED_CODE
```
lyakh: ok, I see now, for the final version would be good to have it a bit better to avoid breaking bisection

lyakh (commented):
that's a lot of rework!

src/audio/igo_nr/igo_nr.c
lgirdwood: We can change this define to something like ```IGO_NR_IPC4_INTEGRATION_TODO``` as place a large comment next to the macro declaration stating this is work that still needs to be done for ipc4/module API that needs to be done by someone with the latest IGO NR binaries (and know what they are doing).

singalsu: I think I can fix this for audio stream type module adapter client even if it's temporary version.

src/audio/igo_nr/igo_nr.c
singalsu: Sorry, I'll clean up this patch to be less confusing.

src/audio/igo_nr/igo_nr.c
singalsu: Yep, I'll work out the intermediate version from this final. The function  wasn't essential for running the component so I just disabled it for first version.

src/audio/igo_nr/igo_nr.c
```diff
@@ -275,124 +277,96 @@ static inline int32_t set_capture_func(struct comp_dev *dev)
...
 	}
 
-	comp_set_drvdata(dev, cd);
-
 	/* Handler for configuration data */
 	cd->model_handler = comp_data_blob_handler_new(dev);
 	if (!cd->model_handler) {
-		comp_cl_err(&comp_igo_nr, "igo_nr_new(): comp_data_blob_handler_new() failed.");
+		comp_err(dev, "igo_nr_init(): comp_data_blob_handler_new() failed.");
 		goto cd_fail;
```
lyakh, line 463: do we need to set an error code here too?

src/audio/igo_nr/igo_nr.c
```diff
@@ -275,124 +277,96 @@ static inline int32_t set_capture_func(struct comp_dev *dev)
...
 	}
 
 	/* Get configuration data */
-	ret = comp_init_data_blob(cd->model_handler, bs, ipc_igo_nr->data);
+	ret = comp_init_data_blob(cd->model_handler, bs, cfg->data);
 	if (ret < 0) {
-		comp_cl_err(&comp_igo_nr, "igo_nr_new(): comp_init_data_blob() failed.");
+		comp_err(dev, "igo_nr_init(): comp_init_data_blob() failed.");
 		goto cd_fail;
```
lyakh, line 470: and here

src/audio/igo_nr/igo_nr.c
```diff
@@ -409,24 +383,6 @@ static int32_t igo_nr_params(struct comp_dev *dev,
...
-
-	/* Use empirical add +10 for target frames number to avoid xruns and
-	 * distorted audio in beginning of streaming. It's slightly more than
-	 * min. needed and does not increase much peak load and buffer memory
-	 * consumption. The copy() function will find the limits and process
-	 * less frames after the buffers levels stabilize.
-	 */
-	cd->source_frames_max = cd->source_frames + 10;
-	cd->sink_frames_max = cd->sink_frames + 10;
-	cd->frames = MAX(cd->source_frames_max, cd->sink_frames_max);
```
lyakh, line 424: all the above parameters aren't set anywhere any more? If they're now unneeded - should they be removed completely?

src/include/sof/audio/igo_nr/igo_nr_comp.h
```diff
@@ -35,11 +35,6 @@ struct comp_data {
 	uint32_t source_rate;	/* Sample rate in Hz */
 	uint32_t sink_format;	/* For used PCM sample format */
 	uint32_t source_format;	/* For used PCM sample format */
-	int32_t source_frames;	/* Nominal # of frames to process at source */
-	int32_t sink_frames;	/* Nominal # of frames to process at sink */
-	int32_t source_frames_max;	/* Max # of frames to process at source */
-	int32_t sink_frames_max;	/* Max # of frames to process at sink */
-	int32_t frames;		/* IO buffer length */
```
lyakh, line 42: aha, this explains that change in an earlier commit now...

src/audio/igo_nr/igo_nr.c
```diff
@@ -69,179 +69,316 @@ static void igo_nr_lib_process(struct comp_data *cd)
...
+				sample = *x;
+				if (k == cd->config.active_channel_idx)
+					cd->in[i] = sample;
+				else
+					*y1 = sample;
+
+				x++;
+				y1++;
+			}
+			i++;
```
lyakh, line 144: is this also just for the source-sink conversion or is this also changing the algorithm?

singalsu: There was no way to keep the (very inefficient) read/write frag per sample code with source/sink API so I had to rewrite these parts. The performance with stub improved with tens of MCPS as side effect.

src/audio/igo_nr/igo_nr.c
```diff
@@ -275,124 +277,96 @@ static inline int32_t set_capture_func(struct comp_dev *dev)
...
 		goto cd_fail;
 	}
 
-	comp_set_drvdata(dev, cd);
-
 	/* Handler for configuration data */
 	cd->model_handler = comp_data_blob_handler_new(dev);
 	if (!cd->model_handler) {
-		comp_cl_err(&comp_igo_nr, "igo_nr_new(): comp_data_blob_handler_new() failed.");
+		comp_err(dev, "igo_nr_init(): comp_data_blob_handler_new() failed.");
```
btian1, line 462: here for goto cd_fail, need free cd-=>p_handle as well, i.e, cd_fail should have 2 or more stage error handling.

src/audio/igo_nr/igo_nr.c
singalsu: Yep

src/audio/igo_nr/igo_nr.c
singalsu: Yep, I think three levels are needed.

src/audio/igo_nr/igo_nr.c
singalsu: Yep

src/audio/igo_nr/igo_nr.c
singalsu: Yes, it's done

src/include/sof/audio/igo_nr/igo_nr_comp.h
singalsu: Yep!

src/audio/igo_nr/igo_nr.c
```diff
@@ -275,124 +277,104 @@ static inline int32_t set_capture_func(struct comp_dev *dev)
...
+	/* update downstream (playback) or upstream (capture) buffer parameters */
+	mod->verify_params_flags = BUFF_PARAMS_RATE;
+
+	comp_info(dev, "igo_nr created");
 
-	return dev;
+	return 0;
+
+cd_fail3:
+	comp_data_blob_handler_free(cd->model_handler);
```
lyakh, line 483: this is the same as 6 lines below

src/include/sof/audio/igo_nr/igo_nr_comp.h
lyakh: but can this be merged into that commit - the first one in the series?

src/audio/igo_nr/igo_nr.c
```diff
@@ -677,52 +812,50 @@ static void igo_nr_lib_init(struct processing_module *mod)
 }
 
 #if CONFIG_IPC_MAJOR_4
-static void igo_nr_ipc4_params(struct processing_module *mod)
+static  int igo_nr_ipc4_params(struct processing_module *mod, struct sof_source *source,
```
lyakh, line 815: one space, please

src/audio/igo_nr/igo_nr.c
```diff
@@ -558,84 +484,52 @@ static int32_t igo_nr_set_chan(struct comp_dev *dev, struct sof_ipc_ctrl_data *c
...
 	}
 }
 
-static void igo_nr_process(struct comp_dev *dev,
-			   struct comp_buffer *source,
-			   struct comp_buffer *sink,
-			   struct comp_copy_limits *cl,
-			   int32_t frames)
-
+static void igo_nr_print_config(struct processing_module *mod)
```
btian1, line 699: this is an igo_nr specific function, possible to remove and add specific print to each place?

src/audio/igo_nr/igo_nr.c
```diff
@@ -669,64 +563,59 @@ static void igo_nr_print_config(struct comp_dev *dev)
...
-	sinkb = list_first_item(&dev->bsink_list, struct comp_buffer,
-				source_list);
-
 	/* Check for changed configuration */
 	if (comp_is_new_data_blob_available(cd->model_handler))
-		igo_nr_set_igo_params(dev);
-
-	/* Get source, sink, number of frames etc. to process. */
-	comp_get_copy_limits(sourceb, sinkb, &cl);
+		igo_nr_set_igo_params(mod);
```
btian1, line 779: possible put set params to set_config? this is time critical place

src/audio/igo_nr/igo_nr_stub.c
```diff
@@ -6,8 +6,15 @@
 
 #include <sof/audio/igo_nr/igo_nr_comp.h>
 
+/* Set handle_size to Just something instead of 0 from rzalloc() of
+ * component data to avoid rballoc() in igo_nr_init() to fail.
+ */
+#define IGO_NR_STUB_HANDLE_SIZE		42
```
btian1, line 12: but at that PR, you still use 42, instead of this macro. or info->handle_size

btian1 (commented):
that's definitely not one PR work, right? title is convert to module interface, inside PR, at least 3 tasks are done, do you want to split it to several PRs? for better review/merge/track?

singalsu:
> that's definitely not one PR work, right? title is convert to module interface, inside PR, at least 3 tasks are done, do you want to split it to several PRs? for better review/merge/track?

The problem is that the component didn't work even with IPC3. It's impossible to test without all the changes included. Since this is a closed component and not useful without licensing it, I'd like to pass this with small effort to make this testable again. The owner or licencors of this should do the extra polishing effort.

lgirdwood:
> > that's definitely not one PR work, right? title is convert to module interface, inside PR, at least 3 tasks are done, do you want to split it to several PRs? for better review/merge/track?
> 
> The problem is that the component didn't work even with IPC3. It's impossible to test without all the changes included. Since this is a closed component and not useful without licensing it, I'd like to pass this with small effort to make this testable again. The owner or licensors of this should do the extra polishing effort.

Ack - the aim here is not to change functionality or improve performance, but to do the mechanical work to bring it to the module API. Once at module API, it can then be integrated by folks who work on this module.

src/audio/igo_nr/igo_nr.c
singalsu: Oops, yes.

src/audio/igo_nr/igo_nr.c
```diff
@@ -672,52 +807,50 @@ static void igo_nr_lib_init(struct processing_module *mod)
...
+	if (ret)
+		return ret;
 
-	sourceb = list_first_item(&dev->bsource_list, struct comp_buffer, sink_list);
-	ipc4_update_buffer_format(sourceb, &mod->priv.cfg.base_cfg.audio_fmt);
-
-	sinkb = list_first_item(&dev->bsink_list, struct comp_buffer, source_list);
-	ipc4_update_buffer_format(sinkb, &mod->priv.cfg.base_cfg.audio_fmt);
+	ret = sink_set_params(sink, params, true);
+	return ret;
```
lyakh, line 824: `return sink_set_params(sink, params, true);`

src/audio/igo_nr/igo_nr_stub.c
btian1: I forgot as well, let's close this conversasion.

src/audio/igo_nr/igo_nr.c
```diff
@@ -69,179 +69,316 @@ static void igo_nr_lib_process(struct comp_data *cd)
 	}
 }
 
+static void *igo_cir_wrap(void *ptr, void *cir_buf_end, size_t cir_buf_size)
```
btian1, line 72: this comment is for title, what's the sample rate for igo_nr? due to I saw the frame size set to 768, if it is 16ms, means only support 48k input?

src/audio/igo_nr/igo_nr.c
```diff
@@ -69,179 +69,316 @@ static void igo_nr_lib_process(struct comp_data *cd)
...
+{
+	if (ptr > cir_buf_end)
+		ptr = (char *)ptr - cir_buf_size;
+
+	return ptr;
+}
```
btian1, line 78: static inline void *cir_buf_wrap(void *ptr, void *buf_addr, void *buf_end)
{
	if (ptr >= buf_end)
		ptr = (char *)buf_addr +
			((char *)ptr - (char *)buf_end);

	assert((intptr_t)ptr <= (intptr_t)buf_end);

	return ptr;
}
 why not directly use this?

src/audio/igo_nr/igo_nr.c
btian1: so igo_nr already tested and verified at least with testbench?

src/audio/igo_nr/igo_nr.c
```diff
@@ -69,179 +69,316 @@ static void igo_nr_lib_process(struct comp_data *cd)
...
+				y1++;
+			}
+			i++;
 		}
 
-		x = audio_stream_read_frag_s16(source, idx_in + cd->config.active_channel_idx);
-		cd->in[i] = (*x);
-
-		idx_in += nch;
+		x = igo_cir_wrap(x, x_end, x_size);
```
btian1, line 147: based on Marcin's pipe 2.0, at least, input is linear, output maybe not(if it is the last module).

tools/rimage/config/lnl.toml
```diff
@@ -56,7 +56,7 @@ name = "ADSPFW"
 load_offset = "0x40000"
 
 [module]
-count = 25
+count = 26
```
btian1, line 59: maybe this patch need on hold for a while and wait for: #8490 merge, after that, you only need add it to src/audio/igo_nr folder.
currently, this is duplicated.

lgirdwood:
@lyakh @btian1 we only need to focus review on the changes to use the module API. This PR is not about improving the existing IGO code as we need to minimize the change surface here to module API integration only. So yes, there are lots of ways the IGO code can be improved, but we need to get integration with module API done before any other code improvements. Same goes for all 3P modules integrations to module API.

singalsu:
I rebased and fixed rimage related conflicts. This version sounds choppy in playback test (sof-hda-benchmark-igo_nr32-tgl.tplg).

src/audio/igo_nr/igo_nr.c
singalsu: Seems I have fix this.

src/audio/igo_nr/igo_nr.c
singalsu: Best to keep the classification in components as-is.  The owner of the component wanted to keep these traces as debug level.  We should fix our trace if debug traces can no more be activated from trace tool.

src/audio/igo_nr/igo_nr.c
```diff
@@ -676,59 +563,54 @@ static void igo_nr_print_config(struct comp_dev *dev)
 
 }
 
-static void igo_nr_set_igo_params(struct comp_dev *dev)
+static void igo_nr_set_igo_params(struct processing_module *mod)
```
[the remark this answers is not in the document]
singalsu: Yes it woud be better, but sorry, I don't want to change original function name. This patch and PR is for module adapter conversion. The owner should do this kind of improvements.

src/audio/igo_nr/igo_nr.c
singalsu: No, let's not remove and not change the feature. The IPC4 handler should handle the binary blob as ipc3 does.

src/audio/igo_nr/igo_nr.c
```diff
@@ -437,6 +441,7 @@ static int igo_nr_get_config(struct processing_module *mod,
 {
 	struct comp_dev *dev = mod->dev;
 
+#if CONFIG_IPC_MAJOR_3
```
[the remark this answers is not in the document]
singalsu: OK, I'll merge this patch with previous.

src/audio/igo_nr/igo_nr.c
```diff
@@ -519,6 +536,26 @@ static int igo_nr_set_config(struct processing_module *mod, uint32_t param_id,
...
 		return -EINVAL;
 	}
+#elif CONFIG_IPC_MAJOR_4
+	struct sof_ipc4_control_msg_payload *ctl = (struct sof_ipc4_control_msg_payload *)fragment;
+
+	switch (param_id) {
+	case SOF_IPC4_SWITCH_CONTROL_PARAM_ID:
+		comp_info(dev, "igo_nr_cmd_set_config(), SOF_IPC4_SWITCH_CONTROL_PARAM_ID");
+		return igo_nr_set_chan(mod, ctl);
+		return 0;
```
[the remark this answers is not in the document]
singalsu: Yep!

src/audio/igo_nr/igo_nr.c
singalsu: The topologies in topology1 are for 48 kHz. The only clue in code is the `#define IGO_FRAME_SIZE (768)` that would be equivalent to 16 ms.

src/audio/igo_nr/igo_nr.c
singalsu: I don't see issues in what I proposed.

src/audio/igo_nr/igo_nr.c
singalsu: I have run it recently in UPX device. Listened music trough it, no glitches heard. Since TGL does not support DP mode I have had a patch to change it to operate with LL for 48 size frame.

src/audio/igo_nr/igo_nr.c
singalsu: I need this with my tests with TGL with LL pipelines. Let's remove circular capability when we know it's safe to do and the condition you mentioned is acceptable. The overhead from this is wrap minimal.

src/audio/igo_nr/igo_nr.c
singalsu: OK

tools/rimage/config/lnl.toml
singalsu: It's merged, the modularize patch is only for MTL.

src/audio/igo_nr/igo_nr.c
```diff
@@ -458,11 +463,22 @@ static int igo_nr_get_config(struct processing_module *mod,
...
 }
 
 static int32_t igo_nr_set_chan(struct processing_module *mod, void *data)
 {
+#if CONFIG_IPC_MAJOR_3
 	struct sof_ipc_ctrl_data *cdata = data;
+#elif CONFIG_IPC_MAJOR_4
+	struct sof_ipc4_control_msg_payload *cdata = data;
+#else
+#error "CONFIG_IPC_MAJOR_n is not set"
```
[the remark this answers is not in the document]
singalsu: Yep, I removed.

src/audio/igo_nr/igo_nr.c
```diff
@@ -519,6 +536,26 @@ static int igo_nr_set_config(struct processing_module *mod, uint32_t param_id,
...
+		comp_info(dev, "igo_nr_cmd_set_config(), SOF_IPC4_SWITCH_CONTROL_PARAM_ID");
+		return igo_nr_set_chan(mod, ctl);
+		return 0;
+	case SOF_IPC4_ENUM_CONTROL_PARAM_ID:
+		comp_err(dev, "igo_nr_set_config(), illegal control.");
+		return -EINVAL;
+	}
+
+	comp_info(dev, "igo_nr_cmd_set_config(), bytes");
+	ret = comp_data_blob_set(cd->model_handler, pos, data_offset_size, fragment, fragment_size);
```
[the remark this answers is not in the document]
singalsu: This gets executed with IPC4,  here's mtrace from running playback in UPX device:

```
[187056387392.556303] <inf> igo_nr: igo_nr_set_config: comp:0 0x16 igo_nr_set_config()
[187056387392.556311] <inf> igo_nr: igo_nr_set_config: comp:0 0x16 igo_nr_cmd_set_config(), bytes
[187056387392.572400] <inf> igo_nr: igo_nr_check_params: comp:0 0x16 igo_nr_check_params()
[187056387392.572410] <inf> igo_nr: igo_nr_check_params: comp:0 0x16 igo_nr_check_params(), sample rate = 48000
[187056387392.572420] <inf> igo_nr: igo_nr_set_igo_params: comp:0 0x16 igo_nr_set_igo_params()
[187056387392.572428] <inf> igo_nr: igo_nr_set_igo_params: comp:0 0x16 New config detected.


```

src/audio/igo_nr/igo_nr.c
btian1: check https://github.com/thesofproject/sof/blob/main/src/include/sof/audio/audio_stream.h#L414
you don't need define a new one.

btian1 (commented):
LGTM, only one open left.

src/audio/igo_nr/igo_nr.c
singalsu: OK, I've changed and tested it now - pushing new version!

btian1 (approved):

src/audio/igo_nr/igo_nr.c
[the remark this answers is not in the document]
singalsu: Oops, wasn't fixed. Done now.

lgirdwood (approved):

Issue #8265, opened 2023-09-26
[BUG] Unable to load tplg with mixer or DRC component


iuliana-prodan:
**Describe the bug**
Unable to load tplg with mixer or DRC component.
This is the dmesg log:

```
[   13.541416] platform sof-sound-wm8960: deferred probe pending
[   13.709372]     fdt_file=RODEV--MX8MP_evk--1_imx8mp-evk-sof-wm8960.dtb
[   20.411653] sof-audio-of-imx8m 3b6e8000.dsp: DT DSP detected
[   20.419352] sof-audio-of-imx8m 3b6e8000.dsp: sof_debug value: 0x1
[   20.568582] sof-audio-of-imx8m 3b6e8000.dsp: unknown sof_ext_man header type 3 size 0x30
[   20.584024] sof-audio-of-imx8m 3b6e8000.dsp: Firmware info: version 2:7:99-cbc04
[   20.599106] sof-audio-of-imx8m 3b6e8000.dsp: Firmware: ABI 3:28:0 Kernel ABI 3:23:0
[   20.621005] sof-audio-of-imx8m 3b6e8000.dsp: Firmware info: version 2:7:99-cbc04
[   20.628419] sof-audio-of-imx8m 3b6e8000.dsp: Firmware: ABI 3:28:0 Kernel ABI 3:23:0
[   20.763141] sof-audio-of-imx8m 3b6e8000.dsp: Topology: ABI 3:27:0 Kernel ABI 3:23:0
[   20.773756] asoc-simple-card sof-sound-wm8960: ASoC: Parent card not yet available, widget card binding deferred
[   20.789377] sof-audio-of-imx8m 3b6e8000.dsp: tplg: config SAI3 fmt 0x1 mclk 12288000 width 32 slots 2 mclk id 0
[   20.800647] sof-audio-of-imx8m 3b6e8000.dsp: tplg: config SAI3 fmt 0x1 mclk 12288000 width 32 slots 2 mclk id 0
[   20.817712] sof-audio-of-imx8m 3b6e8000.dsp: ipc tx error for 0x30010000 (msg/reply size: 224/0): -22
[   20.826973] sof-audio-of-imx8m 3b6e8000.dsp: Failed to setup widget DRC1.0
[   20.834102] sof-audio-of-imx8m 3b6e8000.dsp: error: tplg component load failed -22
[   20.845282] sof-audio-of-imx8m 3b6e8000.dsp: error: failed to load DSP topology -22
[...]

PR #8303, opened 2023-10-10
Revert "module_adapter: avoid module init crash in case of ipc data invalid"

iuliana-prodan:
Revert "module_adapter: avoid module init crash in case of ipc data invalid"

This reverts 'commit https://github.com/thesofproject/sof/commit/e847c8b270150353551d4720e382502c9a916496 ("module_adapter: avoid module
init crash in case of ipc data invalid")'.

No data is not an invalid case for mixer, for example.

Fixes: https://github.com/thesofproject/sof/issues/8265
Signed-off-by: Iuliana Prodan <iuliana.prodan@nxp.com>

### DIFF
--- a/src/audio/module_adapter/module_adapter_ipc3.c
+++ b/src/audio/module_adapter/module_adapter_ipc3.c
@@ -79,8 +79,6 @@ int module_adapter_init_data(struct comp_dev *dev,
 			return ret;
 		}
 		dst->init_data = dst->data;
-	} else {
-		return -EINVAL;
 	}
 
 	return 0;


src/audio/module_adapter/module_adapter_ipc3.c
```diff
@@ -79,6 +79,9 @@ int module_adapter_init_data(struct comp_dev *dev,
 			return ret;
 		}
 		dst->init_data = dst->data;
+	} else if (size == 0) {
+		/* valid case for mixer */
+		return 0;
```
btian1, line 84: if this is the case, do we still need keep line 85 - 87?

kv2019i: @btian1 true, size can't be negative. So maybe we just do a full revert.

It does open questions how to address the original commit e847c8b270150353551d4720e382502c9a916496 . Do we need a SRC specific patch?

kv2019i (commented):
I'd suggest a revert and then do a follow-up to make a new fix to avoiding module init crashes (if a generic module level solution is possible, not sure it is).

src/audio/module_adapter/module_adapter_ipc3.c
iuliana-prodan: I agree @kv2019i , I'll do a full revert of this https://github.com/thesofproject/sof/commit/e847c8b270150353551d4720e382502c9a916496, but the revert doesn't apply cleanly now because of the ipc3/ipc4 split.

I've kept the check for negative size to not break something on Intel targets - even though the explanation for adding that commit is limited.

dbaluta (approved):

kv2019i (approved):
Looks good.  @btian1 can you do a follow-up?

btian1 (approved):

btian1:
> an you do a follow-up?



> Looks good. @btian1 can you do a follow-up?

https://github.com/thesofproject/sof/commit/e847c8b270150353551d4720e382502c9a916496, for me this is a safeguard to avoid module init crash, I remember there is another patch, I forgot the commit name to fix similar issue.
let's merge this first, then check if have any further issues.

PR #8304, opened 2023-10-11
topolog2: nocodec: set output_pin number

RanderWang:
The num_output_pins was set for pass-through path but missed in normal path. Without it Linux kernel driver wouldn't set second output pin format for dai copier which connects smart amp and gain module.

### DIFF
--- a/tools/topology/topology2/cavs-nocodec.conf
+++ b/tools/topology/topology2/cavs-nocodec.conf
@@ -375,6 +375,7 @@ IncludeByKey.PASSTHROUGH {
 						out_valid_bit_depth	32
 					}
 
+					num_output_pins 2
 					Object.Base.output_pin_binding.1 {
 						output_pin_binding_name "gain.8.1"
 					}
--- a/tools/topology/topology2/cavs-nocodec.conf
+++ b/tools/topology/topology2/cavs-nocodec.conf
@@ -307,48 +307,67 @@ IncludeByKey.PASSTHROUGH {
 						}
 						Object.Widget.module-copier.1 {
 							stream_name 'Gain Capture 19'
+
 							num_input_audio_formats 2
+							Object.Base.input_audio_format [
+								{
+									in_bit_depth		32
+									in_valid_bit_depth	32
+								}
+								{
+									in_channels		4
+									in_bit_depth		32
+									in_valid_bit_depth	32
+									in_ch_cfg	$CHANNEL_CONFIG_3_POINT_1
+									in_ch_map	$CHANNEL_MAP_3_POINT_1
+								}
+							]
+
 							num_output_audio_formats 2
-							Object.Base.audio_format.1 {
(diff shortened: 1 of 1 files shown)


tools/topology/topology2/cavs-nocodec.conf
```diff
@@ -375,6 +375,7 @@ IncludeByKey.PASSTHROUGH {
 						out_valid_bit_depth	32
 					}
 
+					num_output_pins 2
```
plbossart, line 378: we should also clean-up the audio format part @RanderWang, we need to have different structures for in and out. And I also don't get why the format should be the same on all pins?

RanderWang:
> is there a link to the issue this PR fixes?

No. It is found in loadable smart amp validation. I also check the built-in smart amp. The format setting by set_large_config is also missed for this reason.

RanderWang:
Clean up audio format in cavs-nocodec

tools/topology/topology2/cavs-nocodec.conf
RanderWang: Add clean-up.  @aiChaoSONG  can you help to answer pin format question ?

kv2019i (approved):
Not sure it's good idea to couple a bugfix and cleanup of audio formatting in same PR, but given you already did the work, it seems good.

tools/topology/topology2/cavs-nocodec.conf
```diff
@@ -307,45 +307,54 @@ IncludeByKey.PASSTHROUGH {
...
-							num_output_audio_formats 2
-							Object.Base.audio_format.1 {
+							Object.Base.input_audio_format.1 {
 								in_bit_depth		32
 								in_valid_bit_depth	32
-								out_bit_depth		32
-								out_valid_bit_depth	32
 							}
-							Object.Base.audio_format.2 {
+							Object.Base.input_audio_format.2 {
```
ranj063, line 316: @RanderWang can you please avoid the .1/.2 and use array instead?

RanderWang:
> Not sure it's good idea to couple a bugfix and cleanup of audio formatting in same PR, but given you already did the work, it seems good.

Good idea, thanks!

tools/topology/topology2/cavs-nocodec.conf
RanderWang: sure, thanks

RanderWang:
use audio format array, thanks!

ranj063 (approved):

plbossart (commented):
I'll let @ranj063 review, and still don't get why the formats should be the same on all pins.

tools/topology/topology2/cavs-nocodec.conf
ranj063: @plbossart both the gain and the smart amp only support 32-bit in this case.

PR #8305, opened 2023-10-11
topology2: sof-mtl-rt713-l0-rt1316-l12-rt1713-l3: use default stream name

bardliao:
rt713 and rt1713 are in different SoundWire links. So, no DAI type will be appended to DAI link name. We should use default stream name in the topology.


plbossart (approved):

kv2019i (approved):

kv2019i:
One MTL DUT (HDA) device fails here (similar to https://github.com/thesofproject/sof/pull/8306 ). @fredoh9 @keqiaozhang what's going on? This PR is changing an unrelated tplg, so cannot be connected. 

I'll proceed with merge on this one, but we need to fix the DUT issue.

PR #8308, opened 2023-10-11
topology2: google-rtc-aec: expose component core_id for override

yongzhi1:
RTC AEC module is using DP and rest of echo ref pipeline is LL, and both can run on secondary cores.
In order to assign RTC_AEC DP alone on secondary core, add new macro GOOGLE_AEC_DP_CORE_ID.

### DIFF
--- a/tools/topology/topology2/cavs-rt5682.conf
+++ b/tools/topology/topology2/cavs-rt5682.conf
@@ -96,7 +96,10 @@ Define {
 	BT_PB_DAI_PIPELINE_SRC "copier.host.9.1"
 	BT_PB_PIPELINE_STREAM_NAME "dai-copier.SSP.10.1"
 	GOOGLE_RTC_AEC_SUPPORT		0
+	# assign core for AEC LL pipelines
 	GOOGLE_AEC_CORE_ID		0
+	# assign core for RTC_AE DP task
+	GOOGLE_AEC_DP_CORE_ID		0
 	HEADSET_PCM_NAME		"Headset"
 	HEADSET_PCM_ID			0
 	SPEAKER_PCM_NAME		"Speakers"
--- a/tools/topology/topology2/platform/intel/google-rtc-aec-reference.conf
+++ b/tools/topology/topology2/platform/intel/google-rtc-aec-reference.conf
@@ -10,6 +10,7 @@ Object.Pipeline.google-rtc-aec-capture [
 		}
 
 		Object.Widget.google-rtc-aec.1 {
+			core_id $GOOGLE_AEC_DP_CORE_ID
 			Object.Base.input_pin_binding.1 {
 				input_pin_binding_name	"module-copier.18.1"
 			}


lgirdwood:
@yongzhi1 I guess we need other PRs for FW and SW from @marcinszkudlinski and @ranj063 before we can merge ?

yongzhi1:
> @yongzhi1 I guess we need other PRs for FW and SW from @marcinszkudlinski and @ranj063 before we can merge ?

Yes, that's right. The DP AEC support has dependency on https://github.com/thesofproject/sof/pull/8403; https://github.com/thesofproject/linux/pull/4664 and other fix.

tools/topology/topology2/cavs-rt5682.conf
```diff
@@ -95,7 +95,10 @@ Define {
 	BT_PB_DAI_PIPELINE_SRC "copier.host.9.1"
 	BT_PB_PIPELINE_STREAM_NAME "dai-copier.SSP.10.1"
 	GOOGLE_RTC_AEC_SUPPORT		0
+	# assign AEC LL pipelines to secondary core
 	GOOGLE_AEC_CORE_ID		0
+	# assign RTC_AE DP task to secondary core
+	GOOGLE_AEC_DP_CORE_ID		0
```
kv2019i, line 101: Hmm, @yongzhi1 , I don't understand the comment. "assign to secondary core" but code sets CORE_ID to 0...?

kv2019i (changes requested):

tools/topology/topology2/cavs-rt5682.conf
yongzhi1: Ack, maybe better to say "override in tplg-targets.cmake to run on secondary core"?

yongzhi1:
Update:

Fix the comment to avoid confusion.

kv2019i (approved):
Thanks @yongzhi1 !

tools/topology/topology2/cavs-rt5682.conf
```diff
@@ -95,7 +95,10 @@ Define {
 	BT_PB_DAI_PIPELINE_SRC "copier.host.9.1"
 	BT_PB_PIPELINE_STREAM_NAME "dai-copier.SSP.10.1"
 	GOOGLE_RTC_AEC_SUPPORT		0
+	# assign core for AEC LL pipelines
 	GOOGLE_AEC_CORE_ID		0
```
marcinszkudlinski, line 99: Do we need to run AEC as LL?

kv2019i: @marcinszkudlinski  I think this is for the rest of the pipeline on which AEC is.

yongzhi1: yes, for example, as a test we can run echo ref related pipelines on core 2, and put RTC_AEC module alone on core 1:

lgirdwood: ok, so comment is a little confusing. We should update it to say AEC is using DP and rest of pipeline is LL and both are on different cores.

yongzhi1: Ack, will update commit message as well as the PR description, thanks,

yongzhi1:
Update:
Improved commit description based on reviews.

lgirdwood (approved):

PR #8314, opened 2023-10-12
[stable-v2.2] topology1: Use DYNAMIC for RPL SDW topologies

bardliao:
Use dynamic for all the rpl SDW topologies.

I am not sure if this should be applied to main branch, too.


kv2019i (approved):
No need to submit to main, we'll soon remove the Intel tplg1 files from main.

Dynamic tplg loading has been used for long on Chromebooks already, so makes sense to take this into use for recent platforms here as well.

plbossart:
Maybe I am too conservative but I've recently seen people bisecting issues all the way to 5.10. If we have a topology feature that requires 5.19+ then we will break this sort of effort.

There's no good solution I am afraid. Either we introduce some issues for a small minority or we don't test what we release. Which is worse?

bardliao:
> Maybe I am too conservative but I've recently seen people bisecting issues all the way to 5.10. If we have a topology feature that requires 5.19+ then we will break this sort of effort.
> 
> There's no good solution I am afraid. Either we introduce some issues for a small minority or we don't test what we release. Which is worse?

But 5.10 doesn't support RPL boards, right? There is no `soc-acpi-intel-rpl-match.c` in v5.10 kernel. `soc-acpi-intel-rpl-match.c` is first seen in 6.1

kv2019i:
@plbossart We had discussion on this with Bard before he submitted this and we touched on this dilemma as well. But yeah, a blanket move to DYNAMIC topologies (and deep-buffer) does seem too risky, but partial introduction seems safer. E.g. RPL, I checked both 5.10.y and 5.15.y stable series, and there's no RPL ACPI match file. So any RPL SDW machine needs a newer kernel anyways.

plbossart (approved):
ok for RPL only changes

PR #8316, opened 2023-10-12
[mtl-006] Add loadable smart amp support

RanderWang:
This PR is developed by Guennadi and I add some fixup to make it work with mtl-006 branch


src/include/sof/audio/module_adapter/module/modules.h
```diff
@@ -60,6 +60,15 @@ do { \
...
 	(comp_dynamic_module)->ops.get_attribute = module_adapter_get_attribute; \
+	(comp_dynamic_module)->ops.bind = module_adapter_bind;		\
+	(comp_dynamic_module)->ops.unbind = module_adapter_unbind;	\
+	(comp_dynamic_module)->ops.get_total_data_processed = module_adapter_get_total_data_processed; \
+	(comp_dynamic_module)->ops.dai_get_hw_params = module_adapter_get_hw_params; \
+	(comp_dynamic_module)->ops.position = module_adapter_position;	\
+	(comp_dynamic_module)->ops.dai_ts_config = module_adapter_ts_config_op;	\
+	(comp_dynamic_module)->ops.dai_ts_start = module_adapter_ts_start_op; \
+	(comp_dynamic_module)->ops.dai_ts_stop = module_adapter_ts_stop_op; \
+	(comp_dynamic_module)->ops.dai_ts_get = module_adapter_ts_get_op; \
```
pjdobrowolski, line 71: Please remove unused ops in loadable modules. There is no need for using them.

src/audio/module_adapter/library/native_system_service.c
```diff
@@ -128,12 +128,21 @@ AdspErrorCode native_system_service_get_interface(adsp_iface_id id, system_servi
...
 	.vec_memset = native_system_service_vec_memset,
 	.notification_create = native_system_service_create_notification,
 	.notification_send = native_system_service_send_notif_msg,
-	.get_interface = native_system_service_get_interface
+	.get_interface = native_system_service_get_interface,
+	.comp_verify_params = comp_verify_params,
+	.math_gcd = gcd,
+	.data_blob_handler_new = comp_data_blob_handler_new_ext,
+	.data_blob_handler_free = comp_data_blob_handler_free,
+	.data_blob_set = comp_data_blob_set,
```
pjdobrowolski, line 147: Please remove unused services, there is no call of them in module.

pjdobrowolski (changes requested):

src/audio/module_adapter/library/native_system_service.c
RanderWang: it is used in smart_amp module
````c
static int smart_amp_init(struct processing_module *mod)
{
        ......
	/* component model data handler */
	sad->model_handler = mod->sys_service->data_blob_handler_new(dev);
	if (!sad->model_handler) {
		ret = -ENOMEM;
		goto sad_fail;
	}
    .....
}
````

File: src/include/sof/audio/module_adapter/module/modules.h (hunk 1 above)
Comment (RanderWang) [reply src/include/sof/audio/module_adapter/module/modules.h]: sure, removed

**Comment by RanderWang** [comment]
remove unused ops. Thanks!

File: lmdk/include/component.h [hunk 3]
```diff
@@ -537,15 +537,18 @@ struct comp_driver_info {
...
-	uint32_t periods_source;/**< 0 means variable */
-	uint32_t frame_fmt;	/**< SOF_IPC_FRAME_ */
-	uint32_t xrun_action;	/**< action we should take on XRUN */
+	uint32_t periods_sink;		/**< 0 means variable */
+	uint32_t periods_source;	/**< 0 means variable */
+	uint32_t frame_fmt;		/**< SOF_IPC_FRAME_ */
+	uint32_t xrun_action;		/**< action we should take on XRUN */
+#if CONFIG_IPC_MAJOR_4
+	uint32_t ipc_config_size;	/**< size of a config received by ipc */
+#endif
```
btian1, line 551: please add more commit message for this change, there are multiple changes in this patch, and not sure it is copied from other source file or not.

src/audio/module_adapter/library/native_system_service.c
pjdobrowolski: This way we are mixing module api with component api. System services are module api only.

lmdk/include/buffer.h
```diff
@@ -140,7 +140,7 @@ struct comp_buffer {
 	uint32_t pipeline_id;
 	uint32_t caps;
 	uint32_t core;
-//	struct tr_ctx tctx;			/* trace settings */
+	uint32_t padd[2];
```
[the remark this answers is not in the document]
RanderWang: long story.  loadable module uses different header file compared to base module. The different header files define different structure which cause the same filed with different offset in structure, so the data shared between base module and loadable module will be out of order. Add padd[2] to make sure the offset is correct.

src/audio/module_adapter/library/native_system_service.c
pjdobrowolski: @lyakh, We must discuss in wider audience how system services work and how to make smart amp work in such a narrow environment. System services just expose some interfaces and we cannot expand them anytime when we need it. Plus mixing compnen api with module api is no go.

lmdk/include/sink_api_implementation.h
```diff
@@ -73,6 +73,7 @@ struct sof_sink {
 	const struct sink_ops *ops;	/** operations interface */
 	size_t requested_write_frag_size; /** keeps number of bytes requested by get_buffer() */
 	size_t num_of_bytes_processed; /** processed bytes counter */
+	size_t obs;
```
marcinszkudlinski, line 76: better use last implementation - IBS/OBS has been replaced by more meaningful names min_free_space / min_available

lmdk/include/buffer.h
marcinszkudlinski, line 143: I don't get it, are you trying to have 2 copies of the same structure, with some fields as padds? 
Maintenance of this will be extremely hard

src/audio/module_adapter/library/native_system_service.c
marcinszkudlinski: i'm not sure if this is a case, but in IPC4 blobs should be passed by large_config_set. In case of huge blobs several conf_sets should be sent and re-assembled in the module

lmdk/include/audio_stream.h
```diff
@@ -31,6 +33,36 @@
  *  @{
  */
 
+/**
```
marcinszkudlinski, line 36: again, 2 copies of the same - I don;t see any way to keep them consistant

marcinszkudlinski (changes requested):

src/audio/module_adapter/library/native_system_service.c
mengdonglin: > This way we are mixing module api with component api. System services are module api only.

@ranj063 Can you please comment here? Most FW processing modules need to handle data blobs. Is there a module API function for this need?

lmdk/include/sink_api_implementation.h
lgirdwood: We should probably also string change that change in code, manifest and docs too.

src/audio/module_adapter/library/native_system_service.c
mengdonglin: > i'm not sure if this is a case, but in IPC4 blobs should be passed by large_config_set. In case of huge blobs several conf_sets should be sent and re-assembled in the module

@marcinszkudlinski @lgirdwood @ranj063 
Module API provides large_config_set API for host to pass blobs to the module. Now the questions is that how the module store and convert the huge blobs via several conf_sets to its internal configuration data blob or structure. Built-in modules can leverage blob handler provided by the base FW. But for loadable modules, is there a general way to do so? If there is no general implementation solution, each loadable module needs to implement by itself.

lmdk/include/sink_api_implementation.h
RanderWang: this change is based on MTL-006 branch, so it is copier from base fw part.  When Guennadi come back, he will do it for main branch.

lmdk/include/buffer.h
RanderWang: ideally both base FW and loadable moule use the same structure definition so that they can decode structure correctly. But Now loadable module remove above tctx for trace. This result to inconsistency of the the structure and decoding can't be correct.  Add padd data to avoid structure size and internal offset change.

src/audio/module_adapter/library/native_system_service.c
RanderWang: blob function is a utility and used for module config saving in module initialization function here, so it is not related to large_config_set.

lyakh: data_blob.c provides a library for receiving and sending of large component (or module) configuration sets as well as for run-time querying of that configuration data by component or module drivers. It belongs to the "audio component driver" infrastructure but currently it is used by module adapter drivers too. Its dependency on the component driver API is its use of a `.dev` pointer in `struct comp_data_blob_handler`. That pointer is used for 3 purposes:
1. logging in calls to `comp_dbg(dev,...)`, `comp_warn(dev,...)`, etc.
2. component status checks via `dev->state`
3. calls to `struct comp_data_blob_handler::validator(dev,...)`

The third item is the easiest to eliminate - `.validator` is a member of `struct comp_data_blob_handler` and its prototype can easily be changed to
```
int (*validator)(struct comp_data_blob_handler *blob_handler, void *new_data, uint32_t new_data_size);
```
Items (1) and (2) so far cannot be (easily) eliminated, but they aren't different from how the current module adapter implementation is related to `struct comp_dev` - currently `struct processing_module` also contains a pointer to `struct comp_dev` and it uses it for similar purposes.
So, I don't see why the data blob functions cannot be added to the "native system services" set. That set anyway will include not only the module adapter API native methods, but also mathematics (`gcd()`), memory allocation, logging, etc. It might be true that we can modify the smart amp test module to not use the data blob API (and not include the header), then we don't need to add these functions now and we can think about that when we migrate the first module, that does need to use them.

pjdobrowolski: How many components/modules are using data_blob.c ?

lyakh: > How many components/modules are using data_blob.c ?

```
sof$ git grep "data_blob\.h" src/audio/
src/audio/crossover/crossover.c:#include <sof/audio/data_blob.h>
src/audio/data_blob.c:#include <sof/audio/data_blob.h>
src/audio/dcblock/dcblock.c:#include <sof/audio/data_blob.h>
src/audio/drc/drc.c:#include <sof/audio/data_blob.h>
src/audio/eq_fir/eq_fir.c:#include <sof/audio/data_blob.h>
src/audio/eq_iir/eq_iir.c:#include <sof/audio/data_blob.h>
src/audio/google/google_hotword_detect.c:#include <sof/audio/data_blob.h>
src/audio/google/google_rtc_audio_processing.c:#include <sof/audio/data_blob.h>
src/audio/igo_nr/igo_nr.c:#include <sof/audio/data_blob.h>
src/audio/mfcc/mfcc.c:#include <sof/audio/data_blob.h>
src/audio/mux/mux.c:#include <sof/audio/data_blob.h>
src/audio/rtnr/rtnr.c:#include <sof/audio/data_blob.h>
src/audio/tdfb/tdfb.c:#include <sof/audio/data_blob.h>
```

pjdobrowolski: We must discuss it in bigger group because if I understand correctly blob isn't generic thing and can be different per module. That is why in reference fw, blobs are internal module thing and each module is doing with blobs whatever they want. I'm not against just I need to be sure what extending native_system_services will cause in near future.

pjdobrowolski: I'm no expert at blob handling but as I understand correctly, Can FW process blob and pass it as a pointer by native system service? For me it seams unnecessary to pass all data_blob.c.

lyakh: @pjdobrowolski that's the current data-blob API, it probably can be improved and made slimmer, but this is what we currently have

RanderWang:
no need

PR #8318, opened 2023-10-12
ipc4: Add cross-core binding support

serhiy-katsyuba-intel:
Implements binding of two pipelines from different cores so stream could travel cross-core.

### DIFF
--- a/src/include/sof/schedule/ll_schedule_domain.h
+++ b/src/include/sof/schedule/ll_schedule_domain.h
@@ -44,6 +44,17 @@ struct ll_schedule_domain_ops {
 				 struct task *task, uint32_t num_tasks);
 	void (*domain_enable)(struct ll_schedule_domain *domain, int core);
 	void (*domain_disable)(struct ll_schedule_domain *domain, int core);
+#if CONFIG_CROSS_CORE_STREAM
+	/*
+	 * Unlike domain_disable(), these are intended to temporary block LL from
+	 * starting its next cycle. Triggering (e.g., by means of timer interrupt)
+	 * is still enabled and registered but execution of next cycle is blocked.
+	 * Once unblocked, if triggering were previously registered in a blocked
+	 * state -- next cycle execution could start immediately.
+	 */
+	void (*domain_block)(struct ll_schedule_domain *domain);
+	void (*domain_unblock)(struct ll_schedule_domain *domain);
+#endif
 	void (*domain_set)(struct ll_schedule_domain *domain, uint64_t start);
 	void (*domain_clear)(struct ll_schedule_domain *domain);
 	bool (*domain_is_pending)(struct ll_schedule_domain *domain,
@@ -191,6 +202,20 @@ static inline void domain_disable(struct ll_schedule_domain *domain, int core)
 		atomic_sub(&domain->enabled_cores, 1);
 	}
 }
+
+#if CONFIG_CROSS_CORE_STREAM
+static inline void domain_block(struct ll_schedule_domain *domain)
+{
+	if (domain->ops->domain_block)
+		domain->ops->domain_block(domain);
+}
+
+static inline void domain_unblock(struct ll_schedule_domain *domain)
+{
+	if (domain->ops->domain_unblock)
+		domain->ops->domain_unblock(domain);
+}
+#endif
 
 static inline bool domain_is_pending(struct ll_schedule_domain *domain,
 				     struct task *task, struct comp_dev **comp)
--- a/src/ipc/ipc4/helper.c
+++ b/src/ipc/ipc4/helper.c
@@ -337,9 +337,77 @@ static struct comp_buffer *ipc4_create_buffer(struct comp_dev *src, bool is_shar
 	ipc_buf.size = buf_size;
 	ipc_buf.comp.id = IPC4_COMP_ID(src_queue, dst_queue);
 	ipc_buf.comp.pipeline_id = src->ipc_config.pipeline_id;
-	ipc_buf.comp.core = src->ipc_config.core;
+	ipc_buf.comp.core = cpu_get_id();
 	return buffer_new(&ipc_buf, is_shared);
 }
+
+#if CONFIG_CROSS_CORE_STREAM
+/*
+ * Disabling interrupts to block next LL cycle works much faster comparing using
+ * of condition variable and mutex. Since same core binding is the most typical
+ * case, let's use slower cond_var blocking mechanism only for not so typical
+ * cross-core binding.
+ *
+ * Note, disabling interrupts to block LL for cross-core binding case will not work
+ * as .bind() handlers are called on corresponding cores using IDC tasks. IDC requires
+ * interrupts to be enabled. Only disabling timer interrupt instead of all interrupts
+ * might work. However, as CPU could go to some power down mode while waiting for
+ * blocking IDC call response, it's not clear how safe is to assume CPU can wakeup
+ * without timer interrupt. It depends on blocking IDC waiting implementation. That
+ * is why additional cond_var mechanism to block LL was introduced which does not
+ * disable any interrupts.
+ */
+
+#define ll_block(cross_core_bind) \
+	do { \
+		if (cross_core_bind) \
+			domain_block(sof_get()->platform_timer_domain); \
+		else \
+			irq_local_disable(flags); \
+	} while (0)
+
+#define ll_unblock(cross_core_bind) \
+	do { \
+		if (cross_core_bind) \
+			domain_unblock(sof_get()->platform_timer_domain); \
+		else \
+			irq_local_enable(flags); \
+	} while (0)
+
+/* Calling both ll_block() and ll_wait_finished_on_core() makes sure LL will not start its
+ * next cycle and its current cycle on specified core has finished.
+ */
+static int ll_wait_finished_on_core(struct comp_dev *dev)
+{
+	/* To make sure (blocked) LL has finished its current cycle, it is
+	 * enough to send any blocking IDC to the core. Since IDC task has lower
+	 * priority then LL thread and cannot preempt it, execution of IDC task
+	 * happens when LL thread is not active waiting for its next cycle.
+	 */
+
+	int ret;
+	struct ipc4_base_module_cfg dummy;
+
+	if (cpu_is_me(dev->ipc_config.core))
+		return 0;
+
+	/* Any blocking IDC that does not change component state could be utilized */
+	ret = comp_ipc4_get_attribute_remote(dev, COMP_ATTR_BASE_CONFIG, &dummy);
+	if (ret < 0) {
+		tr_err(&ipc_tr, "comp_ipc4_get_attribute_remote() failed for module %#x",
+		       dev_comp_id(dev));
+		return ret;
+	}
+
+	return 0;
+}
+
+#else
+
+#define ll_block(cross_core_bind)	irq_local_disable(flags)
+#define ll_unblock(cross_core_bind)	irq_local_enable(flags)
+
+#endif
 
 int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 {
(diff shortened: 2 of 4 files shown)


tmleman (commented):
What about LNL? Maybe enable it by default for all ACE platforms.

src/ipc/ipc4/helper.c
```diff
@@ -337,10 +337,78 @@ static struct comp_buffer *ipc4_create_buffer(struct comp_dev *src, bool is_shar
...
+ * as .bind() handlers are called on corresponding cores using IDC tasks. IDC requires
+ * interrupts to be enabled. Only disabling timer interrupt instead of all interrupts
+ * might work. However, as CPU could go to some power down mode while waiting for
+ * blocking IDC call response, it's not clear how safe is to assume CPU can wakeup
+ * without timer interrupt. It depends on blocking IDC waiting implementation. That
+ * is why additional cond_var mechanism to block LL was introduced which does not
+ * disable any interrupts.
+ */
+
+#define block_ll(cross_core_bind) \
```
lgirdwood, line 361: It might be more readable if we used a ```ll_``` prefix for all these internal macros and APIs (instead of ll in the middle or end).

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -89,3 +89,5 @@ CONFIG_DEBUG_COREDUMP_MEMORY_DUMP_MIN=y
 
 CONFIG_PROBE=y
 CONFIG_PROBE_DMA_MAX=2
+
+CONFIG_CROSS_CORE_STREAM=y
```
lgirdwood, line 93: Lets also add this for all IPC4 targets including CAVS 2.5 (i.e. get more test coverage in CI).

src/ipc/ipc4/helper.c
serhiy-katsyuba-intel: Done.

src/ipc/ipc4/helper.c
```diff
@@ -337,10 +337,78 @@ static struct comp_buffer *ipc4_create_buffer(struct comp_dev *src, bool is_shar
...
+		if (cross_core_bind) \
+			domain_unblock(sof_get()->platform_timer_domain); \
+		else \
+			irq_local_enable(flags); \
+	} while (0)
+
+/* Calling both block_ll() and wait_ll_finished_on_core() makes sure LL will not start its
+ * next cycle and its current cycle on specified core has finished.
+ */
+static int wait_ll_finished_on_core(struct comp_dev *dev)
```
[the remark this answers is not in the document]
serhiy-katsyuba-intel: I simply re-use an existing mechanism here: blocking IDC call. Current IDC implementation uses k_p4wq_wait(work, K_FOREVER). Replacing it to some timeout will affect all IDC calls -- not sure if it is OK. However, it seems might be OK to have some big timeout for blocking IDC calls and log an error if no response received.

src/include/sof/schedule/ll_schedule_domain.h
```diff
@@ -192,6 +203,20 @@ static inline void domain_disable(struct ll_schedule_domain *domain, int core)
 	}
 }
 
+#if CONFIG_CROSS_CORE_STREAM
+static inline void domain_block(struct ll_schedule_domain *domain)
+{
+	if (domain->ops->domain_block)
```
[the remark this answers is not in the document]
lgirdwood: ping - I think this will be difficult to quickly root cause or eliminate this as a cause for glitch without this debug logging i.e. only logged when debug log is enabled in Kconfig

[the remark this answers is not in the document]
serhiy-katsyuba-intel: Yes, the log here would be helpful. However, this is a header file, our logging mechanism requires to pass some context parameter, not yet sure how such context var is declared when calling log from header file.

tmleman (approved):
LGTM

src/schedule/zephyr_domain.c
```diff
@@ -67,6 +72,16 @@ static void zephyr_domain_thread_fn(void *p1, void *p2, void *p3)
 		/* immediately go to sleep, waiting to be woken up by the timer */
 		k_sem_take(&dt->sem, K_FOREVER);
 
+#if CONFIG_CROSS_CORE_STREAM
+		if (atomic_get(&zephyr_domain->block)) {
```
kv2019i, line 76: @serhiy-katsyuba-intel what's the thinking behind using an atomic_get()? I think the normal pattern with condvars is to take the mutex always, check the condition (with lock held), if condition is not met, call k_condvar_wait(), else if condition is met, release the mutex and exit.

kv2019i (commented):
Thanks @serhiy-katsyuba-intel , this looks pretty nice. One clarifying question inline, can you check?

src/ipc/ipc4/helper.c
```diff
@@ -337,10 +337,78 @@ static struct comp_buffer *ipc4_create_buffer(struct comp_dev *src, bool is_shar
 	ipc_buf.size = buf_size;
 	ipc_buf.comp.id = IPC4_COMP_ID(src_queue, dst_queue);
 	ipc_buf.comp.pipeline_id = src->ipc_config.pipeline_id;
-	ipc_buf.comp.core = src->ipc_config.core;
+	ipc_buf.comp.core = cpu_get_id();
```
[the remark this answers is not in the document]
serhiy-katsyuba-intel: Yes, this does not change behaviour for same core binding. For same core binding we call process_on_core() if necessary and do everything on core on which components were created. So src->ipc_config.core == sink->ipc_config.core == cpu_get_id(). For cross-core bind src->ipc_config.core != sink->ipc_config.core, we stay on core 0, buffer is created on core 0 and cpu_get_id() == 0.

src/schedule/zephyr_domain.c
kv2019i: Ack @serhiy-katsyuba-intel . Would be good to sanity check the cost -- the uncontented case should be pretty fast for mutex_lock, just taking a spinlock and checking lock owner. The slower path is only taken if lock is held. Cost on xtensa needs to be checked though -- it's true this is a high-call rate place.

If we keep this optimization, a comment would be good. I think there's a race now:
```
core1:   if (atomic_get(&block))   // returns 0, so branch not taken
core2:     mutex_lock(&block_mutex)  
core2:     atomic_set(&block, 1)
core1:   dt->handler(dt->arg);
core2:     // continue execution after ll_block()
```

I think your code is safe as you'll send the extra IDC to complete the current LL task, but this would be good to note as a comment (that this is per design).

lgirdwood: > I guess taking/releasing mutex is quite expensive.

@serhiy-katsyuba-intel Best to also check with @marcinszkudlinski as he has the data showing what is cheap/expensive re locking on xtensa.

serhiy-katsyuba-intel: Added comment with description before that "if".

src/include/sof/schedule/ll_schedule_domain.h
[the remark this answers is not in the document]
serhiy-katsyuba-intel: Added logging to zephyr_domain_block() and zephyr_domain_unblock().

src/schedule/zephyr_domain.c
```diff
@@ -221,9 +246,38 @@ static int zephyr_domain_unregister(struct ll_schedule_domain *domain,
...
 
+#if CONFIG_CROSS_CORE_STREAM
+static void zephyr_domain_block(struct ll_schedule_domain *domain)
+{
+	struct zephyr_domain *zephyr_domain = ll_sch_domain_get_pdata(domain);
+
+	tr_dbg(&ll_tr, "Blocking LL scheduler");
+
+	k_mutex_lock(&zephyr_domain->block_mutex, K_FOREVER);
+	atomic_set(&zephyr_domain->block, 1);
```
lgirdwood, line 257: Btw, does this have to be atomic since all users should need to gain the mutex before use ? i.e. we could save some cycles with a regular int ?

serhiy-katsyuba-intel: It has to be volatile to make sure compiler does not optimize reading from memory when checking block value in second "if".

```
if (atomic_get(&zephyr_domain->block)) {
	k_mutex_lock(&zephyr_domain->block_mutex, K_FOREVER);
	if (atomic_get(&zephyr_domain->block))
		k_condvar_wait(&zephyr_domain->block_condvar,
			       &zephyr_domain->block_mutex, K_FOREVER);
	k_mutex_unlock(&zephyr_domain->block_mutex);
}
```
Without volatile compiler may optimize second read of zephyr_domain->block but we want to actually check block value after locking the mutex.

atomic here is used not for the purpose of atomic read/write, but for a purpose of volatile read/write. I could replace atomic with volatile bool, however, we seems use atomic in similar cases in SOF.

lgirdwood: ok, but should't the code block be (if I'm understanding the locking around ```zephyr_domain->block```)
```c
/* get the global domain mutex */
k_mutex_lock(&zephyr_domain->block_mutex, K_FOREVER);

/* now critical section with lock held - no one should read/write domain->block without holding mutex */
if (zephyr_domain->block) /* can now be an int since lock is held and no one should change without mtex held */
	k_condvar_wait(&zephyr_domain->block_condvar,
		       &zephyr_domain->block_mutex, K_FOREVER);

/* leave the critical section */
k_mutex_unlock(&zephyr_domain->block_mutex);
```

kv2019i (approved):
Thanks @serhiy-katsyuba-intel , looks good now!

src/ipc/ipc4/helper.c
```diff
@@ -418,12 +487,26 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
+
+	if (cross_core_bind) {
+#if CONFIG_CROSS_CORE_STREAM
+		/* Make sure LL has finished on both cores */
+		if (!cpu_is_me(source->ipc_config.core))
+			if (ll_wait_finished_on_core(source) < 0)
+				goto free;
+		if (!cpu_is_me(sink->ipc_config.core))
+			if (ll_wait_finished_on_core(sink) < 0)
+				goto free;
```
lyakh, line 504: in principle with the current LL we can only try to connect complete pipelines, running on different cores. In such a case the scheduler has to be stopped to prevent a running pipeline from detecting a new connection and trying to use it before it's ready. For this to happen one of the pipelines, that we're connecting has to be in an active running state. So, it would be enough to only stop the LL scheduler on the core, corresponding to that running pipeline, not on both cores.

src/ipc/ipc4/helper.c
```diff
@@ -527,17 +614,39 @@ int ipc_comp_disconnect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
+		if (!cpu_is_me(src->ipc_config.core))
+			if (ll_wait_finished_on_core(src) < 0) {
+				ll_unblock(cross_core_unbind);
+				return IPC4_FAILURE;
+			}
+		if (!cpu_is_me(sink->ipc_config.core))
+			if (ll_wait_finished_on_core(sink) < 0) {
+				ll_unblock(cross_core_unbind);
+				return IPC4_FAILURE;
+			}
```
lyakh, line 635: same here - when disconnecting two pipelines, one of them would already be inactive, so only need to block the active one - if any

src/schedule/zephyr_domain.c
lyakh: I think the comment should rather describe *why* checking that atomic variable without locking is enough. The reason is - as @kv2019i has pointed out - the IDC. So what we're doing is something like
```
IDC <core X>				LL <core Y>
set flag				scheduler running
wait-for-scheduler-idle-on-Y
					scheduler completes
			<SYNC POINT>
process IDC				scheduler starts but detects the flag, waits on condvar
signal condvar
			<SYNC POINT>
					scheduler proceeds running the pipeline
```

marcinszkudlinski:
@lyakh, your review, however right, is an optimalization. All we can gain/loose  is a couple of microseconds delay of LL pipeline start. 
But potentially there's an open door for some races, leading to sporadic defect, extremely hard to track

lyakh:
> @lyakh, your review, however right, is an optimalization. All we can gain/loose is a couple of microseconds delay of LL pipeline start.

Yes-yes, absolutely, nothing critical, can be a follow-up. I think there's one more benefit from that change though: maintainability / self-documentation / clarity. I prefer the code that clearly knows what it is doing over "we don't quite know what might happen here so let's just put another layer of safety rails just in case"

> But potentially there's an open door for some races, leading to sporadic defect, extremely hard to track

See above :-) I don't know what "sporadic defects" you're talking about, sorry. If you have any specific race scenarios, they should be documented and handled appropriately.

src/schedule/zephyr_domain.c
kv2019i: That's a good point @lyakh @serhiy-katsyuba-intel . There's two critical pieces of info anyone reading the code needs to understand: 1) the classic condvar process is modified here on purpose (this is documented), 2) this modification is safe because of logic elsewhere (the blocking IDC).

Let's proceed with this now as we got both CI and checks. I'm happy with the (1) part as that's enough for the reviewer/maintainer to alert this is done on purpose.

PR #8319, opened 2023-10-12
Replace use of audio_format objects

ranj063:


ranj063:
SOFCI TEST

tools/topology/topology2/platform/intel/sdw-jack-generic.conf
```diff
@@ -186,56 +186,103 @@ Object.Pipeline.host-gateway-capture [
...
+			CombineArrays.Object.Base.output_audio_format [
+				# array of 2ch formats with range of out valid bit depths
+				{
+					out_bit_depth [
+						32
+					]
+					out_valid_bit_depth [
+						24
+						32
+					]
```
plbossart, line 224: is 

````
out_channels [					
    2
]
````

missing or default?

File: tools/topology/topology2/platform/intel/sdw-jack-generic.conf [hunk 2]
```diff
@@ -186,56 +186,103 @@ Object.Pipeline.host-gateway-capture [
...
+					out_ch_cfg [
+						$CHANNEL_CONFIG_MONO
+					]
+					out_ch_map [
+						$CHANNEL_MAP_MONO
+					]
+				}
+				# 16-bit 2ch
+				{
+					out_bit_depth		16
```
plbossart, line 266: is 
````
out_bit_depth		16
````
the same as 
````
out_bit_depth	[
	16
]
````
if yes, why not always use the former case for single-value configurations?

File: tools/topology/topology2/platform/intel/sdw-jack-generic.conf [hunk 3]
```diff
@@ -186,56 +186,103 @@ Object.Pipeline.host-gateway-capture [
 		Object.Widget.host-copier.1 {
 			stream_name	"Passthrough Capture 0"
 			pcm_id 1
-			num_input_audio_formats 9
+			num_input_audio_formats 3
```
plbossart, line 189: typo in commit message: introduces

tools/topology/topology2/platform/intel/sdw-jack-generic.conf
ranj063: default

tools/topology/topology2/platform/intel/sdw-jack-generic.conf
ranj063: this is a feature or a limitation with the way the CombineArrays feature has been defined in the compilet today. So all values even single values have to be specified with an array. I've updated the patch to use array consistenly now

plbossart (approved):

fredoh9:
SOFCI TEST

ranj063:
> looks good but I wonder if we should rootcause #8320 first before merging.

@kv2019i fix posted in #8323. We can merge this after that one

PR #8322, opened 2023-10-13
Tools: Topology2: Change gain default curve type to linear

singalsu:


andrula-song (approved):

btian1 (approved):
QB passed, means at least, QB does not force volume ramp with other types, for later, if there is no mandatory requirement, we can keep at most 1-2 ramp type.

singalsu:
> QB passed, means at least, QB does not force volume ramp with other types, for later, if there is no mandatory requirement, we can keep at most 1-2 ramp type.

QB is not using topology at all so this PR cannot impact it. QB is using direct IPC commands via Python scripts and slim driver kernel interface.

singalsu:
> @singalsu can you clarify how this is tested? patch2 uses windows_fade for tests but what tests are you referring to?

I meant that some CI Jenkins tests use nocodec topologies. The tests in sof-ci/jenkins/pr-device-test/main-ace contain MTLP_RVP_NOCODEC. Also sof-ci/jenkins/pr-device-test/main-cavs2.5 contain TGLU_RVP_NOCODEC_IPC4ZPH and ADLP_RVP_NOCODEC_IPC4ZPH. These tests will still use Windows Fade ramp except those pipelines mentioned in patch.

None of Jenkins tests are checking presence of any ramp. But if we check MCPS peak with nocodec the peaks will remain higher. The lower MCPS peaks will be seen with normal topologies those would use the old linear ramp from tplg1.

plbossart:
@singalsu I was referring to the sof-test suite, I really don't know anything about Jenkins and how PR tests map to sof-test helpers.

singalsu:
> @singalsu I was referring to the sof-test suite, I really don't know anything about Jenkins and how PR tests map to sof-test helpers.

Sof-test tests whatever topology has been loaded to DUT, I don't think there's a case that would force load of other topology than kernel default. Test DUTs seem to have a good distribution with nocodec, hda, and sdw topologies.

Also none of the sof-test cases measure the ramp shapes. But I learned a while ago that alsabat fails if ramp is made longer than 20 ms, because it can't skip the beginning for detecting sine tones. So, I had to revert previously capture ramp stretch to 40 ms for nocodec. The failing alsabat record test was done only with nocodec case. So we don't have same tests for every DUT type in the test plans.

lgirdwood:
There is currently no way to automatically test this feature via the topology, we have to rely on QB to test the code/algo only via IPC.

PR #8324, opened 2023-10-13
Audio: RTNR: Convert to module API

singalsu:


src/audio/rtnr/rtnr.c
```diff
@@ -214,76 +220,73 @@ static int32_t rtnr_check_config_validity(struct comp_dev *dev,
...
 	cd->model_handler = comp_data_blob_handler_new(dev);
 	if (!cd->model_handler) {
-		comp_cl_err(&comp_rtnr, "rtnr_new(): comp_data_blob_handler_new() failed.");
+		comp_err(dev, "rtnr_new(): comp_data_blob_handler_new() failed.");
+		ret = -ENOMEM;
 		goto cd_fail;
 	}
 
 	/* Get initial configuration from topology */
-	ret = rtnr_set_config_bytes(dev, ipc_rtnr->data, ipc_rtnr->size);
```
singalsu, line 265: @MingJenTai Can you please help with this. Has this worked, I wonder a current kernel provides the configuration blob in comp new() IPC?

src/audio/rtnr/rtnr.c
```diff
@@ -549,9 +573,13 @@ static int rtnr_set_config_bytes(struct comp_dev *dev,
 	return ret;
 }
 
-static int rtnr_set_bin_data(struct comp_dev *dev, struct sof_ipc_ctrl_data *cdata)
```
singalsu, line 552: @MingJenTai Also need helping hand here. Does this component have two binary controls with SOF_RTNR_CONFIG and SOF_RTNR_DATA. There is no `struct sof_ipc_ctrl_data *cdata` in IPC4 so can't separate them by type. Also the kernel controls cache keeps only the last control instance data. Could this be handled in IPC4 with two bytes controls instances?

src/audio/rtnr/rtnr.c
```diff
@@ -741,22 +752,25 @@ static void rtnr_copy_to_sof_stream(struct audio_stream *dst,
 }
 
 /* copy and process stream data from source to sink buffers */
-static int rtnr_copy(struct comp_dev *dev)
```
singalsu, line 744: @MingJenTai The code looks that there could be multiple source buffers and one sink. But it also looks from consume than there would be now one supported. Can you confirm the number of sources to support, one or many?  Thanks!

singalsu:
This is early draft version, shared to find about opens related to module conversion. Currently if fails in init() due to not having configuration blob available, that would be sent after init(), before prepare() by kernel.

lgirdwood (commented):
Looks clean, I assume you can test using 3P pass through mode ?

tools/rimage/config/tgl-cavs.toml
```diff
@@ -496,3 +496,19 @@ count = 21
 	pin = [0, 0, 0xfeef, 0xf, 0xf, 0x45ff, 1, 0, 0xfeef, 0xf, 0xf, 0x1ff]
 	# mod_cfg [PAR_0 PAR_1 PAR_2 PAR_3 IS_BYTES CPS IBS OBS MOD_FLAGS CPC OBLS]
 	mod_cfg = [0, 0, 0, 0, 4096, 1000000, 128, 128, 0, 0, 0]
+
```
btian1, line 499: any justification for enable RTNR in TGL? how about config side? 
My opinion is: TGL is already PVed, if we want to verify this, better go to mtl?

src/arch/host/configs/library_defconfig
```diff
@@ -8,9 +8,11 @@ CONFIG_COMP_MFCC=y
 CONFIG_COMP_MODULE_ADAPTER=y
 CONFIG_COMP_MULTIBAND_DRC=y
 CONFIG_COMP_MUX=y
+#CONFIG_COMP_RTNR=y
```
btian1, line 11: suggest directly delete line.

tools/testbench/common_test.c
```diff
@@ -56,6 +56,7 @@ int tb_setup(struct sof *sof, struct testbench_prm *tp)
 	sys_comp_module_eq_iir_interface_init();
 	sys_comp_module_multiband_drc_interface_init();
 	sys_comp_module_mux_interface_init();
+	//sys_comp_module_rtnr_interface_init();
```
btian1, line 59: delete if no use.

src/arch/host/configs/library_defconfig
singalsu: Draft - I left this a reminder to fix an issue in testbench build. There's something that fails to compile. I'd want to be able load and run every component (also with mock) to testbench to do basic checks in CI (scripts/host-testbench.sh) to avoid regressions with more rare use.

If I can't solve the build issue I will drop all testbench related.

tools/rimage/config/tgl-cavs.toml
singalsu: There's quite an effort for occasional need to use to create these, so I'd like to keep it as available. My plan is to keep kconfig default n, because this is not a real usable component as mock-up version. But with manual kconfig enable this can be loaded with suitable topology to test in real DUT.

And TGL is my choice because it's the best I have available for my work. Once this is ready I can duplicate to MTL and after.

tools/testbench/common_test.c
singalsu: Same, I have a problem with testbench build so I just commented out. Something in this code breaks it.

tools/rimage/config/tgl-cavs.toml
lgirdwood: We need to add for SW dev reasons. i.e. easier to test/fix any failure if you have HW on your desk and manifest already has the module.

singalsu:
> Looks clean, I assume you can test using 3P pass through mode ?

I need to modify further the code to run it, now the problem with binary control arrival after init() prevents the init() to pass. I think it's too early to attempt to initialize the algorithm there. Also, it does not impact mockup, but there's an issue in controlling two internal features with same binary control. It would be simplest to have two binary controls unless I've understood this wrong.

src/audio/rtnr/rtnr.c
singalsu: Yes, there are two binary controls as seen from https://github.com/thesofproject/sof/blob/main/tools/topology/topology1/sof/pipe-rtnr-google-rtc-audio-processing-capture.m4. The current draft PR version now handles the two control. I've converted the supplied blobs for tplg1 to tplg2 now.

src/audio/rtnr/rtnr.c
singalsu: The topology https://github.com/thesofproject/sof/blob/main/tools/topology/topology1/sof/pipe-rtnr-google-rtc-audio-processing-capture.m4 used one source and sink so I will test this work with similar number of them.

src/audio/rtnr/rtnr.c
```diff
@@ -766,96 +735,95 @@ static int rtnr_copy(struct comp_dev *dev)
...
-			sink_bytes = frames * audio_stream_frame_bytes(&sink->stream);
-
-			buffer_stream_invalidate(source, source_bytes);
-
-			/* Run processing function */
+		/*
+		 * Processing function uses an array of pointers to source streams
+		 * as parameter.
+		 */
+		cd->rtnr_func(mod, (struct audio_stream_rtnr **)&sources_stream,
```
lyakh, line 754: I know this is copied from the current version, but I think, that if you use `sources_stream` instead of `&sources_stream` you get the same result, but a better readability and you won't need the type-cast. In all locations of course

src/audio/rtnr/rtnr.c
```diff
@@ -865,12 +833,20 @@ static int rtnr_prepare(struct comp_dev *dev)
 	sinkb = list_first_item(&dev->bsink_list, struct comp_buffer, source_list);
 	cd->sink_format = audio_stream_get_frm_fmt(&sinkb->stream);
 	cd->sink_stream.frame_fmt = audio_stream_get_frm_fmt(&sinkb->stream);
+	sourceb = list_first_item(&dev->bsource_list, struct comp_buffer, sink_list);
+	igo_nr_set_alignment(&sourceb->stream, &sinkb->stream);
+
+	ret = rtnr_check_params(mod, &sourceb->stream, &sinkb->stream);
+	if (ret)
+		goto err;
+
```
lyakh, line 840: superfluous empty line

src/audio/rtnr/rtnr.c
```diff
@@ -684,8 +684,7 @@ static int rtnr_set_config(struct processing_module *mod, uint32_t param_id,
 #endif
 }
 
-static void rtnr_copy_from_sof_stream(struct audio_stream_rtnr *dst,
-				      struct audio_stream *src)
+void rtnr_copy_from_sof_stream(struct audio_stream_rtnr *dst, struct audio_stream *src)
```
btian1, line 687: these copy can be avoided? if can't, can we use some simple way? like:
memcpy(&dst->size, &src->size, 4*7);

lgirdwood:
@singalsu still draft ?

src/audio/rtnr/rtnr.c
```diff
@@ -768,97 +735,94 @@ static int rtnr_copy(struct comp_dev *dev)
...
-		comp_dbg(dev, "rtnr_copy() source_id: %d, frames = %d", buf_get_id(source),
-			 frames);
+		/* Run processing function */
 
-		if (frames) {
-			source_bytes = frames * audio_stream_frame_bytes(&source->stream);
-			sink_bytes = frames * audio_stream_frame_bytes(&sink->stream);
+		/* copy required data from sof audio stream to RTNR audio stream */
+		rtnr_copy_from_sof_stream(sources_stream[0], source);
+		rtnr_copy_from_sof_stream(sink_stream, sink);
```
lyakh, line 748: is this correct? Need to copy data from a sink stream before processing?

src/audio/rtnr/rtnr.c
```diff
@@ -768,97 +735,94 @@ static int rtnr_copy(struct comp_dev *dev)
...
+		RTKMA_API_Process(cd->rtk_agl, 0, cd->source_rate, MicNum);
 
-			/*
-			 * Processing function uses an array of pointers to source streams
-			 * as parameter.
-			 */
-			cd->rtnr_func(dev, (struct audio_stream_rtnr **)&sources_stream,
-						sink_stream, frames);
+		/* copy required data from RTNR audio stream to sof audio stream */
+		rtnr_copy_to_sof_stream(source, sources_stream[0]);
```
lyakh, line 763: and source stream after processing?

src/audio/rtnr/rtnr_stub.c
```diff
@@ -6,25 +6,56 @@
...
+{
+	struct audio_stream sof_source;
+	struct audio_stream sof_sink;
+
+	rtnr_copy_to_sof_stream(&sof_source, sources[0]);
+	rtnr_copy_to_sof_stream(&sof_sink, sink);
+	audio_stream_copy(&sof_source, 0, &sof_sink, 0,
+			  frames * audio_stream_get_channels(&sof_sink));
+	rtnr_copy_from_sof_stream(sources[0], &sof_source);
+	rtnr_copy_from_sof_stream(sink, &sof_sink);
```
lyakh, line 25: since these repeat throughout the code, they're probably correct and it's just my lack of understanding of how this works

src/audio/rtnr/rtnr.c
lgirdwood: ack - we need to keep code as-is and only touch the parts that interface with the module API to minimize any impact to the validated code.

src/audio/rtnr/rtnr.c
singalsu: The copy function (I've not changed functions names) is converting between SOF stream format and Realtek's library stream format. There's similarity but they are not exactly same. This only converts struct members. It's not moving and PCM data.

src/audio/rtnr/rtnr_stub.c
singalsu: Yes, I'm afraid so :)

singalsu: I've left this PR draft because I found this conversion has started to fail in both reat DUT and testbench. The audio_stream_get_channels(&sof_sink) gives 384 channels in tesbench... Need to fix something in the conversion.

src/audio/rtnr/rtnr.c
lyakh: @singalsu got it, thanks!

singalsu:
> @singalsu still draft ?

There's some regression, the stub library tries to process 384 channels per frame. Function audio_stream_get_channels() returns it. The function is correct but the stream format gets messed up somewhere.

src/audio/rtnr/rtnr_stub.c
```diff
@@ -36,11 +37,16 @@ void RTKMA_API_Prepare(void *Context)
 
 void *RTKMA_API_Context_Create(int sample_rate)
```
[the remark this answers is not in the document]
cujomalainey: @btian1 the point is for it to be empty, its simply a stub for build testing, nothing more. Also I don't know the actual implementation, that is owned by RTK and is proprietary.

src/audio/rtnr/rtnr_stub.c
```diff
@@ -36,11 +37,16 @@ void RTKMA_API_Prepare(void *Context)
...
 {
-	return NULL;
+	/* Allocate something, to avoid return NULL and cause error
+	 * in check of success of this.
+	 */
+	return rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, 42);
```
btian1, line 74: is 42 == sizeof(audio_stream_rtnr)? if so, better use sizeof, or if number are random, you can use 4, 42 is not easy to read.

lgirdwood:
@singalsu any ETA on non draft ?

src/audio/rtnr/rtnr_stub.c
singalsu: I can make it a macro. We don't know the real size used by the closed library, and the run fails if this is NULL. Since I had to choose something, number 42 a reference to plot  of https://en.wikipedia.org/wiki/The_Hitchhiker%27s_Guide_to_the_Galaxy . :)

singalsu:
Changing to ready for review!

The component now works with the stub as ALSA control disable and enable, no more stuck to xrun. I didn't change the rtnr_copy_to/from_sof_stream() from functions where the issue was seen. Apparently there was some regression in SOF that has got fixed.

src/audio/rtnr/rtnr_stub.c
btian1: you may need move 42 definition from:
https://github.com/thesofproject/sof/pull/8301/files#diff-f09311b76861c69cd17f987de389037705c0964aaf78d0d72e0e6da39bcdb2ffR12

to a more common place and reuse it between rtnr and igo_nr module.

btian1: this can be done later, now, I am ok go for merge.

btian1 (approved):

lgirdwood (approved):

PR #8325, opened 2023-10-13
Audio format cleanup

ranj063:


tools/topology/topology2/include/pipelines/cavs/deepbuffer-playback.conf
```diff
@@ -47,28 +48,31 @@ Class.Pipeline."deepbuffer-playback" {
...
-				in_valid_bit_depth	32
-				out_bit_depth		32
-				out_valid_bit_depth	32
-			}
+			Object.Base.output_audio_format [
+				{
+					out_bit_depth		32
+					out_valid_bit_depth	32
+				}
+			]
```
plbossart, line 75: the comments on line 54 and 69 are no longer relevant, remove?

ranj063: this patch was from a different PR, will fix it in a follow up

plbossart (approved):

tools/topology/topology2/include/pipelines/cavs/gain-copier-capture.conf
```diff
@@ -45,59 +46,89 @@ Class.Pipeline."gain-copier-capture" {
...
+
+			Object.Base.output_audio_format [
+				{
+					out_bit_depth		32
+					out_valid_bit_depth	32
+				}
+				{
+					out_bit_depth		16
+					out_valid_bit_depth	16
+				}
```
bardliao, line 102: Does the order matter? The existing code format 1 is 16 bit and format 2 is 32 bit. But after the change output format 1 is 32 bit and output format 2 is 16 bit.

kv2019i (approved):

tools/topology/topology2/include/pipelines/cavs/gain-copier-capture.conf
lgirdwood: order in topology should not matter as FW should iterate over configs to match requested based on copier use case.

lgirdwood (approved):

tools/topology/topology2/include/pipelines/cavs/gain-copier-capture.conf
ranj063: yes, order doesnt matter

PR #8327, opened 2023-10-16
drivers: sdma: Add sdma script code in header file

dbaluta:
For ease of use add sdma script code directly in the sdma header file.


src/include/sof/drivers/sdma.h
```diff
@@ -164,7 +164,183 @@
...
+/*
+ * From "sdma_script_code_imx7d_4_5.h"
+ */
+
+/* SDMA RAM image start address and size */
+#define RAM_CODE_START_ADDR     6144
+#define RAM_CODE_SIZE           1315
+
+/* buffer that holds the SDMA RAM image */
+static const short sdma_code[] = {
```
LaurentiuM1234, line 177: Suggestion: maybe we should encase this in an `#ifdef` block? I'm assuming we're only going to use this if we need to change the "default" script. For MICFIL we could do something like `select SDMA_ALLOW_SCRIPT_CHANGE` in `config IMX_MICFIL`.

dbaluta: For now this can be opted out using CONFIG_HAVE_SDMA_FIRMWARE. So if you want to change the default script just select CONFIG_HAVE_SDMA_FIRMWARE.

LaurentiuM1234: My bad, missed the already existing `#if CONFIG_HAVE_SDMA_FIRMWARE` block.

LaurentiuM1234 (approved):

dbaluta:
@LaurentiuM1234 

As per your suggestion I added one more patch that selects CONFIG_HAVE_SDMA script when IMX_MICFIL is selected.

lgirdwood (approved):
I'm sure @plbossart remebers SDMA scripts on i.MX ;)

iuliana-prodan (approved):

btian1 (approved):

dbaluta:
Some Intel CI tests are failing but this doesn't touch Intel code. 

@lgirdwood  can you please help merge this?

PR #8330, opened 2023-10-17
DP: fix DP thread priority

jxstelter:
After enabling deadline calculation DP threads began to lose priority in relation to logger thread which currently is set to the same priority but has zeroed deadline.
Since processing should be not disturbed by logging operations, the ZEPHYR_DP_THREAD_PRIORITY must be set to lower value than logger.

### DIFF
--- a/src/audio/pipeline/pipeline-schedule.c
+++ b/src/audio/pipeline/pipeline-schedule.c
@@ -48,7 +48,7 @@ DECLARE_SOF_UUID("dp-task", dp_task_uuid, 0xee755917, 0x96b9, 0x4130,
 /**
  * \brief a priority of the DP threads in the system.
  */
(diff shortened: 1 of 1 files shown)


marcinszkudlinski (approved):

lgirdwood (approved):
This is probably worth an inline comment as we update more DP logic to explain why/how we choose a good DP priority.

kv2019i:
Ready to go, just need a clean CI pass. Will kick the CI once just in case the tools build was stuck yesterday.

kv2019i:
SOFCI TEST

dnikodem (approved):

kv2019i:
"sof-ci/jenkins/pr-tools-build" has an unrelated problem with logs publishing. Proceeding to merge after checking tools build manually.

PR #8334, opened 2023-10-17
src_lite: add module

fkwasowi:
Addition of SRC Lite module,
which only supports a subset of conversions
supported by the SRC module.

Purpose of SRC Lite module is memory optimization. Code of SRC Lite is drastically reduced and requires significantly less DSP RAM memory. 
When needed one of defined conversions, driver can choose SRC Lite module instead of SRC module to optimize memory utilization.

48 -> 16kHz
44.1 -> 16 kHz
32 -> 16 kHz
44.1 -> 48


singalsu (approved):
Nice work, thanks!

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -3,6 +3,7 @@ CONFIG_IPC_MAJOR_4=y
```
dbaluta, line 3: Can you please add some justification in the commit message of why this patch is needed?

src/audio/src/src.c
```diff
@@ -64,20 +67,20 @@
 LOG_MODULE_REGISTER(src, CONFIG_SOF_LOG_LEVEL);
 
 /* Calculates the needed FIR delay line length */
-static int src_fir_delay_length(struct src_stage *s)
+int src_fir_delay_length(struct src_stage *s)
```
lgirdwood, line 70: I'm guessing we need to expose these now to common SRC code ? If so can we move the small functions that we want to inline to the header and make static inline (and force inlining) especially for any function we call during ```process()```

src/audio/src/coef/src_lite_ipc4_int32_define.h
```diff
@@ -0,0 +1,22 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright(c) 2023 Intel Corporation. All rights reserved.
+ *
+ */
```
btian1, line 5: should you add author for this module?

src/audio/src/coef/src_lite_ipc4_int32_table.h
```diff
@@ -0,0 +1,43 @@
...
+#include "src_ipc4_int32_1_3_4535_5000.h"
+#include "src_ipc4_int32_3_2_4535_5000.h"
+#include "src_ipc4_int32_8_7_4535_5000.h"
+#include "src_ipc4_int32_16_21_4535_5000.h"
+#include "src_ipc4_int32_20_21_4167_5000.h"
+#include "../src.h"
+#include <stdint.h>
+
+/* SRC table */
+int32_t src_fir_one = 1073741824;
```
btian1, line 24: please comment this with Q format, /* 1 << 30, with Q 2.30 format */

tools/rimage/config/mtl.toml
```diff
@@ -630,3 +630,44 @@ count = 25
 	pin = [0, 0, 0xfeef, 0xf, 0xf, 0x45ff, 1, 0, 0xfeef, 0xf, 0xf, 0x1ff]
 	# mod_cfg [PAR_0 PAR_1 PAR_2 PAR_3 IS_BYTES CPS IBS OBS MOD_FLAGS CPC OBLS]
 	mod_cfg = [0, 0, 0, 0, 4096, 1000000, 128, 128, 0, 0, 0]
+
+	# SRC lite module config
```
btian1, line 634: you may need change count and version for this toml file.

btian1 (commented):
so original SRC module capability does not support listed 4 convertor?  if so, compared with add a new module, possible to add more src parameters to make src module support this?

src/audio/src/coef/src_lite_ipc4_int32_define.h
fkwasowi: Done

tools/rimage/config/mtl.toml
fkwasowi: Done

src/audio/src/coef/src_lite_ipc4_int32_table.h
singalsu: Could do it with `Q_CONVERT_FLOAT(1, 30)`.

app/boards/intel_adsp_ace15_mtpm.conf
lgirdwood: @fkwasowi e.g. best to say we are saving X MCPS or Y memory vs standard config SRC.

src/audio/src/coef/src_lite_ipc4_int32_define.h
```diff
@@ -0,0 +1,23 @@
...
+#define MAX_FIR_DELAY_SIZE 730
+#define MAX_OUT_DELAY_SIZE 900
+#define MAX_BLK_IN 80
+#define MAX_BLK_OUT 40
+#define NUM_IN_FS 16
+#define NUM_OUT_FS 10
+#define STAGE1_TIMES_MAX 32
+#define STAGE2_TIMES_MAX 32
+#define STAGE_BUF_SIZE 672
+#define NUM_ALL_COEFFICIENTS 69224
```
lgirdwood, line 21: I would prefix these with SRC_ as some names are quite common, but I think these are added before this commit so we should revisit this @singalsu ?

singalsu: Yes you are right, I can do it after this PR. These files are automatically generated so better to do the change to Matlab code also.

lgirdwood: ok, thaks @singalsu let add the prefix after this PR is merged.

kv2019i (commented):
I'm good with the change, looks very nicely done! So I think only think missing is some note on why one would choose this instead of full SRC. Doesn't have to be very detailed, but some note to explain why to choose one over the other.

lgirdwood:
@fkwasowi ping, just some minor things and good to merge.

app/boards/intel_adsp_ace15_mtpm.conf
fkwasowi: Describe added.

src/audio/src/src.c
fkwasowi: I moved small functions to src.h as static inline.

src/audio/src/coef/src_ipc4_int32_10_21_2500_5000.h
```diff
@@ -5,7 +5,7 @@
  */
 
 /** \cond GENERATED_BY_TOOLS_TUNE_SRC */
-#include "../src.h"
+
```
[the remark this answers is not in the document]
kv2019i: yup, now that this is enabled for MTL, this will break bisect

kv2019i (commented):
Thanks, my earlier comments addresses. One issue with "src_fir_one" initialization, and mergnig the build fix to one commit, and otherwise looks good to go.

src/audio/src/coef/src_lite_ipc4_int32_table.h
fkwasowi: Done

kv2019i (approved):
Thanks @fkwasowi !

lgirdwood (approved):

btian1 (commented):
@fkwasowi , as commit message mentioned SRC_LITE is saving memory when specific converter happened, could you choose one type and list the memory saving with data:
                   previous code size            with patch code size            previous ram usage           with patch ram usage, we can use 48 --> 16 as a example.

marcinszkudlinski (approved):

PR #8337, opened 2023-10-17
Tools: Topology2: Add more sof-hda-benchmark-<comp><format>-tgl.tplg builds

singalsu:
This PR adds benchmark topologies build for multiband-DRC, gain, and aria with s16/s24/s32 format source and sink.

### DIFF
--- a/tools/topology/topology2/CMakeLists.txt
+++ b/tools/topology/topology2/CMakeLists.txt
@@ -1,4 +1,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
+
+# 3.17 required: foreach(<loop_var>... IN ZIP_LISTS <lists>)
+cmake_minimum_required(VERSION 3.17)
+
 add_custom_target(topologies2)
 
 # Check alsatplg version and build topology2 if alsatplg version is
--- a/tools/topology/topology2/development/tplg-targets-bench.cmake
+++ b/tools/topology/topology2/development/tplg-targets-bench.cmake
@@ -1,0 +1,36 @@
+# SPDX-License-Identifier: BSD-3-Clause
+
+#
+# Append test topologies for multiple platforms and multiple formats
+# to TPLGS array.
+#
+
+set(platforms "TGL" "MTL")
+set(sampleformats "16" "24" "32")
+
+set(components
+	"dcblock"
+	"drc"
+	"eqiir"
+	"eqfir"
+)
+
+set(component_parameters
+	"BENCH_DCBLOCK_PARAMS=default"
+	"BENCH_DRC_PARAMS=enabled"
+	"BENCH_EQIIR_PARAMS=highpass_50hz_0db_48khz"
+	"BENCH_EQFIR_PARAMS=loudness"
+)
+
+foreach(p ${platforms})
+	string(TOLOWER ${p} pl)
+
+	# Add components with all sample formats
+	foreach(sf ${sampleformats})
+		foreach(comp bench_param IN ZIP_LISTS components component_parameters)
+			    set(item "sof-hda-generic\;sof-hda-benchmark-${comp}${sf}-${pl}\;PLATFORM=${p},HDA_CONFIG=benchmark,BENCH_CONFIG=${comp}${sf},${bench_param}")
+			    #message(STATUS "Item=" ${item})
+			    list(APPEND TPLGS "${item}")
+		endforeach()
+	endforeach()
+endforeach()
--- a/tools/topology/topology2/development/tplg-targets-bench.cmake
+++ b/tools/topology/topology2/development/tplg-targets-bench.cmake
@@ -14,6 +14,7 @@ set(components
 	"drc_multiband"
 	"eqiir"
 	"eqfir"
+	"gain"
 )
 
 set(component_parameters
@@ -22,6 +23,7 @@ set(component_parameters
 	"BENCH_DRC_MULTIBAND_PARAMS=default"
 	"BENCH_EQIIR_PARAMS=highpass_50hz_0db_48khz"
 	"BENCH_EQFIR_PARAMS=loudness"
+	"BENCH_GAIN_PARAMS=default"
 )
 
 foreach(p ${platforms})
--- a/tools/topology/topology2/development/tplg-targets-bench.cmake
+++ b/tools/topology/topology2/development/tplg-targets-bench.cmake
@@ -26,6 +26,14 @@ set(component_parameters
 	"BENCH_GAIN_PARAMS=default"
 )
 
+set(components_s32
+	"aria"
+)
+
+set(component_parameters_s32
+	"BENCH_ARIA_PARAMS=default"
+)
+
 foreach(p ${platforms})
 	string(TOLOWER ${p} pl)
 
@@ -36,5 +44,13 @@ foreach(p ${platforms})
 			    #message(STATUS "Item=" ${item})
 			    list(APPEND TPLGS "${item}")
 		endforeach()
+	endforeach()
+
+	# Add components with single format
+	set (sf "32")
+	foreach(comp bench_param IN ZIP_LISTS components_s32 component_parameters_s32)
+		set(item "sof-hda-generic\;sof-hda-benchmark-${comp}${sf}-${pl}\;PLATFORM=${p},HDA_CONFIG=benchmark,BENCH_CONFIG=${comp}${sf},${bench_param}")
+		#message(STATUS "Item=" ${item})
+		list(APPEND TPLGS "${item}")
 	endforeach()
 endforeach()
--- a/tools/topology/topology2/development/tplg-targets-bench.cmake
+++ b/tools/topology/topology2/development/tplg-targets-bench.cmake
@@ -11,6 +11,7 @@ set(sampleformats "16" "24" "32")
 set(components
 	"dcblock"
 	"drc"
+	"drc_multiband"
 	"eqiir"
 	"eqfir"
 )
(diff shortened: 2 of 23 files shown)


marc-hb (commented):
I don't why .conf files should have tabs instead of spaces but I know that this large whitespace commit makes the rest of the PR less convenient to review...

singalsu:
> I don't why .conf files should have tabs instead of spaces but I know that this large whitespace commit makes the rest of the PR less convenient to review...

The topologies build seems to work without this, so I could drop the commit from PR. While being stuck with other issue I did this to eliminate it to cause the build fail. It was in the end one unnecessary include.

tools/topology/topology2/development/tplg-targets.cmake
```diff
@@ -133,6 +133,9 @@ PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-nocodec-bt-mtl-lbm.bin"
 "sof-hda-generic\;sof-hda-benchmark-eqfir16-tgl\;PLATFORM=TGL,HDA_CONFIG=benchmark,BENCH_CONFIG=eqfir16,BENCH_EQFIR_PARAMS=loudness"
 "sof-hda-generic\;sof-hda-benchmark-eqfir24-tgl\;PLATFORM=TGL,HDA_CONFIG=benchmark,BENCH_CONFIG=eqfir24,BENCH_EQFIR_PARAMS=loudness"
 "sof-hda-generic\;sof-hda-benchmark-eqfir32-tgl\;PLATFORM=TGL,HDA_CONFIG=benchmark,BENCH_CONFIG=eqfir32,BENCH_EQFIR_PARAMS=loudness"
+"sof-hda-generic\;sof-hda-benchmark-multiband_drc16-tgl\;PLATFORM=TGL,HDA_CONFIG=benchmark,BENCH_CONFIG=multiband16,BENCH_MULTIBAND_DRC_PARAMS=default"
```
singalsu, line 140: @marc-hb Is there a way to have a PLATFORM loop of TGL, MTL, etc. in cmake? Eventually all of this need to be duplicated to other platforms. Also the 16, 24, 32 format (and float possibly in the future) could be a loop if possible. These should replace sof/tools/test/topology and tools/build_tools/test/topology.

kv2019i (approved):

tools/topology/topology2/include/bench/bench_comp_generate.sh
```diff
@@ -71,32 +71,48 @@ EOF_ROUTE
...
+				#		"default" "include/components/${comp}/default.conf"
+				#	}
+				#}
+				#mixer."1" {
+				#	name '\$ANALOG_PLAYBACK_PCM ${comp^^} switch or volume'
+				#}
+				#enum."1" {
+				#	name '\$ANALOG_PLAYBACK_PCM ${comp^^} enum'
+				#}
 			}
```
lgirdwood, line 92: Meant to be commented out ?

singalsu: Yes, the idea is to un-comment and adapt name of the controls those are supported by the component.

singalsu: Some script could find those out from component class .conf or from module source code but in this step it's manual work.

lgirdwood: ok, this assumption should be commented above the comment block so that everyone knows the usage and future path.

tools/topology/topology2/development/tplg-targets.cmake
marc-hb: > Is there a way to have a PLATFORM loop of TGL, MTL, etc. in cmake?

Yes it should be possible to have a `foreach` loop at the bottom of this file that appends more elements to $TPLGS after the initial, static definition. That should work.

singalsu: Thanks, there's now an effort to do it. Can you please check it if that's a good way to it.

tools/topology/topology2/include/bench/bench_comp_generate.sh
singalsu: Yep, I added the comments.

tools/topology/topology2/development/tplg-targets.cmake
```diff
@@ -147,3 +133,33 @@ PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-tgl-nocodec-crossover.bin,EFX_CROSSOVE
...
+		string(TOLOWER ${p} pl)
+		set(items
+			"sof-hda-generic\;sof-hda-benchmark-dcblock${sf}-${pl}\;PLATFORM=${p},HDA_CONFIG=benchmark,BENCH_CONFIG=dcblock${sf},BENCH_DCBLOCK_PARAMS=default"
+			"sof-hda-generic\;sof-hda-benchmark-drc${sf}-${pl}\;PLATFORM=${p},HDA_CONFIG=benchmark,BENCH_CONFIG=drc${sf},BENCH_DRC_PARAMS=enabled"
+			"sof-hda-generic\;sof-hda-benchmark-eqiir${sf}-${pl}\;PLATFORM=${p},HDA_CONFIG=benchmark,BENCH_CONFIG=eqiir${sf},BENCH_EQIIR_PARAMS=highpass_50hz_0db_48khz"
+			"sof-hda-generic\;sof-hda-benchmark-eqfir${sf}-${pl}\;PLATFORM=${p},HDA_CONFIG=benchmark,BENCH_CONFIG=eqfir${sf},BENCH_EQFIR_PARAMS=loudness"
+			"sof-hda-generic\;sof-hda-benchmark-gain${sf}-${pl}\;PLATFORM=${p},HDA_CONFIG=benchmark,BENCH_CONFIG=gain${sf}"
+			"sof-hda-generic\;sof-hda-benchmark-multiband_drc${sf}-${pl}\;PLATFORM=${p},HDA_CONFIG=benchmark,BENCH_CONFIG=multiband${sf},BENCH_MULTIBAND_DRC_PARAMS=default"
+		)
+		list(APPEND TPLGS "${items}")
```
marc-hb, line 157: Are you sure this works as intended? I know CMake "pseudo-lists" are full of (bad) surprises but I would have expected $items to be unquoted here as in: `list(APPEND TPLGS ${items})`. Right now it looks like it's appending a single and very long string, not appending a list of configs (which is itself an embedded \\; list keywords, oh my...)

There's also a lot of repetition still and I think you can probably kill two birds with one stone. How about this:

```cmake
set(param_suffixes
			"BENCH_CONFIG=dcblock${sf},BENCH_DCBLOCK_PARAMS=default"
			"BENCH_CONFIG=drc${sf},BENCH_DRC_PARAMS=enabled"
    ....
)

foreach(suffix IN ${param_suffixes)
   list APPEND(TPLGS "sof-hda-generic\;sof-hda-benchmark-eqfir${sf}-${pl}\;PLATFORM=${p},HDA_CONFIG=benchmark\;${suffix}")
endforeach()

```

This
- finally gets rid of all the duplication and makes the variable bits much more visible and readable
- This APPENDs a single element at a time which mitigates the usual list&quoting headaches with CMake.

marc-hb (dismissed):
To "divide and conquer", can you create a new, separate `tplg-target-loops.cmake` file and include it from the CMakeLists.txt file? Immediately after the existing one is included.

tools/topology/topology2/development/tplg-targets.cmake
singalsu: > Are you sure this works as intended? I know CMake "pseudo-lists" are full of (bad) surprises but I would have expected $items to be unquoted here as in: `list(APPEND TPLGS ${items})`. Right now it looks like it's appending a single and very long string, not appending a list of configs (which is itself an embedded \; list keywords, oh my...)

I tried it first, of course, It didn't work without quotes. I have no idea why. But I'll try your suggestions, maybe things gets magically fixed.

keqiaozhang:
SOFCI TEST

tools/topology/topology2/development/tplg-targets-bench.cmake
```diff
@@ -0,0 +1,56 @@
...
+set(component_parameters_s32
+	"BENCH_ARIA_PARAMS=default"
+)
+
+foreach(p ${platforms})
+	string(TOLOWER ${p} pl)
+
+	# Add components with all sample formats
+	foreach(sf ${sampleformats})
+		foreach(comp bench_param IN ZIP_LISTS components component_parameters)
```
singalsu, line 42: Suggestion by  @marc-hb , need to bump tools/CMakeLists.txt:cmake_minimum_required(VERSION 3.13) to 3.17 for ZIP_LISTS.

marc-hb: I would have to be `tools/CMakeLists.txt` unfortunately. Right now we have a top-level "tools" CMake project that makes no sense because it's a collection of totally unrelated tools that unfortunately share CMake things like `cmake_minimum_required()` for no good reason.

CMake 3.17 is 2 years old so I think that's not very demanding and acceptable.

marc-hb (commented):
CMake changes LGTM. Please remove the massive whitespace changes.

tools/topology/topology2/CMakeLists.txt
```diff
@@ -1,4 +1,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
+
+# 3.17 required: foreach(<loop_var>... IN ZIP_LISTS <lists>)
+cmake_minimum_required(VERSION 3.17)
```
marc-hb, line 4: Even if you put it here it will still block people trying to build say the logger with CMake 3.16. Just bump this line to 3.99 and try to build the logger and you will see.

So I have mixed feelings about this. It looks local to topology2 but it's not. OK let's say it's useful to have here for 1) documentation purposes, 2) in case someone decouples all the CMakefiles...

marc-hb (commented):
CMake changes approved, dunno about the rest.

lgirdwood (approved):

lgirdwood:
@wszypelt @lrudyX this PR not tested by internal CI, good to merge ?

wszypelt:
@lgirdwood @lrudyX 
I think it will be fine, building and testing went well

PR #8338, opened 2023-10-17
audio: aria: Add set_large_config op

singalsu:
This patch is needed to make the module work with linux and avoid firmware crash. This patch sets the attenuation parameters with the bytes control since the setup parameters can't be passed in init() as in Windows OS.


app/boards/intel_adsp_cavs25_tgph.conf
```diff
@@ -29,6 +29,7 @@ CONFIG_COMP_VOLUME_WINDOWS_FADE=y
 CONFIG_COMP_DRC=y
 CONFIG_COMP_CROSSOVER=y
 CONFIG_COMP_MULTIBAND_DRC=y
+CONFIG_COMP_ARIA=y
```
btian1, line 32: question still same: why choose tgl to enable ARIA?

singalsu: Why not? ARIA is already set to "y" for MTL and after. TGL is the best available generic developer platform for us, the UPX board. We have only high usage remote MTL platforms with usually no proper audio test env for them like I have on my own desk.

app/boards/intel_adsp_cavs25.conf
```diff
@@ -30,6 +30,7 @@ CONFIG_COMP_VOLUME_WINDOWS_FADE=y
 CONFIG_COMP_DRC=y
 CONFIG_COMP_CROSSOVER=y
 CONFIG_COMP_MULTIBAND_DRC=y
+CONFIG_COMP_ARIA=y
```
kv2019i, line 33: I think this needs to be a separate git comit at least. Enabling Aria on tgl/tgl-h is not related to the main commit.

kv2019i (changes requested):
The board config changes need to be split out.

app/boards/intel_adsp_cavs25_tgph.conf
lgirdwood: We need to include new modules today so we can introduce on generic Linux for SOF v2.8

app/boards/intel_adsp_cavs25.conf
singalsu: Yep, it's moved to 2nd commit now!

app/boards/intel_adsp_cavs25.conf
```diff
@@ -30,6 +30,7 @@ CONFIG_COMP_VOLUME_WINDOWS_FADE=y
 CONFIG_COMP_DRC=y
```
kv2019i, line 30: @singalsu checkpatch does not like commits without description. as other checks are green, I think I'll let this pass through, but note for future.

kv2019i (approved):

btian1 (approved):

kv2019i:
Ignoring "sof-ci/jenkins/pr-tools-build", rest of CI checks are good, proceeding with merge.

PR #8342, opened 2023-10-18
ipc4: set fw_ready flag for host

RanderWang:
Set SOF_IPC_INFO_FW_CONTEXT_SAVE for fw_ready event to
host if IMR_CONTEXT_SAVE is supported so that host can sync up with
FW, .e.g. avoid library reload if IMR_CONTEXT_SAVE is supported.

Currently CAVS platforms don't support IMR_CONTEXT_SAVE but ACE
platforms support it. It doesn't affect windows driver since the
fw_ready data is not used by it.

kernel PR: https://github.com/thesofproject/linux/pull/4643


src/ipc/ipc4/handler.c
```diff
@@ -1325,7 +1326,11 @@ struct ipc_cmd_hdr *ipc_prepare_to_send(const struct ipc_msg *msg)
 void ipc_boot_complete_msg(struct ipc_cmd_hdr *header, uint32_t data)
 {
 	header->pri = SOF_IPC4_FW_READY;
-	header->ext = 0;
+
+	if (IS_ENABLED(CONFIG_ADSP_IMR_CONTEXT_SAVE))
+		header->ext = SOF_IPC_INFO_D3_PERSISTENT;
+	else
+		header->ext = 0;
```
kv2019i, line 1333: @mwasko @marcinszkudlinski please comment. Is this ok, or any other way to rely this information to host in IPC4?

plbossart: Not following why we are using the same flag for different firmware behavior.

RanderWang: Introduce a better flag : SOF_IPC4_FW_CONTEXT_SAVE

src/ipc/ipc4/handler.c
```diff
@@ -1325,7 +1325,11 @@ struct ipc_cmd_hdr *ipc_prepare_to_send(const struct ipc_msg *msg)
 void ipc_boot_complete_msg(struct ipc_cmd_hdr *header, uint32_t data)
 {
 	header->pri = SOF_IPC4_FW_READY;
-	header->ext = 0;
```
ranj063, line 1328: why remove this? why not keep it and only have the if below?

RanderWang: sure, I get your idea. Thanks! update.

RanderWang:
update flags only, thanks

src/ipc/ipc4/handler.c
lgirdwood: @RanderWang @plbossart @ujfalusi lets make sure we add in more useful flags here if needed, i.e. we should be able to determine imporant vendor FW features and config based on the FW Kconfig that wont be in manifest.

src/platform/intel/cavs/platform.c
```diff
@@ -67,7 +67,10 @@ static const struct sof_ipc_fw_ready ready
 		.abi_version = SOF_ABI_VERSION,
 		.src_hash = SOF_SRC_HASH,
 	},
-#if CONFIG_CAVS_IMR_D3_PERSISTENT
+#if CONFIG_CAVS_IMR_D3_PERSISTENT && CONFIG_ADSP_IMR_CONTEXT_SAVE
+	.flags = DEBUG_SET_FW_READY_FLAGS | SOF_IPC_INFO_D3_PERSISTENT |
+		SOF_IPC_INFO_CONTEXT_SAVE,
+#elif CONFIG_CAVS_IMR_D3_PERSISTENT
```
ujfalusi, line 73: We are not reading the IPC3 `sof_ipc_fw_ready` message from mailbox in case of IPC4.
Reference firmware do not passes any information via mailbox for FW_READY, iow IPC4 does not define payload for FW_READY.

ujfalusi: The IPC4 way of fw configuration query is via LAGE_CONFIG_GET:FW_CONFIG message. It has definition for various firmware parameters as tlv tuple bundle.
The reference IPC4 FW_READY notification does not have payload and I think if we want to keep the IPC4 name we should not introduce incompatible changes to the basic and common messages, notifications.

ujfalusi: Yes, I agree. An IPC3 (SOF specific) message in the mailbox as payload for IPC4  FW_READY notification is not going to be backwards compatible with the reference, it is not going to be compatible at all since the reference is _not_ placing anything in the mailbox for FW_READY, especially it is not going to place an IPC3 message in there which is part of the IPC3 protocol - in a similar way that IPC4 FW_READY does not have payload.

RanderWang: @ujfalusi why we can't development IPC4 without breaking windows driver ? And if we are stick to current IPC4, we can't use LAGE_CONFIG_GET:FW_CONFIG  since it doesn't support this query and we can't add more config, right ?

lgirdwood: @RanderWang I think direction now is to take this flag as a manifest flag. We should make it clear and add a comment to the FW ready message that the fields in FW ready should only be for runtime generated information (i.e. build time information would go in the manifest). Thanks !

RanderWang:
no need

PR #8347, opened 2023-10-18
topology2: Enable Google AEC for sof-mtl-max98357a-rt5682

yongzhi1:
These two configs are what we are using to enable Google AEC currently.


yongzhi1:
Sorry wrong branch being used.

PR #8350, opened 2023-10-19
Re-organize EQ IIR header and split source code with main/generic/ipc3/ipc4

btian1:
eq iir code refine.
1. move header from include path to eq iir directory.
2. create generic/ipc3/ipc4 source files and move code to source files accordingly.


src/audio/CMakeLists.txt
```diff
@@ -175,9 +175,11 @@ set(sof_audio_modules mixer volume src asrc eq-fir eq-iir dcblock crossover tdfb
 if(CONFIG_IPC_MAJOR_3)
```
singalsu, line 175: Typo in commit message "mvoe". Also start sentences with capital letter.

src/audio/eq_iir/eq_iir_ipc3.c
```diff
@@ -0,0 +1,336 @@
...
+#include <ipc/topology.h>
+#include <user/eq.h>
+#include <user/trace.h>
+#include <errno.h>
+#include <stddef.h>
+#include <stdint.h>
+
+LOG_MODULE_DECLARE(eq_iir, CONFIG_SOF_LOG_LEVEL);
+
+#if CONFIG_FORMAT_S32LE && CONFIG_FORMAT_S16LE
```
singalsu, line 41: You could add comment that format conversion capability is for now only in IPC3 version, if that was the purpose of having these functions here.

src/audio/eq_iir/eq_iir_ipc4.c
```diff
@@ -0,0 +1,135 @@
...
+#include <ipc/topology.h>
+#include <user/eq.h>
+#include <user/trace.h>
+#include <errno.h>
+#include <stddef.h>
+#include <stdint.h>
+
+LOG_MODULE_DECLARE(eq_iir, CONFIG_SOF_LOG_LEVEL);
+
+static eq_iir_func eq_iir_find_func(struct processing_module *mod)
```
singalsu, line 41: Also add comment that ipc4 functions finding is different because of different formats handling capability. I think nothing technically would prevent to have e.g. s32 to s16 conversion in IPC4 as well. It wasn't possible or working when I converted the module but now things have been fixed.

singalsu: Some explanation why IIR has format conversion:

In early days of SOF the preference for pipelines was 16 bits to save RAM in platforms like Baytrail. However in microphone paths if there was need to digitally boost the gain the quality was bad in topologies where capture DAI was 16 bit and we applied with volume or IIR about 20 dB gain. In practice a 16 bit word got left shifted by some bit positions that effectively made signal like 12 bits. We could achieve a lot better quality by capturing codec and DAI with 24 or 32bits and applying the gain in IIR for the larger word length. Then all 16 bits in the pipelines after DAI and IIR had signal. The IIR was chosen for format conversion because it also canceled the sometimes large DC component (and some lowest non-audible frequencies) in signal. It gave the headroom for signal for amplification.

If IPC4 systems ever need the memory save small 16 bit capture paths the format conversion could be brought back.

src/audio/eq_iir/eq_iir_ipc3.c
btian1: move them to eq_iir_ipc3.c means this is for IPC3 only.

src/audio/eq_iir/eq_iir_ipc4.c
btian1: thanks, will add

src/audio/eq_iir/eq_iir_ipc3.c
singalsu: Yep, this is OK since the component capabilities are different for ipc3 and ipc4.

marc-hb:
Nit: you have pretty good commit messages so you can do better than calling this PR "... refine". "Refine" is almost like "enhance" which means... nothing. It's very rare to make things worse, every PR makes things better :-)

Trying to summarize your commit messages: "Re-organize EQ IIR headers and common code". Something like that.

Also: keep acronyms in upper case. Acronyms are never in lower case, this is confusing.

My 2 cents.

src/audio/eq_iir/eq_iir.c
```diff
@@ -6,11 +6,11 @@
 //         Liam Girdwood <liam.r.girdwood@linux.intel.com>
 //         Keyon Jie <yang.jie@linux.intel.com>
 
+#include "eq_iir.h"
```
kv2019i, line 9: Typically non-system include files come after system include files, so unless there's a reason to have eq_iir.h as first, I'd move this to L38 (or around there).

kv2019i (approved):
The conversion functions indeed make the IPC3 specific code look larger than it really it, but with the comment added, I think this is ok. One change for header include order, but otherwise looks good.

src/audio/CMakeLists.txt
singalsu: You are missing capital letter sentence starts from other commits as well,  this time ok for me but try to remember to write well the commit messages.

singalsu (approved):

lgirdwood (approved):

PR #8351, opened 2023-10-19
Audio: Fix the default alignment setting

andrula-song:
Set the default alignment value after buffer format is set in audio_stream_set_params instead of in audio_stream_init where the buffer format hasn't been set yet.

### DIFF
--- a/src/include/sof/audio/audio_stream.h
+++ b/src/include/sof/audio/audio_stream.h
@@ -349,25 +349,6 @@ static inline void audio_stream_set_buffer_fmt(struct audio_stream *buf,
 	audio_stream_wrap(buffer, (char *)(ptr) + ((idx) * (sample_size)))
 
 /**
- * Applies parameters to the buffer.
- * @param buffer Buffer.
- * @param params Parameters (frame format, rate, number of channels).
- * @return 0 if succeeded, error code otherwise.
- */
-static inline int audio_stream_set_params(struct audio_stream *buffer,
-					  struct sof_ipc_stream_params *params)
-{
-	if (!params)
-		return -EINVAL;
-
-	buffer->runtime_stream_params.frame_fmt = params->frame_fmt;
-	buffer->runtime_stream_params.rate = params->rate;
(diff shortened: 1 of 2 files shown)


andrula-song:
SOFCI TEST

lgirdwood (approved):

btian1 (approved):

andrula-song:
SOFCI TEST

src/include/sof/audio/audio_stream.h
```diff
@@ -433,6 +414,31 @@ static inline void audio_stream_init_alignment_constants(const uint32_t byte_ali
...
+
+	buffer->runtime_stream_params.frame_fmt = params->frame_fmt;
+	buffer->runtime_stream_params.rate = params->rate;
+	buffer->runtime_stream_params.channels = params->channels;
+
+	/* set the default alignment info.
+	 * set byte_align as 1 means no alignment limit on byte.
+	 * set frame_align as 1 means no alignment limit on frame.
+	 */
+	audio_stream_init_alignment_constants(1, 1, buffer);
```
kv2019i, line 437: Hmm, it's difficult to confirm this is safe to do. audio_stream_init() is only called once but audio_stream_set_params() is called from many places via buffer_set_params().

Have you checked that we are not overriding alignment info? 

Not saying old code is correct. It would look we'd at least need to recalculate runtime_stream_params.align_frame_cnt and align_shift_idx if buffer parameters are changed.

kv2019i (commented):
One comment to clarify (see inline).

src/include/sof/audio/audio_stream.h
andrula-song: usually, we set alignment info in prepare function(after internal parameter set and verify) for specific component, I checked the src, eq-iir,eq-fir and volume,aria .etc components, they will change the default value to their own setting.
and yes, once the buffer parameters are changed, we do need to recalculate the align info.

aiChaoSONG:
SOFCI TEST

PR #8352, opened 2023-10-19
Src format condense in tplg file

btian1:


tools/topology/topology2/include/components/src_format_s32_convert_from_48k.conf
```diff
@@ -1,94 +1,26 @@
...
-					out_rate                 96000
-					out_bit_depth            32
-					out_valid_bit_depth      32
-				}
-				# 176.4khz output
-				{
-					out_rate                 176400
-					out_bit_depth            32
-					out_valid_bit_depth      32
-					obs "$[($out_channels * (($[($out_rate + 999)] / 1000) + 4)) * ($out_bit_depth / 8)]"
```
ranj063, line 85: some of these rates have an obs set that you seem to have removed @btian1

tools/topology/topology2/include/components/src_format_s32_to_sxx_convert.conf
```diff
@@ -1,346 +1,57 @@
...
+						192000
+					]
+					out_bit_depth [
+						16
+						32
+					]
+					out_valid_bit_depth [
+						16
+						24
+						32
```
ranj063, line 54: @btian1 we cant do this. This would make several invalid combinations such as 16 for bit_depth and 23 for valid_bit_depth. So please separate them out into 3 different combinearray definitions

tools/topology/topology2/include/components/src_format_sxx_to_s32_convert.conf
```diff
@@ -2,345 +2,56 @@
...
+						96000
+						176400
+						192000
+					]
+					in_bit_depth [
+						16
+						32
+					]
+					in_valid_bit_depth [
+						16
```
ranj063, line 27: same thing here. cannot combine the values of 16/32 for bit_depth with 16/24/32 for valid bit depth

tools/topology/topology2/include/components/src_format_sxx_to_s32_convert.conf
```diff
@@ -2,345 +2,56 @@
 			num_input_audio_formats 42
 
 			Object.Base.input_audio_format [
-				# 8khz input
```
ranj063, line 5: and please also fix the prefix in the commit message to "topology2:..."

tools/topology/topology2/include/components/src_format_s32_convert_from_48k.conf
btian1: sorry, missed this part.

tools/topology/topology2/include/components/src_format_s32_to_sxx_convert.conf
btian1: if separate, is it 3 or 6? since out_bit_depth also need a separation?
Also, will the invalid combination will be used with error or just put it there no use?

ranj063: you might have to figure out the right combination so that we dont end up with invalid audio formats and the exact number as there is currently

tools/topology/topology2/include/components/src_format_s32_convert_from_48k.conf
lgirdwood: @btian1 @ranj063 btw, obs is changing name to be more meaningful. Please check in with @marcinszkudlinski for details.

btian1: @lgirdwood @marcinszkudlinski , is it name change or calculation change different with current implementation?
we can use a separate patch to address this change.

tools/topology/topology2/include/components/src_format_s32_convert_from_48k.conf
```diff
@@ -1,95 +1,28 @@
...
-					out_rate                 8000
-					out_bit_depth            32
-					out_valid_bit_depth      32
-				}
-				# 11.025 khz output
-				{
-					out_rate                 11025
-					out_bit_depth            32
-					out_valid_bit_depth      32
-					obs "$[($out_channels * (($[($out_rate + 999)] / 1000) + 4)) * ($out_bit_depth / 8)]"
```
marcinszkudlinski, line 16: if SRC is run as DP it will be scheduled when 
1) has at least IBS bytes of data on input 
2) has at least OBS free bytes of data on output

so IBS should be exact 
OBS is less important - as buffer is allocated of size 2*MAX(ibs_of_next_module, OBS_of_current_module)
so + 999 is a waste of memory - please set precise value


i.e. if SRC is changing 16000/16/2 to 48000/16/2 and you want to process data in 1ms periods
   IBS = 64, OBS = 192

in the same processing, but if you want to process in 10ms chunks:
IBS=640 OBS=1920

tools/topology/topology2/include/components/src_format_s32_convert_from_48k.conf
ranj063: @btian1 sorry but what you have now is also wrong. The new value of obs is only set for 3 of the frequencies in the original file and not for all frequencies.

tools/topology/topology2/include/components/src_format_sxx_to_s32_convert.conf
ranj063: @bitan the prefix is meant as an indication of the path to the file that is being changed.Can you please remove audio and expand to Topology2 ? POlease look at the log for other patches to see the right prefix

tools/topology/topology2/include/components/src_format_s32_convert_from_48k.conf
btian1: Thanks, @marcinszkudlinski , @singalsu , what's your comments on this? I would like use a separate patch to update obs, in case there is regression for new setting.

ranj063 (approved):

tools/topology/topology2/include/components/src_format_s32_convert_from_48k.conf
ranj063: @btian1 I asgree, this PR is purely for the purpose of using combinearray definition instead of separate audio formats. So its important to keep it identical to the existing audio formats. Any changes to the ibs/obs should be a follow up

btian1:
SOFCI TEST

lgirdwood (approved):
@singalsu @andrula-song pls review for SRC config.
@btian1 ack - we can follow up obs/ibs alignment in next PR.

kv2019i (changes requested):
I'd at least want @singalsu to ack this before merge as this seems to change the formula.

tools/topology/topology2/include/components/src_format_s32_convert_from_48k.conf
ranj063: and it is retained with this PR @kv2019i

tools/topology/topology2/include/components/src_format_s32_convert_from_48k.conf
```diff
@@ -1,95 +1,34 @@
...
-				# 88.2khz output
-				{
-					out_rate                 88200
-					out_bit_depth            32
-					out_valid_bit_depth      32
+			CombineArrays.Object.Base.output_audio_format [
+				{
+					out_rate [
+						8000
+						11025
```
andrula-song, line 8: shouldn't we put this 11.025kHz to the next group?

btian1: wonder why 22050 does not have obs setting @singalsu

tools/topology/topology2/include/components/src_format_s32_convert_from_48k.conf
kv2019i: I stand corrected, so it is.

kv2019i (approved):

btian1:
@singalsu , please review this combined SRC tplg PR.
Btw, please confirm in first patch, there may have a missed obs setting bug for 22050hz, I will try to fix it and re-calculate obs in next PR.

tools/topology/topology2/include/components/src_format_s32_convert_from_48k.conf
```diff
@@ -1,95 +1,34 @@
...
+					out_bit_depth [ 32 ]
+					out_valid_bit_depth [ 32 ]
+				}
+				{
+					out_rate [
+						11025
+						44100
+						88200
+						176400
+					]
```
singalsu, line 27: I think 22050 should also belong here because the 1 ms period does not have integer samples count. It's missing from original topology, but it looks it has been missed.

singalsu (changes requested):

singalsu (approved):
Changed my mind, this produces same formats tables as before. Yes please use the same equation for 22050 Hz in next PR as for other 44.1 kHz family rates.

PR #8359, opened 2023-10-20
topology2: add cs42l43 + cs35l56 topology for MTL and TGL

bardliao:
topology2: add cs42l43 + cs35l56 topology for MTL and TGL

### DIFF
--- a/tools/topology/topology2/avs-tplg/tplg-targets.cmake
+++ b/tools/topology/topology2/avs-tplg/tplg-targets.cmake
@@ -34,6 +34,11 @@ PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-tgl-rt715-rt711-rt1308-mono.bin"
 DMIC1_ID=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,SDW_AMP_FEEDBACK=false,\
 PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-tgl-rt711-rt1308-4ch.bin"
 
+"cavs-sdw\;sof-tgl-cs42l43-l3-cs35l56-l01\;PLATFORM=tgl,NUM_SDW_AMP_LINKS=2,SDW_DMIC=1,\
+NUM_HDMIS=0,SDW_SPK_STREAM=Playback-SmartAmp,SDW_SPK_IN_STREAM=Capture-SmartAmp,\
+SDW_DMIC_STREAM=Capture-SmartMic,SDW_JACK_OUT_STREAM=Playback-SimpleJack,\
+SDW_JACK_IN_STREAM=Capture-SimpleJack,SDW_AMP_FMT_24=true,SDW_JACK_FMT_24=true"
+
 # IPC4 topology for TGL rt712 Headset, Amp and DMIC
 "cavs-sdw\;sof-tgl-rt712\;SDW_JACK_OUT_STREAM=SDW0-Playback-SimpleJack,\
 SDW_JACK_IN_STREAM=SDW0-Capture-SimpleJack,SDW_SPK_STREAM=SDW0-Playback-SmartAmp,\
--- a/tools/topology/topology2/platform/intel/sdw-amp-generic.conf
+++ b/tools/topology/topology2/platform/intel/sdw-amp-generic.conf
@@ -9,6 +9,7 @@ Define {
 	SDW_AMP_IN_BE_ID 3
 	AMP_FEEDBACK_CH 2
 	SDW_AMP_FEEDBACK true
+	SDW_AMP_FMT_24 false
 }
 
 Object.Dai.ALH [
@@ -54,6 +55,19 @@ IncludeByKey.PASSTHROUGH {
 				Object.Widget.alh-copier.1 {
 					stream_name $SDW_SPK_STREAM
 					node_type $ALH_LINK_OUTPUT_CLASS
+					IncludeByKey.SDW_AMP_FMT_24 {
+					"true"	{
+							num_output_audio_formats 1
+							Object.Base.output_audio_format [
+								{
+									out_bit_depth           32
+									out_valid_bit_depth     24
+									out_sample_type		$SAMPLE_TYPE_MSB_INTEGER
+									out_fmt_cfg		"$[($out_channels | ($out_valid_bit_depth * 256))]"
+								}
+							]
+						}
+					}
 				}
 				Object.Widget.gain.1 {
 					Object.Control.mixer.1 {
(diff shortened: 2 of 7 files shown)


tools/topology/topology2/platform/intel/sdw-amp-generic.conf
```diff
@@ -54,6 +55,19 @@ IncludeByKey.PASSTHROUGH {
 				Object.Widget.alh-copier.1 {
 					stream_name $SDW_SPK_STREAM
 					node_type $ALH_LINK_OUTPUT_CLASS
+					IncludeByKey.SDW_AMP_FMT_24 {
```
kv2019i, line 58: @bardliao I'm a bit puzzled by this. The kernel has special logic if only one format is defined by copier (DAI link format mask is forced to this format in this case). But why doesn't this work with multiple formats in copier. If the codec only supports S24_LE (32bit-with-24validbits), format should be constrained already.

@ranj063 Is this correct? Looking at kernel sof_ipc4_prepare_copier_module(), dai copier should always be limited to single output format (as there's no logic to decide which one to use if there is multiple).

bardliao: > If the codec only supports S24_LE (32bit-with-24validbits), format should be constrained already.

@kai In theory, yes. But, somehow AlSA may still select 32 bit format even the codec driver declares it does not support 32 bit format. That's why we call `snd_mask_set_format` in `sof_ipc4_pcm_dai_link_fixup` to specify the BE format.

lgirdwood (approved):
@RanderWang @ranj063 @jsarha in general we need to make sure we add more comments in our topologies like we do with our C cide, especially around any conditional logic.

tools/topology/topology2/platform/intel/sdw-amp-generic.conf
RanderWang: This change fixes output format to S24_LE. why ?  Kernel driver can't choose S24_LE in 3 format array ?

bardliao: > This change fixes output format to S24_LE. why ? Kernel driver can't choose S24_LE in 3 format array ?

Because cs42l43 and cs35l56 can't support 32 bit format. So that we want to use 24 bit format.

bardliao: Chatted with @RanderWang, alternative way is to use 24 bit format for all sdw codecs. There seems to be no much benefit to use 32 bit format. What do you think? @kv2019i @ranj063 @plbossart

plbossart: doesn't the Intel hardware require 32 bits anyways, so why would we restrict the number of valid bits?

bardliao: > doesn't the Intel hardware require 32 bits anyways, so why would we restrict the number of valid bits?

@plbossart The purpose is to let kernel to use the 24 bit format. The related kernel change is https://github.com/thesofproject/linux/pull/4446

bardliao: @kv2019i I need to clarify that the format here means the BE format. Based on my experiment, ALSA will use the audio file's format if no one try to set the format. That is the case for IPC4 before https://github.com/thesofproject/linux/pull/4446 was merged. For example, it will be 32 bit format if we play a 32 bit format .wav file, and it will be 24 bit format if we play a 24 bit .wav file.
I agree that ALSA should not use the format that it is not supported by the codec. Maybe we can add a mechanism to select the format that is supported by both cpu and codec. But it will be a big task.
We set the BE format if there is only one format available in the dai copier after https://github.com/thesofproject/linux/pull/4446 was merged. 
And we have set the BE format in IPC3 in  [sof_ipc3_pcm_dai_link_fixup](https://github.com/thesofproject/linux/blob/topic/sof-dev/sound/soc/sof/ipc3-pcm.c#L237) for a while.

bardliao: @kv2019i Yes, always using 24 bit BE format makes sense, and much simpler. The only open in my mind is that how does kernel know a specific codec want to use 32 bit format?
I looked at `hdac_hda_dai_hw_params`, and found it is not my intention. The params_format(params) is still S32_LE, not S24_LE although `maxbps` is 24. It is more like that ASoC says hey let's use S32_LE format, but the HDA codec driver changes the format by itself. I don't think we can do it for all SDW/I2S codec drivers.
IMHO, the right way is let params_format(params) become S24_LE, and codec drivers will set the format configuration according to params_format(params).
After removing `snd_mask_none` and `snd_mask_set_format` from `sof_ipc3_pcm_dai_link_fixup`, the params_format(params) became the same format as what I play. I need some time to figure out where does params_format(params) be decided, but I think using `snd_mask_set_format` is the right way to config the BE format.
Back to this PR, what would you prefer?
1. Change the default SDW format to S24_LE in topology, and kernel set the BE format according to topology's format.
2. Kernel always set S24_LE format for SDW. Topology can use any format it wants and the format will be used in SOF.

lgirdwood:
@kv2019i @ranj063 comments addressed ?

kv2019i (approved):

tools/topology/topology2/platform/intel/sdw-amp-generic.conf
```diff
@@ -54,6 +55,19 @@ IncludeByKey.PASSTHROUGH {
...
+					"true"	{
+							num_output_audio_formats 1
+							Object.Base.output_audio_format [
+								{
+									out_bit_depth           32
+									out_valid_bit_depth     24
+									out_sample_type		$SAMPLE_TYPE_MSB_INTEGER
```
kv2019i, line 65: @bardliao Sorry for late follow-up. But I'm a bit confused again, this format "MSB_INTEGER" is not S32_LE (24 valid bits left aligned) but rather S24_LE (24bits right-aligned). So what's the format the audio codec does not support?

kv2019i (changes requested):
Change of mind, see inline.

tools/topology/topology2/platform/intel/sdw-amp-generic.conf
kv2019i: And answer from @bardliao  that it's specifically the case of valid_bit_depth 32 not supported by some of the codecs.

tools/topology/topology2/platform/intel/sdw-amp-generic.conf
```diff
@@ -130,21 +144,36 @@ IncludeByKey.PASSTHROUGH {
...
+						num_output_audio_formats 3
+						Object.Base.output_audio_format [
+							{
+								out_bit_depth           16
+								out_valid_bit_depth     16
+							}
+							{
+								out_bit_depth           32
+								out_valid_bit_depth     24
+							}
```
kv2019i, line 169: Should the SAMPLE_TYPE_MSG_INTEGER defs be added to this definition as well?

kv2019i (commented):
@bardliao So end-of-day summary, if you add the MSB_INTEGER to both true and false cases (for 24/32bit), I'm good to go with this. I now understand we may need to signal both options in tplg.

RanderWang (approved):

tools/topology/topology2/platform/intel/sdw-amp-generic.conf
bardliao: > Should the SAMPLE_TYPE_MSG_INTEGER defs be added to this definition as well?

Yes, done.

tools/topology/topology2/avs-tplg/tplg-targets.cmake
```diff
@@ -34,6 +34,10 @@ PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-tgl-rt715-rt711-rt1308-mono.bin"
 DMIC1_ID=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,SDW_AMP_FEEDBACK=false,\
 PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-tgl-rt711-rt1308-4ch.bin"
 
+"cavs-sdw\;sof-tgl-cs42l43-l3-cs35l56-l01\;PLATFORM=tgl,NUM_SDW_AMP_LINKS=2,SDW_DMIC=1,\
+NUM_HDMIS=0,SDW_SPK_STREAM=Playback-SmartAmp,SDW_SPK_IN_STREAM=Capture-SmartAmp,SDW_DMIC_STREAM=Capture-SmartMic,SDW_JACK_OUT_STREAM=Playback-SimpleJack,\
```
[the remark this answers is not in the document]
bardliao: > this line is too long, please rework it

fixed

tools/topology/topology2/sof-ace-tplg/tplg-targets.cmake
```diff
@@ -40,6 +40,10 @@ SDW_DMIC_STREAM=SDW0-Capture"
 # Jack codec + SmartAmp topology. No SDW_DMIC connection
 "cavs-sdw\;sof-mtl-rt713-l0-rt1316-l12\;PLATFORM=mtl,NUM_SDW_AMP_LINKS=2"
 
+"cavs-sdw\;sof-mtl-cs42l43-l0-cs35l56-l12\;PLATFORM=mtl,NUM_SDW_AMP_LINKS=2,SDW_DMIC=1,\
+NUM_HDMIS=0,SDW_SPK_STREAM=Playback-SmartAmp,SDW_SPK_IN_STREAM=Capture-SmartAmp,SDW_DMIC_STREAM=Capture-SmartMic,SDW_JACK_OUT_STREAM=Playback-SimpleJack,\
```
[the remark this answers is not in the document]
bardliao: > too long also

fixed

kv2019i (approved):
Thanks @bardliao , looks good now!

kv2019i:
@ranj063 @plbossart Final call on this soon...

PR #8361, opened 2023-10-20
lnl: sndw: use HDA DMA to transfer stream data 

iganakov:
Since LNL SoundWire interface uses HDA DMA to transfer stream data. 

This approach uses HDA dai and HDA DMA. No changes needed on Zephyr side.
Following solution involves FW based SoundWire link aggregation mechanism which
is also used on MTL for backward compatibility with Windows driver.

This is a working solution. It was tested in E2E scenarios with Windows driver.

### DIFF
--- a/src/audio/copier/copier_dai.c
+++ b/src/audio/copier/copier_dai.c
@@ -64,6 +64,96 @@ static int copier_set_alh_multi_gtw_channel_map(struct comp_dev *dev,
 	return 0;
 }
 
+static int copier_alh_assign_dai_index(struct comp_dev *dev,
+				       void *gtw_cfg_data,
+				       union ipc4_connector_node_id node_id,
+				       struct ipc_config_dai *dai,
+				       int *dai_index,
+				       int *dai_count)
+{
+	struct processing_module *mod = comp_get_drvdata(dev);
+	struct copier_data *cd = module_get_private_data(mod);
+	const struct sof_alh_configuration_blob *alh_blob = gtw_cfg_data;
+	uint8_t *dma_config;
+	size_t alh_cfg_size, dma_config_length;
+	int i, dai_num, ret;
+
+	if (!cd->config.gtw_cfg.config_length) {
+		comp_err(mod->dev, "No gateway config found in blob!");
+		return -EINVAL;
+	}
+
+	switch (dai->type) {
+	case SOF_DAI_INTEL_HDA:
+		/* We use DAI_INTEL_HDA for ACE 2.0 platforms */
+		alh_cfg_size = get_alh_config_size(alh_blob);
+		dma_config = (uint8_t *)gtw_cfg_data + alh_cfg_size;
+		dma_config_length = (cd->config.gtw_cfg.config_length << 2) - alh_cfg_size;
+
+		/* Here we check node_id if we need to use FW aggregation,
+		 * in other words do we need to create multiple dai or not
+		 */
+		if (!is_multi_gateway(node_id)) {
+			/* Find DMA config in blob and retrieve stream_id */
+			ret = ipc4_find_dma_config_multiple(dai, dma_config, dma_config_length,
+							    alh_blob->alh_cfg.mapping[0].alh_id, 0);
+			if (ret != 0) {
+				comp_err(mod->dev, "No sndw dma_config found in blob!");
+				return -EINVAL;
+			}
+			dai_index[0] = dai->host_dma_config[0]->stream_id;
+			return 0;
+		}
+
+		dai_num = alh_blob->alh_cfg.count;
+		if (dai_num > IPC4_ALH_MAX_NUMBER_OF_GTW || dai_num < 0) {
+			comp_err(mod->dev, "Invalid dai_count: %d", dai_num);
+			return -EINVAL;
+		}
+
+		for (i = 0; i < dai_num; i++) {
+			ret = ipc4_find_dma_config_multiple(dai, dma_config,
+							    dma_config_length,
+							    alh_blob->alh_cfg.mapping[i].alh_id, i);
+			if (ret != 0) {
+				comp_err(mod->dev, "No sndw dma_config found in blob!");
+				return -EINVAL;
+			}
+
+			/* To process data on SoundWire interface HD-A DMA is used so it seems
+			 * logical to me to use stream tag as a dai_index instead of PDI.
+			 */
+			dai_index[i] = dai->host_dma_config[i]->stream_id;
+		}
+
+		*dai_count = dai_num;
+		break;
+	case SOF_DAI_INTEL_ALH:
+		/* Use DAI_INTEL_ALH for ACE 1.0 and older */
+		if (!is_multi_gateway(node_id)) {
+			dai_index[0] = IPC4_ALH_DAI_INDEX(node_id.f.v_index);
+			return 0;
+		}
+
+		dai_num = alh_blob->alh_cfg.count;
+		if (dai_num > IPC4_ALH_MAX_NUMBER_OF_GTW || dai_num < 0) {
+			comp_err(mod->dev, "Invalid dai_count: %d", dai_num);
+			return -EINVAL;
+		}
+
+		for (i = 0; i < dai_num; i++)
+			dai_index[i] = IPC4_ALH_DAI_INDEX(alh_blob->alh_cfg.mapping[i].alh_id);
+
+		*dai_count = dai_num;
+		break;
+	default:
+		comp_err(mod->dev, "Invalid dai type selected: %d", dai->type);
+		return -EINVAL;
+	}
+
+	return 0;
+}
+
 static int copier_dai_init(struct comp_dev *dev,
 			   struct comp_ipc_config *config,
 			   const struct ipc4_copier_module_cfg *copier,
@@ -100,7 +190,8 @@ static int copier_dai_init(struct comp_dev *dev,
 	}
 
 	/* save the channel map and count for ALH multi-gateway */
-	if (type == ipc4_gtw_alh && is_multi_gateway(copier->gtw_cfg.node_id)) {
+	if ((type == ipc4_gtw_alh || type == ipc4_gtw_link) &&
+	    is_multi_gateway(copier->gtw_cfg.node_id)) {
 		ret = copier_set_alh_multi_gtw_channel_map(dev, copier, index);
 		if (ret < 0)
 			return ret;
@@ -182,50 +273,19 @@ int copier_dai_create(struct comp_dev *dev, struct copier_data *cd,
 		break;
 	case ipc4_alh_link_output_class:
 	case ipc4_alh_link_input_class:
+#if defined(CONFIG_ACE_VERSION_2_0)
+		dai.type = SOF_DAI_INTEL_HDA;
+		dai.is_config_blob = true;
+		type = ipc4_gtw_link;
+#else
 		dai.type = SOF_DAI_INTEL_ALH;
 		dai.is_config_blob = true;
 		type = ipc4_gtw_alh;
-
-		/* copier
-		 * {
-		 *  gtw_cfg
-		 *  {
-		 *     gtw_node_id;
-		 *     config_length;
-		 *     config_data
-		 *     {
-		 *	   count;
-		 *	  {
-		 *	     node_id;  \\ normal gtw id
-		 *	     mask;
-		 *	  }  mapping[MAX_ALH_COUNT];
-		 *     }
-		 *   }
-		 * }
-		 */
-		 /* get gtw node id in config data */
-		if (is_multi_gateway(node_id)) {
-			if (copier->gtw_cfg.config_length) {
-				const struct sof_alh_configuration_blob *alh_blob =
-					(const struct sof_alh_configuration_blob *)
-						copier->gtw_cfg.config_data;
-
-				dai_count = alh_blob->alh_cfg.count;
-				if (dai_count > IPC4_ALH_MAX_NUMBER_OF_GTW || dai_count < 0) {
-					comp_err(dev, "Invalid dai_count: %d", dai_count);
-					return -EINVAL;
-				}
-				for (i = 0; i < dai_count; i++)
-					dai_index[i] =
-					IPC4_ALH_DAI_INDEX(alh_blob->alh_cfg.mapping[i].alh_id);
-			} else {
-				comp_err(dev, "No ipc4_alh_multi_gtw_cfg found in blob!");
-				return -EINVAL;
-			}
-		} else {
-			dai_index[dai_count - 1] = IPC4_ALH_DAI_INDEX(node_id.f.v_index);
-		}
-
+#endif /* defined(CONFIG_ACE_VERSION_2_0) */
+		ret = copier_alh_assign_dai_index(dev, cd->gtw_cfg, node_id,
+						  &dai, dai_index, &dai_count);
+		if (ret)
+			return ret;
 		break;
 	case ipc4_dmic_link_input_class:
 		dai.type = SOF_DAI_INTEL_DMIC;
@@ -425,7 +485,6 @@ int copier_dai_params(struct copier_data *cd, struct comp_dev *dev,
 				cd->converter[IPC4_COPIER_GATEWAY_PIN];
 		return ret;
 	}
-
 	/* For ALH multi-gateway case, params->channels is a total multiplexed
 	 * number of channels. Demultiplexed number of channels for each individual
 	 * gateway comes in blob's struct ipc4_alh_multi_gtw_cfg.
--- a/src/ipc/ipc4/dai.c
+++ b/src/ipc/ipc4/dai.c
@@ -26,6 +26,7 @@
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
+#include <sof/audio/module_adapter/module/generic.h>
 
 #include "../audio/copier/copier.h"
 #include "../audio/copier/dai_copier.h"
@@ -84,6 +85,29 @@ int dai_config_dma_channel(struct dai_data *dd, struct comp_dev *dev, const void
 #endif
 		break;
 	case SOF_DAI_INTEL_HDA:
+#if defined(CONFIG_ACE_VERSION_2_0)
+		if (copier_cfg->gtw_cfg.node_id.f.dma_type == ipc4_alh_link_output_class ||
+		    copier_cfg->gtw_cfg.node_id.f.dma_type == ipc4_alh_link_input_class) {
+			struct processing_module *mod = comp_get_drvdata(dev);
+			struct copier_data *cd = module_get_private_data(mod);
+
+			if (!cd->gtw_cfg) {
+				comp_err(dev, "No gateway config found!");
+				return DMA_CHAN_INVALID;
+			}
+
+			channel = DMA_CHAN_INVALID;
+			const struct sof_alh_configuration_blob *alh_blob = cd->gtw_cfg;
+
+			for (int i = 0; i < alh_blob->alh_cfg.count; i++) {
+				if (dai->host_dma_config[i]->stream_id == dai->dai_index) {
+					channel = dai->host_dma_config[i]->dma_channel_id;
+					break;
+				}
+			}
+			break;
+		}
+#endif /* defined(CONFIG_ACE_VERSION_2_0) */
 		channel = copier_cfg->gtw_cfg.node_id.f.v_index;
 		break;
 	case SOF_DAI_INTEL_ALH:
--- a/src/ipc/ipc4/dai.c
+++ b/src/ipc/ipc4/dai.c
@@ -44,7 +44,7 @@ void dai_set_link_hda_config(uint16_t *link_config,
 	case SOF_DAI_INTEL_SSP:
 		link_cfg.full = 0;
 		link_cfg.part.dir = common_config->direction;
-		link_cfg.part.stream = common_config->host_dma_config->stream_id;
+		link_cfg.part.stream = common_config->host_dma_config[0]->stream_id;
 		break;
 	case SOF_DAI_INTEL_DMIC:
 		link_cfg.full = 0;
@@ -57,7 +57,7 @@ void dai_set_link_hda_config(uint16_t *link_config,
 		} else {
 			link_cfg.part.hchan = out_fmt->channels_count - 1;
 		}
-		link_cfg.part.stream = common_config->host_dma_config->stream_id;
+		link_cfg.part.stream = common_config->host_dma_config[0]->stream_id;
 		break;
 	default:
 		/* other types of DAIs not need link_config */
@@ -79,8 +79,8 @@ int dai_config_dma_channel(struct dai_data *dd, struct comp_dev *dev, const void
 	case SOF_DAI_INTEL_DMIC:
 		channel = 0;
 #if defined(CONFIG_ACE_VERSION_2_0)
-		if (dai->host_dma_config->pre_allocated_by_host)
-			channel = dai->host_dma_config->dma_channel_id;
+		if (dai->host_dma_config[0]->pre_allocated_by_host)
+			channel = dai->host_dma_config[0]->dma_channel_id;
 #endif
 		break;
 	case SOF_DAI_INTEL_HDA:
--- a/src/ipc/ipc4/helper.c
+++ b/src/ipc/ipc4/helper.c
@@ -44,6 +44,7 @@
 #include <ipc4/module.h>
 #include <ipc4/error_status.h>
 #include <sof/lib_manager.h>
+#include <sof/tlv.h>
 
 #include <errno.h>
 #include <stdbool.h>
@@ -1046,6 +1047,35 @@ int ipc4_find_dma_config(struct ipc_config_dai *dai, uint8_t *data_buffer, uint3
 	dai->host_dma_config[0] = GET_IPC_DMA_CONFIG(data_buffer, size);
 #endif
 	return IPC4_SUCCESS;
+}
+
+int ipc4_find_dma_config_multiple(struct ipc_config_dai *dai, uint8_t *data_buffer,
+				  uint32_t size, uint32_t device_id, int dma_cfg_idx)
+{
+	uint32_t end_addr = (uint32_t)data_buffer + size;
+	struct ipc_dma_config *dma_cfg;
+	struct sof_tlv *tlvs;
+
+	for (tlvs = (struct sof_tlv *)data_buffer; (uint32_t)tlvs < end_addr;
+	     tlvs = tlv_next(tlvs)) {
+		dma_cfg = tlv_value_ptr_get(tlvs, GTW_DMA_CONFIG_ID);
+		if (!dma_cfg)
+			continue;
+
+		/* To be able to retrieve proper DMA config we need to check if
+		 * device_id value (which is alh_id) is equal to device_address.
+		 * They both contain SNDW master id and PDI. If they match then
+		 * proper config is found.
+		 */
+		for (uint32_t i = 0; i < dma_cfg->channel_map.device_count; i++) {
+			if (dma_cfg->channel_map.map[i].device_address == device_id) {
+				dai->host_dma_config[dma_cfg_idx] = dma_cfg;
+				return IPC4_SUCCESS;
+			}
+		}
+	}
+
+	return IPC4_INVALID_REQUEST;
 }
 
 void ipc4_base_module_cfg_to_stream_params(const struct ipc4_base_module_cfg *base_cfg,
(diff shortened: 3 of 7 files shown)


plbossart:
Erm, sorry, whaaaaat? 

LNL aggregation already works without any firmware support. The aggregation is handled with the host driver programming the PCMsCM registers in the multi-link configuration. The firmware should only deal with ONE stream ID and does not need to know whether any aggregation is handled since the hardware deals with aggregation.

Why is a firmware-based solution introduced now?

@iganakov @lgirdwood @mwasko ?

src/audio/copier/copier_dai.c
```diff
@@ -182,50 +256,19 @@ int copier_dai_create(struct comp_dev *dev, struct copier_data *cd,
 		break;
 	case ipc4_alh_link_output_class:
 	case ipc4_alh_link_input_class:
+#if defined(CONFIG_ACE_VERSION_2_0)
+		dai.type = SOF_DAI_INTEL_HDA;
```
plbossart, line 260: if I follow the same logic, then the dai.type would need to be INTEL_HDA for 

````c
case ipc4_i2s_link_output_class:
case ipc4_i2s_link_input_class:
case ipc4_dmic_link_input_class:
````

So this change in isolation is a change of convention. Either ALL LNL DAIS become INTEL_HDA, or none become INTEL_HDA.

**Review by plbossart — CHANGES_REQUESTED** [review]
on top of my earlier comment that aggregation does not need any firmware support, the change is inconsistent for the DMIC/SSP support that already works fine as well.

**Comment by iganakov** [comment]
@plbossart I've been anticipating such kind of questions that's why I submitted a draft :)
I didn't add any firmware-based aggregation I just used what was done for MTL. Firmware-based aggregation is needed for compatibility with Windows driver. 

My intention was to use current FW infrastructure to enable SNDW on LNL using HDA DMA and since HDA dai doesn't program any hardware I decided to use it for SNDW configuration on SOF side.

File: src/audio/copier/copier.c [hunk 2]
```diff
@@ -78,7 +80,33 @@ static int copier_init(struct processing_module *mod)
...
+	if (copier->gtw_cfg.config_length) {
+		gtw_cfg_size = (copier->gtw_cfg.config_length << 2);
+		gtw_cfg = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
+				  gtw_cfg_size);
+		if (!gtw_cfg) {
+			ret = -ENOMEM;
+			goto error_cd;
+		}
+
+		cd->gtw_cfg = gtw_cfg;
```
lyakh, line 100: I would assign only in success case - below `memcpy_s()` - to avoid a dangling pointer

src/audio/copier/copier.c
```diff
@@ -78,7 +80,33 @@ static int copier_init(struct processing_module *mod)
...
+		goto error_cd;
+	}
+
+	/* Allocate memory and store gateway_cfg in runtime. Gateway cfg has to
+	 * be kept even after copier is created e.g. during SET_PIPELINE_STATE
+	 * IPC when dai_config_dma_channel() is called second time and DMA
+	 * config is used to assing dma_channel_id value.
+	 */
+	if (copier->gtw_cfg.config_length) {
+		gtw_cfg_size = (copier->gtw_cfg.config_length << 2);
```
lyakh, line 92: superfluous parentheses

src/audio/copier/copier.c
```diff
@@ -78,7 +80,33 @@ static int copier_init(struct processing_module *mod)
...
+		if (!gtw_cfg) {
+			ret = -ENOMEM;
+			goto error_cd;
+		}
+
+		cd->gtw_cfg = gtw_cfg;
+
+		if (memcpy_s(cd->gtw_cfg, gtw_cfg_size,
+			     &copier->gtw_cfg.config_data, gtw_cfg_size) < 0) {
+			ret = -EINVAL;
```
lyakh, line 104: propagate error

src/audio/copier/copier.c
```diff
@@ -64,6 +64,8 @@ static int copier_init(struct processing_module *mod)
 	struct module_data *md = &mod->priv;
 	struct ipc4_copier_module_cfg *copier = (struct ipc4_copier_module_cfg *)md->cfg.init_data;
 	struct comp_ipc_config *config = &dev->ipc_config;
+	uint32_t *gtw_cfg;
```
lyakh, line 67: `void *`

src/include/ipc4/copier.h
```diff
@@ -242,6 +242,7 @@ struct copier_data {
 	 * from the IPC data.
 	 */
 	struct ipc4_copier_module_cfg config;
+	uint32_t *gtw_cfg;
```
lyakh, line 245: `void *`

src/audio/copier/copier_dai.c
```diff
@@ -64,6 +64,79 @@ static int copier_set_alh_multi_gtw_channel_map(struct comp_dev *dev,
 	return 0;
 }
 
+static int copier_alh_assign_dai_index(struct comp_dev *dev,
+				       uint32_t *gtw_cfg_data,
```
lyakh, line 68: `void *` and remove the cast in line 77

src/ipc/ipc4/dai.c
```diff
@@ -83,6 +86,27 @@ int dai_config_dma_channel(struct dai_data *dd, struct comp_dev *dev, const void
...
+		if (copier_cfg->gtw_cfg.node_id.f.dma_type == ipc4_alh_link_output_class ||
+		    copier_cfg->gtw_cfg.node_id.f.dma_type == ipc4_alh_link_input_class) {
+
+			if (!cd->gtw_cfg) {
+				comp_err(dev, "No gateway config found!");
+				return DMA_CHAN_INVALID;
+			}
+
+			const struct sof_alh_configuration_blob *alh_blob =
+				(const struct sof_alh_configuration_blob *)cd->gtw_cfg;
```
lyakh, line 99: this type-cast would go too

src/ipc/ipc4/helper.c
```diff
@@ -880,11 +881,43 @@ int ipc4_find_dma_config(struct ipc_config_dai *dai, uint8_t *data_buffer, uint3
...
+{
+#if defined(CONFIG_ACE_VERSION_2_0)
+	uint32_t end_addr = (uint32_t)data_buffer + size;
+	struct ipc_dma_config *dma_cfg;
+	struct sof_tlv *tlvs;
+	int dma_cfg_count = 0;
+
+	for (tlvs = (struct sof_tlv *)data_buffer; (uint32_t)tlvs < end_addr;
+	     tlvs = tlv_next(tlvs)) {
+		dma_cfg = (struct ipc_dma_config *)tlv_value_ptr_get(
```
lyakh, line 900: no need to type-cast a `void *` pointer to another pointer type

src/ipc/ipc4/helper.c
```diff
@@ -880,11 +881,43 @@ int ipc4_find_dma_config(struct ipc_config_dai *dai, uint8_t *data_buffer, uint3
...
+			   tlvs, GTW_DMA_CONFIG_ID);
+
+		if (!dma_cfg)
+			continue;
+
+		for (uint32_t i = 0; i < dma_cfg->channel_map.device_count; i++) {
+			if (dma_cfg->channel_map.map[i].device_address == device_id) {
+				dai->host_dma_config[dma_cfg_count] = dma_cfg;
```
lyakh, line 908: this is a somewhat strange counter: `dma_cfg_count` how many valid (with a non-NULL `dma_cfg` pointer) didn't match the requested device and then assigns to an array with that index... The connection isn't obvious. Could you clarify?

src/audio/copier/copier_dai.c
```diff
@@ -64,6 +64,78 @@ static int copier_set_alh_multi_gtw_channel_map(struct comp_dev *dev,
...
+	if (!is_multi_gateway(node_id)) {
+		ret = ipc4_find_dma_config_multiple(dai, dma_config, dma_config_length,
+						    alh_blob->alh_cfg.mapping[0].alh_id);
+		if (ret != 0) {
+			comp_err(mod->dev, "No sndw dma_config found in blob!");
+			return -EINVAL;
+		}
+		dai_index[0] = dai->host_dma_config[0]->stream_id;
+
+		return ret;
```
lyakh, line 99: here it's always `return 0`

src/ipc/ipc4/dai.c
```diff
@@ -83,6 +84,30 @@ int dai_config_dma_channel(struct dai_data *dd, struct comp_dev *dev, const void
...
+		if (copier_cfg->gtw_cfg.node_id.f.dma_type == ipc4_alh_link_output_class ||
+		    copier_cfg->gtw_cfg.node_id.f.dma_type == ipc4_alh_link_input_class) {
+			struct processing_module *mod = comp_get_drvdata(dev);
+			struct copier_data *cd = module_get_private_data(mod);
+
+			channel = DMA_CHAN_INVALID;
+
+			if (!cd->gtw_cfg) {
+				comp_err(dev, "No gateway config found!");
+				return channel;
```
lyakh, line 97: I think it would be more obvious to have `return DMA_CHAN_INVALID` here and move line 93 to around line 100 below

RanderWang (commented):
Looks good to me. BTW we support llp update with GP-DMA, how about HD-DMA ?

lgirdwood:
@plbossart @iganakov do you guys have alignment now or still opens around the latest patch updates ?

plbossart:
> @plbossart @iganakov do you guys have alignment now or still opens around the latest patch updates ?

No alignment, there's no clarity on what exactly the gateway configuration should be. Both @bardliao and I looked into this - no reply on our comments so far.

src/ipc/ipc4/helper.c
plbossart: that is absolutely not what we discussed with @mmaka1. device_address was supposed to be the PCMsCMy FIFO address.

src/audio/copier/copier.c
```diff
@@ -78,7 +80,31 @@ static int copier_init(struct processing_module *mod)
...
+
+	/* Allocate memory and store gateway_cfg in runtime. Gateway cfg has to
+	 * be kept even after copier is created e.g. during SET_PIPELINE_STATE
+	 * IPC when dai_config_dma_channel() is called second time and DMA
+	 * config is used to assign dma_channel_id value.
+	 */
+	if (copier->gtw_cfg.config_length) {
+		gtw_cfg_size = copier->gtw_cfg.config_length << 2;
+		gtw_cfg = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
+				  gtw_cfg_size);
```
lyakh, line 94: no need for `rzalloc()` - it's completely overwritten 6 lines below. Just use `rmalloc()`

src/audio/copier/copier_dai.c
```diff
@@ -182,50 +255,19 @@ int copier_dai_create(struct comp_dev *dev, struct copier_data *cd,
...
-		 *     {
-		 *	   count;
-		 *	  {
-		 *	     node_id;  \\ normal gtw id
-		 *	     mask;
-		 *	  }  mapping[MAX_ALH_COUNT];
-		 *     }
-		 *   }
-		 * }
-		 */
```
lyakh, line 205: is this... description not valid any more? I'm not entirely sure what it meant, but some documentation is often better than no documentation

iganakov:
@lgirdwood @plbossart sorry guys I was on a sick leave. I'm in touch with @bardliao we're making some additional experiments

src/audio/copier/copier_dai.c
```diff
@@ -64,6 +64,78 @@ static int copier_set_alh_multi_gtw_channel_map(struct comp_dev *dev,
...
+	struct copier_data *cd = module_get_private_data(mod);
+	const struct sof_alh_configuration_blob *alh_blob = gtw_cfg_data;
+	int i, dai_num;
+
+	if (!cd->config.gtw_cfg.config_length) {
+		comp_err(mod->dev, "No gateway config found in blob!");
+		return -EINVAL;
+	}
+
+#if defined(CONFIG_ACE_VERSION_2_0)
```
kv2019i, line 84: Here again, we have very Intel specific code in copier_dai.c which would need to be generic. We've had DAI type specific code segments before in src/audio , but this is expanding the amount of code quite significantly.

kv2019i (commented):
Hmm. A generic concern is that this is exploding amount of seemingly Intel specific code in generic sof/src/audio code. This is hinting the split of what is done in Zephyr versus SOF is not right with the DAI interfaces, if we need to do this much platform specific ifdefs in generic SOF code. Many bits have existed before this PR, but this PR does double/triple the size.

lgirdwood:
@iganakov thinking aloud, do we need a new boot ID flag to indicate the presence of this feature i.e. supported modes ?

iganakov:
@lgirdwood We have internal discussions about common blob format for Linux and Windows. I'll update this PR when proper solution will appear.

plbossart:
> @lgirdwood We have internal discussions about common blob format for Linux and Windows. I'll update this PR when proper solution will appear.

@iganakov Is there an ETA for this solution? The initial patches were 1.5 month ago, we are December 5 so this is gradually pushing into 2024...

iganakov:
@plbossart The ETA is the end of this week.

src/ipc/ipc4/helper.c
```diff
@@ -880,11 +881,42 @@ int ipc4_find_dma_config(struct ipc_config_dai *dai, uint8_t *data_buffer, uint3
...
-	dai->host_dma_config = GET_IPC_DMA_CONFIG(data_buffer, size);
+	dai->host_dma_config[0] = GET_IPC_DMA_CONFIG(data_buffer, size);
 #endif
 	return IPC4_SUCCESS;
 }
 
+int ipc4_find_dma_config_multiple(struct ipc_config_dai *dai, uint8_t *data_buffer,
+				  uint32_t size, uint32_t device_id)
+{
+#if defined(CONFIG_ACE_VERSION_2_0)
```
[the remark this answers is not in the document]
iganakov: @kv2019i I found a way to make code more generic, please check a new version.

src/audio/copier/copier.c
iganakov: Fixed in PR https://github.com/thesofproject/sof/pull/8574

src/audio/copier/copier_dai.c
```diff
@@ -64,6 +64,92 @@ static int copier_set_alh_multi_gtw_channel_map(struct comp_dev *dev,
...
+	int i, dai_num, ret;
+
+	if (!cd->config.gtw_cfg.config_length) {
+		comp_err(mod->dev, "No gateway config found in blob!");
+		return -EINVAL;
+	}
+
+	switch (dai->type) {
+	case SOF_DAI_INTEL_HDA:
+		/* We use DAI_INTEL_HDA for ACE 2.0 platforms */
```
plbossart, line 88: This comment doesn't exactly match with the code, or the connection is not obvious.

What in this piece of code is dependent on ACE2.0?
I don't see any test on the hardware generation....

src/audio/copier/copier_dai.c
```diff
@@ -64,6 +64,92 @@ static int copier_set_alh_multi_gtw_channel_map(struct comp_dev *dev,
...
+
+		for (i = 0; i < dai_num; i++) {
+			ret = ipc4_find_dma_config_multiple(dai, dma_config,
+							    dma_config_length,
+							    alh_blob->alh_cfg.mapping[i].alh_id);
+			if (ret != 0) {
+				comp_err(mod->dev, "No sndw dma_config found in blob!");
+				return -EINVAL;
+			}
+			dai_index[i] = dai->host_dma_config[i]->stream_id;
```
plbossart, line 122: what does the host_dma config have to do with the dai index?

There can be cases where we have mixers and the host_dma has multiple streams but there's only one stream_id for the link DMA. The reverse is also true on capture.

src/ipc/ipc4/dai.c
```diff
@@ -84,6 +85,29 @@ int dai_config_dma_channel(struct dai_data *dd, struct comp_dev *dev, const void
...
+				comp_err(dev, "No gateway config found!");
+				return DMA_CHAN_INVALID;
+			}
+
+			channel = DMA_CHAN_INVALID;
+			const struct sof_alh_configuration_blob *alh_blob = cd->gtw_cfg;
+
+			for (int i = 0; i < alh_blob->alh_cfg.count; i++) {
+				if (dai->host_dma_config[i]->stream_id == dai->dai_index) {
+					channel = dai->host_dma_config[i]->dma_channel_id;
```
plbossart, line 104: same here, what does 'host_dma_config' mean in this context?

If it's meant to refer to the link DMA configuration setup by the host, it's really poor naming. All the cAVS/ACE documentation refers to host_dma by opposition to the link_dma. I am afraid the naming blurs the concepts at best, and is possibly wrong.

src/audio/copier/copier_dai.c
plbossart: There was no answer to this comment from 2 months ago....

src/audio/copier/copier_dai.c
iganakov: We select `DAI_INTEL_HDA` for LNL and beyond.  For MTL and older platforms we use `DAI_INTEL_ALH`  because they have ALH HW. In line 272 there is a check for `CONFIG_ACE_VERSION_2_0`

src/audio/copier/copier_dai.c
iganakov: `host_dma_config` holds pointer to DMA config (`struct ipc_dma_config`) which we receive in Copier INIT_INSTANCE IPC and it contains only link DMA related data. It  was added previously when I2S, DMIC and ALH dai were changed to use HDA DMA for LNL. I agree the name host_dma_config could be misleading

src/ipc/ipc4/dai.c
iganakov: `host_dma_config ` contains pointer to  `struct ipc_dma_config` and it's meant to refer to link DMA config. Please check my explanation above.

src/ipc/ipc4/helper.c
iganakov: Yes, device_address should be PCMsCMy, but Windows drv always set this value and I need to deal with it.

src/audio/copier/copier_dai.c
plbossart: The main open is that your PR is called "lnl: sndw: use HDA DMA to transfer stream data"

But that's not really an option, this is what the hardware requires, and not only for Soundwire but DMIC/SSP as well.

So the point is that it's still unclear why we are special-casing SoundWire when from a DMA perspective the same support is required as SSP.

It's even more confusing since we are now adding aggregation concepts in the HDaudio case, when aggregation is a concept that only exists for SoundWire.

lgirdwood:
Adding @bardliao as IIUC he has to comment on using dynamic array for blob data impact for MTL.

lgirdwood:
> Adding @bardliao as IIUC he has to comment on using dynamic array for SDW blob data impact for MTL, unless this is a separate issue with no impact for this PR ? @plbossart @iganakov

bardliao:
> Adding @bardliao as IIUC he has to comment on using dynamic array for blob data impact for MTL unless no impact for SDW blob data usage for this PR ? @plbossart @iganakov

My test PR is https://github.com/thesofproject/linux/pull/4752. I tested it on my Dell MTL laptop, and no regression found. PR test result also looks good.

lgirdwood:
> > Adding @bardliao as IIUC he has to comment on using dynamic array for blob data impact for MTL unless no impact for SDW blob data usage for this PR ? @plbossart @iganakov
> 
> My test PR is [thesofproject/linux#4752](https://github.com/thesofproject/linux/pull/4752). I tested it on my Dell MTL laptop, and no regression found. PR test result also looks good.

Great - @plbossart what are the next steps now ?

plbossart:
Just to be clear, what tests did you run @bardliao? 

we have 3 possible configurations:

a) existing firmware with https://github.com/thesofproject/linux/pull/4752
b) firmware with this PR with existing kernel
c) firmware with this PR and with https://github.com/thesofproject/linux/pull/4752

I am mostly concerned about a) since that's a clear example of "we don't break user setups".

src/ipc/ipc4/helper.c
plbossart: I am still lost by all this.

what exactly is this 'device_id'

a) the link_id as I can read in the November 3 comment
b) the address of the PCMsCMy register
c) the PDI index written in PCMsCMy?

Also we all agree that the firmware does not program the PCMsCMy register, so what exactly does the firmware do with the device_id.

How many engineers does it take to define what a 'device' is?


Two additional questions: what happens if we do

One additional question: the 'device_id' aka PDI ID is specific to a link. The DMA configuration is not link specific, we can use whatever stream_tag we want.

So how do we know which PDI

plbossart (commented):
I don't really have any comments on the code proper, but there's no description of what it does and what the TLV needs to be.

bardliao:
> Just to be clear, what tests did you run @bardliao?
> 
> we have 3 possible configurations:
> 
> a) existing firmware with [thesofproject/linux#4752](https://github.com/thesofproject/linux/pull/4752) b) firmware with this PR with existing kernel c) firmware with this PR and with [thesofproject/linux#4752](https://github.com/thesofproject/linux/pull/4752)
> 
> I am mostly concerned about a) since that's a clear example of "we don't break user setups".

I tested a) The existing firmware (v2.7) with [thesofproject/linux#4752](https://github.com/thesofproject/linux/pull/4752)

src/ipc/ipc4/helper.c
```diff
@@ -1002,11 +1003,44 @@ int ipc4_find_dma_config(struct ipc_config_dai *dai, uint8_t *data_buffer, uint3
...
+		}
+
+		for (uint32_t i = 0; i < dma_cfg->channel_map.device_count; i++) {
+			if (dma_cfg->channel_map.map[i].device_address == device_id) {
+				dai->host_dma_config[dma_cfg_count] = dma_cfg;
+				return IPC4_SUCCESS;
+			}
+		}
+
+		dma_cfg_count++;
```
bardliao, line 1038: @iganakov We should only do `dma_cfg_count++` when we assign `dai->host_dma_config[dma_cfg_count] = dma_cfg;`.
IMHO, dma_cfg_count should be an argument of `ipc4_find_dma_config_multiple` not a local variable. `ipc4_find_dma_config_multiple` will return immediately when it find the dma config with the device_id. We need to specify dai->host_dma_config[] index from the caller.
I do meet an issue if the dma config is not found in the first tlv. In that case,  we don't assign `dai->host_dma_config[dma_cfg_count] = dma_cfg;` in the first `for (uint32_t i = 0; i < dma_cfg->channel_map.device_count; i++)` loop, and `dma_cfg_count++`. As a result, `dai->host_dma_config[0]` = NULL and dma_cfg is assigned to `dai->host_dma_config[1]` that is not what we want.

lgirdwood:
@iganakov @bardliao any update ? Do we have any working updates now ?

src/ipc/ipc4/helper.c
iganakov: Good catch. I'll change `ipc4_find_dma_config_multiple()` in next revision of this patch. Thanks @bardliao

iganakov:
@lgirdwood I prepared some minor improvements and will update this PR soon. As I know @bardliao modified Linux blob and is able to use FW aggregation for playback but still has some issues with capture.

lgirdwood:
@bardliao any update for capture ?
@plbossart blob format good for you ?

src/ipc/ipc4/helper.c
```diff
@@ -1031,11 +1032,35 @@ int ipc4_find_dma_config(struct ipc_config_dai *dai, uint8_t *data_buffer, uint3
...
+	struct sof_tlv *tlvs;
+
+	for (tlvs = (struct sof_tlv *)data_buffer; (uint32_t)tlvs < end_addr;
+	     tlvs = tlv_next(tlvs)) {
+		dma_cfg = tlv_value_ptr_get(tlvs, GTW_DMA_CONFIG_ID);
+		if (!dma_cfg)
+			continue;
+
+		for (uint32_t i = 0; i < dma_cfg->channel_map.device_count; i++) {
+			if (dma_cfg->channel_map.map[i].device_address == device_id) {
```
plbossart, line 1054: can we add a comment on what a 'device' is, what a 'device_address' corresponds to?

Every time I look at this PR I forgot what the verbal answer was and have to re-ask

src/ipc/ipc4/dai.c
```diff
@@ -79,8 +79,8 @@ int dai_config_dma_channel(struct dai_data *dd, struct comp_dev *dev, const void
 	case SOF_DAI_INTEL_DMIC:
 		channel = 0;
 #if defined(CONFIG_ACE_VERSION_2_0)
-		if (dai->host_dma_config->pre_allocated_by_host)
-			channel = dai->host_dma_config->dma_channel_id;
+		if (dai->host_dma_config[0]->pre_allocated_by_host)
+			channel = dai->host_dma_config[0]->dma_channel_id;
```
plbossart, line 84: I would break this patch in two, first the addition of this array for the host_dma_config....

src/ipc/ipc4/helper.c
```diff
@@ -1031,11 +1032,35 @@ int ipc4_find_dma_config(struct ipc_config_dai *dai, uint8_t *data_buffer, uint3
...
+				dai->host_dma_config[dma_cfg_idx] = dma_cfg;
+				return IPC4_SUCCESS;
+			}
+		}
+	}
+
+	return IPC4_INVALID_REQUEST;
+}
+
```
plbossart, line 1063: ... and add this function in a separate patch, or when it's actually used.

The current code partitioning makes it hard to figure out what this brings or why this helper was needed.

src/audio/copier/copier_dai.c
```diff
@@ -64,6 +64,92 @@ static int copier_set_alh_multi_gtw_channel_map(struct comp_dev *dev,
...
+
+		for (i = 0; i < dai_num; i++) {
+			ret = ipc4_find_dma_config_multiple(dai, dma_config,
+							    dma_config_length,
+							    alh_blob->alh_cfg.mapping[i].alh_id, i);
+			if (ret != 0) {
+				comp_err(mod->dev, "No sndw dma_config found in blob!");
+				return -EINVAL;
+			}
+			dai_index[i] = dai->host_dma_config[i]->stream_id;
```
plbossart, line 122: Please explain the connection between dai_index and stream_id.

Stream_id looks related to the stream_tag used for HDaudio, but it's absolutely not the DAI for SoundWire - which is rather defined in terms of links/PDI.

Or at the very least a comment to say that you intentionally overloaded the concept of DAI but this dai_index is only used for the DMA handling.

plbossart (changes requested):
I still have objections on the clarity of the code
- What is a device?
- Can we clarify the design and why a DMA 'stream_id' becomes a 'dai_index'

when it looks confusing it's awful to maintain...

bardliao:
> @bardliao any update for capture ? 

I am still struggling with aggregation capture. I always get pcm IO error even with exact the same blob that was verified with slim driver.

tmleman (changes requested):

src/ipc/ipc4/helper.c
```diff
@@ -1031,11 +1032,35 @@ int ipc4_find_dma_config(struct ipc_config_dai *dai, uint8_t *data_buffer, uint3
...
+		dma_cfg = tlv_value_ptr_get(tlvs, GTW_DMA_CONFIG_ID);
+		if (!dma_cfg)
+			continue;
+
+		for (uint32_t i = 0; i < dma_cfg->channel_map.device_count; i++) {
+			if (dma_cfg->channel_map.map[i].device_address == device_id) {
+				dai->host_dma_config[dma_cfg_idx] = dma_cfg;
+				return IPC4_SUCCESS;
+			}
+		}
```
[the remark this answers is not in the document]
iganakov: I can't use `tlv_value_get()` because it is able to find only the first tlv with the same type, but in case of aggregation there  will be two or more. Therefore, I will need to call `tlv_value_get()` in a loop which doesn't make any sense.

src/audio/copier/copier_dai.c
iganakov: @plbossart as you mentioned above `stream_tag` used for HDaudio, and because we use HD-A DMA to process data for SoundWire it looked logical for me to use `stream_tag` as a `dai_index` instead of link/PDI.

plbossart (approved):
@bardliao must approve this one.
I would still add a comment, see below.

tmleman (approved):

lgirdwood (approved):

bardliao (approved):

dnikodem (approved):

plbossart (approved):
Thanks @iganakov

marc-hb:
Causing regression:
- #8816

PR #8365, opened 2023-10-20
sof_api: lmdk: Extraction of api headers for loadable modules

softwarecki:
I separated some of the headers into a separate modules directory. They will be shared between sof and native loadable modules.


src/include/sof_api/ipc4/base-config.h
```diff
@@ -0,0 +1,240 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
```
lgirdwood, line 1: So I agree having an API directory for modules is good, but can we call it include/module/ ? This way all our modules would be including e.g.
```c
#include <module/ipc4/base-config.h>
#include <module/module.h>
```
and so on. i.e. we make it clear this API is for modules and make it more obvious when new APIs are added if they are intended for baseFW or are intened for client modules.

src/include/ipc4/base-config.h
```diff
@@ -1,6 +1,6 @@
-/* SPDX-License-Identifier: BSD-3-Clause
- *
- * Copyright(c) 2021 Intel Corporation. All rights reserved.
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
```
lgirdwood, line 2: We dont need to change the date here if we are moving headers from A -> B

lgirdwood (commented):
Fwiw, I think git has a cmd line hint to help it detect file moves, but GH may not interpret it.

src/include/sof_api/ipc4/base-config.h
pjdobrowolski: Idea behind it is to keep this files as an external header package and calling it sof_api doesn't leave place for mistake.

src/include/ipc4/base-config.h
softwarecki: I always try to update the date when I edit the file. In this case, I cut most of the content, but also added a one new line.

dbaluta:
@lgirdwood @softwarecki 

Now that Zephyr supports loadable module is there a plan to use this feature with SOF?

https://docs.zephyrproject.org/3.5.0/services/llext/index.html

src/include/sof_api/ipc4/base-config.h
softwarecki: Modules include are limited only to include/sof_api directory, so its include will looks like:
`#include <ipc4/base-config.h>
#include <module/generic.h>`

From the sof perspective it is:
`#include <sof_api/ipc4/base-config.h>
#include <sof_api/module/generic.h>`

lgirdwood: > Modules include are limited only to include/sof_api directory, so its include will looks like:

I'm good with that. My issue is that ```sof_api`` name is quite ambiguous i.e. API for SOF what ? all of SOF or just part of SOF. In this case it needs to be clear that this API is only for modules only.

lgirdwood:
> @lgirdwood @softwarecki
> 
> Now that Zephyr supports loadable module is there a plan to use this feature with SOF?
> 
> https://docs.zephyrproject.org/3.5.0/services/llext/index.html

Yep, we are working on several fronts to provide 
1) backwards compatible modules/libraries API used by Windows (this PR and others from @softwarecki  and @pjdobrowolski) and to be enabled soon for MTL for SOF.
2) proof of concept with the new Zephyr module APIs to see how this can help and work on SOF DSP architectures (@lyakh and @teburd) to be reviewed post MTL.

We should be able to have both methods usuable above via Kconfig. Your welcome to beta test :)

src/include/sof_api/ipc4/base-config.h
pjdobrowolski: These headers will be kept in this repo for CI needs. If any changes will be required we will create headers release package for loadable modules and change api version number to keep on track changes. For that will have to create some kind of ABI script only for this cathalog.

pjdobrowolski: > > Modules include are limited only to include/sof_api directory, so its include will looks like:
> 
> I'm good with that. My issue is that `sof_api` name is quite ambiguous i.e. API for SOF what ? all of SOF or just part of SOF. In this case it needs to be clear that this API is only for modules only.

sof_api because we are using module_api, component_api and that headers contains essential data structures for modules to be loaded. If naming is such a problem we can move them to lmdk/include

dbaluta:
> We should be able to have both methods usuable above via Kconfig. Your welcome to beta test :)

Great. Thanks. Will test the new features with Cadence libraries.

src/include/sof/audio/module_adapter/module/generic.h
```diff
@@ -1,11 +1,12 @@
-/* SPDX-License-Identifier: BSD-3-Clause
```
marcinszkudlinski, line 1: Lets find a better name for "generic" - it says nothing at all
module_base.h ?

src/include/sof/audio/sink_api_implementation.h
```diff
@@ -11,72 +11,7 @@
 #include <stdint.h>
 #include <stdbool.h>
 
-/* forward def */
-struct sof_audio_stream_params;
```
marcinszkudlinski, line 15: As we agreed some times ago, we'll give up with total separation of sink/src internals and externals (like private/public in c++). 
All inlines should go to sinlk_api.h / source_api.h
__api_implementation.h  is no longer needed

src/include/sof_api/audio/sink_api.h
```diff
@@ -0,0 +1,178 @@
...
+ *	  if commit_size is bigger than the amount of data obtained before by get_buffer(), only
+ *	  the amount obtained before will be committed. That means - if somebody obtained a buffer,
+ *	  filled it with data and wants to commit it in whole, it may simple call
+ *	  commit_buffer with commit_size==MAXINT
+ * @return proper error code (0  on success)
+ */
+static inline int sink_commit_buffer(struct sof_sink *sink, size_t commit_size)
+{
+	int ret;
+
```
marcinszkudlinski, line 134: Isn't it too much to inline?

src/audio/module_adapter/iadk/system_agent.cpp
```diff
@@ -14,7 +14,7 @@
 #include <stdint.h>
 #include <rtos/string.h>
 #include <utilities/array.h>
-#include <adsp_error_code.h>
```
marcinszkudlinski, line 17: so now we have sof_api and iadk ?

Should not IADK go to sof_api as well?

marcinszkudlinski (changes requested):

src/include/sof_api/ipc4/base-config.h
lgirdwood: The direction of travel is component api is deprecated and will be removed, we will only have module api for modules so this programming surface must be obvious for module developers. In the future we will have the following include directories (not showing IADK atm)
1) src/include/module - all APIs used for modules
2) src/include/sof - base FW apis, not used by modules.
3) src/audio/<module_name>/ - local module private headers, used by module only.

pjdobrowolski (approved):

lmdk/cmake/ldscripts/text_linker_script.txt
```diff
@@ -11,7 +11,7 @@ PHDRS {
 SECTIONS {
     .text : ALIGN(4096) {
         _text_start = ABSOLUTE(.);
-        *(.buildinfo)
+        KEEP(*(.buildinfo))
```
pjdobrowolski, line 14: Does it really part of headers?

src/include/sof_api/ipc4/base-config.h
softwarecki: @lgirdwood I will rename `sof_api` to `module` soon.

lmdk/cmake/ldscripts/text_linker_script.txt
softwarecki: You are right! I will move it to #8399.

src/audio/module_adapter/iadk/system_agent.cpp
cujomalainey: what is the difference?

src/include/module/audio/audio_stream.h
```diff
@@ -0,0 +1,53 @@
...
+#include "../ipc/stream.h"
+
+/**
+ * set of parameters describing audio stream
+ * this structure is shared between audio_stream.h and sink/source interface
+ * TODO: compressed formats
+ */
+struct sof_audio_stream_params {
+	enum sof_ipc_frame frame_fmt;	/**< Sample data format */
+	enum sof_ipc_frame valid_sample_fmt;
```
cujomalainey, line 23: can this get a comment as well? or is this a copy

src/include/module/audio/format.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include <stdint.h>
+#include "../ipc/stream.h"
+
+static inline uint32_t get_sample_bytes(enum sof_ipc_frame fmt)
+{
+	switch (fmt) {
+	case SOF_IPC_FRAME_S16_LE:
+		return 2;
+	case SOF_IPC_FRAME_S24_3LE:
```
cujomalainey, line 22: SOF_IPC_FRAME_S24_4LE?

src/include/module/ipc4/base-config.h
```diff
@@ -0,0 +1,240 @@
...
+
+struct ipc4_pin_list_info {
+	uint32_t pin_count;
+	struct ipc4_pin_props pin_info[1];
+} __attribute__((packed, aligned(4)));
+
+/* structure describing module instance properties used in response
+ * to module LargeConfigGet with MOD_INST_PROPS parameter.
+ */
+struct ipc4_module_instance_props {
```
cujomalainey, line 204: per field descriptions?

src/include/module/audio/audio_stream.h
softwarecki: This structure was moved as is from [src/include/sof/audio/audio_stream.h](https://github.com/thesofproject/sof/pull/8365/files#diff-0dba4adf6342b75cd53dcd502dc280cf1d47d47fc12685180f6e1b873a7d0825L44)

src/include/module/audio/format.h
softwarecki: This function was moved as is from [src/include/sof/audio/format.h](https://github.com/thesofproject/sof/pull/8365/files#diff-6d0f97b1dc37dbd4b9302e4c4456638dd19426a8126cc311837e0225b4e7827eL173)

src/include/module/ipc4/base-config.h
softwarecki: This structure was moved as is from [src/include/ipc4/base-config.h](https://github.com/thesofproject/sof/pull/8365/files#diff-171a61681a099f4c7c767584771f5dd146174c4b93ea31f3501c37af338b17a5L205)

src/include/sof/audio/coefficients/up_down_mixer/up_down_mixer.h
```diff
@@ -6,8 +6,6 @@
 // Author: Adrian Bonislawski <adrian.bonislawski@intel.com>
 
 #include <ipc4/up_down_mixer.h>
-#include <sof/audio/format.h>
-#include <stdint.h>
```
lyakh, line 10: stdint.h is needed for the `int32_t` type, used in this header

src/include/sof/audio/component.h
```diff
@@ -18,30 +18,13 @@
...
 #include <ipc/control.h>
-#include <ipc/stream.h>
 #include <sof/ipc/topology.h>
 #include <kernel/abi.h>
-#include <user/trace.h>
 
-#include <errno.h>
-#include <stdbool.h>
-#include <stddef.h>
-#include <stdint.h>
```
lyakh, line 44: I'm very suspicious about all these removals, how did you figure out, that all these headers aren't needed?

src/include/sof/audio/component.h
```diff
@@ -18,30 +18,13 @@
 
 #include <sof/audio/buffer.h>
 #include <sof/audio/format.h>
-#include <sof/audio/pipeline.h>
```
lyakh, line 21: e.g. this one is needed for a `pipeline_xrun()` prototype

src/include/module/module/interface.h
```diff
@@ -0,0 +1,220 @@
...
+};
+
+/**
+ * \struct input_stream_buffer
+ * \brief Input stream buffer
+ */
+struct input_stream_buffer {
+	void *data; /* data stream buffer */
+	uint32_t size; /* size of data in the buffer */
+	uint32_t consumed; /* number of bytes consumed by the module */
```
lyakh, line 47: stdint.h is needed

src/include/module/module/interface.h
```diff
@@ -0,0 +1,220 @@
...
+
+	/* Indicates end of stream condition has occurred on the input stream */
+	bool end_of_stream;
```
lyakh, line 50: stdbool.h

src/include/module/module/generic.h
```diff
@@ -0,0 +1,138 @@
...
+	void *module_adapter; /**<loadable module interface handle */
+	uint32_t module_entry_point; /**<loadable module entry point address */
+#endif /* SOF_MODULE_PRIVATE */
+};
+
+/* module_adapter private, runtime data - public version*/
+struct processing_module {
+	struct module_data priv; /**< module private data */
+	uint32_t period_bytes; /** pipeline period bytes */
+#ifdef SOF_MODULE_API_PRIVATE
```
lyakh, line 56: is this macro only going to be defined in the base firmware and not in modules? I had problems with conflicting structure definitions when failing to keep them in sync, so this can be dangerous

lyakh (changes requested):

src/include/sof/audio/component.h
pjdobrowolski: I removed headers one by one and check build on sof main. I don't know why there a so many headers not being used or included somewhere else and being redundantly called.

lyakh: This isn't the criterion how we decide whether a header is needed in a file or not. For every feature used in a file, if that feature requires a header, we include that header explicitly in that file instead of relying on indirect inclusion. I can think of two exceptions to this rule:
1. For structure pointers a forward declaration of that structure is enough
2. Some headers explicitly must only be included in a single header that wraps them. This is typical for platform files e.g. src/platform/meteorlake/include/platform/lib/cpu.h

src/include/sof_api/audio/sink_api.h
marc-hb: > Isn't it too much to inline?

`inline` is required in .h files to avoid the One Definition Rule:

https://stackoverflow.com/questions/39652884/one-definition-rule-multiple-definition-of-inline-functions
https://en.wikipedia.org/wiki/Inline_function

Every time I checked, the _optimizer_ did not care at all about `inline`.

src/include/sof/audio/component.h
marc-hb: > I removed headers one by one and check build on sof main.

This proves nothing because they are frequently included indirectly still. For real solutions check

- https://github.com/zephyrproject-rtos/zephyr/issues/41543

src/include/module/ipc4/base-config.h
```diff
@@ -0,0 +1,240 @@
...
+	uint32_t  dp_queue_type;
+	uint32_t  queue_alignment;
+	uint32_t  cp_usage_mask;
+	uint32_t  stack_bytes;
+	uint32_t  bss_total_bytes;
+	uint32_t  bss_used_bytes;
+	uint32_t  ibs_bytes;
+	uint32_t  obs_bytes;
+	uint32_t  cpc;
+	uint32_t  cpc_peak;
```
btian1, line 215: @marcinszkudlinski @pblaszko , @lgirdwood , on our previous discussion, we all agreed:
CPC = max(peak), however, here have two variables, named with cpc and cpc_peak, what's the exact meaning and how to get two different CPC?

src/include/sof_api/audio/sink_api.h
lyakh: @marc-hb I think what @marcinszkudlinski meant is that if the function is too large to be inlined, it could be placed as a normal exported function in a .c file instead

softwarecki: This function was moved to src/module/audio/sink_api.c

src/include/sof/audio/component.h
lgirdwood: The full CI build of FW, tooling,docs etc should pick up any issues here, so I would say all green CI would mean header cleanups are good fo current codebase. However, we could get lucky in some places and have to manually update some headers with dependencies as we find errors.

src/include/module/ipc4/base-config.h
lgirdwood: Only CPC peak is needed according to @pblaszko so we can remove cpc IIUC.

src/include/module/module/interface.h
```diff
@@ -0,0 +1,222 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
```
lgirdwood, line 1: @softwarecki I'm guessing git not detect the move here and this is not a new file ?

src/include/module/module/interface.h
```diff
@@ -0,0 +1,222 @@
...
+	/**
+	 * Module specific bind procedure, called when modules are bound with each other
+	 */
+	int (*bind)(struct processing_module *mod, void *data);
+	/**
+	 * Module specific unbind procedure, called when modules are disconnected from one another
+	 */
+	int (*unbind)(struct processing_module *mod, void *data);
+
+#ifdef SOF_MODULE_API_PRIVATE
```
lgirdwood, line 217: Not a blocker, but we may want to add some commentry here to explain what private means.

src/include/module/module/generic.h
```diff
@@ -0,0 +1,138 @@
...
+
+#ifndef __MODULE_MODULE_GENERIC__
+#define __MODULE_MODULE_GENERIC__
+
+#include <stdint.h>
+#include <stddef.h>
+#include <stdbool.h>
+
+#include "interface.h"
+#include "../ipc4/base-config.h"
```
lgirdwood, line 17: ```<module/ipc4/base-config.h>``` ?

src/include/sof/audio/module_adapter/module/generic.h
lgirdwood: yep, agree. base.h or core.h are more meaningfull

src/include/kernel/abi.h
```diff
@@ -30,7 +30,7 @@
 /** \brief SOF ABI version major, minor and patch numbers */
 #define SOF_ABI_MAJOR 3
 #define SOF_ABI_MINOR 28
-#define SOF_ABI_PATCH 0
+#define SOF_ABI_PATCH 1
```
lgirdwood, line 33: Whay are we touching IPC3, we've not changed any IPC3 here ?

src/CMakeLists.txt
```diff
@@ -34,4 +36,4 @@ endif()
 
 if (CONFIG_LOADABLE_MODULES)
 	add_subdirectory(library_manager)
-endif()
+endif()
```
lgirdwood, line 39: needs a newline at EOF.

lgirdwood (commented):
Just a few opens from me. Almost ready.

src/include/sof/audio/component.h
lyakh: @lgirdwood sorry, I humbly disagree. We could have a single header, that include all other imaginable headers, and have all .c files just include that single header. That would probably pass all build- and run-time tests (with some tweaking probably). I don't think that's what we want.

pjdobrowolski: We had this discussion before with @mmaka1 and @marcinszkudlinski  and creating monster header with all headers is not good idea.

lgirdwood: @lyakh I agree, I'm not sure we can remove all the POSIX headers, i.e. I'm pretty sure we use stdint.h for all structure types, but the other SOF API headers, if they are NOT used in that header file then we can remove (and yes it wont be 100% test build-able as we cant yet independently build each header on its own).  Forward structure/type declarations will no doubt be needed too.

What I mean here is we have a CI pass meaning we are build testing the best we can today and this is not a proof that all headers are build-able after this change (if that is even true before this PR today), but an indication that we have some confidence. 

I think we all agree that we are not 
1) Doing monster header files.
2) Duplicating header files in multiple locations.
3) Creating derived header files with copy and paste structure and API definitions (but we can move structures to align with "needed by modules" rule).

src/include/module/ipc4/base-config.h
softwarecki: In this PR, I only moved this structure to another file. Renaming/deleting a field is beyond the scope of this PR.

src/include/module/module/generic.h
softwarecki: This file is used by both sof and a loadable module. From the sof perspective it is accessible by `<module/ipc4/base-config.h>`, and from the loadable module perspective it is `<ipc4/base-config.h>`. That's why I used a relative path to handle both cases.

src/include/kernel/abi.h
softwarecki: To make CI / checkpatch happy

plbossart: Not following either, these tools shouldn't care.

src/include/module/module/generic.h
lgirdwood: We should set this correctly in the build for baseFW and for modules. i.e. we use ```<module/ipc4/base-config.h>``` and use the ```-I``` directive depending on build target

marc-hb (dismissed):
> For brand new directories use add_subdirectory(../src/module module_install_ignored) and a new src/module/CMakeLists.txt file, see example in commit https://github.com/thesofproject/sof/commit/ef563ebfe3d5725e197d7de7f46b7e7a7672cf78

Another, more recent example in PR #8494

src/include/module/module/generic.h
softwarecki: @lgirdwood If I specify the path to `src/include` instead of `src/include/module` via `-I` when building a loadable module, the modules will be able to include any sof header file, which is what we want to avoid. Should I move the `module` directory to another place that will be added to the include directories by both sof and modules?

src/include/module/module/interface.h
softwarecki: I moved some parts of `src/include/sof/audio/module_adapter/module/module_interface.h` to this file, so it's a copy rather than a move.

> Git does not do rename tracking nor copy tracking, which means **it doesn't record renames or copies**. What it does instead is **rename and copy detection**.
https://stackoverflow.com/a/1043566

src/include/sof/audio/module_adapter/module/generic.h
softwarecki: renamed to base.h

marcinszkudlinski (approved):

src/include/sof/audio/format.h
```diff
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  *
- * Copyright(c) 2016 Intel Corporation. All rights reserved.
+ * Copyright(c) 2023 Intel Corporation. All rights reserved.
```
kv2019i, line 3: Ouch, 2016 lost.

softwarecki:
@kv2019i I restored copyright years.

From my perspective, it was just cutting parts of the files and pasting them into a new files. I thought that the review in this case would be a formality. In the future, in a similar situation, I will break it into a series of atomic PRs.

kv2019i (approved):
Thanks @softwarecki , good for me!

src/include/module/module/base.h
```diff
@@ -0,0 +1,142 @@
...
+#endif /* SOF_MODULE_PRIVATE */
+};
+
+/* module_adapter private, runtime data - public version*/
+struct processing_module {
+	struct module_data priv; /**< module private data */
+	uint32_t period_bytes; /** pipeline period bytes */
+
+	/* Fields below can only be accessed by the SOF */
+#ifdef SOF_MODULE_API_PRIVATE
```
lgirdwood, line 60: Is this new, wont it change the ABI if its defined for some users and not others ?

lgirdwood (commented):
I'm good with 99%, but we cant have conditional code in the structures as a marker for PRIVATE and public APIs, this will mean different ABI for different users. We should use a comment here whilst any cleanup is in progress.

kv2019i:
@softwarecki wrote:
> From my perspective, it was just cutting parts of the files and pasting them into a new files. I thought that the review in this case would be a formality.

Note for future: Yes, this generally applies and this is how I would treat a PR like you describe. But for this to work well, the PR needs to be doing just this. If you mix monotonic changes (like renames or moves) with small functional changes, then either you need to explain that very well in the cover letter ("one functional commit and rest are just renames"), or otherwise you will freak out the reviewers when they browse through the code and note changes that go beyond the simple. I mean this PR has also not so trivial changes, like the changes to header inclusion dependencies and changes to lmdk build rules. So when title says "api extraction", but I see commits that go beyond this, then this is no longer a trivial PR. This is ok as well, but just requires more time to review each iteration.

src/include/module/module/base.h
softwarecki: @lgirdwood This symbol is defined in the file included by sof. Loadable modules should not define it and nor refer to this part of the structure. This structures should be divided in a future so sof can keep their private fileds in a different structure. It will take a lot of work and will require a lot of changes. Would it be enough for now to add a comment that this is a temporary solution until these fields are separated into a new structure?

marc-hb (commented):
> In the future, in a similar situation, I will break it into a series of atomic PRs.

I don't see why it's too late. You never have to submit all your local commits all at once. I `git push myfork HEAD~5:refs/heads/notEverythingAllAtOnce` all the time, keeping the last 5 commits (as an example) for later.  The later commits provide more, local test coverage for the submitted commits.

You can still do this now.

Sometimes I push a second invisible/draft PR for even more test coverage (this is what I did in #8495)

src/include/module/module/base.h
lgirdwood: Yeah, the comment need to be very visible and explain the roadmap forwards. That way its really obvious to everyone working on modules.

src/include/module/module/generic.h
lgirdwood: Modules today can and do include any header file (we still have a lot of header refactoring work todo) and we have vice versa where non module code (testbench, UTs, plugin) will need to include module headers.

softwarecki:
@lgirdwood @marc-hb @lyakh 
I did a rebase to resolve the conflicts and expanded the description in the comments regarding the added #ifdefs.

lgirdwood (approved):

PR #8367, opened 2023-10-20
[SKIP SOF-TEST] .github/zephyr: try to re-use already installed msys64

marc-hb:
This should save 20-30s and reduce the number of variables

### DIFF
--- a/.github/workflows/zephyr.yml
+++ b/.github/workflows/zephyr.yml
@@ -333,14 +333,12 @@ jobs:
           ninja.exe --version
 
       # MSYS2 provides gcc x64_86 toolchain & openssl
-      # Installs in D:/a/_temp/msys64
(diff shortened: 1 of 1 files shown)


marc-hb:
Works fine but is actually SLOWER! https://github.com/thesofproject/sof/actions/runs/6593075605/job/17914953491?pr=8367

Need to try not using this action at all.

lgirdwood:
> Works fine but is actually SLOWER! https://github.com/thesofproject/sof/actions/runs/6593075605/job/17914953491?pr=8367
> 
> Need to try not using this action at all.

Any update, what's our direction for this now ?

marc-hb:
This was just an experiment. Low priority, closing. Please don't look at drafts :-)

PR #8372, opened 2023-10-23
audio: copier: move host and dai to copier folder

btian1:
Host and dai are already been integrated into copier, it is not standalone modules, so move it to copier folder.


lgirdwood (commented):
LGTM

ranj063:
> Host and dai are already been integrated into copier, it is not standalone modules, so move it to copier folder.

But the host and DAI components for IPC3 are not the same as the copier

btian1:
> components for IPC3 are not

How about further separate with ipc3/ipc4 folder for host/dai-legacy and host/dai-zephyr?

RanderWang (approved):
LGTM

src/ipc/ipc4/dai.c
```diff
@@ -28,6 +27,7 @@
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
+#include <../audio/copier/dai_copier.h>
```
kv2019i, line 30: Please use "../audio/copier" so this is relative to this source file.

kv2019i (changes requested):
Hmm, this looks a bit messy as we have non-IPC elements that don't fit very nicely under "copier/" folder. @lgirdwood Do we really need to move copier? I don't think any module developer will use this as an example for anything.

src/ipc/ipc4/dai.c
btian1: I am still using: target_include_directories(sof_options INTERFACE ${RIMAGE_TOP}/src/include)
othewise, it should be:  "../../audio/copier"

these include path come from:
target_include_directories(SOF INTERFACE ${RIMAGE_TOP}/src/include)
target_include_directories(SOF INTERFACE ${SOF_SRC_PATH}/include)
in zephyr/CMakeList.txt.

lgirdwood:
> Hmm, this looks a bit messy as we have non-IPC elements that don't fit very nicely under "copier/" folder. @lgirdwood Do we really need to move copier? I don't think any module developer will use this as an example for anything.

Can you give an example of non-IPC elements that are causing concern ?

kv2019i:
@lgirdwood wrote:
> @kv2019i wrote:
> > Hmm, this looks a bit messy as we have non-IPC elements that don't fit very nicely under "copier/" folder. @lgirdwood Do we really need to move copier? I don't think any module developer will use this as an example for anything.
> 
> Can you give an example of non-IPC elements that are causing concern ?

Sorry, typo there, i mean "non-IPC3 elements". "dai-legacy.c" is not related to "copier" so "audio/copier/ipc3/dai-legacy.c" path is confusing. Also, "dai-zephyr.c" is not specific to IPC4, so having that in ipc4 folder is confusing (implying IPC4 and Zephyr would be coupled, which is not the case).

btian1:
> ements that a
@kv2019i your concern is reasonable, host and dai are not always exactly included by copier.

I also realized this, it is not easy to centralize these code, below is some possible option:
1. create two new folder in src/audio with name host / dai, and move all related code to src/audio/host src/audio/dai.
2. drop patch 1.

these host/dai code simply locate in src/audio looks not good, this is my start point.
let me know what do you think? @kv2019i @lgirdwood

src/ipc/ipc4/dai.c
lgirdwood: This is good as its gives us more visibility on where we have stickyness around legacy abstractions.

lgirdwood (approved):
LGTM, pls remeber this is a next step (but not final) in seperation of module code from RTOS/IPC code.

src/ipc/ipc4/dai.c
kv2019i: @btian1 @lgirdwood I'd still vote for "#include "../audio/copier" . and not "#include <.../audio/...". Mixing relative include paths and system includes is not clean -> https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html

kv2019i (changes requested):
Thanks @btian1 , this looks a lot cleaner. One small nit on using system-header syntax to refer to in-tree headers, see discussion inline.

lgirdwood (approved):

lgirdwood:
@kv2019i changes made.

kv2019i (approved):
Thanks @btian1 !

kv2019i:
The Zephyr main fails are unrelated to this PR, proceeding with merge.

PR #8374, opened 2023-10-23
Tools: Tune: TDFB: Add howto instructions to generated tplg blobs

singalsu:


singalsu:
In case wondering why the howto texts for dual beam blobs are for Matlab, the current Octave version in Ubuntu 23.04 segfaults with the setup script. Need to find out later why. I think it used to work with Octave earlier.

tools/tune/tdfb/example_circular_array.m
```diff
@@ -49,5 +49,7 @@ function example_circular_array()
 % Design
 bf = bf_filenames_helper(bf);
 bf = bf_design(bf);
+bf.export_note = 'Created with script example_circular_array.m';
```
lgirdwood, line 52: @singalsu @ShriramShastry do you guys know if matlab/octave has an equvalent of C argv[0] to get the program/script name ?
@singalsu we also need to rename all these scripts with a ```sof-``` prefix i.e. ```sof-iir.m```, ```sof-fir.m```, ```sof-tdfb.m``` and so on e.g. a combination of the sof prefix and the module it generates data for.

singalsu: @lgirdwood Yes there is, e.g.

mfilename('fullpath') - same in Octave and Matlab path and filename without .m
mfilename('class') - different in Octave and Matlab can't be used
mfilename - same in Octave and Matlab, filename without .m

I can add mfilename to make these texts

lgirdwood: great, thanks @singalsu - then next step is to do the renames. Once all renames are done then we can add to CMakefile install rules so that ```sof-tdfb-ca.m``` (for above file) is installed in target bin $PATH for developers. Intention is that we should have all tools installed locally for developers.

lgirdwood: btw, do matlab/octave scripts always need the ```.m``` i.e. can the shell execute the correct script interpreter for them without the .m  if they are invoked on the command line e.g.
```$sof-tdfb-ca -h```

singalsu: I think the shell launch works for /usr/bin/octave, like for Python. We should be able to do it for most of the top level scripts, and for rest once everything works with Octave.

lgirdwood: ok, lets default our upstream scripts to octave when invoked directly from shell.

btian1 (approved):

lgirdwood (approved):

PR #8378, opened 2023-10-24
topology2 : add DTS topology and changes for MTL

joechengxperi:
Add DTS topology and related changes for MTL with rt5682.  Will enable DTS later until rt5650 is ready.


tools/topology/topology2/cavs-mixin-mixout-eqiir-dts-ssp.conf
```diff
@@ -0,0 +1,129 @@
...
+#
+# PCM0 ---> gain ----> Mixin ----> Mixout ----> gain ----> EQ IIR ----> DTS ----> SSP0
+# PCM1 ---> gain ----> Mixin ----> Mixout ----> gain ----> EQ IIR ----> DTS ----> SSP1
+
+# Pipeline ID:1 PCM ID: 0
+Object.Pipeline {
+	# playback pipelines
+	host-copier-gain-mixin-playback [
+		{
+			index	1
```
lgirdwood, line 12: @jsarha @bardliao @plbossart do we still need index, IIRC we were (planning?) removing index from topology to be automatic.

lgirdwood (commented):
LGTM, just a question about index.

tools/topology/topology2/cavs-mixin-mixout-eqiir-dts-ssp.conf
ranj063: @lgirdwood yes it is still needed. There's no plan to make it automatic right now because widgets are identified with the index for top-level routes.

lgirdwood (approved):

tools/topology/topology2/cavs-mixin-mixout-ssp.conf
```diff
@@ -0,0 +1,95 @@
...
+			}
+			Object.Widget.gain.1 {
+				Object.Control.mixer.1 {
+					name	'Pre Mixer $SPEAKER_PCM_NAME Playback Volume'
+				}
+			}
+		}
+	]
+
+        mixout-gain-dai-copier-playback [
```
ranj063, line 40: this seems a bit too much duplication. Can we simply not create a new pipeline with dts called mixout-eqiir-dts-playback and then replace this with mixout-gain-dai-copier-playback when needed?

joechengxperi: I have one question. For "insert" you mentioned, do you mean using IncludeByKey to choose different pipeline or route at the top-level?

tools/topology/topology2/cavs-rt5682.conf
```diff
@@ -17,6 +17,7 @@
 <io-gateway-capture.conf>
 <host-copier-gain-mixin-playback.conf>
 <mixout-gain-dai-copier-playback.conf>
+<mixout-gain-eqiir-dts-dai-copier-playback.conf>
```
kv2019i, line 20: I guess this is referring to a file that is not yet available in the tree (added in the next patch), so this would break git bisect (e.g. after each git commit, the tree should build and be functional).

kv2019i (commented):
Ranjani and Jyri are the experts with a change like this, but a few comments from me as well. Mostly looks good, one smaller thing and then the question of how much existing definitions can be reused.

tools/topology/topology2/cavs-rt5682.conf
joechengxperi: This file is added in the first patch(b18b769ac2a81067d268fe08831e92e62ead085f). It shouldn't break the build. But it makes sense to move it to next patch.

joechengxperi:
> Ranjani and Jyri are the experts with a change like this, but a few comments from me as well. Mostly looks good, one smaller thing and then the question of how much existing definitions can be reused.

host-copier-gain-mixin-playback is duplicated. For mixout-gain-dai-copier-playback and mixout-gain-eqiir-dts-dai-copier-playback, dai copier and gain are duplicated. This is the question Ranjani asked.

I roughly have an idea to use IncludeByKey to implement Ranjani's suggestion. The difficult part is to just insert eq iir -> dts only. But the structure might become a little bit complicate comparing to current patch having two mixout-gain pipeline.

And I'm curious about the importance to avoid duplication here. I had two versions of top-level topology at the beginning.  The first one is more likely to follow cavs-nocodec.conf. And the second is to follow sof-hda-generic.conf.  Comparing these two version, I think sof-hda-generic.conf is more easy to read and could add more playback pipelines easily if necessary. But the drawback is too much duplication.

tools/topology/topology2/include/components/dts.conf
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	# Default attribute values for smart_amp widget
+
+	# UUID can be overridden in the top-level topology file
+	uuid			"4f:c3:5f:d9:0f:37:c7:4a:bc:86:bf:dc:5b:e2:41:e6"
+	type			"effect"
+	no_pm 			"true"
+	core_id			0
+	cpc                     5000
```
btian1, line 104: cpc only takes 5000? that's a small number.

tools/topology/topology2/include/components/dts/dts_spk.conf
```diff
@@ -0,0 +1,148 @@
+Object.Base.data."dts_spk" {
```
btian1, line 1: add comments for what's table used for?

joechengxperi: This table is the initial parameters to DTS library.

johnylin76: This looks like a full-spec config, which should be set from external rather than by default.
Why don't we use an empty config for default similar to topology1?

joechengxperi: @johnylin76 Yes, this is the full config with default parameters for DTS library. To use empty config or not set the parameters lead to some problems here. Instead, to put the default parameters here works fine.  We'll still need to set the parameters by UCM afterward since it varies by project.

johnylin76: The empty config I meant is the ABI-header-only blob, e.g. https://github.com/thesofproject/sof/blob/cavs2.5-001-drop-stable/tools/topology/topology1/m4/dts_codec_adapter.m4#L5
```
bytes "
0x53,0x4f,0x46,0x00,0x00,0x00,0x00,0x00,
0x00,0x00,0x00,0x00,0x00,0xa0,0x01,0x03,
0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,
0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00"
```
Have you tried this config and still led to problems?

I think we can set the full config here for developing purposes, but should not push this to upstream. Instead, a minimal and least effective config as default is preferred.

p.s. if the config above doesn't work, maybe try this instead:
```
bytes "
0x53,0x4f,0x46,0x00,0x00,0x00,0x00,0x00,
0x0c,0x00,0x00,0x00,0x00,0xa0,0x01,0x03,
0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,
0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,
0x00,0x00,0x00,0x00,0x0c,0x00,0x00,0x00,
0x00,0x00,0x00,0x00"
```

which contains a 12-byte module config {paramId=0, paramSize=0xc, paramData=[0x0]}

joechengxperi: @johnylin76 
I tried both of them before. They didn't work. But I notice a minor difference here now that makes the second one work. I'll test it further and update it.
```
bytes "
0x53,0x4f,0x46,0x34,0x00,0x00,0x00,0x00,
0x0c,0x00,0x00,0x00,0x00,0xa0,0x01,0x03,
0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,
0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,
0x01,0x00,0x00,0x00,0x0c,0x00,0x00,0x00,
0x00,0x00,0x00,0x00"
'''

joechengxperi: @johnylin76 Updated the table now.

johnylin76: The updated bytes will pass a param entry {paramId=1, paramSize=0xc, paramData=[0x0]} to DTS library. Given that paramId=1 is used in the full config, which should be the effective param ID for DTS library.

My previous idea is to pass {paramId=0, paramSize=0xc, paramData=[0x0]} which avoids the effective ID for DTS. However it sounds that errors are produced by paramId=0.

As long as DTS library has no side effect by setting {paramId=1, paramSize=0xc, paramData=[0x0]}, that looks good to me.

johnylin76 (approved):

kv2019i:
The ipc3-stub build failure is unrelated, proceeding with merge.

PR #8379, opened 2023-10-24
module_adapter : audio : fix DTS support on MTL

joechengxperi:
Add the commits to support DTS processing on MTL.


src/audio/CMakeLists.txt
```diff
@@ -169,7 +169,8 @@ check_optimization(fma -mfma -ftree-vectorize -DOPS_FMA)
 check_optimization(hifi2ep -mhifi2ep "" -DOPS_HIFI2EP)
```
dbaluta, line 169: Please split this into multiple patches. One patch per logical change.

kv2019i (commented):
Thanks @joechengxperi ! Code looks good, but like @dbaluta noted, it would be good to split the series. This is bit of a special case as DTS is only enabled if the overlay is selected, but in principle fixes should come first and the code enable for IPC4 as the last commit.

app/overlays/mtl/dts_overlay.conf
```diff
@@ -0,0 +1,3 @@
+CONFIG_COMP_IIR=y
```
btian1, line 1: I noticed this may be followed from:
./src/arch/xtensa/configs/override/mt8195_chrome_dts.config:4:CONFIG_DTS_CODEC=y
./src/arch/xtensa/configs/override/renoir_chrome_dts.config:3:CONFIG_DTS_CODEC=y
./src/arch/xtensa/configs/override/mt8186_chrome_dts.config:3:CONFIG_DTS_CODEC=y

my question is: is IIR must follow with DTS codec? they should have no relation from module perspecitve.

joechengxperi: We use EQ IIR + DTS to support customer project. For different platforms, EQ IIR might be default enabled or disabled. So we add it to make sure IIR is enabled. On MTL, IIR is default enabled. I could remove it if you think it's better.

marc-hb:
This is a good page

joechengxperi:
> Thanks @joechengxperi ! Code looks good, but like @dbaluta noted, it would be good to split the series. This is bit of a special case as DTS is only enabled if the overlay is selected, but in principle fixes should come first and the code enable for IPC4 as the last commit.

@kv2019i I've tried to split the patches and adjusted the order of commits. Please let me know if it's okay. Thanks.

joechengxperi:
> This is a good page
> 
> https://docs.zephyrproject.org/latest/contribute/contributor_expectations.html#defining-smaller-prs

This is super helpful. Thanks @marc-hb!

src/audio/CMakeLists.txt
joechengxperi: I've tried to split the patches. Please let me know if it's okay or not. Thanks.

src/include/sof/audio/dts/DtsSofInterfaceVersion.h
```diff
@@ -8,10 +8,10 @@
 
 #define DTS_SOF_INTERFACE_PRODUCT_NAME "DtsSofInterface"
 #define DTS_SOF_INTERFACE_PRODUCT_VERSION_MAJOR 1
-#define DTS_SOF_INTERFACE_PRODUCT_VERSION_MINOR 0
-#define DTS_SOF_INTERFACE_PRODUCT_VERSION_PATCH 7
+#define DTS_SOF_INTERFACE_PRODUCT_VERSION_MINOR 1
+#define DTS_SOF_INTERFACE_PRODUCT_VERSION_PATCH 3
 #define DTS_SOF_INTERFACE_PRODUCT_VERSION_BUILD 0
-#define DTS_SOF_INTERFACE_PRODUCT_VERSION_EXTRA "Dev"
+#define DTS_SOF_INTERFACE_PRODUCT_VERSION_EXTRA "RC2"
```
lgirdwood, line 14: Will RC2 be final for extra version ?

src/include/sof/audio/component.h
```diff
@@ -728,6 +728,8 @@ void sys_comp_module_src_interface_init(void);
 void sys_comp_module_tdfb_interface_init(void);
 void sys_comp_module_volume_interface_init(void);
 
+void sys_comp_module_dts_interface_init(void);
```
lgirdwood, line 731: nitpick: this should be in line above, next to the others.

app/overlays/mtl/dts_overlay.conf
lgirdwood: So the Kconfig can set a dependecy, we could use ```depends``` in Kconfig to state this relationship between DTS and IIR. 
@singalsu fyi.

lgirdwood (commented):
Looks good, I think only some minor opens from me.

btian1 (commented):
one suggestion： I suggest create a new folder under src/audio with codec/dts, and move out dts related into this folder.
Reason: now, most of modules are module interface, modules does not need put it under:
src/audio/module_adapter/module directory, it should have a standalone folder to cover.

ideally, all modules will be under src/audio with integrated all its source code in future.

src/include/sof/audio/dts/DtsSofInterfaceVersion.h
joechengxperi: We haven't found any other issue in current version. That's supposed to be.

src/audio/Kconfig
```diff
@@ -199,7 +199,7 @@ config COMP_IIR
 	bool "IIR component"
 	select COMP_BLOB
 	default y
-	depends on COMP_MODULE_ADAPTER
+	depends on COMP_MODULE_ADAPTER || DTS_CODEC
```
kv2019i, line 202: I think this is the wrong way. I'd be happy with your original solution and just have this in the overlay as this is not a strict dependency at code level, but a requirement at the topologoy level, so I think the overlay is a better place.

kv2019i (commented):
Excellent, thanks @joechengxperi , looks very good. Only issue with the last patch and I know you are addressing to feedback, so @lgirdwood @btian1 I'd propose we drop the last patch... ok?

src/audio/Kconfig
btian1: I also think this is not a good place, IIR can't depend on DTS, it is a standalone feature.

btian1 (commented):
@kv2019i @joechengxperi , please check my another comments, I am suggesting move out dts to src/audio/codec/dts, please let me know what do you think about this?

src/audio/Kconfig
joechengxperi: @lgirdwood May I know your feedback?

joechengxperi:
> one suggestion： I suggest create a new folder under src/audio with codec/dts, and move out dts related into this folder. Reason: now, most of modules are module interface, modules does not need put it under: src/audio/module_adapter/module directory, it should have a standalone folder to cover.
> 
> ideally, all modules will be under src/audio with integrated all its source code in future.

It sounds a reasonable move to me. If it will be the target to do, I'll work on it in another PR.

src/audio/module_adapter/module/dts/dts.c
```diff
@@ -429,11 +429,19 @@ dts_codec_set_configuration(struct processing_module *mod, uint32_t config_id,
...
 		return ret;
+	}
 
-	/* return if more fragments are expected or if the module is not prepared */
-	if ((pos != MODULE_CFG_FRAGMENT_LAST && pos != MODULE_CFG_FRAGMENT_SINGLE) ||
-	    md->state < MODULE_INITIALIZED)
+	/* return if more fragments are expected */
+	if (pos != MODULE_CFG_FRAGMENT_LAST && pos != MODULE_CFG_FRAGMENT_SINGLE) {
 		return 0;
+	}
```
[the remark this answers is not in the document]
joechengxperi: Done. Remove them now.

src/audio/module_adapter/module/dts/dts.c
```diff
@@ -429,11 +429,19 @@ dts_codec_set_configuration(struct processing_module *mod, uint32_t config_id,
...
+
+#if CONFIG_IPC_MAJOR_3
+	// return if the module is not prepared
+	if (md->state < MODULE_INITIALIZED) {
+		return 0;
+	}
```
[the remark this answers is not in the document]
joechengxperi: Done. Remove them now.

src/audio/module_adapter/module/dts/dts.c
```diff
@@ -427,18 +427,22 @@ dts_codec_set_configuration(struct processing_module *mod, uint32_t config_id,
 
 	ret = module_set_configuration(mod, config_id, pos, data_offset_size, fragment,
 				       fragment_size, response, response_size);
-	if (ret < 0)
+	if (ret < 0) {
```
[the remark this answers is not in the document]
joechengxperi: Done. Put them in the same commit now.

src/include/sof/audio/component.h
cujomalainey: not in alphabetical order

zephyr/CMakeLists.txt
```diff
@@ -711,6 +711,9 @@ zephyr_library_sources_ifdef(CONFIG_COMP_GOOGLE_HOTWORD_DETECT
 zephyr_library_sources_ifdef(CONFIG_DTS_CODEC
 	${SOF_AUDIO_PATH}/module_adapter/module/dts/dts.c
 )
+if (CONFIG_DTS_CODEC)
+zephyr_library_import(DtsCodec ${SOF_AUDIO_PATH}/module_adapter/lib/release/libdts-sof-interface-i32.a)
```
cujomalainey, line 715: zephyr_library_sources_ifdef so you don't need external if

src/audio/Kconfig
cujomalainey: @joechengxperi I agree, I am not sure why the config is defined this way, you should `select` the IIR component from the DTS component.

cujomalainey (changes requested):

zephyr/CMakeLists.txt
joechengxperi: Do you mean I could remove "if (CONFIG_DTS_CODEC)" and keep zephyr_library_import() only?

src/audio/Kconfig
joechengxperi: Thanks @cujomalainey. It looks better using select as you mentioned in this case. But after thinking about this a few days, I prefer to move it back to overlay rather than config. As @kv2019i said, it's not a strict dependency. Also, this dependency may probably change in the future, depends on customer requirement or platform. It seems using overlay and topology is more flexible.

johnylin76: They should not depend on each other.
Besides I don't think this meets your intention "make sure IIR is enabled on different platforms when DTS_CODEC is enabled", which is "if DTS=y; set IIR=y". If so, you should state "config DTS select IIR"

src/audio/CMakeLists.txt
```diff
@@ -169,7 +169,8 @@ check_optimization(fma -mfma -ftree-vectorize -DOPS_FMA)
...
 check_optimization(hifi3 -mhifi3 "" -DOPS_HIFI3)
 
-set(sof_audio_modules mixer volume src asrc eq-fir eq-iir dcblock crossover tdfb drc multiband_drc mfcc mux)
+set(sof_audio_modules mixer volume src asrc eq-fir eq-iir dcblock crossover
+    tdfb drc multiband_drc mfcc mux dts)
```
johnylin76, line 173: May I know why is this commit required?
If yes, may it be moved to src/audio/module_adapter/CMakeLists.txt?

zephyr/CMakeLists.txt
johnylin76: @cujomalainey I don't think zephyr_library_sources_ifdef can work since this is to import external libraries

joechengxperi: Yes, zephyr_library_sources_ifdef doesn't work to import library. I also tested it again to confirm it can't build pass. This leads to the question in my last comment.

src/audio/CMakeLists.txt
joechengxperi: I think it could be removed now as it's already handled in src/audio/module_adapter/CMakeLists.txt.

src/audio/Kconfig
marc-hb: Quoting another thread:
> We use EQ IIR + DTS to support customer project.

Overlays is where you define that, not in Kconfig files (and overlays don't have to be in this git repo, they can come from anywhere).

The purpose of Kconfig is to make sure the code compiles and does not crash and nothing else (and managing these basic dependencies is difficult and error-prone enough already, random Kconfig problem of the day: #8425). Kconfig is only for "hard science" and universal dependencies that affect _everyone_ using the code base, never for "preferences" and even less for "personal" preferences. 

While Kconfig came from the Linux kernel, Zephyr has IMHO better Kconfig documentation:

https://docs.zephyrproject.org/latest/build/kconfig/index.html
https://www.kernel.org/doc/html/latest/kbuild/kconfig-language.html

They should be bug-for-bug compatible so both documentations should apply.

I see the current version does not seem to change Kconfig
https://github.com/thesofproject/sof/commits/1ef2f59db0a

zephyr/CMakeLists.txt
andyross: Yeah, zephyr's cmake integration is elaborate and has lots of great hidden features like implicit ifdef/kconfig integration (also lots of hidden "you must use this" traps), but it's not as orthogonal as you'd want.  No one ever added a `zephyr_library_import_ifdef()`.

And in fact this function itself is a little obscure, and I had to look it up myself.  It's not actually used in-tree, even Zephyr's "external library" sample code doesn't use it, and makes do with cmake-native "add_library()/target_link_libraries()" code.  But it certainly looks like it does the right thing.  It just lacks kconfig integration. 

(That was way too much text.  This looks fine to me basically)

andyross: One nitpick might be the file location though: stuffing it down under src/audio seems weird?  Aren't we trying to put binary artifacts in /third_party?  Not the expert. Don't trust my review.

cujomalainey: yea i think i bad lookup on the existence of a function, apologies for the noise

cujomalainey: ditto to @andyross please import from third party folder

src/audio/Kconfig
joechengxperi: @marc-hb I removed that commit which added the wrong dependency and restored the original solution in overlay yesterday.

zephyr/CMakeLists.txt
btian1: @joechengxperi , why there is no directory and file with this path? module_adapter/lib/release/libdts-sof-interface-i32.a, or you are adding in a separate PR?

joechengxperi: @btian1 It's DTS proprietary library. I can't submit it here. Currently, it's only shared to the partners who need to test DTS functionality or Google for Chromebook project release.

joechengxperi: @cujomalainey I wonder if I could move the import to another PR. 
1. This is more like code reorganization rather than functionality.
2. I noticed that several google header files are under third_party folder. I'm thinking if it's necessary to move DTS header files here as well.
3. There is also one suggestion from @btian1  (https://github.com/thesofproject/sof/pull/8379#pullrequestreview-1698605677) to move DTS code to src/audio/
I think it might be better to put them together in another thread to discuss. Let me know how you think.

johnylin76: +1 for moving the DTS library header files and proprietary libraries (locally) to third_party/. So in the future one wouldn't expect the existence while seeing the library is imported from the path of third_party.

As for moving DTS code to src/audio/, it makes sense but I don't think we should do it now. Two reasons:
1) module API migration is a long-term project, the related optimization task cab be put off unless it blocks some thing.
2) it will be difficult to cherry-pick commits to production branch in the future once you move codes

marc-hb: > it will be difficult to cherry-pick commits to production branch in the future once you move codes

As long as you don't change the code too much while moving it, git is actually pretty good at cherry-picking changes across moves and renames. Easy enough to try.

Example with `sof/rimage` -> `sof/tools/rimage`  https://github.com/thesofproject/rimage/commit/b38fef772bf5ae6de2610ff5ff6e3a983a817f2

btian1:
> Excellent, thanks @joechengxperi , looks very good. Only issue with the last patch and I know you are addressing to feedback, so @lgirdwood @btian1 I'd propose we drop the last patch... ok?

This PR purpose is enable DTS on MTL, if drop last one, then DTS will not be enabled on MTL, my opinion is we can go for now, this feature does not have too much relation with sof other modules, we can go with this and fix new issue if have in future.

btian1 (approved):

kv2019i (approved):
This is fine to me. The location of the DTS source files should be moved to a better place, but I'm ok to do that in a follow-up PR (as we are anyways moving files around during SOF2.8 cycles that is now ongoing).

kv2019i:
@joechengxperi wrote:
> @lgirdwood For this PR, may I know if there is still anything I have to do for merge? Thanks.

@cujomalainey has a -1 on this, that's blocking the merge now. Otherwise looks good to go.

lgirdwood:
> @joechengxperi wrote:
> 
> > @lgirdwood For this PR, may I know if there is still anything I have to do for merge? Thanks.
> 
> @cujomalainey has a -1 on this, that's blocking the merge now. Otherwise looks good to go.

@joechengxperi I assume all of @cujomalainey comments are aligned/closed now ? if so, he should be able to re-review.

cujomalainey (approved):

zephyr/CMakeLists.txt
```diff
@@ -711,6 +711,9 @@ zephyr_library_sources_ifdef(CONFIG_COMP_GOOGLE_HOTWORD_DETECT
 zephyr_library_sources_ifdef(CONFIG_DTS_CODEC
 	${SOF_AUDIO_PATH}/module_adapter/module/dts/dts.c
 )
+if (CONFIG_DTS_CODEC)
```
cujomalainey, line 714: This broke the stub build. Please fix

cujomalainey (changes requested):
sorry spotted one more thing

zephyr/CMakeLists.txt
marc-hb: > This broke the stub build. Please fix

@joechengxperi it is the responsibility of the submitter to look at failures, not the reviewers. Ask for help if you don't understand some failures (or worse: if you don't have access to some)

https://github.com/thesofproject/sof/actions/runs/6718130608/job/18257259780?pr=8379

The last error messages can be misleading. Always scroll up and look at the first errors first. Parallel builds may require scrolling quite a way up.

cujomalainey (approved):

zephyr/CMakeLists.txt
joechengxperi: Fixed. Apologize that I did not notice that build error.

joechengxperi:
> > @joechengxperi wrote:
> > > @lgirdwood For this PR, may I know if there is still anything I have to do for merge? Thanks.
> > 
> > 
> > @cujomalainey has a -1 on this, that's blocking the merge now. Otherwise looks good to go.
> 
> @joechengxperi I assume all of @cujomalainey comments are aligned/closed now ? if so, he should be able to re-review.

Thank you @lgirdwood @kv2019i. Yes, I think those are all aligned now.

cujomalainey:
On device test failure is unrelated to this change, merging

PR #8382, opened 2023-10-24
mtl: lnl: Kd topology failure fix

jxstelter:
KD topology tests fail on MTL and LNL due to insufficient size of
heap memory. During creation of KD topologies with
4ch audio format, FW fails on memory allocation.
The patch increases HEAPMEM size.


marcinszkudlinski (approved):
as for now looks like we do have enough memory...

tmleman (approved):

lgirdwood (approved):
@jxstelter @marcinszkudlinski @kv2019i btw, whats the plan to remove this hard coding as I assume Zephyr heap should be able to be set to max SRAM size (and dynamically power on/off memory banks).

jxstelter:
> @jxstelter @marcinszkudlinski @kv2019i btw, whats the plan to remove this hard coding as I assume Zephyr heap should be able to be set to max SRAM size (and dynamically power on/off memory banks).
I don't think that heap should be set to max SRAM. Large part of memory will be used by loadable libraries and this is not heap allocated space but space mapped from unused SRAM and mapped to particular virtual space. The space is then assigned to module memory domain.

PR #8383, opened 2023-10-24
Tools: Testbench: Add xt-testbench build for mtl

singalsu:
See commits for changes descriptions

### DIFF
--- a/scripts/rebuild-testbench.sh
+++ b/scripts/rebuild-testbench.sh
@@ -57,6 +57,7 @@ setup_xtensa_tools_build()
     BUILD_TYPE=xt
     BUILD_TARGET=
     BUILD_DIR_NAME=build_xt_testbench
+    COMPILER="xt-xcc"
 
     # check needed environment variables
     test -n "${XTENSA_TOOLS_ROOT}" || die "XTENSA_TOOLS_ROOT need to be set.\n"
@@ -72,11 +73,10 @@ setup_xtensa_tools_build()
     test -n "${XTENSA_CORE}" ||
         die "Illegal platform $BUILD_PLATFORM, no XTENSA_CORE found.\n"
 
-    compiler="xt-xcc"
     install_bin=install/tools/$XTENSA_TOOLS_VERSION/XtensaTools/bin
     tools_bin=$XTENSA_TOOLS_ROOT/$install_bin
     testbench_sections="-Wl,--sections-placement $BUILD_TESTBENCH_DIR/testbench_xcc_sections.txt"
-    export CC=$tools_bin/$compiler
+    export CC=$tools_bin/$COMPILER
     export LD=$tools_bin/xt-ld
     export OBJDUMP=$tools_bin/xt-objdump
     export LDFLAGS="-mlsp=sim $testbench_sections"
--- a/scripts/set_xtensa_params.sh
+++ b/scripts/set_xtensa_params.sh
@@ -81,4 +81,10 @@ case "$platform" in
 	HOST="xtensa-mt8195-elf"
 	XTENSA_TOOLS_VERSION="RI-2019.1-linux"
 	;;
+    mtl)
+	PLATFORM="mtl"
+	XTENSA_CORE="ace10_LX7HiFi4_2022_10"
+	XTENSA_TOOLS_VERSION="RI-2022.10-linux"
+	COMPILER="xt-clang"
+	;;
 esac
--- a/tools/testbench/testbench.c
+++ b/tools/testbench/testbench.c
@@ -353,13 +353,14 @@ static int parse_input_args(int argc, char **argv, struct testbench_prm *tp)
 			break;
 
 		/* print usage */
-		default:
-			fprintf(stderr, "unknown option %c\n", option);
-			ret = -EINVAL;
-			/* fallthrough */
 		case 'h':
 			print_usage(argv[0]);
 			exit(EXIT_SUCCESS);
+
+		default:
+			fprintf(stderr, "unknown option %c\n", option);
+			print_usage(argv[0]);
+			ret = -EINVAL;
 		}
 
 		if (ret < 0)
--- a/tools/tplg_parser/CMakeLists.txt
+++ b/tools/tplg_parser/CMakeLists.txt
@@ -64,7 +64,7 @@ endif()
 
 # TODO: add IPC4 option when it's ready.
 target_compile_options(sof_tplg_parser PRIVATE
-	-g -O -Wall -Werror -Wl,-EL -fPIC -DPIC
+	-g -O -Wall -Werror -fPIC -DPIC
 	-Wmissing-prototypes ${implicit_fallthrough}
 	-DCONFIG_LIBRARY -D${tplg_ipc})
 
--- a/tools/testbench/CMakeLists.txt
+++ b/tools/testbench/CMakeLists.txt
@@ -41,7 +41,7 @@ if (supports_implicit_fallthrough)
     set(implicit_fallthrough -Wimplicit-fallthrough)
 endif()
 
-target_compile_options(testbench PRIVATE -g -O3 -Wall -Werror -Wl,-EL -Wmissing-prototypes
+target_compile_options(testbench PRIVATE -g -O3 -Wall -Werror -Wmissing-prototypes
   ${implicit_fallthrough} -DCONFIG_LIBRARY -DCONFIG_LIBRARY_STATIC -imacros${config_h})
 
 target_link_libraries(testbench PRIVATE -lm)


scripts/set_xtensa_params.sh
```diff
@@ -81,4 +81,11 @@ case "$platform" in
 	HOST="xtensa-mt8195-elf"
 	XTENSA_TOOLS_VERSION="RI-2019.1-linux"
 	;;
+    mtl)
+	PLATFORM="mtl"
+	XTENSA_CORE="ace10_LX7HiFi4_2022_10"
+	XTENSA_TOOLS_VERSION="RI-2022.10-linux"
+	COMPILER="xt-clang"
+	LDEXTRAFLAGS=
+	;;
```
singalsu, line 90: @marc-hb Do you have ideas how to clean-up set_xtensa_params.sh. I guess this was duplicated from zephyr build when it was still a shell script.

scripts/rebuild-testbench.sh
```diff
@@ -56,6 +56,8 @@ setup_xtensa_tools_build()
     BUILD_TYPE=xt
     BUILD_TARGET=
     BUILD_DIR_NAME=build_xt_testbench
+    COMPILER="xt-xcc"
+    LDEXTRAFLAGS="-Wl,-LE"
```
marc-hb, line 60: This flag never had any effect because there is no `./E/` directory. I'm dropping it in #8384

tools/tplg_parser/CMakeLists.txt
```diff
@@ -64,7 +64,7 @@ endif()
 
 # TODO: add IPC4 option when it's ready.
 target_compile_options(sof_tplg_parser PRIVATE
-	-g -O -Wall -Werror -Wl,-EL -fPIC -DPIC
+	-g -O -Wall -Werror -fPIC -DPIC
```
marc-hb, line 67: Do you remember why you specified the endianness in commit 0c35f1efde17bbd27d?

scripts/rebuild-testbench.sh
singalsu: (facepalm) good that you spotted it, I was yesterday trying to find from manual what it's doing :)

scripts/set_xtensa_params.sh
singalsu: I think there's a lot of unused environment variables like HOST. Or at least MTL built without setting it, but maybe it's for gcc crosscompiler for other platforms. Currently testbench build is not supporting it.

tools/tplg_parser/CMakeLists.txt
singalsu: No, I don't. I have probably copied it from somewhere else from that time SOF, and it looks totally unnecessary. Thanks for checking this draft!

src/include/sof/trace/preproc.h
```diff
@@ -158,8 +158,10 @@
 		return_t META_CONCAT(prefix, postfix) (args)
 
 /* counteract compiler warning about unused variables */
+#ifndef UNUSED
 #define UNUSED(arg1, ...) do { META_RECURSE( \
 	META_MAP_AGGREGATE(1, _META_VOID2, _META_VOID(arg1), __VA_ARGS__)); \
 	} while (0)
+#endif
```
singalsu, line 165: I feel this change is not good. Would it be better to rename UNUSED() to SOF_UNUSED() or SOF_TRACE_UNUSED()?

lgirdwood (approved):
I guess you can deal with teh UNUSED macro name in a subsequent PR ?

src/include/sof/trace/preproc.h
marc-hb: Yes please rename this to SOF_TRACE_UNUSED() in a separate commit (or even separate PR).

> I Think the pre-processing code is now only used for logging, could be deprecated...

... for Intel.

tools/testbench/testbench.c
```diff
@@ -353,13 +353,14 @@ static int parse_input_args(int argc, char **argv, struct testbench_prm *tp)
...
-			ret = -EINVAL;
-			/* fallthrough */
 		case 'h':
 			print_usage(argv[0]);
 			exit(EXIT_SUCCESS);
+
+		default:
+			fprintf(stderr, "unknown option %c\n", option);
+			print_usage(argv[0]);
+			ret = -EINVAL;
```
marc-hb, line 363: You're getting away with this problem with a refactoring this time around but it's "selfish"! How about other statements elsewhere in the code?

Could you try to find something portable across both gcc and clang?
https://stackoverflow.com/questions/45129741/gcc-7-wimplicit-fallthrough-warnings-and-portable-way-to-clear-them

marc-hb (dismissed):
SOF_TRACE_UNUSED

tools/testbench/testbench.c
singalsu: I'll try if there's something that works for all compilers with help of that link (yesterday I didn't find).

src/include/sof/trace/preproc.h
```diff
@@ -158,7 +158,7 @@
 		return_t META_CONCAT(prefix, postfix) (args)
 
 /* counteract compiler warning about unused variables */
-#define UNUSED(arg1, ...) do { META_RECURSE( \
```
singalsu, line 161: @marc-hb  I added this separately as #8395 so I will remove it this commit.

tools/testbench/testbench.c
marc-hb: Can you please try something like this? It supposedly works with both gcc and clang.

```C
#if defined __has_cpp_attribute && __has_cpp_attribute(fallthrough)
#  define SOF_FALLTHRU __attribute__((__fallthrough__)
#elif <FUTURE_UNUSUAL_COMPILER>
#  define SOF_FALLTHRU <something unusual>
#else
#  warning unknown compiler, undefined SOF_FALLTHRU
#endif

```

marc-hb: https://github.com/zephyrproject-rtos/zephyr/tree/main/include/zephyr/toolchain uses `__attribute__((__fallthrough__)` for both gcc and clang.

src/include/sof/trace/trace.h
```diff
@@ -93,7 +93,7 @@ static inline struct trace *trace_get(void)
 
 /* Silences compiler warnings about unused variables */
 #define trace_unused(class, ctx, id_1, id_2, format, ...) \
-	UNUSED(ctx, id_1, id_2, ##__VA_ARGS__)
+	SOF_TRACE_UNUSED(ctx, id_1, id_2, ##__VA_ARGS__)
```
btian1, line 96: please remove this patch, since you submitted with a separate PR.

marc-hb:
`-Wl,-EL` déjà vu:
https://github.com/thesofproject/rimage/commit/bcbcec79e582

tools/testbench/testbench.c
singalsu: It's used with #define like this:

#ifndef __fallthrough
#if __GNUC__ >= 7
#define __fallthrough        __attribute__((fallthrough))
#else
#define __fallthrough
#endif	/* __GNUC__ >= 7 */
#endif

singalsu: I still feel my code re-order is simplest way to address this. The fall-through is unnecessary trickery for a very simple thing.

singalsu: Also if you look the original code, there's a bug that I'm fixing the same time:

```
		default:
			fprintf(stderr, "unknown option %c\n", option);
			ret = -EINVAL;
			/* fallthrough */
		case 'h':
			print_usage(argv[0]);
			exit(EXIT_SUCCESS);
		}

		if (ret < 0)
			return ret;

```

The `-EINVAL `is not returned because of `exit(EXIT_SUCCESS)`. Fall-through is not correct here.

src/include/sof/trace/trace.h
singalsu: Yep, rebase dropped it.

tools/testbench/testbench.c
marc-hb: > unnecessary trickery for a very simple thing

I never said your change was wrong, I said it was "selfish" and I stand by that. This type of issue is painful to test because you need multiple toolchains and many people use only one. But you already have the environment, skills and a good test case. 

That "trickery" would really not be that complicated (try `git grep META` for some _real_ trickery!) and it would be only in one or two .h files; just one very simple SOF_FALLTHRU keyword everywhere else.


> `#if __GNUC__ >= 7` 

We don't need stuff like that for `testbench.c`, so it could be very simple for now.

marc-hb: > Also if you look the original code, there's a bug that I'm fixing the same time: [...]
> Fall-through is not correct here.

The current commit message still refers to fall through only.

singalsu: @marc-hb I just updated the commit text. Is it OK now?

marc-hb: Sure. Just for the record I never formally "requested" fallthrough changes :-)

marc-hb (approved):

PR #8385, opened 2023-10-25
mtk: fix memory allocation problem

wenliangwu:
Similar to the issue described in https://github.com/thesofproject/sof/pull/7982, MediaTek is also facing out-of-memory problems in our use cases. We have implemented the same solution to address this issue. 
Additionally, I fixed the compilation error "redefinition of i" in module adaptor while using xtensa toolchain.


iuliana-prodan (approved):

yaochunhung (approved):

chunxu-li (approved):

src/audio/module_adapter/module_adapter.c
```diff
@@ -1450,9 +1450,9 @@ int module_adapter_reset(struct comp_dev *dev)
 #endif /* CONFIG_ZEPHYR_DP_SCHEDULER */
 	if (IS_PROCESSING_MODE_SINK_SOURCE(mod)) {
 		/* for both LL and DP processing */
-		for (int i = 0; i < mod->num_of_sources; i++)
+		for (i = 0; i < mod->num_of_sources; i++)
```
lgirdwood, line 1453: @singalsu @iuliana-prodan did we not recently change this in another place ? I think the newer xtensa compilers are possibly stricter here and hence we are hitting this now.

src/platform/mt8186/include/platform/lib/memory.h
```diff
@@ -115,7 +115,7 @@
 #define HEAP_RT_COUNT64		32
 #define HEAP_RT_COUNT128	32
 #define HEAP_RT_COUNT256	32
-#define HEAP_RT_COUNT512	4
+#define HEAP_RT_COUNT512	32
```
lgirdwood, line 118: Fwiw, Intel has not had to make a change like this since switching RTOS to Zephyr.

lgirdwood (approved):
LGTM, but would like to get confirmation from @singalsu about prior fix for int i redefinition.

kv2019i:
@wszypelt There is one Intel test case failing (test_003_01_kd_topology_host_gw), but as this PR doesn't really touch any Intel code, I can't see how this is related. So can I proceed to merge, or can you rerun the test just to be sure?

src/platform/mt8186/include/platform/lib/memory.h
btian1: https://github.com/thesofproject/sof/blob/main/src/platform/library/include/platform/lib/memory.h#L97
is this intel's definition? still used by ipc4 build?

Btw, 4-->32 too big? do you observe around 15k bytes needed to enlarge?

wenliangwu: After confirming that 15k is only a small portion of the heap buffer, I am uncertain whether there may be other use cases that require a larger buffer size. Therefore, I decided to follow the solution in #7982.

wszypelt:
@kv2019i I'll try to test PR again, results should be available within 30 minutes

PR #8387, opened 2023-10-25
[SOF-v2.7] backport fixes for CI builds

kv2019i:
Fix the Zephyr Windows build and Intel QB TGL build with following backports:

ec0667a2b4f  xtensa-build-zephyr: remove ipc option
4e74d2b606 .github/zephyr: de-hardcode the remote name in the Windows build too

No changes to functionality.

### DIFF
--- a/.github/workflows/zephyr.yml
+++ b/.github/workflows/zephyr.yml
@@ -310,7 +310,7 @@ jobs:
           # Get some tags to fix `git describe`, see build-linux comments above.
           cd zephyr
           $_rev = "$(git rev-parse HEAD)"
-          git fetch --filter=tree:0 zephyrproject "${_rev}:_branch_placeholder"
+          git fetch --filter=tree:0 "$(west list -f '{url}' zephyr)" "${_rev}:_branch_placeholder"
           git branch -D _branch_placeholder
 
 
--- a/scripts/xtensa-build-zephyr.py
+++ b/scripts/xtensa-build-zephyr.py
@@ -176,9 +176,6 @@ def parse_args():
 						help="List of platforms to build")
 	parser.add_argument("-d", "--debug", required=False, action="store_true",
 						help="Shortcut for: -o sof/app/debug_overlay.conf")
-	parser.add_argument("-i", "--ipc", required=False, choices=["IPC4"],
-			    help="""Applies --overlay <platform>/ipc4_overlay.conf
-and a different rimage config. Valid only for IPC3 platforms supporting IPC4 too.""")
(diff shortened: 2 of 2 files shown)


kv2019i:
@marc-hb Uh, this is hopeless, it would seem we would need to backport a lot of github workflow stuff to get the Windows builds working. Now we have issues with removing the IPC4 option, and the problem with ninja tool missing. If we get the QB build to pass, I think I'll stop here.

lgirdwood (approved):

marc-hb:
I don't think Github Actions are hopeless on this branch: in fact some Linux configurations _and_ some Windows configurations did compile in https://github.com/thesofproject/sof/actions/runs/6639325877/job/18037461117?pr=8387! I'll take a look at the failures today. We'll most likely need my `choco install` workaround.


I don't understand why you're cherry-picking the IPC4 removal though?

> If we get the QB build to pass, I think I'll stop here.

On the contrary, I think _that_ is hopeless. QB never new anything about branches, see internal issue PTS-45980

Granted, the failure in https://sof-ci.01.org/sof-pr-viewer/#/build/PR8387/build13001947 looks like a temporary glitch, not like a real incompatibility and problem. But it's only a matter of time. That's because Github Actions are branched at the same time and QB code (or Jenkins code for that matter) is not.

marc-hb:
SOFCI TEST

marc-hb:
> I don't think Github Actions are hopeless on this branch: in fact some Linux configurations and some Windows configurations did compile in https://github.com/thesofproject/sof/actions/runs/6639325877/job/18037461117?pr=8387! I'll take a look at the failures today. We'll most likely need my choco install workaround.

I looked at these logs. If you drop the IPC4 removal and cherry-pick the `choco install` (#8254) then everything should Just Work (TM)

kv2019i:
@marc-hb wrote:
> I don't understand why you're cherry-picking the IPC4 removal though?

It has magic code to use the IPC4 rimage toml file if you build for "tgl" and do not pass any IPC version on the command-line.

Without this, QB will build a IPC4 SOF binary, but use IPC3 toml file to prepare the ri file and resulting binary will not boot.

>> If we get the QB build to pass, I think I'll stop here.
> 
> On the contrary, I think _that_ is hopeless. QB never new anything about branches, see internal issue PTS-45980

Maybe so. I was checking the QB build log and to my eyes, it picked all the correct stuff (with west). 

Now the build failed to some other problem (west init), so maybe this is a hopeless road after all.

marc-hb:
Works fine in #8406

PR #8389, opened 2023-10-25
topology1: remove Intel GLK/TGL/ADL/RPL topology definitions

kv2019i:
These topologies are no longer supported by mainline firmware code for these platforms. The stable-v2.2 branch continues to support these targets with topology1. Also SOF main branch provides topology2 targets for a subset of these configurations.


lgirdwood:
> @singalsu any thoughts what we should do with testbench. It now fails due to missing definitions:
> 
> https://github.com/thesofproject/sof/actions/runs/6640839670/job/18042051210?pr=8389
> 
> ```
> m4:test-capture.m4:9: cannot open `dmic.m4': No such file or directory
> m4:test-all.m4:18: cannot open `platform/intel/bxt.m4': No such file or directory
> m4:test-capture.m4:19: cannot open `platform/intel/bxt.m4': No such file or directory
> ```

I think testbench CI check would need to use topology2 examples. @singalsu any suggestions ?

kv2019i:
@singalsu @lgirdwood Or one alternative is we maintain a minimal set of Intel tplg1 m4 files to use in testbench. Then we could do some form of IPC3 debug via testbench. This would still ensure tplg patches will not be sent to wrong branches (e.g. tplg1 Intel stuff should go to stable-v2.2 now).

marc-hb:
> This would still ensure tplg patches will not be sent to wrong branches (e.g. tplg1 Intel stuff should go to stable-v2.2 now).

To achieve that specific objective, trimming the CMakeLists.txt files is enough, no need to touch the .m4 file. Except for people who edit code and submit changes without even compiling them first but I genuinely think these rare people would get what they deserve :-)

lyakh (approved):
@singalsu I presume this is fine with you too?

singalsu:
> @singalsu I presume this is fine with you too?

The test topologies build still fails. @kv2019i should remove generate of topologies those depend on dmic.m4. It's probably the shell scripts in tools/test/topology.

I've been doing to tools/topology/topology2/include/bench replacement for tplg1 topololgies for test but the work is not yet ready. ETA for ipc4 & tplg2 testbench is end of this Q.

singalsu:
@kv2019i I just sent to you proposal how to fix test topologies build for scripts/host-testbench.sh usage.

kv2019i:
V2 pushed:
 - include Seppo's change to testbench

singalsu (approved):

PR #8394, opened 2023-10-26
[stable-v2.2] intel: ssp: implement mclk always-on feature

ujfalusi:
In the earliest design, mclk is turned on in set_config() and never turned off. The behavior is modified in this commit:

d5840a920073 ("ssp: move mclk request/release to pre_start and post_stop")

Now the mclk is enabled only during audio streaming so some codec may pop. This commit adds the always-on behavior back as an new feature.

Signed-off-by: Brent Lu <brent.lu@intel.com>
(cherry picked from commit 8dfc008b8b99a1bdbe7707739b62714aaa65abf6)


ujfalusi:
This patch missed the stable-v2.2 branch, picked it from main (it is also pick to cavs2.5-001-drop-stable)

Vamshigopal (approved):

kv2019i (approved):

PR #8240, opened 2023-09-21
topology2: merge multi-core test into nocodec topology

RanderWang:
The multi-core topology was created since multi-core feature had a few bugs at that time so it was for validation test. Now this feature works fine so merge it into nocodec topology then it will be tested by github CI. And we don't need to maintain a dedicated multi-core topology.
Screenshot from 2023-09-21 16-00-45

Issue #8056, opened 2023-08-17
[BUG] FW reported error: 104 - Other failure of module instance initialization request


fredoh9:
**Describe the bug**
During check-suspend-resume-with-playback-5.sh test, FW reported error: 104 is occurred.

```
[ 4410.777465] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx reply: 0x60000068|0x10036: MOD_INIT_INSTANCE
[ 4410.777468] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: FW reported error: 104 - Other failure of module instance initialization request
[ 4410.777519] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc error for msg 0x40010004|0x10036
[ 4410.777548] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: failed to create module alh-copier.SDW0-Playback.0
```

Intel internal daily test link:

- planresultdetail/30296?model=MTLP_RVP_SDW&testcase=check-suspend-resume-with-playback-5
- planresultdetail/30296?model=MTLP_SDW_AIOC&testcase=check-suspend-resume-with-playback-5

**To Reproduce**
TPLG=/lib/firmware/intel/sof-ace-tplg/sof-mtl-rt711-4ch.tplg MODEL=MTLP_RVP_SDW SOF_TEST_INTERVAL=5 ~/sof-test/test-case/check-suspend-resume-with-audio.sh -l 5 -m playback

**Reproduction Rate**
Starting from today. It happened in MTLP_RVP_SDW and MTLP_SDW_AIOC.

**Environment**

- Kernel Branch: topic/sof-dev
- Kernel Commit: [b49e881b4db3](https://github.com/thesofproject/linux/commit/b49e881b4db3)
- SOF Branch: main
- SOF Commit: [eac5a0a12ef4](https://github.com/thesofproject/sof/commit/eac5a0a12ef4)

**Screenshots or console output**

fredoh9:
Found in ADLP_SKU0B00_SDCA also from today's daily test.

Intel internal daily test link:
planresultdetail/30853?model=ADLP_SKU0B00_SDCA-cavs2_5-ipc4&testcase=check-suspend-resume-with-playback-5

fredoh9:
Not reproduced for 3 days

keqiaozhang:
[...]

PR #8398, opened 2023-10-26
mailbox: use uncached address to write sw regs

RanderWang:
mailbox_sw_regs_write is used by dai to update llp node. In multi-core test, the updated llp node is invisable to other cores although dcache_writeback_region is used on both cavs & ace platforms. Now use uncached address to write memory directly which is aligned with other mailbox access functions.

Validated on cavs & ace platforms.

multi-core topology: https://github.com/thesofproject/sof/pull/8240

### DIFF
--- a/src/platform/intel/ace/include/ace/lib/mailbox.h
+++ b/src/platform/intel/ace/include/ace/lib/mailbox.h
@@ -80,12 +80,12 @@ static inline uint64_t mailbox_sw_reg_read64(size_t offset)
 
 static inline void mailbox_sw_regs_write(size_t offset, const void *src, size_t bytes)
 {
-	int regs_write_err __unused = memcpy_s((void *)(MAILBOX_SW_REG_BASE + offset),
-					       MAILBOX_SW_REG_SIZE - offset, src, bytes);
+	uint32_t __sparse_cache *ptr_c;
+	uint32_t *ptr;
 
-	assert(!regs_write_err);
-	dcache_writeback_region((__sparse_force void __sparse_cache *)(MAILBOX_SW_REG_BASE +
-								       offset), bytes);
+	ptr_c = (uint32_t __sparse_cache *)(MAILBOX_SW_REG_BASE + offset);
+	ptr = cache_to_uncache(ptr_c);
+	memcpy_s(ptr, MAILBOX_SW_REG_SIZE - offset, src, bytes);
 }
 
 #endif /* __ACE_LIB_MAILBOX_H__ */
--- a/src/platform/intel/cavs/include/cavs/lib/mailbox.h
+++ b/src/platform/intel/cavs/include/cavs/lib/mailbox.h
@@ -108,12 +108,12 @@ static inline uint64_t mailbox_sw_reg_read64(size_t offset)
 
 static inline void mailbox_sw_regs_write(size_t offset, const void *src, size_t bytes)
 {
-	int regs_write_err __unused = memcpy_s((void *)(MAILBOX_SW_REG_BASE + offset),
-					       MAILBOX_SW_REG_SIZE - offset, src, bytes);
+	uint32_t __sparse_cache *ptr_c;
+	uint32_t *ptr;
 
-	assert(!regs_write_err);
-	dcache_writeback_region((__sparse_force void __sparse_cache *)(MAILBOX_SW_REG_BASE +
-								       offset), bytes);
+	ptr_c = (uint32_t __sparse_cache *)(MAILBOX_SW_REG_BASE + offset);
+	ptr = cache_to_uncache(ptr_c);
+	memcpy_s(ptr, MAILBOX_SW_REG_SIZE - offset, src, bytes);
 }
 
 #endif /* __CAVS_LIB_MAILBOX_H__ */


RanderWang:
please check the following log.   core1 and core2 allocate the same slot offset and have the different content of mailbox (the bold lines should be same). The time stamp show that there is no competition risk.

[    0.449035] <inf> ipc: dai_init_llp_info: comp:10 0xe0004 +++ alloc slot 77c, now 19
[    0.476428] <err> ipc: dai_get_unused_llp_slot: comp:11 0xf0004 used slot c00
**[    0.476431] <err> ipc: dai_get_unused_llp_slot: comp:11 0xf0004 used slot c20**
[    0.476440] <err> ipc: dai_get_unused_llp_slot: comp:11 0xf0004 used slot b00
[    0.476445] <err> ipc: dai_get_unused_llp_slot: comp:11 0xf0004 *** core[1] confirm slot d10
[    0.476448] <inf> ipc: dai_init_llp_info: comp:11 0xf0004 +++ alloc slot 77c, now 18
[    0.594686] <err> ipc: dai_get_unused_llp_slot: comp:16 0x140004 used slot c00
**[    0.594688] <err> ipc: dai_get_unused_llp_slot: comp:16 0x140004 used slot d00**
[    0.594696] <err> ipc: dai_get_unused_llp_slot: comp:16 0x140004 used slot b00
[    0.594701] <err> ipc: dai_get_unused_llp_slot: comp:16 0x140004 used slot c10
[    0.594705] <err> ipc: dai_get_unused_llp_slot: comp:16 0x140004 *** core[2] confirm slot d20

src/platform/intel/ace/include/ace/lib/mailbox.h
```diff
@@ -80,12 +80,12 @@ static inline uint64_t mailbox_sw_reg_read64(size_t offset)
...
-					       MAILBOX_SW_REG_SIZE - offset, src, bytes);
+	uint32_t __sparse_cache *ptr_c;
+	volatile uint32_t *ptr;
 
-	assert(!regs_write_err);
-	dcache_writeback_region((__sparse_force void __sparse_cache *)(MAILBOX_SW_REG_BASE +
-								       offset), bytes);
+	ptr_c = (uint32_t __sparse_cache *)(MAILBOX_SW_REG_BASE + offset);
+	ptr = cache_to_uncache(ptr_c);
+	memcpy_s(ptr, MAILBOX_SW_REG_SIZE - offset, src, bytes);
```
kv2019i, line 88: This is quite worrying finding, how come this has impact? Is the register read via cached address by other cores? This still doesn't explain why this PR would have any impact. We should reallly understand why dcache_writeback_region() is not working, we rely on this in many places.

kv2019i (commented):
@marcinszkudlinski @lyakh please take a look, we'd have to understand why this PR has an impact.

src/platform/intel/ace/include/ace/lib/mailbox.h
RanderWang: Yes, I am also very surprised. The reading method is mailbox_sw_reg_read, so it should be ok. This change use the same method for mailbox_sw_reg_read. 

And there is a PR to switch to uncached memory by who ?

RanderWang: Linux driver doesn't access llp in mailbox now

RanderWang:
llp code in dai

````c
	key = k_spin_lock(&sof_get()->fw_reg_lock);

	/* find unused llp slot offset with node_id of zero */
	for (i = 0; i < max_slot; i++, offset += sizeof(slot)) {
		uint32_t node_id;

		node_id = mailbox_sw_reg_read(offset);
		if (!node_id)
			break;
	}

	if (i >= max_slot) {
		comp_err(dev, "can't find free slot");
		k_spin_unlock(&sof_get()->fw_reg_lock, key);
		return -EINVAL;
	}

	memset_s(&slot, sizeof(slot), 0, sizeof(slot));
	slot.node_id = node->dw & IPC4_NODE_ID_MASK;
	mailbox_sw_regs_write(offset, &slot, sizeof(slot));

	k_spin_unlock(&sof_get()->fw_reg_lock, key);
````

File: src/platform/intel/ace/include/ace/lib/mailbox.h (hunk 1 above)
Comment (lyakh) [reply src/platform/intel/ace/include/ace/lib/mailbox.h:88]: > dcache writeback is not enough, dcache_invalidate is needed on the reader core.

This would be a logical explanation, yes. Can we add such an invalidation on the reader side? But in principle - yes, a cached write plus a writeback shouldn't be faster than just a plain simple uncached write. We'd only benefit from cached access there if the same core were accessing that data multiple times without dropping caches - reading or writing (without writing back) data there. But I doubt that we ever do that.

Comment (kv2019i) [reply src/platform/intel/ace/include/ace/lib/mailbox.h:88]: The reg reads seem to use uncached addresses, so these seem correct. Nevertheless, i don't think the writeback makes given the range of addresses may not be cache aligned, so best to change the register write routines to use uncached addresses.

Comment (kv2019i) [reply src/platform/intel/ace/include/ace/lib/mailbox.h:88]: @lyakh these are "sw regs", so I guess it can be whatever sw/fw agree upon. Do you think the writeback is safe here?

Comment (lyakh) [reply src/platform/intel/ace/include/ace/lib/mailbox.h:88]: > @lyakh these are "sw regs", so I guess it can be whatever sw/fw agree upon. Do you think the writeback is safe here?

@kv2019i my understanding of "**software** registers" is that they're just generic locations in RAM used purely by the software without any direct hardware effects? If that's correct, then at the physical level they should also be handled like ordinary RAM, i.e. all cache management applies?

File: src/platform/intel/ace/include/ace/lib/mailbox.h [hunk 2]
```diff
@@ -80,12 +80,12 @@ static inline uint64_t mailbox_sw_reg_read64(size_t offset)
 
 static inline void mailbox_sw_regs_write(size_t offset, const void *src, size_t bytes)
 {
-	int regs_write_err __unused = memcpy_s((void *)(MAILBOX_SW_REG_BASE + offset),
-					       MAILBOX_SW_REG_SIZE - offset, src, bytes);
+	uint32_t __sparse_cache *ptr_c;
+	volatile uint32_t *ptr;
```
lgirdwood, line 84: volatile inserts a memory barrier when ptr is dereferenced, but in this case I think its compiled out with the call to ```memcpy_s()```.

src/platform/intel/ace/include/ace/lib/mailbox.h
lgirdwood: My thoughs here are 
1) cache ops here are not being aligned or forced on cache line boundaries or sizes.
2) No obvious locking, so can race (and corrupt with 1 above).
3) The uncache alias is probably fixing this issue as the HW IO arbitration is probably enforcing serialised IO on the memory for all cores (i.e. the HW bus arbitration is a HW spinlock).

lgirdwood (commented):
@RanderWang can you remove the volatile and use the uncache alias only - does this help ?

src/platform/intel/ace/include/ace/lib/mailbox.h
RanderWang: > 3\. The



> My thoughs here are
> 
>     1. cache ops here are not being aligned or forced on cache line boundaries or sizes.

sw_regs_write may be no cache aligned since the llp slot size is 20 bytes

>     2. No obvious locking, so can race (and corrupt with 1 above).

we have lock to protect sw_regs_write

>     3. The uncache alias is probably fixing this issue as the HW IO arbitration is probably enforcing serialised IO on the memory for all cores (i.e. the HW bus arbitration is a HW spinlock).

src/platform/intel/ace/include/ace/lib/mailbox.h
RanderWang: thanks

RanderWang:
removed volatile , thanks!

Tested on TGL & MTL

RanderWang:
It may be could improve https://github.com/thesofproject/sof/issues/8056 since this PR also fixes dai module initialization failure.

kv2019i (approved):

lgirdwood (approved):

PR #8399, opened 2023-10-26
Update build system for native loadable modules

softwarecki:
- Changed the compiler used to build loadable modules to clang. 
- Update path to rimage include directory
- Added a necessary symbols in the linker script which allows use a memory allocation functions from libc (malloc, calloc, etc.).
- Changed linker options to enable linking with libgcc and libc.
- Suggested rimage to ignore lack of the .bss section in loadable modules.
- Removed alignment of .rodata section,
- Changed .module section flags using objcopy to ignore it by rimage


pjdobrowolski (approved):

lmdk/cmake/ldscripts/bss_linker_script.txt
```diff
@@ -16,5 +16,18 @@ SECTIONS {
         _next_end = ABSOLUTE(.);
         *(.bss)
         *(.bss.*)
+
+        /* Heap memory */
+        . = ALIGN(4);
+        __stack = .;
```
lyakh, line 22: why stack?

lmdk/cmake/ldscripts/bss_linker_script.txt
```diff
@@ -16,5 +16,18 @@ SECTIONS {
...
+        _end = .;
```
lyakh, line 23: are these symbols used?

lmdk/cmake/build.cmake
```diff
@@ -15,6 +15,9 @@ foreach(MODULE ${MODULES_LIST})
 
 ###	set_target_properties(${MODULE} PROPERTIES OUTPUT_NAME ${MODULE}.mod)
```
lyakh: how is dynamic memory allocation enabled by this? How do libc / libgcc implement it in Xtensa toolchains?

jxstelter (commented):
Looks ok except the .rodata alignment removal question?

lmdk/cmake/ldscripts/data_linker_script.txt
jxstelter, line 1: I see here the problem with access permission settings for both regions.
The .rodata are read only and should be configured such way, while .data are read/write. Since permission programing is page granularity this alignment was set here.

lmdk/cmake/ldscripts/bss_linker_script.txt
softwarecki: Xtensa implementation of the `_sbrk_r` use `__stack` symbol to check that a new allocated memory from a heap not overlaps with stack. Without this symbol it simply won't link successfully.

lmdk/cmake/ldscripts/bss_linker_script.txt
softwarecki: Start of heap memory is marked by the `_end` symbol.

lmdk/cmake/ldscripts/data_linker_script.txt
softwarecki: I fully agree with you. Currently the `.data` and `.rodata` sections are placed together in a single block of firmware image. From the firmware side we also have no support for readonly data section. Until we add RODATA block support to rimage and sof, separating these sections is just a waste of memory.

lmdk/cmake/build.cmake
softwarecki: Xtensa toolchain have libc with `malloc` implementation. The libsim implements `_sbrk_r` function used by `malloc`.

lyakh: So this won't work with the Zephyr toolchain? First this alone doesn't look like a good approach, secondly we used a SOF native heap implementation before, now we're using one from Zephyr, and this seems to add another one for loadable modules. At the very least this has to be well documented, but I'm not sure at all that it's an acceptable solution.

pjdobrowolski: If it would be me I would avoid any memory allocation in loadable modules. It is hard to debug, leaves many questionable scenarios and it is not so needed after all. But when smartamp creates requirements for loadable modules that is sotultion which we provide. It is not must to used.

softwarecki: Loadable libraries will not be able to allocate any memory from the sof heap. However, they can declare a private heap in the .bss section. We don't care what dynamic allocation algorithm they use to handle this heap. They can use malloc from libc, one of the allocators provided by zephyr, allocator from sof, or even create their own solution. I proposed the simplest solution by using a allocator built into the toolchain. If someone doesn't like it, they can implement their own solution. We will not add an allocator to the system service that would allocate memory from the memory area pointed by the module.

lgirdwood (commented):
We do need to keep our client module API consistent between built-in and module, having reduced API surface and limitations around resources for modules vs built-in is going to be difficult. The latter could be solved by making sure we have some resource allocation API available in the services.

lmdk/cmake/xtensa-toolchain.cmake
```diff
@@ -18,11 +18,11 @@ cmake_path(APPEND XTENSA_TOOLCHAIN_PATH "XtensaTools" OUTPUT_VARIABLE TOOLCHAIN_
 set(CROSS_COMPILE xt-)
 
 # clang or xcc
-find_program(CMAKE_C_COMPILER NAMES ${CROSS_COMPILE}xcc
+find_program(CMAKE_C_COMPILER NAMES ${CROSS_COMPILE}clang
```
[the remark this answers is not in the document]
marc-hb: > How do we align compiler with Zephyr and base FW ?

By leveraging the Zephyr build system and writing CMake code that integrates with it. The current lmdk/CMake code totally ignores it and does everything from scratch, so it's hopeless right now.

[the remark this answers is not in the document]
pjdobrowolski: I don't see any objections against integrating lmdk build system with python build scripts which we use in sof. I thought that is it better to start small and end big?

lmdk/cmake/ldscripts/data_linker_script.txt
softwarecki: I think that when we have a solution, we can add Kconfig.

lmdk/cmake/xtensa-toolchain.cmake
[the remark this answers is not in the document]
softwarecki: I prefer an iterative approach to make PRs easy to review. We are still working on loadable modules and step by step we are reaching the final solution.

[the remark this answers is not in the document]
lgirdwood: Ok, so we should warn users in the Cmakefile that today CC flags and CC name/version are not automatically aligned with the BaseFW/Zephyr kconfig, but its a TODO.  Ultimately we will have a golden Kconfig and west tag for each product that should be easy to align with.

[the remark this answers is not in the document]
marc-hb: > I prefer an iterative approach to make PRs easy to review. We are still working on loadable modules and step by step we are reaching the final solution.

Sounds great as long as the longer term direction is clearly defined, agreed and documented somewhere[*]. At the very very least with some FIXME and TODOs pointing at the longer term so:
- Small, incremental steps never go backwards or in circles. Typically: from "drive-by", ignorant contributors but not just.
- Reviewers are never confusing work in progress with the final, desired solution(s)

> Ok, so we should warn users in the Cmakefile that today CC flags and CC name/version are not automatically aligned with the BaseFW/Zephyr kconfig, but its a TODO.

Perfect example.

[*] which does NOT mean "set in stone".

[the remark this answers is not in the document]
andyross: Marc dragged me over to this one too.  FWIW, another option would be to decide on a SDK/ABI/toolchain spec for each hardware variant, and write a little glue code here and in the Zephyr app to enforce it.

It can be as simple as a section of a header that fails to build if conditions aren't met.  Almost all (but surely there are exception) ABI/compatibility-impactful build settings are visible in the preprocessor.   That is "I'm told I need to build and link on a device with this XCHAL_CORE_ID, xt-clang version 76.8.22 or higher, using the CALL0 ABI, yada yada".

Then the problem becomes the simpler one of synchronizing those header-check blobs between the projects (maybe have them both import a single tree via west, etc...).  For extra credit it could contain a hash or something that gets emitted into the object files and you could validate that it matches at load time on the device.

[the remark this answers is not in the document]
lgirdwood: So there are some good options here to align FW, modules, CC, tools - I think the 1st step is to put large comments in the Cmake and in the Kconfig as a warning as this unblocks and gets more users of the code who will also be able to contribute to alignment processes.

lgirdwood (commented):
I think we just need to Kconfig and Cmake comments (and maybe in main lmdk header), more of a reminder for developers when they debug i.e. have you manually confirmed you are using same config for base FW and modules before further debug

lmdk/cmake/xtensa-toolchain.cmake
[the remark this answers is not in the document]
pjdobrowolski: I don't get what are concerns. These modules are build to binary being loaded and being totally independent from FW.

kv2019i (approved):
The assumptions on xtensa and particular toolchain limit applicability of lmdk quite a bit, but I don't see this PR really making this aspect more/less severe, so no have no particular objections to this PR.

lmdk/cmake/xtensa-toolchain.cmake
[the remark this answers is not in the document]
marc-hb: > loaded and being totally independent from FW.

@pjdobrowolski can you explain why ABIs don't matter here?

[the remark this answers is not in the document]
pjdobrowolski: I don't say that ABIs don't matter however since this feature is developed by our team only I must say that we have it under control quite good and things like that are important and we keep them in mind. That is why we've extracted headers for modules, we've created separate cmake files and did so many attempts in that matter.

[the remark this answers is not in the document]
marc-hb: > I don't say that ABIs don't matter

Thanks for clarifying.

> since this feature is developed by our team only I must say that we have it under control quite good and things like that are important and we keep them in mind.

I understand but people, teams and assignments come and go while open-source projects tend to last much longer. As you can see here and in #8365 there is also clear interest for this beyond just a single company. ABI issues can cost weeks of work and project delays, so having subtle things "kept in mind" is clearly not enough. The very least and all we've been asking at this early stage is just some comments in the source like "Warning: ABI dragons here. Keep toolchains and flags synchronized". I don't get the resistance; now this discussion took longer to type :-)

Issue #8336, opened 2023-10-17
Writing the sections to .ri file in reversal order in rimage creation process.


saisurya-ch:
Hi,

I have observed that the order of writing the sections to .ri got reversed from the rimage [commit ID](https://github.com/thesofproject/rimage/commit/539c2b388cc54ef6e7427ea55c6a77d7636bf99b). 

cause:
At the end of the SOF build process, while finding the sections and listing the valid sections, the address of the last section is stored as the first section. Because of this, the last section is taken first when writing into the output .ri file. This reversal of sections in rimage is causing the SOF loading issues on the AMD platforms.

could you please resolve this issue.

lgirdwood:
@saisurya-ch thanks, not sure why this has not hit us on other platforms yet, it could just be the build config and output on your platform triggers it and means a boot failure. Sounds like it has potential to impact all platforms though.
@softwarecki it looks like the new parser is getting the section address wrong for the last ELF section and writing them in reverse order. Can you fix @mwasko fyi

softwarecki:
@saisurya-ch Can you provide some more information about this issue which will allow me to reproduce it? What platform is affected? Can you share the input files for rimage? Can you paste output from rimage?

lgirdwood:
@marc-hb fyi

marc-hb:
> @saisurya-ch Can you provide some more information about this issue which will allow me to reproduce it?

@saisurya-ch do not discard the entire bug report template
https://github.com/thesofproject/sof/issues/new?assignees=&labels=bug&projects=&template=bug_report.md&title=%5BBUG%5D
Specific reproduction steps and sample data are required.

saisurya-ch:
Bug:
I have observed that the order of writing the sections to .ri got reversed from the rimage [commit ID](https://github.com/thesofproject/rimage/commit/539c2b388cc54ef6e7427ea55c6a77d7636bf99b).

To Reproduce:
Clone the latest SOF recursively. After adding the GCC toolchain path to the PATH variable, try building for the AMD/rembrandt platform using the command "./scripts/xtensa-build-all.sh rmb". (rmb platform is taken as an example here.)

Reproduction Rate:
The issue gets reproduced every time when the latest SOF is built.

Expected behavior:
The order of the elf sections should not be changed while writing them to the .ri file. They need to be written to the output ri file in the same order they were read.

Impact:
This causes improper SOF loading on all AMD platforms. (I haven't tried on other platforms.)

Environment:
[...]

PR #8401, opened 2023-10-26
rimage: module: Fix section order in output image

softwarecki:
Fixed the order in which sections are placed in the output firmware image for platforms using a simple manifest.

Fixes: #8336

### DIFF
--- a/tools/rimage/src/include/rimage/module.h
+++ b/tools/rimage/src/include/rimage/module.h
@@ -63,8 +63,9 @@ struct module_sections_info {
 	/* sections count */
 	unsigned int count;
 
-	/* First section */
+	/* sections list */
 	struct module_section *first_section;
+	struct module_section *last_section;
 };
 
 /*
--- a/tools/rimage/src/module.c
+++ b/tools/rimage/src/module.c
@@ -218,22 +218,31 @@ static void sections_info_init(struct module_sections_info *info)
  * Adds section to module_sections_info structure
  * 
  * @param info Pointer to a module_sections_info structure
- * @param address section address
- * @param size section size
+ * @param section module_section structure
  */
-static void sections_info_add(struct module_sections_info *info, const uint32_t address,
-			      const size_t size)
+static void sections_info_add(struct module_sections_info *info, struct module_section *section)
 {
-	const uint32_t end = address + size;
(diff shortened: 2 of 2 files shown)


kv2019i (approved):

kv2019i:
@saisurya-ch can you verify?

marc-hb:
@softwarecki can you explain why this didn't affect Intel? Because we don't use a "simple" (?) manifest?

btian1 (approved):

pjdobrowolski (approved):

saisurya-ch (approved):
.ri file is written in the correct order as expected.

kv2019i:
This would be ready to go, but we need a clean run of the mandatory CI checks. @softwarecki  can you check?

softwarecki:
> @softwarecki can you explain why this didn't affect Intel? Because we don't use a "simple" (?) manifest?

It appears that Intel platforms use functions from the `manifest.c` file to generate a firmware image. They work by copying content of sections to allocated memory block, taking into account the starting address of each section. The mentioned amd platforms apparently use functions from `file_simple.c`. That's why I described it as a "simple" manifest. They take sections from a input file and write their content to a output file. This is actually a simpler approach.

marc-hb:
@kv2019i , @lgirdwood , based on @softwarecki's description of `file_simple.c` and `manifest.c` we won't need to backport this to rimage/stable-v2.7, agreed? Or should we do it anyway just to be on the safe side?

lgirdwood:
> @kv2019i , @lgirdwood , based on @softwarecki's description of `file_simple.c` and `manifest.c` we won't need to backport this to rimage/stable-v2.7, agreed? Or should we do it anyway just to be on the safe side?

Not impacting v2.7.1, but can be added to v2.7.2 if we need to.

lgirdwood (approved):

kv2019i:
@wszypelt @softwarecki The fails look unrelated to this PR, can you take a lot or give go-ahead to merge as this is urgent for non-Intel configurations.

wszypelt:
> @wszypelt @softwarecki The fails look unrelated to this PR, can you take a lot or give go-ahead to merge as this is urgent for non-Intel configurations.

all green in Internal Intel CI System :)

PR #8403, opened 2023-10-26
topology2: google-rtc-aec: adjust the IBS/OBS size based on process needs

yongzhi1:
Add new token to adjust the input/output buffer size for process modules.

### DIFF
--- a/tools/topology/topology2/include/components/google-rtc-aec.conf
+++ b/tools/topology/topology2/include/components/google-rtc-aec.conf
@@ -14,6 +14,8 @@
 # Where N is the unique instance number for the google-rtc-aec object within the same alsaconf node.
 Define {
 	AEC_UUID	"a6:a0:80:b7:9f:26:6f:46:b4:77:23:df:a0:5a:f7:58"
+	# AEC requires 10ms buffers for processing
+	BUFFER_PERIOD_MS	10
 }
 
 Class.Widget."google-rtc-aec" {
@@ -96,6 +98,7 @@ Class.Widget."google-rtc-aec" {
 			input_pin_index		0
 			in_bit_depth            16
 			in_valid_bit_depth      16
+			ibs "$[(($in_channels * ($[($in_rate + 999)] / 1000)) * ($in_bit_depth / 8)) * $BUFFER_PERIOD_MS ]"
 		}
 		# 4ch 16-bit on Pin 0
 		{
@@ -105,12 +108,14 @@ Class.Widget."google-rtc-aec" {
 			in_channels		4
 			in_ch_cfg		$CHANNEL_CONFIG_3_POINT_1
 			in_ch_map		$CHANNEL_MAP_3_POINT_1
+			ibs "$[(($in_channels * ($[($in_rate + 999)] / 1000)) * ($in_bit_depth / 8)) * $BUFFER_PERIOD_MS]"
 		}
 		# 2ch 16-bit on Pin 1
 		{
 			input_pin_index		1
 			in_bit_depth            16
 			in_valid_bit_depth      16
+			ibs "$[(($in_channels * ($[($in_rate + 999)] / 1000)) * ($in_bit_depth / 8)) * $BUFFER_PERIOD_MS]"
 		}
 	]
 
@@ -118,6 +123,8 @@ Class.Widget."google-rtc-aec" {
 		{
 			out_bit_depth           16
 			out_valid_bit_depth     16
+			obs "$[(($out_channels * ($[($out_rate + 999)] / 1000)) * ($out_bit_depth / 8)) * $BUFFER_PERIOD_MS]"
+
 		}
 		# 4ch 16-bit
 		{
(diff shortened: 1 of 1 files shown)


yongzhi1:
This is written by @ranj063 , posted for review upon her request.

tools/topology/topology2/include/common/tokens.conf
```diff
@@ -181,4 +181,8 @@ Object.Base.VendorToken {
 		node_type		1980
 		deep_buffer_dma_ms	1981
 	}
+	"19" {
+		name "process"
```
btian1, line 185: seems currently, you only want this used for RTC-AEC, is it possible to make it as a common attribute?
CPC calculation need set different input/output buffer size, so maybe we can start from here.

RanderWang:
please check my comments in Linux kernel side, thanks!

tools/topology/topology2/include/components/google-rtc-aec.conf
```diff
@@ -138,4 +143,6 @@ Class.Widget."google-rtc-aec" {
 	is_pages		1
         num_input_pins          2
         num_output_pins         1
+        # AEC requires 10ms buffers for processing
+        buffer_period_size	10
```
lgirdwood, line 147: Should this be buffer_period_ms ?

tools/topology/topology2/include/common/tokens.conf
yongzhi1: Rander suggested to hide IBS/OBS adjustments inside the tplg instead of asking driver to run the math, which is fine.

tools/topology/topology2/include/components/google-rtc-aec.conf
yongzhi1: I can use BUFFER_PERIOD_MS in the 2nd version.

yongzhi1:
V2 Update:

Taka a different approach to update buffer_size with IBS/OBS adjusted in tplg.

tools/topology/topology2/include/components/google-rtc-aec.conf
```diff
@@ -96,6 +98,7 @@ Class.Widget."google-rtc-aec" {
 			input_pin_index		0
 			in_bit_depth            16
 			in_valid_bit_depth      16
+			ibs "$[(($in_channels * ($[($in_rate + 999)] / 1000)) * ($in_bit_depth / 8)) * $M ]"
```
btian1, line 101: @lgirdwood @singalsu @ranj063  @marcinszkudlinski , we may need unify these ibs/obs setting.
currently, when we try to set ibs/obs, we always try to do some rounding, like "+999" or "+4", based on discuss in:
https://github.com/thesofproject/sof/pull/8352/commits/77a12d548c6823407152baf36b6f638c2a61b2ee

from DP perspective, those rounding is not necessary, i, e we should use:
ibs "$[(($in_channels * ($[($in_rate] / 1000)) * ($in_bit_depth / 8)) * $M ]"

this can reduce delay for pipeline, if with rounding, take 11025 as example, it need 12 samples, if current is 11, it have to wait until next come to meet 12 samples, then if set ibs as 11 based, when input 11 samples arrived, it can go without waiting.

btian1 (commented):
BTW, may I know this IBS setting is for AEC module intermediate buffer or AEC's source buffer size?

tools/topology/topology2/include/components/google-rtc-aec.conf
```diff
@@ -14,6 +14,8 @@
 # Where N is the unique instance number for the google-rtc-aec object within the same alsaconf node.
 Define {
 	AEC_UUID	"a6:a0:80:b7:9f:26:6f:46:b4:77:23:df:a0:5a:f7:58"
+	# AEC requires 10ms buffers for processing
+	M		10
```
RanderWang, line 18: M is too simple, is it duration or period?

RanderWang (commented):
please change the PR title. It look good for me

tools/topology/topology2/include/components/google-rtc-aec.conf
yongzhi1: How about BUFFER_PERIOD_MS?

tools/topology/topology2/include/components/google-rtc-aec.conf
yongzhi1: Ack, thanks for the inputs, can we make a cross-tree changes with the improved formula in a seperate patch?

yongzhi1:
> BTW, may I know this IBS setting is for AEC module intermediate buffer or AEC's source buffer size?

This should be the source buffer size according to the definition found here
https://github.com/thesofproject/sof-docs/blob/63e3905c7b376f11ab828ac4246f0dc3e0d1bc6d/developer_guides/rimage/ipc4_extended_module_config.rst#module-config

RanderWang (approved):

RDharageswari (approved):

kv2019i (approved):

PR #8404, opened 2023-10-26
[DNM] Test Zephyr with SSP RX FIFO drain on startup change

ujfalusi:
Only drain the FIFO on startup if it is in overflow condition.


ujfalusi:
FYI: @kv2019i

PR #8412, opened 2023-10-30
Src tplg fix

btian1:


lgirdwood (commented):

andrula-song (approved):

tools/topology/topology2/include/components/src_format_s32_convert_from_48k.conf
```diff
@@ -27,7 +27,7 @@
 					]
 					out_bit_depth [ 32 ]
 					out_valid_bit_depth [ 32 ]
-					obs "$[($out_channels * (($[($out_rate + 999)] / 1000) + 4)) * ($out_bit_depth / 8)]"
+					obs "$[$out_channels * (($out_rate / 1000) + 1) * ($out_bit_depth / 8)]"
```
kv2019i, line 30: I think this would benefit from more explanation and examples with some values. So the base class version of obs calculation is not enough (output_audio_format.conf), but why is 1 extra word enough in this case. I think a similar documentation note as in output_audio_format.conf would be in order here.

kv2019i (changes requested):

tools/topology/topology2/include/components/src_format_s32_convert_from_48k.conf
btian1: I double checked with Seppo, current +1 is for ASRC, somehow mixed with ASRC and SRC, for SRC, the calculation is quite complex, SRC converter is composed with two set of polyphase filter, it is 21/20 and 7/8, which means 20 --> 21, the second is 8 --> samples, Seppo helped generated below table to get better understanding for the max obs:
image

we should set obs as 56 for LL pipe, for other cases, we need more parameter for set, like deep buffer.
Let's keep implementation for LL for now

singalsu: @btian1 Have you been able to run e.g. the SRC capture and playback in sof-nocodec topology with clean audio with this patched topology with any of these impacted rates?

singalsu: Can't listen audio in real time with nocodec, but the topology is arranged such that you can capture from PCM the playback to PCM via SSP loopback. Launch arecord from one shell window and aplay from other. Then copy the recording to you PC and listen if it's clean and with correct rate/duration.

singalsu (approved):
I ran a test with SRCs in sof-tgl-nocodec.tplg. All supported rates combinations worked OK without glitches seen in  "host -> SRC -> DAI (48 kHz) -> SRC -> host" processing chain.

lgirdwood (approved):

lgirdwood:
@btian1 can you check IPC3 build test result, it looks like not related to this PR. Thanks

btian1:
> @btian1 can you check IPC3 build test result, it looks like not related to this PR. Thanks

it is smart amp error, should already be fixed with latest, let me force push again to get pass result.

btian1:
SOFCI TEST

tools/topology/topology2/include/components/src_format_s32_convert_from_48k.conf
```diff
@@ -27,7 +27,8 @@
 					]
 					out_bit_depth [ 32 ]
 					out_valid_bit_depth [ 32 ]
-					obs "$[($out_channels * (($[($out_rate + 999)] / 1000) + 4)) * ($out_bit_depth / 8)]"
+					# actual buffer size in sof is obs * 2, so here simply set to its period size
+					obs "$[$out_channels * ($out_bit_depth / 8) * (($out_rate / 1000)) * ($period / 1000)]"
```
kv2019i, line 31: @btian1 But isn't this the same definition we have in the baseline? So if this doesn't require custom obs logic, we can remove this altogether?

kv2019i (approved):
Not sure why we need to override "obs" anymore if there is no change to baseclass definition, but the comment might be useful to remind why  extra space is not required.
I'm not 100% sure how wise it is to rely on 2*obs definition. To me, if we have a field for output block size, modules should specify what they actually need, and not assume that the user of this information will anyways double the allocation. But seems others are happy, so let's go with this and leave the comment in place.

tools/topology/topology2/include/components/src_format_s32_convert_from_48k.conf
btian1: obs "$[($out_channels * ($[($out_rate + 999)] / 1000)) * ($out_bit_depth / 8)]"

this adding period into count, however, this may change further,  a lot of related changes are happening.
let's wait and see.

PR #7345, opened 2023-03-29
dai-zephyr: refactor dai params

juimonen:
Refactor dai params into more feasible functions and remove duplicate code. No functional change.

PR #8415, opened 2023-10-30
dai-zephyr: refactor dai params

RanderWang:
Refactor dai params into more feasible functions and remove duplicate code. No functional change.

The original pr: https://github.com/thesofproject/sof/pull/7345 was done by Jaska but was not merged. Since he has leaved intel, I will help to rebase it and submit it again .

### DIFF
--- a/src/audio/dai-zephyr.c
+++ b/src/audio/dai-zephyr.c
@@ -565,173 +565,32 @@ static int dai_verify_params(struct dai_data *dd, struct comp_dev *dev,
 	return 0;
 }
 
-/* set component audio SSP and DMA configuration */
-static int dai_playback_params(struct dai_data *dd, struct comp_dev *dev, uint32_t period_bytes,
-			       int32_t period_count)
+static int dai_set_sg_config(struct dai_data *dd, struct comp_dev *dev, uint32_t period_bytes,
+			     uint32_t period_count)
 {
 	struct dma_sg_config *config = &dd->config;
-	struct dma_config *dma_cfg;
-	struct dma_block_config *dma_block_cfg;
-	struct dma_block_config *prev = NULL;
 	uint32_t local_fmt = audio_stream_get_frm_fmt(&dd->local_buffer->stream);
 	uint32_t dma_fmt = audio_stream_get_frm_fmt(&dd->dma_buffer->stream);
 	uint32_t fifo, max_block_count, buf_size;
-	int i, err = 0;
+	int err = 0;
 
-	/* set processing function */
-	dd->process = pcm_get_conversion_function(local_fmt, dma_fmt);
+	/* set up DMA configuration */
+	if (dev->direction == SOF_IPC_STREAM_PLAYBACK) {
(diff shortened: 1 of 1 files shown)


btian1 (commented):
the code already mismatched, if possible, I would suggest to split with several patches for better check and bisect.

RanderWang:
All tests pass!  the original PR got 3 approve. Thanks!

src/audio/dai-zephyr.c
```diff
@@ -956,17 +820,64 @@ int dai_common_params(struct dai_data *dd, struct comp_dev *dev,
...
+		comp_err(dev, "dai_zephyr_params(): set dma config failed.");
+out:
+	/*
+	 * Make sure to free all allocated items, all functions
+	 * can be called with null pointer.
+	 */
+	if (err < 0) {
+		buffer_free(dd->dma_buffer);
+		dma_sg_free(&config->elem_array);
+		rfree(dd->z_config);
```
lyakh, line 877: if I understand it correctly, `dd` is persistent, isn't it? Here in error case we free several buffers, but we keep pointers to them. Should we reset them all to `NULL`? Except `config->elem_array.elems` - it's already reset inside `dma_sg_free()`

RanderWang: dd is allocated by dai_new and released in dai_free. If dai_common_params is failed,  dai_reset will be called and these pointer will be set to null
````c
void dai_common_reset(struct dai_data *dd, struct comp_dev *dev)
{
	struct dma_sg_config *config = &dd->config;

	dma_sg_free(&config->elem_array);
	if (dd->z_config) {
		rfree(dd->z_config->head_block);
		rfree(dd->z_config);
		dd->z_config = NULL;
	}

	if (dd->dma_buffer) {
		buffer_free(dd->dma_buffer);
		dd->dma_buffer = NULL;
	}
}
````

kv2019i (approved):

btian1:
> > ggest to split with several patches for better check and bisect.
> 
> NO, it is based on latest code

mismatch means code are mismatched, hard to review and revert, after this patch merged, and then following dai code changed continue, once you want to revert this patch due to a regression bug based on latest sof code, it will be hard, have to use reset.

if possible, split will make easy to read/review/maintain.

src/audio/dai-zephyr.c
lyakh: @RanderWang that's even worse then. That would mean that we free the same pointers twice, which is wrong and can lead to run-time errors

lyakh (changes requested):
double free is a bug

src/audio/dai-zephyr.c
RanderWang: o, thanks!

RanderWang:
fixed the double free, thanks!

lyakh (approved):

kv2019i:
The smart-amp stub build fail is unrelated, proceeding with merge.

PR #8416, opened 2023-10-30
Asrc module convert

andrula-song:
convert ASRC to module interface with IPC4.

### DIFF
--- a/src/audio/asrc/asrc.c
+++ b/src/audio/asrc/asrc.c
@@ -31,6 +31,7 @@
 #include <stdint.h>
 
 #if CONFIG_IPC_MAJOR_4
+#include <sof/audio/module_adapter/module/generic.h>
 #include <ipc4/base-config.h>
 #include <ipc4/asrc.h>
 #endif
@@ -645,6 +646,56 @@ static int asrc_dai_find(struct comp_dev *dev, struct comp_data *cd)
 	return 0;
 }
 
+#if CONFIG_IPC_MAJOR_4
+static int asrc_dai_configure_timestamp(struct comp_data *cd)
+{
+	if (cd->dai_dev) {
+		struct processing_module *mod = comp_get_drvdata(cd->dai_dev);
+		struct module_data *md = &mod->priv;
+
+		return md->ops->endpoint_ops->dai_ts_config(cd->dai_dev);
+	}
+
+	return -EINVAL;
+}
+
+static int asrc_dai_start_timestamp(struct comp_data *cd)
+{
+	if (cd->dai_dev) {
+		struct processing_module *mod = comp_get_drvdata(cd->dai_dev);
+		struct module_data *md = &mod->priv;
+
+		return md->ops->endpoint_ops->dai_ts_start(cd->dai_dev);
+	}
+
+	return -EINVAL;
+}
+
+static int asrc_dai_stop_timestamp(struct comp_data *cd)
+{
+	if (cd->dai_dev) {
+		struct processing_module *mod = comp_get_drvdata(cd->dai_dev);
+		struct module_data *md = &mod->priv;
+
+		return md->ops->endpoint_ops->dai_ts_stop(cd->dai_dev);
+	}
+
+	return -EINVAL;
+}
+
+static int asrc_dai_get_timestamp(struct comp_data *cd,
+				  struct dai_ts_data *tsd)
+{
+	if (cd->dai_dev) {
+		struct processing_module *mod = comp_get_drvdata(cd->dai_dev);
+		struct module_data *md = &mod->priv;
+
+		return md->ops->endpoint_ops->dai_ts_get(cd->dai_dev, tsd);
+	}
+
+	return -EINVAL;
+}
+#else
 static int asrc_dai_configure_timestamp(struct comp_data *cd)
 {
 	if (cd->dai_dev)
@@ -669,17 +720,14 @@ static int asrc_dai_stop_timestamp(struct comp_data *cd)
 	return -EINVAL;
 }
 
-#if CONFIG_ZEPHYR_NATIVE_DRIVERS
-	static int asrc_dai_get_timestamp(struct comp_data *cd, struct dai_ts_data *tsd)
-#else
-	static int asrc_dai_get_timestamp(struct comp_data *cd, struct timestamp_data *tsd)
-#endif
+static int asrc_dai_get_timestamp(struct comp_data *cd, struct timestamp_data *tsd)
 {
 	if (!cd->dai_dev)
 		return -EINVAL;
 
 	return cd->dai_dev->drv->ops.dai_ts_get(cd->dai_dev, tsd);
 }
+#endif
 
 static int asrc_trigger(struct comp_dev *dev, int cmd)
 {
--- a/src/audio/asrc/asrc.c
+++ b/src/audio/asrc/asrc.c
@@ -3,6 +3,7 @@
 // Copyright(c) 2019-2022 Intel Corporation. All rights reserved.
 
 #include <sof/audio/asrc/asrc_farrow.h>
+#include <sof/audio/module_adapter/module/generic.h>
 #include <sof/audio/buffer.h>
 #include <sof/audio/component.h>
 #include <sof/audio/format.h>
@@ -31,7 +32,6 @@
 #include <stdint.h>
 
 #if CONFIG_IPC_MAJOR_4
-#include <sof/audio/module_adapter/module/generic.h>
 #include <ipc4/base-config.h>
 #include <ipc4/asrc.h>
 #endif
@@ -48,13 +48,11 @@
 #define COEF_C1		Q_CONVERT_FLOAT(0.01, 30)
 #define COEF_C2		Q_CONVERT_FLOAT(0.99, 30)
 
-typedef void (*asrc_proc_func)(struct comp_dev *dev,
+typedef void (*asrc_proc_func)(struct processing_module *mod,
 			       const struct audio_stream *source,
 			       struct audio_stream *sink,
 			       int *consumed,
 			       int *produced);
-
-static const struct comp_driver comp_asrc;
 
 LOG_MODULE_REGISTER(asrc, CONFIG_SOF_LOG_LEVEL);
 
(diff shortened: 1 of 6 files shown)


src/audio/asrc/asrc_ipc3.c
```diff
@@ -0,0 +1,1027 @@
+// SPDX-License-Identifier: BSD-3-Clause
```
lyakh, line 1: can we extract common code?

andrula-song: is there any else to extract? Thanks for pointing out in detail.

src/audio/asrc/asrc_ipc4.c
```diff
@@ -33,7 +33,6 @@
 #include <ipc4/base-config.h>
```
singalsu: It would be simpler to convert both ipc3 and ipc4 versions to module adapter. You can build and run the ipc3 version in testbench. The difference in them is very small.

singalsu (changes requested):

src/audio/asrc/asrc_ipc4.c
singalsu: The ALSA controls handling is the biggest difference in ipc3 and ipc4 but this component does not have them.

src/audio/asrc/asrc_ipc3.c
btian1: @lyakh , we can use another PR to do the extraction and ipc3/ipc4 isolation, this PR can only focus on module convertion.

src/audio/asrc/asrc_ipc4.c
btian1: I also suggest this PR only focus on module converter, then do ipc3/ipc4 isolation.

src/audio/asrc/asrc_ipc3.c
lyakh: > @lyakh , we can use another PR to do the extraction and ipc3/ipc4 isolation, this PR can only focus on module convertion.

@btian1 sorry, I don't think it's a good approach to first duplicate the code to then merge it again

src/audio/asrc/asrc_ipc4.c
andrula-song: @singalsu  thanks for telling how to test IPC3 version, I will do the IPC3 version.

src/audio/asrc/asrc_common.c
```diff
@@ -0,0 +1,799 @@
...
+	struct comp_dev *dev_dai;
+	int pid;
+
+	/* Get current pipeline ID and walk to find the DAI */
+	pid = dev_comp_pipe_id(dev);
+	cd->dai_dev = NULL;
+	if (cd->mode == ASRC_OM_PUSH) {
+		/* In push mode check if sink component is DAI */
+		do {
+			sinkb = list_first_item(&dev->bsink_list, struct comp_buffer, source_list);
```
lyakh, line 315: am I missing something or how is this advancing with loop iterations? Wouldn't it loop permanently if the first iteration doesn't find the DAI?

src/audio/asrc/asrc_common.c
```diff
@@ -0,0 +1,799 @@
...
+			}
+
+		} while (dev_comp_type(dev_dai) != SOF_COMP_DAI);
+	} else {
+		/* In pull mode check if source component is DAI */
+		do {
+			sourceb = list_first_item(&dev->bsource_list, struct comp_buffer,
+						  sink_list);
+
+			dev_dai = sourceb->source;
```
lyakh, line 336: ditto

src/audio/asrc/asrc_common.c
```diff
@@ -0,0 +1,799 @@
...
+		} while (dev_comp_type(dev_dai) != SOF_COMP_DAI);
+	}
+
+	/* Point dai_dev to found DAI */
+	cd->dai_dev = dev_dai;
+
+	return 0;
+}
+
+#if CONFIG_IPC_MAJOR_4
```
lyakh, line 356: why do you have this in the *common* file? confused...

src/audio/asrc/asrc_common.c
```diff
@@ -0,0 +1,799 @@
...
+	if (!src_obj)
+		return;
+
+	if (src_obj->bit_depth == 32)
+		for (ch = 0; ch < src_obj->num_channels; ch++) {
+			buf_32 = src_obj->ring_buffers32[ch];
+
+			if (buf_32) {
+				rfree(buf_32);
+				src_obj->ring_buffers32[ch] = NULL;
```
lyakh, line 289: assuming no race is possible, this could be as simple as
```
for (ch = 0; ch < src_obj->num_channels; ch++) {
	rfree(src_obj->ring_buffers32[ch]);
	src_obj->ring_buffers32[ch] = NULL;
}
```
but if this is indeed trying to protect against a race, then it should first set address to `NULL` and then free memory. Otherwise unclear why these temporary variables are needed. Same below

src/audio/asrc/asrc_ipc3.c
```diff
@@ -321,92 +106,22 @@ static struct comp_dev *asrc_new(const struct comp_driver *drv,
...
 {
 	struct comp_data *cd = comp_get_drvdata(dev);
 
 	comp_info(dev, "asrc_free()");
 
-	rfree(cd->buf);
-	asrc_release_buffers(cd->asrc_obj);
-	rfree(cd->asrc_obj);
-	rfree(cd);
+	asrc_free_general(cd);
```
lyakh, line 124: `asrc_free_common()`?

src/audio/asrc/asrc_ipc3.c
```diff
@@ -493,170 +207,21 @@ static int asrc_params(struct comp_dev *dev,
...
-	static int asrc_dai_get_timestamp(struct comp_data *cd, struct dai_ts_data *tsd)
-#else
-	static int asrc_dai_get_timestamp(struct comp_data *cd, struct timestamp_data *tsd)
-#endif
-{
-	if (!cd->dai_dev)
-		return -EINVAL;
-
-	return cd->dai_dev->drv->ops.dai_ts_get(cd->dai_dev, tsd);
+	return asrc_param_general(cd, dev, &sourceb->stream, &sinkb->stream);
```
lyakh, line 210: I think "common" would be a better description for all these functions - you have several of them

src/audio/asrc/asrc_ipc3.c
```diff
@@ -978,20 +294,7 @@ static int asrc_reset(struct comp_dev *dev)
...
-	if (cd->track_drift)
-		asrc_dai_stop_timestamp(cd);
-
-	/* Free the allocations those were done in prepare() */
-	asrc_release_buffers(cd->asrc_obj);
-	rfree(cd->asrc_obj);
-	rfree(cd->buf);
-	cd->asrc_obj = NULL;
-	cd->buf = NULL;
+	asrc_reset_general(cd, dev);
```
lyakh, line 297: so first a 1000+ line long file is created, then almost 700 lines get moved out...

andrula-song: it is because I didn't have environment to test IPC3 version before, just want to keep the original code to reduce errors imported by me. Since Seppo help to enable the asrc test bench, I can convert IPC3 too. Thanks for reviewing fast, there will be a convert with both IPC3/IPC4 soon.

src/audio/module_adapter/module_adapter.c
```diff
@@ -1393,6 +1391,8 @@ int module_adapter_trigger(struct comp_dev *dev, int cmd)
 		dev->state = COMP_STATE_ACTIVE;
 		return PPL_STATUS_PATH_STOP;
 	}
+	if (md->ops->trigger)
+		return md->ops->trigger(mod, cmd);
```
btian1, line 1395: these  two lines can be a separate patch, it is not only for ASRC.

src/audio/asrc/asrc.c
```diff
@@ -1080,19 +1035,31 @@ static int asrc_copy(struct comp_dev *dev)
...
-	if (cd->source_frames && cd->sink_frames)
-		asrc_process(dev, source, sink);
+	if (cd->source_frames && cd->sink_frames) {
+		int consumed = 0;
+		int produced = 0;
+
+		/* consumed bytes are not known at this point */
+		buffer_stream_invalidate(source, audio_stream_get_size(source_s));
+		cd->asrc_func(mod, source_s, sink_s, &consumed,
+				&produced);
```
lyakh, line 1045: since you'll be splitting this commit into two: first adding `.trigger()` to module-adapter and then the actual ASRC conversion, you could also merge this into one line while at it

btian1 (approved):
@andrula-song ， please change status to open, this PR should be done now.

andrula-song:
SOFCI TEST

lyakh (approved):

src/audio/asrc/asrc.c
```diff
@@ -439,59 +416,49 @@ static void asrc_release_buffers(struct asrc_farrow *src_obj)
 			buf_32 = src_obj->ring_buffers32[ch];
 
 			if (buf_32) {
-				rfree(buf_32);
 				src_obj->ring_buffers32[ch] = NULL;
+				rfree(buf_32);
```
singalsu, line 420: Interesting, was this a memory violation fix? Please explain and move to a separate commit.

andrula-song: ah, it is because the comment from @lyakh  https://github.com/thesofproject/sof/pull/8416#discussion_r1381368050

singalsu: OK fair enough!

singalsu (approved):

src/audio/asrc/asrc.c
lyakh: @singalsu the original order keeps a theoretical race window open: while you still have a pointer to a freed memory in a globally accessible persistent object, someone might access it at that time, causing a use-after-free bug scenario. Although we don't know any such specific possibilities here, but since we already have a temporary variable with that pointer, we could as well use it to protect against that race possibility. I.e.
```
free(obj->mem);
// if someone now accesses obj->mem, it's a use-after-free bug
obj->mem = NULL;
```
that one has a race window, whereas
```
x = obj->mem;
obj->mem = NULL;
free(x);
```
this one doesn't

singalsu: Thanks for elaborating @lyakh !

Issue #8260, opened 2023-09-25
[FEATURE][Zephyr] Remove duplication of CMake rules between Zephyr and XTOS


kv2019i:
**Is your feature request related to a problem? Please describe.**
When a software module is added to SOF that can be used both in XTOs and in Zephyr builds, the module sources need to be added to CMake rules twice:
 - first in sof/src/foo/CMakeLists.txt (for XTOS), and 
 - second time to sof/zephyr/CMakeLists.txt (for Zephyr)

This split of Cmake rules dates back to the time when Zephyr support was introduced and codebase was not cleanly split between operating systems. Now we have a cleaner split at top-level source directories, so it should be possible to start cleaning up this overlap in definitions.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**
Platforms where transition to Zephyr is still in progress might cause some extra challenges to doing this makefile cleanup. E.g. see [XTOS-only a](https://github.com/thesofproject/sof/issues/7248)

- https://github.com/thesofproject/sof/issues/7248
- #8606

kv2019i:
FYI @dbaluta @lgirdwood @marc-hb @marcinszkudlinski -- not tagging to v2.8 yet. A bit more feasibility and work effort analysis is needed to gauge the effort and options. But sooner or later we need to get rid of the duplication.

lyakh:
the best would be of course if we could re-use CMakeLists.txt in subdirectories for both XTOS and Zephyr. If this is impossible, we could move XTOS specific code to xtos.cmake, add Zephyr to zephyr.cmake and put any common code in CMakeLists.txt

marc-hb:
> if we could re-use CMakeLists.txt in subdirectories for both XTOS and Zephyr.

I noticed `sof/zephyr/CMakeLists.txt` uses `zephyr_sources()`, not the plain `target_sources()`. I'm not sure why. I found these:
- https://github.com/zephyrproject-rtos/zephyr/issues/8439
- https://github.com/zephyrproject-rtos/zephyr/discussions/43650

`zephyr/samples/**/CMakeLists.txt` and `zephyr/tests/**/CMakeLists.txt` use `target_sources()`.

lgirdwood:
Can we drop target_sources everywhere and use zephyr sources e.g. somethimg like for xtos
```
#define zephyr_sources(x) target_sources(x)
```

marc-hb:
> I agree this location is inconvenient. Let me think about it.

[...]

PR #8422, opened 2023-10-31
Start decentralizing zephyr/CMakeLists.txt

marc-hb:
2 commits.

Should help with:
- #8260

### DIFF
--- a/src/init/CMakeLists.txt
+++ b/src/init/CMakeLists.txt
@@ -1,5 +1,18 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
+is_zephyr(it_is)
+if(it_is)  ### Zephyr ###
+
+
+zephyr_library_sources(
+	init.c
+	ext_manifest.c
+)
+
+
+else()  ### Not Zephyr ###
+
+
 add_local_sources(sof init.c)
 
 add_library(ext_manifest STATIC "")
@@ -21,3 +34,6 @@ sof_append_relative_path_definitions(ext_manifest)
 
 target_link_libraries(ext_manifest sof_options)
 target_link_libraries(sof_static_libraries INTERFACE ext_manifest)
+
+
+endif() # Zephyr
--- a/src/ipc/CMakeLists.txt
+++ b/src/ipc/CMakeLists.txt
@@ -6,9 +6,31 @@ elseif (CONFIG_IPC_MAJOR_4)
         add_subdirectory(ipc4)
 endif()
 
-add_local_sources(sof
+set(base_files
 	ipc-common.c
 	ipc-helper.c
+)
+
+is_zephyr(it_is)
+if(it_is) ###  Zephyr ###
+
+zephyr_library_sources(
+	${base_files}
+	dma-copy.c
+)
+
+if (CONFIG_SOC_SERIES_INTEL_CAVS_V25 OR CONFIG_SOC_SERIES_INTEL_ACE)
+	zephyr_library_sources(
+		ipc-zephyr.c
+	)
+endif()
+
+
+else() ###  Not Zephyr ###
+
+
+add_local_sources(sof
+	${base_files}
 )
 
 if (CONFIG_TRACE)
@@ -19,3 +41,6 @@ endif()
 if (CONFIG_LIBRARY)
 	return()
 endif()
+
+
+endif() # Zephyr
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -128,6 +128,8 @@ macro(is_zephyr ret)
 endmacro()
 
 add_subdirectory(../src/init/ init_unused_install/)
+add_subdirectory(../src/ipc/  ipc_unused_install/)
+
 
 
 # Old way below: all .c files added by this giant CMake file.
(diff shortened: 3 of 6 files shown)


lgirdwood (approved):
Nice start !

src/init/CMakeLists.txt
```diff
@@ -1,5 +1,17 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
+if(CONFIG_SOF)  ### Zephyr ###
+
+
+zephyr_library_sources(
```
andyross, line 6: This is a spot where Zephyr's promiscuous cmake layer hurts pretty badly, because you can't add files to the project in the same way in both modes.  Again, maybe work around this by adding yet another layer of indirection and have a "sof_sources()".

Also, FWIW:  the "_library_sources" variant really doesn't do anything useful beyond make the build slower by taking the time to create a static library.  There is (AFAIK) exactly one spot in Zephyr where we have to do this for obscure linkage reasons[1], but for general C code it's almost always faster and cleaner to link all the object files at once.

I've tried to make that point upstream and basically failed: "zephyr_library_sources()" is a pervasive paradigm.  But really it's needless except in specific situations.

(For reference: the reason to use it is that a static library only presents internal files to the link if one of their symbols is accessed, so you can build it and include static stuff like e.g. SYS_INIT() or interrupt declarations, and avoid having those end up in the final output unless something is using the file.  But DTS and kconfig are the right mechanisms for that kind of dependency tracking, doing it with the linker is dangerous and surprising.)

andyross (commented):
Very much needed.  The existing Zephyr kconfig is indeed a giant merge collision factory as it stands.  Some drive-by notes.

src/init/CMakeLists.txt
```diff
@@ -1,5 +1,17 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
+if(CONFIG_SOF)  ### Zephyr ###
```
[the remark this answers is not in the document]
andyross: If you do it in the upstream zephyr module, you'll be before any of the cmake code in SOF proper for sure.  But yeah, if there's a clearer kconfig that's probably even better (I'm not much of an expert on the module layer, but if it's exposed in kconfig it's surely valid to rely on).

src/init/CMakeLists.txt
andyross: Heh, I've spent my bikeshed quota for the month already.

The way this goes is that someone says "libraries make linking faster!" and the response is "only after the ranlib stage builds the table, and in zephyr most symbols are actually needed by the final link", then someone else demands benchmarks, etc...

Really I was mostly just whining about a personal peeve.  I've made my peace with it: per-directory/per-app fine grained static libraries are the way zephyr links itself.

zephyr/CMakeLists.txt
```diff
@@ -114,14 +114,30 @@ zephyr_include_directories(
 # SOC level sources
 # Files that are commented may not be needed.
 
+
+# New, "de-centralized Zephyr" way. Requires "is_zephyr()" conditionals in
+# the decentralized CMakeLists.txt files shared with XTOS.
+
+# Duplicate in sof/scripts/misc.cmake; keep in sync
+macro(is_zephyr ret)
+	if(CONFIG_ZEPHYR_SOF_MODULE STREQUAL "y")
```
andyross, line 123: Just `if(CONFIG_ZEPHYR_SOF_MODULE)`  should be fine, that's the pattern within Zephyr itself.

With kconfigs, there's a validation layer somewhere (don't ask me where) that guarantees that variables of the form CONFIG_xxx are matched affirmatively with kconfig input.  So you can assume that if such a variable exists that it reflects the same thing you'd see in C or Kconfig and evaluates to false if unset.  Pretty sure, though the details I'm sure are more complicated.

marc-hb: Done

btian1 (commented):
1. can you add details for how decentralize zephyr/CMakeList.txt in PR comments?
2. anyway, two patches share exact same comments is not good.
3. is this all for decentralize? or just a beginning?

marc-hb:
> can you add details for how decentralize zephyr/CMakeList.txt in PR comments?

I don't understand the question, can you be more specific? This PR is just moving CMake code from one giant file to multiple CMake files in multiple subdirectories. What about this move don't you understand?

> anyway, two patches share exact same comments is not good.

The two commits do the same thing in two different directories. So the only difference between the two commit messages is the directory name.

> is this all for decentralize? or just a beginning?

Did you look at the title of the PR?

lgirdwood:
@andyross good for you as a starting point to merge, we can factor in more review comments incrementally ?

andyross (approved):
Sorry, realized I never had a +1 on this

src/ipc/CMakeLists.txt
```diff
@@ -6,11 +6,33 @@ elseif (CONFIG_IPC_MAJOR_4)
...
 
+is_zephyr(it_is)
+if(it_is) ###  Zephyr ###
+
+zephyr_library_sources(
+	${base_files}
+	dma-copy.c
+)
+
+if (CONFIG_SOC_SERIES_INTEL_CAVS_V25 OR CONFIG_SOC_SERIES_INTEL_ACE)
```
kv2019i, line 22: This looks strange, but it matches the existing logic, so great to have this become more visible in the cmake code. The long per-product sections in one-giant-zephyr-cmakefile easily hide these things (when we have per-platform compilation of stuff that really shouldn't be per-platform).

kv2019i (approved):
Thanks @marc-hb . This is a great start.

marc-hb:
I fixed the unrelated `zmain` failures in  https://github.com/thesofproject/sof/actions/runs/6725126747/job/18278872844?pr=8422 with unrelated:
- https://github.com/zephyrproject-rtos/zephyr/pull/64712

Not force-pushing not to lose existing approvals :-)

PR #8423, opened 2023-10-31
[DNM] intel_adsp: ace: dynamic clock switching

tmleman:
This pull request introduces a series of enhancements to the power management capabilities of the Intel ADSP platforms. The changes include the implementation of dynamic clock switching, which allows the DSP to adjust its clock frequency based on the idle state of the cores, thereby reducing power consumption during idle periods.

Depends on:

- [x] https://github.com/thesofproject/sof/pull/8019
- [ ] https://github.com/zephyrproject-rtos/zephyr/pull/66295
- [ ] https://github.com/zephyrproject-rtos/zephyr/pull/66092


src/ipc/ipc4/handler.c
```diff
@@ -516,13 +516,18 @@ static int ipc_wait_for_compound_msg(void)
 	int try_count = 30; /* timeout out is 30 x 10ms so 300ms for IPC */
 
 	while (atomic_read(&msg_data.delayed_reply)) {
+		/* preventing clock switching in idle */
+		pm_runtime_disable(CORE_HP_CLK, PLATFORM_PRIMARY_CORE_ID);
 		k_sleep(Z_TIMEOUT_MS(10));
```
lyakh, line 521: I don't know the hardware details, but naively this feels like the opposite to what one would want: when sleeping you'd want preferably to switch off any clocks, do some kind of a wait-for-interrupt operation, and instead we turn the clock on to a maximum frequency?

tmleman: Maybe I implemented it in the wrong function. But the goal here is not to turn on the highest clock but to prevent switching to the lowest one if a higher one is already in use.

lgirdwood (commented):
@tmleman I assume you are happy with the testing/validation ?

btian1 (commented):
is this PR a replacement with:https://github.com/thesofproject/sof/pull/8019

tmleman:
> is this PR a replacement with:#8019

No, this change is not intended to replace it. It's more of an extension. Thanks for asking; I've added this PR as a dependency.

lgirdwood:
> > is this PR a replacement with:#8019
> 
> No, this change is not intended to replace it. It's more of an extension. Thanks for asking; I've added this PR as a dependency.

Thanks @tmleman - pls keep this PR up to date with testing/readyness indicators. Lets see if we can merge this and #8019 for v2.9.

kv2019i:
Stable-v2.9 branched, this didn't make the cut, bumping to 2.10.

zephyr/lib/pm_runtime.c
```diff
@@ -74,6 +80,12 @@ void platform_pm_runtime_enable(uint32_t context, uint32_t index)
 		tr_dbg(&power_tr, "removing prevent on d0i3 (lock is active=%d)",
 		       pm_policy_state_lock_is_active(PM_STATE_RUNTIME_IDLE, PM_ALL_SUBSTATES));
 		break;
+#if CONFIG_ADSP_DYNAMIC_CLOCK_SWITCHING
```
lgirdwood, line 83: Btw, what's the long term plan to move this into Zephyr as I would expect we should all the platform PM logic in Zephyr.

lgirdwood (commented):
LGTM, but we need to plan to move our platform PM logic into Zephyr.

zephyr/lib/pm_runtime.c
tmleman: We won't be able to move everything related to PM to Zephyr. We need to prevent changes to the clock and dynamic power gating on the application side. What can definitely be moved to Zephyr is the custom power policy (several changes in this direction have already been made).

lgirdwood: ack, but we should do that with Zephyr PM APIs. e.g. ```zephyr_pm_prevent_clock_gating(CLOCK_ID)``` so that eventually we have zero IP code in SOF and its all in Zephyr.  I know we cant yet do this today, but just for the record that we may have to extend Zephyr APIs here.

abonislawski: > zephyr_pm_prevent_clock_gating(CLOCK_ID)

btw, we have clock gating, this is about clock switching.

I was wondering if there is really a benefit from **idle** clock switching (with switching overhead) if we have hw clock **gating** anyway

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -15,13 +15,17 @@ CONFIG_PM_DEVICE_RUNTIME=y
 CONFIG_PM_DEVICE_RUNTIME_EXCLUSIVE=n
 CONFIG_PM_DEVICE_POWER_DOMAIN=y
 CONFIG_PM_POLICY_CUSTOM=y
+CONFIG_PM_STATE_CUSTOM_DATA=y
+CONFIG_ADSP_DYNAMIC_CLOCK_SWITCHING=y
```
abonislawski, line 19: I strongly suggest to rename it (in zephyr) to CONFIG_ADSP_**IDLE**_DYNAMIC_CLOCK_SWITCHING or something like that to avoid confusion with all kconfigs about clocks and kcps

src/ipc/ipc4/handler.c
```diff
@@ -536,16 +536,20 @@ static int ipc_wait_for_compound_msg(void)
 {
 	int try_count = 30;
 
+	pm_runtime_disable(CORE_HP_CLK, PLATFORM_PRIMARY_CORE_ID);
```
abonislawski, line 539: Is this still needed? Will this work correctly with KCPS which changes clock during IPC?

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -22,6 +22,8 @@ CONFIG_POWER_DOMAIN_INTEL_ADSP=y
 CONFIG_ADSP_IMR_CONTEXT_SAVE=y
 CONFIG_ADSP_IDLE_CLOCK_GATING=y
 
+CONFIG_ARCH_CPU_IDLE_CUSTOM=y
```
kv2019i, line 27: Btw, why do you need this? I don't see a custom idle handled added in this patch. 

This setting is now enabled in upstream by default (the new icache-workaround), so no need to enable this on SOF side anymore.

src/ipc/ipc4/handler.c
```diff
@@ -536,16 +536,20 @@ static int ipc_wait_for_compound_msg(void)
 {
 	int try_count = 30;
```
kv2019i, line 537: Could we generalize this a bit? I guess this could be a Kconfig to "keep core clock high when responding to IPC". 
The actual implementation will vary on different platforms, so ideally we'd have a Zephyr interface to communicate this constraints, but for the time being we might need glue code on SOF side (e.g. for Intel this is turned into pm_runtime call on CORE_HP_CLK). 

I'm ok to also go with a TODO comment for now if there's no direct Zephyr interface yet.

kv2019i (commented):
A few comments inline.

tmleman:
The changes this PR depends on will not be merged into Zephyr in the near future, so I'm closing it.

PR #8425, opened 2023-11-01
app/stub*.conf: add CONFIG_SAMPLE_SMART_AMP=y 

marc-hb:
4 very small commits but review commit by commit is required.


kv2019i (approved):
Thanks @marc-hb , looks all good!

lyakh (approved):

kv2019i:
@marc-hb @lgirdwood I don't understand why this happens, but I suspect that some this PR and #8426 , when both now merged, is causing IPC3 stub build fails in mainline now. Probably something simple, but shows up in every PR.

marc-hb:
I made the same guess as @kv2019i . I will confirm and submit some solution today, one way or the other.

marc-hb:
Fix submitted:
- #8441

PR #8427, opened 2023-11-01
topology2: change HDMI_ID according to dai links

macchian:
mapping HDMI_ID sequence because dai id creation without sdw-dmic.


bardliao (approved):

kv2019i (commented):
The commit message could be improved. This is fixing an issue right and the HDMI IDs have to be listed explicitly in this case.

bardliao (approved):

plbossart (approved):

lgirdwood (approved):

PR #8428, opened 2023-11-01
basefw: add support for fw config query for context save

RanderWang:
If CONFIG_ADSP_IMR_CONTEXT_SAVE is enabled, base fw will report fw_context_save is supported to host

kernel pr: https://github.com/thesofproject/linux/pull/4680

### DIFF
--- a/src/audio/base_fw.c
+++ b/src/audio/base_fw.c
@@ -112,6 +112,10 @@ static int basefw_config(uint32_t *data_offset, char *data)
 	tlv_value_set(tuple, IPC4_SCHEDULER_CONFIGURATION, sizeof(sche_cfg), &sche_cfg);
 
 	tuple = tlv_next(tuple);
+	tlv_value_uint32_set(tuple, IPC4_FW_CONTEXT_SAVE,
+			     IS_ENABLED(CONFIG_ADSP_IMR_CONTEXT_SAVE));
+
+	tuple = tlv_next(tuple);
 	*data_offset = (int)((char *)tuple - data);
 
 	return 0;
--- a/src/include/ipc4/base_fw.h
+++ b/src/include/ipc4/base_fw.h
@@ -359,6 +359,12 @@ enum ipc4_fw_config_params {
 	IPC4_TELEMETRY_BUFFER_SIZE = 24,
 	/* HW version information  */
 	IPC4_BUS_HARDWARE_ID = 25,
+	/* Total memory size available in debug window */
+	IPC4_DEBUG_WINDOW_SIZE = 26,
+	/* Number of realtime Key Phrase Buffer (KPB) output pins */
+	IPC4_KPB_RT_SINK_COUNT = 27,
+	/* Manual control of DMI L1 state */
+	IPC4_DMI_FORCE_L1_EXIT = 28,
 	/* Total number of FW config parameters  */
 	IPC4_FW_CFG_PARAMS_COUNT,
 	/* Max config parameter id */
--- a/src/include/ipc4/base_fw.h
+++ b/src/include/ipc4/base_fw.h
@@ -365,6 +365,8 @@ enum ipc4_fw_config_params {
 	IPC4_KPB_RT_SINK_COUNT = 27,
 	/* Manual control of DMI L1 state */
 	IPC4_DMI_FORCE_L1_EXIT = 28,
+	/* FW context save on D3 entry */
+	IPC4_FW_CONTEXT_SAVE = 29,
 	/* Total number of FW config parameters  */
 	IPC4_FW_CFG_PARAMS_COUNT,
 	/* Max config parameter id */


src/audio/base_fw.c
```diff
@@ -111,6 +111,11 @@ static int basefw_config(uint32_t *data_offset, char *data)
 	sche_cfg.sys_tick_source = SOF_SCHEDULE_LL_TIMER;
 	tlv_value_set(tuple, IPC4_SCHEDULER_CONFIGURATION, sizeof(sche_cfg), &sche_cfg);
 
+	if (IS_ENABLED(CONFIG_ADSP_IMR_CONTEXT_SAVE)) {
+		tuple = tlv_next(tuple);
+		tlv_value_uint32_set(tuple, IPC4_FW_CONTEXT_SAVE, 1);
+	}
```
ujfalusi, line 117: This has to be a negative flag, 1 if context save is not available and 0 if context save is implemented.

ujfalusi (changes requested):

src/audio/base_fw.c
RanderWang: sure, updated.

RanderWang:
on kernel side,  I change the default value of fw_context_save on TGL & MTL to ture, unless fw overrides it

btian1 (approved):
just a suggestion in comments:
If CONFIG_ADSP_IMR_CONTEXT_SAVE is enabled, base fw will report
fw_context_save is supported to driver. Driver will assume fw doesn't
support this feature if it is not reported.

ujfalusi:
> just a suggestion in comments: If CONFIG_ADSP_IMR_CONTEXT_SAVE is enabled, base fw will report fw_context_save is supported to driver. Driver will assume fw doesn't support this feature if it is not reported.

The assumption is that the context save is supported if IMR is supported. We need negative flagging for this.

src/audio/base_fw.c
```diff
@@ -111,6 +111,12 @@ static int basefw_config(uint32_t *data_offset, char *data)
 	sche_cfg.sys_tick_source = SOF_SCHEDULE_LL_TIMER;
 	tlv_value_set(tuple, IPC4_SCHEDULER_CONFIGURATION, sizeof(sche_cfg), &sche_cfg);
 
+	tuple = tlv_next(tuple);
+	if (IS_ENABLED(CONFIG_ADSP_IMR_CONTEXT_SAVE))
+		tlv_value_uint32_set(tuple, IPC4_FW_CONTEXT_SAVE, 1);
+	else
+		tlv_value_uint32_set(tuple, IPC4_FW_CONTEXT_SAVE, 0);
```
lyakh, line 118: I think it should work to just use `tlv_value_uint32_set(tuple, IPC4_FW_CONTEXT_SAVE, IS_ENABLED(CONFIG_ADSP_IMR_CONTEXT_SAVE))`

RanderWang: is IS_ENABLED work ?

RanderWang:
> > just a suggestion in comments: If CONFIG_ADSP_IMR_CONTEXT_SAVE is enabled, base fw will report fw_context_save is supported to driver. Driver will assume fw doesn't support this feature if it is not reported.
> 
> The assumption is that the context save is supported if IMR is supported. We need negative flagging for this.

@ujfalusi  how about ID now ?   We don't need negative flag sine the default value in kernel is true, not false.

src/include/ipc4/base_fw.h
```diff
@@ -361,8 +361,10 @@ enum ipc4_fw_config_params {
 	IPC4_BUS_HARDWARE_ID = 25,
 	/* Total number of FW config parameters  */
 	IPC4_FW_CFG_PARAMS_COUNT,
+	/* FW context save on D3 entry */
+	IPC4_FW_CONTEXT_SAVE,
 	/* Max config parameter id */
-	IPC4_MAX_FW_CFG_PARAM = IPC4_FW_CFG_PARAMS_COUNT - 1,
+	IPC4_MAX_FW_CFG_PARAM = IPC4_FW_CONTEXT_SAVE - 1,
```
lgirdwood, line 367: @RanderWang @ujfalusi @mwasko @marcinszkudlinski to be on the safe side - can we add our new Intel reserved IDs as part of this PR and I think its going to be better if we also assign the numbers when updating this enum. i.e there is no ambiguos enum,  developers should always see
```
IPC4_SOME_ENUM = NUMBER,
```

lgirdwood (commented):
Just need to align new IDs

src/include/ipc4/base_fw.h
RanderWang: good idea!  I use 100, big enough for 10 years.

RanderWang:
reserve 100 for this config

src/include/ipc4/base_fw.h
lgirdwood: > good idea! I use 100, big enough for 10 years.

be careful, our industry has a bad experinces with quotes like these ;)

RanderWang:
@lgirdwood  sync latest config from ref fw.

RanderWang:
@lgirdwood  I reserved 3 id, how about it ? Thanks!

src/include/ipc4/base_fw.h
```diff
@@ -359,6 +359,12 @@ enum ipc4_fw_config_params {
 	IPC4_TELEMETRY_BUFFER_SIZE = 24,
 	/* HW version information  */
 	IPC4_BUS_HARDWARE_ID = 25,
+	/* Total memory size available in debug window */
```
lgirdwood, line 362: is this now aligned with reference IDs ?

src/include/ipc4/base_fw.h
```diff
@@ -365,6 +365,10 @@ enum ipc4_fw_config_params {
 	IPC4_KPB_RT_SINK_COUNT = 27,
 	/* Manual control of DMI L1 state */
 	IPC4_DMI_FORCE_L1_EXIT = 28,
+	/* Intel reserved */
+	IPC4_INTEL_RESRVED_29 = 29,
+	IPC4_INTEL_RESRVED_30 = 30,
+	IPC4_INTEL_RESRVED_31 = 31,
```
lgirdwood, line 371: If we are using 31 we can name it using its real name.

src/include/ipc4/base_fw.h
RanderWang: yes, sync with latest ID

src/include/ipc4/base_fw.h
RanderWang: sure

RanderWang:
@lgirdwood  update the ID name

lgirdwood (approved):
@mwasko this PR also reserves the in-use IDs until some process is created to align.

lgirdwood:
@ujfalusi pls review, does kernel align or does it require a similar PR ?

src/include/ipc4/base_fw.h
```diff
@@ -365,6 +365,11 @@ enum ipc4_fw_config_params {
 	IPC4_KPB_RT_SINK_COUNT = 27,
 	/* Manual control of DMI L1 state */
 	IPC4_DMI_FORCE_L1_EXIT = 28,
+	/* Intel reserved */
+	IPC4_INTEL_RESRVED_29 = 29,
+	IPC4_INTEL_RESRVED_30 = 30,
```
kv2019i, line 370: This looks a bit odd, why not RESERVED? I don't get why save one 'e'

src/include/ipc4/base_fw.h
```diff
@@ -359,6 +359,12 @@ enum ipc4_fw_config_params {
 	IPC4_TELEMETRY_BUFFER_SIZE = 24,
 	/* HW version information  */
 	IPC4_BUS_HARDWARE_ID = 25,
+	/* Total memory size available in debug window */
+	IPC4_DEBUG_WINDOW_SIZE = 26,
+	/* Number of Realtime KPB output pins */
```
kv2019i, line 364: Other entries in this enum have been explained with acronyms opened, so better to put "Number of realtime Key Phrase Buffer (KPB) output pins."

kv2019i (commented):
A few comments on the code documentation, otherwise looks good.

src/include/ipc4/base_fw.h
lgirdwood: Ah, yeah, lets fix the spelling.

RanderWang: thanks

src/include/ipc4/base_fw.h
RanderWang: thanks

lgirdwood (approved):

lgirdwood:
@wszypelt @lrudyX not expecting an ID update to break, good to merge ? Thanks !

ujfalusi (approved):

kv2019i (approved):
@mwasko @mmaka1 good to go (this reserves 29 for the new cfg id)?

abonislawski (approved):

tmleman (approved):

RanderWang:
SOFCI TEST

RanderWang:
Will update it according to ref fw which has merged a config

RanderWang:
@lgirdwood  @kv2019i  update PR to follow ref fw https://github.com/intel-innersource/drivers.audio.firmware.cavs-ace/pull/1533

kv2019i:
SOFCI TEST

PR #8432, opened 2023-11-02
schedule: Fix input/output error when using zephyr dma domain

TangleZ:
Run cmd: "arecord -Dhw:1,0 -f S16_LE -r 48000 -c 2 -d 4 rec.wav & sleep 2; aplay -Dhw:1,0 S168K2C.wav" on imx8qxp, we get input output error issue.

The issue caused by try to free playback dma chan when stopping record, causing playback can't get right dma chan at begainning only according to the dma_chan status.

Add registerd flag for dma chan and correct this issue.

### DIFF
--- a/xtos/include/sof/lib/dma.h
+++ b/xtos/include/sof/lib/dma.h
@@ -243,6 +243,8 @@ struct dma_chan_data {
 	/* true if this DMA channel is the scheduling source */
 	bool is_scheduling_source;
 
+	bool registered;
+
 	/* device specific data set by the device that requests the DMA channel */
 	void *dev_data;
 
@@ -493,6 +495,21 @@ static inline void dma_chan_reg_update_bits16(struct dma_chan_data *channel,
 static inline bool dma_is_scheduling_source(struct dma_chan_data *channel)
 {
 	return channel->is_scheduling_source;
+}
+
+static inline void dma_chan_register(struct dma_chan_data *channel)
+{
+	channel->registered = true;
+}
+
+static inline void dma_chan_unregister(struct dma_chan_data *channel)
+{
+	channel->registered = false;
+}
+
+static inline bool dma_chan_is_registered(struct dma_chan_data *channel)
+{
+	return channel->registered;
 }
 
 static inline void dma_sg_init(struct dma_sg_elem_array *ea)
--- a/src/schedule/zephyr_dma_domain.c
+++ b/src/schedule/zephyr_dma_domain.c
@@ -273,6 +273,9 @@ static int register_dma_irq(struct zephyr_dma_domain *domain,
 			if (core != crt_chan->core)
 				continue;
 
+			if (dma_chan_is_registered(crt_chan))
+				continue;
+
 			/* get IRQ number for current channel */
 			irq = interrupt_get_irq(dma_chan_irq(crt_dma, j),
 						dma_chan_irq_name(crt_dma, j));
@@ -308,6 +311,8 @@ static int register_dma_irq(struct zephyr_dma_domain *domain,
 			/* bind registrable ptask to channel */
 			chan_data->pipe_task = pipe_task;
 			chan_data->irq_data = crt_irq_data;
+
+			dma_chan_register(crt_chan);
 
 			list_item_append(&chan_data->list, &crt_irq_data->channels);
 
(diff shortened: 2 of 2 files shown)


xtos/include/sof/lib/dma.h
```diff
@@ -243,6 +243,8 @@ struct dma_chan_data {
 	/* true if this DMA channel is the scheduling source */
 	bool is_scheduling_source;
```
iuliana-prodan, line 245: Add a comment here, like: 
`/* true if this DMA channel is registered, or already in use */`

LaurentiuM1234 (commented):
Not sure I fully understand the issue here. Is the problem caused by the fact that the capture and playback pipelines are bound to the same channel due to the fact that there's no checking whether the channel is taken or not during `register_dma_irq`?

Also, why is the `enabled` field needed in `struct dma_chan_data`? Can't we just check if crt_chan is in crt_irq_data's channel list? Is there a performance-related issue here? If not, I don't think it's reasonable enough to put domain-specific information in generic structures (i.e: structures that are not necessarily used by the domains).

TangleZ:
> Not sure I fully understand the issue here. Is the problem caused by the fact that the capture and playback pipelines are bound to the same channel due to the fact that there's no checking whether the channel is taken or not during `register_dma_irq`?
**> yes. There're checking has a bug in 'register_dma_irq'. If you add some log in here, you will be clear about why it is wrong.**
> Also, why is the `enabled` field needed in `struct dma_chan_data`? Can't we just check if crt_chan is in crt_irq_data's channel list? Is there a performance-related issue here? If not, I don't think it's reasonable enough to put domain-specific information in generic structures (i.e: structures that are not necessarily used by the domains).
> **Well, I think 'enabled' field makes people confused about here, cause we already enabled the chan and make it active. And we can't check if crt_chan is in crt_irq_data cause two dma_chan_data can in one crt_irq_data(sdma) or two crt_irq_data(edma).**

LaurentiuM1234:
> And we can't check if crt_chan is in crt_irq_data cause two dma_chan_data can in one crt_irq_data(sdma) or two crt_irq_data(edma).

Not sure I follow. Why does that matter? If the current channel is in any `struct zephyr_dma_domain_irq` you can conclude that the channel is already taken and move on. I guess you'd translate this into code as follows:

```
static int register_dma_irq(...)
{
    ...

    if (!crt_irq_data) {
         ....
    } else if (chan_is_used(crt_irq_data, crt_chan)) {
        continue;
    }
}
```

`crt_irq_data` corresponds to the INTID used by the current channel. As such, the current channel couldn't have been added to another `crt_irq_data`.

dbaluta:
@LaurentiuM1234 so we should iterate over all INTids registered and for those, we need to iterate over all channels registered with at intid.

Yes, this could work. Given that the number of interrupts and channels is limited the overhead is not that important.

LaurentiuM1234:
> so we should iterate over all INTids registered and for those, we need to iterate over all channels registered with at intid.

No need to go over all INTIDs. In theory, current channel should only be in current INTID's list (fetched through `fetch_irq_data()`).

xtos/include/sof/lib/dma.h
lgirdwood: So is this a DMA channel tracking issue i.e. which DMA to free on stream stop ? Fwiw, not seeing this on timer domain with multiple DMA channels in operation.

TangleZ:
> > so we should iterate over all INTids registered and for those, we need to iterate over all channels registered with at intid.
> 
> No need to go over all INTIDs. In theory, current channel should only be in current INTID's list (fetched through `fetch_irq_data()`).

I agree with Daniel. Right now maybe with the change 
```
static int register_dma_irq(...)
{
    ...

    if (!crt_irq_data) {
         ....
    } else if (chan_is_used(crt_irq_data, crt_chan)) {
        continue;
    }
}
```
all can work fine. But there have a bug is crt_irq_data may only keep parts of dma chan, and other irq_data keep parts of dma chan, this may cause problem I think.

dbaluta:
> all can work fine. But there have a bug is crt_irq_data may only keep parts of dma chan, and other irq_data keep parts of dma chan, this may cause problem I think.

Can you clarify what is the bug here. Can you reprhase this sentence I don't understand it :D

TangleZ:
> > all can work fine. But there have a bug is crt_irq_data may only keep parts of dma chan, and other irq_data keep parts of dma chan, this may cause problem I think.
> 
> Can you clarify what is the bug here. Can you reprhase this sentence I don't understand it :D

I mean the code that Laurentiu provided may have a bug: there may be not only one irq_data, for example: 1st irq_data includes dma chan 0,1, and 2st irq_data includes dma chan 2, and when we check 3st irq_data, we need check dam chan that are included 1st irq_data and 2st irq_data, means need to check all dma chans, not only the list in 3st irq_data.

dbaluta:
@TangleZ @LaurentiuM1234 @iuliana-prodan I've sent https://github.com/thesofproject/sof/pull/8470 to replace this PR. Please have a look.

Issue #8100, opened 2023-08-25
[RFC] limiting visibility of internal structures 


marcinszkudlinski:
PROBLEM: 
there are many structures in SOF that should be modified by dedicated functions only

BUT, unless the structure is protected somehow, it is extremely easy for anybody to access them directly.  A developer may not even know that there are special API functions for reading/writing values, 
Review is fine, but sooner or later somebody will, even accidently, violate the rule and make problems with further code flexibility

In C++ the issue is solved by compiler and public: / protected: / private: members of a class

**In C, however, there's no perfect way to do it**

The only way to make some of members private is to limit their range to single .c file (put a struct definition in .c) and export only proper API functions. 
Unfortunately such API functions cannot be inline and therefore have significant performance impact

**The purpose of this RFC is a brainstorm - how to do it to have a pretty good protection without loosing performance?**

In example, FREE Rtos is doing this in a very hard way, having always 2 versions of structures: internal (in .c file) and exported (in .h), obfuscating all struct members names to meaningless dummy1, dummy2, etc. 
i,e, exported:
```
typedef struct xSTATIC_TIMER
{
    void * pvDummy1;
    StaticListItem_t xDummy2;
    TickType_t xDummy3;
    void * pvDummy5;
    TaskFunction_t pvDummy6;
    #if ( configUSE_TRACE_FACILITY == 1 )
        UBaseType_t uxDummy7;
    #endif
    uint8_t ucDummy8;
} StaticTimer_t;
```
internal
```
    typedef struct tmrTimerControl                                               /* The old naming convention is used to prevent breaking kernel aware debuggers. */
    {
        const char * pcTimerName;                                                /**< Text name.  This is not used by the kernel, it is included simply to make debugging easier. */ /*lint !e971 Unqualified char types are allowed for strings and single characters only. */
        ListItem_t xTimerListItem;                                               /**< Standard linked list item as used by all kernel features for event management. */
        TickType_t xTimerPeriodInTicks;                                          /**< How quickly and often the timer expires. */
        void * pvTimerID;                                                        /**< An ID to identify the timer.  This allows the timer to be identified when the same callback is used for multiple timers. */
        portTIMER_CALLBACK_ATTRIBUTE TimerCallbackFunction_t pxCallbackFunction; /**< The function that will be called when the timer expires. */
        #if ( configUSE_TRACE_FACILITY == 1 )
[...]

PR #8433, opened 2023-11-02
Add get buffer ID to sink/src api

marcinszkudlinski:
sink/src API should grow to fulfill all requirements

this PR adds stream ID to sink/src api

### DIFF
--- a/src/audio/sink_api_helper.c
+++ b/src/audio/sink_api_helper.c
@@ -185,3 +185,9 @@ size_t sink_get_min_free_space(struct sof_sink *sink)
 {
 	return sink->min_free_space;
 }
+
+uint32_t sink_get_id(struct sof_sink *sink)
+{
+	return sink->audio_stream_params->id;
+}
+
--- a/src/include/sof/audio/buffer.h
+++ b/src/include/sof/audio/buffer.h
@@ -47,7 +47,7 @@ extern struct tr_ctx buffer_tr;
 #define trace_buf_get_id(buf_ptr) ((buf_ptr)->pipeline_id)
 
 /** \brief Retrieves subid (comp id) from the buffer */
-#define trace_buf_get_subid(buf_ptr) ((buf_ptr)->id)
+#define buf_get_id(buf_ptr) ((buf_ptr)->stream.runtime_stream_params.id)
 
 #if defined(__ZEPHYR__) && defined(CONFIG_ZEPHYR_LOG)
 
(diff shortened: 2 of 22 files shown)


kv2019i (commented):
Code looks good. The commit description is little light on the rationale. Is this enabling something, or is this just moving an attribute to a more correct place?

I'm a bit worried about the conceptual override where sink/src in theory allows to abstract the "id", while current SOF code has pretty specific semantics on the buffer id (e.g. using "IPC4_SINK_QUEUD_ID(buf_id)" macro on it). Nothing wrong with this itself, but I wonder if this is abstracted more (behind src/sink), should we documented the semantics better.

Btw, you have build errors in IPC3 targets in CI.

src/audio/sink_api_helper.c
```diff
@@ -185,3 +185,9 @@ size_t sink_get_min_free_space(struct sof_sink *sink)
 {
 	return sink->min_free_space;
 }
+
+uint32_t sink_get_id(struct sof_sink *sink)
```
lgirdwood, line 189: These should be static inline. There is PR in flow for splitting header directories and we will be able to catch this in the correct folder i.e. for baseFW or module usage.

lgirdwood (commented):
I like the abstraction.

src/audio/sink_api_helper.c
btian1: static inline and put it tou sink_api.h and source_api.h accordingly.

marcinszkudlinski: yes. We agreed lately the full sink/src encapsulation is too expensive and all "small" functions should be inline (https://github.com/thesofproject/sof/issues/8100)

It has been implemented for sink/src as a part of https://github.com/thesofproject/sof/pull/8365. Till 8365 is merged I prefer to follow the obsolete way to avoid merge conflicts

marcinszkudlinski:
> Code looks good. The commit description is little light on the rationale. Is this enabling something, or is this just moving an attribute to a more correct place?
it is extending sink/src - wer need feature match with current audio stream

> 
> I'm a bit worried about the conceptual override where sink/src in theory allows to abstract the "id", while current SOF code has pretty specific semantics on the buffer id (e.g. using "IPC4_SINK_QUEUD_ID(buf_id)" macro on it). Nothing wrong with this itself, but I wonder if this is abstracted more (behind src/sink), should we documented the semantics better.

i'll check if we can return sink/src part of id
by now i found something like this in code - a complete id used:
```
			j = buf_get_id(source);
			if (j == BASE_CFG_QUEUED_ID)
```


> 
> Btw, you have build errors in IPC3 targets in CI.

working to fix

src/include/sof/audio/buffer.h
```diff
@@ -47,7 +47,7 @@ extern struct tr_ctx buffer_tr;
 #define trace_buf_get_id(buf_ptr) ((buf_ptr)->pipeline_id)
 
 /** \brief Retrieves subid (comp id) from the buffer */
-#define trace_buf_get_subid(buf_ptr) ((buf_ptr)->id)
+#define buf_get_id(buf_ptr) ((buf_ptr)->stream.runtime_stream_params.id)
```
lyakh, line 50: is the ID really a runtime parameter? Isn't it persistent?

kv2019i (approved):

btian1 (approved):

src/include/sof/audio/buffer.h
marcinszkudlinski: it is set once, not modified later. Anyway - in rutime
As it is a common parameter for all streams, no matter how buffers are implemented, so I put it to this container
But you're right, it the name of the container may be somehow misleading, I'll change it

lyakh:
Sorry, I do not understand this PR. The code seems to add a parameter that isn't used apart from for logging ant the description "fulfill all requirements" doesn't really clarify it either. Is logging the only reason to add this?

marcinszkudlinski:
@lyakh ???
see how many times buf_get_id() is used - its close to 50
yes, ID is used in logging, but also many other places - copier, google AEC, probe. Its purpose is to identify the pins the buffer is connected to. 
image
the purpose of this PR is to provide access to the ID for components using sink/src interface

lyakh:
> @lyakh ??? see how many times buf_get_id() is used - its close to 50 yes, ID is used in logging, but also many other places - copier, google AEC, probe. Its purpose is to identify the pins the buffer is connected to. image the purpose of this PR is to provide access to the ID for components using sink/src interface

I mean the second commit - an ID is added to source / sink and it doesn't seem to be used

marcinszkudlinski:
rebase only

marcinszkudlinski:
@lyakh it is used - in a PR mentioned above

kv2019i (approved):
PR looks good, but mandatory CI is failing again -> needs to be checked.

marcinszkudlinski:
> PR looks good, but mandatory CI is failing again -> needs to be checked.

lets retry - the logs indicates some issues with test setup
rebasing to head again

PR #8435, opened 2023-11-03
DRC: math: Replace exponential function for performance

ShriramShastry:
For DRC performance, replace `exp_small_fixed()` with `sofm_exp_int32()`.
Included supporting change to include `sofm_exp_int32()` within `exp_fixed()`
and repositioned `exp_small_fixed()` for future use.

### DIFF
--- a/src/include/sof/math/exp_fcn.h
+++ b/src/include/sof/math/exp_fcn.h
@@ -59,5 +59,7 @@
 #define SOFM_EXP_LSHIFT_BITS 0x2000
 
 int32_t sofm_exp_int32(int32_t x);
+int32_t sofm_exp_fixed(int32_t x);
+int32_t sofm_db2lin_fixed(int32_t db);
 
 #endif /* __SOFM_EXP_FCN_H__ */
--- a/src/include/sof/math/exp_fcn.h
+++ b/src/include/sof/math/exp_fcn.h
@@ -5,8 +5,8 @@
  * Author: Shriram Shastry <malladi.sastry@linux.intel.com>
  *
  */
-#ifndef __SOFM_EXP_H__
-#define __SOFM_EXP_H__
+#ifndef __SOFM_EXP_FCN_H__
+#define __SOFM_EXP_FCN_H__
 
 #include <stdint.h>
 
@@ -26,6 +26,38 @@
 
 #endif
 
-int32_t sofm_exp_int32(int32_t x);
+/* TODO: Is there a MCPS difference */
+#define USING_QCONVERT	1
+
+#if USING_QCONVERT
+
+#include <sof/audio/format.h>
+#define SOFM_EXP_FIXED_INPUT_MIN	Q_CONVERT_FLOAT(-11.5, 27)		/* Q5.27 */
+#define SOFM_EXP_FIXED_INPUT_MAX	Q_CONVERT_FLOAT(7.6245, 27)		/* Q5.27 */
+#define SOFM_EXP_TWO_Q27		Q_CONVERT_FLOAT(2.0, 27)		/* Q5.27 */
+#define SOFM_EXP_MINUS_TWO_Q27		Q_CONVERT_FLOAT(-2.0, 27)		/* Q5.27 */
+#define SOFM_EXP_ONE_Q20		Q_CONVERT_FLOAT(1.0, 20)		/* Q12.20 */
+#define SOFM_EXP_MINUS_100_Q24		Q_CONVERT_FLOAT(-100.0, 24)		/* Q8.24 */
+#define SOFM_EXP_LOG10_DIV20_Q27	Q_CONVERT_FLOAT(0.1151292546, 27)	/* Q5.27 */
+
+#else
+
+#define SOFM_EXP_FIXED_INPUT_MIN	-1543503872	/* Q_CONVERT_FLOAT(-11.5, 27) */
+#define SOFM_EXP_FIXED_INPUT_MAX	 1023343067	/* Q_CONVERT_FLOAT(7.6245, 27) */
+#define SOFM_EXP_TWO_Q27		 268435456	/* Q_CONVERT_FLOAT(2.0, 27) */
+#define SOFM_EXP_MINUS_TWO_Q27		-268435456	/* Q_CONVERT_FLOAT(-2.0, 27) */
+#define SOFM_EXP_ONE_Q20		 1048576	/* Q_CONVERT_FLOAT(1.0, 20) */
+#define SOFM_EXP_MINUS_100_Q24		-1677721600	/* Q_CONVERT_FLOAT(-100.0, 24) */
+#define SOFM_EXP_LOG10_DIV20_Q27	 15452387	/* Q_CONVERT_FLOAT(0.1151292546, 27) */
 
 #endif
+
+#define SOFM_EXP_BIT_MASK_LOW_Q27P5 0x0000000008000000
+#define SOFM_EXP_BIT_MASK_Q62P2 0x4000000000000000LL
+#define SOFM_EXP_QUOTIENT_SCALE 0x40000000
+#define SOFM_EXP_TERMS_Q23P9 0x800000
+#define SOFM_EXP_LSHIFT_BITS 0x2000
+
+int32_t sofm_exp_int32(int32_t x);
+
+#endif /* __SOFM_EXP_FCN_H__ */
--- a/src/math/exp_fcn.c
+++ b/src/math/exp_fcn.c
@@ -6,6 +6,7 @@
  *
  */
 
+#include <sof/audio/format.h>
 #include <sof/math/exp_fcn.h>
 #include <sof/math/numbers.h>
 #include <sof/common.h>
@@ -212,4 +213,74 @@ int32_t sofm_exp_int32(int32_t x)
 	}
 	return ts;
 }
-#endif
+
+/* Fixed point exponent function for approximate range -11.5 .. 7.6
+ * that corresponds to decibels range -100 .. +66 dB.
+ *
+ * The functions uses rule exp(x) = exp(x/2) * exp(x/2) to reduce
+ * the input argument for private small value exp() function that is
+ * accurate with input range -2.0 .. +2.0. The number of possible
+ * divisions by 2 is computed into variable n. The returned value is
+ * exp()^(2^n).
+ *
+ * Input  is Q5.27, -16.0 .. +16.0, but note the input range limitation
+ * Output is Q12.20, 0.0 .. +2048.0
+ */
+
+int32_t sofm_exp_fixed(int32_t x)
+{
+	int32_t xs;
+	int32_t y;
+	int32_t y0;
+	int i;
+	int n = 0;
+
+	if (x < SOFM_EXP_FIXED_INPUT_MIN)
+		return 0;
+
+	if (x > SOFM_EXP_FIXED_INPUT_MAX)
+		return INT32_MAX;
+
+	/* x is Q5.27 */
+	xs = x;
+	while (xs >= SOFM_EXP_TWO_Q27 || xs <= SOFM_EXP_MINUS_TWO_Q27) {
+		xs >>= 1;
+		n++;
+	}
+
+	/* sofm_exp_int32() input is Q4.28, while x1 is Q5.27
+	 * sofm_exp_int32() output is Q9.23, while y0 is Q12.20
+	 */
+	y0 = Q_SHIFT_RND(sofm_exp_int32(Q_SHIFT_LEFT(xs, 27, 28)), 23, 20);
+	y = SOFM_EXP_ONE_Q20;
+	for (i = 0; i < (1 << n); i++)
+		y = (int32_t)Q_MULTSR_32X32((int64_t)y, y0, 20, 20, 20);
+
+	return y;
+}
+
+#endif /* EXPONENTIAL_GENERIC */
+
+/* Decibels to linear conversion: The function uses exp() to calculate
+ * the linear value. The argument is multiplied by log(10)/20 to
+ * calculate equivalent of 10^(db/20).
+ *
+ * The error in conversion is less than 0.1 dB for -89..+66 dB range. Do not
+ * use the code for argument less than -100 dB. The code simply returns zero
+ * as linear value for such very small value.
+ *
+ * Input is Q8.24 (max 128.0)
+ * output is Q12.20 (max 2048.0)
+ */
+
+int32_t sofm_db2lin_fixed(int32_t db)
+{
+	int32_t arg;
+
+	if (db < SOFM_EXP_MINUS_100_Q24)
+		return 0;
+
+	/* Q8.24 x Q5.27, result needs to be Q5.27 */
+	arg = (int32_t)Q_MULTSR_32X32((int64_t)db, SOFM_EXP_LOG10_DIV20_Q27, 24, 27, 27);
+	return sofm_exp_fixed(arg);
+}
--- a/src/math/exp_fcn.c
+++ b/src/math/exp_fcn.c
@@ -17,12 +17,7 @@
 
 #if defined(EXPONENTIAL_GENERIC)
 
-#define SOFM_BIT_MASK_Q62P2 0x4000000000000000LL
 #define SOFM_CONVERG_ERROR 28823037607936LL	// error smaller than 1e-4,1/2 ^ -44.7122876200884
-#define SOFM_BIT_MASK_LOW_Q27P5 0x8000000
-#define SOFM_QUOTIENT_SCALE BIT(30)
-#define SOFM_TERMS_Q23P9 8388608
-#define SOFM_LSHIFT_BITS 8192
 
 /* inv multiplication lookup table */
 /* LUT = ceil(1/factorial(b_n) * 2 ^ 63) */
@@ -155,7 +150,7 @@ static inline int64_t lomul_s64_sr_sat_near(int64_t a, int64_t b)
 	uint64_t u64_rlo;
 
 	mul_s64(a, b, &u64_rhi, &u64_rlo);
-	const bool roundup = (u64_rlo & SOFM_BIT_MASK_LOW_Q27P5) != 0;
+	const bool roundup = (u64_rlo & SOFM_EXP_BIT_MASK_LOW_Q27P5) != 0;
 
 	u64_rlo = (u64_rhi << 36 | u64_rlo >> 28) + (roundup ? 1 : 0);
 	return u64_rlo;
@@ -182,8 +177,8 @@ int32_t sofm_exp_int32(int32_t x)
 	uint64_t ou0Lo;
 	int64_t qt;
 	int32_t b_n;
-	int32_t ts =  SOFM_TERMS_Q23P9; /* Q23.9 */
-	int64_t dividend = (x + SOFM_LSHIFT_BITS) >> 14; /* x in Q50.14 */
+	int32_t ts =  SOFM_EXP_TERMS_Q23P9; /* Q23.9 */
+	int64_t dividend = (x + SOFM_EXP_LSHIFT_BITS) >> 14; /* x in Q50.14 */
 	static const int32_t i_emin = -1342177280; /* Q4.28 */
 	static const int32_t o_emin = 56601; /* Q9.23 */
 	static const int32_t i_emax = 1342177280; /* Q4.28 */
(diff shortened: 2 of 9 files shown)


ShriramShastry:
For compilation, this PR requires further fixes.

singalsu:
In my test this patch dropped drc.2.1 average MCPS from 140 to 96 in topology sof-hda-efx-generic-4ch.tplg. Excellent!

The used topology initializes DRC to passthrough, so I applied another with "sof-ctl -n 36 -s threshold_-35_knee_27_ratio_8.txt". The blob that I used  is not in SOF git yet but I'll attach it here:

[threshold_-35_knee_27_ratio_8.txt](https://github.com/thesofproject/sof/files/13251190/threshold_-35_knee_27_ratio_8.txt)

singalsu (changes requested):

src/math/decibels.c
```diff
@@ -105,10 +72,10 @@ int32_t exp_fixed(int32_t x)
 		n++;
 	}
 
-	/* exp_small_fixed() input is Q3.29, while x1 is Q5.27
-	 * exp_small_fixed() output is Q9.23, while y0 is Q12.20
+	/* sofm_exp_int32() input is Q4.28, while x1 is Q5.27
+	 * sofm_exp_int32() output is Q9.23, while y0 is Q12.20
 	 */
-	y0 = Q_SHIFT_RND(exp_small_fixed(Q_SHIFT_LEFT(xs, 27, 29)), 23, 20);
+	y0 = Q_SHIFT_RND(sofm_exp_int32(Q_SHIFT_LEFT(xs, 27, 28)), 23, 20);
```
btian1, line 78: line 78 - 81 still can be optimized further with instrinsic code.

andrula-song: this file didn't include xtensa header file, do you mean add hifi implementation of functions in this file?

btian1: yes, each for loop deserve an intrinsic implementation.

lgirdwood:
@singalsu @ShriramShastry @andrula-song I think we need to have some options where we can decide at build time via Kconfig between speed and accuracy for certain maths ops. i.e. for 16bit output we may not need full 24/32 computations and so on. There may also be usages where speed is more impartnat than bit accuracy, but thats up to you guys to identify and target.

src/audio/Kconfig
```diff
@@ -280,6 +280,7 @@ config COMP_DRC
 	select NUMBERS_NORM
 	select MATH_DECIBELS
```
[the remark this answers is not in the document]
singalsu: I meant don't remove or edit MATH_DECIBELS function but add the needed functions to MATH_EXP. Then you can remove from Kconfig for DRC the dependency to MATH_DECIBELS.

src/math/decibels.c
singalsu: Please check if it improves speed. In some cases the C multiply has been as fast.

singalsu:
> @singalsu @ShriramShastry @andrula-song I think we need to have some options where we can decide at build time via Kconfig between speed and accuracy for certain maths ops. i.e. for 16bit output we may not need full 24/32 computations and so on. There may also be usages where speed is more important than bit accuracy, but thats up to you guys to identify and target.

Yep, I wanted to avoid Sriram to change the old exponent function code since it is used by many other features. The new version for DRC is significantly faster but trades off slightly accuracy. We need to check case by case with exponent function to utilize.

src/math/decibels.c
ShriramShastry: I'll give { `Q_CONVERT_FLOAT` ~, { `Q_SHIFT_RND` ~, { `Q_SHIFT_LEFT` ~, and `Q_MULTSR_32X32` } a shot as well. If I don't succeed, I'll try again with the next PR.

singalsu:
@ShriramShastry Any updates to this? The MCPS saving is large so we should get this cleaned up and merged.

src/math/decibels.c
ShriramShastry: image

ShriramShastry: I have completed the implementation of Q_SHIFT_RND, Q_SHIFT_LEFT, and Q_MULTSR_32X32, and I am currently working on Q_CONVERT_FLOAT.

lgirdwood (approved):
LGTM. @andrula-song pls review.

lgirdwood:
@ShriramShastry some conflicts, pls rebase as CI wont complete.

src/math/decibels.c
ShriramShastry: I'll correct the last few CI errors.

src/math/exp_fcn.c
```diff
@@ -217,4 +221,52 @@ int32_t sofm_exp_int32(int32_t x)
...
+ */
+int32_t sofm_exp_fixed(int32_t x)
+{
+	int32_t xs;
+	int32_t y;
+	int32_t y0;
+	int i;
+	int n = 0;
+
+	if (x < Q_CONVERT_FLOAT(-11.5, 27))
```
andrula-song, line 249: can we use a macro replace those fixed float convert? from my test, the division really costs many cycles.

src/math/exp_fcn_hifi.c
```diff
@@ -278,4 +284,130 @@ int32_t sofm_exp_int32(int32_t x)
...
+ */
+int32_t sofm_exp_fixed(int32_t x)
+{
+	int32_t xs;
+	int32_t y;
+	int32_t y0;
+	int i;
+	int n = 0;
+
+	if (x < XT_Q_CONVERT_FLOAT(-11.5, 27))
```
andrula-song, line 374: here too

ShriramShastry: The updated macro is now in use here.

src/math/exp_fcn.c
ShriramShastry: Yup, We can swap it out, saving about 93% of the cycles compared to Generic C. We can handle this in the future PR as the present one just addresses modifications for exponential. I hope you will consent to this.

andrula-song (approved):

src/math/exp_fcn.c
andrula-song: actually, I'd like use a simple macro like #define THERSHOULD Q_CONVERT_FLOAT(-11.5, 27) to replace the convert. because the result is a fixed statistic.

src/math/exp_fcn_hifi.c
andrula-song: curious about the cycles that XT_Q_CONVERT_FLOAT takes.

ShriramShastry: https://github.com/thesofproject/sof/pull/8435#discussion_r1404095250

src/math/exp_fcn.c
```diff
@@ -20,7 +23,7 @@
 #define SOFM_BIT_MASK_Q62P2 0x4000000000000000LL
 #define SOFM_CONVERG_ERROR 28823037607936LL	// error smaller than 1e-4,1/2 ^ -44.7122876200884
 #define SOFM_BIT_MASK_LOW_Q27P5 0x8000000
-#define SOFM_QUOTIENT_SCALE BIT(30)
+#define SOFM_QUOTIENT_SCALE 0x400000000
```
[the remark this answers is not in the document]
ShriramShastry: Oops, your right !!

src/math/exp_fcn_hifi.c
```diff
@@ -26,10 +27,13 @@
...
+#include <xtensa/tie/xt_FP.h>
+
 #define SOFM_CONVERG_ERROR 28823037624320LL /* error smaller than 1e-4,1/2 ^ -44.7122876209085 */
 #define SOFM_BIT_MASK_LOW_Q27P5 0x0000000008000000
 #define SOFM_BIT_MASK_Q62P2 0x4000000000000000LL
-#define SOFM_QUOTIENT_SCALE BIT(30)
+#define SOFM_QUOTIENT_SCALE 0x40000000
 #define SOFM_TERMS_Q23P9 0x800000
 #define SOFM_LSHIFT_BITS 0x2000
 /*
```
lgirdwood, line 39: Btw, these macros look similar (the same ?) as macros in the previous source file. If so, we should declare them in a exp header.

src/math/exp_fcn_hifi.c
```diff
@@ -278,4 +284,130 @@ int32_t sofm_exp_int32(int32_t x)
 
 	return AE_MOVAD32_L(AE_MOVINT32X2_FROMINT64(ts));
 }
+
+/* Xtensa macros */
+/* Convert a float number to fractional Qnx.ny format. Note that there is no
+ * check for nx+ny number of bits to fit the word length of int. The parameter
+ * qy must be 31 or less.
+ */
+static inline int sofm_q_CONVERT_FLOAT(float a, int b)
```
lgirdwood, line 293: sorry, my mistake - I did not see these were in the hifi file and not public APIs. These can be renamed to ```exp_hifi_q_convert_float()``` all lower case and so on.

keqiaozhang:
SOFCI TEST

keqiaozhang:
SOFCI TEST

src/math/exp_fcn.c
```diff
@@ -217,4 +216,52 @@ int32_t sofm_exp_int32(int32_t x)
 	}
 	return ts;
 }
+
+#define ONE_Q20         Q_CONVERT_FLOAT(1.0, 20)	  /* Use Q12.20 */
+#define TWO_Q27         Q_CONVERT_FLOAT(2.0, 27)	  /* Use Q5.27 */
+#define MINUS_TWO_Q27   Q_CONVERT_FLOAT(-2.0, 27)	  /* Use Q5.27 */
```
btian1, line 222: those definition are too simple, suggest add EXP_ to avoid potential conflicts.
or you may directly use existing definition:
https://github.com/thesofproject/sof/blob/main/src/math/decibels.c

put it to a header file.

src/math/exp_fcn.c
```diff
@@ -217,4 +216,52 @@ int32_t sofm_exp_int32(int32_t x)
 	}
 	return ts;
 }
+
+#define ONE_Q20         Q_CONVERT_FLOAT(1.0, 20)	  /* Use Q12.20 */
```
btian1, line 220: duplicated name, suggest move to common header.

src/math/decibels.c
ShriramShastry: I've finished addressing every CI error.

btian1: @ShriramShastry , still code style error.

src/math/decibels.c
```diff
@@ -105,10 +106,10 @@ int32_t exp_fixed(int32_t x)
 		n++;
 	}
 
-	/* exp_small_fixed() input is Q3.29, while x1 is Q5.27
-	 * exp_small_fixed() output is Q9.23, while y0 is Q12.20
+	/* sofm_exp_int32() input is Q4.28, while x1 is Q5.27
```
btian1: for comments, you mentioned input is Q4.28, however, what is x1? means sofm_exp_int32() output?
if means output, please change x1 to output.

src/math/exp_fcn.c
```diff
@@ -212,4 +213,54 @@ int32_t sofm_exp_int32(int32_t x)
...
+ * the input argument for private small value exp() function that is
+ * accurate with input range -2.0 .. +2.0. The number of possible
+ * divisions by 2 is computed into variable n. The returned value is
+ * exp()^(2^n).
+ *
+ * Input  is Q5.27, -16.0 .. +16.0, but note the input range limitation
+ * Output is Q12.20, 0.0 .. +2048.0
+ */
+
+int32_t exp_fixed(int32_t x)
```
btian1: your patch sequence are quite strange, normally, we should have c version first, then hifi version, seems you already have exp_fixed hifi in previous patch, now this patch comes with c version, this is strange sequence.

src/math/exp_fcn_hifi.c
```diff
@@ -302,10 +367,10 @@ int32_t exp_fixed(int32_t x)
 	int i;
 	int n = 0;
 
-	if (x < Q_CONVERT_FLOAT(-11.5, 27))
+	if (x < exp_hifi_q_convert_float(-11.5, 27))
```
btian1, line 367: this is not build with pre-compiler stage? could you show the asm code to compare?

src/math/exp_fcn_hifi.c
```diff
@@ -277,10 +280,72 @@ int32_t sofm_exp_int32(int32_t x)
...
+	ae_int32 xt_o;
+	int shift;
+
+	shift = XT_SUB(c, b);
+	xt_o = AE_SLAA32(a, shift);
+
+	return xt_o;
+}
+
+#define ONE_Q20         exp_hifi_q_convert_float(1.0, 20)	  /* Use Q12.20 */
```
btian1: these are always repeating definition, please move to one common header.

src/math/exp_fcn_hifi.c
```diff
@@ -318,10 +383,13 @@ int32_t exp_fixed(int32_t x)
...
 	 */
-	y0 = Q_SHIFT_RND(sofm_exp_int32(Q_SHIFT_LEFT(xs, 27, 28)), 23, 20);
+	y0 = exp_hifi_q_shift_rnd(
+							  sofm_exp_int32(
+							  exp_hifi_q_shift_left(xs, 27, 28)),
+							  23, 20);
 	y = ONE_Q20;
 	for (i = 0; i < (1 << n); i++)
-		y = (int32_t)Q_MULTSR_32X32((int64_t)y, y0, 20, 20, 20);
+		y = (int32_t)exp_hifi_q_multsr_32x32((int64_t)y, y0, 20, 20, 20);
```
btian1, line 389: this patch is not related with title(exp), you can't use one PR to cover all the changes, please move out this patch from this PR.

src/math/window.c
```diff
@@ -114,7 +115,7 @@ void win_povey_16b(int16_t win[], int length)
 		/* Calculate x^0.85 as exp(0.85 * log(x)) */
 		x2 = (int32_t)(ln_int32((uint32_t)x1) >> 1) - WIN_LOG_2POW31_Q26;
 		x3 = sat_int32(Q_MULTSR_32X32((int64_t)x2, WIN_085_Q31, 26, 31, 27)); /* Q5.27 */
-		x4 = exp_fixed(x3); /* Q5.27 -> Q12.20 */
+		x4 = sofm_exp_fixed(x3); /* Q5.27 -> Q12.20 */
```
btian1, line 118: somehow, you can rename from the beginning or start from first patch, then this patch will be removed.

zephyr/CMakeLists.txt
```diff
@@ -370,6 +370,8 @@ zephyr_library_sources(
 	${SOF_MATH_PATH}/decibels.c
 	${SOF_MATH_PATH}/numbers.c
 	${SOF_MATH_PATH}/trig.c
+	${SOF_MATH_PATH}/exp_fcn.c
+	${SOF_MATH_PATH}/exp_fcn_hifi.c
```
marc-hb, line 374: heads-up this will conflict with #8548

src/audio/module_adapter/module/generic.c
```diff
@@ -198,7 +198,7 @@ int module_prepare(struct processing_module *mod,
 		   struct sof_source **sources, int num_of_sources,
 		   struct sof_sink **sinks, int num_of_sinks)
 {
-	int ret;
+	int ret = 0;
```
singalsu, line 201: I made without knowing about this separate PR #8605. You can leave this fix out from DRC PR. I will add credits of patch to you since you did this first.

src/math/decibels.c
```diff
@@ -9,42 +9,8 @@
 #include <sof/audio/format.h>
 #include <stdint.h>
 
-#define ONE_Q23         Q_CONVERT_FLOAT(1.0, 23)	  /* Use Q9.23 */
 #define LOG10_DIV20_Q27 Q_CONVERT_FLOAT(0.1151292546, 27) /* Use Q5.27 */
 
-/* Exponent function for small values of x. This function calculates
```
singalsu, line 17: Don't remove this. Both exponent functions should be possible to enable via kconfig. To be able to select both until we can safely remove the slower they should have different function name. DRC should use this new fast function.

Can your exponent function have different name, like have _fast_ in function name or some algorithm specific keyword to differentiate?  Or rename exp_fixed() and all usages of it (except DRC)  to e.g. exp_taylor_fixed(), since it's using Taylor -series.

singalsu (changes requested):

src/math/exp_fcn_hifi.c
```diff
@@ -274,4 +275,54 @@ int32_t sofm_exp_int32(int32_t x)
...
+ * the input argument for private small value exp() function that is
+ * accurate with input range -2.0 .. +2.0. The number of possible
+ * divisions by 2 is computed into variable n. The returned value is
+ * exp()^(2^n).
+ *
+ * Input  is Q5.27, -16.0 .. +16.0, but note the input range limitation
+ * Output is Q12.20, 0.0 .. +2048.0
+ */
+
+int32_t exp_fixed(int32_t x)
```
singalsu: You don't get conflict with the other exponent if you use here other name, e.g. sofm_exp_<your method name>_fixed(). And DRC should use the new function name.

src/math/exp_fcn_hifi.c
ShriramShastry: Sorry, I don't get the necessity in this situation; could you please explain?

src/math/exp_fcn_hifi.c
```diff
@@ -277,10 +280,72 @@ int32_t sofm_exp_int32(int32_t x)
...
+	x = XT_FLOAT_S(val, 0);
+	y = XT_MUL_S(a, x);
+
+	return XT_TRUNC_S(y, 0);
+}
+
+/* Fractional multiplication with shift and round
+ * Note that the parameters px and py must be cast to (int64_t) if other type.
+ */
+static inline int exp_hifi_q_multsr_32x32(int a, int b, int c, int d, int e)
```
singalsu, line 303: This is also suspicious, there's no need for generic C speed optimization with generic fractions. You should do this with inline instrinsics with minimal operations if need fast intrinsic multiply. It's would be usually just an AE multiply instruction with a literal constant shift to adjust the fractional format. Please leave optimizations like this to other PR where we can review them more carefully.

src/math/exp_fcn_hifi.c
```diff
@@ -277,10 +280,72 @@ int32_t sofm_exp_int32(int32_t x)
...
+	res = AE_SRAA64(res, shift);
+	res = AE_ADD64(res, 1);
+	res = AE_SRAI64(res, 1);
+	xt_o = AE_MOVINT32_FROMINT64(res);
+
+	return xt_o;
+}
+
+/* A macro for Q-shifts */
+static inline int exp_hifi_q_shift_rnd(int a, int b, int c)
```
singalsu, line 320: Same for all other fractional C macros replacements.

singalsu:
Seems it's not ensured by C standards that literals floating point macros are calculated in the C pre-processor. Maybe things have changed with xt-clang.

https://stackoverflow.com/questions/21241031/does-the-c-preprocessor-handle-floating-point-math-constants

It would be safest to change macros like `#define SOME_COEF Q_CONVERT_FLOAT(-11.5, 27)` into `#define SOME_COEF -1543503872 /* -11.5 Q5.27 */`.

singalsu:
> Seems it's not ensured by C standards that literals floating point macros are calculated in the C pre-processor. Maybe things have changed with xt-clang.
> 
> https://stackoverflow.com/questions/21241031/does-the-c-preprocessor-handle-floating-point-math-constants
> 
> It would be safest to change macros like `#define SOME_COEF Q_CONVERT_FLOAT(-11.5, 27)` into `#define SOME_COEF -1543503872 /* -11.5 Q5.27 */`.

@ShriramShastry Please check disassembly of your Q_CONVERT_FLOAT() usage. We looked into TGL and MTL builds with @btian1 and we could not find floating point code generation from current usages of the macro. No, float overhead found. So, let's not yet start to replace these before understanding better as I worried in my previous comment.

src/math/exp_fcn_hifi.c
btian1: I mean why not directly use:Q_CONVERT_FLOAT(-11.5, 27)?

test/cmocka/src/math/window/CMakeLists.txt
```diff
@@ -8,4 +8,5 @@ cmocka_test(window
 	${PROJECT_SOURCE_DIR}/src/math/base2log.c
 	${PROJECT_SOURCE_DIR}/src/math/decibels.c
 	${PROJECT_SOURCE_DIR}/src/math/exp_fcn.c
+	${PROJECT_SOURCE_DIR}/src/math/exp_fcn_hifi.c
```
btian1, line 11: please squash this patch with previous one, as Marc said, you need get each patch pass all CI separately, here, there is obvious CI error.

singalsu (approved):

src/include/sof/math/exp_fcn.h
```diff
@@ -26,6 +26,38 @@
 
 #endif
 
-int32_t sofm_exp_int32(int32_t x);
+/* TODO: Is there a MCPS difference */
+#define USING_QCONVERT	1
```
singalsu, line 30: I already approved, but let's check in real devices if setting this to zero speeds up the code. In theory it shouldn't impact since Q_CONVERT_FLOAT macro should be evaluated in C pre-processor. Once confirmed we can remove the direct integers. Or if difference seen, remove the Q_CONVERT_FLOAT part.

lgirdwood: ok, can we do this as next steps in a follow up PR and test drive this now.

lgirdwood (approved):

PR #8438, opened 2023-11-03
Tools: Testbench: Fix ASRC component load

singalsu:
Before the fix the comp->core in ipc_comp_new() is garbage and the check for maximum core count fails. The pointer to struct comp should be passed here instead of asrc.

Fixes:	d9f9340e1a8ebc276fb67bf9e0a0c7bd08e4e52b
	("tplg_parser: split out testbench logic and partition
	features per file.")

### DIFF
--- a/tools/testbench/topology.c
+++ b/tools/testbench/topology.c
@@ -50,7 +50,7 @@ static int tb_register_asrc(struct testbench_prm *tp, struct tplg_context *ctx)
 		asrc->source_rate = tp->fs_in;
 
 	/* load asrc component */
-	if (ipc_comp_new(sof->ipc, ipc_to_comp_new(&asrc)) < 0) {
+	if (ipc_comp_new(sof->ipc, ipc_to_comp_new(asrc)) < 0) {
 		fprintf(stderr, "error: new asrc comp\n");
 		return -EINVAL;
 	}


lgirdwood (approved):

lgirdwood:
@andrula-song @kv2019i

tools/testbench/topology.c
```diff
@@ -50,7 +50,7 @@ static int tb_register_asrc(struct testbench_prm *tp, struct tplg_context *ctx)
 		asrc->source_rate = tp->fs_in;
 
 	/* load asrc component */
-	if (ipc_comp_new(sof->ipc, ipc_to_comp_new(&asrc)) < 0) {
+	if (ipc_comp_new(sof->ipc, ipc_to_comp_new(comp)) < 0) {
```
btian1, line 53: you can simply remove &, it is double pointers if you add &, no need change name to comp

singalsu: Yes, you are right, `asrc` makes sense here better here even if they point to same struct beginning.

singalsu: Fixed!

btian1 (approved):

andrula-song (approved):

PR #8439, opened 2023-11-03
Tools: Test: Audio: Add possibility to test ASRC with src_test.m

singalsu:


tools/test/audio/src_test.m
```diff
@@ -435,7 +441,7 @@
 %%
 
 function test = test_defaults_src(t)
-test.comp = 'src';
+test.comp = t.comp;
```
ShriramShastry, line 444: In function `test_run_src(test, t)`, t is unused.

ShriramShastry (approved):
I reviewed the Matlab changes and added one minor review comment, but the improvements seems to be good to me.
Thank you!

tools/test/audio/src_test.m
lgirdwood: @singalsu pls fix in subsequent PR. I think this is a separate issue.

lgirdwood (approved):

tools/test/audio/src_test.m
singalsu: Thanks, the fix is now in #8486.

PR #8441, opened 2023-11-03
samples: change smart_amp_test prefix `smart_amp_` to `smart_amp_test_`

marc-hb:
Avoids conflicts with the "real" smart_amp.c when the "stub" compilation check enables everything and compiles everything.

Also makes `git grep smart_thing` a bit more convenient in some cases.

Only the IPC3 sample clashed because the real code does not use MODULE_ADAPTER(). Pre-emptively rename the IPC4 sample too for consistency and in case the real code changes its mind later.

Only global symbols are renamed; no need to git churn private symbols.

Fixes commit ba2cfe3daa55 ("app/stub*.conf: add
CONFIG_SAMPLE_SMART_AMP=y") which clashed with commit d4d0a0ca134f ("smart_amp: fix code and re-enable it on stub builds")

CI did not detect the clash because these two commits were merged at the same time.

In case anyone is interested, the CI "fix" to prevent this type of race is a GitHub feature called "merge queue" (resp. "merge train" in Gitlab, "commit queue" in the older Chrome infrastructure, etc.)  It would be a heavy lift, overkill for the small size and concurrency of the sof git repo and how rarely these accidents happen (a couple times/year)

### DIFF
--- a/src/samples/audio/smart_amp_test_ipc3.c
+++ b/src/samples/audio/smart_amp_test_ipc3.c
@@ -21,10 +21,10 @@ static const struct comp_driver comp_smart_amp;
 LOG_MODULE_REGISTER(smart_amp_test, CONFIG_SOF_LOG_LEVEL);
 
 /* 167a961e-8ae4-11ea-89f1-000c29ce1635 */
-DECLARE_SOF_RT_UUID("smart_amp-test", smart_amp_comp_uuid, 0x167a961e, 0x8ae4,
+DECLARE_SOF_RT_UUID("smart_amp-test", smart_amp_test_comp_uuid, 0x167a961e, 0x8ae4,
 		    0x11ea, 0x89, 0xf1, 0x00, 0x0c, 0x29, 0xce, 0x16, 0x35);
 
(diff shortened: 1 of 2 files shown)


kv2019i (approved):
~~I'm 100% not sure if the ability to compile real and non-real smart-amp together in the same build, can cause problems in some case. Previously this was not possible due to build time error, but with this change, one can build such an image and the UUID will be the same. This is good for a build test, but may cause weird'ish errors on more complicated functional checks when/if the same uuid is used.~~ Update: The real smart-amp uuids are different, so not an issue.

Approving as I still am pretty confident this will work.

Another approach would be to make a Kconfig depends for smart_amp_tests so that they could not be enabled if the a real smart-amp is enabled in the build.

cujomalainey (approved):

cujomalainey:
Thanks for fixing up my half bake

andrula-song (approved):

dbaluta:
Thanks for fixing this :) I wanted to send a patch on Friday but couldn't understand exactly what names should I use.

PR #8442, opened 2023-11-03
tools/topology1: mass replace codec_master -> codec_provider

marc-hb:
This removes a lot of warning noise and may help spot other warnings like `ALSA lib ops.c:48:(lookup_ops) wrong kcontrol ops value string ''` and others.

Now that commit 1b1ec6c4920f ("topology/cmake: skip all topologies when alsatplg < 1.2.5") enforces alsatplg version >= 1.2.5, it is finally safe to do this!

I compared the generated .tplg files before/after this commit and they're bit for bit identical. Tested with both alsatplg v2.6.1 from Ubuntu 22 and version 1.2.9 from the SOF docker container.

Note this is a much smaller commit than the reverted commit f50b6fe0adc4 ("topologies: switch all .m4 files to codec_provider and codec_consumer") because most topologies v1 have been removed from the main branch. Most topologies v1 are now only in stable-v2.2 where the warning spam is still a problem :-(


marc-hb:
https://github.com/thesofproject/sof/actions/runs/6750662401/job/18353468993?pr=8442 is unrelated, fixed by #8441

ranj063 (approved):

marc-hb:
v2:
- fix conflict with recent and unrelated b9bd2303bd0d
- added `tools/test/topology/` "forgotten" in v1 (cause it's not built by default)

CBala21 (approved):

kv2019i (approved):

PR #8446, opened 2023-11-06
Add support for One Shot DMA configuration

bandamadhu:
Support for One Shot DMA configuration for AMD Platforms.

### DIFF
--- a/src/audio/Kconfig
+++ b/src/audio/Kconfig
@@ -725,4 +725,10 @@ config COMP_MODULE_SHARED_LIBRARY_BUILD
 	  Select if you want to build modules as shared objects that can be used to run
 	  pipelines on the host with the testbench or the ALSA plugin.
 
+config DISABLE_DESCRIPTOR_SPLIT
+	bool "Disable descriptor split"
+	default n
+	help
+	  This option disbale the descriptor split for host p-table.
+
 endmenu
--- a/src/audio/host-legacy.c
+++ b/src/audio/host-legacy.c
@@ -180,7 +180,7 @@ static uint32_t host_get_copy_bytes_one_shot(struct host_data *hd, struct comp_d
 	copy_bytes = ALIGN_DOWN(copy_bytes, hd->dma_copy_align);
 
 	split_value = host_dma_get_split(hd, copy_bytes);
-	if (split_value)
+	if (!IS_ENABLED(CONFIG_DISABLE_DESCRIPTOR_SPLIT) && split_value)
 		copy_bytes -= split_value;
 
 	local_elem->size = copy_bytes;
--- a/src/drivers/amd/common/acp_dma.c
+++ b/src/drivers/amd/common/acp_dma.c
@@ -72,6 +72,7 @@ static struct dma_chan_data *acp_dma_channel_get(struct dma *dma,
 
 	acp_dma_chan->config[req_chan].rd_size = 0;
 	acp_dma_chan->config[req_chan].wr_size = 0;
+	acp_dma_chan->config[req_chan].size = 0;
 	return channel;
 }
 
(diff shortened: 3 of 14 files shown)


sofci:

src/arch/xtensa/configs/rembrandt_defconfig
```diff
@@ -33,3 +33,4 @@ CONFIG_COMP_TDFB=n
 ##CONFIG_COMP_MUX=n
 CONFIG_COMP_SEL=n
 CONFIG_COMP_MIXER=n
+CONFIG_DISABLE_DESCRIPTOR_SPLIT=n
```
lyakh, line 36: these aren't needed, since the option is disabled by default, are they?

src/audio/Kconfig
```diff
@@ -725,4 +725,10 @@ config COMP_MODULE_SHARED_LIBRARY_BUILD
 	  Select if you want to build modules as shared objects that can be used to run
 	  pipelines on the host with the testbench or the ALSA plugin.
 
+config DISABLE_DESCRIPTOR_SPLIT
```
lyakh, line 728: unless there's a really good reason, it's usually easier to work with "positive" options, i.e. if you had `ENABLE_DESCRIPTOR_SPLIT` you wouldn't need to handle double negations

src/audio/host-legacy.c
```diff
@@ -180,8 +180,11 @@ static uint32_t host_get_copy_bytes_one_shot(struct host_data *hd, struct comp_d
 	copy_bytes = ALIGN_DOWN(copy_bytes, hd->dma_copy_align);
 
 	split_value = host_dma_get_split(hd, copy_bytes);
-	if (split_value)
+	if (split_value) {
+#ifndef CONFIG_DISABLE_DESCRIPTOR_SPLIT
```
lyakh, line 184: `if (!IS_ENABLED(CONFIG_DISABLE_DESCRIPTOR_SPLIT) && split_value)` but in fact you don't need the whole `split_value` variable is splitting is disabled

src/audio/Kconfig
```diff
@@ -725,4 +725,10 @@ config COMP_MODULE_SHARED_LIBRARY_BUILD
...
+	bool "Disable descriptor split"
+	default n
```
lyakh, line 730: does it depend on `!FORCE_DMA_COPY_WHOLE_BLOCK`?

src/arch/xtensa/configs/rembrandt_defconfig
bandamadhu: Correct not needed, will remove!

src/audio/Kconfig
bandamadhu: YES CORRECT, but for AMD platfroms in this scenario we are disabling or bypassing discriptor split, thats why it names
as DISABLE_DESCRIPTOR_SPLIT, this is meaningful too here.

src/audio/Kconfig
bandamadhu: NO, this cahnge required for only AMD platfroms & for AMD platforms it is not enable FORCE_DMA_COPY_WHOLE_BLOCK.

src/audio/host-legacy.c
bandamadhu: Yes, will it modify.

src/audio/host-legacy.c
```diff
@@ -180,7 +180,7 @@ static uint32_t host_get_copy_bytes_one_shot(struct host_data *hd, struct comp_d
 	copy_bytes = ALIGN_DOWN(copy_bytes, hd->dma_copy_align);
 
 	split_value = host_dma_get_split(hd, copy_bytes);
-	if (split_value)
+	if (!IS_ENABLED(CONFIG_DISABLE_DESCRIPTOR_SPLIT) && split_value)
```
btian1, line 183: #if !defined(CONFIG_DISABLE_DESCRIPTOR_SPLIT)
	if (split_value)
		copy_bytes -= split_value;
#endif

more easy and quick to understand?

bandamadhu: if we use this logic, it will throw build error in defined case
Error msg:
src/audio/host-legacy.c:169:11: error: variable 'split_value' set but not used [-Werror=unused-but-set-variable]

bhiregoudar (approved):

sunilkumardommati (approved):

kv2019i:
It would seem there are no blocking issues. I'll refresh the SOF CI runs to get a clean run.

src/audio/Kconfig
kv2019i: @bandamadhu I think that aligns with what @lyakh proposes. Adding a "depends on !FORCE_DMA_COPY_WHOLE_BLOCK" ensure this option is not enabled in configurations where it does not work. On AMD platforms this check will be true and DISABLE_DESCRIPTOR_SPLIT can be selected.

PR #8448, opened 2023-11-06
waves: store config blob in a cache in waves.c

barry-waves:
Store/apply config blob in a cache to avoid that
cfg.data will be released after prepare.

### DIFF
--- a/src/audio/module_adapter/module/waves/waves.c
+++ b/src/audio/module_adapter/module/waves/waves.c
@@ -41,7 +41,8 @@ struct waves_codec_data {
 	uint32_t                response_max_bytes;
 	uint32_t                request_max_bytes;
 	void                    *response;
-	struct module_config setup_cfg;
+	uint32_t                config_blob_size;
+	void                    *config_blob;
 };
 
 enum waves_codec_params {
@@ -456,6 +457,51 @@ static int waves_effect_revision(struct processing_module *mod)
 	return 0;
 }
 
+/* cache config blob*/
+static int waves_effect_save_config_blob_to_cache(struct processing_module *mod,
+	void *data, uint32_t size)
+{
+	struct comp_dev *dev = mod->dev;
+	struct module_data *codec = &mod->priv;
+	struct waves_codec_data *waves_codec = codec->private;
+
+	comp_info(dev, "waves_effect_save_config_blob_to_cache() start");
+
+	/* release old cached config blob*/
+	if (waves_codec->config_blob && size != waves_codec->config_blob_size) {
+		comp_info(dev, "waves_effect_save_config_blob_to_cache() release blob");
+		module_free_memory(mod, waves_codec->config_blob);
+		waves_codec->config_blob = NULL;
+		waves_codec->config_blob_size = 0;
+	}
+
+	if (!waves_codec->config_blob) {
+		waves_codec->config_blob = module_allocate_memory(mod, size, 16);
+		if (!waves_codec->config_blob) {
+			comp_err(dev,
+				"waves_effect_save_config_blob_to_cache() failed to allocate %d bytes for config blob",
+				size);
+			return -ENOMEM;
+		}
+		waves_codec->config_blob_size = size;
+	}
+
+	int ret = memcpy_s(waves_codec->config_blob, waves_codec->config_blob_size,
+			data, size);
+	if (ret) {
+		comp_err(dev,
+			"waves_effect_save_config_blob_to_cache(): failed to copy config blob %d",
+			ret);
+		module_free_memory(mod, waves_codec->config_blob);
+		waves_codec->config_blob = NULL;
+		waves_codec->config_blob_size = 0;
+		return ret;
+	}
+
+	comp_dbg(dev, "waves_effect_save_config_blob_to_cache() done");
+	return 0;
+}
+
 /* apply MaxxEffect message */
 static int waves_effect_message(struct processing_module *mod, void *data, uint32_t size)
 {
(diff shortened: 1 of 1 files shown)


src/audio/module_adapter/module/waves/waves.c
```diff
@@ -456,6 +458,50 @@ static int waves_effect_revision(struct processing_module *mod)
...
+			memset(waves_codec->config_blob, 0, size);
+			ret = memcpy_s(waves_codec->config_blob, waves_codec->config_blob_size,
+						data, size);
+			if (ret) {
+				comp_err(dev,
+						"waves_effect_cache_config_blob(): failed to copy config blob %d",
+						ret);
+				module_free_memory(mod, waves_codec->config_blob);
+				waves_codec->config_blob_size = 0;
+				return ret;
```
btian1, line 497: replace line 497 with 
ret = -ENOMEM;

this can: 
1. keep same with if part.
2. one return in line 502 is enough for this case.

src/audio/module_adapter/module/waves/waves.c
```diff
@@ -659,6 +719,23 @@ static int waves_codec_init(struct processing_module *mod)
...
+	struct comp_dev *dev = mod->dev;
+	struct module_data *codec = &mod->priv;
+	struct waves_codec_data *waves_codec = codec->private;
+	int ret = 0;
+
+	comp_dbg(dev, "waves_effect_apply_config_blob_cache()");
+
+	if (waves_codec->config_blob) {
+		ret = waves_effect_message(mod, waves_codec->config_blob,
+								waves_codec->config_blob_size);
```
btian1, line 734: line 734 should align with line 733

src/audio/module_adapter/module/waves/waves.c
```diff
@@ -679,6 +756,9 @@ static int waves_codec_prepare(struct processing_module *mod,
 	if (!ret)
 		ret = waves_effect_setup_config(mod);
 
+	if (!ret)
+		ret = waves_effect_apply_config_blob_cache(mod);
+
```
btian1, line 761: this is not an effective way for such logic, take example
if line 748 return 1, then with current code logic, it have to continue check 5 times ret before real return.
suggest use goto label for each return.

also, line 762 -765 also have logic error, 765 always print.
need else before line 765.

src/audio/module_adapter/module/waves/waves.c
```diff
@@ -456,6 +458,50 @@ static int waves_effect_revision(struct processing_module *mod)
...
+		waves_codec->config_blob_size = 0;
+	}
+
+	if (!waves_codec->config_blob) {
+		waves_codec->config_blob = module_allocate_memory(mod, size, 16);
+		if (!waves_codec->config_blob) {
+			comp_err(dev,
+				"waves_effect_cache_config_blob() failed to allocate %d bytes for config blob",
+				size);
+			ret = -ENOMEM;
```
lyakh, line 485: I'd just do `return -ENOMEM` here (and below) and also avoid a `done` debug print in an error case, remove the `ret` variable

src/audio/module_adapter/module/waves/waves.c
```diff
@@ -456,6 +458,50 @@ static int waves_effect_revision(struct processing_module *mod)
...
+					"waves_effect_cache_config_blob(): failed to copy config blob %d",
+					ret);
+				module_free_memory(mod, waves_codec->config_blob);
+				waves_codec->config_blob_size = 0;
+				ret = -ENOMEM;
+			}
+		}
+	}
+	comp_dbg(dev, "waves_effect_cache_config_blob() done");
+	return ret;
```
lyakh, line 502: and just `return 0` here

src/audio/module_adapter/module/waves/waves.c
```diff
@@ -659,6 +719,23 @@ static int waves_codec_init(struct processing_module *mod)
...
+{
+	struct comp_dev *dev = mod->dev;
+	struct module_data *codec = &mod->priv;
+	struct waves_codec_data *waves_codec = codec->private;
+	int ret = 0;
+
+	comp_dbg(dev, "waves_effect_apply_config_blob_cache()");
+
+	if (waves_codec->config_blob) {
+		ret = waves_effect_message(mod, waves_codec->config_blob,
```
lyakh, line 733: `return waves_effect_message(...)` here and `return 0` below

lyakh (approved):

btian1 (approved):

src/audio/module_adapter/module/waves/waves.c
```diff
@@ -456,6 +458,48 @@ static int waves_effect_revision(struct processing_module *mod)
...
+	if (!waves_codec->config_blob) {
+		waves_codec->config_blob = module_allocate_memory(mod, size, 16);
+		if (!waves_codec->config_blob) {
+			comp_err(dev,
+				"waves_effect_cache_config_blob() failed to allocate %d bytes for config blob",
+				size);
+			return -ENOMEM;
+		}
+		waves_codec->config_blob_size = size;
+		memset(waves_codec->config_blob, 0, size);
```
lyakh, line 487: please, remove. Makes no sense clearing the buffer before completely overwriting it

lyakh (changes requested):

src/audio/module_adapter/module/waves/waves.c
```diff
@@ -659,6 +717,22 @@ static int waves_codec_init(struct processing_module *mod)
 	return ret;
 }
 
+/* apply config blob */
+static int waves_effect_apply_config_blob_cache(struct processing_module *mod)
```
johnylin76, line 721: I beg to state that the commit complicates the design. After that there will be both waves_effect_apply_config() and waves_effect_apply_config_cache(). Moreover, waves_effect_config(), waves_effect_cache_config(), and waves_effect_message().

(Please correct me if there is devil in the details.) I do believe the logic is fair enough as I mentioned in the former comment. That is, two reaction chain in total which can be triggered from external:
1. set config blob from external:
```
(...) -> waves_codec_set_configuration() -> waves_effect_apply_config() -> waves_effect_config_save_to_cache()
```
2. start module (trigger prepare) from external: 
```
(...) -> waves_codec_prepare() -> waves_effect_config_load_from_cache()
```
then handle what if...
   - cache is empty when calling `waves_effect_config_load_from_cache()`?
   - module is processing when calling `waves_effect_config_save_to_cache()`?

src/audio/module_adapter/module/waves/waves.c
```diff
@@ -41,7 +41,9 @@ struct waves_codec_data {
 	uint32_t                response_max_bytes;
 	uint32_t                request_max_bytes;
 	void                    *response;
-	struct module_config setup_cfg;
+	struct module_config    setup_cfg;
```
johnylin76, line 44: I feel certain that setup_cfg is no longer used

src/audio/module_adapter/module/waves/waves.c
```diff
@@ -456,6 +458,48 @@ static int waves_effect_revision(struct processing_module *mod)
...
+		}
+	}
+	waves_codec->config_blob_size = size;
+	int ret = memcpy_s(waves_codec->config_blob, waves_codec->config_blob_size,
+			data, size);
+	if (ret) {
+		comp_err(dev,
+			"waves_effect_cache_config_blob(): failed to copy config blob %d",
+			ret);
+		module_free_memory(mod, waves_codec->config_blob);
```
lyakh, line 494: need to set `waves_codec->config_blob` to `NULL` too

src/audio/module_adapter/module/waves/waves.c
```diff
@@ -669,21 +743,31 @@ static int waves_codec_prepare(struct processing_module *mod,
...
+	if (ret)
+		goto error;
 
+	ret = waves_effect_apply_config_blob_cache(mod);
 	if (ret)
-		comp_err(dev, "waves_codec_prepare() failed %d", ret);
+		goto error;
 
 	comp_dbg(dev, "waves_codec_prepare() done");
 	return ret;
```
lyakh, line 766: this can be `return 0` now

src/audio/module_adapter/module/waves/waves.c
johnylin76: +1 from my outdated comment https://github.com/thesofproject/sof/pull/8448#discussion_r1387899199

btw github is so not friendly to bury unaddressed comments and label them as outdated when the new-updated commit may change the relation.

src/audio/module_adapter/module/waves/waves.c
```diff
@@ -492,37 +536,62 @@ static int waves_effect_message(struct processing_module *mod, void *data, uint3
...
+	comp_info(dev, "waves_effect_apply_config_blob_from_cache()");
+
+	if (waves_codec->config_blob) {
+		return waves_effect_message(mod, waves_codec->config_blob,
+				waves_codec->config_blob_size);
+	}
+	return 0;
+}
+
+static int waves_effect_handle_param_message(struct processing_module *mod,
```
johnylin76, line 555: How would the waves library handle the case that client calls MaxxEffect_Message() during processing? For example, the normal case:
```
prepare()               process() *n                   reset()
MaxxEffect_Message() -> MaxxEffect_Process() *n ... -> MaxxEffect_Reset()
```
the specific case that the config setting is interrupted during the process sequence (the interrupted config is valid and could be either equal to the existing one or not):
```
prepare()               process() *n               apply_config()          process() *n                   reset()
MaxxEffect_Message() -> MaxxEffect_Process() *n -> MaxxEffect_Message() -> MaxxEffect_Process() *n ... -> MaxxEffect_Reset()
```
If that can be handled properly (no risk for producing processing errors or sound distortion) then it's fine; otherwise, you may consider to de-couple "save-to-cache" and "apply-from-cache", i.e. only "save-to-cache" on handle_param_message() and "apply-from-cache" on prepare(), which prevents the interrupt while deferring the new config to the next prepare()

barry-waves: Hi Johny,
It would not produce processing errors or sound distortion. The "apply-from-cache" will call "waves_effect_message()", it will let our tuner to hear the sound change immediately during audio playback.

johnylin76: I see. It's fine as long as that is the intended behavior.

lyakh (approved):

johnylin76:
> @johnylin76 are all review comments aligned with the latest update ?

Yes. Thanks

kv2019i:
SOFCI TEST

PR #8451, opened 2023-11-06
schedule: cancel a task before freeing it

### LINKED ISSUE #8445: Stopping/Pausing a CHAIN_DMA stream takes 100ms
Adding a Linux issue based on https://github.com/thesofproject/sof-test/pull/1126 comments

"ChainDMA enabled PCMs where the time to stop the stream takes more than 100ms (from trigger stop to PCM close)."

this needs to be root-caused, 100ms is not an acceptable response time.

@ujfalusi @lgirdwood we might transfer this issue to SOF at some point.

lyakh:
When a chain DMA stream is the only one running, stopping it currently takes 100ms. The reason is that the chain DMA driver is trying to stop the task directly from the IPC context instead of letting the task exit on its next scheduling event. The IPC thread is indeed trying to wait on a semaphore for the task to exit, but that doesn't work either, because zephyr_domain_unregister() terminates the thread before it signals the semaphore. To fix this we swap the order and signal the semaphore before terminating the thread. Fixes #8445

### DIFF
--- a/src/schedule/zephyr_ll.c
+++ b/src/schedule/zephyr_ll.c
@@ -57,6 +57,8 @@ static void zephyr_ll_assert_core(const struct zephyr_ll *sch)
 static void zephyr_ll_task_done(struct zephyr_ll *sch,
 				struct task *task)
 {
+	struct zephyr_ll_pdata *pdata = task->priv_data;
+
 	list_item_del(&task->list);
 
 	if (!sch->n_tasks) {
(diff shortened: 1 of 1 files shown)


marcinszkudlinski (approved):

abonislawski (approved):

src/schedule/zephyr_ll.c
```diff
@@ -57,6 +57,8 @@ static void zephyr_ll_assert_core(const struct zephyr_ll *sch)
 static void zephyr_ll_task_done(struct zephyr_ll *sch,
```
kv2019i, line 57: The patch prefix is a bit odd as this is not touching any chain-dma code.

kv2019i (approved):
@lyakh Minor note on the commit prefix, but otherwise looks good.

src/schedule/zephyr_ll.c
lyakh: @kv2019i oops, yeah, a left-over from the original version... Too bad this cannot be changed in a "low impact" way - without re-running the CI. But I guess yes, it's wrong enough to deserve a re-push

kv2019i: @lyakh let's change. we have not been perfect with the prefixes for sure, but I think we set a better example here

PR #8452, opened 2023-11-06
audio: google_rtc_audio_processing: Extract audio formats from base c…

ranj063:
…onfig extn

The reference audio format should be extracted from the base config extension. The current code works only by accident when the input and reference capture use the same format.

### DIFF
--- a/src/audio/google/google_rtc_audio_processing.c
+++ b/src/audio/google/google_rtc_audio_processing.c
@@ -40,6 +40,7 @@
 #include <google_rtc_audio_processing_sof_message_reader.h>
 
 #define GOOGLE_RTC_AUDIO_PROCESSING_FREQENCY_TO_PERIOD_FRAMES 100
+#define GOOGLE_RTC_NUM_INPUT_PINS 2
 
 LOG_MODULE_REGISTER(google_rtc_audio_processing, CONFIG_SOF_LOG_LEVEL);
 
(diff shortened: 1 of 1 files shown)


src/audio/google/google_rtc_audio_processing.c
```diff
@@ -385,9 +390,21 @@ static int google_rtc_audio_processing_init(struct processing_module *mod)
 	md->private = cd;
 
 #if CONFIG_IPC_MAJOR_4
-	struct module_config *cfg = &md->cfg;
+	const struct ipc4_base_module_extended_cfg *base_cfg = md->cfg.init_data;
+	struct ipc4_input_pin_format reference_fmt, output_fmt;
+	const size_t size = sizeof(struct ipc4_input_pin_format);
+
+	memcpy_s(&cd->config.base_cfg, sizeof(struct ipc4_base_module_cfg),
+		&base_cfg->base_cfg, sizeof(struct ipc4_base_module_cfg));
```
btian1, line 398: any concern you don't want to use below code?
cd->config.base_cfg = base_cfg->base_cfg;

afraid inside pointers?

cujomalainey (approved):
LGTM although I ditto @btian1's comment about just using the assign

src/audio/google/google_rtc_audio_processing.c
ranj063: yeah that would work as well. fixed now

btian1 (approved):

yongzhi1 (approved):

PR #8456, opened 2023-11-08
[NOT REVIEW] copier: add support for windows driver

RanderWang:
Windows driver always set sample type of MSB for 24/32 format but SOF FW supports 24/32 LSB type for non-copier modules. So FW will convert the input MSB 24/32 to LSB 24/32 and process it and convert it back to MSB 24/32 when exiting gtw.


RanderWang:
@dnikodem  can you help to check this PR with windows ? thanks

lgirdwood (commented):
@RanderWang does this mean we need to do extra or differet conversions for windows in the copier ? i.e. no net gain in MCPS.

RanderWang:
> @RanderWang does this mean we need to do extra or differet conversions for windows in the copier ? i.e. no net gain in MCPS.

As you know, Windows has a different layout of 24/32 compared to SOF implementation. Windows always set MSB_TYPE all the modules while Linux topology set correct type for each module, so we need to **adjust** format to choose the correct conversion for this case. It will not affect performance since the conversion function is the same as Linux but the sample type is not fit for SOF.

lgirdwood:
> > @RanderWang does this mean we need to do extra or differet conversions for windows in the copier ? i.e. no net gain in MCPS.
> 
> As you know, Windows has a different layout of 24/32 compared to SOF implementation. Windows always set MSB_TYPE all the modules while Linux topology set correct type for each module, so we need to **adjust** format to choose the correct conversion for this case. It will not affect performance since the conversion function is the same as Linux but the sample type is not fit for SOF.

Thank you for clarifying - just a different conversion in FW with no MCPS impact. Lets wait on @dnikodem review now.

src/audio/copier/copier_generic.c
```diff
@@ -274,10 +274,39 @@ pcm_converter_func get_converter_func(const struct ipc4_audio_format *in_fmt,
...
-		out_valid = SOF_IPC_FRAME_S24_4LE_MSB;
+	if (in_fmt->s_type == IPC4_TYPE_MSB_INTEGER && in_valid == SOF_IPC_FRAME_S24_4LE) {
+		switch (type) {
+		case ipc4_gtw_host:
+			if (dir == ipc4_playback)
+				in_valid = SOF_IPC_FRAME_S24_4LE_MSB;
+			break;
+		case ipc4_gtw_alh:
+		case ipc4_gtw_link:
+		case ipc4_gtw_ssp:
```
dnikodem, line 285: Missing gtw_dmic type

src/audio/copier/copier_generic.c
```diff
@@ -274,10 +274,39 @@ pcm_converter_func get_converter_func(const struct ipc4_audio_format *in_fmt,
...
+
+	if (out_fmt->s_type == IPC4_TYPE_MSB_INTEGER && out_valid == SOF_IPC_FRAME_S24_4LE) {
+		switch (type) {
+		case ipc4_gtw_host:
+			if (dir == ipc4_capture)
+				out_valid = SOF_IPC_FRAME_S24_4LE_MSB;
+			break;
+		case ipc4_gtw_alh:
+		case ipc4_gtw_link:
+		case ipc4_gtw_ssp:
```
dnikodem, line 302: missing gtw_dmic type

dnikodem (changes requested):

src/audio/copier/copier_generic.c
RanderWang: Thanks, if you have validate all the cases, I can start to refine it now

RanderWang:
Thanks for @dnikodem's hard work on validation on windows platforms. Pass the test, so it is ready for review.

src/audio/copier/copier_generic.c
```diff
@@ -274,10 +274,45 @@ pcm_converter_func get_converter_func(const struct ipc4_audio_format *in_fmt,
...
 
-	if (in_fmt->s_type == IPC4_TYPE_MSB_INTEGER && in_valid == SOF_IPC_FRAME_S24_4LE)
-		in_valid = SOF_IPC_FRAME_S24_4LE_MSB;
-	if (out_fmt->s_type == IPC4_TYPE_MSB_INTEGER && out_valid == SOF_IPC_FRAME_S24_4LE)
-		out_valid = SOF_IPC_FRAME_S24_4LE_MSB;
+	/* use MSB sample type to select conversion function if the data is enter or exit dsp.
+	 * In playback case, host input and dai output and in capture case, host output and
+	 * dai input.
+	 */
+	if (in_fmt->s_type == IPC4_TYPE_MSB_INTEGER && in_valid == SOF_IPC_FRAME_S24_4LE) {
```
btian1, line 281: this condition looks strange， in_fmt is MSB integer, however, its valid bit is 24 with LSB?

RanderWang: Thanks Windows

lgirdwood (approved):

lgirdwood:
Seeing some cluster failures with nocodec, lets be sure.

lgirdwood:
SOFCI TEST

RanderWang:
Multicore failure. It will be fixed by https://github.com/thesofproject/sof/pull/8515

lgirdwood:
> Multicore failure. It will be fixed by #8515

Now merged - will rerun the CI.

lgirdwood:
SOFCI TEST

RanderWang:
ignore sof-ci/jenkins/pr-device-test/main-cavs2.5 which is a deprecated test item for old PR and not used by latest PR.
The failure case sof-ci/jenkins/pr-device-test/main-cavs  : multiple-pause-resume-50.sh  and  check-suspend-resume-with-capture.sh  are not related to this PR

src/audio/copier/copier_generic.c
```diff
@@ -274,10 +274,45 @@ pcm_converter_func get_converter_func(const struct ipc4_audio_format *in_fmt,
...
+		case ipc4_gtw_link:
+		case ipc4_gtw_ssp:
+		case ipc4_gtw_dmic:
+			if (dir == ipc4_playback)
+				out_valid = SOF_IPC_FRAME_S24_4LE_MSB;
+			break;
+		default:
+			break;
+		}
+	}
```
btian1, line 315: can we combine two switch case to one and move two if() inside case?

lgirdwood: lets take that as an optimization for followup in another PR.

lgirdwood (approved):

PR #8457, opened 2023-11-08
TDFB code reorganize

btian1:
move it from include path to tdfb module folder.


src/audio/tdfb/tdfb.c
```diff
@@ -15,7 +15,6 @@
 #include <sof/audio/ipc-config.h>
```
singalsu, line 15: Commit text - use Capital letters in sentences start.

src/audio/tdfb/tdfb.h
```diff
@@ -17,6 +17,15 @@
 #define SOF_TDFB_MAX_ANGLES 360		/* Up to 1 degree precision for 360 degrees coverage */
 #define SOF_TDFB_MAX_MICROPHONES 16	/* Up to 16 microphone locations */
 
+/* The driver assigns running numbers for control index. If there's single control of
+ * type switch, enum, binary they all have index 0.
+ */
+#define CTRL_INDEX_PROCESS		0	/* switch */
```
singalsu, line 23: These would be good to prefix with TDFB_.

src/audio/tdfb/tdfb_ipc3.c
```diff
@@ -0,0 +1,229 @@
...
+#if TDFB_ADD_DIRECTION_TO_GET_CMD
+	cd->ctrl_data->chanv[0].channel = 0;
+	cd->ctrl_data->chanv[0].value = cd->az_value_estimate;
+	cd->ctrl_data->num_elems = 1;
+#endif
+
+	ipc_msg_send(cd->msg, cd->ctrl_data, false);
+}
+
+int tdfb_sub_init(struct processing_module *mod)
```
singalsu, line 62: The new function name is not descriptive, it should describe better what is done here.

src/audio/tdfb/tdfb_ipc4.c
```diff
@@ -0,0 +1,198 @@
...
+static void tdfb_send_notification(struct ipc_msg *msg, uint32_t val)
+{
+	struct tdfb_notification_payload *ipc_payload;
+
+	ipc_payload = (struct tdfb_notification_payload *)msg->tx_data;
+	ipc_payload->control_value.value = val;
+	ipc_msg_send(msg, NULL, false);
+}
+
+int tdfb_sub_init(struct processing_module *mod)
```
singalsu, line 86: Update this function name too.  Perhaps tdfb_common_notification_init()?

src/audio/tdfb/tdfb_ipc4.c
```diff
@@ -0,0 +1,198 @@
...
+void tdfb_params(struct processing_module *mod)
+{
+	struct sof_ipc_stream_params *params = mod->stream_params;
+	struct comp_buffer *sinkb, *sourceb;
+	struct comp_dev *dev = mod->dev;
+
+	ipc4_base_module_cfg_to_stream_params(&mod->priv.cfg.base_cfg, params);
+	component_set_nearest_period_frames(dev, params->rate);
+
+	sourceb = list_first_item(&dev->bsource_list, struct comp_buffer, sink_list);
```
singalsu, line 192: I realized here's similar issue as there was with my work with rtc aec. This is similar as original, but it's not handling correctly a typical case where source has more channels than sink. Please add a reminder /* TODO:  */ note for it. I can fix it later when I have suitable topology to use for test.

src/audio/tdfb/tdfb_ipc4.c
```diff
@@ -0,0 +1,198 @@
...
+}
+
+void tdfb_send_notification_common(struct processing_module *mod)
+{
+	struct tdfb_comp_data *cd = module_get_private_data(mod);
+
+	tdfb_send_notification(cd->msg, cd->az_value_estimate);
+}
+
+int tdfb_get_config_common(struct processing_module *mod,
```
lyakh, line 107: sorry, not sure why you call these functions `*_common()`? They differ between IPC3 and IPC4, I would use "common" for functions shared between IPC versions?

lgirdwood (approved):

src/audio/tdfb/tdfb_ipc4.c
lgirdwood: ack, sounds like they are not common, otherwise it LGTM. Lets fix the ipc/common naming.

src/audio/tdfb/tdfb_ipc4.c
cujomalainey: @lkoenig FYI

src/audio/tdfb/tdfb_ipc4.c
btian1: both ipc3/ipc4 have this function implementation, my purpose is:
keep interface function defined in main source file, i.e, keep: tdfb_get_config & tdfb_set_config in tdfb.c.

Otherwise, I can move tdfb_get_config & tdfb_get_config to ipc3/ipc4 separately, this will make tdfb.c looks lack something

lyakh: the code separation is good, we're only proposing to change the function naming. "common" in this case suggests that those functions are *common* for IPC3 and IPC4. The code in tdfb.c is common for IPC3 and IPC4. But functions, implemented separately for IPC3 and IPC4 are IPC-version-specific. I'm not sure I can propose a good name suffix for those version-specific function parts, how about `..._private` or `..._priv` or `..._specific`?

btian1: no perfect name actually, let's keep for now?

lyakh: > no perfect name actually, let's keep for now?

personally each of the proposed options seems better to me than "common," e.g. I'd go with "priv"

kv2019i: @btian1 @lyakh Hmm, I agree. "common" is very misleading as the functions actually are different and not common. Naming is hard, let's go with "priv", not great, but good enough.

kv2019i (changes requested):
Looks good @btian1 . Let's change the "common" to "priv" and then I think this is good to go.

src/audio/tdfb/tdfb_ipc4.c
```diff
@@ -0,0 +1,198 @@
...
+
+	return 0;
+}
+
+void tdfb_send_notification_common(struct processing_module *mod)
+{
+	struct tdfb_comp_data *cd = module_get_private_data(mod);
+
+	tdfb_send_notification(cd->msg, cd->az_value_estimate);
+}
```
lyakh, line 105: two more non-"_common" functions

singalsu (approved):
Looks good to me, thanks for the clean up!

src/audio/tdfb/tdfb_ipc4.c
singalsu: I recall I proposed some functions names change, I wonder what would be better names @lyakh for the layers needed for ipc3 and ipc4 handling?

singalsu: Ah, there was same discussion yesterday. @kv2019i proposed "priv".

src/audio/tdfb/tdfb_comp.h
```diff
@@ -158,6 +144,41 @@ static inline void tdfb_cdec_s16(int16_t **ptr, int16_t *start, size_t size)
...
+}
+#endif /* CONFIG_FORMAT_S24LE */
+#if CONFIG_FORMAT_S32LE
+static inline void set_s32_fir(struct tdfb_comp_data *cd)
+{
+	cd->tdfb_func = tdfb_fir_s32;
+}
+#endif /* CONFIG_FORMAT_S32LE */
+
+int tdfb_common_notification_init(struct processing_module *mod);
```
kv2019i, line 171: There's still common here, but let me propose something better:
tdfb_ipc_notification_init()

src/audio/tdfb/tdfb_comp.h
```diff
@@ -158,6 +144,41 @@ static inline void tdfb_cdec_s16(int16_t **ptr, int16_t *start, size_t size)
...
+void tdfb_send_notification_common(struct processing_module *mod);
```
kv2019i, line 172: tdfb_send_ipc_notification()

src/audio/tdfb/tdfb_comp.h
```diff
@@ -158,6 +144,41 @@ static inline void tdfb_cdec_s16(int16_t **ptr, int16_t *start, size_t size)
...
+int tdfb_get_private_config(struct processing_module *mod,
```
kv2019i, line 173: tdfb_get_ipc_config()

src/audio/tdfb/tdfb_comp.h
```diff
@@ -158,6 +144,41 @@ static inline void tdfb_cdec_s16(int16_t **ptr, int16_t *start, size_t size)
...
+			    uint32_t param_id, uint32_t *data_offset_size,
+			    uint8_t *fragment, size_t fragment_size);
+int tdfb_set_private_config(struct processing_module *mod, uint32_t param_id,
```
kv2019i, line 176: tdfb_set_ipc_config()

kv2019i (changes requested):
Still a few not-common common functions. While at it, I propose a better alternative to "private". Especially if "private" is put in the middle of the function (like get_private_params()), this implies we are getting some private parameters (i..e attribute of params, and not attribution of the function itself).

src/audio/tdfb/tdfb_ipc4.c
singalsu: The merged fix for it is here: https://github.com/thesofproject/sof/pull/8452

lyakh (approved):

kv2019i (approved):
Thanks @btian1 , looks good now.

PR #8462, opened 2023-11-08
DP: Fix dp period calculation for DP modules

jxstelter:
The OBS parameter for module could be set to higher value
than the one defined by stream output format. This could result
in wrong calculation of period parameter.
Due to this issue we observe SRC tests regression on LNL platform.

This patch adds output stream format to period calculation in addition
to the one calculated from OBS. Lower value is selected.
Also period is round down to represent integer multiple of LL ticks.

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -19,6 +19,7 @@
 #include <sof/audio/sink_source_utils.h>
 #include <sof/audio/dp_queue.h>
 #include <sof/audio/pipeline.h>
+#include <sof/schedule/ll_schedule_domain.h>
 #include <sof/common.h>
 #include <sof/platform.h>
 #include <sof/ut.h>
@@ -240,9 +241,17 @@ static int module_adapter_dp_queue_prepare(struct comp_dev *dev)
 			 &sink_buffer->stream.runtime_stream_params,
 			 sizeof(sink_buffer->stream.runtime_stream_params));
 		/* calculate time required the module to provide OBS data portion - a period */
-		unsigned int sink_period = 1000000 * sink_get_min_free_space(mod->sinks[i]) /
+		uint32_t min_free_size =
+			sink_get_frame_bytes(mod->sinks[i]) *
+			(sink_get_rate(mod->sinks[i]) / (USEC_PER_SEC / LL_TIMER_PERIOD_US));
+		min_free_size = MIN(min_free_size, sink_get_min_free_space(mod->sinks[i]));
+
+		unsigned int sink_period = 1000000 * min_free_size /
 					   (sink_get_frame_bytes(mod->sinks[i]) *
 					   sink_get_rate(mod->sinks[i]));
+		/* Round down period to integer number of LL ticks */
+		sink_period = ROUND_DOWN(sink_period, LL_TIMER_PERIOD_US);
+
 		/* note the minimal period for the module */
 		if (period > sink_period)
 			period = sink_period;
--- a/src/audio/src/src_ipc4.c
+++ b/src/audio/src/src_ipc4.c
@@ -16,6 +16,7 @@
 #include <sof/audio/sink_api.h>
 #include <sof/audio/source_api.h>
 #include <sof/audio/sink_source_utils.h>
+#include <sof/schedule/ll_schedule_domain.h>
 #include <rtos/panic.h>
 #include <sof/ipc/msg.h>
 #include <rtos/alloc.h>
@@ -107,9 +108,17 @@ int src_set_params(struct processing_module *mod, struct sof_sink *sink)
 	 * as SRC uses period value to calculate its internal buffers,
 	 * it must be done here, right after setting sink parameters
 	 */
-	if (dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_DP)
-		dev->period = 1000000 * sink_get_min_free_space(sink) /
+	if (dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_DP) {
+		uint32_t min_free_size =
+			sink_get_frame_bytes(sink) * (sink_get_rate(sink) /
+					(USEC_PER_SEC / LL_TIMER_PERIOD_US));
+		min_free_size = MIN(min_free_size, sink_get_min_free_space(sink));
+
+		dev->period = 1000000 * min_free_size /
 		      (sink_get_frame_bytes(sink) * sink_get_rate(sink));
+		/* Round down period to integer number of LL ticks */
+		dev->period = ROUND_DOWN(dev->period, LL_TIMER_PERIOD_US);
+	}
 
 	comp_info(dev, "SRC DP period calculated as: %u", dev->period);
 


src/audio/module_adapter/module_adapter.c
```diff
@@ -240,9 +241,17 @@ static int module_adapter_dp_queue_prepare(struct comp_dev *dev)
...
+			sink_get_frame_bytes(mod->sinks[i]) *
+			(sink_get_rate(mod->sinks[i]) / LL_TIMER_PERIOD_US);
+		min_free_size = MIN(min_free_size, sink_get_min_free_space(mod->sinks[i]));
+
+		unsigned int sink_period = 1000000 * min_free_size /
 					   (sink_get_frame_bytes(mod->sinks[i]) *
 					   sink_get_rate(mod->sinks[i]));
+		/* Round down period to integer number of LL ticks */
+		sink_period = (sink_period / LL_TIMER_PERIOD_US) * LL_TIMER_PERIOD_US;
+
```
btian1, line 254: you may not need so much complicated calculation, please check:
https://github.com/thesofproject/sof/pull/8412/commits/aa04f1d385c5f778bb2cacb3ad759fb1fb6c7abd
src obs are simplified, this is due to actual buffer size = obs * 2, so do not need left 5 words redundancy.

src/audio/module_adapter/module_adapter.c
```diff
@@ -240,9 +241,17 @@ static int module_adapter_dp_queue_prepare(struct comp_dev *dev)
 			 &sink_buffer->stream.runtime_stream_params,
 			 sizeof(sink_buffer->stream.runtime_stream_params));
 		/* calculate time required the module to provide OBS data portion - a period */
-		unsigned int sink_period = 1000000 * sink_get_min_free_space(mod->sinks[i]) /
+		uint32_t min_free_size =
+			sink_get_frame_bytes(mod->sinks[i]) *
+			(sink_get_rate(mod->sinks[i]) / LL_TIMER_PERIOD_US);
```
lyakh, line 246: this doesn't look correct. Apparently you're trying to calculate number of bytes per period. `sink_get_rate()` is the number of frames per second, `LL_TIMER_PERIOD_US` is the number of microseconds per period. Since the latter is usually 1000, it happens to work, but what you really need is to divide the number of frames per second by the number of periods per second, i.e.
```
sink_get_rate(mod->sinks[i]) / (MICROSECONDS_PER_SECOND / LL_TIMER_PERIOD_US)
```

lyakh (changes requested):

src/audio/module_adapter/module_adapter.c
jxstelter: You are right, but there are configurations that sets OBS to value which gives period as a fraction of LL timer period (eg. from reference firmware). To keep compatibility I added this round down calculation.

src/audio/module_adapter/module_adapter.c
jxstelter: fixed

src/audio/src/src_ipc4.c
```diff
@@ -107,9 +108,17 @@ int src_set_params(struct processing_module *mod, struct sof_sink *sink)
 	 * as SRC uses period value to calculate its internal buffers,
 	 * it must be done here, right after setting sink parameters
 	 */
-	if (dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_DP)
-		dev->period = 1000000 * sink_get_min_free_space(sink) /
+	if (dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_DP) {
```
lgirdwood, line 111: @jxstelter @singalsu this looks the same as the previous patch in this PR. i.e. we could make this a helper API call to avoid duplication (but can be done as an improvement in another PR).

lgirdwood (approved):

tmleman (approved):

src/audio/module_adapter/module_adapter.c
```diff
@@ -240,9 +241,17 @@ static int module_adapter_dp_queue_prepare(struct comp_dev *dev)
...
+		uint32_t min_free_size =
+			sink_get_frame_bytes(mod->sinks[i]) *
+			(sink_get_rate(mod->sinks[i]) / (USEC_PER_SEC / LL_TIMER_PERIOD_US));
+		min_free_size = MIN(min_free_size, sink_get_min_free_space(mod->sinks[i]));
+
+		unsigned int sink_period = 1000000 * min_free_size /
 					   (sink_get_frame_bytes(mod->sinks[i]) *
 					   sink_get_rate(mod->sinks[i]));
+		/* Round down period to integer number of LL ticks */
+		sink_period = (sink_period / LL_TIMER_PERIOD_US) * LL_TIMER_PERIOD_US;
```
lyakh, line 253: `ROUND_DOWN()`

src/audio/module_adapter/module_adapter.c
```diff
@@ -240,9 +241,17 @@ static int module_adapter_dp_queue_prepare(struct comp_dev *dev)
 			 &sink_buffer->stream.runtime_stream_params,
 			 sizeof(sink_buffer->stream.runtime_stream_params));
 		/* calculate time required the module to provide OBS data portion - a period */
-		unsigned int sink_period = 1000000 * sink_get_min_free_space(mod->sinks[i]) /
+		uint32_t min_free_size =
+			sink_get_frame_bytes(mod->sinks[i]) *
+			(sink_get_rate(mod->sinks[i]) / (USEC_PER_SEC / LL_TIMER_PERIOD_US));
+		min_free_size = MIN(min_free_size, sink_get_min_free_space(mod->sinks[i]));
+
+		unsigned int sink_period = 1000000 * min_free_size /
```
lyakh, line 249: can this overflow? If `min_free_size` is a bit more than 4K, it will overflow already?

src/audio/src/src_ipc4.c
```diff
@@ -107,9 +108,17 @@ int src_set_params(struct processing_module *mod, struct sof_sink *sink)
...
+		uint32_t min_free_size =
+			sink_get_frame_bytes(sink) * (sink_get_rate(sink) /
+					(USEC_PER_SEC / LL_TIMER_PERIOD_US));
+		min_free_size = MIN(min_free_size, sink_get_min_free_space(sink));
+
+		dev->period = 1000000 * min_free_size /
 		      (sink_get_frame_bytes(sink) * sink_get_rate(sink));
+		/* Round down period to integer number of LL ticks */
+		dev->period = (dev->period / LL_TIMER_PERIOD_US) * LL_TIMER_PERIOD_US;
```
lyakh, line 120: same comments for overflow and `ROUND_DOWN()`

jxstelter: fixed

src/audio/module_adapter/module_adapter.c
jxstelter: I think there is no case where the minimal size will be close to 4K value.

src/audio/module_adapter/module_adapter.c
```diff
@@ -240,9 +241,17 @@ static int module_adapter_dp_queue_prepare(struct comp_dev *dev)
 			 &sink_buffer->stream.runtime_stream_params,
 			 sizeof(sink_buffer->stream.runtime_stream_params));
 		/* calculate time required the module to provide OBS data portion - a period */
-		unsigned int sink_period = 1000000 * sink_get_min_free_space(mod->sinks[i]) /
+		uint32_t min_free_size =
```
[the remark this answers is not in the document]
jxstelter: > @jxstelter , please specify which converter have problem. This is quite tricky for set sink period, @singalsu have a table to show different produce. it may not easy to set here, check below 48 -->88.2 example: image
> 
> max is 98, we may need more discuss for SRC period setting, it is more complex than we imagine.

Currently the issue is observed only on LNL FPGA setup. This is because on FPGA there is much lower MCPS margin available. In that case simply Copier->SRC->Copier pipeline is affected.  When sink_period is set for too high value, DP module get processing time later than it should and on FPGA it can not finish processing on time, causing glitches. 
The problem is not observed on RVP yet, however it could be observed when core utilization will be much higher.

src/audio/module_adapter/module_adapter.c
btian1: take 192khz, 32 bit 8 ch with 10ms period data as example:
192 * 4 * 8 * 10 = 61440, even 1ms data size is 6144.

lgirdwood (commented):
@jxstelter is this pending more testing before non draft ?

jxstelter:
> @jxstelter is this pending more testing before non draft ?

Yes. Actually all tests passed with this change, but we agreed with @marcinszkudlinski, that the resolution should be different. This is WIP.

src/audio/module_adapter/module_adapter.c
btian1: this is in prepare, so it is one time, no worry on performance.

PR #8468, opened 2023-11-09
sink_src: add ipc4 update format functions

marcinszkudlinski:
sink/src API must grow to provide similar functionality to comp_buffers.

This commit adds 2 helpers for IPC4 format set


RanderWang (approved):
LGTM.  The source and sink are very similar, can we unify them ?

lyakh:
I still prefer adding code only when it's needed, not to provide similar functionality.

kv2019i (approved):

marcinszkudlinski:
@lyakh  - yes it is needed, look at 8469

PR #8466, opened 2023-11-09
idc: check if secondary core is down before sending IDC

marcinszkudlinski:
If for any reason a secondary core is down, IDC message has no chance to be processed. This may lead to process hang in case of blocking calls and to undefined actions in case of non blocking.

This commit adds a check and error log message in case of target core is down.

PR #8469, opened 2023-11-09
[DNM] Google aec sink src dp

marcinszkudlinski:
This PR modifies google AEC shimm to use sink/src interface and to be run as DP module in 10ms period

it contains all required not merged changes as a squash commit

**it still contains a bit to do - do not merge**

- [x] https://github.com/thesofproject/sof/pull/8468
- [x] https://github.com/thesofproject/sof/pull/8467
- [x] https://github.com/thesofproject/sof/pull/8466
- [x] https://github.com/thesofproject/sof/pull/8433
- [x] https://github.com/thesofproject/sof/pull/8636
- [ ] https://github.com/thesofproject/sof/pull/8575
- [ ] https://github.com/thesofproject/sof/pull/8668


src/audio/google/google_rtc_audio_processing.c
```diff
@@ -333,6 +336,7 @@ static int google_rtc_audio_processing_init(struct processing_module *mod)
 
 	cd->config.reference_fmt = reference_fmt.audio_fmt;
 	cd->config.output_fmt = output_fmt.audio_fmt;
+	cd->config = *(const struct sof_ipc4_aec_config *)cfg->init_data;
```
lyakh, line 339: this overwrites all previous assignments of individual `cd->config.*` fields, can they all be removed?

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -382,36 +386,31 @@ static int google_rtc_audio_processing_init(struct processing_module *mod)
...
-	cd->raw_mic_buffer = rballoc(
-		0, SOF_MEM_CAPS_RAM,
-		cd->num_frames * cd->num_capture_channels * sizeof(cd->raw_mic_buffer[0]));
-	if (!cd->raw_mic_buffer) {
+	buf_size = cd->num_frames * cd->num_capture_channels * sizeof(cd->process_buffer[0]);
+	comp_dbg(dev, "Allocating process_buffer of size %u", buf_size);
+	cd->process_buffer = rballoc(0, SOF_MEM_CAPS_RAM, buf_size);
+	if (!cd->process_buffer) {
+		comp_err(dev, "Allocating process_buffer failure");
 		ret = -EINVAL;
```
lyakh, line 394: can fix it to `-ENOMEM` while at it

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -480,29 +476,38 @@ static int google_rtc_audio_processing_prepare(struct processing_module *mod,
 
 	comp_info(dev, "google_rtc_audio_processing_prepare()");
 
-	google_rtc_audio_processing_params(mod);
+	if (num_of_sources != GOOGLE_RTC_NUM_INPUT_PINS) {
+		comp_err(dev, "Expecting 2 sources - ref and mic, got %u", num_of_sources);
```
lyakh, line 480: Let's use the macro to print how many were expected

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -480,29 +476,38 @@ static int google_rtc_audio_processing_prepare(struct processing_module *mod,
...
+		return -EINVAL;
+	}
+
+	if (num_of_sinks != GOOGLE_RTC_NUM_OUTPUT_PINS) {
+		comp_err(dev, "Expecting 1 sink, got %u", num_of_sinks);
```
lyakh, line 485: ditto

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -546,13 +551,36 @@ static int google_rtc_audio_processing_prepare(struct processing_module *mod,
...
+			  sink_get_min_free_space(sinks[0]),
+			  cd->num_frames * sink_get_frame_bytes(sinks[0]));
+	}
+	if (cd->num_frames * source_get_frame_bytes(sources[cd->aec_reference_source]) !=
+	    source_get_min_available(sources[cd->aec_reference_source])) {
+		comp_warn(dev, "Incorrect IBS on reference source: %d, expected %u",
+			  source_get_min_available(sources[cd->aec_reference_source]),
+			  cd->num_frames *
+				source_get_frame_bytes(sources[cd->aec_reference_source]));
+	}
```
lyakh, line 574: a bit strange - we say that inputs are wrong, but we don't fail and don't fix them?

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -563,125 +591,191 @@ static int google_rtc_audio_processing_reset(struct processing_module *mod)
...
+	for (int i = 0; i < cd->num_frames; i++) {
+		for (channel = 0; channel < cd->num_aec_reference_channels; ++channel) {
+			cd->aec_reference_buffer_ptrs[channel][i] =
+					convert_uint16_to_float_hifi(ref[channel]);
 		}
-		num_samples_remaining -= n;
-		ref = audio_stream_wrap(ref_stream, ref);
+		ref += cd->num_aec_reference_channels;
+		if ((void *)ref >= (void *)ref_buf_end)
+			ref = (void *)ref_buf_start;
```
lyakh, line 686: I guess this won't be a lot, but we could optimise this a bit: remove this `if` in the loop and calculate first till the potential end of the buffer, and then wrap, or use an encompassing loop as in other modules. Same for source and sink below

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -563,125 +591,191 @@ static int google_rtc_audio_processing_reset(struct processing_module *mod)
...
+					       struct sof_source **sources, int num_of_sources,
+					       struct sof_sink **sinks, int num_of_sinks)
 {
-	struct google_rtc_audio_processing_comp_data *cd = module_get_private_data(mod);
-	int16_t *src, *dst, *ref;
-	uint32_t num_aec_reference_frames;
-	uint32_t num_aec_reference_bytes;
+
+	int ret;
+	uint16_t const *src;
```
andrula-song, line 625: shouldn't these pointer be signed? our samples are signed.

src/audio/google/google_rtc_audio_processing_mock.c
```diff
@@ -140,46 +138,40 @@ int GoogleRtcAudioProcessingReconfigure(GoogleRtcAudioProcessingState *const sta
...
+	int n, chan, ref_chan;
+
+	for (chan = 0; chan < state->num_output_channels; chan++) {
+		for (n = 0; n < state->num_frames; ++n) {
+			float mic_save = mic[chan][n];	/* allow same in/out buffer */
+
+			if (chan < state->num_aec_reference_channels)
+				dest[chan][n] = mic_save + ref[n + (chan * state->num_frames)];
+			else
+				dest[chan][n] = mic_save;
```
andrula-song, line 156: would move this if statement out of the second for loop be faster?
```
for (chan = 0; chan < state->num_output_channels; chan++) {
        if (chan < state->num_aec_reference_channels) 
		for (n = 0; n < state->num_frames; ++n) 
			dest[chan][n] = mic[chan][n] + ref[n + (chan * state->num_frames)];
       else
		for (n = 0; n < state->num_frames; ++n) 
			dest[chan][n] = mic[chan][n];
}
```

src/audio/google/google_rtc_audio_processing.c
andrula-song: from my test before, wrap would be faster than we check in the for loop.

src/audio/google/google_rtc_audio_processing_mock.c
lyakh: ...or even swapping the loops and splitting the one over channels into two loops:
```
for (n = 0; n < state->num_frames; ++n) {
    for (chan = 0; chan < state->num_aec_reference_channels; chan++)
        ...
    for (chan = state->num_aec_reference_channels; state->num_output_channels; chan++)
        ...
}
```

marcinszkudlinski: this is mock, test only, nobody cares for performance ;)

src/audio/google/google_rtc_audio_processing.c
marcinszkudlinski: incorrect merge, to be fixed

src/audio/google/google_rtc_audio_processing.c
marcinszkudlinski: yes, long story. 
As description says, the PR is not completely finished

src/audio/google/google_rtc_audio_processing_mock.c
cujomalainey: Agreed, mock is mostly for testing or bringup without the actual block, readability over performance here

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -94,19 +93,16 @@ static void google_rtc_audio_processing_params(struct processing_module *mod)
...
-	sinkb = list_first_item(&dev->bsink_list, struct comp_buffer, source_list);
-	ipc4_update_buffer_format(sinkb, &mod->priv.cfg.base_cfg.audio_fmt);
+	/* TODO - it does not work, to be checked before merging!!
+	 * ipc4_update_source_format(ref, &cd->config.reference_fmt);
+	 */
+	source_set_channels(ref,
+				CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_NUM_AEC_REFERENCE_CHANNELS);
+	/* ipc4_update_source_format(mic, &mod->priv.cfg.base_cfg.audio_fmt); */
+	source_set_channels(mic, CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_NUM_CHANNELS);
+	/* ipc4_update_sink_format(out, &mod->priv.cfg.base_cfg.audio_fmt); */
```
lkoenig, line 103: Why aren't we using the config from the stream as number of channels ?
We could use the stream parameter to configure the number of channels on the AEC.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -640,125 +591,191 @@ static int google_rtc_audio_processing_reset(struct processing_module *mod)
...
+
+		src += cd->num_capture_channels;
+		if ((void *)src >= (void *)src_buf_end)
+			src = (void *)src_buf_start;
+	}
+
+	source_release_data(src_stream, num_of_bytes_to_process);
+
+	/* call the library, use same in/out buffers */
+	GoogleRtcAudioProcessingProcessCapture_float32(cd->state,
```
lkoenig, line 715: Are we sure the playback and the capture are in sync: ie they get 10ms of data at the same time ? I see nothing in the code ensuring it is the case: I might have missed it.
Otherwise we take care of that by calling the `AnalyseRender` only when 10ms of playback data, and only call the `ProcessCapture` when we have 10ms of raw microphone data.

src/audio/google/google_rtc_audio_processing_mock.c
```diff
@@ -140,46 +138,40 @@ int GoogleRtcAudioProcessingReconfigure(GoogleRtcAudioProcessingState *const sta
 	return 0;
 }
 
-int GoogleRtcAudioProcessingProcessCapture_int16(GoogleRtcAudioProcessingState *const state,
-						 const int16_t *const src,
-						 int16_t *const dest)
```
lkoenig, line 145: We can keep both: Both exists in the library.

src/audio/google/google_rtc_audio_processing_mock.c
```diff
@@ -140,46 +138,40 @@ int GoogleRtcAudioProcessingReconfigure(GoogleRtcAudioProcessingState *const sta
...
-
-		ref += state->num_aec_reference_channels;
-		out += state->num_output_channels;
-		mic += state->num_capture_channels;
 	}
 	return 0;
 }
 
-int GoogleRtcAudioProcessingAnalyzeRender_int16(GoogleRtcAudioProcessingState *const state,
-						const int16_t *const data)
```
lkoenig, line 175: Both exisyts in the library and can be mixed.

lkoenig (commented):
This is just the more pressing question I have.

src/audio/google/google_rtc_audio_processing.c
marcinszkudlinski: this PR is not yet finished
Currently - for unknown reasons - the number of channels at inputs are set incorrectly and I has to be overwritten. Root cause is under investigation.

lgirdwood:
@marcinszkudlinski any update ?

src/audio/google/google_rtc_audio_processing.c
lkoenig: What if 10ms playback is ready 9ms after 10ms capture are ready ? That will leave a deadline of 1ms wouldn't it ?

src/audio/google/google_rtc_audio_processing_mock.c
lkoenig: The library accepts both float32 between -1 and 1 and 16bits signed PCM. It uses another format internally so it would make sense to not do too much conversion outside and pass the data as straight forward as possible.

src/audio/google/google_rtc_audio_processing.c
lkoenig: Not to mentioned the case when the rates are not exactly the same on render and capture (I think we can omit this case for now).

marcinszkudlinski:
> @marcinszkudlinski any update ?

I'm waiting for feedback from 007 branch.

src/audio/module_adapter/module_adapter.c
```diff
@@ -1068,24 +1071,33 @@ static int module_adapter_copy_dp_queues(struct comp_dev *dev)
 		dp_queue = dp_queue_get_next_item(dp_queue);
 	}
 
-	dp_queue = dp_queue_get_first_item(&mod->dp_queue_dp_to_ll_list);
-	list_for_item(blist, &dev->bsink_list) {
+	if (!mod->DP_startup_delay) {
```
lyakh, line 1074: please do
```
if (mod->DP_startup_delay)
    return 0;
```
instead  and it should be `dp_startup_delay` lower-case too.

src/schedule/zephyr_dp_schedule.c
```diff
@@ -35,10 +35,12 @@ struct scheduler_dp_data {
 
 struct task_dp_pdata {
 	k_tid_t thread_id;		/* zephyr thread ID */
-	uint32_t period_clock_ticks;	/* period the task should be scheduled in Zephyr ticks */
+	uint32_t deadline_clock_ticks;	/* dp module deadline in Zephyr ticks */
+	uint32_t deadline_ll_cycles;	/* dp module deadline in LL cycles */
 	k_thread_stack_t __sparse_cache *p_stack;	/* pointer to thread stack */
 	struct k_sem sem;		/* semaphore for task scheduling */
 	struct processing_module *mod;	/* the module to be scheduled */
+	uint32_t LL_cycles_to_deadline;  /* current number of LL cycles till deadline */
```
lyakh, line 43: lower case

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -91,17 +93,15 @@ static void google_rtc_audio_processing_params(struct processing_module *mod)
...
-			ipc4_update_buffer_format(sourceb, &mod->priv.cfg.base_cfg.audio_fmt);
-	}
-
-	sinkb = list_first_item(&dev->bsink_list, struct comp_buffer, source_list);
-	ipc4_update_buffer_format(sinkb, &mod->priv.cfg.base_cfg.audio_fmt);
+	/* TODO - it does not work, to be checked before merging!!
+	 * ipc4_update_source_format(ref, &cd->config.reference_fmt);
+	 */
+	source_set_channels(ref,
+				CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_NUM_AEC_REFERENCE_CHANNELS);
```
lyakh, line 100: either align, or one line

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -563,125 +591,191 @@ static int google_rtc_audio_processing_reset(struct processing_module *mod)
...
+static float convert_uint16_to_float_hifi(int16_t data)
+{
+	const xtfloat ratio = 2 << 14;
+	xtfloat x0 = data;
+	float x;
+
+	x = XT_DIV_S(x0, ratio);
+
+	return x;
+}
```
lyakh, line 616: AFAIK we usually make HiFi code depend on availability devinitions perhaps something like `XCHAL_HAVE_HIFI3` and also provide pure C alternatives

marcinszkudlinski: no need here, this is a shmm for 3rd party lib that is using HiFI and will always be compiled using clang

marc-hb: > using HiFI and will always be compiled using clang

HiFI and clang are two totally different things, don't mix them up, it's confusing. For instance this #8637 _requires_ clang (it literally can't compile with GCC) and it absolutely can't use HiFI.

I  think you mean "closed-source Cadence Xtensa toolchain" (or any shorter version of this.

marc-hb: > I think you mean "compiled only by closed-source Cadence Xtensa toolchain" (or any shorter version of this).

And this is in fact wrong:

https://github.com/thesofproject/sof/actions/runs/7226339953/job/19691685718?pr=8637

```
[208/259] Building modules_sof.dir/home/workspace/sof/src/audio/drc/drc_generic.c.obj
[209/259] Building modules_sof.dir/home/workspace/sof/src/audio/google/google_rtc_audio_processing_mock.c.obj
[210/259] Building modules_sof.dir/home/workspace/sof/src/audio/igo_nr/igo_nr_stub.c.obj
[211/259] Building modules_sof.dir/home/workspace/sof/src/audio/multiband_drc/multiband_drc.c.obj
[212/259] Building modules_sof.dir/home/workspace/sof/src/audio/google/google_rtc_audio_processing.c.obj

```

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -90,3 +90,6 @@ CONFIG_DEBUG_COREDUMP_MEMORY_DUMP_MIN=y
 
 CONFIG_PROBE=y
 CONFIG_PROBE_DMA_MAX=2
+
+CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING=y
+CONFIG_COMP_STUBS=y
```
marc-hb, line 95: Missing newline, please configure your editor.

src/audio/google/Kconfig
```diff
@@ -16,6 +16,7 @@ config COMP_GOOGLE_RTC_AUDIO_PROCESSING
 	bool "Google Real Time Communication Audio processing"
 	select COMP_BLOB
 	select GOOGLE_RTC_AUDIO_PROCESSING_MOCK if COMP_STUBS
+	depends on IPC_MAJOR_4
```
marc-hb, line 19: @cujomalainey, @andyross I think this is how this PR excluded  `google_rtc_audio_processing.c` from the compilation check: 

https://github.com/thesofproject/sof/actions/runs/6815375008/job/18534330110?pr=8469

```
warning: COMP_GOOGLE_RTC_AUDIO_PROCESSING (defined at
/home/runner/work/sof/sof/workspace/sof/zephyr/../src/audio/google/Kconfig:15) was assigned the
value 'y' but got the value 'n'. Check these unsatisfied dependencies: IPC_MAJOR_4 (=n)
```

`CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING=y` is in `app/stub_build_all_ipc3.conf` but not in `app/stub_build_all_ipc4.conf`. I've noticed other differences between these two .conf files; you may want to bring them somewhat back in sync again.

marcinszkudlinski:
closing by now, a new PR will be opened soon

PR #8470, opened 2023-11-09
schedule: zephyr_dma_domain: Do not use a registered channel

dbaluta:
Current implementation only checks if a channel is used with current interrupt. But that channel might have been registered with another interrupt earlier.

So, we need a way to skip channels already used. For this, we iterate over all registered interrupts and check existing registered channels.

This fixes scenarios where running:

$ arecord -Dhw:0,0 -f S32 -c 2 -r 48000 -t raw |
aplay -Dhw:0,0 -f S32_LE -c 2 -r 48000

will result in an I/O error.


LaurentiuM1234 (approved):

TangleZ:
Looks good to me.

TangleZ (approved):

PR #8474, opened 2023-11-10
Audio: DRC: Add processing enable switch control

singalsu:


singalsu:
@johnylin76 @cujomalainey Does adding this control and the logic of letting the control to impact to look OK to you? The operation should be as before if topology is not defining a switch - the blob will determine enabled / disabled.

When there is a switch control in topology, processing is enabled only if blob has enable set and user space is set to enabled. This is the behaviour what I'm not fully sure about. Other way would be to enable the processing with the blob definitions whenever user requests it even if blob has enable set to false.

For lowest disabled MCPS I was testing also a version that uses bypass copy function (drc_default_pass) for disabled but there would be a click when enabled/disabled so I dropped for now the idea.

Note that multiband-DRC is somewhat different in enable switch control handling. It's not changing enabled/disable during streaming. I'd like to update it to be similar as what is defined for DRC to they would be kind of plug-in replacements for each other. DRC would be the light alternative for multiband-DRC (that runs min. 2 band filterbank always).

src/audio/drc/drc.c
```diff
@@ -235,6 +266,16 @@ static void drc_set_alignment(struct audio_stream *source,
...
 }
 
+static bool drc_has_config(struct drc_comp_data *cd)
+{
+	return cd->config ? true : false;
+}
+
+static bool drc_get_config_enabled(struct drc_comp_data *cd)
+{
+	return cd->config->params.enabled ? true : false;
```
lyakh, line 276: this is boolean, so can just `return cd->config->params.enabled`

src/audio/drc/drc.c
```diff
@@ -261,6 +302,10 @@ static int drc_process(struct processing_module *mod,
 		}
 	}
 
+	/* Control pass-though in processing function with switch control */
+	if (drc_has_config(cd))
+		cd->enabled = (drc_get_config_enabled(cd) && cd->enable_switch) ? true : false;
```
lyakh, line 307: `&&` is a logical operation, its result is already a `true` or a `false`, so a simple `cd->enabled = drc_get_config_enabled(cd) && cd->enable_switch` isenough

src/audio/drc/drc.c
```diff
@@ -261,6 +302,10 @@ static int drc_process(struct processing_module *mod,
 		}
 	}
 
+	/* Control pass-though in processing function with switch control */
+	if (drc_has_config(cd))
```
johnylin76, line 306: Not sure if drc_has_config() is necessary. In my point of view, the code can be
```
static bool drc_get_config_enabled(struct drc_comp_data *cd)
{
	return (cd->config) && (cd->config->params.enabled);
}
...
	/* Control pass-through in processing function with switch control */
	cd->enabled = drc_get_config_enabled(cd) && cd->enable_switch;

```
The present DRC is designed with the assumption that cd->config is always valid while in processing. The enable flag in parameter is static in our use cases, while considering the switch control support the in-param flag can be treated as the default state, and the switch flag is the dynamic signal to unmask the state if needed.

If cd->config is invalid, it makes more senses that DRC should be in disabled (pass-through) state. This is not handled by this commit.

johnylin76 (commented):
It looks good to me only with one code suggestion. Thanks

src/audio/drc/drc.c
```diff
@@ -235,6 +266,16 @@ static void drc_set_alignment(struct audio_stream *source,
 	audio_stream_init_alignment_constants(1, 1, sink);
 }
 
+static bool drc_has_config(struct drc_comp_data *cd)
```
btian1, line 269: "static inline"

lyakh: on ACE and cAVS the compiler is now allowed to inline automatically. If this is an issue on other platforms, I'd rather enable `CONFIG_COMPILER_INLINE_FUNCTION_OPTION` on them too, than adding `inline` to individual functions.

lgirdwood (commented):
LGTM when all current comments resolved.

src/audio/drc/drc.c
```diff
@@ -202,15 +209,39 @@ static int drc_free(struct processing_module *mod)
...
+	struct comp_dev *dev = mod->dev;
+
+	comp_dbg(dev, "drc_set_config()");
+
+#if CONFIG_IPC_MAJOR_4
+	struct sof_ipc4_control_msg_payload *ctl = (struct sof_ipc4_control_msg_payload *)fragment;
+
+	switch (param_id) {
+	case SOF_IPC4_SWITCH_CONTROL_PARAM_ID:
+		if (ctl->id == 0 && ctl->num_elems == 1) {
```
lyakh, line 227: shouldn't control IDs be macros?

singalsu: Yep, it's a good idea

cujomalainey:
@johnylin76 will adding additional controls like this to EQ/MDRC cause any overhead in your offload logic?

lgirdwood:
@wszypelt @lrudyX CI been pending a while. Tests passing ? (dont think DRC is tested on Internal CO today).

src/audio/drc/drc.c
```diff
@@ -261,6 +298,9 @@ static int drc_process(struct processing_module *mod,
 		}
 	}
 
+	/* Control pass-though in processing function with switch control */
+	cd->enabled = drc_get_config_enabled(cd) && cd->enable_switch;
```
btian1, line 302: better to directly as:
cd->enabled = cd->config && cd->config->params.enabled
                         && cd->enable_switch;

lyakh (approved):

wszypelt:
@lgirdwood  
Our current CI is all green.
As for DRC, the module's integration was not synchronized with validation so unfortunately at the moment we do not have any scenarios for this module.

lgirdwood:
@singalsu one conflict.

src/audio/drc/drc.c
```diff
@@ -183,6 +183,13 @@ static int drc_init(struct processing_module *mod)
 	}
 
 	drc_reset_state(&cd->state);
+
+	/* Initialize DRC to enabled. If defined by topology, a control may set
+	 * enabled to false before prepare() or during streaming with the switch
+	 * control from user space.
+	 */
+	cd->enabled = true;
+	cd->enable_switch = true;
```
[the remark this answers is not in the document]
singalsu: I don't know. Does kernel re-apply ALSA controls after reset? And how could that be tested, do you know?

btian1 (approved):

PR #8475, opened 2023-11-10
Tools: Tune: DRC: Add calculation of blob for speaker processing

singalsu:
First commit adds m4 blob for testbench, 2nd commit adds same for tplg2, 3rd commit adds same for sof-ctl, 4th commit contains the Matlab code changes to produce blobs.


tools/tune/drc/drc_gen_coefs.m
```diff
@@ -29,6 +29,7 @@
 % Calculate makeup gain coefficients
 full_range_makeup_gain = (1 / coefs.ratio_base) ^ 0.6; % Empirical/perceptual tuning
 coefs.master_linear_gain = db2mag(params.post_gain) * full_range_makeup_gain;
+coefs.master_linear_gain_db = 20*log10(coefs.master_linear_gain);
```
andrula-song, line 33: why do we add master_linear_gain_db?  seems we only use master_linear_gain when processing.

lgirdwood (commented):
LGTM, but is there an instruction/comments on how to change the gains ?

tools/topology/topology1/m4/drc_coef_threshold_-25_knee_15_ratio_10_post_3.m4
```diff
@@ -0,0 +1,21 @@
+# Exported Control Bytes 10-Nov-2023
```
btian1, line 1: can we add comments for this blob usage here?
maybe filename not need so complicated, we can add its usage and specification here.

tools/topology/topology2/include/components/drc/threshold_-25_knee_15_ratio_10_post_3.conf
```diff
@@ -0,0 +1,22 @@
+# Exported Control Bytes 10-Nov-2023
```
btian1, line 1: same here, what is it and how this table was used?

tools/tune/drc/drc_gen_coefs.m
btian1: following code seems does not use this db.

tools/tune/drc/example_drc.m
```diff
@@ -31,6 +31,14 @@ function example_drc()
 params.enabled = 1;
 drc_coefs_and_config_export(params, 'enabled');
 
+% Export experimental configuration for a small speaker
```
johnylin76, line 34: In my view the only purpose for these lines is to generate the config "threshold_-25_knee_15_ratio_10_post_3", they are not supposed to land upstream. Alternatively, the applied param set could be put into the comment together with the generated config files.
Another point is that the generated config will be changed by any parameter which is not specified here, e.g. attack and release.

johnylin76:
> LGTM, but is there an instruction/comments on how to change the gains ?

The DRC design document may help:
https://docs.google.com/document/d/1T8Hr2l6DdL1Wij3S-qNtvwDHBiiprFbUQwxLZsjo69M

one can also check the freq response curve by this GUI website:
http://www.audio-tuning.appspot.com/
The parameter set is aligned with the one specified in example_drc.m

tools/topology/topology1/m4/drc_coef_threshold_-25_knee_15_ratio_10_post_3.m4
singalsu: A generic name like drc_coef_generic_notebook_speaker.m4 / .conf might do after I can decide what are the final parameters to use.

tools/tune/drc/example_drc.m
singalsu: Yep, there's more parameters. I guess I should change the file name, but the intention of this is to become the default notebook speaker (hda-generic PC) DRC so used by their default topology so it should be in upstream. The parameters are still likely to change with more subjective testing. I used these parameters for an internal demo to show how DRC can be used to make a very audible loudness change. The EQ in the pipeline is also part of the enhance effect.

tools/tune/drc/drc_gen_coefs.m
singalsu: I added it to know the dB gain achieved with all the parameters. Due to above empirical/perceptual tuning commented equation it can't be directly seen from the parameters input. I need to add some print to make this used. I guess I checked it with debugger or print of the same code line with `;` removed.

singalsu: Sorry, the master_linear_gain_db is actually printed in line 66, so it is used.

```
% Print out coefs
coefs

```

tools/tune/drc/example_drc.m
johnylin76: Thanks. It should be a good idea to separate them into multiple files, e.g. example_drc_default.m, example_drc_notebook.m, and others for different use cases. I found that the same way is applied under eq/ and tfdb/

johnylin76 (approved):
LGTM with one suggestion left

andrula-song (approved):

tools/tune/drc/example_drc.m
singalsu: I agree, that would be a small and ease maintenance of the tools files.

lgirdwood (approved):

Issue #8016, opened 2023-08-08
[FEATURE] Implement SOF_IPC4_FW_PARAM_SYSTEM_TIME to sync host and FW timestamps


kv2019i:
**Is your feature request related to a problem? Please describe.**
Host is sending SOF_IPC4_FW_PARAM_SYSTEM_TIME to synchronize the host and DSP timestamps so that timestamps in logs can be compared directly. 

**Describe the solution you'd like**
Implement IPC4 handle to set the timestamp epoch and implement a custom timestamp generator and set this using Zephyr log_set_timestamp_func() (in order for logging subsystem to use this timestamp).

**Describe alternatives you've considered**
Setting the Zephyr kernel time is one option, but there is currently no API to do this from applications and this could create many issues (i.e. time going back for kernel code).

**Additional context**

kv2019i:
@btian1 this is one potential, please assign to yourself if you have bandwidth

plbossart:
@ujfalusi and I already discussed this, I can't recall what the outcome was and where this discussion took place.

ujfalusi:
@plbossart , @kv2019i, the kernel is sending the time when the mtrace is enabled.
The message setup starts here:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/sound/soc/sof/ipc4-mtrace.c#n433
and the send is
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/sound/soc/sof/ipc4-mtrace.c#n444

The epoch is adjusted as per specification, but I'm no longer sure if we should do that and just send the time directly.

ujfalusi:
This is the SOF/Zephyr side handling of the message, right?

lgirdwood:
@kv2019i IIUC, driver is sending the message but we need to make sure we get teh message to Zephyr. @abonislawski @mwasko @nashif we will need this for PSS.

kv2019i:
@alex-cri @lgirdwood @mengdonglin proposing to move to v2.8 . work not yet started, so this did not make the v2.7 cut

RanderWang:
[...]

PR #8480, opened 2023-11-14
base_fw: sync time between host and fw for logging

RanderWang:
It is a general implementation for logging and it doesn't use intel audio hardware feature like ART counter. 64bits timerstamp is needed for accurate since the timestamp used by host is beyond 32bits in most cases.

For feature: https://github.com/thesofproject/sof/issues/8016 **Implement SOF_IPC4_FW_PARAM_SYSTEM_TIME to sync host and FW timestamps**

The demo:
FW log:
[18014.307276] <inf> ipc: ipc_cmd: rx	: 0x11000005|0x0
[18014.307286] <inf> pipe: pipeline_new: pipeline new pipe_id 0 priority 0
[18014.307910] <inf> ipc: ipc_cmd: rx	: 0x40000004|0x15
[18014.307990] <inf> dma: dma_get: dma_get() ID 0 sref = 1 busy channels 0
[18014.308556] <inf> ipc: ipc_cmd: rx	: 0x40000006|0x10
[18014.309081] <inf> ipc: ipc_cmd: rx	: 0x44000006|0x30000018
[18014.309613] <inf> ipc: ipc_cmd: rx	: 0x40000002|0xa
[18014.310131] <inf> ipc: ipc_cmd: rx	: 0x11010004|0x0
[18014.310141] <inf> pipe: pipeline_new: pipeline new pipe_id 1 priority 0

Kernel log:
[18014.307175] sof-audio-pci-intel-tgl 0000:00:1f.3: Create widget pipeline.1 instance 0 - pipe 1 - core 0
[18014.307196] sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx      : 0x11000005|0x0: GLB_CREATE_PIPELINE
[18014.307788] sof-audio-pci-intel-tgl 0000:00:1f.3: Create widget host-copier.0.playback instance 0 - pipe 1 - core 0
[18014.307811] sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx      : 0x40000004|0x15: MOD_INIT_INSTANCE [data size: 84]
[18014.308437] sof-audio-pci-intel-tgl 0000:00:1f.3: widget host-copier.0.playback setup complete
[18014.308451] sof-audio-pci-intel-tgl 0000:00:1f.3: Create widget gain.1.1 instance 0 - pipe 1 - core 0
[18014.308466] sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx      : 0x40000006|0x10: MOD_INIT_INSTANCE [data size: 64]
[18014.309003] sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx      : 0x44000006|0x30000018: MOD_LARGE_CONFIG_SET [data size: 
[18014.309501] sof-audio-pci-intel-tgl 0000:00:1f.3: widget gain.1.1 setup complete
[18014.309519] sof-audio-pci-intel-tgl 0000:00:1f.3: Create widget mixin.1.1 instance 0 - pipe 1 - core 0
[18014.309530] sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx      : 0x40000002|0xa: MOD_INIT_INSTANCE [data size: 40]


lgirdwood (approved):
Now the nest step is for the python experts to upstream a small script that takes the kernel log + FW log and muxes them based on the timestamp. @jsarha @marc-hb @aiChaoSONG

kv2019i (approved):
Excellent @RanderWang , looks good!

plbossart (commented):
Sorry, I don't get the concept at all, and not even what this PR tries to fix.

We already have an IPC4 mechanism for passing the kernel time to the firmware, precisely to make sure the logs are aligned. We had a long discussion on this with @ujfalusi 

So what does this PR do that wasn't done before?

kv2019i:
@plbossart wrote:
> We already have an IPC4 mechanism for passing the kernel time to the firmware, precisely to make sure the logs are aligned. We had a long discussion on this with @ujfalusi

Nothing happened before. Kernel sent a message and FW put it to /dev/null

This PR actually implements something to make FW react to said IPC message kernel sends.

plbossart:
> @plbossart wrote:
> 
> > We already have an IPC4 mechanism for passing the kernel time to the firmware, precisely to make sure the logs are aligned. We had a long discussion on this with @ujfalusi
> 
> Nothing happened before. Kernel sent a message and FW put it to /dev/null
> 
> This PR actually implements something to make FW react to said IPC message kernel sends.

Hummm, I am rather puzzled. This time alignment was added for APL back in 2017 or 2018, it's a well-known requirement from the automotive days, and the gregorian Windows time was used. It would be a big miss in the firmware validation if this code was missing.... @mwasko please chime in please.

And the 0.120ms remains a mystery. I can't relate it to any usual time constant.

src/audio/base_fw.c
```diff
@@ -252,6 +263,15 @@ static uint32_t basefw_set_system_time(uint32_t param_id,
 	global_system_time_info.dsp_time.val_l = (uint32_t)(current_dsp_time);
 	global_system_time_info.dsp_time.val_u = (uint32_t)(current_dsp_time >> 32);
 
+	/* use default timestamp if 64bit is not enabled since 64bit is necessary for host time */
+	if (!IS_ENABLED(CONFIG_LOG_TIMESTAMP_64BIT))
+		return IPC4_SUCCESS;
```
marc-hb, line 268: Never silently drop things on the floor. Log something.

src/audio/base_fw.c
```diff
@@ -234,12 +236,21 @@ static int basefw_mem_state_info(uint32_t *data_offset, char *data)
 	return 0;
 }
 
+/* Assumed time interval between host sends timestamp and fw process it */
+#define TIME_SETTING_LATENCY_NS	120000
```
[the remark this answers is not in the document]
marc-hb: Guessing is worse than doing nothing because it will sometimes ahead and sometimes behind, super confusing. Please leave this to zero so the skew is always one way.

Please correct me: in general I think it's pretty obvious that receiving some IPC happened _after_ it was sent, I mean neither IPC3 not IPC4 has a lot of concurrency, has any? 

(also: don't leave 30 lines between the definition and the only use)

src/audio/base_fw.c
```diff
@@ -32,6 +33,7 @@ DECLARE_SOF_RT_UUID("basefw", basefw_comp_uuid, 0xe398c32, 0x5ade, 0xba4b,
 DECLARE_TR_CTX(basefw_comp_tr, SOF_UUID(basefw_comp_uuid), LOG_LEVEL_INFO);
 
 static struct ipc4_system_time_info global_system_time_info;
+static uint64_t time_delta;
```
marc-hb, line 36:
```suggestion
static uint64_t host_cycles_delta;
```

ujfalusi:
> And the 0.120ms remains a mystery. I can't relate it to any usual time constant.

:stopwatch: perhaps?

app/boards/intel_adsp_cavs25_tgph.conf
```diff
@@ -74,3 +74,4 @@ CONFIG_DMA_DW_FIFO_PARTITION=y
 CONFIG_DMA_INTEL_ADSP_GPDMA_HAS_LLP=y
 
 CONFIG_HEAP_MEM_POOL_SIZE=8192
+CONFIG_LOG_TIMESTAMP_64BIT=y
```
marc-hb, line 77: How about changing the default in Kconfig instead?

src/audio/base_fw.c
```diff
@@ -252,6 +263,15 @@ static uint32_t basefw_set_system_time(uint32_t param_id,
...
 	global_system_time_info.dsp_time.val_u = (uint32_t)(current_dsp_time >> 32);
 
+	/* use default timestamp if 64bit is not enabled since 64bit is necessary for host time */
+	if (!IS_ENABLED(CONFIG_LOG_TIMESTAMP_64BIT))
+		return IPC4_SUCCESS;
+
+	host_time = k_ns_to_cyc_ceil64(*(uint64_t *)data + TIME_SETTING_LATENCY_NS);
+	time_delta = host_time - current_dsp_time;
+	log_set_timestamp_func(basefw_get_timestamp,
+			       sys_clock_hw_cycles_per_sec());
```
marc-hb, line 273: What happens when the DSP frequency changes?

Also, what happens on IMR resume?

src/audio/base_fw.c
```diff
@@ -234,12 +236,21 @@ static int basefw_mem_state_info(uint32_t *data_offset, char *data)
...
+	return k_cycle_get_64() + time_delta;
+}
+
 static uint32_t basefw_set_system_time(uint32_t param_id,
 				       bool first_block,
 				       bool last_block,
 				       uint32_t data_offset,
 				       const char *data)
 {
+	uint64_t host_time;
```
marc-hb, line 252:
```suggestion
	uint64_t host_time_cycles;
```

src/audio/base_fw.c
```diff
@@ -234,12 +236,21 @@ static int basefw_mem_state_info(uint32_t *data_offset, char *data)
...
+
```
marc-hb, line 253: > /* TODO: configurate time to logging subsystem */

Done? If not why not?

src/audio/base_fw.c
```diff
@@ -252,6 +263,15 @@ static uint32_t basefw_set_system_time(uint32_t param_id,
 	global_system_time_info.dsp_time.val_l = (uint32_t)(current_dsp_time);
 	global_system_time_info.dsp_time.val_u = (uint32_t)(current_dsp_time >> 32);
 
+	/* use default timestamp if 64bit is not enabled since 64bit is necessary for host time */
+	if (!IS_ENABLED(CONFIG_LOG_TIMESTAMP_64BIT))
+		return IPC4_SUCCESS;
+
+	host_time = k_ns_to_cyc_ceil64(*(uint64_t *)data + TIME_SETTING_LATENCY_NS);
```
marc-hb, line 270: This is very strange: 10 lines earlier `data` is cast to `struct ipc4_system_time`. But here the same `data` is cast to `uint64_t`?

marc-hb (changes requested):
I don't agree with TIME_SETTING_LATENCY_NS, see below.

I don't understand why the same function is setting `global_system_time_info` from `data`, but now using the same `data` in a different way to set up something  totally different and new for logging. Why isn't one thing enough?

marc-hb:
> small script that takes the kernel log + FW log and muxes them based on the timestamp

That would be awesome but the `TIME_SETTING_LATENCY_NS` is actually a very hard problem to solve without a shared, hardware reference (which probably exists somewhere). It's the same reason why [NTP software](https://en.wikipedia.org/wiki/Network_Time_Protocol) is so complex.

> Nothing happened before. Kernel sent a message and FW put it to /dev/null

`global_system_time_info` == /dev/null ?

app/boards/intel_adsp_cavs25_tgph.conf
RanderWang: it is shared by all vendors, so I do it only for intel

src/audio/base_fw.c
RanderWang: The time is queried from wall clock hardware, not based on DSP frequency, so no problem. For IMR resume case, host will set timestamp again and we will sync time again.

src/audio/base_fw.c
RanderWang: sure, thanks

src/audio/base_fw.c
RanderWang: thanks for advice

src/audio/base_fw.c
RanderWang: sure, good idea.

src/audio/base_fw.c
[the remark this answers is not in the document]
RanderWang: this is a common issue that host builds the time and sends to fw which process it a little later. At the time fw receives the timestamp, the timestamp is out of date since host time is going forward. The latency value is built from many experiments.  It is not accurate and universal. I have a better idea to do it in kernel. Let's wait for it.
````
host build timestamp ------>  host send timestamp by ipc msg -------> fw recieve and process it
             |   <-------          60 us        --------> |   <--------       about 60 us      ------>      |
````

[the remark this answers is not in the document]
Comment (RanderWang) [reply src/audio/base_fw.c:240]: please check https://github.com/thesofproject/linux/pull/4706. Kernel will build the latency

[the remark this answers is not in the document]
Comment (plbossart) [reply src/audio/base_fw.c]: You're assuming the IPC latency is symmetric. It's clearly not, the Linux driver deals with a single interrupt and will check stream, IPC and SoundWire (in that order). If the host driver is busy with a SoundWire IPC, the IPC will be handled *after* renabling the global interrupt and going back to the interrupt thread.

On the firmware side, one would hope that Zephyr has a real-time behavior, but IIRC the IPCs are only processed at every firmware tick to avoid disrupting processing, so there some scheduling delay to be added too.

In other words, averaging the IPC latency is just as bad as adding a random fudge factor to the timestamps. The goal is not precise configuration but a means to correlate kernel and firmware. I don't think there's an expectation that there is a better than 1ms alignment, is there?

Back to requirements I guess :-)

File: app/boards/intel_adsp_cavs25_tgph.conf (hunk 4 above)
Comment (marc-hb) [reply app/boards/intel_adsp_cavs25_tgph.conf:77]: Intel is the only one using IPC4 for now and I see nothing Intel-specific so far so please just change  the default in the KConfig file unless there is a specific reason not to have this by default.

File: src/audio/base_fw.c (hunk 8 above)
Comment (marc-hb) [reply src/audio/base_fw.c]: Sorry I didn't make myself clear. My main question is: why is `data` going to two different places? Why isn't one place enough?

**Review by marc-hb — COMMENTED** [review]
> The latency value is built from many experiments. It is not accurate and universal. I have a better idea to do it in kernel

@RanderWang please stop trying to re-invent https://www.intel.com/content/www/us/en/docs/programmable/683410/current/precision-time-protocol-ptp-synchronization.html and similar.

Because of interrupt latencies and other issues it's simply impossible to get something accurate, so keep  things simple and don't add ANY offset, it will make everything simpler to understand and interpret. Maybe log a message clarifying that there is no PTP involved here, that this is just a loose synchronization for mere convenience.

PS: why is `data` going to two different places? Why is one place not enough?

File: src/audio/base_fw.c (hunk 2 above)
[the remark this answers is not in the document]
Comment (RanderWang) [reply src/audio/base_fw.c]: @plbossart  sure, I fully understand your assumption. But consider the real usage: when we sync timestamp ? Driver just do it after all audio initialization is done after powered on, At this time we don't have any IPC or any audio stream, and FW just boot up and have nothing to do.  Yes, it is still not very accurate, but without it we will found FW receive IPC msg before kernel send it. This will confuse developer.  Currently we have a time error of 100 us,  if 1ms error can expected, I am happy to remove the calibration.

File: app/boards/intel_adsp_cavs25_tgph.conf (hunk 4 above)
Comment (RanderWang) [reply app/boards/intel_adsp_cavs25_tgph.conf:77]: I see NXP is working on IPC4 support and reported bugs about intel special code in framework.

File: src/audio/base_fw.c (hunk 8 above)
Comment (RanderWang) [reply src/audio/base_fw.c]: to simplify the code,  I don't want to pack two 32bit number to 64bit one, so reuse data

File: src/audio/base_fw.c [hunk 9]
```diff
@@ -234,24 +236,46 @@ static int basefw_mem_state_info(uint32_t *data_offset, char *data)
...
+
 	if (!(first_block && last_block))
 		return IPC4_INVALID_REQUEST;
 
 	global_system_time_info.host_time.val_l = ((const struct ipc4_system_time *)data)->val_l;
 	global_system_time_info.host_time.val_u = ((const struct ipc4_system_time *)data)->val_u;
 
-	uint64_t current_dsp_time = sof_cycle_get_64();
+	current_dsp_cycle = sof_cycle_get_64();
+	current_dsp_time = k_cyc_to_us_ceil64(current_dsp_cycle);
```
marc-hb, line 261: This is changing the unit of `global_system_time_info.dsp_time`, why?

marc-hb (commented):
The  PR description of commit da5edbdf7d1f73 which added `global_system_time_info` stated "It is then used by FW to translate event timestamps e.g. logs to host system time domain". 


So this PR is adding `time_delta` which duplicates of `global_system_time_info.host_time`, is changing the unit of `global_system_time_info.dsp_time` (?) and seems to be generally  duplicating the da5edbdf7d1f73 feature without really explaining why it should be duplicated.

If what this PR is simpler and better then why not just delete the `global_system_time_info` from da5edbdf7d1f73?

It's very confusing to have both features that seem to duplicate each other - in the same function!

marc-hb:
I had a look at some of the timestamps in https://sof-ci.01.org/sofpr/PR8480/build14946/devicetest/index.html and they never seem to match.

marc-hb:
OMG I just discovered #8016 by chance just now:
- https://github.com/thesofproject/sof/issues/8016

It's incredibly time-consuming to review code when there is the (lack of) explanations do not even include the URL to the corresponding discussion.

src/audio/base_fw.c
RanderWang: both host & dsp time should be in us, not hw cycle.

RanderWang:
> The PR description of commit [da5edbd](https://github.com/thesofproject/sof/commit/da5edbdf7d1f73bfaccca0bfc2e19fe46097b0d0) which added `global_system_time_info` stated "It is then used by FW to translate event timestamps e.g. logs to host system time domain".
> 
> So this PR is adding `time_delta` which duplicates of `global_system_time_info.host_time`, is changing the unit of `global_system_time_info.dsp_time` (?) and seems to be generally duplicating the [da5edbd](https://github.com/thesofproject/sof/commit/da5edbdf7d1f73bfaccca0bfc2e19fe46097b0d0) feature without really explaining why it should be duplicated.
> 
> If what this PR is simpler and better then why not just delete the `global_system_time_info` from [da5edbd](https://github.com/thesofproject/sof/commit/da5edbdf7d1f73bfaccca0bfc2e19fe46097b0d0)?
> 
> It's very confusing to have both features that seem to duplicate each other - in the same function!

 time_delta is not a duplication of global_system_time_info.host_time. global_system_time_info.host_time is for time query but time_delta is for log time adjus

RanderWang:
> OMG I just discovered #8016 by chance just now:
> 
>     * [[FEATURE] Implement SOF_IPC4_FW_PARAM_SYSTEM_TIME to sync host and FW timestamps #8016](https://github.com/thesofproject/sof/issues/8016)
> 
> 
> It's incredibly time-consuming to review code when the (lack of) explanations do not even include the URL to the corresponding discussion.

Thanks, updated.

btian1:
> Now the nest step is for the python experts to upstream a small script that takes the kernel log + FW log and muxes them based on the timestamp. @jsarha @marc-hb @aiChaoSONG

May I ask question here?

1. do you mean mtrace will output sof and kernel logs both based on unified timestamp?
2. if so, do we need set the log output memory larger in order to avoid potential memory overflow for sudden lots of logs?
3. if so, like performance analyzer also need change, currently , it based on two input log files.

RanderWang:
> > Now the nest step is for the python experts to upstream a small script that takes the kernel log + FW log and muxes them based on the timestamp. @jsarha @marc-hb @aiChaoSONG
> 
> May I ask question here?
> 
>     1. do you mean mtrace will output sof and kernel logs both based on unified timestamp?
mtrace only produces fw trace, not kernel logs which is provided by dmesg.   Yes, both fw log & kernel log will use unified timestamp

>     2. if so, do we need set the log output memory larger in order to avoid potential memory overflow for sudden lots of logs?
This is defined by debug memory slots 4KB, we can't change it. you can hack the code to enlarge it.

>     3. if so, like performance analyzer also need change, currently , it based on two input log files.

RanderWang:
@kv2019i  The failure on TGL is caused by SET_DX for secondary core and is not linked to this PR, thanks!

kv2019i:
I think this looks still pretty good. The sync is not designed to be super accurate (there's no actual time sync protocol even attempted here, just one time setting of the epoc from host to DSP). Commit https://github.com/thesofproject/sof/commit/da5edbdf7d1f73bfaccca0bfc2e19fe46097b0d0 is indeed partial, but I don't see why we should block this PR due to that. Latter allows to get the hw-specific counters (ART for Intel) from DSP to host and that functionality is still working after this PR.

src/audio/base_fw.c
marc-hb: > both host & dsp time should be in us, not hw cycle.

If you're fixing a bug in `global_system_time_info.dsp_time`, then it must be described at the very least in the commit message of a separate and earlier commit (and even better: also described in a GitHub issue). Not secretly hidden in the addition of a new feature.

app/boards/intel_adsp_cavs25_tgph.conf
marc-hb: Why would NXP and others not want this with IPC4? Please explain.

src/audio/base_fw.c
marc-hb: Don't resolve comments with no answer.

src/audio/base_fw.c
```diff
@@ -234,24 +236,46 @@ static int basefw_mem_state_info(uint32_t *data_offset, char *data)
...
 	global_system_time_info.dsp_time.val_u = (uint32_t)(current_dsp_time >> 32);
 
+	/* use default timestamp if 64bit is not enabled since 64bit is necessary for host time */
+	if (!IS_ENABLED(CONFIG_LOG_TIMESTAMP_64BIT)) {
+		LOG_WRN("64bits timestamp is disabled, so use default timestamp");
+		return IPC4_SUCCESS;
+	}
+
+	host_time = global_system_time_info.host_time.val_l |
+		((uint64_t)global_system_time_info.host_time.val_u << 32);
```
marc-hb, line 273: Why not just read this directly from `*data` as you were earlier?

marc-hb (commented):
> > global_system_time_info.host_time is for time query but time_delta is for log time adjust

This is very confusing because this PR "hijacks" the function of an existing feature (which does not work?) to add another, new and different feature in the same function, with zero comment explaining that these are two different features and how much they overlap with each other (if at all).
 
Please add some minimal comments explaining what these two different features are, why we need two different features implemented in the same function, and how they relate to each other or not. I suspect one may be more focused on Linux while the other not? If that's true then say that.

> I had a look at some of the timestamps in https://sof-ci.01.org/sofpr/PR8480/build14946/devicetest/index.html and they never seem to match.

Please explain.

There are also unanswered questions in https://github.com/thesofproject/sof/issues/8016

src/audio/base_fw.c
RanderWang: you commented it was weird so I made this more clear

app/boards/intel_adsp_cavs25_tgph.conf
RanderWang: Why they can support it ? or need it ?   It needs 64 bits hardware clock support. I am not familiar with other vendor hardware so I can't do it

lgirdwood (commented):
LGTM, just need to use floor()

src/audio/base_fw.c
RanderWang: sure, thanks

lgirdwood (approved):

kv2019i:
@marc-hb Please check again, this is one of the last features considered for v2.8.

In the end, this is not really critical for the v2.8 release as this only affects debugging/development, but to improve our debug capabilities (e.g. in SOF CI), I don't see a strong reason not to go ahead and merge. This doesn't touch the EXT_TIME implementation, and adds a capability the Linux driver can utilize. 

I would like to see some more progress on the kernel part: https://github.com/thesofproject/sof/pull/8480 -- the discussion does not seem to be converging, and unless we have a line-of-sight for the kernel part, merging this FW part is not useful. If you have this in FW, but not in kernel, I'd argue the logs are less readable.

@RanderWang Do you have any data on the runtime cost. I.e. how many more DSP cycles does a single log statement take with this PR compared to baseline? That would be also useful. If the cost is small, even less reason to hold this back (as ability easily compare events between kernel and FW logs, is obviously very nice thing to have).

marc-hb:
> I would like to see some more progress on the kernel part: https://github.com/thesofproject/sof/pull/8480 -

I think you meant:
- https://github.com/thesofproject/linux/pull/4706


> and unless we have a line-of-sight for the kernel part, merging this FW part is not useful. If you have this in FW, but not in kernel, I'd argue the logs are less readable.

Agreed 100%.

src/audio/base_fw.c
marc-hb: What's weird is the same `*data` goes  in the same function to two different places that seem to do more or less the same thing without a word of explanation of why these two places are needed and how they compare to each other. That has never changed in this PR and it has never been explained.

RanderWang:
> @marc-hb Please check again, this is one of the last features considered for v2.8.
> 
> In the end, this is not really critical for the v2.8 release as this only affects debugging/development, but to improve our debug capabilities (e.g. in SOF CI), I don't see a strong reason not to go ahead and merge. This doesn't touch the EXT_TIME implementation, and adds a capability the Linux driver can utilize.
> 
> I would like to see some more progress on the kernel part: #8480 -- the discussion does not seem to be converging, and unless we have a line-of-sight for the kernel part, merging this FW part is not useful. If you have this in FW, but not in kernel, I'd argue the logs are less readable.
> 
> @RanderWang Do you have any data on the runtime cost. I.e. how many more DSP cycles does a single log statement take with this PR compared to baseline? That would be also useful. If the cost is small, even less reason to hold this back (as ability easily compare events between kernel and FW logs, is obviously very nice thing to have).

@kv2019i  it only add one add operation for " +  time_detla" compared to original code, very light DSP cycles

kv2019i:
@RanderWang wrote:
> @kv2019i it only add one add operation for " + time_detla" compared to original code, very light DSP cycles

I was also thinking of posisble added code on Zephyr logging subsys side, but checked that now as well, and indeed nothing extra done when the custom timestamp function is used, so indeed it's just the one add. So basicly this is all good to go.

src/audio/base_fw.c
```diff
@@ -240,14 +240,18 @@ static uint32_t basefw_set_system_time(uint32_t param_id,
...
 	/* TODO: configurate time to logging subsystem */
 	if (!(first_block && last_block))
 		return IPC4_INVALID_REQUEST;
 
 	global_system_time_info.host_time.val_l = ((const struct ipc4_system_time *)data)->val_l;
 	global_system_time_info.host_time.val_u = ((const struct ipc4_system_time *)data)->val_u;
 
-	uint64_t current_dsp_time = sof_cycle_get_64();
+	current_dsp_cycle = sof_cycle_get_64();
+	current_dsp_time = k_cyc_to_us_floor64(current_dsp_cycle);
```
ujfalusi, line 263: Why do you need `current_dsp_cycle`?
```c
current_dsp_time = k_cyc_to_us_floor64(sof_cycle_get_64());
```
is not working?

RanderWang: we save current_dsp_cycle to dsp_time in the following code

src/audio/base_fw.c
```diff
@@ -256,6 +265,19 @@ static uint32_t basefw_set_system_time(uint32_t param_id,
...
+	/* use default timestamp if 64bit is not enabled since 64bit is necessary for host time */
+	if (!IS_ENABLED(CONFIG_LOG_TIMESTAMP_64BIT)) {
+		LOG_WRN("64bits timestamp is disabled, so use default timestamp");
+		return IPC4_SUCCESS;
+	}
+
+	host_time = global_system_time_info.host_time.val_l |
+			((uint64_t)global_system_time_info.host_time.val_u << 32);
+	host_time_cycle = k_us_to_cyc_ceil64(host_time);
+	time_delta = host_time_cycle - current_dsp_cycle;
```
ujfalusi, line 277: The naming is quite confusing:
current_dsp_cycle - cycles
current_dsp_time - time in us
host_time - time in us
host_time_cycle - cycles
time_delta - cycles
?

why not:
current_dsp_cycle - cycles
current_dsp_time - time in us
host_time - time in us
host_cycle - cycles
cycle_delta - cycles

src/audio/base_fw.c
ujfalusi: I see, but in this patch it is not needed? At the end it does not going to matter, but I usually prefer to not introduce stale variables that is going to be used by upcoming patch without pointing out that `current_dsp_cycle` is introduced for the next patch.

src/audio/base_fw.c
aiChaoSONG: @marc-hb https://github.com/thesofproject/sof/commit/da5edbdf7d1f73bfaccca0bfc2e19fe46097b0d0 add get/set ops for system time, they will be called on IPC, and could not be deleted. And this PR did something further, do some calculation for time delta, and use it to make log timestamp more accurate.

src/audio/base_fw.c
btian1: uint32_t should be enough, can delta > max(uint32_t), it is 112 seconds.

src/audio/base_fw.c
btian1: unit seems not aligned.

src/audio/base_fw.c
btian1: 38.4mhz clock

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -90,3 +90,4 @@ CONFIG_DEBUG_COREDUMP_MEMORY_DUMP_MIN=y
 
 CONFIG_PROBE=y
 CONFIG_PROBE_DMA_MAX=2
+CONFIG_LOG_TIMESTAMP_64BIT=y
```
btian1, line 93: I would suggest config change as the third patch for enabling, and the second for code changes.

src/audio/base_fw.c
RanderWang: updated

src/audio/base_fw.c
RanderWang: updated, thanks

src/audio/base_fw.c
RanderWang: it is not related to precision. Log system use 64bits log so use 64bit

app/boards/intel_adsp_ace15_mtpm.conf
lgirdwood: @RanderWang ack - this will make it more bisectable.

lgirdwood (commented):
@RanderWang LGTM, I would just split the Kconfig into a 3rd patch and then I can merge.

src/audio/base_fw.c
marc-hb: If it's done then remove the TODO

app/boards/intel_adsp_ace15_mtpm.conf
RanderWang: done

RanderWang:
@lgirdwood  move the kconfig setting into to 3rd commit.

marc-hb (commented):
Looks OK but please don't merge before:
- https://github.com/thesofproject/linux/pull/4706

It's not even approved yet.

src/audio/base_fw.c
```diff
@@ -240,14 +240,16 @@ static uint32_t basefw_set_system_time(uint32_t param_id,
 				       uint32_t data_offset,
 				       const char *data)
 {
+	uint64_t current_dsp_time;
```
marc-hb: This was not necessary, we don't need to be compatible with C89.

kv2019i:
@RanderWang kernel part merged now, this can be resumed now

marc-hb:
SOFCI TEST

src/audio/base_fw.c
RanderWang: thanks, updated

RanderWang:
> > base_fw: convert dsp hw cycle to dsp time
> > Dsp time is in format of micro second, not hw cycle.
> 
> Where does this (new?) information come from?

It is a usage in host windows driver. I didn't find any documents.

RanderWang:
Remove TODO and restore the code style for C89

src/audio/base_fw.c
```diff
@@ -35,6 +36,7 @@ DECLARE_SOF_RT_UUID("basefw", basefw_comp_uuid, 0xe398c32, 0x5ade, 0xba4b,
 DECLARE_TR_CTX(basefw_comp_tr, SOF_UUID(basefw_comp_uuid), LOG_LEVEL_INFO);
 
 static struct ipc4_system_time_info global_system_time_info;
+static uint64_t cycle_delta;
```
kv2019i, line 39: global_cycle_delta or g_cycle_delta would increase readability

src/audio/base_fw.c
```diff
@@ -241,23 +243,46 @@ static int basefw_mem_state_info(uint32_t *data_offset, char *data)
...
+		LOG_WRN("64bits timestamp is disabled, so use default timestamp");
+		return IPC4_SUCCESS;
+	}
 
-	global_system_time_info.dsp_time.val_l = (uint32_t)(current_dsp_time);
-	global_system_time_info.dsp_time.val_u = (uint32_t)(current_dsp_time >> 32);
+	host_time = global_system_time_info.host_time.val_l |
+			((uint64_t)global_system_time_info.host_time.val_u << 32);
+	host_cycle = k_us_to_cyc_ceil64(host_time);
+	cycle_delta = host_cycle - dsp_cycle;
```
kv2019i, line 283: It took me a moment to figure out that host_time and host_cycle are local and cycle_delta is global. A minor thing, but given so many similarly named variables, naming cycle_delta "global_cycle_delta" would increase readability.

kv2019i (approved):
Looks good. I have a one more style nit (please check inline). If you need to update for any reason, please update that as well. Otherwise good, thanks!

marc-hb:
https://sof-ci.01.org/sofpr/PR8480/build1544/devicetest/index.html and https://sof-ci.01.org/sofpr/PR8480/build1545/devicetest/index.html are empty, no tests have run.

This is because the corresponding kernel build is marked as FAIL.

@keqiaozhang , after a lot of digging I found this socwatch.tar error below in some Jenkins log. I don't whether it's relevant but it should be fixed anyway.

```
00:50:09   ...
00:50:09    CC [M]  /srv/home/jenkins/workspace/linux_config_build/socwatch_build/socwatch/socwatch_driver/socwatch2_15.mod.o
00:50:09    LD [M]  /srv/home/jenkins/workspace/linux_config_build/socwatch_build/socwatch/socwatch_driver/socwatch2_15.ko
00:50:09  make[1]: Leaving directory '/srv/home/jenkins/workspace/linux_config_build/build/default'
00:50:09  ************ Built drivers are copied to /srv/home/jenkins/workspace/linux_config_build/socwatch_build/socwatch/drivers directory ************
00:50:12  /srv/home/jenkins/workspace/linux_config_build/build
00:50:12  ├── config-default.txt
00:50:12  ├── debbuild.log
00:50:12  ├── kernelbuild.json
00:50:12  ├── linux-headers-6.7.0-rc3-gde56b759c320_6.7.0-rc3-default-topic-sof-dev_amd64.deb
00:50:12  ├── linux-image-6.7.0-rc3-gde56b759c320_6.7.0-rc3-default-topic-sof-dev_amd64.deb
00:50:12  ├── linux-libc-dev_6.7.0-rc3-default-topic-sof-dev_amd64.deb
00:50:12  ├── linux-upstream_6.7.0-rc3-default-topic-sof-dev_amd64.buildinfo
00:50:12  └── linux-upstream_6.7.0-rc3-default-topic-sof-dev_amd64.changes
00:50:12  
00:50:12  0 directories, 8 files
00:50:13  sending incremental file list
00:50:13  config-default.txt
00:50:13  debbuild.log
00:50:13  kernelbuild.json
00:50:13  linux-headers-6.7.0-rc3-gde56b759c320_6.7.0-rc3-default-topic-sof-dev_amd64.deb
00:50:14  linux-image-6.7.0-rc3-gde56b759c320_6.7.0-rc3-default-topic-sof-dev_amd64.deb
00:50:15  linux-libc-dev_6.7.0-rc3-default-topic-sof-dev_amd64.deb
00:50:15  linux-upstream_6.7.0-rc3-default-topic-sof-dev_amd64.buildinfo
00:50:15  linux-upstream_6.7.0-rc3-default-topic-sof-dev_amd64.changes
00:50:15  
00:50:15  sent 33,210,864 bytes  received 168 bytes  13,284,412.80 bytes/sec
00:50:15  total size is 33,684,535  speedup is 1.01
00:50:15  sending incremental file list
00:50:15  rsync: [sender] link_stat "/srv/home/jenkins/workspace/linux_config_build/socwatch_build/socwatch.tar.gz" failed: No such file or directory (2)
00:50:15  
00:50:15  sent 18 bytes  received 12 bytes  20.00 bytes/sec
00:50:15  total size is 0  speedup is 0.00
00:50:15  rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1338) [sender=3.2.7]
```

marc-hb:
SOFCI TEST

src/audio/base_fw.c
RanderWang: thanks, good idea!

src/audio/base_fw.c
RanderWang: updated

kv2019i:
@wszypelt Strange error in quickbuild TGL IPC4 tests. I see an error rsyncing log files and one smoketest fails to

12:47:41,146 INFO  - Unhandled exception during verification: System.BadImageFormatException: Field token out of range.

...  the quality test "16_00_TestSdwTransferLoopback\m0_sourcePDI2_sinkPDI3_output_loopback_2ch_48000Hz_24b.wav" seems to pass before the exception is triggered.

lgirdwood (approved):

lgirdwood:
> @wszypelt Strange error in quickbuild TGL IPC4 tests. I see an error rsyncing log files and one smoketest fails to
> 
> 12:47:41,146 INFO - Unhandled exception during verification: System.BadImageFormatException: Field token out of range.
> 
> ... the quality test "16_00_TestSdwTransferLoopback\m0_sourcePDI2_sinkPDI3_output_loopback_2ch_48000Hz_24b.wav" seems to pass before the exception is triggered.

Internal CI now passing, must have been an unrelated issue.

PR #8482, opened 2023-11-15
Audio: DRC: Add HiFi4 implementation of DRC

andrula-song:
Add HiFi4 implementation of DRC.


btian1 (commented):
Do we have a tplg to perform DRC test? if have, how about generic/hifi3/hifi4 results？

src/audio/drc/drc_hifi4.c
```diff
@@ -0,0 +1,911 @@
...
+					}
+					inc++;
+				}
+
+				if (++i == count)
+					break;
+
+				for (j = 0; j < 4; j++)
+					x[j] = drc_mult_lshift(x[j], r4, LSHIFT_QX30_QY30_QZ30);
+			}
```
lyakh, line 379: alternatively this could be written as
```
for (i = 0; i < count; i++) {
    for (j = 0; j < 4; j++) {
        ...
        if (i < count - 1)
            x[j] = drc_mult_lshift(x[j], r4, LSHIFT_QX30_QY30_QZ30);
    }
}
```
but I'm not sure if it wins anything in terms of execution time, at least instead of 2 loops and an `if` internally it then would have 1 loop and an `if`. Same 3 more times below of course

andrula-song: I tried before and checked again, wouldn't better, so keep the original version

btian1 (commented):
now, we can try with turn off IRQ to check performance, maybe, there is a good results waiting for you.

dbaluta:
off topic question here: On which platforms does Intel have hifi4?

andrula-song:
> off topic question here: On which platforms does Intel have hifi4?
mtl

lgirdwood (approved):
LGTM, @cujomalainey fyi.

src/audio/drc/drc_hifi4.c
btian1: it is not be better because j < 4, if it is j < 48 or larger, it would bring better cache performance.
this is due to origin code need D cache load more more data than proposed.

I agree keep as current, what's 4 means here?

src/audio/drc/drc_hifi4.c
```diff
@@ -0,0 +1,911 @@
...
+#endif /* CONFIG_FORMAT_S24LE */
+
+#if CONFIG_FORMAT_S32LE
+	{ SOF_IPC_FRAME_S32_LE, drc_s32_default },
+#endif /* CONFIG_FORMAT_S32LE */
+};
+
+const size_t drc_proc_fncount = ARRAY_SIZE(drc_proc_fnmap);
+
+#endif /* DRC_HIFI3 */
```
marc-hb, line 911:
```suggestion
#endif /* DRC_HIFI4 */
```

PR #8484, opened 2023-11-16
peakvol: set the correct vol_min and vol_max values

fkwasowi:
The current values result in incorrect signal
amplitudes at the peakvol module output.

### DIFF
--- a/src/audio/volume/volume_ipc4.c
+++ b/src/audio/volume/volume_ipc4.c
@@ -94,16 +94,8 @@ static void init_ramp(struct vol_data *cd, uint32_t curve_duration, uint32_t tar
 						  Q_CONVERT_FLOAT(1.0 / 10000, 31), 0, 31, 0);
 	}
 
-	if (!cd->initial_ramp) {
-		/* In case when initial ramp time is equal to zero, vol_min and
-		 * vol_max variables should be set to target_volume value
-		 */
-		cd->vol_min = target_volume;
-		cd->vol_max = target_volume;
-	} else {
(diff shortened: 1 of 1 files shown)


singalsu (approved):
Thanks, this looks correct to me. Hope this also passes QB tests. There could be some other issues in IPC4 volume that has been worked around with the strange code.

dnikodem (approved):

src/audio/volume/volume_ipc4.c
```diff
@@ -94,16 +94,8 @@ static void init_ramp(struct vol_data *cd, uint32_t curve_duration, uint32_t tar
...
-		 * vol_max variables should be set to target_volume value
-		 */
-		cd->vol_min = target_volume;
-		cd->vol_max = target_volume;
-	} else {
-		cd->vol_min = VOL_MIN;
-		cd->vol_max = VOL_MAX;
-	}
+	cd->vol_min = VOL_MIN;
+	cd->vol_max = VOL_MAX;
```
kv2019i, line 98: @RanderWang can you review, this is removing the logic in commit 751d0a967e765ac193e7e75b38c30b11bbc37c51

kv2019i (approved):
Looks good to me, but would like to understand what was the problem this code originally solved. See inline

src/audio/volume/volume_ipc4.c
RanderWang: sure.

RanderWang (approved):

btian1 (approved):

PR #8489, opened 2023-11-17
Fixes for compilation errors with the SOF plugin

ranj063:


RanderWang (approved):
LGTM

lgirdwood (approved):

lgirdwood:
@lrudyX @wszypelt not expecting this to fail as it's build fixes (and IIUC would fail on all platforms if broken). Good to merge ?

lrudyX:
> @lrudyX @wszypelt not expecting this to fail as it's build fixes (and IIUC would fail on all platforms if broken). Good to merge ?

It looks like QB problem, I will rerun this.

src/include/ipc4/module.h
```diff
@@ -64,7 +64,7 @@ union ipc4_extended_param_id {
 		uint32_t parameter_type     : 8;
 		uint32_t parameter_instance : 24;
 	} part;
-} __packed __aligned(4);
+} __attribute__((packed, aligned(4)));
```
btian1, line 67: I noticed kernel use left expression for alignment, and sof fw use right ones, any history on this?

kv2019i: "__packed" is a linux kernel shortcut, not necessarily available in generic toolchains (like for SOF plugin).

kv2019i (approved):
We really need to add a CI test to check the SOF plugin build. Otherwise this will be continuous....

PR #8491, opened 2023-11-17
Audio: DRC: Change DRC to use lookup table based sine function

singalsu:
This change saves in TGL platform about 13 MPCS, from 83 to 70 MCPS. In MTL platform the saving is 12 MCPS, from 46 to 34 MCPS.

The .bss RAM usage increases by 1 kB from selecting CONFIG_MATH_LUT_TRIG_FIXED.


src/math/lut_trig.c
```diff
@@ -0,0 +1,574 @@
...
+#include <sof/audio/format.h>
+#include <sof/math/lut_trig.h>
+#include <stdint.h>
+
+#define SOFM_LUT_SINE_C_Q20 341782638 /* 2*SINE_NQUART/pi in Q12.20 */
+#define SOFM_LUT_SINE_NQUART 512 /* Must be 2^N */
+#define SOFM_LUT_SINE_SIZE (SOFM_LUT_SINE_NQUART + 1)
+
+/* An 1/4 period of sine wave as Q1.31 */
+const int32_t sofm_lut_sine_table[SOFM_LUT_SINE_SIZE] = {
```
singalsu, line 16: TODO: Should test if this could be int16_t and size 257 for 16 bit sine.

src/include/sof/audio/drc/drc_math.h
```diff
@@ -59,9 +60,7 @@ static inline int32_t drc_sin_fixed(int32_t x)
 {
 	const int32_t lshift = drc_get_lshift(30, 30, 28);
 	int32_t denorm_x = drc_mult_lshift(x, PI_OVER_TWO_Q30, lshift);
-	int32_t sin_val = sin_fixed_16b(denorm_x);
-
-	return sin_val << 16;
+	return sofm_lut_sin_fixed_32b(denorm_x);
```
lyakh, line 63: some of the tooling might complain about a missing line between variable definitions and statements - same below

src/math/lut_trig.c
btian1: one is check wether can use 16-bit， another thing is whether can shorten the size? do we really need 512? 
such a small usage cost 2k bytes table, it is big cost.

Also, could you add comments for how this table calculated? sin(i/2pi)? then we can further figure out whether 512 is must.

btian1 (commented):
do we have a float version drc source code? I want to first check float version, the fixed version, then optimized version.

src/math/Kconfig
```diff
@@ -9,6 +9,15 @@ config CORDIC_FIXED
 	  Select this to enable sin(), cos(), asin(), acos(),
 	  and cexp() functions as 16 bit and 32 bit versions.
 
+config LUT_TRIG_FIXED
```
lgirdwood, line 12: We need a cleanup at some point where we have ```MATH_TRIG_``` prefix and likewise convention for all maths APIs, macros, Kconfigs etc.

src/math/lut_trig.c
lgirdwood: static ?

src/math/Kconfig
```diff
@@ -9,6 +9,15 @@ config CORDIC_FIXED
...
+	bool "Lookup table based trigonometric functions"
+	default n
+	help
+	  Select this to enable sofm_lut_sin_fixed_32b() function. The
+	  calculation is using 1/4 wave lookup and interpolation.
+	  This option consumes 2052 bytes .bss RAM for the lookup
+	  table.
```
lgirdwood, line 19: Can we offer advice on when each trig type should be used. i.e. I would expect we export the same public API for all maths, but the internal calculations will depend on which Kconfig is selected by the user at build time.

src/math/Kconfig
singalsu: Yep, done

src/math/Kconfig
singalsu: I added some text to Kconfig about preferring the lookup sine when used in hot code parts.

singalsu:
> do we have a float version drc source code? I want to first check float version, the fixed version, then optimized version.

There used to be long ago in git first version that was float C by Sebastiano and Johny but it was replaced by fixed point code when the work proceeded. A float version should be found from ChromeOS sources. The float code and fixed conversion work for this contribution is owned by team Google so we don't review it here.

I've used the scripts in tools/test/audio to evaluate objective steady signal audio parameters for DRC and we've not seen difference in team Intel's optimizations. DRC has complex transient signal characteristics, and we currently don't have other but subjective expert listening test method for that. It means for me to listen an album of music with this processing in DUT and try to spot any issues.

zephyr/CMakeLists.txt
```diff
@@ -458,6 +458,11 @@ zephyr_library_sources_ifdef(CONFIG_LOG_BACKEND_SOF_PROBE
 
 # Optional SOF sources - depends on Kconfig - WIP
 
+# Optional math utility
+zephyr_library_sources_ifdef(CONFIG_MATH_LUT_TRIG_FIXED
+	${SOF_MATH_PATH}/lut_trig.c
+)
+
```
marc-hb, line 465: Please move this next to the other SOF_MATH_PATH (#8620)

src/audio/drc/Kconfig
```diff
@@ -3,6 +3,7 @@
 config COMP_DRC
 	bool "Dynamic Range Compressor component"
 	select CORDIC_FIXED
+        select MATH_LUT_TRIG_FIXED
```
ShriramShastry, line 6: Can it be `MATH_LUT_SINE_FIXED` instead of `MATH_LUT_TRIG_FIXED`?

src/math/Kconfig
```diff
@@ -9,6 +9,16 @@ config CORDIC_FIXED
 	  Select this to enable sin(), cos(), asin(), acos(),
 	  and cexp() functions as 16 bit and 32 bit versions.
 
+config MATH_LUT_TRIG_FIXED
```
ShriramShastry, line 12: Can it be `MATH_LUT_SINE_FIXED` instead of` MATH_LUT_TRIG_FIXED`?

ShriramShastry (approved):
I have reviewed the changes and they appear to be in good standing. I hope that LUT's size is good to others.

src/math/Kconfig
```diff
@@ -9,6 +9,16 @@ config CORDIC_FIXED
...
+	bool "Lookup table based trigonometric functions"
+	default n
```
btian1, line 14: doe it have dependency? I mean there should have one module use this, and this config should depend on that module.

src/audio/drc/Kconfig
singalsu: Yep, I'll change the config name. There is no need for other functions now.

src/math/Kconfig
singalsu: DRC component has "select MATH_LUT_TRIG_FIXED". OK? I will rename that as Sriram suggested but keep it that way.

zephyr/CMakeLists.txt
singalsu: Wrong PR? There's no change to CMakeLists.txt in that.

singalsu:
> I have reviewed the changes and they appear to be in good standing. I hope that LUT's size is good to others.

I tried 256 size LUT but the quality dropped a lot. With 512 the quality is a tiny bit better than in 16 bit cordic, so there should be no negative audio quality impact from this.

src/math/lut_trig.c
```diff
@@ -0,0 +1,108 @@
...
+	 65067,  65091,  65114,  65137,  65159,  65180,  65200,  65220,  65240,  65259,  65277,
+	 65294,  65311,  65328,  65343,  65358,  65373,  65387,  65400,  65413,  65425,  65436,
+	 65447,  65457,  65467,  65476,  65484,  65492,  65499,  65505,  65511,  65516,  65521,
+	 65525,  65528,  65531,  65533,  65535,  65535,  65535
+};
+
+/* Sine lookup table read */
+static inline int32_t sofm_sine_lookup_16b(int idx)
+{
+	uint16_t s;
```
lyakh, line 74: why not `int16_t`?

src/math/lut_trig.c
```diff
@@ -0,0 +1,108 @@
...
+	int i1;
+
+	i1 = idx & (2 * SOFM_LUT_SINE_NQUART - 1);
+	if (i1 > SOFM_LUT_SINE_NQUART)
+		i1 = 2 * SOFM_LUT_SINE_NQUART - i1;
+
+	s = sofm_lut_sine_table_s16[i1];
```
lyakh, line 81: you could even define `sofm_lut_sine_table_s16[]` as signed, but you know better, what's more logical

src/math/lut_trig.c
```diff
@@ -0,0 +1,108 @@
...
+	int32_t frac;
+	int32_t delta;
+	int32_t s0;
+	int32_t s1;
+	int64_t idx_tmp;
+
+	/* Q4.28 x Q12.20 -> Q16.48 --> Q16.31*/
+	idx_tmp = ((int64_t)w * SOFM_LUT_SINE_C_Q20) >> 17;
+	idx = (idx_tmp >> 31); /* Shift to Q0 */
+	frac = (int32_t)(idx_tmp - (idx << 31)); /* Get fraction Q1.31*/
```
lyakh, line 102: that seems to boil down to
```
idx_tmp - (idx_tmp & 0xffffffff80000000ULL) ==
idx_tmp & 0x7fffffff
```
would the compiler optimise that out by itself?

test/cmocka/src/math/trig/lut_sin_16b_fixed.c
```diff
@@ -0,0 +1,55 @@
...
+#define _M_PI		3.14159265358979323846	/* pi */
+
+static void test_math_trig_lut_sin_fixed(void **state)
+{
+	(void)state;
+
+	int theta;
+
+	for (theta = 0; theta < 360; ++theta) {
+		double rad = _M_PI * (theta / 180.0);
```
lyakh, line 31: hopefully the compiler will calculate `_M_PI / 180.0` at compile time, but parentheses might actually prevent it from doing that and make it a (redundant) run-time calculation

src/math/Kconfig
btian1: not sure need add: depends on COMP_DRC, because this table is large, if wrongly included by other modules,  it will occupy more space.

src/math/lut_trig.c
singalsu: I managed to get almost true 16 bit accuracy this way.  The lookup table with only non-negative numbers are equal to 17 bit signed equivalent accuracy. Unsigned to signed conversions give some small MCPS penalty for this code but I did not want to propose option to cordic 16 bit sine that is less accurate.

src/math/lut_trig.c
singalsu: It's using 0 - 65535 non-negative integers, wouldn't that violate C standard. I don't like unsigned types but I got some measurable advantage from it.

src/math/lut_trig.c
singalsu: I was thinking that but it looks awkward in arithmetic that is not bit-banging to HW registers etc. The shifts have association to Qx.y format. But if that gives MCPS advantage I can change, and comment what happens, I'll try.

test/cmocka/src/math/trig/lut_sin_16b_fixed.c
singalsu: It's cmocka test code so we don't care about performance even if there would be emulated floats.

src/math/Kconfig
singalsu: I don't see how other modules would select this accidentally. E.g. MFCC and FFT use trig functions only in initialization code, where the cordic version as 32 bits is ideal.

If some other component would need sine function in hot code then this function should be available for them also. It's not making sense to make this private for DRC.

src/math/lut_trig.c
singalsu: With this modification 69.976 to 69.925 MCPS, not worth it I think, because of bit-and awkwardness here. I think our perf measurement works down to 0.1 MCPS level, below that it's probably noise.

src/math/lut_trig.c
lyakh: uhm, right, yes, sorry, I missed that the function is returning 32 bits and it's type-casting to 32 bits... I thought it was 16 bits throughout, sorry, temporary loss of sight.

btian1 (approved):

lgirdwood:
@singalsu can you check CI. Thanks !

kv2019i (approved):

Issue #8481, opened 2023-11-14
[BUG] DMIC capture failed to start


yongzhi1:
[](url)**Describe the bug**

Launch arecord on DMIC Raw, and observe error returned immediately.

**To Reproduce**
Steps to reproduce the behavior: (e.g. list commands or actions used to reproduce the bug)

```
arecord -Dhw:0,99 /tmp/test.wav -c4 -r48000 -fs32_le -d10 
```
**Reproduction Rate**
100%

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel: {https://chromium.googlesource.com/chromiumos/third_party/kernel/+log/refs/heads/chromeos-6.1/sound}
    * SOF: {https://github.com/thesofproject/sof/commit/a8db157b9a2dee8}
2) Name of the topology file
    * Topology: {sof-mtl-max98357a-rt5682-ssp2-ssp0.tplg}
3) Name of the platform(s) on which the bug is observed.
    * Platform: {MTL/Rex}

**Screenshots or console output**
If applicable, add a screenshot (drag-and-drop an image), or console logs

[mtrace-main-dmic.txt](https://github.com/thesofproject/sof/files/13357969/mtrace-main-dmic.txt)
[dmesg-dmic-err.txt](https://github.com/thesofproject/sof/files/13357973/dmesg-dmic-err.txt)

lgirdwood:
@singalsu can you check this on main, any thoughts ?

lgirdwood:
@yongzhi1 can you confirm if stereo is working with your device, i.e. is issue only with 4 channels and 32bit data ? Thanks

singalsu:
Seems there's no data coming to host. Do you know if this is a regression?

singalsu:
The `<err> dai_intel_dmic_nhlt:` seen in mtrace from DMIC is not an error but wrong log function usage. I will create a PR to change it to LOG_INF().

abonislawski:
Reverting this PR can help in this scenario:
https://github.com/thesofproject/sof/pull/8084

@ujfalusi can you take a quick look if we are missing something obvious?

yongzhi1:
> @yongzhi1 can you confirm if stereo is working with your device, i.e. is issue only with 4 channels and 32bit data ? Thanks

[...]

PR #8504, opened 2023-11-20
ipc4: handler: maintain IPC set_pipeline_state order

kv2019i:
Fix a side-effect of commit 068f14381026 ("ipc4: handler: Use separate loops for pipeline state phases") where the order of pipeline triggers was changed when the trigger handling was delayed to the pipeline thread. If host asked for pipelines to be triggered in order 0, 1, 2, the actual trigger order might end up as 2, 1, 0 based on how the pipes were connected.

Fix the issue by adding a blocking wait to the IPC thread between each delayed trigger.

Link: https://github.com/thesofproject/sof/issues/8481

### DIFF
--- a/src/ipc/ipc4/handler.c
+++ b/src/ipc/ipc4/handler.c
@@ -630,6 +630,17 @@ static int ipc4_set_pipeline_state(struct ipc4_message_request *ipc4)
 			ipc_compound_pre_start(state.primary.r.type);
 			ret = ipc4_pipeline_trigger(ppl_icd, cmd, &delayed);
 			ipc_compound_post_start(state.primary.r.type, ret, delayed);
+			if (delayed) {
+				/* To maintain pipeline order for triggers, we must
+				 * do a blocking wait until trigger is processed.
+				 * This will add a max delay of 'ppl_count' LL ticks
+				 * to process the full trigger list.
+				 */
+				if (ipc_wait_for_compound_msg() != 0) {
+					ipc_cmd_err(&ipc_tr, "ipc4: fail with delayed trigger");
+					return IPC4_FAILURE;
+				}
+			}
 		}
 
 		if (ret != 0)


kv2019i:
This is a bit heavyweight approach, let's see how this fares in CI.

Alternative approach would be to stop propagating pipeline triggers across pipelines in IPC4 (as in IPC4, host sends a separate IPC action to each pipe while in IPC3, only edge nodes of the graph are triggered and DSP traverses the graph to trigger connected pipes).

RanderWang (commented):
If pipeline priority is set,  can zephyr schedule pipeline by priority ?

plbossart:
> If pipeline priority is set, can zephyr schedule pipeline by priority ?

scheduling for LL is not done at the pipeline level, the priority only defines the order in which LL modules will be handled.

RanderWang:
> > If pipeline priority is set, can zephyr schedule pipeline by priority ?
> 
> scheduling for LL is not done at the pipeline level, the priority only defines the order in which LL modules will be handled.

For cSOF,  pipeline is the basic schedule elements, not LL module (only pipeline calls schedule_task_init_ll, not module) . please check pipeline_task which triggers each module in the pipeline from source to sink.

plbossart:
> > > If pipeline priority is set, can zephyr schedule pipeline by priority ?
> > 
> > 
> > scheduling for LL is not done at the pipeline level, the priority only defines the order in which LL modules will be handled.
> 
> For cSOF, pipeline is the basic schedule elements, not LL module (only pipeline calls schedule_task_init_ll, not module) . please check pipeline_task which triggers each module in the pipeline from source to sink.

This will make no sense when LL and DP modules are mixed...

RanderWang:
As I know, pipeline2.0 is being prepared, maybe it change the game

src/ipc/ipc4/handler.c
```diff
@@ -630,6 +630,15 @@ static int ipc4_set_pipeline_state(struct ipc4_message_request *ipc4)
...
 			ipc_compound_post_start(state.primary.r.type, ret, delayed);
+			if (delayed) {
+				/* To maintain pipeline order for triggers,
+				 * must wait until trigger action is done
+				 */
+				if (ipc_wait_for_compound_msg() != 0) {
+					ipc_cmd_err(&ipc_tr, "ipc4: fail with delayed trigger");
+					return IPC4_FAILURE;
+				}
+			}
```
lyakh, line 641: hmm, so that means that in your example the IPC thread might be delayed by up to 3 (or more for more pipelines) ms? I guess it won't break anything, but maybe at least mention this in the comment above?

kv2019i:
@plbossart wrote:
> @RanderWang wrote:
> > If pipeline priority is set, can zephyr schedule pipeline by priority ?
> 
> scheduling for LL is not done at the pipeline level, the priority only defines the order in which LL modules will be handled.

Ack on that. For LL tasks, we are using a single Zephyr thread per core, so we need to implement the expected behaviour within SOF LL scheduler logic, i.e. run the tasks in the priority order as requested by host when pipelines are set. And this is what also happens, so no problem here.

src/ipc/ipc4/handler.c
kv2019i: @lyakh Ack. I extended the comment to explain this, see updated patch.

kv2019i:
V2 and V3:
- extended the inline comment based on Guennadi's input
- clarified the git commit to explain that the order is changed if host asks for trigger order that is different from the order LL will schedule the pipeline tasks (based on their priority)

kv2019i:
Dropped the title link to bug #8481  . This will help to avoid the dmic -ENODATA error seen in that case, but this change is required anyways for FW to comply to host requested pipeline trigger order.

lyakh (approved):
a bit scary, but we don't have a better solution so far

RanderWang (approved):

kv2019i:
SOFCI TEST

lgirdwood (approved):

PR #8506, opened 2023-11-21
audio: pipeline: do not propagate pipeline triggers for IPC4

kv2019i:
One major difference in IPC4 versus IPC3 is that each pipeline will get a separate SET_PIPELINE_STATE command. This is in stark contrast to IPC3 where host sends STREAM_TRIG_START and firmware is expected to propagate the state change to all connected pipelines in the graph.

Change the code such that when pipeline component trigger is executed in IPC4 build, propagation is stopped whenever we reach another pipeline.

This prevents bugs stemming from IPC3 related logic to propagate pipeline triggers, interfering with IPC4 usages like in bug 8481.

Link: https://github.com/thesofproject/sof/issues/8481

### DIFF
--- a/src/audio/pipeline/pipeline-stream.c
+++ b/src/audio/pipeline/pipeline-stream.c
@@ -47,6 +47,17 @@ pipeline_should_report_enodata_on_trigger(struct comp_dev *rsrc,
 	struct pipeline_data *ppl_data = ctx->comp_data;
 	struct comp_dev *pipe_source = ppl_data->start->pipeline->source_comp;
 
+	/* In IPC3, FW propagates triggers to connected pipelines, so
+	 * it can have determistic logic to conclude no data is
+	 * available.
+	 * In IPC4, host controls state of each pipeline separately,
+	 * so FW cannot reliably detect case of no data based on
+	 * observing state of src->pipeline here.
+	 */
+#if CONFIG_IPC_MAJOR_4
+	return false;
+#endif
+
 	/* only applies to capture pipelines */
 	if (dir != SOF_IPC_STREAM_CAPTURE)
 		return false;
@@ -388,8 +399,13 @@ static int pipeline_comp_trigger(struct comp_dev *current,
 
 	/* trigger should propagate to the connected pipelines,
 	 * which need to be scheduled together
+	 *
+	 * IPC4 has a SET_PIPELINE_STATE for each pipeline, so FW
+	 * should not propagate triggers on its own.
+	 * IPC3 has commands only for graph edges, so propagation is
+	 * needed in many cases.
 	 */
-	if (!is_single_ppl && !is_same_sched) {
+	if (!is_single_ppl && (!is_same_sched || IS_ENABLED(CONFIG_IPC_MAJOR_4))) {
 		pipe_dbg(current->pipeline,
 			 "pipeline_comp_trigger(), current is from another pipeline");
 


src/audio/pipeline/pipeline-stream.c
```diff
@@ -352,6 +352,20 @@ static int pipeline_comp_trigger(struct comp_dev *current,
...
+	 * should not propagate triggers on its own.
+	 * IPC3 has commands only for graph edges, so propagation is required.
+	 */
+	if (!is_single_ppl) {
+		pipe_dbg(current->pipeline,
+			 "pipeline_comp_trigger(), stop trigger propagation at comp.id = %u",
+			 dev_comp_id(current));
+		return 0;
+	}
+#endif
```
RanderWang, line 367: line 403 also discuss the propagation, can we combine it ?

keqiaozhang:
SOFCI TEST

src/audio/pipeline/pipeline-stream.c
lyakh: This seems a rather complicated piece of machinery to me. So I think it needs good documentation. It should describe, that:
Pipeline triggering is performed in two stages:
1. from the IPC context: there we set trigger flags for all affected pipelines and submit them for scheduling. For IPC4 we already check and only do that for pipelines, explicitly requested by the host, with no propagation, as determined by https://github.com/thesofproject/sof/blob/ee80da7b00f32170715ba91302b35aad858d75fe/src/audio/pipeline/pipeline-stream.c#L197 . The pipeline order was unimportant until now, but with #8504 the IPC thread will wait when setting each flag, so the order does matter with that. And the order is as sent by the host. And the host orders them output-to-input. So for a 2-pipeline playback stream we first trigger the DAI pipeline, and only on the next tick we trigger the host pipeline. But they're added to the scheduling list in the priority order, which is from input to output, so the scheduling will be first the host and then the DAI pipeline.
2. in the pipeline scheduler context, as described above, we actually execute the trigger operation. But when doing this we trigger not only the pipeline, where we've found the flag, but all connected pipelines too, and that breaks the ordering. This PR addresses that. With it we should get:
```

- IPC: add DAI for triggering, wait
- scheduler (tick 1):
  * process trigger DAI
  * no data - no copying
- IPC: complete waiting, add host for triggering - host task scheduled before DAI
- scheduler (tick 2):
  * process trigger host
  * produce data by host
  * copy data by DAI
```

As for @RanderWang 's proposal - I agree, let's merge it with 403 into something like
```
if (!is_single_ppl && (!is_same_sched || IS_ENABLED(CONFIG_IPC_MAJOR_4))) {
```
(I assume that the `pipeline_should_report_enodata_on_trigger()` logic will still work correctly).

kv2019i: @lyakh Ack on all the above, except I don't think enodata is good. It does work with current Linux SOF driver, so we can debate this, but fundamentally in IPC4, it's the host that controls when a pipeline is set RUNNING, so I just can't see how the ENODATA check can be reliably done in FW with IPC4.

src/audio/pipeline/pipeline-stream.c
```diff
@@ -60,10 +71,6 @@ pipeline_should_report_enodata_on_trigger(struct comp_dev *rsrc,
 	if (pipe_source && dev_comp_type(pipe_source) == SOF_COMP_DAI)
 		return false;
 
-	/* source pipeline may not be active since priority is not higher than current one */
-	if (rsrc->pipeline->priority <= ppl_data->start->pipeline->priority)
-		return false;
```
lyakh, line 65: is this intended or a left-over?

lyakh:
@kv2019i but is this really an alternative to #8504 or are they both needed? Without #8504 this doesn't achieve the goal - the IPC will submit all pipelines from the IPC list and then the scheduler will trigger them on the same tick in the scheduling order?

src/audio/pipeline/pipeline-stream.c
lyakh: seems to be a bad hunk

lyakh (changes requested):

src/audio/pipeline/pipeline-stream.c
RanderWang: A issue found on windows

RanderWang (approved):
LGTM

src/audio/pipeline/pipeline-stream.c
kv2019i: This is intended @lyakh . @RanderWang  If IPC4 build, we will return "false" already at the beginning of the function, so this should fix the issue you saw on Windows.

kv2019i:
@lyakh wrote:
> @kv2019i but is this really an alternative to #8504 or are they both needed? Without #8504 this doesn't achieve the goal - the IPC will submit all pipelines from the IPC list and then the scheduler will trigger them on the same tick in the scheduling order?

No, both are needed. Both PRs will actually "fix" 8504, and thus I marked these as options to fix the same issue before. But with updated findings, both of these PR make sense on their own.

src/audio/pipeline/pipeline-stream.c
kv2019i: In case of IPC3, the remove check to compare priorities seems wrong. IPC3 has not separate trigger order, so the check should actually be reversd for IPC3. If "rsrc" has higher priotiy, then we cannot check it state yet. In practise IPC3 pipelines have priority set to 0 so this is not hit, but I think it's better to remove.

kv2019i: @lyakh  Ok let me remove this. I think the hunk was added to solve a IPC4 specific problem (but didn't cause harm in IPC3 builds), but I admit it's very nonobvious toconclude why this code can be removed. It apparently does no harm for IPC3, so let's just limit the complexity of the patch and simply not touch this code.

kv2019i: @lyakh bad hunk removed

src/audio/pipeline/pipeline-stream.c
```diff
@@ -47,6 +47,17 @@ pipeline_should_report_enodata_on_trigger(struct comp_dev *rsrc,
...
+	/* In IPC3, FW propagates triggers to connected pipelines, so
+	 * it can have determistic logic to conclude no data is
+	 * available.
+	 * In IPC4, host controls state of each pipeline separately,
+	 * so FW cannot reliably detect case of no data based on
+	 * observing state of src->pipeline here.
+	 */
+#if CONFIG_IPC_MAJOR_4
+	return false;
+#endif
```
lyakh, line 59: nitpick / not a blocker: this is a bit crude too... We usually enclose the whole function in that preprocessor conditional bracket. I know I'm too late, approving anyway

lyakh (approved):

abonislawski (approved):

PR #8507, opened 2023-11-21
schedule: zephyr_ll: update zephyr_ll_task_insert_unlocked comments

kv2019i:
Clarify the inline comment in zephyr_ll_task_insert_unlocked() explaining the task priority semantics. The old text is not clear on what is "high priority". Make it explicit that 0 is the highest priority and that the tasks are executed in order of task priority.

### DIFF
--- a/src/schedule/zephyr_ll.c
+++ b/src/schedule/zephyr_ll.c
@@ -96,9 +96,10 @@ static void zephyr_ll_task_insert_unlocked(struct zephyr_ll *sch, struct task *t
 	task->state = SOF_TASK_STATE_QUEUED;
 
 	/*
-	 * Tasks are added into the list from highest to lowest priority. This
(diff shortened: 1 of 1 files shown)


kv2019i:
V2: tiny rewording of the comment
 - based on @ujfalusi 's feedback, fine-tuned the wording -> use "the largest priority value" to discuss the absolute values and mathematical MAX() seen, and use "the highest priority" when talking about the semantics (e.g. that smaller values have higher priority)

kv2019i:
V3 uploaded:
- text still too complicated, simplified the comment further

ranj063 (approved):

iuliana-prodan (approved):

ujfalusi (approved):

dbaluta (approved):

PR #8508, opened 2023-11-21
DP: bugix - lists initialization was done too late

marcinszkudlinski:
in case module prepare fails, dp_queue lists were left uninitialized.
It may lead to crash if module reset was performed later


jxstelter (approved):

lgirdwood (approved):

src/audio/module_adapter/module_adapter.c
```diff
@@ -168,6 +168,9 @@ static int module_adapter_dp_queue_prepare(struct comp_dev *dev)
 	 * first, set all parameters by calling "module prepare" with pointers to
```
kv2019i, line 168: "bugix" ? Also "module_adapter: " would be a better prefix. But not showstoppers, change itself looks good and git commit is understandable.

kv2019i (approved):

PR #8509, opened 2023-11-21
topology2: pipeline: document pipeline.priority attribute

kv2019i:
Add a note that priority 0 is the highest priority and indicates pipeline that should be run first. This matches with the SOF scheduler implementation.

### DIFF
--- a/tools/topology/topology2/include/components/pipeline.conf
+++ b/tools/topology/topology2/include/components/pipeline.conf
@@ -57,10 +57,14 @@ Class.Widget."pipeline" {
 		}
 	}
 
-	# pipeline priority
+	# pipeline priority (0 = highest priority, 7 = lowest)
 	DefineAttribute."priority" {
 		# Token reference and type
 		token_ref	"scheduler.word"
+		constraints {
+			min	0
+			max	7
+		}
 	}
 
 	# core that the pipeline should be scheduled on


tools/topology/topology2/include/components/pipeline.conf
```diff
@@ -57,7 +57,7 @@ Class.Widget."pipeline" {
 		}
 	}
 
-	# pipeline priority
+	# pipeline priority (0 = highest priority)
```
btian1, line 60: can we add more? like priority range: 0 ~ ?

lgirdwood: ok, yeah, also useful to add "lowest = 7" iirc.

lgirdwood (commented):
Needed addition.

tools/topology/topology2/include/components/pipeline.conf
kv2019i: @btian1 @lgirdwood I agree but this is complex. 0 = highest applies to all implementations, but fine details of other semantics are still a  bit unclear to me, including what is the lowest. in IPC4 this is 7, but we don't have such for IPC3. but fair enough, let me investigate more to come up with the documentation.

kv2019i:
V2: 
- document the 0..7 range used by IPC4 (and covers all IPC3 usage as well)
- add explicit constraints (as we now can do this)

bardliao (approved):

kv2019i:
SOFCI TEST

PR #8513, opened 2023-11-22
topology: Add support for buffer flags

cujomalainey:
Flags have been a part of IPC3 for ages, add support for them in the topology.

### DIFF
--- a/src/include/kernel/tokens.h
+++ b/src/include/kernel/tokens.h
@@ -33,6 +33,7 @@
 /* buffers */
 #define SOF_TKN_BUF_SIZE			100
 #define SOF_TKN_BUF_CAPS			101
+#define SOF_TKN_BUF_FLAGS			102
 
 /* DAI */
 /* Token retired with ABI 3.2, do not use for new capabilities
--- a/tools/topology/topology1/m4/buffer.m4
+++ b/tools/topology/topology1/m4/buffer.m4
@@ -5,13 +5,16 @@ dnl Define the macro for buffer widget
 dnl N_BUFFER(name)
 define(`N_BUFFER', `BUF'PIPELINE_ID`.'$1)
 
-dnl W_BUFFER(name, size, capabilities)
+dnl W_BUFFER(name, size, capabilities, [core], [flags])
 define(`W_BUFFER',
 `SectionVendorTuples."'N_BUFFER($1)`_tuples" {'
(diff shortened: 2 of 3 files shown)


cujomalainey:
@andyross FYI

lgirdwood (approved):
Good find - I guess we missed that one. @ranj063 good for you ?

kv2019i (commented):
Code looks good but I wonder why we did not expose this in the end. The two flags we have in FW in buffer flags are SOF_BUF_UNDERRUN_PERMITTED and SOF_BUF_OVERRUN_PERMITTED . I have no personal recollection of what these are but these were added in 2020 in commit ee3f7d9737f00246a0ea881624c4335e5e3a5c3a and as far as I can tell, never used (either via IPC, or by setting the underrun/overrun permitted flag with internal logic).

Do we have some fresh need for these flags, or need to add more flags?

I ask because in IPC4, the buffers get added by FW and @marcinszkudlinski  has been working on the "audio stream" replacement of buffers -- to allow more complex implementations of buffers (e.g. to save SRAM by pooling memory of buffers in a single pipeline). So I wonder if such behaviour changing flags would better be put to the pipeline nodes versus in the connections (buffers).

OTOH, no harm, so if there's concrete need, I see no issue in proceeding with.

plbossart:
I could be wrong, IIRC the flags were added so that "complicated cases" such as the echo reference stream would be allowed to overrun if the stream is not opened by the host.
Indeed it's anyone's guess what this means for IPC4....

cujomalainey:
> I could be wrong, IIRC the flags were added so that "complicated cases" such as the echo reference stream would be allowed to overrun if the stream is not opened by the host. Indeed it's anyone's guess what this means for IPC4....

That was my assumption too, hence my comment in the other PR about "how is echo ref even working?"

ranj063:
> > I could be wrong, IIRC the flags were added so that "complicated cases" such as the echo reference stream would be allowed to overrun if the stream is not opened by the host. Indeed it's anyone's guess what this means for IPC4....
> 
> That was my assumption too, hence my comment in the other PR about "how is echo ref even working?"

@cujomalainey but how does this PR help? Adding this token in topology means the job is still only half done isnt it? Don't you also need a kernel counterpart?

ranj063 (approved):

PR #8518, opened 2023-11-24
topology2:cavs-es83x6: Add support to Es8326 codec + HDMI-in capture

CBala21:
Add Es8326 codec support and HDMI-in capture via I2S.


tools/topology/topology2/cavs-es83x6.conf
```diff
@@ -0,0 +1,254 @@
...
+			direction	capture
+
+			Object.Widget.dai-copier."1" {
+				dai_index	$HEADSET_SSP_DAI_INDEX
+				dai_type	"SSP"
+				type		dai_out
+				copier_type	"SSP"
+				stream_name	"$HEADSET_CODEC_NAME"
+				node_type	$I2S_LINK_INPUT_CLASS
+				Object.Base.audio_format.1 {
```
bardliao, line 203: Please use `Object.Base.input_audio_format` and `Object.Base.output_audio_format`

tools/topology/topology2/cavs-es83x6.conf
```diff
@@ -0,0 +1,254 @@
...
+				stream_name	'$HEADSET_CAPTURE_PCM_STREAM_NAME'
+				pcm_id $HEADSET_PCM_ID
+			}
+		}
+	]
+
+	io-gateway-capture [
+		{
+			index	6
+			direction	capture
```
bardliao, line 194: `direction capture` is not needed for `io-gateway-capture`

bardliao (commented):
Object.Base.audio_format is deprecated. Please use Object.Base.input_audio_format and Object.Base.output_audio_format

tools/topology/topology2/cavs-es83x6.conf
CBala21: Updated, please check the latest

tools/topology/topology2/cavs-es83x6.conf
CBala21: Corrected. please check the latest.

bardliao (approved):

lgirdwood (approved):

tools/topology/topology2/cavs-es83x6.conf
```diff
@@ -0,0 +1,259 @@
...
+
+# include HDMI config if needed.
+IncludeByKey.NUM_HDMIS {
+	"[3-4]"	"platform/intel/hdmi-generic.conf"
+}
+
+# include deep buffer config if buffer size is in 1 - 1000 ms.
+IncludeByKey.DEEPBUFFER_FW_DMA_MS{
+	"([1-9]|[1-9][0-9]|[1-9][0-9][0-9]|1000)" "platform/intel/deep-buffer.conf"
+}
```
plbossart, line 105: why do you need this?

CBala21: Really we are not using in our platform but try to make as similar topology(RT5682) and in future people can use it for other platforms. please let me know if you feel its unnecessary and want to remove it.

CBala21: Please check my above comments and do suggest. Thanks.

plbossart: well, so if the intent is to track an existing topology then adding a new one is really not the direction. There's no such thing as a reference and 'not drift too far', as soon as you fork a new topology it will bit-rot and diverge. 

You'd want to use the same setup really and add the HDMI input that is probably the only thing missing?

CBala21: @plbossart : As you suggested, i have removed the unused feature and uploaded the latest. please review and help to get it merge.

PR #8520, opened 2023-11-24
nxp: imx93_a55: Switch to timer domain and Zephyr native drivers

LaurentiuM1234:
With this PR, the i.MX93 platform will now start using the timer domain in conjunction with the Zephyr native drivers.

Zephyr dependencies:

1. https://github.com/zephyrproject-rtos/zephyr/pull/65671 (EDMA)
2. https://github.com/zephyrproject-rtos/zephyr/pull/65187 (DAI)
3. https://github.com/zephyrproject-rtos/zephyr/pull/65679 (removal of SAI3 and EDMA static mappings)
4. https://github.com/zephyrproject-rtos/zephyr/pull/65684 (DAI_DIR_{TX/RX} value adjustment)

**Important notes**:

1. Currently, the PR is split in multiple commits. This is done to ease the reviewing process, but most of the commits will be squashed into a single one to avoid breaking git bisect.
2. If git dependency 3 is merged before this PR and the Zephyr revision is bumped up this will lead to SOF no longer working on i.MX93 because the SAI and EDMA address spaces will no longer have any mappings, leading to translation faults.
3. The transition to timer domain is mandatory because the DMA domain is unusable with the Zephyr native drivers ATM.

### DIFF
--- a/app/boards/imx93_evk_mimx9352_a55.conf
+++ b/app/boards/imx93_evk_mimx9352_a55.conf
@@ -26,3 +26,5 @@ CONFIG_DMA_NXP_EDMA_ENABLE_HALFMAJOR_IRQ=y
 # needed because EDMA uses same interrupt for multiple
 # channels.
 CONFIG_SHARED_INTERRUPTS=y
+
+CONFIG_ZEPHYR_NATIVE_DRIVERS=y
--- a/app/boards/imx93_evk_mimx9352_a55.conf
+++ b/app/boards/imx93_evk_mimx9352_a55.conf
@@ -14,3 +14,15 @@ CONFIG_DCACHE_LINE_SIZE=64
 
 CONFIG_IMX93_A55=y
 CONFIG_TRACE=n
+
+# DAI-related configurations
+CONFIG_SAI_HAS_MCLK_CONFIG_OPTION=y
+CONFIG_SAI_IMX93_ERRATA_051421=y
+
+# DMA-related configurations
+CONFIG_DMA=y
+CONFIG_DMA_NXP_EDMA_ENABLE_HALFMAJOR_IRQ=y
+
+# needed because EDMA uses same interrupt for multiple
+# channels.
+CONFIG_SHARED_INTERRUPTS=y
--- a/app/boards/imx93_evk_mimx9352_a55.overlay
+++ b/app/boards/imx93_evk_mimx9352_a55.overlay
@@ -24,18 +24,6 @@
 		reg = <0x42430000 DT_SIZE_K(64)>;
 	};
 
-	sai3: memory@42660000 {
-		reg = <0x42660000 DT_SIZE_K(64)>;
-	};
-
-	edma2_ch0: memory@42010000 {
-		reg = <0x42010000 DT_SIZE_K(32)>;
-	};
-
-	edma2_ch1: memory@42018000 {
-		reg = <0x42018000 DT_SIZE_K(32)>;
-	};
-
 	outbox: memory@ce100000 {
 		reg = <0xce100000 DT_SIZE_K(4)>;
 	};
@@ -58,4 +46,22 @@
 	host_ram: memory@80000000 {
 		reg = <0x80000000 DT_SIZE_M(1024)>;
 	};
+
+	host_dma: dma {
+		compatible = "nxp,sof-host-dma";
+		dma-channels = <32>;
+		#dma-cells = <0>;
+	};
+};
+
+&sai3 {
+	tx-fifo-watermark = <2>;
+	rx-fifo-watermark = <96>;
+	fifo-depth = <96>;
+	rx-sync-mode = <1>;
+	status = "okay";
+};
+
+&edma4 {
+	status = "okay";
 };
--- a/src/platform/Kconfig
+++ b/src/platform/Kconfig
@@ -128,7 +128,6 @@ config IMX93_A55
 	select BUILD_OUTPUT_BIN
 	select ZEPHYR_LOG
 	select HOST_PTABLE
-	select DMA_DOMAIN
 	help
 	 Select if your target platform is imx93-compatible.
 
--- a/src/platform/imx93_a55/platform.c
+++ b/src/platform/imx93_a55/platform.c
@@ -118,12 +118,6 @@ int platform_init(struct sof *sof)
 	if (ret < 0)
 		return -ENODEV;
 
-	/* Initialize DMA domain */
-	sof->platform_dma_domain = zephyr_dma_domain_init(&sof->dma_info->dma_array[0],
-							  1,
-							  PLATFORM_DEFAULT_CLOCK);
-	zephyr_ll_scheduler_init(sof->platform_dma_domain);
-
 	/* Initialize IPC */
 	ipc_init(sof);
 
--- a/tools/topology/topology1/sof-imx93-wm8962.m4
+++ b/tools/topology/topology1/sof-imx93-wm8962.m4
@@ -58,14 +58,14 @@ dnl     period, priority, core, time_domain)
 DAI_ADD(sof/pipe-dai-playback.m4,
 	1, SAI, 3, sai3-wm8962,
 	PIPELINE_SOURCE_1, 2, s32le,
-	1000, 0, 0, SCHEDULE_TIME_DOMAIN_DMA)
+	1000, 0, 0, SCHEDULE_TIME_DOMAIN_TIMER)
 
 # capture DAI is SAI1 using 2 periods
 # Buffers use s32le format, with 48 frame per 1000us on core 0 with priority 0
 DAI_ADD(sof/pipe-dai-capture.m4,
 	2, SAI, 3, sai3-wm8962,
 	PIPELINE_SINK_2, 2, s32le,
-	1000, 0, 0)
+	1000, 0, 0, SCHEDULE_TIME_DOMAIN_TIMER)
 
 
 # PCM Low Latency, id 0
--- a/west.yml
+++ b/west.yml
@@ -43,7 +43,7 @@ manifest:
 
     - name: zephyr
       repo-path: zephyr
-      revision: 8858a024c0151e3b4cd35262bed935248211d6b1
+      revision: 66b475a3aa4d73f4ddd5be5a7a1e0f3c7028e539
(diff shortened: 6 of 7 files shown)


lgirdwood:
@LaurentiuM1234 I'm guessing someone from NXP will test  after Zephyr PRs merge and update topologies to use timer ?

LaurentiuM1234:
> @LaurentiuM1234 I'm guessing someone from NXP will test after Zephyr PRs merge and update topologies to use timer ?

The Zephyr PRs will only affect i.MX93 which has one topology that's updated in this PR

src/audio/host-zephyr.c
```diff
@@ -213,6 +215,19 @@ static int host_copy_one_shot(struct host_data *hd, struct comp_dev *dev, copy_c
...
+		/* SRC/DEST addresses have changed so the DMAC needs
+		 * to be re-configured.
+		 */
+		blk.source_address = local_elem->src;
+		blk.dest_address = local_elem->dest;
+		blk.block_size = local_elem->size;
+		/* head block currently points to a stack address so use new
+		 * block for the re-configuration to avoid any "surprises".
+		 */
+		hd->z_config.head_block = &blk;
```
lyakh, line 228: sorry, what exactly does your comment above mean? Indeed this code stores a stack address into a persistent object and that indeed isn't a good idea and should be fixed, so which new block is your comment referring to, that fixes any surprises?

LaurentiuM1234: Ah, what I wanted to say with that comment is that on each `host_copy_one_shot()` you use a new block placed on the stack (i.e: the `blk` variable).

LaurentiuM1234: Fixed. Hopefully it's a bit better now?

lyakh: @LaurentiuM1234 but what my comment actually was around, is that you shouldn't store (at least not permanently) a pointer to an object on stack (`blk`) in a permanent object (`hd->z_config`). So that assignment isn't a very good idea. You can make it kind of acceptable if you reset the pointer to NULL again *inside* the same function before leaving it like
```
if (IS_ENABLED(CONFIG_DMA_NXP_SOF_HOST_DMA)) {
    ...
    hd->z_config.head_block = &blk;
    ...
}
...
ret = dma_config(hd->chan->dma->z_dev, hd->chan->index, &hd->z_config);
/* *** clear the pointer to a stack object! *** */
hd->z_config.head_block = NULL;
```
assuming that your `dma_config()` doesn't "steal" that pointer to store it in a different persistent object, but that would be a bug of its own then.

LaurentiuM1234: ACK. Then I think I'm going to go for allocating the block configuration on the heap just like we do on `dai-zephyr`.

LaurentiuM1234: dcd2bd31bf0de9cd698f3342440d8151abcaa1a8 should address this now

src/audio/host-zephyr.c
```diff
@@ -842,11 +842,21 @@ int host_common_params(struct host_data *hd, struct comp_dev *dev,
 		hd->dma_buffer_size = audio_stream_get_size(&hd->dma_buffer->stream);
 	}
 
+	dma_block_cfg = rballoc(SOF_MEM_FLAG_COHERENT,
+				SOF_MEM_CAPS_RAM | SOF_MEM_CAPS_DMA,
+				sizeof(struct dma_block_config));
```
lyakh, line 857: please use `rzalloc()` and remove `memset()` below

src/audio/host-zephyr.c
lyakh: if that object has to be persistent, then yes, it has to be allocated on heap. But if it's only used by the function and from other functions, used from it, i.e. while the stack frame is valid, then you can keep it on stack and then reset the pointer to `NULL` at the end of the function.

src/audio/host-zephyr.c
LaurentiuM1234: done

LaurentiuM1234:
**V2 updates**

1. dma_block_cfg is now allocated on the heap.
2. `CONFIG_DAI=y` has been removed from i.MX93's configuration file. It's already enabled via `prf.conf`
3. Removed `CONFIG_SAI_HAS_XCR4_CHMOD` as a consequence to it being removed from the Zephyr driver.

keqiaozhang:
SOFCI TEST

LaurentiuM1234:
**V3 updates**

1. Removed selection of `CONFIG_DMA_DOMAIN` from `CONFIG_IMX93_A55`. DMA domain is no longer used on i.MX93 so no need to compile `zephyr_dma_domain.c`.

lgirdwood:
@LaurentiuM1234 @dbaluta I've penciled in for v2.9.

dbaluta:
There is some issues with compilation on Windows:

https://github.com/thesofproject/sof/actions/runs/7555242327/job/20569791161?pr=8520

LaurentiuM1234:
> There is some issues with compilation on Windows:
> 
> https://github.com/thesofproject/sof/actions/runs/7555242327/job/20569791161?pr=8520

That's normal ATM. Since we're still missing the EDMA driver, I can't bump up the Zephyr revision.

dbaluta:
Thanks! Now i see the DNM tag.

app/boards/mimx93_evk_a55.overlay
```diff
@@ -58,4 +46,47 @@
...
+		fifo-depth = <96>;
+		interrupt-parent = <&gic>;
+		interrupts = <GIC_SPI 171 IRQ_TYPE_LEVEL IRQ_DEFAULT_PRIORITY>;
+		dai-index = <3>;
+	};
+
+	edma4: dma@42000000 {
+		compatible = "nxp,edma";
+		reg = <0x42000000 (DT_SIZE_K(64) * 32)>;
+		valid-channels = <0>, <1>;
```
iuliana-prodan, line 79: If you have here `<0>, <1>` shouldn't be `dma-channels` based on this: https://github.com/zephyrproject-rtos/zephyr/pull/65671/commits/bba525577eb0e0528e1460cd2e43aa952958400d#diff-7e174a2aefc1b54220f6b815c04028a782f004bc4d5d03a39f604703f09b3234R16?

app/boards/mimx93_evk_a55.overlay
```diff
@@ -58,4 +46,47 @@
...
+		 * request a DMA transfer every 1ms (which is the period size)
+		 */
+		rx-fifo-watermark = <96>;
+		/* set to 1 because we want to avoid multiple DMA transfer requests
+		 * when the SAI is started. This way, the SAI will request data every
+		 * 1ms (which is the period size).
+		 */
+		tx-fifo-watermark = <1>;
+		/* can't let SOF use the actual FIFO depth (128 FIFO words) because
+		 * the DMA transfer size isn't divizible by it so the driver would
```
iuliana-prodan, line 66: s/divizible/divisible

src/audio/host-zephyr.c
```diff
@@ -878,12 +890,11 @@ int host_common_params(struct host_data *hd, struct comp_dev *dev,
 	hd->chan->period = config->period;
 
 	memset(dma_cfg, 0, sizeof(*dma_cfg));
-	memset(&dma_block_cfg, 0, sizeof(dma_block_cfg));
```
iuliana-prodan, line 903: You can do the allocation for `dma_block_cfg` here, and there's no need to add `rfree(dma_block_cfg)` above (line 858 and 877)

app/boards/mimx93_evk_a55.overlay
iuliana-prodan: > Do you mean having something like: `dma-channels = <2>` instead of `valid-channels = <0>, <1>`? 

Yes.

>If so, both approaches are correct. For a (**very**) high number of channels, the `dma-channels` "way" would be better since the lookup function would be O(1), but since this will never be the case I'm not sure choosing one approach over the other would make that much difference. If you believe the `dma-channels` approach is more intuitive then I don't mind switching to it.

Ok, let's keep` valid-channels`, is not much difference for 2 channels. 
We'll change the list (`<0>, <1>`) when we'll have support for multiple channels and the update will be easier having already the `valid-channels` option.

src/audio/host-zephyr.c
LaurentiuM1234: fixed, thanks!

LaurentiuM1234:
**V5 updates**:

1. Most Zephyr dependencies have been merged, added west.yml bump. Waiting on https://github.com/zephyrproject-rtos/zephyr/pull/67907

LaurentiuM1234:
**V6 updates**

1. Dependencies merged, updated Zephyr manifest and 64-bit CI build is now clean.
2. Switched to RX SYNC - TX ASYNC mode. See comments from overlay for explanation.

A few notes:

1. Current PR layout will break git bisect. Since 93 is not officially supported in the kernel and is mostly used to test out new stuff I would be inclined to keep the layout as-is as long as this won't affect anyone else.
2. Bunch of CI failures, some of them caused by the recent SMP changes from Zephyr.
3. Need to re-test everything to make sure the native switch works ok. Will lift the DNM tag once this is done and the SMP-related issues are fixed. Also need to make sure that this PR isn't a cause for the other non-SMP-related CI failures.

LaurentiuM1234:
@lgirdwood @kv2019i do we, by any chance, have an ETA on the fix for the SMP-related issues? It would be very helpful. Thanks!

kv2019i:
We are very close to unblocking Zephyr updates now, but as there are still issues, we'll do SOF2.9 with old Zephyr baseline. I'll bump this to 2.10.

LaurentiuM1234:
**V7 updates**

1. EDMA4 and SAI3 nodes have been moved to Zephyr (see https://github.com/zephyrproject-rtos/zephyr/pull/70045). All we do here is enable and configure them for SOF usage.
2. Removed commits that were already merged via https://github.com/thesofproject/sof/pull/8859
3. Removed cleanup commits. This will be handled separately.

LaurentiuM1234:
**V8 updates**

1) All dependencies have been merged. Removing the [DNM] tag.
2) Added manifest update

dbaluta (approved):

PR #8529, opened 2023-11-27
Fix i.MX93 compilation warnings related to mis-formatting / casting 64-bit types

LaurentiuM1234:
Currently, building SOF on a 64-bit system (i.e: i.MX93) leads to compilation warnings such as:

"format '%d' expects argument of type 'int', but argument 5 has type 'long unsigned int'" or

"cast to pointer from integer of different size"

This is because pointers are 64-bit and size_t resolves to a 64-bit type. As such, casting a pointer to uint32_t when printing leads to int-to-pointer-cast warnings. Also, printing a size_t variable using the "%d" or "%u" formats leads to format warnings.

To fix this, cast all size_t variables to uint32_t when printing. Pointers shall also be casted to uintptr_t before casting to uint32_t. This is fine to do because i.MX93 uses 32-bit addresses anyways.

This in conjunction with https://github.com/thesofproject/sof/pull/8528 fixes all compilation warnings for i.MX93.
This fix is not that great since it forces 64-bit types to 32-bit and counts on the fact that you won't have values that need 64-bit encoding, but it's a start I guess.

### DIFF
--- a/src/audio/data_blob.c
+++ b/src/audio/data_blob.c
@@ -184,7 +184,7 @@ int comp_data_blob_set(struct comp_data_blob_handler *blob_handler,
 	}
 #endif
 
-	comp_dbg(blob_handler->dev, "comp_data_blob_set_cmd() pos = %d, fragment size = %d",
+	comp_dbg(blob_handler->dev, "comp_data_blob_set_cmd() pos = %d, fragment size = %zu",
 		 pos, fragment_size);
 
 	/* Check that there is no work-in-progress previous request */
--- a/src/audio/module_adapter/module/generic.c
+++ b/src/audio/module_adapter/module/generic.c
@@ -30,8 +30,8 @@ int module_load_config(struct comp_dev *dev, const void *cfg, size_t size)
 	comp_dbg(dev, "module_load_config() start");
 
 	if (!cfg || !size) {
-		comp_err(dev, "module_load_config(): wrong input params! dev %x, cfg %x size %zu",
-			 (uint32_t)dev, (uint32_t)cfg, size);
+		comp_err(dev, "module_load_config(): wrong input params! dev %zx, cfg %zx size %zu",
+			 (size_t)dev, (size_t)cfg, size);
 		return -EINVAL;
 	}
 
--- a/src/audio/module_adapter/module/generic.c
+++ b/src/audio/module_adapter/module/generic.c
@@ -30,7 +30,7 @@ int module_load_config(struct comp_dev *dev, const void *cfg, size_t size)
 	comp_dbg(dev, "module_load_config() start");
 
 	if (!cfg || !size) {
-		comp_err(dev, "module_load_config(): wrong input params! dev %x, cfg %x size %d",
+		comp_err(dev, "module_load_config(): wrong input params! dev %x, cfg %x size %zu",
 			 (uint32_t)dev, (uint32_t)cfg, size);
 		return -EINVAL;
 	}
@@ -448,7 +448,7 @@ int module_set_configuration(struct processing_module *mod,
 			return 0;
 
 		if (md->new_cfg_size > MAX_BLOB_SIZE) {
-			comp_err(dev, "module_set_configuration(): error: blob size is too big cfg size %d, allowed %d",
+			comp_err(dev, "module_set_configuration(): error: blob size is too big cfg size %zu, allowed %d",
 				 md->new_cfg_size, MAX_BLOB_SIZE);
 			return -EINVAL;
 		}
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -47,8 +47,8 @@ struct comp_dev *module_adapter_new(const struct comp_driver *drv,
 	comp_cl_dbg(drv, "module_adapter_new() start");
 
 	if (!config) {
-		comp_cl_err(drv, "module_adapter_new(), wrong input params! drv = %x config = %x",
-			    (uint32_t)drv, (uint32_t)config);
+		comp_cl_err(drv, "module_adapter_new(), wrong input params! drv = %zx config = %zx",
+			    (size_t)drv, (size_t)config);
 		return NULL;
 	}
 
--- a/src/audio/eq_fir/eq_fir.c
+++ b/src/audio/eq_fir/eq_fir.c
@@ -262,7 +262,7 @@ static int eq_fir_init(struct processing_module *mod)
 	 * blob size is sane.
 	 */
 	if (bs > SOF_EQ_FIR_MAX_SIZE) {
-		comp_err(dev, "eq_fir_init(): coefficients blob size = %u > SOF_EQ_FIR_MAX_SIZE",
+		comp_err(dev, "eq_fir_init(): coefficients blob size = %zu > SOF_EQ_FIR_MAX_SIZE",
 			 bs);
 		return -EINVAL;
 	}
--- a/src/audio/eq_iir/eq_iir.c
+++ b/src/audio/eq_iir/eq_iir.c
@@ -59,7 +59,7 @@ static int eq_iir_init(struct processing_module *mod)
 
 	/* Check first before proceeding with dev and cd that coefficients blob size is sane */
 	if (bs > SOF_EQ_IIR_MAX_SIZE) {
-		comp_err(dev, "eq_iir_init(), coefficients blob size %u exceeds maximum", bs);
+		comp_err(dev, "eq_iir_init(), coefficients blob size %zu exceeds maximum", bs);
 		return -EINVAL;
 	}
 
--- a/src/audio/kpb.c
+++ b/src/audio/kpb.c
@@ -414,7 +414,7 @@ static int kpb_set_verify_ipc_params(struct comp_dev *dev,
 		       ipc_config->size);
 
 	if (ret) {
-		comp_err(dev, "kpb_new(): cannot memcpy_s %d bytes into sof_kpb_config (%d)\n",
+		comp_err(dev, "kpb_new(): cannot memcpy_s %d bytes into sof_kpb_config (%zu)\n",
 			 ipc_config->size, sizeof(kpb->config));
 		return -EINVAL;
 	}
@@ -478,7 +478,7 @@ static struct comp_dev *kpb_new(const struct comp_driver *drv,
 
 	/* make sure data size is not bigger than config space */
 	if (ipc_config_size > kpb_config_size) {
-		comp_cl_err(&comp_kpb, "kpb_new(): ipc config size %u too big",
+		comp_cl_err(&comp_kpb, "kpb_new(): ipc config size %zu too big",
 			    ipc_config_size);
 		return NULL;
 	}
@@ -579,7 +579,7 @@ static size_t kpb_allocate_history_buffer(struct comp_data *kpb,
 			/* We managed to allocate a block of ca_size.
 			 * Now we initialize it.
 			 */
-			comp_cl_info(&comp_kpb, "kpb new memory block: %d",
+			comp_cl_info(&comp_kpb, "kpb new memory block: %zu",
 				     ca_size);
 			allocated_size += ca_size;
 			hb->start_addr = new_mem_block;
@@ -628,7 +628,7 @@ static size_t kpb_allocate_history_buffer(struct comp_data *kpb,
 		}
 	}
 
-	comp_cl_info(&comp_kpb, "kpb_allocate_history_buffer(): allocated %d bytes",
+	comp_cl_info(&comp_kpb, "kpb_allocate_history_buffer(): allocated %zu bytes",
 		     allocated_size);
 
 	return allocated_size;
@@ -1339,7 +1339,7 @@ static int kpb_copy(struct comp_dev *dev)
 
 			comp_update_buffer_consume(source, copy_bytes);
 		} else {
-			comp_warn(dev, "kpb_copy(): buffering skipped (no data to copy, avail %d, free %d",
+			comp_warn(dev, "kpb_copy(): buffering skipped (no data to copy, avail %d, free %zu",
 				  audio_stream_get_avail_bytes(&source->stream),
 				  kpb->hd.free);
 		}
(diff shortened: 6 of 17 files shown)


dbaluta (approved):

lgirdwood (approved):
Don't think we can do this yet as we are more optimised for size vs portability.

https://stackoverflow.com/questions/1403074/printf-with-sizeof-on-32-vs-64-platforms-how-do-i-handle-format-code-in-platfor

@LaurentiuM1234 @dbaluta btw, can get this platform as a build check in CI ?

LaurentiuM1234:
> Don't think we can do this yet as we are more optimised for size vs portability.

Understood. Frankly, I think forcing 64-bit types to 32-bit is not really the way to go so if anyone thinks this is a bad idea or opposes to this being merged then I'm all for dropping this and dealing with it when it becomes necessary. It would have been nice to get rid of all of those annoying warnings as it's become very hard to find new warnings in the sea of already existing warnings.


> @LaurentiuM1234 @dbaluta btw, can get this platform as a build check in CI ?

we already have a CI job building SOF for i.MX93. It's just that it's separate from the one building it for all other i.MX platforms.

LaurentiuM1234:
Marking this as a draft for now as I'm not sure about this solution

marc-hb:
Anything wrong with `%zd`? I haven't actually run it but it seems to compile with both Zephyr and XTOS.

marc-hb:
> Anything wrong with %zd? I haven't actually run it but it seems to compile with both Zephyr and XTOS.

This probably requires skipping the `z` character in `sof/tools/logger/convert.c#process_params()`. The old sof-logger only supports 32bits anyway.

As long as you don't plan to actually use `sof/tools/logger` with `aarch64` then it should be fine.

src/audio/data_blob.c
```diff
@@ -185,7 +185,7 @@ int comp_data_blob_set(struct comp_data_blob_handler *blob_handler,
 #endif
 
 	comp_dbg(blob_handler->dev, "comp_data_blob_set_cmd() pos = %d, fragment size = %d",
-		 pos, fragment_size);
+		 pos, (uint32_t)fragment_size);
```
lyakh, line 188: how about just using `%zd` or `%zu`? Same everywhere for `ssize_t` or `size_t` printing respectively

lyakh (commented):
let's use `%zu` please

src/audio/data_blob.c
LaurentiuM1234: Done. Went for `%zu` as `ssize_t` doesn't seem to be used anywhere in the places that the compiler raises warnings for.

dbaluta (approved):

LaurentiuM1234:
> This probably requires skipping the `z` character in `sof/tools/logger/convert.c#process_params()`. The old sof-logger only supports 32bits anyway.
> 
> As long as you don't plan to actually use `sof/tools/logger` with `aarch64` then it should be fine.

sof-logger won't be used on i.MX93 but it's currently being used on the other i.MX boards. I've tested the logger on i.MX8QM and seems like it works although this is probably because my test scenario doesn't reach the modified print statements. @dbaluta @iuliana-prodan what's your take on this? Do we proceed with this and risk breaking sof-logger? Theoretically we should be ready to switch to serial debugging via Zephyr on all boards (except ULP). I'm assuming this won't break the internal tests as afaik sof-logger isn't used there.

I'm assuming Intel no longer uses sof-logger, right?

iuliana-prodan:
> > This probably requires skipping the `z` character in `sof/tools/logger/convert.c#process_params()`. The old sof-logger only supports 32bits anyway.
> > As long as you don't plan to actually use `sof/tools/logger` with `aarch64` then it should be fine.
> 
> sof-logger won't be used on i.MX93 but it's currently being used on the other i.MX boards. I've tested the logger on i.MX8QM and seems like it works although this is probably because my test scenario doesn't reach the modified print statements. @dbaluta @iuliana-prodan what's your take on this? Do we proceed with this and risk breaking sof-logger (assuming this would even break it)? Theoretically we should be ready to switch to serial debugging via Zephyr on all boards (except ULP)(also, this isn't the case for XTOS). I'm assuming this won't break the internal tests as afaik sof-logger isn't used there.
> 
> I'm assuming Intel no longer uses sof-logger, right?
> 
> **EDIT**: this may impact MTK and AMD though but I'm not 100% sure what their setup is so I can't really tell.

You tested sof-logger, with these changes on 8QM and is ok, right?
imx93 is the closest to zephyr native drivers so we will use serial debugger there.
And, as @marc-hb pointed out, the sof-logger might break on arch64, so for the rest it should be fine.

Therefore, I'm approving this!

iuliana-prodan (approved):

LaurentiuM1234:
> You tested sof-logger, with these changes on 8QM and is ok, right?

Yep, seems like it works fine, but please keep in mind that I've only tested a couple of aplays and nothing more so I can't guarantee that this won't break sof-logger

> And, as @marc-hb pointed out, the sof-logger might break on arch64, so for the rest it should be fine.

ACK.

If everyone's OK with this then I'm going to remove from draft. As for the pointer downcast to 32-bit: we should be OK for now as there's no 64-bit platform using addresses encoded in 64 bits.

marc-hb:
> I've tested the logger on i.MX8QM and seems like it works although this is probably because my test scenario doesn't reach the modified print statements

Can you modify one statement that is actually in use (something outside error handling) and print some magic number with `%zu` in that statement? If that one works and that number is logged properly then I think you can blindly change the other ones.

> I'm assuming Intel no longer uses sof-logger, right?

Not in the main branch, no. Still in use in stable-v2.2.

src/audio/module_adapter/module_adapter.c
```diff
@@ -48,7 +48,7 @@ struct comp_dev *module_adapter_new(const struct comp_driver *drv,
 
 	if (!config) {
 		comp_cl_err(drv, "module_adapter_new(), wrong input params! drv = %x config = %x",
-			    (uint32_t)drv, (uint32_t)config);
+			    (uint32_t)(uintptr_t)drv, (uint32_t)(uintptr_t)config);
```
marc-hb, line 51: Double casts are very rarely needed so they need at least a short comment/justification. I see you mentioned something in the commit message but that's too far away from the code.

Also: why not use `%zu` / `(size_t)` here?

src/audio/module_adapter/module/generic.c
```diff
@@ -30,8 +30,8 @@ int module_load_config(struct comp_dev *dev, const void *cfg, size_t size)
 	comp_dbg(dev, "module_load_config() start");
 
 	if (!cfg || !size) {
-		comp_err(dev, "module_load_config(): wrong input params! dev %x, cfg %x size %d",
-			 (uint32_t)dev, (uint32_t)cfg, size);
+		comp_err(dev, "module_load_config(): wrong input params! dev %x, cfg %x size %zu",
+			 (uint32_t)(uintptr_t)dev, (uint32_t)(uintptr_t)cfg, size);
```
marc-hb, line 34: See above

src/include/sof/drivers/mu.h
```diff
@@ -72,12 +72,12 @@ enum imx_mu_xsr {
...
 static inline uint32_t imx_mu_read(uint32_t reg)
 {
-	return *((volatile uint32_t*)(MU_BASE + reg));
+	return *((volatile uint32_t*)((uintptr_t)(MU_BASE + reg)));
 }
 
 static inline void imx_mu_write(uint32_t val, uint32_t reg)
 {
-	*((volatile uint32_t*)(MU_BASE + reg)) = val;
+	*((volatile uint32_t*)((uintptr_t)(MU_BASE + reg))) = val;
```
marc-hb, line 80: Again, MU_BASE is a literal so it is really not intuitive why this is needed. Please add a short explanation in the code (and this one does not seem explained by the commit message)

src/audio/data_blob.c
```diff
@@ -184,7 +184,7 @@ int comp_data_blob_set(struct comp_data_blob_handler *blob_handler,
 	}
 #endif
 
-	comp_dbg(blob_handler->dev, "comp_data_blob_set_cmd() pos = %d, fragment size = %d",
+	comp_dbg(blob_handler->dev, "comp_data_blob_set_cmd() pos = %d, fragment size = %zu",
```
btian1, line 187: how about lld?

marc-hb: zd is arch-dependent, lld is typically not.

src/include/sof/drivers/mu.h
marc-hb: Please do this instead (tested)

```diff
--- a/src/platform/imx93_a55/include/platform/lib/memory.h
+++ b/src/platform/imx93_a55/include/platform/lib/memory.h
@@ -49,7 +49,7 @@
 #define HEAPMEM_SIZE 0x00010000
 
 /* SOF uses A side of the WAKEUPMIX MU */
-#define MU_BASE 0x42430000
+#define MU_BASE ((uintptr_t)0x42430000)
 
 /* SOF uses EDMA2 (a.k.a EDMA4 in the TRM) */
 #define EDMA2_BASE 0x42010000

```

Fun fact: C++23 allows `0x42430000Z`

marc-hb (dismissed):
> > I've tested the logger on i.MX8QM and seems like it works although this is probably because my test scenario doesn't reach the modified print statements

> Can you modify one statement that is actually in use (something outside error handling) and print some magic number with `%zu` in that statement? If that one works and that number is logged properly then I think you can blindly change the other ones.

Please test `%zu` with the sof-logger "for real" in at least once place before merging this great-looking but massive search/replace.

marc-hb:
The only failure in https://sof-ci.01.org/sof-pr-viewer/#/build/PR8529/build13187149 (to which you don't have access, sorry about that) was a "west update" failure, so don't worry about it.

src/audio/module_adapter/module_adapter.c
lyakh: one could use `PRIxPTR` to print `uintptr_t` directly

dbaluta:
@LaurentiuM1234 Let's also fix this:

```
WARNING: line length of 101 exceeds 100 columns
#190: FILE: src/audio/mux/mux.c:218:
+		comp_err(dev, "mux_init(): blob size %zu exceeds %zu", cfg->size, MUX_BLOB_MAX_SIZE);
```

LaurentiuM1234:
> Please test `%zu` with the sof-logger "for real" in at least once place before merging this great-looking but massive search/replace.

With the following change:

```
diff --git a/src/ipc/ipc3/handler.c b/src/ipc/ipc3/handler.c
index b7884fcda..b41038933 100644
--- a/src/ipc/ipc3/handler.c
+++ b/src/ipc/ipc3/handler.c
@@ -1637,6 +1637,8 @@ void ipc_cmd(struct ipc_cmd_hdr *_hdr)
                tr_info(&ipc_tr, "ipc: new cmd 0x%x", hdr->cmd);
        }
 
+       tr_err(&ipc_tr, "Hello, world with number %zu", 0xdeadbeef);
+
        type = iGS(hdr->cmd);
 
        switch (type) {
```

I get the following messages in the logger (trace, -t):

```
src/ipc/ipc3/handler.c:1640 ERROR Hello, world with number 3735928559
```

which seems fine.

Without trace (-t), the logger has the following output:

```
rerror: Invalid filename length 1852399458 or ldc file does not match firmware
error: read_entry_from_ldc_file(0x20013100) returned -22
error: fetch_entry() failed with: -22, aborting
Skipped 0 bytes after the last statement. Potential mailbox wrap, check the start of the output for later logs.
```

which doesn't seem to be related.

src/audio/module_adapter/module_adapter.c
LaurentiuM1234: Thanks for the suggestions! Decided to go with Marc's suggestion for now. Avoiding the PRI* macros for now because of some interesting points from https://github.com/zephyrproject-rtos/zephyr/issues/46032.

src/include/sof/drivers/mu.h
LaurentiuM1234: Fixed this by adding the `ULL` suffix. This ought to make `MU_BASE` 64-bit.

LaurentiuM1234:
> @LaurentiuM1234 Let's also fix this:
> 
> ```
> WARNING: line length of 101 exceeds 100 columns
> #190: FILE: src/audio/mux/mux.c:218:
> +		comp_err(dev, "mux_init(): blob size %zu exceeds %zu", cfg->size, MUX_BLOB_MAX_SIZE);
> ```

should be fixed with cc4eb17bb37f5fd343ace2509ebede974891f289. We could also squash this into the initial PR that causes this issue if you'd like.

dbaluta:
> should be fixed with [cc4eb17](https://github.com/thesofproject/sof/commit/cc4eb17bb37f5fd343ace2509ebede974891f289). We could also squash this into the initial PR that causes this issue if you'd like.

I'm fine that no need to squash it.

marc-hb (approved):

marc-hb:
> I'm fine that no need to squash it.

Checkpatch is only checking... submitted patches one by one. So submitting a separate fixup commit will not turn things red in this PR and it will not help other, unrelated PRs either.

A separate fixup commit will only help if/when someone changes the same line again. But then they're likely to break that line themselves anyway.

LaurentiuM1234:
> > I'm fine that no need to squash it.
> 
> Checkpatch is only checking... submitted patches one by one. So submitting a separate fixup commit will not turn things green in this PR and it will not help other, unrelated PRs either.
> 
> A separate fixup commit will only help if/when someone changes the same line again. But then they're likely to break that line themselves anyway.

squashed it after all. The issue was introduced by b1fabd45ff55cd944903d82b20198bea240a987f anyways.

keqiaozhang:
SOFCI TEST

kv2019i (approved):

kv2019i:
Ready to go, waiting for CI to complete, but based on earlier runs (and PR content), this should pass.

src/audio/module_adapter/module/generic.c
```diff
@@ -30,8 +30,8 @@ int module_load_config(struct comp_dev *dev, const void *cfg, size_t size)
 	comp_dbg(dev, "module_load_config() start");
 
 	if (!cfg || !size) {
-		comp_err(dev, "module_load_config(): wrong input params! dev %x, cfg %x size %zu",
-			 (uint32_t)dev, (uint32_t)cfg, size);
+		comp_err(dev, "module_load_config(): wrong input params! dev %zx, cfg %zx size %zu",
+			 (size_t)dev, (size_t)cfg, size);
```
lyakh, line 34: why are we actually not just printing these as pointers per `%p`?

marc-hb: I remember some confusion about whether sof-logger supports `%p` (I think it does)

kv2019i:
The actual build part of "sof-ci/jenkins/pr-build" seems have been successful, so I'll proceed to merge this.

marc-hb:
Awesome, now that we have a warnings-free build, let's keep it that way. Please review very small -Werror PR:
- #8559

PR #8531, opened 2023-11-27
rimage: manifest: Improve handling of empty segments

softwarecki:
Fixed output segment size calculation. If a given output segment type did not contain any input section (was empty), its size was incorrectly calculated as 1 and then rounded to the page size (4096). This resulted in an unnecessary segment in a output file.

Improved handling of empty RODATA segment. If the RODATA segment does not contain any data, its type is now marked as empty. Cleared the readonly flag of RODATA segment because this segment also contains .data sections.


pjdobrowolski (approved):

tools/rimage/src/module.c
```diff
@@ -252,7 +249,7 @@ static void sections_info_add(struct module_sections_info *info, struct module_s
  */
 static void sections_info_finalize(struct module_sections_info *info)
 {
-	info->file_size = info->end - info->start;
+	info->file_size = (info->end > info->start) ? info->end - info->start : 0;
```
lyakh, line 252: sorry, don't understand. You say, previously this produced the result of 1, i.e. `info->end - info->start == 1` which means `info->end > info->start` so now this still would calculate 1. What am I missing?

tools/rimage/src/manifest.c
```diff
@@ -368,14 +369,26 @@ static int man_module_create(struct image *image, struct manifest_module *module
...
 	pages = module->file.data.file_size / MAN_PAGE_SIZE;
 
 	man_module->segment[SOF_MAN_SEGMENT_RODATA].flags.r.length = pages;
+	if (pages) {
+		man_module->segment[SOF_MAN_SEGMENT_RODATA].v_base_addr = module->file.data.start;
+		man_module->segment[SOF_MAN_SEGMENT_RODATA].file_offset = module->foffset +
+									  module->text_fixup_size;
+		/* Copy data sections content */
+		err = man_copy_elf_sections(image, module, &man_module->segment[SOF_MAN_SEGMENT_RODATA],
+					    module->file.data.first_section);
```
lyakh, line 383: looks like you forgot to delete lines 393-397 below

lyakh (changes requested):

tools/rimage/src/module.c
softwarecki: The initial state is
`info->start = UINT32_MAX;
info->end = 0;`

lyakh: is there a particular reason for that? Could you just initialise both to 0?

softwarecki: The segment size increases as more sections are added from the input file. When adding, the `start` and `end` markers are moved. The `start` decreases if the section starts earlier than the current start. The `end` increases if the end of the section extends beyond the current segment end. This is a classic min(start) and max(end) algorithms, hence the initial values.

softwarecki:
@lyakh 
> We only have one section in the manifest, that contains both .data and .rodata, right? This is in fact not good, can this be fixed?

I'm afraid there is no good solution for this. The manifest only supports max three segments. Currently we use them like this:
1. `.text`
2. `.rodata` `.data`
3. .`bss`
I think this is due to the assumption that loadable modules don't have a `.data` section, only `.text`, `.rodata` and `.bss`.

lyakh:
> @lyakh
> 
> > We only have one section in the manifest, that contains both .data and .rodata, right? This is in fact not good, can this be fixed?
> 
> I'm afraid there is no good solution for this. The manifest only supports max three segments. Currently we use them like this:
> 
>     1. `.text`
> 
>     2. `.rodata` `.data`
> 
>     3. .`bss`
>        I think this is due to the assumption that loadable modules don't have a `.data` section, only `.text`, `.rodata` and `.bss`.

I don't really understand this assumption - modules are allowed to have .rodata, i.e. `const` objects, e.g. operations, and they're allowed to have .bss, i.e. writable initialised to 0 data, why aren't they then supposed to have writable but initialised to non-0 values data?

tools/rimage/src/module.c
softwarecki: @lyakh The `size` variable is initialized to 0. The `start` marker is initialized to the maximum value to simplify calculation of the minimum starting address, and to distinguish a valid value of 0 from the initial state.

lgirdwood:
> > @lyakh
> > > We only have one section in the manifest, that contains both .data and .rodata, right? This is in fact not good, can this be fixed?
> > 
> > 
> > I'm afraid there is no good solution for this. The manifest only supports max three segments. Currently we use them like this:
> > ```
> > 1. `.text`
> > 
> > 2. `.rodata` `.data`
> > 
> > 3. .`bss`
> >    I think this is due to the assumption that loadable modules don't have a `.data` section, only `.text`, `.rodata` and `.bss`.
> > ```
> 
> I don't really understand this assumption - modules are allowed to have .rodata, i.e. `const` objects, e.g. operations, and they're allowed to have .bss, i.e. writable initialised to 0 data, why aren't they then supposed to have writable but initialised to non-0 values data?

I think we have been merging .rodata and .data in the same manifest section to date using the linker and rimage.

keqiaozhang:
SOFCI TEST

tools/rimage/src/manifest.c
```diff
@@ -240,20 +240,21 @@ static int man_get_module_manifest(struct image *image, struct manifest_module *
...
 	/* data segment */
 	segment = &man_module->segment[SOF_MAN_SEGMENT_RODATA];
 	segment->flags.r.contents = 1;
 	segment->flags.r.alloc = 1;
 	segment->flags.r.load = 1;
-	segment->flags.r.readonly = 1;
+	segment->flags.r.readonly = 0; /* rodata segment contains also writtable data */
 	segment->flags.r.data = 1;
-	segment->flags.r.type = 1;
+	segment->flags.r.type = SOF_MAN_SEGMENT_RODATA;
```
lyakh, line 252: How immutable is this ABI? Can it be extended to have both .data and .rodata? At least we could rename `SOF_MAN_SEGMENT_RODATA` to something that isn't confusing

softwarecki: Hard to say. The rimage uses fixed segment indexes for specific type, module loading in zephyr is based on information about the segment type/flags without taking the index into account. The library manager ignores types/flags and assumes that the specified type is at a specific fixed index. In #8544 I modified the library manager to use segment type/flags instead of const index, so theoretically we can separate `.data` from `.rodata`, but the manifest only supports 3 segments! I don't know how the rom behaves, but I suspect that it was easier to implement flag support rather than constant indexes.

kv2019i (approved):
Looks ok. @lyakh please take a look at the clarifications.

tools/rimage/src/manifest.c
lyakh: would be good to clarify and hopefully fix this

PR #8538, opened 2023-11-27
Fix support for hybrid host-dsp pipeline with SOF plugin

ranj063:


ranj063:
> do we actually have any CI testing of the plugin? Building / run-time?

not yet but I'm working to add build test soon

lgirdwood (approved):

RanderWang (approved):

lgirdwood:
@lrudyX @wszypelt this code is not tested by CI today, are we good to merge ?

src/audio/copier/ipcgtw_copier.h
```diff
@@ -95,13 +95,22 @@ struct ipc4_ipc_gateway_cmd_data_reply {
...
-			 struct sof_ipc_stream_params *params);
-
-void copier_ipcgtw_reset(struct comp_dev *dev);
-
 int copier_ipcgtw_create(struct comp_dev *dev, struct copier_data *cd,
 			 const struct ipc4_copier_module_cfg *copier, struct pipeline *pipeline);
 
+#if CONFIG_IPC4_GATEWAY
 void copier_ipcgtw_free(struct copier_data *cd);
+int copier_ipcgtw_params(struct ipcgtw_data *ipcgtw_data, struct comp_dev *dev,
```
kv2019i, line 103: Shouldn't this be fixed in copier.c, or if done here, ifdef out all copier_ipcgtw_foo calls? This feels a bit adhoc to ifdef some functions out. Might be easier to verify in copier.c that code still works correctly if the feature is disabled (or gives proper errors if we reach branches of code that we really shouldn't when gateway support is not built in).

tools/plugin/modules/alsa.c
```diff
@@ -266,23 +266,7 @@ static int set_params(struct comp_dev *dev)
 		return err;
```
kv2019i, line 266: Little light on the git commit on why capture and playback are handled differently.

tools/plugin/modules/alsa.c
```diff
@@ -596,11 +604,14 @@ static int aplay_copy(struct comp_dev *dev)
 		/* write PCM samples to PCM */
 		frames = snd_pcm_writei(cd->handle, pos, frames);
 		if (frames < 0) {
+			if (frames == -EPIPE) {
```
kv2019i, line 607: Should we have some debug/error print for xruns?

kv2019i (commented):
Some comments inline.

lrudyX:
> @lrudyX @wszypelt this code is not tested by CI today, are we good to merge ?

Don't see any blockers, good to merge.

marc-hb:
> @lrudyX @wszypelt this code is not tested by CI today, are we good to merge ?

It is compile-tested in #8560, better merge 8560 first.

tools/plugin/README.md
```diff
@@ -27,7 +27,7 @@ IPC3 is not functional and not supported
 cd sof
 mkdir build_plugin
 cd build_plugin
-cmake ../tools/plugin -DPLUGIN_IPC4=ON
+cmake ../tools/plugin
```
marc-hb, line 30: Please drop this commit, see #8560

marc-hb (commented):
Interesting that the first commit says "fix compilcation errors" while I did not see any nor did https://github.com/thesofproject/sof/actions/runs/7052569908/job/19197888072?pr=8560

CONFIG_ differences maybe?

marc-hb:
> Another build race condition? Really does not look like one.

Yet sometimes it fails and sometimes it builds in my workspace too - on the same commit.

marc-hb:
> Yet sometimes it fails and sometimes it builds in my workspace too - on the same commit.

The best place to observe how often it failed in Github is https://github.com/thesofproject/sof/actions/workflows/tools.yml

There you can see it failed every time except: 1) in this PR; 2) in #8560 which added the GitHub workflow in the first place (see screenshot) below. 

I went and looked at every failure in that screenshot: they all failed with the same `comp_ops` failure above. I see now the `#if CONFIG_IPC_MAJOR_3 || CONFIG_LIBRARY` fix in this PR but it still does not explain why the build compiled in https://github.com/thesofproject/sof/actions/runs/7052569908 and "frequently" in my workspace. Very scary.


https://github.com/thesofproject/sof/actions/workflows/tools.yml

Screenshot 2023-12-04 at 15 36 58

marc-hb:
Mystery solved (I think): it was @lyakh's fa03f058586b6 which was merged at the SAME TIME as #8560 and _immediately_ regressed it. I really didn't expect that much of a coincidence!

> Yet sometimes it fails and sometimes it builds in my workspace too - on the same commit.

I must have not "cleaned" when switching commit or maybe I missed the error.

src/include/sof/audio/component.h
```diff
@@ -338,7 +338,7 @@ struct comp_ops {
 	int (*dai_config)(struct dai_data *dd, struct comp_dev *dev,
 			  struct ipc_config_dai *dai_config, const void *dai_spec_config);
 
-#if CONFIG_IPC_MAJOR_3
+#if CONFIG_IPC_MAJOR_3 || CONFIG_LIBRARY
```
marc-hb, line 341: @ranj063 could you submit this one-line fix in a separate PR so it can be fast-tracked?

kv2019i (approved):

PR #8539, opened 2023-11-27
coherent.h: assert(sizeof(struct coherent)  <= DCACHE_LINE_SIZE)

marc-hb:
Supersedes the "best effort" use of `packed`, being removed by #8528

### DIFF
--- a/src/include/sof/coherent.h
+++ b/src/include/sof/coherent.h
@@ -69,6 +69,14 @@ struct coherent {
 	struct list_item list;	/* coherent list iteration */
 } __coherent;
 
+#if CONFIG_INCOHERENT
+#  ifdef __ZEPHYR__
+BUILD_ASSERT(sizeof(struct coherent)  <= DCACHE_LINE_SIZE, "DCACHE_LINE_SIZE too small");
+#  else
+STATIC_ASSERT(sizeof(struct coherent) <= DCACHE_LINE_SIZE,  DCACHE_LINE_SIZE_too_small);
+#  endif
+#endif
+
 /* debug address aliases */
 #ifdef COHERENT_CHECK_ALIAS
 #define ADDR_IS_INCOHERENT(_c) assert(!is_uncached(_c))


marc-hb:
https://sof-ci.01.org/sofpr/PR8539/build499/devicetest/index.html all green

`10.91.113.127:443&TESTS=tests/avs/fw_02_base_fw_api/test_01_fw_large_config_get.py::TestFwLargeConfigGet::test_02_01_fw_config --platform*TGL` seems to have timed out in https://sof-ci.01.org/sof-pr-viewer/#/build/PR8539/build13183624, everything else green.

src/include/sof/coherent.h
```diff
@@ -69,6 +69,14 @@ struct coherent {
 	struct list_item list;	/* coherent list iteration */
 } __coherent;
 
+#if CONFIG_INCOHERENT
```
LaurentiuM1234, line 72: Do we need to document this configuration somewhere?

LaurentiuM1234 (approved):

marc-hb:
> so maybe we should rather aim at eliminating coherent.h completely rather than investing more work into it

Good to know but this just a one-line compile-time assert.

src/include/sof/coherent.h
marc-hb: The documentation for every `CONFIG_` can be found at:

```
git grep -A10 INCOHERENT '*/Kconfig'

src/platform/Kconfig:config INCOHERENT
src/platform/Kconfig-   bool "Enable cached data access via the Coherent API"
src/platform/Kconfig-   default y if CAVS || ACE
src/platform/Kconfig-   default n
src/platform/Kconfig-   help
src/platform/Kconfig-     The architecture is cache incoherent. i.e FW has to manually manage
src/platform/Kconfig-     cache coherency amongst objects that are used on multiple cores.
src/platform/Kconfig-     This setting should only be disabled for cache incoherent
src/platform/Kconfig-     architectures for testing without cached access to shared data.


```

marc-hb: Also at:

```
ninja -C build-mtl menuconfig
/
INCOHERENT
```

```

lgirdwood (approved):
TBH, coherent.h will go away as it should come from RTOS eventually.

kv2019i (approved):
Mandatory CI failing, otherwise good to go.

lgirdwood:
@lrudyX @wszypelt good to merge ? fwiw, this is a compile time check so expectation would be a pass if other builds are passing ? Thanks !

marc-hb:
> fwiw, this is a compile time check so expectation would be a pass if other builds are passing ? 

All builds did pass, the timeout was a run-time / test time.

PR #8542, opened 2023-11-28
module_adapter: Make free, prepare and reset methods optional

softwarecki:
Simple modules may not need to provide `free`, `prepare` and `reset` functions. To avoid having to provide dummy methods, added an if to check if a module implemented these methods. This will prevent null reference if a module doesn't implement these methods.

### DIFF
--- a/src/audio/up_down_mixer/up_down_mixer.c
+++ b/src/audio/up_down_mixer/up_down_mixer.c
@@ -398,27 +398,6 @@ err:
 	return ret;
 }
 
-/* just stubs for now. Remove these after making these ops optional in the module adapter */
-static int up_down_mixer_prepare(struct processing_module *mod,
-				 struct sof_source **sources, int num_of_sources,
-				 struct sof_sink **sinks, int num_of_sinks)
-{
-	struct up_down_mixer_data *cd = module_get_private_data(mod);
-	struct comp_dev *dev = mod->dev;
-
-	if (!cd->mix_routine) {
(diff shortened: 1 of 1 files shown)


ranj063 (approved):

btian1 (approved):

pjdobrowolski (approved):

marcinszkudlinski (changes requested):
see generic.c
int module_init(struct processing_module *mod, const struct module_interface *interface)

there's a check:

```
	/*check interface, there must be one and only one of processing procedure */
	if (!interface->init || !interface->prepare ||
	    !interface->reset || !interface->free ||
	    (!!interface->process + !!interface->process_audio_stream +
	     !!interface->process_raw_data != 1)) {
		comp_err(dev, "module_init(): comp %d is missing mandatory interfaces",
			 dev_comp_id(dev));
		return -EIO;
	}

```

so free and prepare are still obligatory after your change

src/audio/up_down_mixer/up_down_mixer.c
```diff
@@ -458,9 +437,7 @@ up_down_mixer_process(struct processing_module *mod,
 
 static const struct module_interface up_down_mixer_interface = {
 	.init = up_down_mixer_init,
-	.prepare = up_down_mixer_prepare,
```
btian1, line 461: can we add comments to interface.h reset/prepare/free and mark these 3 are optional?

tmleman (approved):

marcinszkudlinski (approved):

PR #8544, opened 2023-11-28
library_manager: Update module load flow

softwarecki:
The previous code have several issues:
- Did not support empty `data` segment.
- It assumed that certain types of segments would be at fixed indexes, without taking into account their flags/types.
- Incorrect handling of mapping error. If the virtual address cannot be mapped because is already in use, the error handler will unmap this address.
- If there is an error loading one of the modules marked as `lib_code`, previously loaded modules are not unloaded.

This commit fixes the above issues, but it probably introduces new problems :)

### DIFF
--- a/src/library_manager/lib_manager.c
+++ b/src/library_manager/lib_manager.c
@@ -51,11 +51,12 @@ static struct ext_library loader_ext_lib;
 
 #define PAGE_SZ		CONFIG_MM_DRV_PAGE_SIZE
 
-static int lib_manager_load_data_from_storage(void __sparse_cache *vma, void *s_addr,
-					      uint32_t size, uint32_t flags)
+static int lib_manager_load_data_from_storage(void __sparse_cache *vma, void *s_addr, uint32_t size,
+					      uint32_t flags)
 {
-	int ret = sys_mm_drv_map_region((__sparse_force void *)vma, POINTER_TO_UINT(NULL),
-					size, flags);
+	/* Region must be first mapped as writable in order to initialize its contents. */
+	int ret = sys_mm_drv_map_region((__sparse_force void *)vma, POINTER_TO_UINT(NULL), size,
+					flags | SYS_MM_MEM_PERM_RW);
 	if (ret < 0)
 		return ret;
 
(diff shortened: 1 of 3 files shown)


jxstelter (approved):

pjdobrowolski (approved):
Good move to split that long procedure into separate functions.

src/library_manager/lib_manager.c
```diff
@@ -54,7 +54,7 @@ static int lib_manager_load_data_from_storage(void __sparse_cache *vma, void *s_
 					      uint32_t size, uint32_t flags)
 {
 	int ret = sys_mm_drv_map_region((__sparse_force void *)vma, POINTER_TO_UINT(NULL),
-					size, flags);
+					size, SYS_MM_MEM_PERM_RW);
```
lyakh, line 57: looks wrong. You make an effort calculating flags below and here you overwrite them

src/library_manager/lib_manager.c
```diff
@@ -68,114 +68,133 @@ static int lib_manager_load_data_from_storage(void __sparse_cache *vma, void *s_
...
+	for (idx = 0; idx < ARRAY_SIZE(mod->segment); ++idx) {
+		if (!mod->segment[idx].flags.r.load)
+			continue;
+
+		flags = 0;
+
+		if (mod->segment[idx].flags.r.code)
+			flags = SYS_MM_MEM_PERM_EXEC;
+		else if (!mod->segment[idx].flags.r.readonly)
+			flags |= SYS_MM_MEM_PERM_RW;
```
lyakh, line 91: `|` superfluous

src/library_manager/lib_manager.c
```diff
@@ -68,114 +68,133 @@ static int lib_manager_load_data_from_storage(void __sparse_cache *vma, void *s_
...
-			if (module_entry->type.lib_code) {
-				ret = lib_manager_load_module(lib_id << LIB_MANAGER_LIB_ID_SHIFT |
-							      idx, mod, desc);
-				if (ret < 0)
-					goto err;
-			}
+/* There are modules marked as lib_code. This is code shared between several modules inside
+ * the library. Load all lib_code modules with first none lib_code module load.
+ */
+static int lib_manager_load_code_modules(const uint32_t module_id,
```
lyakh, line 140: My preference would be to leave "lib_code" support out until we have at least one module to test with. But if we do end up re-integrating this, can you please come up with better names - "code" doesn't tell me much at least

lyakh (changes requested):

src/library_manager/lib_manager.c
softwarecki: A region must be first mapped as writable in order to initialize its contents. Next, the expected access flags should be set (todo on line 67), but currently the tlb driver does not support this. In fact, the tlb driver currently completely ignores access flags and maps all regions as rwx.

src/library_manager/lib_manager.c
softwarecki: Yep. This needs fixing. Region code should be readonly.

src/library_manager/lib_manager.c
lgirdwood: does it make more send to use text, data, bss in the API naming ?

lyakh: @lgirdwood if I understood you correctly, this isn't about loading specific sections, it's about loading a complete module / library, but a library of a specific type - not one, implementing some module adapter driver for specific audio processing, but an auxiliary library to be used by user modules, like fft, math, etc.

keqiaozhang:
SOFCI TEST

src/library_manager/lib_manager.c
softwarecki: lyakh is right. I'll try to find a better name for these functions.

kv2019i (approved):
I'd have preferred separate commits as there are distinct separate issues you are addressing. Now the diff is really hard to follow so I instead just read the new code on its own and seems fine.

src/library_manager/lib_manager.c
```diff
@@ -68,114 +68,133 @@ static int lib_manager_load_data_from_storage(void __sparse_cache *vma, void *s_
...
-	size_t st_rodata_size = mod->segment[SOF_MAN_SEGMENT_RODATA].flags.r.length;
+	const struct ext_library *const ext_lib = ext_lib_get();
+	const uint32_t lib_id = LIB_MANAGER_GET_LIB_ID(module_id);
+	const uintptr_t load_offset = POINTER_TO_UINT(ext_lib->desc[lib_id]);
+	void *src, *va_base;
+	size_t size;
+	uint32_t flags;
+	int ret, idx;
+
+	for (idx = 0; idx < ARRAY_SIZE(mod->segment); ++idx) {
```
lyakh, line 82: on the one hand it's good to be flexible, to parse the input data safely, but OTOH aren't segment array index values a part of the ABI? Can `SOF_MAN_SEGMENT_*` values really be changed freely?

src/library_manager/lib_manager.c
```diff
@@ -68,114 +68,133 @@ static int lib_manager_load_data_from_storage(void __sparse_cache *vma, void *s_
...
+		if (!mod->segment[idx].flags.r.load)
+			continue;
```
lyakh, line 84: this skips .bss, so effectively you made a loop for 2 elements. Should we go a step further and integrate .bss here too by either copying data for .data and .text or clearing the buffer for .bss?

src/library_manager/lib_manager.c
```diff
@@ -68,114 +68,133 @@ static int lib_manager_load_data_from_storage(void __sparse_cache *vma, void *s_
...
 
 	return ret;
 }
 
-static int lib_manager_unload_module(uint32_t module_id, struct sof_man_module *mod,
-				     struct sof_man_fw_desc *desc)
+/* There are modules marked as lib_code. This is code shared between several modules inside
+ * the library. Unload all lib_code modules with last none lib_code module unload.
+ */
+static int lib_manager_unload_libcode_modules(const uint32_t module_id,
```
lyakh, line 178: let's leave this out until we have a test case

lgirdwood: I think CI will test this ? Btw , I dont see this as a blocker if missing from CI today.

src/library_manager/lib_manager.c
lgirdwood: ok, this is probably worth commenting in the code to avoid misunderstandings.

src/library_manager/lib_manager.c
lyakh: @lgirdwood CI will only test this if a test is created? This is a particular functionality that I'm not sure if we get a user in the next year or two or... And without a single user this will just make reviewing and maintaining more difficult and increase the code size needlessly.

lgirdwood: There will be Windows tests for this feature and others not targeting Linux..

lgirdwood: @softwarecki whats the ETA for client code that uses the extended/unused APIs today? I think we can quickly solve this with a KCONFIG that builds the unload logic when selected (and we can fully enable when client code is upstream). e.g. KCONFIG_LIB_MANAGER_UNLOAD.

lyakh: @lgirdwood a `CONFIG_...` option would help but not for unloading - this is just one function from the API in question: it's about both `lib_manager_load_libcode_modules()` and `lib_manager_unload_libcode_modules()`.

softwarecki:
> I'd have preferred separate commits as there are distinct separate issues you are addressing.

I identified the mentioned issues and simply write a new code from scrach. There were no incremental changes that I could separate.

src/library_manager/lib_manager.c
softwarecki: @lyakh The .bss section is handled differently. Its size is divided by the maximum number of module instances and only the selected fragment for a given instance is mapped.

src/library_manager/lib_manager.c
softwarecki: @lyakh  I checked the ext rom code and it looks like it uses static segment indexes.

src/library_manager/lib_manager.c
lyakh: well, that special .bss handling seems strange to me - it's too ad-hoc IMHO. This means, that you cannot have the same file used both as a loadable module and built-in, because its .bss will be handled differently. Besides it's too special - developers would need to think about it every time they put something in .bss. We discussed this with @pjdobrowolski briefly and maybe we can and should drop that behaviour.
And if we don't and we keep .bss out and only handle .text and .data here, we effectively get a loop over three elements to then throw away one of them, leaving a loop over two iterations - you could just keep it as is unfolded.

lgirdwood: @softwarecki @lyakh any update here, should we put a big inline comment here spelling out this difference and its expected behavior from ROM i.e. its done for current compatibility reasons

pjdobrowolski: any communication between modules, build-in or loadable should be done by using native-system-services. I'm no sure if making shared bss is good practice for such behavior. Especially if we load and unload modules on fly, what will happen with common bss? Each module will leave it own stuff in there until we will have an overflow/

lyakh: @pjdobrowolski what do you mean by "shared bss?" The "normal" programmer doesn't care about ELF sections. Those are just executable format details. They just put `static int x;` at the file level and they understand that there will be exactly one instance of that `x` on the entire system. The size of that `x` would usually be 4 bytes. If your module specifies maximum instance count of 8, you end up with half a byte of that `x` per instance. What is that good for? If the programmer wants to make per-instance data, they make `static int x[max_instance_count]` and then they count instances in their instantiation routine and use the respective member of the array. How are you going to use a "part of .bss" in your software?

pjdobrowolski: DSP has restricted 3MB of ram which is shared between modules. BSS is defining how much of we use. Can you explain me how removing BSS sections from elf will improve working on such small space?

jxstelter: @lyakh You wrote: "This means, that you cannot have the same file used both as a loadable module and built-in...". Where you see use case to have the same code used as loadable and built-in at the same time? I think we do loadable to not have given module built into base FW all the time. 
Another think, there are other reasons for having separate .bss section for each module. 
One is backward compatibility with reference FW design.
The other even more important is possibility to use in the feature some memory protection features. Mixing up the module .bss with base FW general .bss section will block us from introducing such solution. 
The third reason is the debuggability of base FW with presence of external modules. If we mix up the memory spaces, tracking any failures will be more difficult since external modules could corrupt base FW data.

lyakh: @pjdobrowolski @jxstelter I think there's some confusion here. I'm not proposing to remove .bss or to share .bss *between modules* or anything special of that kind. I just propose to do what C does: you program
```
static int module_x;
```
and it's placed in .bss and it's available there to that *file* (because it's `static`). Or you have a global
```
int global_x;
```
and it's also in .bss and accessible globally. In all cases there's exactly 1 copy of that object. What I don't understand why loadable modules are linking those objects with module *instances*. That means, that e.g. if you make volume a loadable module. And you activate 3 pipelines and each of them has a volume component in it so you have 3 instances. So then you give to each *instance* i.e. to each volume component in each pipeline a separate *part* of that .bss, i.e. a part of those objects. I understand *why* it is done, but I don't find this a good idea. It is done to "emulate" a poor-man allocation, so you want to put something like
```
static struct module_private_data[MAX_MODULE_INSTANCES];
```
in those modules and then give each instance one of those objects. But I don't find that a good approach. I think it's confusing and misinterpreting the meaning of the ELF .bss section

pjdobrowolski: If you would look more carefully in history of commits, you would find out, that connecting .bss with instances was introduce much before idea of "poor-man allocation".

lgirdwood: ok, so I think we probably need to make relationship between bss and instances more visible with inline comments in the code here, this way we can avoid any misunderstandings as we continue and add logic for services API and IADK modules.

lgirdwood:
@softwarecki fyi some CI failures, there has been a recent west update so you may want to rebase and rerun the CI.

kv2019i:
@softwarecki  This is ready to go, but we'd need a green run for CI, can you check?

lgirdwood:
> @softwarecki This is ready to go, but we'd need a green run for CI, can you check?

@kv2019i sparse errors ? Do we need a west update to fix them ?

kv2019i:
@lgirdwood wrote:
> > @softwarecki This is ready to go, but we'd need a green run for CI, can you check?
> 
> @kv2019i sparse errors ? Do we need a west update to fix them ?

"Internal Intel CI System/merge/build — Tests failed " -> https://sof-ci.01.org/sof-pr-viewer/#/build/PR8544/build13245909

softwarecki:
@lgirdwood i think its ready

lgirdwood:
Zephr main failures are known and unrelated.

PR #8546, opened 2023-11-28
lmdk: dwmix: Add example of loadable down mixer module

softwarecki:
This PR adds the example of down mixer module ported from iadk. Its build as a native loadable sof module.

Ported additional error code definitions from iadk module API. These codes are not unique and their meaning depends on the context / which function returns it.

### DIFF
--- a/lmdk/modules/down_mixer/downmixer.c
+++ b/lmdk/modules/down_mixer/downmixer.c
@@ -1,0 +1,520 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/*
+ * Copyright(c) 2023 Intel Corporation. All rights reserved.
+ * Author: Adrian Warecki <adrian.warecki@intel.com>
+ */
+
+#include <string.h>
+#include <stdint.h>
+#include <stddef.h>
+
+#include <module/base.h>
+#include <module/api_ver.h>
+#include <iadk/adsp_error_code.h>
+#include <rimage/sof/user/manifest.h>
+#include <audio/source_api.h>
+#include <audio/sink_api.h>
+
+#include "downmixer.h"
+
+/* Logging is temporary disabled */
+#define LOG_MESSAGE(...) 
+
+static const struct native_system_agent *native_sys_agent;
+static struct module_self_data self_data;
+
+/**
+ * Module specific initialization procedure.
+ */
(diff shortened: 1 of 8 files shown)


serhiy-katsyuba-intel (approved):

lmdk/modules/down_mixer/downmixer.c
```diff
@@ -0,0 +1,520 @@
...
+	divider_input_0 = self->config_.divider_input_0;
+	divider_input_1 = self->config_.divider_input_1;
+
+	if (self->processing_mode_ == MODULE_PROCESSING_BYPASS) {
+		divider_input_0 = self->input0_channels_count_;
+		/* input1_channels is already set to 0 in BYPASS mode */
+	}
+
+	frames_processed = 0;
+	while (frames_processed < frames_to_process) {
```
pjdobrowolski, line 331: I'm struggling with that frames processing. For an outsider it is a bit complex and without knowing how pipelines work looks like magic. @marcinszkudlinski do we have any docs for that?

src/include/module/iadk/adsp_error_code.h
```diff
@@ -36,4 +36,27 @@ typedef uint32_t AdspErrorCode;
 /* Service is not supported on target platform. */
 #define ADSP_SERVICE_UNAVAILABLE 143
 
+#define ADSP_MAX_VALUE ADSP_FATAL_FAILURE
+
+/* SystemAgentInterface */
+#define ADSP_MODULE_CREATION_FAILURE (ADSP_MAX_VALUE + 1)
```
lgirdwood, line 42: Wont the max value be changed here if we add more errors ?

lmdk/modules/down_mixer/downmixer.c
```diff
@@ -0,0 +1,520 @@
+// SPDX-License-Identifier: BSD-3-Clause
```
lgirdwood, line 1: Lets not duplicate the module sources in the code base, we should use the same module code regardless of basefw or module/library build.

lgirdwood (changes requested):
We need to use the same module code regardless of built-in baseFW or library module.

softwarecki:
Closing PR as we need to use the same module code for built-in and loadable module.

PR #8548, opened 2023-11-29
zephyr/cmake: move math/*.c files to now common math/CMakeLists.txt

marc-hb:
CMake decentralization per #8260

EDIT: There seems to be some confusion happening in the review comments right now.

Clarification: this CMake PR is BUG-FOR-BUG compatible.

There are Cmake and CONFIG bugs in zephyr/CMakeLists.txt right now that this https://github.com/thesofproject/sof/issues/8260 effort is exposing. The purpose of this PR is to carefully NOT fix any of these bugs. The commits in this PR and all other 8260 commits in general are way too big to hide small functional changes inside. You really don't want your git bisect to land on a giant CONFIG_ commit like these. So this PR is strictly "bug-preserving". Garbage in, exact same garbage out.

If this PR made you discover some existing CMake or CONFIG bug, then please submit a smaller PR either before this PR is merged (and I will rebase and adjust), or after this PR. But please stop asking for bug fixes here.

### DIFF
--- a/src/math/CMakeLists.txt
+++ b/src/math/CMakeLists.txt
@@ -5,9 +5,13 @@ if(BUILD_LIBRARY)
 	return()
 endif()
 
+is_zephyr(it_is_zephyr)
+
 add_local_sources(sof numbers.c)
 
-if(CONFIG_CORDIC_FIXED)
+# Up to now, trig.c has never been optional in Zephyr.
+# Maybe it should be in the future.
+if(CONFIG_CORDIC_FIXED OR it_is_zephyr)
         add_local_sources(sof trig.c)
 endif()
 
@@ -15,10 +19,13 @@ add_local_sources_ifdef(CONFIG_SQRT_FIXED sof sqrt_int16.c)
 
 add_local_sources_ifdef(CONFIG_MATH_EXP sof exp_fcn.c exp_fcn_hifi.c)
 
-if(CONFIG_MATH_DECIBELS)
+# Up to now, decibels.c has never been optional in Zephyr.
+# Maybe it should be in the future.
+if(CONFIG_MATH_DECIBELS OR it_is_zephyr)
         add_local_sources(sof decibels.c)
 endif()
 
+if(NOT it_is_zephyr) # So far none of these has ever been enabled in Zephyr.
 add_local_sources_ifdef(CONFIG_NATURAL_LOGARITHM_FIXED sof log_e.c)
 
 add_local_sources_ifdef(CONFIG_COMMON_LOGARITHM_FIXED sof log_10.c)
@@ -26,10 +33,12 @@ add_local_sources_ifdef(CONFIG_COMMON_LOGARITHM_FIXED sof log_10.c)
 add_local_sources_ifdef(CONFIG_POWER_FIXED sof power.c)
 
 add_local_sources_ifdef(CONFIG_BINARY_LOGARITHM_FIXED sof base2log.c)
+endif()
 
 add_local_sources_ifdef(CONFIG_MATH_FIR sof fir_generic.c fir_hifi2ep.c fir_hifi3.c)
 
-if(CONFIG_MATH_FFT)
+# So far this directory has never been enabled in Zephyr.
+if(CONFIG_MATH_FFT AND NOT it_is_zephyr)
 	add_subdirectory(fft)
 endif()
 
@@ -39,6 +48,7 @@ add_local_sources_ifdef(CONFIG_MATH_IIR_DF2T sof
 add_local_sources_ifdef(CONFIG_MATH_IIR_DF1 sof
 	iir_df1_generic.c iir_df1_hifi3.c iir_df1.c)
 
+if(NOT it_is_zephyr) # So far none of these has ever been enabled in Zephyr.
 if(CONFIG_MATH_WINDOW)
 	 add_local_sources(sof window.c)
 endif()
@@ -54,3 +64,4 @@ endif()
 if(CONFIG_MATH_DCT)
 	 add_local_sources(sof dct.c)
 endif()
+endif() # not Zephyr
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -56,7 +56,6 @@ set(SOF_SAMPLES_PATH "${SOF_SRC_PATH}/samples")
 set(SOF_LIB_PATH "${SOF_SRC_PATH}/lib")
 set(SOF_DRIVERS_PATH "${SOF_SRC_PATH}/drivers")
 set(SOF_DEBUG_PATH "${SOF_SRC_PATH}/debug")
-set(SOF_MATH_PATH "${SOF_SRC_PATH}/math")
 set(SOF_TRACE_PATH "${SOF_SRC_PATH}/trace")
 
 set(RIMAGE_TOP ${sof_top_dir}/tools/rimage)
@@ -146,6 +145,7 @@ endmacro()
 
 add_subdirectory(../src/init/ init_unused_install/)
 add_subdirectory(../src/ipc/  ipc_unused_install/)
+add_subdirectory(../src/math/ math_unused_install/)
 
 
 
@@ -366,11 +366,6 @@ zephyr_include_directories(${SOF_PLATFORM_PATH}/${PLATFORM}/include)
 # Commented files will be added/removed as integration dictates.
 zephyr_library_sources(
 
-	# SOF math utilities
-	${SOF_MATH_PATH}/decibels.c
-	${SOF_MATH_PATH}/numbers.c
-	${SOF_MATH_PATH}/trig.c
-
 	# SOF library - parts to transition to Zephyr over time
 	${SOF_LIB_PATH}/clk.c
 	${SOF_LIB_PATH}/notifier.c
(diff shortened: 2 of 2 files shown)


keqiaozhang:
SOFCI TEST

marc-hb:
Only 1 suspend/resume timeout in https://sof-ci.01.org/sofpr/PR8548/build585/devicetest/index.html

https://sof-ci.01.org/sof-pr-viewer/#/build/PR8548/build13196824 all green

New i915 GSC firmware needed in https://sof-ci.01.org/sofpr/PR8548/build584/devicetest/index.html, everything else green.

src/math/CMakeLists.txt
```diff
@@ -5,31 +5,41 @@ if(BUILD_LIBRARY)
 	return()
 endif()
 
+is_zephyr(it_is_zephyr)
+
 add_local_sources(sof numbers.c)
 
-if(CONFIG_CORDIC_FIXED)
+if(CONFIG_CORDIC_FIXED OR it_is_zephyr)
```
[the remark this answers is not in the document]
marc-hb: `is_zephyr` is only a check for CONFIG_ZEPHYR_SOF_MODULE. The indirection is for brevity, flexibility and "future-proofness".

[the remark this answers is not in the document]
lgirdwood: why not just use ```CONFIG_ZEPHYR_SOF_MODULE``` if they are equivalent.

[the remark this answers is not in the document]
marc-hb: They are equivalent now but who knows in the future. I'm currently spreading this conditional across almost all `sof/**/CMakeLists.txt` files, so this very thin indirection protects us against any future change. If this CONFIG_ namve ever changes in Zephyr in the future we could even make the macro support both the old and new for a seamless transition.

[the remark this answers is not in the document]
lgirdwood: If they are equivalent now then pls use, otherwise this is really just unneeded indirection today.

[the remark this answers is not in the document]
marc-hb: It's a bit late now, this `is_zephyr()` macro was reviewed, approved and added in #8422 and is already in use in `src/init/`, `src/ipc` and `lmdk/cmake`. So not using it would be inconsistent. 

Speaking of consistency, another aspect of the problem is that there are at least two macros: `CONFIG_SOF` and `CONFIG_ZEPHYR_SOF_MODULE`. Which one is "best"? Maybe there's even a another and better one, who knows. I looked in the documentation and in the source a bit but no one really knows for sure; it's not a common problem to solve. Hence the indirection which can be fixed with a one-line change instead of a mass search/replace. Again this was discussed in #8422 cc: @andyross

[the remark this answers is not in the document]
andyross: Tend to agree that the existing kconfigs aren't super well named for this purpose.  They're artifacts of the way the upstream Zephyr module system works: ZEPHYR_SOF_MODULE is automatically generated, and 'config SOF' is added in the Kconfig for the module.  FWIW I NAK'd the latter because in this context it means the opposite of what it says ("CONFIG_SOF" might be expected to mean "xtos/traditional SOF and not Zephyr", when it means the opposite).

My personal vote would be a SOF-internal API that contains "zephyr" in the name like this one.  Whether it's a kconfig or a cmake macro doesn't seem all that important I guess?  Would be easy to fix in a different patch too.

kv2019i (approved):

src/math/CMakeLists.txt
ShriramShastry, line 12: Using the same name for managing math functions under Zephyr is not a good idea.

src/math/CMakeLists.txt
```diff
@@ -5,31 +5,41 @@ if(BUILD_LIBRARY)
...
         add_local_sources(sof trig.c)
 endif()
 
 add_local_sources_ifdef(CONFIG_SQRT_FIXED sof sqrt_int16.c)
 
 add_local_sources_ifdef(CONFIG_MATH_EXP sof exp_fcn.c exp_fcn_hifi.c)
 
-if(CONFIG_MATH_DECIBELS)
+if(CONFIG_MATH_DECIBELS OR it_is_zephyr)
```
ShriramShastry, line 20: Using the same name for managing math functions under Zephyr is not a good idea.

src/math/CMakeLists.txt
```diff
@@ -5,31 +5,41 @@ if(BUILD_LIBRARY)
...
 add_local_sources_ifdef(CONFIG_NATURAL_LOGARITHM_FIXED sof log_e.c)
 
 add_local_sources_ifdef(CONFIG_COMMON_LOGARITHM_FIXED sof log_10.c)
 
 add_local_sources_ifdef(CONFIG_POWER_FIXED sof power.c)
 
 add_local_sources_ifdef(CONFIG_BINARY_LOGARITHM_FIXED sof base2log.c)
+endif()
 
+if(it_is_zephyr)
```
ShriramShastry, line 35: Same here!! Using the same name for managing math functions under Zephyr is not a good idea.

ShriramShastry (commented):
I've added my review comments. I welcome you making changes.

src/math/CMakeLists.txt
marc-hb: I'm not sure I understand @ShriramShastry. The purpose of this PR is to be very careful and NOT make any config "management" change.  That can come later after this clean up. Currently, math function are NOT managed in Zephyr.

ShriramShastry: OK, What I mean is that `it_is_zephyr` is used to enable and disable math functions. If the function is called in the code, you enable `it_is_zephyr`, otherwise you turn it off. What I'm not persuaded of is the use of the same flag (`it_is_zephyr`) across Zephyr code . If you have a solid justification (`That's Zephyr`) for using the same name, that's fine; otherwise, I advocate using a distinct naming policy for each math function.

marc-hb: >  that it_is_zephyr is used to enable and disable math functions

No, that's not what it is. No one is going to "turn Zephyr on/off" to enable/disable some math functions. No one can decide whether they use Zephyr or not, it's not a choice.

Before this PR:
- CORDIC is not optional in Zephyr

After this PR:
- CORDIC is not optional in Zephyr

The purpose of this PR is to carefully NOT change that.

If you want make to make CORDIC or other math code CONFIGurable in Zephyr then please submit a separate PR. This PR is a pure clean-up, it does not make any CONFIG change.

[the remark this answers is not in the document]
marc-hb: `CONFIG_CORDIC OR CONFIG_ZEPHYR_...` would also give the wrong impression that both are choices, see discussion below. The second one is obviously not a CONFIGuration choice.

src/math/CMakeLists.txt
```diff
@@ -5,31 +5,41 @@ if(BUILD_LIBRARY)
...
+add_local_sources_ifdef(CONFIG_COMP_FIR sof fir_generic.c fir_hifi2ep.c fir_hifi3.c)
```
andrula-song, line 36: why not CONFIG_MATH_FIR?  TDFB will select  MATH_FIR too.

src/math/CMakeLists.txt
```diff
@@ -5,31 +5,41 @@ if(BUILD_LIBRARY)
...
+else()
 add_local_sources_ifdef(CONFIG_MATH_FIR sof fir_generic.c fir_hifi2ep.c fir_hifi3.c)
+endif()
 
-if(CONFIG_MATH_FFT)
+# Never used with Zephyr.
+if(CONFIG_MATH_FFT AND NOT it_is_zephyr)
```
andrula-song, line 42: actually this would be called in MFCC, as well as window and dct, though MFCC is not included in zephyr now, but can we keep as original?

src/math/CMakeLists.txt
marc-hb: Because it was `CONFIG_MATH_FIR` in Zephyr, please pay attention to the previous code.

Please read the previous discussion: the purpose of this PR is to carefully NOT change the CONFIGuration.

If you want to fix the configuration then please submit a separate PR and I will adjust this one.

src/math/CMakeLists.txt
marc-hb: You lost me sorry: what is different from the original? I tried really hard not to change anything.

src/math/CMakeLists.txt
andrula-song: I think it is a mistake in zephyr/Cmakelists.txt, the fir lib files should be like 
```
zephyr_library_sources_ifdef(CONFIG_MATH_FIR
	${SOF_MATH_PATH}/fir_generic.c
	${SOF_MATH_PATH}/fir_hifi2ep.c
	${SOF_MATH_PATH}/fir_hifi3.c
)
```

src/math/CMakeLists.txt
andrula-song: okay, since the dependency of MFCC libs haven't been included in zephyr yet, it is okay to modify code latter if we want to add mfcc to zephyr.

src/math/CMakeLists.txt
andrula-song: here is the fix https://github.com/thesofproject/sof/pull/8581

marc-hb:
Demoting to draft until @andrula-song 's bug fix is reviewed:
- #8581

src/math/CMakeLists.txt
lyakh: my question of the day to these changes: I understand that it's just carried over from the present global sof/zephyr/CMakeLists.txt, but now that we move to these local files, shouldn't we rather just `select MATH_DECIBELS` somewhere with Zephyr?

ShriramShastry: Yes, I am thinking about it, and it makes more sense to me to have a select option with Zephyr.

marc-hb:
Rebased on top of #8551. More comments added, thanks for the feedback!

src/math/CMakeLists.txt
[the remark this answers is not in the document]
marc-hb: > For Zephyr stuff, we need something CONFIG_SOF_ZEPHYR_X the precedence being this is a SOF Kconfig and NOT a Zephyr Kconfig.
> @marc-hb can you incrementally change the is_zephyr to use the Kconfig, this way is easier for everyone to follow.

You lost me sorry.

marc-hb:
I volunteered to do this on my "spare time" but this is proving too time consuming, it needs proper planning. Maybe another time?

src/math/CMakeLists.txt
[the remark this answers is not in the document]
marc-hb: I just found that some places started using `CONFIG_ZEPHYR_NATIVE_DRIVERS`...

PR #8550, opened 2023-11-30
[cavs2.5-001] topology1: sof-hda-generic: enable dts playback

macchian:
Enabling dts pipeline on HDA0 speaker/headphone playaback.
In this case, using defined macro W_PIPELINE_TOP() to add missing dai pipeline is due to topology not found it for PCM0P.
Verified dts test blobs via sof-ctl.

### DIFF
--- a/tools/topology/topology1/CMakeLists.txt
+++ b/tools/topology/topology1/CMakeLists.txt
@@ -27,6 +27,7 @@ set(TPLGS
 	"sof-hda-generic\;sof-hda-generic-2ch\;-DCHANNELS=2\;-DHSPROC=volume\;-DDMICPROC_FILTER1=eq_iir_coef_highpass_40hz_20db_48khz.m4\;-DDMIC16KPROC_FILTER1=eq_iir_coef_highpass_40hz_20db_16khz.m4\;-DDYNAMIC=1"
 	"sof-hda-generic\;sof-hda-generic-3ch\;-DCHANNELS=4\;-DHSPROC=volume\;-DDMICPROC_FILTER1=eq_iir_coef_highpass_40hz_20db_48khz.m4\;-DDMIC16KPROC_FILTER1=eq_iir_coef_highpass_40hz_20db_16khz.m4\;-DDYNAMIC=1"
 	"sof-hda-generic\;sof-hda-generic-4ch\;-DCHANNELS=4\;-DHSPROC=volume\;-DDMICPROC_FILTER1=eq_iir_coef_highpass_40hz_20db_48khz.m4\;-DDMIC16KPROC_FILTER1=eq_iir_coef_highpass_40hz_20db_16khz.m4\;-DDYNAMIC=1"
+	"sof-hda-generic\;sof-hda-generic-4ch-dts\;-DCHANNELS=4\;-DHSPROC=volume\;-DDMICPROC_FILTER1=eq_iir_coef_highpass_40hz_20db_48khz.m4\;-DDMIC16KPROC_FILTER1=eq_iir_coef_highpass_40hz_20db_16khz.m4\;-DDYNAMIC=1\;-DDTS=`DTS'"
 	## end HDaudio codec topologies
 
 	"sof-hda-generic-idisp\;sof-hda-generic-idisp\;-DCHANNELS=0\;-DDYNAMIC=1"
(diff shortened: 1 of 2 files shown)


tools/topology/topology1/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ set(TPLGS
 	"sof-hda-generic\;sof-hda-generic-2ch\;-DCHANNELS=2\;-DHSPROC=volume\;-DDMICPROC_FILTER1=eq_iir_coef_highpass_40hz_20db_48khz.m4\;-DDMIC16KPROC_FILTER1=eq_iir_coef_highpass_40hz_20db_16khz.m4\;-DDYNAMIC=1"
 	"sof-hda-generic\;sof-hda-generic-3ch\;-DCHANNELS=4\;-DHSPROC=volume\;-DDMICPROC_FILTER1=eq_iir_coef_highpass_40hz_20db_48khz.m4\;-DDMIC16KPROC_FILTER1=eq_iir_coef_highpass_40hz_20db_16khz.m4\;-DDYNAMIC=1"
 	"sof-hda-generic\;sof-hda-generic-4ch\;-DCHANNELS=4\;-DHSPROC=volume\;-DDMICPROC_FILTER1=eq_iir_coef_highpass_40hz_20db_48khz.m4\;-DDMIC16KPROC_FILTER1=eq_iir_coef_highpass_40hz_20db_16khz.m4\;-DDYNAMIC=1"
+	"sof-hda-generic\;sof-hda-generic-4ch-dts\;-DCHANNELS=4\;-DHSPROC=volume\;-DDMICPROC_FILTER1=eq_iir_coef_highpass_40hz_20db_48khz.m4\;-DDMIC16KPROC_FILTER1=eq_iir_coef_highpass_40hz_20db_16khz.m4\;-DDYNAMIC=1\;-DDTS=`DTS'"
```
plbossart, line 30: I have mixed feelings about this, all existing HDaudio topologies don't have any processing, so new topologies with processing should be handled in a well-identified block. @singalsu was also working on changes to HDaudio, we should place all 'enhanced' topologies together IMHO.

At a minimum this should be a different comment block, worst-case a different file.

plbossart: @macchian the point is that DTS is not an open-sourced library that is distributed as part of sof-bin, so we want this topology to be identified and not released by sof-bin (or released in a very clear directory that says CHROME_ONLY or something).

Adding @marc-hb for further comments.

lgirdwood: This is needed for Chrome cavs-2.5 and main (tplg2) branches but not needed in stable-v2.2.

lgirdwood (approved):
@mwasko fyi - HDA processing topology for CAVS-2.5.

tools/topology/topology1/CMakeLists.txt
lgirdwood: @macchian do you have a similar PR with enhanced HDA for main branch ? We cant use 3P for HDA main, but we can use EQ and other open modules. Please work with @singalsu when he is back next week on module/pipeline definition (that can be used in CI) around HDA. Thanks !

marc-hb: > > the point is that DTS is not an open-sourced library that is distributed as part of sof-bin, so we want this topology to be identified and not released by sof-bin
> > Adding @marc-hb for further comments.

> This is needed for Chrome cavs-2.5 and main (tplg2) branches but not needed in stable-v2.2.

It is also my understanding that this cavs25 branch will never be released to sof-bin.

Now we should probably _still_ have a good answer to @plbossart for the longer term (new subdirectory?) but not needed in the cavs2.5 branch AFAIK.

Even if it were accidentally released to sof-bin, how much damage would that do?

lgirdwood: cavs-2.5 branch is downstream Chrome so not in sof-bin. 
We can make sure a similar PR for main uses the EQ (or similar open module) as an example module.

marc-hb: > Otherwise, there's a risk of someone loading a topology that will result in an IPC error when trying to access a non-existent module.

Is a kernel parameter the only way to attempt this?

lgirdwood:
Adding @abonislawski since @mwasko is out.

PR #8552, opened 2023-11-30
Audio: up_down_mixer: move up_down_mixer header files to module folder

btian1:
This is a clean up, purpose is de-cluster headers, toml files,
Readme.md etc per module basis, since today everything is scattered
in current code base.

### DIFF
--- a/src/audio/up_down_mixer/up_down_mixer.c
+++ b/src/audio/up_down_mixer/up_down_mixer.c
@@ -5,8 +5,6 @@
 // Author: Bartosz Kokoszko <bartoszx.kokoszko@intel.com>
 // Author: Adrian Bonislawski <adrian.bonislawski@intel.com>
 
-#include <sof/audio/coefficients/up_down_mixer/up_down_mixer.h>
-#include <sof/audio/up_down_mixer/up_down_mixer.h>
 #include <sof/audio/buffer.h>
 #include <sof/audio/format.h>
 #include <sof/audio/module_adapter/module/generic.h>
@@ -27,6 +25,9 @@
 #include <errno.h>
 #include <stddef.h>
 #include <stdint.h>
+
+#include "up_down_mixer_coef.h"
+#include "up_down_mixer.h"
 
 LOG_MODULE_REGISTER(up_down_mixer, CONFIG_SOF_LOG_LEVEL);
 
--- a/src/audio/up_down_mixer/up_down_mixer.h
+++ b/src/audio/up_down_mixer/up_down_mixer.h
@@ -12,11 +12,12 @@
 #include <sof/audio/ipc-config.h>
 #include <sof/common.h>
 #include <ipc/stream.h>
-#include <ipc4/up_down_mixer.h>
 #include <ipc4/module.h>
 #include <ipc4/base-config.h>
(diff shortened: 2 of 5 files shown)


RanderWang (commented):
Can you explain why we need to do it ? Thanks!

src/audio/up_down_mixer/coef/up_down_mixer.h
```diff
@@ -5,7 +5,7 @@
 // Author: Bartosz Kokoszko <bartoszx.kokoszko@intel.com>
 // Author: Adrian Bonislawski <adrian.bonislawski@intel.com>
 
-#include <ipc4/up_down_mixer.h>
+#include "../up_down_mixer_ipc4.h"
```
lyakh, line 8: can we just name this file "coef.h" and place it at the same level?

btian1: this is not coef header, coef header put it to coef/up_down_mixer.h
this header have definition:
ipc4_up_down_mixer_module_cfg

btian1:
> Can you explain why we need to do it ? Thanks!

purpose is all module content in the same folder: header/source/kconfig/CMakeList, etc, this is part of our Q4 OKR.

src/audio/up_down_mixer/coef/up_down_mixer.h
lyakh: yes, I mean the coefficient header, i.e. the file, which is being patched here, not the one, included by this instruction, sorry for the confusion

btian1: Do you mean change coef/up_down_mixer.h --> up_down_mixer_coef.h ?

btian1: it is not coef.h, it is in coef folder with name up_down_mixer.h. 
downmix only have one header file, however, for src/asrc, we have hundred of header files, change file name is huge cost.
how about keep it with coef/up_down_mixer.h?

lyakh: > it is not coef.h, it is in coef folder with name up_down_mixer.h. downmix only have one header file, however, for src/asrc, we have hundred of header files, change file name is huge cost. how about keep it with coef/up_down_mixer.h?

@btian1 I know that your current version calls it "coef/up_down_mixer.h" and that's what I'm proposing to change for two reasons: (1) two files with the same name in adjacent directories look confusing, (2) no need to create an additional directory level just for one file. So, originally I proposed to just do
```
git mv coef/up_down_mixer.h coef.h
```
to which you proposed to call it up_down_mixer_coef.h instead which is also fine with me, but not using the same name as in your current version.

btian1: ok, use up_down_mixer_coef.h is better since coef.h is too common for file name.

lgirdwood (approved):

kv2019i:
@btian1 Build errors in CI, please check, e.g https://github.com/thesofproject/sof/actions/runs/7086932318/job/19286139196?pr=8552

lyakh (approved):

PR #8554, opened 2023-11-30
Add Loadable smart amp support

RanderWang:


lmdk/include/adsp_stddef.h
```diff
@@ -0,0 +1,108 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// Copyright(c) 2021 Intel Corporation. All rights reserved.
+
+
+#ifndef _ADSP_STDDEF_H_
+#define _ADSP_STDDEF_H_
```
kv2019i, line 7: @RanderWang Why do we need to copy headers? It seems in this example copying headers is not needed -> https://github.com/thesofproject/sof/pull/8546

I thought now that we have headers in separate module folder, we no longer needed this. And 8546 seems to show this is now needed.

Something I'm missing?

softwarecki (changes requested):
I would mention that all these modifications in `module_adapter` are unnecessary. Native loadable modules should not use the `module_adapter`. @pjdobrowolski is working on the necessary fix in `lib_manager`, until then I'm using this workaround https://github.com/softwarecki/sof/commit/26f4b475d30c81477ad8b4eadf29abc48ab1e31e

lmdk/include/adsp_stddef.h
RanderWang: o, I did part time work on loadable smart amp long time ago and am out of date. Let me check the latest status

RanderWang:
The smart amp is still using input output stream buffer so we have such issue. Now plan to first convert stream buffer to sink|source buffer.

lyakh:
> I would mention that all these modifications in `module_adapter` are unnecessary. Native loadable modules should not use the `module_adapter`. @pjdobrowolski is working on the necessary fix in `lib_manager`, until then I'm using this workaround [softwarecki@26f4b47](https://github.com/softwarecki/sof/commit/26f4b475d30c81477ad8b4eadf29abc48ab1e31e)

@softwarecki you probably mean "shouldn't use modules.c," not use the module adapter, because the module adapter is an essential API that cannot be dropped. As for the modules.c wrapping - removing it would be good, yes. Looking forward to it!

PR #8555, opened 2023-11-30
[cavs2.5-001] waves: store config blob in a cache in waves.c

barry-waves:
Store/apply config blob in a cache to avoid that
cfg.data will be released after prepare.


johnylin76 (approved):
@mwasko commit cherry-picked from main https://github.com/thesofproject/sof/pull/8448

PR #8558, opened 2023-11-30
copier: store gateway config in copier_data

plbossart:
Calculate actual size of copier module gateway config. Copy and store it in copier_data. It needs to be available even after copier is created e.g. during SET_PIPELINE_STATE IPC when dai_config_dma_channel() is called second time and DMA config is used to assign dma_channel_id value.

### DIFF
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -64,6 +64,8 @@ static int copier_init(struct processing_module *mod)
 	struct module_data *md = &mod->priv;
 	struct ipc4_copier_module_cfg *copier = (struct ipc4_copier_module_cfg *)md->cfg.init_data;
 	struct comp_ipc_config *config = &dev->ipc_config;
+	void *gtw_cfg = NULL;
+	size_t gtw_cfg_size;
 	int i, ret = 0;
 
 	cd = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*cd));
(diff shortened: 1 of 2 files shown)


fredoh9:
I tested multiple times but still seeing `FW reported error: 7 - Unsupported operation requested`
Doesn't seem improve any thing.

RanderWang:
@plbossart  please check 4cc9d5418b0e9389a8ebdaf81eda13ea2c1d24e2. At first we save the gateway config, but it was deleted
````
+       /*
+        * Don't copy the config_data[] variable size array, we don't need to
+        * store it, it's only used during IPC processing, besides we haven't
+        * allocated space for it, so don't "fix" this!
+        */
````

File: src/audio/copier/copier.c [hunk 1]
```diff
@@ -78,7 +80,31 @@ static int copier_init(struct processing_module *mod)
...
+
+	/* Allocate memory and store gateway_cfg in runtime. Gateway cfg has to
+	 * be kept even after copier is created e.g. during SET_PIPELINE_STATE
+	 * IPC when dai_config_dma_channel() is called second time and DMA
+	 * config is used to assign dma_channel_id value.
+	 */
+	if (copier->gtw_cfg.config_length) {
+		gtw_cfg_size = copier->gtw_cfg.config_length << 2;
+		gtw_cfg = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
+				  gtw_cfg_size);
```
lyakh, line 94: please use `rmalloc()` - you overwrite it completely 5 lines later

lyakh (changes requested):

iganakov:
@plbossart @fredoh9 Let me submit a proper fix for DMIC and SSP, we've already tested it internally. SNDW needs additional changes related to blob format which is now different for Linux and Windows

PR #8534, opened 2023-11-27
cache handling in volume - address #6285

lyakh:
clarify IPC-version specific processing paths and remove a redundant cache invalidation. This should resolve #6285

PR #8560, opened 2023-11-30
[SKIP SOF-TEST] .github/tools.tml: add build of SOF ALSA plugin

marc-hb:
2 commits


kv2019i (approved):
Thanks @marc-hb , this is great! This is a very different type of build, so without a CI job, this is bound to get broken.

ranj063 (approved):

marc-hb:
> Thanks @marc-hb , this is great! This is a very different type of build, so without a CI job, this is bound to get broken.

And to prove yourself right you merged @lyakh's  #8534 at the same time which _did_ immediately break this build :-D

Longer story and longer confusion (sorry) in #8538

cc: @ranj063

marc-hb:
Oh and now I just saw that you did guess that right and shared it!
- https://github.com/thesofproject/sof/pull/8518#issuecomment-1838089696

Way too many notifications from Github, impossible to cope. I especially tend to ignore the ones with a "topology" title.

PR #8561, opened 2023-12-01
base_fw: add DMI_FORCE_L1_EXIT FW config (depends on Zephyr 66042)

fkwasowi:
Add new parameter for SW to force DMI L1 exit on IPC request.
Associated with: https://github.com/zephyrproject-rtos/zephyr/pull/66042

Separating two new functions force and allow l1 to have the current state with separated functions in the power manager so that SOF can call these functions via IPC DMI_FORCE_L1_EXIT. DMA sometimes slows down and does not perform, therefore the workaround is to switch off l1. Change related to the addition of a new parameter to force DMI L1 exit on IPC request.

### DIFF
--- a/src/audio/base_fw.c
+++ b/src/audio/base_fw.c
@@ -13,6 +13,9 @@
 #include <sof/lib/cpu.h>
 #include <rtos/init.h>
 #include <platform/lib/clk.h>
+#if defined(CONFIG_SOC_SERIES_INTEL_ACE)
+#include <intel_adsp_hda.h>
+#endif
 
 #if CONFIG_ACE_V1X_ART_COUNTER || CONFIG_ACE_V1X_RTC_COUNTER
 #include <zephyr/device.h>
@@ -391,6 +394,42 @@ static int basefw_power_state_info_get(uint32_t *data_offset, char *data)
 	return 0;
 }
 
+static int fw_config_set_force_l1_exit(const struct sof_tlv *tlv)
+{
+#if defined(CONFIG_SOC_SERIES_INTEL_ACE)
+	const uint32_t force = tlv->value[0];
+
+	if (force) {
+		tr_info(&basefw_comp_tr, "FW config set force dmi l0 state");
+		intel_adsp_force_dmi_l0_state();
+	} else {
+		tr_info(&basefw_comp_tr, "FW config set allow dmi l1 state");
+		intel_adsp_allow_dmi_l1_state();
+	}
+
+	return 0;
+#else
+	return IPC4_UNAVAILABLE;
+#endif
+}
+
+static int basefw_set_fw_config(bool first_block,
+				bool last_block,
+				uint32_t data_offset,
+				const char *data)
+{
+	const struct sof_tlv *tlv = (const struct sof_tlv *)data;
+
+	switch (tlv->type) {
+	case IPC4_DMI_FORCE_L1_EXIT:
+		return fw_config_set_force_l1_exit(tlv);
+	default:
+		break;
+	}
+	tr_warn(&basefw_comp_tr, "returning success for Set FW_CONFIG without handling it");
+	return 0;
+}
+
 static int basefw_get_large_config(struct comp_dev *dev,
 				   uint32_t param_id,
 				   bool first_block,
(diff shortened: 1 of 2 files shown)


src/audio/base_fw.c
```diff
@@ -391,6 +392,35 @@ static int basefw_power_state_info_get(uint32_t *data_offset, char *data)
...
+static int basefw_set_fw_config(bool first_block,
+				bool last_block,
+				uint32_t data_offset,
+				const char *data)
+{
+	const struct sof_tlv *tlv = (const struct sof_tlv *)data;
+
+	switch (tlv->type) {
+	case IPC4_DMI_FORCE_L1_EXIT:
+		return fw_config_set_force_l1_exit(tlv);
```
plbossart, line 416: It wouldn't hurt to describe when host software is supposed to send this. My understanding of L1 is that this is a link/DMA state that has very large impacts on platform power consumption - or rather than staying in L0 can be problematic.

To the best of my knowledge this isn't supported today in the Linux driver.

@kv2019i is this on your radar?

lgirdwood: @fkwasowi can you add an entry into the logs when host enables/disables this features to help anyone debugging FW issues.

fkwasowi: I added log info and a more complete description of the PR change.

plbossart (approved):

lgirdwood (approved):
@fkwasowi pls ping once Zephyr side merged and then we can merge here..

iganakov (approved):
Looks good to me

kv2019i (changes requested):
I'm afraid we have multiple build errors for "lnl" and "tgl". We need a fixup to Zephyr upstream, it seems the L1_EXIT define is not available for lnl and for "tgl" inclusion of adsp_shim.h is causing problems.

src/audio/base_fw.c
```diff
@@ -13,6 +13,7 @@
 #include <sof/lib/cpu.h>
 #include <rtos/init.h>
 #include <platform/lib/clk.h>
+#include <intel_adsp_hda.h>
```
kv2019i, line 16: This is problematic. This file is not Intel specific so any Intel-specifc code and headers need to be ifdeffed (ideally put to a different file, but that is a bigger reworked and not needed now). But to not break builds, thi needs ifdef guards.

src/audio/base_fw.c
```diff
@@ -391,6 +392,38 @@ static int basefw_power_state_info_get(uint32_t *data_offset, char *data)
 	return 0;
 }
 
+static int fw_config_set_force_l1_exit(const struct sof_tlv *tlv)
+{
+	const uint32_t force = tlv->value[0];
+
+	if (force) {
+		tr_info(&basefw_comp_tr, "FW config set force dmi l0 state");
+		intel_adsp_force_dmi_l0_state();
```
kv2019i, line 401: You need to change the order of the two git commits. I.e. the west update needs to be first, and this commit needs to be second. Otherwise this breaks git-bisect. I.e. if somebody checks out the tree between your commits (which can happen in git-bisect), the tree cannot be build as intel_adsp_force_dmi_l0_state() is not defined).

kv2019i (changes requested):
Zephyr PR is now good and lnl compiles but still some issues on the SOF side changes, please see inline (and CI results).

kv2019i:
@fkwasowi The tgl/tgl-h still seems to fail, can you check:
https://github.com/thesofproject/sof/actions/runs/7245402754/job/19736029658?pr=8561

src/audio/base_fw.c
```diff
@@ -391,6 +394,42 @@ static int basefw_power_state_info_get(uint32_t *data_offset, char *data)
 	return 0;
 }
 
+static int fw_config_set_force_l1_exit(const struct sof_tlv *tlv)
+{
+	const uint32_t force = tlv->value[0];
+
+	if (force) {
+		tr_info(&basefw_comp_tr, "FW config set force dmi l0 state");
+#if defined(CONFIG_SOC_SERIES_INTEL_ACE)
```
kv2019i, line 403: Not sure what to return if the functionality is not available. Alternatively you could #ifdef the whole funciton body and at the end:

```
#if defined(CONFIG_SOC_SERIES_INTEL_ACE)
const uint32_t force = ..
...
#else
return IPC4_UNAVAILABLE;
#endif
```

kv2019i (approved):
Code looks good now. One suggestion about how to handle platforms that do not support this feature. I wonder if return an error would be more future-proof, but this is not blocking issue to me.

PR #8567, opened 2023-12-04
audio: perf: add module ibs and obs for performance profiling log

btian1:
Add ibs and obs information to performance logs, they will be used for update module CPC based on different ibs and obs.

### DIFF
--- a/src/include/sof/audio/component.h
+++ b/src/include/sof/audio/component.h
@@ -236,12 +236,13 @@ enum {
 		  (uint32_t)((pcd)->cpu_delta_peak))
 
 #define comp_perf_avg_info(pcd, comp_p)					\
-	comp_info(comp_p, "perf comp_copy samples %u period %u cpu avg %u peak %u %u",\
+	comp_info(comp_p, "perf comp_copy samples %u period %u cpu avg %u peak %u ibs %u obs %u",\
 		  (uint32_t)((comp_p)->frames),            \
 		  (uint32_t)((comp_p)->period),			    \
 		  (uint32_t)((pcd)->cpu_delta_sum),			\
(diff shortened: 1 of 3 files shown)


lyakh (dismissed):

src/include/sof/audio/component.h
```diff
@@ -236,12 +236,14 @@ enum {
 		  (uint32_t)((pcd)->cpu_delta_peak))
 
 #define comp_perf_avg_info(pcd, comp_p)					\
-	comp_info(comp_p, "perf comp_copy samples %u period %u cpu avg %u peak %u %u",\
+	comp_info(comp_p, "perf comp_copy samples %u period %u cpu avg %u peak %u %u %u %u",\
```
aiChaoSONG, line 239: can we at least add iobs in the log?

`perf comp_copy samples %u period %u cpu avg %u peak %u %u iobs %u %u`

btian1: just want to minimize the change for sof and script both, since parsing can directly pick ibs obs with a fixed offset.

aiChaoSONG: But readability is also important, no one knows what they are until looking and understanding the code. Even you add iobs, the values are still at some fixed offset.

lyakh: is the "old" (xtos / sof-logger) limit of a maximum of 4 variables in logs now completely abandoned? What about all the platforms still using XTOS?

btian1: @lgirdwood , yes it is runtime read out from each module.

@lyakh , this is good point, I did not test IPC3 case, if consider IPC3 case, we may need separate logs to two lines or remove some variable print.

PR #8527, opened 2023-11-27
topology2: sof-mtl-rt713-l0-rt1316-l12-rt1713-l3: Enable ChainDMA

ujfalusi:
Enable chain DMA support for HDMI links.

Tested on device with 6.7.0 based kernel.

PR #8568, opened 2023-12-04
topology2: Enable ChainDMA for HDMI by default

ujfalusi:
Hi,

going through all configurations and set the flag one-by-one does not scale and the HDA topologies have the ChainDMA enabled for HDMI for quite some time now while some SDW enables it, some does not.

It can be selectively disabled for configurations if needed, but ChainDMA should be stable now to be default on display side.

Replaces: #8527


kv2019i (approved):

lgirdwood (approved):

ujfalusi:
Changes since v1:
- Move the `HDMI_USE_CHAIN_DMA "true"` to `tools/topology/topology2/platform/intel/hdmi-default.conf` instead of scattered around in different config files (for HDA, SDW, etc)

plbossart (approved):
Nice cleanup, thanks @ujfalusi

PR #8573, opened 2023-12-05
zephyr: cpu: ipc reinit after rtd3

tmleman:
This patch enables PM_DEVICE_RUNTIME_EXCLUSIVE. So far, some devices have been suspended by PM during power transition to D3 (this was actually only done for the IPC device). This device suspend step was skip if device was busy, but system suspend was continued. This behavior resulted in an incorrect IPC device state after the context was restored.

Since D3 is system power-off, all devices are restarted and we don't need to do any IPC suspend. SOF already check conditions for successful shutdown and if everything is OK we can just go to power-off in idle.

Until the IPC driver is reworked to enable normal initialization without additional conditions we need to manually trigger suspend and resume actions.

Depends on:

- [ ] https://github.com/zephyrproject-rtos/zephyr/pull/66135
- [ ] zephyr update in west


app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -12,7 +12,7 @@ CONFIG_COMP_MULTIBAND_DRC=y
 CONFIG_PM=y
 CONFIG_PM_DEVICE=y
 CONFIG_PM_DEVICE_RUNTIME=y
-CONFIG_PM_DEVICE_RUNTIME_EXCLUSIVE=n
+CONFIG_PM_DEVICE_RUNTIME_EXCLUSIVE=y
```
kv2019i, line 15: Some explanation about this option would be beneficial (the one-liner from Zephyr Kconfig should be ok). Also to understand this patch, it helps to understand this is set by default and this was disabled for specific Intel platforms only.

zephyr/lib/cpu.c
```diff
@@ -63,6 +63,7 @@ static FUNC_NORETURN void secondary_init(void *arg)
 #if CONFIG_ZEPHYR_NATIVE_DRIVERS
 #include <sof/trace/trace.h>
 #include <rtos/wait.h>
+#include <intel_adsp_ipc.h>
```
kv2019i, line 66: This needs to be wrapped into ifdef CONTEXT_SAVE. This is in general a bit iffy to have Intel specific code here when this code should be generic for all vendors. But alas we have the context save code already in this file, but let's at least use that so you can compile this file for other platforms.

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -12,7 +12,7 @@ CONFIG_COMP_MULTIBAND_DRC=y
 CONFIG_PM=y
```
kv2019i, line 12: I'm having a bit hard time to understand the commit:
```
Since D3 is system power-off, all devices are restarted and we don't
need to do any IPC suspend. SOF already check conditions for successful
shutdown and if everything is OK we can just go to power-off in idle.
```

Is this talking about case of IMR context-save case or regular D3? I'm guess in regular case, suspend is not needed as we'll start from scratch anyways. But in context-save case,  why do we actually have to suspend? The state of the IPC driver should be restored from IMR, so it should just continue from whatever state it was before state save. OTOH, if there's a reason, then this patch seems to be the way to do it, but the commit is not very easy to follow.

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -12,7 +12,7 @@ CONFIG_COMP_MULTIBAND_DRC=y
...
 CONFIG_PM_DEVICE=y
 CONFIG_PM_DEVICE_RUNTIME=y
-CONFIG_PM_DEVICE_RUNTIME_EXCLUSIVE=n
+CONFIG_PM_DEVICE_RUNTIME_EXCLUSIVE=y
 CONFIG_PM_DEVICE_POWER_DOMAIN=y
 CONFIG_PM_POLICY_CUSTOM=y
```
kv2019i, line 18: And this as well:
```
Until the IPC driver is reworked to enable normal initialization without
additional conditions we need to manually trigger suspend and resume
actions.
```

Does this cover context-save only, or some more general problem? What are the "additional conditions"?

kv2019i (commented):
Some comments inline.

lgirdwood (commented):
@tmleman we should probably comment in the code why we need to manually trigger suspend and resume actions so that its obvious to everyone (and it will be easier for other to follow the next steps).

tmleman:
I forgot about that pull request and created a new one, which has already been merged. I'm closing this one.

Issue #8275, opened 2023-09-29
[BUG] LNL has FW reported error: 7 with multiple pipeline playback


fredoh9:
**Describe the bug**
First LNL issue.
For LNLM_RVP_NOCODEC, multiple pipeline playback fails with FW reported error: 7 - Unsupported operation requested.

```
[ 1328.146968] kernel: snd_sof:sof_ipc4_set_pipeline_state: sof-audio-pci-intel-lnl 0000:00:1f.3: ipc4 set pipeline instance 6 state 3
[ 1328.146983] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-lnl 0000:00:1f.3: ipc tx      : 0x13060003|0x0: GLB_SET_PIPELINE_STATE
[ 1328.147386] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-lnl 0000:00:1f.3: ipc tx reply: 0x33000000|0x0: GLB_SET_PIPELINE_STATE
[ 1328.147493] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-lnl 0000:00:1f.3: ipc tx done : 0x13060003|0x0: GLB_SET_PIPELINE_STATE
[ 1328.147521] kernel: snd_sof:sof_ipc4_set_pipeline_state: sof-audio-pci-intel-lnl 0000:00:1f.3: ipc4 set pipeline instance 6 state 4
[ 1328.147532] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-lnl 0000:00:1f.3: ipc tx      : 0x13060004|0x0: GLB_SET_PIPELINE_STATE
[ 1328.149150] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-lnl 0000:00:1f.3: ipc tx reply: 0x33000007|0x0: GLB_SET_PIPELINE_STATE
[ 1328.149162] kernel: sof-audio-pci-intel-lnl 0000:00:1f.3: FW reported error: 7 - Unsupported operation requested
[ 1328.149388] kernel: sof-audio-pci-intel-lnl 0000:00:1f.3: ipc error for msg 0x13060004|0x0
[ 1328.149420] kernel: sof-audio-pci-intel-lnl 0000:00:1f.3: ASoC: error at soc_dai_trigger on SSP1 Pin: -22
[ 1328.149439] kernel:  Port1: ASoC: trigger FE cmd: 1 failed: -22
```

[...]

PR #8574, opened 2023-12-05
copier: store gateway config in copier_data

iganakov:
We need to keep `gtw_cfg.config_data` during Copier lifetime to be able to retrieve dma_channel_id when process SET_PIPELINE_STATE(RUN) IPC. Please refer to the detailed explanation below:
https://github.com/thesofproject/sof/issues/8275#issuecomment-1822333476

Use previously saved gateway config to find DMA config and assign correct pointer to `host_dma_config` for SSP and DMIC.
This PR should fix issues reported in https://github.com/thesofproject/sof/issues/8275 for SSP and DMIC interfaces.

### DIFF
--- a/src/audio/copier/copier.c
+++ b/src/audio/copier/copier.c
@@ -64,6 +64,8 @@ static int copier_init(struct processing_module *mod)
 	struct module_data *md = &mod->priv;
 	struct ipc4_copier_module_cfg *copier = (struct ipc4_copier_module_cfg *)md->cfg.init_data;
 	struct comp_ipc_config *config = &dev->ipc_config;
+	void *gtw_cfg = NULL;
+	size_t gtw_cfg_size;
 	int i, ret = 0;
 
 	cd = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*cd));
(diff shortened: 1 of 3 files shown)


serhiy-katsyuba-intel (approved):

src/audio/copier/copier.h
```diff
@@ -242,6 +242,7 @@ struct copier_data {
 	 * from the IPC data.
 	 */
 	struct ipc4_copier_module_cfg config;
+	void *gtw_cfg;
```
lgirdwood, line 245: if this is IPC4 struct, we can use IPC4 type instead of void ?

lgirdwood (commented):
Just one minor question. @plbossart did you have a similar PR ? good for you ?

plbossart:
@fredoh9 mentioned he'd give this a try to see if indeed it solves issue #8275, let's wait for the test results.

fredoh9:
No luck with this pr. I tested it with LNLM_SDW_AIOC. multiple-pipeline test has same fw error 7.
linux commit id = [266ee31b71f8](https://github.com/thesofproject/linux/commit/266ee31b71f88434f317d78a9ce3e3aec31fdbd8)
sof commit id = 7a76e146fae1db8c8830c4c9c9ab1644addde18e

[dmesg.txt](https://github.com/thesofproject/sof/files/13568283/dmesg.txt)
[mtrace.txt](https://github.com/thesofproject/sof/files/13568141/mtrace.txt)

plbossart:
> No luck with this pr. I tested it with LNLM_SDW_AIOC. 

@fredoh9 this PR is ONLY for SSP and DMIC, using LNL_SDW_AIOC will not work since the work is not complete just yet for SoundWire.

RanderWang (approved):

src/audio/copier/copier.c
```diff
@@ -172,6 +200,8 @@ static int copier_free(struct processing_module *mod)
 		break;
 	}
 
+	if (cd)
+		rfree(cd->gtw_cfg);
```
btian1, line 204: can we put line 203 and 204 to copier_ipcgtw_free? put here is not correct, since cd is true without gtw case.

iganakov: `cd->gtw_cfg ` is used in dai configuration so I'd rather move it to `copier_dai_free()`. Does that sound ok?

fredoh9:
Single pipeline playack/capture in LNL NOCODEC have been broken with DSP panic.
Anyway I tried this PR on LNLM_RVP_NOCODEC. but still seeing DSP panic.

fredoh9:
With the fix for DSP panic https://github.com/zephyrproject-rtos/zephyr/pull/66215 and this pr, 
playback and capture work fine. Never been like this stable.
For multiple pipeline test BOTH playback and capture work fine too. iIeration parameter is 50.
Looks great.

src/audio/copier/copier.c
btian1: that's fine if you confirmed there is no case for host.
and add if (cd->ipcgtw) in dai as well.

src/audio/copier/copier.c
```diff
@@ -78,7 +80,31 @@ static int copier_init(struct processing_module *mod)
...
+		goto error_cd;
+	}
+
+	/* Allocate memory and store gateway_cfg in runtime. Gateway cfg has to
+	 * be kept even after copier is created e.g. during SET_PIPELINE_STATE
+	 * IPC when dai_config_dma_channel() is called second time and DMA
+	 * config is used to assign dma_channel_id value.
+	 */
+	if (copier->gtw_cfg.config_length) {
+		gtw_cfg_size = copier->gtw_cfg.config_length << 2;
```
kv2019i, line 92: Took me a while to figure out why the "<< 2" here by apparently config_length is in dwords.

kv2019i (approved):

PR #8575, opened 2023-12-05
module: Add support for saving and retrieving the base config extension

ranj063:
Modules are bound with one another with a buffer in between. Today, this buffer's size is based on the source module's base config obs and sink module's base config ibs. But this is incorrect because this assumes that the connection is always from output pin 0 of the source module and input pin 0 of the sink module.

To determine the buffer size correctly, the buffer size should be based on the pin-based obs/ibs of the source/sink module. To allow this, save the base config extension for the module during init so that it can be retrieved during module binding to get the obs/ibs based on the source/sink pin ID.

Modify the google rtc and smart amp modules to save the base config extension during their init.

### DIFF
--- a/src/audio/google/google_rtc_audio_processing.c
+++ b/src/audio/google/google_rtc_audio_processing.c
@@ -392,18 +392,36 @@ static int google_rtc_audio_processing_init(struct processing_module *mod)
 #if CONFIG_IPC_MAJOR_4
 	const struct ipc4_base_module_extended_cfg *base_cfg = md->cfg.init_data;
 	struct ipc4_input_pin_format reference_fmt, output_fmt;
-	const size_t size = sizeof(struct ipc4_input_pin_format);
+	size_t in_fmt_size = sizeof(struct ipc4_input_pin_format);
+	size_t out_fmt_size = sizeof(struct ipc4_output_pin_format);
+	size_t size;
 
 	cd->config.base_cfg = base_cfg->base_cfg;
 
 	/* Copy the reference format from input pin 1 format */
-	memcpy_s(&reference_fmt, size,
-		 &base_cfg->base_cfg_ext.pin_formats[size], size);
-	memcpy_s(&output_fmt, size,
-		 &base_cfg->base_cfg_ext.pin_formats[size * GOOGLE_RTC_NUM_INPUT_PINS], size);
+	memcpy_s(&reference_fmt, in_fmt_size,
+		 &base_cfg->base_cfg_ext.pin_formats[in_fmt_size], in_fmt_size);
+	memcpy_s(&output_fmt, in_fmt_size,
+		 &base_cfg->base_cfg_ext.pin_formats[in_fmt_size * GOOGLE_RTC_NUM_INPUT_PINS],
+		 in_fmt_size);
 
 	cd->config.reference_fmt = reference_fmt.audio_fmt;
 	cd->config.output_fmt = output_fmt.audio_fmt;
+
+	/* save the base config extension */
+	size = base_cfg->base_cfg_ext.nb_input_pins * in_fmt_size +
+		base_cfg->base_cfg_ext.nb_output_pins * out_fmt_size;
+
+	md->cfg.basecfg_ext = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
+				      sizeof(struct ipc4_base_module_cfg_ext) + size);
+	if (!md->cfg.basecfg_ext) {
+		ret = -ENOMEM;
+		goto fail;
+	}
+
+	memcpy_s(md->cfg.basecfg_ext, size + sizeof(struct ipc4_base_module_cfg_ext),
+		 &base_cfg->base_cfg_ext,
+		 size + sizeof(struct ipc4_base_module_cfg_ext));
 #endif
 
 	cd->tuning_handler = comp_data_blob_handler_new(dev);
@@ -510,6 +528,9 @@ fail:
 		rfree(cd->memory_buffer);
 		rfree(cd->raw_mic_buffer);
 		comp_data_blob_handler_free(cd->tuning_handler);
+#if CONFIG_IPC_MAJOR_4
+		rfree(md->cfg.basecfg_ext);
+#endif
 		rfree(cd);
 	}
 
@@ -530,6 +551,11 @@ static int google_rtc_audio_processing_free(struct processing_module *mod)
 	rfree(cd->memory_buffer);
 	rfree(cd->raw_mic_buffer);
 	comp_data_blob_handler_free(cd->tuning_handler);
+#if CONFIG_IPC_MAJOR_4
+	struct module_data *md = &mod->priv;
+
+	rfree(md->cfg.basecfg_ext);
+#endif
 	rfree(cd);
 	return 0;
 }
--- a/src/audio/module_adapter/module_adapter_ipc4.c
+++ b/src/audio/module_adapter/module_adapter_ipc4.c
@@ -148,6 +148,22 @@ int module_adapter_get_attribute(struct comp_dev *dev, uint32_t type, void *valu
 		memcpy_s(value, sizeof(struct ipc4_base_module_cfg),
 			 &mod->priv.cfg.base_cfg, sizeof(mod->priv.cfg.base_cfg));
 		break;
+	case COMP_ATTR_BASE_CONFIG_EXT:
+	{
+		struct ipc4_base_module_cfg_ext *basecfg_ext = mod->priv.cfg.basecfg_ext;
+		size_t size;
+
+		if (!basecfg_ext) {
+			comp_err(mod->dev, "No base config extn set for module");
+			return -EINVAL;
+		}
+
+		size = basecfg_ext->nb_input_pins * sizeof(struct ipc4_input_pin_format) +
+				basecfg_ext->nb_output_pins * sizeof(struct ipc4_output_pin_format);
+		memcpy_s(value, sizeof(*basecfg_ext) + size,
+			 basecfg_ext, sizeof(*basecfg_ext) + size);
+		break;
+	}
 	default:
 		return -EINVAL;
 	}
--- a/src/include/sof/audio/component_ext.h
+++ b/src/include/sof/audio/component_ext.h
@@ -16,6 +16,7 @@
 #ifndef __SOF_AUDIO_COMPONENT_INT_H__
 #define __SOF_AUDIO_COMPONENT_INT_H__
 
+#include <sof/audio/module_adapter/module/generic.h>
 #include <sof/audio/component.h>
 #include <ipc/topology.h>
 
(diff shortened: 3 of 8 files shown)


src/audio/google/google_rtc_audio_processing.c
```diff
@@ -404,6 +404,21 @@ static int google_rtc_audio_processing_init(struct processing_module *mod)
 
 	cd->config.reference_fmt = reference_fmt.audio_fmt;
 	cd->config.output_fmt = output_fmt.audio_fmt;
+
+	/* save the base config extension */
+	size = base_cfg->base_cfg_ext.nb_input_pins * sizeof(struct ipc4_input_pin_format) +
```
yongzhi1, line 409: cannot assign to variable 'size' with const-qualified type 'const size_t'

ranj063: fixed. thanks!

src/ipc/ipc4/helper.c
```diff
@@ -453,32 +459,85 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
+		src_basecfg_ext = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
+					  sizeof(*src_basecfg_ext) + size);
+		if (!src_basecfg_ext)
+			return IPC4_OUT_OF_MEMORY;
+
+		ret = comp_get_attribute(source, COMP_ATTR_BASE_CONFIG_EXT, src_basecfg_ext);
+		if (!ret) {
+			size_t in_fmt_size = sizeof(struct ipc4_input_pin_format);
+			size_t out_fmt_size = sizeof(struct ipc4_output_pin_format);
+			output_fmt_offset =
```
btian1, line 479: I think use uint8_t pin_formats[] is not suitable, it should be a structure? do you agree? as we don't know the exact size and data inside.

src/ipc/ipc4/helper.c
```diff
@@ -453,32 +459,85 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
+				src_basecfg_ext->nb_input_pins * in_fmt_size +
+				bu->extension.r.src_queue * out_fmt_size;
+
+			memcpy_s(&out_fmt, sizeof(struct ipc4_output_pin_format),
+				 &src_basecfg_ext->pin_formats[output_fmt_offset],
+				 sizeof(struct ipc4_output_pin_format));
```
btian1, line 485: seems we may not need use memcpy_s, can we directly use 
out_format = (struct ipc4_output_pin_format *)&src_basecfg_ext->pin_formats[output_fmt_offset];

src/include/sof/audio/component_ext.h
```diff
@@ -216,31 +217,59 @@ struct get_attribute_remote_payload {
 static inline int comp_ipc4_get_attribute_remote(struct comp_dev *dev, uint32_t type,
 						 void *value)
 {
-	struct ipc4_base_module_cfg *base_cfg;
+	struct ipc4_base_module_cfg_ext *basecfg_ext = NULL;
+	struct ipc4_base_module_cfg *base_cfg = NULL;
```
lyakh, line 221: you never need both pointers, right? So you could just use one, the only difference is the size, otherwise you'd reuse the same code.

src/ipc/ipc4/helper.c
lyakh: @btian1 is proposing to have `out_fmt` as a pointer and just point it into the array to read out the `obs`

src/ipc/ipc4/helper.c
```diff
@@ -453,32 +459,85 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
+
+		ret = comp_get_attribute(sink, COMP_ATTR_BASE_CONFIG_EXT, sink_basecfg_ext);
+		if (!ret) {
+			input_fmt_offset =
+				bu->extension.r.dst_queue * sizeof(struct ipc4_input_pin_format);
+
+			memcpy_s(&in_fmt, sizeof(struct ipc4_input_pin_format),
+				 &sink_basecfg_ext->pin_formats[input_fmt_offset],
+				 sizeof(struct ipc4_input_pin_format));
+			ibs = in_fmt.ibs;
```
lyakh, line 522: same here with a pointer

src/ipc/ipc4/helper.c
marcinszkudlinski: This is sequence of 2 diffrent structures 
better to have a simple offset, anyway I agree that this is not easy to read :(

marcinszkudlinski:
LGTM

src/include/sof/audio/component_ext.h
lgirdwood: two pointers does help with type correctness, in case we ever have to deref. however, I cant see from here if one struct is embedded in the other (then 1 ptr makes sense).

src/ipc/ipc4/helper.c
lgirdwood: yeah, some inline comments would help.

src/ipc/ipc4/helper.c
```diff
@@ -453,32 +459,85 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
-	 * in case of DP -> LL
-	 *	size = 2*ibs of destination (LL) module. DP queue will handle obs of DP module
-	 */
-	uint32_t buf_size;
+	/* get ibs from the base config extension if the sink queue ID is non-zero */
+	if (bu->extension.r.dst_queue) {
+		struct ipc4_input_pin_format in_fmt;
+		size_t input_fmt_offset;
+
+		sink_basecfg_ext = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
```
lgirdwood, line 509: Btw, if these are always copied and freed (in the same func), why not just have on the stack ?

lgirdwood (commented):
Just some minor questions from me.

src/include/sof/audio/component_ext.h
ranj063: no they're not embedded. i will move the free's to the respective switches and remove the inits here

src/ipc/ipc4/helper.c
ranj063: because basecfg_ext structure contains a flexible array member for the number of pin formats

udaymb:
This patch is verified on Chrome rex system and is working without issue

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -392,18 +392,36 @@ static int google_rtc_audio_processing_init(struct processing_module *mod)
...
+	memcpy_s(&output_fmt, in_fmt_size,
+		 &base_cfg->base_cfg_ext.pin_formats[in_fmt_size * GOOGLE_RTC_NUM_INPUT_PINS],
+		 in_fmt_size);
 
 	cd->config.reference_fmt = reference_fmt.audio_fmt;
 	cd->config.output_fmt = output_fmt.audio_fmt;
+
+	/* save the base config extension */
+	size = base_cfg->base_cfg_ext.nb_input_pins * in_fmt_size +
+		base_cfg->base_cfg_ext.nb_output_pins * out_fmt_size;
```
lyakh, line 413: nitpick: you could have
```
	size = base_cfg->base_cfg_ext.nb_input_pins * in_fmt_size +
		base_cfg->base_cfg_ext.nb_output_pins * out_fmt_size +
		sizeof(struct ipc4_base_module_cfg_ext);
```

src/audio/module_adapter/module_adapter_ipc4.c
```diff
@@ -148,6 +148,22 @@ int module_adapter_get_attribute(struct comp_dev *dev, uint32_t type, void *valu
...
+		struct ipc4_base_module_cfg_ext *basecfg_ext = mod->priv.cfg.basecfg_ext;
+		size_t size;
+
+		if (!basecfg_ext) {
+			comp_err(mod->dev, "No base config extn set for module");
+			return -EINVAL;
+		}
+
+		size = basecfg_ext->nb_input_pins * sizeof(struct ipc4_input_pin_format) +
+				basecfg_ext->nb_output_pins * sizeof(struct ipc4_output_pin_format);
```
lyakh, line 162: similar here with `size`

src/include/sof/audio/component_ext.h
```diff
@@ -216,31 +217,59 @@ struct get_attribute_remote_payload {
...
-			 base_cfg, sizeof(struct ipc4_base_module_cfg));
+	if (!ret) {
+		switch (type) {
+		case COMP_ATTR_BASE_CONFIG:
+			memcpy_s(value, sizeof(struct ipc4_base_module_cfg),
+				 base_cfg, sizeof(struct ipc4_base_module_cfg));
+			rfree(base_cfg);
+			break;
+		case COMP_ATTR_BASE_CONFIG_EXT:
+			memcpy_s(value, size, basecfg_ext, size);
```
lyakh, line 267: here without `sizeof(*basecfg_ext)`, right?

src/ipc/ipc4/helper.c
```diff
@@ -453,32 +459,88 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
+	if (bu->extension.r.src_queue) {
+		struct ipc4_output_pin_format *out_fmt;
+		size_t output_fmt_offset;
+
+		src_basecfg_ext = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
+					  sizeof(*src_basecfg_ext) + size);
+		if (!src_basecfg_ext)
+			return IPC4_OUT_OF_MEMORY;
+
+		ret = comp_get_attribute(source, COMP_ATTR_BASE_CONFIG_EXT, src_basecfg_ext);
```
lyakh, line 475: this looks scary... Can we add a `size` argument to `comp_get_attribute()`?

src/ipc/ipc4/helper.c
lyakh: could we add a component method to get a pin format?

src/ipc/ipc4/helper.c
```diff
@@ -453,32 +459,88 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
+	if (!ibs) {
+		ret = comp_get_attribute(sink, COMP_ATTR_BASE_CONFIG, &sink_src_cfg);
+		if (ret < 0) {
+			tr_err(&ipc_tr, "failed to get base config for sink module %#x",
+			       dev_comp_id(sink));
+			return IPC4_FAILURE;
+		}
+
+		ibs = sink_src_cfg.ibs;
+	}
```
kv2019i, line 540: Another minor note: would be nice to split out into subfunctions somehow, this function is getting very, very long.

kv2019i (commented):
Please check the one functional comment. Multiple minor ones, but not necessary to address in this PR.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -392,18 +392,36 @@ static int google_rtc_audio_processing_init(struct processing_module *mod)
...
 	/* Copy the reference format from input pin 1 format */
-	memcpy_s(&reference_fmt, size,
-		 &base_cfg->base_cfg_ext.pin_formats[size], size);
-	memcpy_s(&output_fmt, size,
-		 &base_cfg->base_cfg_ext.pin_formats[size * GOOGLE_RTC_NUM_INPUT_PINS], size);
+	memcpy_s(&reference_fmt, in_fmt_size,
+		 &base_cfg->base_cfg_ext.pin_formats[in_fmt_size], in_fmt_size);
+	memcpy_s(&output_fmt, in_fmt_size,
+		 &base_cfg->base_cfg_ext.pin_formats[in_fmt_size * GOOGLE_RTC_NUM_INPUT_PINS],
+		 in_fmt_size);
```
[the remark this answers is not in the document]
btian1: and since size is same, I would prefer use memcpy(dst, src, size) to reduce one parameters.

src/audio/google/google_rtc_audio_processing.c
btian1: @lyakh , no , see line 416, extra struct size are added when rzalloc

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -392,18 +392,36 @@ static int google_rtc_audio_processing_init(struct processing_module *mod)
...
+	md->cfg.basecfg_ext = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
+				      sizeof(struct ipc4_base_module_cfg_ext) + size);
+	if (!md->cfg.basecfg_ext) {
+		ret = -ENOMEM;
+		goto fail;
+	}
+
+	memcpy_s(md->cfg.basecfg_ext, size + sizeof(struct ipc4_base_module_cfg_ext),
+		 &base_cfg->base_cfg_ext,
+		 size + sizeof(struct ipc4_base_module_cfg_ext));
```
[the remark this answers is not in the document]
btian1: this is not a complex operation, we should make code more straightforward to read.
total_size = size + sizeof(struct ipc4_base_module_cfg_ext);

src/audio/google/google_rtc_audio_processing.c
lyakh: @btian1 that's exactly why I'm proposing that

src/audio/google/google_rtc_audio_processing.c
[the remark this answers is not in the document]
lgirdwood: yeah, the array index [in_fmt_size], looks like copy paste error (even though its replacing a similar size/index bit of code) ?

marcinszkudlinski (approved):
Changes verified om 007, working very well, please proceed with merging

lgirdwood:
> Changes verified om 007, working very well, please proceed with merging

Thanks @marcinszkudlinski for validating !

@ranj063 fyi, some conflicts.

andyross:
It's probably way too disruptive for me to -1 this given that it's already been pushed into mtl-007-drop-stable, but see #8685 for a different take on the same problem that IMHO is a lot cleaner long-term.

Basically: I... kinda really, really hate the API picked here: Module code (literally every module that might in principle be connected to more than two other components) needs to make a copy of raw IPC memory that it finds in the module_config, set it **back** on a different field in module_config (for the benefit of ibs/obs buffer sizing, which it can't control), and then set that metadata back onto the connected buffers at prepare time.  Literally none of those steps should be the responsibility of module code.

Module code is written by dumb third party people like me who don't understand the module_adapter internals!

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -392,18 +392,36 @@ static int google_rtc_audio_processing_init(struct processing_module *mod)
...
 
 	cd->config.reference_fmt = reference_fmt.audio_fmt;
 	cd->config.output_fmt = output_fmt.audio_fmt;
+
+	/* save the base config extension */
+	size = base_cfg->base_cfg_ext.nb_input_pins * in_fmt_size +
+		base_cfg->base_cfg_ext.nb_output_pins * out_fmt_size;
+
+	md->cfg.basecfg_ext = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
+				      sizeof(struct ipc4_base_module_cfg_ext) + size);
```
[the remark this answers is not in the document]
marcinszkudlinski: @andyross that is not true in this case.
All operations - create/configure/run/delete/whatever will be performed by a code the module is bound to. No exceptions. Using cached structures is safe

ranj063:
replaced by #8685

PR #8537, opened 2023-11-27
ipc4: mixin/mixout: Switch to use sink/source API

serhiy-katsyuba-intel:
Modifications to mixin/mixout implementation to switch to use sink/source API.

PR #8580, opened 2023-12-06
 Add dummy Loadable smart amp support

RanderWang:


lmdk/modules/smart_amp_test/smart_amp_test.c
```diff
@@ -0,0 +1,317 @@
...
+{
+	const struct ipc4_base_module_extended_cfg *base_cfg;
+	struct module_data *mod_data = &mod->priv;
+	struct smart_amp_data *sad;
+	uint8_t *pin_data;
+	int bs;
+	int ret;
+
+	base_cfg = mod_data->cfg.init_data;
+	sad = &smart_amp_priv;
```
lyakh, line 37: can and shall we add a check that this function isn't called when 1 instance of the module is already running?

RanderWang: Only one smart amp in topology and the pcm can't be used two times at the same time, so no need to check it.

src/audio/module_adapter/module_adapter.c
```diff
@@ -1107,6 +1114,8 @@ static int module_adapter_sink_source_copy(struct comp_dev *dev)
 
 	comp_dbg(dev, "module_adapter_sink_source_copy(): start");
 
+	module_adapter_sink_src_update(dev, mod);
```
marcinszkudlinski, line 1117: Why you want to get new set of sink/src pointers before each copy? 
What is "dynamic multi-source issue"?

lmdk/cmake/build.cmake
```diff
@@ -25,6 +25,7 @@ foreach(MODULE ${MODULES_LIST})
 		"${LMDK_BASE}/include"
 		"${RIMAGE_INCLUDE_DIR}"
 		"${SOF_BASE}/src/include/module"
+		"${SOF_BASE}/src/include"
```
lgirdwood, line 28: Is this fix to be squashed ? (its missing a commit message)

src/audio/module_adapter/module/modules.c
```diff
@@ -143,6 +143,7 @@ static int modules_init(struct processing_module *mod)
 					(struct module_interface *)md->module_adapter;
 
 		ret = mod_in->init(mod);
+		mod->priv.ops = mod_in;
```
lgirdwood, line 146: Will need a commit message too.

lmdk/modules/smart_amp_test/smart_amp_test.c
```diff
@@ -0,0 +1,317 @@
+// SPDX-License-Identifier: BSD-3-Clause
```
lgirdwood, line 1: This needs to be in modules directory and we need to reuse code, we cant have 2 implementations of dummy smart amp. i.e. I would use Kconfig to select which functions to build in library mode and which to build in as I imagine most of the code will be the same.

lmdk/modules/smart_amp_test/smart_amp_test.h
```diff
@@ -0,0 +1,122 @@
+/* SPDX-License-Identifier: BSD-3-Clause
```
lgirdwood, line 1: This file needs to be in the common modules directory alongside the dummy smart amp implementation.

lgirdwood (changes requested):
Some of this PR is fixed and can merge with commit messages, but we need reuse and a single dummy smart amp code base. 
We should only have 1 code base for each module regardless of how it's used (built-in or module)

src/audio/module_adapter/module_adapter.c
serhiy-katsyuba-intel: This commit does fix the CI issue reported for https://github.com/thesofproject/sof/pull/8537 PR. Verified on draft PR https://github.com/thesofproject/sof/pull/8576 with https://github.com/thesofproject/sof/pull/8580/commits/1855581b2353403caaa307e9597659751e316f46 commit included.

Here is an example use case that requires this commit: consider two pipelines, one with mixin and the other with mixout, mixin is connected to mixout. After pipelines started, .prepare() is called for mixin and mixout. So now mixout has num_of_sources set to 1. After some time driver creates third pipeline containing mixin, binds that mixin to mixout and starts this new pipeline. .prepare() will be called for this new mixin, however, .prepare() will not be called for mixout!!! And mixout's num_of_sources is still 1, which is now wrong. This is because mixout pipeline was already running. To call .prepare() for mixout, its pipeline should be paused, then reset and then started again. However, driver does not do this.

So this commit fixes the problem. However, you are right, it is excessive to update sinks/sources at every copy call for each module. It would be better to move the module_adapter_sink_src_update() to bind() and unbind() handlers.

src/audio/module_adapter/module/modules.c
ranj063: @RanderWang even if this is missing, everywhere we call mod_in in prepare/process, we always assign before using. What error does this fix?

lmdk/libraries/smart_amp_test/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+cmake_minimum_required(VERSION 3.20)
```
ranj063, line 1: @RanderWang do we have instructions on building a loadable module in our documentation? If so, could you please point me to it? If not, can you please add it?

Also, once this PR is merged, will our CI automatically start testing the smartamp as a loadable module? or will that be a follow up?

mengdonglin: @ranj063 Once this PR is merged, our CI will automatically start testing it as a loadable module.

src/audio/module_adapter/module/modules.c
RanderWang: Do we assign it before using in prepare or process ? For built-in module, it is defined by the module declaration, but for loadable module it is defined by a default interface, so we need to overwrite it by private loadable module interface
````c
/* Processing Module Adapter API*/
static const struct module_interface interface = {
        .init = modules_init,
        .prepare = modules_prepare,
        .process_raw_data = modules_process,
        .set_processing_mode = modules_set_processing_mode,
        .get_processing_mode = modules_get_processing_mode,
        .set_configuration = modules_set_configuration,
        .get_configuration = modules_get_configuration,
        .reset = modules_reset,
        .free = modules_free,
};
````

File: lmdk/libraries/smart_amp_test/CMakeLists.txt (hunk 7 above)
Comment (RanderWang) [reply lmdk/libraries/smart_amp_test/CMakeLists.txt:1]: we have a batch shell to build it, I will also push it.    To test it in CI ? we need to first disable built-in smart amp and then remove smart amp in toml file ( waiting for toml file rework ?)

Comment (RanderWang) [reply lmdk/libraries/smart_amp_test/CMakeLists.txt:1]: BTW this is demo in lmdk directory, not in our building system, so more work to do for CI

Comment (mengdonglin) [reply lmdk/libraries/smart_amp_test/CMakeLists.txt:1]: @RanderWang  @ranj063  CI automated testing can be planned as a follow up. My previous understanding was wrong. We needs to update CI storage and deployment to support loadable modules.

File: src/audio/module_adapter/module_adapter.c (hunk 2 above)
Comment (RanderWang) [reply src/audio/module_adapter/module_adapter.c:1117]: sure, I agree to move it to bind & unbind. Actually I did it in bind & unbind. Let's change it.

Comment (RanderWang) [reply src/audio/module_adapter/module_adapter.c:1117]: for smart amp, the reference buffer from capture will attach to playback stream, so the source count will be changed. This is the issue

**Review by marcinszkudlinski — CHANGES_REQUESTED** [review]

File: src/audio/module_adapter/module/modules.c (hunk 4 above)
Comment (RanderWang) [reply src/audio/module_adapter/module/modules.c:146]: I find this change is not correct. The overwrite will lost some functions

File: src/audio/module_adapter/module/generic.c [hunk 8]
```diff
@@ -519,6 +546,9 @@ int module_unbind(struct processing_module *mod, void *data)
 {
 	struct module_data *md = &mod->priv;
 
+	if (IS_PROCESSING_MODE_SINK_SOURCE(mod))
```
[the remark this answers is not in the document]
RanderWang: thanks

src/audio/module_adapter/module/generic.c
```diff
@@ -506,10 +506,37 @@ int module_set_configuration(struct processing_module *mod,
 	return ret;
 }
 
+static void module_check_source_sink(struct processing_module *mod)
```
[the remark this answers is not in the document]
RanderWang: thanks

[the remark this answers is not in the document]
RanderWang: thanks

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -162,234 +138,237 @@ static inline int smart_amp_get_config(struct processing_module *mod,
...
-	struct comp_dev *dev = mod->dev;
-
-	comp_dbg(dev, "smart_amp_reset()");
-
 	return 0;
 }
 
+static smart_amp_proc get_smart_amp_process(struct sof_sink *sink)
+{
+	switch (sink->audio_stream_params->valid_sample_fmt) {
```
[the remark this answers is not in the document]
RanderWang: sure, thanks

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -162,234 +138,237 @@ static inline int smart_amp_get_config(struct processing_module *mod,
...
+			uint32_t size)
 {
-	struct smart_amp_data *sad = module_get_private_data(mod);
-	struct audio_stream *source = bsource->data;
-	struct audio_stream *sink = bsink->data;
-	int16_t *src;
-	int16_t *dest;
-	uint32_t in_frag = 0;
-	uint32_t out_frag = 0;
+	uint32_t *src_addr = (uint32_t *)src_ptr;
```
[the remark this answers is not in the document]
RanderWang: thanks, updated

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -162,234 +138,237 @@ static inline int smart_amp_get_config(struct processing_module *mod,
...
-	struct comp_dev *dev = mod->dev;
-
-	comp_dbg(dev, "smart_amp_free()");
-	comp_data_blob_handler_free(sad->model_handler);
-	rfree(sad);
 	return 0;
 }
 
-static int smart_amp_params(struct processing_module *mod)
+static void process_s16(uint8_t *src_ptr,
```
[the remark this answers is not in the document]
RanderWang: thanks, updated

src/audio/module_adapter/module/modules.c
```diff
@@ -174,7 +174,7 @@ static int modules_prepare(struct processing_module *mod,
 		struct module_interface *mod_in =
 					(struct module_interface *)mod->priv.module_adapter;
 
-		ret = mod_in->prepare(mod, NULL, 0, NULL, 0);
+		ret = mod_in->prepare(mod, sources, num_of_sources, sinks, num_of_sinks);
```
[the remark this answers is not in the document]
RanderWang: updated

src/audio/module_adapter/module/modules.c
RanderWang: changed

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -171,160 +155,156 @@ static int smart_amp_free(struct processing_module *mod)
...
+			int8_t const *src_end,
+			int8_t *dst_ptr,
+			int8_t const *dst_begin,
+			int8_t const *dst_end,
+			uint32_t size)
 {
-	struct sof_ipc_stream_params *params = mod->stream_params;
-	struct comp_dev *dev = mod->dev;
-	int ret;
+	int16_t *src_addr = (int16_t *)src_ptr;
```
marcinszkudlinski, line 151: still, don't override const:
`int16_t const *src_addr = (int16_t const*)src_ptr;`

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -171,160 +155,156 @@ static int smart_amp_free(struct processing_module *mod)
...
+			uint32_t size)
 {
-	struct smart_amp_data *sad = module_get_private_data(mod);
-	struct audio_stream *source = bsource->data;
-	struct audio_stream *sink = bsink->data;
-	int16_t *src;
-	int16_t *dest;
-	uint32_t in_frag = 0;
-	uint32_t out_frag = 0;
+	int32_t *src_addr = (int32_t *)src_ptr;
```
marcinszkudlinski, line 176: ditto

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -171,160 +155,156 @@ static int smart_amp_free(struct processing_module *mod)
...
-		return smart_amp_process_s16;
-	case SOF_IPC_FRAME_S24_4LE:
-	case SOF_IPC_FRAME_S32_LE:
-		return smart_amp_process_s32;
-	default:
-		comp_err(dev, "smart_amp_process() error: not supported frame format");
-		return NULL;
-	}
-}
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
```
marcinszkudlinski, line 252: There's a MIN macro defined systemwide - sof/match/numbers.h

marcinszkudlinski (changes requested):
don't remap source pointers from const *, accidental modification of source buffers may result in hard to track issues (even to cache incoherency in if dpQueue is used)

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -37,57 +53,48 @@ typedef void (*smart_amp_proc)(int8_t const *src_ptr,
...
 {
 	struct smart_amp_data *sad;
-	struct comp_dev *dev = mod->dev;
 	struct module_data *mod_data = &mod->priv;
-	const size_t in_size = sizeof(struct ipc4_input_pin_format) * SMART_AMP_NUM_IN_PINS;
-	const size_t out_size = sizeof(struct ipc4_output_pin_format) * SMART_AMP_NUM_OUT_PINS;
 	int ret;
 	const struct ipc4_base_module_extended_cfg *base_cfg = mod_data->cfg.init_data;
 
-	comp_dbg(dev, "smart_amp_init()");
```
lgirdwood, line 52: lets keep all logging in the built-in code otherwise its more difficult to integrate/debug. i.e. this one can be move below to line 72.

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -96,23 +103,16 @@ static int smart_amp_set_config(struct processing_module *mod, uint32_t config_i
...
-	comp_dbg(dev, "smart_amp_set_config()");
-
 	switch (config_id) {
 	case SMART_AMP_SET_MODEL:
-		return comp_data_blob_set(sad->model_handler, pos,
-					 data_offset_size, fragment, fragment_size);
+		return 0;
 	case SMART_AMP_SET_CONFIG:
-		if (fragment_size != sizeof(sad->config)) {
-			comp_err(dev, "smart_amp_set_config(): invalid config size %u, expect %u",
```
lgirdwood, line 126: Need to keep error logs too.

lgirdwood (changes requested):
need to keep all logging otherwise we will have difficulty with debug

src/samples/audio/smart_amp_test_ipc4.c
RanderWang: sure, thanks!

src/samples/audio/smart_amp_test_ipc4.c
RanderWang: thanks, updated

src/samples/audio/smart_amp_test_ipc4.c
RanderWang: thanks, updated

src/samples/audio/smart_amp_test_ipc4.c
RanderWang: thanks ,updated

RanderWang:
SOFCI TEST

lgirdwood (approved):
@marcinszkudlinski good for you ?

lgirdwood:
@RanderWang can you check CI, some failures. Thanks

RanderWang:
> @RanderWang can you check CI, some failures. Thanks

https://github.com/thesofproject/sof/pull/8594 was failed and it should be merged first

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -346,42 +337,21 @@ static int smart_amp_prepare(struct processing_module *mod,
 	struct comp_buffer *sink_buffer;
 	struct list_item *blist;
 	int ret;
+	int i;
 
-	ret = smart_amp_params(mod);
```
lyakh, line 350: how is all the configuration now handled? Does this mean, that we're now only handling a single fixed configuration here? Are we losing any flexibility?

lgirdwood: @RanderWang can you put the #ifdef around this. Thanks.

marcinszkudlinski (approved):
Just please remove changes that have been extracted to https://github.com/thesofproject/sof/pull/8594

src/audio/module_adapter/module_adapter.c
```diff
@@ -119,39 +119,6 @@ struct comp_dev *module_adapter_new(const struct comp_driver *drv,
 	return NULL;
 }
 
-static int module_adapter_sink_src_prepare(struct comp_dev *dev)
```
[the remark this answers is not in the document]
RanderWang: sure, it is a test with 8594 and will be remove it if 8594 is merged

src/samples/audio/smart_amp_test_ipc4.c
RanderWang: The param is set in module adapter layer, so remove it, please check https://github.com/thesofproject/sof/pull/8612. Why we need smart_amp_params(mod) to support different configurations ?  User can check "struct smart_amp_data" to get current config and process it.

And for source & sink buffer with loadable module, we can't set params in loadable module since no such interface is available.  BTW why users need to set such detail params ? They want a simple framework.

src/audio/module_adapter/module/modules.c
```diff
@@ -384,7 +405,11 @@ static int modules_reset(struct processing_module *mod)
 static const struct module_interface interface = {
 	.init = modules_init,
 	.prepare = modules_prepare,
-	.process_raw_data = modules_process,
+#if CONFIG_IPC_MAJOR_4
+	.process = modules_process,
+#else
+	.process_raw_data = modules_raw_process,
+#endif
```
lyakh, line 412: wouldn't https://github.com/thesofproject/sof/pull/8180/commits/28735353cbdaa52eaa0d2ad2ac8cf4a89dec3c28 be a better solution?

src/samples/audio/smart_amp_test_ipc4.c
lyakh: do you mean, that after #8612 is merged, the configuration will be handled at the module-adapter level? Then this PR should be marked as dependent on that one, and maybe removing (then redundant) configuration setting here should be a separate commit

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -37,57 +55,50 @@ typedef void (*smart_amp_proc)(int8_t const *src_ptr,
...
 	struct smart_amp_data *sad;
-	struct comp_dev *dev = mod->dev;
 	struct module_data *mod_data = &mod->priv;
-	const size_t in_size = sizeof(struct ipc4_input_pin_format) * SMART_AMP_NUM_IN_PINS;
-	const size_t out_size = sizeof(struct ipc4_output_pin_format) * SMART_AMP_NUM_OUT_PINS;
 	int ret;
 	const struct ipc4_base_module_extended_cfg *base_cfg = mod_data->cfg.init_data;
 
-	comp_dbg(dev, "smart_amp_init()");
+	LOG_DBG("smart_amp_init()");
```
lyakh, line 73: cannot you just declare `domp_dbg()` and friends as dummies themselves for modular builds?

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -37,57 +55,50 @@ typedef void (*smart_amp_proc)(int8_t const *src_ptr,
...
+	sad = &smart_amp_priv;
+#endif
 	mod_data->private = sad;
 
-	/* component model data handler */
-	sad->model_handler = comp_data_blob_handler_new(dev);
-	if (!sad->model_handler) {
-		ret = -ENOMEM;
-		goto sad_fail;
-	}
```
lyakh, line 64: I think it was decided that the smart amp test doesn't really need and isn't using the data blob API, right? Can we make removing it a separate commit?

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -37,57 +55,50 @@ typedef void (*smart_amp_proc)(int8_t const *src_ptr,
...
+		*((struct ipc4_input_pin_format *)base_cfg->base_cfg_ext.pin_formats + 1);
+	sad->ipc4_cfg.output_pin =
+		*(struct ipc4_output_pin_format *)(base_cfg->base_cfg_ext.pin_formats
+			+ sizeof(sad->ipc4_cfg.input_pins));
 
 	return 0;
 
 sad_fail:
-	comp_data_blob_handler_free(sad->model_handler);
-	rfree(sad);
```
lyakh, line 85: you still need `rfree()` for monolithic builds, don't you?

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -146,12 +148,8 @@ static inline int smart_amp_get_config(struct processing_module *mod,
 
 static int smart_amp_free(struct processing_module *mod)
 {
-	struct smart_amp_data *sad = module_get_private_data(mod);
-	struct comp_dev *dev = mod->dev;
+	LOG_DBG("smart_amp_free()");
 
-	comp_dbg(dev, "smart_amp_free()");
-	comp_data_blob_handler_free(sad->model_handler);
-	rfree(sad);
```
lyakh, line 170: ditto

src/samples/audio/smart_amp_test_ipc4.c
```diff
@@ -358,8 +349,32 @@ static const struct module_interface smart_amp_test_interface = {
 	.free = smart_amp_free
 };
 
+#ifndef __MODULE_BUILD__
```
lyakh, line 352: please have a look at https://github.com/thesofproject/sof/pull/8180/commits/03aad36103cc93c17b1fb32f76611910a0188a22 and the first hunk in https://github.com/thesofproject/sof/pull/8180/commits/9e6dd7ec71d34de8c8fe722ba6de0b5333adafda

lyakh (commented):
We plan to enable two ways to build the smart amp test as a module: as in this PR, using the system services / LMDK approach, and as in #8180 . It should be possible to select in Kconfig which way the module is built. Some aspects of this work are common, like enabling tristate Kconfig options, and I find that some of them are better resolved in #8180. I propose to first merge at least those common commits and then merge these PRs in any order. Of course, the one, that gets merged as second, will have to be adjusted a bit

src/audio/module_adapter/module/modules.c
RanderWang: please check this https://github.com/thesofproject/sof/pull/8554#pullrequestreview-1757520083.  So I didn't use it

src/samples/audio/smart_amp_test_ipc4.c
RanderWang: No. The problem is that dev is not available for loadable module build. please check "struct processing_module".   I spent much effort to make compiling pass.

RanderWang:
Alsabat test issue in CI test. debugging

src/samples/audio/smart_amp_test_ipc4.c
lgirdwood: OK, we need to avoid common names to avoid collision with Zephyr, can we rename to ```__SOF_MODULE_SERVICE_BUILD__``` and this would indicate a SOF macro for modules using the services API. Not a blocker for this PR, can be incrementally updated.

keqiaozhang:
SOFCI TEST

src/samples/audio/smart_amp_test_ipc4.c
RanderWang: sure, thanks

src/samples/audio/smart_amp_test_ipc4.c
RanderWang: sure, this is for a demo. We still need a lot of work to include loadable module building system

RanderWang:
This PR depends on https://github.com/thesofproject/sof/pull/8594

RanderWang:
Test this PR with https://github.com/thesofproject/sof/pull/8594

src/audio/module_adapter/module/modules.c
lyakh: @RanderWang ok, we can use that work-around, but then you shouldn't need to touch modules.c ops at all - they're replaced by smart-amp's ops, aren't they?

src/samples/audio/smart_amp_test_ipc4.c
lyakh: @RanderWang sorry, don't understand. You replace `comp_dbg()` with `LOG_DBG()` and then redefine it to a NOP for loadable modules. So, why don't you just redefine the original `comp_dbg()` to a NOP? If you do, you won't care about `dev`

RanderWang: @lyakh I understand your idea. if I use comp_dbg(), it will be more ifdef in the code. So I don't prefer it
````c
#ifdef  MODULE
stuct comp_dev dev = mode->dev;
#endif
.....
````c

File: src/audio/module_adapter/module/modules.c (hunk 21 above)
Comment (RanderWang) [reply src/audio/module_adapter/module/modules.c:412]: @lyakh  this modules.c is used by loadable module which use "struct module_interface interface" as ops and in this ops to invoke loadable module's ops, right ?

Build-in smart amp doesn't use this ops.

File: src/samples/audio/smart_amp_test_ipc4.c (hunk 22 above)
Comment (lyakh) [reply src/samples/audio/smart_amp_test_ipc4.c:73]: then define `comp_dbg()` to something, that keeps `dev` "used" and prevents compiler unused variable warnings?

File: src/audio/module_adapter/module/modules.c (hunk 21 above)
Comment (lyakh) [reply src/audio/module_adapter/module/modules.c:412]: if you use the work-around, proposed by @softwarecki in https://github.com/softwarecki/sof/commit/26f4b475d30c81477ad8b4eadf29abc48ab1e31e then these ops won't be used by the smart-amp test when built as a loadable module

File: src/samples/audio/smart_amp_test_ipc4.c (hunk 22 above)
Comment (RanderWang) [reply src/samples/audio/smart_amp_test_ipc4.c:73]: this return to the first question dev is not available for loadable module

File: src/audio/module_adapter/module/modules.c (hunk 21 above)
Comment (RanderWang) [reply src/audio/module_adapter/module/modules.c:412]: @lyakh  Actually I first used this patch and we have a discussion https://github.com/thesofproject/sof/pull/8580#discussion_r1417587494

**Comment by RanderWang** [comment]
https://github.com/thesofproject/sof/pull/8594 was merged ,  ready for test

File: src/audio/module_adapter/module/modules.c (hunk 21 above)
Comment (lgirdwood) [reply src/audio/module_adapter/module/modules.c:412]: So we have 2 months until v2.9, lets refine this further as module support is still moving quickly. i.e. we can see how things work in practice with service API modules and Zephyr modules and see what solution for above supports both best.

File: src/samples/audio/smart_amp_test_ipc4.c (hunk 22 above)
Comment (lgirdwood) [reply src/samples/audio/smart_amp_test_ipc4.c:73]: This should be resolved by using native Zephyr logging, but this is WIP.

**Review by lgirdwood — APPROVED** [review]
I think we are good to go, as the landing for the initial service API module user. This will no doubt be improved and refined as we head towards v2.9, but this now get some example code that can be tested.

**Comment by lgirdwood** [comment]
@lrudyX @wszypelt not sure if you test the dummy smart amp in CI today ? if not, good to merge ?

**Comment by RanderWang** [comment]
updated.  The failure was caused by FDK library tests which uses different process function compared to sof loadable module. I have to reuse patch:  https://github.com/thesofproject/sof/pull/8580/commits/22c398c77702909f20fe58687fc28802cd128931. I checked the QT internal state , all the tests passed, but why it still show error ?

**Review by lyakh — DISMISSED** [review]
Sorry, commit number 4 in this series now looks like too intrusive to me. As it stands it seems to modify the current smart-amp testing module a lot and I don't quite understand why. As far as I understand we don't want to change this module's functionality, we just want to make it usable as a loadable module. I understand that with system services that means eliminating most calls to the base firmware. But I'd think, that this should be done by disabling that code using `#ifdef`s for this build, not by removing it, as requested e.g. in https://github.com/thesofproject/sof/pull/8580#discussion_r1423713295 . Also comparing to https://github.com/thesofproject/sof/pull/8180/commits/3163edf4f351d096ad2cab88ec38e9fb6de2d5e9#diff-37cb5e9335e5fb23ba003096bb4c98cd7cbd91d550e280bbb06a188175974e6f which only adds a hook for loading and doesn't modify any code, this really looks very intrusive.

File: src/samples/audio/smart_amp_test_ipc4.c (hunk 22 above)
Comment (lyakh) [reply src/samples/audio/smart_amp_test_ipc4.c:73]: > this return to the first question dev is not available for loadable module

I mean something like
```
#define dev_dbg(dev, ...) do { (void)dev; } while (0)
```

Comment (lyakh) [reply src/samples/audio/smart_amp_test_ipc4.c:73]: The previous one wouldn't work probably because `dev` isn't available at all as a variable. So I think a simple
```
#define dev_dbg(...) do {} while (0)
```
should do. If for some reason that doesn't work, maybe
```
#define dev_dbg(dev, ...) do { int dev; (void)dev; } while (0)
```
would.

wszypelt:
@lgirdwood We have tests for dummy smart amp but only for built-in module. after re-test all tests on green

lgirdwood:
@marcinszkudlinski @mwasko good to cherry-pick

keqiaozhang:
> @keqiaozhang are you sure about the bisect to this PR?

@plbossart I confirmed it again this morning, issue https://github.com/thesofproject/sof/issues/8641 cannot be reproduced after reverting these 2 commits: 68c63571fd6829a57814e25f5d5f73ba601a977e and 73317455bf6afb86f3aa19a4e497349bf2c18e1d.

Issue #8414, opened 2023-10-30
[BUG][LNL] DSP panic during the stress capture test


keqiaozhang:
**Describe the bug**
This issue happens on LNL-NOCODEC platforms. It runs to DSP panic after dozens of recording tests. The reproduction rate is almost 100%.

This issue cannot be reproduced in stress playback test. 

**dmesg**
```
[ 2519.683135] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-lnl 0000:00:1f.3: ipc tx done : 0x13000003|0x0: GLB_SET_PIPELINE_STATE
[ 2519.683141] kernel: snd_sof:sof_ipc4_set_pipeline_state: sof-audio-pci-intel-lnl 0000:00:1f.3: ipc4 set pipeline instance 0 state 4
[ 2519.683145] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-lnl 0000:00:1f.3: ipc tx      : 0x13000004|0x0: GLB_SET_PIPELINE_STATE
[ 2519.686193] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-lnl 0000:00:1f.3: ipc rx      : 0x1b0a0000|0x0: GLB_NOTIFICATION|EXCEPTION_CAUGHT
[ 2519.686197] kernel: sof-audio-pci-intel-lnl 0000:00:1f.3: ------------[ DSP dump start ]------------
[ 2519.686201] kernel: sof-audio-pci-intel-lnl 0000:00:1f.3: DSP panic!
[ 2519.686203] kernel: sof-audio-pci-intel-lnl 0000:00:1f.3: fw_state: SOF_FW_BOOT_COMPLETE (7)
[ 2519.686216] kernel: sof-audio-pci-intel-lnl 0000:00:1f.3: ROM status: 0x0, ROM error: 0x0
[ 2519.686218] kernel: sof-audio-pci-intel-lnl 0000:00:1f.3: ROM debug status: 0x0, ROM debug error: 0x0
[ 2519.686221] kernel: sof-audio-pci-intel-lnl 0000:00:1f.3: ROM feature bit enabled
[ 2519.686252] kernel: snd_sof:sof_ipc4_find_debug_slot_offset_by_type: sof-audio-pci-intel-lnl 0000:00:1f.3: Slot type 0x4c455400 is not available in debug window
[...]

PR #8583, opened 2023-12-06
module: sink: source: Prevent division by zero issue

jxstelter:
This patch prevents division by zero, when
source_get_frame_bytes() routine returns 0.

This change fixes issue https://github.com/thesofproject/sof/issues/8414. The problem happens due to race condition during pipeline pause operation. The exact failure mechanism require additional debug, but checking frame_bytes value before division prevents exceptions.

### DIFF
--- a/src/module/audio/source_api.c
+++ b/src/module/audio/source_api.c
@@ -53,6 +53,10 @@ size_t source_get_frame_bytes(struct sof_source *source)
 
 size_t source_get_data_frames_available(struct sof_source *source)
 {
-	return source_get_data_available(source) /
-			source_get_frame_bytes(source);
+	uint32_t frame_bytes = source_get_frame_bytes(source);
+
+	if (frame_bytes > 0)
+		return source_get_data_available(source) / frame_bytes;
+	else
+		return 0;
 }


marcinszkudlinski (approved):

src/module/audio/source_api.c
```diff
@@ -53,6 +53,10 @@ size_t source_get_frame_bytes(struct sof_source *source)
 
 size_t source_get_data_frames_available(struct sof_source *source)
 {
-	return source_get_data_available(source) /
-			source_get_frame_bytes(source);
+	uint32_t frame_bytes = source_get_frame_bytes(source);
+
+	if (frame_bytes > 0)
+		return source_get_data_available(source) / frame_bytes;
+	else
```
ranj063, line 60: else is not needed after a return. and since frame bytes is uint32_t the condition can simply be if (frame_bytes)

src/module/audio/source_api.c
```diff
@@ -53,6 +53,10 @@ size_t source_get_frame_bytes(struct sof_source *source)
 
 size_t source_get_data_frames_available(struct sof_source *source)
 {
-	return source_get_data_available(source) /
-			source_get_frame_bytes(source);
+	uint32_t frame_bytes = source_get_frame_bytes(source);
+
+	if (frame_bytes > 0)
```
lyakh, line 58: `frame_bytes` can only be 0 if the source's number of channels is 0, which does look like a race

PR #8585, opened 2023-12-06
topology2: google-rtc-aec: Add Byte control

yongzhi1:
Add binary controls for AEC tuning data with
max size set to 2048.

### DIFF
--- a/tools/topology/topology2/include/components/google-rtc-aec.conf
+++ b/tools/topology/topology2/include/components/google-rtc-aec.conf
@@ -47,6 +47,15 @@ Class.Widget."google-rtc-aec" {
 		token_ref	"comp.word"
 	}
 
+	Object.Control.bytes."1" {
+		access [
+			tlv_write
+			tlv_callback
+		]
+		name 'Config'
+		max 2048
+	}
+
 	# Attribute categories
 	attributes {
 		#


tools/topology/topology2/include/components/google-rtc-aec.conf
```diff
@@ -47,6 +47,15 @@ Class.Widget."google-rtc-aec" {
 		token_ref	"comp.word"
 	}
 
+	Object.Control.bytes."1" {
+		access [
+			tlv_write
+			tlv_callback
+		]
+		name 'Config'
```
ranj063, line 55: @yongzhi1 can the name be changed to something specific like google-rtc-config instead?

yongzhi1: The generated name is 'google-rtc-aec.18.1 Config' right now, then 'google-rtc-aec.18.1 google-rtc-config' is a little too verbose?

kv2019i (approved):

dbaluta (approved):

andyross:
FWIW there's hopefully some rework coming soon that will be moving to more fine-grained settings for enable and gain (and to get stream parameters from SOF), though the blob mechanism will likely still be required for future-proofing.

PR #8587, opened 2023-12-07
audio: crossover: crossover code refactor

btian1:


src/audio/multiband_drc/user/multiband_drc.h
```diff
@@ -9,10 +9,11 @@
 #define __USER_MULTIBAND_DRC_H__
 
 #include <stdint.h>
-#include <user/crossover.h>
 #include <user/drc.h>
 #include <user/eq.h>
 
+#include "../../crossover/crossover_user.h"
```
lyakh, line 15: not sure how strongly crossover and multiband DRC are related, but maybe it would be better to extract required common parts and keep it in headers under src/include/sof/audio?

src/audio/crossover/crossover.c
```diff
@@ -374,13 +334,11 @@ static int crossover_init(struct processing_module *mod)
 		goto cd_fail;
 	}
 
-#if CONFIG_IPC_MAJOR_4
-	ret = crossover_init_output_pins(mod);
+	ret = crossover_outputpin_init(mod);
```
lyakh, line 337: an underscore between `output_pin`

src/audio/crossover/CMakeLists.txt
```diff
@@ -1,2 +1,7 @@
 add_local_sources(sof crossover.c)
 add_local_sources(sof crossover_generic.c)
+if(CONFIG_IPC_MAJOR_3)
+	add_local_sources(sof crossover_ipc3.c)
+elseif(CONFIG_IPC_MAJOR_4)
+	add_local_sources(sof crossover_ipc4.c)
+endif()
```
lyakh, line 7: please end files with a single new line

src/audio/crossover/crossover_ipc3.c
```diff
@@ -0,0 +1,113 @@
...
+	return 0;
+}
+
+/**
+ * \brief IPC4 specific component prepare, updates source and sink buffers formats from base_cfg
+ */
+void crossover_params(struct processing_module *mod)
+{
+}
+
```
lyakh, line 113: one newline is good, two newlines are too many

src/audio/crossover/crossover_ipc3.c
```diff
@@ -0,0 +1,113 @@
...
+#include <user/trace.h>
+#include <user/eq.h>
+#include <errno.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <limits.h>
+
+#include "crossover_algorithm.h"
+#include "crossover_user.h"
+#include "crossover.h"
```
lyakh, line 39: 30+ headers for 70 lines of code... sorry, cannot believe that

src/audio/crossover/crossover_ipc4.c
```diff
@@ -0,0 +1,150 @@
...
+
+	sourceb = list_first_item(&dev->bsource_list, struct comp_buffer, sink_list);
+	ipc4_update_buffer_format(sourceb, &mod->priv.cfg.base_cfg.audio_fmt);
+
+	list_for_item(sink_list, &dev->bsink_list) {
+		sinkb = container_of(sink_list, struct comp_buffer, source_list);
+		ipc4_update_buffer_format(sinkb, &mod->priv.cfg.base_cfg.audio_fmt);
+	}
+}
+
```
lyakh, line 150: one newline

src/audio/crossover/crossover_ipc4.c
```diff
@@ -0,0 +1,150 @@
...
+#include <user/trace.h>
+#include <user/eq.h>
+#include <errno.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <limits.h>
+
+#include "crossover_algorithm.h"
+#include "crossover_user.h"
+#include "crossover.h"
```
lyakh, line 39: looks like too many headers. Just to expand a bit: getting headers perfectly isn't always simple, and probably we don't expect that. But please try to have them mostly right - include all the headers that are needed *in this file* and avoid any headers, that aren't needed here

src/audio/crossover/crossover_ipc3.c
btian1: forget to remove this part, sorry.

src/audio/multiband_drc/user/multiband_drc.h
btian1: for #include <user/drc.h>

it also need referenced by: ../../drc/drc_user.h, the structure is used from drc_user.h, combined with above, if put common to one header, seems also strange, image a header with crossover definition and drc structure definition, we may consult Seppo once he is back.

marc-hb:
Never a dull moment... this PR confuses the disassembler. For the first time since forever, it introduces a disassembly difference between Linux and Windows - with the same toolchain of course.

Note I have seen other issues before with parent `#include "../..."`, this looks suspicious.

https://github.com/thesofproject/sof/actions/runs/7129022432/job/19412552108?pr=8587

```diff
--- lin/build-sof-staging/sof-info/mtl/zephyr.lst	2023-12-07 13:46:40
+++ win/build-sof-staging/sof-info/mtl/zephyr.lst	2023-12-07 13:49:28
@@ -138437,58 +138437,47 @@
 	return 0;
 a006c11c:	020c                          	movi.n	a2, 0
 }
 a006c11e:	f01d                          	retw.n
 
 a006c120 <multiband_drc_reset_state>:
 {
 a006c120:	004136                      	entry	a1, 32
 	for (i = 0; i < PLATFORM_MAX_CHANNELS; i++)
 a006c123:	88a052                      	movi	a5, 136
 a006c126:	328b                          	addi.n	a3, a2, 8
 a006c128:	525a                          	add.n	a5, a2, a5
-static inline void crossover_reset_state_lr4(struct iir_state_df2t *lr4)
-{
-	rfree(lr4->coef);
-	rfree(lr4->delay);
-
 	lr4->coef = NULL;
 a006c12a:	040c                          	movi.n	a4, 0
 	rfree(lr4->coef);
 a006c12c:	0023a2                      	l32i	a10, a3, 0
 a006c12f:	e1a425                      	call8	a004db70 <rfree>
 	rfree(lr4->delay);
 a006c132:	0123a2                      	l32i	a10, a3, 4
 a006c135:	e1a3a5                      	call8	a004db70 <rfree>
 	lr4->coef = NULL;
 a006c138:	0349                          	s32i.n	a4, a3, 0
 	lr4->delay = NULL;
 a006c13a:	1349                          	s32i.n	a4, a3, 4
 a006c13c:	10c332                      	addi	a3, a3, 16
 a006c13f:	e99357                      	bne	a3, a5, a006c12c <multiband_drc_reset_state+0xc>
 a006c142:	30a362                      	movi	a6, 0x330
 a006c145:	30c252                      	addi	a5, a2, 48
 a006c148:	626a                          	add.n	a6, a2, a6
 	lr4->coef = NULL;
 a006c14a:	040c                          	movi.n	a4, 0
 a006c14c:	000c06                      	j	a006c180 <multiband_drc_reset_state+0x60>
 a006c14f:	00                               	.byte 00
 	rfree(lr4->coef);
 a006c150:	2223a2                      	l32i	a10, a3, 136
- */
-static inline void crossover_reset_state_ch(struct crossover_state *ch_state)
-{
-	int i;
-
-	for (i = 0; i < CROSSOVER_MAX_LR4; i++) {
 a006c153:	10c332                      	addi	a3, a3, 16
 a006c156:	e1a1a5                      	call8	a004db70 <rfree>
 	rfree(lr4->delay);
 a006c159:	1f23a2                      	l32i	a10, a3, 124
 a006c15c:	e1a125                      	call8	a004db70 <rfree>
 	rfree(lr4->coef);
 a006c15f:	2a23a2                      	l32i	a10, a3, 168
 	lr4->coef = NULL;
 a006c162:	1e6342                      	s32i	a4, a3, 120
 	lr4->delay = NULL;
 a006c165:	1f6342                      	s32i	a4, a3, 124
 	rfree(lr4->coef);

```

lgirdwood (approved):

btian1:
SOFCI TEST

src/audio/multiband_drc/multiband_drc.h
```diff
@@ -8,14 +8,14 @@
...
 #include <sof/audio/module_adapter/module/generic.h>
-#include <sof/audio/crossover/crossover.h>
 #include <sof/audio/drc/drc.h>
 #include <sof/math/iir_df2t.h>
 #include <sof/audio/component.h>
 #include <sof/audio/data_blob.h>
 #include <sof/platform.h>
 #include <stdint.h>
 #include "user/multiband_drc.h"
+#include "../crossover/crossover.h"
```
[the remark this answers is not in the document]
btian1: @marc-hb , how about add one more path to zephyr/CMakeList with below diff:
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -90,6 +90,7 @@ ExternalProject_Add(sof_logger_ep
 # default SOF includes
 target_include_directories(SOF INTERFACE ${RIMAGE_TOP}/src/include)
 target_include_directories(SOF INTERFACE ${SOF_SRC_PATH}/include)
+target_include_directories(SOF INTERFACE ${SOF_SRC_PATH}/audio)
 target_include_directories(SOF INTERFACE ${SOF_SRC_PATH}/arch/${ARCH}/include)
 target_include_directories(SOF INTERFACE ${sof_top_dir}/third_party/include)


then all these can be changed with <crossover/crossover.h> ?

lyakh (approved):
headers look roughly reasonable now, thanks!

src/audio/multiband_drc/multiband_drc.h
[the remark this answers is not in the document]
kv2019i: @marc-hb we do! I think ".." is better in this case to make it very clear this module is reusing definitions of a peer module (see my comments to https://github.com/thesofproject/sof/pull/8593 ).

src/audio/crossover/crossover.c
```diff
@@ -46,43 +46,11 @@ DECLARE_SOF_RT_UUID("crossover", crossover_uuid, 0x948c9ad1, 0x806a, 0x4131,
```
kv2019i, line 46: s/de-cluster/declutter/

kv2019i (commented):
Code looks good, but we need to discuss and reach a consensus how we handle cross-module dependencies (and their headers). See inline and my comments to DRC refactor PR.

src/audio/multiband_drc/multiband_drc.h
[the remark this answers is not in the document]
marc-hb: >  how about add one more path to zephyr/CMakeList with below diff:

(I edited your diff, edit it again to see how)

Yes please, something like that but probably with some `CMakeLists.txt:if(CONFIG_...)`. It would be interesting to check whether that solves the disassembly difference between Linux and Windows. That can be done with test draft PR (no reviewers / no github spam)



No no such disassembly difference in https://github.com/thesofproject/sof/actions/runs/7137933378/job/19438902490?pr=8593 even though it has one `#include "../"`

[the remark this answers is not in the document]
marc-hb: > to make it very clear this module is reusing definitions of a peer module

There is no requirement for all `target_include_directories()` to be in the top-level makefile.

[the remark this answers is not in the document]
btian1: @kv2019i @marc-hb , yes we need a conclusion for the cross module header usage:
1. use "../" or "../../"
2. add new target_include_directories with +target_include_directories(SOF INTERFACE ${SOF_SRC_PATH}/audio)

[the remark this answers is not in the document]
singalsu: @johnylin76 @btian1 I have in my improvement ideas list move of crossover filter bank to math library to avoid the tie between DRC and crossover. Need to do it latest when we next need other recursive filter bank. So this include would become include from math headers.

@btian1 But I don't think it fits this PR scope, should be done after since it may need small changes to algorithms code.  If you can test our option 2. it would be nice.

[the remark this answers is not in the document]
kv2019i: @marc-hb @btian1 @singalsu To me, adding a non-include folder into target_include_directories is opening a can of worms. Then the build is searching for headers in the implementation parts of the code base and this can lead to surprises. I personally prefer the explicit "../../foo.h" includes, to make it very explicit something unusual is happening here, and only specific foo.h is included. This really only affects modules that are directly reusing portions of other peer modules.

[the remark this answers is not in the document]
singalsu: Thanks for thoughts @kv2019i ! Makes sense. I hope we can soon clean up this dependency into maths library. It's been causing difficulty with testbench build (shared libraries time).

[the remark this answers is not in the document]
marc-hb: > I think ".." is better in this case to make it very clear this module is reusing definitions of a peer module

I tested this and it Just Works:
```
zephyr_library_include_directories_ifdef(CONFIG_COMP_MULTIBAND_DRC
  ${SOF_AUDIO_PATH}/crossover
)
```

This expresses the multiband->crossover dependency (or any other similar dependency) very clearly and in a regular, "CMake-native" way.


> To me, adding a non-include folder into target_include_directories is opening a can of worms. 
> Then the build is searching for headers in the implementation parts of the code base and this can lead to surprises.

I don't see what could go wrong: there are very few .h files in the .c directories after all.

In fact the "real" question is: why is the .h file of an _external module API_ located in the same directory as some .c files? This here is really the root-cause/of all evil.

If a "peer module" wants to be re-usable then I'm sure it can afford some `include/` subdirectory, that's really not vey demanding and it clarifies what .h is external versus internal. Then:

```
zephyr_library_include_directories_ifdef(CONFIG_COMP_MULTIBAND_DRC
  ${SOF_AUDIO_PATH}/crossover/include
)
```

and job done.

marc-hb (changes requested):
Blocking merge until the build is fixed.

>  but we need to discuss and reach a consensus how we handle cross-module dependencies (and their headers). See inline and my comments to DRC refactor PR.

I don't have a strong opinion but I recommended a couple, simple solutions that don't require any `#include "../foo.h"` and that look like a more "normal" dependency in CMake.

I do have a much stronger opinion on partially breaking the disassembly on Windows, see example above. Reminder in case anyone didn't notice: this PR "breaks the build"!

I have a very strong suspicion that something somewhere in the toolchain when it runs on Windows does not like the unusual `#include "../foo.h"`, probably because of some backslash-forward slash issue. Something similar already happened in ed7c57ba32000a. Indeed, if you look at some Windows build log you will find a mix of forward slashes and backslashes: https://github.com/thesofproject/sof/actions/runs/7176152352/job/19540597509?pr=8587

btian1:
> Blocking merge until the build is fixed.
> 
> Also, a Googler's name is all over the `git blame` yet no one from Google has reviewed yet. According to @cujomalainey in #8593, @johnylin76 should be back soon.
> 
> > but we need to discuss and reach a consensus how we handle cross-module dependencies (and their headers). See inline and my comments to DRC refactor PR.
> 
> I don't have a strong opinion about `#include` but I recommended a couple, simple solutions that don't require any `#include "../foo.h"` and that look like a more "normal" dependency in CMake.
> 
> I do have a much stronger opinion on partially breaking the disassembly on Windows, see example above. Reminder in case anyone didn't notice: this PR "breaks the build"!
> 
> I have a very strong suspicion that something somewhere in the toolchain when it runs on Windows does not like the unusual `#include "../foo.h"`, probably because of some backslash-forward slash issue. Something similar already happened in [ed7c57b](https://github.com/thesofproject/sof/commit/ed7c57ba32000a73f89a74ca1bacc4d4a861b087). Indeed, if you look at some Windows build log you will find a mix of forward slashes and backslashes: https://github.com/thesofproject/sof/actions/runs/7176152352/job/19540597509?pr=8587

@marc-hb ， I also ahve concern on this error difference, @kv2019i , I can add one more patch: move out header file from user folder, and rename it with _user.h, then include will only have "../" instead of "../../", this is a intermediate solution, do you agree on this?

singalsu:
@marc-hb The dependency from multiband-DRC to crossover will be cleaned up by moving the common parts to maths library (IIR filterbank). But it can't be done in this PR.

keqiaozhang:
SOFCI TEST

src/audio/crossover/crossover.c
```diff
@@ -30,13 +28,16 @@
...
 #include <user/trace.h>
-#include <user/crossover.h>
 #include <user/eq.h>
 #include <errno.h>
 #include <stddef.h>
 #include <stdint.h>
 #include <limits.h>
 
+#include "crossover_algorithm.h"
+#include "crossover_user.h"
```
cujomalainey, line 38: is there a doc/comment/anything explaining the new expected layout of the headers?

kv2019i: @cujomalainey  I was looking for same here -> please see  https://github.com/thesofproject/sof/pull/8557#discussion_r1414487762 explanation from @lgirdwood 

Let me file an enhancement retrospectively -- these cross-module usages require alignment for sure.

cujomalainey: Ok what about how modules that depend on each other, is the plan to just to reach into their private files?

marc-hb:
The disassembly is still different in https://github.com/thesofproject/sof/actions/runs/7207650079/job/19635312145?pr=8587

btian1:
> The disassembly is still different in https://github.com/thesofproject/sof/actions/runs/7207650079/job/19635312145?pr=8587

yes, I think we should first resolve the module dependency before review and merge this and DRC PR.
@singalsu @kv2019i 

I will submit patch next week try to resolve the module dependency.

src/audio/crossover/crossover.c
btian1: @cujomalainey ，this is the GH link: https://github.com/thesofproject/sof/issues/8630
we may need discuss how to implement modules that depend on each other.

marc-hb:
Not sure what happened but Windows and Linux are now the same in https://github.com/thesofproject/sof/actions/runs/7258301081/job/19773565761?pr=8587

lgirdwood:
@btian1 can you fix the conflicts and make non draft. thanks.

btian1:
closing this, due to there are much conflicts than expected, I create new PR for continue review:
https://github.com/thesofproject/sof/pull/8652

PR #8593, opened 2023-12-08
DRC code refactor

btian1:


lgirdwood (approved):

src/audio/multiband_drc/multiband_drc.c
```diff
@@ -32,6 +31,8 @@
 #include <errno.h>
 #include <stddef.h>
 #include <stdint.h>
+
+#include "../drc/drc_algorithm.h"
```
lyakh, line 35: as @marc-hb also has pointed out elsewhere, such includes aren't very pretty, would be better to keep shared between modules headers in an "include" directory

kv2019i: This isn't a fully clearcut case either. Here multiband_drc is directly building on top of drc (and also uses crossover).

I think we should also ensure this type of cross-module resuse is easy to do and code stays maintainable. 

@lgirdwood Any input, should we keep such reused headers (like drc_algorithm.h) in common code?

I think I'd personally still code with the approach in this PR. The modules are deeply dependent on each other, so having a "../drc/drc_algorithm.h" makes sense. Having alone drc_algorithm.h in sof/audio/ seems harder to understand ("why is this header here but not others?").

src/audio/drc/drc.c
```diff
@@ -44,7 +44,7 @@ DECLARE_SOF_RT_UUID("drc", drc_uuid, 0xb36ee4da, 0x006f, 0x47f9,
```
kv2019i, line 44: spelling in commit: "in-compatible->incompatible"

kv2019i (commented):
The shared headers questions warrant a bit more discussion, see inline. Feedback from @johnylin76 as module owner also very welcome.

So the background is to make module implementation more standalone, with the module directory having all essential bits (code, headers, docs, Kconfig and possible meta data like a toml file in the future). As DRC is used as a basic for Multiband-DRC, this doesn't fit the goal as nicely -- but I think sharing module code is even more important, so we shouldn't go overboard trying to fit everything to the same one-directory model.

src/audio/multiband_drc/multiband_drc.c
lyakh: I don't have a huge problem with this approach, just looking for the best one. As alternatives we could either
1. have an include directory under src/audio/include for such shared headers or
2. add src/audio to the list of include paths

I think I'd find (2) preferable

cujomalainey:
Please wait till @johnylin76 is back to review this as he is the original author of this code.

src/audio/multiband_drc/multiband_drc.c
btian1: Seppo suggested move the common part to math folder in a separate PR, I will do it once those PRs get approval and merged.

johnylin76 (approved):
LGTM, thanks.

Multiband-DRC is a compound arithmetic model that consists of DRC, EQ, and CROSSOVER. (EQ is referenced from src/math/ which shouldn't bet a problem.)
I have some follow-up questions. Any advice will be appreciated.
1. Do we plan the same code refactoring as DRC for CROSSOVER?
2. What would be the good practice for such compound model for future reference?
    - Is it proper to move the shared function, i.e. (drc|crossover)_algorithm, to src/math?
    - I used to consider to assemble them in topology level, but soon after then I changed my mind due to the technical difficulties for [structure](https://github.com/thesofproject/sof/blob/main/src/audio/multiband_drc/multiband_drc_generic.c#L192) in topology

src/audio/multiband_drc/multiband_drc.c
marc-hb: > The modules are deeply dependent on each other, so having a "../drc/drc_algorithm.h" makes sense. 

I measured how frequent are these "dot-dot" includes across all the default `zephyr/west.yml` repos. It's about 1.5%. So it's not "exotic" but it's definitely not "mainstream" either. As I just mentioned in #8630, there is `zephyr_library_include_directories()` and friends.

```
$ west grep '^[[:blank:]]*#[[:blank:]]*include[[:blank:]]*[<"]\.\.' | wc -l
2914
$ west grep '^[[:blank:]]*#[[:blank:]]*include[[:blank:]]*[<"][[:alnum:]]' | wc -l
228292
```

marc-hb:
> What would be the good practice for such compound model for future reference?

- #8630

src/audio/multiband_drc/multiband_drc.c
btian1: how  about non-zephyr cases? i.e: CONFIG_ZEPHYR_NATIVE_DRIVERS is not defined, will other build also support zephyr syntax?

btian1:
> LGTM, thanks.
> 
> Multiband-DRC is a compound arithmetic model that consists of DRC, EQ, and CROSSOVER. (EQ is referenced from src/math/ which shouldn't bet a problem.) I have some follow-up questions. Any advice will be appreciated.
> 
> 1. Do we plan the same code refactoring as DRC for CROSSOVER?
    Yes, see: #8587
> 2. What would be the good practice for such compound model for future reference?
>    
>    * Is it proper to move the shared function, i.e. (drc|crossover)_algorithm, to src/math?
>    * I used to consider to assemble them in topology level, but soon after then I changed my mind due to the technical difficulties for [structure](https://github.com/thesofproject/sof/blob/main/src/audio/multiband_drc/multiband_drc_generic.c#L192) in topology
yes, move to src/math is one way for this cross module case, another way may need add zephyr specific inlcude, see：
https://github.com/thesofproject/sof/issues/8630

lgirdwood (approved):

PR #8594, opened 2023-12-08
module: prepare sink & source in bind & unbind function

RanderWang:
Module will update source & sink information when bind & unbind event happen.

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -119,39 +119,6 @@ err:
 	return NULL;
 }
 
-static int module_adapter_sink_src_prepare(struct comp_dev *dev)
-{
-	struct processing_module *mod = comp_get_drvdata(dev);
-	struct list_item *blist;
-	int ret;
-	int i;
-
-	/* acquire all sink and source buffers, get handlers to sink/source API */
-	i = 0;
-	list_for_item(blist, &dev->bsink_list) {
-		struct comp_buffer *sink_buffer =
-				container_of(blist, struct comp_buffer, source_list);
-		mod->sinks[i] = audio_stream_get_sink(&sink_buffer->stream);
-		i++;
-	}
-	mod->num_of_sinks = i;
-
-	i = 0;
-	list_for_item(blist, &dev->bsource_list) {
-		struct comp_buffer *source_buffer =
-				container_of(blist, struct comp_buffer, sink_list);
-
-		mod->sources[i] = audio_stream_get_source(&source_buffer->stream);
-		i++;
-	}
-	mod->num_of_sources = i;
-
-	/* Prepare module */
-	ret = module_prepare(mod, mod->sources, mod->num_of_sources, mod->sinks, mod->num_of_sinks);
-
-	return ret;
-}
-
 #if CONFIG_ZEPHYR_DP_SCHEDULER
 static int module_adapter_dp_queue_prepare(struct comp_dev *dev)
 {
--- a/src/include/sof/audio/module_adapter/module/modules.h
+++ b/src/include/sof/audio/module_adapter/module/modules.h
@@ -66,6 +66,8 @@ static inline void declare_dynamic_module_adapter(struct comp_driver *drv,
 	drv->ops.set_large_config = module_set_large_config;
 	drv->ops.get_large_config = module_get_large_config;
 	drv->ops.get_attribute = module_adapter_get_attribute;
+	drv->ops.bind = module_adapter_bind;
+	drv->ops.unbind = module_adapter_unbind;
 }
 
 #endif /* __SOF_AUDIO_MODULES__ */
--- a/src/audio/module_adapter/module_adapter_ipc3.c
+++ b/src/audio/module_adapter/module_adapter_ipc3.c
@@ -287,3 +287,36 @@ int module_adapter_cmd(struct comp_dev *dev, int cmd, void *data, int max_data_s
 	comp_dbg(dev, "module_adapter_cmd() done");
 	return ret;
 }
+
+int module_adapter_sink_src_prepare(struct comp_dev *dev)
+{
+	struct processing_module *mod = comp_get_drvdata(dev);
+	struct list_item *blist;
+	int ret;
+	int i;
+
+	/* acquire all sink and source buffers, get handlers to sink/source API */
+	i = 0;
+	list_for_item(blist, &dev->bsink_list) {
+		struct comp_buffer *sink_buffer =
+				container_of(blist, struct comp_buffer, source_list);
+		mod->sinks[i] = audio_stream_get_sink(&sink_buffer->stream);
+		i++;
+	}
+	mod->num_of_sinks = i;
+
+	i = 0;
+	list_for_item(blist, &dev->bsource_list) {
+		struct comp_buffer *source_buffer =
+				container_of(blist, struct comp_buffer, sink_list);
+
+		mod->sources[i] = audio_stream_get_source(&source_buffer->stream);
+		i++;
+	}
+	mod->num_of_sources = i;
+
+	/* Prepare module */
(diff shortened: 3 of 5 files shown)


marcinszkudlinski (changes requested):
Don't duplicate identical functionality - pointers to sink and sources are set twice, using identical code

remove ``module_adapter_sink_src_prepare`` from module adapter - consider applying https://github.com/thesofproject/sof/commit/d0bf41b6b3df45c40dfcd30a7329e284250c3346 together with this patch

RanderWang:
> Don't duplicate identical functionality - pointers to sink and sources are set twice, using identical code
> 
> remove `module_adapter_sink_src_prepare` from module adapter - consider applying [d0bf41b](https://github.com/thesofproject/sof/commit/d0bf41b6b3df45c40dfcd30a7329e284250c3346) together with this patch

@marcinszkudlinski  Thanks, integrate your suggestion.

marcinszkudlinski (approved):

kv2019i (approved):

src/audio/module_adapter/module_adapter.c
```diff
@@ -198,7 +165,7 @@ static int module_adapter_dp_queue_prepare(struct comp_dev *dev)
 			goto err;
 		dp_queue_append_to_list(dp_queue, &mod->dp_queue_ll_to_dp_list);
 
-		/* it will override source pointers set by module_adapter_sink_src_prepare
+		/* it will override source pointers set before
```
btian1, line 168: in line 180 and 222, mod->num_of_sources and mod->num_of_sinks also re-calculated, is it possible to keep one?

RanderWang: It is for dp_queue. I am not sure about this. @marcinszkudlinski  is working on dp queue.

src/include/sof/audio/module_adapter/module/modules.h
```diff
@@ -66,6 +66,8 @@ static inline void declare_dynamic_module_adapter(struct comp_driver *drv,
 	drv->ops.set_large_config = module_set_large_config;
 	drv->ops.get_large_config = module_get_large_config;
 	drv->ops.get_attribute = module_adapter_get_attribute;
+	drv->ops.bind = module_adapter_bind;
+	drv->ops.unbind = module_adapter_unbind;
```
lyakh, line 70: just for clarity: this change only affects loadable modules, for built-in modules `.bind` and `.unbind` are already set in generic.h

lyakh (approved):

src/audio/module_adapter/module_adapter.c
marcinszkudlinski: Yes, it should be changed - creating of "shadow dpQueues" should also be moved to bind/unbind. 
It is not hurting DP processing at the moment and I don't want to delay this PR.  - so I approved the PR and will change it in separate PR

Target solution is Pipeline 2.0 where sink/src will be a main data exchanging API

kv2019i:
This looks good to go otherwise, but the testbench failure needs to be fixed. If it's a problem in testbench, we can have a fix in separate PR, but we need a rootcause before we can merge this.

singalsu:
The fail with testbench (and likely all IPC3) is because num_of_sources and num_of_sinks is zero when calling src_prepare().

Screenshot from 2023-12-11 16-25-18

singalsu:
@RanderWang Does the SRC work in a normal IPC4 device with this PR? I'm not sure we have in CI test that runs a SRC pipeline. Looks that mod->num_of_sources and mod->num_of_sinks() are set after calling module_prepare() in module_adapter_dp_queue_prepare().

For testbench in module_adapter_prepare() the first if true is not taken because mod->dev->ipc_config.proc_domain is not COMP_PROCESSING_DOMAIN_DP. Then in the second else if the num_of_ arguments are zeros.

```
	if (IS_PROCESSING_MODE_SINK_SOURCE(mod) &&
	    mod->dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_DP)
		ret = module_adapter_dp_queue_prepare(dev);

	else if (IS_PROCESSING_MODE_SINK_SOURCE(mod) &&
		 mod->dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_LL)
		ret = module_prepare(mod, mod->sources, mod->num_of_sources,
				     mod->sinks, mod->num_of_sinks);

```

src/include/sof/audio/module_adapter/module/modules.h
RanderWang: sure

RanderWang:
> @RanderWang Does the SRC work in a normal IPC4 device with this PR? I'm not sure we have in CI test that runs a SRC pipeline. Looks that mod->num_of_sources and mod->num_of_sinks() are set after calling module_prepare() in module_adapter_dp_queue_prepare().
> 
> For testbench in module_adapter_prepare() the first if true is not taken because mod->dev->ipc_config.proc_domain is not COMP_PROCESSING_DOMAIN_DP. Then in the second else if the num_of_ arguments are zeros.
> 
> ```
> 	if (IS_PROCESSING_MODE_SINK_SOURCE(mod) &&
> 	    mod->dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_DP)
> 		ret = module_adapter_dp_queue_prepare(dev);
> 
> 	else if (IS_PROCESSING_MODE_SINK_SOURCE(mod) &&
> 		 mod->dev->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_LL)
> 		ret = module_prepare(mod, mod->sources, mod->num_of_sources,
> 				     mod->sinks, mod->num_of_sinks);
> ```

I tested SRC. but it is working LL mode not in DP mode

RanderWang:
I also tested SRC in DP mode, but no issue found

btian1 (approved):

RanderWang:
Even I simply this PR to one which just updates source & sink info in bind &unbind function but mixin & mixout will be failed.   Actually sink & source are not used by mixin & mixout.  .......

abonislawski (approved):

singalsu:
> I also tested SRC in DP mode, but no issue found

Thanks, then it looks like this is only issue for IPC3 or testbench. Can you check if adding for IPC3 build some of the removed num_of calculations fixes the test issue? In the above code snippet, for the else if call before calling module_prepare().

RanderWang:
> > I also tested SRC in DP mode, but no issue found
> 
> Thanks, then it looks like this is only issue for IPC3 or testbench. Can you check if adding for IPC3 build some of the removed num_of calculations fixes the test issue? In the above code snippet, for the else if call before calling module_prepare().

thanks, I got the same idea. IPC3 doesnt has bind & unbind

kv2019i:
Looks better now, let's wait for the mandatory CI runs to complete and if all good, this is ready to be merged.

kv2019i:
CI almost clean. I do see one fail here:
https://sof-ci.01.org/sofpr/PR8594/build1044/devicetest/index.html?model=TGLU_RVP_SDW-ipc4&testcase=check-signal-stop-start-playback-10

Looks like a case of https://github.com/thesofproject/sof/issues/8056 but triggered in a different test case.

kv2019i:
@RanderWang Can you check the above fail? I browsed through history of PR tests and I don't see this failing for any other PR.

RanderWang:
> @RanderWang Can you check the above fail? I browsed through history of PR tests and I don't see this failing for any other PR.

@kv2019i  it was a old bug hard to be reproduced. Let run it again.  
````c
[   53.980498] <err> dai_comp: dai_common_new: comp:1 0x10004 dai_new(): dai_get() failed to create DAI.
[   53.980518] <err> copier: copier_dai_create: comp:1 0x10004 failed to create dai
````

RanderWang:
SOFCI TEST

RanderWang:
@kv2019i  pass. It is unlucky that it can't be reproduced again.

PR #8595, opened 2023-12-08
MUX code refactor

btian1:
This is a clean up, purpose is de-cluster headers, toml files, Readme.md etc per module basis, since today everything is scattered in current code base.


src/audio/CMakeLists.txt
```diff
@@ -191,6 +192,7 @@ elseif(CONFIG_IPC_MAJOR_4)
 	set(tdfb_sources tdfb/tdfb_ipc4.c)
 	set(tdfb_sources multiband_drc/multiband_drc_ipc4.c)
 	set(dcblock_sources dcblock/dcblock_ipc4.c)
+	set(mux_sources mux/mux_ipc4.c)
```
kv2019i, line 195: Hmm, I noted about this is another PR (I think @andrula-song  did that), but why do we need to enumerate sources twice, once in src/audio/CMakeLists.txt and then in src/audio/mux/CMakeLists.txt ? FYI to @marc-hb  If we want to clean up, shouldn't this be in one place (or two with Zephyr until Marc unifies this between XTOS/Zephyr). Now we seem to have three definitions. Can we fix this?

kv2019i (commented):
One question about the CMakeFiles.txt definitions. Shouldn't we define the sources only twice? Otoh, this duplication was before this PR, so this should be done separately .

lgirdwood (commented):
Just one comment otherwise LGTM

src/audio/CMakeLists.txt
btian1: yes, we can remove it in a separate PR, current change is based on "grep" result, add changes to each mux.c place.

marc-hb: Don't search/replace code without testing it. Compilation does not take much time.

As always the first step is to break it. Insert garbage in that spot, clean and re-compile. If it does not break then it means you're not testing it. If you're not testing it then don't touch it.

btian1: 1. local does not have all the build.
2. I suggested in other PR, we will use a separate PR to converge src/audio/CMakeList.txt and src/audio/module/CMakeList.txt

marc-hb: > local does not have all the build.

I don't understand, can you rephrase?

marc-hb: > why do we need to enumerate sources twice, once in src/audio/CMakeLists.txt and then in src/audio/mux/CMakeLists.txt ? 

Looks like @btian1 didn't know (I'm afraid he's submitting uncompiled search/replace) so I had a look at this myself. The answer is: `src/audio/CMakeLists.txt` code after line 119 is only for `CONFIG_COMP_MODULE_SHARED_LIBRARY_BUILD`. There is a very confusing `return()` on that line which makes the code after it not indented. So this is one of the explanation for the mass copy/paste/diverge.

Good thing I just added a Github Action for the ALSA plugin. It may still be incorrect but at least it will still compile?

This part of the CMake code is significantly more messy that the rest, it's going to be a huge effort to clean all that technical debt (if ever)

lyakh (approved):

kv2019i (approved):

PR #8598, opened 2023-12-08
scripts/xtensa-build-zephyr.py: Correct XTENSA_SYSTEM path

andyross:
When XTENSA_TOOLS_ROOT is set in the environment (which is a standard way for the user to specify the SDK installation directory), this script was setting the XTENSA_SYSTEM variable (the location of the registry of cores in a specific toolchain) to:

    $XTENSA_TOOLS_ROOT/install/builds/<toolchain_ver>/

But that seems to be incorrect, and causes the compiler to fail with a fairly reasonable runtime error that it can't find the requested core. (And THAT then causes the initial Zephyr cmake compiler check to fail in weird ways, as it tries to run "xt-clang" with no arguments and eats the output, reporting that you have no compiler installed).

The correct $XTENSA_SYSTEM path that works is:

    $XTENSA_TOOLS_ROOT/install/tools/<toolchain_ver>/XtensaTools/config

I'm guessing most builds run the tools from standard locations and don't set XTENSA_TOOLS_ROOT?

### DIFF
--- a/scripts/xtensa-build-zephyr.py
+++ b/scripts/xtensa-build-zephyr.py
@@ -699,9 +699,12 @@ def build_platforms():
 			# Set variables expected by xcc toolchain. CMake cannot set (evil) build-time
 			# environment variables at configure time:
 			# https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#how-can-i-get-or-set-environment-variables
-			XTENSA_BUILDS_DIR=str(pathlib.Path(xtensa_tools_root_dir, "install", "builds",
-				TOOLCHAIN_VER).absolute())
-			XTENSA_SYSTEM = str(pathlib.Path(XTENSA_BUILDS_DIR, XTENSA_CORE, "config").absolute())
+			XTENSA_SYSTEM = str(pathlib.Path(xtensa_tools_root_dir,
+							 "install",
+							 "tools",
+							 TOOLCHAIN_VER,
+							 "XtensaTools",
+							 "config").absolute())
 			platf_build_environ["XTENSA_SYSTEM"] = XTENSA_SYSTEM
 
 		platform_build_dir_name = f"build-{platform}"


andyross:
Been sitting on this for a while and figure I should get it submitted.  I doubt many people are using XTENSA_TOOLS_ROOT as an input variable, but I find I end up with toolchains in funny places.

marc-hb (changes requested):
I'm afraid this was done on purpose to avoid confusion when installing multiple toolchains in the same place. Blocking this PR conservatively.

The `tools/` directory has a `default-params` symbolic link and concept that we don't want:
```
ls -l tools/RG-2017.8-linux/*/config/*params | cut -d' ' -f5-
14592 Feb 18  2023 tools/RG-2017.8-linux/XtensaTools/config/cavs2x_LX6HiFi3_2017_8-params
   29 Apr 16  2021 tools/RG-2017.8-linux/XtensaTools/config/default-params -> cavs2x_LX6HiFi3_2017_8-params
14431 Feb 18  2023 tools/RG-2017.8-linux/XtensaTools/config/LX4_langwell_audio_17_8-params
14627 Feb 18  2023 tools/RG-2017.8-linux/XtensaTools/config/X4H3I16w2D48w3a_2017_8-params
14520 Feb 18  2023 tools/RG-2017.8-linux/XtensaTools/config/X6H3CNL_2017_8-params
```
A default toolchain may be useful in interactive use but it's not OK in automation where we never want to have any such sort of silent "fallback". Automation must specify what it wants, otherwise the build should fail. Recent example:
- https://github.com/zephyrproject-rtos/zephyr/pull/65898

Compare with the `builds/` directory which is unambiguous:

```
ls -l builds/RG-2017.8-linux/*/config/*params | cut -d' ' -f5-
14592 Feb 18  2023 builds/RG-2017.8-linux/cavs2x_LX6HiFi3_2017_8/config/cavs2x_LX6HiFi3_2017_8-params
14592 Feb 18  2023 builds/RG-2017.8-linux/cavs2x_LX6HiFi3_2017_8/config/default-params
14431 Feb 18  2023 builds/RG-2017.8-linux/LX4_langwell_audio_17_8/config/default-params
14431 Feb 18  2023 builds/RG-2017.8-linux/LX4_langwell_audio_17_8/config/LX4_langwell_audio_17_8-params
14627 Feb 18  2023 builds/RG-2017.8-linux/X4H3I16w2D48w3a_2017_8/config/default-params
14627 Feb 18  2023 builds/RG-2017.8-linux/X4H3I16w2D48w3a_2017_8/config/X4H3I16w2D48w3a_2017_8-params
14520 Feb 18  2023 builds/RG-2017.8-linux/X6H3CNL_2017_8/config/default-params
14520 Feb 18  2023 builds/RG-2017.8-linux/X6H3CNL_2017_8/config/X6H3CNL_2017_8-params
```

> But that seems to be incorrect, and causes the compiler to fail with a fairly reasonable runtime error that it can't find the requested core

This setup goes many years back to the non-Zephyr, XTOS script (which is still in use outside Intel):  9c9c4a4b172b9c, 6468490d8d381. I don't have any emotional attachment to it but for sure it stood the test of time!


> I doubt many people are using XTENSA_TOOLS_ROOT as an input variable,

`XTENSA_TOOLS_ROOT` is _the_ main feature of this script (and of the XTOS one) that saves configuring all the rest.


See also:
- #8599
- #8578 (pure clean-up, no functional change)

andyross:
> Compare with the builds/ directory which is unambiguous:

But those aren't the directories that get set in XTENSA_SYSTEM?  How does this actually work for you?  The directory I get doesn't exist.

Or you're saying that SOF supports XTENSA_TOOLS_ROOT, but only if it's, I guess, "faked" to point to a directory within builds instead of tools?  Can we at least document how to do that?   What is the exact value of your XTENSA_TOOLS_ROOT at build time?

marc-hb:
> Or you're saying that SOF supports XTENSA_TOOLS_ROOT, but only if it's, I guess, "faked" to point to a directory within builds instead of tools?

Possibly.

> Can we at least document how to do that?

There is some small, closed-source installation script at Intel that is used to configure build systems, let me take a look and check what it actually does. I never expected it to do anything strange because you seem to be the first one to complain about this but maybe a lot of people have been secretly using their custom installation and/or build script(s), it would unfortunately be not the first time? I think I remember this installation script merely feeds some preset values to the main installer...

>  What is the exact value of your XTENSA_TOOLS_ROOT at build time?

See below. You can also infer it from https://sof-ci.01.org/sofpr/PR8598/build954/build/firmware_community/tgl.log

```
tree -d -L 3 $XTENSA_TOOLS_ROOT 
/home/me/XCC
├── install
│   ├── builds
│   │   ├── RD-2012.5-linux
│   │   ├── RF-2016.4-linux
│   │   ├── RG-2017.8-linux
│   │   ├── RI-2020.5-linux
│   │   └── RI-2022.10-linux
│   └── tools
│       ├── RD-2012.5-linux
│       ├── RF-2016.4-linux
│       ├── RG-2017.8-linux
│       ├── RI-2020.5-linux
│       └── RI-2022.10-linux
└── XtDevTools -> .

```

marc-hb:
> > Can we at least document how to do that?

> There is some small, closed-source installation script at Intel that is used to configure build systems, let me take a look and check what it actually does.

Here's the simplified equivalent pseudo-code of what it does. It wasn't me!

```
1. Download:
 XtensaTools_RG_2017_8_linux.tgz

tar xf
  => $XTENSA_TOOLS_ROOT/install/tools/RG-2017.8-linux/

2. Download:
 X6H3CNL_2017_8_linux_redist.tgz
 X4H3I16w2D48w3a_2017_8_linux.tgz
 cavs2x_LX6HiFi3_2017_8_linux_redist.tgz

For each .tgz:

tar xf 
  => $XTENSA_TOOLS_ROOT/install/builds/RG-2017.8-linux/

./install.sh
# 2) Enter the path to the Xtensa Tools directory:
  $XTENSA_TOOLS_ROOT/install/tools/RG-2017.8-linux/XtensaTools/
```

andyross:
OK... I think I see how this goes wrong.   The script has a fixed toolchain version per platform, where my scripting has all the platforms on the same toolchain (well, one of two) and put it on PATH.  In that case, all it needs is an idea of where the installation directory is (because xt-xxx tools are too dumb to autodetect). 

OK, will close this out and come up with something higher level, probably allowing for user control over toolchain version as the front end.

marc-hb:
I'm simplifying and clarifying the corresponding code in:
- #8890 

No functional change, just a cleanup. While doing this I discovered that XTENSA_SYSTEM seems... not required at all. At least not with the way we use toolchains currently.

PR #8571, opened 2023-12-05
Google AEC rework: unbreak/resurrect, mt8195, DP integration

andyross:
This series updates the Google RTC Audio Processing module, which has
always been deployed out of platform branches, to run on SOF main.
The tactic taken is a little different: previous integrations have
relied on getting pipeline dependencies to work, where this does a
little trickery at the component level (both in
google_rtc_audio_processing and demux) to control propagation of
command triggers across pipeline boundaries.  This should evolve
better with IPC4, which in (almost?) all cases refuses to propagate
those commands anyway.

The main advantage is that the pipelines become decoupled: you can
start an AEC'd capture stream without playback running and launch the
playback DMA later which then integrates with AEC dynamically; you can
shut either down in either order, start the playback before
capture, etc...

A secondary advantage is that, except for one logic fix with reset
propagation that I'm all but certain was a bug, no behavior changes
are needed to the pipeline state logic.

It also ends up a little cleaner and a little faster.

This work was done against a mt8195 device.  I'm going to look at MTL
immediately, hopefully nothing will be needed except the format
conversion.

PR #8603, opened 2023-12-11
[mtl-007] AEC fixes from main branch submission

andyross:
Squashed changes from https://github.com/thesofproject/sof/pull/8571 , rebased on top of the mtl-007-drop-stable branch so we can fix the divergences in this code.

Review there.  Not here.

### DIFF
--- a/src/audio/google/google_rtc_audio_processing.c
+++ b/src/audio/google/google_rtc_audio_processing.c
@@ -306,36 +306,19 @@ static int google_rtc_audio_processing_init(struct processing_module *mod)
 	struct module_config *cfg = &md->cfg;
 	const struct ipc4_base_module_extended_cfg *base_cfg = md->cfg.init_data;
 	struct ipc4_input_pin_format reference_fmt, output_fmt;
-	size_t in_fmt_size = sizeof(struct ipc4_input_pin_format);
-	size_t out_fmt_size = sizeof(struct ipc4_output_pin_format);
-	size_t size;
+	const size_t size = sizeof(struct ipc4_input_pin_format);
 
 	cd->config.base_cfg = base_cfg->base_cfg;
 
 	/* Copy the reference format from input pin 1 format */
-	memcpy_s(&reference_fmt, in_fmt_size,
-		 &base_cfg->base_cfg_ext.pin_formats[in_fmt_size], in_fmt_size);
-	memcpy_s(&output_fmt, in_fmt_size,
-		 &base_cfg->base_cfg_ext.pin_formats[in_fmt_size * GOOGLE_RTC_NUM_INPUT_PINS],
-		 in_fmt_size);
+	memcpy_s(&reference_fmt, size,
+		 &base_cfg->base_cfg_ext.pin_formats[size], size);
+	memcpy_s(&output_fmt, size,
+		 &base_cfg->base_cfg_ext.pin_formats[size * GOOGLE_RTC_NUM_INPUT_PINS], size);
 
 	cd->config.reference_fmt = reference_fmt.audio_fmt;
 	cd->config.output_fmt = output_fmt.audio_fmt;
-
-	/* save the base config extension */
-	size = base_cfg->base_cfg_ext.nb_input_pins * in_fmt_size +
-		base_cfg->base_cfg_ext.nb_output_pins * out_fmt_size;
-
-	md->cfg.basecfg_ext = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM,
-				      sizeof(struct ipc4_base_module_cfg_ext) + size);
-	if (!md->cfg.basecfg_ext) {
-		ret = -ENOMEM;
-		goto fail;
-	}
-
-	memcpy_s(md->cfg.basecfg_ext, size + sizeof(struct ipc4_base_module_cfg_ext),
-		 &base_cfg->base_cfg_ext,
-		 size + sizeof(struct ipc4_base_module_cfg_ext));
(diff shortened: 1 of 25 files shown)


marcinszkudlinski (changes requested):
007 branch is a frozen release "stable" branch for MTL, please don't merge this commit.

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -94,4 +94,7 @@ CONFIG_PROBE=y
 CONFIG_PROBE_DMA_MAX=2
 CONFIG_KCPS_DYNAMIC_CLOCK_CONTROL=y
 CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING=y
+CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_NUM_CHANNELS=2
+CONFIG_GOOGLE_RTC_AUDIO_PROCESSING_MIC_BITS=32
+
```
btian1, line 99: should no space line?

andyross: Please review in #8571 . This PR is just a place to deliver the same code against mtl-007-drop-stable.

andyross: Oh, actually this patch got dropped from that branch somehow!  Will fix.  But when I do please review there :)

src/audio/module_adapter/module_adapter.c
```diff
@@ -1088,9 +1088,6 @@ static int module_adapter_copy_dp_queues(struct comp_dev *dev)
 		struct sof_source *data_src = dp_queue_get_source(dp_queue);
 		size_t dp_data_available = source_get_data_available(data_src);
 
-		if (!dp_data_available)
```
marcinszkudlinski, line 1091: this error is a real error - if DP processing is working it will never appear, even during pipelie startup

tools/topology/topology2/include/components/google-rtc-aec.conf
```diff
@@ -47,6 +47,15 @@ Class.Widget."google-rtc-aec" {
 		token_ref	"comp.word"
 	}
 
+	Object.Control.bytes."1" {
+		access [
+			tlv_write
+			tlv_callback
```
cujomalainey, line 53: is callback a kernel or userspace permission?

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -53,21 +52,48 @@ DECLARE_SOF_RT_UUID("google-rtc-audio-processing", google_rtc_audio_processing_u
...
+
+static __aligned(PLATFORM_DCACHE_ALIGN)
+uint8_t aec_mem_blob[CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_MEMORY_BUFFER_SIZE_BYTES];
+
+#define NUM_FRAMES (CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_SAMPLE_RATE_HZ \
+		    / GOOGLE_RTC_AUDIO_PROCESSING_FREQENCY_TO_PERIOD_FRAMES)
+#define CHAN_MAX CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING_CHANNEL_MAX
+
+static __aligned(PLATFORM_DCACHE_ALIGN)
+float refoutbuf[CHAN_MAX][NUM_FRAMES];
```
cujomalainey, line 71: not a fan of having these here long term. Should we want to start supporting >1 AEC instance (e.g. on HS) this will not end well.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -80,28 +106,130 @@ void GoogleRtcFree(void *ptr)
 	return rfree(ptr);
 }
 
-static void google_rtc_audio_processing_params(struct processing_module *mod,
-					       struct sof_source *ref,
-					       struct sof_source *mic,
-					       struct sof_sink *out)
+static ALWAYS_INLINE float s16_to_float(const char *ptr)
```
cujomalainey, line 109: Why are we not using the macros [here](https://github.com/thesofproject/sof/blob/b3b9abffe7ee85bf0884d392d3d243161efadbbe/src/include/sof/audio/format.h#L71)?

andyross: (Again, please continue review in the main PR, but answering here):

This was discussed earlier: Those generate horrifyingly pessimal code (runtime division!) and can't be fixed without breaking the ABI (because as specified they work for 64 bit values).  These seem like they're intended for generating compile time constants for fixed point use.  There's also a pcm_convert API elsewhere in SOF, but that's designed to work in a block on linear arrays of samples (and don't expose inlinable single-sample conversions) where here we need to unify de/interleaving with the conversion.

Basically: SOF just doesn't have an API for this.  I could move this somewhere else and try to unify with an existing library.

tools/topology/topology2/include/components/google-rtc-aec.conf
andyross: This is a cherry pick from main.  And "access" in this context refers (I think!) to the API variant used to access the control's data.  "callback" is what needs to be used for large TLV blocks because they can't be sent in a single IPC command.

andyross:
Updated with current versions of all my in-review PRs, unifying the AEC work in main with the MTL branch.

andyross:
Close this one.   Main has now caught up with 007-drop-stable and this works directly on top of it.  No need to maintain the separate branch anymore.

PR #8610, opened 2023-12-11
topology2: Remove obsolete rate and channels properties from pipeline objects

jsarha:
The channels, channels_min, channels_max, rate, rate_min, and rate_max pipeline object properties are obsolete and not used. Let's get rid of them,

### DIFF
--- a/tools/topology/topology2/include/pipelines/cavs/dai-copier-be.conf
+++ b/tools/topology/topology2/include/pipelines/cavs/dai-copier-be.conf
@@ -9,8 +9,6 @@
 # Object.Pipeline.dai-copier-be."N" {
 # 	period		1000
 # 	time_domain	"timer"
-# 	channels	2
-# 	rate		48000
 # }
 #
 # Where N is the unique pipeline ID within the same alsaconf node.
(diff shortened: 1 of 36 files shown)


singalsu (approved):
Thanks for the cleanup! Not used properties are rather confusing.

kv2019i (approved):
This does beg the question why did we have this all in the definitions. @ranj063 @jsarha Was this planned to be used for something before? Oh well, if everything still works with all this removed, I have no objections.

jsarha:
> This does beg the question why did we have this all in the definitions. @ranj063 @jsarha Was this planned to be used for something before? Oh well, if everything still works with all this removed, I have no objections.

I did not go back in the history to look if these were ever used, but as far as I can tell they are not used anymore.  There was no vendor token definitions for these either, so the attribute values should be copied in palce already by alsatplg if they are used. However, I can only find handling of attributes by these names for pcm_caps objects, which AFAIK should not be used in pipeline objects, so the attribute values from pipeline objects should find their way to such objects. Then again, I still do not fully understand how alsatplg works, so there is a chance I am mistaken, but the CI seems to prove othewise.

ranj063:
> This does beg the question why did we have this all in the definitions. @ranj063 @jsarha Was this planned to be used for something before? Oh well, if everything still works with all this removed, I have no objections.

@kv2019i these were added because they are needed for IPC3 but once we dropped support for IPC3 in tplg2, we never went back to clean it up

PR #8612, opened 2023-12-12
module_adapter: build params first then verify it

RanderWang:
We need to build params based on module base config first then we can verify it since the income params is built for the source of the this module . In comp_verify_params the params is applied to buffer so the sequence is vital. And also remove redundant stream_param set.

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -581,6 +581,8 @@ int module_adapter_params(struct comp_dev *dev, struct sof_ipc_stream_params *pa
 	int ret;
 	struct processing_module *mod = comp_get_drvdata(dev);
 
+	module_adapter_set_params(mod, params);
+
 	ret = comp_verify_params(dev, mod->verify_params_flags, params);
 	if (ret < 0) {
 		comp_err(dev, "module_adapter_params(): comp_verify_params() failed.");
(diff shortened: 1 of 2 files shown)


dbaluta (approved):

jxstelter (approved):

tmleman (approved):

btian1 (approved):
" the this module" something wrong with this in commit message.

kv2019i (commented):
@RanderWang The commit message is hard to read. My immediate reaction is that if the order is wrong, how has this worked before this commit? 

I won't block as others seem to have understood this better, but in future commits, please expand (in the commit message) what problem the patch is solving. And if it's something obvious (like here, we verify before parameter object is ready), reviewers (and people later maintaining the code) would really appreciate an explanation how did this work despite the flaw.

dbaluta:
@kv2019i agree with you, commit messages should explain WHY the change is needed and how in this specific case it worked before.

I Acked this patch only because on IPC3 this patch is a NO-OP and assumed that IPC4-gang reviewed for that case.

RanderWang:
> @RanderWang The commit message is hard to read. My immediate reaction is that if the order is wrong, how has this worked before this commit?
This is a general solution and all the build-in modules do it by itself. Loadable module will use this general solution.

PR #8614, opened 2023-12-12
topology1: pipe-volume-capture.m4: Set minimum number of channels to 2

LaurentiuM1234:
Currently, the minimum number of channels for the capture PCM is set to the same value as the maximum number. This restricts the number of channels supported by the PCM to the maximum number. This is wrong if we want to allow arecord to work with multiple channel values. As such, follow the example of "pipe-volume-playback.m4" and introduce a new macro: LOCAL_CHANNELS_MIN. This will be set to "CHANNELS_MIN" if said macro is defined, otherwise it will be set to 2.


kv2019i (approved):

iuliana-prodan (approved):

dbaluta (approved):

kv2019i:
Intel CI stuck, but as this is tplg1 only, this has not relation, so I'll proceed with merge.

PR #8619, opened 2023-12-13
demux: Support inactive cross-pipeline sinks

andyross:
[Split out from #8571 for separate review]

It may happen that a demux is configured to write to a buffer in a stopped pipeline (consider echo cancellation when the mic is not in use: the output reference stream has nowhere to go).  This doesn't work in the tree currently; the general pipeline design is to try to turn on pipes that are required, but that's incomplete for the case (again, echo cancellation) where the pipelines are oriented in different directions and in any case is undesirable as a microphone stream shouldn't be required to be active for playback to work.

Instead, detect the situation at PRE_START time, configure the output buffers overrun_permitted (as there may not be a reader), and drop the unconfigured streams dynamically at process() time, as the target pipeline may be started up at arbitrary moments.  When it starts, we'll begin emitting output at the first process callback as desired.

### DIFF
--- a/src/audio/mux/mux.c
+++ b/src/audio/mux/mux.c
@@ -270,9 +270,12 @@ static int demux_process(struct processing_module *mod,
 
 	/* produce output, one sink at a time */
 	for (i = 0; i < num_output_buffers; i++) {
-		demux_prepare_active_look_up(cd, sinks_stream[i],
-					     input_buffers[0].data, look_ups[i]);
-		cd->demux(dev, sinks_stream[i], input_buffers[0].data, frames, &cd->active_lookup);
+		if (sinks_stream[i]) {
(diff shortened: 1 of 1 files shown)


kv2019i (approved):
One may argue the need to run the capture pipe was/is a feature not a bug, but I can certainly see value in being more flexible w.r.t. a stopped pipeline.

Added a few reviewers. CI coverage for this especially in IPC3 demux is not great, so hopefully we can get some IPC3 eyes on this. Looks good based on code review.

btian1 (commented):
should use two patches?
1. add sinks_stream[i]
2. add demux trigger

andyross:
> should use two patches?

IMHO no?  There's no working/bisectable configuration that results from just one of them, you need both to get working output.

btian1 (approved):

PR #8628, opened 2023-12-14
[DNM] west.yml: test Zephyr PR 66517

kv2019i:


kv2019i:
@fkwasowi FYI, let's do a test run with your Zephyr PR

PR #8632, opened 2023-12-15
zephyr: lib: alloc: Use cached memory for L3 Heap

jxstelter:
This patch implements recommended hardware flow for Intel ACE platforms.
The L3 heap should be accessed via cached pointers including management data.
This change is fixing D3 save/restore flow on LNL.

### DIFF
--- a/zephyr/lib/alloc.c
+++ b/zephyr/lib/alloc.c
@@ -116,8 +116,7 @@ static inline uintptr_t get_l3_heap_start(void)
 	 * - main_fw_load_offset
 	 * - main fw size in manifest
 	 */
-	return (uintptr_t)z_soc_uncached_ptr((__sparse_force void __sparse_cache *)
-					     ROUND_UP(IMR_L3_HEAP_BASE, L3_MEM_PAGE_SIZE));
+	return (uintptr_t)(ROUND_UP(IMR_L3_HEAP_BASE, L3_MEM_PAGE_SIZE));
 }
 
 /**
@@ -145,14 +144,50 @@ static bool is_l3_heap_pointer(void *ptr)
 	uintptr_t l3_heap_start = get_l3_heap_start();
 	uintptr_t l3_heap_end = l3_heap_start + get_l3_heap_size();
 
-	if (is_cached(ptr))
-		ptr = z_soc_uncached_ptr((__sparse_force void __sparse_cache *)ptr);
-
 	if ((POINTER_TO_UINT(ptr) >= l3_heap_start) && (POINTER_TO_UINT(ptr) < l3_heap_end))
 		return true;
 
 	return false;
 }
+
+static void *l3_heap_alloc_aligned(struct k_heap *h, size_t min_align, size_t bytes)
+{
+	k_spinlock_key_t key;
+	void *ret;
+#if CONFIG_SYS_HEAP_RUNTIME_STATS && CONFIG_IPC_MAJOR_4
+	struct sys_memory_stats stats;
+#endif
+	if (!cpu_is_primary(arch_proc_id())) {
+		tr_err(&zephyr_tr, "L3_HEAP available only for primary core!");
+		return NULL;
+	}
+
+	key = k_spin_lock(&h->lock);
+	ret = sys_heap_aligned_alloc(&h->heap, min_align, bytes);
+	k_spin_unlock(&h->lock, key);
+
+#if CONFIG_SYS_HEAP_RUNTIME_STATS && CONFIG_IPC_MAJOR_4
+	sys_heap_runtime_stats_get(&h->heap, &stats);
+	tr_info(&zephyr_tr, "heap allocated: %u free: %u max allocated: %u",
+		stats.allocated_bytes, stats.free_bytes, stats.max_allocated_bytes);
+#endif
+
+	return ret;
+}
+
+static void l3_heap_free(struct k_heap *h, void *mem)
+{
+	if (!cpu_is_primary(arch_proc_id())) {
+		tr_err(&zephyr_tr, "L3_HEAP available only for primary core!");
+		return;
+	}
+
+	k_spinlock_key_t key = k_spin_lock(&h->lock);
+
+	sys_heap_free(&h->heap, mem);
+	k_spin_unlock(&h->lock, key);
+}
+
 #endif
 
 static void *heap_alloc_aligned(struct k_heap *h, size_t min_align, size_t bytes)
@@ -251,6 +286,17 @@ void *rmalloc(enum mem_zone zone, uint32_t flags, uint32_t caps, size_t bytes)
 	if (caps & SOF_MEM_CAPS_L3) {
 #if CONFIG_L3_HEAP
 		heap = &l3_heap;
+		/* Uncached L3_HEAP should be not used */
+		if (!zone_is_cached(zone)) {
+			tr_err(&zephyr_tr, "L3_HEAP available for cached zones only!");
+			return NULL;
+		}
+		ptr = (__sparse_force void *)l3_heap_alloc_aligned(heap, 0, bytes);
+
+		if (!ptr && zone == SOF_MEM_ZONE_SYS)
+			k_panic();
+
+		return ptr;
 #else
 		k_panic();
 #endif
@@ -335,10 +381,24 @@ EXPORT_SYMBOL(rzalloc);
 void *rballoc_align(uint32_t flags, uint32_t caps, size_t bytes,
 		    uint32_t align)
 {
-	if (flags & SOF_MEM_FLAG_COHERENT)
-		return heap_alloc_aligned(&sof_heap, align, bytes);
+	struct k_heap *heap;
 
-	return (__sparse_force void *)heap_alloc_aligned_cached(&sof_heap, align, bytes);
+	/* choose a heap */
+	if (caps & SOF_MEM_CAPS_L3) {
+#if CONFIG_L3_HEAP
+		heap = &l3_heap;
+		return (__sparse_force void *)l3_heap_alloc_aligned(heap, align, bytes);
+#else
+		k_panic();
+#endif
+	} else {
+		heap = &sof_heap;
+	}
+
+	if (flags & SOF_MEM_FLAG_COHERENT)
+		return heap_alloc_aligned(heap, align, bytes);
+
+	return (__sparse_force void *)heap_alloc_aligned_cached(heap, align, bytes);
 }
 EXPORT_SYMBOL(rballoc_align);
 
@@ -352,7 +412,7 @@ void rfree(void *ptr)
 
 #if CONFIG_L3_HEAP
 	if (is_l3_heap_pointer(ptr)) {
-		heap_free(&l3_heap, ptr);
+		l3_heap_free(&l3_heap, ptr);
 		return;
 	}
 #endif
--- a/zephyr/lib/cpu.c
+++ b/zephyr/lib/cpu.c
@@ -96,10 +96,16 @@ void cpu_notify_state_entry(enum pm_state state)
 		storage_buffer_size += LP_SRAM_SIZE;
 
 		/* allocate IMR buffer and store it in the global pointer */
-		global_imr_ram_storage = rmalloc(SOF_MEM_ZONE_SYS_RUNTIME,
-						 0,
-						 SOF_MEM_CAPS_L3,
-						 storage_buffer_size);
+		global_imr_ram_storage = rballoc_align(0, SOF_MEM_CAPS_L3,
+						       storage_buffer_size,
+						       PLATFORM_DCACHE_ALIGN);
+
+		/* If no IMR buffer we can not recover */
+		if (!global_imr_ram_storage) {
+			tr_err(&zephyr_tr, "failed to allocate global_imr_ram_storage");
+			k_panic();
+		}
+
 #endif /* CONFIG_ADSP_IMR_CONTEXT_SAVE */
 	}
 }
--- a/src/library_manager/lib_manager.c
+++ b/src/library_manager/lib_manager.c
@@ -539,16 +539,14 @@ static void __sparse_cache *lib_manager_allocate_store_mem(uint32_t size,
 	void __sparse_cache *local_add;
 #if CONFIG_L3_HEAP
 	uint32_t caps = SOF_MEM_CAPS_L3 | SOF_MEM_CAPS_DMA;
-
-	/* allocate new buffer: cached alias */
-	local_add = (__sparse_force void __sparse_cache *)rmalloc(SOF_MEM_ZONE_SYS, 0, caps, size);
 #else
-	uint32_t addr_align = PAGE_SZ;
 	uint32_t caps = SOF_MEM_CAPS_DMA;
+#endif
 
+	uint32_t addr_align = PAGE_SZ;
 	/* allocate new buffer: cached alias */
 	local_add = (__sparse_force void __sparse_cache *)rballoc_align(0, caps, size, addr_align);
-#endif
+
 	if (!local_add) {
 		tr_err(&lib_manager_tr, "lib_manager_allocate_store_mem(): alloc failed");
 		return NULL;


zephyr/lib/alloc.c
```diff
@@ -144,14 +144,32 @@ static bool is_l3_heap_pointer(void *ptr)
...
 }
+
+/**
+ * Flush cached L3 heap management data.
+ */
+static inline void l3_heap_flush(struct k_heap *h)
+{
+	if (h == &l3_heap)
+		sys_cache_data_flush_and_invd_range((void *)get_l3_heap_start(),
+						    L3_MEM_PAGE_SIZE);
```
lyakh, line 160: there's 1 page worth of L3 management data in the beginning of L3 memory? What is it?

zephyr/lib/alloc.c
```diff
@@ -163,7 +181,13 @@ static void *heap_alloc_aligned(struct k_heap *h, size_t min_align, size_t bytes
 #endif
 
 	key = k_spin_lock(&h->lock);
+#if CONFIG_L3_HEAP
+	l3_heap_invalidate(h);
+#endif
```
lyakh, line 186: this looks strange... You just took a spinlock at `&h->lock` which presumably will modify RAM contents, i.e. its cache, and now you invalidate caches. Is that correct? Why is this invalidation needed at all?

zephyr/lib/alloc.c
```diff
@@ -212,15 +236,24 @@ static void heap_free(struct k_heap *h, void *mem)
 
 	if (is_cached(mem)) {
 		mem_uncached = z_soc_uncached_ptr((__sparse_force void __sparse_cache *)mem);
+#if CONFIG_L3_HEAP
+		if (h == &l3_heap)
+			mem_uncached = mem;
```
lyakh, line 241: this looks so strange... We take great care in this function to separate cached and uncached pointers and this just mixes them up. Maybe we need to introduce a new type of "cached heap?"

zephyr/lib/alloc.c
jxstelter: The k_heap structure and its spinlock are located in HPSRAM, while l3_heap_invalidate() invalidates cache entries for L3 address range. There is z_heap zephyr internal structure that must be updated each time new alloc/free operation is done. Therefore previous implementation based on uncashed addresses and only pointers to memory buffers itself were translated to cached address. However for L3 Memory on ACE the recommendation is to use always cached addresses.

zephyr/lib/alloc.c
```diff
@@ -341,6 +377,7 @@ void *rballoc_align(uint32_t flags, uint32_t caps, size_t bytes,
 /*
  * Free's memory allocated by above alloc calls.
  */
+__attribute__((optimize("-O0")))
```
lgirdwood, line 380: Best to say why this is needed with an inline comment, It could be tool/platform specific.

lgirdwood (commented):
One question from me on the code.

Thinking long term I guess a lot of this heap mgmt will end up in Zephyr, but this will need memory native API usage 1st.

zephyr/lib/alloc.c
lyakh: indeed... why is this needed at all?

zephyr/lib/alloc.c
```diff
@@ -145,14 +145,89 @@ static bool is_l3_heap_pointer(void *ptr)
...
+static void *l3_heap_alloc_aligned(struct k_heap *h, size_t min_align, size_t bytes)
+{
+	k_spinlock_key_t key;
+	void *ret;
+#if CONFIG_SYS_HEAP_RUNTIME_STATS && CONFIG_IPC_MAJOR_4
+	struct sys_memory_stats stats;
+#endif
+
+	key = k_spin_lock(&h->lock);
+	l3_heap_invalidate(h);
```
lyakh, line 183: sorry, I must be misunderstanding something. To me it seems like - this PR changes L3 heap access to cached, now when a core allocates a new buffer in L3, this function first *drops* all cached data this core could have in the heap? This seems like a sure way to lose cached writes to me?

zephyr/lib/alloc.c
```diff
@@ -145,14 +145,89 @@ static bool is_l3_heap_pointer(void *ptr)
...
+	ret = sys_heap_aligned_alloc(&h->heap, min_align, bytes);
+	l3_heap_flush(h);
```
lyakh, line 185: and after allocation you write back caches (to make changes visible to other cores, that's logical) and then you discard cached data again - why?

zephyr/lib/alloc.c
```diff
@@ -145,14 +145,89 @@ static bool is_l3_heap_pointer(void *ptr)
...
+	if (ptr)
+		ptr = z_soc_cached_ptr((__sparse_force void *)ptr);
+
+	return ptr;
+}
+
+static void l3_heap_free(struct k_heap *h, void *mem)
+{
+	k_spinlock_key_t key = k_spin_lock(&h->lock);
+	sys_cache_data_flush_and_invd_range(mem, get_l3_heap_size());
```
lyakh, line 225: this looks wrong - please either use `sys_cache_data_*_()` directly or your above wrappers consistently everywhere. Then - you manipulate cache for a range, starting at *some* address in L3 memory, but with size equal to entire L3 memory? So you might try to manipulate cache for almost an entire L3 size beyond the L3 heap range? Then - I don't think this is needed at all. Why write back user data if the memory is freed? The data is discarded. Why care to invalidate it here, maybe better do it when allocating a cached buffer

zephyr/lib/alloc.c
```diff
@@ -145,14 +145,89 @@ static bool is_l3_heap_pointer(void *ptr)
...
+	struct sys_memory_stats stats;
+#endif
+
+	key = k_spin_lock(&h->lock);
+	l3_heap_invalidate(h);
+	ret = sys_heap_aligned_alloc(&h->heap, min_align, bytes);
+	l3_heap_flush(h);
+	k_spin_unlock(&h->lock, key);
+
+#if CONFIG_SYS_HEAP_RUNTIME_STATS && CONFIG_IPC_MAJOR_4
```
lyakh, line 188: what does it have to do with IPC?

zephyr/lib/alloc.c
```diff
@@ -115,8 +115,8 @@ static inline uintptr_t get_l3_heap_start(void)
 	 * - main_fw_load_offset
 	 * - main fw size in manifest
 	 */
-	return (uintptr_t)z_soc_uncached_ptr((__sparse_force void __sparse_cache *)
-					     ROUND_UP(IMR_L3_HEAP_BASE, L3_MEM_PAGE_SIZE));
+	return (uintptr_t)((__sparse_force void __sparse_cache *)
+			   ROUND_UP(IMR_L3_HEAP_BASE, L3_MEM_PAGE_SIZE));
```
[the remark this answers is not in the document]
lyakh: Doesn't this mean, that your heap start is now cached, so any allocation from it is now cached, so both cached and uncached allocations return cached buffer addresses?

lyakh (changes requested):

zephyr/lib/alloc.c
lgirdwood: I would suggest adding some extra inline comments to describe the flows/usage. This way it should be easier to follow the code and any assumptions.

lgirdwood (commented):
Still some opens, but I think some inline comments could allow everyone to follow. CI Zephyr failures unrelated to this PR.

zephyr/lib/alloc.c
jxstelter: fixed

zephyr/lib/alloc.c
jxstelter: Yes, this was mistake. Fixed.

zephyr/lib/alloc.c
jxstelter: fixed

zephyr/lib/alloc.c
jxstelter: fixed

zephyr/lib/alloc.c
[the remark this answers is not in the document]
jxstelter: Attempt to allocate uncached will fail. 
https://github.com/jxstelter/sof/blob/cfe9e9b1e32be649456af34d9ad1ed2ce6652dc0/zephyr/lib/alloc.c#L326

zephyr/lib/alloc.c
jxstelter: not valid

zephyr/lib/alloc.c
jxstelter: fixed

zephyr/lib/alloc.c
```diff
@@ -367,6 +447,7 @@ static int heap_init(void)
 
 #if CONFIG_L3_HEAP
 	sys_heap_init(&l3_heap.heap, UINT_TO_POINTER(get_l3_heap_start()), get_l3_heap_size());
+	l3_heap_flush_invalidate(&l3_heap);
```
kv2019i, line 450: I'm not sure if this is safe with cached memory. sys_heap will traverse the metadata and access blocks before and after the allocated one. So even if the individual allocations are aligned to cachelines, when a block is freed/allocated, the sys_heap will access the N-1 and N+1 block metadata. These blocks might be allocated to different cores, and with a non-coherent architecture, there's no way to do this safely. I.e. if the metadata is accessed via cached memory, then there's no way to guarantee coherency --- if multiple cores access the heap. This is why we put the sys_heap metadata on uncached memory.  Right @lyakh ?

I understand this is very limiting, but can we limit L3 heap to a single core only. That would work. 

Otherwise, I don't think we can use sys_heap for L3 on these platforms, but require an alternative implementation that can handle the specific constraints we have here.

kv2019i (commented):
See inline

zephyr/lib/alloc.c
jxstelter: Agree. Currently L3 heap is used only by core 0 in two cases:
- allocate D3 context storage location;
- allocate loadable modules storage;
Both operations could be done only by core 0. 
This patch is provided to unblock LNL SOF testing on RVP.  WE can consider to replace the sys_heap API with for example Memory Slabs https://docs.zephyrproject.org/latest/kernel/memory_management/slabs.html.
I could add explicit check against core number during memory allocation/free.

lyakh: no, spinlocks don't prevent incoherency, they only prevent concurrent access.

zephyr/lib/alloc.c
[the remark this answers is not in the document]
lyakh: @jxstelter ok, but that still doesn't answer the question. I find the cache management calls in this PR very confusing and very likely wrong. If we limit L3 heap to core 0 only, we shouldn't need *any* cache handling here *at all*

zephyr/lib/alloc.c
jxstelter: Oh, I was thinking about combination of spinlock and correct cache flushing/invalidating. But You are right, this could be done from zephyr level when it is exactly known, which data to flush and invalidate. From SOF level it will not work with multiple core access. 
For now limiting L3_HEAP to primary core only.

zephyr/lib/alloc.c
```diff
@@ -145,14 +145,71 @@ static bool is_l3_heap_pointer(void *ptr)
...
+static void *l3_heap_alloc_aligned(struct k_heap *h, size_t min_align, size_t bytes)
+{
+	k_spinlock_key_t key;
+	void *ret;
+#if CONFIG_SYS_HEAP_RUNTIME_STATS && CONFIG_IPC_MAJOR_4
+	struct sys_memory_stats stats;
+#endif
+	if (!cpu_is_primary(arch_proc_id())) {
+		tr_err(&zephyr_tr, "L3_HEAP available only for primary core!");
+		k_panic();
```
lyakh, line 163: how about just returning `NULL`?

zephyr/lib/alloc.c
```diff
@@ -145,14 +145,71 @@ static bool is_l3_heap_pointer(void *ptr)
...
+	 * heap allocation. To ensure no allocated cached buffer
+	 * overlaps the same cacheline with the metadata chunk,
+	 * align both allocation start and size of allocation
+	 * to cacheline. Only cached allocations are supported in
+	 * l3_heap.
+	 */
+	min_align = MAX(PLATFORM_DCACHE_ALIGN, min_align);
+	bytes = ALIGN_UP(bytes, min_align);
+
+	ptr = (void __sparse_cache *)l3_heap_alloc_aligned(h, min_align, bytes);
```
lyakh, line 195: `return (void __sparse_cache *)l3_heap_alloc_aligned(h, min_align, bytes);`

zephyr/lib/alloc.c
```diff
@@ -145,14 +145,71 @@ static bool is_l3_heap_pointer(void *ptr)
...
+
+	return ptr;
+}
+
+static void l3_heap_free(struct k_heap *h, void *mem)
+{
+	if (!cpu_is_primary(arch_proc_id())) {
+		tr_err(&zephyr_tr, "L3_HEAP available only for primary core!");
+		k_panic();
```
lyakh, line 204: just `return`

zephyr/lib/alloc.c
```diff
@@ -251,6 +308,15 @@ void *rmalloc(enum mem_zone zone, uint32_t flags, uint32_t caps, size_t bytes)
 	if (caps & SOF_MEM_CAPS_L3) {
 #if CONFIG_L3_HEAP
 		heap = &l3_heap;
+		/* Uncached L3_HEAP should be not used */
+		if (!zone_is_cached(zone))
+			k_panic();
```
lyakh, line 313: I really wouldn't try to crash the system that often...

zephyr/lib/alloc.c
jxstelter: fixed

zephyr/lib/alloc.c
jxstelter: fixed

zephyr/lib/alloc.c
jxstelter: fixed

zephyr/lib/alloc.c
jxstelter: fixed

zephyr/lib/alloc.c
```diff
@@ -145,14 +145,67 @@ static bool is_l3_heap_pointer(void *ptr)
...
+	/*
+	 * Zephyr sys_heap stores metadata at start of each
+	 * heap allocation. To ensure no allocated cached buffer
+	 * overlaps the same cacheline with the metadata chunk,
+	 * align both allocation start and size of allocation
+	 * to cacheline. Only cached allocations are supported in
+	 * l3_heap.
+	 */
+	min_align = MAX(PLATFORM_DCACHE_ALIGN, min_align);
+	bytes = ALIGN_UP(bytes, min_align);
```
lyakh, line 191: trying to think about it: now we limit the whole L3 heap to core 0, right? Doesn't this mean that we don't need cache-line alignment any more? I don't think Zephyr's heap API explicitly manipulates the caches either, but please double-check. So I would expect this function to be just an alias to `l3_heap_alloc_aligned()`

zephyr/lib/alloc.c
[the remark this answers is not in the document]
lyakh: ...please, remove the type-casts

zephyr/lib/alloc.c
marcinszkudlinski: >I was thinking about combination of spinlock and correct cache flushing/invalidating

coherent.h provides such functionality

marcinszkudlinski (approved):

zephyr/lib/alloc.c
jxstelter: Zephyr heap is not manipulating this, but cache line alignment is required at least for library manager functionality. 
It allocates buffer from L3_HEAP for loadable library storage. After load the library image to this buffer we need to flush cache to have whole library stored in memory. It could be accessed after wake from D3. On the other hands, the D3 context save/restore buffer is also allocated and flushed before D3 entry. Without the cache line alignment I see failures in tests where D3 entry/exit is introduced between Load Library phase and Init Instance phase (this is real E2E scenario).

zephyr/lib/alloc.c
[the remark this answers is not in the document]
jxstelter: fixed

zephyr/lib/alloc.c
lyakh: > cache line alignment is required at least for library manager functionality.

then it's the **caller** who has to specify that alignment, in this case the library manager. That's why we have `alignment` as a function parameter here.

jxstelter: Yes, this is true. Unfortunately alignment parameter is omitted in upper API function
**void *rmalloc(enum mem_zone zone, uint32_t flags, uint32_t caps, size_t bytes)**. I can fulfill your request adding new API function e.g.
**void *rmalloc_alligned(enum mem_zone zone, uint32_t flags, uint32_t caps, size_t bytes, uint32_t alignment)**.
You think it would be ok?

jxstelter: fixed

pjdobrowolski (approved):

abonislawski (approved):

tmleman (approved):

lgirdwood:
@lyakh I think your review comments all addressed now.

lgirdwood (approved):

zephyr/lib/alloc.c
```diff
@@ -335,10 +381,24 @@ EXPORT_SYMBOL(rzalloc);
...
 {
+	struct k_heap *heap;
+
+	/* choose a heap */
+	if (caps & SOF_MEM_CAPS_L3) {
+#if CONFIG_L3_HEAP
+		heap = &l3_heap;
+		return (__sparse_force void *)l3_heap_alloc_aligned(heap, align, bytes);
+#else
+		k_panic();
```
lyakh, line 392: this is kind of theoretical, but maybe a small follow-up PR to convert this and line 301 above to `return NULL`

zephyr/lib/cpu.c
```diff
@@ -96,10 +96,16 @@ void cpu_notify_state_entry(enum pm_state state)
...
-						 SOF_MEM_CAPS_L3,
-						 storage_buffer_size);
+		global_imr_ram_storage = rballoc_align(0, SOF_MEM_CAPS_L3,
+						       storage_buffer_size,
+						       PLATFORM_DCACHE_ALIGN);
+
+		/* If no IMR buffer we can not recover */
+		if (!global_imr_ram_storage) {
+			tr_err(&zephyr_tr, "failed to allocate global_imr_ram_storage");
+			k_panic();
```
lyakh, line 106: this means: the DSP is being powered down and we try to allocate an IMR buffer to safe state, and if we fail, we panic. Would it be possible to just abort saving state and go for a clean power off? Also rather theoretical, but would probably be better not to bother the host with a DSP panic but just to make this a clean shut down? @ujfalusi Can be an incremental PR too

lyakh (approved):

marc-hb:
This PR broke fuzzing, @andyross any idea why?

I fails very quickly and I can reproduce locally without any issue.

https://github.com/thesofproject/sof/actions/runs/7724788048/job/21057626265
https://github.com/thesofproject/sof/actions/runs/7749595113/job/21134436910

It was harder to notice because `zmain` has been broken for a while but it was not _that_ hard to notice.

@jxstelter please pay attention to test results next time, reviewers have one order of magnitude more PRs to pay attention to.


```
#119902	NEW    cov: 729 ft: 3099 corp: 335/18Kb lim: 240 exec/s: 7993 rss: 87Mb L: 61/239 MS: 1 EraseBytes-
Exiting due to fatal error
==5292== ERROR: libFuzzer: fuzz target exited
    #0 0x8134c42 in __sanitizer_print_stack_trace (/home/runner/work/sof/sof/workspace/build-fuzz/zephyr/zephyr.exe+0x8134c42)
    #1 0x80aa4d6 in fuzzer::PrintStackTrace() (/home/runner/work/sof/sof/workspace/build-fuzz/zephyr/zephyr.exe+0x80aa4d6)
    #2 0x8090897 in fuzzer::Fuzzer::ExitCallback() (/home/runner/work/sof/sof/workspace/build-fuzz/zephyr/zephyr.exe+0x8090897)
    #3 0x809082e in fuzzer::Fuzzer::StaticExitCallback() (/home/runner/work/sof/sof/workspace/build-fuzz/zephyr/zephyr.exe+0x809082e)
    #4 0xf783a332  (/lib/i386-linux-gnu/libc.so.6+0x3a332) (BuildId: 0598ef3e075d7653ff4d565675d15666ec9b7b31)
    #5 0xf783a486 in exit (/lib/i386-linux-gnu/libc.so.6+0x3a486) (BuildId: 0598ef3e075d7653ff4d565675d15666ec9b7b31)
    #6 0x8194c5c in posix_exit /home/runner/work/sof/sof/workspace/zephyr/boards/posix/native_posix/main.c:51:2

SUMMARY: libFuzzer: fuzz target exited
```

btian1:
@marc-hb ，Thanks for point out, my PR also got FUZZY error, I am looking for the reason, could you help double confirm below FUZZY error also caused by this:
https://github.com/thesofproject/sof/actions/runs/7751869504/job/21140431875?pr=8822
?

marc-hb:
I debugged this a bit. This PR is probably just the messenger but this fuzzing failure looks like a "good catch" to me.

One of the new k_panic() gets triggered because `caps & SOF_MEM_CAPS_L3` is true even when `CONFIG_L3_HEAP` is false.

I think the reason `caps & SOF_MEM_CAPS_L3` is true is because... `caps` comes directly from untrusted IPC input!?

The panic happens when `ipc_glb_tplg_buffer_new()` does this:
```
ret = ipc_buffer_new(ipc, (struct sof_ipc_buffer *)ipc->comp_data); 
```

At this point `comp_data` looks like it came straight from the fuzzer's untrusted input:

```
(gdb) p /x *desc

$5 = {comp = {hdr = {size = 0x66, cmd = 0x3020ffff}, id = 0xb6b6b600, type = 0xb6b6b6b6, pipeline_id = 0xb6b6b6b6, core = 0xffffb236, 
    ext_data_length = 0xffffffff}, size = 0xffffffff, caps = 0xffffffff, flags = 0xffffffff, reserved = 0xffffffff}
```    


```
#0  rballoc_align (flags=0, caps=4294903040, bytes=86376703, align=64) at sof/zephyr/lib/alloc.c:398
#1  0x0831d38c in buffer_alloc (size=86376703, caps=4294903040, flags=4294967295, align=64, is_shared=false)
    at sof/src/audio/buffer.c:58
#2  0x082e4d6b in buffer_new (desc=0x8b184c0 <heapmem+1728>, is_shared=false) at sof/src/ipc/ipc-helper.c:48
#3  0x082cc78e in ipc_buffer_new (ipc=0x8b183c0 <heapmem+1472>, desc=0x8b184c0 <heapmem+1728>) at sof/src/ipc/ipc3/helper.c:459
#4  0x082b7249 in ipc_glb_tplg_buffer_new (header=807469055) at sof/src/ipc/ipc3/handler.c:1305
#5  0x082b090c in ipc_glb_tplg_message (header=807469055) at sof/src/ipc/ipc3/handler.c:1416
#6  0x082aff8a in ipc_cmd (_hdr=0x8b184c0 <heapmem+1728>) at sof/src/ipc/ipc3/handler.c:1651
#7  0x082fcb1b in ipc_platform_do_cmd (ipc=0x8b183c0 <heapmem+1472>) at sof/src/platform/posix/ipc.c:162
#8  0x082e2827 in ipc_do_cmd (data=0x8b183c0 <heapmem+1472>) at sof/src/ipc/ipc-common.c:328
#9  0x0836a27a in task_run (task=0x8b183e8 <heapmem+1512>) at sof/zephyr/include/rtos/task.h:94
```

kv2019i:
D'oh, my mistake, I did not notice the fail when merging. @jxstelter can you take a quick look so we don't need a revert?

cujomalainey:
@marc-hb thanks for the find, indeed if we are trusting IPC on hw limits then we are not secure

zephyr/lib/alloc.c
cujomalainey: @lyakh this isn't theoretical based on the find by @marc-hb in the fuzzer logs

andyross:
Nice.  Frustrating that k_panic() seems not to appear in the fuzzing backtrace.   I'm guessing that's a misfeature in native_posix where it gets "handled" via an exit(), when what we really want is for it to do something (like crash or abort) that is visible to the framework as an error to be dumped.

marc-hb:
A quick fix seems very unlikely so I filed new issue #8832, please move the discussion there.

PR #8633, opened 2023-12-15
up_down_mixer: sink/source refactor

pjdobrowolski:
For needs of loadable modules up_down_mixer needed to switch from obsolete audio_stream to sink/source API interface.

### DIFF
--- a/src/audio/up_down_mixer/up_down_mixer.c
+++ b/src/audio/up_down_mixer/up_down_mixer.c
@@ -401,44 +401,53 @@ err:
 
 static int
 up_down_mixer_process(struct processing_module *mod,
-		      struct input_stream_buffer *input_buffers, int num_input_buffers,
-		      struct output_stream_buffer *output_buffers, int num_output_buffers)
+		      struct sof_source **input_buffers, int num_input_buffers,
+		      struct sof_sink **output_buffers, int num_output_buffers)
 {
 	struct up_down_mixer_data *cd = module_get_private_data(mod);
 	struct comp_dev *dev = mod->dev;
-	uint32_t source_bytes, sink_bytes;
-	uint32_t mix_frames;
+
+	size_t output_frames, input_frames, ret, input_cirbuf_size, output_cirbuf_size;
+	const uint8_t *input0_pos, *input0_start;
+	uint8_t *output_pos, *output_start;
 
 	comp_dbg(dev, "up_down_mixer_process()");
 
-	mix_frames = audio_stream_avail_frames(mod->input_buffers[0].data,
-					       mod->output_buffers[0].data);
+	output_frames = sink_get_free_frames(output_buffers[0]);
+	input_frames = source_get_data_frames_available(input_buffers[0]);
 
-	source_bytes = mix_frames * audio_stream_frame_bytes(mod->input_buffers[0].data);
-	sink_bytes = mix_frames * audio_stream_frame_bytes(mod->output_buffers[0].data);
+	const size_t output_frame_bytes = sink_get_frame_bytes(output_buffers[0]);
 
-	if (source_bytes) {
-		uint32_t sink_sample_bytes;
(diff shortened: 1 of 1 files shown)


src/audio/up_down_mixer/up_down_mixer.c
```diff
@@ -401,44 +401,52 @@ static int up_down_mixer_init(struct processing_module *mod)
 
 static int
 up_down_mixer_process(struct processing_module *mod,
-		      struct input_stream_buffer *input_buffers, int num_input_buffers,
-		      struct output_stream_buffer *output_buffers, int num_output_buffers)
+		      struct sof_source **input_buffers, int num_input_buffers,
```
btian1, line 404: please add sof_source sof_sink related header file into source header:
There should have: #include <sof/audio/sink_api.h> etc.

src/audio/up_down_mixer/up_down_mixer.c
```diff
@@ -401,44 +401,53 @@ static int up_down_mixer_init(struct processing_module *mod)
 
 static int
```
kv2019i, line 402: Some leftoevers in the git commit message ("update" after signed-off).

src/audio/up_down_mixer/up_down_mixer.c
```diff
@@ -401,44 +401,53 @@ static int up_down_mixer_init(struct processing_module *mod)
...
-		audio_stream_copy_from_linear(cd->buf_out, 0, mod->output_buffers[0].data, 0,
-					      sink_bytes / sink_sample_bytes);
-		mod->output_buffers[0].size = sink_bytes;
-		mod->input_buffers[0].consumed = source_bytes;
+	ret = source_get_data(input_buffers[0], input_frames * input0_frame_bytes,
+			      (const void **)&input0_pos, (const void **)&input0_start,
+			      &input_cirbuf_size);
+	if (ret) {
+		sink_commit_buffer(output_buffers[0], 0);
+		return -ENOMEM;
```
kv2019i, line 433: ENODATA? (ENOMEM is typically used for fatal allocation failures while ENODATA is for more transient no data available now error conditions)

kv2019i (commented):
Minor nit in commit message and one error code, but looks good otherwise.

src/audio/up_down_mixer/up_down_mixer.c
pjdobrowolski: There wasn't any audio_stream.h include so I don't see need adding sink_api.h

src/audio/up_down_mixer/up_down_mixer.c
```diff
@@ -401,44 +401,53 @@ static int up_down_mixer_init(struct processing_module *mod)
...
+	cd->mix_routine(cd, (const void *)input0_start, input_cirbuf_size, (void *)output_start);
+
+	ret = sink_commit_buffer(output_buffers[0], output_frames * output_frame_bytes);
+	if (ret)
+		return ret;
+
+	ret = source_release_data(input_buffers[0], input_frames * input0_frame_bytes);
+	if (ret)
+		return ret;
 	return 0;
```
lyakh, line 445: 4 lines can be replaced with `return source_release_data(input_buffers[0], input_frames * input0_frame_bytes);`

lgirdwood: Ack, this is a nice simplification, lets take it incrementally though as this has passed CI (and it would be good to get this better tested in main) before any CI downtime over the vacation.

lgirdwood (approved):

src/audio/up_down_mixer/up_down_mixer.c
btian1: this is not a reason, we can't make mistakes continue, previous did not add means we lose maintain on this(many reasons), now, we noticed this, we can change this step by step to make each source file correct.

PR #8634, opened 2023-12-15
Enabling probe feature in AMD Platforms

DINESHKUMARK1:
Enabling probe feature on Vangogh and Rembrandt Platforms.


sofci:

kv2019i:
@DINESHKUMARK1 Can you check the CI first, lot of build errors across multiple targets.

src/drivers/amd/common/acp_dma.c
```diff
@@ -22,10 +22,15 @@
 #include <stddef.h>
```
dbaluta, line 22: Can you please add more details about what 'probe' functionality is?

Is the ability to get some statistics/logs  via compress interface?

DINESHKUMARK1: Hi,
This is a debug feature to probe the data in between components of pipeline, for more information refer below 
 https://thesofproject.github.io/latest/developer_guides/debugability/probes/index.html

src/drivers/amd/common/acp_dma.c
```diff
@@ -218,7 +246,18 @@ static int acp_dma_copy(struct dma_chan_data *channel, int bytes, uint32_t flags
 	while (ch_sts.bits.dmachrunsts & dmach_mask)
 		ch_sts = (acp_dma_ch_sts_t)dma_reg_read(channel->dma, ACP_DMA_CH_STS);
 	ret = acp_dma_stop(channel);
-
+	if (ret >= 0)
+		if (acp_dma_chan->config[channel->index].probe_channel == channel->index) {
```
lyakh, line 250: doing `if (ret >= 0 && acp_dma_chan->config[channel->index].probe_channel == channel->index) {` would reduce indentation depth

src/drivers/amd/common/acp_dma.c
```diff
@@ -218,7 +246,18 @@ static int acp_dma_copy(struct dma_chan_data *channel, int bytes, uint32_t flags
...
+			probe_pos_update += bytes;
+			probe_pos += bytes;
+			if (probe_pos >= (PROBE_BUFFER_WATERMARK))	{
+				io_reg_write((PU_REGISTER_BASE +
+					ACP_FUTURE_REG_ACLK_0),
+						(PROBE_UPDATE_POS_MASK | probe_pos_update));
```
lyakh, line 256: 3 instances of superfluous parentheses on 4 lines

src/lib/notifier.c
```diff
@@ -109,6 +109,9 @@ void notifier_unregister(void *receiver, void *caller, enum notify_id type)
 		    (!caller || handle->caller == caller)) {
 			if (!--handle->num_registrations) {
 				list_item_del(&handle->list);
+				#ifdef CONFIG_PROBE
+					list_item_del(&notify->list[NOTIFIER_ID_BUFFER_PRODUCE]);
+				#endif
```
lyakh, line 114: Presumably this would be a bug fix. Are calls to
```
	notifier_unregister(_probe, dev->cb, NOTIFIER_ID_BUFFER_PRODUCE);
```
in probe.c not enough?

If this is indeed needed, then please move `#ifdef` / `#endif` to position 0, and reduce indentation of `list_item_del()`

src/drivers/amd/vangogh/acp_dma.c
```diff
@@ -242,9 +242,12 @@ int dma_setup(struct dma_chan_data *channel,
...
 					dma_config_dscr[dscr_strt_idx].src_addr |
 					ACP_DRAM_ADDR_TRNS;
-				dma_cfg->wr_size = dma_cfg->size;
 				dma_cfg->rd_size = 0;
-			}
+				if (channel->index == dma_cfg->probe_channel)
+					dma_cfg->wr_size = 0;
+				else
+					dma_cfg->wr_size = dma_cfg->size;
+				}
```
lyakh, line 250: wrong indentation

src/ipc/dma-copy.c
```diff
@@ -184,8 +184,6 @@ int dma_copy_new(struct dma_copy *dc)
 	return 0;
 }
 
-#if CONFIG_DMA_GW
-
```
btian1, line 188: this config is not AMD specific, if AMD platform want to use this, better to add this config to AMD build?

src/drivers/amd/common/acp_dma.c
```diff
@@ -49,6 +54,7 @@ void dma_config_descriptor(uint32_t dscr_start_idx, uint32_t dscr_count,
 static struct dma_chan_data *acp_dma_channel_get(struct dma *dma,
 						 unsigned int req_chan)
 {
+	uint32_t *p = NULL;
```
btian1, line 57: uint32_t *p = NULL; -=-> uint32_t *p;

src/drivers/amd/common/acp_dma.c
```diff
@@ -73,6 +79,18 @@ static struct dma_chan_data *acp_dma_channel_get(struct dma *dma,
 	acp_dma_chan->config[req_chan].rd_size = 0;
 	acp_dma_chan->config[req_chan].wr_size = 0;
 	acp_dma_chan->config[req_chan].size = 0;
+	acp_dma_chan->config[req_chan].probe_channel = 0xFF;
+	p = dma->priv_data;
+	if (p) {
```
btian1, line 84: may be can delete p, directly use dma->priv_data

src/ipc/dma-copy.c
DINESHKUMARK1: Enabling CONFIG_DMA_GW will bring in many other API's which will cause failures in DMA operations  and are not required on AMD platform, but dma_copy_set_stream_tag is required for enabling DMA channel based on stream tag, That  is the reason to remove CONFIG_DMA_GW for this API, I believe this won't affect non AMD platforms.

lgirdwood: yeah, this is fine - its mostly generic if anyone is using a DMA tag to identify a DMA channel.

src/lib/notifier.c
```diff
@@ -109,6 +109,9 @@ void notifier_unregister(void *receiver, void *caller, enum notify_id type)
 		    (!caller || handle->caller == caller)) {
 			if (!--handle->num_registrations) {
 				list_item_del(&handle->list);
+#ifdef CONFIG_PROBE
```
lgirdwood, line 112: oh, can you confirm we have the opposite list_add() call for this notifier using the same #ifdef CONFIG_PROBE kconfig ?

src/drivers/amd/common/acp_dma.c
lgirdwood: @dbaluta its already used on Intel platforms. Kernel parts also upstream.

lgirdwood (commented):
LGTM, just 1 question on notifier.

src/lib/notifier.c
DINESHKUMARK1: only for list_item_del(&notify->list[NOTIFIER_ID_BUFFER_PRODUCE]) is used as it is bug for probe buffer to avoid issues for regulator use cases

bhiregoudar (approved):

src/probe/probe.c
```diff
@@ -132,7 +132,6 @@ static int probe_dma_buffer_init(struct probe_dma_buf *buffer, uint32_t size,
  */
```
kv2019i, line 132: The git commit prefixes are not quite correct. SOF follows same style as in Linux kernel and git commit summary line should have a prefix that matches the code that is changed. E.g. for this commit, this should be "probe: local dma_sg_config is no more needed"

Similarly for other commits. Please see "git log src/path/to/file.c" for examples from older commits.

src/lib/notifier.c
kv2019i: There is probably a real bug here, but this does not look a correct fix. The probe code does notify unregister for NOTIFIER_ID_BUFFER_PRODUCE and NOTIFIER_ID_BUFFER_FREE, so the unconditional list_item_del seems out of place here.

kv2019i (commented):
The series looks mostly good, but the git commit prefixes are not correct and the bugfix in "deregister NOTIFIER_ID_BUFFER_PRODUCE" does not look correct. I won-t leave a -1 as I'm OOO next week, but I think these needs to be addressed before merge.

src/lib/notifier.c
lgirdwood: Quick look at the API means we need to call ```notifier_unregister()``` and pass in ```NOTIFIER_ID_BUFFER_PRODUCE``` as type.

DINESHKUMARK1: Updated the patch with proper fix and removed the macro config_probe.

kv2019i (approved):
Thanks @DINESHKUMARK1 , looks good now!

kv2019i:
SOFCI TEST

kv2019i:
@wszypelt Can you check the quickbuild fails, 5 days with no results..?

wszypelt:
@kv2019i It seems that the request was not queued, anyway I turned it on manually, the results should be available within an hour

PR #8636, opened 2023-12-15
DP: LL module should get 1ms data chunk in each cycle

marcinszkudlinski:
DP may produce a huge chunk of output data (i.e. 10 LL cycles), and the following module should be able to consume it in 1 cycle chunks, one by one

unfortunately LL modules are designed to drain input buffer
That leads to issues when DP provide huge data portion

FIX: copy only the following module's IBS in each LL cycle

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -1053,13 +1053,24 @@ static int module_adapter_copy_dp_queues(struct comp_dev *dev)
 	list_for_item(blist, &dev->bsink_list) {
 		/* output - we need to copy data from dp_queue (as source)
 		 * to audio_stream (as sink)
+		 *
+		 * a trick is needed there:
+		 * DP may produce a huge chunk of output data (i.e. 10 LL cycles), and the
+		 * following module should be able to consume it in 1 cycle chunks, one by one
+		 *
+		 * unfortunately LL modules are designed to drain input buffer
+		 * That leads to issues when DP provide huge data portion
+		 *
+		 * FIX: copy only the following module's IBS in each LL cycle
 		 */
 		assert(dp_queue);
 		struct comp_buffer *buffer =
 				container_of(blist, struct comp_buffer, source_list);
 		struct sof_sink *data_sink = audio_stream_get_sink(&buffer->stream);
+		struct sof_source *following_mod_data_source =
+				audio_stream_get_source(&buffer->stream);
 		struct sof_source *data_src = dp_queue_get_source(dp_queue);
-		uint32_t to_copy = MIN(sink_get_free_size(data_sink),
+		uint32_t to_copy = MIN(source_get_min_available(following_mod_data_source),
 				       source_get_data_available(data_src));
 
 		err = source_to_sink_copy(data_src, data_sink, true, to_copy);


src/audio/module_adapter/module_adapter.c
```diff
@@ -1053,13 +1053,24 @@ static int module_adapter_copy_dp_queues(struct comp_dev *dev)
...
 		 */
 		assert(dp_queue);
 		struct comp_buffer *buffer =
 				container_of(blist, struct comp_buffer, source_list);
 		struct sof_sink *data_sink = audio_stream_get_sink(&buffer->stream);
+		struct sof_source *following_mod_data_source =
+				audio_stream_get_source(&buffer->stream);
 		struct sof_source *data_src = dp_queue_get_source(dp_queue);
-		uint32_t to_copy = MIN(sink_get_free_size(data_sink),
+		uint32_t to_copy = MIN(source_get_min_available(following_mod_data_source),
```
btian1, line 1073: there are two possible data listed in DP output buffer:
1. copy 1 LL cycle, then keep read ptr as current position, next time will start from this read ptr.
2. once copy 1 LL cycle from DP output buffer, then move data from current to the beginning of the buffer.
 
2 will give LL modules always fetch data from beginning, i.e, linear buffer,  what do you think about this?

mwasko:
@marcinszkudlinski can you elaborate what issues are observed when DP provide huge data portion to next LL in path? One of the obvious problems I expect is that it will lead to MCPS spike on the following LL component that may result in LL scheduling delay, but are there any other issues?

Generally to reduce latency it might be beneficial to schedule an opportunistic thread with low priority, that in each period will continue processing if there is still any data left in the LL input buffers (and cycles left on CPU). This is of course a separate topic if is a feasible option.

mwasko (approved):

marcinszkudlinski:
@mwasko yes, in addition to load spikes - and possible gaps in data at last point, when LL modules cannot process bigger data portion in 1ms - there are additional problems: 
 - some modules are not prepared for such "fast patch" - i.e. overruns and data drops were reported by SSP DAI
 - if whole 10ms is processed in one step - what's going to happen in next 9 cycles before new data arrive? There'll be no data to process. Some modules are also not ready for his and i.e. are flooding log with error messages "no data"

jxstelter (approved):

src/audio/module_adapter/module_adapter.c
lgirdwood: @marcinszkudlinski @andrula-song I assume this will always give us a size that is aligned to hifi2/3/4/5 SIMD preferred data stride ?

lgirdwood (commented):
LGTM, just 1 Q to confirm.

src/audio/module_adapter/module_adapter.c
andrula-song: from the code it is from IBS/OBS, if we don't set restrictions in tplg, there may be potential issues
`	/*
	 * set min_free_space and min_available in sink/src api of created buffer.
	 * buffer is connected like:
	 *	source_module -> (sink_ifc) BUFFER (source_ifc) -> sink_module
	 *
	 *	source_module needs to set its OBS (out buffer size)
	 *		as min_free_space in buffer's sink ifc
	 *	sink_module needs to set its IBS (input buffer size)
	 *		as min_available in buffer's source ifc
	 */
	sink_set_min_free_space(audio_stream_get_sink(&buffer->stream), source_src_cfg.obs);
	source_set_min_available(audio_stream_get_source(&buffer->stream), sink_src_cfg.ibs);`

marcinszkudlinski:
> @marcinszkudlinski @andrula-song I assume this will always give us a size that is aligned to hifi2/3/4/5 SIMD preferred data stride ?

it is independent from this. It is just copying no more than IBS of a module, not to full size of a buffer.

Issue #8630, opened 2023-12-14
[FEATURE] organize audio module source code to per-module directories


kv2019i:
**Is your feature request related to a problem? Please describe.**
Currently audio module source files are a bit spread around the codebase. Clean up the codebase to so that all key files and definitions for a module are in a single place.

**Describe the solution you'd like**

1) Per module directory which contains its own headers, Kconfig and CMakefile.

2) Any metadata should be in this folder (.e.g IPC4 toml file sections to describe the cycle budgets)

3) A README.MD that describes module features, capabilities and performance.

cc:
- #8260
- #8606

kv2019i:
Related PRs
- https://github.com/thesofproject/sof/pull/8587
- https://github.com/thesofproject/sof/pull/8557
- https://github.com/thesofproject/sof/pull/8514
- https://github.com/thesofproject/sof/pull/8595

kv2019i:
One open issue is how to tackel modules that build on each others functionality directly (like crossover, drc and mutliband-drc). Reusing functionality is something we want to promite as well, so the effort to isolate modules should not make this harder.

marc-hb:
`zephyr_library()` and `zephyr_library_include_directories()` is what Zephyr seems to use to solve this problem. 

```
west grep zephyr.*include_dir | wc -l
805

west manifest --path
zephyr/west.yml
```

I couldn't find documentation about this in https://docs.zephyrproject.org/latest/ but there is a lot in the source of https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/modules/extensions.cmake



cc: @andyross, @cujomalainey , @johnylin76 , @btian1 , @lyakh , @singalsu , @RanderWang

btian1:
@marc-hb ， if we use zephyr syntax， how about non-zephyr case？ I mean CONFIG_ZEPHYR_NATIVE_DRIVERS not defined?

marc-hb:
> how about non-zephyr case？ I mean CONFIG_ZEPHYR_NATIVE_DRIVERS not defined?

Then we need an indirection like the one used in
- #8548

lgirdwood:
> @marc-hb ， if we use zephyr syntax， how about non-zephyr case？ I mean CONFIG_ZEPHYR_NATIVE_DRIVERS not defined?
> 
> There are three ways to cover this:
> 
> 1. keep current ../../
> 2. Marc's proposal.
> 3. move those common to a unify header and store it in math folder.

[...]

PR #8652, opened 2023-12-20
Crossover refactor

btian1:
1. move crossover and multiband DRC common part to math include path.
2. split ipc3 and ipc4 specific code.


src/audio/CMakeLists.txt
```diff
@@ -204,7 +206,7 @@ set(dcblock_sources dcblock/dcblock.c dcblock/dcblock_generic.c dcblock/dcblock_
 set(crossover_sources crossover/crossover.c crossover/crossover_generic.c)
 set(tdfb_sources tdfb/tdfb.c tdfb/tdfb_generic.c tdfb/tdfb_direction.c)
 set(drc_sources drc/drc.c drc/drc_generic.c drc/drc_math_generic.c)
-set(multiband_drc_sources multiband_drc/multiband_drc_generic.c crossover/crossover.c crossover/crossover_generic.c drc/drc.c drc/drc_generic.c drc/drc_math_generic.c multiband_drc/multiband_drc.c )
+set(multiband_drc_sources multiband_drc/multiband_drc_generic.c crossover/crossover.c drc/drc.c drc/drc_generic.c drc/drc_math_generic.c multiband_drc/multiband_drc.c )
```
singalsu, line 209: Not sure it fits scope of this PR but eventually the crossover/crossover.c content should move to maths library where it's enabled by kconfig if crossover or multiband-drc component is built. I think the current sources linking duplicates the crossover .text and consumes more RAM.

singalsu (approved):

src/audio/CMakeLists.txt
singalsu: Not all of crossover.c (e.g. prepare and other component ops) would be in maths library but only the common filter related functions those are really shared.

src/include/sof/math/crossover_common.h
```diff
@@ -0,0 +1,84 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright(c) 2023 Intel Corporation. All rights reserved.
+ *
+ * Author: Baofeng Tian <baofeng.tian@intel.com>
```
lgirdwood, line 5: pls also preserve the original author as this is this will be mostly copy and paste.

src/include/kernel/abi.h
```diff
@@ -29,7 +29,7 @@
 
 /** \brief SOF ABI version major, minor and patch numbers */
 #define SOF_ABI_MAJOR 3
-#define SOF_ABI_MINOR 29
+#define SOF_ABI_MINOR 30
```
lgirdwood, line 32: Why is this changing ABI ?

lgirdwood (commented):
Just one open with the ABI

src/include/sof/math/crossover_common.h
marc-hb: I hope you meant "cut and paste". It would be sad for a refactoring PR to add copy/paste :-)

src/audio/crossover/crossover.c
```diff
@@ -11,8 +11,8 @@
 #include <sof/audio/format.h>
 #include <sof/audio/pipeline.h>
 #include <sof/audio/ipc-config.h>
+#include <sof/math/crossover_common.h>
```
kv2019i, line 14: I'm not a fan of moving crossover_common.h under math. See my rationale at https://github.com/thesofproject/sof/issues/8630#issuecomment-1864940768 

Not blocking on this though if other reviewers are ok.

kv2019i (commented):
Not fan of math folder usage. dct.h and matrix.h fit nicely to math, but putting whole algorithms there. But @singalsu , I do note the iir/fir core headers are there already. so maybe this agreed upon and we should put all "reusable algorithm core" headers here. I won't block the way...

src/audio/crossover/crossover.c
btian1: this is a intermediate solution, Seppo and I will continue try to figure out better refactor solution for this, @singalsu @kv2019i

src/include/kernel/abi.h
btian1: I saw warnings when commit this patch if not changing ABI, warnings list below:
btian1@btian1-linux:~/sof_zephyr_1012/sof$ git commit -s
No codespell typos will be found - file '/usr/share/codespell/dictionary.txt': No such file or directory
WARNING: Please update ABI in accordance with http://semver.org

src/include/sof/math/crossover_common.h
btian1: sorry, I made mistake here, updating.

src/audio/CMakeLists.txt
btian1: I tried to figure out yesterday, however, the common part is prepare / params, not actually the processing functions, not easy to move filter related to math, you can check crossover_common.h
all listed are required by multiband_drc.

Regarding linking, I searched sof assembly, take crossover.c:
static void crossover_s32_default(struct comp_data *cd,
				  struct input_stream_buffer *bsource,
				  struct output_stream_buffer **bsinks,
				  int32_t num_sinks,
				  uint32_t frames)

as example, it only have one time results.

also for crossover_common part, all are embeded as inline, no more duplicated code found in assembly.

src/include/kernel/abi.h
lgirdwood: ok, but tin this case we can ignore since we are not changing the ABI.

src/audio/crossover/crossover.c
lgirdwood: move it to include/module.

lgirdwood (commented):
Lets delete the ABI change and use module include directory for exported public crossover APIs.

src/include/user/crossover.h
```diff
@@ -10,9 +10,7 @@
 
 #include <stdint.h>
 #include <user/eq.h>
-
-/* Maximum Number of sinks allowed in config */
-#define SOF_CROSSOVER_MAX_STREAMS 4
```
kv2019i, line 15: Hmm, this is a bit problematic. The "user/crossover.h" header should not have dependency to implementation parts, it should be independent description of the parameter interface (like user/eq.h).

kv2019i (commented):
One minor note w.r.t. the "user" header. Also, @johnylin76 @cujomalainey please check if this looks reasonable (starts to be ready to merge).

src/include/user/crossover.h
btian1: it was referenced by multiband DRC user header file with:
/* Maximum number of frequency band for Multiband DRC */
#define SOF_MULTIBAND_DRC_MAX_BANDS SOF_CROSSOVER_MAX_STREAMS

src/audio/CMakeLists.txt
btian1: now move it to include/module, seems more reasonable compared with math, since most of common part does not related with math.

lgirdwood (approved):

PR #8653, opened 2023-12-20
schedule: zephyr_ll: Fix schedule bug for multi instances

TangleZ:
When run two instances at same time the sound has noise, for example: playback and record in i.MX8QM or i.MX8ULP platform.

The reason is one dma interrupt will process two task cause the schedule have no mechanism to make sure only handle the task needed to process.

Fix this issue by adding check if task is in pending state.

### DIFF
--- a/src/include/sof/schedule/ll_schedule_domain.h
+++ b/src/include/sof/schedule/ll_schedule_domain.h
@@ -222,7 +222,8 @@ static inline bool domain_is_pending(struct ll_schedule_domain *domain,
 {
 	bool ret;
 
-	assert(domain->ops->domain_is_pending);
+	if (!domain->ops->domain_is_pending)
+		return true;
 
 	ret = domain->ops->domain_is_pending(domain, task, comp);
 
--- a/src/schedule/zephyr_dma_domain.c
+++ b/src/schedule/zephyr_dma_domain.c
@@ -12,6 +12,7 @@
 #include <rtos/alloc.h>
 #include <sof/lib/cpu.h>
 #include <sof/lib/dma.h>
+#include <sof/lib/notifier.h>
 #include <sof/platform.h>
 #include <sof/schedule/ll_schedule.h>
 #include <sof/schedule/ll_schedule_domain.h>
@@ -98,12 +99,15 @@ static int zephyr_dma_domain_register(struct ll_schedule_domain *domain,
 static int zephyr_dma_domain_unregister(struct ll_schedule_domain *domain,
 					struct task *task,
 					uint32_t num_tasks);
+static bool zephyr_dma_domain_is_pending(struct ll_schedule_domain *domain,
+			  struct task *task, struct comp_dev **comp);
 static void zephyr_dma_domain_task_cancel(struct ll_schedule_domain *domain,
 					  struct task *task);
 
 static const struct ll_schedule_domain_ops zephyr_dma_domain_ops = {
 	.domain_register	= zephyr_dma_domain_register,
 	.domain_unregister	= zephyr_dma_domain_unregister,
+	.domain_is_pending	= zephyr_dma_domain_is_pending,
 	.domain_task_cancel	= zephyr_dma_domain_task_cancel
 };
 
@@ -150,6 +154,18 @@ static void zephyr_dma_domain_thread_fn(void *p1, void *p2, void *p3)
 	}
 }
 
+void pipe_task_notify(void *arg, enum notify_id type, void *data)
+{
+	struct pipeline_task *pipe_task = (void *)arg;
+	struct task *task;
+
+	if (!pipe_task)
+		return;
+	task = &pipe_task->task;
+
+	task->state = SOF_TASK_STATE_PENDING;
+}
+
 static void dma_irq_handler(void *data)
 {
 	struct zephyr_dma_domain_irq *irq_data;
@@ -169,8 +185,11 @@ static void dma_irq_handler(void *data)
 	list_for_item(i, &irq_data->channels) {
 		chan_data = container_of(i, struct zephyr_dma_domain_channel, list);
 
-		if (dma_interrupt_legacy(chan_data->channel, DMA_IRQ_STATUS_GET))
+		if (dma_interrupt_legacy(chan_data->channel, DMA_IRQ_STATUS_GET)) {
 			dma_interrupt_legacy(chan_data->channel, DMA_IRQ_CLEAR);
+			notifier_event(chan_data, NOTIFIER_ID_DMA_IRQ,
+				NOTIFIER_TARGET_CORE_LOCAL, NULL, 0);
+		}
 	}
 
 	/* clear IRQ - the mask argument is unused ATM */
@@ -350,6 +369,9 @@ static int register_dma_irq(struct zephyr_dma_domain *domain,
 			*irq_data = crt_irq_data;
 
 			irq_local_enable(flags);
+
+			notifier_register(pipe_task, chan_data, NOTIFIER_ID_DMA_IRQ,
+					  pipe_task_notify, 0);
 
 			return 0;
 		}
@@ -595,6 +617,19 @@ static int zephyr_dma_domain_unregister(struct ll_schedule_domain *domain,
 	/* TODO: thread abortion logic goes here */
 
 	return 0;
+}
+
+static bool zephyr_dma_domain_is_pending(struct ll_schedule_domain *domain,
+			  struct task *task, struct comp_dev **comp)
+{
+	struct pipeline_task *pipe_task;
+
+	pipe_task = pipeline_task_get(task);
+
+	if (!pipe_task->registrable)
+		return true;
+	else
+		return (task->state == SOF_TASK_STATE_PENDING);
 }
 
 static void zephyr_dma_domain_task_cancel(struct ll_schedule_domain *domain,
--- a/src/schedule/zephyr_ll.c
+++ b/src/schedule/zephyr_ll.c
@@ -188,6 +188,7 @@ static void zephyr_ll_run(void *data)
 	for (list = sch->tasks.next; !list_is_empty(&sch->tasks); list = sch->tasks.next) {
 		enum task_state state;
 		struct zephyr_ll_pdata *pdata;
+		struct ll_schedule_domain *domain = sch->ll_domain;
 
 		task = container_of(list, struct task, list);
 		pdata = task->priv_data;
@@ -198,12 +199,18 @@ static void zephyr_ll_run(void *data)
 			continue;
 		}
 
-		pdata->run = true;
-		task->state = SOF_TASK_STATE_RUNNING;
+		if (domain_is_pending(domain, task, NULL)) {
+			pdata->run = true;
+			task->state = SOF_TASK_STATE_RUNNING;
+		}
 
 		/* Move the task to a temporary list */
 		list_item_del(list);
 		list_item_append(list, &task_head);
+
+		/* avoid to execute the task is not ready to run */
+		if (task->state != SOF_TASK_STATE_RUNNING)
+			continue;
 
 		zephyr_ll_unlock(sch, &flags);
 
@@ -211,6 +218,7 @@ static void zephyr_ll_run(void *data)
 		 * task's .run() should only return either
 		 * SOF_TASK_STATE_COMPLETED or SOF_TASK_STATE_RESCHEDULE
 		 */
+
 		state = do_task_run(task);
 		if (state != SOF_TASK_STATE_COMPLETED &&
 		    state != SOF_TASK_STATE_RESCHEDULE) {
@@ -237,6 +245,8 @@ static void zephyr_ll_run(void *data)
 				break;
 			}
 		}
+		/* update task state */
+		task->state = state;
 	}
 
 	/* Move tasks back */


iuliana-prodan:
@TangleZ you hear noise when playing & recoding in parallel?
But why this happens, what changed lately?
What introduced this bug?

When the dma_domain for Zephyr was introduced, I believe @LaurentiuM1234 tested this scenario (aplay || arecord). 
Can you confirm @LaurentiuM1234? 
Any idea why this happens now? Why, so far, we didn't need `is_pending` for Zephyr? 
I know for XTOS we have it.

LaurentiuM1234:
> When the dma_domain for Zephyr was introduced, I believe @LaurentiuM1234 tested this scenario (aplay || arecord).
> Can you confirm @LaurentiuM1234?
> Any idea why this happens now? Why, so far, we didn't need `is_pending` for Zephyr?
> I know for XTOS we have it.

I guess this was bound to happen at some point. I did indeed test the case in which we have multiple pipeline tasks and it didn't happen. **I'm assuming** this didn't happen because the buffers were keeping `pipeline_copy()` from overwriting data from the DMA buffer on DAI's side (i.e: the buffers are full of data so you can't copy anymore from sources to sinks). But yeah this is a known flaw of `zephyr_ll` and `zephyr_dma_domain`.

Anyways, I don't believe this solution is the way to go. It doesn't seem to take into account that you can also have non-registrable pipeline tasks which are not bound to a `chan_data` structure. As such, you'd end up with the non-registrable pipeline tasks not being executed at all.

Also, why are you using the notifier API? It doesn't seem necessary?

lgirdwood (commented):
I've not seen this issues on Intel platforms with 8 streams running at the same time with timer based scheduling, so I suspect relates to DMA based scheduling.
However, both streams with DMA scheduling 
1) need to be independent as they will have different time domains.
2) cant be connected due to 1.
3) may need different priorities to ensure no xruns. 
It looks like here you are using the same DMA IRQ to schedule both pipelines ? if so, maybe better to use timer ?

TangleZ:
@lgirdwood thanks for your comments. And I'm not very clear on what you mean about 1 and 2. but for 3, the answer should be no. In my understanding, every instance should have the same priority when running.
And here it can't say "using the same DMA IRQ to schedule both pipelines", based on different types dma, they may have same DMA IRQ(SDMA) or have different DMA IRQ(EDMA).

What I want to do is to make sure we only process the task that is caused by the right DMA irq.

TangleZ:
@LaurentiuM1234 let us use an example to clear the issue: there are two instances the one is playback(edma 0) and the second(edma 1) is record inthe  qm platform. When interrupt 0 causes the schedule run task, there is no code to guarantee only process this task, it also processes the task belonging to the record. And such a situation also happened in interrupt 1.

"It doesn't seem to take into account that you can also have non-registrable pipeline tasks which are not bound to a chan_data structure."
Actually the task which is not bound to a chan_data they haven't implent pending function, so this will not impact.

"Also, why are you using the notifier API? It doesn't seem necessary?"
I'm not sure notifier API is good, we can discuss to improve this.

lgirdwood:
> @lgirdwood thanks for your comments. And I'm not very clear on what you mean about 1 and 2. but for 3, the answer should be no. In my understanding, every instance should have the same priority when running. And here it can't say "using the same DMA IRQ to schedule both pipelines", based on different types dma, they may have same DMA IRQ(SDMA) or have different DMA IRQ(EDMA).
> 
> What I want to do is to make sure we only process the task that is caused by the right DMA irq.

ok, got you - but the DMA scheduler domain logic should have a callback for each DMA irq source to allow correct irq-> pipeline mapping ?

TangleZ:
> > @lgirdwood thanks for your comments. And I'm not very clear on what you mean about 1 and 2. but for 3, the answer should be no. In my understanding, every instance should have the same priority when running. And here it can't say "using the same DMA IRQ to schedule both pipelines", based on different types dma, they may have same DMA IRQ(SDMA) or have different DMA IRQ(EDMA).
> > What I want to do is to make sure we only process the task that is caused by the right DMA irq.
> 
> ok, got you - but the DMA scheduler domain logic should have a callback for each DMA irq source to allow correct irq-> pipeline mapping ?

Yes, you are right. The problem is as I reply to LaurentiuM1234: "There is no code to guarantee only process the right task". For there are no issues found on Inter platform, I guess the reason is the same as our iMX8MP platform which use SDMA and processes the task that have no data to copy nicely.

src/schedule/zephyr_dma_domain.c
```diff
@@ -597,6 +619,15 @@ static int zephyr_dma_domain_unregister(struct ll_schedule_domain *domain,
 	return 0;
 }
 
+static bool zephyr_dma_domain_is_pending(struct ll_schedule_domain *domain,
+			  struct task *task, struct comp_dev **comp)
+{
+	if (task->state == SOF_TASK_STATE_PENDING)
+		return true;
+	else
+		return false;
```
lyakh, line 628: `return task->state == SOF_TASK_STATE_PENDING;`

src/schedule/zephyr_ll.c
```diff
@@ -198,44 +199,51 @@ static void zephyr_ll_run(void *data)
...
+				switch (task->state) {
+				case SOF_TASK_STATE_CANCEL:
+					zephyr_ll_task_done(sch, task);
+					break;
+				default:
+					break;
+				}
 			}
+			/* update task state */
+			task->state = state;
```
lyakh, line 246: There are a few changes in this commit, including this one, that modify behaviour of the (pipeline) task scheduler which is very central to SOF. If anything this will need a thorough review and extensive testing.

src/include/sof/schedule/ll_schedule_domain.h
```diff
@@ -222,7 +222,8 @@ static inline bool domain_is_pending(struct ll_schedule_domain *domain,
 {
 	bool ret;
 
-	assert(domain->ops->domain_is_pending);
+	if (!domain->ops->domain_is_pending)
+		return true;
```
lyakh, line 226: you add a `.domain_is_pending` method to the DMA scheduling domain, so presumably this change isn't needed?

src/schedule/zephyr_ll.c
lyakh: ...and we have a failure in https://sof-ci.01.org/sofpr/PR8653/build1331/devicetest/index.html?model=MTLP_RVP_NOCODEC&testcase=multiple-pipeline-playback

lyakh: and multiple failures in https://sof-ci.01.org/sofpr/PR8653/build1332/devicetest/index.html

src/schedule/zephyr_ll.c
```diff
@@ -198,44 +199,51 @@ static void zephyr_ll_run(void *data)
...
-		if (state != SOF_TASK_STATE_COMPLETED &&
-		    state != SOF_TASK_STATE_RESCHEDULE) {
-			tr_err(&ll_tr,
-			       "zephyr_ll_run: invalid return state %u",
-			       state);
-			state = SOF_TASK_STATE_RESCHEDULE;
-		}
-
-		zephyr_ll_lock(sch, &flags);
+		if (task->state == SOF_TASK_STATE_RUNNING) {
```
kv2019i, line 211: The diff is a bit hard to read but I can't spot any changes in the function. So could this be nicer to read with:

```
if (task->state != SOF_TASK_STATE_RUNNING)
     continue;
```

kv2019i (commented):
Hmm, I guess this is ok. The diff in zephyr_ll.c looks a bit scary in github view, but as far as I can tell, this should be a no-op for timer-driven use-cases.

Not 100% how suitable the is_pending() interface is to implement this. I think originally this was in the domain interface to cater for one-shot and delayed tasks. Now with move to Zephyr, the recommendation is to use native Zephyr interfaces to do delayed processing and what remains in the LL scheduler is really about running the LL tasks every timer tick and the domain is really only used to create manage the timer thread. 

So in that context, checking of is_pending() is a bit of waste of DSP cycles for the timer driven case.

For DMA driven case,  in theory it should be enough to run the all the tasks once per LL tick (whatever is the first DMA IRQ that fires). But alas, with current arch for DMA driven scheduling, it's not so easy to elect the "main IRQ". This would have to be the IRQ with the highest rate, and if a task is removed, a new "main IRQ" would have to be elected.

So with that, maybe this is the best approach in the end.

src/schedule/zephyr_ll.c
TangleZ: > The diff is a bit hard to read but I can't spot any changes in the function. So could this be nicer to read with:
> 
> ```
**> if (task->state != SOF_TASK_STATE_RUNNING)
>      continue;**
This should better to read.
> ```

Hi @kv2019i , thanks for your review the patches.

src/schedule/zephyr_dma_domain.c
TangleZ: Thanks for the review. Will change this.

src/schedule/zephyr_ll.c
TangleZ: Yes, we need a lot of testing about this change.
About the failures, I can't get any info about what make the CI fail, can you please clarify it to me?

src/schedule/zephyr_ll.c
```diff
@@ -237,6 +244,8 @@ static void zephyr_ll_run(void *data)
 				break;
 			}
 		}
+		/* update task state */
+		task->state = state;
```
kv2019i, line 248: I'm wondering if it's this addition on L247-248 that is causing some unexpected failures. It seems test configurations that use the chain-dma feature (basicly bypasses creation of full audio pipelines and uses a simple copy LL task to push data through two interfaces). I can see DSP panics in configurations where chain-dma is enabled:
https://sof-ci.01.org/sofpr/PR8653/build1416/devicetest/index.html

kv2019i (commented):
Thanks, code looks clean to me now. But, but, tests in CI still fail, it would seem to be linked to one particular user of the LL scheduler, the chain-dmai. See my comment inline.

dbaluta:
@TangleZ can you also please test mixer scenarios on 8qxp or 8qm? I'm heading out for christmas holidays but will follow this PR.

src/schedule/zephyr_ll.c
TangleZ: According to the CI I can't see any info about fails in my side:
_TEST #{{ internalResultID }}
CI build start time:
On-device test with:
Kernel Build Info
SOF Build Info
PR ID: {{ planResult.linuxPrID }}
Linux Branch: {{ planResult.linuxBranch }}
Commit: Merge {{ shortenCommit(planResult.linuxParents.head) }} into {{ shortenCommit(planResult.linuxParents.target) }} ⇒ {{ shortenCommit(planResult.linuxSource.linuxCommit) }}
Kconfig Branch: {{ shortenCommit(planResult.kconfigBranch) }}
Kconfig Commit: {{ shortenCommit(planResult.linuxSource.kconfigCommit) }}
PR ID: {{ planResult.sofPrID }}
SOF Branch: {{ planResult.sofBranch }}
Commit: Merge {{ shortenCommit(planResult.sofParents.head) }} into {{ shortenCommit(planResult.sofParents.target) }} ⇒ {{ shortenCommit(planResult.sofSource.sofCommit) }}
Zephyr Commit: {{ shortenCommit(planResult.sofSource.zephyrCommit) }}
Copy Link_

Can you please help me to understand where show DSP panics and I will try to fix it.

TangleZ:
> @TangleZ can you also please test mixer scenarios on 8qxp or 8qm? I'm heading out for christmas holidays but will follow this PR.

Sure. Can you please send me the test cmd or where I can find it?

TangleZ:
@dbaluta I fixed the issue with the mixer by checking if task is registerable in pending function.

src/schedule/zephyr_ll.c
lgirdwood: @TangleZ can you refresh and enable scripts in your browser for this site, it should then show you logs.

src/schedule/zephyr_ll.c
```diff
@@ -198,19 +199,25 @@ static void zephyr_ll_run(void *data)
...
+		if (domain_is_pending(domain, task, NULL)) {
+			pdata->run = true;
+			task->state = SOF_TASK_STATE_RUNNING;
+		}
 
 		/* Move the task to a temporary list */
 		list_item_del(list);
 		list_item_append(list, &task_head);
 
+		if (task->state != SOF_TASK_STATE_RUNNING)
```
lgirdwood, line 211: ditto, re inline comments - scheduler can be complex so lets add comments to help.

src/schedule/zephyr_ll.c
kv2019i: @TangleZ Sorry for late response, out for xmas. Here's example log showing a panic:
https://sof-ci.01.org/sofpr/PR8653/build1416/devicetest/index.html?model=TGLU_UP_HDA-ipc4&testcase=multiple-pause-resume-50

This is running "~/sof-test/test-case/multiple-pause-resume.sh -r 50" on a Intel TGL based laptop. The panic hits when the test uses the HDMI/DP PCMs that use chain-dma.

src/schedule/zephyr_ll.c
TangleZ: Done.

TangleZ: Also, I can see the fails in CI, and it's related intel test. But I can't do the test in my side cause we have no intel board.
Can you please help? I also will look the patch to see if any I can change to fix the CI fails.

kv2019i: Ack, @TangleZ I'll try to debug this a bit early next week, agreed this is difficult to debug without a board available.

src/schedule/zephyr_dma_domain.c
```diff
@@ -597,6 +619,19 @@ static int zephyr_dma_domain_unregister(struct ll_schedule_domain *domain,
...
 }
 
+static bool zephyr_dma_domain_is_pending(struct ll_schedule_domain *domain,
+			  struct task *task, struct comp_dev **comp)
+{
+	struct pipeline_task *pipe_task;
+
+	pipe_task = pipeline_task_get(task);
+
+	if (!pipe_task->registrable)
```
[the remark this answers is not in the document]
TangleZ: We did have a workaround for this issue, but it not fix the root cause. As I said before, the root cause is schedule bug.
If you think don't need this patch, we'd better switch to native interface + timer domain in next release.

src/schedule/zephyr_ll.c
TangleZ: Thanks for your help!

dbaluta:
We have an workaround for this. We will have a proper fix with the next release. The ideal solution is to align with Intel on this and use timer domain.

PR #8656, opened 2023-12-20
iadk: module_adapter: Check return code for sink/source api

jxstelter:
Add check of return code for sink/source API functions.
The IADK processing module interface uses uint32_t return code
while cSOF is using int type. Checks return value
from processing() method and returns -ENODATA in case of
failure.

### DIFF
--- a/src/audio/module_adapter/iadk/iadk_module_adapter.cpp
+++ b/src/audio/module_adapter/iadk/iadk_module_adapter.cpp
@@ -39,12 +39,12 @@ int IadkModuleAdapter::IadkModuleAdapter_Prepare(void)
 	return 0;
 }
 
-uint32_t IadkModuleAdapter::IadkModuleAdapter_Process(struct sof_source **sources,
-						      int num_of_sources,
-						      struct sof_sink **sinks,
-						      int num_of_sinks)
+int IadkModuleAdapter::IadkModuleAdapter_Process(struct sof_source **sources,
+						 int num_of_sources,
+						 struct sof_sink **sinks,
+						 int num_of_sinks)
 {
-	uint32_t ret = 0;
+	int ret = 0;
 
 	if ((num_of_sources > 0) && (num_of_sinks > 0)) {
 		intel_adsp::InputStreamBuffer input_stream_buffers[INPUT_PIN_COUNT];
@@ -80,7 +80,14 @@ uint32_t IadkModuleAdapter::IadkModuleAdapter_Process(struct sof_source **source
 			new (&output_stream_buffers[i]) intel_adsp::OutputStreamBuffer(osb_data);
 		}
 
-		ret = processing_module_.Process(input_stream_buffers, output_stream_buffers);
+		uint32_t iadk_ret =
+			processing_module_.Process(input_stream_buffers, output_stream_buffers);
+
+		/* IADK modules returns uint32_t return code. Convert to failure if Process
+		 * not successful.
+		 */
+		if (iadk_ret != 0)
+			ret = -ENODATA;
 
 		for (int i = 0; i < num_of_sources; i++) {
 			source_release_data(sources[i], input_stream_buffers[i].size);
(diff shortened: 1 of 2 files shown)


lgirdwood (approved):

lgirdwood:
@jxstelter CI reporting issue, so could be the return code is now passing an error up the stack.

jxstelter:
> @jxstelter CI reporting issue, so could be the return code is now passing an error up the stack.

CI failure seems to be unrelated. Will restart CI.

src/audio/module_adapter/iadk/iadk_module_adapter.cpp
```diff
@@ -57,6 +57,9 @@ uint32_t IadkModuleAdapter::IadkModuleAdapter_Process(struct sof_source **source
 			i_size = source_get_data_available(sources[i]);
 			ret = source_get_data(sources[i], i_size, (const void **)&input,
 					      (const void **)&input_start, &input_end);
+			if (ret != 0)
```
btian1, line 60: there is type wrong for ret, it should not be uint32_t, it should be int. 
since ret maybe negative value, check source_get_data code below:
https://github.com/thesofproject/sof/blob/main/src/module/audio/source_api.c#L18

lgirdwood (commented):
@jxstelter will this break compatibility with IADK with signed/unsigned return value change ?

jxstelter:
> @jxstelter will this break compatibility with IADK with signed/unsigned return value change ?

There should be no regression with this change. Currently we check for success/failure only and If I am correct, we do not have an error code translation between IADK error codes and standard ones.

lgirdwood (approved):

src/audio/module_adapter/iadk/iadk_module_adapter.cpp
```diff
@@ -80,7 +80,14 @@ uint32_t IadkModuleAdapter::IadkModuleAdapter_Process(struct sof_source **source
 			new (&output_stream_buffers[i]) intel_adsp::OutputStreamBuffer(osb_data);
 		}
 
-		ret = processing_module_.Process(input_stream_buffers, output_stream_buffers);
+		uint32_t iadk_ret =
```
btian1, line 83: it is still be negative value for module process, still need use int.
may be should directly return processing_module_.Process(input_stream_buffers, output_stream_buffers);

jxstelter: No. IADK modules are from reference FW. These are using uint32_t return values. 
See: https://github.com/thesofproject/sof/blob/2a0607c658c8b1c487ac63e8e2b5d3410766069a/src/include/sof/audio/module_adapter/iadk/processing_module_interface.h#L230C1-L243C9

src/audio/module_adapter/iadk/iadk_module_adapter.cpp
```diff
@@ -80,7 +80,14 @@ uint32_t IadkModuleAdapter::IadkModuleAdapter_Process(struct sof_source **source
...
+			processing_module_.Process(input_stream_buffers, output_stream_buffers);
+
+		/* IADK modules returns uint32_t return code. Convert to failure if Process
+		 * not successful.
+		 */
+		if (iadk_ret != 0)
+			ret = -ENODATA;
```
kv2019i, line 90: Not really for this PR, but I wonder if it's ok to sink_commit_buffer() with full "output_stream_buffers[i].size" in case of error.

kv2019i (approved):
Some minor nits, but the change itself looks good.

src/audio/module_adapter/iadk/iadk_module_adapter.cpp
jxstelter: Yes it is question whether we plan to recover in case of failure in other way than just restart pipeline. 
Currently we do not recover anyway:
https://github.com/thesofproject/sof/blob/b617388c622f29250221d309c20241c4953efd41/src/audio/pipeline/pipeline-schedule.c#L380C1-L388C2

PR #8490, opened 2023-11-17
toml: modularise TOML configuration

lyakh:
Split TOML configuration files into platform and module parts. Use the C preprocessor to merge them back together.

PR #8660, opened 2023-12-21
rimage: mtl: add copier and eq-iir CPC cases

btian1:
Copier missed ibs(384)/obs(192) case, and eq-iir missed ibs(384)/obs(384) case, adding it to current CPC config.


kv2019i (approved):
Thank you @btian1 for fast update! Can you fix the commit prefix to "rimage: mtl: add copier and eq-iir CPC cases"

This will be racing with https://github.com/thesofproject/sof/pull/8490 , one needs to be udpated after another, FYI @lyakh

kv2019i (changes requested):
Wrong button, adding request changes to fix the git commit.

lgirdwood:
> Thank you @btian1 for fast update! Can you fix the commit prefix to "rimage: mtl: add copier and eq-iir CPC cases"
> 
> This will be racing with #8490 , one needs to be udpated after another, FYI @lyakh

@btian1  fyi TOML PR now merged

singalsu (approved):
Thanks! I can't decode the numbers in my head but looks assuring. Yep I know you got the numbers from tool so they are right :)

kv2019i (approved):
Thanks!

PR #8663, opened 2023-12-21
Pipeline refactor

btian1:


lyakh (changes requested):
No, all these defines, structures and functions are only needed in a single .c file, no need to pull them over in an interface header. This looks like an opposite of "declutter" to me

btian1:
> No, all these defines, structures and functions are only needed in a single .c file, no need to pull them over in an interface header. This looks like an opposite of "declutter" to me

closing, will check in future again, not for now.

PR #8665, opened 2023-12-21
topology2: cavs-nocodec-bt: fix PCM0 and PCM1 capabilities

kv2019i:
The pipelines for PCM0 and PCM1 only support S32_LE audio format. Fix the PCM capability descriptions to match the actual pipeline definitions. This allows to run test suites that enumerate all support PCMs and their formats.


kv2019i:
@keqiaozhang with this, the sof-test plan now completes without errors. The interesting test cases are with PCM2 with this topology. but PCM0 and PCM1 definitions should of course be valid as well.

lgirdwood (approved):

jsarha (approved):

PR #8666, opened 2023-12-21
Preparatory and cosmetic patches for loadable modules

lyakh:
This is a part of #8180 with what should be safe and non-controversial changes. No significant functional changes, no breakage expected.

### DIFF
--- a/src/audio/module_adapter/module/modules.c
+++ b/src/audio/module_adapter/module/modules.c
@@ -62,7 +62,6 @@ static int modules_init(struct processing_module *mod)
 	struct module_data *md = &mod->priv;
 	struct comp_dev *dev = mod->dev;
 	const struct ipc4_base_module_cfg *src_cfg = &md->cfg.base_cfg;
-	int ret = 0;
 	byte_array_t mod_cfg;
 
 	mod_cfg.data = (uint8_t *)md->cfg.init_data;
@@ -136,6 +135,8 @@ static int modules_init(struct processing_module *mod)
 		return -ENOMEM;
 	}
 	md->mpd.out_buff_size = src_cfg->obs;
+
+	int ret;
 
 	/* Call module specific init function if exists. */
 	if (mod->is_native_sof) {
--- a/src/audio/module_adapter/module/modules.c
+++ b/src/audio/module_adapter/module/modules.c
@@ -43,7 +43,7 @@
  *  - Processing Module Adapter - SOF base FW side of ProcessingModuleInterface API
  */
 
-LOG_MODULE_REGISTER(modules, CONFIG_SOF_LOG_LEVEL);
+LOG_MODULE_REGISTER(sof_modules, CONFIG_SOF_LOG_LEVEL);
 /* ee2585f2-e7d8-43dc-90ab-4224e00c3e84 */
 DECLARE_SOF_RT_UUID("modules", intel_uuid, 0xee2585f2, 0xe7d8, 0x43dc,
 		    0x90, 0xab, 0x42, 0x24, 0xe0, 0x0c, 0x3e, 0x84);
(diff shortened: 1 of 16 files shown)


Kconfig.sof
```diff
@@ -207,6 +207,13 @@ config DSP_RESIDENCY_COUNTERS
 	  R0, R1, R2 are abstract states which can be used differently
 	  based on platform implementation.
 
+config MODULES
+	bool "Enable modules"
```
pjdobrowolski, line 211: I think that it is not explanatory enough. Should it be native loadable modules?

pjdobrowolski (changes requested):

Kconfig.sof
lyakh: @pjdobrowolski sure, good idea, thanks, updated

lgirdwood: Yeah, we need to really namespace all our KConfig with ```SOF_``` prefix too, to avoid collisions with Zephyr as we share lots of common names. We are not compliant today.

src/audio/module_adapter/module/modules.c
```diff
@@ -137,6 +136,8 @@ static int modules_init(struct processing_module *mod)
 	}
 	md->mpd.out_buff_size = src_cfg->obs;
 
+	int ret;
+
```
kv2019i, line 140: I know we can technically do this (C99 yay), but style-wise this isn't something we do a lot. Not blocking, but just logging this looks strange to me.

kv2019i (approved):
One style-oddity, but looks good otherwise!

src/audio/module_adapter/module/modules.c
lyakh: @kv2019i I think it's helpful in cases like this one. You define the variable where it's needed. This avoids having a 100 lines of code between the definition and the use where it becomes difficult to verify that the variable is indeed unused all that time (especially if it's some `int i` or even `int ret`) which then prompts the developer to stick in a redundant initialisation because "who knows what that variable is there for."

Kconfig.sof
lyakh: @lgirdwood I think some tooling actually requires `CONFIG_MODULES` to allow the use of `tristate` in Kconfig options. But yes, maybe we need this in Zephyr, I was thinking about that too

lyakh: ...just double-checked: `config MODULES` is indeed needed for kconfiglib, otherwise selecting "m" for `tristate` options doesn't work

lyakh: I pushed https://github.com/zephyrproject-rtos/zephyr/pull/67105 - if it gets accepted in Zephyr, we'll drop it here

Kconfig.sof
```diff
@@ -207,6 +207,13 @@ config DSP_RESIDENCY_COUNTERS
 	  R0, R1, R2 are abstract states which can be used differently
 	  based on platform implementation.
 
+config MODULES
```
lgirdwood, line 210: We really need to namespace these as '''MODULES''' is quite generic and may collide with Zephyr.

lgirdwood (commented):
I think we need to namespace the Kconfig, i.e. with a ```SOF_```I know others dont have this yet, but we need to do this in many places to avoid collisions.

lyakh:
@wszypelt QB failure seems to be an internal connection error? Could you restart please?

lyakh:
CI: style https://github.com/thesofproject/sof/actions/runs/7450293124/job/20268863453?pr=8666 is complaining about `EXPORT_SYMBOL()` and on-device test https://sof-ci.01.org/sofpr/PR8666/build1646/devicetest/index.html has 2 suspend-resume timeouts

Issue #8670, opened 2023-12-22
[BUG] Alsatplg produces incorrect NHLT DMIC blob when second DAI rate is higher than first


singalsu:
**Describe the bug**
Changing DMIC1 rate can corrupt NHLT blob for DMIC0.

**To Reproduce**
Do this simple change to DMIC object. It's not used by topology so it should not impact the used DMIC DAI. Build the topology and copy it to the DUT, reboot.  Use a device that uses NHLT blob from topolopgy (options snd_sof_intel_hda_common sof_use_tplg_nhlt=1) . Capture audio from the supported dai_index 0 DMIC. Rebot

```
$ arecord -Dhw:0,6 -f S32_LE -c 4 -r 48000 -d 5 rec.wav
Recording WAVE 'rec.wav' : Signed 32 bit Little Endian, Rate 48000 Hz, Channels 4
arecord: pcm_read:2221: read error: Invalid argument
```

```
diff --git a/tools/topology/topology2/platform/intel/dmic-generic.conf b/tools/topology/topology2/platform/intel/dmic-generic.conf
index 5ca0ec27c..628e06183 100644
--- a/tools/topology/topology2/platform/intel/dmic-generic.conf
+++ b/tools/topology/topology2/platform/intel/dmic-generic.conf
@@ -39,7 +39,7 @@ Object.Dai.DMIC [
                dai_index               1
                driver_version          $DMIC_DRIVER_VERSION
                io_clk                  $DMIC_IO_CLK
-               sample_rate             16000
+               sample_rate             96000
                clk_min         500000
                clk_max         4800000
                unmute_ramp_time_ms     200

```

**Reproduction Rate**
100%

**Expected behavior**
Working capture.

**Impact**
What impact does this issue have on your progress (e.g., annoyance, showstopper)

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * Kernel: {SHA}
    * SOF: {SHA}
2) Name of the topology file
    * Topology: {FILE}
3) Name of the platform(s) on which the bug is observed.
    * Platform: {PLATFORM}

$ alsatplg --version
alsatplg version 1.2.10
libasound version 1.2.8
libatopology version 1.2.8

**Screenshots or console output**

dmesg

```
[...]

PR #8671, opened 2023-12-22
[SKIP CI][Do not review][Test] Tools: Topology2: Add 2nd 96 kHz DMIC

singalsu:


singalsu:
Note: This fails because of issue #8670 with alsatplg / alsa-utils.

Issue #8639, opened 2023-12-16
[BUG] IPC4 fails to set up stream metadata correctly


andyross:
Component stream state in IPC4 doesn't seem to be set up reliably.  When the sof-mtl-max98357a-rt5682-ssp2-ssp0 topology (built form the mtl-007-drop-stable branch), the Google AEC component is being presented with a reference/playback input source that claims 4 channels, but in fact is only providing enough data for two.

Similarly, the computed "alignment constants" on a buffer's audio stream start out with (I think) garbage, and require that something somewhere call `source/sink_set_alignment_constants()` with arbitrary (and even incorrect, e.g. a byte alignment of 1 was being used!) arguments just to get the fields initialized.

The component has always had code present (see this commit for the current form: https://github.com/thesofproject/sof/pull/8571/commits/bab124dda81aac088fa789a6e1167403b288df90) to fix this at prepare() time.  But it's fragile, undocumented, got dropped during refactoring and took me **DAYS** to figure out; the symptoms were just ring buffer overruns in the downstream components that produced "mild distortion".

This just isn't right.  Components don't own their sinks/sources and shouldn't ever be responsible for configuring them.  The pipeline/IPC layer needs to do this correctly from topology, and in the event that the data is wrong or mismatched, it needs to fail gracefully.  We shouldn't ever be in a situation like this where a source promises 8 byte frames per sample but delivers 4.

lgirdwood:
@marcinszkudlinski can you comment on a solution ?

marcinszkudlinski:
Yes, I know there are inconsistences at number of channels. 
Other parameters are set properly

that's why I introduced a WA to mtl 007 branch:

https://github.com/thesofproject/sof/blob/6d3061c76f23f5e92ec89578476006f6fe3b4cd4/src/audio/google/google_rtc_audio_processing.c#L96

problems are being investigated

andyross:
Add @andrula-song as assignee given expert-seeming comments in #8571

lgirdwood:
@ujfalusi can you give this some priority. Thanks. @plbossart fyi, driver may be sending incorrect IPC4 data, we are not sure yet.

ujfalusi:
@andyross, can you attach kernel and firmware logs with pointers where things go south? I'm not sure if I follow the description.
If the sof-mtl-max98357a-rt5682-ssp2-ssp0  represents the Google AEC with 4 channels  and it should not be like that then the issue is in topology?

[...]

Issue #8340, opened 2023-10-17
[BUG] Few problems with audio_stream_avail_frames_aligned() behaviour


serhiy-katsyuba-intel:
**Describe the bug**

(1)
audio_stream_avail_frames_aligned() uses shifts instead of divisions and so it returns wrong result for formats with frame size not equal to power of 2, e.g. for streams with 3, 5 or 7 channels.

(2)
Default alignment initialization in audio_stream_init() does not work as stream frame format is not yet initialized: it is always set later somewhen after that function call.

(3)
Probably, to overcome (2), every module have to call audio_stream_init_alignment_constants() to initialize alignment. Modules usually do this in .prepare() handler. However, we support binding to a module on a running pipeline without pausing it. In this case .prepare() is not called after new binding as pipeline was not paused or reset and that might left new attached buffer stream with uninitialized alignment.
Another problem is that it is easy to forget to call audio_stream_init_alignment_constants() (must be added to every module!). I see audio_stream_init_alignment_constants() is missing at least in copier, mixin and mixout. It would be better that in this case some default 1 byte alignment initialization take place. 

(4)
if alignment initialization is missing -- because of (2) or (3) -- audio_stream_avail_frames_aligned() returns wrong (higher) number of frames. That frames value is used by module_adapter to invalidate buffer memory and is also passed as parameter to .process() func. Luckily, if all connected modules are on same core, buffer invalidate is skipped. For cross-core case, invalidate with wrong (higher) amount of frames leads to memory corruption and weird bugs.

**Environment**
1) Branch name and commit hash of the 2 repositories: sof (firmware/topology) and linux (kernel driver).
    * SOF: main branch.

audio_stream_avail_frames_aligned() source;
```
static inline uint32_t
audio_stream_avail_frames_aligned(const struct audio_stream *source,
				  const struct audio_stream *sink)
{
	uint32_t src_frames = (audio_stream_get_avail_bytes(source) >>
			source->runtime_stream_params.align_shift_idx) *
					source->runtime_stream_params.align_frame_cnt;
	uint32_t sink_frames = (audio_stream_get_free_bytes(sink) >>
			sink->runtime_stream_params.align_shift_idx) *
					sink->runtime_stream_params.align_frame_cnt;

	return MIN(src_frames, sink_frames);
}
```

andrula-song:
[...]

PR #8672, opened 2023-12-23
audio_stream: Fix uninitialized data in alignment_constants

andyross:
PR was reworked from the version described in the initial version.  See comments below and commit message.  This text is stale:

Audio stream objects are generally allocated out of zero'd heap memory, but the computed alignment fields need non-zero values to have correct behavior.  These were being left as garbage.

Set them to a byte and frame alignmnt of 1, as this corresponds to pervasive convention in the tree.  But note that a byte alignment of 1 is actually technically incorrect, as all existing DSP targets are on architectures which don't allow misaligned loads.  But existing component code is universally coded correctly anyway.

(It's worth pointing out that "set_alignment_constants()" is now exposed as an API call on abstracted source/sink objects.  But the only current implementation is on audio_stream.  Future implementations will need to correctly initialize themselves.)

This is a partial fix for Issue #8639

### DIFF
--- a/src/audio/audio_stream.c
+++ b/src/audio/audio_stream.c
@@ -93,7 +93,7 @@ void audio_stream_recalc_align(struct audio_stream *stream)
 			(is_power_of_2(process_size) ? 31 : 32) - clz(process_size);
 }
 
-void audio_stream_init_alignment_constants(const uint32_t byte_align,
+void audio_stream_set_align(const uint32_t byte_align,
 					   const uint32_t frame_align_req,
 					   struct audio_stream *stream)
 {
@@ -139,7 +139,7 @@ static int audio_stream_source_set_alignment_constants(struct sof_source *source
 {
 	struct audio_stream *audio_stream = container_of(source, struct audio_stream, source_api);
 
-	audio_stream_init_alignment_constants(byte_align, frame_align_req, audio_stream);
+	audio_stream_set_align(byte_align, frame_align_req, audio_stream);
 
 	return 0;
 }
@@ -150,7 +150,7 @@ static int audio_stream_sink_set_alignment_constants(struct sof_sink *sink,
 {
 	struct audio_stream *audio_stream = container_of(sink, struct audio_stream, sink_api);
 
-	audio_stream_init_alignment_constants(byte_align, frame_align_req, audio_stream);
+	audio_stream_set_align(byte_align, frame_align_req, audio_stream);
 
 	return 0;
 }
@@ -195,7 +195,7 @@ void audio_stream_init(struct audio_stream *audio_stream, void *buff_addr, uint3
 	audio_stream->addr = buff_addr;
 	audio_stream->end_addr = (char *)audio_stream->addr + size;
 
-	audio_stream_init_alignment_constants(1, 1, audio_stream);
+	audio_stream_set_align(1, 1, audio_stream);
 	source_init(audio_stream_get_source(audio_stream), &audio_stream_source_ops,
 		    &audio_stream->runtime_stream_params);
 	sink_init(audio_stream_get_sink(audio_stream), &audio_stream_sink_ops,
--- a/src/audio/audio_stream.c
+++ b/src/audio/audio_stream.c
@@ -79,10 +79,10 @@ static uint32_t audio_stream_frame_align_get(const uint32_t byte_align,
 }
 
 
-void audio_stream_init_alignment_constants(const uint32_t byte_align,
-					   const uint32_t frame_align_req,
-					   struct audio_stream *stream)
+void audio_stream_recalc_align(struct audio_stream *stream)
 {
+	const uint32_t byte_align = stream->byte_align_req;
+	const uint32_t frame_align_req = stream->frame_align_req;
 	uint32_t process_size;
 	uint32_t frame_size = audio_stream_frame_bytes(stream);
 
@@ -92,6 +92,16 @@ void audio_stream_init_alignment_constants(const uint32_t byte_align,
 	stream->runtime_stream_params.align_shift_idx	=
 			(is_power_of_2(process_size) ? 31 : 32) - clz(process_size);
 }
+
+void audio_stream_init_alignment_constants(const uint32_t byte_align,
+					   const uint32_t frame_align_req,
+					   struct audio_stream *stream)
+{
+	stream->byte_align_req = byte_align;
+	stream->frame_align_req = frame_align_req;
+	audio_stream_recalc_align(stream);
+}
+
 
 static int audio_stream_release_data(struct sof_source *source, size_t free_size)
 {
@@ -145,11 +155,28 @@ static int audio_stream_sink_set_alignment_constants(struct sof_sink *sink,
 	return 0;
 }
 
+static int source_format_set(struct sof_source *source)
+{
+	struct audio_stream *s = container_of(source, struct audio_stream, source_api);
+
+	audio_stream_recalc_align(s);
+	return 0;
+}
+
+static int sink_format_set(struct sof_sink *sink)
+{
+	struct audio_stream *s = container_of(sink, struct audio_stream, sink_api);
+
+	audio_stream_recalc_align(s);
+	return 0;
+}
+
 static const struct source_ops audio_stream_source_ops = {
 	.get_data_available = audio_stream_get_data_available,
 	.get_data = audio_stream_get_data,
 	.release_data = audio_stream_release_data,
 	.audio_set_ipc_params = audio_stream_set_ipc_params_source,
+	.on_audio_format_set = source_format_set,
 	.set_alignment_constants = audio_stream_source_set_alignment_constants
 };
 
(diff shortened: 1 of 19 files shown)


src/audio/audio_stream.c
```diff
@@ -141,6 +141,7 @@ void audio_stream_init(struct audio_stream *audio_stream, void *buff_addr, uint3
 	audio_stream->addr = buff_addr;
 	audio_stream->end_addr = (char *)audio_stream->addr + size;
 
+	audio_stream_init_alignment_constants(1, 1, audio_stream);
```
btian1, line 144: the essence is to set: align_frame_cnt and align_shift_idx, do you think current function name exactly reflect the content?
I mean good to change function name, due to 1, 1 params make people don't know exactly what it is setting.

src/audio/dcblock/dcblock.c
```diff
@@ -183,17 +183,6 @@ static int dcblock_process(struct processing_module *mod,
...
 
-/* init and calculate the aligned setting for available frames and free frames retrieve*/
-static inline void dcblock_set_frame_alignment(struct audio_stream *source,
-					       struct audio_stream *sink)
-{
-	const uint32_t byte_align = 1;
-	const uint32_t frame_align_req = 1;
-
-	audio_stream_init_alignment_constants(byte_align, frame_align_req, source);
-	audio_stream_init_alignment_constants(byte_align, frame_align_req, sink);
```
btian1, line 194: if this part can be removed, how about others? like eq_fir or eq_iir?

src/audio/audio_stream.c
andrula-song: actually, at this point, I'am afraid the format/channel of audio_stream haven't been set yet, wouldn't the set in audio_stream_set_params enough?

src/audio/dcblock/dcblock.c
andrula-song: I think the purpose of this PR is remove all the default alignment set(all byte_align = 1 && frame_align_req = 1 ) and set the default value while buffer is allocated.

andyross: The eq_fir/iir usage passes a "2" for frame alignment, with a comment indicating that it wants to process pairs of frames (see e.g. https://github.com/thesofproject/sof/blob/main/src/audio/eq_fir/eq_fir.c#L399).  That seems like a legitimate usage to me (not knowing anything else about the algorithm).  The goal with this patch is just to remove the requirement for **all** components to call this even when they don't have special alignment requirements.

src/audio/audio_stream.c
kv2019i: @andyross @andrula-song I wonder if we should start by just applying the first patch. Setting the alignment constraints before channel count and sample format is not correct, but at least we'd get a determistic behaviour in case some module does not correctly set the constraints in its prepare functions. Or alternatively initialize the values to some known invalid value to ensure we catch these cases. 

We discussed earlier that we could have update hooks to recalculate the alignment constants whenever channel count or sample format is updated, but this is not so straightforward currently as the raw alignment input values (the constraints coming from algorithm) are not directly stored in audio stream.

Currently, the modules do need to call this in their prepare.

kv2019i (commented):
Thanks @andyross . I see a few ways how to proceed with this, please see inline.

andyross:
> Setting the alignment constraints before channel count and sample format is not correct, but at least we'd get a determistic behaviour in case some module does not correctly set the constraints in its prepare functions.

See... I think that's just wrong.  It's a design flaw to make components responsible for doing this, because by definition it requires coordination with the component on the other side of the buffer, which module code doesn't own and shouldn't have to know about.  That's why I'm describing this as an "uninitialized data bug" -- the default stream after init doesn't work correctly.

A better scheme would have the component provide "alignment requirements"[1] just after initialization (via a field in the module, API call, whatever), and have the pipeline setup code walk the graph, validate[2] the settings we got from the kernel, and set things up correctly such that the modules can just read/interpret them in prepare().

[1] And all other stream metadata, c.f. the linked bug.

[2] Note that the topology layer is ambiguous here: stream formats in IPC4 are stored on the widget, but the buffer connects two widgets.  So we're storing the same data twice, and it's possible to have the two settings disagree.  AFAICT there's no code layer prepared to detect this.   In fact the framework layer seems not to be inspecting some of the data (the "base_cfg_ext" struct) at all.

kv2019i:
This is certainly different in IPC4 versus IPC3. With IPC3, the kernel set sparams at the graph edges and FW iterates through the graph and propagates the parameters. With IPC4, it's the host that does the graph walk and sets the formats so that they match and sets the pin configuration explicitly for each pin (one format is set for each pin). So modules should not have to bother with this, they just have to follow what host tells them to do (given it gets the orders correctly, e.g. #8575 for multi pin case).  It's possible to mix and match a bit and have e.g. FW do additional logic even when pin configurations come from the host, but so far we have not done a lot of this.

cujomalainey (approved):

src/audio/audio_stream.c
andyross: >  I wonder if we should start by just applying the first patch.

The second patch is a behavioral noop, it only removes the code that was recapitulating the defaults.  Though it would be sad to lose the cleanup.  I feel really strongly at this point that we want **all** of this "just override the topology/framework values to make it work" code out of the components entirely.  Buffers should be write-only from component code.

kv2019i: @andyross Hmm, but the audio_stream_init_alignment_constants() is really a convenience/helper interface between the audio-stream (and now sink/source) and the module. For this helper interface to function correctly, algorithm must use it once it knows the channel count and sample format. The "byte_align==1" is indeed odd, but it's ignored if sample size is larger (and it always is). So in practise these calls set the alignment based on the stream's channel count and sample size. 

Maybe this is correct, but it's not obvious to me the result will be the same alignment in all cases with combination of these two patches. I'd like +1 from @singalsu @andrula-song et al this is valid. The first patch alone seems safe, as it sets the defaults to sane values.

andyross: Oh... OK, now I'm understanding you.  To restate to make sure I have it:  init_alignment_constants() takes as input the frame size from the stream struct, but at creation time that doesn't exist.  So in general it produces different values in when called in this patch vs. at runtime in the components.  But that maybe that's OK, since the components are only doing this to express "don't care" alignment requirements anyway and any value that doesn't break should work.

So I actually agree, we should definitely remove the second patch if so, and replace it with e.g. code in ipc_comp_connect() (both versions) to do this on behalf of default components.

andrula-song: @kv2019i @andyross , if you indeed want to apply an incorrect value first and set the specific value for each component, I'am okay, but I recall some one who report an issue by setting the meaningless value while the format and channel not known here.(sorry, can not find that issue id).

marcinszkudlinski (approved):
looks fine, we should remove align param completely (if possible), alignment should always be a frame size.

kv2019i (changes requested):
This is getting so many approvals, so let me mark -1 for the moment. Please check https://github.com/thesofproject/sof/issues/8340 , I don't think this is safe (at least with the second patch in).

kv2019i:
@marcinszkudlinski  wrote.
> looks fine, we should remove align param completely (if possible), alignment should always be a frame size.

Please check existing uses of audio_stream_init_alignment_constants() like in EQ FIR.

We can get rid of this, but then algorithms that have vector optimized loops, need to duplicate the logic to calculate avail alignments themselves.

Algorithms that process sample by sample will not need this, but that's not the main audience for this interface.

lgirdwood (commented):
I like the second part of this change - lots of removals. Lets align on the fix in teh 1st part where we construct the data (in one place).

andyross (commented):
Just to clarify my comment from above: @kv2019i is right and this can't merge in this form as there needs to be a layer that guarantees that alignment values are recomputed at the time they're needed (which includes other changes to frame size!).  I have a quick mockup working, will get this reworked and submitted over the weekend.

I'd -1 too, but Github won't allow me to do that to my own patch :)

andyross:
OK, reworked version is up.  The cleanup patch is identical, but the alignment constant handling is now modified to be a lazy-evaluated version that can be called safely at any time.  Please re-review.

andyross:
And.... of course the rename patch crossed with new work in main that isn't in mtl-007-drop-stable and so isn't directly testable vs. this work.  I'm getting pretty fed up with having to work across branches like this.  Will fix.

cujomalainey:
> I'd -1 too, but Github won't allow me to do that to my own patch :)

@andyross you can convert to draft (right below reviewers list). Would also prevent submission.

cujomalainey (approved):
love having negative diffs :+1:

kv2019i:
@andrula-song @singalsu please recheck

kv2019i (approved):
Excellent stuff @andyross , so this solves both problems, redundant code to set no-extra-constraints and problem of recalculating the alignment if format is changed!

kv2019i:
Of the failing CI cases, https://sof-ci.01.org/sofpr/PR8672/build1656/devicetest/index.html is a known fail with system PM and can be ignored. https://sof-ci.01.org/sofpr/PR8672/build1655/devicetest/index.html has a fail on MTL I can't explain (no direct connection to the PR, but also it is suspect to have this fail). Let's see how the other CI checks fair.

lgirdwood (approved):

PR #8675, opened 2023-12-25
Work round Fix for noise issue when playback and capturer in parallel.

TangleZ:
Did basic tests(playback&record in parallel, with suspend) in i.MX8QM/i.MX8ULP platform. The fix changed edma driver behavior, so need check not only zephyr but also xtensa build.
@dbaluta Please tell me if you think the fix still impact some tests, I will do the tests you mentioned. Thanks a lot!

### DIFF
--- a/xtos/include/sof/lib/dma.h
+++ b/xtos/include/sof/lib/dma.h
@@ -198,6 +198,8 @@ struct dma_ops {
 	int (*get_data_size)(struct dma_chan_data *channel, uint32_t *avail,
 			     uint32_t *free);
 
+	int (*clear_data_copied)(struct dma_chan_data *channel);
+
 	int (*get_attribute)(struct dma *dma, uint32_t type, uint32_t *value);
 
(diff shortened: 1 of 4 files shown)


dbaluta:
thank you Peng! let’s merge this and go for RC1 with it!

xtos/include/sof/lib/dma.h
```diff
@@ -198,6 +198,8 @@ struct dma_ops {
 	int (*get_data_size)(struct dma_chan_data *channel, uint32_t *avail,
 			     uint32_t *free);
 
+	int (*clear_data_copied)(struct dma_chan_data *channel);
```
TangleZ, line 201: @dbaluta  one question here, there is a file posix/include/sof/lib/dma.h, should we need to implement it?

dbaluta: for us, no. We dont build for posix

PR #8678, opened 2023-12-27
Zephyr: Only print memory information with performance build

btian1:
Due to CONFIG_SYS_HEAP_RUNTIME_STATS set with yes, this part of heap memory information print are always enabled, however, this log information only be used for memory performance analysis, so add CONFIG_PERFORMANCE_COUNTERS restriction to this log print.

### DIFF
--- a/zephyr/lib/alloc.c
+++ b/zephyr/lib/alloc.c
@@ -158,15 +158,14 @@ static void *heap_alloc_aligned(struct k_heap *h, size_t min_align, size_t bytes
 {
 	k_spinlock_key_t key;
 	void *ret;
-#if CONFIG_SYS_HEAP_RUNTIME_STATS && CONFIG_IPC_MAJOR_4
-	struct sys_memory_stats stats;
-#endif
 
 	key = k_spin_lock(&h->lock);
 	ret = sys_heap_aligned_alloc(&h->heap, min_align, bytes);
 	k_spin_unlock(&h->lock, key);
 
-#if CONFIG_SYS_HEAP_RUNTIME_STATS && CONFIG_IPC_MAJOR_4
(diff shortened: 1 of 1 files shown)


kv2019i (commented):
Hmm, CONFIG_PERFORMANCE_COUNTERS is originally only for DSP cycle counting. Not sure we want to mix memory use debug with perf debug.

In your original comimt 371d35108a75c , CONFIG_SYS_HEAP_RUNTIME_STATS was enabled for perf build. Can we just disable this in the non-perf builds (and save some memory and cycles as well).

btian1:
> CONFIG_SYS_HEAP_RUNTIME_STATS

@kv2019i , CONFIG_SYS_HEAP_RUNTIME_STATS is a zephyr feature, althrough disable it will also works for print, however,
check below picture, there is a lot of place use this as well in zephyr, it disable it may beak some of them, when I added this in: https://github.com/thesofproject/sof/commit/371d35108a75cd8788cfadcb00ae9a3028677f22, it only used for performance analysis, memory usage is also part of performance(memory performance).

image

kv2019i:
@btian1 CONFIG_SYS_HEAP_RUNTIME_STATS is only needed in the build if application (in this case SOF) wants to use sys_heap_runtime_stats_get(). So if we disable the only use of this function, we can disable the whole feature and save both memory and runtime cycles. 

What is not clear is that do we have some other users (added after your initial commit) for RUNTIME_STATS? I.e. it's used somewhere else than the performance build. That changes things of course.

btian1:
We can close this until we see memory logs is happening in non-performance build, currently, I double checked with tgl and mtl, it is not set in non-performance build. @kv2019i

PR #8679, opened 2023-12-28
META programming cleanup

btian1:
Meta programming clean up, later will try to align IPC3 logging with zephyr logging interface


lgirdwood:
@btian1 IIRC, I think only 4 macros are used today in the META headers. We could remove the unused macros as part of the 1st step.

btian1:
> @btian1 IIRC, I think only 4 macros are used today in the META headers. We could remove the unused macros as part of the 1st step.

I understand your point, below 4 macros are used in sof code:

META_QUOTE
META_CONCAT
META_MAP
META_COUNT_VARAGS_BEFORE_COMPILE

However, all META definitions are defined together, you even can't remove several, some are not used by SOF, however, it is used by CMOCKA, so this patch remove is quite small.

Next step I will focus on use zephyr logging interface replace existing IPC3 logging interface, this purpose is make SOF code looks more like an zephyr application!

src/include/sof/lib/dai-legacy.h
```diff
@@ -254,7 +254,7 @@ struct dai_type_info {
 /* class (driver) level (no device object) tracing */
 
 #define dai_cl_err(drv_p, __e, ...)		\
-	trace_dev_err(trace_dai_dvr_get_tr_ctx,	\
+	trace_dev_err(trace_dai_drv_get_tr_ctx,	\
```
kv2019i, line 257: Apparently unused code as this could have not compiled. Based on git history, the dai_cl_ family of macros has never been used in the codebase.

kv2019i (approved):

src/include/sof/lib/dai-legacy.h
btian1: same thinking, this is the reason I submit this patch, I will try to check and remove.
Then try to align with zephyr logging interface.

lgirdwood (approved):
Good stuff,  Btw, I think a lot more can be removed when using Zephyr target. This may mean we can move and split the file into the rtos include directories.

btian1:
> Good stuff, Btw, I think a lot more can be removed when using Zephyr target. This may mean we can move and split the file into the rtos include directories.

META are only be used by IPC3 logging, if zephyr, whole META can be removed, like I did for cmocka.
But currently, SOF still need support IPC3 logging, this is the reason to keep left META stuffs.

lgirdwood:
> > Good stuff, Btw, I think a lot more can be removed when using Zephyr target. This may mean we can move and split the file into the rtos include directories.
> 
> META are only be used by IPC3 logging, if zephyr, whole META can be removed, like I did for cmocka. But currently, SOF still need support IPC3 logging, this is the reason to keep left META stuffs.

ok, then we can move the remaining META to xtos include header.

PR #8681, opened 2024-01-01
Simple dependency-free ALSA test rig for PCM capture analysis.

andyross:
Just drop this script on a test device to run it.  No tools to build, no dependencies to install.  Confirmed to run on Python 3.8+ with nothing more than the core libraries and a working libasound.so.2 visible to the runtime linker.

When run without arguments, the tool will record from the capture device for the specified duration, then emit the resulting samples back out the playback device without processing (except potentially to convert the sample format from s32_le to s16_le if needed, and to discard any channels beyond those supported by the playback device).

Passing --chirp-test enables a playback-to-capture latency detector: the tool will emit a short ~6 kHz wave packet via ALSA's mmap interface (which allows measuring and correcting for the buffer latency from the userspace process) and simultaneously loop on short reads from the capture device looking for the moment it arrives.

Passing --echo-test enables a capture-while-playback test.  The script will play a specified .wav file ("noise.wav" by default) for the specified duration, while simultaneously capturing, and report the "power" (in essentially arbitrary units, but it's linear with actual signal energy assuming the sample space is itself linear) of the captured data to stdout at the end of the test.

### DIFF
--- a/scripts/capture-test.py
+++ b/scripts/capture-test.py
@@ -1,0 +1,364 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright 2024 Google LLC
+# Author: Andy Ross <andyross@google.com>
+import os
+import re
+import sys
+import time
+import struct
+import random
+import argparse
+import ctypes as c
+
+# Simple dependency-free ALSA test rig for PCM capture analysis.
+#
+# Just drop this script on a test device to run it.  No tools to
+# build, no dependencies to install.  Confirmed to run on Python 3.8+
+# with nothing more than the core libraries and a working
+# libasound.so.2 visible to the runtime linker.
+#
+# When run without arguments, the tool will record from the capture
+# device for the specified duration, then emit the resulting samples
+# back out the playback device without processing (except potentially
+# to convert the sample format from s32_le to s16_le if needed, and to
+# discard any channels beyond those supported by the playback device).
+#
+# Passing --chirp-test enables a playback-to-capture latency detector:
+# the tool will emit a short ~6 kHz wave packet via ALSA's mmap
+# interface (which allows measuring and correcting for the buffer
+# latency from the userspace process) and simultaneously loop on short
+# reads from the capture device looking for the moment it arrives.
+#
+# Passing --echo-test enables a capture-while-playback test.  The
+# script will play a specified .wav file ("noise.wav" by default) for
+# the specified duration, while simultaneously capturing, and report
+# the "power" (in essentially arbitrary units, but it's linear with
+# actual signal energy assuming the sample space is itself linear) of
+# the captured data to stdout at the end of the test.
+
+opts = argparse.ArgumentParser()
+opts.add_argument("--disable-rtnr", action="store_true", help="Disable RTNR noise reduction")
+opts.add_argument("-c", "--card", type=int, default=0, help="ALSA card index")
+opts.add_argument("--pcm", type=int, default=16, help="Output ALSA PCM index")
+opts.add_argument("--cap", type=int, default=18, help="Capture ALSA PCM index")
+opts.add_argument("--rate", type=int, default=48000, help="Sample rate")
+opts.add_argument("--chan", type=int, default=2, help="Output channel count")
+opts.add_argument("--capchan", type=int, help="Capture channel count (if different from output)")
+opts.add_argument("--capbits", type=int, default=16, help="Capture sample bits (16 or 32)")
+opts.add_argument("--noise", default="noise.wav", help="WAV file containing 'noise' for capture")
+opts.add_argument("--duration", type=int, default=3, help="Capture duration (seconds)")
+opts.add_argument("--chirpcyc", type=int, default=120, help="Repetitions of chirp waveform")
+opts.add_argument("--chirp-test", action="store_true", help="Test latency with synthesized audio")
+opts.add_argument("--echo-test", action="store_true", help="Test simultaneous capture/playback")
+
+opts = opts.parse_args()
+if not opts.capchan:
+    opts.capchan = opts.chan
+opts.base_test = not (opts.chirp_test or opts.echo_test)
+
+# Tiny ctypes stub.  Wraps the alsa API such that errno returns (at
+# least ones that look like an errno) become OSErrors and don't need
+# to be checked.  Includes a generalized alloc() that wraps all the
+# _sizeof() predicates and allocates from the (safe/collected) python
+# heap.  Provides a simple spot for putting (manually-derived)
+# constants.  The ALSA C API is mostly-structless and quite simple, so
+# this tends to work well without a lot of ctypes use except for an
+# occasional constructed integer or byref() pointer.
+class ALSA:
+    PCM_STREAM_PLAYBACK = 0
+    PCM_STREAM_CAPTURE = 1
+    PCM_FORMAT_S16_LE = 2
+    PCM_FORMAT_S32_LE = 10
+    PCM_ACCESS_MMAP_INTERLEAVED = 0
+    PCM_ACCESS_RW_INTERLEAVED = 3
+    def __init__(self):
+        self.lib = c.cdll.LoadLibrary("libasound.so.2")
+    def __getattr__(self, name):
+        if name.startswith("snd_"):
+            fn = getattr(self.lib, name)
+            if name.endswith("_name"): # These return strings!
+                fn.restype = c.c_char_p
+                return lambda *args: fn(*args).decode("utf-8")
+            else:
+                return lambda *args: self.err_wrap(fn(*args))
+    def err_wrap(self, ret):
+        if ret < 0 and ret > -200:
+            raise OSError(os.strerror(-ret))
+        return ret
+    def alloc(self, typ):
+        return (c.c_byte * getattr(self.lib, f"snd_{typ}_sizeof")())()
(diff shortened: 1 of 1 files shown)


andyross:
SOF has sort of a gap with ALSA-API-level tests of playback and capture behavior.  I see multiple single-pipeline tests using aplay/arecord, but nothing that works well for getting numbers out of AEC performance for things like latency and correction gain when using two pipelines in tandem.  This is an attempt to plug the hole.

I'm particularly proud of the all-python (via ctypes) implementation, which is easier for all users, but in particular allows this to be dropped directly on an ARM chromebook without having to decide on a cross compilation strategy for SOF's host tools.

While this is aimed at AEC integration work, there's no dependency there and you can use this for any SOF capture pipe on any system.  There is a Realtek-specific "--disable-rtnr" argument (which does what it says, as noise reduction seems to interfere with the chirp test), but that's just a wrapper around an ALSA control that's a noop if not present.

Finally, I dropped this in the top level ./scripts directory for visibility.  There doesn't seem to be an obvious home in ./tools or ./test.  Let me know if there's a better place.

And please review, obviously.  I baldly claim this is a "simple" tool, but in fact some of the analysis gets a little subtle and I wouldn't be surprised if I had a few units bugs or missing corrections in there.

marc-hb:
> Finally, I dropped this in the top level ./scripts directory for visibility. There doesn't seem to be an obvious home in ./tools or ./test. Let me know if there's a better place.

I think this belongs to either https://github.com/thesofproject/sof-test or (even better) https://github.com/alsa-project/alsa-utils if it's not SOF-specific. Because unlike unit tests, you don't want that script to change when you're git bisecting https://github.com/thesofproject/sof.

`git bisect` is not the only consideration when choosing which git repo to use but it's the simplest and best litmus test.

scripts/capture-test.py
```diff
@@ -0,0 +1,364 @@
...
+opts.add_argument("--chan", type=int, default=2, help="Output channel count")
+opts.add_argument("--capchan", type=int, help="Capture channel count (if different from output)")
+opts.add_argument("--capbits", type=int, default=16, help="Capture sample bits (16 or 32)")
+opts.add_argument("--noise", default="noise.wav", help="WAV file containing 'noise' for capture")
+opts.add_argument("--duration", type=int, default=3, help="Capture duration (seconds)")
+opts.add_argument("--chirpcyc", type=int, default=120, help="Repetitions of chirp waveform")
+opts.add_argument("--chirp-test", action="store_true", help="Test latency with synthesized audio")
+opts.add_argument("--echo-test", action="store_true", help="Test simultaneous capture/playback")
+
+opts = opts.parse_args()
```
marc-hb, line 55: I see no reason for this code to be at the top-level while everything else is in either a class or a function.
```python
def parse_args():
    global opts
    ...
```

scripts/capture-test.py
```diff
@@ -0,0 +1,364 @@
...
+# interface (which allows measuring and correcting for the buffer
+# latency from the userspace process) and simultaneously loop on short
+# reads from the capture device looking for the moment it arrives.
+#
+# Passing --echo-test enables a capture-while-playback test.  The
+# script will play a specified .wav file ("noise.wav" by default) for
+# the specified duration, while simultaneously capturing, and report
+# the "power" (in essentially arbitrary units, but it's linear with
+# actual signal energy assuming the sample space is itself linear) of
+# the captured data to stdout at the end of the test.
```
marc-hb, line 38: Assign this to a new HELP_STRING constant and then:

```python
argparse.ArgumentParser(description=HELP_STRING,...
# or
argparse.ArgumentParser(epilog=HELP_STRING,...




```

There's is stuff like https://stackoverflow.com/questions/35917547/python-argparse-rawtexthelpformatter-with-line-wrap if you want to get fancy but don't worry about it for now. Someone else can fix formatting later :-)

scripts/capture-test.py
```diff
@@ -0,0 +1,364 @@
...
+
+# Tiny ctypes stub.  Wraps the alsa API such that errno returns (at
+# least ones that look like an errno) become OSErrors and don't need
+# to be checked.  Includes a generalized alloc() that wraps all the
+# _sizeof() predicates and allocates from the (safe/collected) python
+# heap.  Provides a simple spot for putting (manually-derived)
+# constants.  The ALSA C API is mostly-structless and quite simple, so
+# this tends to work well without a lot of ctypes use except for an
+# occasional constructed integer or byref() pointer.
+class ALSA:
```
marc-hb, line 68: This looks cool and born to be re-used elsewhere later. So, definitely part of alsa-utils or sof-test.

marc-hb (commented):
> I see multiple single-pipeline tests using aplay/arecord,

sof-test also uses ALSABAT from https://github.com/alsa-project/alsa-utils/tree/master/bat

> SOF has sort of a gap with ALSA-API-level tests of playback and capture behavior. 

BTW:

- https://github.com/thesofproject/sof-test/pull/437
- https://github.com/thesofproject/sof-test/pull/972
- https://github.com/thesofproject/sof-test/issues/1069

https://github.com/thesofproject/sof-test/issues?q=label%3A%22type%3Atest+coverage+gap%22+

scripts/capture-test.py
andyross: Then we need a deployment strategy and can't just drop the script on the target.  I'm all for reuse, but not for complicating interfaces to support code that isn't reused yet.  And it's really extremely small and unlikely to evolve much.

See for example the "Regs" class in cavstool (and now acetool).  Sure, that's technically a python MMIO register block abstraction and not part of a logging tool, but it's likewise tiny and has lived very happily in-script.  And no one seems bothered much by the cut-and-paste, given that the 10x larger enclosing script is itself mostly just a copy:  https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/xtensa/intel_adsp/tools/cavstool.py#L328

As for alsa-utils, I can try, but there's not a line of python in that quarter-century-old source tree, so I don't know if I'm up for the fights involved.

marc-hb: > See for example the "Regs" class in cavstool (and now acetool). Sure, that's technically a python MMIO register block abstraction and not part of a logging tool, but it's likewise tiny and has lived very happily in-script

That looks like a much smaller re-use potential to me. Either way that's a future discussion.

> As for alsa-utils, I can try, but there's not a line of python in that quarter-century-old source tree, so I don't know if I'm up for the fights involved.

Try and immediately fallback on sof-test if that fails. We already have non-SOF specific code in sof-test.

cc: @perexg

marc-hb: > ... just drop the script on the target.

On that topic: alsa-utils is available out of the box (with some usual propagation delays) in every Linux distribution. sof-test is not.

scripts/capture-test.py
```diff
@@ -0,0 +1,364 @@
...
+    PCM_STREAM_PLAYBACK = 0
+    PCM_STREAM_CAPTURE = 1
+    PCM_FORMAT_S16_LE = 2
+    PCM_FORMAT_S32_LE = 10
+    PCM_ACCESS_MMAP_INTERLEAVED = 0
```
cujomalainey, line 73: these should be able to be pulled in from the library API no?

scripts/capture-test.py
```diff
@@ -0,0 +1,364 @@
...
+# convert to playback format, and play it.
+def base_test():
+    bufs = []
+    energy = 0
+    for rec in do_capture(opts.duration):
+        crec = cap_to_playback(rec[1])
+        bufs.append(crec[0])
+        energy += crec[1]
+    play_buf(b''.join(bufs))
+    print(f"Energy {energy}")
```
cujomalainey, line 350: could we maybe get better separation of the pcm code and the analysis code?

scripts/capture-test.py
andyross: They're just preprocessor symbols in alsa headers, they don't appear in the shared library sadly.

scripts/capture-test.py
andyross: That's sort of the idea.  I mean, it's a 300-line script, so everything is near everything else.  But the way it's structured is that there's are "ALSA" routines like init_stream/do_capture/play_buf that work with the low level stuff and the analysis happens up-stack in the code that calls them.

andyross:
Submitted to sof-test here: https://github.com/thesofproject/sof-test/pull/1144

Looking through, I really don't think alsa-utils is the right spot.  Longer term, with some evolution, maybe.  But for now this is really best viewed as a smoke test for SOF capture (and AEC in particular, I'm using that chirp with a whitebox internal to the component that I need to get submitted as soon as I can get the rework PR moving).

Unfortunately I can't add reviewers there, so I'm going to leave this one open for now just so I have a place to whine to an audience.

andyross:
(And I'm pretty sure I addressed the existing comments.  The best I could do for @cujomalainey was to just reorder everything to put all ALSA-related code first, and put "pcm_" or "ctl_" at the front of the routines that touch ALSA APIs.

andyross:
Close this one.  New PR in sof-test is getting lots of review and I think I've addressed all the notes here.

scripts/capture-test.py
marc-hb: > As for alsa-utils, I can try, but there's not a line of python in that quarter-century-old source tree, ...

Proving you sort-of-wrong by sheer chance (was looking for something else):

https://github.com/alsa-project/alsa-tests/tree/master/python

PR #8682, opened 2024-01-02
Volume: Add volume SIMD build option

btian1:
Based on #8787, this patch is adding volume build option for different HIFI levels.

### DIFF
--- a/src/audio/volume/Kconfig
+++ b/src/audio/volume/Kconfig
@@ -8,6 +8,8 @@
 
 if COMP_VOLUME
 
+rsource "Kconfig.simd"
+
 config COMP_VOLUME_WINDOWS_FADE
        bool "Windows Fade shape volume transitions support"
        help
--- a/src/audio/volume/Kconfig.simd
+++ b/src/audio/volume/Kconfig.simd
@@ -1,0 +1,34 @@
+# SPDX-License-Identifier: BSD-3-Clause
+
+comment "volume optimization level select"
+
+choice "VOLUME_SIMD_LEVEL_SELECT"
+	prompt "choose which SIMD level used for volume module"
+	depends on COMP_VOLUME
+	default VOLUME_HIFI_MAX
+
+	config VOLUME_HIFI_MAX
+		prompt "SIMD will selected by toolchain pre-defined header"
+		bool
+		help
+			When this was selected, optimization level will be determined
+			by toolchain pre-defined macros in core isa header file.
+
+	config VOLUME_HIFI_4
+		prompt "choose HIFI4 intrinsic optimized volume module"
+		bool
+		help
+			This option used to build HIFI4 optimized volume code
+
+	config VOLUME_HIFI_3
+		prompt "choose HIFI3 intrinsic optimized volume module"
+		bool
+		help
+			This option used to build HIFI3 intrinsic optimized volume code
+
+	config VOLUME_HIFI_NONE
+		prompt "choose generic C volume module, no HIFI SIMD involved"
+		bool
+		help
+			This option used to build volume generic code.
+endchoice
--- a/src/audio/volume/volume.c
+++ b/src/audio/volume/volume.c
@@ -587,7 +587,7 @@ static int volume_process(struct processing_module *mod,
 		cd->peak_cnt = 0;
 		peak_vol_update(cd);
 		memset(cd->peak_regs.peak_meter, 0, sizeof(cd->peak_regs.peak_meter));
-#ifdef VOLUME_HIFI4
+#if SOF_USE_HIFI(4, VOLUME) || SOF_USE_HIFI(5, VOLUME)
 		memset(cd->peak_vol, 0, sizeof(int32_t) * SOF_IPC_MAX_CHANNELS * 4);
 #endif
 	}
@@ -622,7 +622,7 @@ static vol_zc_func vol_get_zc_function(struct comp_dev *dev,
 static void volume_set_alignment(struct audio_stream *source,
 				 struct audio_stream *sink)
 {
-#if XCHAL_HAVE_HIFI3 || XCHAL_HAVE_HIFI4
+#if SOF_USE_HIFI(3, VOLUME) || SOF_USE_HIFI(4, VOLUME) || SOF_USE_HIFI(5, VOLUME)
 	/* Both source and sink buffer in HiFi 3 or HiFi4 processing version,
 	 * xtensa intrinsics ask for 8-byte aligned. 5.1 format SSE audio
 	 * requires 16-byte aligned.
--- a/src/audio/volume/volume.h
+++ b/src/audio/volume/volume.h
@@ -34,19 +34,6 @@
 struct comp_buffer;
 struct sof_ipc_ctrl_value_chan;
 
-#if defined(__XCC__)
-# include <xtensa/config/core-isa.h>
-# if XCHAL_HAVE_HIFI4
-#  define VOLUME_HIFI4
-# elif XCHAL_HAVE_HIFI3
-#  define VOLUME_HIFI3
-# else
-#  define VOLUME_GENERIC
-# endif
-#else
-# define VOLUME_GENERIC
-#endif
-
 /**
  * \brief In IPC3 volume is in Q8.16 format, in IPC4 in Q1.31, but is converted
  * by firmware to Q1.23 format.
--- a/src/audio/volume/volume_hifi4.c
+++ b/src/audio/volume/volume_hifi4.c
@@ -21,7 +21,7 @@ LOG_MODULE_DECLARE(volume_hifi4, CONFIG_SOF_LOG_LEVEL);
 
 #include "volume.h"
 
-#ifdef VOLUME_HIFI4
+#if SOF_USE_HIFI(4, VOLUME) || SOF_USE_HIFI(5, VOLUME)
 
 #if (!CONFIG_COMP_PEAK_VOL)
 
--- a/src/audio/volume/volume_generic.c
+++ b/src/audio/volume/volume_generic.c
@@ -27,7 +27,7 @@ LOG_MODULE_DECLARE(volume_generic, CONFIG_SOF_LOG_LEVEL);
 
 #include "volume.h"
 
-#ifdef VOLUME_GENERIC
+#if SOF_USE_HIFI(NONE, VOLUME)
 
 #if (!CONFIG_COMP_PEAK_VOL)
 
--- a/src/audio/volume/volume_generic_with_peakvol.c
+++ b/src/audio/volume/volume_generic_with_peakvol.c
@@ -23,7 +23,7 @@ LOG_MODULE_DECLARE(volume_generic, CONFIG_SOF_LOG_LEVEL);
 
 #include "volume.h"
 
-#ifdef VOLUME_GENERIC
+#if SOF_USE_HIFI(NONE, VOLUME)
 
 #if CONFIG_COMP_PEAK_VOL
 
--- a/src/audio/volume/volume_hifi3.c
+++ b/src/audio/volume/volume_hifi3.c
@@ -21,7 +21,7 @@ LOG_MODULE_DECLARE(volume_hifi3, CONFIG_SOF_LOG_LEVEL);
 
 #include "volume.h"
 
-#ifdef VOLUME_HIFI3
+#if SOF_USE_HIFI(3, VOLUME)
 
 #if (!CONFIG_COMP_PEAK_VOL)
 
--- a/src/audio/volume/volume_hifi3_with_peakvol.c
+++ b/src/audio/volume/volume_hifi3_with_peakvol.c
@@ -21,7 +21,7 @@ LOG_MODULE_DECLARE(volume_hifi3, CONFIG_SOF_LOG_LEVEL);
 
 #include "volume.h"
 
-#ifdef VOLUME_HIFI3
+#if SOF_USE_HIFI(3, VOLUME)
 
 #if CONFIG_COMP_PEAK_VOL
 
--- a/src/audio/volume/volume_hifi4_with_peakvol.c
+++ b/src/audio/volume/volume_hifi4_with_peakvol.c
@@ -21,7 +21,7 @@ LOG_MODULE_DECLARE(volume_hifi4, CONFIG_SOF_LOG_LEVEL);
 
 #include "volume.h"
 
-#ifdef VOLUME_HIFI4
+#if SOF_USE_HIFI(4, VOLUME) || SOF_USE_HIFI(5, VOLUME)
 
 #if CONFIG_COMP_PEAK_VOL
 #include <xtensa/tie/xt_hifi4.h>


src/audio/volume/Kconfig
```diff
@@ -76,4 +76,25 @@ config COMP_GAIN
...
+    default n
+    help
+        This option used to enable HIFI3 optimized volume code for sof build
+
+config VOLUME_GENERIC
+    bool
+    default y if !VOLUME_HIFI3 && !VOLUME_HIFI4
+    default n
+    help
+        This option used to enable volume generic code for sof build
```
lyakh, line 98: it makes no sense to enable more than one of these, right? They're mutually-exclusive? should this be a `choice`?

btian1: check core-isa.h generated during build, XCHAL_HAVE_HIFI3 and XCHAL_HAVE_HIFI4 are enabled together.

I am ok with choice, I make this PR as review status because I want to listen whether enable HIFI version based on platform is ok, I don't have other good ideas, if config it in Kconfig, this is the only way I can work out.

src/audio/volume/Kconfig
```diff
@@ -76,4 +76,25 @@ config COMP_GAIN
 	  This option enables gain to change volume. It works
 	  as peak volume without updating peak vol to host
 
+config VOLUME_HIFI4
+    bool
+    default y if ACE
```
kv2019i, line 81: Hmm, this won't work unless the toolchain has hifi support. E.g. Zephyr SDK supports ACE but can't handle hifi code.

kv2019i (commented):
@singalsu  can you check? Not sure this is feasible given we have a hard requirement on the toolchain.

src/audio/volume/Kconfig
btian1: @lyakh , we can' t use choice, as this is not manually determined, it is determined by the toolchain, which is further determined by the platform config, please check this version, I add a new build envionment varaible.

src/audio/volume/Kconfig
```diff
@@ -76,4 +76,25 @@ config COMP_GAIN
 	  This option enables gain to change volume. It works
 	  as peak volume without updating peak vol to host
 
+config VOLUME_HIFI4
+    bool
+    default y if "$(HIFI_VERSION)" = "4"
+    default n
```
andrula-song, line 82: do we really need this default n?

src/audio/volume/Kconfig
andrula-song: but actually, I think we can use a choice with default value depend on tool chain version.

scripts/xtensa-build-zephyr.py
```diff
@@ -743,6 +744,12 @@ def build_platforms():
 			TOOLCHAIN_VER = platform_dict["XTENSA_TOOLS_VERSION"]
 			XTENSA_CORE = platform_dict["XTENSA_CORE"]
 			platf_build_environ["TOOLCHAIN_VER"] = TOOLCHAIN_VER
+			hifi_version_index = XTENSA_CORE.lower().find("hifi")
+			if hifi_version_index < 0:
+				HIFI_VERSION = "generic"
+			else:
+				HIFI_VERSION = XTENSA_CORE[hifi_version_index+4]
```
andrula-song, line 751: may be  we can initialize the HIFI_VERSION = "generic"  and  just 
`			if hifi_version_index >= 0:
				HIFI_VERSION = XTENSA_CORE[hifi_version_index+4]
`

src/audio/volume/Kconfig
```diff
@@ -76,4 +76,25 @@ config COMP_GAIN
...
+config VOLUME_HIFI3
+    bool
+    default y if "$(HIFI_VERSION)" = "3"
+    default n
+    help
+        This option used to enable HIFI3 optimized volume code for sof build
+
+config VOLUME_GENERIC
+    bool
+    default y if "$(HIFI_VERSION)" = "generic" || "$(HIFI_VERSION)" = ""
```
andrula-song, line 95: seems better if we initialize the HIFI_VERSION = "generic" in scripts/xtensa-build-zephyr.py

marc-hb (dismissed):
Do not use environment variables for build configuration: they're evil. They're evil because they're invisible/"secret" and make build reproduction very difficult - even when you're lucky and found about them.

For instance, configuring the Cadence toolchain requires environment variables and it's a constant pain and maintenace overhead, examples:
- 25fc45143b5cca
- 44d9acb72d681824b9c
- etc.

Kconfig is not always easy to use but it is very flexible and powerful and enough for this, so you don't need to change the build script. Keep all the build configuration in a SINGLE place: Kconfig.

https://docs.zephyrproject.org/latest/build/kconfig/tips.html

Keep in mind this build script is OPTIONAL, many people invoke `west build` directly.

We also have CONFIG overlays for fine-tuning if needed.


@andyross could you provide some Kconfig / HiFi toolchain guidance for this?

scripts/xtensa-build-zephyr.py
btian1: I don't find >= in python....

btian1:
> Do not use environment variables for build configuration: they're evil. They're evil because they're invisible/"secret" and make build reproduction very difficult - even when you're lucky and found about them.
> 
> For instance, configuring the Cadence toolchain requires environment variables and it's a constant pain and maintenace overhead, examples:
> 
> * [25fc451](https://github.com/thesofproject/sof/commit/25fc45143b5cca1ac61eede6f19077fd8a5c8e73)
> * [44d9acb](https://github.com/thesofproject/sof/commit/44d9acb72d681824b9c1fcbf6dc07ff19d851aea)
> * etc.
> 
> Kconfig is not always easy to use but it is very flexible and powerful and enough for this, so you don't need to change the build script. Keep all the build configuration in a SINGLE place: Kconfig.
> 
> https://docs.zephyrproject.org/latest/build/kconfig/tips.html
> 
> Keep in mind this build script is OPTIONAL, many people invoke `west build` directly.
> 
> We also have CONFIG overlays for fine-tuning if needed.
> 
> @andyross could you provide some Kconfig / HiFi toolchain guidance for this?

Thanks, your point is also my vision to do this, however, I just can't find a better way to implement this, it is not easy to distinguish hifi3 and hifi4 in Kconfig, please share if you have better ideas.

I also noticed set_xtensa_params.sh for maintain, this is a patch work, also need add a new variable to do this.

marc-hb:
>  it is not easy to distinguish hifi3 and hifi4 in Kconfig,

I see you have used `choice` in https://github.com/thesofproject/sof/pull/8682/commits/818ecfd90b2d6c9b2, why is that not enough?
What's wrong with `sof/scripts/xtensa-build-zephyr.py mtl -C=-DCONFIG_VOLUME_HIFI3` ?

Also, what is the purpose of overriding the toolchain default with Kconfig? Who's going to use this? Why is something like `SRC_AUTOARCH` in e8afa16c33a17 not enough?

> this change makes volume version select based on

We see what this PR does but we don't see what problem(s) it tries to solve.

btian1:
> > it is not easy to distinguish hifi3 and hifi4 in Kconfig,
> 
> I see you have used `choice` in [818ecfd90b2d6c9b2](https://github.com/thesofproject/sof/commit/818ecfd90b2d6c9b2), why is that not enough? What's wrong with `sof/scripts/xtensa-build-zephyr.py mtl -C=-DCONFIG_VOLUME_HIFI3` ?
>  because we need set default as its origin toolchain HIFI version.
If add -D, current CI build need change, because default would be generic, I am not sure this is the right way.

> Also, what is the purpose of overriding the toolchain default with Kconfig? Who's going to use this? Why is something like `SRC_AUTOARCH` in [e8afa16](https://github.com/thesofproject/sof/commit/e8afa16c33a171ed3b7c9a15709f49a78f29bd45) not enough?
> its not overriding, it gives developer an option to select which version you want to have a try, sometimes, can be used for quick rootcause a bug that only happened in generic or HIFI version.

> > this change makes volume version select based on
> 
> We see what this PR does but we don't see what problem(s) it tries to solve.
it give developer flexbility to select which HIFI version for one module.

marc-hb:
> it gives developer an option to select which version you want to have a try, sometimes, can be used for quick rootcause a bug that only happened in generic or HIFI version.

If this is only for developers to debug then even Kconfig is not required because they can just change the source that they are already editing anyway. See the comment at the top of `src/audio/src_config.h` (commit e8afa16c33a171ed)

Also, why do it only for VOLUME? What about other code?

> If add -D, current CI build need change, because default would be generic, I am not sure this is the right way.

I don't understand this sorry. How is CI related to this?

kv2019i:
@btian1 @marc-hb I see two ways to add Kconfig support:

1) add a "VOLUME_HIGHEST_OPT_LEVEL" Kconfig option that is the default and implementaion is the current one (use highest optimization level the toolchain supports for this target)

2) move generic build to an overlay and require builds (e.g. in github CI flows) to use the overlay to pick up generic implemenations for all componets. the default build would use hifi3/4 for Intel targets, but you could only build the default configuration with xtensa toolchain

... I'd personally prefer (1). That keeps the current logic, but allows to explicitly control optimization with Kconfig if user wants to.

lgirdwood:
Kconfig needs to be able to override tooling (for downgrade) and rtos specific headers (like core-isa.h) to select the SIMD version based on ARCH. 

The intention is that Kconfig will select the SIMD config (i.e. be the single source of all configuration) for each component rather than XCC/core-isa.h (today) the extra benefit is we can downgrade SIMD (per module) when performance or logic bugs are detected as the SIMD code changes are large in many files.

kv2019i:
@lgirdwood One big practical problem is that we have a large set of users (e.g. public SOF, Zephyr CIs) that build e.g. "mtl" target but use a public toolchain that cannot support hifi builds. Then we have another large set of users who use the same config (e.g. "mtl") with xtensa toolchain. Currently the different code will be built (hifi enabled when building with xtensa toolchain).

Would my proposal (1) above we sufficient? We'd provide Kconfig options to override the implementation, but default would be kept as automatic. Otherwise we need to fork the configs (have "mtl" and "mtl_generic") and modify all CI instances and other infra to the new convention (this would be proposal (2) above).

lgirdwood:
> @lgirdwood One big practical problem is that we have a large set of users (e.g. public SOF, Zephyr CIs) that build e.g. "mtl" target but use a public toolchain that cannot support hifi builds. Then we have another large set of users who use the same config (e.g. "mtl") with xtensa toolchain. Currently the different code will be built (hifi enabled when building with xtensa toolchain).
> 
> Would my proposal (1) above we sufficient? We'd provide Kconfig options to override the implementation, but default would be kept as automatic. Otherwise we need to fork the configs (have "mtl" and "mtl_generic") and modify all CI instances and other infra to the new convention (this would be proposal (2) above).

Can we not reuse the ```config_compiler_version``` overlay that we can use for GCC and CI users. We also need to support integrating the public upstream clang HiFi support as it lands and this would mean we would have

1) Default config overlay - uses xcc/xt-clang (platform dependent), sets HiFi SIMD configs.
2) GCC config overlay - disables all Hifi SIMD.
3) upstream clang overlay - enables Hifi on a feater per feature basis as it lands upstream

marc-hb:
> add a "VOLUME_HIGHEST_OPT_LEVEL" Kconfig option that is the default and implementation is the current one (use highest optimization level the toolchain supports for this target)

Agreed: no code change and no overlay for the best, optimized default. If it ain't broke then don't fix it. This is similar to SRC_AUTOARCH=1 in e8afa16c33a171ed3b7c9.


If anyone or anything wants to temporarily downgrade VOLUME to HIFI3 or generic, then they just  run `xtensa-build-zephyr.py -C=-DCONFIG_VOLUME_OPT_HIFI3 mtl` or `xtensa-build-zephyr.py -C=-DCONFIG_VOLUME_OPT_GENERIC mtl` and that's it. Simple.

As usual with -DCONFIG_, the same can also be achieved with overlays but none is required.

btian1:
#8697 , I made this feature request to record possible solutions and we can discuss there for technical details, @marc-hb, @kv2019i @lgirdwood

lgirdwood:
> > add a "VOLUME_HIGHEST_OPT_LEVEL" Kconfig option that is the default and implementation is the current one (use highest optimization level the toolchain supports for this target)
> 
> Agreed: no code change and no overlay for the best, optimized default. If it ain't broke then don't fix it. This is similar to SRC_AUTOARCH=1 in [e8afa16](https://github.com/thesofproject/sof/commit/e8afa16c33a171ed3b7c9a15709f49a78f29bd45).
> 
> If anyone or anything wants to temporarily downgrade VOLUME to HIFI3 or generic, then they just run `xtensa-build-zephyr.py -C=-DCONFIG_VOLUME_OPT_HIFI3 mtl` or `xtensa-build-zephyr.py -C=-DCONFIG_VOLUME_OPT_GENERIC mtl` and that's it. Simple.
> 
> As usual with -DCONFIG_, the same can also be achieved with overlays but no overlay is needed for this.

This is ad-hoc. We all need to use the same single config for each compiler and it must come from Kconfig. 

The overlay worked well in the past and I now need to support 2 clang flavours - Cadence and Upstream clang.

src/audio/volume/Kconfig
```diff
@@ -8,6 +8,8 @@
 
 if COMP_VOLUME
 
+rsource "Kconfig.hifi"
```
lgirdwood, line 11: Kconfig.simd

src/audio/volume/Kconfig.hifi
```diff
@@ -0,0 +1,32 @@
+# SPDX-License-Identifier: BSD-3-Clause
+
+comment "volume optimization level select"
+
+choice "VOLUME_HIFI_LEVEL_SELECT"
+	prompt "choose which hifi level used for volume module"
+
+	config VOLUME_DEFAULT_OPT_LEVEL
```
lgirdwood, line 8: VOLUME_SIMD_NONE

src/audio/volume/Kconfig.hifi
```diff
@@ -0,0 +1,32 @@
...
+		prompt "dummy optimized option for volume module"
+		bool
+		help
+			This option is dummy option for volume module, when this was
```
lgirdwood, line 12: It's not dummy, it just generic C using standard arch ISA

src/audio/volume/Kconfig.hifi
```diff
@@ -0,0 +1,32 @@
...
+			selected, optimization level will be determined by compiler.
+
+	config VOLUME_HIFI3
+		prompt "choose hifi3 intrinsic optimized volume module"
```
lgirdwood, line 16: also needs to depend on ARCH_XTENSA && ARCH_HAS_HIFI3 (can be defined in SOC Kconfig), same for the other Hifi levels. These will both come from Zephyr, but I've no issue adding this PR now whilst the Zephyr Kconfigs are WIP for HiFi and then updating when Zephyr is merged.

src/audio/volume/Kconfig.hifi
```diff
@@ -0,0 +1,32 @@
+# SPDX-License-Identifier: BSD-3-Clause
+
+comment "volume optimization level select"
+
+choice "VOLUME_HIFI_LEVEL_SELECT"
+	prompt "choose which hifi level used for volume module"
+
+	config VOLUME_DEFAULT_OPT_LEVEL
+		prompt "dummy optimized option for volume module"
```
lgirdwood, line 9: "No SIMD optimization."

src/audio/volume/Kconfig.hifi
```diff
@@ -0,0 +1,32 @@
...
+		help
+			This option used to build HIFI3 intrinsic optimized volume code
+
+	config VOLUME_HIFI4
+		prompt "choose hifi4 intrinsic optimized volume module"
+		bool
+		help
+			This option used to build HIFI4 optimized volume code
+
+	config VOLUME_GENERIC
```
lgirdwood, line 27: This is the same as no SIMD

src/audio/volume/volume.h
```diff
@@ -36,15 +36,17 @@ struct sof_ipc_ctrl_value_chan;
...
+#  define CONFIG_VOLUME_HIFI4    1
+#elif XCHAL_HAVE_HIFI3
+#  define CONFIG_VOLUME_HIFI3    1
 #else
-# define VOLUME_GENERIC
+#  define CONFIG_VOLUME_GENERIC  1
+#endif
+#endif
+#else
+#  define CONFIG_VOLUME_GENERIC  1
```
lgirdwood, line 49: All this is NOT needed any more.

src/audio/volume/Kconfig.hifi
kv2019i: I think the intention here is "VOLUME_SIMD_AUTOMATIC" or "VOLUME_SIMD_TOOLCHAIN_MAX". I.e. option that captures the current behaviour, enabled SIMD and use the highest level supported by toolchain.

src/audio/volume/Kconfig.hifi
kv2019i: So not dummy, but not generic either, see above.

src/audio/volume/volume.h
kv2019i: @lgirdwood It is needed still, otherwise this PR cannot be submitted before we have defined a generic overlay and converted the github flows to use it.

kv2019i (commented):
We are still not on the same page @lgirdwood @btian1 . If we drop the toolchain based ifdefs for default buiod, then we need a new overlay and convert the github workflows to use the new overlay for generic builds, We cannot merge the PR otherwise.

The approach in this PR could be merged (with the hifi/simd namign changed).

src/audio/volume/volume.h
lgirdwood: @kv2019i @btian1 the important part is to remove the dependence on xtos headers above and duplication of Kconfig, so this has to go.
@btian1 we will need for volume currently a
1) ```app/overlays/<platform>/cc_default.conf``` - uses default production values for xcc/xt-clang.
2) ```app/overlays/<platform>/gcc_default.conf``` - uses non HIFI values for xtensa GCC.
Once 1 & 2 are ready then we can plumb into the GH CI workflows.

btian1:
@lgirdwood @kv2019i 
In short, if want to implement based on Liam's idea, we need have a default loading conf for each platform.
Liam's proposal is app/overlay/platform/platform.conf, however, this need modify python build script, right?

Do you accept put simd configs in the bottom of file(of course, each platform will add): https://github.com/thesofproject/sof/blob/main/app/boards/intel_adsp_ace15_mtpm.conf

there already have conf file for each platform, we can simply append it,

src/audio/volume/volume.h
kv2019i: @btian1  @marc-hb  Discussed this with @lgirdwood  and one new idea. So starting premise is we want to get rid of the core-isa.h based logic in the source files and drive this logic from Kconfig). 

To make transition easier, what if we reverse the logic in volume.h and detect a Zephyr SDK build (plain gcc), and if detected, force VOLUME_GENERIC. 

This would allow to keep our (and Zephyr's) github workflows unmodified, but move to Kconfig-based selection of SIMD level for non-cloud builds.

src/audio/volume/Kconfig
btian1: simd means single instruction multiple data, for example, a 32-bit register, it can load 2 16-bit to perform 2 times multiply with one instruction, no matter hifhi3/4/5, both support same simd, however, the difference is the numbers: i.e. how many multiply unit it have in current HIFI DSP, for example, HIFI3 can perform 3 32-bit multiply in one cycle, HIFI4 can perform 4 times 32-bit multiple in one cycle. hifi is more accurate than simd, we are choosing hifi version with this kconfig.

marc-hb:
I think I understand the benefit of using Kconfig to downgrade Hifi on a per-component basis for testing, troubleshooting and measuring performance, I imagine it's worth the very large KConfig additions.

On the other hand, I don't understand why this implies dropping automated, core-isa.h-based configuration that has been successful since forever, down to new, manual, long (per-toolchain?) overlays that manually duplicate the information from core-isa.h. Eventually people will make some mistake or forget something in these new overlays for one component for one platform and no one will notice.

Also, this will catch by surprise people not using SOF build scripts (= everyone in the Zephyr team) and will add an extra configuration burden past the element of surprise. Then they will forget and be surprised again.

In any case these brand new overlays are a different request that can be done later.

On the other, other hand, I totally understand being fed-up with the massive XCHAL_HAVE / core-isa.h  duplication. Having to "pollute" every component.h with that same `#if XCHAL_HAVE_HIFIx...` block of macros has always been ridiculous. So I used DCBLOCK as another, lead example and I just shared #8720 which shows how we can easily remove all that duplication - and add a new CONFIG_DCBLOCK_HIFI at the same time.

marc-hb:
> ... for non-cloud builds.

Toolchains topics are complex, messy and confusing enough so let's please not use creative and changing names like "cloud builds",  "plain gcc", "github workflow",... 

AFAIK there 3 toolchain types:

- Zephyr SDK. 
- Cadence toolchains (`#ifdef __XCC__`) Can be either xt-gcc or xt-clang: doesn't matter here.
- Custom SOF toolchains. Never used with Zephyr. Unfortunately misnamed "GCC" in https://thesofproject.github.io/latest/introduction/index.html#toolchain-faq 

And that's all the toolchain names we need?

Similarly, please also name build configurations for what they actually are, not after who's using them the most. I mean there is no such thing as "CI build configuration" or any other secret CI configuration. Either some build configuration is valid or it's not. If it's not valid then it should be removed from every CI. If it's valid then it's not just for CI and everyone must be able to configure and use it in less than 15 seconds. 

As you know sof-jenkins CI uses Cadence toolchains whereas Github uses the Zephyr SDK so "CI build configuration" is very meaningless in the current context.

plbossart:
I am clearly missing the problem statement. 

If we are on a platform with HiFi3, use HiFi3
If we are on a platform with HiFi4, use HiFi4.

Why would anyone start changing the type of code generated on a per-component basis? This is asking for trouble, no?

src/audio/volume/Kconfig
lgirdwood: simd is a generic term, we will be using the same method/file for ARM and x86 SIMD so pls use SIMD

lgirdwood:
> I am clearly missing the problem statement.
> 
> If we are on a platform with HiFi3, use HiFi3 If we are on a platform with HiFi4, use HiFi4.
> 
> Why would anyone start changing the type of code generated on a per-component basis? This is asking for trouble, no?

In general yes, but use of SIMD is growing alongside versions, architectures, complexity and compiler support. i.e. the codebase is diverging around SIMD and this means we will have mature and immature implementations as we further grow our usage. i.e. we may need to force downgrade a module due to a bug whilst keeping other modules at HW ISA level.

Notwithstanding the above, we also have to integrate upstream clang HiFi compiler support and that will be on an incremental module and isa level basis i.e. we will have a Kconfig for this that will determine what upstream clang can safely build around Hifi SIMD.

Finally, Kconfig will be the single source of configuration for all architectures and this is not true today.

marc-hb:
A future `MTL.HIFIdefconfig` can be the single source of HiFi truth *and* be as simple as: "Use the max HiFi level available by default _except_ for these two modules VOLUME and DCBLOCK". Two-lines long, done! (Just an example).

This does not require duplicating the max available HiFi level for every module and every platform, accidentally missing some modules, making typos in the name of others and forgetting to copy/paste some name changes. CONFIGuration mistakes are frequent but much more so in long, duplicated files.

btian1:
@marc-hb @lgirdwood @kv2019i 

https://github.com/zephyrproject-rtos/zephyr/pull/67513/commits/512407edc9d8cffe80d7fb9a75cc923b49b2f2fb

From above link information, at least, NXP still have dependency on IPC3 xtos build, which means we still can't fully depend on
zephyr to pass HIFIn information to module, maybe next year is a suitable time to remove core-isa dependency.
 
So we can take this PR as first step, i.e:
1. Still use core-isa as first choice for HIFI version.
2. Add a Choice( in this patch) for each module to make each module are configurable first.
3. In 2025, after all users are switched to zephyr based SOF, remove core-isa dependency with #67513.

Please share your comments on this, above is my rough thought, if you are ok with this, I will address some review comments and update PR.

marc-hb:
> Still use core-isa as first choice for HIFI version.

This is the simplest way to make incremental progress and this is what I did in my very first attempt: #8720 . Dropping the `#include core-isa.h` is not required to achieve per-component Kconfigurability, that's a different requirement which clashes with XTOS platforms.

btian1:
> > Still use core-isa as first choice for HIFI version.
> 
> This is the simplest way to make incremental progress and this is what I did in my very first attempt: #8720 . Dropping the `#include core-isa.h` is not required to achieve per-component Kconfigurability, that's a different requirement which clashes with XTOS platforms.

ok, we can use two step to meet orginal thought, 

1. use current PR to add build option for each module.
2. once all SOF users are switched to zephyr, remove core-isa dependency( may need wait for one year, as Daniel mentioned)

lgirdwood (approved):
LGTM, I assume @marc-hb also aligned ?

src/audio/volume/volume.h
```diff
@@ -36,15 +36,17 @@ struct sof_ipc_ctrl_value_chan;
...
-#  define VOLUME_HIFI3
+#  define CONFIG_VOLUME_HIFI3    1
 # else
-#  define VOLUME_GENERIC
+#  define CONFIG_VOLUME_GENERIC  1
 # endif
+#endif
 #else
-# define VOLUME_GENERIC
+#  define CONFIG_VOLUME_GENERIC  1
```
marc-hb, line 49: We don't have to keep duplicating this `#ifdef` logic in every single C file. I just submitted #8787 which is equivalent to the current state of this PR but without this duplication in every C file.

src/audio/volume/Kconfig.simd
```diff
@@ -0,0 +1,32 @@
...
+
+	config VOLUME_HIFI3
+		prompt "choose HIFI3 intrinsic optimized volume module"
+		bool
+		help
+			This option used to build HIFI3 intrinsic optimized volume code
+
+	config VOLUME_HIFI4
+		prompt "choose HIFI4 intrinsic optimized volume module"
+		bool
```
andrula-song, line 23: I still think we should display the choices depend on the specific compiled target devices, for example, we should not show the VOLUME_HIFI4 if compile a tgl binary.

marc-hb: Yes this would be better but it requires duplicating all the `core-isa.h` files in Kconfig TWICE: once in Zephyr and another time in XTOS. This has been just discussed above, in my prototype https://github.com/zephyrproject-rtos/zephyr/pull/67513 + #8733 and elsewhere.

For now let's just provide the per-component selection (that most users should not change) and revisit later.

marc-hb: It would be a fairly minor convenience: very few people are expected to change these and the few people who will use these are expected to know what is the default that they are manually overriding.

src/audio/volume/volume.c
```diff
@@ -587,7 +587,7 @@ static int volume_process(struct processing_module *mod,
 		cd->peak_cnt = 0;
 		peak_vol_update(cd);
 		memset(cd->peak_regs.peak_meter, 0, sizeof(cd->peak_regs.peak_meter));
-#ifdef VOLUME_HIFI4
+#if CONFIG_VOLUME_HIFI4
```
marc-hb, line 590: why not this?
```suggestion
#if SOF_USE_HIFI(4, VOLUME)
```

src/audio/volume/volume_hifi4.c
```diff
@@ -21,7 +21,7 @@ LOG_MODULE_DECLARE(volume_hifi4, CONFIG_SOF_LOG_LEVEL);
 
 #include "volume.h"
 
-#ifdef VOLUME_HIFI4
+#if SOF_USE_HIFI(4, VOLUME)
```
marc-hb, line 24: Ask @andrula-song to confirm but I think you need this until volume_hifi5.c exists.
```suggestion
#if SOF_USE_HIFI(4, VOLUME) || SOF_USE_HIFI(5, VOLUME)
```

src/audio/volume/volume_hifi4.c
```diff
@@ -21,7 +21,7 @@ LOG_MODULE_DECLARE(volume_hifi4, CONFIG_SOF_LOG_LEVEL);
 
 #include "volume.h"
 
-#ifdef VOLUME_HIFI4
+#if SOF_USE_HIFI(4, VOLUME) || SOF_USE_HIFI(5, VOLUME)
```
kv2019i, line 24: Not sure if we want to add the HIFI5 case in same patchset, but no big harm either,

kv2019i (approved):

src/audio/volume/volume.c
marc-hb: Do not resolve comments without a single word or single change

src/audio/volume/volume_hifi4.c
marc-hb: `SOF_USE_HIFI(5,... )` is now required because I already added HIFI5 in common.h. So, without `SOF_USE_HIFI(5,... )` here, the linker would fail to compile any platform that HAS_HIFI5 with the error "volume code not found".

> I already added HIFI5 in common.h.

Was that premature? If it was premature then feel free to remove it.

marc-hb (dismissed):
Please explain the CONFIG_VOLUME_HIFI4 oddity. I spent a couple more minutes looking at it and I couldn't explain it.

Maybe it's correct for some reason but it needs at least an explanation.

src/audio/volume/volume.c
btian1: yes, just too much comments and I don't want to use a new PR to submit this patch, let's go for merge.

src/audio/volume/volume_hifi4.c
btian1: Add HIFI5 have below advantage:
1. no error, since it is || relation.
2. If add new platform build, like PTL and NVL in near future, there is no HIFI5 implementation for some modules, with this adding, it can make sure still use HIFI4 as best for each module.
3. it is also a reminder for later module HIFI5 adding.

btian1 (commented):
Thanks, @kv2019i @marc-hb , it is ready for merge now.

andrula-song:
please modify the alignment set in volume_set_alignment as well to fit your new style, thanks.

marc-hb (approved):

kv2019i:
One pause-resume fail in https://sof-ci.01.org/sofpr/PR8682/build2430/devicetest/index.html , but does not seem related to this PR. Rest of CI passing, so proceeding with merge.

PR #8685, opened 2024-01-03
Better handling of module base_config_ext data

andyross:
This is a different take on a fix for #8639 than the workaround in #8575 on which it was originally coded.  It removes the requirement for components to have to muck with module internals[1] and places the complicated byte-offset math into just one place (in the module code that owns the struct format).  Really I think this is a lot cleaner.  If you guys want to merge the first version for clarity (given it's already "shipped"[2]), I can submit this as cleanup on top of that instead (which, again, is how this was written originally anyway).

With this, the big IPC4 workaround patch at the end of the AEC work is down to just one line, hope to have that cleaned up by tomorrow.  (see https://github.com/andyross/sof/commits/mtl-aec-tmp/ for the current status of my working tree).

[1] Basically, the first version: (1) requires every module to make a copy of the untyped "init_data" field of module_config, then (2) store that pointer **back** into a different field of module_config (because the first one points into the IPC command buffer and won't survive) where (3) it will be eventually used, **not** by the module code itself but by the ipc4/module layer that created it in the first place.  This avoids all that by having the module_adapter init code just parse and store the data at the start.

[2] In the MTL "stable" branch, where it's already been merged.  Can I just whine a bit about how frustrating it is to have to work across branches like this?  I didn't realize until I was about to submit that the code I was modifying wasn't actually upstream!

### DIFF
--- a/src/audio/module_adapter/module_adapter.c
+++ b/src/audio/module_adapter/module_adapter.c
@@ -1333,6 +1333,10 @@ int module_adapter_reset(struct comp_dev *dev)
 	rfree(mod->stream_params);
 	mod->stream_params = NULL;
 
+#if CONFIG_IPC_MAJOR_4
+	rfree(mod->priv.cfg.input_pins);
+#endif
+
 	comp_dbg(dev, "module_adapter_reset(): done");
 
 	return comp_set_state(dev, COMP_TRIGGER_RESET);
--- a/src/audio/module_adapter/module_adapter_ipc4.c
+++ b/src/audio/module_adapter/module_adapter_ipc4.c
@@ -38,19 +38,39 @@ int module_adapter_init_data(struct comp_dev *dev,
 			     const struct comp_ipc_config *config,
 			     const void *spec)
 {
-	if (dev->drv->type == SOF_COMP_MODULE_ADAPTER) {
-		const struct ipc_config_process *ipc_module_adapter = spec;
+	const struct ipc_config_process *args = spec;
+	const struct ipc4_base_module_extended_cfg *cfg = (void *)args->data;
+	size_t cfgsz = args->size;
 
-		dst->init_data = ipc_module_adapter->data;
-		dst->size = ipc_module_adapter->size;
-		dst->avail = true;
+	assert(dev->drv->type == SOF_COMP_MODULE_ADAPTER);
+	if (cfgsz < sizeof(cfg->base_cfg))
+		return -EINVAL;
 
-		memcpy_s(&dst->base_cfg,  sizeof(dst->base_cfg), ipc_module_adapter->data,
-			 sizeof(dst->base_cfg));
-	} else {
-		dst->init_data = spec;
+	dst->base_cfg = cfg->base_cfg;
+	dst->size = cfgsz;
+
+	if (cfgsz >= sizeof(*cfg)) {
+		int n_in = cfg->base_cfg_ext.nb_input_pins;
+		int n_out = cfg->base_cfg_ext.nb_output_pins;
+		size_t pinsz = (n_in * sizeof(*dst->input_pins))
+			     + (n_out * sizeof(*dst->output_pins));
+
+		if (cfgsz == (sizeof(*cfg) + pinsz)) {
+			dst->nb_input_pins = n_in;
+			dst->nb_output_pins = n_out;
+			dst->input_pins = rmalloc(SOF_MEM_ZONE_RUNTIME_SHARED,
+						  0, SOF_MEM_CAPS_RAM, pinsz);
+			if (!dst->input_pins)
+				return -ENOMEM;
+
+			dst->output_pins = (void *)&dst->input_pins[n_in];
+			memcpy_s(dst->input_pins, pinsz,
+				 &cfg->base_cfg_ext.pin_formats[0], pinsz);
+		}
 	}
 
+	dst->init_data = cfg; /* legacy API */
+	dst->avail = true;
 	return 0;
 }
 
--- a/src/include/module/module/base.h
+++ b/src/include/module/module/base.h
@@ -30,6 +30,10 @@ struct module_config {
 	const void *init_data; /**< Initial IPC configuration. */
 #if CONFIG_IPC_MAJOR_4
 	struct ipc4_base_module_cfg base_cfg;
+	uint8_t nb_input_pins;
+	uint8_t nb_output_pins;
+	struct ipc4_input_pin_format *input_pins;
+	struct ipc4_output_pin_format *output_pins;
 #endif
 };
 
--- a/src/ipc/ipc4/helper.c
+++ b/src/ipc/ipc4/helper.c
@@ -8,7 +8,9 @@
 #include <sof/audio/buffer.h>
 #include <sof/audio/component.h>
 #include <sof/audio/component_ext.h>
+#include <sof/audio/module_adapter/module/generic.h>
 #include <sof/audio/pipeline.h>
+#include <module/module/base.h>
 #include <sof/common.h>
 #include <rtos/interrupt.h>
 #include <sof/ipc/topology.h>
@@ -431,6 +433,9 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 	struct ipc4_base_module_cfg source_src_cfg;
 	struct ipc4_base_module_cfg sink_src_cfg;
 	uint32_t flags;
+	uint32_t ibs = 0;
+	uint32_t obs = 0;
+	uint32_t buf_size;
 	int src_id, sink_id;
 	int ret;
 
@@ -453,17 +458,42 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 	if (!cpu_is_me(source->ipc_config.core) && !cross_core_bind)
 		return ipc4_process_on_core(source->ipc_config.core, false);
 
-	/* these might call comp_ipc4_get_attribute_remote() if necessary */
-	ret = comp_get_attribute(source, COMP_ATTR_BASE_CONFIG, &source_src_cfg);
-	if (ret < 0) {
-		tr_err(&ipc_tr, "failed to get base config for module %#x", dev_comp_id(source));
-		return IPC4_FAILURE;
+	struct processing_module *srcmod = comp_get_drvdata(source);
+	struct processing_module *dstmod = comp_get_drvdata(sink);
+	struct module_config *dstcfg = &dstmod->priv.cfg;
+	struct module_config *srccfg = &srcmod->priv.cfg;
+
+	/* get obs from the base config extension if the src queue ID is non-zero */
+	if (bu->extension.r.src_queue && bu->extension.r.src_queue < srccfg->nb_output_pins)
+		obs = srccfg->output_pins[bu->extension.r.src_queue].obs;
+
+	/* get obs from base config if src queue ID is 0 or if base config extn is missing */
+	if (!obs) {
+		/* these might call comp_ipc4_get_attribute_remote() if necessary */
+		ret = comp_get_attribute(source, COMP_ATTR_BASE_CONFIG, &source_src_cfg);
+
+		if (ret < 0) {
+			tr_err(&ipc_tr, "failed to get base config for src module %#x",
+			       dev_comp_id(source));
+			return IPC4_FAILURE;
+		}
+		obs = source_src_cfg.obs;
 	}
 
-	ret = comp_get_attribute(sink, COMP_ATTR_BASE_CONFIG, &sink_src_cfg);
-	if (ret < 0) {
-		tr_err(&ipc_tr, "failed to get base config for module %#x", dev_comp_id(sink));
-		return IPC4_FAILURE;
+	/* get ibs from the base config extension if the sink queue ID is non-zero */
+	if (bu->extension.r.dst_queue && bu->extension.r.dst_queue < dstcfg->nb_input_pins)
+		ibs = dstcfg->input_pins[bu->extension.r.dst_queue].ibs;
+
+	/* get ibs from base config if sink queue ID is 0 or if base config extn is missing */
+	if (!ibs) {
+		ret = comp_get_attribute(sink, COMP_ATTR_BASE_CONFIG, &sink_src_cfg);
+		if (ret < 0) {
+			tr_err(&ipc_tr, "failed to get base config for sink module %#x",
+			       dev_comp_id(sink));
+			return IPC4_FAILURE;
+		}
+
+		ibs = sink_src_cfg.ibs;
 	}
 
 	/* create a buffer
@@ -472,12 +502,10 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 	 * in case of DP -> LL
 	 *	size = 2*ibs of destination (LL) module. DP queue will handle obs of DP module
 	 */
-	uint32_t buf_size;
-
 	if (source->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_LL)
-		buf_size = source_src_cfg.obs * 2;
+		buf_size = obs * 2;
 	else
-		buf_size = sink_src_cfg.ibs * 2;
+		buf_size = ibs * 2;
 
 	buffer = ipc4_create_buffer(source, cross_core_bind, buf_size, bu->extension.r.src_queue,
 				    bu->extension.r.dst_queue);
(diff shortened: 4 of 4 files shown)


andyross:
> I didn't realize until I was about to submit that the code I was modifying wasn't actually upstream!

As in, I wrote up a whole commit message carefully explaining and advocating my removal of an API that never merged, heh:

https://github.com/andyross/sof/commit/37f784f7af6369c77e3d09d9551dae620e2b7d2e

src/audio/module_adapter/module_adapter_ipc4.c
```diff
@@ -38,19 +38,39 @@ int module_adapter_init_data(struct comp_dev *dev,
...
+			dst->nb_input_pins = ni;
+			dst->nb_output_pins = no;
+			dst->input_pins = rmalloc(SOF_MEM_ZONE_RUNTIME_SHARED,
+						  0, SOF_MEM_CAPS_RAM, pinsz);
+			if (!dst->input_pins)
+				return -ENOMEM;
+
+			dst->output_pins = (void *) &dst->input_pins[ni];
+			memcpy_s(dst->input_pins, pinsz,
+				 &cfg->base_cfg_ext.pin_formats[0], pinsz);
```
btian1, line 68: memcpy? since size are same.

src/audio/module_adapter/module_adapter_ipc4.c
```diff
@@ -38,19 +38,39 @@ int module_adapter_init_data(struct comp_dev *dev,
 			     const struct comp_ipc_config *config,
 			     const void *spec)
 {
-	if (dev->drv->type == SOF_COMP_MODULE_ADAPTER) {
-		const struct ipc_config_process *ipc_module_adapter = spec;
+	const struct ipc_config_process *cp = spec;
```
btian1, line 41: cp is too simple, how about ipc_config_data?

src/audio/module_adapter/module_adapter_ipc4.c
```diff
@@ -38,19 +38,39 @@ int module_adapter_init_data(struct comp_dev *dev,
...
+	dst->base_cfg = cfg->base_cfg;
+	dst->size = cp->size;
+
+	if (cp->size >= sizeof(*cfg)) {
+		int ni = cfg->base_cfg_ext.nb_input_pins;
+		int no = cfg->base_cfg_ext.nb_output_pins;
+		size_t pinsz = (ni * sizeof(*dst->input_pins))
+			+ (no * sizeof(*dst->output_pins));
+
+		if (cp->size == (sizeof(*cfg) + pinsz)) {
```
btian1, line 58: suggest add some comment for this condition.

src/audio/module_adapter/module_adapter_ipc4.c
```diff
@@ -38,19 +38,39 @@ int module_adapter_init_data(struct comp_dev *dev,
...
+			dst->nb_input_pins = ni;
+			dst->nb_output_pins = no;
+			dst->input_pins = rmalloc(SOF_MEM_ZONE_RUNTIME_SHARED,
+						  0, SOF_MEM_CAPS_RAM, pinsz);
+			if (!dst->input_pins)
+				return -ENOMEM;
+
+			dst->output_pins = (void *) &dst->input_pins[ni];
```
btian1, line 66: line 66 also need comments.

src/ipc/ipc4/helper.c
```diff
@@ -453,17 +453,42 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
+
+	/* get ibs from base config if sink queue ID is 0 or if base config extn is missing */
+	if (!ibs) {
+		ret = comp_get_attribute(sink, COMP_ATTR_BASE_CONFIG, &sink_src_cfg);
+		if (ret < 0) {
+			tr_err(&ipc_tr, "failed to get base config for sink module %#x",
+			       dev_comp_id(sink));
+			return IPC4_FAILURE;
+		}
+
```
btian1, line 490: line 475 does not have one more space line, this can be removed as well for align.

src/ipc/ipc4/helper.c
```diff
@@ -453,17 +453,42 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
...
 
-	/* these might call comp_ipc4_get_attribute_remote() if necessary */
-	ret = comp_get_attribute(source, COMP_ATTR_BASE_CONFIG, &source_src_cfg);
-	if (ret < 0) {
-		tr_err(&ipc_tr, "failed to get base config for module %#x", dev_comp_id(source));
-		return IPC4_FAILURE;
+	struct processing_module *srcmod = comp_get_drvdata(source);
+	struct processing_module *dstmod = comp_get_drvdata(sink);
+	struct module_config *dstcfg = &dstmod->priv.cfg;
+	struct module_config *srccfg = &srcmod->priv.cfg;
```
kv2019i, line 459: Something missing in the patch, "struct processing_module" definition is not known and there are lot of fails in CI (defined in module/base.h). Now this is supposed to be private to module implementations, which is why I think @ranj063  opted to add a new property so it can be "set from outside".

kv2019i (commented):
This does look cleaner. I'll let others chime in. Ext-cfg is optional, but given it's more common to have it, don't see why not to handle it in common code. There are some corner cases I'd like others to chime in.

src/audio/module_adapter/module_adapter_ipc4.c
```diff
@@ -38,19 +38,39 @@ int module_adapter_init_data(struct comp_dev *dev,
...
-	if (dev->drv->type == SOF_COMP_MODULE_ADAPTER) {
-		const struct ipc_config_process *ipc_module_adapter = spec;
+	const struct ipc_config_process *cp = spec;
+	const struct ipc4_base_module_extended_cfg *cfg = (void *)cp->data;
+	size_t cfgsz = cp->size;
 
-		dst->init_data = ipc_module_adapter->data;
-		dst->size = ipc_module_adapter->size;
-		dst->avail = true;
+	assert(dev->drv->type == SOF_COMP_MODULE_ADAPTER);
```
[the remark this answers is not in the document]
kv2019i: It would seem this is ok. Most modules use "DECLARE_MODULE_ADAPTER()" to register the driver (which sets the type correctly) and library_manager calls "declare_dynamic_module_adapter(drv, SOF_COMP_MODULE_ADAPTER, ..)". So I don't see any use for IPC4 where the type would not be set to adapter. It still bugs me a bit a specific branch was added to cover this case and I can't find from git history any use for this, but maybe this was some variant of loadable modules. @jxstelter @ranj063 @RanderWang  @softwarecki @pjdobrowolski please do comment if I got this wrong.

src/ipc/ipc4/helper.c
ranj063: @andyross I do like the idea of parsing/saving the base_cfg_ext data in the module adapter instead of doing it in the individual modules. But I think you're going to have to add a new property to retrieve the extn data just like we do for the base_cfg data today.

src/audio/module_adapter/module_adapter_ipc4.c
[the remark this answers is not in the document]
andyross: That was my read, yeah.  The original had a check for that that then set the init_data unconditionally, which seems a little dangerous (what if it's not?  what's in the data and how do we prevent treating it like pin config?".  I grepped around and couldn't find any situations where it would happen, so I replaced it with an assert to be sure, figuring that CI would tell me if I missed something.

src/ipc/ipc4/helper.c
andyross: I see the failures, presumably that's just a header ordering issue?  Or if processing_module is an illegal header in some contexts, can you provide details?  It would be easy enough to wrap these behind an API.  That seems unlikely though, as the alternate code was already banging directly on the mod->priv.cfg.

@ranj063 you mean the COMP_ATTR_BASE_CONFIG_EXT from the version in your first PR?  That's actually deliberately omitted: the only spot that would use it is using the original/kernel-derived data in the new parsed fields.    The reason for making this a property would presumably be that components could then override it, but (and, OK, this is becoming a crusade of mine I guess) I think that's emphatically bad and should be avoided.  If topology lied and gave us an invalid value, the job of the firmware is to fail gracefully and get the integrator to fix the topology.

src/audio/module_adapter/module_adapter_ipc4.c
andyross: I actually agree as a point of style (the code computes the size once, validates against the input blob size, then memcpy's the amount of data it allocates).  Nonetheless there's a checkpatch warning about the use of bare memcpy(), so it seems like I can't win. :)

src/audio/module_adapter/module_adapter_ipc4.c
```diff
@@ -38,19 +38,39 @@ int module_adapter_init_data(struct comp_dev *dev,
...
+		return -EINVAL;
 
-		memcpy_s(&dst->base_cfg,  sizeof(dst->base_cfg), ipc_module_adapter->data,
-			 sizeof(dst->base_cfg));
-	} else {
-		dst->init_data = spec;
+	dst->base_cfg = cfg->base_cfg;
+	dst->size = cp->size;
+
+	if (cp->size >= sizeof(*cfg)) {
```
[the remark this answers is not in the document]
andyross: Well to be fair, this check is needed to prevent a buffer overrun during parsing.  It's true that there's also a typesafety issue in knowing "what kind of extra data do we have", but that's a protocol bug not really addressed by this PR :)

andyross:
Cleaned up the backport from mtl-007-drop-stable (the only place I can test this) that rebases this on top of a reversion that should match main better.  Hopefully will fix the build issues.

src/ipc/ipc4/helper.c
ranj063: @andyross I think originally processing_module was intended to be private data accessible only for the module itself. But given that we are already accessing the base_cfg and base_cfg_ext are being accessed by the common code, I guess it makes sense to make it shared. By the same argument maybe we should remove the COMP_ATTR_BASE_CONFIG attribute as well.

src/ipc/ipc4/helper.c
```diff
@@ -472,8 +502,6 @@ int ipc_comp_connect(struct ipc *ipc, ipc_pipe_comp_connect *_connect)
 	 * in case of DP -> LL
 	 *	size = 2*ibs of destination (LL) module. DP queue will handle obs of DP module
 	 */
-	uint32_t buf_size;
-
 	if (source->ipc_config.proc_domain == COMP_PROCESSING_DOMAIN_LL)
 		buf_size = source_src_cfg.obs * 2;
```
ranj063, line 506: shouldn't this be changed to obs * 2? and similarly ibs * 2 below?

ranj063: but more simply isnt it enough to do MAX(ibs * 2, obs * 2)?

src/audio/module_adapter/module_adapter_ipc4.c
```diff
@@ -38,19 +38,39 @@ int module_adapter_init_data(struct comp_dev *dev,
...
 
-		memcpy_s(&dst->base_cfg,  sizeof(dst->base_cfg), ipc_module_adapter->data,
-			 sizeof(dst->base_cfg));
-	} else {
-		dst->init_data = spec;
+	dst->base_cfg = cfg->base_cfg;
+	dst->size = cfgsz;
+
+	if (cfgsz >= sizeof(*cfg)) {
+		int ni = cfg->base_cfg_ext.nb_input_pins;
```
cujomalainey, line 53: little more verbose with names here please

src/audio/module_adapter/module_adapter.c
```diff
@@ -1333,6 +1333,10 @@ int module_adapter_reset(struct comp_dev *dev)
 	rfree(mod->stream_params);
 	mod->stream_params = NULL;
 
+#if CONFIG_IPC_MAJOR_4
+	rfree(mod->priv.cfg.input_pins);
+#endif
+
```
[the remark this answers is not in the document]
kv2019i: @andyross  ^^ We can do this in a follow-up as well, but in general we want to move the IPC specifc code in separate compilation units to keep the amount of ifdefs in generic code in check.

src/audio/module_adapter/module_adapter_ipc4.c
kv2019i: Hmm, I'm good with this even without comments. To follow this code, reader needs to have definition of the parsed binary blobs at hand.

kv2019i (approved):
CI looks good now. By directly accessing module internals is a change in convention, but we seem to have consensus this is ok in this case.

src/include/module/module/base.h
```diff
@@ -30,6 +30,10 @@ struct module_config {
 	const void *init_data; /**< Initial IPC configuration. */
 #if CONFIG_IPC_MAJOR_4
 	struct ipc4_base_module_cfg base_cfg;
+	uint8_t nb_input_pins;
```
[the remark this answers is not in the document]
andyross: They're actually different values from different sources.  The fields in processing_module count the number of actual comp_buffer objects connected to the module.  The one here is a field set in an IPC message from the kernel based on topology input.   In the case where the base_cfg_ext is provided, they surely should be the same value.  But the extended config is actually not always present (some components have their own config structs here instead), so they might differ at runtime and we need to know how much memory was actually allocated behind the pointers here.

src/ipc/ipc4/helper.c
andyross: Agreed, I'll look at that as a followup cleanup.  The "kpb" component is the only spot I see in current code that's implementing the BASE_CONFIG attribute, and it's just echoing back the default base config anyway.

andyross:
Grrr... sorry.  Came back to check CI and realized that I'd mis-squashed the naming fixup and put it in the wrong patch.  I've already torn my tree up for different work though so can't get back to fix it in the immediate term.  (I'm juggling like six PRs right now, heh).  Don't merge until I get the corrected version in

src/include/module/module/base.h
```diff
@@ -30,6 +30,10 @@ struct module_config {
...
+	uint8_t nb_output_pins;
```
lyakh, line 34: nitpick: is "nb" an abbreviation for "number?" If so, it seems a bit non-standard to me, maybe just "n" or "num" or even "number_of..."

src/ipc/ipc4/helper.c
ranj063: @andyross not changing the buffer sizing according to the ibs/obs based on pin numbers would be incorrect isnt it? Let's assume we're binding modules with a non-zero pin ID's, what would sink_src_cfg.ibs or source_src_cfg.obs contain?

src/include/module/module/base.h
andyross: It's the pre-existing name for the field in struct ipc4_base_module_cfg_ext from which this is derived.  Personally I don't like it much either, but I'm trying to be a good citizen and not rock the boat too much.  :)

src/ipc/ipc4/helper.c
andyross: Oh.... gotcha.  I had missed that the calculation here was still done with the source_src_cfg value and not the recalculated-and-now-pin-aware values from above.   Interestingly, this patch actually started as a refactor of the code in mtl-007-drop-stable, and it actually has exactly the same bug: https://github.com/thesofproject/sof/blob/mtl-007-drop-stable/src/ipc/ipc4/helper.c#L494

ranj063: I dont think mtl-007 has the bug. in that patch, the buf_size is calculated from the ibs/obs set based on the pin ID i.e from the base config for pin ID 0 or from the base config extn if pin ID > 0

andyross: Should be fixed.  Please check.

ranj063 (approved):

jxstelter (approved):

kv2019i:
@marcinszkudlinski ?

src/audio/module_adapter/module_adapter_ipc4.c
```diff
@@ -38,19 +38,39 @@ int module_adapter_init_data(struct comp_dev *dev,
...
+				return -ENOMEM;
+
+			dst->output_pins = (void *)&dst->input_pins[n_in];
+			memcpy_s(dst->input_pins, pinsz,
+				 &cfg->base_cfg_ext.pin_formats[0], pinsz);
+		}
 	}
 
+	dst->init_data = cfg; /* legacy API */
+	dst->avail = true;
```
lgirdwood, line 73: This block could do with an inline comment describing what its doing.

src/audio/module_adapter/module_adapter_ipc4.c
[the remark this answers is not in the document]
lgirdwood: Is this really fatal ? I think its probably survivable where we return the error up the stack for this module/pipeline only and allow other pipelines/streams to continue.

lgirdwood (approved):
My changes can be fixed incrementally, I'd rather get this merged and tested more as priority.

lgirdwood (approved):
My changes can be fixed incrementally, I'd rather get this merged and tested more as priority.

marcinszkudlinski (approved):

marc-hb:
Fix c3da4e3d623a91c for double-free regression was just merged in #8790

PR #8688, opened 2024-01-03
Tools: Topology2: Add DMIC1 and other rates support

singalsu:


singalsu:
Note: This patch needs https://github.com/alsa-project/alsa-utils/pull/250

wszypelt:
@singalsu cmake tools failed in inner intel service

singalsu:
> @singalsu cmake tools failed at internal intel service

The topologies build fails with current alsa-utils version since I wanted to use defines like $DMIC1_RATE for format attributes.

tools/topology/topology2/development/tplg-targets.cmake
```diff
@@ -86,6 +86,51 @@ PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-hda-efx-generic-4ch.bin,\
 DEEPBUFFER_FW_DMA_MS=100,EFX_FIR_PARAMS=passthrough,EFX_IIR_PARAMS=passthrough,\
 EFX_DRC_COMPONENT=multiband,EFX_MBDRC_PARAMS=passthrough"
 
+# With 16 kHz DMIC1
```
RanderWang, line 89: One question:  can we have a flag to build these DMIC1 ?  Currently we do use these DMIC1 tplgs but need to add burden for CI

RanderWang (commented):
LGTM

tools/topology/topology2/development/tplg-targets.cmake
singalsu: @RanderWang @fredoh9 @mengdonglin We had a testing gap and missed that alsatplg did not work correctly for two DMIC DAI case. I'd like to add some cases with DMIC1 and additionally with other than default 16 kHz to increase coverage. Suggestions of which platforms to use and what kind of topologies set to build are welcome. I used here two platforms I had hands-on access to. A MTL RVP platform and LNL might make better sense. What do you think?

tools/topology/topology2/development/tplg-targets.cmake
```diff
@@ -86,6 +86,51 @@ PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-hda-efx-generic-4ch.bin,\
...
+DMIC1_ENABLE=passthrough,DMIC1_RATE=96000,\
+PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-mtl-sdw-cs42l42-l0-max98363-l2-48k-96k.bin,\
+BT_NAME=SSP1-BT,BT_INDEX=1,BT_PCM_ID=20,BT_ID=8,BT_PCM_NAME=Bluetooth,ADD_BT=true,\
+NUM_SDW_AMP_LINKS=1,SDW_SPK_STREAM=SDW2-Playback,SDW_AMP_FEEDBACK=false,\
+SDW_JACK_CAPTURE_CH=1,DEEPBUFFER_FW_DMA_MS=100,DEEPBUFFER_D0I3_COMPATIBLE=true"
+
+# SDW + DMIC + HDMI, with 96 kHz DMIC0
+"cavs-sdw\;sof-mtl-sdw-cs42l42-l0-max98363-l2-96k\;PLATFORM=mtl,\
+NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,DMIC0_ID=3,DMIC1_ID=4,\
+DMIC1_ENABLE=passthrough,DMIC1_RATE=96000,\
```
singalsu, line 128: Oops, here's a mistake, should be with DMIC0_RATE=96000 and without DMIC1_ENABLE.

tools/topology/topology2/development/tplg-targets.cmake
RanderWang: Yes, MTL RVP is preferred. SDW platforms use SDW codec DMIC in current situation, do we need to test with it ?


Please check this in HDA machine driver,  does your PR work with this ?
````c
	{
		.name = "dmic16k",
		.id = 7,
		.dpcm_capture = 1,
		.no_pcm = 1,
		SND_SOC_DAILINK_REG(dmic16k, dmic_codec, platform),
	},
````

lgirdwood (approved):
LGTM, I assume you need more testing ?

singalsu:
I got more comments about DMIC to #8728 that also contains these patches, so better to continue there and close this.

singalsu:
> LGTM, I assume you need more testing ?

This can't be built with current alsatplg version in CI.

PR #8689, opened 2024-01-03
Add initial Google CTC component

eddy1021:
Add CTC component


lgirdwood (commented):
@eddy1021 can you update to us the module API instead of component API (now deprecated) ? Module API works with both IPC3 & 4, the changes are mainly in registration/init with some functions using changed names and args. @singalsu can give you more details if needed. Thanks !

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,369 @@
...
+							  google_ctc_audio_processing_config_size);
+		if (ret) {
+			comp_err(dev, "GoogleCtcAudioProcessingReconfigure failed: %d",
+				 ret);
+			return ret;
+		}
+	}
+
+	if (num_channels_present) {
+		cd->num_channels = num_channels;
```
andyross, line 173: Why is num_channels an external tuneable at all?  Channel count is a runtime property of the SOF stream, shouldn't we be configuring this automatically to eliminate the possibility of getting wrong all the way out in Linux userspace (where it's a static binary blob that can't be programmatically controlled except at image generation time, no less!)?

Maybe it's too late to fix the external library API.  But at the very least I'd suggest validating this vs. the "real" value we get from the SOF input source and using that one if there's a conflict.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,369 @@
...
+	in_streamb = &input_buffers[0];
+	in_stream = in_streamb->data;
+	src = audio_stream_get_rptr(in_stream);
+	out_streamb = &output_buffers[0];
+	out_stream = out_streamb->data;
+	dst = audio_stream_get_rptr(out_stream);
+
+	for (frame = 0; frame < cd->num_frames; frame++) {
+		for (channel = 0; channel < cd->num_channels; channel++) {
+			GoogleCtcAudioProcessingProcess(cd->state, src, dst);
```
andyross, line 230: So... the interface to the CTC library is a 3-argument C function that needs to be called for Every Individual Sample in the stream?  Is that not a performance disaster?  Can't we rework it so it takes a buffer pointer?

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,369 @@
...
+
+			src += num_channels;
+			dst += num_channels;
```
andyross, line 233: Maybe I'm just confused over terminology?  The inner loop here is processing one channel in the frame, but you're incrementing the pointers by a full frame and not just the one sample?  Shouldn't this just be `src++; dst++;`?

Also note that this is ignoring buffer wrap and assuming that the full num_frames*num_channels block of data is contiguous in the buffers starting at the provided pointers.  That's only going to be true in the general case if the buffer size exactly matches the processing size here, otherwise you end up with a buffer overrun.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,369 @@
...
+	comp_dbg(mod->dev, "google_ctc_audio_processing_reset()");
+
+	return 0;
+}
+
+static int google_ctc_audio_processing_process(struct processing_module *mod,
+					       struct input_stream_buffer *input_buffers,
+					       int num_input_buffers,
+					       struct output_stream_buffer *output_buffers,
+					       int num_output_buffers)
```
andyross, line 203: This is implementing the older process_audio_stream API, not the newer process() one which works with the sink/source API.  I got dinged on this with the AEC PR too, in theory either can still be used, but sink/source is futureproof.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,369 @@
...
+
+	in_streamb = &input_buffers[0];
+	in_stream = in_streamb->data;
+	src = audio_stream_get_rptr(in_stream);
+	out_streamb = &output_buffers[0];
+	out_stream = out_streamb->data;
+	dst = audio_stream_get_rptr(out_stream);
+
+	for (frame = 0; frame < cd->num_frames; frame++) {
+		for (channel = 0; channel < cd->num_channels; channel++) {
```
andyross, line 229: Just to repeat from elsewhere that this num_channels value may have no relation to the channel count of the stream you're actually processing.  At the very least we should be validating in a prepare() method.

andyross (commented):
I guess I can declare myself expert enough to start pontificating a bit.  A few spots seem like they might be errors, most are nitpicks.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,369 @@
...
+
+	cd->num_frames = CONFIG_COMP_GOOGLE_CTC_AUDIO_PROCESSING_NUM_FRAMES;
+
+	cd->input =
+		rballoc(0, SOF_MEM_CAPS_RAM, cd->num_frames * sizeof(cd->input[0]));
+	if (!cd->input) {
+		ret = -ENOMEM;
+		goto fail;
+	}
+	bzero(cd->input, cd->num_frames * sizeof(cd->input[0]));
```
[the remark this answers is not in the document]
cujomalainey: At least last time I checked rzalloc can't pull from the buffer zone like rballoc, but I might be wrong.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,369 @@
...
+
+	if (!comp_is_current_data_blob_valid(cd->tuning_handler) &&
+	    !comp_is_new_data_blob_available(cd->tuning_handler)) {
+		/*
+		 * The data blob hasn't been available once so far.
+		 *
+		 * This looks redundant since the same check will be done in
+		 * comp_get_data_blob() below. But without this early return,
+		 * hundreds of warn message lines are produced per second by
+		 * comp_get_data_blob() calls until the data blob is arrived.
```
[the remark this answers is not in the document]
cujomalainey: I thought this was moved up into the module layer?

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,369 @@
...
+			  cd->num_channels);
+	}
+
+	return 0;
+}
+
+static int google_ctc_audio_processing_prepare(struct processing_module *mod)
+{
+	struct comp_dev *dev = mod->dev;
+	struct google_ctc_audio_processing_comp_data *cd = module_get_private_data(mod);
```
cujomalainey, line 185: unused

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,369 @@
...
+	comp_info(dev, "google_ctc_audio_processing_prepare()");
+
+	return 0;
+}
+
+static int google_ctc_audio_processing_reset(struct processing_module *mod)
+{
+	comp_dbg(mod->dev, "google_ctc_audio_processing_reset()");
+
+	return 0;
```
cujomalainey, line 196: there is no way to reset the state of the filters?

third_party/include/google_ctc_audio_processing.h
```diff
@@ -0,0 +1,45 @@
...
+    int num_frames, int partition_size, int impulse_size, int chunk_frames,
+    int sample_rate, int is_symmetric, const uint8_t *const config,
+    int config_size);
+
+// Frees all allocated resources in `state`.
+void GoogleCtcAudioProcessingFree(GoogleCtcAudioProcessingState *state);
+
+// Apply CTC on `src` and produce result in `dest`.
+void GoogleCtcAudioProcessingProcess(GoogleCtcAudioProcessingState *const state,
+				     const float *const src, float *const dest);
```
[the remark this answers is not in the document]
cujomalainey: If this was a copy paste of the header then we should leave it. Otherwise i agree it is a bit weird.

lgirdwood (commented):
@eddy1021 good to see a mock as part of this PR :)
@singalsu pls review.

third_party/include/google_ctc_audio_processing.h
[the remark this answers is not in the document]
eddy1021: The API of the algorithm can still be changed. I'll remove the postfix const in both places.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,369 @@
...
+	}
+	bzero(cd->input, cd->num_frames * sizeof(cd->input[0]));
+
+	cd->output =
+		rballoc(0, SOF_MEM_CAPS_RAM, cd->num_frames * sizeof(cd->output[0]));
+	if (!cd->output) {
+		ret = -ENOMEM;
+		goto fail;
+	}
+	bzero(cd->output, cd->num_frames * sizeof(cd->output[0]));
```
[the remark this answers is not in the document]
eddy1021: Removed bzero

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,369 @@
...
+
+	return 0;
+fail:
+	comp_err(dev, "google_ctc_audio_processing_init(): Failed");
+	if (cd) {
+		rfree(cd->input);
+		rfree(cd->output);
+		if (cd->state)
+			GoogleCtcAudioProcessingFree(cd->state);
+		rfree(cd);
```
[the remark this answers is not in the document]
eddy1021: Thanks! Add a new `ctc_free` function.

src/audio/google/google_ctc_audio_processing.c
eddy1021: Changed to the source/sink API.

src/audio/google/google_ctc_audio_processing.c
eddy1021: Changed to process for the whole chunk. Thanks for pointing this out.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,369 @@
...
+	struct comp_buffer *input;
+	struct comp_buffer *output;
+	uint32_t num_frames;
+	uint32_t num_channels;
+	GoogleCtcAudioProcessingState *state;
+	struct comp_data_blob_handler *tuning_handler;
+	bool reconfigure;
+};
+
+static int google_ctc_audio_processing_init(struct processing_module *mod)
```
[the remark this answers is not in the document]
eddy1021: Shorten it to ctc_*, also for other APIs.

src/audio/google/google_ctc_audio_processing.c
eddy1021: Yes, `num_channels` should not be an tuneable value. I'll remove it. Excuse me for the confusion.

src/audio/google/google_ctc_audio_processing.c
eddy1021: Removed.

src/audio/google/google_ctc_audio_processing.c
eddy1021: Removed `num_channels` in the comp_data now. Excuse me for the confusion.

src/audio/google/google_ctc_audio_processing.c
eddy1021: Changed the CTC library to take buffer pointers instead.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,360 @@
...
+	if (cd) {
+		rfree(cd->input);
+		rfree(cd->output);
+		if (cd->state) {
+			GoogleCtcAudioProcessingFree(cd->state);
+			cd->state = NULL;
+		}
+		rfree(cd);
+	}
+	rfree(dev);
```
singalsu, line 46: The processing modules do not need to free mod->dev. It's done in module_adapter_free() in module_adapter.c.

singalsu: @eddy1021 Also please check the codestyle / checkpatch results from CI (long lines, white space).

eddy1021: Thanks! I've fixed the long lines and white space issues.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,359 @@
...
+	}
+
+	in_stream = sources[0];
+	out_stream = sinks[0];
+
+	num_of_bytes_to_process = cd->num_frames * source_get_frame_bytes(in_stream);
+
+	ret = source_get_data(in_stream, num_of_bytes_to_process, (const void **)&in,
+			      (const void **)&in_buf_start,  &in_buf_size);
+	assert(!ret);
```
lyakh, line 216: I wouldn't crash the system in such cases, just return the error code. Same below

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,359 @@
...
+
+	source_release_data(in_stream, num_of_bytes_to_process);
+	sink_commit_buffer(out_stream, num_of_bytes_to_process);
+
+	return 0;
+}
+
+#if CONFIG_IPC_MAJOR_3
+static int ctc_cmd_set_data(struct processing_module *mod,
+						    struct sof_ipc_ctrl_data *cdata)
```
lyakh, line 232: indentation - multiple cases

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,359 @@
...
+		if (ret)
+			return ret;
+		/* Accept the new blob immediately so that userspace can write
+		 * the control in quick succession without error.
+		 * This ensures the last successful control write from userspace
+		 * before prepare/copy is applied.
+		 * The config blob is not referenced after reconfigure() returns
+		 * so it is safe to call comp_get_data_blob here which frees the
+		 * old blob. This assumes cmd() and prepare()/copy() cannot run
+		 * concurrently which is the case when there is no preemption.
```
lyakh, line 249: concurrently - on different cores or preempting each other? I'd say, that the only assumption that you can make now is that IPCs are serialised. I.e. a new IPC command cannot be sent before the previous one has been answered. And an IPC is answered only after its processing has been completed by the DSP. The only processes that can run concurrently or preempt each other are the streaming / copying taking place in the background and new IPCs (stop, pause, configure, etc.). Then you have to consider which event can preempt which context...

src/audio/google/google_ctc_audio_processing_mock.c
```diff
@@ -0,0 +1,53 @@
...
+	s->impulse_size = 256;
+	s->chunk_frames = 480;
+	s->sample_rate = 48000;
+	s->is_symmetric = 0;
+	return s;
+}
+
+void GoogleCtcAudioProcessingFree(GoogleCtcAudioProcessingState *state)
+{
+	if (state)
```
lyakh, line 42: unneeded, `rfree(NULL)` is fine

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,359 @@
...
+	return 0;
+}
+
+static int ctc_process(struct processing_module *mod,
+					       struct sof_source **sources, int num_of_sources,
+					       struct sof_sink **sinks, int num_of_sinks)
+{
+	struct google_ctc_audio_processing_comp_data *cd = module_get_private_data(mod);
+
+	int16_t const *in;
```
johnylin76, line 186: It looks to have the assumption that the sof source stream is 16-bit format. Does that make sense?

src/audio/google/google_ctc_audio_processing.c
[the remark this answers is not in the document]
eddy1021: Excuse me, I think CTC does not fall into this special case actually.

src/audio/google/google_ctc_audio_processing.c
eddy1021: I changed the process function to avoid that. Thanks!

src/audio/google/google_ctc_audio_processing_mock.c
eddy1021: Removed.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,359 @@
...
+	comp_info(dev, "ctc_init()");
+
+	/* Create private component data */
+	cd = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*cd));
+	if (!cd) {
+		ret = -ENOMEM;
+		goto fail;
+	}
+
+	md->private = cd;
```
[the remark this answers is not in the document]
eddy1021: Changed

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,359 @@
...
+	struct google_ctc_audio_processing_comp_data *cd = module_get_private_data(mod);
+
+	if (cd) {
+		rfree(cd->input);
+		rfree(cd->output);
+		if (cd->state) {
+			GoogleCtcAudioProcessingFree(cd->state);
+			cd->state = NULL;
+		}
+		rfree(cd);
```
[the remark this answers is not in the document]
eddy1021: changed

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,359 @@
...
+	}
+
+	comp_info(dev, "google_ctc_audio_processing_reconfigure(): New tuning config %p (%zu bytes)",
+		  config, size);
+
+	cd->reconfigure = false;
+
+	uint8_t *google_ctc_audio_processing_config;
+	size_t google_ctc_audio_processing_config_size;
+	bool google_ctc_audio_processing_config_present;
```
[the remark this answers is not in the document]
eddy1021: Thanks, I make it shorter for reading friendly.

marc-hb:
@eddy1021 , minor Github trick:

> Respond to comments using the “Start Review” and “Add Review” green buttons in the “Files changed” view. This allows responding to multiple comments and publishing the responses in bulk. This reduces the number of emails sent to reviewers.

kv2019i (commented):
This PR seems to be close to be merge-shape. I see nothing blocking in code (one TODO, added an inline comment about that). The testbench build failures are compile warnings that are probably easy to fix:
https://github.com/thesofproject/sof/actions/runs/8456582880/job/23166679582?pr=8689
@singalsu and @marc-hb can help with these if needed.
Otherwise, I'll wait someone from Google to +1 before merge.

singalsu:
> The testbench build failures are compile warnings that are probably easy to fix:

Yes, there's an unused variable.

```
/home/runner/work/sof/sof/src/audio/google/google_ctc_audio_processing.c: In function ‘ctc_init’:
/home/runner/work/sof/sof/src/audio/google/google_ctc_audio_processing.c:265:29: error: unused variable ‘md’ [-Werror=unused-variable]
  265 |         struct module_data *md = &mod->priv;
      |                             ^~

```

lgirdwood:
@eddy1021 any update - think we are almost good to go !

cujomalainey:
> > The testbench build failures are compile warnings that are probably easy to fix:
> 
> Yes, there's an unused variable.
> 
> ```
> /home/runner/work/sof/sof/src/audio/google/google_ctc_audio_processing.c: In function ‘ctc_init’:
> /home/runner/work/sof/sof/src/audio/google/google_ctc_audio_processing.c:265:29: error: unused variable ‘md’ [-Werror=unused-variable]
>   265 |         struct module_data *md = &mod->priv;
>       |                             ^~
> cc1: all warnings being treated as errors
> ```

@eddy1021 FYI

singalsu:
> > This PR seems to be close to be merge-shape. I see nothing blocking in code (one TODO, added an inline comment about that). The testbench build failures are compile warnings that are probably easy to fix: https://github.com/thesofproject/sof/actions/runs/8456582880/job/23166679582?pr=8689 @singalsu and @marc-hb can help with these if needed. Otherwise, I'll wait someone from Google to +1 before merge.
> 
> Yes we are still blocked internally on the backend's full end to end test. once we have that we will push this. Thanks for your patience

The CI testbench build has error with function non-static "void GoogleCtcAudioProcessingParseSofConfigMessage". A function prototype add is a simple fix for it, or perhaps it should be static?

cujomalainey:
> > > This PR seems to be close to be merge-shape. I see nothing blocking in code (one TODO, added an inline comment about that). The testbench build failures are compile warnings that are probably easy to fix: https://github.com/thesofproject/sof/actions/runs/8456582880/job/23166679582?pr=8689 @singalsu and @marc-hb can help with these if needed. Otherwise, I'll wait someone from Google to +1 before merge.
> > 
> > 
> > Yes we are still blocked internally on the backend's full end to end test. once we have that we will push this. Thanks for your patience
> 
> The CI testbench build has error with function non-static "void GoogleCtcAudioProcessingParseSofConfigMessage". A function prototype add is a simple fix for it, or perhaps it should be static?

no that means that likely likely is missing the header before a usage

eddy1021:
The CI testbench build error has been fixed.

The internal building issue is fixed now and the feature is tested end-to-end. The review for this PR should be un-blocked now. Thanks for the long wait.

src/audio/google/google_ctc_audio_processing.toml
```diff
@@ -0,0 +1,17 @@
...
+	affinity_mask = "0x1"
+	instance_count = "40"
+	domain_types = "0"
+	load_type = "0"
+	module_type = "9"
+	auto_start = "0"
+	sched_caps = [1, 0x00008000]
+	// # pin = [dir, type, sample rate, size, container, channel-cfg]
+	pin = [0, 0, 0xfeef, 0xf, 0xf, 0x45ff, 1, 0, 0xfeef, 0xf, 0xf, 0x1ff]
+	// # mod_cfg [PAR_0 PAR_1 PAR_2 PAR_3 IS_BYTES CPS IBS OBS MOD_FLAGS CPC OBLS]
```
marc-hb, line 14: This drops comments from the pre-processed file. You can use the new `REM #` prefix instead to avoid that.

app/overlays/mtl/ctc_overlay.conf
```diff
@@ -0,0 +1,4 @@
+CONFIG_COMP_STUBS=n
+CONFIG_COMP_MODULE_ADAPTER=y
+CONFIG_COMP_GOOGLE_CTC_AUDIO_PROCESSING=y
+CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING=n
```
marc-hb, line 4: Are RTC and CTC mutually exclusive? If they are then Kconfig files should make sure of that always - not just when using this particular .conf file.

If they are not mutually exclusive then why is RTC being turned off here?

Kconfig and .conf issues are always a nightmare to debug and have a CTC.conf file sneakily and silently turning off RTC is a recipe for confusion.

At the very least this line deserves a comment.

marc-hb: Actually: is anyone or anything outside Google ever going to test CTC and use this ctc.conf file? If not then please keep it private; it's only 3-4 lines long.

zephyr/CMakeLists.txt
```diff
@@ -785,6 +785,25 @@ if(CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING AND NOT CONFIG_GOOGLE_RTC_AUDIO_PROCE
...
+if(CONFIG_COMP_GOOGLE_CTC_AUDIO_PROCESSING AND NOT CONFIG_GOOGLE_CTC_AUDIO_PROCESSING_MOCK)
+	zephyr_include_directories(../third_party/include)
+	target_link_directories(SOF INTERFACE ../third_party/lib)
+	target_link_libraries(SOF INTERFACE google_ctc_audio_processing)
+	target_link_libraries(SOF INTERFACE c++)
+	target_link_libraries(SOF INTERFACE c++abi)
+	target_link_libraries(SOF INTERFACE m)
+	target_link_libraries(SOF INTERFACE c)
+	target_link_libraries(SOF INTERFACE gcc)
+endif()
```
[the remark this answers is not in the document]
cujomalainey: @marc-hb I believe that is the intention since one is for XTOS and the other for Zephyr, unless this has been resolved then we need both

app/overlays/mtl/ctc_overlay.conf
cujomalainey: No they are not mutually exclusive

marc-hb: > No they are not mutually exclusive

So the `CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING=n` here? It's sneaky.

zephyr/CMakeLists.txt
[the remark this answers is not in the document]
marc-hb: You can have both Zephyr and XTOS in `src/audio/google/CMakeLists.txt` now, see example in `src/ipc/CMakeLists.txt`.

For more examples: `git grep is_zephyr`

There is still a lot of de-duplication work to do but please try not to increase that work.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,574 @@
...
+		cd->input[cd->input_ptr++] = convert_int16_to_float(src[i]);
+	src = audio_stream_wrap(source, src + samples_to_process);
+	input_buffers->consumed += audio_stream_frame_bytes(source) * samples_to_process / n_ch;
+
+	while (cd->input_ptr >= cd->num_frames * n_ch) {
+		GoogleCtcAudioProcessingProcess(cd->state, cd->input, cd->output + cd->output_ptr,
+						cd->num_frames, n_ch);
+		cd->output_ptr += cd->num_frames * n_ch;
+
+		cd->input_ptr -= cd->num_frames * n_ch;
```
lyakh, line 147: `.input_ptr` isn't really a pointer from how it's used, it's rather a counter

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,574 @@
...
+		for (int i = 0; i < cd->input_ptr; ++i)
+			cd->input[i] = cd->input[i + cd->num_frames * n_ch];
```
lyakh, line 149: this seems rather inefficient - moving samples to the beginning of both input and output buffers (below). Maybe you could make this a circular buffer?

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,574 @@
...
+	int ret;
+
+	comp_info(mod->dev, "ctc_prepare()");
+
+	source = list_first_item(&dev->bsource_list, struct comp_buffer, sink_list);
+	cd->ctc_func = ctc_find_func(audio_stream_get_frm_fmt(&source->stream));
+	if (!cd->ctc_func) {
+		comp_err(mod->dev, "ctc_prepare(): No suitable processing function found.");
+		ret = -EINVAL;
+		goto err;
```
lyakh, line 426: `return -EINVAL;`

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,574 @@
...
+static inline int16_t convert_float_to_int16(float data)
+{
+#if XCHAL_HAVE_HIFI3
+	const xtfloat ratio = 2 << 15;
+	xtfloat x0 = data;
+	xtfloat x1;
+	int16_t x;
+
+	x1 = XT_MUL_S(x0, ratio);
+	x = XT_TRUNC_S(x1, 0);
```
andyross, line 66: FWIW, the use of intrinsics for scalar float usage is minimally useful.  The compiler generates great (generally identical) code if you just use C types, and has the advantage of working correctly on HIFI4/5 devices without all the hardware muckery.  The code in the AEC patch is solving basically identical problems and generating more or less optimal output code with only floats.

Also the !HIFI case (which corresponds to "open source toolchain" in practice) is sort of pessimal.  Q_CONVERT_FLOAT() is going to go through a double precision conversion, where if you just let gcc see the floats directly it'll call into its own libgcc soft-float routines at single precision.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,574 @@
...
+	int32_t *src = audio_stream_get_rptr(source);
+	int32_t *dest = audio_stream_get_wptr(sink);
+
+	int samples_to_process = MIN(samples, audio_stream_samples_without_wrap_s32(source, src));
+	int samples_to_consume = 0;
+
+	for (int i = 0; i < samples_to_process; ++i)
+		cd->input[cd->input_ptr++] = convert_int32_to_float(src[i]);
+	src = audio_stream_wrap(source, src + samples_to_process);
+	input_buffers->consumed += audio_stream_frame_bytes(source) * samples_to_process / n_ch;
```
andyross, line 224: It suddenly strikes me that this all almost 100% identical to the same code in AEC, though this has 24 bit sample support.  Post-merge we should do a unification pass.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,574 @@
...
+ * \param fmt the frames' format of the source and sink buffers
+ */
+static ctc_func ctc_find_func(enum sof_ipc_frame fmt)
+{
+	int i;
+
+	/* Find suitable processing function from map */
+	for (i = 0; i < ctc_fncount; i++) {
+		if (fmt == ctc_fnmap[i].fmt)
+			return ctc_fnmap[i].func;
```
andyross, line 274: Nitpick, but I can all but guarantee that the code size required for the table and this function is bigger than the if() expression needed to just select it directly. :)    IMHO there's a time and place for gold-plated engineering and there are times where clarity demands the simple choice.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,574 @@
...
+
+static int ctc_free(struct processing_module *mod)
+{
+	struct google_ctc_audio_processing_comp_data *cd = module_get_private_data(mod);
+
+	if (cd) {
+		rfree(cd->input);
+		rfree(cd->output);
+		if (cd->state) {
+			GoogleCtcAudioProcessingFree(cd->state);
```
andyross, line 288: Freeing a NULL pointer (which is all GoogleCtcAudioProcessingFree() is a noop, so you can eliminate the if() expression. Also: why set "state" to NULL but not input or output, which are now dangling pointers? 

Normally a "free()" call is a destructor and you assume that no one will touch the struct in the future.  But here you seem to assume that it needs to be re-initialized, which makes me worry about the other fields too.  Basically best to pick one convention and stick to it.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,574 @@
...
+		ret = google_ctc_audio_processing_reconfigure(mod);
+		if (ret)
+			return ret;
+	}
+
+	cd->ctc_func(cd, source, sink, &input_buffers[0], &output_buffers[0], frames);
+	return 0;
+}
+
+#if CONFIG_IPC_MAJOR_3
```
andyross, line 466: Recent SOF idiom for this kind of thing, where you have large APIs that differ between IPC3/4, is to put the divergent handler code in separate files with a unified API to avoid the big block ifdefs.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,574 @@
...
+
+	comp_info(mod->dev, "google_ctc_audio_processing_ctrl_get_data(): %u", cdata->cmd);
+
+	switch (cdata->cmd) {
+	case SOF_CTRL_CMD_BINARY:
+		return comp_data_blob_get_cmd(cd->tuning_handler, cdata, max_data_size);
+	default:
+		comp_err(mod->dev,
+			 "google_ctc_audio_processing_ctrl_get_data(): Only binary controls supported %d",
+			 cdata->cmd);
```
andyross, line 505: FWIW: that -EINVAL is propagated all the way out to the userspace process that sets the ALSA control, so it won't be missed.  Not clear to me that it's worth space in the firmware binary just to log "unsupported".  IMHO "ERROR" logs in firmware should be limited to genuine runtime errors, not routine checking of correct software usage.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,574 @@
...
+#endif
+}
+
+static struct module_interface google_ctc_audio_processing_interface = {
+	.init  = ctc_init,
+	.free = ctc_free,
+	.process_audio_stream = ctc_process,
+	.prepare = ctc_prepare,
+	.set_configuration = ctc_set_config,
+	.get_configuration = ctc_get_config,
```
andyross, line 567: Just to repeat from elsewhere: the pattern would be you have a xxx-ipc3.c and xxx-ipc4.c, each of which implements global "google_ctc_s/get_config()" functions which you then put here.  Then you select which one to build in cmake.

src/audio/google/google_ctc_audio_processing.toml
```diff
@@ -0,0 +1,17 @@
...
+	instance_count = "40"
+	domain_types = "0"
+	load_type = "0"
+	module_type = "9"
+	auto_start = "0"
+	sched_caps = [1, 0x00008000]
+	REM # pin = [dir, type, sample rate, size, container, channel-cfg]
+	pin = [0, 0, 0xfeef, 0xf, 0xf, 0x45ff, 1, 0, 0xfeef, 0xf, 0xf, 0x1ff]
+	REM # mod_cfg [PAR_0 PAR_1 PAR_2 PAR_3 IS_BYTES CPS IBS OBS MOD_FLAGS CPC OBLS]
+	mod_cfg = [0, 0, 0, 0, 4096, 1000000, 128, 128, 0, 0, 0]
```
andyross, line 15: Not really related to this PR, but I'm seeing this pattern abused in the TOML stuff and it bugs me: the TOML entries in the manifest are stored per-type, not per-component.  They shouldn't have stuff that is already covered by the per-component topology data (and especially shouldn't have stuff that duplicates and might conflict with topology).

But people are doing it anyway, because topology doesn't exist in a general form down here in the firmware, where the TOML-generated manifest entries do.  It's a design wart.

src/audio/google/google_ctc_audio_processing_mock.c
```diff
@@ -0,0 +1,82 @@
...
+{
+	struct GoogleCtcAudioProcessingState *s =
+		rballoc(0, SOF_MEM_CAPS_RAM, sizeof(GoogleCtcAudioProcessingState));
+	if (!s)
+		return NULL;
+
+	s->num_frames = CONFIG_COMP_GOOGLE_CTC_AUDIO_PROCESSING_NUM_FRAMES;
+	s->partition_size = 64;
+	s->impulse_size = 256;
+	s->sample_rate = 48000;
```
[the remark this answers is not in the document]
andyross: Actually... what is this for?  This is dead code.  That "State" struct is local to this file and not visible outside it except as an opaque pointer.   And you initialize it here, but.... nothing reads it anywhere. :)

Just remove.  It's reasonable to return a zero-lenght heap block for tracking purposes just to be sure it's unique and acts like the non-mock component.  But the dead code seems silly.

src/audio/google/google_ctc_audio_processing_mock.c
```diff
@@ -0,0 +1,82 @@
...
+
+void GoogleCtcAudioProcessingFree(GoogleCtcAudioProcessingState *state)
+{
+	rfree(state);
+}
+
+void GoogleCtcAudioProcessingProcess(GoogleCtcAudioProcessingState *state,
+				     const float *src, float *dest,
+				     int num_frames, int num_channels)
+{
```
andyross, line 64: To repeat from elsewhere: note that the "state" parameter is ignored, so there's no point in putting any info in it.

andyross (commented):
Haven't come through on this one in a while.  Some nitpickery, nothing fatal.

app/overlays/mtl/ctc_overlay.conf
marc-hb: BTW why is this new file placed in the `mtl/` subdirectory? It does not look like it has anything MTL-specific from a technical perspective.

Maybe Google wants this only on MTL right now from a "project management" perspective? In this case, `ctc_overlay.conf` should live in some Google-specific subdirectory or even better - in private configuration files.

src/audio/google/Kconfig
```diff
@@ -93,4 +93,61 @@ config GOOGLE_RTC_AUDIO_PROCESSING_MOCK
...
+	select COMP_BLOB
+	select GOOGLE_CTC_AUDIO_PROCESSING_MOCK if COMP_STUBS
+	default n
+	help
+	  Select for Google crosstalk cancellation audio processing. It
+	  uses the Google real-time audio processing library to perform
+	  crosstalk cancellation.
+
+config COMP_GOOGLE_CTC_AUDIO_PROCESSING_NUM_FRAMES
+	depends on COMP_GOOGLE_CTC_AUDIO_PROCESSING
```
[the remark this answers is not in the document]
eddy1021: Done. Thanks!

src/audio/google/google_ctc_audio_processing.c
eddy1021: Changed to the same version as AEC.

src/audio/google/google_ctc_audio_processing.c
eddy1021: Renamed to `input_samples`.

src/audio/google/google_ctc_audio_processing.c
eddy1021: Redesign the buffer to contain exactly `num_frames * n_ch` and consume them as soon as there are enough samples to pass to the CTC function.
In this way, the samples don't need to be moved.
Thanks for pointing this out!

src/audio/google/google_ctc_audio_processing.c
eddy1021: Changed to inline ifs.

src/audio/google/google_ctc_audio_processing.c
eddy1021: Fixed.

src/audio/google/google_ctc_audio_processing.c
eddy1021: Fixed

src/audio/google/google_ctc_audio_processing.c
eddy1021: I'll leave it as a follow up change.

src/audio/google/google_ctc_audio_processing.c
eddy1021: Split to google_ctc_audio_processing_ipc{3, 4}.c

src/audio/google/google_ctc_audio_processing.c
eddy1021: I agree. Thanks!

src/audio/google/google_ctc_audio_processing.c
eddy1021: Done. Thanks!

src/audio/google/google_ctc_audio_processing_mock.c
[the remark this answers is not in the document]
eddy1021: removed. Thanks!

src/audio/google/google_ctc_audio_processing_mock.c
eddy1021: Removed.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,357 @@
...
+}
+
+static int ctc_free(struct processing_module *mod)
+{
+	struct google_ctc_audio_processing_comp_data *cd = module_get_private_data(mod);
+
+	if (cd) {
+		rfree(cd->input);
+		rfree(cd->output);
+		GoogleCtcAudioProcessingFree(cd->state);
```
lyakh, line 167: isn't this missing `rfree(cd)` here?

lyakh (changes requested):
simple to fix, but better fix it before merge

src/audio/google/google_ctc_audio_processing.c
cujomalainey: also the new handler should be designed such that if we don't have cd then init failed and free shouldn't need to be called.

eddy1021: Added `rfree(cd)`.

lyakh:
@eddy1021 BTW, this PR adds a mock, but all Kconfig options are off by default, so this will never be tested by the CI, not even for compilation, right?

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,358 @@
...
+#include <google_ctc_audio_processing_sof_message_reader.h>
+
+#include "google_ctc_audio_processing.h"
+
+LOG_MODULE_REGISTER(google_ctc_audio_processing, CONFIG_SOF_LOG_LEVEL);
+
+/* bf0e1bbc-dc6a-45fe-bc90-2554cb137ab4 */
+DECLARE_SOF_RT_UUID("google-ctc-audio-processing", google_ctc_audio_processing_uuid,
+		    0xbf0e1bbc, 0xdc6a, 0x45fe, 0xbc, 0x90, 0x25, 0x54, 0xcb,
+		    0x13, 0x7a, 0xb4);
```
kv2019i, line 27: This could now be moved to the new UUID system @andyross added - see commit 17f46a71ee22505949af5d2de8def513b16b7021 and 51a5fb9c8098207527a4ef8b5afbc78a6f3b9bda

kv2019i (approved):
I think this is starting to be ready. Having a version in the tree could help with maintainance and get changes like the recent UUID registry and cleanup of the xtos/zephyr makefiles, "for free". These need to be addressed, but I'm ok to do these as follow-ups.

marc-hb:
> this PR adds a mock, but all Kconfig options are off by default, so this will never be tested by the CI, not even for compilation, right?

I think you mean "open-source CI". I think (and hope) there are many closed CIs we don't all know about.

lyakh:
> > this PR adds a mock, but all Kconfig options are off by default, so this will never be tested by the CI, not even for compilation, right?
> 
> I think you mean "open-source CI". I think (and hope) there are many closed CIs we don't all know about.

@marc-hb I mean the SOF GitHub PR CI - the one we all see results of when submitting our PRs.

eddy1021:
> @eddy1021 BTW, this PR adds a mock, but all Kconfig options are off by default, so this will never be tested by the CI, not even for compilation, right?

I turned the Mock to default on now. Thanks for pointing this out!

lgirdwood (approved):

lgirdwood:
@eddy1021 CI has some scheduled downtime, I'll manually rerun CI on Monday.

marc-hb:
> @eddy1021 CI has some scheduled downtime, I'll manually rerun CI on Monday.

There was a downtime but that is not the main reason Jenkins did not test this. The main reason is because I had to add the magic keywords `[SKIP SOF-TEST]` in the past because of rapid-fire force pushes that were saturating our scarce hardware devices. It seems more quiet now; dropping it.

marc-hb:
SOFCI TEST

marc-hb:
SOFCI TEST

andyross (approved):
Sneak in and refresh my vote

src/audio/google/google_ctc_audio_processing_mock.c
```diff
@@ -0,0 +1,74 @@
...
+		 src, sizeof(float) * num_frames * num_channels);
+}
+
+void GoogleCtcAudioProcessingParseSofConfigMessage(uint8_t *message,
+						   size_t message_size,
+						   uint8_t **config,
+						   size_t *config_size,
+						   bool *config_present)
+{
+	*config = NULL;
```
cujomalainey, line 65: leak?

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,405 @@
...
+#include "google_ctc_audio_processing.h"
+
+LOG_MODULE_REGISTER(google_ctc_audio_processing, CONFIG_SOF_LOG_LEVEL);
+
+SOF_DEFINE_REG_UUID(google_ctc_audio_processing);
+
+DECLARE_TR_CTX(google_ctc_audio_processing_tr, SOF_UUID(google_ctc_audio_processing_uuid),
+	       LOG_LEVEL_INFO);
+
+static inline float clamp_rescale(float max_val, float x)
```
cujomalainey, line 29: we should consolidate these utils with RTC in a followup PR.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,405 @@
...
+	}
+	if (written_samples > 0) {
+		dest = audio_stream_wrap(sink, dest + written_samples);
+		output_buffers->size += audio_stream_frame_bytes(sink) * written_samples / n_ch;
+	}
+	src = audio_stream_wrap(source, src + samples_to_process);
+	input_buffers->consumed += audio_stream_frame_bytes(source) * samples_to_process / n_ch;
+}
+
+static void ctc_s24_default(struct google_ctc_audio_processing_comp_data *cd,
```
cujomalainey, line 109: these should be wrapped with `CONFIG_FORMAT_SXXLE` ifdef

src/audio/google/google_ctc_audio_processing.c
cujomalainey: +1

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,405 @@
...
+
+static int ctc_free(struct processing_module *mod)
+{
+	struct google_ctc_audio_processing_comp_data *cd = module_get_private_data(mod);
+
+	if (cd) {
+		rfree(cd->input);
+		rfree(cd->output);
+		GoogleCtcAudioProcessingFree(cd->state);
+		rfree(cd);
```
cujomalainey, line 213: nit: you can simply code below so you can drop this if

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,405 @@
...
+	struct google_ctc_audio_processing_comp_data *cd;
+	int ret;
+	int buf_size;
+
+	comp_info(dev, "ctc_init()");
+
+	/* Create private component data */
+	cd = rzalloc(SOF_MEM_ZONE_RUNTIME, 0, SOF_MEM_CAPS_RAM, sizeof(*cd));
+	if (!cd) {
+		ret = -ENOMEM;
```
cujomalainey, line 232: just return here to simply code

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,405 @@
...
+		goto fail;
+	}
+	cd->output = rballoc(0, SOF_MEM_CAPS_RAM, buf_size);
+	if (!cd->output) {
+		ret = -ENOMEM;
+		goto fail;
+	}
+
+	cd->input_samples = 0;
```
cujomalainey, line 254: unneeded, `rzalloc` will init to 0.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,405 @@
...
+	struct google_ctc_audio_processing_comp_data *cd = module_get_private_data(mod);
+	struct comp_dev *dev = mod->dev;
+	uint8_t *config;
+	size_t size;
+	int ret;
+
+	comp_dbg(dev, "google_ctc_audio_processing_reconfigure()");
+
+	config = comp_get_data_blob(cd->tuning_handler, &size, NULL);
+	if (size == 0) {
```
cujomalainey, line 290: are more checks done inside the lib?

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,405 @@
...
+	struct google_ctc_audio_processing_comp_data *cd = module_get_private_data(mod);
+	struct audio_stream *source = input_buffers[0].data;
+	struct audio_stream *sink = output_buffers[0].data;
+	uint32_t frames = input_buffers[0].size;
+
+	int ret;
+
+	comp_dbg(mod->dev, "ctc_process()");
+
+	if (cd->reconfigure) {
```
cujomalainey, line 382: recommend you use `config` or `reconfigure` as you have 2 flags to indicate you need to reconfigure when you could use 1

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,405 @@
...
+	struct google_ctc_audio_processing_comp_data *cd = module_get_private_data(mod);
+	struct comp_dev *dev = mod->dev;
+	struct comp_buffer *source;
+	enum sof_ipc_frame fmt;
+
+	comp_info(mod->dev, "ctc_prepare()");
+
+	source = list_first_item(&dev->bsource_list, struct comp_buffer, sink_list);
+	fmt = audio_stream_get_frm_fmt(&source->stream);
+	if (fmt == SOF_IPC_FRAME_S16_LE)
```
cujomalainey, line 345: use a switch case here, also your catch all should fail in case new formats are added

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,405 @@
...
+	}
+
+	return 0;
+}
+
+static int ctc_reset(struct processing_module *mod)
+{
+	comp_dbg(mod->dev, "ctc_reset()");
+
+	// TODO(eddyhsu): handle the reset.
```
cujomalainey, line 363: you should call into the lib to purge the registers

zephyr/CMakeLists.txt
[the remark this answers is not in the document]
cujomalainey: @eddy1021 the check we talked about is `is_zephyr` and please try to merge this.

cujomalainey (changes requested):

src/audio/google/Kconfig
```diff
@@ -88,4 +88,68 @@ config GOOGLE_RTC_AUDIO_PROCESSING_MOCK
...
+	  cancellation audio processing.
+
+config COMP_GOOGLE_CTC_AUDIO_PROCESSING_CHUNK_FRAMES
+	int "Chunk frames to process for Google Crosstalk Cancellation Audio processing"
+	default 48
+	help
+	  Sets the chunk frames to process in the Google crosstalk
+	  cancellation audio processing.
+
+config COMP_GOOGLE_CTC_AUDIO_PROCESSING_SAMPLE_RATE
```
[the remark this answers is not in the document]
eddy1021: Removed. Thanks!

src/audio/google/google_ctc_audio_processing_mock.c
eddy1021: I think it's irrelevant in a mock function.

zephyr/CMakeLists.txt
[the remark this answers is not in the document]
cujomalainey: @marc-hb when was this enabled? I am only a week behind and src/audio cmake files are not touched by the zephyr build for me

lgirdwood:
@cujomalainey good for you now ?

zephyr/CMakeLists.txt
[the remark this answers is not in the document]
marc-hb: My bad, it's not used in `audio/` yet. It could be added to `audio/google/` but nevermind.

src/audio/google/google_ctc_audio_processing.c
```diff
@@ -0,0 +1,422 @@
...
+#endif
+	default:
+		comp_err(mod->dev, "ctc_prepare(), invalid frame_fmt");
+		return -EINVAL;
+	}
+
+	num_channels = audio_stream_get_channels(&source->stream);
+	buf_size = cd->chunk_frames * sizeof(cd->input[0]) * num_channels;
+
+	cd->input = rballoc(0, SOF_MEM_CAPS_RAM, buf_size);
```
cujomalainey, line 350: address leak here as discussed

cujomalainey (changes requested):

src/audio/google/google_ctc_audio_processing.c
eddy1021: Fixed it. Move the buffer allocation to `ctc_init`

cujomalainey (approved):

lgirdwood:
> I'll push submit tomorrow unless anyone else fines any issues.

I've rerun failing GH actions (now passed) and CI is showing expected results (this module not tested in CI to date). Pls merge when ready.

PR #8690, opened 2024-01-03
C++ & AEC support for Zephyr builds

andyross:
Split out from #8571 as it should be relatively uncontroversial.

See commit messages, basically this just forward ports some pre-existing trickery to make Google AEC (which is a C++ library) build with existing toolchains.  As the commit message indicates, really this is something the Zephyr toolchain layer should be prepared to provide, but it doesn't (yet).

### DIFF
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -751,6 +751,17 @@ zephyr_library_sources_ifdef(CONFIG_GOOGLE_RTC_AUDIO_PROCESSING_MOCK
 	${SOF_AUDIO_PATH}/google/google_rtc_audio_processing_mock.c
 )
 
+if(CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING AND NOT CONFIG_GOOGLE_RTC_AUDIO_PROCESSING_MOCK)
+	zephyr_include_directories(../third_party/include)
+	target_link_directories(SOF INTERFACE ../third_party/lib)
+	target_link_libraries(SOF INTERFACE google_rtc_audio_processing)
+	target_link_libraries(SOF INTERFACE c++)
+	target_link_libraries(SOF INTERFACE c++abi)
+	target_link_libraries(SOF INTERFACE m)
+	target_link_libraries(SOF INTERFACE c)
+	target_link_libraries(SOF INTERFACE gcc)
+endif()
+
 zephyr_library_sources_ifdef(CONFIG_COMP_IGO_NR
 	${SOF_AUDIO_PATH}/igo_nr/igo_nr.c
 )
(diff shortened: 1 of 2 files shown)


zephyr/CMakeLists.txt
```diff
@@ -877,6 +877,17 @@ zephyr_library_sources_ifdef(CONFIG_SOF_BOOT_TEST
        boot_test.c
 )
 
+if(CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING AND NOT CONFIG_GOOGLE_RTC_AUDIO_PROCESSING_MOCK)
```
marc-hb, line 880: Can this CMake addition be moved to the relevant subdirectory?
- #8630

andyross: So, I did try.  But unfortunately no, as src/audio/google/CMakeList.txt is written to be included from src/audio/CMakeLists.txt, and that file isn't Zephyrized yet, only src/{init,ipc,module} so far.  And as much as I agree with the aim here, I just don't have the cmake patience required to make that work.

marc-hb: Thanks for trying! 

> no, as src/audio/google/CMakeList.txt is written to be included from src/audio/CMakeLists.txt,

In that case can you please at least move this addition close to the existing `CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING / ${SOF_AUDIO_PATH}/google/google_rtc_audio_processing.c`?

It will be clearer and also  help when #8630 starts looking at SOF_AUDIO_PATH.

marc-hb (changes requested):

marcinszkudlinski (approved):
Looks good. 
"Zephyrsation" of all cmake files is on todo list, single zephyr/CMakeLists.txt is not a good approach.

kv2019i:
@marcinszkudlinski wrote:
> Looks good. "Zephyrsation" of all cmake files is on todo list, single zephyr/CMakeLists.txt is not a good approach.

Not just on todo list but tagged for v2.9 and under work -> https://github.com/thesofproject/sof/issues/8260

kv2019i (approved):
Please check @marc-hb comment still, but otherwise looks good.

lgirdwood (commented):
Ack, LGTM, just the one improvement as mentioned by Marc

andyross:
Sorry, forgot to push this yesterday.  Should be OK now.

marc-hb:
Bad luck: it looks like CI was very busy when https://sof-ci.01.org/sofpr/PR8690/build1594/devicetest/index.html and https://sof-ci.01.org/sofpr/PR8690/build1595/devicetest/index.html ran

marc-hb:
SOFCI TEST

PR #8692, opened 2024-01-04
Audio: ARIA: Add HiFi5 implementation

andrula-song:
Add HiFi5 implementation of aria algorithm functions, compared with HiFi3 version, can reduce about 10% cycles.

### DIFF
--- a/src/audio/aria/aria.h
+++ b/src/audio/aria/aria.h
@@ -30,7 +30,9 @@
 
 #if defined(__XCC__)
 # include <xtensa/config/core-isa.h>
-# if XCHAL_HAVE_HIFI3 || XCHAL_HAVE_HIFI4
+# if XCHAL_HAVE_HIFI5
+#  define ARIA_HIFI5
+# elif XCHAL_HAVE_HIFI3 || XCHAL_HAVE_HIFI4
 #  define ARIA_HIFI3
 # else
 #  define ARIA_GENERIC
--- a/src/audio/aria/aria_hifi5.c
+++ b/src/audio/aria/aria_hifi5.c
@@ -1,0 +1,188 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// Copyright(c) 2024 Intel Corporation. All rights reserved.
+
+#include "aria.h"
+
+#ifdef ARIA_HIFI5
+#include <xtensa/config/defs.h>
+#include <xtensa/tie/xt_hifi5.h>
+
+/**
+ * \brief Aria gain index mapping table
+ */
+static const int32_t index_tab[] = {
+		0,    1,    2,    3,
+		4,    5,    6,    7,
+		8,    9,    0,    1,
+		2,    3,    4,    5,
+		6,    7,    8,    9,
+		0,    1,    2,    3
+};
+
+/* Setup circular buffer 0 */
+static inline void set_circular_buf0(const void *start, const void *end)
+{
+	AE_SETCBEGIN0(start);
+	AE_SETCEND0(end);
+}
+
+/* Setup circular for buffer 1 */
+static inline void set_circular_buf1(const void *start, const void *end)
+{
+	AE_SETCBEGIN1(start);
+	AE_SETCEND1(end);
+}
+
+inline void aria_algo_calc_gain(struct aria_data *cd, size_t gain_idx,
+				struct audio_stream *source, int frames)
+{
+	/* detecting maximum value in data chunk */
+	ae_int32x2 in_sample;
+	ae_int32x2 in_sample1;
+	ae_int32x2 max_data = AE_ZERO32();
+	int32_t att = cd->att;
+	ae_valignx2 inu = AE_ZALIGN128();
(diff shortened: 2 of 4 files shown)


andrula-song:
here is the xtensa simulator result:
Screenshot from 2024-01-04 10-28-01
compared with hifi3 version, the HiFi5 version can save:
about 5% for aria_algo_get_data_odd_channel;
about 10% for aria_algo_get_data_even_channel;
about 15% for aria_algo_calc_gain.

src/audio/aria/aria_hifi5.c
```diff
@@ -0,0 +1,183 @@
...
+#include "aria.h"
+
+#ifdef ARIA_HIFI5
+#include <xtensa/config/defs.h>
+#include <xtensa/tie/xt_hifi5.h>
+
+/**
+ * \brief Aria gain index mapping table
+ */
+const uint8_t INDEX_TAB[] = {
```
lgirdwood, line 14: this is being cast to a uint32_t in the code and should simplify the L32 loads

andrula-song: from the xtensa simulator I didn't see any improvement, the purpose I use the uint8_t is to save memory. But it is okay to change it to a uint32_t table if we don't mind the memory usage.

lgirdwood: interesting, the L32 of a uint32_t is 1 op, but to load a byte is a L32 and a shift/mask.

src/audio/aria/aria_hifi5.c
```diff
@@ -0,0 +1,188 @@
...
+#include "aria.h"
+
+#ifdef ARIA_HIFI5
+#include <xtensa/config/defs.h>
+#include <xtensa/tie/xt_hifi5.h>
+
+/**
+ * \brief Aria gain index mapping table
+ */
+const int32_t INDEX_TAB[] = {
```
lgirdwood, line 14: static and naming convention would be lower case.

lgirdwood (commented):
@singalsu needs your review.

src/audio/aria/aria_hifi5.c
```diff
@@ -0,0 +1,188 @@
...
+ * \brief Aria gain index mapping table
+ */
+static const int32_t index_tab[] = {
+		0,    1,    2,    3,
+		4,    5,    6,    7,
+		8,    9,    0,    1,
+		2,    3,    4,    5,
+		6,    7,    8,    9,
+		0,    1,    2,    3
+};
```
singalsu, line 21: This looks like modulo 10 for 0 to 23 index. Likely not possible but if gains gains table size ARIA_MAX_GAIN_STATES  could be 16 or other 2^N value this could be bit-and instead of modulo. But it's out-of-scope of this PR to change functionality.

singalsu: > This looks like modulo 10 for 0 to 23 index. Likely not possible while keeping the algorithm the same but if gains gains table size ARIA_MAX_GAIN_STATES could be 16 or other 2^N value this could be bit-and instead of modulo. But it's out-of-scope of this PR to change functionality.

singalsu (approved):
Looks good to me, I got the patch and compared to aria_hifi3.c.

src/audio/aria/aria_hifi5.c
btian1: @lgirdwood @andrula-song , there is load byte instruction, so no need use load32 for this table, save memory is always good:

Load 8-bit Unsigned L8UI
Xtensa Instruction Set Architecture (ISA) Reference Manual 449
Instruction Word (RRI8)
Required Configuration Option
Core Architecture (See Section 4.2 on page 51)
Assembler Syntax
L8UI at, as, 0..255
Description
L8UI is an 8-bit unsigned load from memory. It forms a virtual address by adding the
contents of address register as and an 8-bit zero-extended constant value encoded in
the instruction word. Therefore, the offset ranges from 0 to 255. Eight bits (one byte) are
read from the physical address. This data is then zero-extended and written to address
register at.
If the Region Translation Option (page 185) or the MMU Option (page 199)is enabled,
the virtual address is translated to the physical address. If not, the physical address is
identical to the virtual address. If the translation or memory reference encounters an
error (for example, protection violation or non-existent memory), the processor raises
one of several exceptions (see Section 4.4.2.5 on page 105)

lgirdwood (approved):

src/audio/aria/aria.h
```diff
@@ -30,7 +30,9 @@
 
 #if defined(__XCC__)
 # include <xtensa/config/core-isa.h>
-# if XCHAL_HAVE_HIFI3 || XCHAL_HAVE_HIFI4
+# if XCHAL_HAVE_HIFI5
+#  define ARIA_HIFI5
+# elif XCHAL_HAVE_HIFI3 || XCHAL_HAVE_HIFI4
```
marc-hb, line 35: Please switch to the new, de-duplicated approach from #8783

PR #8695, opened 2024-01-04
Audio: Component: HiFi5 implementation of functions.

andrula-song:
1. Add HiFi5 implementation of audio_stream_copy, an save about 29% cycles.
2. Add HiFi5 implementation of cir_buf_copy, can save about 40% cycles.


andrula-song:
with the xtensa simulator, here are the test result:
Screenshot from 2024-01-04 14-40-56

src/audio/component.c
```diff
@@ -180,7 +182,57 @@ void comp_get_copy_limits_frame_aligned(const struct comp_buffer *source,
...
+	int shorts_dst;
+	int shorts_copied;
+	int left, m, i;
+	ae_int16x4 in_sample1;
+	ae_int16x4 in_sample2;
+	ae_valignx2 inu;
+	ae_valignx2 outu = AE_ZALIGN128();
+
+	/* copy with 16bit as the minimum unit since the minimum sample size is 16 bit*/
+	while (shorts) {
```
lgirdwood, line 206: unsigned ? easy to loop forever if we undershoot ```shorts``` ?

lgirdwood (commented):
Good stuff, we can address the Kconfig work later as part of teh work that @btian1 is doing.

src/audio/component.c
```diff
@@ -240,7 +240,7 @@ int audio_stream_copy(const struct audio_stream *source, uint32_t ioffset,
 	int ssize = audio_stream_sample_bytes(source); /* src fmt == sink fmt */
 	ae_int16x4 *src = (ae_int16x4 *)((int8_t *)audio_stream_get_rptr(source) + ioffset * ssize);
 	ae_int16x4 *dst = (ae_int16x4 *)((int8_t *)audio_stream_get_wptr(sink) + ooffset * ssize);
-	int shorts = samples * ssize >> 1;
+	size_t shorts = samples * ssize >> 1;
```
lyakh, line 287: sorry, I don't understand this. Firstly, I think `size_t` is supposed to be used as a number of bytes, whereas here it's a number of `short`s. Secondly, "reducing the risk of a forever loop" - shouldn't such a risk actually be made to be equal to 0? Besides, I don't see how this change affects that risk at all - the loop is over `while (shorts)` so it doesn't seem to be affected by this type change. Thirdly, IIUC `AE_MIN_32_signed(shorts, shorts_copied)` is a signed comparison, so I don't understand why `shorts` now has to be unsigned.

singalsu (approved):

singalsu:
Good improvement! Approve - with fix suggested by Guennadi.

src/audio/component.c
andrula-song: so better keep the signed type and use while(short > 0) to avoid the risk, though it shouldn't happen.

PR #8700, opened 2024-01-05
topology2: Deep Buffer cleanups

ujfalusi:
Hi,

Set DEEPBUFFER_D0I3_COMPATIBLE to **true** by default for all configuration and remove the individual topology parameters.
Reasoning:
The DeepBuffer stream is D0I3 compatible for some time now on all IPC4 platforms, The CAVS2.5 topologies just did not got updated.

Remove the DEEPBUFFER_FW_DMA_MS=100 from individual topology parameter list as it is redundant, the default value is 100.


plbossart (approved):
@ranj063 must approve this one before mergng

kv2019i (approved):

jsarha (approved):
Looks technically sound. There is couple of topologies that used to have DEEPBUFFER_D0I3_COMPATIBLE=false before this, but I guess that is intentional. Wonder, if that would still be worth mentioning explicitly int he commit message.

ujfalusi:
> Looks technically sound. There is couple of topologies that used to have DEEPBUFFER_D0I3_COMPATIBLE=false before this, but I guess that is intentional. Wonder, if that would still be worth mentioning explicitly int he commit message.

Hrm, I think we  only have DEEPBUFFER_D0I3_COMPATIBLE=true specified for topologies. True that when it was not set those were taking the false, but this is mentioned in the commit message. I have re-checked again and I did not removed any DEEPBUFFER_D0I3_COMPATIBLE=false

lgirdwood (approved):

tools/topology/topology2/development/tplg-targets.cmake
```diff
@@ -28,16 +28,13 @@ PLATFORM=adl"
 
 # SSP topology for MTL
 "cavs-nocodec\;sof-mtl-nocodec\;PLATFORM=mtl,NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,\
-PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-mtl-nocodec.bin,DEEPBUFFER_FW_DMA_MS=100,\
-DEEPBUFFER_D0I3_COMPATIBLE=true"
```
ranj063, line 32: I think this flag is controversial even for MTL. The intention of the D0i3_COMPATIBLE flag for a pcm was to ensure that while the stream is suspended by ALSA, it can still keep running the DSP. While this works for streams like wov, we have no idea how to do this for playback. So until we figure out how to do this, my vote would be to set this to false for all platforms or even better just remove it altogether for now. @plbossart @kv2019i what do you think?

ranj063: @plbossart what I was referring to was enabling S0ix playback. If we set a stream as D0i3 compatible and suspend the system to S0ix, the stream will get suspended by ALSA but can it really keep running in the DSP in this case? I suppose we have never tried this.

plbossart: @ranj063 I suggest we try walking before running. the first issue to solve is to see if we can detect 'idleness' in S0/D0i0 and transition to S0/D0i3 without any issues in audio rendering.

The S0i1 case is different, the host side will need to periodically refill buffers and that means applications need to keep on running. I am not sure this can be done either, but that's the second level of optimization.

ranj063: @plbossart my point is that the deep buffer stream with this flag set will not be able to resume from S0ix. We don't test this in our CI today, so we havent seen it fail.

ujfalusi: @ranj063, should then it be disabled? The DEEPBUFFER_D0I3_COMPATIBLE is applied to DeepBuffer PCMs and they are playback PCMs, so the flag is wrong and it is incorrectly enabled?

PR #8701, opened 2024-01-05
aria: change to S24_LE format processing

fkwasowi:
After the following change:
"copier: add support for windows driver (383d17a19e25f804d2ba37b6b621131918544e6a)"
aria stopped working properly. It is necessary to adapt the module
S24_LE format processing.

### DIFF
--- a/src/audio/aria/aria_generic.c
+++ b/src/audio/aria/aria_generic.c
@@ -40,10 +40,9 @@ inline void aria_algo_calc_gain(struct aria_data *cd, size_t gain_idx,
 		src = audio_stream_wrap(source, src + n);
 		samples -= n;
 	}
-
 	/*zero check for maxis not needed since att is in range <0;3>*/
-	if (max_data > (0x7fffffff >> att))
-		gain = (0x7fffffffULL << 32) / max_data;
+	if (max_data > (0x007fffff >> att))
+		gain = (0x007fffffULL << 32) / max_data;
 
 	cd->gains[gain_idx] = (int32_t)(gain >> (att + 1));
 }
@@ -83,7 +82,7 @@ void aria_algo_get_data(struct processing_module *mod,
 		for (i = 0; i < n; i += ch_n) {
 			for (ch = 0; ch < ch_n; ch++) {
 				in_sample = *in++;
-				out[ch] = q_multsr_sat_32x32(in_sample, gain, shift);
+				out[ch] = q_multsr_sat_32x32_24(in_sample, gain, shift);
 			}
 			gain += step;
 			out += ch_n;
--- a/src/audio/aria/aria_hifi3.c
+++ b/src/audio/aria/aria_hifi3.c
@@ -53,8 +53,8 @@ inline void aria_algo_calc_gain(struct aria_data *cd, size_t gain_idx,
 		samples -= n;
 	}
 	/*zero check for maxis not needed since att is in range <0;3>*/
-	if (max > (0x7fffffff >> att))
-		gain = (0x7fffffffULL << 32) / max;
+	if (max > (0x007fffff >> att))
+		gain = (0x007fffffULL << 32) / max;
 
 	/* normalization by attenuation factor to obtain fractional range <1 / (2 pow att), 1> */
 	cd->gains[gain_idx] = (int32_t)(gain >> (att + 1));
@@ -74,7 +74,6 @@ void aria_algo_get_data_odd_channel(struct processing_module *mod,
 	int32_t gain_end = cd->gains[INDEX_TAB[gain_state_add_3]];
 	size_t samples = frames * audio_stream_get_channels(sink);
 	ae_int32x2 *out = audio_stream_get_wptr(sink);
-	int32_t att = cd->att;
 	ae_int32x2 *in = (ae_int32x2 *)cd->data_ptr;
 	ae_valign inu = AE_ZALIGN64();
 	ae_valign outu = AE_ZALIGN64();
@@ -82,6 +81,8 @@ void aria_algo_get_data_odd_channel(struct processing_module *mod,
 	const int inc = sizeof(ae_int32);
 	ae_int32x2 gain;
 	const int ch_n = cd->chan_cnt;
+	const int shift_bits = 31 - cd->att - 24;
+	ae_int64 out1;
 
 	for (i = 1; i < ARIA_MAX_GAIN_STATES - 1; i++) {
 		if (cd->gains[INDEX_TAB[gain_state_add_2 + i]] < gain_begin)
@@ -102,9 +103,9 @@ void aria_algo_get_data_odd_channel(struct processing_module *mod,
 			/*process data one by one if ch_n is odd*/
 			for (ch = 0; ch < ch_n; ch++) {
 				AE_L32_XP(in_sample, (ae_int32 *)in, inc);
-				out_sample = AE_MULFP32X2RS(in_sample, gain);
-
-				out_sample = AE_SLAA32S(out_sample, att);
+				out1 = AE_MUL32_HH(in_sample, gain);
+				out1 = AE_SRAA64(out1, shift_bits);
+				out_sample = AE_ROUND24X2F48SSYM(out1, out1);
 				AE_S32_L_XP(out_sample, (ae_int32 *)out, inc);
 			}
 			gain = AE_ADD32S(gain, step);
@@ -131,13 +132,14 @@ void aria_algo_get_data_even_channel(struct processing_module *mod,
 	int32_t gain_end = cd->gains[INDEX_TAB[gain_state_add_3]];
 	size_t samples = frames * audio_stream_get_channels(sink);
 	ae_int32x2 *out = audio_stream_get_wptr(sink);
-	int32_t att = cd->att;
 	ae_int32x2 *in = (ae_int32x2 *)cd->data_ptr;
 	ae_valign inu = AE_ZALIGN64();
 	ae_valign outu = AE_ZALIGN64();
 	ae_int32x2 in_sample, out_sample;
 	ae_int32x2 gain;
 	const int ch_n = cd->chan_cnt;
+	const int shift_bits = 31 - cd->att - 24;
+	ae_int64 out1, out2;
 
 	for (i = 1; i < ARIA_MAX_GAIN_STATES - 1; i++) {
 		if (cd->gains[INDEX_TAB[gain_state_add_2 + i]] < gain_begin)
@@ -158,9 +160,11 @@ void aria_algo_get_data_even_channel(struct processing_module *mod,
 			/*process 2 samples per time if ch_n is even*/
 			for (ch = 0; ch < ch_n; ch += 2) {
 				AE_LA32X2_IP(in_sample, inu, in);
-				out_sample = AE_MULFP32X2RS(in_sample, gain);
-
-				out_sample = AE_SLAA32S(out_sample, att);
+				out1 = AE_MUL32_HH(in_sample, gain);
+				out1 = AE_SRAA64(out1, shift_bits);
+				out2 = AE_MUL32_LL(in_sample, gain);
+				out2 = AE_SRAA64(out2, shift_bits);
+				out_sample = AE_ROUND24X2F48SSYM(out1, out2);
 				AE_SA32X2_IP(out_sample, outu, out);
 			}
 			gain = AE_ADD32S(gain, step);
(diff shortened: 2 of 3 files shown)


src/audio/aria/aria_generic.c
```diff
@@ -83,7 +83,8 @@ void aria_algo_get_data(struct processing_module *mod,
 		for (i = 0; i < n; i += ch_n) {
 			for (ch = 0; ch < ch_n; ch++) {
 				in_sample = *in++;
-				out[ch] = q_multsr_sat_32x32(in_sample, gain, shift);
+				out[ch] =
+				 sat_int32((((int64_t)in_sample * gain) >> (shift - 1)) >> 1);
```
andrula-song, line 87: seems the shift and truncation could be 
`out[ch] =sat_int32((int64_t)in_sample * gain) >>shift); `

src/audio/aria/aria_hifi3.c
```diff
@@ -102,9 +105,13 @@ void aria_algo_get_data_odd_channel(struct processing_module *mod,
...
-				out_sample = AE_MULFP32X2RS(in_sample, gain);
-
-				out_sample = AE_SLAA32S(out_sample, att);
+				out1 = AE_MUL32_HH(in_sample, gain);
+				out1 = AE_SRAA64(out1, shift_bits);
+				out1 = AE_SLAA64(out1, 16);
+				out2 = AE_MUL32_LL(in_sample, gain);
+				out2 = AE_SRAA64(out2, shift_bits);
+				out2 = AE_SLAA64(out2, 16);
+				out_sample = AE_ROUND32X2F48SSYM(out1, out2);
```
andrula-song, line 114: sorry forgot to tell you if for the odd channel, only out1 is enough since the H&L part of in_sample is the same. it could be like:
```
				out1 = AE_MUL32_HH(in_sample, gain);
				out1 = AE_SRAA64(out1, shift_bits);
				out1 = AE_SLAA64(out1, 16);
                                out_sample = AE_ROUND32X2F48SSYM(out1, out1);
```

kv2019i (commented):
The change looks ok, but I'm surprised aria module doesn't have any check to validate the input it gets? If the implementation only supports one variant of 24bit (either LSB or MSB encoding), it should raise an error if it is fed with data in wrong format.

src/audio/aria/aria_generic.c
```diff
@@ -40,7 +40,7 @@ inline void aria_algo_calc_gain(struct aria_data *cd, size_t gain_idx,
 		src = audio_stream_wrap(source, src + n);
 		samples -= n;
 	}
-
+	max_data = max_data << 8;
```
singalsu, line 43: Doesn't this overflow with s32_le format?

src/audio/aria/aria_generic.c
```diff
@@ -83,7 +83,7 @@ void aria_algo_get_data(struct processing_module *mod,
 		for (i = 0; i < n; i += ch_n) {
 			for (ch = 0; ch < ch_n; ch++) {
 				in_sample = *in++;
-				out[ch] = q_multsr_sat_32x32(in_sample, gain, shift);
+				out[ch] = sat_int32(((int64_t)in_sample * gain) >> shift);
```
singalsu, line 86: Why is the rounding removed? If really needed, could also add q_mults_sat_32x32() function to format.h. Seems such does not exist.

src/audio/aria/aria_hifi3.c
```diff
@@ -52,6 +52,7 @@ inline void aria_algo_calc_gain(struct aria_data *cd, size_t gain_idx,
 		in = audio_stream_wrap(source, in);
 		samples -= n;
 	}
+	max = max << 8;
```
singalsu, line 55: same s32_le format worry as for generic

src/audio/aria/aria_hifi3.c
```diff
@@ -82,6 +83,8 @@ void aria_algo_get_data_odd_channel(struct processing_module *mod,
 	const int inc = sizeof(ae_int32);
 	ae_int32x2 gain;
 	const int ch_n = cd->chan_cnt;
+	const int shift_bits = 31 - cd->att;
```
singalsu, line 86: Does this mean that aria needs different "att" initialize for different format? It would be more efficient to have a different processing core for s24_4le and s32_le formats as for many other components.

singalsu:
Also the PR title is confusing. It would be better to refer to e.g. ALSA format names like S24_LE and such.

src/audio/aria/aria_hifi3.c
```diff
@@ -102,9 +105,10 @@ void aria_algo_get_data_odd_channel(struct processing_module *mod,
 			/*process data one by one if ch_n is odd*/
 			for (ch = 0; ch < ch_n; ch++) {
 				AE_L32_XP(in_sample, (ae_int32 *)in, inc);
-				out_sample = AE_MULFP32X2RS(in_sample, gain);
-
-				out_sample = AE_SLAA32S(out_sample, att);
+				out1 = AE_MUL32_HH(in_sample, gain);
+				out1 = AE_SRAA64(out1, shift_bits);
+				out1 = AE_SLAA64(out1, 16);
```
btian1, line 110: you can save one shift here, 
1. right shift: 32 - atten 
2. left shift: 16

means you can simply >> 32 - atten - 16, I don't think attenuation > 16, right?

src/audio/aria/aria_hifi3.c
```diff
@@ -158,9 +164,13 @@ void aria_algo_get_data_even_channel(struct processing_module *mod,
 			/*process 2 samples per time if ch_n is even*/
 			for (ch = 0; ch < ch_n; ch += 2) {
 				AE_LA32X2_IP(in_sample, inu, in);
-				out_sample = AE_MULFP32X2RS(in_sample, gain);
-
-				out_sample = AE_SLAA32S(out_sample, att);
+				out1 = AE_MUL32_HH(in_sample, gain);
+				out1 = AE_SRAA64(out1, shift_bits);
+				out1 = AE_SLAA64(out1, 16);
```
btian1, line 169: same

src/audio/aria/aria_hifi3.c
```diff
@@ -158,9 +164,13 @@ void aria_algo_get_data_even_channel(struct processing_module *mod,
...
+				out2 = AE_MUL32_LL(in_sample, gain);
+				out2 = AE_SRAA64(out2, shift_bits);
+				out2 = AE_SLAA64(out2, 16);
```
btian1, line 172: same

src/audio/aria/aria_generic.c
```diff
@@ -40,7 +40,7 @@ inline void aria_algo_calc_gain(struct aria_data *cd, size_t gain_idx,
 		src = audio_stream_wrap(source, src + n);
 		samples -= n;
 	}
-
+	max_data = max_data << 8;
 	/*zero check for maxis not needed since att is in range <0;3>*/
 	if (max_data > (0x7fffffff >> att))
 		gain = (0x7fffffffULL << 32) / max_data;
```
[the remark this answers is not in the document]
tlissows: The both occurrences of `0x7fffffff` (line 45 and line 46) need to be changed.

src/audio/aria/aria_hifi3.c
tlissows, line 1: This patch leaves two unused `int32_t att` variables - one in function `aria_algo_get_data_odd_channel` and the other in function `aria_algo_get_data_even_channel`.

tlissows (approved):

kv2019i:
@kv2019i wrote:
> The change looks ok, but I'm surprised aria module doesn't have any check to validate the input it gets? If the implementation only supports one variant of 24bit (either LSB or MSB encoding), it should raise an error if it is fed with data in wrong format.

@fkwasowi any comments to this? It seems before the change, this module supports 32bit samples (same code can handle 24bit in 32bit-MSB). Now there's a change that starts using this module in a configuration where it will get 24bit LSB data, but doesn't this mean this code no longer works with SOF_IPC_FRAME_S32_LE and SOF_IPC_FRAME_S24_4LE_MSB. 

Given this module is used in limited topologies, I'm ok to limit its applicability (e.g. to only SOF_IPC_FRAME_S24_4LE), but I think some check should be added to raise an error that if anyone tries to feed this module unsupported formats, it will raise an error.

src/audio/aria/aria_hifi3.c
fkwasowi: The concept of the fix has changed and this problem no longer exists.

src/audio/aria/aria_hifi3.c
fkwasowi: The concept of the fix has changed and this problem no longer exists.

src/audio/aria/aria_hifi3.c
fkwasowi: The concept of the fix has changed and this problem no longer exists.

src/audio/aria/aria_hifi3.c
fkwasowi: The concept of the fix has changed and this problem no longer exists.

fkwasowi:
> Also the patch title is confusing. It would be better to refer to e.g. ALSA format names like S24_LE and such and what issue is being fixed. The title should be such that is helpful when looking afterwards git history of patch titles.

I have changed the title and description.

lgirdwood (commented):
@fkwasowi @andrula-song I assume no MCPS impact ?

fkwasowi:
> @kv2019i wrote:
> 
> > The change looks ok, but I'm surprised aria module doesn't have any check to validate the input it gets? If the implementation only supports one variant of 24bit (either LSB or MSB encoding), it should raise an error if it is fed with data in wrong format.
> 
> @fkwasowi any comments to this? It seems before the change, this module supports 32bit samples (same code can handle 24bit in 32bit-MSB). Now there's a change that starts using this module in a configuration where it will get 24bit LSB data, but doesn't this mean this code no longer works with SOF_IPC_FRAME_S32_LE and SOF_IPC_FRAME_S24_4LE_MSB.
> 
> Given this module is used in limited topologies, I'm ok to limit its applicability (e.g. to only SOF_IPC_FRAME_S24_4LE), but I think some check should be added to raise an error that if anyone tries to feed this module unsupported formats, it will raise an error.

I added a condition to check the format

src/audio/aria/aria_generic.c
fkwasowi: The concept of the fix has changed and this problem no longer exists.

fkwasowi:
> I assume no MCPS impact ?

The change does not affect MCPS

lgirdwood (approved):

iganakov (approved):
Looks good to me

kv2019i (approved):
Thanks @fkwasowi , looks good now. Let's wait for the CI runs to complete and then this is ready to go.

src/audio/aria/aria_generic.c
```diff
@@ -40,10 +40,9 @@ inline void aria_algo_calc_gain(struct aria_data *cd, size_t gain_idx,
 		src = audio_stream_wrap(source, src + n);
 		samples -= n;
 	}
-
 	/*zero check for maxis not needed since att is in range <0;3>*/
-	if (max_data > (0x7fffffff >> att))
-		gain = (0x7fffffffULL << 32) / max_data;
+	if (max_data > (0x007fffff >> att))
+		gain = (0x007fffffULL << 32) / max_data;
```
btian1, line 45: << 32 or << 24? since we are handling with 24LE sample.

src/audio/aria/aria_generic.c
```diff
@@ -40,10 +40,9 @@ inline void aria_algo_calc_gain(struct aria_data *cd, size_t gain_idx,
 		src = audio_stream_wrap(source, src + n);
 		samples -= n;
 	}
-
 	/*zero check for maxis not needed since att is in range <0;3>*/
-	if (max_data > (0x7fffffff >> att))
-		gain = (0x7fffffffULL << 32) / max_data;
+	if (max_data > (0x007fffff >> att))
```
btian1, line 44: there is definition for int24, you can reuse it:
src/include/sof/audio/format.h:33:#define INT24_MAXVALUE  8388607
src/include/sof/audio/format.h:34:#define INT24_MINVALUE -8388608

fkwasowi:
SOFCI TEST

fkwasowi:
SOFCI TEST

src/audio/aria/aria_generic.c
kv2019i: @btian1 I think the PR seems correct, the resulting is gain is same with max S24 signal value. @fkwasowi please ack

PR #8708, opened 2024-01-07
common.h: Guard definition of IS_ALIGNED for Zephyr builds

yonsch:
common.h defines the IS_ALIGNED macro, which conflicts with a Zephyr macro of the same name. Add guards to define it only if it is not already defined by Zephyr.

### DIFF
--- a/src/include/sof/common.h
+++ b/src/include/sof/common.h
@@ -14,8 +14,14 @@
 /* callers must check/use the return value */
 #define __must_check __attribute__((warn_unused_result))
 
+#ifdef __ZEPHYR__
+#include <zephyr/sys/util.h>
+#endif
+
 /* Align the number to the nearest alignment value */
+#ifndef IS_ALIGNED
 #define IS_ALIGNED(size, alignment) ((size) % (alignment) == 0)
+#endif
 
 /* Treat zero as a special case because it wraps around */
 #define is_power_of_2(x) ((x) && !((x) & ((x) - 1)))


sofci:

btian1 (approved):
There is another place in Zephyr also defined this in zephyr code, could you change zephyr code as well?
https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/crypto/crypto_intel_sha_priv.h#L33

kv2019i (approved):
Thanks @yonsch !

src/include/sof/common.h
```diff
@@ -15,7 +15,11 @@
 #define __must_check __attribute__((warn_unused_result))
 
 /* Align the number to the nearest alignment value */
+#ifndef __ZEPHYR__
 #define IS_ALIGNED(size, alignment) ((size) % (alignment) == 0)
+#else
+#include <zephyr/sys/util.h>
+#endif
```
lyakh, line 22: this is referring to https://github.com/zephyrproject-rtos/zephyr/pull/67243 which isn't merged yet, right? So if we merge this first this will break Zephyr builds? This cannot be merged like this. A simple solution would be to use `#ifnder IS_ALIGNED` as a guard.

lyakh (changes requested):

src/include/sof/common.h
lgirdwood: Long term we move this and other common Zephyr re-definitions to the xtos header directory for remaining xtos users and pull this directly from Zephyr.

kv2019i (approved):
Refresh +1 , thanks @yonsch ! 

@lyakh can you check again?

lyakh (approved):

lgirdwood:
Looks like some CI has timed out -  restart CI.

lgirdwood:
SOFCI TEST

kv2019i:
@wszypelt This PR seems to have got stuck in quickbuild, can you check.
For other results, the one fail in https://sof-ci.01.org/sofpr/PR8708/build1825/devicetest/index.html can be ignored.

src/include/sof/common.h
```diff
@@ -14,8 +14,14 @@
 /* callers must check/use the return value */
 #define __must_check __attribute__((warn_unused_result))
 
+#ifdef __ZEPHYR__
+#include <zephyr/sys/util.h>
+#endif
+
 /* Align the number to the nearest alignment value */
+#ifndef IS_ALIGNED
```
marc-hb, line 22: Except for "extreme" cases, `#ifndef` is always best avoided because it makes the usual, horrible macro experience even worse. Zephyr forbids it: https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html#rule-a-3-macro-name-collisions

As requested in https://github.com/zephyrproject-rtos/zephyr/pull/67243#discussion_r1454373787, please rename this `SOF_ALIGN()` and make everything simpler: backwards and forwards compatibility and trannsitions

marc-hb: > Can you do a follow-up on SOF side. 

Submitted in #8755

Issue #8683, opened 2024-01-02
[FEATURE] move to platform specific HDA tplg files


kv2019i:
**Is your feature request related to a problem? Please describe.**
With IPC4, the tplg files can contain platform specific configuration objects (like NHLT blobs). In practise, all Intel tplg files that have DAIs for DMIC or SSP, will have such blobcs. E.g. topology file like "sof-hda-generic-2ch.tplg" is no longer platform agnostic, so for platforms newer than Intel MTL, we need to ship multiple versions of the these topologies, and provide a unique naming so that kernel can look up the correct topology file.

Additionally, we must maintain backward compatibility for platforms already shipping (e.g. for IPC4, this means Intel MTL platforms that are included in SOF2.8 release).

**Describe the solution you'd like**
Add a platform prefix to all topologies that have platform dependent configuration objects. Keep old naming for platforms like Intel MTL that are already shipping.

**Describe alternatives you've considered**
Using different directories and duplicating all tplgs. This adds overhead to release process if we need to duplicate all topologies (even when they are in fact the same).

**Additional context**
Related to SOF2.8 release process lessons learnt, kernel PR https://github.com/thesofproject/linux/pull/4382 and discussion about better system to package the SOF topology files for sof-bin distribution.

ujfalusi:
@kv2019i, my proposal for the IPC4 tplg paths, which will scale well for the future and past (I will make a sof-docs update with this):
The collision is really about the generic HDA topologies with DMIC (and some development ones using generic names). We could just try to separate these per architecture/platform but I feel that creating a clear rule for that is not going to be easy. We would need to define the 'generic', have different path lookup for them, etc.
I would rather go via a simple and unimaginative approach and store all tplg files in separate directories. This would give us backwards compatibility with older kernels.

Option **A**
```
.
└── intel
    ├── sof-ace-tplg -> sof-ipc4-tplg/ace1
    └── sof-ipc4-tplg
        ├── ace1
        │   ├── sof-hda-generic-2ch.tplg
        │   ├── sof-hda-generic-4ch.tplg
        │   ├── sof-hda-generic-idisp.tplg
        │   ├── sof-hda-generic.tplg
        │   ├── sof-lnl-rt711-4ch.tplg
        │   ├── sof-lnl-rt711-l0-rt1316-l23-rt714-l1.tplg
        │   ├── sof-mtl-cs42l43-l0-cs35l56-l12.tplg
[...]

PR #8712, opened 2024-01-08
topology2: Merge avs-tplg and sof-ace-tplg under production directory

ujfalusi:
Merge the avs-tplg and sof-ace-tplg under a common production directory.

After a successful build CMake will copy the topology files to a target
directory from where they can be copied to DUT/release:

$ tree tools/build_tools/topology/topology2/target
tools/build_tools/topology/topology2/target
├── development
│   ├── cavs-sdw-hdmi.tplg
│   ├── cavs-sdw-src-gain-mixin.tplg
        ...
│   ├── sof-tgl-nocodec-rtcaec.tplg
│   └── sof-tgl-nocodec.tplg
├── sof-ace-tplg -> sof-ipc4-tplg
└── sof-ipc4-tplg
    ├── sof-adl-rt711-4ch.tplg
    ├── sof-adl-rt711-l0-rt1316-l12-rt714-l3.tplg
        ...
    ├── sof-tgl-rt712.tplg
    └── sof-tgl-rt715-rt711-rt1308-mono.tplg

The sof-hda-generic-2/4ch.tplg will be generated without embedded NHLT as
it is not used under normal circumstance.
Two flavor of the generic topology is generated for CAVS2.5 and ACE1/2 with
included NHLT binary in case it is used by existing users, but it is
unlikely.

As noted in the documentation, on the deployed system a symlink is needed
for ACE1/2 platforms for backwards compatibility:
sof-ace-tplg -> sof-ipc4-tplg

Link: https://github.com/thesofproject/sof-docs/blob/master/getting_started/intel_debug/introduction.rst#2-topology-file


ujfalusi:
This is going to fail in CI as all IPC4 firmware files are under sof-ipc4-tplg.
On upx and RVPs where the BIOS does not contain NHLT blob, the appropriate hda-generic topology needs to be specified as topology file name override:
CAVS2.5 platforms: sof-cavs25-nhlt-hda-generic-2ch.tplg | sof-cavs25-nhlt-hda-generic-4ch.tplg
AVS1/2 platforms: sof-ace1-nhlt-hda-generic-2ch.tplg | sof-ace1-nhlt-hda-generic-4ch.tplg

The generic HDA topologies does not contain NHLT binary!

lgirdwood:
> This is going to fail in CI as all IPC4 firmware files are under sof-ipc4-tplg. On upx and RVPs where the BIOS does not contain NHLT blob, the appropriate hda-generic topology needs to be specified as topology file name override: CAVS2.5 platforms: sof-cavs25-nhlt-hda-generic-2ch.tplg | sof-cavs25-nhlt-hda-generic-4ch.tplg AVS1/2 platforms: sof-ace1-nhlt-hda-generic-2ch.tplg | sof-ace1-nhlt-hda-generic-4ch.tplg
> 
> The generic HDA topologies does not contain NHLT binary!

@ujfalusi What is your proposed solution to fix the issue in CI ?

marc-hb:
> This is going to fail in CI as all IPC4 firmware files are under sof-ipc4-tplg.

If CI needs to (stop!) renaming some directories or create some symbolic link(s) then just specify very clearly what it must do or stop doing. Then we'll merge all changes at about the same time.

The only purpose of tests and CI is to find problems before customers do and every CI hack and CI "secret" has been for too long a failure to do that (admittedly not helped by a lack of concern and documentation before @ujfalusi stood up)

Let's worry about sof-bin /lib/firmware/sof/ on Ubuntu, Fedora, etc. including backwards compatibility with older kernels! That is the important and difficult part. Temporary CI failures can and must be fixed by aligning with releases and customers.

ujfalusi:
@marc-hb, @lgirdwood, @kv2019i, with this PR CI (and sof-bin) should do this for all IPC4 platforms:
```
ssh root@DUT
cd /lib/firmware/intel
mkdir sof-ipc4-tplg
ln -sf sof-ipc4-tplg sof-ace-tplg
exit
scp tools/build_tools/topology/topology2/sof-ipc4-tplg/*.tplg root@DUT:/lib/firmware/intel/sof-ipc4-tplg/
```
**sof-bin note:** remove the CAVS2.5 and ACE2 topologies before release (sof-tgl/adl/etc and sof-cavs25-*) for now?

After this DUT should just work, but...

We have machines in CI (UPX I think) which does not have NHLT blob in BIOS and we  were using the topology embedded one.
`sof-hda-generic-2ch.tplg` and `sof-hda-generic-4ch.tplg` no longer contains NHLT blob, so on devices where we need NHLT from the topology file (only affecting sof-hda-gneeric-2/4ch users!!!) we need to override the topology name:
on CAVS2.5 platforms:
`sof-cavs25-nhlt-hda-generic-2ch.tplg` instead of `sof-hda-generic-2ch.tplg`
`sof-cavs25-nhlt-hda-generic-4ch.tplg` instead of `sof-hda-generic-4ch.tplg`

on ACE1 and ACE2 platforms:
`sof-ace1-nhlt-hda-generic-2ch.tplg` instead of `sof-hda-generic-2ch.tplg`
`sof-ace1-nhlt-hda-generic-4ch.tplg` instead of `sof-hda-generic-4ch.tplg`

The reason is that the blob format differs between architectures, so the sof-hda-generic* is converted to be generic, it works on systems where the NHLT is in BIOS, however for developers (and unfortunate users) we have the topologies with NHLT.

lgirdwood:
@ujfalusi lets base solution on what the upstream driver loads today for MTL, we can adapt kernel topology preference from LNL onwards. I would ideally like to leave MTL (and everything before) with the naming convention it uses today and land any new schema from LNL onwards.

ujfalusi:
> @ujfalusi lets base solution on what the upstream driver loads today for MTL, we can adapt kernel topology preference from LNL onwards. I would ideally like to leave MTL (and everything before) with the naming convention it uses today and land any new schema from LNL onwards.

@lgirdwood, I disagree. We should transition to a documented and standard layout, we can provide the fallback for existing kernels without complicated rules and logistics.
The `sof-ace-tplg` directory was a mistake and it has to be corrected while we can with as small pain as possible.

lgirdwood:
> > @ujfalusi lets base solution on what the upstream driver loads today for MTL, we can adapt kernel topology preference from LNL onwards. I would ideally like to leave MTL (and everything before) with the naming convention it uses today and land any new schema from LNL onwards.
> 
> @lgirdwood, I disagree. We should transition to a documented and standard layout, we can provide the fallback for existing kernels without complicated rules and logistics. The `sof-ace-tplg` directory was a mistake and it has to be corrected while we can with as small pain as possible.

This means managing different kernel version and distros through any transition and will be cumbersome. Whilst the current upstream schema may not be perfect for all needs, its often a lot more effort than its worth to back port the improvements. I think its better to leave it as it is (since its working for MTL ?) and make the changes for LNL+

marc-hb:
> The sof-ace-tplg directory was a mistake and it has to be corrected while we can with as small pain as possible.

One of the main disagreements seems to be on the definition of "too late". @ujfalusi please spell out your precise "too late" definition. I guess this should include some kernel version(s)?

I personally don't see any difference between "One platform (MTL) uses the `sof-ace-tplg/` mistake versus "Three platforms use the `sof-ace-tplg/` mistake". For me the line is drawn between 0 and 1 platform, not some place between 1 and 5.

> On upx and RVPs where the BIOS does not contain NHLT blob, the appropriate hda-generic topology needs to be specified as topology file name override.

These are development boards, not "regular" customers but developers so a small hack for these seems acceptable to me if absolutely needed. Also: CAVS25+TGL is NOT officially supported.


> I think its better to leave it as it is (since its working for MTL ?) and make the changes for LNL+

I don't see what's wrong with this either. I admittedly don't understand all the details and nuances.

ujfalusi:
> > > @ujfalusi lets base solution on what the upstream driver loads today for MTL, we can adapt kernel topology preference from LNL onwards. I would ideally like to leave MTL (and everything before) with the naming convention it uses today and land any new schema from LNL onwards.
> > 
> > 
> > @lgirdwood, I disagree. We should transition to a documented and standard layout, we can provide the fallback for existing kernels without complicated rules and logistics. The `sof-ace-tplg` directory was a mistake and it has to be corrected while we can with as small pain as possible.
> 
> This means managing different kernel version and distros through any transition and will be cumbersome. Whilst the current upstream schema may not be perfect for all needs, its often a lot more effort than its worth to back port the improvements. I think its better to leave it as it is (since its working for MTL ?) and make the changes for LNL+

@lgirdwood, no backporting is needed to support released distro kernels for MTL. We have only MTL released currently, we have the opportunity _now_ to change for a layout which is going to work in the future.
I don't want to duplicate the content here, but see my comment on the documentation PR:
https://github.com/thesofproject/sof-docs/pull/482#discussion_r1446979683

We need to define where the topology files are going to be before we can release new platforms. The firmware file location is now settled and it is scaling up to any new IPC4 platform but the topology location is not and if we don't settle on a layout which is going to scale then we are going to have the burden to deal with an increasing mess overtime.
imho

kv2019i:
@lgirdwood wrote:
> This means managing different kernel version and distros through any transition and will be cumbersome. Whilst the current 

@ujfalusi already replied, but let me add a note that I also had this concern, but this is not a problem in the PR as we will always have the "sof-ace-tplg" symlink in sof-bin (this we cannot change anymore). The idea is just to put all Intel IPC4 tplg files in one directory and build our CI/sof-bin automation around that idea. Whether it's called "sof-ipc4-tplg" or "sof-ace-tplg" is less important.

kv2019i:
One scaled down alternative B is to just strip the tplg NHLT blobs from the sof-ace-tplg/sof-hda-generic*tplg files and not converge the folder. This will make "sof-ace-tplg" more future proof (solve https://github.com/thesofproject/sof/issues/8683), but not trying to clean up naming or help releasing the cavs2.5 IPC4 tplg files.

Or alternative C: no action now. Fork "sof-ace-tplg" to "sof-bar-tplg" if/when a platform comes that makes the sof-hda-generic*tplg incompatible and use sof-ace-tplg name until then.

ujfalusi:
I know that it is intuitive for most of the people
that a firmware in `sof` is paired with topology from `sof-tplg`
and
that a firmware in `sof-ipc4` is paired with topology from `sof-ace-tplg` _or_ from `sof-ipc4-tplg` _or_ from `sof-foo-tplg` _or_ from `sof-bar-tplg`
but it is not to me..

A simple rule of `sof-ipc4` is paired with topology from `sof-ipc4-tplg` with a symlink for already released `sof-ace-tplg` (MTL) to it makes it clean. The LNL in kernel can be changed from sof-ace-tplg to sof-ipc4-tplg, it is not yet released.

plbossart:
> A simple rule of `sof-ipc4` is paired with topology from `sof-ipc4-tplg` with a symlink for already released `sof-ace-tplg` (MTL) to it makes it clean. The LNL in kernel can be changed from sof-ace-tplg to sof-ipc4-tplg, it is not yet released.

there's no reason to treat MTL and LNL differently at this point. let's keep the two grouped.

ujfalusi:
> there's no reason to treat MTL and LNL differently at this point. let's keep the two grouped.

We have changed the CAVS2.5 paths, we could do that because CAVS2.5 is not released with IPC4.
LNL is not released either.

To build the firmware, you need different platform target: `lnl` vs `mtl`, right?
To build the topologies you specify different platform also: `lnl` vs `mtl`, right?

What is wrong with saying that IPC4 topology files are under `sof-ipc4-tplg` directory with a caveat that existing distribution kernels are still looking for the MTL topologies under `sof-ace-tplg` so we provide a backwards compatible symlink to make sure that a new sof-bin release is not causing regressions with existing installations relying on that?

ujfalusi:
> > To build the topologies you specify different platform also: `lnl` vs `mtl`, right?
> 
> no we don't. see build-tools.sh

https://github.com/thesofproject/sof/blob/main/tools/topology/topology2/sof-ace-tplg/tplg-targets.cmake#L19
VS
https://github.com/thesofproject/sof/blob/main/tools/topology/topology2/sof-ace-tplg/tplg-targets.cmake#L96

```
"cavs-sdw\;sof-mtl-rt711-4ch\;PLATFORM=mtl,NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,\
DMIC0_ID=2,DMIC1_ID=3,NUM_HDMIS=0,PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-mtl-rt711-4ch.bin"
```
VS
```
"cavs-sdw\;sof-lnl-rt711-4ch\;PLATFORM=lnl,NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,\
DMIC0_ID=2,DMIC1_ID=3,NUM_HDMIS=0,PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-lnl-rt711-4ch.bin"
```

plbossart:
> > > To build the topologies you specify different platform also: `lnl` vs `mtl`, right?
> > 
> > 
> > no we don't. see build-tools.sh
> 
> https://github.com/thesofproject/sof/blob/main/tools/topology/topology2/sof-ace-tplg/tplg-targets.cmake#L19 VS https://github.com/thesofproject/sof/blob/main/tools/topology/topology2/sof-ace-tplg/tplg-targets.cmake#L96
> 
> ```
> "cavs-sdw\;sof-mtl-rt711-4ch\;PLATFORM=mtl,NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,\
> DMIC0_ID=2,DMIC1_ID=3,NUM_HDMIS=0,PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-mtl-rt711-4ch.bin"
> ```
> 
> VS
> 
> ```
> "cavs-sdw\;sof-lnl-rt711-4ch\;PLATFORM=lnl,NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,\
> DMIC0_ID=2,DMIC1_ID=3,NUM_HDMIS=0,PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-lnl-rt711-4ch.bin"
> ```

all topologies will be built.

ujfalusi:
Changes since v1:
- the target is changed from `sof-ipc4-tplg` to `production`
- build_tools.sh updated to create the target directory which can be copied as it is (including symlink)
- the NHLT included hda-generic topology names changed:
 - sof-cavs25-nhlt-hda-generic-2/4ch -> sof-hda-generic-cavs25-2/4ch
 - sof-ace1-nhlt-hda-generic-2/4ch -> sof-hda-generic-ace1-2/4ch

plbossart:
The changes look ok, but I am not Cmake expert, and I don't quite get how CI is going to deal with this change across all devices.

ujfalusi:
Changes since v2:
- Change the target path to `tools/build_tools/topology/topology2/production/target`
- _move_ the tplg files from tools/build_tools/topology/topology2/production/` to `tools/build_tools/topology/topology2/production/target/sof-ipc4-tplg/` instead of copying them (to reduce duplication)

ujfalusi:
> The changes look ok, but I am not Cmake expert, and I don't quite get how CI is going to deal with this change across all devices.

It almost looks good on the storage disk in web view but for some reason the symlink is not transferred there, the target is OK with all the tplg files in place:
```
[ICO]		Name			Last modified	Size	Description

[PARENTDIR]	Parent Directory	 	- 	 
[DIR]		sof-ipc4-tplg/		2024-01-10 13:37 	- 	 
```
@marc-hb, do you have idea why the symlink is missing?

ujfalusi:
@marc-hb, @aiChaoSONG, @mengdonglin, can you try to plan a solution for CI to use  the new location of the built tplg files and to prepare the DUTs?

ujfalusi:
Changes since v3:
- Rebased on main to fix the conflict

plbossart (approved):
Conditional approval, contingent on successful CI tests on MTL/LNL

mengdonglin:
Hi @ujfalusi @lgirdwood  @plbossart @marc-hb,  @keqiaozhang is going to see how to update CI to support this change.

mengdonglin:
@ujfalusi Do we still need a kernel PR to change the default topology path for cavs2.5 platforms and LNL?

ujfalusi:
> @ujfalusi Do we still need a kernel PR to change the default topology path for cavs2.5 platforms and LNL?

CAVS2.5 is already changed to use `sof-ipc4-tplg`
I will change LNL after the CI change, when we have on DUTs:
```
sof-ipc4-tplg
sof-ace-tplg -> sof-ipc4-tplg
```
This is to avoid temporary regression and to not need to sync PRs.

keqiaozhang:
@ujfalusi , CI only need to change the tplg path in software configuration. If both kernel and sof PRs are ready, I think to avoid potential risks, we can ask help from Bard to merge these PRs tomorrow morning (PRC time), I will do the configuration changes accordingly.

ujfalusi:
> @ujfalusi , CI only need to change the tplg path in software configuration. If both kernel and sof PRs are ready, I think to avoid potential risks, we can ask help from Bard to merge these PRs tomorrow morning (PRC time), I will do the configuration changes accordingly.

@keqiaozhang, you also need to override the topology filename for devices where the NHLT is coming from the topology file (like upx-i11 boards afaik).
topologies w/o NHLT (generic):
```
sof-hda-generic.tplg
sof-hda-generic-2ch.tplg
sof-hda-generic-4ch.tplg
```
With CAVS2.5 NHLT:
```
sof-hda-generic-cavs25-2ch.tplg
sof-hda-generic-cavs25-4ch.tplg
```
With ACE1/2 NHLT:
```
sof-hda-generic-ace1-2ch.tplg
sof-hda-generic-ace1-4ch.tplg
```
Wherever CI sets the `sof_use_tplg_nhlt` parameter, you also need to select the appropriate hda-generic topology which includes NHLT.

keqiaozhang:
> you also need to override the topology filename for devices where the NHLT is coming from the topology file.

Sure, I will also change the tplg file name in hardware configuration for HDA platforms.

ujfalusi:
The kernel PR to change the topology path: https://github.com/thesofproject/linux/pull/4779
The sof-docs PR to clarify the ipc4 path: https://github.com/thesofproject/sof-docs/pull/483

scripts/build-tools.sh
```diff
@@ -146,12 +146,20 @@ main()
                 exit 0
         fi
 
-        # Keep 'topologies' first because it's the noisiest.
-        for util in topologies tests; do
-                if eval '$DO_BUILD_'$util; then
-                        make_tool $util
-                fi
-        done
+        if eval $DO_BUILD_topologies; then
```
marc-hb, line 149: This is suspicious, does this really work? I would expect this new code to perform like this:

```
eval $DO_BUILD_topologies;
eval _topologies;

=> `_topologies` command not found, return `false` always.
```

I would have expected the original quotes to be preserved: 
```suggestion
        if eval '$DO_BUILD_'topologies; then
```

While `eval` may stretch `shellcheck` a bit, it's always, always worth asking `shellcheck` what it thinks.

marc-hb: Sorry, my bad. I shouldn't review after a certain hour.

The `eval` is not required anymore, just drop it. It's confusing and confused me.

```suggestion
        if ${DO_BUILD_topologies}; then
```
Also run `shellcheck -x` - always.

scripts/build-tools.sh
```diff
@@ -146,12 +146,20 @@ main()
...
+
+                TPLG2_RELEASE_PATH="$BUILD_TOOLS_DIR/topology/topology2/production/target/"
+                echo ""; echo "Target directory for topology2: $TPLG2_RELEASE_PATH"
+                mkdir -p "$TPLG2_RELEASE_PATH/sof-ipc4-tplg"
+                mv $BUILD_TOOLS_DIR/topology/topology2/production/*.tplg $TPLG2_RELEASE_PATH/sof-ipc4-tplg/
+
+                ln -rsf $TPLG2_RELEASE_PATH/sof-ipc4-tplg $TPLG2_RELEASE_PATH/sof-ace-tplg
+        fi
+
+        if eval $DO_BUILD_tests; then
```
marc-hb, line 160:
```suggestion
        if ${DO_BUILD_tests}; then
```

marc-hb (changes requested):
Drop the eval, test and then dismiss my review. Didn't have time yet to really review the rest or answer questions. Tomorrow if not merged yet.

scripts/build-tools.sh
```diff
@@ -146,12 +146,20 @@ main()
                 exit 0
         fi
 
-        # Keep 'topologies' first because it's the noisiest.
```
marc-hb, line 149: > `# Keep 'topologies' first because it's the noisiest.`

Why drop that comment?

scripts/build-tools.sh
```diff
@@ -146,12 +146,20 @@ main()
...
-        for util in topologies tests; do
-                if eval '$DO_BUILD_'$util; then
-                        make_tool $util
-                fi
-        done
+        if eval $DO_BUILD_topologies; then
+                make_tool topologies
+
+                TPLG2_RELEASE_PATH="$BUILD_TOOLS_DIR/topology/topology2/production/target/"
```
marc-hb, line 152:
```suggestion
                local TPLG2_RELEASE_PATH="$BUILD_TOOLS_DIR/topology/topology2/production/target/"
```

tools/topology/topology2/production/CMakeLists.txt
```diff
@@ -22,6 +25,6 @@ foreach(tplg ${TPLGS})
 	  "${CMAKE_CURRENT_SOURCE_DIR}/../${input}" "${output}"
 	  "${CMAKE_CURRENT_SOURCE_DIR}/../" "${defines}")
 
-	add_custom_target(topology2_avs_${output} DEPENDS ${output}.tplg)
-	add_dependencies(topology2_cavs topology2_avs_${output})
+	add_custom_target(topology2_${output} DEPENDS ${output}.tplg)
+	add_dependencies(topology2 topology2_${output})
```
marc-hb, line 29:
```suggestion
	add_dependencies(topology2_prod topology2_${output})
```

scripts/build-tools.sh
```diff
@@ -146,12 +146,20 @@ main()
...
+                echo ""; echo "Target directory for topology2: $TPLG2_RELEASE_PATH"
+                mkdir -p "$TPLG2_RELEASE_PATH/sof-ipc4-tplg"
+                mv $BUILD_TOOLS_DIR/topology/topology2/production/*.tplg $TPLG2_RELEASE_PATH/sof-ipc4-tplg/
```
marc-hb, line 155: Finally noticing this, sorry for the belated review.

```suggestion
                mv $BUILD_TOOLS_DIR/topology/topology2/production/*.tplg $TPLG2_RELEASE_PATH/sof-ipc4-tplg/
```

Nope.

Never move, delete or rename the files that CMake/ninja just built. They belong to CMake/ninja; `$BUILD_TOOLS_DIR` is EXCLUSIVE property, don't trespass with some outsider script unknown to CMake.  This will cause at least these problems:
- spurious rebuilds
- invoking ninja directly behaves differently from invoking this script. Debugging layers of indirections is painful enough, in no circumstances should this behave differently.
- other, unknown and unexpected CMake issues because no one ever does that ever pulls the rug under CMake/ninja like this.

These files can be at most copied but the best is really to structure the source directories to match what you want the build directories to be. In other words: move and rename the sources to match.

I hate git renames more than the average but this one case where they must be performed.

If you really, really, really can't move sources then look at CMake's "install" feature. Spoiler alert: you don't want that.

marc-hb (changes requested):
Late review sorry. Moving files inside `$BUILD_TOOLS_DIR/` behind CMake's back is a non-starter.

scripts/build-tools.sh
marc-hb: > I hate git renames more than the average but this one case where they must be performed.

That does not mean moving **everything**. Moving only the CMakeLists.txt file and changing the `include()` should work. Or tricks like `CMAKE_CURRENT_BINARY_DIR/../better_dir/` might work too. As long as it does not mess with the build directory _behind CMake's back_ then it's worth a try.

On the other hand, renaming everything would be more consistent with your overarching and admirable quest for... consistency and simplification.

keqiaozhang:
I just verified that there is no risk in applying TPLG changes in CI, we just need to update the TPLG name and path accordingly in the hardware/software configurations.
IPC4 platforms: https://sof-ci.ostc.intel.com/#/result/planresultdetail/37085
IPC3 platforms: https://sof-ci.ostc.intel.com/#/result/planresultdetail/37086

@ujfalusi please help to address the comments from Marc, I will merge these PRs next Monday.

ujfalusi:
Changes since v4:
- Do not touch `build-tools.sh`, do everything within CMake domain
- the target path is changed to `tools/build_tools/topology/topology2/target/` 
- the target directory is created and populated by commands  in add_custom_target(topology2 ...)
- the target directory contains copy of the production and development _topology_ files only
- the target directory contains the `sof-ace-tplg -> sof-ipc4-tplg` symlink needed for ACE1
- The production target in the CMakeList.txt files named precisely as suggested by @marc-hb 
- The build directories (development, production) is retained, no files removed, moved.

**Note:** The target path is changed from previous version!!

tools/topology/topology2/CMakeLists.txt
```diff
@@ -3,7 +3,15 @@
...
-add_custom_target(topologies2)
+add_custom_target(topologies2 ALL
+    # Create directory structure to be provided for deployment
+    COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/target/sof-ipc4-tplg
+    COMMAND ${CMAKE_COMMAND} -E create_symlink sof-ipc4-tplg ${CMAKE_CURRENT_BINARY_DIR}/target/sof-ace-tplg
+    COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/target/development
+    # copy the topology files only to target
+    COMMAND ${CMAKE_COMMAND} -E copy_if_different production/*.tplg ${CMAKE_CURRENT_BINARY_DIR}/target/sof-ipc4-tplg/
+    COMMAND ${CMAKE_COMMAND} -E copy_if_different development/*.tplg ${CMAKE_CURRENT_BINARY_DIR}/target/development/
+)
```
kv2019i, line 14: Seems to do the trick. Given the baseline has no install logic at all, this seems like a huge value-add. 

It would seem the dependencies are also set for topologies2 target, so good in that sense as well.

kv2019i (approved):

tools/topology/topology2/production/CMakeLists.txt
```diff
@@ -22,6 +25,6 @@ foreach(tplg ${TPLGS})
 	  "${CMAKE_CURRENT_SOURCE_DIR}/../${input}" "${output}"
 	  "${CMAKE_CURRENT_SOURCE_DIR}/../" "${defines}")
 
-	add_custom_target(topology2_avs_${output} DEPENDS ${output}.tplg)
-	add_dependencies(topology2_cavs topology2_avs_${output})
+	add_custom_target(topology2_${output} DEPENDS ${output}.tplg)
+	add_dependencies(topology2_prod topology2_${output})
```
marc-hb, line 29: Nit
```suggestion
	add_dependencies(topology2_prod topology2_prod_${output})
```
See the difference in `ninja -C tools/build_tools/ help`

(I know, should have said earlier)

tools/topology/topology2/CMakeLists.txt
marc-hb: So this is basically a DIY "install". I suspect there is a slightly more "native" and performant way to do this but this is very simple and easy to maintain later so plenty good enough to get started.

marc-hb (approved):

marc-hb (commented):
This PR did way too many things in a single commit: renaming AND splitting files AND making HDA code changes at the same time. When you add the usual branch divergence on top, this is proving near impossible to backport (#8858).

git is really bad at tracking file renames, so a best practice in complex cases like this one is to keep renames "pure" in one commit and then make changes in the next commit - EVEN if that first commit does not work temporarily. Just flag it as "don't use"; `git bisect` has a `git bisect skip` command for this.

PR #8720, opened 2024-01-10
Add new CONFIG_DCBLOCK_HIFI and move generic XCHAL_HAVE logic to common.h

marc-hb:
1. The `if XCHAL_HAVE_HIFI4` macro logic is duplicated across many source files. Starting with dcblock.h, make it generic and move it to common.h

2. Add the ability to override the max available HIFI level using Kconfig

### DIFF
--- a/src/audio/dcblock/Kconfig
+++ b/src/audio/dcblock/Kconfig
@@ -6,3 +6,10 @@ config COMP_DCBLOCK
 	help
 	  Select for DC Blocking Filter component. This component filters out
 	  the DC offset which often originates from a microphone's output.
+
+
+config DCBLOCK_HIFI
+	int "DC Blocking Filter HIFI level (-1 = max)"
+	default -1
+	help
+	  0: no HIFI, -1 (default): max available
--- a/src/include/sof/common.h
+++ b/src/include/sof/common.h
@@ -177,6 +177,46 @@
 #define IS_ENABLED_STEP_3(ignore, value, ...) (!!(value))
 #endif
 
+#define SOF_AUTO_HIFI -1
+
+/* Generic, no HIFI. This could be anything but it is zero for
+ * user-friendliness. Unfortunately, an accidentally undefined
+ * CONFIG_${comp}_HIFI will silently be converted to 0 and will not be
+ * reported. To catch such issues, simply set this to -2 temporarily.
+ */
+#define SOF_NO_HIFI 0
+
+/* True if CONFIG_${comp}_HIFI has been explicitly configured to value 'v'.
+ *
+ * We unfortunately cannot used `defined()` here:
+ *  - https://bugs.webkit.org/show_bug.cgi?id=167643
+ *  - https://lists.llvm.org/pipermail/cfe-commits/Week-of-Mon-20160118/147239.html
+ */
+#define SOF_CONFIG_HIFI(v, comp) ((v) == (CONFIG_ ## comp ## _HIFI))
+
+/* True if either explicitly specified in Kconfig as above, _or_
+ *   it is the max HIFI available in the HAL.
+ */
+#define SOF_USE_HIFI(v, comp) SOF_CONFIG_HIFI(v, comp) || \
+	(SOF_CONFIG_HIFI(SOF_AUTO_HIFI, comp) && SOF_MAX_XCHAL_HIFI == (v))
+
+#ifndef __XCC__ // Cadence toolchains
+#  define SOF_MAX_XCHAL_HIFI SOF_NO_HIFI
+#else
+#  include <xtensa/config/core-isa.h>
+// Maybe we could make this fully generic (and less readable!) using
+// IS_ENABLED() above.
+#  if XCHAL_HAVE_HIFI5
+#    define SOF_MAX_XCHAL_HIFI 5
+#  elif XCHAL_HAVE_HIFI4
+#    define SOF_MAX_XCHAL_HIFI 4
+#  elif XCHAL_HAVE_HIFI3
+#    define SOF_MAX_XCHAL_HIFI 3
+#  else
+#    define SOF_MAX_XCHAL_HIFI SOF_NO_HIFI
+#  endif
+#endif
+
 #ifndef __GLIBC_USE
 #define __GLIBC_USE(x) 0
 #endif
--- a/src/audio/dcblock/dcblock.h
+++ b/src/audio/dcblock/dcblock.h
@@ -15,20 +15,6 @@
 #include <module/module/base.h>
 #include <module/module/interface.h>
 
-/* __XCC__ is both for xt_xcc and xt_clang */
-#if defined(__XCC__)
-# include <xtensa/config/core-isa.h>
-# if XCHAL_HAVE_HIFI4
-#  define DCBLOCK_HIFI4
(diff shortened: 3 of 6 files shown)


src/audio/dcblock/Kconfig
```diff
@@ -6,3 +6,10 @@ config COMP_DCBLOCK
 	help
 	  Select for DC Blocking Filter component. This component filters out
 	  the DC offset which often originates from a microphone's output.
+
+
+config DCBLOCK_HIFI
+	int "DC Blocking Filter HIFI level (-1 = max)"
+	default -1
```
lgirdwood, line 13: can we depend on ARCH_XTENSA and make this a Kconfig choice (where each choice would depend on ARCH_XTENSA_HIFI_n set by top level PLATFORM Kconfig) ? 

Btw, I think Peter Mitsis will have the HiFi flavour landing in Zephyr soon as part of the HiFi context save/restore work. i.e. we can reuse RTOS supplied Kconfig data.

Whatever methodology we use, it should also scale to x86 and ARM SIMD.

src/include/sof/common.h
```diff
@@ -177,6 +177,46 @@
 #define IS_ENABLED_STEP_3(ignore, value, ...) (!!(value))
 #endif
 
+#define SOF_AUTO_HIFI -1
```
lgirdwood, line 180: Will need to namespace these as ```SOF_ARCH_XTENSA_feature```

src/audio/dcblock/Kconfig
btian1: where is ARCH_XTENSA and ARCH_XTENSA_HIFI_n definition? is it already exist or added later?

src/include/sof/common.h
```diff
@@ -177,6 +177,46 @@
...
+ *   it is the max HIFI available in the HAL.
+ */
+#define SOF_USE_HIFI(v, comp) SOF_CONFIG_HIFI(v, comp) || \
+	(SOF_CONFIG_HIFI(SOF_AUTO_HIFI, comp) && SOF_MAX_XCHAL_HIFI == (v))
+
+#ifndef __XCC__ // Cadence toolchains
+#  define SOF_MAX_XCHAL_HIFI SOF_NO_HIFI
+#else
+#  include <xtensa/config/core-isa.h>
+// Maybe we could make this fully generic (and less readable!) using
```
btian1, line 207: I am not sure this is best way to go, with this, module HIFI config still depend on core-isa, this is centuralized. 

The logic is more complex and with more dependency, like __XCC__, more complex SOF_USE_HIFI macro definition, if expand this to other dsp platform(non-xtensa and X86), this part will need change again and hard to make changes.

src/audio/dcblock/Kconfig
btian1: also, there is no HIFI option for this module, no HIFI3 and HIFI4 option.

src/include/sof/common.h
marc-hb: > module HIFI config still depend on core-isa, this is centuralized

You haven't summarized why these are bad.

We can duplicate `core-isa.h` in various places and I'm looking at options but `core-isa.h` will always remain the origin, true reference for HIFI facts.

> with more dependency, like `__XCC__`

`__XCC__` is the only reliable way to detect Cadence toolchains and it's what Zephyr uses.
https://github.com/thesofproject/sof/pull/8682#discussion_r1446760651

Which other dependency? 


> The logic is more complex [...] and hard to make changes.

If you find this logic complex then I'm afraid nothing can be done for you.

src/audio/dcblock/Kconfig
marc-hb: > also, there is no HIFI option for this module, no HIFI3 and HIFI4 option.

In this prototype you enter '3' or '4' or any version you like. @lgirdwood already requested a `Kconfig choice`

Please spend more time reading and make sure you understand better before commenting.

peter-mitsis: Just to confirm, I am indeed working on adding HiFi DSP support in Zephyr so that its registers will be saved/restored on relevant context switches.

andyross: Nice!   Can't resist a preemptive kibitz:  When you do, please do it as a generic hook to save coprocessor state, as Xtensa has some existing FPU context logic that should share the same backend.  And FWIW: lazy-evaluated switching based on the coprocessor enabled exception is really only a nice-to-have in SOF, as the overwhelming majority of processing is done on core-pinned threads that don't preempt each other; storing them all on a switch (but not ISR!) is totally OK.   Similarly will want a hook on interrupt entry to make sure the kernel can't touch the registers incorrectly.  But note eventually this will want userspace support, and syscalls probably **will** want SIMD access, so the syscall entry code will need a spot to do the lazy save.

andyross: Oh, and remember that the thread struct is uncached, which makes it a poor spot on intel_adsp for saving big blocks like this, you'll want to pass it through arch_xtensa_cached_ptr() and to do your flush/invalidations manually if you want it there, or otherwise find a cached spot for it.

andyross: I'll shut up now.

marc-hb:
> can we depend on ARCH_XTENSA and make this a Kconfig choice (where each choice would depend on ARCH_XTENSA_HIFI_n set by top level PLATFORM Kconfig) ?

I did this in a separate PR because the code is totally different: no macros and everything in Kconfig:
- #8720 

It's not as tedious and verbose as I was afraid it would be BUT:
- It is less generic and does require significantly more Kconfig code
- I think 8720 relies on some undocumented Kconfig feature. If that feature can't be trusted then the KConfig code will become about twice more verbose
- It depends on Zephyr Kconfig changes https://github.com/zephyrproject-rtos/zephyr/pull/67513 (just a prototype for now)
- XTOS will need an equivalent of 67513. I can't write and test this XTOS equivalent because all Intel XTOS has been removed from this branch, see #8720 for the longer story.

src/audio/dcblock/Kconfig
lgirdwood: All - @peter-mitsis will need to define a SOC level Kconfig to indicate the HIFI level on xtensa. i.e. this will say hifi3, hifi4, hifi5 etc as the different hifi ISA add additional register that must be saved/restored.
This same Kconfig can be reused here to indicate the highest HIFI SIMD level for each platform/SOC i.e. some menuconfig options will depend on this Zephyr SOC Kconfig.

marc-hb: > All - @peter-mitsis will need to define a SOC level Kconfig to indicate the HIFI level on xtensa.

My attempt to do this:
- https://github.com/zephyrproject-rtos/zephyr/pull/67513

Just prototype and draft but a functional one so please review.

btian1: Thank you all,  then this open will be resolved soon.
let's wait for Peter's PR to zephyr and got merged.

@peter-mitsis , could you share PR link here once you submit here? or add me into review list for the zephyr PR？

marc-hb: #8787 is similar to this but with a Kconfig `choice` instead.

marc-hb:
Alternative #8787 has been merged.

PR #8722, opened 2024-01-10
Google aec sink src dp 

marcinszkudlinski:
this is a replacement of  https://github.com/thesofproject/sof/pull/8469
all review changes from it have been included

This PR modifies google AEC shimm to use sink/src interface and to be run as DP module in 10ms period

This is upstream of AEC changes from 007 branch and:
 - Legacy implementation kept as google_rtc_audio_processing_ll.c, compiled when DP processing is disabled
 - changes required after merging https://github.com/thesofproject/sof/pull/8685 change

please ignore checkpatch warnings in google_rtc_audio_processing_ll.c, this file is - and should stay - 1:1 copy of legacy google_rtc_audio_processing.c.


andyross (commented):
No objection in principle to merging this first, as long as there's a sincere commitment from Intel to consume and validate the other version too.  If the only reason for merging this is to cut another "stable drop" branch that is cut off from SOF main, that seems like a bad trade.  We need to agree on a single implementation and that takes work from both sides.

As far as code: it would be cleaner to avoid all the code motion (i.e. leave it all in one file, avoid the IPC3 removals), which really isn't needed to support the unified AEC and will only confuse the git history.  (Edit: but that's not a -1, it's no harder to squash a big change than a small one, the only impact is to the pour soul trying to figure out what happened from the commit log)

app/boards/intel_adsp_ace15_mtpm.conf
```diff
@@ -92,3 +92,6 @@ CONFIG_DEBUG_COREDUMP_MEMORY_DUMP_MIN=y
 CONFIG_PROBE=y
 CONFIG_PROBE_DMA_MAX=2
 CONFIG_LOG_TIMESTAMP_64BIT=y
+
+CONFIG_COMP_GOOGLE_RTC_AUDIO_PROCESSING=y
+CONFIG_COMP_STUBS=y
```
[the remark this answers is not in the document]
marc-hb: This is a recurring problem. If your editor can't be be fixed, maybe your git client can somehow show this or even catch this before git push?

andyross (changes requested):
>  If the only reason for merging this is to cut another "stable drop" branch that is cut off from SOF main, that seems like a bad trade. We need to agree on a single implementation and that takes work from both sides.

And... it turns out from sideband communication that this is exactly what is happening. 
 I think this has to be a -1 from me.    I've been following the rules in #8571, addressing review comments, splitting out changes, chasing down workarounds and external/framework issues instead of just working around them inline.  It's neither fair nor good engineering practice for Intel to ram this through just to be able to cut a release without having to work with me.

Can you guys please actually try the code in #8571 first before merging this without review?  It works as-is, though I need to push up a respin without the separately-submitted changes.  I'll get this out this afternoon.  If it doesn't work, then report problems and let's work together on fixing them.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -561,121 +575,172 @@ static int google_rtc_audio_processing_reset(struct processing_module *mod)
...
+	}
 
-			src += audio_stream_get_channels(mic_stream);
-			dst += audio_stream_get_channels(out_stream);
-		}
-		num_frames_remaining -= n;
-		src = audio_stream_wrap(mic_stream, src);
-		dst = audio_stream_wrap(out_stream, dst);
+#if CONFIG_COMP_GOOGLE_RTC_USE_32_BIT_FLOAT_API
+	GoogleRtcAudioProcessingAnalyzeRender_float32(
```
RanderWang, line 680: we use different proc function for different format and select proc function based on module config
.e.g.
````c
const struct drc_proc_fnmap drc_proc_fnmap[] = {
/* { SOURCE_FORMAT , PROCESSING FUNCTION } */
#if CONFIG_FORMAT_S16LE
	{ SOF_IPC_FRAME_S16_LE, drc_s16_default },
#endif /* CONFIG_FORMAT_S16LE */

#if CONFIG_FORMAT_S24LE
	{ SOF_IPC_FRAME_S24_4LE, drc_s24_default },
#endif /* CONFIG_FORMAT_S24LE */

#if CONFIG_FORMAT_S32LE
	{ SOF_IPC_FRAME_S32_LE, drc_s32_default },
#endif /* CONFIG_FORMAT_S32LE */
};
````

**Comment by marcinszkudlinski** [comment]
why _LL ?

because **IPC4 != DP_scheduling**

this PR does:
 - LL version for IPC3 and IPC4
 - DP version (IPC4 only - no choice here)

**Comment by mwasko** [comment]
> No objection in principle to merging this first, as long as there's a sincere commitment from Intel to consume and validate the other version too. 

@andyross the purpose of this change is to enable AEC module with DP Scheduler using IPC4. Since Intel has already transition to IPC4 then we have no plan to enable and validate this feature with IPC3. However, there are no objections to follow up with IPC3 support as a next step, but the validation and enabling will have to be done on your end.

> As far as code: it would be cleaner to avoid all the code motion (i.e. leave it all in one file, avoid the IPC3 removals), which really isn't needed to support the unified AEC and will only confuse the git history. (Edit: but that's not a -1, it's no harder to squash a big change than a small one, the only impact is to the pour soul trying to figure out what happened from the commit log)

I talked with @marcinszkudlinski  and "Remove IPC3 support" is unfortunate naming and should be squashed, because the intention here is not to remove any IPC3 functionality, but rather limit the new feature to IPC4 (at least in this PR).

File: src/audio/google/google_rtc_audio_processing.c (hunk 2 above)
Comment (marcinszkudlinski) [reply src/audio/google/google_rtc_audio_processing.c:680]: this is a separate option - just use google's API 16 or 32. It is independent of out CONFIG_FORMAT_xxxx flags.
In fact, AEC 16 and 32 bit API differ a bit more than just number of bits - they also accept diffrent set of parameters: 

int GoogleRtcAudioProcessingAnalyzeRender_float32(
    GoogleRtcAudioProcessingState *const state, const float *const *src);
int GoogleRtcAudioProcessingAnalyzeRender_int16(
    GoogleRtcAudioProcessingState *const state, const int16_t *const src);

const float *const *src vs const float *const src  - a pointer to a table of pointers to channels or just a pointer to first channel

**Comment by andyross** [comment]
> Since Intel has already transition to IPC4 then we have no plan to enable and validate this feature with IPC3

This isn't an Intel project and SOF isn't an Intel deliverable though.  I understand you don't want to support IPC3 anymore (I don't either, FWIW), but we still have devices shipping with it and they need to run with current code still.  Which is why #8571 exists: it has all the features this PR does, plus IPC3 support, plus dynamic stream format handling, it's smaller, faster, and it does it all in a single source file without having to copy and paste all the future changes between this and a "legacy" version.  If there are features you want to see there, just add review comments and let's work together on fixing them.

I absolutely understand there are business worries about what software which entity is willing to support.   And that may have to be dealt with out of band.  I'll remove the -1 here since it's not really code-related and replace it with a proper review trying to point out the collisions and how it compares with the competing PR.

File: src/audio/google/google_rtc_audio_processing.c (hunk 2 above)
Comment (andyross) [reply src/audio/google/google_rtc_audio_processing.c:680]: FWIW, this is basically the architecture in #8571 (though it's selected with a branch and isn't table driven), there are a bunch of function pointers for doing the optimized sink/source de/interleave+convert+copy for each format variant, without any cut/paste of algorithm code.  See https://github.com/andyross/sof/blob/mt8195-aec/src/audio/google/google_rtc_audio_processing.c#L112

**Comment by andyross** [comment]
> Can we split this file into google_rtc_audio_processing_ipc3.c and google_rtc_audio_processing_ipc4.c like smart_amp ?

To repeat, there's really just not that much divergence.  The only major code block that changes is that the control handling has a rather different API in IPC4 (this was pre-existing, and absolutely can be cleaned up a ton).  Beyond that,  there's some sink/source validation code that obviously doesn't apply to IPC3 and a 1-line workaround for the broken reference channel count (which is a protocol bug really, will submit a proposed fix once we get things merged).

Basically see here and scan for the use of IPC_MAJOR: https://github.com/andyross/sof/blob/mt8195-aec/src/audio/google/google_rtc_audio_processing.c

**Comment by marcinszkudlinski** [comment]
...so here's a merge IPC3+IPC4
IPC4 works with DP scheduling only

**Comment by mwasko** [comment]
> This isn't an Intel project and SOF isn't an Intel deliverable though. I understand you don't want to support IPC3 anymore (I don't either, FWIW), but we still have devices shipping with it and they need to run with current code still. 

@andyross you are completely right and because of that you cannot expect Intel to enable and validate all possible configurations that are supported in SOF. This change introduce DP scheduling support for IPC4 because this is the configuration we can validate and guarantee the quality of it.

**Comment by andyross** [comment]
> you cannot expect Intel to enable and validate all possible configurations that are supported in SOF.

Of course not.  I just want you to review and approve a PR.  If it doesn't work on your supported platform, provide comments and test results or whatever and we'll work together to fix it.  I'll handle the other stuff.

File: src/audio/google/google_rtc_audio_processing.c [hunk 3]
```diff
@@ -53,9 +53,7 @@ DECLARE_TR_CTX(google_rtc_audio_processing_tr, SOF_UUID(google_rtc_audio_process
 			   LOG_LEVEL_INFO);
 
 struct google_rtc_audio_processing_comp_data {
-#if CONFIG_IPC_MAJOR_4
```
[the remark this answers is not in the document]
btian1: agree, I did not touch goole RTC part for ipc3/ipc4 split, because last quarter, a lot of patches are changing this module, if you need me to do this, we can do it later, this module indeed need a separation.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -639,6 +730,167 @@ static int google_rtc_audio_processing_reset(struct processing_module *mod)
 	return 0;
 }
 
+#if CONFIG_IPC_MAJOR_4
+static inline int16_t convert_google_aec_format_to_int16(BUF_TYPE data)
+{
+#if CONFIG_COMP_GOOGLE_RTC_USE_32_BIT_FLOAT_API
+#if XCHAL_HAVE_HIFI3
+	const xtfloat ratio = 2 << 14;
```
btian1, line 738: 2 << 14 is strange, 1 << 15 means this fraction is Q15 format for int16 type.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -639,6 +730,167 @@ static int google_rtc_audio_processing_reset(struct processing_module *mod)
...
+#else /* CONFIG_COMP_GOOGLE_RTC_USE_32_BIT_FLOAT_API */
+	return data;
+#endif
+}
+
+static inline BUF_TYPE convert_int16_to_google_aec_format(int16_t data)
+{
+#if CONFIG_COMP_GOOGLE_RTC_USE_32_BIT_FLOAT_API
+#if XCHAL_HAVE_HIFI3
+	const xtfloat ratio = 2 << 14;
```
btian1, line 759: here same.

mwasko:
> > you cannot expect Intel to enable and validate all possible configurations that are supported in SOF.
> 
> Of course not. I just want you to review and approve a PR. If it doesn't work on your supported platform, provide comments and test results or whatever and we'll work together to fix it. I'll handle the other stuff.

That is the plan. I just wanted to clarify that we do not have Intel setups for AEC DP testing with IPC3 protocol so this configuration has to be verified by Google or MediaTek. What we can do is cross validate your PR with IPC4 configuration to confirm that it does not introduce any regression (in progress).

src/audio/google/google_rtc_audio_processing_mock.c
```diff
@@ -140,34 +145,64 @@ int GoogleRtcAudioProcessingReconfigure(GoogleRtcAudioProcessingState *const sta
...
+			int16_t mic_save = mic[n + (chan * state->num_frames)];
 
-		ref += state->num_aec_reference_channels;
-		out += state->num_output_channels;
-		mic += state->num_capture_channels;
+			if (chan < state->num_aec_reference_channels)
+				dest[n + (chan * state->num_frames)] =
+						mic_save + ref[n + (chan * state->num_frames)];
+			else
+				dest[n + (chan * state->num_frames)] = mic_save;
```
lyakh, line 202: I wouldn't rely on the compiler and use a local variable for `n + (chan * state->num_frames)` explicitly

mwasko (approved):

kv2019i (approved):
@andyross It seems both this and https://github.com/thesofproject/sof/pull/8571 are progressing and it's really a question in which order we merge. Looking at both PRs again, your PR is doing more things and has comments open, while this PR is taking a safer path and only touching IPC4 specific parts. But please check the comments on 8571 and consider can we proceed with merging this first.

lkoenig (changes requested):

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -390,20 +381,26 @@ static int google_rtc_audio_processing_init(struct processing_module *mod)
...
-		 &base_cfg->base_cfg_ext.pin_formats[size], size);
-	memcpy_s(&output_fmt, size,
-		 &base_cfg->base_cfg_ext.pin_formats[size * GOOGLE_RTC_NUM_INPUT_PINS], size);
+	cd->config.output_fmt = mod->priv.cfg.input_pins[SOF_AEC_DMIC_QUEUE_ID].audio_fmt;
+	cd->config.reference_fmt = mod->priv.cfg.input_pins[SOF_AEC_FEEDBACK_QUEUE_ID].audio_fmt;
 
-	cd->config.reference_fmt = reference_fmt.audio_fmt;
-	cd->config.output_fmt = output_fmt.audio_fmt;
+	cd->num_aec_reference_channels = cd->config.reference_fmt.channels_count;
+	cd->num_capture_channels = cd->config.output_fmt.channels_count;
```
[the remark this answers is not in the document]
lgirdwood: @lkoenig I'm good with a Google AEC Kconfig option to define the max channels here as I assume different Google AEC configurations would be used on different HW (with different max channel Kconfigs). The runtime channels count would come from topology definition (which host drv and FW would additionally verify).
@marcinszkudlinski ok for you ?

marcinszkudlinski:
COMP_GOOGLE_RTC_AUDIO_PROCESSING_CHANNEL_MAX added and tested
The lib I got in testing indeed is running twice faster when processing 2 channels (as expected)
@andyross  A tip:
if you're **certain** that the library will finish processing faster than 10ms (i.e. in 6), you can lower its declared deadline by adding a line into google_rtc_audio_processing_prepare:

```
#if CONFIG_IPC_MAJOR_4
dev->period =  6 * LL_TIMER_PERIOD_US;
#endif
```

the  AEC processing will still be called in 10ms period (this is determined by buffer sizes), but as the declared deadline is shorter, FW will keep shorter "safe buffers" and the overall processing latency will be reduced.

a default deadline is the longest allowed deadline == period calculated from in/out buffer sizes

src/audio/google/google_rtc_audio_processing.c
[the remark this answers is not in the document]
marcinszkudlinski: @lgirdwood ok, already upstreamed here

src/audio/google/google_rtc_audio_processing_mock.c
marcinszkudlinski: you mean a local var for performance? This is test mock, performance does not matter at all.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -639,6 +733,180 @@ static int google_rtc_audio_processing_reset(struct processing_module *mod)
...
+			ref = (void *)ref_buf_start;
+	}
+
+#if CONFIG_COMP_GOOGLE_RTC_USE_32_BIT_FLOAT_API
+	GoogleRtcAudioProcessingAnalyzeRender_float32(cd->state,
+						      (const float **)
+							cd->aec_reference_buffer_ptrs);
+#else
+	GoogleRtcAudioProcessingAnalyzeRender_int16(cd->state,
+						    (const int16_t *)cd->aec_reference_buffer);
```
lkoenig, line 843: looks like `aec_reference_buffer` is not correctly populated properly for `int16` processing.

I would say: either we keep both option working (which complexify testing) or we choose only int16 or float32 (in that case my preference would go for the float32 one.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -639,6 +733,180 @@ static int google_rtc_audio_processing_reset(struct processing_module *mod)
...
+		 * number of incoming channels may be <cd->num_capture_channels
+		 */
+		src += cd->config.output_fmt.channels_count;
+		if ((void *)src >= (void *)src_buf_end)
+			src = (void *)src_buf_start;
+	}
+
+	source_release_data(src_stream, num_of_bytes_to_process);
+
+	/* call the library, use same in/out buffers */
```
lkoenig, line 870: Same comment about interleaving of samples.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -639,6 +733,180 @@ static int google_rtc_audio_processing_reset(struct processing_module *mod)
...
+
+	/* same number of bytes to process for output stream as for mic stream */
+	ret = sink_get_buffer(dst_stream, num_of_bytes_to_process, (void **)&dst,
+			      (void **)&dst_buf_start, &dst_buf_size);
+	assert(!ret);
+	dst_buf_end = dst_buf_start + dst_buf_size;
+
+	/* process all channels in output stream */
+	for (int i = 0; i < cd->num_frames; i++) {
+		for (channel = 0; channel < cd->config.output_fmt.channels_count; channel++)
```
lkoenig, line 889: interleaving again.

src/audio/google/google_rtc_audio_processing_mock.c
```diff
@@ -140,34 +145,64 @@ int GoogleRtcAudioProcessingReconfigure(GoogleRtcAudioProcessingState *const sta
...
-		for (io = 0; io < state->num_output_channels; io++) {
-			out[io] = sat_int16(
-				(im < state->num_capture_channels ? (int32_t)mic[im++] : 0) +
-				(ir < state->num_aec_reference_channels ? (int32_t)ref[ir++] : 0));
-		}
+	int n, chan;
+
+	for (chan = 0; chan < state->num_output_channels; chan++) {
+		for (n = 0; n < state->num_frames; ++n) {
+			int16_t mic_save = mic[n + (chan * state->num_frames)];
```
lkoenig, line 196: int16 is interleaved !
`n * cd->num_capture_channels + chan`

src/audio/google/google_rtc_audio_processing_mock.c
```diff
@@ -140,34 +145,64 @@ int GoogleRtcAudioProcessingReconfigure(GoogleRtcAudioProcessingState *const sta
...
+
+	for (channel = 0; channel < state->num_aec_reference_channels; channel++) {
+		memcpy_s(&state->aec_reference[channel * state->num_frames], buffer_size,
+			 data[channel], buffer_size);
+	}
+
+	return 0;
+}
+#else /* CONFIG_COMP_GOOGLE_RTC_USE_32_BIT_FLOAT_API */
 int GoogleRtcAudioProcessingProcessCapture_int16(GoogleRtcAudioProcessingState *const state,
```
lkoenig, line 186: `int16` processing is *_interleaved_* !

lkoenig (changes requested):

src/audio/google/google_rtc_audio_processing_mock.c
lyakh: I think we should distance a bit from "mock performance doesn't matter." I understand that it doesn't matter when running it, but I think we should still keep the code "reasonably optimal" when this doesn't sacrifice clarity and doesn't require too much work. Code gets copied around and can well end up in production algorithms. So I'd try to avoid code that clearly and easily can be replaced with a faster version.

marc-hb:
The build failures with the zephyr main branch are expected. This other one isn't:

https://github.com/thesofproject/sof/actions/runs/7628619323/job/20780113474?pr=8722

```
/zep_workspace/sof/src/audio/google/google_rtc_audio_processing.c
/zep_workspace/sof/src/audio/google/google_rtc_audio_processing.c: In function 'google_rtc_audio_processing_init':
/zep_workspace/sof/src/audio/google/google_rtc_audio_processing.c:370:13: error: unused variable 'channel' [-Werror=unused-variable]
  370 |         int channel;
      |             ^~~~~~~
```

marcinszkudlinski:
@andyross @lkoenig @marc-hb all request addressed, pls proceed

abonislawski (approved):

mwasko:
@lkoenig, @andyross your change requests has been addressed, please update your review so we can move forward with this PR.

lkoenig:
Sorry I was on vacation. Will have a look tomorrow.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -453,7 +449,36 @@ static int google_rtc_audio_processing_init(struct processing_module *mod)
...
+	if (!cd->process_buffer) {
+		comp_err(dev, "Allocating process_buffer failure");
+		ret = -EINVAL;
+		goto fail;
+	}
+	bzero(cd->process_buffer, buf_size);
+	buf_size = cd->num_frames * sizeof(cd->aec_reference_buffer[0]) *
+			cd->num_aec_reference_channels;
+	comp_dbg(dev, "Allocating aec_reference_buffer of size %u", buf_size);
+	cd->aec_reference_buffer = rballoc(0, SOF_MEM_CAPS_RAM, buf_size);
```
lkoenig, line 466: nit picky here: Would be nice to have the computation ` cd->num_frames * sizeof(cd->aec_reference_buffer[0]) * cd->num_aec_reference_channels` instead of `buf_size` as the buffer size might be different for render and for capture.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -639,6 +734,190 @@ static int google_rtc_audio_processing_reset(struct processing_module *mod)
...
+		if (ret)
+			return ret;
+	}
+
+	src_stream = sources[cd->raw_microphone_source];
+	ref_stream = sources[cd->aec_reference_source];
+	dst_stream = sinks[0];
+
+	num_of_bytes_to_process = cd->num_frames * source_get_frame_bytes(ref_stream);
+	ret = source_get_data(ref_stream, num_of_bytes_to_process, (const void **)&ref,
```
lkoenig, line 813: Is it always guaranteed that the input buffers are int16 ? because otherwise the pointer arithmetic to move to the next frame won't work if let's say the input is int32_t.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -639,6 +734,190 @@ static int google_rtc_audio_processing_reset(struct processing_module *mod)
...
+	assert(!ret);
+	src_buf_end = src_buf_start + src_buf_size;
+
+	buffer_offset = 0;
+	for (int i = 0; i < cd->num_frames; i++) {
+		for (channel = 0; channel < cd->num_capture_channels; channel++)
+#if CONFIG_COMP_GOOGLE_RTC_USE_32_BIT_FLOAT_API
+			cd->process_buffer_ptrs[channel][i] = convert_int16_to_float(src[channel]);
+#else /* CONFIG_COMP_GOOGLE_RTC_USE_32_BIT_FLOAT_API */
+		cd->process_buffer[buffer_offset++] = src[channel];
```
lkoenig, line 865: Nit: indentation

lkoenig:
I think we are getting there. Any update ?

lkoenig:
Sorry I clicked the wrong button.

lkoenig:
Sorry I clicked the wrong button.

src/audio/google/google_rtc_audio_processing.c
```diff
@@ -639,6 +733,180 @@ static int google_rtc_audio_processing_reset(struct processing_module *mod)
...
+			return ret;
+	}
+
+	src_stream = sources[cd->raw_microphone_source];
+	ref_stream = sources[cd->aec_reference_source];
+	dst_stream = sinks[0];
+
+	num_of_bytes_to_process = cd->num_frames * source_get_frame_bytes(ref_stream);
+	ret = source_get_data(ref_stream, num_of_bytes_to_process, (const void **)&ref,
+			      (const void **)&ref_buf_start, &ref_buf_size);
```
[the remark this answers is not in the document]
marcinszkudlinski: DP should NOT process more than its period

why
 - DP start in async to LL, so it may happen that between DP becomes ready (10ms in each of buffers),  there's another LL tick and one of the pipelines will provide additional data portions. 

TICK
 - Core 0 LL finishes processing, AEC has 10ms in both buffers => ready to running
 TICK 
 - core 0 starts another LL cycle, core1 starts DP (and is late a bit- i.e. was processing IPC, does not matter why)
 - **core0 finishes some processing before DP starts**, there's additional 1ms data in one of inputs (or even on both)

but "additional 1ms" belong to a next cycle. Process it in next cycle

src/audio/google/google_rtc_audio_processing.c
marcinszkudlinski: CONFIG_FORMAT_S32LE is not implemented

src/audio/google/google_rtc_audio_processing.c
lkoenig: Sorry I must have been tired when I did the review. It looks good as it is.

src/audio/google/google_rtc_audio_processing.c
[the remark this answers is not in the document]
lkoenig: Thank you for the clarification.

marcinszkudlinski:
I was on medical leave, please proceed

lkoenig (approved):

andyross (commented):
Sorry, thought I was already +1.  Fine with me, though not looking forward to having to piecewise-rework the other PR.

lgirdwood (approved):

lgirdwood:
Re run the CI build check. CI went down on Monday so manually restart.

app/boards/intel_adsp_ace15_mtpm.conf
marc-hb, line 97: ?

From `src/audio/Kconfig`
"CONFIG_STUBS: This should only be used in testing environments like fuzzers or CI."

Good example of #9386

cc: @cujomalainey

marc-hb: Now being reverted in:
- #9410

marc-hb: #9410 results and discussion seem to show a test seems to depend on this!

PR #8723, opened 2024-01-10
audio: dai-zephyr: Fix unused variable warning

dbaluta:
Guard sink_list in order to avoid the following variable unused warning:

$ sof/src/audio/dai-zephyr.c: In function 'dai_dma_cb': $ sof_dir/sof/src/audio/dai-zephyr.c:259:35: warning: unused variable 'sink_list' [-Wunused-variable]
  259 |                 struct list_item *sink_list;
      |                                   ^~~~~~~~~

### DIFF
--- a/src/audio/dai-zephyr.c
+++ b/src/audio/dai-zephyr.c
@@ -255,8 +255,6 @@ dai_dma_cb(struct dai_data *dd, struct comp_dev *dev, uint32_t bytes,
 		ret = dma_buffer_copy_to(dd->local_buffer, dd->dma_buffer,
 					 dd->process, bytes);
 	} else {
-		struct list_item *sink_list;
-
 		audio_stream_invalidate(&dd->dma_buffer->stream, bytes);
 		/*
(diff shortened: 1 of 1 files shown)


ranj063 (approved):

iuliana-prodan (approved):

marc-hb (commented):
Why not just use a better C99 declaration and move it right after the already existing `#if CONFIG_IPC_MAJOR_4`? We're in 2024 now...

C89 declarations are ugly and the `#ifdef` makes it even uglier.

In case you're wondering: we have C99 declarations in plenty other places and they never caused any issue with any toolchain.

This topic was discussed in
- https://github.com/orgs/thesofproject/discussions/7395 (2021!)

kv2019i:
@dbaluta can you check, fix is good, but moving the declaration down avoids adding an ifdef...

dbaluta:
> @dbaluta can you check, fix is good, but moving the declaration down avoids adding an ifdef...

done.

PR #8725, opened 2024-01-11
Remove sof logging ctx

btian1:


lgirdwood (commented):
@btian1 for changes like this is best to show before and after in the logs as part of your commit message so everyone can follow.

btian1:
@dbaluta @iuliana-prodan , could you help check from IPC3 side? as this change only have impact on IPC3 and I don't have a way to check.

dbaluta:
Will try to have a look this week.

btian1:
> Will try to have a look this week.

Any progress? @dbaluta , this DRAFT patch try to remove each module's ctx dependency for further one logging interface.
And since ctx is only used by IPC3 logging, request help from you here for double confirm, if it is acceptable:
1. All DECLARE_TR_CTX will be removed.
2. Zephyr and non-zephyr logging will have same logging interface.

dbaluta (approved):

dbaluta:
@btian1 tested this on imx8mp. Logging works fine after your patch.

lgirdwood (approved):

lgirdwood:
@btian1 draft ?

dbaluta:
If i remember correctly this is only for testing purposes.

PR #8727, opened 2024-01-11
rimage: loadable_modules: Set correct image type

softwarecki:
The image header contains a bitfield with flags specifying a image type. This commit sets correct image type depending on whether firmware or a loadable module is being signed.

### DIFF
--- a/tools/rimage/src/manifest.c
+++ b/tools/rimage/src/manifest.c
@@ -1380,7 +1380,8 @@ int man_write_fw_ace_v1_5(struct image *image)
 	m->desc.header.build_version = image->fw_ver_build;
 
 	m->desc.header.feature_mask = 0x2; // -> should be feature mask - to fix
-	m->desc.header.fw_image_flags = 0x2; // -> should be feature mask - to fix
+	m->desc.header.fw_image_flags.fields.image_type = image->loadable_module ?
+		SOF_MAN_FW_HDR_IMG_TYPE_LIB : SOF_MAN_FW_HDR_IMG_TYPE_MAIN_FW;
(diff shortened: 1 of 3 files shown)


pjdobrowolski (approved):

tools/rimage/src/manifest.c
```diff
@@ -1380,7 +1380,8 @@ int man_write_fw_ace_v1_5(struct image *image)
 	m->desc.header.build_version = image->fw_ver_build;
 
 	m->desc.header.feature_mask = 0x2; // -> should be feature mask - to fix
-	m->desc.header.fw_image_flags = 0x2; // -> should be feature mask - to fix
+	m->desc.header.fw_image_flags.fields.image_type = image->loadable_module ?
+		SOF_MAN_FW_HDR_IMG_TYPE_LIB : SOF_MAN_FW_HDR_IMG_TYPE_MAIN_FW;
```
kv2019i, line 1384: So we have set the flag incorrectly before (used 0x2 for main image)?

kv2019i (approved):
Good catch @softwarecki  , let's see how the CI fairs, this should fail at boot if at all.

tools/rimage/src/manifest.c
softwarecki: @kv2019i  0x02 was correct value for a main firmware image

lyakh (approved):

lgirdwood:
@softwarecki can you check CI. Thanks !

lgirdwood (approved):

tools/rimage/src/manifest.c
kv2019i: @softwarecki Aa, right, the old 0x02 was for the raw value.

PR #8728, opened 2024-01-11
Tools: Topology2: Add other rates support for DMIC and HDA

singalsu:


tools/topology/topology2/development/tplg-targets.cmake
```diff
@@ -81,6 +81,68 @@ PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-hda-efx-generic-4ch.bin,\
...
+# With 4ch 96 kHz DMIC0 and HDA
+"sof-hda-generic\;sof-hda-generic-2ch-96k\;\
+HDA_CONFIG=mix,DEEPBUFFER_FW_DMA_MS=false,HDA_ANALOG_CAPTURE_RATE=96000,HDA_ANALOG_PLAYBACK_RATE=96000,\
+PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-hda-generic-2ch-96k.bin,\
+NUM_DMICS=2,PDM1_MIC_A_ENABLE=0,PDM1_MIC_B_ENABLE=0,DMIC0_RATE=96000"
+
+# With 96 kHz HDA
+"sof-hda-generic\;sof-hda-generic-96k\;\
+HDA_CONFIG=mix,DEEPBUFFER_FW_DMA_MS=false,HDA_ANALOG_CAPTURE_RATE=96000,HDA_ANALOG_PLAYBACK_RATE=96000,\
+PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-hda-generic-2ch-96k.bin,\
```
singalsu, line 117: No NHLT for HDA only.

lgirdwood (commented):
LGTM, @RanderWang @bardliao @ujfalusi fyi - starting point for 96kHz topologies.

tools/topology/topology2/platform/intel/dmic-default.conf
```diff
@@ -36,4 +37,7 @@ Define {
 	DMIC_CORE_ID			0
 	PASSTHROUGH			"false"
 	DMIC_PIPELINE_PRIORITY		0
+	DMIC0_RATE                      48000
+	DMIC1_RATE                      16000
```
btian1, line 41: DMIC0/1 rate are defined by spec or pipeline?

tools/topology/topology2/platform/intel/dmic-generic.conf
```diff
@@ -39,7 +40,7 @@ Object.Dai.DMIC [
                dai_index               1
                driver_version          $DMIC_DRIVER_VERSION
                io_clk                  $DMIC_IO_CLK
-               sample_rate             16000
+	       sample_rate             $DMIC1_RATE
```
btian1, line 43: your commit message only mentioned DMIC0 can adjust with 8-96k,  Does not mention DMIC1, and DMIC1 also set from external ternal, what's DMIC1's rate setting range?

Also, I think line 43 changes should go to the next patch or a separate patch, what do you think?

tools/topology/topology2/cavs-mixin-mixout-hda.conf
```diff
@@ -7,6 +7,8 @@ Define {
 	DEEP_BUFFER_PIPELINE_SRC	'mixin.15.1'
 	DEEP_BUFFER_PIPELINE_SINK	'mixout.2.1'
 	DEEP_BUFFER_PCM_NAME		'Deepbuffer HDA Analog'
+        HDA_ANALOG_CAPTURE_RATE         48000
+        HDA_ANALOG_PLAYBACK_RATE        48000
```
btian1, line 11: just ask a question here: ACE platform should also have HDA config platform, do we need change ACE platform as well?

tools/topology/topology2/development/tplg-targets.cmake
```diff
@@ -81,6 +81,68 @@ PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-hda-efx-generic-4ch.bin,\
 EFX_FIR_PARAMS=passthrough,EFX_IIR_PARAMS=passthrough,\
 EFX_DRC_COMPONENT=multiband,EFX_MBDRC_PARAMS=passthrough"
 
+# With 16 kHz DMIC1
+"sof-hda-generic\;sof-hda-generic-4ch-48k-16k\;\
+HDA_CONFIG=mix,DEEPBUFFER_FW_DMA_MS=100,\
```
ujfalusi, line 86: `DEEPBUFFER_FW_DMA_MS=100`  can be dropped, it is the default

tools/topology/topology2/development/tplg-targets.cmake
```diff
@@ -81,6 +81,68 @@ PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-hda-efx-generic-4ch.bin,\
 EFX_FIR_PARAMS=passthrough,EFX_IIR_PARAMS=passthrough,\
 EFX_DRC_COMPONENT=multiband,EFX_MBDRC_PARAMS=passthrough"
 
+# With 16 kHz DMIC1
+"sof-hda-generic\;sof-hda-generic-4ch-48k-16k\;\
```
ujfalusi, line 85: I think it might consider namespacing the development topologies for different DSP architectures.
If I'm not mistaken, these are for CAVS2.5 and will not work on ACE1/2?

I have a PR (https://github.com/thesofproject/sof/pull/8712#) to merge the production topologies and it introduces namespacing for hda-generic topologies:
topologies w/o NHLT (generic):
```
sof-hda-generic.tplg
sof-hda-generic-2ch.tplg
sof-hda-generic-4ch.tplg
```
With CAVS2.5 NHLT:
```
sof-hda-generic-cavs25-2ch.tplg
sof-hda-generic-cavs25-4ch.tplg
```
With ACE1/2 NHLT:
```
sof-hda-generic-ace1-2ch.tplg
sof-hda-generic-ace1-4ch.tplg
```
Probably introduce similar naming right now or later when #8712 is merged?

tools/topology/topology2/cavs-mixin-mixout-hda.conf
singalsu: I think we use this same .conf for ACE too. The naming is confusing.

tools/topology/topology2/development/tplg-targets.cmake
singalsu: Yep, your naming change looks very good to me. I've been wondering how I could differentiate TGL and MTL tplg builds. I will do this update.

tools/topology/topology2/platform/intel/dmic-default.conf
singalsu: I first thought that we would review DMIC with #8688 but maybe it's better that I close it and keep editing this.

So, these are the default values. We can specify other rates in build cmake if need. Does this look OK to you?

tools/topology/topology2/platform/intel/dmic-generic.conf
singalsu: I need to correct the commit text, both DMIC0 and DMIC1 can support 8 - 96 kHz.

I don't know what would be best, where DMIC1_RATE is changed is not critical.  Since the macro is already defined, what's the issue to  not keep it here?

tools/topology/topology2/cavs-mixin-mixout-hda.conf
btian1: agree, if also use this, we may need change name and line 9 to remove HDA config

singalsu: I hope the filenames change for ACE could be other PR. There's quite many cavs-*.conf files.

Maybe the line numbers no more match. What's the issue with HDA config?

ujfalusi: if the topology does not have NHLT blob then the filename can be the same.
In any case the development topologies should not be copied to the production directory (I just did that on my setup, so don't follow my lead).
If these worked on ACE then there is no NHLT or the NHLT is not used from topology, it can be ripped out?

My topology change PR is not touching development only creates a production target and merges the avs-tplg and sof-ace-tplg under it.

singalsu: I don't think there's 96 kHz DMIC blobs in BIOS so it's safest to use add the blob to topology.

ujfalusi: @singalsu, my point is: you cannot use the same NHLT blob on CAVS2.5 and ACE1/2, so if the NHLT is taken from the topology file then you must have different files. This is nothing new, I just pointed out that it looks to me that shis change is only supporting CAVS2.5 and the generic once needs namespacing.

singalsu:
Note: This PR needs https://github.com/alsa-project/alsa-utils/pull/250

RanderWang:
> There was a mistake in previous PR version with sof-mtl-max98360a-rt5682 PCM IDs. DMIC0 should be 99 (it was 100) and I decided to set DMIC1 PCM ID to 100. The PCM name in all cases is "DMIC Raw2nd".
> 
> With sof-mtl-sdw-cs42l42-l0-max98363-l2 and HDA generic 2ch or 4ch the added DMIC1 capture PCM ID is default 11. It's the same as with wake-on-voice, but ultrasound and WoV should be mutually exclusive for DMIC1.
> 
> @jsarha @plbossart @RanderWang Do these names and IDs loook OK?

Name is good for me. It is for pass-through pipeline, we can rename it for real case.

plbossart:
for SoundWire, best to follow the list I came up with a couple of years ago...

[PCMDeviceList.txt](https://github.com/thesofproject/sof/files/14021187/PCMDeviceList.txt)

Device 11 is "just" for 16khz, for KPB/WOV it's a different device really...

singalsu:
> for SoundWire, best to follow the list I came up with a couple of years ago...
> 
> [PCMDeviceList.txt](https://github.com/thesofproject/sof/files/14021187/PCMDeviceList.txt)
> 
> Device 11 is "just" for 16khz, for KPB/WOV it's a different device really...

Thanks, I'll change it to PCM22 for 96 kHz capture. That's very useful plan for other applications too.

lgirdwood (commented):
@ujfalusi @singalsu is the ace/cavs NHLT blob naming now resolved ?

fredoh9:
SOFCI TEST

kv2019i (commented):
Code looks good, CI issues need address, e.g. https://github.com/thesofproject/sof/actions/runs/8017346835/job/21901028754?pr=8728

tools/topology/topology2/platform/intel/dmic-generic.conf
```diff
@@ -198,16 +211,112 @@ IncludeByKey.PASSTHROUGH {
...
+						}
+						{
+							out_rate		$DMIC0_RATE
+							out_channels		4
+							out_bit_depth		32
+							out_valid_bit_depth	32
+							out_ch_cfg	$CHANNEL_CONFIG_3_POINT_1
+							out_ch_map	$CHANNEL_MAP_3_POINT_1
+						}
+					]
```
plbossart, line 245: this part adds a lot more than the simple addition of a non-default 48kHz.

That should be moved in a separate patch with explanations on why we didn't need this earlier?

tools/topology/topology2/platform/intel/dmic-generic.conf
```diff
@@ -198,16 +211,112 @@ IncludeByKey.PASSTHROUGH {
...
```
plbossart, line 319: same, separate patch?

tools/topology/topology2/platform/intel/dmic-generic.conf
```diff
@@ -40,7 +40,7 @@ Object.Dai.DMIC [
                dai_index               1
                driver_version          $DMIC_DRIVER_VERSION
                io_clk                  $DMIC_IO_CLK
-               sample_rate             16000
+	       sample_rate             $DMIC1_RATE
```
plbossart, line 43: commit message: 

"
The sample rate of DMIC1 DAI can be set to 8 - 96 kHz with
macro DMIC1_RATE.
"

I thought DMIC1_RATE had to be lower than DMIC0_RATE? Are the two independent?

tools/topology/topology2/development/tplg-targets.cmake
```diff
@@ -81,6 +81,61 @@ PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-hda-efx-generic-4ch.bin,\
...
+"sof-hda-generic\;sof-hda-generic-cavs25-4ch-48k-16k\;HDA_CONFIG=mix,\
+PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-hda-generic-cavs25-4ch-48k-16k.bin,\
+NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,DMIC1_ENABLE=passthrough,DMIC1_RATE=16000"
+
+"sof-hda-generic\;sof-hda-generic-ace1-4ch-48k-16k\;HDA_CONFIG=mix,PLATFORM=mtl,\
+PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-hda-generic-ace1-4ch-48k-16k.bin,\
+NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,DMIC1_ENABLE=passthrough,DMIC1_RATE=16000"
+
+# SDW + DMIC + HDMI, with 16 kHz DMIC1
+"cavs-sdw\;sof-mtl-sdw-cs42l42-l0-max98363-l2-48k-16k\;PLATFORM=mtl,\
```
plbossart, line 148: the naming isn't great, there's nothing that tells us that the 48k-16k is related to dmics. same comment for all SoundWire platforms.

tools/topology/topology2/development/tplg-targets.cmake
```diff
@@ -84,17 +84,71 @@ EFX_DRC_COMPONENT=multiband,EFX_MBDRC_PARAMS=passthrough"
...
+DMIC1_RATE=16000,DMIC1_PCM_ID=24"
 
 "sof-hda-generic\;sof-hda-generic-ace1-4ch-48k-16k\;HDA_CONFIG=mix,PLATFORM=mtl,\
 PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-hda-generic-ace1-4ch-48k-16k.bin,\
-NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,DMIC1_ENABLE=passthrough,DMIC1_RATE=16000"
+NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,DMIC1_ENABLE=passthrough,\
+DMIC1_RATE=16000,DMIC1_PCM_ID=24"
+
+# With 4ch 96 kHz DMIC1 and HDA
+"sof-hda-generic\;sof-hda-generic-cavs25-4ch-48k-96k\;HDA_CONFIG=mix,\
```
plbossart, line 96: it's the same here, now 96kHz applies to both HDAudio and DMIC...

And it'll be an absolute horror if we start adding heaphone specific streams, which can have a different rate to both the HDaudio speaker and DMIC paths...

tools/topology/topology2/platform/intel/dmic-generic.conf
singalsu: Yep, splitting this to two.

tools/topology/topology2/platform/intel/dmic-generic.conf
singalsu: yep

tools/topology/topology2/platform/intel/dmic-generic.conf
singalsu: Yes, the DMIC0 and DMIC1 FIR are identical, they share the same CIC. There can be some restrictions for rates from DMIC bus clock constraints (min, max Hz, duty cycle) and IO clock (higher clock usually better). If an impossible rates combination is attempted the topology build will fail.

tools/topology/topology2/development/tplg-targets.cmake
singalsu: This is for development folder to show an example with multiple rates options, what the platform is capable to. The production topologies for ultrasound should have nicer simple names when the multi-rate architecture in decided.

tools/topology/topology2/development/tplg-targets.cmake
singalsu: I could add "dmic" but on the other hand we have already 2ch/4ch without indication that it's for microphones count. I'm not sure what to do with this issue and next.

tools/topology/topology2/development/tplg-targets.cmake
singalsu: Done!

tools/topology/topology2/platform/intel/dmic-generic.conf
plbossart: my question was : can DMIC0 run at 16kHz and DMIC1 at 96kHz? In other words is the selection arbitrary? IIRC there was a second stage FIR, is this enabled for both DMIC0 and DMIC1?

tools/topology/topology2/development/tplg-targets.cmake
plbossart: you could use e.g. 4ch-48k-16k, i.e. channel, DMIC0 rate, DMIC1 rate in that order. That's different enough from the convention used for SoundWire links.

tools/topology/topology2/development/tplg-targets.cmake
```diff
@@ -84,17 +84,71 @@ EFX_DRC_COMPONENT=multiband,EFX_MBDRC_PARAMS=passthrough"
...
 PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-hda-generic-cavs25-4ch-48k-16k.bin,\
-NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,DMIC1_ENABLE=passthrough,DMIC1_RATE=16000"
+NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,DMIC1_ENABLE=passthrough,\
+DMIC1_RATE=16000,DMIC1_PCM_ID=24"
 
 "sof-hda-generic\;sof-hda-generic-ace1-4ch-48k-16k\;HDA_CONFIG=mix,PLATFORM=mtl,\
 PREPROCESS_PLUGINS=nhlt,NHLT_BIN=nhlt-sof-hda-generic-ace1-4ch-48k-16k.bin,\
-NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,DMIC1_ENABLE=passthrough,DMIC1_RATE=16000"
+NUM_DMICS=4,PDM1_MIC_A_ENABLE=1,PDM1_MIC_B_ENABLE=1,DMIC1_ENABLE=passthrough,\
+DMIC1_RATE=16000,DMIC1_PCM_ID=24"
```
plbossart, line 93: Why not "PCM11  capture	DMIC16kHz" ?

tools/topology/topology2/platform/intel/dmic-generic.conf
singalsu: Yes, DMIC0 as 16 kHz and DMIC1 as 96 kHz is OK. The FIR decimators are in parallel and share same CIC, so they can be set independently to any rate as long as there is a suitable common CIC output rate. There's no option to place DMIC1 FIR after DMIC0 FIR even if it might have been a good idea for efficiency.

plbossart: Ah ok, thanks for educating me. I thought the two parts were serial with a 1/N downsampler in the second DMIC1.

tools/topology/topology2/development/tplg-targets.cmake
singalsu: I changed it because I understood from your earlier comment that 11 for WoV. This 24 would be "PCM24  capture Voice  (Communication with a human)" but I'll change this back to 11.

tools/topology/topology2/development/tplg-targets.cmake
lgirdwood: Development folder is fine - we have a chance to circle back for v2.10 and make improvements to overall topology suggested by @plbossart

kv2019i (approved):
My concerns addressed. CI tests still need to pass, but e.g. quickbuild fails should have nothing to do with tplg changes, so that is likely a false fail.

plbossart:
@singalsu sorry if I am perceived as the painful reviewer, but I have to ask a background technical question:

In a recent thread (https://github.com/thesofproject/sof/issues/8834), @perexg mentioned that "The limitation (on some hardware) is that the speaker is usually connected to one DAC with a fixed rate (48000Hz) and the headphones to other (more rates up to 192kHz are available). "

So my question is "do we know for sure if the 96kHz rate is supported on the 'analog' endpoint on all HDAudio hardware platforms"? If not, then we would need some sort of hardware discovery at the kernel level to filter/supersede topology definitions.

I don't mind if we add 96kHz for 'development' topologies, I just wonder if this will remain an experimental capability for the foreseeable future or if we can also add it to mainstream HDaudio topologies at some point.

lgirdwood:
@wszypelt @lrudyX good to merge ? I dont think these files are tested by internal CI today ?

singalsu:
> @singalsu sorry if I am perceived as the painful reviewer, but I have to ask a background technical question:
> 
> In a recent thread (#8834), @perexg mentioned that "The limitation (on some hardware) is that the speaker is usually connected to one DAC with a fixed rate (48000Hz) and the headphones to other (more rates up to 192kHz are available). "
> 
> So my question is "do we know for sure if the 96kHz rate is supported on the 'analog' endpoint on all HDAudio hardware platforms"? If not, then we would need some sort of hardware discovery at the kernel level to filter/supersede topology definitions.
> 
> I don't mind if we add 96kHz for 'development' topologies, I just wonder if this will remain an experimental capability for the foreseeable future or if we can also add it to mainstream HDaudio topologies at some point.

I don't know. I can't find from this specification that 96 kHz would be a mandatory feature of all HD audio codecs. Does someome know? https://www.intel.com/content/dam/www/public/us/en/documents/product-specifications/high-definition-audio-specification.pdf

But I'd assume that the the driver would error if 96 kHz would be attempted on a non-96 kHz capable system. Also this patch enables non-48 kHz only for test topologies in development folder. They are not part of FW release.

wszypelt:
@lgirdwood @singalsu Unfortunately, Intel Internal CI shows an error in the tools step: make tools cmake

singalsu:
> @lgirdwood @singalsu Unfortunately, Intel Internal CI shows an error in the tools step: make tools cmake

Is the alsa-utils up to date there? It needs to have recent changes, commit 906a56f9ff136211e53db201383a47d9b0ee445d.

singalsu:
> Just to clarify, is the topology going to support both 48+96kHz, or just 96kHz? The latter case is problematic, I have no real issues with the former.

The changes common to all (HDA) topology definitions cause no issues to build topologies for 48 kHz like we today build. The rate is just set explicitly in formats in addition to word lengths to enable changing the rate for PCM0 and in DMIC case the PCMs for it (DMIC0 and DMIC1). For simplicity there is no SRC placeholder and macros for multiple rates. So when build for 96 kHz is defined, I'm afraid this is " just 96kHz". I was asked to make this example for HDA and DMIC. I didn't want to make anything complicated, just something that allowed colleagues at Intel to further define this for customer.

Can you draw a picture of how these ultrasound development enabler topologies should look like ("both 48+96kHz").

plbossart:
> > Just to clarify, is the topology going to support both 48+96kHz, or just 96kHz? The latter case is problematic, I have no real issues with the former.
> 
> The changes common to all (HDA) topology definitions cause no issues to build topologies for 48 kHz like we today build. The rate is just set explicitly in formats in addition to word lengths to enable changing the rate for PCM0 and in DMIC case the PCMs for it (DMIC0 and DMIC1). For simplicity there is no SRC placeholder and macros for multiple rates. So when build for 96 kHz is defined, I'm afraid this is " just 96kHz". I was asked to make this example for HDA and DMIC. I didn't want to make anything complicated, just something that allowed colleagues at Intel to further define this for customer.
> 
> Can you draw a picture of how these ultrasound development enabler topologies should look like ("both 48+96kHz").

Now that I think of it, we can only support multiple rates if we remove the mixer. When we have a mixer in place we have to pick ONE sample rate, otherwise it's a mess if different streams with different rates get mixed.

But assuming we remove the deep buffer for playback (which no one uses), then we can make the entire pipeline support multiple of 48kHz, as done for SSP/BT. If the hardware doesn't support a rate, that would be rejected by the codec DAI so there's a clear error sent by to apps.

singalsu:
> > > Just to clarify, is the topology going to support both 48+96kHz, or just 96kHz? The latter case is problematic, I have no real issues with the former.
> > 
> > 
> > The changes common to all (HDA) topology definitions cause no issues to build topologies for 48 kHz like we today build. The rate is just set explicitly in formats in addition to word lengths to enable changing the rate for PCM0 and in DMIC case the PCMs for it (DMIC0 and DMIC1). For simplicity there is no SRC placeholder and macros for multiple rates. So when build for 96 kHz is defined, I'm afraid this is " just 96kHz". I was asked to make this example for HDA and DMIC. I didn't want to make anything complicated, just something that allowed colleagues at Intel to further define this for customer.
> > Can you draw a picture of how these ultrasound development enabler topologies should look like ("both 48+96kHz").
> 
> Now that I think of it, we can only support multiple rates if we remove the mixer. When we have a mixer in place we have to pick ONE sample rate, otherwise it's a mess if different streams with different rates get mixed.
> 
> But assuming we remove the deep buffer for playback (which no one uses), then we can make the entire pipeline support multiples of 48kHz, as done for SSP/BT. If the hardware doesn't support a rate, that would be rejected by the codec DAI so there's a clear error sent by to apps.

Removing the deep buffer from these 96 kHz topologies builds should be possible, I'll check it and update.

singalsu:
> Removing the deep buffer from these 96 kHz topologies builds should be possible, I'll check it and update.

Actually the deep buffer is already disabled with DEEPBUFFER_FW_DMA_MS=false. I just forgot about it, sorry.

plbossart:
> DEEPBUFFER_FW_DMA_MS=false

Does this actually disable the deep buffer path or just default to the regular buffer size for the deep-buffer path?

that's pretty bad...

singalsu:
> > DEEPBUFFER_FW_DMA_MS=false
> 
> Does this actually disable the deep buffer path or just default to the regular buffer size for the deep-buffer path?
> 
> that's pretty bad...

The first one, everything related to deep buffer is removed. See here:
https://github.com/thesofproject/sof/blob/6f33aee2709ac619c4913a231f40e5b27d4390a5/tools/topology/topology2/cavs-mixin-mixout-hda.conf#L16

plbossart:
> > > DEEPBUFFER_FW_DMA_MS=false
> > 
> > 
> > Does this actually disable the deep buffer path or just default to the regular buffer size for the deep-buffer path?
> > that's pretty bad...
> 
> The first one, everything related to deep buffer is removed. See here:
> 
> https://github.com/thesofproject/sof/blob/6f33aee2709ac619c4913a231f40e5b27d4390a5/tools/topology/topology2/cavs-mixin-mixout-hda.conf#L16

wow. I thought IncludeByKey was using a true/false variable. I didn't know it could be used for integers.  Oh well, you live and learn.

singalsu:
> Ping @singalsu -- this pending ack from @plbossart and CI check is failing.

@kv2019i  Thanks for following this. The fail is because is CI at IPG has not updated their alsa-utils. I've reported it and they are working on it ( @wszypelt ).

I rebased and pushed this again to see if it works now. There's no changes to "code".

lgirdwood (approved):

lgirdwood:
@wszypelt I assume this is good for internal CI as AFAIK none of the changes here are tested by Internal CI ?

wszypelt:
@lgirdwood @singalsu we managed to fix alsa-utils in CI, everything works :)

PR #8730, opened 2024-01-12
META clean up

btian1:


src/include/sof/trace/trace.h
```diff
@@ -21,7 +21,7 @@
 #endif
 #include <sof/common.h>
 #include <rtos/sof.h>
-#include <sof/trace/preproc.h>
+#include <preproc/preproc.h>
```
lgirdwood, line 24: Looks like next steps could be to move trace.h to xtos header too since it will be all xtos specific APIs.

lgirdwood (commented):
LGTM

src/include/sof/trace/trace.h
btian1: trace.h header file both include zephyr and non-zephyr definitons, if want to move to xtos/zephyr directory, we need duplicate it like other header files under xtos/include or zephyr/include.

zephyr/include/rtos/string_macro.h
```diff
@@ -0,0 +1,19 @@
...
+#ifndef __ZEPHYR_RTOS_STRING_MACRO_H__
+#define __ZEPHYR_RTOS_STRING_MACRO_H__
+
+/* treat x as string while forcing x expansion beforehand */
+#define _STRING_QUOTE(x) #x
+#define STRING_QUOTE(x) _STRING_QUOTE(x)
+
+/* concat x and y while forcing x and y expansion beforehand */
+#define _STRING_CONCAT_BASE(x, y) x##y
+#define STRING_CONCAT(x, y) _STRING_CONCAT_BASE(x, y)
```
lyakh, line 17: no. This header should just `#include <zephyr/toolchain.h>` and `#include <zephyr/sys/util_macro.h>` and you should just use `STRINGIFY()` and `CONCAT()` directly in SOF code and define **them** in posix and xtos headers.

lyakh (changes requested):

xtos/include/sof/trace/preproc-private-inc.h
```diff
@@ -0,0 +1,98 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright(c) 2016 Intel Corporation. All rights reserved.
+ *
+ * Author: Michal Jerzy Wierzbicki <michalx.wierzbicki@intel.com>
+ */
+
+/* THIS FILE SHOULD NOT BE INCLUDED DIRECTLY */
```
lyakh, line 8: hm, so now we have two copies of these rather large headers. These are headers and they aren't included simultaneously of course, they don't affect run-time, etc., but this isn't very good. Why are they needed under posix? Something like testbench or mocka?

btian1: yes, testbench and cmocka need it from posix path, I have to add it as well. I guess testbench and cmocka build can't be changed to xtos, since the file inside are not same, @singalsu , do you have comments on this question?

singalsu: I hope we could remove need for xtos headers for testbench, topology parser and all other in SOF/tools. The xtos headers like often come with low-level xtensa stuff dependencies and cause difficulty in writing portable code.

singalsu (approved):
Looks good to me, thanks! Locally all my testbench runs seem to be OK, everything works.

kv2019i (approved):

kv2019i:
Worrying failures in https://sof-ci.01.org/sofpr/PR8730/build2035/devicetest/index.html but hardly caused by this PR.

PR #8731, opened 2024-01-12
dai: add error log for failed path

RanderWang:
The error log is convenient for developer to check failed case in stress test.

We have a bug : https://github.com/thesofproject/sof/issues/8056. The failed log provided limited information.
````
[  766.013925] <err> dai_comp: dai_common_new: comp:1 0x10004 dai_new(): dai_get() failed to create DAI.
[  766.013945] <err> copier: copier_dai_create: comp:1 0x10004 failed to create dai
[  766.013978] <err> copier: copier_init: comp:1 0x10004 unable to create dai
````
It is hard to be reproduced in private environment, so add error log for CI to catch it

### DIFF
--- a/src/lib/dai.c
+++ b/src/lib/dai.c
@@ -208,8 +208,11 @@ struct dai *dai_get(uint32_t type, uint32_t index, uint32_t flags)
 	struct dai *d;
 
 	dev = dai_get_zephyr_device(type, index);
-	if (!dev)
+	if (!dev) {
+		tr_err(&dai_tr, "dai_get: failed to get dai with index %d type %d",
+		       index, type);
 		return NULL;
+	}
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by kv2019i — APPROVED** [review]

**Review by abonislawski — APPROVED** [review]

File: src/lib/dai.c [hunk 1]
```diff
@@ -222,6 +225,8 @@ struct dai *dai_get(uint32_t type, uint32_t index, uint32_t flags)
 	dai_set_device_params(d);
 
 	if (dai_probe(d->dev)) {
+		tr_err(&dai_tr, "dai_get: failed to probe dai with index %d type %d",
+		       index, type);
```
lyakh, line 229: as an option, you could print the returned error code

lgirdwood (approved):

src/lib/dai.c
lgirdwood: Lets add this incrementally, would be good to get weekend tests done with this log.

PR #8737, opened 2024-01-12
telemetry: systick info

tobonex:
This patch introduces telemetry by adding systick info measurements. Also adds basic implementations of ipc's used for by our tests for systick info measurements.
Currently this implementation only handles measurement of LL execution time, as measuring DP tasks may be a bit tricky. 

This patch also requires increasing the size of memory window 2, which is on the Zephyr side. It may pass tests without it, but the results will be incorrect as it may write on memory occupied by Zephyr logs. This memory window size change is being taken care of also. Will link any related Zephyr PR's once they are created. 

I'm a bit uncertain about some parts, so I'll self review with some questions.


src/ipc/ipc4/handler.c
```diff
@@ -927,7 +1020,8 @@ static int ipc4_get_large_config_module_instance(struct ipc4_message_request *ip
 
 	if (ret < 0)
 		return IPC4_FAILURE;
-
+	dcache_invalidate_region((__sparse_force void __sparse_cache *)MAILBOX_HOSTBOX_BASE,
+				 config.extension.r.data_off_size);
```
tobonex, line 1024: config_get now uses payload, so we need to invalidate hostbox cache here. Only vendor config get uses payload for now , so we could move this to the vendor_config_get function

src/debug/telemetry/telemetry.c
```diff
@@ -0,0 +1,137 @@
+// SPDX-License-Identifier: BSD-3-Clause
```
tobonex, line 1: Looking for a good home for this file, and the .h one. I created telemetry folder in debug, but if you think there's a more fittting place let me know.

src/debug/telemetry/telemetry.c
```diff
@@ -0,0 +1,137 @@
...
+#include <zephyr/debug/sparse.h>
+#include <sof/debug/telemetry/telemetry.h>
+#include <sof/audio/module_adapter/module/generic.h>
+#include "adsp_debug_window.h"
+
+/* systic vars */
+int systick_counter;
+int prev_ccount;
+int perf_period_sum_;
+int perf_period_cnt_;
```
tobonex, line 23: If i remember correctly, each core has an instance of an LL scheduler, right? Those variables in original firmware were in CoreServices class, which I'm assuming was there to provide some storage and functionality per core. As those variables here are used in update_telemetry function, if there are more LL schedulers running, can be interruped and vars can have their value changed. I'm guessing in this case providing one set of those per core should fix the problem then?

src/audio/base_fw.c
```diff
@@ -455,16 +461,86 @@ static int basefw_set_fw_config(bool first_block,
...
+	schedulers_info->scheduler_count++;
+	scheduler_props = (struct scheduler_props *)(data + *data_off_size);
+	scheduler_get_task_info_ll(scheduler_props, data_off_size);
+
+	/* ===================== LL_DMA SCHEDULER INFO ============================ */
+	/* Could not make this one work */
+	/* schedulers_info->scheduler_count++;
+	 * scheduler_props = (struct scheduler_props *)(data + *data_off_size);
+	 * scheduler_get_task_info_ll_dma(scheduler_props, data_off_size);
+	 */
```
tobonex, line 497: Did not manage to get any info from LL_DMA scheduler. I was not sure what type of struct scheduler_get_data(SOF_SCHEDULE_LL_DMA) would return. Are there any tasks there anyway?

src/audio/base_fw.c
```diff
@@ -455,16 +461,86 @@ static int basefw_set_fw_config(bool first_block,
...
 
+int schedulers_info_get(uint32_t *data_off_size,
+			char *data,
+			uint32_t core_id)
+{
+	/* TODO
+	 * Core id parameter is not yet used. For now we only get scheduler info from current core
+	 * Other cores info can be added by implementing idc request for this data.
+	 * Do this if Schedulers info get ipc has uses for accurate info per core
+	 */
```
tobonex, line 472: I left this one in a bit of an awkward state. IPC works now, but it gets data only from the core that it on is currently, which is random. It was enough for my needs, though.

src/ipc/ipc4/handler.c
```diff
@@ -914,6 +914,99 @@ static int ipc4_unbind_module_instance(struct ipc4_message_request *ipc4)
...
+		const uint8_t *end_offset = data_in + *data_off_size;
+		size_t produced_data = 0;
+
+		while ((const uint8_t *)input_tl < end_offset) {
+			/* we go to next output tlv with each iteration */
+			uint32_t data_off_size_local;
+			struct sof_tlv *output_tlv = (struct sof_tlv *)(data_out + produced_data);
+
+			if (produced_data + input_tl->max_length > MAILBOX_DSPBOX_SIZE)
+				return IPC4_FAILURE;
```
tobonex, line 944: I'm guessing ipc response finds its way to dspbox in the end, so is comparing this to dspbox size correct?

tobonex (commented):
Self-review with some questions/things to look at.

src/schedule/schedule.c
```diff
@@ -75,3 +76,33 @@ void scheduler_init(int type, const struct scheduler_ops *ops, void *data)
...
+		scheduler_props->task_count++;
+		task_props = (struct task_props *)
+			((uint8_t *)scheduler_props + sizeof(struct scheduler_props));
+
+		/* Fill TaskProps */
+		task_props->task_id = 0;
+		*data_off_size += sizeof(struct task_props);
+
+		/* Left unimplemented */
+		task_props->module_instance_count = 0;
```
lyakh, line 106: this is a dummy so far. Not getting any real information

src/debug/telemetry/telemetry.c
lyakh: yes, one per core and uncached, which is also the default for static writable data

src/schedule/zephyr_ll.c
```diff
@@ -252,6 +254,17 @@ static void zephyr_ll_run(void *data)
 		       NOTIFIER_TARGET_CORE_LOCAL, NULL, 0);
 }
 
+static void schedule_ll_callback(void *data)
+{
+	const uint32_t begin_count = xthal_get_ccount();
```
lyakh, line 259: I would expect `xthal_get_ccount()` to actually go via Zephyr

src/debug/telemetry/telemetry.c
```diff
@@ -0,0 +1,137 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// Copyright(c) 2023 Intel Corporation. All rights reserved.
```
btian1, line 3: 2024

src/debug/telemetry/telemetry.c
```diff
@@ -0,0 +1,137 @@
...
+#include <limits.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <adsp_debug_window.h>
+#include <mem_window.h>
+#include <zephyr/debug/sparse.h>
+#include <sof/debug/telemetry/telemetry.h>
+#include <sof/audio/module_adapter/module/generic.h>
+#include "adsp_debug_window.h"
```
btian1, line 17: convention is more like longer header file name first.

src/schedule/zephyr_ll.c
```diff
@@ -252,6 +254,17 @@ static void zephyr_ll_run(void *data)
 		       NOTIFIER_TARGET_CORE_LOCAL, NULL, 0);
 }
 
+static void schedule_ll_callback(void *data)
```
btian1, line 257: there already have a similar ccount calculation, not sure you can directly use it:
https://github.com/thesofproject/sof/blob/main/src/schedule/zephyr_ll.c#L137C1-L137C1

line 137 - line 146, it also use ccount for zephyr, same data, you may need modify this part if you want to use it.

src/audio/base_fw.c
```diff
@@ -504,6 +504,28 @@ int schedulers_info_get(uint32_t *data_off_size,
...
+			(struct telemetry_wnd_data *)ADSP_DW->slots[DW_TELEMETRY_SLOT];
+	struct system_tick_info *systick_info =
+			(struct system_tick_info *)wnd_data->system_tick_info;
+
+	switch (state) {
+	case IPC4_PERF_MEASUREMENTS_DISABLED:
+		break;
+	case IPC4_PERF_MEASUREMENTS_STOPPED:
+		for (int i = 0; i < CONFIG_MAX_CORE_COUNT; i++)
+			systick_info[i].peak_utilization = 0;
```
btian1, line 521: I am not sure what exactly you want for peak_utilization, is it for pipeline or core?
do you have a github feature link for telemetry?

src/include/sof/tlv.h
```diff
@@ -25,6 +25,11 @@ struct sof_tlv {
 	char value[];
 } __packed __aligned(4);
 
+struct sof_tl {
+	uint32_t type;
+	uint32_t max_length;
+} __packed __aligned(4);
```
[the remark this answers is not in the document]
lyakh: whenever available, `__packed` and similar macros are recommended by checkpatch over explicit `__attribute__(())` versions

src/debug/telemetry/telemetry.c
lyakh: the preferred order is alphabetic. Often alphabetic within groups, like first all `<zephyr/...>` headers, then all `<sof/...>` headers, then all libc headers

src/audio/base_fw.c
```diff
@@ -462,9 +462,14 @@ static int basefw_get_large_config(struct comp_dev *dev,
 				   uint32_t *data_offset,
 				   char *data)
 {
+	/* We can use extended param id for both extended and standard param id */
+	union ipc4_extended_param_id extended_param_id;
```
lgirdwood, line 537: @tobonex I assume the vendor config get means that it is flexible and can be used to get/set any custom data for any vendor ? If so, is this the right structure, i.e. the telemetry data will be ABI standard data and part of a generic telemetry ABI.

src/audio/base_fw.c
lgirdwood: I would call the Zephyr (low level data) & SOF (high level data) APIs here to get info on LL and DP schedulers as 1st step. I dont think we have any users of the DMA driven scheduling on Zephyr.

src/include/ipc4/base_fw.h
```diff
@@ -671,4 +671,34 @@ enum ipc4_power_state_type {
...
+	struct task_props  task_info[];
+};
+
+struct schedulers_info {
+	/* Specifies number of items in scheduler_info array. */
+	uint32_t        scheduler_count;
+	/* Array of scheduler properties. */
+	struct scheduler_props  scheduler_info[];
+};
+
```
lgirdwood, line 703: This looks like the ABI that will be sent to host ? if so , it should go in the ABI headers and be packed/aligned and have a small id & size header for each struct..

src/debug/telemetry/telemetry.c
```diff
@@ -0,0 +1,137 @@
...
+	/* systick_init */
+	uint8_t slot_num = DW_TELEMETRY_SLOT;
+	volatile struct adsp_debug_window *window = ADSP_DW;
+	struct telemetry_wnd_data *wnd_data = (struct telemetry_wnd_data *)ADSP_DW->slots[slot_num];
+	struct system_tick_info *systick_info =
+			(struct system_tick_info *)wnd_data->system_tick_info;
+
+	window->descs[slot_num].type = ADSP_DW_SLOT_TELEMETRY;
+	window->descs[slot_num].resource_id = 0;
+	wnd_data->separator_1 = 0x0000C0DE;
```
lgirdwood, line 69: some magic number ?

src/debug/telemetry/telemetry.c
```diff
@@ -0,0 +1,137 @@
...
+		(struct system_tick_info *)wnd_data->system_tick_info;
+
+	systick_info[prid].count = systick_counter;
+	systick_info[prid].last_sys_tick_count = current_ccount - begin_ccount;
+	systick_info[prid].max_sys_tick_count =
+			MAX(current_ccount - begin_ccount,
+			    systick_info[prid].max_sys_tick_count);
+	systick_info[prid].last_ccount = current_ccount;
+
+	#ifdef PERFORMANCE_MEASUREMENTS
```
lgirdwood, line 102: indentation.

src/debug/telemetry/telemetry.c
```diff
@@ -0,0 +1,137 @@
...
+				MAX(systick_info[prid].peak_utilization,
+				    measured_systick);
+			systick_info[prid].peak_utilization_4k =
+				MAX(systick_info[prid].peak_utilization_4k,
+				    measured_systick);
+			systick_info[prid].peak_utilization_8k =
+				MAX(systick_info[prid].peak_utilization_8k,
+				    measured_systick);
+		}
+		if ((systick_counter % 0x1000) == 0)
```
lgirdwood, line 127: hopefully the CC will see the constant is a power of 2 and do a bitwise op instead of a %.

src/include/sof/debug/telemetry/telemetry.h
```diff
@@ -0,0 +1,86 @@
...
+/* Slot in memory window 2 to be used as telemetry slot */
+#define DW_TELEMETRY_SLOT 1
+/* Memory of average algorithm of performance queue */
+#define AVG_PERF_MEAS_DEPTH 64
+/* Number of runs taken to calculate average (algorithm resolution) */
+#define AVG_PERF_MEAS_PERIOD 16
+/* disables calculating systick_averages */
+#define PERFORMANCE_MEASUREMENTS
+
+struct system_tick_info {
```
lgirdwood, line 18: external ABI structures ? if so we need packing etc.

src/schedule/zephyr_ll.c
lgirdwood: yep, we need to use Zephyr API here.

src/audio/base_fw.c
```diff
@@ -504,6 +504,28 @@ int schedulers_info_get(uint32_t *data_off_size,
...
+void set_perf_meas_state(const char *data)
+{
+	enum ipc4_perf_measurements_state_set state = *data;
+
+	struct telemetry_wnd_data *wnd_data =
+			(struct telemetry_wnd_data *)ADSP_DW->slots[DW_TELEMETRY_SLOT];
+	struct system_tick_info *systick_info =
+			(struct system_tick_info *)wnd_data->system_tick_info;
+
+	switch (state) {
```
lgirdwood, line 516: We probably want to print the state change for telemetry in the logs as it may have impact on performance or on issue reproducibility. i.e. a developer will know telemetry state when debugging.

lgirdwood (commented):
LGTM, I just have some opens mainly on the ABI.

src/schedule/zephyr_ll.c
tobonex: Yes, it's probably also based on telemetry from reference firmware. The difference is, the old one does not have all the necessary metrics and it outputs to trace. Mine puts output to memory window and its compatible with reference firmware.

As for ccount for zephyr, seems like i need to actually use the zephyr one so I'll take that from the old one.

src/schedule/zephyr_ll.c
tobonex: Ah, you mean I should use functions from the Zephyr side like arch_timing_counter_get() ?

src/schedule/schedule.c
tobonex: Yeah, there's a lot of holes here. The info that reference firmware sends here and info we have here in csof don't seem to be 1 to 1 comparable. Still, telemetry tests use this IPC so I need this one.

src/debug/telemetry/telemetry.c
```diff
@@ -0,0 +1,137 @@
...
+#include <stdint.h>
+#include <adsp_debug_window.h>
+#include <mem_window.h>
+#include <zephyr/debug/sparse.h>
+#include <sof/debug/telemetry/telemetry.h>
+#include <sof/audio/module_adapter/module/generic.h>
+#include "adsp_debug_window.h"
+
+/* systic vars */
+int systick_counter;
```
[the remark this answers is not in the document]
tobonex: Checkpatch told me not to

src/debug/telemetry/telemetry.c
tobonex: Separators were defined in reference FW. Those are only  used to identify telemetry structs when looking at the memory directly (e.g. using RWE), so those can be anything really.

src/debug/telemetry/telemetry.c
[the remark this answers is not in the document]
lyakh: > why you don't init as 0?

@btian1 it's a static variable. And it's C. All static variables are initialised to 0 in C (unless explicitly initialised to something else).

src/audio/base_fw.c
lyakh: but we actually need all the cores, don't we. Need a plan how to get that

src/debug/telemetry/telemetry.c
```diff
@@ -0,0 +1,137 @@
...
+int prev_ccount;
+int perf_period_sum_;
+int perf_period_cnt_;
+struct perf_queue perf_queue = {0};
```
lyakh, line 24: btw, if all these variables are only used in this file, please mark them static, otherwise please namespace them with a prefix like `telemetry_*`. You can add the prefix even if they're local in fact too.

src/debug/telemetry/telemetry.c
```diff
@@ -0,0 +1,137 @@
...
+		systick_info[i].last_ccount = 0;
+		systick_info[i].avg_utilization = 0;
+		systick_info[i].peak_utilization = 0;
+		systick_info[i].peak_utilization_4k = 0;
+		systick_info[i].peak_utilization_8k = 0;
+	}
+	return 0;
+}
+
+void update_telemetry(uint32_t begin_ccount, uint32_t current_ccount)
```
lyakh, line 85: should it be `telemetry_update()` to use a uniform `telemetry_*` namespace, especially for exported functions?

src/debug/telemetry/telemetry.c
```diff
@@ -0,0 +1,137 @@
...
+		perf_period_sum_ += measured_systick;
+		perf_period_cnt_ = (perf_period_cnt_ + 1) % AVG_PERF_MEAS_PERIOD;
+		if (perf_period_cnt_ == 0) {
+			/* Append average of last AVG_PERF_MEAS_PERIOD runs */
+			perf_queue_append(&perf_queue, perf_period_sum_ / AVG_PERF_MEAS_PERIOD);
+			perf_period_sum_ = 0;
+			/* Calculate average from all buckets */
+			systick_info[prid].avg_utilization = perf_queue_avg(&perf_queue);
+		}
+		if (systick_counter > 1) {
```
lyakh, line 116: looks wrong: you already know, that it's `> 2`

src/include/sof/debug/telemetry/telemetry.h
```diff
@@ -0,0 +1,86 @@
...
+	 * uint32_t separator_exception;
+	 * CoreExceptionRecord core_exception_record[FW_REPORTED_MAX_CORES_COUNT];
+	 */
+};
+
+/* Reference FW used a normal Queue here.
+ * Implementing simplified queue just for avg calculation.
+ * Queue is circular, oldest element replaced by latest
+ */
+struct perf_queue {
```
lyakh, line 73: `telemetry_` prefix?

src/schedule/zephyr_ll.c
```diff
@@ -15,6 +15,8 @@
 #include <sof/lib/perf_cnt.h>
 #include <zephyr/kernel.h>
 #include <ipc4/base_fw.h>
+#include <sof/debug/telemetry/telemetry.h>
+
```
lyakh, line 19: no need for an additional empty line

src/debug/telemetry/telemetry.c
lgirdwood: ok, so to be scalable (where FW can add and modify data payloads in a way that wont break host based tooling) we would need to put a small header as the separator to convey extra information instead of just a magic number e.g.
```c
struct telem_hdr {
    uint16_t magic; // 0xc0de
    uint16_t type; // type of C struct, i.e. what am I.
    uint16_t words;  // size of data elem in words
    uint16_t elem; // > 1 if payload is array
} __packed;
// payload follows here
```
This could convey all the info tooling needs and would allow FW to grow data and not break tooling along teh way. i.e. if tooling does not know about a ```type``` then it can go to the next header (and so on).

src/schedule/zephyr_ll.c
btian1: sof_cycle_get_64

src/audio/base_fw.c
tobonex: This made me think a bit. This IPC doesn't do much as of yet. It does some additional things in reference firmware, but nothing directly related to systick measurement (tests still use this one so we need it). It seems that systick measurement  in reference FW was on all the time, same here. This does not seem like it would take a lot of time but still...
I could modify this IPC and add a bool so that systick measurement is stopped unless it's triggered by this IPC.

src/debug/telemetry/telemetry.c
tobonex: The structure of telemetry data in memory window 2 here is the same as in reference firmware. If we want to be compatible it would be best to leave it as it is.

src/debug/telemetry/telemetry.c
tobonex: PERFORMANCE_MEASUREMENTS  I just ported from reference FW, for now there is no reason for it it seems. Will probably just delete this one. I added TELEMETRY_CONFIG instead to disable the whole thing.

src/audio/base_fw.c
tobonex: This one is just a special case of large_config_get that reference FW used. Scheduler_info_get IPC uses this format. Normal param_id uses one byte out of 4 and extended one uses the remaining bytes as space for additional argument.

src/debug/telemetry/telemetry.c
```diff
@@ -0,0 +1,138 @@
...
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+/* systic vars */
+int systick_counter;
+int prev_ccount;
+int perf_period_sum_;
+int perf_period_cnt_;
+struct telemetry_perf_queue perf_queue = {0};
```
lyakh, line 25: please make all above `static` and use a consistent namespace

lyakh (dismissed):

src/debug/telemetry/telemetry.c
lgirdwood: We do need to maintain compatability but we also need to scale and grow as new features upstream. We can wrap any legacy/reference structure as-is, but they would need to be pre-fixed with above header so that new data could be added (without impacting any legacy data).

src/audio/base_fw.c
lgirdwood: Thanks, this would just be informational to the developer. They would at least know if telemetry was on/off when an issue occurred.

tobonex:
Update:
- changed ccount function to sof_cycle_get_64
- telemetry.c variables made static and added a set per core
- other minor fixes

src/debug/telemetry/telemetry.c
tobonex: Still not sure if we can just change the separators. Pre-fixing with the header would also move the fields from their designated places in the MW slot, and things would just stop working.

About scaling: memory window 2 is pretty scalable by itself. We can still extend the window to get more slots where we can add new structures.

src/debug/telemetry/telemetry.c
tobonex: I think we can write this as 
`if((systick_counter & (0x1000 -1)) == 0)`
`if((systick_counter & (0x2000 -1)) == 0)`
 if we want to be sure it is optimized

src/debug/telemetry/telemetry.c
```diff
@@ -0,0 +1,143 @@
...
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+/* Systic variables, one set per core */
+static int telemetry_systick_counter[CONFIG_MAX_CORE_COUNT];
+static int telemetry_prev_ccount[CONFIG_MAX_CORE_COUNT];
+static int telemetry_perf_period_sum[CONFIG_MAX_CORE_COUNT];
+static int telemetry_perf_period_cnt[CONFIG_MAX_CORE_COUNT];
+static struct telemetry_perf_queue telemetry_perf_queue[CONFIG_MAX_CORE_COUNT] = {0};
```
lyakh, line 25: `static` objects are always initialised to 0, unless initialised otherwise explicitly, so this initialisation is redundant

src/debug/telemetry/telemetry.c
```diff
@@ -0,0 +1,143 @@
...
+		q->sum += element;
+		q->index++;
+		q->size++;
+		if (q->index >= AVG_PERF_MEAS_DEPTH) {
+			q->index = 0;
+			q->size = AVG_PERF_MEAS_DEPTH;
+			q->full = true;
+		}
+	} else {
+		/* no space, pop tail */
```
lyakh, line 40: are actually two indications of the storage being full needed? You have `q->full` and `q->index >= AVG_PERF_MEAS_DEPTH` - aren't they the same? If they are, maybe better to use only one of them, to improve clarity and to avoid any potential issues keeping them synchronised

tobonex: The queue here is circular. After the queue is full it stays full. We delete the oldest data point when a new one is given.

The first `q->index >= AVG_PERF_MEAS_DEPTH` is used to determine if the queue is full.  After that the index becomes the tail of the queue and `q->index >= AVG_PERF_MEAS_DEPTH` is used to check if the tail reached the last element. The tail then circles back to 0.

src/audio/base_fw.c
tobonex: Yeah i have some info on those two. Removing LL DMA info from here for now.

src/include/sof/debug/telemetry/telemetry.h
tobonex: Other memory window structs have __packed only. I'll add it.

zephyr/CMakeLists.txt
```diff
@@ -900,6 +900,10 @@ zephyr_library_sources_ifdef(CONFIG_SOF_BOOT_TEST
        boot_test.c
 )
 
+zephyr_library_sources_ifdef( CONFIG_TELEMETRY
+       ${SOF_SRC_PATH}/debug/telemetry/telemetry.c
```
marc-hb, line 904: Do need add C files in this top-level file. CMake and Kconfig belong to subdirectories, I'm trying to scatter this giant file (#8260).

Use Zephyr conditionals in `src/debug/telemetry/CMakeLists.txt` like ef563ebfe3d572 did in `src/init/CMakeLists.txt`

BTW are you testing BOTH Zephyr and non-Zephyr? If you're testing only one then add only the one you tested.

zephyr/Kconfig
```diff
@@ -77,4 +77,10 @@ config SOF_BOOT_TEST
 	  initialized. After that SOF will continue running and be usable as
 	  usual.
 
+config TELEMETRY
+	bool "enable telemetry"
+	default n
+	help
+	  Enable Telemetry
```
marc-hb, line 84: This should probably be in a subdirectory too?

marc-hb (changes requested):

RanderWang (commented):
Since basefw will be shared by all vendors,  we should figure out a method to move intel specific code out

src/audio/base_fw.c
```diff
@@ -455,6 +459,50 @@ static int basefw_set_fw_config(bool first_block,
...
+
+	/* ===================== LL_DMA SCHEDULER INFO ============================ */
+	/* Could not make this one work */
+	/* schedulers_info->scheduler_count++;
+	 * scheduler_props = (struct scheduler_props *)(data + *data_off_size);
+	 * scheduler_get_task_info_ll_dma(scheduler_props, data_off_size);
+	 */
+
+	/* ===================== DP SCHEDULER INFO ============================ */
+#if !CONFIG_TIGERLAKE
```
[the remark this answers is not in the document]
tobonex: I'll replace this one with CONFIG_ZEPHYR_DP_SCHEDULER

zephyr/CMakeLists.txt
tobonex: I'm testing Zephyr only, building with xtensa-build-zephyr.py. First time adding files here so my first idea was adding CmakeLists in subdirectories but it didn't work for me - it's not even parsed . Should I then use `zephyr/CMakeLists.txt` and delete the subdirectory one or is `zephyr/CMakeLists.txt` still a no-go?

src/audio/base_fw.c
tobonex: Got some more info, systick info measurement works all the time by design.  So there's no on/off switch. Still, what I can do is adding the log at the init of telemetry indicating that systick info measurement is active.

zephyr/CMakeLists.txt
marc-hb: Thanks for the additional information. I also took a closer look. Try this:
```diff
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -146,6 +146,7 @@ endmacro()
 
 add_subdirectory(../src/init/ init_unused_install/)
 add_subdirectory(../src/ipc/  ipc_unused_install/)
+add_subdirectory(../src/debug/telemetry/  telemetry_unused_install/)
 add_subdirectory(test/)
```

Then add_sources the new C file from its subdirectory _only_. Funny enough you were already doing that but it had no effect yet. This line above should fix that.

So you were adding the same C file from two different places. That's why I wondered whether you were testing the non-Zephyr/XTOS build too.

Generally speaking, don't add CMake code that is already dead! TEST every line of CMake change by temporarily deleting it (much easier and faster than testing C code)

src/debug/telemetry/telemetry.c
kv2019i: I think we have to live with the fixed location (slot + slot offset + separator) to keep compatibility.

src/debug/telemetry/telemetry.c
```diff
@@ -0,0 +1,143 @@
...
+	systick_info[prid].last_sys_tick_count = current_ccount - begin_ccount;
+	systick_info[prid].max_sys_tick_count =
+			MAX(current_ccount - begin_ccount,
+			    systick_info[prid].max_sys_tick_count);
+	systick_info[prid].last_ccount = current_ccount;
+
+#ifdef PERFORMANCE_MEASUREMENTS
+	const size_t measured_systick = begin_ccount - telemetry_prev_ccount[prid];
+
+	telemetry_prev_ccount[prid] = begin_ccount;
```
kv2019i, line 107: This is interesting how we share the implementation with existing routines. We have similar (but not same) calculation in LL scheduler to calculate the LL load. This is done with a very lean implementation to allow it to be left on in product builds. How is this code for PERFORMANCE_MEASUREMENTS ? Is this designed to be fast enough to be left on in release builds?

src/include/sof/debug/telemetry/telemetry.h
```diff
@@ -0,0 +1,86 @@
...
+};
+
+/*
+ * This is the structure of telemetry data in memory window.
+ * If you need to define a field, you should also define the fields before it to
+ * keep the internal structures aligned with each other.
+ */
+struct telemetry_wnd_data {
+	uint32_t separator_1;
+	struct system_tick_info system_tick_info[CONFIG_MAX_CORE_COUNT];
```
kv2019i, line 37: How is this aligned with the coredump backend. See commit 0bff695d2c997363eb0c6cbf9421b056d645ea4a in zephyr by @pjdobrowolski . If we have users of the same slot both in Zephyr and SOF, that is a potential problem and conflicts are possible.

kv2019i (commented):
Generally this looks pretty ok. I don't mind the gaps in implementation and am ok to start merging this with gaps in implementation. We do need to solve the overall approach on where the debug slow content is managed. This would seem to bo conflict with the coredump backend (which may be ok if we treat coredump as a special case, but needs to be very explicit in the code).

pjdobrowolski:
Separators are useful to read data in f.e. http://rweverything.com/ in real life.

src/include/sof/debug/telemetry/telemetry.h
pjdobrowolski: Possible conflicts must be investigate. After  /* uint32_t separator_exception; */ we should have space where core dump should be placed without problems between zephyr and sof.

src/debug/telemetry/telemetry.c
lyakh: `& 0xfff` and `& 0x1fff`

src/include/sof/debug/telemetry/telemetry.h
tobonex: Yes, coredump was here before we had telemetry structure, it's supposed to be placed at the end of it. Currently it just overwrites the structure, but only after the firmware crashes, so this memory is not really used anymore by then. Not much of an issue for now, it still works. We can realign coredump later.

pjdobrowolski: We can do it in this PR. It is connected by space in MW.

src/debug/telemetry/telemetry.c
lgirdwood: This is a debug mechanism that has to grow and scale as FW features, use cases and capabilities grow and scale (and they will), its way simpler and cheaper to fix the debug tooling (i.e. look for a header to find start, all other data is the same) than to continue with an ABI that cant grow with the FW.

dbaluta:
@tobonex is there a Linux corresponding PR?

src/include/ipc4/module.h
```diff
@@ -66,6 +66,13 @@ union ipc4_extended_param_id {
 	} part;
 } __attribute__((packed, aligned(4)));
 
+struct vendor_error {
```
dbaluta, line 69: How about calling this ipc4_vendor_error?

src/include/ipc4/module.h
```diff
@@ -66,6 +66,13 @@ union ipc4_extended_param_id {
...
+	/* Index of the failed parameter */
+	uint32_t failed_param_index;
+	/* Error code */
+	uint32_t vendor_error_code;
```
dbaluta, line 73: This struct is already called vendor, I think using just `err_code` is fine. Also, you can just use `param_index` because we already know the context from struct name. 

```

struct ipc4_vendor_error {
uint32_t param_idx;
uint32_t err_code
};

```

src/audio/base_fw.c
```diff
@@ -455,6 +459,50 @@ static int basefw_set_fw_config(bool first_block,
...
+	schedulers_info->scheduler_count++;
+	scheduler_props = (struct scheduler_props *)(data + *data_off_size);
+	scheduler_get_task_info_ll(scheduler_props, data_off_size);
+
+	/* ===================== LL_DMA SCHEDULER INFO ============================ */
+	/* Could not make this one work */
+	/* schedulers_info->scheduler_count++;
+	 * scheduler_props = (struct scheduler_props *)(data + *data_off_size);
+	 * scheduler_get_task_info_ll_dma(scheduler_props, data_off_size);
+	 */
```
dbaluta, line 499: Please don't leave commented code in your PR.

app/boards/intel_adsp_ace20_lnl.conf
```diff
@@ -37,6 +37,7 @@ CONFIG_INTEL_ADSP_TIMER=y
 CONFIG_MM_DRV_INTEL_ADSP_TLB_REMAP_UNUSED_RAM=y
 CONFIG_AMS=y
 CONFIG_COUNTER=y
+CONFIG_TELEMETRY=y
```
dbaluta, line 41: Configs should be enabled in a separate patch.

lgirdwood:
> Separators are useful to read data in f.e. http://rweverything.com/ in real life.

Not disagreeing for immutable data formats, but we will want to grow over time and this means we have to have a method to communicate to tooling about data format changes. We can probably grow  by updating the below 2 items to indicate ABI version.
```c
window->descs[slot_num].type = ADSP_DW_SLOT_TELEMETRY;
window->descs[slot_num].resource_id = 0;
```

zephyr/Kconfig
tobonex: Added Kconfig to src/debug

zephyr/CMakeLists.txt
```diff
@@ -146,9 +146,12 @@ endmacro()
 
 add_subdirectory(../src/init/ init_unused_install/)
 add_subdirectory(../src/ipc/  ipc_unused_install/)
+add_subdirectory(../src/debug/telemetry/  telemetry_unused_install/)
 add_subdirectory(test/)
 
 
+
+
```
marc-hb, line 154: Noise

src/debug/Kconfig
```diff
@@ -0,0 +1,8 @@
+# SPDX-License-Identifier: BSD-3-Clause
+
+config TELEMETRY
+	bool "enable telemetry"
+	default n
+	help
+	  Enable Telemetry
```
marc-hb, line 7: If you really want to take a shortcut then do `rsource debug/telemetry/Kconfig` at the top-level but not this. Keep telemetry files in the telemetry subdirectory.

marc-hb (dismissed):

zephyr/CMakeLists.txt
tobonex: Yes, this works, thanks! Check if it looks good now, if you can.

src/debug/Kconfig
tobonex: Like this?

src/debug/telemetry/Kconfig
```diff
@@ -0,0 +1,8 @@
+# SPDX-License-Identifier: BSD-3-Clause
+
+config TELEMETRY
```
marc-hb, line 3:
```suggestion
config SOF_TELEMETRY
```
... otherwise high risk of clashing with Zephyr or some Zephyr dependency.

https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html#rule-a-3-macro-name-collisions

This is not theoretical : we have already experienced name clashes like this.

marc-hb (commented):
TELEMETRY -> SOF_TELEMETRY

Everything else in CMake and Kconfig looks good to me.

src/debug/telemetry/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+# SPDX-License-Identifier: BSD-3-Clause
+
+if (CONFIG_TELEMETRY)
+	add_local_sources(sof telemetry.c)
+endif()
```
marc-hb, line 5: Shortcut:
```
add_local_sources_ifdef(CONFIG_[SOF_]TELEMETRY  sof
   telemetry.c)

```

app/boards/intel_adsp_ace20_lnl.conf
tobonex: Separate commit or separate PR?

marc-hb: patch = commit.

The lines between commits is very rarely ambiguous. The line between PRs is less well-defined.

Drawing the lines between commits is easy: just think about `git cherry-pick`, `git revert`, `git bisect` etc. There are many pages about this, this is my favorite: https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes

Drawing the line between PRs is more difficult. (1) put yourself in the shoes of reviewers (2) think "Continuous Integration".

In this case: would it be good to have one or two round of daily tests with the code merged but disabled?

This is the most relevant page:

src/ipc/ipc4/handler.c
```diff
@@ -914,6 +914,99 @@ static int ipc4_unbind_module_instance(struct ipc4_message_request *ipc4)
...
+{
+	const struct sof_tl *input_tl = (struct sof_tl *)data_in;
+	int ret;
+	struct ipc4_vendor_error *error;
+
+	if (init_block && final_block) {
+		/* we use data_off_size as in/out,
+		 *  save value to new variable so it can be used as out size
+		 */
+		const uint8_t *end_offset = data_in + *data_off_size;
```
marc-hb, line 935:
```
/home/runner/work/sof/sof/src/ipc/ipc4/handler.c:930:45:
 error: pointer targets in initialization of ‘const uint8_t *’
 {aka ‘const unsigned char *’} from ‘const char *’ differ in signedness [-Werror=pointer-sign]
  930 |                 const uint8_t *end_offset = data_in + *data_off_size;
      |                                             ^~~~~~~
```

src/ipc/ipc4/handler.c
```diff
@@ -914,6 +914,99 @@ static int ipc4_unbind_module_instance(struct ipc4_message_request *ipc4)
...
+		size_t produced_data = 0;
+
+		while ((const uint8_t *)input_tl < end_offset) {
```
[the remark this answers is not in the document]
tobonex: Yes, we can't implicitly comapre different pointers, compiler will throw a warning.

src/ipc/ipc4/handler.c
```diff
@@ -914,6 +914,99 @@ static int ipc4_unbind_module_instance(struct ipc4_message_request *ipc4)
...
+			/* local size is in/out: max msg len goes in, msg len goes out */
+			data_off_size_local = input_tl->max_length;
+			ret = drv->ops.get_large_config(dev, input_tl->type,
+							true,
+							true,
+							&data_off_size_local, output_tlv->value);
+			if (ret) {
+				/* This is how the original FW handled error here. Currently
+				 * no memory is allocated for output in case of error,
+				 * so this may be obsolete.
```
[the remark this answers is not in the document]
tobonex: I'll change it to "reference firmware"

src/ipc/ipc4/handler.c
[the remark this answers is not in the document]
marc-hb: Casts don't magically make all problems go away...  The compiler can do crazy things with undefined behaviors. Comparing pointers is well defined only with arrays and inside structs. If this is an array then use one.

src/ipc/ipc4/handler.c
tobonex: Yeah, this looks better. I would still move '*data_off_size/sizeof(struct sof_tl)' to a const var, so that we don't divide each iteration.

src/ipc/ipc4/handler.c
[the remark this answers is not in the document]
lyakh: > Comparing pointers is well defined only with arrays and inside structs.

@marc-hb Sorry, not sure I understand this your statement. Pointers are routinely compared for any "buffer" type objects. You can say, that they are (implicit) arrays but often enough they aren't declared as such. You can do anything like
```
uint8_t *x = malloc(1024);
read(file, x, 1024);
uint8_t *y = x;
while (y < x + 1024)
    y = process_next_obj(y);
```
Why wouldn't that be well defined?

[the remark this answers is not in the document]
marc-hb: The `uint8_t[]` returned from malloc is a single object in this case, which means it's probably OK?

But more generally speaking, comparing pointers that belong to _different_ objects is not specified - hence the compiler warning above.

> Why wouldn't that be well defined?

I don't know, ask people who write the standards... Maybe for yet another, obscure backwards compatibility reason like segmentation?

Or yet some other crazy stuff like:
https://queue.acm.org/detail.cfm?id=3212479
https://blog.llvm.org/2011/05/what-every-c-programmer-should-know.html
https://libc.llvm.org/dev/undefined_behavior.html

If it's an array then just use an array - problem solved :-)

src/debug/telemetry/telemetry.c
tobonex: Yes, in reference firmware it was always on, so we can assume it was fast enough.

src/schedule/zephyr_ll.c
```diff
@@ -14,6 +14,10 @@
 #include <rtos/task.h>
 #include <sof/lib/perf_cnt.h>
 #include <zephyr/kernel.h>
+#include <ipc4/base_fw.h>
+#ifdef CONFIG_SOF_TELEMETRY
+#include <sof/debug/telemetry/telemetry.h>
+#endif
```
marc-hb, line 20: `telemetry.h` does not seem to have anything out of the ordinary so I see no reason to make this conditional. Kconfig combinatorial explosion and lack of coverage is already painful enough: no need to add even more coverage gaps and realize even later than some symbols haven't been namespaced and collide.

marc-hb (commented):
Multiple lacks of namespacing and risk of collisions
https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html#rule-a-3-macro-name-collisions

marc-hb:
CAVS25 https://sof-ci.01.org/sofpr/PR8737/build2569/devicetest/index.html looks OK.

On the other hand, MTL NOCODEC tests failed across the board: https://sof-ci.01.org/sofpr/PR8737/build2568/devicetest/index.html
IPC timeout.

src/audio/base_fw.c
```diff
@@ -16,6 +16,14 @@
...
 #include <intel_adsp_hda.h>
 #endif
+#include <sof/audio/module_adapter/module/generic.h>
+#include <sof/schedule/dp_schedule.h>
+#include <sof/schedule/ll_schedule.h>
+#ifdef CONFIG_SOF_TELEMETRY
+#include <sof/debug/telemetry/telemetry.h>
+#include "adsp_debug_window.h"
+#include "mem_window.h"
+#endif
```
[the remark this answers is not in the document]
tobonex: In case of telemetry.h , yes, you are right.
In case of other ones: I excluded some platforms from this patch, as there are some issues with memory window defines being already defined in platform configs.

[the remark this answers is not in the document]
marc-hb: > I excluded some platforms from this patch, as there are some issues with memory window defines being already defined in platform configs.

So you are indeed "sweeping under the rug" exactly the sort of collision problem I was describing. This problem should at the very least be mentioned with a FIXME in the code, maybe pointing to a new bug. Maybe fixing this is less urgent than this PR, I don't know about priorities but you can't just **silently** hide collisions with a comment-less `#ifdef CONFIG_xxx #include` like this.

Kconfig's combinatorial explosion is always an untestable 
nightmare "by design"[*] and hiding problems is only making it worse for later.


[*] even more so considering kconfiglib.py does not implement "randconfig"

src/schedule/zephyr_ll.c
```diff
@@ -14,6 +14,10 @@
 #include <rtos/task.h>
 #include <sof/lib/perf_cnt.h>
 #include <zephyr/kernel.h>
+#include <ipc4/base_fw.h>
+//#ifdef CONFIG_SOF_TELEMETRY
+#include <sof/debug/telemetry/telemetry.h>
+//#endif
```
marc-hb, line 20: Temporary hack

marc-hb (changes requested):

src/debug/telemetry/telemetry.c
```diff
@@ -0,0 +1,145 @@
...
+
+	++telemetry_systick_counter[prid];
+
+	struct telemetry_wnd_data *wnd_data =
+		(struct telemetry_wnd_data *)ADSP_DW->slots[SOF_DW_TELEMETRY_SLOT];
+	struct system_tick_info *systick_info =
+		(struct system_tick_info *)wnd_data->system_tick_info;
+
+	systick_info[prid].count = telemetry_systick_counter[prid];
+	systick_info[prid].last_sys_tick_count = current_ccount - begin_ccount;
```
andyross, line 98: Just to whine a bit, the terminology here seems pretty mixed up:

* The code right here is calling it a "ccount", which implies that it's sourced from the local core's CCOUNT SR.  It isn't[1], it's derived from sof_cycle_get_64(), which is a wrapper around Zephyr k_cycle_get_64(), which is returning the HDA wall clock time.

* It's then being assigned to a variable called "tick_count", implying that it might be a Zephyr tick (12 kHz) or SOF scheduler quantum (1 kHz).  It's not those either.

* The PR is using the term "systick", which isn't a standard thing AFAIK?  Where it's used in the industry, it generally is a reference to the ARM Cortex M internal timer device.  It's not that either. :)

[1] And really can't be; CCOUNT isn't synchronized between cores and thus can't be used as a timestamp.

src/audio/base_fw.c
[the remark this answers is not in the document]
andyross: Agreed.  What's being multiply defined here?  Maybe it's just a header missing an include guard?  As I read it this is only going to build on intel_adsp anyway, what are the other platforms in question?

andyross (commented):
@marc-hb pulled me in.  Some whining about timing terminology, which is sort of a peeve of mine upstream.

andyross (commented):
Also regarding timing: I don't see anywhere in the protocol where it's self-describing enough to actually define the units in question.  The interpreting kernel code just needs to know what the relevant clock rate that the DSP is sampling?  Shouldn't we store that somewhere?

src/audio/base_fw.c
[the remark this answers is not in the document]
tobonex: Yes, I understand. I mentioned the exclusion here on this page somewhere but adding fixme slipped my mind. I'll write something.

The problem is mainly on tgl. We have defines like  HP_SRAM_WIN0_BASE, HP_SRAM_WIN0_SIZE placed in src/platform/tigerlake/include/platform/lib/memory.h , for some reason.  I don't want to just remove them in case there is some context i don't know about. Other platforms get those defines from including the .h files mentioned here.

src/debug/telemetry/telemetry.c
tobonex: Since this is a port from the reference firmware I just kept the original field names. 

About the ccount function, earlier in this review there was a talk about this. Originally I used  ` xthal_get_ccount();` but I was told to use a Zephyr side function, and it ended on using `sof_cycle_get_64()`. Is it really a good replacement? I'm guessing the scaling may be wrong.

About your other question about DSP clock rate: had to check the tests how it was handled and yes, you are right, clock rate was hard coded per platform. It seems that "the other side has to just know" was the idea when this was originally implemented.

marc-hb: > Since this is a port from the reference firmware I just kept the original field names.

The reference firmware is not based on a separate project like Zephyr and it's not multi-vendors either. So it has totally different collision and consistency constraints.

https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html#rule-a-3-macro-name-collisions

Preserving register names and other vendor-specific symbols is great. Preserving names of high level concepts like `tick_count` is not good. It's much more important for these to be consistent with Zephyr and industry standards than with some closed-sourced, self contained project.

So I think this is the key question: is this new feature SOF_TELEMETRY or rather SOF_INTEL_TELEMETRY? Very important question.

tobonex: @andyross Used ccount (xthal_get_ccount) before i was told to change it to sof_cycle_get_64. It was all made with ccount in mind so I'm not sure if it's a good replacement. What is the actual difference between output values of those two? 
 Also why is ccount not good for timestamps? Isn't there one ccount value shared for all cores?

tmleman (changes requested):

src/ipc/ipc4/handler.c
tobonex: I'll add logs in a bit.
Just wanted to be sure if data actually ends up copied to the window. Ipc code seems a bit messy so it's kinda hard to find the exact moment this happens.

lgirdwood:
> Also regarding timing: I don't see anywhere in the protocol where it's self-describing enough to actually define the units in question. The interpreting kernel code just needs to know what the relevant clock rate that the DSP is sampling? Shouldn't we store that somewhere?

@andyross my understanding would be the timestamps/units would be the same used in logging. @tobonex can you confirm ?

src/audio/base_fw.c
```diff
@@ -570,6 +572,8 @@ static int basefw_set_large_config(struct comp_dev *dev,
 	switch (param_id) {
 	case IPC4_FW_CONFIG:
 		return basefw_set_fw_config(first_block, last_block, data_offset, data);
+	case IPC4_PERF_MEASUREMENTS_STATE:
+		return 0;
```
tmleman: Why does the IPC4_PERF_MEASUREMENTS_STATE case in this switch statement always return success (0)? According to the documentation, the driver can check this state via LargeConfigGet and change it with LargeConfigSet.

tmleman (changes requested):

src/ipc/ipc4/handler.c
```diff
@@ -909,6 +909,100 @@ static int ipc4_unbind_module_instance(struct ipc4_message_request *ipc4)
...
+							&data_off_size_local, output_tlv->value);
+			if (ret) {
+				/* This is how the reference firmware handled error here. Currently
+				 * no memory is allocated for output in case of error,
+				 * so this may be obsolete.
+				 */
+				error = (struct ipc4_vendor_error *)data_out;
+				error->param_idx = input_tl[i].type;
+				error->err_code = IPC4_FAILURE;
+				*data_off_size = sizeof(struct ipc4_vendor_error);
```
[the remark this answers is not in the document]
marc-hb: Not just more obvious but also more future-proof.

src/audio/base_fw.c
tobonex: This IPC has limited use for now, but it's used in test cases for systick. But now that you mention it I forgot to add default case returning error to the `set_perf_meas_state` function, so it won't always return 0 (I'll push changes soon).

tobonex:
Update:

- Memory window size increase doesn't have to be done on Zephyr side, so I'm adding a new commit here instead.
- Fixed one issue in scheduler_get_task_info() in schedule.c

src/include/sof/debug/telemetry/telemetry.h
```diff
@@ -0,0 +1,85 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright(c) 2024 Intel Corporation. All rights reserved.
+ */
+
+#ifndef __SOF_TELEMETRY_H__
+#define __SOF_TELEMETRY_H__
+
+/* Slot in memory window 2 (Debug Window) to be used as telemetry slot */
+#define SOF_DW_TELEMETRY_SLOT 1
```
marcinszkudlinski, line 10: Mem window  slots usage cannot be speeded through the code. 
Slot 0 is used for mtrace by Zephyr:


```
log_backend_adsp_mtrace.c:

static void mtrace_init(void)
{
	if (ADSP_DW->descs[0].type == MTRACE_LOGGING_SLOT_TYPE(MTRACE_CORE)) {
		return;
	}

	ADSP_DW->descs[0].type = MTRACE_LOGGING_SLOT_TYPE(MTRACE_CORE);
}
```

using a magic number... :( 

that will lead to a mess - "which slot is free?? Who knows..."

please add a define for MTRACE and put both defines in adsp_debug_window.h

marcinszkudlinski (changes requested):

src/audio/base_fw.c
tmleman: Since you're adding support for this IPC in a later commit, this one line can be skipped in this.

tmleman (approved):
I believe all my comments have been addressed.

src/schedule/schedule.c
```diff
@@ -75,3 +76,38 @@ void scheduler_init(int type, const struct scheduler_ops *ops, void *data)
...
+	 */
+	struct task_props *task_props;
+	struct list_item *tlist;
+
+	scheduler_props->core_id = cpu_get_id();
+	scheduler_props->task_count = 0;
+	*data_off_size += sizeof(*scheduler_props);
+
+	task_props = (struct task_props *)((uint8_t *)scheduler_props + sizeof(*scheduler_props));
+	list_for_item(tlist, tasks) {
```
lyakh, line 96: do any locks have to be held while traversing the list

tobonex: I only do reading here but if the list size changes in some other thread that might be a problem. I'm guessing i should just use the locks to be safe. Not in this function though, but in the scheduler specific functions that call this general one, as each scheduler uses different locks. I'll get to it.

src/ipc/ipc4/handler.c
```diff
@@ -909,6 +909,100 @@ static int ipc4_unbind_module_instance(struct ipc4_message_request *ipc4)
...
+						data_off_size, output_buffer);
+
+		/* on error report which param failed */
+		if (ret) {
+			error = (struct ipc4_vendor_error *)data_out;
+			error->param_idx = input_tl->type;
+			error->err_code = IPC4_FAILURE;
+			*data_off_size = sizeof(*error);
+			ipc_cmd_err(&ipc_tr, "error: get_large_config returned %d", ret);
+			tr_err(&ipc_tr, "error: get_large_config returned %d", ret);
```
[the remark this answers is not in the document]
tobonex: Removed the extra print.

src/schedule/schedule.c
tobonex: Added locks to zephyr_ll and dp files, where this function is called.

softwarecki (approved):

src/debug/telemetry/telemetry.c
kv2019i: I think thing is still open. We already have overload of "tick" as this is widely used term with a very specific meaning in Zephyr, and on SOF side, IPC4 has added definition of tick in base_fw.h that is not the same as Zephyr. Now in this code, a third type of "tick" is added.

Given this is a very hardware specific interface to begin with, I'd actually just use "ccount" directly to make this explicit. CCOUNT is not shared by cores, and with DSP clock rate changes, clock/power gating, the ccount will not be a usable timestamp even for a single core. That's why we use sof_cycle_get_64() (which is not ccount) for timestamps in SOF. But if the telemetry interface expects ccount, then that's what you need to fill.

But what are the systick fields here? I've seen some uses of "DSP ticks" (= CCOUNT), so I guess that is possible. I'd then add comments to the code as this will be very confusing as this is the third type of "tick" in the same codebase.

If you compile Zephyr with CONFIG_TIMING_FUNCTIONS, then you can get CCOUNT in a more portable way with timing_counter_get() on xtensa. However, we disable CONFIG_TIMING_FUNCTIONS by default in the build and only use it performance overlay builds. so this will require some additional investigation to added (e.g. @abonislawski disabled the DSP clock gating for perf overlay -> this is not ok for telemetry if it needs to be enabled by default).



Use of word "tick" is very confusing as we have established meaning of ticks in Zephyr and before this PRGiven this is a very hardware specific interface

src/include/sof/debug/telemetry/telemetry.h
kv2019i: @tobonex Can you confirm telemetry will not overwrite the coredump? This is minimum requirement to merge this PR.

pjdobrowolski (approved):

gbernatxintel (approved):

src/audio/base_fw.c
```diff
@@ -455,16 +467,85 @@ static int basefw_set_fw_config(bool first_block,
...
+
+	/* ===================== LL_TIMER SCHEDULER INFO ============================ */
+	schedulers_info->scheduler_count++;
+	scheduler_props = (struct scheduler_props *)(data + *data_off_size);
+	scheduler_get_task_info_ll(scheduler_props, data_off_size);
+
+	/* ===================== DP SCHEDULER INFO ============================ */
+#if CONFIG_ZEPHYR_DP_SCHEDULER
+	schedulers_info->scheduler_count++;
+	scheduler_props = (struct scheduler_props *)(data + *data_off_size);
```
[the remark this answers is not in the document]
tobonex: Data pointer does not move, it's just the begining of data. I recalculate the pointer to the structs using updated data_off_size.  data_off_size changes inside those functions, so there's no overwriting

src/audio/base_fw.c
```diff
@@ -16,6 +16,10 @@
 #if defined(CONFIG_SOC_SERIES_INTEL_ADSP_ACE)
 #include <intel_adsp_hda.h>
 #endif
+#include <sof/debug/telemetry/telemetry.h>
```
jsarha: "sof/debug/telemetry/telemetry.h" is added couple commit later in "telemetry: Add systick_info part of telemetry", so this sure breaks build and bisect here.

jsarha (changes requested):
Some open issues still unaddressed, and one more from me. Otherwise looks Ok to me. But in the future I think we should write some generic debug slot reservation code to Zephyr size so that all the slots would be allocated in the same place.

src/include/sof/debug/telemetry/telemetry.h
tobonex: Currently it's the coredump that overwrites telemetry in case of a crash. Not great but functionally coredump should work the same. Still, my plan was to put coredump in its designed place in the window after this gets merged. Coredump is currently at the beginning of the window and apparently should be somewhere close to the end.

tobonex:
> Some open issues still unaddressed, and one more from me. Otherwise looks Ok to me. But in the future I think we should write some generic debug slot reservation code to Zephyr size so that all the slots would be allocated in the same place.

Yes, with @marcinszkudlinski we were planning to clean up the slots in the near future. We were thinking about either defining all the slot owners in one file or writing this reservation process.

src/debug/telemetry/telemetry.c
kv2019i: @tobonex At a minimum, we need a comment to explain the tick semantics and why we are putting non-ccount values into variable called ccount. I can accept to do this on purpose, but you need to have a comment that explains why this is done.

src/include/sof/debug/telemetry/telemetry.h
kv2019i: Ack @tobonex . I just wanted to confirm it is tested that after coredump overwrites the telemetery, we won't have telemetry again overwrite the coredump. This would make coredump unsable in the build.

kv2019i (commented):
Thanks @tobonex . So based on my count, there are two opens left:
- the tick/ccount naming needs to be clarified (at a minimum with a comment explain why we are mixing units)
- fix bisect issue noted by Jyri -> you should not use header/definitions in the series that are only added later in the series. at every git commit, the tree should build

src/debug/telemetry/telemetry.c
tobonex: @kv2019i  One more thing here, if sof_cycle_get_64 is the final pick here, what is it's tick rate compared to ccount? I'll probably need to rescale the timestamp if it's different.
As for the names, I'll probably just rename the fields.

src/include/sof/debug/telemetry/telemetry.h
tobonex: I'll try to crash it a few times to check if it ever happens.

src/debug/telemetry/telemetry.c
kv2019i: @tobonex it will depend on build target, but with most SOF targets (including all Intel ones), sof_cycle_get_64 will be mapped to k_cycle_get_64() which is defined as "* This routine returns the current time in 64-bits, as measured by the                                                                                                   system's hardware clock, if available.  "
This again depends on the build and target, but on Intel platforms this is the SoC timer which runs at rate of CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC (typically 38.4Mhz).

But it really depends on how the telemetry interface is defined. It is defined to track number of DSP cycles, then you have to use ccount() (or similar).

If it's tracking time, then you have to use sof_cycle_get_64() and provide out-of-band means for telemetry use to discover the value of CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC. 

Neither is really correct or incorrect, and if the telemetry interface documents the semantics, both are ok to me. This is only a blocker if semantics say one thing and implementation does something else.

tobonex: @kv2019i Originally (in reference FW) it used ccount, and that was my initial idea here. Tests compare those measurements to the dsp clock rate.
Our tests basically compute the resulting percentage like this:
core_clock = 40000000
max_kpcs = core_clock/1000  
percentage=(ccount/max_kpcs)*100 
Where ccount is the one we measure here for the zephyr_ll_run, read by host from memwindow.

Other reviewers said to use sof_cycle_get_64, but I still think ccount is more appropriate.

kv2019i: It's very dependent on what you are measuring. The DSP clock can be changed at runtime and clock can be gated. But that's not necessary a problem if you calculate percentage of how the load compares to maximum the DSP can do. If you want to calculate how much time DSP spent active, then this formula won't work, but that's not necessarily a problem.

Not sure who recommended sof_cycle_get_64, I can't see in the public comments. It really al depends on what the interface tries to measure. Both ccount and sof_cycle_get_64 will work, but you are reporting different things.

src/audio/base_fw.c
tobonex: fixed

src/debug/telemetry/telemetry.c
tobonex: I'll leave sof_cycle_get_64 here for now then. Systick will be revisited later anyway,  so there's time for potential changes.
About the naming, I changed variable names to be more generic. Added a comment at systick struct to differentiate this "tick" from other ticks.

jsarha (approved):
I only checked that my earlier comment was properly addressed, but for my part this looks good now.

src/Kconfig
```diff
@@ -21,3 +21,5 @@ rsource "ipc/Kconfig"
 rsource "math/Kconfig"
 
 rsource "library_manager/Kconfig"
+
+rsource "debug/telemetry/Kconfig"
```
kv2019i, line 25: This should be in the previous patch "telemetry: Add systick_info part of telemetry". No functional impact as this commit is the first user, but just noting.

kv2019i (approved):
Thanks @tobonex , looks good now. Let's see if the CI passes. There's one issue (see inline), but we can merge with this if there are no other issues. If you need to update for some reason, please fix this as well.

src/Kconfig
marc-hb: However you must make sure every commit can compile and run basic tests (not telemetry tests)

wszypelt:
SOFCI TEST

src/Kconfig
tobonex: Every commit here does compile

Issue #8642, opened 2023-12-18
[BUG] Various IPC timed out with multi-core run simultaneously


keqiaozhang:
**Describe the bug**
Observed this issue on both main branch and mtl-007-drop-stable when checking the multiple-pipeline stress test. The reproduction rate is almost 100%. Sometimes it fails with a simple I/O error and no error logs in kernel message, but sometimes it fails with different IPC timed out.

**dmesg**
```
[  311.736404] kernel: snd_sof:sof_ipc4_widget_setup: sof-audio-pci-intel-mtl 0000:00:1f.3: Create pipeline pipeline.1 (pipe 1) - instance 0, core 0
[  311.736409] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc tx      : 0x11000005|0x0: GLB_CREATE_PIPELINE
[  312.241983] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ipc timed out for 0x11000005|0x0
[  312.242016] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: Attempting to prevent DSP from entering D3 state to preserve context
[  312.242021] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ IPC dump start ]------------
[  312.242067] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: Host IPC initiator: 0x91000005|0x0|0x0, target: 0x1b060000|0x0|0x0, ctl: 0x3
[  312.242094] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ IPC dump end ]------------
[  312.242113] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ------------[ DSP dump start ]------------
[  312.242132] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: IPC timeout
[  312.242147] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: fw_state: SOF_FW_BOOT_COMPLETE (7)
[  312.242187] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ROM status: 0x5, ROM error: 0x0
[  312.242205] kernel: sof-audio-pci-intel-mtl 0000:00:1f.3: ROM debug status: 0x50000005, ROM debug error: 0x0
[...]

PR #8748, opened 2024-01-17
topology2: add nocodec support for real case validation

RanderWang:
Sof-mtl-max98357a-rt5682 is used by chrome project. This patch uses nocodec backend for validation so that our CI test can check real production case.


RanderWang:
sof-mtl-max98357a-rt5682-nocodec

mengdonglin (approved):

mengdonglin:
Thank you @RanderWang ! This help us simulate Chrome use scenario in our testing. Maybe we can name it "sof-mtl-mock-chrome-nocodec"? But I feel current name is also good.

plbossart:
Not following @RanderWang, why not use a 'nocodec' topology we already have? Why do we need multiple flavors of nocodec?

Also what does this bring in terms of validation?

You haven't quite described the problem statement and alternatives....

mengdonglin:
@plbossart The background is that we want to evaluate the impact of multicore timeout issue https://github.com/thesofproject/sof/issues/8642 for Chrome by running a similar topology as Chrome on MTL RVP.  That issue is reproduced on MTL RVP with nocodec topology (3 cores, dummy smart amp, 11 concurrent streams). This PR made a topology for 2 cores, no smart amp, AEC on core 2, and less streams.

lgirdwood:
> Thank you @RanderWang ! This help us simulate Chrome use scenario in our testing. Maybe we can name it "sof-mtl-mock-chrome-nocodec"? But I feel current name is also good.

Ack - this will differentiate and allow stress and developer debug/test.

PR #8753, opened 2024-01-17
idc: zephyr: set of fixes to IDC in multicore scenario

kv2019i:


src/idc/zephyr_idc.c
```diff
@@ -120,6 +129,18 @@ int idc_send_msg(struct idc_msg *msg, uint32_t mode)
...
 	int idc_send_memcpy_err __unused;
 
+	if (!cpu_is_core_enabled(target_cpu)) {
+		tr_err(&zephyr_idc_tr, "Core %u is down, cannot sent IDC message", target_cpu);
+		return -EACCES;
+	}
+
+	if (zmsg->work_status == WORK_SUBMITTED) {
+		work->sync = 1;
+		ret = k_p4wq_wait(work, K_USEC(IDC_TIMEOUT));
```
kv2019i, line 139: This is bit of double-counting, the internal semaphore in p4wq should in theory be sufficient. The only case where this matters is the very first usage. The work item is zeroed out at init, so k_p4wq_wait() on uninitialized work item is undefined behaviour, so we need some outside bookeeping. Let me see if I can clean this up further.

src/idc/zephyr_idc.c
```diff
@@ -73,8 +73,7 @@ static void idc_handler(struct k_p4wq_work *work)
 	int payload = -1;
 	k_spinlock_key_t key;
 
-	/* A message is received from another core, invalidate local cache */
-	sys_cache_data_invd_range(msg, sizeof(*msg));
+	__ASSERT_NO_MSG(!is_cached(msg));
```
kv2019i, line 83: I could just remove these asserts in the final patch. It's nice here in the PR review, but I wonder should this be just dropped.

@lyakh any comment, you added these cache ops in the original version?

src/idc/zephyr_idc.c
```diff
@@ -120,6 +129,18 @@ int idc_send_msg(struct idc_msg *msg, uint32_t mode)
 	int ret;
 	int idc_send_memcpy_err __unused;
 
+	if (!cpu_is_core_enabled(target_cpu)) {
+		tr_err(&zephyr_idc_tr, "Core %u is down, cannot sent IDC message", target_cpu);
+		return -EACCES;
+	}
+
+	if (zmsg->work_status == WORK_SUBMITTED) {
+		work->sync = 1;
```
lyakh, line 138: `true`

src/idc/zephyr_idc.c
lyakh, line 139: don't other `work` fields have to be set? E.g. `.handler` is called unconditionally.

src/idc/zephyr_idc.c
```diff
@@ -120,6 +129,18 @@ int idc_send_msg(struct idc_msg *msg, uint32_t mode)
...
+		ret = k_p4wq_wait(work, K_USEC(IDC_TIMEOUT));
+		if (ret < 0 && zmsg->work_status == WORK_SUBMITTED)
+			return ret;
```
lyakh, line 141: have we ever seen this race occurring? Between what and what? Non-blocking operations usually return an error when the action cannot be completed immediately. Wouldn't that be better?
In fact this seems racy too. What if the handler completes between checking for `WORK_SUBMITTED` and calling `k_p4wq_wait()`? Then we'll never be woken up?

kv2019i: I can hit this on a MTL with modified multicore tplg. Still trying to find more info on the scenario as I agree this should not happen normally (and it's rare!). I now reworked the wait logic to only rely o p4wq semaphore, so this should be safe against races (and dropped the work_status field).

kv2019i:
This is not necessarily fixing rootcauses, but I believe this is making the code more robust and can help in debug. Without this PR, if an ID is sent before previous one is compltely handled by the p4wq thread (of the destionation core), we enter unchartered territory. In debug builds, one will get following assert in these cases:

```
[    3.259940] <inf> ipc: ipc_cmd: rx   : 0x13010003|0x0
ASSERTION FAIL [item->thread == ((void *)0)] @ ZEPHYR_BASE/lib/os/p4wq.c:236
```

This PR ensure we will not reuse the work object until p4wq is finished with it.

src/idc/zephyr_idc.c
kv2019i: Droped this code in V2 of the patchset.

kv2019i:
@lyakh Here's one concrete case where this is hit. A set of set-pipeline-state IPCs, sent over to core1 with a non-blocking IDC. Processign one is delayed (I suspect same rootcause as https://github.com/thesofproject/sof/pull/8756 ), and when the next set-pipeline-state is sent, the IDC work queue is busy:

The real rootcause is most probably IPC processing taking too much time, but without this PR, the work item is silently reused in the IDC thread and results can be unpredictable.

```
[   57.796150] <inf> ll_schedule: zephyr_domain_thread_fn: ll core 1 timer avg 5635, max 23095, overruns 1
[   58.820255] <inf> ll_schedule: zephyr_domain_thread_fn: ll core 1 timer avg 5533, max 8715, overruns 0
[   59.844440] <inf> ll_schedule: zephyr_domain_thread_fn: ll core 1 timer avg 5528, max 8226, overruns 0
[   59.862658] <inf> ipc: ipc_cmd: rx   : 0x13000003|0x0
IDC@1: TX 4000000 (IPC 4010f380)
IDC@1: RX 4000000 done (IPC 4010f380)
[   59.863898] <inf> pipe: pipeline_trigger: pipe:0 0x0 pipe trigger cmd 2
[   59.864890] <inf> pipe: pipeline_trigger: Sum of KCPS consumption: 6388, core: 1
[   59.866810] <inf> ll_schedule: zephyr_ll_task_done: task complete 0xa0115680 0x20210U
[   59.866895] <inf> ll_schedule: zephyr_ll_task_done: num_tasks 2 total_num_tasks 2
[   59.867320] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_common_copy(): Copy_bytes 0 + avail bytes 80 < period bytes 384, possible glitch
[   59.867366] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_zephyr_copy(): nothing to copy
[   59.868305] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_common_copy(): Copy_bytes 0 + avail bytes 0 < period bytes 384, possible glitch
[   59.868350] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_zephyr_copy(): nothing to copy
[   59.869365] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_common_copy(): Copy_bytes 0 + avail bytes 0 < period bytes 384, possible glitch
[   59.869410] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_zephyr_copy(): nothing to copy
[   59.870390] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_common_copy(): Copy_bytes 0 + avail bytes 0 < period bytes 384, possible glitch
[   59.870435] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_zephyr_copy(): nothing to copy
[   59.871400] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_common_copy(): Copy_bytes 0 + avail bytes 0 < period bytes 384, possible glitch
[   59.871446] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_zephyr_copy(): nothing to copy
[   59.872293] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_common_copy(): Copy_bytes 0 + avail bytes 0 < period bytes 384, possible glitch
[   59.872338] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_zephyr_copy(): nothing to copy
[   59.873373] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_common_copy(): Copy_bytes 0 + avail bytes 0 < period bytes 384, possible glitch
[   59.873420] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_zephyr_copy(): nothing to copy
[   59.874380] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_common_copy(): Copy_bytes 0 + avail bytes 0 < period bytes 384, possible glitch
[   59.874425] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_zephyr_copy(): nothing to copy
[   59.875361] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_common_copy(): Copy_bytes 0 + avail bytes 0 < period bytes 384, possible glitch
[   59.875406] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_zephyr_copy(): nothing to copy
[   59.876711] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_common_copy(): Copy_bytes 0 + avail bytes 0 < period bytes 384, possible glitch
[   59.876758] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_zephyr_copy(): nothing to copy
[   59.877400] <inf> ipc: ipc_cmd: rx   : 0x13010003|0x0
IDC@1: TX 4000000 (IPC 4010f380)
[   59.877545] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_common_copy(): Copy_bytes 0 + avail bytes 0 < period bytes 384, possible glitch
[   59.877583] <err> zephyr_idc: idc_send_msg: idc_send_msg error -16, target core 1
[   59.877596] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_zephyr_copy(): nothing to copy
[   59.877635] <err> ipc: ipc_cmd: ipc4: FW_GEN_MSG failed with err 4
[   59.878371] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_common_copy(): Copy_bytes 0 + avail bytes 0 < period bytes 384, possible glitch
[   59.878416] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_zephyr_copy(): nothing to copy
```

RanderWang (approved):
LGTM. I like defensive programming.

abonislawski (approved):

lyakh:
> A set of set-pipeline-state IPCs, sent over to core1 with a non-blocking IDC. Processign one is delayed (I suspect same rootcause as #8756 ), and when the next set-pipeline-state is sent, the IDC work queue is busy:

@kv2019i this seems to be a problem right there: the next IPC *must not* be sent by the host before the previous one has been processed and responded to! We've spent **a lot** of effort tracing all possible out-of-order IPC processing flows (LL-thread, other core, LL-thread on other core) and making sure we **really** complete processing one IPC before the next one can arrive!

lyakh (changes requested):
let me -1 this to make sure it isn't merged until fully understood

kv2019i:
@lyakh wrote.
>> A set of set-pipeline-state IPCs, sent over to core1 with a non-blocking IDC. Processign one is delayed (I suspect same 
>> rootcause as #8756 ), and when the next set-pipeline-state is sent, the IDC work queue is busy:
> 
> @kv2019i this seems to be a problem right there: the next IPC _must not_ be sent by the host before the previous one has been processed and responded to! We've spent **a lot** of effort tracing all possible out-of-order IPC processing flows (LL-thread, other core, LL-thread on other core) and making sure we **really** complete processing one IPC before the next one can arrive!

I'm not claiming the calls to idc_send_msg() are valid and the PR will not allow such cases to pass. But what this PR does is to explicitly catch misuse of the interface and raise an error instead of silently proceeding to use a worker object on core X while ownership is still on core Y (as per p4wq interface semantics).

In the specific case, these are no multiple IPCs concurrently. A set-pipeline IPC to core1 is handled, a reply is sent from core1, but core1 is busy with other work and the p4wq thread does not signal work to be ready before core0 sends the next one IPC to core1. 
This is not a normal case, core1 is overloaded for some reason, but my point remains that we should not silently override the busy IDC work object in this case (especially as in assert-enabled build this will hit an assert but in release builds it will not).

Here's one log where this happens:
```
KV: first set-pipeline state
[   14.790320] <inf> ipc: ipc_cmd: rx   : 0x13000003|0x0
KV: passed to core 1
[   14.790400] <err> zephyr_idc: idc_send_msg: IDC@1: TX 4000000 (IPC 4010f380)
KV: at 14.791390 the core1 p4wq thread start running, and handles the IPC 
[   14.791390] <err> zephyr_idc: idc_handler: IDC@1: TX 4000000 worker (IPC 4010f380)
[   14.791471] <inf> pipe: pipeline_trigger: pipe:0 0x0 pipe trigger cmd 2
[   14.791908] <inf> pipe: pipeline_trigger: Sum of KCPS consumption: 6388, core: 1
[   14.793371] <inf> ll_schedule: zephyr_ll_task_done: task complete 0xa0115680 0x20210U
[   14.793418] <inf> ll_schedule: zephyr_ll_task_done: num_tasks 2 total_num_tasks 2
[   14.794043] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_common_copy(): Copy_bytes 0 + avail bytes 48 < period bytes 384, possible glitch
[   14.794148] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_zephyr_copy(): nothing to copy
KV: IPC reply is sent to host, host can send a new IPC (but p4wq worker has NOT signaled the semaphore yet)
[   14.794401] <inf> ipc: ipc_cmd: tx-reply     : 0xb3000000|0x0
[   14.795236] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_common_copy(): Copy_bytes 0 + avail bytes 0 < period bytes 384, possible glitch
[   14.795281] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_zephyr_copy(): nothing to copy
[   14.796288] <err> zephyr_idc: idc_handler: IDC@1: TX 4000000 done (IPC 4010f380)
[   14.796066] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_common_copy(): Copy_bytes 0 + avail bytes 0 < period bytes 384, possible glitch
[   14.796113] <wrn> dai_comp: dai_common_copy: comp:1 0x10004 dai_zephyr_copy(): nothing to copy
KV: next set-pipeline-state IPC comes from host
[   14.796140] <inf> ipc: ipc_cmd: rx   : 0x13010003|0x0
KV: passed to core1
[   14.796225] <err> zephyr_idc: idc_send_msg: IDC@1: TX 4000000 (IPC 4010f380)
KV: and bam, the error. obviously something is wrong on core1 as the semaphore has not been signaled, but without this PR we'd start using the worker object while it's still held by the p4wq worker thread 
[   14.796275] <err> zephyr_idc: idc_send_msg: idc_send_msg error -16, target core 1
[   14.796325] <err> ipc: ipc_cmd: ipc4: FW_GEN_MSG failed with err 4
```

src/idc/zephyr_idc.c
lyakh: @kv2019i looks correct

src/idc/zephyr_idc.c
```diff
@@ -120,6 +120,32 @@ int idc_send_msg(struct idc_msg *msg, uint32_t mode)
...
+		tr_err(&zephyr_idc_tr, "Core %u is down, cannot sent IDC message", target_cpu);
+		return -EACCES;
+	}
+
+	/*
+	 * Handler is NULL when work object has never been submitted.
+	 * In all other cases, we must use k_p4wq_wait() before reuse
+	 * of the object.
+	 */
+	if (work->handler) {
```
lyakh, line 133: the thing is that IIUC this check is unprotected, so 2 cores can simultaneously check this and proceed to use the `work` object simultaneously.

kv2019i: @lyakh This field is only set once object lifetime. It is NULL initially (as a static object in BSS). When the first message is sent to a core, handler is set to "handler = idc_handler".

I.e. this check is only to prevent p4wq_wait on an uninitialized work queue object.

I guess I could just skip this and and call p4wq_wait() always, but this is depending on p4wq internals a bit to work as expect (so that wait returns immediately if the object is fresh out from BSS).

src/idc/zephyr_idc.c
lgirdwood: best to keep and modify any inline comments around cache handling code. i.e. why are we checking the ptr type.

lgirdwood (approved):
LGTM, I can take any new inline comments in a subsequent update. Would like to see this in daily testing.

src/idc/zephyr_idc.c
lyakh: @kv2019i rrright, yes, sorry, forgot about this, but I still think there's a race possibility here: it's the "check-and-take" operation that should be atomic and safe, right? And here it isn't. You added a wait in line 141, but before we submit a new work in line 171, it might be unsafe again? Another core might send an IDC to the same target core and the target core might already be processing it again?

lyakh: @andyross is my understanding of this P4WQ use-case correct? What's the right way to handle this (see above comment)?

lgirdwood: @nashif @ceolin @teburd can anyone help on P4WQ question above ?

lgirdwood:
> let me -1 this to make sure it isn't merged until fully understood

@lyakh are you aligned now ?

lyakh:
> > let me -1 this to make sure it isn't merged until fully understood
> 
> @lyakh are you aligned now ?

@lgirdwood I still think there's a possibility for a race here, e.g. if two secondary cores try to send an IDC message to the primary core at the same time

lgirdwood:
> > > let me -1 this to make sure it isn't merged until fully understood
> > 
> > 
> > @lyakh are you aligned now ?
> 
> @lgirdwood I still think there's a possibility for a race here, e.g. if two secondary cores try to send an IDC message to the primary core at the same time

ok, but is this worse 

> > > let me -1 this to make sure it isn't merged until fully understood
> > 
> > 
> > @lyakh are you aligned now ?
> 
> @lgirdwood I still think there's a possibility for a race here, e.g. if two secondary cores try to send an IDC message to the primary core at the same time

ok, but is this issue less likely than the issue we are seeing today. i.e. is this a stability improvement and yes we can circle back with an update once P4WQ opens are resolved ?

src/idc/zephyr_idc.c
andyross: Just seeing this question now, apologies: The API promises that after p4wq_wait() returns the item has been executed and the queue itself won't touch it anymore.  It doesn't provide any synchronization otherwise, in particular if this code itself may be executed from more than one core then it can indeed race on the single work item.

It looks to me like there's one work item instantiated for each target CPU, so on 3+ core systems that seems like a real risk unless the is more synchronization at a higher level.  You could fix that by implementing a synchronized pool of such work items (c.f. k_mem_slab), or by having one for each unique sender/receiver pair (i.e. 6 on MTL, 12 on TGL).

lyakh:
> but is this issue less likely than the issue we are seeing today. i.e. is this a stability improvement and yes we can circle back with an update once P4WQ opens are resolved ?

@lgirdwood if @kv2019i says that this improves the test results for him, then apparently yes, it makes the race window narrower. In general I don't find changes to address a problem that end up addressing it only partially particularly good, but if we add a "FIXME" comment here and open a bug for this, we could take this perhaps.

kv2019i:
It seems this has multiple issues, so closing this. FYI @abonislawski

Issue #8651, opened 2023-12-20
[BUG] DSP panic when testing multiple-pause-resume on LNL/MTL-NOCODEC


keqiaozhang:
**Describe the bug**
Observed this issue on LNL-NOCODEC platform, we had some similar DSP panic issues before, but this one seems to be a new one.
The reproduction rate is about 50%. 

**dmesg**
```
[ 7757.883706] kernel: snd_sof:sof_ipc4_set_pipeline_state: sof-audio-pci-intel-lnl 0000:00:1f.3: ipc4 set pipeline instance 0 state 4
[ 7757.883713] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-lnl 0000:00:1f.3: ipc tx      : 0x13000004|0x0: GLB_SET_PIPELINE_STATE
[ 7757.889677] kernel: snd_sof:sof_ipc4_log_header: sof-audio-pci-intel-lnl 0000:00:1f.3: ipc rx      : 0x1b0a0000|0x0: GLB_NOTIFICATION|EXCEPTION_CAUGHT
[ 7757.889689] kernel: sof-audio-pci-intel-lnl 0000:00:1f.3: ------------[ DSP dump start ]------------
[ 7757.893295] kernel: sof-audio-pci-intel-lnl 0000:00:1f.3: DSP panic!
[ 7757.894681] kernel: sof-audio-pci-intel-lnl 0000:00:1f.3: fw_state: SOF_FW_BOOT_COMPLETE (7)
[ 7757.896605] kernel: sof-audio-pci-intel-lnl 0000:00:1f.3: ROM status: 0x5, ROM error: 0x0
[ 7757.898418] kernel: sof-audio-pci-intel-lnl 0000:00:1f.3: ROM debug status: 0x0, ROM debug error: 0x0
[ 7757.900502] kernel: sof-audio-pci-intel-lnl 0000:00:1f.3: ROM feature bit enabled
[ 7757.902203] kernel: snd_sof:sof_ipc4_find_debug_slot_offset_by_type: sof-audio-pci-intel-lnl 0000:00:1f.3: Slot type 0x4c455400 is not available in debug window
[ 7757.902205] kernel: sof-audio-pci-intel-lnl 0000:00:1f.3: ------------[ DSP dump end ]------------
[ 7757.904205] kernel: snd_sof:sof_set_fw_state: sof-audio-pci-intel-lnl 0000:00:1f.3: fw_state change: 7 -> 8
[...]

Issue #5794, opened 2022-05-10
[FEATURE][meta] Use Zephyr native interfaces


kv2019i:
This is a meta issue to track work to move SOF code to use Zephyr native interfaces.

Change in brief: SOF application code will transition to Zephyr application interfaces. Change is done incrementally so that build with the "in-tree XTOS" continuous to work.

Guard rails:
 - code in sof/src/audio is the primary application code, should only use Zephyr interfaces
 - drivers in sof/src/drivers  should be moved to Zephyr upstream
 - arch code in sof/src/arch is not used for SOF Zephyr builds, this is only for XTOS builds

Transition tools:
 - sof/zephyr/wrapper.c  - xtos-to-zephyr wrapping that was used for initial introduction of Zephyr. this will grow smaller and smaller and eventually be removed  
 - src/xtensa-wrapper -  zephyr-to-xtos wrapper that can be used to build SOF app code on XTOS
 
Ongoing work:
- [ ] PM runtime move to Zephyr PM model
- [ ] Module init with native Zephyr interface #5406
- [x] Spinlocks #5286 
- [x] System clock for timing Wallclock Clocks #5393 
- [ ] Logging, partially done in #5707 
- [x] Wait #5770

lenghonglin:
> > is there any plan to move code in sof/src/platform/ to zephyr？ I want development drivers and platforms to be out of tree
> 
> Yes - we are just working out the stages for this. Topic at TSC tomorrow.

Is there something news to share after TSC？

lgirdwood:
@lenghonglin apologies, I and some other TSC member were travelling last week. 

Please see minutes here. https://github.com/orgs/thesofproject/teams/steering-committee/discussions/31

lenghonglin:
> @lenghonglin apologies, I and some other TSC member were travelling last week.
> 
> Please see minutes here. https://github.com/orgs/thesofproject/teams/steering-committee/discussions/31

but zephyr does not support DAI interface

lgirdwood:
@kv2019i we should add the DAI/DMA integration here too ?

kv2019i:
Marked multiple items as done. Added one new item for platform code init.

lgirdwood:
@abonislawski @mwasko @dbaluta any items you can add/update here ?

kv2019i:
[...]

PR #8755, opened 2024-01-17
Rename IS_ALIGN() to SOF_IS_ALIGN()

marc-hb:
This is the simplest, cheapest and safest to avoid the collision with
the incoming Zephyr macro IS_ALIGN() submitted in
https://github.com/zephyrproject-rtos/zephyr/pull/67243

As SOF depends on Zephyr, most SOF symbols should be prefixed SOF_ or
something specific.

This is compliant with:
https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html#rule-a-3-macro-name-collisions

### DIFF
--- a/src/audio/dai-zephyr.c
+++ b/src/audio/dai-zephyr.c
@@ -633,7 +633,7 @@ static int dai_set_sg_config(struct dai_data *dd, struct comp_dev *dev, uint32_t
 				 period_count);
 			buf_size = period_count * period_bytes;
 			do {
-				if (IS_ALIGNED(buf_size, max_block_count)) {
+				if (SOF_IS_ALIGNED(buf_size, max_block_count)) {
 					period_count = max_block_count;
 					period_bytes = buf_size / period_count;
 					break;
--- a/src/include/sof/common.h
+++ b/src/include/sof/common.h
@@ -14,14 +14,8 @@
 /* callers must check/use the return value */
 #define __must_check __attribute__((warn_unused_result))
 
-#ifdef __ZEPHYR__
-#include <zephyr/sys/util.h>
-#endif
-
 /* Align the number to the nearest alignment value */
-#ifndef IS_ALIGNED
 #define IS_ALIGNED(size, alignment) ((size) % (alignment) == 0)
-#endif
 
 /* Treat zero as a special case because it wraps around */
 #define is_power_of_2(x) ((x) && !((x) & ((x) - 1)))
--- a/src/include/sof/common.h
+++ b/src/include/sof/common.h
@@ -15,7 +15,7 @@
 #define __must_check __attribute__((warn_unused_result))
 
 /* Align the number to the nearest alignment value */
-#define IS_ALIGNED(size, alignment) ((size) % (alignment) == 0)
+#define SOF_IS_ALIGNED(size, alignment) ((size) % (alignment) == 0)
 
 /* Treat zero as a special case because it wraps around */
 #define is_power_of_2(x) ((x) && !((x) & ((x) - 1)))
--- a/src/audio/kpb.c
+++ b/src/audio/kpb.c
@@ -1895,7 +1895,7 @@ static void kpb_convert_24b_to_32b(const void *linear_source, int ioffset,
 	int i = 0;
 	ae_int24x2 d24 = AE_ZERO24();
 
-	if (!IS_ALIGNED((uintptr_t)out_ptr, 8)) {
+	if (!SOF_IS_ALIGNED((uintptr_t)out_ptr, 8)) {
 		AE_LA24_IP(d24, align_in, in);
 		ae_int32x2 d320 = d24;
 		int higher = AE_MOVAD32_H(d320);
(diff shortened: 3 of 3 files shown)


marc-hb:
Failures to compile with the Zephyr main branch are known and unrelated:
https://github.com/thesofproject/sof/actions/runs/7562861040/job/20594168710?pr=8754

`implicit declaration of function 'z_smp_thread_init';`

They're caused by backwards-incompatible change:
- https://github.com/zephyrproject-rtos/zephyr/pull/64755

marc-hb:
Unrelated pause-resume failure in ACE https://sof-ci.01.org/sofpr/PR8755/build1982/devicetest/index.html . pause-resume is not reliable right now. Maybe #8651?

Unrelated `FW reported error: 104 - Other failure of module instance initialization request` in https://sof-ci.01.org/sofpr/PR8755/build1983/devicetest/index.html

Everything else green.

src/include/sof/common.h
```diff
@@ -14,14 +14,8 @@
...
 #define __must_check __attribute__((warn_unused_result))
 
-#ifdef __ZEPHYR__
-#include <zephyr/sys/util.h>
-#endif
-
 /* Align the number to the nearest alignment value */
-#ifndef IS_ALIGNED
 #define IS_ALIGNED(size, alignment) ((size) % (alignment) == 0)
-#endif
```
lyakh, line 24: if https://github.com/zephyrproject-rtos/zephyr/pull/67243 happens to be merged before this (yes, I know that it's currently blocked) and we move to a Zephyr version after it, this would break compilation. Why not just merge these two commits?

src/audio/dai-zephyr.c
```diff
@@ -633,7 +633,7 @@ static int dai_set_sg_config(struct dai_data *dd, struct comp_dev *dev, uint32_t
 				 period_count);
 			buf_size = period_count * period_bytes;
 			do {
-				if (IS_ALIGNED(buf_size, max_block_count)) {
+				if (SOF_IS_ALIGNED(buf_size, max_block_count)) {
```
lyakh, line 636: is it the plan then after https://github.com/zephyrproject-rtos/zephyr/pull/67243 is merged to move to the "proper" Zephyr-provided `IS_ALIGNED()` and to rename back?

lyakh (commented):
We want to use native Zephyr APIs and this seems to be taking us away from it

src/include/sof/common.h
kv2019i: we probably should hold of until we can get the Zephyr interface merged in. We have the SMP interface in flight (plus the regression), so the air is getting too thick with incompatible combinations.

src/audio/dai-zephyr.c
kv2019i: I think an alternative would be to take the Zephyr change with IS_ALIGNED in, then add IS_ALIGNED definition only for xtos builds in SOF, and call it a day.

The current ifdef can carry us until then.

Or we merge this, but I agree this is going against our long term plans. We want to use native Zephyr interfaces in generic SOF code whenever possible (e.g. #5794).

kv2019i (commented):
I think this we should consider also moving this to rtos abstraction and keep using native interfaces.

marc-hb (commented):
> We want to use native Zephyr APIs and this seems to be taking us away from it

In theory yes. In practice these are 4 lines in 2 files.

Every time we want to use a "Zephyr API", we have to implement some indirection logic for XTOS that we (Intel) can't test any more.

Calling this one-liner macro a "Zephyr API" is blowing it out of proportion.

I'm aware the "air is getting thick" and this is precisely why this PR is the simplest, cheapest and safest solution in the short term. I understand and agree in general but this particular example is just 4 lines; let's not over-engineer it.

marc-hb:
> In theory yes. In practice these are 4 lines in 2 files.

4 lines including... 3 lines in unused and never compiled `kpb.c`. I had to change the defconfig to compile-test it.

I think we have much bigger problems to solve than making this one-liner macro "API compliant"

lyakh (changes requested):
Sorry, no, I don't like this. I really don't think we should add a `SOF_*` version for every Zephyr macro. We want to use them directly and for non-Zephyr builds define them in compatibility headers.

marc-hb:
>  I really don't think we should add a SOF_* version for every Zephyr macro.

I forgot how many times I wrote "just this one, with have bigger problems"

> We want to use them directly and for non-Zephyr builds define them in compatibility headers.

Agreed, please send a better (and tested) PR since you have the time.

marc-hb:
After a chat with @lyakh I realized I forgot to explicit the main concern behind this PR: if the SOF definition is guarded with the (evil) `#ifndef` and the Zephyr definition is not (cause it's evil), then compilation success depends on which _order_ the .h files are included. Anything that depends on inclusion order is of course super evil too.

I admit I didn't check whether inclusion order this an _actual_ problem right now, it was much faster (and safer) to submit this PR. Actually testing this inclusion order problem could require testing multiple defconfig variations again.
 
Even if not today, the inclusion order could very easily become a real problem in any future refactoring. Hence this dead simple PR to move on with our life.

lyakh:
> After a chat with @lyakh I realized I forgot to explicit the main concern behind this PR: if the SOF definition is guarded with the (evil) `#ifndef` and the Zephyr definition is not (cause it's evil), then compilation success depends on which _order_ the .h files are included. Anything that depends on inclusion order is of course super evil too.
> 
> I admit I didn't check whether inclusion order this an _actual_ problem right now, it was much faster (and safer) to submit this PR. Actually testing this inclusion order problem could require testing multiple defconfig variations again.
> 
> Even if not today, the inclusion order could very easily become a real problem in any future refactoring. Hence this dead simple PR to move on with our lives.

@marc-hb no, there's no order problem there. The header is already included before the `#ifndef`: https://github.com/thesofproject/sof/blob/0a3b816c033cd2741f422d9892c952fa3e43fea6/src/include/sof/common.h#L17-L24

marc-hb:
OK, let's not worry then.

PR #8756, opened 2024-01-18
Revert "ipc4: relax the IPC timeout checks and be nicer to other thre…

RanderWang:
…ads"

When a stream is triggered to start, host kernel first sendis trigger start ipc message to fw and then start host dma for this stream. Ipc_wait_for_compound_msg is used to wait for all pipelines in the stream to be complete and need to be done fast since it blocks host to start hda dma. The reverted commit adds a 10 ms delay and results to host copier xrun warning for it can't get data from host dma. And this commit also makes a negative effect for stream_start_offset calculation, so revert it.

This reverts commit 909a3277f1271d91314736fc1eaff35f362176d1.

one fix for https://github.com/thesofproject/linux/issues/4781

### DIFF
--- a/src/ipc/ipc4/handler.c
+++ b/src/ipc/ipc4/handler.c
@@ -510,13 +510,12 @@ static void ipc_compound_msg_done(uint32_t msg_id, int error)
 	}
 }
 
-/* wait for IPCs to complete on other cores and be nice to any LL work */
 static int ipc_wait_for_compound_msg(void)
 {
-	int try_count = 30; /* timeout out is 30 x 10ms so 300ms for IPC */
+	int try_count = 30;
 
 	while (atomic_read(&msg_data.delayed_reply)) {
-		k_sleep(Z_TIMEOUT_MS(10));
+		k_sleep(Z_TIMEOUT_US(250));
 
 		if (!try_count--) {
 			atomic_set(&msg_data.delayed_reply, 0);


RanderWang:
fw warning with reverted commit:
````
[ 4580.800270] <inf> copier: copier_comp_trigger: comp:0 0x4 copier_comp_trigger() cmd 7
[ 4580.800293] <inf> copier: copier_comp_trigger: comp:0 0x4 copier_comp_trigger() cmd 1
[ 4580.800328] <err> copier: copier_comp_trigger: comp:0 0x4 ------- dai pos 4288, dai link 527
[ 4580.800350] <inf> host_comp: host_get_copy_bytes_normal: comp:0 0x4 no bytes to copy, available samples: 0, free_samples: 192
[ 4580.801273] <inf> host_comp: host_get_copy_bytes_normal: comp:0 0x4 no bytes to copy, available samples: 0, free_samples: 192
[ 4580.802271] <inf> host_comp: host_get_copy_bytes_normal: comp:0 0x4 no bytes to copy, available samples: 0, free_samples: 192
[ 4580.803271] <inf> host_comp: host_get_copy_bytes_normal: comp:0 0x4 no bytes to copy, available samples: 0, free_samples: 192
[ 4580.804271] <inf> host_comp: host_get_copy_bytes_normal: comp:0 0x4 no bytes to copy, available samples: 0, free_samples: 192
[ 4580.805271] <inf> host_comp: host_get_copy_bytes_normal: comp:0 0x4 no bytes to copy, available samples: 0, free_samples: 192
[ 4580.806273] <inf> host_comp: host_get_copy_bytes_normal: comp:0 0x4 no bytes to copy, available samples: 0, free_samples: 192
[ 4580.807271] <inf> host_comp: host_get_copy_bytes_normal: comp:0 0x4 no bytes to copy, available samples: 0, free_samples: 192
[ 4580.808271] <inf> host_comp: host_get_copy_bytes_normal: comp:0 0x4 no bytes to copy, available samples: 0, free_samples: 192
[ 4580.809253] <inf> host_comp: host_get_copy_bytes_normal: comp:0 0x4 no bytes to copy, available samples: 0, free_samples: 192
[ 4580.810258] <inf> host_comp: host_get_copy_bytes_normal: comp:0 0x4 no bytes to copy, available samples: 0, free_samples: 192
[ 4580.811248] <inf> host_comp: host_get_copy_bytes_normal: comp:0 0x4 no bytes to copy, available samples: 0,
````

**Review by kv2019i — APPROVED** [review]
Thanks @RanderWang , this can help potentially in other cases as well. I don't think this was the original intent of the change, but this will delay any IPC that requires to be handled in pipeline context, by 10ms.

**Comment by kv2019i** [comment]
Note, the original patch has been referred to fix https://github.com/thesofproject/sof/issues/7774 so we need to track test results carefully.

**Comment by RanderWang** [comment]
> Note, the original patch has been referred to fix #7774 so we need to track test results carefully.

As I was working on the issue #7774 at that time, I tested the original patch but made no effect.

**Comment by RanderWang** [comment]
#7774 should be fixed by https://github.com/thesofproject/sof/pull/8143

File: src/ipc/ipc4/handler.c [hunk 1]
```diff
@@ -510,13 +510,12 @@ static void ipc_compound_msg_done(uint32_t msg_id, int error)
...
 
-/* wait for IPCs to complete on other cores and be nice to any LL work */
 static int ipc_wait_for_compound_msg(void)
 {
-	int try_count = 30; /* timeout out is 30 x 10ms so 300ms for IPC */
+	int try_count = 30;
 
 	while (atomic_read(&msg_data.delayed_reply)) {
-		k_sleep(Z_TIMEOUT_MS(10));
+		k_sleep(Z_TIMEOUT_US(250));
```
btian1, line 518: btian1 

250 --> 10ms? it is 40 times enlarge? may miss something during sleep? how about 1ms?

above is the comments I added when the change first submitted, my question is: how 250us * 30 = 7.5ms comes from?

is it based on experience value to make total wait time < 10ms?

RanderWang: it is set by experiments with test and log. Most waiting time is less than 1ms.  I don't want to use 1ms since it will miss one schedule period in FW.

lgirdwood: What context is this called from ? My expectation is the IPC thread and if so not sure how we can block LL from running ?

kv2019i: @lgirdwood This is the IPC thread. It doesn't block the LL thread, but it will block processing of IPC messages until the LL timer fires next time, runs the IPC processing (that needs to happen in LL context), acks it as done, and then the IPC thread can continue. Now this adds a mandatory 10ms delay to each such handover from IPC thread to LL thread.

lgirdwood: ok, so we have a timing dependency between LL thread and IPC thread and this is convergence. This need to be spelled out in the inline comments alongside setting a better tries count as 30 means about 7.5 LL ticks. i.e. we need to say > 1.5 LL ticks would be a significant failure

btian1: @kv2019i , @RanderWang  is it better if 150 * 50? to make IPC response more quick?

kv2019i: @lgirdwood This was the mechanism added by @lyakh to synchronize pipeline trigger processing with LL thread (so that pipeline state changes are not done async with the pipeline exection). @lgirdwood @btian1 We can improve the logic (or even devise a different way to wait for next ll tick, I think the original concern of trying to avoid unnecessary thread context switches is as valid as ever), but can we do that as a follow-up and merge this plain revert first? This helps with multiple pressing issues (AV sync and this seems to help with multicore scenarios), and we have done significant testing of this version with @RanderWang . If we want a modified new mechanism, that needs to be tested well, pushing the merge out.

serhiy-katsyuba-intel (approved):

src/ipc/ipc4/handler.c
lgirdwood: I'm fine with a revert, but we need a folloup PR that spells out the relationships with some inline commentary.

lgirdwood (approved):

kv2019i:
One fail in https://sof-ci.01.org/sofpr/PR8756/build1997/devicetest/index.html?model=MTLP_RVP_NOCODEC&testcase=multiple-pause-resume-50 -- no FW/drv errors in the logs. This looks like a case of https://github.com/thesofproject/sof/issues/8770 . 

I'll proceed wit merge

marc-hb:
According to @keqiaozhang this revert causes the following regression:
- #8801

PR #8741, opened 2024-01-15
Add library authentication feature for MTL and LNL

jxstelter:
These changes will enable loadable library authentication before load.
Untrusted libraries will be not loaded.
Feature is available only on Intel platforms since it uses rom_ext module interface to
perform validation.

Signed-off-by: Jaroslaw Stelter [Jaroslaw.Stelter@intel.com](mailto:Jaroslaw.Stelter@intel.com)

PR #8759, opened 2024-01-18
1st part of llext #8180

lyakh:
7 first commits from #8180 not requiring a Zephyr update


marc-hb:
Thanks for this smaller PR. Still a draft?

src/include/sof/lib_manager.h
```diff
@@ -81,7 +81,9 @@ struct ipc_lib_msg {
 struct ext_library {
 	struct k_spinlock lock;	/* last locking CPU record */
 	struct sof_man_fw_desc *desc[LIB_MANAGER_MAX_LIBS];
+#ifdef __SOF_MODULE_SERVICE_BUILD__
```
kv2019i, line 86: This is first time this define is used in basefw build (this is defined when building modules with lmdk but not set for base fw build). I think this has to be a Kconfig for the main build to select loadable module behaviour and/or support both.

kv2019i (commented):
Lot of good changes, thanks for splitting these out. One question about the config define.

src/include/sof/lib_manager.h
lyakh: @kv2019i right, yes, I agree that ideally we should be able to build firmware that can load all kinds of modules: IADK, native system-services, and native dynamic. But initially I'd integrate all three approaches into the firmware and then check how we can identify modules at run-time

lyakh: in fact indeed it isn't correct. `-D__SOF_MODULE_SERVICE_BUILD__=1` is only used to build the **module** itself, not the complete firmware. Converted to draft for a fix

src/library_manager/lib_manager.c
```diff
@@ -47,6 +47,42 @@ struct lib_manager_dma_ext {
 
 static struct ext_library loader_ext_lib;
 
+struct lib_manager_mod_ctx {
+	struct sof_man_fw_desc *desc;
+	size_t segment_size[3];
```
kv2019i, line 52: Minor: The magic 3 sticks out a but, some SOF_MAN_SEGMENT_COUNT ?

kv2019i (approved):
Thanks, looks good now!

src/library_manager/lib_manager.c
```diff
@@ -155,23 +191,20 @@ static void lib_manager_unload_aux(struct sof_man_module *mod,
...
+	size_t st_text_size = ctx->segment_size[SOF_MAN_SEGMENT_TEXT];
 	void __sparse_cache *va_base_rodata = (void __sparse_cache *)
 		mod->segment[SOF_MAN_SEGMENT_RODATA].v_base_addr;
-	void *src_rodata =
-		(void *)(mod->segment[SOF_MAN_SEGMENT_RODATA].file_offset + load_offset);
-	size_t st_rodata_size = mod->segment[SOF_MAN_SEGMENT_RODATA].flags.r.length;
+	void *src_rodata = (void *)(load_base +
+				    mod->segment[SOF_MAN_SEGMENT_RODATA].file_offset);
+	size_t st_rodata_size = ctx->segment_size[SOF_MAN_SEGMENT_RODATA];
 	int ret;
```
jxstelter, line 207: Page alignment for sections is caused by secirity reasons. TEXT should be read_only while BSS read/write. Additionally if we want to introduce module level access rights, the PAGE size is minimal quantity we can control.

src/library_manager/lib_manager.c
lyakh: @kv2019i well, here's how they're defined https://github.com/thesofproject/sof/blob/d324af52bd3ceaf502afa15bef12023af595491e/tools/rimage/src/include/rimage/sof/user/manifest.h#L47-L51 and it is the same as https://github.com/thesofproject/sof/blob/d324af52bd3ceaf502afa15bef12023af595491e/tools/rimage/src/include/rimage/sof/user/manifest.h#L103 so, yes, we could add a separate commit to define them and use them in both those structures, but that would modify manifest.h. Even though it wouldn't *actually* change anything, still that probably would be considered an ABI change? Or now?

src/library_manager/lib_manager.c
```diff
@@ -249,19 +326,26 @@ uint32_t lib_manager_allocate_module(const struct comp_driver *drv,
...
 		return 0;
 	}
 
 	mod = (struct sof_man_module *)((char *)desc + SOF_MAN_MODULE_OFFSET(entry_index));
 
+	for (unsigned int i = 0; i < ARRAY_SIZE(ctx->segment_size); i++) {
+		ctx->segment_size[i] = mod->segment[i].flags.r.length * PAGE_SZ;
+
+		if (i == SOF_MAN_SEGMENT_BSS && mod->type.load_type != SOF_MAN_MOD_TYPE_LLEXT)
+			ctx->segment_size[i] /= mod->instance_max_count;
```
softwarecki, line 346: I would like move this if to `lib_manager_allocate_module_instance`

src/library_manager/lib_manager.c
```diff
@@ -110,79 +137,135 @@ static int lib_manager_load_module(uint32_t module_id, struct sof_man_module *mo
...
+						 SYS_MM_MEM_PERM_RW | SYS_MM_MEM_PERM_EXEC);
+	if (ret < 0)
+		return ret;
+
+	/* .text contains instructions and it also often contains local data */
+	dcache_writeback_region(va_base_text, st_text_size);
+	icache_invalidate_region(va_base_text, st_text_size);
+
+	/* Copy RODATA */
+	ret = lib_manager_load_data_from_storage(va_base_rodata, src_rodata,
```
softwarecki, line 211: Lack of support for readonly segment and empty data segment.

softwarecki (changes requested):

src/library_manager/lib_manager.c
```diff
@@ -110,79 +137,135 @@ static int lib_manager_load_module(uint32_t module_id, struct sof_man_module *mo
...
 		for (size_t idx = 0; idx < desc->header.num_module_entries;
 				++idx, ++module_entry) {
 			if (module_entry->type.lib_code) {
-				ret = lib_manager_load_module(lib_id << LIB_MANAGER_LIB_ID_SHIFT |
-							      idx, mod, desc);
+				int ret = lib_manager_load_module(lib_id << LIB_MANAGER_LIB_ID_SHIFT |
+								  idx, mod, desc);
 				if (ret < 0)
-					goto err;
+					return ret;
```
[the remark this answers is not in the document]
lyakh: @softwarecki It should be freed, yes, but I just moved the code from below. And I think some other PRs will rewrite this anyway. So I just moved it out of the way for now to let those other PRs work on this.

src/library_manager/lib_manager.c
lyakh: @softwarecki sorry, is this something that this commit changes? That's how it is now, isn't it?

src/library_manager/lib_manager.c
lyakh: @softwarecki I guess this could be done, and it wouldn't affect the LLEXT flow anyway, but I think it's more robust to have final sizes in this table and to be able to use them with no further modifications wherever needed?

src/library_manager/lib_manager.c
```diff
@@ -110,79 +137,135 @@ static int lib_manager_load_module(uint32_t module_id, struct sof_man_module *mo
...
+	icache_invalidate_region(va_base_text, st_text_size);
+
+	/* Copy RODATA */
+	ret = lib_manager_load_data_from_storage(va_base_rodata, src_rodata,
+						 st_rodata_size, SYS_MM_MEM_PERM_RW);
+	if (ret < 0)
+		goto e_text;
+
+	/* Some data can be accessed as uncached, in fact that's the default */
+	dcache_writeback_region(va_base_rodata, st_rodata_size);
```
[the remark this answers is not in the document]
lyakh: @softwarecki I'd rather not do that. As you see there's a difference between .text and .data - we need to additionally invalidate instruction cache for .text. If I only move this write-back call into the mapping function, it will become less logical - for .text the writeback will be inside the function and the icache invalidation will be outside

lyakh:
CI: a suspend-resume test failed on TGL and a pause-resume-capture failed on MTL, none should be related

src/library_manager/lib_manager.c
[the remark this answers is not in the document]
lyakh: ...and yes, I meant #8544 by @softwarecki

src/library_manager/lib_manager.c
```diff
@@ -47,59 +47,86 @@ struct lib_manager_dma_ext {
...
+	uint32_t lib_id = LIB_MANAGER_GET_LIB_ID(module_id);
+	struct ext_library *_ext_lib = ext_lib_get();
+
+	return _ext_lib->desc[lib_id];
+}
+
+struct sof_man_fw_desc *lib_manager_get_library_module_desc(int module_id)
+{
+	struct lib_manager_mod_ctx *ctx = lib_manager_get_mod_ctx(module_id);
+	uint8_t *buffptr = (uint8_t *)(ctx ? ctx->desc : NULL);
```
softwarecki, line 66: superfluous double checks. Why not just `if (!ctx) return NULL;`?

pjdobrowolski:
I have a mixed feeling about that llext. I only can guess that llext means that it is loadable libraries external however I didn't see any draft of documentation whatsoever. Changes themselves are cosmetic or irrelevant from IADK/Native loadable modules perspective and I would like to review that PR properly. For me without any documentation it is no-go.

src/library_manager/lib_manager.c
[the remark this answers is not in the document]
softwarecki: In that case, let's not modify the code that will be changed by another PR. There is no point in introducing a merge conflicts in another PR.

src/library_manager/lib_manager.c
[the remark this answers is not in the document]
softwarecki: Why do you moved a call to `dcache_writeback_region` from the `lib_manager_load_data_from_storage` to this function? i'm ok with icache.

src/library_manager/lib_manager.c
pjdobrowolski: Is ctx->segment_size calculated value for module or is it size of loaded binary? Can you clarify it?

pjdobrowolski (dismissed):

src/library_manager/lib_manager.c
softwarecki: I would prefer all bss-related calculations to be in one place. In my opinion, the name and location (in ctx) of this array suggests storing values read from the manifest. While reviewing the changes in the `lib_manager_allocate_module_instance` function, I was surprised that the division by the number of instances disappeared.

softwarecki (changes requested):

src/library_manager/lib_manager.c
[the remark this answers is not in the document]
softwarecki: In that case, let's not modify the code that will be changed by another PR. There is no point in introducing a merge conflicts in another PR.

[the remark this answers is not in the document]
softwarecki: In that case, let's not modify the code that will be changed by another PR. There is no point in introducing a merge conflicts in another PR.

lyakh:
> I have a mixed feeling about that llext. I only can guess that llext means that it is loadable libraries external however I didn't see any draft of documentation whatsoever. Changes themselves are cosmetic or irrelevant from IADK/Native loadable modules perspective and I would like to review that PR properly. For me without any documentation it is no-go.

https://docs.zephyrproject.org/latest/services/llext/index.html

src/library_manager/lib_manager.c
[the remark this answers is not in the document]
lyakh: @softwarecki I cannot and shall not sit and wait because I know that somebody else's work is in progress. If your PR gets merged first - I'll update mine, if mine is merged first - you'll update yours

src/library_manager/lib_manager.c
lyakh: @softwarecki I wouldn't really call this superfluous. An alternative would look like
```
if (!ctx || !ctx->desc)
	return NULL;
uint8_t *buffptr = ctx->desc;
```
so you anyway have to check up to two conditions - just as well as now. I'd say it's a matter of preference

src/library_manager/lib_manager.c
lyakh: > Is ctx->segment_size calculated value for module or is it size of loaded binary? Can you clarify it?

@pjdobrowolski that should be rather obvious from the code, sorry. It's calculated from section sizes of the loaded binary.

lyakh: > I would prefer all bss-related calculations to be in one place. In my opinion, the name and location (in ctx) of this array suggests storing values read from the manifest. While reviewing the changes in the `lib_manager_allocate_module_instance` function, I was surprised that the division by the number of instances disappeared.

@softwarecki currently that division is performed in two locations: in `lib_manager_allocate_module_instance()` and in `lib_manager_free_module_instance()`, after this change it's done only in one location, I think it's better.

src/library_manager/lib_manager.c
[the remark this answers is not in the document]
lyakh: @softwarecki to make it logical - `lib_manager_load_data_from_storage()` allocates memory and copies whatever data it's asked to handle into it, it doesn't know cache requirements of that data, that is handled by the caller, because the caller knows what data that is. Alternatively we'd need to add some kind of flags to the function to let it know what cache handling it has to perform

tools/rimage/src/include/rimage/sof/user/manifest.h
```diff
@@ -25,6 +25,7 @@
 /* module type load type */
 #define SOF_MAN_MOD_TYPE_BUILTIN	0
 #define SOF_MAN_MOD_TYPE_MODULE		1
+#define SOF_MAN_MOD_TYPE_LLEXT		2
```
plbossart, line 28: maybe explain what LL and EXT stand for? it's not obvious to me. Low Latency, Loadable Library, External, Extension, Exploding,  who knows?

Acronyms are evil.

src/library_manager/lib_manager.c
```diff
@@ -69,6 +69,68 @@ static int lib_manager_load_data_from_storage(void __sparse_cache *vma, void *s_
...
+	return 0;
+}
+
+static void lib_manager_unload_aux(struct sof_man_module *mod,
+				   struct sof_man_fw_desc *desc, uint32_t lib_id)
+{
+	struct ext_library *ext_lib = ext_lib_get();
+
+	/* There are modules marked as lib_code. This is code shared between several modules inside
+	 * the library. Unload all lib_code modules with last none lib_code module unload.
```
plbossart, line 161: @ujfalusi didn't we make the assumption that a library contains ONE module?

IIRC the firmwares are identified with a UUID, not sure how the library is identified in Linux

src/library_manager/lib_manager.c
```diff
@@ -69,6 +69,68 @@ static int lib_manager_load_data_from_storage(void __sparse_cache *vma, void *s_
...
+	 * the library. Load all lib_code modules with first none lib_code module load.
+	 */
+	if (!mod->type.lib_code)
+		ext_lib->mods_exec_load_cnt++;
+
+	if (ext_lib->mods_exec_load_cnt == 1) {
+		struct sof_man_module *module_entry =
+		    (struct sof_man_module *)((char *)desc + SOF_MAN_MODULE_OFFSET(0));
+
+		for (size_t idx = 0; idx < desc->header.num_module_entries;
```
plbossart, line 141: why the size_t for an index incremented by one?

src/library_manager/lib_manager.c
```diff
@@ -229,7 +250,7 @@ static void __sparse_cache *lib_manager_get_instance_bss_address(uint32_t module
 static int lib_manager_allocate_module_instance(uint32_t module_id, uint32_t instance_id,
 						uint32_t is_pages, struct sof_man_module *mod)
 {
-	uint32_t bss_size =
+	size_t bss_size =
```
plbossart: could be a separate patch

plbossart (commented):
Minor comments from a completely clueless reviewer...

src/library_manager/lib_manager.c
lyakh: @plbossart no idea, not my code, moved it as is from the current implementation

src/library_manager/lib_manager.c
lyakh: @plbossart in an earlier version I removed this whole "lib_code" support because, well, it's unused so far and it's unclear if and when and how it would be used. I got complains. So, I keep it as is, just move out to functions

pjdobrowolski:
> > I have a mixed feeling about that llext. I only can guess that llext means that it is loadable libraries external however I didn't see any draft of documentation whatsoever. Changes themselves are cosmetic or irrelevant from IADK/Native loadable modules perspective and I would like to review that PR properly. For me without any documentation it is no-go.
> 
> https://docs.zephyrproject.org/latest/services/llext/index.html

It is just explanation what is llext, there is no sofdocs of its implementation in sof.

src/library_manager/lib_manager.c
```diff
@@ -554,9 +619,6 @@ static void __sparse_cache *lib_manager_allocate_store_mem(uint32_t size,
 		return NULL;
 	}
 
-	dcache_invalidate_region(local_add, size);
-	icache_invalidate_region(local_add, size);
-
 	return local_add;
 }
```
pjdobrowolski, line 624: lib_manager.c is slowly hitting 1k lines of code. It is hard to understand what and where is jumping, especially when new functions were introduce. Is it a good time to split lib_manager.c for more readable format?

lyakh:
> > > I have a mixed feeling about that llext. I only can guess that llext means that it is loadable libraries external however I didn't see any draft of documentation whatsoever. Changes themselves are cosmetic or irrelevant from IADK/Native loadable modules perspective and I would like to review that PR properly. For me without any documentation it is no-go.
> > 
> > 
> > https://docs.zephyrproject.org/latest/services/llext/index.html
> 
> It is just explanation what is llext, there is no sofdocs of its implementation in sof.

@pjdobrowolski SOF uses many Zephyr APIs: threads, mutexes, work queues, timers, etc., we don't have sofdocs for each of them. But sure, we can add it as a paragraph in SOF loadable module implementation. In short - we use LLEXT for ELF parsing and linking. Also keep in mind, that this PR is just a preparation, it doesn't use LLEXT yet

src/library_manager/lib_manager.c
ujfalusi: No, the kernel side does not assume anything, it is prepared for the module bundles. The kernel tracks the library itself and then checks the module(s) inside:
b700672e2250 ("ASoC: SOF: Intel/IPC4: Support for external firmware libraries")

The lib_code does not have UUID, but in case a library have multiple modules then there should be symlinks:
```
module_bundle.bin
<UUID1>.bin -> module_bundle.bin
<UUID2>.bin -> module_bundle.bin
<UUID3>.bin -> module_bundle.bin
```
The kernel will load a library for a specific module UUID, so multiple UUID pointing to the same library will load that library with all of it's modules once.

tools/rimage/src/include/rimage/sof/user/manifest.h
ujfalusi: this conversation made me curious, what is llext stands for? ;) @lyakh

marc-hb: https://www.bing.com/search?q=zephyr+llext

lyakh:
New version. I'm forking lib_manager.c for LLEXT. Sharing it with other loadable module implementations didn't work out. @pjdobrowolski @softwarecki you're personally cordially invited to review this version.

tools/rimage/src/include/rimage/sof/user/manifest.h
kv2019i: But it's a fair point this file has nothing to do with Zephyr, so it's not so obvious to figure out this is a term coming from Zephyr.

src/library_manager/llext_manager.c
```diff
@@ -0,0 +1,241 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// Copyright(c) 2022 Intel Corporation. All rights reserved.
+//
+// Author: Jaroslaw Stelter <jaroslaw.stelter@intel.com>
+//         Pawel Dobrowolski<pawelx.dobrowolski@intel.com>
+
+/*
+ * Dynamic module loading functions.
```
kv2019i, line 9: So this could be more explict and mention "Dynamic module loading functions using Zephyr Linkable Loadable Extensions (LLEXT) interface".

kv2019i (approved):
Some minor things, but nothing blocking merge. I do wonder how we share https://github.com/thesofproject/sof/pull/8741 -- that's needed here as well, right?

lyakh:
> Some minor things, but nothing blocking merge. I do wonder how we share #8741 -- that's needed here as well, right?

@kv2019i it's not *needed* for this to work, it's needed for SOF to make sure we check what we run, right? And yes, it should work with this PR too, since auth is added to the IPC copying path, which is shared with system-services

src/library_manager/llext_manager.c
lyakh: @kv2019i I propose to address any cosmetic improvements in a follow-up PR, will do one first thing after this is merged

lyakh:
CI failures: a single known multiple-pause-resume failure https://sof-ci.01.org/sofpr/PR8759/build2309/devicetest/index.html?model=TGLU_RVP_NOCODEC-ipc4&testcase=multiple-pause-resume-50
zmain builds fail due to a Zephyr "main" regression
checkpatch complains about `extern` in a C file and a missing SPDX header in a .toml.h, both can be ignored.

kv2019i:
We got the auth bit merged, and now this can go it. @pjdobrowolski @softwarecki you didn't respond the pings this week but Guennadi has now addressed the main concerns and the code is now in separate file, so I assume we are good now. The documentation bit definitely needs to be supported but I think the Zephyr llext docs plus #8180 gives a pretty good view of what this is -- Linux-kernel style loadbale modules support using Zephyr llext mechanism. This won't be for all purposes, but is one option to build modules.

Issue #7593, opened 2023-05-11
[FEATURE] stop using the internal kernel API on the SOF side when enabling secondary cores.


tmleman:
SOF should use the public zephyr API for power management of CPU cores. The currently used functions are internal kernel functions and should not be used from within the application. The code in file [zephyr/lib/cpu.c](https://github.com/thesofproject/sof/blob/main/zephyr/lib/cpu.c) needs to be refactored.

- [x] https://github.com/zephyrproject-rtos/zephyr/pull/65394

- [x]  https://github.com/zephyrproject-rtos/zephyr/pull/64755

- [x] https://github.com/zephyrproject-rtos/sof/pull/32

EDIT updated 4th Dec with latest tracking.

lgirdwood:
@tmleman @kv2019i ping, any update ?

RanderWang:
@lgirdwood  @kv2019i  @tmleman  I investigate this issue. Let's first review our last state:

[kv2019i](https://github.com/kv2019i) [Apr 28, 2023](https://github.com/thesofproject/sof/pull/7174#discussion_r1180209343)

Given this is an open-coded of z_smp_start_cpu() (see comment above), I don't fully understand why we have a different implementation here, and if this is correct, why this is not needed in the Zephyr kernel/smp.c version. Wouldn't the same problem be there? Or is this specific to a configuration where state is saved to persistant memory? If yes, this would need to be ifdef'ed out so this applies only to applicable platforms.

[tmleman](https://github.com/tmleman) [May 8, 2023](https://github.com/thesofproject/sof/pull/7174#discussion_r1187424001)

I don't know how zephyr is handling it but right now it's not working properly. After secondary core exits D3 it will repeat the path that led him to this state. Power will not be shut off so eventually it will return to the Idle state but this is not correct behavior. Maybe zephyr doesn't support situation in which each core can be turned off and on multiple times.

My conclusion:
we define a private version of z_smp_start_cpu for two reasons:
(1) we need a point to initialize sof environment for secondary core after it boots up. Now we do it by our own callback function secondary_init which is called immediately after core boot up. For z_smp_start_cpu, it uses its own function: smp_init_top and we can't redefine it.  Actually we can initialize sof environment in callback function :cpu_notify_state_entry or in primary core boot up stage since these initialization only use get_cpu_id() to init some data array then we can only init one time

[...]

PR #8760, opened 2024-01-18
audio: dai-zephyr: update channel verification logic

dnikodem:
This commit updates the channel verification logic in dai-zephyr.c. The previous condition checked for an exact match between hw_params.channels and params->channels.

The new condition allows for hw_params.channels to be greater than or equal to params->channels, providing more flexibility in channel configuration.

### DIFF
--- a/src/audio/dai-zephyr.c
+++ b/src/audio/dai-zephyr.c
@@ -553,7 +553,7 @@ static int dai_verify_params(struct dai_data *dd, struct comp_dev *dev,
 		return -EINVAL;
 	}
 
-	if (hw_params.channels && hw_params.channels != params->channels) {
+	if (hw_params.channels && hw_params.channels < params->channels) {
 		comp_err(dev, "dai_verify_params(): pcm channels parameter %d does not match hardware channels %d",
 			 params->channels, hw_params.channels);
 		return -EINVAL;


src/audio/dai-zephyr.c
```diff
@@ -553,7 +553,7 @@ static int dai_verify_params(struct dai_data *dd, struct comp_dev *dev,
 		return -EINVAL;
 	}
 
-	if (hw_params.channels && hw_params.channels != params->channels) {
+	if (hw_params.channels && !(hw_params.channels >= params->channels)) {
```
btian1, line 556: 1. if this is the logic, why not:
    (hw_params.channels && (hw_params.channels < params->channels))

2. please also update the err log message, it is still use "match" words.

dnikodem: Thanks for feedback.

1. I agree, it does look better with your version, thank you.

2. In my opinion, the log is appropriate. It specifies that the pcm channels value is incorrect. We could, of course, add a note that the pcm channels value should be less than or equal to the hardware channels value, but this would make the log much longer. At this point, both values are printed, so they should be useful during debugging.
Of course, if you have a suggestion for presenting this log differently, I am open to it :)

tmleman (approved):

btian1 (approved):

dbaluta (approved):

dbaluta:
There are some build failures. Can you please check?

dnikodem:
> There are some build failures. Can you please check?
> 
> https://github.com/thesofproject/sof/actions/runs/7581488658/job/20649189700?pr=8760

I've noticed that the Zephyr / build-linux for (zmain, lnl/tgl/mtl) is failing on various PRs as well – the reason is unclear. Regarding CI, I am awaiting a rerun.

kv2019i:
@dbaluta  @dnikodem The mainlin Zephyr fails are known and related to this interface change (https://github.com/zephyrproject-rtos/sof/pull/32 , https://github.com/thesofproject/sof/issues/7593 and https://github.com/zephyrproject-rtos/zephyr/pull/64755 ).

src/audio/dai-zephyr.c
```diff
@@ -553,7 +553,7 @@ static int dai_verify_params(struct dai_data *dd, struct comp_dev *dev,
 		return -EINVAL;
 	}
 
-	if (hw_params.channels && hw_params.channels != params->channels) {
+	if (hw_params.channels && hw_params.channels < params->channels) {
```
kv2019i, line 556: Hmm, with some DAI implementations (like Intel PCH dmic), the channel count really is fixed, so the old check is correct. But it's true if this constrained cannot be assume to hold for all DAIs, we can enforce the condition here. DAI specific checks needs to return failure later.

kv2019i (changes requested):
Some concern how this impacts cases where the DAI really only supports a specific fixed channel count. 

Could the existing mechanism to use 0 to signal that the DAI driver can support a variable property, be used here instead of this patch?

src/audio/dai-zephyr.c
lgirdwood: @dnikodem might be worth an inline comment here to explain. Also it might be worth a 
```c
switch (dai_type) {
case dynamic_dai:
   // compare using new flow
    break;
default:
  // compare using old flow for fixed dais.
   break;
}
```

dnikodem: Colleagues, after reading your comments, I'm not sure if my change is correct – I'm working on confirming my doubts. Thanks for the feedback

dnikodem:
After further analysis of the code and feedback from Liam and Kai, I have concluded that this condition may remain unchanged. Therefore, I am closing this pull request. Thank you.

PR #8747, opened 2024-01-17
west.yml: update zephyr to main branch commit 53c121381f1a

marc-hb:
530 Zephyr commits in total including:

```
4204ca9bcb3f ace: fix DSP panic during startup  (fixes c3a6274bf5e4)
d4b0273ab0c4 cmake: sparse.template: add COMMAND_ERROR_IS_FATAL
ca12fd13c6d3 xtensa: intel_adsp: fix a cache handling error
0ee1e28a2f5f xtensa: polish doxygen and add to missing doc
035c8d8ceb4b xtensa: remove sys_define_gpr_with_alias()
a64eec6aaeec xtensa: remove XTENSA_ERR_NORET
1c0c900cbbc3 intel_adsp: ace15: Enhance HST domain power-down sequence
c3a6274bf5e4 intel_adsp: ace: power: Prevent HST domain power gating
4aa0e7af6834 west: sign.py: add "REM" support to pass comments through cpp
a27e8f9a1976 west: sign.py: explain why `-P` is passed to cpp
2ee6c26d15a0 west: sign.py: rename new `generated/platf.toml` to `rimage_config.toml`
991a85361ee0 llext: use generic attributes instead of GCC's
5993192ec1ac llext: fix incorrect elf64_sym structure
```

PR #8764, opened 2024-01-22
[DNM] Zephyr smp rework

RanderWang:
Update secondary core setting based on zephyr smp changes https://github.com/zephyrproject-rtos/zephyr/pull/64755.

Waiting for : https://github.com/thesofproject/sof/pull/8747

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -43,7 +43,7 @@ manifest:
 
     - name: zephyr
       repo-path: zephyr
-      revision: d7af6f371034f31b9440b27c694b0be3c87491d3
+      revision: 6a0b1da158a4f8bc5f2ca5058637dce26a38660e
       remote: zephyrproject
 
       # Import some projects listed in zephyr/west.yml@revision
--- a/zephyr/lib/cpu.c
+++ b/zephyr/lib/cpu.c
@@ -19,6 +19,7 @@
 /* Zephyr includes */
 #include <version.h>
 #include <zephyr/kernel.h>
+#include <zephyr/kernel/smp.h>
 #include <zephyr/device.h>
 #include <zephyr/drivers/mm/mm_drv_intel_adsp_mtl_tlb.h>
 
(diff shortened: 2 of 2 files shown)


zephyr/lib/cpu.c
```diff
@@ -149,19 +123,13 @@ int cpu_enable_core(int id)
 	 * and the idle thread stack.
 	 */
 	if (pm_state_next_get(id)->state == PM_STATE_ACTIVE)
-		z_init_cpu(id);
+		cpu_resume = false;
```
lyakh, line 126: could just do
```
	if (pm_state_next_get(id)->state == PM_STATE_ACTIVE) {
		k_smp_cpu_start(id, secondary_init, NULL);
		return 0;
	}
```
here and avoid adding `cpu_resume` altogether.

RanderWang: need to consider #if ZEPHYR_VERSION(3, 0, 99) <= ZEPHYR_VERSION_CODE

RanderWang: thanks, updated

zephyr/lib/cpu.c
```diff
@@ -239,29 +206,24 @@ int cpu_enable_core(int id)
...
-
-	atomic_clear(&start_flag);
-	atomic_clear(&ready_flag);
-
-	arch_start_cpu(id, z_interrupt_stacks[id], CONFIG_ISR_STACK_SIZE,
-		       secondary_init, &start_flag);
-
-	while (!atomic_get(&ready_flag))
-		k_busy_wait(100);
+	/* During kernel initialization, the next pm state is set to ACTIVE. By checking this
```
kv2019i, line 213: Should we remove the ZEPHYR_VERSION check? Afer this change, we won't support older Zephyr 3.0 anymore in SOF main, so might as well drop this ifdef.

west.yml
```diff
@@ -43,7 +43,7 @@ manifest:
 
     - name: zephyr
       repo-path: zephyr
-      revision: d7af6f371034f31b9440b27c694b0be3c87491d3
+      revision: 04de43b1a119665532b6b2a0c6bbcbefd47eb7d6
```
kv2019i, line 46: For the mergeable version, this needs to be combined with the previous git commit. Otherwise git-bisect will be broken as the first commit cannot be compiled with old zephyr so this has to be a single git commit to change the API and modify west.yml.

kv2019i (changes requested):
@RanderWang can you check why TGL builds are failing? https://sof-ci.01.org/sofpr/PR8764/build2107/build/index.html

This shouldn't happen and now we don't get any test results for this version. If this would pass the CI, we could just merge this and drop the older Zephyr baseline update PR.

west.yml
RanderWang: thanks

zephyr/lib/cpu.c
RanderWang: sure, thanks

tmleman:
> QuickBuild test case pass, but got failed result, why ?

@RanderWang I've reviewed the last four builds for this pull request on QB. In the older ones, TGL doesn't build, but in the latest one, this issue is no longer present. However, there is a failure in the gain_quality test (also TGL). The log indicates that there was an issue in the final step during the audio quality verification. @wszypelt , could you check if this is a problem with the test?

PS. I've scheduled a rebuild and rerun of tests.

kv2019i (approved):
Mandatory tests are now passing. The MTL error is known, so only open now is the cavs2.5 fails and whether there is more of them with this update than we have in baseline (we do have fails with pause-resume in mainline as well).

UPDATE: e.g. in this recent PR test, all these cavs2.5 tests pass:
https://sof-ci.01.org/sofpr/PR8756/build1998/devicetest/index.html
Compared to this PR:
https://sof-ci.01.org/sofpr/PR8764/build2195/devicetest/index.html

west.yml
```diff
@@ -43,7 +43,7 @@ manifest:
 
     - name: zephyr
       repo-path: zephyr
-      revision: d7af6f371034f31b9440b27c694b0be3c87491d3
+      revision: 6a0b1da158a4f8bc5f2ca5058637dce26a38660e
```
lgirdwood, line 46: can you make this a separate patch for bisect-ability.

lgirdwood (commented):
LGTM, just need to split the west update into another patch.

west.yml
tmleman: I think that if we separate the west update into a separate commit, then we'll lose the ability to bisect because these changes won't compile without updating Zephyr.
@RanderWang  correct me if I'm wrong.

kv2019i: @lgirdwood west update needs to be in same commit, otherwise bisect is broken.

kv2019i:
Marking as DNM so this doesn't get accidentally merged. We need to understand first why we have so many fails in https://sof-ci.01.org/sofpr/PR8764/build2195/devicetest/index.html

kv2019i:
https://github.com/thesofproject/sof/issues/8792 filed for the chain-dma failures seen. I'll submit a fix soon to Zephyr.

kv2019i:
https://github.com/thesofproject/sof/pull/8790 submitted to fix the double-free.

PR #8769, opened 2024-01-22
zephyr: use arch_xtensa_uncached_ptr/arch_xtensa_cached_ptr

nashif:
Use direct call to function instead of going over a redefine.

Signed-off-by: Anas Nashif <anas.nashif@intel.com>

### DIFF
--- a/zephyr/lib/alloc.c
+++ b/zephyr/lib/alloc.c
@@ -116,7 +116,7 @@ static inline uintptr_t get_l3_heap_start(void)
 	 * - main_fw_load_offset
 	 * - main fw size in manifest
 	 */
-	return (uintptr_t)z_soc_uncached_ptr((__sparse_force void __sparse_cache *)
+	return (uintptr_t)arch_xtensa_uncached_ptr((__sparse_force void __sparse_cache *)
 					     ROUND_UP(IMR_L3_HEAP_BASE, L3_MEM_PAGE_SIZE));
 }
 
@@ -146,7 +146,7 @@ static bool is_l3_heap_pointer(void *ptr)
 	uintptr_t l3_heap_end = l3_heap_start + get_l3_heap_size();
 
 	if (is_cached(ptr))
-		ptr = z_soc_uncached_ptr((__sparse_force void __sparse_cache *)ptr);
+		ptr = arch_xtensa_uncached_ptr((__sparse_force void __sparse_cache *)ptr);
 
 	if ((POINTER_TO_UINT(ptr) >= l3_heap_start) && (POINTER_TO_UINT(ptr) < l3_heap_end))
 		return true;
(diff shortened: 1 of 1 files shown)


zephyr/lib/alloc.c
```diff
@@ -116,7 +116,7 @@ static inline uintptr_t get_l3_heap_start(void)
 	 * - main_fw_load_offset
 	 * - main fw size in manifest
 	 */
-	return (uintptr_t)z_soc_uncached_ptr((__sparse_force void __sparse_cache *)
+	return (uintptr_t)arch_xtensa_uncached_ptr((__sparse_force void __sparse_cache *)
```
kv2019i, line 119: Hmm, this seems to go to the wrong direction, zephyr/lib/alloc.c is generic code that should work e.g. on x86 or arm. 

OTOH, many architectures will not support such cached/uncached mappings to same memory, so this will have to:
   - be no-op with architectures with only single addresss (but cache coherency between cores)
   - raise an error on architecture that are have no coherency between cores and do not allow uncached mapping

z_soc_uncached_ptr() seemd like a suitably generic name for such wrapping. Should we add a SOF abstraction for this after all?

kv2019i (changes requested):
This brings some fundamental questions on how we isolate xtensa-specific code, please see question inline.

dbaluta (approved):

dbaluta:
I think we are fine with this, but it won't compile on arm64. 

Cc: @LaurentiuM1234 I wonder why we havent added a build for imx93 here?

LaurentiuM1234:
> Cc: @LaurentiuM1234 I wonder why we havent added a build for imx93 here?

we do have that. It's just that the code being changed here is not used by 93 (because of some configs not being selected). At first  glance, this code shouldn't compile anyways on 93 because `z_soc_uncached_ptr()` is an Intel-only thing as far as I can tell.

zephyr/lib/alloc.c
lgirdwood: ack - we would need an ARCH generic Zephyr API for this since it lives in a generic file.

nashif: see https://github.com/zephyrproject-rtos/zephyr/pull/68140

lgirdwood: Thanks @nashif - once https://github.com/zephyrproject-rtos/zephyr/pull/68140 is merged, this PR can be updated to use it and we can merge as ARCH generic code.

lgirdwood: > Thanks @nashif - once [zephyrproject-rtos/zephyr#68140](https://github.com/zephyrproject-rtos/zephyr/pull/68140) is merged, this PR can be updated to use it and we can merge as ARCH generic code.

Ok, Zephyr PR is now merged. @nashif looks like we are good to go with update when you are ready.
@dbaluta this will become arch generic so should build for you on arm64 when the new update lands.

dbaluta (approved):

PR #8772, opened 2024-01-23
vmh: fix config pasing issue

dabekjakub:
Config was not correctly read by API. Fix config copying.


dabekjakub:
Cannot fix checkpatch in a way it would pass - left best possible and readable way.

zephyr/lib/regions_mm.c
```diff
@@ -69,23 +69,24 @@ struct vmh_heap *vmh_init_heap(const struct vmh_heap_config *cfg,
 		goto fail;
 
 	/* If no config was specified we will use default one */
-	if (!cfg) {
+	if (!cfg)
 		vmh_get_default_heap_config(new_heap->virtual_region, &new_config);
-		cfg = &new_config;
-	}
+	else
+		memcpy_s(&new_config, sizeof(struct vmh_heap_config),
```
lyakh, line 75: `sizeof(new_config)`

lyakh (commented):
sorry, don't understand what this is fixing. What exactly was wrong with the existing version?

zephyr/lib/regions_mm.c
dabekjakub: Old code resulted in the following behavior:
cfg was read correctly only in the if context and reads done after that returned garbage.
I do not understand the origin of the issue but sticking to new_config which was created and zeroed in the function body seems to fix the issue.

lyakh: > Old code resulted in the following behavior: cfg was read correctly only in the if context and reads done after that returned garbage. I do not understand the origin of the issue but sticking to new_config which was created and zeroed in the function body seems to fix the issue.

@dabekjakub sorry, I have a substantial issue with fixing something that we don't understand, especially if that's our own open-source software, not some 3rd-party black box.

dabekjakub: Well, I agree but it is either this or the code not working. Which do we choose? This can be investigated further but this fixes heap creation. 

Both ptr should be stack ptrs since cfg should be created runtime and not assigned space on the heap. At least that is an assumption I am making it is a client that will decide from where he will pass config.

marc-hb:
> Cannot fix checkpatch in a way it would pass - left best possible and readable way.

https://www.kernel.org/doc/html/latest/dev-tools/checkpatch.html

> Checkpatch is not always right. Your judgement takes precedence over checkpatch messages. If your code looks better with the violations, then its probably best left alone.


> line length of 102 exceeds 100 columns

Looks good enough to me.

dabekjakub:
@lyakh Was right here. After another debug session narrowed it down to testing issue. Closing.

PR #8774, opened 2024-01-23
Audio: IGO_NR: Fix build fail variable size data structure

singalsu:
This change avoids error, seen with testbench build for MTL platform with "scripts/rebuild-testbench.sh -p mtl":

sof/src/include/sof/audio/igo_nr/igo_nr_comp.h:29:27: error: field 'config' with variable sized type
'struct sof_igo_nr_config' not at the end of a struct or class is a GNU extension [-Werror,-Wgnu-variable-sized-type-not-at-end]

### DIFF
--- a/src/include/sof/audio/igo_nr/igo_nr_comp.h
+++ b/src/include/sof/audio/igo_nr/igo_nr_comp.h
@@ -26,7 +26,6 @@ struct comp_data {
 	struct IgoStreamData igo_stream_data_ref;
 	struct IgoStreamData igo_stream_data_out;
 	struct comp_data_blob_handler *model_handler;
-	struct sof_igo_nr_config config;	    /**< blob data buffer */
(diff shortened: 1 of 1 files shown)


singalsu:
Note: There is no usage in the open-source parts of component for component data member config.data so it's not known if this could impact operation. The component data in init() is allocated only for size of struct comp_data, nothing more for flexible array member.

If it would feel more safe to remove "int16_t data[];" it would be another way to fix this. Build fail is anyway a blocker for xt-testbench, so need to fix this or disable build from kconfig.

lgirdwood (approved):
Ok, good catch for host GCC.

btian1 (approved):

lgirdwood:
@wszypelt @lrudyX good to merge, dont think this is tested by internal CI today ?

wszypelt:
@lgirdwood good to merge

PR #8775, opened 2024-01-23
Virtual Memory Heaps boot testing improvements

dabekjakub:
Change boot test trigger from ipc to zephyr API call.
Change existing vmh tests and expand it.

### DIFF
--- a/zephyr/boot_test.c
+++ b/zephyr/boot_test.c
@@ -4,9 +4,17 @@
  */
 
 #include <zephyr/logging/log.h>
+#include <sof/boot_test.h>
 
 #include <zephyr/ztest.h>
 
 LOG_MODULE_REGISTER(sof_boot_test, LOG_LEVEL_DBG);
 
 ZTEST_SUITE(sof_boot, NULL, NULL, NULL, NULL, NULL);
+
+void sys_run_boot_tests(void)
+{
+	ztest_run_all(NULL, false, 1, 1);
+}
+SYS_INIT(sys_run_boot_tests, APPLICATION, 99);
+
--- a/zephyr/test/vmh.c
+++ b/zephyr/test/vmh.c
@@ -17,81 +17,8 @@
 
 LOG_MODULE_DECLARE(sof_boot_test, CONFIG_SOF_LOG_LEVEL);
 
-#define ALLOC_SIZE1 1616
-#define ALLOC_SIZE2 26
-
-static int vmh_test_single(bool span)
-{
-	struct vmh_heap *h = vmh_init_heap(NULL, MEM_REG_ATTR_CORE_HEAP, 0, span);
-
-	if (!h)
-		return -EINVAL;
-
-	char *buf = vmh_alloc(h, ALLOC_SIZE1);
-	int ret1;
-
-	if (buf) {
-		buf[0] = 0;
-		buf[ALLOC_SIZE1 - 1] = 15;
-
-		ret1 = vmh_free(h, buf);
-		if (ret1 < 0)
-			goto out;
-	} else if (span) {
-		ret1 = -ENOMEM;
-		LOG_ERR("Failed to allocate %u in contiguous mode", ALLOC_SIZE1);
-		goto out;
-	} else {
-		LOG_WRN("Ignoring failure to allocate %u in non-contiguous mode",
-			ALLOC_SIZE1);
-	}
-
-	buf = vmh_alloc(h, ALLOC_SIZE2);
-
-	if (!buf) {
-		ret1 = -ENOMEM;
-		LOG_ERR("Failed to allocate %u", ALLOC_SIZE2);
-		goto out;
-	}
-
-	buf[0] = 0;
-	buf[ALLOC_SIZE2 - 1] = 15;
-
-	ret1 = vmh_free(h, buf);
-	if (ret1 < 0)
-		LOG_ERR("Free error %d", ret1);
-
-out:
-	int ret2 = vmh_free_heap(h);
-
-	if (ret2 < 0)
-		LOG_ERR("Free heap error %d", ret2);
-
-	if (!ret1)
-		ret1 = ret2;
-
-	return ret1;
-}
-
-static int vmh_test(void)
-{
-	int ret = vmh_test_single(false);
-
-	if (ret < 0) {
-		LOG_ERR("Non-contiguous test error %d", ret);
-		return ret;
-	}
-
-	ret = vmh_test_single(true);
-	if (ret < 0)
-		LOG_ERR("Contiguous test error %d", ret);
-
-	return ret;
-}
 
 ZTEST(sof_boot, virtual_memory_heap)
 {
-	int ret = vmh_test();
-
-	TEST_CHECK_RET(ret, "virtual_memory_heap");
+	TEST_CHECK_RET(true, "virtual_memory_heap");
 }
--- a/zephyr/test/vmh.c
+++ b/zephyr/test/vmh.c
@@ -7,6 +7,7 @@
 
 #include <errno.h>
 #include <stdbool.h>
+#include <stdlib.h>
 
 #include <adsp_memory_regions.h>
 #include <sof/boot_test.h>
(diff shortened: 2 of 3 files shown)


zephyr/test/vmh.c
```diff
@@ -1,12 +1,13 @@
 // SPDX-License-Identifier: BSD-3-Clause
 /*
- * Copyright(c) 2023 Intel Corporation. All rights reserved.
- *
- * Author: Guennadi Liakhovetski <guennadi.liakhovetski@linux.intel.com>
- */
+* Copyright(c) 2023 Intel Corporation. All rights reserved.
+*
+* Author: Guennadi Liakhovetski <guennadi.liakhovetski@linux.intel.com>
+*/
```
lyakh, line 6: the indentation was correct, no need to change

zephyr/test/vmh.c
```diff
@@ -17,8 +18,258 @@
...
+		}
+	else
+		zassert_is_null(heap, "Heap initialization expected to fail but succeeded");
+
+	if (heap) {
+		int ret = vmh_free_heap(heap);
+
+		zassert_equal(ret, 0, "Failed to free heap");
+	}
+}
```
lyakh, line 42: can we have all this as a change to the current version instead of a rewrite?  It would be much easier to review changes than complete new code.

zephyr/test/vmh.c
```diff
@@ -17,8 +18,258 @@
...
+					bool expect_success)
+{
+	void *ptr = vmh_alloc(heap, alloc_size);
+
+	if (expect_success)
+		zassert_not_null(ptr, "Allocation expected to succeed but failed");
+	else {
+		zassert_is_null(ptr, "Allocation expected to fail but succeeded");
+		return;
+	}
```
lyakh, line 75: no. I don't think you need a `return` after an assert, asserts **are** checks themselves, so all this could be a single line like
```
__ASSERT(expect_success == !!ptr, "...");
```

lyakh (commented):
would be better as a change, not a rewrite

zephyr/test/vmh.c
```diff
@@ -17,8 +18,258 @@
...
 ZTEST(sof_boot, virtual_memory_heap)
 {
+	test_heap_creation();
+	test_vmh_init_all_heaps();
+	test_alloc_on_configured_heap(true);
+	test_alloc_on_configured_heap(false);
+	test_vmh_alloc_free(true);
+	test_vmh_alloc_free(false);
+	test_vmh_alloc_multiple_times(true);
+	test_vmh_alloc_multiple_times(false);
```
lyakh, line 272: This is quite a bit of testing, taking into account that many of these functions test multiple allocations too. Do we know how long this takes? I'm not sure we need that much testing - I'd basically test each (or most) configuration once - once in contiguous mode, once without, once with success, once with failure. Something rather lightweight for regular debug-overlay testing, possibly add a new Kconfig for extensive VMH testing. So do all these tests now pass? Can we re-enable boot-time testing by default?

zephyr/test/vmh.c
```diff
@@ -17,8 +18,258 @@
...
+	/* Try to setup with pre defined heap config */
+	struct vmh_heap_config config;
+
+	config.block_bundles_table[0].block_size = 4;
+
+	config.block_bundles_table[0].number_of_blocks = 80;
+
+	config.block_bundles_table[1].block_size = 4;
+
+	config.block_bundles_table[1].number_of_blocks = 80;
```
lyakh, line 214: if you're already testing 2 bundles, maybe configure them differently too?

zephyr/test/vmh.c
```diff
@@ -17,8 +18,258 @@
...
+	/* Test initializing all types of heaps */
+	for (i = 0; i < num_regions; i++) {
+
+		/* Test without continuous allocation */
+		test_vmh_init_and_free_heap(virtual_memory_region[i].attr, NULL, 0, false,
+		true);
+
+		/* Test if it fails when heap already exists */
+		test_vmh_init_and_free_heap(virtual_memory_region[i].attr, NULL, 0, true,
+		false);
```
lyakh, line 259: I don't see a single contiguous heap test which is supposed to succeed. In fact why isn't this supposed to succeed? The comment says "already exists," but the test frees the heap internally?

zephyr/test/vmh.c
```diff
@@ -17,8 +18,258 @@
...
+		zassert_true(ptrs[i] != NULL,
+		"Allocation of size %u expected to succeed but failed",
+		alloc_size);
+
+	if (ptrs[i] != NULL) {
+		LOG_INF("Test allocation size: %d", alloc_size);
+
+		memset(ptrs[i], 0xAA, alloc_size);
+
+
```
lyakh, line 132: one empty line

zephyr/test/vmh.c
```diff
@@ -17,8 +18,258 @@
...
+		uint8_t *byte_ptr = (uint8_t *)ptrs[i];
+
+		for (uint32_t j = 0; j < alloc_size; j++) {
+			zassert_equal(byte_ptr[j], 0xAA,
+			"Memory content verification failed at position %u", j);
```
lyakh, line 137: I *really* don't think we should check every byte in every allocation in multiple allocations in multiple tests in multiple heaps

lyakh (commented):
extensive testing is good, but I think this is taking it a bit too far. And *many* indentation and curly brace issues

zephyr/boot_test.c
```diff
@@ -4,9 +4,19 @@
...
 ZTEST_SUITE(sof_boot, NULL, NULL, NULL, NULL, NULL);
+
+#if CONFIG_SOF_BOOT_TEST
+void sys_run_boot_tests(void)
+{
+	TEST_RUN_ONCE(ztest_run_test_suite, sof_boot);
+}
+SYS_INIT(sys_run_boot_tests, APPLICATION, 99);
+
+#endif
```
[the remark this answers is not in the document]
dabekjakub:
```
SYS_INIT(sys_run_boot_tests, APPLICATION, 99);
```
99 here guarantees execution after everything else. Works in my testing.

zephyr/test/vmh.c
dabekjakub: clang formater automatically changed it - reverting

zephyr/test/vmh.c
dabekjakub: Old code was not reused in any way in the vmh.c file so the change would only confuse the reader who did not write this code before. The new approach enhances existing tests but does it in a parametrized fashion and I see no way to do it as a change in a readable way.
I specifically did it as a rewrite to make it clear in the review.

zephyr/test/vmh.c
dabekjakub: Also good catch on the bug in tests ;) block size can only be power of 2.

zephyr/test/vmh.c
dabekjakub: Overall I assumed that this boot test feature is not intended to be enabled in the productized version. So we will only use the CONFIG_SOF_BOOT_TEST flag for testing only and are not that much impacted by test time.

However, for allocator testing, this is quite concise. Should not take that much time - will have better data on it as soon as I clean up all the issues those tests found.

Amount of tests: This test approach allows for clear identification of the issues. I tried to add tests for each case I encountered. For example, multiple allocations failed in my internal testing during development so now I want it tested extensively. Writing data to the allocated space also checks for correctly assigned physical pages.

Can we enable boot testing? No - I am still working out the issues in vmh API #8772 Once this is done It can be pushed.

zephyr/test/vmh.c
dabekjakub: good catch need to rewrite this test

zephyr/test/vmh.c
dabekjakub: The problem I am trying to catch here is that allocation worked but physical pages were not mapped correctly. I am afraid I do not see any other option to be sure.
If our discussion ends with CONFIG_SOF_BOOT_TEST being time constrained we can think about timing it down. Let's come back to it after we clear that up in the comments above.

lgirdwood: I think the comments is possible this could be faster and more optimal with words instead of bytes ? or even use memcmp() now that we have a C library in Zephyr

zephyr/boot_test.c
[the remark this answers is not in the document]
lyakh: 99 doesn't *guarantee* that - others might also use 99 (even if it's the maximum accepted number there). And it's only the ordering of init calls, whereas in my testing something was missing, that was done after all the init calls. But if it's working now - that's good, something must've changed since I was testing it

zephyr/test/vmh.c
lyakh: Good, that this helps you find issues with the VMH! Maybe we should first have all the issues fixed and then merge and enable this?

zephyr/test/vmh.c
dabekjakub: I will try to rewrite the check to use bigger data, however, memcmp compares memory regions so it is not applicable if I have no buffer with the pattern.

zephyr/test/vmh.c
dabekjakub: I am working on it so I see no problem In pushing both fixes and this at the same time.
Linked pr is for fixes I will do.

lgirdwood (commented):
@dabekjakub is this bisect-able with the remove and add patches being different i.e. will it build/run ?
Also please state in commit message why new test are better etc, i.e. what are you adding that makes the test better.

zephyr/boot_test.c
```diff
@@ -4,9 +4,19 @@
...
 #include <zephyr/logging/log.h>
+#include <sof/boot_test.h>
 
 #include <zephyr/ztest.h>
 
 LOG_MODULE_REGISTER(sof_boot_test, LOG_LEVEL_DBG);
 
 ZTEST_SUITE(sof_boot, NULL, NULL, NULL, NULL, NULL);
+
+#if CONFIG_SOF_BOOT_TEST
```
lyakh, line 15: this isn't needed - this file isn't built if `CONFIG_SOF_BOOT_TEST` isn't selected

lgirdwood (commented):
@dabekjakub commit message should say why we remove old test (i.e. what the gaps were) and why the new tests are better (what we test now and not before etc).

dabekjakub:
@lgirdwood @lyakh After zephyr update there is a build error in the current implementation. 

I will update PR tomorrow with both review fixes and new zephyr ztest suite run implementation.

dabekjakub:
As soon as this and https://github.com/thesofproject/sof/pull/8849 is merged we can reenable boot tests. However, they do take a lot of time so if we enhance those we should think about separating it as a test-only build.

lyakh:
> However, they do take a lot of time so if we enhance those we should think about separating it as a test-only build.

@dabekjakub maybe we could have a define (Kconfig?) to select between a short and a long VMH test, a short one would skip most configurations and only test a couple of them

zephyr/test/vmh.c
lgirdwood: It does not matter how long the tests take to run as they should not be used in a production mode. i.e. we are not going to add latency to audio use case.
CI should be able to build this Kconfig and all tests should run for all subsystems.

lgirdwood (approved):

zephyr/test/vmh.c
kv2019i: I'd prefer to enable this in debug overlay, so this would get tested automatically with the current infra we have. That would require the tests to pass. Otherwise the code will very easily get broken (like it has when we temporarily disabled it in the debug overlay).

kv2019i (approved):
Given the existing code doesn't build anymore, I'm ok to merge this version. 

I do think we should enable this in some configuration CI tests (or add one, either in Intel internal/quickbuild). Otherwise the code will get broken and we won't know e.g. regressions with Zephyr. Having this enabled in CI would be very powerful e.g. to test Zephyr baseline updates.

PR #8767, opened 2024-01-22
CHAIN_DMA: fix memory leaks

plbossart:
The SoundWire BPT/BRA loop tests exposes a DMA buffer allocation issue, which was root-caused after 4 days of my life that I am not getting back to a 20-byte memory leak. After a while this causes the DMA buffer to fail, likely because of alignment/fragmentation.

This PR also suggests fixes for the XRUN_NOTIFICATIONS_ENABLE option, but since this Kconfig is not used it's anyone's guess if the code actually works. Our firmware overlords might decide to completely remove this bad case of bitrot.

PR #8780, opened 2024-01-23
chain DMA: remove redundant code

lyakh:
No need to re-calculate cdma_comp in ipc4_process_chain_dma(), keep using the already calculated value.

### DIFF
--- a/src/ipc/ipc4/handler.c
+++ b/src/ipc/ipc4/handler.c
@@ -715,10 +715,6 @@ static int ipc4_process_chain_dma(struct ipc4_message_request *ipc4)
 		return IPC4_SUCCESS;
 	}
 
-	cdma_comp = ipc_get_comp_by_id(ipc, comp_id);
-	if (!cdma_comp)
-		return IPC4_FAILURE;
-
 	ret = ipc4_chain_dma_state(cdma_comp->cd, &cdma);
 	if (ret < 0)
 		return IPC4_INVALID_CHAIN_STATE_TRANSITION;


plbossart (approved):

RanderWang (approved):

src/ipc/ipc4/handler.c
```diff
@@ -715,10 +715,6 @@ static int ipc4_process_chain_dma(struct ipc4_message_request *ipc4)
 		return IPC4_SUCCESS;
 	}
 
-	cdma_comp = ipc_get_comp_by_id(ipc, comp_id);
```
btian1, line 718: compare the if part, there are below difference:

1. comp_free(cdma_comp->cd);, this added in line 711, however, missed in following error handling.
2. can we confirm that ipc4_chain_dma_state need twice calling? if line 693 condition meet, if part will call this.
    and following line 722 will also call this, seems a little not so reasonable.

lyakh:
CI: a single unrelated suspend-resume failure on cAVS

src/ipc/ipc4/handler.c
lgirdwood: @lyakh is this resolved/valid ?

lgirdwood (commented):
LGTM, lets resolve the question.

src/ipc/ipc4/handler.c
lyakh: @lgirdwood sorry, I don't understand whether that's a comment or a question or a change request? @btian1 can you clarify?

lyakh: The only thing I can state here, is that I'm pretty sure that that code is unneeded, a repeated call to `ipc_get_comp_by_id()` there is unneeded and the repeated check is redundant

btian1: it is further ask for this function, simple for this patch, it is ok go for merge, I mean we can study and discuss here to add more patches if aligned, or you simply want to go with this single patch?
@RanderWang @ujfalusi please share your comments on below ask:

1. comp_free(cdma_comp->cd);, this added in line 711, however, missed in following error handling.
2. can we confirm that ipc4_chain_dma_state need twice calling? if line 693 condition meet, if part will call this.
and following line 722 will also call this, seems a little not so reasonable.

ujfalusi (approved):
@lyakh, this has been bugging me for some time, thanks for fixing it!

src/ipc/ipc4/handler.c
ujfalusi: @btian1, I agree that the error handling is wrong but it is not the topic of this PR.
Furthermore: a simple comp_free() is inadequate, it is leaking memory for ChainDMA, see: https://github.com/thesofproject/sof/pull/8767/commits/d2ffc0b227b5213acaa260402b4157833f50c214

We need a new wrapper to properly clean up after ipc4_chain_manager_create() and to be precise after  ipc4_add_comp_dev().

PR #8784, opened 2024-01-24
topology2: mtl: Add HDMI-in capture support for SSP-based no-headset …

CBala21:
…codec products.

Adding support for the product which doesn't have no ssp-based audio codec but need to support HDMI audio playback and HDMI-in capture via I2S.


lgirdwood (commented):
@CBala21 I guess this is a HW I2S/HDMI bridge ? - if so, do we need to spell out this as a special case in more inline comments. Thinking aloud, I would not want anyone to copy this example for native HDMI or native I2S as it could cause confusion ?

CBala21:
> @CBala21 I guess this is a HW I2S/HDMI bridge ? - if so, do we need to spell out this as a special case in more inline comments. Thinking aloud, I would not want anyone to copy this example for native HDMI or native I2S as it could cause confusion ?

>Its not HDMI bridge(converter chip will convert and send samples to I2S pins), for intel soc its just a input to I2S pins and similar as BT offload use case where we get audio samples via I2S pins. So we will handle only SSP configuration in slave mode.

The changes are added to generate the topology configuration for the product which doesn't have I2S audio codec but they need HDMI-In audio capture via I2S interface.

tools/topology/topology2/production/tplg-targets-ace1.cmake
```diff
@@ -70,7 +70,13 @@ GOOGLE_RTC_AEC_SUPPORT=1,DEEP_BUF_SPK=true,PLAYBACK_PIPELINE_SRC=dts"
...
 "cavs-es83x6\;sof-mtl-es83x6-ssp1-hdmi-ssp02\;PLATFORM=mtl,PREPROCESS_PLUGINS=nhlt,\
 NHLT_BIN=nhlt-sof-mtl-es83x6-ssp1-hdmi-ssp02.bin,HEADSET_SSP_DAI_INDEX=1,\
-HEADSET_CODEC_NAME=SSP1-Codec,HDMI_IN_CAPTURE=true"
+HEADSET_CODEC=true,HEADSET_CODEC_NAME=SSP1-Codec,HDMI1_ID=3,HDMI2_ID=4,HDMI3_ID=5,\
+HDMI_IN_CAPTURE=true"
+
+#No HeadsetCodec+HDMI-IN
+"cavs-es83x6\;sof-mtl-hdmi-ssp02\;PLATFORM=mtl,PREPROCESS_PLUGINS=nhlt,\
+NHLT_BIN=nhlt-sof-mtl-hdmi-ssp02.bin,HEADSET_CODEC=false,HDMI_IN_CAPTURE=true,\
+HDMI_IN_1_ID=0,HDMI_IN_2_ID=1"
```
bardliao, line 79: LGTM, but please split the commit to 1. add HEADSET_CODEC macro. 2. Add the sof-mtl-hdmi-ssp02 topology

CBala21: done

bardliao (approved):

lgirdwood (approved):

lgirdwood:
@plbossart any comments ?

plbossart (approved):

PR #8785, opened 2024-01-24
[DNM] PR8764 + Fix to module adapter free

kv2019i:
https://github.com/thesofproject/sof/pull/8764

plus fix identified by analyzing the PR failures and specifically this:
https://sof-ci.01.org/sofpr/PR8764/build2195/devicetest/index.html?model=TGLU_RVP_NOCODEC-ipc4&testcase=multiple-pause-resume-50

PR #8786, opened 2024-01-24
[DNM] PR8785 with more debug

kv2019i:
https://github.com/thesofproject/sof/pull/8785 + more debug

The chain-dma bugs are probably related to recent Zephyr patches, add debug to confirm (as I can't get these on local devices despite hours of testing -- same kernel, same tplg, same FW, but still fails).


kv2019i:
Closing, seems we have two different bugs that show up with new Zephyr. A chain-dma regression on Intel cAVS2.5 systems and existing problem with module-adapter gets triggered at higher rate (see https://github.com/thesofproject/sof/pull/8790 ).

PR #8789, opened 2024-01-25
topology2: enable HDMI for all MTL topologies

plbossart:
Remove the temporary workaround and add HDMI on all MeteorLake products. If this breaks stuff, well we need to fix hit ASAP.


plbossart:
@kv2019i probably need to add this to a 2.8.x release?

kv2019i (approved):
@plbossart I'll tag it with v2.9 to keep track of this. If we do 2.8.1. I'll cherry-pick this to it.

bardliao (approved):

kv2019i:
https://github.com/thesofproject/sof/issues/8056 was hit once, otherwise looks good:
https://sof-ci.01.org/sofpr/PR8789/build2267/devicetest/index.html?model=TGLU_RVP_SDW-ipc4&testcase=check-signal-stop-start-playback-10

Issue #8606, opened 2023-12-11
[FEATURE] remove duplicated CMake rules for audio modules


kv2019i:
**Is your feature request related to a problem? Please describe.**
The list of files to build is duplicated for the shared-library build. On time in top-level
src/audio/CMakeLists.txt and second time in subfolders.

marc-hb:
Found in
- #8595
- https://github.com/thesofproject/sof/pull/8514/files#r1404344228

Should probably be done before zephyr cleanup:
- #8260

kv2019i:
Stable-v2.9 branched, this didn't make the cut, bumping to 2.10.

kv2019i:
Stable branched for 2.10, so moving this to 2.11.

kv2019i:
Feature cutoff for v2.12, moving this to v2.13.

kv2019i:
https://github.com/thesofproject/sof/issues/8260 was completed for v2.13, but this won't be ready. Moving to v2.14.

PR #8791, opened 2024-01-25
Audio: Fix the kconfig error of mixin_mixout.

andrula-song:
The original content of mixin_mixout kconfig is incorrect, that is for aria. This commit fix that error.


src/audio/mixin_mixout/Kconfig
```diff
@@ -1,12 +1,8 @@
...
-        bool "ARIA component"
-        default n
-        depends on IPC_MAJOR_4
-        help
-          Select for Automatic Regressive Input Amplifier Module component
-	  ARIA applies variable gain into incoming signal.
-	  Applied gain is in range <1, 2 power attenuation>
-	  Currently ARIA introduces gain transition and algorithmic
-	  latency equal to 1 ms.
+config COMP_MIXER
```
kv2019i, line 3: Hmm, it doesn't seem correct to have two COMP_MIXER definitions with different semantics. This works now as Zephyr and XTOS builds use different Kconfigs (and this is getting slowly fixed in #8606 ), but this seems very confusing. Maybe better to just fix the extra ARIA defintiion and leave COMP_MIXER intact.

kv2019i (dismissed):
Aria change looks good, but I wonder if COMP_MIXER just causes more room for confusion.  @marc-hb please check

marc-hb (changes requested):
First, add a `Fixes commit abcdef ("...")` line so we know for how long this didn't work.

> I also don't think it would be a good choice to use 2 COMP_MIXER definitions, 

Why do we need two different directories?

- Either the IPC3 "mixer" and the IPC4 "mixin_mixout" are significantly different features that need a different name. Then don't add a new IPC4 COMP_MIXER. Instead a new, IPC4 `COMP_MIXIN_MIXOUT` option. Keep the single IPC3 COMP_MIXER.

- Or, the IPC3 "mixer" and the IPC4 "mixin_mixout" are the same feature. Then merge the two directories with a `git mv  mixin_mixout/*  mixer/`. Single COMP_MIXER for both in the same directory. You can preserve the filenames, they're not part of the Kconfig user interface.

> but don't know which solution is better.

I don't have any opinion as long as the Kconfig names are consistent with the directory names.

How was this done for other components?

marc-hb:
> The original content of mixin_mixout kconfig is incorrect, that is for aria.

BTW it's not obvious what "that" is in this sentence. I suspect this contributed to Kai's confusion.

>  This commit fix that error.

Please use a basic grammar checker, it helps too. There seems to be one embedded in everywhere nowadays.

So I think you meant this:
> Due to a copy/paste error in commit abcdedf ("..."), the content of `mixin_mixout/Kconfig` has been a copy of `aria/Kconfig` for the last 2 months.
> ...

The mixer seems like a pretty important component. So what have been the effects of this error for the last 2 months? The consequences (big or small) must be described briefly in the commit message. Or even described in a Github issue if the problem is too complex for a commit message (hopefully not this time).

lgirdwood (approved):

andrula-song:
hi, @kv2019i  @marc-hb , can you help to review?

zephyr/CMakeLists.txt
```diff
@@ -551,19 +551,17 @@ zephyr_library_sources_ifdef(CONFIG_COMP_KPB
...
+zephyr_library_sources_ifdef(CONFIG_COMP_MIXER
+	${SOF_AUDIO_PATH}/mixer/mixer.c
+	${SOF_AUDIO_PATH}/mixer/mixer_generic.c
+	${SOF_AUDIO_PATH}/mixer/mixer_hifi3.c
+)
+
+zephyr_library_sources_ifdef(CONFIG_COMP_MIXIN_MIXOUT
+	${SOF_AUDIO_PATH}/mixin_mixout/mixin_mixout.c
+	${SOF_AUDIO_PATH}/mixin_mixout/mixin_mixout_generic.c
+	${SOF_AUDIO_PATH}/mixin_mixout/mixin_mixout_hifi3.c
```
marc-hb, line 563: This makes the conditionals totally different from `src/audio/CMakeLists.txt` ?!?

marc-hb (changes requested):
Sorry I missed your force-pushed. Also, you didn't update the description which didn't help.

So aceef1a153255e77 was initially  was a minor Kconfig fix without a bug description.

Now there is a commit message in 9633be79e8cda641b64 that explains the change but:
1. Still no explanation of how this was broken for 2 months without anyone noticing. Is anyone using this code? 

2. Much more importantly, commit 9633be79e8 now makes a new, BIG logical change: the CMake conditionals totally new and totally different from `src/audio/CMakeLists.txt` - without a single word of CMake explanation.

What is going on?

zephyr/CMakeLists.txt
andrula-song: we use mmixin_mixout code to achieve the mixing function for IPC4 version and mixer code to achieve the mixing function for IPC3, but the infrastructure and implementation and even ability are very different for those 2 components.

> Either the IPC3 "mixer" and the IPC4 "mixin_mixout" are significantly different features that need a different name. Then don't add a new IPC4 COMP_MIXER. Instead a new, IPC4 COMP_MIXIN_MIXOUT option. Keep the single IPC3 COMP_MIXER.

so I think we should better use 2 different kconfig name with the dependence of IPC version for them, and since I move the IPC version dependence into kconfig, we have to change the src/audio/CMakeLists.txt to fit the kconfig modification.

marc-hb: Great explanations, thank you! They _really_ deserve to be in the commit message.

> and since I move the IPC version dependence into kconfig, we have to change the src/audio/CMakeLists.txt to fit the kconfig modification.

... and as usual in these situations, the problem is: how are you going to TEST all these CMake changes? This requires compiling in many different ways.

Always remember: break it before you fix it. It's very common to edit some of our CMake files while accidentally NOT testing the changes because they're in a different use case or configuration than the current one. If you break it and nothing happens, then you're not testing it.

marc-hb: Here's another very useful test trick in C files:

```C
#warning this volume Hifi7 file is being compiled
```

`#error` works too.

marc-hb:
> we use mmixin_mixout code to achieve the mixing function for IPC4 version and mixer code to achieve the mixing function for IPC3

So once again: how come no one noticed this Kconfig mistake for 2 months?

zephyr/CMakeLists.txt
andrula-song: actually even with the copy/past error, we won't found that error because the COMP_MIXER is default yes, we will always compile the mixin_mixout or mixer code. I noticed this because I want to change this to new hifi config style.

marc-hb:
Last week was all firefighting and Monday was off. I promise I'll look at this again Tuesday first thing.

marc-hb (dismissed):
- The code change looks good but it is incomplete: please align `src/audio/CMakeLists.txt` not to ignore your new `CONFIG_COMP_*`. As already requested by @kv2019i and myself 3 weeks ago, don't leave `src/audio/CMakeLists.txt` inconsistent with `zephyr/CMakeLists.txt`

You can easily test `src/audio/CMakeLists.txt` and Kconfig with commands like these (just an example)

```
./scripts/docker-run.sh ./scripts/xtensa-build-all.sh rmb
grep MIX build_rmb_gcc/generated/.config 
./scripts/docker-run.sh make -C build_rmb_gcc/ menuconfig
```

- The commit title is misleading, this is not just "fixing a kconfig error", this PR is significantly more than that! It's introducing a brand new `COMP_MIXIN_MIXOUT` that never existed before instead of relying directly on the IPC level. The error that is getting fixed does not even matter much because the fixed code was never used. So that really threw me off for a long time.

- The commit message is much better now but still too short. Just summarize the answers to the most important questions already discussed. in this PR and add that to the commit message.

- As requested earlier, Github does not automatically sync the PR title and PR message with the commit title and PR message. It would speed reviews a bit if you could do this manually.

- To make overbusy reviewers less confused and future reviews *much faster*, anticipate reviewers questions and answer them in advance in the commit message. This is the purpose of the commit message. https://wiki.openstack.org/wiki/GitCommitMessages

Anticipating reviewers' questions is especially faster when reviews happen across timezones with no possibility of interactive, real-time discussions. Ask teammates in your timezone for a quick review and answer their interactive questions in the commit message Use a metaphorical rubber duck for code reviews: https://en.wikipedia.org/wiki/Rubber_duck_debugging It's not just for debugging.

lgirdwood:
> * The code change looks good but it is incomplete: please align `src/audio/CMakeLists.txt` not to ignore your new `CONFIG_COMP_*`. As already requested by @kv2019i and myself 3 weeks ago, don't leave `src/audio/CMakeLists.txt` inconsistent with `zephyr/CMakeLists.txt`

Lets fix this incrementally as the next step, it better to get this fix in rather than waiting longer.

PR #8795, opened 2024-01-25
Fix building of native loadable modules

softwarecki:
This PR reverts changes in:
    src/module/audio/sink_api.c
    src/module/audio/source_api.c
made by commit 12d958a2eb8e3fa2d7a534f1ab1b59bab2a1fb64. These changes made it impossible to build native loadable modules.

### DIFF
--- a/src/module/audio/sink_api.c
+++ b/src/module/audio/sink_api.c
@@ -5,7 +5,6 @@
 
 #include <module/audio/sink_api.h>
 #include <module/audio/audio_stream.h>
-#include <rtos/symbol.h>
 
 /* This file contains public sink API functions that were too large to mark is as inline. */
 
@@ -24,7 +23,6 @@ int sink_get_buffer(struct sof_sink *sink, size_t req_size,
 		sink->requested_write_frag_size = req_size;
 	return ret;
 }
-EXPORT_SYMBOL(sink_get_buffer);
 
 int sink_commit_buffer(struct sof_sink *sink, size_t commit_size)
 {
@@ -46,7 +44,6 @@ int sink_commit_buffer(struct sof_sink *sink, size_t commit_size)
 	sink->num_of_bytes_processed += commit_size;
 	return ret;
 }
-EXPORT_SYMBOL(sink_commit_buffer);
 
 int sink_set_frm_fmt(struct sof_sink *sink, enum sof_ipc_frame frame_fmt)
 {
--- a/src/module/audio/source_api.c
+++ b/src/module/audio/source_api.c
@@ -6,7 +6,6 @@
 
 #include <module/audio/source_api.h>
 #include <module/audio/audio_stream.h>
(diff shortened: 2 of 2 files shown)


pjdobrowolski (approved):

jxstelter (approved):

pjdobrowolski:
Native loadable modules are building without Zephyr and they are not finding zephyr headers under define.

pjdobrowolski:
You are right. That is our fault that building native modules is not in CI to block such changes however it blocks us to add it now.

kv2019i:
@pjdobrowolski wrote:
> Native loadable modules are building without Zephyr and they are not finding zephyr headers under define.

But this is not module code, this is basefw, which makes this confusing. If I build up-down-mixer with lmdk, these files dont' need to be compiled, right? 

If this gets compiled with lmdk, toget forward, can you add a ifdef around these lines. ifdef LMDK build, do not export the symbols. Basefw should build and be free to use native Zephyr services.

lyakh:
> You are right. That is our fault that building native modules is not in CI to block such changes however it blocks us to add it now.

can you just put
```
#ifndef __SOF_MODULE_SERVICE_BUILD__
```
around those things and be done with it

lgirdwood:
Let me try and summarize - stand alone module build (without SOF base FW repo or Zephyr repo or maybe both ?) is failing as it has some undefined macros/functions ?
@softwarecki are you able to share the build output and build BKM.

lyakh:
> Let me try and summarize - stand alone module build (without SOF base FW repo or Zephyr repo or maybe both ?) is failing as it has some undefined macros/functions ? @softwarecki are you able to share the build output and build BKM.

@lgirdwood I don't think it depends on whether the build is stand-alone or with a base firmware build available. In fact I still think that isolated builds (not using a base FW build environment) aren't a good idea. But in any case, I think the reason here is simply, that "native system-service modules" link with these files, effectively they make a copy of them, in addition to those files being also linked with the base firmware.

softwarecki:
@lyakh @kv2019i @lgirdwood This files are shared between basefw and lmdk modules. Its contains common helper functions to handle sink/source interface.

softwarecki:
> can you just put
> 
> ```
> #ifndef __SOF_MODULE_SERVICE_BUILD__
> ```
> 
> around those things and be done with it

@lyakh can you create PR with it?

dbaluta:
@softwarecki  Can you please add better description to this patch. This needs to explain *WHY* previous patch needs to be reverted. What is the error and how does this patch fixes it.

lyakh:
> > can you just put
> > ```
> > #ifndef __SOF_MODULE_SERVICE_BUILD__
> > ```
> >   
> > around those things and be done with it
> 
> @lyakh can you create PR with it?

https://github.com/thesofproject/sof/pull/8812

PR #7959, opened 2023-07-18
IPC4: simplify pcm conversion function and add LSB & MSB support

RanderWang:
Currently cSOF only supports S24_LE, now add S24_BE to support sample type IPC4_TYPE_MSB_INTEGER

PR #8800, opened 2024-01-26
fixup! topology2: Use SAMPLE_TYPE_MSB_INTEGER for ALH copier

bardliao:
Some ALH DAI copier missed SAMPLE_TYPE_MSB_INTEGER setting.

### DIFF
--- a/tools/topology/topology2/platform/intel/sdw-amp-generic.conf
+++ b/tools/topology/topology2/platform/intel/sdw-amp-generic.conf
@@ -452,6 +452,8 @@ IncludeByKey.SDW_AMP_FEEDBACK {
 									in_valid_bit_depth	32
 									in_ch_cfg	$CHANNEL_CONFIG_3_POINT_1
 									in_ch_map	$CHANNEL_MAP_3_POINT_1
+									in_sample_type		$SAMPLE_TYPE_MSB_INTEGER
+									in_fmt_cfg		"$[($in_channels | ($in_valid_bit_depth * 256))]"
 								}
 							]
 							Object.Base.output_audio_format [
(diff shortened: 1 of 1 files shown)


plbossart (approved):

tools/topology/topology2/platform/intel/sdw-amp-generic.conf
```diff
@@ -461,6 +463,8 @@ IncludeByKey.SDW_AMP_FEEDBACK {
 									out_valid_bit_depth	32
 									out_ch_cfg	$CHANNEL_CONFIG_3_POINT_1
 									out_ch_map	$CHANNEL_MAP_3_POINT_1
+									out_sample_type		$SAMPLE_TYPE_MSB_INTEGER
+									out_fmt_cfg		"$[($out_channels | ($out_valid_bit_depth * 256))]"
```
kv2019i, line 467: This should have no practical impact given out_valid_bit_depth is 32. The bit-alignment only takes effect when valid-bits is smaller than the container size.

kv2019i (approved):
This should have no functional impact (see inline comment), but +1 to keep consistency with other definitions in the file.

tools/topology/topology2/platform/intel/sdw-amp-generic.conf
bardliao: > This should have no practical impact given out_valid_bit_depth is 32. The bit-alignment only takes effect when valid-bits is smaller than the container size.

Right, but it is better to be consistent. Other ALH dais set SAMPLE_TYPE_MSB_INTEGER no matter the valid_bit_depth is 32 bit or not. :)

lgirdwood (approved):
@bardliao can you followup with some inline topology comments here explaining the logic. We can merge today as this fix unblocks a lot of other tests.

bardliao:
> @bardliao can you followup with some inline topology comments here explaining the logic. We can merge today as this fix unblocks a lot of other tests.

@lgirdwood Sorry, I didn't get your comment. The PR is a fixup of https://github.com/thesofproject/sof/commit/2541e55f2fb534bc1eac37b686767edfae12ca8a that is a commit of https://github.com/thesofproject/sof/pull/8359. The request is that we should set SAMPLE_TYPE_MSB_INTEGER for 24 bit ALH DAIs. It is no impact if we set it for 16 or 32 bit format because the valid bits and the container bits are the same in the cases. However, we set SAMPLE_TYPE_MSB_INTEGER for all ALH DAIs for consistent. PR https://github.com/thesofproject/sof/pull/7959 may explain more.

PR #8804, opened 2024-01-26
[DNM] Zephyr update to 6a0b1da158a + WA to 8764

kv2019i:
https://github.com/thesofproject/sof/pull/8764 + SOF side workaround to fix the failures on cAVS2.5 (other attempts to solve this issue have failed this week)

The WA works in local test -- during 1h test, the failing condition was hit twice, and in both cases, the WA recovered. Let's see how this fares in CI with wider testing.

PR #8805, opened 2024-01-26
west.yml: update Zephyr (subset of PR8804 , no SMP changes)

kv2019i:


kv2019i:
One configuration was missed due to missing, but results are good:
https://sof-ci.01.org/sofpr/PR8805/build2381/devicetest/index.html
https://sof-ci.01.org/sofpr/PR8805/build2380/devicetest/index.html

The pause-resume fail seen with newer Zephyr revisions in https://github.com/thesofproject/sof/pull/8804 and https://github.com/thesofproject/sof/pull/8764 is not seen in results for this PR, so the the SMP interface change is causing the new failure on pause-resume.

src/audio/chain_dma.c
```diff
@@ -121,9 +121,28 @@ static int chain_host_stop(struct comp_dev *dev)
 	int err;
 
 	err = dma_stop(cd->chan_host->dma->z_dev, cd->chan_host->index);
-	if (err < 0)
+	if (err < 0) {
+		if (err == -EBUSY) {
+			int retries = 100;
```
lyakh, line 126: you probably want to fix the misplaced brace at line 143, and while at it (optionally) you could also consider doing
```
for (int i = 0; i < 100 && err == -EBUSY; i++) {
    if (!i)
        comp_warn(...);
    k_busy_wait(10);
    err = dma_stop(...);
}
if (err < 0)
    return err;
```

src/audio/chain_dma.c
```diff
@@ -121,9 +121,28 @@ static int chain_host_stop(struct comp_dev *dev)
...
+			comp_warn(dev, "dma_stop() fail chan_index = %u, retrying...",
+				  cd->chan_link->index);
+
+			/*
+			 * FIXME: ugly workaround for
+			 * https://github.com/thesofproject/sof/issues/8686
+			 */
+			while (err == -EBUSY && --retries) {
+				k_busy_wait(10);
+				err = dma_stop(cd->chan_link->dma->z_dev, cd->chan_link->index);
```
kv2019i, line 137: Actually, @ujfalusi noted this is likely not doing actual stop from second iterations onwards, so this is a no-op operation (with the HD-DMA driver). Let me revise the workaround to be more explicit...

kv2019i:
Workaround requires work, converting to a draft.

src/audio/chain_dma.c
ujfalusi: I guess this would be identical and more explicit:
```diff
diff --git a/src/audio/chain_dma.c b/src/audio/chain_dma.c
index 696973146404..cea2a2e5c8f3 100644
--- a/src/audio/chain_dma.c
+++ b/src/audio/chain_dma.c
@@ -121,7 +121,7 @@ static int chain_host_stop(struct comp_dev *dev)
 	int err;
 
 	err = dma_stop(cd->chan_host->dma->z_dev, cd->chan_host->index);
-	if (err < 0)
+	if (err < 0 && err != -EBUSY)
 		return err;
 
 	comp_info(dev, "chain_host_stop(): dma_stop() host chan_index = %u",
```
With added comment

PR #8806, opened 2024-01-26
cleanup: remove the need to use drivers/interrupt.h on some platforms

LaurentiuM1234:
The point of this patch series is to allow some Zephyr platforms which don't require `drivers/interrupt.h` to finally remove it


iuliana-prodan (approved):

kv2019i (approved):
This is nice, thanks @LaurentiuM1234 !

PR #8807, opened 2024-01-27
[SKIP SOF-TEST] .github: fallback on rebuild-testbench -j 1 when failing

marc-hb:
Parallel builds are fast but they bury errors, for instance good luck finding the cause of this failure:
 https://github.com/thesofproject/sof/actions/runs/7667629748/job/20897861756?pr=8019

Use the well known technique of falling back on a single thread when failing.

### DIFF
--- a/.github/workflows/testbench.yml
+++ b/.github/workflows/testbench.yml
@@ -49,7 +49,8 @@ jobs:
              ./scripts/build-tools.sh -t
 
       - name: build testbench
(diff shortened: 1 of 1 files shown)


lgirdwood (approved):
@kv2019i

marc-hb:
Another buried failure this PR would show:
https://github.com/thesofproject/sof/actions/runs/7693205487/job/20961494936?pr=8019

kv2019i (approved):

PR #8810, opened 2024-01-29
Audio: Aria: Use new ARIA HiFi config style  Use the new CONFIG_ARIA_HIFI style to choose right processing file.

andrula-song:
Use the new CONFIG_ARIA_HIFI style to choose right processing file.

### DIFF
--- a/src/audio/aria/aria.c
+++ b/src/audio/aria/aria.c
@@ -162,8 +162,10 @@ static void aria_set_stream_params(struct comp_buffer *buffer,
 	const struct ipc4_audio_format *audio_fmt = &mod->priv.cfg.base_cfg.audio_fmt;
 
 	ipc4_update_buffer_format(buffer, audio_fmt);
-#ifndef ARIA_GENERIC
+#ifdef ARIA_HIFI3
 	audio_stream_set_align(8, 1, &buffer->stream);
+#elif ARIA_HIFI5
+	audio_stream_set_align(16, 1, &buffer->stream);
 #endif
 }
 
--- a/src/audio/aria/aria.c
+++ b/src/audio/aria/aria.c
@@ -162,9 +162,9 @@ static void aria_set_stream_params(struct comp_buffer *buffer,
 	const struct ipc4_audio_format *audio_fmt = &mod->priv.cfg.base_cfg.audio_fmt;
 
 	ipc4_update_buffer_format(buffer, audio_fmt);
-#ifdef ARIA_HIFI3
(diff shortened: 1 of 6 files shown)


lgirdwood (commented):
@andrula-song I think the 1st patch probably needs to go in with the Aria Hifi5 PR if not merged.

src/audio/aria/aria.c
```diff
@@ -162,8 +162,10 @@ static void aria_set_stream_params(struct comp_buffer *buffer,
 	const struct ipc4_audio_format *audio_fmt = &mod->priv.cfg.base_cfg.audio_fmt;
 
 	ipc4_update_buffer_format(buffer, audio_fmt);
-#ifndef ARIA_GENERIC
+#ifdef ARIA_HIFI3
 	audio_stream_set_align(8, 1, &buffer->stream);
+#elif ARIA_HIFI5
```
marc-hb, line 167: Same question as in https://github.com/thesofproject/sof/pull/8682/files#r1469048792 : where and how are these defined?

andrula-song: the ARIA_HIFI5 is defined in aria.h (https://github.com/thesofproject/sof/pull/8692) while adding HiFi5 implementation.

andrula-song:
> @andrula-song I think the 1st patch probably needs to go in with the Aria Hifi5 PR if not merged.

the PR for for ARIA HiFi5 implementation has been merged

src/audio/aria/aria.c
marc-hb: I was confused because 307921c7ac24 was using the new way but its parent was still using the old way. Resolved, everything new now.

src/audio/aria/aria.c
```diff
@@ -162,9 +162,9 @@ static void aria_set_stream_params(struct comp_buffer *buffer,
 	const struct ipc4_audio_format *audio_fmt = &mod->priv.cfg.base_cfg.audio_fmt;
 
 	ipc4_update_buffer_format(buffer, audio_fmt);
-#ifdef ARIA_HIFI3
+#if SOF_USE_HIFI(3, ARIA) || SOF_USE_HIFI(4, ARIA)
```
btian1, line 165: patch1 can be squashed into patch2?

andrula-song: I don't think it is a good idea, patch 1 adds the align limit for hifi5, patch 2 is for new style, they are totally different things

btian1 (approved):

lgirdwood (approved):

PR #8811, opened 2024-01-29
Tools: Testbench: Convert file component to module adapter

singalsu:
This is done as preparation for testbench IPC4 support. The update to IPC4 is simpler for a module adapter component.

### DIFF
--- a/src/include/ipc/topology.h
+++ b/src/include/ipc/topology.h
@@ -240,8 +240,16 @@ struct sof_ipc_comp_process {
 
 /* IPC file component used by testbench only */
 struct sof_ipc_comp_file {
+	/* These need to be the same as in above sof_ipc_comp_process */
 	struct sof_ipc_comp comp;
 	struct sof_ipc_comp_config config;
+	uint32_t size;	/**< size of bespoke data section in bytes */
+	uint32_t type;	/**< sof_ipc_process_type */
+
+	/* reserved for future use */
+	uint32_t reserved[7];
+
+	/* These are additional parameters for file */
 	uint32_t rate;
 	uint32_t channels;
 	char *fn;
--- a/tools/testbench/common_test.c
+++ b/tools/testbench/common_test.c
@@ -1,33 +1,35 @@
 // SPDX-License-Identifier: BSD-3-Clause
 //
-// Copyright(c) 2018 Intel Corporation. All rights reserved.
+// Copyright(c) 2018-2024 Intel Corporation. All rights reserved.
 
-#include <stdint.h>
-#include <stddef.h>
-#include <time.h>
-#include <stdio.h>
-#include <stdlib.h>
-#include <rtos/string.h>
-#include <math.h>
-#include <rtos/sof.h>
-#include <rtos/task.h>
-#include <rtos/alloc.h>
-#include <sof/lib/notifier.h>
+#include <sof/audio/component_ext.h>
+#include <sof/audio/pipeline.h>
 #include <sof/ipc/driver.h>
 #include <sof/ipc/topology.h>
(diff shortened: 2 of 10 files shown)


src/include/ipc/topology.h
```diff
@@ -240,8 +240,16 @@ struct sof_ipc_comp_process {
...
 /* IPC file component used by testbench only */
 struct sof_ipc_comp_file {
+	/* These need to be the same as in above sof_ipc_comp_process */
 	struct sof_ipc_comp comp;
 	struct sof_ipc_comp_config config;
+	uint32_t size;	/**< size of bespoke data section in bytes */
+	uint32_t type;	/**< sof_ipc_process_type */
+
+	/* reserved for future use */
+	uint32_t reserved[7];
```
btian1, line 250: I would suggest put reserved to the bottom of struct definition.

tools/testbench/common_test.c
```diff
@@ -1,33 +1,35 @@
...
 #include <sof/schedule/ll_schedule_domain.h>
 #include <sof/schedule/schedule.h>
+#include <rtos/alloc.h>
+#include <rtos/sof.h>
+#include <rtos/string.h>
+#include <rtos/task.h>
 #include <rtos/wait.h>
-#include <sof/audio/pipeline.h>
-#include <sof/audio/component_ext.h>
+#include <tplg_parser/topology.h>
```
btian1, line 22: 1. maybe it is a good time to check whether all the headers are all must to put it here.
2. suggest a sequence for header include, from long to short.

tools/testbench/topology.c
```diff
@@ -369,23 +379,33 @@ static int tb_register_fileread(struct testbench_prm *tp,
...
+	file_uuid->b = 0x75aa;
+	file_uuid->c = 0x4ce8;
+	file_uuid->d[0] = 0x9d;
+	file_uuid->d[1] = 0xbe;
+	file_uuid->d[2] = 0xd8;
+	file_uuid->d[3] = 0xda;
+	file_uuid->d[4] = 0x08;
+	file_uuid->d[5] = 0xa6;
+	file_uuid->d[6] = 0x98;
+	file_uuid->d[7] = 0xc2;
```
btian1, line 398: how about a strcpy to replace above?

btian1 (commented):
one question: will testbench support IPC3 and IPC4 both or after change, testbench only support IPC4?

src/include/ipc/topology.h
singalsu: I did it this way since currently module adapter defaults to process type IPC unless I add exception handling to module adapter.  The extra IPC data for file component is passed in the end of this struct.

I think I will try to another approach with CONFIG_LIBRARY exceptions added for SOF_COMP_FILEREAD and SOF_COMP_FILEWRITE. Not sure which is better.

Also this will be bit different with IPC4. I'm keeping this draft until I know better.

tools/testbench/common_test.c
singalsu: I'm following a rule for alphabetic sort for 4 levels, 3 levels, 2 levels, standard C headers last. I got such suggestion some time ago in review. Though these could go to other cosmetic changes patch.

tools/testbench/topology.c
singalsu: Yes, could be better, or at least better readable.

singalsu:
> one question: will testbench support IPC3 and IPC4 both or after change, testbench only support IPC4?

I think there need to be separate executable build for IPC versions. The SOF headers are constructed in such way that I can't have both in the same time.

singalsu:
Closing, this change is part of #9025

PR #8812, opened 2024-01-29
lmdk: fix building system-service modules

lyakh:
System-service modules cannot include Zephyr headers and don't use exported symbols, they link sink_api.c and source_api.c into their images. Disable symbol exporting for them.

### DIFF
--- a/src/module/audio/sink_api.c
+++ b/src/module/audio/sink_api.c
@@ -5,7 +5,17 @@
 
 #include <module/audio/sink_api.h>
 #include <module/audio/audio_stream.h>
+/*
+ * When building native system-service modules only exported module headers can
+ * be included, autoconf.h isn't included either. To identify such a build we
+ * need any symbol that is guaranteed to be defined with any SOF build.
+ * CONFIG_CORE_COUNT is such a symbol, it is always defined to an integer number
+ */
+#ifndef CONFIG_CORE_COUNT
+#define EXPORT_SYMBOL(...)
+#else
 #include <rtos/symbol.h>
+#endif
 
 /* This file contains public sink API functions that were too large to mark is as inline. */
 
--- a/src/module/audio/source_api.c
+++ b/src/module/audio/source_api.c
@@ -3,10 +3,19 @@
  * Copyright(c) 2023 Intel Corporation. All rights reserved.
  */
 
-
 #include <module/audio/source_api.h>
 #include <module/audio/audio_stream.h>
+/*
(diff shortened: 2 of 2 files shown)


kv2019i (approved):

softwarecki (dismissed):

softwarecki (changes requested):
This PR doesn't solve the problem. The `__SOF_MODULE_SERVICE_BUILD__` symbol is defined only in one example module. We need more universal solution.

lyakh:
> This PR doesn't solve the problem. The `__SOF_MODULE_SERVICE_BUILD__` symbol is defined only in one example module. We need more universal solution.

@softwarecki you don't use it when building LMDK modules? Ok, I thought you did. Is there anything that you do define for all system-service modules?

softwarecki:
> Ok, I thought you did. Is there anything that you do define for all system-service modules?

We don't define it in our examples of lmdk modules. Please check files in `lmdk/cmake` or use some definition used by llmodules.

lgirdwood:
@softwarecki btw, why does system services build and link to a base FW file (e.fg the source/sink API) ? Shouldn't these be unresolved in the library and then be resolved when the library is loaded ?

src/module/audio/sink_api.c
```diff
@@ -5,7 +5,11 @@
 
 #include <module/audio/sink_api.h>
 #include <module/audio/audio_stream.h>
+#ifdef MAJOR_IADSP_API_VERSION
+#define EXPORT_SYMBOL(...)
+#else
 #include <rtos/symbol.h>
+#endif
```
marc-hb, line 12: I find this not obvious at all so I think most the commit message should be moved to a comment here.

softwarecki (approved):
This is a very surprising solution, but it works so I accept it.

lyakh:
I think we can safely state, that this PR cannot be a reason for all the CI failures:
- cAVS suspend-resume https://sof-ci.01.org/sofpr/PR8812/build2551/devicetest/index.html
- ACE multiple pipelines and multiple pause-resume with nocodec https://sof-ci.01.org/sofpr/PR8812/build2550/devicetest/index.html
- Broken fuzzing is a recent regression
- Failing zmain builds are a known Zephyr breakage too

marc-hb:
> This is a very surprising solution, but it works so I accept it.

Moreover it's short, well documented and self-contained in only 2 places to change if needed in the future. So very low technical debt if any!

pjdobrowolski (approved):

PR #8815, opened 2024-01-29
topology2: Add topology for testing SOF plugin

ranj063:
The topology uses passthrough pipelines for playback/capture and supports only 48k, 2ch, 16bit format for now.

### DIFF
--- a/tools/plugin/README.md
+++ b/tools/plugin/README.md
@@ -41,16 +41,16 @@ Ex: "~/work/sof/sof/build_plugin/sof_ep/install/lib:~/work/sof/sof/build_plugin/
 And set the environment variable SOF_PLUGIN_TOPOLOGY_PATH to point to the directory containing the topology binary
 ```
 
-Code can then be run by starting sof-pipe with your desired topology
+Code can then be run by starting sof-pipe with your desired topology (Ex: sof-plugin.tplg)
 
 ```
- ./sof-pipe -T sof-tgl-nocodec.tplg
+ ./sof-pipe -T sof-plugin.tplg
 ```
 
 At this point the sof-pipe daemon is waiting for IPC. Audio applications can now invoke sof-pipe processing via
 
 ```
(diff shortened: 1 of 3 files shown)


tools/plugin/README.md
```diff
@@ -66,9 +66,9 @@ The above example needs to be 48k as example pipe has no SRC/ASRC.
 Likewise
 
 ```
-arecord -Dsof:tgl-nocodec:1:default:default:48k2c16b -f dat file.wav
+arecord -Dsof:plugin:1:default:default:48k2c16b -f dat file.wav
```
plbossart, line 69: it:would:not:hurt:to:document:all:these:options:never:seen:this:before

tools/topology/topology2/development/sof-plugin.conf
```diff
@@ -0,0 +1,170 @@
...
+<output_audio_format.conf>
+
+Object.Pipeline.io-gateway-capture [
+	{
+		index		10
+		direction	capture
+
+		Object.Widget.dai-copier."1" {
+			dai_index	1
+			dai_type	"SSP"
```
plbossart, line 36: does the concept of SSP make any sense for a plugin at the ALSA-lib level? Why is this needed?

plbossart (commented):
I have zero background on the plugin....

tools/plugin/README.md
ranj063: @plbossart it is documented in the plugin usage help message and in the same README file below

tools/topology/topology2/development/sof-plugin.conf
ranj063: it is not really needed for the plugin but these are mandatory attributes for a dai-copier and I cant create one without it. I will have to update the host/dai copier to remove these  but for now, we have no choice

lgirdwood (commented):
Plugin is still at the "staging" level, so we need to continue and capture our TODOs e.g. needing SSP for copier is a new TODO

singalsu (approved):
Thanks, really useful to have a working example to start other topologies plugin development.

PR #8822, opened 2024-01-31
Trace: Remove ctx dependency from current logging system

btian1:
This is a preparation for SOF logging to zephyr interface, IPC3 logging need context support, it is not must, and verified with #8725 , so remove these dependencies and will use another new PR to make SOF with zephyr logging interface.

### DIFF
--- a/posix/include/rtos/spinlock.h
+++ b/posix/include/rtos/spinlock.h
@@ -77,8 +77,6 @@ typedef uint32_t k_spinlock_key_t;
 extern uint32_t lock_dbg_atomic;
 extern uint32_t lock_dbg_user[DBG_LOCK_USERS];
 
-extern struct tr_ctx sl_tr;
-
 /* panic on deadlock */
 #define spin_try_lock_dbg(lock, line) \
 	do { \
@@ -115,14 +113,14 @@ extern struct tr_ctx sl_tr;
 
 #define spin_lock_dbg(line) \
 	do { \
-		tr_info(&sl_tr, "LcE"); \
-		tr_info(&sl_tr, "line: %d", line); \
+		tr_info("LcE"); \
+		tr_info("line: %d", line); \
 	} while (0)
 
 #define spin_unlock_dbg(line) \
 	do { \
-		tr_info(&sl_tr, "LcX"); \
-		tr_info(&sl_tr, "line: %d", line); \
+		tr_info("LcX"); \
+		tr_info("line: %d", line); \
 	} while (0)
 
 #else  /* CONFIG_DEBUG_LOCKS_VERBOSE */
--- a/posix/include/rtos/spinlock.h
+++ b/posix/include/rtos/spinlock.h
@@ -88,9 +88,9 @@ extern struct tr_ctx sl_tr;
 				break;	/* lock acquired */ \
 		} \
 		if (__tries == 0) { \
-			tr_err_atomic(&sl_tr, "DED"); \
-			tr_err_atomic(&sl_tr, "line: %d", line); \
-			tr_err_atomic(&sl_tr, "user: %d", (lock)->user); \
+			tr_err_atomic("DED"); \
+			tr_err_atomic("line: %d", line); \
+			tr_err_atomic("user: %d", (lock)->user); \
 			panic(SOF_IPC_PANIC_DEADLOCK); /* lock not acquired */ \
 		} \
 	} while (0)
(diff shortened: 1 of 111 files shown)


lyakh (commented):
the last commit in the series is huge, impossible to review (it seems trivial, but still would be good to at least have a chance), and has a high breakage probability - almost any other PR, that gets merged before this one would break it.
What I personally would do instead, I'd add a new interface (maybe `trace_*()` or `log_*()`) and gradually move files over to it - one subsystem at a time, one PR for IPC, one PR for schedulers, etc. Then after all are migrated, you just remove the `tr_*()` API

lgirdwood:
> the last commit in the series is huge, impossible to review (it seems trivial, but still would be good to at least have a chance), and has a high breakage probability - almost any other PR, that gets merged before this one would break it. What I personally would do instead, I'd add a new interface (maybe `trace_*()` or `log_*()`) and gradually move files over to it - one subsystem at a time, one PR for IPC, one PR for schedulers, etc. Then after all are migrated, you just remove the `tr_*()` API

This would simplify the merge, i.e. not a big bang and reduced change of conflict races with other PRs.
@btian1 one more thing, I would show the log output before/after the change in the PR introduction.

dbaluta:
Can you please try to address at least some of the checkpatch.pl errors:

for example this:

```
WARNING: macros should not use a trailing semicolon
#243: FILE: src/include/sof/trace/trace_nonzephyr.h:137:
+#define trace_check_size_uint32(a) \
+	CT_ASSERT(sizeof(a) <= sizeof(uint32_t), "error: trace argument is bigger than a uint32_t");

```

I think it is relevant.

btian1:
> Can you please try to address at least some of the checkpatch.pl errors:
> 
> for example this:
> 
> ```
> WARNING: macros should not use a trailing semicolon
> #243: FILE: src/include/sof/trace/trace_nonzephyr.h:137:
> +#define trace_check_size_uint32(a) \
> +	CT_ASSERT(sizeof(a) <= sizeof(uint32_t), "error: trace argument is bigger than a uint32_t");
> ```
> 
> I think it is relevant.

@dbaluta , this is keep as origin, I change nothing,  you can check:
https://github.com/thesofproject/sof/blob/main/src/include/sof/trace/trace.h#L272

still need help from you for IPC3 logging compare as Liam requested, it should be fine, I will try to finish this patch tomorrow.

dbaluta:
@btian1 Please let me know when the patches are in a final form and will try to give them a test

btian1:
> @btian1 Please let me know when the patches are in a final form and will try to give them a test



> @btian1 Please let me know when the patches are in a final form and will try to give them a test

Thanks, @dbaluta , could you try with #8725, 8725 is test patch, and it is ahead of #8822, if #8725 ok, I will submit more patches.

btian1:
@kv2019i @lgirdwood @lyakh @dbaluta , codestyle problem is origin issue, if we want to fix, we can fix with another PR, this PR does not change those code actually.

Regarding the fuzzy error, it may related with #8632 patch merge, please notice.

btian1:
@dbaluta , could you try with this PR for IPC3 logging? I don't want make pause here due to CNY holiday.

dbaluta:
@LaurentiuM1234 can you please try to see if logging still works after this PR?

LaurentiuM1234:
> @LaurentiuM1234 can you please try to see if logging still works after this PR?
> 
> @btian1 can you please confirm the testing conditions? Is it with XTOS right?

tested on QM with Zephyr hybrid build and all's good. For reference, I'm attaching a snippet of how the logs look like w/ this patch. I'm assuming this behaves the same as XTOS since they use the same infrastructure.

Screenshot from 2024-02-07 14-15-48

dbaluta:
@LaurentiuM1234 can you post a snippet of the log w/o the patch. It looks like now the component is shown as "Unknown". I wonder if w/o the patch the real component is correctly shown or there is an earlier bug.

LaurentiuM1234:
> @LaurentiuM1234 can you post a snippet of the log w/o the patch. It looks like now the component is shown as "Unknown". I wonder if w/o the patch the real component is correctly shown or there is an earlier bug.

sorry for the long delay. Attaching snippet (8QM, WM8690, SOF main (557e581ca8a607ace0833b30c3ec5e0c87471ff4)):
Screenshot from 2024-02-12 13-49-09

dbaluta:
@btian1 looks like with your patch we lose ability to get the component for which the log was generated. Is this intended right?

If so then maybe we can remove that column from sof-logger output. 

What's your take on this?

btian1:
> then maybe we can remove that column from sof-logger output.
> 
> What's your take on this?

Thanks, @dbaluta @LaurentiuM1234 , due to UID was removed, then seems sof-logger unable to find each component name for logging.

So, if you are ok, we can modify sof-logger to remove this column, or keep as current until IPC3 support ended? @lgirdwood

